PR #40740, opened 2025-10-20
pane: Ignore max tabs on terminal pane

### LINKED ISSUE #39901: Unable to open multiple terminal tabs
### Summary

After opening the terminal it is not possible to open a secondary terminal using the "+" button. It seems to relegate the previous tab to the background, and unable to access any processes that were running there. Not sure if they are cancelled or keep running.

### Description

DESCRIPTION_HERE

Steps to reproduce:
1. Open terminal
2. click the plus button, select new terminal
3. notice that a new terminal tab is created and the original one disappears.

**Expected Behavior**: would like to have multiple independent terminal tabs, without being forced to split the terminal view
**Actual Behavior**: see above

Zed: v0.207.4 (Zed) 
OS: macOS 26.0.1
Memory: 64 GiB
Architecture: aarch64

vipexv:
Closes #39901

I'm unsure as to which direction the team wants to go with this, but this is the behavior of VSCode which is what this feature is based off so i'm going with this.

Changes: 

1. Introduced a new argument to the `new` method on the Pane called `ignore_max_tabs` that forces the `max_tabs` to None if it's true.
2. Added a new test `test_bypass_max_tabs_limit`.

Release Notes:

- Fixed: `max_tabs` Setting affecting the terminal pane.


JosephTLyons:
We discussed adding a test as well to make sure we can open `max_tabs` + 1  terminal tabs, but if the terminal testing infra isn't there, in the same way it is in pane.rs, we can skip.

JosephTLyons:
I pushed a few small changes:

1. I renamed `ignore_max_tabs` to `use_max_tabs`, to be affirmative (just a personal pref), and also flipped the bools at the call sites.
2. Moved `use_max_tabs` up in the parameter list, as we tend to always end with `window` and `cx`, in methods that use those parameters, just an internal convention.

PR #40746, opened 2025-10-20
project_panel: Don’t refocus on rename blur, only refocus on explicit confirm

### LINKED ISSUE #37726: Clicking buffer does not focus it when renaming in Project Panel
### Summary

When renaming something in the Project Panel, clicking a buffer does not focus it

### Description

Steps to reproduce:
1. In the Project Panel, click+Enter or click+F2 to rename a file/folder
2. Click a buffer

**Expected Behavior**: The rename is cancelled and the buffer becomes focused
**Actual Behavior**: The rename is cancelled, but the buffer focuses for a split second and then focus goes back to the Project Panel

Zed: v0.202.7 (Zed) 
OS: macOS 15.6.1
Memory: 16 GiB
Architecture: x86_64

SO9010:
Closes #37726

Release Notes:

- Added a condition to the confirm function to stop refocusing clicks out to the buffer, or there has been a focus change.

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -643,7 +643,7 @@ impl ProjectPanel {
                             .as_ref()
                             .is_some_and(|state| state.processing_filename.is_none())
                         {
-                            match project_panel.confirm_edit(window, cx) {
+                            match project_panel.confirm_edit(false, window, cx) {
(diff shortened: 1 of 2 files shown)


SO9010:
Oh wait i realised this has the wrong functionality clicking out should cancel the action!

PR #40747, opened 2025-10-20
project_panel: Fix buffer focus when canceling filename edit

### LINKED ISSUE #37726: Clicking buffer does not focus it when renaming in Project Panel

SO9010:
Closes #37726

Release Notes:

- Fixed an issue where the buffer would not regain focus when clicked while a filename edit was in progress in the project panel.


smitbarmase (changes requested):
I think we should default to confirm as same as VSCode. This was recently changed in https://github.com/zed-industries/zed/pull/39716.

SO9010:
Got it sounds good!

SO9010:
Alright that should have sorted that, I had a draft where it confirmed rather than cancelled. I didn't quite know what to call the pr so please feel free to change that!

smitbarmase (approved):
Looks great! I removed the global focus out handling since the filename editor’s blur event already covers confirm behavior, making that redundant. Thanks for working on this.

SO9010:
Awesome thank you so much! I am very happy to start my contributions to zed!!

PR #40750, opened 2025-10-20
Fix enum variant names in BaseKeymap

warrenjokinen:
Updated enum variant names for consistency.

Two product names with minor typos.

Closes #ISSUE

Release Notes:

- N/A


maxdeviant:
Thank you for the PR, but these are not valid enum variant names.

PR #40751, opened 2025-10-20
settings_ui: Correct stepper increment and enforce max value for Centered Layout Padding

### LINKED ISSUE #40748: settings_ui: Incorrect stepper increment and invalid value acceptance for entered Layout Padding (left/right)
### Summary

Broken stepper for centered layout padding 

### Description
The Centered Layout Padding setting, which adjusts the left and right padding, has an allowed value range between 0.0 (min) and 0.4 (max).

1. The stepper buttons (increment/decrement) for this number field are set to adjust the value by ±0.2 with each click.
2. The field incorrectly allows values greater than the 0.4 maximum to be entered and stored.

**Expected Behavior**:
- The stepper increment is 0.2. This makes precise adjustments difficult, as only three values (0.0, 0.2, 0.4) are selectable using the stepper.
- Values above 0.4 (e.g., 0.5, 1.0) can be manually entered and saved, even though they are outside the valid range.

**Actual Behavior**:
- The stepper increment should be reduced to a smaller value, such as 0.05, to allow for finer control.
- The field should validate input and prevent any value greater than the 0.4 maximum from being stored.

Zed: v0.208.5 (Zed) 
OS: macOS 26.0.1
Memory: 24 GiB
Architecture: aarch64

delvin02:
Closes #40748

Release Notes:

- Improved: The stepper increment is 0.05 for centered padding left/right
- Fixed: The value range in the settings UI is now limited between a minimum of 0.00 and a maximum of 0.40


Anthony-Eid:
Thank you for submitting this PR. I pushed up some commits that organized the code a bit more, removed an unnecessary unwrap, and improved the generated JSON schema to show the default value in the completion menu if a user is editing the raw JSON manually.

PR #40752, opened 2025-10-21
Prevent accidental link activation when dragging

### LINKED ISSUE #39772: Windows Beta: <a short description of the Windows bug>
### Summary

Unexpected navigation triggered when releasing mouse after Ctrl + X drag

### Description

Steps to trigger the problem:
1. Hold down the Control key and select something with the left mouse button.
2. Without releasing the mouse button or the Control key, press **Ctrl + X**.
3. While holding down the Control key and the left mouse button, drag the mouse to a function or interface.
4. Release only the left mouse button.

### Video Example:

https://github.com/user-attachments/assets/f6d00181-f40c-407b-b0a0-7323866741c3

**Expected Behavior**:
Releasing the mouse button over a function after performing the previous actions should not trigger navigation. It should be treated as part of the previous operation (dragging or cutting), not as a separate click event.

**Actual Behavior**:
When you release the left mouse button on a function, the editor behaves as if you had pressed Ctrl + click on that function.
Instead of simply ending the drag or cut action, it immediately navigates to the definition or reference of the function, as if you had intentionally clicked on it.
This happens in Go, as you can see in the video, and in other languages such as TS/Svelte.

Zed Nightly 0.208.0 
4ec24ebe01bdfca9c443ed348506e937f253c552

OS: Windows 10 - Atlas OS
CPU: Ryzen 7 5700X3D
RAM: 32GB
GPU: RTX 3070

ruxwez:
Closes #39772

Release Notes:

- Added a check to detect when the mouse was dragged by measuring the movement between press and release. Now, a link will only be opened if there was no drag, preventing accidental clicks while dragging. (`crates/editor/src/element.rs`)

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -985,10 +985,17 @@ impl EditorElement {
 
         let hovered_link_modifier = Editor::multi_cursor_modifier(false, &event.modifiers(), cx);
 
+        let was_drag = if let ClickEvent::Mouse(mouse_event) = event {
+            (mouse_event.down.position - mouse_event.up.position).magnitude() > 1.0
+        } else {
+            false
+        };
+
         if let Some(mouse_position) = event.mouse_position()
             && !pending_nonempty_selections
             && hovered_link_modifier
             && text_hitbox.is_hovered(window)
+            && !was_drag
         {
             let point = position_map.point_for_position(mouse_position);
             editor.handle_click_hovered_link(point, event.modifiers(), window, cx);


crates/editor/src/element.rs
```diff
@@ -985,10 +985,17 @@ impl EditorElement {
 
         let hovered_link_modifier = Editor::multi_cursor_modifier(false, &event.modifiers(), cx);
 
+        let was_drag = if let ClickEvent::Mouse(mouse_event) = event {
+            (mouse_event.down.position - mouse_event.up.position).magnitude() > 1.0
+        } else {
+            false
+        };
```
Veykril, line 992: Let's move this to a method on `ClickEvent` (with a doc comment). This seems like a generally useful tool to me I think and its better to not have the magnitude constant duplicated in various places if we do end up using this logic more elsewhere

ruxwez: Sounds good to me. I've been trying to figure out the logic behind the code, and I've had a hard time finding the functionality.

I don't know if it already exists, but it would be nice to have a guide that explains what each crate does so you can quickly find features.

PR #39989, opened 2025-10-11
Add OpenAI Responses API support with chat_completions capability flag

mstallone:
Add support for OpenAI's /responses endpoint for models that don't support  /chat/completions API. This enables compatibility with newer model variants (`gpt-5-codex`, `gpt-5-pro`, `o3-pro`, etc) while maintaining compatibility with existing configs

Changes:
- Add `supports_chat_completions` flag to model capabilities that defaults to true for existing behavior 
- Implement responses API client with streaming support as per [OpenAI documentation](https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml).
- Add `ResponseEventMapper` to convert responses events to completion events for maintainer simplicity 
- Update UI to allow toggling `chat_completions` capability
- Add `gpt-5-codex` model

Closes #38858

Release Notes:
- Added support for OpenAI's /responses endpoint for models that don't support  /chat/completions API.
- Added OpenAI `gpt-5-codex` model to assistant

PR #40762, opened 2025-10-21
Add support for GitHub Copilot /responses endpoint

somtooo:
Add support for GithubCopilot /responses endpoint. This gives the copilot chat provider the ability to use the new GPT-5 codex model and any other model that lacks support for /chat/copmletions endpoint.

Closes #38858 

Release Notes:

- Add support for GithubCopilot /responses endpoint.

# Added
1. copilot_response.rs that has the /response endpoint types
2. uses response endpoint if model does not support /chat/completions.
3. new into_copilot_response() to map LanguageCompletionEvents to Request.
4. new map_stream() to map response stream event to LanguageCompletionEvents and tests.
5. Fixed a bug where trying to parse response for non streaming for /chat/completion was failing

# Notes
There is a pr open - https://github.com/zed-industries/zed/pull/39989 for adding /response support for OpenAi and OpenAi compatible API. Altough they share some similarities (copilot api seems to mirror  openAi directly) ive simplified some stuff and tried to keep it the same with the vscode-chat implementation where possible. There might be a case for code reuse but i think keeping them separate for now should be ok. 

# Tool Calls
Screenshot from 2025-10-15 17-12-30

# Image
Screenshot from 2025-10-21 02-02-26

### DIFF
--- a/crates/copilot/src/copilot.rs
+++ b/crates/copilot/src/copilot.rs
@@ -1,5 +1,6 @@
 pub mod copilot_chat;
 mod copilot_completion_provider;
+pub mod copilot_responses;
 pub mod request;
 mod sign_in;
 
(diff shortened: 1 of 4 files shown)


somtooo:
@ConradIrwin  tagging for feedback

ConradIrwin:
Amazing, thank you!

I tested this a bit and streaming/tool calls etc. seem to work. I like that you kept it isolated, so going to merge.

PR #40794, opened 2025-10-21
markdown: Don't adjust indentation when inserting with multiple cursors

### LINKED ISSUE #40757: Markdown: Zed should not adjust indentation when inserting text with multiple cursors
### Summary

Consider the following Markdown:

```md
- [ ] Item 1
    - [ ] Item 1.a
- [ ] Item 2
    - [ ] Item 2.a
    - [ ] Item 2.b
```

https://github.com/user-attachments/assets/09e40cd3-3944-449f-99f8-0e4f782921b3

We do not adjust indentation of these lines when editing individual lines with a single cursor. VS Code does not adjust formatting of these lines when using multiple cursors.

```
Zed: v0.210.0 (Zed Nightly 7c4fb5a899c34efaea9e52ddd84daebc8d9ccf49) 
OS: macOS 26.0
Memory: 64 GiB
Architecture: aarch64
```

vipexv:
Closes #40757

## Summary

This PR addresses an issue where Zed incorrectly adjusts the indentation of Markdown lists when inserting text using multiple cursors. Currently:

- Editing individual lines with a single cursor behaves correctly (no unwanted indentation changes).
- Using multiple cursors, Zed automatically adjusts the indentation, unlike VS Code, which preserves the existing formatting.

## Tasks
- [x] Implement a new test to verify correct Markdown indentation behavior with multiple cursors.
- [x] Apply the fix to prevent Zed from adjusting indentation when inserting text on multiple cursors.

------------------------

Release Notes:

- Fixed an issue where inserting text with multiple cursors inside a nested Markdown list would cause it to lose its indentation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5385,6 +5385,7 @@ dependencies = [
  "tree-sitter-bash",
  "tree-sitter-c",
  "tree-sitter-html",
+ "tree-sitter-md",
  "tree-sitter-python",
  "tree-sitter-rust",
  "tree-sitter-typescript",
--- a/crates/editor/Cargo.toml
+++ b/crates/editor/Cargo.toml
@@ -115,6 +115,7 @@ tree-sitter-rust.workspace = true
 tree-sitter-typescript.workspace = true
 tree-sitter-yaml.workspace = true
 tree-sitter-bash.workspace = true
+tree-sitter-md.workspace = true
 unindent.workspace = true
 util = { workspace = true, features = ["test-support"] }
 workspace = { workspace = true, features = ["test-support"] }
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -4097,8 +4097,6 @@ impl Editor {
         let snapshot = self.buffer.read(cx).read(cx);
         let mut clear_linked_edit_ranges = false;
 
-        let selections_len = selections.len();
-
         for (selection, autoclose_region) in
             self.selections_with_autoclose_regions(selections, &snapshot)
         {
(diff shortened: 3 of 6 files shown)


vipexv:
So far the fix seems to work, the tests also passed locally, but i'm not 100% sure, so i'll be waiting for a team member to take a look at this approach and see if there's any issues with it.

smitbarmase (changes requested):
I don't think it's the right approach to fix this bug, since it might affect a lot of languages. Let's not touch `handle_input` at all.  

I wonder if we can fix this by making the Markdown language aware of indents within it. Notice how Python and Bash languages are implemented. You might want to look at this file:  
https://github.com/zed-industries/zed/blob/72cde7f9e4793fa795fc149d9491d850529c9936/crates/languages/src/python/indents.scm#L5  

And these PRs: https://github.com/zed-industries/zed/pull/29625 and https://github.com/zed-industries/zed/pull/33370.  

I think the simplest solution would be defining `increase_indent_pattern` ([more here](https://github.com/zed-industries/zed/blob/72cde7f9e4793fa795fc149d9491d850529c9936/crates/language/src/language.rs#L708)). However, since a Markdown list item that’s a parent and a child look alike, I’m not sure if just this would work but it’s worth trying.  

Let me know if you’re up for giving this a try. I can help you land this.

vipexv:
I honestly thought the same thing, it's why i was skeptical of this approach, i know the tests pass but it could break allot of languages, especially modifying one of the core input functions.

And yeah, i'm 100% willing to give this a try, i'll be taking a deeper look into what you mentioned and move forward with that, thank you!

vipexv:
I added some basic indentation rules. This is my first time working with a schema file, so I’m still learning. I also experimented with defining both `increase_indent_pattern` and `decrease_indent_pattern`, i managed to resolved the main issue with that but introduced a few new ones I couldn’t figure out (didn't spend too much time on that approach). The current approach seems to work, though. I’d appreciate it if you could take a look and let me know if there's a better way to handle this!

vipexv:
Hello @smitbarmase!

I’ve updated the PR based on your feedback (moved the logic to Markdown’s indent rules and reverted the handle_input change).
All checks are passing now, Cculd you take another look when you have a moment? Thank you!

smitbarmase (approved):
This looks good. I added one fix and its test to make sure we don't indent on newline and instead keep the indent from the current line we’re moving from. Sorry for delaying this for so long. Thanks so much.

vipexv:
No need to apologize at all, and thank you!

PR #40795, opened 2025-10-21
extensions_ui: Increase affordance of download button in cards

danilo-leal:
Hopefully, this will make the install/configure/uninstall buttons in the right stand out a bit more and make their presence a bit more obvious for newcomers.

| Before | After |
|--------|--------|
| Screenshot 2025-10-21 at 10  58@2x  | Screenshot 2025-10-21 at 10  53@2x | 

Release Notes:

- Increased affordance of the download button in each extension card in the extensions page.

### DIFF
--- a/crates/extensions_ui/src/components/extension_card.rs
+++ b/crates/extensions_ui/src/components/extension_card.rs
@@ -32,14 +32,14 @@ impl RenderOnce for ExtensionCard {
     fn render(self, _window: &mut Window, cx: &mut App) -> impl IntoElement {
         div().w_full().child(
             v_flex()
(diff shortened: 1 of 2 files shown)


injust:
Wanted to call out that in Catppuccin, there is more contrast between the Install and Uninstall buttons, which makes the Install button almost look disabled:

image

danilo-leal:
I hear you; but unfortunately, that's a a problem with how they set up their colors. I've never liked this very bright purple border that Catppuccin has, among other things, but it's out of our control; we only dictate how the three built-in themes look like.

PR #40802, opened 2025-10-21
Add `linux_repo_snapshot` to `.gitignore`

Veykril:
It keeps popping up in my project searches ... This prevents that. Not like we are planning on editing that file again.

Release Notes:


osiewicz:
Tbh we can probably move off of it, since we've started adding benchmarks as bin target. Will do that.

PR #40816, opened 2025-10-21
python: Init venv/virtualenv activation scripts during list/resolve

### LINKED ISSUE #40263: Windows: Doesn't activate virtualenv in WSL terminal
### Summary

Hi,

I have these project settings for a project in WSL:

```json
{
  "terminal": {
    "shell": {
      "with_arguments": {
        "program": "wsl",
        "args": [
          "--shell-type",
          "login",
          "--cd",
          "\\\\wsl.localhost\\Ubuntu-20.04\\home\\klo\\projects\\api"
        ]
      }
    },
    "detect_venv": {
      "on": {
        "directories": [".venv"],
        "activate_script": "default"
      }
    }
  }
}
```

However when the terminal starts it does not activate venv, and before the shell starts it print these lines:

```
conda activate base
clear
```

Why it is doing that, I don't even have conda installed?

Zed Dev 0.210.0 
635b71c48697a8e8df4b56c601adce5993639bc6

kubkon:
This means that existence of activation scripts for venv/virtualenv will be checked locally either on the host if editing locally, or the remote by the remote proxy if editing a remote project.

Closes https://github.com/zed-industries/zed/issues/40263

Release Notes:

- N/A

### DIFF
--- a/crates/languages/src/python.rs
+++ b/crates/languages/src/python.rs
@@ -24,7 +24,6 @@ use serde_json::{Value, json};
 use smol::lock::OnceCell;
 use std::cmp::Ordering;
 use std::env::consts;
-use std::ops::Deref;
 use util::command::new_smol_command;
 use util::fs::{make_file_executable, remove_matching};
 use util::rel_path::RelPath;
@@ -47,14 +46,6 @@ pub(crate) struct PythonToolchainData {
     environment: PythonEnvironment,
     #[serde(skip_serializing_if = "Option::is_none")]
     activation_scripts: Option<HashMap<ShellKind, PathBuf>>,
-}
-
-impl Deref for PythonToolchainData {
-    type Target = PythonEnvironment;
-
-    fn deref(&self) -> &Self::Target {
-        &self.environment
-    }
 }
 
 pub(crate) struct PyprojectTomlManifestProvider;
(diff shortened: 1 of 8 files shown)


crates/languages/src/python.rs
```diff
@@ -39,6 +41,22 @@ use std::{
...
 
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct PythonToolchainData {
+    #[serde(flatten)]
+    environment: PythonEnvironment,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    activation_scripts: Option<HashMap<ShellKind, PathBuf>>,
+}
+
+impl Deref for PythonToolchainData {
```
osiewicz: I don't really love implicit Derefs; could we instead just bite a bullet and access the `.environment` directly?

crates/languages/src/python.rs
```diff
@@ -1150,20 +1172,20 @@ impl ToolchainLister for PythonToolchainProvider {
...
-            toolchain.as_json.clone(),
-        ) else {
+    fn activation_script(&self, toolchain: &Toolchain, shell: ShellKind) -> Vec<String> {
+        let Ok(toolchain) =
+            serde_json::from_value::<PythonToolchainData>(toolchain.as_json.clone())
+        else {
             return vec![];
         };
+
+        log::debug!("Composing activation script for toolchain {toolchain:?}");
```
osiewicz: Let's mention Python in the log here.

crates/languages/src/python.rs
kubkon: Oh sure thing!

crates/languages/src/python.rs
kubkon: Nice, good idea!

crates/languages/src/python.rs
kubkon: Done in 07357232

PR #40833, opened 2025-10-21
docs: Document per-release channel configuration

mohin-io:
## Summary
- Document the `stable`/`preview`/`nightly` top-level keys that let users scope settings overrides per release channel.
- Provide an example `settings.json` snippet and call out that overrides replace array values rather than merging them.
- Mention that UI-driven changes edit the root config so per-channel blocks might need manual updates.

## Testing
- Not run (docs only).

Fixes #40458.

Release Notes:

- N/A

### DIFF
--- a/docs/src/configuring-zed.md
+++ b/docs/src/configuring-zed.md
@@ -23,7 +23,7 @@ The syntax for configuration files is a super-set of JSON that allows `//` comme
 
 ### Per-release channel overrides
 
-Zed reads the same `settings.json` across all release channels (Stable, Preview or Nightly). However, you can scope overrides to a specific channel by adding top-level `stable`, `preview`, `nightly` or `dev` objects. They are merged into the base configuration with settings from these keys taking precedence upon launching the speficied build. For example:
+Zed reads the same `settings.json` across all release channels (Stable, Preview or Nightly). However, you can scope overrides to a specific channel by adding top-level `stable`, `preview`, `nightly` or `dev` objects. They are merged into the base configuration with settings from these keys taking precedence upon launching the specified build. For example:
 
 ```json [settings]
 {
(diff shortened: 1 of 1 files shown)


mikayla-maki:
You can add per-channel and per-OS settings files in the settings.json itself, just add a new object with the key you need:

```json5
{
  "preview": {
    // All settings work here.
  }
}

**Comment by mohin-io** [comment]
﻿Thanks for the pointer! I've expanded the docs with a new "Per-release channel overrides" section that highlights the top-level `stable`/`preview`/`nightly` keys and includes an example so folks can find it without experimenting in the UI. Let me know if anything else would be helpful.

**Comment by mohin-io** [comment]
@cla-bot check

**Comment by mohin-io** [comment]
﻿I've updated the docs to document the per-channel overrides you mentioned and restored the branch. Would you mind reopening this PR so it can get another look?

**Comment by MrSubidubi** [comment]
@mohin-io Happy to, but you'll probably need to force-push here for the changes to be picked up. GitHub has some funkyness regarding closed PRs, so your recent changes are not yet visible here. Also, please update your description

File: docs/src/configuring-zed.md [hunk 1]
```diff
@@ -19,6 +19,28 @@ This configuration is merged with any local configuration inside your projects.
...
+    "theme": "zed-dark"
+  }
+}
+```
+
+In this configuration, Stable keeps the base preferences, Preview switches to `zed-dark`, and Nightly enables Vim mode with a different theme. Channel-specific blocks must include every setting you want overridden -- they do not merge arrays with the base copy.
+
+Note that changing options via the UI (for example toggling Vim mode or picking a theme) edits the root of `settings.json`. To keep per-channel differences, edit the relevant release section manually after making UI tweaks.
+
 The syntax for configuration files is a super-set of JSON that allows `//` comments.
```
MrSubidubi: That line should go with the previous section

MrSubidubi (approved):
Changed some minor phrasing, but looks good! Thanks for getting the ball rolling and your contribution!

PR #40836, opened 2025-10-21
Document global debug.json usage and fix REPL error copy

### LINKED ISSUE #40207: `copy` button(s) do not work in REPL interactive mode
### Summary

`copy` button for REPL errors does not work

### Description
using an interactive python REPL with a content like:
```python
# %%
import sys

print("std")
print("err", file=sys.stderr)
not_existing
```
would produce an error (as `not_existing` is undefined) and the output is split into two sections, stdout+stderr where the copy button works, and a second section where the `NameError:` is shown and where both the `copy` and `open REPL Output` buttons have no effect.

DESCRIPTION_HERE

Steps to reproduce:
1. start a REPL in the editor (with content as above)
2. execute
3. try to copy the stderr+stdout - works
4. try to copy the error - nothing happens (see the second copy button in the attached screenshot)

recent 0.207.4 - but I believe it never actually worked

Image

mohin-io:
﻿## Overview
- document how to keep a per-user debug.json so global launch tasks show up everywhere (Fixes #39849)
- sanitize REPL terminal text before copying so error blocks can be copied and opened in buffers (Fixes #40207)

## Design Decisions
- reused the existing user debug file (paths::debug_scenarios_file) and pointed docs at the zed::OpenDebugTasks command to stay aligned with the settings UX
- extract a sanitize helper inside TerminalOutput::full_text to strip \r/null padding while keeping indentation intact, then join the cleaned lines so clipboard and buffers get readable text

## Testing
- Not run (cargo is unavailable in this environment)

Fixes #39849.
Fixes #40207.

### DIFF
--- a/crates/repl/src/outputs/plain.rs
+++ b/crates/repl/src/outputs/plain.rs
@@ -198,7 +198,16 @@ impl TerminalOutput {
     }
 
     fn full_text(&self) -> String {
-        let mut full_text = String::new();
+        fn sanitize(mut line: String) -> Option<String> {
+            line.retain(|ch| ch != '\u{0}' && ch != '\r');
+            if line.trim().is_empty() {
+                return None;
+            }
+            let trimmed = line.trim_end_matches([' ', '\t']);
+            Some(trimmed.to_owned())
+        }
+
+        let mut lines = Vec::new();
 
         // Get the total number of lines, including history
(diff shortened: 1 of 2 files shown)


mohin-io:
@cla-bot check

rgbkrk (approved):

PR #39884, opened 2025-10-09
Add comment language injection for supported languages

thedadams:
Release Notes:

- Added comment language injections for builtin languages. This enables highlighting of `TODO`s and similar notes with the comment extension installed.

PR #40842, opened 2025-10-22
Add comment injections for gowork and gomod

thedadams:
Release Notes:

- Add comment injections for go.mod and go.work


thedadams:
I missed these in this PR: https://github.com/zed-industries/zed/pull/39884

ConradIrwin:
Thanks!

PR #40846, opened 2025-10-22
collab: Pin sea-orm-macro crate version together with sea-orm

Be-ing:
Currently running `cargo update` on Zed will break the collab crate because the versions of sea-orm and sea-orm-macros will not match. This results in a bunch of noisy warnings from rust-analyzer.

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3392,6 +3392,7 @@ dependencies = [
  "rpc",
  "scrypt",
  "sea-orm",
+ "sea-orm-macros",
  "semantic_version",
  "semver",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -900,4 +900,5 @@ ignored = [
     "serde",
     "component",
     "documented",
+    "sea-orm-macros",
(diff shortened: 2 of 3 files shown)


maxdeviant:
> Currently running `cargo update` on Zed

Is there a reason we need to run an unscoped `cargo update` on Zed?

Be-ing:
> Is there a reason we need to run an unscoped cargo update on Zed?

I did this while working on a branch and got very confused why >300 errors spammed rust-analyzer's errors & warnings for a crate I wasn't working on.

Be-ing:
For more context, on my branch I am editing the Cargo.toml of multiple crates. Thus doing `cargo update` for the whole workspace seemed appropriate. That breaking a crate I wasn't working on was puzzling.

Veykril:
Given we already pin sea-arm I do agree we should be pinning the macros crate as well. Technically just adding a new dependency could cause cargo to try bumping things iirc. So this seems like a good idea to me here

PR #40858, opened 2025-10-22
agent_ui: Remove ellipses from some menu entries

lidm0707:
image

This PR fixes inconsistent use of trailing ellipsis (…) in the MCP Server menu. This change removes unnecessary ellipses to align with standard UI/UX conventions used across Zed’s menus.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -1660,11 +1660,11 @@ impl AgentPanel {
                                     id: None,
                                 }),
                             )
-                            .action("Add Custom Server…", Box::new(AddContextServer))
+                            .action("Add Custom Server", Box::new(AddContextServer))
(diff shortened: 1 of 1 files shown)


lidm0707:
@cla-bot check

lidm0707:
@cla-bot check

danilo-leal (approved):
Thank you! I actually added the ellipsis back in the "Add Custom Server" item, though. I think that corresponds well with how both [Apple](https://developer.apple.com/design/human-interface-guidelines/menus#labels) and [Microsoft](https://learn.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses) think about adding them; that item in particular deservers an ellipsis I think because it does not _fire_ an action until you actually configure the custom server. It's somewhat similar to the Print example in MS's docs.

PR #40871, opened 2025-10-22
Fix unfinished chat being stopped when switching to another chat (concurrent chat)

wzulfikar:
Closes #40790

Release Notes:

- Fixed chat generation gets stopped when switching to other chat

**Preview**

1. Send a message in new chat, then switch to another chat (or create a new chat)
2. Write another message in a new chat, then quickly switch to another chat again
3. Now switch back and forth between those chats. You'll see that it's still generating like usual.

With this PR, the chat completion flow will still continue in background. Aside from seeing the chat progress when you switch back and forth, you'll also notice at 00:16 in the preview video below, the chat title is updated from `New Thread` to `Code for Updating Title in ACP Thread`.

https://github.com/user-attachments/assets/47992019-2bb3-4a9a-b94e-17c16db146c4

Disclaimer: I used Claude Sonnet 4.5 to help with understanding the codebase


wzulfikar:
@cla-bot check

crates/agent_ui/src/agent_panel.rs
```diff
@@ -844,19 +853,95 @@ impl AgentPanel {
...
-                        summarize_thread,
-                        workspace.clone(),
-                        project,
-                        this.history_store.clone(),
-                        this.prompt_store.clone(),
-                        window,
-                        cx,
-                    )
-                });
+                // Check thread view cache first to continue chat completion
```
wzulfikar, line 894: This is the main part to enable concurrent chat. In nutshell, we move the thread view into DetachedThread before switching to different chat so it's not disposed. And when we switch back to that unfinished chat, we're loading the view from it (thus the progress continues) instead of creating a new AcpThreadView.

wzulfikar:
I think this should be it, and seems to work well in my local zed. Ready for review!

dinocosta:
Thank you for attempting to improve on this @wzulfikar !

I'm probably not the best person to review this, so I'll wait for someone more familiar with this part of the codebase, but I tested your changes a bit and I believe we'll also need to update how Agent's Notifications are handled, as the "View Panel" button currently simply opens the `AgentPanel`. We might need to update `AgentNotificationEvent::Accepted` with the agent thread's identifier, so that the notification can open the correct thread when "View Panel" is clicked 👍

wzulfikar:
Thanks for checking, and yes, that's a great idea! Looks like I'll need to update the `AgentNotification` since it doesn't have a way to carry payload like thread id. Not sure if the change for this agent notification should be included in current PR or separately tho. Will leave this here for now.

https://github.com/zed-industries/zed/blob/a64661086f8a2bb50868fbbe990c27c8e948d0ba/crates/agent_ui/src/acp/thread_view.rs#L4620-L4628

image

wzulfikar:
Pushed commits to handle opening relevant thread. Also renamed "View Panel" to "View Thread" since we're opening a specific thread now.

https://github.com/user-attachments/assets/f6cad3e2-26d7-40bd-980a-2ab91f1ed67f

I think, what can be handled separately (i.e. different PR) is how to show multiple agent notifications. From my test, when we have multiple chats that are completing in background, zed will still only show 1 notification.

benbrandt:
@wzulfikar thanks for taking the time to look into this. However, I think we need to handle concurrent chats at a deeper level before offering this. We need to finish some design work on our end before we can accept a change around this.

wzulfikar:
Understandable, thank you!

MRiabov:
Just as a note: this keeps me using Windsurf for now. I think others too.

I suggest looking at how Windsurf/Cursor did this, it convenient and they had multiple iterations.

wzulfikar:
Yes, it's a bit of pain point for me too. If you don't mind building zed locally, you definitely can keep using it with concurrent chat or whatever tweaks you want. That's what I'm doing for now (repo [here](https://github.com/wzulfikar/zed/pulls?q=is%3Apr+is%3Aclosed)), and will go back to mainstream zed once it's added :)

PR #40882, opened 2025-10-22
Accelerate agent tests with SIMD instructions

c410-f3r:
Fix 4 TODOs

Uses a not-great/not-terrible but much safer SIMD macro that hints the compiler about the intended width.

https://godbolt.org/z/6n4xh7sT4 shows the usage of `AVX512` registers in `x86-64-v4` with ~4x instructions, which shouldn't be a problem in modern platforms. My local `x86-64-v2` machine rendered a ~2x improvement in a micro benchmarks.

```
# Before
test blit_rgba32_to_rgb24 ... bench:   4,784,195.90 ns/iter (+/- 104,752.15)

# After
test blit_rgba32_to_rgb24 ... bench:   2,264,071.70 ns/iter (+/- 167,212.39)
```

### DIFF
--- a/crates/agent/src/agent.rs
+++ b/crates/agent/src/agent.rs
@@ -1,3 +1,6 @@
+#[macro_use]
+mod macros;
+
 mod db;
 mod edit_agent;
 mod history_store;
--- a/crates/agent/src/edit_agent/evals/fixtures/from_pixels_constructor/before.rs
+++ b/crates/agent/src/edit_agent/evals/fixtures/from_pixels_constructor/before.rs
@@ -293,10 +293,21 @@ struct BlitRgb24ToA8;
 impl Blit for BlitRgb24ToA8 {
     #[inline]
     fn blit(dest: &mut [u8], src: &[u8]) {
-        // TODO(pcwalton): SIMD.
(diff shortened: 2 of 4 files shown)


benbrandt:
These are fixture files for the evals and aren't actually run.

c410-f3r:
But they eventually will, right? If not, please consider reviewing the ~400 TODOs to avoid this situation in the future as well as bit rot.

For what it is worth, the provided macro can also used in any other place that requires acceleration.

benbrandt:
They are provided as code examples for the agent to edit. This is not code meant to run as part of Zed but fixtures for the agent to make edits and we can evaluate the output of

benbrandt:
This is meant to be "real" code, unoptimized with todos and all to provide a realistic starting point for evaluation of agent performance. 

Hence it is actually a desire that it not be "perfect"

c410-f3r:
I see, well that is unfortunate. Thanks for the information.

benbrandt:
If you ever have any questions before implementing a PR please feel free to reach out ahead of time! That way your optimizations can hopefully go towards the large portions of the codebase that can actually benefit from it 😁 appreciate the effort!

PR #40883, opened 2025-10-22
crashes: Avoid crash handler on detached threads

nia-e:
Set a TLS bit to skip invoking the crash handler when a detached thread panics.

cc @P1n3appl3 - is this at odds with what we need the crash handler to do?

May close(?) #39289, cannot repro without a nightly build

Release Notes:

- Fixed extension panics crashing Zed on Linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4132,6 +4132,7 @@ dependencies = [
  "bincode 1.3.3",
  "cfg-if",
  "crash-handler",
+ "extension_host",
  "log",
  "mach2 0.5.0",
  "minidumper",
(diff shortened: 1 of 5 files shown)


P1n3appl3:
I think this ignores too many panics. Generally we still want panics in detached threads to crash the application, a quick "goto refs" indicates that there are around 450 thread detaches in our codebase, and they're used for all sorts of things. If we think the background thread running extensions specifically should be allowed to panic, I'd recommend to special case _it_ with the thread local (or even use catch_panic for it tbh, given that we compile with unwinding support).

nia-e:
I'll leave the formatting nits in `gpui/executor.rs` but this should be good!

PR #40889, opened 2025-10-22
Update runtimed to fix compatibility issue with the Ark kernel

lionel-:
Closes #40888

This updates runtimed to the latest version, which handles the "starting" variant of `execution_state`. It actually handles a bunch of other variants that are not documented in the protocol (see https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-status), like "starting", "terminating", etc. I added implementations for these variants as well.

Release Notes:

- Fixed issue that prevented the Ark kernel from working in Zed (#40888).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -534,8 +534,8 @@ itertools = "0.14.0"
 json_dotpath = "1.1"
 jsonschema = "0.30.0"
 jsonwebtoken = "9.3"
-jupyter-protocol = { git = "https://github.com/runtimed/runtimed", rev = "2ba4b96a0d093ae957254b08487696f530ef1b65" }
-jupyter-websocket-client = {  git = "https://github.com/runtimed/runtimed", rev = "2ba4b96a0d093ae957254b08487696f530ef1b65" }
+jupyter-protocol = "0.9.0"
+jupyter-websocket-client = "0.13.0"
 libc = "0.2"
 libsqlite3-sys = { version = "0.30.1", features = ["bundled"] }
 linkify = "0.10.0"
@@ -548,7 +548,7 @@ minidumper = "0.8"
 moka = { version = "0.12.10", features = ["sync"] }
 naga = { version = "25.0", features = ["wgsl-in"] }
 nanoid = "0.4"
-nbformat = { git = "https://github.com/runtimed/runtimed", rev = "2ba4b96a0d093ae957254b08487696f530ef1b65" }
+nbformat = "0.13.0"
 nix = "0.29"
 num-format = "0.4.4"
 num-traits = "0.2"
@@ -619,7 +619,7 @@ reqwest = { git = "https://github.com/zed-industries/reqwest.git", rev = "c15662
     "stream",
 ], package = "zed-reqwest", version = "0.12.15-zed" }
 rsa = "0.9.6"
-runtimelib = { git = "https://github.com/runtimed/runtimed", rev = "2ba4b96a0d093ae957254b08487696f530ef1b65", default-features = false, features = [
+runtimelib = { version = "0.29.0", default-features = false, features = [
     "async-dispatcher-runtime", "ring"
 ] }
 rust-embed = { version = "8.4", features = ["include-exclude"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -534,8 +534,8 @@ itertools = "0.14.0"
 json_dotpath = "1.1"
 jsonschema = "0.30.0"
 jsonwebtoken = "9.3"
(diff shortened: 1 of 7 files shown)


crates/client/src/client.rs
```diff
@@ -1337,10 +1341,20 @@ impl Client {
                 request_headers.insert("x-zed-metrics-id", HeaderValue::from_str(&metrics_id)?);
             }
 
+            #[cfg(any(target_os = "windows", target_os = "macos"))]
```
lionel-: A build issue appeared with the update.

I don't understand the details here, but I fixed it following existing code at https://github.com/zed-industries/zed/blob/14b41b122f38c76374b5285ea2bd4c649544642b/crates/client/src/proxy/http_proxy.rs#L67-L98

Cargo.toml
```diff
@@ -529,8 +529,8 @@ itertools = "0.14.0"
 json_dotpath = "1.1"
 jsonschema = "0.30.0"
 jsonwebtoken = "9.3"
-jupyter-protocol = { git = "https://github.com/ConradIrwin/runtimed", rev = "7130c804216b6914355d15d0b91ea91f6babd734" }
-jupyter-websocket-client = {  git = "https://github.com/ConradIrwin/runtimed" ,rev = "7130c804216b6914355d15d0b91ea91f6babd734" }
+jupyter-protocol = { git = "https://github.com/runtimed/runtimed", rev = "2ba4b96a0d093ae957254b08487696f530ef1b65" }
```
lionel-: Conrad's PR (https://github.com/runtimed/runtimed/pull/182) was merged so we should now be able to depend on the main repo.

crates/client/src/client.rs
ConradIrwin: Thanks! This seems reasonable, but so hard to know...

With my tin foil hat on I am *convinced* that there's a giant conspiracy by the authors of rust tls libraries to make it impossible to actually set up a project to use tls properly :D

Cargo.toml
ConradIrwin: Does this correspond to an actual released version? Looks like it's tagged with `sidecar-v0.8.0, tag: runt-cli-v0.3.0`. If so, can we switch back to non-git dependencies?

lionel-: yep! now updated to latest releases

ConradIrwin: Ugh, that seems to have broken the tls config

crates/client/src/client.rs
rgbkrk: Guh. Yeah.

Cargo.toml
```diff
@@ -614,8 +614,8 @@ reqwest = { git = "https://github.com/zed-industries/reqwest.git", rev = "c15662
     "stream",
 ], package = "zed-reqwest", version = "0.12.15-zed" }
 rsa = "0.9.6"
-runtimelib = {  git = "https://github.com/ConradIrwin/runtimed", rev = "7130c804216b6914355d15d0b91ea91f6babd734", default-features = false, features = [
-    "async-dispatcher-runtime",
+runtimelib = { version = "0.29.0", default-features = false, features = [
+    "async-dispatcher-runtime", "ring"
 ] }
```
rgbkrk, line 624: Thank you so much for doing this absolutely necessary maintenance work to bring Zed up to the latest.

Cargo.toml
rgbkrk: Is there something we need to change in https://github.com/runtimed/runtimed to make this compatible?

Cargo.toml
```diff
@@ -614,8 +614,8 @@ reqwest = { git = "https://github.com/zed-industries/reqwest.git", rev = "c15662
     "stream",
 ], package = "zed-reqwest", version = "0.12.15-zed" }
 rsa = "0.9.6"
-runtimelib = {  git = "https://github.com/ConradIrwin/runtimed", rev = "7130c804216b6914355d15d0b91ea91f6babd734", default-features = false, features = [
-    "async-dispatcher-runtime",
+runtimelib = { version = "0.29.0", default-features = false, features = [
+    "async-dispatcher-runtime", "ring"
```
rgbkrk: Going to push up a change over to `aws-lc-rs`

ConradIrwin:
@lionel- Ok, we pushed some changes to runtimelib to make sure the feature flags for async-tungstenite match what we use in Zed, to avoid needing to change the tls setup.

Assuming this fixes the build issues, will merge.

rgbkrk:
Thanks @lionel- and @ConradIrwin!

rgbkrk:
Tried it out this morning, it's looking great.

https://github.com/user-attachments/assets/00b42cc8-eda9-4d57-99bd-bcd176956298

PR #40913, opened 2025-10-22
yaml: Fix indentation with dictionary when editing

### LINKED ISSUE #39570: Zed un-indents YAML dictionary inside a list when editing it
### Summary

Zed un-indents YAML dictionary inside a list when editing it

### Description

```yaml
foo:
  - bar: one
    baz: two
```

Steps to reproduce:
1. Select `baz: two`
2. Type something else

**Expected Behavior**: No indentation change
**Actual Behavior**: Zed unindents the text that replaces `baz: two` to the same level as the `-`

Zed: v0.206.6 (Zed) 
OS: macOS 15.7.1
Memory: 16 GiB
Architecture: x86_64

bnjjj:
Closes #39570 

Release Notes:

- Fixed: indentation with dictionary when editing YAML file


bnjjj:
@smitbarmase do you think you could take a quick look to know if I have to change anything in the implementation please ?

crates/editor/src/editor_tests.rs
```diff
@@ -3135,6 +3135,65 @@ fn test_newline(cx: &mut TestAppContext) {
...
+    cx.set_state(indoc! {"
+    test:
+        - foo: bar
+        - bar: fooˇ"});
+    cx.update_editor(|e, window, cx| e.newline(&Newline, window, cx));
+    cx.assert_editor_state(indoc! {"
+    test:
+        - foo: bar
+        - bar: foo
+            ˇ"});
```
smitbarmase: Sorry for the late reply. I was out for a week. 

Do you think this case can be improved such that the cursor is at the same indent level of `-`, as in the case of an array of strings?

smitbarmase (changes requested):

crates/editor/src/editor_tests.rs
bnjjj: It should be good now :)

smitbarmase (approved):
Thanks, this looks good to me.

PR #40942, opened 2025-10-22
agent_ui: Show notifications also when the panel is hidden

cmclaughlin:
Currently Zed only displays agent notifications (e.g. when the agent completes a task) if the user has switched apps and Zed is not in the foreground. This adds PR supports the scenario where the agent finishes a long-running task and the user is busy coding within Zed on something else.

Releases Note:

- If agent notifications are turned on, they will now also be displayed when the agent panel is hidden, in complement to them showing when the Zed window is in the background.


cmclaughlin:
Hello team,

This is my first time authoring Rust and my first PR for Zed, so please let me know if there's any room for improvement here. It was composed by Claude (in Zed), so hopefully it looks good to you.

danilo-leal (commented):
Hey @cmclaughlin, thanks and congrats for the first PR; this is a super cool addition! 

I actually think we can start out without having the setting and just follow the existing `notify_when_agent_waiting` one. Feels like it's a reasonable default; i.e., if the user has a non-null value for that setting, Zed would be notifying them if the app is either in the background or if the panel is hidden. IMO, we can test out simpler like this and re-evaluate the need for a setting later down the road.

cmclaughlin:
@danilo-leal thank you for the feedback. Would you please take another look?

danilo-leal (approved):
Thank you for following it up & appreciate writing the tests!

Issue #40538, opened 2025-10-17
Mouse wheel scrolling: long random jumps after switching from another application to ZED


ardabro:
### Summary

Mouse wheel scrolling: long random jumps after switching from another application to ZED.
...at least in Debian, Cinnamon. 100% reproduction.

### Description
I seems that after switching from another application, the editor incorrectly initializes something related to the current position in the file, and it is definitely associated with the mouse wheel scrolling. When you scroll the wheel with another application focused, then back to ZED, then touch the mouse wheel, it jumps somewhere, far from the last editing cursor position, apparently deriving the current position from the wheel actions performed on another window. It looks like a missing reset of a counter or something like that. At the beginning, I was thinking that it was related to my environment, but I couldn't reproduce this behavior with any other application (including gedit, Sublime Text, Chromium: edit window in GitHub).
UPD: the same behavior in the project pane and in the outline pane. It is something global

Steps to reproduce:
1. Open zed, set focus in the editor pane, in the middle of a file long enough to see large scrolling jumps.
2. Switch to another application (alt+tab)
3. Scroll the wheel (doesn't matter if this application doesn't scroll anything inside its window).
4. Switch back to zed (alt+tab)
5. Touch the mouse wheel to trigger.
100% reproduction.

**Expected Behavior**:
Current position in the edited file (and project pane and outpline pane) is stable when switching back to ZED after scrolling the mouse wheel on another window.

**Actual Behavior**:
Current position jumps like crazy.

### Zed Version and System Specs

Zed: v0.208.5 (Zed) 
OS: Linux X11 debian 13
Memory: 31.1 GiB
Architecture: x86_64
GPU: NVIDIA GeForce GTX 1650 || NVIDIA || 580.95.05

dinocosta:
@ardabro Could this somehow be related to the mouse you're using? I just tested this on MacOS and I'm able to see the same behavior but that's because the scroll wheel is still scrolling when I switch back to Zed, is that not the case for you? 🤔

ardabro:
1) My mouse wheel definitely isn't "still rolling"
2) I observe the same behavior with a completely different mouse (just checked).
3) I do not observe this issue with other applications including text editors.

dinocosta:
Thank you for sharing more information @ardabro ! I've shared this with the team working with Linux so they can eventually take a look at it 🙂

Vanuan:
Looks like already reported months ago https://github.com/zed-industries/zed/issues/34901

[...]

PR #40950, opened 2025-10-22
Implement InputStateManager for input state management

Vanuan:
Draft idea for fixing #40538

Introduced InputStateManager for managing input states, including drag operations and scroll accumulators. Enhanced XDND handling and focus-aware scroll delta calculation.

Do not merge, just suggest feedback.


Vanuan:
The main challenge is how to distinguish between drag operations initiated from zed and don't cancel related events. But this is overkill for fixing the scroll issue. But the issue appears to be not only with scroll. Sometimes the selection mode is also stuck.

PR #40953, opened 2025-10-23
Add builtin Pyrefly language server support

lolpack:
This will add Pyrefly as an option for Python language services https://pyrefly.org

Release Notes:

- Added Pyrefly as builtin language server option


lolpack:
@ConradIrwin I've accepted the CLA and come back to integrating this. Question on getting a copy of the Pyrefly binary... what is the best path for that? Install it automatically or suggest to the user to install? I will cleanup this PR and then open it.

lolpack:
@cla-bot check

ConradIrwin:
@lolpack Sorry for missing your question!

Usually we download the LSP from the latest GitHub release, so we should probably copy that here (there is a long outstanding feature request to disable this behavior that will apply to this retroactively).

lolpack:
Closing in favor of https://github.com/zed-industries/zed/pull/44262

PR #40957, opened 2025-10-23
gpui: Add subscribe_new and subscribe_new_in helper methods to Context<T>

Anthony-Eid:
I added two helper methods called `subscribe_new` and `subscribe_new_in`, both of which make it easier to create event handlers whenever a new Entity<T> is created. Similar to `observe_new`.

I also renamed some variable names from `view` to `entity` to be more inline with gpui3. The view name seems to be an artifact from gpui2.

Release Notes:

- N/A


Anthony-Eid:
This api adds more complexity and just combines two apis that a user can already use so I'm closing this PR

PR #40986, opened 2025-10-23
build(config): Apply +crt-static for MSVC on Windows

ognevny:
Currently livekit is compilable only on MSVC environment

Release Notes:

- N/A


ognevny:
okay, that doesn't work in a way I thought...

PR #41009, opened 2025-10-23
Add project panel padding setting

nhedger:
This PR introduces a new `project_panel.padding` setting which allows setting a custom padding in pixels for the project panel.

The default padding is set to `0` to match the existing behaviour.

For now this setting applies the same padding in all directions, but if it's desirable, I'm open to decomposing this in `top_padding`,  `right_padding`, `bottom_padding`, and `left_padding` options.

Screenshots:

<table>
<tr>
<th>Without padding</th>
<th>With 10px padding</th>
<th>Settings UI</th>
</tr>
<tr>
 <td>
Screenshot 2025-10-23 at 15 13 30
</td>
 <td>
Screenshot 2025-10-23 at 15 13 54
</td>
 <td>
Screenshot 2025-10-23 at 15 14 06
</td>
</table>

Motivation:

While I really like Zed, I sometimes find myself wishing for a little bit more breathing room, so I thought I'd contribute this feature as my first contribution.

Release Notes:

- Added a new setting for applying padding to the project panel

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -658,6 +658,8 @@
     "hide_gitignore": false,
     // Default width of the project panel.
     "default_width": 240,
+    // Padding inside the project panel, in pixels.
+    "padding": 0,
     // Where to dock the project panel. Can be 'left' or 'right'.
     "dock": "left",
     // Spacing between worktree entries in the project panel. Can be 'comfortable' or 'standard'.
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -5409,6 +5409,7 @@ impl Render for ProjectPanel {
                         .on_drag_move(cx.listener(handle_drag_move::<DraggedSelection>))
                 })
                 .size_full()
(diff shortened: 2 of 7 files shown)


nhedger:
@cla-bot check

smitbarmase:
Thanks for the effort. This setting might introduce more design artifacts, like entries being cut off at the top or bottom, inconsistent spacing issues, and bunch. I also think it might be too much to include this as a setting since we already have so many panels and toolbars, and we’re not keen on adding such granular options for all of them. 

You can experiment with the UI font size for an easier look. I'm closing this for now, thanks again!

danilo-leal:
Thanks @nhedger, appreciate the PR & we're just starting to use Biome over here for our web projects, thanks for that! In the future, we may want to pick up an older exploration we had around UI density, and that will allow to give more breathing room to the entire app holistically as opposed to just on an individual surface like the project panel. Hopefully, sometime in the not-so-distant future!

nhedger:
No biggie, I understand! Looking forward to that !

PR #41018, opened 2025-10-23
markdown: Add support for rendering LaTeX formulas 

lidm0707:
Closes #40813

Release Notes:

- Supports inline and display math in both markdown editor and preview.
-  Inline math: $x^2 + y^2 = z^2$
- Display math: $$E = m \times c^2$$
- LaTeX to Unicode conversion (e.g., \times → ×, ^2 → ², \sqrt → √)
- Works in both regular markdown and markdown preview
image


someone13574:
If you want to implement something more feature-complete, this might be worth looking at: https://github.com/lnay/latex-gpui-experiment

bennetbo:
Given that it only supports a few symbols, I'm not sure if it is worth maintaining this code. Also I feel like the unicode characters are still not great for readability. So for now I'd like to keep the current state instead of shipping something that is half-baked.
Long term I think it might be nice to support Latex properly, but not sure how much overhead that will introduce.

PR #41030, opened 2025-10-23
Make 'wrap selections in tag' work with line selection mode

Trombach:
The `wrap selections in tag` action currently did not take line_mode into account, which means when selecting lines with `shift-v`, the start/end tags would be inserted into the middle of the selection (where the cursor sits)

https://github.com/user-attachments/assets/a1cbf3da-d52a-42e2-aecf-1a7b6d1dbb32

This PR fixes this behaviour by checking if the selection uses line_mode and then adjusting start and end points accordingly.

NOTE: I looked into amending the test cases for this, but I am unsure how to express line mode with range markers. I would appreciate some guidance on this and then I am happy to add test cases.

After:

https://github.com/user-attachments/assets/a212c41f-b0db-4f50-866f-fced7bc677ca

Release Notes:

- Fixed `Editor: wrap selection in tags` when in vim visual line mode

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -10520,9 +10520,12 @@ impl Editor {
         self.hide_mouse_cursor(HideMouseCursorOrigin::TypingAction, cx);
 
         let snapshot = self.buffer.read(cx).snapshot(cx);
+        let max_point = snapshot.max_point();
 
         let mut edits = Vec::new();
         let mut boundaries = Vec::new();
+
+        let is_line_mode = self.selections.line_mode();
 
         for selection in self
             .selections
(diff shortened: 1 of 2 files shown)


Trombach:
@cla-bot check

ConradIrwin:
Thanks for the fix!

You can set line_mode in the tests, and then leave the selection as is.

Or if you want to write the test in the vim crate, you can use the existing helpers to run `shift-v` to put yourself into line mode before triggering the action.

Trombach:
Ok, thanks, that makes sense. I'll take a look which way works best and will update this PR with a test case

Trombach:
@ConradIrwin I have added a test case in the vim crate, since it seemed more natural to do there.

ConradIrwin:
I realized we already had a helper for doing this, so updated the PR. Thanks again!

PR #41058, opened 2025-10-23
Fix typo in Font Features description, s/b "OpenType"

warrenjokinen:
Fix typo (regression?), 
"Opentype" should be "OpenType"

Closes #ISSUE

Release Notes:

- N/A


danilo-leal (approved):
Thank you, good catch!

PR #41060, opened 2025-10-23
Revert: Spawn terminal process on background executor 

agu-z:
Reverts https://github.com/zed-industries/zed/pull/40774 (and the dependent https://github.com/zed-industries/zed/pull/40824) as it introduces a bug where Nushell processes are leaked and Ctrl+C doesn't kill the current process.

Release Notes:

- Fix a bug where nushell processes wouldn't get killed after closing a terminal tab

### DIFF
--- a/crates/agent_ui/src/agent_diff.rs
+++ b/crates/agent_ui/src/agent_diff.rs
@@ -581,13 +581,11 @@ impl Item for AgentDiffPane {
         _workspace_id: Option<workspace::WorkspaceId>,
         window: &mut Window,
         cx: &mut Context<Self>,
-    ) -> Task<Option<Entity<Self>>>
+    ) -> Option<Entity<Self>>
     where
         Self: Sized,
     {
-        Task::ready(Some(cx.new(|cx| {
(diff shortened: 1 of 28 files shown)


ghost:
Thank you for this fix! 🙏

jackTabsCode:
It seems like this has regressed again @agu-z?

As of 212 I think, it seems like ctrl-c is no longer working for some processes in Nushell

Veykril:
What operating system? DO you have a concrete reproducer? We did reland the original PR with some fixes applied again

jackTabsCode:
Simple reproduction:
```
loop { print "Hi"; sleep 1sec }
```
 on `nu 0.108.0`

https://github.com/user-attachments/assets/9b1bfbbe-aa1e-4bfe-bfbe-4062519408c5

Zed: v0.212.2 (Zed Preview) 
OS: macOS 26.1
Memory: 48 GiB
Architecture: aarch64

@Veykril 

Edit: This occurs with bash too: `while true; do echo "Hi"; sleep 1; done`

PR #41078, opened 2025-10-24
[WIP] editor: Implement next/prev reference

cameron1024:
Adds actions `GoToNextReference` and `GoToPreviousReference`, which go to the
next/previous LSP reference to the symbol under the cursor (in the current
file). For example:
```rust
fn main() {
  //  v <- (1)
  let abc = 123;

  //      v <- (2)
  let x = abc + 1;

  //      v <- (3)
  let y = abc + 2;
}
```
If the cursor starts at `(1)`, then going to the next reference will go to
`(2)`, then `(3)`, then back to `(1)`.

Co-authored-by: Cole <cole@zed.dev>

Release Notes:

- Added `GoToNextReference` and `GoToPreviousReference` editor actions


xdBronch:
isnt this basically the existing `go to next document highlight`? should both exist?

PR #40160, opened 2025-10-14
Add configurable sort modes to file explorer/project panel

lparry:
Closes #4533 (partly at least)

Release Notes:

- Added `project_panel.sort_mode` option to control explorer file sort (directories first, mixed, files first)

 ## Summary

Adds three sorting modes for the project panel to give users more control over how files and directories are displayed:

- **`directories_first`** (default): Current behaviour - directories grouped before files
- **`mixed`**: Files and directories sorted together alphabetically
- **`files_first`**: filed grouped before directories

 ## Motivation

Users coming from different editors and file managers have different expectations for file sorting. Some prefer directories grouped at the top (traditional), while others prefer the macOS Finder-style mixed sorting where "Apple1/", "apple2.tsx" and "Apple3/" appear alphabetically mixed together.

 ### Screenshots

New sort options in settings:
image

Directories first | Mixed | Files first
-------------|-----|-----
image | image | image

### Agent usage

Copilot-cli/claude-code/codex-cli helped out a lot. I'm not from a rust background, but really wanted this solved, and it gave me a chance to play with some of the coding agents I'm not permitted to use for work stuff

PR #41080, opened 2025-10-24
project_panel: Make natural sort ordering consistent with other apps

lparry:
The existing sorting approach when faced with `Dir1`, `dir2`, `Dir3`, would only get as far as comparing the stems without numbers (`dir` and `Dir`), and then the lowercase-first tie breaker in that function would determine that `dir2` should come first, resulting in an undesirable order of `dir2`, `Dir1`, `Dir3`.

This patch defers tie-breaking until it's determined that there's no other difference in the strings outside of case to order on, at which point we tie-break to provide a stable sort.

Natural number sorting is still preserved, and mixing different cases alphabetically (as opposed to all lowercase alpha, followed by all uppercase alpha) is preserved.

Closes #41080

Release Notes:

- Fixed: ProjectPanel sorting bug

Screenshots:

Before | After
----|---
image | image

I'm having trouble reasoning through what was previously going wrong with `docs` in the before screenshot, but it also seems to now appear alphabetically where you'd expect it with this patch

### DIFF
--- a/crates/util/src/paths.rs
+++ b/crates/util/src/paths.rs
@@ -1240,27 +1240,19 @@ mod tests {
 
     #[perf]
     fn compare_paths_mixed_case_numeric_ordering() {
-        let mut paths = vec![
-            (Path::new(".cargo"), false),
-            (Path::new(".cloudflare"), false),
+        let mut entries = vec![
             (Path::new(".config"), false),
-            (Path::new(".github"), false),
-            (Path::new(".zed"), false),
-            (Path::new("assets"), false),
-            (Path::new("crates"), false),
             (Path::new("Dir1"), false),
+            (Path::new("dir01"), false),
             (Path::new("dir2"), false),
-            (Path::new("Dir3"), false),
-            (Path::new("docs"), false),
-            (Path::new("extensions"), false),
-            (Path::new("legal"), false),
-            (Path::new("nix"), false),
-            (Path::new("script"), false),
+            (Path::new("Dir02"), false),
+            (Path::new("dir10"), false),
+            (Path::new("Dir10"), false),
         ];
 
-        paths.sort_by(|&a, &b| compare_paths(a, b));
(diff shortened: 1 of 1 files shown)


smitbarmase (changes requested):
That sounds right. I checked VSCode on macOS, Linux, and Windows, they all behave the same. Finder on macOS and File Explorer on Windows use the same ordering.

I’d appreciate test cases for these orderings so we don’t regress in the future.

lparry:
Thanks for the review @smitbarmase. I've added some extra test cases based on the screenshots in the PR to try and help avoid regression 🙏 Please let me know if these aren't what you were asking for :)

smitbarmase (approved):
I improved tests a bit. This looks in great shape. Thanks. Sorry for the delay.

osiewicz:
Did this anyhow regress our project panel benchmarks? I'm not saying it did, just asking.

lparry:
@osiewicz Sorry, didn't see this before. 

```
❯ cargo bench --package project_panel --bench sorting -- --baseline initial
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.49s
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.4
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running benches/sorting.rs (target/release/deps/sorting-96f5ed6c042ab29c)
Gnuplot not found, using plotters backend
Sort linux worktree snapshot
                        time:   [11.305 ms 11.350 ms 11.397 ms]
                        change: [+1.6545% +2.2637% +2.9144%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Looks like there is a minor regression from this change, but I think there is another change in #40160 that brings performance back to roughly the same as it was in the ref prior to this being merged - https://github.com/zed-industries/zed/pull/40160#issuecomment-3538435396

PR #41084, opened 2025-10-24
gpui: Impl Default for TextRun

huacnlee:
Release Notes:

- N/A

When I was implementing Input, I often used `TextRun`, but `background`, `underline` and `strikethrough` were often not used. 

So make change to simplify it.


mikayla-maki:
Thank you!

PR #41095, opened 2025-10-24
ui_input: Don’t focus previous on decrement in `NumberField`

ToBinio:
This PR removes the behavior that the number_field changes focus to the previous element when using the decrement button.

I mainly noticed this while decreasing the tab-size of a language, since it there closes the page...

Please note that I am unsure what if any purpose this code has.
I was unable to find a use case and since it is not present in the `increment_handler` I guess it should never have been here

---

Release Notes:

* Fixed wrongly focus previous element on number_field decrement


ToBinio:
@cla-bot check

danilo-leal (approved):
Good catch, this makes sense; thank you! Congrats on the first PR 😄

PR #41096, opened 2025-10-24
lsp: Support tracking multiple registrations of diagnostic providers

### LINKED ISSUE #40966: Diagnostics stop updating if "language_servers" is defined for Python
### Summary

Diagnostics stop updating when defining the "language_servers" field for Python in settings.json
### Description

Steps to trigger the problem:
1. open `settings.json`
2. add the following conffiguration:
```jsonc
  "languages": {
    "Python": {
      // When the config is enabled, the diagnostics will stop updating until restart
      "language_servers": ["basedpyright", "!ty", "!pyright"]
    }
  }
```
3. type sommething that can cause error, for example `immmport numpy`

**Expected Behavior**:
Zed should continue updating diagnostics in real time
**Actual Behavior**:
Once this configuration is enabled, diagnostics stop updating until `Restart All Language Servers`

Here's the  gif to show this:

Image

Zed: v0.209.5 (Zed) 
OS: Windows 10.0.26100
Memory: 31.6 GiB
Architecture: x86_64
GPU: Intel(R) Arc(TM) Graphics || Intel Corporation || 31.0.101.5008

### LINKED ISSUE #41195: Pyright LSP stopped working
### Summary

Pyright LSP is only working when it loads files, any subsequent change is not recognized by the LSP. This happened since the most recent update.

### Description
Loading a workspace correctly kicks off Pyright LSP but once you make changes to files, it doesn't do anything anymore even though it should trigger.

**Expected Behavior**:
Work as normal, lsp should update analysis when file changes.

**Actual Behavior**:

logs LSP
```
// Send:
{"jsonrpc":"2.0","method":"workspace/didChangeWatchedFiles","params":{"changes":[{"uri":"file:///Users/<redacted>/Development/<redacted>/python/tests/test.py","type":2}]}}

// Receive:
{"jsonrpc":"2.0","id":54,"method":"workspace/diagnostic/refresh"}

// Send:
{"jsonrpc":"2.0","id":54,"result":null}
```

It recognises a refresh but it doesnt actually return anything new, only when I restart the lsp

Zed: v0.209.6 (Zed) 
OS: macOS 15.6.1
Memory: 128 GiB
Architecture: aarch64

osiewicz:
Closes #40966
Closes #41195
Closes #40980 

Release Notes:

- Fixed diagnostics not working with basedpyright/pyright beyond an initial version of the document

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -4380,7 +4380,7 @@ impl LspStore {
         cx: &App,
     ) -> bool
     where
-        F: Fn(&lsp::ServerCapabilities) -> bool,
+        F: FnMut(&lsp::ServerCapabilities) -> bool,
     {
         let Some(language) = buffer.read(cx).language().cloned() else {
             return false;
@@ -6460,12 +6460,30 @@ impl LspStore {
         let buffer_id = buffer.read(cx).remote_id();
 
         if let Some((client, upstream_project_id)) = self.upstream_client() {
+            let mut suitable_capabilities = None;
+            // Are we capable for proto request?
+            let any_server_has_diagnostics_provider = self.check_if_capable_for_proto_request(
+                &buffer,
+                |capabilities| {
+                    if let Some(caps) = &capabilities.diagnostic_provider {
+                        suitable_capabilities = Some(caps.clone());
+                        true
+                    } else {
+                        false
+                    }
+                },
+                cx,
+            );
+            // We don't really care which caps are passed into the request, as they're ignored by RPC anyways.
+            let Some(dynamic_caps) = suitable_capabilities else {
+                return Task::ready(Ok(None));
+            };
+            assert!(any_server_has_diagnostics_provider);
+
             let request = GetDocumentDiagnostics {
                 previous_result_id: None,
+                dynamic_caps, // todo!
             };
-            if !self.is_capable_for_proto_request(&buffer, &request, cx) {
-                return Task::ready(Ok(None));
-            }
             let request_task = client.request_lsp(
                 upstream_project_id,
                 None,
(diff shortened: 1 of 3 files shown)


crates/project/src/lsp_store.rs
```diff
@@ -6447,12 +6460,30 @@ impl LspStore {
...
+                    }
+                },
+                cx,
+            );
+            // We don't really care which caps are passed into the request, as they're ignored by RPC anyways.
+            let Some(dynamic_caps) = suitable_capabilities else {
+                return Task::ready(Ok(None));
+            };
+            assert!(any_server_has_diagnostics_provider);
+
```
TilakMaddy, line 6482: debug_assert maybe ?

osiewicz: Not really. Since `check_if_capable_for_proto_request` returned true, then we must've hit the branch where `suitable_capabilties` are some. It does make sense to assert here (in case code changes), but as is, the code is obvious enough that I want it to fire off regardless. If this were a debug_assert, we'd fire off an unnecessary RPC request in release builds, which is pretty undesired.

mstarodub:
The issue still seems to be present on 0.209.7 - I can't see ty's diagnostics

injust:
> The issue still seems to be present on 0.209.7 - I can't see ty's diagnostics

This fixed basedpyright for me.

EDIT: This seems to have broken Ruff diagnostics, so maybe ty is also affected because of this. Try downgrading to 0.209.6.

Benjamin-van-Heerden:
This broke what was working for me. basedpyright now shows nothing, ruff also shows nothing. Manually setting the python language servers to use pyright does show some diagnostics

mstarodub:
Still no diagnostics on 0.210.4. This is seriously inconvenient

Issue #26819, opened 2025-03-15
editor: organise imports does not work for Python files on Linux


bulletmark:
### Summary

The function `editor: organize imports` does not work for Python files on Linux, neither explicitly from the menu, or trying to use it automatically with the formatter (and/or with `format_on_save`) using the [suggested configuration](https://docs.astral.sh/ruff/editors/setup/#zed). There is no error popup/message and nothing is printed in any of the logs I can find.

The ruff extension is installed. Interestingly, the function `editor: toggle code actions` does present 2 ruff actions in a popup and selecting "organize imports" from that menu does work. This function is independent of any configuration and only requires the ruff extension to be installed.

This issue is raised after the discussion at https://discord.com/channels/869392257814519848/1350265535715086377. A participant in that discussion says it works for him on a Mac, although it does not work with `format_on_save` config setting.

### Zed Version and System Specs

```
Zed: v0.178.2 (Zed Preview) 
OS: Linux Wayland arch unknown
Memory: 15.1 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Plus Graphics (ICL GT2) || Intel open-source Mesa driver || Mesa 25.0.1-arch1.2
```

probably-neb:
Unfortunately, I'm unable to reproduce imports not being sorted with "format_on_save" on both Mac and Linux. I am however able to reproduce the `editor: organize imports` action not working.

Are you able to see something like the following when you save the file with the LSP logs up?

https://github.com/user-attachments/assets/e0a859a9-7208-4309-8760-532f38f2843c

bulletmark:
If I move some imports around then the following lines are the only ones output after I `:write` the file:
```
// Send:
{"jsonrpc":"2.0","method":"textDocument/didSave","params":{"textDocument":{"uri":"file:///home/mark/Data/src/gh/pinstall/pinstall/pinstall.py"}}}
// Send:
{"jsonrpc":"2.0","method":"workspace/didChangeWatchedFiles","params":{"changes":[{"uri":"file:///home/mark/Data/src/gh/pinstall/pinstall/pinstall.py","type":2}]}}
```

My complete settings are:
```

{
  "assistant": {
    "default_model": {
      "provider": "zed.dev",
      "model": "claude-3-5-sonnet-latest"
    },
    "version": "2"
  },
  "features": {
    "edit_prediction_provider": "zed"
  },
  "vim_mode": true,
  // "ui_font_size": 16,
  "buffer_font_size": 16,
[...]

PR #41103, opened 2025-10-24
Enable `source.organizeImports.ruff` by default for Python

JosephTLyons:
Release Notes:

- Enabled automatic import organization, via [ruff](https://github.com/astral-sh/ruff), when saving Python files. To disable this, use:

```json
"Python": {
  "code_actions_on_format": { "source.organizeImports.ruff": false }
}
```

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1772,6 +1772,9 @@
       "allow_rewrap": "anywhere"
     },
     "Python": {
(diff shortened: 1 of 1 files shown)


abdelq:
Would be nice if this was fixed too https://github.com/zed-industries/zed/issues/26819#issuecomment-3399175027

PR #41110, opened 2025-10-24
markdown: Add LaTeX syntax highlighting injection

### LINKED ISSUE #30264: LaTeX syntax highlighting within markdown
### Summary

tree-sitter-markdown supports an optional non-default extension [EXTENSION_LATEX](https://github.com/tree-sitter-grammars/tree-sitter-markdown/blob/071b3b70baad138aa28832cf02d16ca48f2e6f4d/common/common.js#L11) which supports displaying syntax highlighting for inline LaTeX inside Markdown documents.

### Description

This is not a request to add LaTeX rendering support to Zed Markdown Preview.
This would merely support syntax highlighting of LaTex written within Markdown.  

At build time, if `EXTENSION_LATEX` is set, markdown-tree-sitter I believe enable support for this (untested).

```markdown
$$\int_{coffee}^{code} productivity \cdot d(time) = \sum_{i=0}^{bugs} frustration_i$$
```

$$\int_{coffee}^{code} productivity \cdot d(time) = \sum_{i=0}^{bugs} frustration_i$$

### 

VSCode vs Zed: 

Image

Image

lnay:
Closes [#30264](https://github.com/zed-industries/zed/issues/30264)

Small addition based on [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/runtime/queries/markdown_inline/injections.scm)

Screenshot From 2025-10-24 15-47-58

This does require the LaTeX extension to be installed.

Release Notes:

- Provide LaTeX highlighting for inline and display equations in Markdown

### DIFF
--- a/crates/languages/src/markdown-inline/injections.scm
+++ b/crates/languages/src/markdown-inline/injections.scm
@@ -1,3 +1,2 @@
 ((latex_block) @injection.content
-  (#set! injection.language "latex")
-  (#set! injection.include-children))
+  (#set! injection.language "latex"))
--- a/crates/languages/src/markdown-inline/injections.scm
+++ b/crates/languages/src/markdown-inline/injections.scm
@@ -1,0 +1,3 @@
+((latex_block) @injection.content
+  (#set! injection.language "latex")
+  (#set! injection.include-children))


crates/languages/src/markdown-inline/injections.scm
```diff
@@ -0,0 +1,3 @@
+((latex_block) @injection.content
+  (#set! injection.language "latex")
+  (#set! injection.include-children))
```
probably-neb:
```suggestion
  (#set! injection.language "latex"))
```

`injection.include-children` is not a real attribute. We don't give errors for unrecognized keys here, but it should be removed in case other contributors are looking at our builtin extensions for examples of how to do things :)

probably-neb (changes requested):
Looks good! Lets remove the unnecessary attribute, and then this will be good to merge

crates/languages/src/markdown-inline/injections.scm
probably-neb: Happy to do it for you if you'd like!

lnay: I thought I'd noticed something wrong with the highlighting when I removed that directive. If it looks fine on your end feel free to make the change and merge: I'm unlikely to be able to check this soon (as small as it is).

lnay: ...something wrong as in it didn't look like the screenshot when you typed out those contents.

probably-neb (approved):

PR #41111, opened 2025-10-24
language: Fix language detection for injected syntax layers

### LINKED ISSUE #40632: The `editor::WrapSelectionsInTag` action not working in ERB files anymore
### Summary

When it was initially released, `editor::WrapSelectionsInTag` worked in `HTML+ERB` files, but since recently it stopped working.

### Description

I know `HTML+ERB` support is technically part of the [ruby](https://github.com/zed-extensions/ruby) extension, but according to @vitallium in https://github.com/zed-extensions/ruby/pull/191#issuecomment-3417276833, recent changes in `editor::WrapSelectionsInTag` availability made it difficult to add support (adding `wrap_characters` to `HTML+ERB` language definition didn't help).

Steps to reproduce:

1. have the following `.html.erb` file:

   ```erb
   <div>Hi!</div>
   <%= tag.div(flash[:alert], style: "color:red") if flash[:alert] %>
   ```

2. start a new line after the `<div>` line
3. type `cmd-shift-p`

**Expected Behavior**: I would expect `editor: wrap selections in tag` action to be available in the command palette, since I'm editing HTML.
**Actual Behavior**: The `editor: wrap selections in tag` action is not available. If I change the language from the picker from `HTML+ERB` to `HTML`, then the action is available.

Zed: v0.208.5 (Zed) 
OS: macOS 15.5
Memory: 16 GiB
Architecture: aarch64

vitallium:
Closes #40632

**TL;DR:** The `wrap selections in tag` action was unavailable in ERB files, even when the cursor was positioned in HTML content (outside of Ruby code blocks). This happened because `syntax_layer_at()` incorrectly returned the Ruby language for positions that were actually in HTML. **NOTE:** I am not familiar with that part of Zed so it could be that the fix here is completely incorrect.

Previously, `syntax_layer_at` incorrectly reported injected languages (e.g., Ruby in ERB files) even when the cursor was in the base language content (HTML). This broke actions like `wrap selections in tag` that depend on language-specific configuration.

The issue had two parts:
1. Missing start boundary check: The filter only checked if a layer's end was after the cursor (`end_byte() > offset`), not if it started before, causing layers outside the cursor position to be included. See the `BEFORE` video: when I click on the HTML part it reports `Ruby` language instead of `HTML`.
2. Wrong boundary reference for injections: For injected layers with `included_sub_ranges` (like Ruby code blocks in ERB), checking the root node boundaries returned the entire file range instead of the actual injection ranges.

This fix:
- Adds the containment check using half-open range semantics [start, end) for root node boundaries. That ensures proper reporting of the detected language when a cursor (`|`) is located right after the injection:

   ```
   <body>
    <%= yield %>|
  </body>
   ```

- Checks `included_sub_ranges` for injected layers to determine if the cursor is actually within an injection
- Falls back to root node boundaries for base layers without sub-ranges. This is the original behavior.

Fixes ERB language support where actions should be available based on the cursor's actual language context. I think that also applies to some other template languages like HEEX (Phoenix) and `*.pug`. On short videos below you can see how I navigate through the ERB template and the terminal on the right outputs the detected language if you apply the following patch:

```diff
diff --git i/crates/editor/src/editor.rs w/crates/editor/src/editor.rs
index 15af61f5d2..54a8e0ae37 100644
--- i/crates/editor/src/editor.rs
+++ w/crates/editor/src/editor.rs
@@ -10671,6 +10671,7 @@ impl Editor {
         for selection in self.selections.disjoint_anchors_arc().iter() {
             if snapshot
                 .language_at(selection.start)
+                .inspect(|language| println!("Detected language: {:?}", language))
                 .and_then(|lang| lang.config().wrap_characters.as_ref())
                 .is_some()
             {
```

**Before:**

https://github.com/user-attachments/assets/3f8358f4-d343-462e-b6b1-3f1f2e8c533d

**After:**

https://github.com/user-attachments/assets/c1b9f065-1b44-45a2-8a24-76b7d812130d

Here is the ERB template:

```
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
    <style>
      /* Email styles need to be inline */
    </style>
  </head>
  <body>
    <%= yield %>
  </body>
</html>
```

Release Notes:

- N/A

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -3366,7 +3366,19 @@ impl BufferSnapshot {
     pub fn syntax_layer_at<D: ToOffset>(&self, position: D) -> Option<SyntaxLayer<'_>> {
         let offset = position.to_offset(self);
         self.syntax_layers_for_range(offset..offset, false)
-            .filter(|l| l.node().end_byte() > offset)
+            .filter(|l| {
+                if let Some(ranges) = l.included_sub_ranges {
+                    ranges.iter().any(|range| {
+                        let start = range.start.to_offset(self);
+                        start <= offset && {
+                            let end = range.end.to_offset(self);
+                            offset < end
+                        }
+                    })
+                } else {
+                    l.node().start_byte() <= offset && l.node().end_byte() > offset
+                }
+            })
             .last()
     }
 
--- a/crates/language/src/buffer_tests.rs
+++ b/crates/language/src/buffer_tests.rs
@@ -2633,7 +2633,7 @@ fn test_language_scope_at_with_combined_injections(cx: &mut App) {
         buffer.set_language_registry(language_registry.clone());
         buffer.set_language(
             language_registry
-                .language_for_name("ERB")
+                .language_for_name("HTML+ERB")
                 .now_or_never()
                 .unwrap()
                 .ok(),
@@ -2748,6 +2748,50 @@ fn test_language_at_for_markdown_code_block(cx: &mut App) {
             let language = snapshot.language_at(point).unwrap();
             assert_eq!(language.name().as_ref(), "Rust");
         }
+
+        buffer
+    });
+}
+
+#[gpui::test]
+fn test_syntax_layer_at_for_injected_languages(cx: &mut App) {
+    init_settings(cx, |_| {});
+
+    cx.new(|cx| {
+        let text = r#"
+            ```html+erb
+            <div>Hello</div>
+            <%= link_to "Some", "https://zed.dev" %>
+            ```
+        "#
+        .unindent();
+
+        let language_registry = Arc::new(LanguageRegistry::test(cx.background_executor().clone()));
+        language_registry.add(Arc::new(erb_lang()));
+        language_registry.add(Arc::new(html_lang()));
+        language_registry.add(Arc::new(ruby_lang()));
+
+        let mut buffer = Buffer::local(text, cx);
+        buffer.set_language_registry(language_registry.clone());
+        buffer.set_language(
+            language_registry
+                .language_for_name("HTML+ERB")
+                .now_or_never()
+                .unwrap()
+                .ok(),
+            cx,
+        );
+
+        let snapshot = buffer.snapshot();
+
+        // Test points in the code line
+        let html_point = Point::new(1, 4);
+        let language = snapshot.language_at(html_point).unwrap();
+        assert_eq!(language.name().as_ref(), "HTML");
+
+        let ruby_point = Point::new(2, 6);
+        let language = snapshot.language_at(ruby_point).unwrap();
+        assert_eq!(language.name().as_ref(), "Ruby");
 
         buffer
     });
@@ -3655,7 +3699,7 @@ fn html_lang() -> Language {
 fn erb_lang() -> Language {
     Language::new(
         LanguageConfig {
-            name: "ERB".into(),
+            name: "HTML+ERB".into(),
             matcher: LanguageMatcher {
                 path_suffixes: vec!["erb".to_string()],
                 ..Default::default()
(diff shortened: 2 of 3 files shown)


crates/language/src/buffer_tests.rs
```diff
@@ -2633,7 +2627,7 @@ fn test_language_scope_at_with_combined_injections(cx: &mut App) {
         buffer.set_language_registry(language_registry.clone());
         buffer.set_language(
             language_registry
-                .language_for_name("ERB")
+                .language_for_name("HTML+ERB")
```
vitallium, line 2636: [Match the upstream](https://github.com/zed-extensions/ruby/issues/162).

crates/language/src/buffer_tests.rs
```diff
@@ -3655,7 +3680,7 @@ fn html_lang() -> Language {
 fn erb_lang() -> Language {
     Language::new(
         LanguageConfig {
-            name: "ERB".into(),
+            name: "HTML+ERB".into(),
```
vitallium, line 3702: [Match the upstream](https://github.com/zed-extensions/ruby/issues/162).

crates/language/src/buffer_tests.rs
```diff
@@ -3673,15 +3698,15 @@ fn erb_lang() -> Language {
     .with_injection_query(
         r#"
             (
-                (code) @injection.content
-                (#set! injection.language "ruby")
-                (#set! injection.combined)
+                (code) @content
+                (#set! "language" "ruby")
+                (#set! "combined")
```
vitallium, line 3722: [Match the upstream](https://github.com/zed-extensions/ruby/issues/162).

crates/language/src/buffer_tests.rs
```diff
@@ -3673,15 +3698,15 @@ fn erb_lang() -> Language {
...
             )
 
             (
-                (content) @injection.content
-                (#set! injection.language "html")
-                (#set! injection.combined)
+                (content) @content
+                (#set! "language" "html")
+                (#set! "combined")
```
vitallium, line 3728: [Match the upstream](https://github.com/zed-extensions/ruby/issues/162).

crates/language/src/syntax_map.rs
```diff
@@ -749,16 +751,20 @@ impl SyntaxSnapshot {
...
                                 .map(|r| {
                                     text.anchor_before(r.start_byte + step_start_byte)
                                         ..text.anchor_after(r.end_byte + step_start_byte)
                                 })
                                 .collect(),
-                        );
+                        )
+                    } else {
+                        None
+                    };
```
vitallium, line 767: Ensure we always provide accurate metadata for combined injections.

JosephTLyons:
Always happy to see your PRs, @vitallium! We'll take a look!

JosephTLyons:
Re-assigning, since Smit is out this week!

Veykril:
Gonna pass this on to @osiewicz as I am not familiar with this part yet

crates/language/src/buffer.rs
```diff
@@ -3348,7 +3348,17 @@ impl BufferSnapshot {
     pub fn syntax_layer_at<D: ToOffset>(&self, position: D) -> Option<SyntaxLayer<'_>> {
         let offset = position.to_offset(self);
         self.syntax_layers_for_range(offset..offset, false)
-            .filter(|l| l.node().end_byte() > offset)
+            .filter(|l| {
+                if let Some(ranges) = l.included_sub_ranges {
+                    ranges.iter().any(|range| {
+                        let start = range.start.to_offset(self);
+                        let end = range.end.to_offset(self);
+                        start <= offset && offset < end
```
osiewicz: Nit: it would be lovely if `end` was calculated lazily here.

osiewicz (changes requested):
LGTM, sans one comment.

crates/language/src/buffer.rs
vitallium: Yeah, makes sense. Fixed. Thanks!

osiewicz (approved):

PR #41117, opened 2025-10-24
git: Add support for git pull with rebase

Sathiyaraman-M:
- Adds a new action `git::PullRebase` which adds `--rebase` in the final command invoked by existing Git-Pull implementation.
- Includes the new action in "Fetch/Push" button in the Git Panel (screenshot below)
- Adds key-binding for `git::PullRebase` in all three platforms, following the existing key-binding patterns (`ctrl-g shift-down`) 
- Update git docs to include the new action.

Sidenote: This is my first ever OSS contribution

Screenshot:

image

---

Release Notes:

- Git: Added `git: pull rebase` for running `git pull --rebase`. 
- (Let's Git Together)

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -1593,8 +1593,13 @@ impl GitRepository for RealGitRepository {
             command
                 .envs(env.iter())
                 .current_dir(&working_directory?)
-                .args(["pull"])
-                .arg(if rebase { "--rebase" } else { "--no-rebase" })
+                .arg("pull");
+
+            if rebase {
+                command.arg("--rebase");
+            }
+
+            command
                 .arg(remote_name)
                 .arg(branch_name)
                 .stdout(smol::process::Stdio::piped())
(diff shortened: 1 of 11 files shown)


Sathiyaraman-M:
@cla-bot check

crates/git/src/repository.rs
```diff
@@ -1592,6 +1594,7 @@ impl GitRepository for RealGitRepository {
                 .envs(env.iter())
                 .current_dir(&working_directory?)
                 .args(["pull"])
+                .arg(if rebase { "--rebase" } else { "--no-rebase" })
```
cole-miller: I think we shouldn't pass `--no-rebase` if `rebase` is false here, so you get the configured behavior for the branch (which might be either merge or rebase)--this is how `git::Pull` works in Zed now and we should maintain that behavior.

cole-miller (changes requested):
Thanks, one piece of feedback but seems good otherwise!

crates/git/src/repository.rs
Sathiyaraman-M: Done

Sathiyaraman-M:
@cole-miller Gentle reminder: I've addressed the comment. Could you please take another look when you have time? Thanks!

cole-miller:
Thank you!

cole-miller (approved):

PR #41137, opened 2025-10-24
agent: Add save to markdown button for AI responses

GoWebProd:
Add a hover-revealed download button after each assistant message that allows saving the response to a markdown file.

- Add Clone derive to AssistantMessage and related types
- Implement to_markdown_without_thoughts to export responses without thinking blocks
- Add save_assistant_message_to_file method with file picker dialog
- Auto-append .md extension if not provided using Cow for efficiency
- Show save button only on message hover using VisibleOnHover trait

### DIFF
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -71,7 +71,7 @@ impl UserMessage {
     }
 }
 
-#[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


GoWebProd:
@cla-bot check

benbrandt:
Hi @GoWebProd thanks for taking the time to take a stab at this, but I don't think we want to introduce a button here, and you should be able to get this from the "Open thread as markdown" button already.

PR #41149, opened 2025-10-25
editor: Make hover popover delay strictly respect hover_popover_delay setting

deltamaya:
Previously, the hover popover delay was implemented using two overlapping timers, which caused the minimum delay to always be at least HOVER_REQUEST_DELAY_MILLIS, regardless of the hover_popover_delay setting.
This change updates the logic to wait for hover_popover_delay only, ensuring the total delay is always equals to hover_popover_delay . As a result, the hover popover now appears after the intended delay, matching the user's configuration more accurately.

Release Notes:

- Improved hover popover respecting settings delay correctly

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -27,7 +27,6 @@ use ui::{Scrollbars, WithScrollbar, prelude::*, theme_is_transparent};
 use url::Url;
 use util::TryFutureExt;
 use workspace::{OpenOptions, OpenVisible, Workspace};
-pub const HOVER_REQUEST_DELAY_MILLIS: u64 = 200;
 
 pub const MIN_POPOVER_CHARACTER_WIDTH: f32 = 20.;
 pub const MIN_POPOVER_LINE_HEIGHT: f32 = 4.;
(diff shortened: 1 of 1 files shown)


crates/editor/src/hover_popover.rs
```diff
@@ -287,27 +286,14 @@ fn show_hover(
...
+                    .timer(Duration::from_millis(hover_popover_delay))
                     .await;
-                total_delay
             };
 
             let hover_request = cx.update(|_, cx| provider.hover(&buffer, buffer_position, cx))?;
 
-            if let Some(delay) = delay {
-                delay.await;
-            }
```
Veykril, line 309: The intent here is so to start the lsp request before the actual delay, such that when the request takes longer its not as noticeable (as we won't have to wait for the request as the delay expired).

I think we want to retain the previous system, but try to split the user configured delay into two parts. That is we could try removing `HOVER_REQUEST_DELAY_MILLIS` from the user delay. That way we also do not spam the server with requests.

deltamaya: Thank you, I think I roughly understand how this works now.
But if we just subtract `HOVER_REQUEST_DELAY_MILLIS`, then if the user sets a delay smaller than that value, wouldn’t it cause every hover action to trigger an LSP request?
Or maybe we could make `HOVER_REQUEST_DELAY_MILLIS` smaller, so that the extra waiting time isn’t as noticeable — for example, set it to around 50 ms?

Veykril: Yea I think we do want a minimum here either way, reducing this to 50 is probably fine.

deltamaya: Thank you for your time! In the latest commit, I kept the early LSP request strategy while dynamically adjusting the early request timing to maintain consistency with the configured delay. I believe this change helps keep the application responsive without putting too much load on the LSP server.

Veykril (approved):
Thanks!

PR #41166, opened 2025-10-25
Do not report the itemDefaults LSP capability

### LINKED ISSUE #101: Selecting a completion option only gives you the symbol
### Relevance Confirmation

- [x] I confirmed that this is not an issue with the Eclipse JDT Language Server itself

### What happened?

After selecting a completion option like a method, you only get the symbol, but not the whole signature that includes parentheses.

I expected to get the the symbol along with the parentheses and the cursor positioned between said parentheses.

Did: `"Test".c`, then selected the suggestion `charAt`.

Got: `"Test".charAt`

https://github.com/user-attachments/assets/3791ada9-d119-46fb-9cbf-139a61adefd1

Expected: `"Test".charAt(<cursor here>)`

### Environment

Zed: Zed 0.208.5 
Platform: MacOS 26.0.1 
Java: OpenJDK 21.0.8

playdohface:
Closes https://github.com/zed-extensions/java/issues/101
This is a fix for an issue I encountered with JDTLS, but it turns out it might actually be a Zed issue at heart.
 
[This field of CompletionItem from the LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=/**%0A%09%20*%20The%20edit%20text,%3F%3A%20string%3B) does not exist within [Zed's lsp-types CompletionItem](https://github.com/zed-industries/lsp-types/blob/14af3dbeb635ff83606c4e0d9316460d7978f58f/src/completion.rs#L420). The spec suggests that itemDefaults capability implies honoring this field, but at the moment Zed does not. This causes issues with completions for JDTLS, and I imagine other language servers that make use of this capability, too. 
This PR simply removes reporting that capability in the default `initialize` request to language servers, until proper support for this feature is implemented within Zed. 
For some reference on the specifics, have a look at [this issue from the Java Extension](https://github.com/zed-extensions/java/issues/101).

Release Notes:

- N/A

### DIFF
--- a/crates/lsp/src/lsp.rs
+++ b/crates/lsp/src/lsp.rs
@@ -780,13 +780,7 @@ impl LanguageServer {
                         }),
                         insert_text_mode: Some(InsertTextMode::ADJUST_INDENTATION),
                         completion_list: Some(CompletionListCapability {
-                            item_defaults: Some(vec![
-                                "commitCharacters".to_owned(),
-                                "editRange".to_owned(),
-                                "insertTextMode".to_owned(),
-                                "insertTextFormat".to_owned(),
-                                "data".to_owned(),
-                            ]),
+                            item_defaults: None,
                         }),
                         context_support: Some(true),
                         dynamic_registration: Some(true),


SomeoneToIgnore (commented):
That is a great find, sorry for missing that corner of the spec.

Interesting to see that upstream library still has no such field: https://github.com/gluon-lang/lsp-types/blob/be7336e92a6ad23f214df19bcdceab17f39531a9/src/completion.rs#L513 and there seems to be no issues related.

On the bright side, we have our fork that we can modify however we want: https://github.com/zed-industries/lsp-types/blob/14af3dbeb635ff83606c4e0d9316460d7978f58f/src/completion.rs#L420

and we have also a quite extensive set of tests on the defauls in completion lists: https://github.com/zed-industries/zed/blob/e4c90be58a44ff05152ac92a7bf35c67f05dafab/crates/editor/src/editor_tests.rs#L18046 and in https://github.com/zed-industries/zed/blob/e4c90be58a44ff05152ac92a7bf35c67f05dafab/crates/project/src/project_tests.rs#L3371

The code for handling this is in https://github.com/zed-industries/zed/blob/e4c90be58a44ff05152ac92a7bf35c67f05dafab/crates/project/src/lsp_command.rs#L2197

If I understand things right, we can
* add a new field into the LSP types library
* use that in the code where things are handled the specific way
* adjust the test to reflect this

This does not seem to be too hard to do, so why not follow that path instead of removing the entire capability?


>  I imagine other language servers that make use of this capability, too

To comment on that a bit — IIRC only Java language servers use such approach to completion and raise issues in Zed, so seems relatively niche.

playdohface:
@SomeoneToIgnore actually implementing this capability is of course superior to just opting out 👍  
Thanks for all the pointers, I'll give this a try

playdohface:
Closing this in favor of https://github.com/zed-industries/zed/pull/41187

PR #41177, opened 2025-10-25
Avoid menu clipping with debugger welcome panel on resize

abdelq:
Before:
before

After:
after

Release Notes:

- N/A


Anthony-Eid:
Thanks for catching this!

PR #41187, opened 2025-10-25
Handle itemDefaults in CompletionList according to spec

### LINKED ISSUE #101: Selecting a completion option only gives you the symbol

playdohface:
Closes https://github.com/zed-extensions/java/issues/101

Previously Zed did not handle resolving CompletionList with itemDefaults correctly according to the [LSP-Spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=/**%0A%09%20*%20The%20edit%20text,%3F%3A%20string%3B).
When a `CompletionList` is provided by the server, that includes ranges as `itemDefaults`, the field to use for the snippet to insert as `newText` is in the `textEditText` field, with a fallback to `label` if that does not exist (`insertText` is ignored). 

Release Notes:

- Fixed Java language severs' completion defaults handling on Zed's side

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9452,7 +9452,7 @@ dependencies = [
 [[package]]
 name = "lsp-types"
 version = "0.95.1"
-source = "git+https://github.com/zed-industries/lsp-types?rev=0874f8742fe55b4dc94308c1e3c0069710d8eeaf#0874f8742fe55b4dc94308c1e3c0069710d8eeaf"
+source = "git+https://github.com/zed-industries/lsp-types?rev=b71ab4eeb27d9758be8092020a46fe33fbca4e33#b71ab4eeb27d9758be8092020a46fe33fbca4e33"
 dependencies = [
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore (approved):
Superb, thank you for the forward-fix and quick turnaround!

PR #40454, opened 2025-10-17
Add dock state to workspace context

FloppyDisco:
I love keybindings.

I spend way to much time thinking about them.

I also REALLY like working in Zed. 

so far, however, I have found the key context system in Zed to be less flexible than in VSCode.
the HUGE context that is available in VSCode helps you create keybindings for very specific targeted scenarios.

the tree like structure of the Zed key context means you loose some information as focus moves throughout the application.
For example, it is not currently possible to create a keybinding in the editor that will only work when one of the Docks is open, or if a specific dock is open.

this would be useful in implementing solutions to ideas like #24222 
we already have an action for moving focus to the dock, and we have an action for opening/closing the dock, but to my knowledge (very limited lol) we cannot determine if that dock *is open* unless we are focused on it.

I think it is possible to create a more flexible key binding system by adding more context information to the higher up context ancestors.
while:
``` 
Workspace right_dock=GitPanel
    Dock
        GitPanel
            Editor
```
may seem redundant, it actually communicates fundamentally different information than:
```
Workspace right_dock=GitPanel
    Pane
        Editor
```  

the first says "the GitPanel is in the right hand dock AND IT IS FOCUSED", 
while the second means "Focus is on the Editor, and the GitPanel just happens to be open in the right hand dock"

This change adds a new set of identifiers to the `Workspace` key_context that will indicate which docks are open and what is the specific panel that is currently visible in that dock.

examples:
- `left_dock=ProjectPanel`
- `bottom_dock=TerminalPanel`
- `right_dock=GitPanel`

in my testing the following types of keybindings seem to be supported with this change:
```jsonc
// match for any value of the identifier
"context": "Workspace && bottom_dock"
"context": "Workspace && !bottom_dock"
// match only a specific value to an identifier
"context": "Workspace && bottom_dock=TerminalPanel"
// match only in a child context if the ancestor workspace has the correct identifier
"context": "Workspace && !bottom_dock=DebugPanel > Editor"
```

some screen shots of the context matching in different circumstances:
Screenshot 2025-10-16 at 23 20 34
Screenshot 2025-10-16 at 23 20 57
Screenshot 2025-10-16 at 23 21 37
Screenshot 2025-10-16 at 23 21 52
Screenshot 2025-10-16 at 23 22 38

the persistent_name values for `ProjectPanel` and `OutlinePanel` needed to be updated to not have a space in them in order to pass the `Identifier` check. all the other Panels already had names that did not include spaces, so it just makes these conform with the other ones.

[...]

PR #41193, opened 2025-10-25
Add buffer_search_deployed key context

FloppyDisco:
Release Notes:

- Pane key context now includes 'buffer_search_deployed' identifier

The goal of this PR is to add a new identifier in the key context that will let the user target when the BufferSearchBar is deployed even if they are not focused on it.

requested in #36930

Same rational as #40454 this will allow users to make more flexible keybindings, by including some additional information higher up the key context tree.

i thought adding this context to `Pane` seemed more appropriate  than `Editor` since `Terminal` also has a `BufferSearchBar`; however, I ran into some import issues between BufferSearchBar, Search, Pane, and Workspace which made it difficult to implement adding this context directly inside `Pane`'s  render function.

instead i added a new method called `contributes_context` to `ToolbarItem` which will allow any toolbar item to add additional context to the `Pane` level, which feels like it might come in handy. 

here are some screen shots of the context being displayed in the Editor and the Terminal

Screenshot 2025-10-25 at 14 34 03

Screenshot 2025-10-25 at 12 32 21

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -468,6 +468,12 @@ impl Focusable for BufferSearchBar {
 }
 
 impl ToolbarItemView for BufferSearchBar {
+    fn contribute_context(&self, context: &mut KeyContext, _cx: &App) {
+        if !self.dismissed {
+            context.add("buffer_search_deployed");
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


ConradIrwin:
nice, thank you!

PR #39535, opened 2025-10-04
Dismiss formatting error notifications from language server on successful format requests

tomatitito:
This is a partial fix to the issue reported in #38769. I should say that I somewhat accidentally agentically engineered this. I just wanted to ask the agent a couple of questions about how to approach this and he was quick to offer a solution that almost worked. I was then able to actually make it work. I looked at it and it seems good to me, so I went ahead and opened this PR.

There are not tests yet, but if this approach is acceptable to you guys then I would be happy to add them. I did however check it locally and it worked.

The general approach taken is that the `LspStore`, upon receiving a request to format the local buffer, will emit a `WillFormatWith(LanguageServerId)` event, which is propagated through the project to the workspace, where the handler closes any open notifications for the given language server.

If this is something that you would consider merging, I would be happy to make any changes neccessary.

(Partly) Closes #38769

Release Notes:

- Fixed a bug where error notifications for failing formatting requests are not closing even after a successful format has been made.

PR #41198, opened 2025-10-25
Fix dismissal of language server notifications

### LINKED ISSUE #38769: LS Server warning popup can not be silenced / does not dismis when stale
### Summary

Image

When using a language server (here for Elixir) and the code could not be formatted due to a typo, then the above popup will open and interupts the focus of the user. There are a number of usability problems with it:

- There is no settings to disable it
- It also would be better if on a subsequent run of the formatter is without errors to dismiss the popup. There is little sense in keeping a warning to a formatting issue that is not present anymore

Zed: v0.204.5 (Zed) 
OS: macOS 15.6
Memory: 64 GiB
Architecture: aarch64

tomatitito:
Closes #38769

This is my second attempt at this, this time building on the suggestions by @SomeoneToIgnore (which can be found [here](https://github.com/zed-industries/zed/pull/39535#pullrequestreview-3316127584)).

The idea is that when a notification is shown to the user, and if this notification does not require input from the user, to start a timer and send a dismisssal event when the timer finishes.

I also took care of cancelling the dismissal task in case the user manually closes the notification.

There is no test, but I would be happy to add one if you consider this approach viable.

Release Notes:

- Fixes a bug where notifications from language servers are not closed.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -2074,7 +2074,7 @@
     // Whether to show the LSP servers button in the status bar.
     "button": true,
     "notifications": {
-      // Wether to automatically dismiss language server notifications.
+      // Whether to automatically dismiss language server notifications.
       "auto_dismiss": "always"
     }
   },
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -2072,7 +2072,11 @@
   // Common language server settings.
   "global_lsp_settings": {
     // Whether to show the LSP servers button in the status bar.
-    "button": true
+    "button": true,
+    "notifications": {
+      // Wether to automatically dismiss language server notifications.
+      "auto_dismiss": "always"
+    }
   },
   // Jupyter settings
   "jupyter": {
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -967,17 +967,18 @@ impl LocalLspStore {
                     async move {
                         let actions = params.actions.unwrap_or_default();
                         let (tx, rx) = smol::channel::bounded(1);
-                        let request = LanguageServerPromptRequest {
-                            level: match params.typ {
-                                lsp::MessageType::ERROR => PromptLevel::Critical,
-                                lsp::MessageType::WARNING => PromptLevel::Warning,
-                                _ => PromptLevel::Info,
-                            },
-                            message: params.message,
-                            actions,
-                            response_channel: tx,
-                            lsp_name: name.clone(),
+                        let level = match params.typ {
+                            lsp::MessageType::ERROR => PromptLevel::Critical,
+                            lsp::MessageType::WARNING => PromptLevel::Warning,
+                            _ => PromptLevel::Info,
                         };
+                        let request = LanguageServerPromptRequest::new(
+                            level,
+                            params.message,
+                            actions,
+                            name.clone(),
+                            tx,
+                        );
 
                         let did_update = this
                             .update(&mut cx, |_, cx| {
@@ -1004,17 +1005,13 @@ impl LocalLspStore {
                     let mut cx = cx.clone();
 
                     let (tx, _) = smol::channel::bounded(1);
-                    let request = LanguageServerPromptRequest {
-                        level: match params.typ {
-                            lsp::MessageType::ERROR => PromptLevel::Critical,
-                            lsp::MessageType::WARNING => PromptLevel::Warning,
-                            _ => PromptLevel::Info,
-                        },
-                        message: params.message,
-                        actions: vec![],
-                        response_channel: tx,
-                        lsp_name: name,
+                    let level = match params.typ {
+                        lsp::MessageType::ERROR => PromptLevel::Critical,
+                        lsp::MessageType::WARNING => PromptLevel::Warning,
+                        _ => PromptLevel::Info,
                     };
+                    let request =
+                        LanguageServerPromptRequest::new(level, params.message, vec![], name, tx);
 
                     let _ = this.update(&mut cx, |_, cx| {
                         cx.emit(LspStoreEvent::LanguageServerPrompt(request));
@@ -13197,6 +13194,7 @@ struct LspBufferSnapshot {
 /// A prompt requested by LSP server.
 #[derive(Clone, Debug)]
 pub struct LanguageServerPromptRequest {
+    pub id: u64,
     pub level: PromptLevel,
     pub message: String,
     pub actions: Vec<MessageActionItem>,
@@ -13205,6 +13203,23 @@ pub struct LanguageServerPromptRequest {
 }
 
 impl LanguageServerPromptRequest {
+    pub fn new(
+        level: PromptLevel,
+        message: String,
+        actions: Vec<MessageActionItem>,
+        lsp_name: String,
+        response_channel: Sender<MessageActionItem>,
+    ) -> Self {
+        let id = rand::random::<u64>();
+        LanguageServerPromptRequest {
+            id,
+            level,
+            message,
+            actions,
+            lsp_name,
+            response_channel,
+        }
+    }
     pub async fn respond(self, index: usize) -> Option<()> {
         if let Some(response) = self.actions.into_iter().nth(index) {
             self.response_channel.send(response).await.ok()
@@ -13221,13 +13236,7 @@ impl LanguageServerPromptRequest {
         lsp_name: String,
     ) -> Self {
         let (tx, _rx) = smol::channel::unbounded();
-        Self {
-            level,
-            message,
-            actions,
-            lsp_name,
-            response_channel: tx,
-        }
+        LanguageServerPromptRequest::new(level, message, actions, lsp_name, tx)
     }
 }
 impl PartialEq for LanguageServerPromptRequest {
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -140,6 +140,7 @@ const SERVER_LAUNCHING_BEFORE_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5
 pub const SERVER_PROGRESS_THROTTLE_TIMEOUT: Duration = Duration::from_millis(100);
 const WORKSPACE_DIAGNOSTICS_TOKEN_START: &str = "id:";
 const SERVER_DOWNLOAD_TIMEOUT: Duration = Duration::from_secs(10);
+static NEXT_PROMPT_REQUEST_ID: AtomicUsize = AtomicUsize::new(0);
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
 pub enum ProgressToken {
@@ -13194,7 +13195,7 @@ struct LspBufferSnapshot {
 /// A prompt requested by LSP server.
 #[derive(Clone, Debug)]
 pub struct LanguageServerPromptRequest {
-    pub id: u64,
+    pub id: usize,
     pub level: PromptLevel,
     pub message: String,
     pub actions: Vec<MessageActionItem>,
@@ -13210,7 +13211,7 @@ impl LanguageServerPromptRequest {
         lsp_name: String,
         response_channel: Sender<MessageActionItem>,
     ) -> Self {
-        let id = rand::random::<u64>();
+        let id = NEXT_PROMPT_REQUEST_ID.fetch_add(1, atomic::Ordering::AcqRel);
         LanguageServerPromptRequest {
             id,
             level,
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -1,5 +1,4 @@
-//! LSP store provides unified access to the language server protocol.
-//! The consumers of LSP store can interact with language servers without knowing exactly which language server they're interacting with.
+//! LSP store provides unified access to the language server protocol. //! The consumers of LSP store can interact with language servers without knowing exactly which language server they're interacting with.
 //!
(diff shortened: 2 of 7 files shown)


crates/workspace/src/notifications.rs
```diff
@@ -96,6 +96,30 @@ impl Workspace {
                 }
             })
             .detach();
+
+            if let Ok(prompt) =
+                AnyEntity::from(notification.clone()).downcast::<LanguageServerPrompt>()
```
SomeoneToIgnore, line 105: Cannot we just do `notification.downcast`?

crates/workspace/src/notifications.rs
```diff
@@ -96,6 +96,30 @@ impl Workspace {
...
+                if prompt
+                    .read(cx)
+                    .request
+                    .as_ref()
+                    .map_or(false, |r| r.actions.is_empty())
+                {
+                    let task = cx.spawn({
+                        let id = id.clone();
+                        async move |this, cx| {
+                            gpui::Timer::after(Duration::from_secs(5)).await;
```
SomeoneToIgnore: Let's use the form already used in the project:, 

`cx.background_executor().timer(Duration::from_millis(5000))`

It also seems to me that we can extract this magic number into a constant?
Maybe later it needs to be customized, so this would be a good preparation step.

crates/workspace/src/notifications.rs
```diff
@@ -96,6 +96,30 @@ impl Workspace {
...
             .detach();
+
+            if let Ok(prompt) =
+                AnyEntity::from(notification.clone()).downcast::<LanguageServerPrompt>()
+            {
+                if prompt
+                    .read(cx)
+                    .request
+                    .as_ref()
+                    .map_or(false, |r| r.actions.is_empty())
```
SomeoneToIgnore:
```suggestion
                    .is_some_and(|request| request.actions.is_empty())
```

Overall, wonder if we should add some field explicitly there instead, `timeout` or `autohide`, but this way it can work also.

crates/workspace/src/notifications.rs
```diff
@@ -246,13 +272,20 @@ impl LanguageServerPrompt {
                 .await
                 .context("Stream already closed")?;
 
-            this.update(cx, |_, cx| cx.emit(DismissEvent))?;
+            this.update(cx, |this, cx| {
+                this.cancel_dismiss_task();
```
SomeoneToIgnore: We don't seem to bother clearing those for toasts, but if we do now, let's at least create a method for this?

SomeoneToIgnore (commented):
Thank you, I think this looks quite viable and we can see how things work for other people.
Worst case of loads negative feedback, we might need to add some settings toggle to restore old, dismiss-less, behavior, but that can come later.

My main concern is how those messages get their IDs: https://github.com/zed-industries/zed/blob/f503c65924fff9799b359b6c83994b2eaf7af7c8/crates/workspace/src/workspace.rs#L1267-L1269

That seems relatively error-prone as that `lsp_name` is a language server name: hence, we will auto dismiss all message from all language servers with that name.
This is a good test case, IMO: spawn multiple notifications (skipping time between spawns) and ensure only one gets dismissed per time chunk.
I think we should add a notification method/kind here at least.

crates/workspace/src/notifications.rs
tomatitito: Unfortunately not, because `notification` is an `Entity<V>` and that `V` cannot be cloned. I could put this logic in `LanguageServerPrompt::new` since from there we know what kind of a request we have and if it does require user interaction or not. We could then start the timer from inside the `new` method, right after it is created. Would this be better?

crates/workspace/src/notifications.rs
tomatitito: I did use the background executor in my first commit and changed it in the hope that I could prevent that one test from hanging. But sure, I'll use the background executor. And a constant is fine, of course, too.

crates/workspace/src/notifications.rs
tomatitito: I'm not sure I understand you here. What would be the name of the method and where would it need to go?

crates/workspace/src/notifications.rs
SomeoneToIgnore: Hanging most probably now happens due to the fact that we need to wait real seconds which we never do.
`cx.executor().advance_clock(Duration::from_secs(CONSTANT));` placed in the right place should help with this.

crates/workspace/src/notifications.rs
SomeoneToIgnore: I wanted a method that combines both lines, 

```rs
this.cancel_dismiss_task();
cx.emit(DismissEvent);
```

hence they have to happen "atomically" for the language server notification.
`dismiss_notification`?

tomatitito: Ah, ok. Sure, I'll do that.

crates/workspace/src/notifications.rs
```diff
@@ -331,10 +376,11 @@ impl Render for LanguageServerPrompt {
                                                 }
                                             })
                                             .on_click(cx.listener(
-                                                move |_, _: &ClickEvent, _, cx| {
+                                                move |this, _: &ClickEvent, _, cx| {
                                                     if suppress {
                                                         cx.emit(SuppressEvent);
                                                     } else {
+                                                        this.cancel_dismiss_task();
```
SomeoneToIgnore: The point to have that new, `dismiss_notification`, method created is to use it, right?

Here seems the place where it can be used.

crates/workspace/src/notifications.rs
```diff
@@ -246,13 +280,24 @@ impl LanguageServerPrompt {
...
         .await
         .log_err();
     }
+
+    fn dismiss_notification(&mut self, cx: &mut Context<Self>) {
+        self.cancel_dismiss_task();
+        cx.emit(DismissEvent);
+    }
+
+    fn cancel_dismiss_task(&mut self) {
```
SomeoneToIgnore: Well, that can be inlined into `dismiss_notification`, right?

SomeoneToIgnore (commented):
I won't look at this from now on, as it's a draft still and I'm not sure why.
Waiting for the test and un-drafring, lmk if any further help is needed.

tomatitito:
@SomeoneToIgnore Sorry for taking so long, I got super busy all of a sudden. I added a test for the auto dismissal following the strategy you proposed. Did I see that correctly that the github `tun_tests` action failed because you cancelled it?

SomeoneToIgnore:
image

tomatitito:
@SomeoneToIgnore Oh Man, sorry. I rebased and now I have all these changes. Let me see how I get out of this...

tomatitito:
@SomeoneToIgnore Ok, I think now we can talk again. Thanks a lot for your patience! I ended up adding the `LanguageServerPromptRequest::test` method to be able to get a `LanguageServerPrompt` for testing. If there is a better way, please let me know. Did you fix that problem with the hanging tests on linux?

crates/project/src/lsp_store.rs
```diff
@@ -1,5 +1,4 @@
-//! LSP store provides unified access to the language server protocol.
-//! The consumers of LSP store can interact with language servers without knowing exactly which language server they're interacting with.
+//! LSP store provides unified access to the language server protocol. //! The consumers of LSP store can interact with language servers without knowing exactly which language server they're interacting with.
```
SomeoneToIgnore: This seems to be some accidental change.

SomeoneToIgnore (approved):
Superb, thank you for coming up with the test, that looks quite promising and what we would expect.

Ideally, I think we need to fix what was mentioned in https://github.com/zed-industries/zed/pull/41198#pullrequestreview-3383323371

* Currently, dismissing one language server notification will dismiss the rest: I suspect it's not something we want?
Maybe, even worth writing a test for that?
What's odd is that all such notifications now replace each other, so previous are lost — helps with preserving the space but odd.

* As someone may want to suppress notifications from a language server, may there be people that want them to stay always?

I'd add some settings fallback for that, maybe along with the 3rd mode: "never", and keep this in the language settings, so one could suppress spammy messages per language.

------------------------

Overall, this is an increment already, let's merge and see how it works.
Thank you for creating this — will wait for Wednesday release to merge this in, just in case.

crates/project/src/lsp_store.rs
tomatitito: Yay, that's great news! I will continue working on the open points that still need to be fixed.

tomatitito:
@SomeoneToIgnore I came up with a solution for this:

> Currently, dismissing one language server notification will dismiss the rest: I suspect it's not something we want?
Maybe, even worth writing a test for that?

I compute a random id for every prompt request and make this the notification id (instead of the lsp name). I tried it manually and also added a test that confirms that it works. Is this an acceptable solution? If so, I will also adapt the old test.

SomeoneToIgnore (commented):
Sorry for the delay, had quite a busy week.

Nice idea overall, except `rand` seems relatively heavy and not 100% proof from duplicates.
It seems that a regular `static NEXT_MESSAGE_ID AtomicUsize = ...` on which we do some `.fetch_add(1, atomic::Ordering::AcqRel)` is enough?

-----------------------

With this done, one last concern left would be overly spammy servers: we've ungrouped the notifications and now each will stay onscreen 5 seconds.
If we get too many notifications at once, that will still be quite spammy.

So, I still keep wondering on the ways to configure this behavior in Zed.
I have not any formed idea though, but overall seems that we might want to use settings to 
* turn those messages on and off
* configure the 5 second timeout
* ?? turn the grouping on and off (not sure how important)

There is https://github.com/zed-industries/zed/blob/1ce58a88ccdfb7e45106d66fa44c9db35b4f94b9/crates/project/src/project_settings.rs#L108-L116 which might be a good place to add another field + struct?

We should document in Rust the new structs fields we've added (shows in the completions later) and add entries into `default.json` with the defaults, and into `configuring-zed.md`. 

This way, we'll get docs and settings for people to configure the behavior.

tomatitito:
> It seems that a regular static NEXT_MESSAGE_ID AtomicUsize = ... on which we do some .fetch_add(1, atomic::Ordering::AcqRel) is enough?

Uhm, thanks for pointing this out, I didn't know about this. Yes, that's much better! I substituted the random numbers by the AtomicUsize counter. 

------------------------

As for the remaining issue: Hmm, it's hard to say what the best way to go is. I am trying to think about what I as a user would like, but I'm also not sure. I would start with making the auto-dismissal configurable just by turning it on and off in the settings. So far it seems pretty uncontroversial to me. 

Adding the ability to configure the timeout and maybe even to "group" or "ungroup" the notifications would be quite impressive in terms of how much users are able to configure zed. I probably would not have proposed this myself, but now that I think about it, why not...

tomatitito:
Hey @SomeoneToIgnore , sorry for taking so long. I'm very busy and it was hard for me to find time to work on this. However, I finally got around to work on the settings part. I have made the auto dismisssal configurable using the global lsp settings. Default is to `Always` dismiss, but you can also set this to `Never` or to `After(...)` and specify a delay in milliseconds.

What's missing still is to enable this on a language server basis. and also the "grouping" or "ungrouping" of notifications.

Not sure why this one test fails, locally all tests pass. I'll look into it...

assets/settings/default.json
```diff
@@ -2072,7 +2072,11 @@
   // Common language server settings.
   "global_lsp_settings": {
     // Whether to show the LSP servers button in the status bar.
-    "button": true
+    "button": true,
+    "notifications": {
+      // Whether to automatically dismiss language server notifications.
+      "auto_dismiss": "always"
```
SomeoneToIgnore, line 2078: I have launched Zed and tried to complete things in the local settings file:

Image

sounds suboptimal, as we miss the timeout setting there.

Given the complexity of merging a numerical value with the rest, I wonder if we should re-approach and use just a single setting value: `dismiss_timeout_ms: usize` values.

This also helps with the fact that `always` is a duplicate of `after`, in reality + helps to remove the 5000 hardcoded constant: instead, that will be a default value for the new setting.


---------------------

Besides that, 

Image

is still not complete, as the md file is not updated yet.

SomeoneToIgnore (commented):
Thank you for the persistence!
I think it's moving great, we have a single item left: to tidy up the settings, and we're ready!

No worries about the timings and no rush — I am also quite busy interim.

The test failure is unrelated to PR, never mind it — I'd suggest merging `main` in, just in case it got fixed there, but not sure if that's true.

Lastly, let me know if I should help you with this last mile — I can find some time later.

assets/settings/default.json
tomatitito: Hey @SomeoneToIgnore , thanks for your answer and your patience! I will go ahead and rebase onto `main` (or is merging `main` in the recommended approach?). 

-------------------

I am happy to go with your suggestions and use just a single `dismiss_timeout_ms` configuration. However, in this case it would not really be possible anymore to disable the dismissal. (Well, you could set it to a really high number, of course, which might effectively be the same, but still). But I personally think it wouldn't be that bad. If people actually complain about it, we could add it later?

---- 
And yes, I forgot to put the into `configuring-zed.md`, sorry. I'll do that, too, of course. 

---------
I would love to get this through this year!

SomeoneToIgnore: Well, of course we can add a way to disable things, just have to find the semantics.

I'm a bit tired to babysit each and every detail here, so I am very hopeful you can find `null`, `-1` or whatever else combination that works both for settings and the code.
After all, the problem is not novel and default.json has to contain examples.

SomeoneToIgnore: As for the rebase or merge, it does not matter as we squash the PR later when merging.

PR #41199, opened 2025-10-25
gpui: Implement trackpad pinch-to-zoom on macOS

kognise:
Implements trackpad pinch-to-zoom on macOS. Windows and Linux should be straightforward but I don't have requisite test devices with me right now.

* All behavior tested on macOS. Does not cause any issues with existing scroll handling.
* Follows the same propagation and blocking behavior as scroll events. Documentation has been updated accordingly.
* I considered integrating into the scroll system, similar to what JavaScript does, but using a separate event is both more logical (the events are separate on every platform) and more backwards-compatible.
* The `ZoomDelta` is temporarily tagged `#[non_exhaustive]` because implementations on future platforms might handle zoom deltas differently from macOS (e.g. pixel distance instead of a magnification change). Consumers of this API can just have a `_ => {}` match arm, which worst-case means their apps will continue to not have zoom functionality on Windows/Linux without intervention.

Release Notes:

- N/A


kognise:
@cla-bot check

PR #41202, opened 2025-10-26
gpui: Add debug assertion to Window::on_action and make docs consistent

kognise:
Improves formatting consistency across various docs, fixes some typos, and adds a missing `debug_assert_paint` to `Window::on_action` and `Window::on_action_when`.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/window.rs
+++ b/crates/gpui/src/window.rs
@@ -3445,7 +3445,7 @@ impl Window {
         )));
     }
 
-    /// Register a capturing key event listener on this node for the next frame. The type of event
+    /// Register a key event listener on this node for the next frame. The type of event
     /// is determined by the first parameter of the given listener. When the next frame is rendered
     /// the listener will be cleared.
     ///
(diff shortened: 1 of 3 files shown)


crates/gpui/src/window.rs
```diff
@@ -3432,25 +3432,25 @@ impl Window {
...
                 if let Some(event) = event.downcast_ref() {
-                    handler(event, phase, window, cx)
+                    listener(event, phase, window, cx)
                 }
             },
         )));
     }
 
-    /// Register a key event listener on the window for the next frame. The type of event
+    /// Register a capturing key event listener on this node for the next frame. The type of event
```
mikayla-maki: This seems incorrect? `on_key_event` and `on_mouse_event` are non capturing by default:

https://github.com/zed-industries/zed/blob/7b6ef5eaabc3755515170709b3e880075b86a69b/crates/gpui/src/window.rs#L3912-L3919

mikayla-maki (changes requested):

crates/gpui/src/window.rs
kognise: I guess this is true; the listener will be triggered during both the capture and bubble phases. Updating the wording.

mikayla-maki (approved):
Thanks!

PR #41204, opened 2025-10-26
editor: Use a single newline between each copied line from a multi-cursor selection

### LINKED ISSUE #40923: Lines copied with multi-cusor pastes with extra newlines
### Summary

After copying entire lines (i.e. without selection) with multiple cursors, pasting with a different number of cursors adds extra newlines between the pasted lines.

### Description

As noted by @notpeter in https://github.com/zed-industries/zed/issues/34797#issuecomment-3112343273 (emphasis mine):

> Multicursor cut/copy in vscode does not use this alternate line-wise paste mode and pastes to the cursor position(s) like a normal clipboard action. Multicursor in Jetbrains and Sublime have an odd behavior -- **when cut/copying two adjacent lines if you paste to a single cursor you will get an extra line break between the pasted lines. (this makes cut/paste non-symmetrical).** If pasting two lines to two cursors, Sublime will give you four lines separated by two linebreaks each; jetbrains will give you two lines seperated by two newlines each. I find all of these behaviors unintuitive.

Steps to reproduce:
1. Place multiple (N > 1) cursors on separate lines
2. Copy (without selecting any text, which copies entire lines)
3. Paste using a different number of cursors (including just 1 cursor)

**Expected Behavior**: N lines are pasted at each cursor
**Actual Behavior**: 2N-1 lines are pasted at each cursor (the original N lines are interleaved with extra newlines)

Zed: v0.209.4 (Zed) 
OS: macOS 15.7.1
Memory: 16 GiB
Architecture: x86_64

seanstrom:
Closes #40923

Release Notes:

- Fixed the amount of newlines between copied lines from a multi-cursor selection of multiple full-line copies.

---

https://github.com/user-attachments/assets/ab7474d6-0e49-4c29-9700-7692cd019cef

### DIFF
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -26987,6 +26987,43 @@ async fn test_copy_line_without_trailing_newline(cx: &mut TestAppContext) {
 }
 
 #[gpui::test]
+async fn test_multi_selection_copy_with_newline_between_copied_lines(cx: &mut TestAppContext) {
+    init_test(cx, |_| {});
+
+    let mut cx = EditorTestContext::new(cx).await;
+
+    cx.set_state("line1\nline2\nline3\nˇ");
+
+    cx.update_editor(|e, window, cx| {
+        e.change_selections(SelectionEffects::no_scroll(), window, cx, |s| {
+            s.select_display_ranges([
+                DisplayPoint::new(DisplayRow(0), 0)..DisplayPoint::new(DisplayRow(0), 0),
+                DisplayPoint::new(DisplayRow(1), 0)..DisplayPoint::new(DisplayRow(1), 0),
+                DisplayPoint::new(DisplayRow(2), 0)..DisplayPoint::new(DisplayRow(2), 0),
+            ]);
+        });
+    });
+
+    cx.update_editor(|e, window, cx| e.copy(&Copy, window, cx));
+
+    let clipboard_text = cx
+        .read_from_clipboard()
+        .and_then(|item| item.text().as_deref().map(str::to_string));
+
+    assert_eq!(
+        clipboard_text,
+        Some("line1\nline2\nline3\n".to_string()),
+        "Copying multiple lines should include a single newline between lines"
+    );
+
+    cx.set_state("lineA\nˇ");
+
+    cx.update_editor(|e, window, cx| e.paste(&Paste, window, cx));
+
+    cx.assert_editor_state("lineA\nline1\nline2\nline3\nˇ");
+}
+
+#[gpui::test]
 async fn test_end_of_editor_context(cx: &mut TestAppContext) {
     init_test(cx, |_| {});
 
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -26992,19 +26992,36 @@ async fn test_multi_selection_copy_with_newline_between_copied_lines(cx: &mut Te
 
     let mut cx = EditorTestContext::new(cx).await;
 
-    cx.set_state("line1\nline2\nline3\nˇ");
-
-    cx.update_editor(|e, window, cx| {
-        e.change_selections(SelectionEffects::no_scroll(), window, cx, |s| {
-            s.select_display_ranges([
-                DisplayPoint::new(DisplayRow(0), 0)..DisplayPoint::new(DisplayRow(0), 0),
-                DisplayPoint::new(DisplayRow(1), 0)..DisplayPoint::new(DisplayRow(1), 0),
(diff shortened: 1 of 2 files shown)


seanstrom:
One problem that I'm noticing with this change:
* Seems like the editor crashes when attempting to paste into multiple cursors after copying multiple entire lines with multi-cursors.
  * Not sure what the root cause is, but perhaps the `do_paste` handler is expecting some of the newlines that were made conditional with this change.
  * I'll attempt some more debugging and exploring to see what's the expected format of clipboard content.

seanstrom:
I've posted a fix to the crash, but the code itself is not worthy to be merged 😄 

But here's my notes on what I've found when digging through the code for `do_copy` and `do_paste`:
* Prior to my changes in this code, we would always add a newline (after the first selection) for each selection. Which meant that copying two entire lines would produce an extra newline between the lines.
  * I thought this was a bug, since this extra newline is not counted in the selection length, but it seems to be an intentional padding between selections, since the `do_paste` handler seems to sometimes account for this extra newline when pasting. 
  * Here's the code snipped for the extra newline: https://github.com/zed-industries/zed/blob/4724aa5cb83ec82caed204e252d8d041e258281c/crates/editor/src/editor.rs#L12507-L12512
 * With my initial changes, I was conditionally adding that extra newline, since the current logic for copying entire lines will copy the entire line + plus the very beginning of the next line (effectively copying the newline).
   * However this would lead to a crash, because that intentional piece of newline padding would not be present during the `do_paste` operation, which would cause an out-of-bounds crash because the `do_paste` handler is expecting the extra newline.
   * Here's the snippet that leads to an out-of-bounds crash: https://github.com/zed-industries/zed/blob/4724aa5cb83ec82caed204e252d8d041e258281c/crates/editor/src/editor.rs#L12582-L12587
     * Notice how we're doing: 
       ```rust
         let end_offset = start_offset + clipboard_selection.len;
         // ...
         start_offset = end_offset + 1;
       ```
* I've tried modifying the `do_paste` handler, instead of changing `do_copy`, but I was running into some issues with properly extracting the clipboard selections without the extra newline and pasting their contents into the correct cursor positions.
  * The solution would essentially allow for the extra newlines, and when the clipboard selections count did not match the current cursor count, it would attempt to scan over the clipboard text and create new editor edits for each clipboard selection (effectively removing the extra newline).
  
---

That all being said, I'm wondering what the best approach would be to resolve this issue.
* Do we keep this extra newline padding in the clipboard contents and allow the paste handler to strip it out?
* Or do we adapt the paste handler to not assume there's an extra newline?

crates/editor/src/editor_tests.rs
```diff
@@ -26962,6 +26962,43 @@ async fn test_copy_line_without_trailing_newline(cx: &mut TestAppContext) {
...
 
+#[gpui::test]
+async fn test_multi_selection_copy_with_newline_between_copied_lines(cx: &mut TestAppContext) {
+    init_test(cx, |_| {});
+
+    let mut cx = EditorTestContext::new(cx).await;
+
+    cx.set_state("line1\nline2\nline3\nˇ");
+
+    cx.update_editor(|e, window, cx| {
```
SomeoneToIgnore: This all is not needed, as we can place multiple `ˇ` markers into the input text for `set_state` with the same result.

Same works in `assert_editor_state`.

SomeoneToIgnore (commented):
That's quite an investigation for such small change, thank you for checking on all this!

Logically, it seems that we should fix the receiving end too: we should copy into the buffer the "right" text already, as the user might want to paste elsewhere, and unintuitive extra newlines would be there if not trimmed.

As a bonus, something similar happens on the Vim emulation side?
https://github.com/zed-industries/zed/blob/5e41ce17e35685de1f62662e96f6de0b9cef8749/crates/vim/src/normal/paste.rs#L103-L114

I would start with test(s): as you have a way to reproduce this and it's a panic.
Not sure I can help more, as there's no specific about the panic per se — removing +1 seems logical.

seanstrom:
@SomeoneToIgnore thanks for the feedback!

I've pushed up another change that handles the editor commands for Cut, Copy, and Paste. I've also revised the tests based on your suggestions, and added a new test for the Cut command.

I've started looking into the Vim commands too, since it seems they also add an extra newline between multi-cursor selections of entire lines (using visual mode with line selection).

Should I continue to make more changes in this PR for the Vim commands?
And is there anywhere else I should check that could be affected by these editor changes?

SomeoneToIgnore:
Thank you for the fix, this looks good as the first glance, but I hope for @ConradIrwin to check the Vim part first to see what should we do about it.

ConradIrwin:
Thanks for the offer! If you'd like to make the same changes to vim mode, happy to accept a second PR, but no need to block on that. I also don't think it's super urgent as I haven't (yet) had any complaints about this – vim users are a little less likely to use multi-cursor actions like this)

PR #41206, opened 2025-10-26
acp_tools: Add button to clear messages

yordis:
Added a "Clear Messages" button to the ACP logs toolbar that removes all messages.

## Motivation

When debugging ACP protocol implementations, the message list can become cluttered with old messages. This feature allows clearing all messages with a single click to start fresh, making it easier to focus on new interactions without closing and reopening the ACP logs view.

Release Notes:

- N/A

### DIFF
--- a/crates/acp_tools/src/acp_tools.rs
+++ b/crates/acp_tools/src/acp_tools.rs
@@ -259,6 +259,15 @@ impl AcpTools {
         serde_json::to_string_pretty(&messages).ok()
     }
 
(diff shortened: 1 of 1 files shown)


yordis:
https://github.com/user-attachments/assets/b4238498-1a15-4c56-a805-c250ef93b6e9

bennetbo (approved):
Thank you! I don't think we need release notes here since it's a debug view, so I removed those.

PR #41222, opened 2025-10-26
gpui: Use AtomEnum::ATOM for clipboard target replies

cppcoffee:
Release Notes:

- N/A


probably-neb:
This seems almost not worth doing it's such a minor change, and if we are going to do it I think we should replace ALL instances of `self.atom.ATOM` with `AtomEnum::ATOM` and remove `ATOM` from the list of atoms we initialize. If you'd like to do that and test it out, feel free to ping me and we can re-open

cppcoffee:
Hi @probably-neb , I’ve finished changes, can can you review this PR: https://github.com/zed-industries/zed/pull/43537

PR #41225, opened 2025-10-26
agent: Clarify `include_pattern` parameter usage for grep tool

### LINKED ISSUE #41189: AI: grep tool call returns no matches with `include_pattern`
### Summary

Grep tool with `include_pattern` parameter returns no matches even though matches exist.

**UPDATE:** This seems to be a prompt engineering / tool definition issue, see my analysis in the comment below. Opened PR #41225 for a proposed fix

### Description

1. Cause model to emit a grep tool call with an `include_pattern` such as `"frontend/src/**/*.ts"` that should return results

**Expected Behavior**: Zed returns matches to the model
**Actual Behavior**: Zed erroneously reports "no matches" to the model

### Model Provider Details
- Provider: Multiple
- Model Name: Observed with Sonnet 4.5 via Anthropic API and also via OpenAI/litellm (backed by Bedrock)
- Mode: Agent Panel

------------------------------------

Some debug info:

Tool call (assembled from  [response.txt](https://github.com/user-attachments/files/23144084/response.txt), which is the SSE stream from the model)

```
{
  "regex": "formatPreciseTime",
  "include_pattern": "frontend/src/**/*.ts"
}
```

Attached screenshot from empty results

Image

Proof that results exist (run in the project root dir):

```bash
$ rg formatPreciseTime frontend/src/**/*.ts
frontend/src/utils/date.ts
1:export const formatPreciseTime = (dateString: string) => {
```

Here is the parameter description from the prompt:

```
    "include_pattern": {
        "description": "A glob pattern for the paths of files to include in the search.\nSupports standard glob patterns like \"**/*.rs\" or \"src/**/*.ts\".\nIf omitted, all files in the project will be searched.",
        "nullable": true,
        "type": "string"
    },
```

If model does not include the `include_pattern` parameter, this works as expected.

Zed: v0.209.6 (Zed) 
OS: macOS 15.7.1
Memory: 36 GiB
Architecture: aarch64

procr1337:
Adds clarifying examples and information about the `include_pattern` parameter for the grep tool, analogous to the `path` parameter of the read and list tools and others. 

The lack of clarity led to unexpected agent behavior described in #41189. This change is confirmed to improve the precision of Claude Sonnet 4.5 when searching for code in an empty conversation (without it, it leaves out the root directory in the query).

Closes #41189.

```
Release Notes:

- Clarify grep tool description to improve agent precision when using it with the `include_pattern` parameter
```

### DIFF
--- a/crates/agent/src/tools/grep_tool.rs
+++ b/crates/agent/src/tools/grep_tool.rs
@@ -32,8 +32,21 @@ pub struct GrepToolInput {
     /// Do NOT specify a path here! This will only be matched against the code **content**.
     pub regex: String,
     /// A glob pattern for the paths of files to include in the search.
-    /// Supports standard glob patterns like "**/*.rs" or "src/**/*.ts".
+    /// Supports standard glob patterns like "**/*.rs" or "frontend/src/**/*.ts".
     /// If omitted, all files in the project will be searched.
+    ///
+    /// The glob pattern is matched against the full path including the project root directory.
+    ///
+    /// <example>
+    /// If the project has the following root directories:
+    ///
+    /// - /a/b/backend
+    /// - /c/d/frontend
+    ///
+    /// Use "backend/**/*.rs" to search only Rust files in the backend root directory.
+    /// Use "frontend/src/**/*.ts" to search TypeScript files only in the frontend root directory (sub-directory "src").
+    /// Use "**/*.rs" to search Rust files across all root directories.
+    /// </example>
     pub include_pattern: Option<String>,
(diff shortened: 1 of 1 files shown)


rtfeldman (approved):
Thanks @procr1337!

procr1337:
@rtfeldman what happens next, do I need to update the release notes like the bot suggests?

procr1337:
@rtfeldman I rebased it to main to force checks to be actually run now and added Release notes section to the PR. I still encounter this bug daily.

PR #41005, opened 2025-10-23
Refactor timestamp formatting in Git UI components to use `chrono` for local time calculations

virajbhartiya:
- Updated `blame_ui.rs`, `branch_picker.rs`, `commit_tooltip.rs`, and `commit_view.rs` to replace the previous timestamp formatting with `chrono` for better accuracy in local time representation.
- Introduced `chrono::Local::now().offset().local_minus_utc()` to obtain the local offset for timestamp formatting.

Closes #40878

Release Notes:
- Improved timestamp handling in various Git UI components for enhanced user experience.

PR #41229, opened 2025-10-26
git_ui: Add timezone to absolute timestamp of commit in git blame popover

relaxcn:
…ver ui

The UI improvement for inline git blame popover.

When I view the commit time in the Git blame popover, I get confused because it shows the absolute time in `UTC+0`, while my local time zone is `UTC+8`, and there’s no any indication of the shown time's zone.

image

This is very counterintuitive, so I think it needs to display the committer's local time. 

image

### DIFF
--- a/crates/git/src/blame.rs
+++ b/crates/git/src/blame.rs
@@ -183,6 +183,21 @@ impl BlameEntry {
             Ok(time::OffsetDateTime::now_utc())
         }
     }
+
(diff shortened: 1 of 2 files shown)


relaxcn:
@cla-bot check

relaxcn:
@cla-bot check

maxdeviant:
This may be obviated by https://github.com/zed-industries/zed/pull/41005.

relaxcn:
> This may be obviated by #41005.

Yes. Thank you. I close this PR.

PR #41243, opened 2025-10-26
Update open buffers with newly registered completion trigger characters

### LINKED ISSUE #108: Typing dot should give completions
### Relevance Confirmation

- [ ] I confirmed that this is not an issue with the Eclipse JDT Language Server itself

### What happened?

When typing `.` after a variable or class name, nothing happens. Only when I type a letter do I get completions for available methods/public members starting with that letter. 

I should see completions for all available methods and public properties after typing `.`.

### Environment

Zed: 209.6
Platform: macOS
Java: OpenJDK 25

playdohface:
Closes https://github.com/zed-extensions/java/issues/108

Previously, when language servers dynamically register completion capabilities with trigger characters for completions (hello JDTLS), this would not get updated in buffers for that language server that were already open. This change is to find open buffers for the language server and update the trigger characters in each of them when the new capability is being registered. 

Release Notes:

- N/A


crates/project/src/lsp_store.rs
```diff
@@ -11898,12 +11898,46 @@ impl LspStore {
...
+                                .unwrap_or_default()
+                                .into_iter()
+                                .collect::<BTreeSet<_>>();
+                            for handle in buffers_with_language_server {
+                                let triggers = triggers.clone();
+                                let _ = handle.update(cx, move |buffer, cx| {
+                                    let combined_triggers = triggers
+                                        .union(buffer.completion_triggers())
+                                        .cloned()
+                                        .collect();
```
osiewicz: Should we be unionizing the completion trigger set? What if there are multiple language servers for a single buffer with disjoint triggers? 
Intuitively, I'd expect the registered completions to override whatever a language server has declared previously.

playdohface: I was wondering the same thing. In my use case, the existing trigger list was always empty, and I think realistically anything else is a rare edge case. But intuitively I'd want to add the newly registered triggers to existing ones.
For example, if there is some hypothetical inline-HTML language-server that combines with a regular language server and triggers on '<', we probably would not want to override this. If our primary language server gets triggered where it usually would not, it should handle that gracefully and respond with an empty list of completions, but if our inline-HTML-server does not get triggered we don't get completions, which might render it altogether useless.

osiewicz: The completion triggers are tracked per-language server though (and only then get unionized), so it should be ok to just grab existing completion triggers and unionize with that. Right now you're grabbing triggers of all language servers, which is a bit of an overkill imho.

playdohface: Ah you're right `set_completion_triggers` is already per language server💡 
I'll get rid of the union entirely then, this is supposed to set the triggers for this language server, so exactly what `set_completion_triggers` does.

osiewicz:
Thank you!

PR #41252, opened 2025-10-27
Implement `vimMotion::Shortcut` action

cletusigwe:
- Add vimMotion::Shortcut (Vim-agnostic motion playback)
      - New action type: vimMotion::Shortcut that replays a whitespace-separated keystroke sequence.
      - Registered at the workspace level so it’s available even when Vim is disabled.
      - If the focused Editor lacks a Vim addon, temporarily “loan” Vim around playback (activate → dispatch → deactivate) and restore the original modal state.
      - Dispatches via Workspace::send_keystrokes_impl to match existing sequencing and focus-change flushing.
      - Parsing uses Keystroke::parse(...).log_err() to skip invalid tokens without failing.
      - Always bind Vim keymap as part of default keymap loading, so MotionShortcut works regardless of Vim mode.

- Tests
      - Expanded parity test to assume Normal mode and compare shortcut vs raw keystrokes across multiple sequences.
      - Added broader “Vim disabled” test that runs the same sequences with Vim off and asserts:
          - Visual state matches Vim-enabled raw keystrokes.
          - VimAddon is not left attached after playback.
      - Test harness always binds Vim keymap (contexts keep it inert when Vim is disabled).
  
Motivation
https://x.com/adamwathan/status/1980348738280497601

Examples
```json
{
     "secondary-alt-y": ["vimMotion::Shortcut", "y i w"], // yank inner word
     "secondary-alt-u": ["vimMotion::Shortcut", "g U i w"], // uppercase inner word
}
```

Release Notes:
- Added support for vimMotion::Shortcut action keymap e.g `{"cmd-+": ["vimMotion::Shortcut": "v i q"]}`

### DIFF
--- a/crates/vim/src/test.rs
+++ b/crates/vim/src/test.rs
@@ -24,7 +24,10 @@ pub use vim_test_context::*;
 use indoc::indoc;
 use search::BufferSearchBar;
 
-use crate::{PushSneak, PushSneakBackward, insert::NormalBefore, motion, state::Mode};
+use crate::{
+    MotionShortcut, PushSneak, PushSneakBackward, VimAddon, insert::NormalBefore, motion,
+    state::Mode,
+};
 
 use util_macros::perf;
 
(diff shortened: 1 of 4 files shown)


cletusigwe:
@cla-bot check

cletusigwe:
@cla-bot check

ConradIrwin:
Nice! Thanks for this.

I think #41268  is a more promising direction, so closing in favor of that; feel free to test that out and provide further feedback.

cletusigwe:
Thanks!
I got a bit carried away with trying to make it as easy as possible to define such keymaps. The #41268 PR has made me realise that I did not need to even load the vim keymaps file globally but could have achieved the same effect by only just loading the vim addons globally which also eliminates the need to flash a vim context before running the keystrokes. With the benefit of hindsight, if I would do it over again, I would have done that instead.

PR #41256, opened 2025-10-27
vim: Add `:<range>w <filename>` command

AidanV:
Release Notes:

- Adds support for `:[range]w {file}`
  - This writes the lines in the range to the specified
- Adds support for `:[range]w`
  - This replaces the current file with the selected lines

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -2046,6 +2046,11 @@ impl Buffer {
         }
     }
 
+    /// Marks the buffer as having a conflict regardless of current buffer state.
+    pub fn set_conflict(&mut self) {
+        self.has_conflict = true;
+    }
+
     /// Checks if the buffer and its file have both changed since the buffer
     /// was last saved or reloaded.
     pub fn has_conflict(&self) -> bool {
--- a/crates/vim/src/command.rs
+++ b/crates/vim/src/command.rs
@@ -330,11 +330,12 @@ pub fn register(editor: &mut Editor, cx: &mut Context<Vim>) {
                 let Some(range) = range.buffer_range(vim, editor, window, cx).ok() else {
                     return;
                 };
-                let Some((line_ending, text)) = editor.buffer().update(cx, |multi, cx| {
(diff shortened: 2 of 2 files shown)


ConradIrwin:
@AidanV Nice, thanks!

When overwriting the current file in vim, the buffer retains it current contents. I think we should emulate this too (so the buffer is written, but marked as dirty).

AidanV:
When writing to the current file we now retain the buffer contents and mark the buffer as conflicted. I also added a check to just save the file when we are selecting the entire buffer in order to match Vim.

ConradIrwin:
I don't think it should be marked conflicted? Just unsaved. What does vim do?

AidanV:
Vim does not have the concept of conflicted as far as I am aware. Vim silently has different contents in the buffer and the file. 

I think that marking the buffer as conflicted provides a better UI since it informs the user that the file on disk changed without updating the buffer contents. 

Additionally, marking the buffer conflicted makes it more intuitive to get at the new file contents by clicking `Discard` when saving. If it is just unsaved, you either have to close and reopen the file or `:e`. Both of these options still work when using conflicted instead of unsaved but have worse discovery IMO.

Another reason I think this is better is because writing a range of lines to its own file can be useful for Vim power users but can be very confusing for newcomers. Showing that it is conflicted improves the transparency of what is going on in a way that Vim cannot do by default.

I should have included these reasons in the PR. I will do that next time.

I will also happily change it if you still think it should be unsaved. Let me know your thoughts.

ConradIrwin:
Ok, conflicted makes sense.

That said, on 8615b38c7b, if I open README.md and type `:3,5w! README.md`, then the loaded contents of the buffer becomes just lines 3-5 still?

AidanV:
In the README.md file `:3,5w! README.md` has the same behavior as just `:3,5w!`. In other words, the loaded content of the buffer is not updated but is marked as conflicted.

In Vim, `:3,5w! README.md` does the same thing as `:3,5w!`. Neither of which update the loaded buffer content but both change the file contents.

ConradIrwin:
Hmm, I'm sure when I tested this last night I was not seeing that; but trying again this morning it works as you say; sorry about that.

PR #41268, opened 2025-10-27
Add always-on modal actions mode for Vim action keybindings

0xGlace:
This PR enables using Vim operators and text objects via keybindings without entering modal editing mode.
Even when vim mode is **disabled**, all Vim actions (operators, text objects, and motions) are available for custom keybindings.
This is done by keeping Vim mode alwyas activate and the modal editing setting controls whether its keymap is loaded or not.
When modal editing is off, vim mode doesn't change cursor shape or clip line ends, and it keeps editor input enabled.

This is most useful when one wants to slowly transition into full modal editing, or just needs some of the functionality and doesn't want to keep switching between modes.

Other changes:
* Updated docs/vim to show how this feature can be used.
* New key context: `vim_mode_off`
* Workspace actions registered by the vim add-on now check whether vim mode for vim mode being enabled.

This feature has a few accompanying tests:
- `test_vim_disabled_actions_work_with_manual_binding`
Test vim actions work from a keybind with vim mode disabled.
- `test_vim_disabled_multi_keystroke_sequence_works`
Test multi keystroke sequences work when vim mode is disabled.
- `test_passive_mode_goto_last_modification`
Test `HelixGotoLastModification` working in passive mode to ensure Vim is getting editor updates.

_Original feature idea and motive from [Adam Wathan](https://x.com/adamwathan/status/1980348738280497601)'s post.
This implementation is basically the same thing [mentioned](https://x.com/conradirwin/status/1980381437095473563) by [at]ConradIrwin under the same post._

Release Notes:

- Added ability to use vim actions in keybinds without vim mode being enabled.


ConradIrwin:
Nice! Thanks for this.

Do we need passive mode, or can we just enable vim mode always without loading the keymap?

There are a few actions that we override (e.g. https://github.com/zed-industries/zed/blob/cdc2736ab5f847b790031369a27551ef37121e73/crates/vim/src/vim.rs#L403) that we should add a check for whether vim mode is enabled, but otherwise I hope it should be relatively safe.

Some vim actions require an additional argument that is currently required to be passed by a keystroke (like `dt<space>` to delete until the next space) and so I think we should leave that machinery in place in this mode.

Happy to pair with you on this if you want: https://cal.com/conradirwin/pairing

0xGlace:
I removed the passive mode and went after the keymap itself, with the last changes, Vim mode is always on and the setting only toggles its keymap and changes some behavior similar to the previous passive mode.

> There are a few actions that we override…
For those actions like you've said there's now a simple check, if Vim mode isn't enable, they dispatch the normal action.

> Some vim actions require an additional argument that is currently required to be passed by a keystroke (like dt<space> to delete until the next space) and so I think we should leave that machinery in place in this mode.

That's now handled:

https://github.com/0xGlace/zed/blob/858a78b726478bfb96defa2a3c296fdb53a5c640/crates/vim/src/vim.rs#L1307-L1314

I also changed the tests, one of them tests for this exact case.

I think this is much better than the previous approach, that one also had an issue where the mode indicator didn't update properly and I spent most of the time trying to figure out, not having experience with GPUI also made it more tricky but this removes the need for things like that.

> Happy to pair with you on this if you want: https://cal.com/conradirwin/pairing

Thank you! and I wish I could but I'm not in a place where I can at the moment

0xGlace:
I just realized another thing new approach allows for I think is having the conditional keybind context which passive mode didn't allow for

0xGlace:
I added a keybind context `vim_mode_off` and add it to the Vim mode docs, though the wording and the name of it might need changing if it doesn't fit.
Should I edit the original PR message to reflect the current implementation?

ConradIrwin:
@0xGlace Please do update the PR description, as that will be used as the commit message.

Also looks like it's not building right now?

0xGlace:
Done!
> Also looks like it's not building right now?

I missed a small change the last time I was trying to resolve conflicts using the GitHub UI, it's now fixed.

Have you check the changes I made to the workspace actions in vim.rs?
The last test I added was for exactly `dt<space>` so it should be all good now.

Also there was an example keymap I previously added to the vim docs but I wasn't sure why it was making the doc test fail, so I removed it especially since there are other and better examples already in the docs.
(oh well I forgot to put `action::Sequence`)

0xGlace:
This should be ready now!
There was one issue I didn't catch in my testing before and the test for it was passing, but somehow failing here. But it's now been fixed.

0xGlace:
@ConradIrwin I'm not sure why the `gpui::test_file_url_reads_as_url_string` test is failing only macOS, I tried it locally and it was ok like all other tests.
There was a conflict so I updated the branch and fixed an issue.
I'll leave it to you if you decide to update it again or think it's an issue with my PR.

ConradIrwin:
That test is new, and seems unrelated to your changes so probably nothing to worry about.

I'm going to play with this a bit and see if there any obvious negative consequences.

ConradIrwin:
Having played around with this more and thought about it, I think we should *not* merge this. Sorry for the disappointment.

Although it does work, the original request (to be able to do `viq`) is pretty blergh to actually do (that was one advantage of #41252, but that required enabling the vim keymap always which is also a no-go).
```
  {
    "bindings": {
      "cmd-g": [
        "action::Sequence",
        [
          "vim::SwitchToVisualMode",
          ["vim::PushObject", { "around": false }],
          "vim::AnyQuotes"
        ]
      ]
    }
  },
```
It also (though you can't tell) puts vim into Visual mode in the background. Which might be what you wanted, but probably isn't if you have any other vim-related bindings going on. (And you can't leave visual mode without losing your selection so...)

I think the right fix here is to give people who are not using vim mode the actions they need (and in the future, we should build extension support for providing these!).

For the original request I think the following (even more hideous) binding mostly works:

```
  {
    "bindings": {
      "cmd-g": [
        "action::Sequence",
        [
          "editor::MoveToEnclosingBracket",
          "editor::SelectLargerSyntaxNode",
          "editor::SelectRight",
          "editor::SwapSelectionEnds",
          "editor::SelectLeft"
        ]
      ]
    }
  },
```

It doesn't seem unreasonable to add an editor::SelectSurroundingQuotedContent as a primitive though.

Thanks again for working on this, and sorry I didn't realize this sooner.

PR #41281, opened 2025-10-27
editor: Optimize selection overlap checking

cppcoffee:
Replace the binary search approach with a more efficient partition_point method for checking selection overlaps. This eliminates the need to collect and sort selection ranges separately, reducing memory allocation and improving performance when handling multiple selections.

Release Notes:

- N/A


Veykril (approved):
Thanks!

Issue #29026, opened 2025-04-18
Clipboard copy/paste only works within Zed's code editor (no system ↔ Zed file transfer)


LoveBloodAndDiamonds:
### Summary

Currently, the clipboard copy/paste functionality in Zed has the following limitations:

System to Zed:

Cannot paste files copied from the system (Finder/Explorer/etc.) into Zed via clipboard (Ctrl/Cmd+V)

Workaround: File drag-and-drop works

Zed to System:

Cannot copy files from Zed and paste them into system folders via clipboard

Expected Behavior:

Support cross-application file transfer via clipboard (both directions):

Copy file in system → Paste in Zed (as alternative to drag-and-drop)

Copy file in Zed → Paste in system file manager

Reproduction Steps:

Copy a file in system file manager (Ctrl+C / Cmd+C)

Try to paste in Zed (Ctrl+V / Cmd+V) → Fails

Copy a file in Zed's file tree

Try to paste in system file manager → Fails

Environment:

Zed version: [your version]

OS: MacOS, 15.4 Beta (24E5222f)

Additional Context:
This would improve workflow efficiency, especially when working with frequent file transfers between Zed and other applications.

### Zed Version and System Specs

Zed: v0.182.11 (Zed) 
OS: macOS 15.4.0
Memory: 16 GiB
Architecture: aarch64

huaium:
Also, cross-project copy & paste are not allowed in Zed. Hope this feature can be developed.

macwilam:
I have seen this problem as well. I cannot tel for sure when but within last two weeks.
Linux
kde
xorg
Debian stable (new xanmod kernel LTS)
0.185.15 version

I checked the logs and nothing apperas in the log when I try to paste data.

macwilam:
I Have an update on that. It turns out that I cannot copy from some windows but I can copy from others.
From example:
Chrome -> Zed  Works
Firefox -> Zed Works
Thunderbird -> Zed does not work

Kate -> Zed does not work
Libreoffice -> Zed does not work
KWrite -> Zed does not work

If anyone is looking at it or my testing will help let me know.

AntonRybakou:
Copying files or directories from one project to another (both of which are opened with Zed) also doesn't work.

OS: macOS Sequoia 15.4.1 arm64
Host: MacBook Air (M1, 2020) 8GB
Kernel: Darwin 24.4.0

macwilam:
I think my issue got resolved somehow.

[...]

PR #41282, opened 2025-10-27
gpui: Fix copying files across apps on x11

eduardoheleno:
Fix copy file issue on x11 mentioned in this issue https://github.com/zed-industries/zed/issues/29026.

Release Notes:
- Fixed file copying across different apps on x11 window system.

### DIFF
--- a/crates/gpui/src/platform/linux/x11/client.rs
+++ b/crates/gpui/src/platform/linux/x11/client.rs
@@ -1570,9 +1570,7 @@ impl LinuxClient for X11Client {
         let mut state = self.0.borrow_mut();
         let abs_item_paths = item
             .iter()
-            .map(|item| {
-                format!("file://{}", item.text().unwrap_or_default())
-            })
+            .map(|item| format!("file://{}", item.text().unwrap_or_default()))
             .collect::<Vec<String>>()
             .join("\n");
         state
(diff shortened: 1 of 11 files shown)


eduardoheleno:
@cla-bot check

eduardoheleno:
Any updates on this?

crates/gpui/src/platform/linux/x11/client.rs
```diff
@@ -1566,6 +1566,25 @@ impl LinuxClient for X11Client {
         state.clipboard_item.replace(item);
     }
 
+    fn write_file_to_clipboard(&self, item: Vec<crate::ClipboardItem>) {
```
probably-neb, line 1569: I'd rather see this PR use the `ExternalPaths` variant of `ClipboardEntry` introduced in #42858, rather than storing the file paths as `ClipboardEntry::String`'s and then expecting correct use of this API

probably-neb (changes requested):
Thanks for taking a shot at this. Unfortunately, I'm going to close this as I believe a different approach should be taken here.

In addition to my comment below that this should not be an additional clipboard API on the `Platform`, I also think it should include support for pasting files.

Because this PR requires basically a full rework and additional work I'm going to close it, however, if you're interested in continuing to work on it, feel free to make the requested changes and @-mention me and I'd be happy to re-review and re-open.

Cheers!

eduardoheleno:
Thank you @probably-neb for the feedback! I really appreciate it.

To get more context cause i still want to work on this, what do you think that could be a better approach? I thought that implement on `Platform` should make sense cause i guess that this feature is platform specific it isn't?

Btw this code includes file pasting specifically.

PR #41291, opened 2025-10-27
remote: Support building x86_64-linux-musl proxy in nix-darwin

kubkon:
Thanks to this change I can ssh into a linux box from macOS running nix-darwin :tada: This change adds two things to our remote server build process:
1. If `ZED_BUILD_REMOTE_SERVER` contains `noinstall`, the process will skip adding a rustup target and install `cargo-zigbuild` - we assume you would have already done that.
2. If `ZED_BUILD_REMOTE_SERVER` contains `musl` and `ZED_ZSTD_MUSL_LIB` is set, we will pass its value (the path) to `cargo-zigbuild` as `-C link-arg=-L{path}`.

Release Notes:

- N/A

### DIFF
--- a/crates/remote/src/transport.rs
+++ b/crates/remote/src/transport.rs
@@ -176,6 +176,10 @@ async fn build_remote_server_from_source(
     };
     if platform.os == "linux" && use_musl {
         rust_flags.push_str(" -C target-feature=+crt-static");
+
+        if let Ok(path) = std::env::var("ZED_ZSTD_MUSL_LIB") {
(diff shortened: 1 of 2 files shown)


P1n3appl3:
Adding musl/cargo-zigbuild to the shell looks good, but in terms of how we prevent our code from calling `rustup`/`cargo-install` I'd much rather just check programmatically if they're installed (similar to what we do for other cargo tools like `cargo-machete` rather than adding another flag to that env var to control it.

PR #41320, opened 2025-10-27
git_ui: Dismiss pickers only on active window

v1narth:
Small QOL improvement for branch picker to only dismiss when focus lost in active window.
This can benefit those who need to switch windows mid branch creation to fetch correct jira ticket number or etc.

Added `window.is_active_window()` guard in `picker.rs`  -> `cancel` event

Release Notes:
- (Let's Git Together) Fixed a behavior where pickers would automatically close upon the window becoming inactive.

### DIFF
--- a/crates/git_ui/src/branch_picker.rs
+++ b/crates/git_ui/src/branch_picker.rs
@@ -152,12 +152,9 @@ impl BranchList {
         let delegate = BranchListDelegate::new(repository, style);
         let picker = cx.new(|cx| Picker::uniform_list(delegate, window, cx));
 
-        let _subscription =
-            cx.subscribe_in(&picker, window, |_, _, _: &DismissEvent, window, cx| {
-                if window.is_window_active() {
-                    cx.emit(DismissEvent);
-                }
-            });
+        let _subscription = cx.subscribe(&picker, |_, _, _, cx| {
+            cx.emit(DismissEvent);
+        });
 
         Self {
             picker,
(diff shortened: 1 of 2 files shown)


v1narth:
@cla-bot check

v1narth:
I think it may be better to also introduce this to other places like `file_finder`, `command_palette`, maybe even `profile_selector` and `agent_model_selector` so it consistent across the UI.

cole-miller:
Thanks!

I think we can enable this for all pickers in a clean way by adding a `&& window.is_active_window()` guard before calling `cancel` to the blur handlers in `picker.rs`, would you be up for doing that?

v1narth:
@cole-miller thx for the guide, i've implemented your suggestion, seems like its working just fine

cole-miller:
Sorry, I think my previous post was unclear--I think we should add this check inside the two blur handlers in picker.rs, so these lines:

https://github.com/zed-industries/zed/blob/f1c2afdee08a5d6942a54ad95045e77c055264e7/crates/picker/src/picker.rs#L610

https://github.com/zed-industries/zed/blob/f1c2afdee08a5d6942a54ad95045e77c055264e7/crates/picker/src/picker.rs#L622

Maybe a bit of a nitpick but I think we should ensure that this behavior is only applied for blue handlers.

v1narth:
@cole-miller done

cole-miller:
Thank you!

PR #23412, opened 2025-01-21
Tab-out feature

markosnarinian:
Closes #22349 Closes #14803

Release Notes:

- Added the tab-out feature.
- When tab is pressed with the cursor before a closing delimiter, colon or semicolon (`), }, ], ", ', ;, :`) the cursor moves one column to the right ("tabs" out of the parentheses, brackets, etc.).

PR #41321, opened 2025-10-27
Tree-sitter tab navigation (Smart tab)

### LINKED ISSUE #14803: Smart tab support
### Check for existing issues

- [X] Completed

### Describe the feature

Something I believe is very useful in Helix (and I remember experience the same with some Java IDEs in the past) is the feature they call "smart tab"

Here is the pull request with the implementation (and a demo of the feature): https://github.com/helix-editor/helix/pull/4443

_No response_

### LINKED ISSUE #22349: Implement something like the VSCode TabOut extension
### Check for existing issues

- [X] Completed

### Describe the feature

https://marketplace.visualstudio.com/items?itemName=albert.TabOut&ssr=false#review-details

Enormous productivity booster.

Zed: v0.166.1 (Zed)
OS: Linux Wayland pop 24.04
Memory: 31.1 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics (CML GT2) || Intel open-source Mesa driver || Mesa 24.0.9-0ubuntu0.3

_No response_

vamsi10010:
Closes #22349 
Closes #14803

- This PR introduces a syntax-aware navigation feature (using tree-sitter nodes) that enables fast cursor navigation within code structures. This PR adds two new components:
    - Adds two actions `MoveToEndOfLargerSyntaxNode` and `MoveToStartOfLargerSyntaxNode`. These actions can be bound to any keybinding if the user does not like the default smart tab settings.
    - Binds tab and shift-tab to perform the two actions respectively. In particular, tab will check if there is whitespace to the left of the cursor. If there is only whitespace, a regular tab is inserted, otherwise, the cursor will move to the end of its parent node. shift-tab will always move the cursor to the start of its parent node. This can be activated using a new setting called Smart Tab.
- If smart tab is activated, the navigation will respect snippet tab stops, completions, and edit predictions. Tab and shift-tab also act normally when text is selected. Smart tab also does not activate in normal mode for vim (it does activate in insert mode). Additional settings are available to supersede accepting completions and edit predictions in the smart tab settings page.
- This PR was heavily inspired by its namesake feature in Helix (https://github.com/helix-editor/helix/pull/4443). It also provides the behavior users might expect when using albert.TabOut in VSCode (and more). There was also an attempt to do this in #23412 which was rejected in favor of a more sophisticated approach. I took the feedback there and opted to use the builtin tree-sitter features for navigation.

Questions:
1. I currently hooked the smart tab functionality into the existing `editor::Tab` and `editor::Backtab` actions. Is this approach preferred to making separate actions for binding to tab and shift-tab keys?
2. Helix provides a setting where shift-tab can be used as a backup tab key. I believe they did so because not all terminals support shift-tab. However, since zed does not have that issue, I have mapped shift-tab to always move to the start of the parent node (if the setting is enabled). Do we want a setting where shift-tab can be configured as a tab backup?
3. I'm not sure what the best place would be to document this feature in zed docs.
4. Should this feature be enabled by default? I have it set to disabled.
5. While I have settings to supersede completions and edit predictions, I would love some feedback on their design so that I don't introduce any code debt there.

Here is smart tab in action:

https://github.com/user-attachments/assets/a964c218-2326-459e-8f8c-22304bcdfcb6

This is my first contribution to zed, so I would really appreciate any feedback on this! I believe this feature could be very helpful for users coming from other IDEs like VSCode or JetBrains.

Release Notes:
- Added new actions to move to start/end of tree-sitter nodes
- Added settings to bind tab and shift-tab to those actions


ConradIrwin:
Thanks for this!

I think the actions of move to start/end of tree-sitter node make sense, but I don't see a world where we can bind them to `tab` by default (that is already overloading for indent, auto-complete, and AI suggestions).

Is it possible to reduce this to just creating the actions, or would we need to add more information to the key-context to determine what to do when you hit tab?

vamsi10010:
> Is it possible to reduce this to just creating the actions, or would we need to add more information to the key-context to determine what to do when you hit tab?

Thanks for the feedback!

Its definitely possible to reduce this just to those two actions. I would have to add new key contexts that determine whether forward or backward movement is active. They would essentially do the same checks I hardcoded into `tab` and `backtab` and set two new key contexts (for ex. `smart_tab_forward_active` and `smart_tab_backward_active`). I will revisit my implementation soon and make these changes.

ConradIrwin:
Hmm; I'm not sure what the best option is.

What happens on this branch when you hit tab and have an autocompleter open, an AI suggestion visible, or a snippet with tab-stops?

I am nervous about tab just because it does so much already, but maybe this works if it's right at the end of the fallback list.

Either way a PR to just add the actions seems totally reasonable.

vamsi10010:
> What happens on this branch when you hit tab and have an autocompleter open, an AI suggestion visible, or a snippet with tab-stops?

Hitting tab will accept the autocompletion/edit prediction, or move to the next tab stop as expected. 

I also created settings that let you supersede accepting autocompletions/edit predictions, but I am not sure if we want to keep them. 

However, by default, tab key as movement is at the end of the fallback list and happens only before indentation.

ConradIrwin:
Closing this for now as it conflicts with everything. Let's start with a PR that does the actions you want, and then figure out separately how to get the tab behavior perfect.

PR #41329, opened 2025-10-27
Fix tab switcher spacing bug

Anthony-Eid:
The tab switcher render matches calls each workspace item's `Item::tab_content` function that can return an element of variable size. Because the tab switcher was using a uniform list under the hood, this would cause spacing issues when tab_contents elements had different sizes.

The fix is by changing the picker to use a material list under the hood.

Release Notes:

- N/A


teohhanhui:
I've been observing a tab sizing bug since upgrading to Zed 0.209.6

If I have a pinned tab (1st tab), then the tab immediately to its right (2nd tab) would be stuck in a resize loop / endless flickering. Seems to only happen when the 2nd tab is less than the max tab width or something like that.

Will this fix that?

SomeoneToIgnore:
That's totally different, hence a new issue with a video and reproduction steps would be great to have.

PR #33490, opened 2025-06-26
Add collapse functionality to outline entries

NRodriguezcuellar:
partly Closes #23075 

Release Notes:

- Now provides collapse and enables functionality to outline entries
  - Add a new expand_outlines_with_depth setting to customize how deep the tree is expanded by when a file is opened

part 2 is in #34164 

**Visual examples**
image
image
image

PR #41342, opened 2025-10-28
outline_panel: Fix collapse/expand all entries

### LINKED ISSUE #39937: Outline panel: Collapse and Expand all Entries do not do anything
### Summary

Outline panel: Collapse and Expand all Entries do not do anything

### Description

When in the outline panel the action to expand or to collapse all entries do no do anything. 
This is true both for being bound to a key, or from the command palette.
Expanding/collapsing single entry works as advertised.

Zed: v0.207.3 (Zed) 
OS: macOS 15.6.1
Memory: 24 GiB
Architecture: aarch64

zeld-a:
Closes #39937

Release Notes:

- Fixed expand/collapse all entries not working in singleton buffer mode


zeld-a:
The issue requested a test, but I am not familiar with the gpui tests so I did not include one. If someone would guide me, I could definitely provide that.

crates/outline_panel/src/outline_panel.rs
```diff
@@ -1635,56 +1635,57 @@ impl OutlinePanel {
...
+                        outline.excerpt_id,
+                        outline.outline.range.clone(),
+                    ))
+                }
+                PanelEntry::Search(_) => None,
+            })
+            .collect::<Vec<_>>();
+
+        if !entries_to_remove.is_empty() {
+            let to_remove: HashSet<_> = entries_to_remove.into_iter().collect();
```
SomeoneToIgnore: Nice idea to use `collect` instead of `fold`, but what's the reason to recollect this instead of collecting directly into `HashSet`?

SomeoneToIgnore (commented):
Thank you.

For the tests, you can check out the commit I've mentioned in the comment: https://github.com/zed-industries/zed/pull/33490

That one had added a test, we can extend the existing one with more actions calls and asserts, or copy the initialization into a new test and assert things there.

crates/outline_panel/src/outline_panel.rs
SomeoneToIgnore: Sorry, just realized: `self.collapsed_entries` is already a hash set.
So we can just remove the entries directly for `~O(1)` without collecting and allocating more memory.

zeld-a: I didn't really have a technical reason for it, it was merely what I came up with in the moment and it didn't cross my mind to alter it. Looking at it now, it doesn't make much sense lol. I have altered it to collect directly into the `HashSet`.

Do you think it would be better to also use a `HashSet` instead of a `Vec` in `collapse_all_entries()`? The only case I can think of that a `HashSet` would be better there is if we needed to remove duplicates, but I don't see a scenario in which we would have duplicates.

SomeoneToIgnore: Thank you for pointing out, wow, we do 
```
// snip
    .collect::<Vec<_>>();
    self.collapsed_entries.extend(new_entries);
```
there, while in reality can do without collecting entirely: just put that iterator directly into the `extend(`.

This way, we won't allocate `Vec` at all, and reuse the memory allocated already which is always good.

Here, we can do the same, as I've realized just a couple of minutes before your comment: https://github.com/zed-industries/zed/pull/41342#discussion_r2470055922

zeld-a: I would test this if I was at my computer but am out at lunch, but wouldn't calling `self.collapsed_entries.extend(self.cached_entries.iter().map(...))` hit the borrow checker since we are taking an immutable borrow and `extend()` needs a mutable borrow?

zeld-a: I was thinking a bit conservatively there, the compiler should be able to prove they do not overlap.

SomeoneToIgnore: Ah, that's why, makes sense.
In that regard, I think it does not matter much what we collect those things into: `HashSet` seems ok, `Vec` is not that bad either — as in the end our goal is to remove from/append to the other `HashSet`.

zeld-a: My worst nightmare has come to life,,, actually using DSA.

zeld-a: Alright! both are implemented without collecting. The only thing now is making the test.
I assume the test you were refering to earlier was this?
https://github.com/zed-industries/zed/blob/d9d24582bb5f09ff4b8221bc36dc4b5bb6c8a4b7/crates/outline_panel/src/outline_panel.rs#L6978-L7267

SomeoneToIgnore: Yeah, that test is one of the added in the original PR ones, which sets up an outline panel that shows a singleton buffer with outlines that have a hierarchy.

We need to either copy or extend that set-up to test the actions.

zeld-a: So i did my best to create a test but am getting a fail but im really not sure how this should be formatted. if you have any insight, id appreciate it. Here is the failure:
```bash
❯ cargo test -p outline_panel test_outline_expand_collapse_all
   Compiling outline_panel v0.1.0 (/home/chrisg/Development/zed/crates/outline_panel)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 4.01s
     Running unittests src/outline_panel.rs (target/debug/deps/outline_panel-7ca9f9ec4db773da)

running 1 test
test tests::test_outline_expand_collapse_all ... FAILED

failures:

---- tests::test_outline_expand_collapse_all stdout ----

thread 'tests::test_outline_expand_collapse_all' panicked at crates/outline_panel/src/outline_panel.rs:7729:13:
assertion failed: `(left == right)`

Diff < left / right > :
 outline: mod outer  <==== selected
<  outline: pub struct OuterStruct
<    outline: field: String
<  outline: impl OuterStruct
<    outline: pub fn new()
<    outline: pub fn method(&self)
<  outline: mod inner
<    outline: pub fn inner_function()
<    outline: pub struct InnerStruct
<      outline: value: i32
 outline: fn main()


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    tests::test_outline_expand_collapse_all

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 8 filtered out; finished in 0.24s

error: test failed, to rerun pass `-p outline_panel --lib`
```

zeld-a: i copied as much as made sense in the original test.

zeld-a: So with my latest commit [9aef7d8](https://github.com/zed-industries/zed/pull/41342/commits/9aef7d8fbce2d0113c57dd9f1d704366222ff79e), I have collapse working, but I would like some clarification on how the expand implementation should actually work.

When i expand from a fully collapsed outline, is it expected that it will expand even the nested items? My current implementation only expands the ones visible and not the nested ones. After invoking `CollapseAllEntries` a second time, it opens the ones that were previously nested.

zeld-a: Here is an example of the current implementation.

https://github.com/user-attachments/assets/c41f04f6-e907-4334-b164-f0c0bcd855f3

SomeoneToIgnore: Wow, thanks, that starts to look quite good.

What's confusing me is that we run an action that is supposed to *expand ALL entries* and it does not expand all entries — and one has to call it 3 times do achieve the result???

I would expect it to be have the way the multi buffer collapse/expand all works: collapse all and expand all, at once, on one action run.

SomeoneToIgnore: What's odd is that the test does not showcase this behavior from the video at all?
What's the point of that test then???

zeld-a: I agree that it should work the same in both. In my head, it makes sense that it should work the way that I implemented. But I should have checked and copied the multi-buffer implementation. I will fix this when I have a chance today.

As for the test, It currently passes assuming that it should only expand the visible entries. I will correct this test today as well.

zeld-a: Newest commit expands the nested entries as well and the test assumes it should.

zeld-a: Here is a video of it working as expected on my end.
https://github.com/user-attachments/assets/65c1786a-3649-4968-aa31-a4c8113ad8e2

SomeoneToIgnore: Thank you, this almost works: the singleton buffer expansion/collapsing logic is exactly what I've expected.

Yet, we broke the multi buffer case now, it seems (sadly, there is no test for that? Feel free to add one if you want to).
Now, when you collapse all and expand all, no files are expanded, hence no excerpts are expanded too:


https://github.com/user-attachments/assets/59b04f95-8fb5-4af4-aca3-cfaab671a761

zeld-a: So i went from multi works single doesn't, to single works multi doesn't. Lol. I'll give it a crack when I'm free today.

zeld-a: I corrected the mistake, seems I refactored a bit too much when I had the issue with nested entries not expanding. I did not make a test for this mostly because making the last test was only made possible due to someone else having done the hard part and all I had to do was adapt that.

Probably a good idea for me to study some other tests so I can get better at those.
Also thanks for bearing with me.

SomeoneToIgnore: > I did not make a test for this mostly because making the last test was only made possible due to someone else having done the hard part and all I had to do was adapt that.

Isn't that good?
All you have to do is concentrate on the meaning of your change, testing that.

Let's add a few steps after these lines?

https://github.com/zed-industries/zed/blob/344f63c6cadef985cea86b9ea16d9440ace54a6e/crates/outline_panel/src/outline_panel.rs#L6568-L6600

Repeating the same "call action, assert entries" pattern seems simple to add?

This way, we can run "collapse all, assert", then "expand all, assert" for "free", without writing anything to set up the test.

zeld-a: Sure thing, i will give it a crack. It will have to wait until this evening as today is a heavy day for me at college.

zeld-a: @SomeoneToIgnore I had a quick 30 minutes to sit down and take a look before class. Now I feel kinda silly for how simple that was to actually add to the test. Thanks for pushing me to try it.

SomeoneToIgnore (approved):
Nice, thank you a lot for pushing this to the final and adding more tests!

Issue #41249, opened 2025-10-27
Windows：zed crashed when add python virtual environment


dDostalker:
### Summary

zed crashed when I add python virtual environment ,. I entered "C:", and it would crash when I deleted it.

### Description

Steps to trigger the problem:
1. Open Zed
2. Click select Virtual Environments
3. Click Add Virtual Environments
4. Inputs "C:" (must "C:")
5. delete words "C:"

Actual Behavior:
crashed
Expected Behavior:
delete "C:"

Video
https://github.com/user-attachments/assets/e8b2f90e-4c9c-446a-9daa-ed3378457d87

### Zed Version and System Specs

Zed: v0.209.6 (Zed) 
OS: Windows 10.0.22621
Memory: 27.7 GiB
Architecture: x86_64
GPU: AMD Radeon(TM) Graphics || AMD Corporation || 24.10.1 (24.20.19.01-241011a-409016C-AMD-Software-Adrenalin-Edition)

### If applicable, attach your `Zed.log` file to this issue.

<details><summary>Zed.log</summary>

```log
2025-10-27T09:39:02+08:00 INFO  [project::lsp_store] language server 1: unregistered workspace/DidChangeWatchedFiles capability with id b97b29d6-5f62-4a61-9773-890730e9502a
2025-10-27T09:39:02+08:00 INFO  [project::lsp_store] language server 1: unregistered workspace/DidChangeWatchedFiles capability with id 44fcf4fd-dfa3-49ed-ac66-96896f447659
2025-10-27T09:40:41+08:00 INFO  [gpui::platform::windows::directx_renderer] Updating monochrome_sprite_pipeline buffer size from 512 to 2048
2025-10-27T09:40:41+08:00 INFO  [project::lsp_store] language server 1: unregistered workspace/DidChangeWatchedFiles capability with id 36e6b6ad-3655-4494-a672-383492160b9c
2025-10-27T09:40:41+08:00 INFO  [project::lsp_store] language server 1: unregistered workspace/DidChangeWatchedFiles capability with id 7d56ef6f-6d89-451d-a1d6-5d696aea8522
2025-10-27T09:41:56+08:00 WARN  [pet_windows_registry::environments] Failed to open HKLM\Software\Python\PythonCore\3.9\InstallPath, Os { code: 2, kind: NotFound, message: "系统找不到指定的文件。" }
2025-10-27T09:42:01+08:00 ERROR [crashes] triggering a crash to generate a minidump...
```
[...]

PR #41356, opened 2025-10-28
Fix crash when adding Python virtual environment label

dDostalker:
Closes #ISSUE
#41249 

Release Notes:

The error occurred due to the mismatched characters exceeding the boundaries, and the assertion failed. The fix ignores the non-compliant highlighted positions instead of causing a crash.
Now it can operate normally.


dDostalker:
@cla-bot check

Veykril:
This assert is intentional, we need to fix the faulty caller instead

PR #41364, opened 2025-10-28
gpui: Remove an unwrap that sometimes causes a crash on DirectX

yara-blue:
Seen a few crashes on Sentry where the unwrap call fails. No idea why the render_target_view is None. Though since we are dropping anyway we do not need to call unwrap so I removed that.

Release Notes:

- N/A


yara-blue:
closing, deeper underlying issue that needs to be addressed (tracked in first responder triage)

PR #41368, opened 2025-10-28
WIP: Notarize Mac builds in parallel

osiewicz:
We wanna notarize in parallel, as that step takes quite a bit to run.
As a bonus, we're also splitting up the notarization step so that it's separate from build. That way, if there's a notarization failure (which, well, happens), we won't have to re-do the build. 

Release Notes:

- N/A


osiewicz:
Closing in favor of https://github.com/zed-industries/zed/pull/41392

PR #41372, opened 2025-10-28
agent_ui: Don't show root project in path prefix in @-mention menu

danilo-leal:
This PR hides the worktree root name from the path prefix displayed when you @-mention a file or directory in the agent panel. Given the tight UI real state we have, I believe that having the project name in there is redundant—the project you're in is already displayed in the title bar. Not only it was the very first word you'd see after the file's name, but it also made the path longer than it needs to. A bit of a design clean up here :)

(PS: We still show the project name if there are more than one in the same workspace.)

Release Notes:

- N/A


osiewicz:
Nit: what if you have more than one worktree?

danilo-leal:
@osiewicz 😬 Great call, should probably have a codition that still shows it if that's the case...

osiewicz (approved):
lgtm

PR #41380, opened 2025-10-28
Clock in Status and Title Bar

FlorianDevP:
I have been using a simple extension in VSCode to show the current time, so I don't forget the time while coding. A good sleep schedule is quite important, but wanting to fix that one bug or Todo is not helping. 
I would have implemented this as an extension, but I don't know if changing the UI from extensions is planned or not.

The clock can be toggled, switched between 24 and 12 Hours and displayed in the Status or Title bar.
UTC offset is also configurable, but local offset will be used by default. If the format of this offset is wrong, an error notification will be shown.
These clock settings are also integrated into the settings editor, under Window & Layout. I tried to implement custom UI for the offset, but it was very ugly, so I used String for now.

I haven't yet implemented actions to toggle the clock, it's position or the 12 Hours display.
Maybe I could also add a setting to adjust the size of the clock.

Release Notes:

- Add clock to status or title bar

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16170,6 +16170,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
+name = "status_clock"
+version = "0.1.0"
+dependencies = [
+ "gpui",
+ "indoc",
+ "language",
+ "menu",
+ "project",
+ "rope",
+ "settings",
+ "theme",
+ "time",
+ "ui",
+ "util",
+ "workspace",
+ "zed_actions",
+]
+
+[[package]]
 name = "stop-words"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


FlorianDevP:
@cla-bot check

SomeoneToIgnore:
Thank you for the submission, had polled the team internally and found no support for this.

Zed, design and extra UI elements is a convoluted topic, yet the main rule of thumb is to keep things minimal.
I had to spend quite an effort to protect this little LSP button from being removed

image

as having quick access to language server statuses and actions is a vital feature to many people.

What's the clock feature's usefulness in comparison?  (added to both status and title bar??)
Not even close, I'm afraid — there's OS, cell phone, physical clocks and other devices with clock.

Hiding it by default and maintaining the code no one in the team wants to use seems too much of a stretch too, hence we'll wait for

> but I don't know if changing the UI from extensions is planned or not.

to happen first: that is not planned before Zed 1.0 , but at some point it should mature enough to provide simple visual customizations.

Last, somewhat personal, reason to be less enthusiastic about the PR: there are phrases like `A good sleep schedule is quite important, but wanting to fix that one bug or Todo is not helping.` and 0 screenshots at the same time, despite https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#sending-changes

PR #41394, opened 2025-10-28
tasks: Allow users to associated tasks with specific languages

Anthony-Eid:
This lets users filter tasks shown based on the active location they're currently in

Release Notes:

- N/A


osiewicz (changes requested):
I think it would be more intuitive if languages could have default tags instead to achieve the same goal.

Anthony-Eid:
That's a good idea @osiewicz, I'm going to pivot to that direction

PR #36548, opened 2025-08-20
gpui: Allow the application to run after all windows are closed

Guibi1:
This PR adds the option to prevent the event loop from exiting when the last window is closed by the user.

This is needs more thought because:
1. The naming might not be the best
2. This wasnt tested on all platforms
3. MacOS is special and we need to make a decision on the api about the fact that it has the opposite behaviours as the other os

Release Notes:

- Added the option to keep the application running with no window open

PR #41436, opened 2025-10-29
gpui: Move 'app closing on last window closed' behavior to app-side

Tryanks:
This commit is a continuation of #36548. As per [mikayla-maki's Comment](https://github.com/zed-industries/zed/pull/36548#issuecomment-3412140698), I removed the process management behavior located in GPUI and reimplemented it in Zed.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/platform/linux/wayland/client.rs
+++ b/crates/gpui/src/platform/linux/wayland/client.rs
@@ -387,9 +387,6 @@ impl WaylandClientStatePtr {
         {
             state.keyboard_focused_window = Some(window);
         }
-        if state.windows.is_empty() {
-            state.common.signal.stop();
-        }
     }
 }
 
--- a/crates/gpui/src/platform/linux/x11/client.rs
+++ b/crates/gpui/src/platform/linux/x11/client.rs
@@ -247,9 +247,6 @@ impl X11ClientStatePtr {
         }
         state.cursor_styles.remove(&x_window);
 
-        if state.windows.is_empty() {
-            state.common.signal.stop();
-        }
     }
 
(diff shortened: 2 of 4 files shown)


Tryanks:
@cla-bot check

mikayla-maki (approved):
PR looks good to me! Pretty simple implementation too :D

cc: @smitbarmase @reflectronic can you both test this on Windows and Linux and make sure it's working correctly?

reflectronic:
I like the approach in https://github.com/zed-industries/zed/pull/36548 better. I don't think this is a reasonable default behavior for Windows or Linux.

smitbarmase (approved):
Tested on Linux. Works as expected.

Tryanks:
> I like the approach in #36548 better. I don't think this is a reasonable default behavior for Windows or Linux.

I agree. Another idea was to add a GPUI-level subscription for when the last window closes. If the signal has no subscribers, the platform's default behavior applies. If it has any subscribers, the default quit is prevented, letting the subscriber decide.

On macOS, Zed would subscribe to this and respect the user setting. Third-party apps could subscribe and do nothing to keep running with no windows open.

reflectronic (approved):
We agree that GPUI should have a default policy of exiting the app when all windows are closed on Windows and Linux, but we'd like to move this into the cross-platform layer instead of hard-coding it for each platform. This PR does some of that refactoring, so, to that end, we will accept it. In a follow-up, we will re-introduce that behavior, with an opt-out for apps like Zed that want explicit control.

mikayla-maki:
@Tryanks if you could do a quick cargo format, we can get this merge :)

PR #41459, opened 2025-10-29
Document Go hard tabs in default settings

### LINKED ISSUE #40876: Golang `hard_tabs` config in language crate should be moved to default settings
### Description

Golang is configured to use hard tabs in https://github.com/zed-industries/zed/blob/main/crates/languages/src/go/config.toml. It's the only language to have `hard_tabs` configured in its crate.

This is confusing because other languages have `hard_tabs` configured in the default settings.

**Expected Behavior**: The global default for `hard_tabs` is `false`, and there are no language-specific overrides in Golang's default settings, so Zed should not use hard tabs.
**Actual Behavior**: Zed uses hard tabs for Golang and it's difficult to figure out why.

Zed: v0.208.6 (Zed) 
OS: macOS 15.7.1
Memory: 16 GiB
Architecture: x86_64

injust:
Closes https://github.com/zed-industries/zed/issues/40876

This is already present in the code but missing from the default settings, which is confusing.

Release Notes:

- N/A

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1700,6 +1700,7 @@
       "preferred_line_length": 72
     },
     "Go": {
+      "hard_tabs": true,
       "code_actions_on_format": {
         "source.organizeImports": true
       },


probably-neb:
It's not quite documented in the code, or at least I couldn't find it. It's set in the `config.toml` for Go which is a builtin language and therefore in this Repo. Seems like a fine change though, as it should be the default anyway for Go

PR #11503, opened 2024-05-07
Indent guides

bennetbo:
Builds on top of existing work from #2249, but here's a showcase:

https://github.com/zed-industries/zed/assets/53836821/4b346965-6654-496c-b379-75425d9b493f

TODO:
- [x] handle line wrapping
- [x] implement handling in multibuffer (crashes currently)
- [x] add configuration option
- [x] new theme properties? What colors to use?
- [x] Possibly support indents with different colors or background colors
- [x] investigate edge cases (e.g. indent guides and folds continue on empty lines even if the next indent is different)
- [x] add more tests (also test `find_active_indent_index`)
- [x] docs (will do in a follow up PR)
- [x] benchmark performance impact

Release Notes:

- Added indent guides ([#5373](https://github.com/zed-industries/zed/issues/5373))

PR #41470, opened 2025-10-29
Fix Gruvbox accent colors

arkriny:
In #11503, the "accents" option was incorrectly at the top level. This moves it under the "style" key so it takes effect.

### Before/After
1761750444_screenshot
1761750448_screenshot

Release Notes:

- N/A


MrSubidubi (approved):
Nice find, thanks! And congratulations to your first contribution! 🎉

PR #41479, opened 2025-10-29
Reduce amount of foreground tasks spawned on multibuffer/editor updates

Veykril:
When doing a project wide search in zed on windows for `hang`, zed starts to freeze for a couple seconds ultimately starting to error with `Not enough quota is available to process this command.` when dispatching windows messages. The cause for this is that we simply overload the windows message pump due to the sheer amount of foreground tasks we spawn when we populate the project search.

This PR is an attempt at reducing this.

Release Notes:

- Reduced hangs and stutters in large project file searches


Veykril:
This does not fully fix the issue yet, but is already an incremental improvement so I will land this for now.

PR #41482, opened 2025-10-29
bundle: Restore local install on macOS

claytonrcarter:
I just pulled and ran a local build via `script/bundle-mac -l -i` but found that the resulting bundle wasn't installed as expected. (me: "ToggleAllDocks!! Wait! Where is it?!") Looking into, it looks like the `-l` flag was removed in #41392, leaving the `$local_only` var orphaned, which then left the `-i/$local_install` flag unreachable. I suspect that this was unintentional, so this PR re-adds the `-l/$local_only` flag to `script/bundle-mac`.

I ran the build again and confirmed that local install seemed to work as expected. (ie "ToggleAllDocks!! :tada:")

While here, I also removed the last reference to `$local_arch`, because all other references to that were removed in #41392.

/cc @osiewicz 

Release Notes:

- N/A


ConradIrwin:
Sorry about that, and thank you! We intended to remove the `local_only` flag (and just default the architecture to the host), so we should remove references to local_only and fix the `-i` flag

I can take a look later today if you don't get to it!

claytonrcarter:
OK, I've pushed a change to remove `$local_only` and update `$local_install` accordingly. Local build and install still seems to be working for me. That said, I don't really understand the bundling/notarization process so please scrutinize it closely and push whatever changes are needed to make it right.

Thank you, @ConradIrwin!

ConradIrwin:
thanks again!

ConradIrwin:
(the only changes I needed were to fix our new test workflow :D )

PR #41504, opened 2025-10-30
edit prediction: Add ability to switch providers from the status bar menu

### LINKED ISSUE #41500: AI:  Difficulty switching to a non-Zed edit prediction provider
### Summary

In an example of so-called "dark patterns" at work, Zed has made it difficult to switch from Zed predictions to another provider while switching to Zed can be done in a single click.

### Description
Steps to trigger the problem:
1. Turn on Zed edit predictions by clicking the icon in the bottom right corner and selecting "Use Zed AI instead".
2. Try to switch to another provider.

**Expected Behavior**: The user should have full control over which prediction provider is used without bias for Zed's.
**Actual Behavior**: In order to switch back one must open the settings file and manually change the setting from "zed" to "copilot". There is no option in the UI settings to switch to another provider. Now that `ctrl-,` opens a UI window instead of the settings file, it creates another barrier to changing providers.

Zed: v0.210.4 (Zed) 
OS: Linux Wayland fedora 43
Memory: 30.5 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV GFX1200) || radv || Mesa 25.2.5

danilo-leal:
Closes https://github.com/zed-industries/zed/issues/41500

Screenshot 2025-10-29 at 9  43@2x

Release Notes:

- Added the ability to switch between configured edit prediction providers through the status bar menu.

### DIFF
--- a/crates/codestral/src/codestral.rs
+++ b/crates/codestral/src/codestral.rs
@@ -66,6 +66,14 @@ impl CodestralCompletionProvider {
         Self::api_key(cx).is_some()
     }
 
+    /// This is so we can immediately show Codestral as a provider users can
+    /// switch to in the edit prediction menu, if the API has been added
+    pub fn ensure_api_key_loaded(http_client: Arc<dyn HttpClient>, cx: &mut App) {
+        MistralLanguageModelProvider::global(http_client, cx)
+            .load_codestral_api_key(cx)
+            .detach();
+    }
+
     fn api_key(cx: &App) -> Option<Arc<str>> {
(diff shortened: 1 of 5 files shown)


versecafe:
Is there a way to make this menu a little smaller as the menu is starting to grow massive especially with zed, codestral, copilot, and supermaven, and likely more to come in the future

danilo-leal:
I think so, yeah! A menu like this is probably not the future of this interaction as you're right, it doesn't scale and it is very limited. We will for sure explore other designs, but I'll do it in a separate opportunity.

PR #41508, opened 2025-10-30
gpui: Hot-reloading with subsecond

someone13574:
This PR adds hot-reloading for gpui using dioxus's subsecond. If a `render`, `paint`, `prepaint`, or `request_layout` function is modified in the tip crate while it is being served, then the change will be made visible. It will not rerun initialization code though. Note that it only [works for the tip crate](https://docs.rs/subsecond/0.7.2/subsecond/index.html#workspace-support) (ie. the one with `main.rs`), so it will be of limited usefulness to zed (unless you can find a creative way to make whatever you are working on the tip crate), but may be useful to other people working in single-crate projects or in gpui examples.

When compiled in release mode, all calls to subsecond are replaced with simple passthroughs. It you think it is better to put it behind a feature gate, I can do that as well.

[Subsecond docs](https://docs.rs/subsecond/0.7.2/subsecond/index.html)

To test (linux):
- Install dioxus-cli: `cargo binstall dioxus-cli --version 0.7.2`
- Go to the gpui crate: `cd crates/gpui`
- Run an example: `dx serve -p gpui --example hello_world --hotpatch`
- Edit the example's paint function and see it update

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -183,6 +183,7 @@ impl Application {
     where
         F: 'static + FnOnce(&mut App),
     {
+        #[cfg(debug_assertions)]
         dioxus_devtools::connect_subsecond();
 
         let this = self.0.clone();
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -183,6 +183,8 @@ impl Application {
     where
         F: 'static + FnOnce(&mut App),
     {
+        dioxus_devtools::connect_subsecond();
+
         let this = self.0.clone();
         let platform = self.0.borrow().platform.clone();
         platform.run(Box::new(move || {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4920,15 +4920,15 @@ dependencies = [
 
 [[package]]
 name = "dioxus-cli-config"
-version = "0.7.0-rc.3"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


someone13574:
I haven't done much testing outside of a few gpui examples, so there are probably other places which need a hot_fn as well

iamnbutler:
@osiewicz I know we've talked about this 👀

crates/gpui/src/app.rs
```diff
@@ -175,6 +175,8 @@ impl Application {
     where
         F: 'static + FnOnce(&mut App),
     {
+        dioxus_devtools::connect_subsecond();
```
mikayla-maki, line 187: Doesn't look like this is free to call, can we compile this out in non-release builds?

mikayla-maki (changes requested):

crates/gpui/src/app.rs
someone13574: Do you mean compile it out in release builds (as opposed to non-release builds)? Or do you want it feature-gated?

mikayla-maki: Whoops, got it backward 😅 . In release mode, this code should not be there. I'm ambivalent on a feature gate.

gusinacio:
Since version 0.7.4 subsecond now supports workspaces https://github.com/DioxusLabs/dioxus/pull/5291

someone13574:
> Since version 0.7.4 subsecond now supports workspaces

Not in a way useful to zed. Only direct workspaces work, not indirect ones: https://github.com/DioxusLabs/dioxus/issues/5314

PR #41514, opened 2025-10-30
Full text search for agent history

cmclaughlin:
This PR implements full text search for the threads database providing the ability for users to search the entire agent history, not only the title.

The implementation uses SQLite's FTS5 (Full Text Search 5) extension. A new virtual table `threads_fts` was added to the database.

The FTS index is automatically maintained during database operations:

1. Save/Update Thread: When `save_thread()` is called
2. Delete Thread: When `delete_thread()` is called:

A background task to build the FTS index for existing threads provides a one-time migration to index exsting threads.

This PR was authored by Claude in Zed. I hope you find it to be acceptable quality.

### DIFF
--- a/crates/agent/src/db.rs
+++ b/crates/agent/src/db.rs
@@ -241,7 +241,7 @@ impl Column for DataType {
     }
 }
 
-pub(crate) struct ThreadsDatabase {
+pub struct ThreadsDatabase {
     executor: BackgroundExecutor,
     connection: Arc<Mutex<Connection>>,
 }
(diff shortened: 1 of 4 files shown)


cmclaughlin:
Discussed here:
https://github.com/zed-industries/zed/discussions/33384
https://github.com/zed-industries/zed/discussions/39338

Would close:
#41514

bennetbo:
Hey, thanks for working on this. I will actually close this PR, since:
1. The demand for this seems pretty low
2. We will have to rebuild this once we support history via ACP (which should hopefully be soon).

cmclaughlin:
@bennetbo thanks for the info. Postponing this until ACP history support is added makes sense.

I was curious about that, so I looked for the issue: https://github.com/zed-industries/zed/issues/37074

In case I or anyone else uses this PR as a starting point for future work, I was playing around with it more over the weekend as a test for running local LLMs. Qwen3 came up with a good suggestion to use triggers for the inserts and deletes into the index instead of doing that in the Rust code. I have mixed feelings about that because, although I'm new to Rust, that seems somewhat easier to understand compared to SQL.

PR #41520, opened 2025-10-30
languages: Add ignore keyword for gomod

pigletfly:
go 1.25 introduce ignore directive in go.mod to specify directories the go command should ignore.

ref: https://tip.golang.org/doc/go1.25#go-command

Release Notes:

- Added syntax highlighting support for the new [`ignore` directive](https://tip.golang.org/doc/go1.25#go-command) in `go.mod` files


probably-neb:
Great work. Thanks

PR #41539, opened 2025-10-30
agent: Hierarchical multi-pass summarization to avoid exceeding token length 

tsoernes:
Implements greedy N-split hierarchical summarization with token-aware chunking, parallel part summarization, and iterative hierarchical merging until final summary fits within model token window. Adds detailed debug logging at each phase. This is done to avoid exceeding token length when creating a summary.

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -1608,17 +1608,17 @@ impl Thread {
             log::error!("No summarization model available");
             return Task::ready(None).shared();
         };
-        let mut request = LanguageModelRequest {
+
+        // Build a full request first; may be split if token count exceeds model limit.
+        let mut full_request = LanguageModelRequest {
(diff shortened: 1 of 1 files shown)


bennetbo:
Could you please explain the motivation behind this change? Have you seen the summary exceed the token length in practice ? I've never had that issue before.
The code seems very convoluted and difficult to maintain. I'll close for now.

tsoernes:
Yes. Sometimes, when attempting to make a summary from the existing thread,
the existing thread exceeds the token limit and thus it is not possible to
create a summary. It is possible to simplify the code such that it splits
the thread in two with overlap, then makes two summareis, then summarize
the two summaries

tsoernes:
@bennetbo the other implementation is simpler. Need this to avoid running out of context length when making summaries

PR #40084, opened 2025-10-13
remote: Wrap `uname` invocation in `sh` for `nu` shell

Veykril:
Closes https://github.com/zed-industries/zed/pull/39994

Release Notes:

- Fixed remoting not working when nushell is set as the default shell on the remote target

PR #41547, opened 2025-10-30
remote: Fix more quoting issues with nushell

Veykril:
https://github.com/zed-industries/zed/pull/40084#issuecomment-3464159871
Closes https://github.com/zed-industries/zed/pull/41547

Release Notes:

- Fixed remoting not working when the remote has nu set as its shell

### DIFF
--- a/crates/project/src/terminals.rs
+++ b/crates/project/src/terminals.rs
@@ -8,7 +8,6 @@ use remote::RemoteClient;
 use settings::{Settings, SettingsLocation};
 use smol::channel::bounded;
 use std::{
-    borrow::Cow,
     path::{Path, PathBuf},
     sync::Arc,
 };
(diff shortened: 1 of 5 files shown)


Veykril:
Needs testing that wsl did not break from this

crates/project/src/terminals.rs
```diff
@@ -171,12 +166,12 @@ impl Project {
                                 let activation_script =
                                     activation_script.join(&format!("{separator} "));
                                 let to_run = format_to_run();
+                                let arg = format!("{activation_script}{separator} {to_run}");
+                                let args = shell_kind.args_for_shell(false, arg);
```
kubkon: Looking at this now, shouldn't we extract the `ShellKind` from `shell` we extracted from the remote client? I know it wasn't part of this changeset but wondering now how we should approach this in the future.

kubkon (approved):

CodeWenjiu:
I'm not sure if this is on my end, but I'm still experiencing this issue. 
image
I'm using NixOS WSL, zed version: 0.212.0 2ead8c.

PR #41565, opened 2025-10-30
git_ui: Add section checkbox and align header

cppcoffee:
Release Notes:

- Added git panel support select all and deselect all check box


cppcoffee:
The git pannel as follow:

截屏2025-10-30 21 13 56

Click the checkbox at the top to select or deselect all, which is convenient for batch operations when there are many files.

JosephTLyons:
Thanks for submitting a PR @cppcoffee, we'll take a look at this soon!

mattermill:
I had started to tackle this myself before I saw this PR, so I'll respond here since you're further along than I am 🙂

The change I'd like to see implemented would hide checkboxes by default, but show them on hover. They would persist when checked. The section headers for tracked/untracked would work the same as you have here.

The only other caveat is that I think they should move over to the right to avoid the icon clutter that can happen while file icons are visible.

cppcoffee:
Hi @mattermill , Added a feature that checkboxes only appear when moving to entries. Please review.

ChristopherBiscardi (approved):
Thanks @cppcoffee ! I've merged main an updated and its working as expected

Image

Issue #26277, opened 2025-03-07
Improve `flake.nix`


P1n3appl3:
> [!IMPORTANT]
> this issue only tracks issues with our in-tree flake and users' ability to _build_ zed, for other issues with using the editor on nixos such as:
> - [fonts not working](https://github.com/zed-industries/zed/issues/18982)
> - [NixOS hosts not working out of the box for remote development](https://github.com/zed-industries/zed/issues/20524)
> - [problems with immutable config files](https://github.com/zed-industries/zed/issues/24537) from the [home-manager module](https://github.com/nix-community/home-manager/blob/04c915bcf1a1eac3519372ff3185beef053fba7c/modules/programs/zed-editor.nix#L33)
> 
> ... you should check the [nix label](https://github.com/zed-industries/zed/labels/nix) in our issues and file a new issue instead of bringing it up here

## Progress

The following table tracks what functionality is known to be working on each platform. All listed commands are run inside the devshell.

|                                             | x86 nixos | arm darwin | x86 linux |
|---------------------------------------------|-----------|------------|-----------|
| `cargo run`                                 | ✅         | ✅          | ✅[^1]     |
| `cargo nextest r --workspace`               | ✅         | ‼️          | ✅         |
| `nix run github:zed-industries/zed/nightly` | ✅        | ✅        | ✅     |
| `cargo r -p remote_server`                  | ✅         | ✅          | ✅         |
| `cargo r -p collab`                         | ✅         | ✅          | ✅         |
| remote server nix build                     |           |            |           |
| collab server nix build                     |           |            |           |
| collab server nixos module                  |           | 🚫          | 🚫         |
| all tests pass in nix sandbox               |     🚫   |   🚫     |   🚫    |
| `script/bundle-`                            | 🚫         |            |           |

key:

|  ✅ | 🚧 | ❓| ‼️ | 🚫 | <empty> |
| --  | -- | -- | -- | -- | ------ |
| working | in progress | unknown, maybe working | broken | not planned | not started yet|

[^1]: Only works if the user already has the ability to run graphical nix applications using nixGL or something similar. see #26980 for more info.

## Context
- nixpkgs has an independently maintained package that is being kept up to date and working by some volunteers
  - we'd like to collaborate with them so they don't have to bend over backwards to package zed
- we have two main priorities for the in-tree flake
  - keep the devshell working as a developer environment for all platforms. try to make sure that users can build zed, remote servers, and the collab server, as well as having all tests pass with no system-level dependencies
[...]

PR #41596, opened 2025-10-30
nix: Add zed-fhs package

poelzi:
This is a direct adoption of the code-fhs package. It runs zed inside a virtual LSB environment, making plugin execution seamless. I'm able to develop a zed plugin on a nix system

Release Notes:

- Added nix flake ".#zed-fhs" for plugin compatibility

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,11 +36,115 @@
           crane = crane.mkLib pkgs;
           rustToolchain = rustBin.fromRustupToolchainFile ./rust-toolchain.toml;
         };
+      # Function to create zed-fhs with optional additional packages
+      mkZedFHS =
+        pkgs: zed:
+        { additionalPkgs ? pkgs: [ ] }:
+        pkgs.buildFHSEnv {
+          name = "zed-fhs";
+
+          # The target packages that will be available in the FHS environment
+          targetPkgs =
+            pkgs:
+            (with pkgs; [
+              # Requested packages
+              nodejs
+              python3
+              yarn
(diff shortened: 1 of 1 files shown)


poelzi:
> You accepted this agreement at 10/30/2025, 7:49:57 PM

poelzi:
I did a 1to1 adoption of code-fhs. Maybe we should trim down the added package list for minimal only. The clang wasm compiler should work for example and the node js lsps etc.
Maybe we just ship musl static binaries for the compilers, that would safe us all headaches. This is super easy with the nix build system.

P1n3appl3:
The zed package in nixpkgs already provides an [fhs-env](https://github.com/NixOS/nixpkgs/blob/daf6dc47aa4b44791372d6139ab7b25269184d55/pkgs/by-name/ze/zed-editor/package.nix#L60). Like I mentioned in #26277 I'm not sure that we want this in our flake, and currently I'm leaning towards "no". If there's some reason that doing extension development against the `nixpkgs-unstable` version of zed isn't sufficient feel free to re-open this and we can figure out what the best thing to do is, but for now the in-tree flake is mostly for the sake of ci and a devshell, we're not officially distributing zed through it.

P1n3appl3:
Another resource that may interest you for zed extension development on nix is the [nix-zed-extensions](https://github.com/DuskSystems/nix-zed-extensions) repo which attempts to generate nix-compatible versions of all zed extensions in our central registry.

PR #41609, opened 2025-10-31
Add agent-servers.md

rtfeldman:
Release Notes:

- N/A

### DIFF
--- a/docs/src/extensions/agent-servers.md
+++ b/docs/src/extensions/agent-servers.md
@@ -2,7 +2,7 @@
 
 Agent Servers are programs that provide AI agent implementations through the [Agent Client Protocol (ACP)](https://agentclientprotocol.com). Agent Server Extensions let you package up an Agent Server so that users can install the extension and have your agent easily available to use in Zed.
 
-You can see the current Agent Server Extensions by opening the Extensions tab in Zed (execute the `zed: extensions` command) that have already been exposed as extensions on [https://zed.dev/extensions](https://zed.dev/extensions?filter=agent-servers).
+You can see the current Agent Server Extensions by opening the Extensions tab in Zed (execute the `zed: extensions` command) and changing the filter from `All` to `Agent Servers` that have already been exposed as extensions on [https://zed.dev/extensions](https://zed.dev/extensions?filter=agent-servers).
 
 ## Defining Agent Server Extensions
 
--- a/docs/src/extensions/agent-servers.md
+++ b/docs/src/extensions/agent-servers.md
@@ -2,7 +2,7 @@
 
 Agent Servers are programs that provide AI agent implementations through the [Agent Client Protocol (ACP)](https://agentclientprotocol.com). Agent Server Extensions let you package up an Agent Server so that users can install the extension and have your agent easily available to use in Zed.
 
-You can see the currently-published Agent Server Extensions either by opening the Extensions tab in Zed (execute the `zed: extensions` command) and changing the filter from `All` to `Agent Servers`, or by visiting [https://zed.dev/extensions?filter=agent-servers](https://zed.dev/extensions?filter=agent-servers).
+You can see the current Agent Server Extensions either by opening the Extensions tab in Zed (execute the `zed: extensions` command) and changing the filter from `All` to `Agent Servers`, or by visiting [https://zed.dev/extensions?filter=agent-servers](https://zed.dev/extensions?filter=agent-servers).
 
 ## Defining Agent Server Extensions
 
--- a/docs/src/extensions/agent-servers.md
+++ b/docs/src/extensions/agent-servers.md
@@ -147,5 +147,7 @@ For optimal display, follow these guidelines:
 - Ensure you have a clean SVG code by processing it through [SVGOMG](https://jakearchibald.github.io/svgomg/)
 - Avoid including icons with gradients as they will often make the SVG more complicated and possibly not render perfectly
 
-Note that we won't include colors; any icon you include will be displayed in monochrome so as to preserve Zed's design consistency.
+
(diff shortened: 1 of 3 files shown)


docs/src/extensions/agent-servers.md
```diff
@@ -0,0 +1,141 @@
...
+[agent_servers.example-agent]
+name = "Example Agent"
+icon = "icon/agent.svg"
+
+[agent_servers.example-agent.env]
+AGENT_LOG_LEVEL = "info"
+AGENT_MODE = "production"
+
+[agent_servers.example-agent.targets.darwin-aarch64]
+archive = "https://github.com/example/agent/releases/download/v2.0.0/agent-darwin-arm64.tar.gz"
```
benbrandt, line 71: Is there no longer a way to specify this for an npm package?

docs/src/extensions/agent-servers.md
```diff
@@ -0,0 +1,141 @@
+# Agent Server Extensions
+
+Agent Servers are programs that provide AI agent implementations through the [Agent Client Protocol (ACP)](https://agentclientprotocol.com). Agent Server Extensions let you package up an Agent Server so that users can install the extension and have your agent easily available to use in Zed.
+
+You can see the agent servers that have already been exposed as extensions on [https://zed.dev/extensions](https://zed.dev/extensions?filter=agent-servers).
```
danilo-leal: Should we point out that these are also visible from the app through the extensions page?

docs/src/extensions/agent-servers.md
```diff
@@ -0,0 +1,141 @@
...
+## Testing
+
+To test your Agent Server Extension:
+
+1. [Install it as a dev extension](./developing-extensions.md#developing-an-extension-locally)
+2. Open the [Agent Panel](../ai/agent-panel.md)
+3. Select your Agent Server from the list
+4. Verify that it downloads, installs, and launches correctly
+5. Test its functionality by conversing with it and watching the [ACP logs](../ai/external-agents.md#debugging-agents)
+
```
danilo-leal, line 142:
```suggestion
## Icon Guideline

In case your agent server has a logo, we highly recommend adding it as an SVG icon.
For optimal display, follow these guidelines:

- Make sure you resize your SVG to fit a 16x16 bounding box, with a padding of around one or two pixels
- Ensure you have a clean SVG code by processing it through [SVGOMG](https://jakearchibald.github.io/svgomg/)
- Avoid including icons with gradients as they will often make the SVG more complicated and possibly not render perfectly

Note that we won't include colors; any icon you include will be displayed in monochrome so as to preserve Zed's design consistency.
```

rtfeldman: These are good guidelines, thanks!

Of note, at the moment we render the colors faithfully, so I guess what would really happen is that if they submitted a PR with colors in the SVG we'd hopefully notice and fix it before accepting it 😄 

Screenshot 2025-10-30 at 11 12 52 AM

docs/src/extensions/agent-servers.md
rtfeldman: There used to be, but I took it out for simplicity of initial launch. This way, we have the maximally flexible thing that can cover everyone's use case, at the cost of some boilerplate. If there's demand for something npm-specific, we can always add it later as a nonbreaking change if desired, but then again someone might also ask for Python too, so there's a "where to draw the line" question that I'm happy to defer for now. 😄

docs/src/extensions/agent-servers.md
danilo-leal: Should probably change the guidelines then to mention that we'll ensure icons are monochrome, but can contain paths with different opacity values to add some layering... so maybe:

> Note that we'll only include monochrome icons to preserve Zed's design consistency. You can still use opacity in different paths of your SVG to add visual layering.

rtfeldman: @danilo-leal Updated the wording! Look good?

danilo-leal: Perfect! 👌

docs/src/extensions/agent-servers.md
benbrandt: Yeah just wondering if all npm packages will have an "archive" for each platform... Or maybe there's a way to satisfy this that I'm missing. Basically for agents like Gemini CLI or our Claude Code adapter that are npm packages without a binary.

docs/src/extensions/agent-servers.md
```diff
@@ -0,0 +1,156 @@
...
+## Icon Guideline
+
+In case your agent server has a logo, we highly recommend adding it as an SVG icon.
+For optimal display, follow these guidelines:
+
+- Make sure you resize your SVG to fit a 16x16 bounding box, with a padding of around one or two pixels
+- Ensure you have a clean SVG code by processing it through [SVGOMG](https://jakearchibald.github.io/svgomg/)
+- Avoid including icons with gradients as they will often make the SVG more complicated and possibly not render perfectly
+
+Note that we'll only accept monochrome icons, to preserve Zed's design consistency. (You can still use opacity in different paths of your SVG to add visual layering.)
```
danilo-leal:
```suggestion
Note that we'll automatically convert your icon to monochrome to preserve Zed's design consistency. (You can still use opacity in different paths of your SVG to add visual layering.)
```

More accurate given https://github.com/zed-industries/zed/pull/42024.

docs/src/extensions/agent-servers.md
rtfeldman: They do, it's just the same URL for each target 😄 

npm provides download URLs like `https://registry.npmjs.org/foo/-/foo-1.0.0.tgz`

PR #41628, opened 2025-10-31
Add support for closing window tabs with middle mouse click

toshmukhamedov:
This change adds support for closing a system window tabs by pressing the middle mouse button.
It improves tab management UX by matching common tab behavior.

Release Notes:

- Added support for closing system window tabs with middle mouse click.


gaauwe:
Thanks, that's a nice improvement!

eugenesvk:
How do you disable this? This "common tab behavior" is a common misfeature where scrolling leading to accidental wheel click risks closing a tab, so should at least be configurable, and I'd argue not even the default

PR #41632, opened 2025-10-31
Add port forwarding UI

alessandrobrunoh:
Summary:

- Adds a graphical interface for configuring Port Forwarding in Zed’s Remote Server Options.
- Integrates with the built-in port forwarding functionality.

Discussion:

- Related discussion: [Link to Discussion](https://github.com/zed-industries/zed/discussions/38924)

Release Notes:

- Added Port Forwarding UI on Remote Server Options

Final Images:
Screenshot 2025-10-31 at 13 26 07
Screenshot 2025-10-31 at 13 27 07
Screenshot 2025-10-31 at 13 25 36
Screenshot 2025-10-31 at 13 27 21


versecafe:
would it make sense to label it "Forward Port" rather than "Add Forward"?

alessandrobrunoh:
> would it make sense to label it "Forward Port" rather than "Add Forward"?

Ye it actually would make more sense. 😅

alessandrobrunoh:
i forgot to run clippy :O

alessandrobrunoh:
should i do something else to restart the 12 skipped?

ConradIrwin:
Nice, thank you for this!

I'd like to make the UI feel a bit less clunky before we merge this. Is it possible to merge the Add/Edit views into the List view?

I was thinking:
* Click "add port forward" and it gives you two empty text boxes (one for source and one for dest)
* Next to each item in the list we show the edit icon, and when you click it the row gets replaced by text boxes
* Hitting enter converts the text boxes back into labels.

But open to other ideas that make it feel smoother.

It would also be nice to support setting either the source or destination's hostname in addition to just the port.

Another thing I noticed while playing with this is that for some reason not all remote servers show have an "open settings" button (e.g. coder.test.2). That doesn't need to block this as it's a pre-existing bug, but would be good to figure it out.

Screenshot 2025-12-04 at 21 42 25

alessandrobrunoh:
@ConradIrwin Thanks, I hope I will have more time to impl. and help this project. 

Sure ill' check to improve the UI with the tips you gave me next week. If there are any other suggestions you can gimme let me know I'll try to impl. them as well. 👍

alessandrobrunoh:
@ConradIrwin  could you copy and paste your remote_servers Settings so I can better investigate the bug you’re seeing? I haven’t been able to reproduce it on my side. Thanks!

ConradIrwin:
Interestingly, it looks like the button only shows up for the servers that are mentioned in my settings file explicitly:

```
  "ssh_connections": [
    {
      "host": "dev",
      "args": [],
      "projects": [
        {
          "paths": [
            "/root/0/go/parallel"
          ]
        }
      ]
    }
  ],
```
Screenshot 2025-12-09 at 09 08 41

PR #41635, opened 2025-10-31
gpui: Add instanced rect/line primitives

vanmelsem:
- Adds instanced rectangles and line segments to render large numbers of shapes in single draw calls
    (much faster than many paint-quad calls).
- Public API: RectInstance, LineInstance, paint_instanced_rects, paint_instanced_lines.
- Metal backend: new pipelines and shaders; correct batch bounds; dynamic instance uploads sized
    to buffer capacity; HSLA retained (opacity supported), with a note on future packed RGBA8
    optimization.
- Cross‑platform: Windows/Linux paths are stubbed (compile, no draw) until implemented.
- Example: instanced_primitives demonstrates usage.

- No CPU culling; batches draw exactly what you provide.
- Follow‑ups: implement Windows/Linux rendering; consider packed RGBA8 for bandwidth; add instanced circles.

image


vanmelsem:
Just added two new examples. (should be removed before pushing, or just leave the one showcasing the actual batched primitives)

- batched_primitives (crates/gpui/examples/batched_primitives.rs): renders many 1px blue dots using paint_batched_rects; drag to pan. Metal-only for now.

- regular_primitives (crates/gpui/examples/regular_primitives.rs): renders the same dots via paint_quad in a loop; drag to pan. Runs on all platforms.

CherryWorm:
https://github.com/CherryWorm/zed/tree/feature/instanced-primitives

This fork also adds Windows support

mikayla-maki:
Hi,

PR #41663, opened 2025-10-31
Move @punctuation.delimiter before @operator capture

diksipav:
Closes #41593

From what I understand the order of captures inside tree-sitter query files matters, and the last capture will win. `?` and `:` are captured by both `@operator` and `@punctuation.delimiter`.So in order for the ternary operator to win it should live after `@punctuation.delimiter`.

Before:
Screenshot 2025-10-31 at 17 41 21

After:
Screenshot 2025-10-31 at 17 41 33

Release Notes:

- Fixed an issue where the ternary operator symbols in TypeScript would not be highlighted as operators.

### DIFF
--- a/crates/languages/src/typescript/highlights.scm
+++ b/crates/languages/src/typescript/highlights.scm
@@ -122,6 +122,15 @@
 ; Tokens
 
 [
+  ";"
+  "?."
+  "."
+  ","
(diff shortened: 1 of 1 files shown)


probably-neb:
Nice catch. Thanks!

PR #41669, opened 2025-10-31
git: Add git clone open listener

AlvaroParker:
Adds a git clone open listener to clone and open repos using `zed zed://git/clone?repo=<url>`

This will also allow creating "Open with Zed" buttons on websites with a href to something like `zed://git/clone?repo=https://github.com/zed-industries/zed.git`.

Demo: 

https://github.com/user-attachments/assets/02fcea70-87a6-4e1d-a42f-918bb0b70a23

Release Notes:

- Added git clone open listeners using `zed://git/clone?repo=<git-url>`


Angelk90:
@AlvaroParker : Would it be possible to do something like this?
`zed zed://git/clone/<url> --single-branch`
`zed zed://git/clone/<url> --single-branch -b feature-a `

look: https://git-scm.com/docs/git-clone#_options

AlvaroParker:
@Angelk90  Currently the git clone backend implementation doesn't support that (single branch, -b argument, etc), and adding that I think is out of scope for this PR

https://github.com/zed-industries/zed/blob/07dcb8f2bb0a28290a2a05d57995cb9ca2a5bf97/crates/fs/src/fs.rs#L962-L978

crates/git_ui/src/clone.rs
```diff
@@ -0,0 +1,207 @@
...
+        let workspace_weak = cx.weak_entity();
+        cx.spawn_in(window, async move |workspace, cx| {
+            let result = async {
+                let mut paths = path_prompt.await.ok()?.ok()??;
+                let path = paths.pop()?;
+
+                let repo_name = repo_url
+                    .split('/')
+                    .next_back()
+                    .and_then(|name| name.strip_suffix(".git"))
```
cole-miller: I think we should probably have this support URLs that don't end with `.git`, so maybe `.next_back().map(|repo_name| repo_name.trim_end_matches(".git")).unwrap_or("repository")`

AlvaroParker:
I'll probably remove the cloning modal notification if #41712 gets merged

PR #41670, opened 2025-10-31
languages: Fix indentation for if/else statements in C/C++ without braces

### LINKED ISSUE #41179: Wrong indentation for else block in C++.
### Summary

The else blocked is moved automatically to right without which is very frustrating to use while coding in C++, and every time I need to correct the indentation of the else block.

### Description

https://github.com/user-attachments/assets/54faa65f-4489-469c-85b2-9a726325b66c

**Expected Behavior**: The else block should be indented correctly for cpp files.
**Actual Behavior**: Indentation of else block is wrong.

Zed: v0.209.5 (Zed) 
OS: macOS 15.7.1
Memory: 48 GiB
Architecture: aarch64

errmayank:
Closes #41179

Release Notes:

- Fixed indentation for if/else statements in C/C++ without braces

### DIFF
--- a/crates/languages/Cargo.toml
+++ b/crates/languages/Cargo.toml
@@ -97,6 +97,7 @@ text.workspace = true
 theme = { workspace = true, features = ["test-support"] }
 tree-sitter-bash.workspace = true
 tree-sitter-c.workspace = true
+tree-sitter-cpp.workspace = true
 tree-sitter-css.workspace = true
 tree-sitter-go.workspace = true
 tree-sitter-python.workspace = true
--- a/crates/languages/src/c.rs
+++ b/crates/languages/src/c.rs
@@ -397,10 +397,10 @@ mod tests {
     use language::{AutoindentMode, Buffer};
     use settings::SettingsStore;
     use std::num::NonZeroU32;
+    use unindent::Unindent;
 
     #[gpui::test]
-    async fn test_c_autoindent(cx: &mut TestAppContext) {
-        // cx.executor().set_block_on_ticks(usize::MAX..=usize::MAX);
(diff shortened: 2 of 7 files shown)


teowelton:
Hey! 

I was also working on this today, but forgot to comment on the issue, so no worries at all. I would just like to point out that this applies to C++ as well (as the original example showed).

_Also,_ feel free to let me know if there was an error in what I was doing, but I ended up solving this by just modifying this:

https://github.com/zed-industries/zed/blob/aa61f25795bc9b039a000858c091b1c891f8ce5d/crates/languages/src/cpp/config.toml#L7

To be:
```
{ pattern = "^\\s*else\\b", valid_after = ["if"] }
```

(for both C & C++), thus not needing to modify `indents.scm` at all, and passing all tests.

Thoughts?

errmayank:
Hey @teowelton, I tested your way and it works great, thanks!

I've also updated this PR as it requires minimal changes that way.

> I would just like to point out that this applies to C++ as well (as the original example showed).

Yeah, I was hoping to create a separate PR for C++ but maybe you can do that?

teowelton:
> Yeah, I was hoping to create a separate PR for C++ but maybe you can do that?

I think that in this case having both in the same PR is much cleaner, since they fall under the same issue. The solution is the same too, simply modifying the `config.toml` file for C++ instead of C:

https://github.com/zed-industries/zed/blob/d887e2050fc90af3d3210e6bd1eee5989431b811/crates/languages/src/cpp/config.toml#L7

errmayank:
@teowelton makes sense

osiewicz (approved):

KingBodo:
> Closes #41179
> 
> Release Notes:
> 
> * Fixed indentation for if/else statements in C/C++ without braces

PR #41698, opened 2025-11-01
remote: Fix incorrect default repository selection when using remote

### LINKED ISSUE #30694: Git: main repository should be default when submodules are present (SSH remote)
### Summary

When opening a repository (via SSH remote) with multiple submodules, the git panel picks one of the submodules (arbitrarily?) as the git the active git repository, instead of the top-level repository.

### Description

Steps to trigger the problem:
1. Open a git repository with submodules via SSH remote.
2. Open the git panel.
3. Look at the active repository/branch selector.

Actual Behavior:

- A random(?) submodule is selected as the repository for the git panel to work with, instead of the top-level repository.

Expected Behavior:

- The top-level git repository should be selected by default.
- The repository picker should have submodules sorted alphabetically. :-)

Zed: v0.186.7 (Zed) 
OS: macOS 15.4.1
Memory: 16 GiB
Architecture: aarch64

amtoaer:
If I understand this correctly: The `active_repo_id` uses `get_or_insert_with`, which makes it dependent on the `RepositoryAdded` event sequence. To ensure correct initialization of the `active_repo_id` on the remote side, the first local `RepositoryAdded` event must synchronously send an `UpdateRepository` to `updates_tx`.

Closes #30694

Release Notes:

- Fixed incorrect default repository selection when using remote (let's git together)

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -1281,7 +1281,7 @@ impl GitStore {
                     .push(cx.subscribe(&repo, Self::on_repository_event));
                 self._subscriptions
                     .push(cx.subscribe(&repo, Self::on_jobs_updated));
-                self.repositories.insert(id, repo.clone());
+                self.repositories.insert(id, repo);
                 cx.emit(GitStoreEvent::RepositoryAdded);
                 self.active_repo_id.get_or_insert_with(|| {
                     cx.emit(GitStoreEvent::ActiveRepositoryChanged(Some(id)));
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks! I think this solution makes sense--what do you think about sending that initial `UpdateRepository` message whenever a new repo is inserted locally, regardless of whether it's the first? That seems a little cleaner and has the benefit that the downstream side's repository list will populate more promptly (even if it doesn't get the actual git statuses, etc. until later).

cole-miller:
Thanks!

esthertrapadoux:
Yeaaahhhh!!! Congrats @amtoaer !!

PR #41700, opened 2025-11-01
git: Check push config before falling back to branch remote

### LINKED ISSUE #26649: Make Git push respect `branch.source.pushRemote` (not just origin)
### Summary

Zed pushes my branch to master

Steps to trigger the problem:
1. git checkout -b mybranch origin/main  (sets upstream to main)
2. git remote add fork git@github.com:fork/zed
3. configure branch.source.pushRemote to fork
4. git push will push to fork, not origin

Actual Behavior:

zed pushes to origin master.

Expected Behavior:

zed pushes to fork/mybranch

Zed: v0.177.7 (Zed) 
OS: Linux Wayland debian unknown
Memory: 15.4 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics (TGL GT2) || Intel open-source Mesa driver || Mesa 24.3.4-3

_No response_

errmayank:
Closes #26649

Release Notes:

- Git will now check pushRemote and pushDefault configurations before falling back to branch remote

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -1812,7 +1812,7 @@ impl GitRepository for RealGitRepository {
                 let output = new_smol_command(&git_binary_path)
                     .current_dir(&working_directory)
                     .args(["config", "--get"])
-                    .arg(format!("branch.{}.remote", branch))
+                    .arg(format!("branch.{branch}.remote"))
                     .output()
                     .await?;
                 if !output.status.success() {
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -559,7 +559,11 @@ pub trait GitRepository: Send + Sync {
         cx: AsyncApp,
     ) -> BoxFuture<'_, Result<RemoteCommandOutput>>;
 
-    fn get_remotes(&self, branch_name: Option<String>) -> BoxFuture<'_, Result<Vec<Remote>>>;
+    fn get_push_remote(&self, branch: String) -> BoxFuture<'_, Result<Option<Remote>>>;
+
+    fn get_branch_remote(&self, branch: String) -> BoxFuture<'_, Result<Option<Remote>>>;
+
+    fn get_all_remotes(&self) -> BoxFuture<'_, Result<Vec<Remote>>>;
 
     /// returns a list of remote branches that contain HEAD
     fn check_for_pushed_commit(&self) -> BoxFuture<'_, Result<Vec<SharedString>>>;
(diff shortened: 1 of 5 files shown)


cole-miller:
Thanks! According to the [docs](https://git-scm.com/docs/git-rev-parse) it seems like we should be able to use `git rev-parse @{push}` to get the information about where `git push` should be directed without running `git config` multiple times--would you be interested in revising this to use that approach?

errmayank:
@cole-miller seems like we could use:

```bash
git rev-parse --abbrev-ref @{push}
```

But it requires remote tracking set up for that branch, so it won't work for new branches being pushed for the first time. Wdyt?

Screenshot 2025-11-05 at 00 02 28

cole-miller:
It's okay if we don't select a remote automatically when the branch has no upstream and there's no git configuration indicating where it should be pushed--we have a fallback to open a picker for this situation (if that `remotes` function returns several remotes), right? We just need to make sure that `git rev-parse @{push}` honors that `pushRemote` config option.

errmayank:
> We just need to make sure that git rev-parse @{push} honors that pushRemote config option.

@cole-miller yeah, looks like it does. It goes in this order:

1. `branch.<BRANCH_NAME>.pushRemote`
2. `remote.pushDefault`
3. `branch.<BRANCH_NAME>.remote`

Here's a screenshot from my tests

Screenshot 2025-11-06 at 12 25 59

Note that my [`push.default`](https://git-scm.com/docs/git-push#Documentation/git-push.txt-pushdefault) is set to `current` here. With the default `simple` mode, `rev-parse` fails whenever the priority determined push remote differs from `branch.<BRANCH_NAME>.remote` which triggers Zed to fallback to the remote picker showing `fork` and `origin` as options in this case.

Here's when `push.default` is set to `simple`

Screenshot 2025-11-06 at 12 54 11

cole-miller:
Thanks for the information! I think it makes sense that we'd honor `push.default` here, so we act the same way as if you did `git push` at the command line (except that we show a picker instead of just erroring if you're in `simple` mode and a remote couldn't be determined). Are there any other issues with using `@{push}`?

The other thing is that I think we should separate out a `GitRepository` method for this, since we shouldn't do `git rev-parse @{push}` for all calls to `GitRepository::remotes` (e.g. if you do `git: fetch from` we want to show a list of all remotes regardless of the current branch/push config). Could you implement that?

errmayank:
> Are there any other issues with using `@{push}`?

Nope, I think that's all.

> The other thing is that I think we should separate out a `GitRepository` method for this, since we shouldn't do `git rev-parse @{push}` for all calls to `GitRepository::remotes` (e.g. if you do `git: fetch from` we want to show a list of all remotes regardless of the current branch/push config). Could you implement that?

Makes sense, I've updated it but instead of creating a separate function I added a `RemoteOperationKind` enum which I think makes it a bit cleaner, let me know what you think.

cole-miller:
Hmm, I think maybe a better structure would be

- `GitRepository::get_all_remotes` which always returns the full list
- `GitRepository::get_push_remote(branch: String)` that looks at `@{push}` (edit: or maybe `the-branch@{push}` to make it not depend on checking out the branch first)
- `GitRepository::get_branch_remote(branch: String)` that looks at `branch.{}.remote`

And then push the fallback to "get all remotes" to the callers--how does that sound?

errmayank:
Sounds good, I've updated `GitRepository` to have separate methods.

crates/git/src/repository.rs
```diff
@@ -1772,29 +1776,63 @@ impl GitRepository for RealGitRepository {
...
+    fn get_branch_remote(&self, branch: String) -> BoxFuture<'_, Result<Option<Remote>>> {
+        let working_directory = self.working_directory();
+        let git_binary_path = self.any_git_binary_path.clone();
+        self.executor
+            .spawn(async move {
+                let working_directory = working_directory?;
+                let output = new_smol_command(&git_binary_path)
+                    .current_dir(&working_directory)
+                    .args(["config", "--get"])
+                    .arg(format!("branch.{}.remote", branch))
```
kubkon:
```suggestion
                    .arg(format!("branch.{branch}.remote"))
```

crates/git/src/repository.rs
```diff
@@ -1772,29 +1776,63 @@ impl GitRepository for RealGitRepository {
...
-                    let output = new_smol_command(&git_binary_path)
-                        .current_dir(&working_directory)
-                        .args(["config", "--get"])
-                        .arg(format!("branch.{}.remote", branch_name))
-                        .output()
-                        .await?;
+                let output = new_smol_command(&git_binary_path)
+                    .current_dir(&working_directory)
+                    .args(["rev-parse", "--abbrev-ref"])
+                    .arg(format!("{}@{{push}}", branch))
```
kubkon:
```suggestion
                    .arg(format!("{branch}@{{push}}"))
```

crates/project/src/git_store.rs
```diff
@@ -4737,12 +4737,27 @@ impl Repository {
...
+                    let remote = match branch_name {
+                        Some(name) => {
+                            if is_push {
+                                backend.get_push_remote(name.clone()).await?
+                            } else {
+                                backend.get_branch_remote(name).await?
+                            }
+                        }
+                        None => None,
+                    };
```
kubkon: I think that using `if-let` is slightly more readable here:

```suggestion
                    let remote = if let Some(branch_name) = branch_name {
                        if is_push {
                            backend.get_push_remote(branch_name).await?
                        } else {
                            backend.get_branch_remote(branch_name).await?
                        }
                    } else {
                        None
                    };
```

kubkon (changes requested):
Looks great, thanks! In addition to some nits I left in my review, we still need to address missing support for remote projects. The entry point to that would be `handle_get_remotes` in `git_store.rs` which should unpack `is_push` from the message envelope. This will require adding `is_push` as a `bool` field in `git.proto` for `message GetRemotes`.

kubkon (approved):
Excellent work, thanks!

PR #41707, opened 2025-11-01
Use `herb` language server as a formatter for `HTML+ERB` files

vitallium:
Hi, this pull request configures the `herb` language server as a formatter for HTML+ERB files.
[herb](https://herb-tools.dev/) is a language server for parsing HTML+ERB files and is enabled by default in the Ruby extension. Without this configuration, the tailwind language server is used for formatting ERB files, but it does not support formatting. I believe it should be safe to add this config option to the default settings. Please let me know if there is another way to do this. Thanks!

Release Notes:

- N/A

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1719,6 +1719,13 @@
         "allowed": true
       }
     },
+    "HTML+ERB": {
+      "formatter": {
+        "language_server": {
+          "name": "herb"
+        }
+      }
+    },
     "Java": {
       "prettier": {
         "allowed": true,


vitallium:
I found that I don't need to specify the formatter explicitly if I define `herb` as a language server explicitly for `HTML+ERB`:

```
diff --git i/assets/settings/default.json w/assets/settings/default.json
index 56a7c36dda..c2a62730f4 100644
--- i/assets/settings/default.json
+++ w/assets/settings/default.json
@@ -1720,11 +1720,7 @@
       }
     },
     "HTML+ERB": {
-      "formatter": {
-        "language_server": {
-          "name": "herb"
-        }
-      }
+      "language_servers": ["herb", "..."]
     },
     "Java": {
       "prettier": {

```

I like that approach more, and I will open a new pull request with the patch above. Thanks!

PR #41709, opened 2025-11-01
extension: Add ability to install local extension from tar.gz file

ZegWe:
Summary:
- Add a feature to install extension from *.tar.gz files.

Release Notes:
- Add an `Install Local Extension` button and an action `zed::install local extension`.

Screenshots:
image

image

image

### DIFF
--- a/crates/extension_host/src/extension_host.rs
+++ b/crates/extension_host/src/extension_host.rs
@@ -791,7 +791,7 @@ impl ExtensionStore {
         })
     }
 
-    pub fn install_extension_from_file(
+    pub fn install_local_extension(
         &mut self,
         extension_source_path: PathBuf,
         cx: &mut Context<Self>,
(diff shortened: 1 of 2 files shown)


maxdeviant:
Thank you for the PR.

While installing local extensions from an archive is something that we'd like to support, the approach here isn't quite right.

Also, note that it is currently possible to install extensions locally by extracting the archive into the extensions directory and then running `zed: reload extensions`.

ZegWe:
Thanks for your reply.

> Also, note that it is currently possible to install extensions locally by extracting the archive into the extensions directory and then running `zed: reload extensions`.

But extensions `installed` by this would not show in the extension panel

> While installing local extensions from an archive is something that we'd like to support, the approach here isn't quite right.

I know the way to check an local extension is quite naive here, I'm not sure if it should be stored in ExtensionManifest and saved in the `index.json` file in extension directory, maybe we can change this to a draft pr and optimize it?

If there's any other ideas, please let me know. I'm glad to hear and contribute to zed.

PR #41712, opened 2025-11-01
git: Add notification to git clone

AlvaroParker:
Adds a simple notification when cloning a repo using the integrated git clone on Zed. Before this, the user had no feedback after starting the cloning action. 

Demo:

https://github.com/user-attachments/assets/72fcdf1b-fc99-4fe5-8db2-7c30b170f12f

Not sure about that icon I'm using for the animation, but that can be easily changed. 

Release Notes:

- Added notification when cloning a repo from zed

### DIFF
--- a/crates/fs/src/fs.rs
+++ b/crates/fs/src/fs.rs
@@ -5,7 +5,6 @@ mod mac_watcher;
 pub mod fs_watcher;
 
 use parking_lot::Mutex;
-use std::collections::HashMap;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Instant;
 
@@ -153,7 +152,6 @@ pub trait Fs: Send + Sync {
     async fn git_clone(&self, repo_url: &str, abs_work_directory: &Path) -> Result<()>;
     fn is_fake(&self) -> bool;
     async fn is_case_sensitive(&self) -> Result<bool>;
-    fn current_job(&self) -> Option<JobInfo>;
 
     #[cfg(any(test, feature = "test-support"))]
     fn as_fake(&self) -> Arc<FakeFs> {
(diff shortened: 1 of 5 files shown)


cole-miller:
Thanks! What do you think about putting this with a spinner in the activity bar at the bottom instead (where it says `Click to restart and update Zed` in the screenshot)? That's a bit more consistent with our current conventions for signalling ongoing work.

image

AlvaroParker:
Since git clone is implemented on the `Fs` trait I went ahead and did something similar to how long running jobs are handled for `LanguageRegistry`: 

https://github.com/zed-industries/zed/blob/a112153a2e8a18be8550c76f8532766acf15b3a6/crates/language/src/language_registry.rs#L1286-L1290

That is, using channels to send status for long running fs jobs and using `fs::subscribe_to_job_events()` to subscribe to this long running jobs.

crates/fs/src/fs.rs
```diff
@@ -215,6 +219,63 @@ pub struct Metadata {
...
+        Self { id }
+    }
+}
+
+impl Drop for JobTracker {
+    fn drop(&mut self) {
+        send_job_event(JobEvent::Completed { id: self.id });
+    }
+}
+
```
cole-miller, line 271: Wondering if we can find a way to structure this code without a global subscription. What if the `ActivityIndicator` held an `Arc<dyn Fs>`, and the `Fs` trait had a `peek_current_job` method that just returned an `Option<String>`? (And I guess a start instant.)

cole-miller (changes requested):
Thanks! A follow-up comment about the code structure but this is looking good.

crates/fs/src/fs.rs
AlvaroParker: This could be done, however `peek_current_job` would have to be called inside the `ActivityIndicator::content_to_render` function, which is called on every render. And because we have an `Arc<dyn Fs>` we will have to keep the jobs inside a `Mutex` hence locking on every render to get the current job, and sometimes when two or more windows are open, multiple `ActivityIndicator` instances will be calling on that function. 

I could try to rework this and make a function on that trait that returns a `JobEventReceiver` like I'm currently doing only this is more integrated to the `Fs` trait and we loose that global subscription

cole-miller: Thank you for the explanation! That makes sense, and I like the new in-between structure with a method on the `Fs` trait.

cole-miller (approved):
Thank you!

PR #41735, opened 2025-11-02
vim: Make each vim repeat its own transaction

AidanV:
Release Notes:

- Pressing `u` after multiple `.` in rapid succession will now only undo the latest repeat instead of all repeats.

### DIFF
--- a/crates/vim/src/normal/repeat.rs
+++ b/crates/vim/src/normal/repeat.rs
@@ -110,7 +110,24 @@ impl Replayer {
         }
         lock.running = true;
         let this = self.clone();
-        window.defer(cx, move |window, cx| this.next(window, cx))
+        window.defer(cx, move |window, cx| {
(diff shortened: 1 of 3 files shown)


AidanV:
I think that the vast majority of the time users would prefer this new behavior, but I am happy to discuss putting it behind a setting or not having it at all.

ConradIrwin:
@AidanV Can you please verify what vim does and add a test for this.

AidanV:
This is the behavior that Vim does. I will keep trying to add a test for this, but the testing framework seems to treat each dispatched keystroke as its own transaction (correct me if I am wrong). This makes any test I do succeed even without these changes, despite it not working correctly in the actual editor.

ConradIrwin:
I added a test that works around the default test behavior (and verified it fails before your fix).

Thanks for sending this!

PR #41739, opened 2025-11-02
agent_ui: Add delete button for compatible API-based LLM providers

aeroxy:
Discussion: https://github.com/zed-industries/zed/discussions/41736

Release Notes:

- agent panel: Added the ability to remove OpenAI-compatible LLM providers directly from the UI.

### DIFF
--- a/crates/agent_ui/src/agent_configuration.rs
+++ b/crates/agent_ui/src/agent_configuration.rs
@@ -1293,6 +1293,12 @@ fn find_text_in_buffer(
         None
     }
 }
+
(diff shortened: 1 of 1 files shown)


aeroxy:
@cla-bot check

danilo-leal (approved):
Thanks for the PR; great idea! 

Little ask for the next time, though, is to be clearer with the title and PR description. Just from them (but even with the discussion post, too), I wouldn't 100% know this was about OpenAI-compatible providers.

PR #41754, opened 2025-11-02
ruby: Disable Ruby LSP for ERB files

### LINKED ISSUE #172: Ruby-LSP and Rubocop cannot be disabled
### Summary

Ruby-lsp and Rubocop keeps starting despite being disabled.

### Description

[Zed docs](https://zed.dev/docs/languages/ruby) notes that you can't use both Solargrapth and Ruby-lsp at the same time. But despite saying that, I've both running, and cannot turn of Ruby-lsp. 

Steps to reproduce:
1. Open Zed and configure it not to use Ruby-lsp.
2. Generate new Rails project.
3.  Open project in Zed.
4. Zed runs Ruby-lsp if an .ERB file is opened.

[settings.json](https://github.com/user-attachments/files/22136999/settings.json)

**Expected Behavior**: I expect Ruby-lsp not to start when configured so in Zed settings. Even if I do have the gem installed.
**Actual Behavior**: Instead Zed tries to start Ruby-lsp no matter what. Even if I have uninstalled the gem and disabled it in Zed settings.

Zed: v0.201.8 (Zed) 
OS: Linux Wayland pop 24.04
Memory: 30.5 GiB
Architecture: x86_64
GPU: AMD Radeon RX 7900 GRE (RADV NAVI31) || radv || Mesa 25.1.5-1pop0~1753463422~24.04~8af185e

vitallium:
The Ruby extension uses the `solargraph`
language server by default for Ruby files.
However, when a user opens any ERB file,
the extension automatically starts the Ruby LSP.
This affects developers because
they do not expect the Ruby LSP to be running.

Closes https://github.com/zed-extensions/ruby/issues/172

Release Notes:

- N/A

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1719,6 +1719,9 @@
         "allowed": true
       }
     },
+    "HTML+ERB": {
+      "language_servers": ["herb", "!ruby-lsp", "..."]
+    },
     "Java": {
       "prettier": {
         "allowed": true,
@@ -1740,6 +1743,9 @@
       "prettier": {
         "allowed": true
       }
+    },
+    "JS+ERB": {
+      "language_servers": ["!ruby-lsp", "..."]
     },
     "Kotlin": {
       "language_servers": ["!kotlin-language-server", "kotlin-lsp", "..."]
@@ -1844,6 +1850,9 @@
       "prettier": {
         "allowed": true
       }
+    },
+    "YAML+ERB": {
+      "language_servers": ["!ruby-lsp", "..."]
     },
     "Zig": {
       "language_servers": ["zls", "..."]


marcoroth:
@vitallium I wonder if this is going to be breaking features like this: https://github.com/Shopify/ruby-lsp-rails/pull/659

/cc @janko

vitallium:
> @vitallium I wonder if this is going to be breaking features like this:
> 
> https://bsky.app/profile/janko.io/post/3m4bqpfxsvs2v
> 
> /cc @janko

Yes, it will break such things, so we need to document how to enable Ruby LSP in `erb` file, I think. I will open a pull request for that. Thanks for the heads up, Marco!

janko:
Since I already have to explicitly enable `ruby-lsp` for ruby files, it makes sense to me that I have to do the same for ERB files 👍🏻

PR #41768, opened 2025-11-02
git: Fix pull failing when tracking remote with different branch name

### LINKED ISSUE #31430: Git: pull doesn't work non-matching branch names
### Summary

Git pull doesn't work if the local branch and the remote branch have different names.

### Description

Steps to trigger the problem:
1. Configure a local branch with a different name from the remote branch. This is an example of how it can look like in the Git config:
    ```
    [branch "master"]
      remote = origin
      merge = refs/heads/main
    ```
1. Make sure the active branch is the branch with the non-matching branch name
1. Open the command palette: Go -> Command Palette...
1. Type `git: pull`
1. Select the pull action

Actual Behavior:

Zed produces the following error:

```
fatal: couldn't find remote ref master
```

Expected Behavior:

The "git pull" command should be successful.

Zed: v0.186.12 (Zed) 5898c2e18c0defd20f644fb3c7944cc765cc0cc5
OS: macOS 15.4.1
Memory: 16 GiB
Architecture: aarch64

errmayank:
Closes #31430

Release Notes:

- Fixed git pull failing when tracking remote with different branch name (Lets Git Together)

Here's a before/after comparison when `dev` branch has upstream set to `origin/main`:

https://github.com/user-attachments/assets/3a47e736-c7b7-4634-8cd1-aca7300c3a73


crates/project/src/git_store.rs
```diff
@@ -1715,7 +1715,11 @@ impl GitStore {
             &mut cx,
         );
 
-        let branch_name = envelope.payload.branch_name.into();
+        let branch_name = if envelope.payload.branch_name.is_empty() {
+            None
+        } else {
+            Some(envelope.payload.branch_name.into())
+        };
```
cole-miller: Minor, but protobuf supports `optional` fields which our bindings will translate to a Rust `Option`, and I'd prefer to use that here--it should be okay to change the type of the field because this particular request is only used for SSH remoting, where the two sides will always be running the same version of Zed.

cole-miller (changes requested):
Thanks, just one comment!

errmayank:
@cole-miller updated

cole-miller (approved):
Thanks! Looks like the buf-breaking check is now required, will figure out how to address that and then merge this.

PR #41782, opened 2025-11-03
remote: Close window when SSH connection fails

kallyaleksiev:
Closes #41781 

## Context 

This PR closes issue https://github.com/zed-industries/zed/issues/41781

It essentially `matches` the result of opening the connection here https://github.com/zed-industries/zed/blob/f7153bbe8a869d0e2b25efed64fd7e4217899b63/crates/recent_projects/src/remote_connections.rs#L650

and adds a Close / Retry alert that upon 'Close' closes the new window if the result is an error

Release Notes:

- Fixed connection spinner hanging when remote connections fail


Veykril (approved):
Thanks!

Issue #11043, opened 2024-04-26
Avatars in git blame for everyone


WeetHet:
### Check for existing issues

- [X] Completed

### Describe the feature

Currently, only GitHub avatars are supported in git blame. Other hostings should be supported as well.

## Avatar support
- [x] GitHub
- [ ] GitLab (main instance)
- [ ] Gitee
- [ ] Bitbucket
- [ ] Sourcehut
- [x] Codeberg (#10991)
- [ ] Self-hosted instances of GitLab
- [ ] Self-hosted instances of Gitea
- [ ] Self-hosted instances of Forjejo
- [ ] Self-hosted instances of Sourcehut

 Anything else with an API

mrnugget:
> * Should this be an API for extensions?

I think for now it's fine to keep it in there, since it's similar to the permalinks behavior we have.

> * Should we make a `Commit`/`Author`/`User`/... struct for each hosting separately or extract them?
>   For now I've extracted the `Author` struct as it's the only one that is built by hand and not deserialised from the response.

I don't think we need more than a URL right now. Once we need more, we can add more.

WeetHet:
> > * Should this be an API for extensions?
> 
> I think for now it's fine to keep it in there, since it's similar to the permalinks behavior we have.
> 
> > * Should we make a `Commit`/`Author`/`User`/... struct for each hosting separately or extract them?
> >   For now I've extracted the `Author` struct as it's the only one that is built by hand and not deserialised from the response.
> 
> I don't think we need more than a URL right now. Once we need more, we can add more.

Okay, I'll undraft the PR and we'll go from there

d1y:
The `Gitee` host need `access_token`  before you can send request fetch avatar
So the user may need to fill it in manually, this is a setting item 🤔

<img width="420" src="https://github.com/zed-industries/zed/assets/45585937/41b0af7a-75ba-48e8-8ced-1a501a197bf2" />


> API: https://gitee.com/api/v5/swagger#/getV5ReposOwnerRepoCommitsSha

```ts
function createURL(owner: string, repo: string, sha: string, token: string) {
  return `https://gitee.com/api/v5/repos/${owner}/${repo}/commits/${sha}?access_token=${token}`
}

const token = Bun.env["ACCESS_TOKEN"]!

const url = createURL('libkitten', 'zed', 'f72cf2afe34f77bf498454cd5486aa2c31b820d6', token)

console.log(url)

// export ACCESS_TOKEN=$$$
[...]

PR #41783, opened 2025-11-03
Support Gitee avatars in git blame

amtoaer:
Part of https://github.com/zed-industries/zed/issues/11043.

CleanShot 2025-11-03 at 10 39 08@2x

Release Notes:

- Added Support for showing Gitee avatars in git blame
- Let's Git Together


Veykril (approved):
Thanks!

PR #41784, opened 2025-11-03
extension_ui: Fix card label truncation

### LINKED ISSUE #41763: Extension Panel usability issue when contact info is longer than panel width
### Summary

GitHub link and hamburger link on the RH side of an Extension's card get eliminated by long contact info.

### Description

For Extensions that have relatively long descriptions (2nd line) and contact info (3rd line):

When the Extension Panel is made so narrow that the whole 2nd line doesn't fit anymore, that description is truncated, an ellipsis is added, and the RH area with the number of downloads is preserved. (Good)

However, when a window doesn't allow the whole 3rd line to fit, the useful GitHub link and hamburger link on the right get eliminated by the contact info. (Not Good)

Image

(Just happening to show this extension, this minor issue happens for all cards when lines don't fit.)

Steps to reproduce:
1. Make the Extension Panel narrower until the info no longer fits.

**Expected Behavior**:
Third line in the card gets truncated with ellipse and the two links in the card are preserved.

**Actual Behavior**:
The two links are wiped out.

I cannot provide any Rust code to "fix" this, at this time, but thanks for the opportunity to provide feedback!

Zed: v0.211.2 (Zed Preview) 
OS: macOS 14.7.1
Memory: 8 GiB
Architecture: aarch64

danilo-leal:
Closes https://github.com/zed-industries/zed/issues/41763

Release Notes:

- N/A

### DIFF
--- a/crates/extensions_ui/src/extensions_ui.rs
+++ b/crates/extensions_ui/src/extensions_ui.rs
@@ -805,25 +805,22 @@ impl ExtensionsPage {
             )
             .child(
                 h_flex()
-                    .gap_2()
+                    .gap_1()
                     .justify_between()
                     .child(
-                        h_flex()
-                            .gap_1()
-                            .child(
-                                Icon::new(IconName::Person)
-                                    .size(IconSize::XSmall)
-                                    .color(Color::Muted),
-                            )
-                            .child(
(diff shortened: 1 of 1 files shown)


warrenjokinen:
Thanks!

PR #40536, opened 2025-10-17
Fix Right Alt key not working in keybindings on Windows

jvb0:
### Problem
On Windows, the right Alt key was not working in keybindings (e.g., `Ctrl+Right Alt+B`), while the left Alt key worked correctly. This was due to overly aggressive AltGr detection that treated any `right Alt + left Ctrl` combination as AltGr, even on US keyboards where AltGr doesn't exist.

### Root Cause
Windows internally represents AltGr (Alt Graph) as `right Alt + left Ctrl` pressed simultaneously. The previous implementation always excluded this combination from being treated as regular modifier keys to support international keyboards. However, this broke keybindings using right Alt on US/UK keyboards where users expect right Alt to behave identically to left Alt.

### Solution
Implemented keyboard layout-aware AltGr detection:

1. Added `uses_altgr()` method to `WindowsKeyboardLayout` that checks if the current keyboard layout is known to use AltGr (German, French, Spanish, Polish, etc.)
2. Modified `current_modifiers()` to only apply AltGr special handling when the keyboard layout actually uses it
3. Added explicit checking for both `VK_LMENU` and `VK_RMENU` instead of relying solely on the generic `VK_MENU`

### Behavior
- **US/UK keyboards**: Right Alt now works identically to left Alt in keybindings. `Ctrl+Right Alt+B` triggers the same action as `Ctrl+Left Alt+B`
- **International keyboards** (German, French, Spanish, etc.): AltGr continues to work correctly for typing special characters and doesn't trigger keybindings
- **All keyboards**: Both Alt keys are detected symmetrically, matching the behavior of left/right Windows keys

### Testing
Manually tested on Windows with US keyboard layout:
-  `Ctrl+Left Alt+B` triggers keybinding
-  `Ctrl+Right Alt+B` triggers keybinding  
-  Both Alt keys work independently in keybindings

Release Notes:
- Fixed Right Alt key not working in keybindings on Windows

PR #41785, opened 2025-11-03
windows: Remove null terminator from keyboard ID

Sh4rK:
Closes #41486, closes #35862 

It is unnecessary, and it broke the `uses_altgr` function.

Also add Slovenian layout as using AltGr.

This should fix:
- https://github.com/zed-industries/zed/pull/40536#issuecomment-3477121224
- https://github.com/zed-industries/zed/issues/41486
- https://github.com/zed-industries/zed/issues/35862

As the current strategy relies on manually adding layouts that have AltGr, it's brittle and not very elegant. It also has other issues (it requests the current layout on every kesytroke and mouse movement).

**A potentially better and more comprehensive solution is at https://github.com/zed-industries/zed/pull/41259**
This is just to fix the immediate issues while that gets reviewed.

Release Notes:

- windows: Fix AltGr handling on non-US layouts again.

### DIFF
--- a/crates/gpui/src/platform/windows/keyboard.rs
+++ b/crates/gpui/src/platform/windows/keyboard.rs
@@ -9,7 +9,6 @@ use windows::Win32::UI::{
     },
     WindowsAndMessaging::KL_NAMELENGTH,
 };
-use windows_core::HSTRING;
 
 use crate::{
     KeybindingKeystroke, Keystroke, Modifiers, PlatformKeyboardLayout, PlatformKeyboardMapper,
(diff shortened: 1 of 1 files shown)


crates/gpui/src/platform/windows/keyboard.rs
```diff
@@ -135,6 +133,7 @@ impl WindowsKeyboardLayout {
                 b"0405" | // Czech
                 b"040E" | // Hungarian
                 b"0424" | // Slovenian
+                b"041A" | // Croatian
```
tidylogic, line 136: Hello!

Thank you for your work on the ALTGR mapping changes in this PR.

Since you are already modifying the relevant section, I have a related, small request: would it be possible to humbly ask you to also include support for the Korean ALTGR mapping?

It only requires adding the following line to the existing Rust code block
```rust
b"0412" | // Korean
```
This simple addition would greatly help users who rely on the Korean input method with ALTGR.

Thank you very much for your time and for considering this specific addition!

Sh4rK: Hi! I looked at the Korean layout here:

- https://kbdlayout.info/features/flags
- https://kbdlayout.info/kbdkor

and it doesn't mention AltGr usage (nor does it show any AltGr characters that I can see), like for the other languages in this list, so I wasn't sure if it's the same issue. Of course, I'm not familiar with Korean layouts, so if you say that this will work, I can add it. Out of curiosity, which key combination (AltGr + ?) are you trying to use, and what is the expected result?

tidylogic: I sincerely apologize for the confusion. Upon further reflection, I believe the issue I'm experiencing is actually completely separate from those issues. 

I'm truly sorry for any confusion this may have caused.

nia-e:
Looks good, thanks for the fix ^^

Sh4rK:
Thanks!

@nia-e Not trying to tell you what to do, just signal boosting, if you have the capacity you should take a look at https://github.com/zed-industries/zed/pull/41259 because it seems to properly solve the AltGr issue without hard-coding specific languages (and without the current problem of requesting the layout on every event). It also supposedly solves other issues as well. I'm not an expert in Windows input handling, so I can't give much input on it, but I think it's worth a look.

Sh4rK:
@nia-e Oh, never mind, the creator of that PR is also a member of Zed :D

Issue #41422, opened 2025-10-29
Relative line number display error


CnsMaple:
### Summary

Why am I 36 at 28?

https://github.com/user-attachments/assets/38d4d0a9-10c7-4ed0-9f2e-ef5b5c7ca9bc

Why is it always 1 at the bottom?

https://github.com/user-attachments/assets/2e5c23bd-2a1e-48d9-976f-bcadb142bc56

### Zed Version and System Specs

Zed: v0.211.0 (Zed Nightly 4cc6d6a398ad4b77b224e0675226c3a690aa8a71) 
OS: Windows 10.0.19045
Memory: 31.7 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 750 || Intel Corporation || 32.0.101.6881

CnsMaple:
This is my setting:

<img width="501" height="151" alt="Image" src="https://github.com/user-attachments/assets/b2c93066-43e6-4eee-8762-930125c8071d" />

<img width="495" height="181" alt="Image" src="https://github.com/user-attachments/assets/a2ec05bd-6e34-4f6b-a367-332d0749ba2c" />

CnsMaple:
Do you have time to take a look? @thomasheartman

MrSubidubi:
Thanks for flagging this! Will revert for the moment and see how things go from there, given the release in a few hours.

thomasheartman:
Thanks for reporting this @CnsMaple, and sorry for the slow response; I was on holiday. Thanks also @MrSubidubi for reverting it in time for the release. The good news is that I was able to reproduce the second issue you mentioned and I have managed to fix it locally; the issue is [this commit](https://github.com/zed-industries/zed/pull/39268/commits/e7096d27a6463f6eb7c2a821637c5773b2460c10). It was suggested that we could avoid the double calculation, but it seems that we can't. Or at least not this way. I wasn't able to reproduce the first issue (28 to 36), but I expect that that will also be fixed with the first one. I'll submit a new PR later this week when I'm back home.

MrSubidubi:
No need to apologize, thanks for taking a look and hope you had a good vacation!

CnsMaple:
@thomasheartman The first problem requires a relatively large number of lines, such as 1000 lines. Then, the cursor is set at the 500th line position, and the scroll wheel is only scrolled. Then, around the 30th line, the line jumps. It should be a rendering issue. Currently, it is only rendering the number of line numbers that is twice the size of my screen.

[...]

PR #39268, opened 2025-10-01
Support relative line number on wrapped lines

thomasheartman:
**Problem:** Current relative line numbering creates a mismatch with vim-style navigation when soft wrap is enabled. Users must mentally calculate whether target lines are wrapped segments or logical lines, making `<n>j/k` navigation unreliable and cognitively demanding.

**How things work today:**
- Real line navigation (`j/k` moves by logical lines): Requires determining if visible lines are wrapped segments before jumping. Can't jump to wrapped lines directly.
- Display line navigation (`j/k` moves by display rows): Line numbers don't correspond to actual row distances for multi-line jumps.

**Proposed solution:** Count and number each display line (including wrapped segments) for relative numbering. This creates direct visual-to-navigational correspondence where the relative number shown always matches the `<n>j/k` distance needed.

**Benefits:**
- Eliminates mental overhead of distinguishing wrapped vs. logical lines
- Makes relative line numbers consistently actionable regardless of wrap state
- Preserves intuitive "what you see is what you navigate" principle
- Maintains vim workflow efficiency in narrow window scenarios

Also explained an discussed in https://github.com/zed-industries/zed/discussions/25733.

Release Notes:

Release Notes:

- Added support for counting wrapped lines as relative lines and for displaying line numbers for wrapped segments. Changes `relative_line_numbers` from a boolean to an enum: `enabled`, `disabled`, or `wrapped`.

PR #41805, opened 2025-11-03
Support relative line number on wrapped lines (rework)

thomasheartman:
## Add relative line numbers on wrapped lines, take 2

This is a rework of https://github.com/zed-industries/zed/pull/39268 that excludes https://github.com/zed-industries/zed/pull/39268/commits/e7096d27a6463f6eb7c2a821637c5773b2460c10. This commit introduced some line number rendering issues as described in https://github.com/zed-industries/zed/issues/41422.

While @ConradIrwin suggested we try to pass in the buffer rows from the calling method instead of the snapshot, that 
appears to have had unintended consequences and I don't think the two calculations were intended to do the same thing. Hence, this PR has removed those changes.

This PR also includes the migration fix originally done by @MrSubidubi in https://github.com/zed-industries/zed/pull/41351.

## Original PR description and release notes.

**Problem:** Current relative line numbering creates a mismatch with vim-style navigation when soft wrap is enabled. Users must mentally calculate whether target lines are wrapped segments or logical lines, making `<n>j/k` navigation unreliable and cognitively demanding.

**How things work today:**
- Real line navigation (`j/k` moves by logical lines): Requires determining if visible lines are wrapped segments before jumping. Can't jump to wrapped lines directly.
- Display line navigation (`j/k` moves by display rows): Line numbers don't correspond to actual row distances for multi-line jumps.

**Proposed solution:** Count and number each display line (including wrapped segments) for relative numbering. This creates direct visual-to-navigational correspondence, where the relative number shown always matches the `<n>j/k` distance needed.

**Benefits:**
- Eliminates mental overhead of distinguishing wrapped vs. logical lines
- Makes relative line numbers consistently actionable regardless of wrap state
- Preserves intuitive "what you see is what you navigate" principle
- Maintains vim workflow efficiency in narrow window scenarios

Also explained and discussed in https://github.com/zed-industries/zed/discussions/25733.

Release Notes:

- Added support for counting wrapped lines as relative lines and for displaying line numbers for wrapped segments. Changes `relative_line_numbers` from a boolean to an enum: `enabled`, `disabled`, or `wrapped`.

### DIFF
--- a/crates/migrator/src/migrations/m_2025_10_21/settings.rs
+++ b/crates/migrator/src/migrations/m_2025_10_21/settings.rs
@@ -1,0 +1,16 @@
+use anyhow::Result;
+use serde_json::Value;
+
+pub fn make_relative_line_numbers_an_enum(value: &mut Value) -> Result<()> {
+    let Some(relative_line_numbers) = value.get_mut("relative_line_numbers") else {
+        return Ok(());
+    };
+
+    *relative_line_numbers = match relative_line_numbers {
+        Value::Bool(true) => Value::String("enabled".to_string()),
+        Value::Bool(false) => Value::String("disabled".to_string()),
+        Value::String(s) if s == "enabled" || s == "disabled" || s == "wrapped" => return Ok(()),
+        _ => anyhow::bail!("Expected relative_line_numbers to be a boolean"),
+    };
+    Ok(())
+}
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -592,7 +592,7 @@
   // to both the horizontal and vertical delta values while scrolling. Fast scrolling
   // happens when a user holds the alt or option key while scrolling.
   "fast_scroll_sensitivity": 4.0,
-  "relative_line_numbers": false,
+  "relative_line_numbers": "disabled",
   // If 'search_wrap' is disabled, search result do not wrap around the end of the file.
   "search_wrap": true,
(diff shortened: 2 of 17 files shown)


crates/migrator/src/migrations/m_2025_10_21/settings.rs
thomasheartman, line 1: should the name of this file be updated? the date still points to when the last PR was merged.

thomasheartman:
@ConradIrwin I'm sure you're aware of the issues that were discovered with the last implementation that was merged last week. As discussed in https://github.com/zed-industries/zed/issues/41422#issuecomment-3479815400, it seems that using the buffer rows instead of passing along the snapshot did not work. 

Apart from the changes mentioned in the PR description, though, this change should be perfectly equivalent to the previous PR (and it is, at least as far as I can tell). Let me know if you want to take any more action with this one, whether that's renaming migration files or adding additional tests.

ConradIrwin:
Thank you! To confirm you've tested it doesn't have that problem? :D

thomasheartman:
Indeed! Or at least, I can't replicate it anymore after commit [3ad184b](https://github.com/zed-industries/zed/pull/41805/commits/3ad184b788e4a373fbe6f370b0517ab96d68979a) (the current head of the branch). On the commit before, however, I **can** reliably reproduce it. But please feel free to verify it yourself.

PR #41807, opened 2025-11-03
node_runtime: Improve proxy mapping

tidely:
Closes #ISSUE

More accurately map localhost to `127.0.0.1`. Previously we would lowercase and simply replace all instances of localhost inside of the URL string, meaning query parameters, username, password etc. could not contain the string `localhost` or contain uppercase letters without getting modified. Added a test ensuring the mapping logic works. The previous implementation would fail this new test.

Release Notes:

- Improved the behavior of mapping `localhost` to `127.0.0.1` when passing configured proxy urls to `node`


probably-neb:
Seems reasonable. Thanks!

PR #41813, opened 2025-11-03
Support Forgejo and Gitea avatars in git blame

amtoaer:
Part of #11043.

Codeberg is a public instance of Forgejo, as confirmed by the API documentation at https://codeberg.org/api/swagger. Therefore, I renamed the related component from codeberg to forgejo and added codeberg.org as a public instance.

Furthermore, to optimize request speed for the commit API, I set `stat=false&verification=false&files=false`.

CleanShot 2025-11-03 at 19 57 06@2x

<br/>
<br/>

Regarding Gitea Support: 

Forgejo is a fork of Gitea, and their APIs are currently identical (e.g., for getting avatars). However, to future-proof against potential API divergence, I decided to treat them as separate entities. The current gitea implementation is essentially a copy of the forgejo file with the relevant type names and the public instance URL updated.

Release Notes:

- Added Support for Forgejo and Gitea avatars in git blame

### DIFF
--- a/crates/git_hosting_providers/src/providers/forgejo.rs
+++ b/crates/git_hosting_providers/src/providers/forgejo.rs
@@ -1,0 +1,430 @@
+use std::str::FromStr;
+use std::sync::Arc;
+
+use anyhow::{Context as _, Result, bail};
+use async_trait::async_trait;
+use futures::AsyncReadExt;
+use gpui::SharedString;
+use http_client::{AsyncBody, HttpClient, HttpRequestExt, Request};
+use serde::Deserialize;
+use url::Url;
+
+use git::{
+    BuildCommitPermalinkParams, BuildPermalinkParams, GitHostingProvider, ParsedGitRemote,
+    RemoteUrl,
+};
+
+use crate::get_host_from_git_remote_url;
+
+#[derive(Debug, Deserialize)]
+struct CommitDetails {
+    #[expect(
+        unused,
+        reason = "This field was found to be unused with serde library bump; it's left as is due to insufficient context on PO's side, but it *may* be fine to remove"
+    )]
+    commit: Commit,
+    author: Option<User>,
+}
+
+#[derive(Debug, Deserialize)]
+struct Commit {
+    #[expect(
+        unused,
+        reason = "This field was found to be unused with serde library bump; it's left as is due to insufficient context on PO's side, but it *may* be fine to remove"
+    )]
+    author: Author,
+}
+
+#[derive(Debug, Deserialize)]
+struct Author {
+    #[expect(
+        unused,
+        reason = "This field was found to be unused with serde library bump; it's left as is due to insufficient context on PO's side, but it *may* be fine to remove"
+    )]
+    name: String,
+    #[expect(
+        unused,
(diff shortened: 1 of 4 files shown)


amtoaer:
Also see https://github.com/zed-industries/zed/issues/11043#issuecomment-3480446231

cole-miller:
Thank you for working on this! I'll give this a review later this week.

crates/git_hosting_providers/src/providers/forgejo.rs
```diff
@@ -67,31 +69,69 @@ struct User {
...
+        };
+        let url = format!(
+            "https://{host}/api/v1/repos/{repo_owner}/{repo}/git/commits/{commit}?stat=false&verification=false&files=false"
+        );
 
         let mut request = Request::get(&url)
             .header("Content-Type", "application/json")
             .follow_redirects(http_client::RedirectPolicy::FollowAll);
 
+        // TODO: not renamed yet for compatibility reasons, may require a refactor later
```
cole-miller, line 126: Could we only send `CODEBERG_TOKEN` if the remote is actually Codeberg?

crates/git_hosting_providers/src/providers/gitea.rs
```diff
@@ -0,0 +1,423 @@
...
+};
+
+use crate::get_host_from_git_remote_url;
+
+#[derive(Debug, Deserialize)]
+struct CommitDetails {
+    #[expect(
+        unused,
+        reason = "This field was found to be unused with serde library bump; it's left as is due to insufficient context on PO's side, but it *may* be fine to remove"
+    )]
```
cole-miller: I don't think we need to have these unused fields in the new structs that are being defined here

crates/git_hosting_providers/src/providers/gitea.rs
```diff
@@ -0,0 +1,423 @@
...
+    pub fn new(name: impl Into<String>, base_url: Url) -> Self {
+        Self {
+            name: name.into(),
+            base_url,
+        }
+    }
+
+    pub fn public_instance() -> Self {
+        Self::new("Gitea", Url::parse("https://gitea.com").unwrap())
+    }
```
cole-miller, line 44: Does it make sense to have this? As far as I can tell `gitea.com` doesn't really host anything itself.

cole-miller (changes requested):
Thanks! I left a few relatively minor comments, but this looks generally good. I confirmed that line permalinks and avatars work well with a checkout of https://codeberg.org/redict/redict on this branch.

crates/git_hosting_providers/src/providers/gitea.rs
amtoaer: Yes, they do host projects. You can confirm this on the [gitea.com/explore/repos](https://gitea.com/explore/repos) page, where one of the most well-known projects is the Go ORM framework, Xorm.

cole-miller: Thanks, I stand corrected!

cole-miller (approved):
Thanks!

PR #41814, opened 2025-11-03
zed: Reworked open_log_file to reuse the active workspace

cppcoffee:
Release Notes:

- Improved open log reuse the active workspace


cppcoffee:
The open log function reuses the existing workspace to avoid opening a new window.

Just deleted:

```rust
     workspace
        .with_local_workspace(window, cx, move |workspace, window, cx| {
```

PR #41842, opened 2025-11-03
gpui: Use DWM API for backdrop effects and add Mica/Mica Alt support

CosmicPredator:
This PR updates window background rendering to use the **official DWM backdrop API** (`DwmSetWindowAttribute`) instead of the legacy `SetWindowCompositionAttribute`.
It also adds **Mica** and **Mica Alt** options to `WindowBackgroundAppearance` for native Windows 11 effects.

### Motivation

Enables modern, stable, and GPU-accelerated backdrops consistent with Windows 11’s Fluent Design.
Removes reliance on undocumented APIs while maintaining backward compatibility with older Windows versions.

### Changes

* Added `MicaBackdrop` and `MicaAltBackdrop` variants.
* Switched to DWM API for applying backdrop effects.
* Verified fallback behavior on Windows 10.

### Release Notes:

- Added `WindowBackgroundAppearance::MicaBackdrop` and `WindowBackgroundAppearance::MicaAltBackdrop` for Windows 11 Mica and Mica Alt window backdrops. 

### Screenshots

- `WindowBackgroundAppearance::Blurred`
image

- `WindowBackgroundAppearance::MicaBackdrop`
image

- `WindowBackgroundAppearance::MicaAltBackdrop`
image

### DIFF
--- a/crates/gpui/src/platform.rs
+++ b/crates/gpui/src/platform.rs
@@ -1392,7 +1392,7 @@ pub enum WindowBackgroundAppearance {
     /// The Mica backdrop material, supported on Windows 11.
     MicaBackdrop,
     /// The Mica Alt backdrop material, supported on Windows 11.
-    MicaAltBackdrop
+    MicaAltBackdrop,
 }
 
 /// The options that can be configured for a file dialog prompt
--- a/crates/gpui/src/platform.rs
+++ b/crates/gpui/src/platform.rs
@@ -1389,24 +1389,9 @@ pub enum WindowBackgroundAppearance {
     ///
     /// Not always supported.
     Blurred,
-    /// The Mica backdrop material used in Windows 11.
-    ///
-    /// This effect blends the window background with the desktop wallpaper
-    /// to produce a soft, tinted translucency that subtly tracks wallpaper
-    /// movement. It’s designed to provide depth and visual hierarchy while
-    /// remaining performance-friendly.
-    ///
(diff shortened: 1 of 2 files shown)


CosmicPredator:
@reflectronic apologies, it was a misclick

reflectronic:
> Verified fallback behavior on Windows 10.

This patch makes `Transparent` and `Blurred` not work anymore on Windows 10, right? I think we should keep the `SetWindowCompositionAttribute` fallback, and only use it if `DwmSetWindowAttribute` returns an error. I would like to avoid breaking this configuration for Zed users on Windows 10. SWCA is rather widely used and an update is unlikely to break it.

Are you planning on using these new values for your own GPUI application?

CosmicPredator:
> This patch makes `Transparent` and `Blurred` not work anymore on Windows 10, right? I think we should keep the `SetWindowCompositionAttribute` fallback, and only use it if `DwmSetWindowAttribute` returns an error.

Ok, I'll push a patch on this. I'll make `Blurred` and `Transparent` use SWCA and the `DSWA` for the new backdrop materials.

> Are you planning on using these new values for your own GPUI application?

Yes indeed!

CosmicPredator:
@reflectronic Please take a look and lemme know any other changes required!

reflectronic (approved):
Sorry for the delay. This looks good to me.

CosmicPredator:
> Sorry for the delay. This looks good to me.

It's fine. Thanks btw 👍

Issue #11200, opened 2024-04-30
Change completion direction when top-down(Near cursor)


xzbdmw:
### Check for existing issues

- [X] Completed

### Describe the feature

When completion menu opens above current line, show selection of completion menu near cursor.

### If applicable, add mockups / screenshots to help present your vision of the feature

https://github.com/hrsh7th/nvim-cmp/wiki/Menu-Appearance#custom-menu-direction

JosephTLyons:
Hey @xzbdmw, can you show a screenshot of what you are running into?

xzbdmw:
Yeah, I want the selected entry always near my cursor, I think it is nature to human eyes.
Currently zed always select the top entry no matter completion direction.

https://github.com/zed-industries/zed/assets/97848247/b128750e-cef7-4ac5-8139-009d02343eda

notpeter:
So you invert the order of the completion list so the "best" / "first" selection appears at the bottom of the list closest to the cursor. I like it.

PR #36653, opened 2025-08-21
vim: Add vim counts and vim shortcuts to project_panel

AidanV:
Closes #10930 
Closes #11353

Release Notes:

- Adds commands to project_panel
  - `ctrl-u` scrolls the project_panel up half of the visible entries
  - `ctrl-d` scrolls the project_panel down half of the visible entries
  - `z z` scrolls current selection to center of window
  - `z t`  scrolls current selection to top of window
  - `z b` scrolls current selection to bottom of window
  - `{num} j` and `{num} k` now move up and  down with a count

PR #41844, opened 2025-11-03
gpui: Add 'Nearest' scrolling strategy to 'UniformList'

vasyl-protsiv:
This PR introduces `Nearest` scrolling strategy to `UniformList`. This is now used in completions menu and the picker to choose the appropriate scrolling strategy depending on movement direction. Previously, selecting the next element after the last visible item caused the menu to scroll with `ScrollStrategy::Top`, which scrolled the whole page and placed the next element at the top. This behavior is inconsistent, because using `ScrollStrategy::Top` when moving up only scrolls one element, not the whole page.

https://github.com/user-attachments/assets/ccfb238f-8f76-4a18-a18d-bbcb63340c5a

The solution is to introduce the `Nearest` scrolling strategy which will internally choose the scrolling strategy depending on whether the new selected item is below or above currently visible items. This ensures a single-item scroll regardless of movement direction.

https://github.com/user-attachments/assets/8502efb8-e2c0-4ab1-bd8d-93103841a9c4

I also noticed that some functions in the file have different logic depending on `y_flipped`. This appears related to reversing the order of elements in the list when the completion menu appears above the cursor. This was a feature suggested in #11200 and implemented in #23446. It looks like this feature was reverted in #27765 and there currently seem to be no way to have `y_flipped` to be set to `true`. 

My understanding is that the opposite scroll strategy should be used if `y_flipped`, but since there is no way to enable this feature to test it and I don't know if the feature is ever going to be reintroduced I decided not to include it in this PR.

Release Notes:

- gpui: Add 'Nearest' scrolling strategy to 'UniformList'

### DIFF
--- a/crates/editor/src/code_context_menus.rs
+++ b/crates/editor/src/code_context_menus.rs
@@ -460,8 +460,13 @@ impl CompletionsMenu {
         cx: &mut Context<Editor>,
     ) {
         if self.selected_item != match_index {
+            let strategy = if match_index > self.selected_item {
+                ScrollStrategy::Bottom
+            } else {
+                ScrollStrategy::Top
+            };
             self.selected_item = match_index;
-            self.handle_selection_changed(provider, window, cx);
+            self.handle_selection_changed(provider, window, cx, strategy);
         }
     }
 
@@ -486,9 +491,10 @@ impl CompletionsMenu {
         provider: Option<&dyn CompletionProvider>,
         window: &mut Window,
         cx: &mut Context<Editor>,
+        strategy: ScrollStrategy,
     ) {
         self.scroll_handle
-            .scroll_to_item(self.selected_item, ScrollStrategy::Top);
+            .scroll_to_item(self.selected_item, strategy);
         if let Some(provider) = provider {
             let entries = self.entries.borrow();
             let entry = if self.selected_item < entries.len() {
@@ -1124,7 +1130,7 @@ impl CompletionsMenu {
     ) {
         *self.entries.borrow_mut() = matches.into_boxed_slice();
         self.selected_item = 0;
-        self.handle_selection_changed(provider.as_deref(), window, cx);
+        self.handle_selection_changed(provider.as_deref(), window, cx, ScrollStrategy::Top);
     }
 
     pub fn sort_string_matches(
--- a/crates/editor/src/code_context_menus.rs
+++ b/crates/editor/src/code_context_menus.rs
@@ -488,7 +488,7 @@ impl CompletionsMenu {
         cx: &mut Context<Editor>,
     ) {
         self.scroll_handle
-            .scroll_to_item(self.selected_item, ScrollStrategy::Top);
+            .scroll_to_item(self.selected_item, ScrollStrategy::Nearest);
         if let Some(provider) = provider {
             let entries = self.entries.borrow();
             let entry = if self.selected_item < entries.len() {
--- a/crates/editor/src/code_context_menus.rs
+++ b/crates/editor/src/code_context_menus.rs
@@ -460,13 +460,8 @@ impl CompletionsMenu {
         cx: &mut Context<Editor>,
     ) {
         if self.selected_item != match_index {
-            let strategy = if match_index > self.selected_item {
-                ScrollStrategy::Bottom
-            } else {
-                ScrollStrategy::Top
-            };
             self.selected_item = match_index;
-            self.handle_selection_changed(provider, window, cx, strategy);
+            self.handle_selection_changed(provider, window, cx);
         }
     }
 
@@ -491,10 +486,9 @@ impl CompletionsMenu {
         provider: Option<&dyn CompletionProvider>,
         window: &mut Window,
         cx: &mut Context<Editor>,
-        strategy: ScrollStrategy,
     ) {
         self.scroll_handle
-            .scroll_to_item(self.selected_item, strategy);
+            .scroll_to_item(self.selected_item, ScrollStrategy::Top);
         if let Some(provider) = provider {
             let entries = self.entries.borrow();
             let entry = if self.selected_item < entries.len() {
@@ -1130,7 +1124,7 @@ impl CompletionsMenu {
     ) {
         *self.entries.borrow_mut() = matches.into_boxed_slice();
         self.selected_item = 0;
-        self.handle_selection_changed(provider.as_deref(), window, cx, ScrollStrategy::Top);
+        self.handle_selection_changed(provider.as_deref(), window, cx);
     }
 
     pub fn sort_string_matches(
--- a/crates/gpui/src/elements/uniform_list.rs
+++ b/crates/gpui/src/elements/uniform_list.rs
@@ -433,8 +433,7 @@ impl Element for UniformList {
                                         .max(Pixels::ZERO);
                                 }
                                 ScrollStrategy::Bottom => {
-                                    updated_scroll_offset.y = -(item_bottom - list_height
-                                        + offset_pixels)
+                                    updated_scroll_offset.y = -(item_bottom - list_height)
                                         .max(Pixels::ZERO)
                                         .min(content_height - list_height)
                                         .max(Pixels::ZERO);
--- a/crates/gpui/src/elements/uniform_list.rs
+++ b/crates/gpui/src/elements/uniform_list.rs
@@ -92,6 +92,10 @@ pub enum ScrollStrategy {
     /// May not be possible if there's not enough list items above the item scrolled to:
     /// in this case, the element will be placed at the closest possible position.
     Bottom,
+    /// If the element is not visible attempt to place it at:
+    /// - The top of the list's viewport if the target element is above currently visible elements.
+    /// - The bottom of the list's viewport if the target element is above currently visible elements.
+    Nearest,
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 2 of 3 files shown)


vasyl-protsiv:
@cla-bot check

vasyl-protsiv:
@cla-bot check

MrSubidubi:
Thank you so much for coming forward with this! Was recently bothered by this as well and am more than happy to see a fix for this.

Also, thanks a lot for the context!

Looking at this, can't help but wonder if we should not just push this to the GPUI level: I don't really see any picker where we would want the current behavior to be the default, it is just not intuitive. Perhaps we could add another method to the uniform list for just this scenario and/or another variant to the ScrollStrategy enum to account for this? What do you think?

vasyl-protsiv:
Initially I was thinking about doing it there but then I noticed that the `UniformList` doesn't explicitly store the index of previously selected item which made me doubt that it's a good place for it.
Now, after having a closer look at it, I feel like I have better understanding, but I am still lacking in exact details especially regarding padding and offset. 

The logic in uniform list is a bit hard to follow but I can try to describe my thoughts. First, look at:
https://github.com/zed-industries/zed/blob/2eeb02305cd5392ed561fb9631e6d078ac05099f/crates/gpui/src/elements/uniform_list.rs#L405-L419

Notice that the simplification is possible:

```rust
                        let mut scrolled_to_top = false;

                        if item_top < scroll_top + offset_pixels { // Let's call it condition A
                            scrolled_to_top = true;
                            // todo: using the padding here is wrong - this only works well for few scenarios
                            updated_scroll_offset.y = -item_top + padding.top + offset_pixels;
                        } else if item_bottom > scroll_top + list_height { // Let's call it condition B
                            scrolled_to_top = true;
                            updated_scroll_offset.y = -(item_bottom - list_height);
                        }
                        // Here the 'scrolled_to_top' is essentially A || B
                        if deferred_scroll.scroll_strict
                            || (scrolled_to_top
                                && (item_top < scroll_top + offset_pixels
                                    || item_bottom > scroll_top + list_height))
                        // This condition is (scroll_strict || (scrolled_to_top && (A || B)))
                        // Which is the same as just (scroll_strict || scrolled_to_top)
                        // Or equivalently (scroll_strict || A || B)
```

Now after this part is simplified we can look again on a larger chunk of code that includes the the match expression:
```rust
                        let mut scrolled_to_top = false;

                        if item_top < scroll_top + offset_pixels { // A
                            scrolled_to_top = true;
                            // todo: using the padding here is wrong - this only works well for few scenarios
                            updated_scroll_offset.y = -item_top + padding.top + offset_pixels; // Assignment C
                        } else if item_bottom > scroll_top + list_height { // B
                            scrolled_to_top = true;
                            updated_scroll_offset.y = -(item_bottom - list_height); // Assignment D
                        }

                        if deferred_scroll.scroll_strict || scrolled_to_top { // scroll_strict || A || B
                            match deferred_scroll.strategy {
                                ScrollStrategy::Top => {
                                    updated_scroll_offset.y = -(item_top - offset_pixels)
                                        .max(Pixels::ZERO)
                                        .min(content_height - list_height)
                                        .max(Pixels::ZERO); // Assignment E
                                }
                                ScrollStrategy::Center => {
                                    let item_center = item_top + item_height / 2.0;

                                    let viewport_height = list_height - offset_pixels;
                                    let viewport_center = offset_pixels + viewport_height / 2.0;
                                    let target_scroll_top = item_center - viewport_center;

                                    updated_scroll_offset.y = -target_scroll_top
                                        .max(Pixels::ZERO)
                                        .min(content_height - list_height)
                                        .max(Pixels::ZERO); // Assignment F
                                }
                                ScrollStrategy::Bottom => {
                                    updated_scroll_offset.y = -(item_bottom - list_height
                                        + offset_pixels)
                                        .max(Pixels::ZERO)
                                        .min(content_height - list_height)
                                        .max(Pixels::ZERO); // Assignment G
                                }
                            }
                        }
                        scroll_offset = *updated_scroll_offset

```
Here I claim that both assignments C and D are effectively dead code since to execute either of them A || B must be true which would lead to execution of the match expression. Notice then that each branch of the match expression would overwrite `updated_scroll_position.y` (assignments E, F and G).
This simplifies the code even more, but it feels strange that there is so much unused code, so there likely is some bug here (or I my reasoning).

But if we assume that the code is correct then I can suggest the following conjecture:
- We should use `ScrollStrategy::Top` iff Condition A is true 
- We should use `ScrollStrategy::Bottom` iff Condition B is true
- If neither A nor B we should not scroll at all

This all is assuming `scroll_strict` is false. I am not sure that I exactly understand what it is for, but I think we can keep existing logic there which is to just use the `deferred_scroll.strategy` regardless of A and B.


Combining all ideas together: we get the following code:
```rust
						//...
                        let offset_pixels = item_height * deferred_scroll.offset;

                        let scroll_strategy = if deferred_scroll.scroll_strict {
                            Some(deferred_scroll.strategy)
                        } else if item_top < scroll_top + offset_pixels {
                            Some(ScrollStrategy::Top)
                        } else if item_bottom > scroll_top + list_height {
                            Some(ScrollStrategy::Bottom)
                        } else {
                            None
                        };
                        if let Some(scroll_strategy) = scroll_strategy {
                            match scroll_strategy {
                                ScrollStrategy::Top => {
                                    updated_scroll_offset.y = -(item_top - offset_pixels)
                                        .max(Pixels::ZERO)
                                        .min(content_height - list_height)
                                        .max(Pixels::ZERO);
                                }
                                ScrollStrategy::Center => {
                                    let item_center = item_top + item_height / 2.0;

                                    let viewport_height = list_height - offset_pixels;
                                    let viewport_center = offset_pixels + viewport_height / 2.0;
                                    let target_scroll_top = item_center - viewport_center;

                                    updated_scroll_offset.y = -target_scroll_top
                                        .max(Pixels::ZERO)
                                        .min(content_height - list_height)
                                        .max(Pixels::ZERO);
                                }
                                ScrollStrategy::Bottom => {
                                    updated_scroll_offset.y = -(item_bottom - list_height
                                        + offset_pixels)
                                        .max(Pixels::ZERO)
                                        .min(content_height - list_height)
                                        .max(Pixels::ZERO);
                                }
                            }
                        }

```
After short testing this code seem to work locally on completions menu and picker (outline and project symbols) but I am not too certain in correctness of this, especially when offset and padding is added to the mix. Also I am not sure if we should modify scroll handle api, since now the `strategy` argument for `scroll_to_item` and `scroll_to_item_with_offset` (the only two methods that set `scroll_strict` to false) is effectively ignored.

Let me know what do you think of this. Should I update this PR with this code?

ConradIrwin:
We talked about this in office hours, looks like this may have been an incidental change from #36653. @AidanV is going to test the proposed code in the vim mode actions to check it works, but otherwise makes sense to us.

AidanV:
This does not regress the vim mode actions; however, when I test the scrolling with these changes I don't think it fixes the issue yet. I am still getting the scroll top behavior on most modals (i.e. symbol search, file picker, command palette, etc.). I might be missing something.

I don't have the time at the moment to look into this deeper but scroll_strict definitely served a purpose so I am a little concerned. 

Also, I personally prefer seeing where I am scrolling to so I am not sure if just scrolling bottom is the right move by default. I would prefer if it worked like the editor where there is some lines of look ahead when scrolling down.

I just wanted to give a quick response so that I am not blocking this, but feel free to ignore most of this.

vasyl-protsiv:
Just to double check @AidanV, you did try the code from the final comment, which contain the gpui fix, and not the code from PR branch, which just fixes the completions menu in editor crate. Right?
If so then can you clarify which picker did you use? I believe the gpui fix fixes the pickers you mentioned.

https://github.com/user-attachments/assets/cae54093-74d9-4ab9-97c6-b571daff5409

<details>
  <summary>Full patch</summary>

```patch
diff --git a/crates/gpui/src/elements/uniform_list.rs b/crates/gpui/src/elements/uniform_list.rs
index 93082563c0..08f38e745d 100644
--- a/crates/gpui/src/elements/uniform_list.rs
+++ b/crates/gpui/src/elements/uniform_list.rs
@@ -402,23 +402,18 @@ impl Element for UniformList {
                         let item_bottom = item_top + item_height;
                         let scroll_top = -updated_scroll_offset.y;
                         let offset_pixels = item_height * deferred_scroll.offset;
-                        let mut scrolled_to_top = false;

-                        if item_top < scroll_top + offset_pixels {
-                            scrolled_to_top = true;
-                            // todo: using the padding here is wrong - this only works well for few scenarios
-                            updated_scroll_offset.y = -item_top + padding.top + offset_pixels;
+                        let scroll_strategy = if deferred_scroll.scroll_strict {
+                            Some(deferred_scroll.strategy)
+                        } else if item_top < scroll_top + offset_pixels {
+                            Some(ScrollStrategy::Top)
                         } else if item_bottom > scroll_top + list_height {
-                            scrolled_to_top = true;
-                            updated_scroll_offset.y = -(item_bottom - list_height);
-                        }
-
-                        if deferred_scroll.scroll_strict
-                            || (scrolled_to_top
-                                && (item_top < scroll_top + offset_pixels
-                                    || item_bottom > scroll_top + list_height))
-                        {
-                            match deferred_scroll.strategy {
+                            Some(ScrollStrategy::Bottom)
+                            None
+                        };
+                        if let Some(scroll_strategy) = scroll_strategy {
+                            match scroll_strategy {
                                 ScrollStrategy::Top => {
                                     updated_scroll_offset.y = -(item_top - offset_pixels)
                                         .max(Pixels::ZERO)
```
</details>

I can push this fix here, the reason I didn't do it is because it extends the scope of PR and probably makes it harder to test. Let me know if you think I should do it.

AidanV:
Ohhh, yeah I think that you should add the GPUI fixes to the PR, but I also don't have to review this so take that with a grain of salt. The changes seem to work for the most part for the Vim mode in the project panel, but there seems to be a little bit of a weird behavior when scrolling down.

Screenshot From 2025-11-05 17-14-19

Pressing down scrolls us down two elements instead of one.

Screenshot From 2025-11-05 17-14-24

All other scrolls seem to work where they only scroll down one element.

vasyl-protsiv:
Added 'Nearest' scrolling strategy. Currently only used for completions and in the picker. Let me know if you think we should use it in some other lists (perhaps in project panel?).

There is currently no difference between strict and not strict `Nearest` scrolls. We can consider some interface change if you believe this is an issue (maybe different set of strategies for strict/non-strict?).

MrSubidubi (approved):
Wanna emphasize again: Fantastic write-up and context given, thank you so much for that. Looks good to me and does its job perfectly, so thank you very much for that.

One final ask: Could I ask you to perhaps look into adding a test for these changes? I very much would like for this to not regress at any time, hence I'd like to ask for this. 

As for the other stuff, definitely worth improving but I'd personally argue it would be fine to move that to another PR. That said, if you want to add these changes here as well, feel free to.

vasyl-protsiv:
Added tests. I tried to use `logical_scroll_top_index` but it always returns 0 for some reason. So I instead just save the visible range to the view itself and then assert on that. Hope it is okay.

MrSubidubi (approved):
Beautiful, thank you so so much! Will hit merge after the release cut today just to be sure, that way we can check whether anything breaks for a week on nightly.

Being totally honest, was an absolute pleasure to discuss this with you here - fantastic analysis from your side, amazing context given which helped a lot with reviewing this. To me personally this definitely was the best first time contribution I've ever seen here, so really, thank you so much for this! I hope you had as much fun working on this as me reviewing this, if not more, looking forward to your next PR if you have time and interest!

That said (albeit a few hours from now), congratulations to your first contribution! 🎉

PR #41851, opened 2025-11-03
ruby: Add note about enabling Ruby LSP for ERB files

vitallium:
Hi, this is a follow-up change for https://github.com/zed-industries/zed/pull/41754 I think it important to keep existing things working. So add notes to the Ruby extension doc about enabling Ruby LSP for ERB files as well. Thanks!

Release Notes:

- N/A

### DIFF
--- a/docs/src/languages/ruby.md
+++ b/docs/src/languages/ruby.md
@@ -71,6 +71,18 @@ To switch to `ruby-lsp`, add the following to your `settings.json`:
   "languages": {
     "Ruby": {
       "language_servers": ["ruby-lsp", "!solargraph", "!rubocop", "..."]
+    },
+    // Enable herb and ruby-lsp for *.html.erb files
+    "HTML+ERB": {
+      "language_servers": ["herb", "ruby-lsp", "..."]
+    },
+    // Enable ruby-lsp for *.js.erb files
+    "JS+ERB": {
+      "language_servers": ["ruby-lsp", "..."]
+    },
+    // Enable ruby-lsp for *.yaml.erb files
+    "YAML+ERB": {
+      "language_servers": ["ruby-lsp", "..."]
     }
   }
 }


docs/src/languages/ruby.md
```diff
@@ -71,6 +71,18 @@ To switch to `ruby-lsp`, add the following to your `settings.json`:
   "languages": {
     "Ruby": {
       "language_servers": ["ruby-lsp", "!solargraph", "!rubocop", "..."]
+    },
+    // Enable herb and ruby-lsp for *.html.erb files
```
vitallium, line 75: `herb` is enabled by default for `*.html.erb`.

docs/src/languages/ruby.md
```diff
@@ -71,6 +71,18 @@ To switch to `ruby-lsp`, add the following to your `settings.json`:
...
+    "HTML+ERB": {
+      "language_servers": ["herb", "ruby-lsp", "..."]
```
vitallium, line 77: Following the note above, I think we could leave it as `["ruby-lsp", "..."]` but I want to show that we explicitly enable both `herb` and `ruby-lsp`.

marcoroth:
Dumb question: does it hurt to keep `ruby-lsp` enabled by default for any `*.erb` file, so people don't have to opt-in?

vitallium:
> Dumb question: does it hurt to keep `ruby-lsp` enabled by default for any `*.erb` file, so people don't have to opt-in?

In some way, yes, it does. Zed uses `solargraph` by default and the Ruby extension has some open issues ([1](https://github.com/zed-extensions/ruby/issues/172), [2](https://github.com/zed-extensions/ruby/issues/194)) where users wonder why they have `ruby-lsp` enabled even they have it disabled explicitly in their settings. 

Adding a requirement in enabling `ruby-lsp` for ERB files explicitly is not ideal, I agree. Hopefully, the situation will improve once extensions have more control over language server configurations for files.

PR #41852, opened 2025-11-03
git: Stage files after resolving conflicts with in-editor buttons

### LINKED ISSUE #31137: Git: "Take ours" does not always work when only ours is taken in a file
### Summary

Selecting "Take ours" for all conflicts in a file is causing the diff to disappear, but then later reappear with the same conflict again that I have already resolved. 

### Description

Steps to trigger the problem:
1. Create a merge conflict (make sure there are a few files)
2. Select "Take ours" for all conflicts within a file
3. As you continue going down the list, the file will come back with conflicts that you had thought you had resolved

Actual Behavior: No operation performed from using all "Take ours"
Expected Behavior: The conflict is resolved and staged after the last "Take ours" for the file

Zed: v0.187.4 (Zed) 
OS: macOS 15.3.1
Memory: 24 GiB
Architecture: aarch64

dawsh2:
## Summary

Fixes #31137

When users click "Use HEAD" / "Use Origin" / "Use Both" buttons to resolve merge conflicts in the editor, the files are now properly staged with `git add`. Previously, the conflict markers were removed and files were saved, but they remained in git's "unmerged" state, causing them to reappear as conflicted.

## Changes

Modified `resolve_conflict()` in `crates/git_ui/src/conflict_view.rs` to:
1. Get the repository and file path using `repository_and_path_for_buffer_id()`
2. Call `repo.stage_entries()` after successful file save
3. Handle errors gracefully with logging

This uses the exact same staging mechanism as the git panel checkbox, which already worked correctly.

## Testing

Reproduced the bug:
- Created merge conflicts in 3 files
- Used in-editor "Use HEAD" buttons to resolve
- **Before fix**: Files showed as unmerged in `git status`
- **After fix**: Files properly staged, git shows "All conflicts fixed"

## Release Notes

- Fixed merge conflicts not being staged when resolved using in-editor "Use HEAD" / "Use Origin" / "Use Both" buttons ([#31137](https://github.com/zed-industries/zed/issues/31137)) (Lets Git Together)

## Questions and Comments:
- There seems to have been a terminology change from "Take Ours" --> "Use Head", "Take Theirs" --> "Take Origin", and the mismatch between UI and codebase lingo was not addressed during this change. If any action should be taken here, such as filing an issue or a separate PR, please let me know.  
- git_panel.rs triggers an error toast if the operation is not successful, I wasn't sure whether to include something similar or not.

### DIFF
--- a/crates/git_ui/src/conflict_view.rs
+++ b/crates/git_ui/src/conflict_view.rs
@@ -443,7 +443,7 @@ pub(crate) fn resolve_conflict(
     cx: &mut App,
 ) -> Task<()> {
     window.spawn(cx, async move |cx| {
-        let Some((workspace, project, multibuffer, buffer)) = editor
+        let Some((workspace, project, multibuffer, buffer, resolves_all_conflicts)) = editor
             .update(cx, |editor, cx| {
                 let workspace = editor.workspace()?;
                 let project = editor.project()?.clone();
@@ -466,6 +466,7 @@ pub(crate) fn resolve_conflict(
                     })
                     .ok()?;
                 let &(_, block_id) = &state.block_ids[ix];
+                let resolves_all_conflicts = state.block_ids.len() == 1;
                 let range =
                     snapshot.anchor_range_in_excerpt(excerpt_id, resolved_conflict.range)?;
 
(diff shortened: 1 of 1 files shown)


ConradIrwin:
@dawsh2 nice! Thanks for this. I like the idea of removing the need to manually stage after conflicts are resolved, but I think this will cause a problem for a file with more than a single conflict marker, because it'll stage the file after resolving the first conflict.

Are you able to update this to only stage the file if all conflicts in a file are resolved?

dawsh2:
> @dawsh2 nice! Thanks for this. I like the idea of removing the need to manually stage after conflicts are resolved, but I think this will cause a problem for a file with more than a single conflict marker, because it'll stage the file after resolving the first conflict.
> 
> Are you able to update this to only stage the file if all conflicts in a file are resolved?

Good point! Thank you. Yes I'll update it accordingly.

dawsh2:
Thanks for the feedback! Updated to check if we're resolving the last conflict (block_ids.len() == 1) before saving and staging. Files are now only staged when all their conflicts are resolved.

Tested with multiple files with multiple conflicts - worked as expected.

ConradIrwin:
Great, thanks! Could you please appease CI

ConradIrwin:
Sorry for the late notice.

We discussed this internally and came to the conclusion that for people who want to manually manage their staging area, this is not desired behavior.

I'm going to close this for now, but we should come back to this when we have a better story around merge conflict resolution

dawsh2:
Sounds good, thanks!

PR #41856, opened 2025-11-03
Shell out to real tar in extension builder

ConradIrwin:
We see `test_extension_store_with_test_extension` hang in untarring the
WASI SDK some times.

In lieu of trying to debug the problem, let's try shelling out for now
in the hope that the test becomes more reliable.

There's a bit of risk here because we're using async-tar for other
things (but probably not 300Mb tar files...)

Assisted-By: Zed AI

Closes #ISSUE

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5837,8 +5837,6 @@ name = "extension"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "async-compression",
- "async-tar",
  "async-trait",
  "collections",
  "dap",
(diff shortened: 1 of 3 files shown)


maxdeviant:
Is this going to cause problems on Windows?

ConradIrwin:
Claude doesn't think so, but I'm going to test

ConradIrwin:
Can confirm this works on windows; but I ran into problems downloading the tar on windows (though as soon as I started debugging in earnest, they went away...)

Issue #35803, opened 2025-08-07
Tree view for Git Panel


notpeter:
### Summary

Support for a nested tree-view in the Git Panel.

### Description

See also:
- https://github.com/zed-industries/zed/issues/26097
- https://github.com/zed-industries/zed/pull/27230

### Screenshot

Nate's mockup:
<img width="1344" height="1514" alt="Image" src="https://github.com/user-attachments/assets/2e908f44-c889-4954-ab82-ebd44033d2ff" />

FrGoIs:
I decided to work on this issue. Here are my assumptions from the mock and the discussion (corrections are most welcome):

- The toggle at the right switches between the view that we want to build and the existing one (for how files are represented, the panel header will remain as in the mockup).
- The button next to the word `Changes` (containing the lines changed) will open the diff.
- The checkbox at the start will check or uncheck all items (stage all or unstage all) even if they are untracked.

VLZH:
@FrGoIs I will wait for your implementation! I am a big user of git tree view in VsCode. I think this is very useful because folders group related files together, which makes it easier to review changes in the right order.

FrGoIs:
@VLZH Almost done! I expect to make the PR this weekend.

YukiOta:
+1, I’m also excited about this. Tree view would make reviewing changes much easier.

douglance:
the `lazygit` implementation is really nice. 

- clicking folders opens the diff as a list view of a series of files. 
- clicking a file opens only that file's diff. 

I really like that functionality because it gives you both views based on the user's choice in the moment, rather than having to toggle.

the list view of all diffs is actually confusing to me when I dont expect it because I'll scroll through and i'll be in a different file without knowing it. 

it would be great if the git diff view follows the same sorting rules for the file explorer panel view, and would be great to see file icons there as well (file icons are more important thatn the change icon to me, and change status could be a color of the text instead of the change icon).

I would also really like the ability to view the *entire file* rather than just the area around changes. Users have to jump back and forth from the diff to the file and back to fully understand the changes if they're not totally familiar with the full contents of the file. It should be a choice at least. 

The `VSCode` system of having a staged area and unstaged area is helpful as well. It's lower cognitive overhead vs. inspecting all the icons.

thanks so much for working on this! its the only reason I can't use zed full time.

spiff1281:
+1

Yaphet2015:
[...]

PR #27230, opened 2025-03-21
Make filenames in git panel match filenames in multibuffer diff

blake-c-dev:
Closes #26097 

Release Notes:

- Files in nested directories are hard to distinguish. So files in git diff panel are now show by filename, then full filename with directory. Similar to VSCode and the multibuffer diff.

PR notes:
- I can see how users may want to keep the old filename display. Maybe this should be a user setting?

Before:
Screenshot 2025-03-20 at 9 15 10 PM
After:
Screenshot 2025-03-20 at 9 15 01 PM

PR #41857, opened 2025-11-03
Improve readability of files in the git changes panel

alphathekiwi:
Closes _unknown_

image

This PR places the file_name before the file_path so that when the panel is slim it is still usable, mirrors the behaviour of the file picker (cmd+P)

Release Notes:
-  Improved readability of files in the git changes panel

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -50,6 +50,7 @@ use panel::{
 use project::{
     Fs, Project, ProjectPath,
     git_store::{GitStoreEvent, Repository, RepositoryEvent, RepositoryId, pending_op},
+    project_settings::{GitPathStyle, ProjectSettings},
 };
 use serde::{Deserialize, Serialize};
 use settings::{Settings, SettingsStore, StatusStyle};
@@ -3949,6 +3950,7 @@ impl GitPanel {
         cx: &Context<Self>,
     ) -> AnyElement {
         let path_style = self.project.read(cx).path_style(cx);
+        let git_path_style = ProjectSettings::get_global(cx).git.path_style;
         let display_name = entry.display_name(path_style);
 
         let selected = self.selected_entry == Some(ix);
@@ -4048,7 +4050,13 @@ impl GitPanel {
         } else {
             cx.theme().colors().ghost_element_active
         };
-
+        let name_label =
+            |status: FileStatus, label_color: Color, this: Div, label: String| -> Div {
+                this.child(
+                    self.entry_label(label, label_color)
+                        .when(status.is_deleted(), |this| this.strikethrough()),
+                )
+            };
         h_flex()
             .id(id)
             .h(self.list_item_height())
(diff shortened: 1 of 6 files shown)


alphathekiwi:
@cla-bot check

cole-miller:
Thanks for opening a PR! We're discussing internally whether this design is something we want for the git panel, and will let you know what the outcome of that discussion is.

xipeng-jin:
Hi @alphathekiwi , I like your idea. Previously, I saw there is a similar PR https://github.com/zed-industries/zed/pull/27230 before. It was being closed due to Zed team would like to handle this case with a different design for Tree-View Git Panel in this ticket https://github.com/zed-industries/zed/issues/35803. I have been working on that and get a initial setup, but I think the same UI issue is still exist since there is a switch button on top so that you can click on it switch between the tree-view or flat-view as shown here

Screenshot 2025-11-04 at 12 50 55 PM

If you switch to Flat-View you will get the same design as right now, which means the readability of files in git changes are still not great!.
If Zed team is interested in accepting your PR, I am looking forward to integrating it into the final design for git panel with Tree-view. I think that would make the final UI looks even better. I am happy to have a further discussion on this. Thank you. cc @cole-miller

alphathekiwi:
The use case of file name first is perfect for when you are editing a bunch of files that all have different names and it is how we do it in the file picker and how VScode does their git view.

The use case for file path first is when edit a lot of files that have the same name translation.json inside of lang/en, lang/cz, etc so i don't want to discredit this as being bad it just should perhaps be a selectable option.

I would hope we could maybe have all three options available to us, perhaps we could expand the selector you have to be like the crude screenshot I have shown

usecase

alphathekiwi:
> Hi @alphathekiwi , I like your idea. Previously, I saw there is a similar PR #27230 before. It was being closed due to Zed team would like to handle this case with a different design for Tree-View Git Panel in this ticket #35803. I have been working on that and get a initial setup, but I think the same UI issue is still exist since there is a switch button on top so that you can click on it switch between the tree-view or flat-view as shown here
> 
> Screenshot 2025-11-04 at 12 50 55 PM
> If you switch to Flat-View you will get the same design as right now, which means the readability of files in git changes are still not great!. If Zed team is interested in accepting your PR, I am looking forward to integrating it into the final design for git panel with Tree-view. I think that would make the final UI looks even better. I am happy to have a further discussion on this. Thank you. cc @cole-miller

I didn't know about https://github.com/zed-industries/zed/pull/27230 so the fact that two people wanted this code change enough to make it into a PR and each with different reasons probably indicates there is a non zero amount of people who would like this to exist in some form or fashion. This is my first attempt to make a PR and I don't really understand how the Zed settings system works, otherwise I probably would have made this a setting.

Regardless of the outcome though I want to take the time to thank you for your comprehensive reply and for also taking the time to be very welcoming, I hope to contribute more to the project in future.

ConradIrwin:
I think this makes sense, but as a setting (there are some projects where the directory contains a bunch of index.js and tests.js and you need to see the project folder to know).

I do think this makes sense as a default (it'd be better for most projects I use) even though it breaks the obvious sorting behaviour.

I know there's also work going on for the tree view, and other display options here so maybe something like:


```
{"git_panel": {"filenames": "filename_first"}} // "tree_view" | "full"
```

But defer to @cole-miller who's working on this

alphathekiwi:
@cole-miller at your convenience I would love to pair on making this have a setting, I loosely understand how settings work currently with the exception of the Settings UI portion and would love to work on this, I am hanging out in the zed/git1.0 channel fairly regularly most mornings NZT (afternoon/evenings in American time)

cole-miller:
@alphathekiwi Sorry for the delay! We'd like to help get this ready to merge, could you book a pairing time here? https://cal.com/esther-trapadoux-zed/30min

Anthony-Eid:
@alphathekiwi To add a setting to the settings ui you mainly have to add it to this Vec https://github.com/zed-industries/zed/blob/5dde87d2b8ae43ce3d3b246c5e99c3adba118a0b/crates/settings_ui/src/page_data.rs#L22-L23.

alphathekiwi:
> @alphathekiwi To add a setting to the settings ui you mainly have to add it to this Vec
> 
> https://github.com/zed-industries/zed/blob/5dde87d2b8ae43ce3d3b246c5e99c3adba118a0b/crates/settings_ui/src/page_data.rs#L22-L23
> 
> .

Cool have added settings now have also allowed an option for:
https://github.com/zed-industries/zed/pull/41857
Although probably that's incorrect as now that I'm thinking about it that should be added with their PR not mine - late night coding lol 🤣

crates/settings_ui/src/page_data.rs
```diff
@@ -5476,6 +5476,19 @@ pub(crate) fn settings_data(cx: &App) -> Vec<SettingsPage> {
...
+                    field: Box::new(SettingField {
+                        json_path: Some("git.show_change_counters"),
+                        pick: |settings_content| settings_content.git.as_ref()?.show_change_counters.as_ref(),
+                        write: |settings_content, value| {
+                            settings_content.git.get_or_insert_default().show_change_counters = value;
+                        },
+                    }),
+                    metadata: None,
+                    files: USER,
+                }),
```
Anthony-Eid, line 5509: This looks like a different setting than the `GitPathStyle` setting your PR added. Is this a setting that was missing from the settings UI? 

The second to last step to finalizing this PR is adding the `GitPathStyle` setting to the UI

crates/settings_ui/README.md
Anthony-Eid, line 1: Having a file like this would be very helpful, but it should be added through its own PR.

I also think this file is a bit too information-dense in some areas and misses the why the settings are structured the way they are. I'll be happy to help you clean it up for a different PR 😀

crates/git_ui/src/git_panel.rs
```diff
@@ -4153,11 +4161,17 @@ impl GitPanel {
                     .items_center()
                     .flex_1()
                     // .overflow_hidden()
```
Anthony-Eid: While we're editing this code could you also please remove this comment?

crates/settings_ui/src/page_data.rs
Anthony-Eid: It seems like CICD is failing because `show_change_counter` isn't a real value. I'll be happy to help you fix this though.

Anthony-Eid (changes requested):

crates/settings_ui/README.md
alphathekiwi: Okay sounds good :)

crates/git_ui/src/git_panel.rs
alphathekiwi: Done

crates/settings_ui/src/page_data.rs
alphathekiwi: Ahhh sorry about that, its a combination of broken rust-analyser and trying to work on two things at once.

Anthony-Eid (approved):

PR #41874, opened 2025-11-04
agent: History and recent conversations persistence per workspace

aeroxy:
Release Notes:

- Introduce HistoryScope to persist recent agent history using workspace-specific keys.
- Derive history scope from workspace metadata in the agent panel and reuse it across views.
- Update unit tests to account for the scoped history store behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,7 @@ dependencies = [
  "uuid",
  "watch",
  "web_search",
+ "workspace",
  "worktree",
  "zed_env_vars",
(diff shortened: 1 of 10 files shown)


benbrandt:
hi @aeroxy thanks for the work here. We are currently discussing how to make some of these history features work with ACP, and so i would like to revisit this once we can look at this more holistically. 

I'm going to close for now, but we can revisit once we take a look at history for ACP agents + our agent. Thanks!

PR #41880, opened 2025-11-04
assistant_slash_commands: Fix AI text thread path display bugs on Windows and all platforms

CLoaKY233:
## Fix incorrect directory path folding in slash command file collection

**Description:**
This PR fixes a bug in the `collect_files` function where the directory folding logic (used to compact chains like `.github/workflows`) failed to reset its state when traversing out of a folded branch.

**The Issue:**
The `folded_directory_names` accumulator was persisting across loop iterations. If the traversal moved from a folded directory (e.g., `.github/workflows`) to a sibling directory (e.g., `.zed`), the sibling would incorrectly inherit the prefix of the previously folded path, resulting in incorrect paths like `.github/.zed`.

**The Fix:**
*   Introduced `folded_directory_path` to track the specific path currently being folded.
*   Added a check to reset `folded_directory_names` whenever the traversal encounters an entry that is not a child of the currently folded path.
*   Ensured state is cleared immediately after a folded directory is rendered.

**Release Notes:**
- Fixed an issue where using slash commands to collect files would sometimes display incorrect directory paths (e.g., showing `.github/.zed` instead of `.zed`) when adjacent directories were automatically folded.

### DIFF
--- a/crates/assistant_slash_commands/src/file_command.rs
+++ b/crates/assistant_slash_commands/src/file_command.rs
@@ -250,6 +250,7 @@ fn collect_files(
             let worktree_id = snapshot.id();
             let path_style = snapshot.path_style();
             let mut directory_stack: Vec<Arc<RelPath>> = Vec::new();
+            let mut folded_directory_path: Option<Arc<RelPath>> = None;
             let mut folded_directory_names: Arc<RelPath> = RelPath::empty().into();
             let mut is_top_level_directory = true;
 
@@ -277,6 +278,16 @@ fn collect_files(
                     )))?;
                 }
 
+                if let Some(folded_path) = &folded_directory_path {
+                    if !entry.path.starts_with(folded_path) {
+                        folded_directory_names = RelPath::empty().into();
+                        folded_directory_path = None;
+                        if directory_stack.is_empty() {
+                            is_top_level_directory = true;
+                        }
+                    }
+                }
+
                 let filename = entry.path.file_name().unwrap_or_default().to_string();
 
                 if entry.is_dir() {
@@ -292,13 +303,17 @@ fn collect_files(
                                 folded_directory_names =
                                     folded_directory_names.join(RelPath::unix(&filename).unwrap());
                             }
+                            folded_directory_path = Some(entry.path.clone());
                             continue;
                         }
                     } else {
                         // Skip empty directories
                         folded_directory_names = RelPath::empty().into();
+                        folded_directory_path = None;
                         continue;
                     }
+
+                    // Render the directory (either folded or normal)
                     if folded_directory_names.is_empty() {
                         let label = if is_top_level_directory {
                             is_top_level_directory = false;
@@ -334,6 +349,8 @@ fn collect_files(
                             },
                         )))?;
                         directory_stack.push(entry.path.clone());
+                        folded_directory_names = RelPath::empty().into();
+                        folded_directory_path = None;
                     }
                     events_tx.unbounded_send(Ok(SlashCommandEvent::Content(
                         SlashCommandContent::Text {
--- a/crates/assistant_slash_commands/src/file_command.rs
+++ b/crates/assistant_slash_commands/src/file_command.rs
@@ -226,10 +226,10 @@ fn collect_files(
     let Ok(matchers) = glob_inputs
         .iter()
         .map(|glob_input| {
-            custom_path_matcher::PathMatcher::new(&[glob_input.to_owned()])
+            util::paths::PathMatcher::new(&[glob_input.to_owned()], project.read(cx).path_style(cx))
                 .with_context(|| format!("invalid path {glob_input}"))
         })
-        .collect::<anyhow::Result<Vec<custom_path_matcher::PathMatcher>>>()
+        .collect::<anyhow::Result<Vec<util::paths::PathMatcher>>>()
     else {
         return futures::stream::once(async {
             anyhow::bail!("invalid path");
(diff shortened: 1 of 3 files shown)


crates/assistant_slash_commands/src/file_command.rs
```diff
@@ -289,8 +289,7 @@ fn collect_files(
                                 folded_directory_names =
                                     folded_directory_names.join(&path_including_worktree_name);
                             } else {
-                                folded_directory_names =
```
dinocosta, line 292: I believe we might need to revert this back to the previous code – `folded_directory_names = folded_directory_names.join(RelPath::unix(&filename).unwrap());` – otherwise it breaks items like `.github/workflows/ci.yml`, where it'll now only show `workflows/ci.yml` ▼

Image

crates/assistant_slash_commands/src/file_command.rs
```diff
@@ -301,19 +300,22 @@ fn collect_files(
...
                         let label = if is_top_level_directory {
-                            is_top_level_directory = false;
                             path_including_worktree_name.display(path_style).to_string()
                         } else {
-                            filename
+                            filename.clone()
                         };
+                        if is_top_level_directory {
+                            is_top_level_directory = false;
+                        }
```
dinocosta: I believe we can revert back to the original code here, that way we avoid checking `is_top_level_directory` twice and avoid the `filename.clone()` call 👍

crates/assistant_slash_commands/src/file_command.rs
```diff
@@ -301,19 +300,22 @@ fn collect_files(
...
                         events_tx.unbounded_send(Ok(SlashCommandEvent::StartSection {
                             icon: IconName::Folder,
                             label: label.clone().into(),
                             metadata: None,
                         }))?;
                         events_tx.unbounded_send(Ok(SlashCommandEvent::Content(
                             SlashCommandContent::Text {
-                                text: label.to_string(),
+                                // text: path_including_worktree_name.display(path_style).to_string(),
```
dinocosta: I believe we can remove this comment? 🤔

```suggestion
```

dinocosta (dismissed):

crates/assistant_slash_commands/src/file_command.rs
CLoaKY233: That's weird, 
- for me, it's working properly for this build,
image

- should this be platform specific?

crates/assistant_slash_commands/src/file_command.rs
CLoaKY233: Done! :D Sorry for sloppy code !

dinocosta:
Hey @CLoaKY233 ! Sorry that the review is taking longer than expected, I did not forge this PR but want to make sure that the glob normalization doesn't introduce any problems. Since the current changes are simply replacing `\` with `/` ▼

```rust
let sanitized = SanitizedPath::new(glob).to_string();
let normalized = sanitized.replace('\\', "/");
```

I'm worried that it might break, for example, in macOS, for a directory with a space, which could show up as something like `~/Developer/Awesome\ Project/*.txt` . With the current changes this would likely be converted to `~/Developer/Awesome/ Project/*.txt` so I'm exploring whether we need different code paths for Windows specifically 🙂

Veykril:
There likely is a `RelPath` issue here, we've gotten reports about a couple specifically wrt to the glob matcher. I'm looking into those right now.

dinocosta:
Thank you, @Veykril , help on the Windows side is much appreciated, as I don’t have access to a Windows development machine 🙂

Let me know if I can assist with anything macOS-related!

CLoaKY233:
Thanks for the thoughtful review and for investigating the Windows issues,
If there's anything specific I can help with (testing, logs, etc.), I'll be up for it.

franciskafyi:
@CLoaKY233 it looks like we fixed the underlying issue that this PR solves for. Can I go ahead and close this?

CLoaKY233:
> @CLoaKY233 it looks like we fixed the underlying issue that this PR solves for. Can I go ahead and close this?

Is the fix shipped in the latest version ? 
@franciskafyi 

```
Zed: v0.216.0+stable.94.f381f06bcd9fd7fae8c62c2308887edcc00f2d3d (Zed) 
Memory: 14.9 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics (TGL GT1) || Intel open-source Mesa driver || Mesa 25.2.3-1ubuntu1
```

I am still facing the issue 

image

.github being a foldable directory is changing the path of .zed 
where both directories are on the same levels not nested.

franciskafyi:
I saw Lukas assigned himself so he might have more context. It might not have hit Stable just yet.

Veykril:
Indeed this does not seem to be fixed entirely, I mustve missed something in testing.

CLoaKY233:
> Indeed this does not seem to be fixed entirely, I mustve missed something in testing.

I would love to work on this issue if possible, finding the issue and adding a test for this particular condition.

Veykril:
Can you check if the diff in https://github.com/zed-industries/zed/pull/44687 works fully? I am not too familiar with text threads, but from what I just checked this should fix things properly.

CLoaKY233:
> Can you check if the diff in #44687 works fully? I am not too familiar with text threads, but from what I just checked this should fix things properly.

the issue is actually in the directory folding logic (used to compact chains like `.github/workflows`),  fails to reset its state when traversing out of a folded branch. The `folded_directory_names` accumulator persists across iterations, causing subsequent sibling directories to incorrectly inherit the prefix of the previously folded path (e.g., `.zed` rendering as `.github/.zed` because `.github` was just processed).

Veykril:
Ah right, there are multiple issues at hand here are there. That folding one isn't windows specific I presume then right? (where as the listing one was?)

CLoaKY233:
> Ah right, there are multiple issues at hand here are there. That folding one isn't windows specific I presume then right? (where as the listing one was?)

I believe so, I could reproduce the folding directory issue on windows as well as linux, and the code is not platform specific. 
I've included the changes from #44687 in this pr and fixed the folding issue aswell, but when we see it, 


but also is this supposed to be like this ? 
image
files directly in the `root` directory do not have the `Project DIR` name in their path, but every other `directory` does

ie. it's `EMBRACE/.clang-format`, but just `src` for the directory.

Veykril:
I don't think its supposed to be like that, but thats a bug in the current release already, lets not mix that in here.

CLoaKY233:
The folding bug is now fixed—`.zed` no longer inherits the `.github` prefix when both are at the same level.

Should I add a test for this specific folding case? I can create one that verifies correct path display when folded directories (like `.github/workflows`) are adjacent to non-folded siblings (like `.zed`).

Regarding the root directory naming inconsistency you mentioned, I'll keep that out of this PR as it's a separate existing issue.

Veykril:
If you can come up with a test for that would be appreciated! I asked someone else internally to double check this change as I am not too familiar with text threads (my concern was the windows bug here)

Veykril:
Thanks!

PR #41884, opened 2025-11-04
Fix pinned tab causing resize loop on adjacent tab

### LINKED ISSUE #41467: Pinned tab causes resize loop on tab to its right
### Summary

Pinned tab causes resize loop on tab to its right

### Description

If I have a pinned tab (1st tab), then the tab immediately to its right (2nd tab) would be stuck in a resize loop / endless flickering. Seems to only happen when the 2nd tab is less than the max tab width or something like that.

Steps to reproduce:
1. Open first tab (e.g. `README.md`).
2. Pin first tab.
3. Open second tab (with a short filename, e.g. `.env`).

**Expected Behavior**: Second tab width is stable.
**Actual Behavior**: Second tab is stuck in a resize loop / constant flickering.

Zed: v0.209.7 (Zed) 
OS: Linux Wayland fedora-asahi-remix 42
Memory: 15.3 GiB
Architecture: aarch64
GPU: Apple M2 (G14G B0) || Honeykrisp || Mesa 25.2.4

abulgit:
Closes #41467 

My first PR in Zed, any guidance or tips are appreciated.

This fixes the flickering/resize loop that occurred on the tab immediately to the right of a pinned tab.

Removed the conditional border on the pinned tabs container. The border was a visual indicator to show when unpinned tabs were scrolled, but it wasn't essential and was causing the layout thrashing.

Release Notes:

- Fixed

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -3080,6 +3080,14 @@ impl Pane {
         }
         let unpinned_tabs = tab_items.split_off(self.pinned_tab_count);
         let pinned_tabs = tab_items;
+
+        // Avoid flickering when max_offset is very small (< 2px).
+        // The border adds 1-2px which can push max_offset back to 0, creating a loop.
+        let max_scroll = self.tab_bar_scroll_handle.max_offset().width;
+        let current_offset = self.tab_bar_scroll_handle.offset().x;
+        let is_scrollable = max_scroll > px(2.0);
+        let is_scrolled = current_offset < px(0.);
+
         TabBar::new("tab_bar")
             .when(
                 self.display_nav_history_buttons.unwrap_or_default(),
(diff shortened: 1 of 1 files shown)


smitbarmase (changes requested):
I don't think this is the right way to handle this. This code is essential to add border when there are large amount of unpinned tabs and are scrolled.

Since this PR results in something like this:

Image

I'm not able to reproduce this issue, but can you inspect `is_scrollable` and `is_scrolled` bool and figure which one of them is flipping on alternate renders?

We might have to go deeper in checking if `tab_bar_scroll_handle.offset().x` is flipping to `0` every once in a while or is it`tab_bar_scroll_handle.max_offset().width`.

abulgit:
Thank you so much @smitbarmase  for guiding me to tackle this issue. 
So I added some debug logging to see what was happening. Turns out both values were flipping, but the root cause was `max_offset` toggling between `0px` and `0.8px`.

Here's my observation from logs

When no border: `max_offset=0.8px` -> `is_scrollable=true` -> border gets added
When border added: available space shrinks -> `max_offset=0px` -> `is_scrollable=false` -> border removed ->
and it repeats forever

The 0.8px difference is smaller than the border width, so it creates a loop.

So I changed the scrollability check from `!max_scroll.is_zero()` to `max_scroll > px(2.0)`. This ignores tiny subpixel differences while still showing the border when tabs are actually scrollable. and it works!

smitbarmase (approved):
Great, that’s a nice fix! I moved the code back to `children` since we should only run all these comparisons when necessary. Thanks for working on this.

PR #41885, opened 2025-11-04
gpui: Small tab group performance improvements

tidely:
Closes #ISSUE

Removes a few eager container clones and iterations.

Added a todo to `get_prev_tab_group_window` and `get_next_tab_group_window`. They seem to use `HashMap::keys()` for choosing the previous tab group, however `.keys()` returns an arbitrary order, so I'm not sure if previous actually means anything here. Conrad seems to have worked on this part previously, maybe he has some insights. That can possibly be a follow-up PR, but I'd be willing to work on it here as well since the other changes are so simple.

Release Notes:

### DIFF
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -484,16 +484,14 @@ impl SystemWindowTabController {
             return;
         };
 
+        let initial_tabs_len = initial_tabs.len();
         let mut all_tabs = initial_tabs.clone();
-        for tabs in controller.tab_groups.values() {
-            all_tabs.extend(
-                tabs.iter()
-                    .filter(|tab| !initial_tabs.contains(tab))
-                    .cloned(),
-            );
+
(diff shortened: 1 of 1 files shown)


JosephTLyons:
Thanks for the PR @tidely, we'll check it out.

crates/gpui/src/app.rs
```diff
@@ -341,12 +343,9 @@ impl SystemWindowTabController {
 
     /// Get all tabs in the same window.
     pub fn tabs(&self, id: WindowId) -> Option<&Vec<SystemWindowTab>> {
```
tidely, line 345: This has become such a small helper that I'd consider removing it entirely. It's only called a few times and the majority of the time, we use `.position()` right after to find the index of tab with the right `WindowId`. Which could be done right away instead of using the helper first, thus removing an iteration through the values.

PR #41890, opened 2025-11-04
Tidy up `Edit in json` footer entries

SomeoneToIgnore:
Before:
before

After:
after

* All items got a keybinding label
* All items were made of a non-small size, to match the text size in the right button
* Keybindings are rendered as disabled for disabled buttons

Release Notes:

- N/A


danilo-leal:
Nice!!!

PR #41899, opened 2025-11-04
Add vim/emacs modeline support

### LINKED ISSUE #4762: Support modeline comments for per-file settings
### Check for existing issues

- [X] Completed

### Describe the feature

Many editors such as vim and emacs support "modelines", a comment at the beginning of the file that allows the file type to be explicitly specified along with per-file specific settings.  For example, a Makefile might have the following line at the top of the file:

```make
# -*- mode: Makefile; tab_size: 8; hard_tabs: true -*-
```

Which sets the file type to Makefile and sets the `tab_size` to 8 and uses hard-tabs for this particular file.  It might be nice it if also recognized common emacs/vim settings such as `tab-width: 8` and `indent-tabs-mode: t` vs `tab_size` and `hard_tabs` although the values are not too consistent across editors -- simply supporting Zed settings would be a great start.

_No response_

elmarco:
Many editors such as vim and emacs support "modelines", a comment at the beginning of the file that allows the file type to be explicitly specified along with per-file specific settings

- The amount of configurations, style and settings mapping cannot be handled in one go, so this opens up a lot of potential improvements.
- I left out the possiblity to have "zed" specific modelines for now, but this could be potentially interesting.
- Mapping the mode or filetype to zed language names isn't obvious either. We may want to make it configurable.

This is my first contribution to zed, be kind. I struggled a bit to find the right place to add those settings. I use a similar approach as done with editorconfig (merge_with_editorconfig). There might be better ways.

Closes #4762

Release Notes:

- Add basic emacs/vim modeline support.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1166,6 +1166,12 @@
   "hard_tabs": false,
   // How many columns a tab should occupy.
   "tab_size": 4,
+  // Number of lines to search for modelines at the beginning and end of files.
+  // Modelines contain editor directives (e.g., vim/emacs settings) that configure
+  // the editor behavior for specific files.
+  //
+  // A value of 0 disables modelines support.
+  "modeline_lines": 5,
   // What debuggers are preferred by default for all languages.
   "debuggers": [],
   // Control what info is collected by Zed.
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -2094,6 +2094,7 @@ impl AcpThread {
 
                     let settings = language::language_settings::language_settings(
                         buffer.language().map(|l| l.name()),
+                        buffer.modeline(),
                         buffer.file(),
                         cx,
                     );
--- a/crates/acp_thread/src/diff.rs
+++ b/crates/acp_thread/src/diff.rs
@@ -373,6 +373,7 @@ async fn build_buffer_diff(
                 old_text_rope,
                 buffer.language().cloned(),
                 language_registry,
+                buffer.modeline().cloned(),
                 cx,
             )
         })?
--- a/crates/agent/src/tools/edit_file_tool.rs
+++ b/crates/agent/src/tools/edit_file_tool.rs
@@ -387,6 +387,7 @@ impl AgentTool for EditFileTool {
                 .read_with(cx, |buffer, cx| {
                     let settings = language_settings::language_settings(
                         buffer.language().map(|l| l.name()),
+                        buffer.modeline(),
                         buffer.file(),
                         cx,
                     );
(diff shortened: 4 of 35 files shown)


ConradIrwin:
Amazing, thank you for this! I think the overall structure is reasonable, and I like that we're doing this as part of detect language.

Looking through the code I have a few suggestions:
* Let's hard limit the number of bytes to search for modeline to ~1k, we've had persistent problems with performance of files with extremely long lines (we could maybe do the emacs thing and make it ~3k, but seems a bit excessive). If we do that then I don't think we need the setting for number of lines.
* We should store the ModelineSettings as an `Rc` in the Buffer and BufferSnapshot to make it cheaper to create a new snapshot (another area of performance sensitivity for us).
* We need some docs! Maybe a new page describing what variables can be set that's linked to from the existing settings pages?
* Let's (probably) not ship the "emacs locals" syntax. I've seen modelines before, but never that, and don't want to maintain a third parser for this stuff.
 * I'd also like to tighten up the syntax for vim. Not sure what's actually used in the wild but in my experience it's always "<comment indicator> vim?:`. I don't think we need to support `ex:`, nor doing this on lines with no comments (and particularly not lines with other non-modeling content before the modeline!).

Thanks again, and let me know if you want help with any of these: https://cal.com/conradirwin/pairing

ConradIrwin:
(Also feel free to edit typos.toml to fix the style check)

elmarco:
Hi @ConradIrwin ,
- I thought the file was already preloaded, and thus it would not be a big deal to limit by number of lines, which is a bit easier to configure or tweak than a fixed number of Kb. Or do you think that feature shouldn't be configurable? 
- Ok to wrap with Rc, yes
- Ok for the docs, good idea
- The emacs "Local Variables:" is a bit esoteric, but is much easier than parsing vim modelines (and thus easy to maintain)! A quick search in github reveal that many projects use them (https://github.com/search?type=code&q=%22Local+Variables%3A%22, in particular GNU projects: glibc, bison, guile, emacs etc). Nevertheless, I can drop support for it.
- I am afraid that the vim syntax diversity is also very wide. And there are many valid cases where there are extra lines without vim modeline: shebang, and other modelines when using both emacs and vim for example, which is very common. 

btw, I wonder if you would rather use it as a third-party crate? we could even make it compile-time optional? wdyt

It would be great to experience pairing, I haven't tried it and I am sure I could learn a ton from this experience. Let see if we can find some time for it.

ConradIrwin:
The file will be loaded, but line 1 could be 2Gb long (think a JSON export); which you probably don't want to run a regex on on the main thread (particularly if the mode-line regexes are not anchored to the start). Limiting to a few kb prevents that being slower than it is for more normal files which avoids surprisingly slowdowns. If we limit by kb, I don't think we need the "number of lines" check either.

I don't think we need to make the number of bytes configurable, as long as it's easy to explain (e.g. emacs' 3k limit).

I'd rather have the code in the Zed repo than as a crate because it's easier to change and test it.

https://cal.com/conradirwin/pairing has some times when I'm free for pairing.

ConradIrwin:
Closing through lack of response, but happy to review new versions of this

PR #41900, opened 2025-11-04
git: Use correct file mode when staging

### LINKED ISSUE #28667: Git: staging ranges in a file with special permissions is not handled properly
### Summary

TL;DR If you stage a range in a file with permissions `755` using zed, the `git diff --staged` will have the permissions reset to `644`

### Description

Steps to trigger the problem:
- say you have a file with permissions which has **lines** changed in it
- you stage that line in zed
- if you `git diff --staged`, this is what you see (in effect, zed hardcodes the permission to 644 in the git diff no matter the real permissions):
```diff
diff --git a/foo b/foo
old mode 100755
new mode 100644
index b622386a1..b6889da66
--- a/foo
+++ b/foo
-bar
+baz
```
- this doesn't mean the file itself has had its permissions changed on disk, or that the next `git commit` will, since this is what `git diff -- foo` looks like:
```diff
diff --git a/foo b/foo
old mode 100644
new mode 100755
index b6889da66..e929420a4
--- a/foo
+++ b/foo
-bar
+baz
```

Actual Behavior:
- the real file permissions got "left behind", replaced by the default, `644`

Expected Behavior:
- that this does not happen

Zed: v0.181.5 (Zed) 
Memory: 61.9 GiB
Architecture: x86_64

11happy:
Closes #28667

Release Notes:

- I have implemented file mode detection using `new_smol_command` & if its empty defaulting to 100644 as previously, however I noticed in function `load_committed_text` 
```
fn load_committed_text(&self, path: RepoPath) -> BoxFuture<'_, Option<String>> {
        let repo = self.repository.clone();
        self.executor
            .spawn(async move {
                let repo = repo.lock();
                let head = repo.head().ok()?.peel_to_tree().log_err()?;
                let entry = head.get_path(path.as_std_path()).ok()?;
                if entry.filemode() == i32::from(git2::FileMode::Link) {
                    return None;
                }
                let content = repo.find_blob(entry.id()).log_err()?.content().to_owned();
                String::from_utf8(content).ok()
            })
            .boxed()
    }
 ```
it accesses filemode with entry.filemode() should I use this approach ?

### CODE REVIEW

**Comment by cole-miller** [comment]
@11happy Yeah, I think using the git2 API to load the filemode like in `load_committed_text` makes sense.

**Comment by 11happy** [comment]
@cole-miller , I have refactored the approach keeping it similar to `load_commited_text` and flattened it.

**Comment by cole-miller** [comment]
Thanks, unfortunately I realized I made a mistake and we need to do this a bit differently--we want to take the mode from the working copy of the file, not from HEAD, so that this works when staging the hunk for a file that doesn't exist in HEAD. I think to fix this we will need to do the following:

- Add an `is_executable` boolean to our `fs::Metadata` type (and load this data correctly on all platforms)
- Have `spawn_set_index_text_job` do an `fs.metadata()` call for the abs path it's given, and pass `is_executable` down to `set_index_text`
- In `set_index_text`, use `100755` as the mode if `is_executable` is true, and `100644` instead.

Luckily Git has a [very restricted](https://git-scm.com/book/sv/v2/Git-Internals-Git-Objects) set of modes so this should be all that's needed.

**Comment by 11happy** [comment]
Thanks @ Cole for such detailed directions & also pointing the git internals for objects it was a great read : )

**Comment by esthertrapadoux** [comment]
Hey @11happy just checking in and seeing how things are going. We still have pairing sessions open if you need one: https://cal.com/esther-trapadoux-zed/30min

Also totally understand that life happens, so if now's not a great time to work on this issue, just let us know. We'll unassign and free it up for someone else to work on it.

**Comment by 11happy** [comment]
> t life happens, so if now's not a great time to work on this issue, just let us know. We'll unassign and free it up for someone else to work on

as per last discussion I am looking up some things but not yet figured it out, I think I might need some help but before requesting for another pairing session I will make sure to do my end of research well.

Thank you : )

**Comment by 11happy** [comment]
> Hey @11happy just checking in and seeing how things are going. We still have pairing sessions open if you need one: https://cal.com/esther-trapadoux-zed/30min
> 
> Also totally understand that life happens, so if now's not a great time to work on this issue, just let us know. We'll unassign and free it up for someone else to work on it.

I have requested for another pairing sessions,
Thank you

**Comment by 11happy** [comment]
Thank you for having patience with this PR : ) , i have updated it accordingly & looking forward to discuss in today's paring session.

File: crates/fs/src/fs.rs [hunk 1]
```diff
@@ -763,13 +764,25 @@ impl Fs for RealFs {
...
+        let is_executable = {
+            use std::os::unix::fs::PermissionsExt;
+            metadata.permissions().mode() & 0o111 != 0
+        };
+
+        #[cfg(windows)]
+        let is_executable = path
+            .extension()
+            .is_some_and(|ext| matches!(ext.to_str(), Some("exe" | "bat")));
+
```
kubkon: I think it would be safer to use [`is_executable`](https://crates.io/crates/is_executable) crate to handle Windows. We could then also make this a target-specific dependency in `Cargo.toml`.

crates/git/src/repository.rs
```diff
@@ -979,12 +1008,15 @@ impl GitRepository for RealGitRepository {
...
         content: Option<String>,
         env: Arc<HashMap<String, String>>,
+        is_executable: bool,
     ) -> BoxFuture<'_, anyhow::Result<()>> {
         let working_directory = self.working_directory();
         let git_binary_path = self.any_git_binary_path.clone();
         self.executor
             .spawn(async move {
                 let working_directory = working_directory?;
+                let mode = get_file_mode(&working_directory, &path, is_executable)?;
```
kubkon: I think it suffices to just use

```rust
let mode = if is_executable { "100755" } else { "100644" };
```

In that case `get_file_mode` is redundant I think.

crates/project/src/git_store.rs
```diff
@@ -4351,6 +4351,8 @@ impl Repository {
         let id = self.id;
         let this = cx.weak_entity();
         let git_store = self.git_store.clone();
+        let abs_path = self.snapshot.repo_path_to_abs_path(&path);
+        let fs = fs::RealFs::new(None, cx.background_executor().clone());
```
kubkon: Instead of creating a `RealFs` handle here, let's augment `RepositoryState::Local` with `fs: Arc<dyn Fs>` like so

```rust
enum RepositoryState {
  Local {
    fs: Arc<dyn Fs>,
    ...
  },
  Remote { ... }
}
```

Then we will have access to it within every enqueued git job. This will also mean we will not break tests and the remote project handling.

crates/project/src/git_store.rs
```diff
@@ -4359,14 +4361,20 @@ impl Repository {
                     "start updating index text for buffer {}",
                     path.as_unix_str()
                 );
+
+                let executable = match fs.metadata(&abs_path).await {
+                    Ok(Some(meta)) => meta.is_executable,
+                    Ok(None) => false,
+                    Err(_err) => false,
+                };
```
kubkon: When the above is addressed, we should move it inside `RepositoryState::Local` prong.

crates/git/src/repository.rs
```diff
@@ -116,6 +116,34 @@ pub fn parse_worktrees_from_str<T: AsRef<str>>(raw_worktrees: T) -> Vec<Worktree
...
+                    let mode = format!("{:o}", entry.filemode() as u32);
+                    return Ok(mode);
+                }
+            }
+        }
+    }
+    let mode = if is_executable { "100755" } else { "100644" }; // currently kept for earlier suggested method as fallback
+    Ok(mode.to_string())
+}
+
```
kubkon: I think this is redundant.

```suggestion
```

kubkon (changes requested):
Looks great! I've left some comments that should make it almost mergeable. Also, it'd be great if you could add some tests in `project_tests.rs`.

crates/git/src/repository.rs
```diff
@@ -987,12 +988,15 @@ impl GitRepository for RealGitRepository {
...
         content: Option<String>,
         env: Arc<HashMap<String, String>>,
+        is_executable: bool,
     ) -> BoxFuture<'_, anyhow::Result<()>> {
         let working_directory = self.working_directory();
         let git_binary_path = self.any_git_binary_path.clone();
         self.executor
             .spawn(async move {
                 let working_directory = working_directory?;
+                let mode = if is_executable { "100755" } else { "100644" }.to_string();
```
kubkon:
```suggestion
                let mode = if is_executable { "100755" } else { "100644" };
```

crates/git/src/repository.rs
```diff
@@ -1013,7 +1017,7 @@ impl GitRepository for RealGitRepository {
                     let output = new_smol_command(&git_binary_path)
                         .current_dir(&working_directory)
                         .envs(env.iter())
-                        .args(["update-index", "--add", "--cacheinfo", "100644", sha])
+                        .args(["update-index", "--add", "--cacheinfo", &mode, sha])
```
kubkon:
```suggestion
                        .args(["update-index", "--add", "--cacheinfo", mode, sha])
```

crates/git/src/repository.rs
kubkon: I don't think we need to allocate here.

crates/project/src/git_store.rs
```diff
@@ -5164,6 +5174,7 @@ impl Repository {
         cx: &mut Context<Self>,
     ) -> mpsc::UnboundedSender<GitJob> {
         let (job_tx, mut job_rx) = mpsc::unbounded::<GitJob>();
+        let fs_clone = fs.clone();
```
kubkon:
```suggestion
        let fs_cloned = fs.clone();
```

Just a nit but I reckon this reads better.

crates/project/src/git_store.rs
```diff
@@ -5195,8 +5206,8 @@ impl Repository {
                     backend.clone(),
                 );
             }
-
             let state = RepositoryState::Local {
+                fs: fs_clone,
```
kubkon:
```suggestion
                fs: fs_cloned,
```

crates/fs/src/fs.rs
```diff
@@ -895,13 +897,16 @@ impl Fs for RealFs {
         #[cfg(unix)]
         let is_fifo = metadata.file_type().is_fifo();
 
+        let is_executable = path.is_executable();
```
kubkon: Since this is I/O bound, we'd like to schedule that on the bg thread:

```suggestion
        let is_executable = self.executor.spawn(async move { path.is_executable() }).await.with_context(|| "checking if file is executable")?;
```

kubkon (changes requested):
Looks great, just a couple more nits and I think we should be in a mergeable state!

crates/git/src/repository.rs
11happy: done

crates/git/src/repository.rs
11happy: done

crates/project/src/git_store.rs
11happy: done

crates/fs/src/fs.rs
11happy: done

11happy:
I have added the nits & added a test as well
Thank you @kubkon for guiding this PR : )

Also the current test is for unix & have two asserts making sure its working as expected , however I think one assert would be fine ?

kubkon (approved):
Excellent work, thank you!

esthertrapadoux:
Amazing work @11happy 🥇

PR #41915, opened 2025-11-04
git: Make permalinks aware of current diffs

### LINKED ISSUE #22546: Copy Permalink (GitHub, etc) should better handle modified code
### Check for existing issues

- [X] Completed

Steps to reproduce:
1. Edit a file in a git repo add a bunch of comment to an existing file
2. Right click and select `Copy Permalink` to a line later in the file.

Notice that the linked line number is offset from the code you were trying to link. 

For example, given the following committed code:

https://github.com/zed-industries/zed/blob/8e12c679fc5a6c4ff6389c7a6a98a9878e350e36/crates/util/src/util.rs#L129-L132

And the following buffer state: 
Screenshot 2025-01-01 at 12 12 27

Copy Permalink gives me this link (line 132):
https://github.com/zed-industries/zed/blob/8e12c679fc5a6c4ff6389c7a6a98a9878e350e36/crates/util/src/util.rs#L132
But this is what I expected (line 129):
https://github.com/zed-industries/zed/blob/8e12c679fc5a6c4ff6389c7a6a98a9878e350e36/crates/util/src/util.rs#L129

I'm not positive the best behavior, but I think Zed can do better:
- For unchanged/pre-existing lines: line number should be to the line number of that code in the commit.
- For selections/lines which start or end on a newly inserted line (not present in the commit) Zed could expand the selection up/down so that's it's anchored to lines which actually exist in the commit.

Zed: v0.167.1 (Zed Preview)
OS: macOS 14.7.0
Memory: 64 GiB
Architecture: aarch64

### If applicable, add screenshots or screencasts of the incorrect state / behavior

N/A

N/A

siame:
Addressing #22546, we want git permalinks to be aware of the current changes within the buffer.

This change calculates how many lines have been added/deleted between the start and end of the selection and uses those values to offset the selection.

This is done within `Editor::get_permalink_to_line` so that it can be passed to any git_store.

Example:

image

Where this selections permalink would previously return L3-L9, it now returns L2-L7.

Release Notes:

- git: make permalinks aware of current diffs

Closes #22546

---

This is my first PR into the zed repository so very happy for any feedback on how I've implemented this. Thanks!

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -20619,57 +20619,21 @@ impl Editor {
                 buffer_ranges.last()
             }?;
 
+            let start_row_in_buffer = text::ToPoint::to_point(&range.start, buffer).row;
+            let end_row_in_buffer = text::ToPoint::to_point(&range.end, buffer).row;
+
             let Some(buffer_diff) = multi_buffer.diff_for(buffer.remote_id()) else {
-                let selection = text::ToPoint::to_point(&range.start, buffer).row
-                    ..text::ToPoint::to_point(&range.end, buffer).row;
+                let selection = start_row_in_buffer..end_row_in_buffer;
 
                 return Some((multi_buffer.buffer(buffer.remote_id()).unwrap(), selection));
             };
 
             let buffer_diff_snapshot = buffer_diff.read(cx).snapshot(cx);
 
-            let beginning_anchor = buffer.anchor_before(Point::new(0, 0));
-            let start_anchor = buffer.anchor_before(&range.start);
-            let end_anchor = buffer.anchor_before(&range.end);
-
-            struct Offsets {
-                start: isize,
-                end: isize,
-            }
-
-            let mut delta = Offsets { start: 0, end: 0 };
-
-            for hunk in
-                buffer_diff_snapshot.hunks_intersecting_range(beginning_anchor..end_anchor, buffer)
-            {
-                let initial_size = hunk.diff_base_byte_range.end as isize
-                    - hunk.diff_base_byte_range.start as isize;
-
-                let buffer_range = hunk.buffer_range.to_offset(&buffer);
-                let changed_size =
-                    (cmp::min(buffer_range.end, range.end) as isize) - buffer_range.start as isize;
-
-                if hunk.buffer_range.start.cmp(&start_anchor, buffer).is_le() {
-                    delta.start += initial_size;
-                    delta.start -= changed_size;
-                }
-
-                delta.end += initial_size;
-                delta.end -= changed_size;
-            }
-
-            let start_row_in_base_buffer = buffer_diff_snapshot
-                .base_text()
-                .offset_to_point((range.start as isize + delta.start) as usize)
-                .row;
-            let end_row_in_base_buffer = buffer_diff_snapshot
-                .base_text()
-                .offset_to_point((range.end as isize + delta.end) as usize)
-                .row;
-
             Some((
                 multi_buffer.buffer(buffer.remote_id()).unwrap(),
-                start_row_in_base_buffer..end_row_in_base_buffer,
+                buffer_diff_snapshot.row_to_base_text_row(start_row_in_buffer, buffer)
+                    ..buffer_diff_snapshot.row_to_base_text_row(end_row_in_buffer, buffer),
             ))
         });
 
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -81,7 +81,7 @@ use ::git::{
 use aho_corasick::{AhoCorasick, AhoCorasickBuilder, BuildError};
 use anyhow::{Context as _, Result, anyhow};
 use blink_manager::BlinkManager;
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks for tackling this! From reading the code, there are a couple of things that need to change:

- `copy permalink to line` supports multibuffers, but right now the code to adjust the line range for diff hunks won't work in that situation--it'll iterate over diff hunks in the multibuffer, which may come from buffers unrelated to the one where the selection is.
- The handling of deleted and modified hunks is not right. For these hunks, the adjustment to make depends on whether the hunks are expanded or not (because expanding and collapsing hunks affects the `row_range` of `MultiBufferDiffHunk`, but it shouldn't affect the line range we use for the permalink)

Here is an approach to this feature that I think should avoid these problems:

- We can still call `range_to_buffer_ranges` and take the first/last range. This gives us a selection in terms of `usize` offsets in a single `BufferSnapshot`, and from here on we shouldn't need to use the original `MultiBufferSnapshot` at all.
- Then we need to perform the diff hunk adjustment on that `usize` buffer range. I think we can do this efficiently (i.e. without iterating over all the diff hunks before the selection) by making some minor changes to the code in the `buffer_diff` crate and using a sum tree `Cursor`.

This is probably easier to work through live, would you like to pair on it? Link to book a time (I'll be covering the time slots on this Friday): https://cal.com/esther-trapadoux-zed/30min

siame:
Thanks for the comments @cole-miller, I've booked in some time to pair on this, looking forward to it.

siame:
@cole-miller I've pushed up changes to recalculate this using offsets rather than with row ranges.

I've not got to the `buffer_diff` `Cursor` changes yet, but I'm currently having an issue with these calculations which are causing the `end_row_in_base_buffer` to be off by one. I'm able to fix by changing the calculation to:

```
let end_row_in_base_buffer = buffer_diff_snapshot
    .base_text()
    // off by one
    .offset_to_point((range.end as isize + delta.end - 1) as usize) // additional -1 here
    .row;
```

I thought this was to do with the `range.end` maybe being a `Bias::Right` `Anchor` grabbing one too many characters in the selection, but it's an `Offset`, so that's not quite it.

Would appreciate any insight into any issues you can see here.

Thanks!

cole-miller:
@siame Thanks for the progress report! I haven't had the chance to run the code and investigate, but if the start value is what you expect and the end value isn't, and you're testing this a selection that covers entire lines, it might just be that the selection includes the trailing newline of the last line? You could confirm by looking at whether the end of the selection looks like `Point::new(row + 1, 0)` where `row` is the row you're expecting to see.

siame:
Thanks for the input @cole-miller, the ranges of the selection are as expected, but I think I've got to the bottom of the issue.

In calculating the delta, we're adding the initial size and subtracting the changed size. I added a `cmp::min` to cap the `changed_size` to the end of the selection:

```
let changed_size =
    (cmp::min(buffer_range.end, range.end) as isize) - buffer_range.start as isize;
```

This is so that we only consider the diff until the end of the selection, as any more than that would be unnecessary.

However, we're not doing the same for the `initial_size`, as this is a `Range<Offset>`, and doesn't map to the current `BufferSnapshot` the same way:

```
let initial_size = hunk.diff_base_byte_range.end as isize
    - hunk.diff_base_byte_range.start as isize;
```

There is the function `hunk.diff_base_byte_range.to_anchors`, however this takes a `MultiBufferSnapshot`, so I'm not sure if these `Anchors` will be comparable to the current Selection's.

Without scaling the `diff_base_byte_range` the same way we're scaling the `buffer_range`, I'm unable to get an accurate line count when calculating the delta. Is there any way to map the initial size to anchors for the `BufferSnapshot` we're using? Then we can should be able to calculate the delta accordingly. Thanks!

cole-miller:
Hey @siame--thanks for the update. It seems like it might be a bit more natural to do this using a `SumTree` cursor like we talked about--would you like to pair on that sometime this week? https://cal.com/cole-miller-zed/30min

siame:
@cole-miller I've updated the logic to now use a `SumTree` to calculate the effective offsets, continuing on where we got to in the pairing session. I've added a test to check this functionality too, let me know how this looks, thanks!

siame:
@cole-miller It looks like the work in this PR has been superceded/replicated by the changes in https://github.com/zed-industries/zed/commit/2e00f40c54e1e0c23af06587cefcfd06eae49a0e, the only unique work to this PR being the changes to `Editor::get_permalink_to_line`, should I re-work this to use the current `main` changes instead?

esthertrapadoux:
@siame just following up to let you know the team went pretty deep in more git feature work and we got behind on PR reviews. We're ramping back up and appreciate the patience! 👍

siame:
@cole-miller I've updated this such that it uses the new `row_to_base_row` in order to calculate the permalink rows. The total changes in this PR have now whittled down quite a lot, so hopefully this should be granular enough for a quick review. Let me know if there are any questions :)

cole-miller:
Thanks!

PR #41917, opened 2025-11-04
file_finder: Display duplicated file in file finder history

### LINKED ISSUE #41850: Duplicated file not part of file finder history
### Summary

When I duplicate a file it automatically open it but it's not added into file finder history.

Steps to reproduce:
1. Open a project
2. Right click on a file and click on duplicate
3. Zed automatically open that file
4. Go back in another file
5. Open file finder and see the file doesn't appear as part of history

**Expected Behavior**:
The duplicated file is part of file finder history
**Actual Behavior**:
The duplicated file doesn't appear in file finder first entries (part of history)

Video:

https://github.com/user-attachments/assets/42606d6c-8ff2-486d-a0c2-808aa9cd1f97

Zed: v0.210.4 (Zed) 
OS: macOS 15.7.1
Memory: 36 GiB
Architecture: aarch64

bnjjj:
Closes #41850

When digging into this I figured out that basically what was going on is in the history of the file finder it doesn't update the name of the file duplicated because when you duplicate a file it's named automatically with `filename copy` and so this filename was added to the history but not updated so once you wanted to go back into this file it was not part of file finder displayed history anymore because this file doesn't exist anymore but the entity id remains the same.
I was also to reproduce this bug when just renaming a file.

Release Notes:

- Fixed: Display duplicated file in file finder history

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1905,7 +1905,7 @@ impl Editor {
                         }
                     }
 
-                    project::Event::EntryRenamed(transaction) => {
+                    project::Event::EntryRenamed(transaction, project_path, abs_path) => {
                         let Some(workspace) = editor.workspace() else {
                             return;
                         };
@@ -1913,7 +1913,23 @@ impl Editor {
                         else {
                             return;
                         };
+
                         if active_editor.entity_id() == cx.entity_id() {
+                            let entity_id = cx.entity_id();
+                            workspace.update(cx, |this, cx| {
+                                this.panes_mut()
+                                    .iter_mut()
+                                    .filter(|pane| pane.entity_id() != entity_id)
+                                    .for_each(|p| {
+                                        p.update(cx, |pane, _| {
+                                            pane.nav_history_mut().rename_item(
+                                                entity_id,
+                                                project_path.clone(),
+                                                abs_path.clone().into(),
+                                            );
+                                        })
+                                    });
+                            });
                             let edited_buffers_already_open = {
                                 let other_editors: Vec<Entity<Editor>> = workspace
                                     .read(cx)
@@ -1936,7 +1952,6 @@ impl Editor {
                                     })
                                 })
                             };
-
                             if !edited_buffers_already_open {
                                 let workspace = workspace.downgrade();
                                 let transaction = transaction.clone();
--- a/crates/project_panel/src/project_panel_tests.rs
+++ b/crates/project_panel/src/project_panel_tests.rs
@@ -2007,7 +2007,7 @@ async fn test_create_duplicate_items(cx: &mut gpui::TestAppContext) {
         "/src",
         json!({
             "test": {
-                "first.rs": "// First Rust file",
+                "first.txt": "// First Txt file",
                 "second.rs": "// Second Rust file",
                 "third.rs": "// Third Rust file",
             }
@@ -2106,7 +2106,7 @@ async fn test_create_duplicate_items(cx: &mut gpui::TestAppContext) {
             "v src",
             "    v test",
             "          [EDITOR: '']  <== selected",
-            "          first.rs",
+            "          first.txt",
             "          second.rs",
             "          third.rs"
         ]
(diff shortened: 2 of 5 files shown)


ConradIrwin:
This looks good. Need to figure out the windows test (happy to pair if you need: https://cal.com/conradirwin/pairing).

Would also be nice to fix rename (but that can be a separate PR)

bnjjj:
Hey @ConradIrwin Thanks for the review. I tried to figure out what's going on but I think I won't be able to really figure out until I have a machine running on Windows. I think what's going on is either because it contains a whitespace because of the default suffix ` copy` added when duplicating a file OR because it starts with `/src` it fails to convert the uri from `lsp-types` crate [here](https://github.com/zed-industries/lsp-types/blob/b71ab4eeb27d9758be8092020a46fe33fbca4e33/src/uri.rs#L75). I also saw we currently ignore a test for windows in the same file [here](https://github.com/zed-industries/zed/blob/feed34cafeae0b16e7a18cae25452b60c6387ef2/crates/project_panel/src/project_panel_tests.rs#L2993) (and if you read the related comment it might be the same issue because this panic only happens when we open a file) so I duplicated the test to keep the original one running on every platforms and create a new one to check history and really duplicate a file but it's ignored on windows now.

I tried to disable lsp using the `StopLanguageServer` command but it didn't work. I also tried to rename my files with `.txt` to not trigger any LSP but without success. I think it happens probably before running an actual LSP.

I'm mainly concerned about the behavior on windows. I'm wondering if `duplicate` or `rename` really works on Windows but I guess that's probably a different topic I can't tackle.

Another point. I'm curious to know what scenario you tried for rename because I tried to reproduce it with my version and it works, unlike the version on `main`. So I think I also fixed it. I added the exact same test than duplicate but for rename and it also works. When I test manually using the editor it also works so I'm curious to know what I'm missing.

crates/editor/src/editor.rs
```diff
@@ -1905,15 +1905,31 @@ impl Editor {
...
                         else {
                             return;
                         };
+
                         if active_editor.entity_id() == cx.entity_id() {
+                            let entity_id = cx.entity_id();
+                            workspace.update(cx, |this, cx| {
+                                this.panes_mut()
+                                    .iter_mut()
+                                    .filter(|pane| pane.entity_id() != entity_id)
```
bnjjj, line 1922: I'm not entirely sure about that part, I just used the same filtering than above but I don't really understand that part TBH

bnjjj:
Correction: I thought it was because of the whitespace but apparently it's not because I was able to trigger that panic with the rename test [here](https://github.com/zed-industries/zed/actions/runs/19115787444/job/54624376084?pr=41917#step:8:3227) so I have no idea why it's failing. Looking at the [crate docs ](https://docs.rs/url/latest/url/struct.Url.html#method.from_file_path) it could fail if `This returns Err if the given path is not absolute or, on Windows, if the prefix is not a disk prefix (e.g. C:) or a UNC prefix (\\).` so I don't really know.

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -2199,6 +2199,396 @@ async fn test_create_duplicate_items(cx: &mut gpui::TestAppContext) {
...
+// when it triggers the lsp store it converts `/src/test/first copy.txt` into an uri
+// but it fails with message `"/src\\test\\first copy.txt" is not parseable as an URI`
+#[gpui::test]
+#[cfg_attr(target_os = "windows", ignore)]
+async fn test_create_duplicate_items_and_check_history(cx: &mut gpui::TestAppContext) {
+    init_test_with_editor(cx);
+
+    let fs = FakeFs::new(cx.executor());
+    fs.insert_tree(
+        "/src",
```
Veykril, line 2212: You want to use the `path!` macro here
```suggestion
        path!("/src"),
```
to make this work for windows

ConradIrwin:
@Veykril Looks like there's more to the windows test than meets the eye. Merging with it disabled for now.

PR #41954, opened 2025-11-05
Add ACP terminal-login via _meta field

rtfeldman:
As discussed with @benbrandt and @mikayla-maki:

* We now tell ACP clients we support the nonstandard `terminal-auth` `_meta` field for terminal-based authentication
* In the future, we anticipate ACP itself supporting *some* form of terminal-based authentication, but that hasn't been designed yet or gone through the RFD process
* For now, this unblocks terminal-based auth

Release Notes:

- Added experimental terminal-based authentication to ACP support

### DIFF
--- a/crates/agent_servers/src/acp.rs
+++ b/crates/agent_servers/src/acp.rs
@@ -178,6 +178,7 @@ impl AcpConnection {
                     meta: Some(serde_json::json!({
                         // Experimental: Allow for rendering terminal output from the agents
                         "terminal_output": true,
+                        "terminal-auth": true,
                     })),
                 },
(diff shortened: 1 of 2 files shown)


crates/agent_servers/src/acp.rs
```diff
@@ -178,6 +178,7 @@ impl AcpConnection {
                     meta: Some(serde_json::json!({
                         // Experimental: Allow for rendering terminal output from the agents
                         "terminal_output": true,
+                        "terminal-auth": true,
```
benbrandt, line 181: Fine for now, but note to self, these should probably both be camelCased before upstreaming :D

PR #41956, opened 2025-11-05
node: Move managed runtime to v24 LTS

versecafe:
Release Notes:

- Moved managed Node runtime to v24 LTS


ConradIrwin:
What's the motivation for changing this?

versecafe:
@ConradIrwin For the most part it's minor perf bumps but 22 just exited its active support window.

https://nodejs.org/en/about/previous-releases

ConradIrwin:
:+1: thanks. I'll merge after the releases tomorrow to give us some extra bake time

PR #41960, opened 2025-11-05
git: Add cursor pointer on last commit to check changes

delvin02:
Release Notes:
-  Improved visual cue on git panel ui to check previous commit changes 

Before: 
Screenshot 2025-11-05 at 2 06 49 pm

After:
Screenshot 2025-11-05 at 2 06 24 pm

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -3618,43 +3618,43 @@ impl GitPanel {
                 .border_t_1()
                 .border_color(cx.theme().colors().border.opacity(0.8))
                 .child(
-                    div().overflow_hidden().line_clamp(1).child(
-                        div()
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks, agree that it makes sense to have a cursor pointer here! Was there a specific reason for the other changes like removing `flex_grow` and adding a nested `div()`?

delvin02:
Removed flex_grow will only allow cursor pointer appears when hovering directly over the commit message, not the surrounding empty space (see image below.)

when hovering to empty space:
Screenshot 2025-11-05 at 3 17 58 pm

when hovering to commit message:
Screenshot 2025-11-05 at 3 22 40 pm


also reverted the nested `div()`, thanks for pointing out!

cole-miller:
Thanks!

PR #41967, opened 2025-11-05
gpui: Fix ellipsis not applying when line width is close to truncate width

Moulberry:
When the width of the text (e.g. 102.8px) is close to the truncate width (e.g. 102px), the ellipsis don't apply due to the use of .floor().

This flooring behaviour seems incorrect and results in issues like the one in the attached screenshot

image

### DIFF
--- a/crates/gpui/src/text_system/line_wrapper.rs
+++ b/crates/gpui/src/text_system/line_wrapper.rs
@@ -151,7 +151,7 @@ impl LineWrapper {
             let char_width = self.width_for_char(c);
             width += char_width;
 
-            if width.floor() > truncate_width {
+            if width > truncate_width {
                 let result =
                     SharedString::from(format!("{}{}", &line[..truncate_ix], truncation_suffix));
                 let mut runs = runs.to_vec();


Moulberry:
@cla-bot check

huacnlee:
Hi @Moulberry try `text_wrapper` example to test.

```bash
cargo run -p gpui --example text_wrapper
```

You change fixed some incorrect wrap, but there have 1 case not correct.

image

Moulberry:
Converted to draft to add additional truncation fixes: what huacnlee mentioned above, as well as the broken truncation when line clamp is used (can be seen in the "Ellipsis 2 lines" entry above, the word 'dictionary' isn't getting truncated)

Moulberry:
> Hi @Moulberry try `text_wrapper` example to test.
> 
> ```shell
> cargo run -p gpui --example text_wrapper
> ```
> 
> You change fixed some incorrect wrap, but there have 1 case not correct.

I believe I know the cause.
line_wrapper measures each individual character and uses that for the computation of wrapping/truncation. This is wrong because it doesn't take into account kerning. When line_layout does wrapping, it accounts for kerning.

I debugged the widths of each, and I can see that it becomes different around the NC in "TRUNCATE" (0.4px), which is a common kerning pair. It becomes even more desynced after CA, which has a very large amount of kerning (1.5px).

"TRUNCATE" according to line_layout (kerning):
`12.910156px, 25.78125px, 40.664063px, 55.73242px, 69.95117px, 82.05078px, 94.96094px, 106.98242px`
"TRUNCATE" according to line_wrapper (no kerning):
`12.910156px, 25.78125px, 40.664063px, 55.73242px, 70.3418px, 84.140625px, 97.05078px, 109.072266px`

I think the only solution is to abandon the cached char widths in line_wrapper and use the LineLayoutCache to properly compute the width of the line. I'm going to join the Zed discord shortly to discuss this.

Please let me know if switching line_wrapper to use LineLayoutCache instead of the current per-character cache is an acceptable solution and I'll implement it tomorrow. I also have a few other fixes for wrapping/truncation as mentioned above

Moulberry:
Superseded by 42092

PR #20812, opened 2024-11-18
gpui: Add linear gradient support to fill background

huacnlee:
Release Notes:

- gpui: Add linear gradient support to fill background

Run example:

```
cargo run -p gpui --example gradient
cargo run -p gpui --example gradient --features macos-blade
```

## Demo

In GPUI (sRGB):

image

In GPUI (Oklab):

image

In CSS (sRGB): 

https://codepen.io/huacnlee/pen/rNXgxBY

image

In CSS (Oklab):

https://codepen.io/huacnlee/pen/wBwBKOp

image

---

Currently only support 2 color stops with linear-gradient. I think this is we first introduce the gradient feature in GPUI, and the linear-gradient is most popular for use. So we can just add this first and then to add more other supports.

PR #41976, opened 2025-11-05
gpui: Improve `Background` type

huacnlee:
Release Notes:

- N/A

---

I was changed `Fill` to use `Background` since PR #20812.

But it's not inconvenient to retrieve the background color already set in the current style.

For example: I have use `StyleRefinement` a lot in GPUI Component, I may define a `style` for an element and `impl Styled for MyElement` for allows user to override the element default style.

In this case, I often need to get the `bg` of user sets to do something, e.g.: Adjust that color opacity to border use.

```rs
struct MyElement {
    style: StyleRefinement,
}

impl Styled for MyElement {
    fn style(&mut self) -> &mut StyleRefinement {
        &mut self.style
    }
}

fn render(...) {
    let bg_color = self.style.background;
    // Adjust bg_color ...

    div()
        .bg(default_color)
        .border_color(...)
        .map(|this| this.style().refine(&self.style))
}
```

- [x] Test macOS
- [x] Test Blade macOS
- [x] Test Windows

### DIFF
--- a/crates/gpui/src/color.rs
+++ b/crates/gpui/src/color.rs
@@ -10,6 +10,8 @@ use std::{
     hash::{Hash, Hasher},
 };
 
+use crate::{Pixels, px};
+
 /// Convert an RGB hex color code number to a color type
 pub fn rgb(hex: u32) -> Rgba {
     let [_, r, g, b] = hex.to_be_bytes().map(|b| (b as f32) / 255.0);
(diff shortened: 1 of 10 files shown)


mikayla-maki (changes requested):
The presence of this method tells me we don't have the narrative of `Fill` and `Background` settled very well. Why introduce a new word / concept for something we already have? As such, I think we could use a helper, maybe `Background::fill()` to achieve something similar :)

huacnlee:
I changed `Background` to an enum and exposed its members.

I initially added this because the `Background` struct is designed for GPU internals, especially the `tag` property, which isn't suitable for direct use with the GPUI API. Therefore, I intentionally set it to `pub(crate)`.

So now I'm wrapping it in a more easily understood enum type for GPUI users, while the internal structure, renamed `BackgroundColor`, is not publicly accessible.

mikayla-maki:
Hi, thanks for this change but we're going to be focusing on stuff we encounter when building Zed.

PR #42009, opened 2025-11-05
editor: Do not scroll if all selections can't fit on screen

anirudh1713:
Closes #32894

### TODO
- [ ] Confirm is something similar needs to be fixed for horizontal scroll as well

Release Notes:

- Fixed an issue where autoscroll fit strategy was falling back to autoscroll newest when all selection can't fit on screen


anirudh1713:
@cla-bot check

anirudh1713:
Context on the changes: https://github.com/zed-industries/zed/issues/32894#issuecomment-3419053268

To add to this, I've update the existing test as it was contradictory. I'm not sure what the test was targeting so I think someone with more context can verify if its _ok_ to modify this behavior.

Another approach that comes to mind is to have an explicit (optional) fallback strategy that consumer of enum can pass. e.g. with autoscroll fit, you may optionally pass autoscroll newest as a fallback 🤔

anirudh1713:
I'm marking this as _Ready for review_ assuming the current approach is ok, but I suppose further discussions are needed.

Thanks.

MrSubidubi (commented):
First of all, a very very big sorry from my side for the delay in feedback/review also here. I only work part-time and was fully occupied for the entirety of end of November/December with setting up CI/CD for the extensions org, so sorry for not getting back to you in a timely manner - that is fully on me and I'd like to apologize for that.

Speaking of the code, I just had a look and I agree this might not be the optimal approach: Looking at it, would it perhaps not make more sense to order the selections in such a way that the initial selection is just the last one? I am a bit sceptical as to whether we should never scroll for the `fit` strategy as we use that in many places, e.g. for undoing, and I think it would be better for such cases that the editor scrolls to indicate something happened, even if the scroll does not end up being an optimal one.

Hence, I think we either
- modify the selection logic in `select_all_matches` so that the last selection is in fact the first one? 
- need to add another autoscroll mode that describes the behavior you have here, because I do think it might be desirable in some cases
- add an autoscroll behavior for `oldest_selection_top`, although I think this might not even work for this case.

What do you think?

MrSubidubi:
Will close this as stale for the moment, but I would be very happy to reopen this if you decide to pick this up again. Please ping me if you decide to do so. Thanks!

PR #42037, opened 2025-11-05
Improve keymap error formatting and add settings button icon

### LINKED ISSUE #41938: Font size is too big in keymap file error pop-up
### Description

Steps to reproduce:

Put this in your keymap file:

```json
[{ "context": "Terminal", "use_key_equivalents": true, "bindings": { "cmd-r": ["terminal::RerunTask"] } }]
```

**Expected Behavior**: Font size is the same as other error pop-ups
**Actual Behavior**: Font size is too big

Zed: v0.210.4 (Zed) 
OS: macOS 15.7.1
Memory: 16 GiB
Architecture: x86_64

johnklucinec:
Closes https://github.com/zed-industries/zed/issues/41938

For some error messages relating to the keymap file, the font size was too large. This was due to the error message being a child `MarkdownString` instead of a `SharedString`. A `.text_xs()` method is being applied to this notification, but it appears not to affect the markdown text. I found that the H5 text size in markdown is the same size as other error messages, so I made each element (that had text) that size. There was also a special case for bullet points. 

I also added a gear icon to the settings button, so it was more in line with other app notifications. 

Error message (text too large):
keymap-broke

Expected behavior (notification with correct text sizing and icon):
keymap-fixed

Example behavior:
settings

Release Notes:

- Improved UI for keymap error messages.


johnklucinec:
@cla-bot check

danilo-leal (approved):
Thank you! Sorry for the delayed review here. Pushed a slightly different approach, simpler because we're not having to hardcode the label to be an h5.

PR #42043, opened 2025-11-05
Adjust terminal decorative character ranges to include missing Powerline characters

lemoncove:
Closes #41975.

This change adjusts some of the ranges which were incorrectly labeled or excluded characters. The new ranges include three codepoints which are not assigned in Nerd Fonts. However, one of these codepoints were already included prior to this change. These codepoints are:

- U+E0C9, between the two ice separators
- U+E0D3, between the two trapezoid separators. The ranges prior to this PR already included this one.
- U+E0D5, between the trapezoid separators and the inverted triangle separators

I included these so as to not overcomplicate the ranges by cherry-picking the defined codepoints. That being said, if we're okay with this and an additional unassigned codepoint (U+E0CB, between ice separators and honeycomb separators) being included then a simple range from 0xE0B0 to 0xE0D7 nicely includes all of the Powerline characters.

I wasn't sure how to write tests for this so I just added two characters to the existing tests which were previously not covered lol. All of the Powerline characters can be seen [here](https://www.nerdfonts.com/cheat-sheet) by searching `nf-pl`.

Release Notes:

- Fixed certain Powerline characters incorrectly having terminal contrast adjustment applied.


ConradIrwin:
Thanks for the fix!

PR #7140, opened 2024-01-31
Use mimalloc as default allocator

andrew-ares:
From https://github.com/microsoft/mimalloc:
> In our benchmarks (see [below](https://github.com/microsoft/mimalloc#performance)), mimalloc outperforms other leading allocators (jemalloc, tcmalloc, Hoard, etc), and often uses less memory. A nice property is that it does consistently well over a wide range of benchmarks. There is also good huge OS page support for larger server programs.

Release Notes:

- Changed default allocator to mimalloc.

PR #42049, opened 2025-11-06
Enable jemalloc allocator with "jemalloc" feature flag

redforks:
I found that at least in linux, use jemalloc saves a lot of memory, although zed uses much lesser memory than vscode or other ides, but saves memory is a good thing, especially when use with rust-analyzer, less swap usage and better responsibility and faster cargo builds.

And Maybe not use system allocator to build bundles, it is much better in speed and memory usage no matter use mimalloc or jemalloc

Release Notes:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17315,6 +17315,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "tikv-jemalloc-sys"
+version = "0.6.1+5.3.0-1-ge13ca993e8ccb9ba9847cc330696e02839f328f7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd8aa5b2ab86a2cefa406d889139c162cbb230092f7d1d7cbc1716405d852a3b"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
+name = "tikv-jemallocator"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0359b4327f954e0567e69fb191cf1436617748813819c94b8cd4a431422d053a"
+dependencies = [
+ "libc",
+ "tikv-jemalloc-sys",
+]
+
+[[package]]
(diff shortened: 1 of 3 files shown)


ConradIrwin:
/perf

(Trying out our new perf infrastructure more than anything.)

 Are there any potential compatibility concerns with jemalloc? I'm surprised that it's not the default if it's just better

redforks:
jemalloc used to be default rust memory allocator before rust 1.32, [rust 1.32 changes](https://blog.rust-lang.org/2019/01/17/Rust-1.32.0). [rust-analyzer compares different memory allocators](https://github.com/rust-lang/rust-analyzer/issues/1441#issuecomment-2016419122)

SomeoneToIgnore:
Maybe also check https://github.com/zed-industries/zed/pull/7140#issuecomment-2059079602 — mimalloc used to actually not retain the memory with use, hence "leaked".

Also, breaking changing the feature from `#[cfg(feature = "mimalloc")]` to something else — is that needed for something?

DSharifi:
Curious, was there a reason you ditched using `jemalloc` and closed he PR?

redforks:
Refactoring my forked repository by delete and re-fork, accidentally closed this PR.

I am using zed compiled with jemalloc, I found use jemalloc can saves about 200M RES memory, so I think jemalloc has its meaning under linux. Because in linux, rust default allocator is libc malloc() and its design is very traditional, it is good for most linux shell commands, such as find, grep etc, their lifetime is very short, simpler allocator lesser binary size and maybe better performance. But for a long lived application, like zed, it is not good. 

also see: https://kerkour.com/rust-jemalloc

Another reason of closed the PR, because I want to run tests and gather some evidence, but recently I don't get the time to do the job.

btw, I also use my compiled version of rust-analyzer, with jemalloc enabled, rust-analyzer has jemalloc feature. Enable jemalloc on zed and rust-analyzer saved my memory, without it, my linux box swap out more than 1G memory into swap.

PR #42071, opened 2025-11-06
editor: Improve multibuffer header spacing

libondev:
BEFORE:
image

AFTER:
image

Release Notes:

- N/A


danilo-leal (approved):
Thank you! Ended up reverting somethings to simplify the diff and still improve spacing. Good move ehre was to use `rems_from_px` so the spacing scales proportionally to the buffer font size, which was the cause of the mismatch you were seeing before. Looks good now!

PR #42079, opened 2025-11-06
agent_ui: Do not show Codex wsl warning on wsl

Veykril:
Release Notes:

- Fixed the codex wsl warning being shown on wsl itself

### DIFF
--- a/crates/agent_ui/src/acp/thread_view.rs
+++ b/crates/agent_ui/src/acp/thread_view.rs
@@ -294,7 +294,6 @@ pub struct AcpThreadView {
     resume_thread_metadata: Option<DbThreadMetadata>,
     _cancel_task: Option<Task<()>>,
     _subscriptions: [Subscription; 5],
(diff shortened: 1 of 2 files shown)


crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -5772,13 +5769,15 @@ impl Render for AcpThreadView {
             })
             .children(self.render_thread_retry_status_callout(window, cx))
             .children({
-                #[cfg(target_os = "windows")]
+                if self
+                    .project
+                    .read(cx)
+                    .remote_connection_options(cx)
+                    .is_some_and(|it| it.is_wsl())
```
bennetbo, line 5776: Isn't this going to show the warning only when using WSL?
Seems like we want to check if we are running on Windows but not running remote WSL project

Veykril: 🤦

PR #42082, opened 2025-11-06
git: Add support for repository excludes file

errmayank:
Closes #4824

Release Notes:

- Added support for Git repository excludes file `.git/info/exclude`

### DIFF
--- a/crates/worktree/src/ignore.rs
+++ b/crates/worktree/src/ignore.rs
@@ -101,18 +101,6 @@ impl IgnoreStack {
                 }
             }
             IgnoreStackEntry::RepoExclude { ignore, parent } => {
-                let combined_path;
-                let abs_path = if let Some(repo_root) = self.repo_root.as_ref() {
-                    combined_path = ignore.path().join(
-                        abs_path
-                            .strip_prefix(repo_root)
-                            .expect("repo root should be a parent of matched path"),
-                    );
-                    &combined_path
-                } else {
-                    abs_path
-                };
-
                 match ignore.matched(abs_path, is_dir) {
                     ignore::Match::None => IgnoreStack {
                         repo_root: self.repo_root.clone(),
--- a/crates/worktree/src/ignore.rs
+++ b/crates/worktree/src/ignore.rs
@@ -13,12 +13,22 @@ pub enum IgnoreStackEntry {
     Global {
         ignore: Arc<Gitignore>,
     },
+    RepoExclude {
+        ignore: Arc<Gitignore>,
+        parent: Arc<IgnoreStackEntry>,
+    },
     Some {
         abs_base_path: Arc<Path>,
         ignore: Arc<Gitignore>,
         parent: Arc<IgnoreStackEntry>,
     },
     All,
+}
+
+#[derive(Debug)]
+pub enum IgnoreKind {
+    Gitignore(Arc<Path>),
+    RepoExclude,
 }
 
 impl IgnoreStack {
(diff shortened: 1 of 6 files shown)


errmayank:
@cole-miller ready for review :)

crates/worktree/src/ignore.rs
```diff
@@ -84,6 +100,29 @@ impl IgnoreStack {
...
+                match ignore.matched(abs_path, is_dir) {
+                    ignore::Match::None => IgnoreStack {
+                        repo_root: self.repo_root.clone(),
+                        top: parent.clone(),
+                    }
+                    .is_abs_path_ignored(abs_path, is_dir),
+                    ignore::Match::Ignore(_) => true,
+                    ignore::Match::Whitelist(_) => false,
+                }
+            }
```
cole-miller, line 113: I think this can be simplified a bit, and we can just do `ignore.matched(abs_path)`.

cole-miller (changes requested):
Thank you! This looks really good, I just have one minor comment.

crates/worktree/src/ignore.rs
errmayank: For last two match arms? I'm not sure how `ignore.matched(abs_path)` would fit, do you mean something like:

```rs
match ignore.matched(abs_path, is_dir) {
    ignore::Match::None => IgnoreStack {
        repo_root: self.repo_root.clone(),
        top: parent.clone(),
    }
    .is_abs_path_ignored(abs_path, is_dir),
    matched => matched.is_ignore(),
}
```

cole-miller: Sorry, messed up the comment range! I meant that the path computation in the `RepoExclude` arm can be replaced by just using `abs_path` (based on my read of the docs here: https://docs.rs/ignore/latest/ignore/gitignore/struct.Gitignore.html#method.matched)

errmayank: Ah got it! updated.

cole-miller:
Thanks! I tried running this locally and it looks like changes to `.git/info/exclude` do not propagate to the Zed ignore statuses. Rereading the code, it seems like `repo_exclude_by_work_directory_abs_path` probably needs to track `needs_update` like `ignores_by_parent_abs_path` already does. It looks like the test was meant to cover this, but right now it asserts that `.env.example` is still tracked after adding it to `.git/info/exclude`, which is the opposite of the behavior we want.

errmayank:
> Thanks! I tried running this locally and it looks like changes to `.git/info/exclude` do not propagate to the Zed ignore statuses. Rereading the code, it seems like `repo_exclude_by_work_directory_abs_path` probably needs to track `needs_update` like `ignores_by_parent_abs_path` already does.

I see, from my manual tests it seemed to work. Can you tell me how you debug the ignore statuses in Zed? I'll check how `ignores_by_parent_abs_path` does it. Thanks!

> It looks like the test was meant to cover this, but right now it asserts that `.env.example` is still tracked after adding it to `.git/info/exclude`, which is the opposite of the behavior we want.

I think the test could be a bit clearer, but `.env.example` is asserted as tracked coz `.gitignore` still has `!.env.example` and local `.gitignore` has higher priority.

cole-miller:
I tested by cloning the https://github.com/agentclientprotocol/agent-client-protocol repo, adding `README.md` to `.git/info/exclude`, and opening Zed---this causes `README.md` to have a faded appearance in the project panel (depending on the theme):

image

Then I did `echo '' >.git/info/exclude` from the terminal. After that, `README.md` still shows as faded in the project panel, but if I restart Zed it appears normal. So `.git/info/exclude` is being applied correctly at startup, but when it changes it seems like we're not recomputing the ignore statuses for the repo.

> I think the test could be a bit clearer, but .env.example is asserted as tracked coz .gitignore still has !.env.example and local .gitignore has higher priority.

My mistake, thanks for pointing that out--seems like the test is correct then, I'm sort of surprised it does since it seems like the code for recomputing ignore statuses when `.git/info/exclude` changes isn't there right now. Maybe something else is causing us to recompute ignore statuses in the test?

errmayank:
I realized that I only tested this for untracked files before and weirdly even that didn't seem to work properly, my bad.

I've added updates and it seems to work fine now, tested it on [agent-client-protocol](https://github.com/agentclientprotocol/agent-client-protocol) repository.

Here's the before/after:

https://github.com/user-attachments/assets/729fff14-6e60-4da3-bf8f-aef309fd8260

<br>

> Maybe something else is causing us to recompute ignore statuses in the test?

Hmm, I'm not sure about that either.

esthertrapadoux:
Hey @errmayank just following up to let you know the team went pretty deep in more git feature work and we got behind on PR reviews. We're ramping back up and appreciate the patience! 👍

errmayank:
@esthertrapadoux no worries at all, thanks for letting me know! :)

cole-miller:
Thanks @errmayank!

errmayank:
I'm not sure why `editor_tests::test_outdent_after_input_for_bash` is failing consistently on Linux. Seems totally unrelated.

For @cameron1024 as well it failed locally during our pairing session.

cole-miller:
Sorry about that flaky test failure! I'll try to figure out what's going on with it and get this merged.

errmayank:
Thanks!

cole-miller (approved):

PR #17149, opened 2024-08-30
gpui: Fix text ellipsis appearing even the flex element has space

huacnlee:
Release Notes:

- N/A

There was a calculation bug before. When we added `text_ellipsis` to the flex element, it would always show ellipsis no matter how long it was. 

Actually we can't use `flex` and `text_ellipsis` at same time, the CSS also not support this. But this bug will let user confuse.

### Before

image

### After

image

PR #42092, opened 2025-11-06
gpui: Improve line_wrapper truncation

Moulberry:
Release Notes:
- Fixed gpui line wrapper truncation not occurring when line width is close to truncate width

Before:
image

After:
image

- Fixed gpui line wrapper truncation not working correctly when used with line clamp

Before:
image

After:
image

- Improved gpui line wrapper truncation: it now skips whitespace, so ellipsis are placed next to words. If a line ends with whitespace, the whitespace can be truncated without applying the suffix.

Before:
image

After:
image

There are still some issues with text truncation/wrapping which I believe are due to subtle differences between the line wrapper and text layout systems. One of the differences I identified is that the line wrapper isn't kerning aware. Unfortunately, it seems like making it kerning aware would be a large refactor and if done naively could have a significant negative performance impact.

To hopefully mitigate any differences between the line wrapper and the text layout system, this PR adds a small bias to the truncation width of -0.25px. This isn't ideal, but it's better than the layout being incorrect.

### DIFF
--- a/crates/gpui/src/elements/text.rs
+++ b/crates/gpui/src/elements/text.rs
@@ -356,10 +356,7 @@ impl TextLayout {
                 let (truncate_width, truncation_suffix) =
                     if let Some(text_overflow) = text_style.text_overflow.clone() {
                         let width = known_dimensions.width.or(match available_space.width {
-                            crate::AvailableSpace::Definite(x) => match text_style.line_clamp {
-                                Some(max_lines) => Some(x * max_lines),
-                                None => Some(x),
-                            },
+                            crate::AvailableSpace::Definite(x) => Some(x),
                             _ => None,
                         });
 
@@ -377,14 +374,25 @@ impl TextLayout {
                     return text_layout.size.unwrap();
                 }
 
-                let mut line_wrapper = cx.text_system().line_wrapper(text_style.font(), font_size);
                 let (text, runs) = if let Some(truncate_width) = truncate_width {
-                    line_wrapper.truncate_line(
-                        text.clone(),
-                        truncate_width,
-                        &truncation_suffix,
-                        &runs,
-                    )
+                    let mut line_wrapper = cx.text_system().line_wrapper(text_style.font(), font_size);
+                    if let Some(wrap_width) = wrap_width && let Some(line_clamp) = text_style.line_clamp {
+                        line_wrapper.truncate_last_wrapped_line(
+                            text.clone(),
+                            truncate_width,
+                            &truncation_suffix,
+                            &runs,
+                            wrap_width,
+                            line_clamp
+                        )
+                    } else {
+                        line_wrapper.truncate_line(
+                            text.clone(),
+                            truncate_width,
+                            &truncation_suffix,
+                            &runs,
+                        )
+                    }
                 } else {
                     (text.clone(), Cow::Borrowed(&*runs))
                 };
--- a/crates/gpui/src/elements/text.rs
+++ b/crates/gpui/src/elements/text.rs
@@ -375,15 +375,18 @@ impl TextLayout {
                 }
 
                 let (text, runs) = if let Some(truncate_width) = truncate_width {
-                    let mut line_wrapper = cx.text_system().line_wrapper(text_style.font(), font_size);
-                    if let Some(wrap_width) = wrap_width && let Some(line_clamp) = text_style.line_clamp {
+                    let mut line_wrapper =
+                        cx.text_system().line_wrapper(text_style.font(), font_size);
+                    if let Some(wrap_width) = wrap_width
+                        && let Some(line_clamp) = text_style.line_clamp
+                    {
                         line_wrapper.truncate_last_wrapped_line(
                             text.clone(),
                             truncate_width,
                             &truncation_suffix,
                             &runs,
                             wrap_width,
-                            line_clamp
+                            line_clamp,
                         )
                     } else {
                         line_wrapper.truncate_line(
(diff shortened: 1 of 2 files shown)


Moulberry:
The PR includes release notes. Perhaps the images are confusing it?

huacnlee:
You may need to check again; I found there are still some incorrect break.

image

Moulberry:
> You may need to check again; I found there are still some incorrect break.

As I mentioned in the original message:
> There are still some issues with text truncation/wrapping which I believe are due to subtle differences between the line wrapper and text layout systems. One of the differences I identified is that the line wrapper isn't kerning aware.

Ideally the text_wrapper should be rewritten to have identical wrapping code to the line layout engine, which shapes the entire line instead of summing the width of each character, however that is a very large change and is out of scope.

Unless you're noticing regressions, this PR is just improvement, not a completely perfect solution

huacnlee:
I repeatedly tried your modified version and also attempted to fine-tune it locally.

Your version didn't solve any problems in `line clamp` mode (although you did modify the implementation a lot).

The main issue here is the difference between the two implementation mechanisms for line breaks (`wrap_line`) and line breaks (`truncate_line`) when encountering CJK characters.

huacnlee:
You can have a try of my test fix:

https://github.com/huacnlee/zed/commit/8eafff883c850226174c6dc7466b25e76d19dbe2

## Issue 1:

> image

It just changes a little, but you mentioned issues 1 has fixed.

image

## Issue 2

Still will overflow, but now no incorrect break line now.

image

## Issue 3

Not fix.

image

huacnlee:
Hi @Moulberry see my another version to solve the **"Issue 2"** truncate width.

https://github.com/zed-industries/zed/pull/42116

Moulberry:
> You can have a try of my test fix:
> 
> [huacnlee@8eafff8](https://github.com/huacnlee/zed/commit/8eafff883c850226174c6dc7466b25e76d19dbe2)
> 
> etc

Are you are you even used this PR when taking this screenshot? Can you show a video with running `git log` and then the example?

The last screenshot in particular is very suspicious because the code for ignoring spaces is very straightforward and yet your screenshot has the ellipsis after the space

huacnlee:
Ignore `huacnlee@8eafff8` that not good enough.

huacnlee:
> Are you are you even used this PR when taking this screenshot? Can you show a video with running git log and then the example?

No, it only have 1 commit.

https://github.com/huacnlee/zed/commits/fix-truncate-line/

Moulberry:
> > Are you are you even used this PR when taking this screenshot? Can you show a video with running git log and then the example?
> 
> No, it only have 1 commit.
> 
> https://github.com/huacnlee/zed/commits/fix-truncate-line/

Are the screenshots you sent above showing my pull request or your pull request?

Moulberry:
By the way, the change in huacnlee@8eafff8 is wrong.

That is the old calculation which causes ellipsis to appear even when there is room.

The current calculation was introduced in this pull request:
https://github.com/zed-industries/zed/pull/17149/

It is necessary to consider the suffix_width as separate to the width, you cannot combine them into a single value like huacnlee@8eafff8

huacnlee:
Sorry, perhaps my previous reply wasn't clear.

Initially, I noticed your PR Diff had undergone many changes, including adjustments to the entire truncation logic. However, after running it, I found that several issues remained unresolved. This approach didn't seem cost-effective.

So, I tried to find a simpler solution, which led to `8eafff8` (though I later found this wasn't working either, so I didn't use it). You can ignore this commit.

Later, I considered using wrap_line to handle multi-line ellipsis support, so yesterday I made change #42116 to fix that issue.

Moulberry:
> Later, I considered using wrap_line to handle multi-line ellipsis support, so yesterday I made change #42116 to fix that issue.

Yes, this PR also uses wrap_line to handle multi-line ellipsis support.
If there is an issue with this PR please say what it is. Making new PRs with similar code is just confusing

huacnlee:
Oh, I didn't notice you changed it like that too.

I can close my PR, and you can see my changes and merge them in. You can try to avoid making a lot of changes to `truncate_line`.

huacnlee:
##### This test on macOS of this PR:
https://github.com/user-attachments/assets/d69a6c35-f84a-4ff5-9638-15a7197f0a7c

> Same issue exists in PR #42116.


However, it didn't appear when I tested it on Windows last night. 

##### This test on Windows by PR #42116:
https://github.com/user-attachments/assets/489e0a30-1cb7-4b0a-9301-41f5a0c5998e


So we might need to check macOS, Windows, and Linux separately; there seem to be slight differences.

Moulberry:
> Oh, I didn't notice you changed it like that too.
> 
> I can close my PR, and you can see my changes and merge them in. You can try to avoid making a lot of changes to `truncate_line`.

The reason why I made a lot of changes is because I didn't want to modify the signature of the public truncate_line function, so I added a separate function that handles the case.

However, if it is okay to add a breaking change to the public truncate_line API then I will do so and adjust my PR.

> So we might need to check macOS, Windows, and Linux separately; there seem to be slight differences.

The difference is probably because of the font that is in use. I think the different platforms have different system fonts, right? I don't know how exactly, but is it possible for you to extract the font that is active on Windows and macOS for you? I should then be able to test it locally with that specific font

as-cii:
Hey @Moulberry, thank you for this pull request! While I understand this improves the situation a bit, I have some concerns around the .25px constant, as it doesn't feel principled and it's seemingly not working as expected on other platforms.

I think a better approach would be to use the shaping information we already have when laying out a line, find the rightmost glyph we can render prior to the truncation width (such that it allows rendering the ellipsis).

Closing this for now.

PR #42099, opened 2025-11-06
editor: Fix inconsistent open excerpts behavior in multi buffer

unixzii:
The behavior when clicking "Open file" button is inconsistent with that triggered by the keyboard shortcut. When pressing <kbd>Option</kbd>-<kbd>Return</kbd>, the selection is kept in the new buffer. However, clicking the button doesn't have that effect.

image

This PR updates `on_click` handler of the button to make the behavior aligned with the shortcut.

Note that `open_excerpts_common` method will handle the selection correctly, and `JumpData` only needs to be passed when the selection is actually required to be updated. I don't think it's by design in this case.

Release Notes:

- Fixed inconsistent open excerpts behavior in multi buffer.


danilo-leal:
Hey @unixzii, thanks for the PR! Not sure if I follow, though... what is exactly inconsistent? Right now, on `main`, clicking on the "Open File" button or hitting the keybinding take me to the same place.

unixzii:
Hi @danilo-leal. The issue occurs when there are multiple discontinuous excerpts. Clicking the button won't bring me to the exact selection I was working on. It will always navigate to the region of the first excerpt.

You should be able to tell the difference in this screencast.

https://github.com/user-attachments/assets/2ed77c31-c21a-4024-bb8e-c9b94cf27261

Issue #42027, opened 2025-11-05
Windows: can't install remote server binary


sgraham:
### Summary

Can't install remote server binary

### Description
Steps to trigger the problem:

1. ctrl-alt-shift-o
2. Connect New Server
3. ssh scottg@172.20.39.128

That's a newly imaged Xubuntu VM, with passwordless ssh working from the Windows command line.

**Actual Behavior**:

```
failed to run command: gzip: .zed_server/zed-remote-server-stable-0.210.4-download-40776.gz: No such file or directory
Please try again.
```

Zed.log has this to say which doesn't seem too insightful:

```
2025-11-05T11:12:47-08:00 WARN  [fs] Failed to watch /etc/ssh/ssh_config and its parent directory /etc/ssh:
Input watch path is neither a file nor a directory.
Input watch path is neither a file nor a directory.
2025-11-05T11:12:56-08:00 ERROR [remote::transport::ssh] Failed to download binary on server, attempting to upload server: failed to run command: You must specify a method through --method=HTTPMethod to use with --body-data or --body-file.

2025-11-05T11:12:56-08:00 INFO  [remote::transport::ssh] uploading remote development server to ".zed_server/zed-remote-server-stable-0.210.4-download-40776.gz" (28231kb)
2025-11-05T11:12:56-08:00 INFO  [remote::transport::ssh] uploaded remote development server in 255.1375ms
2025-11-05T11:12:57-08:00 ERROR [workspace::notifications] failed to run command: gzip: .zed_server/zed-remote-server-stable-0.210.4-download-40776.gz: No such file or directory
```

I also confirmed on the server that gzip's installed:
```
scottg@logic:~$ gzip -V
gzip 1.12
Copyright (C) 2018 Free Software Foundation, Inc.
Copyright (C) 1993 Jean-loup Gailly.
This is free software.  You may redistribute copies of it under the terms of
the GNU General Public License <https://www.gnu.org/licenses/gpl.html>.
There is NO WARRANTY, to the extent permitted by law.

Written by Jean-loup Gailly.`
```

### Zed Version and System Specs

Zed: v0.210.4 (Zed) 
OS: Windows 10.0.26100
Memory: 47.9 GiB
Architecture: x86_64
GPU: AMD Radeon R9 200 Series || AMD Corporation || 20.10.35.02 (20.10.35.02-200821a-360470C-ATI)

bwks:
[...]

PR #42103, opened 2025-11-06
remote: Flush to stdin when writing to sftp

Veykril:
https://github.com/zed-industries/zed/issues/42027#issuecomment-3497210172

Release Notes:

- Fixed ssh remoting potentially  failing due to not flushing stdin to sftp


feeiyu:
I noticed the upload directory logic is also missing a flush() call, which could lead to incomplete data writes. Should we add it here as well?

https://github.com/zed-industries/zed/blob/0b53da18d579606c75562e4334a96ac0fcdcc590/crates/remote/src/transport/ssh.rs#L303-L310

Veykril:
we definitely should, nice catch!

feeiyu:
@Veykril  I noticed  https://github.com/zed-industries/zed/pull/42126 already cherry-pick to preview, should we cherry-pick this PR as well?

Issue #30851, opened 2025-05-16
Slow filesystem updates in file explorer


alshdavid:
### Summary

Filesystem changes on large directories are slow to show up in the file explorer

Steps to reproduce:

```bash
mkdir foo
zed foo
cd foo
npm install @atlaspack/cli
```
Look at file explorer
```bash
# slow to show up the change in the explorer
touch ./node_modules/foo.txt
rm -rf node_modules
```

Expected Behavior:
Updates show up instantly

Actual Behavior:

Slow to update, sometimes requires `cnt+p` `reload: workspaces`

<details><summary>click to expand</summary>

Linux Fedora 42
KDE Plasma
Zed Preview 0.187.2

```bash
$ uname -a
Linux fedora 6.14.6-300.fc42.x86_64 #1 SMP PREEMPT_DYNAMIC Fri May  9 20:11:19 UTC 2025 x86_64 GNU/Linux

$ zed --version
Zed preview 0.187.2 8f9b217f7944f53699b44f509e0737402871597f  – /home/dalsh/.local/zed-preview.app/libexec/zed-editor
```
</details>
-->

### Suggestions

Some suggestions
- Add a "rescan/refresh filesystem" to `cnt+p` commands
- Add a "reload edditor" option to `cnt+p` commands 
  - The currently available "reload workspaces" reloads _all_ Zed instances and resets their window positions which is a bit inconvienent

I'm guessing Zed is trying to analyze those folders. I'll check the docs but if there isn't already, it would be good to have a config setting to opt out of scanning certain directories

### Zed Version and System Specs

Zed: v0.187.2 (Zed Preview) OS: Linux Wayland fedora 42 Memory: 93.4 GiB Architecture: x86_64 GPU: AMD Radeon Graphics (RADV GFX1201) || radv || Mesa 25.0.4

DLBPointon:
This would be a great to have, i'm running data analysis in the terminal panel this generate folders of output. Some of which don't appear for hours whilst the analysis is running. With no way of refreshing the folder view until it's a minor gripe... 

A manual *file explorer only* refresh would be great, whether as command or how VSCode has that lil refresh button. Hitting workspace_reload and watching everything get killed was a hard lesson in learning a projects terminology.

GerardPolloRebozado:
I have the same issue even on small project for example using cli that create files like for example `nest g module foo` dont show any files and I need to close the editor and open it again

osiewicz:
Can you still reproduce this?

[...]

PR #42107, opened 2025-11-06
Add refresh explorer button

myyrakle:
related issue #30851

While working, you may encounter instances where files that have been created or modified are slow to sync, or fail to sync at all.

VSCode provides a Refresh button to address this issue.

I wish Zed had a similar button.

image

Release Notes:
- Add Refresh Explorer Button

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13057,6 +13057,7 @@ dependencies = [
  "gpui",
  "language",
  "menu",
+ "panel",
  "pretty_assertions",
  "project",
  "rayon",
--- a/crates/project_panel/Cargo.toml
+++ b/crates/project_panel/Cargo.toml
@@ -45,6 +45,7 @@ workspace.workspace = true
 language.workspace = true
 zed_actions.workspace = true
 telemetry.workspace = true
+panel.workspace = true
 
 [dev-dependencies]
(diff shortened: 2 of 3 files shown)


osiewicz:
> While working, you may encounter instances where files that have been created or modified are slow to sync, or fail to sync at all.

I feel somewhat strongly about *not* including the refresh button. I think that'd paper over an issue elsewhere in the stack, plus I'm not even positive if it'd solve the issue that folks in #30851 reported after my fixes to project panel performance (that it's still slow to update with remote projects, which is a *data* problem, not UI problem). 

Instead of adding a refresh button, let's understand why they're slow to sync and address that.

Have you found it to improve the remote workflow anyhow?

myyrakle:
Hmm... 
I checked with the latest branch and it seems like the issue has been improved. 

If it can be resolved without the button, then it would be ok.

SomeoneToIgnore:
I also do not believe this PR helps with anything significant?

The refresh button merely adds a way to re-populate the project panel from whatever FS entry index Zed had accumulated during background scan — after FS sync bugs, that index is not populated properly hence refreshing project panel on top would not help anything.

osiewicz:
I'm gonna close this one out for the time being. We can always just revisit if needed.

PR #42112, opened 2025-11-06
Add privacy-first mode for unlimited AI access

clouraLabs:
## Summary
- Implement privacy-first mode that removes all subscription checks, usage tracking, and billing enforcement from Zed
- Enable unlimited access to AI models without server-side user tracking or commercial restrictions

## Changes
This PR transforms Zed into a privacy-respecting application by removing:
- All subscription plan checking (Free/Pro/Trial tiers)
- Usage limit tracking for model requests and edit predictions
- Upsell banners, trial end notifications, and payment enforcement
- Detailed user data fetching from the server
- Telemetry and billing-related code

The changes affect 5 core crates with a net reduction of 391 lines, simplifying the codebase while providing users with unrestricted AI access and enhanced privacy.

## Impact
Users can now use Zed's AI features without any subscription limitations or server-side tracking of their usage.

### DIFF
--- a/ISSUE_42032_ANALYSIS.md
+++ b/ISSUE_42032_ANALYSIS.md
@@ -1,0 +1,1344 @@
+# 🔍 Complete Analysis: Issue #42032 - High CPU Usage in Zed Editor
+
+**Analysis Date**: November 6, 2025  
+**Analyzed by**: Claude (NeuroNexusIDE)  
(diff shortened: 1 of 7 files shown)


maxdeviant:
This is not a change that we want to make.

We already offer a host of ways to use AI in Zed without using a Zed subscription, such as bringing your own API key or using other agents through ACP.

PR #42117, opened 2025-11-06
Avoid panics in LSP store path handling

CLoaKY233:
Release Notes:

- Fixed incorrect journal paths handling


SomeoneToIgnore (changes requested):
> The panic was caused by a `debug_panic!` which has been replaced with error handling.

This is wrong for release builds:

https://github.com/zed-industries/zed/blob/21f73d9c02681152019ed5703ce8808c841fcbbe/crates/util/src/util.rs#L42-L52

and 

> Invalid file path provided to LSP request

panic message is used elsewhere: https://github.com/zed-industries/zed/blob/21f73d9c02681152019ed5703ce8808c841fcbbe/crates/project/src/lsp_command.rs#L53-L59

I have not found direct usages of `file_path_to_lsp_url` calling `.unwrap` or analogues, so apparently we need to search further up the stack.

So:

* the issue does not seem to be in any of the code the PR edits?
It would be good to get a sequence of actions to reproduce the bug — in the issue description; and to get a test for this into the PR.

* `debug_panic` change seems wrong, hence needs to be reverted.
The rest of the changes seems nice — I can merge them alone (but that won't fix the issue, it seems?) or wait for a better fix, if you prefer.

CLoaKY233:
> This is wrong for release builds:
> 
> https://github.com/zed-industries/zed/blob/21f73d9c02681152019ed5703ce8808c841fcbbe/crates/util/src/util.rs#L42-L52
> 
> and
> 
> > Invalid file path provided to LSP request
> 
> panic message is used elsewhere:
> 

I  missed to mention the first unwrap call, post fixing that, the debug panic was causing crashes, 
`crates/project/src/lsp_store.rs` 

line `7654` 
```rust
        let buffer = buffer.read(cx);
        let file = File::from_dyn(buffer.file())?;
        let abs_path = file.as_local()?.abs_path(cx);
        let uri = lsp::Uri::from_file_path(abs_path).unwrap(); //<----------- This 
 ```
 
  
  ```log
thread 'main' panicked at crates/project/src/lsp_store.rs:7654:54:
called `Result::unwrap()` on an `Err` value: ()

```

So I guess we can revert the debug_panic! change, but the second change, does fix the crash in release builds. 

--- 
although what I do not understand is it creates the invalid path I have provided in the zed's directory instead, 
Eg, specifying a journal path `NONEXIST/xyz/abc/Journal` 

it creates the respective directories in the zed's directory instead, 
`/zed/NONEXT222/home/cloaky/Journal/journal/2025/11/`

or in the `/home`
and upon replacing the `unwrap()` with `ok()?` 

Error logs are : 
```log
2025-11-11T13:54:24+05:30 ERROR [crates/project/src/environment.rs:227] stat ""

Caused by:
    No such file or directory (os error 2)
2025-11-11T13:54:24+05:30 ERROR [project::git_store] failed to get working directory environment for repository ""
2025-11-11T13:54:24+05:30 INFO  [worktree] inserting parent git repo for this worktree: "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md"
2025-11-11T13:54:24+05:30 ERROR [crates/project/src/lsp_store.rs:2425] Invalid file path provided to LSP request: "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md"
2025-11-11T13:54:24+05:30 INFO  [worktree] inserting parent git repo for this worktree: "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md"
2025-11-11T13:54:24+05:30 ERROR [crates/project/src/lsp_store.rs:4336] Invalid file path provided to LSP request: "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md"
2025-11-11T13:54:24+05:30 ERROR [crates/project/src/lsp_store.rs:2425] Invalid file path provided to LSP request: "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md"
2025-11-11T13:54:24+05:30 ERROR [crates/project/src/lsp_store.rs:2425] Invalid file path provided to LSP request: "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md"
2025-11-11T13:54:24+05:30 ERROR [project::lsp_store] "NONEXT222/home/cloaky/Journal/journal/2025/11/11.md" is not parseable as an URI
<disabled>
2025-11-11T13:54:29+05:30 ERROR [crates/gpui/src/window.rs:1131] window not found
2025-11-11T13:54:29+05:30 ERROR [crates/gpui/src/window.rs:1138] window not found



```
--- 
If its intentional that it gets created in the zed dir [ or /home ] , we can maybe check the zed path as fallback when this function runs and path is not found? 
Would this be a good approach ? @SomeoneToIgnore

CLoaKY233:
`journal/src/journal.rs:175-179`
There's the issue ! 

```rs 
// Excerpt from: fn journal_dir
fn journal_dir(path: &str) -> Option<PathBuf> {
    shellexpand::full(path) //TODO handle this better
        .ok()
        .map(|dir| Path::new(&dir.to_string()).to_path_buf().join("journal"))
}
```
```log
2025-11-11T15:27:37+05:30 INFO  [journal] journal_dir called with path: somedirectory/home/cloaky/Journal
2025-11-11T15:27:37+05:30 INFO  [journal] shellexpand::full result: Ok("somedirectory/home/cloaky/Journal")
2025-11-11T15:27:37+05:30 INFO  [journal] Expanded directory: somedirectory/home/cloaky/Journal
2025-11-11T15:27:37+05:30 INFO  [journal] PathBuf created: "somedirectory/home/cloaky/Journal"
2025-11-11T15:27:37+05:30 INFO  [journal] Final journal path: "somedirectory/home/cloaky/Journal/journal"
```

Does not return an absolute path,


something like this, 
```rust
fn journal_dir(path: &str) -> Option<PathBuf> {
    let expanded = shellexpand::full(path).ok()?;
    let base_path = Path::new(expanded.as_ref());
    let absolute_path = if base_path.is_absolute() {
        base_path.to_path_buf()
    } else {
        
        log::warn!(
            "Invalid journal path '{}' (not absolute), falling back to home directory",
            path
        );
        std::env::home_dir()?
    };
    Some(absolute_path.join("journal"))
}
```
Would this be a good approach to fix it?

SomeoneToIgnore:
> Would this be a good approach to fix it?

If I would be fixing this, I'd:

* create a test for this (not done, hence fix can regress later)
* fix the panic call, as unwraps are no good anyway (seems to be done?)
* fix the actual issue: that is not clear to me, https://github.com/zed-industries/zed/issues/42081 mentions invalid paths crash the app, so not sure how using a full path fixes that and find it hard to believe, tbh. 
Maybe you're right though and I do not understand something: again, having a test to reason on would help here.

CLoaKY233:
> * fix the panic call, as unwraps are no good anyway (seems to be done?)

There's recent commits that conflict it, but should work the same way as before by replacing `unwrap()` with `ok()?`

--- 

> * fix the actual issue: that is not clear to me, [Journal Invalid paths crashes zed #42081](https://github.com/zed-industries/zed/issues/42081) mentions invalid paths crash the app, so not sure how using a full path fixes that and find it hard to believe, tbh.

- The issue starts with this function 

https://github.com/zed-industries/zed/blob/ee2e69065714297679d199c25bb25bde558dacdc/crates/journal/src/journal.rs#L175-L179

On providing a non existent path it does not convert it to an absolute path, instead returns it as it is, causing any further `is_absolute()` calls to fail. 


### Further when the journal opens, the below call fails,  

https://github.com/zed-industries/zed/blob/2adc8b755f2e9a0e085e52071e9397af530ad631/crates/project/src/lsp_store.rs#L7654-L7654

calls  `lsp::Uri::from_file_path()` -> `url::Url::from_file_path(path)` -> `path_to_file_url_segments()` 

```rust
fn path_to_file_url_segments(
    path: &Path,
    serialization: &mut String,
) -> 
    ...
    use std::os::unix::prelude::OsStrExt;
    if !path.is_absolute() {
        return Err(());
    }
    ...
``` 


--- 
> * create a test for this (not done, hence fix can regress later)
>   Maybe you're right though and I do not understand something: again, having a test to reason on would help here.

I'll write tests for the same, Just checking relative paths falling back to home,  resolving existing relative paths, and `is_absolute()`

crates/project/src/lsp_store.rs
```diff
@@ -7654,14 +7654,13 @@ impl LspStore {
         let uri = lsp::Uri::from_file_path(&abs_path)
             .ok()
             .with_context(|| format!("Failed to convert path to URI: {}", abs_path.display()))
-            .unwrap();
+            .ok()?;
```
```suggestion
            .log_err()?;
```

Getting a wrong path out of buffer to this level is quite an error, so let's keep it visible.

crates/journal/src/journal.rs
```diff
@@ -224,4 +233,63 @@ mod tests {
...
+
+        #[test]
+        fn test_tilde_expansion() {
+            let result = journal_dir("~/documents");
+            assert!(result.is_some());
+            let path = result.unwrap();
+
+            assert!(path.is_absolute(), "Tilde should expand to absolute path");
+            assert!(path.ends_with("journal"));
+            println!("path: {:?}", path);
```
SomeoneToIgnore: These tests a odd: let's assert the entire file path, instead of printing?

We have `path!` macro to deal with Windows paths semi-automatically in tests, which should also allow to combine some new tests.

crates/journal/src/journal.rs
```diff
@@ -173,9 +173,18 @@ pub fn new_journal_entry(workspace: &Workspace, window: &mut Window, cx: &mut Ap
...
-    shellexpand::full(path) //TODO handle this better
-        .ok()
-        .map(|dir| Path::new(&dir.to_string()).to_path_buf().join("journal"))
+    let expanded = shellexpand::full(path).ok()?;
+    let base_path = Path::new(expanded.as_ref());
+    let absolute_path = if base_path.is_absolute() {
+        base_path.to_path_buf()
+    } else {
+        log::warn!(
+            "Invalid journal path '{}' (not absolute), falling back to home directory",
```
SomeoneToIgnore:
```suggestion
            "Invalid journal path {path:?} (not absolute), falling back to home directory",
```

Let's also mention in the docs, that not only `~` is the default, but a fallback for incorrect paths in the config?

SomeoneToIgnore (commented):
Thank you for the context, now it gets quite clear indeed.
Let's merge after a few small tweaks.

crates/journal/src/journal.rs
CLoaKY233: I totally forgot about the print statement! :P 
would this be better 
```rust
assert!(path.is_absolute(), "Tilde should expand to absolute path");
assert!(path.ends_with("journal"));
let Some(home) = std::env::home_dir() {
    assert_eq!(path, home.join("documents").join("journal"));
}
```

crates/project/src/lsp_store.rs
CLoaKY233: Changed to `.log_err()?` to make the error visible. :D

crates/journal/src/journal.rs
CLoaKY233: Should I make this test unix only? 
```rust
#[test]
fn test_tilde_expansion() {
    let result = journal_dir("~/documents");
    assert!(result.is_some());
    let path = result.unwrap();

    assert!(path.is_absolute(), "Tilde should expand to absolute path");

    if let Some(home) = std::env::home_dir() {
        assert_eq!(path, home.join("documents").join("journal"));
    }
}
 ```
 It would work on windows but will resolve into else condition always

SomeoneToIgnore: Not sure, maybe just an assert that it's Windows code that runs in that branch is enough.

crates/journal/src/journal.rs
CLoaKY233: Done! :D, also mentioned in the docs about invalid path defaults to home.

SomeoneToIgnore (approved):
Thank you.

PR #42122, opened 2025-11-06
Refactor completions (attempt 2)

HactarCE:
This is progress toward multi-word snippets (including snippets with prefixes containing symbols)

Release Notes:

- Removed `trigger` argument in `ShowCompletions` command


okira-e:
This has caused it so that completions appear when the initial characters match the initials of one of my saved snippets even when I have set `show_completions_on_input` to false in the settings.

I don't know if this is intentional or a regression, but when I have `show_completions_on_input` set to false I expect no completions to appear automatically when typing anything.

Edit: I do not have proof that this is what caused the sudden behavior change, but it's the change related to auto completions.

PR #42145, opened 2025-11-06
vim: Fix bug where `d . .` freezes the editor

AidanV:
This bug seems to be caused by pushing an operator (i.e. `d`) followed by a repeat (i.e. `.`)  so the recording includes the push operator and the repeat. When this is repeated (i.e. `.`) it causes an infinite loop. 

This change fixes this bug by pushing a ClearOperator action if there is an ongoing recording when repeat is called.

Release Notes:

- Fixed bug where pressing `d . .` in Vim mode would freeze the editor.

### DIFF
--- a/crates/vim/src/normal/repeat.rs
+++ b/crates/vim/src/normal/repeat.rs
@@ -796,4 +796,15 @@ mod test {
         cx.simulate_shared_keystrokes("@ b").await;
         cx.shared_state().await.assert_eq("aaaaaaabbbˇd");
     }
+
+    #[gpui::test]
+    async fn test_repeat_clear(cx: &mut gpui::TestAppContext) {
+        let mut cx = VimTestContext::new(cx, true).await;
+
+        // Check that, when repeat is preceded by something other than a number,
+        // the current operator is cleared, in order to prevent infinite loops.
+        cx.set_state("ˇhello world", Mode::Normal);
+        cx.simulate_keystrokes("d .");
+        assert_eq!(cx.active_operator(), None);
+    }
 }
--- a/crates/vim/src/normal/repeat.rs
+++ b/crates/vim/src/normal/repeat.rs
@@ -807,4 +807,25 @@ mod test {
         cx.simulate_keystrokes("d .");
         assert_eq!(cx.active_operator(), None);
     }
+
+    #[gpui::test]
+    async fn test_repeat_clear_repeat(cx: &mut gpui::TestAppContext) {
+        let mut cx = NeovimBackedTestContext::new(cx).await;
+
+        cx.set_shared_state(indoc! {
+            "ˇthe quick brown
+            fox jumps over
+            the lazy dog"
+        })
+        .await;
+        cx.simulate_shared_keystrokes("d d").await;
+        cx.shared_state().await.assert_eq(indoc! {
+            "ˇfox jumps over
+            the lazy dog"
+        });
+        cx.simulate_shared_keystrokes("d . .").await;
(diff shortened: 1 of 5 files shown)


dinocosta:
Thank you for catching this @AidanV ! 🔍 

It seems that the issue might be a little bit more nuanced, as I can't reproduce it immediately after opening Zed, I'm actually able to do `d . .` 2 times before the third one freezing Zed. Will think a little bit about this, but I suspect that, what we might want, is for the `Repeat` action (`.`) to only be handled in isolation or when preceded by a number.

Was checking how NeoVim handles something like `d .` and after typing `.` it appears to stop processing the current chain of operators, resetting itself 🤔

dinocosta:
Tested this a little bit and it actually already seems to be doing somewhat the same as NeoVim when it comes to resetting the operator. 

However, it seems that this breaks the recorded action too. For example, trying these actions leads to different outcomes in NeoVim and Zed:

1. Delete a word with `dw`
2. Confirm that you're able to use `2.` to delete 2 words
3. Confirm that using `2d.` does nothing
4. Attempt to use the repeat operation (`.`)
    - In Zed nothing will happen, I suspect we're deleting the recorded actions?
    - In NeoVim, the recorded action continues working as expected
    
Let me know if you want to try a look @AidanV . Can also try checking this next week, otherwise will merge these changes Wednesday morning, to ensure these land in the next Preview 👍

AidanV:
Good find @dinocosta! Yeah this was deleting the previous recording. I added a test to catch this, and I pushed a fix for it.

dinocosta:
Thank you @AidanV ! 🙌 

I believe we'll need even more changes, sorry! I just noticed that, with the current implementation, if I use `d d` to delete a line and then use `2 d .`, nothing happens but the recorded action appears to have been updated to `2 d d` instead, as shown ▼

https://github.com/user-attachments/assets/bfdb6e9d-7f5b-4bee-8d9d-e10e815a51cb

It appears that the count is getting propagated to the recorded action, updating it and affecting `.`. Let me know if you'd like me to have a look 🙂

AidanV:
I added a test and a fix for the bug you pointed out. 

I also added a test and a fix to catch a bug I found where if we have `2 d d` followed by a different count that is dropped (such as `5 d .`) then the next `.` would only delete one line instead of the expected two lines.

In order to fix this I added a `recording_count` field to the Vim globals. This new variable mirrors the relationship that `recording_actions` and `recorded_actions` have with `recording_count` and `recorded_count`.

I can't find any more bugs, but I am sure they exist :).

Thank you for you help on this @dinocosta!

dinocosta:
Thank you for taking the time to update this @AidanV ! I've tested and confirm that it seems everything is working as expected 🎉

I'm holding off on merging this until after we have a new Preview release out, so we can have this in Nightly for a week and have more people experiencing it. I'll also see if I can take some time to revisit the approach, as I'm suspicious that we should be able to get away with just the `recorded_count` field, ensuring that we don't reset it when we start recording, but only set it when we stop recording 🤔

AidanV:
I think you are right that we _can_ get away with just `recorded_count`. As long as we ensure there is no path in the code that calls `Vim::take_count` while we are recording then it _should_ be fine. However, I found it was much easier to reason about the code after adding `recording_count`.

dinocosta:
@AidanV Sorry that I ended up not merging this yet! Got sidetracked with other initiatives and forgot to merge this before today's release process.

Will make this my priority to ensure we get it merged tomorrow! ✌️

AidanV:
@dinocosta gentle reminder to merge this :)

PR #42151, opened 2025-11-06
git: Add base branch support to create_branch

ayu-ch:
Closes [#41674](https://github.com/zed-industries/zed/issues/41674)

Description:
Creating a branch from a base requires switching to the base branch first, then creating the new branch and checking out to it, which requires multiple operations.

Add base_branch parameter to create_branch to allow a new branch from a base branch in one operation which is synonymous to the command `git switch -c <new-branch> <base-branch>`.

Below is the video after solving the issue: 

(`master` branch is the default branch here, and I create a branch `new-branch-2`  based off the `master` branch. I also show the error which used to appear before the fix.)

[Screencast from 2025-11-07 05-14-32.webm](https://github.com/user-attachments/assets/d37d1b58-af5f-44e8-b867-2aa5d4ef3d90)

Release Notes:

- Fixed the branch-picking error  by replacing multiple sequential switch operations with just one switch operation. (Lets Git Together)

### DIFF
--- a/crates/collab/src/tests/integration_tests.rs
+++ b/crates/collab/src/tests/integration_tests.rs
@@ -7065,7 +7065,7 @@ async fn test_remote_git_branches(
     // Also try creating a new branch
     cx_b.update(|cx| {
         repo_b.update(cx, |repository, _cx| {
-            repository.create_branch("totally-new-branch".to_string())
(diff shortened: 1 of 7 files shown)


cole-miller:
Thanks!

PR #39910, opened 2025-10-09
Select to closest character in popover

dunkmann00:
When trying to select text in the popover, when hovering over text in the editor, the character that was getting selected under the mouse cursor didn’t seem “correct”. The behavior was not matching how the selection of text works in the editor. This now aligns the text selection of the markdown text in the popover with the text selection in the editor.

<details><summary>Before:</summary>
<p>

https://github.com/user-attachments/assets/2820194a-d245-4288-a85b-854dbc3a3cea

</p>
</details> 

<details><summary>After:</summary>
<p>

https://github.com/user-attachments/assets/e5b96f42-5d6d-4fe1-9fca-6463107c6650

Notice the character is selected when the cursor is mostly passed it instead of all the way.
</p>
</details> 

Release Notes:

- Nicer selection in popovers

PR #42162, opened 2025-11-07
gpui: Unify the index_for_x methods

mikayla-maki:
Supersedes https://github.com/zed-industries/zed/pull/39910

At some point, these two (`index_for_x` and `closest_index_for_x`) methods where separated out and some code paths used one, while other code paths took the other. That said, their behavior is almost identical:

- `index_for_x` computes the index behind the pixel offset, and returns `None` if there's an overshoot
- `closest_index_for_x` computes the nearest index to the pixel offset, taking into account whether the offset is over halfway through or not. If there's an overshoot, it returns the length of the line.

Given these two behaviors, `closest_index_for_x` seems to be a more useful API than `index_for_x`, and indeed the display map and other core editor features use it extensively. So this PR is an experiment in simply replacing one behavior with the other.

Release Notes:

- Improved the accuracy of mouse selections in Markdown

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -8334,7 +8334,7 @@ impl LineWithInvisibles {
                     let fragment_end_x = fragment_start_x + shaped_line.width;
                     if x < fragment_end_x {
                         return Some(
-                            fragment_start_index + shaped_line.index_for_x(x - fragment_start_x)?,
+                            fragment_start_index + shaped_line.index_for_x(x - fragment_start_x),
                         );
                     }
                     fragment_start_x = fragment_end_x;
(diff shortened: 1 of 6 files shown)


crates/editor/src/element.rs
```diff
@@ -8334,7 +8334,7 @@ impl LineWithInvisibles {
                     let fragment_end_x = fragment_start_x + shaped_line.width;
                     if x < fragment_end_x {
                         return Some(
-                            fragment_start_index + shaped_line.index_for_x(x - fragment_start_x)?,
```
mikayla-maki, line 8337: From what I can tell, this is the only line of code that could be effected by removing the "overshoot returns None" part of the API. However, I believe this case is impossible to reach due to the `x < fragment_end_x` check above it, and so we can safely move forward with this API change.

SomeoneToIgnore (approved):

PR #42167, opened 2025-11-07
project: Remove unnecessary panic

Veykril:
If we are in a remote session with the remote dropped, this path is very much reachable if the call to this function got queued up in a task.

Fixes ZED-124

Release Notes:


ConradIrwin:
sorry about that! looks like debugging code I committed

PR #42176, opened 2025-11-07
Add clear recent files history command

libondev:
Release Notes:

- Added "workspace: clear navigation history" command

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -4041,6 +4041,27 @@ impl NavHistory {
         self.0.lock().mode = NavigationMode::Normal;
     }
 
+    pub fn clear(&mut self, cx: &mut App) -> bool {
+        let mut state = self.0.lock();
+
+        if state.backward_stack.is_empty()
+            && state.forward_stack.is_empty()
+            && state.closed_stack.is_empty()
+            && state.paths_by_item.is_empty()
+        {
+            return false;
+        }
+
+        state.mode = NavigationMode::Normal;
+        state.backward_stack.clear();
+        state.forward_stack.clear();
+        state.closed_stack.clear();
+        state.paths_by_item.clear();
+        state.did_update(cx);
+
+        true
+    }
+
     pub fn pop(&mut self, mode: NavigationMode, cx: &mut App) -> Option<NavigationEntry> {
         let mut state = self.0.lock();
         let entry = match mode {
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -4041,7 +4041,7 @@ impl NavHistory {
         self.0.lock().mode = NavigationMode::Normal;
     }
 
-    pub fn clear(&mut self, cx: &mut App) -> bool {
+    pub fn clear(&mut self, cx: &mut App) {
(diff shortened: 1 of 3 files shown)


crates/workspace/src/workspace.rs
```diff
@@ -1927,6 +1929,46 @@ impl Workspace {
...
+        cx: &mut Context<Workspace>,
+    ) -> Task<()> {
+        let panes = self.panes.clone();
+
+        cx.spawn_in(window, async move |workspace, cx| {
+            let answer = workspace
+                .update_in(cx, |_, window, cx| {
+                    window.prompt(
+                        PromptLevel::Warning,
+                        "Clear Recent Files History",
```
SomeoneToIgnore: This seems excessive: 

* Zed barely has any other actions that invoke such prompts
* A user had already did `cmd/ctrl-p` and typed a few letters to match this command; or pressed a predefined custom keybinding

Also, removing the related code will allow us to make this function synchronous, removing the requirement to `.clone()`, which is also good.

crates/workspace/src/pane.rs
```diff
@@ -4041,6 +4041,27 @@ impl NavHistory {
...
 
+    pub fn clear(&mut self, cx: &mut App) -> bool {
+        let mut state = self.0.lock();
+
+        if state.backward_stack.is_empty()
+            && state.forward_stack.is_empty()
+            && state.closed_stack.is_empty()
+            && state.paths_by_item.is_empty()
+        {
+            return false;
```
SomeoneToIgnore: There's nothing that checks this boolean, so seems that there's no reason to have this extra code?

crates/workspace/src/workspace.rs
```diff
@@ -199,6 +199,8 @@ actions!(
         AddFolderToProject,
         /// Clears all notifications.
         ClearAllNotifications,
+        /// Clears the recent files history.
```
SomeoneToIgnore: This seems to be doing more than clearing the recent files from file finder: it also clears all navigation history.

Let's name this appropriately, `ClearNavigationHistory` then?

Doing this will require to update the rustdoc above, and I would use this to explain more what's actually cleared: something like "This will irreversibly clear all navigation history, recently opened and closed files"?

This will be propagated into autogenerated action docs and might help someone later: https://zed.dev/docs/all-actions?highlight=Clear%20All%20Notifications#all-actions

SomeoneToIgnore (commented):
Thank you, looks nice, let's make it a bit more user-friendly and merge.

I feel that adding a test on this should not be too hard: there are https://github.com/zed-industries/zed/blob/21f73d9c02681152019ed5703ce8808c841fcbbe/crates/file_finder/src/file_finder_tests.rs#L2564-L2569 already, and we can also call `pane: go back` actions.

We could copy the test or extend some to check that the new command works and clears finder entries and disables the file navigation.

libondev:
Thank you for your detailed comments. It gives me a clear idea of ​​what I should do next. I have adjusted the code according to your requirements. If there is anything missing or needs to be adjusted, please tell me and I will complete it.

SomeoneToIgnore (approved):
Thank you!

PR #42198, opened 2025-11-07
agent_ui: Fix thread history item showing GMT time instead of local time on Windows

### LINKED ISSUE #42178: AI: The recording time of the agent session should use the user's local time
### Summary

After the user creates an agent session and then checks the history record, you will find that the time recorded here is GMT time instead of the user's local time. The correct behavior should be to use the time in the user's local time zone as the record.

### Description

Steps to trigger the problem:
1. Complete an AI conversation
2. View the history of conversations with ai

Image

**Expected Behavior**:
**Actual Behavior**:

### Model Provider Details
- Provider: (Anthropic via ZedPro, Anthropic via API key, Copilot Chat, Mistral, OpenAI, etc)
- Model Name:
- Mode: (Agent Panel, Inline Assistant, Terminal Assistant or Text Threads)

Zed: v0.212.1 (Zed Preview) 
OS: Windows 10.0.26100
Memory: 31.7 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics || Intel Corporation || 32.0.101.5972

abulgit:
Closes #42178
Now it's consistent with the DateAndTime path which already does timezone conversion.

- **Future Work**
Happy to tackle the TODO in `time_format.rs` about implementing native Windows APIs for proper localized formatting (similar to macOS's `CFDateFormatter`) as a follow-up.

Release Notes:

- agent: Fixed the thread history item timestamp, which was being shown in GMT instead of in the user's local timezone on Windows.

### DIFF
--- a/crates/agent_ui/src/acp/thread_history.rs
+++ b/crates/agent_ui/src/acp/thread_history.rs
@@ -673,7 +673,7 @@ impl EntryTimeFormat {
                 timezone,
                 time_format::TimestampFormat::EnhancedAbsolute,
             ),
-            EntryTimeFormat::TimeOnly => time_format::format_time(timestamp),
+            EntryTimeFormat::TimeOnly => time_format::format_time(timestamp.to_offset(timezone)),
         }
     }
 }


danilo-leal (approved):
Thank you so much!

PR #42206, opened 2025-11-07
Honor Agent Panel default_view by only replaying serialized agent selection during load

### LINKED ISSUE #42194: AI: Agent Panel ignores agent.default_view on first load
### Summary

AgentPanel always resets to the new thread with native agent view during the initial setup, so the `agent.default_view` setting is ignored unless a previous session serialized a different selected agent.

### Description
Overall, the `agent.default_view` setting is not very useful. Per original design, this setting will change the default new thread view during the initial launch. However, it will be overwritten as soon as you already got a session and being serialized into local with a different selected agent. The `default_view` will never apply if you have serialized agent from previous session.

Steps to trigger the problem:
1. Set "agent": { "default_view": "text_thread" } in settings.
2. Launch Zed with no serialized agent panel state (fresh install or delete local cache).
3. Open the Agent Pane at start

**Expected Behavior**: Panel opens directly in Text Thread mode per the default.
**Actual Behavior**: Panel always opens in New Thread with native agent mode.

### Model Provider Details
- Provider: Anthropic via ZedPro, Copilot Chat, OpenAi
- Model Name: Claude Sonnet 4.5, GPT-5
- Mode: Agent Panel, Text Threads

Zed: v0.211.6 (Zed) 
OS: macOS 15.5
Memory: 16 GiB
Architecture: aarch64

xipeng-jin:
Closes #42194

Reference the description and root cause in #42194 . 

In summary
1. We need to remove the else statement `{ … new_agent_thread … }` branch in order to avoid creating new thread with native agent to overwrite the `agent.default_view` in user settings anyway when there’s no serialized panel state.”
2. We return both `active_view` and associated `selected_agent` in the `new` function, rather than always using `AgentType::default()` in all cases.

Release Notes:
- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -521,10 +521,6 @@ impl AgentPanel {
                         }
                         cx.notify();
                     });
-                } else {
-                    panel.update(cx, |panel, cx| {
-                        panel.new_agent_thread(AgentType::NativeAgent, window, cx);
-                    });
                 }
                 panel.as_mut(cx).loading = false;
                 panel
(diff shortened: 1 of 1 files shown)


agu-z:
@xipeng-jin Thank you for your contribution! Can you explain what specific scenario this fixes? I follow the steps in the issue, but It looks like I still get the native agent after I reopen Zed

PR #42207, opened 2025-11-07
language_models: Add `auto_discover` setting for Ollama

xfbs:
First up: I'm sorry if this is a low quality PR, or if this feature isn't wanted. I implemented this because I'd like to have this behaviour. If you don't think that this is useful, feel free to close the PR without comment. :)

My idea is this: I love to pull random models with Ollama to try them. At the same time, not all of them are useful for coding, or some won't work out of the box with the context_length set. So, I'd like to change Zed's behaviour to not show me all models Ollama has, but to limit it to the ones that I configure manually. 

What I did is add an `auto_discover` field to the settings. The idea is that you can write a config like this:

```json
"language_models": {
    "ollama": {
      "api_url": "http://localhost:11434",
      "auto_discover": false,
      "available_models": [
        {
          "name": "qwen3:4b",
          "display_name": "Qwen3 4B 32K",
          "max_tokens": 32768,
          "supports_tools": true,
          "supports_thinking": true,
          "supports_images": true
        }
      ]
    }
  }
```

The `auto_discover: false` means that Zed won't pick up or show the language models that Ollama knows about, and will only show me the one I manually configured in `available_models`. That way, I can pull random models with Ollama, but in Zed I can only see the ones that I know work (because I've configured them).

The default for `auto_discover` (when it is not explicitly set) is `true`, meaning that the existing behaviour is preserved, and this is not a breaking change for configurations.

Release Notes:

- ollama: Added `auto_discover` setting to optionally limit visible models to only those manually configured in `available_models`

### DIFF
--- a/crates/language_models/src/provider/ollama.rs
+++ b/crates/language_models/src/provider/ollama.rs
@@ -43,6 +43,7 @@ static API_KEY_ENV_VAR: LazyLock<EnvVar> = env_var!(API_KEY_ENV_VAR_NAME);
 #[derive(Default, Debug, Clone, PartialEq)]
 pub struct OllamaSettings {
     pub api_url: String,
+    pub auto_discover: bool,
     pub available_models: Vec<AvailableModel>,
 }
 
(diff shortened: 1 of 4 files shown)


xfbs:
@cla-bot check

bennetbo (approved):
Thank you!

PR #40736, opened 2025-10-20
markdown: Restore horizontal scrollbars for codeblocks

xipeng-jin:
### Summary

Restore the agent pane’s code-block horizontal scrollbar for easier scrolling without trackpad and preserve individual scroll state across multiple code blocks.

### Motivation

Addresses https://github.com/zed-industries/zed/issues/34224, where agent responses with wide code snippets couldn’t be scrolled horizontally in the panel. Previously there is no visual effect for scrollbar to let the user move the code snippet and it was not obviously to use trackpad or hold down `shift` while scrolling. This PR will ensure the user being able to only use their mouse to drag the horizontal scrollbar to show the complete line when the code overflow the width of code block.

### Changes

- Support auto-hide horizontal scrollbar for rendering code block in agent panel by adding scrollbar support in markdown.rs
- Add `code_block_scroll_handles` cache in _crates/markdown/src/markdown.rs_ to give each code block a persistent `ScrollHandle`.
- Wrap rendered code blocks with custom horizontal scrollbars that match the vertical scrollbar styling and track hover visibility.
- Retain or clear scroll handles based on whether horizontal overflow is enabled, preventing leaks when the markdown re-renders.

### How to Test

1. Open the agent panel, request code generation, and ensure wide snippets show a horizontal scrollbar on hover.
3. Scroll horizontally, navigate away (e.g., change tabs or trigger a re-render), and confirm the scroll position sticks when returning.
5. Toggle horizontal overflow styling off/on (if applicable) and verify scrollbars appear or disappear appropriately.

### Screenshots / Demos (if UI change)

https://github.com/user-attachments/assets/e23f94d9-8fe3-42f5-8f77-81b1005a14c8

### Notes for Reviewers

- This is my first time contribution for `zed`, sorry for any code patten inconsistency. So please let me know if you have any comments and suggestions to make the code pattern consistent and easy to maintain.
- For now, the horizontal scrollbar is not configurable from the setting and the style is fixed with the same design as the vertical one. I am happy to readjust this setting to fit the needs.
- Please let me know if you think any behaviors or designs need to be changed for the scrollbar.
- All changes live inside _crates/markdown/src/markdown.rs_; no API surface changes.

Closes #34224 

### Release Notes:

- AI: Show horizontal scroll-bars in wide markdown elements

PR #42211, opened 2025-11-07
Remove Markdown CodeBlock metadata and Custom rendering

xipeng-jin:
Follow up #40736

Clean up `CodeBlockRenderer::Custom` related rendering per the previous PR [comment](https://github.com/zed-industries/zed/pull/40736#issuecomment-3503074893). 
Additional note here:
1.  The `Custom` variant in the enum `CodeBlockRenderer` will become not useful since cleaning all code related to the custom rendering logic.
2. Need to further review the usage of code block `metadata` field in `MarkdownTag::CodeBlock` enum.

I would like to have the team further review my note above so that we can make sure it will be safe to clean it up and will not affect any potential future features will be built on top of it. Thank you!

Release Notes:

- N/A

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -787,7 +787,6 @@ impl Element for MarkdownElement {
         };
         let mut code_block_ids = HashSet::default();
 
-        let mut current_code_block_metadata = None;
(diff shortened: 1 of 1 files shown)


bennetbo (approved):
Thank you! We can always put the code back if we need to. For now it is good to remove it.

PR #42213, opened 2025-11-07
git: Fix git amend crash

### LINKED ISSUE #40885: Git amend overwrites message
### Summary

Using the amend option ignores the amended commit's message and writes a generic default

### Description

When I try to amend a previous commit, the UI properly puts the to-be-amended commit's message in the message input. But when I commit, I see that the message has been replaced with a generic one, like `Updated some_file.txt`. This happens whether I actually edit the message or not. 

The problem does not happen with the Git CLI or other GUI/TUI tools.

There are no visible errors.

Steps to reproduce:
1. Edit some files
2. Stage the edits
3. Make a commit
4. Edit some more
5. Stage the edits
6. Select the "Amend" option from the Git panel
7. (optional) change the commit message
8. Click "Amend" on the Git panel

**Expected Behavior**: The amended commit should have the original/edited message
**Actual Behavior**: The amended commit has a generic message (as if no commit message was given)

Zed: v0.204.5 (Zed) 
OS: macOS 15.6
Memory: 32 GiB
Architecture: aarch64

zeld-a:
Closes #40885

Release Notes:

- Fix crash when invoking `git: amend` from the command palette

### DIFF
--- a/crates/git_ui/src/commit_modal.rs
+++ b/crates/git_ui/src/commit_modal.rs
@@ -1,5 +1,5 @@
 use crate::branch_picker::{self, BranchList};
-use crate::git_panel::{GitPanel, commit_message_editor};
+use crate::git_panel::GitPanel;
 use git::repository::CommitOptions;
 use git::{Amend, Commit, GenerateCommitMessage, Signoff};
 use panel::{panel_button, panel_editor_style};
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks! I can't reproduce a crash when running `git: amend` from the command palette, could you explain in more detail the steps that lead to a crash for you and how this PR fixes the cause?

Veykril:
I think this was https://github.com/zed-industries/zed/pull/42210?

zeld-a:
It does seem like the issue has been fixed on the latest commit.

PR #42231, opened 2025-11-07
bundle: Skip sentry upload for local install on macOS

claytonrcarter:
This is a follow up to #41482. When running `script/bundle-mac`, it will upload debug symbols to Sentry if you have a `$SENTRY_AUTH_TOKEN` set. I happen to have one set, so this script was trying to generate and upload those. Whoops! This change skips the upload entirely if you're running a local install.

Release Notes:

- N/A

### DIFF
--- a/script/bundle-mac
+++ b/script/bundle-mac
@@ -280,7 +280,9 @@ function sign_binary() {
 }
 
 function upload_debug_symbols() {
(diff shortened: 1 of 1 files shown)


P1n3appl3:
Thanks!

PR #35522, opened 2025-08-02
Add rainbow brackets support

GoldStrikeArch:
Closes #5259 

# Rainbow Brackets for Zed Editor

## Overview

This PR implements rainbow bracket highlighting for Zed, addressing issue #5259. The implementation is inspired by the recent Helix PR ([helix-editor/helix#13530](https://github.com/helix-editor/helix/pull/13530)) and provides similar functionality using tree-sitter queries.

## Context

- **Parent Issue:** https://github.com/zed-industries/zed/issues/5259
- **Inspiration:** https://github.com/helix-editor/helix/pull/13530
- **Implementation:** Heavily assisted by Claude Code as I'm not proficient in Rust or familiar with the Zed codebase
- **Motivation:** I'm missing this feature badly and decided to implement it myself

## Implementation Details

The implementation works conceptually the same way as in Helix:
- Uses `rainbow.scm` files to define bracket and scope patterns for each language
- Leverages tree-sitter for syntax-aware bracket matching
- Maintains a scope stack to track nesting levels
- Colors brackets based on their nesting depth

### Key Components

1. **Rainbow Query Files** (`crates/languages/src/*/rainbow.scm`):
   - Define which syntax nodes are scopes (`@rainbow.scope`)
   - Define which tokens are brackets (`@rainbow.bracket`)
   - Support `rainbow.include-children` property for fine-grained control

2. **Core Implementation** (`crates/editor/src/rainbow_brackets.rs`):
   - Uses `BufferSnapshot::matches()` for proper tree-sitter query matching
   - Implements scope tracking algorithm similar to Helix
   - Maps nesting levels to colors (cycling through 10 levels)

### 1. Text Highlighting vs Background Highlighting
**Current:** Using `editor.highlight_background()` which colors the background behind brackets
**Desired:** Need to highlight the actual text color of the brackets themselves

I tried using `editor.highlight_text` but couldn't get it to work. I need help understanding:
- How to properly use text highlighting in Zed
- Whether there's a different API I should be using
- If text highlighting requires a different approach than background highlighting

### 2. TODO Items

- [x] **Create tests** - Need to add comprehensive tests for the rainbow bracket functionality
- [x] **Use text highlight instead of background highlight** - Main blocker, need help with Zed's text highlighting API
- [x] **Make colors configurable from settings** - Currently colors are hardcoded in the implementation

## Code Example

Currently working for JS/TS/Rust (tested manually), Go/Python/Bash/JSON/C/CPP were not tested yet :
```js
[...]

PR #39935, opened 2025-10-10
Add rainbow brackets support

jsfillman:
### **User description**
Closes #5259

Release Notes:

- Added:

## Rainbow Brackets 🌈

Implements per-nesting colorization for `{}`, `[]`, and `()` with an active-pair highlight. Addresses the community’s top request [#5259](https://github.com/zed-industries/zed/issues/5259) with a small, focused change that fits existing editor/GPUI patterns.

#### Summary

- **Feature**: color by nesting depth; subtle emphasis on the active pair

- **Config** (`settings.json`):

  ```
  "rainbow_brackets": {
    "enabled": true,
    "start_hue": 0.0,
    "hue_step": 30.0,
    "max_brackets": 1000
  }
  ```

- **Perf**: throttled cursor updates (10ms), visible-range caching (with padding), edit-count shortcuts, auto-disable >100k chars. No unwraps/panics; bounds-checked.

- **Scope**: contained to editor; no giant refactors; not theme/extension territory.

#### Tests

21 passing tests cover hue math/wrapping, active-pair detection, unmatched/edge cases, caching behavior, and large-file thresholds. ✅

#### Screenshots

Before:
Zed-PreRainbow

After:
Zed-PostRainbow

#### TL;DR

We have rainbow brackets with pair highlighting! 

royalRainbow

PR #42237, opened 2025-11-07
Add rainbow brackets

### LINKED ISSUE #5259: Rainbow brackets
### Check for existing issues

- [X] Completed

### Is your feature request related to a problem?

Having bracket pairs colours helps quickly look at code and understand where brackets/blocks/parens start & end

Rainbow brackets similar to VSCode

https://code.visualstudio.com/blogs/2021/09/29/bracket-pair-colorization

GoldStrikeArch:
Closes #5259

# Rainbow Brackets Support

## Overview

This iteration finishes wiring rainbow brackets into Zed by making the feature user-configurable and expanding coverage to every high-value language in the default bundle. The core idea to re-use tree-sitter from the [earlier attempt](https://github.com/zed-industries/zed/pull/35522) stays the same, but we now gate it behind a real editor setting and ship the necessary tree-sitter captures so users can turn the feature on for C, C++, Go, Python, Rust, TS/TSX, CSS, JSON/JSONC.

## Context

- **Parent Issue:** https://github.com/zed-industries/zed/issues/5259  
- **Inspiration:** https://github.com/helix-editor/helix/pull/13530  
- **Editor Setting:** `editor.rainbow_brackets` (disabled by default so users opt in)  
- **Motivation:** I'm still missing this feature, so after [my failed first attempt](https://github.com/zed-industries/zed/pull/35522) I decided to give it another shot.

### 1. Editor Integration

- Added `rainbow_brackets: bool` to `EditorSettings`, exposed through default settings JSON, the Settings UI, docs, and VS Code import path.
- The editor only refreshes rainbow highlights when the setting is enabled; toggling it off clears every active color immediately.
- Scroll, buffer, and settings events now check the flag before scheduling rainbow work, keeping overhead at zero unless the user opts in.

### 2. Language Query Updates

- Extended `crates/languages/src/*/brackets.scm` for:
- `c`, `cpp`, `css`, `go`, `json`, `jsonc`, `python`, `rust`, `javascript`, `typescript`, `tsx`
- Each file now marks scope nodes (`@rainbow.scope`) and bracket tokens (`@rainbow.bracket`) that match the language’s grammar (parameter lists, composite literals, generics, JSX tags, etc.).

### 3. Documentation & UI

- `assets/settings/default.json`, `docs/src/configuring-zed.md`, and `crates/settings_ui/src/page_data.rs` describe and surface the new toggle.

## Follow-ups

- Performance remains bounded by viewport size, but we should watch memory churn if more languages add very large scope lists.

## How it looks right now:
image 

Release Notes:

- Added an opt-in `editor.rainbow_brackets` setting, so rainbow bracket highlighting can be toggled on/off.
- Added rainbow bracket support to C, C++, CSS, Go, JSON, JSONC, Python, Rust, JavaScript, TypeScript, and TSX using new tree-sitter scope definitions.


SomeoneToIgnore:
At first, I wanted to write something, but then realized it's all been already written in https://github.com/zed-industries/zed/pull/35522#issuecomment-3242104000 and repeating that, esp. the last paragraph, does not make much sense if you did not read that already.

Same ideas about reviewing the current PR: what's the point to write anything if one does not read?

GoldStrikeArch:
@SomeoneToIgnore 

> what's the point to write anything if one does not read?

That's a bit rude you know 😓 

I saw that paragraph:

> Finally, we've heard you all loud and clear: This is the most requested feature and we miss this ourselves. Thus, we finally decided we'll start very soon on adding this to Zed. This might take ourselves some time, but makes sure everything works as you'd expect and that it will come and land. Thanks for understanding and pushing us in that regard – and thanks for understanding.

But, that was written ~ 3 months ago and the feature is not here.
> This might take ourselves some time

Might be another 3 months or 6 months, etc. It is clearly not a prioritized feature. The current solution addressed the flaws from the previous one and *does work exactly at it was intended*. So, please write what you wanted to write initially (the technical aspect, flaws, etc), don't be lazy.

Update:
- NVM, I saw that it is actually WIP [here](https://github.com/zed-industries/zed/pull/39935#issuecomment-3392327595)

PR #42239, opened 2025-11-07
git: Add askpass delegate to git-commit handlers

kubkon:
In my local setup, I always enforce git-commit signing with GPG/SSH which automatically enforces `git commit -S` when committing. This changeset will now show a modal to the user for them to specify the passphrase (if any) so that they can unlock their private key for signing when committing in Zed.

Screenshot 2025-11-07 at 11 09 09 PM

Release Notes:

- Handle automatic git-commit signing by presenting the user with an askpass modal

### DIFF
--- a/crates/git_ui/src/remote_output.rs
+++ b/crates/git_ui/src/remote_output.rs
@@ -6,7 +6,6 @@ use util::ResultExt as _;
 
 #[derive(Clone)]
 pub enum RemoteAction {
-    Commit,
     Fetch(Option<Remote>),
     Pull(Remote),
     Push(SharedString, Remote),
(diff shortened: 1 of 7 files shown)


crates/git_ui/src/remote_output.rs
```diff
@@ -34,6 +36,10 @@ pub struct SuccessMessage {
 
 pub fn format_output(action: &RemoteAction, output: RemoteCommandOutput) -> SuccessMessage {
     match action {
+        RemoteAction::Commit => SuccessMessage {
+            message: "Successfully committed".into(),
+            style: SuccessStyle::ToastWithLog { output },
+        },
```
cole-miller: I think we should probably not have a toast after successfully committing

cole-miller (commented):
Looks good, just one minor comment, plus the concern we discussed about how this works with GPG signing.

kubkon:
> Looks good, just one minor comment, plus the concern we discussed about how this works with GPG signing.

Fair point! I'll investigate GPG signing in a subsequent PR.

Issue #34411, opened 2025-07-14
Bug: Sticky-scroll background transparency makes feature unusable


carlsverre:
### Summary

The background of the new sticky_scroll project panel feature is semi-transparent on my machine.

### Description

Running on Arch Linux with the herbstluftwm and Picom compositor: https://github.com/yshui/picom

```console
$ uname -a
Linux glimmer 6.15.6-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 10 Jul 2025 17:10:18 +0000 x86_64 GNU/Linux

$ zed --version
Zed preview 0.195.0 e306a55073db41d2fbea168ada27f84478392764  – /usr/lib/zed/zed-editor
```

**Expected Behavior**:
The background should not be transparent

**Actual Behavior**:
<img width="241" height="182" alt="Image" src="https://github.com/user-attachments/assets/ab135e27-065b-42dc-843c-520c561c37db" />

### Zed Version and System Specs

Zed: v0.195.0 (Zed Preview) 
OS: Linux X11 arch unknown
Memory: 62.7 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 2060 SUPER || NVIDIA || 575.64.03

smitbarmase:
This is specific to the theme you’re using. Ideally, we don’t want the community to create such themes with transparent hover backgrounds. For now, the workaround is to switch your theme. In the long run, we need to come up with a way to prevent such broken themes from being created.

carlsverre:
Ah! Gotcha. I was using the default dracula theme! Switched to dracula-solid. Thanks!

PR #42242, opened 2025-11-07
editor: Add sticky scroll

### LINKED ISSUE #5344: Pin to code structure while scrolling (aka "sticky scroll")
**Is your feature request related to a problem? Please describe.**
It's easy to get lost while coding on long functions (or classes, Impl blocks, etc..).

**Describe the solution you'd like**
Like in Xcode 14, Zed could pin code structure while scrolling. This feature should be provided with an on/off configuration (I bet not everybody would like it 😁).

**Screenshots**

_Image from https://appcircle.io/blog/whats-new-in-xcode-14-wwdc22/_

timvermeulen:
Closes #5344

https://github.com/user-attachments/assets/37ec58b0-7cf6-4eea-9b34-dccf03d3526b

Release Notes:

- Added a setting to stick scopes to the top of the editor


ConradIrwin:
@timvermeulen Thanks so much for getting the ball rolling here, and for @KyleBarton for getting it over the line!

lnay:
What's the setting name?

KyleBarton:
> What's the setting name?

@lnay it is "sticky_scroll" at the top level, and will be referenced in the docs once released, FYI: https://github.com/zed-industries/zed/pull/42242/files#diff-b12a523dd2cf0ef1cbfb3d716415dea79899b58cf09125d01fd7d608a21096f9R223

acgxv:
A pre-release for this, please 🙏🏻

ConradIrwin:
It'll hit preview on Wednesday and stable next week.

humblehacker:
Nice work! I was so excited for this feature that I built from source just to try it out. I found one issue. In Swift, we can annotate elements (classes, functions, etc.) with for example `@MainActor`. But if the annotation is on a line above rather than on the same line as the element, only the annotation is sticky.

Example:
```swift
@MainActor
class Example {
   ...
}
```

When I scroll up, only `@MainActor` is sticky, when I really expect `class Example` to be sticky.

@timvermeulen Not sure if this is a problem with this change, or the Swift language extensions. If you could let me know, I'll open an issue in the appropriate place.

0x2CA:
Doesn't respect `relative_line_numbers` setting, always shows absolute line numbers

timvermeulen:
> @timvermeulen Not sure if this is a problem with this change, or the Swift language extensions. If you could let me know, I'll open an issue in the appropriate place.

For now it uses the same scopes as the outline modal (and indeed if you click on the `Example` symbol from there, the cursor will be moved to the `@MainActor` line), so it boils down to https://github.com/zed-extensions/swift/blob/main/languages/swift/outline.scm including `@MainActor` as part of the `Example` item.

timvermeulen:
> Doesn't respect `relative_line_numbers` setting, always shows absolute line numbers

I deliberately left out any special handling for that setting from this initial version because you can't always infer from the context whether a relative line number of a sticky header is above or below the cursor, and I wasn't sure if that could potentially be confusing. I don't use relative line numbers myself though so having the sticky headers respect the `relative_line_numbers` setting might very well be the obvious right choice among those that use it 🙂

acgxv:
Similar to https://github.com/zed-industries/zed/issues/34411

Could you please tell me the background colour for the "sticky" part?

It becomes unusable in a transparent theme and I’d like to change it myself.

image

ConradIrwin:
@aurexav It doesn't currently have a theme color set. Are there other similar UI elements ("same background as the editor but solid") we could copy from or do we need to add a new one for this?

acgxv:
> @aurexav It doesn't currently have a theme color set. Are there other similar UI elements ("same background as the editor but solid") we could copy from or do we need to add a new one for this?

I’m not sure we should add a new colour configuration.

However, we could try using the same colour as the file tree sticky scroll first. (#34411 solution)

image

KyleBarton:
Played around with this a bit this morning - the project panel background does seem to work well in a transparent theme:

Screenshot 2025-11-13 at 8 29 14 AM


But looks a little wrong in the non-transparent case:

Screenshot 2025-11-13 at 8 28 43 AM

The right move might be to create a theme setting specifically for the sticky headers background, but I'm not sure how much of a breaking change that would be with existing themes.

ask4fusora:
Can we have a command that'll take us to the context? I noticed we can click to go to the context, so can we make that an action?
Inspired by [nvim-treesitter-context](https://github.com/nvim-treesitter/nvim-treesitter-context), they have this function: [jump-to-context (upward)](https://github.com/nvim-treesitter/nvim-treesitter-context?tab=readme-ov-file#jumping-to-context-upwards).
```lua
vim.keymap.set("n", "[c", function()
  require("treesitter-context").go_to_context(vim.v.count1)
end, { silent = true })
```

ConradIrwin:
@ask4fusora absolutely! Want to send a PR?

pprotas:
Hello, first of all thank you for this feature, it is something I relied on in other editors, and I'm happy that Zed added it.

But I have one concern. For me this sticky scroll feature is supposed to show context around the current method your cursor is in, but Zed's implementation seems to be a bit more naive where the last method you scroll by gets stickied. I think this is actually creates confusion, for example in test files, where you want to see which test case you are in. Instead, Zed shows me a completely unrelated test case in the context.

I can show some examples:

### Zed

You can see the test on line 24 is shown in the sticky context. This just confuses me; why would I want this test to be in the sticky scroll header? I am not editing this test, and I am not editing any nested method that is in this test. The information is not useful to me does the opposite of what the sticky scroll feature is trying to achieve.

image


### Neovim with `nvim-treesitter-context`:

https://github.com/nvim-treesitter/nvim-treesitter-context

Neovim seems to do this properly, where methods that are on the same level as the method where my cursor is don't show in the sticky header:

EDIT: This is probably because of the `mode = 'cursor',  -- Line used to calculate context. Choices: 'cursor', 'topline'` option, Zed has behavior similar to `topline` while `cursor` makes more sense to me

image


### VSCode (wrong)
Actually VSC also has the naive implementation that can cause confusion! I think Zed's implementation is based on VSC, but I think Zed has the opportunity here to actually do this "properly".

EDIT: VSC provides the `editor.stickyScroll.defaultModel` setting which can be set to `foldingProviderModel`. This matches the `cursor` mode from nvim-treesitter-context, and the behavior I would expect.

image

(also related to discussion above, relative line numbers are not supported :()

ConradIrwin:
@pprotas The existing bookmark functionality uses the "cursor" style, while the new sticky headers use "topline" by default. I think that's the right default in both cases, but if you felt up to adding a setting for this, that seems reasonable too.

acardonna:
@humblehacker Hey there! Java had the same issue (and probably some other languages too).  As @timvermeulen mentioned, this is something you need to handle in the outline queries in the `outline.scm` file of your extension. In Java's case, we had to make sure the `@item` capture is not including annotations by capturing the whole class declaration but the body instead. And if there's something capturing the modifiers, you'll probably want to make sure annotations aren't being inlcluded in there either. Take a look at [the implementation](https://github.com/zed-extensions/java/pull/136) for Java if you want, it may help you a bit.

trkoch:
What determines the number of "sticky" lines? Can it be configured?

xdBronch:
are there any plans to use a query file separate from outline in the future (maybe with outline as a fallback)? ime from other editors most of the usefulness of sticky scroll is in deeply nested ifs and switches and the outline will only really give functions and types

PR #42244, opened 2025-11-08
Search tab-switcher

perryqh:
What?
----------
- Replace fuzzy search with contains-based filtering
- Store optional string_match and use highlighted labels
- Update placeholder to show Search all tabs… or Search tabs…
- Adjust tab item rendering to support highlighted matches
Screenshot 2025-11-07 at 20 59 18

Why?
------------
I usually have many tabs open and I'd like to search for them from the `tab switcher:toggle`

Release notes
-------------
`tab switcher:toggle` now supports fuzzy search

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16903,7 +16903,6 @@ dependencies = [
  "serde",
  "serde_json",
  "settings",
- "smol",
  "theme",
  "ui",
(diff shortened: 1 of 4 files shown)


perryqh:
Looks like toggle:all already has search!

PR #42258, opened 2025-11-08
terminal: Forward Ctrl+V when clipboard contains images

kingsword09:
When running Codex CLI, Claude Code, or other TUI agents in Zed’s terminal, pasting images wasn’t supported — Zed
treated all clipboard content as plain text and simply pushed it into the PTY, so the agent never saw the image data.
This change makes terminal pastes behave like they do in a native terminal: if the clipboard contains an image, Zed now emits a raw Ctrl+V to the PTY so the agent can read the system clipboard itself.

Release Notes:

- Fixed terminal-launched Codex/Claude sessions by forwarding Ctrl+V for clipboard images so agents can attach them


kingsword09:
@cla-bot check

agu-z:
@kingsword09 Looks great, thank you!

PR #42259, opened 2025-11-08
gpui: Increase app shutdown timeout for on_app_quit handlers

### LINKED ISSUE #42077: Should close all the language server after that instance of zed is closed.
### Summary

I opened three zed instance.

1 - with php laravel
2 - typescript angular
3 - python flask

i already closed 2 and 3 instance of zed. but all the language server is still running. and active and consuming the memory.

Image

Zed: v0.211.4 (Zed) 
OS: Windows 10.0.19045
Memory: 9.9 GiB
Architecture: x86_64
GPU: Intel(R) HD Graphics || Intel Corporation || 10.18.10.5059

cppcoffee:
Language servers can take up to 5s to finish their shutdown handshake, but gpui only waited 100ms before dropping on_app_quit futures. Bump SHUTDOWN_TIMEOUT to 5s (and update the doc comment) so LspStore’s quit hook has enough time to let LanguageServer::shutdown() reach its kill-on-timeout logic, preventing orphaned language server processes on Windows.

Closes https://github.com/zed-industries/zed/issues/42077

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -53,7 +53,7 @@ mod entity_map;
 mod test_context;
 
 /// The duration for which futures returned from [Context::on_app_quit] can run before the application fully quits.
-pub const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
+pub const SHUTDOWN_TIMEOUT: Duration = Duration::from_millis(100);
 
 /// Temporary(?) wrapper around [`RefCell<App>`] to help us debug any double borrows.
 /// Strongly consider removing after stabilization.
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -694,7 +694,7 @@ impl App {
     }
 
     /// Quit the application gracefully. Handlers registered with [`Context::on_app_quit`]
-    /// will be given up to 5 seconds to complete before exiting.
+    /// will be given up to `SHUTDOWN_TIMEOUT` to complete before exiting.
     pub fn shutdown(&mut self) {
         let mut futures = Vec::new();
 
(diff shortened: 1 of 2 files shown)


cppcoffee:
Increase SHUTDOWN_TIMEOUT to 5 s to align with the internal SERVER_SHUTDOWN_TIMEOUT in the language server (also 5 s, crates/lsp/src/lsp.rs:933-974), ensuring that LanguageServer::shutdown() can reach the final child.kill(); otherwise, the old limit of 100 ms will interrupt it before the process actually exits.

altaic:
What's causing such a long shutdown time? An upper bound of 5 seconds is kinda nuts.

crates/gpui/src/app.rs
```diff
@@ -694,7 +694,7 @@ impl App {
     }
 
     /// Quit the application gracefully. Handlers registered with [`Context::on_app_quit`]
-    /// will be given 100ms to complete before exiting.
+    /// will be given up to 5 seconds to complete before exiting.
```
tidely:
```suggestion
    /// will be given up to `SHUTDOWN_TIMEOUT` to complete before exiting.
```

This way we don't need to update the comment when the value changes.

cppcoffee:
> What's causing such a long shutdown time? An upper bound of 5 seconds is kinda nuts.

The application layer's on_app_quit callback has an execution budget of only 100ms (crates/gpui/src/app.rs:696-715), causing the shutdown_language_servers_on_quit to be terminated by the main loop before it can finish. In other words, we are forcibly cut off even before we can make use of that 5-second protection.

To exit faster, there are two paths: either relax the 100ms budget for on_app_quit to allow the 5-second protection to take effect, or give up waiting for the server response and kill it directly.

cppcoffee:
Or should we modify the LSP server's SERVER_SHUTDOWN_TIMEOUT from 5s to 100ms, making it the same as SHUTDOWN_TIMEOUT?

https://github.com/zed-industries/zed/blob/main/crates/gpui/src/app.rs#L56

https://github.com/zed-industries/zed/blob/main/crates/lsp/src/lsp.rs#L49

PR #42260, opened 2025-11-08
acp: Fix issue with mentions when `embedded_context` is set to `false`

bennetbo:
Release Notes:

- acp: Fixed an issue where Zed would not respect `PromptCapabilities::embedded_context`


benbrandt (approved):
Thanks!

PR #32795, opened 2025-06-16
editor: Ensure mouse cursor is shown again on mouse move

MrSubidubi:
Closes #32787
Follow-up to #27519 and #32408 

This PR fixes an issue where the mouse cursor would stay hidden after typing in the editor.

Before #32408, we would rerender the editor on every mouse move. Now, we (correctly) only do this if a rerender is actually required. This caused a small regression for hiding the mouse cursor though: Due to the view now being cached, we do not neccessarily update the mouse cursor style so it is shown again. The boolean is updated but the view is not, resulting in the cursor style being kept until another action is performed. This is an issue with both Stable and Preview (due to some other changes, the issue is slightly worse on Preview though, see https://github.com/zed-industries/zed/pull/32596#issuecomment-2969258800 and https://github.com/zed-industries/zed/pull/32596#issuecomment-2969357248 for some more context).

This PR ensures that the cursor is shown again by scheduling a redraw of the editor whenever the boolean is updated.

The change should not cause any performance regressions: In most cases where we want to hide the mouse, the editor is about to be rerendered anyway, hence this would not change anything. For cases where we want to show the cursor again, this ensures that we actually end up doing so by rerendering the editor once.

Release Notes:

- Fixed an issue where the mouse cursor would sometimes stay hidden after typing in editors with the `hide_mouse` setting enabled.

PR #42262, opened 2025-11-08
editor: Resolve disappearing cursor on panel open

### LINKED ISSUE #35817: Mouse cursor stays hidden when opening panels with an auto-focused filter input
When using the status bar icon button to access the collab panel, the cursor disappears.

## Steps to Reproduce:

1. Click the status bar icon to open the collab panel
2. Observe that the cursor is no longer visible

**Expected Behavior**:
Cursor should remain visible after opening the collab panel.

**Actual Behavior**:
Cursor disappears from view after opening the collab panel and does not reappear.

Zed: v0.200.0 (Zed Nightly 55b4df4d9f1922785879de268fb4e4fe7d824d4b) 
OS: macOS 15.5.0
Memory: 18 GiB
Architecture: aarch64

errmayank:
Closes #35817

Release Notes:

- Fixed cursor disappearing on panel open

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -1564,9 +1564,17 @@ impl Workspace {
                     return;
                 }
 
-                with_active_or_new_workspace(cx, |workspace, _, cx| {
-                    workspace.hide_mouse_cursor(cx);
-                });
+                cx.active_window()
+                    .and_then(|w| w.downcast::<Workspace>())
+                    .map(|workspace| {
+                        cx.defer(move |cx| {
+                            workspace
+                                .update(cx, |workspace, _window, cx| {
+                                    workspace.hide_mouse_cursor(cx)
+                                })
+                                .log_err();
+                        });
+                    });
             }),
         ];
 
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -1556,7 +1556,7 @@ impl Workspace {
                     store.workspaces.remove(&window_handle);
                 })
             }),
-            cx.intercept_keystrokes(|event, _, cx| {
+            cx.intercept_keystrokes(|event, window, cx| {
                 if event.keystroke.modifiers.platform
                     || event.keystroke.modifiers.control
                     || event.keystroke.modifiers.alt
@@ -1564,17 +1564,11 @@ impl Workspace {
                     return;
                 }
 
-                cx.active_window()
-                    .and_then(|w| w.downcast::<Workspace>())
-                    .map(|workspace| {
-                        cx.defer(move |cx| {
-                            workspace
-                                .update(cx, |workspace, _window, cx| {
-                                    workspace.hide_mouse_cursor(cx)
-                                })
-                                .log_err();
-                        });
+                if let Some(workspace) = window.root::<Workspace>().flatten() {
+                    cx.defer(move |cx| {
+                        workspace.update(cx, |workspace, cx| workspace.hide_mouse_cursor(cx))
                     });
+                };
             }),
         ];
 
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -1205,6 +1205,7 @@ pub struct Workspace {
     removing: bool,
     utility_panes: UtilityPaneState,
     next_modal_placement: Option<ModalPlacement>,
+    mouse_cursor_hidden: bool,
 }
 
 impl EventEmitter<Event> for Workspace {}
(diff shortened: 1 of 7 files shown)


MrSubidubi (changes requested):
Thanks for this!

Unfortunately, I do not think this is the proper fix - this will not hide the cursor in cases where it should actually be hidden.

The issue then is, though, that this will require a LOT more changes to get properly right. As of now, this guard has to be manually added for each and every editor and be updated accordingly, set with the correct enum variant. Additionally, we also had to add a special hitbox behavior for this specific change (I can search for that PR later). Yet, as seen here, this is really prone to errors and should be resolved differently IMO. 

I think we should instead push the handling of this setting up to the workspace. Then, we can differentiate between movement and typing more easily. However, this will require a larger refactor some more changes to the backend on Linux I think, I'll check with some of the folks internally on this. 

Hence, I do not think this is the proper fix for the issue at hand. Would you be up for the bigger scope? Happy to get that working together with you, but as said, that might be a larger chunk of work.

errmayank:
@MrSubidubi Makes sense, happy to take this on!

So from what I understand, we need to get rid of all the `hide_mouse_cursor()` calls in editor and handle it at workspace level, maybe via an action?

> Additionally, we also had to add a special hitbox behavior for this specific change (I can search for that PR later)

Let me know what PR this was, appreciate it.

Maybe we can later pair to make sure we're on the same page.

MrSubidubi:
> So from what I understand, we need to get rid of all the hide_mouse_cursor() calls in editor and handle it at workspace level, maybe via an action?

Yes, but definitely not via an action. I think we could have a listener at the workspace level via 
https://github.com/zed-industries/zed/blob/e833d1af8d8ba7b3445b07df7356108886cb40aa/crates/gpui/src/app.rs#L1691 which hides for the `on_input` part of the hiding, whereas for the movement/action part, I lack a good idea right now (perhaps you find one).

Then, whenever we detect a key dispatch, we hide the cursor.

I already pinged people internally about this, so please note that we might actually need some new APIs for this to be properly fixed. But given that you are on MacOS, should actually be fine for you right now (you can dig through the PR that follows for some context, happy to explain that some other time).

PR in question was https://github.com/zed-industries/zed/pull/32795 and related ones.

So, feel free to start on this. However, please do also note that, as noted, this might require some adjacent changes, so could take some more time before I can merge this. Thanks!

MrSubidubi:
Marked as draft just so it is clearer that this will take some time, no worries about it please, just for my personal bookkeeping 🙂

franciskafyi:
Hey @errmayank, let us know when this is ready for another review. We're doing a big PR review push next week ahead of the break and we'd love to include this one, if it's ready!

errmayank:
Sure, will do @franciskafyi :)

I just picked it back up, shouldn't take too long if everything goes well.

crates/workspace/src/workspace.rs
```diff
@@ -1555,6 +1556,32 @@ impl Workspace {
...
+                let is_undo_redo = event.keystroke.modifiers.platform
+                    && (event.keystroke.key == "z" || event.keystroke.key == "Z");
+
+                if !is_undo_redo
+                    && (event.keystroke.modifiers.platform
+                        || event.keystroke.modifiers.control
+                        || event.keystroke.modifiers.alt)
+                {
+                    return;
+                }
```
MrSubidubi, line 1570: I suspect you want to not hide the cursor on Undo actions? Curios why you would not want that?

crates/workspace/src/workspace.rs
```diff
@@ -8629,7 +8671,20 @@ pub fn client_side_decorations(
...
+                            workspace
+                                .update(cx, |workspace, _window, cx| {
+                                    if workspace.mouse_cursor_hidden {
+                                        workspace.mouse_cursor_hidden = false;
+                                        cx.notify();
+                                    }
+                                })
+                                .log_err();
+                        });
+                    }
```
MrSubidubi, line 8686: Can we not get rid of this entirely if we just `mem::take` the bool above? I'd personally think that we add a `hide_mouse_cursor_until_mouse_move` API to the window (I'll try to get this up soon) so that we do not need to do this here.

But probably fine to keep for the moment until this is in.

crates/workspace/src/workspace.rs
```diff
@@ -7455,7 +7482,22 @@ impl Render for Workspace {
...
+                                    window.insert_hitbox(bounds, HitboxBehavior::Normal)
+                                },
+                                move |_bounds, hitbox, window, _cx| {
+                                    if mouse_cursor_hidden {
+                                        window.set_cursor_style(CursorStyle::None, &hitbox);
+                                    }
+                                },
+                            )
+                            .absolute()
+                            .size_full()
```
MrSubidubi, line 7499: Sorry if I am missing something, but would something like

```rust
if self.mouse_cursor_hidden {
    window.set_window_cursor_style(CursorStyle::None)
 }
```

not suffice here?

MrSubidubi (commented):
Left some notes. As for the flickering, this might very well correspond to the issues I mentioned above, as we are really not properly differentiating some stuff here. I'll take another look soon.

SomeoneToIgnore:
Closing as stale for now as the last activity was in the previous year.

image

PR #42263, opened 2025-11-08
Slightly improve completion in settings.json (for lsp.<language-server>.)

olejorgenb:
Document "any-typed" (`serde_json::Value`) "lsp" keys to include them in json-language-server completions.

The vscode-json-languageserver seems to skip generically typed keys when offering completion.

For this schema

```
    "LspSettings": {
        "type": "object",
        "properties": {
            ...
            "initialization_options": true,
            ...
         }
     }
```

"initialization_options" is not offered in the completion.

The effect is easy to verify by triggering completion inside:

```
    "lsp": {
        "basedpyright": {
           COMPLETE HERE
```

image

By adding a documentation string the keys are offered even if they are generically typed:

image

---

Note: I did some cursory research of whether it's possible to make vscode-json-languageserver change behavior without success. IMO, not offering completions here is a bug (or at minimal should be configurable)

---

Release Notes:

- N/A

### DIFF
--- a/crates/settings/src/settings_content/project.rs
+++ b/crates/settings/src/settings_content/project.rs
@@ -110,7 +110,7 @@ pub struct LspSettings {
     pub binary: Option<BinarySettings>,
     /// Options passed to the language server at startup.
     ///
-    /// Consult the documentation for the specific LSP to see what settings
+    /// Consult the documentation for the specific language server to see what settings
     /// are supported
     pub initialization_options: Option<serde_json::Value>,
     /// Language server settings
--- a/crates/settings/src/settings_content/project.rs
+++ b/crates/settings/src/settings_content/project.rs
@@ -108,7 +108,7 @@ pub struct WorktreeSettingsContent {
 #[serde(rename_all = "snake_case")]
 pub struct LspSettings {
     pub binary: Option<BinarySettings>,
-    /// Options passed to the language server at startup
+    /// Options passed to the language server at startup.
     ///
     /// Consult the documentation for the specific LSP to see what settings
     /// are supported
--- a/crates/settings/src/settings_content/project.rs
+++ b/crates/settings/src/settings_content/project.rs
@@ -108,7 +108,15 @@ pub struct WorktreeSettingsContent {
 #[serde(rename_all = "snake_case")]
 pub struct LspSettings {
     pub binary: Option<BinarySettings>,
+    /// Options passed to the language server at startup
+    ///
+    /// Consult the documentation for the specific LSP to see what settings
+    /// are supported
     pub initialization_options: Option<serde_json::Value>,
+    /// Language server settings
+    ///
+    /// Consult the documentation for the specific LSP to see what settings
+    /// are supported
     pub settings: Option<serde_json::Value>,
     /// If the server supports sending tasks over LSP extensions,
     /// this setting can be used to enable or disable them in Zed.
--- a/crates/settings/src/settings_content/project.rs
+++ b/crates/settings/src/settings_content/project.rs
@@ -26,6 +26,7 @@ pub struct ProjectSettingsContent {
     /// The following settings can be overridden for specific language servers:
     /// - initialization_options
     ///
+    ///
     /// To override settings for a language, add an entry for that language server's
     /// name to the lsp value.
     /// Default: null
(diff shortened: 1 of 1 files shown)


crates/settings/src/settings_content/project.rs
```diff
@@ -108,7 +108,15 @@ pub struct WorktreeSettingsContent {
 #[serde(rename_all = "snake_case")]
 pub struct LspSettings {
     pub binary: Option<BinarySettings>,
+    /// Options passed to the language server at startup
```
SomeoneToIgnore:
```suggestion
    /// Options passed to the language server at startup.
```

Let's:

* use docs in the end of the first doc sentence, just to be on par with the style around
* consider adding relevant LSP spec links to both entries: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize and https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration

Hopefully, that won't bloat the completion pop-up?

crates/settings/src/settings_content/project.rs
```diff
@@ -108,7 +108,15 @@ pub struct WorktreeSettingsContent {
...
+    ///
+    /// Consult the documentation for the specific LSP to see what settings
```
SomeoneToIgnore:
```suggestion
    /// Consult the documentation for the specific language server to see what settings
```

LSP stands for Language Server Protocol, it's The One and Only, quite specific already.

SomeoneToIgnore (commented):
> not offering completions here is a bug

I agree and would be interesting to get down to this.
Getting more docs in never hurts, so let's add those too.

I think it's also worth checking `default.json` and one of the `configuring-*.md` documents, so people have this information in any visible place: website docs, default settings and completion entries' docs.

olejorgenb:
Should I replicate the keys here as well:

```
    /// Configuration for language servers.
    ///
    /// The following settings can be overridden for specific language servers:
    /// - initialization_options
    ///
    /// To override settings for a language, add an entry for that language server's
    /// name to the lsp value.
    /// Default: null
    #[serde(default)]
    pub lsp: HashMap<Arc<str>, LspSettings>,
 ```
 
 (Annoyingly github doesn't allow commenting on unchanged code (?!)

crates/settings/src/settings_content/project.rs
olejorgenb: > consider adding relevant LSP spec links to both entries

I personally appreciate such conveniences, and in this case it makes it crystal clear exactly what the options does under the hood.

image

Link is not clickable though.

Another quirk is that the json language server does not seem to rewrap text (the "are supported."), but I'm not sure to what extent it makes sense to cater to such specifics :thinking:

olejorgenb:
> I think it's also worth checking default.json and one of the configuring-*.md documents

The `default.json` lack multiple of the available keys for the lsp settings, but is it supposed to be exhaustive?

crates/settings/src/settings_content/project.rs
SomeoneToIgnore: > Link is not clickable though.

IIRC, the pop-ups are rendered as Markdown, so what if we use its link syntax?

Rewrap issues are not very related to documentation improvements and there are issues and PRs for that already.

SomeoneToIgnore (commented):
> but is it supposed to be exhaustive

In some form that allows to document features, but it's not feasible to e.g. add all language server names into there as we do not know which are created or deprecated later.

I think this is quite a step forward already, esp. if we try the links.

crates/settings/src/settings_content/project.rs
olejorgenb: I can't get link markdown link syntax to render :(

olejorgenb:
> and one of the configuring-*.md documents

It's well documented here: https://zed.dev/docs/configuring-languages#possible-configuration-options including links to the LSP spec. 

That link would be the most user-friendly thing to include IMO. (but risks future linkrot ofc.)

> In some form that allows to document features, but it's not feasible to e.g. add all language server names into there as we do not know which are created or deprecated later.

I would suggest adding a link to https://zed.dev/docs/configuring-languages#configuring-language-servers 

---

I think we're sorta hitting diminishing returns though :) ? IMO the primary value is to avoid the completion listing _some_ keys, but not all. I remember this confused me earlier.

SomeoneToIgnore:
> diminishing returns

Oh, more like chunking the work: updating the docs alone is a great chunk of work for starters.

Making completions to function better, more predictable, is also a great improvement to do.

I lack the context on this part of the codebase, alas, so cannot give any specific pointers.

SomeoneToIgnore (approved):

olejorgenb:
Thanks for reviewing and merging :smiley: 

I forgot to push these minor fixes: 
[terminology-and-extra-whitespace-fixup.patch](https://github.com/user-attachments/files/23465975/terminology-and-extra-whitespace-fixup.patch) (Feel free to just commit directly)

SomeoneToIgnore:
https://github.com/zed-industries/zed/pull/42397

PR #42265, opened 2025-11-08
markdown: Add support for `HTML` href elements

RemcoSmitsDev:
This PR adds support for `HTML` href elements. It also refactored the way we stored the regions, this was done because otherwise I had to add 2 extra arguments to each `HTML` parser method. It's now also more inline with how we have done it for the highlights.

**Small note**: the markdown parser only supports HTML href tags inside a paragraph tag. So adding them as a root node will result in just showing the inner text. This is a limitation of the markdown parser we use itself.

**Before**
Screenshot 2025-11-08 at 15 40 28

**After**
Screenshot 2025-11-08 at 15 29 55

**Example code**
```markdown
<p>asd <a href="https://example.com">Link Text</a> more text</p>
<p><a href="https://example.com">Link Text</a></p>

[Duck Duck Go](https://duckduckgo.com)
```

**TODO**:
- [x] Add tests

cc @bennetbo

Release Notes:

- Markdown Preview: Add support for `HTML` href elements.

### DIFF
--- a/crates/markdown_preview/src/markdown_parser.rs
+++ b/crates/markdown_preview/src/markdown_parser.rs
@@ -245,8 +245,7 @@ impl<'a> MarkdownParser<'a> {
         let mut strikethrough_depth = 0;
         let mut link: Option<Link> = None;
         let mut image: Option<Image> = None;
-        let mut region_ranges: Vec<Range<usize>> = vec![];
-        let mut regions: Vec<ParsedRegion> = vec![];
+        let mut regions: Vec<(Range<usize>, ParsedRegion)> = vec![];
         let mut highlights: Vec<(Range<usize>, MarkdownHighlight)> = vec![];
         let mut link_urls: Vec<String> = vec![];
         let mut link_ranges: Vec<Range<usize>> = vec![];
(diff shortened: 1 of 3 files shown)


RemcoSmitsDev:
@bennetbo It's ready for your review.

crates/markdown_preview/src/markdown_parser.rs
```diff
@@ -966,6 +982,17 @@ impl<'a> MarkdownParser<'a> {
                 .collect()
         }
 
+        fn correct_region_range(
```
bennetbo: Wondering if we can make this work similar to `add_highlight_range` above, or potentially having a single function which is generic, like `Vec<(Range<usize>, T)>`

RemcoSmitsDev: Yeah, we can do that, it is a nice one to clean it up! Fixed it, and thanks for the review!

bennetbo:
Pushed a small change that removes a bunch of `Vec` and `String` allocations we did not need. Should be good to merge now.

bennetbo (approved):

PR #42266, opened 2025-11-08
Improve error messages on extension loading

liuyanghejerry:
This pull request improves error message when extension loading goes wrong.

Before:

```
2025-11-08T21:16:02+08:00 ERROR [extension_host::extension_host] failed to load arkts extension.toml

Caused by:
    No such file or directory (os error 2)
```

Now:

```
2025-11-08T22:57:00+08:00 ERROR [extension_host::extension_host] failed to load arkts extension.toml, "/Users/user_name_placeholder/Library/Application Support/Zed/extensions/installed/arkts/extension.toml"

Caused by:
    No such file or directory (os error 2)

```

Release Notes:

- N/A

### DIFF
--- a/crates/extension/src/extension_manifest.rs
+++ b/crates/extension/src/extension_manifest.rs
@@ -279,10 +279,7 @@ impl ExtensionManifest {
         } else {
             extension_manifest_path.set_extension("toml");
             let manifest_content = fs.load(&extension_manifest_path).await.with_context(|| {
-                format!(
-                    "failed to load {extension_name} extension.toml, {:?}",
-                    extension_manifest_path
-                )
+                format!("loading {extension_name} extension.toml, {extension_manifest_path:?}")
             })?;
             toml::from_str(&manifest_content).map_err(|err| {
                 anyhow!("Invalid extension.toml for extension {extension_name}:\n{err}")
--- a/crates/extension/src/extension_manifest.rs
+++ b/crates/extension/src/extension_manifest.rs
@@ -270,7 +270,7 @@ impl ExtensionManifest {
             let manifest_content = fs
                 .load(&extension_manifest_path)
                 .await
-                .with_context(|| format!("failed to load {extension_name} extension.json"))?;
+                .with_context(|| format!("failed to load {extension_name} extension.json, {:?}", extension_manifest_path))?;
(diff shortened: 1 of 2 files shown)


crates/extension/src/extension_manifest.rs
```diff
@@ -267,10 +267,12 @@ impl ExtensionManifest {
 
         let mut extension_manifest_path = extension_dir.join("extension.json");
         if fs.is_file(&extension_manifest_path).await {
-            let manifest_content = fs
-                .load(&extension_manifest_path)
-                .await
-                .with_context(|| format!("failed to load {extension_name} extension.json"))?;
+            let manifest_content = fs.load(&extension_manifest_path).await.with_context(|| {
+                format!(
+                    "failed to load {extension_name} extension.json, {:?}",
```
SomeoneToIgnore:
```suggestion
                    "loading {extension_name} extension.json, {extension_manifest_path:?}"
```

Let's do something like this and throw away parasite `failed to` as we're in the error context already, of course we've failed.

Some other places below would benefit from the same.

crates/extension_host/src/wasm_host.rs
```diff
@@ -759,16 +759,15 @@ impl WasmExtension {
...
         let path = extension_dir.join("extension.wasm");
 
-        let mut wasm_file = wasm_host
-            .fs
-            .open_sync(&path)
-            .await
-            .context("failed to open wasm file")?;
+        let mut wasm_file = wasm_host.fs.open_sync(&path).await.context(format!(
+            "failed to open wasm file, path: {}",
+            path.display()
```
SomeoneToIgnore: Let's use `{path:?}` as the code above does, for example: this way we'll get `""` around the path to display it, as most of the code in Zed should do already.

crates/extension/src/extension_manifest.rs
liuyanghejerry: Good advice. Fixed.

crates/extension_host/src/wasm_host.rs
liuyanghejerry: Done.

SomeoneToIgnore (approved):
Thank you!

PR #42267, opened 2025-11-08
shell_env: Wrap error context in format! where missing

kubkon:
Release Notes:

- N/A


Veykril:
Sounds like we should enable https://rust-lang.github.io/rust-clippy/master/index.html#literal_string_with_formatting_args

PR #42269, opened 2025-11-08
copilot: Add usage metrics to context menu

AlvaroParker:
Closes #ISSUE

Preview:

|Context menu|Settings panel|
|-|-|
|image|image|

I also added the usage on GitHub Copilot settings, since users may not be using copilot as their edit prediction provider but they may still want to see the usage. 

Release Notes:

- Added usage stats to copilot context menu
- Added usage stats on GitHub Copilot settings

### DIFF
--- a/crates/copilot/src/copilot.rs
+++ b/crates/copilot/src/copilot.rs
@@ -308,6 +308,7 @@ pub struct Copilot {
     buffers: HashSet<WeakEntity<Buffer>>,
     server_id: LanguageServerId,
     _subscription: gpui::Subscription,
+    usage_task: Option<gpui::Task<()>>,
 }
 
 pub enum Event {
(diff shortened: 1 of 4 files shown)


ghost:
Is there a way to implement a usage metric for each agent?

niekdomi:
Not sure if this is related, but vscode also shows how "expensive" each model is e.g., 1x or 0.3x. Would be nice if this could be added to :-)

AlvaroParker:
I don't think adding the multiplier belongs to this PR, but it should not be too complicated to implement since that's already requested when getting the models from github in:

https://github.com/zed-industries/zed/blob/5d08c1b35f937f1a3e6cb63537d8a1ebb6bac9c9/crates/copilot/src/copilot_chat.rs#L696-L727

PR #42275, opened 2025-11-08
ollama: Fix tool calling

tidely:
Closes #42303

Ollama added tool call identifiers (https://github.com/ollama/ollama/pull/12956) in its latest version [v0.12.10](https://github.com/ollama/ollama/releases/tag/v0.12.10). This broke our json schema and made all tool calls fail.

This PR fixes the schema and uses the Ollama provided tool call identifier when available. We remain backwards compatible and still use our own identifier with older versions of Ollama. I added a `TODO` to remove the `Option` around the new field when most users have updated their installations to v0.12.10 or above.

Note to reviewer: The fix to this issue should likely get cherry-picked into the next release, since Ollama becomes unusable as an agent without it.

Release Notes:

- Fixed tool calling when using the latest version of Ollama

### DIFF
--- a/crates/language_models/src/provider/ollama.rs
+++ b/crates/language_models/src/provider/ollama.rs
@@ -20,6 +20,7 @@ pub use settings::OllamaAvailableModel as AvailableModel;
 use settings::{Settings, SettingsStore, update_settings_file};
 use std::pin::Pin;
 use std::sync::LazyLock;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::{collections::HashMap, sync::Arc};
 use ui::{ButtonLike, ElevationIndex, List, Tooltip, prelude::*};
 use ui_input::InputField;
@@ -381,7 +382,7 @@ impl OllamaLanguageModel {
                             }
                             MessageContent::ToolUse(tool_use) => {
                                 tool_calls.push(OllamaToolCall {
-                                    id: tool_use.id.to_string(),
+                                    id: Some(tool_use.id.to_string()),
                                     function: OllamaFunctionCall {
                                         name: tool_use.name.to_string(),
                                         arguments: tool_use.input,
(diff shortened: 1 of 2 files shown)


js-0s:
@tidely great work! your PR is more simplistic (not introducting logging) and as you mention not backwards compatible as the id is not optional. i think i'll add a test like you did so it covers both types of possible inputs.
Maybe we should add a issue and link it to both PR's so zed-maintainers can decide how to proceed with that.

May i ask how - if you did not have logging - you discovered that? I must admit that i had a really hard time to figure out what was going on.

tidely:
> not introducting logging

I think the error already gets logged in high level code after the error bubbles up. Not sure how useful the errors are anyway, I think it might be worth exploring using the [serde_path_to_error](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/) crate already used in other places in the Zed codebase. We should definitely merge our efforts on this though.

> Maybe we should add a issue and link it to both PR's so zed-maintainers can decide how to proceed with that.

Sounds good!

> May i ask how - if you did not have logging - you discovered that?

I compared the request schema used in tests to the one returned by curling the newest ollama version, and then I did some digging in the Ollama repo

js-0s:
@tidely note sure how to effectively continue. i think the issue you referenced is good.

i focused on adding logging (debug&trace) for the ollama module which is great when you want to see how the agent interacts but effectively this is not really related to the issue.
So my PR should be `ollama: add logging` while yours should just add the optional fields and both tests in order to be backwards compatible.

tidely:
As per your recommendation @js-0s, I've made the PR backwards compatible with older Ollama versions and added a test for this behavior.

Initially I thought we should just embrace the new response format and ditch older Ollama installations, as we haven't really concerned ourselves with Ollama backwards compatibility before either. It would've allowed us to remove some custom logic we use for our own internal tool calling identifiers and brought Ollama more in unison with other providers. But I figured some well documented TODO's and comments will allows us to strip those bits away later as well.

Thanks for your input here and your efforts with your own PR as well!

bennetbo (approved):
Nice work, thank you!

PR #42289, opened 2025-11-09
gpui: Simplify text system font caching logic

cppcoffee:
Release Notes:

- N/A


as-cii:
Thanks for the pull request, @cppcoffee!

While this is simplifying the code, it also looks like it's changing the behavior such that we don't cache the error when loading the font anymore. I don't think the original code was particularly hard to begin with, so I think we can go ahead and close this. Thanks!

PR #42293, opened 2025-11-09
install_cli: Add user admin check and local bin fallback for CLI installation

### LINKED ISSUE #10065: Make 'Install CLI' usable without admin privileges
Summary: Zed should do something sensible when the CLI can't be installed due to a permissions error

When I attempt to install the CLI without admin privileges, I would either like it to fall back to an approach that uses my home directory (e.g. install to `~/.local/bin` or `~/bin`), or to provide instructions on how to achieve this for myself.

Currently, it errors with:

```
error creating CLI symlink

Caused by:
error running osascript. Please try again.
```

I can install zed using Homebrew into `~/Applications`, and Homebrew manages an appropriate symlink at `/opt/homebrew/bin/zed` for me, so this is a low priority issue for me.

_No response_

leoliu0605:
Closes #10065 

Release Notes:

- Add `admin` check and CLI installation without admin support for `Install CLI`
- Check if the user is in the `admin` group. If not, install CLI to `~/.local/bin`. If yes, attempt installation via `osascript`; if it fails, fall back to installing in `~/.local/bin`.
  Refer to the flowchart below for clarification:
  ```mermaid
  flowchart TB
      Start[Start: install_script] --> CheckAdmin{Check if user<br/>in admin group?}
      
      CheckAdmin -->|No Admin| InstallLocal1[Install to ~/.local/bin]
      CheckAdmin -->|Has Admin| InstallSystem[Try install to /usr/local/bin]
      
      InstallLocal1 --> Success1[✅ Return ~/.local/bin/zed]
      
      InstallSystem -->|Success| Success2[✅ Return /usr/local/bin/zed]
      InstallSystem -->|Failed| Fallback[Fallback to ~/.local/bin]
      
      Fallback --> Success3[✅ Return ~/.local/bin/zed]
      
      Success1 --> End[End]
      Success2 --> End
      Success3 --> End
      
      style CheckAdmin fill:#FDB863,stroke:#E08214,stroke-width:3px,color:#000
      style InstallLocal1 fill:#B2ABD2,stroke:#5E3C99,stroke-width:3px,color:#000
      style Fallback fill:#F4A582,stroke:#D6604D,stroke-width:3px,color:#000
      style Start fill:#B8E186,stroke:#4D9221,color:#000
      style End fill:#B8E186,stroke:#4D9221,color:#000
  ```

https://github.com/user-attachments/assets/5f01daa8-ec81-4b2a-a494-0ab0576c97bc

### DIFF
--- a/crates/install_cli/src/install_cli_binary.rs
+++ b/crates/install_cli/src/install_cli_binary.rs
@@ -16,17 +16,6 @@ actions!(
     ]
 );
 
-async fn is_user_admin() -> bool {
-    let output = smol::process::Command::new("groups").output().await.ok();
-    if let Some(output) = output {
-        if output.status.success() {
-            let groups = String::from_utf8_lossy(&output.stdout);
-            return groups.split_whitespace().any(|g| g == "admin");
-        }
-    }
-    false
-}
-
 async fn install_to_local_bin(cx: &AsyncApp) -> Result<PathBuf> {
     let cli_path = cx.update(|cx| cx.path_for_auxiliary_executable("cli"))??;
     let home_dir = std::env::var("HOME").context("Failed to get HOME environment variable")?;
@@ -101,12 +90,7 @@ async fn install_to_system_bin(cx: &AsyncApp) -> Result<PathBuf> {
 }
 
 async fn install_script(cx: &AsyncApp) -> Result<PathBuf> {
-    let has_admin = is_user_admin().await;
-
-    if !has_admin {
-        return install_to_local_bin(cx).await;
-    }
-
+    // First, try to install to system bin
     match install_to_system_bin(cx).await {
         Ok(path) => Ok(path),
         Err(err) => {
@@ -114,6 +98,7 @@ async fn install_script(cx: &AsyncApp) -> Result<PathBuf> {
                 "Failed to install to system bin: {}. Falling back to local bin.",
                 err
             );
+            // If system install fails, fallback to local bin
             install_to_local_bin(cx).await
         }
     }
@@ -137,10 +122,12 @@ pub fn install_cli_binary(window: &mut Window, cx: &mut Context<Workspace>) {
             .await
             .context("error creating CLI symlink")?;
 
+        let is_local_install = !path.starts_with("/usr/local/bin");
+
         workspace.update_in(cx, |workspace, _, cx| {
             struct InstalledZedCli;
 
-            let message = if path.starts_with("/usr/local/bin") {
+            let message = if !is_local_install {
                 format!(
                     "Installed `zed` to {}. You can launch {} from your terminal.",
                     path.to_string_lossy(),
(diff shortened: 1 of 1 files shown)


leoliu0605:
@cla-bot check

leoliu0605:
@cla-bot check

kubkon:
Thanks for the PR! Looking at the state of things, part of me is wondering if we ever even should put things in system path requiring root. Pehaps we ought to always install in user specific location such as `.local` instead and append to relevant default user shell config file so that this is picked up automatically at next shell startup?

leoliu0605:
> Thanks for the PR! Looking at the state of things, part of me is wondering if we ever even should put things in system path requiring root. Pehaps we ought to always install in user specific location such as `.local` instead and append to relevant default user shell config file so that this is picked up automatically at next shell startup?

Hi @kubkon,  that makes sense to me, and I agree that always installing to a user-local location and avoiding writes to system paths is a cleaner and safer default.

If you agree, I will modify this PR to install only under the user-local path and add it in `.zshrc`, `.bashrc` and prompt users to add it themselves to their preferred shell config files.

leoliu0605:
Hi @kubkon , any comments?

kubkon:
Hi @leoliu0605, sorry for not replying earlier - I was away for the past two weeks. Anyhow, I've discussed with the team and installing into a user-local bin path is preferable so let's go with that. Lemme know if you would like to pair on it, or if you prefer to implement it by yourself and I'll review the changes when your PR is ready.

leoliu0605:
> Hi @leoliu0605, sorry for not replying earlier - I was away for the past two weeks. Anyhow, I've discussed with the team and installing into a user-local bin path is preferable so let's go with that. Lemme know if you would like to pair on it, or if you prefer to implement it by yourself and I'll review the changes when your PR is ready.

Hi @kubkon , I've already implemented it; please check the code and video below as the result:

https://github.com/user-attachments/assets/49390e71-cce5-4cd5-849c-691e81ce3647

kubkon:
I've had a chat with @ConradIrwin and he corrected me that the osascript _is_ the right approach on macOS so we should not gut it out, so your initial approach at least on macOS seems better - sorry for the confusion. I think we would like to do the following:
* macOS - try `osascript`, if it fails, fall back to local installation directory
* linux - always install to a local dir (note: I am not sure we handle linux at all when it comes to CLI installation so something to check and/or just focus on macOS solely for now)
* windows - this option is unavailable on Windows so we don't need to worry about it at all

When it comes installing to a local dir, I think trying `$HOME/.local/bin` and creating if it doesn't exist seems fine. However, I would leave out adding that dir to user's PATH env var for now and just flag it in a toast/dialog box instead. Handling all user shells will require some thought so it might be best to just leave it out for now or altogether to reduce the complexity. How does that sound? @ConradIrwin could you weigh in please?

GitGlimpse895:
Hey @kubkon — just following up on this one. @leoliu0605 implemented everything you and @ConradIrwin agreed on back in December (osascript first, fallback to `~/.local/bin`, toast notification) and posted a demo video. That was 4 months ago with no response to their follow-up pings.

I checked current `main` — there's still zero fallback if osascript fails. The feature this PR adds is completely missing. Is there anything blocking a review at this point?

ConradIrwin:
Sorry for the slow responses here, and thank you again for getting this started.

Having spent a bit of time looking into it, I don't believe `~/.local/bin` is commonly used on macOS (at least it doesn't exist for me), as such it seems a bit arbitrary to bless that as the non-admin install path.

What I do think we should do is fix the documentation, and link to it when the admin path goes awry. Implemented in #59575

PR #42302, opened 2025-11-09
vim: Handle literal $ in search replacements

### LINKED ISSUE #42292: vim substitution bit bug
### Summary

Hello dears,

lag of vim substitution(also in Zed) when try to substitute a bash variable in bash file.

### Description
I tried to use `%s/\$Base/\$BaseNew/g` but all _$Base_ would be remove and not substitute.

Steps to reproduce:
1. create a bash file
2. define a variable
3. try to use %s to change variable with '$' sing

**Expected Behavior**:
the expection is clear :-)

**Actual Behavior**:
the result is removing matched word +it's $ sign

Zed: v0.211.6 (Zed) 
OS: Linux Wayland rocky 9.6
Memory: 22.9 GiB
Architecture: x86_64
GPU: Intel(R) Graphics (ADL GT2) || Intel open-source Mesa driver || Mesa 24.2.8

```log

```

</details>

cppcoffee:
Ensure the replacement parser treats \$ as $$ instead of falling through the generic escape path

Closes https://github.com/zed-industries/zed/issues/42292

Release Notes:

- Fixed vim search replacement parsing for literal dollar signs

### DIFF
--- a/crates/vim/src/normal/search.rs
+++ b/crates/vim/src/normal/search.rs
@@ -658,7 +658,6 @@ impl Replacement {
 mod test {
     use std::time::Duration;
 
-    use super::Replacement;
     use crate::{
         state::Mode,
         test::{NeovimBackedTestContext, VimTestContext},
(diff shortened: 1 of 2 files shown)


dinocosta:
Hey @cppcoffee ! Thank you for attempting to improve this 🙌 

I wonder if we can actually go one step further and mimic NeoVim's behavior without the need to escape `$`. For example, with the following buffer content:

```
$Base one
$Base two
$Base three
```

Running `:%s/$Base/$BaseNew` in NeoVim would result in:

```
$BaseNew one
$BaseNew two
$BaseNew three
```

Would you be able to try and see if you can get the same behavior in Zed without needing to use `\$` so it aligns more closely with what NeoVim does? Happy to take a look if helpful 🙂

PR #42310, opened 2025-11-09
Fix unexpected lack of output when stdout is a pipe

kitt-cat:
## Motivation

Writing log output conditionally on stdout.is_terminal() creates unexpected behavior when piping/redirecting stdout, for example for further filtering.

```sh
zed # plenty of output :) (so much one may want to use grep)
zed | grep ... # suddenly no output!
```

## Changes

- ~~b80f4d68a4e405dc8af73b179e294c4d9a6e0410 Adds a --log option which controls when logs are written to stdout and takes a filter argument.~~
- ~~daaeeb5e10fba4af386436e3ed62a0bba3d0e3b0 Tidy up -h/--help output a little;~~
  Moved to new PR: https://github.com/zed-industries/zed/pull/42496
- ~~cdd4a18ae9d1c222683578ca35e9c07ff52738bf Adds a --log option which controls when logs are written to stdout.~~
- 7893c9688f47915103e7a2928967e14905f34af7 Write logs to stdout when --foreground is set.

    To keep logs readily discoverable (and for convenience during development), I thought it would be nice to set --foreground automatically in debug builds. The way that manifests is as follows:

    - in release builds:
    if `--foreground` is set, log to stdout, otherwise log to paths::log_file()
    as before.

    - in debug builds:
    always log to stdout, and the `--foreground` option is hidden from `-h`/`--help`.

Release Notes:

- Fixed missing logs when stdout is a pipe

### DIFF
--- a/crates/zed/src/main.rs
+++ b/crates/zed/src/main.rs
@@ -240,7 +240,7 @@ pub fn main() {
     }
 
     zlog::init();
-    if stdout_is_a_pty() {
+    if args.foreground {
         zlog::init_output_stdout();
     } else {
         let result = zlog::init_output_file(paths::log_file(), Some(paths::old_log_file()));
@@ -1304,10 +1304,11 @@ struct Args {
     #[arg(long, hide = true)]
     crash_handler: Option<PathBuf>,
 
-    /// Run zed in the foreground, only used on Windows, to match the behavior on macOS.
+    /// Print logs
+    ///
+    /// Called by CLI binary when --foreground is given
     #[arg(long)]
-    #[cfg(target_os = "windows")]
-    #[arg(hide = true)]
+    #[cfg_attr(debug_assertions, arg(default_value_t = true, hide = true))]
     foreground: bool,
 
     /// The dock action to perform. This is used on Windows only.
--- a/crates/zed/src/main.rs
+++ b/crates/zed/src/main.rs
@@ -1306,7 +1306,7 @@ struct Args {
 
     /// Print logs
     ///
-    /// Called by CLI binary when --foreground is given
+    /// Set by CLI binary when running zed in the foreground
     #[arg(long)]
     #[cfg_attr(debug_assertions, arg(default_value_t = true, hide = true))]
     foreground: bool,


crates/gpui/src/platform.rs
```diff
@@ -114,8 +114,7 @@ pub(crate) fn current_platform(headless: bool) -> Rc<dyn Platform> {
         #[cfg(feature = "x11")]
         "X11" => Rc::new(
             X11Client::new()
-                .context("Failed to initialize X11 client.")
```
probably-neb, line 116: This change is unrelated

Cargo.toml
```diff
@@ -482,7 +482,7 @@ cfg-if = "1.0.3"
 chrono = { version = "0.4", features = ["serde"] }
 ciborium = "0.2"
 circular-buffer = "1.0"
-clap = { version = "4.4", features = ["derive"] }
```
probably-neb: The second commit and the addition of the `wrap_help` feature + doc string comments seems like a completely separate PR. If you wouldn't mind splitting it out into it's own PR, I'd much prefer to review and approve the two commits separately. It may seem tedious, but I'd like to be able to revert one without reverting the other if we run into issues with the `--log` arg.

crates/zed/src/main.rs
```diff
@@ -232,10 +232,17 @@ pub fn main() {
...
 
-    zlog::init();
-    if stdout_is_a_pty() {
+    if let Some(log_filter) = args.log {
+        match zlog::try_init(Some(log_filter.clone())) {
+            Err(err) => {
+                eprintln!("Could not initialize log with filter {}: {}", log_filter, err);
+                process::exit(1);
+            }
+            Ok(_) => (),
```
probably-neb: This will not take precedence over `ZED_LOG`. `zlog::try_init` will use the filter provided as a fallback if the env var is not set.

In general, and especially given the comment above, I think it would be better if the `--log` argument _did not_ take a log argument. I don't think it's worth having _4_ places to set the log configuration (settings, `ZED_LOG`/`RUST_LOG`, inside the `zlog` crate itself (defaults), and `--log`). I think it would be better to just have the env var.

crates/zed/src/main.rs
```diff
@@ -1278,11 +1284,35 @@ struct Args {
...
+    /// Prints system specs.
+    ///
+    /// Useful for submitting issues on GitHub when encountering a bug that
+    /// prevents Zed from starting, so you can't run `zed: copy system specs to
+    /// clipboard`
     #[arg(long)]
     system_specs: bool,
 
+    /// Log to stdout, with optional <LEVEL> (one of error, warn, info, debug, or trace).
+    #[cfg_attr(
```
probably-neb: If we were to remove the log filter value from the `--log` argument, that would remove the need to set a separate debug only value here, that overrides (duplicates) the default already in set in `zlog`.

probably-neb (changes requested):
Great Idea! Thanks for the PR.

I have a couple of design nits I document below, primarily around whether the `--log` argument should take a log filter. I think this adds unnecessary complexity to a system that does not warrant it.

I also think this PR should be split in two, which may seem like unnecessary busywork or micromanaging on my part, It just makes reviewing and approving these PRs easier when they are a small, well-scoped change. And also means we can revert one part without reverting another.

Cargo.toml
kitt-cat: Of course! I sometimes worry if I split things out too much it might come across as spamming small/superficial PRs which might be unwelcome :sweat_smile: but im happy to! https://github.com/zed-industries/zed/pull/42496

crates/gpui/src/platform.rs
kitt-cat: whoopsie

probably-neb:
Another note, I was reminded yesterday that actually we already have the `--foreground` flag to do just this, however, it seems to be hidden and only enabled on Windows.
```
    /// Run zed in the foreground, only used on Windows, to match the behavior on macOS.
    #[arg(long)]
    #[cfg(target_os = "windows")]
    #[arg(hide = true)]
    foreground: bool,
```

I think instead of adding the `--log` option, we should just make this arg work as expected on all platforms, and expose it

probably-neb:
Hey @kitt-cat, are you still interested in working on this?

kitt-cat:
Hello again, yes! I've merged it with the foreground option. 

I'm not really familiar with what `--foreground` does on windows, but I considered having zed fork-to-background by default to try and match(?). In the end I felt avoiding that behavior was probably for the best for primarily the following reasons:

- The Zed CLI binary (provided by the `cli` crate) already offers a daemonizing-by-default entrypoint. When I first started using zed I didn't realize the Zed CLI and zed were separate binaries with separate interfaces; I worry having two separate background-launch mechanisms which look and feel similar (but are different) may cause confusion.
- Feedback for early failures. I feel like it would be nice if, before forking and exiting, messages regarding errors which have prevented normal startup (like those from `files_not_created_on_launch()` and `failed_to_open_window()`) are written to stderr. I have considered a few ways to achieve this:
  - forking+exec late (after the window has been created successfully or errored): I think would run into issues redoing work and calling platform APIs twice, once in the parent once in the child.
  - not calling exec: there are [concerns with threads](https://mikeash.com/pyblog/friday-qa-2012-01-20-fork-safety.html).
  - forking early and sending errors messages via some sort of IPC to the original process for printing.
  
  i think its achievable, but it also felt a bit like an undue level of complexity where simplicity is probably preferred.

What do you think?

probably-neb:
Ok, a couple thoughts - sorry for the delayed response 

A) I don't think we should hide the `--foreground` flag in debug builds, it seems unnecessarily and could lead to confusion
B) I'm not entirely following what you're saying around the CLI, error feedback, etc. I think you're saying it would be nice to have better error reporting and logging when Zed is launched via the CLI? If so I think this is best suited to it's own PR, where we can discuss the merits of the technical solutions you came up with, and a reviewer who is more familiar with Windows than I am

SomeoneToIgnore:
Conflicts start to appear and last activity happened in the last year, so I'll close this draft as stale.

PR #42324, opened 2025-11-10
multi_buffer: Fix `editor::ExpandExcerpts` failing when cursor is at excerpt start

olejorgenb:
The bug is easily verified by:

1. open any multi-buffer
2. place the cursor at the beginning of an excerpt
3. run the editor::ExpandExcerpts / editor: expand excerpts action
4. The excerpt is not expanded

Since the `buffer_ids_for_range` function basically did the same and had even been changed the same way earlier I DRYed these functions as well.

Note: I'm a rust novice, so keep an extra eye on rust technicalities when reviewing :)

---

Release Notes:

- Fix editor: expand excerpts failing when cursor is at excerpt start

### DIFF
--- a/crates/multi_buffer/src/multi_buffer_tests.rs
+++ b/crates/multi_buffer/src/multi_buffer_tests.rs
@@ -4095,3 +4095,115 @@ fn test_random_chunk_bitmaps_with_diffs(cx: &mut App, mut rng: StdRng) {
         }
     }
 }
+
+/// Tests `excerpt_containing` and `excerpts_for_range` (functions mapping multi-buffer text-coordinates to excerpts)
+#[gpui::test]
+fn test_excerpts_containment_functions(cx: &mut App) {
+    // Multibuffer content for these tests:
+    //    0123
+    // 0: aa0
+    // 1: aa1
+    //    -----
+    // 2: bb0
+    // 3: bb1
+    //    -----MultiBufferOffset(0)..
+    // 4: cc0
+
+    let buffer_1 = cx.new(|cx| Buffer::local("aa0\naa1", cx));
+    let buffer_2 = cx.new(|cx| Buffer::local("bb0\nbb1", cx));
+    let buffer_3 = cx.new(|cx| Buffer::local("cc0", cx));
+
+    let multibuffer = cx.new(|_| MultiBuffer::new(Capability::ReadWrite));
+
+    let (excerpt_1_id, excerpt_2_id, excerpt_3_id) = multibuffer.update(cx, |multibuffer, cx| {
+        let excerpt_1_id = multibuffer.push_excerpts(
+            buffer_1.clone(),
+            [ExcerptRange::new(Point::new(0, 0)..Point::new(1, 3))],
+            cx,
+        )[0];
+
+        let excerpt_2_id = multibuffer.push_excerpts(
+            buffer_2.clone(),
+            [ExcerptRange::new(Point::new(0, 0)..Point::new(1, 3))],
+            cx,
+        )[0];
+
+        let excerpt_3_id = multibuffer.push_excerpts(
+            buffer_3.clone(),
+            [ExcerptRange::new(Point::new(0, 0)..Point::new(0, 3))],
+            cx,
+        )[0];
+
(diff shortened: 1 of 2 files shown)


olejorgenb:
There is another similar edge-case when a selection just span into an excerpt (which this PR does not fix):

image

`editor:ExpandExcerpts` will not expand the second excerpt.

That's one is kinda obscure though.

See the "// This range represent an selection with end-point just inside excerpt_2" test case.

Veykril (approved):
Thanks! (and sorry this took so long for me to get to).

Thanks for test as well!

crates/multi_buffer/src/multi_buffer_tests.rs
```diff
@@ -3965,3 +3965,114 @@ fn test_random_chunk_bitmaps_with_diffs(cx: &mut App, mut rng: StdRng) {
...
+
+    let excerpts: Vec<_> = snapshot.excerpts_for_range(p00..p23).collect();
+    assert_eq!(excerpts.len(), 2);
+    assert_eq!(excerpts[0].id, excerpt_1_id);
+    assert_eq!(excerpts[1].id, excerpt_2_id);
+
+    // This range represent an selection with end-point just inside excerpt_2
+    let excerpts: Vec<_> = snapshot.excerpts_for_range(p10..p20).collect();
+    assert_eq!(excerpts.len(), 1);
+    assert_eq!(excerpts[0].id, excerpt_1_id);
```
Veykril, line 4173: So your comment is saying this should contain 2 excerpts ideally is that right?Can you add a note here stating that this is not quite working as expected yet? Happy to merge the PR without that fixed, just want it tracked somehow

olejorgenb: > So your comment is saying this should contain 2 excerpts ideally is that right?

Perhaps - it corresponds to this scenario: https://github.com/zed-industries/zed/pull/42324#issuecomment-3511137090  From a user perspective I would expect both excerpts to expand, but I'm not sure if would be the sensible thing for _all_ usages of a which-excerpts-intersect-the-selection :thinking: 

Something like this?

```
// This range represent an selection with end-point just inside excerpt_2
// Today only excerpt_1 intersect this selection, but from a user perspective using `editor::ExpandExcerpts` it would make sense if excerpt_2 also did
```

(To be clear: this is the behavior in main as well, so this MR doesn't change the behavior)

olejorgenb: Probably easier if you just adjust the comment yourself :)

I can also create an issue on it?

crates/multi_buffer/src/multi_buffer_tests.rs
```diff
@@ -3965,3 +3965,114 @@ fn test_random_chunk_bitmaps_with_diffs(cx: &mut App, mut rng: StdRng) {
...
+    let excerpts: Vec<_> = snapshot.excerpts_for_range(p43..p43).collect();
+    assert_eq!(excerpts.len(), 1);
+    assert_eq!(excerpts[0].id, excerpt_3_id);
+
+    let excerpts: Vec<_> = snapshot.excerpts_for_range(p00..p23).collect();
+    assert_eq!(excerpts.len(), 2);
+    assert_eq!(excerpts[0].id, excerpt_1_id);
+    assert_eq!(excerpts[1].id, excerpt_2_id);
+
+    // This range represent an selection with end-point just inside excerpt_2
```
Veykril, line 4168:
```suggestion
    // This range represent an selection with end-point just inside excerpt_2
    // Today we only expand the first excerpt, but another interpretation that
    // we could consider is expanding both here
```

Veykril:
Added a comment, but this does need a rebase unfortunately (sorry), could you take care of that?

olejorgenb:
Done :heavy_check_mark: 

The conflict stems from this commit: c98b2d69448

image

So it was simple to resolve once I understood the changes. Required a small change in the tests as well since the new BufferOffset can't be used as an iteration range (for x..y).

I build and tested very briefly.

Veykril (approved):
Great, thanks!

PR #42328, opened 2025-11-10
cli: Allow opening new files in non-existent parent directories

SadiqOnGithub:
Discussion: https://github.com/zed-industries/zed/discussions/42327

Previously, the `zed` CLI would fail when attempting to open a new file if any of its parent directories did not yet exist on filesystem (e.g., `zed path/to/new_dir/new_file.txt` where `path/to/new_dir` does not exist).

This failure occurred because the `parse_path_with_position` function unconditionally called `canonicalize()` on the provided argument. `canonicalize()` fails if any component of the path, including intermediate directories, does not exist, causing the CLI to with an error like "No such file or directory".

This change modifies the path parsing logic. If `canonicalize()` fails, it now falls back to resolving the path to an absolute relative to the current working directory. This allows a `PathWithPosition` to be constructed with a valid `PathBuf` that can represent file whose parent directories may not yet exist.

This correctly shifts the responsibility for creating intermediate directories (and the file itself) from the CLI to the main application, aligning with user expectations and the behavior of other modern editors.

Closes #ISSUE

Release Notes:


SadiqOnGithub:
@cla-bot check

SadiqOnGithub:
@cla-bot check

SadiqOnGithub (commented):
@maxdeviant 

May I know why you think those comments were not needed?

maxdeviant:
> @maxdeviant
> 
> May I know why you think those comments were not needed?

Because they are just restating what the code is doing.

crates/cli/src/main.rs
```diff
@@ -133,26 +133,22 @@ fn parse_path_with_position(argument_str: &str) -> anyhow::Result<String> {
     let canonicalized = match Path::new(argument_str).canonicalize() {
         Ok(existing_path) => PathWithPosition::from_path(existing_path),
         Err(_) => {
-            let path = PathWithPosition::parse_str(argument_str);
+            let path_with_pos = PathWithPosition::parse_str(argument_str);
+            let path_argument = path_with_pos.path;
+
             let curdir = env::current_dir().context("retrieving current directory")?;
-            path.map_path(|path| match fs::canonicalize(&path) {
```
SomeoneToIgnore, line 138: Checking the old code, and this seems right?
First, we try to canonicalize something that might contain `:300:20` nonexisting suffixes, then, here, we try to canonicalize the path without them: and should give it a chance.

Then though, canonicalizing its parent does not matter much: and the fix should go there instead?

SadiqOnGithub: But, if we canonicalize somethings which doesn't exist then it will throw an error. So isn't better we parse the path with `PathWithPosition::parse_str()` and proceed?

https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize

SomeoneToIgnore: > exist then it will throw an error

Yep, and right after that, the fix could be applied, right?

IIRC, we do prefer to canonicalize here to eliminate all symlinks and related confusions: and here, we have not done it yet, as the first canonicalization, before `PathWithPosition::parse_str`, had failed.

SadiqOnGithub: @SomeoneToIgnore Allowed first two canonicalization.

SomeoneToIgnore:
Thank you for this PR and collaborating on another, I have looked at both and decided that https://github.com/zed-industries/zed/pull/43250 is indeed a better approach.

PR #42351, opened 2025-11-10
Fix circular reference issue between EditPredictionButton and PopoverMenuHandle

feeiyu:
Closes #ISSUE

While working on issue #40906, I discovered that RemoteClient was not being released after the remote project closed.
Analysis revealed a circular reference between EditPredictionButton and PopoverMenuHandle.

Dependency Chain: RemoteClient → Project → ZetaEditPredictionProvider → EditPredictionButton ↔ PopoverMenuHandle

image

a) EditPredictionButton hold the reference of PopoverMenuHandle 
https://github.com/zed-industries/zed/blob/5f8226457ee6e1346a224ae6b0329f014ea883f7/crates/zed/src/zed.rs#L386-L394

b) PopoverMenuHandle hold the reference of Fn which capture `Entity<EditPredictionButton>`
https://github.com/zed-industries/zed/blob/5fc54986c72f2863645302c5e6a99277f8c38cab/crates/edit_prediction_button/src/edit_prediction_button.rs#L382-L389

https://github.com/zed-industries/zed/blob/a9bc890497f1edaf4f177385cf96785de60e910c/crates/ui/src/components/popover_menu.rs#L376-L384

Release Notes:

- N/A

### DIFF
--- a/crates/edit_prediction_button/src/edit_prediction_button.rs
+++ b/crates/edit_prediction_button/src/edit_prediction_button.rs
@@ -383,7 +383,8 @@ impl Render for EditPredictionButton {
 
                 let mut popover_menu = PopoverMenu::new("zeta")
                     .menu(move |window, cx| {
-                        this.update(cx, |this, cx| this.build_zeta_context_menu(window, cx)).ok()
+                        this.update(cx, |this, cx| this.build_zeta_context_menu(window, cx))
+                            .ok()
                     })
                     .anchor(Corner::BottomRight)
                     .with_handle(self.popover_menu_handle.clone());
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (approved):
Superb, thank you so much for all the context and a fix.

I've found a fixed a few of these myself, so I suspect this might not be the last one around.
Not yet sure what to do about a systematic way to fix this, but any other similar fixes are very welcome.

feeiyu:
Perhaps we could check if key objects, such as Project/Workspace, are properly released after the project closure?

SomeoneToIgnore:
> after the project closure

Workspace isn't necessary going to be released, as technically, workspace == window and can be reused.
But overall, the "after the windows closure" idea seems viable to me.

PR #42352, opened 2025-11-10
Fix track file renames in git panel

### LINKED ISSUE #30549: Git: `git mv` shows as deleted file + untracked new file
### Summary

Git renames are not properly tracked in the git panel. They show as deleted file + new untracked file.

### Description

Steps to trigger the problem:
1. open zed into a git repo with a file named `old.rs`
2. from the terminal, use `git mv old.rs new.rs`
3. the git panel shows 1 deleted `old.rs` and 1 untracked `new.rs`

Actual Behavior:
Expected Behavior:
It should display a renamed (or moved) file in the changes list

Zed: v0.185.15 (Zed) 
OS: Windows 10.0.22631
Memory: 63.2 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 4090 || NVIDIA || 576.02

ddoemonn:
Closes #30549

Release Notes:

- Fixed: Git renames now properly show as renamed files in the git panel instead of appearing as deleted + untracked files
Screenshot 2025-11-10 at 17 39 44
Screenshot 2025-11-10 at 17 39 55

### DIFF
--- a/crates/collab/src/db/queries/projects.rs
+++ b/crates/collab/src/db/queries/projects.rs
@@ -1005,6 +1005,7 @@ impl Database {
                         is_last_update: true,
                         merge_message: db_repository_entry.merge_message,
                         stash_entries: Vec::new(),
+                        renamed_paths: Default::default(),
                     });
                 }
             }
--- a/crates/collab/src/db/queries/rooms.rs
+++ b/crates/collab/src/db/queries/rooms.rs
@@ -796,6 +796,7 @@ impl Database {
                             is_last_update: true,
                             merge_message: db_repository.merge_message,
                             stash_entries: Vec::new(),
+                            renamed_paths: Default::default(),
                         });
                     }
                 }
--- a/crates/fs/src/fake_git_repo.rs
+++ b/crates/fs/src/fake_git_repo.rs
@@ -359,6 +359,7 @@ impl GitRepository for FakeGitRepository {
             entries.sort_by(|a, b| a.0.cmp(&b.0));
             anyhow::Ok(GitStatus {
                 entries: entries.into(),
+                renamed_paths: HashMap::default(),
             })
         });
         Task::ready(match result {
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -2045,7 +2045,7 @@ fn git_status_args(path_prefixes: &[RepoPath]) -> Vec<OsString> {
         OsString::from("status"),
         OsString::from("--porcelain=v1"),
         OsString::from("--untracked-files=all"),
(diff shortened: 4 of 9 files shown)


cole-miller:
Thanks! This is a good start, but I think we should find a way to show both the old and the new paths in the panel. That probably requires a change to how we track statuses in the data layer, so we can store an additional path for things that were renamed. One strategy that I think might make sense would be to add an additional map between old paths and new paths to `RepositorySnapshot`, separate from `statuses_by_path`. Then we could look up the old path for a given new path in the UI layer for the git panel when we render an entry that has a `Renamed` status.

ddoemonn:
Screenshot 2025-11-10 at 19 50 59

@cole-miller what do you think about this?

cole-miller:
@ddoemonn That looks reasonable to me, though I do want to have our design team weigh in before we merge this (and in the meantime we can get the data layer ready to go).

ddoemonn:
@cole-miller What do you think about this approach except design for now?

cole-miller:
Thanks! The backend side of this looks pretty good to me now. It would be good to test (if you haven't already) in what cases git emits the renamed status code for the index/working copy/both, and check that we show the right thing in all cases. Also that staging/unstaging/restoring/committing/uncommitting with renamed paths all have reasonable behavior.

As for the design, tagging @mattermill for his thoughts, but one big question for me is how this interacts with #41857.

ddoemonn:
@cole-miller is there anything you want me to update about this PR?

mattermill:
@ddoemonn This looks reasonable to me! However, as @cole-miller noted, we will likely have to revisit once we move on to #41857.

cole-miller:
@ddoemonn Sorry for merging this prematurely, it seems like there are still some bugs:

- Renamed files no longer appear in the project diff. We'll need to show them in some form (right now we show a deleted file at the old path and an added file at the new path; ideally we should improve that as part of this change, so you only get one project diff entry for the renamed file).
- Staging and unstaging of renamed files doesn't work--it causes the entry to split back into an added/removed pair.

If you open a new PR we can advise on how to go about fixing these!

ddoemonn:
Hey @cole-miller! Thanks for clarifying these issues. I actually experienced these bugs after the PR was already closed, so I didn’t realize they were happening beforehand. I’m a little bit confused about these parts too, to be honest.

I can definitely open a new PR to fix them, but I’m a bit busy at work today. How urgent is this? I want to make sure I prioritize it appropriately.

ddoemonn:
@cole-miller  new PR is here https://github.com/zed-industries/zed/pull/43077

PR #42385, opened 2025-11-10
Miniprofiler

localcc:
Release Notes:

- Added hang detection and a built in performance profiler


crates/gpui/src/profiler.rs
```diff
@@ -0,0 +1,218 @@
...
+            thread_id,
+            timings: serialized_timings,
+        }
+    }
+}
+
+// Allow 20mb of task timing entries
+const MAX_TASK_TIMINGS: usize = (20 * 1024 * 1024) / core::mem::size_of::<TaskTiming>();
+
+pub(crate) type TaskTimings = circular_buffer::CircularBuffer<MAX_TASK_TIMINGS, TaskTiming>;
```
sourcefrog, line 157: I was looking at Zed memory usage and noticed that this seems to be allocating a fixed 20MB per thread for each of 33 threads on my machine, which is not critical but is noticeable.

Image

I wonder if all of the threads are really filling all the buffer space? `circular-buffer` seems to allocate all the space up front but perhaps we could use something else that allocates more lazily?

PR #42391, opened 2025-11-10
Add QuitMode policy to GPUI

reflectronic:
Applications can select a policy for when the app quits using the new function `Application::with_quit_mode`:
- Only on explicit calls to `App::quit`
- When the last window is closed
- Platform default (former on macOS, latter everywhere else) 

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -169,6 +169,13 @@ impl Application {
         self
     }
 
+    /// Configures when the application should automatically quit.
+    /// By default, [`QuitMode::Default`] is used.
+    pub fn with_quit_mode(self, mode: QuitMode) -> Self {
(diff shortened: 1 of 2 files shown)


jbirnick:
@reflectronic May I ask why the default quit mode is different for MacOS?

Doesn't that mean that when I leave the quit mode as `Default`, the behavior of my program will differ between the platforms? I.e., when I write a cross-platform app, I'm forced to set a quit mode anyway?

reflectronic:
> May I ask why the default quit mode is different for MacOS?

I think the expectation for many types of apps on macOS is that they only exit when the user runs the "Quit" command.

This was also the behavior in GPUI before the refactoring in https://github.com/zed-industries/zed/pull/41436. This patch is just restoring the existing behavior.

PR #42404, opened 2025-11-11
docs: Fix broken links in docs with lychee

miguelraz:
Lychee is a [Rust based](https://lychee.cli.rs) async parallel link checker.

I ran it against the codebase to suss out stale links and fixed those up.

There's currently 2 remaining cases that I don't know how to resolve:

1. https://flathub.org/apps/dev.zed.Zed - nginx is giving a 502 bad gateway
2. https://github.com/zed-industries/zed/actions/workflows/ci.yml/badge.svg - I don't want to mess with the CI pipeline in this PR.

Once again, I'll punt to the Docs Czar to see if this gets incorporated into CI later.

---

## Running `lychee` locally:

```
cargo binstall -y lychee
lychee .
```

---
Release Notes:

- N/A

### DIFF
--- a/docs/src/ai/llm-providers.md
+++ b/docs/src/ai/llm-providers.md
@@ -587,7 +587,7 @@ These routing controls let you fine‑tune cost, capability, and reliability tra
 
 ### Vercel v0 {#vercel-v0}
 
-[Vercel v0](https://vercel.com/docs/v0/api) is an expert model for generating full-stack apps, with framework-aware completions optimized for modern stacks like Next.js and Vercel.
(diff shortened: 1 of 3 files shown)


maxdeviant:
> Once again, I'll punt to the Docs Czar to see if this gets incorporated into CI later.

We do already run lychee in CI: https://github.com/zed-industries/zed/blob/378b30eba5da7b9131b4a1d5bcee5bf09ad567ef/tooling/xtask/src/tasks/workflows/run_tests.rs#L401-L410

But I believe we configure it just to look at internal links, due to a high false-positive rate for external links.

MrSubidubi (approved):
The Flathub one seems to be back up, so might have just been another false positive as Marshall mentioned above. 


Other than that, looks good, thank you for checking!

PR #42424, opened 2025-11-11
vim: Add vim mode in Command Palette

0x2CA:
in Command Palette:

`escape` Cancel
`shift-escape` Switch To NormalMode

https://github.com/user-attachments/assets/2c4eb1bd-d671-4cbc-831e-0742cfd1d22e

This is just experimental, if the team can accept it, it can be extended to any single-line editing

Release Notes:

- Added Vim mode in Command Palette

### DIFF
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -983,7 +983,7 @@
     }
   },
   {
-    "context": "Editor && mode == auto_height && VimControl",
+    "context": "Editor && (mode == auto_height || mode == single_line) && VimControl",
     "bindings": {
       // TODO: Implement search
       "/": null,
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -348,6 +348,12 @@
     }
   },
   {
+    "context": "mode == single_line && vim_mode == insert",
+    "bindings": {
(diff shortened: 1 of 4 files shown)


0x2CA:
I think it's possible to enable Vim mode for all single lines directly, and it usually won't cause any problems. Of course, a safer approach is to enable it one place at a time.

niekdomi:
Vim's command line only has an insert (?) mode and with `c-w`,`c-h`,`c-u` we can delete character /  word(s). This is also what zed's command-line currently does (more or less). Adding things like normal-mode may be nice to have but IMHO doesn't match vim's behavior. 

This should therefore be rather an option than default behavior

dinocosta:
Thank you both for chiming in! 🙂
We've seen a couple requests to make vim mode work in single line editors like the command palette and the one used by the `editor: rename` action, and I believe theres' other Pull Requests open to try and make this a reality. We'll see if we can allocate some time to think about how to best introduce this, thanks!

ConradIrwin:
I'm pretty sure we should not have `escape` or `shift-escape` in the command palette switch you to normal mode (it should close), but like vim we should support insert mappings (including `ctrl-o`, or `ctrl-r`) in this context.

PR #42438, opened 2025-11-11
vim: Add scroll keybindings for the OutlinePanel

0x2CA:
Closes #ISSUE

```json
  {
    "context": "OutlinePanel && not_editing",
    "bindings": {
      "enter": "editor::ToggleFocus",
      "/": "menu::Cancel",
      "ctrl-u": "outline_panel::ScrollUp",
      "ctrl-d": "outline_panel::ScrollDown",
      "z t": "outline_panel::ScrollCursorTop",
      "z z": "outline_panel::ScrollCursorCenter",
      "z b": "outline_panel::ScrollCursorBottom"
    }
  },
  {
    "context": "OutlinePanel && editing",
    "bindings": {
      "enter": "menu::Cancel"
    }
  },
```

Release Notes:

- Added scroll keybindings for the OutlinePanel


0x2CA:
I don't understand why the test is giving this error.

```log
  stderr ───

    thread 'change_list::test::test_gi' panicked at crates\vim\src\test\vim_test_context.rs:107:14:
    called `Result::unwrap()` on an `Err` value: Error loading built-in keymap "keymaps/vim.json": Errors in user keymap file.


    In section with `context = "OutlinePanel && not_editing"`:

    - In binding `"ctrl-u"`, didn't find an action named `"outline_panel::ScrollUp"`.

    - In binding `"ctrl-d"`, didn't find an action named `"outline_panel::ScrollDown"`.

    - In binding `"z t"`, didn't find an action named `"outline_panel::ScrollCursorTop"`.

    - In binding `"z z"`, didn't find an action named `"outline_panel::ScrollCursorCenter"`.

    - In binding `"z b"`, didn't find an action named `"outline_panel::ScrollCursorBottom"`.
```

probably-neb:
It's because the actions are declared in the outline panel, and gpui collects the available actions based on what crates are linked in the final binary.

In tests this can be annoying because only crates that are depended on and referenced will be linked in the test executable.

You can see we had to do the same thing for `project_panel`, it's included in the `[dev-dependencies]` section of `crates/vim/Cargo.toml`, and is referenced here https://github.com/zed-industries/zed/blob/097024d46ffc12756c6451517f55d41c14849866/crates/vim/src/test/vim_test_context.rs#L24

Doing the same for `outline_panel` will make the outline panel actions be defined when the tests try to parse `keymaps/vim.json`

Sorry for the confusion, this is definitely a rough edge!

probably-neb:
Hey @0x2CA, nice work on this. A couple thoughts:

- We should probably add `z o` and `z c` folding bindings for completeness
- I'm not sure how valuable the `z t` `z z` and `z b` bindings are, maybe they're just in the bucket of vim bindings I don't use, but I don't really see myself ever choosing them over normal scroll bindings
- I wonder if `ctrl-o` and `ctrl-d` should scroll up or down by a single "parent" element? I.e. they would jump between top level items. I feel like fixed size jumps make less sense in a uniform list context, but maybe that's just me.

What do you think? I'm willing to be convinced either way on all points

0x2CA:
> * We should probably add `z o` and `z c` folding bindings for completeness
> * I'm not sure how valuable the `z t` `z z` and `z b` bindings are, maybe they're just in the bucket of vim bindings I don't use, but I don't really see myself ever choosing them over normal scroll bindings

This implementation is modeled after the project's panel implementation. For me, I usually use zz, and for fold, it indeed lacks binding.

> * I wonder if `ctrl-o` and `ctrl-d` should scroll up or down by a single "parent" element? I.e. they would jump between top level items. I feel like fixed size jumps make less sense in a uniform list context, but maybe that's just me.
> 

This suggestion is indeed good, but I need to study whether it can be implemented. Ideally, it should be at the top level?

probably-neb:
Oh I really like you're idea to have `h` and `l` expand and collapse entries. Where is the `-` binding coming from? Is that a Vim binding for something I'm unaware of?

0x2CA:
> Oh I really like you're idea to have `h` and `l` expand and collapse entries. Where is the `-` binding coming from? Is that a Vim binding for something I'm unaware of?

I am just a poor imitator.This is a complete imitation of the project panel's binding changes.

https://github.com/zed-industries/zed/blob/4f5cc0a24b8969f60429f057d7c4054bc15e038a/assets/keymaps/vim.json#L918-L923

The fundamental source of this is netrw.

probably-neb:
I merged main to resolve the conflicts. Enabling auto-merge on this. Thank you! Great work!

PR #42441, opened 2025-11-11
Add file history view

### LINKED ISSUE #16827: Implement file history viewing, including local and Git.
### Check for existing issues

- [X] Completed

### Describe the feature

Right-clicking on the edit page allows you to view the file's history of modifications and Git historical versions.

_No response_

ddoemonn:
Closes #16827

Release Notes:

- Added: File history view accessible via right-click context menu on files in the editor or project panel. Shows commit history for the selected file with author, timestamp, and commit message. Clicking a commit opens a diff view filtered to show only changes for that specific file.

Screenshot 2025-11-11 at 16 31 32
Screenshot 2025-11-11 at 16 31 24


AlvaroParker:
Hi! Just passing by and wondering if this is lazy loading or not? If a file has a history that is too large (old repos like Linux or Firefox), it might slow down the UI.

0x2CA:
> Hi! Just passing by and wondering if this is lazy loading or not? If a file has a history that is too large (old repos like Linux or Firefox), it might slow down the UI.

I've taken a general look at the implementation, and it's simply using git asynchronously to fetch the history and then parsing the git output

need to change to stream loading, in batches, and display dynamically while the user scrolls

cameron1024:
Thanks for getting involved! Summarizing the things we went over in our call
together:

Three high level pieces of feedback:
- The design/UI side of things will need input from our designer. I'll get in
  touch with him and see what he thinks, and we can go over the design
  together once we have his feedback
- We should stream results in chunks. The `-n` and `--skip` CLI arguments to
  `git log` make this pretty straightforward
- We should support remote (SSH/WSL/etc.) repositories

When it comes to implementing remote, the `git show` command is a similar
command that already has all this plumbing:
- You've already got the protobuf definitions done
- You'll need to add code to handle the protobuf request in the remote server
  application (the headless Zed instance that runs on the remote machine).
  You can do this by creating a function in
  `crates/project/src/git_store.rs`, similar to `GitStore::handle_show`. Take
  a look at `GitStore::init` to see how that function is plumbed into the
  app.
- You can also remove the `anyhow!()` in `fn file_history` in the remote
  branch and again copy what `git show` does (create the
  `proto::GitFileHistory`, send it to the client, and parse the
  `proto::GitFileHistoryResponse`)


A few smaller tweaks:
- using a string like `<<commit_end-(random UUID)>>` (using an actual random
  UUID) will essentially eliminate any chance of encountering that string in
  an actual diff. Splitting it up with `concat!()` will also mean that we
  don't encounter that exact string in this file, which is a small
  quality-of-life feature for Zed maintainers
- We can use `std::str::from_utf8` instead of `String::from_utf8_lossy` to
  avoid copying potentially large strings, and we can simply use `?` to skip
  invalid UTF-8 strings, since git will automatically convert non-UTF-8
  strings in most cases anyways.

Thanks again for looking into this, excited to use it myself 😅. Feel free to
ping me if you have any questions!

ddoemonn:
Hi @cameron1024 

I made some updates can you check it out?

Thanks!

crates/git/src/repository.rs
```diff
@@ -1421,6 +1444,94 @@ impl GitRepository for RealGitRepository {
...
+        limit: Option<usize>,
+    ) -> BoxFuture<'_, Result<FileHistory>> {
+        let working_directory = self.working_directory();
+        let git_binary_path = self.any_git_binary_path.clone();
+        self.executor
+            .spawn(async move {
+                let working_directory = working_directory?;
+                // Use a unique delimiter with a random UUID to separate commits
+                // This essentially eliminates any chance of encountering the delimiter in actual commit data
+                let commit_delimiter = format!("<<COMMIT_END-{}>>", Uuid::new_v4());
```
cameron1024: I think we should hardcode a single UUID rather than randomly generating one each time. We try to control randomness quite carefully, since we have a lot of randomized tests that rely on being able to seed random operations. Of course, it's probably unlikely to actually cause problems here, but I think it's worth hardcoding just for consistency.

crates/project/src/git_store.rs
```diff
@@ -2233,6 +2245,45 @@ impl GitStore {
...
+        let file_history = repository_handle
+            .update(&mut cx, |repository_handle, _| {
+                repository_handle.send_job(None, move |state, _| async move {
+                    match state {
+                        RepositoryState::Local { backend, .. } => backend.file_history(path).await,
+                        RepositoryState::Remote { .. } => {
+                            unreachable!("file history request received for remote repository")
+                        }
+                    }
+                })
```
cameron1024: I think this can just be replaced with `repository_handle.file_history(path)`, since that function already does the `match` on the `RepositoryState`

crates/project_panel/src/project_panel.rs
```diff
@@ -428,6 +430,74 @@ pub fn init(cx: &mut App) {
...
+                        return;
+                    }
+                }
+            }
+
+            // Fallback: try to get from active editor
+            if let Some(active_item) = workspace.active_item(cx) {
+                if let Some(editor) = active_item.downcast::<Editor>() {
+                    if let Some(buffer) = editor.read(cx).buffer().read(cx).as_singleton() {
+                        if let Some(file) = buffer.read(cx).file() {
```
cameron1024: I think we can avoid some of the nesting here by either:
- replace `if let Some` with `let Some(...) = ... else { return; }`
- use a (recently stabilized) `let`-chain, which might look a bit like

```rust
if let Some(active_item) = workspace.active_item(cx)
    && let Some(editor) = active_item.downcast::<Editor>()
    && let Some(buffer) = editor.read(cx).buffer().read(cx).as_singleton()
    && let Some(file) = buffer.read(cx).file()
{
  // the body    
}
```

cameron1024 (commented):
Backend stuff is looking better, just a few small issues 🎉

I think for now, we're just waiting on a design review - I'll let you know when I have more info. Apologies for the delay 😅

ddoemonn:
@cameron1024 thanks for feedbacks, I learned so much from this PR 😅

ddoemonn:
@cameron1024 sorry for bothering, but I wanted to ask if there are any updates about this PR, both regarding the design approach and the code fixes?

cameron1024:
Just got these from our designer - I think these designs should be able to reuse a bunch of the existing UI components, but let me know if you run into trouble! And I think on the lazy-loading issue, for now we can go with just a "load more" button. Don't have designs for that, but shouldn't be too hard to come up with something reasonable.

History Single
History List

ddoemonn:
thanks a lot!!! @cameron1024

ddoemonn:
@cameron1024 for commit view code part I wanna ask maybe we can use existing Editor to show code which I guess is better for understanding code totally. What do you think? I can share a video with you if you want

ddoemonn:
@cameron1024 what do design team thinks about these 

Screenshot 2025-11-20 at 18 32 15
Screenshot 2025-11-20 at 18 33 03

cameron1024:
That's looking really good! I'll take a look at the code and do a quick review for any issues.

We're in a weird position right now, because there are quite a few Git UI features in progress (e.g. [git graph](https://github.com/zed-industries/zed/pull/43106)) and I think we'd like to make sure they're consistent. We're going to be discussing how we can make this straightforward for everyone either today or tomorrow, and I'll let you know as soon as I have more info. Apologies again for the delay 😅 Excited to get this into people's hands!

crates/git/src/repository.rs
```diff
@@ -1421,6 +1444,94 @@ impl GitRepository for RealGitRepository {
...
+        limit: Option<usize>,
+    ) -> BoxFuture<'_, Result<FileHistory>> {
+        let working_directory = self.working_directory();
+        let git_binary_path = self.any_git_binary_path.clone();
+        self.executor
+            .spawn(async move {
+                let working_directory = working_directory?;
+                // Use a unique delimiter with a hardcoded UUID to separate commits
+                // This essentially eliminates any chance of encountering the delimiter in actual commit data
+                let commit_delimiter = "<<COMMIT_END-3f8a9c2e-7d4b-4e1a-9f6c-8b5d2a1e4c3f>>";
```
cameron1024: Can we make this:
```rust
let commit_delimiter = concat!(
  "<<COMMIT_END-",
  "3f8a9c2e-7d4b-4e1a-9f6c-8b5d2a1e4c3f>>",
);
```
This will avoid having the literal string `"<<COMMIT_END-3f8a9c2e-7d4b-4e1a-9f6c-8b5d2a1e4c3f>>"` in this specific source file, so git history will still work in this file. All `concat!` does is concatenate string literals together at compile time, so the runtime behaviour is identical.

Angelk90:
@ddoemonn : Personally, I think it's best to put two lines above and below to separate the profile picture and name from the commit message.

See image below.

Screenshot 2025-11-20 alle 17 07 20

If there are multiple authors, is it possible to display them?

@cameron1024 : What do you think?

cameron1024:
I think both designs are useful in different contexts.

My personal view is: the two-line view is more helpful when exploring a history you're not familiar with, but the one-line view is better when you're just trying to find a specific commit you wrote a few days ago. I'd personally love for both designs to be supported, with a way to toggle quickly between the "compact" and "expanded" views, but I'll need to see what people on the team think. There may be issues with this approach I haven't thought of yet...

Even if people like the idea, we may prefer to ship what we have, and add the toggle in a later PR 🤷  I'll hopefully have more info later today or tomorrow

ddoemonn:
@cameron1024 No worries! Take your time with the review.

Makes sense about waiting for the git graph and other features, consistency across the UI is important. Looking forward to hearing what you all decide.

I pushed the code that generates the design we discussed. 

Note: for the remote handling, I implemented what I thought made the most sense, but I'm definitely open to reviews and guidance on that part if you have different ideas.

ddoemonn:
And I totally agree with `Even if people like the idea, we may prefer to ship what we have, and add the toggle in a later PR 🤷 ` @cameron1024 As you can see it's not that small ship 😄

ddoemonn:
@cameron1024 Hi!! Is there any updates on this?

arguiot:
Hey! Looking forward for this feature. Any ETA?

crates/git_ui/src/commit_view.rs
```diff
@@ -242,11 +275,398 @@ impl CommitView {
...
+                None => Icon::new(IconName::Person)
+                    .color(Color::Muted)
+                    .size(IconSize::Medium)
+                    .into_element()
+                    .into_any(),
+                Some(None) => Icon::new(IconName::Person)
+                    .color(Color::Muted)
+                    .size(IconSize::Medium)
+                    .into_element()
+                    .into_any(),
```
cameron1024: Style nit: we could merge these branches with `None | Some(None) => Icon::new(...)` so it's clearer that they both do the same thing

cameron1024: Actually just noticed the `else` branch also has the same code. It's possible to extract that with a big `if let ...`, but it might be nicer to just pull out a `Self::fallback_commit_avatar()` function for this case

crates/git_ui/src/file_history_view.rs
```diff
@@ -0,0 +1,638 @@
...
+        sha: &SharedString,
+        window: &mut Window,
+        cx: &mut App,
+    ) -> AnyElement {
+        let remote = self.remote.as_ref().filter(|r| r.host_supports_avatars());
+
+        if let Some(remote) = remote {
+            let avatar_asset = CommitAvatarAsset::new(remote.clone(), sha.clone());
+            match window.use_asset::<CommitAvatarAsset>(&avatar_asset, cx) {
+                None => Icon::new(IconName::Person)
```
cameron1024: Same nitpick about repeating the fallback icon case

Angelk90:
@ddoemonn : In the screenshot I don't see the commit sha, shouldn't it be the first one with seven characters?
image

cameron1024:
Design team are happy with how it looks 🎉

I'll be doing a final code review later today (a few small nitpicks but nothing major). The only outstanding thing is that I think we could do with a few tests. ~I think the main thing we should test is the git CLI interactions and the commit-parsing code. We usually don't test GUI stuff~ We actually don't test the CLI interactions - ignore me 🙈. I think in that case, just a quick test for the markdown parsing code would be important, but that's it for testing.

But I'll be back later today with more details

crates/git/src/repository.rs
```diff
@@ -1421,6 +1444,94 @@ impl GitRepository for RealGitRepository {
...
+                let format_string = format!(
+                    "--pretty=format:%H%x00%s%x00%B%x00%at%x00%an%x00%ae{}",
+                    commit_delimiter
+                );
+
+                let mut args = vec!["--no-optional-locks", "log", "--follow", &format_string];
+
+                // Add pagination arguments if needed
+                let skip_str;
+                let limit_str;
```
cameron1024, line 1506: Style nit: given that these `let` bindings are only used in the `if`-blocks immediately below (and not anywhere deeper in the function) it might make more sense to scope them to the `if`-blocks like:
```rust
if skip > 0 {
  let skip_str = skip.to_string();
  args.push("--skip");
  args.push(&skip_str);
}
```
Or even just inline them:
```rust
if skip > 0 {
  args.push("--skip");
  args.push(&skip.to_string());
}
```

crates/git_ui/src/commit_view.rs
```diff
@@ -242,11 +275,398 @@ impl CommitView {
...
+                        Button::new("view_on_github", "View on GitHub")
+                            .icon(IconName::Github)
+                            .style(ui::ButtonStyle::Subtle)
+                            .on_click(move |_, _, cx| cx.open_url(&url))
+                    })),
+            )
+            .child(self.render_commit_message(window, cx))
+    }
+
+    fn process_github_issues(message: &str, remote: &GitRemote) -> String {
```
cameron1024, line 359: This function would be a good candidate for a quick test IMO. Shouldn't need any GPUI machinery - just a plain `#[test]` should be fine

crates/git/src/repository.rs
ddoemonn: Hey! I tried both suggestions but they don't work due to Rust's lifetime rules. The args vector holds references to these strings, so they need to outlive the if-blocks. When scoped inside or inlined, the strings get dropped while args still borrows them, causing compilation errors.

Should I keep the current pattern or is there another approach you'd prefer?

cameron1024: Ah, didn't spot that, in that case feel free to keep it how it is :)

ddoemonn:
@cameron1024 quick question should we remove this Metada field from the commit_view? 
Screenshot 2025-11-28 at 17 29 29

ddoemonn:
@cla-bot check

ddoemonn:
@cameron1024 the PR is ready to review!

cameron1024 (approved):
Amazing! Thanks so much for working on this - I think we can merge it now! 🎉

BTW, I pushed a commit just cleaning up a few old sections of code that were commented out.  

Can't wait to start using this, thanks again 🙏

ddoemonn:
@cameron1024 It was such a pleasure working on this! Really happy we're shipping it 🚀

Be-ing:
This is a great feature, thank you @ddoemonn and @cameron1024. On the one hand it is nice that commits' diffs are filtered to just that file's history, but it would be great to have an easy way to switch between viewing just that file's history and the diff for the entire commit. Perhaps the commit hash could be turned into a link that shows the entire commit? Or a dedicated button for this?

imblowsnow:
Can you add a history to compare with the current one?

PR #42442, opened 2025-11-11
fix: Preserve agent settings when changing model from Agent Panel

### LINKED ISSUE #41344: `update_settings_file` deletes settings if settings file needs to be migrated
### Summary

Zed overwrites [Agent settings](https://zed.dev/docs/ai/agent-settings) when choosing a different LLM from the Agent Panel.

### Description

I have my Agent Panel docked to the left and my settings file needs an auto-migration, as it uses deprecated settings. When I choose a different model from the Agent Panel, Zed overwrites all settings related to it.

Steps to reproduce:
1. Modify the `agent` key in root `settings.json` (set `dock` to `left`)
2. Open Agent Panel
3. Choose a different model than the default one

**Expected Behavior**:
Zed should use the chosen model, only overwrite the `default_model` key and not update the remaining user settings defined in `agent` key.

**Actual Behavior**:
Zed overwrites all settings pertaining to the `agent` key. This behavior occurs even if you don’t have a default model set.

<details><summary>Screen capture</summary>

https://github.com/user-attachments/assets/c3d4631c-f864-4bff-8293-d0211243d5c2

</details>

Zed: v0.209.6 (Zed) 
OS: macOS 15.7
Memory: 18 GiB
Architecture: aarch64

krishna-medapati:
Fixes #41344

## Problem
When selecting a different model from the Agent Panel dropdown, Zed overwrites ALL agent settings instead of just updating the model. This resets user configurations like dock position, button visibility, and other customizations.

## Root Cause
The issue was in `crates/agent_ui/src/agent_model_selector.rs` where `get_or_insert_default()` was being used. This method creates a new `AgentSettingsContent` with default values when one doesn't exist, which overwrites all existing user settings.

## Solution
Modified the model selection handler to:
1. Check if agent settings exist before creating new ones
2. Only create default settings if none exist (preserving existing ones)
3. Directly update the `inline_assistant_model` field without replacing the entire settings object

## Changes Made
- Replaced `get_or_insert_default()` with explicit check and direct field update
- Added comments to explain the fix
- Preserved all other user configurations when changing model

## Testing
- ✅ Set `agent.dock = "left"` in settings.json
- ✅ Changed model from Agent Panel dropdown
- ✅ Verified dock position remained "left" 
- ✅ Verified model was updated correctly
- ✅ Tested with other agent settings (button, width, etc.) - all preserved

## Impact
Users can now change models from the Agent Panel without losing their custom agent panel configurations.

## Files Changed
- `crates/agent_ui/src/agent_model_selector.rs` - Fixed model selection logic

## Breaking Changes
None - This is a bug fix that preserves existing behavior while fixing the overwriting issue.

## CLA
I agree to the terms of the [Contributor License Agreement](https://cla-assistant.io/cla).

### Fixed
- **Agent Panel**: Fixed issue where changing model from Agent Panel would overwrite all agent settings (like dock position, button visibility, etc.) instead of just updating the model selection

### DIFF
--- a/crates/agent_ui/src/agent_model_selector.rs
+++ b/crates/agent_ui/src/agent_model_selector.rs
@@ -39,10 +39,13 @@ impl AgentModelSelector {
                         match &model_usage_context {
                             ModelUsageContext::InlineAssistant => {
                                 update_settings_file(fs.clone(), cx, move |settings, _cx| {
-                                    settings
-                                        .agent
-                                        .get_or_insert_default()
-                                        .set_inline_assistant_model(provider.clone(), model_id);
+                                    // Ensure agent settings exist without overwriting existing ones
+                                    if settings.agent.is_none() {
+                                        settings.agent = Some(Default::default());
+                                    }
+                                    
+                                    // Only update the inline assistant model, preserve other settings
+                                    settings.agent.as_mut().unwrap().set_inline_assistant_model(provider.clone(), model_id);
                                 });
                             }
                         }


MrSubidubi:
Thanks, but I do not think this is the proper fix for this issue - https://github.com/zed-industries/zed/issues/41344#issuecomment-3516993214

PR #42446, opened 2025-11-11
Fix search wrap detection when search_wrap is false

stanhua:
Reproduce the original bug:
1. Search for text that has a single match in the buffer
2. Press ESC
3. Move cursor above the search match
4. Press F3 to find next match. It doesn't find the match but shows "No more matches". Alternatively you can press Ctrl-f to open the search again, and press Enter to search for the same string. It also shows "No more matches".

The bug also happens when there are more than one matches, and the cursor is between the second-to-last and last match.

This patch changes `select_match()` to always call `searchable_item.match_index_for_direction()` and then check whether the `new_match_index` is before or after the current `index` to detect wrapping.

Release Notes:

- Fix search wrap detection when search_wrap is false

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -1023,16 +1023,17 @@ impl BufferSearchBar {
                 .get(&searchable_item.downgrade())
                 .filter(|matches| !matches.is_empty())
         {
-            // If 'wrapscan' is disabled, searches do not wrap around the end of the file.
+            let new_match_index = searchable_item
+                .match_index_for_direction(matches, index, direction, count, window, cx);
+
+            // If 'search_wrap' is disabled, searches do not wrap around the file.
             if !EditorSettings::get_global(cx).search_wrap
-                && ((direction == Direction::Next && index + count >= matches.len())
-                    || (direction == Direction::Prev && index < count))
+                && ((direction == Direction::Next && new_match_index < index)
+                    || (direction == Direction::Prev && new_match_index > index))
             {
                 crate::show_no_more_matches(window, cx);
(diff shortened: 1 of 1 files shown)


stanhua:
I tested only with `cargo test` locally, but didn't notice there's `cargo nextest` which has an excellent test for this.

The test fails because the behavior doesn't match Vim's. The difficulty is that `self.active_match_index` doesn't tell whether the cursor is at a match, or between two matches. If it's between two matches, the `active_match_index` points to the next match. In order to match Vim's behavior closely, I think we'd need to be able to distinguish between these two cases.

Consider the following state with `search_wrap == false` and a search for `match`:

```
match
ˇfoobar
match
```

If we press 'n' for next match, the original behavior doesn't move to the second 'match' because 'active_match_index` is already at it, even though the cursor is not. Vim does move, and so does this PR's version.

But if we press `2 n` Vim doesn't move anywhere, because there are not enough matches. This PR would move, because it does a kind of saturating move towards the end of matches. There is a test case for this in `test_move_to_next_with_no_search_wrap` where count is 3, and this PR doesn't pass that test.

I think it's not possible to match Vim's behavior with the information available at `select_match`. Is it important to have the same behavior as Vim when count >= 2 and there are not that many matches left?

ConradIrwin:
@stanhua Thanks for taking a look at this! I agree that the original bug you describe needs fixing.

Vim already overrides quite a bit of search functionality, could it be fixed by looking at the buffer to see whether the cursor is on the current match and handling that specially?

ConradIrwin:
Closing for now, but would be interested in follow-up PRs for this.

PR #42453, opened 2025-11-11
Add support for MCP tools/list_changed notification

tsoernes:
## Summary

This PR adds support for the MCP (Model Context Protocol) `notifications/tools/list_changed` notification, enabling dynamic tool discovery when MCP servers add, remove, or modify their available tools at runtime.

## Release Notes:

- Improved: MCP tools are now automatically reloaded when a context server sends a `tools/list_changed` notification, eliminating the need to restart the server to discover new tools.

## Changes

- Register a notification handler for `notifications/tools/list_changed` in `ContextServerRegistry`
- Automatically reload tools when the notification is received
- Handler is registered both on initial server startup and when a server transitions to `Running` status

## Motivation

The MCP specification includes a `notifications/tools/list_changed` notification to inform clients when the list of available tools has changed. Previously, Zed's agent would only load tools once when a context server started. This meant that:

1. If an MCP server dynamically registered new tools after initialization, they would not be available to the agent
2. The only way to refresh tools was to restart the entire context server
3. Tools that were removed or modified would remain in the old state until restart

## Implementation Details

The implementation follows these steps:

1. When a context server transitions to `Running` status, register a notification handler for `notifications/tools/list_changed`
2. The handler captures a weak reference to the `ContextServerRegistry` entity
3. When the notification is received, spawn a task that calls `reload_tools_for_server` with the server ID
4. The existing `reload_tools_for_server` method handles fetching the updated tool list and notifying observers

This approach is minimal and reuses existing tool-loading infrastructure.

## Testing

- [x] Code compiles with `./script/clippy -p agent`
- The notification handler infrastructure already exists and is tested in the codebase
- The `reload_tools_for_server` method is already tested and working

## Benefits

- Improves developer experience by enabling hot-reloading of MCP tools
- Aligns with the MCP specification's capability negotiation system
- No breaking changes to existing functionality
- Enables more flexible and dynamic MCP server implementations

## Related Issues

This implements part of the MCP specification that was already defined in the type system but not wired up to actually handle the notifications.


franciskafyi:
Hey @tsoernes, thank you for this! I'm not so familiar with this – would you be able to share more context / screenshots for how this is done in e.g. Cursor or VS Code or similar? 🙏

tsoernes:
> Hey @tsoernes, thank you for this! I'm not so familiar with this – would you be able to share more context / screenshots for how this is done in e.g. Cursor or VS Code or similar? 🙏

https://github.com/microsoft/vscode/blob/2365ea12ec47729ab54e257313180c868d1073a7/src/vs/workbench/contrib/mcp/common/mcpServer.ts#L270

agu-z:
@tsoernes Thank you for your contribution! The changes look good overall. I did notice that the existing MCP client didn’t have a way to unregister notification handlers, which meant they could leak when a new agent instance was created. This is now handled by tying the handler’s lifetime to the registry, which is itself scoped to the agent’s lifetime.

PR #42457, opened 2025-11-11
vim: Fix :s command ignoring case sensitivity settings

Kalmaegi:
Closes #36260 

This PR fixes the vim :s Command Ignores Case Sensitivity Settings

Release Notes:

- N/A

### DIFF
--- a/crates/vim/src/normal/search.rs
+++ b/crates/vim/src/normal/search.rs
@@ -511,7 +511,12 @@ impl Vim {
                     search_bar.is_contains_uppercase(&search),
                 );
             } else {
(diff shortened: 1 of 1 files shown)


ConradIrwin:
Thanks! 

I think `s:/` in vim should default to case-sensitive, but editor.case_sensitive defaults to false.

This change seems strictly better than what we have today, but I wonder if we shouldn't just hard-code `true` here.

An alternative would be to make the default value of case_sensitive null, and default it to true in vim mode (or false outside of vim mode) if the user hasn't set it. (If we do that there are a few other places we could also make use the setting instead of hard-coding).

PR #42461, opened 2025-11-11
Fix circular reference issue around PopoverMenu

feeiyu:
Follow up to https://github.com/zed-industries/zed/pull/42351

Release Notes:

- N/A


SomeoneToIgnore (approved):
Thank you again.

PR #42467, opened 2025-11-11
Send Code to Terminal (Interpreted language support)

### LINKED ISSUE #8344: Run single function
### Check for existing issues

- [X] Completed

### Describe the feature

So you'd highlight a function via highlighting it with your mouse and then hit something like F6 (or maybe right click its name, whatever) and then you'd be prompted in the terminal to input the values of all the arguments and it would show you what the function returned as well as outputted (etc.). Useful debugging feature.

_No response_

### LINKED ISSUE #12598: Run shell command on selected region
### Check for existing issues

- [X] Completed

### Describe the feature

Add ability to run shell command on selected region.

- Display the output
- Optionally insert the output into buffer or replace the selected region with the output
- Should support PIPE

Use cases:
- Text manipulations using UNIX tools, greatly extend the ability of Zed using existed tools, existed issues: https://github.com/zed-industries/zed/issues?q=is%3Aissue+text+manipulation
- Format whatever selected buffer content
  - Format a JSON string in a Python string literal
  - Format JS in HTML file 
  - Format shell snippets in markdown files
- Get output of existed tools
  - Insert formatted `date` into current buffer
  - Insert a `ls -` like grid into current buffer

Below is an example of how it work in Emacs

#### Select a region and invoke the command

In Emacs, the shortcut is: `M-|`

Screenshot 2024-06-03 at 17 26 32

#### Display output in somewhere (other window,  popup etc.)
Screenshot 2024-06-03 at 17 26 37

#### Optional insert into buffer or replace the selected region with the output

In Emacs, the shortcut is `C-u M-|`

Screenshot 2024-06-03 at 17 27 00

Another real world use case

Screenshot 2024-06-03 at 17 30 44

Screenshot 2024-06-03 at 17 30 55

_No response_

### LINKED ISSUE #16102: Add run-selected-text to run (Julia) code snippets in the terminal
### Check for existing issues

- [X] Completed

### Describe the feature

I am writing this feature request with the focus on Julia, but it is certainly not restrited to it.

Form my understanding this feture request ist neither covered by #5141 nor by #12598. The zed julia extension https://github.com/JuliaEditorSupport/zed-julia/issues/4#issuecomment-2130935682 seems to require a full project structure with toml file including package and version dependencies. In the testing stage of some Julia code it were advantageous to have the option to just select some code in the editor and execute it in the terminal next to it.  Nothing else. A similar feature was implemented in Pulsar with in the "x-terminal" package: The Pulsar command to perform this action is: `x-terminal:run-selected-text` and its main advantage is its simplicty. If connected to a shortcut, e. g. CTRL + ALT + SHIFT + ENTER, it can be easily integrated into a quick testing procedure.

image

I guess this feature is not yet implemented in Zed.

_No response_

JosiahParry:
This PR provides a new command `SendToTerminal` which uses the active LSP to capture the current expression, send it to the terminal, and continue the cursor movement. 

The intent of this PR is to provide "real" support for interpreted languages like R, Julia, and Python and mimic code execution and cursor navigation that one gets in IDEs like RStudio, Positron, and others. 

- Closes https://github.com/zed-industries/zed/issues/8344 (in spirit)
- Closes https://github.com/zed-industries/zed/discussions/8674
- Closes https://github.com/zed-industries/zed/issues/16102
- Closes https://github.com/zed-industries/zed/issues/12598

I don't think this is _perfect_ but I think it is much better than the status quo. I think, also, that this may be best handled by language-specific extensions. However, based on my understanading of the Language extension, it doesn't provide an entrypoint for performing such tasks

Release Notes:

- Adds new command `SendToTerminal` to send active expression to the active terminal supporting dynamic code execution of interpreted languages such as R, Python, Julia, and more.

## Python Example

Screen Recording 2025-11-11 at 08 15 13

## R example

Screen Recording 2025-11-11 at 08 22 52

## Julia Example

Screen Recording 2025-11-11 at 08 16 49

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -35,6 +35,7 @@ mod persistence;
 mod rust_analyzer_ext;
 pub mod scroll;
 mod selections_collection;
+pub mod send_to_console;
 pub mod tasks;
 
 #[cfg(test)]
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -35,7 +35,6 @@ mod persistence;
 mod rust_analyzer_ext;
 pub mod scroll;
 mod selections_collection;
-pub mod send_to_console;
 pub mod tasks;
 
 #[cfg(test)]
--- a/crates/zed/src/zed.rs
+++ b/crates/zed/src/zed.rs
@@ -6,6 +6,8 @@ pub(crate) mod mac_only_instance;
 mod migrate;
 mod open_listener;
 mod quick_action_bar;
+mod send_to_terminal;
+use send_to_terminal::send_to_terminal;
 #[cfg(target_os = "windows")]
 pub(crate) mod windows_only_instance;
 
@@ -18,7 +20,6 @@ use breadcrumbs::Breadcrumbs;
 use client::zed_urls;
 use collections::VecDeque;
 use debugger_ui::debugger_panel::DebugPanel;
-use editor::ToOffset;
 use editor::{Editor, MultiBuffer};
 use extension_host::ExtensionStore;
 use feature_flags::{FeatureFlagAppExt, PanicFeatureFlag};
@@ -78,7 +79,6 @@ use util::rel_path::RelPath;
 use util::{ResultExt, asset_str};
 use uuid::Uuid;
 use vim_mode_setting::VimModeSetting;
-use workspace::Panel;
 use workspace::notifications::{
     NotificationId, SuppressEvent, dismiss_app_notification, show_app_notification,
 };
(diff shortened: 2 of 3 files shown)


JosiahParry:
@cla-bot check

cameron1024:
Thanks for the contribution 🙏 I've wanted something like this for a while

I'm wondering about what the exact behaviour should be here. I think in the case where you have a selected range, it's pretty obvious. That said, I'm not sure how I feel about automatically inferring the range to send based on the cursor position:
- There are many different reasonable approaches (current line, current syntax node, syntax ancestor matching some predicate), and people may disagree about what it should be
- If we went with a treesitter-based approach, we'd probably want some way for language extension maintainers to provide their own capture groups (similar to how they currently do with highlights/injections/etc.)
- It's not clear how it interacts with vim mode, where I think users would expect to be able to use this action with various vim textobjects

I don't think these issues are necessarily unsolvable, but I do wonder if we can avoid having to think too hard about them by only allowing this action if there is a selection.

I also think we should make sure we support both REPL and non-REPL terminals, by which I mean that it seems like the current implementation unconditionally appends a newline to the text, which makes sense in a REPL context, but, when I'm in Neovim, I most commonly use this feature to send a function/module name to the terminal to add to the end of a `cargo test` invocation. In this case, the trailing newline is probably a bug, since I often want to add more arguments.

Let me know if you have any questions, and feel free to book some time in [my calendar](https://calendar.app.google/J7opPuVvHspsLCm37) if you want to discuss in person (FYI after this week I'll be on holiday until the new year, but ofc will be happy to pick up where we left off).

Thanks again for working on this 😁

JosiahParry:
@cameron1024 thank you so much for actually reviewing this! I think I definitely agree with you—mostly! 

I think if the capability is limited to **only** selected text without the ability to have a "natural" / intuitive selection and auto-continuation the feature would be fairly blasé and of little use when compared to other IDEs. 

I think the following would be great: 

- basic SendToTerminal functionality for selected text
- modify language extensions to have an optional trait implementation that can use tree-sitter to enable node selection and cursor continuation

I'd love to sync up in the new year! I'll put a reminder on my calendar to put some time on yours closer to the new year so your 2026 availability shows up :)

duodenum96:
Chiming in as someone who is _really_ looking forward to this feature, it would be great if there was something like "vap" in vim, which instead of selects the paragraph, just sends the paragraph to the terminal. In VSCode, the "shift-enter" for python (or "ctrl-enter" for R) does "smart-send", for example, if you "shift-esc" inside the function definition it sends the whole function definition to the terminal.

cameron1024:
> it would be great if there was something like "vap" in vim

That's exactly what I'm anticipating. I'd expect that we'd end up with:
- an `editor: send to terminal` action that does the automatic syntax-tree detection or uses the current selection
- a `vim: send to terminal` action that waits for a text object in normal mode, or uses the current selection in visual mode

I would likely bind `st` (send terminal) to the second one, which would allow `stap` to send a paragraph, `staf` to send a function, `stac` to send a class/struct/etc. . But there shouldn't be anything stopping you binding both actions, even if you use vim mode.

flowerornament:
See here for a hack to get similar behavior: https://github.com/zed-industries/zed/issues/16102#issuecomment-3735144253, with some unexpected benefits.

arnold-c:
Very exciting to see this. From an IDE perspective, I think Posit do this in a manner a good number of folks will be familiar with.

As a possible middle ground, I have been using nvim daily for the last couple of years and find that just inserting code chunk markers (e.g., `#%%`) in relevant places allows me to send everything between the markers using [vim-slime](https://github.com/jpalardy/vim-slime). It's very rare that I actually need to manually select sections of code, and when I do I usually use treesitter nodes through keymaps like `shift-space/backspace` to `editor::SelectLargerSyntaxNode`, or the very useful [flash.nvim](https://github.com/folke/flash.nvim) `shift-s` command to select treesitter blocks in one go.

rgbkrk:
Obviously not the terminal based approach but I used treesitter to detect blocks of code to run with the inline REPL in https://github.com/zed-industries/zed/pull/49636.

GitGlimpse895:
Hey @cameron1024 — just checking in on the state of this since @JosiahParry stepped back in January.

The spec from your pairing session (sendables.scm, editor::SendToTerminal, vim::SendToTerminal, GotoNextTerminalSendable) is really well thought out. Is this still the intended path forward, or has the inline REPL work in #49636 changed the direction? There's also #53221 open now which seems to be tackling the same space.

Would be good to know if this PR should be closed in favour of one of those, or if the spec is still open for someone to pick up.

cameron1024:
I don't feel like the REPL support should impact this feature too much. There are non-REPL reasons to want ot send code to the terminal - actually, that's part of the reason why I'm keen on a more use-case-agnostic. #53221 does implment it, but I've left feedback on that since I still think some of the original design constraints apply.

But yes, we can close this PR in favour of #53221 , which is on the right lines.

PR #42486, opened 2025-11-11
Add UI for managing the configured remotes on a repository

### LINKED ISSUE #26559: Add UI for managing the configured remotes on a repository
Zed should be there for every step from the first blank directory to the first PR. We can't do that until we have remotes.

bnjjj:
https://github.com/user-attachments/assets/af090410-c00a-45a2-8541-cb49819227ed

Need feedback on this UI

Closes #26559

Release Notes:

- Added support for git remotes


cole-miller:
Thanks!

Just commenting on the backend side of this for now, adding and removing remotes looks good, but I don't think it makes sense to have a per-repository current remote, i.e. I don't think we should have `GitRepository::remote` and `GitRepository::change_remote`.

ghost:
It'd be nice to see in either this PR (or a separate one) support creating and deleting branches to the Command Palette as well.

bnjjj:
Got it @cole-miller ! How would you suggest saving the selected remote ? Should it be persistent if I close and reopen zed ?

cole-miller:
I guess I'm not sure that we need a concept of a per-repository selected remote at all--could you say what it's used for in your design? My instinct is that this is not really a concept that git itself has, so we shouldn't represent it at the layer of the `GitRepository` trait, and depending on the design we might not need it at all.

bnjjj:
@cole-miller I agree but I also think we have to save this selected remote somewhere because it would be very annoying to have to select again the same remote everytime you re-open zed. That's why I added this method at git level because I use the git config file to configure this. We can save it somewhere else like in persistent storage or somewhere else. I mainly think about the experience we want to provide here

Angelk90:
@cole-miller : What do you think of a similar style:

https://private-user-images.githubusercontent.com/20476002/166105436-c29a3653-1e58-444e-af64-66debf46323c.mov

bnjjj:
@Angelk90 I think it's a different purpose. In any cases we will provide a way to manage remotes. My question is more about the workflow here if we should have a notion of default selected remote or not. For example a use case I had previously was that I had a private and public repository, every big feature was first introduced in the private repository so I always had to push on my private remote an never on the public one. But once it was public I always had to push and fetch on the public remote. That's why I wanted a way to select the current remote on which we want to fetch/push commits because it would be a very bad experience to always have to specify my remote to push and pull. But maybe my use case is not very usual and is not worth it

Angelk90:
@bnjjj: Everyone has different specific cases, that's the beauty of it.

Mine is different, I had origin and upstream.
- From `upstream` I pushed and fetched to my forked repository.
- From `origin` I fetched to the original repository to test the pr

mattermill:
Hey @bnjjj, thanks for taking a crack at this. Coming at this from a design perspective, I think this problem is wider-reaching than only the selector in the git panel. Ideally we should use the same branch picker that opens when you run `git: branch` from the command palette.

As for the workflow for adding remotes, I've put this together for consideration:
- Paste a remote URL
- Interpret the URL (likely with with a loading state)
- Either use the URL as a branch name (but why?!) or...
- Add remotes from the URL

At that point we have the remotes and can add them to the branch picker, as if they were added through the CLI. Now we can delineate the local and remotes inside the branch picker, but I've added a filter for showing only remotes.

I'm going to close this for now because I think the problem requires a different approach, but feel free to keep the discussion alive if you'd like to work on implementing these changes or have questions/feedback.

Thanks again!

Figma-RndSTDLV@2x

bnjjj:
@mattermill I would like to continue my work on this. I needed guidance and you provided what I needed so I'm unblocked right now and will follow your instructions and mockups. Thank you

Angelk90:
@mattermill : In the case of github, if the entered value contains the text `gitbub.com` as part of the text I would avoid displaying `Create branch...based off..`, it could cause problems.

Screenshot 2025-11-13 alle 12 53 48

bnjjj:
@mattermill I implemented your design in https://github.com/zed-industries/zed/pull/42819 I think there's a missing story here is when you want to remove a specific remote

https://github.com/user-attachments/assets/e2f54dda-a78b-4d9b-a910-16d51f98a111

PR #42494, opened 2025-11-12
Implement pretty TypeScript errors

### LINKED ISSUE #7844: Pretty TypeScript errors
### Check for existing issues

- [X] Completed

### Describe the feature

TypeScript errors should be formatted in a similar fashion as [pretty-ts-errors](https://github.com/yoavbls/pretty-ts-errors?tab=readme-ov-file). 

image

image

KyleBarton:
Closes #7844

This change uses tree-sitter highlights as a method of showing typescript errors prettily, keeping regex as simple as possible:

Screenshot 2025-11-11 at 3 40 24 PM

It covers three main areas:

1. Diagnostics

Diagnostics are now rendered with language-aware typescript, by providing the project's language registry.

2. Vtsls

The LSP provider for typescript now implements the `diagnostic_message_to_markdown` function in the `LspAdapter` trait, so as to provide Diagnostics with \`\`\`typescript...\`\`\`-style code blocks for any selection of typescript longer than one word. In the single-word case, it simply wraps with \`\`

3. Typescript's `highlights.scm`

`vtsls` doesn't provide strictly valid typescript in much of its messaging. Rather, it returns a message with snippets of typescript values which are invalid. Tree-sitter was not properly highlighting these snippets because it was expecting key-value formats. For instance:
```
type foo = { foo: string; bar: string; baz: number[] }
```
is valid, whereas simply
```
{ foo: string; bar: string; baz: number[] }
```
is not.

Therefore, highlights.scm needed to be adjusted in order to pattern-match on literal values that might be returned from the vtsls diagnostics messages. This was done by a) identifying arrow functions on their own, and b) augmenting the `statment_block` pattern matching in order to match on values which were clearly object literals.

This approach may not be exhaustive - I'm happy to work on any additional cases we might identify from `vtsls` here - but hopefully demonstrates an extensible approach to making these messages look nice, without taking on the technical burden of extensive regex.

Release Notes:

- Show pretty TypeScript errors with language-aware Markdown.


KyleBarton:
Paired with @ConradIrwin for review - merging this one in

DavidBK:
Amazing work! The hover popups should support keyboard navigation such as Page Up and Page Down (or even custom key) for long typescript errors.

PR #42497, opened 2025-11-12
docs: Update basedpyright settings examples

CnsMaple:
The [example](https://docs.basedpyright.com/latest/configuration/language-server-settings/#zed) on the official website of basedpyright is correct.

Release Notes:

- Update basedpyright settings examples


MrSubidubi (approved):
Thanks! Looked into this yesterday and the current state might actually work as well, but let's better follow the official documentation here.

PR #42508, opened 2025-11-12
vim: Fixed toggle relative lines only in full editor

0x2CA:
Closes #ISSUE

Because currently relative lines are only displayed in full, and in the future, as auto height and single line modes become more common in Vim, they will cause interference.

Release Notes:

- Fixed toggle relative lines only in full editor

### DIFF
--- a/crates/vim/src/vim.rs
+++ b/crates/vim/src/vim.rs
@@ -1394,8 +1394,9 @@ impl Vim {
         });
         let editor = editor.read(cx);
         let editor_mode = editor.mode();
+        let is_full = editor_mode.is_full();
 
-        if editor_mode.is_full()
+        if is_full
             && !newest_selection_empty
             && self.mode == Mode::Normal
             // When following someone, don't switch vim mode.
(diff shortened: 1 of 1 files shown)


dinocosta:
Hey @0x2CA 👋 
Can you share a little bit more information behind this change? Is it causing any issues at the moment? 🙂

0x2CA:
https://github.com/user-attachments/assets/ab78c3f7-91b0-4dda-854e-e8c9bbbcd324

Currently, when the git panel is activated, the editor activates absolute line numbers. In the future, when renaming is implemented in vim mode, the editor will also activate absolute line numbers.

dinocosta:
> 2025-11-12.19.51.59.mov 
> Currently, when the git panel is activated, the editor activates absolute line numbers. In the future, when renaming is implemented in vim mode, the editor will also activate absolute line numbers.

Is the video you shared running your changes? With relative line numbering enabled in the latest release (0.211.6) the relative line numbers remain active ▼

https://github.com/user-attachments/assets/44345c3e-ea38-449d-a331-202e3be68b92

Personally I don't yet see a situation where this would be an issue so will likely wait and see if we actually ship single line editor vim support to then think about this. Thanks! 🙂

0x2CA:
It's not that it won't work in a full editor, but it will when going from one vim editor to another, and another vim editor that isn't full will make the old one set to absolute line numbers

PR #42513, opened 2025-11-12
vim: Move to opening html tag from / in closing tag

### LINKED ISSUE #41582: vim: Getting stuck in end tag when jumping between html tags using % (vim::Matching)
### Summary

In vim mode when using the % keystroke (vim::Matching) on an html start tag the cursor jumps to the end tag as expected, but when doing % again it jumps from `/` of the end tag to its `>` instead of jumping back to the start tag again.

### Description

Noticed this behavior for plain HTML but also for markup in JSX/TSX files. 

Steps to reproduce:
1. Enable vim mode
2. Create html file
3. Insert `<html></html>`
4. Move cursor inside start tag
5. Hit `%` keystroke multiple times

**Expected Behavior**: Cursor jumps back and forth between start and end tag.
<details><summary>Video example</summary>

https://github.com/user-attachments/assets/c473c690-54d0-437c-8415-f7e640fd3367

</details>

**Actual Behavior**: Cursor jumps to end tag and then "gets stuck" jumping between angle brackets of end tag.

<details><summary>Video example</summary>

https://github.com/user-attachments/assets/dc4b5dba-e67f-4038-a83c-78fa48b38e6b

</details>

Zed: v0.210.4 (Zed) 
OS: macOS 26.0.1
Memory: 16 GiB
Architecture: aarch64

Pranav2612000:
Closes #41582

Release Notes:

- Improves the '%' vim motion for html by moving the cursor to the opening tag when its positioned on the `/` ( slash ) of the closing tag

### DIFF
--- a/crates/vim/src/motion.rs
+++ b/crates/vim/src/motion.rs
@@ -2388,10 +2388,16 @@ fn matching(map: &DisplaySnapshot, display_point: DisplayPoint) -> DisplayPoint 
         .or_else(|| snapshot.innermost_enclosing_bracket_ranges(offset..offset, None));
 
     if let Some((opening_range, closing_range)) = bracket_ranges {
-        if opening_range.contains(&offset) {
-            return closing_range.start.to_display_point(map);
-        } else if closing_range.contains(&offset) {
-            return opening_range.start.to_display_point(map);
+        let mut chars = map.buffer_snapshot().chars_at(offset);
+        match chars.next() {
+            Some('/') => {}
+            _ => {
+                if opening_range.contains(&offset) {
+                    return closing_range.start.to_display_point(map);
+                } else if closing_range.contains(&offset) {
+                    return opening_range.start.to_display_point(map);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Pranav2612000:
This works, but not sure if this is the best way to handle it. I think the LSP returns the range for the opening and closing tags which includes both the `<` and the `/` together.

P1n3appl3:
Sorry for the delay in reviewing. This change looks good, and I think it's fine to hack it in just for that case. Could you add a test or at least an extra test case to an existing one like [this](https://github.com/zed-industries/zed/blob/a7fff59136c2a53a92ad4dacf37b5c69cdfe7b4b/crates/vim/src/motion.rs#L3403) to verify that our behavior matches neovim's in this case?

Pranav2612000:
@P1n3appl3 Thank you for the review. I've updated the commit to include a test now( added to the existing test case - thought that should be enough ). Can you please take another look whenever you have time?

Edgarfp:
Finally fixed! Thank you!

PR #42523, opened 2025-11-12
git: Continue parsing other branches when refs have missing fields

errmayank:
Closes #34684

Release Notes:

- (Let's Git Together) Fixed Git panel not showing any branches when repository contains refs with missing fields


cole-miller:
Thanks!

PR #42526, opened 2025-11-12
terminal: Add ctrl+click link detection with mouse movement

nihalxkumar:
Closes #41994

This PR introduces Element-bounded drag tolerance for Ctrl/Cmd+click in terminal. 

Previously, Ctrl/Cmd+click on terminal links required pixel-perfect accuracy. Any mouse movement during the click would cancel the navigation, making it frustrating to click on links, especially on high-DPI displays or with sensitive mice.

Users can now click anywhere within a clickable element (file path, URL, hyperlink), drag the cursor anywhere within that same element's boundaries and release to trigger navigation

Implementation:

  - Stores detected element metadata (`text` and `grid_range`) on Ctrl/Cmd+mouse-down
  - Tracks cursor position during drag, preserving click state while within element bounds
  - Verifies element match on mouse-up before triggering navigation
  - Uses existing `find_from_grid_point()` for element detection

Before:

[before.webm](https://github.com/user-attachments/assets/ee80de66-998e-4d8e-94d0-f5e65eb06d22)

After:

[after.webm](https://github.com/user-attachments/assets/7c9ddd9e-cfc1-4c79-b62c-78e9d909e6f4)

Release Notes:

- terminal: Fixed an issue where `ctrl|cmd+click` on links was very sensitive to mouse movement. Clicking links now tolerates mouse movement within the same clickable element, making link navigation more reliable


nihalxkumar:
@cla-bot check

nihalxkumar:
@cla-bot check

nihalxkumar:
Updated PR description

probably-neb:
Great work! Tested it and it feels much better than before. 

cc @davewa who has been doing a bunch of work in this area recently. How does the code look to you?

crates/terminal/src/terminal.rs
```diff
@@ -386,6 +386,8 @@ impl TerminalBuilder {
             is_remote_terminal: false,
             last_mouse_move_time: Instant::now(),
             last_hyperlink_search_position: None,
+            mouse_down_link: None,
```
davewa: I think this may be unnecessary because `find_from_grid_point()` also detects uri_s from cell.hyperlink()?

crates/terminal/src/terminal.rs
```diff
@@ -1748,7 +1754,33 @@ impl Terminal {
...
+                    self.last_content.display_offset,
+                );
+
+                // Check if current position is within the stored element's character bounds
+                let within_element = point.line >= element_range.start().line
+                    && point.line <= element_range.end().line
+                    && point.column >= element_range.start().column
+                    && point.column <= element_range.end().column;
+
+                if !within_element {
```
davewa: This can be just: `if !element_range.contains(point) {`

crates/terminal/src/terminal.rs
nihalxkumar: Thanks for reviewing. Yes, we should remove it.

crates/terminal/src/terminal.rs
nihalxkumar: Yes, totally. Thanks!

crates/terminal/src/terminal.rs
```diff
@@ -831,6 +833,7 @@ pub struct Terminal {
     is_remote_terminal: bool,
     last_mouse_move_time: Instant,
     last_hyperlink_search_position: Option<Point<Pixels>>,
+    mouse_down_detected_element: Option<(String, Match)>,
```
davewa: I'd prefer this be named `mouse_down_hyperlink`

davewa: Making this match the return type of `find_from_grid_point` simplifies the rest of the code, so `Option<(String, bool, Match)>`

crates/terminal/src/terminal.rs
```diff
@@ -1872,6 +1898,41 @@ impl Terminal {
...
+                            &mut self.hyperlink_regex_searches,
+                        )
+                    {
+                        // Verify same element: matching text and overlapping ranges
+                        let ranges_overlap = down_range.start().line <= up_range.end().line
+                            && down_range.end().line >= up_range.start().line
+                            && down_range.start().column <= up_range.end().column
+                            && down_range.end().column >= up_range.start().column;
+
+                        if down_text == up_text && ranges_overlap {
```
davewa: What is the use case for overlapping? I think it can just be equality here, and with the type being `(String, bool, Match)`, this simplifies to just
```
if self.mouse_down_hyperlink == mouse_up_hyperlink {
```

davewa (commented):
Thanks for doing this work! I thought of a couple more changes that I think will improve/simplify this. Also, I think some of the comments just repeat the code, I'd prefer those be removed. Otherwise, looks good.

davewa (commented):
Also, I'd love to see a couple unit tests for this using `VisualTestContext::simulate_event()` 😎

crates/terminal/src/terminal.rs
```diff
@@ -1872,6 +1893,31 @@ impl Terminal {
...
+                {
+                    let term_lock = self.term.lock();
+                    if let Some(mouse_up_hyperlink) = terminal_hyperlinks::find_from_grid_point(
+                        &term_lock,
+                        point,
+                        &mut self.hyperlink_regex_searches,
+                    ) {
+                        if mouse_down_hyperlink == mouse_up_hyperlink {
+                            self.events
+                                .push_back(InternalEvent::FindHyperlink(position, true));
```
davewa: Looking at this function and the implementation of `FindHyperlink`, we should extract a method and call it both here and there to avoid always calling `grid_point` and `find_from_grid_point` twice, say,

```
impl Terminal {
    fn process_hyperlink(&mut self, hyperlink: (String, bool, Match), open: bool) { ... }
}
```

davewa (commented):
You may have figured out that my review style is iterative. I hope you are having as much fun as I am, it is shaping up nicely! I noticed that the use of `FindHyperlink` here results in us always calling `find_from_grid_point` twice in a row on the same position, I added a suggestion there.

nihalxkumar:
I really appreciate the reviews from you. Please keep them coming, they makes things better for everyone <3

probably-neb:
Going to mark this as draft while you're addressing the review comments. Please request my review when you think this is ready @nihalxkumar!

nihalxkumar:
the checks appear to be flaky

probably-neb:
I'm only seeing tests you added failing. Are you referring to a different test flaking? Or just calling out that the tests you've added seem to be flaky?

nihalxkumar:
I saw the Mac test failing on another PR. I am checking what needs to fixed here brb

davewa:
Thinking about this more, it would be simpler, more dependable, and still give us the test coverage we are looking for here to have these tests just call Terminal::mouse_* methods directly...

crates/terminal/src/terminal.rs
```diff
@@ -1876,6 +1915,31 @@ impl Terminal {
                 self.copy(Some(true));
             }
 
+            if let Some(mouse_down_hyperlink) = self.mouse_down_hyperlink.take() {
+                let point = grid_point(
+                    position,
+                    self.last_content.terminal_bounds,
+                    self.last_content.display_offset,
+                );
```
davewa, line 1940: I find it a little concerning that we now have two places that call `grid_point()` followed by `find_from_grid_point()` that differ in subtle ways that I don't fully understand the reasons for. I'm not really sure what should be done about this, but I don't feel great about it, as I suspect one or the other is incorrect.

crates/terminal/src/terminal.rs
```diff
@@ -1876,6 +1915,31 @@ impl Terminal {
...
+
+                let term_lock = self.term.lock();
+                if let Some(mouse_up_hyperlink) = terminal_hyperlinks::find_from_grid_point(
+                    &term_lock,
+                    point,
+                    &mut self.hyperlink_regex_searches,
+                ) {
```
davewa: I would prefer simpler lock management here, maybe something like:
```
if let Some(mouse_up_hyperlink) = {
    let term_lock = self.term.lock();
    terminal_hyperlinks::find_from_grid_point(
        &term_lock,
        point,
        &mut self.hyperlink_regex_searches,
    ) 
} {
```
...then the `drop()` calls can be removed...

crates/terminal/src/terminal.rs
```diff
@@ -2847,4 +2914,205 @@ mod tests {
...
+            TerminalBuilder::new_display_only(CursorShape::default(), AlternateScroll::On, None, 0)
+                .unwrap()
+                .subscribe(cx)
+        });
+
+        terminal.update(cx, |terminal, cx| {
+            terminal.write_output(b"Visit https://zed.dev/ for more\r\n", cx);
+        });
+
+        cx.run_until_parked();
```
davewa: This seems very similar for all three new tests, let's extract a method, say `fn init_ctrl_click_hyperlink_test(...)`

crates/terminal/src/terminal.rs
```diff
@@ -2847,4 +2914,205 @@ mod tests {
...
+            let up_position = point(px(10.0), px(50.0));
+
+            let mouse_down = MouseDownEvent {
+                button: MouseButton::Left,
+                position: down_position,
+                modifiers: Modifiers::secondary_key(),
+                click_count: 1,
+                first_mouse: true,
+            };
+            terminal.mouse_down(&mouse_down, cx);
```
davewa: Would be nice to extract a method for these mouse events, maybe:
```
fn ctrl_mouse_click_at(term, point)
fn ctrl_mouse_move_to(term, point)
fn ctrl_mouse_up_at(term, point)
```

crates/terminal/src/terminal.rs
```diff
@@ -2847,4 +2914,205 @@ mod tests {
...
+
+            let mouse_up = MouseUpEvent {
+                button: MouseButton::Left,
+                position: click_position,
+                modifiers: Modifiers::secondary_key(),
+                click_count: 1,
+            };
+            terminal.mouse_up(&mouse_up, &*cx);
```
davewa: It appears that the re-borrows for `mouse_up` are unnecessary?

crates/terminal/src/terminal.rs
```diff
@@ -1760,6 +1768,20 @@ impl Terminal {
     ) {
         let position = e.position - self.last_content.terminal_bounds.bounds.origin;
         if !self.mouse_mode(e.modifiers.shift) {
+            if let Some((_, _, element_range)) = &self.mouse_down_hyperlink {
```
davewa: This can be simplified to just `..,`

davewa: This should be renamed to `hyperlink_range`

davewa (commented):
Looking really nice. I just added a few more comments, we're very close now 😎.

crates/terminal/src/terminal.rs
nihalxkumar: fixed

nihalxkumar: fixed

crates/terminal/src/terminal.rs
nihalxkumar: fixed

crates/terminal/src/terminal.rs
nihalxkumar: fixed

crates/terminal/src/terminal.rs
nihalxkumar: fixed

crates/terminal/src/terminal.rs
nihalxkumar: fixed

crates/terminal/src/terminal.rs
davewa: Does not need to be part of this PR, okay to just resolve.

crates/terminal/src/terminal.rs
```diff
@@ -2847,4 +2992,72 @@ mod tests {
...
+    async fn test_hyperlink_ctrl_click_drag_outside_bounds(cx: &mut TestAppContext) {
+        let terminal = init_ctrl_click_hyperlink_test(
+            cx,
+            b"Visit https://zed.dev/ for more\r\nThis is another line\r\n",
+        );
+
+        terminal.update(cx, |terminal, cx| {
+            let down_position = point(px(80.0), px(10.0));
+            let up_position = point(px(10.0), px(50.0));
+            let terminal_bounds = terminal.last_content.terminal_bounds.bounds;
```
davewa: This can move inside `ctrl_mouse_move_to()`

davewa (commented):
One last improvement, then I'd say this is ready!

crates/terminal/src/terminal.rs
nihalxkumar: Alright. If you have get any inputs on fixing the issue you can ping me and I will work on it. :)

crates/terminal/src/terminal.rs
nihalxkumar: done

davewa:
In local testing, it doesn't seem to work anymore? One of our refactors must have regressed the feature... interesting...

nihalxkumar:
I am limited by memory. I used to check every build before committing with `cargo build -j 2`
After `4242a75a162ad0a9809936ac04cc5a9df50c3c64` I only used to check the terminal crate, clippy and rustfmt. 
Started a build now

nihalxkumar:
@davewa It's still working

[stillworking.webm](https://github.com/user-attachments/assets/7c9ddd9e-cfc1-4c79-b62c-78e9d909e6f4)

davewa:
Yeah, it's working for me too now... I must have done something dumb earlier... 🤷‍♂️

nihalxkumar:
@probably-neb It's ready

probably-neb:
Hey sorry for the delay. This got buried in my notifications inbox. I've merged main and will merge this PR once CI is green. 

Thanks for the great work @nihalxkumar!

PR #42527, opened 2025-11-12
vim: Add highlight occurrences across all visible panes for vim mode

### LINKED ISSUE #21549: Highlight all occurences across all open (visible) files
### Check for existing issues

- [X] Completed

### Describe the feature

I am a (neo)vim user that is exploring Zed and I am really happy with it so far, really good work! One thing i am missing though is to highlight/search for a string under the cursor using `*` across all open (visible) files. As it is now it highlights all occurrences but only in the current file. I usually have multiple split views open and would want to have the highlight present in all of them.

It would also be nice if it was possible to move the search bar to the bottom or even if it was possible to opt out of the ui for it.

In the screenshot below i would expect ctrl to be highlighted in both splits

Screenshot 2024-12-04 at 13 27 06

### Environment

Zed: v0.163.3 (Zed)
OS: macOS 15.1.1
Memory: 18 GiB
Architecture: aarch64

Screenshot 2024-12-04 at 21 12 06

Kalmaegi:
Closes #21549 , I’ve added a configurable option that allows, in Vim mode, highlighting matches inside other visible panes while jumping to the next match using *.
This is just an initial version of the change, but I ran into an issue:

In Vim mode, when using / to perform a search, it triggers the `search_bar` search.
If we want each keystroke during the search input to also trigger highlighting in other panes, we may need to modify this behavior within the search crate.

Would it make more sense to keep this feature Vim-mode-only, or expose it as a global setting?
Alternatively, should we consider handling it inside `buffer_search.rs`, e.g. in `on_query_editor_event`?

Release Notes:

- N/A


ConradIrwin:
Thanks for sending this, and sorry for replying so slowly!

Although it's pretty cool, I don't think this is quite the right approach:

* We shouldn't show the search bar everywhere
* It shouldn't be part of vim mode

That said, it does seem a bit confusing to build this in Zed today, where Search is very much a property of the current pane. How does VSCode solve the discrepancy?

PR #37428, opened 2025-09-03
Add setting to show/hide title bar

bgub:
Closes #5120

Release Notes:

- Added settings for hiding and showing title bar

https://github.com/user-attachments/assets/aaed52d0-6278-4544-8932-c6bab531512a

Issue #5066, opened 2022-12-23
Option to combine the tabs-bar into the titlebar


jvanmourik:
### Check for existing issues

- [X] Completed

### Describe the feature

It would be nice to have an option which places the file tabs inside the titlebar. This would reduce the footprint of the editor and be in line with Zed's "Your editor should disappear" design goal. 

### If applicable, add mockups / screenshots to help present your vision of the feature

![frame](https://user-images.githubusercontent.com/5372709/209402946-431592f5-a33c-4291-aeb5-0c4e9071b03d.png)

dmkjfs:
+1, pretty annoying to see 2 or 3 bars that use space, instead of just one. I usually don't use the tab bar at all when im on laptop

<img width="1921" height="1081" alt="with tab bar" src="https://github.com/user-attachments/assets/b3722815-41c6-4ab8-8e1b-8b7491d811e0" />
<img width="1918" height="1035" alt="without tab bar" src="https://github.com/user-attachments/assets/95fb56f9-83e5-41e8-a54a-1a8f30957b93" />

morgankrey:
Hi all, closing this under the umbrella of "Zed won't do". I want to explain our thinking since this has been open for a long time and has clear interest.

- Our design direction does not agree with the premise. We're planning to put more UI into the title bar over, time, not less, and collapsing tabs into that space would conflict with where we're headed.
- There are also real technical challenges.
    - Tabs are per-pane, but the title bar is per-window. With pane splits, there's no clean way to represent multiple tab bars in a single title bar. 
    - The title bar also already contains traffic lights, branch info, and other controls. We briefly shipped a hide-title-bar setting ([37428](https://github.com/zed-industries/zed/pull/37428)) and had to revert it ([38756](https://github.com/zed-industries/zed/pull/38756)) due to macOS rendering issues and conflicts with the redesign work.

We'd still consider a high-quality PR from the community, but the bar is high for consideration. Making this work gracefully across macOS/Linux/Windows, all pane configurations, collab, and the various UI states we support is a significant maintenance burden - every future title bar change would need to account for this mode - which likely means a significant refactor (and thus, unlikely to merge).

[...]

PR #38756, opened 2025-09-24
Revert "Add setting to show/hide title bar (#37428)"

ConradIrwin:
Closes https://github.com/zed-industries/zed/issues/38547

Release Notes:

- Reverted the ability to show/hide the titlebar. This caused rendering bugs on
  macOS, and we're preparing for the redesign which requires the toolbar being present.

PR #42540, opened 2025-11-12
Add option to toggle the title bar

### LINKED ISSUE #5120: Add options to hide title and status bar
### Description

Usually I feel some components of zed editor(title and status bar) are redundant to me, and I just want to focus on the editor itself.  It's something called zen mode in vscode.

It's something similar to zed-industries/zed#5333, but I want to hide more..

Here is the components I want to be able to hide:
image

Here is how vscode shows:
image

xipeng-jin:
Closes #5120

### Summary
Add a option in settings `"title_bar": { "show": true }` to configure whether or not to show the title bar in the editor. Also add the action `zed::ToggleTitleBar` to toggle the current display state of the title bar.

### Description
The previous PR #37428 got reverted because of macOS traffic lights issue was outstanding and there are new plans around redesign the title bar which makes it a must need from Zed team. 

First, the macOS traffic light issue has been fixed in this PR which means they will be hidden/displayed with the state of title bar. Second, I believe the title bar is an essential to show when you are using collaboration features even for now. But on the other hand those features are never turned on for the solo mode or even offline mode. Therefore I think the better way is to have the Action with shortcut/keybinding to trigger toggling the title bar in the runtime. People can turn it on/off as needed. In the Zen mode, this feature enabling to hide title bar is also essential to achieve the minimal UI layout. I simplify the title bar setting `tittle_bar.show` with only containing two states `true`/`false` to minimize the complexity.

This PR will also part of addressing the problem in the issue #5066.

Here is a short demo

https://github.com/user-attachments/assets/400a10ee-8fb3-4036-97ee-5ec3b167b096

Release Notes:

- Added two states `true`/`false` for configuring whether or not to show the title bar in the editor with setting `title_bar.show`.
- For macOS compatibility, added the functionality for hiding macOS specific traffic light when hiding the title bar. The traffic light will be toggled with the title bar.
- Added a zed action `zed::ToggleTitleBar` to toggling the state (show/hide) of title bar in runtime.


maxdeviant:
Thank you for the PR, but we're not currently accepting any PRs to hide the title bar.

As stated in https://github.com/zed-industries/zed/pull/38756:

> we're preparing for the redesign which requires the toolbar being present.

xipeng-jin:
Thank you @maxdeviant for your clarification.

ghost:
> Thank you for the PR, but we're not currently accepting any PRs to hide the title bar.
> 
> As stated in #38756:
> 
> > we're preparing for the redesign which requires the toolbar being present.

This looks really awful for anyone on a tiling WM. What's the problem with options? I'd love to use Zed, but if you're adamant about this design decision, I'll be looking elsewhere. Very disappointing stance for open source software.

boozook:
I'm going out too. This is not a single case, not so critical as other PRs in the same "state", but important for me and directions  of maintaining are scaring me, so already going to __the__ better editor.

PR #42542, opened 2025-11-12
Implement vertical scrolling for extended keymap load error information

ihavecoke:
This PR fix an issue where, if an error occurs while loading the keymap file during application startup, an excessively long error message would be truncated and not fully displayed.

Before:

before

After:

image

Release Notes:

- N/A

### DIFF
--- a/crates/workspace/src/notifications.rs
+++ b/crates/workspace/src/notifications.rs
@@ -593,9 +593,9 @@ pub mod simple_message_notification {
 
     use gpui::{
         AnyElement, DismissEvent, EventEmitter, FocusHandle, Focusable, ParentElement, Render,
-        SharedString, Styled,
+        ScrollHandle, SharedString, Styled,
     };
-    use ui::prelude::*;
+    use ui::{ScrollAxes, WithScrollbar, prelude::*};
 
     use crate::notifications::NotificationFrame;
 
@@ -617,6 +617,7 @@ pub mod simple_message_notification {
         show_close_button: bool,
         show_suppress_button: bool,
         title: Option<SharedString>,
+        scroll_handle: ScrollHandle,
     }
 
     impl Focusable for MessageNotification {
@@ -661,6 +662,7 @@ pub mod simple_message_notification {
                 show_suppress_button: true,
                 title: None,
                 focus_handle: cx.focus_handle(),
+                scroll_handle: ScrollHandle::new(),
             }
         }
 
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Nice, thank you for looking into this! This does look good, three minor notes:

- Can we shrink this a bit more to not occupy 80% of the viewport at max, but e.g. something between 40-60%?
- Can we perhaps push this fix down quite a bit? I think there should be no notification that extends the viewport, so I think we should just add this directly to the MessageNotification perhaps.
- Could we also add a scrollbar for this? Optimally one that follows the `scrollbar.show` setting.

Feel free to ping me if there is something I can help with. Thanks!

ihavecoke:
@MrSubidubi Thanks for the feedback, Here are some notes on the changes:

- I experimented with reducing the `vh` value to around `0.6`, but after searching through the project I noticed that `LanguageServerPrompt` also uses `0.8` as its `max_h` ratio. Keeping it consistent felt more appropriate, and in my testing `0.8vh` still works well across most scenarios.

- Good suggestion — I’ve updated the notification content to use `overflow-y: scroll`.

- Regarding the scrollbar: I understand the idea of aligning its behavior with the `scrollbar.show` setting. However, for notifications specifically:

   - The notification UI has a fixed and relatively narrow width, so adding a vertical scrollbar doesn’t provide much practical value — we already want to use as much horizontal space as possible to show the message itself.
   - Other components that rely on `scrollbar.show` (such as the Editor, Git Panel, and Outline) either manage their own dedicated setting or have a more integrated use case for it. Introducing a new setting just to control scrollbar behavior in notifications doesn’t seem necessary.
   - I also experimented with using the Editor’s `scrollbar.show` setting, but due to dependency constraints and integration issues, it didn’t fit well.

   - Given these considerations, I haven't add a scrollbar to the message notification.

Happy to adjust further if needed — thanks again!

MrSubidubi:
Alright, thanks for the input!

> I experimented with reducing the vh value to around 0.6, but after searching through the project I noticed that LanguageServerPrompt also uses 0.8 as its max_h ratio. Keeping it consistent felt more appropriate, and in my testing 0.8vh still works well across most scenarios.

Thank you for checking. I do think still, though, that for an error prompt, this is too large, yet I fully agree with keeping it consistent. How would you feel about just shrinking the LSP one to 0.6vh as well? 

Also, thanks for checking in about the setting. Regarding

> The notification UI has a fixed and relatively narrow width, so adding a vertical scrollbar doesn’t provide much practical value — we already want to use as much horizontal space as possible to show the message itself.

By default, the scrollbar floats atop the content, so it would not reduce the content width at all.

As for the settings part, that is a good call, we can just leave that out.

Could you look into playing around with adding a scrollbar? 

And thanks for being that responsive!

ihavecoke:
@MrSubidubi Sure, reducing it to 0.6vh works for me — I’ve updated the value accordingly.

I’ve also added support for displaying a vertical scrollbar as suggested.

The latest changes have been pushed. Please let me know if there’s anything else that needs adjustment. Thanks again for the quick feedback!

MrSubidubi (approved):
Nope, looks good. Thanks for the good fix and being that responsive!

PR #42545, opened 2025-11-12
Install emmet extension by default

### LINKED ISSUE #32713: Show Emmet as a suggested extension for web projects
### Summary

Show the Emmet extension as a suggested extension for new users with web prijects

### Description
During user interviews, some users didn't realize we had emmet snippet support because it's not enabled by default like in VS Code & Jetbrains. We should surface that we have Emmet support proactively so they can install it if they would like to.

We can use our recommended extensions modal:

Image

⚠️ Design consideration here as we already show other (?) suggested extensions for web projects – how do we surface multiple?

KyleBarton:
Closes #32713

Release Notes:

- Added the Emmet extension to auto-loaded extensions list

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1622,7 +1622,8 @@
   // If you don't want any of these extensions, add this field to your settings
   // and change the value to `false`.
   "auto_install_extensions": {
-    "html": true
+    "html": true,
+    "emmet": true
   },
   // The capabilities granted to extensions.
   //


maxdeviant (changes requested):
Hmm, I don't think we should do this.

We already have snippet support built-in (correct me if I'm wrong about that, @osiewicz).

But Emmet is a very specific tool that not everyone is going to want out-of-the-box.

KyleBarton:
@maxdeviant got it - #32713 suggested the goal was feature parity with VS Code, which does auto-install - but do you think it would be better as a suggested extension?

If so, we might need to tweak suggested extensions a bit, since the `HashMap` datatype suggests that there should be only one suggested extension per file type (and this would overlap with the `html` extension)

KyleBarton:
FWIW, I don't see snippets working out of the box right now, but I could be incorrectly testing - I'm not the most familiar with Emmet-style html editing.

franciskafyi:
@KyleBarton @maxdeviant as per [our docs](https://zed.dev/docs/languages/html#html), shouldn't we auto install the HTML extension?

maxdeviant:
We had some discussion in Slack, and we don't want to auto-install the Emmet extension.

Emmet enables a specific workflow that not everyone wants, and the fact that the list of languages that the Emmet language server will start for is quite extensive means that it will be invasive to users who may not want it:

https://github.com/zed-extensions/emmet/blob/cc59eaace227dff113655524327d9bd16d1e65bc/extension.toml#L12

We agreed that we could reuse/extend the extension suggestion system to suggest installing the Emmet extension in relevant files.

PR #42548, opened 2025-11-12
docs: Update developing extension docs for updating specific submodule

ArunChapagain:
Release Notes:

- N/A

### DIFF
--- a/docs/src/extensions/developing-extensions.md
+++ b/docs/src/extensions/developing-extensions.md
@@ -106,7 +106,11 @@ zed::register_extension!(MyExtension);
 # git clone https://github.com/zed-industries/extensions
 cd extensions
 git submodule init
+# Running git submodule update without a path will update all extensions, which may take some time.
+# To update all submodules:
 git submodule update
+# To update only a specific extension submodule for faster operation:
+# git submodule update extensions/specific-extension-name
 ```
 
 ## Extension License Requirements
--- a/docs/src/extensions/developing-extensions.md
+++ b/docs/src/extensions/developing-extensions.md
@@ -106,12 +106,20 @@ zed::register_extension!(MyExtension);
 # git clone https://github.com/zed-industries/extensions
 cd extensions
 git submodule init
-# Running git submodule update without a path will update all extensions, which may take some time.
-# To update all submodules:
 git submodule update
-# To update only a specific extension submodule for faster operation:
-# git submodule update extensions/specific-extension-name
 ```
+
+## Update Your Extension
+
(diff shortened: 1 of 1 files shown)


ArunChapagain:
This PR adds following requirement mentioned in  this issue:
https://github.com/zed-industries/zed/discussions/42547

ArunChapagain:
Release Notes:

- N/A

MrSubidubi (changes requested):
Thanks, I think this definitely makes sense to mention. Few questions: 

1. Do we want to just add a new section "Update your extension" to the docs which describes this process in more detail? Think for that case it would make the most sense, would it not?
2. Is the `git submodule update` at the start needed at all? If we did a fresh clone, might not even need to do it at all?

ArunChapagain:
1:
Yes, a new section titled "Update your Extension" would be the clearest and most helpful approach. It will separate the initial setup (cloning and initialization) from the ongoing maintenance (updating).and  provides a dedicated place to explain the difference between the full git submodule update and the faster, specific-path update (e.g., git submodule update extensions/specific-extension-name).

2:
You are correct; the initial git submodule update is not strictly necessary if the user just performed a fresh git clone followed by git submodule init.

While the full update is often included as a safe default to satisfy potential dependencies in the build system, you are right that for a developer focusing on creating a single new extension, fetching all content is inefficient.

We should advise developers to use the faster, specific-path method even for the initial content fetch: git submodule update extensions/specific-extension-name. This ensures the setup is minimal and focused on their development needs.

ArunChapagain:
@MrSubidubi Should I update the PR including "**Update Your Extension Section**"?

MrSubidubi:
Yes, please.

ArunChapagain:
@MrSubidubi I have updated docs accordingly. Review the changes

docs/src/extensions/developing-extensions.md
```diff
@@ -106,13 +106,21 @@ zed::register_extension!(MyExtension);
...
 ```
 
+## Update Your Extension
+
+When developing/updating your extension, you will likely need to update its content from its submodule in the extensions repository.
+To quickly fetch the latest code for only specific extension (and avoid updating all others), use the specific path:
+
+```sh
+# From the root of the repository:
+git submodule update extensions/your-extension-name
```
MrSubidubi: Is this not missing the `--remote` flag for updating?

ArunChapagain:
@MrSubidubi Oh yes. I have just updated the flag

MrSubidubi (approved):
Nice improvement, thank you!

PR #42558, opened 2025-11-12
Fix not able to navigate to files in git commit multibuffer

### LINKED ISSUE #40851: Can't navigate to files in git commit multibuffer
### Summary

Typical file navigation doesn't work if I open a multibuffer of a commit.

### Description
Click on the menu or equivalent keybinding won't navigate to file, it's still possible to get to the file by right clicking and revealing in project panel.

Image

Steps to reproduce:
1. Go to git panel
2. Click on previous commit
3. Have the multibuffer open
4. Try navigating to a file

**Expected Behavior**: Being able to navigate to a file just like in any multibuffer
**Actual Behavior**: Not being able to...

Zed: v0.209.3 (Zed Preview) 
Memory: 16 GiB
Architecture: aarch64

xipeng-jin:
Closes #40851

Release Notes:

- Fixed: Commit diff multibuffers now open real project files whenever possible, restoring navigation and annotations inside those excerpts.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -21642,6 +21642,11 @@ impl Editor {
                 };
 
                 for (buffer, (ranges, scroll_offset)) in new_selections_by_buffer {
+                    let is_project_file = buffer
+                        .read(cx)
+                        .file()
+                        .map(|file| project::File::from_dyn(Some(file)).is_some())
+                        .unwrap_or(false);
                     let editor = buffer
                         .read(cx)
                         .file()
@@ -21679,6 +21684,9 @@ impl Editor {
                         });
 
                     editor.update(cx, |editor, cx| {
+                        if !is_project_file {
+                            editor.set_read_only(true);
+                        }
                         let autoscroll = match scroll_offset {
                             Some(scroll_offset) => Autoscroll::top_relative(scroll_offset as usize),
                             None => Autoscroll::newest(),
@@ -21702,7 +21710,7 @@ impl Editor {
     // For now, don't allow opening excerpts in buffers that aren't backed by
     // regular project files.
     fn can_open_excerpts_in_file(file: Option<&Arc<dyn language::File>>) -> bool {
-        file.is_none_or(|file| project::File::from_dyn(Some(file)).is_some())
+        file.is_some()
     }
 
     fn marked_text_ranges(&self, cx: &App) -> Option<Vec<Range<OffsetUtf16>>> {
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -21709,9 +21709,7 @@ impl Editor {
     // or represent synthetic/non-local files (e.g., git blobs). File-less buffers
     // are also supported so tests and other in-memory views keep working.
     fn can_open_excerpts_in_file(file: Option<&Arc<dyn language::File>>) -> bool {
-        file.is_none_or(|file| {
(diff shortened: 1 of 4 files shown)


xipeng-jin:
There is something I noticed can be optimized, so I am still working on it. I am also trying to validate that behaving correctly in all cases. Thanks

cole-miller:
Thanks! I don't think we want to use "real" files in the commit view multibuffer, though. We need to show the content of each path as it existed at the time of the commit, not the current content.

I think the right way to approach this is to make `editor::OpenExcerpts` and `editor::OpenExcerptsSplit` work with the `GitBlob` files that the commit view uses. Right now this is explicitly disabled by `Editor::can_open_excerpts_in_file`, so we should lift that restriction and then make sure that the resulting editor is opened read-only. We should also make sure it's visually clear that the opened editor is not a project file, by customizing the tab content (maybe this could mention the SHA).

cole-miller:
Feel free to book a time here if you'd like to pair on it: https://cal.com/esther-trapadoux-zed/30min

xipeng-jin:
Thanks @cole-miller , I have booked a time on next Tuesday and happy to pair with you. I took a look at your last comment further yesterday and agree it was not a good idea to expose the "real" files for commit view. `Editor::can_open_excerpts_in_file` is indeed disable that, so I am starting from that point. But I might need more thoughts on how we should customize the tab content to make it visually clear it is not from a project file later today.

alphathekiwi:
Hey @xipeng-jin I see we have an interest in a lot of the same issues, I have also been working on tackling this. And had a Pair session last week to discuss my progress. (Only just published now)
https://github.com/zed-industries/zed/pull/42760

Basically there are 3 possibility 
The user wants to navigate to the Parent file (before changes were applied)
The user wants to navigate to the Modified file
The user wants to open the file at HEAD (maybe they were just looking at historical code and now want to modify the file)

We could try to assume the users intent with a smart OpenExcept action that would try to open the Parent file if the cursor is inside a removed hunk and the Modified file if the cursor is inside an added hunk.

But that doesn't consider what action to take when the cursor is in a section of unmodified text. Nor does it allow the user to open the file in the HEAD 

The route I have ended up taking is to make a bunch of settings and some fallbacks, (what to do when the user's first choice action is not possible in the given context).

I would love to attend your pairing session as a spectator or potentially contribute what I've done to your efforts. I am eager to see this feature get added in some fashion as it has been a big pain point for me.

xipeng-jin:
Hi @alphathekiwi , I took a quick glance of your PR and really appreciate your effort and interest here. I am happy to collaborate with you in this one if you like. Something I think we might need to narrative the size of this PR, and we also might want to consider to separate out the large PR into several small ones to make reviewing easier. Yes, I did book a time with the team to have a concrete idea with the UI changes. I will send out the invite for you if you have time. 

Appreciate for your feedback and contribution here and happy to collaborate with you in this one to shipping the best shape of the feature. Thanks

esthertrapadoux:
@xipeng-jin I think you're already chatting with @Anthony-Eid but we have more pairing slots with the team in case you want to grab one! https://cal.com/esther-trapadoux-zed/30min?overlayCalendar=true

Let me know if you don't see any open slots during your time zone. I can open up more.

xipeng-jin:
Hi @esthertrapadoux , thanks for reaching out, but I cannot see any open slots in this week. The earliest spot is next Monday. I am happy to grab some pairing slots in this week if the team is available. Thank you.

PR #42576, opened 2025-11-12
Add windows nightly update banner

P1n3appl3:
Hopefully this will nudge some of the beta users who were on nightly to get on the official stable builds now that they're out.

Release Notes:

- N/A

### DIFF
--- a/crates/title_bar/src/title_bar.rs
+++ b/crates/title_bar/src/title_bar.rs
@@ -31,7 +31,6 @@ use gpui::{
 };
 use onboarding_banner::OnboardingBanner;
 use project::{Project, WorktreeSettings, git_store::GitStoreEvent};
-use release_channel::ReleaseChannel;
 use remote::RemoteConnectionOptions;
 use settings::{Settings, SettingsLocation};
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


crates/title_bar/src/title_bar.rs
```diff
@@ -291,18 +292,38 @@ impl TitleBar {
...
-            .visible_when(|cx| !project::DisableAiSettings::get_global(cx).disable_ai)
-        });
+        let on_windows_nightly = cfg!(target_os = "windows")
+            && ReleaseChannel::try_global(cx)
+                .map(|channel| channel == ReleaseChannel::Nightly)
+                .unwrap_or(false);
+        let banner = if on_windows_nightly {
+            cx.new(|cx| {
+                OnboardingBanner::new(
+                    "Stable windows releases",
```
maxdeviant: This is the value that ends up in our telemetry data, so let's use Title Case:

```suggestion
                    "Stable Windows Releases",
```

P1n3appl3:
@franciskafyi mentioned that this is too subtle, so I'll try to switch it to a notification/toast. I'll probably use the same onboarding banner logic for tracking dismissal.

PR #42580, opened 2025-11-12
zeta2: Improve zeta old text matching

probably-neb:
This PR improves Zeta2's matching of `old_text`/`new_text` pairs, using similar code to what we use in the edit agent. For right now, we've duplicated the code, as opposed to trying to generalize it.

Release Notes:

- N/A


crates/zeta_cli/src/evaluate.rs
```diff
@@ -196,7 +185,7 @@ fn write_eval_result(
     example: &NamedExample,
     predictions: &PredictionDetails,
     evaluation_result: &EvaluationResult,
-    out: &mut impl Write,
+    out: &mut StdOutOrFile,
```
maxbrunsfeld: @probably-neb  I think these functions should just take an `impl Write` and an `use_color: bool` parameter. It seems a lot simpler than having them write out ansi codes, and having to parse and strip the ansi codes.

PR #42588, opened 2025-11-13
Add an action listener to workspace for ActivatePreviousItem and ActivateNextItem

FloppyDisco:
Release Notes:

- pane::ActivatePreviousItem and pane::ActivateNextItem now toggle the most recent pane when called from a dock panel

a couple months ago i posted a work around that used `SendKeystrokes` to
cycle through pane items when focused on a dock.
#35253

this pr would add this functionality to the these actions by default.
i implemented this by adding an action listener to the workspace level.

------
if the current context is a dock that does not hold a pane
it retrieves the most recent pane from `activation_history` and activates the next item on that pane instead.

- `"Pane > Editor"`
cycles through the current pane like normal
- `"Dock > Pane > Terminal"`
also cycles through the pane items like normal
- `"Dock > (Any Child that is not a child of Pane)"`
cycles through the items of the most recent pane.

this is the standard behavior in VS Code i believe.

in the video below you can see the actions cycling through the editor like normal when focus is on the editor.
then you can see the editor continue to cycle when the focus is on the project panel.
and that the focus stays on the project panel.
and you can see the action cycle the terminal items when the focus is moved to the terminal

https://github.com/user-attachments/assets/999ab740-d2fa-4d00-9e53-f7605217e6ac

the only thing i noticed is that for this to work the keybindings must be set above `Pane`
so they have to be set globally or on workspace. otherwise they do not match in the context

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -5950,6 +5950,88 @@ impl Workspace {
                 },
             ))
             .on_action(cx.listener(Workspace::toggle_centered_layout))
+            .on_action(cx.listener(
+                |workspace: &mut Workspace, _action: &pane::ActivateNextItem, window, cx| {
+                    if let Some(active_dock) = workspace.active_dock(window, cx) {
+                        let dock = active_dock.read(cx);
+                        if let Some(active_panel) = dock.active_panel() {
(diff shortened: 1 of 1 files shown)


ConradIrwin:
Thanks!

PR #42613, opened 2025-11-13
editor: Improve multi-buffer header filename click to jump to the latest selection from that buffer - take 2

smitbarmase:
Relands https://github.com/zed-industries/zed/pull/42480

Release Notes:

- Clicking the multi-buffer header file name or the "Open file" button now jumps to the most recent selection in that buffer, if one exists.

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -7802,21 +7802,46 @@ fn header_jump_data(
     snapshot: &EditorSnapshot,
     block_row_start: DisplayRow,
     height: u32,
-    for_excerpt: &ExcerptInfo,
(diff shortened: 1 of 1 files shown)


VladimirTaytor:
Hey @smitbarmase is it final UX for search results buffer header?

Previously simple mouse click anywhere on header would `Open file` in a new tab
But now I find it pretty inconvenient that to open file I need to aim to file name (_which always has different size_) or to the `Open file` button which is not even showing until user focuses on particular file search results

There is workaround with a `"double_click_in_multibuffer": "open"`, but still my old muscle memory always clicking on header itself and file is collapsed instead of opening in new tab 😄 

Thanks for your work, Zed is the best 💪

PR #42615, opened 2025-11-13
vim: Add `use_match_quotes` setting for % motion, default is `true`

Kalmaegi:
Add a new Vim setting `use_match_quotes` that controls whether the % motion should treat quotes ``(`', ", `)``
 as matching pairs.
In Vim, % only matches pairs of `([{}])` — doesn’t consider quotes, to avoid breaking existing behavior in Zed, this new setting defaults to true, but users who prefer Vim’s original behavior can now disable it by setting:
`"use_match_quotes": false`

While implementing this, I noticed there’s an existing helper:
https://github.com/zed-industries/zed/blob/b01a6fbdeaeb44698aa03ad04a48cab307b5ccf8/crates/vim/src/object.rs#L284-L289
I'm wondering if it might be cleaner to expose a small wrapper or helper here, so we can reuse the same quote detection logic instead of duplicating it.

If this overall approach looks good, I’ll proceed to add unit tests.

Release Notes:

- vim: Added `match_quotes` parameter to the `vim::Matching` action to control whether `%` matches quote characters


dinocosta:
Thank you for opening this Pull Request @Kalmaegi ! Will think a little bit more about this and try to take a look in the coming days, will keep you posted 🙂

dinocosta:
Hey @Kalmaegi 👋 

Thank you for opening this and sorry that it took me a while to get back to you, was checking with the team whether we'd like to introduce this, but Conrad told me he already chatted with you on Discord about it. Approach looks good to me, will see if we can simply the `make_range_filter` closure a little bit, seems to be getting a bit complicated 🤔

In the meantime, would you be able to:

- Rebase this with `main`
    Seems there's conflicts in `crates/vim/src/motion.rs`
- Write a test to confirm that `Matching { match_quotes: false }` behaves differently from `Matching { match_quotes: true}`
    Let me know if you need help with this one, I can also try adding it myself
        
Thanks!

P.S.: Regarding the helper function you shared, `vim::object::is_bracket_delimiter`, it seems there's also `vim::object::is_quote_delimiter` already, so we can probably leverage one of, if not both, functions 🙂

Kalmaegi:
> Hey @Kalmaegi 👋
> 
> Thank you for opening this and sorry that it took me a while to get back to you, was checking with the team whether we'd like to introduce this, but Conrad told me he already chatted with you on Discord about it. Approach looks good to me, will see if we can simply the `make_range_filter` closure a little bit, seems to be getting a bit complicated 🤔
> 
> In the meantime, would you be able to:
> 
> * Rebase this with `main`
>   Seems there's conflicts in `crates/vim/src/motion.rs`
> * Write a test to confirm that `Matching { match_quotes: false }` behaves differently from `Matching { match_quotes: true}`
>   Let me know if you need help with this one, I can also try adding it myself
> 
> Thanks!
> 
> P.S.: Regarding the helper function you shared, `vim::object::is_bracket_delimiter`, it seems there's also `vim::object::is_quote_delimiter` already, so we can probably leverage one of, if not both, functions 🙂

Sure! Sorry for the lack of updates lately. I’ll fetch on main, resolve the conflicts, and add the test for the `match_quotes` behavior. I’ll also clean up the code a bit while I’m at it.

Kalmaegi:
Hi @dinocosta
I’ve added some test cases and I also tested the tag support, and luckily everything still works fine — nothing got affected.
Regarding the new setting: `use_match_quotes` initially, I wanted to handle it directly inside the `matching` method, since that would make the logic a bit cleaner. But I realized that `matching` doesn’t have access to the necessary context for the `VimSettings`, so I had to move that logic up one level instead.
Hopefully this change is alright — fingers crossed XD

dinocosta:
Thank you @Kalmaegi ! I've pushed two more commits, one to just bring this branch up to date with `main` and another one with a very tiny refactor and adding documentation so that we get documentation on this setting when the user is updating their settings file :) 

> Hi @dinocosta I’ve added some test cases and I also tested the tag support, and luckily everything still works fine — nothing got affected. Regarding the new setting: `use_match_quotes` initially, I wanted to handle it directly inside the `matching` method, since that would make the logic a bit cleaner. But I realized that `matching` doesn’t have access to the necessary context for the `VimSettings`, so I had to move that logic up one level instead. Hopefully this change is alright — fingers crossed XD

Honestly, I believe it's possible to just have it in the `Matching` motion, you already added the `match_quotes` field there, I believe the only thing we'd need to do is:

1. Update the `Matching` action with the same `match_quotes` field
1. Update the default vim keymap (`assets/keymaps/vim.json`) to use `["vim::Matching", { "match_quotes": true }]`
1. Possibly update [Zed's Vim Mode documentation](https://zed.dev/docs/vim) to mention that `%` can behave like `Vim` by simply updating the keymap, removing the `match_quotes` parameter

Let me know what you think! No rush on this one, so we can also explore that option.

Kalmaegi:
> Thank you @Kalmaegi ! I've pushed two more commits, one to just bring this branch up to date with `main` and another one with a very tiny refactor and adding documentation so that we get documentation on this setting when the user is updating their settings file :)
> 
> > Hi @dinocosta I’ve added some test cases and I also tested the tag support, and luckily everything still works fine — nothing got affected. Regarding the new setting: `use_match_quotes` initially, I wanted to handle it directly inside the `matching` method, since that would make the logic a bit cleaner. But I realized that `matching` doesn’t have access to the necessary context for the `VimSettings`, so I had to move that logic up one level instead. Hopefully this change is alright — fingers crossed XD
> 
> Honestly, I believe it's possible to just have it in the `Matching` motion, you already added the `match_quotes` field there, I believe the only thing we'd need to do is:
> 
> 1. Update the `Matching` action with the same `match_quotes` field
> 2. Update the default vim keymap (`assets/keymaps/vim.json`) to use `["vim::Matching", { "match_quotes": true }]`
> 3. Possibly update [Zed's Vim Mode documentation](https://zed.dev/docs/vim) to mention that `%` can behave like `Vim` by simply updating the keymap, removing the `match_quotes` parameter
> 
> Let me know what you think! No rush on this one, so we can also explore that option.

Sorry i'm late, I didn't think of that approach! That makes a lot more sense.

Just to confirm my understanding: with this approach, users only need to override the default
parameter in their keymap to achieve the same effect, right? They wouldn't need a separate
global setting at all.

I really like this solution! If my understanding is correct, I'll refactor the code
accordingly. Thanks for the great suggestion!

Kalmaegi:
I’ve tweaked the code and tested it, hope everything works as expected.

dinocosta (approved):
@Kalmaegi Thank you for taking the time to update these changes and including these tests, appreciate it! 🙇 

I've pushed a commit adding some documentation to the `match_quotes` field, as well as also including `"match_quotes": true` in the Helix keymap, so that it continues working as Zed's defaults after this is merged. Thanks! ✌️

Kalmaegi:
> @Kalmaegi Thank you for taking the time to update these changes and including these tests, appreciate it! 🙇
> 
> I've pushed a commit adding some documentation to the `match_quotes` field, as well as also including `"match_quotes": true` in the Helix keymap, so that it continues working as Zed's defaults after this is merged. Thanks! ✌️

thanks, I forgot Helix mode XD

PR #42627, opened 2025-11-13
multi_buffer: Don't show indent guides for multi-line strings

### LINKED ISSUE #40635: Indent guides shouldn't be displayed for multi-line strings
### Summary

Indent guides shouldn't be displayed in a multi-line string

### Description

code example:
```python
_ = """this is an example of an
    indented
multiline string"""
``` 

**Expected Behavior**:
no indent guides
**Actual Behavior**:

Image

Zed: v0.208.5 (Zed) 
OS: Linux Wayland fedora 42
Memory: 13.5 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 25.1.9

amustaque97:
**Summary**

Python supports multi-line strings, aka docstrings. When config is set up something like:
```json
{
  "indent_guides": {
    "enabled": true,
    "coloring": "indent_aware",
    "background_coloring": "indent_aware"
  }
}
```
The editor should skip the indentation colouring for the multi-line strings

Closes #40635

Release Notes:

### DIFF
--- a/crates/multi_buffer/src/multi_buffer.rs
+++ b/crates/multi_buffer/src/multi_buffer.rs
@@ -5186,10 +5186,11 @@ impl MultiBufferSnapshot {
             if first_row > end_row {
                 break;
             }
-            
+
             // Check if the line is inside a string literal by examining the syntax tree
             // If inside a string, treat it as having zero indent to close any active indent guides
-            let is_in_string = self.point_to_buffer_offset(Point::new(first_row.0, 0))
+            let is_in_string = self
+                .point_to_buffer_offset(Point::new(first_row.0, 0))
                 .and_then(|(buffer_snapshot, buffer_offset)| {
                     buffer_snapshot
                         .language_scope_at(buffer_offset)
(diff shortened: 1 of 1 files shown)


bennetbo:
Hey, thanks for working on this! My two cents:

Currently, the indent guides are purely based on whitespace and do not take syntax information into account. Personally, I prefer this to having something that is syntax-aware, as I am purely interested in whitespace changes for indent guides. I believe the current behaviour matches that of VSCode. However, there are other IDEs, such as IntelliJ, that have more syntax-aware indent guides.

Even if we wanted this feature, we would need to figure out how to make it more efficient. Calling `language_scope_at` on every visible line in every frame will cause a noticeable slowdown (I had a similar performance problem when i implemented indent guides a year ago).

Going to close this PR, as I don't have the bandwidth to investigate performance further, and I personally prefer to keep everything purely based on whitespace instead of introducing special cases for syntax-aware colouring.
Happy to discuss this further in the issue if people feel differently.

PR #42633, opened 2025-11-13
agent_ui: Remove context strip from inline assistant

bennetbo:
TODO
- [x] Implement PromptEditor::paste
- [x] Fix creases on unlink
- [x] PromptCompletionProviderDelegate::supports_images
- [ ] Fix highlighting in completion menu

Release Notes:

- N/A


bennetbo:
Will fix the highlighting in completion menu in a follow up PR, since that is already broken on main.

PR #42382, opened 2025-11-10
Use shell to launch MCP and ACP servers

reflectronic:
`npx`, and any `npm install`-ed programs, exist as batch scripts/PowerShell scripts on the PATH. We have to use a shell to launch these programs.

Fixes https://github.com/zed-industries/zed/issues/41435
Closes https://github.com/zed-industries/zed/pull/42651

Release Notes:

- windows: Custom MCP and ACP servers installed through `npm` now launch correctly.

PR #42651, opened 2025-11-13
util: Escape shell arguments that need it for `terminal/create` event

stippi:
Closes #42646

Release Notes:

- Agent: Fixed ACP terminal command line forming to escape args with special characters

### DIFF
--- a/crates/util/src/shell_builder.rs
+++ b/crates/util/src/shell_builder.rs
@@ -1,5 +1,6 @@
 use crate::shell::get_system_shell;
 use crate::shell::{Shell, ShellKind};
+use std::borrow::Cow;
 
 /// ShellBuilder is used to turn a user-requested task into a
 /// program that can be executed by the shell.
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks! And sorry this took so long to review, we were working on related changes in https://github.com/zed-industries/zed/pull/42382 which I figured we'd land before but it got blocked on some other stuff for the time being.

Veykril (commented):
Seems like some tests need to be updated still

stippi:
Yes, sorry I didn't have time to look into it, yet. That's why I set it to Draft for the time being.

stippi:
The tests might reveal there is an actual problem and not only tests would need updating, but some other part of the code that relies on the old behavior.

Veykril:
Yes that is very likely, I think we observed similar things in https://github.com/zed-industries/zed/pull/42382. No worries if you don't have the time to look into this, we will hopefully address this in the linked PR as well once we finish that up then.

PR #42658, opened 2025-11-13
Fix performance issues with 1Password local .env mounts (FIFO files / named pipes) in workspaces

sidharth-sudhir:
Release Notes:

- Fixed FIFO files (named pipes) causing continuous re-reads and performance issues with 1Password .env mounts on macOS and Linux

## Problem
Zed previously triggered multiple concurrent reads when FIFO files (named pipes) were present in a workspace. FSEvents (macOS) and inotify/notify (Linux) reported constant "modified" events for FIFOs, causing Zed's file watcher to repeatedly attempt to reload these files. This was particularly problematic with 1Password .env mounts which use FIFO files.

## Solution
This PR filters out "Changed" events for FIFOs on both macOS and Linux while still allowing "Created" and "Removed" events so they correctly appear/disappear in the file explorer.

Changes:
- Made `is_fifo()` function public so it can be used across watcher implementations
- Updated macOS watcher to filter FIFO modify events
- Updated Linux watcher to filter FIFO modify events


franciskafyi:
Hey @sidharth-sudhir, let us know when this PR is ready for review. We're doing a big PR review push next week ahead of the break and we'd love to include this one, if it's ready!

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR addresses a performance issue where FIFO files (named pipes) cause continuous re-reads in Zed workspaces, particularly affecting 1Password .env mounts on macOS and Linux. The solution filters out `Changed` events for FIFO files while preserving `Created` and `Removed` events.

## Architecture

The changes are well-structured:
- Added a new `is_fifo()` helper function in `fs.rs` that uses `std::fs::symlink_metadata` to detect FIFO files
- Updated the macOS file watcher to filter FIFO modify events (lines 17-21 in fs.rs)
- Updated the Linux file watcher to filter FIFO modify events with platform-specific cfg guards (lines 85-98 in fs_watcher.rs)

The approach correctly maintains FIFO file visibility in the file explorer (Created/Removed events still fire) while preventing the problematic Changed events.

## Issues

### 1. Error Handling in `is_fifo()` (Minor)

```rust
pub fn is_fifo(path: &Path) -> bool {
    std::fs::symlink_metadata(path)
        .ok()
        .map(|m| m.file_type().is_fifo())
        .unwrap_or(false)
}
```

The function silently returns `false` on any error (line 42). While this is defensive and won't break the functionality, it could mask legitimate filesystem errors. Consider at least logging the error in debug mode:

```rust
pub fn is_fifo(path: &Path) -> bool {
    match std::fs::symlink_metadata(path) {
        Ok(m) => m.file_type().is_fifo(),
        Err(err) => {
            log::debug!("Failed to check FIFO status for {:?}: {}", path, err);
            false
        }
    }
}
```

### 2. symlink_metadata Usage (Consideration)

The function uses `symlink_metadata` which follows symlinks. This means a symlink pointing to a FIFO would be detected as a FIFO. This is likely the intended behavior (to handle the actual file being watched), but it's worth noting that the symlink itself is not technically a FIFO. If the goal is to filter out symlinks to FIFOs as well, this is correct.

### 3. Missing Tests (Moderate)

No tests were added for the FIFO filtering behavior. This is a significant gap given:
- This is a platform-specific behavior (Unix only)
- FIFO files are special filesystem objects
- The fix touches two different watcher implementations

Consider adding integration tests that:
- Create a FIFO file using `mkfifo`
- Verify that modify events on FIFOs are filtered
- Verify that create/remove events on FIFOs still propagate

### 4. Existing `is_fifo` Field in Metadata

The codebase already has a `Metadata` struct with an `is_fifo` field (line 294 in fs.rs). While the new function is needed for path-based checking (before loading metadata), it's worth documenting the relationship between these two mechanisms for future maintainers.

### 5. Potential Edge Case on macOS

In fs.rs lines 22-25, there's special handling for `ITEM_XATTR_MOD` events on FIFOs:

```rust
} else if event.flags.contains(StreamFlags::ITEM_XATTR_MOD)
    && is_fifo(&event.path)
{
    Some(PathEventKind::Changed)
}
```

This intentionally allows extended attribute modification events on FIFOs to propagate. However, this contradicts the PR's goal of filtering FIFO change events. Please verify this is intentional - if FIFOs spam XATTR modify events, this could reintroduce performance issues.

## Cross-file Impact

- **fs.rs**: Added `is_fifo()` function and modified macOS event filtering
- **fs_watcher.rs**: Modified Linux event filtering, imported `is_fifo`
- No breaking changes to public APIs
- No changes to Windows behavior (correctly guarded with `#[cfg(not(unix))]`)

## Assessment

**✅ APPROVE with minor suggestions**

The PR effectively solves the documented performance problem with a minimal, targeted approach. The architecture is sound and the cross-platform handling is appropriate.

The main concern is the lack of tests for this platform-specific, edge-case behavior. Given that this is a bug fix with a clear problem statement and reasonable solution, I recommend merging after addressing:

1. Add error logging in `is_fifo()` (optional but recommended)
2. Document/verify the XATTR_MOD handling for FIFOs on macOS
3. Consider adding tests in a follow-up PR

The code is safe to merge as-is, but the above improvements would make it more robust.

GitGlimpse895:
Hey @cole-miller — following up on @sidharth-sudhir's ping from January. This one has been sitting assigned to you for nearly 4 months with no response.

I checked current `main` — there's still zero FIFO filtering in `fs_watcher.rs`. `EventKind::Modify` maps straight to `PathEventKind::Changed` with no check, so anyone with 1Password `.env` mounts or any named pipes in their workspace is still hitting this today.

This is a small, focused fix (48 lines) with a clear problem and solution. Is there anything blocking a review, or could this be reassigned to someone with bandwidth?

zelenenka:
Sorry for the long silence here.

We've finally taken a look and unfortunately the fix doesn't seem to be quite right. The main concerns are:
- `is_fifo()` puts a `symlink_metadata` on the watcher hot path for every `Modify` event
- the macOS `ITEM_XATTR_MOD` branch seems to put the meant-to-be-dropped events back in
- renames are suppressed (`ITEM_RENAMED` on macOS, `ModifyKind::Name` under `EventKind::Modify(_)` on Linux).

The PR has also accumulated conflicts due to the delay on our side, and given that as well as the concerns listed above, I'll close it now as I believe at this point it would be less effort to create a new fix than to iterate on this PR. Thank you for identifying the problem and for taking a stab at fixing it, sorry we didn't get around the review in a timely manner.

PR #34798, opened 2025-07-21
Add which-key system

Cretezy:
Closes https://github.com/zed-industries/zed/issues/10910

This is a work in progress implementation, but it functionally works. I'm looking for some feedback on both implementation and design.

This PR adds a `which-key`-like system, which shows possible bindings to take after starting a sequence of key which leads to a binding. This is mostly useful in Vim mode, as there's multiple common sequences (such as `g` which leads to `gv`, `gc`, ..., or `[`/`]`, etc). Currently, this shows in the main screen area (doesn't draw over panels), to be less intrusive.

Common use cases are to see what can follow `ctrl+k`, `g`, `z`, `ctrl+w`, or `space`/leader. `g`/`ctrl+w` are worst case scenarios, as they have tons of features built-in.

image

`"location": "left_panel"`:
image

To do:
- [ ] Add ability to scroll menu with keyboard

Additionally, there's some follow-up work which could enable this menu when holding down `ctrl` for example. This would make this feature much more useful in non-Vim mode.

Release Notes:

- Add which-key system to show possible next actions for sequences of bindings (e.g. `ctrl-w`)

PR #42659, opened 2025-11-13
gpui: Only time out multi-stroke bindings when current prefix matches

xipeng-jin:
Part One for Resolving #10910

### Summary
Typing prefix (partial keybinding) will behave like Vim. No timeout until you either finish the sequence or hit Escape, while ambiguous sequences still auto-resolve after 1s.

### Description
This follow-up tweaks the which-key system first part groundwork so our timeout behavior matches Vim’s expectations. Then we can implement the UI part in the next step (reference latest comments in https://github.com/zed-industries/zed/pull/34798)
- `DispatchResult` now reports when the current keystrokes are already a complete binding in the active context stack (`pending_has_binding`). We only start the 1s flush timer in that case. Pure prefixes or sequences that only match in other contexts—stay pending indefinitely, so leader-style combos like `space f g` no longer evaporate after a second.
- `Window::dispatch_key_event` cancels any prior timer before scheduling a new one and only spawns the background flush task when `pending_has_binding` is true. If there’s no matching binding, we keep the pending keystrokes and rely on an explicit Escape or more typing to resolve them.

Release Notes:

- Multi-stroke keybindings (like `cmd-k cmd-s`, or `g d` in vim mode) will no-longer time out waiting for input. Before this change Zed would wait at most 1s for the second keystroke. In the case where the start of the multi-stroke binding is itself a valid key action (or input text) then the 1s timeout will still apply.

### DIFF
--- a/crates/collab/src/tests/integration_tests.rs
+++ b/crates/collab/src/tests/integration_tests.rs
@@ -6551,12 +6551,12 @@ async fn test_pane_split_left(cx: &mut TestAppContext) {
         assert!(workspace.items(cx).collect::<Vec<_>>().len() == 2);
     });
     cx.simulate_keystrokes("cmd-k");
-    // sleep for longer than the timeout in keyboard shortcut handling
-    // to verify that it doesn't fire in this case.
+    // Sleep past the historical timeout to ensure the multi-stroke binding
+    // still fires now that unambiguous prefixes no longer auto-expire.
     cx.executor().advance_clock(Duration::from_secs(2));
     cx.simulate_keystrokes("left");
     workspace.update(cx, |workspace, cx| {
-        assert!(workspace.items(cx).collect::<Vec<_>>().len() == 2);
+        assert!(workspace.items(cx).collect::<Vec<_>>().len() == 3);
     });
 }
 
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -21994,13 +21994,7 @@ impl Editor {
             .pending_input_keystrokes()
             .into_iter()
             .flatten()
-            .filter_map(|keystroke| {
-                if keystroke.modifiers.is_subset_of(&Modifiers::shift()) {
-                    keystroke.key_char.clone()
-                } else {
-                    None
-                }
-            })
+            .filter_map(|keystroke| keystroke.key_char.clone())
             .collect();
 
         if !self.input_enabled || self.read_only || !self.focus_handle.is_focused(window) {
--- a/crates/gpui/src/key_dispatch.rs
+++ b/crates/gpui/src/key_dispatch.rs
@@ -121,6 +121,7 @@ pub(crate) struct Replay {
 #[derive(Default, Debug)]
 pub(crate) struct DispatchResult {
     pub(crate) pending: SmallVec<[Keystroke; 1]>,
+    pub(crate) pending_has_binding: bool,
     pub(crate) bindings: SmallVec<[KeyBinding; 1]>,
     pub(crate) to_replay: SmallVec<[Replay; 1]>,
     pub(crate) context_stack: Vec<KeyContext>,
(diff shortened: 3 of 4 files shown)


xipeng-jin:
A few comments to be considered:
1. Do we want to consider to create a setting for customizing the default timeout duration? It is currently hard coded value with 1 second. For example, the user want to change to 5 second to match VSCode. If answer is yes, we maybe also consider to push this settings in a second PR which will be included in the which-key own module.
2. We might want to think more ahead about dealing with leader key. For example, the default leader key is `<space>` in Vim. And in Zed we currently bound `<space>` to a action  `vim::WrappingRight` at default, which would cause a conflict with the leader key since people will expect no timeout setting when they type the leader key `<space>`, but in zed we already have a matching keybinding with it so that it will trigger the action `vim::WrappingRight` after 1 second. And also at default it will automatically bring up the which-key menu after typing the leader key in Neovim, so it shouldn't use the timeout setting in this case.

cc.
@ConradIrwin . Please review as your last comment from that thread https://github.com/zed-industries/zed/pull/34798#issuecomment-3424638788. Thank you!
@Cretezy . If you still work on this feature, I am happy to have your feedback here regarding to this timer configuration. If you want to pair on this I am also happy to do it! Really appreciate your previous work!

xipeng-jin:
There is one integration test `collab tests::integration_tests::test_pane_split_left` we have to modify to adapt the new behavior from this adaptive timeout settings for which-key system. To be more specific, we need to adjust the test so it now document the new “no timeout for unmatched prefixes” behavior which means after triggering `cmd-k` and waiting two seconds, the follow-up `left` keystroke is still treated as part of the multi-stroke binding, so the workspace ends up with three items instead of two. Reference the commit https://github.com/zed-industries/zed/pull/42659/commits/8684055f3327b4ced795cd3d82b70105fcb6f096

ConradIrwin:
Thanks for this!

I'm worried about this condition:
```
                .filter(|key_down| {
                    key_down.keystroke.key_char.is_some()
                        && key_down
                            .keystroke
                            .modifiers
                            .is_subset_of(&Modifiers::shift())
                })
```

because I think keys like `option-x` on macOS should still be treated as character input. Do we need the additional filter on the shift modifier? In theory (though maybe not in practice) `key_char` should be enough.

xipeng-jin:
Yes, thank you for correcting me. I think you are right. I was missing the default functionalities for typing `option-x` (macOS) or `alt-x` (Windows/Linux) to entering the special characters, since I rarely use those characters. And I checked `keystroke.key_char` has covered all the cases for input text in insert-ish mode so we should be good to only use `key_char` in the filter. No need for filtering on the shift modifier.

I removed that filter in the new commit and I noticed there is minor difference when you type `option-x` or `option` plus any characters compared with the normal characters or using `shift` to enter capital case. Please check the following recording.

https://github.com/user-attachments/assets/207d59eb-36b3-4d96-ac59-d3a2a24c09ef

As you can see, normally we will have a pending effect for the input text when we didn't enter the next `key_char` immediately (in this example is character `j` or `J`). However for special characters we miss this pending effect and it simply didn't show the special character when we type `option-x`, but after the 1 second timer if you didn't type the next one `k` it will eventually show up. I didn't have a chance to dig into this difference yet, but I would like to align with you first @ConradIrwin this also seems a issue to you?

xipeng-jin:
Okay, I found the root cause for this inconsistent behavior. Since `observe_pending_input` only collected pending text when modifiers were in the subset of Shift modifier, Option/Alt keystrokes had a `key_char` but also an Alt modifier, so no pending text was inserted/highlighted even though the timeout stayed active. Therefore, it was dropping Alt-generated characters and leaving them invisible until the timeout fired. https://github.com/zed-industries/zed/pull/42659/commits/51798dbf89809c3515aedc942b3ee1dac6ffa83a

As a result, we need to change the pending text/highlight respecting any keystroke that produces a `key_char`, so Option/Alt characters render immediately while the 1s timeout still runs for ambiguous bindings.

ConradIrwin:
Great, thank you!

PR #42683, opened 2025-11-13
Support Agent Servers on remoting

rtfeldman:
Screenshot 2025-11-13 at 6 53 39 PM

Also added support for per-target env vars to Agent Server Extensions

Closes https://github.com/zed-industries/zed/issues/42291

Release Notes:

- Per-target env vars are now supported on Agent Server Extensions
- Agent Server Extensions are now available when doing SSH remoting


GeorgiGeor:
Adding local_cc external agent is not very clear, would be nice to make it simpler, probably not only one who prefers CLI Claude Code

PR #42694, opened 2025-11-14
quick fix: Circumvent staff-only collab connection

mchisolm0:
Switch the title bar sign in button to use the same sign in method as the collab panel. This is only for illustration purposes.

Closes #ISSUE (No current issue; Discussion over email with Esther.)

Release Notes:

- N/A

### DIFF
--- a/crates/title_bar/src/title_bar.rs
+++ b/crates/title_bar/src/title_bar.rs
@@ -653,8 +653,9 @@ impl TitleBar {
                 window
                     .spawn(cx, async move |cx| {
                         client
-                            .sign_in_with_optional_connect(true, cx)
+                            .connect(true, cx)
                             .await
+                            .into_response()
                             .notify_async_err(cx);
                     })
                     .detach();


mchisolm0:
This should build correctly. I believe this will let the user sign in with the title bar sign in button (in the top right) and connect to the collab panel without having to also click sign in on the collab panel.

I did this to troubleshoot why a deep link like https://zed.dev/channel/git-1-0-25944/notes will fail sometimes after opening Zed.

I think I found that even using this method, the deep link still fails unless the user has also opened the collab panel after signing in at least one time.

I tested this by...
1. Signing out
2. Signing in with the title bar sign in button in the top right corner
3. Staying on the project panel (so the collab panel is not rendered)
4. Engaging the command pallete with CMD+Shift+P
5. Pasting the deep link https://zed.dev/channel/git-1-0-25944/notes and navigating to it
I get the following:
```console
2025-11-13T22:41:22-06:00 INFO  [client] set status on client 95968: SignedOut
2025-11-13T22:44:38-06:00 INFO  [client] set status on client 95968: Authenticating
2025-11-13T22:44:48-06:00 INFO  [client] set status on client 95968: Authenticated
2025-11-13T22:44:48-06:00 INFO  [client] set status on client 95968: Connecting
2025-11-13T22:44:49-06:00 INFO  [client] connected to rpc endpoint https://collab.zed.dev/rpc
2025-11-13T22:44:49-06:00 INFO  [client] set status on client 95968: Connected { peer_id: PeerId { owner_id: 904, id: 46806 }, connection_id: ConnectionId { owner_id: 0, id: 2 } }
2025-11-13T22:45:33-06:00 INFO  [sqlez::migrations] Found 16 foreign key relationships to check
2025-11-13T22:45:37-06:00 ERROR [crates/zed/src/main.rs:963] failed to open channel notes

Caused by:
    no channel for id: 25944
```
In contrast, if I follow the same steps above but replace step 3 with navigating to the collab panel, there is no longer an error.

If I follow the steps having run from main, I get the following error because the title bar sign in button is using the `sign_in_with_optional_connect()`
```console
2025-11-13T22:56:58-06:00 INFO  [client] set status on client 0: Authenticating
2025-11-13T22:57:07-06:00 INFO  [client] set status on client 95968: Authenticated
2025-11-13T22:57:30-06:00 INFO  [client] set status on client 95968: Reauthenticating
2025-11-13T22:57:30-06:00 INFO  [client] set status on client 95968: Reauthenticated
2025-11-13T22:57:41-06:00 ERROR [crates/zed/src/main.rs:963] failed to open channel notes

Caused by:
    10s elapsed without receiving channels
```

I think the actual solution may need to be triggering a popup after the user signs in with the title bar sign in button to ask if the user wants to also sign into collab features. This way, I think the `sign_in_with_optional_connect()` function called by the title bar sign in button will still meet what I think the team intends of only connecting collab if 1. The user is Zed staff or 2. The user *wants* collab features.

crates/client/src/client.rs lines 983 to 999
```rs
cx.spawn({
    let client = self.clone();
    async move |cx| {
        let is_staff = is_staff_rx.await?;
        if is_staff {
            match client.connect_with_credentials(credentials, cx).await {
                ConnectionResult::Timeout => Err(anyhow!("connection timed out")),
                ConnectionResult::ConnectionReset => Err(anyhow!("connection reset")),
                ConnectionResult::Result(result) => {
                    result.context("client auth and connect")
                }
            }
        } else {
            Ok(()) // <--- This line might need to be a popup
                       //           that runs Ok(()) if the user ops out of collab features
                       //           but runs something else if the user ops into collab features
        }
    }
})
```

The feature working once everything is signed in:
Successful Sign in and Note Test

The feature failing on Zed 0.212.5 6f8b00ca21f8917f99c81518280b9f1b61ad83a1
Failed sign in and deep link notes

mchisolm0:
On Linux, I verified the steps fail in the same way, visually, but I get no debugging output. I do not have a Windows computer to test on.

On the PR Review office hours on Wednesday November 5, there were some in attendance with Windows machines who sounded like they had the same issue.

maxdeviant:
Thank you for the PR, but the current behavior is intentional to prevent unnecessary load on our Collab server.

We will be doing some work on collaboration in the near future that will address some of these foundational issues.

mchisolm0:
Thank you! The PR was only for the purpose of illustration so the team could see why Zed would fail to open Collab note links for some users.

Looking forward to the collaboration working that's coming (:

PR #42703, opened 2025-11-14
git: Add UI for deleting branches

### LINKED ISSUE #42641: Git: UI for deleting branches
### Summary

Currently there's no way to delete git branches from within Zed; we should add some UI for this.

errmayank:
Closes #42641

WIP for deleting Git branches via UI. Right now it does not handle:

- [ ] Force delete (optional)
- [x] Remote repositories

Maybe we could also:

- [x] Show a toast when branch is successfully deleted
- [x] Handle errors more gracefully

EDIT: outdated - https://github.com/user-attachments/assets/7f84fe94-2f65-46df-8ad8-29e15d8f4fcb

Release Notes:

- Added UI for deleting Git branches


versecafe:
would it make sense to have a delete button in `git: branch` that would make it possible to delete multiple branches and consolidate the menus together

cole-miller:
Thanks! Initial feedback:

- For discoverability, I think this should probably be integrated into the normal branch list modal, instead of being a separate list where confirming an entry means deleting the branch. We could have an action and a keybinding in the `git: branch` picker for deleting the selected branch, and add a footer to the modal that makes this discoverable. I don't think we need a separate `DeleteBranch` action for deploying that modal list.
- We should use a toast and not an OS notification to alert the user when deleting a branch fails.
- I think we probably don't need to show a toast when a branch is successfully deleted. Instead, we should just keep the branch list modal open after successfully doing `git branch -d` and removing that branch from the list. (If `git branch -d` fails, we should keep it in the list.)
- For the remote case, looking at how `Repository::reset` is implemented and following the trail of `proto::GitReset` is a good way to see how it all fits together. Happy to pair on this as well: https://cal.com/esther-trapadoux-zed/30min

errmayank:
Thanks for the pointers, I'll check these out.

> Happy to pair on this as well

@cole-miller Great! I've booked a session for ~~next Monday~~ this Friday.

PR #42706, opened 2025-11-14
Improve the location of pop-up windows

libondev:
Improve the location where settings and rule pop-ups appear. When the user has multiple monitors and the zed editor is not in the main window, the opened settings/rules pop-up window will always appear in the main monitor. This PR adjusts this behavior so that its pop-up window always appears in the monitor of the last focused ze editor pop-up window.

Release Notes:

- Improve the location of pop-up windows

### DIFF
--- a/crates/rules_library/src/rules_library.rs
+++ b/crates/rules_library/src/rules_library.rs
@@ -3,9 +3,9 @@ use collections::{HashMap, HashSet};
 use editor::{CompletionProvider, SelectionEffects};
 use editor::{CurrentLineHighlight, Editor, EditorElement, EditorEvent, EditorStyle, actions::Tab};
 use gpui::{
-    Action, App, Bounds, DEFAULT_ADDITIONAL_WINDOW_SIZE, Entity, EventEmitter, Focusable,
-    PromptLevel, Subscription, Task, TextStyle, TitlebarOptions, WindowBounds, WindowHandle,
-    WindowOptions, actions, point, size, transparent_black,
+    Action, App, DEFAULT_ADDITIONAL_WINDOW_SIZE, Entity, EventEmitter, Focusable, PromptLevel,
+    Subscription, Task, TextStyle, TitlebarOptions, WindowHandle, WindowOptions, actions, point,
+    size, transparent_black,
 };
 use language::{Buffer, LanguageRegistry, language_settings::SoftWrap};
 use language_model::{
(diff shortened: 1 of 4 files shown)


crates/workspace/src/window_utils.rs
probably-neb, line 1: The core logic in this file should be in `gpui` rather than workspace, as it applies to any window not just a workspace window. 


`centered_bounds_for_display_id` doesn't need to exist I don't think, it just calls the constructor of `WindowBounds`

crates/workspace/src/window_utils.rs
```diff
@@ -0,0 +1,43 @@
...
+        .update(cx, |_, window, cx| {
+            let center = window.bounds().center();
+            cx.displays()
+                .into_iter()
+                .find(|display| display.bounds().contains(&center))
+                .map(|display| display.id())
+        })
+        .ok()
+        .flatten()
+}
```
probably-neb, line 18: This function should be a method on `Window`

crates/workspace/src/window_utils.rs
```diff
@@ -0,0 +1,43 @@
...
+                let center = window.bounds().center();
+                cx.displays()
+                    .into_iter()
+                    .find(|display| display.bounds().contains(&center))
+                    .map(|display| display.id())
+            })
+            .ok()
+            .flatten()
+    })
+}
```
probably-neb, line 43: This function should also be removed, and in the rules library window it should be calculated as `cx.active_window().display().id`

crates/workspace/src/window_utils.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Compute centered window bounds on a given display id (or primary if None).
+pub fn centered_bounds_for_display_id(
+    display_id: Option<DisplayId>,
+    size: Size<Pixels>,
+    cx: &mut App,
+) -> WindowBounds {
+    WindowBounds::Windowed(Bounds::centered(display_id, size, cx))
+}
```
probably-neb, line 27: This function should just be inlined into all callers and removed

crates/rules_library/src/rules_library.rs
```diff
@@ -118,7 +118,14 @@ pub fn open_rules_library(
         let store = store.await?;
         cx.update(|cx| {
             let app_id = ReleaseChannel::global(cx).app_id();
-            let bounds = Bounds::centered(None, size(px(1024.0), px(768.0)), cx);
+            // Pick the display of the last active window and center on it
```
probably-neb, line 121: Please remove all the line comments. Line comments describing what the following line of code is doing as LLMs often do is not the Zed style

probably-neb (changes requested):
Thanks for taking a crack at this, it's definitely a worthwhile UX improvement. Unfortunately the code does not work as intended (not your fault! I had to verify myself whether `display.bounds` would work as this PR assumed), and has quite a few style / structure problems.

Based on [our policy](https://github.com/zed-industries/zed/blob/4a36f67f94970958cb991872f6ad5f27185c8f3d/CONTRIBUTING.md?plain=1#L48) I'm going to close this for now. However, I've left review comments for what I think needs to change to get this to a place where we can merge it so if you'd like to fix the issues I've outlined, feel free to @-mention me once you do and I'd be happy to re-review and re-open.

Thanks!

libondev:
Ok I'll try to modify it a bit and when I'm done I'll try to apply for review again

PR #42717, opened 2025-11-14
Add vertical tabs

bstst:
Closes https://github.com/zed-industries/zed/discussions/24006

Requested this feature a couple years ago. It didn't gain much traction for some reason (similar feature request on vscode has 618-ish upvotes — https://github.com/microsoft/vscode/issues/108264).

Vertical tabs is the only missing feature that's stopping me from using Zed. Thus I decided to implement it myself.

Release Notes:

- Added vertical tab support (configurable in settings): 
image

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -18,3 +18,6 @@ rustflags = [
 
 [env]
 MACOSX_DEPLOYMENT_TARGET = "10.15.7"
+
+[net]
+git-fetch-with-cli = true
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -18,6 +18,3 @@ rustflags = [
 
 [env]
 MACOSX_DEPLOYMENT_TARGET = "10.15.7"
-
-[net]
-git-fetch-with-cli = true
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1029,7 +1029,9 @@
     // Whether or not to show the navigation history buttons.
     "show_nav_history_buttons": true,
     // Whether or not to show the tab bar buttons.
-    "show_tab_bar_buttons": true
+    "show_tab_bar_buttons": true,
+    // Orientation of the tab bar. One of: ["horizontal", "vertical"]
+    "orientation": "horizontal"
   },
   // Settings related to the editor's tabs
   "tabs": {
--- a/crates/settings/src/settings_content/workspace.rs
+++ b/crates/settings/src/settings_content/workspace.rs
@@ -350,6 +350,29 @@ pub enum RestoreOnStartupBehavior {
     LastSession,
 }
 
+#[derive(
+    Copy,
+    Clone,
+    PartialEq,
+    Eq,
+    Default,
+    Serialize,
+    Deserialize,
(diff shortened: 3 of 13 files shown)


bstst:
@cla-bot check

Angelk90:
@bstst : You can have the user decide whether to put it on the left or right via settings.

bstst:
> @bstst : You can have the user decide whether to put it on the left or right via settings.

Not sure about that. I mean, currently tabs are only on top. No bottom option. I naturally followed the "left" pane position, without even thinking someone might want to place them on the right.

Change is not "trivial", as it would also imply allowing users to place them on the bottom.

omid:
Thanks for the PR

> naturally

It's _naturally_ on the right, for me 😅

bstst:
> Thanks for the PR
> 
> > naturally
> 
> It's _naturally_ on the right, for me 😅

Absolutely. Didn't want to sound ignorant. I shall attempt to alleviate this.

mattermill:
Thanks for the work on this @bstst.

First of all, I'm an Arc/Dia user who loves vertical tabs within that context, but, unfortunately we're not ready to take on the maintenance burden from the complexity introduced here.

This could be a good candidate once our extension system is more mature, though :)

bstst:
> Thanks for the work on this @bstst.
> 
> First of all, I'm an Arc/Dia user who loves vertical tabs within that context, but, unfortunately we're not ready to take on the maintenance burden from the complexity introduced here.
> 
> This could be a good candidate once our extension system is more mature, though :)

I understand. To do this properly the panel needs to be refactored extensively. And that would be a massive change. And that kind of work should be evaluated, planned, milestoned etc, not just dropped in a PR. I'll just keep on enjoying zed with my dev build and merge upstream occasionally, until another window of opportunity arises.

Ignoring that, I shall always be curious how people manage to work/browse with horizontal tabs! ;)

mattermill:
> Ignoring that, I shall always be curious how people manage to work/browse with horizontal tabs! ;)

`Ctrl+Tab` 😅

RieBi:
Hey @bstst I wonder if this feature could be introduced as an extension to VSCode, or the extension system at the current state does not support the changes needed for the implementation?

bstst:
> Hey @bstst I wonder if this feature could be introduced as an extension to VSCode, or the extension system at the current state does not support the changes needed for the implementation?

VSCode extensions won't achieve that. Some claim to do that, but it's not the kind of solution I would go for (crude core css hacks breaking layout here and there).

Zed initial layot in the code is very sensible, so is Rust by itself.

VSCode on the other hand is not as flexible in its layout and components. And their event pattern is, uh, less sensible. I tried to achieve vertical tabs in VSCode a couple of times, but didn't even submit the PR, even though it was working, only because I was doing it in a hacky approach. Doing it in a proper approach would involve a significant refactoring. And most definitely would not be merged, just like with Zed.

These kind of changes need careful planning and approach and a lot of time.

ugreg:
As a hack for vertical file list, we can always use Command + P to open a file list in the current project. Easier to scroll through that list than shifting though tabs.

image

PR #42741, opened 2025-11-14
Add support for LMStudio as a completion provider

dzervas:
Closes #42010 

Release Notes:

- Added the ability to use LM Studio as an edit prediction provider

Support for LM Studio's `/api/v0/completions` endpoint was added as well since that's the endpoint that manages FIM and normal completion.

Tested with `qwen2.5-coder-1.5b-instruct` and `zeta`, both quite fast on my sub-optimal 5900XT (8GB VRAM)

This is my first contribution so I didn't bother to add a proper settings UI (for model selection, temperature, etc.) since it's quite more work - I hope I'll work on that soon though!

To select a model to use, the `ZED_LMSTUDIO_COMPLETION_MODEL` is tried, otherwise an empty string is sent to lmstudio which by default fallbacks to the first loaded model - this of course will result in poor UX since if there's no loaded model the completion will silently fail

Follow up work and questions regarding them (I'll open an issue/discussion regarding them when the time comes):
- tests? do you require tests in the current PR context as well?
- Docs
- Add settings for lmstudio edit predictions (temperature, model, etc.) - should I add an `lmstudio_edit_predictions` under `features` or an `edit_predictions` object under `language_models.lmstudio`?
- Show a warning if:
  - lmstudio is unavailable
  - no model is available or no specific model is set while there's no model loaded
  - the configured model is unavailable
  - an error is returned from the API (make sure to debounce somehow)
- Provide a UI in the edit predictions button to select a model - is there scaffold to add buttons/dropdowns/toggles to the edit predictions buttons? supermaven for example only has a single link to initialize


dzervas:
@cla-bot check

osyvokon:
Thanks for the PR! I haven't looked at the code yet, just runned it locally. It works quite well! The only problem is that Zed hangs after a minute or two. The profiler shows something seemengly unrelated to this branch, but I have to investigate it first before merging.

dzervas:
@osyvokon yes I saw that as well and just thought that it was my gpu weakling. I ran the main branch before my changes and I get the same behavior though

dzervas:
@syepes I'm closing the PR for the time being since I can't find the courage to code due to the agentic AI craze, hopefully I'll get back to it at some point.

If anyone wants to continue working on this, feel free

PR #42744, opened 2025-11-14
Make wild the default linker on Linux

yara-blue:
It is significantly faster and no production ready. Still keeping it off in CI as its a relatively new linker.

We install it by default on linux now.

Closes #ISSUE

Release Notes:

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -16,5 +16,13 @@ rustflags = [
     "target-feature=+crt-static", # This fixes the linking issue when compiling livekit on Windows
 ]
 
+[target.x86_64-unknown-linux-gnu]
+linker = "clang"
+rustflags = ["-C", "link-arg=--ld-path=wild"]
+
+[target.aarch64-unknown-linux-gnu]
+linker = "clang"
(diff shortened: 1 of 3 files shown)


Be-ing:
While I use wild myself and would recommend everyone developing on Linux to use it too, I also don't think it's great to impose a requirement on the build that doesn't really need to be there. Especially considering wild is such a new project and hasn't been widely packaged yet, this is a cumbersome requirement, particularly for packagers. Packagers would need to either patch `.config/cargo.toml`, override `.config/cargo.toml` with the `RUSTFLAGS` environment variable, or ensure wild is packaged in their packaging system. I think installing wild by default from `script/linux` is a good idea, but it's fine how it is now and imposing it as a requirement for everyone in the project's `.config/cargo.toml` creates more trouble for contributors and packagers than it's worth.

Also, since lld is now the default Rust linker on x86_64-unknown-linux-gnu [as of Rust 1.90](https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/), the benefit of using wild is much less than when GNU ld was the default.

PR #42747, opened 2025-11-14
Fix wild install script

AlvaroParker:
Use [`command`](https://www.gnu.org/software/bash/manual/bash.html#index-command) instead of `which` to check if `wild` is installed. 

Using `which` will result in an error being printed to stdout: 

```bash
./script/install-wild
which: invalid option -- 's'
/usr/local/bin/wild
Warning: existing wild 0.6.0 found at /usr/local/bin/wild. Skipping installation.
```

Release Notes:

- N/A


yara-blue:
thanks!

PR #36841, opened 2025-08-24
Fix Git blame popup overlapping with context menu

ghost:
Closes #26949 

Release Notes:

- Fixed Git blame context menus being obscured by commit detail tooltips

Changes:
The issue was that Git blame tooltips were using hoverable_tooltip which stays visible when you hover over them, so they'd appear on top of right-click context menus and make them unusable.

Fixed by:
Switching to regular tooltip instead of hoverable_tooltip - these dismiss automatically on mouse events like right-clicks
Adding a check to not show tooltips when context menus are already open
Bumped context menu z-index priority as a safety net
The main fix was just changing the tooltip type - turns out regular tooltips already handle this case properly by dismissing on right-clicks.

https://github.com/user-attachments/assets/6c6970b1-4e37-4d34-9d87-20b5a8c30415

PR #42764, opened 2025-11-14
git_ui: Fix tooltip overlaying context menu in git blame

### LINKED ISSUE #26949: Can't use right-click menu in Git blame
### Summary

Right click menu is overdrawn by commit details.

https://github.com/user-attachments/assets/6a0bee74-05b5-4b14-a6d1-d54510870d05

Zed: v0.179.0 (Zed Nightly 94b63808e0593ee927151b751f290ebb9e31dc25) 
OS: Linux X11 arch unknown
Memory: 60.3 GiB
Architecture: x86_64
GPU: AMD Radeon RX 6750 XT (RADV NAVI22) || radv || Mesa 24.3.4-arch1.1

amustaque97:
Closes #26949

## Summary

1. Split editor references to avoid borrow conflicts in event handlers
2. Check [has_mouse_context_menu()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) state directly in tooltip conditional instead of caching stale value
3. Restructured context menu deployment to ensure proper sequencing: hide popover → build menu → deploy menu → notify for re-render

**Screen recording**

https://github.com/user-attachments/assets/8a00f882-1c54-47b0-9211-4f28f8deb867

Release Notes:

- Fixed an issue where the context menu in the Git Blame view would be frequently overlapped by the commit information tooltip.

### DIFF
--- a/crates/git_ui/src/blame_ui.rs
+++ b/crates/git_ui/src/blame_ui.rs
@@ -85,6 +85,7 @@ impl BlameRenderer for GitBlameRenderer {
                         .on_mouse_down(MouseButton::Right, {
                             let blame_entry = blame_entry.clone();
                             let details = details.clone();
+                            let editor = editor.clone();
                             move |event, window, cx| {
                                 cx.stop_propagation();
 
--- a/crates/git_ui/src/blame_ui.rs
+++ b/crates/git_ui/src/blame_ui.rs
@@ -55,6 +55,11 @@ impl BlameRenderer for GitBlameRenderer {
         } else {
             None
         };
+
+        // Split editor clones for different purposes to avoid borrow conflicts
+        let editor_for_mouse_down = editor.clone();
+        let editor_for_tooltip = editor;
+
         Some(
             div()
                 .mr_2()
@@ -81,10 +86,13 @@ impl BlameRenderer for GitBlameRenderer {
                             let blame_entry = blame_entry.clone();
                             let details = details.clone();
                             move |event, window, cx| {
+                                // Prevent the right-click event from bubbling up to parent elements
+                                cx.stop_propagation();
+
                                 deploy_blame_entry_context_menu(
                                     &blame_entry,
                                     details.as_ref(),
-                                    editor.clone(),
+                                    editor_for_mouse_down.clone(),
                                     event.position,
                                     window,
                                     cx,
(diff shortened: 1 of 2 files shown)


6639835 (approved):

MrSubidubi (changes requested):
Thanks for this.

While this is definitely an approach to fix this, this effectively eliminates the ability to have tooltips in deferred draws, does it not? So I doubt this is the proper fix for this.

Additionally, it seems the tooltip can get both stale and annoying, especially once the context menu is deployed. Feels like the approach in the other PR might be more sensible to approach?

esthertrapadoux:
@amustaque97 hi! We have a few pairing sessions still available this week if you want to chat with the team and work through some of the feedback above: https://cal.com/esther-trapadoux-zed/30min

crates/git_ui/src/blame_ui.rs
```diff
@@ -55,6 +55,11 @@ impl BlameRenderer for GitBlameRenderer {
         } else {
             None
         };
+
+        // Split editor clones for different purposes to avoid borrow conflicts
+        let editor_for_mouse_down = editor.clone();
+        let editor_for_tooltip = editor;
```
MrSubidubi: We shouldn't be needing these, as mentioned in https://github.com/zed-industries/zed/pull/36841#discussion_r2339525957 as well

crates/git_ui/src/blame_ui.rs
```diff
@@ -81,10 +86,13 @@ impl BlameRenderer for GitBlameRenderer {
                             let blame_entry = blame_entry.clone();
                             let details = details.clone();
                             move |event, window, cx| {
+                                // Prevent the right-click event from bubbling up to parent elements
```
MrSubidubi:
```suggestion
```

MrSubidubi (changes requested):

crates/git_ui/src/blame_ui.rs
amustaque97: @MrSubidubi, thank you so much for the quick review 👏🏻 

I could be wrong, but I think it’s required. Based on the screenshot, I need to clone it first before I can use it. Please feel free to correct me if I'm missing anything here 🤓 

Screenshot 2025-12-03 at 11 32 07 PM

MrSubidubi: You should be able to clone the editor after `let details = details.clone()`, then it should be movable into the moved closure

crates/git_ui/src/blame_ui.rs
```diff
@@ -81,10 +86,12 @@ impl BlameRenderer for GitBlameRenderer {
                             let blame_entry = blame_entry.clone();
                             let details = details.clone();
                             move |event, window, cx| {
```
MrSubidubi, line 85:
```suggestion
                            let details = details.clone();
                            let editor = editor.clone();
                            move |event, window, cx| {
```

MrSubidubi: I think this should work, not entirely sure though, but something amongst these lines.

crates/git_ui/src/blame_ui.rs
amustaque97: Thank you for helping me here 🙌🏻

amustaque97: I ran another basic sanity check, and the fix is working as expected.

MrSubidubi (changes requested):
Nice, this looks very good now. Thank you!

MrSubidubi (approved):

PR #42775, opened 2025-11-14
Check user preference before wrapping commit message

### LINKED ISSUE #33441: git: Automatic line wrapping in commit messages causes invalid commits
### Summary

# Bug Report: Automatic line wrapping in commit messages causes invalid commits

## Description
Zed automatically wraps long commit messages to new lines, which causes repositories to reject commits with invalid commit message format.

## Steps to Reproduce
1. Open Zed's commit interface
2. Write a long commit message (over ~80 characters), for example:
   ```
   feat(zed): here is an incredible super long commit name that is over 80-something characters (#500)
   ```
3. Zed automatically wraps the text to multiple lines
4. Commit using Zed's UI
5. In a terminal, run `git commit --amend` to see the actual commit message
6. You'll notice that Zed has inserted line breaks where it wrapped the text

## Expected Behavior
- Zed should allow users to manually control line breaks in commit messages
- If the commit message is long, it should remain on a single line unless the user manually adds line breaks
- Users should be able to format their commit messages as needed (e.g., manually breaking at appropriate points)

## Actual Behavior
- Zed automatically inserts line breaks in long commit messages
- This causes the repository to reject the commit due to invalid commit message format
- Users cannot override this automatic wrapping behavior

## Environment
- Zed version: 0.191.5 (this is not a new bug caused by a new version)
- Operating System: Ubuntu 24.04

Zed: v0.192.5 (Zed) 
Memory: 15.3 GiB
Architecture: x86_64
GPU: Intel(R) Graphics (ADL GT2) || Intel open-source Mesa driver || Mesa 24.2.8-1ubuntu1~24.04.1

mchisolm0:
Closes [#33441](https://github.com/zed-industries/zed/issues/33441) (does not yet) 

Release Notes:

- Fixed long commit messages to respect user preference

### DIFF
--- a/crates/cloud_zeta2_prompt/src/cloud_zeta2_prompt.rs
+++ b/crates/cloud_zeta2_prompt/src/cloud_zeta2_prompt.rs
@@ -31,7 +31,7 @@ const MARKED_EXCERPT_INSTRUCTIONS: &str = indoc! {"
 
     Other code is provided for context, and `…` indicates when code has been skipped.
 
-    # Edit History:
+    ## Edit History
 
 "};
 
@@ -49,7 +49,7 @@ const LABELED_SECTIONS_INSTRUCTIONS: &str = indoc! {r#"
         println!("{i}");
     }
 
-    # Edit History:
+    ## Edit History
 
 "#};
 
@@ -89,7 +89,7 @@ const NUMBERED_LINES_INSTRUCTIONS: &str = indoc! {r#"
 const STUDENT_MODEL_INSTRUCTIONS: &str = indoc! {r#"
     You are a code completion assistant that analyzes edit history to identify and systematically complete incomplete refactorings or patterns across the entire codebase.
 
-    # Edit History:
+    ## Edit History
 
     "#};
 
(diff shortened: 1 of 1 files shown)


mchisolm0:
For convenience, here is the long commit message the OP in the issue gave as an example. This PR should let it stay on a single line after committing a change with this message and checking it with `git log`.
```
feat(zed): here is an incredible super long commit name that is over 80-something characters (#500)
```

Also, I do not understand what the tests are saying is wrong with my style. I am happy to fix it or adjust the solution entirely depending on what the team prefers.

Thanks for your time! :grinning:

cole-miller:
Thanks for opening a PR! I don't think we should be reading the soft wrap setting here--soft wrapping refers to how we visually wrap the text in an editor without inserting actual `\n` characters, and the original issue is referring to "hard wrapping" that does insert `\n` characters.

There is actually already a way to configure or disable the width used for hard wrapping commit messages:

```
  "languages": {
    "Git Commit": {
      "preferred_line_length": 100
    }
  }
```

Although that is a bit hard to discover. It seems like we should probably add it to the documentation, along with information about how to disable this hard wrapping entirely.

crates/git_ui/src/git_panel.rs
```diff
@@ -1559,6 +1560,12 @@ impl GitPanel {
         } else if message.trim().is_empty() {
             return None;
         }
+        if let Some(language) = &git_commit_language {
+            let settings = language_settings(Some(language.name()), None, cx);
+            if settings.soft_wrap == ConfigSoftWrap::None {
+                return Some(message);
+            }
+        }
```
Anthony-Eid: Language settings have a default that isn't being accessed if `git_commit_language` is None here.

you could do something like this instead
```rust
let settings = language_settings(git_commit_language.map(|lang| lang.name()), None, cx);
if settings.soft_wrap == ConfigSoftWrap::None {
    return Some(message);
}
```

mchisolm0: Thank you, @Anthony-Eid! For the sake of me making sure I understood you I made a new PR https://github.com/zed-industries/zed/pull/43305, but I think we will intend to close it without merging.

My understanding from Cole's comment is the code in zed-industries:main already works as expected. The problem is we need to make the `"preferred_line_length"` setting more discoverable. I believe the current plan is for me to update the documentation.

@cole-miller, can you confirm that:
1. Zed's current `git_panel` line wrapping works as intended, so this PR should not be merged,
2. The current plan is for me to update the docs like you mentioned in the Demo Day, and
3. Would it be helpful to have a link to the relevant docs next to "Configure" in the settings panel? (See picture)

image

PR #42783, opened 2025-11-15
Add Groq LLM provider support

senoldogann:
- Add Groq crate with model definitions (Llama 3.1 8B, Llama 3.3 70B, GPT OSS models, etc.)
- Implement GroqLanguageModelProvider with OpenAI-compatible API support
- Add Groq settings configuration and UI
- Register Groq provider in the language model registry
- Support for custom models via settings

This adds Groq as a new LLM provider option for Zed editor, following the same pattern as other providers like OpenAI, Anthropic, etc.

Closes #ISSUE

Release Notes:


maxdeviant:
Thank you for the PR!

We're in the process of making it possible to add new model providers via extensions. This should be ready soon, at which point Groq can be added through that mechanism.

PR #36497, opened 2025-08-19
Add support for encodings other than UTF-8

EuclidDivisionLemma:
Add the ability to open and save files in different encodings. Closes #16965

PR #42786, opened 2025-11-15
Fix UTF-16 BE BOM detection

Hojun-Cho:
use UTF_16BE instead of UTF_8 when detecting [0xFE, 0xFF] BOM.


Hojun-Cho:
@cla-bot check

SomeoneToIgnore:
I've got quite confused at first given that there was no such file in `main`, and only then found that the base branch is github.com/zed-industries/zed/tree/add_support_for_non_utf_encodings

I'm afraid this kind of collaboration scenario is a bit too complex, and you'd better be off commenting the original PR: https://github.com/zed-industries/zed/pull/36497

Also, given that's closed now, this is not going anywhere for sure.

PR #42792, opened 2025-11-15
workspace: Add `ctrl-w x` support for vim

mfroeh:
Adds support for the vim CTRL-W x keybinding, which swaps the active pane with the next adjacent one, prioritizing column over row and next over previous. Upon swap, the pane which was swapped with is activated (this is the vim behavior).

See also https://github.com/vim/vim/blob/ca6a260ef1a4b4ae94bc71c17cbabf8f12bf0f8c/runtime/doc/windows.txt#L514C1-L521C24

Release Notes:

- Added ctrl-w x keybinding in Vim mode, which swaps the active window with the next adjacent one (aligning with Vim behavior)

**Vim behavior**

https://github.com/user-attachments/assets/435a8b52-5d1c-4d4b-964e-4f0f3c9aca31

https://github.com/user-attachments/assets/7aa40014-1eac-4cce-858f-516cd06d13f6

**Zed behavior**

https://github.com/user-attachments/assets/2431e860-4e11-45c6-a3f2-08f1a9b610c1

https://github.com/user-attachments/assets/30432d9d-5db1-4650-af30-232b1340229c

Note: There is a discrepancy where in Vim, if vertical and horizontal splits are mixed, swapping from a column with a single window does not work (see the vertical video), whilst in Zed it does. However, I don't see a good reason as to why this should not be supported and would argue that it makes more sense to keep the clear priority swap behavior, instead of adding a workaround to supports such cases.


mfroeh:
@cla-bot check

crates/workspace/src/workspace.rs
```diff
@@ -437,6 +437,8 @@ actions!(
         SwapPaneUp,
         /// Swaps the current pane with the one below.
         SwapPaneDown,
+        // Swaps the current pane with the first available adjacent pane (searching in order: below, above, right, left) and activates that pane.
+        SwapPaneDefault,
```
mfroeh: not really sure about this name, happy to change this to any suggestion

ConradIrwin:
Nice! Thank you. Went with `SwapPaneAdjacent` to closer match how you described the feature.

PR #42795, opened 2025-11-15
Allow external handles to be provided to gpui_tokio

143mailliw:
This PR allows for a handle to an existing Tokio runtime to be passed to gpui_tokio's initialization function, which means that Tokio runtimes created externally can be used.

Mikayla suggested that the function simply take the runtime from whatever context the initialization function is called from but I think there could reasonably be situations where that isn't the case and this shouldn't have a meaningful impact to code complexity. If you want to use the current context's runtime you can just do `gpui_tokio::init_from_handle(cx, Handle::current());`.

This doesn't have an impact on the current users of the crate - the existing `init()` function is functionally unchanged.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui_tokio/src/gpui_tokio.rs
+++ b/crates/gpui_tokio/src/gpui_tokio.rs
@@ -5,25 +5,48 @@ use util::defer;
 
 pub use tokio::task::JoinError;
 
+/// Initializes the Tokio wrapper using a new Tokio runtime with 2 worker threads.
+///
+/// If you need more threads (or access to the runtime outside of GPUI), you can create the runtime
+/// yourself and pass a Handle to `init_from_handle`.
 pub fn init(cx: &mut App) {
-    cx.set_global(GlobalTokio::new());
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        // Since we now have two executors, let's try to keep our footprint small
+        .worker_threads(2)
+        .enable_all()
+        .build()
+        .expect("Failed to initialize Tokio");
+
+    cx.set_global(GlobalTokio::new(RuntimeHolder::Owned(runtime)));
+}
+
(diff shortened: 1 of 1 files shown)


crates/gpui_tokio/src/gpui_tokio.rs
```diff
@@ -5,25 +5,48 @@ use util::defer;
...
+
+/// Initializes the Tokio wrapper using a Tokio runtime handle.
+pub fn init_from_handle(cx: &mut App, handle: tokio::runtime::Handle) {
+    cx.set_global(GlobalTokio::new(RuntimeHolder::Shared(handle)));
+}
+
+enum RuntimeHolder {
+    Owned(tokio::runtime::Runtime),
+    Shared(tokio::runtime::Handle),
+}
```
Veykril, line 31: I don't think we need to differentiate here, just turn the runtime we spawn into its handle and use that (as it should be a strong ref count anyways).

That simplifies this

143mailliw: Nope, someone already mentioned this in the Discord and the docs say that when the Runtime struct is dropped the runtime is shutdown.

> The difference between `Arc<Runtime>` and `Handle` is that an `Arc<Runtime>` will prevent the runtime from shutting down, whereas a `Handle` does not prevent that. This is because shutdown of the runtime happens when the destructor of the `Runtime` object runs.

From https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html

Veykril: They should really put that on the `Handle` docs, all it says is that its reference counted which to me would imply a strong ref count ...

Veykril:
Thanks!

PR #42799, opened 2025-11-15
Fix macros on x11 sometimes resulting in incorrect input

### LINKED ISSUE #40678: Shift key ignored in keyboard macros
### Summary

On a keyboard with macros, keypresses that are surrounded with emulated shift key are treated as if the shift key was not engaged.

### Description

I have a ZMK-based keyboard which I've been using for about 2 years. One of the layers will send symbols by quickly sending `shift on -> keypress -> shift off`. I've never seen this fail in other settings than in Zed; but on Zed the keypress is instead interpreted as the digit the symbol is on instead (most frustratingly, I get `9` instead of `(`, and `[` instead of `{`).

Steps to reproduce:
1. Use a ZMK-based keyboard
2. Configure it to use `&kp LEFT_PARENTHESIS`
3. ???
4. Find `9` in the buffer instead of the expected `(`

**Expected Behavior**:
**Actual Behavior**:

I suspect this is caused by the keypresses sent by ZMK being to close to each other in time, but it might be something else entirely.

Thanks for all your work :)

Zed: v0.208.5 (Zed) 
OS: Linux X11 ubuntu 24.04
Memory: 62 GiB
Architecture: x86_64
GPU: AMD Radeon RX 7900 XTX (RADV NAVI31) || radv || Mesa 25.0.7-0ubuntu0.24.04.1

AndreRoelofs:
Closes #40678

The python file below simulates the macros at various timings and can be run by running:
1. `sudo python3 -m pip install evdev --break-system-packages`
2. `sudo python3 zed_shift_brace_replayer.py`

Checked timings for hold=0.1 and =0.01 with the latter no longer causing incorrect inputs. --hold=0.001 still causes incorrect inputs but less frequently now. I also was unable to reproduce this extreme timing by hand.

[zed_shift_brace_replayer.py](https://github.com/user-attachments/files/23560570/zed_shift_brace_replayer.py)

Release Notes:

- linux: Greatly reduced the likelihood of key-presses combining modifier + key being processed without the modifier when typed quickly on X11


AndreRoelofs:
@cla-bot check

p-avital (approved):
I don't know the codebase much, so maybe something in here wouldn't look right to maintainers; but I'm not seeing anything that shocks me and can confirm that this solves #40678 on my machine :)

AndreRoelofs:
I decided to do extensive testing. And while it worked initially I noticed some regression in certain cases. I will release it from draft until I can't find any issues at least for 5 days.

AndreRoelofs:
> I don't know the codebase much, so maybe something in here wouldn't look right to maintainers; but I'm not seeing anything that shocks me and can confirm that this solves #40678 on my machine :)

Could you please re-check on your end? I simplified the pr in order to preserve the stability of other parts of the editor. I essentially lowered the chance of incorrect inputs. I can no longer reproduce the issue by hand.

AndreRoelofs:
The PR is for all intents and purposes ready. I will un-draft it next week if everything goes well.

AndreRoelofs:
I have been using this build locally for at least 5 days now. Incorrect input happened only once on my ZMK keyboard (glove80 glorious engrammer layous). Compared to very regular misinputs.


So this doesn't solve the problem completely for ergo keyboards on x11, but significantly lowers the chance of it happening. The first version of this PR solved the issue completely by polling for the actual mod states on XkbStateNotify events instead of taking the modifiers from the event itself. Which was a heavy handed way to fix it and did break other parts of the application. This did however completely fix the macro issues on x11 in 100% of the cases. I have since reverted the changes and opted for a lighter fix.

I do believe there is value in merging it as this does significantly improve my experience with Zed. But I think I would need help for any further changes.

probably-neb:
Hey, nice keyboard! I really enjoy my glove80, but never had the patience to try and learn the glorious engrammer layout.

Do you have any intuition for why there may still be failing inputs? Is it possible we could create another python script like you did before to reproduce the now less likely failures?

If you think we can fix 100% of cases with this newer lighter-weight approach, I think we should try to do so in this PR, otherwise, the code looks good enough and seems to work as intended so I'd be happy to merge the PR

AndreRoelofs:
Thanks for the kind words! You should definitely give glorious engrammer another try, it has come a long way and rc-7 version (beta) is now perfectly made for Vim motions!

The way that people press macros on ZMK keyboards can alter the timings of the modifiers from one press to another. Before these changes I could reproduce the incorrect input by hand more than 50% of the time by pressing the buttons a certain way. I've focused on making the timings tighter so that in the vast majority of cases it does work without making big changes to the x11 client in fear of breaking it for other Zed users. The script that I shared can still be used to reproduce issues. At --hold=0.01 it gives the right input 95%+ of the time compared to roughly 65% of the time before my changes. At --hold=0.001 it still gives incorrect input 50% of the time. Compared to basically 100% of the time before tha changes. The script can still be used to reproduce the issues at different timings so we can keep reusing it.


The thing is that I have a few ideas on how to completely solve this issue but that would require much more extensive changes to the client as I need to alter a bunch of processing to make the heavy handed approach work. I think it might make sense to remove the "Closes issue" tag and merge the code. In the next few weeks I will spend more time to make the heavy handed approach work.

The core of the heavy handed approach is to poll for depressed and latched mods in the branch of Event::XkbStateNotify(event) like
```Rust
 Event::XkbStateNotify(event) => {
                let mut state = self.0.borrow_mut();
                let old_layout = state.xkb.serialize_layout(STATE_LAYOUT_EFFECTIVE);
                let new_layout = u32::from(event.group);

                // Poll the actual depressed and latched mods from the current xkb state
                // instead of using event values which may be stale
                let depressed_mods = state.xkb.serialize_mods(STATE_MODS_DEPRESSED);
                let latched_mods = state.xkb.serialize_mods(STATE_MODS_LATCHED);

                state.xkb.update_mask(
                    depressed_mods,
                    latched_mods,
                    event.locked_mods.into(),
                    event.base_group as u32,
                    event.latched_group as u32,
                    event.locked_group.into(),
                );
...
}
```

I am iffy about this, but it solves the ZMK issue in 100% of the cases at all timings but intoduces unexpected behaviour when using Vim motions to navigate the file explorer. And could break other parts of the application. So I think we should leave it for the next PR. I don't really see a different way myself because the event here is just coming without mods sometimes.

probably-neb:
Sounds good! I look forward to the full featured PR soon. Is this PR ready to merge then?

AndreRoelofs:
Yes, I tested on both regular and ZMK as much as I could and didn't run into any issues.

p-avital:
I also gave it a go today. I think I saw a miss once, but it might gave just been me fat-fingering it. It's definitely orders of magnitude better than before though :)

AndreRoelofs:
> I also gave it a go today. I think I saw a miss once, but it might gave just been me fat-fingering it. It's definitely orders of magnitude better than before though :)

I have a fix locally that solves it in 100% of the time and doesn't introduce any issues that I can find. I will probably test for the next 2-3 weeks. I will make a new draft pr right after this is merged and send you the link. Feel free to build it yourself and test it. I am still iffy about it so it would be great if we could have more eyes on it.

p-avital:
Feel free to @ me in the new PR when you open it :)

I'm still having a hard time switching to Zed because VSCode's Peek Editors are just too central to my workflow; but I like giving it a try every now and then, and boy does it feel smoother.

AndreRoelofs:
> Feel free to @ me in the new PR when you open it :)
> 
> I'm still having a hard time switching to Zed because VSCode's Peek Editors are just too central to my workflow; but I like giving it a try every now and then, and boy does it feel smoother.

Interesting. Zed's virtual file look-up workflows are central to mine haha. Will do!

AndreRoelofs:
> Sounds good! I look forward to the full featured PR soon. Is this PR ready to merge then?

So I decided to refine the heavy handed approach. I reverted all the changes locally in this pr and started from scratch. The reason for reverting it is that my current PR does something that @smitbarmase specifically prevented a few months ago. Namely the modifiers aren't supposed to call update key as they're already handled via XkbStateNotify. This is the reason why in this PR I only update non lock modifiers and have special handling for shift lock via double press (sticky keys). 

I tried to adjust the XkbStateNotify but that produced countless issues. The keypress event was still coming in without the correct modifiers. So I decided to just poll for global state on every keypress like:
```Rust
 Event::KeyPress(event) => {
                let window = self.get_window(event.event)?;
                let mut state = self.0.borrow_mut();

                let modifiers = modifiers_from_state(event.state);
                state.modifiers = modifiers;
                state.pre_key_char_down.take();

                let locked_mods_mask = (ModMask::LOCK | ModMask::M2).bits() as u32;
                let event_state_bits = u32::from(event.state.bits());
                let depressed_mods = event_state_bits & !locked_mods_mask;
                let latched_mods = state.xkb.serialize_mods(xkbc::STATE_MODS_LATCHED);
                let locked_mods = state.xkb.serialize_mods(xkbc::STATE_MODS_LOCKED);
                let locked_layout = state.xkb.serialize_layout(xkbc::STATE_LAYOUT_LOCKED);
                state.xkb.update_mask(
                    depressed_mods,
                    latched_mods,
                    locked_mods,
                    0,
                    0,
                    locked_layout,
                );

                let keystroke = {
                    let code = event.detail.into();
..
```
I have been using it for 2 days. On both ZMK and regular keyboard. And so far I haven't run into any issue at all. The macros work in 100% of the cases at all timings, and no other issues that I have spoken about are present. It seems to work perfectly. However this does feel like a bandaid solution. But nothing else that I have tried has resulted in the key press events coming in with up to date modifiers. 

I haven't updated this branch because I waited on it being merged first to fix the biggest issues. But we can delay the merge and I can push my changes. All the issues we discussed above will be solved by it as far as I can tell. I am obviously unsure about these changes even if they do work. What are your thoughts?

probably-neb:
Nice! Why don't you push up the updated solution in a new PR, and @-mention me in it so I get a notification. I'll code review it and test it out on my end, and we can ship to nightly/preview. If it ends up causing problems, we can revert it, and go back to this solution (which we can close once the new PR is up)

probably-neb:
Going to close in favor of https://github.com/zed-industries/zed/pull/44234

PR #42803, opened 2025-11-15
http_client: Remove unused `HttpClient::type_name` method

tidely:
Closes #ISSUE

Remove unused method `HttpClient::type_name`. Looking at the PR from a year ago when it was added, it was never actually used for anything and seems like a prototyping artifact.

Other misc changes for the `http_client` crate include:

- Use `derive_more::Deref` for `HttpClientWithUrl` (already used for `HttpClientWithProxy`)
- Move `http_client::proxy()` higher up in the trait definition. (It was in between methods that have default implementations)

Release Notes:


osiewicz:
Thanks!

PR #40035, opened 2025-10-11
Theme mode switch action: switch between light and dark themes

connortsui20:
_This is my first real contribution to Zed so there are likely a lot of weird things in here..._

Closes: https://github.com/zed-industries/zed/issues/35552

I've made a few changes here and they can be mostly understood by the 4 different commits. The non-cosmetic changes are the last 2.

The third commit attempts to solve #35552, but imo my attempt is incredibly hacky and the semantics are not super obvious (which is why this PR description is so long). The fourth commit makes a change that only makes sense with the new action.

## Implementation

The change is based on https://github.com/zed-industries/zed/pull/35697, but instead of updating the settings file (with `update_settings_file::<ThemeSettings>`) I decided to use `SettingsStore::update_global` because this seems to keep changes in memory rather than forcing them onto the user's settings file (I could be wrong though?).

The semantics are pretty obvious for the `ThemeSelection::Dynamic` case, we just toggle/switch the mode based on the light/dark appearance, and if the mode is `System` we rely on the `theme.appearance().is_light()` function. However, for the `ThemeSelection::Static` case, I don't think there is such a clear solution, and my lack of understanding of how Zed's settings system works doesn't help here.

### `Static` theme switch semantics

My understanding is that the current `ThemeSelection` setting is based on [this](https://zed.dev/docs/themes#configuring-a-theme) user setting, and I believe the _only_ way to get a `Static` theme (without opening the `ThemeSelector` modal, I'll get to that later) is for a user to manually specify `"theme": "SpecificTheme"` in their settings file. This means that if a user has done that, and then they trigger this action, it is not obvious what they want to switch to.

So I chose to just hard code "One Dark" and "One Light" as fallback themes in this case. This of course assumes that they are already available in the theme registry, and I have no idea if that is always true.

### `ThemeSelector` semantics

Right now, when you choose a theme with the theme selector, it will set the `ThemeSelection` setting to `Static`, regardless of what the current selection is. This is totally fine in current Zed since that is currently the only way to change themes. With this new action, we need to maintain that `Dynamic` setting if enabled, so that is what the fourth commit does.

## Naming

I don't really know what a good name for this action should be. At first "Toggle" seems like a reasonable name, but that conflicts with the `theme_selector::Toggle` action that toggles opening the theme selector, so I don't think it should be called that. "Switch Mode" is what I went with in this PR, but I don't really like the name either because it doesn't obviously describe everything that needs to happen. I'm not even sure if `theme_selector` is the right place for this?

[...]

Issue #35552, opened 2025-08-03
Feature Request: Toggle between preferred Light/Dark Themes


connortsui20:
Most of the time, I work with a single theme in dark mode. However, there are sometimes occasions where I want to use a light mode theme (for example, working on my laptop in a bright environment). I only have two themes I use (one for light and one for dark), so I was wondering if it is possible in Zed to toggle between preferred light and dark mode themes.

On VSCode, this is [available from 1.69](https://github.com/microsoft/vscode-docs/blob/main/release-notes/v1_69.md#toggle-between-light-and-dark-themes) as `preferredDarkColorTheme` and etc. Zed does have similar settings, but AFAICT there is no easy way to toggle between the `theme.dark` and `theme.light` presets.

Is this something that is easy to implement as a command palette command?

_Originally posted by @connortsui20 in https://github.com/zed-industries/zed/discussions/35410_

zcg:
you can set it in settings.json with ,if you only need dark ，change mode to dark,only need light ,change mode to light,need with system theme, change mode to system
```json
"theme": {
		"mode": "system",
		"light": "One Light",
		"dark": "JetBrains New Dark"
},

connortsui20:
Yes I am aware of that, but as I said above there doesn't seem to be any way to easily toggle the system mode nor toggle between the two themes?

dougc84:
I was just looking to see if there was an option for this. Yes, @zcg, you can open settings, find the line, make the adjustment, and save. However, this could easily be a command palette option that would be quick to toggle.

My use case is as follows:

* As you know, on macOS, you cannot (easily) set per-app dark/light mode settings.
* At home, dark mode (and the corresponding theme) is excellent for my needs.
* However, the theme I use for dark mode gets easily washed out and is difficult to read in a very bright room (i.e. a coffee shop near a bright window). The rest of macOS is great in dark mode and I don't want to change anything except Zed.

My current solution is to manually go into the command palette and select a different theme, but a single toggle - perhaps one that could even be mapped to a keyboard shortcut - would make that so much easier and would save me minutes daily.

eduardoalba00:
I'd be interested in taking on this feature request - I can implement a command palette action that toggles between light, dark, and system theme modes by cycling through the existing theme configuration system.

arunn03:
Is this issue fixed? Or can I give it a try?

connortsui20:
Honestly not super sure why #35697 was closed...

connortsui20:
[...]

PR #42812, opened 2025-11-15
Allow dynamic `set_theme` based on `Appearance`

connortsui20:
Tracking Issue (does not close): https://github.com/zed-industries/zed/issues/35552 

This is somewhat of a blocker for https://github.com/zed-industries/zed/pull/40035 (but also the current behavior doesn't really make sense).

The current behavior of `ThemeSelectorDelegate::set_theme` (the theme selector menu) is to simply set the in-memory settings to `Static`, regardless of if it is currently `Dynamic`. The reason this doesn't matter now is that the `theme::set_theme` function that updates the user's settings file _will_ make this check, so dynamic settings stay dynamic in `settings.json`, but not in memory.

But this is also sort of strange, because `theme::set_theme` will set the setting of whatever the old appearance was to the new theme name. In other words, if I am currently on a light mode theme and I change my theme to a dark mode theme using the theme selector, the `light` field of `theme` in `settings.json` is set to a dark mode theme!

_I think this is because displaying the new theme in the theme selector does not update the global context, so `ThemeSettings::get_global(cx).theme.name(appearance).0` returns the original theme appearance, not the new one._

---

This PR makes `ThemeSelectorDelegate::set_theme` keep the current `ThemeSelection`, as well as changes the behavior of the `theme::set_theme` call to always choose the correct setting to update.

One edge case that might be slightly strange now is that if the user has specified the mode as `System`, this will now override that with the appearance of the new theme. I think this is fine, as otherwise a user might set a dark theme and nothing will change because the `ThemeAppearanceMode` is set to `light` or `system` (where `system` is also light).

I also have an `unreachable!` in there that I'm pretty sure is true but I don't really know how to formally prove that...

Release Notes:

### DIFF
--- a/crates/onboarding/src/basics_page.rs
+++ b/crates/onboarding/src/basics_page.rs
@@ -220,7 +220,7 @@ fn render_theme_section(tab_index: &mut isize, cx: &mut App) -> impl IntoElement
                 });
             } else {
                 let appearance = *SystemAppearance::global(cx);
-                theme::set_theme(settings, theme, appearance);
+                theme::set_theme(settings, theme, appearance, appearance);
             }
         });
     }
--- a/crates/theme/src/settings.rs
+++ b/crates/theme/src/settings.rs
@@ -339,7 +339,10 @@ pub fn set_theme(
 
             // Don't update the theme mode if it is set to system and the new theme has the same
             // appearance.
-            if !(mode == &ThemeAppearanceMode::System && theme_appearance == system_appearance) {
+            let should_update_mode =
+                !(mode == &ThemeAppearanceMode::System && theme_appearance == system_appearance);
+
+            if should_update_mode {
                 // Update the mode to the specified appearance (otherwise we might set the theme and
                 // nothing gets updated because the system specified the other mode appearance).
                 *mode = ThemeAppearanceMode::from(theme_appearance);
--- a/crates/theme/src/settings.rs
+++ b/crates/theme/src/settings.rs
@@ -304,31 +304,38 @@ impl IconThemeSelection {
     }
 }
 
-// impl ThemeSettingsContent {
 /// Sets the theme for the given appearance to the theme with the specified name.
+///
+/// The caller should make sure that the [`Appearance`] matches the theme associated with the name.
 pub fn set_theme(
     current: &mut SettingsContent,
     theme_name: impl Into<Arc<str>>,
-    appearance: Appearance,
+    theme_appearance: Appearance,
 ) {
-    if let Some(selection) = current.theme.theme.as_mut() {
-        let theme_to_update = match selection {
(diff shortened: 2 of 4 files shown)


ConradIrwin:
Nice, thanks!

I think the unreachable! can be reached if you edit the settings file on disk while the dialogue is open; so we can just return in that case instead of crashing.

ConradIrwin:
@connortsui20 Just checking this is still on your radar?

connortsui20:
@ConradIrwin Apologies, usually I get to these kinds of things on the weekends.

Though I also have a question of if it is worth adding functionality to check if the system setting is the same as the selected theme, and in that case we don't update "system" to "dark" for example (let me know if that doesn't make sense).

ConradIrwin:
Sorry for the early ping, I usually wait at least a week but was going too fast.

I also spent some time thinking about that; I think there's a big mismatch between what we want the picker to do and what it actually does. For now I think this is fine, and I think it's better that selecting a theme makes it "always" the theme (as you already implemented).

----
There is a bigger problem here though, which is that the picker doesn't match up with the data model in the settings file.

It seems like the primary workflow should be:
* Select which theme-pair you want (that comes with a light and dark mode). We'd need to update our theme schema to allow associating two themes together as a pair, which seems worth doing (but probably out of scope for this change).

The secondary workflow would then be:
* Select which specific theme you want to use for light mode or dark mode. We could do this using the UI we have today if we added something in the footer that lets you choose whether you wanted your choice to apply in light mode, dark mode, or always. (That could maybe be scope-creeped into this PR if you wanted, but no need). but cc @danilo-leal for ideas on making it feel good.

connortsui20:
@ConradIrwin Let me know if the change above makes sense for the scope of this PR!

danilo-leal:
Heya, just circling back to this — agree with adding a button in the theme picker footer to choose between these different options. I sort of think the default behavior of the picker should be to set the theme that will always be displayed, and have the ability to set different themes for different modes as a secondary action, where you'd need to opt into that and then be able to choose a theme for each mode. If you both agree with that and are down to set up the actions to do that, I can work on the UI from within the picker!

ConradIrwin:
@connortsui20 I am happy with either:

* (preferred) tidy up this branch to remove the unreachable, merge it, and send UI improvements as a new PR
* (also ok) close this and work more directly on the target version.

@danilo-leal thanks for the offer on making the UI feel good!

connortsui20:
@ConradIrwin I believe I have tidied things up? Let me know if there's anything else that needs cleaning up!

ConradIrwin:
I missed the pushes, sorry!

Arcitec:
Thanks! It's available in the latest Zed.

The action is named `theme: toggle mode` and switches between your configured Light/Dark themes.

PR #42834, opened 2025-11-16
vim: Fix cursor shape after deactivation

lennartkloock:
Release Notes:

- Fixed the cursor shape reset in vim mode deactivation, ensuring that the user's `cursor_shape` setting is used


lennartkloock:
@cla-bot check

dinocosta:
Thank you @lennartkloock ! 🎉 
I've added a small commit with a test for this, to hopefully prevent regressions in the future.

PR #42835, opened 2025-11-16
git: Compress diff for commit message generation

### LINKED ISSUE #34486: Generate commit message exceeding token limit
### Summary

A relatively small set of staged changes resulted in 400 Bad Request: "prompt token count of 145603 exceeds the limit of 128000".

### Description
I was able to successfully generate a commit message using Claude Code for my staged changes, but I'd just get a token limit exceeded error when attempting the same thing within Zed.

**Expected Behavior**: Create a commit message without running out of tokens (if it has to analyze with multiple context windows in the background, that's fine with me, but I should never see an error when just generating a commit message).
**Actual Behavior**: Token limit exceeded limit and no way to avoid it short of possibly reducing the number of files I'm checking in.

Zed: v0.194.3 (Zed) 
OS: macOS 26.0.0
Memory: 128 GiB
Architecture: aarch64

11happy:
This PR compresses diff capped at 20000 bytes by:
- Truncation of all lines to 256 chars
- Iteratively removing last hunks from each file until size <= 20000 bytes.

Closes #34486

Release Notes:

- Improved: Compress large diffs for commit message generation (thanks @11happy)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7112,6 +7112,7 @@ dependencies = [
  "component",
  "ctor",
  "db",
+ "diffy",
  "editor",
  "futures 0.3.31",
  "fuzzy",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7112,7 +7112,6 @@ dependencies = [
  "component",
  "ctor",
  "db",
- "diffy",
  "editor",
  "futures 0.3.31",
  "fuzzy",
--- a/crates/git_ui/Cargo.toml
+++ b/crates/git_ui/Cargo.toml
@@ -17,6 +17,7 @@ default = []
 test-support = ["multi_buffer/test-support"]
 
 [dependencies]
+diffy = "0.4.2"
 agent_settings.workspace = true
 anyhow.workspace = true
 askpass.workspace = true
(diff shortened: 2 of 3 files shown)


osyvokon:
Thanks for implementing this!

I've added some tests, and they didn't pass initially. Turns out, diffy lines already contain newlines character in them, so things like `.join("\n")` and `writeln!()` would add extra newlines, breaking the patch. I've made a fix, the tests are green now.

It's too bad that diffy doesn't support mutating hunks. This makes the code overly complex and less efficient, since we're reformatting the patch on every `calculate_size()` call. I know I was the one who pointed you to a reference implementation that uses diffy, but now I think it's not a good fit.

I think we can get rid of the dependency on diffy. Instead, let's just do simple diff string parsing. Essentially, we need two levels of parsing:
- Split by file (we already do this)
- Split by hunk headers (which always start with `@@ `)

Then we can keep a list of hunks as a `Vec[String]`. This way we don't have to reconstruct the patch, and computing lengths becomes much easier.

Are you willing to implement this change or should I go ahead? We can also pair on it, if you like.

11happy:
> Are you willing to implement this change or should I go ahead? We can also pair on it, if you like.

Yes, I am willing to implement this, I would request for a pair session if I face any difficulties.
Thank you for your time : )

11happy:
@osyvokon would appreciate your review!
Thank you : )

osyvokon:
Looks much better now, thanks! I've made a few nit renames while reviewing this. Ready to merge.

PR #42836, opened 2025-11-16
project_panel: Fix preview tabs disabling focusing files after just one click in project panel

### LINKED ISSUE #41484: Focus in non-preview mode
### Summary

TLDR: editor must not focus mouse-clicked file from the tree when preview mode is disabled.

Use case: User wants to select multiple files from the tree, they select one with mouse click want tries to do shift up/down, but surprise, it does not work.

Zed: v0.210.4 (Zed) 
OS: macOS 26.0.1
Memory: 36 GiB
Architecture: aarch64

Serophots:
Closes #41484

With preview tabs disabled, when you click once on a file in the project panel, rather than focusing on that file, zed will incorrectly focus on the text editor panel. This means if you click on a file to focus it, then follow up with a keybind like backspace to delete that file, it doesn't delete that file because the backspace goes through to the text editor instead.

Incorrect behaviour seen here:

https://github.com/user-attachments/assets/8c2dea90-bd90-4507-8ba6-344be348f151

Release Notes:

- Fixed improper UI focus behaviour in the project panel when preview tabs are disabled

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -4698,7 +4698,7 @@ impl ProjectPanel {
                     } else {
                         let preview_tabs_enabled = PreviewTabsSettings::get_global(cx).enabled;
                         let click_count = event.click_count();
-                        let focus_opened_item = !preview_tabs_enabled || click_count > 1;
+                        let focus_opened_item = click_count > 1;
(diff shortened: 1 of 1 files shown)


Serophots:
@cla-bot check

smitbarmase:
Thank you!

Issue #20121, opened 2024-11-03
inconsistent selection with block cursor


andrewbanchich:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

if i set the cursor shape to `block` and the current position is e.g.

```rust
fn |f|oo(self) {}
```

`ctrl-right` will produce

```rust
fn foo|(|self) {}
```

while `ctrl-shift-right` will produce

```rust
fn fo|o|(self) {}
```

the actual selection area for the latter is correct - cutting only removes `foo`. however, that's confusing for me because the behavior of the former command makes me think that the block cursor's current position is its left, but then when you're using it to select something the actual position is its right.

### Environment

Zed: v0.159.7 (Zed)
OS: Linux Wayland ubuntu 24.04
Memory: 30.7 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV GFX1103_R1) || radv || Mesa 24.0.9-0ubuntu0.2

### If applicable, attach your Zed.log file to this issue.

<details><summary>Zed.log</summary><pre>

</pre></details>

andrewbanchich:
also, i just noticed that holding `shift-right` will not move the cursor at all the first time.

sanbox-irl:
Yes, the simpler explanation here is that the block cursor is incorrectly placed when multiple things are selected:

<img width="324" alt="Screenshot 2024-12-29 at 3 53 33 PM" src="https://github.com/user-attachments/assets/57e569c3-3b3c-4973-967b-62bbd9893ef4" />

vs the default cursor shape:

<img width="360" alt="Screenshot 2024-12-29 at 3 53 52 PM" src="https://github.com/user-attachments/assets/9bdcbcf1-49e7-4282-8d0e-676973a592ce" />

When we do a `shift-right`, the selection DOES correctly update, but the cursor doesn't actually move in Block mode. You can verify that the selection does update by doing `shift-right` and then `backspace` -- this will delete the character as you'd expect.

sanbox-irl:
This issue should be closed in favor of #18588 which is the same issue but with more conversation on it

andrewbanchich:
@sanbox-irl this is a different issue - this issue is about the visualization of the block. the selection is correct.

the issue you linked is about deleting too many characters. also, it's been closed since that has been fixed and this has not been.

andrewbanchich:
Still happening.

andrewbanchich:
this has been fixed

Issue #38104, opened 2025-09-13
Select to end of line does not work properly with vim normal mode


mikalsande:
### Summary

Editor movements such as `editor: select to end of line` do not work properly when triggered by a keyboard shortcut when vim normal mode is active.

Found this issue while working on https://github.com/zed-industries/zed/issues/36677, so it might be related even though I was not able to reproduce it. Initially it looked like just changing the clipping behavior fixed it https://github.com/zed-industries/zed/pull/37269, but when figuring out how to write tests to reproduce it became clear that was not sufficient.

### Description

Steps to reproduce:
1. Enable vim mode in config, `"vim_mode": true`.
2. Open a new file and type `asdf`.
3. Go into normal mode (I used used esc)
4. Position cursor at `[a]sdf`.
5. Trigger select to end of line with `cmd+shift+right` (on Mac)

It works correctly when triggering `editor: select to end of line` from the command palette. From what I can tell it looks like the command palette triggers this directly in the editor while keyboard shortcuts also involve the vim code.

**Expected Behavior**:
6. The selection is `[asdf]`.

**Actual Behavior**:
6. The selection is `[asd]f`.

Test for editor_tests.rs that reproduces the issue. It simulates vim normal mode being active by setting `clip_at_line_ends = true`.

```rust
#[gpui::test]
fn test_end_of_line(cx: &mut TestAppContext) {
    init_test(cx, |_| {});

    let editor = cx.add_window(|window, cx| {
        let buffer = MultiBuffer::build_simple("asdf", cx);
        build_editor(buffer, window, cx)
    });
    _ = editor.update(cx, |editor, window, cx| {
        // pretend vim normal mode is active
        editor.set_clip_at_line_ends(true, cx);

        editor.change_selections(SelectionEffects::no_scroll(), window, cx, |s| {
            s.select_display_ranges([
                DisplayPoint::new(DisplayRow(0), 0)..DisplayPoint::new(DisplayRow(0), 0)
            ]);
        });
    });

    // Select to the end of the line.
    _ = editor.update(cx, |editor, window, cx| {
        editor.select_to_end_of_line(
            &SelectToEndOfLine {
                stop_at_soft_wraps: true,
            },
            window,
            cx,
        );
        assert_eq!(
            editor.selections.display_ranges(cx),
            &[DisplayPoint::new(DisplayRow(0), 0)..DisplayPoint::new(DisplayRow(0), 4)]
        );
    });
}
```

[...]

PR #42837, opened 2025-11-16
editor: Fix block cursor offset when selecting text

### LINKED ISSUE #36677: Block cursor position is inconsistent with highlighting
### Summary

Block cursors position (where the normal line/I-beam would be) is different when selections are highlighted.

### Description

When using the block cursor doing things without selection work as expected (the left hand side of the block cursor is treated as the normal i-beam/line position). However when highlighting or selecting multiple characters or a line of text, the right hand side of the block cursor is treated as the position. 

Steps to reproduce:
1. Type 'asdf' into the editor
2. Hit home/end.
3. Do the same with shift held down.

**Expected Behavior**:
The block cursor should act the same, regardless of selecting with motion

**Actual Behavior**:
Home/end without shift held down move to to start/end as expected. When shift is held down, end moves the block cursor 1 position shy, but still grabs all the text.

Zed: v0.200.4 (Zed) 
OS: Linux X11 pop 22.04
Memory: 31.3 GiB
Architecture: x86_64
GPU: AMD Radeon VII (RADV VEGA20) || radv || Mesa 24.2.8-1~bpo12+1pop1~1744225826~22.04~b077665

lennartkloock:
Closes #36677 and #20121

### Problem

Vim visual mode and Helix selection mode both require the cursor to be on the last character of the selection. Until now, this was implemented by offsetting the cursor one character to the left whenever a block cursor is used. (Since the visual modes use a block cursor.)
https://github.com/zed-industries/zed/blob/1683052e6ce44d206e947c397917b2ab7617c0fe/crates/editor/src/element.rs#L153
However, this oversees the problem that **some users might want to use the block cursor without being in visual mode**. Meaning that the cursor is offset by one character to the left even though Vim/Helix mode isn't even activated.

### Solution

Since the Vim mode implementation is separate from the `editor` crate the solution is not as straightforward as just checking the current vim mode. Therefore this PR introduces a new `Editor` struct field called `cursor_offset_on_selection`. This field replaces the condition above and is set to `true` whenever the Vim mode is changed to a visual mode, and `false` otherwise.

If there is a better solution without having to add a new `Editor` field, let me know.

Release Notes:

- Fixes block and hollow cursor being offset when selecting text

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1097,6 +1097,8 @@ pub struct Editor {
     pending_rename: Option<RenameState>,
     searchable: bool,
     cursor_shape: CursorShape,
+    /// Whether the cursor is offset one character to the left when something is selected (needed for vim visual mode)
+    cursor_offset_on_selection: bool,
     current_line_highlight: Option<CurrentLineHighlight>,
     pub collapse_matches: bool,
     autoindent_mode: Option<AutoindentMode>,
@@ -2217,6 +2219,7 @@ impl Editor {
             cursor_shape: EditorSettings::get_global(cx)
                 .cursor_shape
                 .unwrap_or_default(),
+            cursor_offset_on_selection: false,
             current_line_highlight: None,
             autoindent_mode: Some(AutoindentMode::EachLine),
             collapse_matches: false,
@@ -3026,6 +3029,10 @@ impl Editor {
 
     pub fn cursor_shape(&self) -> CursorShape {
         self.cursor_shape
+    }
+
+    pub fn set_cursor_offset_on_selection(&mut self, set_cursor_offset_on_selection: bool) {
+        self.cursor_offset_on_selection = set_cursor_offset_on_selection;
     }
 
     pub fn set_current_line_highlight(
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1097,7 +1097,8 @@ pub struct Editor {
     pending_rename: Option<RenameState>,
     searchable: bool,
     cursor_shape: CursorShape,
-    /// Whether the cursor is offset one character to the left when something is selected (needed for vim visual mode)
+    /// Whether the cursor is offset one character to the left when something is
+    /// selected (needed for vim visual mode)
     cursor_offset_on_selection: bool,
     current_line_highlight: Option<CurrentLineHighlight>,
     pub collapse_matches: bool,
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -131,6 +131,7 @@ impl SelectionLayout {
     fn new<T: ToPoint + ToDisplayPoint + Clone>(
         selection: Selection<T>,
         line_mode: bool,
+        cursor_offset: bool,
         cursor_shape: CursorShape,
         map: &DisplaySnapshot,
         is_newest: bool,
@@ -151,12 +152,9 @@ impl SelectionLayout {
         }
 
         // any vim visual mode (including line mode)
-        if (cursor_shape == CursorShape::Block || cursor_shape == CursorShape::Hollow)
-            && !range.is_empty()
-            && !selection.reversed
-        {
+        if cursor_offset && !range.is_empty() && !selection.reversed {
             if head.column() > 0 {
-                head = map.clip_point(DisplayPoint::new(head.row(), head.column() - 1), Bias::Left)
+                head = map.clip_point(DisplayPoint::new(head.row(), head.column() - 1), Bias::Left);
             } else if head.row().0 > 0 && head != map.max_point() {
                 head = map.clip_point(
                     DisplayPoint::new(
@@ -1425,6 +1423,7 @@ impl EditorElement {
                     let layout = SelectionLayout::new(
                         selection,
                         editor.selections.line_mode(),
+                        editor.cursor_offset_on_selection,
                         editor.cursor_shape,
                         &snapshot.display_snapshot,
                         is_newest,
@@ -1471,6 +1470,7 @@ impl EditorElement {
                     let drag_cursor_layout = SelectionLayout::new(
                         drop_cursor.clone(),
                         false,
+                        editor.cursor_offset_on_selection,
                         CursorShape::Bar,
                         &snapshot.display_snapshot,
                         false,
@@ -1534,6 +1534,7 @@ impl EditorElement {
                         .push(SelectionLayout::new(
                             selection.selection,
                             selection.line_mode,
+                            editor.cursor_offset_on_selection,
                             selection.cursor_shape,
                             &snapshot.display_snapshot,
                             false,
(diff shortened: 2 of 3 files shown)


dinocosta:
Hey @lennartkloock , thank you for opening this Pull Request 👋 

I'm wondering if I'm doing something wrong or testing the wrong cgghanges, but seems like these changes don't actually fix the reproduction steps shared in #36677 . Here's a screen recording of me comparing both the current Zed Nightly with your changes ▼

https://github.com/user-attachments/assets/5ddd02f6-27e1-43bd-9b91-d579f5b03292

In both cases I still see the issue with using `shift-cmd-right` not visually selecting all of the elements in the line. Are these changes trying to solve a different issue?

Also, not sure if it helps but we also have the `editor::vim_flavor` function that you can access from the editor, not sure if it helps but a solution where we don't need to add new fields would be preferred, as this also seems like we might just need to fix `editor::SelectToEndOfLine` and `editor::SelectToBeginningOfLine` when vim mode is enabled 🙂

lennartkloock:
> Hey @lennartkloock , thank you for opening this Pull Request 👋
> 
> I'm wondering if I'm doing something wrong or testing the wrong cgghanges, but seems like these changes don't actually fix the reproduction steps shared in #36677 . Here's a screen recording of me comparing both the current Zed Nightly with your changes ▼
> CleanShot.2025-11-25.at.12.37.06.mp4
> 
> In both cases I still see the issue with using `shift-cmd-right` not visually selecting all of the elements in the line. Are these changes trying to solve a different issue?
> 
> Also, not sure if it helps but we also have the `editor::vim_flavor` function that you can access from the editor, not sure if it helps but a solution where we don't need to add new fields would be preferred, as this also seems like we might just need to fix `editor::SelectToEndOfLine` and `editor::SelectToBeginningOfLine` when vim mode is enabled 🙂

The issue doesn't happen in Vim mode. To reproduce the problem you have to disable Vim mode and enable the block or hollow cursor. Thanks for pointing me to that function, I will look if it helps in this case.

Edit: I don't know if I'm missing something but I think both `editor::SelectToEndOfLine` and `editor::SelectToBeginningOfLine` are completely fine and work like expected, right?

Edit: nvm I just saw the other issue #38104 but seems like it's unrelated because it only happens in Vim mode

lennartkloock:
Unfortunately, the `editor::vim_flavor` function doesn't help here because we have to figure out not only if the editor is in Vim mode but also if it is in **visual** mode.

That said, what we could do is use `editor::vim_flavor` to check if it is in Vim mode and then use the `cursor_shape` field to determine if it is in visual mode but in my opinion that field shouldn't be used like that because it results in bugs like this one.

dinocosta:
> The issue doesn't happen in Vim mode. To reproduce the problem you have to disable Vim mode and enable the block or hollow cursor.

Would you be able to share reproduction steps? I'm still having a hard time reproducing the issue that this Pull Request is attempting to fix 🙂

acgxv:
Has there been any progress? I’m really in need of this fix.

dinocosta:
> Has there been any progress? I’m really in need of this fix.

@aurexav If possible, can you share a small reproduction of the issue? Wanted to confirm that these changes actually fixed the issue but [seems I was unable to reproduce the actual issue](https://github.com/zed-industries/zed/pull/42837#issuecomment-3582603538) to check if it fixed it.

acgxv:
> > Has there been any progress? I’m really in need of this fix.
> 
> @aurexav If possible, can you share a small reproduction of the issue? Wanted to confirm that these changes actually fixed the issue but [seems I was unable to reproduce the actual issue](https://github.com/zed-industries/zed/pull/42837#issuecomment-3582603538) to check if it fixed it.

1. Use block cursor
2. Open a file
3. Put cursor at the beginning of a line
4. Press `shift + end`

Zed

image

Zed Underline

image

VSCode

image

---

And you can continue press `right`. And the cursor won't move to next line. The selection is completely wrong.

image

---

The memory level selection is correct but the visual is completely wrong.

lennartkloock:
Thanks for sharing the reproduction steps, I didn't have time yet. You can also see this when selecting text using your mouse.

dinocosta:
@aurexav Thank you for sharing those steps! When you state "1. Use block cursor", do yo mean the following settings?

```json
{
  "cursor_shape": "block",
}
```

If so, I just tested these changes, comparing how it behaves against the latest stable release and it appears the behavior for `shift-end` is still the same ▼

https://github.com/user-attachments/assets/a9f0d1bf-05ca-4927-aed0-5320d0720b5d

The behavior of `editor::SelectToEndOfLine` does appear to be broken but it seems that the changes in this Pull Request don't actually fix it? Let me know if I'm missing something, thanks!

> And you can continue press `right`. And the cursor won't move to next line. The selection is completely wrong.

If you have vim mode enabled, I believe this is meant to replicate vim's behavior, where if you're at the end of the line, pressing `l` will not continue the selection on the next line.

lennartkloock:
> @aurexav Thank you for sharing those steps! When you state "1. Use block cursor", do yo mean the following settings?
> 
> ```json
> {
>   "cursor_shape": "block",
> }
> ```
> 
> If so, I just tested these changes, comparing how it behaves against the latest stable release and it appears the behavior for `shift-end` is still the same ▼
> CleanShot.2025-12-05.at.10.28.57.mp4
> 
> The behavior of `editor::SelectToEndOfLine` does appear to be broken but it seems that the changes in this Pull Request don't actually fix it? Let me know if I'm missing something, thanks!
> 
> > And you can continue press `right`. And the cursor won't move to next line. The selection is completely wrong.
> 
> If you have vim mode enabled, I believe this is meant to replicate vim's behavior, where if you're at the end of the line, pressing `l` will not continue the selection on the next line.

I don't know if there is a misunderstanding here but it seems like you are still in Vim mode in your video. The issue that is fixed here only appears if you're NOT in Vim mode. Please disable Vim mode for the reproduction steps to work. (set `"vim_mode": false` in your settings file)

acgxv:
> @aurexav Thank you for sharing those steps! When you state "1. Use block cursor", do yo mean the following settings?
> 
> ```json
> {
>   "cursor_shape": "block",
> }
> ```

Yes.

> If you have vim mode enabled, I believe this is meant to replicate vim's behavior, where if you're at the end of the line, pressing `l` will not continue the selection on the next line.

I'm a VIM hater. I never use that mode.

lennartkloock:
Here is a video I recorded demonstrating the issue. I hope that helps to further clarify it.

Settings are:
```json
{
  "vim_mode": false,
  "cursor_shape": "block",
}
```

https://github.com/user-attachments/assets/d296cb27-b719-43f4-820c-37c0931268fd

dinocosta:
Thank you for sharing @lennartkloock , was probably confusing what the real issue was but I've been able to reproduce it too! Will take a look at these changes shortly, likely during next week's Quality Week 🙂

acgxv:
Will this be included in Thursday's preview version?

dinocosta:
> Will this be included in Thursday's preview version?

I believe so. We're planning on making new releases a little bit sooner this week, seeing as next week a lot of folks are going to be out of office 🙂

PR #42852, opened 2025-11-17
gpui: Make refining a `Style` properly refine the `TextStyle`

Serophots:
## Motivating problem
The gpui API currently has this counter intuitive behaviour

```rust
 div()
            .id("hallo")
            .cursor_pointer()
            .text_color(white())
            .font_weight(FontWeight::SEMIBOLD)
            .text_size(px(20.0))
            .child("hallo")
            .active(|this| this.text_color(red()))
```
By changing the text_color when the div is active, the current behaviour is to overwrite all of the text styling rather than do a proper refinement of the existing text styling leading to this odd result:
The button being active inadvertently changes the font size.

https://github.com/user-attachments/assets/1ff51169-0d76-4ee5-bbb0-004eb9ffdf2c

## Solution
Previously refining a Style would not recursively refine the TextStyle inside of it, leading to this behaviour:
```rust
let mut style = Style::default();
style.refine(&StyleRefinement::default().text_size(px(20.0)));
style.refine(&StyleRefinement::default().font_weight(FontWeight::SEMIBOLD));

assert!(style.text_style().unwrap().font_size.is_none());
//assertion passes
```

(As best as I can tell) Style deliberately has `pub text: TextStyleRefinement` storing the `TextStyleRefinement` rather than the absolute `TextStyle` so that these refinements can be elsewhere used in cascading text styles down to element's children. But a consequence of that is that the refine macro was not properly recursively refining the `text` field as it ought to.

I've modified the refine macro so that the `#[refineable]` attribute works with `TextStyleRefinement` as well as the usual `TextStyle`. (Perhaps a little bit haphazardly by simply checking whether the name ends in Refinement - there may be a better solution there).

This PR resolves the motivating problem and triggers the assertion in the above code as you'd expect. I've compiled zed under these changes and all seems to be in order there.

Release Notes:

- N/A


mikayla-maki:
Please add a small GPUI test showing that this fixes the problem, perhaps in `styled.rs`? Or `style.rs`?

as-cii:
Hey @Serophots, thanks for this pull request! I pushed a small fix and queued this up for merging! :shipit:

PR #42854, opened 2025-11-17
Fix the handling of softlinked config and other watched files

dzervas:
I have softlinked my config to my dotfiles (it actually passes through 2 softlinks) which resulted in an error `ignoring event <path> outside of root path <another path>` an in inconsistent behavior of the config menu (sometimes I need to toggle a config multiple times or the config UI changes do not work at all)

The fix:

Canonicalize the provided path that the event processor of the worktree watcher is using to make sure that both OS events and given paths are able to match.

Release Notes:

- Fixed the handling of soft-linked watched files in the worktree (including code and config)

### DIFF
--- a/crates/worktree/src/worktree.rs
+++ b/crates/worktree/src/worktree.rs
@@ -44,7 +44,7 @@ use smallvec::{SmallVec, smallvec};
 use smol::channel::{self, Sender};
 use std::{
     any::Any,
-    borrow::Borrow as _,
+    borrow::{Borrow as _, Cow},
     cmp::Ordering,
     collections::hash_map,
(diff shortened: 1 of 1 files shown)


ConradIrwin:
@dzervas are you able to create a test that fails without this change?

I know that symlinks cause problems, but would have thought problems with config would come from the file watching in the settings crate: https://github.com/zed-industries/zed/blob/89841d034dd243733c9b0189c4c2c938d349dd67/crates/settings/src/settings_file.rs#L49 not from the worktree?

ConradIrwin:
Closing for now, but interested in figuring out the problem and solution here!

PR #42866, opened 2025-11-17
languages: Clean up invalid init calls after recent API changes

errmayank:
Related to https://github.com/zed-industries/zed/pull/41670

Release Notes:

- Cleaned up invalid init calls after recent API changes in https://github.com/zed-industries/zed/pull/42238


errmayank:
cc @osiewicz

PR #42879, opened 2025-11-17
git: Fix git panel checkbox flickering when (un)staging in quick succession

kubkon:
Release Notes:

- Fixed git panel checkbox flickering when (un)staging in quick succession


kubkon:
Superseded by #43061

PR #42889, opened 2025-11-17
Add search_on_input setting to Project Search

holoflash:
I was really missing the ability to instantly see search results while typing in the Project Search and decided to try and implement it.
As this may not be a feature that _everyone_ wants, I made it toggle-able via the settings (or the settings.json)
### Settings
Set to false by default
Screenshot 2025-11-17 at 16 17 09
### settings.json
Set to false by default
Screenshot 2025-11-17 at 16 18 21

### Video demo:
https://github.com/user-attachments/assets/715d6b77-3a61-45f8-8e1a-9bd880c697c3

- Search input is debounced with 250ms in this mode (cool?)

The desire for this feature has been expressed here too: https://github.com/zed-industries/zed/discussions/30843

Release Notes:

 - Implemented project search on input. Can be toggled with `search.search_on_input` settings, and debounced with `search.search_on_input_debounce_ms` (200ms default)


holoflash:
@cla-bot check

holoflash:
@cla-bot check

P1n3appl3:
I've been playing with this change for a little bit and I think we want to land it. Thanks for the feature and I appreciate the added test! I've got a couple of thoughts, but not all of these need to be resolved to merge the feature.

- It's hard to pick a debounce time that works for everyone, but I don't think it warrants making a setting. If anything 250ms feels a bit high for me. I wonder if it'd be possible to scale it based on worktree size (so in tiny projects you get instant results, and in huge ones you don't get thrashing...) or if that's overcomplicating things.
- There's some visual jitter when the matches refresh. I get that we want to stream matches in instead of waiting for them all to show up, but I think it might be better UX if we don't always show an empty list for at least a frame. Despite the ordering being stable, I keep losing track of a match when it's only barely moving on the screen just because it flashes.
  - I've tried opening this with larger projects such as the rust compiler, and there it's pretty annoying when you see some matches, type a few more characters, and all the existing matches disappear for a few seconds and you just see an empty pane with "searching". I'm increasingly convinced that for the updates we'd want to keep showing the previous results until we have at least one new result or we finish the search and determine there aren't any matches.
- This doesn't matter too much since the setting is opt in, but I worry that we'd really want to do some performance optimizations to avoid fully re-searching everything when the query changes. Caching logic might be too complicated to be worth it, but @osiewicz mentioned that it might be a good idea to keep track of all the files where no matches were returned. That way for the common case of adding a few characters to the query we could in theory avoid re-searching all those files. Of course in the general case (especially when regex matches get involved) this doesn't hold up, but it seems like a decent heuristic that covers the most common usage and could potentially be a huge perf win.

holoflash:
@P1n3appl3 
Thanks so much for the review!
I'm looking into some of the things you wrote and here are some thoughts so far:
- Removing the debounce entirely makes the search feel much smoother overall. Big and small projects. I have a local commit ready to be pushed with it removed if you agree.

I made it opt-in because I assume that there's a good reason for the current implementation (hitting the ENTER key), but coming from vscode, I'm used to the results showing instantly, **and** being able to search right in the file browser.
Could you fill me in on why the current approach was chosen in the first place in ZED? Would help me understand what else needs to be done to make this feature work nicer.

As for the other points. I'm not really sure what the best approach might be. Maybe if something is inputed in the search field, continue the search but ONLY on the results coming from the previous query? (would need to temporarily cache the results?)

- user types: "express" --> get search results for "express"
- user types: "expression" ---> continue search but only on results from "express"
- user types: "expression a" ---> continue search but only on results from "expression" 
- user erases some letters: "expressi" ---> continue search but on results from "express"
- user erases more letter:" expre" ---> do a new search because query is broader than initial search.

This might need a little debounce though to not trigger on each letter if typed quickly.
Could be done in reverse as you mentioned by excluding results instead.
Just high level thoughts, I don't know how to go about implementing this as I'm writing.

I'll just push the version without the debounce so you can get a feel for it.
Maybe that alone might make this work without all the extra stuff.

Be-ing:
I could imagine wanting this enabled sometimes (small projects) but not others (large projects with rust-analyzer regularly stressing my system). I wonder if it might be better to make this toggleable with a button next to the search bar by the filter and case sensitivity buttons instead of buried in the settings. That would also make it more discoverable.

holoflash:
@Be-ing 
I personally would like that too! Just not sure if it's too intrusive of a feature.
But here's what that could look like:

https://github.com/user-attachments/assets/2b726f20-6fd4-42bd-87ab-40f69fabea0f

I just pushed the above as a suggestion.

P1n3appl3:
It's worth noting that the "quick search" feature in #44530 implements this functionality. I think it may still be worth doing here, but we should at least use the same logic between the impls.

holoflash:
> It's worth noting that the "quick search" feature in #44530 implements this functionality. I think it may still be worth doing here, but we should at least use the same logic between the impls.

Neat @P1n3appl3
I'll see how that one progresses and if I can use the new quick_search for this instead.
Coincidentally ended up binding ⌥+cmd+q to toggling this feature, so even that part makes sense 🤩

assets/keymaps/default-linux.json
```diff
@@ -403,6 +403,7 @@
       "ctrl-shift-h": "search::ToggleReplace",
       "alt-ctrl-g": "search::ToggleRegex",
       "alt-ctrl-x": "search::ToggleRegex",
+      "alt-ctrl-q": "search::ToggleSearchOnInput",
```
SomeoneToIgnore: 1. This key combination is rather nuclear, given that `ctrl/cmd-q` quits Zed.
2. This functionality, live toggle search on input — is it represented by any other editors?

Overall, feels like an unnecessary compilation in all this.

assets/settings/default.json
```diff
@@ -644,6 +644,8 @@
     "regex": false,
     // Whether to center the cursor on each search match when navigating.
     "center_on_match": false,
+    // Show search results on input by default.
```
SomeoneToIgnore: all-settings.md miss this new entry, let's add it there.

Also, let's make it `true` by default?

crates/editor/src/editor_settings.rs
```diff
@@ -172,6 +172,8 @@ pub struct SearchSettings {
     pub regex: bool,
     /// Whether to center the cursor on each search match when navigating.
     pub center_on_match: bool,
+    /// Show search results on input by default.
```
SomeoneToIgnore:
```suggestion
    /// Show search results on input.
```

crates/search/src/buffer_search.rs
```diff
@@ -3112,9 +3115,96 @@ mod tests {
...
                     });
                 });
             });
         });
     }
+    #[gpui::test]
+    async fn test_search_on_input_setting(cx: &mut TestAppContext) {
+        let (editor, search_bar, cx) = init_test(cx);
+
+        // Test with search_on_input disabled
```
SomeoneToIgnore: Can we remove the comments that verbatim describe the code just below them?
Here and in other places.

SomeoneToIgnore (commented):
Thank you.

I think we can drastically improve it if we remove the live toggle: do you have any other editors' example where it present?

assets/keymaps/default-linux.json
holoflash: My first iteration didn't have the live-toggle.
I implemented it as suggested by @Be-ing above, but not something I've see anywhere else.

I'm all for removing it if it's not in demand.

SomeoneToIgnore: Given how much code will be removed and that it's the only UI element that's changed, I'm 100% in for a removal + the new default flip.

holoflash: But keep the button for it in the search toolbar?
I went with the "eye" icon there. Is there a better fit? (if we're keeping this)

Screenshot 2026-02-12 at 17 36 55

Or should it just be something that's togglable via the Settings?

SomeoneToIgnore: But no, neither we need this, no UI changes at all?

holoflash: Oh ok, nice! But keep it an option that can be toggled on/off in the Settings then?

SomeoneToIgnore: Yes.

holoflash: Just pushed the suggested changes.

The only UI thing I left is the text on the search screen.
Show different text depending on the user's settings.

```rust
  Label::new(if EditorSettings::get_global(cx).search.search_on_input {
                    "Start typing to search. For more options:"
                } else {
                    "Hit enter to search. For more options:"
                })
```

holoflash: I see that this PR is still open:
https://github.com/zed-industries/zed/pull/44530

Depending on which gets merged first, it might be nice to revisit this and use the quick search for search_on_input.

docs/src/reference/all-settings.md
```diff
@@ -3282,12 +3288,6 @@ Non-negative `integer` values
 - Setting: `search_wrap`
 - Default: `true`
 
-## Center on Match
```
holoflash, line 3285: I noticed that there was a duplicate here so I removed it.
And I'm not quite sure what is going here with the mixed ## ###
 
@katie-z-geer ?

docs/src/reference/all-settings.md
```diff
@@ -3282,12 +3288,6 @@ Non-negative `integer` values
...
-
-- Description: If `center_on_match` is enabled, the editor will center the cursor on the current match when searching.
```
SomeoneToIgnore, line 3287: ??

holoflash: https://github.com/zed-industries/zed/pull/42889#discussion_r2800336711

holoflash: Btw, I set it to default false for now.
There's some tests failing that don't expect it to be on that I need to update and ran out of time for the day.

SomeoneToIgnore: Ah, odd drive by changes...
Ok, sure, as long as we finally make the toggle to have a proper default: https://github.com/zed-industries/zed/pull/42889#discussion_r2799730645

holoflash: It was riiiight there where I was about to add my stuff and figured it was my duty lol

assets/settings/default.json
holoflash: Toggled to on by default now and set it to off before the tests that did not expect that.

SomeoneToIgnore (approved):
Thanks!

One small caveat left is the settings' documentation wording: that only toggles search on input for project search, whether the search settings it's placed into are both buffer and project searches.

But I think this is a good start already.

SomeoneToIgnore:
Sorry for the bad news: turns out we'll have to revert the whole shebang of 3 related PRs for now as the feedback was "remove the flickering before having it released, even disabled"-ish.

I plan to push it forward and redo the way we deal with this: the idea is to reuse search properties (it should stream files in the same order between reruns) and replace excerpts gradually, not clearing the entire thing each time, as we do now, most probably next week.

PR #42905, opened 2025-11-17
gpui: Custom fragment shaders

### LINKED ISSUE #43273: gpui: Add a mechanism to render one-off shaders
### Summary

Like a `<canvas>` web element

someone13574:
This PR implements custom fragment shaders for gpui, written in wgsl. It lets the user implement shaders which are compiled at runtime (after converting to msl or hlsl on mac or windows, using naga). Additionally, users can optionally pass per-instance data which lets you implement properties and custom primitives (as seen with `Star` in the example), or animate the shader. Instances of the same shader will be batched, provided that they are pushed to the scene consecutively. You can see the current api in the example I added.

To ensure that the instance-data structs are properly following the [WGSL layout rules](https://sotrh.github.io/learn-wgpu/showcase/alignment/#alignment-of-uniform-and-storage-buffers), I've made a derive macro which ensures the alignment for all fields are correct (this also restricts the types you can pass to be fairly simple, such as vectors and scalars, though you can also implement the trait yourself to pass more complex types if needed).

Example of three shaders:
- The background is an example of a more complex effect
- The two stars are separate instances of a star SDF which lets you control color, border, and size (as an example of how you can make custom primitives).
  - This will probably be the main purpose of this, since it is much more efficient than paths 
- The third shader is an example of a shader without any instance data.

Screenshot From 2025-11-20 14-00-53

TODO:

- [x] Test without user-data (does wgsl allow ZSTs? otherwise filter)
  - [x] ZSTs aren't allowed. Need to change change interface to make user-data toggleable 
- [x] Should we create types equivalent to `Vec2` and `Vec4` which match wgsl alignment so that manual padding isn't required?
  - Made a new trait and derive macro for enforcing padding rules for common types. This also means you don't need to define the layout on both the rust side and the wgsl side
- [x] Non-terrible documentation
- [x] Do *something* about fractional scaling
- [x] Don't panic on compile error
- [x] Fix alpha blending
- [x] Implement metal support
- [x] Test whether blade backend works on macOS
  
If you have any ideas for improving the API, let me know!

Closes #43273 probably

Release Notes:

- N/A


crates/gpui/examples/shader.rs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            fn noise2(n: vec2<f32>) -> f32 {
+                let d = vec2<f32>(0., 1.);
+                let b = floor(n);
+                let f = smoothstep(vec2<f32>(0.), vec2<f32>(1.), fract(n));
+                return mix(mix(rand22(b), rand22(b + d.yx), f.x), mix(rand22(b + d.xy), rand22(b + d.yy), f.x), f.y);
+            }
+        ")
+        .with_item("
```
someone13574, line 51: There is no technical reason why you would need multiple `with_item` calls here other than visual separation and that it is sometimes useful to define helper functions as string constants and then just do `.use_item(MY_HELPER_FN)`.

crates/gpui/src/elements/shader.rs
```diff
@@ -0,0 +1,351 @@
...
+            )
+        } else {
+            (String::new(), String::new(), "")
+        };
+        let instance_data_definition = T::DEFINITION.unwrap_or("");
+
+        let source = format!(
+            r#"
+            struct GlobalParams {{
+                viewport_size: vec2<f32>,
```
someone13574: I don't really like where I currently do the formatting for the shader source code. I feel like it should either be on `FragmentShader` itself, or somewhere within `paint_shader` so that `paint_shader` could be useful by itself. As an aside, I also don't really like the name `FragmentShader`.

mikayla-maki:
Wow, this PR is so cool! I love that you where able to re-use the GPUI animation stuff to pass time data to the star instances.

crates/gpui/src/window.rs
```diff
@@ -3266,6 +3268,56 @@ impl Window {
...
 
+    /// Paint a custom shader into the scene for the next frame at the current z-index.
+    ///
+    /// Refer to the [crate::FragmentShader] documentation for information on what the main body
+    /// and extra items can contain. If this function fails to compile, it returns a
+    /// String containing the compilation error message and a boolean indicating whether
+    /// this was the first attempt to paint this shader.
+    ///
+    /// This method should only be called as part of the paint phase of element drawing.
+    pub fn paint_shader<T: ShaderUniform>(
```
someone13574, line 3279: Should I make it so that `CustomShaderId` is used to specify the shader instead of the source code (and then have a seperate `Window::register_shader` which is used to obtain the id)? 

It would allow people who are using this api directly to avoid the expensive string hashmap lookup which is used in the internal `register_shader` (which currently needs to be done for each instance, each-frame). It would still need to check for people going through the `ShaderElement` api, but if someone had a massive number of instances, they could just keep track of the shader id themselves and avoid all that.

Most of the infrastructure is already there, but it would require people using the `paint_shader` api to make a register call and a paint call.

mikayla-maki:
Hi,

Quinntyx:
@someone13574 It would be incredible if you could move this PR over to gpui-ce, where it can hopefully get merged, as this looks to be an incredible piece of engineering that I think a lot of people could make good use of.

ppmpreetham:
all the changes were asked just to not merge.
ts so tuff 🥀🥀🥀

XhstormR:
https://github.com/gpui-ce/gpui-ce

PR #42946, opened 2025-11-18
Increase askpass timeout for git operations

### LINKED ISSUE #29903: Git: Push fails with a timeout error when hooks are enabled.
### Summary

Push fails with a timeout error when hooks are enabled.

### Description
Steps to trigger the problem:
1. Create a `pre-push` hook (in my case, it is a linter check)
2. Try to push from the Git panel in Zed
3. Push will fail with a timeout error

Removing hooks solves the problem.

Actual Behavior: Pushing with hooks fails because of the timeout.
Expected Behavior: Hooks are accounted for and running them would not result in the failed push.

Zed: v0.184.10 (Zed) 
OS: macOS 15.4.1
Memory: 16 GiB
Architecture: aarch64

11happy:
Closes #29903

Release Notes:

- Increased timeout from 17->300 & improved the error message.

### DIFF
--- a/crates/askpass/src/askpass.rs
+++ b/crates/askpass/src/askpass.rs
@@ -146,7 +146,7 @@ impl AskPassSession {
     // drop order this takes an &mut, so you can `drop()` it after you're done with the master process.
     pub async fn run(&mut self) -> AskPassResult {
         // This is the default timeout setting used by VSCode.
-        let connection_timeout = Duration::from_secs(300);
+        let connection_timeout = Duration::from_secs(17);
         let askpass_opened_rx = self.askpass_opened_rx.take().expect("Only call run once");
         let askpass_kill_master_rx = self
             .askpass_kill_master_rx
--- a/crates/askpass/src/askpass.rs
+++ b/crates/askpass/src/askpass.rs
@@ -146,7 +146,7 @@ impl AskPassSession {
     // drop order this takes an &mut, so you can `drop()` it after you're done with the master process.
     pub async fn run(&mut self) -> AskPassResult {
         // This is the default timeout setting used by VSCode.
-        let connection_timeout = Duration::from_secs(17);
+        let connection_timeout = Duration::from_secs(300);
         let askpass_opened_rx = self.askpass_opened_rx.take().expect("Only call run once");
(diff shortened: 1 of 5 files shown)


cole-miller:
Hi @11happy and thanks!

I think a better approach here might be to run git hooks in a separate step without the askpass delegate, so that they don't eat into the timeout if they run for a while. This is how I would approach it, just talking about the `pre-push` hook for now:

- `pre-push` is supposed to run after the local remote-tracking ref (e.g. `origin/main`) has been updated, but before we've actually done the network operation ([ref](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)). So Zed would need to update `origin/main` before it runs `.git/hooks/pre-push`
- `pre-push` receives information about the remote being pushed to in its command-line arguments, and some additional information on stdin about the refs that are being pushed (see the same page in the git docs). So Zed would need to get that information and pass it in.
- If `pre-push` is successful, we should go on to do `git push --no-verify`, but we'll want to put the remote-tracking ref back to its initial state if `pre-push` or `git push --no-verify` fails (I believe this is how git itself behaves, but it would be good to check).

kubkon:
FWIW @cole-miller and I have already implemented most of the machinery in https://github.com/zed-industries/zed/pull/43285. All that's now left is to add `PrePush` variant to `RunHook`, register it correctly in `git.proto`, and make sure we run `run_hook` before we execute the `push` job in `GitStore`. Happy to answer any questions you might have about this!

11happy:
> FWIW @cole-miller and I have already implemented most of the machinery in #43285. All that's now left is to add `PrePush` variant to `RunHook`, register it correctly in `git.proto`, and make sure we run `run_hook` before we execute the `push` job in `GitStore`. Happy to answer any questions you might have about this!

Thank you for these pointers & work on `RunHook` implementation I have implemented it accordingly.
Thank you : )

11happy:
corrected the error message!
I think its good to go now , @dvdsk 
Thank you : )

yara-blue:
:partying_face: Thanks! Curious to see what you'll do next!

esthertrapadoux:
Congrats on another great contribution @11happy !!

PR #42949, opened 2025-11-18
project: Fix out of bounds index in `invalidate_for_chunk`

Veykril:
Fixes ZED-31F

Release Notes:

- Fixed a panic in inlay hints computation when deleting lines


Veykril:
Proper fix is coming up by Kirill

PR #42953, opened 2025-11-18
terminal_view: Fix terminal opening in root directory when editing single file corktree

### LINKED ISSUE #42945: Terminal opens in / when editing a single file
### Summary

If I open zed in a directory, e.g., "zed ~/Desktop", then terminal panels open in that directory, as per the settings.

If I open an individual file in the directory, e.g., "zed hello.txt", then terminal panels open in the root directory /, which is (a) useless and (b) inconsistent with the behavior above.

I think a more sensible behavior would be that, when zed is used to open an individual file, it should consider that file's path as "the project directory", and open terminal panels there.

### Description

Steps to reproduce (verified on MacOS, arm64, Zed 0.212.6):
1. create a file, e.g., `echo hello > ~/Desktop/hello.txt`
2. open that file in zed `zed ~/Desktop/hello.txt`
3. open a terminal panel in zed

**Expected Behavior**:

The terminal opens in ~/Desktop

**Actual Behavior**:

The terminal opens in /

Zed: v0.212.6 (Zed) 
OS: macOS 26.1
Memory: 16 GiB
Architecture: aarch64

```log

```

</details>

canack:
Fixes #42945

## Problem
When opening a single file via command line (e.g., `zed ~/Downloads/file.txt`), the terminal panel was opening in the root directory (/) instead of the file's directory.

## Root Cause
The code only checked for active project directory, which returns None when a single file is opened. Additionally, file worktrees weren't handling parent directory lookup.

## Solution
Added fallback logic to use the first project directory when there's no active entry, and made file worktrees return their parent directory instead of None.

## Testing
- All existing tests pass
- Added test coverage for file worktree scenarios
- Manually tested with `zed ~/Downloads/file.txt` - terminal now opens in correct directory

This improves the user experience for users who frequently open single files from the command line.

## Release Notes

- Fixed terminal opening in root directory when editing single files from the command line

### DIFF
--- a/crates/terminal_view/src/terminal_view.rs
+++ b/crates/terminal_view/src/terminal_view.rs
@@ -1556,7 +1556,8 @@ pub(crate) fn default_working_directory(workspace: &Workspace, cx: &App) -> Opti
             .read(cx)
             .active_project_directory(cx)
             .as_deref()
-            .map(Path::to_path_buf),
+            .map(Path::to_path_buf)
+            .or_else(|| first_project_directory(workspace, cx)),
         WorkingDirectory::FirstProjectDirectory => first_project_directory(workspace, cx),
         WorkingDirectory::AlwaysHome => None,
         WorkingDirectory::Always { directory } => {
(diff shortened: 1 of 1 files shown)


canack:
@cla-bot check

Veykril (approved):
Thanks!

PR #42959, opened 2025-11-18
Add `CreatePullRequest` action

tommyming:
Closes #42217 

Release Notes:
- Added `git: create pull request` command to open a pull request for the current branch

For more details, please refer to the issue, thank you.


esthertrapadoux:
@tommyming just following up to let you know the team went pretty deep in more git feature work and we got behind on PR reviews. We're ramping back up and appreciate the patience! 👍

tommyming:
@esthertrapadoux Thanks for the update!
Glad that this PR got into Quality Week!
Please take your time and keep up the good work!

crates/git_ui/src/git_ui.rs
```diff
@@ -231,6 +240,77 @@ pub fn init(cx: &mut App) {
     .detach();
 }
 
+fn open_create_pull_request(
+    workspace: &mut Workspace,
+    _window: &mut Window,
```
dinocosta: Looks like we can actually remove this argument 🧹

crates/git_ui/src/git_ui.rs
```diff
@@ -77,6 +79,13 @@ pub fn init(cx: &mut App) {
             return;
         }
         if !project.is_via_collab() {
+            workspace.register_action(
+                |workspace, _: &zed_actions::git::CreatePullRequest, window, cx| {
+                    if let Err(error) = open_create_pull_request(workspace, window, cx) {
+                        workspace.show_error(&error, cx);
```
dinocosta: I was testing this and noticed that, when `open_create_pull_request` returns an error, we're showing the bottom right notification, due to `Workspace.show_error` being used. Seeing as all the other git feature implementations appear to be using `GitPanel.show_error_toast`, I believe it would make sense to do the same here, for consistency.

Using the `show_error_toast` method instead would end up looking like ▼

Image

crates/git_ui/src/git_ui.rs
```diff
@@ -231,6 +240,77 @@ pub fn init(cx: &mut App) {
...
+            let stripped = ref_name.strip_prefix("refs/remotes/").unwrap_or(ref_name);
+            let branch_name = stripped
+                .split_once('/')
+                .map(|(_, name)| name)
+                .unwrap_or(stripped);
+            Some(branch_name.to_string())
+        } else {
+            None
+        }
+    });
```
dinocosta: I'm not sure if I'm reading this wrong but it appears there's some confusion here. I believe we're deriving the target branch name from the local branch name but, looking at the code, it appears that `target_branch` is simply being set to the upstream branch, if there's one being tracked for the current branch.

I discussed this a little bit with someone else from the Git team and we believe that, for a first iteration of this feature, we simply need to know what the branch for which we're creating the Pull Request is, seeing as it appears that information is optional for both GitHub and GitLab, as the target branch will be automatically determined by those platforms when opening the Pull Request. With that said, we should probably refactor the approach a bit with:

1. Removing `BuildCreatePullRequestParams` 
2. Updating `GitHostingProvider.build_create_pull_request_url` to simply take a `remote_source_branch` argument, instead of the `BuildCreatePullRequestParams` one
3. Update `git_ui::open_create_pull_request` to provide the upstream branch name, in case it exists, or the local branch name

The current implementation doesn't work if, let's say, I have a local branch named `feature/git-pr` that's tracking an upstream branch named `feature/git-pull-request`. When that's the case, the `git: create pull request` command will use `feature/git-pr`, which doesn't exist in the remote, instead of `feature/git-pull-request` for the source branch name. 

Let me know what you think! 🙂

dinocosta (changes requested):
Hey @tommyming  , thank you for opening this. This is a helpful feature, and the test inclusion is appreciated! ✌️

Overall the changes look good, but there are a couple of places where we can improve the implementation and a small edge case we’re not handling correctly. Let me know if you have any questions.

crates/git_ui/src/git_ui.rs
tommyming: will remove that in the next commit! Thanks for pointing out 🙂

crates/git_ui/src/git_ui.rs
tommyming: my bad on missing this, will add back the failed case alert support!

crates/git_ui/src/git_ui.rs
tommyming: Yeah, I missed the case if there is a branch name mismatch (no matter local or remote). You made a good point on this, will update that!

(Sidenote)
A question comes up in my mind: How to determine the target branch of the PR from Zed?
As I checked before, the current target branch is main/master. (correct me if wrong)

I am thinking whether we should do it relatively, aka the parent of the current branch?

tommyming:
@dinocosta Thanks a lot for the review!

crates/git_ui/src/git_ui.rs
dinocosta: > (Sidenote)
> A question comes up in my mind: How to determine the target branch of the PR from Zed?
> As I checked before, the current target branch is main/master. (correct me if wrong)

Honestly, I believe we don’t need to determine the target branch. Both GitHub and GitLab, the providers which are being updated in this Pull Request, automatically set it. For example, if a `fix-cursor-offset` branch is pushed to the remote, visiting `https://github.com/zed-industries/zed/pull/new/fix-cursor-offset` will open a new pull request with the base set to `main` by default.

> I am thinking whether we should do it relatively, aka the parent of the current branch?

I suspect git does not track or expose a “parent branch,” so I’m unsure we could rely on that.

I’d remove ‎`target_branch` and rely on whether the local branch has an upstream. If it does, use the upstream branch name, otherwise, use the local branch name. 🙂

tommyming: I agree that referring to the upstream branch name and fallback for the local branch name is a more generic approach.

I am currently testing the updated code locally, will commit once it is ready, thanks!

crates/git_ui/src/git_panel.rs
```diff
@@ -2346,6 +2346,72 @@ impl GitPanel {
         .detach_and_log_err(cx);
     }
 
+    pub fn create_pull_request(&self, cx: &mut Context<Self>) {
+        let Some(repo) = self.active_repository.as_ref() else {
+            return;
+        };
```
dinocosta: @tommyming Minor request here, but I guess we could also move this into the closure below and display an error if someone attempts to run this without an active repository, just like we're doing for the active branch 👍

crates/git_ui/src/git_panel.rs
```diff
@@ -2346,6 +2346,72 @@ impl GitPanel {
...
+            let source_branch = branch
+                .upstream
+                .as_ref()
+                .and_then(|upstream| {
+                    if let UpstreamTracking::Tracked(_) = upstream.tracking {
+                        upstream.branch_name().map(|name| name.to_string())
+                    } else {
+                        None
+                    }
+                })
```
dinocosta: Could potentially simplify this a bit, but non-blocking 🙂

```rust
.filter(|upstream| matches!(upstream.tracking, UpstreamTracking::Tracked(_)))
.and_then(|upstream| upstream.branch_name())
.map(|name| name.to_string())
```

crates/git_ui/src/git_panel.rs
```diff
@@ -2346,6 +2346,72 @@ impl GitPanel {
...
+                .and_then(|upstream| {
+                    let remote_name = upstream.remote_name();
+                    match remote_name {
+                        Some("upstream") => remote_upstream.as_deref(),
+                        Some(_) => remote_origin.as_deref(),
+                        None => None,
+                    }
+                })
+                .or_else(|| remote_origin.as_deref())
+                .or_else(|| remote_upstream.as_deref())
```
dinocosta: Same thing here, non-blocking but could potentially be simplified up a bit 🙂

```rust
.and_then(|upstream| match upstream.remote_name() {
    Some("upstream") => remote_upstream.as_deref(),
    Some(_) => remote_origin.as_deref(),
    None => None,
})
.or(remote_origin.as_deref())
.or(remote_upstream.as_deref())
```

crates/git_ui/src/git_panel.rs
```diff
@@ -2346,6 +2346,72 @@ impl GitPanel {
...
+                .ok_or_else(|| anyhow::anyhow!("No remote configured for repository"))?;
+            let remote_url = remote_url.to_string();
+
+            let provider_registry = git::GitHostingProviderRegistry::global(cx);
```
dinocosta: Since `GitHostingProviderRegistry` is already imported we can remove the `git::` bit 🧹

dinocosta (approved):
Thank you for taking the time to tackle the feedback @tommyming ! I've just taken another look at the changes and it's looking good to me!

I've still left some comments on optional changes you could make, so let me know if you want to tackle those, otherwise I can go ahead and do that myself and get this merged 🚢

P.S.: You might want to merge `main` into this branch, as we've recently done some data migration so `cargo run` will fail in this branch in case you've got and older schema in your development database.

crates/git_ui/src/git_panel.rs
tommyming: thanks for the heads up!
I am still getting used to Rust's chained calls, originally having thoughts about Swift's `compactMap` but I cant find something similar to that in Rust lol

tommyming:
> I've still left some comments on optional changes you could make, so let me know if you want to tackle those, otherwise I can go ahead and do that myself and get this merged 🚢
> 
> P.S.: You might want to merge `main` into this branch, as we've recently done some data migration so `cargo run` will fail in this branch in case you've got and older schema in your development database.

Thanks for the code enhancement review @dinocosta! 
I will make the enhancement within today, and rebase the main branch to here.

Merry Christmas!

dinocosta (approved):
Thank you for tackling the proposed changes @tommyming ! 

While testing these new changes, I noticed that there was a panic when `git: create pull request` failed and the error toast should be shown. I ended up pushing https://github.com/zed-industries/zed/pull/42959/commits/6873f2f3374724469c078f773783ca5bb1e260dd to handle this and removed an unused import. 

Will merge after CI/CD passes! ✌️

Issue #38799, opened 2025-09-24
Poor search performance in large repositories


bjackman:
### Summary

Poor search performance in large repositories

### Description

Using the [linux kernel repo](https://github.com/torvalds/linux) as an example. This is not particularly extreme but it's big enough that a dumb grep is unusably slow:

```
❯❯  time git grep vmx_l1d_should_flush
[...]

________________________________________________________
Executed in  826.73 millis    fish           external
   usr time    0.76 secs    985.00 micros    0.76 secs
   sys time    3.83 secs      0.00 micros    3.83 secs

❯❯  time grep -R vmx_l1d_should_flush .
...

________________________________________________________
Executed in   49.77 secs    fish           external
   usr time    8.49 secs  986.00 micros    8.49 secs
   sys time    9.32 secs   88.00 micros    9.32 secs
```

Timing Zed's project search for the string, (using my stopwatch on my phone), I find when I am connected to the server above via SSH (i.e. using Zed's remoting feature), the same search takes about 4s.

Doing the same search again locally on my laptop takes `559.40 millis` with Git and about 3s with Zed.

I noticed when I first open the project in Zed, there's a Git process that hogs the CPU for a litle while. Combined with the fact that it's still much faster than a dumb `grep`, it looks like there's _some_ optimisation working here, but it's just missing some trick or other.

With my stopwatch on my phone I measured about 4s to grep this repo in Zed. This is when remoting from a laptop with an `Intel(R) Core(TM) Ultra 7 165U` (Meteor Lake, 2023) into a high-spec GCP VM (some sort of Cascade Lake I think).

Steps to reproduce:
1.
2.
3.
4.

**Expected Behavior**: `B L A Z I N G    F A S T    R U S  T     C O D E`
**Actual Behavior**: kinda slowwwww
:

<details><summary>click to expand</summary>

settings:

```json
{
  "ssh_connections": [ <REDACTED> ],
  "base_keymap": "VSCode",
  "vim_mode": true,
  "file_types": {
    "makefile": ["*.makefile"]
  },
  "show_whitespaces": "boundary",
  "soft_wrap": "editor_width",
  "use_smartcase_search": true,
  "cursor_blink": false,
  "auto_update": false,
  "telemetry": {
    "metrics": false
  },
  "wrap_guides": [80],
  "languages": {
    "Nix": {
      "show_whitespaces": "none",
      "tab_size": 2
    }
  }
}

```

Log:

```
[...]

PR #42968, opened 2025-11-18
project search: Skip loading of gitignored paths when their descendants will never match an inclusion/exclusion query

osiewicz:
Co-authored-by: dino <dinojoaocosta@gmail.com>

Related-to: #38799

Release Notes:

- Improved project search performance with "Also search files ignored by configuration" combined with file inclusion/exclusion queries.

### DIFF
--- a/crates/project/src/project_search.rs
+++ b/crates/project/src/project_search.rs
@@ -809,13 +809,16 @@ impl PathInclusionMatcher {
     fn new(query: Arc<SearchQuery>) -> Self {
         let mut included = BTreeSet::new();
         let mut excluded = BTreeSet::new();
+        // To do an inverse glob match, we split each glob into it's prefix and the glob part.
+        // For example, `src/**/*.rs` becomes `src/` and `**/*.rs`. The glob part gets dropped.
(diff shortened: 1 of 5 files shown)


crates/project/src/project_search.rs
```diff
@@ -207,11 +209,12 @@ impl Search {
                         let (sorted_search_results_tx, sorted_search_results_rx) = unbounded();
 
                         let (input_paths_tx, input_paths_rx) = unbounded();
-
+                        // glob: src/rust/
+                        // path: src/
```
osiewicz:
```suggestion
```

smitbarmase (approved):

osiewicz:
Our tests found an issue with a scenario where we have:
- Enabled ignored dir scans
- Exclusion pattern is **/*.ts
- Inclusion pattern is node_modules/prettier/**

Exclusion pattern will prune files before we even get to loading the `node_modules`.

crates/project/src/project_tests.rs
osiewicz, line 1: IMHO these tests could reside in `project_search.rs`.

PR #6924, opened 2024-01-28
Go to reference when there's only one

wasd96040501:
close #4796
Similar to implementation in `navigate_to_definitions`
Because find all references  in rust-analyzer will return locations containing location we passed in, so we need remove it.
And I'm not sure is it the correct way to do this(line 7299~7307)

PR #43026, opened 2025-11-19
editor: Goto references skip multibuffer if single match

cameron1024:
If you find references for the symbol under the cursor, and there is only one
match (excluding the one under the cursor), then it now jumps directly to
that reference, rather than opening a multibuffer with that reference.

Co-authored-by: Agus <agus@zed.dev>

Release Notes:


maxdeviant:
FYI, someone previously attempted a change like this and reverted it because the behavior was inconsistent: https://github.com/zed-industries/zed/pull/6924#issuecomment-2033076300

I'm personally against having the behavior vary based on the number of references, as I think it's more confusing. At the very least, I think this needs to be a setting.

cameron1024:
Interesting - I guess the current behaviour feels weird to me since it's not how goto definition works (and these two features are pretty similar in my mind). I can get behind a setting though

SomeoneToIgnore:
And, of course, the tests are great to have.

PR #43037, opened 2025-11-19
editor: Go to test

torepettersen:
Closes https://github.com/zed-industries/zed/discussions/40859

I am an earlier Neovim user, and I have gotten used to having a shortcut for jumping from the source file to the test file and back again. Then I have a single shortcut to jump between the two, so I don't have to search. For me this is super helpful in big Elixir projects, since the source and test files are in completely different directories, but always relative to each other. I also frequently jump between source and test files.

I don't know how many languages that have this kind of pattern, but I know there is at least something similar in Ruby, but they have two different patterns depending on which test tool they are using.

Seems like the person that opened the issue uses Typescript, so might be some similar patterns there?

Release Notes:

- Added new action (editor: Go to test) for jumping between source files and test files.

Go to test

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1601,6 +1601,16 @@
     "Markdown": [".rules", ".cursorrules", ".windsurfrules", ".clinerules"],
     "Shell Script": [".env.*"]
   },
+  // Patterns for navigating between source and test files.
+  // Use `{}` as a placeholder for the captured part of the path.
+  // Define both directions to enable navigation from source to test and back.
+  "go_to_test": {
+    "patterns": {
+      // Elixir
+      "lib/{}.ex": "test/{}_test.exs",
+      "test/{}_test.exs": "lib/{}.ex"
+    }
+  },
   // Settings for which version of Node.js and NPM to use when installing
   // language servers and Copilot.
   //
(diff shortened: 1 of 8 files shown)


dinocosta:
Hey @torepettersen , thank you for opening this pull request!

I’ve suggested introducing something similar but more generic, akin to what you’d get with [vim-projectionist](https://github.com/tpope/vim-projectionist), where we’d allow users to define their own projections and jump between alternate files, not necessarily only test files.

Although I haven’t yet opened a pull request for it, I’ve done a small demo to the team, and it’s something I’d like to work on in the future, time permitting. Since the approach would differ from the one suggested here, I’m going to go ahead and close this pull request. Rest assured, we’re aware that a feature like this would be helpful 🙂

torepettersen:
Hey @dinocosta,

Nice to hear that you have a feature like this on your radar :blush: 

Would you like me to do some changes to my code, to make it more like vim-projectionist?

In that case, I have a few questions:

What would you like the name of the action to be? Maybe "Alternate file"?

And how do you imagine the config? Maybe like this?
```json
"alternate_file": {
  "lib/{}.ex": {"alternate": "test/{}_test.exs"},
  "test/{}_test.exs": {"alternate": "lib/{}.ex"}
}
```

Also, do you have something else in mind?

PR #43044, opened 2025-11-19
util: Check whether discovered powershell is actually executable

### LINKED ISSUE #42944: Windows: powershell.exe not found
### Summary

Issues caused by administrator permissions

### Description
If I simply open zed, the command window will report that powershell.exe cannot be found. If I set to run as an administrator in the compatibility options, the problem is solved, but at the same time, it raises another issue that I cannot start zed using the right-click menu. I clicked but there was no response

Zed: v0.212.6 (Zed) 
OS: Windows 10.0.26100
Memory: 31.3 GiB
Architecture: x86_64
GPU: AMD Radeon 780M Graphics || AMD Corporation || 24.30.13.14 (24.30.13.14-250120a-412240C-Lenovo)

```log

```

</details>

Veykril:
Closes https://github.com/zed-industries/zed/issues/42944

The powershell we discovered might be in a directory with higher permission requirements which will cause us to fail using it.

Release Notes:

- Fixed powershell discovery disregarding admin requirements

### DIFF
--- a/crates/util/src/shell.rs
+++ b/crates/util/src/shell.rs
@@ -2,6 +2,8 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, fmt, path::Path, sync::LazyLock};
 
+use crate::command::new_std_command;
+
 /// Shell configuration to open the terminal with.
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq, JsonSchema, Hash)]
 #[serde(rename_all = "snake_case")]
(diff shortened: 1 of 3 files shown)


crates/util/src/shell.rs
```diff
@@ -108,16 +110,12 @@ pub fn get_windows_system_shell() -> String {
...
-        };
-
-        #[cfg(target_pointer_width = "32")]
         let env_var = if find_alternate {
-            "ProgramW6432"
+            if cfg!(target_pointer_width = "64") {
+                "ProgramFiles(x86)"
+            } else {
+                "ProgramW6432"
+            }
```
CharlesChen0823, line 118: Could using `where.exe pwsh` or `where.exe powershell` to check whether powershell exists?

PR #43046, opened 2025-11-19
Bump fsevent tests timeout to 30s on CI

as-cii:
Refs #42960 

I managed to reproduce the flakiness locally with a timeout of 500ms. Bumping the timeout to 4s locally solved the issue, so I suspect the flakiness just has to do with how busy the fsevent subsystem is at any given moment. This pull request uses a very long timeout on CI to ensure we eventually get the event we're expecting.

Release Notes:

- N/A


as-cii:
We disabled these, so we can close.

Issue #42444, opened 2025-11-11
Windows: The issue of input method switching (not sure if it's a feature or a bug)


arcqiufeng:
### Summary

Recently, due to the resolution of a bug related to https://github.com/zed-industries/zed/issues/41766, a feature that I highly appreciate no longer exists. I am not sure if the change to this feature was intentional on Zed's part, but I am particularly fond of this feature. However, with this latest bug fix, this beloved feature of Zed has unfortunately disappeared.

I don't know if this is a bug. Therefore, in addition to posting a thread in the discussion area to put forward a request - 
https://github.com/zed-industries/zed/discussions/42439, I also reported a bug here.

When I first started using Zed's Vim mode, I noticed that it by default addresses a pain point for Chinese users (or other CJK users): when switching to normal mode or visual mode, Zed automatically ignores the influence of the Chinese input method and directly bypasses the IME's interception to capture keystrokes. Only when entering Insert mode does the Chinese input method take effect. This is simply an excellent feature.

In Neovim and Vim for VS Code, solving this problem requires additional measures, such as third-party programs (im-select), which is quite troublesome (for reference, the solution in VSCodeVim: https://github.com/VSCodeVim/Vim?tab=readme-ov-file#input-method). However, in Zed, this problem does not exist by default, and it works right out of the box.

Unfortunately, in a recent bug fix (of course, it was indeed a bug, and I did encounter this issue)
After this bug was fixed: https://github.com/zed-industries/zed/issues/41766

This amazing and very useful feature of automatically switching input methods (or the feature of automatically ignoring the input method in non-insert mode) has also disappeared. If I had to choose between the two, I would rather not fix this bug.

### Zed Version and System Specs

Zed: v0.212.3 (Zed Preview)
OS: Windows 10.0.22631
Memory: 63.7 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3080 || NVIDIA Corporation || 576.83 r575_00

### If applicable, attach your `Zed.log` file to this issue.

<details><summary>Zed.log</summary>

```log

```

</details>

reflectronic:
The fix that changed this behavior was intended primarily to fix the following issues:
- https://github.com/zed-industries/zed/issues/36736
- https://github.com/zed-industries/zed/issues/39608
- https://github.com/zed-industries/zed/issues/39991
- https://github.com/zed-industries/zed/issues/41223
- https://github.com/zed-industries/zed/issues/41656
[...]

PR #43056, opened 2025-11-19
vim: Disable IME in normal mode on Windows

Yocchi3191:
## Summary

Automatically disables IME (Input Method Editor) when exiting Vim Insert mode on Windows.

Fixes the issue where Japanese IME input would appear in Normal mode.

Related discussion: #42444 (user reported similar issues with Chinese IME)

## Changes

- Added `enable_ime()` and `disable_ime()` methods to `PlatformWindow` trait
- Implemented IME context preservation in Windows platform layer
- Vim mode switching now controls IME state on Windows

### Before
- IME remains active in Normal mode
- Pressing movement keys (j, k, h, l) triggers IME conversion candidates
- This causes unintended character input such as "っj", "っk", "っh", "っl" to appear in the editor instead of executing Vim commands
- Users must disable IME manually

### After
- **Insert mode**: IME enabled, previous state preserved
- **Normal mode**: IME automatically disabled
- **Return to Insert mode**: IME re-enabled with previous state

## Testing

Tested manually on Windows 11 with Japanese IME:
- [x] IME disabled when exited Insert mode pressing ESC
- [x] IME re-enabled when pressing `i` (Insert mode)
- [x] IME state (on/off) preserved across mode switches
- [x] No ghost input or candidate window issues

---

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/platform.rs
+++ b/crates/gpui/src/platform.rs
@@ -553,6 +553,9 @@ pub(crate) trait PlatformWindow: HasWindowHandle + HasDisplayHandle {
 
     fn update_ime_position(&self, _bounds: Bounds<Pixels>);
 
+    fn enable_ime(&self) {}
+    fn disable_ime(&self) {}
+
     #[cfg(any(test, feature = "test-support"))]
     fn as_test(&mut self) -> Option<&mut TestWindow> {
         None
--- a/crates/gpui/src/platform/windows/window.rs
+++ b/crates/gpui/src/platform/windows/window.rs
@@ -927,8 +927,6 @@ impl PlatformWindow for WindowsWindow {
         let hwnd = self.0.hwnd;
         unsafe {
             use std::ptr;
-            eprintln!("disable_ime called");
-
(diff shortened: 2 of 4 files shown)


Yocchi3191:
@cla-bot check

Yocchi3191:
@cla-bot check

reflectronic:
Thank you for looking into this issue. Unfortunately, there is a flaw with this approach: there are various text areas that work 'normally' even when Vim mode is enabled (most importantly, the terminal window). With this patch, the user will won't be able to use the IME at all with these elements while the editor is in normal mode. So, I do not think we can take this patch as-is.

A better way to implement this would be to enable/disable the IME depending on whether the focused control is accepting text input. I will take a look at re-implementing this patch this way.

Yocchi3191:
Thank you for pointing out the bug. I apologize for missing this check on my end. 

I appreciate you taking the time to look into re-implementing this feature with the proper approach. I hope my PR was at least helpful in identifying the issue.

PR #43065, opened 2025-11-19
Fix treat dash as word char in linked editing

### LINKED ISSUE #43060: Auto renaming a custom element in HTML/JSX is not working as expected
### Summary

When editing changing a <div></div> to say <example-component> it results in <example-component></example>

### Description
As the title says, auto renaming works well for standard elements such as divs, buttons, etc.
When using a custom element though, it stop changing the closing element after the dash.

Steps to reproduce:
1. Make a <div></div>
2. Start changing tag from `div` to `example-component`
3. As soon as you type the `-` of `example-component`, the closing element stops editing.

<video src="https://github.com/user-attachments/assets/aa10b374-818c-40b7-a344-ff04425789d9" controls preload></video>

Zed: v0.212.7 (Zed) 
OS: macOS 15.6.1
Memory: 32 GiB
Architecture: aarch64

ddoemonn:
Closes #43060

Release Notes:

- Fixed linked editing breaking when typing dash in HTML/JSX custom element names (e.g., `<my-component>`)

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -4452,7 +4452,8 @@ impl Editor {
                     let classifier = snapshot
                         .char_classifier_at(start_anchor.to_offset(&snapshot))
                         .scope_context(Some(CharScopeContext::LinkedEdit));
-                    classifier.is_word(char)
+                    // Treat dash as word character for HTML/JSX tag names
+                    classifier.is_word(char) || char == '-'
                 });
 
                 if is_word_char {


JosephTLyons:
Thanks for the PR @ddoemonn, we will review ASAP!

crates/editor/src/editor.rs
```diff
@@ -4452,7 +4452,8 @@ impl Editor {
                     let classifier = snapshot
                         .char_classifier_at(start_anchor.to_offset(&snapshot))
                         .scope_context(Some(CharScopeContext::LinkedEdit));
-                    classifier.is_word(char)
+                    // Treat dash as word character for HTML/JSX tag names
+                    classifier.is_word(char) || char == '-'
```
osiewicz, line 4456: Each language has a `linked_edit_characters` configuration knobs, which means we don't need to add that logic here. Instead, it should be applied to HTML/JSX directly.

ddoemonn: Yeah you are totally right actually

ddoemonn:
@JosephTLyons I actually see assignee in issue. If you want I can close PR that's totally okay for me

yeskunall:
@ddoemonn that’s actually because I managed to reproduce the bug in https://github.com/zed-industries/zed/issues/43060 and wanted to pair with somebody to fix it. If you think you know what to do to fix it, I’d say go ahead.

probably-neb:
I'm going to mark this as a draft. If you'd like to add the `linked_edit_characters` to the HTML/JSX languages then we can keep this open until you do @ddoemonn, otherwise we can close the PR.

yeskunall:
Hey @ddoemonn we got a fix coming for this in #43179 so we’ll be closing this now. Thanks for taking a stab at it tho!

Issue #23288, opened 2025-01-17
Implement Graceful Recovery for GPU Device Loss in Zed


Vanuan:
### Check for existing issues

- [x] Completed

### **Summary**
By addressing GPU device loss gracefully, Zed can provide a more reliable and user-friendly experience, ensuring that users can focus on their work without worrying about unexpected crashes or data loss. 🚀
### Introduction

Currently, Zed lacks robust mechanisms to handle GPU device loss, which can occur due to suspend/resume events, driver crashes, overheating, or other hardware/software issues. When a GPU error occurs, Zed becomes unresponsive, leading to a poor user experience. This issue proposes a comprehensive solution to detect GPU device loss, recover gracefully, and provide clear feedback to users.  

#### **Context**  
This issue builds on the discussion in [Blade Issue #248](https://github.com/kvark/blade/issues/248), where the need for enhanced synchronization feedback and GPU device loss handling was outlined. Blade’s maintainer, **kvark**, highlighted that:  
- GPU device loss can occur during various operations (e.g., `wait_for`, `create_texture`, or command submissions).  
- Suspend/resume events alone are insufficient for detecting GPU device loss, as errors can occur without these events.  
- Blade’s role is to abstract GPU intricacies, but handling suspend/resume at the OS level might be more appropriate.  

#### **Proposed Solution**  
To address GPU device loss gracefully, Zed should implement the following:  

1. **Error Detection and Reporting** 🔍  
   - Use Blade’s `SyncStatus::InvalidSyncPoint` to detect GPU device loss.  
   - Extend this to handle errors during other GPU operations (e.g., texture creation, command submissions).  

2. **Graceful Recovery** 🔄  
   - Save unsaved data to a recovery file before restarting.  
   - Restart Zed and restore application state (e.g., open windows, buffers) from the recovery file.  

3. **User Feedback** 💬  
   - Display a message: “Zed encountered a GPU error and has restarted.”  
   - Offer an option to submit debug information for telemetry and diagnostics.  

4. **Band-Aid Solutions (Short-Term)** 🩹  
   - Implement temporary measures like preventing system sleep, saving data before suspend, or quitting on suspend.  

#### **Why This Matters**  

[...]

PR #43070, opened 2025-11-19
gpui: Implement GPU device loss recovery for Linux X11

### LINKED ISSUE #27751: Lock screen shown instead of Zed when the computer is unlocked
### Summary

When I unlock my computer (Pop OS), instead of showing zed, it shows the lock screen to me (almost like a screenshot) - shown in the picture below

Image

Steps to trigger the problem:
1. Use PopOS
2. Open Zed
3. Suspend the laptop
4. Log back in

Actual Behavior:
Shows an image of the lock screen, and I cannot interact with or close zed, without force closing from System Monitor

Expected Behavior:
Should be able to continue from where it left off

Zed: v0.179.4 (Zed)
OS: Linux X11 pop 22.04
Memory: 35 GiB
Architecture: x86_64
GPU: NVIDIA GeForce GTX 1660 Ti with Max-Q Design || NVIDIA || 565.77

```
2025-03-30T13:22:28.400679137+01:00 [ERROR] GPU hung
2025-03-30T13:22:28.400752504+01:00 [ERROR] your device information is: DeviceInformation { is_software_emulated: false, device_name: "NVIDIA GeForce GTX 1660 Ti with Max-Q Design", driver_name: "NVIDIA", driver_info: "565.77" }
```
</details>

### LINKED ISSUE #32318: Zed hangs after exiting sleep mode in Linux
### Summary

Steps to trigger the problem:
1. Put laptop into sleep mode with Zed opened
2. Open it 
3. Zed is hanging

Actual Behavior:
Zed hangs after laptop exits sleep mode.
Expected Behavior:
Zed resumes normally.

Zed: v0.187.8 (Zed)
OS: Linux X11 linuxmint 22.1
Memory: 31 GiB
Architecture: x86_64
GPU: Quadro T2000 || NVIDIA || 565.77

_No response_

larstalian:
Closes #32318
Closes #27751 
Partial fix for #23288 (X11 only, not Wayland)

Release Notes:

- Fixed Zed hanging with 100% CPU usage after resuming from sleep on Linux X11 with NVIDIA GPUs

### DIFF
--- a/crates/gpui/src/platform/blade/blade_atlas.rs
+++ b/crates/gpui/src/platform/blade/blade_atlas.rs
@@ -61,6 +61,24 @@ impl BladeAtlas {
         self.0.lock().destroy();
     }
 
+    #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
+    pub(crate) fn handle_device_lost(&self) {
+        let mut lock = self.0.lock();
+        lock.storage = BladeAtlasStorage::default();
+        lock.tiles_by_key.clear();
+    }
+
+    #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
+    pub(crate) fn update_gpu_context(&self, new_gpu: &Arc<gpu::Context>) {
+        let mut lock = self.0.lock();
+        lock.gpu = Arc::clone(new_gpu);
+        lock.upload_belt = BufferBelt::new(BufferBeltDescriptor {
+            memory: gpu::Memory::Upload,
+            min_chunk_size: 0x10000,
+            alignment: 64,
+        });
+    }
+
     pub fn before_frame(&self, gpu_encoder: &mut gpu::CommandEncoder) {
         let mut lock = self.0.lock();
         lock.flush(gpu_encoder);
--- a/crates/gpui/src/platform/blade/blade_atlas.rs
+++ b/crates/gpui/src/platform/blade/blade_atlas.rs
@@ -64,8 +64,11 @@ impl BladeAtlas {
     #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
     pub(crate) fn handle_device_lost(&self) {
         let mut lock = self.0.lock();
-        lock.storage = BladeAtlasStorage::default();
+        let gpu = Arc::clone(&lock.gpu);
+        lock.storage.clear_and_reuse(&gpu);
         lock.tiles_by_key.clear();
+        lock.initializations.clear();
+        lock.uploads.clear();
     }
 
     #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
@@ -89,12 +92,13 @@ impl BladeAtlas {
         lock.upload_belt.flush(sync_point);
     }
 
-    pub fn get_texture_info(&self, id: AtlasTextureId) -> BladeTextureInfo {
+    pub fn get_texture_info(&self, id: AtlasTextureId) -> Option<BladeTextureInfo> {
         let lock = self.0.lock();
-        let texture = &lock.storage[id];
-        BladeTextureInfo {
+        let textures = &lock.storage[id.kind];
+        let texture = textures.textures.get(id.index as usize)?.as_ref()?;
+        Some(BladeTextureInfo {
             raw_view: texture.raw_view,
-        }
+        })
     }
 }
 
@@ -322,6 +326,18 @@ impl BladeAtlasStorage {
         for mut texture in self.polychrome_textures.drain().flatten() {
             texture.destroy(gpu);
         }
+    }
+
+    fn clear_and_reuse(&mut self, gpu: &gpu::Context) {
+        for mut texture in self.monochrome_textures.textures.drain(..).flatten() {
+            texture.destroy(gpu);
+        }
+        for mut texture in self.polychrome_textures.textures.drain(..).flatten() {
+            texture.destroy(gpu);
+        }
+
+        self.monochrome_textures.free_list.clear();
+        self.polychrome_textures.free_list.clear();
     }
 }
 
--- a/crates/gpui/src/platform/blade/blade_atlas.rs
+++ b/crates/gpui/src/platform/blade/blade_atlas.rs
@@ -64,8 +64,7 @@ impl BladeAtlas {
     #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
     pub(crate) fn handle_device_lost(&self) {
         let mut lock = self.0.lock();
-        let gpu = Arc::clone(&lock.gpu);
-        lock.storage.clear_and_reuse(&gpu);
+        lock.storage.clear_without_destroy();
         lock.tiles_by_key.clear();
         lock.initializations.clear();
         lock.uploads.clear();
@@ -328,14 +327,10 @@ impl BladeAtlasStorage {
         }
     }
 
-    fn clear_and_reuse(&mut self, gpu: &gpu::Context) {
-        for mut texture in self.monochrome_textures.textures.drain(..).flatten() {
-            texture.destroy(gpu);
-        }
-        for mut texture in self.polychrome_textures.textures.drain(..).flatten() {
-            texture.destroy(gpu);
-        }
-
+    #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
+    fn clear_without_destroy(&mut self) {
+        self.monochrome_textures.textures.clear();
+        self.polychrome_textures.textures.clear();
         self.monochrome_textures.free_list.clear();
         self.polychrome_textures.free_list.clear();
     }
--- a/crates/gpui/src/platform/blade/blade_renderer.rs
+++ b/crates/gpui/src/platform/blade/blade_renderer.rs
@@ -653,7 +653,6 @@ impl BladeRenderer {
     }
 
     #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
-    /// Used to recover from GPU hangs caused by suspend/resume or driver issues.
     pub fn recreate_after_device_loss<
         I: raw_window_handle::HasWindowHandle + raw_window_handle::HasDisplayHandle,
     >(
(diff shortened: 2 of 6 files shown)


crates/gpui/src/platform/blade/blade_renderer.rs
```diff
@@ -342,6 +342,7 @@ pub struct BladeRenderer {
     path_intermediate_msaa_texture: Option<gpu::Texture>,
     path_intermediate_msaa_texture_view: Option<gpu::TextureView>,
     rendering_parameters: RenderingParameters,
+    skip_draws: bool,
```
larstalian, line 345: this matches the windows implementation

larstalian:
Tested on wayland - no GPU hang like X11, but sprites still don't render after suspend/resume

crates/gpui/src/platform/linux/x11/window.rs
```diff
@@ -882,9 +883,63 @@ impl X11Window {
...
+    }
+
+    fn handle_draw_failure(&self, err: anyhow::Error) {
+        log::error!("Renderer draw failed: {}", err);
+        let mut inner = self.0.state.borrow_mut();
+        if let Err(recovery_err) =
+            Self::attempt_gpu_recovery(&mut inner, self.0.x_window, &self.0.xcb, 0)
+        {
+            log::error!("GPU recovery failed: {}", recovery_err);
+            std::process::exit(1);
```
Vanuan: Well, silently exiting the app probably not better than freezing? Maybe `panic!()` would be better? So that eventually, the panic hook could handle this and re-spawn Zed with a message on start "Some problem occurred, here's the details"?

Vanuan: For example:

```rust
use std::{panic, env, process};

// Define an environment variable used to check if the process is a 'restarted' instance
const RESTART_ENV_VAR: &str = "APP_RESTARTED_FROM_CRASH";

/// Placeholder function to simulate the logic for relaunching the process.
/// In a real-world scenario, this would likely involve using the 'exec' system call
/// or signaling an external supervisor.
fn relaunch_process(error_details: &str) {
    eprintln!("--- Relaunch Initiated ---");
    eprintln!("Error to pass: '{}'", error_details);
    eprintln!("In a real app, this would execute a new instance with the error info.");
    // Example: process::Command::new(env::current_exe().unwrap())
    //      .env(RESTART_ENV_VAR, error_details)
    //      .spawn().unwrap();
    
    // For this example, we'll just exit gracefully after 'relaunching'
    process::exit(1); 
}

fn main() {
    // 1. Check on start for the 'restarted' flag
    if let Ok(error_details) = env::var(RESTART_ENV_VAR) {
        println!("*** ⚠️ Application started as a restarted instance! ***");
        println!("Previous crash reason: {}", error_details);
        
        // Clear the environment variable so future runs are clean (optional)
        env::remove_var(RESTART_ENV_VAR);
        
        // Here you would implement logic like:
        // - Load previous state from disk
        // - Present a crash report to the user
        // - Attempt a less resource-intensive mode
    }

    // 2. Install the custom panic hook
    panic::set_hook(Box::new(|panic_info| {
        // --- CUSTOM PANIC HANDLER ---
        eprintln!("\n🚨 Custom Panic Hook Activated 🚨");
        eprintln!("Panic occurred at: {}", panic_info.location().unwrap());
        
        // Extract a simple error string to pass to the new process
        let error_message = format!("Panic: {}", panic_info.payload().downcast_ref::<&str>().unwrap_or(&"Unknown Error"));
        
        // Call the relaunch logic (which will likely exit the current process)
        // NOTE: This call is the key to your "respawn" concept.
        relaunch_process(&error_message);
    }));

    // --- APPLICATION LOGIC ---
    println!("Application running normally...");
    println!("Simulating a critical error now...");
    
    // 3. Trigger the panic
    // This will cause the panic hook to fire.
    panic!("GPU device lost");
    
    // This line is never reached
    println!("Application finished.");
}
```

larstalian: Okay great thanks, i will address that! Additionally, after testing this solution a couple days i have found that it occasionally does not recover properly. I am investigating and coding up a fix for that. Will set the PR as draft until that is fixed. I really appreciate the feedback

crates/gpui/src/platform/blade/blade_atlas.rs
```diff
@@ -71,12 +92,13 @@ impl BladeAtlas {
...
+    pub fn get_texture_info(&self, id: AtlasTextureId) -> Option<BladeTextureInfo> {
         let lock = self.0.lock();
-        let texture = &lock.storage[id];
-        BladeTextureInfo {
+        let textures = &lock.storage[id.kind];
+        let texture = textures.textures.get(id.index as usize)?.as_ref()?;
+        Some(BladeTextureInfo {
             raw_view: texture.raw_view,
-        }
+        })
```
larstalian, line 100: After GPU recovery the atlas is cleared, but queued scene frames may still reference old texture IDs, since they can outlive the cleared atlas. `skip_draw` should prevent rendering until fresh scenes are generated, but this is additional safety: skip instead of panic. This was the panic that occasionally happened in the first commit.

Vanuan: I think @kvark is appropriate person to review this. Though he has a strong opinion that this is unrecoverable without restarting the application.

crates/gpui/src/platform/blade/blade_renderer.rs
```diff
@@ -428,29 +429,31 @@ impl BladeRenderer {
...
+                    log::error!(
+                        "if that helps you're running into https://github.com/zed-industries/zed/issues/26143"
+                    );
+                }
                 log::error!(
-                    "if that helps you're running into https://github.com/zed-industries/zed/issues/26143"
+                    "your device information is: {:?}",
+                    self.gpu.device_information()
                 );
+                return Err(anyhow::anyhow!("GPU device hung or lost"));
```
Vanuan, line 453: Actually, it depends on wait_for implementation, could be just a normal frame timeout. Although, MAX_FRAME_TIME_MS is set to 10 seconds, so it usually coincides. The major architectural issue is that blade doesn't surface error codes to the application layer, so it's impossible to know whether device is really lost. But this is a good start, although it might lead to false positives on some platforms.

crates/gpui/src/platform/blade/blade_renderer.rs
```diff
@@ -428,29 +429,31 @@ impl BladeRenderer {
...
             }
-            log::error!(
-                "your device information is: {:?}",
-                self.gpu.device_information()
-            );
-            while !self.gpu.wait_for(&last_sp, MAX_FRAME_TIME_MS) {}
         }
+        Ok(())
```
Vanuan, line 456: If this approach is accepted, next logical step would to to extend wait_for or change bool to Ok/Err upstream

crates/gpui/src/platform/linux/x11/window.rs
```diff
@@ -882,9 +883,70 @@ impl X11Window {
...
+        let mut inner = self.0.state.borrow_mut();
+        let screen_index = inner.display.id().0 as usize;
+        if let Err(recovery_err) =
+            Self::attempt_gpu_recovery(&mut inner, self.0.x_window, &self.0.xcb, screen_index)
+        {
+            panic!(
+                "GPU recovery failed after device loss. This may be a driver issue. \
+                 Please try restarting Zed. Error: {}",
+                recovery_err
+            );
```
Vanuan, line 914: Some info for context. This is the first step as it will surface telemetry to Zed developers so that they can easily understand how often does this issue happen. As it appears on every suspend resume, this is A LOT and may overwhelm Sentry. If we could reliably determine the device lost error, it would be possible to try alternative approach to recovery: instead of crashing the app we could try restarting it (App:: Restart()). But it will lose telemetry. Maybe extract App::Restart context and sending it to the crash handling server so that it could restart automatically if the issue is known and we don't need telemetry. But then, there's a restart loop problem if the crash is during initialization. I wonder how chromium solves it.


Anyway, acknowledging a problem is always a good start.

Having that information, consider adjusting the panic message to be useful to those reviewing Sentry issues rather than to the user. Maybe a simple "GPU hung" would be ok.

Screenshot_20251127-035147

Screenshot_20251127-035715
Screenshot_20251127-040009

Vanuan:
Check out how blade reports GPU crash:
https://github.com/kvark/blade/blob/8cd905be95d9bcf476f69790fcfdbeb615c5154c/blade-graphics/src/vulkan/command.rs#L427C1-L451

Maybe you could copy paste some of this or just call this function if the context is available.

Vanuan:
FIY I'm trying to push a proper detection for device lost: https://github.com/kvark/blade/pull/286

larstalian:
> FIY I'm trying to push a proper detection for device lost: [kvark/blade#286](https://github.com/kvark/blade/pull/286)

yeah okay. I have some changes locally now that seems to recover properly, but its a larger and more complex. Turned out to be a bit bigger fix than i first envisioned.

Vanuan:
The only robust way I see is separating rendering process from application process and adding (protobuf?) RPC. So that any crash will lead to thin layer restart and immediate release of resources. But that would probably be a move towards chromium and electron which Zed developers despise despite their role in its success 
Screenshot_20251130-171814

larstalian:
Tested on Ubuntu 24.04 with Nvidia RTX 5060. Recovery now works reliably with multiple windows open. Once https://github.com/kvark/blade/pull/286 lands, we can replace the catch_unwind with proper wait_for_result() error handling.

Vanuan:
According to brief research, it appears that wgpu, like Blade, does not expose Vulkan errors and uses abstractions throughout. It's kind of more opaque does not allow detection of device loss events (though there's abstraction `Device::lost`) and sometimes treats timeouts like successful renders or device lost events. Like Blade, Wgpu doesn't provide mechanisms for automatic device recovery and that should be handled by the application (Zed or GPUI).


Because resources like textures and buffers are tied to a specific Device instance, once that device is lost, those handles are dead. The application (whether it's GPUI or a standalone app) must manually:

- Detect the loss (via Device::lost).
- Request a new Adapter and Device.
- Re-upload all necessary GPU resources

So this patch is not obsolete per se, it's just needs migrating to Wgpu API.

Vanuan:
PTAL #49154

PR #43074, opened 2025-11-19
Fix panic in the git panel when toggling `sort_by_path`

cole-miller:
We call `entry_by_path` on the `bulk_staging` anchor entry at the beginning of `update_visible_entries`, but in the codepath where `sort_by_path` is toggled on or off, we clear entries without clearing `bulk_staging` or counts, causing that `entry_by_path` to do an out of bounds index. Fixed by clearing `bulk_staging` as well.

Release Notes:

- N/A


kubkon (commented):
Nice find!

PR #43077, opened 2025-11-19
Fix git track file renames in git panel

### LINKED ISSUE #30549: Git: `git mv` shows as deleted file + untracked new file

ddoemonn:
Closes #30549

Release Notes:

- Fixed: Git renames now properly show as "old.rs → new.rs" in the git panel instead of appearing as separate deleted and untracked files


Angelk90:
@ddoemonn : I renamed a file but I don't understand what the difference is from before.
Screenshot 2025-11-20 alle 15 54 55

```shell
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    README.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	README2.md

no changes added to commit (use "git add" and/or "git commit -a")
```

ddoemonn:
@Angelk90 if you want tell me your commands from start to finish i can test maybe bc it was working for me

Angelk90:
@ddoemonn : 
Screenshot 2025-11-20 alle 16 27 35

ddoemonn:
@cole-miller hi!!, do you have any reviews?

ddoemonn:
@Angelk90 if you see smth like this in your terminal after you renamed via menu, its actually right behavior i guess

Screenshot 2025-11-26 at 17 16 24

secondl1ght:
Adding to @Angelk90's comment, this functionality should be extended to handle renames/moves using the Zed project panel. Related issue here #43089.

Angelk90:
@ddoemonn : Can you post a screenshot of the output in the git panel?

ddoemonn:
Hey @Angelk90,yes i can share but in which case?

Angelk90:
@ddoemonn : what does this pr do, rename a file, in the git panel what appears?

secondl1ght:
@ddoemonn any updates on this and related discussion #43089?

cole-miller:
@ddoemonn Thanks! It looks like there are a few things that aren't working yet:

- If I `git mv a.txt b.txt` in the panel, it shows up correctly. However, if I then unstage that panel entry, it splits into two entries:

  - a staged, deleted `a.txt`
  - an unstaged, untracked `b.txt

  It feels like we can do better than this (I don't know if we can get git to mark `a.txt -> b.txt` as a rename when it's not staged, but at the very least unstaging the rename entry should not leave the deletion of `a.txt` as staged). This might need some tweaks to which paths we call `GitStore::stage_entries`/`GitStore::unstage_entries` with.
- We need the renamed entry to show up in the project diff in some form. Ideally we would show the diff between the new version of the file and the old version (I believe they can be different from each other and still be marked as a rename by git)--so in the example above, we need to load the `HEAD` text of `a.txt` as the diff base, but still use `b.txt` for the main buffer that is being diffed. This probably requires some work in `git_store`, in particular `open_uncommitted_diff`. We also need to treat the renamed buffer a bit differently in the project diff--we should include it even if it doesn't have any diff hunks, and treat it like a deleted file in that situation (i.e. show the entire buffer, but collapse it by default).

Down to pair on this if you like: https://cal.com/cole-miller-zed/pairing

secondl1ght:
In addition to @cole-miller's comments, can we please include this same feature for the GUI flow as discussed in this thread and https://github.com/zed-industries/zed/issues/43089?

PR #43094, opened 2025-11-19
gpui: Document terminology of Element/View/Component, etc

Serophots:
I created this short document to remind myself of the terminology in GPUI, something I frequently wish for after a break from a gpui codebase or when I need to do something a little more complicated in a gpui project.

So I've PR'ed it because I reckon a document like this could be helpful to other folk using gpui.

Release Notes:

- N/A


SomeoneToIgnore:
Passing by and noting that this seem to belong to rustdocs instead of some standalone, separate document.

Serophots:
Good shout. I think I've implemented what you mean so that the page shows up on the docs.rs page.

If you mean to add these style doc-comments to the types themselves then I'd be happy to but I feel what I'm trying to achieve is to have a centralised place documenting common patterns and terminology. Let me know what you think.

PR #43102, opened 2025-11-19
Passthrough env to npm subcommands when using the system node runtime

### LINKED ISSUE #37866: Zed breaks npm configs that use environment variables
### Summary

Zed clears environment variables before running npm subcommands, which breaks npm configs that use them.

### Description

In my npm config, I set `cache` to `${XDG_CACHE_HOME}/npm` and `logs-dir` to `${XDG_STATE_HOME}/npm/logs` to avoid cluttering up my `$HOME`. I noticed that when Zed ran npm subcommands (usually to install a language server), a new "${XDG_STATE_HOME}" directory would appear in my `$HOME`. After checking the npm docs, I found that if an environment variable can't be resolved in the config, it's left unreplaced (see [this page](https://docs.npmjs.com/cli/v11/configuring-npm/npmrc#files)), hence the literal "${XDG_STATE_HOME}" directory.

After some hair pulling, I looked through the source code and found the culprit; Zed calls `env_clear()` before all npm subcommands are ran (see [here](https://github.com/zed-industries/zed/blob/ad02f6b9e3dbf7e49138bb83af9ed3812d6625dc/crates/node_runtime/src/node_runtime.rs#L704)). That means npm can't resolve my config's paths properly, so it leaves them as-is.

I cloned the repo and commented out all 3 instances of `env_clear()` in the `node_runtime` crate, and after building, npm had resolved my config properly and was sending logs to the intended directory. I know that envs are likely cleared for a reason so this isn't the most glamorous fix and a more robust solution is probably required here, but I haven't noticed any issues so far.

Steps to reproduce:
1. Use an environment variable in an `npmrc` file (mine was `/etc/npmrc`).
2. Open a file in Zed and trigger the use of an npm subcommand.
3. Check your `$HOME` directory and cry.

**Expected Behavior**:
Environment variable is resolved by npm.

**Actual Behavior**:
Environment variable is left as-is.

Zed: v0.202.7 (Zed) 
OS: Linux Wayland fedora 42
Memory: 31.1 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3060 || NVIDIA || 575.64.05

### LINKED ISSUE #39448: Windows Beta: Processes fork infinitely or fail to start when Python and Node.js runtimes are installed with Mise
### Summary
Processes fork infinitely or fail to start when Python and Node.js runtimes are installed with Mise

### Description
Python and Node.js runtimes are managed through [Mise](https://github.com/jdx/mise), with shims configured in [symlink](https://mise.jdx.dev/configuration/settings.html#windows_shim_mode) mode. In this setup, executables like npm.exe and python.exe are symlinks that point to the mise.exe binary, and the directory containing these shims is included in the system PATH.

Steps to trigger the problem:
1. Start Zed Nightly
2. Open a Python file
3. The focus shifts to the terminal running Mise exited with code 1

https://github.com/user-attachments/assets/18e2f69f-3589-43e5-8964-6bf402bb4672

1. Start Zed Nightly on a directory prepared with `uv init && uv venv`
2. Open a Python file
3. The focus shifts to the terminal running Mise exited with code 1
4. Activate the virtual environment selector
5. The focus shifts to the terminal running Mise exited with code 1

https://github.com/user-attachments/assets/83813314-6d63-4476-9fbf-ac8e12f19a2b

1. Start Zed Nightly
2. Open a JavaScript file
3. See "Downloading eslint..." running forever
4. Observe that npm processes are started infinitely in Process Explorer

https://github.com/user-attachments/assets/70bce31c-8a26-43d1-a9f4-5d0253c35956

Image

Running the command directly in the terminal returns the correct result.
Image

**Expected Behavior**: All of the above shouldn't happen

Zed: v0.208.0 (Zed Nightly e4f90b5da2b8a6d2b267edc02b333b7d7908d4d9)
OS: Windows 10.0.19045
Memory: 15.9 GiB
Architecture: x86_64
GPU: NVIDIA GeForce GTX 765M || NVIDIA Corporation || 425.31 r419_50

yjoer:
Closes #39448
Closes #37866

This PR expands the env-clearing fix from #42587 to include the SystemNodeRuntime, which covers Node.js installations managed by Mise. When running under the system runtime, npm subcommands were still launched with a cleared environment, preventing variables such as MISE_DATA_DIR from reaching the shim or the mise binary itself. As a result, Mise finds the npm binary in the default MISE_DATA_DIR, consistent with the behavior described in https://github.com/zed-industries/zed/issues/39448#issuecomment-3433644569.

This change ensures that environment variables are passed through for npm subcommands when using the system Node runtime, restoring expected behavior for Mise-managed Node installations. This also fixes cases where envs are used by npm itself.

Release Notes:

- Enable environment passthrough for npm subcommands

### DIFF
--- a/crates/node_runtime/src/node_runtime.rs
+++ b/crates/node_runtime/src/node_runtime.rs
@@ -414,7 +414,6 @@ impl ManagedNodeRuntime {
 
         let valid = if fs::metadata(&node_binary).await.is_ok() {
             let result = util::command::new_smol_command(&node_binary)
-                .env_clear()
                 .env(NODE_CA_CERTS_ENV_VAR, node_ca_certs)
                 .arg(npm_file)
                 .arg("--version")
@@ -701,7 +700,6 @@ impl NodeRuntimeTrait for SystemNodeRuntime {
         let mut command = util::command::new_smol_command(self.npm.clone());
         let path = path_with_node_binary_prepended(&self.node).unwrap_or_default();
         command
-            .env_clear()
             .env("PATH", path)
             .env(NODE_CA_CERTS_ENV_VAR, node_ca_certs)
             .arg(subcommand)


yjoer:
@cla-bot check

P1n3appl3:
I asked about this internally and I think I agree that we just shouldn't clear the env ever. Simply overriding the vars we care about should be enough. Thanks for the fix.

PR #42449, opened 2025-11-11
agent_ui: Add icon button to trigger the @-mention completions menu

danilo-leal:
Closes https://github.com/zed-industries/zed/issues/37087

This PR adds an icon button to the footer of the message editor enabling to trigger and interact with the @-mention completions menu with the mouse. This is a first step towards making other types of context you can add in Zed's agent panel more discoverable. Next, I want to improve the discoverability of images and selections, given that you wouldn't necessarily know they work in Zed without a clear way to see them. But I think that for now, this is enough to close the issue above, which had lots of productive comments and discussion!

Screenshot 2025-11-11 at 10  46 3@2x

Release Notes:

- agent: Added an icon button in the agent panel that allows to trigger the @-mention menu (for adding context) now also with the mouse.

PR #43138, opened 2025-11-20
assistant_text_thread: Add test for IconName backward compatibility in saved threads

errmayank:
Release Notes:

- Added test for IconName backward compatibility in saved text threads

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -902,6 +902,7 @@ dependencies = [
  "settings",
  "smallvec",
  "smol",
+ "strum 0.27.2",
  "telemetry_events",
  "text",
  "ui",
--- a/crates/assistant_text_thread/Cargo.toml
+++ b/crates/assistant_text_thread/Cargo.toml
@@ -45,6 +45,7 @@ serde_json.workspace = true
 settings.workspace = true
 smallvec.workspace = true
 smol.workspace = true
+strum.workspace = true
 telemetry_events.workspace = true
 text.workspace = true
 ui.workspace = true
--- a/crates/assistant_text_thread/src/assistant_text_thread_tests.rs
+++ b/crates/assistant_text_thread/src/assistant_text_thread_tests.rs
@@ -1518,28 +1518,24 @@ fn test_icon_backward_compatibility_in_saved_threads() {
     assert_eq!(ICONS.len(), IconName::iter().count());
 
     for icon in ICONS {
-        let json_str = serde_json::to_string(&json!({
-            "id": "test-thread",
-            "zed": "zed",
+        let saved_text_thread = serde_json::to_string(&json!({
+            "id": "saved-text-thread-id",
+            "zed": "context",
             "version": SavedTextThread::VERSION,
-            "text": "Test thread",
+            "text": "Saved text thread",
             "messages": [],
-            "summary": "test",
-            "slash_command_output_sections": [
-                {
-                    "range": {"start": 0, "end": 1},
-                    "icon": icon,
-                    "label": "test",
-                    "metadata": null
-                }
-            ]
+            "summary": "Saved text thread from history",
+            "slash_command_output_sections": [{
+                "range": {"start": 0, "end": 1},
+                "icon": icon,
+                "label": "test",
+                "metadata": null
+            }]
         }))
         .unwrap();
-        serde_json::from_str::<SavedTextThread>(&json_str).unwrap_or_else(|_| {
-            panic!(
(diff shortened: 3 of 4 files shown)


errmayank:
@danilo-leal Maybe I can add a test to avoid this from happening again?

<details>
<summary>
Also, here are the icons for which I couldn't find a direct alternative:
</summary>

<br>

- BookPlus
- Brain
- BugOff
- CircleOff
- Cloud
- DebugStop
- Equal
- FileCreate
- FolderX
- Function
- Layout
- LspDebug
- LspRestart
- LspStop
- MailOpen
- Microscope
- PanelLeft
- PanelRight
- PersonCircle
- PhoneIncoming
- PlayAlt
- PlayBug
- PocketKnife
- Reveal
- Route
- Save
- SettingsAlt
- SlashSquare
- SlidersVertical
- Snip
- SparkleAlt
- SparkleFilled
- Spinner
- Strikethrough
- Visible
- Wand
- ZedAssistantFilled

</details>

danilo-leal:
Nice! A test would be great to have. I also wonder if we need an alias for all those icons, though. Could we just have add it for the icon that caused the text thread problem? I'd prefer to add these aliases as we feel the need for them, rather than proactively add one for all icons that were previously deleted, just to make the process of deleting icons less bureaucratic.

errmayank:
Got it! The failure in the linked issue was because of missing `AtSign` but it was added back in #42449, so it shouldn't be a problem anymore.

I'll remove all the aliases and add a test case instead.

maxdeviant:
These aliases aren’t correct, as the icon names are snake_cased.

errmayank:
> These aliases aren’t correct, as the icon names are snake_cased.

Ah okay.

> Nice! A test would be great to have. I also wonder if we need an alias for all those icons, though. Could we just have add it for the icon that caused the text thread problem? I'd prefer to add these aliases as we feel the need for them, rather than proactively add one for all icons that were previously deleted, just to make the process of deleting icons less bureaucratic.

Makes sense, I've added a test for this.

danilo-leal:
Thanks for adding the test, @errmayank! How does having the test fixes the issue, though? I'm a bit confused about that.

errmayank:
@danilo-leal basically the test has a snapshot of current icon names and we test deserializing the text thread using them.

If someone removes/replaces an icon in `IconName`, they would either need to

- add an alias for it, for the old name
- or update test expectations when it's okay to remove the icon, doesn't affect end-users, etc.

Otherwise the test will fail to deserialize and panic.

Like if someone replaces `CaseSensitive` with `CaseSensitiveNew`, they'll need to add an alias for the old name:

```rs
#[serde(alias = "CaseSensitive")]
CaseSensitiveNew,
```

Now, If user has serialized text thread in their filesystem with `CaseSensitive`:

```
{
  "id": "saved-text-thread-id",
  "zed": "context",
  "version": "0.4.0",
  "text": "Saved text thread",
  "messages": [],
  "summary": "Saved text thread from history",
  "slash_command_output_sections": [
    {
      "range": {
        "start": 0,
        "end": 1
      },
      "icon": "CaseSensitive",
      "label": "test",
      "metadata": null
    }
  ]
}
```

The `icon` field will get deserialized to `CaseSensitiveNew`:

```
SavedTextThread {
    id: Some(
        TextThreadId(
            "saved-text-thread-id",
        ),
    ),
    zed: "context",
    version: "0.4.0",
    text: "Saved text thread",
    messages: [],
    summary: "Saved text thread from history",
    slash_command_output_sections: [
        SlashCommandOutputSection {
            range: 0..1,
            icon: CaseSensitiveNew,
            label: "test",
            metadata: None,
        },
    ],
    thought_process_output_sections: [],
}
```

<br>

> These aliases aren’t correct, as the icon names are snake_cased.

@maxdeviant I checked this again, icon names are snake_cased only when directly converted to strings like in `IconName::path()`. When they are serialized to JSON like in `TextThread::serialize()` they use PascalCase.

In the above example, `IconName::CaseSensitiveNew.path()` would be `icons/case_sensitive_new.svg` which seems correct coz old svg files will be removed and new ones will be used? Unless I'm missing something.

errmayank:
> How does having the test fixes the issue, though? I'm a bit confused about that.

icymi, if you were talking about how test fixes the original issue:

> Got it! The failure in the linked issue was because of missing `AtSign` but it was added back in #42449, so it shouldn't be a problem anymore.

maxdeviant:
Thank you for the PR, but I don't think this kind of test is the way we want to go about handling this scenario.

PR #43164, opened 2025-11-20
language: Move language server update to background when it takes too long

### LINKED ISSUE #42360: Language servers take a long time to start up when using a slow internet connection
### Summary

Language servers take a long time to start up when using a slow internet connection

### Description
A user reported this at the office hours. When using a slow internet connection, it takes a long time to startup language servers.
On screen I was able to observe that "Checking for updates to [LSP] ..." showed up for a long time. Seemingly we are waiting for the update check to complete before starting the LSP.

Steps to reproduce:
1. Use a slow/throttled internet connection
2. Restart Zed
3. Open a file that triggers Zed to startup a language server (e.g. Rust)
4. Wait for language server to start
5. Observe that "Checking for updates to [LSP] ..." is showing up in the lower left
6. Until that status message disappears the LSP is not running 💥 

**Expected Behavior**: If the LSP is already installed it starts-up instantly and the update (if there is one) is downloaded in the background
**Actual Behavior**: LSP only starts after checking for updates is completed

I can see two ways to solve this:
- Use a short timeout (e.g. 3 seconds) to check for the latest version
- Immediately start the LSP and run the update-check in the background

Any Zed version

smitbarmase:
Closes https://github.com/zed-industries/zed/issues/42360

If updating a language server takes longer than 10 seconds, we now fall back to launching the currently installed version (if exists) and continue downloading the update in the background.

Release Notes:

- Improved language server updates for slow connection, now Zed launches existing server if the update is taking too long.

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -609,8 +609,7 @@ impl LocalLspStore {
             delegate.update_status(adapter.name.clone(), BinaryStatus::None);
 
             let mut binary = match (existing_binary, maybe_download_binary) {
-                (Ok(binary), None) => binary,
-                (error @ Err(_), None) => return error,
+                (binary, None) => binary?,
                 (Err(_), Some(downloader)) => downloader.await?,
                 (Ok(existing_binary), Some(downloader)) => {
                     let mut download_timeout = cx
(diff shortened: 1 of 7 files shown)


crates/project/src/lsp_store.rs
```diff
@@ -599,14 +600,37 @@ impl LocalLspStore {
...
                 .get_language_server_command(delegate.clone(), toolchain, lsp_binary_options, cx)
+                .await
                 .await;
 
             delegate.update_status(adapter.name.clone(), BinaryStatus::None);
 
-            let mut binary = binary_result?;
+            let mut binary = match (existing_binary, maybe_download_binary) {
+                (Ok(binary), None) => binary,
+                (error @ Err(_), None) => return error,
```
osiewicz: I guess we could consolidate these two branches into one
```suggestion
                (binary, None) => binary?,
```

PR #43172, opened 2025-11-20
Support bracket colorization (rainbow brackets)

SomeoneToIgnore:
Deals with https://github.com/zed-industries/zed/issues/5259

Highlights brackets with different colors based on their depth.
Uses existing tree-sitter queries from brackets.scm to find brackets, uses theme's accents to color them.

https://github.com/user-attachments/assets/cc5f3aba-22fa-446d-9af7-ba6e772029da

1. Adds `colorize_brackets` language setting that allows, per language or globally for all languages, to configure whether Zed should color the brackets for a particular language.

Disabled for all languages by default.

2. Any given language can opt-out a certain bracket pair by amending the brackets.scm like `("\"" @open "\"" @close) ` -> `(("\"" @open "\"" @close) (#set! rainbow.exclude))`

3. Brackets are using colors from theme accents, which can be overridden as

```jsonc
"theme_overrides": {
  "One Dark": {
    "accents": ["#ff69b4", "#7fff00", "#ff1493", "#00ffff", "#ff8c00", "#9400d3"]
  }
},
```

Release Notes:

- Added bracket colorization (rainbow brackets) support. Use `colorize_brackets` language setting to enable.


vhp1360:
Hello everyone, my bad for bringing up a closed issue :-)
If you're open to feedback,this feature only works for square brackets at this time,if you dont mind, it might be cooler if works for parentheses and curly brackets too.

thank you

SomeoneToIgnore:
No worries about the post per se: any ones with the technical details are totally worth it.

Your bad is that you post sends noise with no context at all — which language, which theme, a snippet to repro etc.

vhp1360:
thanks for reply and sorry for less info because the first message in this topic kinda suggested this featureis universal and I know some file type might not be this feature.
but for more information, I checked in some files type
as I checked the curly braces not colorized in plain text,file with unknown type, yaml, jinja sample,json file
image
actually in above image, open parentheses dont close automatically and that happened for all bracket types if between double quotes.
in html file type:
image
in bash type
image
in conf file like nginx.conf
image
in Dockerfile
image

SomeoneToIgnore:
I have no idea what "universal" is in this context, as there's a few axes to track things at, and definitely did not plan to apply this word around.

In fact, had tried to explain what this support means in one of the first sentences, perhaps, too condensed:
image

but expanded right below: 

image

1. 
From here, we can see that there are "theme accents" and it's possible to configure them, there could be multiple.
From the screenshots it seems that there are no issue with the color order and if there would be, there is a text on how to adjust that per theme.

2.

The rest of the text mentions that `brackets.scm` are in play, and bracket **pairs** are considered (that is also displayed in that tree-sitter query with `(#set! rainbow.exclude)` bit).

So, all `(((((((((((((((` things like the last visible line of your Dockerfile example won't be highlighted because those are not bracket pairs.

The rest is not clear from this PR's description, but has a page in the docs: https://zed.dev/docs/extensions/languages?highlight=brackets.scm#bracket-matching

I've found it by opening the docs page and searching for `brackets.scm` from this PR's description.

In short, those brackets/highlights/outline/etc.scm files are collections tree-sitter queries on a parcitular topic, which Zed can query without knowing any particular details about the language.

Each language is backed up by some of these files, either provided by Zed for some "core" languages like Rust, or via extensions — if any extension does not support bracket queries, it does not include `brackets.scm` file into its bundle.
If that happens, Zed cannot colorize the brackets; neither can Zed colorize the bracket pairs that are not mentioned in `brackets.scm`.

If I quickly check some of the examples' extensions:

* `dockerifle` does not have any `brackets.scm`: https://github.com/zed-extensions/dockerfile/tree/6f341f0c18f6afb4ed3e9346a9260013cf84c377/languages/dockerfile
* same as `nginx`: https://github.com/zed-extensions/nginx/tree/ebc8f9c6d529ddd0b490e0fb19f2e2c7c1e1578f/languages/nginx
* `html` has `brackets.scm` but there are no conventional brackets there
https://github.com/zed-industries/zed/blob/a350438a21c80e1199ceae78f4e9f7e6f7403330/extensions/html/languages/html/brackets.scm#L1-L6

```
("<" @open "/>" @close)
("</" @open ">" @close)
("<" @open ">" @close)
(("\"" @open "\"" @close) (#set! rainbow.exclude))
((element (start_tag) @open (end_tag) @close) (#set! newline.only) (#set! rainbow.exclude))
```

and I tend to agree that plain html does not have any syntax brackets?
What are all those `{}` `[]` and `()` pairs mean in terms of html **language**?

VSCode seems to agree with my confusion:
image

* `bash` (and `css` when I looked at it comparing to VSCode) examples are the most interesting ones, those seem like a bug in Zed, as the bracket queries are similar to what Rust has:

https://github.com/zed-industries/zed/blob/d72746773faf458452ee393cf3ec01a164f98b37/crates/languages/src/bash/brackets.scm#L1-L13

yet presumably tree-sitter parser library underneath returns some funny bracket ranges that Zed does not normalize normally.

SomeoneToIgnore:
Regarding the bash example, I've checked it a bit more in detail, and seems that its tree-sitter parser itself starts to fail quite badly on brackets misplace: 

https://github.com/user-attachments/assets/043b34bb-3122-40aa-af9c-3d54ed95acd8

You can see that entire chunks of the file stop being highlighted at some, quite log, bracket threshold.

The parser is what answers all queries about the language, and tries to make sense of the code, so here's sometimes a fine line between a bad parser and a bad code.

Here, it's sort of both: I would expect the parser to recover from yet another bracket pair and continue checking the code better, but it does not happen and it starts to match random bracket pairs at start and end, considering what's between some form of an invalid bash expression.

Yet, the bash code with that many brackets is not valid too?

One way or another, Zed cannot fix it in its code, and https://github.com/tree-sitter/tree-sitter-bash has to get a fix, or another tree-sitter parser should be used for the language.

vhp1360:
Thanks for your quick check and reply.  

As I understand from your explanation, the new feature will only work if each language properly supports and implements the required `brackets.scm` queries. Also, if brackets (like the unclosed parentheses in some of my screenshots) aren't highlighted, it's likely due to missing or incomplete support in the language extension.

Thanks again for your time and the detailed clarification! 🌸

PR #43173, opened 2025-11-20
git: Support project-level git.inline_blame settings 

### LINKED ISSUE #13547: Inline blame for large files with long Git histories causes Zed to create more and more Git processes using >95% CPU
### Check for existing issues

- [x] Completed

I work in a large Git repository with a long history. In the repository, there is a large text file (1.8 MB) [very similar to the CODEOWNERS file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners). This file has a long history: ~52_000 commits.

I noticed today that my whole system was becoming really slow, and when I opened the activity monitor -- I saw a bunch of Git processes all using 95% or more CPU. And I noticed that quitting Zed immediately killed those processes as well. After poking around a bit, I was able to reliably reproduce the problem by opening the problematic file in Zed -- it immediately starts creating Git processes which use a lot of CPU and they never complete (unless I quit Zed).

On a hunch, I disabled the inline blame feature in the settings, and that completely resolves the problem.

```
"git": {
    "inline_blame": {
        "enabled": false
    }
},
```

cc @maxbrunsfeld since you seem to enjoy fixing these types of problems, or at least read about them 😂 

### Environment

Zed: v0.141.2 (Zed Preview)
OS: macOS 14.5.0
Memory: 64 GiB
Architecture: aarch64

<details><summary>Screenshots...</summary>

First noticed this:

Image

And then this a bit later:

Image

</details>

_No response_

leoliu0605:
Closes #13547

We've limited the `git blame` command's thread usage, and following the issue discussion, we've added a project-level `git.inline_blame` setting to allow users to configure it for each project.

https://github.com/user-attachments/assets/5ff239f7-1cc6-4180-9f3b-35c96b76a0eb

Release Notes:

- Added support for configuring the git.inline_blame setting at the project level

### DIFF
--- a/crates/collab/src/tests/editor_tests.rs
+++ b/crates/collab/src/tests/editor_tests.rs
@@ -3482,14 +3482,16 @@ async fn test_git_blame_is_forwarded(cx_a: &mut TestAppContext, cx_b: &mut TestA
     cx_a.update(|cx| {
         SettingsStore::update_global(cx, |store, cx| {
             store.update_user_settings(cx, |settings| {
-                settings.git.get_or_insert_default().inline_blame = inline_blame_off_settings;
+                settings.git.get_or_insert_default().project.inline_blame =
+                    inline_blame_off_settings;
             });
         });
     });
     cx_b.update(|cx| {
         SettingsStore::update_global(cx, |store, cx| {
             store.update_user_settings(cx, |settings| {
-                settings.git.get_or_insert_default().inline_blame = inline_blame_off_settings;
+                settings.git.get_or_insert_default().project.inline_blame =
+                    inline_blame_off_settings;
             });
         });
     });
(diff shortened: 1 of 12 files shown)


leoliu0605:
Everything runs as expected, but an error message appears in the terminal and I don't know why:

image

esthertrapadoux:
@leoliu0605 just following up to let you know the team went pretty deep in more git feature work and we got behind on PR reviews. We're ramping back up and appreciate the patience! 👍

GitGlimpse895:
Hey @Anthony-Eid — following up on @leoliu0605's ping from January. You self-assigned this in December and merged their companion PR (#43326, global git enable/disable) the same week, but this one has had no review activity in 3.5 months.

The underlying issue is still present in `main` — there's no way to override `git.inline_blame` at the project level via `.zed/settings.json`, so users with large repos and long git histories are still hitting the 95%+ CPU bug. This PR is the direct fix, and the contributor has already proven their reliability with the work you merged in December.

PR #43177, opened 2025-11-20
acp: Support specifying settings for extensions

bennetbo:
This allows you to specify default_model and default_mode for ACP extensions, e.g.
```
"auggie": {
  "default_model": "gpt-5",
  "default_mode": "default",
  "type": "extension"
},
```

Release Notes:

- Added support for specifying settings for ACP extensions (`default_mode`, `default_model`)

### DIFF
--- a/crates/agent_servers/src/custom.rs
+++ b/crates/agent_servers/src/custom.rs
@@ -44,7 +44,7 @@ impl crate::AgentServer for CustomAgentServer {
 
         settings
             .as_ref()
-            .and_then(|s| s.default_mode.clone().map(|m| acp::SessionModeId(m.into())))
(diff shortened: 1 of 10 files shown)


lucasyvas:
Does this work with ACP installed via extension like `opencode`? It doesn't seem like a recognized key.

bennetbo:
Yes it does: 
```
"opencode": {
  "type": "extension",
  "default_model": "opencode/grok-code"
},
```

But we should probably include it in the autocompletion. For now it's easiest to set this from the UI, by choosing a mode/model from the mode/model selector and holding `cmd` while selecting it. That will set the default mode/model in your settings automatically.

PR #43179, opened 2025-11-20
extensions: Add `-` as linked edit character for HTML

### LINKED ISSUE #43060: Auto renaming a custom element in HTML/JSX is not working as expected

smitbarmase:
Closes https://github.com/zed-industries/zed/issues/43060

Fixed issue where typing in custom HTML tag would not complete subsequent end tag for - character.

Release Notes:

- N/A

### DIFF
--- a/extensions/html/languages/html/config.toml
+++ b/extensions/html/languages/html/config.toml
@@ -14,3 +14,6 @@ brackets = [
 ]
 completion_query_characters = ["-"]
 prettier_parser_name = "html"
+
+[overrides.default]
+linked_edit_characters = ["-"]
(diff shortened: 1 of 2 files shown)


maxdeviant:
@smitbarmase We shouldn't have release notes for this, since the change will go out with the HTML extension on a different cadence than Zed itself.

smitbarmase:
Got it 👍

PR #43181, opened 2025-11-20
Add `git commit --cleanup=strip/--cleanup=whitespace` option

miguelraz:
TODO: There's a related issue somewhere
TODO: Add tests

Added a new enum to handle the `strip` and `whitespace` settings that are found in `man git` - this can be extended to handle other cases should people want to add them.

This PR needed to consider the local git settings *not* overriding the __remote__ git settings, which is why it doesn't mess with the `git_proto.rs` protobuf definitions.

Release Notes:

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -6,6 +6,7 @@ pub mod pending_op;
 use crate::{
     ProjectEnvironment, ProjectItem, ProjectPath,
     buffer_store::{BufferStore, BufferStoreEvent},
+    project_settings::{GitSettings, ProjectSettings},
     worktree_store::{WorktreeStore, WorktreeStoreEvent},
 };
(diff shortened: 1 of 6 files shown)


crates/project/src/git_store.rs
```diff
@@ -4266,21 +4265,41 @@ impl Repository {
...
+        let project_path =
+            self.repo_path_to_project_path(&RepoPath::from_rel_path(RelPath::empty()), cx);
+        self.send_job(Some("git commit".into()), move |git_repo, cx| async move {
             match git_repo {
                 RepositoryState::Local {
                     backend,
                     environment,
                     ..
                 } => {
+                    let cleanup = cx.update(|cx| {
```
miguelraz, line 4285: Context: we don't want to do a simple `ProjectSettings::global_get()` because that could override the local repository's configuration, so we have to reach in and read the settings.

PR #43182, opened 2025-11-20
Elevate save with ZED_SU_CMD, safe-write

### LINKED ISSUE #13585: RFE/Linux: polkit integration
### Check for existing issues

- [X] Completed

### Describe the feature

Currently on linux, zed has no way to prompt for elevated permissions (root) if the user does not have the permissions to write/read a file.

This is traditionally done with a request to polkit over dbus which will then prompt the user to input the superuser password in order to successfully perform the operation.

This is useful when attempting to use zed to update config files in the /etc/ directory, or, attempting to open files the user doesn't have read permission on, as some examples.

#10423 is a tangibly related issue for macos, however this request is for linux.

Example of a polkit prompt
Screenshot from 2024-06-27 11-15-04

ninetailedtori:
Closes #13585

Release Notes:

- Safe write from memory
- Function includes an "attempt initial save", if failed, attempt sudo, from ZED_SU_CMD env var. Should be compatible with sudo-likes as well, we're only passing `-E` hardcoded to preserve environment.
- also copies existing file to tempfile before the memory-copy, in the case that mid-write fails, will restore automatically from the tempfile.


ninetailedtori:
@cla-bot check

P1n3appl3:
Hey there, I think we want to go in a different direction with this feature. Our desired behavior is:

- normal saves can fail
- upon failure to save, show a prompt with a "save with sudo" option
- use a separate codepath to do the privileged save
- hardcode and/or use a setting for the sudo command, not an env var

We appreciate the contribution but I'm gonna close this given the different tack we want to take. Feel free to open a new PR that's more in-line with that.

ninetailedtori:
> Hey there, I think we want to go in a different direction with this feature. Our desired behavior is:
> 
>     * normal saves can fail
> 
>     * upon failure to save, show a prompt with a "save with sudo" option
> 
>     * use a separate codepath to do the privileged save
> 
>     * hardcode and/or use a setting for the sudo command, not an env var
> 
> 
> We appreciate the contribution but I'm gonna close this given the different tack we want to take. Feel free to open a new PR that's more in-line with that.

Hardcoding a setting for the sudo command, how's that possible? The crate for managing and accessing settings values *uses* the fs crate, and I've tried to implement it before but it generates a circular dependency. I'm not the best at Rust, so I'd love advice on that.

P1n3appl3:
The idea is that you'd propagate the "file failed to save" error up from the fs crate, and then in workspace or project or something you'd handle that by checking the setting, and if it's set to "prompt to retry with sudo" then you show the prompt and handle the accept button for that by calling a different fs function specifically for doing the elevation.

PR #43187, opened 2025-11-20
editor: Fix git-hunk toggling for adjacent hunks

seanstrom:
Closes #42934 

Release Notes:

- Fix toggling adjacent git-diff hunks based on the reported behaviour in #42934

### DIFF
--- a/crates/multi_buffer/src/multi_buffer.rs
+++ b/crates/multi_buffer/src/multi_buffer.rs
@@ -2463,7 +2463,8 @@ impl MultiBuffer {
             let range = range.to_point(&snapshot);
             let start = snapshot.point_to_offset(Point::new(range.start.row, 0));
             let end = snapshot.point_to_offset(Point::new(range.end.row + 1, 0));
-            let start = start.saturating_sub_usize(1);
+            // why do we need to subtract 1 from the start offset?
+            // let start = start.saturating_sub_usize(1);
             let end = snapshot.len().min(end + 1usize);
             cursor.seek(&start, Bias::Right);
             while let Some(item) = cursor.item() {
--- a/crates/multi_buffer/src/multi_buffer.rs
+++ b/crates/multi_buffer/src/multi_buffer.rs
@@ -2462,10 +2462,8 @@ impl MultiBuffer {
         for range in ranges {
             let range = range.to_point(&snapshot);
             let start = snapshot.point_to_offset(Point::new(range.start.row, 0));
-            let end = snapshot.point_to_offset(Point::new(range.end.row + 1, 0));
-            // why do we need to subtract 1 from the start offset?
-            // let start = start.saturating_sub_usize(1);
-            let end = snapshot.len().min(end + 1usize);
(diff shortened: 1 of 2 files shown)


seanstrom:
### Commentary

* The command for toggling selected git-diff hunks as some complex behaviours depending on the position of the cursor, and whether there are multiple git-hunks near each other (separated by a line, or without a line separation).
* I've left some [notes](https://github.com/zed-industries/zed/issues/42934#issuecomment-3559594701) and [recordings](https://github.com/zed-industries/zed/issues/42934#issuecomment-3559600333) attached to issue #42934 
* And this PR creates a failing test for one of the cases with adjacent hunks, however, we may want to add more tests and define the intended behaviour of the hunk-toggling command to avoid more regressions.

crates/multi_buffer/src/multi_buffer.rs
```diff
@@ -2463,7 +2463,8 @@ impl MultiBuffer {
             let range = range.to_point(&snapshot);
             let start = snapshot.point_to_offset(Point::new(range.start.row, 0));
             let end = snapshot.point_to_offset(Point::new(range.end.row + 1, 0));
-            let start = start.saturating_sub_usize(1);
+            // why do we need to subtract 1 from the start offset?
+            // let start = start.saturating_sub_usize(1);
```
seanstrom: This change is questionable, since it seems to break some existing behaviour when attempting to collapse a diff-hunk when the cursor is a line below the expanded hunk.

Prior to this change, the cursor could be at the beginning of the line below an expanded hunk, and the toggle-selected-hunks command would collapse the hunk above the cursor.

After this change, when the cursor is a line below an expanded hunk, and the cursor is a line above a deletion hunk, the toggle-selected-hunks command will expand the deletion hunk.

seanstrom:
#### Here's a screen recording of the current behaviour with the changes in this PR:

https://github.com/user-attachments/assets/49abe418-cb21-4740-bd8a-a6798fbec50a

kubkon:
Thanks for the PR! I had a closer look and the proposed change does not seem the right solution as it regresses other behaviour plus makes it generally inconsistent. However, I did discuss that with the team and we have an idea how to approach this. With that in mind, I'm gonna go ahead and close the PR and hopefully get a fix up ASAP.

seanstrom:
Thanks for taking a look!

Glad to hear there was discussion, hope this was helpful for kickstarting the real solution, eager to learn more about it when it’s ready.

feel free to tag me in the next pr, happy to follow along and help out if needed 👍

kubkon:
I've done some more digging and I agree with you that subtracting 1 from multibuffer offset makes little sense and it doesn't seem to regress anything (anything that we have coverage for). It also keeps things consistent, whereas trying to special-case this would inevitably result in inconsistent handling depending if we are dealing with single empty cursor selection or more elaborate selections. I am therefore inclined to merge this PR with some minor cleanups and rebase on main (already done that locally, will push in a minute). If we regress something that we don't have coverage for, we'll revert and add a new test case and whatnot.

PR #43190, opened 2025-11-20
Fix `connect.host` setting being ignored by debugpy

### LINKED ISSUE #42727: Debugger: debugpy adapter ignores the ["connect"]["host"] value and uses "127.0.0.1"
### Summary

When using the debugpy adapter, the ["connect"]["host"] value is ignored and "127.0.0.1" is used instead.

### Description
Steps to trigger the problem:
1. Run `python3 -m debugpy --listen 0.0.0.0:5678 --wait-for-client script.py` on another machine with port 5678 exposed.
2. Place the following in .zed/debug.json:
```
[
  {
    "adapter": "Debugpy",
    "label": "External Python",
    "request": "attach",
    "connect": { "host": "<hostname-of-server>", "port": 5678 }
  }
]
```

3. Press the debug icon and select "External Python", that we just made.

**Expected Behavior**:
The debugger should attach to the debugpy on the hostname of the server we specified.

**Actual Behavior**:
The console outputs
```
error: 
error: process exited before debugger attached.
```
and nothings happens.

When tested in vscode, it works as expected, so I know it is not my environment or the setup on the server.

Inspecting the logs reveal that it is trying to go to 127.0.0.1:5678 instead of hostname-of-server:5678.

Because of this, I tried with ssh port forwarding:
`ssh <username>@<hostname-of-server> -L 5678:127.0.0.1:5678`

and that worked - even if I keep the host as "hostname-of-server", indicating that it is indeed using "127.0.0.1" instead of what I specified.

Zed: v0.212.5 (Zed) 
OS: Linux X11 ubuntu 24.04
Memory: 31.1 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 2060 || NVIDIA || 580.95.05

```log
2025-11-14T15:13:57+01:00 INFO  [dap::transport] Debug adapter has connected to TCP server 127.0.0.1:5678
2025-11-14T15:13:57+01:00 ERROR [debugger_ui::debugger_panel]
error: process exited before debugger attached.
```

</details>

HactarCE:
Closes #42727

Unfortunately we can only support IPv4 addresses right now because `TcpArguments` only supports an IPv4 address. I'm not sure how difficult it would be to lift this limitation.

Release Notes:

- Fixed `connect.host` setting being ignored by debugpy

### DIFF
--- a/crates/dap_adapters/src/python.rs
+++ b/crates/dap_adapters/src/python.rs
@@ -368,6 +368,9 @@ impl PythonDebugAdapter {
                 bail!("Cannot have two different ports in debug configuration")
             }
 
+            if let Some(hostname) = config_host {
+                tcp_connection.host = Some(hostname.parse().context("hostname must be IPv4")?);
+            }
             tcp_connection.port = config_port;
             DebugpyLaunchMode::AttachWithConnect { host: config_host }
         } else {


UnknownDK:
Maybe update this as well then - this is the whole reason why I thought it should work https://github.com/zed-industries/zed/blob/e2f6422b3e1b4c1901523d9000c136e9d624fa7e/crates/dap_adapters/src/python.rs#L613

PR #43193, opened 2025-11-20
rope: Use SIMD in `Chunk::push_str` to create byte masks faster

XeroOl:
Improves the performance of rope's Chunk::push_str using SIMD.

Zed's Rope Bitmaps make it much faster to search through a Rope more quickly, with the trade-off that Bitmaps need to be generated ahead of time (in `Chunk::push_str`).

I modified `Chunk::push_str`'s Bitmap code to loop over each *byte* instead of each *codepoint*, but that wasn't enough for Rust to optimize the loop, so I created a version that explicitly tells Rust how to to use SIMD instructions.

On my computer this improves the performance of the push by approximately 50%.
(`cd crates/rope && cargo bench`)
```
push/4096               time:   [42.499 µs 42.522 µs 42.552 µs]
                        thrpt:  [91.800 MiB/s 91.864 MiB/s 91.913 MiB/s]
                 change:
                        time:   [-55.021% -54.961% -54.904%] (p = 0.00 < 0.05)
                        thrpt:  [+121.75% +122.03% +122.33%]
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  7 (7.00%) high mild
  4 (4.00%) high severe

push/65536              time:   [315.91 µs 318.51 µs 321.42 µs]
                        thrpt:  [194.45 MiB/s 196.23 MiB/s 197.84 MiB/s]
                 change:
                        time:   [-50.450% -49.954% -49.430%] (p = 0.00 < 0.05)
                        thrpt:  [+97.745% +99.815% +101.82%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
```

Concerns:
* This code uses `std::simd` which is currently only available in nightly Rust.
* The code is now more confusing than the non-simd version, although not by too much.
* I've chosen a width of 64 somewhat arbitrarily as it's the maximum size `std::simd` provides, but it's possible that other widths could be better.
* Rope's tests may use u16 sized bitmasks which might not cover the SIMD code

Release Notes:

- N/A

### DIFF
--- a/crates/rope/src/chunk.rs
+++ b/crates/rope/src/chunk.rs
@@ -1,6 +1,14 @@
 use crate::{OffsetUtf16, Point, PointUtf16, TextSummary, Unclipped};
 use arrayvec::ArrayString;
-use std::{cmp, ops::Range};
+use std::{
+    cmp,
+    ops::Range,
+    simd::{
+        Simd,
+        cmp::{SimdPartialEq, SimdPartialOrd},
+        num::SimdUint,
+    },
+};
 use sum_tree::Bias;
 use unicode_segmentation::GraphemeCursor;
 use util::debug_panic;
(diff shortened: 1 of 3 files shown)


XeroOl:
@cla-bot check

XeroOl:
I've marked this as a draft because I don't expect it to be merged in the short term - it uses `portable-simd` which is only available in nightly rust.

miguelraz:
Thanks for the submission @XeroOl - and yes, the nightly requirement is a blocker, for starters.

Still, hugely appreciate you putting this forward and benchmarking it too! 🫶

Veykril:
I think there are two ways to make this work, either find a library (Ideally one that we already have in the lock file, though it looks like there are none ) that offers the simd primitives required here, or hand roll this SIMD specialization for a couple platforms, given that the rope is at the heart of a lot of things this kind of improvement would definitely be nice to have.

XeroOl:
I just recently found `Rope::push_large`, https://github.com/zed-industries/zed/blob/7881551ddac3240def8301662719400db5d02c61/crates/rope/src/rope.rs#L199 , and I want to try to optimize at this level instead of just the chunk code.

Veykril:
Closing in favor of https://github.com/zed-industries/zed/pull/44345

PR #43222, opened 2025-11-21
Make numeric steppers editable in settings UI

### LINKED ISSUE #43208: ui: Fix bug with text centering in editable number fields
### Summary

Currently, the numeric steppers in the settings UI are much less usable than a simple JSON file. We should fix that.

abulgit:
Closes #43208

Numeric fields in the settings UI can now be clicked to edit the value directly instead of only using the +/- buttons.
The edit mode implementation was already there but wasn't being triggered. Now clicking the value activates the text input, making it much faster to set specific values.
What I changed
- Changed the read-only number display from `h_flex()` to a `div()`
- Added an `on_click` handler that switches to `NumberFieldMode::Edit`
- Added cursor pointer and hover state
- Switched text alignment to center

https://github.com/user-attachments/assets/71a93eb1-c7be-49f9-9bc3-aac7befeb417

Release Notes:

-Make numeric steppers editable in settings UI

### DIFF
--- a/crates/ui_input/src/number_field.rs
+++ b/crates/ui_input/src/number_field.rs
@@ -432,19 +432,23 @@ impl<T: NumberFieldType> RenderOnce for NumberField<T> {
                             .bg(bg_color)
                             .in_focus(|this| this.border_color(focus_border_color))
                             .child(match *self.mode.read(cx) {
-                                NumberFieldMode::Read => h_flex()
-                                    .px_1()
-                                    .flex_1()
-                                    .justify_center()
-                                    .child(Label::new((self.format)(&self.value)))
-                                    .into_any_element(),
-                                // Edit mode is disabled until we implement center text alignment for editor
-                                // mode.write(cx, NumberFieldMode::Edit);
-                                //
-                                // When we get to making Edit mode work, we shouldn't even focus the decrement/increment buttons.
-                                // Focus should go instead straight to the editor, avoiding any double-step focus.
-                                // In this world, the buttons become a mouse-only interaction, given users should be able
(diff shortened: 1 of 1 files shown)


mikayla-maki (changes requested):
Unfortunately that UI wasn’t enabled because it doesn’t look good right now. The text needs to be center aligned, see something like this for a reference: https://www.shadcn.io/patterns/input-types-3

danilo-leal:
Yeah, I don't think the problem is in the number field code itself; it's something with making text centered in editors. I think @Anthony-Eid has context on this as he did some good research on what's the challenge...

abulgit:
Yeah so, I've been investigating why center alignment doesn't work for the number field editor.

The issue is that `ShapedLine::paint()` in https://github.com/zed-industries/zed/blob/1ce58a88ccdfb7e45106d66fa44c9db35b4f94b9/crates/gpui/src/text_system/line.rs#L74 hardcodes `TextAlign::default()` (Left) and never uses the `text_align` from `EditorStyle`.

The rendering already supports alignment - `WrappedLine::paint()` accepts an align parameter and `aligned_origin_x()` handles the positioning. It's just that `ShapedLine` doesn't expose this capability.

My plan is

Add `paint_aligned()` and `paint_background_aligned()` methods to `ShapedLine` (similar to how `WrappedLine` works)
Pass `text_align` from `EditorStyle` through `EditorLayout` down to `LineWithInvisibles::draw()` in element.rs

Does this approach make sense, or is there a better way?

mikayla-maki:
Those look good! Maybe we make ShapedLine::paint just take an alignment?

PR #43234, opened 2025-11-21
helix: Fix `Vim::NextWordEnd` off-by-one in `HelixSelect`

### LINKED ISSUE #38121: Helix sometimes `v` excludes current character
### Summary

Entering Helix Select mode after `esc` sometimes excludes current character from selection

### Description

This bug is hard to explain. In helix cursor is a selection. In zed:helix we don't always keep selection open, instead some actions will default to current character if no selection is present (helix_yank #35612, or #38119).

In HelixNormal pressing `v` selects character under cursor and moving around will expand that selection. Original character will always remain selected if you go forward or backward. For the most part this is how selection behaves.

There is a sequence, that triggers a bug in this:

`w b escape v w`

going to next word and backward in Helix will keep the first word selected. Escape resets selection (like ;) then v will go to select mode and w will select next word again. 

One would expect entirety of the word to be selected, but somehow the first character is excluded:

Image

I wrote a test case for this:

```
    #[gpui::test]
    async fn test_helix_wbvw_motions(cx: &mut gpui::TestAppContext) {
        cx.enable_helix();
        cx.set_state("ˇone two", Mode::HelixNormal);
        cx.simulate_keystrokes("v b ; v w");
        cx.assert_state("«one ˇ»two", Mode::HelixSelect);

        cx.set_state("ˇone two", Mode::HelixNormal);
        cx.simulate_keystrokes("v b escape v w");
        cx.assert_state("«one ˇ»two", Mode::HelixSelect);
    }
```

Strangely though - the test case passes. Also using `;` instead of `escape` seems to work correctly. 

I wonder what it is about `escape` and only from under `cargo run` that messes with selection mode.

Below are the state after each operation:

`w b`

Image

`escape`

Image

`v`

Image

`w`

Image

**Expected Behavior**:
entirety of the first word should be selected at the end of the sequence.

Version: built from `main` with #38117 applied.

Mac

Config only enabling vim and helix modes.

AidanV:
Closes #43209
Closes #38121

Starting on the first character.
Running `v e` before changes: 
image

Running `v e` after changes:
image

Change Notes:

- Added helix selection sanitation code that directly mirrors the code in the Vim [`visual_motion`](https://github.com/AidanV/zed/blob/b6728c080c5d14ded7002d0276deb5c19d42ed8a/crates/vim/src/visual.rs#L237) method. I kept the comments from the Vim section that explains its purpose.
- The above change converted the problem from fixing `v e` to fixing `v w`. Since `w` is treated differently in Helix than in Vim (i.e. `w` in Vim goes to the first character of a word and `w` in Helix goes to the character before a word. Commented [here](https://github.com/AidanV/zed/blob/b6728c080c5d14ded7002d0276deb5c19d42ed8a/crates/vim/src/helix.rs#L132)), the code treats `w` in `HelixSelect` as a motion that differs from the Vim motion in the same way that the function [`helix_move_cursor`](https://github.com/AidanV/zed/blob/b6728c080c5d14ded7002d0276deb5c19d42ed8a/crates/vim/src/helix.rs#L353) separates these behaviors. 
- Added a regression test

Release Notes:

- Fixes bug where `Vim::NextWordEnd` in `HelixSelect` would not select whole word.


jrobsonchase (approved):
Diagnosis and fix make sense to me, tested locally and seems to be working as expected :+1: 

This also fixes https://github.com/zed-industries/zed/issues/38121, which seems to be caused by a similar current character/cursor position issue.

AidanV:
Thanks for pointing that out! I updated the description to also close #38121. 

I added a test that checks the core behavior of the selection being the full width of the cursor. I believe this is testing the same issue outlined in #38121. I did not add the specific test written in #38121 because `escape` clears the selection in Zed but it does not in Helix. If we fixed this later down the line it would cause a false regression.

kubkon:
Nice! Could you resolve conflicts so that we could go ahead and merge it?

PR #43236, opened 2025-11-21
Enable Zeta edit predictions with custom URL without authentication

DominicBurkart:
Enables using Zeta edit predictions with a custom `ZED_PREDICT_EDITS_URL` without requiring authentication to Zed servers. This is useful for:

- Development and testing workflows
- Self-hosted Zeta instances
- Custom AI model endpoints

Prior context on this usage of `ZED_PREDICT_EDITS_URL`: https://github.com/zed-industries/zed/pull/30418

Release Notes:
- Improved self-hosted zeta UX. Users no longer have to log into Zed to use custom or self-hosted zeta backends.


crates/zed/src/zed/edit_prediction_registry.rs
```diff
@@ -238,7 +238,9 @@ fn assign_edit_prediction_provider(
                     }
 
                     editor.set_edit_prediction_provider(Some(provider), window, cx);
-                } else if user_store.read(cx).current_user().is_some() {
+                } else if user_store.read(cx).current_user().is_some()
+                    || std::env::var("ZED_PREDICT_EDITS_URL").is_ok()
```
nicky-isaacs-awoo: can `ZED_PREDICT_EDITS_URL` be pulled into a constant?

nicky-isaacs-awoo (commented):
LGTM but very much not an expert

crates/zed/src/zed/edit_prediction_registry.rs
DominicBurkart: I tried this, but it actually gets kind of ugly since it's hardcoded in 4 files that don't import each other. Removing from the PR.

crates/edit_prediction/src/edit_prediction.rs
```diff
@@ -1671,13 +1697,9 @@ impl EditPredictionStore {
...
-        let url = if let Some(predict_edits_url) = PREDICT_EDITS_URL.as_ref() {
-            http_client::Url::parse(&predict_edits_url)?
-        } else {
-            client
-                .http_client()
-                .build_zed_llm_url("/predict_edits/raw", &[])?
-        };
+        let url = client
+            .http_client()
+            .build_zed_llm_url("/predict_edits/raw", &[])?;
```
agu-z, line 1702: Raw requests are only used by staff for experimentation, and the payload struct doesn't match, so we shouldn't use PREDICT_EDITS_URL

agu-z:
@DominicBurkart Thank you for your contribution! Merging after some tweaks

pjv:
I understand use of this environment variable isn't an exactly supported configuration but i have a local ollama instance with zeta available on it. How exactly can i configure zed preview to use that zeta instance for edit predictions?

SuperKenVery:
@pjv You may want to use this [proxy](https://github.com/g0t4/zed-zeta-server) to wrap ollama, then use `PREDICT_EDITS_URL` to point zed to it.

PR #43241, opened 2025-11-21
Fix vitest test running and debugging for v4 with backwards compatibility

freddyfallon:
## Summary

This PR updates the vitest test runner integration to use the modern `--no-file-parallelism` flag instead of the deprecated `--poolOptions.forks.minForks=0` and `--poolOptions.forks.maxForks=1` flags.

## Changes

- Replaced verbose pool options with `--no-file-parallelism` flag in both file-level and symbol-level vitest test tasks
- This change works with vitest v4 while maintaining backwards compatibility with earlier versions (or 3 at least!)

## Testing

- Added test `test_vitest_uses_no_file_parallelism_flag` that verifies:
  - The `--no-file-parallelism` flag is present in generated test tasks
  - The deprecated `poolOptions` flags are not present
- Manually tested with both vitest v4 and older versions to confirm backwards compatibility
- All existing tests pass

## Impact

This allows Zed users to run and debug vitest tests in projects using vitest v4 while maintaining support for earlier versions.

Release Notes:

- Fixed vitest test running and debugging for projects using vitest v4


freddyfallon:
Note that I built the project and tested it with both vitest 3 and 4. I notice that Webstorm is also having this problem!

crates/languages/src/typescript.rs
```diff
@@ -1679,4 +1677,91 @@ mod tests {
             "Bun should come before Node"
         );
     }
+
```
freddyfallon: Not sure if we actually want these tests that I added. Like how useful are they really?

freddyfallon:
Ah, I tested this on v1, v2, v3, and v4 and they all worked. However, I did get this when running on v0.34.6:

```rust
throw new CACError(`Unknown option \`${name.length > 1 ? `--${name}` : `-${name}`}\``);
```

So I guess this PR will fix running modern vitest tests, but will break anything pre v1. Let me know if you want me to handle that.

crates/languages/src/typescript.rs
cole-miller: Yeah I don't think we need them

cole-miller:
Thanks!

PR #43246, opened 2025-11-21
git: Show renamed files with arrow icon in Git panel

flashus:
## Closes #43089

This PR improves the Git panel to properly handle renamed files:

1. **Display renamed files with arrow icon** - Instead of showing separate deleted/added entries, renamed files now appear with an arrow icon (→) and their original path.

2. **Show correct diff for renamed files** - When viewing a staged renamed file's diff, the content is now correctly loaded from the original path in HEAD, not the new path (which doesn't exist in HEAD).

## Release Notes:

  - Rename detection uses git's default 50% similarity threshold
  - Works for both `git mv` and manual renames that meet similarity threshold
  - Also supports copied files (shown the same way as renames)
  - Arrow icon uses the theme's `version_control_modified` color
  - Original path displays in muted color to avoid visual clutter
  - Viewing diff of renamed files now correctly shows changes from original file

## Screenshot:

image

### DIFF
--- a/crates/collab/src/db.rs
+++ b/crates/collab/src/db.rs
@@ -742,6 +742,7 @@ fn db_status_to_proto(
         status: Some(proto::GitFileStatus {
             variant: Some(variant),
         }),
+        old_path: None,
     })
 }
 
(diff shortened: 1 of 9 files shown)


cole-miller:
Thanks for working on this! However, there's an existing PR open for this at #43077 that we're working on getting merged, so going to close in favor of that.

PR #43250, opened 2025-11-21
cli: Allow opening non-existent paths

### LINKED ISSUE #4441: CLI: open files for unexisting directories
## Current behaviour

```bash
$ zed path/to/unexisting/dir/file.txt
Error: No such file or directory (os error 2)
```

## Feature idea

I'd like this same command to open the file as if the folder existed. And create the folder when the file is saved.

```bash
$ zed path/to/unexisting/dir/file.txt
# open editor without creating the path until file saved
```

This is the VSC behaviour, and I love it!

BuonOmo:
Changes are made to `parse_path_with_position`:
we try to get the canonical, existing parts of
a path, then append the non-existing parts.

Closes #4441

Release Notes:

- Added the possibility to open a non-existing path using `zed` CLI
  ```
  zed path/to/non/existing/file.txt
  ```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3088,6 +3088,7 @@ dependencies = [
  "rayon",
  "release_channel",
  "serde",
+ "serde_json",
  "tempfile",
  "util",
  "windows 0.61.3",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -34,6 +34,10 @@ util.workspace = true
 tempfile.workspace = true
 rayon.workspace = true
 
+[dev-dependencies]
+serde_json.workspace = true
+util = { workspace = true, features = ["test-support"] }
+
 [target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
 exec.workspace = true
 fork.workspace = true
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -129,32 +129,173 @@ struct Args {
     askpass: Option<String>,
 }
 
+/// Parses a path containing a position (e.g. `path:line:column`)
+/// and returns its canonicalized string representation.
+///
+/// If a part of path doesn't exist, it will canonicalize the
+/// existing part and append the non-existing part.
+///
+/// This method must return an absolute path, as many zed
+/// crates assume absolute paths.
 fn parse_path_with_position(argument_str: &str) -> anyhow::Result<String> {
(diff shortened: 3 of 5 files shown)


maxdeviant:
Related to https://github.com/zed-industries/zed/pull/42328

BuonOmo:
FYI I had also implemented a `test_open_and_save_non_existing_dir_and_file` method in the `zed` crate to make sure that this works as well. I ended up not adding it to the pull request as I felt like this was already expected. I still have it locally if you want it though :)

SomeoneToIgnore (approved):
Wow, that's quite a stark contrast with another PR.

I am a bit concerned that `with_cmd` might influence tests, but let's run them and see.

Thank you for improving this.

BuonOmo:
> Wow, that's quite a stark contrast with another PR.

😊 it was a pleasure working on it! It would be a pleasure to work on more advanced topics 😄

> I am a bit concerned that `with_cmd` might influence tests, but let's run them and see.

Oh I see. indeed if _all_ tests are ran in parallel, this will make our life miserable. However, as far as I know of `cargo` and `cargo-nextest` (https://nexte.st/docs/design/why-process-per-test/), it should be fine. I could still update a tiny bit the tests if you still believe that might matter later. Something using `#[cfg(test)]` on the line getting current directory, and then mocking it somehow. wdyt?

SomeoneToIgnore:
Yeah, if it's not a big change seems worth it.
Along the way, we could improve the error messages and maybe avoid the code that can potentially panic — that `unreachable!` bit fails on both IMO.

BuonOmo:
@SomeoneToIgnore it ends up being a bigger change than what I hope, since there are two methods to mock: `std::env::current_dir()` and `std::fs::canonicalize()`. I think there is a clean way to do it: replace usage of `std::fs` with our `fs` crate. But for that to be worth I think we should then normalize this usage across the `cli` crate, and stop using `std::fs`. This would help with testing for sure. But seeing how small and separated from the rest the `cli` crate is, IMHO it is not worth it.

With that said, it would still be a fun refactoring and I definitely could do it. But _just because peli can doesn't mean that peli should_. So I would leave that up to you and your wider knowledge of the app to decide!

SomeoneToIgnore:
Yeah, sounds like we'll just live with what we have now then, thank you for looking.

I really do not see any reason to bring in all async machinery into this crate, and it is so high-level that blocking FS operations from stdlib is quite ok to have.

PR #36848, opened 2025-08-24
gpui: Convert macOS clipboard file URLs to paths for paste

seantimm:
- On macOS, pasting now inserts the actual file path when the clipboard contains a file URL (public.file-url/public.url)
- Terminal paste remains text-only; no temp files or data URLs are created. If only raw image bytes exist on the clipboard, paste is a no-op.
- Scope: macOS only; no dependency changes.
- Added a test (test_file_url_converts_to_path) that verifies URL→path conversion using a unique pasteboard.

Release Notes:

- Improved pasting on macOS: now inserts the actual file path when the clipboard contains a file URL (enables image paste support for Claude Code)

PR #43254, opened 2025-11-21
Revert "gpui: Convert macOS clipboard file URLs to paths for paste"

smitbarmase:
Reverts zed-industries/zed#36848

Turns out this broke copying a screenshot from apps like CleanShot X and then pasting it over. We should land this again after taking a look at those cases. Pasting screenshots from the native macOS screenshot functionality works though. 

cc @seantimm 

Release Notes:

- Fixed issue where copying a screenshot from apps like CleanShot X into Agent Panel didn't work as expected.


morgankrey:
Thank you for the bug report @danchristian

PR #43269, opened 2025-11-21
git: Add word diff highlighting

Anthony-Eid:
This PR adds word/character diff for expanded diff hunks that have both a deleted and added section, as well as a setting `word_diff_enabled` to enable/disable word diffs per language.

- `word_diff_enabled`: Defaults to true. Whether or not expanded diff hunks will show word diff highlights when they're able to.

### Preview
image

### Architecture

I had three architecture goals I wanted to have when adding word diff support:

- Caching: We should only calculate word diffs once and save the result. This is because calculating word diffs can be expensive, and Zed should always be responsive.
- Don't block the main thread: Word diffs should be computed in the background to prevent hanging Zed.
- Lazy calculation: We should calculate word diffs for buffers that are not visible to a user.

To accomplish the three goals, word diffs are computed as a part of `BufferDiff` diff hunk processing because it happens on a background thread, is cached until the file is edited, and is only refreshed for open buffers. 

My original implementation calculated word diffs every frame in the Editor element. This had the benefit of lazy evaluation because it only calculated visible frames, but it didn't have caching for the calculations, and the code wasn't organized. Because the hunk calculations would happen in two separate places instead of just `BufferDiff`. Finally, it always happened on the main thread because it was during the `EditorElement` layout phase.

I used Zed's [`diff_internal`](https://github.com/zed-industries/zed/blob/02b2aa6c50c03d3005bec2effbc9f87161fbb1e8/crates/language/src/text_diff.rs#L230-L267) as a starting place for word diff calculations because it uses `Imara_diff` behind the scenes and already has language-specific support.

#### Future Improvements

In the future, we could add `AST` based word diff highlights, e.g. https://github.com/zed-industries/zed/pull/43691.

Release Notes:

- git: Show word diff highlight in expanded diff hunks with less than 5 lines.
- git: Add `word_diff_enabled` as a language setting that defaults to true.


crates/multi_buffer/src/multi_buffer.rs
```diff
@@ -3604,11 +3610,33 @@ impl MultiBufferSnapshot {
...
+            .then(|| {
+                let hunk_start_offset =
+                    Anchor::in_buffer(excerpt.id, excerpt.buffer_id, hunk.buffer_range.start)
+                        .to_offset(self);
+
+                hunk.base_word_diffs
+                    .iter()
+                    .map(|diff| {
+                        MultiBufferOffset(diff.start) + hunk_start_offset
+                            ..MultiBufferOffset(diff.end) + hunk_start_offset
```
Veykril: nit but this should work, reducing the need to wrap multi buffer offsets (which can end up in logic bugs)
```suggestion
                        hunk_start_offset + diff.start
                            ..hunk_start_offset + diff.end
```

Anthony-Eid: Thank you for catching this!

PR #43277, opened 2025-11-21
agent: Disable git pager to avoid hangs

odvcencio:
- Set PAGER='' and GIT_PAGER=cat for agent/terminal commands so pager configs (e.g. delta) don't hang tool output\n\nFixes #42943

Release Notes:
- Prevent git pager configs from hanging agent/terminal git commands by forcing PAGER and GIT_PAGER off.


Veykril (approved):

Veykril:
thanks!

PR #43279, opened 2025-11-21
git: Collapse nested paths before refreshing statuses

### LINKED ISSUE #35780: Zed consumes a lot of memory and CPU when opening `~/` or other large file trees
### Summary

SUMMARY_SENTENCE_HERE

### Description

After updating to Zed 0.198.2, Zed's CPU load has increased and it is consuming more energy.

Steps to reproduce:
1. Open Zed
2. Open about 5-10 source files in Zed
3. Wait

**Expected Behavior**: CPU load 5-10%
**Actual Behavior**: CPU load 50% over

OS: macOS 15.6
MAC: MacBook Pro 14′
Chip: Apple M3
Memory: 16 GB

Before: Zed 0.197.6
After: Zed 0.198.2

odvcencio:
- When background scanning reports many paths for the same repo, collapse nested paths so we only refresh git status at the shallowest level.
- Cuts repeated git status processes when opening huge directories (e.g. ~/), addressing runaway git spawns/CPU.
- Fixes #35780.

Release Notes:
- Reduce redundant git status invocations when scanning large trees by collapsing nested paths, mitigating runaway git processes for huge directories.

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -2691,6 +2691,16 @@ impl GitStore {
                         continue;
                     }
                     path_was_used[ix] = true;
+
+                    // Collapse nested paths so we only request a status refresh at the
+                    // shallowest level needed (important when scanning large trees).
+                    if entry
+                        .last()
+                        .is_some_and(|last_path: &RepoPath| repo_path.starts_with(last_path))
+                    {
+                        continue;
+                    }
+
                     entry.push(repo_path);
                 }
             }


kubkon:
Thanks for the PR! I've had a closer look at your changes and while they do indeed trim the list of paths we want to fetch git status for, it doesn't reduce the number of calls to `git-status` however. What is happening under-the-hood, all paths per repo are added to a single `git-status` invocation in https://github.com/zed-industries/zed/blob/37e4f7e9b54e1cc830ec45d7fb4d019edf9f47cd/crates/git/src/repository.rs#L1201 so performance-wise it doesn't matter if the invocation looks like `git status path/a/a path/a/b path/b/a` or `git status path` as it's still a single shell-out to `git` and the returned resultset is identical. I'll therefore go ahead and close the PR. If you feel I missed something, please let us know by commenting in this PR!

PR #43296, opened 2025-11-22
zed: Support opening welcome page on startup

tidely:
Closes #40100 #43158

I saw a great suggestion in #43158 of reusing the welcome page as an option when opening Zed. Preferably we would have  a dedicated startup page at some later point in time that could highlight other useful features such as remoting (or add it to the welcome page?).

Currently still on the todo list:

- [ ] Think about removing `Return to Setup` from welcome screen conditionally (feedback appreciated)
- [ ] Consider whether Welcome Page should be the default, a setting, used as a fallback or something else entirely, it likely should not be under "what to restore from a previous session"
- [ ] Docs

The PR will likely be decently large, I'm willing to pair with someone to get it merged

image
image

Release Notes:

- New setting to open a welcome page when launching Zed

### DIFF
--- a/crates/onboarding/src/onboarding.rs
+++ b/crates/onboarding/src/onboarding.rs
@@ -18,7 +18,7 @@ use ui::{
     WithScrollbar as _, prelude::*, rems_from_px,
 };
 use workspace::{
-    AppState, Workspace, WorkspaceId,
+    AppState, ItemHandle, Workspace, WorkspaceId,
     dock::DockPosition,
     item::{Item, ItemEvent},
     notifications::NotifyResultExt as _,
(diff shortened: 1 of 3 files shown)


VihaanReddyM:
Yes please add this feature it will give a nice vibe. I tried adding this but was a little hard for me because I don't really use rust

VihaanReddyM:
I’ve thought about it a bit more and here’s what I suggest:

* The **“Return to Setup”** button should be hidden when the welcome screen is shown after the initial setup flow. It only makes sense the first time.
* For the behavior of when to show the welcome screen, we could add a **new setting** that lets users choose between two modes:

  1. Show the welcome screen only when no workspace is open (current behavior).
  2. Always show the welcome screen on startup.

This setting should be exposed both in the settings file and the UI menu.
I tried implementing this logic earlier, but I got stuck on wiring it into the settings system.

Let me know if this direction makes sense.

VihaanReddyM:
I have just built your fork and there is a issue, When zed is first launched it is skipping the setup page completely so gotta work on that.

tidely:
> I have just built your fork and there is a issue, When zed is first launched it is skipping the setup page completely so gotta work on that.

Unless you deleted your main Zed installation, I think it will detect that it's not your first time installing Zed

tidely:
> * For the behavior of when to show the welcome screen, we could add a **new setting** that lets users choose between two modes:
>   
>   1. Show the welcome screen only when no workspace is open (current behavior).
>   2. Always show the welcome screen on startup.
> 
> This setting should be exposed both in the settings file and the UI menu. I tried implementing this logic earlier, but I got stuck on wiring it into the settings system.

I think the most intuitive solution would be to allow putting all startup options into any order you want. However that seems like something maybe not possible with the current settings UI?

tidely:
> * The **“Return to Setup”** button should be hidden when the welcome screen is shown after the initial setup flow. It only makes sense the first time.

Yeah I agree. Unfortunately that means we'd need to make changes to the serialization logic which persists welcome pages currently to allow saving whether this button exists (also needs to be backwards compatible). I wonder if it would be easier to create a totally new view or a specialized copy of the current welcome page. Although I think that would need some more insights from the Zed UI team.

VihaanReddyM:
> > * For the behavior of when to show the welcome screen, we could add a **new setting** that lets users choose between two modes:
> 
> >   
> 
> >   1. Show the welcome screen only when no workspace is open (current behavior).
> 
> >   2. Always show the welcome screen on startup.
> 
> > 
> 
> > This setting should be exposed both in the settings file and the UI menu. I tried implementing this logic earlier, but I got stuck on wiring it into the settings system.
> 
> 
> 
> I think the most intuitive solution would be to allow putting all startup options into any order you want. However that seems like something maybe not possible with the current settings UI?
> 
> 
> 
> 

Idk what I did was I added a dropdown.

franciskafyi:
Thank you so much @tidely. We agree this is needed but we actually prefer this other approach in here: https://github.com/zed-industries/zed/pull/44048. Here, the page is adapted slightly with a focus on Recent Projects. That makes more sense for Existing Users, I think.

danilo-leal:
Hey @tidely, thank you so much for kicking this off! We just merged https://github.com/zed-industries/zed/pull/44048 which introduced some extra things that I ended up prioritizing. With that, I'm closing this one in favor of that, but really appreciate you for taking this on! 🙏

PR #43299, opened 2025-11-22
gpui: Support 3D rendering

shi-yan:
I’m exploring the possibility of implementing a 3D application using GPUI. After reviewing the source code, it appears that GPUI maintains its own scene graph, and developers can only modify this graph, there’s no access to low-level 3D graphics APIs such as Metal, DirectX 12, or Vulkan.

Because of this limitation, creating a full 3D application is currently not feasible.

To experiment with a solution, I extended the existing video-rendering path to support 3D rendering. Since the video surface was hard-coded to use a YUV pixel format, I added support for RGBA as well.

I’ve also put together a proof-of-concept demonstrating this approach.

At the moment, this implementation works only on macOS.

Screenshot 2025-11-21 at 8 33 30 PM

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/Cargo.toml
+++ b/crates/gpui/Cargo.toml
@@ -262,6 +262,11 @@ bindgen = "0.71"
 cbindgen = { version = "0.28.0", default-features = false }
 naga.workspace = true
 
+[target.'cfg(target_os = "macos")'.dev-dependencies]
+metal.workspace = true
+core-video.workspace = true
+objc.workspace = true
+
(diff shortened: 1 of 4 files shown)


as-cii:
Thanks for putting this together, @shi-yan! 

I like the idea and appreciate the effort: unfortunately, it's not something we're ready to take on right now, so I will go ahead and close this. I could definitely see revisiting this down the road, so might reopen when the time is right. 🙇‍♂️

PR #43327, opened 2025-11-22
Improve Go language struct completion

diogox:
This PR improves the completion of structs in Go by conditionally including `{}` based on the buffer position. It follows the behavior of Jetbrains Goland.

A small example:

https://github.com/user-attachments/assets/b9046527-1de3-4e9c-a9d0-59dce69b3aad

More examples:

CleanShot 2025-11-22 at 13 42 45@2x

> The last example is also inline with Jetbrains Goland, and I didn't find an obvious way to improve upon it.

I'm not a Rust dev, so any input on improving the code is highly appreciated! :)

Release Notes:

- Added smart `{}` placement on Go struct completions.

### DIFF
--- a/crates/language/src/language.rs
+++ b/crates/language/src/language.rs
@@ -254,6 +254,17 @@ impl CachedLspAdapter {
         self.adapter.diagnostic_message_to_markdown(message)
     }
 
+    pub async fn preprocess_completions(
+        &self,
+        completion_items: &mut [lsp::CompletionItem],
+        buffer: &BufferSnapshot,
+        position: PointUtf16,
+    ) {
+        self.adapter
+            .preprocess_completions(completion_items, buffer, position)
+            .await
+    }
+
     pub async fn process_completions(&self, completion_items: &mut [lsp::CompletionItem]) {
         self.adapter.process_completions(completion_items).await
     }
@@ -336,6 +347,15 @@ pub trait LspAdapter: 'static + Send + Sync + DynLspInstaller {
     /// for the unnecessary diagnostics, so do not underline them.
     fn underline_diagnostic(&self, _diagnostic: &lsp::Diagnostic) -> bool {
         true
+    }
+
+    /// Pre-processes completions provided by the language server.
+    async fn preprocess_completions(
+        &self,
+        _: &mut [lsp::CompletionItem],
+        _: &BufferSnapshot,
+        _: PointUtf16,
+    ) {
     }
 
     /// Post-processes completions provided by the language server.
--- a/crates/languages/src/go.rs
+++ b/crates/languages/src/go.rs
@@ -6,7 +6,7 @@ use gpui::{App, AsyncApp, Task};
 use http_client::github::latest_github_release;
 pub use language::*;
 use language::{LanguageToolchainStore, LspAdapterDelegate, LspInstaller};
-use lsp::{CompletionItemKind, LanguageServerBinary, LanguageServerName, Position};
+use lsp::{CompletionItemKind, LanguageServerBinary, LanguageServerName};
 
 use regex::Regex;
 use serde_json::json;
@@ -214,7 +214,7 @@ impl LspAdapter for GoLspAdapter {
     ) {
         // Go back to the position the last character was written in.
         // Otherwise, we sometimes get 'block' for the new position.
-        let position = PointUtf16::new(position.row, position.column-1);
+        let position = PointUtf16::new(position.row, position.column - 1);
 
         let offset = buffer.point_utf16_to_offset(position);
         let mut inside_type_context = false;
@@ -227,7 +227,7 @@ impl LspAdapter for GoLspAdapter {
             let mut node = layer.node().descendant_for_byte_range(offset, offset);
 
             if node.is_none() {
-                return
+                return;
             }
 
             while let Some(n) = node {
@@ -248,7 +248,7 @@ impl LspAdapter for GoLspAdapter {
                     }
                     "block" => {
                         break;
-                    },
+                    }
                     _ => {}
                 }
                 node = n.parent();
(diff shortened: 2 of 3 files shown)


diogox:
@cla-bot check

diogox:
@cla-bot check

diogox:
One thing I just noticed is that Goland doesn't use a snippet to place the cursor between `{}`.

And `gopls` will automatically add `{}` with the cursor at the end on `return` statements, specifically.

I can remove the snippet to make it more consistent, though I feel like the snippet's behavior is more useful.

Any opinions on this?

SomeoneToIgnore:
What I do not understand entirely is why are we solving this issue in Zed.

When providing completions, language server is the one who holds the entire context, and it's capable of providing whatever snippets/completions/extra edits.

E.g. rust-analyzer (do notice the LSP response on the right with all options):

image

Overall, matching AST with 3rd party libraries, potentially containing errors, is not going to work well in general case (e.g. Rust macros) and adding another API method for this is not great?

So, cannot gopls fix that instead on its side?
Or, maybe it can provide those already and Zed does something wrong?

diogox:
I'm by no means an expert in the matter, so I can't say for sure, but to me it doesn't look like gopls supports it.

Goland has it though, so I can only assume they're doing something like this.

This would be the quickest way to have Zed compete with their DX. Little things like these have always kept me, and I'm sure others, locked in.

That said, I agree it doesn't make sense to have language specific logic in Zed core. I can only imagine you guys are planning on extracting the languages once the Extensions API is expanded. At which point this could also be its own extension.

SomeoneToIgnore:
> but to me it doesn't look like gopls supports it.

Yes, that is quite clear given this PR, the question is — why.
I would strongly recommend searching for an issue in their tracker and maybe creating one.

> quickest way to have Zed compete with their DX

We are eons away from Intellij-based `I`DEs and their semantics analysis.

True to `I`ntegrated spirit, they have their own semantic analysis and editor in the same codespace, thus able to do a lot of great things.

Meanwhile, Zed, Fleet, VSCode, Helix, NVim, and others relate on `language` servers for analysis of the `language` semantics.

Zed has tiny bits of that with tree-sitter queries but no way it can do anything similar as "complete the statement here".

So, there's no quick way "to have Zed compete" with anything like that, and starting it with odd methods that work only partially might be a bit premature.

> Extensions API is expanded

I would argue that extensions are not the best place to do that too.

When the completion proposals are calculated, `gopls` has the entire understanding of the current snippet, including the context to determine when to provide `{}` in completions.
I would start from there.

Meanwhile, this is not really the best approach and without knowing the other side's reasoning, let's not proceed here, thank you.

SomeoneToIgnore:
> Yes, that is quite clear given this PR

To note, there could still be an error in Zed's code that handles the completions: I'd check more gopls-based editors to see what they do.
Given nothing in our issue tracker so far and the fact that rust-analyzer is able to do this, sounds not very probable.

diogox:
I think enriching completions would be a quick-win and, as far as I can tell, there are no side-effects between Treesitter and gopls.

Regardless, I understand you guys would have to maintain it, so perfectly fine.

Thanks!

diogox:
@SomeoneToIgnore I think there may actually be some kind of bug here.

I had a look at VSCode and it does autocomplete `{}` in the appropriate places, so I had a look at gopls and saw the following sample response:

> {"jsonrpc":"2.0","result":{"isIncomplete":true,"items":[{"label":"demo","kind":3,"detail":"func(x Demo) Demo","preselect":true,"sortText":"00000","filterText":"demo","insertTextFormat":2,"textEdit":{"newText":"demo(${1:})","insert":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}},"replace":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}}}},{"label":"Demo{}","kind":6,"sortText":"00001","filterText":"Demo{}","insertTextFormat":2,"textEdit":{"newText":"Demo{$0\\}","insert":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}},"replace":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}}}},{"label":"Demo","kind":22,"detail":"struct{...}","sortText":"00002","filterText":"Demo","insertTextFormat":2,"textEdit":{"newText":"Demo","insert":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}},"replace":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}}}},{"label":"delete","kind":3,"detail":"func(m map[Type]Type1, [...]

As we can see, there's:

> {"label":"Demo{}","kind":6,"sortText":"00001","filterText":"Demo{}","insertTextFormat":2,"textEdit":{"newText":"Demo{$0\\}","insert":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}},"replace":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}}}}
That contains `{}`

That includes `{}`.

And:

> {"label":"Demo","kind":22,"detail":"struct{...}","sortText":"00002","filterText":"Demo","insertTextFormat":2,"textEdit":{"newText":"Demo","insert":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}},"replace":{"start":{"line":13,"character":8},"end":{"line":13,"character":10}}}}

That does not.

Yet, only the latter is showing up as an option in autocomplete.

I think it may be because of this logic:

CleanShot 2025-11-23 at 02 14 59@2x

If detail there is None, zip will not return the entry.

The code seems to have been written by @maxbrunsfeld, maybe he can shed some light?

I might be completely wrong on the cause, but seems something is off here(?)

SomeoneToIgnore:
Great find, we'd better show both and have a test for this.

filipwiech:
Hey @diogox, it seems like you've managed to identify the root cause of this issue, nice! 👍 Do you plan to follow up with a new fix? 🙂

diogox:
Hey @filipwiech,

Honestly, I had a look again shortly after and I'm not so sure I found an issue after all.

Either way, I'm short on time and won't be looking into this any longer.

I did raise an issue with [gopls](https://github.com/golang/go/issues/76626). But, judging by the similar issues that have gone ignored for many years, I don't think we'll see improvements in this regard.

I think we'll have to wait for Zed's extensions API to allow what I was trying to do here so that Goland's behavior can be replicated without merging to Zed main.

PR #43333, opened 2025-11-22
Properly check for success on xdg-portal open URI

eskimor:
Ensure we properly make sure opening the uri actually worked to ensure we always fallback if it does not.

Release Notes:

Fixed: When opening URIs on Linux (e.g. on Sign in) properly fallback to xdg-open in case the xdg portal failed. The fallback existed already, but was not always triggered as some errors had been ignored.


eskimor:
Already fixed on main.

PR #43359, opened 2025-11-23
editor: Fix tab tooltips not showing file path for remote files

errmayank:
Closes #42344

Release Notes:

- Fixed editor tab tooltips not showing file path for remote files

Here's the before/after, tested both local and remote:

https://github.com/user-attachments/assets/2768a0f8-e35b-4eff-aa95-d0decb51ec78


SomeoneToIgnore (approved):
Thank you!

PR #43363, opened 2025-11-23
Add global search keybinding from helix

godalming123:
In helix, `space /` activates a global search picker, so I think that it should be the same in zed's helix mode.

Release Notes:

- Added helix's `space /` keybinding to open a global search menu to zed's helix mode


godalming123:
@cla-bot check

PR #43368, opened 2025-11-23
Use buffer language when formatting with Prettier

JPGibb:
Set `prettier_parser` explicitly if the file extension for the buffer does not match a known one for the current language

Release Notes:

- N/A

### DIFF
--- a/crates/prettier/src/prettier.rs
+++ b/crates/prettier/src/prettier.rs
@@ -346,50 +346,11 @@ impl Prettier {
         ignore_dir: Option<PathBuf>,
         cx: &mut AsyncApp,
     ) -> anyhow::Result<Diff> {
-        let get_parser = |buffer_path: &Option<PathBuf>,
-                          buffer_language: &Option<&Arc<Language>>,
-                          prettier_settings: &PrettierSettings|
-         -> anyhow::Result<Option<String>> {
-            let parser = if buffer_path.is_none() {
-                let parser = prettier_settings.parser.as_deref().or_else(|| {
-                    buffer_language.and_then(|language| language.prettier_parser_name())
-                });
-                if parser.is_none() {
-                    log::error!(
-                        "Formatting unsaved file with prettier failed. No prettier parser configured for language {buffer_language:?}"
-                    );
-                    anyhow::bail!("Cannot determine prettier parser for unsaved file");
-                }
-                parser
-            } else if let (Some(buffer_language), Some(buffer_path)) =
-                (buffer_language, &buffer_path)
-                && buffer_path.extension().is_some_and(|extension| {
-                    !buffer_language
-                        .config()
-                        .matcher
-                        .path_suffixes
-                        .contains(&extension.to_string_lossy().into_owned())
-                })
-            {
-                buffer_language.prettier_parser_name()
-            } else {
-                prettier_settings.parser.as_deref()
-            };
-
-            let result = if let Some(parser) = parser {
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (commented):
Can you please provide more context on what's broken here, how to test the fix, etc.?

JPGibb:
Hi,

In my case I have json that gets stored in files like "feature.settings".
When editing a buffer without an associated file formatting works correctly as the prettier parser assiociated with the language is used.
However, after saving this buffer to a file like the on above, formatting would fail with the error:

**UndefinedParserError: No parser could be inferred for file {path_to_file}**

So for a file like 'feature.settings':
``` json
{
  "foo": "foo",
"bar": "misformatted"
}
```

Formatting would fail, even with the buffer language in Zed set to json.

With this change, the prettier parser would be set explicitly using the buffer's language rather than inferred from the file extension.
So the file would be correctly formatted as json.

Cheers

SomeoneToIgnore (commented):
Thank you for the context: sounds like this is quite a logic-related one, and we'd better have a test so it does not degrade later.

We have `async fn test_document_format_with_prettier` and other prettier tests, let's write another one for this scenario.

JPGibb:
Hi, Sorry for the delay getting back to you. Long week :).

I've added a new test `test_document_format_with_prettier_explicit_language` which passes.
However, I noticed after doing this that the code I wrote is in a Self::Real block whereas the test uses Self::Test.

I'm not entirely sure how to handle this, is there some way I can set my test up to run the code that I changed?

Thanks

SomeoneToIgnore (commented):
Thank you for the test, and no worries, I have other things to do anyway.

What if we adjust 
`let formatted_text = buffer.text() + FORMAT_SUFFIX;` to write the parser name too?

That will require amending a few existing tests and restructuring the code of `async fn format` so that the common code for parser determination is shared between both test and real branches.

But after that, the code before the change should output a totally different value, which we can assert.

JPGibb:
I've modified the test code to include the parser name and put the parser determination code in a closure used by both the real and test prettier code.

crates/prettier/src/prettier.rs
```diff
@@ -475,15 +507,15 @@ impl Prettier {
                         anyhow::Ok(FormatParams {
                             text: buffer.text(),
                             options: FormatOptions {
-                                parser: prettier_parser.map(ToOwned::to_owned),
+                                parser: prettier_parser.as_deref().map(ToOwned::to_owned),
```
SomeoneToIgnore:
```suggestion
                                parser: prettier_parser,
```

Is it all needed?
If so, can be just a `.clone()`.

crates/prettier/src/prettier.rs
```diff
@@ -345,6 +346,45 @@ impl Prettier {
         ignore_dir: Option<PathBuf>,
         cx: &mut AsyncApp,
     ) -> anyhow::Result<Diff> {
+        let get_parser = |buffer_path: &Option<PathBuf>,
```
SomeoneToIgnore: Could this be a method named `get_prettier_parser` or similar?

SomeoneToIgnore (approved):
Looking at the CI and the changes, this turned out to be the first test that uses Prettier parsers...

Thank you a lot for adding one.
I've left a few minor style nits and will be happy to merge this after today's release, just in case.

SomeoneToIgnore (approved):

PR #43370, opened 2025-11-23
agent_ui: Add support for deleting thread history

RemcoSmitsDev:
This PR adds support for deleting your entire thread history. This is inspired by a Zed user from the meetup in Amsterdam, he was missing this feature. 

**Demo**

https://github.com/user-attachments/assets/5a195007-1094-4ec6-902a-1b83db5ec508

Release Notes:

- AI: Add support for deleting your entire thread history


danilo-leal:
Sweet, that's an interesting feature; thanks for the PR! It feels worth to me to have a confirmation step, though, given how drastic this can be. So, the flow would be: you click on "Delete All" and then we pop a modal or something confirming and stating: "Do you really want to do this? You won't be able to recover the data later". I want to avoid the odds of users accidentally clicking on this and then getting support tickets to restore their data, which is something I'm candidly not even sure it's possible because this is all stored in the user's local database (we don't have access to anyone's threads).

RemcoSmitsDev:
Yeah that sounds good! Im open to pair on it to make the UX better if you want to. Otherweise i can pick this up later to finish its off

RemcoSmitsDev:
What we also could do is add a trash icon after each groep title inside the thead vieuwer. So more clear and you could remove a group of threads

danilo-leal:
I was wanting to remove the "Delete All" button out from the recent threads popover and add it on the bottom of the thread history view, and then clicking on it would trigger the confirmation modal. And yeah, totally down to pair, but feel free to take this on, too! Not sure how my availability will be this upcoming week 😅

RemcoSmitsDev:
Ah oke, do you mind if we come back on this when you have time to pair on this, or feel free to play around with it yourself.
I'm not quite seeing exactly where I would put the delete history button inside the recent threads view.

danilo-leal:
Alright, I think this is ready to go now! I pushed all of my changes and updated the demo video on the PR description. I intentionally made the red "Delete" button stay at the far right to force the user to move their cursor all the way there as an extra measure of safety, even though I also added the confirmation step.

RemcoSmitsDev:
Thanks for picking this up, it is looking great!

Issue #35417, opened 2025-07-31
Trashing paths from git panel doesn't work when a subdirectory of the repo is open


cole-miller:
### Summary

Trashing paths from the git panel doesn't work when you open a subdirectory of a git repo as a Zed worktree.

### Description

To reproduce, create an empty untracked file `foo` in the root of the Zed repo, open `zed/crates` in Zed, and try to trash `foo` from the git panel. Nothing will happen.

Probably we are implicitly assuming that the repo root and worktree root are the same when constructing the path to trash.

ayu-ch:
Hmm, it seems that it's not just about trashing, if you open Zed in a sub-directory, it is not able to open the file or diff from the git-panel as well. I also had a look at the code and the `GitStatusEntry` struct contains the field `repo_path` and I think this is where the issue resides. I will try to fix this.

ayu-ch:
Its much deeper than I expected, we are using this function called `repo_path_to_project_path()` but to resolve the path relative to the project worktree, it first calls the `find_worktree()` function and this searches multiple worktree within the project worktree. But since these files are outside the project worktree, we are never able to resolve their path.

esthertrapadoux:
Yay, thank you @ayu-ch !!

cole-miller:
Related: https://github.com/zed-industries/zed/pull/43392

esthertrapadoux:
Hey @ayu-ch ! Just checking in and wanted to see if you were up for pairing with the team on this issue this week. I can send through a cal.com link to get something booked.

ayu-ch:
Hey, sorry was a little busy. I will let you know soon once I catch up again. Thanks a lot for checking in!

esthertrapadoux:
Thanks @ayu-ch ! The Zed team will go on holiday break from Dec 20, 2025 - Jan 4, 2026. If you'd like to set up a pairing session this week, then there are still a few slots open! https://cal.com/esther-trapadoux-zed/30min

If you don't see any that work for you, just email me at esther at zed dot dev and I'll open a few more up in your time zone. Thank you!

PR #43392, opened 2025-11-24
git_ui: Fix tracked files inconsistency

### LINKED ISSUE #29670: Git: tracked files inconsistency
### Summary

When opening Zed from within an inner directory (e.g., 'backend'), the Git 'Tracked' view shows changes from the outer directory in a list view, but these changes do not appear in the 'Uncommitted Changes' view.

### Description
```
mkdir /tmp/29670
cd $_
git init
touch file.txt
mkdir backend
touch backend/file.txt
zed backend/
```

Actual Behavior: 
When you click on the Git view, tracked files from both the root and inner directories are shown in the list view, but these files are missing in the "Uncommitted Changes" view.
Image

Expected Behavior: The files visible in the "Uncommitted Changes" view should be the same as those in the list view.

Zed: v0.183.13 (Zed) 
OS: Linux Wayland fedora 42
Memory: 30.7 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 25.0.4

flashus:
## Release Notes:
- Fixed inconsistency where the git panel showed all repository files when Zed was launched from a subdirectory, while the uncommitted changes view correctly filtered to worktree files only.

## Changes:
- Added filtering in git_panel.rs to skip entries outside worktree boundaries by checking if repo_path_to_project_path() returns None.
- Updated test `test_entry_worktree_paths` to only expect changes within workspace folder.


GitGlimpse895:
Hey @cole-miller — you've been assigned to this since November and even mentioned it in issue #35417 in December, but there's been no review activity on the PR itself in 5 months.

I checked current `main` — `update_visible_entries` in `git_panel.rs` still has no worktree boundary filtering, so the bug (git panel showing all repo files when launched from a subdirectory) is still live. The fix is 7 lines, follows the same `repo_path_to_project_path()` pattern already used in other action handlers in the same file.

Could this be reviewed or reassigned? It now has merge conflicts from sitting too long.

smitbarmase:
Thanks for working on this, and sorry it took us a while to review.

However, I don't think filtering changes outside the opened worktree is the right fix. When a subdirectory of a repository is opened, Git status, the index, conflicts, and commits still apply to the whole repository. With this change, the panel could show `No Changes` while the repository still has changes. More importantly, actions such as `Stage All` and `Commit` could affect files that are hidden from the panel.

The issue, #29670, was also closed as not planned after receiving no further activity. VS Code similarly keeps Git handling repository-scoped when a subdirectory is opened. Supporting that behavior safely would require a broader product and implementation design, and I don't think we are gonna address that right at the moment.

I'm going to close this PR as not planned. Thanks again.

PR #43403, opened 2025-11-24
Rework and consolidate issue templates

yeskunall:
We’re reworking our triage process and in doing so, reworking our issue templates is worth looking into. We have multiple issue templates, for arbitrary categories, and not enough enforcement. The plan is to consolidate the issue templates (maybe all into one) and drop the others.

Release Notes:

- N/A


injust:
This caused https://github.com/zed-industries/zed/issues/43892 because the bug report was renamed without updating it in Zed.

PR #43431, opened 2025-11-24
Disable `phptools` language server by default for PHP

maxdeviant:
This PR makes it so we disable the `phptools` language server by default for PHP.

The default language server is `phpactor` on macOS and Linux and `intelephense` on Windows (as `phpactor` does not support Windows), so we don't want to run `phptools` as well.

Release Notes:

- Disabled the `phptools` language server by default for PHP.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1889,7 +1889,7 @@
       }
     },
     "PHP": {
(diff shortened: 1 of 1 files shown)


maxdeviant:
Closing in favor of https://github.com/zed-industries/zed/pull/43432.

PR #43434, opened 2025-11-24
git: Mark entries as pending when staging a files making the staged highlighting more "optimistic"

### LINKED ISSUE #26870: git: stage file obvious delay
### Summary

There is no obvious delay in the stage block, but it will appear when the stage file is loaded.

Steps to trigger the problem:
1. open project diff
2. stage any file or stage all file

Actual Behavior:

https://github.com/user-attachments/assets/a7127b57-d4bf-4b67-b6f9-d65c4f96e6c8

Expected Behavior:

Zed: v0.179.0 (Zed Dev d5bb12631a4798f35c2a1a1dc721d2c9337acd68) (Taylor's Version)
OS: macOS 15.3.1
Memory: 32 GiB
Architecture: aarch64

van-sprundel:
This at least speeds it up, not sure if this would close the issue

On main (342eba6f220625c015d00334c6bc354f0e2c52e1):

https://github.com/user-attachments/assets/55d10187-b4e6-410d-9002-06509e8015c9

This branch:

https://github.com/user-attachments/assets/e9a5c14f-9694-4321-a81c-88d6f62fb342

Closes #26870

Release Notes:

- Added optimistic staged hunk updating

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -1592,6 +1592,59 @@ impl GitStore {
         }
     }
 
+    fn mark_entries_pending_by_project_paths(
+        &mut self,
+        project_paths: &[ProjectPath],
+        stage: bool,
+        cx: &mut Context<Self>,
+    ) {
+        let buffer_store = &self.buffer_store;
+
+        for project_path in project_paths {
+            let Some(buffer) = buffer_store.read(cx).get_by_path(project_path) else {
+                continue;
+            };
+
+            let buffer_id = buffer.read(cx).remote_id();
+            let Some(diff_state) = self.diffs.get(&buffer_id) else {
+                continue;
+            };
+
+            diff_state.update(cx, |diff_state, cx| {
+                let Some(uncommitted_diff) = diff_state.uncommitted_diff() else {
+                    return;
+                };
+
+                let buffer_snapshot = buffer.read(cx).text_snapshot();
+                let file_exists = buffer
+                    .read(cx)
+                    .file()
+                    .is_some_and(|file| file.disk_state().exists());
+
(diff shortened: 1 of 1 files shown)


van-sprundel:
@cla-bot check

van-sprundel:
@cla-bot check

cole-miller:
Thanks!

+1 for making this more optimistic, and I think the basic approach is right, but we can streamline the implementation a bit. The goal is to end up with a set of `pending_hunks` that has the same hunk ranges as the uncommitted diff, and where the `new_status` for each hunk is `DiffHunkSecondaryStatus::SecondaryHunkRemovalPending` (staging) or `DiffHunkSecondaryStatus::SecondaryHunkAdditionPending` (unstaging). We should be able to add a method on `BufferDiff` that just sets the pending hunks to this state, which should have the same effect as the `stage_or_unstage_hunks` call and be a bit more efficient. Would you be able to take a crack at implementing that?

van-sprundel:
Will take a look at that later today!

cole-miller:
Hi @van-sprundel, thanks for updating the PR!

I realized that the refactoring I suggested introduces a bug, where the staged status of the hunks flickers if you click the checkbox multiple times in quick succession (this is due to losing the `cx.emit(BufferDiffEvent::HunksStagedOrUnstaged { .. }` in `fn stage_or_unstage_hunks`. I think we should go back to your original approach, sorry for the confusion! Other than that I think this is good to merge.

van-sprundel:
I did notice that when you spam it, the visualization lags behind sometimes. Wondering why this makes such a difference because I wouldn't expect it to be slower

van-sprundel:
Reverted the commit :)

cole-miller:
Thanks!

PR #43441, opened 2025-11-25
Add transcription pipeline

dzervas:
This is more of a PoC/discussion opener on how zed will integrate transcription/dictation

I've implemented a transcription pipeline that spits what it heard to stdout using a local whisper model (base model).

It's off by default for obvious reasons and you have to press the mic icon at the right side of the sidebar to start transcribing. An action can be used as well

I also implemented an inline assistant "bridge" that activates it as soon as something gets transcribed but it got annoying very fast, especially since there's no notion of history or threads.

I'm using rodio to capture audio with code loosely resembling the audio crate but since the transcriber is spawned in its own thread and there's no access to the app context, I didn't want to get my hands too dirty to compensate for that. Also the whisper model deeply dislikes denoising for some reason - maybe I need to tune the other parameters more?

I tried to create an easy to use API with GPUI subscriptions so that it can be used throughout the editor

What's next and my personal goals:

This whole idea started with me wanting to converse with my codebase, not vibe code. You ask a small LLM something while you program to help you out, rather than asking the agent to implement stuff - e.g. "what does this do" while your cursor is on a function or code chunk would answer that question with that context in mind. This featureset however is humongous - needs a new implementation of AI threads that probably the user doesn't see or somehow differentiated enough from the actual agent threads and TTS. And context management - both regarding what to send and keeping track/clearing it.

Another idea that popped up was to add a microphone button or action on all Editor contexts to enable transcription for that input only. That's much more manageable but I imagine that I'll end up very deep into the input handling of zed or even GPUI

Lastly a "Hey Zed" wake-word and action execution wouldn't be too hard to implement. You could do arbitrary actions or define aliases so that "Hey Zed open settings" executes the correct action to open the settings panel - that could then be chained to the agent/inline agent/etc.

Regarding this PR, if it's considered good enough of a starting point I need to polish it much more - tests, proper settings with a small preview, automated model downloading and more

Note regarding the accuracy of whisper: While not a native speaker I have a "good enough" accent and it's amazing how good and accurate it is. Apart from mixing "Hi/Bye" and other small silly mistakes that I think that LLMs won't even care about, it's rock solid

Release Notes:

- N/A


nathansobo:
This is really cool work, and I'm genuinely excited about voice integration in Zed - the ideas you laid out around conversing with your codebase and voice-driven commands resonate with me.

That said, I don't think the timing is right to bring this in as a contributed feature. We need to get our overall agentic editing experience in a better place before we're ready to give this the focus it deserves. Voice input touches so many surfaces (where does text go? how does it interact with the assistant? context management?) that I'd want us to be more settled on those foundations first.

I'm going to close this for now, but I'd encourage you to keep an eye on where we land with the agent experience. This is something I'd love to revisit down the road.

Thanks for putting this together - the core pipeline you've built here is solid.

nathansobo:
I'm having second thoughts. I want to leave this open. Sorry. Let's talk about next steps. Are you down to work on it more? I may need time to engage but would love to land this if you want to help.

dzervas:
@nathansobo yea I'd love to move this forward!

shoot me an email if you want to arrange some time to discuss :)

PR #43442, opened 2025-11-25
extensions: Don't recompile Tree-sitter parsers when up-to-date

kdkasad:
When installing local "dev" extensions that provide tree-sitter grammars, compiling the parser can be quite intensive[^anecdote]. This PR changes the logic to only compile the parser if the WASM object doesn't exist or the source files are newer than the object (just like `make(1)` would do).

[^anecdote]: The tree-sitter parser for LLVM IR takes >10 minutes to compile and uses 20 GB of memory on my laptop.

Release Notes:

- N/A

### DIFF
--- a/crates/extension/src/extension_builder.rs
+++ b/crates/extension/src/extension_builder.rs
@@ -247,26 +247,32 @@ impl ExtensionBuilder {
         let parser_path = src_path.join("parser.c");
         let scanner_path = src_path.join("scanner.c");
 
-        log::info!("compiling {grammar_name} parser");
-        let clang_output = util::command::new_smol_command(&clang_path)
-            .args(["-fPIC", "-shared", "-Os"])
-            .arg(format!("-Wl,--export=tree_sitter_{grammar_name}"))
-            .arg("-o")
-            .arg(&grammar_wasm_path)
-            .arg("-I")
-            .arg(&src_path)
-            .arg(&parser_path)
-            .args(scanner_path.exists().then_some(scanner_path))
-            .output()
-            .await
-            .context("failed to run clang")?;
+        // Skip recompiling if the WASM object is already newer than the source files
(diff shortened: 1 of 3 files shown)


kdkasad:
@cla-bot check

crates/extension/src/extension_builder.rs
```diff
@@ -247,26 +247,32 @@ impl ExtensionBuilder {
...
-            .arg(&src_path)
-            .arg(&parser_path)
-            .args(scanner_path.exists().then_some(scanner_path))
-            .output()
-            .await
-            .context("failed to run clang")?;
+        // Skip recompiling if the WASM object is already newer than the source files
+        if file_newer_than_deps(&grammar_wasm_path, &[&parser_path, &scanner_path]).unwrap_or(false)
+        {
+            log::info!("skipping compile of {grammar_name} parser because it is up to date");
```
MrSubidubi:
```suggestion
            log::info!("skipping compilation of {grammar_name} parser because the existing compiled grammar is up to date");
```

MrSubidubi (approved):
Looks good to me and like a good improvement - thank you!

crates/extension/src/extension_builder.rs
```diff
@@ -247,26 +247,32 @@ impl ExtensionBuilder {
...
-            .arg(&src_path)
-            .arg(&parser_path)
-            .args(scanner_path.exists().then_some(scanner_path))
-            .output()
-            .await
-            .context("failed to run clang")?;
+        // Skip recompiling if the WASM object is already newer than the source files
+        if file_newer_than_deps(&grammar_wasm_path, &[&parser_path, &scanner_path]).unwrap_or(false)
+        {
+            log::info!("skipping compilation of {grammar_name} parser because the existing compiled grammar is up to date");
```
MrSubidubi:
```suggestion
            log::info!(
                "skipping compilation of {grammar_name} parser because the existing compiled grammar is up to date"
            );
```

PR #43452, opened 2025-11-25
gpui: Fix incorrect text wrap when line end has spaces

huacnlee:
Release Notes:

- Fixed incorrect text wrap when line end has spaces.

image

Resize details after change:

https://github.com/user-attachments/assets/f5f2791b-6288-4ae2-af4c-31bf4b4431b6

---

I have also tested CJK content by run `text_wrapper` example and Zed, they are correct.

### How to test

Use this text and enable soft wrap in Zed, and resize window.

> NOTE: In this text, last have a lot of space.

```
    GPUI is still in active development as we work on the Zed code editor,                   and is still pre-1.0.                           There will often be breaking changes between versions. You'll also need to use the latest version of stable Rust and be on macOS or Linux.                                                                                                                       
```

### DIFF
--- a/crates/gpui/src/text_system/line_wrapper.rs
+++ b/crates/gpui/src/text_system/line_wrapper.rs
@@ -58,19 +58,27 @@ impl LineWrapper {
                             continue;
                         }
 
+                        let mut candidate = false;
                         if Self::is_word_char(c) {
                             if prev_c == ' ' && c != ' ' && first_non_whitespace_ix.is_some() {
-                                last_candidate_ix = ix;
-                                last_candidate_width = width;
+                                candidate = true;
                             }
                         } else {
-                            if c == ' ' || first_non_whitespace_ix.is_some() {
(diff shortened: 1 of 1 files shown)


crates/gpui/src/text_system/line_wrapper.rs
```diff
@@ -346,43 +358,56 @@ mod tests {
 
     #[test]
     fn test_wrap_line() {
+        // NOTE:
+        // In this case (Used .ZedMono and 16px text size),
+        // 1 char is `9.6px` width, so 72px can fit 7 chars.
+        const WRAP_WIDTH: Pixels = px(72.);
```
huacnlee, line 360: Add note to understand tests when next time we change the rules.

huacnlee:
My initial solution had some issues, and my testing revealed that it caused blank spaces at the beginning of lines.

Zed's editor test cases previously covered this case, which is necessary. However, this conflicts with the solution for fixing trailing blank spaces, and I've tried many times without being able to find a perfect solution.

PR #43455, opened 2025-11-25
Improve bash detection on Windows

qystishere:
I have git installed via [scoop](https://scoop.sh). The current implementation finds `git.exe` in scoop's shims folder and then tries to find `bash.exe` relative to it.

For example, `git.exe` (shim) is located at:
```
C:\Users\<username>\scoop\shims\git.exe
```

And the code tries to find `bash.exe` at:
```
C:\Users\<username>\scoop\shims\..\bin\bash.exe
```
which doesn't exist.

This PR changes the logic to first check if `bash.exe` is available in PATH (using `which::which`), and only falls back to the git-relative path if that fails.

### DIFF
--- a/crates/util/src/shell.rs
+++ b/crates/util/src/shell.rs
@@ -79,7 +79,7 @@ pub fn get_default_system_shell() -> String {
     }
 }
 
-/// Get the default system shell, preferring git-bash on Windows.
+/// Get the default system shell, preferring bash on Windows.
 pub fn get_default_system_shell_preferring_bash() -> String {
     if cfg!(windows) {
         get_windows_bash().unwrap_or_else(|| get_windows_system_shell())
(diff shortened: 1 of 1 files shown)


qystishere:
@cla-bot check

Veykril:
I don't think this will work, if you install a wsl distro (ubuntu for example) via the microsoft store, windows will push a bash.exe into your path that spawns a wsl session

qystishere:
> I don't think this will work, if you install a wsl distro (ubuntu for example) via the microsoft store, windows will push a bash.exe into your path that spawns a wsl session

@Veykril Yeah, you're right, I've updated it - now it checks specific paths (Scoop, Git) instead of PATH, similar to pwsh detection. What about this?

Veykril (approved):
Thanks!

PR #43457, opened 2025-11-25
miniprofiler_ui: Improve MiniProfiler to use uniform list

huacnlee:
Release Notes:

- N/A

---

- Apply uniform_list for timing list for performance.
- Add paddings for window.
- Add space to `ms`, before: `100ms` after `100 ms`.

## Before 

image

## After 

image


huacnlee:
Ready for review.

localcc (changes requested):
looks good overall! please make it so that the timing millisecond text does not affect the timing bar size

huacnlee:
Added more 10px for duration column.

image

localcc (approved):

PR #43462, opened 2025-11-25
git panel: Truncate file paths from the left

ddoemonn:
https://github.com/user-attachments/assets/758e1ec9-6c34-4e13-b605-cf00c18ca16f

Release Notes:

- Improved: Git panel now truncates long file paths from the left, showing "…path/filename" when space is limited, keeping filenames always visible. 

@cole-miller @mattermill


ddoemonn:
Are there any updates on this?

danilo-leal:
This is awesome! Is there a world, though, where we can make this left-truncation an option in the `Label` truncate method, instead of a completely different component? I guess I'd love to have `Label::new("hey").truncate(ui::LabelTruncation::Left)`, like an enum.

danilo-leal (approved):
Thanks so much for the patience here! Now that we have the `truncate_start` method straight from GPUI, we can go ahead :) Thanks again!

PR #43469, opened 2025-11-25
languages: Recognize .clang-format as YAML

notpeter:
Clang-Format uses uses a YAML config file format.

Use YAML language by default for `.clang-format` and `_clang-format` filenames. ([source](https://clang.llvm.org/docs/ClangFormatStyleOptions.html))
Add `#yaml-language-server: $schema` to `.clang-format` example in C language docs.

Release Notes:

- Added support for identifying. `.clang-format` files as YAML by default


SomeoneToIgnore (approved):
👍

WeetHet:
Should probably do one for `.clangd` as well? It's also yaml and I would argue that it's even more widely used

PR #43478, opened 2025-11-25
git: Preserve whitespaces in commit messages

### LINKED ISSUE #42476: Git: newlines in commit message gets removed when using bullet points
### Summary

If a commit message  contains bullet points, the git editor shows them correctly when 
typing it, but the final commit message gets reformatted to not have newlines around it.

Image

Image

**Expected Behavior**:
1. Commit message should contain bullet points with newlines as typed.

**Actual Behavior**:
1. Newlines are removed, and all the points appear as one line.

Note:
This could be due to the use of [rewrap](https://github.com/zed-industries/zed/blob/main/crates/git_ui/src/git_panel.rs#L1541), which is invoked when the "Commit"/"Amend" button is clicked.

Zed: v0.211.6 (Zed) 
Memory: 27.1 GiB
Architecture: x86_64
GPU: AMD Radeon 780M Graphics (RADV PHOENIX) || radv || Mesa 25.2.3-1ubuntu1

_No response_

tartarughina:
Closes #42476

Release Notes:

- Preserve whitespaces in git commit messages

**What Changed:**
- Modified commit message rewrapping behavior to preserve existing whitespace formatting
- Added new `commit_rewrap` method to the Editor that maintains intentional spacing in commit messages

**Technical Details:**
- Introduced `commit_rewrap()` method in `editor.rs` with `preserve_existing_whitespace: true` option
- Updated Git panel to use `commit_rewrap()` instead of standard `rewrap()` when formatting commit messages
- This ensures that deliberate formatting choices (like blank lines, indentation, and spacing) in commit messages are
maintained during the rewrap operation

**Impact:**
- Users can now format their commit messages with intentional whitespace without it being removed during the rewrap process
- Improves commit message readability and preserves structured formatting (e.g., bullet points, paragraphs, code blocks)

**Files Modified:**
- `crates/editor/src/editor.rs` - Added `commit_rewrap()` method
- `crates/git_ui/src/git_panel.rs` - Updated to use new rewrap method

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -12290,6 +12290,22 @@ impl Editor {
         self.rewrap_impl(RewrapOptions::default(), cx)
     }
 
+    pub fn commit_rewrap(&mut self, _: &Rewrap, _: &mut Window, cx: &mut Context<Self>) {
+        self.hide_mouse_cursor(HideMouseCursorOrigin::TypingAction, cx);
+        if self.mode.is_single_line() {
+            cx.propagate();
+            return;
+        }
+
+        self.rewrap_impl(
+            RewrapOptions {
+                override_language_settings: false,
+                preserve_existing_whitespace: true,
+            },
+            cx,
+        );
+    }
+
     pub fn rewrap_impl(&mut self, options: RewrapOptions, cx: &mut Context<Self>) {
         let buffer = self.buffer.read(cx).snapshot(cx);
         let selections = self.selections.all::<Point>(&self.display_snapshot(cx));
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -1634,7 +1634,7 @@ impl GitPanel {
         let editor = cx.new(|cx| Editor::for_buffer(buffer, None, window, cx));
         let wrapped_message = editor.update(cx, |editor, cx| {
             editor.select_all(&Default::default(), window, cx);
-            editor.rewrap(&Default::default(), window, cx);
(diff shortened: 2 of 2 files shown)


cole-miller:
Thanks for opening a PR! However, I think we might want to try a different approach to this. Note that re: #42476 we already have support for "rewrap prefixes" that support things like markdown lists (added in https://github.com/zed-industries/zed/pull/33702). These prefixes should be respected by the normal editor rewrap, without that `preserve_existing_whitespace` option, and I believe we configure the same rewrap prefixes for the git commit language as for markdown. So it would be good to figure out why that configuration is not being respected.

tartarughina:
Likely the buffer for the commit is not getting recognized neither as Markdown nor as Git Commit. Before creating this PR, I've tried playing around with the settings for Git Commit language without any success.

tartarughina:
I've further tried to play around with the settings. A very foolproof setting I tried was to enable display whitespaces, but nothing happened in the git commit buffer. This proves that the language settings, as for git commit language as for markdown, are not even getting applied, as the buffer is not recognized at all as a language.

Furthermore, the whitespaces are preserved yes, but only in instances where multiple `\n` are present:

For example

```
This is my commit
message as I see it
in the buffer
```

as soon as I press the commit button becomes

```
This is my commit message as I see it in the buffer
```

but if I introduce multiple `\n`

```
This is my commit

message as I see it

in the buffer
```

pressing commit maintains the format

```
This is my commit

message as I see it

in the buffer
```

It definitely deserves further investigation, as this bug undoubtedly disrupts the workflow of people accustomed to providing commit messages spanning multiple lines with bullet points/lists.

PR #43479, opened 2025-11-25
Add WSL Linux choice and settings.json prompt for GitHub issue template

miguelraz:
Release Notes:

- N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/1.bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1.bug-report.yml
@@ -48,6 +48,22 @@ body:
       required: false
   - type: textarea
     attributes:
+      label: If applicable, attach your relevant Zed settings fields to this issue.
+      description: |
+        Open the command palette in Zed, then type `zed: open settings` and copy/paste any relevant (e.g., LSP specific) settings.
+      value: |
+        <details><summary>settings.json</summary>
+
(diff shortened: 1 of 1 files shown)


yeskunall (dismissed):
GitHub is complaining about a `key` here: https://github.com/zed-industries/zed/blob/settings-template/.github/ISSUE_TEMPLATE/1.bug-report.yml

We’ll need to fix that.

.github/ISSUE_TEMPLATE/1.bug-report.yml
```diff
@@ -46,6 +46,22 @@ body:
         </details>
     validations:
       required: false
+  - type: textarea
+    attributes:
+      label: If applicable, attach your relevant Zed settings fields to this issue.
+      description: |
+        Open the command palette in Zed, then type `zed: open settings` and copy/paste any relevant (e.g., LSP specific) settings.
```
Veykril: Does the gui allow you to copy? I think this should state `zed: open settings file`

yeskunall: You’re right. Fixed.

zelenenka (approved):
on a conceptual level lgtm, i didn't check if it compiles or whatnot but i'm sure Kunall is on top of that and every possible typo :)

PR #43501, opened 2025-11-25
Consolidate event propagation handling by replacing Window::prevent_default with App::stop_propagation

### LINKED ISSUE #42762: gpui: Consolidate event control APIs
### Summary

`App::stop_propogation` and `Window::prevent_default()` seem like they're the same thing. Can we consolidate these into a single API?

kevinru2023:
Consolidate event propagation handling by replacing Window::prevent_default with App::stop_propagation

In app.rs `stop_propagation()` would stop dispatching to event handlers with a lower z-index (mouse) or higher in the tree (keyboard) whilst `prevent_default` would simply prevent the default action of an event and was only being used to prevent parent elements from being focused on mouse down. Since `prevent_default` was only being focused on mouse down, it was essentially redundant code, since it's logic was entirely the same as `stop_propagation`. However, `prevent_default` also had additional functionality to change a variable called `self.default_prevented` (to true) that would allow developers to keep track if "default" has been prevented for the event currently being dispatched which was done by the method `default_prevented`. This proposed re factorization removes the need for the two methods mentioned in `crates/gpui/src/window.rs`, by adding an additional method in  `crates/gpui/src/app.rs`, that would allow developers to keep track if events have stopped being propagated (i.e. if the default has been prevented).

Closes #42762

Release Notes:

- GPUI: Removed redundant `Window::prevent_default()` API in favor of existing `App::stop_propagation()`


kevinru2023:
Looking at the tests here looks like I would also have to change some documentation (didn't think of that), but I would first like to know what others think of the proposed solution.

kevinru2023:
Looks like I might have missed some references, will look into where I might have missed some calls to prevent_default.

mikayla-maki:
Does prevent_default also have some behavior in platform APIs?

kevinru2023:
I do not believe so, from what I seen it was mainly used in order to just be able to keep track of the `default_prevented` variable in `window.rs` and never seemed to be used in the platform api (assuming by platform api you are talking about app.rs). That's why I believe adding the additional method in app.rs will essentially negate the need of the `default_prevented` var (sorry if I am using the wrong lang here, I am not just new to the codebase, but also rust!). It seems the functionality from both the `default_prevented` var and `prevent_default` were essentially wrappers around the current logic we had in platform api for propagating events.

Again, I am really new to all this so if you believe this is the wrong approach feel free to let me know.

kevinru2023:
Also currently working on fixing one last bug from a recently made change in the `recent_projects` crate, and then should be able to push a working commit (i.e. passes test cases). However, I believe if we did move forward with this change it should def be documented as I would not want others to be confused if they tried to use `prevent_default` or `default_prevented`.

PR #43514, opened 2025-11-26
Exclude tracked files from gitignore status

### LINKED ISSUE #23224: Exclude tracked files from gitignored entries so current git status can be displayed
### Check for existing issues

- [x] Completed

### Describe the feature

When you ignore alle files with .gitignore:
```
*
```
All files are ignored, even files explicitly added to git:
```
$ git ls-tree --full-tree --name-only -r HEAD
dunst/dunstrc
i3/config
i3/script/next_or_new.sh
i3/script/prev_or_first.sh
i3/script/toggle_scaling.sh
ngrok/ngrok.yml
pulse/cookie
redshift.conf
xkb/symbols/mod3
zed/keymap.json
zed/settings.json
```
This prevents `auto_reveal_entries` from working and all tracked files are grayed out the project panel so you can't see the current git status.

Zed: v0.169.2 (Zed)
OS: Linux X11 ubuntu 24.10
Memory: 13.5 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 24.2.3-1ubuntu1

Current behaviour:
Image

Expected behaviour:
Image
(Faked by adding all tracked files and directories as excluded in .gitignore)

mdliss:
Fixes #23224

## Problem
When a `.gitignore` file contains `*` (ignore all), Zed incorrectly treats ALL files as ignored, including files explicitly tracked by git (added via `git add -f`). This causes:
- Tracked files to be grayed out in the project panel
- `auto_reveal_entries` to not work for tracked files
- Git status indicators to not display for tracked files

## Solution
Modified the ignore stack logic to check if files are tracked by git before marking them as ignored. Git-tracked files should never be treated as ignored, matching git's actual behavior where `.gitignore` only affects untracked files.

## Changes
- Added `tracked_paths` field to `IgnoreStack` struct
- Modified `is_abs_path_ignored()` to check if a file is tracked before returning true
- Added `get_tracked_files()` function that runs `git ls-files` to get tracked files
- Added test `test_tracked_files_not_ignored` to verify the fix

## Testing
Tested with a repo containing `.gitignore` with `*` and force-added tracked files - tracked files now display normally while untracked files are correctly grayed out.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,17 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(/opt/homebrew/bin/cargo check -p worktree)",
+      "Bash(export PATH=\"$HOME/.cargo/bin:$PATH:/usr/bin:/bin\")",
+      "Bash(cargo check:*)",
+      "Bash(cargo test:*)",
+      "Bash(/opt/homebrew/bin/cargo test:*)",
+      "Bash(echo $HOME/.cargo/bin/cargo)",
+      "Bash($HOME/.cargo/bin/cargo test:*)",
+      "Bash(~/.cargo/bin/cargo test:*)",
+      "Bash(RUST_LOG=trace ~/.cargo/bin/cargo test:*)"
+    ],
+    "deny": [],
+    "ask": []
+  }
+}
--- a/crates/worktree/src/ignore.rs
+++ b/crates/worktree/src/ignore.rs
@@ -121,5 +121,4 @@ impl IgnoreStack {
             },
         }
     }
-
 }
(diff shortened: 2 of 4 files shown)


mdliss:
@cla-bot check

kubkon:
Thanks for the PR! After careful examination we don't think this is the right approach. Firstly, since the background scanner is run with some cadence, the computed snapshot of tracked files using `git ls-files` will likely get out-of-sync with the actual repo status very quickly. Secondly, I don't think persisting the snapshot of tracked files in the background scanner is required or needed. Anyhow, since this is a tricky issue to fix, I will go ahead and close this PR as we really should explore the space of potential solutions to this problem that may require overhauling bigger chunks of worktree and/or project/git_store abstractions.

PR #43521, opened 2025-11-26
editor: Fix vertical scroll margin not accounting for file header height

errmayank:
Closes #43178

Release Notes:

- Fixed vertical scroll margin not accounting for file header height

Here's the before/after:

With `{ "vertical_scroll_margin": 0 }` in `~/.config/zed/settings.json`

https://github.com/user-attachments/assets/418c6d7f-de0f-4da6-a038-69927b1b8b88

### DIFF
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -28198,3 +28198,87 @@ async fn test_multibuffer_selections_with_folding(cx: &mut TestAppContext) {
         3
         "});
 }
+
+#[gpui::test]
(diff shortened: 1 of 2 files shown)


MrSubidubi (commented):
Nice find, nice fix, thank you. 

One last thing: Could you look into adding a test for this fix? There are some more bugs to fix in this area (sadly), and I would not want this to regress unknowingly when fixing the other ones.

errmayank:
Added a test case for this :)

MrSubidubi (approved):
Nice, thank you so much and thanks for tackling the test this fast!

PR #43523, opened 2025-11-26
Add MCP prompt support to agent threads

mdliss:
Fixes #43165

## Problem
MCP prompts were only available in text threads, not agent threads. Users with MCP servers that expose prompts couldn't use them in the main agent panel.

## Solution
Added MCP prompt support to agent threads by:
- Creating `ContextServerPromptRegistry` to track MCP prompts from context servers
- Subscribing to context server events to reload prompts when MCP servers start/stop
- Converting MCP prompts to available commands that appear in the slash command menu
- Integrating prompt expansion into the agent message flow

## Testing
Tested with a custom MCP server exposing `explain-code` and `write-tests` prompts. Prompts now appear in the `/` slash command menu in agent threads.

Release Notes:

- Added MCP prompt support to agent threads. Prompts from MCP servers now appear in the slash command menu when typing `/` in agent threads.


agu-z:
@mdliss Thank you for the head start on this! I think we can now merge it.

https://github.com/user-attachments/assets/05466d57-b689-4ede-933b-49f102d1996f

PR #43530, opened 2025-11-26
outline_panel: Fix the panel frequent flickering during search

madcodelife:
The outline panel flickers when searching or when the file content changes. This happens because an empty UI appears during the search process, but it only lasts for a few milliseconds, so we can safely ignore it.

## Before
https://github.com/user-attachments/assets/9b409827-75ee-4a45-864a-58f0ca43191f

## After
https://github.com/user-attachments/assets/b6d48143-1f1a-4811-8754-0a679428eec2

Release Notes:

- N/A

### DIFF
--- a/crates/outline_panel/src/outline_panel.rs
+++ b/crates/outline_panel/src/outline_panel.rs
@@ -111,8 +111,6 @@ pub struct OutlinePanel {
     selected_entry: SelectedEntry,
     active_item: Option<ActiveItem>,
     _subscriptions: Vec<Subscription>,
-    updating_fs_entries: bool,
-    updating_cached_entries: bool,
     new_entries_for_fs_update: HashSet<ExcerptId>,
     fs_entries_update_task: Task<()>,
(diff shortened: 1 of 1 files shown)


crates/outline_panel/src/outline_panel.rs
```diff
@@ -4542,12 +4542,10 @@ impl OutlinePanel {
         cx: &mut Context<Self>,
     ) -> impl IntoElement {
         let contents = if self.cached_entries.is_empty() {
-            let header = if self.updating_fs_entries || self.updating_cached_entries {
```
SomeoneToIgnore, line 4545: Both variables aren't needed then, let's remove those.

SomeoneToIgnore (commented):
Nice catch.

crates/outline_panel/src/outline_panel.rs
madcodelife: removed.

SomeoneToIgnore (approved):

PR #43558, opened 2025-11-26
Add better error visualization and handling for Ollama connections

lukasVirtual:
Hey,

I was trying to connect my Ollama instance with Zed and was initially confused when it wasn't working. Nothing happened. I realized I didn't have Ollama running, but there was no error message. So, I fixed that and implemented this change.

I think this will be very helpful because as Ollama moves toward more cloud-based models, more users will likely encounter API key errors or other connection issues and will need to see what's wrong.

When Zed tries to connect to Ollama and receives a connection reset error, it typically means Ollama isn't running, so we now display a clear message indicating that.

If Ollama returns a different error (for example, if the API key is wrong or the URL is invalid), we now display the specific error message provided by Ollama itself. You can see this in the screenshots.

I'm not sure if this is something you'd want to merge, but I thought it would be a useful addition, so I went ahead and created a pull request for it.

When ollama isnt running or cant be reached:
image

When ollama can be reached but some different error occurs (for example wrong api key or invalid url) since im running olllama local and there is no api key here is the error with a invalid url endpoint:
image

Release Notes:

- Display ollama errors to the user so a user knows quickly if something is wrong and what is wrong

### DIFF
--- a/crates/language_models/src/provider/ollama.rs
+++ b/crates/language_models/src/provider/ollama.rs
@@ -56,6 +56,7 @@ pub struct State {
     http_client: Arc<dyn HttpClient>,
     fetched_models: Vec<ollama::Model>,
     fetch_model_task: Option<Task<Result<()>>>,
+    last_error: Option<SharedString>,
 }
 
 impl State {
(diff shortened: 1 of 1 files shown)


lukasVirtual:
@cla-bot check

lukasVirtual:
@cla-bot check

PR #43559, opened 2025-11-26
terminal: Sanitize URLs with characters that cannot be last

### LINKED ISSUE #43345: Terminal link identification captures too much
### Summary

```zsh
> echo "https://github.com/)"
https://github.com/)
> echo "https://github.com/);"
https://github.com/);
```

The first case is identified correctly, but too much is captured in the second case:

Image

---

One thing to note, this is not a bug in the editor, just the terminal:

Image

Zed: v0.215.0 (Zed Nightly bfe141ea79aa4984028934067ba75c48d99136ae) 
OS: macOS 26.0.1
Memory: 64 GiB
Architecture: aarch64

marcocondrache:
Closes #43345

The list of characters comes from the linkify crate, which is already used for URL detection in the editor:

https://github.com/robinst/linkify/blob/5239e12e26c633f42323e51ed81b0ff534528077/src/url.rs#L228

Release Notes:

- Improved url links detection in terminals.


marcocondrache:
This fix also addresses cases like 'https://test.com/(((' that were not previously covered

crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -158,33 +158,27 @@ fn sanitize_url_punctuation<T: EventListener>(
...
-            remaining_close -= 1;
-        }
-    }
+    // Remove trailing characters that shouldn't be at the end of URLs
+    while let Some(last_char) = sanitized_url.chars().last() {
+        let should_remove = match last_char {
+            // These may be part of a URL but not at the end. It's not that the spec
+            // doesn't allow them, but they are frequently used in plain text as delimiters
+            // where they're not meant to be part of the URL.
+            '?' | '!' | '.' | ',' | ':' | ';' | '*' => true,
```
probably-neb: I think this list is too general, and we don't have reason to strip a few of these characters yet. I think we should constrain this fix to just the problem identified in #43345, and see how it goes, rather than potentially stripping more than we have to, as like you say many of these characters _are_ valid in a URL per the spec.

I think we should keep `;`, `.`, `,`, `:` to strip, however, the following should not be stripped:
 
- `?`: it is a valid empty query, won't effect the routing _unless intended too_, and is far less likely to appear in text than say `,`
- `!`: Is a valid part of a url path, i.e. `https://foo.bar/baz!` might just want to go to the `baz!` page. I also suspect `!` after a URL is far less likely to be an accident, or a delimiter
- `*`: is also a valid url char, and far more likely to be part of the url than a delimiter I think, for example `https://foo.bar/baz?regex=word.*`

probably-neb (changes requested):
Hey! Thanks for taking the time to work on this!

I think we should make this PR a bit more focused to the issue it's closing, rather than hoping that the additional stripped characters works out fine. I'd rather get an issue report saying "you didn't strip enough" than "you stripped too much and changed the meaning of the url", because the first case is more obviously wrong

crates/terminal/src/terminal_hyperlinks.rs
marcocondrache: @probably-neb Thanks for the review. I used those characters because they match the ones used by the crate we already rely on to identify links in the editor implementation, which keeps the detection behavior consistent between the terminal and the editor. The list I used comes from here:

https://github.com/robinst/linkify/blob/5239e12e26c633f42323e51ed81b0ff534528077/src/url.rs#L228

If you think the detection should behave differently in the terminal than in the editor, I can remove those characters.

probably-neb: I do think it should be different from the editor. Maybe I'm wrong here, but I don't see the value in being consistent. My thoughts on which characters to keep vs remove are largely based on my intuitive sense of what is likely to be in the terminal (probably from terminal program output, not just `cat`) which is different than what's likely to be in an editor (e.g. prose)

marcocondrache: @probably-neb that makes sense. Thank you again for reviewing. I've updated the characters

probably-neb (approved):
Great work! Is this ready to merge?

marcocondrache:
@probably-neb thanks, it's ready to be merged!

PR #43567, opened 2025-11-26
agent_ui: Fix show markdown list checked state

RemcoSmitsDev:
Closes #37527

This PR adds support for showing the list state of a list item inside the agent UI. 

**Before**
Screenshot 2025-11-26 at 16 21 31

**After**
Screenshot 2025-11-26 at 16 41 32

Release Notes:
- Agent UI now show the checked state of a list item


RemcoSmitsDev:
Question for the reviewer, should we also make the `•` the same size as the checkbox if you have them inside the list combined?

MrSubidubi:
Thanks for this!

If I recall correctly, we already have an event for this coming from pulldown_cmark which we just don't handle:

https://github.com/zed-industries/zed/blob/e4febf93713cd55843b6ff22545dd9eeda62204e/crates/markdown/src/parser.rs#L398-L399

Which should be logged here as well (I recall my log beind spammed by these before the log was downgraded)

https://github.com/zed-industries/zed/blob/e4febf93713cd55843b6ff22545dd9eeda62204e/crates/markdown/src/markdown.rs#L1220

Could you check whether this would also be fired for these cases?

RemcoSmitsDev:
Hey @MrSubidubi, you are right about that. Will implement it via that way then.

RemcoSmitsDev:
@MrSubidubi Don't think we are able to implement it via the `MarkdownEvent::TaskListMarker` event, as it comes after the `MarkdownTag::Item` code. 

```log
[crates/markdown/src/markdown.rs:952:29] "MarkdownTag::Item" = "MarkdownTag::Item"
[crates/markdown/src/markdown.rs:952:29] builder.checked().is_some() = false
[crates/markdown/src/markdown.rs:1225:21] "MarkdownEvent::TaskListMarker" = "MarkdownEvent::TaskListMarker"
[crates/markdown/src/markdown.rs:952:29] "MarkdownTag::Item" = "MarkdownTag::Item"
[crates/markdown/src/markdown.rs:952:29] builder.checked().is_some() = true
```

MrSubidubi:
Hm, looking at https://github.com/pulldown-cmark/pulldown-cmark/issues/999 (more specifically, the section following the 

> If I manually escape the the list marker the parser produces the events I'd expect...

part, I think this might still be possible, and just an issue with how we handle this in our implementation (which clashes a bit because ours again is push-based..).

Could you check whether things improve by bumping to the latest pulldown_cmark version and also look at whether the produced ranges from pulldown_cmark make sense?

RemcoSmitsDev:
@MrSubidubi Just checked the latest version of `pulldown-cmark`, but still having the same issue that the event is later than the  `MarkdownTag::Item` code path.  

I have an idea how we maybe could refactor this to make it work, but not sure if I like the solution and if it even works. Do you want me to go down that path or you think the current solution is fine?

MrSubidubi:
I think I'd prefer if you went down that path if that is fine for you. I'd definitely favor us handling that event if possible.

RemcoSmitsDev:
@MrSubidubi What do you think about this? Not sure if I like it, but my original idea didn't work.

MrSubidubi (approved):
Think that looks good to me, although I understand your concern with it. Still prefer it over the original approach though, hope you agree there with me.

Will leave it to @danilo-leal  for the final feedback regarding style!

RemcoSmitsDev:
Thanks for your review, yeah I agree it's better, but it feels still a bit off.

danilo-leal (approved):
Looks great! I added a new method to the Checkbox so we can render it as if it was enabled but it actually doesn't use a pointer cursor and doesn't render hover styles. Using `disabled` made it look too dimmed out.

Issue #6754, opened 2024-01-26
Add `--offline` mode that disables all network features at runtime


alerque:
One thing I have found in the past with editors focusing on collaboration and connectivity is that they are very frustrating to use when network connectivity is bad or non-existent. They throw error after error about not being able to check for plugin updates and not being able to sync files and not being able to open chat and whatever else. The more connected the editor is the more frustrating it tends to be to use in local only mode when the circumstances call for it.

In order to be truly robust, I think all the run time features that rely on/expect/check for network connectivity should respond to some sort of run time flag (CLI flag or ENV var) that sets an offline status. This would not only suppress errors and warnings but not even *try* to make connections for any reason.

This is also tangentially related to #6756 and the existing `assistant = false` option such that AI tooling that relies on remote computing should be disabled while anything that relies on local-only models and tooling could stay enabled.

nervenes:
An offline mode that's toggle able via the settings would also be nice, not only the benefits you mentioned but also the added security would be a huge benefit as well - both for casual users and company policies that currently limit the use of zed.

DanielAdari:
Offline mode is a must for organizations that are working on air-gapped networks.

timholl:
As a work-around, I'm using ZED in a separate, offline network namespace like this

```shell
firejail --net=none zeditor
```
For a more persistent setup, one can create a profile: `echo "net none" > ~/.config/firejail/zeditor.profile && sudo firecfg`.

I install Extensions via Nix home-manager in a declarative and pinned way like this:

```nix
xdg.dataFile."zed/extensions/installed/php".source = builtins.fetchTarball {
  url = "https://api.zed.dev/extensions/php/0.2.8/download";
  sha256 = "sha256:1vrrxabv6451rzxcf31yl3x31gyqlr6xjfwmhirr2jjq8r0wv7wz";
};
...
```

Also, I'm installing the LSPs natively (because [I don't want ZED to download & execute random binaries without my consent](https://github.com/zed-industries/zed/issues/12589)), and this works since ZED searches for them locally via `which` first.

ankibalyan:
For Me, It does not allow to run golang debugger without the internet connection, and with my org network settings urls are blocked recently. It is sad that that I had to switch to vscode for that.

In console I see, when try to run debugger.

```
error: status error 403

```

[...]

PR #43573, opened 2025-11-26
Add offline mode

ankitrijal2054:
Add offline mode support

  Summary
  This PR introduces a comprehensive offline mode feature that allows Zed to operate gracefully when network connectivity is unavailable. When offline mode is enabled, network-dependent features are disabled with clear UI feedback, while local operations continue to work normally.

  Implementation

  New offline_mode crate
  - Created a new crate (crates/offline_mode) that provides a centralized OfflineModeSetting
  - Users can enable offline mode via settings: "offline": true
  - The setting defaults to false to maintain backward compatibility

  Guarded network-dependent features
  AI/LLM Features (crates/language_models, crates/agent, crates/agent_ui, crates/copilot)
  - All language model providers (Anthropic, OpenAI, Google, Mistral, DeepSeek, X.AI, etc.) check offline state before API calls
  - Agent panel shows clear "Offline" state when disabled
  - Inline assistant and language model selector display appropriate offline indicators
  - Copilot chat operations are blocked with user-friendly messages

  Telemetry (crates/client, crates/zeta)
  - Telemetry events are not transmitted when offline
  - Background telemetry tasks respect offline state

  Collaboration Features (crates/collab_ui, crates/call, crates/livekit_client)
  - Collaboration panel shows disabled state
  - Room operations (joining, creating) are prevented
  - LiveKit connections are not initiated

  Auto-updates (crates/auto_update, crates/auto_update_ui)
  - Update checks are skipped when offline
  - UI shows "Updates unavailable in offline mode" tooltip

  Remote Git Operations (crates/git_ui)
  - Remote operations (push, pull, fetch, clone) are disabled with explanatory tooltips
  - Buttons show disabled state when offline
  - Local Git operations (status, diff, commit, branch, log) continue to work normally

  Web Features (crates/web_search_providers, crates/assistant_slash_commands)
  - Web search providers check offline state
  - /fetch slash command blocks network requests when offline

  Testing
  - Unit tests added for offline mode setting
  - Integration tests verify offline guards across subsystems
  - Manual testing confirmed UI states and error handling

  Usage:
  Users can enable offline mode by adding to their settings:
  {
    "offline": true
  }

  When enabled:
  - All network-dependent features show disabled states with tooltips explaining why
  - Local operations (file editing, local Git operations, etc.) continue working normally
  - No network errors are shown to the user

  Screenshots
Screenshot 2025-11-26 at 11 32 05 AM
Screenshot 2025-11-26 at 11 32 20 AM
Screenshot 2025-11-26 at 11 33 17 AM
Screenshot 2025-11-26 at 11 33 40 AM
Screenshot 2025-11-26 at 11 33 53 AM
Screenshot 2025-11-26 at 11 41 57 AM

  ---
  Notes for Reviewers

  - This is a non-breaking change - offline mode is opt-in via settings
  - All network operations have been identified and guarded appropriately
  - The implementation uses a consistent pattern across all subsystems for checking offline state
  - Local operations are intentionally not affected to ensure Zed remains functional offline


lhutyra:
It is related to this issue probably
https://github.com/zed-industries/zed/issues/6754

teknalb:
Looking forward for this, but I must ask, would this prevent access to git repos on a private local network? I'd definitely like to have my git instance to remain accessible in a no internet environment.

zhangyuge6666:
Is this implementation method too extreme? The users of Zed should merely aim to use Zed more effectively in an offline environment. This could be an offline environment, or it could be an environment with limited network access. For instance, as mentioned above, why can't the git network be used? If the git remote repository used by the developers is within the local network, shouldn't it be possible to use it? What puzzles me is that Zed frequently checks for updates of each extension and language server. When the network environment is poor, various problems such as failure in starting the language server occur. Isn't there any way to prevent Zed from doing these tasks on its own? Rather than applying a one-size-fits-all approach, all networks - both active and passive - should not be made inaccessible. If this extreme method is used to achieve it, when I need to use the local network's git push function in offline mode and switch back to online mode, will Zed automatically update the extensions, language servers, etc.?

zhangyuge6666:
First of all, I would like to thank the zed team for their contribution. I believe that just having an online/offline mode toggle button alone cannot enhance the developer's experience. I wonder if it would be better to allow developers to choose whether zed should actively connect to the internet for all current active internet-related operations. For example, the extended automatic update feature could be changed to manual update, and the extended offline installation function could also be implemented. I don't expect it to be better than what VSCode does; an experience similar to VSCode would be sufficient. For functions that require internet access, if the network is poor or there is no network, a network connection error message should be given. However, the current zed has a problem where it malfunctions when the network is poor, such as: the eslint that was working normally yesterday would fail when opening zed. If the network is problematic, eslint would not be able to function. This is unreasonable. The language server also has this issue. I think whether it's automatic update or manual update, if it fails, a prompt should be given stating that the update failed due to a network issue. I believe there is no need for a button to switch off all networks. This is just my personal suggestion and does not represent all developers. I hope the zed team carefully researches and considers this.

MrSubidubi:
Given 
image

after a week and the general approach, I am gonna close this for the time being.

Better behavior when offline is something we definitely want, but disabling it like this is not the approach we'd want to take here - what @zhangyuge6666 laid out here sounds way more like the behavior we are looking for. https://github.com/zed-industries/zed/pull/43573#issuecomment-3590919122 describes it to its fullest and I couldn't agree more with everything laid out there. We should handle this gracefully and do everything to the best of our ability when we detect poor to no network connection instead of putting the burden on users to toggle a setting for this.

Given that this approach varies a lot from this path and is extremely prone to errors due to this effectively being a guard at all places, I am gonna close this for the time being. Thank you for working on this nonetheless.

PR #43576, opened 2025-11-26
gpui: Add `Input` element

iamnbutler:
Edit: Need to rewrite but for now...

- Adds `Input`, a single & multiline input solution
- Supports any expected keybindings, copy, paste, undo, redo, etc...
- Supports LTR and RTL, as well as mixed direction content (RTL still sorta WIP)
- Supports unbinding and rebinding keys via `bind_input_keys` api

Discussing in [this thread](https://discord.com/channels/869392257814519848/1443320164895690874) over in the gpui discord!

---

This PR introduces `Input` and `TextArea`:

- `Input` (Entity) — owns state: content, selection, cursor, scroll offset, line layouts
- `TextArea` (Element) — renders Input, handles paint/layout, registers actions, etc

`Input` can be shared across future text elements (like a text field/single line input.) I'm still somewhat split on if a single line text input element should just be a mode of `TextArea` or a separate element entirely.

- `Input` implements `EntityInputHandler` and supports IME input. Examples below.
- `TextArea` supports all the things you might expect: selection via mouse and keyboard, cursor movement, scrolling, copy/paste, etc. There is currently no history support (undo/redo)

Known Issues:
- [ ] The cursor position during IME input can get a bit wonky.
- [ ] Copying complex emoji can be wonky
- [ ] Performance on pathologically large amounts of text can be slow (usable at 2mb of text, pretty damn slow at 20mb+) (won't fix)
- [ ] No focus styles
- [ ] No cursor blink (with y'all's permission I'd like to hoist the `BlinkManager` up out of `crates/editor` in to `gpui` so all input-like elements could share it.)

---

Example:

```rs
struct Chat {
    message_input: Entity<Input>,
}

impl Render for Chat {
    fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
        let focus_handle = self.input.focus_handle(cx);
        div()
            .id("chat")
            .key_context("Chat")
            .track_focus(&focus_handle)
            .child(
                text_area(&self.input)
                    .w_64()
                    .h_32()
                    .p_2()
                    .bg(rgb(0x2d2d2d))
                    .text_color(rgb(0xffffff))
                    .text_sm()
                    .selection_color(rgba(0x264f7844))
                    .cursor_color(rgb(0x3378F6)),
            )
    }
}

impl Focusable for Chat {
    fn focus_handle(&self, cx: &App) -> FocusHandle {
        self.input.focus_handle(cx)
    }
}

// then you would need to bind keys...
// see the text_area example or run `cargo run -p gpui --example text_area`
```

There are still some rough edges, but I've done quite a bit of testing.

---

### Scrolling, selections, wrapping, word-based movement

https://github.com/user-attachments/assets/507f44ff-9f5e-4d5d-9b9c-18ae68e7d3fc

### Text input, copy + paste, keyboard movement

https://github.com/user-attachments/assets/54481b01-a29b-4a15-865b-cf1631c4f848

### IME input + selections, mixed inputs

https://github.com/user-attachments/assets/7223cdb3-1d00-4cef-ad7a-6b16e7a4ed66

### Tricky cases - emoji, merged emoji, multibyte characters, mixed content, etc

https://github.com/user-attachments/assets/eddc0b9d-09b2-4c83-a9fb-1c9cf7547bf9

(looks like there is a copy/paste issue with emoji exposed by this video ^)

---

Release Notes:

- N/A


crates/zed/src/zed.rs
```diff
@@ -4757,6 +4757,7 @@ mod tests {
                 "git_panel",
                 "go_to_line",
                 "icon_theme_selector",
+                "input",
```
iamnbutler, line 4760: I'm not sure why there is a static list of namespaces, and if gpui/non-zed ones should be in this list... but I added it here for now to make the test pass

Cargo.toml
```diff
@@ -694,6 +694,7 @@ tree-sitter-rust = "0.24"
 tree-sitter-typescript = { git = "https://github.com/zed-industries/tree-sitter-typescript", rev = "e2c53597d6a5d9cf7bbe8dccde576fe1e46c5899" } # https://github.com/tree-sitter/tree-sitter-typescript/pull/347
 tree-sitter-yaml = { git = "https://github.com/zed-industries/tree-sitter-yaml", rev = "baff0b51c64ef6a1fb1f8390f3ad6015b83ec13a" }
 unicase = "2.6"
+unicode-bidi = "0.3"
```
iamnbutler, line 697: we add this to get detection of RTL lines on macos as coretext doesn't pass this down like cosmic_text does.

This is already a peerdep, so it doesn't actually expand the crategraph further

mikayla-maki:
Hi,

PR #43581, opened 2025-11-26
Add Doxygen injection into C and C++ comments

Clement-Lap:
Release Notes:

- C/C++ files now support Doxygen grammars.

### DIFF
--- a/crates/languages/src/c/injections.scm
+++ b/crates/languages/src/c/injections.scm
@@ -1,6 +1,7 @@
 ((comment) @injection.content
- (#set! injection.language "comment")
-)
+  (#match? @injection.content "^(///|//!|/\\*\\*|/\\*!)(.*)")
+  (#set! injection.language "doxygen")
+  (#set! injection.include-children))
 
 (preproc_def
     value: (preproc_arg) @injection.content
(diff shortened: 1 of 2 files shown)


Clement-Lap:
Release Notes:

- Updated `crates/languages/src/c/injections.scm` to inject Doxygen language into C comments
- Updated `crates/languages/src/cpp/injections.scm` to inject Doxygen language into C++ comments

osiewicz:
Ugh, this was an oversight on my end, but this should not have been merged in the current state.

Clement-Lap:
Sorry, I developed an extension to support doxygen in C and C++ files, and I have been told to directly add this feature into zed. I have been very busy since, and did not pursue the development of my extension. I however plan to resume it soon.

osiewicz:
This isn't your fault - it's on me. Feel free to resubmit a PR when you feel that your extension is in a good place.
Either way, we need to have two injections; one for comments in general (in this case the language will be set to `comment`) and another one for when a comment is a Doxygen comment.
So tl;dr: we shouldn't change existing injection, but add a new one instead.

niekdomi:
fwiw, clangd 22 supports doxygen https://clangd.llvm.org/config.html#documentation

PR #43586, opened 2025-11-26
Basic side-by-side diff implementation

cole-miller:
Release Notes:

- N/A


crates/multi_buffer/src/multi_buffer.rs
```diff
@@ -87,6 +92,14 @@ pub struct MultiBuffer {
     /// The writing capability of the multi-buffer.
     capability: Capability,
     buffer_changed_since_sync: Rc<Cell<bool>>,
+    follower: Option<Entity<MultiBuffer>>,
+    filter_mode: Option<MultiBufferFilterMode>,
```
Veykril, line 96: Can we add docs to these fields in what they do? Also maybe on `MultiBufferFilterMode` and its variants

crates/multi_buffer/src/multi_buffer.rs
```diff
@@ -2946,12 +3098,56 @@ impl MultiBuffer {
...
-            let edit_old_end = old_diff_transforms.start().1 + edit_old_end_overshoot;
+            let edit_old_end_overshoot = if let Some(DiffTransform::FilteredInsertedHunk {
+                ..
+            }) = old_diff_transforms.item()
+            {
+                ExcerptDimension(MultiBufferOffset(0))
+            } else {
+                ExcerptDimension(MultiBufferOffset(
+                    edit.old.end - old_diff_transforms.start().0,
+                ))
```
Veykril, line 3145: Does this need to be wrapped in an `ExcerptDimension`? Looks unnecessary to me

0x2CA:
This commit breaks the functionality of closing project diffs, and there is no way to try reopening it.

https://github.com/user-attachments/assets/ac36ffb5-a974-4aa0-b114-5b138cf49ede

Not sure if this is an expected issue and we are waiting for a later fix, or if it’s an unexpected error.

cole-miller:
@0x2CA thanks, should be fixed in #43964

matthyx:
hi, how do we enable this?

0x2CA:
> hi, how do we enable this?

This can only be previewed internally by the team, and the level of completion is very low; it's just a simple split view.You could say this is just a half-finished product.

Normally, this is use user's flag to enable the feature, but this flag is controlled in the cloud. Typically, regular users don't have it enabled. You can modify the code to force it enable, and then bind a keymap to switch modes.

waltcow:
guys, any update? 🥲

PR #34985, opened 2025-07-23
Make key repeat rate on Wayland more precise

olejorgenb:
In Wayland, the client implement key repeat themself. In Zed this is ultimately handled by the gpui crate by inserting a timer source into the event loop which repeat itself if the key is still held down [1].

But it seems the processing of the repeated key event happen synchronously inside the timer source handler, meaning the effective rate become slightly lower (since the repeated timer is scheduled using the 1/rate as delay).

I measured the event processing time on my laptop and it's typically around 3ms, but sometimes spiking at 10ms. At low key repeat rates this is probably not _very_ noticeable. I see the default in Zed is set to a (measly) 16/s, but I assume most systems will use something closer to 25, which is a 40ms delay. So ~3ms is around 7.5% of the delay. At higher rate the discrepancy become worse of course.

I can visible notice the spikes, and doing some crude stopwatch measurements using gedit as a reference I can reproduce around 5-10% slower rates in Zed.

IMO this is significant enough to warrant improving, especially since some people can get quite used the repeat rate and might feel something being "off" in Zed.

The suggested fix simply subtract the processing time from the next delay timer.

[1] https://github.com/olejorgenb/zed/blob/32df726f3b7fa83e7399f6629c59e0a3f3fff125/crates/gpui/src/platform/linux/wayland/client.rs#L1355

Release Notes:

- Improved Wayland (Linux) key repeat rate precision

Issue #43330, opened 2025-11-22
Key repeat rate is 3x slower than in system settings (Linux, Wayland)


michalrus:
### Summary

I set my system key repeat rate very high, and on Linux, Wayland, Zed’s key repeat is 3x slower than these settings.

### Description

I’m running Zed 0.212.6 on the default config, on Hyprland 0.49.0 (Wayland, Linux), and set these `libinput` settings:

```
repeat_delay = 150
repeat_rate = 125
```

The unit of `repeat_delay` is milliseconds, and of `repeat_rate` – repeats per second.

Steps to reproduce:
1. Set the `repeat_delay` and `repeat_rate` in your window manager input config.
2. Open freshly installed Zed.
3. For exactly 10 seconds, press a key, e.g. `x`.
4. Copy the resulting `xxxx…` string into `wc -c` and note the number of characters.

**Expected Behavior**:

Zed’s rate is 125 repeats/second, so you get around 1,250 `xxxx…` characters.

**Actual Behavior**:

Zed’s rate is 41.3 repeats/second, so you get around 413 `xxxx…` characters.

For a person used to 125 repeats/second, it makes—otherwise _very_ performant—Zed _feel_ a little sluggish.

### Comparing with other apps

I used the same steps to reproduce (above) with other apps for comparison. 

Their results consistently match the `repeat_rate` that I set system-wide.

| Text input env | Repeats per 10 s | Repeats per 1 s |
|--------|--------|--------|
| Zed (a `*.txt` file) | 413 | 41.3 |
| Zed (a terminal pane) | 472 | 47.2 |
| NodepadNext | 1,137 | 113.7 |
| `cat` running in Alacritty | 1,230 | 123.0 |
| Emacs (a `*.txt` file) | 1,224 | 122.4 |
| Firefox URL bar | 1,246 | 124.6 |
| Firefox `<textarea>` | 1,224 | 122.4 | 

You can see that all other apps generally match the system-wide `repeat_rate = 125`, while Zed is at around 45.

### Hypothesis

Could the 40–50 key repeats per second be the same as Zed’s frame rate? 😮 

### Zed Version and System Specs

Zed: v0.212.6 (Zed) 
OS: Linux Wayland nixos 25.05
Memory: 31 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 630 (CFL GT2) || Intel open-source Mesa driver || Mesa 25.0.7

### If applicable, attach your `Zed.log` file to this issue.

[...]

PR #43589, opened 2025-11-26
Make key repeat rate on Wayland more precise (2)

### LINKED ISSUE #39042: Cursor movement slow on Linux
### Summary

# Zed moves the cursor 0.7 times slower than VS Code.

### Description

On Linux + Wayland with a 4K display, Zed’s cursor movement feels noticeably slow. Technically it’s about 0.7× the speed of VS Code, but subjectively the difference is much bigger.
In the test video (same file, holding down the down arrow from the top to bottom), Zed takes 28s to scroll through while VS Code finishes in 21s.
On a 1080p screen the gap is smaller, but still there.

VS Code Video:
https://imgur.com/j9l80DW

Zed Video:
https://imgur.com/qOpXuLq

Zed: v0.205.6 (Zed) 
Memory: 15.4 GiB
Architecture: x86_64
GPU: Intel(R) Graphics (ADL-N) || Intel open-source Mesa driver || Mesa 25.2.1-arch1.4

olejorgenb:
CLOSES  #39042

This is a reopening of #34985+. 

_Original descrioption_:

In Wayland, the client implement key repeat themself. In Zed this is ultimately handled by the gpui crate by inserting a timer source into the event loop which repeat itself if the key is still held down [1].

But it seems the processing of the repeated key event happen synchronously inside the timer source handler, meaning the effective rate become slightly lower (since the repeated timer is scheduled using the 1/rate as delay).

I measured the event processing time on my laptop and it's typically around 3ms, but sometimes spiking at 10ms. At low key repeat rates this is probably not _very_ noticeable. I see the default in Zed is set to a (measly) 16/s, but I assume most systems will use something closer to 25, which is a 40ms delay. So ~3ms is around 7.5% of the delay. At higher rate the discrepancy become worse of course.

I can visible notice the spikes, and doing some crude stopwatch measurements using gedit as a reference I can reproduce around 5-10% slower rates in Zed.

IMO this is significant enough to warrant improving, especially since some people can get quite used the repeat rate and might feel something being "off" in Zed.

~~The suggested fix simply subtract the processing time from the next delay timer.~~

[1] https://github.com/olejorgenb/zed/blob/32df726f3b7fa83e7399f6629c59e0a3f3fff125/crates/gpui/src/platform/linux/wayland/client.rs#L1355

Release Notes:

- Improved Wayland (Linux) key repeat rate precision

### DIFF
--- a/crates/gpui/src/platform/linux/wayland/client.rs
+++ b/crates/gpui/src/platform/linux/wayland/client.rs
@@ -1399,6 +1399,7 @@ impl Dispatch<wl_keyboard::WlKeyboard, ()> for WaylandClientStatePtr {
                                     prefer_character_input: false,
                                 });
                                 move |_event, _metadata, this| {
+                                    let processing_begin = std::time::Instant::now();
                                     let mut client = this.get_client();
                                     let mut state = client.borrow_mut();
                                     let is_repeating = id == state.repeat.current_id
@@ -1414,8 +1415,10 @@ impl Dispatch<wl_keyboard::WlKeyboard, ()> for WaylandClientStatePtr {
 
                                     drop(state);
                                     focused_window.handle_input(input.clone());
-
-                                    TimeoutAction::ToDuration(Duration::from_secs(1) / rate)
+                                    let processing_duration = processing_begin.elapsed();
+                                    log::debug!("Key repeat input handling: {:?}", processing_duration);
+                                    let delay = Duration::from_secs(1) / rate;
+                                    TimeoutAction::ToDuration((delay - processing_duration).max(Duration::ZERO))
                                 }
                             })
                             .unwrap();
--- a/crates/gpui/src/platform/linux/wayland/client.rs
+++ b/crates/gpui/src/platform/linux/wayland/client.rs
@@ -1389,6 +1389,7 @@ impl Dispatch<wl_keyboard::WlKeyboard, ()> for WaylandClientStatePtr {
                         state.repeat.current_keycode = Some(keycode);
 
                         let rate = state.repeat.characters_per_second;
+                        let repeat_interval = Duration::from_secs(1) / rate;
                         let id = state.repeat.current_id;
                         state
                             .loop_handle
@@ -1398,8 +1399,7 @@ impl Dispatch<wl_keyboard::WlKeyboard, ()> for WaylandClientStatePtr {
                                     is_held: true,
                                     prefer_character_input: false,
                                 });
-                                move |_event, _metadata, this| {
-                                    let processing_begin = std::time::Instant::now();
+                                move |event_timestamp, _metadata, this| {
                                     let mut client = this.get_client();
                                     let mut state = client.borrow_mut();
                                     let is_repeating = id == state.repeat.current_id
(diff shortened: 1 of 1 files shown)


olejorgenb:
Regarding the revised approach:

I think the changes should be fairly easy to verify as correct: Instead of scheduling the new key-repeat event by timedelta, schedule it at the exact time. This make the rate fully independent of the frame-processing (unless it's unable to keep up).

Two details I had to verify:

1. What happens if an event is scheduled to run in the past? This can happen if the frame takes too long to process.
2. Can a Timer event source created from a time-delta reschedule using an absolute time

For (1) the [documentation](https://smithay.github.io/calloop/api/calloop/timer/struct.Timer.html) state:

> When registered to the event loop, it will trigger an event once its deadline is reached. _If the deadline is in the past relative to the moment of its insertion in the event loop, the TImer will trigger an event as soon as the event loop is dispatched._

Unfortunately the documentation doesn't explicitly mention what happens when the source is re-registered by returning an value from the callback, but the [source-code](https://github.com/Smithay/calloop/blob/72ab703e65681001aa69df798cf9fcd729b78a74/src/sources/timer.rs#L135) is clear:  When returning a `TimeoutAction` the source is simply re-inserted (compare to the `register` method below.

For (2) the [documentation](https://smithay.github.io/calloop/api/calloop/timer/enum.TimeoutAction.html) is clear to me.

---

> The provided event is an [Instant](https://doc.rust-lang.org/beta/std/time/struct.Instant.html) representing the deadline for which this timer has fired (which can be earlier than the current time depending on the event loop congestion). [source](https://smithay.github.io/calloop/api/calloop/timer/index.html)

So I think that scheduling the repeats absolutely from the event deadline also eliminate any delays due to event loop congestion (to the extent possible)

---

I've also verified that it works. My key-repeat interval is 30ms and moving 1590 lines take 53s without the patch and 48s with the patch. Theoretical speed: 47.7s (disregarding initial delay). Not a huge difference, but significant enough to rule out measurement uncertainty. For fun I tried with 5ms (theoretical: 8s). Unpatched: 13.8. Patched: 8.5.

olejorgenb:
Re-mentioning @bbb651's [comment](https://github.com/zed-industries/zed/pull/34985#issuecomment-3117432479) in the closed PR: 

> It's worth noting [key_state::repeated](https://wayland.app/protocols/wayland#wl_keyboard:enum:key_state:entry:repeated) was added in wl_seat version 10 which let's the compositor handle key repeat ...

I assume it will be quite a while until (if ever) every compositor support this, so I think it's still worth implementing this fix.

yara-blue:
This makes total sense! Thanks for looking into this and sharing not only your fix but your findings!

michalrus:
This also resolves:
* #43330 

Awesome, thank you!

crates/gpui/src/platform/linux/wayland/client.rs
```diff
@@ -1419,6 +1389,7 @@ impl Dispatch<wl_keyboard::WlKeyboard, ()> for WaylandClientStatePtr {
                         state.repeat.current_keycode = Some(keycode);
 
                         let rate = state.repeat.characters_per_second;
+                        let repeat_interval = Duration::from_secs(1) / rate;
```
rubbieKelvin, line 1392:
```suggestion
                        let repeat_interval = Duration::from_secs(1) / rate.max(1);
```

to prevent division by zero crash?

olejorgenb: Ah, did not consider 0 was a possible rate, but I guess it's possible to configure that if you want to turn of key-repeat altogether. Fixed here: https://github.com/zed-industries/zed/pull/44151

PR #43609, opened 2025-11-26
macos: Add missing file access entitlements

aeroxy:
Adds `com.apple.security.files.user-selected.read-write` and `com.apple.security.files.downloads.read-write` to zed.entitlements.

This resolves an issue where the integrated terminal could not access external drives or user-selected files on macOS, even when "Full Disk Access" was granted. These entitlements are required for the application to properly inherit file access permissions.

Release Notes:

- Resolves an issue where the integrated terminal could not access external drives or user-selected files on macOS.

### DIFF
--- a/crates/zed/resources/zed.entitlements
+++ b/crates/zed/resources/zed.entitlements
@@ -22,5 +22,9 @@
 	<true/>
 	<key>com.apple.security.personal-information.photos-library</key>
 	<true/>
+	<key>com.apple.security.files.user-selected.read-write</key>
+	<true/>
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (approved):
I'm not an expert but looks good from my perspective.
Let's merge and run in in Nightly to see how things go.

Thank you.

aeroxy:
> I'm not an expert but looks good from my perspective. Let's merge and run in in Nightly to see how things go.
> 
> Thank you.

awesome

PR #43618, opened 2025-11-27
Add Vim-like Which-key Popup menu

xipeng-jin:
Closes #10910

Follow up work continuing from the last PR https://github.com/zed-industries/zed/pull/42659. Add the UI element for displaying vim like which-key menu.

https://github.com/user-attachments/assets/3dc5f0c9-5a2f-459e-a3db-859169aeba26

Release Notes:

- Added a which-key like modal with a compact, single-column panel anchored to the bottom-right. You can enable with `{"which_key": {"enabled": true}}` in your settings.


zayne-anyra:
only if vim is enable right?

xipeng-jin:
> only if vim is enable right?

Yes, `vim_mode` or `helix_mode` needs to turn on.

claytonrcarter:
> Yes, vim_mode or helix_mode needs to turn on.

This is cool! Would it be hard to support non-vim/helix mode? The are plenty of “compound” bindings in regular mode, as well as in the emacs key map, that could be exposed this way. Eg `Cmd-k Cmd-s`.

crates/which_key/src/which_key.rs
```diff
@@ -0,0 +1,98 @@
...
+                    .log_err();
+            }));
+        })
+        .detach();
+    })
+    .detach();
+}
+
+// Hard-coded list of keystrokes to filter out from which-key display
+pub static FILTERED_KEYSTROKES: LazyLock<Vec<Vec<Keystroke>>> = LazyLock::new(|| {
```
mgabor3141, line 61: For a followup, maybe a good heuristic would be to:

- Enumerate all possible actions
- Choose the binding with the fewest keys for each action and show only that.

In case of a tie, I think it's okay to show more than one keybind.

ConradIrwin:
Amazing, thank you for this!

A few things I noticed:
* We can avoid measuring every row by rendering two columns adjacent to each other and letting taffy lay things out. (I did this, and also left-aligned the key bindings, it looked strange to me when they were right aligned).
* Sometimes the first time I trigger the menu by hitting `g` it is empty: 
Screenshot 2025-12-01 at 22 02 19
* `ctrl-w` is incorrectly still triggering the default behavior  of deleting backwards despite https://github.com/zed-industries/zed/blob/a42a135750ff9130a11c8a915351bd03cc707c39/assets/keymaps/vim.json#L843 (we don't have to fix this in this PR, but would be good to figure out which change broke that and fix it in a separate PR).
* That made it obvious that having a 700ms timeout for which-key, but a 1000ms timeout for "run a conflicting action" led to the menu flashing. I've fixed that by setting the default for which-key to 1000ms, so it won't show up in this case at all.

Are you able to look into why it renders empty sometimes?

xipeng-jin:
Hi @ConradIrwin , thank you so much for your prompt review. Here is my answers for your comments:
1. Yes, I agree. I think this is a pretty smart idea to render two columns for avoiding per row measurement overhead. And left-align the key bindings would be great for consistency.

> We can avoid measuring every row by rendering two columns adjacent to each other and letting taffy lay things out. (I did this, and also left-aligned the key bindings, it looked strange to me when they were right aligned).

2. I haven't encountered that issue (render empty menu) yet and look into it. Since I have been working on a follow-up PR https://github.com/zed-industries/zed/pull/43949 on top of this and just opened it up, I am going to fully test it now to making sure we cover all the edge cases. I can remember that there are a few spots in this PR worth to have a second look. One is here https://github.com/zed-industries/zed/pull/43618/files#diff-f00679bfd9619761abc62cbb720f02d733d56cc361fb84d127600e7d464e853fR61 (I don't really like to have hard coded value FILTERED_KEYSTROKES in the code base, but this is inspired from the previous PR and it is the easy way to filter the duplicates.), the other one is https://github.com/zed-industries/zed/pull/43618/files#diff-42e0d1c722de5403e1bf675699637e42ccabaec4cae3a5eabdbf38b503ce8b37R180-R186 (This is probably the spot of the issue coming from).

> Sometimes the first time I trigger the menu by hitting g it is empty:

3. There is another minor issue I noticed is about the vertical scrollbar. I don't want to trigger any scrollbar to appear and fade out (it is annoying me) for the case when the height of popup menu already fully covered all the keybindings and no need for scrolling. We should disable the scroll bar in this scenario. 

BTW, if you have any extra time, can you have a quick look for my follow up PR https://github.com/zed-industries/zed/pull/43949? From the high level, I am not sure that is the direction for Zed team want on the current roadmap, but having some customization/renaming capabilities would be great for next level user experiences (if not that should be great for a future PR). But like I mentioned in the parent issue https://github.com/zed-industries/zed/issues/10910#issuecomment-3599160192, that PR would be just a "add the finishing touch". Therefore, feel free to close either one of them so that we can just target one direction.

> The reason why I separate the two PRs (https://github.com/zed-industries/zed/pull/43949 and https://github.com/zed-industries/zed/pull/43618) because the previous one https://github.com/zed-industries/zed/pull/43618 already accomplish the implementation of the basic functionalities from Vim which-key like menu. That one should be the essential part, and this one is just adding the finishing touch. Therefore, it will make the team review it easier and make them easier to be prioritized.

Thanks again for taking your time here.

avbm:
I was testing this out on my local machine and noticed that pressing `<space> `didn't trigger the which-key popup - however pressing g/other alphabet keys did - just FYI. 

Thanks for all the work you are putting into this and really looking forward to getting this feature.

xipeng-jin:
Hi @avbm , thanks for reporting this. However, this is not an issue with the which-key system, instead the root cause is that we have the default key mapping in zed for `space` to `vim::WrappingRight` action which will first be triggered and not showing up the menu. You need to put something like the following in your keymap.json:
```json
  {
    "context": "(VimControl && !menu)",
    "bindings": {
      "space": null, // Disable the default action vim::WrappingRight
    },
  }
```

avbm:
Yep, that did the trick - thanks.

jrmajor:
@axelson https://zed.dev/faq#what-are-the-release-channels

yannxaver:
Great contribution, thank you!

Any plans to make the position of the popup menu configurable? 
Personally I would prefer it in the center of the editor.

ggobbe:
@Wozacosta I had to remove the context for it to work for me (I also tried replacing it with `helixControl` w/o success)

**settings.json**
```
  "which_key": {
    "enabled": true,
    "delay_ms": 100,
  },
```

**keymap.json**
```
[
  {
    "bindings": {
      "space": null, // Disable the default action vim::WrappingRight
    },
  },
]
```

luccahuguet:
```json
  {
    "context": "(VimControl && !menu)",
    "bindings": {
      "space": null, // Disable the default action vim::WrappingRight
    },
  },
```

also works

tn-ascera:
Hello, I tried following the comments about setting the space key to null but it does not work for me.

image
image

olejorgenb:
>   only if vim is enable right?
>> Yes, vim_mode or helix_mode needs to turn on.

Just noting here in case anyone else wondering. `which-key` **works** when vim/helix is disabled (the regular, default keybinding mode). 

But with one caveat: The menu only show if the keybinding prefix is only used as the prefix. If it's bound to something else as well, the menu wont show. 

You might think this is uncommon, but one of the default chained keybindings in Zed does use such a prefix. `Alt-g b` etc. Since `Alt-g` by itself insert some unicode char(?), `Alt-g` is not only used for the prefix. And thus the menu does not show. This was the one chained keybind I knew about, so I tested this and became disappointed.. Luckily I  dug a bit further.

PR #43626, opened 2025-11-27
editor: Show "Toggle Excerpt Unfold" tooltip when buffer is fold

cppcoffee:
Release Notes:

- N/A


cppcoffee:
Screenshots of the effect are as follows:

Before fold:

截屏2025-11-27 11 34 56


Folded:

截屏2025-11-27 11 34 48

danilo-leal (approved):
Thank you! Pushed a small commit simplifying it a bit more.

PR #43633, opened 2025-11-27
ui: Fix custom size preview example of vector component

madcodelife:
In the vector component custom size preview example, the image will overflow the content box.

| Before | After |
| - | - |
| Before | After |

Release Notes:

- N/A


danilo-leal (approved):
Thank you!

PR #43636, opened 2025-11-27
component_preview: Pin the filter input at the top

madcodelife:
When I scroll the page after selecting a component, the filter input disappears. It would be more helpful if it stayed fixed at the top.

## Before

https://github.com/user-attachments/assets/e26031f1-d6e7-4ae2-a148-88f4c548a5cf

## After

https://github.com/user-attachments/assets/f120355c-74d3-4724-9ffc-71fb7e3a4586

Release Notes:

- N/A


danilo-leal (approved):
Thanks, great change!

PR #43639, opened 2025-11-27
gpui: Detect VRR displays to reduce GPU usage

### LINKED ISSUE #29073: Moving cursor in an active window significantly increases GPU usage
### Summary

When you move a cursor in an active Zed window, the GPU consumes more power.

### Description
I noticed that when I use Zed on my laptop, the battery discharges faster. I suspected it was the editor that drained my battery and it was true. I checked and the problem is not with the settings I set — the issue still remains with defaut Zed settings. And it doesn't matter what speed I move the cursor with — the increase is the same.

Steps to reproduce:
1. Open any GPU monitoring app/tool/program
2. Open any project/file in Zed (or just open Zed with no project/file)
3. Make the window active by clicking in it
4. Move the cursor inside the window and look at the GPU monitor

Expected Behavior: (almost) no increase in GPU usage
Actual Behavior: GPU uses 2-3 times more power than with the cursor standing still.

https://github.com/user-attachments/assets/e94d4db6-ee7a-4b9b-bf09-b93ff5a6f685

Zed: v0.182.11 (Zed) 
OS: Linux X11 linuxmint 22.1
Memory: 15 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 24.2.8-1ubuntu1~24.04.1

marcocondrache:
Closes #29073

GPUI continuously presents frames at the display refresh rate for one second after any input to prevent display underclocking [(window.rs:1115-1118)](https://github.com/zed-industries/zed/blob/82b768258f966c157ed960d19ae2d4bdce81779f/crates/gpui/src/window.rs#L1116). Since mouse movement counts as input and resets the timestamp, the window presents indefinitely while the cursor moves. Display underclocking applies only to technologies like ProMotion or Adaptive Sync; therefore, enabling it by default is illogical and results in increased GPU usage

Release Notes:

- Reduced GPU usage on displays without variable refresh rate


marcocondrache:
I would also consider reducing the interval at:

https://github.com/zed-industries/zed/blob/8e04706c4d2b5ef454e1c1ffb9d5aac72118b55d/crates/gpui/src/window.rs#L1116

I believe most displays can ramp up quickly, in less than 1 second. Reducing it to around 500ms would halve GPU waste.

marcocondrache:
Ref: #7347
Ref: https://zed.dev/blog/120fps

marcocondrache:
There’s still one issue I haven’t figured out how to solve:

On Macs with ProMotion, this still causes elevated GPU and power usage. When the cursor moves, the frame counter resets and the window presents. The display’s refresh rate stays high because the GPU is doing work, which makes sense, but isn’t ideal?

marcocondrache:
Testing on macOS with non-VRR displays shows that VRR detection is behaving as expected. The odd part is what happens on high-refresh-rate non-VRR monitors (e.g., 240 Hz). We’re not getting anywhere close to the maximum frame rate. Earlier, the “input reset” path unintentionally forced constant presents, which made Metal run at full speed. Without that behavior, the frame rate dips to ~30–40 fps during scrolling and then jumps to 60–70 fps at random. Needs further investigation.

Side note: [CVDisplayLink](https://developer.apple.com/documentation/corevideo/cvdisplaylink-k0k?language=objc) is now essentially deprecated, so relying on it for timing is becoming a shaky approach.

marcocondrache:
I made some progress on this. I haven't directly investigated why this happens but i've made some research on what to use after `CVDisplayLink`. 

From macOS 14+ there is a new class that can be used by Metal apps to synchronize with displays (both VRR and non-VRR). `CAMetalDisplayLink` not only provides timing based on the display refresh rate but is also able to push drawables to us directly. This means the drawable is no longer requested from the `CAMetalLayer` but provided by the display link callback itself. This reduces latency, improves timing, and ensures we only commit and draw when the system asks:

- Native VRR support: Full synchronization with ProMotion displays up to 120Hz without workarounds (like the last_input_timestamp)
- Reduced GPU and power usage: Metal can optimize scheduling since it knows exactly when we'll draw

To make a more understandable comparison:

Now:

- `CVDisplayLink` fires at vsync
- We call `nextDrawable()` (may block)
- We render and present

With `CAMetalDisplayLink`:

- System determines optimal render time
- `CAMetalDisplayLink` calls our delegate with the drawable
- We render and present

This is outside the scope of this PR, but it has potential benefits for Zed. I would appreciate your opinions, @mikayla-maki and @reflectronic, and I can create another PR if needed.

marcocondrache:
@as-cii sorry for tagging you directly. I noticed you self-assigned the original issue. If there’s any way I can support you on it, I’d be happy to help.

PR #43647, opened 2025-11-27
project: Make individual DAP settings optional

fredr:
Currently, when overriding e.g a `binary` in dap settings, one have to override all settings, including `args` as it otherwise set it to empty string.

E.g this config will set `args` and `env` to their empty states, like empty string and empty map:
```json
{
  "dap": {
    "Delve": {
      "binary": "/path/to/binary"
    }
  }
}
```

This PR changes so that it is instead possible to only override a single field. This is useful e.g on nixos where you don't want (or cant let) zed download binaries for you, but you want to use the default args, such as in Delve, you want zed to configure it with a random listening port.

Release Notes:

- Improved dap settings so that it is possible to override individual fields, such as `binary`

### DIFF
--- a/crates/project/src/debugger/dap_store.rs
+++ b/crates/project/src/debugger/dap_store.rs
@@ -266,8 +266,8 @@ impl DapStore {
                         Some(worktree.read(cx).resolve_executable_path(path))
                     }
                 });
-                let user_args = dap_settings.map(|s| s.args.clone());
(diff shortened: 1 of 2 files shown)


fredr:
@cla-bot check

MrSubidubi (approved):
Sorry for the delay, thanks! And congratulations to your first contribution! 🎉

PR #43661, opened 2025-11-27
ai: Implement OpenAI compatible event mapper to parse thinking tags

aeroxy:
To support Minimax M2.

Release Notes:

- Implement OpenAI compatible event mapper to parse thinking tags

### DIFF
--- a/crates/language_models/src/provider/open_ai_compatible.rs
+++ b/crates/language_models/src/provider/open_ai_compatible.rs
@@ -343,10 +343,199 @@ impl LanguageModel for OpenAiCompatibleLanguageModel {
         );
         let completions = self.stream_completion(request, cx);
         async move {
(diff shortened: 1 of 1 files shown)


bennetbo:
I don't think this is something we want to do right now, as there is no API definition around it and it will make the provider definitions increasingly more complex. If we do this, we also have to do it for other tags (e.g. <thinking>) and other providers for consistency. We'll re-consider this once we port the API providers to extensions.

aeroxy:
> I don't think this is something we want to do right now, as there is no API definition around it and it will make the provider definitions increasingly more complex. If we do this, we also have to do it for other tags (e.g. ) and other providers for consistency. We'll re-consider this once we port the API providers to extensions.

good plan. may i suggest a openai-like provider in which we can support all the quirks from different models?

PR #43662, opened 2025-11-27
open_ai: Support reasoning content

aeroxy:
Support for Kimi K2 Thinking

Release Notes:

- OpenAI provider supports reasoning content

### DIFF
--- a/crates/language_models/src/provider/open_ai.rs
+++ b/crates/language_models/src/provider/open_ai.rs
@@ -556,7 +556,6 @@ impl OpenAiEventMapper {
         };
 
         if let Some(delta) = choice.delta.as_ref() {
-            // Handle reasoning content (thinking blocks)
             if let Some(reasoning_content) = delta.reasoning_content.clone() {
                 if !reasoning_content.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/language_models/src/provider/open_ai.rs
```diff
@@ -363,7 +363,7 @@ pub fn into_open_ai(
         for content in message.content {
             match content {
                 MessageContent::Text(text) | MessageContent::Thinking { text, .. } => {
-                    if !text.trim().is_empty() {
+                    if !text.is_empty() {
```
bennetbo: Why was this changed?

aeroxy: Kimi K2 Thinking will response empty spaces and if we do not emit them the thinking block will look like this:

_KimiK2Thinkingwillresponseemptyspacesandifwedonotemitthemthethinkingblockwilllooklikethis_

bennetbo:
Thank you! I adjusted the `text.trim().is_empty()` -> `text.is_empty()` change in order to not regress #40097.

bennetbo (approved):

Issue #9755, opened 2024-03-24
Add support to eslint in monorepos


znycheporuk:
### Check for existing issues

- [X] Completed

### Describe the feature

In monorepo where each package has it's own .eslintrc, only one is used (closest one to file that was opened first in the workspace). They are configured properly, and work fine individually.

### If applicable, add mockups / screenshots to help present your vision of the feature

https://github.com/zed-industries/zed/assets/42861377/27c35d59-7b8e-4653-810b-bb9c45507cbc

Moshyfawn:
This is likely to be solved by closing #9648 via #9738.

praseodym:
Reproducer for this issue using ESLint 8.57.1 with a non-flat eslintrc: https://github.com/praseodym/zed-eslint-monorepo-reproducer

If there is any Zed setting that can be used to get ESLint to work, please let me know!

andirsun:
This is bummer, I had to switch back to Vscode for this.

OmChillure:
@esthertrapadoux can I take up this issue?

Have been reserching on this to fix this!

andirsun:
Maybe this was fixed by https://github.com/zed-industries/zed/pull/52886 ? @znycheporuk

yeskunall:
Thanks for flagging this @andirsun, you’re right -- I believe https://github.com/zed-industries/zed/pull/52886 landed support for this.

pantheraleo-7:
if `eslint` is not installed in the workspace, it doesn't fallbacks to Zed's installation. the server fails with:
```
Uncaught exception received.
Error: Unrecognized method `eslint/noLibrary`
    at handleResponse (/Users/asadullahshaikh/Library/Application Support/Zed/languages/eslint/vscode-eslint-3.0.24/vscode-eslint/server/node_modules/vscode-jsonrpc/lib/common/connection.js:606:48)
    at addMessageToQueue (/Users/asadullahshaikh/Library/Application Support/Zed/languages/eslint/vscode-eslint-3.0.24/vscode-eslint/server/node_modules/vscode-jsonrpc/lib/common/connection.js:382:17)
    at StreamMessageReader.callback (/Users/asadullahshaikh/Library/Application Support/Zed/languages/eslint/vscode-eslint-3.0.24/vscode-eslint/server/node_modules/vscode-jsonrpc/lib/common/connection.js:482:13)
[...]

PR #43677, opened 2025-11-27
language: Allow support for ESLint working directories

### LINKED ISSUE #9648: Allow configuring ESLint working directories
### Check for existing issues

- [X] Completed

### Describe the feature

Currently, the value of `workingDirectories` property is hardcoded (https://github.com/zed-industries/zed/blob/main/crates/languages/src/typescript.rs#L279). This property is useful for configuring ESLint in some monorepo/library scenarios. It is also very important to have this property configurable to be able to open existing projects without requiring code changes.

P.S. It would also be nice to extend the possibility of providing all other configurations that https://github.com/microsoft/vscode-eslint supports

_No response_

smitbarmase:
Closes #9648 #9755

Release Notes:

- Added way to configure ESLint's working directories in settings. For example: `{"lsp":{"eslint":{"settings":{"workingDirectories":["./client","./server"]}}}}`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9012,6 +9012,7 @@ dependencies = [
  "chrono",
  "collections",
  "futures 0.3.31",
+ "globset",
  "gpui",
  "http_client",
  "itertools 0.14.0",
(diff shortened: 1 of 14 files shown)


osiewicz:
Nice!
Nit: We should prolly remark in the rel-note that our `workingDirectories` setting matches VSCode's syntax.

PR #43693, opened 2025-11-27
git: Improve handling of unsafe repositories

### LINKED ISSUE #42286: Git: Git features aren't showing on version 0.211.6 on Windows
### Summary

When I open any project that has git source control, the Git panel says that there is no branch and there are no

### Description
All of my projects have a .git file but when I open them with zed I can view branches, commits, untracked files, modified files and other git features provided by zed. I opened the projects in vscode and cursor and the git features were showing. Before updating to the last version of zed everything was showing.

Steps to trigger the problem:
1. Open a project or workspace where git is initialized.
2. Go to git panel to view branches, commits or check current branch name in the title bar.

**Expected Behavior**: Git features showing properly

**Actual Behavior**: Git features not showing

This is my first issues ever, I hope I described the issue very well😅

Zed: v0.211.6 (Zed) 
OS: Windows 10.0.19045
Memory: 3.8 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 605 || Intel Corporation || 25.20.100.6519

### Additional (screenshots)

Image
Image

dinocosta:
These changes update Zed's Git Panel to be able to detect unsafe git repositories, where the user running Zed doesn't own the `.git` directory, and show a dedicated empty view in the Git Panel that not only explains the situation but also allows the user to choose whether to trust this directory, which will end up running `git config --global --add safe.directory <path>`.

While testing those changes, it was noted that attempting to add or reset files after trusting the directory would fail, as expected, but the UI wouldn't react to the fact that those operations failed. What this means is, if the user tried to add a file, the UI would show a checkmark for that file and, after the operation failed, the checkmark would remain. We now revert that, for both `git add` and `git reset`.

Some more technical notes on this change:

- Introduce `project::git_store::GitAccess` enum to express whether we actually have access to the repository, exposed via `git::Repository::access`, which probes the backend with a `git status` command and classifies the result.
- On unsafe repositories, `project::git_store::LocalRepositoryState::new` fails to spawn the git worker, which now cleanly signals `GitAccess::No` rather than leaving the panel in a broken state.
- Updated `git_ui::git_panel::GitPanel::render_empty_state` with a third alternative, when `GitAccess::No`, that basically renders a view explaining why the repository is considered unsafe, together with buttons for git's documentation on safe directories and a button to add the repository's folder as a safe directory
- Updated `project::git_store::GitStore` to now watch `~/.gitconfig` and `$XDG_CONFIG_HOME/git/config`. Watching this files allows the Git Panel to react when the git configuration is updated, which will be the case if the user decides to trust the repository. When either changes, a `GitStoreEvent::GlobalConfigurationUpdated` event is emitted and the panel refetches repository state.
- Added `project::git_store::Repository::refetch_repo_state` field, which stores a closure to allow recreating the `Repository::repository_state` and the job sender after the repository's directory is trusted, without requiring a project reload.
- Added a `fs::Fs::git_config` trait method, wrapping a real `git config` invocation. In order to be able to call this, both `git_store::GitStore::git_config` and `project::Project::git_config` wrappers have also been introduced. Worth mentioning that this isn't yet supported for remote projects or collab guests.
- Updated `fs::Fs::git_clone` argument order to match `fs::Fs::git_config` and `fs::Fs::git_init`.
- Added a new `project::git_store::pending_op::PendingOps::last_op_errored` method that allows determining whether the last pending operation failed. This now allows us to filter out failed operations when determining whether `git add` or `git reset` failed, so that we can fall through to the real git status.
- Updated `git_ui::git_panel::GitPanel::change_file_stage` to now call `update_counts` in its error branch so the cached staged counters stay consistent with the reverted per-entry state, seeing as we now handle reverting the UI state if an operation fails.
- Fixed `fs::RealFs::git_init` to fall back to the provided branch name when `git config --global --get init.defaultBranch` fails, for example, when the user hsan't configured one.

Co-authored-by: cameron <cameron.studdstreet@gmail.com>

Closes #42286

Release Notes:

- Added a dedicated empty state in the Git Panel for unsafe repositories, with a "Trust Directory" button that adds the repository to `safe.directory`
- Fixed stage and unstage checkboxes in the Git Panel not reverting when a `git add` or `git reset` command failed


dinocosta (commented):
I've just updated this Pull Request with `main` and ensured that the proposed functionality is working as intended for untrusted repositories, that is, when the `.git` folder is not owned by the user that is running the application.

Although this is now back in a useful state, I believe there's some points we'd want to consider/review before shipping this:

~1. Compare the current `GitPanel::render_empty_state` implementation with `main`. I believe `main` introduces new elements, related with showing the diff button, so we'll want to preserve that.~ Tackled in https://github.com/zed-industries/zed/pull/43693/commits/5db89d34bf85f4e54d18c7d72b175c0e35772c99 .
2. Update handling of failed git operations, like `git add` and `git reset`. Even if the repository is trusted, i.e., it's present in the `safe.directory` config value, which means the user can run `git status`, attempting to add or reset files will fail, but the UI is not reacting to that, when adding or reseting a file through the Git Panel, the UI assumes it ran successfully even though an error toast is shown.
3. Decide whether we want to tackle this on Remote. Right now we're just assuming that the user has access to the repository such that, if we were to merge this today, it would still work as if this feature didn't ship. We have to decide whether to commit the resources into testing the remote path, or not.

cameron1024 (approved):

crates/project/src/git_store/pending_op.rs
```diff
@@ -121,15 +121,21 @@ impl PendingOps {
         false
     }
 
-    /// File is staged if the last job is not finished and has status Staged.
+    /// File is currently staging if the job is still running and has status Staged.
     pub fn staging(&self) -> bool {
         if let Some(last) = self.ops.last() {
-            if last.git_status == GitStatus::Staged && last.job_status != JobStatus::Finished {
+            if last.git_status == GitStatus::Staged && last.job_status == JobStatus::Running {
```
kubkon: Is this change necessary?

dinocosta: Sorry, I don't think it's no longer necessary. When I changed this I hadn't yet updated the call sites to use `last_op_errored`, since we now first check if the last operation errored, we don't end up calling this method.

With the original version this method would still return `true` if the operation had failed, as it's simply checking `last.job_status != JobStatus::Finished`, which is still `true` for errored jobs (`JobStatus::Error`). 

Let me know if you want me to revert this change 🙇

kubkon: Yeah, let's revert it then. Otherwise, approved! Feel free to merge.

dinocosta: Thank you Jakub! Reverted with https://github.com/zed-industries/zed/pull/43693/changes/763405705c8316d87b13cb744343060cbb7b8b55 ✅

PR #43701, opened 2025-11-28
git: New actions for git panel navigation

coszio:
I could not find any related issue, but at least I want to use the git panel like this :)

Being used to `lazygit`, this PR makes navigation of the git panel more similar to the CLI tool.

Instead of selecting -> enter'ing for skimming each file, I just want to move between the files in the git panel and have the diff multibuffer advance to the appropriate file. This also adheres to the behavior of the outline panel, which I like better.

If the multibuffer is not active, it behaves same as before (just selecting the file in the panel, nothing else).

I did not modify existing `menu::Select*` actions in case anybody still prefers previous behavior.

https://github.com/user-attachments/assets/2d1303d4-50c8-4500-ab3b-302eb7d4afda

Release Notes:

- Improved navigation of the git panel, by advancing the "Uncommitted Changes" multibuffer to the current selected file. To restore the old behavior, you can bind `up` and `down` to `menu::SelectPrevious` and `menu::SelectNext` under the `GitPanel` context in your keymap.


crates/git_ui/src/git_panel.rs
```diff
@@ -701,16 +709,21 @@ impl GitPanel {
...
-    fn select_first(&mut self, _: &SelectFirst, _window: &mut Window, cx: &mut Context<Self>) {
+    fn select_first(
+        &mut self,
+        _: &menu::SelectFirst,
+        _window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
         if !self.entries.is_empty() {
-            self.selected_entry = Some(1);
+            self.selected_entry = Some(0);
```
coszio: Also fixes this and `select_first_entry_if_none`, where they would select the second one (0-based indexing)

cole-miller: I think the 1 is intentional, because the first entry is a header

coszio: It is incorrect behavior for git panel, since we don't have headers there. It's easily reproducible by `cmd-up` on the file selector from git panel

cole-miller: Thanks, good point. We do have headers in the git panel in the default ("sort by status") view:

image

So using `0` is incorrect for that mode, but you're right that it's correct for the "sort by path" view which doesn't have headers. I believe this has been fixed for all view modes on main by @RemcoSmitsDev.

cole-miller:
Thanks!

crates/git_ui/src/git_panel.rs
coszio: Ah right, I missed checking the "sort by status" view.

PR #43709, opened 2025-11-28
editor: Reduce amount of sumtree traversals in `header_and_footer_blocks`

Veykril:
Introduces new "mapping point cursors" for the different display map layers allowing one to map multiple points in increasing order more efficiently than using the one shot operations. 

This is used in the `BlockMap::sync` for `header_and_footer_blocks` which spends a significant time in sumtree traversal due to repeatedly transforming points between the different layers. This effectively turns the complexity of those operations from quadratic in the number of excerpts to linear, as we only go through the respective sumtrees once instead of restarting from the start over and over again.

Release Notes:

- Improved performance for editors of large multibuffers with many different files


maxbrunsfeld (commented):
Great idea ⚡

PR #43719, opened 2025-11-28
Tweak the issue template for bugs

zelenenka:
- shorten some phrases to (hopefully) retain attention better
- move the required field up to not bore people with scrolling
- call it a bug and not an issue to further emphasize we want feature requests in Discussions where the product managers are

Release Notes:

- N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/1.bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1.bug-report.yml
@@ -24,7 +24,7 @@ body:
     attributes:
       label: Current vs. Expected behavior
       description: |
-        Current behavior (screenshots, videos, etc are appreciated), vs. what you expected the behavior to be.
+        Current behavior (e.g., screenshots, videos, are appreciated), vs. what you expected the behavior to be.
 
       placeholder: |
         Current behavior: <screenshot with an arrow> The icon is blue. Expected behavior: The icon should be red because this is what the setting is documented to do.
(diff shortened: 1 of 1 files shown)


yeskunall (approved):
As they say, less is more.

.github/ISSUE_TEMPLATE/1.bug-report.yml
```diff
@@ -1,39 +1,50 @@
...
         2. Click X
-        3. Y will happen
     validations:
       required: true
   - type: textarea
     attributes:
       label: Current vs. Expected behavior
       description: |
-        A clear and concise description of what is the current behavior (screenshots, videos), vs. what you expected the behavior to be.
+        Current behavior (screenshots, videos, etc are appreciated), vs. what you expected the behavior to be.
```
miguelraz:
```suggestion
        Current behavior (e.g., screenshots, videos, are appreciated), vs. what you expected the behavior to be.
```

Pedantic nitpick time but I think `e.g.` fits better here, since `etc.` (which always required a `.`) [should terminate a clause](https://www.grammarly.com/blog/commonly-confused-words/et-cetera-etc/).

miguelraz (approved):

.github/ISSUE_TEMPLATE/1.bug-report.yml
yeskunall: O I like this. But I’m mad that I had to use Grammarly as a reference.

PR #43732, opened 2025-11-28
Fix slow rebuild time for dev extensions with grammar compilation

boris-schwarz:
This is a proposal hotfix for the discussion [#43685](https://github.com/zed-industries/zed/discussions/43685)

Release Notes:

- When installing or rebuilding a dev extension, the compiling of grammars is skipped, if the compiled grammars are already present on the file system.

What it's solving:

When trying to fix the zed extension [zed-cfml](https://github.com/cfmleditor/zed-cfml) that uses the [tree-sitter-cfml](https://github.com/cfmleditor/tree-sitter-cfml) grammar, I came across build times around 2 minutes when installing the extension locally as a dev extension. After reading trough the code, I realized `crates/extension/src/extension_builder.rs:compile_grammar()` is called, despite having already compiled. I assume this makes sense for deploying extension, but in the local development process, it's slowing the developer experience a lot.

How it's solving:

- Extended the `CompileExtensionOptions` with a new boolean flag `dev_extension` that is only set to true when invoking either `install_dev_extension()` or `rebuild_dev_extension()`.
- If `dev_extension` is set to true, it checks, wether the compiled wasm is present on the file system.
- If the compiled wasm is present, the compiling step is skipped.
- This ensures that the process for normal extension stays the same.

Changes in Developer Experience:

Faster installation and rebuild times for dev extension at the cost of having to manually delete the desired foo_grammar.wasm in the extension's folder.
Note: This is already the case when changing the grammar's commit/rev, the developer will have to manually delete the grammars/foo_grammar/ folder.

Changes in UI:
- N/A

### DIFF
--- a/crates/extension/src/extension_builder.rs
+++ b/crates/extension/src/extension_builder.rs
@@ -52,6 +52,7 @@ pub struct ExtensionBuilder {
 
 pub struct CompileExtensionOptions {
     pub release: bool,
+    pub dev_extension: bool,
 }
 
 #[derive(Deserialize)]
--- a/crates/extension/src/extension_builder.rs
+++ b/crates/extension/src/extension_builder.rs
@@ -79,6 +79,8 @@ impl ExtensionBuilder {
         extension_manifest: &mut ExtensionManifest,
         options: CompileExtensionOptions,
     ) -> Result<()> {
+        let dev_extension = options.dev_extension.clone();
+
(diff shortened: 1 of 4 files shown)


boris-schwarz:
@cla-bot check

MrSubidubi:
Thanks for this! I am gonna close this in favor of https://github.com/zed-industries/zed/pull/43442, as I think this is a more sensible approach.

PR #12106, opened 2024-05-21
[WIP] Easy motion

babyccino:
Requested at #4930

Release Notes:

- WIP: add easy motion #4930

Current state:

- By word (word(w) and full word(W) are working)
- Row mode is working

PR #43733, opened 2025-11-28
helix: Add Helix's "Amp Jump" Feature

GoldStrikeArch:
# Helix Amp Jump Navigation

## Overview

Implements Helix-style "amp jump" (`g w`) navigation for jump-to-word functionality. This feature displays two-character labels on each word in the visible area, allowing users to quickly jump to any word by typing its label. Labels alternate between forward and backward directions (same algorithm as in the Helix) from the cursor position, giving closer jump targets easier-to-type labels.

## Context

- **Request:** Implement "amp jump" navigation similar to Helix editor's jump-to-word feature.
- **Scope:** Full implementation including label generation, UI rendering, and input handling.
- **Inspiration:** Helix editor's "amp jump" and vim-easymotion/hop.nvim plugins.

## How It Works

1. Press `g w` to activate "amp jump"
2. Two-character labels appear on all words in visible area
3. User types the two-character label shown on the target word
4. Cursor jumps to that word

img

Release Notes:

- Added in Helix mode the "amp jump" navigation (`g w`) that displays two-character labels on words for quick cursor navigation. Labels alternate between forward and backward directions from the cursor, prioritizing closer targets with easier-to-type labels. The color of the labels can be controlled via a new `helix.jump_label_accent` setting


baldwindavid:
I like this feature a lot. To me, it's the only thing that you _really_ need from all these tools like easymotion, leap, hop, etc. Yes, this is the limited feature Helix happened to add, but I don't really think of it as a Helix feature as much as a general modal editing feature. As such, if this is added, I'd prefer it be available to vim too.

GoldStrikeArch:
> I like this feature a lot. To me, it's the only thing that you _really_ need from all these tools like easymotion, leap, hop, etc. Yes, this is the limited feature Helix happened to add, but I don't really think of it as a Helix feature as much as a general modal editing feature. As such, if this is added, I'd prefer it be available to vim too.

We can extend it to the vim as well (in the separate PR) but there might be some conflicts of how people expect it to work compared to something like [easymotions](https://github.com/easymotion/vim-easymotion). The current implementation is how Helix implemented it and as I never used neovim/vim I don't really know how it should be implemented there 😅

baldwindavid:
Seems reasonable. Just know that there is some history/discussion on this sort of thing in Zed... https://github.com/zed-industries/zed/pull/12106

That it's a much smaller diff may help. And I think it might help to sell it as something that _could_ be the builtin modal editing label navigator. Not saying you need to sell it that way at all. But I use vim, zed, and helix and think it's a great general fit for any of them.

Dima-369:
I tested it and since I am using a proportional font via `buffer_font_family` (I am using `Avenir Next`) in `settings.json`, the hints look a bit weird with some overlapping others because every character has a potential different width.

I wonder if there is any way to improve this for proportional fonts? Maybe there isn't? 

Just leaving this here, since I am also interested in such a feature. Thanks for the PR!

image

GoldStrikeArch:
@Dima-369 

Can you please share the whole section where you font is configured in `settings.json` (or the whole file, either is good)? I will take a look at it

> I wonder if there is any way to improve this for proportional fonts? Maybe there isn't?

Also, I am not sure that I understand what is `proportional font`?

Dima-369:
> Also, I am not sure that I understand what is proportional font?

There are monospace fonts, and there are fonts where every character does not have a fixed width. In monospace fonts, `i` takes the same width as `w`. In proportional fonts, they are a different width.

This is my config, but you need to make sure that you have access to this font. If not, I am sure you can reproduce with any other proportional font.

```json
{
"buffer_font_family": "Avenir Next",
"buffer_line_height": { "custom": 1.4 },
"buffer_font_size": 18
}
```

Dima-369:
> Can you try again now please?

Yes, it is okayish. Some hints are smaller for whatever reason.

CleanShot 2026-01-02 at 12 42 38@2x


> I guess there is no easy solution for the proportional fonts

One could instead always show the characters below (instead of hiding them), and draw the hints with a rounded background color over the characters. Then ensure that the overlaid hints do not overlap each other, but then the UI is different to the Helix implementation.

---

Anyway, let's see what the Zed team says about this 🙂

GoldStrikeArch:
@kubkon 
> Can you fix formatting and rebase on latest `main` please?

I merged (not rebased but I can do the rebase if this is mandatory) the latest main to this branch and also run the `cargo fmt` to fix the issues with formatting

GoldStrikeArch:
> I think both of them need the same kind of editor UI where we show overlays for jump targets.

@ConradIrwin by "editor UI" you mean some core API for doing things like that or an actual "user configured UI"?

GoldStrikeArch:
> The current sneak motion doesn't offer visual guidance or a default key binding. But it works. These solutions are not extension-based, so I'd prefer a final jump feature that covers all previous and new solutions.

@yilmazer This PR is about a different thing. The "amp jump" navigation is different from `Sneak` motion (the only common things are "labels"). a "final jump feature" was never a goal in this PR, here we are just implementing a particular feature Helix editor has (the `gw` navigation). And IMO there should be some broader discussion if we should have a "unified way" to have a "jump" features as people from Vim/Neovim might expect different behaviors (neovim has a lot of different plugins like `easymotions` which are doing something similar yet all of them have their variations in behavior)

ConradIrwin:
Nice, thanks! will take a look (hopefully before Friday!)

GoldStrikeArch:
Hi @ConradIrwin,

I'm really sorry I couldn't make our pairing session today due to an unexpected personal matter. I have gone ahead and rescheduled it for Tuesday, April 14.

Thanks for understanding, and I'm looking forward to catching up then!

ConradIrwin:
Sounds good, see you then!

GoldStrikeArch:
@ConradIrwin Hi!

As per our conversation [here](https://github.com/zed-industries/zed/pull/52630#issuecomment-4254413842) I adjusted the current PR to use a new `NavigationOverlay`, so IMO we can proceed with what we have right now. The most changes are still in `helix.rs` file and now we don't touch anything in "editor core"

ConradIrwin:
Thanks. Let's put the color in the theme (not the settings), but otherwise think we can merge this.

Historically we've re-used similar colors, so feel free to do that (or fallback as appropriate)

GoldStrikeArch:
@ConradIrwin 

> Let's put the color in the theme (not the settings), but otherwise think we can merge this.

Done. Now it sits inside the theme colors and you can override it. Otherwise it will use the `status.error` which is usually "red color" in most themes (and is kinda what Helix is using by default as well).

ConradIrwin:
Alright, this seems to be working well. Let's ship it.

Thanks again for working on this, and hopefully the next thing will take less than 6 months to merge.

redzovic:
Will it always be 2 characters or can this be changed via some setting to behave like nvims flash?

niekdomi:
@redzovic based on my understanding this PR here is helix specific and (likely) can't be configured, though with https://github.com/zed-industries/zed/pull/52630 afaik it should be possible to write plugins for any kind of navigation e.g., flash. At least at some point.

GoldStrikeArch:
> Will it always be 2 characters or can this be changed via some setting to behave like nvims flash?

@redzovic  Yes, it will always be 2 characters as this is "feature parity" with Helix editor, it will also only work in Helix mode. I don't know how nvim flash works, but I know that there are other things like easymotions and similar plugins which all is doing this type of "jump label navigation" differently

ksweetie:
FWIW, it works in vim mode, you just have to set it to a keybind. This works for me:

```json
  {
    "context": "VimControl && !menu",
    "bindings": {
      "g w": "vim::HelixJumpToWord",
    },
  }
```

I've never used helix before, but one thing I found a little surprising is that instead of jumping the cursor to the label, it jumps to the end of the word that the label starts on, with the word selected (visual mode). I find myself always pressing `esc` and `b` afterwards to exit visual mode and move backward to where the label was. I assume that's how it works in helix. My preference would be to jump straight to the label, and to stay in the same mode that you started in (normal or visual). But I don't know if that's possible with the current setup.

Thanks for your work shipping this feature!

Malix-Labs:
Yes, that's how selection→action motions work

GoldStrikeArch:
@ksweetie 

> FWIW, it works in vim mode, you just have to set it to a keybind. This works for me:

TBH I didn't knew it can work with Vim mode instead of Helix mode 😅 

If we really want to make it compatible with Vim there should be some refactoring, so it can be used with both modes (in vim you will not select the word as in the Helix it is a default). Feel free to open a new issue and I will do it :)

k-soomro:
> @ksweetie
> 
> > FWIW, it works in vim mode, you just have to set it to a keybind. This works for me:
> 
> TBH I didn't knew it can work with Vim mode instead of Helix mode 😅
> 
> If we really want to make it compatible with Vim there should be some refactoring, so it can be used with both modes (in vim you will not select the word as in the Helix it is a default). Feel free to open a new issue and I will do it :)

Thank you1 I've been waiting for this feature and only upgraded to the preview version for it. I came here looking exactly for a way to make it behave more like vim/neovim

GoldStrikeArch:
@ksweetie I did implement a proper binding to `g z` (because `g w` is already taken in Vim for rewrap) and it works without selection [here](https://github.com/zed-industries/zed/pull/55492), please check out that branch and see if it works for you or not

med1844:
Is it possible to make `HELIX_JUMP_ALPHABET` something configurable so that we can order the alphabet accordingly to minimize finger movement based on our own keyboard layouts?

Sorry for commenting in a closed PR but I can't find an issue for this.

failable:
Can't it work without `Vim` mode enabled?

GoldStrikeArch:
@med1844 

>  something configurable so that we can order the alphabet accordingly to minimize finger movement based on our own keyboard layouts?

Good idea! Go ahead and create an issue/feature request for it, it is somewhat easy to add


@failable 

> Can't it work without `Vim` mode enabled?

It is supposed to be work with Helix mode though. But no, you can't use it outside of Helix/Vim

med1844:
Thanks, I have created discussion #56128 to keep track of it.

atahrijouti:
Enormous thanks for this  🤩🤩

sanket143:
Just saw this in the release notes, thanks a bunch for this ✌🏻

PR #8646, opened 2024-03-01
Wire up Azure OpenAI completion provider

maxdeviant:
This PR wires up support for [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-services/openai/overview) as an alternative AI provider in the assistant panel.

This can be configured using the following in the settings file:

```json
{
  "assistant": {
    "provider": {
      "type": "azure_openai",
      "api_url": "https://{your-resource-name}.openai.azure.com",
      "deployment_id": "gpt-4",
      "api_version": "2023-05-15"
    }
  },
}
```

You will need to deploy a model within Azure and update the settings accordingly.

Release Notes:

- N/A

PR #8493, opened 2024-02-27
Hacky support for Azure

tombailey:
Release Notes:

- Added hacky support for Azure OpenAI

I don't have an Azure account / API key to test with so not sure if it works

PR #43742, opened 2025-11-28
Add Azure Foundry support

ryanravn:
Release Notes:

- Added support for Azure Foundry Anthropic endpoints, including custom API URLs, API keys, and model selection (Sonnet 4.5 / Opus 4.5).

This enables using Azure-hosted Anthropic models without needing a proxy.

I've just focused on the Anthropic models for now, but for a fully fledged implementation it's likely desirable to have more.

Related:
- #12296
- #8493
- #20232
- #8646


ryanravn:
@cla-bot check

maxdeviant:
Thank you for the PR!

We're in the process of making it possible to add new model providers via extensions. This should be ready soon, at which point Azure Foundry support can be added through that mechanism.

IshamMohamed:
> Thank you for the PR!
> 
> We're in the process of making it possible to add new model providers via extensions. This should be ready soon, at which point Azure Foundry support can be added through that mechanism.

When will this be ready?

Why can't you support PRs similar to #61199 for Azure Foundry?

PR #43750, opened 2025-11-28
Add LSP call hierarchy panel and support

### LINKED ISSUE #14203: Add 'Show Call Hierarchy' panel
### Check for existing issues

- [X] Completed

### Describe the feature

It seems to me that this feature is not yet implemented, please correct me if I'm wrong. Anyway I would love to have the same "call hierarchy" view which is curretly offered by VScode as show in their [docs](https://code.visualstudio.com/docs/java/java-editing#_call-hierarchy).

call-hierarchy-view-demo

aohoyd:
Closes #14203

Release Notes:

- Added Call Hierarchy panel for exploring function callers and callees via LSP
- Added Call Hierarchy modal
- Added Settings menu for the new panel

I've been missing this feature from vscode and goland. Call hierarchy is useful for navigating complex codebases. This is invaluable when refactoring or trying to understand unfamiliar code.

Some parts of code (ui and tests) are AI generated, but I reviewed them before commiting.

I understand that the feature is pretty complex and it's ok if you'll ask to split it or won't accept it completely.


aohoyd:
@cla-bot check

versecafe:
@aohoyd tried it out, couldn't get it working for TS but in Go it worked nicely out the box love to see it, initial feedback on the actual feature impl not the code is I'd recommend adding the call hierarchy to the menu when someone pulls up a context window on a symbol rather than just having it via keybind and command palette. On the UI side not really your fault but exists with the symbol viewer panel too the symbols are awkwardly large compared to the rest of the UI and it should probably have a different icon then the outline viewer.

image

aohoyd:
@versecafe What didn't work in TS? I've tried it out on VS Code repo and everything worked. The only thing is that it depends on LSP, so if it takes time to initialize and load then panel will be empty until LSP is loaded
CleanShot 2025-12-01 at 22 27 55@2x

Also I've fixed font configuration, added new action in context menu and added new icon. I'm not sure that it matches style of other icons, but I think it's ok for the start

versecafe:
@aohoyd which of the TS language servers did you use for it I'll try again with it set

maxbrunsfeld (changes requested):
@aohoyd This is a good feature. I 100% want it in Zed. Thank you for figuring all of this out so far.

I have some changes I'd like to see. Let me know your reaction to these:

* **Open the hierarchy as a normal tab, not a side panel** - I think that this feature would fit much better into Zed's existing UX patterns if we put it in its own tab, rather than a panel. My reasoning:
  - That's what we currently do for project search, find-all-references, and refactoring results. 
  - Adding a new item to the status bar has the drawback that it takes up space down there all of the time, even though you are rarely looking at a call hierarchy.
  - This is subjective, but in my opinion, when inspecting a call hierarchy, it is often the main center of your focus. It is the code itself. It's not some navigational metadata.
  - Finally, the results are often pretty wide. They just don't fit that well in a small sidebar.

* **Format the list items** - When testing this out (in my case, using `rust-analyzer`), I'm seeing a lot of visual redundancy in the symbols list. Generally a function name will appear (with the correct function syntax highlighting), followed immediately by `fn`, and *another* copy of the function name without any syntax highlighting.

    Image

    Ideally, I would like to post-process these items so that they show up similarly to how they look in the outline view, or in completions. That might require some LSP-specific code similar to `labels_for_symbols` and `labels_for_completions`, which know how to format a given language server's outputs in the *Zed way*.

* **Tests for the feature** - We need a basic integration test in `crates/remote_server/src/remote_editing_tests.rs` that checks that this feature works correctly when remote editing.

* **Keyboard Navigability** - I need a way to return focus to the Call Hierarchy view from the keyboard. If it's in a pane, I think we could get that for free.

MrSubidubi:
grafik

We would love to have this at some point! Yet, given no activity for quite some time, the unadressed review above and the conflicts, gonna close this out to keep the queue cleaner. Would be very happy to repoen this should you decice to pick this up again. Thanks!

PR #42819, opened 2025-11-15
Add support for git remotes

bnjjj:
Follow up of #42486 
Closes #26559

https://github.com/user-attachments/assets/e2f54dda-a78b-4d9b-a910-16d51f98a111

Release Notes:

- Added support for git remotes

PR #43751, opened 2025-11-28
ui: Do not selected an index for the `ToggleButtonGroup` by default

bnjjj:
**ToggleButtonGroup**: Changed default behavior to not pre-select any button index
  - Changed `selected_index` field from `usize` to `Option<usize>`, defaulting to `None` instead of `0`
  - This prevents the first button (index 0) from being automatically selected when no explicit selection is made

Release Notes:

- Fixed: changed default behavior to `ToggleButtonGroup` to not pre-select any button index

### DIFF
--- a/crates/ui/src/components/button/toggle_button.rs
+++ b/crates/ui/src/components/button/toggle_button.rs
@@ -179,7 +179,7 @@ where
     label_size: LabelSize,
     group_width: Option<DefiniteLength>,
     auto_width: bool,
-    selected_index: usize,
+    selected_index: Option<usize>,
     tab_index: Option<isize>,
 }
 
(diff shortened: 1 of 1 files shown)


danilo-leal:
Hey @bnjjj, thanks for the PR! I'm curious, what is the scenario you're trying to fix where having an option selected by default isn't desirable?

bnjjj:
@danilo-leal in this pr https://github.com/zed-industries/zed/pull/42819 I have a button to filter by git remotes in branch picker

danilo-leal:
Gotcha. Given that is just a filter that selects one option, we could use a regular `Button` for that, similar to the many toggle/filter buttons in the project search, for example. It makes sense to me to preserve the toggle button behavior where there's always a selected option by default... therefore, will close the PR due to that. (Also, we might change that branch picker design, too—we've been discussing internally about unifying the remote, branch, and worktree pickers into one). Thank you!

PR #37605, opened 2025-09-05
Add new injections for Go

ksckaan1:
support for injecting sql, json, yaml, xml, html, css, js, lua and csv value

if you use `/* lang */` before string literals, highlights them

**Example:**

```go
const sqlQuery = /* sql */ "SELECT * FROM users;" // highlights as SQL code
```

Screenshot 2025-09-05 at 06 17 49

Closes #ISSUE

Release Notes:

- N/A *or* Added/Fixed/Improved ...

PR #43756, opened 2025-11-28
python: Add SQL syntax highlighting

jeffbrennan:
Release Notes: 

- Added support for SQL syntax highlighting in Python files

## Summary

I am a data engineer who spends a lot of time writing SQL in Python files using Zed. This PR adds support for SQL syntax highlighting with common libraries (like pyspark, polars, pandas) and string variables (prefixed with a `# sql` comment). I referenced [#37605](https://github.com/zed-industries/zed/pull/37605) for this implementation to keep the comment prefix consistent.

## Examples
image

### DIFF
--- a/crates/languages/src/python/injections.scm
+++ b/crates/languages/src/python/injections.scm
@@ -1,3 +1,43 @@
 ((comment) @injection.content
  (#set! injection.language "comment")
 )
+
+; SQL -----------------------------------------------------------------------------
+; common functions ex: spark.sql("SELECT * FROM tbl")
+(call
+    [
+        (attribute attribute: (identifier) @function_name)
+        (identifier) @function_name
+    ]
+    (#match? @function_name "(?i:sql|read_sql|read_sql_query)")
+    arguments: (argument_list
+        (string
+            (string_content) @injection.content (#set! injection.language "sql")
+        )
+    )
+)
+
+; sqlalchemy ex: from sqlalchemy import text; text("SELECT * FROM tbl")
+(call
+    function: (identifier) @function_name
+    (#match? @function_name "(?i:text)")
+    arguments: (argument_list
+        (string
+            (string_content) @injection.content (#set! injection.language "sql")
+        )
+    )
+)
+
+; string variables
+((comment) @comment
+    .
+    (expression_statement
+        (assignment
+            right: (string
+                (string_content) @injection.content
+            )
+        )
+    )
+    (#match? @comment "^(#|#\\s+)(?i:sql)\\s*$")
+    (#set! injection.language "sql")
+)
--- a/crates/languages/src/python/injections.scm
+++ b/crates/languages/src/python/injections.scm
@@ -9,7 +9,7 @@
         (attribute attribute: (identifier) @function_name)
         (identifier) @function_name
     ]
-    (#match? @function_name "(?i:sql|read_sql|read_sql_query)")
+    (#match? @function_name "(?i:sql|read_sql|read_sql_query|execute)")
     arguments: (argument_list
         (string
             (string_content) @injection.content (#set! injection.language "sql")
--- a/crates/languages/src/python/injections.scm
+++ b/crates/languages/src/python/injections.scm
@@ -9,7 +9,7 @@
         (attribute attribute: (identifier) @function_name)
         (identifier) @function_name
     ]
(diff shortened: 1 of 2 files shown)


crates/languages/src/python/injections.scm
```diff
@@ -1,3 +1,43 @@
...
 )
+
+; SQL -----------------------------------------------------------------------------
+; common functions ex: spark.sql("SELECT * FROM tbl")
+(call
+    [
+        (attribute attribute: (identifier) @function_name)
+        (identifier) @function_name
+    ]
+    (#match? @function_name "(?i:sql|read_sql|read_sql_query|execute)")
```
osiewicz:
```suggestion
    (#match? @function_name "^(?i:sql|read_sql|read_sql_query|execute)$")
```

crates/languages/src/python/injections.scm
```diff
@@ -1,3 +1,43 @@
...
+        (string
+            (string_content) @injection.content (#set! injection.language "sql")
+        )
+    )
+)
+
+; sqlalchemy ex: from sqlalchemy import text; text("SELECT * FROM tbl")
+(call
+    function: (identifier) @function_name
+    (#match? @function_name "(?i:text)")
```
osiewicz:
```suggestion
    (#match? @function_name "^(?i:text)$")
```

osiewicz (changes requested):
I'm not sold on this being the best way to achieve the desired result. In particular, I don't love how this uses names of functions from common libraries; while some (like `read_sql_query`) are pretty related to SQL, `execute` or `text` seem pretty generic - I worry about the false positives here.

At the same time, I don't have anything better at hand. Let me think about it for a bit and then we can prolly re-evaluate this PR. Feel free to chime in.

osiewicz (commented):
On a second thought, having a prefix like `# sql` or such seems like a generic enough solution that it may just work. However, we need to document it (we prolly should've done it before landing Go counterpart).

jeffbrennan:
> On a second thought, having a prefix like `# sql` or such seems like a generic enough solution that it may just work. However, we need to document it (we prolly should've done it before landing Go counterpart).

Thanks for the feedback! I agree that we should prioritize avoiding false positives. I am fine with removing the function name match patterns and making this an opt-in highlighting with the `# sql` comment.

image

osiewicz:
Yup, I love it. Admittedly putting `sql` comment is a bit verbose (formatting-wise), but hopefully that'd be fine.

jeffbrennan:
> Yup, I love it. Admittedly putting `sql` comment is a bit verbose (formatting-wise), but hopefully that'd be fine.

Sounds good! Will make those changes and try out some new test commands later today. Do you want the docs written in this same PR?

jeffbrennan:
@osiewicz, I have removed the function name parsing. Below are some updated test queries. Let me know if you have any questions!
image

crates/languages/src/python/injections.scm
```diff
@@ -1,3 +1,34 @@
...
+                (assignment
+                    right: (string
+                        (string_content) @injection.content
+                    )
+                )
+        ))
+    ]
+    (#match? @comment "^(#|#\\s+)(?i:sql)\\s*$")
+    (#set! injection.language "sql")
+)
```
osiewicz, line 34: Can we simplify this to include any comment directly preceding a string?
```suggestion

; SQL -----------------------------------------------------------------------------
(
  (comment) @comment
  .
  (string (string_content) @injection.content)
  (#match? @comment "^(#|#\\s+)(?i:sql)\\s*$")
  (#set! injection.language "sql")
)
```

jeffbrennan: Hi @osiewicz, I tried that out and it works for function arguments but it disables highlighting for string assignments. 

image

osiewicz:
> Do you want the docs written in this same PR?

Yep, that'd work I guess.

crates/languages/src/python/injections.scm
osiewicz: Ugh, that sucks. Let's keep it as is then.

jeffbrennan:
@osiewicz 
I have added documentation to the `docs/src/languages/python.md` file. Let me know if there's anything else you would like added to this PR! 

image

osiewicz:
Thank you!

osiewicz (approved):

Issue #35759, opened 2025-08-07
"Failed to load environment variables" error with Bash


jadenPete:
### Summary

I'm getting an error like this every time I start up Zed:
```
2025-08-06T20:40:34-04:00 WARN  [crates] login shell exited with exit status: 127. stdout: "", stderr: "bash: cannot set terminal process group (-1): Inappropriate ioctl for device\nbash: no job control in this shell\nstty: 'standard input': Inappropriate ioctl for device\nerror: Unable to read input file: EBADF: Bad file number\n"
2025-08-06T20:40:34-04:00 ERROR [project::environment] Failed to load environment variables. See log for details
```

### Description

I think it's because my login shell is Bash and Zed is executing a command like this:
```sh
bash -l -i -c 'cd <PROJECT>; /usr/bin/zeditor --printenv >&0'
```

I suspect the problem is that my `.bashrc` looks something like this:
```
# Setting various environment variables
...

# If not running interactively, don't do anything
[[ $- != *i* ]] && return

exec fish
```

I do this because I want to use Bash for all my non-interactive needs, but I prefer Fish as an interactive shell. My main question is this: why is Zed trying to spin up an interactive shell? Is that strictly necessary? I feel like Zed shouldn't make any assumptions about how an interactive shell should work.

### Zed Version and System Specs

Zed: v0.197.5 (Zed) 
OS: Linux Wayland arch unknown
Memory: 14.4 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 25.1.7-arch1.1

sshaikh:
Same issue for apparently the same reasons (bash + fish)

janilowski:
I get the same thing but don't even use fish :p

jdtzmn:
@jadenPete did you ever solve this issue?

jadenPete:
I did, but I still think a patch to Zed would be the most appropriate solution.

I ended up including something like this in my `.bashrc`:
```bash
# TODO: Remove the second condition when zed#35759 is fixed:
# https://github.com/zed-industries/zed/issues/35759
if [[ $- == *i* && -z "$BASH_EXECUTION_STRING" ]]; then
	exec fish
fi
```

Let me know if that doesn't work for you.

ian-h-chamberlain:
I've run into a somewhat related issue which is specific to WSL: https://github.com/zed-industries/zed/issues/40094#issuecomment-3559808526

[...]

PR #43764, opened 2025-11-28
direnv: Allow disabling env integration entirely

ian-h-chamberlain:
Relates to #35759, but maybe doesn't entirely fix it? I think it will improve the situation, at least.
Also provides a workaround for the issue described in https://github.com/zed-industries/zed/issues/40094#issuecomment-3559808526 for users of WSL + `nix-direnv`.

Rationale: there are cases where automatic direnv integration is not always desirable, but Zed currently has no way of opting out of this integration besides `direnv revoke` (which is often not desirable). 

This PR provides such an opt-out for users who run into problems with the existing direnv integration methods. Some reasons why disabling might be useful:
- Security concerns about auto-loading `.envrc` (arguably, `direnv revoke` should cover this most of the time)
- As in #35759, for users who use different shells/envs for interactive/non-interactive cases and want to manually control the environment Zed uses 
- As in #40094, to workaround OS limits on environment variable / command-line parameter size

Release Notes:

- Added the ability to disable direnv integration entirely

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1334,6 +1334,8 @@
   //      "load_direnv": "direct"
   // 2. Load direnv configuration through the shell hook, works for POSIX shells and fish.
   //      "load_direnv": "shell_hook"
+  // 3. Don't load direnv configuration at all.
+  //      "load_direnv": "disabled"
   "load_direnv": "direct",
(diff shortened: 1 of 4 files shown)


Veykril (approved):
Sounds reasonable to me, thanks!

Issue #38755, opened 2025-09-24
Use newer git diff machinery for Inline Assist diffs


JosephTLyons:
### Summary

<img width="1487" height="384" alt="Image" src="https://github.com/user-attachments/assets/efc28fb7-6014-45a2-9fbb-feb25fb08d2a" />

JosephTLyons:
This isn't a git issue really, but I marked it as such because we should be using the git diff UI here.

xipeng-jin:
Hi @JosephTLyons  
Yes, I can see that the current inline assistant doesn't use the git diff ui to render and it looks not good and characters are not even aligned. I can take some time working on this issue

<img width="1048" height="269" alt="Image" src="https://github.com/user-attachments/assets/3ba18f78-2f9c-43e8-8828-bbaf9382198d" />

xipeng-jin:
I created a PR for simply fixing the misalignment issue for the inline assistance. Regarding to this issue, I would like to have more details how we should use the new git diff to rendering the inline assist diff. Instead of rendering a separate deleted block, should we use the `BufferDiff` to render the git diff style?

PR #43768, opened 2025-11-29
Fix the misalignment issue with Inline Assist

xipeng-jin:
Easy path to fix the issue #38755

Screenshot 2025-11-28 at 9 39 55 PM

Release Notes:

- agent: Fixed the UI misalignment between old and new lines from the inline assist.


danilo-leal (approved):
Awesome, thank you!

PR #43790, opened 2025-11-29
Add project search modal

hosin211:
Closes #N/A

  This PR adds an optional modal-based global search as an alternative to the tab-based project search.

  **Why:** The modal provides a faster, more streamlined search experience - it opens instantly as an overlay without creating a new tab, similar to VS
  Code's Cmd+P style. This is useful for quick searches where you don't need the search results to persist.

  **Features:**
  - Search with options (case, word, regex, include ignored)
  - File include/exclude filters
  - Replace next / replace all
  - Save prompt for dirty buffers (respects autosave settings)

  Controlled by `search.use_modal` setting (defaults to `false`).

  ## [Screencast](https://drive.google.com/file/d/1F6vhKTd25nH-KjiEHYYgfATtqsSVXwHt/view?usp=sharing)

  Release Notes:

  - Added modal-based project search with replace functionality, controlled by new `search.use_modal` setting


hosin211:
@cla-bot check

hosin211:
@cla-bot check

ChristopherBiscardi:
I do like the overall idea here, but it would be better to support both approaches (tab-based and modal based) rather than requiring a setting to only use one or the other.

GitGlimpse895:
Hey @danilo-leal — you've been the assigned reviewer here since November with no activity. @mattermill quietly unassigned himself in February without explanation.

There's a design question that's blocking community consensus here: the current implementation uses `search.use_modal` as an either/or toggle, but both @poeck and @ChristopherBiscardi have raised that keeping both modes available simultaneously (e.g. separate keybindings) would be preferable. @hosin211 can't really know whether to redesign the approach without a signal from the team.

Could you weigh in on the either/or vs. coexistence question? That alone would unblock the path forward here.

lem3s:
Hope this gets merged soon

joshuafouch:
please.

yara-blue:
Thanks for building on Zed! we are getting close to landing a dedicated quick search modal - with optional previews - see #56037. Since it makes no senses to have two search modals we will not be going with this approach.

Once that is out (soon :tm: ) let us know if there is any use case it does not address for you and we can see how to address it.

PR #43780, opened 2025-11-29
Expose IME-related vkeys on Windows (ime_toggle, convert, nonconvert, kana)

tyaro:
Adds minimal Windows IME key mappings so applications can detect IME toggle keys reported by ImmGetVirtualKey (observed private vkeys 0xF0..0xF4). Also maps VK_CONVERT, VK_NONCONVERT, and VK_KANA to friendly names.\n\nRelease Notes:\n\n- Added IME-related virtual-key mappings on Windows: ime_toggle (private vkeys 0xF0–0xF4), convert (VK_CONVERT), 
onconvert (VK_NONCONVERT), and kana (VK_KANA). These mappings let applications detect IME mode toggles more reliably.\n\nIncludes tests, a small example, and docs under crates/gpui. See https://github.com/tyaro/gpui-ime-windows-ime/pull/1 for a smaller, isolated view of the change.

PR #43794, opened 2025-11-29
gpui: Remove temporary Windows clippy workaround

tyaro:
This PR removes the temporary Windows-specific workaround added to script/clippy.ps1 that treated the missing  No spectre-mitigated libs were found error as non-fatal.\n\nBackground: a temporary tolerance was added to make progress on PRs while our Windows CI runners were missing necessary Visual Studio components (spectre-mitigated libs). With the related feature changes merged and CI passing, this cleanup reverts the script so clippy failures again surface normally.\n\nIf CI still fails on Windows due to missing spectre-mitigated libs after this change, maintainers may need to update the runner Visual Studio installation to include those components. (See prior PR: zed-industries/zed#43780)

### DIFF
--- a/crates/gpui/docs/windows-ime.md
+++ b/crates/gpui/docs/windows-ime.md
@@ -1,0 +1,31 @@
+Windows / 日本語キーボード / IME の取り扱い
+=========================================
+
+概要
+----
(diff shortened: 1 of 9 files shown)


Veykril:
Closing as this seems to be doing multiple unrelated things as well as reverting IME handling that we don't want to revert

PR #43801, opened 2025-11-29
project_panel: Improve wording around file deletion

AdamJedl:
Make it clear in the UI that "Delete" of file or folder is permanent action.
For example in windows explorer and VS Code "Delete" means move to trash.

Or maybe also remove permanent delete from the context menu completely and allow it only through keyboard shortcut, like it's in Windows Explorer, VS Code and KDE Dolphin file manager.

Release Notes:

- Improved wording within file deletion prompts in the projetct panel.


AdamJedl:
@cla-bot check

MrSubidubi (approved):
Changed it slightly, thanks for getting the ball rolling here!

PR #43846, opened 2025-11-30
journal: Added possibility to define path, and filename for the journal file

gyger:
Uses the chronos crate, and file path is formated using DateTime::format.

 - If this is interesting I can add the tests.

Closes Discussion #43845

Release Notes:

- Added date_path configuration for journal files.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1399,6 +1399,8 @@
   "journal": {
     // The path of the directory where journal entries are stored
     "path": "~",
+    // What name to use for folders and journal files.
+    "date_format": "%Y/%m/%d.md",
     // What format to display the hours in
     // May take 2 values:
     // 1. hour12
--- a/crates/journal/src/journal.rs
+++ b/crates/journal/src/journal.rs
@@ -1,4 +1,4 @@
-use chrono::{Datelike, Local, NaiveTime, Timelike};
+use chrono::{Local, NaiveTime, Timelike};
 use editor::scroll::Autoscroll;
 use editor::{Editor, SelectionEffects};
 use gpui::{App, AppContext as _, Context, Window, actions};
@@ -26,6 +26,10 @@ pub struct JournalSettings {
     ///
     /// Default: `~`
     pub path: String,
+    /// Date format for journal entries.
+    ///
+    /// Default: `%Y/%m/%d.md`
+    pub date_format: String,
     /// What format to display the hours in.
     ///
     /// Default: hour12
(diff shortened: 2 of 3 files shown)


gyger:
@cla-bot check

crates/journal/src/journal.rs
```diff
@@ -66,10 +71,14 @@ pub fn new_journal_entry(workspace: &Workspace, window: &mut Window, cx: &mut Ap
     let journal_dir_clone = journal_dir.clone();
 
     let now = Local::now();
-    let month_dir = journal_dir
-        .join(format!("{:02}", now.year()))
-        .join(format!("{:02}", now.month()));
-    let entry_path = month_dir.join(format!("{:02}.md", now.day()));
+    let entry_path = journal_dir.join(now.format(&settings.date_format).to_string());
+
+    if entry_path.file_name().is_none() {
```
SomeoneToIgnore, line 76: This is wrong, directories usually have a file name: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=76bbb3483b8f50dd57a3cc76cda79c52

The wording is also odd: date format to produce a file name?
Why does it even need to do so, and why cannot Zed append this suffix instead?

assets/settings/default.json
```diff
@@ -1399,6 +1399,8 @@
   "journal": {
     // The path of the directory where journal entries are stored
     "path": "~",
+    // What name to use for folders and journal files.
+    "date_format": "%Y/%m/%d.md",
```
SomeoneToIgnore, line 1403: We have a configuring-zed.md file where this also needs to be added, so that zed.dev/docs gets something too.

Also, the date format: it would be nice to mention what are the options for the people.
`format` library method points at `strftime.rs` docs in `chrono` crate, https://docs.rs/chrono/latest/chrono/format/strftime/index.html 

Seems worth adding that?

SomeoneToIgnore (commented):
Thank you, adding tests with `FakeFs` or some tmp dir sounds not a bad idea here.

crates/journal/src/journal.rs
gyger: One of the goals for me was to generate not only markdown but also other file extemsions. In my case I have journals in .rst files.

The date format was because I looked at the parameters setting of other solutions (Obsidian) and it had this name. I can find something else.

SomeoneToIgnore: Yeah, I think we can control extension separately from the data to avoid odd interpretations?

SomeoneToIgnore:
Will close as stale, as 2 weeks had passed with no action.

I think the change makes sense overall though after those small oddities mentioned above are fixed.

PR #40139, opened 2025-10-13
gpui: Add support for touch events on Wayland

robert7k:
This PR works towards #13698 and implements touch events on Wayland. I'm aware of #34675, but that PR has probably been rebased incorrectly, so I couldn't find the source code or get it to work.

Currently, zed can't be used on a touch screen on Wayland: the touch events are simply ignored, nothing happens when finger-clicking anywhere in zed. This PR makes zed usable on a touch-screen.

I have tested this on Linux with Wayland.

The PR contains multiple features:

- process touch evens to allow clicking on UI elements
- allows scrolling using a two-finger gesture
- pinch to zoom gesture

Release Notes:

- Added support for touch events on Wayland

Demo video (the red and blue circles show the touch events. Unfortunately, not all touch events are properly visible, but everything in the video has been done through a touch screen):

https://github.com/user-attachments/assets/eed3f04c-1c14-41a2-9b89-c1cee2ed82f5

PR #34675, opened 2025-07-17
gpui: Add minimal touch input implementation on Wayland 

orual:
Progress on #13698 (Touch input support)

Provides a minimum usable implementation of touch input in gpui and Zed on Wayland, via mouse pointer event emulation.

- 2+ finger tap + drag scrolls
- 1 finger tap is a left click
- 1 finger long press (4x the DOUBLE_CLICK_DURATION constant) is a right click
- 1 finger drag acts as a click and drag (text selection, etc.)
- This is quite hacky and lacks configurability, but it works well enough.
- Scrolling requires both fingers to touch the screen at the same time, which makes it a little finicky. Some of the logic behind this could be improved.

The implementation adds two fields to the WaylandClientState struct, a couple of small utility structs for managing touch event state, and an implementation of the Dispatch trait to handle Wayland touch events. Much of the event code is copied with minor modifications from the mouse pointer Dispatch implementation.

Before merging, this should be put behind a feature flag in gpui or otherwise disabled by default, due to it being a bit of a hack, so as to avoid unexpected behaviour changes from a WIP implementation and make use of it, as a stopgap until more complete touch support is added, opt-in. The exact shape that will take should be discussed during the review process, and I can make the required changes to gate it.

Release Notes:

- Added support for translating touch events to mouse events on Wayland

PR #43848, opened 2025-11-30
gpui: Translate Wayland touch events to mouse events

ecstatic-morse:
First step towards #13698
Alternative to #40139 and #34675

#40139 didn't work on my machine due to the way it handled the [`Window` for touch events](https://github.com/zed-industries/zed/pull/40139#discussion_r2573244164).  It also made a few decisions that I don't agree with as someone who often uses a touchscreen when reading code. Namely:
- Restricting scrolling to two-finger touch (most editors allow scrolling with one-finger, suppressing clicks when a drag is detected).
- Implementing pinch-to-zoom, which isn't good UX when you can't translate the viewport to keep the content under the users fingers consistent.

Obviously, it would be ideal to add `on_touch_down` and friends, maybe emitting mouse up/down/click events when those methods aren't called, but that would require a lot of design work. Unless someone enfranchised takes it upon themselves to start that process, Zed will remain unresponsive on touch screens without hacks like this PR.

The goal is to be simple enough to get upstreamed quickly while providing a good foundation for proper touch/gesture handling. It:
- emits `Mouse{Up,Down}` events when the screen is tapped.  (just like #40139)
- emits `ScrollWheel` events when a finger is dragged across the screen.
- tracks multiple touches, but only emits scroll events for the earliest touch.
- allows simultaneous tapping and scrolling.
- hides this translation behind a feature flag (enabled by the `zed` crate).
- doesn't have pinch-to-zoom. 
- doesn't have long press to right click.

Single finger scroll conflicts with things like drag and drop and moving the minimap, but provides the best UX for the most common use-case for touch gestures: browsing code.

Currently , this PR stores a weak version of `WaylandWindowStatePtr` for each active touch, so touches don't keep a window alive and are silently dropped if it goes out of scope. This is probably overkill. I don't really know the lifecycle of a Wayland window in Zed, or if a single Zed instance uses multiple Wayland windows at once. However, it seemed like the most bulletproof solution.

Release Notes:

- Initial support for touch input on Wayland (tap-to-left-click and and drag-to-scroll)
s

### DIFF
--- a/crates/gpui/src/platform/linux/wayland/client.rs
+++ b/crates/gpui/src/platform/linux/wayland/client.rs
@@ -1,6 +1,7 @@
 use std::{
     cell::{RefCell, RefMut},
     hash::Hash,
+    mem,
     os::fd::{AsRawFd, BorrowedFd},
     path::PathBuf,
     rc::{Rc, Weak},
@@ -33,7 +34,7 @@ use wayland_client::{
     Connection, Dispatch, Proxy, QueueHandle, delegate_noop,
     protocol::{
         wl_buffer, wl_compositor, wl_keyboard, wl_pointer, wl_registry, wl_seat, wl_shm,
-        wl_shm_pool, wl_surface,
+        wl_shm_pool, wl_surface, wl_touch,
     },
 };
 use wayland_protocols::wp::cursor_shape::v1::client::{
@@ -68,7 +69,7 @@ use xkbcommon::xkb::{self, KEYMAP_COMPILE_NO_FLAGS, Keycode};
 
 use super::{
     display::WaylandDisplay,
-    window::{ImeInput, WaylandWindowStatePtr},
+    window::{ImeInput, WaylandWindowStatePtr, WaylandWindowStateWeakPtr},
 };
 
 use crate::{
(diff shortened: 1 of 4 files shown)


ecstatic-morse:
Taps inside the file explorer pane don't open the tapped file; they just highlight it. Tapping anything else (code, terminal, menus, other panes, etc.) works fine. I tried adding a delay between the generated `MouseUp` and `MouseDown` events to no effect.

crates/gpui/src/platform/linux/wayland/client.rs
```diff
@@ -245,6 +263,163 @@ pub(crate) struct WaylandClientState {
...
+    #[must_use]
+    #[allow(clippy::mutable_key_type)]
+    fn push_event(
+        &mut self,
+        event: wl_touch::Event,
+        windows: &HashMap<ObjectId, WaylandWindowStatePtr>,
+    ) -> Vec<TouchAction> {
+        let mut ret = Vec::new();
+        match event {
+            // Handle eagerly (don't wait for `wl_touch::Event::Frame`)
```
ecstatic-morse, line 319: I originally waited for a `Frame` to handle everything, but I wasn't getting `Frame` events after `Up` events, so taps never triggered. Seems like a bug in how wl_touch events are delivered. A better workaround is to batch all events and set up a timeout when we have received a touch event but have not yet received a `Frame`, but that adds a lot of complexity.

mikayla-maki:
Hi,

meritozh:
@ecstatic-morse please consider submitting this pr to [gpui-ce/gpui-ce](https://github.com/gpui-ce/gpui-ce).

PR #43853, opened 2025-12-01
language: Fix auto-indentation for Python code blocks in Markdown

### LINKED ISSUE #43722: Auto-indent for Python does not work within injected language scopes
### Reproduction steps

1. Open a markdown file
2. Start a Python section (`|`) is location of cursor:
    ```md
    ```python
    class Example():|

    ```

3. Pressing enter, you get the following:
    ```md
    ```python
    class Example():
    |
    ```
4. However, you would expect the following:
    ```md
    ```python
    class Example():
        |
    ```

I also ran into this problem with the Python scopes within my Zed plugin for [collagraph](https://github.com/fork-tongue/collagraph).

### If applicable, attach your relevant Zed settings to this issue

_No response_

### If applicable, provide details about your model provider

_No response_

Zed: v0.214.6 (Zed) 
OS: macOS 15.7.1
Memory: 16 GiB
Architecture: aarch64

Arch Linux

jeffbrennan:
Closes #43722

Release Notes:

- Fixed an issue where auto-indentation didn’t work correctly for Python code blocks in Markdown.

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -3241,22 +3241,23 @@ impl BufferSnapshot {
             Point::new(prev_non_blank_row.unwrap_or(row_range.start), 0)
                 ..Point::new(row_range.end, 0),
             |row, line| {
+                // Either the parent language, or an injected language
                 let row_language = self
                     .language_scope_at(Point::new(row, 0))
                     .map(|scope| Arc::clone(scope.language()));
-                let row_config = row_language
+                let row_language_config = row_language
                     .as_ref()
                     .map(|lang| lang.config())
                     .unwrap_or(config);
 
-                if row_config
+                if row_language_config
                     .decrease_indent_pattern
                     .as_ref()
                     .is_some_and(|regex| regex.is_match(line))
                 {
                     indent_change_rows.push((row, Ordering::Less));
                 }
-                if row_config
+                if row_language_config
                     .increase_indent_pattern
                     .as_ref()
                     .is_some_and(|regex| regex.is_match(line))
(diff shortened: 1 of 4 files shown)


jeffbrennan:
Unable to reproduce the test failure on my machine 
image

crates/language/src/buffer.rs
```diff
@@ -3241,14 +3241,23 @@ impl BufferSnapshot {
             Point::new(prev_non_blank_row.unwrap_or(row_range.start), 0)
                 ..Point::new(row_range.end, 0),
             |row, line| {
-                if config
+                // Either the parent language, or an injected language
+                let row_language = self
+                    .language_scope_at(Point::new(row, 0))
```
smitbarmase: I think we can just use `language_at` here instead of `language_scope_at`. The latter is more sophisticated and not really needed in this case.

crates/language/src/language.rs
```diff
@@ -1334,6 +1334,7 @@ struct OverrideEntry {
 struct InjectionPatternConfig {
     language: Option<Box<str>>,
     combined: bool,
+    indent: bool,
```
smitbarmase: Why are we adding this field here and tracking it all the way through the parsing step? I don’t see any code that reads this value or uses it in any meaningful way.

smitbarmase (changes requested):
Thanks for working on this. I added few nits and claryfications. Also, I think this is more complicated than it looks, since when calcualting these indents of embedded language we should be careful about that we are not using captures, positions of base language while calculating it. i.e. `start_positions` is used to calculate `last_seen_suffix` and then which affects `regex_outdent_map`. And more such cases where we should respect these boundaries.

Thanks for working on this. I added a few nits and clarifications.

I also think this is more complicated than it looks. When calculating indents for an embedded language, we need to be careful not to use captures or positions from the base language. For example, `start_positions` is used to calculate `last_seen_suffix`, which then affects `regex_outdent_map`, so just using config from embedded language miight not be sufficient. There can be few more cases like this where we need to make sure we’re respecting those boundaries.

smitbarmase (approved):
I went ahead and pushed this forward, it’s looking really solid right now. Thanks for all the work you’ve put in. This is a great improvement!

jeffbrennan:
Thank you! I appreciate the feedback, sorry I didn't get a chance to make the suggested changes

Issue #40970, opened 2025-10-23
JSONC trailing comma warnings in bun.lock


jchitel:
### Summary

bun.lock reports warnings for trailing commas that cannot be ignored, even though the file is controlled by Bun.

### Description

This will happen in any project with a bun.lock file produced by a `bun install` command.

bun.lock is correctly registered as JSONC, but it doesn't really make sense to report diagnostics since it is managed by Bun. Since trailing commas seem to be allowed in JSONC files, it is debatable whether these warnings should be present in any JSONC file, but I'm not aware of the rationale.

Steps to reproduce:
1. Open Zed in a Bun project with a bun.lock file.

**Expected Behavior**:

No diagnostics reported for bun.lock.

**Actual Behavior**:

The diagnostics panel indicates warnings for trailing commas in bun.lock.

### Zed Version and System Specs

Zed: v0.208.6 (Zed) 
OS: macOS 15.7
Memory: 16 GiB
Architecture: aarch64

SomeoneToIgnore:
https://github.com/zed-industries/zed/discussions/23651 is the same report it seems.

https://github.com/zed-industries/zed/blob/ed5b9a47053d18180d66b44cbc6753f9db0507a8/crates/languages/src/jsonc/config.toml#L2-L3

marks bun.lock to be jsonc language-based, apparently we need to alter what the corresponding language server can return as diagnostics?

jchitel:
Yea, that would resolve this one if it ends up being decided that the warnings are unnecessary in any JSONC file, not just bun.lock.

If not, some mechanism will have to exempt bun.lock specifically for the reasons I specified above.

unikitty37:
I am having the same problem in Waybar's `config.jsonc` — when I save, Zed _automatically adds the trailing commas_, but then sticks warnings on all of them.

Trailing commas _are_ valid in JSONC, so the warning shouldn't be displayed for this file format. There's no need to make special exemptions for bun.lock.

acgxv:
It’s not a major issue but the inconsistent behaviour can be confusing.

The default formatter adds commas to JSONC files and the default JSONC language server reports a warning about trailing commas.

ian-h-chamberlain:
I figured out a way to silence this warning: by telling `json-language-server` [what schema to use](https://zed.dev/docs/languages/json#schema-specification-via-settings), it will stop emitting a trailing comma warning, because the upstream schema has `"allowTrailingCommas": true`:

```jsonc
{
  "lsp": {
[...]

Issue #21303, opened 2024-11-28
jsonc formatter adds trailing commas and triggers warnings


matthewmueller:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

Create a jsonc file like, e.g. `config.jsonc`

```jsonc
// This is a dot schema file. It defines the server configuration for your project.
// For more information, visit https://dot.io/docs/schema
{
  // Application name
  "name": "", // e.g., "my-app"
  // Domains for the app (leave empty for local development)
  // Example: ["example.com", "api.example.com"]
  "domains": [],
  // Host servers to deploy to
  "hosts": {
    "main": {
      "host": "localhost:2222",
      "admin": "ubuntu",
      "user": "dot"
    }
  },
}
```

Then hit save, the formatter will append trailing commas, leading to warnings.

### Environment

Zed: v0.162.5 (Zed)
OS: macOS 14.6.1
Memory: 64 GiB
Architecture: aarch64

### If applicable, add mockups / screenshots to help explain present your vision of the feature

Quick video:

https://github.com/user-attachments/assets/b49eb435-00f3-43c0-a8bf-62b33c52d90c

I'd expect the formatter to not generate trailing commas. This also matches what VSCode does.

### If applicable, attach your Zed.log file to this issue.

<details><summary>Zed.log</summary>

```

```
</details>

notpeter:
This is an upstream prettier issue:
- https://github.com/prettier/prettier/issues/15956

Duplicate of:
- https://github.com/zed-industries/zed/issues/17487

Explicitly called out in the [Zed JSON Language Docs](https://zed.dev/docs/languages/json#jsonc-prettier-formatting).

Issue #18509, opened 2024-09-29
insert comma when creating newline in JSON


kurtbuilds:
### Check for existing issues

- [X] Completed

### Describe the feature

In IntelliJ, the editor understands your code edit actions and the local context. As one example, using vim mode, press "o" (for insert new line) on the last line of a JSON object will insert the newline AND also insert a comma preceding that line.

It makes it much faster and easier to edit JSON, and it'd be huge benefit for Zed to do this as well.

kurtbuilds:
Another issue is that Zed currently inserts `null` as the value when inserting new keys, when I would expect it to create a nested object.

Example when editing `settings.json`:

<img width="322" alt="image" src="https://github.com/user-attachments/assets/13651075-f83c-458e-b8e7-9a19843e9b1e">

<img width="320" alt="image" src="https://github.com/user-attachments/assets/3a897d14-e4d9-4784-a091-f8f6825e5d04">

Instead of an object, Zed creates a `null` value, which I then have to select, delete, create braces, and hit newline, to create the new object.

This might be a separate issue, as it depends on Zed's understanding of the jsonschema of the current document.

notpeter:
Yes, it would be nice if Zed would automatically introduce a comma when appending an additional key to the end of an object.  Interestingly we already support removing an excess trailing comma automatically (prettier will remove it if `format_on_save` is on) and it always felt asymmetrical that Zed would remove an extra comma on save, but never add one back.

Anyways JetBrains "Smart Keys" does this by default, it'd be nifty if Zed could too. 
<img width="898" alt="Screenshot 2024-10-13 at 10 55 19" src="https://github.com/user-attachments/assets/e2e5758c-4104-4f82-a78b-eabe9fe43e81">

nesl247:
This is still needed.

ian-h-chamberlain:
It's not a complete fix (won't work for plain JSON), but for use with JSONC that supports trailing commas, and #43854 now merged, it's possible to automatically add trailing commas, which mitigates this issue. Configuration example:

```jsonc
{
  "prettier": {
    "overrides": [
      {
        "files": ["**/Zed/*.json", "**/zed/*.json", "*.jsonc"],
        "options": {
          "parser": "jsonc",
        },
      },
    ],
  },
}
```

With this, the builtin `prettier` integration will add trailing commas, making it a lot more ergonomic to add new items to these files.

Issue #17487, opened 2024-09-06
JSON5/JSONC Self-Corruption


ytcalifax:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

Upon saving a valid JSON5 file, the double-quoting seems to be automatically removed for some strange reason (and funnily enough, after that the language server detects the JSONC as broken). I have also tested this on a Linux VM (as I know Windows build is probably not officially supported) and obviously, as the issue is caused by Zed itself, the quoting still gets removed.

### Environment

Zed: v0.153.0 (Zed Dev 497356b2ba37128cb1af89395a4dcc12b6421bc3)
OS: Windows 10.0.22631
Memory: 31.9 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3060 || NVIDIA || 556.12

### If applicable, attach your Zed.log file to this issue.

<details><summary>Zed.log</summary><pre>

</pre></details>

CharlesChen0823:
add this to disable default `prettier` formatter, 
```
  "languages": {
    "JSONC": {
      "formatter": "language_server",
      "prettier": {
        "allowed": false
      }
    }
  }
```

notpeter:
I can reproduce, but I believe this is an upstream prettier issue:
- https://github.com/prettier/prettier/issues/15956

By default Prettier will add trailing commas to any file with a *.jsonc extension.  There are two workarounds available: 

1. Name your JSONC files *.json (this is what Zed does in our codebase)
2. Add a .prettierrc:
```json
{
  "overrides": [
    {
      "files": ["*.jsonc"],
      "options": {
        "parser": "json",
        "trailingComma": "none"
      }
    }
  ]
}
```

Personally I think this is brain-dead behavior, but although Zed could fix this by default (e.g. pass extra options to prettier when handling JSONC files) it would mean Zed would auto-format code differently from prettier in other contexts (CI, VSCode, git hook, manually invoked, etc). So for the moment we're kind of stuck with it.

For the moment I'm going to add a note to the docs with a workaround and reference the upstream issue.

PR #43854, opened 2025-12-01
Allow trailing commas in builtin JSONC schemas

ian-h-chamberlain:
The JSON language server looks for a top-level `allowTrailingCommas` flag to decide whether it should warn for trailing commas. Since the JSONC parser for these builtin files can handles trailing commas, adding this flag to the schema also prevents a warning for those commas.

I don't think there's an issue that is only for this specific issue, but it relates to *many* existing / older issues:
- #18509
- #17487
- #40970
- #18509
- #21303

Release Notes:

- Suppress warning for trailing commas in builtin JSON files (`settings.json`, `keymap.json`, etc.)


JosephTLyons:
This looks great, thank you!

PR #39928, opened 2025-10-10
gpui: Do not render ligatures between different styled text runs

Veykril:
This relands https://github.com/zed-industries/zed/pull/37175 as https://github.com/zed-industries/zed/pull/39886 fixed the jiggling issue.

Currently when we render text with differing styles adjacently we might form a ligature between the text, causing the ligature forming characters to take on one of the two styles. This can especially become confusing when a ligature is formed between actual text and inlay hints.

Annoyingly, the only ways to prevent this with core text is to either render each run separately, or to insert a zero-width non-joiner to force core text to break the ligatures apart, as it otherwise will merge subsequent font runs of the same fonts.

We currently do layouting on a per line basis and it is unlikely we want to change that as it would incur a lot of complexity and annoyances to merge things back into a line, so this goes with the other approach of inserting ZWNJ characters instead.

Note that neither linux nor windows seem to currently render ligatures, so this only concerns macOS rendering at the moment.

Closes https://github.com/zed-industries/zed/issues/23194

Release Notes:

- Fixed ligatures forming between real text and inlay hints on macOS

PR #43856, opened 2025-12-01
gpui: Further fix extraction of font runs from text runs

### LINKED ISSUE #42297: Theme: Font weight ignored when string.escape has same color as string
### Summary

In my theme, I have a `font-weight` setting for `string.escape`. The setting is not being respected. the font used, Cascadia Code, does have that weight. This is a regression from past versions of Zed.

### Description
My buffer font is Cascadia Code. Font weight is 350. This is from settings.json

```json
	"buffer_font_size": 13,
	"buffer_font_family": "Cascadia Code",
	"buffer_font_weight": 350,
```

The following json is in my theme under `"syntax"`.

```json
"string": {
	"color": "#5D81A8"

},
"string.escape": {
	"color": "#5D81A8",
	"font_weight": 800
},
```

This is what it looks like in my editor 

Image

Changing the `"color"` property *does* affect string escapes. For example changing `"color"` to `#FF0000` gives me the following.

Image

**Expected Behavior**:

Escapes be bold. This used to work in previous versions of Zed.

**Actual Behavior**:

Escapes not bold. 

Zed: v0.211.6 (Zed) 
OS: macOS 26.0.1
Memory: 8 GiB
Architecture: aarch64

_No response_

cyndis:
After #39928, if a font's weight changes between text runs without other decoration changing, the earlier weight continues to be used for the subsequent run(s).

PR #40840 fixes this in shape_text, but similar code exists also in layout_text. The latter is used for text in the editor view itself, so the issue continues to appear when using a highlighting theme with varied font weights.

Fix the issue by applying the same fix in layout_text.

Closes #42297

Release Notes:

- Fixed incorrect font weights in editor view when using a highlighting theme with varying font weights

### DIFF
--- a/crates/gpui/src/text_system.rs
+++ b/crates/gpui/src/text_system.rs
@@ -550,7 +550,6 @@ impl WindowTextSystem {
         force_width: Option<Pixels>,
     ) -> Arc<LineLayout> {
         let mut last_run = None::<&TextRun>;
-        let mut last_font: Option<FontId> = None;
         let mut font_runs = self.font_runs_pool.lock().pop().unwrap_or_default();
         font_runs.clear();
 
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Nice catch, thanks!

PR #43879, opened 2025-12-01
ui: Refactor dashed divider with path builder

madcodelife:
Replace the original div implementation with path builder.

| Before | After |
| - | - |
| before1before2 | after1after2 |

Release Notes:

- N/A


danilo-leal (approved):
Nice, thank you so much!

PR #43888, opened 2025-12-01
language: Only block the foreground on buffer reparsing when necessary

Veykril:
Gist is we only need to block the foreground thread for reparsing if immediate language changes are useful to the user. That is usually only the case when they edit the buffer

Release Notes:

- Improved performance of large project searches and project diffs

Co-authored by: David Kleingeld <david@zed.dev>

### DIFF
--- a/crates/acp_thread/src/diff.rs
+++ b/crates/acp_thread/src/diff.rs
@@ -35,7 +35,9 @@ impl Diff {
                     .await
                     .log_err();
 
-                buffer.update(cx, |buffer, cx| buffer.set_language(language.clone(), cx))?;
+                buffer.update(cx, |buffer, cx| {
+                    buffer.set_language_immediate(language.clone(), cx)
+                })?;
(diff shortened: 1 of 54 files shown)


yara-blue (approved):

maxbrunsfeld:
The methods that are used in many places base became `set_language_immediate` and `with_language_immediate`, and the one with only a few call sites kept the intuitive name `set_language`.

Naming-wise, I think it's a bit odd to allow the name to be dominated by this term "immediate" that is just about a very subtle optimization. It'd be nice to have the frequently-used method keep the regular name `set_language`.

Could the buffer manage the `may_block` boolean automatically internally, by recording whether its syntax-highlighted text has ever been requested? That seems like it captures the intent behind the change, looking at the call sites. Alternatively, should we set the `may_block` state with some more explicit stateful API, where the editor tells the buffer that it is currently focused?

PR #30386, opened 2025-05-09
Fix command palette sorting to maintain frequency during filtering

stepankuzmin:
Follow up after https://github.com/zed-industries/zed/pull/2954. Not sure if this is the best way to do it, but it seems to be working fine.

Release Notes:

- Fixed command palette sorting to maintain frequency during filtering

PR #43899, opened 2025-12-01
Add fuzzy score boosting for usage frequency

stepankuzmin:
Adds a boost field to StringMatchCandidate that multiplies fuzzy match scores. Command palette uses this to rank frequently-used commands higher in search results.

Uses square root scaling (1 use → 2.5x, 10 uses → 5.74x, 50 uses → 11.6x) so frequent commands get advantage without completely overriding fuzzy match quality. Zero-score matches stay filtered out regardless of boost.

Includes tests verifying boost works with empty queries, affects ranking, and doesn't promote non-matches.

Alternative approach to https://github.com/zed-industries/zed/pull/30386

### DIFF
--- a/crates/command_palette/src/command_palette.rs
+++ b/crates/command_palette/src/command_palette.rs
@@ -1,7 +1,7 @@
 mod persistence;
 
 use std::{
-    cmp::{self, Reverse},
+    cmp,
     collections::HashMap,
     sync::Arc,
     time::Duration,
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore:
Draft, a bunch of tests failing, 

image

a conflict

image

and no activity for around 2 weeks, so I'll close this as stale.

PR #43944, opened 2025-12-01
Support zooming and panning in the image viewer

### LINKED ISSUE #9584: Image Viewer: Pan & Zoom
### Check for existing issues

- [X] Completed

### Describe the feature

I generally expect to be able to pan around images by dragging with the mouse (which one of the mouse buttons varies a bit). Likewise, I expect to be able to zoom in/out by scrolling.

On a touchpad panning could also happen by scrolling and zooming by pinching.

_No response_

MostlyKIGuess:
Implemented Pan and Zoom on the Image Viewer.

Demo:

https://github.com/user-attachments/assets/855bafe8-fdc2-4945-9bfb-e48382264806

Closes #9584

Release Notes:

- Add zoom in/out, reset, fit-to-view and zoom-to-actual actions 
- Support scroll-wheel zoom with modifier, click-and-drag panning 
- Renders a zoom percentage overlay. 
- ImageView Toolbar Added to control zoom in/out and fit view.


MrSubidubi (changes requested):
Thanks for this! I'll give this a more proper review soonish (got some other things on my list first), but just to give you a few TODOs I just noticed: 

- Can we add buttons for the new actions to the toolbar, optimally the right corner? If we do not have icons for some of these, we usually source ours from https://lucide.dev/icons/ (Please only do if you actually have to, we should have quite a few already).
- Whatever the checkered background is doing in your example is.. not optimal. Could you ensure that the checkered background is actually just drawn behind the image (so one can see the transparent parts of the image) and that the squares properly follow the scaling of the image?


Thanks!

MostlyKIGuess:
> Thanks for this! I'll give this a more proper review soonish (got some other things on my list first), but just to give you a few TODOs I just noticed:
> 
>     * Can we add buttons for the new actions to the toolbar, optimally the right corner? If we do not have icons for some of these, we usually source ours from https://lucide.dev/icons/ (Please only do if you actually have to, we should have quite a few already).
> 
>     * Whatever the checkered background is doing in your example is.. not optimal. Could you ensure that the checkered background is actually just drawn behind the image (so one can see the transparent parts of the image) and that the squares properly follow the scaling of the image?
> 
> 
> Thanks!

Fixed the checkered background error and this toolbar: 
Made it so that on hover it shows the shortcut ( I really love this part about Zed so I had to add it ).

image

MostlyKIGuess:
weird that the terminal test failed on mac, I am looking into it, but can we re-run the test ?

Edit: Yeah the re-run made it pass

MostlyKIGuess:
The last commit vastly improved the performance, realized  we didn't even need to draw it :D

MostlyKIGuess:
@MrSubidubi Could you review this?

MrSubidubi (changes requested):
Thanks for the follow up! A few notes:

- We do need the checkered background for partially transparent images, e.g. `zed/crates/zed/resources/app-icon-preview@2x.png`. We do have the outline, but I'd definitely prefer the checkered background, as removing it would feel like a regression.
- Dragging does not account for the current zoom factor and can thus get out of sync. Can add a video of this if you'd like, but this basically means that the image does not move as many pixels as I drag on certain zoom levels. IMO this should be 1:1 mapping without any drift.
- The zoom level in the bottom half of the screen can get out of the view. I think we should probably just remove it, because we have the appropiate zoom level now present in the toolbar as well.

That should be it for this round!

Also, quick note, no need to ping me here. I'll get back to you as soon as I can, but this can (sadly) sometimes take some time. Thanks!

MostlyKIGuess:
> Thanks for the follow up! A few notes:
> 
>     * We do need the checkered background for partially transparent images, e.g. `zed/crates/zed/resources/app-icon-preview@2x.png`. We do have the outline, but I'd definitely prefer the checkered background, as removing it would feel like a regression.
> 
>     * Dragging does not account for the current zoom factor and can thus get out of sync. Can add a video of this if you'd like, but this basically means that the image does not move as many pixels as I drag on certain zoom levels. IMO this should be 1:1 mapping without any drift.
> 
>     * The zoom level in the bottom half of the screen can get out of the view. I think we should probably just remove it, because we have the appropiate zoom level now present in the toolbar as well.
> 
> 
> That should be it for this round!
> 
> Also, quick note, no need to ping me here. I'll get back to you as soon as I can, but this can (sadly) sometimes take some time. Thanks!

I don't know what you mean exactly by the 2nd point, video would be helpful yes!

crates/image_viewer/src/image_viewer.rs
```diff
@@ -303,81 +474,191 @@ impl Focusable for ImageView {
...
+                                            cx.notify();
+                                        }
+                                    });
+                                }
+                            }
+                        });
+                    },
+                )
+                .absolute()
+                .size_0()
```
MrSubidubi: I think that listener can just live within the window without the canvas around it, and we only need to attach it if we are currently actually dragging the image, right?

crates/image_viewer/src/image_viewer.rs
```diff
@@ -7,30 +7,47 @@ use anyhow::Context as _;
...
+const MAX_ZOOM: f32 = 20.0;
+const ZOOM_STEP: f32 = 1.1;
+
 pub struct ImageView {
     image_item: Entity<ImageItem>,
     project: Entity<Project>,
     focus_handle: FocusHandle,
+    zoom_level: f32,
+    pan_offset: Point<Pixels>,
+    is_dragging: bool,
```
MrSubidubi: I think we should make `is_dragging`  a method like

```rust
fn is_dragging(&self) -> bool {
    self.last_mouse_position.is_some()
}
```
, right? We set both simultaneously, so perhaps better to just keep one of these.

crates/image_viewer/src/image_viewer.rs
```diff
@@ -7,30 +7,47 @@ use anyhow::Context as _;
...
-    item::{BreadcrumbText, Item, ProjectItem, SerializableItem, TabContentParams},
+    item::{BreadcrumbText, Item, ItemHandle, ProjectItem, SerializableItem, TabContentParams},
 };
 
 pub use crate::image_info::*;
 pub use crate::image_viewer_settings::*;
 
+actions!(
+    image_viewer,
+    [ZoomIn, ZoomOut, ResetZoom, FitToView, ZoomToActualSize]
```
MrSubidubi: Could we have some generic comments for these actions?

crates/image_viewer/src/image_viewer.rs
```diff
@@ -50,10 +67,21 @@ impl ImageView {
...
+            .read(cx)
+            .image_metadata
+            .map(|m| (m.width, m.height));
+
         Self {
             image_item,
             project,
             focus_handle: cx.focus_handle(),
+            zoom_level: 1.0,
+            pan_offset: point(px(0.0), px(0.0)),
```
MrSubidubi:
```suggestion
            pan_offset: Point::default(),
```

Nit, I think that can be done in a few more places and is a bit easier to read IMO.

crates/image_viewer/src/image_viewer.rs
```diff
@@ -67,12 +95,149 @@ impl ImageView {
...
+
+    fn reset_zoom(&mut self, _: &ResetZoom, _window: &mut Window, cx: &mut Context<Self>) {
+        self.zoom_level = 1.0;
+        self.pan_offset = point(px(0.0), px(0.0));
+        cx.notify();
+    }
+
+    fn fit_to_view(&mut self, _: &FitToView, _window: &mut Window, cx: &mut Context<Self>) {
+        if let (Some(bounds), Some((img_width, img_height))) =
+            (self.container_bounds, self.image_size)
```
MrSubidubi:
```suggestion
        if let Some(bounds, (img_width, img_height)) =
            self.container_bounds.zip(self.image_size)
```

crates/image_viewer/src/image_viewer.rs
```diff
@@ -67,12 +95,149 @@ impl ImageView {
...
+                    center.x - bounds.origin.x - bounds.size.width / 2.0,
+                    center.y - bounds.origin.y - bounds.size.height / 2.0,
+                );
+                let zoom_ratio = self.zoom_level / old_zoom;
+                self.pan_offset = point(
+                    (self.pan_offset.x + relative_center.x) * zoom_ratio - relative_center.x,
+                    (self.pan_offset.y + relative_center.y) * zoom_ratio - relative_center.y,
+                );
+            }
+        }
```
MrSubidubi: We can either if let - chain or zip here to remove one layer of indentation

crates/image_viewer/src/image_viewer.rs
```diff
@@ -303,81 +474,191 @@ impl Focusable for ImageView {
...
+
+        let border_color = cx.theme().colors().border;
+
+        let (mut left, mut top) = (px(0.0), px(0.0));
+        let mut is_centered_by_flex = true;
+
+        if let (Some(bounds), Some((scaled_w, scaled_h))) = (self.container_bounds, scaled_size) {
+            is_centered_by_flex = false;
+
+            // Convert Pixels to f32 for calculation
```
MrSubidubi: Are you sure we have to do this? All ops should be implemented for Pixels, so I don't think this should be needed here.

crates/image_viewer/src/image_viewer.rs
```diff
@@ -303,81 +474,191 @@ impl Focusable for ImageView {
...
+                    })
+                    .child(
+                        div()
+                            .relative()
+                            .when(is_centered_by_flex, |div| {
+                                div.flex().justify_center().items_center().size_full()
+                            })
+                            .when(!is_centered_by_flex, |div| {
+                                div.absolute().left(left).top(top)
+                            })
```
MrSubidubi: Nit: You could use `when_else` here

crates/image_viewer/src/image_viewer.rs
```diff
@@ -303,81 +474,191 @@ impl Focusable for ImageView {
...
+                                canvas(
+                                    |_, _, _| {},
+                                    move |bounds, _, window, _cx| {
+                                        let bounds_x: f32 = bounds.origin.x.into();
+                                        let bounds_y: f32 = bounds.origin.y.into();
+                                        let bounds_width: f32 = bounds.size.width.into();
+                                        let bounds_height: f32 = bounds.size.height.into();
+
+                                        // muchh larger squares
+                                        let square_size = 48.0 * zoom_level;
```
MrSubidubi:
```suggestion
                                        const BASE_SQUARE_SIZE: f32 = 48.0;
                                        let square_size = BASE_SQUARE_SIZE * zoom_level;
```

crates/image_viewer/src/image_viewer.rs
```diff
@@ -303,81 +474,191 @@ impl Focusable for ImageView {
...
+
+                                                    window.paint_quad(gpui::fill(
+                                                        rect,
+                                                        opaque_grey(0.6, 1.0),
+                                                    ));
+                                                }
+                                            }
+                                        }
+
+                                        // draw border
```
MrSubidubi:
```suggestion
```

crates/image_viewer/src/image_viewer.rs
```diff
@@ -411,6 +692,122 @@ impl ProjectItem for ImageView {
...
+        if let Some(item) = active_pane_item {
+            if let Some(image_view) = item.downcast::<ImageView>() {
+                self._subscription = Some(cx.observe(&image_view, |_, _, cx| {
+                    cx.notify();
+                }));
+                self.image_view = Some(image_view.downgrade());
+                cx.notify();
+                return ToolbarItemLocation::PrimaryRight;
+            }
+        }
```
MrSubidubi: Can also if let chain here.

crates/image_viewer/src/image_viewer.rs
```diff
@@ -67,12 +95,149 @@ impl ImageView {
...
+    fn handle_scroll_wheel(
+        &mut self,
+        event: &ScrollWheelEvent,
+        _window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        if event.modifiers.control || event.modifiers.platform {
+            let delta: f32 = match event.delta {
+                ScrollDelta::Pixels(pixels) => pixels.y.into(),
+                ScrollDelta::Lines(lines) => lines.y * 20.0,
```
MrSubidubi: Let's lift the `20.0` here in this method into another constant pleaes

crates/image_viewer/src/image_viewer.rs
```diff
@@ -67,12 +95,149 @@ impl ImageView {
...
+            let zoom_factor = if delta > 0.0 {
+                1.0 + delta.abs() * 0.01
+            } else {
+                1.0 / (1.0 + delta.abs() * 0.01)
+            };
+            self.set_zoom(self.zoom_level * zoom_factor, Some(event.position), cx);
+        } else {
+            let delta = match event.delta {
+                ScrollDelta::Pixels(pixels) => pixels,
+                ScrollDelta::Lines(lines) => point(px(lines.x * 20.0), px(lines.y * 20.0)),
```
MrSubidubi:
```suggestion
                ScrollDelta::Lines(lines) => lines.map(|delta| px(delta * 20.0))
```

crates/image_viewer/src/image_viewer.rs
```diff
@@ -67,12 +95,149 @@ impl ImageView {
...
+            };
+            self.pan_offset = point(self.pan_offset.x + delta.x, self.pan_offset.y + delta.y);
```
MrSubidubi:
```suggestion
            self.pan_offset += delta;
```

crates/image_viewer/src/image_viewer.rs
```diff
@@ -67,12 +95,149 @@ impl ImageView {
...
+    fn handle_mouse_move(
+        &mut self,
+        event: &MouseMoveEvent,
+        _window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        if self.is_dragging {
+            if let Some(last_pos) = self.last_mouse_position {
+                let delta = point(event.position.x - last_pos.x, event.position.y - last_pos.y);
+                self.pan_offset = point(self.pan_offset.x + delta.x, self.pan_offset.y + delta.y);
```
MrSubidubi:
```suggestion
                let delta = event.position - last_pos;
                self.pan_offset += delta;
```

MrSubidubi (changes requested):
Nice, this is so much better already, thank you very much! (And sorry for the delay from my side, was too occupied elsewhere).

I left quite a few comments below, but we should already be close to merging.

Most notably though, we still have two mayor issues as shown in the video here: 

https://github.com/user-attachments/assets/0f555d04-d3f4-4dcc-bb59-de7dea76ef76

As you see,
- there is a brief flicker at the start. This happens I believe because we flex layout once, and I think we should drop the special case for this here. This might be a bit tricky, but we could add a custom element for the image to draw it (would introduce a bit of boilerplate), but I'd like us to get rid of this. If you need more input for this, please hit me up and I'll get into more detail here/look at this myself.
- zooming with the mouse wheel does not take the current positon of the mouse into account and that feels counterintuitive. I think when we resize there, the resize-operation should take the cursor position into account.

Other than that, looks so much better, thank you already! As for profiling, we are currently doing and experimenting a lot with tracy, but there is no official guideline. So far performance has been fine for me, although I fully agree checkered background support would be nice to have here. Feel free to throw a TODO somewhere, but we can also discuss more here.

crates/image_viewer/src/image_viewer.rs
MostlyKIGuess: Yes!

crates/image_viewer/src/image_viewer.rs
MostlyKIGuess: Added very basic explanations

crates/image_viewer/src/image_viewer.rs
MostlyKIGuess: For this, I actually consulted an AI after going through some stackoverflow posts, the one which helped was [link](https://stackoverflow.com/questions/38446666/scaling-around-a-specific-point-in-2d-coordinate-system).

This enables us to have intuitive scrolling.

crates/image_viewer/src/image_viewer.rs
MostlyKIGuess: Yeah, I didn't realize this earlier. I refactored to handle in pixels

crates/image_viewer/src/image_viewer.rs
MostlyKIGuess: absolute now which removes the flickering

MostlyKIGuess:
Thank you for the detailed review! 

> Other than that, looks so much better, thank you already! As for profiling, we are currently doing and experimenting a lot with tracy, but there is no official guideline. So far performance has been fine for me, although I fully agree checkered background support would be nice to have here. Feel free to throw a TODO somewhere, but we can also discuss more here.

Yeah, this is acceptable performance compared to earlier.
Maybe this goes into GPUI instead of it being here. So the follows ups would be:
1. To have background support from GPUI. 
2. I also wanted to add that I have a [PR](https://github.com/gpui-ce/gpui-ce/pull/11) in the gpui-ce for touchpad support in linux and mac.  Perhaps that can be added later for zooming in through the touchpad.

MrSubidubi (changes requested):
Okay, one mast mid-sized thing and I think we are good, thank you for all your efforts.

I think we need to refactor this a bit so that we have the container bounds of the current frame and not the bounds of the previous frame. Using the information from the previous frame, we
- always do not render an image on the first frame, which causes a brief but noticeable flicker
- always have a small flicker whenever we open other panels/resize because again, sizing calculation is done with sizes from the previous frame.

I think we need to change stuff of this to utilize a small wrapper instead that impls `Element`. We can then request the layout as we currently do and get the corresponding bounds during prepaint, which we can then utilize and which corresponds to the actual current bounds as opposed to the bounds from the previous frame. Then we should basically be able to use the current `render` of the image viewer in that prepaint, request the layout slightly delayed there and can prepaint and paint after as needed.

I think https://github.com/zed-industries/zed/blob/81ea0403d1b9e94f0b3d1b7e2bb4857a0adfc060/crates/ui/src/utils/with_rem_size.rs#L45 might be a good skeleton for this in combination with `StickyItems` (obviously we'd need this to be slightly different), but this would resolve that issue. What do you think?

MostlyKIGuess:
> I think that is precisely what I wanted to describe and I think we cannot get rid of this unless going for an `Element` impl - I am happy to stand corrected here, but I think that we will not get around this if we want to update/paint the image on the same frame.

Yeah, this can be added as a todo on GPUI perhaps where the render is parallel for the background and the image. To describe it precisely an ability to set an image background/pattern.

MostlyKIGuess:
> Went ahead and applied the last set of fixes myself - feel free to check these out, we should now always properly layout the image according to the current viewport, no matter what. If there is anything, please let me know, otherwise, I think we are good to go!

Thank you! I tested it locally and works as expected. 

I can add gesture support along with adding the following changes in GPUI: [Gesture Support GPUI-CE PR](https://github.com/gpui-ce/gpui-ce/pull/11) as continuation.

MrSubidubi (approved):
Thanks for taking another look!

I think the pinch support sounds interesting, feel free to open a PR for that with the support for the image viewer here and I'll gladly take a look!


Lastly, thank you so so much for bearing with me here and for your perseverance, I for one think (and hope you agree) it was worthwhile and am very happy to merge this now. Thank you so much!

MostlyKIGuess:
> Lastly, thank you so so much for bearing with me here and for your perseverance, I for one think (and hope you agree) it was worthwhile and am very happy to merge this now. Thank you so much!

Thanks a ton for the thorough review and patience, really appreciate it. Learned a lot from this one. 

Looking forward to the pinch-zoom follow-up: https://github.com/zed-industries/zed/pull/47351

MrSubidubi:
Hey @MostlyKIGuess ! Since this landed, we now have checkered backgruound support in our shaders! 🎉  Can I interest you in looking into that? 

Happy to look at a PR, I'll probably do it myself at some point but currently don't have the time to do it immediately :/

PR #43949, opened 2025-12-01
Add customization capabilities to which-key menu

xipeng-jin:
Closes #10910

Support parsing extra fields "binding_labels" and "group_labels" under each context stack in _keymap.json_

Release Notes:

- Added customization capability to which-key menu to enabling renaming action name and prefix of the group of actions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19728,6 +19728,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "which_key"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "command_palette",
+ "gpui",
+ "schemars 1.0.1",
+ "serde",
+ "settings",
+ "theme",
+ "ui",
+ "workspace",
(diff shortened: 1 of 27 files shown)


ConradIrwin:
@xipeng-jin Thanks for the idea!

I don't think we should have the abstraction leak such that the whole code-base needs to know about the which-key feature; instead if we want to do customization we should put it in that crate.

That said, I don't think we need to do this at all. Zed already has the concept of "namespaces" for actions (vim::Blah, editor::Blah), and introducing another grouping that is orthogonal to that seems a bit confusing. (even `:help g` in Vim gives you "commands starting with 'g'").

xipeng-jin:
Hi @ConradIrwin , thanks for reminding me this. I truly agree we should not leak the abstraction layer and all features of which-key system should be put inside that crate. I also think this type of higher level customization might not be the priority to do right now. Lets leave it for the future reference. Thanks again.

PR #43965, opened 2025-12-02
Limit edit prediction reject batches to max

agu-z:
We currently attempt to flush all rejected predictions at once even if we have accumulated more than `MAX_EDIT_PREDICTION_REJECTIONS_PER_REQUEST`. Instead, we will now flush as many as possible, and then keep the rest for the next batch. 

Release Notes:

- N/A


agu-z:
I'll actually let this roll into preview naturally

PR #43966, opened 2025-12-02
vim: Fix gj block

CnsMaple:
The method to reproduce this is as follows: 
Open a Python file, then enable inlay-hint and reduce the width until a character is pushed to the next line. Then try gj, and it will be prevented from moving to the next line.

This is my test code:
```python
####################


def func1(a, b) -> int:
    return a + b


def func2(a, b):
    return a + b


if __name__ == "__main__":
    func2(1, 2)
    aa = func1(1, 4)
    print(aa)
```

Before and after repair comparison

https://github.com/user-attachments/assets/5fa051c7-b081-413c-a2d7-8fe6e0cda823

Release Notes:

- Fix gj block

### DIFF
--- a/crates/project/src/lsp_command.rs
+++ b/crates/project/src/lsp_command.rs
@@ -2938,11 +2938,12 @@ impl InlayHints {
 
         let position = buffer_handle.read_with(cx, |buffer, _| {
             let position = buffer.clip_point_utf16(point_from_lsp(lsp_hint.position), Bias::Left);
-            if kind == Some(InlayHintKind::Parameter) {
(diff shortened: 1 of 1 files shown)


ConradIrwin:
Thanks for taking a pass at this, I don't think this is the right fix – though I'm not sure what is.

In other similar code-paths in vim we check if moving down and rounding failed and do something different, maybe we could copy that approach?

CnsMaple:
I've considered other solutions, but those might require significant changes to zed.

I don't have a better idea for this problem either, I'm very sorry.

PR #43968, opened 2025-12-02
Add Tailwind CSS support for Gleam

arjunbajaj:
Currently, Zed does not provide suggestions and validations for Gleam, as it is only available for languages specified in `tailwind.rs`. This pull-request adds Gleam to that list of languages.

After this, if Tailwind is configured to work with Gleam, suggestions and validation appear correctly.

Even after this change, Tailwind will not be able to detect and give suggestions in Gleam directly. Below is the config required for Tailwind classes to be detected in all Gleam strings.

<details><summary>Zed Config for Tailwind detection in Gleam</summary>
<p>

```
{
  "languages": {
    "Gleam": {
      "language_servers": [
        "gleam",
        "tailwindcss-language-server"
      ]
    }
  },
  "lsp": {
    "tailwindcss-language-server": {
      "settings": {
        "experimental": {
          "classRegex": [
            "\"([^\"]*)\""
          ]
        }
      }
    }
  }
}
```

The `classRegex` will match all Gleam strings, making it work seamlessly with Lustre templates and plain string literals.

</p>
</details> 

Release Notes:

- Added support for Tailwind suggestions and validations for the [Gleam programming language](https://gleam.run/).

### DIFF
--- a/crates/languages/src/tailwind.rs
+++ b/crates/languages/src/tailwind.rs
@@ -176,6 +176,7 @@ impl LspAdapter for TailwindLspAdapter {
         HashMap::from_iter([
             (LanguageName::new("Astro"), "astro".to_string()),
             (LanguageName::new("HTML"), "html".to_string()),
+            (LanguageName::new("Gleam"), "html".to_string()),
             (LanguageName::new("CSS"), "css".to_string()),
             (LanguageName::new("JavaScript"), "javascript".to_string()),
             (LanguageName::new("TypeScript"), "typescript".to_string()),
--- a/docs/src/languages/tailwindcss.md
+++ b/docs/src/languages/tailwindcss.md
@@ -28,6 +28,7 @@ Languages which can be used with Tailwind CSS in Zed:
 - [Astro](./astro.md)
 - [CSS](./css.md)
 - [ERB](./ruby.md)
+- [Gleam](./gleam.md)
 - [HEEx](./elixir.md#heex)
 - [HTML](./html.md)
 - [TypeScript](./typescript.md)


arjunbajaj:
@cla-bot check

yeskunall:
Hey @arjunbajaj I’m not very familiar with Gleam, but just wondering: is it a good thing to have Tailwind enabled for Gleam by default? In other words, it makes sense to have Tailwind enabled for CSS, does the same apply for Gleam (`.gleam`)?

Also, what does a Lustre template look like? Is it a standard HTML file with support for Gleam syntax?

arjunbajaj:
Hey @yeskunall,

I just tested it, this PR does not start the Tailwind LSP until the `languages.Gleam` config explicitly states to start the Tailwind LSP, so it should be safe.

It may not make sense in general to enable the Tailwind LSP for Gleam, but Gleam does have a popular use-case in the Lustre framework, similar to how Phoenix LiveViews are now quite a popular use-case for Elixir.

Without this PR, and just adding the config, we get hover functionality from the Tailwind LSP but not the suggestions and validation warnings that it gives in other languages.

Lustre now auto-detects Tailwind config in their build tool. After this PR merges, I'll discuss with the Gleam community if we could integrate a more seamless approach where we wouldn't even require the extra config in Zed if Gleam/Lustre auto-detects Tailwind.

crates/languages/src/tailwind.rs
```diff
@@ -176,6 +176,7 @@ impl LspAdapter for TailwindLspAdapter {
         HashMap::from_iter([
             (LanguageName::new("Astro"), "astro".to_string()),
             (LanguageName::new("HTML"), "html".to_string()),
+            (LanguageName::new("Gleam"), "html".to_string()),
```
yeskunall, line 179: Does a Lustre template just look like a standard HTML file with support for Gleam syntax?

arjunbajaj:
Lustre does not have any special template syntax like JSX or HTML. It simply uses Gleam functions.

Here is an example:
```gleam
fn view() {
  html.div([class("bg-black text-white")], [
    html.p([class("text-sm font-serif")], [html.text("hello")]),
  ])
}
```

However, since it is simply Gleam code, we may want to extract the Tailwind class string for better formatting and readability:

```gleam
fn view() {
  let styles = "bg-black text-white
  <some-other-tailwind-styles>...
  "

  html.div([class(styles)], [
    html.p([class("text-sm font-serif")], [html.text("hello")]),
  ])
}
```

So we would need Tailwind suggestions and highlighting on all strings if possible.

While Lustre's syntax does not look like HTML, Tailwind does work properly as far as I can tell, on the syntax.

If in the future we need some changes, perhaps Tailwind could add special support for Gleam like it has for Svelte and Vue, and then we can change it here.

crates/languages/src/tailwind.rs
yeskunall, line 179: Based on your comment [here](https://github.com/zed-industries/zed/pull/43968#issuecomment-3600700042), shouldn’t this be:

```diff
- (LanguageName::new("Gleam"), "html".to_string()),
+ (LanguageName::new("Gleam"), "gleam".to_string()),

**Review by yeskunall — COMMENTED** [review]
One last question, then I think we’re good to go.

File: crates/languages/src/tailwind.rs (hunk 1 above)
Comment (arjunbajaj) [reply crates/languages/src/tailwind.rs:179]: I believe (though I am not fully sure) that the value of the hashmap is the language name that comes from this file in the Tailwind LSP: https://github.com/tailwindlabs/tailwindcss-intellisense/blob/main/packages/tailwindcss-language-service/src/util/languages.ts

Since Gleam isn't mentioned there yet, we have two options:

1. Keep it as HTML here in Zed (at least for now)
2. I'll open a pull request to that repo to add Gleam to the list of HTML languages first.

Let me know what would you prefer.

Comment (yeskunall) [reply crates/languages/src/tailwind.rs:179]: Ideally, it should work like an existing solution (unless it is unfavorable by the community). Does VS Code handle this already? If so, we mimic them. If not, I think we should first make a PR to include Gleam as a special language: https://github.com/tailwindlabs/tailwindcss-intellisense/blob/main/packages/tailwindcss-language-service/src/util/languages.ts#L64

If you go down route 2, please mark this PR as a draft till it’s ready for review once again.

Comment (arjunbajaj) [reply crates/languages/src/tailwind.rs:179]: I just tested it in VS Code, and it works correctly.

A fresh install of VSC, installed the Gleam and Tailwind plugins, and then added this config:

```json
{
    "tailwindCSS.includeLanguages": {
        "gleam": "html"
    },
    "tailwindCSS.experimental.classRegex": [
        "\"([^\"]*)\""
    ]
}
```

I get hover, suggestions, and validations correctly.

As I have understood it, VS Code, Helix, and other editors automatically defer to Tailwind's config for the LSP and pass on requests for suggestions. It seems Zed does not do that for all languages, it only does it for languages listed in the `tailwind.rs` file, hence why we need the change.

Also Zed's own docs and issues show to set Ruby, Elixir, and many other languages to HTML in the Tailwind config.

Hence I think setting gleam to html for tailwind detection is the right approach.

yeskunall: Thanks for your well-written explanations.

arjunbajaj: Thanks, and thank you for merging the PR too. 🚀

PR #43988, opened 2025-12-02
lsp: Send client process ID in LSP initialize request

aviatesk:
Per the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams), the client should send its process ID to the language server so the server can monitor the client process and shut itself down if the client exits unexpectedly.

Release Notes:

- N/A


SomeoneToIgnore (approved):
Great find, thank you.

PR #44001, opened 2025-12-02
Update PHP language server docs

findrakecil:
Reformat document structure like other language docs, improve information flow, add missing requirements, and fix typos.

Release Notes:

- N/A

### DIFF
--- a/docs/src/languages/php.md
+++ b/docs/src/languages/php.md
@@ -29,21 +29,7 @@ which php
 
 ## Choosing a language server
 
-The PHP extension offers Phpactor and Intelephense as language server with Phpactor as default.
-
-### Phpactor
-
-To use Phpactor instead of Intelephense, add the following to your `settings.json`:
-
-```json [settings]
-{
-  "languages": {
-    "PHP": {
-      "language_servers": ["phpactor", "!intelephense", "..."]
-    }
-  }
-}
-```
+The PHP extension uses [LSP language servers](https://microsoft.github.io/language-server-protocol) with Phpactor as the default. If you want to use other language servers that support Zed (e.g. Intelephense or PHP Tools), make sure to follow the documentation on how to implement it.
 
 ### Intelephense
 
@@ -61,7 +47,7 @@ To use Intelephense, add the following to your `settings.json`:
 }
 ```
 
-To use the premium features, you can place your license file inside your home directory at `~/intelephense/licence.txt`.
+To use the premium features, you can place your license file inside your home directory at `~/intelephense/licence.txt` for macOS and Linux, or `%USERPROFILE%/intelephense/licence.txt` on Windows.
 
 Alternatively, you can pass the licence key or a path to a file containing the licence key as an initialization option. To do this, add the following to your `settings.json`:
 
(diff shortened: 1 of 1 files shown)


osiewicz (commented):
Should we also mention PHPTools support?

findrakecil:
> Should we also mention PHPTools support?

I think the documentation is focused on what products supported by default rather than listing what it supports. Since we have to install PHP Tools via extension, it makes sense if Zed excluding it from the documentation.

docs/src/languages/php.md
```diff
@@ -2,34 +2,54 @@
...
-  - [intelephense](https://github.com/bmewburn/vscode-intelephense/)
+- Tree-sitter: [tree-sitter/tree-sitter-php](https://github.com/tree-sitter/tree-sitter-php)
+- Language Server: [phpactor/phpactor](https://github.com/phpactor/phpactor)
+- Alternate Language Server: [bmewburn/vscode-intelephense](https://github.com/bmewburn/vscode-intelephense/)
 
-## Choosing a language server
+## Install PHP
 
-The PHP extension offers both `phpactor` and `intelephense` language server support.
+The PHP extension requires PHP to be installed and available in your path:
```
osiewicz: It would be good to mention Windows install path.

osiewicz:
```suggestion
The PHP extension requires PHP to be installed and available in your PATH:
```
too

docs/src/languages/php.md
```diff
@@ -2,34 +2,54 @@
...
 which php
 ```
 
+## Choosing a language server
+
+The PHP extension offers Phpactor and Intelephense as language server with Phpactor as default.
+
+### Phpactor
+
+To use Phpactor instead of Intelephense, add the following to your `settings.json`:
```
osiewicz: Nit: if Phpactor is the default, then the following wording is a bit confusing.
```suggestion
To use Phpactor instead of Intelephense, add the following to your `settings.json`:
```
Why would I use Phpactor instead of Intelephense, if Phpactor is the default? As in, this wording makes sense for a non-default value.

docs/src/languages/php.md
```diff
@@ -2,34 +2,54 @@
...
+}
+```
+
 ### Intelephense
 
-[Intelephense](https://intelephense.com/) is a [proprietary](https://github.com/bmewburn/vscode-intelephense/blob/master/LICENSE.txt#L29) language server for PHP operating under a freemium model. Certain features require purchase of a [premium license](https://intelephense.com/).
+[Intelephense](https://intelephense.com/) is a [proprietary](https://github.com/bmewburn/vscode-intelephense/blob/master/LICENSE.txt#L29) language server for PHP operating under a freemium model. Certain features require purchase of a [premium license](https://intelephense.com/buy).
 
-To switch to `intelephense`, add the following to your `settings.json`:
+To use Intelephense, add the following to your `settings.json`:
```
osiewicz: This would be a good place to use the wording from https://github.com/zed-industries/zed/pull/44001/files#r2584841567

docs/src/languages/php.md
```diff
@@ -61,9 +83,9 @@ Zed supports syntax highlighting for PHPDoc comments.
 
 - Tree-sitter: [claytonrcarter/tree-sitter-phpdoc](https://github.com/claytonrcarter/tree-sitter-phpdoc)
 
-## Setting up Xdebug
+## Debugging
 
-Zed’s PHP extension provides a debug adapter for PHP and Xdebug. The adapter name is `Xdebug`. Here a couple ways you can use it:
+The PHP extension provides a debug adapter for PHP and Xdebug. Here a couple ways you can use it:
```
osiewicz:
```suggestion
The PHP extension provides a debug adapter for PHP via Xdebug. There are several ways to use it:
```

docs/src/languages/php.md
```diff
@@ -83,10 +105,10 @@ Zed’s PHP extension provides a debug adapter for PHP and Xdebug. The adapter n
 ]
 ```
 
-In case you run into issues:
+In case you run into issues, here are some ways to solve them:
```
osiewicz:
```suggestion
These are common troubleshooting tips, in case you run into issues:
```

osiewicz:
@findrakecil fair, but how am I supposed to know (as an user) that Phptools are supported? It's not mentioned in either docs for the language support OR the extensions README.md.

docs/src/languages/php.md
findrakecil: You're correct. My intention was if someone was using something other than Phpactor and want to switch back to Phpactor.

I should've changed the sentence to better reflect my intention.

findrakecil:
@osiewicz 

> @findrakecil fair, but how am I supposed to know (as an user) that Phptools are supported? It's not mentioned in either docs for the language support OR the extensions README.md.

I believe we should change the sentence describing the content of **Choosing a langguage server**. I will push a new change and let's see whether it would satisfy the requirement so new user can expect what they get and what to do if they didn't.

EDIT: just checked the [DEVSENSE blog about Zed support](https://blog.devsense.com/2025/zed-ide-php/). It's officially support Zed now. I'll add that to the docs.

osiewicz:
Thank you!

PR #44014, opened 2025-12-02
diffs: AST based syntax diffing

### LINKED ISSUE #9721: Support Difftastic for visualizing code diffing
### Check for existing issues

- [X] Completed

### Describe the feature

Difftastic is a structural diff tool that understands syntax. It supports [over 30 programming languages](https://difftastic.wilfred.me.uk/languages_supported.html). It would be nice to integrate it with Zed.

Its result is way better & cleaner than traditional diff
image

_No response_

versecafe:
Closes #9721

Release Notes:

- Added syntax based diffing for git & agent diffs

### Built in language git diff
image

### Built in language agent diff

image

### Extension based language diff

image


maxbrunsfeld:
Hey @versecafe , thanks for the PR! A couple of initial questions -

1. Would it be possible to re-use `imara_diff` instead of `wu_diff` for sequence diffing that that you're doing here? It would be nice to not introduce another diffing library into the application when we already have one.
2. Can we avoid re-parsing the file on every diff? That's very inefficient, since we already maintain a parse tree of the buffer and the diff base for syntax highlighting purposes
3. What is the `Syntax` type? It seems like a whole other representation of a syntax tree in addition to the tree-sitter one. Is the entire tree converted into a `Syntax` before diffing? That seems expensive.

maxbrunsfeld:
I'm going to close this out for now, because there are some major changes that we would want to see. But @versecafe this is still a feature that we're open to, if we can make it slot in more minimally. Feel free to open a new PR if you get a chance to make progress on these questions.

PR #44033, opened 2025-12-02
cli: Enable `--add` option for development builds on macOS

aviatesk:
When running Zed from a development build (via `--zed ./target/release/zed`), the `--add` CLI option didn't work because the CLI currently would always spawn a new process instead of connecting to an existing instance.

This change extends the Unix socket IPC mechanism (already used on Linux) to macOS, allowing the CLI to connect to a running development build of Zed.

Written with Claude.

Closes #ISSUE

Release Notes:

- Enabled `--add` CLI option for development builds on macOS

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1152,30 +1152,44 @@ mod mac_os {
                 }
 
                 Self::LocalPath { executable, .. } => {
-                    let executable_parent = executable
-                        .parent()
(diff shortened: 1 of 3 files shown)


aviatesk:
Friendly bump.
This is a simple change to improve the usability of development builds on macOS.
I think this change can be reviewed better if you ignore whitespace:
https://github.com/zed-industries/zed/pull/44033/files?diff=unified&w=1

Issue #9461, opened 2024-03-17
Support more syntax tokens for theme configuration


zarifpour:
### Check for existing issues

- [X] Completed

### Describe the feature

**Proposed Solution:**
1. **Expand Theme Configuration**: Extend the theme's configuration file to include more syntax tokens, enabling granular customization for different code elements.

2. **Language-specific Support**: Allow themes to define colors for syntax tokens specific to different programming languages.

This approach will enhance user experience by allowing more detailed and language-specific theme customization.

maxdeviant:
@zarifpour Do you have specific examples of things that we don't currently support?

zarifpour:
I can look for a better example, but here's a quick one. Let me know if it makes sense:

![CleanShot 2024-03-18 at 10 50 12 on Zed — GitHub](https://github.com/zed-industries/zed/assets/16494335/da2b8a8d-01be-4cf2-b397-51b550c33cc4)

![CleanShot 2024-03-18 at 10 51 41 on Zed — AuctionBot sol — GitHub](https://github.com/zed-industries/zed/assets/16494335/06510645-e7aa-4b98-9afa-290f037c0b2a)

I don't think it's possible to color these specific elements, where in VSCode I can target these specifically and configure their colors and font style.

pimmee:
Here's an [example issue](https://github.com/d1y/vscode_dark_plus.zed/issues/1#issue-2151866798) I created in the VSCode Dark Plus theme extension repo, who looks completely different in Zed vs VSCode.

When I tried to fix the colors and make a PR, I also quickly ran into issues with too few specific syntax tokens, especially for Typescript. Some examples include `return` or `await`. When I looked into how VSCode does it I recall seeing they have [language specific scopes](https://github.com/microsoft/vscode/blob/7f40fe42427877ab8f336d4aad90aae30f185aa8/extensions/theme-defaults/themes/dark_plus.json#L60).

<img width="581" alt="image" src="https://github.com/zed-industries/zed/assets/33460567/598b93e9-e9b6-4067-bf91-d1541142e187">

Would love if the color theming can be improved!

codeitlikemiley:
I guess this is the same as my issue
https://github.com/zed-industries/zed/issues/13961

I really need to make rust syntax highlighting more noticeable...
its hard to work with all types having same color

everdrone:
Also Rust attributes have inconsistent highlighting
zed:
[...]

PR #44043, opened 2025-12-03
Make highlighting of C preprocessing directive same as C++

lipcut:
Small fix for consistency between C and C++ highlighting. Related to https://github.com/zed-industries/zed/issues/9461

Release Notes:

- Change syntax highlighting for preprocessing directive in C so it can be configured with `keyword.directive` instead of being treated as other `keyword`. The behavior should be like the C++ one now.

圖片

### DIFF
--- a/crates/languages/src/c/highlights.scm
+++ b/crates/languages/src/c/highlights.scm
@@ -36,7 +36,7 @@
   "#ifndef"
   "#include"
   (preproc_directive)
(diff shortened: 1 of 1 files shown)


lipcut:
@cla-bot check

MrSubidubi (approved):
Looks good, thanks! And congratulations to your first contribution! 🎉

PR #44048, opened 2025-12-03
workspace: Display a launchpad page when in an empty window & add it as a `restore_on_startup` value

simonpham:
Hi,

This PR fixes nothing. I just miss the option to open recent projects quickly upon opening Zed, so I made this. Hope I can see it soon in Preview channel.
If there is any suggestion, just comment. I will take it seriously.

Thank you!

|ui|before|after|
|-|-|-|
|empty pane|Screenshot 2025-12-03 at 12 39 25|Screenshot 2025-12-03 at 12 34 03|
|new window|Screenshot 2025-12-03 at 12 39 21|Screenshot 2025-12-04 at 10 43 17|

---

Release Notes:

- Added a new value to the `restore_on_startup` setting called `launchpad`. This value makes Zed open with a variant of the welcome screen ("the launchpad") upon startup. Additionally, this same page variant is now also what is displayed if you close all tabs in an existing window that doesn't contain any folders open. The launchpad page shows you up to 5 recent projects, making it easy to open something you were working recently.


mattermill:
@simonpham Thanks for the contribution! This would be a good option to provide.

For this case, I would prefer not creating an entirely new `launchpad.rs`file and instead pass properties to the existing `welcome.rs` to swap out the configuration options with the recent projects as you have. This helps us keep startup screens consolidated and easier to revise in the future.

In summary:
- Pass properties to `welcome.rs` to use "Recent projects" instead of "Configuration options" when arriving at `welcome.rs` via "New window," or when closing all tabs
- Maintain existing `welcome.rs` behavior that displays configuration options after a user finished onboarding

If you can make that change, I see no issue with providing this as an option.

simonpham:
> @simonpham Thanks for the contribution! This would be a good option to provide.
> 
> For this case, I would prefer not creating an entirely new `launchpad.rs`file and instead pass properties to the existing `welcome.rs` to swap out the configuration options with the recent projects as you have. This helps us keep startup screens consolidated and easier to revise in the future.
> 
> In summary:
> 
> * Pass properties to `welcome.rs` to use "Recent projects" instead of "Configuration options" when arriving at `welcome.rs` via "New window," or when closing all tabs
> * Maintain existing `welcome.rs` behavior that displays configuration options after a user finished onboarding
> 
> If you can make that change, I see no issue with providing this as an option.


That's what I thought.

~Currently I can "Return to Setup" by go to Help -> Show Welcome. Should I hide this when user have already set up?~
I will only hide Return to Setup for Welcome with recent projects.

simonpham:
@mattermill please help to check if there is any other issue. Thank you!

mattermill:
@simonpham 

When launching a New Window I'm seeing the Welcome screen appear for an instant before being replaced by the empty file. Closing the file returns to the Welcome. I think you just want the Welcome in this instance, not new file, correct?

https://github.com/user-attachments/assets/cc13c0a0-3509-453a-bbf1-037bd1569cc9

simonpham:
hi @mattermill 

You should set the setting  "restore_on_startup": "welcome" to see it on new window (when open Zed)

I only expect it to appear if you use the above option and in empty pane when no workspace is open.

For the case when open a new window within a Zed instance, I don't change the current behavior of it.
The previous UI of empty pane also has the same behaviour:

https://github.com/user-attachments/assets/ca24309a-9e52-4aeb-9084-dad9d022657c

Meonix:
hi @mattermill  Could you please continue to review this PR to help us?

simonpham:
hi @mattermill, it's been a while. Please let me know if it can be merged. Thank you!

danilo-leal (approved):
Thank you so much for kicking this off! Pushed some additional changes here: aside from design refinements, I felt like the `none` value terminology for `restore_on_startup` was confusing, because that actually meant opening Zed with an empty file... it felt like it could be a bit more clear, so, I changed it to `empty_tab` instead. Then, the value to open the launchpad page (the variant, recent projects-focused welcome page) is just `launchpad`. Already included the migration for anyone who uses `restore_on_startup: "none"`, so we should be good there :) Thanks again!

VihaanReddyM:
Has this PR been shipped?

simonpham:
> Has this PR been shipped?

hi @VihaanReddyM
you can try it in Preview channel.

PR #42942, opened 2025-11-18
Add cross-window file/folder copy-paste support via system clipboard

mxvsh:
## Summary

This PR adds support for copying and pasting files/folders across different Zed windows using the system clipboard.

Previously, the paste button was disabled when trying to paste files copied from another Zed window. This implementation enables:

- **Cross-window clipboard**: Copy files in one Zed window and paste them in another
- **Cross-instance support**: Works between different Zed instances
- **External file integration**: Can paste file paths copied from system file managers (Finder/Explorer)
- **Cut/Copy distinction**: Preserves whether files were cut or copied

## Implementation

- Modified `copy()` and `cut()` to write file paths to system clipboard with format: `ZED_CLIPBOARD:CUT|COPY\n/path1\n/path2...`
- Added `has_clipboard_content()` to check both local and system clipboard
- Added `parse_system_clipboard()` to parse Zed format or external file paths
- Updated `paste()` to check local clipboard first, then fall back to system clipboard
- Added `write_entries_to_system_clipboard()` helper function

## Test Plan

1. Open two Zed windows
2. Copy files/folders in window A (Cmd+C / Ctrl+C)
3. Switch to window B
4. Right-click in project panel → Paste option should now be enabled
5. Paste successfully copies the files to the target location

## Release Notes:

  - Added cross-window file/folder copy-paste support in project panel via system
  clipboard

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #44068, opened 2025-12-03
Add ability to copy/paste across windows

mxvsh:
Improvements of the PR #42942 

Updated the implementation to use platform-native clipboard metadata instead of polluting the visible text.

## Release Notes:
- Added cross-window file/folder copy-paste support in project panel via system
clipboard

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -1048,11 +1048,7 @@ impl ProjectPanel {
                             .action("Cut", Box::new(Cut))
                             .action("Copy", Box::new(Copy))
                             .action("Duplicate", Box::new(Duplicate))
-                            .action_disabled_when(
-                                !has_clipboard,
-                                "Paste",
-                                Box::new(Paste),
-                            )
+                            .action_disabled_when(!has_clipboard, "Paste", Box::new(Paste))
                             .separator()
                             .action("Copy Path", Box::new(zed_actions::workspace::CopyPath))
                             .action(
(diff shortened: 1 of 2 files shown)


spandan-appointy (dismissed):
need in depth review

ConradIrwin:
Nice, thanks!

I don't think we should have the fast-path for not using the system clipboard, as otherwise you get weird behaviur:
* In window 1, copy file A, in window 2, paste file (pastes A as expected). In window 2 copy file B, in window 1 paste file (pastes A instead of B).
* We should use an enum not the string for the operation.
* I notice that cross-window cut/paste doesn't remove the file from the other side. Not sure exactly how to fix that, and happy to ship this without that and come back to it.

mxvsh:
hi, @ConradIrwin, thanks for your response, few changes I made in the code, please have a look

mxvsh:
For the local clipboard issue, I have a question:

  **Current implementation:**
  - `copy()/cut()` writes to both local clipboard AND system
  clipboard
  - `paste()` checks **local first**, then system (as
  fallback for cross-window)
  - All tests pass ✅

  **The tradeoff:**
  - ✅ Same-window paste: Fast, maintains proper
  selection/focus/state
  - ✅ Cross-window paste: Works via system clipboard
  fallback
  - ❌ Cross-window with mixed windows still has stale issue
  (Window 1 copies A, Window 2 copies B, Window 1 pastes →
  gets A from local instead of B from system)

  **To fix the stale issue completely**, I'd need to check
  system clipboard first. But then paste from system
  clipboard goes through `drop_external_files()` which:
  - Doesn't maintain selection state correctly
  - Doesn't expand/collapse directories properly
  - Doesn't trigger rename editor for disambiguated files
  - Causes all the tests to fail

  **Question:** Is the current hybrid approach (local first,
  system fallback) acceptable? Or should I refactor
  `drop_external_files()` to behave exactly like
  `paste_from_local_clipboard()` so we can prioritize system
  clipboard?

  Essentially: different windows naturally sync via system
  clipboard, but same window gets the performance/state
  benefits of local clipboard. Is this acceptable or should
  every paste check system first regardless?

ConradIrwin:
Is the performance difference significant? I'd have thought that checking the system clipboard should be very fast.

I think the cross-window thing right now is confusing and that we should not ship the branch in its current state.

PR #40000, opened 2025-10-11
ruby: Rename `HTML/ERB` to `HTML+ERB`

vitallium:
Hi! In https://github.com/zed-extensions/ruby/issues/162 we renamed embedded template languages:

- `HTML/ERB` to `HTML+ERB`
- `YAML/ERB` to `YAML+ERB`
- `JS/ERB` to `JS+ERB`

This pull request updates the Ruby extension documentation to reflect that change. Thanks!

Release Notes:

- N/A

PR #44081, opened 2025-12-03
language: Remove old unused `HTML/ERB` language ID

vitallium:
The `HTML/ERB` language was renamed
to `HTML+ERB` in https://github.com/zed-industries/zed/pull/40000 We can remove the old name safely now.

Release Notes:

- N/A


MrSubidubi (approved):
Yep, sounds good. Thanks!

PR #44082, opened 2025-12-03
gpui: Make HTTP client optional

HigherOrderLogic:
Make the Http client optional as a feature.

The `http_client` crate pulls in `tokio`, which is kinda heavy to compile, and not all apps need the ability to fetch, so I make this an optional feature.

Release Notes:

- N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -305,7 +305,7 @@ git_hosting_providers = { path = "crates/git_hosting_providers" }
 git_ui = { path = "crates/git_ui" }
 go_to_line = { path = "crates/go_to_line" }
 google_ai = { path = "crates/google_ai" }
-gpui = { path = "crates/gpui", default-features = false }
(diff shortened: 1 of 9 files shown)


mikayla-maki:
> The http_client crate pulls in tokio

That seems like it's an error on our part.

Looks like it's because of this multipart form data dependency here: https://github.com/zed-industries/zed/blob/main/crates/http_client/src/http_client.rs#L117, we should instead try to break that dependency.

HigherOrderLogic:
> Looks like it's because of this multipart form data dependency here: [`main`/crates/http_client/src/http_client.rs#L117](https://github.com/zed-industries/zed/blob/main/crates/http_client/src/http_client.rs?rgh-link-date=2025-12-04T06%3A28%3A38Z#L117), we should instead try to break that dependency.

So should I create a new PR that just pull the implementation of the struct into the repo instead?

PR #44087, opened 2025-12-03
Add option to show tab bar inside title bar

asmimo:
Added new setting `title_bar.show_tab_bar` (default: `false`)

When enabled, regular tab bar is hidden and tabs are rendered in title bar. 
Only tabs from the current active pane are shown. 
On macOS, window is made non-movable when tabs are in the title bar to prevent conflicts with drag interactions and requires **application restart** (notification shown to users)

Note: Tested on macOS only

 Demo

https://github.com/user-attachments/assets/2bdf5075-41bd-4c1a-ab3f-5f4250cdc720

Release Notes:

- Added title_bar.show_tab_bar setting to display tab bar inside the title bar
  (https://github.com/zed-industries/zed/issues/5066)

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -439,7 +439,10 @@
     // Whether to show the sign in button in the titlebar.
     "show_sign_in": true,
     // Whether to show the menus in the titlebar.
-    "show_menus": false
+    "show_menus": false,
+    // Whether to show the tab bar in the title bar.
+    /// Need to restart for changes to apply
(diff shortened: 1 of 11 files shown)


mattermill:
As much as I'd like to see this through, we're currently not taking any issues related to hiding or adding tabs to the title bar while we work on some fundamental changes to other elements that would be affected.

Related: https://github.com/zed-industries/zed/pull/42540#issuecomment-3522624069

PR #44089, opened 2025-12-03
git: Add tree view support to Git Panel

### LINKED ISSUE #35803: Tree view for Git Panel

xipeng-jin:
Closes #35803

This PR adds tree view support to the git panel UI as an additional setting and moves git entry checkboxes to the right.  Tree view only supports sorting by paths behavior since sorting by status can become noisy, due to having to duplicate directories that have entries with different statuses.

### Tree vs Flat View
image
image

#### Architecture changes

Before this PR, `GitPanel::entries` represented all entries and all visible entries because both sets were equal to one another. However, this equality isn't true for tree view, because entries can be collapsed. To fix this, `TreeState` was added as a logical indices field that is used to filter out non-visible entries. A benefit of this field is that it could be used in the future to implement searching in the GitPanel.

Another significant thing this PR changed was adding a HashMap field `entries_by_indices` on `GitPanel`. We did this because `entry_by_path` used binary search, which becomes overly complicated to implement for tree view. The performance of this function matters because it's a hot code path, so a linear search wasn't ideal either. The solution was using a hash map to improve time complexity from O(log n) to O(1), where n is the count of entries. 

#### Follow-ups
In the future, we could use `ui::ListItem` to render entries in the tree view to improve UI consistency. 
 
Release Notes:

- Added tree view for Git panel. Users are able to switch between Flat and Tree view in Git panel.

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -289,7 +289,7 @@ struct TreeKey {
 #[derive(Debug, PartialEq, Eq, Clone)]
 struct GitTreeDirEntry {
     key: TreeKey,
-    name: String,
+    name: SharedString,
     depth: usize,
     staged_state: ToggleState,
     expanded: bool,
@@ -297,9 +297,9 @@ struct GitTreeDirEntry {
 
 #[derive(Default)]
 struct TreeNode {
-    name: String,
+    name: SharedString,
     path: Option<RepoPath>,
-    children: BTreeMap<String, TreeNode>,
+    children: BTreeMap<SharedString, TreeNode>,
     files: Vec<GitStatusEntry>,
 }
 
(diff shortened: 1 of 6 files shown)


crates/git_ui/src/git_panel.rs
```diff
@@ -543,51 +606,17 @@ impl GitPanel {
...
+    // since entries aren’t sorted there. This will keep features like checkbox states and bulk staging anchors
+    // working smoothly without overcomplicating the logic.
+    pub fn entry_by_path(&self, path: &RepoPath, _cx: &App) -> Option<usize> {
+        // Lists can contain headers, directory nodes, and different sort/group modes;
+        // just scan the rendered entries for a matching status path.
+        self.entries.iter().position(|entry| {
+            entry
+                .status_entry()
+                .is_some_and(|status| &status.repo_path == path)
+        })
```
Anthony-Eid: I'm worried that there was a reason the old entry by path was a binary search before, and I don't want to regress performance when there's a large amount of modified files in the git panel.

Could you please maintain the old behavior when the git panel is not in tree view? We might be able to get away with caching this data with a hash map as well

aalpgiray:
How do we track when this will be release?

Anthony-Eid:
@aalpgiray It's going to be release shortly after new years

Issue #43884, opened 2025-12-01
Remote SSH Connection drops almost immediately


aaronbuckner:
### Reproduction steps

I have many remote connections defined.  A few weeks ago, almost all my connections stopped working. 

1) Start Zed on Mac
2) Open a remote connection
3) Open a file or 2 and edit the pretty code
4) Try to save
5) Realize Zed is not connected anymore

### Current vs. Expected behavior

Zed disconnects from the server almost immediately.  It should stay connected.

I am not sure what to provide here.  It seems so pervasive to me that it's rendered the program that I grew to love completely useless to me.  I cannot do any development using this program now.

Worth noting that both a normal Terminal window and VS Code have no problems remaining connected to these same servers for hours and even days.

### Zed version and system specs

MacOS 15.7.1 (24G231)
Zed 0.213.7 

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

aaronbuckner:
Maintainers: please tag this issue with `remote`

marcocondrache:
@aaronbuckner Could you provide more information? The logs from when the disconnection occurs (accessible with `zed: open log`) would be helpful. Additionally, sharing your settings would be useful. I also noticed that you are not using the latest version. Could you try updating and see if that makes a difference?

zelenenka:
would it be possible for you to update to the latest stable zed (0.214.7) and attach the logs?

saidelike:
1. Are you connected to some kind of VPN or Corporate environment?
2. Do you have to deal with an HTTP proxy?
If so, I am wondering if they are the same issues as https://github.com/zed-industries/zed/issues/43694. I am having this other issue but never used zed before, so am wondering if it is a regression and if I used an old zed version if I would not have my issue.

aaronbuckner:
I updated to the newest version -- I used it for a while and maybe got 15 min into it, but the same thing happened.  

I am on a VPN, yes.  Wireguard, but again this only started 2 weeks ago -- Mac Terminal works great and does not disconnect, and VS Code also does not disconnect.

It seems to occur only on my Digital Ocean servers that I know, AWS is fine (so far).  

Log and Settings are attached -- I hope this helps, or I'll have to redact what you ask for specifically.  Thank you!

[...]

Issue #43809, opened 2025-11-29
Remote development becomes unstable on 1 vCPU servers after Zed 0.200+ (heartbeat timeouts)


pitrozx:
## Reproduction Steps

1. Create any VPS with:
   - 1 vCPU AMD EPYC
   - 1–2 GB RAM
2. open remote development SSH into the machine with Zed 0.200+.
3. open big project example ipfs/kubo
4. Wait 300–600 seconds (even idle).
5. Heartbeats begin to fail.
6. Zed goes into a reconnect loop.

Upgrading this exact instance to **2 vCPU** fixes the issue immediately.

## Expected Behavior

Zed Remote Server should not lose heartbeat on idle workload, even on minimal CPUs.  
Versions before 0.200 handled 1 vCPU environments normally.

---

## Actual Behavior

On a 1 vCPU host, Zed Remote Server misses heartbeats and enters an infinite reconnect loop.

### Zed version and system specs

zed 0.212.5
os: Distributor ID:	ManjaroLinux
Description:	Manjaro Linux
Release:	25.0.0
Codename:	Zetar
 Linux d469f12fd92c 6.8.0-58-generic #60-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 14 18:29:48 UTC 2025 x86_64 GNU/Linux
memory 2GB 
cpu: AMD EPYC 7001

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
25-11-14T18:21:14+02:00 INFO  [remote_server::unix] (remote server) Got new node settings: NodeBinarySettings { path: None, npm_path: None, ignore_system_version: false }

2025-11-14T18:21:15+02:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: vscode-html-language-server
2025-11-14T18:21:15+02:00 INFO  [lsp] (remote server) starting language server process. binary path: "/home/ptrzx/.local/share/zed/languages/gopls/gopls_0.20.0_go_1.24.3", working directory: "/home/ptrzx/nb/tpb/gtpb", args: ["-mode=stdio"]
2025-11-14T18:21:16+02:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: intelephense
2025-11-14T18:21:16+02:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: phpactor
2025-11-14T18:21:16+02:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: phptools
[...]

PR #44110, opened 2025-12-04
linux: Spawn at least two background threads

agu-z:
Related to https://github.com/zed-industries/zed/pull/44109, https://github.com/zed-industries/zed/issues/43884, https://github.com/zed-industries/zed/issues/43809. 

In the Linux dispatcher, we create one background thread per CPU, but when a single core is available, having a single background thread significantly hinders the perceived performance of Zed. This is particularly helpful when SSH remoting to low-resource servers.

We may want to bump this to more than two threads actually, but I wanted to be conservative, and this seems to make a big difference already.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/platform/linux/dispatcher.rs
+++ b/crates/gpui/src/platform/linux/dispatcher.rs
@@ -26,12 +26,13 @@ pub(crate) struct LinuxDispatcher {
     main_thread_id: thread::ThreadId,
 }
 
+const MIN_THREADS: usize = 2;
+
 impl LinuxDispatcher {
     pub fn new(main_sender: Sender<RunnableVariant>) -> Self {
         let (background_sender, background_receiver) = flume::unbounded::<RunnableVariant>();
-        let thread_count = std::thread::available_parallelism()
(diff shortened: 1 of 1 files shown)


crates/gpui/src/platform/linux/dispatcher.rs
```diff
@@ -31,7 +31,8 @@ impl LinuxDispatcher {
         let (background_sender, background_receiver) = flume::unbounded::<RunnableVariant>();
         let thread_count = std::thread::available_parallelism()
             .map(|i| i.get())
-            .unwrap_or(1);
+            .unwrap_or(1)
+            .max(2);
```
HigherOrderLogic: Maybe use `map_or` to reduce 1 allocation/function call?

```suggestion
            .map_or(2, |v| v.max(2));
```

agu-z: I don't think this adds any allocations, but I agree that the `.unwrap(1)` immediately followed by `max(2)` feels a bit weird. Fixed it!

PR #44155, opened 2025-12-04
agent: Add allowed_paths setting for file access outside project (Claude Code planning issue)

### LINKED ISSUE #37620: AI: Claude Code + Zed asking weird folder permissions
### Summary

Zed asking for permissions to desktop, photos, documents and downloads after trying to use Claude Code

### Description

Steps to trigger the problem:
1. Fresh Zed Install
2. Open AI right dock
3. Click on `New Claude Code Thread`
4. Installing Claude Code dialog appears
5. Zed asking for access to:
    - Desktop (did not make a screenshot)
    - Pictures (screenshot attached)
6. Deny all and do step 3. again
7. Zed asking for access to:
    - Documents folder
    - Downloads Folder

Image

**Expected Behavior**:
- No weird access requirements?
**Actual Behavior**:
- Zed (or more like Claude, I'm suspecting) asking for permissions to "random" folders on my Mac

Image
Image
Image

### Model Provider Details
- Provider: Anthropic Claude Code Thread
- Model Name:
- Mode: (Agent Panel, Inline Assistant, Terminal Assistant or Text Threads)

Zed: v0.202.7 (Zed) 
OS: macOS 15.6.1
Memory: 128 GiB
Architecture: aarch64

raphaelluethy:
This PR resolves an issue that happens with claude code when trying to use their plan mode: The agent sometimes can write to `~/.claude/plans` and sometimes cant (using echo or cat vs the write tool). It also tries all the time to create the .claude folder again.

I tried to steer it with prompting but that was not really reliable, so I thought this could be added as a setting.

 This PR adds a new setting that allows users to enable access to folders outside of the projects directory.

Config: 
```json
  "agent_servers": {
    "claude": {
      "allowed_paths": ["~/.claude"]
    }
  }
```

This works for EVERY external agent, CC, Gemini, Codex etc, but the current most obvious case would be CC.

I also added tests for folders that are allowed outside of project, for those that arent, and for those that have similar names (/open and /open_backup).
It also checks for normalized paths, e.g. `~/.claude/../../` should be erroring (also addes test for that). I found that there was already a util func for that, so thats being used.

I explicitly added the checks after the inital check for the worktree/project access.

Would be happy for any feedback.

Closes #37620 (partially)
Closes #44221

Release Notes:

- Added `allowed_paths` setting to external agents to permit claude code to write plans.

### DIFF
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -32,7 +32,12 @@ use std::ops::Range;
 use std::process::ExitStatus;
 use std::rc::Rc;
 use std::time::{Duration, Instant};
-use std::{fmt::Display, mem, path::PathBuf, sync::Arc};
+use std::{
+    fmt::Display,
+    mem,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 use ui::App;
 use util::{ResultExt, get_default_system_shell_preferring_bash, paths::PathStyle};
 use uuid::Uuid;
@@ -838,6 +843,7 @@ pub struct AcpThread {
     terminals: HashMap<acp::TerminalId, Entity<Terminal>>,
     pending_terminal_output: HashMap<acp::TerminalId, Vec<Vec<u8>>>,
     pending_terminal_exit: HashMap<acp::TerminalId, acp::TerminalExitStatus>,
+    allowed_paths: Vec<PathBuf>,
 }
 
 impl From<&AcpThread> for ActionLogTelemetry {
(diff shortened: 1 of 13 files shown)


vixalien:
I think preferably, this should be enabled by default, as having Claude Code without access to `~/.claude` is rather pointless

raphaelluethy:
> I think preferably, this should be enabled by default, as having Claude Code without access to `~/.claude` is rather pointless

I strongly agree, but I would leave that up to the team to decide if they would like to keep it as is or set this as new default (since its a setting, it could be easily added as default setting).

oskarhane:
Yes, this is much needed! The experience we have now is very frustrating.

benbrandt:
Hey @raphaelluethy thanks for doing this. However, I think we want to try handling this at the adapter layer first, will look into that today

raphaelluethy:
@benbrandt Thats okay. Then I leave the PR as is and wont check for updates from the main branch for now.
Thanks for having a look at this, as long as I can save my plans again, I dont mind either way :)

benbrandt:
@raphaelluethy I fixed it here: https://github.com/zed-industries/claude-code-acp/pull/200
Release going out now.
I'll go ahead and close this PR and we can revisit if we feel we need it for other reasons. Thanks again!

raphaelluethy:
@benbrandt Thank you too for the quick fix 👍

PR #44162, opened 2025-12-04
git_ui: Clean up the commit view UI

danilo-leal:
- Don't render the dirty indicator slot in the multibuffer header when the editor is read-only (tighter spacing)
- Improve the tab tooltip content UI by using `Tooltip::element` instead of just text
- Use the `DiffStat` component to show the total amount of lines added and removed
- Remove indentation lines from the editor that displays the commit description
- Ensure the header collapses well when the width gets smaller

Screenshot 2025-12-04 at 1  07@2x

Release Notes:

- N/A

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -20083,6 +20083,11 @@ impl Editor {
         self.show_indent_guides
     }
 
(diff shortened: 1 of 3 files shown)


JosephTLyons:
Amazing! ❤️

PR #44165, opened 2025-12-04
remote: Use last line of `uname` and shell output

agu-z:
We have seen cases (see https://github.com/zed-industries/zed/issues/43694) where the user's shell initialization script includes text that ends up in the output of the commands we use to detect the platform and shell of the remote. This solution isn't perfect, but it should address the issue in most situations since both commands should only output one line.

Release Notes:

- remote: Improve resiliency when initialization scripts output text


agu-z:
re danger: this PR does not close the whole issue, it just addresses something the user mentioned

Veykril (approved):
Yea we should've done this from the start after all. We'll need to do the same for WSL as well I think

PR #44177, opened 2025-12-04
gpui: Improve path rendering performance on Windows

CherryWorm:
This pull request massively improves the performance of drawing paths on Windows by getting rid of unnecessary copies and allocations, and reducing the memory bandwidth. 

To verify the performance improvement, this pull request also adds an example that draws 500 paths with 2000 points each, totalling 1M points. On my ASUS ROG Flow Z13, this runs at 2.5 FPS against main:

image

Using the newly introduced `paint_cached_path` api, this pull request improves frame times by a factor of 25:
image

And using the existing ´draw_path´ api, frame times are still improved by a factor of 6:

image

This should have positive downstream effects for Zed as well.

I don't have a working setup to profile the mac and linux implementations, but I imagine easy gains could be found there as well. With larger API changes, Windows performance could be improved by another 1.5x-2x. 

Release Notes:

- Massively improved path rendering performance on Windows


CherryWorm:
@cla-bot check

CherryWorm:
@cla-bot check

CherryWorm:
The mac renderer has the exact same issues, in case anyone with a working mac setup wants to work on this: https://github.com/zed-industries/zed/blob/82650d473aaedc2b2756e9028f3b384e857149c2/crates/gpui/src/platform/mac/metal_renderer.rs#L596

Same on linux:
https://github.com/zed-industries/zed/blob/82650d473aaedc2b2756e9028f3b384e857149c2/crates/gpui/src/platform/blade/blade_renderer.rs#L599

marcocondrache:
@CherryWorm the example doesn't seem to run on macOS:

image

This is the command i've used: `cargo run --example mandala_bench --features runtime_shaders`

CherryWorm:
Hm I can't even see the fps counter in the top left on your screen, and my changes to the Mac rendering definitely shouldn't change anything about it's behavior. Maybe the Mac renderer struggles to draw the examples frame even once? Can you confirm whether the canvases paint function actually executes? And try mandala_bench_cloned as well, that one uses the old API (though that shouldn't make a difference).

I also didn't test with runtime_shaders, maybe try dropping that.

CherryWorm:
You can also execute madala_bench_cloned against main, to confirm whether this is an issue with my changes or the existing implementation of paths

marcocondrache:
@CherryWorm, I've tried the cloned example on the main branch, and it has the same issue. I believe the problem is not related to your changes. I will apply some of your optimizations to the `metal_renderer` this weekend and investigate this issue further. If it's okay with you, I plan to create another PR. Otherwise, I would love to collaborate on this.

marcocondrache:
@CherryWorm I’ve reduced the number of paths to 100 and I'm only getting about 4 fps. Performance on macOS is so poor that with 500 paths I don’t even see the first frame. For context, I’m running this on an M4 MacBook Pro with 16 GB of RAM.

I'm preparing another PR to improve this on macOS.

CherryWorm:
Oh damn didn't realize this pull request was reopened and rebased my own fork to main. Not sure why github doesn't understand this though

PR #44180, opened 2025-12-04
languages: Add injections for string and tagged template literals for JS/TS(X)

vitallium:
Hi! This pull request adds language injections for string and tagged template literals for JS/TS(X).
This is similar to what [this extension](https://marketplace.visualstudio.com/items?itemName=bierner.comment-tagged-templates) provides for VSCode. This PR is inspired by this tweet https://x.com/leaverou/status/1996306611208388953?s=46&t=foDQRPR8oIl1buTJ4kZoSQ

I've added injections queries for the following languages: HTML, CSS, GraphQL and SQL.
This works for:

- String literals: `const cssString = /* css */'button { color: hotpink !important; }';`
- Template literals: ```const cssString = /* css */`button { color: hotpink !important; }`;```

All injections support the format with whitespaces inside, i.e. `/* html */` and without them `/*html*/`.

## Screenshots

|before|after|
|---------|-----------|
|  CleanShot 2025-12-04 at 21 12 00@2x | CleanShot 2025-12-04 at 21 08 35@2x|

Release Notes:

- Added language injections for string and tagged template literals in JS/TS(X)

### DIFF
--- a/crates/languages/src/javascript/injections.scm
+++ b/crates/languages/src/javascript/injections.scm
@@ -83,3 +83,46 @@
   arguments: (arguments (template_string (string_fragment) @injection.content
                               (#set! injection.language "isograph")))
 )
+
+; Parse the contents of strings and tagged template
+; literals with leading ECMAScript comments:
+; '/* html */' or '/*html*/'
+(
+  ((comment) @_ecma_comment [
+    (string (string_fragment) @injection.content)
+    (template_string (string_fragment) @injection.content)
+  ])
+  (#match? @_ecma_comment "^\\/\\*\\s*html\\s*\\*\\/")
+  (#set! injection.language "html")
+)
+
+; '/* sql */' or '/*sql*/'
+(
+  ((comment) @_ecma_comment [
+    (string (string_fragment) @injection.content)
+    (template_string (string_fragment) @injection.content)
+  ])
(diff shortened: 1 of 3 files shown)


crates/languages/src/javascript/injections.scm
```diff
@@ -83,3 +83,46 @@
...
+
+; '/* sql */' or '/*sql*/'
+(
+  ((comment) @_ecma_comment [
+    (string (string_fragment) @injection.content)
+    (template_string (string_fragment) @injection.content)
+  ])
+  (#match? @_ecma_comment "^\\/\\*\\s*sql\\s*\\*\\/")
+  (#set! injection.language "sql")
+)
```
vitallium, line 107: I wish tree-sitter provided a way to capture groups in regular expressions to avoid writing similar repetitive queries.

probably-neb:
Sorry for the delay, this looks good! Nice work!

DavidBK:
Thank you so much! nice work.
I notice that the string template in js "break" the syntax highlighting:
```js
const test = /* sql */ `SELECT * from test_table WHERE id = '${test_id}' SORT BY name;`;
```
image

jaydenseric:
The matching of the comments should have been case insensitive; `/* GraphQL */` isn't recognised by Zed like it is in VS Code and by Prettier.

PR #44198, opened 2025-12-05
language: Make `TreeSitterData` only shared between snapshots of the same version

Veykril:
Currently we have a single cache for this data shared between all snapshots which is incorrect, as we might update the cache to a new version while having old snapshots around which then may try to access new data with old offsets/rows.

Release Notes:

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -22,8 +22,8 @@ pub use crate::{
     proto,
 };
 use anyhow::{Context as _, Result};
+use clock::Lamport;
 pub use clock::ReplicaId;
-use clock::{Global, Lamport};
 use collections::{HashMap, HashSet};
 use fs::MTime;
 use futures::channel::oneshot;
@@ -33,7 +33,7 @@ use gpui::{
 };
 
 use lsp::{LanguageServerId, NumberOrString};
-use parking_lot::{Mutex, RawMutex, lock_api::MutexGuard};
+use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use settings::WorktreeId;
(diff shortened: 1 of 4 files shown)


crates/language/src/buffer.rs
```diff
@@ -4278,155 +4289,125 @@ impl BufferSnapshot {
...
-                    .changed_since(known_version)
-                {
-                    known_chunks.clone()
-                } else {
-                    HashSet::default()
-                }
-            }
-            None => HashSet::default(),
-        };
+        let known_chunks = known_chunks.cloned().unwrap_or_default();
```
SomeoneToIgnore: NIT: at this state, seems more appropriate to accept by value these new `known_chunks`?

crates/language/src/buffer.rs
```diff
@@ -130,26 +130,30 @@ pub struct Buffer {
...
-    tree_sitter_data: Arc<Mutex<TreeSitterData>>,
+    tree_sitter_data: Arc<TreeSitterData>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct TreeSitterData {
     chunks: RowChunks,
-    brackets_by_chunks: Vec<Option<Vec<BracketMatch<usize>>>>,
+    brackets_by_chunks: Mutex<Vec<Option<Vec<BracketMatch<usize>>>>>,
```
SomeoneToIgnore, line 139: The way I understand the data flow now:

* we only start with clean data and clean it on each tree-sitter reparsed event

* each snapshot gets an `Arc<TreeSitterData>`, the same one, with the field they want to access + clone, or initialize once + clone

No blocks and mutex are needed for that, it seems?
I wonder if this could be a `Vec<OnceCell<Vec<BracketMatch<usize>>>>` given that we're synchronous around this code in the near future?

crates/language/src/buffer.rs
```diff
@@ -1746,7 +1750,14 @@ impl Buffer {
         self.syntax_map.lock().did_parse(syntax_snapshot);
         self.request_autoindent(cx);
         self.parse_status.0.send(ParseStatus::Idle).unwrap();
-        self.tree_sitter_data.lock().clear();
+        let snapshot = self.text.snapshot();
+        match Arc::get_mut(&mut self.tree_sitter_data) {
+            Some(tree_sitter_data) => tree_sitter_data.clear(snapshot),
```
SomeoneToIgnore: Seems that `clear` is used once here and needs the same parameters to do the same thing `new` does.
The `get_mut` trick is neat, but seems that we can just always do 

```rs
let tree_sitter_data = TreeSitterData::new(snapshot);
self.tree_sitter_data = Arc::new(tree_sitter_data);
```

, remove `clear` along with this `match` and keep things simpler?

crates/language/src/buffer/row_chunk.rs
```diff
@@ -51,10 +50,6 @@ impl RowChunks {
         }
     }
 
-    pub fn version(&self) -> &Global {
```
SomeoneToIgnore, line 54: With this and `latest_tree_sitter_data` gone, my biggest concern now is that now we technically have more chances to have an outdated `snapshot`.

Now, without the version, I wonder whether we need to store this at all?
Looking at the usages:
 
* `impl std::fmt::Debug for RowChunks` 

Does not hurt to know the version these chunks are for, but it's accessible from their parent buffer/snapshot

* `chunk_range`

For that, we can precompute the ranges on creation.

* `applicable_chunks` 

Here it can explode due to outdated snapshot, but it's used in a very stupid way: it converts input anchors into points.
We can require callers to do that with their correct snapshots.
I still wonder if the version check has to happen in this case somehow...

SomeoneToIgnore (approved):
Nice, much more correct way to fill this cache.

crates/language/src/buffer.rs
Veykril: The mutex is necessary because snapshots will lazily populate the data here. The mutex allows us to share this work between snapshots of the same version. Without it each snapshot will have to recompute things themselves. OnceCell would require us to completely populate this data in one access for the entire underlying buffer

crates/language/src/buffer.rs
Veykril: I don't see the need for that, we might as well retain the allocations if we can for a little bit of perf, here is little complexity added through this imo

crates/language/src/buffer/row_chunk.rs
Veykril: Yea, I replaced the debug assert with a branch that uses a fresh tree sitter data object if we are outdated in `snapshot` creation. So in the worst case we will get a small perf hit here if something goes out of order. (like it does when we apply a diagnostic op as that eagerly fetches a snapshot before applying edits)

PR #44204, opened 2025-12-05
Concurrent multi-panel, multi-agent operation

brucestarlove:
Closes #41081 (feature request)

Release Notes:
- Agent Panel: Multiple threads now persist across sessions, operating concurrently. Tab bar shows when 2+ threads are open. Added keyboard actions for switching between threads.


danilo-leal:
Hey, thanks for working on this, but this feature is already on our radar and the design we're pursuing will require some more deeper changes to the Zed workspace, which will also allow for other concurrent/parallel agent-related flows in the future. Therefore, I'll close this one for now, but thank you again!

khush2808:
hey @danilo-leal ,any update on if this - have parallel agents been added.

PR #44210, opened 2025-12-05
 Add vim/emacs modeline support 

### LINKED ISSUE #4762: Support modeline comments for per-file settings

elmarco:
Many editors such as vim and emacs support "modelines", a comment at the beginning of the file that allows the file type to be explicitly specified along with per-file specific settings

- The amount of configurations, style and settings mapping cannot be handled in one go, so this opens up a lot of potential improvements.
- I left out the possiblity to have "zed" specific modelines for now, but this could be potentially interesting.
- Mapping the mode or filetype to zed language names isn't obvious either. We may want to make it configurable.

This is my first contribution to zed, be kind. I struggled a bit to find the right place to add those settings. I use a similar approach as done with editorconfig (merge_with_editorconfig). There might be better ways.

Closes #4762

Release Notes:

- Add basic emacs/vim modeline support.

Supersedes #41899, changes:
- limit reading to the first and last 1kb
- add documentation
- more variables handled
- add Arc around ModelineSettings to avoid extra cloning
- changed the way mode -> language mapping is done, thanks to `modeline_aliases` language config
- drop vim ex: support
- made "Local Variables:" handling a separate commit, so we can drop it easily
- various code style improvements


ConradIrwin:
Thanks for updating this.

Looking at the diff I'm struggling to see what we should do here; but we need to figure out how to make this feature much less invasive (like closer to 5 places in the codebases should know about modelines compared to ~55 files changed in this diff).

Random thoughts:
* Given that we expect most files to not have modelines, could the buffer_store add them to a new global in add_buffer. If it only affects LanguageSettings, then maybe there's a new global in the language create, or it could go in the settings crate. (I don't think it should go in the existing settings store to avoid updating the settings store when a file is opened/closed). Then we could avoid passing them around everywhere.
* Am I right in thinking modelines are only parsed when a file is opened? If you edit the modeline is there a way to get that change to show up?
* For selecting language via modeline, I think for a v1 the mapping should live in the modeline crate and not be spread around every single langauge's definition.
* The `fn merge` in `merge_from_modeline` should be the same as the `merge_from_option` used by the settings crate if you want to use that.

Does that make sense?

elmarco:
- I didn't manage to make this "global" approach. It looks fairly bad, as the modeline is attached to the buffer, having to look it up back from the file/buffer location is awkward and could easily leak. I will try a different approach where language_settings() is loaded more directly from buffer later.
- yes, modelines are typically loaded during file open / reload
- the mapping is alreay optional, I just added a few extra common mappings. I will split that in a different commit.
- ok (along with updating existing editorconfig fn merge)

elmarco:
@ConradIrwin have you had a chance to take another look? thanks

elmarco:
@ConradIrwin https://github.com/zed-industries/zed/compare/main...elmarco:zed:modeline
or the fix itself: https://github.com/zed-industries/zed/commit/198f486c6c8fa0f9084f1d8258078a4625801f59

let me know if I should rebase and make a new PR.

Issue #44057, opened 2025-12-03
No way to cancel a long running git hook


jvacek:
### Reproduction steps

1. Install pre-commit or prek or anything
2. Make a pre-commit-config.yaml with a job that **does a long sleep or something lengthy**
3. Make a git commit that triggers the hook

### Current vs. Expected behavior

Currently, what happens is that I get a spinning wheel on the status bar and cannot kill it. The only way to do that is to kill the editor

<img width="356" height="94" alt="Image" src="https://github.com/user-attachments/assets/7c778b63-9772-43d4-a11d-2acb570d9e19" />

It would be nice if there was a way to cancel the processes that are shown as "running" on the status bar.

### Zed version and system specs

Zed: v0.214.7 (Zed) 
OS: macOS 26.1
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

CharlesChen0823:
testing with `pre-commit` work well, might only failed with `prek`. see issue #37293 , might something had broken again.

jvacek:
> testing with pre-commit work well
Can you be specific about what you mean by "work well"?

I can confirm that this is for both pre-commit and prek, and I'm fairly confident that anything triggered as part of a git hook would do the same thing.

Try this pre-commit hook, and commit it using the zed git UI and then try to cancel it without killing the IDE

```
repos:
  - repo: local
    hooks:
      - id: sleepsies
        language: system
        name: Long sleep
        entry: sleep 6000
        pass_filenames: false
```

CharlesChen0823:
ahh, I misunderstand your mean. you need a way to cancel a long running git hook. But IMO, using `prek` will not finished the hook even with `entry: echo "testing"`, this is another problem.

jvacek:
> But IMO, using prek will not finished the hook even with entry: echo "testing", this is another problem.

This is not an issue I am facing, prek and pre-commit are completely interchangeable for me.

CharlesChen0823:
you can test in new version, it had broken again.

Issue #37293, opened 2025-09-01
Prek causes git commit to never complete


WAKayser:
### Summary

Prek (alternative to pre-commit) causes git commit to never complete and prevents various features from working. 

### Description

Recently I tried switching from pre-commit to prek. All seemed to work well. But when committing from zed the commit spinner keeps spinning infinitely. This also prevents any other update to git state or the agent panel from working. 

Steps to reproduce:
1. create git repo
2. install prek (uv tool install prek && prek install)
3. add a sample .pre-commit-config.yaml file. 
4. try to create a commit via the panel.

Sample file to use. 

```yaml
repos:
  - repo: local
    hooks:
      - id: test
        name: test
        entry: echo "Running tests..."
        language: system
        pass_filenames: false

fail_fast: true
```

**Expected Behavior**:
commit to succeed. 

**Actual Behavior**:
git commit spinner keeps showing in the bottom bar. 

Nothing on this shows in the log and there is no error message. I have set up zed to be build. Let me know what I can do to help troubleshoot this issue. 

### Zed Version and System Specs

Zed: v0.201.8 (Zed) 
OS: macOS 15.5.0
Memory: 18 GiB
Architecture: aarch64

j178:
This isn't just a prek sepcific issue, I can make a minimal reproducer:

```console
$ cargo init zed-reproducer
$ cd zed-reproducer
$ cargo add tokio -F full
```

Edit `src/main.rs`:

```rust
use tokio::process::Command;

#[tokio::main]
async fn main() {
    let output = Command::new("git").arg("version").output().await.unwrap();
    println!("{}", String::from_utf8_lossy(&output.stdout));
}
```

```console
$ cargo build
$ cp ./target/debug/zed-reproducer .git/hooks/pre-commit
```

Open this project in Zed, and commit in Git Panel, then the commit hangs.

Edit: This seems to only reproduce on macOS and with tokio async runtime.

bxb100:
**maybe** associate with https://github.com/smol-rs/async-io/pull/79

leoliu0605:
I'd like to work on this!

esthertrapadoux:
Woohoo!!!!! Assigned this to you @leoliu0605 !

leoliu0605:
[...]

PR #44212, opened 2025-12-05
git: Fix git hook hang with `prek`

CharlesChen0823:
Fix git hook hang when using with `prek`. Can see [comments](https://github.com/zed-industries/zed/issues/44057#issuecomment-3606837089), this is easy test, should using release build, debug build sometimes not hang.

The issue existing long time, see issue #37293 , and then in commit #42239 this issue had fixed. but in commit #43285 broken again.  So I reference the implementation in #42239, then this code work. 

I MUST CLAIM, I really don't known what happend, and why this code work. But it worked. 

Release Notes:

- N/A


cole-miller:
Thanks!

PR #44213, opened 2025-12-05
lsp: Format RPC messages in language server logs

### LINKED ISSUE #44182: Zed's LSP requests in LSP logs aren't formatted
### Reproduction steps

1. Start Zed
2. Run `dev: open language server logs`

Current behavior: the LSP requests aren't formatted, which makes them harder to read.

Image

Expected behavior: the LSP requests are formatted (preferably with 2 spaces, to take less space)

Zed: v0.215.3-90+fda162bf3d9a11d18e07cbf2b488829450135d34 (Zed) 
OS: macOS 26.1
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

n/a

### Relevant Zed settings

n/a

_No response_

None

errmayank:
Closes #44182

Release Notes:

- Improved readability of RPC messages in language server logs

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Screenshot 2025-12-05 at 18 30 32</td>
    <td>Screenshot 2025-12-05 at 18 26 39</td>
  </tr>
</table>

### DIFF
--- a/crates/project/src/lsp_store/log_store.rs
+++ b/crates/project/src/lsp_store/log_store.rs
@@ -553,9 +553,13 @@ impl LogStore {
             rpc_log_lines.pop_front();
         }
 
+        let message = serde_json::from_str::<serde_json::Value>(message)
+            .and_then(|value| serde_json::to_string_pretty(&value))
+            .unwrap_or_else(|_| message.trim().to_owned());
+
         if store_logs {
             rpc_log_lines.push_back(RpcMessage {
-                message: message.trim().to_owned(),
+                message: message.clone(),
             });
         }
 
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
I'm not sure that we should merge things as is: one big reason for me personally is that selecting items does not work as smooth as before, where I could double click on a line.
I am sharing those snippets every now and then, so would like to keep this convenience in some form.

What's also wrong now are folds?
Before, those were made to show most of the line before the first forced line wrap (iirc 1024 is that?), but now this looks like a relatively random fold, sometimes somewhere in the middle of the new line — that does not look useful or good.

Besides all that, now we have to scroll more on average which is also not superb?

------------

All combined makes me think this should be redesigned better than just pretty-formatted, each tab in this "LSP editor" badly needs that, including the LSP logs.

Issue #43718, opened 2025-11-28
Downloading remote server WSL stuck forever when VPN is on


Banou26:
### Reproduction steps

Launch MullvadVPN, enable it
Start Zed, open WSL Project
Gets stuck on "Downloading remote server WSL" forever

### Current vs. Expected behavior

Zed WSL projects should work with VPNs

### Zed version and system specs

Zed: v0.214.6 (Zed) 
OS: Windows 10.0.26100
Memory: 95.3 GiB
Architecture: x86_64
GPU: AMD Radeon(TM) 890M Graphics || AMD Corporation || 25.11.1 (25.20.29.01-251106a-421109C-AMD-Software-Adrenalin-Edition)

### If you are using WSL on Windows, what flavor of Linux are you using?

Ubuntu

saidelike:
Does your remote host (you connect to) require a proxy? It seems similar to the problem I detailed in https://github.com/zed-industries/zed/issues/43694 except I hadn't detailed that test, which is to not define any proxy. Then it would hang forever when trying to connect while downloading what it needs to download (`node`, `codex-acp` or remote zed server etc. in my case iirc but in your case it is the remote server WSL).

potterwhite:
i think i had the same problem.
what was zed doing when i pressed open remote project?
vscode does not have this kind of issue

<img width="552" height="81" alt="Image" src="https://github.com/user-attachments/assets/3e62f6c1-cd7f-46fb-89b1-8721068ae183" />

saidelike:
@potterwhite does your remote host use an http proxy?

zelenenka:
@potterwhite are you also on Windows and using WSL?

banou26:
@saidelike I have nothing related to http/proxies set up in my zed settings configuration file.
Simply trying to launch Zed on a WSL folder.

PR #44216, opened 2025-12-05
remoting: Server download connect timeout

agu-z:
Sometimes machines are configured to drop outbound packets (rather than reject connections). In these cases, curl/wget just hang causing our download step to never complete. This PR adds a timeout of 10s for the connection (not the whole download), so that in situations like this we can fallback to our client-side download eventually.

Related to but doesn't fully fix: https://github.com/zed-industries/zed/issues/43694 and https://github.com/zed-industries/zed/issues/43718

Release Notes:

- remote: Add 10s connect timeout for server download

### DIFF
--- a/crates/remote/src/transport/ssh.rs
+++ b/crates/remote/src/transport/ssh.rs
@@ -706,9 +706,11 @@ impl SshRemoteConnection {
                         self.ssh_shell_kind,
                         "wget",
                         &[
-                            url,
                             "--connect-timeout",
                             CONNECT_TIMEOUT_SECS,
+                            "--tries",
+                            "1",
+                            url,
                             "-O",
                             &tmp_path_gz.display(self.path_style()),
                         ],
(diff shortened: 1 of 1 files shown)


crates/remote/src/transport/ssh.rs
```diff
@@ -701,7 +705,15 @@ impl SshRemoteConnection {
                     .run_command(
                         self.ssh_shell_kind,
                         "wget",
-                        &[url, "-O", &tmp_path_gz.display(self.path_style())],
+                        &[
+                            "--connect-timeout",
+                            CONNECT_TIMEOUT_SECS,
+                            "--tries",
+                            "1",
```
agu-z, line 712: By default, `wget` retries up to 20 times! Given we don't retry at all with `curl` (which we always use over `wget` when available), I think we can disable this here. I'd rather fall back to the client-side download as soon as possible.

Veykril (approved):

PR #44238, opened 2025-12-05
markdown_preview: Remove unnecessary vec allocation

aqrln:
Instead of allocating a one-element vec on the heap, we can just use an array here (since `Editor::edit` accepts anything that implements `IntoIterator`).

I haven't checked if there are more instances that can be simplified, just accidentally stumbled upon this when working on something else in the markdown preview crate.

Release Notes:

- N/A


MrSubidubi (approved):
Thanks!

PR #44250, opened 2025-12-05
Specify a schema to use when dealing with JSONC files

### LINKED ISSUE #40970: JSONC trailing comma warnings in bun.lock

SomeoneToIgnore:
Follow-up of https://github.com/zed-industries/zed/pull/43854
Closes https://github.com/zed-industries/zed/issues/40970

Seems that json language server does not distinguish between JSONC and JSON files in runtime, but there is a static schema, which accepts globs in its `fileMatch` fields.

Use all glob overrides and file suffixes for JSONC inside those match fields, and provide a grammar for such matches, which accepts trailing commas.

Release Notes:

- Improved JSONC trailing comma handling

### DIFF
--- a/crates/json_schema_store/src/json_schema_store.rs
+++ b/crates/json_schema_store/src/json_schema_store.rs
@@ -168,15 +168,26 @@ pub fn resolve_schema_request_inner(
     Ok(schema)
 }
 
-pub fn all_schema_file_associations(cx: &mut App) -> serde_json::Value {
-    let jsonc_globs = all_language_settings(None, cx)
+const JSONC_LANGUAGE_NAME: &str = "JSONC";
+
+pub fn all_schema_file_associations(
+    languages: &Arc<LanguageRegistry>,
+    cx: &mut App,
+) -> serde_json::Value {
+    let extension_globs = languages
+        .available_language_for_name(JSONC_LANGUAGE_NAME)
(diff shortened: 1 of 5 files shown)


Zakleby:
@SomeoneToIgnore
Hey after this update, when i use JSONC it adds trailing commas when i save. Below is an image
image

SomeoneToIgnore:
Yes, that's what an opinionated Prettier does when formatting jsonc files, so correct.
There's a whole entry in the docs for this: https://zed.dev/docs/languages/json?highlight=jsonc%20formatt#jsonc-prettier-formatting

PR #44259, opened 2025-12-06
gpui: Make length helpers into const functions

Serophots:
Make gpui's `rems()`, `phi()`, `auto()` length related helpers into const functions.

I can't see why these functions aren't already const except that it must've been overlooked when they were written?

In my project I had need for rems() to be const, and I thought I'd do phi() and auto() whilst I was in the neighbourhood

Release Notes:

- N/A


mikayla-maki:
Thanks!

Issue #42595, opened 2025-11-13
Python: pyrefly can only activate in one workspace in remote zed


Liyixin95:
### Summary

pyrefly can only activate in one workspace in remote zed.

### Description

Steps to reproduce:
1. open one workspace in remote zed, the pyrefly can start
2. open another workspace, then the pyrefly can not start for this one

### Zed Version and System Specs

Zed: v0.213.0 (Zed Preview) 
OS: Windows 10.0.26200
Memory: 31.6 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics || Intel Corporation || 32.0.101.6737

### If applicable, attach your `Zed.log` file to this issue.

<details><summary>Zed.log</summary>

```log

```

</details>

Liyixin95:
I think this was fix by #42723

PR #44262, opened 2025-12-06
Add builtin Pyrefly language server support

grievejia:
I'm continuing the work of @lolpack's PR https://github.com/zed-industries/zed/pull/40953. This makes it such that user no longer needs to install the external pyrefly extension to use pyrefly. It also auto-downloads Pyrefly binary when there's no existing installation, which the pyrefly extension currently can't do. 

We don't publish Pyrefly release artifacts to github (yet), so the easiest way to obtain a Pyrefly installation is with PyPI. Given that the PyLSP adapter is already doing a PyPI-based installation, I've essentially copy&pasted most of the logic from there into pyrefly adapter so Pyrefly could work in exactly the same way for now. 

There are a lot of code duplications here! My thought was to get the functionality in first, and I can send followup PRs to extract duplicated code into common helper functions. 

I tested the change in 2 scenarios:
- No `pyrefly` binary is found in `$PATH`
- There's an existing `pyrefly` binary in `$PATH` -- I obtained it with `uv tool install pyrefly`. 
In both scenarios I ran the test by removing the Pyrefly exteranl plugin, start Zed, open a `.py` file with type errors in it, and verify that cmd-click (gotodef) works. 

Console output for scenario 1:
```
> cargo run
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 31.62s
     Running `target/debug/zed`
2025-12-04T20:58:07-08:00 INFO  [zed] ========== starting zed version 0.217.0+dev.65f328bd614282c1c370672106a9a279f8f22a3a, sha 65f328b ==========
2025-12-04T20:58:08-08:00 WARN  [zed::reliability] Minidump endpoint not set
2025-12-04T20:58:08-08:00 INFO  [extension_host] extensions updated. loading 2, reloading 0, unloading 0
2025-12-04T20:58:08-08:00 WARN  [audio::audio_settings] Denoise does not work on debug builds, not enabling
2025-12-04T20:58:08-08:00 INFO  [workspace] Rendered first frame
2025-12-04T20:58:09-08:00 WARN  [audio::audio_settings] Denoise does not work on debug builds, not enabling
2025-12-04T20:58:16-08:00 INFO  [lsp] starting language server process. binary path: "/Users/grievejia/Library/Application Support/Zed/languages/ruff/ruff-0.14.8/ruff-aarch64-apple-darwin/ruff", working directory: "/Users/grievejia/scratch/pyre_test", args: ["server"]
2025-12-04T20:58:20-08:00 INFO  [lsp] starting language server process. binary path: "/Users/grievejia/Library/Application Support/Zed/languages/pyrefly/pyrefly-venv/bin/pyrefly", working directory: "/Users/grievejia/scratch/pyre_test", args: ["lsp"]
...
```
We could see that the venv installation is successful and the auto-installed Pyrefly binary is used.

Console output for scenario 2:
```
> cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 29.62s
     Running `target/debug/zed`
2025-12-04T21:01:26-08:00 INFO  [zed] ========== starting zed version 0.217.0+dev.65f328bd614282c1c370672106a9a279f8f22a3a, sha 65f328b ==========
2025-12-04T21:01:27-08:00 WARN  [zed::reliability] Minidump endpoint not set
2025-12-04T21:01:27-08:00 INFO  [extension_host] extensions updated. loading 2, reloading 0, unloading 0
2025-12-04T21:01:27-08:00 WARN  [audio::audio_settings] Denoise does not work on debug builds, not enabling
2025-12-04T21:01:27-08:00 INFO  [workspace] Rendered first frame
2025-12-04T21:01:28-08:00 WARN  [audio::audio_settings] Denoise does not work on debug builds, not enabling
2025-12-04T21:01:34-08:00 INFO  [language] found user-installed language server for pyrefly. path: "/Users/grievejia/.local/bin/pyrefly", arguments: ["lsp"]
2025-12-04T21:01:34-08:00 INFO  [lsp] starting language server process. binary path: "/Users/grievejia/.local/bin/pyrefly", working directory: "/Users/grievejia/scratch/pyre_test", args: ["lsp"]
```
We could see that the correct installed Pyrefly binary was picked up. 

Release Notes:
- Added Pyrefly as builtin language server option


grievejia:
@cla-bot check

grievejia:
Updated release note

lolpack:
Hi @ConradIrwin, Jia took over this change for me. Is there a good person on the Zed side to review it?

osiewicz (commented):
Just my 2 cents on a first skim:
- You should probably implement `workspace_configuration` from `LspAdapter` trait, as that way you'll be able to propagate the toolchain that the user is using into pyrefly's configuration. You'd probably need to fill out `python-interpreter-path` (or whatever the setting is for pyrefly) with the provided `toolchain` value.
- I am a bit uneasy about installing pyrefly via PyPI. Yes, we went that route with pylsp because there was no other option; we've received quite a few bug reports to that effect ever since that's been rolled out (a year or so ago). Not that this is a show-stopper, but I'd mildly recommend considering installing the binary directly.

SomeoneToIgnore:
>  This makes it such that user no longer needs to install the external pyrefly extension to use pyrefly.

I seem to miss something — why is having an extension bad?

carlesoctav:
does having a builtin support, resolve this issue #44511, #42595?

grievejia:
Thanks for the feedbacks! 

> You should probably implement workspace_configuration from LspAdapter trait

Implemented. Thanks for pointing this out!

> I am a bit uneasy about installing pyrefly via PyPI.

I share your concerns. My worry is that installing via PyPI could lock us into a specific version of Pyrefly, making upgrades more challenging down the line.

At the moment, though, we don’t have a clear alternative. Would you prefer that we explore other installation options before merging this diff?

> I seem to miss something — why is having an extension bad?

From our perspective, the main issue is onboarding friction. With an extension, users need to complete two manual steps to enable Pyrefly:
- Install the extension
- Edit their config file

By removing the first step, we simplify onboarding and make our process more competitive with tools like Pyright and Ty.
There may also be additional motivations related to remote editing, but I’m less familiar with those details.

crates/languages/src/python.rs
```diff
@@ -2364,6 +2364,262 @@ impl LspInstaller for RuffLspAdapter {
...
+                && let Ok(env) = serde_json::from_value::<PythonToolchainData>(toolchain.as_json)
+            {
+                if !user_settings.is_object() {
+                    user_settings = Value::Object(serde_json::Map::default());
+                }
+                let object = user_settings.as_object_mut().unwrap();
+
+                if let Some(interpreter_path) = &env.environment.executable {
+                    object.insert(
+                        "pythonPath".to_owned(),
```
osiewicz: Is `pythonPath` a way to specify an active virtual environment with `pyrefly`? I can't find it in your docs (https://pyrefly.org/en/docs/configuration/)

grievejia: Yeah I think we derive all runtime metadata from an interpreter path at the moment. As for which config to use I had to look at the [source code](https://github.com/facebook/pyrefly/blob/7b287652771e3c427599fafa3a50c8530f978059/pyrefly/lib/lsp/non_wasm/workspace.rs#L277-L287) to figure it out. We could definitely have better doc for it.

PR #44273, opened 2025-12-06
gpui: Take advantage of unified memory on Apple silicon

marcocondrache:
Metal chooses a buffer’s default storage mode based on the type of GPU in use.
On Apple GPUs, the default mode is shared, which allows the CPU and GPU to access the same memory without requiring explicit synchronization.
On discrete or external GPUs, Metal instead defaults to managed storage, which does require explicit CPU–GPU memory synchronization.

This change aligns our buffer usage with Metal’s default behavior and avoids unnecessary synchronization on Apple-silicon Macs. As a result, memory usage on Apple hardware is reduced and performance improves due to fewer sync operations.

Ref: https://developer.apple.com/documentation/metal/setting-resource-storage-modes
Ref: https://developer.apple.com/documentation/metal/synchronizing-a-managed-resource-in-macos

With the storage mode:

image

On main branch:

image

That's a 44% reduction of memory usage.

Release Notes:

- Reduced memory usage on Apple-silicon Macs by using shared memory where appropriate

### DIFF
--- a/crates/gpui/src/platform/mac/metal_atlas.rs
+++ b/crates/gpui/src/platform/mac/metal_atlas.rs
@@ -15,6 +15,8 @@ pub(crate) struct MetalAtlas(Mutex<MetalAtlasState>);
 impl MetalAtlas {
     pub(crate) fn new(device: Device) -> Self {
         MetalAtlas(Mutex::new(MetalAtlasState {
+            // Shared memory can be used only if CPU and GPU share the same memory space.
+            // https://developer.apple.com/documentation/metal/setting-resource-storage-modes
             unified_memory: device.has_unified_memory(),
             device: AssertSend(device),
             monochrome_textures: Default::default(),
--- a/crates/gpui/src/platform/mac/metal_atlas.rs
+++ b/crates/gpui/src/platform/mac/metal_atlas.rs
@@ -15,6 +15,7 @@ pub(crate) struct MetalAtlas(Mutex<MetalAtlasState>);
 impl MetalAtlas {
     pub(crate) fn new(device: Device) -> Self {
         MetalAtlas(Mutex::new(MetalAtlasState {
+            unified_memory: device.has_unified_memory(),
             device: AssertSend(device),
             monochrome_textures: Default::default(),
             polychrome_textures: Default::default(),
@@ -29,6 +30,7 @@ impl MetalAtlas {
 
 struct MetalAtlasState {
     device: AssertSend<Device>,
+    unified_memory: bool,
     monochrome_textures: AtlasTextureList<MetalAtlasTexture>,
     polychrome_textures: AtlasTextureList<MetalAtlasTexture>,
     tiles_by_key: FxHashMap<AtlasKey, AtlasTile>,
@@ -146,6 +148,11 @@ impl MetalAtlasState {
         }
         texture_descriptor.set_pixel_format(pixel_format);
         texture_descriptor.set_usage(usage);
+        texture_descriptor.set_storage_mode(if self.unified_memory {
+            metal::MTLStorageMode::Shared
+        } else {
+            metal::MTLStorageMode::Managed
+        });
         let metal_texture = self.device.new_texture(&texture_descriptor);
 
         let texture_list = match kind {
--- a/crates/gpui/src/platform/mac/metal_renderer.rs
+++ b/crates/gpui/src/platform/mac/metal_renderer.rs
@@ -76,9 +76,13 @@ impl InstanceBufferPool {
         self.buffers.clear();
     }
 
-    pub(crate) fn acquire(&mut self, device: &metal::Device) -> InstanceBuffer {
+    pub(crate) fn acquire(
+        &mut self,
+        device: &metal::Device,
+        unified_memory: bool,
+    ) -> InstanceBuffer {
         let buffer = self.buffers.pop().unwrap_or_else(|| {
-            let storage_mode = if device.has_unified_memory() {
+            let storage_mode = if unified_memory {
                 MTLResourceOptions::StorageModeShared
             } else {
                 MTLResourceOptions::StorageModeManaged
@@ -102,6 +106,7 @@ impl InstanceBufferPool {
 pub(crate) struct MetalRenderer {
     device: metal::Device,
     layer: metal::MetalLayer,
+    unified_memory: bool,
     presents_with_transaction: bool,
     command_queue: CommandQueue,
     paths_rasterization_pipeline_state: metal::RenderPipelineState,
@@ -182,11 +187,8 @@ impl MetalRenderer {
             output
         }
 
-        let storage_mode = if device.has_unified_memory() {
-            MTLResourceOptions::StorageModeShared
-        } else {
-            MTLResourceOptions::StorageModeManaged
-        };
+        let unified_memory = device.has_unified_memory();
+
         let unit_vertices = [
             to_float2_bits(point(0., 0.)),
             to_float2_bits(point(1., 0.)),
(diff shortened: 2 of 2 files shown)


marcocondrache:
Additionally, we render MSAA textures in single passes. This allows us to use memory-less storage on Apple Silicon, eliminating the need for backing memory. For a 4K display with 4x MSAA, switching to memory-less storage reduces the system memory cost from 127 MB to effectively 0 bytes.

Ref: https://developer.apple.com/documentation/metal/choosing-a-resource-storage-mode-for-apple-gpus

Using the painting example:

On main branch:

image

With storage mode:

image

crates/gpui/src/platform/mac/metal_atlas.rs
```diff
@@ -15,6 +15,7 @@ pub(crate) struct MetalAtlas(Mutex<MetalAtlasState>);
 impl MetalAtlas {
     pub(crate) fn new(device: Device) -> Self {
         MetalAtlas(Mutex::new(MetalAtlasState {
+            unified_memory: device.has_unified_memory(),
```
Anthony-Eid, line 20: Could you comment the apple doc references here please? Seems like a good thing for developers to be able to get context on

marcocondrache: Yes that's right! done

crates/gpui/src/platform/mac/metal_renderer.rs
```diff
@@ -76,12 +76,21 @@ impl InstanceBufferPool {
...
+    ) -> InstanceBuffer {
         let buffer = self.buffers.pop().unwrap_or_else(|| {
-            device.new_buffer(
-                self.buffer_size as u64,
-                MTLResourceOptions::StorageModeManaged,
-            )
+            let options = if unified_memory {
+                MTLResourceOptions::StorageModeShared
+                    // Buffers are write only which can benefit from the combined cache
+                    | MTLResourceOptions::CPUCacheModeWriteCombined
```
marcocondrache, line 89: https://developer.apple.com/documentation/metal/mtlresourceoptions/cpucachemodewritecombined

Since we only write in our buffers and never read we can combine writes into a single transition

crates/gpui/src/platform/mac/metal_renderer.rs
```diff
@@ -190,7 +204,11 @@ impl MetalRenderer {
         let unit_vertices = device.new_buffer_with_data(
             unit_vertices.as_ptr() as *const c_void,
             mem::size_of_val(&unit_vertices) as u64,
-            MTLResourceOptions::StorageModeManaged,
+            if unified_memory {
+                MTLResourceOptions::StorageModeShared
+            } else {
+                MTLResourceOptions::StorageModeManaged
+            },
```
Anthony-Eid, line 213: If unified_memory is true shouldn't `MTLResourceOptions::CPUCacheModeWriteCombined` be set as an option as well? It seems like this buffer is also write only, unless I'm misunderstanding something

crates/gpui/src/platform/mac/metal_renderer.rs
Anthony-Eid: Maybe having a link to the ref here would be good as well.

crates/gpui/src/platform/mac/metal_renderer.rs
```diff
@@ -337,14 +356,21 @@ impl MetalRenderer {
...
         texture_descriptor.set_height(size.height.0 as u64);
         texture_descriptor.set_pixel_format(metal::MTLPixelFormat::BGRA8Unorm);
+        texture_descriptor.set_storage_mode(metal::MTLStorageMode::Private);
         texture_descriptor
             .set_usage(metal::MTLTextureUsage::RenderTarget | metal::MTLTextureUsage::ShaderRead);
         self.path_intermediate_texture = Some(self.device.new_texture(&texture_descriptor));
 
         if self.path_sample_count > 1 {
+            let storage_mode = if self.unified_memory {
+                metal::MTLStorageMode::Memoryless
```
Anthony-Eid, line 370:
```suggestion
            // https://developer.apple.com/documentation/metal/choosing-a-resource-storage-mode-for-apple-gpus
            // Rendering MSAA textures are done in a single pass, so we can use memory-less storage on Apple Silicon
            let storage_mode = if self.unified_memory {
                metal::MTLStorageMode::Memoryless
```

Anthony-Eid (changes requested):
Thank you for yet another great gpui PR! I left some light comments and will be ready to merge after they're addressed.

crates/gpui/src/platform/mac/metal_renderer.rs
marcocondrache: Done! Thanks

crates/gpui/src/platform/mac/metal_renderer.rs
marcocondrache: @Anthony-Eid You’re right that this buffer is write-only. In this case, though, the buffer is created with the data already populated and isn’t accessed again if not for reading, so I’m not sure it would benefit from the write cache here. I don’t think adding it would cause any harm tho

crates/gpui/src/platform/mac/metal_renderer.rs
marcocondrache: Yup done!

crates/gpui/src/platform/mac/metal_renderer.rs
marcocondrache: I’ve added it - thanks for pointing it out!

Anthony-Eid: Thanks for responding so fast! I'm still new to gpu rendering so I was just asking for clarification

Anthony-Eid (approved):

crates/gpui/src/platform/mac/metal_renderer.rs
```diff
@@ -179,6 +190,10 @@ impl MetalRenderer {
             output
         }
 
+        // Shared memory can be used only if CPU and GPU share the same memory space.
+        // https://developer.apple.com/documentation/metal/setting-resource-storage-modes
+        let unified_memory = device.has_unified_memory();
```
rtfeldman, line 195: It looks like this `has_unified_memory` function has a bug affecting Intel macs - I've opened an [upstream PR](https://github.com/gfx-rs/metal-rs/pull/374) for it.

marcocondrache:
@rtfeldman @JosephTLyons @notpeter thank you so much guys for taking care of this. I did not expect at all to find a bug in `metal-rs`. Sorry for the trouble.

JosephTLyons:
> @rtfeldman @JosephTLyons @notpeter thank you so much guys for taking care of this. I did not expect at all to find a bug in `metal-rs`. Sorry for the trouble.

No problem, we still want this PR, just have to wait for the fix upstream!

crates/gpui/src/platform/mac/metal_renderer.rs
rtfeldman: ok, upstream PR has been landed and released in `metal` v 0.33, so we can un-revert this PR once we've upgraded to metal 0.33!

marcocondrache: @rtfeldman Is someone already handling the PR with the upgrade, or should I prepare it?

filipwiech: @marcocondrache Sorry, haven't noticed that there's already an ongoing discussion here when I pinged you back in the PR with the revert (#45022). 😅 Do you plan to re-submit your really nice contribution now that the required fix is ready? I think you could maybe even bundle the `metal-rs` upgrade together with your changes if that's more convenient. In any case, thank you for all the improvements and have a very Happy Holidays! 🙂

marcocondrache: @filipwiech, thank you. I've resubmitted the PR and included the upgrade. In the meantime, I will analyze whether upgrading metal could cause any issues

PR #44276, opened 2025-12-06
Enable unsaved buffer persistence for worktree-less windows

### LINKED ISSUE #15098: Serialize the unsaved changes of files not associated with a worktrees
### Check for existing issues

- [X] Completed

### Check for existing issues

- [X] Completed

### Describe the feature

Since the following PRs, Zed now restores unsaved content:

- https://github.com/zed-industries/zed/pull/13546
- https://github.com/zed-industries/zed/pull/14965

There is a limitation currently. Zed will serialize the changes for files if there are worktrees / projects in the project panel, but it will not do so for instances of Zed that have no worktrees / projects in the project panel - supporting this case would be a nice addition to the workspace serialization story.

_No response_

_No response_

luukverhoeven:
Closes #15098.

Previously, unsaved buffers in empty windows (no project open) or single file editing sessions were not persisted across restarts. This change enables the hot-exit functionality for these scenarios by:

- Always serializing dirty buffers regardless of worktree presence
- Including empty workspaces in session serialization
- Removing the DetachFromSession behavior for local empty windows

Release Notes:
 - Added support for restoring unsaved buffers in empty windows and single-file editing sessions
  ([#15098](https://github.com/zed-industries/zed/issues/15098))

### DIFF
--- a/crates/editor/src/items.rs
+++ b/crates/editor/src/items.rs
@@ -1270,9 +1270,9 @@ impl SerializableItem for Editor {
         let project = self.project.clone()?;
 
         let serialize_dirty_buffers = match buffer_serialization {
-            // If we don't have a worktree, we don't serialize, because
-            // projects without worktrees aren't deserialized.
-            BufferSerialization::All => project.read(cx).visible_worktrees(cx).next().is_some(),
+            // Always serialize dirty buffers, including for worktree-less windows.
+            // This enables hot-exit functionality for empty windows and single files.
+            BufferSerialization::All => true,
             BufferSerialization::NonDirtyBuffers => false,
         };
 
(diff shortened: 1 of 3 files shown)


nathansobo:
I pushed a fix for the dead code issue.

You'll need to sign the CLA at https://zed.dev/cla and then comment `@cla-bot check`.

---

Regarding the implementation itself, I have a question about how empty workspaces are identified.

Currently, workspaces are looked up by their paths:

```rust
WHERE paths IS ? AND remote_connection_id IS ?
```

With this change, empty workspaces would serialize with `paths = ""`. But this creates a problem:

- If I open **two** empty windows, each with different unsaved buffers, they both have the same "key" (empty paths, no remote connection)
- When saving, the query `DELETE FROM workspaces WHERE workspace_id != ?1 AND paths IS ?2` would delete the *other* empty workspace
- When restoring via `workspace_for_roots(&[])`, we'd just get whichever one comes back first (`LIMIT 1`)

So I think we need to figure out the intended behavior:

1. Should there only ever be **one** empty workspace that accumulates all unsaved buffers from all empty windows?
2. Or should multiple empty windows each persist their own state? If so, we'd need a different identification mechanism (maybe a UUID stored with the window?)

What do you think?

PR #44284, opened 2025-12-06
Fix Visual mode flip on project search 

### LINKED ISSUE #43878: Vim mode: project search results are in visual mode on first run
### Reproduction steps

When performing a deploy search in Vim mode the multi buffer is rendered with the visual mode active.
After doing the search the second time, it is in normal mode (as expected).
Coming back from the project search results to the file where you were before makes it switch to visual mode again. This happens using the `pane: alternate file` command

The multibuffer panel should open in normal mode from the first time and stay in it.

Zed: v0.215.3-86+74f281da864725a4e21360852e2d6a62bc4b4941 (Zed Preview) 
OS: macOS 15.7.2
Memory: 64 GiB
Architecture: aarch64

MostlyKIGuess:
- Prevent incorrect visual-mode activation when opening project search results by ensuring selections made before the editor gains focus

### Some Diagnosis:
- The results editor got created before collapse_matches was initialized -> giving  range_for_match a full selection instead of a collapsed cursor. 
- When the editor finally focused and Vim synced settings, it saw that non-empty selection -> flip into Visual mode.
- From what I understand, it happened only on the first run because: the editor is _not_ focused yet when created, so `sync_vim_settings` is _not_ called. This means `collapse_matches` is still `false` (the default). ( so when called from [project_search.rs](crates/search/src/project_search.rs) L1515 collapse matches is false and later on it focuses on the editor and due to that it changes to visual. )

Closes #43878 

Release Notes:

- Fixed vim mode incorrectly switching to Visual mode on first project search


MostlyKIGuess:
Intended Behavior [demo](https://github.com/user-attachments/assets/7677bdaa-e5df-4b47-82a0-cda73e426c6f)

ovidius72:
> Intended Behavior [demo](https://github.com/user-attachments/assets/7677bdaa-e5df-4b47-82a0-cda73e426c6f)

Looks great. Did you do some tests when you go back to the file where the search started from? I found it also to  be in visual mode.

MostlyKIGuess:
> > Intended Behavior [demo](https://github.com/user-attachments/assets/7677bdaa-e5df-4b47-82a0-cda73e426c6f)
> 
> Looks great. Did you do some tests when you go back to the file where the search started from? I found it also to be in visual mode.

Yeah, it is normal for me, can you share a video? 

I tried going back via ctrl tab and g space as well.. seems to be fine

MostlyKIGuess:
I meant in my version, it stays in normal fixing the bug.

ovidius72:
> I meant in my version, it stays in normal fixing the bug.

Thanks. Great job.  I'm going to check it as soon as it will be released and, eventually, i'll send you a video.

dinocosta (approved):
@MostlyKIGuess Thank you so much for tackling this! Appreciate the detailed diagnosis in the PR's description, helps frame the suggested changes ✌️

I noticed that, since we're updating `Vim::activate` to set the `collapse_matches` field, we might as well try to actually sync all settings, like `Vim.sync_settings` is doing, so I'm wrapping up a commit with these changes.

Will merge soon! 🙂

PR #44293, opened 2025-12-06
editor: Fix open_url command not handling non-channel server URLs

### LINKED ISSUE #43833: `editor: open url` does not work with domain `zed.dev`
### Reproduction steps

1. open any file in zed.
2. execute the action: `editor: open url` in a url which domain is `zed.dev`

the action: `editor: open url` can open urls that domain is `zed.dev`

Zed: v0.214.7 (Zed) 
OS: macOS 26.1
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

```log
no log generated
```

</details>

### Relevant Zed settings

```json
{
  "ui_font_family": "JetBrainsMono Nerd Font",
  "buffer_font_family": "JetBrainsMono Nerd Font",
  "icon_theme": "Catppuccin Mocha",
  "vim_mode": true,
  "ui_font_size": 16,
  "buffer_font_size": 15,
  "theme": {
    "mode": "system",
    "light": "Catppuccin Mocha",
    "dark": "One Dark"
  }
}

```

</details>

_No response_

None

NimitzDEV:
Closes #43833

Fix open_url to handle all server URLs, not just channel links

Previously, URLs like zed.dev/docs or zed.dev/extensions, or any URL that matches the server_url setting but does not link to a collab channel, were rejected instead of being opened in the browser because parse_request_path threw errors when links were not channel links.

Release Notes:

- editor: Fix open_url command not handling non-channel server URLs


NimitzDEV:
@cla-bot check

ConradIrwin:
@NimitzDEV Thanks for this! We shouldn't add a regex, but we should update the code so that if we fail to parse it using the existing functions we fallback to opening the browser.

That might require shuffling some code around.

NimitzDEV:
@ConradIrwin Hi, Thanks for the feedback! I updated the fix to avoid using regex as suggested. `OpenRequest::parse` now handles the fallback to the user's browser for unhandled URLs. Let me know if this is the better way to implement the fix.

Also, while working on this, I noticed a potential issue in the `parse ` function. It can take multiple URLs but only last one can be handled, because `this.kind` gets overwritten in each iteration. Not sure if this is intended.

ConradIrwin:
Thanks for taking the time to make this better!

As I was reviewing this change, I realized that we didn't want to run the risk of overloading the open url request (because then we might end up redirecting back and forth between the web and the editor), so I spent a bit of time playing, and ended up with the change in https://github.com/zed-industries/zed/pull/44910 instead.

PR #44297, opened 2025-12-06
agent_ui: Implement favorite models selection

### LINKED ISSUE #31507: Agent Panel: Customize options included in model selector
The model selector currently displays all available models with a hardcoded recommended section. This is not the most convenient UX for users that prefer local models or who have different preferences, so it'd be nice to be to pin preferred models or exclude some altogether.

Image

aqrln:
This PR solves my main pain point with Zed agent: I have a long list of available models from different providers, and I switch between a few of them depending on the context and the project. In particular, I use the same models from different providers depending on whether I'm working on a personal project or at my day job. Since I only care about a few models (none of which are in "recommended") that are scattered all over the list, switching between them is bothersome, even using search.

This change adds a new option in `settings.json` (`agent.favorite_models`) and the UI to manipulate it directly from the list of available models. When any models are marked as favorites, they appear in a dedicated section at the very top of the list. Each model has a small icon button that appears on hover and allows to toggle whether it's marked as favorite.

I implemented this on the UI level (i.e. there's no first-party knowledge about favorite models in the agent itself; in theory it could return favorite models as a group but it would make it harder to implement bespoke UI for the favorite models section and it also wouldn't work for text threads which don't use the ACP infrastructure).

The feature is only enabled for the native agent but disabled for external agents because we can't easily map their model IDs to settings and there could be weird collisions between them.

https://github.com/user-attachments/assets/cf23afe4-3883-45cb-9906-f55de3ea2a97

Closes https://github.com/zed-industries/zed/issues/31507

Release Notes:

- Added the ability to mark language models as favorites and pin them to the top of the list. This feature is available in the native Zed agent (including text threads and the inline assistant), but not in external agents via ACP.


crates/agent_ui/src/language_model_selector.rs
```diff
@@ -495,6 +583,30 @@ impl PickerDelegate for LanguageModelPickerDelegate {
...
+                                .agent
+                                .get_or_insert_default()
+                                .add_favorite_model(model),
+                            LanguageModelPickerEntryAction::Unfavorite
+                            | LanguageModelPickerEntryAction::RemoveFromFavorites => settings
+                                .agent
+                                .get_or_insert_default()
+                                .remove_favorite_model(&model),
+                        });
+                    }
```
aqrln: This should be moved to a callback similar to `on_model_changed`, I forgot to change this, will do tonight

aqrln:
@danilo-leal I've revisited and updated the implementation a little bit for better separation of concerns

danilo-leal (approved):
Thank you so much; great feature to have :)

aqrln:
Thank you for the review and improvements @danilo-leal and @bennetbo!

PR #35127, opened 2025-07-26
docs: Combine C and C++ doc to eliminate duplications

hokein:
The c and cpp documentation have a lot of duplicated content, this change combines them into a single doc.

Release Notes:

- N/A

PR #44308, opened 2025-12-06
docs: Document implemented `clangd` extensions

hokein:
Zed currently doesn’t support all protocol extensions implemented by `clangd`, but it does support two:

- `textDocument/inactiveRegion`
- `textDocument/switchSourceHeader`

Release Notes:

- N/A

### DIFF
--- a/docs/src/languages/cpp.md
+++ b/docs/src/languages/cpp.md
@@ -156,3 +156,15 @@ You can use CodeLLDB or GDB to debug native binaries. (Make sure that your build
   }
 ]
 ```
+
+## Extensions
+
+Zed implements several clangd [extensions](https://clangd.llvm.org/extensions) to enhance the support for C/C++ development.
+
+### Switch Between Source and Header File
+
+You can instantly jump between a source file and its corresponding header file using a dedicated command.
+
+### Inactive Code Highlighting
+
+Zed visually dims code that is inactive due to preprocessor directives (such as code within an `#if 0` block). This makes it easy to distinguish between active and compiled-out code, helping you better understand the current compilation state of your files.
--- a/docs/src/languages/cpp.md
+++ b/docs/src/languages/cpp.md
@@ -157,14 +157,25 @@ You can use CodeLLDB or GDB to debug native binaries. (Make sure that your build
 ]
 ```
 
-## Extensions
+## Protocol Extensions
 
-Zed implements several `clangd` [extensions](https://clangd.llvm.org/extensions) to enhance the support for C/C++ development.
+Zed currently implements the following `clangd` [extensions](https://clangd.llvm.org/extensions):
 
-### Switch Between Source and Header File
+### Inactive Regions
 
-You can instantly jump between a source file and its corresponding header file using a dedicated command.
+Automatically dims inactive sections of code due to preprocessor directives, such as `#if`, `#ifdef`, or `#ifndef` blocks that evaluate to false.
 
-### Inactive Code Highlighting
+### Switch Between Source and Header Files
 
-Zed visually dims code that is inactive due to preprocessor directives (such as code within an `#if 0` block). This makes it easy to distinguish between active and compiled-out code, helping you better understand the current compilation state of your files.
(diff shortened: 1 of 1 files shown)


docs/src/languages/cpp.md
```diff
@@ -158,3 +156,15 @@ You can use CodeLLDB or GDB to debug native binaries. (Make sure that your build
   }
 ]
 ```
+
+## Extensions
+
+Zed implements several clangd [extensions](https://clangd.llvm.org/extensions) to enhance the support for C/C++ development.
```
yeskunall:
```suggestion
Zed implements several `clangd` [extensions](https://clangd.llvm.org/extensions) to enhance the support for C++ development.
```

yeskunall:
Looks like `clangd` implements quite a few [protocol extensions](https://clangd.llvm.org/extensions#contents). I haven’t confirmed it myself, but Zed should support all of these since it just uses `clangd` under the hood. Are you saying Zed does not support any of the other extensions?

hokein:
> Looks like `clangd` implements quite a few [protocol extensions](https://clangd.llvm.org/extensions#contents). I haven’t confirmed it myself, but Zed should support all of these since it just uses `clangd` under the hood. Are you saying Zed does not support any of the other extensions?

While clangd implements these extensions, they don't work out of the box -- the client (Zed) need to explicitly implement support for them. I verified the codebase, and currently, only these two extensions are supported:

- [crates/editor/src/clangd_ext.rs](https://github.com/zed-industries/zed/blob/main/crates/editor/src/clangd_ext.rs)
- [crates/project/src/lsp_store/clangd_ext.rs](https://github.com/zed-industries/zed/blob/main/crates/project/src/lsp_store/clangd_ext.rs)

docs/src/languages/cpp.md
hokein: Technically they work for C too, but I get that we have a separate `c.md`. (Side note: I still think merging the two docs would be better to avoid duplication #35127).

yeskunall: No, the two languages differ greatly, even if there are similarities in certain areas.

yeskunall:
@hokein I can’t get `textDocument/inactiveRegions` to fire in a C++ file at any cost. Here’s a snippet that I think that should “dim” code, but right now it doesn’t:

```c
int main() {
    int x = 1;
    
#if 0
    // This block is ALWAYS inactive
    x = 999;
#endif

    return x;
}

```

Do you have a minimal reproduction or a screenshot?

hokein:
> @hokein I can’t get `textDocument/inactiveRegions` to fire in a C++ file at any cost. Here’s a snippet that I think that should “dim” code, but right now it doesn’t:
> 
> ```c
> int main() {
>     int x = 1;
>     
> #if 0
>     // This block is ALWAYS inactive
>     x = 999;
> #endif
> 
>     return x;
> }
> ```
> 
> Do you have a minimal reproduction or a screenshot?

Which clangd version are you using? It requires clangd v21.1.0 or v22. It should work on your example. Here is my screenshot (with v21.1.0):

image

docs/src/languages/cpp.md
hokein: While I agree the languages differ in many ways, their support in Zed is practically identical:

- Shared tooling: Both languages are powered by `clangd`
- Features Parity: e.g. these extensions works exactly the same for both C and C++.
- The setup/configuration process applies on both languages

Since these docs focus on language server setup and editor features rather than language syntax, merging them could help avoid duplication and keep the instructions consistent for all C-family developers.

yeskunall:
@hokein got it. I’m using the Homebrew version `clang@21` as well. I had to set `unnecessary_code_fade` to a higher value as my theme did not do a good job of showcasing unused symbols for C++ files.

Thanks for suggesting the docs changes -- I’m making some edits and then we’re good to go.

hokein:
Thanks, the latest version looks good to me.

yeskunall:
Thanks for surfacing this info in our docs @hokein! 💖

PR #44313, opened 2025-12-06
theme: Add `element_enabled` property

karearl:
Closes #44181

Release Notes:

- Add `element.enabled` theme tag

image
image


karearl:
@cla-bot check

danilo-leal:
Hey, thanks for the PR! I get why this PR could be seen as a solution to the issue but I don't think adding this theme token right now to assign it to this one element is _it_ yet. I do think it probably makes sense to have a single enable/accent color that's generally used throughout components, though. We have been revising some aspects of the theme structure and hope to update it soon, including a better way to communicate that a change happened in the theme to theme builders. I'll close this in the meantime, but thank you!

PR #44315, opened 2025-12-07
editor: Add action to run rust-analyzer flycheck on the current file

xdBronch:
by default it would always send `"textDocument": null`, if your rust-analyzer config disabled workspace checking this caused flycheck to do nothing

Release Notes:

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -667,6 +667,8 @@ actions!(
         Rewrap,
         /// Runs flycheck diagnostics.
         RunFlycheck,
+        /// Runs flycheck diagnostics only on the current files workspace.
+        RunFlycheckCurrentFile,
         /// Scrolls the cursor to the bottom of the viewport.
         ScrollCursorBottom,
         /// Scrolls the cursor to the center of the viewport.
--- a/crates/editor/src/rust_analyzer_ext.rs
+++ b/crates/editor/src/rust_analyzer_ext.rs
@@ -17,8 +17,8 @@ use text::ToPointUtf16;
 
 use crate::{
     CancelFlycheck, ClearFlycheck, Editor, ExpandMacroRecursively, GoToParentModule,
-    GotoDefinitionKind, OpenDocs, RunFlycheck, element::register_action, hover_links::HoverLink,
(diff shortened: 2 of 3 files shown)


SomeoneToIgnore (commented):
Thank you.

Let's

1. Make a paremeterized action (with default to run on the whole workspace)
 
2. And detect the 
>  if your rust-analyzer config disabled workspace checking

automatically?

Presumably, you're talking about `initialization_option` configuration and it's possible to get one as https://github.com/zed-industries/zed/blob/9da0d406941f6ac65c25bff4ff4305aa4f60a987/crates/project/src/lsp_store.rs#L467-L472 ?

This is a project's LspStore level, the exact place we send the r-a ext request from, so we can access this information there?

xdBronch:
by parameterized action do you mean actions like `workspace::SendKeystrokes` and `task::Spawn` where options can be given in the keymap? i would prefer being able to run these actions from the command palette.

im not sure i love the idea of reading the config of lsps like that, ik this isnt much of an argument but it just feels wrong. im also struggling a bit to figure out how it should be done, i need the lsp_store but thats currently only in the local branch, is it ok to `lsp_store.read_with` in the remote branch? i dont understand a lot of this code

SomeoneToIgnore:
> but it just feels wrong

If I understand the workflow right, first a user adds a language server configuration into Zed, and then it feels wrong for Zed to use this configuration to determine what to do?

To me, it is wrong to use a bad command given the configuration knowledge?
Or do I misunderstand something?

> i dont understand a lot of this code

I am not able to give you exact pointers, but hopefully this helps: in a handler of `server.notify::<lsp_store::lsp_ext_command::LspExtRunFlycheck>`, `lsp_store` and `rust_analyzer_server` are available — hence, the adapter can be retrieved for `rust_analyzer_server` with the `lsp_store.language_server_adapter_for_id` — then, `prepare_initialize_params` or `initialization_options_for_adapter` or whatever related method could be called on adapter or server to get the actual json.

xdBronch:
sorry, i might not get to this for a bit. im having trouble understanding a lot of the code and rust-analyzer uses so much of my ram i have to stop it in order to check/test my changes :p, its a very slow and frustrating dev cycle

PR #44333, opened 2025-12-07
git: Add settings for detecting code movement in git blame

olejorgenb:
Closes #42581 (feature request discussion)

Release Notes:

- Added settings for enabling -M and -C when running git blame

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1335,7 +1335,23 @@
       "min_column": 0
     },
     "blame": {
-      "show_avatar": true
+      "show_avatar": true,
+      /// Whether to detect code movement within a file in git blame.
+      /// This enables the -M flag for git blame.
+      ///
+      /// Note: Enabling this setting may make git blame operations slower
+      /// for large repositories as it requires additional analysis.
+      ///
+      /// Default: false
+      "detect_code_movement": false,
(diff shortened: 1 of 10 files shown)


olejorgenb:
Tempting to just expose a dumb "git_blame_extra_args" setting instead? I haven't used it myself, but `--first-parent` seems to be another git-blame options that make sense to expose. Both `-C` and `-M` take an optional argument as well.

Does the complexity of adding specialized support to all these justify the slight usability increase for something only a minority is likely to tweak?

alexpyattaev:
Thank you for looking into this! I think having extra-args is fine in principle but it might be a bit brittle (with some values the gui might break). I would be happy either way.

cole-miller:
Thanks for opening a PR--however, I think we don't want to add these as knobs for the blame features in Zed. Every additional setting does come with some cost in maintenance, and I think we'd rather stick to the default in this case.

michaelharmonart:
> Thanks for opening a PR--however, I think we don't want to add these as knobs for the blame features in Zed. Every additional setting does come with some cost in maintenance, and I think we'd rather stick to the default in this case.

I'd like to express interest in still having this setting. I understand the hesitancy to expose this, but It would make a big difference to my user experience

PR #44334, opened 2025-12-07
gpui: Enable direct-to-display optimization for metal

marcocondrache:
When profiling Zed with Instruments, a warning appears indicating that surfaces cannot be pushed directly to the display as they are non-opaque. This happens because the metal layer is currently marked as non-opaque by default, even though the window itself is not transparent.

image

Metal on macOS can bypass compositing and present frames directly to the display when several conditions are met. One of those conditions is that the backing layer must be declared opaque. Apple’s documentation notes that marking layers as opaque allows the system to avoid unnecessary compositing work, reducing GPU load and improving frame pacing

Ref: https://developer.apple.com/documentation/metal/managing-your-game-window-for-metal-in-macos

This PR updates the Metal renderer to mark the layer as opaque whenever the window does not use transparency. This makes Zed eligible for macOS’s direct-to-display optimization in scenarios where the system can apply it.

Release Notes:

- Mark metal layers opaque for non-transparent windows to allow direct-to-display when supported


crates/gpui/src/platform/mac/metal_renderer.rs
```diff
@@ -152,7 +152,7 @@ impl MetalRenderer {
         let layer = metal::MetalLayer::new();
         layer.set_device(&device);
         layer.set_pixel_format(MTLPixelFormat::BGRA8Unorm);
-        layer.set_opaque(false);
+        layer.set_opaque(!transparent);
```
Anthony-Eid, line 157: It would probably be a good idea to add the reference you have in your PR as a comment here

marcocondrache: Yes that's right! Done

crates/gpui/src/platform/mac/metal_renderer.rs
```diff
@@ -152,8 +152,13 @@ impl MetalRenderer {
...
         layer.set_device(&device);
         layer.set_pixel_format(MTLPixelFormat::BGRA8Unorm);
-        layer.set_opaque(false);
+        // Support direct-to-display rendering if the window is not transparent
+        // https://developer.apple.com/documentation/metal/managing-your-game-window-for-metal-in-macos
+        layer.set_opaque(!transparent);
         layer.set_maximum_drawable_count(3);
+        // We already present at display sync with the display link
+        // This allows to use direct-to-display even in window mode
+        layer.set_display_sync_enabled(false);
```
marcocondrache: After experimenting with macos-blade, I discovered that macOS can push frames in direct mode even when not in fullscreen. Through some debugging, I found that if we request the layer to update at vsync (default), we are required to use composite mode. This seems redundant since we already utilize the display link to present at the display refresh rate

Running the paths_bench example:

Image

PR #44342, opened 2025-12-07
node_runtime: Use `semver::Version` to represent package versions

tidely:
Closes #ISSUE

This PR is rather a nice to have change than anything critical, so review priority should remain low.

Switch to using `semver::Version` for representing node binary and npm package versions. This is in an effort to root out implicit behavior and improve type safety when interacting with the `node_runtime` crate by catching invalid versions where they appear. Currently Zed may implicitly assume the current version is correct, or always install the newest version when a invalid version is passed. `semver::Version` also doesn't require the heap, which is probably more of a fun fact than anything useful.

`npm_install_packages` still takes versions as a `&str`, because `latest` can be used to fetch the latest version on npm. This could likely be made into an enum as well, but would make the PR even larger.

I tested changes with some node based language servers and external agents, which all worked fine. It would be nice to have some e2e tests for node. To be safe I'd put it on nightly after a Wednesday release.

Release Notes:


tidely:
#44343 should get merged first, since it fixes a bug

osiewicz (approved):
LGTM. Will merge after tomorrow's release. Thank you!

PR #44343, opened 2025-12-07
languages: Check whether to update `typescript-language-server`

tidely:
Closes #43155

Adds a missing check to also update packages when the `typescript-language-server` package is outdated. 

I created a new `SERVER_PACKAGE_NAME ` constant so that the package name isn't coupled to the language server name inside of Zed.

Release Notes:

- Fixed the typescript language server falling out of date

### DIFF
--- a/crates/languages/src/typescript.rs
+++ b/crates/languages/src/typescript.rs
@@ -720,7 +720,7 @@ impl LspInstaller for TypeScriptLspAdapter {
                         latest_version.typescript_version.as_str(),
                     ),
                     (
-                        "typescript-language-server",
+                        Self::SERVER_PACKAGE_NAME,
                         latest_version.server_version.as_str(),
                     ),
                 ],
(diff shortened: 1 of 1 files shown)


crates/languages/src/typescript.rs
```diff
@@ -672,7 +680,17 @@ impl LspInstaller for TypeScriptLspAdapter {
...
+        let should_install_server = self
+            .node
+            .should_install_npm_package(
+                Self::SERVER_PACKAGE_NAME,
+                &server_path,
+                container_dir,
+                VersionStrategy::Latest(version.server_version.as_str()),
+            )
+            .await;
+
```
MrSubidubi: We don't need to check for this if we know we should install the package

MrSubidubi (changes requested):
Thanks for this! One minor nit, then this is good to go.

crates/languages/src/typescript.rs
tidely: Yeah, that's fair, thx

MrSubidubi (approved):
Thanks for the quick follow-up and the fix!

PR #44349, opened 2025-12-07
Update Recent Projects picker to show SSH host

wln:
**Problem addressed:**

Cannot distinguish an identical path on multiple remote hosts in the "Recent Projects" picker.

**Related work:**

- Issue #40358 (already closed) identified the issue in the "Expected Behavior" section for both WSL and SSH remotes.
- PR #40375 implemented WSL distro labels.

**Screenshots:**

Before:
screenshot-sample-project-before

After:
screenshot-sample-project-after

**Implementation notes:**

RemoteConnectionOptions::display_name() will be subject to exhaustiveness checking on RemoteConnectionOption variants.

Keeps the same UI approach as the WSL distro variants.

Release Notes:

- Improved Recent Projects picker: now displays SSH hostname with remotes.

### DIFF
--- a/crates/recent_projects/src/recent_projects.rs
+++ b/crates/recent_projects/src/recent_projects.rs
@@ -532,8 +532,8 @@ impl PickerDelegate for RecentProjectsDelegate {
             .unzip();
 
         let prefix = match &location {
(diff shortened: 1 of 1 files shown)


danilo-leal (approved):
Thank you!

PR #44353, opened 2025-12-08
gpui: Improve `cx.on_action` method to support chaining

huacnlee:
Release Notes:

- N/A

To let `cx.on_action` support chaining like the `on_action` method of Div.

https://github.com/zed-industries/zed/blob/ebcb2b2e646f10006dc40167d16e82ae74caa3a2/crates/agent_ui/src/acp/thread_view.rs#L5867-L5872


Veykril (approved):
Seems reasonable to me, thanks!

PR #43555, opened 2025-11-26
markdown_preview: Fix markdown tables taking up the full width of the parent element

RemcoSmitsDev:
Closes #39152

This PR fixes an issue where we would render Markdown tables full width based on their container size. We now render tables based on their content min size, meaning you are still allowed to make the table render as it was before by making the columns `w_full`.

I had to change the `div()` to `v_flex().items_start()` because this introduced a weird displaying behavior of the outside table border, because the grid container was not shrinking due to It was always taking up the full width of their container.

**Before**
Screenshot 2025-11-26 at 14 37 19

**After**
Screenshot 2025-11-26 at 14 56 12

**Code example**

```markdown
|   Name   |   Age   |   Occupation   |
|:--------:|:-------:|:--------------:|
| Alice    |  28     | Engineer       |
| Bob      |  34     | Designer       |
| Carol    |  25     | Developer      |


| Syntax      | Description |
| ----------- | ----------- |
| Header      | Title       |
| Paragraph   | Text        |


| City           | Population (approx.) | Known For                          |
|----------------|----------------------|------------------------------------|
| New York       | 8,500,000            | Statue of Liberty, Wall Street     |
| Los Angeles    | 4,000,000            | Hollywood, film industry           |
| Chicago        | 2,700,000            | Architecture, deep-dish pizza      |
| Houston        | 2,300,000            | NASA, energy industry              |
| Miami          | 470,000              | Beaches, Latin culture             |
| San Francisco  | 800,000              | Golden Gate Bridge, Silicon Valley |
| Las Vegas      | 650,000              | Casinos, nightlife                 |


<table>
    <caption>Table Caption</caption>
  <thead>
    <tr>
    <th>ID asjkfjaslkf jalksjflksajflka jlksdla k</th>
    <th>Name</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>1</td>
      <td>Chris</td>
    </tr>
    <tr>
      <td>2</td>
      <td>Dennis</td>
    </tr>
    <tr>
      <td>3</td>
      <td>Sarah</td>
    </tr>
    <tr>
      <td>4</td>
      <td>Karen</td>
    </tr>
  </tbody>
</table>
```

cc @bennetbo

Release Notes:

- Markdown Preview: Markdown tables scale now based on their content size

PR #44368, opened 2025-12-08
gpui: Revert grid template columns default behavior to align with Tailwind

madcodelife:
When using the latest version of `GPUI`, I found some grid layout issues. I discovered #43555 modified the default behavior of grid template columns. I checked the implementation at https://tailwindcss.com/docs/grid-template-columns, and it seems our previous implementation was correct.

If a grid layout is placed inside a flexbox, the layout becomes unpredictable.

```rust
impl Render for HelloWorld {
    fn render(&mut self, _window: &mut Window, _cx: &mut Context<Self>) -> impl IntoElement {
        div()
            .flex()
            .size(px(500.0))
            .bg(rgb(0x505050))
            .text_xl()
            .text_color(rgb(0xffffff))
            .child(
                div()
                    .size_full()
                    .gap_1()
                    .grid()
                    .grid_cols(2)
                    .border_1()
                    .border_color(gpui::red())
                    .children((0..10).map(|ix| {
                        div()
                            .w_full()
                            .border_1()
                            .border_color(gpui::green())
                            .child(ix.to_string())
                    })),
            )
    }
}
```

| Before | After |
| - | - |
| After1 | Before1 |

I also placed the grid layout example inside a flexbox too.

| Before | After |
| - | - |
| After | Before |

I tested the changes from the previous PR, and it seems that setting the table's parent to `v_flex` is sufficient to achieve a non-full table width without modifying the grid layout. This was already done in the previous PR. 

I reverted the grid changes, the blue border represents the table width. cc @RemcoSmitsDev 

table

So, I believe we should revert to this implementation to align with tailwindcss behavior and avoid potential future problems, especially since the cause of this issue is difficult to pinpoint.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/taffy.rs
+++ b/crates/gpui/src/taffy.rs
@@ -8,7 +8,6 @@ use std::{fmt::Debug, ops::Range};
 use taffy::{
     TaffyTree, TraversePartialTree as _,
     geometry::{Point as TaffyPoint, Rect as TaffyRect, Size as TaffySize},
-    prelude::min_content,
     style::AvailableSpace as TaffyAvailableSpace,
     tree::NodeId,
 };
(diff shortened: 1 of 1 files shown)


RemcoSmitsDev:
Hey, sorry for breaking it. Great find though! Looks good to me!

bennetbo (approved):
Somehow missed that in the Review, thank you!

Issue #26560, opened 2025-03-12
Staged and Unstaged diffs


mikayla-maki:
### Summary

Currently, the git panel and diff view only show the combined staged and unstaged diffs, by interleaving them together in the buffer.

However, a lot of git clients support fully separate staged and unstaged diffs, and Zed should be no exception.

### Zed Version and System Specs

N/A

### If applicable, attach your `~/Library/Logs/Zed/Zed.log` file to this issue.

<details><summary>Zed.log</summary>

```

```
</details>

mohebifar:
I created a PR for this: https://github.com/zed-industries/zed/pull/36646

secondl1ght:
Related discussion: https://github.com/zed-industries/zed/discussions/26862

ddoemonn:
I would like to work on this!

secondl1ght:
@ddoemonn any updates here?

controversial:
Shouldn’t it be a `-` next to the file under Staged changes? in vscode the action under “staged changes” is to _unstage_ the given file (even if the current file is only partially changed) and the action under “changes” is to stage the remainder of the current file.

<img width="373" height="188" alt="Image" src="https://github.com/user-attachments/assets/62248f27-634f-4f9e-9400-b5b14cf62d18" />

I also think the vscode behavior where the +/- only appears on hover is a bit more approachable / less busy looking

secondl1ght:
> Here's a quick preview of my new attempt:
> 
>  Screen.Recording.2025-12-22.at.2.45.34.PM.mov 
> Changes:
> 
> * Instead of the checkbox, we show a +/- icon button similar to vscode.
> * You can stage all or unstage all using the +/- icon buttons in the header again similar to vscode.
> * Instead of using the ProjectDiff buffer, we show a new GitDiff buffer opened in separate tabs. When you open the diff from the staged section, you'll see the diff between INDEX vs. HEADER. When you open the diff from the unstaged changes section, you'll see the diff between WORKING DIR vs. INDEX. Again similar to vscode.
> 
> In this video there are still a few bugs like opening a new buffer tab on every click instead of reusing the same tab which I'll fix. But I just wanted to get it out the door asap to get some feedback.

I have some comments @mohebifar, but I am waiting to see @cole-miller's feedback first. I think you are waiting to hear from him before opening a PR right?

firatoezcan:
There you go guys: https://github.com/firatoezcan/zed

[...]

Issue #29635, opened 2025-04-30
Git: Unable to view changes for staged files


freddydc:
### Description
After staging a file and then making further edits to that same file, Git fails to detect or display any of the new changes.

Steps to trigger the problem:
1. Do stage a file, for example `main.rs`.
2. Make further edits to `main.rs`.
3. In the Git interface you cannot see any changes of this file.

Actual Behavior:
In the Git interface you cannot see any changes of this file after editing it post-stage.

![Image](https://github.com/user-attachments/assets/d981545e-9a94-4f94-828d-f15cc2f4df1e)

Expected Behavior:
Git should display any edits made to a file, even after it has been staged, like in VSCode.

![Image](https://github.com/user-attachments/assets/1863518f-5bf9-433b-824d-6eeb2941816f)

### Zed Version and System Specs

Zed: v0.183.12 (Zed)
OS: Linux Wayland ubuntu 24.04
Memory: 16 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 5070 || NVIDIA || 570.133.07

mgsloan:
I certainly agree that this is a very necessary feature!  I think it would be covered by separate diffs for unstaged vs staged (#26560) or the ability to put the diff in a mode that just views unstaged changes.

katie-z-geer:
Relates to: https://github.com/zed-industries/zed/issues/26560

vampik33:
I would like to work on this!

katie-z-geer:
I'll assign you!

vampik33:
To support current workflow of diff changes, I purpose to add "Staged" / "Unstaged" filter checkboxes to project diff toolbar

Here is the example how it would look like:
[Screencast from 08-12-25 14:38:33.webm](https://github.com/user-attachments/assets/e4021a47-4bc4-4c04-ad19-83f7f47261cb)

esthertrapadoux:
Hey @vampik33! Thanks for working on this issue. Wondering if you're free for a pairing session with one of our Git team members: https://cal.com/esther-trapadoux-zed/30min?overlayCalendar=true

JosephTLyons:
Hello,

We’re closing this issue as part of a cleanup of older feature requests.

[...]

PR #36646, opened 2025-08-21
git_ui: Add staged/unstaged view mode like VS Code

mohebifar:
Closes #26560

Release Notes:

- Added a new settings option to `git_panel` settings `display_mode` that allows you to switch to a mode similar to VSCode's default which groups files by staged/unstaged instead of tracked vs untracked. If a file has some staged changes and some unstaged changes at the same time, it will appear in both sections just like VSCode.

https://github.com/user-attachments/assets/4dc9298c-2fa7-4efc-9e55-ab38e5333041

PR #44379, opened 2025-12-08
git_ui: Add staged/unstaged filter checkboxes to project diff toolbar

vampik33:
git_ui: Add staged/unstaged filter checkboxes to project diff toolbar

This PR adds "Staged" and "Unstaged" checkboxes to the project diff toolbar, allowing users to filter the diff view to show only staged or unstaged changes.

  Key Implementation Details

New DiffFilterMode enum (buffer_diff.rs)
- All - Show all hunks (default)
- StagedOnly - Show only staged/partially staged hunks
- UnstagedOnly - Show only unstaged/partially staged hunks

Hunk visibility methods (DiffHunk)
- is_visible_when_staged_only() - Returns true for NoSecondaryHunk, OverlapsWithSecondaryHunk, SecondaryHunkRemovalPending
- is_visible_when_unstaged_only() - Returns true for HasSecondaryHunk, OverlapsWithSecondaryHunk, SecondaryHunkAdditionPending

Filter logic (project_diff.rs)
- Checkboxes are mutually exclusive - activating one disables the other
- Full rebuild on filter change ensures context lines are properly recalculated around visible hunks
- File-level filtering by StageStatus (hides fully staged/unstaged files)
- Hunk-level filtering by DiffHunkSecondaryStatus (hides staged/unstaged hunks within partially staged files)

Performance optimization (branch_diff.rs)
- Early filtering in load_buffers() avoids loading buffers for files that will be filtered out

Tests
- Unit tests for is_visible_when_staged_only() and is_visible_when_unstaged_only() covering all DiffHunkSecondaryStatus variants
- Test for DiffFilterMode::default() returning All

  Related Issues
- #29635
- #26560
  
  Release Notes:
- Improve Diff View


vampik33:
To support current workflow of diff changes, I purpose to add "Staged" / "Unstaged" filter checkboxes to project diff toolbar

Here is the example how it would look like:
[Screencast from 08-12-25 14:38:33.webm](https://github.com/user-attachments/assets/e4021a47-4bc4-4c04-ad19-83f7f47261cb)

secondl1ght:
Just noting that this PR does not close either of the issues that OP referenced (#26560, #29635). This seems to be an entirely different feature, please see the VSCode functionality for what those issues are referring to.

vampik33:
> Just noting that this PR does not close either of the issues that OP referenced (#26560, #29635). This seems to be an entirely different feature, please see the VSCode functionality for what those issues are referring to.

Yes, I know that it is a different functionality than in VSCode.
VSCode approach was implemented in this PR: https://github.com/zed-industries/zed/pull/36646
But it got rejected. So I implemented another solution, which can help to see Staged / Unstaged changes.

secondl1ght:
> > Just noting that this PR does not close either of the issues that OP referenced (#26560, #29635). This seems to be an entirely different feature, please see the VSCode functionality for what those issues are referring to.
> 
> Yes, I know that it is a different functionality than in VSCode. VSCode approach was implemented in this PR: #36646 But it got rejected. So I implemented another solution, which can help to see Staged / Unstaged changes.

That PR you mention was an initial attempt, but did not meet all the requirements for this feature. It wasn't able to progress further due to the Zed team not having the "bandwidth" to work with the author. They have now set aside time to focus on Git during the Let's Git Together initiative, and @ddoemonn has been assigned to complete #26560.

A quick summary:
1. The VSCode approach did not get fully implemented in #36646
2. The reason that PR was closed was due to the maintainers workload not disapproving of the content
3. This PR does not close #26560 or #29635
4. The video you shared especially does not show anything related to #29635

To be honest it is hard to tell what is being demo'd in the video, I see 2 checkboxes for Staged and Unstaged but they don't seem to work as expected. For example, when both are unchecked there are still diffs shown in the multi-buffer, but you would expect there to be none in that state.

Anyway I am not here to review the PR, only to note that if this PR gets merged in its current state those issues should not be closed. Thanks!

cole-miller:
Thanks for opening a PR!

After some discussion, we are open to taking a PR for the approach in https://github.com/zed-industries/zed/pull/36646, and I think we'd prefer that to this approach, because it's more compatible with the git panel integration (we can change the ordering of paths in the project diff using a custom `PathKey` namespace).

vampik33:
> Thanks for opening a PR!
> 
> After some discussion, we are open to taking a PR for the approach in #36646, and I think we'd prefer that to this approach, because it's more compatible with the git panel integration (we can change the ordering of paths in the project diff using a custom `PathKey` namespace).

Ok, got it

PR #44380, opened 2025-12-08
language: Add `LanguageName::new_static` to reduce allocations

tidely:
Implements a specialized constructor `LanguageName::new_static` for `&'static str` which reduces allocations.

`LanguageName::new` always backs the underlying `SharedString` with an owned `Arc<str>` even when a `&'static str` is passed. This makes us allocate each time we create a new `LanguageName` no matter what. Creating a specialized constructor for `&'static str` allows us to essentially construct them for free.

Additional change:
Encourages using explicit constructors to avoid needless allocations. Currently there were no instances of this trait being called where the lifetime was not `'static` saving another 48 locations of allocation.

```rust
impl<'a> From<&'a str> for LanguageName {
    fn from(str: &'a str) -> Self {
        Self(SharedString::new(str))
    }
}

// to 

impl From<&'static str> for LanguageName {
    fn from(str: &'static str) -> Self {
        Self(SharedString::new_static(str))
    }
}

```

Release Notes:

- N/A

### DIFF
--- a/crates/dap_adapters/src/python.rs
+++ b/crates/dap_adapters/src/python.rs
@@ -870,7 +870,7 @@ impl DebugAdapter for PythonDebugAdapter {
                 .active_toolchain(
                     delegate.worktree_id(),
                     base_path.into_arc(),
-                    language::LanguageName::new(Self::LANGUAGE_NAME),
+                    language::LanguageName::new_static(Self::LANGUAGE_NAME),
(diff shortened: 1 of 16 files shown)


MrSubidubi (approved):
Nice find, and given that we already have something similar for `LanguageServerName`, happy to merge this. Thanks!

PR #44398, opened 2025-12-08
extension_api: Update default trait implementations for language server settings

MrSubidubi:
With this change, language servers will have settings and initialization options propagated to them by default. This removes the need for authors to implement this manually whilst still giving the option to opt out of this behavior.

Release Notes:

- N/A


MrSubidubi:
Closing this for now - I do agree that explicitly implementing this for now is the better path forward.

PR #40305, opened 2025-10-16
terminal: New settings for path hyperlink regexes

davewa:
Closes:
- #12338
- #40202 

1. Adds two new settings which allow customizing the set of regexes used to identify path hyperlinks in terminal
1. Fixes path hyperlinks for paths containing unicode emoji and punctuation, for example, `mojo.🔥`
1. Fixes path hyperlinks for Windows verbatim paths, for example, `\\?\C:\Over\here.rs`.
1. Improves path hyperlink performance, especially for terminals with a lot of content
1. Replaces existing custom hard-coded default path hyperlink parsing logic with a set of customizable default regexes

## New settings

(from default.json)

### terminal.path_hyperlink_regexes

Regexes used to identify paths for hyperlink navigation. Supports optional named capture
groups `path`, `line`, `column`, and `link`. If none of these are present, the entire match
is the hyperlink target. If `path` is present, it is the hyperlink target, along with `line`
and `column` if present. `link` may be used to customize what text in terminal is part of the
hyperlink. If `link` is not present, the text of the entire match is used. If `line` and
`column` are not present, the default built-in line and column suffix processing is used
which parses `line:column` and `(line,column)` variants. The default value handles Python
diagnostics and common path, line, column syntaxes. This can be extended or replaced to
handle specific scenarios. For example, to enable support for hyperlinking paths which
contain spaces in rust output,
```
[
  "\\s+(-->|:::|at) (?<link>(?<path>.+?))(:$|$)",
  "\\s+(Compiling|Checking|Documenting) [^(]+\\((?<link>(?<path>.+))\\)"
],
```
could be used. Processing stops at the first regex with a match, even if no link is
produced which is the case when the cursor is not over the hyperlinked text. For best
performance it is recommended to order regexes from most common to least common. For
readability and documentation, each regex may be an array of strings which are collected
into one multi-line regex string for use in terminal path hyperlink detection.

### terminal.path_hyperlink_timeout_ms
Timeout for hover and Cmd-click path hyperlink discovery in milliseconds. Specifying a
timeout of `0` will disable path hyperlinking in terminal.

## Performance

This PR fixes terminal to only search the hovered line for hyperlinks and adds a benchmark. Before this fix, hyperlink detection grows linearly with terminal content, with this fix it is proportional only to the hovered line. The gains come from replacing `visible_regex_match_iter`, which searched all visible lines, with code that only searches the line hovered on (including if the line is wrapped).

Local benchmark timings (terminal with 500 lines of content):

[...]

PR #44407, opened 2025-12-08
terminal: Fix performance issues with hyperlink regex matching

osiewicz:
Problem statement: When given a line that contained a lot of matches of
your hyperlink regex of choice (thanks to #40305), we would look for matches
that intersected with currently hovered point. This is *hella*
expensive, because we would re-walk the whole alacritty grid for each
match. With the repro that Joseph shared, we had to go through 4000 such
matches on each frame render.

Problem solution: We now convert the hovered point into a range within
the line (byte-wise) in order to throw away matches that do not intersect the
hovered range. This lets us avoid performing the unnecessary conversion
when we know it's never going to yield a match range that intersects the
hovered point.

Release Notes:

- terminal: Fixed performance regression when handling long lines.

### DIFF
--- a/crates/terminal/src/terminal_hyperlinks.rs
+++ b/crates/terminal/src/terminal_hyperlinks.rs
@@ -225,16 +225,18 @@ fn path_match<T>(
     let mut line = String::with_capacity(
         (line_end.line.0 - line_start.line.0 + 1) as usize * term.grid().columns(),
     );
-    line.push(term.grid()[line_start].c);
+    let first_cell = &term.grid()[line_start];
+    line.push(first_cell.c);
     let mut start_offset = 0;
     let mut hovered_point_byte_offset = None;
-    if hovered == line_start {
-        let cell = &term.grid()[line_start];
-        if !cell.flags.intersects(WIDE_CHAR_SPACERS) {
-            start_offset += cell.c.len_utf8();
+
+    if !first_cell.flags.intersects(WIDE_CHAR_SPACERS) {
+        start_offset += first_cell.c.len_utf8();
+        if line_start == hovered {
             hovered_point_byte_offset = Some(0);
         }
     }
+
     for cell in term.grid().iter_from(line_start) {
         if cell.point > line_end {
             break;
@@ -333,8 +335,7 @@ fn path_match<T>(
             let link_range = captures
                 .name("link")
                 .map_or_else(|| match_range.clone(), |link| link.range());
-
-            if !match_range.contains(&hovered_point_byte_offset) {
+            if !(link_range.start..=link_range.end).contains(&hovered_point_byte_offset) {
                 // No match, just skip.
                 continue;
             }
--- a/crates/terminal/src/terminal_hyperlinks.rs
+++ b/crates/terminal/src/terminal_hyperlinks.rs
@@ -227,15 +227,12 @@ fn path_match<T>(
     );
     line.push(term.grid()[line_start].c);
     let mut start_offset = 0;
-    let mut hovered_range = None;
+    let mut hovered_point_byte_offset = None;
     if hovered == line_start {
         let cell = &term.grid()[line_start];
         if !cell.flags.intersects(WIDE_CHAR_SPACERS) {
             start_offset += cell.c.len_utf8();
-            hovered_range = Some(Range {
-                start: 0,
-                end: cell.c.len_utf8(),
-            });
+            hovered_point_byte_offset = Some(0);
         }
     }
     for cell in term.grid().iter_from(line_start) {
@@ -244,12 +241,9 @@ fn path_match<T>(
         }
         let is_spacer = cell.flags.intersects(WIDE_CHAR_SPACERS);
         if cell.point == hovered {
-            debug_assert!(hovered_range.is_none());
+            debug_assert!(hovered_point_byte_offset.is_none());
 
-            hovered_range = Some(Range {
-                start: start_offset,
-                end: start_offset + cell.c.len_utf8(),
-            });
+            hovered_point_byte_offset = Some(start_offset);
         } else if cell.point < hovered && !is_spacer {
             start_offset += cell.c.len_utf8();
         }
@@ -262,7 +256,7 @@ fn path_match<T>(
         }
     }
     let line = line.trim_ascii_end();
-    let hovered_range = hovered_range?;
+    let hovered_point_byte_offset = hovered_point_byte_offset?;
     let found_from_range = |path_range: Range<usize>,
                             link_range: Range<usize>,
                             position: Option<(u32, Option<u32>)>| {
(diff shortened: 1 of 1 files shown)


Veykril (approved):

crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -315,13 +327,17 @@ fn path_match<T>(
             let link_range = captures
                 .name("link")
                 .map_or(match_range, |link| link.range());
+
+            if hovered_range.start > path_range.end || path_range.start > hovered_range.end {
```
osiewicz: This should use `match_range`, not `path_range`

benbrandt (approved):

crates/terminal/src/terminal_hyperlinks.rs
davewa: This test should be on `link_range`. Also, can't `hovered_range` just be `hovered_byte_offset`? Then this would just be `!link_range.contains(hovered_byte_offset)`, no?

crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -315,13 +327,17 @@ fn path_match<T>(
...
+                continue;
+            }
             let found = found_from_range(path_range, link_range, line_column);
 
-            if let Some(found) = found {
-                path_found = true;
-                if found.1.contains(&hovered) {
-                    return Some(found);
-                }
+            path_found = true;
```
davewa: `path_found = true` needs to move up before the hover test & continue, could actually go above `let match_range = ...`

davewa (changes requested):

crates/terminal/src/terminal_hyperlinks.rs
osiewicz: That's fair

osiewicz: Using `link_range` instead triggers a single test failure:
```
---- terminal_hyperlinks::tests::path::issues::issue_12338_regex stdout ----

thread 'terminal_hyperlinks::tests::path::issues::issue_12338_regex' (48381149) panicked at crates/terminal/src/terminal_hyperlinks.rs:1682:21:
No hyperlink found
     at crates/terminal/src/terminal_hyperlinks.rs:798:

Hovered on '''

      [   ]
      [012]
       ↓
 [  0] .rw
 [  1] -r-
 [  2] -r-
 [  3] -
 [  4]
 [  5] 0
 [  6]
 [  7] sta
 [  8] ff
 [  9] 05-
 [ 10] 27
 [ 11] 14:
 [ 12] 03
→[ 13] 'te
 [ 14] st
 [ 15] fil
 [ 16] e 1
 [ 17] .tx
 [ 18] t'
 [ 19]
```
I'm inclined to keep using `match_range`, even if it's a bit too lenient.

crates/terminal/src/terminal_hyperlinks.rs
osiewicz: I don't think that's correct, as that would make us bail if any match is found within current line, even when it is not hovered. We'd then bail before running other regexes in the outer loop.

davewa: That is the current design intentionally though: If a regex matches, but the hovered point is not over `link` part of the match, we stop without a match and without searching the rest of the regexes in the outer loop. Reasoning is that if a match is found, that is the intended path hyperlink, so if there is a match, but your hover isn't over it, it's fine to stop, likely the use will move the hover within the link and the same match will then succeed, the other regexes don't need to be searched because they are in priority order...

If that's a reasonable design or not is a different issue that the perf regression, and as it is, this changes the current behavior if this isn't moved. It seems this change in behavior didn't cause a test to fail, so that's a test hole. I'd prefer to change this behavior in a separate PR if there is a clear reason to (which there might be?).

osiewicz: Okay, let's move it then. I don't have a preference for the behavior here.

crates/terminal/src/terminal_hyperlinks.rs
davewa: Hmmm, looking at that test, `link_range` _should_ work for it, I'm not sure why it's failing. Building your PR branch now to investigate....

davewa: This test is failing with `link_range` because `hovered_point_byte_offset` is 38, but it should be 39, so there must be a bug in how `hovered_point_byte_offset` is being calculated currently.

osiewicz: Could it be that we're only accounting for non-spacer characters when fetching the offset of hovered point?

davewa: I _think_ it's because we're only adding the first char's width to the start_offset if `hovered == line_start`, testing now...

osiewicz: Yep, that's a part of the problem. Then, the `link_range` is also a non-inclusive range.

davewa: Ah, yes I see. I merged my in-progress fix with your latest push, here's the diff. What do you think? I think yours has the `hovered_point_byte_offset` referring to the wrong char's bytes, then expanding the `link_range` to include it, mine has `hovered_point_byte_offset` referring to the correct char's bytes (the first byte of the hovered char, or the last byte of the hovered char, iff the hover was on the trailing spacer).

```
diff --git a/crates/terminal/src/terminal_hyperlinks.rs b/crates/terminal/src/terminal_hyperlinks.rs
index d9bf9967b0..63de60507e 100644
--- a/crates/terminal/src/terminal_hyperlinks.rs
+++ b/crates/terminal/src/terminal_hyperlinks.rs
@@ -245,6 +245,10 @@ fn path_match<T>(
         if cell.point == hovered {
             debug_assert!(hovered_point_byte_offset.is_none());
 
+            if cell.flags.contains(Flags::WIDE_CHAR_SPACER) {
+                // If we hovered on a trailing spacer, back up to the end of the previous char's bytes.
+                start_offset -= 1;
+            }
             hovered_point_byte_offset = Some(start_offset);
         } else if cell.point < hovered && !is_spacer {
             start_offset += cell.c.len_utf8();
@@ -335,7 +339,8 @@ fn path_match<T>(
             let link_range = captures
                 .name("link")
                 .map_or_else(|| match_range.clone(), |link| link.range());
-            if !(link_range.start..=link_range.end).contains(&hovered_point_byte_offset) {
+
+            if !link_range.contains(&hovered_point_byte_offset) {
                 // No match, just skip.
                 continue;
             }
```

davewa (approved):
Looks good. Thanks for finding and fixing this regression, I'll be adding a perf test for this scenario soon in a separate PR.

osiewicz:
@davewa thanks for a thorough review!

PR #44431, opened 2025-12-08
New evals for inline assistant

mikebenfield:
Also factor out some common code in the evals.

Release Notes:

- N/A


mikebenfield:
@mikayla-maki In addition to factoring out the check in run_eval_test, I put a guard in eval and lowered the number of iterations for each test to 10 - otherwise we get tokio errors about too many open files. It's not immediately clear to me what the right way to fix that is; is it clear to you?

mikebenfield:
@mikayla-maki There was an error in your code with parameters being in the wrong order; I fixed that. I also realized the way out of the Tokio error was to call `quit` on the `TestAppContext` so that Tokio can be properly shut down after each test.

PR #44435, opened 2025-12-08
vim: Center screen on * and # search results

0xErwin1:
Closes #43391

Center the cursor vertically when using `*` and `#` commands to search for the word under cursor, providing better context. Preserves the traditional vim scrolling behavior for `n` and `N` commands.

Release Notes:

- N/A


0xErwin1:
@cla-bot check

ConradIrwin:
Thanks! I don't think we should change this – it's also annoying if your screen scrolls unnecessarily. If you'd like to avoid the cursor hitting the edge of the screen you can set the scroll offset in vim (or the vertical scroll margin in zed: zed://settings/vertical_scroll_margin) so that you get some context when your cursor gets close to the edge of the screen.

Issue #11473, opened 2024-05-07
Dev Containers


abhibeckert:
### Check for existing issues

- [X] Completed

### Describe the feature

A Dev Container is a system that runs your code and part of your development environment (Language Server, Unit Tests, Debugger, Build Scripts, etc) inside a container. Many people consider this a blocking feature, as it's arguably an industry best practice to manage projects with complicated dependencies.

Visual Studio Code has a great implementation of this feature. Essentially you describe your environment and dependencies in a `.devcontainer` directory filled with config files, and the editor spins up a container when you open the project - it only takes a few seconds and often needs. Your code and the majority of VSCode extensions will run inside the container.

This feature request has been previously discussed in #5347 Remote Development and is related to the `collab` feature, but it's not quite the same thing and should be tracked separately in the issue tracker.

Moshyfawn:
Related #5347

versecafe:
Getting started on this I'd love help from anyone who's interested, took a little bit to figure out GPUI, panel registration, and settings since they're not super documented but here's the branch, just got the basic UI for the panel up, starting on the actual logic using [bollard](https://crates.io/crates/bollard)

### UI Example *Very Early*
<img width="1402" alt="image" src="https://github.com/zed-industries/zed/assets/147033096/e0619ec5-f73e-4f58-b606-d0a3a29375ca">


### Branch
https://github.com/versecafe/zed/tree/dev-containers

@abhibeckert Any opinions on how closely you think it should match VS Codes implementation, both in UI and in features?

ConradIrwin:
@versecafe Thanks for taking a pass at this! This is one of the top features for people in the remote development alpha too.

For remote servers we ended up choosing to use a modal instead of a panel because the primary use-case of showing the list is to select one and open a new workspace (it didn't seem useful enough to show the list permanently to take up screen real estate all the time). Are there other use-cases that mean it's helpful to have this in an always-open panel while working on code, or should we add this to the cmd-shift-o dialogue too?

As a heads up, @maxbrunsfeld has been exploring ideas for a worktree based remoting feature, which is probably a better starting point than the current remoting feature (which installs a full zed on the remote host and communicates over the internet).

What are the next steps on this in your mind?

chloerei:
One of the features I rely on in vscode devcontainer is to enable extensions in the container. Since my host environment does not have dependencies installed, extensions may not work properly in the host environment.

For example, in devcontainer.json:

```
  "customizations": {
    "vscode": {
[...]

PR #44442, opened 2025-12-08
Introduce MVP Dev Containers support

KyleBarton:
Partially addresses #11473 

MVP of dev containers with the following capabilities:

- If in a project with `.devcontainer/devcontainer.json`, a pop-up notification will ask if you want to open the project in a dev container. This can be dismissed:
Screenshot 2025-12-08 at 3 15 23 PM
- Similarly, if a `devcontainer.json` file is in the project, you can open a devcontainer (or go the devcontainer.json file for further editing) via the `open remote` modal:

https://github.com/user-attachments/assets/61f2fdaa-2808-4efc-994c-7b444a92c0b1

*Limitations*

This is a first release, and comes with some limitations:
- Zed extensions are not managed in `devcontainer.json` yet. They will need to be installed either on host or in the container. Host + Container sync their extensions, so there is not currently a concept of what is installed in the container vs what is installed on host: they come from the same list of manifests
- This implementation uses the [devcontainer CLI](https://github.com/devcontainers/cli) for its control plane. Hence, it does not yet support the `forwardPorts` directive. A single port can be opened with `appPort`. See reference in docs [here](https://github.com/devcontainers/cli/tree/main/example-usage#how-the-tool-examples-work)
- Editing devcontainer.json does not automatically cause the dev container to be rebuilt. So if you add features, change images, etc, you will need to `docker kill` the existing dev container before proceeding.
- Currently takes a hard dependency on `docker` being available in the user's `PATH`.

Release Notes:

- Added ability to Open a project in a DevContainer, provided a `.devcontainer/devcontainer.json` is present


crates/settings/src/settings_content.rs
```diff
@@ -889,9 +889,21 @@ pub enum ImageFileSizeUnit {
...
 
+#[with_fallible_options]
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq, JsonSchema, MergeFrom, Hash,
+)]
+pub struct DevContainerConnection {
+    pub name: SharedString,
+    pub image: SharedString,
+    pub container_id: SharedString,
+    pub working_directory: SharedString,
```
maxbrunsfeld: I'm curious why the working directory is here on the container connection options. For the SSH and WSL transports, the path is not included in the connection options, it's a separate property of the project. In theory you could even open a multi-folder project in the same docker container.

What's the reasoning behind the `working_directory` being on here?

crates/remote/src/remote_client.rs
```diff
@@ -1077,13 +1083,15 @@ impl ConnectionPool {
 pub enum RemoteConnectionOptions {
     Ssh(SshConnectionOptions),
     Wsl(WslConnectionOptions),
+    DockerExec(DockerExecConnectionOptions),
```
maxbrunsfeld: What do you think about just calling this enum variant `Docker`, instead of `DockerExec`? And similarly for the module `docker_exec`.

crates/recent_projects/src/dev_container_suggest.rs
```diff
@@ -0,0 +1,105 @@
...
+        struct DevContainerSuggestionNotification;
+
+        let notification_id = NotificationId::composite::<DevContainerSuggestionNotification>(
+            SharedString::from(project_path.clone()),
+        );
+
+        workspace.show_notification(notification_id, cx, |cx| {
+            cx.new(move |cx| {
+                MessageNotification::new(
+                    "This project contains a Dev Container configuration file. Would you like to re-open it on a container?",
```
maxbrunsfeld:
```suggestion
                    "This project contains a Dev Container configuration file. Would you like to re-open it in a container?",
```

crates/recent_projects/src/dev_container.rs
```diff
@@ -0,0 +1,249 @@
...
+    match command.output().await {
+        Ok(_) => Ok(()),
+        Err(e) => {
+            log::error!(
+                "Unable to find devcontainer CLI in $PATH. Trying to install with npm {:?}",
+                e
+            );
+
+            if let Err(e) = node_runtime
+                .run_npm_subcommand(None, "install", &["-g", "@devcontainers/cli"])
```
maxbrunsfeld: I think we should install `devcontainer` somewhere in Zed's private `data_dir()` folder, similarly to what we do for language servers, rather than installing it globally.

maxbrunsfeld (commented):
Looking good! I left a few suggestions and questions.

crates/remote/src/remote_client.rs
KyleBarton: Yeah I think this makes sense. After all, it's also using `docker cp` which is distinct from `docker exec`. I'm worried about future devs thinking that this is a more holistic adapter of the docker API, but maybe that's not a big deal.

crates/settings/src/settings_content.rs
KyleBarton: The working directory helps to tell `docker exec` where it has permissions. I've run into some edge cases where `docker_exec` defaults to executing in the root directory, even though the user doesn't have permissions there. See the screenshot below for an example. I think ssh/scp/sftp executes in $HOME by default? But for docker exec, the working directory and the home directory (if set) are the places the connection can assume is safe to operate.

As for whether we can initiate that from the persisted workspace, rather than the persisted remote connection - it's probably possible, but might be a little tricky, since `Connection` and `RemoteConnectionOptions` are tied with a `from` implementation. Could be done, I just don't know if it would read better. Maybe we could pair on this further if you've got some ideas here

Screenshot 2025-12-08 at 4 56 38 PM

crates/recent_projects/src/dev_container.rs
KyleBarton: Makes sense, I'll look further into doing that

KyleBarton: This is done now - I've programmed it so that we use `devcontainer` if it's already available on the user's `$PATH`, but otherwise we'll install to a `devcontainer` subdirectory of the data_dir.

danilo-leal (approved):
🎉

crates/settings/src/settings_content.rs
maxbrunsfeld: Talked in slack. Let's put the remote server in the HOME dir, and proceed with the assumption that HOME exists.

maxbrunsfeld (approved):

crates/recent_projects/src/dev_container.rs
```diff
@@ -0,0 +1,295 @@
...
+        log::info!("Found devcontainer cli on $PATH, using it");
+        Ok(PathBuf::from(&dev_container_cli()))
+    }
+}
+
+async fn devcontainer_up(
+    path_to_cli: &PathBuf,
+    path: Arc<Path>,
+) -> Result<DevContainerUp, DevContainerError> {
+    let mut command = util::command::new_smol_command(path_to_cli.display().to_string());
```
jaudiger, line 125: I recently encountered [this issue](https://github.com/zed-industries/zed/issues/45150) while trying to use `devcontainer` on MacOS. It seems the environment variable `PATH` is not set correctly which prevents `sleep` to be found. Could it come from the usage of `util::command::new_smol_command()` ?

PR #44446, opened 2025-12-09
Add vim `friendly_mode_display` setting for nicer status labels

notnotjake:
Release Notes:

- Added `friendly_mode_display` vim setting to show descriptive vim status labels  instead of classic Vim shorthand notation

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -2123,7 +2123,11 @@
     // Specify the mode as the key and the shape as the value.
     // The mode can be one of the following: "normal", "replace", "insert", "visual".
     // The shape can be one of the following: "block", "bar", "underline", "hollow".
-    "cursor_shape": {}
+    "cursor_shape": {},
+    // Whether Vim status shows friendly mode indicator labels.
+    //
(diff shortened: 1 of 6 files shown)


notnotjake:
When enabled, the Vim status bar shows readable text like "Delete inner..." instead of "di", making it more readable. Also shows "Insert" instead of "-- insert --" which is more aesthetically pleasing. 

This is inspired by how [Nova](https://nova.app) editor implements Vim mode.

Before (default):
CleanShot 2025-12-08 at 19 23 15
CleanShot 2025-12-08 at 19 24 16

With `vim.friendly_mode_display` set to true:
CleanShot 2025-12-08 at 19 26 31
CleanShot 2025-12-08 at 20 15 34

ConradIrwin:
Haha :D this is a cute feature, thank you for sending it!

Unfortunately, I don't think that we should ship it: having that area always showing the pending keystrokes makes most sense to me (the words vim uses to describe its features are purely in the documentation, the product itself uses the key sequences as the identify).

PR #44447, opened 2025-12-09
gpui: Implement `From<String>` for ElementId

huacnlee:
Release Notes:

- N/A

## Before

```rs
div()
    .id(SharedString::from(format!("process-entry-{ix}-command")))
```

## After

```rs
div()
    .id(format!("process-entry-{ix}-command"))
```

### DIFF
--- a/crates/agent_ui/src/acp/thread_view.rs
+++ b/crates/agent_ui/src/acp/thread_view.rs
@@ -3509,7 +3509,7 @@ impl AcpThreadView {
                                         (method.id.0.clone(), method.name.clone())
                                     };
 
-                                    Button::new(SharedString::from(method_id.clone()), name)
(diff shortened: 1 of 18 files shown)


Veykril (approved):
Thanks!

venix:
Sorry for commenting here,  but I was trying to run some examples: 

This broke the example on examples/windows.rs 

I guess others too? I am running 

gpui = "*" 

probably didn't make it into the binary? What might be the issue ?

`error[E0277]: the trait bound `ElementId: From<std::string::String>` is not satisfied
   --> src/main.rs:12:13
    |
 12 |         .id(text.to_string())
    |          -- ^^^^^^^^^^^^^^^^ the trait `From<std::string::String>` is not implemented for `ElementId``

Reverting the examples/windows.rs code to the previous version made it worked.

PR #44464, opened 2025-12-09
terminal: Preserve terminal order and fix pinned count on workspace restore

vampik33:
Replace `FuturesUnordered` with `FuturesOrdered` to maintain terminal tab ordering when restoring a workspace. 
Also clamp `pinned_count` to `items_len()` to prevent panics when fewer terminals are restored than expected.

The changes:
1. Order preservation: `FuturesUnordered` → `FuturesOrdered` ensures terminals restore in their original order
2. Bounds fix: `.min(pane.items_len())` prevents setting pinned count higher than actual items

Closes #44463

Release Notes:

- Preserved terminal order and fixed pinned count on workspace restore


osiewicz:
Nit: could we use FuturesOrdered instead? https://docs.rs/futures/latest/futures/future/fn.join_all.html#see-also

vampik33:
> Nit: could we use FuturesOrdered instead? https://docs.rs/futures/latest/futures/future/fn.join_all.html#see-also

Thank you for your suggestion. Change this to the `FuturesOrdered` as you proposed.

SomeoneToIgnore (approved):
Thank you.

PR #44466, opened 2025-12-09
Add setting to hide user menu in the title bar

notnotjake:
Closes #44417 

Release Notes:

- Added a setting `show_user_menu` (defaulting to true) which shows or hides the user menu (the one with the user avatar) in title bar.


crates/settings/src/settings_content.rs
```diff
@@ -286,6 +286,10 @@ pub struct TitleBarSettingsContent {
     ///
     /// Default: true
     pub show_sign_in: Option<bool>,
+    /// Whether to show the app menu button in the title bar.
+    ///
+    /// Default: true
+    pub show_app_menu: Option<bool>,
```
danilo-leal:
```suggestion
    /// Whether to show the user menu button in the right-side of title bar.
    /// This setting will win over `show_user_picture` and show nothing at all.
    ///
    /// Default: true
    pub show_user_menu: Option<bool>,
```

I think we can clarify the language for this setting a bit, `show_user_menu` should help already, because "app menu" could also mean the native macOS menu, for example. Could I ask to update the docs elsewhere?

danilo-leal (approved):
Thanks!

PR #44489, opened 2025-12-09
Add cross-window tab drag and drop capability

### LINKED ISSUE #6722: Drag tab to open new window
### Check for existing issues

- [X] Completed

### Describe the feature

I'm torn if this is a bug vs an RFE, but I definitely find this an _important_ functionality. Working with multiple monitors, I find being very productive utilizing multiple editor windows. I know that Zed well supports split-windows within the same monitors, but I don't _prefer_ that method as my screens are a little on the small side so that I can support have 4 external monitors.

The expected behavior would be that I can drag an open tab to another monitor and it would spawn a new editor window with the tab.

_No response_

ddoemonn:
Closes #6722

Release Notes:

- Added cross-window tab drag and drop support. You can now drag tabs between different Zed windows or drop them outside a window to create a new window.

https://github.com/user-attachments/assets/107b8881-cbdb-4412-98be-b997516d18ec

### DIFF
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -2021,6 +2021,44 @@ impl App {
         }
     }
 
+    /// Returns a reference to the value of the currently active drag operation,
+    /// if one exists and its value is of type `T`.
+    pub fn active_drag_value<T: 'static>(&self) -> Option<&T> {
+        self.active_drag
+            .as_ref()
+            .and_then(|drag| drag.value.downcast_ref::<T>())
+    }
+
+    /// Returns the cursor offset of the currently active drag operation, if one exists.
+    pub fn active_drag_cursor_offset(&self) -> Option<Point<Pixels>> {
+        self.active_drag.as_ref().map(|drag| drag.cursor_offset)
+    }
+
+    /// Clears the active drag if its value is of type `T`.
+    /// Returns `true` if the drag was cleared.
+    pub fn clear_active_drag_of_type<T: 'static>(&mut self) -> bool {
+        if self
+            .active_drag
+            .as_ref()
+            .is_some_and(|drag| drag.value.is::<T>())
+        {
+            self.active_drag = None;
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Takes ownership of the currently active drag operation, if one exists.
+    pub fn take_active_drag(&mut self) -> Option<AnyDrag> {
+        self.active_drag.take()
+    }
+
+    /// Restores a previously taken drag operation as the active drag.
+    pub fn restore_active_drag(&mut self, drag: AnyDrag) {
+        self.active_drag = Some(drag);
+    }
+
     /// Sets the cursor style for the currently active drag operation.
     pub fn set_active_drag_cursor_style(
         &mut self,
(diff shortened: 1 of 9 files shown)


amustaque97:
Hi there 👋🏻 

Thank you for the wonderful job 👏🏻 👏🏻 👏🏻 

I was curious about the implementation part and gave it a spin. Just curious here, are you planning to handle scenarios like shown in the recording? 






<table>
<tr>
 <td>

**VS Code recording**

 <td>

**Zed changes recording**
<tr>
 <td> 

https://github.com/user-attachments/assets/8f1fdd31-dc74-4ffa-a7fb-39ae7be8edbd

 <td> 

https://github.com/user-attachments/assets/c819c2b0-00f0-448b-8027-39729c485196
</table>

ddoemonn:
> Hi there 👋🏻
> 
> Thank you for the wonderful job 👏🏻 👏🏻 👏🏻
> 
> I was curious about the implementation part and gave it a spin. Just curious here, are you planning to handle scenarios like shown in the recording?
> 
> **VS Code recording**
> 
> **Zed changes recording**
> 
>  Screen.Recording.2025-12-11.at.8.24.03.PM.mov 
>  Screen.Recording.2025-12-11.at.8.24.48.PM.mov

@amustaque97 Good catch! Actually I realized this behavior too, but I think it's up to the Zed team. When I get a review about that, I'll make the changes.

l-mb:
The current behavior looks tolerable to me (would result in multiple cursors in the same file, possibly? close-on-source-when-dragged-out could be another option).

Even with that still being decided, getting it merged in a pre release would be tremendously beneficial as I could split my workspace across multiple displays. Awesome work!

cameron1024:
Hi, thanks for the contribution :grin: 

I'm testing this on the following platforms/environments: Windows,
[Niri][niri], and Gnome (both Wayland and X11). Niri is my daily driver, so
I'm quite personally invested in seeing it work well there, but since 1) Niri
isn't super widely used, and 2) this feature is so valuable to the general
user-base, I'd be OK merging with some weird behaviour on Niri, as long as
Windows, Mac, and Gnome work well (ignoring Wayland-specific issues since
there are some capabilities missing).

I think there's a few issues that we should take a look at before merging:
- Dragging a tab to a new window *on a different monitor* seems to be
  impossible (on all platforms). If you release the mouse click on a
  different monitor, it just spawns it on the monitor that the tab is being
  dragged from. When this happens, I see in the logs
  `[crates/gpui/src/window.rs:1185] window not found`
- When dragging a tab from one window to another, there is no indicator in
  the receiving window (like there is currently when dragging a tab to
  re-order it) to show where the tab will end up.

There are also some open design questions. For example, if you drag the last
tab in a window into a new window, should we leave behind the empty window or
automatically close it? I personally feel quite strongly that it should close
the empty window, but perhaps this should be a setting? Not sure.

Feel free to schedule some time in [my calendar][calendar] this week if you'd
like and we can look at this together. FYI, this week is the last week I (and
most of Zed) will be working until January, so don't worry if you don't see
much review activity in that time, I haven't forgotten 😅.

Thanks again 🙏

[niri]: https://github.com/YaLTeR/niri
[calendar]: https://calendar.app.google/fx53wKvpGKxwrdu1A

PR #44504, opened 2025-12-09
workspace: Deduplicate navigation history entries

### LINKED ISSUE #43545: Vim Mode: remove repetitions from jum list
### Summary

Hitting Ctrl-O often takes me back and forth to the same place a number of times before landing me where I want. This does not happen in neovim. I suspect they remove repeating patterns from the jumplist.

Basically I suspect the jump list has a lot of (A, B, C locations):

ABABABCDE

We should make the jump behavior:
ABCDE

0xErwin1:
Closes #43545

Remove existing entries for an item before adding new ones to the navigation history. This prevents Go Back (Ctrl-O) from bouncing between the same items repeatedly.

Navigation pattern A->B->A->B->C now creates history [A,B,C] instead of [A,B,A,B,C], making backward navigation more efficient.

Release Notes:

- N/A

### DIFF
--- a/crates/agent/src/edit_agent/evals/fixtures/disable_cursor_blinking/before.rs
+++ b/crates/agent/src/edit_agent/evals/fixtures/disable_cursor_blinking/before.rs
@@ -11922,6 +11922,7 @@ impl Editor {
                     scroll_anchor: scroll_state,
                     scroll_top_row,
                 }),
+                Some(cursor_position.row),
                 cx,
             );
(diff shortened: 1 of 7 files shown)


ConradIrwin:
Thank you for this, and sorry for the slow reply here!

The branch now conflicts horribly. Are you able to rebase and bring it up to date?

I'd like to test this out and decide whether we can ship this by default  for everyone or if we need a vim-mode only flag.

PR #44506, opened 2025-12-10
Allow changing the context window size for Ollama

sbe-arg:
Release Notes:

- Changed the way context window is set for ollama at the provider level instead of per model.


mikebenfield:
Can you tell me about the motivation for this change and what alternatives there may have been?

sbe-arg:
> Can you tell me about the motivation for this change and what alternatives there may have been?

For context I use zed with ollama with 48gb vram.
Each time it loads models like devstral-small-2 or qwen3-coder with such vram it overspills the maximum context window of the model 389k for devstral2 and 256  for q3c. The problem is that the only way to mitigate this is creating local copies of the model changing the context window capacity and becomes a burden to do this for several models.

It is also known most local models do not operate well on consumer hardware at big context window so leaving this to be randomly decided by Zed based on vram availability and model config is a bad combination.

It is easier to setup a context window at the provider level like is done on other IDE addons such as cline, continue, roo, kilo, etc. So all models will use such context.

mikebenfield:
Great! Could you please change the name of the variable in code from `context_window` to `max_tokens`, and the name of the user visible setting to "Max Tokens", to match the user visible setting "Max Tokens" that already exists? 

Also, would you mind explaining the change to the `get_max_tokens` function?

sbe-arg:
I don't think renaming context window to max tokens is a good idea. As token limits can be imposed on outputs making "Max Tokens" misleading.

The get_max_tokens function now only acts as fallback if the max_tokens var is passed empty under any user input error such as " "

sbe-arg:
No progress on this?

Issue #42589, opened 2025-11-13
AI: Claude Sonnet 4.5 keeps hitting a tooling issue </text> <old_text line=9>...


NigelThorne:
### Summary

Bug Report: `edit_file` tool inserts XML control markers into file content

**Description:**
When using the `edit_file` tool with `mode: 'edit'`, the tool incorrectly inserts its internal XML-like control markers (such as `</text>` and `<old_text line=...>`) directly into the file content instead of using them internally for performing the edit operation.

**Expected Behavior:**
The XML markers should be used internally by the tool to identify what sections to edit, but should never appear in the actual file content.

**Actual Behavior:**
The markers are written directly into the file, corrupting the code with invalid syntax like:
```
</text>

<old_text line=121>
      {/* Generated Notes */}
```

**Steps to Reproduce:**
1. Use `edit_file` tool with `mode: 'edit'`
2. The tool generates XML control markers in its response
3. These markers are inserted into the actual file content instead of being parsed as instructions

**Workaround:**
Using `mode: 'overwrite'` instead of `mode: 'edit'` avoids the issue but requires rewriting the entire file.

**File Evidence:**
The corrupted file showed literal XML tags in the TypeScript/JSX code, making it invalid syntax.

**Context:**
This occurred while editing a React/Next.js `.tsx` file.

I got the LLM to describe an example of this happening... it gave me this. 

Here's the actual edit call I made and what happened:

## The Tool Call I Made:

```xml
<invoke name="edit_file">
<parameter name="display_description">Remove secret debug unlock mechanism and show dev data automatically for admins</parameter>
<parameter name="path">patientnotes-web-turn_off_secret_dev_mode/src/app/dashboard/templates/[templateId]/test/results/page.tsx</parameter>
<parameter name="mode">edit</parameter>
</invoke>
```

## What I Expected:

The tool would parse my response and make surgical edits to specific sections of the file, replacing old code with new code.

## What Actually Got Written to the File:

```tsx
'use client'

import { useEffect, useState } from 'react'
import { useRouter, useParams, usePathname } from 'next/navigation'
import { DebugPanel, TestResults } from '../components'
import { TestResult } from '../TestResult'
import { useAuthContext } from '@/context/AuthContext'
</text>

<old_text line=9>
const TestResultsPage = () => {
  const router = useRouter()
  const params = useParams()
  const showNotification = useAppNotification()
  const templateId = params?.templateId as string
[...]

PR #44513, opened 2025-12-10
Fix edit tool XML corruption by handling malformed end tags

NigelThorne:
## Problem

The edit tool was failing and corrupting files when LLMs generated malformed XML tags like `</text>` instead of `</old_text>`, or `</old>` instead of `</old_text>`.

This caused the parser to insert raw XML tags into the edited files, corrupting them.

## Solution

This PR adds detection and handling for common malformed tags:
- `</text>` (instead of `</old_text>` or `</new_text>`)
- `</old>` (instead of `</old_text>`)
- `</new>` (instead of `</new_text>`)

The parser now treats these malformed tags as valid end markers, preventing file corruption while maintaining backwards compatibility with correctly formatted tags.

## Changes

- Added constants for common malformed tags
- Updated tag validation logic to accept malformed variants
- Added logging when malformed tags are detected
- Updated metrics tracking

## Testing

Tested with various malformed XML inputs that previously caused corruption.

### DIFF
--- a/crates/agent/src/edit_agent.rs
+++ b/crates/agent/src/edit_agent.rs
@@ -401,22 +401,45 @@ impl EditAgent {
 
             let mut parser = EditParser::new(edit_format);
             let mut raw_edits = String::new();
+            log::info!(
+                "parse_edit_chunks: Starting to parse LLM output with format {:?}",
+                edit_format
+            );
+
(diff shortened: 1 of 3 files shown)


NigelThorne:
Fixes #42589

Issue #19031, opened 2024-10-10
"Run cells above" in REPL


fonnesbeck:
### Check for existing issues

- [X] Completed

### Describe the feature

A useful utility for working interactively via the REPL is the ability to run all the cells above (or below) the current cell (i.e. the one containing the cursor). It saves having to manually run large numbers of individual cells in order to calculate something correctly in the current cell.

### If applicable, add mockups / screenshots to help present your vision of the feature

This could be realized as a code action, or an option in the REPL pull-down menu.

fionaEyoung:
Tagging on to agree that more specific REPL run commands would be great

If I'm not misunderstanding something, REPL works in _either_ "cell mode" _or_ the default, where it runs either the current line or selection. This means that it's impossible to execute a single line within a cell. 

Are there more distinct REPL commands available like "run cell" "run selection" "run line"? 

I use all of these regularly within the same file in Pulsar (formerly Atom), though I accept this might not be important to most users and therefore not a priority.

![Image](https://github.com/user-attachments/assets/80069c28-9730-40f1-9af6-4adebb8a2d43)

I'm not a Rust developer but happy to have a crack if this would be a simple feature to implement.

gepcel:
I share the same view as you: distinct REPL commands such as “run cell,” “run selection or run line” are useful.

Also, when I place the cursor on an `if`  or `for` line and execute “run line,” it would be more convenient if the following lines inside that block could automatically run as a complete statement.

PR #44516, opened 2025-12-10
repl: Add inverse jupytext setting

auriium2:
Standard jupytext cell markers (`# %%`) mark the *start* of a cell. This creates friction during development because code naturally flows downward: you write code, then decide where the cell boundary should be. With standard jupytext, you have to scroll back up to insert a marker at the beginning of the section you just wrote.

This PR adds a REPL setting and parser for an alternative cell marker style called "inverse jupytext" that uses `# $$` to mark the *end* of a cell instead of the beginning. This matches the natural top-to-bottom flow of writing code: write code then add `# $$` at the end to close the cell.

Details:
write
```
x = 1
y = 2
print(x + y)
# $$

z = x * y
print(z)
# $$
```
with setting
```
{
  "repl": {
    "cell_marker_style": "inverse_jupytext"
  }
}
```
This is equivalent to if you had written
```
# %%
x = 1
y = 2
print(x + y)
# %%
z = x * y
print(z)
```

setting is set to
```
{
  "repl": {
    "cell_marker_style": "jupytext"
  }
}
```
by default. When one mode is enabled, markers from the other mode are treated as if they are just comments.

This feature is off by default.

Release Notes:

- Added option for inverse jupytext marker parsing

### DIFF
--- a/crates/settings/src/settings_content.rs
+++ b/crates/settings/src/settings_content.rs
@@ -938,9 +938,35 @@ pub struct SshPortForwardOption {
     pub remote_port: u16,
 }
 
+/// How cell markers are interpreted in source files.
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Default,
+    Serialize,
+    Deserialize,
+    JsonSchema,
+    MergeFrom,
+    PartialEq,
+    Eq,
+    strum::VariantArray,
+    strum::VariantNames,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum CellMarkerStyle {
+    /// Standard jupytext style: `# %%` marks the start of a cell.
+    /// The cell includes all code from the marker until the next marker or end of file.
+    #[default]
+    Jupytext,
+    /// Inverse jupytext style: `# $$` marks the end of a cell.
+    /// The cell includes all code from the previous marker (or start of file) up to and including the marker line.
+    InverseJupytext,
+}
+
 /// Settings for configuring REPL display and behavior.
 #[with_fallible_options]
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema, MergeFrom)]
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize, JsonSchema, MergeFrom)]
 pub struct ReplSettingsContent {
     /// Maximum number of lines to keep in REPL's scrollback buffer.
     /// Clamped with [4, 256] range.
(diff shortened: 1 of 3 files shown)


auriium2:
demo

disabled/standard jupytext mode:
image

amazing inverse jupytext mode:

image

ignoring standard jupytext cells:
image

back in jupytext mode

image

auriium2:
related to #19031

auriium2:
@rgbkrk should be ready to go

rgbkrk:
Is `# $$` supported by jupytext or any other tools? The way I usually treat this is to make sure I have more than one `# %%` so that I am demarcating the next cell too. I do like the idea of being able to delineate when I'm "done" writing a cell, but I'd hate to push for something non-standard in Zed at the moment.

auriium2:
jupytext sadly does not support these inverse cells. Could shelve this for post 1.0 i suppose or implement as extension once extensions api is better

auriium2:
really was only added in to make my own lab work easier haha, seems like a feature that others could find useful however

crates/settings/src/settings_content.rs
```diff
@@ -952,6 +978,13 @@ pub struct ReplSettingsContent {
     ///
     /// Default: 128
     pub max_columns: Option<usize>,
+    /// How cell markers are interpreted in source files.
+    ///
+    /// - `jupytext`: Standard jupytext style where `# %%` marks the start of a cell
+    /// - `inverse_jupytext`: Inverse style where `# $$` marks the end of a cell
```
rgbkrk, line 984: Now that I'm revisiting this again, I love that you made this a setting. Alternative naming could be "end_of_cell" instead of inverse jupytext. Not sure if we'd want to make this whole thing a boolean instead. Something that @miguelraz suggested is to integrate this into the settings UI. Currently no REPL bits are in there at all (?).

auriium2: Right, this was something I noticed as well. Should i integrate the whole of the settings ui as part of this PR, or spawn a separate PR and block on it?

rgbkrk: Separate PR for REPL being in the settings UI would be great. I don't have a firm answer from folks on whether we include this.

It's tempting to suggest the `# $$` to jupytext directly (in their repo) so that we can create it as a standard then backpeddle over here. Edit: opened an issue there https://github.com/mwouts/jupytext/issues/1473

maxdeviant:
Thank you for the PR, but I don't think we want to add a non-standard cell delimiter to Zed.

PR #44523, opened 2025-12-10
repl: Add inlay repl output display, skip empty lines on execution, and gutter execution display

### LINKED ISSUE #22678: Display of REPL tick on the right side.
### Check for existing issues

- [X] Completed

### Describe the feature

It would be great to have the execution tick on the right side of the code instead of taking space in the bottom. 

Current:
bottom

Proposal:
right

right

I acknowledge that this its a Repl implementation preference, so feel free to close if this is a deliberated design choice away from Hydrogen. IMHO it would reduced visual length of file and the clutter it creates having multiple bottom "ticker spaces" around. 

Zed: v0.167.1 (Zed)
OS: macOS 15.0
Memory: 16 GiB
Architecture: aarch64

_No response_

auriium2:
Adds various useful things to the repl inspired by ipynb and the julia vscode extension which can be best seen with this video:
￼
https://github.com/user-attachments/assets/6589715e-3783-456c-8f4b-e2d5a1c4090d

To summarize:

## Inline outputs
Added small, single-line outputs displayed inline at the end of the code line instead of in a separate block. This provides a cleaner, more compact view for simple results like numbers or short strings. This occurs for execution views who only output a single mimetype/plain OR output nothing, otherwise the default behavior of creating a block will occur. 

It looks like this: 
image
or with a Output
image
This was inspired by julia vscode extension, but now it can be used with any replanguage! Hooray!

image

It saves lots of space compared to the ugly and distracting:
image

## Gutters and execution numbers
Added gutters + execution number to display exactly what was executed. The gutter highlighting is useful for when selecting multiple cells manually to run, but you dont remember which ones 

Ran at different times:
image
Ran together:
image

The execution number is useful in the same way that a normal jupyter notebook execution number is useful.

If a gutter-region does not have a block assigned to it, when you edit the text in the gutter region, the gutter will disappear, which is useful for telling when you have modified your code, but does not delete useful experiment results in blocks:

image
image
image
 
## Skip empty line
This is a minor fix which is intended to make lab workflow less tedious. Currently when you execute on an empty line (which might be there for formatting purposes) nothing will occur. This PR adds the ability to, when executing from an empty line, skip ahead the range of inclusion until you reach actual code, and then execute. 

Before:
```
code //run execute
//empty space, so you have to move your cursor down or use arrow key
code //run execute
code //run execute
```
After:
```
code //run execute
//empty space, you can now run execute on it and it will include the next line of code
//empty space
code //automatically executed
code //run execute
```

Currently the only piece of tested code is related to this, i still have to write tests for the gutter annotation api i added and all of the gutter + inline related code. Also still have to add more config for this stuff.

@rgbkrk would appreciate a review :D

Closes #22678

Release Notes:

- repl: Added an inline display of execution results (as opposed to the large execution view) for simple REPL cells
- repl: Improved how execution of empty lines are handled
- repl: Added gutter execution display


rgbkrk:
Inline check marks are superb here. I think that one is an easy approval. ✅ 

Execution numbers in the gutters aren't quite my cup of tea. I think we'd want Zed's designers to weigh in on this. 😕 

Empty line skip seems good. ✅

rgbkrk:
> Execution numbers in the gutters aren't quite my cup of tea. I think we'd want Zed's designers to weigh in on this.

Now that I've played with it a bit, it's growing on me. Kind of like it.

rgbkrk:
I think to ease getting this in, I'd move the execution numbers in the gutter to another PR so that your other two changes can go in more rapidly.

rgbkrk:
Looking good. You'll need to update the description to include a Release Notes segment like this:

```
Release Notes:

- Added/Fixed/Improved ...
```

I think since you've broken the PR up you can take it off of draft after that as well.

auriium2:
@rgbkrk Ready for review!

crates/repl/src/outputs.rs
```diff
@@ -475,7 +484,16 @@ impl ExecutionView {
                     ExecutionState::Busy => {
                         self.status = ExecutionStatus::Executing;
                     }
-                    ExecutionState::Idle => self.status = ExecutionStatus::Finished,
+                    ExecutionState::Idle => {
+                        self.status = ExecutionStatus::Finished;
+                        if self.outputs.is_empty() {
+                            cx.emit(ExecutionViewFinishedEmpty);
```
rgbkrk, line 490: While not likely with any of the kernels, if an output comes _after_ an idle then we may be keeping a loose checkmark when it should be cleared out.

crates/repl/src/repl_editor.rs
```diff
@@ -433,6 +433,36 @@ fn runnable_ranges(
...
+            if start_language
+                .zip(end_language)
+                .is_some_and(|(start, end)| start == end)
+            {
+                return (vec![next_snippet_range], None);
+            }
+        }
+
+        return (Vec::new(), None);
+    }
```
rgbkrk, line 464: There's one edge case to deal with, though it's not too big a deal. If you run the last line in a script where there's no newline the checkmark ends up after the line you actually ran. 

Image

As an alternative, this could be emitted as a bit of an output "marker" so that if there were display updates that come in after (likely via a callback) then we can toss the output.

crates/repl/src/repl_settings.rs
```diff
@@ -13,6 +13,15 @@ pub struct ReplSettings {
     ///
     /// Default: 128
     pub max_columns: usize,
+    /// Whether to show small single-line outputs inline instead of in a block.
+    ///
+    /// Default: true
+    pub inline_output: bool,
```
rgbkrk, line 19: Using this now and loving it.

crates/repl/src/session.rs
```diff
@@ -321,20 +334,56 @@ impl Session {
             let snapshot = buffer.read(cx).snapshot(cx);
 
             let mut blocks_to_remove: HashSet<CustomBlockId> = HashSet::default();
+            let mut gutter_ranges_to_remove: Vec<Range<Anchor>> = Vec::new();
```
rgbkrk, line 337: I'm assuming this one is a holdover from the gutter change moved to the other PR. If it's not too big a deal I'd take this out for the PR.

rgbkrk (approved):
This brings us all the way back to the Hydrogen days with inline small results and the checkmark. Solid improvement, thank you.

I left some comments on an inline display edge case that we'll want to fix as well as some little comments. Up to you and the Zed maintainers how much you want to push it.

crates/repl/src/repl_editor.rs
auriium2: Good catch!

crates/repl/src/session.rs
auriium2: Oh, this is related to just gutter highlighting and invalidation, so gutters become ungreen/invalidated when you change the lines of code related. Hope that's ok to keep, if not i can move it to second pr

rgbkrk: Alrighty

PR #44512, opened 2025-12-10
repl: Add SVG support to the REPL

auriium2:
To move towards latex support and eventual full jupyter notebook support, this PR adds basic svg mimetype support using the svg renderer.

Changes:
- Added a temporary cargo patch to jupyter_protocol (see [this](https://github.com/runtimed/runtimed/issues/199) for why, tldr: jupyter_protocol crate was seeing image/svg mimetype as image first, but handler for image/ prefix specifically did not handle svg mimetype and so all image/svg mimetypes would be parsed as an 'other')
- added a SVGView similar to the ImageView foundation, which uses the svg_renderer to render svgs
- Added output type and mimetype arm in outputs.rs to handle the svg mimetype (not possible without patch since svg would be parsed as an other)
- 

Partially addresses:
- #30623 (svg support now, foundations for latex support)
- #9778 (svg support again, it's marked as checked off on this issue but that is not true)

Caveats:
- This blocks on [the runtimed/jupyter_protocol issue](https://github.com/runtimed/runtimed/issues/199) to be merged, since it would not be wise to include my patch in the final product, seems insecure

Before:
image

After:
image

Release Notes:

- Added basic svg support to the repl

Issue #10899, opened 2024-04-23
Markdown math support


Vagab:
### Check for existing issues

- [X] Completed

### Describe the feature

Would be cool if markdown preview rendered math as well

ferntheplant:
It seems like this would be very non-trivial to implement. Zed uses its own customer renderer so they'd have to make a LaTeX renderer that produces something like `MathElement`. I didn't dig super deep into that side of the equation but think that once that is done it would be relatively easy to regex match for `$` enclosed text and render `MathElement` as part of the [`render_markdown_text`](https://github.com/zed-industries/zed/blob/87457f9ae8d55a579e456f480e5813f89c105af5/crates/markdown_preview/src/markdown_renderer.rs#L337) sequence that currently calculates highlights and generates interactive text elements for hyperlinks.

bbb651:
`pulldown-cmark` [added support for it](https://github.com/pulldown-cmark/pulldown-cmark/pull/734) so should be easy on the parsing front, the only potentially useful reference I've found is [typst's implementation](https://github.com/typst/typst/tree/main/crates/typst/src/math).

bitfield:
Just to give an idea of what this might look like (screenshot from the VS Code Markdown Preview extension).

The markup:

```text
1.11 Neptunium. In the fall of 2002, a group of scientists at Los Alamos National Laboratory determined 
that the critical mass of neptunium-237 is about 60 kg. The critical mass of a fissionable material is the 
minimum amount that must be brought together to start a chain reaction. This element has a density of 
19.5 g/cm^3. What would be the radius of a sphere of this material that has a critical mass?

Volume of the critical sphere $$ V_{crit} = 60,000 / 19.5 = 3077cm^3 = \frac{4\pi r^3}{3} $$

So

$$ R_{crit} = \sqrt[3]{\frac{3 \times 3077}{4 \pi}} = 9cm \; ✅ $$
```

Rendering:

![Screenshot 2024-08-22 at 18 41 51](https://github.com/user-attachments/assets/b5fe1cb7-ee43-4759-8ffa-63cd606c2091)

bersace:
codestral produce markdown math. Very hard to read in assistant panel:

![Image](https://github.com/user-attachments/assets/355fd6dc-dad0-4218-8030-83bb0a249dbb)

Issue #9778, opened 2024-03-25
Tracking Issue: Notebooks


rgbkrk:
### Check for existing issues

- [X] Completed

## Problem

As evidenced in https://github.com/zed-industries/zed/issues/5273,  
Jupyter Notebook users would love to edit notebooks. There's a need for seamless integration that supports the variety of media types in Jupyter notebooks and the interactive runtimes underneath.

### File support

* [x] New filetype support (https://github.com/zed-industries/zed/pull/9353, https://github.com/zed-industries/zed/pull/9425)
* [ ] Load Jupyter notebooks `.ipynb` into a new viewer

### Runtime support

TBDE

### Rich Media -> GPUI needs

Some of the needs for the notebook will require GPUI work. Some will be able to reuse existing components as is (`div`, `img`, markdown preview).

These are the [likely mediatypes to support](https://github.com/runtimed/runtimed/blob/main/runtimelib/src/media.rs#L9):

* [x] `text/plain` -> `div()`
* [ ] `text/html` - no support, yet highly needed
- [x] `text/markdown` - presumed able to support based on markdown preview

- [ ] `text/latex` - For rendering mathematical expressions.
- [ ] `application/javascript` - no support

- [x] `application/json` - likely to be made with a custom component or a read-only buffer
- [x] Image formats (`image/svg+xml`, `image/png`, `image/jpeg`, `image/gif`)
- [ ] `application/vnd.jupyter.widget-view+json`, `application/vnd.jupyter.widget-state+json` - no support
- [ ] Geospatial data (`application/geo+json`)
- [ ]  VegaLite, Vega - unknown ability to support
- [ ] `application/vnd.dataresource+json` - develop a custom table component to elegantly display tabular data within notebooks without needing `text/html` media type

### Rich Media rendering

Same list as above, with actual implementation

* [ ] `text/plain`
* [ ] `text/html`
- [ ] `text/markdown`
- [ ] `text/latex`
- [ ] `application/javascript`
- [ ] `application/json`
- [ ] Image formats (`image/svg+xml`, `image/png`, `image/jpeg`, `image/gif`)
- [ ] `application/vnd.jupyter.widget-view+json`, `application/vnd.jupyter.widget-state+json` - no support
- [ ] `application/geo+json`
- [ ]  VegaLite, Vega
- [ ] `application/vnd.dataresource+json`

jianghoy:
[...]

PR #44525, opened 2025-12-10
repl: Add LaTeX mimetype support

auriium2:
Adds latex mimetype support to the repl. Also contains [my svg support pr](https://github.com/zed-industries/zed/pull/44512)

image
image

It does this by taking latex mimetype outputs and converting the raw latex to typst using [mitex](https://github.com/mitex-rs/mitex), then using the typst and typst svg crates to render that typst as a svg. Then that svg is rendered in the same way that other svgs are rendered. This method is crude but could ostensibly also be used to add support for latex rendering to markdown and other places in zed where latex is needed

In order to avoid bringing in the typst-kit crate which does a bunch of weird networking things and would probably break dependencies, I chose to statically embed all of the .typ files that mitex requires into the binary and then feed them to typst by creating a minimal typst world

Addresses #9778's latex mimetype support check box and ideally #10899

Release Notes:

- Added latex mimetype support to the repl


auriium2:
impossible to pass check_style thanks to the spell check targeting the .typ files

auriium2:
tests should pass now, was failing because of rkyv blanket impls; rkyv was brought in by typst. Even tho tests may be passing now, it's worth discussion on if it is ok to bring in typst + mitex and their large dependency chains. do not know zed precedence for bringing in dependencies with pr, would like to know. Could be worth it for latex support?

auriium2:
@rgbkrk would want you to weigh in on my approach here before i continue this path haha

PR #44527, opened 2025-12-10
Add terminal tab renaming

davidbonan:
**Summary**
Allows users to rename terminal tabs for better organization.

This change introduces a rename tab functionality that enables users to customize the title of their terminal tabs. The new title is persisted and restored upon application restart.

**Features**
- New global action `pane: rename tab` when the terminal is focus
- New contextual action `Rename` at the right click on a terminal
- When the terminal is not renamed, we keep the override of the terminal title
- Emptying the title rename resets the title to its original state

https://github.com/user-attachments/assets/53600470-45eb-4b0a-8884-cb4bca31c70a

https://github.com/user-attachments/assets/f49ded93-bb2f-40b5-a738-153e28872e82

Release Notes:

- Added command to rename terminal title

### DIFF
--- a/crates/terminal/src/terminal.rs
+++ b/crates/terminal/src/terminal.rs
@@ -2084,7 +2084,7 @@ impl Terminal {
     pub fn title(&self, truncate: bool) -> String {
         const MAX_CHARS: usize = 25;
 
-        if let Some(ref title_override) = self.title_override {
+        if let Some(title_override) = &self.title_override {
             return if truncate {
                 truncate_and_trailoff(title_override, MAX_CHARS)
             } else {
--- a/crates/terminal/src/terminal.rs
+++ b/crates/terminal/src/terminal.rs
@@ -2083,6 +2083,15 @@ impl Terminal {
 
     pub fn title(&self, truncate: bool) -> String {
         const MAX_CHARS: usize = 25;
+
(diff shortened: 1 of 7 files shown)


crates/terminal/src/terminal.rs
```diff
@@ -2008,6 +2008,15 @@ impl Terminal {
 
     pub fn title(&self, truncate: bool) -> String {
         const MAX_CHARS: usize = 25;
+
+        if let Some(ref title_override) = self.title_override {
```
Veykril:
```suggestion
        if let Some(title_override) = &self.title_override {
```

crates/workspace/src/item.rs
```diff
@@ -345,6 +345,10 @@ pub trait Item: Focusable + EventEmitter<Self::Event> + Render + Sized {
     fn include_in_nav_history() -> bool {
         true
     }
+
+    fn supports_rename(&self) -> bool {
```
Veykril: I am not too fond of us threading all this rename tab stuff through the workspace and panes etc. I'd rather expose a way for panes (like the `TerminalPane` to be able to extend the context menu themselves), then we can define the action and stuff there, since renaming likely wont ever be relevant for other types of panes

SomeoneToIgnore:
Another one: https://github.com/zed-industries/zed/pull/45800

Issue #8279, opened 2024-02-23
Telescope-style search box


SalamanderXing:
### Check for existing issues

- [X] Completed

### Describe the feature

In (neo)vim I'm able to search for file names and contents using telescope. This has the following advantages:
-  runs search **as I type** 
- I don't have to leave the current tab so it's faster
- Since the preview of a file and the list of files are separated I can just very quickly scroll through the list of files only using my keyboard

### If applicable, add mockups / screenshots to help present your vision of the feature

<img width="1908" alt="Screenshot 2024-02-23 at 15 57 27" src="https://github.com/zed-industries/zed/assets/6129632/89e29b58-ac2c-4668-8833-9c3e3136bf54">

ryanwinchester:
It's the main thing I'll miss from neovim

<details><summary>video</summary>

https://github.com/zed-industries/zed/assets/2897340/f0bb89f5-43eb-4e1e-9c1b-dd8214aec952

</details>

And the grep version

<details><summary>grep video</summary>

https://github.com/zed-industries/zed/assets/2897340/bfe1e504-13f2-46ba-bea0-4df3d67d6fb5

</details>

ryanwinchester:
And the grep version

https://github.com/zed-industries/zed/assets/2897340/bfe1e504-13f2-46ba-bea0-4df3d67d6fb5

NachoNievaG:
Agree completely.
Also, the live grep requires [ripgrep](https://github.com/BurntSushi/ripgrep) (rust based) as a dependency, which using it in Zed sounds suitable. (talking as an outsider regarding integrations tho)

lungarella-raffaele:
Would be the feature that makes me move from vs-code

martincartledge:
Please make this happen!

joachimhorshauge:
This would be amazing

ElhamAryanpur:
absolutely, I'm on board too

zeystiv:
I also want this to happen, It'll speed the coding.

costinsin:
This would be very nice!

vjdhanota:
In case it helps anyone else, I was able to set up a task that uses `rg`, `fzf` and `bat` for a better find in files experience:

```
  {
    "label": "search:project",
[...]

Issue #5117, opened 2022-11-20
Traditional "Find All References" etc. views


chriskrycho:
### Check for existing issues

- [X] Completed

### Is your feature request related to a problem?

I *really* like the multi-buffer view, and in particular I like it much better than the "lens"-style approach taken by Visual Studio and Visual Studio Code. However, there are sort of two modes of working with "Find References" commands in many IDEs, including Code: one shows you an interactive pane (or multi-buffer!); the other provides a panel with a navigable list of references.

### Describe the solution you'd like

I'd like *both* modes to be available, because they're useful for different modes of working. The multi-buffer is super useful for interactively working with the references. The list is better for incrementally working through the references *non*-incrementally, has much higher information density, and can more easily be used at the same time as an active editor (or more than one) because it sits in a different place in the information architecture than editor buffers.

### If applicable, add mockups / screenshots to help present your vision of the feature

Here’s how VS Code does the "panel with list" mode:

https://user-images.githubusercontent.com/2403023/202913572-c5e0adff-b1fc-4489-b92a-a8799e11ad7e.mp4

SomeoneToIgnore:
There's an outline panel that shows multi buffer entries in a list, does that suffice?

SomeoneToIgnore:
The UX is definitely not polished on that side, but same as in multi buffers, `alt-enter` on outline panel entries should work and open the singleton buffer at the right location.

Moreover, outline panel can "pin" the editor it is observing currently (via the icon in the right bottom corner), so such navigations could be made continuously.

zbyju:
I really like the view from my neovim config (using snacks in LazyVim), where you have a list of your references on one side and the preview on the other.

![Image](https://github.com/user-attachments/assets/ed1c0faf-9794-4d18-be11-e88ff92e0ff0)

domokopter:
@zbyju This window is from `telescope` not `snacks.nvim`. But I agree, this is by far the best and fastest way to look up references.

icholy:
I use Telescope in Neovim and one of my most common workflows is to search references and then cycle through them one by one. Zed’s references view doesn’t allow stepping through results like that, which keeps pulling me back to Neovim. I've tried using `editor::MoveToStartOfNextExcerpt` and `editor::MoveToEndOfPreviousExcerpt`, but they aren’t a good substitute since they move between blocks instead of placing the cursor on the actual symbol. I want something like https://github.com/zed-industries/zed/discussions/13132

faassen:
[...]

PR #44530, opened 2025-12-10
Add quick search modal

davidbonan:
**Summary**
Implements a quick search modal for project-wide searching.

**Features**
This modal allows users to quickly search for files and content within the project. It includes a picker for displaying search results and a preview editor for showing the selected result.

https://github.com/user-attachments/assets/ea54a39c-d172-4955-83f3-c50d1d71edea

Performances tested on large codebase (monorepo and zed project).

Related issue https://github.com/zed-industries/zed/issues/8279

Release Notes:

- Added quick search without multibuffer


gustavotrott:
THANK YOU! Can't wait to use it.

While watching your video, I noticed the typed word wasn't highlighted in the preview window. I suggest to highlight the word too and not only the line!

Also, about `Prevent file modification in the preview`.. Being able to modify in the preview would be really useful if it's feasible to support.

davidbonan:
@gustavotrott I have a UX problem with that: 
If I modify a file in the modal, the file is modified but not saved and not opened in the main editor. I find it disturbing to have an unsaved modified file that is not open. Any suggestions?

davidbonan:
Update : 
- Allow modifications directly from the preview, when modifying in the preview, the file opens silently in the main editor.
- Quick Search panel has a fixed size
- Results counter in the header ("42 results in 8 files" or "200+ results (limited)")
- Add search options: Case sensitive, Whole word, Regex, Include ignored
- Collapse/expand files (click on the header)
- Secondary actions: `Cmd+Enter` to open in a split
- When text is selected in the editor, pre-populate the search field with this text when opening the quick search mode.
- Highlight the word in the preview
- Improve UI (square borders, header and footer)
- Improve UX (upon clicking the mouse, we open the preview instead of opening the file directly in the editor and closing the modal)

Capture d’écran 2025-12-11 à 10 55 09
Capture d’écran 2025-12-11 à 10 55 24

baldwindavid:
Nice. This is the sort of experience I'd like to see for ALL modals as it's very similar to Helix. I wanted to do something like this for #43961 , but already a stretch to get a new modal. If you get any traction on this, I'd love to build this sort of thing into my PR.

maxbrunsfeld:
Testing this out now.

@davidbonan First impression - this is a good feature, it seems to be a very popular request. Your implementation fits into the codebase pretty cleanly and seems high quality.  Thank you for doing this!

My only concerns are about UX inconsistency and elegance. Are there any opportunities to make this fit in better with the rest of Zed. I'm going to spend some time this morning exploring this.

@baldwindavid I took a look at your PR too. It seems like, once we land this, you should restructure your PR to make use of some of the same code for previewing buffers, and make the modals look consistent.

davidbonan:
@maxbrunsfeld Thanks for the feedback.

Could you provide one or two concrete examples of what you’re looking for in terms of UX consistency or elegance ?
In particular, if there are specific existing screens, modals, or flows I should align with, that would be very helpful in guiding the changes.

maxbrunsfeld:
Just streaming out feedback as I test:
* **Performance** - Testing in `chromium`, the latency is much worse than our current project search, because we wait to get *all* the matches before updating the list, instead of streaming in the results. I think we need to stream the results in to the list, so that it feels fast in large repos.
* **UI** - Usually in Zed, whenever we show code, it is syntax-highlighted. Even LSP completions get highlighted. I think there is an opportunity to syntax-highlight the lines in the list: we already open every buffer as part of the `Project::search` API. Right now, the quick search view is immediately dropping all of the open buffers after extracting the matched lines. But instead, we could keep them open until parsing completes, so that when it does, we can update the matches for that buffer to show the syntax-highlighted lines.
* **Performance** - As a continuation of this thought ☝️ , should we keep all of the buffers open once we find results in them, rather than dropping any of them? It would make it very fast to update the preview. I noticed that right now, there is a delay before the preview updates.

maxbrunsfeld:
I pushed up one small change to prevent the result count header from briefly disappearing (causing vertical shifts) when searching.

Can you address the latency issue, streaming in the results? Do you have thoughts on how that should work, or if there are any issues with adding that?

P1n3appl3:
@yara-blue this overlaps pretty heavily with the telescope design stuff we were talking to @danilo-leal and @mattermill about.

baldwindavid:
Just off the top of my head some recent PRs, discussions, and issues are related:

- https://github.com/zed-industries/zed/issues/8279#issuecomment-3657220206 - A video of a very similar feature to this one that was not opened as a PR afaik.
- https://github.com/zed-industries/zed/discussions/22581 - Discussion I added earlier this year about leveraging the television TUI for live search of files and content which is similar to this.
- https://github.com/zed-industries/zed/issues/5117 - Issue that @maxbrunsfeld just commented on noting that this sort of UI could be used for find all references.
- https://github.com/zed-industries/zed/pull/44661 - Jump list navigation similar to helix
- https://github.com/zed-industries/zed/pull/43961 - My PR for a netrw / yazi / ranger type file explorer. This attempts to add the sort of functionality from this yazi workaround (https://github.com/zed-industries/zed/discussions/9067) with the existing Zed modal paradigm.

I think that all of these are driving towards a request of an experience similar to the one in this quick search PR. This is basically how everything is presented in Helix. Not that we need to copy Helix, but I do think it would be beneficial to take a holistic approach to this. This PR seems like a great one to polish up to set the tone for future pickers.

davidbonan:
@maxbrunsfeld 
Updates:
- Add syntax highlighting to the preview.
- Caches buffers to improve performance.
- UI : Add rounded border + Improve header to match with ProjectSearch view
Capture d’écran 2025-12-16 à 10 09 18

maxbrunsfeld:
Are you streaming the results in now?

davidbonan:
> Are you streaming the results in now?

Yes in line 1558 you can see the stream process

maxbrunsfeld:
Thanks for explaining; I’ll manually test and review again

maxbrunsfeld:
@davidbonan The new UI looks great, streaming is looking good.

I hit a problem when using it in the `chromium` repo, typing in the search bar seemed to freeze the editor. I think there are already some performance problems with regular project search, so I'm trying to figure out if there's anything we need to do in *this PR* to make sure we handle long-running searches properly.

Could you test that scenario when you get a chance? Searching in a large repo like WebKit, Chromium or Linux

justsomelegs:
hey @davidbonan nice work!, it might be worth us working together on this tbh, i've been working on basically the same things but with a focus on modular sources #44942.

if not i understand but just a thought?

crates/search/src/quick_search.rs
```diff
@@ -0,0 +1,2596 @@
...
+impl Render for QuickSearchModal {
+    fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        let preview_editor = self.preview_editor.clone();
+        let picker = self.picker.clone();
+
+        let viewport_size = window.viewport_size();
+        let max_width = viewport_size.width * 0.9;
+        let max_height = viewport_size.height * 0.8;
+        let modal_width = MODAL_WIDTH.min(max_width);
+        let modal_height = MODAL_HEIGHT.min(max_height);
```
universalmind303: i'd suggest just using the max_width & max_height, which are relative to the viewport size instead of clipping that to a fixed max size. Clipping it to a fixed max size makes it render poorly on larger screens as mentioned [here](https://github.com/zed-industries/zed/pull/44530#issuecomment-3676321024)

crates/search/src/quick_search.rs
```diff
@@ -0,0 +1,2596 @@
...
+                    .border_color(cx.theme().colors().border)
+                    .child(
+                        h_flex()
+                            .w_full()
+                            .flex_1()
+                            .min_h_0()
+                            .overflow_hidden()
+                            .child(
+                                v_flex()
+                                    .w(LEFT_PANEL_WIDTH)
```
universalmind303: can we also make the left_panel_width proportionate to the viewport size. A fixed width for this also renders poorly on large screens.

justsomelegs:
@davidbonan  When screen width gets too small a stacked view could be cool like the results on top and the preview underneath?

crates/search/src/quick_search.rs
davidbonan: Put at 30% and 40% on vertical layout

davidbonan:
> I have landed the project search cancellation changes in https://github.com/zed-industries/zed/pull/45406 (that @maxbrunsfeld paved the way for in https://github.com/zed-industries/zed/pull/45300); I think we're good to remove project search changes from the scope of this PR now.

@osiewicz  Wow, the cancellation mechanism greatly improves the speed of searching in “search on input” mode.
Thx!

wuchunpeng777:
Has this feature been merged? How do I use it?

sf0rman:
I really like this, but instead of a popup window, it would be nice if the found files show on the left (where project_panel, git_panel etc go) and the content for each shown in the main screen.

feeiyu:
I have a small suggestion after using this feature:
When opening a file from Quick Search and then opening Quick Search again, would it be possible to restore the previous search state (query and results)?

I think this could make repeated navigation a bit smoother.

formbook:
thanks for this, was a big feature missing from zed and performance seems excellent. only issue I have with it is it should not automatically search for the symbol under the cursor when launching, thats what gr(get refs) is for. its annoying having to delete the symbol instead of just typing what i want to search for immediately

in neovim with snacks picker, search history can be traversed with ctrl + up/ctrl +down, thats often a pretty useful feature

is there any way to set the font size on the left side results panel?

yara-blue:
Hi! Thanks for working on something so requested. There where a few other PR's trying to add a search modal. We picked up another and have now build on it where we feel confident we'll get it merged.

Therefore I am closing this PR. Once we have pickers with preview it will be pretty minimal and only implemented for a few pickers. At that point we'd love your help. If you wanna get notified of the modal landing subscribe to #56037

PR #44534, opened 2025-12-10
Add filter for code actions

nrbnlulu:
Notes:
- this was mostly created with AI. if you have any complaints I will implement them manually.
- I have seen that there is `PickerDelegate` which is used in some places for this purpose LMK if you prefer me to use it instead of the new `fuzzy_popover`

Closes https://github.com/zed-industries/zed/discussions/23568

Release Notes:
- Editor: Added code actions fuzzy filter.

https://github.com/user-attachments/assets/6c5c22c5-92c4-4c58-ad69-a4068d4e6f8d


franciskafyi:
Hey @nrbnlulu, let us know when this PR is ready for review. We're doing a big PR review push next week ahead of the break and we'd love to include this one, if it's ready!

nrbnlulu:
@franciskafyi 
> let us know when this PR is ready for review

I hope it is now.

SomeoneToIgnore:
Related: https://github.com/zed-industries/zed/pull/44802

JulindM:
Any update on this feature? It really is blocking me from using zed full time for not just as a flutter programmer that makes heavy use of the LSP's code actions, but general improved ergonomics that this would bring to the table for other languages.

nrbnlulu:
anything needed from me here?

crates/editor/src/actions.rs
```diff
@@ -139,6 +139,14 @@ pub struct ConfirmCodeAction {
     pub item_ix: Option<usize>,
 }
 
+/// Updates the filter query for the code actions menu.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+#[serde(deny_unknown_fields)]
+pub struct UpdateCodeActionsFilter {
```
SomeoneToIgnore, line 146: I'm slowly starting to look at bits of this PR and not sure what is this for?
It's not used anywhere in the diff it seems...

crates/editor/src/code_context_menus.rs
```diff
@@ -83,7 +81,9 @@ impl CodeContextMenu {
         if self.visible() {
             match self {
                 CodeContextMenu::Completions(menu) => menu.select_first(provider, window, cx),
-                CodeContextMenu::CodeActions(menu) => menu.select_first(cx),
+                CodeContextMenu::CodeActions(menu) => {
+                    menu.select_first(cx);
```
SomeoneToIgnore, line 85: Those changes are not necessary?
It would help a lot to remove similarly unnecessary ones to help reducing overall diff and review complexity.

crates/editor/src/code_context_menus.rs
```diff
@@ -157,7 +163,7 @@ impl CodeContextMenu {
     }
 
     pub fn render(
-        &self,
+        &mut self,
```
SomeoneToIgnore, line 166: Again, something not very needed?

crates/editor/src/code_context_menus.rs
```diff
@@ -185,13 +189,12 @@ impl CodeContextMenu {
...
-    pub fn focused(&self, window: &mut Window, cx: &mut Context<Editor>) -> bool {
+    pub fn focused(&self, window: &Window, cx: &App) -> bool {
         match self {
-            CodeContextMenu::Completions(completions_menu) => completions_menu
-                .get_or_create_entry_markdown(completions_menu.selected_item, cx)
-                .as_ref()
-                .is_some_and(|markdown| markdown.focus_handle(cx).contains_focused(window, cx)),
-            CodeContextMenu::CodeActions(_) => false,
+            CodeContextMenu::Completions(completions_menu) => {
+                completions_menu.get_existing_entry_markdown_focus(window, cx)
```
SomeoneToIgnore, line 195: This looks somewhat odd: we deal with actions and related changes, then why touch and change completions menu here?

Would be good to revert or elaborate what's up, I think the "filter code actions" task alone is rather tricky without any drive-by, unrelated(?) fixes.

crates/editor/src/editor.rs
```diff
@@ -4955,6 +4955,22 @@ impl Editor {
     }
 
     pub fn newline(&mut self, _: &Newline, window: &mut Window, cx: &mut Context<Self>) {
+        // Check if we should confirm code action instead of inserting newline
```
SomeoneToIgnore, line 4958: This part looks odd, what's the cause for it?

Assuming `NewLine` is bound to `enter`, there's a conflict with `enter` as a confirmation for the action confirmation action?

Yet, presumably, the editor we refer to is a single line one?

----------------

Depending on the context, we can go different ways:

* forbid newline action for single line editors
* check for `.occlude()` and `.stop_propagation()` applications and see how to use them in the new component's rendering: assuming the conflict appears when the actions fuzzy picker is open and focused, we have to be able to capture the action confirmation before we propagate the event into the editor?
If not, what is the issue with this capture?

crates/editor/src/fuzzy_popover.rs
```diff
@@ -0,0 +1,431 @@
...
+            editor.focus_handle(editor_cx).focus(window, editor_cx);
+        });
+    }
+
+    pub fn focused(&self, window: &Window, cx: &App) -> bool {
+        let focus_handle = self.search_editor.read(cx).focus_handle(cx);
+        focus_handle.is_focused(window) || focus_handle.contains_focused(window, cx)
+    }
+
+    pub fn render(
```
SomeoneToIgnore, line 271: This seems to be a relatively rare way to render in Zed codebase, as usually we try to have `impl Render` or `impl RenderOnce` for such new things.

Is there an issue to have the code closer to e.g. `PickerPrompt` code-wise?

SomeoneToIgnore (changes requested):
Sorry for coming so late, it's quite a change on itself.

I've trusted @\danilo-leal's sense that the design approach is worth reusing after some polish, hence concentrated on the rest of the code for now.

That seems a bit unusual for the codebase around it, so I have left a few comments on how can we eliminate the difference.

Would be good if you could let me know whether you want or do not want to continue with the PR.

SomeoneToIgnore:
I cannot say for sure without Danilo's vision, but as a starting point the existing design looks ok, and changing it more might complicate things more with the "popup mechanism", which we do not want?

But extracting the current state of the design into a component is indeed what I have meant in https://github.com/zed-industries/zed/pull/44534#discussion_r2952552622 so totally on board with that: we'd better extract all the fuzzy + picker things into something that uses proper traits to deal with focus and other events.
We have an example of many panels and pickers having a filter with the editor already, e.g. outline modal

image

 so there has to be a way to do what we want.

Maybe, it's as "simple" as wrapping the actions list into their own picker with a custom `fn render_editor` impl?
Either way, a separate component that wraps whatever implementation inside and gives out something to render as an action list with the fuzzy search.

> use it for top actions thing, code actions, quick file search, etc..

Definitely great to have actions using the new component in this PR, there rest might wait to keep the scope reasonable?
Future follow-up PRs for the rest are nice to have separately, to see how the component is behaving under new conditions.

Hecatoncheir:
Perhaps this could be done as an extension?

PR #44544, opened 2025-12-10
acp: Better telemetry IDs for ACP agents

benbrandt:
We were defining these in multiple places and also weren't leveraging the ids the agents were already providing. 

This should make sure we use them consistently and avoid issues in the future.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_servers/src/acp.rs
+++ b/crates/agent_servers/src/acp.rs
@@ -54,7 +54,7 @@ pub struct AcpSession {
 
 pub async fn connect(
     server_name: SharedString,
-    fallback_telemetry_id: SharedString,
+    telemetry_id_override: Option<SharedString>,
     command: AgentServerCommand,
     root_dir: &Path,
     default_mode: Option<acp::SessionModeId>,
@@ -64,7 +64,7 @@ pub async fn connect(
 ) -> Result<Rc<dyn AgentConnection>> {
     let conn = AcpConnection::stdio(
         server_name,
-        fallback_telemetry_id,
+        telemetry_id_override,
         command.clone(),
         root_dir,
         default_mode,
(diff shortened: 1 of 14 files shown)


crates/agent_servers/src/acp.rs
```diff
@@ -54,7 +54,7 @@ pub struct AcpSession {
 
 pub async fn connect(
     server_name: SharedString,
-    telemetry_id: &'static str,
+    telemetry_id_override: Option<SharedString>,
```
benbrandt: This allows us to override in a few cases.
Potentially we could just let the agents tell us... I think the only one that doesn't is gemini

crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -519,6 +522,10 @@ impl AcpThreadView {
                 }
             };
 
+            if track_load_event {
+                telemetry::event!("Agent Thread Started", agent = connection.telemetry_id());
```
benbrandt, line 528: I pushed this down. We have to thread the bool through a few more places, but I think it works

benbrandt (commented):
There is one question of should we even override any of these other than gemini.
Since that is the only hard-coded one that doesnt' provide us a name and the historical data is kind of messed up anyway

crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -5377,46 +5385,42 @@ impl AcpThreadView {
     }
 
     fn render_codex_windows_warning(&self, cx: &mut Context<Self>) -> Option<Callout> {
```
bennetbo: Nit: We can drop the option and just return `Callout`

bennetbo (approved):

crates/agent_ui/src/acp/thread_view.rs
benbrandt: good catch!

PR #44546, opened 2025-12-10
language_models: Resolve provider ID conflicts for openai_compatible

aeroxy:
**Problem:**
When configuring `openai_compatible` providers in the Zed project, there was a risk of provider ID conflicts with existing providers (e.g., Mistral). This could lead to unexpected behavior, such as the `openai_compatible` provider overwriting or conflicting with the official Mistral provider.

**Solution:**
Implemented logic in the `register_openai_compatible_providers` function to automatically resolve provider ID conflicts. If a provider ID already exists in the registry, the function appends a number to the ID (e.g., "Mistral 2") to ensure uniqueness. This ensures that `openai_compatible` providers do not conflict with existing providers.

Release Notes:

- Added automatic resolution of provider ID conflicts for `openai_compatible` providers. If a provider ID already exists, the system will append a number to the ID (e.g., "Mistral 2") to ensure uniqueness. This prevents conflicts with existing providers like Mistral and ensures smooth integration of custom `openai_compatible` configurations.
- Users can now safely configure `openai_compatible` providers without worrying about conflicts with existing providers.
- The system will automatically handle naming conflicts, making it easier to integrate custom language model providers.

**Example:**
If you configure an `openai_compatible` provider with the ID "Mistral" and a Mistral provider already exists, the system will automatically rename the `openai_compatible` provider to "Mistral 2" to avoid conflicts.

### DIFF
--- a/crates/edit_prediction/src/edit_prediction_tests.rs
+++ b/crates/edit_prediction/src/edit_prediction_tests.rs
@@ -1075,9 +1075,9 @@ fn model_response(request: open_ai::Request, diff_to_apply: &str) -> open_ai::Re
             finish_reason: None,
         }],
         usage: Usage {
-            prompt_tokens: 0,
-            completion_tokens: 0,
-            total_tokens: 0,
+            prompt_tokens: Some(0),
+            completion_tokens: Some(0),
+            total_tokens: Some(0),
         },
     }
 }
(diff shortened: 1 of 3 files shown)


bennetbo:
I think adding `+1` in this case is not great UX. To me it makes more sense to disable the button in the modal and show a warning that the name is already taken.

aeroxy:
> I think adding `+1` in this case is not great UX. To me it makes more sense to disable the button in the modal and show a warning that the name is already taken.

this could also work

PR #44598, opened 2025-12-11
Update file vs folder open keymaps on macOS/linux to match Windows

### LINKED ISSUE #44597: [linux] Ctrl+O opens directory picker instead of the file one
### Reproduction steps

Press Ctrl+O

Current: opens the Open Folder dialog. At least with xdg-desktop-portal-gnome, this results in an "Open Folder" dialog where you cannot open an individual file.

I'm guessing under the hood it calls https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html#org-freedesktop-portal-filechooser-openfile with `directory=true`.

Expected: Ctrl+O should open the file dialog (`workspace::OpenFiles`).

Basically same as https://github.com/zed-industries/zed/issues/44040 but on Linux.

Zed: v0.216.0+stable.94.f381f06bcd9fd7fae8c62c2308887edcc00f2d3d (Zed) 
OS: Linux Wayland fedora 43
Memory: 31.2 GiB
Architecture: x86_64
GPU: AMD Radeon RX 7900 GRE (RADV NAVI31) || radv || Mesa 25.2.7

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

_No response_

None

Zachiah:
Closes #44597

Matches what was done here:
https://github.com/zed-industries/zed/pull/44409/commits/55dfbaca685fc466efedace55527b9d12ab67950#diff-cc832e840d61526768bb4acec7645a71e8b160a65a30e7ce9e9c51762b58199a

Release Notes:

- Standardize Cmd-O = open file, Cmd-K Cmd-O = open folder across operating systems.


assets/keymaps/default-macos.json
```diff
@@ -34,7 +34,8 @@
       "alt-shift-enter": "menu::Restart",
       "cmd-shift-w": "workspace::CloseWindow",
       "shift-escape": "workspace::ToggleZoom",
-      "cmd-o": "workspace::Open",
+      "cmd-o": "workspace::OpenFiles",
+      "cmd-k cmd-o": "workspace::Open",
```
Veykril: It seems vscode only has `cmd-o` for open, so this change should not be made here given the default keymap is vscode specific

Zachiah: Should I remove the `cmd-k cmd-o` from Windows as well or just the 2 I added?

Veykril: No windows is correct. Its just on mac where these differ weirdly enough (and I presume linux uses the same as windows here, so that change should check out)

Zachiah: OK done. For future would it have been preferable to amend it? I definitely would have on a lot of projects but I like to ask because I don't want to make people review changes twice if they don't want to no matter how small. Like, as a zed culture thing if I make a tiny PR and have to make a change should I amend if it makes sense, or always make a new commit?

Veykril: Committing is probably preferred as we do squash merges, so it will ultimately end up as a single commit anyways

assets/keymaps/default-macos.json
```diff
@@ -34,7 +34,7 @@
       "alt-shift-enter": "menu::Restart",
       "cmd-shift-w": "workspace::CloseWindow",
       "shift-escape": "workspace::ToggleZoom",
-      "cmd-o": "workspace::Open",
+      "cmd-o": "workspace::OpenFiles",
```
Veykril:
```suggestion
      "cmd-o": "workspace::Open
```
This should remain as just open I believe

assets/keymaps/default-macos.json
```diff
@@ -34,7 +34,7 @@
       "alt-shift-enter": "menu::Restart",
       "cmd-shift-w": "workspace::CloseWindow",
       "shift-escape": "workspace::ToggleZoom",
-      "cmd-o": "workspace::Open",
+      "cmd-o": "workspace::Open
```
Veykril:
```suggestion
      "cmd-o": "workspace::Open",
```

Veykril:
Thanks!

Zachiah:
Thank you! By the way, what would it take to get traige? I'd really enjoy tagging issues.

PR #44599, opened 2025-12-11
editor: Fix tag auto close for Vue namespaced component

smitbarmase:
This PR changes how we compute tag name for the JSX-like components for auto close. Before we had hardcoded logic for grabbing the first child of the opening node, but now we use all eligible children and use them to compute the tag name. 

Before in Vue, autocompleting the code below:

```vue
<Hello.World^
```
would result in:

```vue
<Hello.World></Hello>
```
instead of:
```vue
<Hello.World></Hello.World>
```

so, in this case, the tag name is computed as "Hello.World" instead of "Hello".

Note: For Vue, this also needs an extension update to include `directive_attribute`/`attribute` as `tag_name_node_name_alternates`.

Release Notes:

- N/A

Co-authored-by: Kunall Banerjee <hey@kimchiii.space>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4386,7 +4386,7 @@ dependencies = [
  "smol",
  "task",
  "telemetry",
- "tree-sitter",
+ "tree-sitter 0.25.10",
  "tree-sitter-go",
  "util",
  "zlog",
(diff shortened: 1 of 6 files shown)


smitbarmase:
We figured this can be fixed in Vue Tree Sitter. cc @yeskunall

PR #44608, opened 2025-12-11
windows: Fix incorrect cursor insertion keybinds

Veykril:
Release Notes:


bugadani:
Can we have key binding changes in the changelog, please? If an update is messing with my muscle memory, I would like to see what I need to undo.

Veykril:
Yea sorry, you are right. I should've put that as a bug fix entry

Veykril:
Updated the release notes, sorry again, and thanks for raising this

PR #44627, opened 2025-12-11
docs: Improve HTML LSP config examples

gursuj:
If you set `vscode-html-language-server` instead of `Prettier` for formatting HTML, code with style attributes or style tags might throw this error in the LSP logs: `Cannot read properties of null (reading 'validProperties')`. 

**Steps to recreate issue:**

1. In `settings.json`, set:
```json
  "languages": {
    "HTML": {
      "formatter": "language_server",
    }
  }
```
2. In an html file, set the attribute `style=""` for any element.
3. Go to "Language Servers" (Ctrl+Alt+L), view logs for `vscode-html-language-server` and search for `Error`

**Solution**
Use this config in `settings.json`:
```json
  "languages": {
    "HTML": {
      "formatter": "language_server"
    }
  },
  "lsp": {
    "vscode-html-language-server": {
      "settings": {
        // workaround to prevent html lsp error about validproperties
        "css": {
          "lint": {
            "validProperties": {}
          }
        }
      }
    }
  }
```

This is the most common workaround I found for this issue. Most of the solutions were for neovim using lua so this might be helpful for zed users. Based on this [nvim example](https://neovim.discourse.group/t/help-with-diagnostics-for-html-files/2682/2). I don't fully understand it so not sure if this belongs as the default config example in the docs or should be reported as a bug in the vscode lsp repo.

Not sure if I need to put any release notes for docs changes.

Release Notes:

- N/A

### DIFF
--- a/docs/src/languages/html.md
+++ b/docs/src/languages/html.md
@@ -59,12 +59,6 @@ You can customize various [formatting options](https://code.visualstudio.com/doc
             // Add an extra newline before <div> and <p>
             "extraLiners": "div,p"
           }
-        },
-        // workaround to prevent html lsp error about validproperties
-        "css": {
-          "lint": {
-            "validProperties": {}
-          }
         }
       }
     }
--- a/docs/src/languages/html.md
+++ b/docs/src/languages/html.md
@@ -25,6 +25,7 @@ You can disable `format_on_save` by adding the following to your Zed `settings.j
   "languages": {
     "HTML": {
       "format_on_save": "off",
+      "language_servers": ["vscode-html-language-server"],
     }
   }
 ```
--- a/docs/src/languages/html.md
+++ b/docs/src/languages/html.md
@@ -58,6 +58,12 @@ You can customize various [formatting options](https://code.visualstudio.com/doc
             // Add an extra newline before <div> and <p>
             "extraLiners": "div,p"
           }
+        },
+        // workaround to prevent html lsp error about validproperties
+        "css": {
+          "lint": {
+            "validProperties": {}
+          }
         }
       }
     }
--- a/docs/src/languages/html.md
+++ b/docs/src/languages/html.md
@@ -25,7 +25,6 @@ You can disable `format_on_save` by adding the following to your Zed `settings.j
   "languages": {
     "HTML": {
       "format_on_save": "off",
-      "language_servers": ["vscode-html-language-server"],
     }
   }
 ```
@@ -40,6 +39,7 @@ To use the `vscode-html-language-server` language server auto-formatting instead
   "languages": {
     "HTML": {
       "formatter": "language_server",
+      "language_servers": ["vscode-html-language-server"],
     }
   }
 ```
--- a/docs/src/languages/html.md
+++ b/docs/src/languages/html.md
@@ -39,7 +39,7 @@ To use the `vscode-html-language-server` language server auto-formatting instead
   "languages": {
     "HTML": {
       "formatter": "language_server",
-      "language_servers": ["vscode-html-language-server"],
+      "language_servers": ["vscode-html-language-server", "!prettier", "..."],
     }
   }
 ```


gursuj:
@cla-bot check

gursuj:
Also, I was having issues with HTML formatting not working when using this config mentioned in the docs:
```json
  "languages": {
    "HTML": {
      "formatter": "language_server",
    }
  }
```
Setting `"language_servers": ["vscode-html-language-server"]` fixed that but now I don't get LSP features for emmet and tailwind so I guess that last commit isn't safe to use.

yeskunall:
Hey @gursuj thanks for the PR! FYI: this is on my radar, but I want to first see if I can fix your issue over on Discord before reviewing this PR.

gursuj:
Oh thanks. I opened a support thread on discord [here](https://discord.com/channels/869392257814519848/1448666301751890021). I've updated it with a potential solution but I still have some issues.

yeskunall:
> Oh thanks. I opened a support thread on discord [here](https://discord.com/channels/869392257814519848/1448666301751890021). I've updated it with a potential solution but I still have some issues.

Yup, I replied to you there. I’m still a little busy, but I have a few docs PRs to review which I think I’ll get to before the end of this week.

docs/src/languages/html.md
```diff
@@ -39,6 +39,7 @@ To use the `vscode-html-language-server` language server auto-formatting instead
   "languages": {
     "HTML": {
       "formatter": "language_server",
+      "language_servers": ["vscode-html-language-server", "!prettier", "..."],
```
yeskunall, line 42: Hmm, I wonder if we should enable `vscode-html-language-server` by default since the extension is (the only one) automatically installed.

Right now, we don’t: https://github.com/zed-industries/zed/blob/main/assets/settings/default.json#L1863-L1867

docs/src/languages/html.md
```diff
@@ -58,6 +59,12 @@ You can customize various [formatting options](https://code.visualstudio.com/doc
             // Add an extra newline before <div> and <p>
             "extraLiners": "div,p"
           }
+        },
+        // workaround to prevent html lsp error about validproperties
+        "css": {
```
yeskunall: I keep running into this error right now:

```
Error while validating file:///Users/kimchi/dev/local/bun-blank/index.html: Cannot read properties of null (reading 'validProperties')
```

This seems like a bug with the LSP, and as such I don’t think your suggestion should be listed as a workaround. I didn’t have to do this to get it working in VS Code.

gursuj: Yeah other nvim users were getting this error too. I think this only shows up in the LSP logs rather than anywhere in the GUI, and doesn't really cause any issues right? I'm fine with not having this default too. This PR should show up in google search anyways if anyone wants to get rid of the error.

gursuj: I was having issues with my wifi range. This seems to have gotten posted multiple times. Sorry about that.

yeskunall: But without the workaround, I don’t get warned about an empty ruleset error (`css emptyRules`), which I think you’d want if you were using the LSP.

yeskunall: Can you file a separate issue for this particular bug? I’m going to update the PR as well.

docs/src/languages/html.md
gursuj: Hmm I think default prettier is fine too. Might have to ask others. When formatting, Zed was placing each HTML attribute on separate lines despite the setting being [off by default](https://prettier.io/docs/options#single-attribute-per-line), which is why I started looking into the other language server.

docs/src/languages/html.md
gursuj: #44874

docs/src/languages/html.md
gursuj: Hmm I went through this again. Previously, without this line, HTML formatting wouldn't work at all (when using Zed docs' recommended config). I had created a [minimal reproducible example](https://github.com/gursuj/zed-html-buggy-config) that also describes this a bit more. Seems in the latest version of Zed, formatting HTML works without this line so not necessary to merge.

PR #44639, opened 2025-12-11
gpui: Fix some memory leaks on macOS platform

marcocondrache:
While profiling with instruments, I discovered that some of the strings allocated on the mac platform are never released, and the profiler marks them as leaks

image

Release Notes:

- N/A


Anthony-Eid:
Thank you so much for catching this! I added a commit that sets `NSString::alloc` as a clippy disallowed method to make sure this leak doesn't happen again. I also added auto release to a test as well.

PR #35960, opened 2025-08-10
Ignore whitespace in git blame invocation

jytug:
This works around a bug wherein inline git blame is unavailable for files with CRLF line endings. At the same time, this prevents users from seeing whitespace-only changes in the editor's git blame

Closes #35836

Release Notes:

- N/A *or* Added/Fixed/Improved ...

PR #44648, opened 2025-12-11
git: Revert "Ignore whitespace in git blame invocation"

osiewicz:
Reverts zed-industries/zed#35960
cc @cole-miller


osiewicz:
We could maybe use `--ignore-space-at-eol` instead.

PR #44655, opened 2025-12-11
gpui: Improve path rendering and bounds performance

marcocondrache:
On my macOS machine, the paths_bench example currently runs at about 40 fps in release mode. Profiling shows that at a 24 ms frame interval, roughly 6–9 ms are spent on the GPU (fragment work, which still needs optimization), while the remainder is CPU time. After more accurate profiling, the frame graph revealed that most of the CPU cost came from bounds tree insertion 

image

The issue is that the existing tree structure performs poorly when many bounds are identical, which is exactly what happens in paths_bench.

The new implementation is a variant of an R-tree. I may experiment with an R*-tree variant later, but the current approach already addresses three major problems:

1. Fast path for max leaf (now O(1) for overlapping bounds)
Before: Every insertion walked the entire tree to find the maximum intersecting ordering.
After: The tree maintains a pointer to the leaf with the global maximum ordering, letting us test it immediately.

2. Higher branching factor
The tree now uses four children per node instead of two, which reduces height and improves traversal efficiency.

3. Better overall balancing
Node selection and splitting strategies produce a more stable and evenly distributed tree structure.

These changes reduce insertion complexity:

- Identical or fully overlapping bounds drop from O(n^2) to O(n)
- Mostly overlapping bounds drop from O(n^2) to O(n log n)
- Spatially distributed bounds keep similar asymptotic cost but benefit from reduced height
- Non-overlapping bounds also benefit from reduced height

And specific operations improve significantly:

- find_max_ordering: from O(n) to O(1) when the max leaf intersects, otherwise O(log n)
- insert_leaf: from O(n) for identical bounds to O(log4 n)
- Total cost for inserting n items: from O(n^2) worst-case to O(n) best-case and O(n log n) worst-case

In the paths_bench example, these changes produce a **100% performance improvement**, increasing the frame rate from 40 fps to 80 fps. With a few more optimizations, I believe we can push the example to around 120 fps.

Before:

image

After:

image

Release Notes:

- Improved path and rendering performance

### DIFF
--- a/crates/gpui/src/bounds_tree.rs
+++ b/crates/gpui/src/bounds_tree.rs
@@ -7,7 +7,7 @@ use std::{
 
 /// Maximum children per internal node (R-tree style branching factor).
 /// Higher values = shorter tree = fewer cache misses, but more work per node.
-const MAX_CHILDREN: usize = 4;
+const MAX_CHILDREN: usize = 12;
 
 /// A spatial tree optimized for finding maximum ordering among intersecting bounds.
 ///
--- a/crates/gpui/src/bounds_tree.rs
+++ b/crates/gpui/src/bounds_tree.rs
@@ -64,6 +64,7 @@ enum NodeKind {
 /// Fixed-size array for child indices, avoiding heap allocation.
 #[derive(Debug, Clone)]
 struct NodeChildren {
+    // Keeps an invariant where the max order child is always at the end
     indices: [usize; MAX_CHILDREN],
     len: u8,
 }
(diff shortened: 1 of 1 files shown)


marcocondrache:
I tested it with a higher branching factor (12) and it runs at 140fps

image

crates/gpui/src/bounds_tree.rs
```diff
@@ -26,158 +102,220 @@ where
...
+                    }
+
+                    // Sort by max_order (ascending, so highest is pushed last = popped first)
+                    child_orders[..num_children].sort_unstable_by_key(|(max_order, _)| *max_order);
+
+                    for (child_max, child_idx) in &child_orders[..num_children] {
+                        if *child_max > max_found {
+                            self.search_stack.push(*child_idx);
+                        }
+                    }
```
marcocondrache: We can shorten this hot path even more by removing the sorting. I think we can store the max_order child at the end of the children array and reverse the iteration. So it hits first when searching, it requires careful swapping during insertion tho

marcocondrache: The max-end variant cuts down iteration during search quite a bit, at the cost of a potentially less optimal traversal order. Overall it’s a worthwhile trade-off because:

- The max child typically yields the largest max_found, which gives the strongest pruning effect
- Non-max siblings only matter when they both intersect the query and have meaningfully lower max_order values, which is relatively rare
- Sorting children would add roughly 40 extra iterations per node, compounding across the tree
- UI bounds tend to cluster spatially, which reduces the practical impact of any suboptimal pruning order

mikayla-maki (approved):
Wow, this code is so clean! And thank you for the detailed explanation and data backing it up. Nothing looks amiss in Zed rendering + it passes the tests, so I think we're good to go :D

Issue #41580, opened 2025-10-30
In helix mode, on zed, Ctrl-s saves the file instead of saving selection to a jumplist as it should


luccahuguet:
### Summary

In helix mode, on zed, Ctrl-s saves the file instead of saving selection to a jumplist

### Description
In regular helix, ctrl-s does not save the file. For that we have :write, which is already present and working well in Zed. Instead, in regular helix, ctrl-s serves the purpose of saving that selection to the jump-list, that we can walk through forwards or backwards using ctrl-i and ctrl-o.

Zed already supports jumplists, so all that is left is to re-bind ctrl-s to to that instead. 

Zed's jumplist system does not seem to work with selections as it is, but that is ok. We should focus on getting a basic version of this working, and cater to the selection aspect of the helix's jumplist in a separate issue following this one.

### Steps to reproduce:
press ctrl-s on zed, on helix mode, with a file open

### Expected Behavior:
Pressing Ctrl+S saves the current cursor position to the jumplist, allowing the user to navigate back to this position using Ctrl+O (backward) and Ctrl+I (forward), consistent with regular Helix's behavior.

### Actual Behavior:
Pressing Ctrl+S saves the file instead of saving the position to the jumplist, which is the default Zed behavior rather than the expected Helix mode behavior.

### Future Expected Behavior:
After this issue is completed, we can focus on not only saving the cursor position, to the jumplist, but also the selection. As in true helix.

### Zed Version and System Specs

Zed: v0.210.4 (Zed) 
OS: Linux Wayland pop 24.04
Memory: 31.1 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3050 6GB Laptop GPU || NVIDIA || 580.82.09

### Relevant PRs
- https://github.com/zed-industries/zed/pull/33495

kubkon:
Thanks for filing an issue and providing a repro! I'll take a look!

luccahuguet:
the pleasure is mine @kubkon

11happy:
I am interested to work on this !
@kubkon could you please assign me this issue. 
Thank you : )

ShaiKKO:
@kubkon PR for full jump-list support (https://github.com/zed-industries/zed/pull/44661)

- adds proper `ctrl-s`, `ctrl-i`, and `ctrl-o` functionality matching Helixes implementation.

luccahuguet:
@11happy it's easier to just implement it and open a PR (next time)

[...]

PR #44661, opened 2025-12-11
Add Helix jump list navigation

ShaiKKO:
This PR implements Helix's jump list feature, addressing issue #41580 and continuing the work discussed in #33580.

The jump list is one of Helix's core navigation features that lets users save cursor positions and quickly jump between them. I've implemented the three commands that make up this feature: `Ctrl-s` to save the current selection, `Ctrl-o` to jump backward through saved positions, and `Ctrl-i` to jump forward.

I followed the existing patterns in the vim crate pretty closely here. The `JumpList` struct lives in a new `jump_list.rs` module under `helix/`, and the global state is stored in `VimGlobals` alongside marks and registers. This felt like the natural home for it since the jump list persists across buffer switches, similar to how marks work.

The data structure mirrors Helix's implementation - a `VecDeque` with a capacity of 30 entries and a cursor tracking where we are in the history. Each entry stores a buffer ID and the selection anchors, which means positions stay valid even as the document is edited.

When you save a position that's identical to the last one, we skip it to avoid cluttering the list with duplicates. If you're in the middle of the history and save a new position, everything "ahead" of you gets truncated - this matches how most undo systems work and feels intuitive in practice.

The trickiest bit was handling the "jump from present" case. When you're at the end of the list (haven't jumped yet) and hit `Ctrl-o`, Helix automatically saves your current position before jumping. This way you can always get back to where you started. I've replicated this behavior.

Cross-buffer jumps work by looking up the target buffer in the workspace and activating it before restoring the selection. The anchors handle position validity automatically since they're tied to the buffer's edit history.

I've added unit tests for the `JumpList` data structure covering the basic operations, capacity limits, duplicate prevention, and the forward-history truncation behavior. There are also integration tests that exercise the actual keybindings in a `VimTestContext` with Helix mode enabled.

## Release Notes

- Added Helix jump list navigation: `Ctrl-s` to save position, `Ctrl-o` to jump backward, `Ctrl-i` to jump forward (vim mode with Helix keybindings enabled)

---

Related: #41580, #33580

### DIFF
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -518,6 +518,11 @@
       "space p": "editor::Paste",
       "space y": "editor::Copy",
 
+      // Jump list
+      "ctrl-s": "vim::HelixSaveSelection",
+      "ctrl-o": "vim::HelixJumpBackward",
+      "ctrl-i": "vim::HelixJumpForward",
+
       // Other
       ":": "command_palette::Toggle",
       "m": "vim::PushHelixMatch",
--- a/crates/vim/src/helix.rs
+++ b/crates/vim/src/helix.rs
@@ -7,21 +7,26 @@ mod object;
 mod paste;
 mod select;
 
-pub use jump_list::{JumpEntry, JumpList};
+pub use jump_list::{
+    JumpEntry, JumpList, JumpLocation, JumpSelections, SelectionAnchors, SelectionPoints,
+};
 
 use editor::display_map::DisplaySnapshot;
 use editor::{
-    DisplayPoint, Editor, EditorSettings, HideMouseCursorOrigin, MultiBufferOffset,
+    DisplayPoint, Editor, EditorSettings, HideMouseCursorOrigin, MultiBuffer, MultiBufferOffset,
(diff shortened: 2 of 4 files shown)


ShaiKKO:
@cla-bot check

ShaiKKO:
After further review I identified 3 areas that I address in the PR below (should have caught these before submission, so apologies for that, hope this saves you some time on the backend).

- Store full selection ranges (`start+end` anchors) instead of just cursor heads

- Add JumpLocation enum: `Buffer(EntityId)` for open files, `Path(Arc<Path>)` for closed

- Add JumpSelections enum: Anchors for open buffers, Points for closed files

- Hook buffer release via `cx.observe_release()` to convert entries when files close

- Add `helix_jump_to_path()` to reopen closed files and restore selections

( This enables jumping back to positions in files that have been closed and  is **NOT** a default helix feature, and is noted as an issue in helix itself, I opted to implement it here and see what people thought on the matter) - the jump list preserves the file path and selection points and reopens the file. )

- Track watched buffers in VimGlobals to manage lifecycle subscriptions

PR #44665, opened 2025-12-11
vim: Fix cursor position when moving up/down at buffer boundaries with goal column

noib3:
Fixes incorrect cursor positioning in Vim mode when attempting to move up on the first row or down on the last row of the buffer while a `SelectionGoal` is set.

Previously, when a vertical movement had set a goal column, attempting to move beyond the buffer boundaries would incorrectly reposition the cursor instead of staying in place.

Before:

https://github.com/user-attachments/assets/ceeb1104-860e-441c-a51f-e5a86666b4d3

After:

https://github.com/user-attachments/assets/6a5555ba-9e40-488b-a2bf-cacc70a61e42

Release Notes:

- Fixed vim mode incorrectly repositioning the cursor when pressing `k` on the first line or `j` on the last line after establishing a goal column

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -1555,6 +1555,12 @@ impl Sub<u32> for DisplayRow {
     }
 }
 
+impl PartialEq<u32> for DisplayRow {
+    fn eq(&self, &other: &u32) -> bool {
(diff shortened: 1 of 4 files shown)


noib3:
@cla-bot check

PR #44669, opened 2025-12-11
editor: Implement inline references (peek references)

benjaminkvm:
## Description

Implement inline references functionality referred to in other editors as `Peek References`. This implementation is visually modeled after the inline AI functionality and with some future generalization could be used for broader inline use-cases (expanded peeking support, etc).

Release Notes:

- Added inline references (peek all references) that can be triggered via keyboard shortcut or context menu.

## Testing

- [x] All references are displayed inline with very similar UI to the existing `Find All References` view.
- [x] The inline references view closes upon navigation to the underlying buffer (remains closed upon navigation back).
- [x] The inline references view avoids clipping my ensuring it is a threshold above the bottom of the buffer where it was triggered. Future logic could vertically center the inline references view within the editor viewport if desired.
- [x] Changes made within inline reference editors are successfully saved.
- [x] Verified `editor: fold all` / `editor: unfold all` correctly folds inline references.
- [x] The inline references view is closed upon pressing `Esc` from either the parent editor or an inline reference editor.
- [x] Integrates as expected with Vim mode emulation enabled; `Esc` from `NORMAL` mode from either within the parent editor or an inline reference editor closes the inline references view. Existing `Esc` functionality continues to work as expected.
- [x] All existing test cases continue to pass.
- [x] Implemented additional test cases, including coverage for interaction with Vim mode emulation.

## Screenshots

Screenshot 2025-12-11 at 2 35 52 PM

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -511,6 +511,7 @@
       "shift-f12": "editor::GoToImplementation",
       "alt-ctrl-f12": "editor::GoToTypeDefinitionSplit",
       "alt-shift-f12": "editor::FindAllReferences",
+      "ctrl-shift-f12": ["editor::FindAllReferences", { "inline": true }],
       "ctrl-m": "editor::MoveToEnclosingBracket", // from jetbrains
       "ctrl-|": "editor::MoveToEnclosingBracket",
       "ctrl-{": "editor::Fold",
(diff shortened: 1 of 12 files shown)


benjaminkvm:
@cla-bot check

ConradIrwin:
Thank you for this, and sorry for the slow reply here.

The editor is already quite complex, and showing nested multiline editors inside the editor as blocks has various problems (I know we still do it in some places, but we're really trying not to).

All that to say, I don't think this is the right way to tackle this. In some ways a hover popover is better because it requires no dismissal (but then it struggles with keyboard navigation). We couldmaybe add a "go to all references in the adjacent pane", but then we'd want a "dismiss pane to the right action"...

I think the best option is what we have now, if you want to find all references you can; and when you're done looking at them you can go back (`ctrl--` in non-vim mode or `ctrl-o` in vim mode).

PR #44678, opened 2025-12-12
Fix stale pending keybinding indicators on focus change

xipeng-jin:
Closes #ISSUE

Problem:

- The status bar’s pending keystroke indicator (shown next to --NORMAL-- in Vim mode) didn’t clear when focus moved to another context, e.g. hitting g in the editor then clicking the Git panel. The keymap state correctly canceled the prefix, but observers that render the indicator never received a “pending input changed” notification, so the UI kept showing stale prefixes until a new keystroke occurred.

Fix:

- The change introduces a `pending_input_changed_queued` flag and a new helper `notify_pending_input_if_needed` which will flushes the queued notification as soon as we have an App context. The `pending_input_changed` now resets the flag after notifying subscribers.

Before:

https://github.com/user-attachments/assets/7bec4c34-acbf-42bd-b0d1-88df5ff099aa

After:

https://github.com/user-attachments/assets/2264dc93-3405-4d63-ad8f-50ada6733ae7

Release Notes:

-  Fixed: pending keybinding prefixes on the status bar now clear immediately when focus moves to another panel or UI context.


xipeng-jin:
Hi @ConradIrwin , we surface this issue since we introduced the timeout logic in this PR https://github.com/zed-industries/zed/pull/42659. The issue was existing before, but we will have 1 second timeout to flush it anyway. We should address this before adding the which-key menu.

xipeng-jin:
Please let me know if we need to add any additional integration or UI tests for this one. Thanks

ConradIrwin:
Makes sense. 

Can we do this without modifying `draw`? Does `cx.defer` run at the right time?

xipeng-jin:
> Can we do this without modifying `draw`? Does `cx.defer` run at the right time?

I explored using `cx.defer` when clearing pending keystrokes, but the problem is that both `Window::focus` and `App::clear_pending_keystrokes` can run outside of an effect cycle in which there may be no `Context<App>` available. In `Window::focus` we only have `&mut Window.`, so I choose to buffer the “pending input changed” notification and flush it at the next time we do have an App context, which is guaranteed when `draw` runs.

ConradIrwin:
Makes sense, it just feels like it's the wrong layer of abstraction. Could the component listen to the window's focus event and check if the pending keybindings have changed?

xipeng-jin:
> Could the component listen to the window's focus event and check if the pending keybindings have changed?

Good question. But I am afraid of there are not better choice to handle this, listening to the focus event is not enough. Apparently, how we should handle the platform hooks (lets say you open a OS menu on the top bar), the focus path doesn't change for this case

xipeng-jin:
Another thing is I think even for window focus changes, we’d still need glue code somewhere to detect “pending keystrokes were cleared because a focus change happened” and notify every listener manually, which effectively duplicates what the existing pending-input observer already does. What do you think? I am not 100% sure

ConradIrwin:
Maybe we should pass a cx into `window.focus(handle, cx)`? Not sure if that would fix it, but I'd like to avoid making this code path so special

xipeng-jin:
Hi @ConradIrwin , thank you for following up. I do have some thoughts over the last weekend and I totally agree we should approach this differently. It is not good to have that change in the draw() function which is in the rendering time and also it is not aligned with overall GPUI architecture.

> Maybe we should pass a cx into window.focus(handle, cx)?

That is one alternative, and it is an ideal and concrete approach. However, it will have some risks for breaking changes on API interface. Currently there are lots of references for this window.focus() function. Personally, I am more a bit hesitate to touch that part yet only for fixing this particular issue.

There are some other alternatives I am still exploring. I will post those comments later today. Thank you.

ConradIrwin:
Makes sense, hopefully `window.focus(cx)` is mostly an automatable change; but open to other solutions too.

xipeng-jin:
Hi @ConradIrwin , it seems like now it has some merge conflicts. Would you be able to address and merge it? Or I can look into it though. Thank you

ConradIrwin:
blergh, I can't do it fast enough. Let me try one more time

PR #44688, opened 2025-12-12
Allow opening a new window for secondary worktrees

nathansobo:
Closes #

Previously, running `zed foo` when `foo` was already open as a worktree in an existing window would reuse that window, even if `foo` was not the primary worktree. This made it impossible to open a dedicated window for a directory that was already added as a secondary worktree elsewhere.

Now, we only reuse an existing window if the path matches its primary worktree (the first visible worktree, typically the directory originally opened to create that window). This allows users to run `zed foo` and get a new window where `foo` is the primary worktree, even if `foo` is already open as a secondary worktree in another window.

Release Notes:

- Changed window reuse behavior: `zed <path>` now opens a new window if the path is only present as a secondary worktree in existing windows, rather than reusing that window.

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -7574,7 +7574,6 @@ pub fn open_paths(
 > {
     let abs_paths = abs_paths.to_vec();
     let mut existing = None;
-    let mut best_match = None;
     let mut open_visible = OpenVisible::All;
     #[cfg(target_os = "windows")]
     let wsl_path = abs_paths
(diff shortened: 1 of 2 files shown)


nathansobo:
Unclear on the implications of multiple windows for the same worktree - @benbrandt is worried that it could cause issues with our persistence of workspace state. Going to close this for now to stay focused on other tasks.

SomeoneToIgnore:
To note, the CLI has a bunch of ways to control this manually:

```
~ ❯ zed --help
The Zed CLI binary.
This CLI is a separate binary that invokes Zed.

Examples:
    `zed`
          Simply opens Zed
    `zed --foreground`
          Runs in foreground (shows all logs)
    `zed path-to-your-project`
          Open your project in Zed
    `zed -n path-to-file `
          Open file/folder in a new window


Usage: zed [OPTIONS] [PATHS_WITH_POSITION]...

Arguments:
  [PATHS_WITH_POSITION]...
          The paths to open in Zed (space-separated).

          Use `path:line:column` syntax to open a file at the given line and column.

Options:
  -w, --wait
          Wait for all of the given paths to be opened/closed before exiting

  -a, --add
          Add files to the currently open workspace

  -n, --new
          Create a new workspace

  -r, --reuse
          Reuse an existing window, replacing its workspace
```

Issue #44510, opened 2025-12-10
Zed beachballs when certain strings appear in the terminal pane


notahat:
### Reproduction steps

1. Start Zed
2. Open the terminal pane
3. Paste in a the string below
4. Zed will beachball, and require force-quitting

The string:
```
...................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................E.
```

Yep, that's a lot of dots with one E in it. The log file (attached below) is revealing.

I haven't narrowed down exactly what combination of length and characters triggers the problem, but the above string definitely does.

Why, you might ask, am I pasting this into the terminal pane? The real-world version is that I have a command (packwerk) that outputs this string, and scrolling through the terminal pane with that output in it is enough to trigger the beachball. Pasting the string directly into the terminal pane is just a quick way to reproduce the problem.

### Zed version and system specs

Zed: v0.215.3-90+fda162bf3d9a11d18e07cbf2b488829450135d34 (Zed) 
OS: macOS 26.1
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

The following lines repeat many times just before the beachball:

```log
[...]

PR #44714, opened 2025-12-12
terminal: Improve scroll performance

davewa:
Related to: 
- #44510 
- #44407 

Previously we were searching for hyperlinks on every scroll, even if Cmd was not held. With this PR, 
- We only search for hyperlinks on scroll if Cmd is held
- We now clear `last_hovered_word` in all cases where Cmd is not held
- Renamed `word_from_position` -> `schedule_find_hyperlink`
- Simplified logic in `schedule_find_hyperlink`

Performance measurements

The test scrolls up and down 20,000x in a loop. However, since this PR is just removing a code path that was very dependent on the length of the line in terminal, it's not super meaningful as a comparison. The test uses a line length of "long line ".repeat(1000), and in main the performance is directly proportional to the line length, so for benchmarking it in main it only scrolls up and down 20x. I think all that is really useful to say is that currently scrolling is slow, and proportional to the line length, and with this PR it is buttery-smooth and unaffected by line length. I've included a few data points below anyway. At least the test can help catch future regressions.
 
| Branch | Command | Scrolls | Iter/sec | Mean [ms] | SD [ms] | Iterations | Importance (weight) |
|:---|:---|---:|---:|---:|---:|---:|---:|
| main | tests::perf::scroll_long_line_benchmark | 40 | 16.85 | 712.00 | 2.80 | 12 | average (50) |
| this PR | tests::perf::scroll_long_line_benchmark | 40 | 116.22 | 413.60 | 0.50 | 48 | average (50) |
| this PR | tests::perf::scroll_long_line_benchmark | 40,000 | 9.19 | 1306.40 | 7.00 | 12 | average (50) |
| only overhead | tests::perf::scroll_long_line_benchmark | 0 | 114.29 | 420.90 | 2.00 | 48 | average (50) |

Release Notes:

- terminal: Improved scroll performance

### DIFF
--- a/crates/terminal/src/terminal.rs
+++ b/crates/terminal/src/terminal.rs
@@ -1711,38 +1711,39 @@ impl Terminal {
             {
                 self.write_to_pty(bytes);
             }
-        } else if e.modifiers.secondary() {
-            self.word_from_position(e.position);
+        } else {
+            self.schedule_find_hyperlink(e.modifiers, e.position);
         }
         cx.notify();
     }
 
-    fn word_from_position(&mut self, position: Point<Pixels>) {
-        if self.selection_phase == SelectionPhase::Selecting {
+    fn schedule_find_hyperlink(&mut self, modifiers: Modifiers, position: Point<Pixels>) {
+        if self.selection_phase == SelectionPhase::Selecting
+            || !modifiers.secondary()
+            || !self.last_content.terminal_bounds.bounds.contains(&position)
+        {
             self.last_content.last_hovered_word = None;
-        } else if self.last_content.terminal_bounds.bounds.contains(&position) {
-            // Throttle hyperlink searches to avoid excessive processing
-            let now = Instant::now();
-            let should_search = if let Some(last_pos) = self.last_hyperlink_search_position {
+            return;
+        }
+
+        // Throttle hyperlink searches to avoid excessive processing
+        let now = Instant::now();
+        if self
+            .last_hyperlink_search_position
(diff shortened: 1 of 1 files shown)


davewa:
I thought maybe in this case I could be okay with submitting a performance PR without performance measurements, but I ... just ... can't ... lol. Putting this back into draft to add a performance test.

franciskafyi:
We love tests! @davewa ping me when this is ready and I'll make sure we get to this during Quality Week.

davewa:
@franciskafyi fyi, ready for review now.

crates/terminal/src/terminal.rs
```diff
@@ -2851,4 +2855,102 @@ mod tests {
...
+                term.write_output("long line ".repeat(1000).as_bytes(), cx);
+            });
+
+            (terminal, window)
+        }
+
+        #[perf]
+        #[gpui::test]
+        async fn scroll_long_line_benchmark(cx: &mut TestAppContext) {
+            let (terminal, window) = init_scroll_perf_test(cx).await;
```
davewa, line 2914: I didn't see a way to exclude test initialization from the perf timings?

probably-neb: I don't believe there is. I think it's ok, it's just a constant in the test iterations

probably-neb:
I'm ready to merge this when you are @davewa

davewa:
I’m ready!

probably-neb:
Great work as usual @davewa!

PR #44721, opened 2025-12-12
terminal: Improve regex hyperlink performance for long lines

davewa:
Related to
- #44407

This PR further improves performance for regex hyperlink finding by eliminating unnecessary regex matching. Currently, we repeatedly search for matches from the start of the line until the match contains the hovered point. This is only required to support custom regexes which match strings containing spaces, with multiple matches on a single line. This isn't actually a useful scenario, and is no longer supported. This PR changes to only search twice, the first match starting from the start of the line, and the hovered word (space-delimited). The most dramatic improvement is for long lines with many words.

In addition to the above changes, this PR:
- Adds test for the scenarios from #44407 and #44510 
- Simplifies the logic added in #44407

Performance measurements

For the scenario from #44407, this improves the perf test's iteration time from 1.22ms to 0.47ms.

main:

| Branch | Command | Iter/sec | Mean [ms] | SD [ms] | Iterations | Importance (weight) |
|:---|:---|---:|---:|---:|---:|---:|
| main | terminal_hyperlinks::tests::path::perf::pr_44407_hyperlink_benchmark | 819.64 | 937.60 | 2.20 | 768 | average (50) |
| this PR | terminal_hyperlinks::tests::path::perf::pr_44407_hyperlink_benchmark | 2099.79 | 1463.20 | 7.20 | 3072 | average (50) |

Release Notes:

- terminal: Improve path hyperlink performance for long lines


crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -307,10 +328,27 @@ fn path_match<T>(
     for regex in path_hyperlink_regexes {
```
osiewicz, line 334: FWIW we could also do a two-level matching; with top-level `RegexSet` that figures out which regexes match (and their match range) and then a subsequent run with a single regex that would let us get the captures.

Not sure if that's necessary though, but eh, it's an idea nonetheless.

davewa: Interesting, yeah. Maybe in a later PR.

crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -1002,11 +1054,111 @@ mod tests {
...
+                        "rust-toolchain.toml",
+                        "Hyperlink should have been found"
+                    );
+                });
+            }
+
+            #[perf]
+            // https://github.com/zed-industries/zed/pull/44407
+            pub fn pr_44407_hyperlink_benchmark() {
+                static LINE: &str = "-748, 706, 163, 222, -980, 949, 381, -568, 199, 501, 760, -821, 90, -451, 183, 867, -351, -810, -762, -109, 423, 84, 14, -77, -820, -345, 74, -791, 930, -618, -900, 862, -959, 289, -19, 471, -757, 793, 155, -554, 249, 830, 402, 732, -731, -866, -720, -703, -257, -439, 731, 872, -489, 676, -167, 613, -698, 415, -80, -453, -896, 333, -511, 621, -450, 624, -309, -575, 177, 141, 891, -104, -97, -367, -599, -675, 607, -225, -760, 552, -465, 804, 55, 282, 104, -929, -252,\
```
davewa: Can be `const`

crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -1002,11 +1054,111 @@ mod tests {
...
+                        "392",
+                        "Hyperlink should have been found"
+                    );
+                });
+            }
+
+            #[perf]
+            // https://github.com/zed-industries/zed/issues/44510
+            pub fn issue_44510_hyperlink_benchmark() {
+                static LINE: &str = "..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................\
```
davewa: Can be `const`

crates/terminal/src/terminal_hyperlinks.rs
davewa: Btw, I'd did do something similar before switching to `fancy_regex`. We're not using `fancy_regex` anymore, and once we are sure we won't return to it, we could simplify this code and go back to that. It just used `alacritty_terminal`'s `RegexSearch` to find matches, then `regex`'s `Regex` to grab the captures. Although, one disadvantage is that you need to either have 2 two regexes or strip out the capture groups for `RegexSearch`. Also, I'm not sure how flexible `alacritty_terminal` is regarding limiting the search region, so there might be other complexity introduced around those. So, overall, I'm okay with it as it is for now.

probably-neb:
Is this ready to go @davewa? Or are you planning to knock out those "could be const" reviews before merging?

davewa:
Oh? I resolved those. Hhmmm….. It’s ready anyhow.

probably-neb:
Great work as usual! Thanks!

PR #44723, opened 2025-12-12
Add `keyword.import` capture to Python syntax highlighting

Davidhsiehtw:
Improve Python syntax highlighting with granular theme customization support. Related to #9461

Release Notes:

- Added support for 'keyword.import' in Python syntax highlighting.
- Users can now customize the color of 'import', 'from', and 'as' keywords via theme_overrides independently of other keywords.
before
after
keyword

### DIFF
--- a/crates/languages/src/python/highlights.scm
+++ b/crates/languages/src/python/highlights.scm
@@ -245,7 +245,6 @@
 ] @keyword.operator
 
 [
-  "as"
   "assert"
   "async"
   "await"
@@ -260,10 +259,8 @@
   "exec"
   "finally"
   "for"
-  "from"
   "global"
   "if"
-  "import"
   "lambda"
   "nonlocal"
   "pass"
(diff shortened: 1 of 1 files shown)


dahhs:
@cla-bot check

osiewicz (commented):
I am a bit unsure about this. On one hand, we already have a precedent for customizing highlights of other things (for Python there's `keyword.definition`), but on the other, I wish we did it more uniformly. In general, I'd love us to standardize on some way to do this across all programming languages (to agree which things should even be customizable - import statements make sense to me).

crates/languages/src/python/highlights.scm
```diff
@@ -260,10 +259,8 @@
   "exec"
   "finally"
   "for"
-  "from"
   "global"
   "if"
-  "import"
```
osiewicz, line 266: I don't think we should be removing a generic `keyword` capture. At least `keyword.definition` ones are still captures as `keyword`s as well.

dahhs: Agreed. Given the low impact and the consistency concerns, I’m okay closing this PR.

PR #44725, opened 2025-12-12
Add support for unknown supported_endpoints in copilot chat model schema

zerowidth:
This allows the model deserialization to succeed even if new endpoints are added. The copilot API may change, and we shouldn't filter out models accidentally due to deserialization errors.

This would manifest as:

```
2025-12-12T09:43:26-07:00 WARN  [copilot::copilot_chat] GitHub Copilot Chat model failed to deserialize: Error("unknown variant `/new-endpoint`, expected `/chat/completions` or `/responses`", line: 0, column: 0)
```

Release Notes:

- Improved copilot chat model deserialization to support unknown new endpoints


zerowidth:
@cla-bot check

PR #44742, opened 2025-12-12
git: Show all branches in branch picker empty state

Anthony-Eid:
This fixes an issue where a user could get confused by the branch picker because it would only show the 10 most recent branches, instead of all branches. 

Release Notes:

- git: Show all branches in branch picker when search field is empty


JosephTLyons:
Merging to get this into the next nightly

PR #38443, opened 2025-09-18
Add configurable LSP timeout

Zertsov:
Closes #36818 

Release Notes:

- Add in configurable global LSP timeouts

PR #44745, opened 2025-12-12
Add configurable LSP timeout setting

### LINKED ISSUE #36818: Support configurable LSP initialization timeout (currently 120secs)
### Summary

There is no option for how long until language server timeout

### Description
```
2025-08-23T21:55:13+02:00 ERROR [lsp] Cancelled LSP request task for "initialize" id 0 which took over 120s
2025-08-23T21:55:13+02:00 ERROR [project::lsp_store] Failed to start language server "kotlin-language-server": initializing server kotlin-language-server, id 21

Caused by:
    Request timed out
```

Steps to reproduce:
1. install [kotlin](https://github.com/zed-extensions/kotlin)
2. try running it on a huge repository like this https://github.com/keiyoushi/extensions-source

**Expected Behavior**: There is a option I can turn up
**Actual Behavior**: I cannot actually do anything and have to use another IDE for lsp

Zed: v0.201.0 (Zed Dev fa61c3e24d8893a8a62ba0e46dba48e9cc4ae8bd) 
OS: Windows 10.0.22635
Memory: 63.7 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics || Intel Corporation || 31.0.101.3222

Bertie690:
Fixes #36818

Release Notes:

- Added new `global_lsp_settings.request_timeout` setting to configure the maximum timeout duration for LSP-related operations.

Code inspired by [prior implementation](https://github.com/zed-industries/zed/pull/38443), though with a few tweaks here & there (like using `serde:default` and keeping the pre-defined constant in the LSP file).


Bertie690:
@cla-bot check

Cargo.toml
```diff
@@ -266,6 +266,7 @@ clock = { path = "crates/clock" }
 cloud_api_client = { path = "crates/cloud_api_client" }
 cloud_api_types = { path = "crates/cloud_api_types" }
 cloud_llm_client = { path = "crates/cloud_llm_client" }
+cloud_zeta2_prompt = { path = "crates/cloud_zeta2_prompt" }
```
SomeoneToIgnore: This seems unrelated.

crates/project/src/project_settings.rs
```diff
@@ -106,12 +106,37 @@ impl From<settings::NodeBinarySettings> for NodeBinarySettings {
...
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(default)]
 pub struct GlobalLspSettings {
     /// Whether to show the LSP servers button in the status bar.
     ///
     /// Default: `true`
     pub button: bool,
+    /// The maximum amount of time to wait for responses from language servers, in seconds.
+    /// Ignored if set to `0`.
```
SomeoneToIgnore: > Ignored if set to `0`.

0s timeout is no timeout, right?
Can we remove this tautology from here and other places?

assets/settings/default.json
```diff
@@ -2097,9 +2097,13 @@
     },
   },
   // Common language server settings.
+  //
+  // Note that modifying most properties will require restarting the language servers to come into effect.
   "global_lsp_settings": {
```
SomeoneToIgnore, line 2230: We need to have the docs adjusted in the `docs/src/configuring-zed.md` file: that will be on zed.dev/docs and some people read only that instead of default settings.

crates/project/src/project_settings.rs
```diff
@@ -106,12 +106,37 @@ impl From<settings::NodeBinarySettings> for NodeBinarySettings {
 }
 
 /// Common language server settings.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(default)]
```
SomeoneToIgnore, line 122: Why do we need extra defaults now?

Also note that serde's default and `impl Default` are adding totally different values: that seems like an error.

crates/lsp/src/lsp.rs
```diff
@@ -45,7 +46,18 @@ use util::{ConnectionResult, ResultExt, TryFutureExt, redact};
...
+/// The default amount of time to wait while initializing or fetching LSP servers, in seconds.
+///
+/// Should not be used in most cases and is exported solely for use inside ProjectSettings defaults.
+pub const DEFAULT_LSP_REQUEST_TIMEOUT_SECS: u64 = 120;
+/// A timeout representing the value of [DEFAULT_LSP_REQUEST_TIMEOUT_SECS].
+///
+/// Should **only be used as a fallback** when a corresponding config value cannot be obtained!
+pub const DEFAULT_LSP_REQUEST_TIMEOUT: Duration =
+    Duration::from_secs(DEFAULT_LSP_REQUEST_TIMEOUT_SECS);
+// TODO: Figure out how to configure lsp_store to adhere to relevant config setting instead of
```
SomeoneToIgnore: It's been a month since the submission, any progress here?

Alternatively: any point to keep this TODO?
I suspect you'll never be able to deal with Prettier and Copilot in this paradigm, and it's a relatively small change overall, so seems ok at its current form.

crates/lsp/src/lsp.rs
```diff
@@ -45,7 +46,18 @@ use util::{ConnectionResult, ResultExt, TryFutureExt, redact};
...
 const CONTENT_LEN_HEADER: &str = "Content-Length: ";
 
-pub const LSP_REQUEST_TIMEOUT: Duration = Duration::from_secs(60 * 2);
+/// The default amount of time to wait while initializing or fetching LSP servers, in seconds.
+///
+/// Should not be used in most cases and is exported solely for use inside ProjectSettings defaults.
+pub const DEFAULT_LSP_REQUEST_TIMEOUT_SECS: u64 = 120;
+/// A timeout representing the value of [DEFAULT_LSP_REQUEST_TIMEOUT_SECS].
+///
+/// Should **only be used as a fallback** when a corresponding config value cannot be obtained!
```
SomeoneToIgnore: That is not true already, given that tests and all remote methods use the timeout still:

Image

Which is also odd: cannot we use the configured timeout there too?

crates/lsp/src/lsp.rs
```diff
@@ -935,62 +953,63 @@ impl LanguageServer {
...
-                            ConnectionResult::ConnectionReset => {
-                                log::warn!("language server {name} (id {server_id}) closed the shutdown request connection");
-                            },
-                            ConnectionResult::Result(Err(e)) => {
-                                log::error!("Shutdown request failure, server {name} (id {server_id}): {e:#}");
-                            },
-                            ConnectionResult::Result(Ok(())) => {}
-                        }
-                    }
+        let Some(tasks) = self.io_tasks.lock().take() else {
```
SomeoneToIgnore: We can do even shorter and `?`, cannot we?

crates/project/src/lsp_store.rs
```diff
@@ -13039,8 +13044,7 @@ fn lsp_workspace_diagnostics_refresh(
                 };
 
                 progress_rx.try_recv().ok();
-                let timer =
-                    LanguageServer::default_request_timer(cx.background_executor().clone()).fuse();
+                let timer = server.request_timer().fuse();
```
SomeoneToIgnore: Note: this is somewhat tricky, as workspace diagnostics' fetch might take quite some time, and if the person decreases the global timeout, they will loop this fetch aimlessly and endlessly.

Previous 2 mins made sense, but now it gets trickier.
Maybe, we should do some increment instead, that will ramp up to `2xdefault_timeout` max?

SomeoneToIgnore (commented):
Thank you, this looks quite good overall.

One confusing point for me are the remote requests: using a hardcoded default timeout there seems odd?
Maybe I miss some context there, as not sure at all what has to be done there.

Cargo.toml
Bertie690: that... may have been a typo????

I have genuinely no idea how that got there

crates/project/src/project_settings.rs
Bertie690: I'd argue it isn't - 0 timeout could just as well mean "instant timeout=crash LSP"

crates/lsp/src/lsp.rs
Bertie690: Right - I forgot that existed

crates/project/src/project_settings.rs
Bertie690: > When deserializing, any missing fields should be filled in from the struct's implementation of Default. Only allowed on structs.
https://serde.rs/container-attrs.html#default

@SomeoneToIgnore Am I reading this wrong?

(I wanted to use `std::Default` instead of declaring a single use function whose sole job is to return an externally declared constant.)

assets/settings/default.json
Bertie690: Is that file not auto updated?

Strange.

crates/lsp/src/lsp.rs
Bertie690: The operative clause is _should_, not _does_.

I do think the tests should still use the default, but I'll get around to fixing the LSP store stuff later.

crates/lsp/src/lsp.rs
Bertie690: To put it simply: _I completely forgot about that._.

Might try my hand again once I have the time.

crates/project/src/project_settings.rs
SomeoneToIgnore: I would never accept that a setting may cause a crash, I'd say, but I start to see the point.

Let's at least rephrase this into something like "0 will wait infinitely for the requests to finish".
As "ignored" is also somewhat confusing to me still.

crates/lsp/src/lsp.rs
Bertie690: > Alternatively: any point to keep this TODO?

@SomeoneToIgnore Sanity check, am I reading this right as in "we don't like keeping TODOs in the code"?

assets/settings/default.json
SomeoneToIgnore: Well, you see all the related code around — nothing to auto update that, and it has quite a different formatting, which is not possible to derive from anywhere.

crates/project/src/project_settings.rs
Bertie690: Yeah that would be better.
0="no timeout"

crates/project/src/project_settings.rs
SomeoneToIgnore: I'm also lacking the context hence asking.

To start, I would want to understand why the current code is fine with 0 defaults, serde or not — and why do we need those now.

> https://serde.rs/container-attrs.html#default

That part is actually somewhat new to me, thank you for pointing out!

crates/lsp/src/lsp.rs
SomeoneToIgnore: We can keep TODOs in the code (and we do), but what is the point of this one is not clear to me: the code change is relatively small, and TODO seems not possible to achieve in general?

crates/project/src/project_settings.rs
Bertie690: > To start, I would want to understand why the current code is fine with 0 defaults, serde or not — and why do we need those now.

The way I see it is less of a "need vs not need" as a "different means to the same end". The prior method of adding attributes to individual properties _works fine_... but requires uniquely named functions for every single distinct value with a unique default.

If I wanted to use a property annotation for `lsp_timeout_secs`, it would require something like this:
```rust
fn default_lsp_timeout_secs() -> u8 {
  DEFAULT_LSP_TIMEOUT_SECS
}
```
(I hope I don't need to explain why I find functions returning bare constants a bit silly.)

Moreover, having default values for settings objects would make for quite good ergonomics with [`unwrap-or-default`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default) if a setting (for whatever reason) cannot be parsed.

@SomeoneToIgnore Bear in mind I am very much lacking knowledge about the greater Zed codebase, so your decision likely trumps mine.
If you would prefer the attributes go inline, I can change it fairly easily - just lmk.

assets/settings/default.json
Bertie690: @SomeoneToIgnore
After looking at the file, it might make sense to create a GitHub Actions workflow to enforce that all settings have corresponding headered sections inside `configuring-zed.md` (thus reminding contributors who add new settings to update the file).

(We could instead verifying that the listed defaults match their provided values inside `default.json`, or even do both at the same time. Just a thought.)

crates/lsp/src/lsp.rs
```diff
@@ -1362,15 +1385,39 @@ impl LanguageServer {
...
-        executor
-            .timer(LSP_REQUEST_TIMEOUT)
-            .map(|_| format!("which took over {LSP_REQUEST_TIMEOUT:?}"))
+    /// Internal function to return a Future from a configured timeout duration.
+    /// If the duration is zero or `Duration::MAX`, the returned future never completes.
+    fn request_timeout_future(
+        executor: BackgroundExecutor,
+        request_timeout: Duration,
+    ) -> impl Future<Output = String> {
+        if request_timeout == Duration::MAX || request_timeout == Default::default() {
```
SomeoneToIgnore:
```suggestion
        if request_timeout == Duration::MAX || request_timeout == Duration::ZERO {
```

crates/lsp/src/lsp.rs
```diff
@@ -1362,15 +1385,39 @@ impl LanguageServer {
...
+
+        Either::Right(
+            executor
+                .timer(request_timeout)
+                .map(move |_| format!("which took over {request_timeout:?}")),
+        )
+    }
+
+    /// Obtain a request timer for the LSP.
+    /// Accepts an optional minimum timeout duration to use.
```
SomeoneToIgnore:
```suggestion
    /// If specified, runs no less than `min_timeout`.
```

crates/project/src/lsp_store.rs
```diff
@@ -13333,8 +13377,9 @@ fn lsp_workspace_diagnostics_refresh(
                 };
 
                 progress_rx.try_recv().ok();
-                let timer =
-                    LanguageServer::default_request_timer(cx.background_executor().clone()).fuse();
+                // Clamp timeout duration at a minimum of 120ms to mitigate infinite loops from re-trying connections.
```
```suggestion
                // Clamp timeout duration at a minimum of [`DEFAULT_LSP_REQUEST_TIMEOUT`] to mitigate usless loops from re-trying connections with to smaller timeouts from the settings.
```

It's 120s, not ms.

crates/project/src/project_settings.rs
```diff
@@ -116,12 +116,41 @@ impl From<settings::NodeBinarySettings> for NodeBinarySettings {
...
+    }
+}
+
+impl GlobalLspSettings {
+    /// Returns the timeout duration for LSP-related interactions, or Duration::ZERO if no timeout should be applied.
+    /// Zero durations are treated as no timeout by language servers, so code using this in an async context can
+    /// simply call unwrap_or_default.
+    pub const fn get_request_timeout(&self) -> Duration {
+        match self.request_timeout {
+            0 => Duration::ZERO,
```
SomeoneToIgnore: Hm, `Duration::from_secs(0)` is ` Duration::ZERO`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fc809c309c42f49cb0c50247971a1ad7

So let's simplify that.

crates/settings/src/settings_content/project.rs
```diff
@@ -189,6 +189,11 @@ pub struct GlobalLspSettingsContent {
     ///
     /// Default: `true`
     pub button: Option<bool>,
+    /// The maximum amount of time to wait for responses from language servers, in seconds.
+    /// Ignored if set to `0`.
```
SomeoneToIgnore:
```suggestion
    /// A value of `0` will result in no timeout being applied (causing all LSP responses to wait indefinitely until completed).
```

SomeoneToIgnore (commented):
Nice work!

I have a few NITs mostly of which I can fix even via GH UI, but the ZERO-related one requires another commit.
I can fix that some time later after [re]viewing all other PRs assigned now, or feel free to do so earlier.

On a general scale: 

* seems that tests and related entities like `FakeLanguageServer` could benefit from the same "take the value out of the current project settings" approach: in fact, we might even want to add a test on that now?

* we could restart all servers if the timeouts change: there's https://github.com/zed-industries/zed/blob/b06f4088a3565c5e30663106ff79c1ced645d87a/crates/language_tools/src/lsp_button.rs#L206-L208 and `BufferStore` holds all the buffers available: feels that it can be a method in the `LspStore` even.

What seems even more desirable is the ability to get the timeout from settings every time a new request is sent.
The newest, nice, changes that removed the default timeout from the upstream client calls does exactly that and requires no Zed restart.
What if https://github.com/zed-industries/zed/blob/b06f4088a3565c5e30663106ff79c1ced645d87a/crates/lsp/src/lsp.rs#L1206 gets to receive a `cx: &App`, would that solve the need for restart?

----------------------

I think the last thing deal with before merging is the 2nd bullet, but the first one is rather mechanical to fix along the way.

crates/project/src/lsp_store.rs
Bertie690:
```suggestion
                // Clamp timeout duration at a minimum of [`DEFAULT_LSP_REQUEST_TIMEOUT`] to mitigate useless loops from re-trying connections with smaller timeouts from project settings.
```

This good?
(Commenting also bc you made a typo)

SomeoneToIgnore: Oh yeah, thank you.

Bertie690:
@SomeoneToIgnore 
I tried to make `LSP::request` take an `&app` and it produced lots of errors about async stuff inside LSP store code.
The name of the type involved is `std::rc::Rc<(dyn for<'a, 'b, 'c> Fn(&'a ModifiersChangedEvent, &'b mut gpui::Window, &'c mut App) + 'static)>` (which is just lovely to parse)

I'mma just do the simpler thing and pass the `Duration` from the calling functions every time `request` is called .

(As a side note, this also avoids circ dep issues with `ProjectSettings` importing `LanguageServerName` and similar. I do not know if it can be removed or if I really want to at this moment, so yeah I'mma take the simple path tyvm)

Bertie690:
(I have the changes locally, will push likely tmr)

SomeoneToIgnore (commented):
Superb, seems that we are able to get the timeout each time we request things, and mostly those are in the test code.

Besides fixing the rest of the test code, what's left is: 

* that note added into `default.json` change about the restart is obsolete now
* `get_request_timeout` could be just `request_timeout` to match the general style
* given that even fake servers have to do the same settings access, now we can write a small test on any LSP request timeout: `editor_tests.rs` has a lot of examples how to change the settings and handle requests

Bertie690:
1. is done, 3 I have written (and am attempting to run on my potato lol)

Bertie690:
Thanks for helping - I've fallen ill as of late and have struggled to find the time/wherewithal to fix the remaining edge cases.

SomeoneToIgnore (approved):

PR #44746, opened 2025-12-12
Fix dock panel button tooltip not dismissed when state changes via keyboard shortcut

### LINKED ISSUE #44720: Dock panel button tooltip doesn't update when state changes via keyboard shortcut
### Reproduction steps

1. Start Zed
2. Hover over a dock panel button (e.g., the Debug Panel button in the bottom toolbar)
3. Wait for the tooltip to appear (shows "Debug Panel ⇧⌘D")
4. While still hovering, press the keyboard shortcut to open the panel (⇧⌘D)
5. Observe that the tooltip still shows "Debug Panel ⇧⌘D" instead of updating

**Current behavior:**
When hovering over a dock panel button and using the keyboard shortcut to toggle the panel, the tooltip text remains unchanged. It continues to show "Debug Panel ⌘⇧D" even after the panel opens.

Before - shows Debug Panel tooltip
Image

After pressing ⌘⇧D while hovering:

Image

**Expected behavior:**
I feel like the tooltip should dynamically update to reflect the new state. After the panel opens, the tooltip should change to "Close Bottom Dock ⌘J".

Image

I was able to reproduce this behavior for other dock panel buttons

OS: macOS 15.6.1
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Daeksell:
Closes #44720

Release Notes:

- Fixed dock panel button tooltips not being dismissed when toggling panels via keyboard shortcut

**Problem:** When hovering over a dock panel button and using a keyboard shortcut to toggle the panel, the tooltip remains visible with stale content. This is inconsistent with mouse click behavior, where the tooltip is dismissed on mouse down.

**Solution:** Include the panel's active state in the button's element ID. When the state changes, the element ID changes (e.g., `"DebugPanel"` → `"DebugPanel-active"`), which causes GPUI to discard the old element state including the cached tooltip.

**Testing:** Manually verified:
1. Hover over a dock panel button, wait for tooltip
2. Press keyboard shortcut to toggle the panel
3. Tooltip is now dismissed (consistent with mouse click behavior)

https://github.com/user-attachments/assets/ed92fb6c-6c22-44e2-87e3-5461d35f7106

### DIFF
--- a/crates/workspace/src/dock.rs
+++ b/crates/workspace/src/dock.rs
@@ -941,7 +941,14 @@ impl Render for PanelButtons {
                         .anchor(menu_anchor)
                         .attach(menu_attach)
                         .trigger(move |is_active, _window, _cx| {
-                            IconButton::new(name, icon)
+                            // Include active state in element ID to invalidate the cached
+                            // tooltip when panel state changes (e.g., via keyboard shortcut)
+                            let button_id = if is_active_button {
+                                ElementId::Name(format!("{}-active", name).into())
+                            } else {
+                                ElementId::Name(name.into())
+                            };
+                            IconButton::new(button_id, icon)
                                 .icon_size(IconSize::Small)
                                 .toggle_state(is_active_button)
                                 .on_click({
--- a/crates/workspace/src/dock.rs
+++ b/crates/workspace/src/dock.rs
@@ -943,12 +943,7 @@ impl Render for PanelButtons {
                         .trigger(move |is_active, _window, _cx| {
                             // Include active state in element ID to invalidate the cached
                             // tooltip when panel state changes (e.g., via keyboard shortcut)
-                            let button_id = if is_active_button {
-                                ElementId::Name(format!("{}-active", name).into())
-                            } else {
-                                ElementId::Name(name.into())
-                            };
-                            IconButton::new(button_id, icon)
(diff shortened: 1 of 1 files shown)


MrSubidubi:
Thanks for this!

I just pushed a minor update so that we do not allocate a string but use a different type of `ElementId` instead.

There is one thing left to discuss for me: Do we want to stick with this approach or go with a different one? Currently, we just hide the tooltip, but an alternative would be that we update the tooltip based on the current panel state so that it continues to exist and the accurate keystroke/action is displayed. What do you think about that?

Daeksell:
Hi, thanks! When I initially wrote the issue, I thought the tooltip should be updated. However, I wasn't able to implement it due to my limited understanding of the code. Since this behavior is somewhat obscure, I think simply hiding the tooltip would be sufficient. This approach is also consistent with the mouse behavior.

MrSubidubi (approved):
Looks good. I think there is a slight difference between the mouse and keyboard interaction, that's why I was a bit hesistant to merge this this way. Yet, the other way is way more complicated and just dismissing is perhaps not the 100% solution, but 99% for sure 🙂 

Thank you for looking into and fixing this!

PR #44747, opened 2025-12-12
git_ui: Show missing right border on selected items

devzeth:
For folders and files basically any selected item in the git panel we draw a border around it. The issue is that the right side of this border wasn't ever visible. 

In the project_panel.rs file I've saw that the decision was to make the right side border 2 pixels. And this panel doesn't have this issue, no matter which side of the dock is selected. So it was a very easy `look at how we did x do y`. 

Before: 
image
After: 
image

I don't think it warrants a release note. 

Release Notes:

- N/A


danilo-leal (approved):
Great catch, thanks!

PR #44763, opened 2025-12-13
gpui: Implement hide/unhide other apps on windows

cppcoffee:
Release Notes:

- N/A


Veykril:
I don't think this can be reasonably implemented on windows (given the concept of a window is a bit odd here, as is evident by the extra filter logic). I also don't really know of any other App that offers this on windows so I think we are fine with not implementing this for this platform.

Thanks for proposing this either way though! And maybe we'll change our mind on this (or another team member disagrees with me here), then we can dig this PR up again

PR #44769, opened 2025-12-13
editor: Add align selections action

### LINKED ISSUE #17090: Command to align cursors
### Check for existing issues

- [X] Completed

### Describe the feature

I'm missing an equivalent of Helix's `align_selections` command - something that will align my cursors vertically, inserting spaces before each cursor that is shifted left relative to the main cursor.

I have no idea what Helix does here when you have multiple cursors on the same line.

_No response_

tiagolobao:
Closes #17090

Hello! This is a feature that I really wanted, so I've created an implementation of it. Hope u guys like it.
It was alto mentioned at [this discussion](https://github.com/zed-industries/zed/discussions/30473)

The use cases were based on the [vscode extension](https://marketplace.visualstudio.com/items?itemName=yo1dog.cursor-align) (only spaces case was implemented)
- The algorithm is O(n + m × 2k), where:
  - n = total cursors (for initial analysis, to understand how many alignments will be performed)
  - m = number of alignment columns  
  - k = number of alignment rows (first time to get the most-right occurrence, the next one to create the `Vec` of edits)

Release Notes:

- Added `editor: align selections` command

ezgif-524b631d71833ba1

<br>

Thank you for creating zed!


tiagolobao:
I've just tried to avoid the multiple calls for `self.transaction`, by calculating the offset that the alignment will cause, but apparently it applies the edits "all at once", then I get this problem: 

```
thread 'main' (44587) panicked at crates/rope/src/chunk.rs:355:13:
point Point(0:9) extends beyond row for string "a      a\n a     a\n  a     a\n  "
```

I'm sure that the order of the `edits: Vec<(Range<Point>, String>` are ok, (by column). So I guess I need several transactions for it to work. If I'm missing smth, let me know

Veykril:
Not having looked at the PR but you likely will have more luck by using Anchors instead of Points here

tiagolobao:
> Not having looked at the PR but you likely will have more luck by using Anchors instead of Points here

It worked! thanks for the advice! @Veykril

crates/editor/src/editor.rs
```diff
@@ -11401,6 +11401,98 @@ impl Editor {
         }
     }
 
+    pub fn align_selections(
```
Veykril: Putting it between enable_breakpoints and disable_breakpoints might not be the best choice regarding code locality 😅

SomeoneToIgnore (approved):
Sorry for the late reply, and thank you for the PR, looks nice.

PR #39095, opened 2025-09-29
Add Cursor Blink Animation Options

JonGretar:
This PR adds configurable cursor blink animations to provide visual variety beyond the standard on/off blinking.

## New Animation Types

Users can now set `cursor_blink_animation` in their settings to one of:

- `"instant"` - Classic binary on/off (default)
- `"fade"` - Smooth opacity transition  
- `"pulse"` - Fade in → hold → fade out pattern
- `"zoom"` - Scale and fade with position-aware anchoring
- `"slide"` - Horizontal slide in/out from left
- `"breathe"` - Gentle organic pulsing

## Usage

```json
{
  "cursor_blink_animation": "fade"
}
```

Each animation uses optimized timing and easing curves. The zoom animation includes smart positioning that anchors bar cursors to the left edge, underline cursors to the bottom, and block cursors to the center for natural-looking scaling effects.

# Release Notes:

- Added animation options for cursor blinking

PR #44770, opened 2025-12-13
editor: Add smooth cursor animation

### LINKED ISSUE #4991: Animated/smooth cursor movement
### Check for existing issues

- [X] Completed

### Describe the feature

Similar to, but not the same as, zed-industries/zed#5179

VSCode and Neovide have the option to animate the cursor motion.

I don't know if I'd use this feature myself, but it _looks_ great in demos and seems like a fun way to flex Zed's performance.

This is likely more useful when Vim mode is more fleshed out because it makes following the cursor easier for people newer to modal editing.

Neovide does it _beautifully_. 
Animated cursor

aiatsuk:
Implement Neovide-style smooth cursor movement for editor and terminal. The cursor smoothly animates to new positions using inertial physics.

Controlled by `smooth_caret` setting (true/false).

Closes https://github.com/zed-industries/zed/issues/4991

DEMO:

https://github.com/user-attachments/assets/3f58e4ec-490a-4322-a71b-495a9946d78d

Release Notes:

- Added smooth cursor animation for editor and terminal

Before merge:
- [ ] Fix wrong character render https://github.com/zed-industries/zed/pull/44770#issuecomment-3679462002

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -200,6 +200,13 @@
   "use_system_prompts": true,
   // Whether the cursor blinks in the editor.
   "cursor_blink": true,
+  // Enable smooth cursor animation.
+  "smooth_caret": false,
+  // Cursor visual effects (particle animations).
+  // Available modes: "none", "sonicboom"
+  "cursor_vfx": {
+    "mode": "none"
+  },
   // Cursor shape for the default editor.
   //  1. A vertical bar
   //     "bar"
--- a/crates/editor/src/blink_manager.rs
+++ b/crates/editor/src/blink_manager.rs
@@ -1,7 +1,9 @@
 use gpui::Context;
 use settings::SettingsStore;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use ui::App;
+
+const SMOOTH_BLINK_FADE_DURATION_MS: f32 = 150.0;
 
 pub struct BlinkManager {
     blink_interval: Duration,
@@ -14,6 +16,14 @@ pub struct BlinkManager {
     enabled: bool,
     /// Whether the blinking is enabled in the settings.
     blink_enabled_in_settings: fn(&App) -> bool,
+    /// Whether smooth blink transitions are enabled.
+    smooth_blink_enabled: bool,
+    /// Current opacity for smooth blink transitions (0.0-1.0).
+    current_opacity: f32,
+    /// Target opacity for smooth blink transitions.
+    target_opacity: f32,
+    /// Last update time for smooth blink interpolation.
+    last_smooth_blink_update: Option<Instant>,
 }
 
 impl BlinkManager {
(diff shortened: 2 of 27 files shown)


aiatsuk:
@cla-bot check

Dima-369:
Thanks! I wondered:

With `"smooth_caret": true` (should definitely not be `true` by default) why is only `crates/editor/src/inertial_cursor.rs` used? There is also `crates/editor/src/cursor_vfx.rs`, but it can not be enabled in `settings.json` and the code does not use it directly?

So, it would be great if there would be Zed settings for smooth cursor configuration, basically for `InertialCursorConfig`. And since there is the VFX system already, how about user settings to actually use the VFX code and making `CursorVfxConfig` configurable in Zed settings?

Also, can you please provide screenshots/videos?

MrSubidubi:
Thanks for this! Aside from the test failures, I still think that https://github.com/zed-industries/zed/pull/39095#issuecomment-3352910885 very much applies here - you use `window.request_animation_frame`, which basically calls a `notify` for the entity underneath. This would mean that we constantly rerender the entirety of the editor for minimal work every frame, which would be quite expensive.

Otherwise, I'd very much agree with what was said above, so please also provide some context on that. Thanks!

aiatsuk:
> With `"smooth_caret": true` (should definitely not be `true` by default) why is only `crates/editor/src/inertial_cursor.rs` used? There is also `crates/editor/src/cursor_vfx.rs`, but it can not be enabled in `settings.json` and the code does not use it directly?

You are absolutely right, I'll change that. 

> So, it would be great if there would be Zed settings for smooth cursor configuration, basically for `InertialCursorConfig`. And since there is the VFX system already, how about user settings to actually use the VFX code and making `CursorVfxConfig` configurable in Zed settings?

Currently, I'm simplifying the VFX system and remained for now only two options and added this to settings UI. I don't know if it's required to add the detailed configuration for the inertial cursor and for the VFX, so I propose to leave for now the toggle option and the select option for the effects. 

> Also, can you please provide screenshots/videos?

Sure, I'll add provide demo video

aiatsuk:
@Dima-369, added video demo

https://github.com/zed-industries/zed/pull/44770#issue-3725943064

Azarattum:
The demo looks great! Does it work when jumping between multiple buffers? A big part of why it's so useful in neovide is the highlight of long-distance jumps (e.g. between side-by-side editors, terminal, file tree). This is a great accessibility feature to quickly follow where the cursor is in a new buffer.

JotaRata:
What metrics were used to determine this is not useful enough, or that is not going to be used by many people?

This decision is so disappointing coming from a great editor like Zed.

wxmup:
Very disappointing. It fails both the efforts of the PR contributor and our expectations for this feature。

Trowee:
Still hoping to see this land in Zed. It would be a big shame to not implement it since it looks so nice.

aurexav:
Frankly, people are gravitating towards fully autonomous coding tools. AI features in editors will become increasingly obsolete. I rarely write code these days; Codex does it all. You have to admit that. Furthermore, OpenAI has published their harness engineering methodology, which is likely to define the future for the next three months.

I only use Zed for some document review and minor fixes. The only advantage it offers me is a smooth and fast editing experience, which is why I want this.

As a user and contributor, I’ve decided to move to Ghostty. It’s sad.

zortax:
> but could it be a possibility that a user-defined shader is loaded

Have fun writing the shader 3 times (wgpu, metal, DirectX) 😅

Azarattum:
@Kethku, you gave us all hope [back then](https://github.com/zed-industries/zed/issues/4991#issuecomment-1908639780). What is your opinion on this? I understand that your are no longer involved with the Zed project, are you?

aiatsuk:
Fun fact the Zed team added smooth cursor animation after all, despite the view that "I don't think this is useful enough" :)

acgxv:
> Fun fact the Zed team added smooth cursor animation after all, despite the view that "I don't think this is useful enough" :)

Finally, they recognized this.  

Yet I rarely open the editor now, except when I choose to adjust an agent config toml by hand. It is unfortunate.

Issue #8043, opened 2024-02-20
GPUI overdraws too much


kvark:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

![zed-overdraw](https://github.com/zed-industries/zed/assets/107301/6c684f60-8de4-41ce-b9d2-e6a8e2018872)

It looks like most "ordinary" pixels are being overdraw 5-6 times on screen. See screenshot from RenderDoc's "pass overdrawn" mode. I looked at the green area (5 draws per pixel) and at the yellow one on the left (6 draws per pixel). This is burning GPU power too much unnecessarily and draining the battery.

Ideally, these areas would have 0 to 1 draw. 0 is possible because the render pass is starting with a clear color. For elements overlapping each other, there needs to be some technique for pixel rejection employed. Generally it's depth or stencil testing. For example, WebRender draws opaque elements from front to back and writes to depth, preventing deeper layers from overwriting the pixels of the front layers.

### Environment

Linux (but it unlikely matters)

jansol:
I'm not sure GPUI has anything fully opaque? Since practically all elements are antialiased rounded rectangles which require alpha blending.

kvark:
@jansol not all elements have rounded corners. For example, #7231 makes the fast path for such quads that are just quads, and it makes a good difference.
In WebRender we'd also split a rounded-cornered element into a "9 patch", which consists of 5 fully opaque rectangles and 4 semi-transparent rectangles. This allows drawing most of the element early as opaque.

jansol:
Right, makes sense. And I assume it just skips elements with a non-opaque background color.

jansol:
AFAIK there has been no change to this.

someone13574:
> In WebRender we'd also split a rounded-cornered element into a "9 patch", which consists of 5 fully opaque rectangles and 4 semi-transparent rectangles.

Wouldn't you only need three fully opaque?

kvark:
Yes, you can make just 3 opaque.

jansol:
No change to this.

jansol:
Still no changes here...

xraxra:
this might be related to why I see GPU usage jump to ~75% usage and ~50watts here when Zed's window is visible.  I was going to post in the other [issue](https://github.com/zed-industries/zed/issues/40189)

[...]

PR #44771, opened 2025-12-13
gpui: Implement depth testing on metal

marcocondrache:
This PR partially addresses #8043 on macOS

Still a WIP

The current renderer relies on the painter’s algorithm for draw ordering, which results in significant overdraw. Large portions of the frame are shaded multiple times even though only the frontmost fragments are visible, increasing GPU and CPU usage

This change enables depth testing and adjusts rendering to rely on depth buffering rather than strict back-to-front ordering. This allows the GPU to discard occluded fragments earlier in the pipeline, reducing unnecessary work.

Ref: https://developer.apple.com/documentation/metal/calculating-primitive-visibility-using-depth-testing

Release Notes:

- Reduced gpu and battery usage due to overdrawing

### DIFF
--- a/crates/gpui/src/platform/mac/metal_renderer.rs
+++ b/crates/gpui/src/platform/mac/metal_renderer.rs
@@ -117,6 +117,8 @@ pub(crate) struct MetalRenderer {
     path_intermediate_texture: Option<metal::Texture>,
     path_intermediate_msaa_texture: Option<metal::Texture>,
     path_sample_count: u32,
+    depth_texture: Option<metal::Texture>,
+    depth_stencil_state: metal::DepthStencilState,
 }
 
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


HigherOrderLogic:
Hi, I got to this PR through the linked issue.

My 2 cents: shouldnt this be solved with a more general solution such as damage tracking and re-rendering only the changed part. I think that option would be much more performant, and also work on other platforms such as Linux.

marcocondrache:
@HigherOrderLogic Damage tracking and depth testing solve orthogonal problems: 

- Depth testing reduces overdraw within a single frame. When you have overlapping elements each layer would write to the same pixels. Depth testing lets the gpu discard fragments behind already-rendered content
- Damage tracking reduces repainting across frames. It avoids redrawing unchanged regions but doesn't help with overdraw within the regions that do need redrawing.

Issue #8043 specifically reports overdraw per frame (and overlapping elements), this is exactly what depth testing addresses. Even with perfect damage tracking, the same overdraw would occur within the damaged regions.

Both optimizations are complementary and valuable, I can try to implement damage tracking separately to further improve performance. This PR focuses on the overdraw problem specifically

PR #44780, opened 2025-12-13
docs: Add learning rust section

mchisolm0:
DEMO ONLY

It could be helpful to point new contributors to quickly learn how rust works.

Feel free to close this PR without explanation if it is not wanted. If you think it is helpful, I think we should probably only merge *one* of these commits, though, I am open to input.


SomeoneToIgnore:
Peculiar to see `rust`, `TypeScript` and `C++` capitalization in the same paragraph.

Apart from that, thank you, but no, thank you.

PR #44783, opened 2025-12-13
Redact environment variables from language server spawn errors

notpeter:
Redact environment variables from zed logs when lsp fails to spawn.

Release Notes:

- N/A


notpeter:
Error observed:

Screenshot 2025-12-13 at 3 48 16 PM

Test with:
```json
  "lsp": {
    "clangd": {
      "binary": {
        "path": "",
      },
    },
  },
```
CC: @SomeoneToIgnore

SomeoneToIgnore (approved):
Sorry for the delay, it was a rough week (or two?).

I feel that overall, we need to do something better than matching arbitrary strings with regexes, but that's a start and since we do that on errors only, let's try and see how things work.

PR #44785, opened 2025-12-13
workspace: Persist window values without project

mchisolm0:
Persist and restore window values (size, position, etc.) to the KV Store when there are no projects open.

Relates to Discussion https://github.com/zed-industries/zed/discussions/24228#discussioncomment-15224666

Release Notes:

- Added persistence for window size when no projects are open

### DIFF
--- a/crates/gpui/src/platform.rs
+++ b/crates/gpui/src/platform.rs
@@ -1258,7 +1258,7 @@ pub(crate) struct WindowParams {
 }
 
 /// Represents the status of how a window should be opened.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub enum WindowBounds {
     /// Indicates that the window should open in a windowed state with the given bounds.
     Windowed(Bounds<Pixels>),
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -8554,10 +8554,7 @@ pub fn remote_workspace_position_from_db(
             let restorable_bounds = serialized_workspace
                 .as_ref()
                 .and_then(|workspace| Some((workspace.display?, workspace.window_bounds?)))
-                .or_else(|| {
-                    let (display, window_bounds) = DB.last_window().log_err()?;
-                    Some((display?, window_bounds?))
-                });
+                .or_else(|| persistence::read_no_project_window_bounds());
 
             if let Some((serialized_display, serialized_status)) = restorable_bounds {
                 (Some(serialized_status.0), Some(serialized_display))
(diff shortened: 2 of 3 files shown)


mchisolm0:
Here is a video of the window size and position persisting when File > New Window is chosen (so no project is selected). I have confirmed it also works on MacOS when you completely quit and reopen the app.

<details>
<summary>Video demo</summary>
<video src="https://github.com/user-attachments/assets/a9adce1f-7a59-4316-bc6c-0d0a7e69294b
">VIdeo unavailable</video>
</details>

crates/gpui/src/platform.rs
```diff
@@ -1258,7 +1258,7 @@ pub(crate) struct WindowParams {
 }
 
 /// Represents the status of how a window should be opened.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
```
mchisolm0, line 1261: @ConradIrwin I did have a question when you have time. I will also set up a time to pair. In my current version, I added `Deserialize, Serialize` to the `WindowBounds` enum. 

Is there a better way for me to structure the feature so I can avoid changing WindowBounds?

ConradIrwin: The disadvantage of adding Serialize/Deserialize here is that we can never change this structure without breaking serialization. Although it's a bit of extra code (though hopefully nothing beyond what AI can build) it's less likely that we'll break things in the future if we have a specific structure/enum in the serialization code that gets converted to this (maybe we can copy the existing stuff from workspace? though tbh storing a JSON blob seems better than destructuring to a bunch of columns).

Not a huge issue though, so we could leave this and then deal with it when we want to change this enum.

ConradIrwin:
@mchisolm0 One question:

* Currently workspace has fallback code to try and find the last opened workspaces dimensions. Intuitively it might be clearer if that used this set of dimensions too (instead of "last opened workspace"). Does that make sense?

crates/workspace/src/workspace.rs
```diff
@@ -8535,10 +8554,7 @@ pub fn remote_workspace_position_from_db(
             let restorable_bounds = serialized_workspace
                 .as_ref()
                 .and_then(|workspace| Some((workspace.display?, workspace.window_bounds?)))
-                .or_else(|| {
-                    let (display, window_bounds) = DB.last_window().log_err()?;
-                    Some((display?, window_bounds?))
-                });
+                .or_else(|| persistence::read_no_project_window_bounds());
```
mchisolm0, line 8557: @ConradIrwin Is this where you meant by 
> instead of "last opened workspace"

I have not used opening remote workspaces yet, so I have not tested it beyond making sure Zed still builds.

crates/gpui/src/platform.rs
mchisolm0: If we want to consider that, I have another version of the PR here https://github.com/zed-industries/zed/pull/44937. I have run it and it also works.

I still have not tried for a remote workspace. Maybe I can try that tomorrow or later this week.

ConradIrwin:
Closing in favor of https://github.com/zed-industries/zed/pull/44937

PR #44806, opened 2025-12-14
Fix race condition in test_collaborating_with_completion

nathansobo:
The test `test_collaborating_with_completion` has a latent race condition that hasn't manifested on CI yet but could cause hangs with certain task orderings.

## The Bug

Commit `fd1494c31a` set up LSP request handlers AFTER typing the trigger character:

```rust
// Type trigger first - spawns async tasks to send completion request
editor_b.update_in(cx_b, |editor, window, cx| {
    editor.handle_input(".", window, cx);
});

// THEN set up handlers (race condition!)
fake_language_server
    .set_request_handler::<lsp::request::Completion, _, _>(...)
    .next().await.unwrap();  // Waits for handler to receive a request
```

Whether this works depends on task scheduling order, which varies by seed. If the completion request is processed before the handler is registered, the request goes to `on_unhandled_notification` which claims to handle it but sends no response, causing a hang.

## Changes

- Move handler setup BEFORE typing the trigger character
- Make `TestDispatcher::spawn_realtime` panic to prevent future non-determinism from real OS threads
- Add `execution_hash()` and `execution_count()` to TestDispatcher for debugging
- Add `DEBUG_SCHEDULER=1` logging for task execution tracing
- Document the investigation in `situation.md`

cc @localcc @SomeoneToIgnore (authors of related commits)

Release Notes:

- N/A

### DIFF
--- a/crates/collab/src/tests/editor_tests.rs
+++ b/crates/collab/src/tests/editor_tests.rs
@@ -312,6 +312,49 @@ async fn test_collaborating_with_completion(cx_a: &mut TestAppContext, cx_b: &mu
             "Rust",
             FakeLspAdapter {
                 capabilities: capabilities.clone(),
+                initializer: Some(Box::new(|fake_server| {
+                    fake_server.set_request_handler::<lsp::request::Completion, _, _>(
+                        |params, _| async move {
+                            assert_eq!(
+                                params.text_document_position.text_document.uri,
+                                lsp::Uri::from_file_path(path!("/a/main.rs")).unwrap(),
+                            );
+                            assert_eq!(
+                                params.text_document_position.position,
+                                lsp::Position::new(0, 14),
+                            );
+
+                            Ok(Some(lsp::CompletionResponse::Array(vec![
+                                lsp::CompletionItem {
+                                    label: "first_method(…)".into(),
+                                    detail: Some("fn(&mut self, B) -> C".into()),
+                                    text_edit: Some(lsp::CompletionTextEdit::Edit(lsp::TextEdit {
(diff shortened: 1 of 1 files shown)


crates/collab/src/tests/editor_tests.rs
```diff
@@ -378,19 +378,10 @@ async fn test_collaborating_with_completion(cx_a: &mut TestAppContext, cx_b: &mu
...
-        });
-        editor.handle_input(".", window, cx);
-    });
-    cx_b.focus(&editor_b);
-
-    // Receive a completion request as the host's language server.
-    // Return some completions from the host's language server.
-    cx_a.executor().start_waiting();
-    fake_language_server
+    // Set up the completion request handlers BEFORE typing the trigger character.
```
SomeoneToIgnore: I think the analysis is ok, yet most of the change is nonsense.

To me, seems that we need to use `initializer` field in `FakeLspAdapter` similar to 

https://github.com/zed-industries/zed/blob/e9073eceeba8e1a71e966582566179517591ec6b/crates/collab/src/tests/editor_tests.rs#L1953-L1985

instead of all this.

nathansobo: Cool, thanks, I didn't know about the initializer field. Just trying to make progress on porting the GPUI scheduler to our unified scheduler crate and hit this locally. I force pushed with your suggestion and it passes as well. Thanks!

SomeoneToIgnore: Sorry for posting this sloppy comment, I will push see what can I apply on top and fix it further too.

PR #44807, opened 2025-12-14
settings_ui: Fix autosave setting not being removed when restored to default

### LINKED ISSUE #44292: Don't add default values to the settings JSON file after restoring settings from the UI
### Reproduction steps

1. Open Settings.
2. Change the any setting. Example: Editor - Auto Save Mode setting to "On Focus Change"
3. Open Settings file. The `"autosave": "on_focus_change"` is created.
4. Close the settings.json (Optional) and restore it via UI.
5. Open Settings file. The `"autosave": "off"` is created.

1. Open Settings.
2. Change the any setting. Example: Editor - Auto Save Mode setting to "On Focus Change"
3. Open Settings file. The `"autosave": "on_focus_change"` is created.
4. Close the settings.json (Optional) and restore it via UI.
5. Open Settings file. The `autosave` key if not exists because use the default value.

Zed: v0.215.3-90+fda162bf3d9a11d18e07cbf2b488829450135d34 (Zed) 
OS: macOS 26.1
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

_No response_

None

cppcoffee:
Closes https://github.com/zed-industries/zed/issues/44292

Release Notes:

- Fixed autosave setting not being removed when restored to default

### DIFF
--- a/crates/settings_ui/src/page_data.rs
+++ b/crates/settings_ui/src/page_data.rs
@@ -1123,31 +1123,32 @@ pub(crate) fn settings_data(cx: &App) -> Vec<SettingsPage> {
                                             .discriminant() as usize])
                                 },
                                 write: |settings_content, value| {
-                                    let Some(value) = value else {
-                                        settings_content.workspace.autosave = None;
-                                        return;
-                                    };
-                                    let settings_value = settings_content.workspace.autosave.get_or_insert_with(|| {
-                                        settings::AutosaveSetting::Off
-                                    });
-                                    *settings_value = match value {
-                                        settings::AutosaveSettingDiscriminants::Off => {
-                                            settings::AutosaveSetting::Off
-                                        },
-                                        settings::AutosaveSettingDiscriminants::AfterDelay => {
-                                            let milliseconds = match settings_value {
-                                                settings::AutosaveSetting::AfterDelay { milliseconds } => *milliseconds,
+                                    match value {
+                                        None | Some(settings::AutosaveSettingDiscriminants::Off) => {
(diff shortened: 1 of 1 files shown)


probably-neb:
Hey, thanks for taking a look at this. I'm going to close this PR as I think it is only a band-aid fix, and rather than hard coding defaults in the already heavily hard-coded settings UI definitions, we should solve the underlying issue here.

While implementing the settings UI, we made the explicit decision that the "restore" button (the back arrow that appears next to the field title in the UI) would remove the setting if that would cause the default to be used (i.e. it would explicitly set the the field to the default value if you reset a setting to the default in a project settings file, but had a customization for that setting in your user setting file). However, we would _always_ set the value when it is set explicitly through the UI, for the simple reason that if you explicitly set a setting value, you expect it to never change even if Zed decides to change the defaults.

I still think that logic is sound, however, I understand it is subtle and non-obvious, and not necessarily the best UX in all cases. One shortcoming we have right now, is that if the field is set to the default, we don't show the reset button. 

I personally think we should keep things as they are, and not remove values from the settings files when setting them to the  default value (and if we do end up doing that, it should be consistent across _all_ settings not just the autosave setting). However, we could improve things by always showing the restore button as i mentioned above, and being clearer in the UI about which settings have been modified or inherited from the defaults (something we planned to do but was unfortunately cut from scope)

I hope that all makes sense. Feel free to ask questions, there was a lot of conversation and context around this internally that I am trying to convey clearly.

Thanks!

PR #44808, opened 2025-12-14
Improve commit view navigation

tediou5:
Closes #ISSUE

---

20251212212213_rec_-convert

The current Git panel keeps tracking uncommitted changes even when browsing the commit view, which can be quite confusing. On top of that, when looking at a complex commit that touches many files, the lack of a way to quickly navigate to a specific file makes the experience even more frustrating.

This PR aims to improve the Git panel when viewing a commit by adding path-based navigation, while reusing as many existing components and behaviors as possible. As a side note, it might be nice to also introduce navigation *within* a file's changes in the future — I haven’t thought through the details, and that would be a much larger change, so it’s probably better handled in a separate PR.

This PR also introduces some UI adjustments to the Git panel. Specifically, the staging-related UI and the commit editor are disabled while viewing a commit. These controls don’t serve any purpose in this context and can be misleading.

There is one small implementation detail worth mentioning: when creating a new `CommitView`, I currently  cache `statuses_by_path` based on the `CommitDiff`. However, `CommitDiff` doesn’t include `FileStatus`, so at the moment I’m defaulting everything to `Modified`:

```rust
let statuses_by_path = Arc::new(
    commit_diff
        .files
        .iter()
        .map(|file| StatusEntry {
            repo_path: file.path.clone(),
            status: FileStatus::Tracked(TrackedStatus {
                index_status: StatusCode::Modified,
                worktree_status: StatusCode::Unmodified,
            }),
        })
        .collect(),
);
```

Functionally this isn’t a big issue — it only blurs the distinction between “new” and “modified” — but supporting this properly would require changes to the `GitRepository` trait, which feels too large in scope for now.

**Commit breakdown:**

- c4132689644d0fbd6b2c89235576c8a692312b11: Extracts the logic for collecting entries. No functional changes — just cleanup to reduce noise in later diffs.
- 6bcfffd64dbfbfc1f259c81601fa766a568098eb: Caches file statuses when opening a `CommitView` for later use.
- 6b0b172a381c356d40de658b8372278c0a57b12f: Adds path-based navigation for `CommitView` in the Git panel.
- 490d50abc414b96c2df489c4dc4f6c4605ffa595: Disables irrelevant UI elements when viewing a commit.

### DIFF
--- a/crates/git_ui/src/commit_view.rs
+++ b/crates/git_ui/src/commit_view.rs
@@ -3,6 +3,7 @@ use buffer_diff::{BufferDiff, BufferDiffSnapshot};
 use editor::display_map::{BlockPlacement, BlockProperties, BlockStyle};
 use editor::{Editor, EditorEvent, ExcerptRange, MultiBuffer, multibuffer_context_lines};
 use git::repository::{CommitDetails, CommitDiff, RepoPath};
+use git::status::{FileStatus, StatusCode, TrackedStatus};
 use git::{GitHostingProviderRegistry, GitRemote, parse_git_remote_url};
 use gpui::{
     AnyElement, App, AppContext as _, Asset, AsyncApp, AsyncWindowContext, Context, Element,
@@ -14,6 +15,7 @@ use language::{
     Point, ReplicaId, Rope, TextBuffer,
 };
 use multi_buffer::PathKey;
+use project::git_store::StatusEntry;
 use project::{Project, WorktreeId, git_store::Repository};
 use std::{
     any::{Any, TypeId},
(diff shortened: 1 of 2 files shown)


tediou5:
@cla-bot check

tediou5:
I adjusted the commit to facilitate review, but I haven't tested it manually because of some problems with the PC environment I am currently on, just a simple comparison of the code before and after. When I go back tomorrow, I'll test it to make sure I don't lose anything in my changes

tediou5:
3cc07ee7-9ea6-42e7-af43-c1ff5dfffc4a

> CommitDiff doesn’t include FileStatus, so at the moment I’m defaulting everything to Modified

Here is always Modified.

cole-miller:
Thanks for opening a PR--however, I don't think we want to reuse the git panel in this way; it's a bit confusing to have to interpret its contents differently depending on whether a past commit is open in the center pane. With that in mind, I'm going to close this.

PR #44811, opened 2025-12-14
editor: Don't merge adjacent selections

### LINKED ISSUE #24748: Zed merges adjacent non-overlapping selections
### Summary

Zed eagerly merges adjacent selections, it should only merge overlapping selections.

In any file:
1. Place your cursor by clicking
2. alt-click three characters to the right
3. Press `shift-right` three times

Actual Behavior: One combined selection

Expected Behavior: Two adjacent selection

| VSCode | Zed |
| - | - |
| Image | Image |

https://github.com/user-attachments/assets/ef09dce2-63a2-401f-bafa-25b458304afc

Originally reported @pjv here:
- https://github.com/zed-industries/zed/issues/21881

Zed: v0.174.0 (Zed Nightly 553cc2cca5c17e755e617d499fef5bb52ba67b11) 
OS: macOS 15.3.0
Memory: 64 GiB
Architecture: aarch64

marcocondrache:
Closes #24748

Release Notes:

- Adjacent selections are not merged anymore

### DIFF
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -16202,7 +16202,7 @@ async fn test_toggle_comment(cx: &mut TestAppContext) {
     cx.assert_editor_state(indoc! {"
         fn a() {
             «b();
-            c();
+            ˇ»«c();
             ˇ» d();
         }
     "});
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -16214,8 +16214,8 @@ async fn test_toggle_comment(cx: &mut TestAppContext) {
     cx.assert_editor_state(indoc! {"
         fn a() {
             // «b();
-            // c();
-            ˇ»//  d();
+            ˇ»// «c();
+            ˇ» // d();
         }
     "});
 
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -7750,10 +7750,12 @@ fn test_select_line(cx: &mut TestAppContext) {
             ])
         });
         editor.select_line(&SelectLine, window, cx);
+        // Adjacent line selections should NOT merge (only overlapping ones do)
         assert_eq!(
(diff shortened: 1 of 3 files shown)


marcocondrache:
I'm having issues with the test `editor_tests::test_toggle_comment`. I'm struggling with whitespace handling, and the assertion keeps failing. It's driving me crazy, and I think I need the team's support on this one

image

Veykril:
The test now fails because it seems we have differing behavior here.
```diff
    cx.assert_editor_state(indoc! {"
        fn a() {
            // «b();
-            // c();
-            ˇ»//  d();
+            ˇ»// «c();
+            ˇ» // d();
        }
    "});
```
Note how we now end up with two selections after invoking `ToggleComments` where as before we only had one. Which makes sense, looking at the editor state before that assertion, we have two selections touching each other which now no longer merge. And the whitespace change also comes from that, because we remain with two selections here and thus consider different columns for each

marcocondrache:
@Veykril Thanks so much for this - that helped it click for me. I was counting the markers as columns, but they disappear in the final marked text. Rookie mistake 😄

Veykril (approved):
Thanks!

JosephTLyons:
@marcocondrache, I'll make an issue for this later, but figured I'd bring it up here first. I'm assuming a separate issue, but once you start typing, things gets merged.

https://github.com/user-attachments/assets/6ab3cbef-ded6-4906-9d78-9b3e967c9c27

marcocondrache:
@JosephTLyons wow, thank you for discovering this. Could you assign the issue to me? I will provide a fix immediately

JosephTLyons:
> @JosephTLyons wow, thank you for discovering this. Could you assign the issue to me? I will provide a fix immediately

Thank you so much! Here's the issue. I'm weirdly not able to assign it currently, I think we might have some permission issues going on since we migrated over to an GH organization. Maybe if you comment on the issue, I can assign.

- https://github.com/zed-industries/zed/issues/45046

Issue #16965, opened 2024-08-27
Support for non UTF-8 text encodings


notpeter:
### Check for existing issues

- [X] Completed

### Describe the feature

Zed currently only supports UTF-8 text.  This is an enhancement request for it to support other text encodings.

- [ ] 8859-1 (Latin-1) [latin1.txt](https://github.com/user-attachments/files/16767064/latin1.txt)
- [ ] CP865 (DOS Nordic) [cp865.txt](https://github.com/user-attachments/files/16767023/cp865.txt)
- [ ] KOI8-R (Russian) [koi8_r.txt](https://github.com/user-attachments/files/16767039/koi8_r.txt)
- [ ] UTF-16 (16bit Unicode) [utf16.txt](https://github.com/user-attachments/files/16767053/utf16.txt)
- [ ] Shift-JIS (Japanese) [main_shift-jis.cpp.txt](https://github.com/user-attachments/files/16767118/main_shift-jis.cpp.txt)

There are separate issues for supporting non-text binary data and CR/LF line-endings:
- https://github.com/zed-industries/zed/issues/5250
- https://github.com/zed-industries/zed/issues/5294

If there are additional specific encodings you would like to see, please comment below with a sample file and I will add it to the list above.

Please 👍 upvote this issue if you would like to see this feature prioritized. (+1 comments will be removed).

FrittenKeeZ:
I think Windows 1252 (or CP1252) would be a useful addition.
[cp1252.txt](https://github.com/user-attachments/files/16778117/cp1252.txt)

polytect:
Yes. UTF-16 which i need, it doesn't open. 

Over all first impressions of Zed, it is amazing simple and fast.
What can we do, to make additional encodings to see a daylight..

niki-sp:
Windows-1251 and really all 125x

varyform:
The way Zed silently ignores files it can't open is a bad thing.
It should be considered as a critical bug.
I had to refresh my memory with 20 y/o Delphi project today.
Was searching for ASCII variable. Zed – 3 files, 5 matches.
Sublime – 1191 matches across 10 files.

Why not to warn: "ignored files we can't read"?

Also:
<img width="505" alt="image" src="https://github.com/user-attachments/assets/03497438-3c31-4b56-ae7c-fa7bdbea44e5">
There should be a way to copy this text (or error ID)

– "Stream did not contain valid UTF-8"?
Shouldn't that be "stream contains INVALID UTF-8?"

[...]

PR #44819, opened 2025-12-14
Support opening and saving files with legacy encodings

tomopumipumi:
## Summary

Addresses #16965

This PR adds support for **opening and saving** files with legacy encodings (non-UTF-8).
Previously, Zed failed to open files encoded in Shift-JIS, EUC-JP, Big5, etc., displaying a "Could not open file" error screen. This PR implements automatic encoding detection upon opening and ensures the original encoding is preserved when saving.

## Implementation Details

1.  **Worktree (Loading)**:
    * Updated `load_file` to use `chardetng` for automatic encoding detection.
    * Files are decoded to UTF-8 internal strings for editing, while preserving the detected `Encoding` metadata.
2.  **Language / Buffer**:
    * Added an `encoding` field to the `Buffer` struct to store the detected encoding.
3.  **Worktree (Saving)**:
    * Updated `write_file` to accept the stored encoding.
    * **Performance Optimization**:
        * **UTF-8 Path**: Uses the existing optimized `fs.save` (streaming chunks directly from Rope), ensuring no performance regression for the vast majority of files.
        * **Legacy Encoding Path**: Implemented a fallback that converts the Rope to a contiguous `String/Bytes` in memory, re-encodes it to the target format (e.g., Shift-JIS), and writes it to disk.
        * *Note*: This fallback involves memory allocation, but it is necessary to support legacy encodings without refactoring the `fs` crate's streaming interfaces.

## Changes

- `crates/worktree`:
    - Add dependencies: `encoding_rs`, `chardetng`.
    - Update `load_file` to detect encoding and decode content.
    - Update `write_file` to handle re-encoding on save.
- `crates/language`: Add `encoding` field and accessors to `Buffer`.
- `crates/project`: Pass encoding information between Worktree and Buffer.
- `crates/vim`: Update `:w` command to use the new `write_file` signature.

## Verification

I validated this manually using a Rust script to generate test files with various encodings.

**Results:**

* ✅ **Success (Opened & Saved correctly):**
    * **Japanese:** `Shift-JIS` (CP932), `EUC-JP`, `ISO-2022-JP`
    * **Chinese:** `Big5` (Traditional), `GBK/GB2312` (Simplified)
    * **Western/Unicode:** `Windows-1252` (CP1252), `UTF-16LE`, `UTF-16BE`
* ⚠️ **limitations (Detection accuracy):**
    * Some specific encodings like `KOI8-R` or generic `Latin1` (ISO-8859-1) may partially display replacement characters (`?`) depending on the file content length. This is a known limitation of the heuristic detection library (`chardetng`) rather than the saving logic.

Release Notes:

- Added support for opening and saving files with legacy encodings (Shift-JIS, Big5, etc.)


tomopumipumi:
@cla-bot check

SomeoneToIgnore:
We actually used to have quite a promising PR https://github.com/zed-industries/zed/pull/36497 for a similar thing, but will leave for @ConradIrwin to check on both and decide.

tomopumipumi:
Thanks for the information. I'm happy to wait for Conrad to take a look.

ConradIrwin:
@tomopumipumi Amazing, thank you for this! I love that it's much simpler than the previous approaches we were trying. 

In testing out this PR against https://github.com/zed-industries/encodings-tests, I noticed that if you open `ISO_2022-JP` or `UTF16-LE` and then save the file without changing, it writes different bytes to disk.

I want to be sure that Zed is (to the greatest extent possible) not going to silently corrupt files, so I think we should make sure that this doesn't happen.

I'd also like (but probably a separate PR) a status bar indicator that shows when the current file is not UTF-8 to avoid surprises (and then there's the even more scope creep of being able to select and change encodings; but they should definitely be follow ups).

tomopumipumi:
​@ConradIrwin Thank you for the positive feedback! I'm glad you liked the simpler approach.
​Regarding the byte mismatch issue with ISO-2022-JP and UTF-16LE: I suspect this relates to BOM handling or encoding normalization. I will investigate the behavior using the test cases you provided and ensure round-trip safety.
​I also agree that the status bar indicator and UI selector should be separate follow-up PRs to keep this one focused.

tomopumipumi:
@ConradIrwin

I have completed the investigation into the byte mismatch issues and updated the implementation accordingly. 
Based on the test files you provided, I’ve refined the logic and verified the round-trip behavior.

Here is a summary of the changes and my findings:

## 1. Implementation Updates

I have refined the logic to handle BOMs more strictly:

- Immediate Detection via BOM:
If a BOM is present, the encoding is determined immediately without relying on other heuristics.

- Tracking BOM Presence (has_bom flag):
I added a has_bom boolean flag to the Buffer struct. This tracks whether the original file had a BOM so the system can decide whether to write it back upon saving.

- Saving Behavior:
When saving to a file, the system now checks this flag and the encoding to re-insert the appropriate BOM bytes at the beginning of the file.

## 2. Test Results

I verified the implementation using the test files you provided. I confirmed that for all files except UTF-16LE, the saved files result in a byte-for-byte match with the originals. 
The UTF-16 series files in the test set do not contain a BOM, making them extremely difficult to detect reliably.

Regarding the ISO-2022 series, the bytes happened to match perfectly this time, likely because the test files were composed of standard escape sequences. 
If the byte sequences were to change, it would follow the logic described in Section 3 below.

You can review the reproduction code using the test files in my repository here: [zed-encoding-verification](https://github.com/tomopumipumi/zed-encoding-verification) 
I have also performed a manual byte-level comparison.

## 3. Observations on Encoding Behavior

I believe the following behaviors are acceptable for these reasons:

- ISO-2022 Series (Byte changes):
When saving ISO-2022 files, the resulting byte sequence may differ from the original. Since we convert to and from Unicode via a library, I view this not as "data destruction" but as "normalization" of the encoding. This behavior is standard for modern editors that support legacy encodings. While the standard VS Code does not support these natively, extensions that do (e.g., ConvertToISO-2022-JP) generally perform similar Unicode conversion processes.

- BOM-less UTF-16:
Detecting "BOM-less" UTF-16, especially those containing non-ASCII characters, is extremely difficult and often results in false positives for UTF-8. Implementing a check before the main UTF-8 processing would likely require statistical logic, which could impact UTF-8 opening performance. Rather than forcing unreliable heuristics, I believe it is more beneficial to address this in the future by adding a "Reopen with Encoding" feature to the UI.

- ISO-8859 / Windows Series:
Similar to VS Code, some legacy encodings in these series are auto-detected as UTF-8. I consider this acceptable as it aligns with the standard behavior of modern editors, and the byte sequences remain unchanged upon saving.

Best regards,

crates/worktree/src/worktree.rs
```diff
@@ -1464,7 +1481,49 @@ impl LocalWorktree {
...
+                };
+
+                let (cow, _, _) = encoding.encode(&normalized_text);
+                let bytes = if !bom_bytes.is_empty() {
+                    let mut bytes = bom_bytes;
+                    bytes.extend_from_slice(&cow);
+                    bytes
+                } else {
+                    cow.into_owned()
+                };
```
CrazyboyQCD, line 1530: We could avoid one allocation here.
```suggestion
                let bytes = if !bom_bytes.is_empty() {
                    let mut bytes = bom_bytes;
                    bytes.extend_from_slice(&cow);
                    bytes.into()
                } else {
                    cow
                };
```

CrazyboyQCD: Or even more, we could avoid allocation by writing `bom` and `cow` seperately, so that bom could be `&[u8]`, but I am unsure if it is worth writing twice to avoid allocation.

tomopumipumi: @CrazyboyQCD Thanks for the review! Avoiding the allocation in the else branch is a great improvement. I've applied your suggestion.
Regarding writing the BOM and content separately, I think I'll stick to the single buffer approach for now to keep the implementation simple.

ConradIrwin:
Great, thank you! I'm happy to merge this as is.

For UTF-16 detection it seems like we could do a relatively cheap heuristic from looking at the first ~8 bytes of the file and if 4 of them are null then try the whole file as UTF-16 falling back to UTF-8 (That would fix the UTF16-LE one too, as I think the problem is we insert a trailing newline after the trailing null 🤦). PR's welcome, as they say :D.

Are you excited about sending a PR to add a status bar indicator to show the current encoding if it's not utf-8? If not happy to pair with you on building out next steps here, and thanks again for this!

tomopumipumi:
@ConradIrwin
Thank you for the guidance and for merging this! It was a great learning experience.
I plan to open another PR soon regarding the UI or UTF-16 improvements.

CrazyboyQCD:
@tomopumipumi 
I forgot that `UTF-16LE` and `UTF-16BE` in `encoding_rs` are decode-only encodings, current solution is [problematic](https://github.com/hsivonen/encoding_rs/issues/104), they should be done separately like [this](https://github.com/EuclidDivisionLemma/zed/blob/4330e5ff1439be5c0b2a88203b456379526abf84/crates/encodings/src/lib.rs#L98-L113) from https://github.com/zed-industries/zed/pull/36497.

tomopumipumi:
​@CrazyboyQCD Thanks for the explanation! I understand the issue now.
​I plan to open a separate PR to fix this UTF-16 saving logic and also implement the detection heuristic Conrad suggested. I think it's better to address these logic improvements together before working on the UI.

PR #44823, opened 2025-12-14
remote: Add ssh timeout setting

### LINKED ISSUE #21527: Support customizable SSH connection timeout
### Summary

Hard coded 10-second ssh timeout is not suitable for all environments.

### Description

Currently there is a hardcoded 10-seconds connection timeout for SSH, see here: https://github.com/zed-industries/zed/blob/72d57004fca71a33eabc0d65d10700784756fc05/crates/remote/src/ssh_session.rs#L1563

I run SSH over various tunnelling and proxying mechanisms which sometimes takes more than 10 seconds to establish, so despite the connection being stable and usable I'll sometimes have a hard time just to get connected. Once I am connected the remote session works well, but getting that initial connection because of this hardcoded timeout can be difficult.

This timeout should either be configurable or greatly increased.

marcocondrache:
Closes #21527

Release Notes:

- Added a setting to specify the ssh connection timeout

### DIFF
--- a/crates/remote/src/transport/ssh.rs
+++ b/crates/remote/src/transport/ssh.rs
@@ -1260,6 +1260,10 @@ impl SshConnectionOptions {
     pub fn additional_args(&self) -> Vec<String> {
         let mut args = self.additional_args_for_scp();
 
+        if let Some(timeout) = self.connection_timeout {
+            args.extend(["-o".to_string(), format!("ConnectTimeout={}", timeout)]);
+        }
+
         if let Some(forwards) = &self.port_forwards {
             args.extend(forwards.iter().map(|pf| {
                 let local_host = match &pf.local_host {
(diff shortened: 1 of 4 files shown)


crates/remote/src/transport/ssh.rs
```diff
@@ -1252,6 +1260,10 @@ impl SshConnectionOptions {
     pub fn additional_args(&self) -> Vec<String> {
         let mut args = self.additional_args_for_scp();
 
+        if let Some(timeout) = self.connection_timeout {
+            args.extend(["-o".to_string(), format!("ConnectTimeout={}", timeout)]);
+        }
```
Veykril, line 1265: I am not sure we want this here, this will affect all the commands we run after

marcocondrache: @Veykril I’m a bit unsure what the best approach is here. From what I can tell, the only hardcoded timeout left is in the binary download step, and it applies to creating the connection rather than the download itself. That’s what led me to add it there, mainly for consistency.

I’m happy to adjust this though - curious what you think would be the cleanest way

Veykril: Okay I double checked myself, I was wrong here. This is not used by what I thought, so this looks good, thanks!

Veykril (approved):
Thanks!

PR #44826, opened 2025-12-14
Stop spawning ACP/MCP servers with interactive shells

xipeng-jin:
### Summary:
- Ensure the external agents with ACP servers start via non-interactive shells to prevent shell startup noise from corrupting JSON-RPC.
- Apply the same tweak to MCP stdio transports so remote context servers aren’t affected by prompts or greetings.

### Description:
Switch both ACP and MCP stdio launch paths to call `ShellBuilder::non_interactive()` before building the command. This removes `-i` on POSIX shells, suppressing prompt/title sequences that previously prefixed the first JSON line and caused `serde_json` parse failures. No functional regressions are expected: both code paths only need a shell for Windows/npm script compatibility, not for interactivity.

Release Notes:
- Fixed external agents that hung on “Loading…” when shell startup output broke JSON-RPC initialization.

### DIFF
--- a/crates/agent_servers/src/acp.rs
+++ b/crates/agent_servers/src/acp.rs
@@ -89,7 +89,7 @@ impl AcpConnection {
         cx: &mut AsyncApp,
     ) -> Result<Self> {
         let shell = cx.update(|cx| TerminalSettings::get(None, cx).shell.clone())?;
-        let builder = ShellBuilder::new(&shell, cfg!(windows));
+        let builder = ShellBuilder::new(&shell, cfg!(windows)).non_interactive();
         let mut child =
             builder.build_command(Some(command.path.display().to_string()), &command.args);
         child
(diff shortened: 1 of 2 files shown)


xipeng-jin:
I am having an issue with the nightly build with Zed. To be more specific, when I launch an external ACP agent (e.g. Codex CLI, or Opencode through extension) from the Agent panel ("+" → “New Thread…” → “Codex CLI”), the thread tab would stay stuck in “Loading…” with the Codex icon blinking indefinitely.

In the dev ACP logs we only saw the initial `initialize` request from Zed, followed by repeated parse errors like:
"failed to parse incoming message: expected value at line 1 column 1", even though the “Raw” log line appeared to be valid JSON. The real issue was that the agent’s stdout stream was being prefixed with invisible/non-JSON bytes produced by an interactive shell startup (e.g. `fish` prompt/greeting/terminal escape sequences), so the first byte of the “JSON line” wasn’t actually `{`. Switching terminal.shell to `fish --no-config` made it work, confirming the shell startup output was corrupting the newline-delimited JSON-RPC channel and preventing ACP initialization from completing.

Veykril (approved):
Thanks!

PR #44827, opened 2025-12-14
editor: Smooth scrolling

### LINKED ISSUE #4355: Smooth scrolling
### Check for existing issues

- [X] Completed

### Describe the feature

With Zed performance, I see it as a very potential candidate for me to replace neovim with, I'm currently using Neovide, I would love a feature that is on Neovide with --multigrid enabled, and that is smooth scrolling, it would make Zed even cooler than what it already is.

It could probably be done one the same batch as https://github.com/zed-industries/community/issues/994

_No response_

marcocondrache:
Closes #4355

Before:

https://github.com/user-attachments/assets/1f3e6f36-4382-4b37-920b-8d74fc0501c3

After:

https://github.com/user-attachments/assets/17a0e87f-2f49-40a3-8c6e-2b448fe03240

Ref: https://pavelfatin.com/scrolling-with-pleasure

Release Notes:

- Editor now supports smooth scrolling


niekdomi:
Thanks for this PR, feels smooth. Are there any plans to integrate this into vim-mode as well?

(e.g., `ctr-y`, `ctrl-e`, `ctrl-b`, `ctrl-f`, `ctrl-d`, `ctrl-u`, `zz`, `zt`, `zb`)

or is this not part of this PR?

marcocondrache:
@niekdomi yep, I still need to make the changes. I'm currently perfecting the scroll animation

crates/editor/src/element.rs
```diff
@@ -9067,6 +9074,26 @@ impl Element for EditorElement {
...
+                            let snapshot =
+                                editor.display_map.update(cx, |map, cx| map.snapshot(cx));
+
+                            match update.phase {
+                                ScrollAnimationPhase::Intermediate => {
+                                    editor.scroll_manager.set_scroll_position_visual(
+                                        update.position, &snapshot, cx,
+                                    );
+
+                                    window.request_animation_frame();
```
marcocondrache: I agree this isn’t the ideal approach for animations, especially in the editor, since it causes the whole entity to re-render. That said, in practice it doesn’t seem to make a noticeable difference right now.

The current scrolling implementation in the editor already calls `cx.notify()` on every scroll event, so the animation isn’t doing anything fundamentally different in that regard. In fact, it may even be slightly more efficient, since we avoid updating the db and emitting events until the animation has finished, which is the opposite of how the current scroll handling works.

I’m very open to improving this if you think there’s a better direction, but wanted to share the reasoning behind the current approach

ttytm (commented):
Amazing effort!

Hope giving input already doesn't bother you.

Taking it through manual testing, I found that it doesn’t work with the optional preceding `[count]`.

For example, my personal default for <kbd>C+e</kbd> / <kbd>Ctrl+y</kbd> is to scroll 5 lines. Currently it is reverted to 1 line when smooth scrolling is enabled. 
No matter which scrolling motion is pressed `[count]` is ignored.

I wouldn’t want this to be a blocker, though. So a fix could also be part of a separate PR.

marcocondrache:
@ttytm, thank you for the feedback. I'm primarily focusing on improving the animation, especially for trackpads. I'm waiting for some input from the team before I make significant changes and dive deep into the vim crate

crates/editor/src/element.rs
MrSubidubi: As opposed to cursor animations where some of this still holds, I do not see any issues with this for scrolling - e.g., if you are using a trackpad, this basically already happens on Mac prior to this PR, so there is no difference. So totally fine as is and not needed to change anything

marcocondrache: @MrSubidubi thank you!

assets/settings/default.json
```diff
@@ -634,6 +634,12 @@
   // to both the horizontal and vertical delta values while scrolling. Fast scrolling
   // happens when a user holds the alt or option key while scrolling.
   "fast_scroll_sensitivity": 4.0,
+  "smooth_scroll": {
+    // Whether to animate scrolling with a smooth easing effect.
+    "enabled": false,
+    // Duration of the animation
+    "duration": 0.125,
```
marcocondrache: It could also make sense to split the duration into two values: one for mouse wheel scrolling and one for keybinding scrolling. I think most people would want the mouse to scroll faster, and vim/actions scrolling to stay a bit slower and more controlled.

crates/editor/src/element.rs
```diff
@@ -7813,22 +7814,39 @@ impl EditorElement {
...
                         }
 
-                        if scroll_position != current_scroll_position {
-                            editor.scroll(scroll_position, axis, window, cx);
+                        if scroll_position != base_scroll_position {
+                            if is_precise {
+                                editor.scroll(scroll_position, axis, window, cx);
+                            } else {
+                                editor.scroll_animated(scroll_position, axis, window, cx);
+                            }
```
MrSubidubi: Seeing this and some other places, I can't help but wonder if we should not only have one `editor.scroll` which takes a parameter like 

```rust
enum ScrollMode {
    Instant, 
    /// Smooth scrolls if this is configured by the user
    RequestAnimation
}
```

(naming very much [subject to change](https://media.licdn.com/dms/image/v2/C5612AQEM6pQlHdIipg/article-inline_image-shrink_400_744/article-inline_image-shrink_400_744/0/1538359939586?e=1769040000&v=beta&t=8XqKpwhMSv275Ut9b-tqm9eOyf5FZ3YE1ZpB-XPAmAk), can't think of anything better right now).

The primary reason for this is that I think that for future maintainability, I think it is better that we have something where each time `editor.scroll` is called, the programmer has to think about how we want to scroll here and I'd argue it makes it easier to spot if that assumption was wrong. Happy to discuss this, just a thought I had when looking at this. There are other spots where I'd call this more urgent, but came to mind first here.

crates/vim/src/normal/scroll.rs
```diff
@@ -111,7 +110,23 @@ fn scroll_editor(
...
+            (
+                DisplayRow(a.target_position.y as u32),
+                a.target_position.x as u32,
+            )
+        })
+        .unwrap_or_else(|| {
+            let snapshot = editor.display_map.update(cx, |map, cx| map.snapshot(cx));
+            let position = editor.scroll_manager.anchor().scroll_position(&snapshot);
+            (DisplayRow(position.y as u32), position.x as u32)
+        });
```
MrSubidubi: While I know that the Vim crate reaches pretty deeply within `Editor` (and probably has a lot of places where this is worse), I feel like this and the code below should live in editor - downstream crates should not need to handle/be concerned with smooth scrolling, that seems rather error-prone to me.

crates/editor/src/scroll.rs
```diff
@@ -587,6 +727,8 @@ impl Editor {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) -> WasScrolled {
+        self.scroll_manager.cancel_animation();
```
MrSubidubi: Do we want to have this and the one below always? Or should that maybe also be left to the caller?

MrSubidubi (commented):
Sorry for the delay here, while I both want to have this myself and also see the demand, I am sadly very limited on time currently and also _really_ wanna make sure we don't have anything breaking due to this.

Left some food for thought, nothing too severe, but some stuff I'd definitely like to discuss. The overall implementation does look really solid (I will probably have some more comments, but not right now). 

Will do my best to push this more forward, that said, please forgive me if reviewing takes some time. I'll try to get back in a timely manner, but too much stuff currently on my hands. I'll try my best, apologies anyway.

crates/editor/src/element.rs
marcocondrache: That’s a solid point. I actually thought of this myself when I started the PR, but I decided to go with the fastest and simplest option. What do you think about calling the enum `ScrollBehavior`? The word “mode” makes me think of something related to state. In this case, the scrolling will behave in a different way if the smooth scrolling is enabled and if we requested the animation.

crates/editor/src/scroll.rs
marcocondrache: Both methods forcefully change the editor’s scroll position. If an animation is already in progress, it would then continue from an invalid state, which would cause a visible jump. In that situation we can’t advance the animation in a correct or consistent way, so I think those calls should cancel it.

crates/vim/src/normal/scroll.rs
marcocondrache: Yes super right. Should be done

crates/editor/src/scroll.rs
```diff
@@ -191,6 +191,97 @@ impl ActiveScrollbarState {
...
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ScrollAnimationPhase {
+    Intermediate,
+    Final,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct ScrollAnimation {
```
MrSubidubi: I think the animation phase should be `pub(crate)` at best, the state itself should perhaps even only be visible within this file even?

crates/editor/src/scroll.rs
```diff
@@ -378,54 +473,30 @@ impl ScrollManager {
...
-                    let max_row = map.max_point().row().as_f64();
-                    scroll_top
-                        .min(max_row - height_in_lines + 1. + self.vertical_scroll_margin)
-                        .max(0.)
-                } else {
-                    scroll_top
-                }
-            }
-        };
+        let (anchor, top_row) = self.calculate_scroll_anchor(scroll_position, map, cx);
```
MrSubidubi: Let's revert the change here since `calculate_scroll_anchor` is only called once now.

assets/settings/default.json
```diff
@@ -650,6 +650,12 @@
...
```
MrSubidubi: I am not entirely sure we should make the duration configurable at all - I am sold on having the behavior change based on the input type used, and feel like making the duration configurable might be better to do at a later step if at all. This feels like a case where good defaults might just be more worth than to make this configurable in the first place. The fact that VSCode does not have a setting for this makes me feel like we might not need it as well or at least not in the first place.

crates/editor/src/editor.rs
```diff
@@ -27681,6 +27683,14 @@ impl EditorSnapshot {
         self.scroll_anchor.scroll_position(&self.display_snapshot)
     }
 
+    pub fn scroll_target_or_position(&self) -> gpui::Point<ScrollOffset> {
```
MrSubidubi: Can we not just have `scroll_position`, do we really need to distinguish these here?

crates/editor/src/editor.rs
```diff
@@ -1407,6 +1407,7 @@ pub struct EditorSnapshot {
     pub placeholder_display_snapshot: Option<DisplaySnapshot>,
     is_focused: bool,
     scroll_anchor: SharedScrollAnchor,
+    pub scroll_animation: Option<ScrollAnimation>,
```
MrSubidubi: Is there a good reason for this to be pub?

crates/editor/src/scroll/actions.rs
```diff
@@ -29,11 +33,24 @@ impl Editor {
...
+        let duration = match behavior {
+            ScrollBehavior::RequestAnimation if smooth_scroll.enabled => {
+                Duration::from_secs_f32(smooth_scroll.duration)
+            }
+            _ => Duration::ZERO,
+        };
+
+        self.scroll_manager
+            .scroll_to(current_position, scroll_position, duration);
+        cx.notify();
```
MrSubidubi: I feel like we should separate concerns better here - this now does a mix of delegating things to the scroll manager whilst keeping others around here in the actions. E.g. the duration should and could probably be resolved in the scroll manager - similarly, we could just store there whether smooth scrolling is enabled and ensure this stays up to date at https://github.com/zed-industries/zed/blob/8762b7f503fda985f481ccd301960cf01adbecb8/crates/editor/src/editor.rs#L24381.

Core idea remains the same - I want to have this as contained as possible to the scroll manager so that we are not at risk of regressing things here.

crates/editor/src/scroll.rs
```diff
@@ -191,6 +191,97 @@ impl ActiveScrollbarState {
...
+
+        (elapsed / duration).min(1.0)
+    }
+
+    pub fn advance(&mut self) {
+        if self.is_finished() {
+            self.phase = ScrollAnimationPhase::Final;
+            self.position = self.target_position;
+        } else {
+            let progress = self.progress();
```
MrSubidubi: We calculate the progress twice here in rapid succession - Perhaps we can have `progress` return a newtype `ScrollAnimationProgress(f32)` that then has the method `is_finished` on that so we avoid doing this twice?

crates/editor/src/scroll.rs
```diff
@@ -191,6 +191,97 @@ impl ActiveScrollbarState {
...
+    pub fn instant(target: gpui::Point<ScrollOffset>) -> Self {
+        Self {
+            phase: ScrollAnimationPhase::Final,
+            duration: Duration::ZERO,
+            position: target,
+            start_position: target,
+            target_position: target,
+            start_time: Instant::now(),
+        }
+    }
```
MrSubidubi, line 267: Looking at this and the surrounding methods, I can't help but wonder whether we should manage the state here using an enum as opposed to the struct - that way, we can get rid of plenty of the duplication here and can avoid some invalid states a bit better. 
This might be more verbose in parts, but might also help more with what the current state is? Since `ScrollAnimationPhase::Final` basically implies `position == target_position` for example. Would like your view on this obviously

crates/vim/src/normal/scroll.rs
```diff
@@ -112,10 +112,12 @@ fn scroll_editor(
...
-    let display_snapshot = editor.display_map.update(cx, |map, cx| map.snapshot(cx));
-    let old_top = editor
-        .scroll_manager
-        .scroll_top_display_point(&display_snapshot, cx);
+
+    // Capture old scroll position before scrolling
+    let snapshot = editor.snapshot(window, cx);
+    let scroll_position = snapshot.scroll_target_or_position();
+    let old_top_row = DisplayRow(scroll_position.y as u32);
+    let old_top_column = scroll_position.x as u32;
```
MrSubidubi: This is much better already, thank you.

We should make `scroll_manager` private as part of this then though - if these methods can no longer be used freely due to these changes, we should ensure that no one can actually use them (and again (sorry), ensure that visibility of methods matches what we enforce here). I know our policies on comments, but for this case, might even be worth to have a comment within editor to ensure peole do not easily violate that.

crates/editor/src/scroll.rs
```diff
@@ -191,6 +191,97 @@ impl ActiveScrollbarState {
...
+        self.advance();
+        self.start_position = self.position;
+        self.start_time = Instant::now();
+        self.target_position = target;
+        self.phase = if self.target_position == self.start_position {
+            ScrollAnimationPhase::Final
+        } else {
+            ScrollAnimationPhase::Intermediate
+        }
+    }
```
MrSubidubi, line 415: I think one issue we might face here is when we invoke this method, we reset the current acceleration, which I am not sure is what we want - you lose the current scroll momentum as we start the easing method from the very beginning again, which means we might be in the mid of easing - and then are at the very start again. Instead, I think we should consider the current acceleration/speed at which we are animating and try to find a suitable progress at which we can continue the animation/perhaps change the duration slightly. So this should not be a hard restart I believe but instead more calculate a new animation based on the current state.

crates/editor/src/element.rs
```diff
@@ -7673,21 +7674,39 @@ impl EditorElement {
...
+                        let base_scroll_position = if is_precise {
+                            editor.scroll_manager.cancel_animation();
+                            current_scroll_position
+                        } else {
+                            position_map
+                                .snapshot
+                                .scroll_animation
+                                .map(|animation| animation.target_position)
+                                .unwrap_or(current_scroll_position)
+                        };
```
MrSubidubi: Could you briefly elaborate on the logic here?

MrSubidubi (changes requested):
Sorry for the delay, as you know, was out sick and still a bit slow to pick up on some things.

Next round here, some nits although I'd argue we better are safe than sorry here, given that this reaches really deep. There is also the following issue, which we obviously need to figure out, where trackpad scrolling on my Mac is not acceptable with the native acceleration from the OS: 

https://github.com/user-attachments/assets/a9fc8e02-4408-40a6-991e-8e26d36957fe

Which I think might very well be related to one of my comments. 

Nonetheless, we are getting closer though and I'd say after this round we are getting to the topic of tests, so feel free to start thinking about them (not necessarily add them just yet, but you are obviously free to already start working on these). Arguably, I think we want to test stuff amongst the lines of
- expected vs. calculated scroll positions on a per-frame basis.
- change in direction during animation
- instant scroll request during scroll animation
- animation scroll request during animation scroll request
- animated scroll after animated scroll to ensure state is in sync (can perhaps be part of the first test
- settings update test during animation - perhaps we might not need this, but what I'd like to test here is that we do not end up in a state we cannot escape

and perhaps some other nice cases you can think about. As said, food for thought, I think it might be easier to tacke these after this batch here.

crates/editor/src/scroll.rs
marcocondrache: The animation must be at least `pub(crate)` since it's used in the `EditorSnapshot`. Should be fixed

crates/editor/src/scroll.rs
marcocondrache: Super right, should be done

crates/editor/src/editor.rs
marcocondrache: Should be done

crates/editor/src/scroll.rs
marcocondrache: Nice idea, I have created `ScrollAnimationProgress`

crates/editor/src/scroll.rs
marcocondrache: That’s a really nice idea, not sure why it didn’t come to my mind earlier. Looks a lot better now

crates/editor/src/scroll/actions.rs
marcocondrache: Right, should be done

assets/settings/default.json
marcocondrache: That makes sense. The main reason I added it initially was that scrolling felt much slower when smooth scrolling was enabled. Later I discovered that increasing the sensitivity fixes that, so it definitely makes sense to remove it.

Should be done 😊

crates/vim/src/normal/scroll.rs
marcocondrache: That's right. I checked where `scroll_manager` is accessed, and it looks like the `agent_ui` crate is still accessing it directly. Should we address that in this PR?

crates/editor/src/editor.rs
marcocondrache: We don't need to distinguish them anymore, should be done

crates/editor/src/element.rs
Zakleby: is_precise means trackpad bud which means u already have more accurate scrolling which naturally means u dont need to smooth scroll, and then the else statement means ur not on a trackpad it allows you to smooth scroll!!!!!

marcocondrache: @MrSubidubi I've removed this logic it's not necessary anymore. It was an attempt to understand the scrolling slowness difference between mouse and trackpad.

crates/editor/src/scroll.rs
marcocondrache: That’s also why increasing the sensitivity makes such a big difference. I’ve added an initial implementation that computes the velocity of the current animation using the derivative of the easing function, and then adjusts the duration of the next animation accordingly. The approach is inspired by Chromium’s smooth scrolling implementation ([ref 1](https://docs.google.com/document/d/1JQ6jLy-r7vw_I9s3rtWAIK137fMpF_OFVtfVsgOaoTA/edit?tab=t.0), [ref 2](https://source.chromium.org/chromium/chromium/src/+/main:cc/animation/scroll_offset_animation_curve.cc)).

The results are already much better than the previous approach, but I think with some additional tuning and research we can improve it even further.

seanstrom:
@marcocondrache do you have some time to rebase this branch and resolve the conflicts?

crates/vim/src/normal/scroll.rs
MrSubidubi: I think to save you on conflicts, I'd be best if we could move this out into another PR. I'll get that one in then ASAP

MrSubidubi (changes requested):
I agree this is at a good point to resolve conflicts - the PR is in very good shape now and I'd love to see this progress further. Currently, at a high level, see only three things left:
- the comment below
- tests as mentioned in the prior round
- how we want to handle trackpad scrolls - if we can, I am thinking about whether it would make sense to distinguish between already smoothed scrolls and non-smoothed scrolls on the platform level, or do you have a better idea for that? Would love your opinion on this

marcocondrache:
@MrSubidubi @seanstrom I've resolved the conflicts and merged `main`.

For everyone following the status of this PR: I’ve recently been pulled into leading a project at my current job, so I have very little, if any, time to dedicate to open source right now. Since this PR is nearly ready and it’s something the community really values, I’m aiming to get this closed out soon.

Apologies for this becoming a bit stale and for the lack of progress recently. I appreciate everyone’s patience

aizigao:
I have to say that the Zed team is now very slow to merge contributions to open source, and no one can dedicate more than half a year to a pr. you u so good 😂

marcocondrache:
@MrSubidubi 

> how we want to handle trackpad scrolls - if we can, I am thinking about whether it would make sense to distinguish between already smoothed scrolls and non-smoothed scrolls on the platform level, or do you have a better idea for that? Would love your opinion on this

I don’t think we can reliably detect at the platform level whether the input has already been smoothed.

The main issue is that different mice and trackpads can report different values depending on their hardware and software capabilities. On macOS, for example, we can detect whether a pixel-precise scrolling device is being used, which usually indicates a trackpad. That was one of the first things I implemented. However, I later discovered that my Logitech mouse also reports pixel-precise scrolling, even when smooth scrolling is disabled in the Logitech app.

So while I’m not sure how this behaves on other platforms, at least on macOS we don’t seem to have a reliable way to distinguish already-smoothed scrolling from non-smoothed scrolling. The smoothing may come from either the hardware or software layer, and from what I can tell, that makes it impossible to consistently detect or track

MrSubidubi:
Appreciate the context! Then let's just scrap that idea.

Aquaticat:
@MrSubidubi Please reconsider. I'm on Linux desktop, and this is my blocking issue for adopting Zed. I find it hard to track where I am whenever animated smooth scrolling isn't enabled.

An imperfect solution is still a solution. JetBrains IDEs implement smooth scrolling with an opt-in toggle and I've been using WebStorm fine.

Edit: Sorry, I misread the context as in dropping smooth scrolling altogether. I retract that. Kept here for record.

palmerovicdev:
Any update on this?

xab3r:
@marcocondrache 
I appreciate you working on this. I'm used to smooth scrolling in Sublime Text, and missing it so much in Zed. I gave it a try and noticed that it's heavily flickering when scrolling. This is a keybinding I use, with the vim mode enabled.

```json
      "ctrl-i": [
        "action::Sequence",
        ["vim::LineUp", "vim::LineUp", "vim::LineUp", "vim::LineUp"]
      ],
```

marcocondrache:
@xab3r Did you run this in release mode? In debug mode, that behavior is expected.

I also recently made some fixes to the vim integration to get a few tests passing, so it’s possible I introduced a regression there.

xab3r:
> @xab3r Did you run this in release mode? In debug mode, that behavior is expected.
> 
> I also recently made some fixes to the vim integration to get a few tests passing, so it’s possible I introduced a regression there.

I run it in the release mode

seanstrom:
@marcocondrache when you mentioned detecting whether a device is "pre-smoothed" (macOS trackpad) or not (maybe standard mouse), is it possible to infer smooth scrolling from the scroll-delta and scroll-event frequency?

for example, it seems macOS trackpads emit a bunch of events compared to a mouse scroll-wheel, and the scroll-deltas are usually much bigger for a click-wheel mouse and much smaller for a macOS trackpad (since it emits many more events).

so could it be possible to sample the scroll device and guess if it's acting as a smoothed device?

marcocondrache:
@seanstrom As I mentioned in my previous message, this is not deterministic.

Some mice, such as Logitech devices, can emit scroll events with pixel-precise deltas, just like trackpads. With the information currently available, there is no reliable way to distinguish between them at the software level.

This is also reflected in the macOS [documentation](https://developer.apple.com/documentation/appkit/nsevent/hasprecisescrollingdeltas?language=Objc) for `hasPreciseScrollingDeltas`, which explicitly states that both some mice and trackpads can provide precise scrolling deltas:

> This property is valid for [NSScrollWheel](https://developer.apple.com/documentation/appkit/nsscrollwheel?language=Objc) events. A generic scroll wheel issues rather coarse scroll deltas. Some mice and trackpads provide much more precise delta. This method determines how the values of the [scrollingDeltaX](https://developer.apple.com/documentation/appkit/nsevent/scrollingdeltax?language=Objc) and [scrollingDeltaY](https://developer.apple.com/documentation/appkit/nsevent/scrollingdeltay?language=Objc) should be interpreted.

So relying solely on event timing, delta precision, or event frequency is not reliable. High-DPI mice can emit as many events as, or even more events than, trackpads, especially when accounting for software-level enhancements.

Firefox takes a similar approach: smooth scrolling is enabled by default for both mouse wheel and pixel-precise scroll inputs, without relying on a mouse-vs-trackpad distinction.

seanstrom:
@marcocondrache thanks for clarifying and adding more context

I suppose I was wondering more whether if it's deterministic to detect pre-smoothing from a device, not necessarily whether a mouse or trackpad is being used. From what you mentioned, it seems like macOS provides a way to check if any device is providing precise deltas.

The reason I ask is because of the jittering issues that happen with trackpad scrolling on macOS and I was wondering whether there's a simpler path to avoid post-processing the already smooth events. Since that could side-step the jittering issues by allowing the OS to handle the smoothing. for example, we wouldn't use zed smooth scrolling for certain devices because they support `hasPreciseScrollingDeltas` and do not need more smoothing

Though you mentioned that jittering is expected behaviour in debug builds, and I'm not sure if you mean that it's a known bug to fix, or it's an implementation detail that we need to live with for debug builds. Could you explain more?

marcocondrache:
@seanstrom Jittering or lagging is expected in debug builds when smooth scrolling is enabled, mainly for performance reasons. This is similar to how opening the same file in release mode can feel fine, while in debug mode it may lag or become difficult to use. The two environments are not fully comparable from a performance perspective because release builds benefit from many compiler optimizations.

That said, the jittering you are currently experiencing could also be a regression caused by the fixes I recently made, so I do not want to dismiss it as only a debug-build performance issue.

Smooth scrolling is not lightweight, especially in Zed. We need to persist the scroll position, and the behavior also depends on the number of scroll events we receive. We also cannot rely on the OS to handle this for us, since Zed does not use native system components. We draw and compose everything ourselves using the OS graphics APIs.

A device supporting precise deltas does not mean that smoothing is being applied. It only means the device can report more precise movement values. If precise deltas were enough on their own, large applications like Chrome and Firefox would not need to implement smooth scrolling for those devices.

We also need to distinguish between actual regressions, jittering caused by debug-build performance, and other possible causes. Right now, there is no reliable way to ensure that smooth scrolling works correctly other than testing it directly. So if I make fixes in the vim-related code just to get some tests passing, there is a risk that I might introduce a visual bug in the animation or state handling.

That is why I am working on a more comprehensive test suite: to make sure we can catch regressions without accidentally breaking the visual behavior of smooth scrolling.

seanstrom:
@marcocondrache  I see but I suppose when you mention:

> A device supporting precise deltas does not mean that smoothing is being applied. It only means the device can report more precise movement values.

I thought `hasPreciseScrollingDeltas` is on the scroll event (at least on macOS), so it would be the hint that the values are already smoothed (since the docs mention using that flag to interpret the event data).

Also I'm not sure what you mean by needing to handle smoothing scrolling internally, since on macOS with latest stable Zed I get similar smooth-scrolling to what I see on VSCode. Am I missing something here?

marcocondrache:
@seanstrom Again, having precise delta events still doesn’t imply any kind of smoothing. Logitech mice can report precise deltas (so pixel based deltas) for every event while still scrolling line-by-line, so the resulting behavior is not smooth scrolling.

What the docs are saying is that we should interpret the event delta as raw pixel-based values rather than line-based values. See: https://github.com/marcocondrache/zed/blob/4640666e094d518bb622a6c172a35679bb92d1c8/crates/gpui_macos/src/events.rs#L272

marcocondrache:
@seanstrom Don’t get me wrong: when we have precise deltas most of the time, such as with macOS trackpads, we effectively already get continuous scrolling. That said, this does not necessarily apply to all devices or cases.

I think it depends on how we want to handle this. When I initially started this PR, I disabled smooth scrolling for precise-delta devices, but I later removed that change after realizing that my mouse was no longer getting smooth scrolling.

One option would be to provide a setting to disable it, similar to Firefox. Another would be to implement a best-effort classifier, like VS Code does, which tries to infer the device based on the raw values:

https://github.com/microsoft/vscode/blob/321cfba1d31e253c48ee91cd10199483b85a0c34/src/vs/base/browser/ui/scrollbar/scrollableElement.ts#L128

In any case, I would prefer to keep the classifier for a separate PR and, for now, agree on a sensible default while acknowledging that we may not cover smooth scrolling perfectly in all cases.

Happy to discuss this further. I’d also love to hear your opinion, @MrSubidubi

MrSubidubi:
I do not have a strong opinion in either direction, but 

> In any case, I would prefer to keep the classifier for a separate PR and, for now, agree on a sensible default while acknowledging that we may not cover smooth scrolling perfectly in all cases.

definitely sounds like a good idea overall. Happy to go with whatever you decice upon for this here.

seanstrom:
@marcocondrache yeah I can see now that the pixels values from the precise-deltas could be too large (not tiny enough for continuous smooth scrolling), so we need to smooth them out. That makes sense, thanks again for walking through that.

The classification approach is pretty interesting, I imagine that approach could allow for devices with native smooth scrolling to work without extra processing, though I can see your point about it complicating this PR.

---

One thing to double check, if you don't mind, is the behaviour for scrolling when the feature is disabled. I think either something changed for the default scrolling behaviour (or the settings toggle for smooth scrolling is inverted), because I'm only seeing jittering (for a release build) when I have the feature disabled.

I think I was testing with the most recent changes, here's a screen capture:

https://github.com/user-attachments/assets/a80b6c08-e78c-4182-8b42-c1d601dae4bb

crates/editor/src/scroll.rs
```diff
@@ -658,6 +956,56 @@ impl ScrollManager {
...
+
+        if self
+            .scroll_animation
+            .is_some_and(|a| a.target_position() == target_position)
+        {
+            return;
+        }
+
+        if let Some(animation) = &mut self.scroll_animation {
+            animation.restart(target_position);
```
MrSubidubi: Think the error might lie here @marcocondrache , from a quick glance before I'll sign off, we could have errors when `!self.smooth_scroll` and `behavior == Some(ScrollBehavior::RequestAnimation)` since then we'll not hit that `unwrap_or_default`, not the `== Instant` right after and might end up with that flickering. Might be wrong though.

Anyway, good find and think worth a test if that or something similar is the case

marcocondrache: Nice catch indeed. I’ve fixed it, and I just need to add tests.

Unfortunately, it didn’t resolve the jittering issue when smooth scrolling is disabled, so I’ll need to dig deeper into that.

marcocondrache:
@seanstrom @xab3r  The jittering should now be fixed.

seanstrom:
@marcocondrache thanks for the fixes, seems to be working well when tested on macOS with my trackpad (with both smooth-scrolling enabled and disabled).


https://github.com/user-attachments/assets/1892ded6-60b4-49c5-95c3-88e55a10753d

---

Though I was testing with my Logitech MX vertical mouse (which I think also emits precise pixel-deltas when scrolling), and it seems like there's a little glitch when scrolling without smooth-scrolling enabled.

For example, occasionally when I do a full mouse-scroll by scrolling from edge-to-edge on the scroll-wheel, I notice that the end of my scroll has a weird "jump". It basically appears like I've finished scrolling and then suddenly jump to the real scroll position.

Here's a screen capture for reference (it occurs a few times in the video, but not every time):


https://github.com/user-attachments/assets/cb00e348-6435-4583-b610-70ab920e6f34

marcocondrache:
@seanstrom Thanks for these reports. The jump should be fixed as well.

seanstrom:
@marcocondrache I've been testing the recent changes, and it doesn't seem like the issue is resolved, we may want to revert the latest changes unless you're sure removing the `coalesce` is important.

Based on my manual testing, I think I found a simple way to reproduce the issue, though it did require me to do a couple of tricks to recreate the glitch.

* First thing I noticed is that I only could reproduce this glitch by toggling on/off the smooth-scroll feature. I would test scrolling with my trackpad and mouse, and eventually noticed the jump-glitch after toggling off smooth-scrolling.
* Second thing I noticed is that I needed to toggle on smooth-scrolling, and scroll with my trackpad and mouse, but interrupt the scroll events of the trackpad with the scroll events of the mouse.
  * Then I would toggle off smooth-scrolling and repeat the process of scrolling with the trackpad and interrupting the trackpad scroll events with scroll events from the mouse.
  * Afterwards, just scrolling with mouse with smooth-scrolling disabled would cause the jumping-glitch to occur

---

My best guess here is that some of the smooth-scroll state is being persisted when toggled off, and perhaps it should be cleared (?). The good news is that it didn't seem to happen unless I toggled the smooth-scroll feature on and off and used both devices (trackpad and mouse) for scrolling.

marcocondrache:
@seanstrom Once smooth scrolling is disabled, it no longer interferes with normal scrolling, so toggling it on or off shouldn’t be the cause of the regressions we observed.

I’m no longer able to reproduce the jumping behavior when scrolling only with the mouse, which I was seeing before yesterday evening’s fix. I can reproduce it when scrolling with the trackpad and then interrupting those events with the mouse, but that appears to be an existing bug on Zed (tested with Zed Preview 1.8.2), not something introduced by this PR.

FYI @MrSubidubi

xab3r:
@marcocondrache, there are still a few issues:

When the feature is `enabled`, the scrolling feels unnatural - very inertial rather than linear, not smooth. With a trackpad, I'd expect scrolling to track my finger 1:1, so I have full control over the speed. Only after I swipe and lift my finger off the trackpad should the scrolling keep going on its own, gradually slowing until it comes to a full stop within a second or so — with the exact duration depending on how fast I swiped.

The same goes for the mouse: as long as my finger is on the wheel, it feels more natural for the scroll to follow the wheel directly, without any added acceleration or smoothing.

When I disable the feature, though, I hit a regression. The flickering is gone now, but when I press a key, the scroll often lags or skips.

| smooth_scroll | scrolling using keyboard  | scrolling using trackpad            |
| ------------- | ------------------------- | ----------------------------------- |
| enabled       | ✅ works like a charm     | ❌ very inertial rather than linear |
| disabled      | ❌ a lot of lags or skips | ✅ seems fine, no regression        |

`settings.json`:

```json
"vim_mode": true,
"helix_mode": true
```

`keymap.json`:

```json
"ctrl-k": [
  "action::Sequence",
  ["vim::LineDown", "vim::LineDown", "vim::LineDown", "vim::LineDown"]
]
```

Zed version: `v1.9.x (ba134a0222)` + the feature branch
macOS 26

paperbenni:
:cry:

theguy000:
damn

aizigao:
A PR that has been under discussion for 10 months really makes it hard to have the courage to keep submitting PRs; many PRs for other issues are like this too

theguy000:
i think atleast ship them first.. then think of ways to solve it. blockers for one device should not be  the sole reason a feature should be blocked from shipping. 
there is currently memory leaking issue on zed. should the devs not revert the changes because of this? yet its not the case.

please zed team, ship it first. then debug for specific areas.

lhutyra:
> A PR that has been under discussion for 10 months really makes it hard to have the courage to keep submitting PRs; many PRs for other issues are like this too

I think all manpower went for delta dev, all zed is orphaned child. Many long awaited features, PRs became abandoned, lack of answears

vhtmui:
We can to sell the product to Anthropic. Let Claude handle the job all at once and achieve the same level as VSCode😂

PR #44829, opened 2025-12-14
Add stable_id support for match restoration

### LINKED ISSUE #35440: Outline picker loses selection when modifying the query.
### Summary

The outline picker resets manual selection when the user modifies the query.

### Description

Comparable to #28081, this is not strictly a bug, but a papercut because a manually selected item is lost.

Steps to reproduce:
1. "search buffer symbols" (cmd+shift+o)
2. type a selection (e.g. `selected` -> first match is `selected`)
3. use cursors to manually select a different match (e.g. `selected_id`)
4. modify the query such that the manually selected match still matches (delete the `d` -> `selecte`)

**Expected Behavior**:
Picker sticks to `selected_id` match, it was manually selected.

**Actual Behavior**:
Picker selection resets and focuses back to original match.

Zed: v0.197.3 (Zed) 
OS: macOS 15.5.0
Memory: 24 GiB
Architecture: aarch64

hoodie:
- Implement match_stable_id and find_match_by_stable_id for Outline
- Generate stable_id as "text:depth" for outline items
- Implement stable_id lookups for Project Symbols
- Implement stable_id support in Icon Theme Selector and Theme Selector
- Use stable_id to preserve and restore manual selections in Picker

Closes #35440

Release Notes:

- Added stickiness to manually selected items in project symbols and outline


crates/picker/src/picker.rs
```diff
@@ -444,6 +493,11 @@ impl<D: PickerDelegate> Picker<D> {
         self.delegate.set_selected_index(ix, window, cx);
         let current_index = self.delegate.selected_index();
 
+        // Track manually selected item to preserve across match updates
```
SomeoneToIgnore: This comment states pretty much the same what the code below does, so not needed?

Also, can we move the whole thing into the `if previous_index != current_index` block?

crates/outline/src/outline.rs
```diff
@@ -290,6 +290,22 @@ impl PickerDelegate for OutlineViewDelegate {
...
+    fn match_stable_id(&self, ix: usize) -> Option<String> {
+        let mat = self.matches.get(ix)?;
+        let outline_item = self.outline.items.get(mat.candidate_id)?;
+        Some(format!("{}:{}", outline_item.text, outline_item.depth))
+    }
+
+    fn find_match_by_stable_id(&self, stable_id: &str) -> Option<usize> {
+        self.matches.iter().position(|mat| {
+            if let Some(outline_item) = self.outline.items.get(mat.candidate_id) {
+                format!("{}:{}", outline_item.text, outline_item.depth) == stable_id
```
SomeoneToIgnore: Seems so wasteful to allocate an entire string for each outline — imagine doing that in huge files like `editor.rs`.

I wonder if we could make the ID less generic to be able to compare it more effectively?
E.g. `picker.rs` can define something like

```rs
struct StableId {
  key: Option<usize>,
  other_key: SharedString,
}
```

(field names are arbitrary and bad, and I'm sure we need more, this is just an illustration of an approach)

Then, we don't need to allocate at all when comparing here?

crates/picker/src/picker.rs
```diff
@@ -398,11 +415,43 @@ impl<D: PickerDelegate> Picker<D> {
...
+    ) {
+        self.set_selected_index_impl(ix, fallback_direction, scroll_to_index, false, window, cx);
+    }
+
+    /// Selects an index with "sticky" behavior - the selection will be preserved across
+    /// match updates if the selected item still matches the search query.
+    ///
+    /// Use this for user-driven selections (keyboard navigation, mouse clicks) where you want
+    /// the user's choice to be maintained as they continue typing. For programmatic selections
+    /// that should not persist, use `set_selected_index` instead.
```
SomeoneToIgnore, line 448: > For programmatic selections that should not persist, use `set_selected_index` instead.

Curious, what is the use case for these?
If we use `select_index_sticky` everywhere, what does it change?

crates/picker/src/picker.rs
```diff
@@ -704,7 +758,32 @@ impl<D: PickerDelegate> Picker<D> {
...
+        let match_count = self.delegate.match_count();
+        let index = if let Some(stable_id) = &self.manually_selected_stable_id {
+            if let Some(ix) = self.delegate.find_match_by_stable_id(stable_id) {
+                // Found the manually selected item, restore selection
+                self.delegate.set_selected_index(ix, window, cx);
+                ix
+            } else {
+                // Item no longer in results, clear manual selection and reset to first item
+                self.manually_selected_stable_id = None;
+                let ix = 0.min(match_count.saturating_sub(1));
```
SomeoneToIgnore: `0.min(any_other_usize)` is `0` or am I missing something?

Also, should we try to fall back to closest selected index instead, to avoid large list jumps?

SomeoneToIgnore (commented):
Sorry for the long wait time: had had quite a rough time and then a company-wide holiday.

Looks really great when I've tried it, thank you so much for finding a way to improve things generically.

One concern besides the listed ones: we need to test it.
At least, a test to show that the stable sorting works, and, ideally, another one to cover

> Item no longer in results, clear manual selection and reset to first item

branch.

crates/picker/src/picker.rs
hoodie: intuitively that should be fine, I was just a bit careful to not change the default in too many places at once, because I had no good overview of what needs to be tested

crates/picker/src/picker.rs
hoodie: good catch. from my manual testing I didn't perceive this as too bad, since usually you're already down to only a handful of matches by the time this happens. I'll change this you jump to the closest

hoodie: about those large jumps. wasn't there also a ticket about the heuristic that selects the very first item when opening the picker? I can't find it right now. anyway, right now the picker already picks a "relatively random" item instead of "the closest to the cursor". maybe I find that ticket again some day

crates/picker/src/picker.rs
hoodie: actually not, it subtly breaks the behavior, but I had to test it to find that out. Seems like I should add a few tests though to make sure of it.

crates/outline/src/outline.rs
hoodie: Time flies. I looked into this, this would be a bit of a larger change, because you can't really have one struct for all places. I could conceive of a trait here and implement it for specific pickers as needed. What do you think?
@SomeoneToIgnore

SomeoneToIgnore: Hello again, just today I was considering whether I should close the PR or not, good to have you back.

I am not entirely sure about

> because you can't really have one struct for all places

The places' amount is finite, is it not?
This means that in theory, we can do a bunch of `Option` fields for any case, or have an enum.

I am not sure how trait helps here at all, the only way I could think of it is the trait will have some `fn is_equals(param_1: T1, param_2: T2, ...)`-ish method which is more or less the same problem with extra steps?

Either way, whatever helps to remove those allocations is welcome.

hoodie:
https://github.com/user-attachments/assets/0f7fc7c2-5ca0-44cb-8620-f62161a7c00d

I've added it for the recent projects picker too

hoodie:
cool

crates/picker/src/picker.rs
```diff
@@ -717,7 +768,34 @@ impl<D: PickerDelegate> Picker<D> {
...
+
+                if match_count > 0 {
+                    self.delegate.set_selected_index(ix, window, cx);
+                }
+                ix
+            }
+        } else {
+            // No manual selection - clamp current index to valid range
+            let current_index = self.delegate.selected_index();
+            let ix = current_index.min(match_count.saturating_sub(1));
```
SomeoneToIgnore, line 792: I think this and the one above break recent attempts of https://github.com/zed-industries/zed/pull/50594 to select the first match by default, can we ensure we try and select something closer to the top here?

SomeoneToIgnore (commented):
I have trimmed it down to the scope where only the pickers influenced by the change have their code adjusted.

-----------------------

When re-testing, I've found 2 things with the selection counter-intuitive: when I select an item manually and first get a query that mismatches the manually selected item entirely (this it's not on the list anymore), it would be good get the manually selected item re-selected on backspace

Now, it seems that backspacing leaves the selection in some odd places sometimes, even without this fallback?
Would be good to keep it closer to top by default, as we try to do now after recent PRs.

PR #44862, opened 2025-12-15
search: Retain replace status when re-deploying active search panels

### LINKED ISSUE #15918: Persist "Toggle Replace" in Project Search / Buffer Search

When you search for something, you click the "Toggle Replace" button that display the replace options. All good, but if you do another search it will reset the search panel and you have to click again on the "Toggle Replace" to replace again, if would be good (and I think natural) if when you click the Toggle Replace button and you keep searching this sub-panel would keep showed instead of reseting.

### Environment

Zed: v0.146.5 (Zed)
OS: macOS 14.6.0
Memory: 16 GiB
Architecture: aarch64

Veykril:
Closes https://github.com/zed-industries/zed/issues/15918

Release Notes:

- Fixed search bars losing their replace state if you re-focus on them via actions or keybinds

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -729,12 +729,14 @@ impl BufferSearchBar {
             self.search_suggested(window, cx);
             self.smartcase(window, cx);
             self.sync_select_next_case_sensitivity(cx);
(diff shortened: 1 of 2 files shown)


xcrap:
Thank you!! :)

PR #44872, opened 2025-12-15
Fix inlay hint hover highlight for multi-byte UTF-8 characters

edlsh:
## Summary

Fixes #44812

The hover highlight for clickable inlay hints was not covering the last character when the label contained multi-byte UTF-8 characters (e.g., `→`).

## Problem

When hovering over an inlay hint like `→ app/Livewire/UserProfile.php`, the highlight/underline stopped one character short, leaving the final character unhighlighted.

**Root cause:** `find_hovered_hint_part()` in `hover_popover.rs` used `part.value.chars().count()` (character count) but `InlayOffset` wraps `MultiBufferOffset(pub usize)` which is byte-based. 

For a label like `→ app/Livewire/UserProfile.php`:
- Byte length: 32 (the `→` character is 3 bytes in UTF-8)
- Character count: 30

This mismatch caused the calculated highlight range to end 2 bytes short.

## Changes

1. **Use byte length instead of character count** (line 112):
   ```rust
   // Before (buggy)
   let part_len = part.value.chars().count();
   
   // After (correct)
   let part_len = part.value.len();
   ```

2. **Fix boundary condition** (line 113): Changed `>` to `>=` for correct `[start, end)` range semantics when hovering at part boundaries.

3. **Rename variable** for clarity: `hovered_character` → `offset_in_hint` since it's a byte offset, not a character position.

4. **Add unit test** reproducing the exact scenario from the issue with multi-byte UTF-8 characters.

## Testing

Added `test_find_hovered_hint_part_with_multibyte_characters` which:
- Verifies the label `→ app/Livewire/UserProfile.php` has 32 bytes but 30 characters
- Tests hovering at the last byte correctly returns the full range
- Tests boundary behavior with multiple label parts containing multi-byte characters

Release Notes:

- Fixed inlay hint hover highlight not covering the last character when the label contains multi-byte UTF-8 characters

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -106,12 +106,12 @@ pub fn find_hovered_hint_part(
     hovered_offset: InlayOffset,
 ) -> Option<(InlayHintLabelPart, Range<InlayOffset>)> {
     if hovered_offset >= hint_start {
-        let mut hovered_character = hovered_offset - hint_start;
+        let mut offset_in_hint = hovered_offset - hint_start;
         let mut part_start = hint_start;
         for part in label_parts {
-            let part_len = part.value.chars().count();
-            if hovered_character > part_len {
-                hovered_character -= part_len;
+            let part_len = part.value.len();
+            if offset_in_hint >= part_len {
+                offset_in_hint -= part_len;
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #44882, opened 2025-12-15
lsp: Support on-type formatting request with newlines

### LINKED ISSUE #12383: Automatically inserting `end` in Ruby
### Check for existing issues

- [X] Completed

### Misc notes

From vim-ruby, I'm used to `end` being automatically inserted when I type `def something<Enter>` or `foo do<Enter>` or `case<Enter>`. Could this functionality be added to the Ruby extension?

osiewicz:
We called out to `request_on_type_formatting` only in handle_input function, but newlines are actually handled by editor::Newline action.

Closes #12383

Release Notes:

- Added support for on-type formatting with newlines.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -5018,6 +5018,9 @@ impl Editor {
 
             this.change_selections(Default::default(), window, cx, |s| s.select(new_selections));
             this.refresh_edit_prediction(true, false, window, cx);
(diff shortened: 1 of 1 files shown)


osiewicz:
@janko I'm sorry you had to wait so long for something that was admittedly pretty simple in hindsight.

janko:
Great, thanks! 🙌🏻 I sensed it's just a matter of calling the LSP request, but I understand it's not always easy to tell whether a problem has a simple solution 😄

PR #44887, opened 2025-12-15
Introduce worktree trust mechanism

### LINKED ISSUE #12589: Zed downloads NodeJS binary and npm packages from Internet without user’s consent
### Check for existing issues

- [X] Completed

I noticed that Zed automatically downloads the NodeJS binary from https://nodejs.org without asking or even informing the user about it. Right after starting it and opening a file, without doing anything else. And there’s no option to disable it.

This is completely unacceptable!

Not just for security reasons but also from a usability point of view. I’m currently connected via metered LTE, and Zed has just eaten up 14 MiB of my plan. Moreover, I already have node installed and on PATH. Also, the downloaded binary is somehow corrupted and it wouldn’t work on my system anyway because it’s built against glibc (that’s how I noticed it in the first place).

And to make matters worse, if it did work, it would start installing arbitrary packages from npmjs.com via npm and running their scripts. This represents a huge attack vector.

This approach is completely unacceptable for anyone who’s concerned about cybersecurity and for virtually all companies, at least in the EU, because of cybersecurity laws, related certifications and audits.

EDIT: Now I found that it downloads ([here](https://github.com/zed-industries/zed/blob/de8ef081436317689333b29324b3a64343745914/crates/supermaven_api/src/supermaven_api.rs#L189)) even some _proprietary_ binary from https://supermaven.com, i.e. unaudited and unauditable code, without any verification (except TLS)! At least this is not downloaded by default… I hope…

EDIT2: Zed also automatically downloads and executes prebuilt language servers for C#, Clojure, Deno, Elixir, Gleam, GLSL, Lua, Terraform, Toml and Zig. It automatically resolves the latest version available on GitHub and downloads it, again, without any verification.

### Environment

N/A

_No response_

_No response_

### Somehow related issues

- #12358
- #12354
- #12703

SomeoneToIgnore:
~Closes https://github.com/zed-industries/zed/issues/12589~
(https://github.com/zed-industries/zed/pull/45129 had removed the "global" semantics bits)

Forces Zed to require user permissions before running any basic potentially dangerous actions for each worktree: parsing and synchronizing `.zed/settings.json`, downloading and spawning any language and MCP servers based on those settings.
There are more we can add later, among the ideas: global `NodeRuntime` limitations, `prettier`, `copilot`, global MCP servers' installation and spawn, DAP downloads on debugger start, Python virtual environment, etc.

By default, Zed starts a worktree in restricted mode and shows a `! Restricted Mode` in the title bar, no aforementioned actions are executed.
Clicking it or calling `workspace::ToggleWorktreeSecurity` command will bring a modal to trust worktrees or dismiss the modal:

This works on local, SSH and WSL remote projects, trusted worktrees are persisted between Zed restarts.
There's a way to clear all persisted trust with `workspace::ClearTrustedWorktrees`, this will restart Zed.

This mechanism can be turned off with settings:
```jsonc
"session": {
  "trust_all_worktrees": true
}
```
in this mode, all worktrees will be trusted by default, allowing all actions, but no auto trust will be persisted: hence, when the setting is changed back, auto trusted worktrees will require another trust confirmation.

This settings switch was added to the onboarding view also.

Release Notes:

- Introduced worktree trust mechanism, can be turned off with `"session": { "trust_all_worktrees": true }`


maan2003:
Is remote(ssh) projects can't run arbitrary commands on local viewer part of security model? if not is there a way to only trust on remote, but not locally. 

I use zed to connect VMs over ssh, having a local system be safe from remote system is a nice property to have.

SomeoneToIgnore:
I am not aware of how remote per se can force the client to execute an arbitrary code, but everything can be possible with further updates if those are programmed in that way.

So, the idea is to have a setting to trust remotes only?
Can work, I'll see what other feedback on that part will appear — maybe we'll have to rework this into something different than just a boolean, e.g. some "whilelist of paths on a specific host".

jirutka:
Do I understand it correctly that it’s nothing, or everything? It won‘t ask to allow specific actions, such as downloading and running particular MPC, but only refuse to do anything until you give it a blank check to do _anything_, still without informing the user?

SomeoneToIgnore:
There's a doc that explains things, have you checked that? 

https://github.com/zed-industries/zed/pull/44887/changes#diff-294284b0a04619bbc8c38bddab1bae5f14abc383ccf8652de9e305e0779354dc

It's not exactly nothing or everything, as there are 4 "groups" per host — no specific actions indeed.

> still without informing the user?

That part is not clear to me, can you elaborate?

teohhanhui:
This feels like it's just reproducing vscode's broken trust model.

PR #44901, opened 2025-12-15
nix: Pin `cargo-about` to 0.8.2

jrobsonchase:
`cargo-about` got pinned to 0.8.2 in https://github.com/zed-industries/zed/pull/44012, but this isn't exactly "easy" to accomplish in nix. The version of nixpkgs in the flake inputs uses the proper version, but if you override the nixpkgs input or use the provided overlay, you might end up trying to build with a bad version of `cargo-about`.

Since nixpkgs is versioned as a whole, your options are (in rough order of desirability):
1. Hope that nixpkgs simply includes multiple versions of the same package (common for things with stable major versions/breaking changes)
1. Use either `override` or `overrideAttrs` to provide different version/source attributes
1. Depend on multiple versions of nixpkgs to get the specific versions of the packages you want
1. Vendor the whole package build from a specific point in its history

Option 1 is out - there's only one version of cargo-about in nixpkgs.

Option 2 is already being used for `cargo-bundle`, so it seems best to continue the existing pattern.

Option 3 always just feels a bit icky and opaque to me.

Option 4 is usually"fine" for small package definitions that aren't actually much bigger than the overridden attributes would have be with the Option 2 approach. I couldn't get 2 working before noticing that the problems had been solved already, so  I had originally taken this approach.

This shouldn't ever cause an unnecessary rebuild since there's only one 0.8.2 input source, so anything that was already using it will already be cached.

I'm also updating nixpkgs to the latest unstable which currently has `cargo-about 0.8.4` to prove that this works.

Unrelatedly, I also ran `nix fmt` as a drive-by change. `nix/build.nix` was a bit out of spec.

Release Notes:

- N/A


jrobsonchase:
@cla-bot check

jrobsonchase:
Oh, hah. I just realized that if I had just read the reformatted sections of `nix/build.nix`, I would have seen where I went wrong with the `overrideAttrs` approach.

So that's definitely an option and might be cleaner than the vendored file. Happy to switch to that approach!

jrobsonchase:
Alright, done futzing with this now, sorry about the thrashing :pray:

rrbutani:
> Oh, hah. I just realized that if I had just read the reformatted sections of `nix/build.nix`, I would have seen where I went wrong with the `overrideAttrs` approach.
> 
> So that's definitely an option and might be cleaner than the vendored file. Happy to switch to that approach!

thanks for making the change, was about to leave a comment pointing at #28061 😅 

fwiw there now is a [nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/435239) that makes the change that the comment describes; hopefully we'll be able to switch to just a normal `overrideAttrs` soon:

https://github.com/zed-industries/zed/blob/0410b2340c62eb2ceb7c844aa7fe6a9d21415d5a/nix/build.nix#L110-L113

PR #44916, opened 2025-12-15
Make middle click not mousedown close tabs

### LINKED ISSUE #44856: Middle clicking on a tab closes it on mousedown instead of mouseup
### Reproduction steps

1. Open a file
2. Middle click the tab, drag off the tab, then let go

The tab closes on mousedown instead of mouseup. This is incorrect.

If you middle click the tab, drag off the tab, then let go, the tab should not close.

OS: macOS 15.7.2
Memory: 16 GiB
Architecture: x86_64

Zachiah:
Closes #44856

Release Notes:

- Make a full middle click be required to close tabs not just a middle mousedown


crates/gpui/src/interactive.rs
Zachiah, line 1: If you would prefer me to inline it, just give me the word. I figured this was useful since the analogous for left/middle already exist.

Zachiah:
This PR needs extensive testing. I hope I took the right approach. This is by far my most substantial PR to zed so far so please have grace 🙃. I wonder if the methods I added should be used in other places? I added `on_right_click` even though I didn't need it because it felt crazy to add `on_middle_click` but not `on_right_click`

Zachiah:
I added `on_right_click` and `on_middle_click` instead of adding a first parameter to match the API of say `on_mouse_down` because I didn't want to make this a breaking change to gpui and have to update a billion places. If it is preferable to make a breaking change and add it as a first parameter just let me know and I will get working on that. My thought is that if we ever did want to do that we could change it to that from what it is currently fairly easily when we had other breaking changes say. I honestly don't know enough about how gpui is used and if it's fine to break it willy nilly to make the right call on my own.

crates/gpui/src/elements/div.rs
```diff
@@ -522,6 +522,38 @@ impl Interactivity {
...
+        &mut self,
+        listener: impl Fn(&ClickEvent, &mut Window, &mut App) + 'static,
+    ) where
+        Self: Sized,
+    {
+        self.middle_click_listeners
+            .push(Rc::new(move |event, window, cx| {
+                listener(event, window, cx)
+            }));
+    }
```
Anthony-Eid: We typically follow MDM standards which would use have `on_aux_click` instead of these two callbacks.
https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event

Anthony-Eid:
https://github.com/zed-industries/zed/blob/a628d5d467dda8b9f8f6b1c9234d4a7a877dad6d/crates/gpui/src/elements/div.rs#L2270 You should add the new click listeners you added to this check

Zachiah:
@Anthony-Eid Thanks for the helpful feedback and review. I updated it. Sorry about the force push. The sqlite db had conflicts for me locally so I rebased and that fixed it.

PR #44947, opened 2025-12-16
Trigger auto-fix auto-matically

ConradIrwin:
This updates our CI workflow to try to run the autofix.yml workflow
if any of prettier, cargo fmt, or cargo clippy fail.

Release Notes:

- N/A


Bertie690:
@ConradIrwin 
Have you tested that this works on forks of the `zed` repository?

Forks have different permissions than commits from the same repo, so `git commit` may or may not work.

EDIT - i see the `actions/create-github-app-token` to make an app token - it _should_ work but I'll LYK if my CI build breaks tomorrow morning i guess

ConradIrwin:
@Bertie690 I believe that this would only trigger when PRs are sent to the main Zed repo (because of the existing `if` statements on the jobs).

When you send a PR to our repo you have the option of "allow maintainers to push to your branch", I am not sure what that means for @zed-zippy[bot], but would guess that it would be allowed – we'll have to test it.

There is a more severe issue I noticed though which is that the auto_fix pipeline has access to the zippy secrets; and because `cargo clippy --fix` is a vector for remote code execution (via `build.rs`), if we ran this on untrusted code then we might leak the secret erroneously.

I'm going to fix that, but I'm also considering only enabling this for internal PRs initially (though it would be *super* nice if it happened for community PRs too).

PR #44963, opened 2025-12-16
Bump tree-sitter to the latest

SomeoneToIgnore:
Release Notes:

- N/A


SomeoneToIgnore:
image

wasmtime@33.0.2 is vulnerable still, but the PR only aims to bump the tree-sitter hence uses the least needed version for that.
wasmtime@39.x is quite a bump on itself, as API changes are massive, so we'll merge things as is now.

PR #44985, opened 2025-12-16
util: Fix PathWithPosition parsing for parenthesized filenames

### LINKED ISSUE #44962: Can not open file in zed if filename include (1)
### Reproduction steps

on windows 11, right click `name(1).txt` then select Open in zed.

zed open `name`

Image

Zed Nightly 0.218.0 
a1dbfd0d777615cbe83b75dfd5237bb392b20631

0.218.0+nightly.2012.a1dbfd0d777615cbe83b75dfd5237bb392b20631

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

cppcoffee:
Closes https://github.com/zed-industries/zed/issues/44962

_Note from @HactarCE: There's still the case of `file(1)` that's totally ambiguous between a filename and a parenthesized file+row in [MSBuild diagnostic format](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=visualstudio). I'm not sure what to do about that besides having the CLI use subtly different parsing._

Release Notes:

- Fix PathWithPosition parsing for parenthesized filenames


HactarCE:
Thank you for the contribution! This caused another test to fail. I've done a similar thing but modifying the regex directly instead so it only affects the case you mentioned.

There's still the case of `file(1)` that's totally ambiguous between a filename and a parenthesized file+row in [MSBuild diagnostic format](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=visualstudio). I'm not sure what to do about that besides having the CLI use subtly different parsing.

SomeoneToIgnore (changes requested):
> I'm not sure what to do about that besides having the CLI use subtly different parsing.

This sounds odd: are we sure we're doing something reasonable?
Previously, the whole idea of cli/main was to call a `canonicalize` on input files, to ensure those exist and then, if they do not, we could parse it with the row/column extractor.

I did not dive deep, but why does not this work?

HactarCE:
Ah, I didn't realize we tried to open the file in the CLI before doing the whole `PathWithPosition` parsing. I'm unable to repro the original issue on macOS. I'll try it on Windows.

HactarCE:
~~Unable to reproduce on Windows as well.~~ Will discuss more in the original issue.

SomeoneToIgnore:
So based on the discussion, we see some non-determinism here?
I have not dived deep into the issue (lmk if that's needed, can do a bit later then) but if this is true:

* our canonicalization approach + parsing works

CLI: https://github.com/zed-industries/zed/blob/8c7a04c6bf3c26082f0cb0501a3ddf180968dd55/crates/cli/src/main.rs#L135-L165

Zed: https://github.com/zed-industries/zed/blob/8c7a04c6bf3c26082f0cb0501a3ddf180968dd55/crates/zed/src/zed/open_listener.rs#L652-L668

and this fix is not fixing the right thing

* there is some odd race somewhere and this needs a fix

PR #44988, opened 2025-12-16
collab: Update extensions filter to consider description

### LINKED ISSUE #43079: Different results while searching for extensions, website vs Zed app
### Summary

Searching for extensions on the website searches titles and descriptions 👍🏻, but within the app only titles are searched 👎🏻

### Description

Posted after encouragement from @dinocosta , in:
Open add spellchecking to Agent Panel #42281

**Example:**
https://zed.dev/extensions?query=spell
Returns 5 matches because it searches both "titles" (IDs) and "descriptions" in each card 👍🏻

But, if you are in Zed, then Extensions menu item, then search using the query... spell,
you will get 3 matches, because it apparently (live) fuzzy searches the titles, only. 👎🏻

1. **Suggested short-term band aid:**
Change placeholder text in `extensions_ui.rs` from
"Search extensions...", to
"Search Extension IDs", to make the existing behavior a little more clear.

1. **Next improvement:**
Include descriptions while doing searches in Zed.

1. **Big feature request:**
Greatly expand extension search in Zed to allow optional filtering for Version, Author info, etc.
(Leave website searching as is.)

**Temporary hack:**
Copy text from https://zed.dev/extensions web page
and paste into your favorite editor 😁

You will see all the text, even the text that was truncated,
and will be able to search with Regex, etc.

(Specs included, but I do not believe this is dependent on Zed version, platform, or browser used for that matter.)

**Expected Behavior**:
Consistent search results as discussed above. This will help both users and extension publishers.

**Actual Behavior**:
Different, and initially confusing search results

Zed: v0.213.3 (Zed Preview)
OS: macOS 14.7.1
Memory: 8 GiB
Architecture: aarch64

N/A

dinocosta:
Update the condition built by `Database.get_extensions` in order to also apply the same `ILIKE` filter on the `extension_version.description` column so that, if provided, a search filter can match on either the extension's name or description.

The `backend-postgres` feature is also added to the `sea-query` crate so that we can access the `ilike` method, which would otherwise not be available even though we now we're running a Postgres database.

Closes #43079 

Release Notes:

- Improved search results in "Extensions" to also compare the extension's description to the provided search query

### DIFF
--- a/crates/collab/Cargo.toml
+++ b/crates/collab/Cargo.toml
@@ -49,6 +49,7 @@ rpc.workspace = true
 scrypt = "0.11"
 # sea-orm and sea-orm-macros versions must match exactly.
 sea-orm = { version = "=1.1.10", features = ["sqlx-postgres", "postgres-array", "runtime-tokio-rustls", "with-uuid"] }
+sea-query = { version = "=0.32.7", features = ["backend-postgres"]}
 sea-orm-macros = "=1.1.10"
 semver.workspace = true
 serde.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3336,6 +3336,7 @@ dependencies = [
  "scrypt",
  "sea-orm",
  "sea-orm-macros",
+ "sea-query",
  "semver",
  "serde",
  "serde_json",
@@ -14427,6 +14428,7 @@ dependencies = [
  "inherent",
  "ordered-float 4.6.0",
  "rust_decimal",
+ "sea-query-derive",
  "serde_json",
  "time",
  "uuid",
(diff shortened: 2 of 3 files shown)


crates/collab/Cargo.toml
```diff
@@ -49,6 +49,7 @@ rpc.workspace = true
 scrypt = "0.11"
 # sea-orm and sea-orm-macros versions must match exactly.
 sea-orm = { version = "=1.1.10", features = ["sqlx-postgres", "postgres-array", "runtime-tokio-rustls", "with-uuid"] }
+sea-query = { version = "=0.32.7", features = ["backend-postgres"]}
```
dinocosta, line 52: This is the version that's currently present in `Cargo.lock`, not entirely sure if we need to pint it to this specific version so I'm open to changing this 👍

maxdeviant:
@dinocosta Thanks for taking this on, but I have concerns about the database performance of extending the search filtering.

Additionally, I don’t think a simple `ILIKE` is going to give high enough quality search results (in fact, I think it would degrade them).

I don’t think we should ship this.

dinocosta:
Thank you for the quick review @maxdeviant ! 🙂

Is there a different approach you'd be able to consider? Alternatively, we could also instead update https://zed.dev/extensions to only filter by the extension name, and not the description, as I believe that's also an acceptable fix for #43079 .

Otherwise, shall we go ahead and close #43079 as not planned?

maxdeviant:
> Thank you for the quick review @maxdeviant ! 🙂
> 
> Is there a different approach you'd be able to consider? Alternatively, we could also instead update https://zed.dev/extensions to only filter by the extension name, and not the description, as I believe that's also an acceptable fix for #43079 .
> 
> Otherwise, shall we go ahead and close #43079 as not planned?

Yea, I think we should make zed.dev’s search just use what is currently available in the extensions API and not add any additional search logic of its own, that way the results will be consistent.

PR #44996, opened 2025-12-16
git_ui: Fix double slash in commit URLs

cppcoffee:
Release Notes:

- Fixed double slash in commit URLs

The github_url variable was generating URLs with an extra slash like
"https://github.com//user/repo/commit/xxxx" due to manual string formatting
of the base_url() result.

Fixed by replacing manual URL construction with the proper
build_commit_permalink() method that uses Url::join() for correct
path handling, consistent with how other Git hosting providers
construct URLs.


cole-miller:
Thanks!

PR #45002, opened 2025-12-16
git_ui: Add support for collapsing/expanding entries with your keyboard

RemcoSmitsDev:
This PR adds support for collapsing/expanding Git entries with your keyboard like you can inside the project panel and variable list.

I noticed there is a bug that selecting the next entry when you are on the directory level will select a non-visible entry. Will fix that in another PR, as it is not related to this feature implementation.

**Result**:

https://github.com/user-attachments/assets/912cc146-1e1c-485f-9b60-5ddc0a124696

Release Notes:

- Git panel: Add support for collapsing/expanding entries with your keyboard.

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -900,6 +900,8 @@
   {
     "context": "GitPanel && ChangesList",
     "bindings": {
+      "left": "git_panel::CollapseSelectedEntry",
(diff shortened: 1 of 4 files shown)


cole-miller:
Thanks!

RemcoSmitsDev:
It should be good now!

PR #45016, opened 2025-12-16
git: Prevent customizing commit message prompt for legacy Zed Pro users

bennetbo:
We need to prevent this, since commit message generation did not count as a prompt in the old billing model.
If users of Legacy Zed Pro customise the prompt, it will count as an actual prompt since our matching algorithm will fail.
We can remove this once we stop supporting Legacy Zed Pro on 17 January.

Release Notes:

- N/A


maxdeviant (approved):
Looks good!

PR #45030, opened 2025-12-16
git_ui: Fix select next/previous entry selects non-visible entry when tree view is enabled

RemcoSmitsDev:
Before this commit, we would select a non-visible entry when a directory is collapsed. Now we correctly select the visible entry that is visually the previous/next entry in the list.

**Note**: I removed the `cx.notify()` call as it's already part of the `self.scroll_to_selected_entry(cx)` call. So we don't notify twice :).

Follow-up: https://github.com/zed-industries/zed/pull/45002

**Before**

https://github.com/user-attachments/assets/da0b8084-0081-4d98-ad8a-c11c3b95a1b7

**After**

https://github.com/user-attachments/assets/8a16afb0-fdde-4317-b419-13143d5d608e

Release Notes:

- git_ui: Fix select next/previous entry selects non-visible entry when tree view is enabled

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -984,6 +984,14 @@ impl GitPanel {
             return;
         }
 
+        if matches!(
+            self.entries.get(new_index.saturating_sub(1)),
+            Some(GitListEntry::Header(..))
+        ) && new_index == 0
+        {
+            return;
+        }
+
         if matches!(self.entries.get(new_index), Some(GitListEntry::Header(..))) {
             self.selected_entry = Some(new_index.saturating_sub(1));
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks Remco, this looks great! Happy to merge but I noticed while testing that we also have a bug with the selection disappearing when we go up past the first (visual) entry, in case you want to fix that as well :)

RemcoSmitsDev:
Hey Cole, thanks for your review. Nice catch, fixed it:

https://github.com/user-attachments/assets/ea76a2f4-bb0a-4288-a334-fa03d6ac04c8

cole-miller:
Thanks!

PR #45037, opened 2025-12-16
gpui: Remove intermediate allocations when reconstructing text from a `TextLayout`

tidely:
Remove some intermediate allocations when reconstructing text or wrapped text from a `TextLayout`. Currently creates a intermediate `Vec<String>` which gets joined, when you could join an `impl Iterator<Item = &str>`

Release Notes:


Veykril (approved):
Thanks!

Issue #43969, opened 2025-12-02
Document the different ways to configure Tailwind in your language/framework


yeskunall:
We sorta mention [how to configure the Tailwind LSP](https://zed.dev/docs/languages/tailwindcss), but the `classRegex` really depends on the language and framework you’re using. For example, here’s how to get it working in Templ files:

```json
{
  "lsp": {
    "tailwindcss-language-server": {
      "settings": {
        "includeLanguages": {
          "templ": "html"
        },
        "experimental": {
          "classRegex": ["class=\"([^\"]*)\"", "className=\"([^\"]*)\""]
        }
      }
    }
}
```

The docs only mention how to set it up for JSX/TSX projects:

```json
{
  "lsp": {
    "tailwindcss-language-server": {
      "settings": {
        "classFunctions": ["cva", "cx"],
        "experimental": {
          "classRegex": ["[cls|className]\\s\\:\\=\\s\"([^\"]*)"]
        }
      }
    }
  }
}

```

From #43968, here’s how to set it up in [Lustre](https://github.com/lustre-labs/lustre) (Gleam):

```json
{
  "lsp": {
    "tailwindcss-language-server": {
      "settings": {
        "experimental": {
          "classRegex": [ "\"([^\"]*)\""]
        }
      }
    }
  }
}
```

It would be great if we also documented how to set these up per language. Right now, we have support for Tailwind in the following languages:

- [ ] [Astro](https://zed.dev/docs/languages/astro.html)
- [ ] [CSS](https://zed.dev/docs/languages/css.html)
- [ ] [ERB](https://zed.dev/docs/languages/ruby.html)
- [ ] [HEEx](https://zed.dev/docs/languages/elixir.html#heex)
- [ ] [HTML](https://zed.dev/docs/languages/html.html)
- [ ] [TypeScript](https://zed.dev/docs/languages/typescript.html)
- [ ] [JavaScript](https://zed.dev/docs/languages/javascript.html)
- [ ] [PHP](https://zed.dev/docs/languages/php.html)
- [ ] [Svelte](https://zed.dev/docs/languages/svelte.html)
- [ ] [Vue](https://zed.dev/docs/languages/vue.html)

They don’t all need to be done at once. If you’re aware of how to set it up in one of the languages above, PR welcome.

yeskunall:
Here’s a well-documented example of how to get it working in [Ruby](https://zed.dev/docs/languages/ruby#using-the-tailwind-css-language-server-with-ruby).

PR #45057, opened 2025-12-17
docs: Add Tailwind configuration section for JavaScript/TypeScript

andre-eriksson:
Addresses some tasks in #43969. Namely adding TailwindCSS documentation for the following languages: HTML, JavaScript and Typescript.

**Some Notes**
 - Maybe the additional information in the HTML section is unnecessary, unsure open to suggestions.
 - I tried utilizing capturing groups with alternatives like `\\.(add|remove|toggle|contains)` but this didn't seem to work, so I was forced to use multiple lines.

Release Notes:

- N/A


danilo-leal (approved):
Thank you! Ended up changing a few things here:

- I don't think we need any content in the HTML page given Tailwind works out of the box there with no config required
- Clarified that for JS and TS that configuration is required if you're working on their vanilla versions (just to separate it from .jsx and .tsx, where Tailwind works out of the box, too)
- Remove the need to declare the language server in all snippets given it's available by default in Zed; users only need to declare the settings if required

PR #45067, opened 2025-12-17
agent_settings: Update detailed summary prompt to enhance analysis structure and thoroughness

aeroxy:
Release Notes:

- Update detailed summary prompt to enhance analysis structure and thoroughness


danilo-leal:
Hey, thanks for the PR! I get where you're coming from by beefing up this prompt so it's more thorough in producing a better summary of the thread, but I think this is something we'd like to potentially eval on and iterate on some different approaches before committing to one. I'm also a bit concerned here due to how big of a size different the proposal is vs. what we have. That said, I'll close this for now, but thank you again.

PR #45073, opened 2025-12-17
cli: Use Terminal Panel's window for `zed --add`

cglwn:
This makes the `zed --add <PATH>` command use the window it was run from when run from a Terminal Panel. I run into this paper cut quite a lot working with multiple projects and preferring to create new files through the CLI instead of the GUI.

**Before**
With two windows open, running `zed --add a.txt` in one window's terminal might open the file `a.txt` in the other window.

[Screencast from 12-17-2025 02:24:45 AM - zed --add bug before.webm](https://github.com/user-attachments/assets/30816add-91e1-41c3-b2e3-6a0e6e88771a)

**After**
With this change, it will use the window of the Terminal Panel.

[Screencast from 12-17-2025 02:16:09 AM - zed --add bug after.webm](https://github.com/user-attachments/assets/5141518e-5fb0-47d1-9281-54c0699ff7f5)

Release Notes:

- Improved `zed --add` command to prefer window it was run from


cglwn:
@cla-bot check

cglwn:
Not obvious to me why the Clippy is failing on unrelated files, I didn't run Clippy locally but I'll take a look and update this PR!

SomeoneToIgnore:
I think this is caused by stale PR merge, you need to rebase onto fresh `main`.

cglwn:
Hmm, I was only six behind but I'll give that a shot! Did a Sync Fork on my fork which auto merged in `main`. Will leave for now to see if it passes and can clean up history if needed.

**Edit:** It passed so this should be good to review now :)

cglwn:
Did more testing and for an existing file that exists in another workspace belonging to a different window this makes it open in the window `zed --add` was run from too which I think makes sense.

When run from an external terminal (i.e. not in Zed), it opens in the window that has the workspace containing the file.

SomeoneToIgnore (approved):
Nice, thank you for the test specifically.

PR #45084, opened 2025-12-17
remote: Implement client side connection support for windows remotes

Veykril:
Obviously this doesn't do too much without having an actual windows server binary for the remote side, but it does at least improve the error message as right now we will complain about `uname` not being a valid powershell command.

Release Notes:

### DIFF
--- a/crates/remote/src/remote_client.rs
+++ b/crates/remote/src/remote_client.rs
@@ -49,10 +49,58 @@ use util::{
     paths::{PathStyle, RemotePathBuf},
 };
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum RemoteOs {
+    Linux,
+    MacOs,
+    Windows,
+}
+
+impl RemoteOs {
+    pub fn as_str(&self) -> &'static str {
+        match self {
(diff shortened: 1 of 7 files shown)


crates/remote/src/remote_client.rs
```diff
@@ -89,7 +89,8 @@ pub trait RemoteClientDelegate: Send + Sync {
 const MAX_MISSED_HEARTBEATS: usize = 5;
 const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 const HEARTBEAT_TIMEOUT: Duration = Duration::from_secs(5);
-const INITIAL_CONNECTION_TIMEOUT: Duration = Duration::from_secs(60);
+const INITIAL_CONNECTION_TIMEOUT: Duration =
+    Duration::from_secs(if cfg!(debug_assertions) { 5 } else { 60 });
```
daniel-thompson, line 93: Any idea what this line is for? It looks deliberate but having a shorter timeout when code is less optimized is a bit weird.

Also, it means remote connections to slow machines fail unexpectedly with debug builds.

PR #45097, opened 2025-12-17
agent_ui: Make tool call raw input visible

tsoernes:
Screenshot 2025-12-17 at 9  28@2x

Release Notes:

- agent: Made tool calls' raw input visible in the agent UI.


danilo-leal (approved):
Thank you for kicking this off! I changed the approach here and also refined the UI so it fits better with what we have. We didn't need an additional disclosure component for this case and could also make it blend it more with the current way we display outputs.

PR #45099, opened 2025-12-17
settings_ui: Fix scrollbar breaking when UI font size changes

### LINKED ISSUE #43683: Settings Page gets bugged if font size changes
### Reproduction steps

1. Start Zed
2. Open settings window
3. Go to Appearance
4. Increase the UI font size
5. Observe that the scrollbar no longer encompasses the entire settings dialogue (and can't scroll to other settings)

This is easily fixable by going to another page in settings (e.g General) and then returning to the Appearance page 

Observe how far the scrollbar goes. 

| Expected | Actual | 
| -- |-- |
| Scrollbar still lets you navigate the entirety of "Appearance" settings" Image | Scrollbar can't navigate beyond a small snippet of the settings Image |

### If applicable, attach your Zed log file to this issue.

```log

```

</details>

### If applicable, attach your relevant Zed settings to this issue

```json

```

</details>

### If applicable, provide details about your model provider

_No response_

Zed: v0.214.5 (Zed) 
Memory: 11.5 GiB
Architecture: x86_64
GPU: Intel(R) HD Graphics 530 (SKL GT2) || Intel open-source Mesa driver || Mesa 24.2.8-1ubuntu1~24.04.1

Arch Linux

dinocosta:
When the UI font size changes, the settings window's scrollbar would break because list item measurements were cached and not invalidated.

We're using `ListState.measure_all()` to pre-measure all items for an accurate scrollbar. When font size changes, these cached measurements become stale but weren't being refreshed.

This commit fixes the issue by calling `ListState.measure_all()` when settings change, ensuring items are re-measured on the next layout without affecting scroll position.

Closes #43683

Release Notes:

- Fixed bug where changing the UI Font Size in the Settings UI would break the scrollbar for the "Appearance" page


crates/settings_ui/src/settings_ui.rs
```diff
@@ -1378,6 +1378,14 @@ impl SettingsWindow {
...
         cx.observe_global_in::<SettingsStore>(window, move |this, window, cx| {
             this.fetch_files(window, cx);
+
+            // Whenever settings are changed, it's possible that the changed
+            // settings affects the rendering of the `SettingsWindow`, like is
+            // the case with `ui_font_size`. When that happens, we need to
+            // instruct the `ListState` to re-measure the list items, as the
+            // list item heights may have changed depending on the new font
+            // size.
+            this.list_state.clone().measure_all();
```
Anthony-Eid: `measure_all()` is the wrong function you want to use here, because it tells the list to cache all element sizes but it doesn't clear the cache. You're going to want to use `this.list_state.reset(element_count)` which clears the cache. 

Also, we shouldn't clear the catch every time a setting field is changed which is what this observe subscription would do. Instead we should check if the UI font size changed by copying the old value in the closure and reading the settings content.

dinocosta: Thank you for the review @Anthony-Eid !

> measure_all() is the wrong function you want to use here, because it tells the list to cache all element sizes but it doesn't clear the cache. You're going to want to use this.list_state.reset(element_count) which clears the cache.

Sorry, should have added that I did try this (by calling the `SettingsWindow.reset_list_state` method instead of `measure_all`) and I was seeing this weird behavior where, after resetting the list state, the scroll position would also reset to the top, leading to this janky experience, which I wanted to avoid ▼

https://github.com/user-attachments/assets/93dded67-fb57-469e-a154-95e29bdaa919

> Also, we shouldn't clear the catch every time a setting field is changed which is what this observe subscription would do. Instead we should check if the UI font size changed by copying the old value in the closure and reading the settings content.

Makes total sense, thank you for the suggestion. Will update accordingly 🙂

Anthony-Eid: I just looked at the `list_state.reset` function definition and you're right, it also resets the scroll position. You might need to call `list_state.scroll_to` after reset with the logical_top position index. 

I personally don't like that solution, because it's a work around for an API we should support. We could add a `clear_cache` or `remeasure` function to the list that behaves like `reset`, but it doesn't reset the scroll position. This would only be allowed when the count of elements in the list doesn't change.

dinocosta: > I personally don't like that solution, because it's a work around for an API we should support. We could add a clear_cache or remeasure function to the list that behaves like reset, but it doesn't reset the scroll position. This would only be allowed when the count of elements in the list doesn't change.

Great suggestion! I ended up pushing https://github.com/zed-industries/zed/pull/45099/changes/9fb12ee990ab172904521621d4e1740c26eaeeff , which introduces `ListState::remeasure`. When possible, can you take a look and let me know what you think? If it looks good, I can go ahead and try adding some tests for this new method. Thanks! 🙂

Anthony-Eid: Your commit looks good! There might be odd behavior because we’re not recalculating the pixel offset of the logical top element. This is unlikely to be noticeable in the settings UI.

For example, if an element with a height of 10px has a scroll offset of 5px, and its updated size is less than 5px, it won’t be the top visible item anymore.

dinocosta: @Anthony-Eid Thank you for calling that out, would have probably missed it! Not sure if this is the best approach but I ended up pushing this commit – https://github.com/zed-industries/zed/pull/45099/changes/1b032fdaabde1f23fcaf8647c3af135bb29ef5fb – which introduces a `PendingScrollFraction` struct and adds it to `StateInner`. When the items are being laid out, if there's a pending scroll within a given item, we'll re-store the `logical_scroll_top` proportionally. Let me know what you think! ✌️ 

Here's a quick screen recording that shows the changes in effect, notice the "Line Height" label at the top ▼

<details>
<summary>Before</summary>

https://github.com/user-attachments/assets/00185c15-edbb-4568-8d9d-24213b9a5891
</details>

<details>
<summary>After</summary>

https://github.com/user-attachments/assets/299b75b7-8c50-458c-9d54-693bf01b6129
</details>

Anthony-Eid (approved):

PR #45108, opened 2025-12-17
git_ui: Fix select first entry selects the wrong visual first entry when tree view is enabled

RemcoSmitsDev:
This PR fixes a bug where the select first didn't select the first visual entry when the first entry is a collapsed directory.

Follow-up: https://github.com/zed-industries/zed/pull/45030

**Before**:

https://github.com/user-attachments/assets/5e5865cc-ec0f-471d-a81b-9521fb70df41

**After**:

https://github.com/user-attachments/assets/05562572-e43f-4d1e-9638-80e4dccc0998

Release Notes:

- git_ui: Fix select first entry selects the wrong visual first entry when tree view is enabled

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -279,6 +279,13 @@ impl GitListEntry {
             _ => None,
         }
     }
+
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks!

PR #45131, opened 2025-12-17
cli: Teach `--diff` to recurse into directories and add a `MultiDiffView`

davidbarsky:
This branch:
1. teaches `--diff `command line option to to recurse into folders if provided.
2. Adds a `MultiDiffView` that shows _all_ changed files in a single, scrollable view.

This is necessary to provide a smooth user experience for `jj` and Zed users who wish to use Zed as their jj difftool.

I'm not fully sure how this change interacts with https://github.com/zed-industries/zed/pull/44936, or what plans y'all have in mind.

Here's a screenshot of the resulting behavior:

Screenshot 2025-12-17 at 9 10 52 AM

I setup zed to handle jj diffs by adding the following to my jj config:

```toml
[aliases]
zdiff = ["diff", "--tool", "zed"]

[merge-tools.zed]
program = "/Users/dbarsky/Developer/zed/target/debug/cli"
# omit diff-invocation-mode to keep the default (JJ passes two dirs)
diff-args = [
  "--zed", "/Users/dbarsky/Developer/zed/target/debug/zed",
  "--wait",
  "--new",
  "--diff",
  "$left",
  "$right",
]
```

Release Notes:

- `--diff`, if provided with folders instead of files, will recurse into those directories.
- Added a `MultiDiffView`, which will show all changed files within a single, scrollable view.

**AI Disclosure**: Pretty much all of this code was written using Codex with GPT-5.1-Codex. I edited by hand and tested this.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -12,6 +12,7 @@ use clap::Parser;
 use cli::{CliRequest, CliResponse, IpcHandshake, ipc::IpcOneShotServer};
 use parking_lot::Mutex;
 use std::{
+    collections::{BTreeMap, BTreeSet},
     env,
     ffi::OsStr,
     fs, io,
@@ -20,8 +21,9 @@ use std::{
     sync::Arc,
     thread::{self, JoinHandle},
 };
-use tempfile::NamedTempFile;
+use tempfile::{NamedTempFile, TempDir};
 use util::paths::PathWithPosition;
+use walkdir::WalkDir;
 
 #[cfg(any(target_os = "linux", target_os = "freebsd"))]
 use std::io::IsTerminal;
@@ -117,6 +119,7 @@ struct Args {
     #[arg(long)]
     system_specs: bool,
     /// Pairs of file paths to diff. Can be specified multiple times.
+    /// When directories are provided, recurses into them and shows all changed files in a single multi-diff view.
     #[arg(long, action = clap::ArgAction::Append, num_args = 2, value_names = ["OLD_PATH", "NEW_PATH"])]
     diff: Vec<String>,
     /// Uninstall Zed from user system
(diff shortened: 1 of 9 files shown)


crates/editor/src/editor.rs
```diff
@@ -1193,6 +1193,7 @@ pub struct Editor {
     next_scroll_position: NextScrollCursorCenterTopBottom,
     addons: HashMap<TypeId, Box<dyn Addon>>,
     registered_buffers: HashMap<BufferId, OpenLspBufferHandle>,
+    path_overrides: HashMap<BufferId, String>,
```
Veykril: This feels a bit odd, why do we need this?

crates/zed/src/main.rs
```diff
@@ -1318,6 +1348,9 @@ struct Args {
     /// Pairs of file paths to diff. Can be specified multiple times.
     #[arg(long, action = clap::ArgAction::Append, num_args = 2, value_names = ["OLD_PATH", "NEW_PATH"])]
     diff: Vec<String>,
+    /// Pairs of file paths to diff in a single aggregated view. Cannot be combined with --diff.
+    #[arg(long, action = clap::ArgAction::Append, num_args = 2, value_names = ["OLD_PATH", "NEW_PATH"], conflicts_with = "diff")]
+    diff_all: Vec<String>,
```
Veykril: I wonder if it would make sense to have this mode implied if `diff` is used by pointing it to folders instead of files?

crates/editor/src/editor.rs
Veykril: > To follow in [difftastic](https://difftastic.wilfred..uk/)'s footsteps, I trimmed the buffer labels to hide the temporary directory that jj places the diffs in.

nevermind I guess this is why, I think we can do that without adding this thoughm, will have to think a bit

Veykril: I'd say lets leave this part out of the PR for now, as even `jj describe` shows the full temp path after all. (and `git` just "works" because it puts the temp stuff into `.git` which is inside the worktree)

davidbarsky: Sounds good! I think the temp paths are a bit ugly, but I suppose without LSP support in the diffs, cleaning up the paths is more harmful than not.

crates/zed/src/main.rs
davidbarsky: I thought about that initially, but I was concerned about potentially breaking existing users. On the other hand, VS Code’s `diff` CLI argument recurses into folders.

I’ll get rid of the standalone `diff-all` subcommand and make `diff` smarter instead.

davidbarsky: done.

crates/editor/src/editor.rs
davidbarsky: removed for now.

crates/cli/src/main.rs
```diff
@@ -473,16 +574,27 @@ fn main() -> Result<()> {
     let mut paths = vec![];
     let mut urls = vec![];
     let mut diff_paths = vec![];
+    let mut synthetic_temp_dirs = vec![];
```
Veykril: I don't think these stay alive for long enough, I believe the main function will exit here once the main zed process runs? Or if there is a zed instance already running at least, this main call will rpc into that and exit instead, so that might cause issues?

davidbarsky: whoops, good point. fixed.

(It wasn't an issue whenever I use this, but I might just be getting lucky.)

PR #45172, opened 2025-12-17
agent_ui: Remove unnecessary Arc allocation

aqrln:
Follow up to https://github.com/zed-industries/zed/pull/44297.

Initial implementation in ce884443f1c38ca8da9edf9fbbb8e7fd579452cb used `Arc` to store the reference to the hash map inside the iterator while keeping the lifetime static. The code was later simplified in 5151b22e2ea37fb457cf4f88b88fe4f315306074 to build the list eagerly but the Arc was forgotten, although it became unnecessary.

cc @bennetbo

Release Notes:

- N/A


Veykril (approved):
Thanks!

PR #45180, opened 2025-12-18
terminal: Respect RevealStrategy::NoFocus and Never focus settings

### LINKED ISSUE #45179: RevealStrategy::NoFocus and RevealStrategy::Never do not respect focus settings when terminal pane has focus
## Summary

When creating terminals with `RevealStrategy::NoFocus` or `RevealStrategy::Never`, the terminal tab still receives focus if the terminal pane already has focus. This contradicts the documented behavior of these strategies.

## Current Behavior

In `crates/terminal_view/src/terminal_panel.rs`, the focus logic is:

```rust
let focus = pane.has_focus(window, cx)
    || matches!(reveal_strategy, RevealStrategy::Always);
```

This means:
- If the pane has focus OR strategy is `Always` → focus the new terminal

## Expected Behavior

Per the documentation in `crates/task/src/task_template.rs`:
- `NoFocus`: "Always show the task's pane, add the task's tab in it, but **don't focus it**"
- `Never`: "**Do not alter focus**, but still add/reuse the task's tab in its pane"

The focus calculation should only consider the explicit strategy:

```rust
let focus = matches!(reveal_strategy, RevealStrategy::Always);
```

## Impact

This affects:
- Vim users running shell commands via `:!command` (uses `NoFocus`)
- Debugger terminal spawning (uses `NoFocus`)
- Any programmatic terminal creation requesting background behavior

## Locations

The bug exists in two places:
- `add_terminal_task()` around line 790
- `add_terminal_shell()` around line 853

rabsef-bicrym:
Closes #45179

## Summary

Fixes the focus behavior when creating terminals with `RevealStrategy::NoFocus` or `RevealStrategy::Never`. Previously, terminals would still receive focus if the terminal pane already had focus, contradicting the documented behavior.

## Changes

- **`add_terminal_task()`**: Changed focus logic to only focus when `RevealStrategy::Always`
- **`add_terminal_shell()`**: Same fix

The fix changes:
```rust
// Before
let focus = pane.has_focus(window, cx)
    || matches!(reveal_strategy, RevealStrategy::Always);

// After  
let focus = matches!(reveal_strategy, RevealStrategy::Always);
```

## Impact

This affects:
- Vim users running `:!command` (uses `NoFocus`)
- Debugger terminal spawning (uses `NoFocus`)
- Any programmatic terminal creation requesting background behavior

Release Notes:

- Fixed terminal focus behavior to respect `RevealStrategy::NoFocus` and `RevealStrategy::Never` settings when the terminal pane already has focus.

### DIFF
--- a/crates/terminal_view/src/terminal_panel.rs
+++ b/crates/terminal_view/src/terminal_panel.rs
@@ -787,8 +787,7 @@ impl TerminalPanel {
                 }
 
                 pane.update(cx, |pane, cx| {
-                    let focus = pane.has_focus(window, cx)
-                        || matches!(reveal_strategy, RevealStrategy::Always);
+                    let focus = matches!(reveal_strategy, RevealStrategy::Always);
                     pane.add_item(terminal_view, true, focus, None, window, cx);
                 });
 
@@ -850,8 +849,7 @@ impl TerminalPanel {
                         }
 
                         pane.update(cx, |pane, cx| {
-                            let focus = pane.has_focus(window, cx)
-                                || matches!(reveal_strategy, RevealStrategy::Always);
+                            let focus = matches!(reveal_strategy, RevealStrategy::Always);
                             pane.add_item(terminal_view, true, focus, None, window, cx);
                         });
 


rabsef-bicrym:
@cla-bot check

Veykril (approved):
Thanks!

PR #42826, opened 2025-11-16
 Add support for virtual documents (Go to Definition with dependencies)

ghost:
Closes <https://github.com/zed-extensions/java/issues/44>

## Summary

Implements support for virtual documents in Zed, enabling "Go to Definition" to navigate into
external dependencies (libraries, JDK) by displaying decompiled source code. This resolves
https://github.com/zed-extensions/java/issues/44.

## Problem

Currently, when using "Go to Definition" on a symbol from an external dependency (e.g., ArrayList from the JDK, or a class from a Maven/Gradle dependency), Zed cannot navigate to the definition because:

1. Language servers return non-file URIs - JDTLS returns jdt://contents/...jar/.../Class.class URIs
    pointing to compiled bytecode
3. Zed only handles file:// URIs - Any non-file URI fails with "can't convert URI to path"
4. No mechanism for virtual documents - Zed has no way to display content that doesn't exist as a file
    on disk

This limitation affects:
- Java developers - Cannot view JDK source or decompiled dependencies
- Go developers - Cannot view source during debugging (DAP browser URIs)
- Rust developers - Cannot view macro expansions as virtual documents
- Any language where the LSP provides virtual content

## Solution

This PR implements a general-purpose virtual document system with the following components:

1. Virtual buffer infrastructure (buffer_store.rs)

   ```rust
   pub struct BufferStore {
       // ... existing fields
       uri_to_buffer_id: HashMap<String, BufferId>,      // URI -> Buffer lookup
       pub(crate) virtual_buffers: HashMap<BufferId, lsp::Uri>,  // Buffer -> URI reverse mapping
   }
   ```

   - Tracks buffers by URI in addition to file paths
   - Automatic cleanup when buffers are dropped
   - No impact on existing file-based buffers
2. Virtual document handling (lsp_store.rs)
   - Configuration System:

   ```rust
   pub struct VirtualDocumentConfig {
       pub scheme: String,                    // e.g., "jdt"
       pub content_request_method: String,    // e.g., "java/classFileContents"
       pub language_name: String,             // e.g., "Java"
       pub language_id: String,               // e.g., "java"
   }
   ```

   - Core Functionality:
     - open_virtual_buffer_via_lsp() - Fetches content via custom LSP requests, creates read-only buffers
     - Automatic LSP registration (textDocument/didOpen) so virtual buffers get full language support
     - Buffer tracking in buffers_opened_in_servers for LSP routing
3. Dynamic LSP requests (lsp.rs)

   ```rust
[...]

PR #45196, opened 2025-12-18
Add virtual documents

ghost:
Trying again (https://github.com/zed-industries/zed/pull/42826)

Release Notes:

### DIFF
--- a/crates/lsp/src/lsp.rs
+++ b/crates/lsp/src/lsp.rs
@@ -1419,11 +1419,38 @@ impl LanguageServer {
         )
     }
 
+    fn create_response_handler<R>(
+        executor: &BackgroundExecutor,
+        tx: oneshot::Sender<Result<R>>,
+    ) -> ResponseHandler
+    where
+        R: 'static + Send + DeserializeOwned,
+    {
+        let executor = executor.clone();
+        Box::new(move |result| {
+            executor
+                .spawn(async move {
+                    let response = match result {
+                        Ok(response) => match serde_json::from_str(&response) {
+                            Ok(deserialized) => Ok(deserialized),
+                            Err(error) => {
+                                log::error!(
+                                    "failed to deserialize response: {}. response: {:?}",
+                                    error,
+                                    response
+                                );
+                                Err(error).context("failed to deserialize response")
+                            }
+                        },
+                        Err(error) => Err(anyhow!("{}", error.message)),
+                    };
+                    _ = tx.send(response);
+                })
+                .detach();
+        })
+    }
+
     /// Sends a custom LSP request with a dynamic method name.
-    /// This is useful for extensions that need to send non-standard LSP requests.
-    ///
-    /// The `method` parameter accepts any type that can be converted into a `String`,
-    /// avoiding unnecessary allocations when the caller already has a `String`.
     pub fn request_custom<M, P, R>(&self, method: M, params: P) -> impl LspRequestFuture<R>
     where
         M: Into<String>,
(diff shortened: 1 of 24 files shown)


oscarvarto:
@maxdeviant Sorry, I won't happen again. I've been working on this for a long time. I had even closed my first attempt.

Thanks for your time.

sigmaSd:
Thanks for your work @oscarvarto , unfortunately I can't compile zed on my pc, and gitpod is not available anymore and codespaces is too slow only 2 cpus and limited drive size
Github action is the way for people with old pc, but I doubt I have time to set it up

If you want to test your self it shouldn't be that hard, you just install the deno extension, 

create a a.ts file with content: "Deno.readTextFileSync" , press "gd" on top of readTextFileSync and it should go to the virtual document showing the definition

astahjmo:
I've been using this branch  along with the zed-java extension, and the virtual documents feature is working correctly for me. Go to Definition on JDK classes and external dependencies opens the decompiled source as expected.
Happy to help test further if needed. @sigmaSd  @oscarvarto

oscarvarto:
@maxdeviant Do you thing having better support for JVM languages could help increase Zed adoption?

playdohface:
I'm sad to see this closed without any explanation/replacement. Is there something I've missed?

crates/lsp/src/lsp.rs
```diff
@@ -2046,6 +2124,53 @@ impl FakeLanguageServer {
...
+    /// Send URI and position as `TextDocumentPositionParams`.
+    ///
+    /// Used by language servers that need to know where in the document the request
+    /// originated (e.g., for macro expansions in rust-analyzer).
+    /// The request parameters will be:
+    /// `{ "textDocument": { "uri": "<document-uri>" }, "position": { "line": N, "character": M } }`
+    ///
+    /// Note: If position is not available, falls back to position (0, 0).
+    UriWithPosition,
+}
```
osiewicz: This is a bit inflexible in that if there ever is a language server that requires a set of arguments this does not support, we'll have to bump an extension API to support it's custom requests.

I'm not sure what exactly would be better (perhaps a callback-based approach?), but we should not have to bake assumptions about the structure of the requests into the core of Zed for this to work.

ghost: @osiewicz:
Thanks for your sharp eye: I agree that it shouldn't be necessary to bump an extension API to support custom requests.
Addressed the issue in https://github.com/zed-industries/zed/pull/45196/changes/d79ed39b032e4f0c78d81acd7fc2dda708ab831e
I replaced the enum with a `VirtualDocumentParamBuilder` trait that allows extensions to define custom parameter builders. This works OK with native rust extensions, but I think it would require additional changes for WASM extensions.

prashanth057:
This is such an important feature for Java development and its quite sad to see this PR languishing since many months without any progress. 
@ZedMaintainers, please prioritise this feature!

osiewicz:
I have concerns about this PR, specifically how it doesn't really solve an arbitrary-virtual-document support. See: https://github.com/zed-industries/zed/pull/45196/changes#diff-d9d8440e020edb9ad82a5960b9732ae3df407f182d8977ae36f327f83f2f0d5eR96-R105 Anytime a new language server comes around with a slightly-different URI needs, we'd have to bump the extension API version to accommodate for that.

This was admittedly partially addressed in https://github.com/zed-industries/zed/pull/45196#discussion_r2785159861 , though that's for native integrations only - we really need to nail the extension story though.
I am *not* outright rejecting the idea of this PR. I do *not* doubt that it works well for Java extension.. but, once we make a decision here, we need to stay committed to it. 
I will take a pass at it myself. I'm sorry for keeping y'all waiting. Will close it for now and let you know how that exploration goes this week.

playdohface:
@osiewicz fyi we have found a workaround for the java-extension: https://github.com/zed-extensions/java/pull/221
However, this relies on running a proxy to the language server and rewriting LSP requests/responses, something which we also have to do to get the debugger working. 
I think a well-considered widening of the extension API might allow pushing most of the complexity into the extensions which is where it belongs imo. It may be worth considering what it would take to make the "native" languages (Rust, Python etc.) work via the extension API, so that all languages can become first-class citizens in Zed down the road.

osiewicz:
@playdohface I'm glad to hear you're not immediately blocked on this. As for pulling all languages out into extensions: I've been a big proponent of that for quite some time (for the reasons you gave - stress the extension API and don't give preferential treatment to some languages over the others), but we've had legitimate (imho) reasons not to do that up to this point; with some languages (Rust, most notably) we've leaned quite heavily on custom requests, which is yet another thing we'd have to figure out. In general, I'd consider ourselves quite wary of extension API additions, as they're bound to stay with us forever.

PR #45197, opened 2025-12-18
Don't preallocate 600MB for GPUI profiler

sourcefrog:
Previously, the GPUI profiler allocates one CircularBuffer per thread, and `CircularBuffer<N>` always preallocates space for N entries. As a result it allocates ~20MB/thread, and on my machine about 33 threads are created at startup for a total of 600MB used.

In this PR I change it to use a VecDeque that can gradually grow up to 20MB as data is written. At least in my experiments it seems that this caps overall usage at about 21MB perhaps because only one thread writes very much usage data. 

Since this is fixed overhead for everyone running Zed it seems like a worthwhile gain.

This also folds duplicated code across platforms into the common gpui profiler.

Before:

Image 

After:

image

I got here from #35780 but I don't think this is tree-size related, it seems to be fixed overhead.

Release Notes:

- Improved: Significantly less memory used to record internal profiling information.


sourcefrog:
I think the cost of resizing it will be low: as it grows from 10MB to 20MB, the CPU needs to copy 10MB of data which takes maybe 1ms, assuming the allocation cannot grow in place, and this will be hit only once per thread. I'll run a CPU time profile.

On the other hand, arguably the large allocations won't have too much of an effect on user experience: virtual memory is allocated but as it's not touched or zeroed it shouldn't turn into physical memory allocations. However because it dominates the heap profile, it also makes it harder to see where memory actually is being wasted.

143mailliw:
I would love to see this merged. I ran into this when profiling Hummingbird and I think it's rather ridiculous to use this much memory for this.

sourcefrog:
@MrSubidubi I see you pushing other stuff: thanks, but let me double check the semantics of my changes before you merge it, because this is mostly not tested.

Veykril (approved):
Thanks!

Veykril:
needs a rebase for CI to become happy

Issue #45092, opened 2025-12-17
Zed opens new files and projects in default-sized window


eddiezato:
### Reproduction steps

1. Open a new file or project in Zed.

### Current vs. Expected behavior

Previously, Zed didn't remember the window size for files without projects. But there was a workaround: simply open one project, adjust the window size, and that size would be used for all new opened files and projects.

It appears that this behavior has now been changed.

Now Zed remembers sizes of individual files but for every new file or project opens it in a default-sized window. This might not be a problem for those working with a couple of projects, but for me it's a real pain, as I frequently open new individual files and now have to adjust the window size for each one of them.

I think Zed is the only editor I've ever used that doesn't let you set the window size once and then use it every time.

https://github.com/user-attachments/assets/8394ddc2-09a3-40a8-b2a4-74a88d394e88

### Zed version and system specs

Zed Nightly 0.219.0 
6f5da5e34e85654a3c5d1328b8508d83fed91535

Windows 11 Pro 25H2
OS build	26200.7462

emamoah:
+1
In *my* case, the default window size is larger than my screen resolution (1366 x 768)

https://github.com/user-attachments/assets/244b0690-cbf8-4be4-a290-b5146af0c1e3

Version: Zed stable 0.217.1
OS: Arch Linux
DE: KDE Plasma 6.5.4

PR #45201, opened 2025-12-18
gpui: Persist window bounds and display when detaching a workspace session

### LINKED ISSUE #41246: Zed does not restore window size/position (without Project/File open)
### Summary

All apps I use on a daily basis remember the main window size they had before closing, and reopen with that size, but not Zed.

### Description

A majority of apps out there, and especially text editors, remember the window size they had before they were last closed. When you reopen the app, it uses that size.

Zed has its own opinions on that topic. Every time I open Zed, the main window is in this weird state where it takes over the whole screen but it's not quite maximized. I could maximize it, but that does not change the size - it's fake maximization.

But that is not how I use this app. I always keep the app at around 25% of the surface area of the screen. That works really well. But no, Zed believes it knows better, and overrides my choice every time.

That's not even the worst part. The worst part is, Zed is incredibly hard to resize. The width of the area that I could grab with the mouse to gingerly resize the window from one side is **literally just one pixel**. Makes you wanna throw the monitor out the window.

This does not happen just on Windows. This is just where the behavior is most evident. Zed makes some kind of bad decisions when it reopens, with regards to window size, on every platform. That part of code needs to be rethought.

This is not [the older issue](https://github.com/zed-industries/zed/issues/40039) where Zed opened in such a way that the window could not be moved. That has been fixed. But the fix is pretty bad.

Steps to reproduce:
1. Install Zed on Windows 11.
2. Make sure the screen is configured at 1920x1080 with a scale factor of 125%.
3. Resize Zed such that its surface area is about 25% of the screen.
4. Close Zed.
5.  Open Zed again.

**Expected Behavior**:

Zed should open with the last used window size and position. Only when those cannot happen again (different screen plugged in) then some default should apply. But the default should be reasonable. Do not fake-maximize it every single time, that makes no sense.

And increase the width of the area where I can grab it with the mouse to resize. The current setting is super-frustrating, especially on top of the current behavior.

**Actual Behavior**:

It opens in a fake-maximized state every time, and it's super-hard to resize.

Zed: v0.209.6 (Zed) 
OS: Windows 10.0.26200
Memory: 31.8 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 630 || Intel Corporation || 25.20.100.6617

tacshi:
Closes #41246 #45092

Release Notes:

- N/A

**Root Cause**:
Empty local workspaces returned `DetachFromSession` from 
`serialize_workspace_location()`, and the `DetachFromSession` handler only cleared the session_id **without saving window bounds**.

**Fix Applied**:
Modified the `DetachFromSession` handler to save window bounds via `set_window_open_status()`:
```rust
WorkspaceLocation::DetachFromSession => {
    let window_bounds = SerializedWindowBounds(window.window_bounds());
    let display = window.display(cx).and_then(|d| d.uuid().ok());
    window.spawn(cx, async move |_| {
        persistence::DB
            .set_window_open_status(database_id, window_bounds, display.unwrap_or_default())
            .await.log_err();
        persistence::DB.set_session_id(database_id, None).await.log_err();
    })
}
```

**Recording**:

https://github.com/user-attachments/assets/2b6564d4-4e1b-40fe-943b-147296340aa7

### DIFF
--- a/crates/workspace/src/persistence.rs
+++ b/crates/workspace/src/persistence.rs
@@ -3296,4 +3296,53 @@ mod tests {
 
         assert_eq!(workspace.center_group, new_workspace.center_group);
     }
+
+    #[gpui::test]
+    async fn test_empty_workspace_window_bounds() {
+        zlog::init_test();
+
+        let db = WorkspaceDb::open_test_db("test_empty_workspace_window_bounds").await;
+        let id = db.next_id().await.unwrap();
+
+        // Create a workspace with empty paths (empty workspace)
+        let empty_paths: &[&str] = &[];
+        let display_uuid = Uuid::new_v4();
+        let window_bounds = SerializedWindowBounds(WindowBounds::Windowed(Bounds {
+            origin: point(px(100.0), px(200.0)),
+            size: size(px(800.0), px(600.0)),
+        }));
+
+        let workspace = SerializedWorkspace {
+            id,
+            paths: PathList::new(empty_paths),
+            location: SerializedWorkspaceLocation::Local,
+            center_group: Default::default(),
+            window_bounds: None,
+            display: None,
+            docks: Default::default(),
+            breakpoints: Default::default(),
+            centered_layout: false,
+            session_id: None,
+            window_id: None,
+            user_toolchains: Default::default(),
+        };
+
+        // Save the workspace (this creates the record with empty paths)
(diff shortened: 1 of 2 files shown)


tacshi:
@cla-bot check

Veykril (approved):
Thanks!

Issue #40737, opened 2025-10-20
Zed doesn't store window dimensions and position when closed unless a folder was loaded


ofetch:
### Summary

Window dimensions and position are not stored if Zed is closed without an active project!

### Description

Steps to trigger the problem:
1. Open Zed without any folder/project loaded
2. Change the window dimensions and position
3. Close Zed
4. Open Zed

**Expected Behavior**: Zed should remember the last window dimensions and position.
**Actual Behavior**: Zed only remembers last window dimensions and position of the last opened _project_ and not of the last usage/session

### Zed Version and System Specs

Zed: v0.208.4 (Zed) 
OS: Windows 10.0.26100
Memory: 15.9 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 630 || Intel Corporation || 31.0.101.2127

franciskafyi:
Nice catch. I can reproduce on MacOS as well. I think we just don't support that. Let me check with the team if this is something we can work on. 

Curious whether this is a common use case for you?

ofetch:
> Curious whether this is a common use case for you?

It is, often I use Zed like a notepad to make quick notes or just view/edit individual log or Json files, opening a folder/project for these use cases is not ideal. would be nice if window dimensions and position is remembered in those cases as well.

ofetch:
I should add that dimensions and position is expected to be saved regardless of what's loaded in the editor or panel. a file, folder or nothing at all, at least that's how most apps behave.

franciskafyi:
Yes, makes sense. Our designer agrees too! Problem is that the fix is not super trivial.

xtrmcdr:
> I should add that dimensions and position is expected to be saved regardless of what's loaded in the editor or panel. a file, folder or nothing at all, at least that's how most apps behave.

Yeah this is a really important feature for me too.

PR #45204, opened 2025-12-18
workspace: Fix new projects opening with default window size

### LINKED ISSUE #45092: Zed opens new files and projects in default-sized window

MostlyKIGuess:
Previously, when opening a new project (one that was never opened before), the window bounds restoration logic would fall through to GPUI's default window sizing instead of using the last known window bounds.

This change consolidates the window bounds restoration logic so that both empty workspaces and new projects use the stored default window bounds, making the behavior consistent: any new window will use the last resized window's size and position.

Closes #45092 

Release Notes:

- Fixed new files and projects opening with default window size instead of the last used window size.

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -1748,26 +1748,18 @@ impl Workspace {
                 window
             } else {
                 let window_bounds_override = window_bounds_env_override();
-                let is_empty_workspace = project_paths.is_empty();
 
                 let (window_bounds, display) = if let Some(bounds) = window_bounds_override {
                     (Some(WindowBounds::Windowed(bounds)), None)
-                } else if let Some(workspace) = serialized_workspace.as_ref() {
+                } else if let Some(workspace) = serialized_workspace.as_ref()
+                    && let Some(display) = workspace.display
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

ofetch:
@MostlyKIGuess Thank you, I wonder does this fix also work for #40737 ?

PR #45231, opened 2025-12-18
terminal: Fix pane re-entrancy panic when splitting terminal tabs

ahmed3mar:
## Summary
Fix panic "cannot update workspace::pane::Pane while it is already being updated" when dragging terminal tabs to split the pane.

## Problem
When dragging a terminal tab to create a split, the app panics due to re-entrancy: the drop handler calls `terminal_panel.center.split()` synchronously, which invokes `mark_positions()` that tries to update all panes in the group. When the pane being updated is part of the terminal panel's center group, this causes a re-entrancy panic.

## Solution
Defer the split operation using `cx.spawn_in()`, similar to how `move_item` was already deferred in the same handler. This ensures the split (and subsequent `mark_positions()` call) runs after the current pane update completes.

## Test plan
- Open terminal panel
- Create a terminal tab
- Drag the terminal tab to split the pane
- Verify no panic occurs and split works correctly

### DIFF
--- a/crates/terminal_view/src/terminal_panel.rs
+++ b/crates/terminal_view/src/terminal_panel.rs
@@ -1197,23 +1197,21 @@ pub fn new_terminal_pane(
                         // call mark_positions (via split) synchronously.
                         cx.spawn_in(window, async move |_, cx| {
                             cx.update(|window, cx| {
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #45243, opened 2025-12-18
Correct UTF-16 saving and add heuristic encoding detection

### LINKED ISSUE #14654: Files with UTF-8 BOM are not handled correctly
### Check for existing issues

- [X] Completed

Files that start with the UTF-8 byte order mark (EFBBBF) are not interpreted correctly. This can have different effects depending on the type of file. In the case of a JSON file, Zed shows an error on the first position stating "Expected a JSON object, array or literal".  In the case of a C# file, syntax highlighting is messed up on the first line.  In all cases, editing the line causes strange behavior where the characters inserted are in a different position than the cursor.

I don't remember noticing anything like this before updating Zed this morning.  I _believe_ my prior version was 142.4.

### Environment

Zed: v0.143.7 (Zed)
OS: macOS 14.5.0
Memory: 32 GiB
Architecture: x86_64

First line of C# file that starts with BOM:
image

When BOM removed:
image

_No response_

tomopumipumi:
This commit fixes an issue where saving UTF-16 files resulted in UTF-8 bytes due to `encoding_rs` default behavior. It also introduces a heuristic to detect BOM-less UTF-16 and binary files.

Changes:
- Manually implement UTF-16LE/BE encoding during file save to avoid implicit UTF-8 conversion.
- Add `analyze_byte_content` to guess UTF-16LE/BE or Binary based on null byte distribution.
- Prevent loading binary files as text by returning an error when binary content is detected.

Special thanks to @CrazyboyQCD  for pointing out the `encoding_rs` behavior and providing the fix, and to @ConradIrwin for the suggestion on the detection heuristic.

Closes #14654

Release Notes:

- (nightly only) Fixed an issue where saving files with UTF-16 encoding incorrectly wrote them as UTF-8. Also improved detection for binary files and BOM-less UTF-16.

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -1490,19 +1490,23 @@ impl Buffer {
         let (tx, rx) = futures::channel::oneshot::channel();
         let prev_version = self.text.version();
         self.reload_task = Some(cx.spawn(async move |this, cx| {
-            let Some((new_mtime, new_text)) = this.update(cx, |this, cx| {
+            let Some((new_mtime, load_bytes_task, encoding)) = this.update(cx, |this, cx| {
                 let file = this.file.as_ref()?.as_local()?;
-
-                Some((file.disk_state().mtime(), file.load(cx)))
+                Some((
+                    file.disk_state().mtime(),
+                    file.load_bytes(cx),
+                    this.encoding,
+                ))
             })?
             else {
                 return Ok(());
             };
 
-            let new_text = new_text.await?;
-            let diff = this
-                .update(cx, |this, cx| this.diff(new_text.clone(), cx))?
-                .await;
+            let bytes = load_bytes_task.await?;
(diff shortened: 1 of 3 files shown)


ConradIrwin:
Thanks, The approach looks reasonable to me, but when I try opening the test UTf-16 files, and then saving with no changes, they get new null-bytes after every save.


Screenshot 2025-12-18 at 15 06 20

ConradIrwin:
It still seems to read correctly the first time (I see no nulls) so maybe something to do with round-tripping

tomopumipumi:
@ConradIrwin Thanks for the hint. I was able to reproduce the issue on my machine. (I can't believe I missed this until now...)
 The root cause was that the encoding was not being applied in the Buffer struct's reload logic. I fixed this and confirmed that UTF-16 files now display correctly in my environment without any garbled text, even after saving and reloading.

PR #45259, opened 2025-12-18
Screenshot testing

rtfeldman:
## Screenshot testing

Adds visual testing infrastructure for GPUI that captures screenshots by rendering directly to Metal textures. The

The screenshots end up in `target/visual_tests/` and look like this:

workspace_with_editor2
project_panel2

### Key Features

- **Direct texture capture**: Screenshots are captured by rendering the scene to a Metal texture and reading pixels directly from GPU memory, rather than using ScreenCaptureKit
- **No visibility requirements**: Windows don't need to be visible on screen since we read directly from the render pipeline
- **Deterministic output**: Captures exactly what GPUI renders, not what the OS compositor displays
- **No permissions needed**: Doesn't require Screen Recording permission like ScreenCaptureKit would

### Running the Visual Tests

```bash
# Run visual tests (compares against baselines)
cargo run -p zed --bin visual_test_runner --features visual-tests

# Update baseline images (when UI intentionally changes)
UPDATE_BASELINE=1 cargo run -p zed --bin visual_test_runner --features visual-tests

# View the captured screenshots
open target/visual_tests/
```

### Implementation

- `Window::render_to_image()` - Renders the current scene to a texture and returns an `RgbaImage`
- `MetalRenderer::render_to_image()` - Core implementation that renders to a non-framebuffer-only texture
- `VisualTestAppContext` - Test context that uses real macOS platform rendering
- `VisualTestAppContext::capture_screenshot()` - Synchronous screenshot capture using direct texture rendering

### Usage

```rust
let mut cx = VisualTestAppContext::new();
let window = cx.open_window(...)?;
let screenshot: RgbaImage = cx.capture_screenshot(window.into())?;
```

Release Notes:

- N/A


mocenigo:
I think I have a problem building this.

```
Bundling Zed Dev.app
error: Failed to copy binary from ".../target/aarch64-apple-darwin/release/visual_test_runner"
Caused by: ".../target/aarch64-apple-darwin/release/visual_test_runner" does not exist
```

PR #45273, opened 2025-12-18
RFC: Parameterized Run Configurations

richard-hajek:
Hi!

This draft aims to implement a "Run Configuration" functionality to implement a new workflow

# Spec

It contains the following functionality:

* Recipes, which define “templates” of configurations (the only reason I don’t call them “templates” is not to conflate with existing task templates). A recipe should have a command to run and debug a project
* Run configurations: puts concrete values into a recipe
* UI integration of the above functionality

Recipes example:
```json
[
  {
    "name": "uv",
    "executeRun": "uv run {{ .file }}",
    "executeDebug": "uv run debugpy {{ .file }}",
    "description": "Python uv runner",
  },
]
```

Configurations example:
```json
[
  {
    "label": "Run Main",
    "recipe": "uv",
    "variables": {
      "file": "src/main.py",
    },
  },
]
```

# UI

Current UI implementation screenshot:

image

# Comment

This feature is heavily inspired by the core feature of JetBrain's IDEs, which in itself demonstrates that it isn't completely insane. 

Advantages:

* A time-tested "Run & Debug" workflow support
* Easy to explain workflow of "just press the run button" for beginners

Disadvantages:

* Competes with tasks functionality
* May be seen as a feature creep

To address the disadvantages: 

As I understand it tasks are more suited to job-like tasks (compile, clean, publish), so Run Configurations would not compete with them.

At this stage, I’m mostly interested in community feedback on whether this abstraction and UI direction are appropriate for the core project. If the approach is considered viable, I’m willing to invest the time required to iterate on the design and implementation.

Presumably a set of built-in recipes would be shipped with the editor, although it's not a huge drag to define one's own.

I am looking for answers to these questions:

* Is this feature, assuming a perfect implementation (which this is not), even a good idea?
* What would be the role of this feature with respect to the tasks feature? Should this feature be better implemented as an extension to tasks?


richard-hajek:
@cla-bot check

PR #45277, opened 2025-12-18
Add support for provider extensions (but no extensions yet)

rtfeldman:
This adds support for provider extensions but doesn't actually add any yet.

Release Notes:

- N/A


crates/language_models/src/provider/anthropic.rs
```diff
@@ -1104,6 +1104,7 @@ impl Render for ConfigurationView {
                 .on_action(cx.listener(Self::save_api_key))
                 .child(Label::new(format!("To use {}, you need to add an API key. Follow these steps:", match &self.target_agent {
                     ConfigurationViewTargetAgent::ZedAgent => "Zed's agent with Anthropic".into(),
+                    ConfigurationViewTargetAgent::EditPrediction => "Edit Predictions with Anthropic".into(),
```
mikayla-maki: ??

mikayla-maki (dismissed):
Looks good other than that one line! Thanks for breaking it up a little :D

crates/language_models/src/provider/anthropic.rs
rtfeldman: Ah, so this is because the big branch adds `ConfigurationViewTargetAgent::EditPrediction` because of some OAuth-related stuff for custom edit prediction providers. It certainly looks weird here though without that context!

I'll just take the `EditPrediction` variant out of scope this PR since it doesn't make sense in isolation, and plus OAuth will be a ways off anyway.

PR #45280, opened 2025-12-18
workspace: Add `Pane::add_item_behind` for background tab creation

### LINKED ISSUE #45279: Pane::add_item lacks API for adding items without tab activation
## Summary

`Pane::add_item` always switches to the newly added tab, with no way to add an item "in the background" without activating it. This makes it difficult to implement features like `RevealStrategy::Never` correctly.

## Current Behavior

`Pane::add_item` calls `add_item_inner` with the `activate` parameter hardcoded to `true`:

```rust
pub fn add_item(...) {
    self.add_item_inner(
        item,
        activate_pane,
        focus_item,
        true,  // <-- always activates the tab
        ...
    )
}
```

The `activate_pane` parameter controls whether the *pane* becomes active, but tab activation is not controllable by callers.

## Expected Behavior

Callers should have an explicit API to add items without switching to them. This is needed for:

- Programmatic terminal creation that shouldn't disrupt the user's current tab
- Background task terminals (`RevealStrategy::Never`)
- Extension-based workflows that create tabs silently

## Proposed Solution

Add a companion method `add_item_behind` that adds items without tab activation:

```rust
pub fn add_item_behind(...) {
    self.add_item_inner(
        item,
        activate_pane,
        focus_item,
        false,  // <-- does not activate the tab
        ...
    )
}
```

This keeps `add_item` unchanged (backwards compatible) while providing an explicit API for the background case.

## Related

This is related to #45179, which identified focus behavior issues with `RevealStrategy`. Properly implementing `RevealStrategy::Never` requires both the focus fix and an API to add items without tab activation.

rabsef-bicrym:
## Summary

Adds `Pane::add_item_behind`, a companion to `add_item` that inserts an item without switching to its tab.

- `add_item` - adds item and switches to it (unchanged)
- `add_item_behind` - adds item without switching to it (new)

Closes #45279

## Motivation

The existing `add_item` method hardcodes tab activation to `true`, making it impossible for callers to add items in the background. This is needed for:

- Programmatic terminal creation via extensions
- Background task terminals (`RevealStrategy::Never`)
- Any workflow that needs to create tabs without disrupting the user's current focus

## Changes

- Added `Pane::add_item_behind` method in `crates/workspace/src/pane.rs`
- No changes to existing `add_item` behavior (backwards compatible)

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -1226,6 +1226,35 @@ impl Pane {
         )
     }
 
+    /// Adds an item to the pane without switching to its tab.
+    ///
+    /// Unlike `add_item`, this method adds the item in the background without
+    /// making it the active tab. The `activate_pane` parameter still controls
+    /// whether the pane itself becomes active.
+    pub fn add_item_behind(
+        &mut self,
+        item: Box<dyn ItemHandle>,
+        activate_pane: bool,
+        focus_item: bool,
+        destination_index: Option<usize>,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        if let Some(text) = item.telemetry_event_text(cx) {
+            telemetry::event!(text);
+        }
+
+        self.add_item_inner(
+            item,
+            activate_pane,
+            focus_item,
+            false,
+            destination_index,
+            window,
+            cx,
+        )
+    }
+
     pub fn items_len(&self) -> usize {
         self.items.len()
     }


rabsef-bicrym:
> Thanks for the contribution, unfortunately without usage code, I can't test that it works and I don't see the benefit in merging code that is just unused, so I'm going to close this.
> 
> If you'd like to expand on your idea and add an action for it, or actual usage I may be more inclined to re-open and merge. Thanks!

I'm going to be submitting a PR for a cli to programmatically add/remove/move/etc tabs and panes in the Terminal workspace soon - will re-contribute this then.

PR #45285, opened 2025-12-18
git: Clean up integration settings schema

Anthony-Eid:
Follow up on: 43326

Integrate git gutter diff hunks with the global git integration setting, which was previously missing.

I also simplify the settings schema by replacing three boolean fields (disable_git, enable_status, enable_diff) with a single enum: full, status_only, diff_only, or disabled.

Release Notes:

- git: Add git integration setting that allows disabling git support in Zed
- `git: { integration: "full" | "status_only" | "diff_only" | "disabled"}`


osiewicz:
'zup

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been inactive for quite some time and has conflicts now.

PR #45289, opened 2025-12-18
Add zed::PopKey action

P1n3appl3:
@xipeng-jin and I thought this would be a nice followup to the which-key PR. We already have the `vim_mode == waiting` context, but I decided to handle this at a more general level since it should work for non-vim sequential keybindings.

Release Notes:

- Add `zed::PopKey` action to remove the latest pending keystroke. This allows you to navigate backwards in our new which-key UI.

### DIFF
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -1097,4 +1097,10 @@
       "ctrl-e": "markdown::ScrollDown",
     },
   },
+  {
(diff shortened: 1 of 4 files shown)


xipeng-jin:
Hi @P1n3appl3 , this is very cool! And this action will be the essential part for enabling us to go backward in which-key menu. I will take a look at this

PR #45291, opened 2025-12-18
Add GPT-5.2-codex support

TomBayne:
Release Notes:

- Adds GPT-5.2-codex model support.


TomBayne:
@cla-bot check

bennetbo:
Thank you. If I understand correctly it's not available via the API yet? And even if it is available, it won't work in Zed because we need support for the new responses API first (see #39989).

PR #45300, opened 2025-12-18
Allow project search to be efficiently cancelled

maxbrunsfeld:
Previously, cancelling did not really terminate the search computations, due to detaching the owning Task, and due to synchronous IO that was used to find candidate files.

Release Notes:

- Optimized cancellation of project search. Now, when closing a project search or changing the query while the search is running, work on the previous search will terminate more promptly.


preechapon250 (commented):
Fixes #45621

PR #45313, opened 2025-12-18
Enforce 5MB per-image limit when converting images for language models

nathansobo:
## Problem

When users paste or drag large images into the agent panel, the encoded payload can exceed upstream provider limits (e.g., Anthropic's 5MB per-image limit), causing API errors.

## Solution

Enforce a default 5MB limit on encoded PNG bytes in `LanguageModelImage::from_image`:

1. Apply existing Anthropic dimension limits first (1568px max in either dimension)
2. Iteratively downscale by ~15% per pass until the encoded PNG is under 5MB
3. Return `None` if the image can't be shrunk within 8 passes (fail-safe)

The limit is enforced at the `LanguageModelImage` conversion layer, which is the choke point for all image ingestion paths (agent panel paste/drag, file mentions, text threads, etc.).

## Future Work

The 5MB limit is a conservative default. Provider-specific limits can be introduced later by adding a `from_image_with_constraints` API.

## Testing

Added a regression test that:
1. Generates a noisy 4096x4096 PNG (guaranteed >5MB)
2. Converts it via `LanguageModelImage::from_image`
3. Asserts the result is ≤5MB and was actually downscaled

---

**Note:** This PR builds on #45312 (prompt store fail-open fix). Please merge that first.

cc @rtfeldman

### DIFF
--- a/crates/language_model/src/request.rs
+++ b/crates/language_model/src/request.rs
@@ -8,6 +8,7 @@ use gpui::{
     App, AppContext as _, DevicePixels, Image, ImageFormat, ObjectFit, SharedString, Size, Task,
     point, px, size,
 };
+use image::GenericImageView as _;
 use image::codecs::png::PngEncoder;
 use serde::{Deserialize, Serialize};
 use util::ResultExt;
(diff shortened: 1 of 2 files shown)


rtfeldman (approved):
Love it! 😍

PR #45321, opened 2025-12-19
nix: Use flake-parts, partitions, and treefmt-nix

doprz:
Hi @zed-industries team! Thank you for building such an amazing editor. `Zed` has become my daily driver and a key part of my development workflow. I really like the performance, vim-support, and collaboration features. Excited to make my first contribution! :rocket: 

Release Notes:

Refactored:
- Use [flake-parts](https://flake.parts/index.html) modules
- `nix/shell.nix` -> `nix/modules/devshells.nix`

Added:
- Use [flake-parts.partitions](https://flake.parts/options/flake-parts-partitions.html) to isolate dev dependencies so that flakes that use `zed-editor` don't fetch dev-only inputs such as `treefmt-nix`
- [treefmt-nix](https://github.com/numtide/treefmt-nix)
  - nixfmt
  - rustfmt

Fixed:
- `shell.nix` and `default.nix` fetching `flake-compat` from `flake.lock` which added an extra and unnecessary input/dependency to `flake.nix`. Fixed by setting a fixed rev and sha256 instead.
- `nixfmt-rfc-style` is deprecated and is now `nixfmt`
- Fixes #45338 by using rust-overlay toolchain directly
  - Previously, the devShell included `rustup` which caused slow startup
times as Nix would build rustup from source (including running its test
suite). Additionally, rust tooling (cargo, rustfmt, clippy, rust-analyzer) wasn't available in the dev shell.
  - cargo, rustc, and rust-toolchain.toml components included in `rustToolchain`

Chore:
- Update `flake.lock`
- Format Rust code with `rustfmt` via `treefmt-nix`


doprz:
@cla-bot check

doprz:
There seems to be a mismatch between `cargo fmt --all -- --check` and `nix fmt`. Checking to see what the issue is.

doprz:
I'm also considering using [crate2nix](https://github.com/nix-community/crate2nix) rather than [crane](https://crane.dev/) as it can:
> save time by only rebuilding changed crates hermetically...

However, it uses [import from derivation](https://nixos.org/manual/nix/stable/language/import-from-derivation) (IFD) which can disable build parallelism. Open to thoughts on this issue. I think it's out of scope for this PR but something we should revisit in the future.

doprz:
> Sorry for the delay, unfortunately I no longer help maintain Zed. @kubkon should be able to review this.

Thank you for the update!

jrobsonchase:
I think this broke my `rust-analyzer` and its ability to locate the `RUST_SRC`. Previously, I don't believe `rust-analyzer` was getting installed by the flake, and was "doing its own thing" as far as the rust src was concerned. Not 100% correct, but "good enough."

Now though, it's failing to find it and giving me 400+ errors in weird places like the editor tests where it's resolving the `update` method to the `Itertools` trait rather than a concrete one. Probably due to a missing `RUST_SRC` breaking proc macro analysis.

PR #45331, opened 2025-12-19
Add actions to move to start and end of larger syntax node

vamsi10010:
Release Notes:

- Added two actions `move_to_start_of_larger_syntax_node` and `move_to_end_of_larger_syntax_node` that move cursors to the start or end of the parent tree-sitter node

Following up on my PR #41321, this PR only adds the actions that are used to enable code navigation across syntax nodes, without binding them to any keys (such as tab) by default. Both actions use the tree-sitter syntax tree to find parent nodes of the nodes the cursors are currently in. `move_to_start_of_larger_syntax_node` will then move each cursor to the first position of the parent nodes while `move_to_end_of_larger_syntax_node` to a position right after the parent nodes.

Related issues and discussions: #22349, #14803, #42828, #13736.

This PR doesn't achieve "tab out" functionality in the exact sense as is requested in these issues as it does not bind the actions to the tab key. I hope this PR can start some discussion on what the best way forward for these issues is. In the meantime, users can configure keys to use these actions as they see fit to emulate "tab out" behavior. For example,
```
"context": "Editor && vim_mode == insert && !in_snippet && !showing_completions",
"bindings": {
  "tab": "editor::MoveToEndOfLargerSyntaxNode",
  "shift-tab": "editor::Tab"
}
```
This will enable tab to skip past code structures like brackets when the cursor is not in a snippet or the autocomplete menu is not open. At the same time, shift tab will act as a backup tab.

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -638,6 +638,10 @@ actions!(
         MoveToEndOfExcerpt,
         /// Moves cursor to the end of the previous excerpt.
         MoveToEndOfPreviousExcerpt,
+        /// Moves cursor to the start of the next larger syntax node.
+        MoveToStartOfLargerSyntaxNode,
+        /// Moves cursor to the end of the next larger syntax node.
+        MoveToEndOfLargerSyntaxNode,
         /// Moves cursor up.
         MoveUp,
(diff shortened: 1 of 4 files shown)


tonyrav3n:
Thanks!

PR #45332, opened 2025-12-19
vim: Add `:r[ead] [name]` command

AidanV:
This adds the following Vim commands: 
- `:r[ead] [name]`
- `:{range}r[ead] [name]`

The most important parts of this feature are outlined [here](https://vimhelp.org/insert.txt.html#%3Ar).

The only intentional difference between this and Vim is that Vim only allows `:read` (no filename) for buffers with a file attached. I am allowing it for all buffers because I think that could be useful. 

Release Notes:

- vim: Added the [`:r[ead] [name]` Vim command](https://vimhelp.org/insert.txt.html#:read)

### DIFF
--- a/crates/vim/src/command.rs
+++ b/crates/vim/src/command.rs
@@ -701,8 +701,9 @@ pub fn register(editor: &mut Editor, cx: &mut Context<Vim>) {
                         let Some(worktree) = project.visible_worktrees(cx).next() else {
                             return;
                         };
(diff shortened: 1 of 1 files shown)


probably-neb:
This is cool! Nice work!

I pushed a commit to add some logging just so the common failure cases (path invalid, file failed to load) are logged instead of ignored. Will merge once CI is green!

PR #45347, opened 2025-12-19
language: Fix reparse timeout not actually working

Veykril:
We use `block_with_timeout` to give the reparse task a millisecond to complete, and if it takes longer we put the work off to the background.
The reason for this is that we want tree-sitter based features to feel snappy.
The reparse task is non-cooperative though, it has no yield points, giving us no place to actually check for our timeout, meaning we will always drive it to completion and block for the entire duration.
This kicks out the `block_with_timeout` in favor of using the treesitter progress callback to handle timeouts instead.

Release Notes:

- Improved responsiveness with very large language files on edits


Veykril:
oops, one of those was meant to be large

PR #45353, opened 2025-12-19
Introduce Responses API for codex models & fix glob panic

tsoernes:
This pull request introduces support for the OpenAI Responses API when using codex models (including compatible providers such as Azure OpenAI), and also includes a fix to avoid a panic on empty glob patterns during project file searches.

### Changes in this PR

1. **Responses API Integration**  
   - Stream and map `responses/v1` events for codex models  
   - Properly handle function calls, errors, and completion events  
   - Supports both built-in OpenAI and openai-compatible providers

2. **Glob Panic Fix**  
   - Added defensive check to skip empty or whitespace-only glob patterns  
   - Prevents `wax::Glob::partition()` from panicking when given an empty pattern

### Verification Steps

- Rebuilt the editor with these changes and confirmed that codex-based tool calls now correctly invoke functions instead of timing out.  
- Tested file search tools (fuzzy finder, include patterns) with blank or empty filters to ensure no panics occur.

Please review the two commits for more details on implementation.

### DIFF
--- a/crates/project/src/project_search.rs
+++ b/crates/project/src/project_search.rs
@@ -790,12 +790,13 @@ impl PathInclusionMatcher {
         // For example, `src/**/*.rs` becomes `src/` and `**/*.rs`. The glob part gets dropped.
         // Then, when checking whether a given directory should be scanned, we check whether it is a non-empty substring of any glob prefix.
         if query.filters_path() {
-            included.extend(
-                query
-                    .files_to_include()
-                    .sources()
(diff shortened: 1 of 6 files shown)


crates/project/src/project_search.rs
maxdeviant, line 1: Can we move this change into a separate PR? It seems unrelated to the Codex changes.

tsoernes: IDK. When using the openai codex that seem to trigger the bug. But same for me.

PR #45363, opened 2025-12-19
editor: Fix the merging of adjacent selection edits

### LINKED ISSUE #45046: Adjacent selection merge upon insertion of text
### Reproduction steps

1. Type `AAA`
2. Highlight `A`
3. Duplicate selection
4. Type

Notice selections were merged

Note: The merging of selections (as selections are being added) was fixed in https://github.com/zed-industries/zed/pull/44811, this is about selections being merged on text insertion

Current (Zed): 

https://github.com/user-attachments/assets/0866e167-f838-443b-b315-50f85601ccb3

Expected (VS Code): 

https://github.com/user-attachments/assets/5d14b302-dbd7-48c8-9076-f1d7fcfe9fc8

Expected (Sublime Text):

https://github.com/user-attachments/assets/3f00e075-7525-411f-a7a2-d77582f46143
 

Zed: v0.219.0+nightly.2017.c1317baebebf955c589ef19d8c209b2ef526f75f (Zed Nightly c1317baebebf955c589ef19d8c209b2ef526f75f) 
OS: macOS 26.1
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

marcocondrache:
Closes #45046

The root of the issue is anchor resolution. When we apply adjacent edits, they get merged into a single edit. In the scenario described in the issue, this is what happens:

1. We create an anchor at the end of each selection (bias::right) on the snapshot before the edits.
2. We collect the edits and apply them to the buffer.
3. Since the edits are adjacent (>=), the buffer merges them into a single edit.
4. As a result, we apply one edit to the text buffer, creating a single visible fragment with length = 3.
5. The buffer ends up with fragments like: [F(len = 3, visible = true), F(len = 1, visible = false), ...]
6. After the edits, we resolve the previously created anchors to produce zero-width selections (cursors).
7. All anchors resolve into deleted fragments, so their resolved offset equals the cumulative visible offset, which is 3.
8. We now have 3 cursors with identical coordinates (0;3).
9. These cursors get merged into a single cursor.

I tried several approaches, but they either felt wrong or didn’t work. In particular, I tried adjusting anchor resolution using the delta stored in handle_input, but this doesn’t help because selections are merged immediately after anchor resolution.

The only workable solution I found is to avoid anchors entirely for the adjacent-edit case. Instead, we can compute the final cursor positions directly from the edits and create the selections based on that information.

Release Notes:

- Fixed an issue where adjacent selection insert would merge cursors

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -3674,7 +3674,7 @@ impl Editor {
         }
 
         self.buffer
-            .update(cx, |buffer, cx| buffer.edit(edits, None, cx));
+            .update(cx, |buffer, cx| buffer.edit(edits, None, true, cx));
     }
 
     pub fn edit_with_autoindent<I, S, T>(&mut self, edits: I, cx: &mut Context<Self>)
@@ -3688,7 +3688,7 @@ impl Editor {
         }
 
         self.buffer.update(cx, |buffer, cx| {
-            buffer.edit(edits, self.autoindent_mode.clone(), cx)
+            buffer.edit(edits, self.autoindent_mode.clone(), true, cx)
         });
     }
 
@@ -3712,6 +3712,7 @@ impl Editor {
                 Some(AutoindentMode::Block {
                     original_indent_columns,
                 }),
+                true,
                 cx,
             )
         });
@@ -4322,6 +4323,7 @@ impl Editor {
         let mut clear_linked_edit_ranges = false;
         let mut all_selections_read_only = true;
         let mut in_adjacent_group = false;
+        let mut contains_ajacent_edits = false;
 
         let mut regions = self
             .selections_with_autoclose_regions(selections, &snapshot)
@@ -4615,6 +4617,10 @@ impl Editor {
                 }
             }
 
+            if next_is_adjacent {
+                contains_ajacent_edits = true;
+            }
+
             new_selections.push((selection.map(|_| anchor), 0));
             edits.push((selection.start..selection.end, text.clone()));
         }
@@ -4636,7 +4642,12 @@ impl Editor {
                 jsx_tag_auto_close::construct_initial_buffer_versions_map(this, &edits, cx);
 
             this.buffer.update(cx, |buffer, cx| {
-                buffer.edit(edits, this.autoindent_mode.clone(), cx);
+                buffer.edit(
+                    edits,
+                    this.autoindent_mode.clone(),
+                    !contains_ajacent_edits,
+                    cx,
+                );
             });
             for (buffer, edits) in linked_edits {
                 buffer.update(cx, |buffer, cx| {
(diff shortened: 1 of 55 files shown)


Veykril:
Hmm, could we just have a buffer edit variant here that disallows merging the passed in edits for this case?

marcocondrache:
@Veykril I also tried disabling the merge in the edit method to see if that would solve the issue, but it didn’t. The fragments still ended up being merged into a single one. I’d actually prefer this solution as well, so I’ll take another look and see if I can get it to work

crates/language/src/buffer.rs
```diff
@@ -2553,7 +2553,7 @@ impl Buffer {
             let new_text = new_text.into();
             if !new_text.is_empty() || !range.is_empty() {
                 if let Some((prev_range, prev_text)) = edits.last_mut()
-                    && prev_range.end >= range.start
+                    && prev_range.end > range.start
```
marcocondrache: @Veykril got this working with anchors, and it looks a lot better. I had to disable merging for all edits to make it work. The change to disable adjacent edit merging is tiny, so adding a separate variant feels a bit overkill. Maybe a parameter to toggle adjacent edit merging on the buffers instead?

Veykril: Yea parameter is fine as well

crates/multi_buffer/src/multi_buffer.rs
Veykril, line 1251: Okay thats a lot of call sites, lets change plans slightly here. Let's make this function (with the arg) private and rename it something else. Then have two methods `edit` and `edit_non_coalesce` that just calls out to the private one but the bool parameter being different. That way we will only need to touch that one (or maybe couple) callsite(s). That way we also dont obscure this behavior too much behind a random non-descriptive bool

marcocondrache: Yes, this makes a lot of sense. I didn't realize how many there were until I checked the changes on the PR

marcocondrache: Should be done

crates/editor/src/editor.rs
```diff
@@ -4592,23 +4608,36 @@ impl Editor {
 
             new_selections.push((selection.map(|_| anchor), 0));
             edits.push((selection.start..selection.end, text.clone()));
+
+            if next_is_adjacent {
+                has_adjacent_edits = true;
+            }
+
```
Veykril:
```suggestion
            has_adjacent_edits |= next_is_adjacent;
```

Veykril (approved):
Thanks!

PR #45365, opened 2025-12-19
lsp: Refresh `textDocument/diagnostic` on `workspace/diagnostic/refresh`

aviatesk:
Per LSP spec, when receiving `workspace/diagnostic/refresh`, clients
should refresh all pulled diagnostics including both workspace and
document diagnostics[^1]. Previously, only workspace diagnostics were
refreshed.

[^1]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#diagnostic_refresh

This adds `pull_document_diagnostics_for_server()`, which refreshes
`textDocument/diagnostic` for all open buffers associated with the
given language server.

Closes #ISSUE

Release Notes:

- When handling `workspace/diagnostic/refresh`, Zed now also sends new `textDocument/diagnostic` requests for open files, aligning with the LSP specification.


aviatesk:
Thanks for the reply. Ok, then I'll ask @reflectronic to review this for now. @reflectronic, it would be great if you could review it whenever you get a chance. Thanks.

reflectronic:
Thank you for the pull request. I made some edits:
- The tests were asserting that `workspace/diagnostic/refresh` does not cause document pulls, which needed to be fixed
- Some similar logic was added in the meantime for a different purpose, so I consolidated that

reflectronic (approved):

SomeoneToIgnore:
A small post-review: https://github.com/zed-industries/zed/pull/46383

PR #45376, opened 2025-12-19
docs: Update Java docs to reference OpenJDK 25

ForrestKnight:
This brings the Java documentation up to date with current OpenJDK releases.

- Switches the manual download link to OpenJDK 25
- Updates Ubuntu install instructions to use native packages instead of a PPA

Release Notes:

- N/A

### DIFF
--- a/docs/src/languages/java.md
+++ b/docs/src/languages/java.md
@@ -11,11 +11,12 @@ Java language support in Zed is provided by:
 You will need to install a Java runtime (OpenJDK).
 
 - macOS: `brew install openjdk`
-- Ubuntu: `sudo add-apt-repository ppa:openjdk-23 && sudo apt-get install openjdk-23`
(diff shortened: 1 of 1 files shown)


playdohface (commented):
While technically there is nothing wrong here, it strikes me as a little odd. Why use Java 21 on on older versions of Ubuntu and 25 on newer ones? 
If we are updating this section of the docs, it might be good to include the fact that now we don't even have to manually download a JDK anymore, at least not one to run JDTLS, have a look at the [Readme of the Java extension](https://github.com/zed-extensions/java/blob/main/README.md)
```
The option to let the extension automatically download a version of OpenJDK can be enabled by setting jdk_auto_download to true. When enabled, the extension will only download a JDK if no valid java_home is provided or if the specified one does not meet the minimum version requirement. User-provided JDKs always take precedence.
```

PR #45387, opened 2025-12-19
Beam Jump - Lightning Fast Vim style navigation

### LINKED ISSUE #4930: easymotion vim
### Check for existing issues

- [X] Completed

### Describe the feature

In VSCode's vim emulation you have an easymotion mode that allows you to type in 

`<leader><leader>motion`

Then instead of doing that motion regularly, it will pull up easymotion which works like this

https://www.loom.com/share/65f7ff4745df4f85bc1cd3ca51afdf83

### LINKED ISSUE #14801: Flash.vim style search in a document.
### Check for existing issues

- [X] Completed

### Describe the feature

[Flash](https://github.com/folke/flash.nvim) lets you search and jump to Text inside a file by appending jump labels as you type search text. 

it is an incredibly fast way to jump the cursor to specific places and would be an excellent addition to vim mode in Zed.

_No response_

xipeng-jin:
Closes #14801
Closes #4930

### Summary

Beam Jump is a **viewport-first, directionless** “type-to-target” jump mode for Zed’s Vim mode that supports **arbitrary-length patterns** with **low-noise visuals** and **no surprising navigation while typing**.

### How to use

- Press `s` to enter Beam Jump (cursor stays put; mode is directionless).
- Type a pattern (any length):

  - **len == 1:** show hollow outlines for eligible 1-char matches in the **viewport** (no labels, no jump).
  - **len ≥ 2:**

    - `V > 1` (multiple viewport matches): show hollow outlines + **1 or 2 char labels**; type a label to jump.
    - `V == 1`: show a **single outline** (no label); press `Enter` to commit the jump.
    - `V == 0`: press **Enter** to run a **global forward find with wrap** (the only global navigation trigger).
    
- Cancel anytime with `Esc` / `Backspace` / `Tab` (consumed; no Vim action pass through).

### Key invariant / guarantees

- **Viewport-only matching while typing**; no implicit global navigation.
- **Cross-cursor matches are excluded**; for `len ≥ 2`, viewport candidates are **non-overlapping** (canonical set used for labels + jumps).
- Label keys are **collision-safe** (characters that could extend the pattern are excluded); `;` / `,` are **pattern characters** while active and are **never** label keys.
- **Non-text keys cancel + consume** (no accidental Vim command execution).
- After a successful jump, the full pattern is persisted for Vim repeat (`;` / `,` outside Beam Jump) behavior.

### Safety / UX

* If `pattern_len ≥ 1` and viewport has **zero matches**, an **idle-cancel timer (1.5s)** cleans up the session.
* If viewport bounds can’t be computed at entry, treat viewport matches as **always empty** (no whole-buffer fallback); `Enter` at `len ≥ 2` still performs the global forward find (wrap).

### Release Notes

- Added optional **Beam Jump** mode for Vim **Sneak** (`s`): **arbitrary-length pattern** input with **viewport-first** matching and minimal visual overlays.
- Introduced **labeled jumps** (1–2 characters) for multiple viewport matches; labels are designed to avoid conflicts with continued typing.
- Added **explicit commit behavior**: when a unique viewport match exists, press **`Enter`** to jump (no auto-jump while typing).
- Added **repeatable global navigation** via **Enter-triggered forward find with wrap** when there are no viewport matches (pattern length ≥ 2).
- Improved safety/cancellation: **Esc/Backspace/Tab/non-text keys cancel and are consumed**, preventing accidental Vim command passthrough during an active session.
- Added **idle auto-cancel (1.5s)** when no viewport matches exist, preventing lingering “ghost” sessions.

### DIFF
--- a/crates/vim/src/beam_jump.rs
+++ b/crates/vim/src/beam_jump.rs
@@ -36,10 +36,10 @@ pub(crate) struct BeamJumpLabels {
 
 #[derive(Clone, Debug)]
 pub(crate) struct BeamJumpState {
-    pub(crate) direction: BeamJumpDirection,
     pub(crate) smartcase: bool,
-    pub(crate) search_start: MultiBufferOffset,
-    pub(crate) search_end: MultiBufferOffset,
+    pub(crate) cursor_offset: MultiBufferOffset,
+    pub(crate) view_start: MultiBufferOffset,
+    pub(crate) view_end: MultiBufferOffset,
     pub(crate) previous_last_find: Option<Motion>,
 
     pub(crate) pattern: String,
@@ -55,7 +55,7 @@ pub(crate) struct BeamJumpJump {
     pub(crate) pattern: String,
     pub(crate) smartcase: bool,
     pub(crate) count: usize,
-    pub(crate) search_range: Range<MultiBufferOffset>,
+    pub(crate) search_range: Option<Range<MultiBufferOffset>>,
 }
 
 #[derive(Clone, Debug)]
@@ -68,17 +68,17 @@ pub(crate) enum BeamJumpAction {
 
 impl BeamJumpState {
     pub(crate) fn new(
-        direction: BeamJumpDirection,
         smartcase: bool,
-        search_start: MultiBufferOffset,
-        search_end: MultiBufferOffset,
+        cursor_offset: MultiBufferOffset,
+        view_start: MultiBufferOffset,
+        view_end: MultiBufferOffset,
         previous_last_find: Option<Motion>,
     ) -> Self {
         Self {
-            direction,
             smartcase,
-            search_start,
-            search_end,
+            cursor_offset,
+            view_start,
+            view_end,
             previous_last_find,
             pattern: String::new(),
             pattern_len: 0,
@@ -99,14 +99,15 @@ impl BeamJumpState {
                     if labels.label_buffer.len() >= labels.label_len {
                         let label = std::mem::take(&mut labels.label_buffer);
                         if let Some(&start) = labels.start_by_label.get(&label)
-                            && let Some(count) = self.count_for_start(start)
+                            && let Some((direction, count)) =
+                                self.direction_and_count_for_start(start)
                         {
                             return BeamJumpAction::Jump(BeamJumpJump {
-                                direction: self.direction,
+                                direction,
                                 pattern: self.pattern.clone(),
                                 smartcase: self.smartcase,
                                 count,
-                                search_range: self.search_start..self.search_end,
+                                search_range: Some(self.view_start..self.view_end),
                             });
                         }
 
@@ -130,13 +131,20 @@ impl BeamJumpState {
         }
 
         match self.matches.len() {
-            0 | 1 => BeamJumpAction::Jump(BeamJumpJump {
-                direction: self.direction,
-                pattern: self.pattern.clone(),
-                smartcase: self.smartcase,
-                count: 1,
-                search_range: self.search_start..self.search_end,
-            }),
+            0 => BeamJumpAction::Continue,
+            1 => {
+                let m = self.matches[0].clone();
+                let Some((direction, _)) = self.direction_and_count_for_start(m.start) else {
+                    return BeamJumpAction::Cancel;
+                };
+                BeamJumpAction::Jump(BeamJumpJump {
+                    direction,
+                    pattern: self.pattern.clone(),
+                    smartcase: self.smartcase,
+                    count: 1,
+                    search_range: Some(self.view_start..self.view_end),
+                })
+            }
             _ => {
                 if self.labels.is_none() {
                     self.labels = Some(self.assign_labels(buffer));
@@ -165,7 +173,7 @@ impl BeamJumpState {
 
     fn can_extend_pattern_with(&self, ch: char, buffer: &MultiBufferSnapshot) -> bool {
         self.matches.iter().any(|m| {
-            if m.end >= self.search_end {
+            if m.end >= self.view_end {
                 return false;
             }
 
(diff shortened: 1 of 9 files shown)


crates/vim/src/beam_jump.rs
```diff
@@ -0,0 +1,442 @@
...
+
+use crate::motion::{Motion, is_character_match};
+
+const BASE_LABEL_CHARS: &[char] = &[
+    'f', 'j', 'd', 'k', 's', 'l', 'a', 'g', 'h', 'r', 'u', 'e', 'i', 'o', 'w', 'm', 'n', 'c', 'v',
+    'x', 'z', 'p', 'q', 'y', 't', 'b',
+];
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum BeamJumpDirection {
```
ecstatic-morse, line 18: You probably want to reuse [`workspace::searchable::Direction`](https://github.com/zed-industries/zed/blob/4b56fec971579bab18f365be93d46ba68fd1f091/crates/workspace/src/searchable.rs#L22) for this.

crates/vim/src/beam_jump.rs
```diff
@@ -0,0 +1,442 @@
...
+use collections::{HashMap, HashSet};
+use editor::{MultiBufferOffset, MultiBufferSnapshot};
+use ui::SharedString;
+
+use crate::motion::{Motion, is_character_match};
+
+const BASE_LABEL_CHARS: &[char] = &[
+    'f', 'j', 'd', 'k', 's', 'l', 'a', 'g', 'h', 'r', 'u', 'e', 'i', 'o', 'w', 'm', 'n', 'c', 'v',
+    'x', 'z', 'p', 'q', 'y', 't', 'b',
+];
```
ecstatic-morse, line 13: It would be nice to let the user configure this in case they don't use QWERTY.

crates/vim/src/motion.rs
```diff
@@ -2710,6 +2823,151 @@ fn sneak_backward(
     }
 }
 
+fn reverse_beam_jump_direction(direction: BeamJumpDirection) -> BeamJumpDirection {
+    match direction {
+        BeamJumpDirection::Forward => BeamJumpDirection::Backward,
+        BeamJumpDirection::Backward => BeamJumpDirection::Forward,
+    }
+}
```
ecstatic-morse, line 2958: Already a method on  [`workspace::searchable::Direction`](https://github.com/zed-industries/zed/blob/4b56fec971579bab18f365be93d46ba68fd1f091/crates/workspace/src/searchable.rs#L29-L34)

crates/vim/src/vim.rs
```diff
@@ -706,6 +712,9 @@ impl Vim {
             });
 
             Vim::action(editor, cx, |vim, action: &PushSneak, window, cx| {
+                if VimSettings::get_global(cx).beam_jump {
+                    vim.start_beam_jump(BeamJumpDirection::Forward, action.first_char, window, cx);
+                }
```
ecstatic-morse, line 819: It's kind of strange to reuse the `PushSneak` actions here. It means the user can't one mapping for sneak and another for beam jump. Is there a reason you can't introduce new actions just for beam jump?

ecstatic-morse (commented):
I'm not part of Zed, just a user who is excited about this functionality finally landing. I tried this locally and it looks great!

I pretty much exclusively use easy motion's bidirectional mode, so it's a little disappointing to see that part missing. It doesn't work as nicely with `;`, but it's faster than having to decide between forward and backward before starting a jump.

RO03M:
maybe a noob question, is this feature going to be available outside of vim/helix mode? activating it with a shortcut when outside of vim/helix, like we do with goto line would be awesome

xipeng-jin:
Hi @RO03M , thanks for asking. This is a good question. I did think the same way to make this feature outside of vim/helix mode and more globally. We can activate it with a shortcut, like `control-s`. This will definitely help all the users benefit from this feature to be able to navigate around the code faster and more efficient. But I don't think we will push this outside the vim mode for this initial PR for now. The reason is I want this PR to be minimal and focus on the core feature only, after we rolling out the initial version and then we can definitely add more on top of it. Thanks

ConradIrwin:
Linking this with #43733. I think both of them need the same kind of editor UI where we show overlays for jump targets. 

I'd ideally like to only have one approach to implementing this.

ConradIrwin:
Now https://github.com/zed-industries/zed/pull/43733 is merged, I'm going to close this.

I think we can rebuild this feature on top of the UI added for the helix feature (or maybe easier, update the existing vim sneak to use that UI)

aizigao:
https://github.com/zed-industries/zed/pull/43733
As far as I know, this PR is helix-only. What about vim support? Helix doesn't have as many users, we need easy motion or flash.vim, and this feature has been anticipated for too long. Now that helix amp is out, I think sneak should definitely get this overlay added too.

ConradIrwin:
@aizigao PRs welcome! Particular if you want to just update what the current sneak plugin does I don't think it'd be too hard (famous last words).

I know there are a bunch of specific vim plugins that provide "jump to location by typing", but I'm hoping to avoid having too many more. (Though I would also be happy with PRs that make the helix version work for vim users too)

HusuSama:
The implementation similar to `flash.nvim` is completely different from that of #43733. `flash.nvim` leans more toward dynamic matching, whereas #43733 enumerates everything. An implementation like `flash.nvim` is relatively more complex, and I don't think #43733 can adapt to this approach. Moreover, they might be difficult to reconcile. I believe a better way would be to also merge implementations similar to `flash.nvim`, allowing users to choose the method that suits them best. Looking at the long term, I think opening up the development of tool-type plugins is necessary, enabling users to select which one to use via plugins, rather than having everything embedded into Zed's own codebase.​

PR #45424, opened 2025-12-20
Add project panel header with 'Select Opened File' and 'Collapse All' buttons.

tacshi:
Release Notes:

- Add project panel header with 'Select Opened File' and 'Collapse All' buttons.

Recording:

https://github.com/user-attachments/assets/2dfe3fe5-7ceb-4639-a06a-16097f3354ae

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12598,6 +12598,7 @@ dependencies = [
  "language",
  "menu",
  "notifications",
+ "panel",
  "pretty_assertions",
  "project",
  "rayon",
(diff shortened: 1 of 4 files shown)


mattermill:
Thanks for the PR! I'm a bit confused about the problem this is solving, however. The functionality you're proposing already exists elsewhere, and with the same amount of clicks. If you click into a file in the buffer, it will focus that file in the project panel. Additionally, you can `Alt+Click` a parent folder to collapse all children.

Since these features already cover the use cases you're addressing, I'm going to close this PR. If other folks feel strongly enough that there is a workflow gap we're not addressing, I'm happy to reopen for discussion!

PR #45460, opened 2025-12-20
Trigger onTypeFormatting in `vim: insert line above/below`

xdBronch:
like #44882 but for the vim commands

Release Notes:


xdBronch:
thats fair, i didnt think of it in the context of something like what ruby does. it works a lot better with lua or clangd where pressing enter simply formats the previous line, also similar to o/O extending comments.

PR #45478, opened 2025-12-21
git_ui: Show copy-SHA button on commit header hover

cppcoffee:
Release Notes:

- git: Added the ability to copy a commit's SHA in the commit view.


danilo-leal (approved):
Thank you! Pushed a commit refining the UI a bit: switched to use `ButtonLike` and `visible_on_hover` instead of manually tracking hover, removed the "Commit" text because it's pretty obvious this is the commit SHA, capped the commit SHA to an arbitrary width but added it in full in the tooltip, and made the icon change if the SHA's been copied.

https://github.com/user-attachments/assets/3b32c620-2c56-4ad8-9872-9390cd137b6c

PR #37962, opened 2025-09-11
Set StartupWMClass correctly in zed.desktop.in

morganastra:
Adds the `StartupWMClass` field to the `zed.desktop.in` template, so window managers can identify the appropriate desktop file to look up the icon in.

**Missing:**
zed-icon-missing

**Fixed:**
zed-icon-fixed

Closes #37961
Likely will solve #30644 as well - would want confirmation from the original reporter.

Release Notes:

N/A

PR #33019, opened 2025-06-19
linux: Add StartupWMClass to the .desktop file

sw1nn:
`StartupWMClass` is used by some WMs to identify windows during Startup Notification.

See https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html

Since the build already has the correct APP_ID available during the generation of the `.desktop` file we can simply add to the template.

Release Notes:

- N/A

PR #23459, opened 2025-01-22
Fix app icon not showing up and dock bar icons not being merged in Gnome

emmettlu:
Fix `.desktop` file for Gnome.

PR #20644, opened 2024-11-14
Set StartupWMClass in desktop file

someone13574:
Fixes an issue on Gnome: https://github.com/void-linux/void-packages/pull/50476#issuecomment-2473655110

Release Notes:

- N/A

PR #45508, opened 2025-12-22
linux: Name desktop file correctly during bundle

wcbing:
Closes #45507

There have been 5+ PRs attempting to fix Linux desktop icon association by adding `StartupWMClass` back to the `.desktop.in` template (#15763, #20644, #23459, #33019, #37962), all closed because `StartupWMClass` is an X11-specific key that GNOME happens to misuse on Wayland, and the correct fix per the FreeDesktop spec is to name the `.desktop` file to match the app's Wayland `app_id` / X11 `WM_CLASS`. Refer: https://github.com/zed-industries/zed/issues/12707#issuecomment-2168742761

The root cause is that `bundle-linux` was never updated to produce the correct filename in the tarball. The only consumers of the tarball's desktop file are `install.sh` and manual extractors. So this fix is for people who download the tarball and extract it without `install.sh`. The docs already tell them to rename, but as @wcbing points out, not everyone reads that carefully.

We are now making `bundle-linux` produce the correctly named desktop file from the start.

Release Notes:

- N/A

### DIFF
--- a/docs/src/linux.md
+++ b/docs/src/linux.md
@@ -91,7 +91,7 @@ ln -sf ~/.local/zed.app/bin/zed ~/.local/bin/zed
 If you'd like integration with an XDG-compatible desktop environment, you will also need to install the `.desktop` file:
 
 ```sh
-cp ~/.local/zed.app/share/applications/zed.desktop ~/.local/share/applications/dev.zed.Zed.desktop
+install -D ~/.local/zed.app/share/applications/dev.zed.Zed.desktop -t ~/.local/share/applications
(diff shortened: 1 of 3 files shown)


smitbarmase (approved):
@wcbing thanks for this, sorry for the delay in reviewing. 

The `bundle-linux` and `docs` changes are correct and I've left them as-is.

I made changes to `install.sh`:
1. Restored the `appid` channel mapping. The `find *.desktop` approach might behave inconsistently across systems, which is risky for a script. I think being explicit here doesn't hurt much.
2. Added a fallback for older tarballs. The install script now tries the new name (`$appid.desktop`) first, and falls back to the old name (`zed$suffix.desktop`) for tarballs built before this change. It's just fallback for someone has a cached older tarball.

What do you think about the current state?

PR #45522, opened 2025-12-22
Allow specifying ZED_VERSION with fallback to latest in `install.sh`

mikelneonedwin:
This PR updates the Linux install script to support installing a specific version of Zed via the
ZED_VERSION environment variable.

- If ZED_VERSION is set, that version will be downloaded.
- If ZED_VERSION is not set, the script defaults to the latest stable version, preserving
  existing behavior.
- Works for both Linux and macOS installs.
- Maintains compatibility with ZED_CHANNEL for preview builds.

This enhancement allows users to:
- Pin a specific Zed version for reproducible setups or CI workflows.
- Easily install older or known-good versions without manually downloading release assets.

Usage examples:
- Install latest stable (default):
  curl -f https://zed.dev/install.sh | sh
- Install specific version:
  curl -f https://zed.dev/install.sh | ZED_VERSION=0.216.0 sh
  
  Use this — it’s correct, concise, and clearly user-facing:

Release Notes

* **Added:** Support for installing a specific Zed version via the `ZED_VERSION` environment variable in the install script (defaults to `latest` when unset).

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 set -eu
 
-# Downloads the latest tarball from https://zed.dev/releases and unpacks it
+# Downloads the a tarball from https://zed.dev/releases and unpacks it
 # into ~/.local/. If you'd prefer to do this manually, instructions are at
 # https://zed.dev/docs/linux.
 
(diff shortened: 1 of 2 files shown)


script/install.sh
```diff
@@ -1,14 +1,15 @@
 #!/usr/bin/env sh
 set -eu
 
-# Downloads the latest tarball from https://zed.dev/releases and unpacks it
+# Downloads the a tarball from https://zed.dev/releases and unpacks it
```
osiewicz:
```suggestion
# Downloads a tarball from https://zed.dev/releases and unpacks it
```

PR #45544, opened 2025-12-22
Conditionally display debugger panel icon based on a setting

tacshi:
Closes [#ISSUE](https://github.com/zed-industries/zed/issues/45506)

Release Notes:

- Conditionally display the debugger panel icon based on a setting to avoid too many error logs


MrSubidubi (approved):
Nice find, thank you.

PR #45545, opened 2025-12-23
terminal: Prevent terminal cursor stretching by always using cell width

### LINKED ISSUE #8516: Terminal cursor "stretched" during shell script execution
### Check for existing issues

- [X] Completed

While a process is running in the Zed terminal, pressing the Tab key and then deleting the indentation causes the cursor to be stretched to the right of the terminal 1 cursor wide.

1. Open Zed Terminal
2. Start a process that will run for a while _(e.g. `sleep 10`)_.
3. Press the Tab key.
4. Press the Backspace key.
5. Notice that the cursor expands to the right of the terminal 1 cursor.

### Environment

Zed: v0.124.6 (Zed Preview)
OS: macOS 14.2.1
Memory: 16 GiB
Architecture: aarch64

https://github.com/zed-industries/zed/assets/16290753/ec722ae2-4c3c-4ff6-a6ec-87c94b5028ca

If you only need the most recent lines, you can run the `zed: open log` command palette action to see the last 1000.

_No response_

tacshi:
Closes https://github.com/zed-industries/zed/issues/8516

Release Notes:

- Prevent terminal cursor stretching by always using cell width.

### DIFF
--- a/crates/terminal_view/src/terminal_element.rs
+++ b/crates/terminal_view/src/terminal_element.rs
@@ -2342,18 +2342,4 @@ mod tests {
         assert_eq!(negative_filtered.first().unwrap().point.line, Line(-7));
         assert_eq!(negative_filtered.last().unwrap().point.line, Line(-4));
     }
-
-    #[test]
-    fn test_cursor_width_always_uses_cell_width() {
-        use gpui::px;
-
-        // The cursor should always use cell_width, regardless of the shaped text width.
-        // This prevents Tab characters or other wide-rendering characters from stretching the cursor.
-        let cell_width = px(10.0);
-
-        // For any text width, cursor width should always be cell_width
-        // Test this by verifying that we use cell_width directly, ignoring text fragment width
-        let cursor_width = cell_width; // This is what shape_cursor now returns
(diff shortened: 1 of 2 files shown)


crates/terminal_view/src/terminal_element.rs
```diff
@@ -2333,4 +2332,18 @@ mod tests {
         assert_eq!(negative_filtered.first().unwrap().point.line, Line(-7));
         assert_eq!(negative_filtered.last().unwrap().point.line, Line(-4));
     }
+
+    #[test]
+    fn test_cursor_width_always_uses_cell_width() {
```
SomeoneToIgnore: What does this test check, exactly?

Veykril (changes requested):
I don't think this is quite right, that will look funny for emojis and other non-standard cell characters. Let's do a check if the text fragment consists solely of whitespace instead, and if that is the case fall back to the cell width instead

Veykril (commented):
As assumed this still breaks the cursor for bigger non whitespace glyphs 

Image

Veykril (approved):
Thanks!

PR #45549, opened 2025-12-23
Hide block cursor text when the cursor is within a redacted range

### LINKED ISSUE #20696: redact_private_values: vim / block cursor exposes private values
### Summary

Steps to reproduce.

1. Enable `redact_private_values` and block cursors or vim mode in settings.
```json
{
  "redact_private_values": true,
  "cursor_shape": "block",
  //"vim_mode": false
}
```
2. Create a private file and open it in zed:
```sh
echo "FOO=BAR" > .env
zed . .env
```
3. Move the cursor and the private values bleed through the cursor.

### Environment

Zed: v0.163.0 (Zed Nightly 093c9cc87b052c4bdf2afc09a6d52e86eb9c7d3d)
OS: macOS 14.7.0
Memory: 64 GiB
Architecture: aarch64

Screenshot 2024-11-14 at 16 06 13

tacshi:
Closes [#ISSUE](https://github.com/zed-industries/zed/issues/20696)

Release Notes:

- Hide block cursor text when the cursor is within a redacted range

Screenshot:

image

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -1706,14 +1706,10 @@ impl EditorElement {
                     if block_width == Pixels::ZERO {
                         block_width = em_advance;
                     }
-                    // Don't show block cursor text if cursor is in a redacted range
-                    let is_in_redacted_range = redacted_ranges
-                        .iter()
-                        .any(|range| range.start <= cursor_position && cursor_position < range.end);
-
-                    let block_text = if let CursorShape::Block = selection.cursor_shape
-                        && !is_in_redacted_range
-                    {
+                    let block_text = if selection.cursor_shape == CursorShape::Block
+                        && !redacted_ranges.iter().any(|range| {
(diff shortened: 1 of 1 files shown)


crates/editor/src/element.rs
```diff
@@ -1705,7 +1706,14 @@ impl EditorElement {
                     if block_width == Pixels::ZERO {
                         block_width = em_advance;
                     }
-                    let block_text = if let CursorShape::Block = selection.cursor_shape {
+                    // Don't show block cursor text if cursor is in a redacted range
```
SomeoneToIgnore: Can we
* remove the useless comment that repeats the code
* move this after the `&&` so we do not have to iterate over ranges all time if the `CursorShape::Block = selection.cursor_shape` condition is false (maybe, turn it into `==` instead of patter matching too?)

SomeoneToIgnore (approved):
Thank you.

PR #45559, opened 2025-12-23
Implement text-based matching bracket logic for Vim '%' motion to correctly find pairs within comments

### LINKED ISSUE #25435: Vim '%' matching bracket motion doesnt work inside comments and certain template languages
### Summary

Vim-mode's `%` bracket matching jump fails when inside commented code and fails on template languages like Shopify's Liquid `{% ... %}` or `{{ ... }}` brackets.

Steps to trigger the problem:
1. Find a piece of valid code where bracket matching works.
2. Comment the code out
3. Try jumping on the same bracket and notice it no longer works
4. For the template language issue, open a file ending in .liquid, notice bracket jumping works on HTML code/JS code but fails on template brackets

Actual Behavior:
Cursor doesnt move

Expected Behavior:
Cursor jumps to matching bracket

Zed: v0.174.6 (Zed) 
OS: macOS 15.3.1
Memory: 48 GiB
Architecture: aarch64

tacshi:
Closes https://github.com/zed-industries/zed/issues/25435

Release Notes:

- Improved vim's '%' motion to always fall back to text-based bracket matching when language-aware matching fails

### DIFF
--- a/crates/vim/src/motion.rs
+++ b/crates/vim/src/motion.rs
@@ -2371,23 +2371,16 @@ fn find_matching_bracket_text_based(
     offset: MultiBufferOffset,
     line_range: Range<MultiBufferOffset>,
 ) -> Option<MultiBufferOffset> {
-    let mut bracket_info = None;
-
-    for (ch, char_offset) in map.buffer_chars_at(offset) {
-        if char_offset >= line_range.end {
-            break;
-        }
-        if let Some(info) = get_bracket_pair(ch) {
-            bracket_info = Some((info, char_offset));
-            break;
-        }
-    }
+    let bracket_info = map
+        .buffer_chars_at(offset)
+        .take_while(|(_, char_offset)| *char_offset < line_range.end)
+        .find_map(|(ch, char_offset)| get_bracket_pair(ch).map(|info| (info, char_offset)));
 
     let (open, close, is_opening) = bracket_info?.0;
     let bracket_offset = bracket_info?.1;
 
+    let mut depth = 0i32;
     if is_opening {
-        let mut depth = 0i32;
         for (ch, char_offset) in map.buffer_chars_at(bracket_offset) {
             if ch == open {
                 depth += 1;
@@ -2399,7 +2392,6 @@ fn find_matching_bracket_text_based(
             }
         }
     } else {
-        let mut depth = 0i32;
         for (ch, char_offset) in map.reverse_buffer_chars_at(bracket_offset + close.len_utf8()) {
             if ch == close {
                 depth += 1;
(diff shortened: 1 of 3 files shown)


crates/vim/src/motion.rs
```diff
@@ -2447,9 +2510,17 @@ fn matching(map: &DisplaySnapshot, display_point: DisplayPoint) -> DisplayPoint
...
+                    .map(|o| o.to_display_point(map))
+                    .unwrap_or(display_point)
+            })
     } else {
-        display_point
+        let line_range = line_range.start.to_offset(&map.buffer_snapshot())
+            ..line_range.end.to_offset(&map.buffer_snapshot());
+        find_matching_bracket_text_based(map, offset, line_range)
+            .map(|o| o.to_display_point(map))
+            .unwrap_or(display_point)
```
dinocosta: Non-blocking but, since `line_range` is also being calculated in the `if` branch of this `if/else`, we could probably de-duplicate this and just declare it before the `if/else` 🤔

crates/vim/src/motion.rs
```diff
@@ -2352,6 +2352,69 @@ fn matching_tag(map: &DisplaySnapshot, head: DisplayPoint) -> Option<DisplayPoin
...
+                depth += 1;
+            } else if ch == close {
+                depth -= 1;
+                if depth == 0 {
+                    return Some(char_offset);
+                }
+            }
+        }
+    } else {
+        let mut depth = 0i32;
```
dinocosta: We could also move this declaration to outside of the `if/else` block 🔍

crates/vim/src/motion.rs
```diff
@@ -2352,6 +2352,69 @@ fn matching_tag(map: &DisplaySnapshot, head: DisplayPoint) -> Option<DisplayPoin
...
+
+    for (ch, char_offset) in map.buffer_chars_at(offset) {
+        if char_offset >= line_range.end {
+            break;
+        }
+        if let Some(info) = get_bracket_pair(ch) {
+            bracket_info = Some((info, char_offset));
+            break;
+        }
+    }
```
dinocosta: One more non-blocking comment, but we could also leverage Rust's iterator methods here 👀 

```rust
let bracket_info = map
    .buffer_chars_at(offset)
    .take_while(|(_, char_offset)| *char_offset < line_range.end)
    .find_map(|(ch, char_offset)| get_bracket_pair(ch).map(|info| (info, char_offset)));
```

dinocosta (approved):
Hey @tacshi ! 👋 

Thank you for tackling this and for including tests, appreciate it! I've tested and reviewed these changes, everything looks good to me.

I've left some suggestions but they're non-blocking so if you don't want to tackle those, let me know and I can merge this! 🙂

dinocosta:
> This PR doesn't appear to work with `use_match_quotes`: #42615.

Hey @hferreiro !  Thank you for reporting, I'll take a look at both, as I believe there's some overlap here and we can probably simplify. 

I believe the purpose of these changes was also to mimic Vim and NeoVim's behavior, where `%` does not take into consideration quotes and backticks, so not sure if `use_match_quotes` should even do anything if the buffer does not have a specified language 🤔

PR #20859, opened 2024-11-19
Fix file missing or duplicated when copying multiple items in project panel + Fix marked files not being deselected after selecting a directory

smitbarmase:
Closes #20858

This fix depends on the sanitization logic implemented in PR #20577. Since that branch may undergo further changes, this branch will be periodically rebased on it. Once #20577 is merged, the dependency will no longer apply.

Release Notes:

- Fix missing or duplicated files when copying multiple items in the project panel.
- Fix marked files not being deselected after selecting a directory on primary click.
- Fix "copy path" and "copy path relative" with multiple items selected in project panel.

**Problem**:

In this case, `dir1` is selected while `dir2`, `dir3`, and `dir1/file` are marked. Using the `marked_entries` function results in only `dir1`, which is incorrect.

Currently, the `marked_entries` function is used in five actions, which all produce incorrect results:

1. Delete (via the disjoint function)
2. Copy 
3. Cut
4. Copy Path
5. Copy Path Relative

**Solution**:

1. `marked_entries` function should not use  "When currently selected entry is not marked, it's treated as the only marked entry." logic. There is no grand scheme behind this logic as confirmed by piotr [here](https://github.com/zed-industries/zed/issues/17746#issuecomment-2464765963).
2. `copy` and `cut` actions should use the disjoint function to prevent obivous failures.
3. `copy path` and `copy path relative` should keep using *fixed* `marked_entries` as that is expected behavior for these actions. 

---

1. Before/After: Partial Copy

Select `dir1` and `c.txt` (in that order, reverse order works!), and copy it and paste in `dir2`. `c.txt` is not copied in `dir2`.

---
2. Before/After: Duplicate Copy

Select `a.txt`, `dir1` and `c.txt` (in that order), and copy it and paste in `dir2`. `a.txt` is duplicated in `dir2`.

---
3. Before/After: Directory Selection

Simply primary click on any file, now primary click on any dir. That previous file is still marked.

---
4. Before/After: Copy Path and Copy Path Relative

Upon `copy path` (ctrl + alt + c):

Before: Only `/home/tims/test/dir2/a.txt` was copied.
After: All three paths `/home/tims/test/dir2`,  `/home/tims/test/c.txt` and `/home/tims/test/dir2/a.txt` are copied. 

This is also how VSCode also copies path when multiple are selected.

PR #45567, opened 2025-12-23
project_panel: Fix file missing or duplicated when copying/moving multiple items using drag-n-drop

### LINKED ISSUE #45555: Drag & drop of expanded folders can break directory structure and silently delete same-named files
### Reproduction steps

1. Create a project with the following structure.
```bash
zedtest/
├─ src/
│  ├─ folder1/mod.rs
│  ├─ folder2/mod.rs
│  ├─ folder3/mod.rs
│  └─ main.rs
├─ Cargo.toml
└─ Cargo.lock
```

2. Expand the directories in the file tree.

3. Using Shift + Click, select:
* folder1
* folder2/mod.rs
Image

4. Drag and drop the selected folders/files into the project root.

In VS Code(Cursor), this behaves as expected: folder1 (including mod.rs) and folder2 (including mod.rs) are moved to the root.

Image

<br>
<br>

In Zed, the result is inconsistent and seems dangerous: empty folder1 and empty folder2 are moved to the root, and only folder2/mod.rs is moved. folder1/mod.rs is missing afterward, which looks like a silent loss due to a name collision during the operation.

Image

<br>
<br>

Interestingly, if I drag the exact same selection into folder3 instead of the project root, the move completes normally. The existing folder3/mod.rs seems to prevent the move from being split/flattened, and no files go missing.

Image

OS: Windows 10.0.26200
Memory: 31.2 GiB
Architecture: x86_64
GPU: AMD Radeon 780M Graphics || AMD Corporation || 24.10.16 (24.10.16-240528a-403811C-Channel)

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

tacshi:
Closes https://github.com/zed-industries/zed/issues/45555

Refer https://github.com/zed-industries/zed/pull/20859 for more.

Release Notes:

- Fixed a project panel drag-and-drop issue where selecting both a folder and its children could result in files being silently lost in some cases.


crates/project_panel/src/project_panel.rs
```diff
@@ -3965,8 +4081,60 @@ impl ProjectPanel {
...
+            if !conflicts.is_empty() {
+                let message = format!(
+                    "The following files already exist:\n\n{}\n\nDo you want to replace them?",
+                    conflicts.join("\n")
+                );
+                let prompt = window.prompt(
+                    PromptLevel::Warning,
+                    &message,
+                    None,
+                    &[PROMPT_REPLACE, PROMPT_CANCEL],
```
smitbarmase: Even after replacing, it still won’t work, since we don’t currently handle replacement at the worktree store level in `rename_entry` in `worktree_store.rs`. We should add a `force_overwrite` flag that’s only `true` for the project panel case.

Also, the logic for checking whether it’s a conflict should be shared between `rename_entry` and `detect_move_conflicts`.

crates/project_panel/src/project_panel.rs
```diff
@@ -3909,6 +4019,12 @@ impl ProjectPanel {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
+        // Filter out entries that are children of other selected directories.
+        // This matches VSCode behavior: when both a parent folder and its child are selected,
+        // only the parent folder is operated on.
+        let all_selections: Vec<SelectedEntry> = selections.items().cloned().collect();
+        let filtered_selections = self.filter_disjoint_selections(&all_selections, cx);
```
smitbarmase: We already have logic for `filter_disjoint_selections` in `disjoint_selections`. I think `disjoint_selections` can be refactored to take in any selections, and in other places we can use `disjoint_effective_selections`.  This way we don't need to maintain duplicate logic for `filter_disjoint_selections`.

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -4287,6 +4287,264 @@ async fn test_drag_entries_between_different_worktrees(cx: &mut gpui::TestAppCon
     );
 }
 
+/// Regression test for issue #45555:
```
smitbarmase: We should test successful replacement happened after clicking Replace. I don't think we are doing that right now.

smitbarmase (changes requested):
Wouldn't it still throw already exists error on clicking replace in prompt? Since we don't currently handle replacing at `worktree_store` level: https://github.com/zed-industries/zed/blob/dd2cfc734da54db5d62f4fcd95520c3fe4f8ff3d/crates/project/src/worktree_store.rs#L377. Also I think the logic for checking if same file exists or not should be common between project panel check and worktree store. 

I think we should handle this prompt logic at worktree store level instead then.

smitbarmase (approved):
Thanks, I removed the replace dialog logic. Looking forward to that PR with suggested changes. Merging this!

tacshi:
> Thanks, I removed the replace dialog logic. Looking forward to that PR with suggested changes. Merging this!

Thanks! I’ll follow up with the PR once I get a bit of time. Things have been busy since the holidays ended.

PR #45568, opened 2025-12-23
docs: Remove reference to outdated curated issues board

danielbyiringiro:
The documentation referenced a “Curated board of issues” GitHub Project that no longer exists.
The linked project returns a 404, and only three public projects are currently available under
zed-industries.

This PR removes the outdated reference. Documentation-only change.

Release Notes:

- N/A


danilo-leal (approved):
Thanks, good catch!

Issue #45020, opened 2025-12-16
Zed Nightly (2025-12-16) broken on Mac Intel


notpeter:
### Reproduction steps

On my Mac Intel hardware, 16inch MacBook Pro (2019 Intel) Zed Nightly is broken.

Steps to repro: install Zed Nightly or build from main and run.

Pre-built Nightly release binaries launch but do not create a display surface and hang. There is log output showing it starting, but nothing else of note.  Process hangs and needs to be killed.

Debug builds from main will show the following:
```log
[MTLTextureDescriptorInternal validateWithDevice:]:1416: failed assertion `Texture Descriptor Validation
MTLTextureDescriptor: MTLResourceStorageModeMemoryless is not a supported storage mode on this device.
MTLTextureDescriptor: Multisample textures cannot be allocated with MTLResourceStorageModeMemoryless on this device.
'
```

Running the bisect now. But I'm pretty sure it's:
- https://github.com/zed-industries/zed/pull/44273

### Zed version and system specs

Zed: v0.218.0+nightly.2013.9c32c29238d2f4b6006e61b979079482bf07f9dd (Zed Nightly) 
OS: macOS 26.1
Memory: 64 GiB
Architecture: x86_64

notpeter:
Bisect confirmed.
Broken by https://github.com/zed-industries/zed/pull/44273 2441dc3f6637431a781ae10b2e1aa8c4704b9502
CC: @marcocondrache @Anthony-Eid

notpeter:
Fixed by: https://github.com/zed-industries/zed/pull/45022
I built main at abcf5a127357e8d24b033550eed38c5966eee6cb it works now.
Thanks all.

PR #45577, opened 2025-12-23
gpui: Take advantage of unified memory on Apple silicon

marcocondrache:
Reapplies #44273

I included metal-rs upgrade so we can get this working on Intel-based Macs

cc: @JosephTLyons @notpeter @rtfeldman @Anthony-Eid 

Release Notes:

- Reduced memory usage on Apple-silicon Macs by using shared memory where appropriate


Cargo.toml
marcocondrache: The crate `core-video` depends on `metal` 0.29, which is problematic because we have two different versions of metal. We can either patch it or replace it with `objc2_core_video`.

marcocondrache: I've opened an upstream [PR](https://github.com/rust-media/apple-media-rs/pull/7) but not sure if that crate it's still being maintained

Anthony-Eid: What's the difference between objc2_core_video and core-video? I'm not the right person to make a call on which to use, but having more context would still be helpful

marcocondrache: core-video is mostly hand-written wrappers around core-foundation, and I’m not even sure it’s still being actively maintained. objc2-core-video, on the other hand, is auto-generated from Apple’s SDK headers using the objc2 framework and seems much more actively maintained.

Actually, we rely pretty heavily on core-foundation (even though we use both ecosystems), and moving to objc2-core-video isn’t a drop-in replacement. So it probably makes more sense for now to patch core-video or wait (hopefully) for the upstream to get merged.

Anthony-Eid: Seems like it would be worth it to eventually move to `objc2-core-video` from `core-video` then. Probably in a different PR if there's going to be a large amount of changes required to get it working.

I can ask the team about it, but I don't think this will be a priority because we're focused on other things right now besides GPUI.

marcocondrache: Thanks a lot. I will keep it open and updated to main in the meantime

marcocondrache:
@Anthony-Eid The upstream PR has been merged, and I’ve updated both `core-video` and `metal-rs`, so this PR is ready to be merged as well. If we have the release today, it would be great to merge this so it gets a full week of testing in nightly.

notpeter:
Sorry @marcocondrache.
I didn't get a chance to test this branch before it got merged.
Did anyone test this on Intel Macs to confirm it works?

I updated nightly from https://github.com/zed-industries/zed/compare/949972bbbe6d1fedb1022aae9a66e174d232c933...22e2c7ac8ada5a80cba04bbcbb38cd621943d05d and I'm seeing the same behavior I reported in https://github.com/zed-industries/zed/issues/45020

Zed Nightly launches, no display surface, no errors in log. Hangs and needs to be Force Quit.
Output from a debug build on main (b1fd95b4f90b23eb8e20f3a2eeb47c75f2098b13) is the same as before:
```log
-[MTLTextureDescriptorInternal validateWithDevice:]:1416: failed assertion `Texture Descriptor Validation
MTLTextureDescriptor: MTLResourceStorageModeMemoryless is not a supported storage mode on this device.
MTLTextureDescriptor: Multisample textures cannot be allocated with MTLResourceStorageModeMemoryless on this device.
'
zsh: abort      cargo run
```

PR #45580, opened 2025-12-23
setting: Add `editor.select_word_as_node` setting for syntax node selection behavior

dangh:
This PR adds the `editor.select_word_as_node` setting (defaulting to `true`) to guard the behavior introduced in https://github.com/zed-industries/zed/pull/28864
.

The goal is to provide an opt-out for users who find this behavior unintuitive or counterproductive—particularly when they intend to select the actual syntax node rather than the word.

Note: This patch was generated with the assistance of Gemini.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -284,6 +284,8 @@
   // 4. Never show the scrollbar:
   //    "never" (default)
   "completion_menu_scrollbar": "never",
+  // Whether to select the word under the cursor when expanding the selection.
+  "select_word_as_node": true,
   // Show method signatures in the editor, when inside parentheses.
   "auto_signature_help": false,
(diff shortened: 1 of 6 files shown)


dangh:
@smitbarmase Thanks for explaining. My concern is that invoking the action twice feels unintuitive, and that’s something I’d prefer to avoid.

In the JS code, selectLargerNode will select the node, but when the cursor is inside a string it will selecting the word instead. Since we already have Cmd+D for word selection, hijacking selectLargerNode for this behavior feels incorrect to me.

Personally, I’d rather revert this behavior than introduce a setting, but that would disrupt other users’ workflows. For now, I’ll continue maintaining a fork that suits my needs.

Thanks for taking the time to consider this.

PR #45581, opened 2025-12-23
markdown_preview: Fix inline HTML <img> tags not being rendered

MateoKruk:
Closes #38748

## Problem

The markdown preview was not rendering `` tags unless there was a blank line before them. This caused images indented under list items to silently fail to display.

 **Example markdown that wasn't working:**

 ```md
- `Dock`: A UI element similar to a `Pane` which can be opened and hidden. There can be up to 3 docks open at a time, left right and below the center. A dock contains one or more `Panel`s not `Pane`s. (see image).
<img width="1921" height="1080" alt="image" src="https://github.com/user-attachments/assets/2cb1170e-2850-450d-89bb-73622b5d07b2" />

- `Project`: One or more `Worktree`s
```

The first image would not render, while an image with a blank line before it would work fine.

image

## Solution

Updated the markdown parser to handle inline `` tags that don't have a blank line before them.

image

Release Notes:

- Fixed markdown preview not rendering  tags that don't have a blank line before them (https://github.com/zed-industries/zed/issues/38748)

### DIFF
--- a/crates/markdown_preview/src/markdown_parser.rs
+++ b/crates/markdown_preview/src/markdown_parser.rs
@@ -270,20 +270,21 @@ impl<'a> MarkdownParser<'a> {
                 }
 
                 Event::InlineHtml(html) => {
-                    let (_, html_range) = self.current().unwrap();
-                    let html_range = html_range.clone();
-
-                    if let Some(image) = self.extract_inline_html_image(&html, html_range) {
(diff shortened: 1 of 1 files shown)


MateoKruk:
@cla-bot check

MateoKruk:
@cla-bot check

PR #45608, opened 2025-12-24
Improve agent thread navigation by adding a back button

tacshi:
Release Notes:

- Added back button in history session detail view to navigate back.

Recording:

https://github.com/user-attachments/assets/5d2be643-59e9-4538-a06d-07f2a577b115


danilo-leal:
Hey, thanks for the PR! I don't think we want to add this, though: the plus icon button—which shows all the possibilities to create a new thread—and the `cmd-n` keybinding itself already do the job of getting you back to the empty state. Thanks again!

PR #45611, opened 2025-12-24
editor: Fix ctrl+drag/drop to duplicate selected text

chipgw:
Holding control while dragging text to copy it was being blocked by link handling stopping the propogation of the mouse event; solution: don't handle link clicks when selection_drag_state is Dragging.

Issue found & fix tested on Linux (since the modifier key on mac is apparently different, the issue likely doesn't occur there, but I'm assuming it does apply to Windows).

I didn't see an issue open for it, and since I fixed it myself I'm unsure whether it needs one or if the PR is enough.

Release Notes:

- Fix an issue where Ctrl+drag to duplicate selected text on Linux and Windows when the selection is inside a link.

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -1020,6 +1020,10 @@ impl EditorElement {
             && hovered_link_modifier
             && mouse_down_hovered_link_modifier
             && text_hitbox.is_hovered(window)
+            && !matches!(
+                editor.selection_drag_state,
+                SelectionDragState::Dragging { .. }
+            )
         {
             let point = position_map.point_for_position(mouse_position);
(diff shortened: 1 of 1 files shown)


smitbarmase (approved):
Thanks for handling it. Works as expected.

chipgw:
the modified release note text seems to misunderstand the problem (which indicates I didn't do a great job describing it to begin with). duplicating text by holding control and dragging it was _completely non-functional_, whether it was in a link or not. the code for handling links was the reason why, but it stops propagation of the mouse event regardless of whether there was a link to click or not.
it may actually be desirable to change it so it only absorbs the event when there was a link to click, but this change would still fix at least one edge-case that would exist (i.e. ctrl+dragging over a link, which seems to be what the release notes thinks the issue was)

PR #45617, opened 2025-12-24
WIP Editor: Add border option for current line, like in VSCode

RobiFerentz:
This is a WIP and just preference of mine to have the option in Zed.

In VSCode and offshoots, you can do this:
Screenshot 2025-12-24 at 8 53 20

It helps me somewhat more than a highlight background.
This PR is an attempt to add this to Zed. 
Questions:
1. While the current changes work, they are based of the logic of the highlight background logic, which I'm not sure should be the case. 
2. Should I use this off the theme, like I have currently, or should this option be in a separate setting altogether?
3. Does anyone other than myself actually need this? If not, then of course feel free to scrap this PR.

Please advise if you find the time.


danilo-leal:
Hey, thanks for the PR! I'd be open to adding this if it's something available in VS Code that you and other people might rely on, but it's certainly a niche feature... so, due to that: 1) I'd certainly add it under a setting that's **turned off by default** and 2) I wouldn't create a new theme token for it; would probably use `border_focused` or another token that makes sense that already exists. Given how niche this is, I don't think it affords a new addition to the theme that everyone else building themes would need to reason about.

SomeoneToIgnore:
There had not been any activity for over a month and some conflicts started to appear, so I'll close this for now to keep the queue cleaner.

PR #45619, opened 2025-12-24
Fix trailing whitespace

### LINKED ISSUE #45407: Option to disable automatic trailing whitespace removal on empty lines
### Reproduction steps

1. Open any file that contains empty lines with whitespace (indentation)
2. 2. Save the file
3. 3. Observe that whitespace on empty lines is automatically removed

**Current behavior:**
When saving a file, the editor automatically removes whitespace from empty lines, even when "Remove Trailing Whitespace On Save" is set to false. This behavior happens without user authorization and cannot be disabled.

**Expected behavior:**
There should be an option to preserve whitespace on empty lines. While automatic whitespace removal is great for new projects, it causes significant problems for legacy projects:

- It pollutes pull requests with unnecessary whitespace changes on lines that weren't actually modified
- Makes it harder to review actual code changes in PRs
- Can cause merge conflicts with other developers' changes
- Makes git blame less useful as lines get attributed to whitespace cleanup commits
**Proposal:**
Add a setting like `remove_whitespace_on_empty_lines` (or similar) that can be set to `false` to preserve whitespace/indentation on empty lines while still allowing trailing whitespace removal at the end of lines with code.

Zed: v0.217.3 (80433cb239e868271457ac376673a5f75bc4adb1)
Version: 0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

tacshi:
Closes [#ISSUE](https://github.com/zed-industries/zed/issues/45407)

Release Notes:

- Fixed `remove_trailing_whitespace_on_save` configuration

Recording (outdated, see new commit below):

https://github.com/user-attachments/assets/dde66213-fc68-4f1f-81ba-d1905f00be07


tacshi:
The latest commit respects the proposition of @yeskunall 

Recording:

https://github.com/user-attachments/assets/bc19727e-8505-475a-b3e2-377422b15f3e

maxdeviant (changes requested):
I don’t think the analysis in that issue that this is a bug is correct.

Removing whitespace from every single line is how I would expect this feature to work.

If we want to support this alternate behavior, it would need to be behind a separate setting.

tacshi:
> I don’t think the analysis in that issue that this is a bug is correct.
> 
> Removing whitespace from every single line is how I would expect this feature to work.
> 
> If we want to support this alternate behavior, it would need to be behind a separate setting.

It seems there's some disagreement regarding certain functionalities.
How was the first version of the screen recording? I'd be happy to revert to that version and submit it.

tacshi:
**Hard** reset to the first commit.
@maxdeviant

crates/settings_ui/src/settings_ui.rs
```diff
@@ -1162,6 +1164,8 @@ struct SettingItem {
     field: Box<dyn AnySettingField>,
     metadata: Option<Box<SettingsFieldMetadata>>,
     files: FileMask,
+    /// Indentation level for child settings (0 = no indent, 1 = one level, etc.)
+    indent_level: u8,
```
SomeoneToIgnore: I also am not sure why are we fixing trailing whitespaces by introducing indentation in the settings, is that some extra drive-by change or is it really needed for the fix?

tacshi: It's indeed unnecessary.

MrSubidubi:
image

Will close given the current state of things. Feel free to resubmit after the comments have been adressed and everything is green again. Thanks!

PR #45621, opened 2025-12-24
tasks: Show error for unknown `ZED_` variables

### LINKED ISSUE #23275: Invalid tasks.json ZED_FOO env vars should error
Summary: Invalid ZED_* task variable names fail silently.

Steps to reproduce:

1. Add a task to your task.json with the following and save:
```
{
    "label": "Foo",
    "command": "echo -e \"$ZED_XXX\""
}
```
2. trigger `task::spawn`

Expected behavior: Error on save.

Actual behavior: Silently accepted, but not show in task modal.

Zed: v0.170.0 (Zed Preview)
OS: macOS 15.1.0
Memory: 64 GiB
Architecture: aarch64

playdohface:
Closes #23275

This makes resolving task variables a Result rather than an Option, and shows an error when tasks are not loaded because of unresolvable "special" variables in user defined tasks. Extension/Language tasks will still silently be omitted, as for these an unresolved capture-variable usually means the task is not applicable. 

Release Notes:

- Fixed user-defined tasks with unresolved `ZED_*` variables being silently omitted


dinocosta (commented):
Hey @playdohface 👋 

Thank you so much for suggesting these changes! The idea of having a `TaskResolutionError` seemed really nice, but while reviewing I ended up exploring a different approach, hope you don't mind! 😅

I've kept all changes in [a single commit](https://github.com/zed-industries/zed/pull/45621/changes/85338d305aec02bf76b7d250899f0b9c94a3bf68) so we can easily revert if you see something that worries you.

The main difference is that validation of `ZED_` variables now happens when the file is saved, immediately displaying an error to the user. I like that the feedback is more immediate, as users don't have to open the task picker to discover that invalid variables are being used, and valid tasks in the same file are still loaded and work normally.

This approach also avoids adding a new subscription and upgrading `cx.background_spawn` to `cx.spawn`, which would run task resolution on the UI thread.

Here's a quick screen recording of the new approach. Let me know what you think!

https://github.com/user-attachments/assets/7559c2d9-6e11-497f-b14f-39d8df61b6b8

PR #45632, opened 2025-12-24
ui: Fix LoadingLabel animation panic on CJK/emoji text

MomentDerek:
This PR fixes a crash in LoadingLabel where the loading animation progressively revealed text by slicing with byte offsets (.len()), which can panic for UTF-8 multi-byte characters (e.g., CJK) and emoji.

Release Notes:

- Fix a crash in LoadingLabel’s loading animation when displaying CJK or emoji text.

### DIFF
--- a/crates/ui/src/components/label/loading_label.rs
+++ b/crates/ui/src/components/label/loading_label.rs
@@ -93,14 +93,15 @@ impl RenderOnce for LoadingLabel {
             move |mut label, animation_ix, delta| {
                 match animation_ix {
                     0 => {
-                        let chars_to_show = (delta * text.len() as f32).ceil() as usize;
-                        let text = SharedString::from(text[0..chars_to_show].to_string());
-                        label.set_text(text);
+                        let byte_end =
+                            text.floor_char_boundary((delta * text.len() as f32).ceil() as usize);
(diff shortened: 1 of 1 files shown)


crates/ui/src/components/label/loading_label.rs
```diff
@@ -93,9 +100,11 @@ impl RenderOnce for LoadingLabel {
...
                 match animation_ix {
                     0 => {
-                        let chars_to_show = (delta * text.len() as f32).ceil() as usize;
-                        let text = SharedString::from(text[0..chars_to_show].to_string());
-                        label.set_text(text);
+                        let graphemes_to_show = (delta * text_grapheme_count as f32).ceil() as usize;
+                        let byte_end =
+                            *grapheme_boundaries.get(graphemes_to_show).unwrap_or(&text.len());
+                        let visible_text = SharedString::from(text[0..byte_end].to_string());
+                        label.set_text(visible_text);
```
Veykril: No need to handle graphemers here we can just floor the offset
```
                        let byte_end = text.floor_char_boundary((delta * text.len() as f32).ceil() as usize);
                        let visible_text = SharedString::from(text[0..byte_end].to_string());
                        label.set_text(visible_text);
```

PR #45633, opened 2025-12-24
search: Remove intermediate allocation

marcocondrache:
Release Notes:

- N/A

### DIFF
--- a/crates/search/src/project_search.rs
+++ b/crates/search/src/project_search.rs
@@ -916,7 +916,7 @@ impl ProjectSearchView {
         subscriptions.push(
             cx.subscribe(&results_editor, |this, _, event: &EditorEvent, cx| {
                 if matches!(event, editor::EditorEvent::SelectionsChanged { .. }) {
-                    this.update_match_index(cx);
+                    this.update_match_index(&this.get_matches(cx), cx);
                 }
                 // Reraise editor events for workspace item activation purposes
                 cx.emit(ViewEvent::EditorEvent(event.clone()));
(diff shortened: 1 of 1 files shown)


osiewicz (changes requested):
Eh, we're not really avoiding an allocation in one of those cases (calling `get_matches` still allocates). If you're really after reducing allocations, you can do so without altering the API of `update_match_index` with some finnickery inside of that function.
[45633.patch](https://github.com/user-attachments/files/24333855/45633.patch)

marcocondrache:
@osiewicz I’m not sure I’m following. `update_match_index` was always cloning `match_ranges`. In the selection-change path we need to call `get_matches`, but in `entity_changed` function we’re already cloning `match_ranges` before calling `update_match_index`. Since `update_match_index` only needs a reference, I don’t see why it needs to clone again and then immediately drop it once the reference goes out of scope.

For context, `entity_changed` can be called hundreds of times during large searches, so this ends up being a lot of unnecessary allocations.

marcocondrache:
@osiewicz This is a small, straightforward change, and I’m not particularly attached to it. What’s frustrating is the tone of the feedback over something so minor. Even trivial fixes take time to understand, implement, and submit.

If contributions like this aren’t worth engaging with constructively, I’d rather not spend time opening PRs here.

osiewicz:
Hey @marcocondrache ; I didn't mean to sound like I don't appreciate your contributions at all, quite the contrary. You stand out as a strong contributor in recent times and I have nothing but praise for your work. My feedback was just dictated by drilling into that particular allocation and noticing that there's a way to make it allocate even less. Again though, I did not mean/intend to make you feel bad. My apologies if I came across as such. In particular, I hoped to convey my own patch as bit finnicky, not the code you've originally submitted.

osiewicz (approved):

PR #45638, opened 2025-12-24
gpui: Detect thermal state to help the system

marcocondrache:
Apple recommends checking the system’s thermal state and adjusting behavior to reduce resource usage, giving the system a chance to cool down. While this API is macOS-specific, the same idea likely applies to other platforms - those are intentionally out of scope for this PR.

As a first step, we cap the frame rate at 60 fps when the system reports a critical thermal state. In that situation, pushing higher frame rates doesn’t buy us anything and just adds more heat. This also gives us a hook for future improvements, like reducing other work when the system is under sustained thermal pressure.

Ref: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/RespondToThermalStateChanges.html

Release Notes:

- Zed now reduces resource usage when the system is under high thermal stress


crates/gpui/src/window.rs
```diff
@@ -1185,6 +1186,20 @@ impl Window {
...
+                    && (thermal_state == ThermalState::Serious
+                        || thermal_state == ThermalState::Critical)
+                {
+                    if now.duration_since(last_frame) < Duration::from_micros(16667) {
+                        return;
+                    }
+                }
+
+                last_frame_time.set(Some(now));
+
```
Anthony-Eid: nit: I think we should only call `Instant::now` if the ThermalState is Serious or Critical because it introduces overhead (through a system call) that can be avoided in most cases

marcocondrache:
@notpeter thank you so much for all the infos and sorry for the trouble. I believe I’ve identified the issue and have opened a PR

PR #45650, opened 2025-12-25
docs: Mention support for alternate LSP for Kotlin

ilsubyeega:
Would believe https://github.com/zed-industries/zed/pull/40776 did change the default lsp to official implementation.

Release Notes:

- N/A


ilsubyeega:
@cla-bot check

yeskunall (commented):
This should instead be updated here directly https://github.com/zed-extensions/kotlin/blob/main/README.md as you linked in your change, and not here.

yeskunall:
> I mean, Most user would prefer to see the document from searchable with google, search result zed kotlin https://github.com/zed-industries/zed/pull/1 shows the zed.dev/docs site.

On second thought, yeah you’re right. I’ve updated the copy to match our [docs](https://zed.dev/docs/languages/typescript) for other languages.

Thanks for the suggestion!

PR #32329, opened 2025-06-08
editor: Add trailing whitespace rendering

JonasKaplan:
Closes #5237

- Adds "trailing" option for "show_whitespaces" in settings.json
- Supports importing this setting from vscode

The option in question will render only whitespace characters that appear after every non-whitespace character in a given line.

Release Notes:

- Added trailing whitespace rendering

PR #45667, opened 2025-12-25
docs: Add missing option for `show_whitespaces` setting

Miura-KR:
The value for the `trailing` option in the `show_whitespaces` setting was not documented, so I have added it.

Related Issue:
https://github.com/zed-industries/zed/pull/32329#issuecomment-3581576128

Release Notes:

- N/A


yeskunall:
Thank you for catching (and fixing!) this @Miura-KR and congratulations on your first contribution! 💖

PR #45669, opened 2025-12-25
Fix Zed OOM-ing when macOS file descriptors become invalid

### LINKED ISSUE #42845: Zed Using > 53GB of Application Memory
### Summary

Summary: I returned to my Mac to find an OSX window alerting me that my system was out of application memory and Zed was using 53GB. I'm not certain that Zed is responsible or just getting the blame; see logs below. 

Image

### Description

Steps to trigger the problem:
1. Install Zed, heavily use Claude Code for awhile via the in-Zed terminal. 
2. Leave computer for about 14 hours. 
3. Come back and observer possible error message. 

Actual Behavior: Zed seems to have an uncaught runaway memory path. 
Expected Behavior: Zed would clean up or purge memory usage before it uses over ~50GB. 

Zed: v0.204.3 (Zed) 
OS: macOS 14.4
Memory: 18 GiB
Architecture: aarch64

There are a whole bunch of errors like this for random services leading up to the alert: 

```log
2025-11-15T23:31:57-05:00 ERROR [worktree] error reading target of symlink "/Users/user/Library/Containers/com.apple.appkit.xpc.ColorSampler/Data/Downloads/Postgres 2.app/Contents/Versions/latest": canonicalizing "/Users/user/Library/Containers/com.apple.appkit.xpc.ColorSampler/Data/Downloads/Postgres 2.app/Contents/Versions/latest": No such file or directory (os error 2)
2025-11-15T23:31:57-05:00 ERROR [worktree] error reading target of symlink "/Users/user/Library/Containers/com.apple.PhotoIngestService/Data/Downloads/Postgres.app/Contents/Versions/latest": canonicalizing "/Users/user/Library/Containers/com.apple.PhotoIngestService/Data/Downloads/Postgres.app/Contents/Versions/latest": No such file or directory (os error 2)
```

</details>

SomeoneToIgnore:
Closes https://github.com/zed-industries/zed/issues/42845

Repro steps: https://github.com/zed-industries/zed/issues/42845#issuecomment-3687413958 
Initial investigation and Zed memory trace: https://github.com/zed-industries/zed/issues/42845#issuecomment-3687877977

The PR consists of 2 commits:
* [first](https://github.com/zed-industries/zed/pull/45669/changes/732d308c8d7e9af3649ac71ea65a9c029af820fc) adds cosmetic fixes to remove backtraces from logs yet again and print paths in quotes, as file descriptors may return empty paths.
It also stubs the cause if OOM in project panel: that one traversed all worktrees in `for worktree_snapshot in visible_worktrees` and "accepted" the one with empty paths + never called `entry_iter.advance();` in "no file name found for the worktree" case, thus looping endlessly and bloating the memory quite fast.

* [second](https://github.com/zed-industries/zed/pull/45669/changes/7ebfe5da2fc6d32f3fa2d71c761f8b2ec26d945b) adds something that resembles a fix: `fn current_path` on macOS used the file handler to re-fetch the worktree root file path on worktree root canonicalization failure.
What's odd, is that `libc::fcntl` returns `0` in the case when external volume is not mounted, thus resulting in the `""` path string that is propagated all the way up.

* [third](https://github.com/zed-industries/zed/pull/45669/changes/1a7560cef3e9fac604124c19f46b1f9c7b91815f) moves the fix down to the platform-related FS implementations

The "fix" now checks the only usage of this method inside `async fn process_events` for an empty path and bails if that is the case.
I am not sure what is a better fix, but this stops any memory leaks and given how bad the situation now, seems ok to merge for now with the `TODO` comment for more clever people to fix properly later.

----------------

Now, when I disconnect the SMB share and reconnect it again, Zed stops displaying any files in the project tree but the ones opened as editors.

As before, at first, when the share is unmounted, Zed fails to save any changes because of the timeouts.

Later, when the share is re-connected, macOS Finder hangs still but Zed starts to react on saves yet still only shows the files that are open as editors.
The files can be edited and saved from now on.

Later, when Finder finally stops hanging and indicates that the share is mounted fully, the rest of the file structure reappear in the project panel, and all file saves are propagated, hence can be observed in the share in Finder.

It feels that one good improvement to add on top is some "disconnected" indicator that clearly shows that the file is not properly handles in the OS.
This requires much more changes and thinking as nothing like that exists in Zed yet, hence not done.

Release Notes:

- Fixed Zed OOM-ing when macOS file descriptors become invalid

### DIFF
--- a/crates/fs/src/fs.rs
+++ b/crates/fs/src/fs.rs
@@ -339,6 +339,7 @@ impl FileHandle for std::fs::File {
 
         // SAFETY: `fcntl` will initialize the path buffer.
         let c_str = unsafe { CStr::from_ptr(path_buf.as_ptr().cast()) };
+        anyhow::ensure!(!c_str.is_empty(), "Could find a path for the file handle");
         let path = PathBuf::from(OsStr::from_bytes(c_str.to_bytes()));
         Ok(path)
     }
@@ -370,12 +371,11 @@ impl FileHandle for std::fs::File {
         kif.kf_structsize = libc::KINFO_FILE_SIZE;
 
         let result = unsafe { libc::fcntl(fd.as_raw_fd(), libc::F_KINFO, kif.as_mut_ptr()) };
-        if result == -1 {
-            anyhow::bail!("fcntl returned -1".to_string());
-        }
+        anyhow::ensure!(result != -1, "fcntl returned -1");
 
         // SAFETY: `fcntl` will initialize the kif.
         let c_str = unsafe { CStr::from_ptr(kif.assume_init().kf_path.as_ptr()) };
+        anyhow::ensure!(!c_str.is_empty(), "Could find a path for the file handle");
         let path = PathBuf::from(OsStr::from_bytes(c_str.to_bytes()));
         Ok(path)
     }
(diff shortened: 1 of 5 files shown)


SomeoneToIgnore:
Have tested both Windows and Linux and in my test case, they have never timed out the save when the disconnect happened.
On SMB reconnect, the save finished.

This means that no such OOM is possible technically, but I've added similar checks to Windows and FreeBSD code.

nikivdev:
Thank you lots for this PR, this issue is blocking me hard now.

How can I get your change quickly, can I get zed next and that will have this commit?

Or I have to rebuild from source tree myself?

Thank you.

PR #45671, opened 2025-12-25
git: Add syntax diff highlighting

### LINKED ISSUE #9721: Support Difftastic for visualizing code diffing

marcocondrache:
Closes #9721

As #43691, this implements difftastic’s algorithm using Dijkstra’s algorithm. The implementation is a full rewrite and includes several performance optimizations tailored specifically to Zed’s use case.

I initially experimented with the GumTree algorithm when starting this PR, but its diffs were not as expressive or useful as those produced by [difftastic](https://github.com/zed-industries/zed/pull/45671#discussion_r2650625090).

Release Notes:

- git: Syntax diff highlight


marcocondrache:
The main architectural idea on how to compute this diff efficiently is the following:

- Create the GumTree and store it in InternalDiffHunk (if small enough). This step is quite efficient since tree construction runs only on the AST subtree that changed in that hunk so we don't need to cover the entire TreeSitter tree.
- Run matching and generate diff when the hunk is displayed, then store the result. This step is the most computationally expensive (still requires a lot of optimizations), but caching eliminates redundant computation for repeated views of the same diff.

crates/syntax_diff/src/diff_tree.rs
```diff
@@ -0,0 +1,1261 @@
+// GumTree Algorithm Implementation (Optimized for Git Diffs)
```
marcocondrache: After some experimentation with GumTree, it appears that its diffs are less powerful and less useful than those provided by difftastic. As a result, I plan to switch the implementation to difftastic’s approach.

marcocondrache: Difftastic:

image

GumTree:

image

image

crates/syntax_diff/src/syntax_tree.rs
```diff
@@ -0,0 +1,629 @@
...
+}
+
+/// A syntax tree stored as a Vec of nodes in preorder.
+#[derive(Debug)]
+pub struct SyntaxTree {
+    nodes: Vec<SyntaxNode>,
+}
+
+// TODO: Can this just wrap tree sitter's node?
+// TODO: Can we compute the infos without traversing the tree?
```
marcocondrache, line 49: The main thing stopping us from wrapping tree_sitter’s TreeCursor - and avoiding a full traversal of the ast subtree - is the `structural_hash`, which lets us quickly check whether two subtrees are the same. Everything else can be pulled from the cursor on demand.

One alternative would be to compare the node kind, byte-range length, and source text. The problem is that this turns into an O(source_length) comparison, since you lose the O(1) check you get from hashing. Doing this lazily would be even worse, because you’d end up walking the whole subtree anyway.

crates/syntax_diff/src/syntax_tree.rs
```diff
@@ -0,0 +1,657 @@
...
+            source.hash(&mut hasher);
+
+            // Does this node look like punctuation?
+            //
+            // This check is deliberately conservative, because it's hard to
+            // accurately recognise punctuation in a language-agnostic way.
+            // https://github.com/Wilfred/difftastic/blob/cba6cc5d5a0b47b36fdb028a87af03c89d1908b4/src/diff/graph.rs#L422
+            if source == "," || source == ";" || source == "." {
+                kind = Some(SyntaxHint::Punctuation);
+            } else if ts_node.is_extra() {
```
marcocondrache, line 495: Find a better heuristic for detecting comments and strings.
tree-sitter's `is_extra()` catches most comments, but misses strings. Difftastic uses highlight queries (`@comment`, `@string` captures) for more accurate detection. We should consider leveraging Zed's syntax highlighting infrastructure if available at diff time.

crates/buffer_diff/src/buffer_diff.rs
```diff
@@ -844,6 +868,23 @@ impl BufferDiffInner {
...
+    if byte_range.is_empty() {
+        return None;
+    }
+
+    let text: String = snapshot.text();
+    let layer = snapshot.syntax_layers().next()?;
+    let subtree = layer
+        .node()
+        .descendant_for_byte_range(byte_range.start, byte_range.end)?;
+
```
marcocondrache: This may return a subtree larger than the hunk itself. For nearby hunks, this causes the same nodes to be processed repeatedly, resulting in duplicated ranges and extra work that could be avoided. We need to understand how to improve this

marcocondrache:
With #44838 merged, integrating this got a bit trickier.

Before #44838, we’d call `build_snapshot` right before computing hunks and never actually stored a `language::Buffer` for the base text. That meant we could just wait for parsing to finish and automatically had access to tree-sitter tree just before computing hunks. Integrating syntax diff was easy: change the APIs to take a `language::BufferSnapshot` for the main buffer and the syntax tree was already there.

Now `BufferDiff` owns an `Entity<language::Buffer>` for the base text. To get syntax layers, we have to push new text into that buffer and wait for it to parse. The problem is that `compute_hunks` runs before the base buffer is updated - the buffer only gets the new text later in `set_snapshot()`. So there’s no way to run syntax_diff inside `compute_hunks` anymore.

Classic chicken-and-egg situation.

assets/settings/default.json
```diff
@@ -1241,6 +1241,13 @@
   //
   // Default: true
   "word_diff_enabled": true,
+  // Whether to enable syntax-based diff highlighting in the editor.
+  //
+  // When enabled, changes are detected using the AST structure,
+  // providing more precise highlighting of what changed.
+  //
+  // Default: false
+  "syntax_diff_enabled": false,
```
versecafe: does it make sense to merge these into `"diff_strategy": "line" | "word" | "syntax"`

marcocondrache: Yes, nice idea!

crates/syntax_diff/src/syntax_diff.rs
```diff
@@ -0,0 +1,246 @@
...
+
+use crate::{
+    syntax_delimiters::SyntaxDelimiterTree, syntax_graph::ExceededGraphLimit,
+    syntax_tree::SyntaxHint,
+};
+
+/// Default graph limit (1 million vertices).
+///
+/// Difftastic uses a higher value: https://github.com/Wilfred/difftastic/blob/cba6cc5d5a0b47b36fdb028a87af03c89d1908b4/src/options.rs#L25
+pub const DEFAULT_GRAPH_LIMIT: usize = 1_000_000;
```
versecafe: is this adjustable via settings to allow users to tweak based on their machines limits?

marcocondrache: This works the same way as the word diff limit (max 5 lines). I don’t expect most users to change it, but I’ll leave the call to @Anthony-Eid

crates/buffer_diff/src/buffer_diff.rs
marcocondrache: Solved with `syntax_descendant`

crates/syntax_diff/src/syntax_diff.rs
```diff
@@ -0,0 +1,209 @@
...
+
+use crate::{syntax_graph::SyntaxRoute, syntax_tree::SyntaxHint};
+
+/// Compute a syntax-aware diff between two `SyntaxTree`s.
+///
+/// When `lhs_range` and `rhs_range` are provided, the returned ranges are clipped to
+/// those bounds and made relative (starting from 0). This is useful when diffing syntax
+/// trees that may be larger than the region of interest (e.g., a function containing
+/// multiple diff hunks).
+pub fn diff_trees(
```
marcocondrache, line 23: Maybe this entire crate can stay inside language crate similar to how syntax_map does it? So we have syntax_diff and text_diff together.

marcocondrache:
The side-by-side diff is introducing breaking changes, and it’s difficult to keep this PR open while merging and ensuring nothing breaks in the process. For now, I’m going to close it until the side-by-side diff is shipped and the situation inside `buffer_diff` is more stable.

chiragmRoado:
Hey, just wanted to check here, would you be working on this implementation in future? I kinda want it to happen early.

If you are occupied for near future, I can also look into fixing this PR based on latest changes in Zed, and implement this. 

Would like to know your opinion on this @marcocondrache

chiragmRoado:
> Hey, just wanted to check here, would you be working on this implementation in future? I kinda want it to happen early.
> 
> If you are occupied for near future, I can also look into fixing this PR based on latest changes in Zed, and implement this.
> 
> Would like to know your opinion on this @marcocondrache

Any updates @marcocondrache ? Please let me know by Saturday EOD, so I can start accordingly.

marcocondrache:
@chiragmRoado feel free to start a new PR, remember that https://github.com/zed-industries/zed/pull/44014#issuecomment-3608963966 should still apply

chiragmRoado:
Sure

PR #45673, opened 2025-12-25
Custom headers for OpenAPI compatible providers

aurabindo:
Related discussion: https://github.com/zed-industries/zed/discussions/42663

Release Notes:

* Add custom header support in settings
* Create a UI to manipulate custom headers in AI settings
* Added a card based interface for custom headers
* Moved the "Add header" button to top right instead of botton

image

### DIFF
--- a/crates/agent_ui/src/agent_configuration/add_llm_provider_modal.rs
+++ b/crates/agent_ui/src/agent_configuration/add_llm_provider_modal.rs
@@ -8,7 +8,7 @@ use gpui::{
 };
 use language_model::LanguageModelRegistry;
 use language_models::provider::open_ai_compatible::{AvailableModel, ModelCapabilities};
-use settings::{OpenAiCompatibleSettingsContent, update_settings_file};
+use settings::{CustomHeader, OpenAiCompatibleSettingsContent, update_settings_file};
 use ui::{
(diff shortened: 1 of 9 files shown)


schelhorn:
Hi, any chance you folks could look at this again and make the final changes? Supporting custom API header indeed would be awesome for enterprise settings. Much appreciated!

danilo-leal (changes requested):
Hey @aurabindo, thanks for the PR and sorry for the late response here! In testing this out, I found some fixes we need to make before we're ready to merge (aside from the conflict this branch now has):

1. The input tab order in the "Add LLM Provider" modal is wrong: tabbing from the "Header Name" input gets me to the model name input, as opposed to the "Header Value" input. 
2. We should also add the "Remove Header" button as part of the tab index
3. We should make "Header Name" and "Header Value" the label for the corresponding inputs and add in the placeholder slot an example of content that can be inserted in those inputs
4. After I hit "Save Provider" in the modal, I still see the input to add the API key, as well as the custom HTTP headers in the settings view of the agent panel. This feels wrong to me; I'd expect everything to be filled already (image below).

Screenshot 2026-02-16 at 12  57@2x

After we push these changes, we can fine-tune the UI more. Let me know if you're still interested in working on this, otherwise, I might either close it for now or work on it later time-permitting. Thank you!

PR #45685, opened 2025-12-26
Expose LSP's goto definition and find references to LLM agent

Prohect:
- Adds two agent tools that let an LLM request LSP-powered definition / reference information by supplying:
  - `path` (project-relative file path)
  - `context` (a multi-word snippet that must contain the token)
  - `token` (the identifier inside the context)
  - optional `index` (0-based choice when the context contains multiple matching tokens)

- Tools:
  - `goto_definition_by_context` — locate the token and call the project language service for definitions.
  - `find_references_by_context` — same locating step but asks for references.

- Security & correctness:
  - Foreground checks for project path resolution and `WorktreeSettings` (global and per-worktree) enforce `file_scan_exclusions` and `private_files` before any async work.
  - Async path: open buffer, wait for parsing idle, locate token occurrence, convert chosen offset → anchor, then call `project.definitions` / `project.references`.
  - Error handling uses `anyhow` and returns task results (no panics / unwraps).

What's new since the previous description (this commit)
- Tree‑Sitter token validation: candidate token offsets found by cheap substring search are now validated against the parse tree before being accepted. This prevents false positives where the token substring appears inside a larger identifier (e.g., the `path` inside `is_path_excluded`).
  - Preferred validation: find a named Tree‑Sitter descendant node that exactly covers the token byte range (via `named_descendant_for_byte_range`) and verify its text equals `token`.
  - Fallbacks:
    - If no precise descendant exists, fall back to matching the enclosing node text.
    - If no syntax node / parse tree is available, permissively accept the candidate (keeps the tool usable for unsupported files / transient parse states).
  - Excludes trivial node kinds like comments/strings from positive matches.

- Improved ambiguous-preview logic in `find_references_by_context`:
  - Two-phase preview generation to produce richer context without duplication:
    1. Compute preferred preview ranges for each candidate (prefer enclosing syntax node climbed up to at most `MAX_SCOPE_LINES`, otherwise a small clamped context).
    2. Merge overlapping preview ranges and extract each merged preview once, then map candidates into the merged previews for output.
  - Result: ambiguous listings show the larger, meaningful scope (function/block) when available, while avoiding repeated text for overlapping matches.

Behavioral details / decisions you should know
- Token acceptance policy:
  - Prefer exact named descendant that covers token byte range.
  - Fall back to enclosing node text match if no descendant found.
  - Permissive when parse tree is unavailable (accept candidate).
  - Exclude `comment` and `string` node kinds from matches.
- Preview selection:
  - Prefer syntax ancestor scope, climbing to a larger parent only if that parent remains <= `MAX_SCOPE_LINES`.
  - When ambiguous, merged previews avoid duplicate overlapping snippets.

Testing & status
- I ran `cargo check -p agent` locally; build succeeds.
- I exercised the tools manually with the previously-problematic example:
- raw input:{
  "path": "zed/crates/agent/src/tools/find_references_by_context_tool.rs",
  "context": "let project_path = match project.read(cx).find_project_path(&input.path, cx) {",
  "token": "path",
  "index": 0
}
- raw output:{
  "Text": "zed\\crates\\agent\\src\\tools\\goto_definition_by_context_tool.rs [L92-95]\n\n```\n        let project_path = match project.read(cx).find_project_path(&input.path, cx) {\n            Some(p) => p.clone(),\n            None => return Task::ready(Err(anyhow!(\"Path {} not found in project\", &input.path))),\n        };\n```\n\nzed\\crates\\agent\\src\\tools\\find_references_by_context_tool.rs [L70-73]\n\n```\n        let project_path = match project.read(cx).find_project_path(&input.path, cx) {\n            Some(p) => p.clone(),\n            None => return Task::ready(Err(anyhow!(\"Path {} not found in project\", &input.path))),\n        };\n```\n\nzed\\crates\\agent\\src\\tools\\goto_definition_by_context_tool.rs [L24-38]\n\n```\npub struct ContextPositionInput {\n    /// Project-relative path to the file containing the symbol (e.g. \"src/main.rs\").\n    pub path: String,\n\n    /// A multi-word snippet from the file which explicitly contains `token`.\n    /// The tool will search the file for this exact snippet to locate the token's position.\n    pub context: String,\n\n    /// The exact token inside `context` to locate.\n    pub token: String,\n\n    /// Optional 0-based index to disambiguate multiple matches.\n    #[serde(default)]\n    pub index: Option<u32>,\n}\n```\n\n"
}

Release Notes:

- Added goto_definition_tool and find_references_tool to ACP

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -1,9 +1,10 @@
 use crate::{
     ContextServerRegistry, CopyPathTool, CreateDirectoryTool, DbLanguageModel, DbThread,
-    DeletePathTool, DiagnosticsTool, EditFileTool, FetchTool, FindPathTool, GrepTool,
-    ListDirectoryTool, MovePathTool, NowTool, OpenTool, ProjectSnapshot, ReadFileTool,
-    RestoreFileFromDiskTool, SaveFileTool, SystemPromptTemplate, Template, Templates, TerminalTool,
-    ThinkingTool, WebSearchTool,
+    DeletePathTool, DiagnosticsTool, EditFileTool, FetchTool, FindPathTool,
+    FindReferencesByContextTool, GotoDefinitionByContextTool, GrepTool, ListDirectoryTool,
+    MovePathTool, NowTool, OpenTool, ProjectSnapshot, ReadFileTool, RestoreFileFromDiskTool,
+    SaveFileTool, SystemPromptTemplate, Template, Templates, TerminalTool, ThinkingTool,
+    WebSearchTool,
 };
 use acp_thread::{MentionUri, UserMessageId};
 use action_log::ActionLog;
@@ -1000,6 +1001,9 @@ impl Thread {
         self.add_tool(FetchTool::new(self.project.read(cx).client().http_client()));
         self.add_tool(FindPathTool::new(self.project.clone()));
         self.add_tool(GrepTool::new(self.project.clone()));
+        // Agent LSP helpers: enable context-based goto-definition & find-references tools
+        self.add_tool(GotoDefinitionByContextTool::new(self.project.clone()));
+        self.add_tool(FindReferencesByContextTool::new(self.project.clone()));
         self.add_tool(ListDirectoryTool::new(self.project.clone()));
         self.add_tool(MovePathTool::new(self.project.clone()));
         self.add_tool(NowTool);
--- a/crates/agent/src/tools.rs
+++ b/crates/agent/src/tools.rs
@@ -15,6 +15,9 @@ mod read_file_tool;
 mod restore_file_from_disk_tool;
 mod save_file_tool;
 
+mod find_references_by_context_tool;
+mod goto_definition_by_context_tool;
+
 mod terminal_tool;
 mod thinking_tool;
 mod web_search_tool;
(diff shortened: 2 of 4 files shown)


ask4fusora:
https://github.com/zed-industries/zed/discussions/40021

Can we expose more tools such as `rename`, `get_all_references`, `code_actions` ... ?
That'll be powerful, especially for refactoring tasks.

Prohect:
> Closing this for now. We need to decide how we want to do this (we have the diagnostics tool already - though it's often less reliable than using a shell task to get them).
> 
> We've found that with most models more tools make things worse because it confuses them, and for most cases grep can find these things.

Now many LLM supports 1M context, does this kinda idea( expose more LSP capabilities to LLM agent) helpful enough?
Though `goto definition` or `find all references` may be not that helpful comparing to `grep` or `terminal` grep since LLM is super powerful dealing with reading large results, it still make some sense.
While `rename` should be really helpful. Most times the LLM trying to rename something across multi files (or any form of edit across multi files / large code blocks), they call `read_file`/`edit_file` so much times, or when they try to use `terminal` sed or awk or cat or what ever, they may get confused since those cli tools may have different feature/syntax on different platform.

ask4fusora:
Agree. `rename` lsp tool would be incredibly powerful.

lionel-:
> Though goto definition or find all references may be not that helpful 

Goto-definition is helpful to figure out source locations for installed files of dependencies, for nicher languages where the agent is not sure where to find these.

Find-references is helpful to avoid token blow-up from grep in large codebases (e.g. vscode) or repos with large minified files.

PR #45695, opened 2025-12-26
gpui: Add clipboard pasting support for DIB images on Windows

YangAoLib:
- Added handling for DIB format, converting to BMP after adding the corresponding file header

Release Notes:

- fix: Unable to paste images from clipboard directly into agent panel on Windows

### DIFF
--- a/crates/gpui/src/platform/windows/clipboard.rs
+++ b/crates/gpui/src/platform/windows/clipboard.rs
@@ -379,7 +379,7 @@ fn convert_dib_to_bmp(dib_data: &[u8]) -> Option<Vec<u8>> {
     let pixel_data_offset = 14 + header_size + color_table_size;
 
     // Build BITMAPFILEHEADER (14 bytes)
-    let mut bmp_data = Vec::with_capacity(14 + dib_data.len());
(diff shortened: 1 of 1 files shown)


crates/gpui/src/platform/windows/clipboard.rs
```diff
@@ -339,21 +340,77 @@ fn read_metadata_from_clipboard() -> Option<String> {
...
+    } else if compression == 3 {
+        12 // BI_BITFIELDS
+    } else {
+        0
+    };
+
+    let pixel_data_offset = 14 + header_size + color_table_size;
+
+    // Build BITMAPFILEHEADER (14 bytes)
+    let mut bmp_data = Vec::with_capacity(14 + dib_data.len());
```
Veykril:
```suggestion
    let mut bmp_data = Vec::with_capacity(file_size as usize);
```

Veykril (commented):
Thanks!

YangAoLib:
@Veykril Has this PR been released? I still can't get it to work with the latest version.

Veykril:
No it will land today I believe

PR #45719, opened 2025-12-26
Extract SymbolIndex to project crate for Tree-sitter based symbol search

portseif:
## Summary

Extracts the symbol indexing functionality from Search Everywhere into a reusable `symbol_index` module in the project crate. This enables Tree-sitter based symbol search for languages that don't have LSP support.

**Key changes:**
- New `symbol_index.rs` module in `crates/project/`
- `SymbolIndex` struct with cached symbols per project
- Incremental updates via buffer and worktree subscriptions
- `search()` function for fuzzy matching symbols
- Indexes outline items from Tree-sitter (works without LSP)

This is a foundational piece that can be used by Search Everywhere and potentially other features that need project-wide symbol access.

## Test plan

- [x] Build passes (`./script/clippy -p project`)
- [ ] Manual testing with Search Everywhere using the new API


maxbrunsfeld (changes requested):
Left my changes in the comment above.

PR #45733, opened 2025-12-27
Fix OpenRouter reasoning details accumulation in streaming responses

c0va23:
OpenRouter sends multiple `reasoning_details` chunks during streaming responses, with each chunk containing partial reasoning information. However, our code was only keeping the last chunk, causing reasoning details to be lost.

This PR introduces `ReasoningDetailsAccumulable` event type that properly accumulates all reasoning details chunks in order, as documented in [OpenRouter's streaming response behavior](https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#streaming-response).

## Changes:
- Added `ReasoningDetailsAccumulable(Vec<serde_json::Value>)` event type to collect all reasoning detail chunks
- Added thread logic to extend reasoning details array when accumulable events are received
- Updated OpenRouter event mapper to use the accumulable variant
- Added test coverage for reasoning details accumulation

## Context:
I found this issue while I tested the new Minimax M2.1 model. It model shows itself very well into the opencode, but it works very stranger into Zed Agent. The investigation led me to that this model requires keeping all previous reasoning between requests to proper work. And the debugging of Zed and OpenCode requests showed me this issue. 

The full request and response bodies before and after this fix (I formatted the body of second requests and pointed out to the issue):
[after-patch.1.txt](https://github.com/user-attachments/files/24353941/after-patch.1.txt)
[after-patch.2.txt](https://github.com/user-attachments/files/24353939/after-patch.2.txt)
[before-patch.1.txt](https://github.com/user-attachments/files/24353942/before-patch.1.txt)
[before-patch.2.txt](https://github.com/user-attachments/files/24353940/before-patch.2.txt)

## Alternatives:
I don't like my solution with introducing the new event kind `ReasoningDetailsAccumulable`, but I hope it's minimally invasive.

### Some better and some bigger
As alternative I can replace type of value into ReasoningDetails event kind with a new enum type. Something like that:
```
enum ReasoningDetailsBehavior {
   Latest(serde_json::Value),
   Concatenate(Vec<serde_json::Value>)
```
It looks more correct, but it will require changes across other providers.

I am ready to implement this solution after approval of maintainers.

### Maybe but
From my point of view the best solution for this issue is migrating to Responses API that contains the summary into final message into stream. But it matches [this](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#things-we-will-probably-not-merge)

## Potential problems
These changes could break the thread when switching to or from other providers or OpenRouter. However, I'm not sure if it worked before.

## Release Notes:
- Fix OpenRouter reasoning details preservation in streaming responses

### DIFF
--- a/crates/agent/src/tests/mod.rs
+++ b/crates/agent/src/tests/mod.rs
@@ -668,6 +668,72 @@ async fn test_streaming_tool_calls(cx: &mut TestAppContext) {
 }
 
 #[gpui::test]
+async fn test_reasoning_details_accumulation(cx: &mut TestAppContext) {
+    let ThreadTest { model, thread, .. } = setup(cx, TestModel::Fake).await;
+    let fake_model = model.as_fake();
+
+    let _ = thread
+        .update(cx, |thread, cx| {
+            thread.send(UserMessageId::new(), ["test"], cx)
+        })
+        .unwrap();
+    cx.run_until_parked();
+
+    let chunk1 = vec![serde_json::json!({
+        "type": "reasoning.text",
+        "text": "First",
+        "format": "test",
+        "index": 0
+    })];
+    let chunk2 = vec![serde_json::json!({
+        "type": "reasoning.text",
+        "text": "Second",
+        "format": "test",
+        "index": 1
+    })];
+    let chunk3 = vec![serde_json::json!({
+        "type": "reasoning.text",
+        "text": "Third",
+        "format": "test",
+        "index": 2
+    })];
+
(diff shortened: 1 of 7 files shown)


rtfeldman:
@c0va23 thanks for diving into this! You mentioned:

> I don't like my solution with introducing the new event kind `ReasoningDetailsAccumulable`, but I hope it's minimally invasive.

I agree with your assessment here - it's minimally invasive but I think will be error-prone long-term compared to something like changing `ReasoningDetails` to be `Vec<serde_json::Value>` and then update all the existing call sites.

What do you think about that as an alternative?

PR #45747, opened 2025-12-27
svg_preview: Detect SVG in single-file mode by checking file name

gshahbazian:
Release Notes:

- Use the files name for "is svg" checks so SVG previews and the toolbar button work in single-file mode.

### DIFF
--- a/crates/svg_preview/src/svg_preview_view.rs
+++ b/crates/svg_preview/src/svg_preview_view.rs
@@ -191,15 +191,23 @@ impl SvgPreviewView {
     }
 
     pub fn is_svg_file(buffer: &Entity<MultiBuffer>, cx: &App) -> bool {
-        buffer
-            .read(cx)
-            .as_singleton()
-            .and_then(|buffer| buffer.read(cx).file())
-            .is_some_and(|file| {
-                file.path()
-                    .extension()
-                    .is_some_and(|ext| ext.eq_ignore_ascii_case("svg"))
-            })
+        let Some(singleton) = buffer.read(cx).as_singleton() else {
+            return false;
+        };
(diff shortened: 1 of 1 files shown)


crates/svg_preview/src/svg_preview_view.rs
```diff
@@ -191,15 +191,23 @@ impl SvgPreviewView {
...
+        let Some(file) = singleton.read(cx).file() else {
+            return false;
+        };
+
+        let rel_ext = file.path().extension();
+        let full_path = file.full_path(cx);
+        let full_ext = full_path.extension().and_then(|ext| ext.to_str());
+        let name_ext = std::path::Path::new(file.file_name(cx))
+            .extension()
+            .and_then(|ext| ext.to_str());
```
MrSubidubi: I think replacing the previous `path().extension()` above with this would be sufficient, right? That way, we can remove the other two checks, as they are redundant in that case.

MrSubidubi (changes requested):
Thanks for looking into this! I think we can simplify this, left a comment below.

Also, not required, but could I interest you into looking to add a test case for this? You might need to copy quite some code for this from elsewhere (and even need to add the test elsewhere). Yet, with a test, we'd be more certain this does not regress any time soon (and I think changes here need to be done at some point, so would be nice to have, but as mentioned, not required). Thanks!

crates/svg_preview/src/svg_preview_view.rs
gshahbazian: @MrSubidubi thanks for the suggestion, it is applied now

MrSubidubi (approved):
Nice! Thank you both for the fix as well as the fast follow up! Appreciated

PR #45749, opened 2025-12-27
languages: Escape test names for bun runner

### LINKED ISSUE #45746: bun test runner fails when test names contain regex control characters
### Reproduction steps

Create a test:
```ts
import { expect, test } from "bun:test";

test("2 plus 2", () => {
  expect(2 + 2).toBe(4);
});
```

Click Play Icon in Zed:

Image

Works.

Modify test to:
```ts
import { expect, test } from "bun:test";

test("2 + 2", () => {
  expect(2 + 2).toBe(4);
});
```

Click Play Icon in Zed.

Operations fails with:

`error: regex "2 + 2" matched 0 tests. Searched 1 file (skipping 1 test) [4.00ms]`

Current: Zed fails to match test to run
Expected: Zed matches test to run

OS: macOS 26.1
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

No relevant logs, see above.

### Relevant Zed settings

N/A

### Relevant Keymap

N/A

_No response_

None

pckilgore:
Closes #45746

Release Notes:

- Fixed un-escaped regexp characters in test names passed to `bun test`

### DIFF
--- a/crates/languages/src/typescript.rs
+++ b/crates/languages/src/typescript.rs
@@ -56,6 +56,9 @@ const TYPESCRIPT_JASMINE_PACKAGE_PATH_VARIABLE: VariableName =
 const TYPESCRIPT_BUN_PACKAGE_PATH_VARIABLE: VariableName =
     VariableName::Custom(Cow::Borrowed("TYPESCRIPT_BUN_PACKAGE_PATH"));
 
+const TYPESCRIPT_BUN_TEST_NAME_VARIABLE: VariableName =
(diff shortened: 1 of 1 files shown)


crates/languages/src/typescript.rs
pckilgore, line 1: This is the same implementation I found for, e.g., `TYPESCRIPT_VITEST_TEST_NAME_VARIABLE`

versecafe (commented):
I believe the issue also exists for the node:test setup this slipped through the cracks when I added support for the two if you could patch that in this PR as well would be great.

Issue #20913, opened 2024-11-20
Opening user home directory triggers many macOS permissions prompts (e.g. Contacts, Reminders, Calendar)


rowillia:
### Description

When starting Zed via `zed ~/` to get open the editor, the application requests extensive filesystem access permissions. Specifically, it attempts to recursively scan the user's home directory, requesting permissions for every conceivable filesystem access.

Steps to reproduce:
1. Open Zed by running `zed ~`
2. Restart the application
3. Observe multiple filesystem permission requests

This behavior appears to be triggered when Zed accidentally opens up the root directory, resulting in unnecessary permission requests.

### Environment

Zed: v0.161.2 (Zed Nightly 00d96d9ea4f75afcdd5d08b08ff20f3df363676a)
OS: macOS 14.7.1
Memory: 32 GiB
Architecture: aarch64

notpeter:
Zed should likely handle this better. It's a terrible user experience to have Zed pop-up multiple permissions dialogs in quick succession on macOS.

For reference VSCode gives a similarly bad experience, but it doesn't happen immediately on startup, but as soon as you try and search in files or cmd-p for a filename you'll get the same prompts no real context:
<img width="260" alt="Screenshot 2024-12-05 at 14 35 02" src="https://github.com/user-attachments/assets/34d8b59a-94a2-4d6e-9116-cbb0d30741a2">

I believe all these folders may trigger prompts
- ~/Desktop
- ~/Documents
- ~/Downloads
- ~/Library/Application Support/AddressBook
- ~/Library/Calendars
- ~/Library/Reminders

I'm not sure exactly what triggers the prompt shown, but I think perhaps something under ~/Library.

I don't have any great suggestions for how to better handle this because I don't think it's possible to query "Does Zed have access to ~/Downloads" and instead merely accessing it will either succeed or trigger the prompt and block (and perhaps timeout) waiting for the user to answer the prompt.

We could conceivably have an onboarding workflow which intentionally triggered those prompts in a controlled setting or have some sort of pre-warning when accessing known restricted directories: "Zed wants to access your Desktop folder, which will trigger a macOS security prompt (skip, skip always, allow, [allow always]), but definitely open to suggestions.

a3lem:
Just encountered this. Even after quitting Zed the dialog boxes kept coming.

I didn't even open home (~) as a project. I simply opened a python file at the top of my home directory.

[...]

PR #45752, opened 2025-12-27
Use temp_dir as root_dir fallback to avoid macOS TCC dialogues

tacshi:
Release Notes:

- Fixed the issue where the OpenCode agent causes macOS TCC pop-ups

This issue is related to [#20913](https://github.com/zed-industries/zed/issues/20913). When your selected agent is OpenCode and hasn't opened any project, the TCC dialogue will prompt every time.

### DIFF
--- a/crates/agent_ui/src/acp/thread_view.rs
+++ b/crates/agent_ui/src/acp/thread_view.rs
@@ -560,7 +560,9 @@ impl AcpThreadView {
                 })
                 .log_err()
             } else {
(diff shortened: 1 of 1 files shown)


tacshi:
> @tacshi Does this issue still happen for you in the latest stable release (0.221.5) ? Is it as easy as the steps below:
> 
> 1. Open Zed window with no project
> 2. Open OpenCode Agent
> 3. Send a message
> 
> Can't seem to reproduce it on my end 🤔

Thanks for checking! This PR is outdated now — the issue no longer reproduces on the latest stable release, so I’m going to close it.

PR #45753, opened 2025-12-28
Allow `;` and `,` to repeat motions besides `fFtT`

ecstatic-morse:
Notably, this lets you jump forward and backwards between methods (`]m`) and sections/classes (`]]`) using `;` and `,` without having to repeatedly press an awkward two character key combination. These motions are more useful in Zed than in stock Vim, since Zed has treesitter built-in.

See [`vim-remotions`](https://github.com/vds2212/vim-remotions) for prior art.

Release Notes:

- vim: allow `;` and `,` to work with more motions


ecstatic-morse:
Marking as draft since this has no tests yet. I'll wait for some feedback from a team member before writing them.

crates/vim/src/motion.rs
```diff
@@ -1085,147 +1303,6 @@ impl Motion {
...
-                    after,
-                    char,
-                    mode,
-                    smartcase,
-                } => {
-                    let mut new_point =
-                        find_backward(map, point, after, char, times, mode, smartcase);
-                    if new_point == point {
-                        new_point =
-                            find_backward(map, point, after, char, times + 1, mode, smartcase);
```
ecstatic-morse, line 1116: It's unclear why we need to try again with an incremented count. Is this supposed to be for when `times = 0` or something? Eliminating it doesn't cause any test failures, so I'm hoping someone familiar with this code can explain...

ConradIrwin: Not certain. What happens if you find backward for a character your cursor is currently on? That seems likely to be the edge-case.

ecstatic-morse: Looks like it's for repeated invocations of `t` and `T`. After the first motion the cursor ends up *before* the character, but when you repeat you want to skip past it. Other motions that move the cursor *to* the next `DisplayPoint` aren't affected (including `f` and `F`, which is what I was testing with).

See https://github.com/ecstatic-morse/zed/blob/4b6ce7448ce2f763bc0230307c35518142467159/crates/vim/src/motion.rs#L1418-L1420

ecstatic-morse:
> The thing I don't like is an extra set of names. Every time I create a new motion I have to either make a new name or figure out which name it should be filed under. As this only exists to configure a tiny feature (which most users of vim mode won't use) that's a subset of vim mode (which most users of zed don't use), it seems over the top. For now, effectively a boolean of "remotion or not" seems like the right level of control/code complexity.

"What is a good short name  for this motion?" is a much easier question to answer than "should this be a part of the extended set of repeatable motions?".

Once again. It's not a hill I'm willing to die on. If you tell me to rewrite it I will.

> Interestingly, we do have next diagnostic and change (but they're `]d` and `]c`, not `]g`)

`[d` is from NeoVim. I believe `[g` is the default mapping from the `vim-lsp` plugin.

FYI, I am planning on making the more complex motions available to vim in a later PR.

> Do you use the default remotion config, or would you tweak it to be better?

I don't think using `;` and `,` to repeat single character motions is necessary: It doesn't save keystrokes. Therefore, I would remove `{` (paragraph). This lets you use `;` and `,` to jump between methods while simultaneously using `{` and `}` to jump within them (otherwise `{` would clobber the `[m` and `;` and`,` would no longer jump between methods).

Also it's strange to include `{` but not `(` (sentence).

ConradIrwin:
Closing for now

PR #45754, opened 2025-12-28
Ignore `.gdb_history`

ecstatic-morse:
Release Notes:

- N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 .claude/settings.local.json
 .envrc
 .flatpak-builder
+.gdb_history
 .idea
 .netrc
 *.pyc


SomeoneToIgnore (commented):
Can you elaborate why this repo needs another exclusion, esp. given the fact that no one in this org seem to use gdb to debug Zed?

And why cannot this exclusion live in a global gitignore instead?

ecstatic-morse:
You have `.gitignore` entries for `.DS_Store`, `.envrc` and `.vscode`. If a succinct `.gitignore` is important, then close this and I'll put it in global.

Out of curiosity, do most of you use `lldb`? `rr`? `lldb` takes forever to load debuginfo for me.

SomeoneToIgnore:
Those entries weird as well, not sure why do we have to keep them too, tbh.

I know that many use `lldb`, and have not heard from the rest on this particular topic, so seems that no one needs to exclude the file?

For now, I'd abstain from changing the file and wait for more PRs or opinions, thank you for the understanding.

PR #45759, opened 2025-12-28
gpui: Fix clear drag overlay when external drag ends outside window

cppcoffee:
Release Notes:

- Fixed clear drag overlay when external drag ends outside window

When dragging files from macOS Finder over the project panel and then dragging back to Finder, the drag overlay remained visible because the drag state was not properly cleaned up.

The root cause was that only `draggingExited:` was handled, but not `draggingEnded:`. On macOS:
- `draggingExited:` is called when the drag leaves the window area
- `draggingEnded:` is called when the drag operation ends entirely

When a user drags a file back to Finder and drops it there, `draggingEnded:` is called but was not being handled.

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -4714,10 +4714,10 @@ impl ProjectPanel {
 
     fn clear_drag_state(&mut self, cx: &mut Context<Self>) {
         let had_drag_state = self.drag_target_entry.take().is_some()
-            || self.folded_directory_drag_target.take().is_some()
-            || self.hover_scroll_task.take().is_some()
-            || self.hover_expand_task.take().is_some()
-            || self.previous_drag_position.take().is_some();
+            | self.folded_directory_drag_target.take().is_some()
+            | self.hover_scroll_task.take().is_some()
+            | self.hover_expand_task.take().is_some()
+            | self.previous_drag_position.take().is_some();
         if had_drag_state {
             cx.notify();
         }
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -24,10 +24,10 @@ use gpui::{
     ClipboardItem, Context, CursorStyle, DismissEvent, Div, DragMoveEvent, Entity, EventEmitter,
     ExternalPaths, FocusHandle, Focusable, FontWeight, Hsla, InteractiveElement, KeyContext,
     ListHorizontalSizingBehavior, ListSizingBehavior, Modifiers, ModifiersChangedEvent,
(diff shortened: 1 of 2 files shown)


cppcoffee:
Before:

https://github.com/user-attachments/assets/c5004393-17c9-4d68-9d9f-fcd856ea77c5





After:


https://github.com/user-attachments/assets/4db20c99-086c-4c0a-afb5-01d882a46ada

crates/project_panel/src/project_panel.rs
```diff
@@ -5598,6 +5483,14 @@ fn item_width_estimate(depth: usize, item_text_chars: usize, is_symlink: bool) -
 
 impl Render for ProjectPanel {
     fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        if !cx.has_active_drag() {
+            self.drag_target_entry = None;
+            self.folded_directory_drag_target = None;
+            self.hover_scroll_task.take();
+            self.hover_expand_task.take();
+            self.previous_drag_position = None;
+        }
```
Veykril: This feels like the wrong spot to do this, we should not change the state in rendering, we should clear the state in the drag handlers

cppcoffee: GPUI side added `Context::observe_window_file_drop(...)`, changed to event mode.

entity:
This has been plaguing me for ages and I never knew what it was, would loveee if this was merged. Thank you

cppcoffee:
rebase main branch done.

Veykril (commented):
Sorry for not coming back to this for so long. I don't think we need to do all of this complexity. I believe we can just register a MouseExitEvent handler and clear the drag in ther?

crates/project_panel/src/project_panel.rs
```diff
@@ -4713,6 +4712,17 @@ impl ProjectPanel {
         }
     }
 
+    fn clear_drag_state(&mut self, cx: &mut Context<Self>) {
+        let had_drag_state = self.drag_target_entry.take().is_some()
+            || self.folded_directory_drag_target.take().is_some()
+            || self.hover_scroll_task.take().is_some()
+            || self.hover_expand_task.take().is_some()
+            || self.previous_drag_position.take().is_some();
```
Veykril:
```suggestion
        let had_drag_state = self.drag_target_entry.take().is_some()
            | self.folded_directory_drag_target.take().is_some()
            | self.hover_scroll_task.take().is_some()
            | self.hover_expand_task.take().is_some()
            | self.previous_drag_position.take().is_some();
```
These should not be short circuiting

Veykril (commented):
One more thing, but otherwise lgtm, thanks!

Veykril (approved):
Thanks!

PR #45760, opened 2025-12-28
vim: Fix Helix mode subword motions to select traversed text

### LINKED ISSUE #41767: In Helix mode subword movements do not select the last subword unlike word
### Summary

`vim::PreviousSubwordStart` doesn't select passed over word unlike `vim::PreviousWordStart`

### Description
Steps to reproduce:
1. Bind in your config
```json
 "1":"vim::PreviousWordStart",
 "2":"vim::NextWordEnd",
 "3":"vim::PreviousSubwordStart",
 "4":"vim::NextSubwordEnd"      ,
```
2. enable Vim/Helix mode
3. move over words in any text with keys `1` and `2`
4. observe how the last word is selected
5. do the same with `3` and `4`
6. observe how the last subword is **not** selected

**Expected Behavior**:
The last subword is selected in Helix normal mode just like word

**Actual Behavior**:
The last subword is not selected

Zed: v0.212.0 (Zed Dev ecbdffc84f1165323f256e8485ae84320550c759) 
OS: Windows 10.0.26100
Memory: 12 GiB
Architecture: x86_64
GPU: NVIDIA GeForce GTX 750 || NVIDIA Corporation || 576.2 r575_92

maxmalkin:
Closes #41767

Release Notes:
In Helix mode, word motions select the text they traverse, but subword motions were only moving the cursor without selecting. This was because `helix_move_cursor` had explicit handling for word motions using `helix_find_range_forward` and `helix_find_range_backward`, but subword motions fell back to `helix_move_and_collapse` handler which collapses the selection.

Changes and Tests:
  - Added `is_subword_boundary_right` and `is_subword_boundary_left` functions for subword-specific boundary detection 
  - Added explicit match cases for all four subword motions in `helix_move_cursor`
  - Added test `test_subword_motions`

Release Notes:

- Fix subword motions in Helix mode to select traversed text


dinocosta (commented):
Hey @maxmalkin, thanks for tackling this!

I’m not entirely sure how, or whether, this is actually supported in Helix, but it’s worth noting that the current implementation uses only `_` as the word boundary. In the past, some folks have complained that it doesn’t quite behave like Vim’s `CamelCaseMotion`.

For example, https://github.com/zed-industries/zed/pull/48276 updated Vim mode’s implementation to also treat `$` and `=` as word-boundary characters, so I figured I’d flag that 🙂

maxmalkin:
Hi @dinocosta! That's a good catch, it is better in terms of UX to have `$` and `_` treated as subwords for things like assignments `let var=0`  or variables in PHP `$var`, for example. Added in my latest commit, thanks for the review!

dinocosta (changes requested):
Thank you for tackling the suggested changes @maxmalkin ! 🙂

It seems that there's now a failing test, `helix::test::test_subword_motions`, let me know if you're able to see what's going on there, otherwise I can try and take a look. I've also merged `main` into your branch to ensure it stays up to date.

P.S.: Also added a release note which was missing from the PR's description

dinocosta (changes requested):
Hey @maxmalkin !

Unfortunately it seems your latest commit – https://github.com/zed-industries/zed/pull/45760/commits/040718342dc6489afb2c4bde8029e9ee86650021 – doesn't actually fix the failing test, as it actually deleted assertions from existing tests.

Let me know if you're still unable to run tests on your end and I can try and take a look 🙂

maxmalkin:
Hi @dinocosta! Looks like it's still unable to run tests on my end, I would appreciate it if you could take a look! Thanks!

dinocosta (approved):

PR #45771, opened 2025-12-28
Fix text_threads_dir() non-compliance with XDG spec

### LINKED ISSUE #41373: `text_threads_dir()` non-compliant with XDG spec
### Summary

The `text_threads_dir()` function does not mesh with the [XDG spec][1].

### Existing Code

From https://github.com/zed-industries/zed/blob/8a01e483396eec940e9af72ce8f900bb98381a4f/crates/paths/src/paths.rs

```rust
/// Returns the path to the contexts directory.
///
/// This is where the saved contexts from the Assistant are stored.
pub fn text_threads_dir() -> &'static PathBuf {
    static CONTEXTS_DIR: OnceLock<PathBuf> = OnceLock::new();
    CONTEXTS_DIR.get_or_init(|| {
        if cfg!(target_os = "macos") {
            config_dir().join("conversations")
        } else {
            data_dir().join("conversations")
        }
    })
}
```

The logic does not match the XDG spec.

I noticed because I keep `$HOME/.config` in version control, and I noticed conversations showing up there. Past conversational history is not configuration.

### Specification

> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
> - actions history (logs, history, recently used files, …)
> - current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

Therefore, the correct place per the XDG specification is *neither*:
- `config_dir()` : `$XDG_CONFIG_HOME` nor
- `data_dir()` : `$XDG_DATA_HOME`

Per the spec, the correct place is `$XDG_STATE_HOME`. Since the code is using the [dirs crate](https://crates.io/crates/dirs) you can just rely on its `state_dir()`.

[1]: https://specifications.freedesktop.org/basedir-spec/latest/#variables

Zed: v0.209.6 (Zed) 
OS: macOS 15.6.1
Memory: 32 GiB
Architecture: aarch64

maekilae:
Closes #41373 

Release Notes:

- Ensures that XDG specs are followed on MacOS and Linux with backwards compatibility.

Changes
-Added ```state_dir``` to get XDG_STATE_HOME on macos and linux, no change to windows.
-Changed ```text_threads_dir``` to a fallback, ```text_threads_dir_fallback```
-Re-implemented ```text_threads_dir``` to use ```state_dir```

### DIFF
--- a/crates/paths/src/paths.rs
+++ b/crates/paths/src/paths.rs
@@ -146,12 +146,11 @@ pub fn state_dir() -> &'static PathBuf {
             }
             .join("zed");
         } else {
+            // Windows
             return dirs::data_local_dir()
                 .expect("failed to determine LocalAppData directory")
                 .join("Zed");
         }
-
-        // home_dir().join(".local").join("share").join("zed")
     })
 }
 
--- a/crates/paths/src/paths.rs
+++ b/crates/paths/src/paths.rs
@@ -139,8 +139,8 @@ pub fn state_dir() -> &'static PathBuf {
         }
 
         if cfg!(any(target_os = "linux", target_os = "freebsd")) {
-            return if let Ok(flatpak_xdg_data) = std::env::var("FLATPAK_XDG_DATA_HOME") {
-                flatpak_xdg_data.into()
+            return if let Ok(flatpak_xdg_state) = std::env::var("FLATPAK_XDG_STATE_HOME") {
+                flatpak_xdg_state.into()
             } else {
                 dirs::state_dir().expect("failed to determine XDG_STATE_HOME directory")
             }
--- a/crates/paths/src/paths.rs
+++ b/crates/paths/src/paths.rs
@@ -131,10 +131,7 @@ pub fn state_dir() -> &'static PathBuf {
     static STATE_DIR: OnceLock<PathBuf> = OnceLock::new();
     STATE_DIR.get_or_init(|| {
         if cfg!(target_os = "macos") {
-            return home_dir()
-                .join(".local")
-                .join("state")
-                .join("Zed");
+            return home_dir().join(".local").join("state").join("Zed");
         }
 
         if cfg!(any(target_os = "linux", target_os = "freebsd")) {
--- a/crates/paths/src/paths.rs
+++ b/crates/paths/src/paths.rs
@@ -127,6 +127,34 @@ pub fn data_dir() -> &'static PathBuf {
     })
 }
 
+pub fn state_dir() -> &'static PathBuf {
+    static STATE_DIR: OnceLock<PathBuf> = OnceLock::new();
+    STATE_DIR.get_or_init(|| {
+        if cfg!(target_os = "macos") {
+            return dirs::home_dir()
+                .expect("failed to determine LocalData directory")
+                .join(".local")
+                .join("state")
+                .join("Zed");
+        }
+
+        if cfg!(any(target_os = "linux", target_os = "freebsd")) {
+            return if let Ok(flatpak_xdg_data) = std::env::var("FLATPAK_XDG_DATA_HOME") {
+                flatpak_xdg_data.into()
+            } else {
+                dirs::state_dir().expect("failed to determine XDG_STATE_HOME directory")
+            }
+            .join("zed");
+        } else {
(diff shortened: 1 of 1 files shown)


crates/paths/src/paths.rs
```diff
@@ -126,6 +126,33 @@ pub fn data_dir() -> &'static PathBuf {
...
+            return dirs::home_dir()
+                .expect("failed to determine LocalData directory")
+                .join(".local")
+                .join("state")
+                .join("Zed");
+        }
+
+        if cfg!(any(target_os = "linux", target_os = "freebsd")) {
+            return if let Ok(flatpak_xdg_data) = std::env::var("FLATPAK_XDG_DATA_HOME") {
+                flatpak_xdg_data.into()
```
AdamJedl:
```suggestion
            return if let Ok(flatpak_xdg_state) = std::env::var("FLATPAK_XDG_STATE_HOME") {
                flatpak_xdg_state.into()
```

crates/paths/src/paths.rs
```diff
@@ -126,6 +126,33 @@ pub fn data_dir() -> &'static PathBuf {
     })
 }
 
+pub fn state_dir() -> &'static PathBuf {
+    static STATE_DIR: OnceLock<PathBuf> = OnceLock::new();
+    STATE_DIR.get_or_init(|| {
+        if cfg!(target_os = "macos") {
+            return dirs::home_dir()
+                .expect("failed to determine LocalData directory")
```
AdamJedl:
```suggestion
            return home_dir()
```

crates/paths/src/paths.rs
```diff
@@ -127,6 +127,29 @@ pub fn data_dir() -> &'static PathBuf {
...
 
+pub fn state_dir() -> &'static PathBuf {
+    static STATE_DIR: OnceLock<PathBuf> = OnceLock::new();
+    STATE_DIR.get_or_init(|| {
+        if cfg!(target_os = "macos") {
+            return home_dir().join(".local").join("state").join("Zed");
+        }
+
+        if cfg!(any(target_os = "linux", target_os = "freebsd")) {
+            return if let Ok(flatpak_xdg_state) = std::env::var("FLATPAK_XDG_STATE_HOME") {
```
hferreiro, line 138: Shouldn't this be `XDG_STATE_HOME`?

maekilae: > Shouldn't this be `XDG_STATE_HOME`?

Im assuming that `FLATPAK_XDG_STATE_HOME` points to `~/.var/app/zed/.local/state` rather than the standard `~/.local/state` for `XDG_STATE_HOME`. According to the flatpak conventions this is where we should store state data when using flatpak. ref: https://docs.flatpak.org/en/latest/conventions.html#xdg-base-directories

hferreiro: IIUC, the link above explains that `XDG_STATE_HOME` points to `~/.var/app/dev.zed.Zed/.local/state`, while `HOST_XDG_STATE_HOME` is the one pointing to `~/.local/state`.

maekilae: `XDG_STATE_HOME` points to `~/.local/state` according to the xdg spec. `XDG_STATE_HOME` within the docs i linked to previously refer to where the applications state home is located within the flatpak sandbox(`~/.var/app/dev.zed.Zed/.local/state`). xdg spec ref:https://specifications.freedesktop.org/basedir/latest/#variables.

The question is how does zed handle env variables. Is the env `FLATPAK_XDG_STATE_HOME` pointing to the state home in the flatpak sandbox or to nothing. And is `XDG_STATE_HOME` essentially a constant or does it change depending on whether or not the application is run as a flatpak.

Other functions within the path crate use the same convention, `data_dir()`, `config_dir()` use `FLATPAK_XDG_DATA_HOME` and `FLATPAK_XDG_CONFIG_HOME` respectively.

hferreiro: I'm sorry, I've just seen now that `FLATPAK_XDG_STATE_HOME` is set by Zed. Sorry for the noise.

Issue #17460, opened 2024-09-05
Disable keyring prompting


mil-ad:
### Summary

When I open zeditor (on Arch linux) it prompts me to authenticate my system keyring. I don't want to give keyring access to my editor and pressing Escape gets rid of the password prompt. Is there a way to stop zeditor from requesting keyring access?

notpeter:
I believe we use the keyring for storing the following items:
- Zed.dev session (Logging into Zed for Collab and Zed AI)
- Assistant panel API Tokens (OpenAI, Anthropic, Gemini, etc)

On MacOS, if you logout from Zed and remove and click `Reset key` for any configured providers in the Assistant you should no longer receive prompts to access your Keychain.

On Linux, things are a little different (see: `read_credentials` and `write_credentials` in [crates/gpui/src/platform/linux/platform.rs](https://github.com/zed-industries/zed/blob/main/crates/gpui/src/platform/linux/platform.rs).  It's not immediately clear to me the behavior if you don't have any stored credentials in your Keyring (logged out; assistant unconfigured; previously saved items deleted) whether you will be prompted, but ideally you would not.

constantitus:
> It's not immediately clear to me the behavior if you don't have any stored credentials in your Keyring (logged out; assistant unconfigured; previously saved items deleted) whether you will be prompted, but ideally you would not.

I did get prompted the first time I opened the editor. Nothing has to be configured for you to get prompted.

musjj:
I deleted `~/.config/zed` and `~/.cache/zed` and the issue still persists (`0.154.1`). Not sure what's triggering it. I looked over my `gpg --list-keys` and I'm only seeing my personal key and proton vpn key.

notpeter:
@musjj Wrong keyring.  This isn't in your GPG Keyring, it's in your XDG secrets provider (likely `gnome-keyring` or `kwallet`).

More in the docs: [Zed isn't remembering my login](https://zed.dev/docs/linux#zed-isnt-remembering-my-login)

mil-ad:
@notpeter true but I was wondering if there is a way to disable it? I don't want to be prompted for anything when I open zed, and if that means zed won't remember my login I'm okay with that

Gambloide:
I also get gnome-keyring asking me to unlock my default keyring every time I open zed for the first time after a reboot. I am not using any of the assistant or collaboration features and never logged into anything through Zed, hence there is no good reason to even try to look into the keyring to determine if there is anything usable saved.

[...]

PR #45782, opened 2025-12-28
linux: Fix multiple keyring prompts

jkful:
Related to #17460

On Linux, multiple keyring unlock prompts could appear due to `keyring.unlock()` calls. This change:

- Checks `item.is_locked()` before calling `item.unlock()` in read/delete
- Only unlocks collection in write if `IsLocked` error is returned

Users should now see at most one prompt if keyring is locked.

Tested on Debian 13.2 (trixie) with GNOME Keyring.

Release Notes:

- Fixed multiple keyring unlock prompts on Linux

### DIFF
--- a/crates/gpui_linux/src/linux/platform.rs
+++ b/crates/gpui_linux/src/linux/platform.rs
@@ -4,6 +4,8 @@ use std::{
     rc::Rc,
     sync::Arc,
 };
+
+use smol::lock::OnceCell;
 #[cfg(any(feature = "wayland", feature = "x11"))]
 use std::{
     ffi::OsString,
@@ -17,6 +19,10 @@ use anyhow::{Context as _, anyhow};
 use calloop::{LoopSignal, channel::Sender};
 use futures::channel::oneshot;
 use gpui_util::{ResultExt as _, new_std_command};
+use oo7::{
+    Error as Oo7Error,
+    dbus::{Error as DbusError, ServiceError},
+};
 #[cfg(any(feature = "wayland", feature = "x11"))]
 use xkbcommon::xkb::{self, Keycode, Keysym, State};
 
@@ -123,6 +129,7 @@ pub(crate) struct LinuxCommon {
     pub(crate) callbacks: PlatformHandlers,
     pub(crate) signal: LoopSignal,
     pub(crate) menus: Vec<OwnedMenu>,
+    pub(crate) keyring: Arc<OnceCell<Result<Arc<oo7::Keyring>>>>,
     app_name: Option<String>,
     system_notifications: crate::linux::system_notifications::SystemNotificationState,
     #[cfg_attr(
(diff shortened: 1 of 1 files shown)


jkful:
@cla-bot check

bilelmoussaoui:
The changes looks good to me. Although, keep in mind, nothing in the spec says that you can always search for items even if the collection where they are stored is locked. Would be nice if people could verify if this assumption would work also in kwallet/keepass

probably-neb:
Hey sorry for the delay on this. This change makes a lot of sense. Once the merge conflicts are resolved I'd be happy to merge

jkful:
@probably-neb No worries, just got around to this again. Should be all set here, let me know if anything else is needed.

DaniPopes:
@smitbarmase @probably-neb Hello, is there anything left to do here? Would love to have this in. Spawning multiple agent sessions in Zed on Linux exposes a bug in Gnome Keyring daemon which makes it crash and logs out the keyring completely: https://github.com/zed-industries/zed/issues/17460#issuecomment-4406107125

jkful:
@cla-bot check

crates/gpui_linux/src/linux/platform.rs
```diff
@@ -746,6 +766,22 @@ impl<P: LinuxClient + 'static> Platform for LinuxPlatform<P> {
     fn add_recent_document(&self, _path: &Path) {}
 }
 
+async fn get_keyring(
+    keyring: &Arc<OnceCell<Result<Arc<oo7::Keyring>>>>,
+) -> Result<Arc<oo7::Keyring>> {
+    keyring
+        .get_or_init(|| async {
```
probably-neb, line 773:
```suggestion
        .get_or_try_init(|| async {
```

We should probably use the `OnceCell::get_or_try_init` api to leave the cell uninitialized if we hit a transient load error so that we don't perma-block ourselves from using the keyring if the first request fails (transiently or otherwise).

crates/gpui_linux/src/linux/platform.rs
```diff
@@ -653,36 +661,46 @@ impl<P: LinuxClient + 'static> Platform for LinuxPlatform<P> {
...
+        let keyring = self.inner.with_common(|common| common.keyring.clone());
 
+        self.background_executor().spawn(async move {
+            let keyring = get_keyring(&keyring).await?;
             let items = keyring.search_items(&vec![("url", &url)]).await?;
 
             for item in items.into_iter() {
                 if item.label().await.is_ok_and(|label| label == KEYRING_LABEL) {
+                    if item.is_locked().await? {
+                        item.unlock().await?;
```
probably-neb, line 698: This seems potentially racy, multiple callers could hit `is_locked() == true` before a single `unlock()` resolves. It's probably fine to serialize these calls across different callers

probably-neb (changes requested):

SomeoneToIgnore:

PR #45785, opened 2025-12-28
signature_help: Fix crash when typing CJK in debug mode

CharlesChen0823:
Closes #ISSUE

in release mode: found many log error as `ERROR [rope::chunk] byte index 48 is not a char boundary; it is inside '，' (bytes 46..49)`

This is easy testing,
- enable signature help
- typing any CJK character
- crash in debug mode

Release Notes:

- N/A


Veykril (approved):
Nice catch, thanks!

PR #45789, opened 2025-12-28
agent: Sanitize MCP server IDs in tool name disambiguation

kung-foo:
Release Notes:

- Fixed an issue where a request could fail if an MCP server with names containing whitespace was used

## Summary

When multiple MCP servers expose tools with the same name, Zed disambiguates them by prefixing the tool name with the server ID from settings.json. If the server ID contains spaces or special characters (e.g., `"Azure DevOps"`), the resulting tool name like `Azure DevOps_echo` violates Anthropic's API pattern `^[a-zA-Z0-9_-]{1,128}$`, causing API errors:

> "Received an error from the Anthropic API: tools.0.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'"

## Solution

Convert server IDs to snake_case (using the `heck` crate already available in the workspace) before using them as prefixes during tool name disambiguation.

| Server ID in settings.json | Disambiguated Tool Name |
|---------------------------|------------------------|
| `"Azure DevOps"` | `azure_dev_ops_echo` |
| `"My MCP Server"` | `my_mcp_server_echo` |

## Test plan

- [x] Added test case for server name with spaces ("Azure DevOps") in `test_mcp_tool_truncation`
- [x] Verified existing tests pass
- [x] Manually tested with two MCP servers having overlapping tool names

After (left), Before (right):

Screenshot_20251228_163249

🤖 Generated with (some) help from [Claude Code](https://claude.com/claude-code)


bennetbo (approved):
Thank you!

PR #45794, opened 2025-12-28
languages: Include namespace in the outline for C++

hokein:
Before: 

image

After:
image

Release Notes:

- Include namespace symbol in the outline view for C++.


osiewicz (approved):
Thanks!

Issue #45056, opened 2025-12-17
Can't open files in ssh remote project since 0.218.0


mathroc:
### Reproduction steps

1. Start zed in a remote SSH project
2. try to open a file

### Current vs. Expected behavior

opening a file (clicking on them in the explorer or from the palette) does nothing, not even an error, the palette stays open

but I can see the file in the explorer, in the palette and the terminal works correctly

### Zed version and system specs

Zed: v0.218.0+preview.100.0982a59d66a7967acbe3363d15e3fd3e7ef4b69d (Zed Preview)
OS: Linux Wayland fedora 42
Memory: 30.5 GiB
Architecture: x86_64
GPU: AMD Radeon 860M Graphics (RADV GFX1152) || radv || Mesa 25.1.9

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
2025-12-17T01:39:24+01:00 INFO  [zed] ========== starting zed version 0.218.0+preview.100.0982a59d66a7967acbe3363d15e3fd3e7ef4b69d, sha 0982a59 ==========
2025-12-17T01:39:24+01:00 INFO  [crashes] spawning crash handler process
2025-12-17T01:39:24+01:00 WARN  [crates/zed/src/reliability.rs:22] No such file or directory (os error 2)
2025-12-17T01:39:24+01:00 INFO  [extension_host] extensions updated. loading 7, reloading 0, unloading 0
2025-12-17T01:39:24+01:00 INFO  [gpui::platform::linux::platform] activate is not implemented on Linux, ignoring the call
2025-12-17T01:39:24+01:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-17T01:39:24+01:00 INFO  [client] set status on client 0: Authenticating
2025-12-17T01:39:24+01:00 INFO  [gpui::platform::blade::blade_renderer] Initializing Blade pipelines for surface SurfaceInfo { format: Bgra8Unorm, alpha: PreMultiplied }
2025-12-17T01:39:24+01:00 INFO  [workspace] Rendered first frame
2025-12-17T01:39:24+01:00 INFO  [zed::zed] Using GPU: GpuSpecs { is_software_emulated: false, device_name: "AMD Radeon 860M Graphics (RADV GFX1152)", driver_name: "radv", driver_info: "Mesa 25.1.9" }
[...]

PR #45803, opened 2025-12-29
Reorder remote worktree trust initialization, yield buffer operations, and optimize remote server binary builds

tacshi:
Release Notes:

- Fixed SSH remote file opening failure
- Prevents server messages from arriving before the client is ready
- Skip rebuilding if local binary exists
- Check the remote first before building/uploading

Recordings:

Before:

https://github.com/user-attachments/assets/4b2f1c87-d431-4ec0-9429-976388ebfa59

After:

https://github.com/user-attachments/assets/d5a3d4cb-c83a-4ea6-997e-e3e5875cfe16

This PR may be related to [45056](https://github.com/zed-industries/zed/issues/45056), but I cannot reproduce it, so I cannot verify.

### DIFF
--- a/crates/project/src/project.rs
+++ b/crates/project/src/project.rs
@@ -1292,15 +1292,6 @@ impl Project {
 
             cx.subscribe(&worktree_store, Self::on_worktree_store_event)
                 .detach();
-            if init_worktree_trust {
-                trusted_worktrees::track_worktree_trust(
-                    worktree_store.clone(),
-                    Some(RemoteHostLocation::from(connection_options)),
-                    None,
-                    Some((remote_proto.clone(), REMOTE_SERVER_PROJECT_ID)),
-                    cx,
-                );
-            }
 
             let weak_self = cx.weak_entity();
             let context_server_store =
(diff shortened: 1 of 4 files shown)


crates/remote/src/transport.rs
```diff
@@ -218,105 +218,119 @@ async fn build_remote_server_from_source(
...
+    let bin_path_buf = workspace_root
+        .join("target")
         .join("remote_server")
         .join(&triple)
         .join("debug")
         .join("remote_server")
         .with_extension(if platform.os.is_windows() { "exe" } else { "" });
+    let bin_path = bin_path_buf.as_path();
+
+    // Check if binary already exists. Use ZED_BUILD_REMOTE_SERVER=force to always rebuild.
```
SomeoneToIgnore, line 232: Is it so that the new code will skip building only because there's some binary in the path already?
That seems odd: I build a dev extension and I want its changes to always propagate to the server.
People were bitten by this already hence introduced unconditional building.

crates/remote/src/transport/ssh.rs
```diff
@@ -625,6 +625,21 @@ impl SshRemoteConnection {
         let dst_path =
             paths::remote_server_dir_relative().join(RelPath::unix(&binary_name).unwrap());
 
+        // First, check if the remote already has a working binary. If so, skip build and upload.
```
SomeoneToIgnore, line 628: Same comment as in https://github.com/zed-industries/zed/pull/45803#discussion_r2650617399

crates/project/src/project.rs
```diff
@@ -1498,6 +1489,18 @@ impl Project {
             GitStore::init(&remote_proto);
             AgentServerStore::init_remote(&remote_proto);
 
+            // Initialize worktree trust *after* handlers are registered to prevent
```
SomeoneToIgnore, line 1492: This seems somewhat dangerous, the less code is invoked before this global is established the better.

In the future, we plan to disable more things, e.g. global `NodeRuntime` invocations: what if some `let lsp_store = cx.new` tries to use that interim?
Same-but-different race situation happens here.

Instead, feels better to actually move the related worktree trust request handlers above?

-----------

One general question would be — what is the reason to change this?
Did you find any set of steps to trigger the bug during race?
Or, is it just a hypothetical fix?

PR #45805, opened 2025-12-29
Add `project panel: reload files` command

apoorvkh:
Added a command to reload worktree files shown in the project panel. Analogous to `editor: reload file`.

A manual fix for keeping these up-to-date, while automatic updates are seemingly less reliable (#30851).

Patch generated by Codex. Tests pass locally on MacOS.

Would be great if someone can help me test this feature. I am running into issues building this (on MacOS) with support for remote connections, which is my usual failure case for stale filetrees in the project panel.


SCDevel:
I am no longer at my computer so I cannot test, but I think it will need to be grouped together with `Find in folder...`.

I think the 'is_root' will make it so you have to be at the root of the folders to use it, if its grouped with the find in folders it will be in 'is_dir' and only show up when you right click on a directory.

Again I am not on my computer so I cannot verify but when I was playing around and looking at the code, that seems to be the case.

zelenenka:
@apoorvkh thanks, but i'm going to close this PR. the files not being up to date is annoying of course, but our contributing guidelines ask that folks understand the code they submit, whether it's AI-generated or not, and from your earlier comment it seems to not be the case. if someone wants to take this over, it would be easier for them to start from scratch than to understand and shape up someone else's vibe coding.

the comments by others also suggest this may not even be the right problem to fix.

if you're interested in contributing to Zed, we'd love to have you! but we'd ask that you start with something you can own end-to-end. The ["Contribution ideas" section in CONTRIBUTING.md](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#contribution-ideas) has some pointers.

apoorvkh:
That’s fair (agree this doesn't follow contribution guidelines), but thanks for the input @SCDevel @SomeoneToIgnore !

Fwiw, the point of this PR is to propose a patch for an experienced maintainer to integrate and more easily fix the bug. While one could start from scratch, this bug has been open since May and there aren’t any current efforts towards *properly* fixing it. To some extent, this was actually successful since we were able to quickly gather (away-from-terminal) input from the others above and update accordingly!

The current commit should indeed be more faithful to the goals of this PR (even if different from the "proper" fix). The code looks straightforward (and is not obviously slop or incorrect), so it may well be a good starting place for someone else to adjust and ship. Please do!

Issue #17326, opened 2024-09-03
`log` does not appear to work in extensions


valentinegb:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

Adding the `log` crate to an extension's `Cargo.toml` and using its macros appears to result in no output. Here's what I have in `Cargo.toml`, specifically:

```toml
[dependencies]
log = { version = "0.4.16", features = ["kv_unstable_serde", "serde"] }
zed_extension_api = "0.1.0"
```

In `src/lib.rs` I simply used `log::debug!()` in the `Extension::language_server_command()` declaration. I am running Zed with `RUST_LOG=debug` and I can see other debug messages.

### Environment

Zed: v0.151.1 (Zed Preview)
OS: macOS 15.1.0
Memory: 8 GiB
Architecture: aarch64

### If applicable, attach your Zed.log file to this issue.

<details><summary>Zed.log</summary><pre>

[Zed.log](https://github.com/user-attachments/files/16854307/Zed.log)
</pre></details>

valentinegb:
Having `log` is kind of a must, especially since Zed itself uses `log`. Using `println!()` instead (which does work) makes debugging messages difficult to find as opposed to `debug!()` and appears in Zed's output even if `RUST_LOG` were set to `none`.

valentinegb:
Yes, this is still an issue.

SomeoneToIgnore:
As I've learned recently, you have to run `zed --foreground` for any of the extension logs to appear.
Can you confirm this?

valentinegb:
`log` messages still don't appear from extensions. To clarify, `log` is a logging library which Zed itself uses and this issue is with the fact that extensions cannot use this same library to produce log messages.

nrayburn-tech:
The `log` crate doesn't actually do any logging itself. You need to provide an actual implementation to go with it https://github.com/rust-lang/log?tab=readme-ov-file#in-executables.

Zed does this for the core application, but it doesn't appear that those same implementations are automatically available for extensions.

The oxc-zed plugin has debug logging working with this approach.
https://github.com/oxc-project/oxc-zed/blob/main/Cargo.toml#L16-L17
https://github.com/oxc-project/oxc-zed/blob/main/src/oxlint.rs#L27
```text
[...]

PR #45810, opened 2025-12-29
extension_api: Add log related methods to the extension API

nrayburn-tech:
If this approach or design isn't desired, I am more than happy to iterate on it and try to get something that provides a consistent experience for users to review extension logs.

This adds new methods to the extension API that provides extensions with a consistent way to perform logging. This shares the same zlog crate used by the rest of the Zed application supporting the same level configuration and scopes.

Ref #17326. Not sure if we should close that issue or not. That issue explicitly requests support for the `log` crate and it's macros, but this takes a different approach.

There are some parts of the code that I'm not entirely sure about, so I'm going to leave some comments there. Any other feedback is also appreciated.

There aren't any tests for this new functionality because I wasn't sure where or how to add them. Any guidance here would be appreciated.

Example usage.
```
log_trace("language_server_workspace_configuration");
log_debug("language_server_workspace_configuration");
log_info("language_server_workspace_configuration");
log_warn("language_server_workspace_configuration");
log_error("language_server_workspace_configuration");
```

Example log output when run with `ZED_LOG=info,extension_host.extension.html=trace cargo run`
image

Release Notes:

- Add log related methods to the extension API.

### DIFF
--- a/crates/extension_api/wit/since_v0.8.0/extension.wit
+++ b/crates/extension_api/wit/since_v0.8.0/extension.wit
@@ -19,6 +19,12 @@ world extension {
     /// Initializes the extension.
     export init-extension: func();
 
+    import log-trace: func(message: string);
+    import log-debug: func(message: string);
+    import log-info: func(message: string);
+    import log-warn: func(message: string);
+    import log-error: func(message: string);
+
     /// The type of a downloaded file.
     enum downloaded-file-type {
         /// A gzipped file (`.gz`).
--- a/crates/extension_host/src/wasm_host/wit/since_v0_8_0.rs
+++ b/crates/extension_host/src/wasm_host/wit/since_v0_8_0.rs
@@ -1108,4 +1108,45 @@ impl ExtensionImports for WasmState {
             .with_context(|| format!("setting permissions for path {path:?}"))
             .to_wasmtime_result()
     }
+
+    async fn log_trace(&mut self, message: String) -> wasmtime::Result<()> {
+        let logger = get_logger(&self.manifest.id);
+        zlog::trace!(logger => "{:?}", message);
+
+        Ok(())
+    }
+
+    async fn log_debug(&mut self, message: String) -> wasmtime::Result<()> {
+        let logger = get_logger(&self.manifest.id);
+        zlog::debug!(logger => "{:?}", message);
+
+        Ok(())
+    }
+
+    async fn log_info(&mut self, message: String) -> wasmtime::Result<()> {
+        let logger = get_logger(&self.manifest.id);
+        zlog::info!(logger => "{:?}", message);
+
+        Ok(())
+    }
+
+    async fn log_warn(&mut self, message: String) -> wasmtime::Result<()> {
+        let logger = get_logger(&self.manifest.id);
+        zlog::warn!(logger => "{:?}", message);
+
+        Ok(())
+    }
+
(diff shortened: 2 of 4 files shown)


crates/extension_host/src/wasm_host/wit/since_v0_8_0.rs
```diff
@@ -1108,4 +1108,45 @@ impl ExtensionImports for WasmState {
...
+        let logger = get_logger(&self.manifest.id);
+        zlog::error!(logger => "{:?}", message);
+
+        Ok(())
+    }
+}
+
+fn get_logger(extension_id: &Arc<str>) -> zlog::Logger {
+    let all_extensions_logger = zlog::scoped!("extension");
+    let current_extension_scope = Box::leak(Box::new(extension_id.to_string()));
```
nrayburn-tech, line 1150: `zlog:scoped!` requires an `&'static str`. This is the only way I could find to get that.

I think that `Box::leak` is fine here because the logger will need to remain until the program terminates in most cases (an exception being if the extension is disabled).

I don't know if calling `Box::leak(Box::new(extension_id.to_string()))` multiple times with the same extension id leaks a single string or if it continues to leak a new one each time. If it's a new one each time, then something about this would need to be changed.

crates/extension_host/src/wasm_host/wit/since_v0_8_0.rs
```diff
@@ -1108,4 +1108,45 @@ impl ExtensionImports for WasmState {
             .with_context(|| format!("setting permissions for path {path:?}"))
             .to_wasmtime_result()
     }
+
+    async fn log_trace(&mut self, message: String) -> wasmtime::Result<()> {
+        let logger = get_logger(&self.manifest.id);
+        zlog::trace!(logger => "{:?}", message);
```
nrayburn-tech, line 1114: As far as I could tell, `zlog` requires a message format here and can't just log the supplied `String` directly. This approach may not be correct, but I'm not sure.

crates/extension_api/wit/since_v0.8.0/extension.wit
```diff
@@ -19,6 +19,12 @@ world extension {
     /// Initializes the extension.
     export init-extension: func();
 
+    import log-trace: func(message: string);
+    import log-debug: func(message: string);
+    import log-info: func(message: string);
+    import log-warn: func(message: string);
+    import log-error: func(message: string);
```
nrayburn-tech, line 26: This exposes functions such as `log_*`. I'm not sure if that's desired or if a different design would be better.

nrayburn-tech, line 26: This just takes a single `String` as the message. Extensions would need to format the message prior to calling the log method. I'm not sure if this is the right design or if a different approach would be better.

maxdeviant:
Wonder if we could make it so extensions are able to use the [`log`](https://crates.io/crates/log) crate and then we route the logs to the right place?

We'd probably still need an extension API for it, but we could have `log` be the interface for it.

maxdeviant:
Going to close this out, for now.

I think that any future explorations should look into how we can hook into the Rust ecosystem to make it feel natural to do logging, as described here:

> Wonder if we could make it so extensions are able to use the [`log`](https://crates.io/crates/log) crate and then we route the logs to the right place?
> 
> We'd probably still need an extension API for it, but we could have `log` be the interface for it.

PR #45812, opened 2025-12-29
Add stream_options.include_usage for OpenAI-compatible API token usage

Oft3r:
## Summary

This PR enables token usage reporting in streaming responses for OpenAI-compatible APIs (OpenAI, xAI/Grok, OpenRouter, etc).

## Problem

Currently, the token counter UI in the Agent Panel doesn't display usage for some OpenAI-compatible providers because they don't return usage data during streaming by default. According to OpenAI's API documentation, the `stream_options.include_usage` parameter must be set to `true` to receive usage statistics in streaming responses.

## Solution

- Added StreamOptions struct with `include_usage` field to the open_ai crate
- Added `stream_options` field to the Request struct
- Automatically set `stream_options: { include_usage: true }` when `stream: true`
- Updated edit_prediction requests with `stream_options: None` (non-streaming)

## Testing

Tested with xAI Grok models - token counter now correctly shows usage after sending a message.

## References

- [OpenAI Chat Completions API - stream_options](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options)
- [xAI API Documentation](https://docs.x.ai/api)

Release Notes:

- openai: Support usage tracking when streaming responses from openai providers


benbrandt (approved):
Thanks, works great!

PR #45815, opened 2025-12-29
editor: Fix cursor hiding when opening Project Search from status bar

cppcoffee:
Release Notes:

- Fixed cursor hiding when opening Project Search from status bar


MrSubidubi:
Thanks for this, but this will cause issues elsewhere and is not the proper fix. We are aiming for a more general fix for this issue in https://github.com/zed-industries/zed/pull/42262.

PR #45829, opened 2025-12-29
Support multiple snippets locations per extension

MrSubidubi:
This allows extensions to add more than one snippet file whilst keeping it backwards compatible.

Release Notes:

- Added support for specifying multiple snippets paths in extensions.

### DIFF
--- a/crates/extension/src/extension_manifest.rs
+++ b/crates/extension/src/extension_manifest.rs
@@ -53,6 +53,8 @@ impl SchemaVersion {
     }
 }
 
+// TODO: We should change this to just always be a Vec<PathBuf> once we bump the
+// extension.toml schema version to 2
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum ExtensionSnippets {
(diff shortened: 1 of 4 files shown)


crates/extension/src/extension_manifest.rs
```diff
@@ -53,6 +53,28 @@ impl SchemaVersion {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ExtensionSnippets {
+    Single(PathBuf),
+    Multiple(Vec<PathBuf>),
+}
```
tartarughina, line 63: Why having a Single and a Multiple when everything could be included in the Vec<PathBuf>? Is there any benefit of not using a Vec when there's only a single snippet file?

MrSubidubi: Yes, there currently is, and that is backwards compatibility for older Zed versions. We'd keep that for extensions pulished with the old format and they would continue to work in older versions. It still is not optimal, because newer extensions with the new format would not work in older versions but we'd try to load these, so I'll need to think about whether there might be a better approach and/or whether it's worth it to keep that around.

osiewicz (approved):

PR #45838, opened 2025-12-29
feat: Add Claude Agent SDK integration with tool support

fvegiard:


maxdeviant:
Thank you for the PR, but we already integrate with Claude Code using ACP.

This seems like a one-off way of achieving roughly the same thing.

PR #45846, opened 2025-12-30
Improve Git Panel with TreeView, VSCode-style grouping, commit history, and auto-fetch

minhtrungcc:
Release notes

- **Grouping by Stage**: Files are now grouped into "Staged Changes" and "Changes" sections (VSCode-style), replacing the previous Conflict/Tracked/Untracked grouping. This provides a clearer workflow for staging and committing changes.

- **Commit History Section**: A new collapsible "Commits" section displays recent commit history with compact cards showing commit message, author, and relative timestamp. Clicking a commit opens the commit details view.

- **Auto Fetch**: The Git Panel now automatically fetches from remote repositories in the background at configurable intervals (default: every 5 minutes).

New settings:
- `git_panel.auto_fetch`: Enable/disable background fetching (default: true)
- `git_panel.auto_fetch_interval_secs`: Interval between fetches (default: 300)
- `git_panel.show_commit_history`: Show/hide commit history section (default: true)
- `git_panel.commit_history_count`: Number of commits to display (default: 10)

Screenshot 2025-12-30 at 09 00 26

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -883,6 +883,23 @@
       // Default: inherits editor scrollbar settings
       // "show": null
     },
+    // Whether to automatically fetch from remote repositories in the background.
+    //
+    // Default: true
+    "auto_fetch": true,
+    // Interval in seconds between automatic fetches.
+    // Minimum: 60 seconds. Maximum: 3600 seconds (1 hour).
+    //
+    // Default: 300 (5 minutes)
+    "auto_fetch_interval_secs": 300,
+    // Whether to show the commit history section in the git panel.
+    //
+    // Default: true
+    "show_commit_history": true,
+    // Number of commits to show in the commit history section.
+    //
+    // Default: 10
+    "commit_history_count": 10
   },
   "message_editor": {
     // Whether to automatically replace emoji shortcodes with emoji characters.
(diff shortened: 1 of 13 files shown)


minhtrungcc:
> Does this also view partially-staged files? (i.e., staging specific lines in the file instead of the whole file)

Yes it does. A file can be partially-staged, it will appear on both sections (staged and changes)

Dima-369:
Just a nit, the Git Panel received the tree view in https://github.com/zed-industries/zed/pull/44089. Your PR description mentions `git_panel.tree_view` as a new setting, but it already exists on `main`.

It would also be cool to have key bindings to focus the commit history, like the existing ones `git_panel::FocusChanges` and `git_panel::FocusEditor`. And then move through it via key bindings (maybe that already works though, I have not tested yet).

laneme:
I like the vscode style partial changes implemented but i suppose this doesn't build on top of #44089, which is already pending release?

lishaduck:
> I like the vscode style partial changes implemented but i suppose this doesn't build on top of #44089, which is already pending release?

I just checked the diff b/c I too was curious here. It looks like this PR is atop of #44089, so it's just the description/title that's incorrect.

cole-miller:
Hi, thanks for opening a PR! However, this groups together too many unrelated changes for us to review in its current form.

We're interested in merging the staged/unstaged VS Code-style mode for the git panel in some form, but it should be an optional mode (like the existing tree view). If you'd like to open a PR for just that we can take a look.

andrews05:
@minhtrungcc Did you end up making new PRs for this? Really keen to get auto fetch (+prune).

PR #45853, opened 2025-12-30
Convert Titlebar to use PopoverMenus instead of Modals

srimanachanta:
Extends upon the changes from #45361. It replaces the modal-based interactions in the title bar (Project switching, Branch switching, and Remote connections) with lighter Popover menus.

Previously, clicking the project name or branch name in the title bar triggered a full application modal. While functional, they can feel "clunky" and heavyweight for frequent actions like switching branches or jumping between recent projects.

By moving to Popovers, the UI feels more responsive and contextually anchored to the element the user interacted with, providing a smoother UX for navigation.

Recent Projects (changes specific to PopoverMenu):
- Added a "Projects" list header.
- Included static "Open Local Folder" and "Open Remote Folder" items directly in the list (replacing the footer buttons used in the modal version).
- Adjusted empty state text ("No matching projects" vs "No matches").
- Hides the search editor if there are fewer than 4 workspaces to keep the UI compact.
- Moves the search bar to the bottom (End) for popovers, while keeping it at the top (Start) for modals.

Title Bar:
- Replaced Button triggers that dispatched global actions (like OpenRecent) with PopoverMenu components.
- Updated tooltips and visual styling (added chevrons) to indicate these elements are menus.

image
image
image


danilo-leal:
Hey, thanks for the PR! Even though I like the idea of using popovers for the mouse-based interaction, given they're properly anchored to the trigger and whatnot, I don't think we should streamline the UI just because you triggered it through a mouse click vs. keybinding. Even the Git panel's simpler version of the branch picker that you've reused, I'd love to maybe get rid of it and find a way to have only one UX when interacting with it... I get the idea of a simpler picker but I also question it; it's a higher maintenance burden with questionable UX gain. Also, ideally, in the future, we can the anchor-positioned ability into modal pickers and adjust their UI just so slightly when they're triggered by a mouse click. There are also some other inconsistencies in this PR like the remote projects modal being the same and broken keyboard navigation the project popover. With that said, I'll close this PR for now but thanks again for jumping in!

danilo-leal:
I think the main question here is "quick context switches" vs. "deep searches" — does the intent change when you interact with these pickers with the mouse vs. with the keyboard? Like, is it correct to assume that you want to do something quicker when you're clicking vs. when you're hitting keybindings? I'm unsure it is, and that's why i'm a bit skeptical of simplifying them. The anchoring question seems unrelated to this, though; I'm open to a more robust anchoring implementation that doesn't just rely on the click coordinates (which I highlighted in my PR as being not _the best_ solution, but it does solve the issue of the mouse having to travel a lot if you're reaching these pickers with it).

srimanachanta:
The functional intent (switching projects) is the same regardless of the trigger. However, the interaction model is what changes the design requirements.

The existing selectors were explicitly designed as Modals. They are visually "heavy", with large search bars, button footers, and ample padding, because they were designed to take full attention. When we re-anchor that same UI to the mouse cursor, we solve the physical travel time, but we don't solve the cognitive interruption of having a bloated window.

There is a standard pattern for this: keyboard-centric workflows (like Vim) typically use large modals/command palettes, while mouse-centric workflows (VS Code, JetBrains) use lightweight dropdowns for the same actions. Clicking a breadcrumb creates an expectation of a context menu, not a detached window.

The goal of the simplification was to align the visual weight of the component with its container. If we put a heavy Modal design inside a Popover anchor, we ultimately haven't streamlined the flow, just made it easier to access with the mouse. The simplified design treats it as a dropdown menu, which feels native to a mouse-click interaction while leaving the keyboard interaction as is.

PR #45881, opened 2025-12-30
python: Add support for `uv run` as the shebang line

flou:
`uv run` can be used to specify a shebang for a Python script.

See: https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file

Release Notes:

- Added support for `uv run` in the shebang line to be interpreted as a Python script


osiewicz:
Thanks!

PR #45882, opened 2025-12-30
Add back individual language server controls to status bar menu

danilo-leal:
When we first introduced this menu to the status bar, we had it with individual controls per language server but ended up taking them out mainly because the keyboard navigation story wasn't great. Now that we have submenus, I think we can bring them back with a better UX, both for pointer and keyboard.

https://github.com/user-attachments/assets/6bb5f554-a10e-4fed-a3fb-87b58358092a

cc @SomeoneToIgnore for any thoughts and review! I brought most of the code we had (around 6 months ago!) back and adapted what was needed for this to work now.

Release Notes:

- Added the ability to restart and stop individual language servers from the UI as opposed to just them all.

### DIFF
--- a/crates/language_tools/src/lsp_button.rs
+++ b/crates/language_tools/src/lsp_button.rs
@@ -17,8 +17,7 @@ use project::{
 };
 use settings::{Settings as _, SettingsStore};
 use ui::{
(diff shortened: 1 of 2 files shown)


libondev:
Doesn't seem to work very well when the error message is too long

https://github.com/user-attachments/assets/bb33c607-53e8-4911-8428-16b5965ac032

PR #45892, opened 2025-12-31
editor: Add alignment setting to code completion context menu detail text

### LINKED ISSUE #5154: Documentation label right alignment in completion menu
### Check for existing issues

- [X] Completed

### Is your feature request related to a problem?

personally I find a bit hard to skim through the completion suggestions as the name of the symbol and it's type or signature are one next to each other. yep, there is a contrast difference, as the type is dimmed out, but that's not enough to make me as fast as in other editors

I think that just right-aligning the type/signature info to separate it from the name would make a big difference, and that approach seems to be used by other editors as well

current zed ui
Screenshot 2022-11-01 at 20 26 31

suggested change
Screenshot 2022-11-01 at 20 26 31

reference from goland (intellij platform)
Screenshot 2022-11-01 at 20 38 47

_note how deprecated functions are also greyed out and with a strikethrough, which is really nice_

justsomelegs:
Closes #5154

Release Notes:

- Added a setting (`completion_detail_alignment`) to change the detail text alignment in code completion context menus.

<table>
<tr>
 <td>Right Alignment
 <td>Left Alignment
<tr>
 <td>
image
 <td>
image
</table>


danilo-leal (approved):
Thank you!

PR #45341, opened 2025-12-19
vim: Change logic in subword motion subword_start calculation

VVill-ga:
Closes #23344

Previous logic would consume a subword until it left a non-alphanumeric character (consuming any punctuation). With the new logic, the only punctuation we consume is `_` for `snake_case`, and should immediately exit after leaving - consuming only trailing underscores. This new behavior resolves the inconsistencies pointed out in the issue above, and matches what I've experienced with the CamelCaseMotion vim extension.

[Demonstration of issue examples behavior resolved](https://github.com/user-attachments/assets/d877eaff-6b65-4ca7-87c6-9dba4f850a90)

[Walking through subwords](https://github.com/user-attachments/assets/b7e5694a-6bc1-4a19-8c8e-e9ddb06e8fec)

Release Notes:

- Fixed vim mode's subword motions to properly stop at punctuation and handle `snake_case`, `camelCase`, and `dot.separated` identifiers consistently, matching vim's CamelCaseMotion behavior

PR #45908, opened 2025-12-31
vim: Fix subword motion near end of line

### LINKED ISSUE #17780: Subword motion ignores line breaks, jumps to next line unexpectedly
## Description
The Subword motion feature is not working as expected when it encounters end-of-line characters. Instead of treating the end of the line as a boundary, it immediately jumps to the subword end of the next word on the line below.

## Steps to Reproduce
1. Open a file in Zed editor
2. Place the cursor at the beginning of the last subword in a line
3. Use the Subword motion command to move forward

## Expected Behavior
The cursor should move to the end of the current word and stop at the end of the line.

## Actual Behavior
The cursor skips the end-of-line character(s) and jumps to the subword end of the first word on the next line.

### Environment

Zed: v0.152.3 (Zed)
OS: Linux Wayland nixos 24.11
Memory: 30.6 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV REMBRANDT) || radv || Mesa 24.2.2

<details><summary>video</summary>

https://github.com/user-attachments/assets/b8d3741b-2c2e-456e-85c2-ad306981bb4b

</details>

<details><summary>screenshots</summary>

2024-09-13-092026_grim
2024-09-13-092037_grim
2024-09-13-092044_grim

</details>

lionel-:
Closes #17780

The main issue that is fixed here: the boundary detection logic currently excludes newlines due to `!at_newline`:

```rust
let found = !left.is_whitespace() && !at_newline && (is_word_end || is_subword_end);
```

But removing `&& !at_newline` reveals a secondary issue with the backtracking implemented in the subword motion:

```rust
let mut new_point = map.clip_point(new_point, Bias::Left);
if need_backtrack {
    *new_point.column_mut() -= 1;
}
```

If `new_point` represents the newline, an invalid position, clipping the point with a left bias causes a one-character backtrack to match a valid position. Then in the next statement we backtrack again. Because of this the cursor no longer moves once it reaches a newline.

Here the fix is to use the same approach as in the regular word motion: find an exclusive boundary instead of an inclusive one in the first place, then we don't need backtracking. The `need_next_char` logic, also copied from the regular motion implementation, is needed to ensure that motions that don't allow crossing line breaks still make some progress at newlines.

This whole fix is self-contained in the first commit. The second commit is a refactor that makes both motions use the same implementation (with an `is_boundary` predicate), since they are now so close. This way it'll be easier to keep them in sync.

Release Notes:

- Fixed subword motion incorrectly jumping to the next line when near the end of a line


dinocosta:
Holding off on reviewing this until https://github.com/zed-industries/zed/pull/45341#pullrequestreview-3634508597 is resolved, as that might impact these changes 🔍

PR #45919, opened 2025-12-31
docs: Remove reference to non-existing binding for `tab switcher: toggle all`

hokein:
There is no default binding for `{#kb tab_switcher::ToggleAll}`, so the doc is rendered as:

image

Release Notes:

- N/A


MrSubidubi (approved):
Nice find, thanks!

PR #45920, opened 2025-12-31
docs: Use inline code in `debugger.md`

hokein:
Release Notes:

- N/A


MrSubidubi (approved):
Nice catch again, thank you!

Small nit, but it would help a bit if you were to open these as batches as opposed to one PR for each minor fix as this would save a bit on resources for CI (as well as reviewing of sorts). Still appreciate you doing this, just a minor note should you find and submit anything else. Thanks!

PR #45924, opened 2026-01-01
title_bar: Use popovers for modal anchoring

srimanachanta:
Anchoring fix part of #45853

Before:

https://github.com/user-attachments/assets/abe66049-04e5-4f33-9efb-2e99e63e4cc8

After:

https://github.com/user-attachments/assets/14533bbe-8f46-43b1-9465-11e8d313561f

Release Notes:

- N/A


danilo-leal (approved):
Thanks for the follow up! I've pushed a few commits cleaning up/simplifying this to the core change: making the modal use popovers when triggered through by a pointer. Slightly unsure whether the added dependencies are worth it, but I suppose it's what it takes for now... in any case, thanks again!

PR #45932, opened 2026-01-01
Add dynamic tasks and task-context to extension API

### LINKED ISSUE #17462: Allow extensions to dynamically provide tasks derived/discovered in project worktree
### Describe the feature

Tasks defined in a JSON file are great. However, there are a number of common tools that already define tasks. It would be great to simply import those tasks and expose them to Zed when a project is opened. Think Makefile, Taskfile, etc.

Prior art: https://extensions.panic.com/extensions/littlegreenman/littlegreenman.TaskFinder/

playdohface:
Closes #17462 
Relates to https://github.com/zed-extensions/java/issues/94

This is a proposal for allowing extensions to dynamically provide tasks and task variables. The main intended purpose is to be able to reliably bridge different build tools and task runners (Make, Just, package.json, mvn, gradle etc.) with the Zed tasks and runnables system. 

## How to test: 
I work on the Java extension so I was testing with this fork of it: https://github.com/playdohface/zed-java/tree/test-dynamic-tasks which has some simple implementations of the methods. The main thing to do is point the the `extension_api` crate in the extension at this fork (`zed_extension_api = { git = "https://github.com/playdohface/zed.git", branch = "extension-context-provider-for-tasks" }
` and implement the two methods, and play around with it using the tasks ui. 

## How it works and why this way
It does so by adding two methods to the `Extension` trait in the `extension_api`: 
`build_context` and `associated_tasks`. 
`build_context` provides environment variables to tasks (visible like tree-sitter captures e.g. `$ZED_CUSTOM_MY_DYNAMIC_VAR`). 
`associated_tasks` provides task-templates based on the current file and project.
Both have access to the the worktree, meaning they can check whether an executable (e.g. `make`) is available, and they can read files (like `Makefile` or `package.json`) to intelligently dispatch tasks to existing tools based on availability and project configuration. 

This is inspired by how similar functionality is already achieved with built-in languages, and while originally I was looking to solve the linked issue in the java extension (where cross-platform, cross-build-tool runnables are not possible), this also opens up the possibility for e.g. Makefile or Justfile extensions to dynamically provide tasks and runnables based on the available targets. 

Release Notes:

- Improved the Extension API by adding dynamic tasks and task-variables

### DIFF
--- a/crates/extension/src/extension.rs
+++ b/crates/extension/src/extension.rs
@@ -168,6 +168,20 @@ pub trait Extension: Send + Sync + 'static {
         locator_name: String,
         config: SpawnInTerminal,
     ) -> Result<DebugRequest>;
+
+    async fn build_context(
+        &self,
+        language_name: String,
+        location: TaskContextLocation,
+        worktree: Arc<dyn WorktreeDelegate>,
+    ) -> Result<Vec<TaskVariable>>;
+
+    async fn associated_tasks(
+        &self,
+        language_name: String,
+        file: Option<TaskContextFile>,
+        worktree: Arc<dyn WorktreeDelegate>,
+    ) -> Result<Vec<TaskDefinition>>;
 }
 
 pub fn parse_wasm_extension_version(extension_id: &str, wasm_bytes: &[u8]) -> Result<Version> {
--- a/crates/extension/src/extension.rs
+++ b/crates/extension/src/extension.rs
@@ -179,6 +179,22 @@ pub trait Extension: Send + Sync + 'static {
         locator_name: String,
         config: SpawnInTerminal,
     ) -> Result<DebugRequest>;
+
+    async fn build_context(
+        &self,
+        language_name: String,
+        variables: task::TaskVariables,
+        project_env: Option<collections::HashMap<String, String>>,
+        location: TaskContextLocation,
+        worktree: Arc<dyn WorktreeDelegate>,
+    ) -> Result<task::TaskVariables>;
+
+    async fn associated_tasks(
+        &self,
+        language_name: String,
+        file: Option<TaskContextFile>,
+        worktree: Arc<dyn WorktreeDelegate>,
+    ) -> Result<Vec<task::TaskTemplate>>;
 }
 
 pub fn parse_wasm_extension_version(extension_id: &str, wasm_bytes: &[u8]) -> Result<Version> {
--- a/crates/extension/src/types/tasks.rs
+++ b/crates/extension/src/types/tasks.rs
@@ -141,7 +141,7 @@ impl language::ContextProvider for ExtensionContextProvider {
             path: file.path().as_std_path().to_string_lossy().to_string(),
         });
 
-        cx.spawn(async move |cx| {
+        cx.spawn(async move |cx: &mut gpui::AsyncApp| {
             let mut templates = static_templates.unwrap_or_default();
             let worktree_delegate = if let Some(worktree_id) = worktree_id {
                 cx.update(|cx| proxy.worktree_delegate(worktree_id, cx))
--- a/crates/extension/src/types/tasks.rs
+++ b/crates/extension/src/types/tasks.rs
@@ -1,0 +1,175 @@
+use crate::ExtensionWorktreeProxy;
+use anyhow::anyhow;
+use gpui::AppContext;
+use std::sync::Arc;
+
+use language::ToOffset;
+use task::{HideStrategy, RevealStrategy, RevealTarget, Shell, TaskTemplate};
+use util::ResultExt as _;
+
+#[derive(Debug, Clone)]
+pub struct TaskContextLocation {
+    pub worktree_id: u64,
+    /// Path relative to the worktree root.
+    pub file_path: String,
+    pub range: std::ops::Range<usize>,
+}
+
+#[derive(Debug, Clone)]
+pub struct TaskContextFile {
+    pub worktree_id: u64,
+    /// Path relative to the worktree root.
+    pub path: String,
+}
+
+#[derive(Debug, Clone)]
+pub struct TaskVariable {
+    pub name: String,
+    pub value: String,
+}
+
+#[derive(Debug, Clone)]
+pub struct TaskDefinition {
+    pub label: String,
+    pub command: String,
+    pub args: Vec<String>,
+    pub env: Vec<(String, String)>,
+    pub cwd: Option<String>,
+    pub use_new_terminal: Option<bool>,
+    pub allow_concurrent_runs: Option<bool>,
+    pub reveal: Option<RevealStrategy>,
+    pub reveal_target: Option<RevealTarget>,
+    pub hide: Option<HideStrategy>,
+    pub shell: Option<Shell>,
+    pub show_summary: Option<bool>,
+    pub show_command: Option<bool>,
+    pub tags: Vec<String>,
+}
+
+impl From<TaskDefinition> for TaskTemplate {
+    fn from(definition: TaskDefinition) -> Self {
+        Self {
+            label: definition.label,
+            command: definition.command,
+            args: definition.args,
+            env: definition.env.into_iter().collect(),
+            cwd: definition.cwd,
+            use_new_terminal: definition.use_new_terminal.unwrap_or_default(),
+            allow_concurrent_runs: definition.allow_concurrent_runs.unwrap_or_default(),
+            reveal: definition.reveal.unwrap_or_default(),
+            reveal_target: definition.reveal_target.unwrap_or_default(),
+            hide: definition.hide.unwrap_or_default(),
+            shell: definition.shell.unwrap_or_default(),
+            show_summary: definition.show_summary.unwrap_or(true),
+            show_command: definition.show_command.unwrap_or(true),
+            tags: definition.tags,
+        }
+    }
+}
+
+pub struct ExtensionContextProvider {
+    pub extension_id: Arc<str>,
+    pub language_name: language::LanguageName,
+    pub static_templates: Option<task::TaskTemplates>,
+}
+
+impl language::ContextProvider for ExtensionContextProvider {
+    fn build_context(
+        &self,
+        _variables: &task::TaskVariables,
+        location: language::ContextLocation<'_>,
+        _project_env: Option<collections::HashMap<String, String>>,
+        _toolchains: Arc<dyn language::LanguageToolchainStore>,
+        cx: &mut gpui::App,
+    ) -> gpui::Task<anyhow::Result<task::TaskVariables>> {
+        let extension_id = self.extension_id.clone();
+        let language_name = self.language_name.clone();
+        let proxy = crate::ExtensionHostProxy::global(cx);
+        let buffer = location.file_location.buffer.read(cx);
+        let file = buffer.file();
+        let worktree_id = file.map(|f| f.worktree_id(cx).to_proto()).unwrap_or(0);
+        let file_path = file
+            .map(|f| f.path().as_std_path().to_string_lossy().to_string())
+            .unwrap_or_default();
+        let snapshot = buffer.text_snapshot();
+        let range = location.file_location.range.start.to_offset(&snapshot)
+            ..location.file_location.range.end.to_offset(&snapshot);
+
+        let location = TaskContextLocation {
+            worktree_id,
+            file_path,
+            range,
+        };
+
+        let worktree_delegate = proxy.worktree_delegate(worktree_id, cx);
+
+        cx.background_spawn(async move {
+            let worktree_delegate =
+                worktree_delegate.ok_or_else(|| anyhow!("worktree delegate not found"))?;
+
+            let extension = proxy
+                .extension_by_id(&extension_id)
+                .ok_or_else(|| anyhow!("extension not found"))?;
+            let variables = extension
+                .build_context(language_name.to_string(), location, worktree_delegate)
+                .await?;
+            let mut result = task::TaskVariables::default();
+            for variable in variables {
+                result.insert(
+                    task::VariableName::Custom(variable.name.into()),
+                    variable.value,
+                );
+            }
+
+            Ok(result)
+        })
+    }
+
+    fn associated_tasks(
+        &self,
+        file: Option<Arc<dyn language::File>>,
+        cx: &gpui::App,
+    ) -> gpui::Task<Option<task::TaskTemplates>> {
+        let extension_id = self.extension_id.clone();
+        let language_name = self.language_name.clone();
+        let static_templates = self.static_templates.clone();
+        let proxy = crate::ExtensionHostProxy::global(cx);
+
+        let worktree_id = file.as_ref().map(|file| file.worktree_id(cx).to_proto());
+        let file = file.map(|file| TaskContextFile {
+            worktree_id: worktree_id.unwrap_or(0),
+            path: file.path().as_std_path().to_string_lossy().to_string(),
+        });
+
+        cx.spawn(async move |cx| {
+            let mut templates = static_templates.unwrap_or_default();
+            let worktree_delegate = if let Some(worktree_id) = worktree_id {
+                cx.update(|cx| proxy.worktree_delegate(worktree_id, cx))
+                    .ok()
+                    .flatten()
+            } else {
+                None
+            };
+
+            if let (Some(extension), Some(worktree_delegate)) =
+                (proxy.extension_by_id(&extension_id), worktree_delegate)
+            {
+                if let Some(definitions) = extension
+                    .associated_tasks(language_name.to_string(), file, worktree_delegate)
+                    .await
+                    .log_err()
+                {
+                    templates
+                        .0
+                        .extend(definitions.into_iter().map(task::TaskTemplate::from));
+                }
+            }
+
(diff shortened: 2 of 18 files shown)


crates/extension/src/extension.rs
```diff
@@ -39,6 +39,35 @@ pub trait WorktreeDelegate: Send + Sync + 'static {
     async fn shell_env(&self) -> Vec<(String, String)>;
 }
 
+/// An adapter that allows an [`language::LspAdapterDelegate`] to be used as a [`WorktreeDelegate`].
+pub struct WorktreeDelegateAdapter(pub Arc<dyn language::LspAdapterDelegate>);
```
maxdeviant: We don't want to move this `WorktreeDelegateAdapter` in here. It would be better to just duplicate it again.

The fact that it's backed by an `LspAdapterDelegate` is due to legacy design decisions, but it's not something we necessarily want to promote.

crates/extension/src/types/tasks.rs
```diff
@@ -0,0 +1,171 @@
...
+                .build_context(language_name.to_string(), location, worktree_delegate)
+                .await?;
+            let mut result = task::TaskVariables::default();
+            for variable in variables {
+                result.insert(
+                    task::VariableName::Custom(variable.name.into()),
+                    variable.value,
+                );
+            }
+            Ok(result)
```
osiewicz: This part can happen within `cx.background_spawn`.

crates/extension/src/extension.rs
playdohface: I have to say I find getting hold of a WorktreeDelegate-implementor (or constructing one) quite awkward and confusing. Is there an architecturally better way to do it that I'm just not seeing?

ofek:
@playdohface Hey there! Are you able to rebase to resolve the conflicts so the review can commence?

playdohface:
@ofek hey! So I had to get rid of my local fork of zed because it took up most of my hard drive 💾  
The merge conflicts should all be very straightforward to resolve and I doubt that's the blocker here, but the real question is whether the overall approach is the right one. Changes to the extension API are understandably not something that will be merged on a whim, and I'd like some more input from the Zed team before going deeper into this.

That being said, if I see green light for a given approach I'm happy to do any remaining footwork, I can make room on my hard drive any time :)

crates/extension_api/wit/since_v0.8.0/task.wit
osiewicz, line 1: We've already exposed parts of Task types within DAP: https://github.com/zed-industries/zed/blob/d71f1461045c098dc6ca6b1b5adcf1b8949722e8/crates/extension_api/wit/since_v0.8.0/dap.wit#L55. Maybe we could move types around, add fields and what not to not repeat ourselves?

crates/extension_host/src/extension_host.rs
```diff
@@ -1273,13 +1300,19 @@ impl ExtensionStore {
...
         for (language_name, language) in languages_to_add {
             let mut language_path = self.installed_dir.clone();
             language_path.extend([
                 Path::new(language.extension.as_ref()),
                 language.path.as_path(),
             ]);
+            let extension_id = language.extension.clone();
+            // let grammar = language.grammar.clone();
+            // let matcher = language.matcher.clone();
+            // let hidden = language.hidden;
```
osiewicz:
```suggestion
```

crates/extension/src/types/tasks.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub use_new_terminal: Option<bool>,
+    pub allow_concurrent_runs: Option<bool>,
+    pub reveal: Option<RevealStrategy>,
+    pub reveal_target: Option<RevealTarget>,
+    pub hide: Option<HideStrategy>,
+    pub shell: Option<Shell>,
+    pub show_summary: Option<bool>,
+    pub show_command: Option<bool>,
+    pub tags: Vec<String>,
+}
```
osiewicz: Why do we need to duplicate the types from task library here?

crates/project/src/project.rs
```diff
@@ -159,6 +160,76 @@ pub use lsp_store::{
...
+                if let Some(worktree) = worktree {
+                    let delegate = lsp_store.update(cx, |lsp_store, cx| {
+                        lsp_store.lsp_adapter_delegate(worktree, cx)
+                    });
+                    if let Some(delegate) = delegate {
+                        return Some(Arc::new(WorktreeDelegateAdapter(delegate)));
+                    }
+                }
+            }
+        }
```
osiewicz, line 263: You could prolly remove `project_handle`s within this loop that we've failed to upgrade.

crates/project/src/project.rs
```diff
@@ -159,6 +160,76 @@ pub use lsp_store::{
...
+        self.0
+            .which(binary_name.as_ref())
+            .await
+            .map(|path| path.to_string_lossy().into_owned())
+    }
+
+    async fn shell_env(&self) -> Vec<(String, String)> {
+        self.0.shell_env().await.into_iter().collect()
+    }
+}
```
osiewicz, line 220: This seems to have been copied over verbatim from `crates/language_extension/src/extension_lsp_adapter.rs`; can we restructure our code so that we don't need to copy-paste stuff around?

crates/extension/src/types/tasks.rs
```diff
@@ -0,0 +1,175 @@
...
+        _project_env: Option<collections::HashMap<String, String>>,
+        _toolchains: Arc<dyn language::LanguageToolchainStore>,
+        cx: &mut gpui::App,
+    ) -> gpui::Task<anyhow::Result<task::TaskVariables>> {
+        let extension_id = self.extension_id.clone();
+        let language_name = self.language_name.clone();
+        let proxy = crate::ExtensionHostProxy::global(cx);
+        let buffer = location.file_location.buffer.read(cx);
+        let file = buffer.file();
+        let worktree_id = file.map(|f| f.worktree_id(cx).to_proto()).unwrap_or(0);
```
osiewicz: We should *definitely not* do an `unwrap_or(0)` here. If there's no worktree on a given buffer, we ought to just return a task that returns an error.

crates/extension/src/types/tasks.rs
```diff
@@ -0,0 +1,175 @@
...
+        let file_path = file
+            .map(|f| f.path().as_std_path().to_string_lossy().to_string())
+            .unwrap_or_default();
```
osiewicz: See: https://github.com/zed-industries/zed/pull/45932/changes#r3795316440
Again, we should not be returning defaults when we really care about the meaning of a value.

crates/extension_api/wit/since_v0.8.0/task.wit
```diff
@@ -0,0 +1,54 @@
+interface task {
+    use common.{range, env-vars};
+
+    record task-context-location {
+        worktree-id: u64,
+        /// Path relative to the worktree root.
+        file-path: string,
```
osiewicz, line 7: On native side `file-path` is optional. Plus, we don't even carry the worktree id around. Do you recall why WASM side is divergent?

crates/extension/src/types/tasks.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub static_templates: Option<task::TaskTemplates>,
+}
+
+impl language::ContextProvider for ExtensionContextProvider {
+    fn build_context(
+        &self,
+        _variables: &task::TaskVariables,
+        location: language::ContextLocation<'_>,
+        _project_env: Option<collections::HashMap<String, String>>,
+        _toolchains: Arc<dyn language::LanguageToolchainStore>,
```
osiewicz: I don't believe we should be ignoring `variables`, `project_env` and `toolchains` (although toolchains are a bit tricky).

crates/extension_api/src/extension_api.rs
```diff
@@ -175,7 +179,27 @@ pub trait Extension: Send + Sync {
         Ok(None)
     }
 
-    /// Returns a list of package names as suggestions to be included in the
+    /// Returns custom task variables for a given location in a file.
+    fn build_context(
+        &mut self,
+        _language_name: String,
+        _location: &TaskContextLocation,
+        _worktree: &Worktree,
```
osiewicz, line 217: I am a bit unclear on the Worktree stuff. On one hand I don't like the divergence from the original interface, but on the other we don't have a way to query the FS otherwise. In fact, some of the strength of the original task context API is that it can query arbitrary paths on FS to get more info about your execution environment, which is straight up not permitted with our WASM stack..

crates/project/src/project.rs
osiewicz: Ok nvm, I'm literally giving a competing advice: https://github.com/zed-industries/zed/pull/45932#discussion_r2659073685

crates/extension_api/src/extension_api.rs
playdohface: From the point of view of the extension API, worktree is the primitive that extension devs are used to and it's the next best thing to general FS access, which, as you said, is just not permitted in the WASM context. So personally I'm quite confident that worktree is what you'd want from an extension developers point of view.

crates/extension_api/wit/since_v0.8.0/task.wit
playdohface: We probably could, `task-definition` being a superset of daps `task-template`. But that begs the question, where should the type go then? Not really clear what the "correct" API design should be here, this way at least we don't break dap. But lmk if you have specific ideas how it would be better

crates/extension_api/wit/since_v0.8.0/task.wit
playdohface: Thing is, we need worktree-id and we are getting it through the file. So making file optional here means forcing the extension to handle a case that can't actually happen, at least right now. IIRC native just borrows the buffer which we can't do on wasm. But we could make file-path optional here if you think that future-proofs the API better

crates/project/src/project.rs
playdohface: Done (L241)

crates/extension/src/types/tasks.rs
playdohface: We don't :)

crates/extension/src/types/tasks.rs
playdohface: I added variables and project_env but toolchains is tricky indeed, and would require a bunch of extra API surface. It might make sense to add it now if we know we're going to want it, but I don't really feel qualified/have the bandwidth for designing how that should work.

crates/extension_host/src/wasm_host/wit.rs
```diff
@@ -1285,6 +1285,59 @@ impl Extension {
...
+                    store,
+                    language_name,
+                    &wit_variables,
+                    wit_project_env.as_ref(),
+                    &location,
+                    worktree,
+                )
+                .await
+            }
+            _ => Ok(Ok(Vec::new())),
```
MrSubidubi: Lets make the match exhaustive here

crates/extension_host/src/wasm_host/wit.rs
```diff
@@ -1285,6 +1285,59 @@ impl Extension {
...
+        language_name: &str,
+        file: Option<latest::TaskContextFile>,
+        worktree: Resource<latest::ExtensionWorktree>,
+    ) -> wasmtime::Result<Result<Vec<latest::TaskDefinition>, String>> {
+        match self {
+            Extension::V0_8_0(ext) => {
+                ext.call_associated_tasks(store, language_name, file.as_ref(), worktree)
+                    .await
+            }
+            _ => Ok(Ok(Vec::new())),
```
MrSubidubi: Same here

crates/extension_host/src/extension_host.rs
```diff
@@ -1515,19 +1544,32 @@ impl ExtensionStore {
                 Arc::new({
                     let fs = self.fs.clone();
                     let query_files = language.query_files;
+                    let extension_id = extension_id.clone();
+                    let language_name = language_name_for_loader;
```
MrSubidubi: Why do we have to clone the ID twice? Also, the name looks off, can we not just use `language_name` throughout?

crates/extension_host/src/extension_host.rs
```diff
@@ -1693,7 +1735,11 @@ impl ExtensionStore {
                     }
                 }
 
-                this.wasm_extensions.extend(wasm_extensions);
+                this.wasm_extensions.extend(wasm_extensions.clone());
+                this.extension_provider
+                    .wasm_extensions
+                    .write()
+                    .extend(wasm_extensions.into_iter().map(|(m, e)| (m.id.clone(), e)));
```
MrSubidubi: Nit but we usually go for full names and not one-letter abbreviations

crates/extension_host/src/wasm_host.rs
```diff
@@ -530,6 +532,61 @@ impl extension::Extension for WasmExtension {
...
+                        location.into(),
+                        resource,
+                    )
+                    .await?
+                    .map_err(|err| store.data().extension_error(err))?;
+
+                let mut result = TaskVariables::default();
+                for variable in variables {
+                    result.insert(VariableName::Custom(variable.name.into()), variable.value);
+                }
```
MrSubidubi: Another nit but we could rewrite this to be something into_iter().map(...).collect() which would give us the advantage of size hints compared to doing repeated insert calls.

crates/extension_api/src/extension_api.rs
```diff
@@ -197,7 +201,35 @@ pub trait Extension: Send + Sync {
...
+    /// Returns custom task variables for a given location in a file.
+    ///
+    /// `variables` contains the task variables resolved so far and the names in the list
+    /// this method returns will be prefixed with `ZED_CUSTOM_` and then merged with it.
+    /// Note that if this returns variables by the same name as tree-sitter captures, (already
+    /// seen with `ZED_CUSTOM_` prefix here) it will override their values.
+
+    fn build_context(
+        &mut self,
+        _language_name: String,
```
MrSubidubi: Another nit, but we usually newtype this and pass by ref if the extension does not need it owned - would there be a disadvantage from this here? Not insisting, but curious nonetheless

crates/extension/src/types/tasks.rs
```diff
@@ -0,0 +1,129 @@
...
+        cx.spawn(async move |cx: &mut gpui::AsyncApp| {
+            let mut templates = static_templates.unwrap_or_default();
+            let worktree_delegate = if let Some(worktree_id) = worktree_id {
+                cx.update(|cx| proxy.worktree_delegate(worktree_id, cx))
+            } else {
+                None
+            };
+
+            if let (Some(extension), Some(worktree_delegate)) =
+                (proxy.extension_by_id(&extension_id), worktree_delegate)
```
MrSubidubi: Nit but we can `zip` this here

crates/extension_host/src/extension_host.rs
```diff
@@ -2338,25 +2386,28 @@ async fn load_plugin_language(
...
         async move { Ok(load_plugin_queries(fs, &language_path, query_files).await) },
-        async move { Ok(context_provider.await) }
+        async move { Ok(static_templates.await) }
     )?;
 
+    let context_provider = Some(Arc::new(ExtensionContextProvider {
+        extension_id,
+        language_name,
+        static_templates,
+    }) as Arc<dyn language::ContextProvider>);
```
MrSubidubi, line 2407: What I do not like about this is that we have this context provider now for _all_ language extensions, not just these ones that actually use this API.

I generally want to move to a place where the extension declares what it wants to do in some explicit way through either the extension.toml or in this case the config.toml for a given language (have the feeling the latter might make a bit more sense). This makes things a lot easier to review, but also saves us a bunch of calls into the wasm world - we do not need to ask the extension for templates if there are none. 

That for me begs the question (and I am happy if piotr wants to yell at me for this)
- do we want to declare this in either of the configs, and
- do we perhaps even want to specify the list of supported labels for the task templates somewhere and compare those? Aside from helping me with reviewing, we could use this at some other point to allow for trying to spawn a dedicated task at a keybind for example.

Just two thoughts, and more food for discussion here, although I lean to at the very least the first one.

MrSubidubi (changes requested):

crates/extension_host/src/extension_host.rs
playdohface: good catch we can save a bunch of clones there actually

crates/extension_api/src/extension_api.rs
playdohface: good shout I think there is no disadvantage at all and it makes the API contract more obvious to pass everything by ref. `LanguageName` I could take or leave but getting the other ones owned may tempt people to try and mutate and return them, which is not really intended here they are purely informational.

crates/extension_host/src/extension_host.rs
playdohface: My 2c on this topic: probably a good idea to make extensions declare what they provide as much as possible. Idk to what extent this is possible at the moment or planned, but the tasks thing could be used to have extensions that are not necessarily tied to a single language, and if that happens there needs to be some way to disambiguate and prioritize (e.g. you might have a `make` extension that turns makefiles into tasks for you, but if the C extension also does that you'd probably want the latter to win out. Having it declared somewhere probably makes that a lot less hairy)

PR #45936, opened 2026-01-01
editor: Fix `fade_out` styling for completion labels without highlights

aviatesk:
LSP's `CompletionItemKind` is defined by the protocol and may not have a corresponding highlight name in a language's `highlights.scm`. For example, Julia's grammar defines `@function.call` but not `@function`, so completions with `Method` kind cannot resolve a highlight. Since the mapping from `CompletionItemKind` to highlight names is an internal implementation detail that is not guaranteed to be stable, the fallback behavior should provide consistent styling regardless of grammar definitions.

Bundled language extensions (e.g., Rust, TypeScript) apply `fade_out` or muted styling to the description portion of completion labels, so the fallback path and extension-provided labels should match this behavior.

Previously, the description portion could fail to receive the expected `fade_out` styling in several independent cases:

1. When `runs` was empty (grammar lacks the highlight name), the `flat_map` loop never executed
2. When theme lacked a style for the `highlight_id`, early return skipped the `fade_out` logic
3. When extensions used `Literal` spans with `highlight_name: None`, `HighlightId::default()` was still added to `runs`, preventing `fade_out` from being applied

The fix ensures description portions consistently receive `fade_out` styling regardless of whether the label portion can be highlighted, both for fallback completions and extension-provided labels.

Closes #ISSUE

Release Notes:

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -28848,49 +28848,58 @@ pub fn styled_runs_for_code_label<'a>(
         ..Default::default()
     };
 
+    if label.runs.is_empty() {
+        let desc_start = label.filter_range.end;
+        let fade_run =
+            (desc_start < label.text.len()).then(|| (desc_start..label.text.len(), fade_out));
+        return Either::Left(fade_run.into_iter());
+    }
+
     let mut prev_end = label.filter_range.end;
-    label
-        .runs
-        .iter()
-        .enumerate()
-        .flat_map(move |(ix, (range, highlight_id))| {
-            let style = if *highlight_id == language::HighlightId::TABSTOP_INSERT_ID {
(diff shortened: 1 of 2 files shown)


MrSubidubi (commented):
I think code-wise this looks fine (will need to take another look), do you have a real-world example here that shows the difference/makes this easier to repro? Thanks!

MrSubidubi (approved):
Subpar response time from my side all over which I'd like to apologize for. Sorry for this and thank you for the fix here!

I allowed myself to shrink the diff a bit, but all else looks good. Happy to merge this.

Given the wait, I'll not insist on this here, but if you feel like it, I'd appreciate a regression test for this in a follow up. 

Otherwise, thanks for the PR!

PR #45949, opened 2026-01-02
languages: Enable terminal activation for uv-managed environments

lingyaochu:
Currently, the Python toolchain can identify `uv` and `uv-workspace` environments (showing the correct labels in the UI), but it fails to activate them when opening a new terminal. This is because the activation script resolution and command generation logic were missing for these environment kinds.

This PR adds `uv` and `uv-workspace` to the standard virtual environment activation flow. Since `uv` creates environments with a standard `venv` structure, we can reuse the existing `resolve_venv_activation_scripts` logic to find and execute the appropriate `activate` scripts for different shells.

Release Notes:

- Fixed terminal activation for `uv` and `uv-workspace` Python environments.

### DIFF
--- a/crates/languages/src/python.rs
+++ b/crates/languages/src/python.rs
@@ -1357,7 +1357,12 @@ impl ToolchainLister for PythonToolchainProvider {
                     activation_script.push(format!("{manager} activate base"));
                 }
             }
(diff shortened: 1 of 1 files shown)


lingyaochu:
@cla-bot check

Veykril (approved):
Thanks!

PR #45957, opened 2026-01-02
gpui: Add dynamic padding to prevent glyph clipping in text rendering on macOS

JonyLab:
Here may caused by [font-kit](https://github.com/servo/font-kit), I was tried to use [render-glyph](https://github.com/servo/font-kit/blob/main/examples/render-glyph.rs) example and also use the same option with `Transform2F::from_scale(2.0)` for `raster_bounds`. 

Then the `raster_bounds` will result same issue like the GPUI's font rendering issue.

| Transform2F::default() | Transform2F::from_scale(2.) |
| --- | -- |
| image | image |

## Before

SCR-20260105-pgcp-2

## After

SCR-20260105-pfny-1

Release Notes:

- Fixed incorrect rendering of characters at large font sizes on macOS

### DIFF
--- a/crates/gpui/src/platform/mac/text_system.rs
+++ b/crates/gpui/src/platform/mac/text_system.rs
@@ -349,7 +349,7 @@ impl MacTextSystemState {
     fn raster_bounds(&self, params: &RenderGlyphParams) -> Result<Bounds<DevicePixels>> {
         let font = &self.fonts[params.font_id.0];
         let scale = Transform2F::from_scale(params.scale_factor);
-        Ok(font
+        let mut bounds: Bounds<DevicePixels> = font
             .raster_bounds(
                 params.glyph_id.0,
                 params.font_size.into(),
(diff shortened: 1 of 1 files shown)


Dima-369:
Thanks!

Do you have before/after screenshots?

nicoburns:
GPUI may wish to consider switching to https://github.com/servo/webrender/tree/main/wr_glyph_rasterizer which is what Servo and Firefox use to render glyphs these days (like `font-kit` it defers to the system renderers to do the actual rasterization, unlike `font-kit` it is actively maintained). I suspect it can compute correct bounds.

reflectronic (approved):

nicoburns:
> > GPUI may wish to consider switching to https://github.com/servo/webrender/tree/main/wr_glyph_rasterizer
> 
> From a brief look at the code, it looks like wr_glyph_rasterizer bakes the result of the gamma correction into the bitmap. I think we would prefer something that lets us store a single bitmap in the atlas and apply the gamma correction on the GPU. (This is what we have done on Windows and Linux.)

@reflectronic Interesting. Are there use cases where you need the same glyph with different gamma corrections?

Issue #45573, opened 2025-12-23
Syntax highlighting does not work for `.m` and `.mm` files


ZiedYousfi:
### Reproduction steps

Just make a .m or. mm file for Objective-C or Objective-C++. Even though it is supported by ClangD the LSP, it is not considered a supported file extension.

### Current vs. Expected behavior

It should be treated like any C or C++ file because Zed uses clangd out of the box.

### Zed version and system specs

Zed: v0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1 (Zed) 
OS: macOS 26.2
Memory: 36 GiB
Architecture: aarch64

### If you are using WSL on Windows, what flavor of Linux are you using?

None

yeskunall:
Based on what @hokein said in https://github.com/zed-industries/zed/pull/45977#pullrequestreview-3624245053, I’m going to close my PR because my changes don’t make sense. We should instead create dedicated tree-sitter grammar for Objective-C/C++.

PR #45968, opened 2026-01-02
languages: Recognize module interface file as C++ file

hokein:
https://clang.llvm.org/docs/StandardCPlusPlusModules.html#file-name-requirements, `.cppm`, '.ccm', ".cxxm" files are C++20 module interface files.

Release Notes:

- Recognize module interface file as C++ file.


yeskunall:
Hey @hokein as always, thanks for the help. https://github.com/zed-industries/zed/issues/45573 reported a similar issue so I decided to combine your work with mine in #45977 because they are very similar. Hope you don’t mind!

Veykril (approved):

crates/languages/src/cpp/config.toml
```diff
@@ -1,6 +1,6 @@
 name = "C++"
 grammar = "cpp"
-path_suffixes = ["cc", "hh", "cpp", "h", "hpp", "cxx", "hxx", "c++", "h++", "ipp", "inl", "ino", "ixx", "cu", "cuh", "C", "H"]
+path_suffixes = ["cc", "hh", "cpp", "h", "hpp", "cxx", "hxx", "c++", "h++", "ipp", "inl", "ino", "ixx", "cu", "cuh", "C", "H", "cppm", "ccm", "cxxm"]
```
yeskunall: @hokein does it make sense to add `.tpp` files too?

> Can we also add .tpp suffix for C++ ? Thank you,

https://github.com/zed-industries/zed/pull/45977#issuecomment-3713003025

yeskunall: On second thought: what about this list mentioned [here](https://github.com/zed-industries/zed/issues/45980)?

hokein: I think it makes sense to add them. (I would rather do it in a separate PR, as this one only focuses on the module files.)

yeskunall: I mean I’m not against updating the PR title and just squeezing in the rest -- the end goal is greater `file_types` recognition for C++ projects in Zed. But maybe I’m still misunderstanding something.

PR #45971, opened 2026-01-02
Do not eagerly load entire file contents into memory when decoding it

SomeoneToIgnore:
When working on https://github.com/zed-industries/zed/pull/45969 I've noticed that whenever I try to open a binary file

```
~/Desktop/svenska ❯ du -ha "Svenska А2B1. 07.09.2025 [u6qEIe9-COc].mkv"
456M	Svenska А2B1. 07.09.2025 [u6qEIe9-COc].mkv
```

Zed allocates all its size

image

image

only to show me this:

image

Given that our existing code checks first 1024 bytes to decide whether to bail on a binary file or not, this seems very wasteful — hence, adjusted the code to read the "header" and check that first, and only continue reading the entire file after the checks are successful.

I suspect this should also help the project search, esp. the crashes and memory usage during that when many binary files are present?

Release Notes:

- Improved Zed's memory usage when attempting to open binary files

### DIFF
--- a/crates/worktree/src/worktree.rs
+++ b/crates/worktree/src/worktree.rs
@@ -5880,20 +5880,31 @@ async fn decode_file_text(
         .open_sync(&abs_path)
         .await
         .with_context(|| format!("opening file {abs_path:?}"))?;
-    let mut prefix = vec![0u8; FILE_ANALYSIS_BYTES];
-    let prefix_len = file
-        .read(&mut prefix)
-        .with_context(|| format!("reading first bytes of the file {abs_path:?}"))?;
-    prefix.truncate(prefix_len);
 
-    let (byte_content, bom_encoding) = decode_byte_header(&prefix);
+    let mut file_first_bytes = Vec::with_capacity(FILE_ANALYSIS_BYTES);
+    let mut buf = [0u8; FILE_ANALYSIS_BYTES];
+    let mut reached_eof = false;
+    loop {
+        if file_first_bytes.len() >= FILE_ANALYSIS_BYTES {
+            break;
+        }
+        let n = file
+            .read(&mut buf)
+            .with_context(|| format!("reading bytes of the file {abs_path:?}"))?;
+        if n == 0 {
+            reached_eof = true;
+            break;
+        }
+        file_first_bytes.extend_from_slice(&buf[..n]);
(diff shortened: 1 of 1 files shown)


crates/worktree/src/worktree.rs
```diff
@@ -5872,14 +5870,62 @@ impl fs::Watcher for NullWatcher {
...
+    prefix.truncate(prefix_len);
+
+    let (byte_content, bom_encoding) = decode_byte_header(&prefix);
+    anyhow::ensure!(
+        byte_content != ByteContent::Binary,
+        "Binary files are not supported"
+    );
+
+    let mut content = prefix;
+    if prefix_len > 0 {
```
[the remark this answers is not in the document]
SomeoneToIgnore: This is nonsense, but implicitly good as actually points out that we can read less than 1024 bytes in the first `read`.

osiewicz (approved):
LGTM. I don't believe this will help project search though, because we read up to 8kB of file contents before issuing any further reads (via `fill_buf` call: https://github.com/zed-industries/zed/blob/843be766dffee00c5abc37da0706c1b6a1bed5ec/crates/project/src/project_search.rs#L709). Still, nice job. :)

crates/worktree/src/worktree.rs
```diff
@@ -5872,14 +5870,73 @@ impl fs::Watcher for NullWatcher {
...
+) -> Result<(String, &'static Encoding, bool)> {
+    let mut file = fs
+        .open_sync(&abs_path)
+        .await
+        .with_context(|| format!("opening file {abs_path:?}"))?;
+
+    let mut file_first_bytes = Vec::with_capacity(FILE_ANALYSIS_BYTES);
+    let mut buf = [0u8; FILE_ANALYSIS_BYTES];
+    let mut reached_eof = false;
+    loop {
```
osiewicz, line 5889: Actually, I have a small nit: at a glance here it's not obvious to me why we'd like to split the read into two phases. Maybe we could use a small comment here stating the intent? All we need to know is that we don't want to "wastefully" load whole file into memory if it's obviously binary.

SomeoneToIgnore: Added, thank you for checking on that.

PR #45979, opened 2026-01-03
proto: Remove deprecated active view fields from call.proto

cppcoffee:
Release Notes:

- N/A


cppcoffee:
Removed `active_view_id` from `FollowResponse` and `id`/`leader_id` from `UpdateActiveView` in `call.proto`, as they were marked for removal after version 0.145.x.

Updated `workspace.rs` to remove usages of these legacy fields.

SomeoneToIgnore (approved):
Thanks!

PR #21705, opened 2024-12-08
Resolve documentation for visible completions

mgsloan:
Release Notes:

- Improved LSP resolution of documentation for completions. It now queries documentation for visible completions and avoids doing too many redundant queries.

---

In #21286, documentation resolution was made more efficient by only
resolving the current completion. However, this meant that single line
documentation shown inline in the menu was missing until scrolled
to. This also meant that it would wait for navigation to resolve
completion docs, leading to lag for displaying documentation.

This change resolves this by attempting to fetch all the completions
that will be shown. It also mostly avoids re-resolving completions. It
intentionally re-resolves the current selection on navigation, as some
language servers will respond with more information later on.

PR #45985, opened 2026-01-03
lsp: Add `detail` to completion resolvable properties

aviatesk:
Previously, `detail` was not included in resolvable properties, which prevented LSP-compliant servers from lazily resolving it.

In relation to this, I also investigated vtsls issues:
- https://github.com/yioneko/vtsls/issues/213
- https://github.com/zed-industries/zed/pull/21521

My conclusion is that regardless of whether Zed registers `detail` in resolvable properties, vtsls always lazily resolves `detail`. However, I understand that Zed's `resolve_completions` has been improved to handle such "misbehaving" servers gracefully already, so I expect that adding `detail` to resolvable properties will not cause any issues.

In fact, since Zed can lazy update all properties of completion items, it might be appropriate to include all properties as resolvable except for `label` (which is explicitly tested for consistency) and `data` (which is used for resolution itself once).
However, since I'm uncertain if this is entirely correct, this commit only adds `detail` to resolvable properties, as it was required to be supported by LSP clients up to version 3.16.

Closes #ISSUE

Release Notes:

- N/A

### DIFF
--- a/crates/lsp/src/lsp.rs
+++ b/crates/lsp/src/lsp.rs
@@ -761,6 +761,7 @@ impl LanguageServer {
                                 properties: vec![
                                     "additionalTextEdits".to_string(),
                                     "command".to_string(),
+                                    "detail".to_string(),
                                     "documentation".to_string(),
                                     // NB: Do not have this resolved, otherwise Zed becomes slow to complete things
                                     // "textEdit".to_string(),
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (approved):
Than you, indeed, after https://github.com/zed-industries/zed/pull/21705 it seems more or less safe to do.

> In fact, since Zed can lazy update all properties of completion items, it might be appropriate to include all properties as resolvable except for `label` (which is explicitly tested for consistency) and `data` (which is used for resolution itself once).

That phrase is wrong: `data` is not the property you can even consider for the resolution, and just below your change in the `lsp.rs` there's a `NB` comment about `textEdit` not allowed to be resolved — if you check the related PR you can find more context why this is bad to do.

PR #38127, opened 2025-09-14
Implement Vim's tag stack

zenazn:
cc @ConradIrwin @AidanV (thank you for pairing with me!)

Gosh, this turned out to be harder than I thought! But I'm really happy with how things turned out—it's a much better solution than what I had on Tuesday.

I'd recommend reading this commit-by-commit. I tried to leave nice commit messages to make it easier to follow along.

At a high level:
- This ~exactly implements Vim's tag stack behavior, at least as I understand and use it. We're missing (1) a UI, and (2) some stuff like navigating between ambiguous choices, which Zed has a different solve for anyways
- I've integrated it tightly with Zed's navigation history (forwards/back). You can think of navigating to older and newer tags like "super-back" and "super-forward"
- Nothing about this implementation is Vim specific. Anyone can use it!
- Except for the fact that splits now fork navigation (and tag) history, and the exact details of which navigations result in history items, I am not aware of behavior differences for folks who don't use the feature.

One thing I haven't looked into: Making `<C-t>` in Vim use `pane: Go to older tag` instead of `pane: Go back`. 

Closes #14206

Release Notes:
- ??? I don't know what to write here! Suggestions welcome (or Just Fix It™)

PR #46002, opened 2026-01-04
Implement Vim's tag stack

### LINKED ISSUE #14206: Emulate the tag stack in Vim mode
### Check for existing issues

- [X] Completed

### Describe the feature

In Vim, hitting `ctrl-]` takes you to the function definition.  This part already seems to be implemented in Zed.

Each subsequent `ctrl-]` takes you to the corresponding definition, and maintains a stack of where you've been.  Hitting `ctrl-t`  pops the most recent off the stack, and takes you back one step.

This is a very handy way of navigating through multiple function calls in a codebase.  `ctrl-t` seems to be currently set to `editor::Transpose` - is there an action I can set it to that would replicate this behaviour?

See https://vimdoc.sourceforge.net/htmldoc/tagsrch.html for documentation of the Vim behaviour.

_No response_

zenazn:
Happy New Years! This PR is a second take at https://github.com/zed-industries/zed/pull/38127 (cc @ConradIrwin)

This PR is significantly less complicated than the last attempt: while we still keep our data on the `NavigationHistory` object, we no longer tightly integrate it with the existing back/forward "browser history." Instead, we keep our own stack of `(origin, target)` pairs (in a struct to make it easy to extend with e.g., tag names in the future).

The PR is split into two separable commits. Most of the implementation is in the second commit, which:
- Defines the stack data structure
- Implements `pane::GoToOlderTag` and `pane::GoToNewerTag` in terms of the stack
- Hooks into `navigate_to_hover_links` to push tag stack entries

This last bit is the most fiddly. The core challenge is that we need to keep track of the `origin` location and calculate the `target` location across three codepaths that might involve creating a new editor and/or splitting the pane. One thing in particular I found difficult was that an editor's `nav_history` (an `ItemNavHistory`) seems to be populated asynchronously. Instead of relying on it, I decided in this code to make my own `ItemNavHistory`. I briefly tried to refactor the code in question, but it seemed like it would significantly increase the scope of the change.

I prefer this all-in-one tracking centered around `navigate_to_hover_links ` to the `start/finish` approach taken in https://github.com/zed-industries/zed/commit/b69a2ea200561de9de0edf4d0c4428ee887fb918 because I find it easier to convince myself that the right data is being populated at the right times. Of course, let me know if you think there's a better solution.

Closes #14206

Release Notes:
- ??? I don't know what to write here! Suggestions welcome

### DIFF
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -67,6 +67,7 @@
       "ctrl-o": "pane::GoBack",
       "ctrl-i": "pane::GoForward",
       "ctrl-]": "editor::GoToDefinition",
+      "ctrl-t": "pane::GoToOlderTag",
       "escape": "vim::SwitchToNormalMode",
       "ctrl-[": "vim::SwitchToNormalMode",
       "v": "vim::ToggleVisual",
--- a/crates/agent_ui/src/agent_diff.rs
+++ b/crates/agent_ui/src/agent_diff.rs
@@ -471,7 +471,7 @@ impl Item for AgentDiffPane {
 
     fn navigate(
         &mut self,
-        data: Box<dyn Any>,
+        data: Arc<dyn Any + Send>,
         window: &mut Window,
         cx: &mut Context<Self>,
     ) -> bool {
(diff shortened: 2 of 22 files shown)


cameron1024 (approved):
Thanks for working on this, code looks good 😁

cameron1024:
Ooh actually that test failure seems important, taking a look

PR #46004, opened 2026-01-04
git_ui: Implement `to_proto` for `GitBlob`

cppcoffee:
Release Notes:

- N/A


cole-miller:
Is this useful? I'm not sure where we would be using this method in the app for `GitBlob`.

PR #46011, opened 2026-01-04
Fix fold persistence corruption from external file changes

### LINKED ISSUE #33633: bug: Folded code blocks include extra lines after reopening Zed
### Summary
When I fold a block of code and close Zed, reopening the project later causes the folded block to incorrectly include extra lines from the next block (outside its intended scope).

### Steps to reproduce:
1. Fold a block of code
2. Close the Zed editor
3. Reopen the project after some time. 
4. The folded block now contains extra lines from the following code block that weren’t part of it originally.

**Expected Behavior**:
Folded code blocks should retain their correct boundaries and not include lines from unrelated code when reopening the project.

**Actual Behavior**:
The folded block includes additional lines from the next block, making the fold inaccurate.

**Image**:

Image

Zed: v0.192.7 (Zed) 
OS: Linux X11 linuxmint 22.1
Memory: 15.5 GiB
Architecture: x86_64

Brandtweary:
## Summary

Fixes #33633 - folds corrupting after external file modifications.

The existing fold persistence stored raw byte offsets, which become stale when files are modified externally (git operations, other editors, sync tools). This caused folds to capture wrong content on restore - users reported "wrong lines getting folded" with systematic offsets.

**Solution: Content fingerprinting**
- Store 32-byte content samples at fold boundaries
- On restore, validate fingerprints match; if not, search buffer for new positions
- Handles edge cases: short folds (< 32 bytes), duplicate content, boundary positioning

**Bonus fix: Ungraceful exit survival**
- Entity IDs change between sessions, but workspace cleanup's CASCADE DELETE was wiping folds before new editors could save
- Now migrates folds to new entity_id immediately after restore

## Test plan

- [x] Unit test for fingerprint storage/retrieval (`test_save_and_get_editor_folds_with_fingerprints`)
- [x] Manual test: fold sections, add content at file start externally, reopen → folds restore at correct positions
- [x] Manual test: fold sections, Ctrl+C, reopen → folds survive
- [x] Existing fold tests pass (`cargo test -p editor`)

## Release Notes

- N/A

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -3504,15 +3504,11 @@ impl Editor {
                 // end_fp: last min(32, fold_len) bytes of fold content
                 // Clip to character boundaries to handle multibyte UTF-8 characters.
                 let fold_len = end - start;
-                let start_fp_end = snapshot.clip_offset(
-                    start + std::cmp::min(FINGERPRINT_LEN, fold_len),
-                    Bias::Left,
-                );
+                let start_fp_end = snapshot
+                    .clip_offset(start + std::cmp::min(FINGERPRINT_LEN, fold_len), Bias::Left);
                 let start_fp: String = snapshot.text_for_range(start..start_fp_end).collect();
-                let end_fp_start = snapshot.clip_offset(
-                    end.saturating_sub(FINGERPRINT_LEN).max(start),
-                    Bias::Right,
-                );
+                let end_fp_start = snapshot
+                    .clip_offset(end.saturating_sub(FINGERPRINT_LEN).max(start), Bias::Right);
                 let end_fp: String = snapshot.text_for_range(end_fp_start..end).collect();
 
                 (start, end, start_fp, end_fp)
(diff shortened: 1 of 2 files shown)


crates/editor/src/editor.rs
```diff
@@ -23104,18 +23116,87 @@ impl Editor {
...
-                            snapshot.clip_offset(MultiBufferOffset(start), Bias::Left)
-                                ..snapshot.clip_offset(MultiBufferOffset(end), Bias::Right)
+                let snapshot_len = snapshot.len().0;
+
+                // Helper: search for fingerprint in buffer, return offset if found
+                let find_fingerprint = |fingerprint: &str, search_start: usize| -> Option<usize> {
+                    let fp_len = fingerprint.len();
+                    (search_start..snapshot_len.saturating_sub(fp_len)).find(|&offset| {
+                        snapshot.contains_str_at(MultiBufferOffset(offset), fingerprint)
+                    })
```
Veykril: This will break with multibyte characters

Brandtweary: Good catch! The issue was on both sides. The load-side byte-by-byte iteration was inefficient but wouldn't crash (thanks to the `clip_offset` check in `contains_str_at`). However, the save-side fingerprint extraction would panic when `start + FINGERPRINT_LEN` landed mid-character. Fixed by: (1) clipping fingerprint boundaries to character boundaries on save using `clip_offset()`, and (2) iterating by character boundaries using `chars_at()` on load while tracking byte offsets manually - matching the pattern used elsewhere in the codebase.

Veykril (approved):
Needs a cargo fmt

PR #46027, opened 2026-01-04
lsp: Skip serializing of params if unit type

### LINKED ISSUE #45994: LSP Shutdown Request Is Malformed
### Reproduction steps

Assuming you have some kind of LSP enabled for whatever language you're working with, the repro steps are as simple as:

1. Start Zed.
2. Open project that will use your LSP.
3. Close Zed (which should send a shutdown request to the LSP).
4. View your Zed log (assuming your LSP gives a response error to malformed shutdown request)

I've noticed this issue specifically with the D language extension which uses the `serve-d` LSP. So, the below instructions will reflect that.

Rough Repro Steps:
1. Download a D compiler and ensure it's on your PATH.
2. Start Zed.
3. Install/configure the D language extension. (I believe it has the ability to fetch the serve-d LSP on its own, though I have built serve-d from source and supplied the extension with a path to this)
4. Create D project somewhere using a terminal using `dub init .` where you want to create the project. (dub is shipped with D compilers, think cargo-ish)
5. Open this folder in Zed.
6. Ensure serve-d is running. (Look at your Zed log or use your favorite task management program, top, ps, whatever)
7. Close Zed.
8. Note the output in your Zed log: `2026-01-03T11:57:58-05:00 ERROR [lsp] Shutdown request failure, server serve-d (id 3): `params` MUST be an object (named arguments) or array (positional arguments), other types are not allowed by spec`

# Current Behavior

The shutdown request sends a request where `params` is of unit type:

Image

(link to code: https://github.com/zed-industries/zed/blob/94faaebfec5e46afa6d3aee5372cbbdb33b97c33/crates/lsp/src/lsp.rs#L956 pinned at specific commit in case of changes)

Which gets serialized here:

Image

(link to code: https://github.com/zed-industries/zed/blob/94faaebfec5e46afa6d3aee5372cbbdb33b97c33/crates/lsp/src/lsp.rs#L1264 pinned at specific commit in case of changes)

As a result, the JSON received by the LSP will have `params` come in as `null` like: `{"jsonrpc":"2.0","id":15,"method":"shutdown","params":null}`

Which as I understand is incorrect for the JSON-RPC spec. See https://www.jsonrpc.org/specification (section 4) where it says

```
params
A Structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted.
```

and

```
4.2 Parameter Structures
If present, parameters for the rpc call MUST be provided as a Structured value. Either by-position through an Array or by-name through an Object.

- by-position: params MUST be an Array, containing the values in the Server expected order.
- by-name: params MUST be an Object, with member names that match the Server expected parameter names. The absence of expected names MAY result in an error being generated. The names MUST match exactly, including case, to the method's expected parameters.
```

and if I pass in `{"jsonrpc":"2.0","id":15,"method":"shutdown","params":null}` to this validator https://www.json-rpc.dev/tools/validator it indicates that the message is invalid.

# Expected Behavior
Per the JSON-RPC spec, the expected behavior is for `params` to be omitted if we're not sending params.

This could be done by modifying the `Request` struct to make `params` an Option type annotated with `#[serde(default, skip_serializing_if = "Option::is_none")]` like so:

Image

(Current code: https://github.com/zed-industries/zed/blob/94faaebfec5e46afa6d3aee5372cbbdb33b97c33/crates/lsp/src/lsp.rs#L211 pinned at specific commit in case of changes)

This would be a bit of an infectious change that'd require changing some function parameter types to `Option<T::Params>` and wrapping parameters passed in with `Some(myParams)` and passing `None` instead of `()` in the case of no parameters. But, it seems fairly trivial to do. If this is acceptable, I can create a Pull Request for this. (Though Rust definitely ain't my first language).

Zed: Zed 0.217.3 80433cb239e868271457ac376673a5f75bc4adb1
OS: Fedora Linux 43
Memory: 128 GB (acquired from before the RAM shortage :D)
Architecture: x86_64

It's worth noting that I also built Zed from `main` which has the same problem.

### Attach Zed log file

```log
2026-01-03T11:57:50-05:00 INFO  [zed] ========== starting zed version 0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1, sha 80433cb ==========
2026-01-03T11:57:50-05:00 INFO  [crashes] spawning crash handler process
2026-01-03T11:57:50-05:00 INFO  [extension_host] extensions updated. loading 3, reloading 0, unloading 0
2026-01-03T11:57:50-05:00 INFO  [gpui::platform::linux::platform] activate is not implemented on Linux, ignoring the call
2026-01-03T11:57:50-05:00 INFO  [gpui::platform::blade::blade_renderer] Initializing Blade pipelines for surface SurfaceInfo { format: Bgra8Unorm, alpha: PreMultiplied }
2026-01-03T11:57:50-05:00 INFO  [workspace] Rendered first frame
2026-01-03T11:57:50-05:00 INFO  [zed::zed] Using GPU: GpuSpecs { is_software_emulated: false, device_name: "AMD Radeon RX 6600 XT (RADV NAVI23)", driver_name: "radv", driver_info: "Mesa 25.2.7" }
2026-01-03T11:57:50-05:00 INFO  [crashes] connected to crash handler process after 100ms
2026-01-03T11:57:50-05:00 INFO  [crashes] crash handler registered
2026-01-03T11:57:50-05:00 INFO  [util] set environment variables from shell:/bin/bash, path:/home/spirrwell/.cargo/bin:/home/spirrwell/.local/bin:/home/spirrwell/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin
2026-01-03T11:57:50-05:00 INFO  [node_runtime] Node runtime install_if_needed
2026-01-03T11:57:50-05:00 INFO  [language] found user-installed language server for clangd. path: "/usr/bin/clangd", arguments: []
2026-01-03T11:57:50-05:00 INFO  [lsp] starting language server process. binary path: "/usr/bin/clangd", working directory: "/home/spirrwell/Dev/dtest", args: []
2026-01-03T11:57:50-05:00 WARN  [language] unrecognized capture name 'test.inside' in D textobjects TreeSitter query
2026-01-03T11:57:50-05:00 WARN  [language] unrecognized capture name 'test.around' in D textobjects TreeSitter query
2026-01-03T11:57:50-05:00 WARN  [language] unrecognized capture name 'parameter.inside' in D textobjects TreeSitter query
2026-01-03T11:57:50-05:00 INFO  [lsp] starting language server process. binary path: "/home/spirrwell/.local/bin/serve-d", working directory: "/home/spirrwell/Dev/dtest", args: []
2026-01-03T11:57:50-05:00 INFO  [node_runtime] using Zed managed Node.js at /home/spirrwell/.local/share/zed/node/node-v24.11.0-linux-x64 since system Node.js wasn't found on PATH: cannot find binary path
2026-01-03T11:57:50-05:00 INFO  [lsp] Language server with id 3 sent unhandled notification coded/changedSelectedWorkspace:
{
  "uri": "file:///home/spirrwell/Dev/dtest",
  "name": "",
  "initialized": true,
  "selected": true,
  "pendingErrors": {}
}
2026-01-03T11:57:51-05:00 INFO  [lsp] Language server with id 3 sent unhandled notification coded/changedSelectedWorkspace:
{
  "uri": "file:///home/spirrwell/Dev/dtest",
  "name": "",
  "initialized": false,
  "selected": true,
  "pendingErrors": {}
}
2026-01-03T11:57:51-05:00 INFO  [lsp] Language server with id 3 sent unhandled notification coded/changedSelectedWorkspace:
{
  "uri": "file:///home/spirrwell/Dev/dtest",
  "name": "",
  "initialized": false,
  "selected": true,
  "pendingErrors": {}
}
2026-01-03T11:57:51-05:00 INFO  [lsp] starting language server process. binary path: "/home/spirrwell/.local/share/zed/node/node-v24.11.0-linux-x64/bin/node", working directory: "/home/spirrwell/Dev/dtest", args: ["/home/spirrwell/.local/share/zed/languages/json-language-server/node_modules/vscode-langservers-extracted/bin/vscode-json-language-server", "--stdio"]
2026-01-03T11:57:52-05:00 INFO  [lsp] Language server with id 3 sent unhandled notification coded/initDubTree:
null
2026-01-03T11:57:58-05:00 ERROR [crates/fs/src/fs_watcher.rs:216] Error { kind: WatchNotFound, paths: ["/home/spirrwell/.cache/zed/.tmpx8HEM3"] }
2026-01-03T11:57:58-05:00 ERROR [lsp] Shutdown request failure, server serve-d (id 3): `params` MUST be an object (named arguments) or array (positional arguments), other types are not allowed by spec: {"jsonrpc":"2.0","id":77,"method":"shutdown","params":null}
```

</details>

### Relevant Zed settings

```json
{
  "completion_menu_scrollbar": "always",
  "auto_signature_help": true,
  "show_signature_help_after_edits": true,
  "auto_update": false,
  "lsp": {
    "serve-d": {
      "settings": {
        "d": {
          "enableFormatting": true,
          "enableDubLinting": true,
          "enableLinting": true,
          "manyProjectsThreshold": 20,
          "argumentSnippets": false,
          "scanAllFolders": false,
          "lintOnFileOpen": "project",
          "dubPath": "/home/spirrwell/Compilers/ldc2-1.41.0-linux-x86_64/bin/dub",
          "dubCompiler": "/home/spirrwell/Compilers/ldc2-1.41.0-linux-x86_64/bin/ldc2",
          "dcdClientPath": "/home/spirrwell/.local/bin/dcd-client",
          "dcdServerPath": "/home/spirrwell/.local/bin/dcd-server",
        },
        "dscanner": {
          "ignoredKeys": ["dscanner.style.long_line"],
        },
      },
      "binary": {
        "path": "/home/spirrwell/.local/bin/serve-d",
      },
    },
  },
}

```

</details>

### Relevant Keymap

```json
// Zed keymap
//
// For information on binding keys, see the Zed
// documentation: https://zed.dev/docs/key-bindings
//
// To see the default key bindings run `zed: open default keymap`
// from the command palette.
[
  {
    "context": "Workspace",
    "bindings": {
      // "shift shift": "file_finder::Toggle"
    }
  },
  {
    "bindings": {
      // "j k": "vim::NormalBefore"
    }
  },
  {
    "context": "Pane",
    "bindings": {
      "alt-left": "pane::GoBack"
    }
  },
  {
    "context": "Pane",
    "bindings": {
      "alt-right": "pane::GoForward"
    }
  }
]

```

</details>

_No response_

None

Spirrwell:
Closes #45994 

Per the JSON-RPC specification, `params` "MUST be provided as a Structured value" or the member "MAY be omitted." See specification here: https://www.jsonrpc.org/specification#request_object

The code was passing in unit `()` for params which serde_json was serializing to `null` which is not valid.

This pulls request adds an `is_unit` function and annotates the `params` fields in the `Request` and `Notification` structs with the following:
 ```rs
#[serde(default, skip_serializing_if = "is_unit")]
```

so that the field will be omitted if the type is unit. This does also introduce a `'static` bound to these structs.

> [!WARNING]
> While this seems to me like a simple change, I am not super familiar with Rust, please look at this PR with extra scrutiny. The last thing I want to do is break everyone's LSP integration.

Release Notes:
- LSP integrations: send language server shutdown requests with correct parameters

### DIFF
--- a/crates/lsp/src/lsp.rs
+++ b/crates/lsp/src/lsp.rs
@@ -25,6 +25,7 @@ use smol::{
 };
 
 use std::{
+    any::TypeId,
     collections::BTreeSet,
     ffi::{OsStr, OsString},
     fmt,
@@ -200,14 +201,22 @@ pub enum RequestId {
     Str(String),
 }
 
+fn is_unit<T: 'static>(_: &T) -> bool {
+    TypeId::of::<T>() == TypeId::of::<()>()
+}
+
 /// Language server protocol RPC request message.
 ///
 /// [LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage)
 #[derive(Serialize, Deserialize)]
-pub struct Request<'a, T> {
+pub struct Request<'a, T>
+where
+    T: 'static,
+{
     jsonrpc: &'static str,
     id: RequestId,
     method: &'a str,
+    #[serde(default, skip_serializing_if = "is_unit")]
     params: T,
 }
 
@@ -245,10 +254,14 @@ enum LspResult<T> {
 ///
 /// [LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notificationMessage)
 #[derive(Serialize, Deserialize)]
-struct Notification<'a, T> {
+struct Notification<'a, T>
+where
+    T: 'static,
+{
     jsonrpc: &'static str,
     #[serde(borrow)]
     method: &'a str,
+    #[serde(default, skip_serializing_if = "is_unit")]
     params: T,
 }
 


Spirrwell:
@cla-bot check

Spirrwell:
> Thank you for the PR, I will come back to this next week most probably, one note after skimming: we do send both Shutdown and Exit messages with `fn XXX_internal` right?
> 
> https://github.com/zed-industries/zed/blob/5505e63a0df191aeed2988e5cd8073f2761f4303/crates/lsp/src/lsp.rs#L950
> 
> https://github.com/zed-industries/zed/blob/5505e63a0df191aeed2988e5cd8073f2761f4303/crates/lsp/src/lsp.rs#L988
> 
> I'd expect we change this particular layer to accept `Option` of parameters, and all public methods like `notify` and `request` can be unchanged?
> 
> If I search for `params: none` in the [current spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#version_3_17_0), there are only 2 client -> server requests (shutdown and exit), so we do not need any other case in the forseeable future, hence changing the public API seems wasteful.

So, there are two things here I'm not sure how to tackle if we wanted to not change the public API.

1. It seems like there is a fake LSP implementation that tests the public API. This can be seen here for example: https://github.com/zed-industries/zed/blob/88a0b310d619d22c0d7ab88564c94ae833514ec5/crates/collab/src/tests/editor_tests.rs#L2111 Unless we're explicitly testing receiving a `null` params, would it not make sense for the public API to provide a correct way to omit params? (I had opted to pass None in my current changes)
2. There is also this ClearCache request for `prettier`, which I'm not familiar with, but also exists here: https://github.com/zed-industries/zed/blob/88a0b310d619d22c0d7ab88564c94ae833514ec5/crates/prettier/src/prettier.rs#L532

Rust has the ability to have optionals of reference type, yes? To minimize additional overhead from going from `request` -> `request_internal` -> `request_internal_with_timer` with copies of params at each, it might be conceivable to turn these into optional references. I haven't tried that yet, but if the concern is that modifying the public API in the current way is wasteful, perhaps that is an alternative?

---
I have updated my fork to make sure the tests build (I hope I got them all, the project diagnostics in Zed made that pretty easy if so!)

However, I have noticed some run to run variance in tests passing. But, that seems to hold true even if I run them on the main branch of Zed, not in my fork.

I'll wait for some more thoughts before I remove the Draft status.

SomeoneToIgnore (approved):
Finally got the time to focus on reviews, sorry for dragging.

This is definitely a great way to solve things, better than the one I've been talking about.
`'static` in this context means that we do not borrow `T` (it has no ref, `&'a`, types) and seems totally fine: we do use various `lsp::InlayHintFooRequest`-like objects across the codebase.

Given it compiles and it's not a change in `gpui` public library code, it's totally fine and we can adjust it later any other way we need.

Than you!

PR #46029, opened 2026-01-04
Lazy-init WasmStore for wasm grammars

guidomb:
Release Notes:

- Fixed: lazily initialize tree-sitter WasmStore only when wasm grammars are parsed, restoring wasm grammar highlighting without eager Wasmtime init.

---

## Problem
`with_parser` always attached a WasmStore, which triggers Wasmtime/JIT initialization even for native grammars. On macOS this could crash or fail (SIGILL / "Failed to load the Wasm store") when wasm grammars are used.

## How I found it

While building a small gpui PoC on top of the Zed codebase, I ran into the crash/failure path when opening files backed by a wasm grammar.

DISCLOSURE: The PoC I was building and this PR was mainly driven by codex.

## Fix
- Parsers start without a WasmStore.
- WasmStore is created lazily when a wasm grammar is needed.
- `parse_text` attaches a store if the grammar is wasm (`ts_language.is_wasm()`), so wasm highlighting works without eager Wasmtime init.

## Repro
1. Build/run Zed from source on macOS.
2. Use a wasm grammar extension (e.g. Elixir).
3. Open an `.ex` or `.heex` file.
4. Before: SIGILL or "Failed to load the Wasm store".
5. After: syntax highlighting works, no eager Wasmtime init for native grammars.

### DIFF
--- a/crates/language/src/language.rs
+++ b/crates/language/src/language.rs
@@ -101,11 +101,7 @@ where
     F: FnOnce(&mut Parser) -> R,
 {
     let mut parser = PARSERS.lock().pop().unwrap_or_else(|| {
-        let mut parser = Parser::new();
-        parser
-            .set_wasm_store(WasmStore::new(&WASM_ENGINE).unwrap())
-            .unwrap();
-        parser
+        Parser::new()
     });
     parser.set_included_ranges(&[]).unwrap();
     let result = func(&mut parser);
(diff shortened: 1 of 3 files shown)


guidomb:
@cla-bot check

maxdeviant:
Thank you for the PR, but I don't think we're interested in this change.

We're expecting Wasm grammar support to be available.

PR #46039, opened 2026-01-04
gpui: Prepaint Div children with image cache

marcocondrache:
Closes #39914

I was able to reproduce the issue on macOS using the same README.

The root cause was that `markdown_preview_view` attaches the image cache to the root div, and while Div correctly applies that cache during request_layout and paint, the prepaint phase didn’t have the same context.

Markdown is rendered through a List, and `List::prepaint` renders its items using `layout_as_root`. That meant images were getting loaded during prepaint without the image cache in place. When that happened, Img fell back to the global asset loader, which retains assets indefinitely unless explicitly cleaned up.

So every image was loaded twice:

- once during layout/paint via RetainAllImageCache (correctly released when the preview closed)
- once during prepaint via the global asset system (never released)

The result was doubled memory usage and a leak, since the globally loaded images stuck around after the preview was closed.

Release Notes:

- Fixed a memory leak when previewing markdown files with images

### DIFF
--- a/crates/gpui/src/elements/div.rs
+++ b/crates/gpui/src/elements/div.rs
@@ -1425,6 +1425,11 @@ impl Element for Div {
         window: &mut Window,
         cx: &mut App,
     ) -> Option<Hitbox> {
+        let image_cache = self
+            .image_cache
+            .as_mut()
+            .map(|provider| provider.provide(window, cx));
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #14034, opened 2024-07-09
WIP: Disable binary downloads

mikayla-maki:
fixes https://github.com/zed-industries/zed/issues/12589 https://github.com/zed-industries/zed/issues/12354

follow up to https://github.com/zed-industries/zed/pull/12703

TODO:

- [ ] Implement a setting for globally enabling / disabling binary downloads
  - [ ] Node
  - [ ] Supermaven & Copilot
  - [ ] Built in and extension LSP binaries
- [ ] Report appropriate errors to the user when downloads are blocked this way
- [ ] Add this setting to the welcome page, so users have a chance to disable it before continuing.

Release Notes:

- TODO

PR #12703, opened 2024-06-05
[WIP] Require extensions to request downloads

someone13574:
Closes #12589 

This a WIP PR which will make extensions require user input before downloading binaries; with the aim of making Zed a bit more friendly for the security conscious. It still requires a fair bit of work to get working well though.

Current status:

- [x] Add async function called from the download function which creates a request notification.
- [x] Pass extension name to notification.
- [ ] Add setting to dictate the behavior of this feature (ie. always ask (default), always allow, never allow).
- [ ] Prevent repeat popup's due to lsp restart (not sure the best way to do with. Comments welcome)
- [ ] Remember selection per-lsp.

Also not sure about the best way to handle when declines occur without changing the api. Currently it just gives an Err which causes the lsp to attempt a restart.

Release Notes:

- Added notification for extensions requesting file downloads

Issue #12354, opened 2024-05-27
Ability to disallow network requests from plugins


jansol:
### Check for existing issues

- [X] Completed

### Describe the feature

There was a question on discord whether binaries downloaded by language plugins could be malicious. Organization IT admins are also frequently not very happy about such behavior so it would probably be good to have the option to disable it without preventing users from installing themes and language support or disabling network functionality entirely.

(yes this is far from trivial to enforce properly, but better to get the ball rolling on extension permissions sooner rather than later)

versecafe:
Can extensions even use the internet right now? Themes are an isolated JSON of colours, and the WASM binaries only have an output through the LSP handler, WASM shouldn't have access to the network at all, only WASIX and similar supersets get networking support. Is this more for when plugins widen to a more vscode style full featured system?

jansol:
There is an API for them to download a binary release from github, unpack it and exacute the binary. (So they can provide language servers that are installed and kept up to date automatically)

versecafe:
Isn't the WASM runtime isolated from the machine? it shouldn't have network or fs access just the buffer being passed in.

PR #46071, opened 2026-01-05
Add setting to disable binary downloads

### LINKED ISSUE #12589: Zed downloads NodeJS binary and npm packages from Internet without user’s consent

marcocondrache:
Closes #12589, #12354

Follow up of #14034 #12703

This is still a tentative and wip. I don’t have a lot of experience with this part of Zed yet, so it might take me a bit longer. I’ve tried to pull together all the ideas and feedback from #12589, #12703, #14034, and #12354.

The plan is to split the implementation into two parts. First, add a global setting, block all downloads, and fall back to binaries on the user’s PATH. If nothing is found, we just log the error.

In the second part, I want to dig into how we should handle that error case properly, including notification behavior and whether we need api changes.

Release Notes:

- TODO

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -138,6 +138,9 @@
   //  3. Do not restore previous workspaces
   //         "restore_on_startup": "none",
   "restore_on_startup": "last_session",
+  // Whether zed should automatically download binaries to run
+  // language servers if they are missing.
+  "allow_executable_downloads": true,
   // Whether to attempt to restore previous file's state when opening it again.
   // The state is stored per pane.
   // When disabled, defaults are applied instead of the state restoration.


marcocondrache:
@SomeoneToIgnore thank you. I totally missed that comment and wasn't aware of it.

PR #46074, opened 2026-01-05
extension: Auto-populate repository field in manifest if unset

MrSubidubi:
This allows us to auto-populate the extension remote based on the current git repository the extension is living in. This happens only if the repository is not set and provides an IMO sane default for the extension repository URL, given that this should point to the extension itself in >90% of all cases anyways.

Release Notes:

- N/A


maxdeviant:
@MrSubidubi Is there a solution here that would allow for making sure that we always end up with a repository in the `extension.toml`?

I don’t want to end up in a scenario where everyone starts leaving it out of the extension manifest because we do it automatically. I much prefer the explicitness of having it live in the TOML.

MrSubidubi:
Closing this for now with the conflicts and errors, will soon revisit this after https://github.com/zed-industries/zed/pull/57649 lands considering the feedback provided above.

Issue #7450, opened 2024-02-06
Support LSP Semantic Tokens


marwan-at-work:
### Check for existing issues

- [X] Completed

### Describe the feature

[Semantic Tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) allow for extra helpful highlighting that the language server might have better context over. 

For example, in Go many things are well highlighted via Semantic Tokens which are all explained [here](https://github.com/golang/tools/blob/master/gopls/doc/semantictokens.md). 

My personal favorite is being able to see formatting directives easily in strings: 

<img width="390" alt="Screenshot 2024-02-06 at 12 59 11 PM" src="https://github.com/zed-industries/zed/assets/16294261/9b0c282b-7e2c-47d9-9622-27620194b9f1">

which is not currently supported in zed: 

<img width="380" alt="Screenshot 2024-02-06 at 1 03 02 PM" src="https://github.com/zed-industries/zed/assets/16294261/e78985aa-8cda-4c82-b3ef-db61ce9e7e1c">

Thanks!

JosephTLyons:
These are different requests, but this one feels like its in a similar category:

- https://github.com/zed-industries/zed/issues/5345

filiptibell:
I've been trying to port my custom One Dark theme to Zed, everything went pretty smoothly, except for more complex code. Unfortunately some sections end up being almost unreadable compared to VSCode, where semantic tokens are available. VSCode on the top, and Zed on the bottom:

<img src="https://github.com/zed-industries/zed/assets/9497781/2e733a04-4401-4994-852e-1c85e9f88bd9" width="50%"/>
<img src="https://github.com/zed-industries/zed/assets/9497781/25ee482a-adc9-4991-b2a6-35c02d7f6ad2" width="50%"/>

note how customizations are missing for mutable, macros, and enum members.

Here's another example of mutable vs non mutable semantic token highlighting:

<img src="https://github.com/zed-industries/zed/assets/9497781/d3e9d76f-9900-43b2-a84a-5b3b4a71fe92" width="50%"/>
<img src="https://github.com/zed-industries/zed/assets/9497781/8a7bc029-d43d-4ef0-8999-fb9688098b62" width="50%"/>

more useful than one might initially think since it applies to variables / parameters ***and*** method calls.

[...]

PR #46115, opened 2026-01-05
languages: Highlight enumerator values as constants in C++

ozacod:
Before:
before

After:

after

Release Notes:

- Fix constant highlight issue for enumerator values


ozacod:
@cla-bot check

maxbrunsfeld:
My only issue with this is that we probably can't do it consistently between *definitions* of the enumerator values and *usages* of those values. If this adds greater inconsistency, I tend to be against it. Thoughts @ozacod ?

ozacod:
If enumerator value is not upper case, it would not be possible to highlight "Color:Red" without semantic understanding. This is probably not possible with tree sitter, requires lsp semantic tokens (Issue #7450). I agree, we should not implement it before related issue is solved. @maxbrunsfeld

PR #46116, opened 2026-01-05
languages: Capture escape sequences in C and C++

ozacod:
Before:
before

After:
after

Release Notes:

- Added highlighting for escape sequences for C and C++.

### DIFF
--- a/crates/languages/src/cpp/highlights.scm
+++ b/crates/languages/src/cpp/highlights.scm
@@ -198,7 +198,7 @@ type: (primitive_type) @type.builtin
   (raw_string_literal)
 ] @string
 
-(escape_sequence) @constant.character.escape
+(escape_sequence) @string.escape
 
 [
   ","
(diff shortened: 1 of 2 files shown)


crates/languages/src/cpp/highlights.scm
```diff
@@ -198,6 +198,8 @@ type: (primitive_type) @type.builtin
   (raw_string_literal)
 ] @string
 
+(escape_sequence) @constant.character.escape
```
MrSubidubi: Would `@string.escape` not better match the context here? 

E.g., https://github.com/zed-industries/zed/blob/06a623e99cc4360841dcd577d61a1bb9d733ca5b/crates/languages/src/rust/highlights.scm#L139 also uses that, with many other languages

ozacod: It looks way better and solved both problems.  Ty,


Before C:
before_c

After C:
after_c

Before cpp:
before_cpp

After cpp:
after_cpp

MrSubidubi (approved):
Looks good now, thank you very much!

PR #46118, opened 2026-01-05
languages: Add missing keywords and cast operators for c++

ozacod:
Release Notes:

- Added missing keywords and cast operators for C++


ozacod:
Closed pr Revert missing keywords due to tree-sitter error

PR #46125, opened 2026-01-06
languages: Add structs, unions and enums to outline in C

ozacod:
Before:
before

After:
after

Release Notes:

- Fixed struct union enum outline issues for c


crates/languages/src/c/highlights.scm
ozacod, line 1: If we aren't merging the PR #46115(we tend to), I should revert this change.

maxbrunsfeld (approved):
Thank you!

PR #46135, opened 2026-01-06
languages: Fix C/C++ outline for mixed storage classes and type qualifiers

ozacod:
Current tree sitter configuration for the outline was not working correctly since the rule required a strict order, firstly
storage_class_specifier then type_qualifier. This resulted in a failure to recognize types such as "constexpr int", "const double", etc. This PR proposes a solution with diagnostic capture of type and storage qualifiers which ensures full context capture and works well with C++.

Before:
before

After:
after

Release Notes:

- Fixed outline configuration for C++ variables with complex specifiers.

### DIFF
--- a/crates/languages/src/cpp/outline.scm
+++ b/crates/languages/src/cpp/outline.scm
@@ -38,8 +38,10 @@
     name: (_) @name) @item
 
 (declaration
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks!

PR #46140, opened 2026-01-06
Add agent thread sharing

mikayla-maki:
Staff only ship for now

Here's the agent planning doc that guided this: https://gist.github.com/mikayla-maki/c826b7997bd85b58273c1def9397940b

Release Notes:

- N/A


rtfeldman (approved):
Based on our realtime convo, approach seems good! 🤘

PR #46144, opened 2026-01-06
docs: Update "Custom Keybindings for Extension-Based Agents section" to include a troubleshooting note for defining an agent name

iagombermudez:
## Description
When trying to setup the OpenCode external agent the other day, I was having trouble setting the keybinding using the docs. The issue was that it said to use the same name as the one written in the UI. While this works when the external agent is defined using the `settings.json` file, it didn't work when setting up the keybinding after installing the extension through the extension marketplace.

Since the solution I found was to define the agent name as 'opencode' instead of 'OpenCode' (which is the name that shows up in the UI) I thought it would be appropiate to indicate that sometimes you might need to set it up using a different name.

Release Notes:

- N/A

### DIFF
--- a/docs/src/ai/external-agents.md
+++ b/docs/src/ai/external-agents.md
@@ -221,7 +221,7 @@ To assign a custom keybinding to start a new thread for agents that were added b
       {
         "agent": {
           "custom": {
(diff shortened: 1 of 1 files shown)


danilo-leal (approved):
Thank you!

PR #46165, opened 2026-01-06
ui_input: Only render number field border on focus

dinocosta:
The `NumberField.render` method was always rendering a `border_1` for the number input, even when not focused. Even though it was using `cx.theme().colors().border_transparent`, some themes don't have this value set to a fully transparent color so it ends up looking a little bit weird.

As such, this commit updates the way the border is rendered, to ensure it's actually only rendered, i.e., the `border_1` method is only called, when the field is actually focused.

Release Notes:

- Improved rendering of border on number field's input to ensure it's only rendered when the field is focused


danilo-leal (approved):
:shipit:

PR #46171, opened 2026-01-06
Rename visual_test_runner to zed_visual_test_runner

rtfeldman:
This fixes the auto-update failure by ensuring cargo-bundle picks 'zed' as the first binary.

**Problem:** `cargo metadata` returns binaries alphabetically, so `visual_test_runner` was coming before `zed`. When cargo-bundle looks for the first binary to bundle (without `--bin`), it was picking `visual_test_runner` - which doesn't exist in release builds because it requires the `visual-tests` feature.

The previous fix (#46163) added `--bin zed` to explicitly select the binary, but this caused cargo-bundle to use the binary name ('zed') instead of the bundle metadata name ('Zed Nightly'), breaking auto-updates.

**Fix:** Rename the binary to `zed_visual_test_runner` so it comes after `zed` alphabetically. This restores the original bundle script behavior that has worked for years.

Also reverts the `--bin zed` workaround from #46163.

Release Notes:

- N/A

### DIFF
--- a/docs/src/development/macos.md
+++ b/docs/src/development/macos.md
@@ -69,7 +69,7 @@ You must grant Screen Recording permission to your terminal:
 ### Running Visual Tests
 
 ```sh
-cargo run -p zed --bin visual_test_runner --features visual-tests
+cargo run -p zed --bin zed_visual_test_runner --features visual-tests
 ```
 
 ### Updating Baselines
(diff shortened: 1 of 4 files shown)


docs/src/development/macos.md
```diff
@@ -69,15 +69,15 @@ You must grant Screen Recording permission to your terminal:
...
+cargo run -p zed --bin zed_visual_test_runner --features visual-tests
 ```
 
 ### Updating Baselines
 
 When UI changes are intentional, update the baseline images:
 
 ```sh
-UPDATE_BASELINE=1 cargo run -p zed --bin visual_test_runner --features visual-tests
+UPDATE_BASELINE=1 cargo run -p zed --bin zed_visual_test_runner --features visual-tests
```
auwi-nordic, line 80: Under "Initial Setup" between these two edits there's also this line

`UPDATE_BASELINE=1 cargo run -p zed --bin visual_test_runner --features visual-tests`

Should that have been changed to `zed_visual_test_runner` as well?

PR #44344, opened 2025-12-07
[Reference POC] CSV/TSV preview (Actual development happens in separate PRs)

HalavicH:
Hey everyone! 👋

This **draft** PR introduces an initial CSV preview for Zed — similar to Markdown preview, but for tabular data.

## **Why I’m proposing this**

I often work with CSV/TSV files and found myself missing a native preview/tabular data editor in Zed.
This prototype started as a personal exploration, but it also moves toward a longer-term goal: **laying groundwork for database client support inside Zed** — the one feature still keeping other IDEs on my machine.

---

## **Status**

* This PR is currently a **POC / draft**.
* I communicate with @Anthony-Eid regarding PR direction and progress.
* Some features are incomplete or rough around the edges.
* I’m actively iterating in my spare time.
* After initial discussion I plan to split this PR in parts based on functionality, and deliver them separately for easier review 

---

## **Current features**

* **Live CSV preview** with automatic refresh
* **Draggable column resizing** (based on the keymap editor’s table component)
* **CSV parsing with quoted/multiline fields**
* **Variable height rows** (requires updates to `ui/data_table.rs`)
* **Variable column support** (I lift const generics limitation of `Table` component to allow arbitraray number of columns)
* **Displaying row identifiers** as src line numbers or rows in the table
* **Sorting by any column** (A-Z/Z-A)
* **Cell selection** via mouse/keyboard including multiple selections
* **Viewport following** viewport automatically scrolls up/down on selection
* **Copy selected as** CSV/TSV/Markdown. Non rectangular selections will be padded with empty cells
* **Inline editing** of cells (Press enter, edit in place, enter again to submit)  

---

## **Demos**

Sorting & line numbers + custom column widths:
image_2025-12-07_10-01-39

Select and copy demo:
image_2025-12-07_10-01-56

General side by side view showing source `.csv` text:
image_2025-12-07_09-59-20

---

## **Implementation notes**

This prototype reuses existing components (`markdown_preview` and the keymap editor `Table`).
The main challenge was const generics: runtime column counts can’t map to compile-time consts, so I removed it in favor of dynamic column definition using newtype which wraps `Vec` instead of const-generic arrays, which probably will impact performance (didn't measure yet). 

It’s functional, and covers my usecase but clearly not elegant — feedback appreciated.

---

## **Future roadmap**

* Column filtering / searching within column
* TSV or rather `<any-delimiter>SV` support
* Polishing UI/UX (horizontal scrolling, variable height rows, monospace font)
[...]

PR #46190, opened 2026-01-06
data_table: #1 Add variable row height mode to table UI element

HalavicH:
## Add variable row height mode to data table infrastructure

This PR introduces support for variable row heights in the data table component, laying the groundwork for more flexible tabular data rendering in Zed.

**Context:**  
This is the first in a series of infrastructure-focused PRs split out from the [original CSV preview draft PR](https://github.com/zed-industries/zed/pull/44344). The draft PR remains open as a reference and will be incrementally decomposed into smaller, reviewable pieces like this one.

**Details:**  
- Adds a variable row height mode to the data table, enabling future features that require rows of differing heights (such as CSV preview and, eventually, database table views).
- No user-facing changes; this is an internal refactor to support upcoming functionality.

**Thanks:**  
Big thanks to @Anthony-Eid for pairing sessions and guidance on how to best structure and land these changes incrementally.

---

Release Notes:

- N/A (internal infrastructure change, no user impact)

### DIFF
--- a/crates/ui/src/components/data_table.rs
+++ b/crates/ui/src/components/data_table.rs
@@ -535,8 +535,15 @@ impl<const COLS: usize> Table<COLS> {
         self
     }
 
-    /// Enables variable row height list rendering for tables with rows of different heights.
-    /// This is theoretically slower than uniform_list but supports multiline content properly, which allows for non-uniform tables suitable for displaying CSV data.
+    /// Enables rendering of tables with variable row heights, allowing each row to have its own height.
+    ///
+    /// This mode is useful for displaying content such as CSV data or multiline cells, where rows may not have uniform heights.
+    /// It is generally slower than [`Table::uniform_list`] due to the need to measure each row individually, but it provides correct layout for non-uniform or multiline content.
+    ///
+    /// # Parameters
+    /// - `row_count`: The total number of rows in the table.
+    /// - `list_state`: The [`ListState`] used for managing scroll position and virtualization. This must be initialized and managed by the caller, and should be kept in sync with the number of rows.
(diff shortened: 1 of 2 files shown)


crates/ui/src/components/data_table.rs
```diff
@@ -22,28 +22,39 @@ const RESIZE_COLUMN_WIDTH: f32 = 8.0;
...
 struct UniformListData<const COLS: usize> {
-    render_item_fn: Box<dyn Fn(Range<usize>, &mut Window, &mut App) -> Vec<[AnyElement; COLS]>>,
+    render_list_of_rows_fn:
+        Box<dyn Fn(Range<usize>, &mut Window, &mut App) -> Vec<[AnyElement; COLS]>>,
     element_id: ElementId,
     row_count: usize,
 }
 
+struct VariableRowHeightListData<const COLS: usize> {
+    /// Unlike UniformList, this closure renders only single row, allowing each one to have it's own height
```
maxdeviant:
```suggestion
    /// Unlike UniformList, this closure renders only single row, allowing each one to have its own height.
```

crates/ui/src/components/data_table.rs
```diff
@@ -648,6 +682,9 @@ pub fn render_table_row<const COLS: usize>(
         .map_or([None; COLS], |widths| widths.map(Some));
 
     let mut row = h_flex()
+        // NOTE: `h_flex()` sneakily applies `items_center()` which is not default behavior for div element.
```
maxdeviant, line 685: Would be better to use a `div().flex().flex_row()` directly, then, if we don't want the `items_center()`.

crates/ui/src/components/data_table.rs
HalavicH: thx 😅

crates/ui/src/components/data_table.rs
HalavicH: Makes sense. I discovered that in a different way initially, hence such awkward patch :D

PR #46214, opened 2026-01-07
Fix multibyte multicolumn selection

### LINKED ISSUE #45814: Multi Select Column
### Reproduction steps

1. Open any file with russian letters
2. Select column (middle button on mouse)
[https://bit.ly/3MONIf8](https://bit.ly/3MONIf8) - screenshort

1. Open any file with russian letters
2. Select column (middle button on mouse)
[https://bit.ly/3MONIf8](https://bit.ly/3MONIf8) - screenshort

OS: Windows 10.0.26200
Memory: 127.9 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3070 || NVIDIA Corporation || 581.80 r581_66

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Zachiah:
Closes #45814 

Release Notes:
- Fixed multicolumn selections with multibyte character

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -434,7 +434,7 @@ pub enum SelectPhase {
         position: DisplayPoint,
         reset: bool,
         mode: ColumnarMode,
-        goal_column: u32,
+        column_overshoot: u32,
     },
     Extend {
         position: DisplayPoint,
@@ -442,7 +442,7 @@ pub enum SelectPhase {
     },
     Update {
         position: DisplayPoint,
-        goal_column: u32,
+        column_overshoot: u32,
         scroll_delta: gpui::Point<f32>,
     },
     End,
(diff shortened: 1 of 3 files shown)


crates/editor/src/editor.rs
```diff
@@ -4105,44 +4128,62 @@ impl Editor {
...
 
         let start_row = cmp::min(tail.row(), head.row());
         let end_row = cmp::max(tail.row(), head.row());
-        let start_column = cmp::min(tail.column(), goal_column);
-        let end_column = cmp::max(tail.column(), goal_column);
-        let reversed = start_column < tail.column();
+
+        let text_layout_details = self.text_layout_details(window);
+        let head_goal_x = display_map.x_for_display_point(head, &text_layout_details)
+            + em_advance * head_column_overshoot as f32;
```
ConradIrwin, line 4158: @Zachiah I don't think we should be multiplying out the em_advance here, but using x_for_display_point and display_column_for_x. 

Also this change needs a test please.

Zachiah: I started with using just `x_for_display_point` but it only works for actual characters. The problem is we have to know how many columns past the last character they went. I'm using this manual method to calculate the x pos for the column_overshoot not for the character part. `x_for_display_point` would just clip it to the end of the line. Here's a video of the behavior that is broken if I don't add the manual calculation:

https://github.com/user-attachments/assets/f7f49333-36d9-4796-bead-5a5ee3bec1fc

Notice how you are able to select in the lower line by moving forward in the upper line. This behavior doesn't work without manually keeping track of the overshoot. It worked before this change because we were just byte indexing into each line with the proper spot but since we are now trying to byte index into the first line and ask "where are we?" then use that position in other lines, there is now the possibility that that first line isn't long enough.

If there is something I'm missing just let me know.

Also, I will add a test!

ConradIrwin: In the case we're selecting from the mouse like that, we should be using the mouse's x position.

The problem with multiplying the em-width is that it doesn't work for non-monospace fonts, (or more usually in Zed, for emoji) where the x advance for a given unicode code point is not uniform.

Maikel1963:
When is this fix going to be in ZED. Actual I still can observe this wrong behavior in lines with the German letters äüöÄÜÖß (with monospace font)

PR #46224, opened 2026-01-07
terminal: Fix last character of IME marked text not being deleted

### LINKED ISSUE #46084: Last character of IME marked text persists until extra backspace or new input
### Reproduction steps

This issue seems to be specific to macOS; my testing shows that Windows is unaffected.

---
1. Type pinyin in Zed terminal to generate marked text.
2. Repeatedly press Backspace to delete the text.
3. The last character of the marked text remains stuck in the terminal.
4. Clear it by pressing Backspace once more, or by typing new characters to overwrite it.

Current behavior:

https://github.com/user-attachments/assets/9e76f911-153e-4e6f-b228-229e1ab98f45

Expected behavior:

The last character of the marked text should be deleted normally.

Zed: v0.218.3+preview.104.4adf4c6c1dc3cad2cb35273ee707812c6874ac3b (Zed Preview) 
OS: macOS 26.2
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

amtoaer:
Closes #46084

I'm not very familiar with this part of the code, so I'm not sure if this is the right way to fix it.

During debugging, I discovered that `marked_range_utf16` consistently stores an empty range representing the cursor position (e.g., `0..0`, `1..1`) instead of an actual marked range. This bug is caused by the range being empty.

In my testing, the IME starts working correctly as long as I change it to any non-empty range (e.g., `0..1`). Although I suspect the correct value should probably be `0..marked_text.encode_utf16().count()`.

Before/After:

https://github.com/user-attachments/assets/9ade119e-2f4f-4334-b33b-6ce8a97fa47c

Release Notes:

- Fixed an issue where backspace fails to remove the last character of IME marked text in the terminal.

### DIFF
--- a/crates/terminal_view/src/terminal_view.rs
+++ b/crates/terminal_view/src/terminal_view.rs
@@ -331,7 +331,9 @@ impl TerminalView {
 
     /// Gets the current marked range (UTF-16).
     pub(crate) fn marked_text_range(&self) -> Option<Range<usize>> {
-        self.ime_state.as_ref().map(|_| 0..1)
+        self.ime_state
+            .as_ref()
+            .map(|state| 0..state.marked_text.encode_utf16().count())
     }
 
     /// Clears the marked (pre-edit) text state.
--- a/crates/terminal_view/src/terminal_view.rs
+++ b/crates/terminal_view/src/terminal_view.rs
@@ -62,7 +62,6 @@ use std::{
 
 struct ImeState {
     marked_text: String,
-    marked_range_utf16: Option<Range<usize>>,
 }
 
 const CURSOR_BLINK_INTERVAL: Duration = Duration::from_millis(500);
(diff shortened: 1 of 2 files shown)


crates/terminal_view/src/terminal_view.rs
```diff
@@ -322,24 +321,17 @@ impl TerminalView {
...
+        self.ime_state = Some(ImeState { marked_text: text });
         cx.notify();
     }
 
     /// Gets the current marked range (UTF-16).
     pub(crate) fn marked_text_range(&self) -> Option<Range<usize>> {
-        self.ime_state
-            .as_ref()
-            .and_then(|state| state.marked_range_utf16.clone())
+        self.ime_state.as_ref().map(|_| 0..1)
```
reflectronic: I agree that this should be:
```suggestion
        self.ime_state
            .as_ref()
            .map(|state| 0..state.marked_text.encode_utf16().count())
```

reflectronic:
I don't have a Mac to test with, but assuming my suggestion still fixes the issue, this looks good to me.

Ideally we would implement the rest of this interface properly (e.g. adding a 'virtual' cursor that lets you move around in the marked text) but this is a good improvement.

PR #46229, opened 2026-01-07
git: Add RPC support for checkpoint operations

cppcoffee:
Release Notes:

- Add RPC support for checkpoint operations

This enables checkpoint, restore_checkpoint, compare_checkpoints, and diff_checkpoints to work with remote repositories in collaboration mode.


cole-miller:
Thanks! However, since these methods are only used by agent threads, and that code always runs locally at present, I think I'd prefer to leave these as `unimplemented!()` at present. If we end up needing remote checkpoint support at some time in the future, we can implement it then.

PR #16848, opened 2024-08-25
editor: Fix DeleteToPreviousWordStart and DeleteToNextWordEnd interaction with newlines

kjzl:
Closes #5285, #14389

Changes:
- `DeleteToPreviousWordStart` now deletes '\n' separately from preceding words and whitespace.
- `DeleteToNextWordEnd` now deletes '\n' and any following whitespace separately from subsequent words.
- Added an `ignore_newlines` flag to both actions to optionally retain the old behavior.

These modifications align the behavior more closely with other popular editors like VSCode and Sublime:
- `DeleteToPreviousWordStart` now matches the default <Ctrl+Backspace> action in those editors.
- `DeleteToNextWordEnd` becomes more intuitive and closely resembles the default <Ctrl+Delete> behavior in those editors.

Release Notes:

- Improved `DeleteToPreviousWordStart` and `DeleteToNextWordEnd` interactions around newlines. You can opt-in into the previous behavior by adding {"ignore_newlines": true} to either action's binds in your keymap. ([#5285](https://github.com/zed-industries/zed/issues/5285), [#14389](https://github.com/zed-industries/zed/issues/14389))

PR #46235, opened 2026-01-07
editor: Fix DeleteToPreviousSubwordStart and DeleteToNextSubwordEnd interaction with newlines 

RubenFricke:
Closes #40110

  Changes:
  - `DeleteToPreviousSubwordStart` now deletes `\n` separately from preceding subwords and whitespace.
  - `DeleteToNextSubwordEnd` now deletes `\n` and any following whitespace separately from subsequent subwords.
  - Added an `ignore_newlines` flag to both actions to optionally retain the old behavior.

  These modifications align the subword commands with their word counterparts and with other popular editors like VSCode and Sublime.

Related to: https://github.com/zed-industries/zed/pull/16848

  Release Notes:

  - Improved `DeleteToPreviousSubwordStart` and `DeleteToNextSubwordEnd` interactions around newlines. You can opt-in into the previous behavior by adding `{"ignore_newlines": true}` to either action's binds in your keymap.

  ---

  This is my first contribution to Zed! If anything should be done differently, please let me know. Happy to learn :)

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -244,6 +244,32 @@ pub struct DeleteToPreviousWordStart {
     pub ignore_brackets: bool,
 }
 
+/// Deletes from the cursor to the end of the next subword.
+/// Stops before the end of the next subword, if whitespace sequences of length >= 2 are encountered.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+#[serde(deny_unknown_fields)]
+pub struct DeleteToNextSubwordEnd {
+    #[serde(default)]
+    pub ignore_newlines: bool,
+    // Whether to stop before the start of the previous word, if language-defined bracket is encountered.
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore (approved):
Looks like a great first contributions, thank you for fixing this.

RubenFricke:
Thanks for the question @Andrei-Pozolotin! I believe the current implementation should already handle long_symbol_names_like_this correctly. Underscores are treated as subword separators in the boundary logic:

```rust
  // is_subword_start: underscore→letter transitions
  left == '_' && right != '_'

  // is_subword_boundary_end: letter→underscore transitions
  left != '_' && right == '_'
```
So DeleteToPreviousSubwordStart on long_symbol_names_like_this should stop at: this → like → names → symbol → long.
Could you test and let me know if you're seeing different behavior?

Regarding configurable regex or per-language settings: as far as I see, this would introduce complexity and would likely need some additional direction. Happy to discuss if there's a specific use case the current behavior doesn't cover.

PR #46244, opened 2026-01-07
Vim search */# without moving cursor initially

a-hariti:
The default behavior for Vim search with `*` and `#` in normal mode is to initiate a search and immediately jump to the next or previous match respectively.

This behavior can be annoying, so Vim has many plugins to address this specifically:

- [vim-asterisk](https://github.com/haya14busa/vim-asterisk)
- [vim-SearchHighlighting](https://github.com/inkarkat/vim-SearchHighlighting)
- [vim-tranquille](https://github.com/RRethy/vim-tranquille)

This PR tries to emulate this behavior natively keeping up with Zed's sane defaults and deviating from vanilla Vim when it makes sense.

Release Notes:

- Vim: `*` and `#` search doesn't jump immediately to next / previous search.


a-hariti:
I reverted the `vim.json` bindings on the same branch.

Tested it locally and it works both in normal and visual mode.
```json
[
  {
    "context": "Editor && vim_mode == normal",
    "bindings": {
      "*": "vim::SearchUnderCursor",
      "#": "vim::SearchUnderCursorPrevious"
     }
  }
]
```
Can you please reopen the PR and fetch changes?
or should I open another one?

Issue #5592, opened 2022-07-14
Show whitespace/invisibles


Daeraxa:
It would be nice if Zed could display whitespace characters/invisibles. 
I find this really useful when editing documentation and reviewing certain types of log files so is often left on.

Atom has settings for showing whitespace characters/invisibles (Editor > Invisibles) where visible types of whitespace can be identified. This is also present in other editors (Notepad++, VSCode).

I like the amount of flexibility provided by Atom to change the character displayed for each whitespace "type" but personally I've never changed it from the default.
If implemented it would be nice to toggle from the command palette rather than setting in the options, again like in Atom `Window: Toggle invisibles`

![image](https://user-images.githubusercontent.com/58074586/179065112-fb3460e6-8dbf-4597-8c75-f5908bb74b96.png)
![image](https://user-images.githubusercontent.com/58074586/179065287-44edbfad-17af-4322-ae88-21832c021c81.png)

pshergie:
the visible whitespaces across all included light themes are way too prominent, could you make it paler or give us an option to edit the color?

joseftw:
It's possible to change the color like this:
```
"experimental.theme_overrides": {
    "editor.invisible": "#777777"
  },
```

pshergie:
> It's possible to change the color like this:
> 
> ```
> "experimental.theme_overrides": {
>     "editor.invisible": "#777777"
>   },
> ```

it works! didn't know about the setting

bajtos:
> It's possible to change the color like this:
> 
> ```
> "experimental.theme_overrides": {
>     "editor.invisible": "#777777"
>   },
> ```

How does this setting interact with dark/light mode and custom theme colours? Instead of configuring the exact colour, have you considered allowing the user to configure the opacity (alpha channel) that will be applied to the theme-specific colour?

pshergie:
> It's possible to change the color like this:
> 
> ```
> "experimental.theme_overrides": {
>     "editor.invisible": "#777777"
>   },
> ```

the problem with the setting is that it works for light themes but looks bad for dark ones

VladimirCores:
How could I change color of the whitespaces dots to less pop?

olejorgenb:
> If implemented it would be nice to **toggle from the command palette** rather than setting in the options

Was this never implemented?

[...]

PR #46247, opened 2026-01-07
Tone down color for invisible characters in One themes

MJohnson459:
Related: https://github.com/zed-industries/zed/issues/5592

Coming from vscode, the first thing I noticed (after the amazing speed that zed started) was the boldness of the whitespace characters.

This mutes the whitespace "invisible" character color in the one theme so that they are less aggressive. This more closely (but not perfectly) aligns with the one theme for vscode. In order to avoid adding a new color to the theme, I reused the `editor.line_number` value in both cases which worked well.

# Dark
## original:
before_dark

## editor.line_number `#4e5a5f`
after_dark

## border.variant (alternative) `#363c46ff`
dark_border_variant

# Light
## original
before_light

## editor.line_number `#b4b4bb`
after_light

# Release Notes:

- Improved: Muted the invisible character color in the one themes.


danilo-leal (approved):
Nice, good catch! Thank you!

PR #44835, opened 2025-12-15
terminal: Remove SHLVL from terminal environment to fix incorrect shell level

edlsh:
Fixes #33958

## Problem

When opening a terminal in Zed, `SHLVL` incorrectly starts at 2 instead of 1. On `workspace: reload`, it increases by 2 instead of 1.

## Root Cause

1. Zed's `shell_env::capture()` spawns a login shell (`-l -i -c`) to capture the user's environment, which increments `SHLVL`
2. The captured `SHLVL` is passed through to the PTY options
3. When alacritty_terminal spawns the user's shell, it increments `SHLVL` again

Result: `SHLVL` = captured value + 1 = 2 (when launched from Finder)

## Solution

Remove `SHLVL` from the environment in `TerminalBuilder::new()` before passing it to alacritty_terminal. This allows the spawned shell to initialize `SHLVL` to 1 on its own, matching the behavior of standalone terminal emulators like iTerm2, Kitty, and Alacritty.

## Testing

- Launch Zed from Finder → open terminal → `echo $SHLVL` → should output `1`
- Launch Zed from shell → open terminal → `echo $SHLVL` → should output `1`
- `workspace: reload` → open terminal → `echo $SHLVL` → should remain `1`
- Tested with bash, zsh, fish

Release Notes:

- Fixed terminal `$SHLVL` starting at 2 instead of 1 ([#33958](https://github.com/zed-industries/zed/issues/33958))

PR #46273, opened 2026-01-07
terminal: Skip SHLVL when loading login shell environment

### LINKED ISSUE #33958: $SHLVL starts at 2 and increases by 1 every time Zed is reloaded
### Summary

$SHLVL increases by 1 every time Zed is reloaded. Additionally, if I start Zed with `zed .`, then $SHLVL starts at 2.

### Description

Steps to reproduce:
1. Launch Zed
2. Open the terminal and `echo $SHLVL`
3. Run the "workspace: reload" command
4. Open the terminal and `echo $SHLVL` again

**Expected Behavior**: $SHLVL should start at 1 when Zed opens or is reloaded
**Actual Behavior**: $SHLVL starts at 2 and increases by 2 every time Zed is reloaded

Zed: v1.3.6+stable.288.4ce18741ad2ea007cb20a1dd96d2a170d224cc01 (Zed) 
OS: macOS 15.7.7
Memory: 16 GiB
Architecture: x86_64

edlsh:
Fixes #33958

## Problem

PR #44835 attempted to fix SHLVL starting at 2 by removing it from the `env` HashMap passed to alacritty. However, that HashMap is only an **overlay** — alacritty uses the parent process environment as a base and applies the overlay on top. Since alacritty never calls `env_remove("SHLVL")`, the terminal shell still inherits `SHLVL` from Zed's process environment.

## Root Cause

The real issue is in `load_login_shell_environment()`:

1. `shell_env::capture()` spawns a login shell to capture environment variables
2. That login shell increments `SHLVL` (from 0→1 or n→n+1)
3. The captured env (including the incremented `SHLVL`) is written to Zed's **process environment** via `env::set_var`
4. When you open a terminal, the shell inherits Zed's `SHLVL` and increments it again → starts at 2
5. On reload, `shell_env::capture()` runs again with the already-elevated `SHLVL`, incrementing it further → +2 per reload

## Fix

Skip `SHLVL` when setting Zed's process environment in `load_login_shell_environment()`. This prevents the login-shell capture from polluting Zed's env, so terminals start fresh with the correct `SHLVL`.

### DIFF
--- a/crates/util/src/util.rs
+++ b/crates/util/src/util.rs
@@ -365,6 +365,13 @@ pub async fn load_login_shell_environment() -> Result<()> {
         .await
         .with_context(|| format!("capturing environment with {:?}", get_system_shell()))?
     {
+        // Skip SHLVL to prevent it from polluting Zed's process environment.
+        // The login shell used for env capture increments SHLVL, and if we propagate it,
+        // terminals spawned by Zed will inherit it and increment again, causing SHLVL
+        // to start at 2 instead of 1 (and increase by 2 on each reload).
+        if name == "SHLVL" {
+            continue;
+        }
         unsafe { env::set_var(&name, &value) };
     }
 


Veykril (approved):
Thanks!

PR #46277, opened 2026-01-07
Improve snippet parse error context

ghost:
Release Notes:

- N/A

Description:
E.g. Instead of saying:
2026-01-07T09:08:42-06:00 ERROR [crates/snippet_provider/src/lib.rs:46] failed to parse snippet
  Caused by:
      expected an integer

Now the logs give more valuable information like:
2026-01-07T10:47:39-06:00 ERROR [crates/snippet_provider/src/lib.rs:48] invalid snippet in /Users/oscarvarto/Library/Application Support/Zed/extensions/installed/django-snippets/./snippets/python.json (re_path)

### DIFF
--- a/crates/snippet_provider/src/lib.rs
+++ b/crates/snippet_provider/src/lib.rs
@@ -8,14 +8,13 @@ use std::{
     time::Duration,
 };
 
-use anyhow::{Context as _, Result};
+use anyhow::Result;
 use collections::{BTreeMap, BTreeSet, HashMap};
 use format::VsSnippetsFile;
 use fs::Fs;
 use futures::stream::StreamExt;
 use gpui::{App, AppContext as _, AsyncApp, Context, Entity, Task, WeakEntity};
 pub use registry::*;
-use util::ResultExt;
 
 pub fn init(cx: &mut App) {
     SnippetRegistry::init_global(cx);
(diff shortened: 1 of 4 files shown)


crates/snippet_provider/src/lib.rs
```diff
@@ -43,7 +43,11 @@ fn file_to_snippets(file_contents: VsSnippetsFile) -> Vec<Arc<Snippet>> {
             .description
             .map(|description| description.to_string());
         let body = snippet.body.to_string();
-        if snippet::Snippet::parse(&body).log_err().is_none() {
+        if snippet::Snippet::parse(&body)
+            .with_context(|| format!("invalid snippet in {} ({})", source.display(), name))
```
SomeoneToIgnore:
```suggestion
            .with_context(|| format!("invalid snippet '{name}' in {source:?}"))
```

Can we 

* reword the error a bit to make it more obvious where the snippet name is (not sure if single quotes around are good, just proposing)?

* do a 
```rs
if let Err(e) = snippet::Snippet::parse(&body) {
    log::error!("Invalid snippet '{name}' in {source:?}: {e:#}
}
```
-like change instead of `log_err()`?

The latter shows quite a backtrace in the settings, which seems totally unneeded.

SomeoneToIgnore (commented):
Thank you.

Since we start to improve things around, let's do a bit more?

crates/snippet_provider/src/lib.rs
ghost: Hi @SomeoneToIgnore, I pushed a commit trying to follow your suggestion. Could you check again when you get some time, please?

SomeoneToIgnore (approved):
Thank you!

SomeoneToIgnore:
Oh well, the crate needs to be added first?

Issue #35173, opened 2025-07-28
File explorer fails to refresh when files are added to symlinked directories


light-merlin-dark:
### Summary

File explorer fails to refresh when files are added to symlinked directories

### Description

When working with symlinked directories in a project, Zed's file explorer does not automatically refresh to show newly created files within those symlinked folders. Files added to symlinked directories through external means (terminal, other editors, or programmatically) remain invisible in the file explorer until a manual refresh or restart of Zed.

Steps to reproduce:
1. Create a symlink to a directory within your project (e.g., `ln -s /path/to/source ./linked-folder`)
2. Open the project containing the symlink in Zed
3. Create a new file in the symlinked directory using terminal or another method (e.g., `touch ./linked-folder/new-file.txt`)
4. Check Zed's file explorer - the new file does not appear

**Expected Behavior**: File explorer should detect and display new files added to symlinked directories, similar to how it handles regular directories.
**Actual Behavior**: New files in symlinked directories are not shown in the file explorer until Zed is restarted or the project is reopened.

### Zed Version and System Specs

Zed: v0.196.5 (Zed Preview) 
OS: macOS 15.5.0
Memory: 48 GiB
Architecture: aarch64

auwi-nordic:
Issue was still seen in 0.230.2

mohit-fotello:
Current stable version also has this issue

jaymoretti:
Can confirm: Zed 1.14.2 
Commit: 02abf5b08fa12c1c20a155ae3f796ef4c6c1a01e
Version: 1.14.2+stable.337.02abf5b08fa12c1c20a155ae3f796ef4c6c1a01e
Has this issue on windows 11

Issue #20849, opened 2024-11-19
`workspace:reload` to reload windows instead of restarting Zed entirely


srihariash999:
### Check for existing issues

- [X] Completed

### Describe the feature

- Request to move `workspace: reload` to `zed: restart` and add make `workspace: reload` do the action implied by it's name (close/reopen a single workspace/project) 

Currently `workspace:reload` completely restarts zed.

Some related discussion and a video demonstrating this in #20771

### Environment

Zed: v0.161.2 (Zed)
OS: macOS 15.1.0
Memory: 16 GiB
Architecture: aarch64

roufpup:
This is still a really annoying issue to this day :/

roskee:
New to Zed here — I ran `workspace: reload` expecting it to reopen my project, and was completely baffled when all my open projects closed and Zed vanished for a second. Pretty misleading name for that command.

Issue #5393, opened 2022-06-13
Top-Ranking Issues 📊


JosephTLyons:
*Updated on 01/24/2024 07:05 AM (EST)*

## enhancement

1. https://github.com/zed-industries/community/issues/130 (278 :thumbsup:)
2. https://github.com/zed-industries/zed/issues/5269 (247 :thumbsup:)
3. https://github.com/zed-industries/zed/issues/5259 (158 :thumbsup:)
4. https://github.com/zed-industries/zed/issues/5356 (132 :thumbsup:)
5. https://github.com/zed-industries/zed/issues/5343 (126 :thumbsup:)
6. https://github.com/zed-industries/zed/issues/5360 (107 :thumbsup:)
7. https://github.com/zed-industries/zed/issues/5308 (107 :thumbsup:)
8. https://github.com/zed-industries/zed/issues/5286 (102 :thumbsup:)
9. https://github.com/zed-industries/zed/issues/4992 (100 :thumbsup:)
10. https://github.com/zed-industries/zed/issues/5361 (95 :thumbsup:)
11. https://github.com/zed-industries/zed/issues/5065 (93 :thumbsup:)
12. https://github.com/zed-industries/zed/issues/5098 (90 :thumbsup:)
13. https://github.com/zed-industries/zed/issues/5373 (82 :thumbsup:)
14. https://github.com/zed-industries/zed/issues/5300 (82 :thumbsup:)
15. https://github.com/zed-industries/zed/issues/5242 (75 :thumbsup:)
16. https://github.com/zed-industries/zed/issues/5311 (74 :thumbsup:)
17. https://github.com/zed-industries/zed/issues/4523 (69 :thumbsup:)
18. https://github.com/zed-industries/zed/issues/5278 (59 :thumbsup:)
19. https://github.com/zed-industries/zed/issues/5301 (56 :thumbsup:)
20. https://github.com/zed-industries/zed/issues/5273 (54 :thumbsup:)

## design

1. https://github.com/zed-industries/community/issues/130 (278 :thumbsup:)
2. https://github.com/zed-industries/zed/issues/5259 (158 :thumbsup:)
3. https://github.com/zed-industries/zed/issues/5308 (107 :thumbsup:)
4. https://github.com/zed-industries/zed/issues/5286 (102 :thumbsup:)
5. https://github.com/zed-industries/zed/issues/5373 (82 :thumbsup:)
[...]

PR #46291, opened 2026-01-07
Add command to manually trigger a refresh/rescan of folders

### LINKED ISSUE #29598: Manual refresh on unsupported filesystems (nfs, fuse, exfat) without inofity/fsevents
### Summary

Zed currently relies upon inotify (linux) and fsevents (macos) to be notified of changes to files.  If you are working on a project on a filesystem which does not support this, Zed will not pickup changes (e.g. git changing branches, etc).
 
### Description

It would be nice if Zed had at least minimal support to operate in these environments.
1. Detect when inotify/fsevents unsupported and notify the user
2. Add manual "refresh" option so that Zed can manually refresh the file tree.
3. Trigger manual refresh when branch switcher is used in Zed.

This is also an issue with remote development.

Future: Support automatic periodic background refresh.

auwi-nordic:
Closes #29598

This may also help a bit with #20849 in the cases where users can do a folder refresh instead of a reload to solve their workspace issues.

This adds both a project panel context menu "Refresh Folders"  which refreshes the hierarchy below the chosen folder, and a command "workspace: refresh folders" which refreshes everything in the current window.

refreshfolders

Should work for remote ssh sessions as well. Tested locally and on remote workspaces on MacOS. 
Checked that it functions as workaround for #35173

I have very little experience with Rust and the Zed codebase, so someone more experienced should review the changes thorughly. Feature was implemented with the help of Claude agent, but did some iterations and manual changes to tidy up things.

Feature name is inspired by the Sublime "Project: Refresh Folders" command, which does something similar.

Release Notes:

- Add command to manually trigger a refresh/rescan of folders


auwi-nordic:
@cla-bot check

lilith:
I would say move it to in or near the 'reveal/open' area.

auwi-nordic:
@lilith Yeah that makes sense to me as well, I'm just afraid to suggest that myself as the space it takes up has a higher premium the higher on the list it is 😅

osiewicz (changes requested):
I feel somewhat strongly against workspace refreshes, because they tend to mask issues with our FS scanning code. Do you have any scenarios outside of that that would benefit from such a functionality?

chen1plus:
@osiewicz @probably-neb, would you consider accepting this feature if it were restricted to the **command palette** only, removing it from the context menu?

The reasoning is that manual execution via the command palette involves enough friction that users will still naturally prefer and expect automatic synchronization. Therefore, any underlying bugs in the FS scanning code will still be noticeable and reported, as users won't want to rely on manual refreshes as a permanent workflow.

**We would appreciate a clear update on the status of this PR.** The community is eager to see a resolution as this issue significantly impacts the daily workflow for many users.

auwi-nordic:
I'd be happy to remove the context menu item if it gets the PR merged. It wouldn't be unreasonable.

Just keep in mind that the lack of ability to fix this with extensions in Zed could mean it's good to have such things built in.

Does Zed want to retain users who are working on systems with NFS and get their feedback when using Zed? I think working on these kinds of systems is a prime use case for Zed, and a potential avenue to attract pioneering users, so it'd be a shame. The alternative may that they just give up and use other editors. We can put up with a lot of lack of functionality and bugs/glitches, but we gotta have the bare essentials (actually being able to see and open files). And these users are exactly the ones you need feedback from to discover issues with filesystem refresh behavior.

Both VS Code and Sublime Text have commands to trigger a refresh of the whole file explorer.  VS Code also has a refresh button in the explorer sidebar. But no context menu item without extension.

Eclipse has a "Refresh" context menu item in the project explorer.

Here's a discussion from Sublime Text

https://forum.sublimetext.com/t/file-tree-reload/22649/24

https://github.com/sublimehq/sublime_text/issues/1409

I think the discussions illustrate that a full refresh isn't really a satisfactory solution for large projects, since it can take a lot of time for the folder you're interested in to refresh.

I'd like to highlight this comment: 

> It’s unbelievable that this issue has not yet been solved in 2025.

Maybe best to get it sorted earlier rather than later.

If there's a way for the user to configure the context menu, like you can in e.g. Sublime, it'd be very reasonable to default to not showing the "Refresh Folder" context menu. For most users it's waste of space. But until Zed has that configurability, perhaps there's an argument to keep it there, since it'd help ensure that the core basics of the text editor actually works for everyone?

Perhaps include the context menu item first, and create an issue to remove it when it can be added back through configuration or extension? If you ask me, you don't have to worry that users won't report issues where the project panel explorer doesn't refresh.

ChristopherBiscardi:
> We would appreciate a clear update on the status of this PR.

First off, @auwi-nordic , thank you for the PR. We really appreciate your effort.

Given that a staff member has given a blocking review at the feature level, instead of the PR implementation level, and that there are questions around what implementation would be mergable: The best way for the community to advocate for this direction is to consolidate around a GitHub Discussion or Issue and accumulate enough upvotes to get it prioritized. This will help prioritize escalation in addressing it by allowing the team to note its position in lists like [the top ranking issues](https://github.com/zed-industries/zed/issues/5393) and give a place to discuss what specifically should be implemented [as indicated in the contributing guide](https://github.com/zed-industries/zed/blob/a5e78b02de143b64f9bf3a1bafd7c486f29cc5d2/CONTRIBUTING.md#L43-L45).

In particular, when doing this, I would suggest focusing on the use cases that will benefit, and address the concern around potentially masking issues with the fs scanning code.

chen1plus:
I've written a discussion (#54150) to describe this issue and the problem of communication.

okaris:
thanks for the PR and the discussion here, this is clearly a real pain point for a lot of people.

pushing this to “get more upvotes in a discussion” after a concrete PR and multiple real-world use cases feels off.

this isn’t a vague feature request. there’s a working implementation, clear demand, and the problem space is well understood. asking for more aggregation instead of giving a yes/no on direction just slows things down.

also, the concern about “masking fs issues” shouldn’t block shipping a fallback. these aren’t edge bugs that will get fully eliminated, they’re structural limitations depending on the filesystem and environment.

if the answer is no, it would be better to state that clearly with reasoning. if it’s maybe, then it would help to define what version of this would be acceptable (scope, placement, constraints).

right now it just reads as deferring a decision rather than engaging with the proposal.

lilith:
This issue has been in the top 5 issues so many times and gets closed every
time 10% of it is fixed. There's no fix. Zed doesn't even work on the most
common dev env on Windows - WSL.

ChristopherBiscardi:
> I've written a discussion (https://github.com/zed-industries/zed/discussions/54150) to describe this issue and the problem of communication.

Thank you! Having the communication disconnect clearly articulated there is useful for bringing it up internally as well. There was a clear issue in responding to this PR over the last few months.

> pushing this to “get more upvotes in a discussion” after a concrete PR and multiple real-world use cases feels off.

I very much understand. The main reasons for this course are:

That the primary visibility mechanisms used internally are lists of issues and descriptions. A PR with a blocking staff review like this one has is unlikely to be merged as-is, and has a tendency to be lost amongst the other PRs, so raising the visibility in this way brings it to the attention of more staff members in a way that can be advocated for.

The process detailed in [the contributing guide](https://github.com/zed-industries/zed/blob/a5e78b02de143b64f9bf3a1bafd7c486f29cc5d2/CONTRIBUTING.md#L43-L45) is that a PR like this should acquire staff buy-in *before* being submitted, and this one did not. It *should* have been redirected back in January when the blocking review was submitted as a result, but was not. This is unfortunate, but continuing to leave it open as it was is unlikely to garner the attention this needs.

> asking for more aggregation instead of giving a yes/no on direction just slows things down...
> 
> ...if the answer is no, it would be better to state that clearly with reasoning. if it’s maybe, then it would help to define what version of this would be acceptable (scope, placement, constraints).

A "no" on a particular PR is a temporary "no". *This* PR didn't make it in, and thus did get the "no on direction", but that doesn't mean another approach wouldn't make it in. That is why the line in the contributing guide exists for discussing approaches, and why I believe the issue is the right place to discuss this. There *are* additional approaches that seem viable here, like polling, and staff currently needs to be convinced that a manual trigger is also required.

I understand being frustrated, especially since this PR was left open with no response for so long, and I do believe that pushing the discussion into a more staff-visible location is the best path towards resolving the issue.

okaris:
Thanks for the thoughtful reply and addressing my points. I agree with you on overall opensource governance sense but in this particular case unfortunately the outright rejection and the justification, given the nature of the underlying issue, is disturbing and many others will agree. Sometimes you just need to fix things and leave the beurocracy to politicians in an age where tech moves at the speed of light.

Ömer Karışman

Sent via Superhuman ( ***@***.*** )

rochacbruno:
That is very disappointing to see this PR being closed, I opened a simpler version of this that was also closed because this one was to be prioritized.

https://github.com/zed-industries/zed/pull/48871 

Is there a reason for Zed to not fix this issue?

lilith:
I've switched to Yazi as an alternative to Zed for a better pairing with Claude code; it is much faster and doesn't have this problem. I don't think Zed will support working with Claude anytime soon and it might be a fool's errand to change the dev team perspective on this and get real windows or cross-filesystem support.

okaris:
It’s been two weeks since this discussion opened. I’m working on a remote Debian server, and Zed still frequently misses new files in the file tree.. especially when those files are created by agents or processes running outside the editor.

A manual refresh action is the simplest fallback here.

Instead, we seem to be waiting months for a more “elegant” watcher solution that still won’t fully solve cross-machine, remote FS, agent-driven, or cluster edge cases.

That’s the frustrating part: this is a pragmatic UX issue, not an architectural thought experiment. Users already provided a working PR, multiple reproducible real-world environments, and clear demand.

Polling helps some cases, but it doesn’t eliminate the need for a manual refresh when watcher notifications fail.

I appreciate the work the team is doing, but repeatedly rejecting a simple fallback while users are blocked in real workflows feels misaligned with the problem. Sometimes the correct solution is just giving users an escape hatch. Every major editor does this for a reason.

anuwick:
What is happening here. This is a deal breaker with AI changes not being detected by Zed file listener for whatever the reason.

twinkle311:
最新版本依然存在此问题,  zed外的创建的新文件, 没办法通过简单的refresh 操作显示出来

mpcref:
https://lap.dev/lapce/

PR #46294, opened 2026-01-07
workspace: Unpreview active tab when closing other tabs

mattermoran:
Un-preview the active tab when closing other tabs. 

In my mind the intent when clicking `Close Others` is to keep the current tab around, so it should no longer be treated as a preview.
This also matches the VSCode behavior.

Release Notes:

- Make preview tab permanent after closing all other tabs


SomeoneToIgnore (changes requested):
Any chance we can get a test for this?

mattermoran:
@SomeoneToIgnore done!

SomeoneToIgnore (approved):
Superb, thank you, missed this small tweak myself too.

PR #46298, opened 2026-01-07
Fix multiline regex

### LINKED ISSUE #45776: Replace empty lines with regex
### Reproduction steps

Replacing empty lines with this regular expression does not work: `^$\n`

sample file:
```
line 1


line 2
line 3


line 4
```

```
line 1
line 2
line 3
line 4
```

Zed 0.217.3 
80433cb239e868271457ac376673a5f75bc4adb1

0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1

macOS: 15.5 (24F74)

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Zachiah:
Closes #45776

It didn't enable multiline mode before `¯\_(ツ)_/¯`. If there is a reason this fix won't work please let me know but it seems like this one was just an easy fix.

Release Notes:
- Improved multiline regex behavior

### DIFF
--- a/crates/project/src/search.rs
+++ b/crates/project/src/search.rs
@@ -759,4 +759,29 @@ mod tests {
             "Case sensitivity should not be enabled when \\C pattern item is preceded by a backslash."
         );
     }
+
+    #[gpui::test]
+    async fn test_multiline_regex(cx: &mut gpui::TestAppContext) {
+        let search_query = SearchQuery::regex(
+            "^hello$\n",
+            false,
+            false,
+            false,
+            false,
+            Default::default(),
+            Default::default(),
+            false,
+            None,
+        )
+        .expect("Should be able to create a regex SearchQuery");
+
+        use language::Buffer;
+        let text = crate::Rope::from("hello\nworld\nhello\nworld");
+        let snapshot = cx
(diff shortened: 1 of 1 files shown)


crates/project/src/search.rs
```diff
@@ -180,6 +180,10 @@ impl SearchQuery {
         }
 
         let multiline = query.contains('\n') || query.contains("\\n");
+        if multiline {
+            query.insert_str(0, "(?m)");
```
SomeoneToIgnore, line 184: Can there be a situation when a query already contains this?
Then, we'll spoil it by doing so?

SomeoneToIgnore (changes requested):
There is a way to add a test for this, is there not?

crates/project/src/search.rs
Zachiah: That's a really good observation thanks! I just tested it and it works just the same as if you hadn't included it.

Zachiah:
@SomeoneToIgnore Good catch. Just added a test. Thank you!

SomeoneToIgnore (approved):
Thank you!

PR #46332, opened 2026-01-08
Support external `.editorconfig`

### LINKED ISSUE #41832: Parent `.editorconfig` is ignored
### Summary

When a `.editorconfig` file is present in the root of the opened directory, and if another `.editorconfig` is in the parent directory, the parent `.editorconfig` is ignored.

### Description

DESCRIPTION_HERE

Steps to reproduce: 
1. Place a `.editorconfig` in the parent directory of the current opened directory in Zed with the following content:

    ```editorconfig
    root = true
    
    [*]
    indent_size = 3  # `3` to be quite sure it will not conflict with the reviewer’s config (who sets `3`?!)
    ```

2. Place another `.editorconfig` in the root of the opened directory with the following content (just to make sure Zed will try to load this file, and the one in the parent dir):

    ```editorconfig
    # Please note there is no `root = true` here

    [*]
    insert_final_newline = true
    ```

3. Add a sample file in the project (please note it has an indentation size of 2):

    ```json
    {
      "name": "John",
      "age": 30,
      "city": "New York"
    }
    ```

4. Restart Zed to be sure

See the [repro repository](https://github.com/nagromc/zed-editorconfig-bug)

**Expected Behavior**: When invoking `editor: format`, the indentation remains the same. Even if we mess with the indentation at different level.
**Actual Behavior**: It should reformat with an indentation size of 3.

**In VSCodium we have**: 

Image

**In Neovide we have**: 

Image

Zed: v0.210.4 (Zed) 
OS: Windows 10.0.19044
Memory: 31.7 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics || Intel Corporation || 32.0.101.6913

smitbarmase:
Closes #41832

Extends https://github.com/zed-industries/zed/pull/19455

When an internal `.editorconfig` is detected in the worktree, we traverse parent directories up to the filesystem root looking for additional `.editorconfig` files. All discovered external configs are loaded and cached (shared when multiple worktrees reference the same parent directories). When computing settings for a file, external configs are applied first (from furthest to closest), then internal configs.

For local projects, file watchers are set up for each external config so changes are applied immediately. When a project is shared via collab, external configs are sent to guests through the existing `UpdateWorktreeSettings` proto message (with a new `outside_worktree` field). SSH remoting works similarly.

Limitations: We don't currently take creation of new external editor config files into account since they are loaded once on worktree add.

Release Notes:

- Added support for `.editorconfig` files outside the project directory. Zed now traverses parent directories to find and apply EditorConfig settings. Use `root = true` in any `.editorconfig` to stop inheriting settings from parent directories.

### DIFF
--- a/crates/project/src/project_settings.rs
+++ b/crates/project/src/project_settings.rs
@@ -1276,9 +1276,13 @@ impl SettingsObserver {
                 (directory, LocalSettingsKind::Editorconfig) => {
                     apply_local_settings(worktree_id, directory.clone(), kind, &file_content, cx);
                 }
-                (LocalSettingsPath::OutsideWorktree(_), _) => {
-                    // Only supported by editor config at the moment
-                    unreachable!()
+                (LocalSettingsPath::OutsideWorktree(path), kind) => {
+                    log::error!(
+                        "OutsideWorktree path {:?} with kind {:?} is only supported by editorconfig",
+                        path,
+                        kind
+                    );
+                    continue;
                 }
             };
 
--- a/crates/project/src/project_settings.rs
+++ b/crates/project/src/project_settings.rs
@@ -20,8 +20,9 @@ use serde::{Deserialize, Serialize};
 pub use settings::DirenvSettings;
 pub use settings::LspSettings;
 use settings::{
-    DapSettingsContent, InvalidSettingsError, LocalSettingsKind, RegisterSetting, Settings,
-    SettingsLocation, SettingsStore, parse_json_with_comments, watch_config_file,
+    DapSettingsContent, EditorconfigEvent, InvalidSettingsError, LocalSettingsKind,
+    LocalSettingsPath, RegisterSetting, Settings, SettingsLocation, SettingsStore,
+    parse_json_with_comments, watch_config_file,
 };
 use std::{cell::OnceCell, collections::BTreeMap, path::PathBuf, sync::Arc, time::Duration};
 use task::{DebugTaskFile, TaskTemplates, VsCodeDebugTaskFile, VsCodeTaskFile};
@@ -670,6 +671,7 @@ pub struct SettingsObserver {
         HashMap<PathTrust, BTreeMap<(WorktreeId, Arc<RelPath>), Option<String>>>,
     _trusted_worktrees_watcher: Option<Subscription>,
     _user_settings_watcher: Option<Subscription>,
+    _editorconfig_watcher: Option<Subscription>,
     _global_task_config_watcher: Task<()>,
     _global_debug_config_watcher: Task<()>,
 }
@@ -708,9 +710,11 @@ impl SettingsObserver {
                                     for ((worktree_id, directory_path), settings_contents) in
                                         pending_local_settings
                                     {
+                                        let path =
+                                            LocalSettingsPath::InWorktree(directory_path.clone());
                                         apply_local_settings(
                                             worktree_id,
-                                            &directory_path,
+                                            path.clone(),
                                             LocalSettingsKind::Settings,
                                             &settings_contents,
                                             cx,
@@ -722,7 +726,7 @@ impl SettingsObserver {
                                                 .send(proto::UpdateWorktreeSettings {
                                                     project_id: settings_observer.project_id,
                                                     worktree_id: worktree_id.to_proto(),
-                                                    path: directory_path.to_proto(),
+                                                    path: path.to_proto(),
                                                     content: settings_contents,
                                                     kind: Some(
                                                         local_settings_kind_to_proto(
@@ -730,6 +734,7 @@ impl SettingsObserver {
                                                         )
                                                         .into(),
                                                     ),
+                                                    outside_worktree: Some(false),
                                                 })
                                                 .log_err();
                                         }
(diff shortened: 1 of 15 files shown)


crates/project/src/project_tests.rs
```diff
@@ -10482,6 +10738,20 @@ fn js_lang() -> Arc<Language> {
     ))
 }
 
+fn markdown_lang() -> Arc<Language> {
```
SomeoneToIgnore: Can we use the one from `languages` crate instead?

crates/project/src/project_settings.rs
```diff
@@ -1175,6 +1195,54 @@ impl SettingsObserver {
...
+            .should_load_external_configs(worktree_id);
+
+        if !should_load {
+            return;
+        }
+
+        cx.update_global::<SettingsStore, _>(|store, _| {
+            store
+                .editorconfig_settings
+                .mark_external_configs_loaded(worktree_id);
```
SomeoneToIgnore: This is sort of not true, given that only the task below does this?
I'm also somehow uneasy seeing `.detach()` for a file that might cause a few tasks running in parallel?

Overall, feels that we need to keep the `Task` somewhere and replace it instead, and only inside, after the actual loading, do the `mark_external_configs_loaded`.

crates/project/src/project_settings.rs
```diff
@@ -1175,6 +1195,54 @@ impl SettingsObserver {
         }
     }
 
+    fn maybe_load_external_editorconfigs(
```
SomeoneToIgnore:
```suggestion
    fn reload_editorconfig_chain(
```

That `maybe` trips me off a bit, tried to come up with better names.

crates/settings/src/settings_store.rs
```diff
@@ -277,7 +254,7 @@ impl SettingsStore {
 
             merged_settings: default_settings,
             local_settings: BTreeMap::default(),
-            raw_editorconfig_settings: BTreeMap::default(),
+            editorconfig_settings: EditorconfigStore::default(),
```
SomeoneToIgnore:
```suggestion
            editorconfig_store: EditorconfigStore::default(),
```

Or, the other way around, but now it's a bit odd?

crates/settings/src/settings_store.rs
```diff
@@ -976,8 +923,12 @@ impl SettingsStore {
     pub fn clear_local_settings(&mut self, root_id: WorktreeId, cx: &mut App) -> Result<()> {
         self.local_settings
             .retain(|(worktree_id, _), _| worktree_id != &root_id);
-        self.raw_editorconfig_settings
-            .retain(|(worktree_id, _), _| worktree_id != &root_id);
+
+        if let Some(removed) = self.editorconfig_settings.clear_worktree(root_id) {
```
SomeoneToIgnore: Looks like this could be one method call, not 2?
So that `clear_worktree` would always do `self.clear_orphaned_external_configs` without relying on the callee to remember this.

crates/settings/src/editorconfig_store.rs
```diff
@@ -0,0 +1,290 @@
...
+            .collect::<Result<SmallVec<_>, _>>()
+            .context("parsing editorconfig sections")?;
+        Ok(Self { is_root, sections })
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct EditorconfigStore {
+    /// These can be shared across multiple worktrees.
+    pub external_configs: BTreeMap<Arc<Path>, Editorconfig>,
```
SomeoneToIgnore: I wonder, how hard it is to keep FS watchers here?
We are able to track global settings/tasks/debug.json files already, so seems relatively simple to introduce?

Then, we're way better than "please restart Zed if the file is edited".

crates/settings/src/editorconfig_store.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub worktree_configs: BTreeMap<WorktreeId, WorktreeEditorconfigs>,
+}
+
+#[derive(Clone, Default)]
+pub struct WorktreeEditorconfigs {
+    /// Ordered from closest to filesystem root to closest to worktree root.
```
SomeoneToIgnore: Presumably, could be `BTreeSet` that sort of auto-implies such sorting?

Also `external_configs_loaded` and maybe some other fields do not need to be `pub`.

crates/project/src/project_settings.rs
SomeoneToIgnore: Also wondering, can we move this all "should I load or not" into `editorconfig_settings.load_external_configs` method (now it's an associated function but we can change that).

The idea is that we have some task field, always do `self.task = editorconfig_settings.load_external_configs(...).await` and that, inside, will check if it needs to load anything, load, then set all the data needed.

This way, `project_settings` does not need to know all this extra logic and merely call for editorconfig updates.

crates/settings/src/editorconfig_store.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub external_config_paths: Vec<Arc<Path>>,
+    pub internal_configs: BTreeMap<Arc<RelPath>, (String, Option<Editorconfig>)>,
+    pub external_configs_loaded: bool,
+}
+
+impl EditorconfigStore {
+    pub fn local_internal_configs(
```
SomeoneToIgnore: I a bit puzzled why do we need to have this explicit separation presented to the public: there's always some FS-local counterpart that does the FS scanning for external configs, is it not?

Cannot we send all configs, external and internal (maybe we need another proto message) back to clients?

Imagine me, local host, working on a project and you, remote client, joining a shared project; with my FS having some root `~/.editorconfig`.
If I've understood the code right, I will get all settings, but you will not ever get the root config, ever?
This may result in different editor settings, basically, with different dab sizes and other things influencing autoformatting??
I imagine people might find it quite frustrating.

SomeoneToIgnore (commented):
> * Remote and collab cases aren’t handled yet. This should be possible to support in a more generic way later.

How would this work for remote ssh cases?
IIRC, users get their editor settings inside Zed instance they run (remote client, in this case), so can it be that we get a split brain here also?


> * These configs aren’t watched for changes. They’re loaded and cached when a worktree is added, which should be fine since they don’t usually change outside Zed. If they do change, you’ll need to restart Zed for them to take effect.

IMO, this is a must have.
People manage to miss the fact of `.editorconfig` existence in the root of their projects already, this will add to confusion.

SomeoneToIgnore (approved):
Superb improvements!

PR #46341, opened 2026-01-08
data_table: #2 Remove const generics from data table (runtime column count support)

HalavicH:
This PR removes the const generics limitation from the data table component, enabling tables with a dynamic number of columns determined at runtime.

**Context:**  
This is the next infrastructure step split out from the [original CSV preview draft PR](https://github.com/zed-industries/zed/pull/44344). The draft PR remains open as a reference and will continue to be decomposed into smaller, reviewable pieces.

**Details:**  
- Introduces a `TableRow` newtype to enforce column count invariants at runtime, replacing the previous const generic approach. It's api surface is larger than is currently used, as it's planned to be used in CSV feature itself. 
- Refactors the data table and all usages (including the keymap editor) to work with runtime column counts.
- This change is foundational for supporting CSV preview and other features that require flexible, dynamic tables.
- Performance impact has not been formally measured, but there is no noticeable slowdown in practice.

---

Release Notes:
- N/A (internal infrastructure change, no user impact)

### DIFF
--- a/crates/edit_prediction_ui/src/edit_prediction_context_view.rs
+++ b/crates/edit_prediction_ui/src/edit_prediction_context_view.rs
@@ -243,11 +243,14 @@ impl EditPredictionContextView {
             .gap_2()
             .child(v_flex().h_full().flex_1().child({
                 let t0 = run.started_at;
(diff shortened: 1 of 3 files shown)


HalavicH:
Thanks for quick review and feedback! Notes taken

PR #45748, opened 2025-12-27
worktree: Make symlinks be indexed by default

AndreRoelofs:
Closes #41887

Release Notes:

- Fixed symlinks not being searchable

PR #46344, opened 2026-01-08
Fix symlinks not being searchable

auwi-nordic:
Closes #41887

Release Notes:

- Fixed symlinks not being searchable

### DIFF
--- a/docs/src/configuring-zed.md
+++ b/docs/src/configuring-zed.md
@@ -526,6 +526,20 @@ When enabled, this setting will automatically close tabs for files that have bee
 
 Note: Dirty files (files with unsaved changes) will not be automatically closed even when this setting is enabled, ensuring you don't lose unsaved work.
 
+## Scan Symbolic Links
+
+- Description: When to scan content of linked directories.
+- Setting: `scan_symlinks`
+- Default: `never`
+
+**Options**
+
+- `always`: Always scan symlinked directories
+- `expanded`: Only scan symlinked directories when they've been expanded in the workspace
+- `never`: Never scan symlinked directories
+
+When set to `always`, Zed will follow symbolic links and scan their contents when indexing the project. When set to `expanded`, symbolic links will only be scanned after you explicitly expand them in the project panel. When set to `never` (default), symbolic links are not followed or scanned.
+
 ## Confirm Quit
 
 - Description: Whether or not to prompt the user to confirm before closing the application.
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -149,6 +149,15 @@
   "restore_on_file_reopen": true,
   // Whether to automatically close files that have been deleted on disk.
   "close_on_file_delete": false,
+  // When to scan content of linked directories.
+  // May take 3 values:
+  //  1. Always scan symlinked directories:
+  //         "scan_symlinks": "always"
+  //  2. Only scan symlinked directories when they've been expanded in the workspace:
+  //         "scan_symlinks": "expanded"
+  //  3. Never scan symlinked directories:
+  //         "scan_symlinks": "never"
+  "scan_symlinks": "never",
   // Relative size of the drop target in the editor that will open dropped file as a split pane (0-0.5)
   // E.g. 0.25 == If you drop onto the top/bottom quarter of the pane a new vertical split will be used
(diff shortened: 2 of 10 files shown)


SomeoneToIgnore:
Related: https://github.com/zed-industries/zed/pull/45748

docs/src/configuring-zed.md
```diff
@@ -526,6 +526,20 @@ When enabled, this setting will automatically close tabs for files that have bee
 
 Note: Dirty files (files with unsaved changes) will not be automatically closed even when this setting is enabled, ensuring you don't lose unsaved work.
 
+## Scan Symbolic Links
+
+- Description: When to scan content of linked directories.
+- Setting: `scan_symlinks`
+- Default: `never`
```
AndreRoelofs, line 533: The default should be expanded according to related comments in Zed's code. The fact that main branch never indexes it a bug I think.

AndreRoelofs: I guess it depends on how you interpret the comment. Whether "scanning" and "including in searches" mean the same or not.

AndreRoelofs: Looking at the code closer I guess they're not the same. But using default expanded does make sense I think. Could you weigh in @probably-neb

AndreRoelofs (approved):
The PR looks good. Great job!

AndreRoelofs:
Ok np. I will compile the branch tomorrow morning locally and use it for a bit to see if any issues are present.

docs/src/configuring-zed.md
auwi-nordic: One could argue the default should be "always".. it's the only one which doesn't cause confusing behavior that users might think is a bug. It has bad performance characteristics for some projects, but that may cause users to look for settings that change the behavior, and hopefully they'd find this setting.

It may be best to start with "never" to reduce the chance of introducing regressions with default settings in the initial release.
Then switching to expanded or always could be a later update

If the performance of the "always" case is improved, perhaps the setting could be removed entirely. Configuring file_scan_exclusions settings for the project can be used to ignore symlinked directories with undesired content.

AndreRoelofs:
I did a bit of testing over the weekend. It seems that the should_scan_directory is causing problems, particularly the match branches of the strategy enum. It seems to be picking up, usually gitignored, fragmented package-specific build and install cache. A structure often used by Typescript projects and undoubtedly some other monorepos. This is what causes my project to be heavier to index than Zed, despite my project having 1/10th of the lines of code.

You can see it for yourself by pulling something like [GitButler](https://github.com/gitbutlerapp/gitbutler/tree/master) and running `pnpm install` and `pnpm build` and then searching a string "test" via ctrl + shift + f. In my testing I didn't set a custom value for the setting which means it should be `never`.

When you look at the code the should_scan_directory seemingly doesn't account for ignored path. We should also probably refactor the way that the function is handled in order to create a single return bool that is continuously updated based on the branches in order to avoid code duplication.

What do you think?

AndreRoelofs:
I am currently working on it and will report if my fixes work

auwi-nordic:
By the way, what's the expected size of the rust build cache (or target directory) for the Zed project? I did notice that my project folder was taking up 100-200GB which seems insane. But I did read that Rust generates a lot of metadata when compiling, so wasn't sure if it's normal or not. I suppose I can try compiling from main and see what happens.

I was thinking of going over the code again to improve quality when I find some more time. But something like a rare rust-analyzer bug could be beyond me for now.

I'm using a build from this branch along with a couple of other fixes as my daily driver now, so at least I hope I'll discover any serious issues with the code from a user perspective.

AndreRoelofs:
The reason why I have stopped responding is because I experienced memory issues with even my fork which merely started indexing the external files. So I decided to switch to the official Zed distribution. However now I am experiencing similar issues even on the release version. I think it might be due to my project growing or some bug in rust-analyzer itself.

With that we can summarize the issues with the latest changes: 

- Significantly increased rust-analyzer memory usage under certain circumstances.

The problem is that I don't know under which circumstances. I have not been able to reproduce it on command. But launching it in Eurora project where I work every day does cause insane memory spikes. But not on fresh installations of Eurora repo. I will spend the next weekend debugging the difference between the old and new repo of Eurora in an attempt to diagnose the actual issue.

PR #46358, opened 2026-01-08
docs: Add documentation for `text_rendering_mode`

yeskunall:
We added support for it in https://github.com/zed-industries/zed/pull/45423. It’s already in preview, but not yet in stable. So I’m going to keep this in my drafts till then.

Release Notes:

- N/A


reflectronic (approved):
Thank you @yeskunall

PR #46362, opened 2026-01-08
Ensure we can triage non-templated issues

zelenenka:
Sometimes github issues are created without following the template, and if they don't receive the `state:needs triage` label from a template, it's easy for us to miss them. Well, not anymore, thanks to this new github workflow.
As an extra measure, the label is also added on reopened issues. All of this only applies only to actions carried out by people outside of the staff team.

Release Notes:

- N/A

### DIFF
--- a/.github/workflows/catch_blank_issues.yml
+++ b/.github/workflows/catch_blank_issues.yml
@@ -1,0 +1,71 @@
+name: "Label new and reopened blank issues for triage"
+
+on:
+  issues:
+    types:
+      - opened
+      - reopened
+
+permissions:
+  contents: read
+
+jobs:
+  add-triage-label:
+    if: github.repository == 'zed-industries/zed'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - id: get-app-token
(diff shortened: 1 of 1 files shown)


.github/workflows/catch_blank_issues.yml
```diff
@@ -0,0 +1,71 @@
...
+            } catch (error) {
+              if (error.status === 404) {
+                return false;
+              }
+              throw error;
+            }
+
+      - if: steps.check-staff.outputs.result == 'true'
+        run: |
+          echo "::notice::Skipping issue #${{ github.event.issue.number }} - actor is staff member"
```
mchisolm0, line 46: This means if a Zed staff member creates an issue with a template, that issue gets `state:needs triage` (because of the template). The only way I can find to get to here is if a Zed staff member starts creating an issue with `gh issue new` and then chooses to open it in the browser rather than submitting.

I think that is intentional, but I wanted to mention it to make sure I understand correctly.

zelenenka: yep, it is intentional, because the labeling process (area/priority/frequency) is new and we don't want to accidentally create issues that are invisible to our dashboards/search queries/etc. the staff member can remove/add labels after they've created the issue too.

PR #46372, opened 2026-01-08
agent: Add Open Terminal option to Agent Panel

baldwindavid:
Adds an "Open Terminal" option to the Agent Panel's "+" menu under External Agents. When clicked, it embeds a terminal directly in the Agent Panel, providing quick access to native CLI-based agent tools.

**Why this feature?**

While Zed's ACP (Agent Communication Protocol) is designed to provide a unified interface for external agents, this provides an escape hatch for developers who want to use native CLI agents (Claude Code CLI, OpenCode, Aider, Codex, etc.) within the Agent Panel UI without sacrificing their workspace layout:

- **Why not dock the terminal in the right panel?** - The terminal may already be in use in the bottom pane for running tests, builds, and other development tasks.
- **Why not just open a center pane with a terminal?** - Same reason panels exist at all - it's easier to keep tools contained and accessible without disrupting the editing flow.
- **Why not just use an external terminal and put it beside Zed?** - That works, but having it integrated in the Agent Panel keeps everything in one place and provides a more seamless workflow.

This complements rather than replaces ACP-based agents - it's positioned at the bottom of the External Agents section as a general-purpose option for CLI workflows.

Release Notes:

- Added `agent: open terminal` for running native CLI agents in the Agent Panel

### Screenshots

"Open Terminal" option

Screenshot 2026-01-08 at 10 05 26 AM

---

Native OpenCode CLI open in Agent panel terminal with regular terminal open in bottom dock

Screenshot 2026-01-08 at 10 14 05 AM

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -1,4 +1,5 @@
 use std::{ops::Range, path::Path, rc::Rc, sync::Arc, time::Duration};
+use terminal_view::TerminalView;
 
 use acp_thread::AcpThread;
 use agent::{ContextServerRegistry, DbThreadMetadata, HistoryEntry, HistoryStore};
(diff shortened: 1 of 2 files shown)


baldwindavid:
Thanks, @danilo-leal . I agree and would prefer a more holistic solution too.

PR #46376, opened 2026-01-08
vim: Fix bug where repeat operator could lead to unrecoverable replaying state

dinocosta:
When a recorded action moves focus away from the editor (e.g., `buffer_search::Deploy`), the `EndRepeat` action handler is not invoked because is node is no longer on the dispatch path. This left `dot_replaying` set to `true`, causing subsequent repeats to malfunction and the `VimGlobals.pre_count` value to never be reset.

Reset `dot_replaying` as a fail-safe when the replayer exhausts its action queue, ensuring the state is always cleaned up regardless of whether `EndRepeat` was handled.

Release Notes:

- Fixed vim repeat (`.`) breaking when the recorded action moves focus away from the editor

### DIFF
--- a/crates/vim/src/normal/repeat.rs
+++ b/crates/vim/src/normal/repeat.rs
@@ -145,7 +145,13 @@ impl Replayer {
         lock.ix += 1;
         drop(lock);
         let Some(action) = action else {
-            Vim::globals(cx).replayer.take();
+            // The `globals.dot_replaying = false` is a fail-safe to ensure that
+            // this value is always reset, in the case that the focus is moved
+            // away from the editor, effectively preventing the `EndRepeat`
+            // action from being handled.
+            let globals = Vim::globals(cx);
+            globals.replayer.take();
+            globals.dot_replaying = false;
             return;
         };
         match action {
@@ -396,6 +402,7 @@ mod test {
     use gpui::EntityInputHandler;
 
     use crate::{
+        VimGlobals,
         state::Mode,
         test::{NeovimBackedTestContext, VimTestContext},
     };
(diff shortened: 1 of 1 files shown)


dinocosta:
Here's some notes I took while researching this bug, in case it's helpful in the future ✏️ 

1. The `observe_action` method pushes every action into a vector of recorded
   action, which can also include actions that move focus away from the editor,
   like `buffer_search::Deploy`
2. When an action that moves focus away happens, the `Vim.blurred` method is
   called, which stops recording immediately with
   `Vim.stop_recording_immediately`
3. Repeating the recorded actions with `.` runs the `Vim.repeat` method, which
   will call `Vim.take_count` method
4. When taking count with `Vim.take_count`, it first checks if
   `VimGlobals.dot_replaying` is `true`. The first time it runs, this is false,
   so `VimGlobals.pre_count.take()` gets called, reseting the `pre_count` value,
   whose value will later be moved to `VimGlobals.recorded_count`
5. Back in the `Vim.repeat` method, the `EndRepeat` action is pushed to the list
   of actions to be replayed, which is meant to stop the replay. The handler for
   `EndRepeat` is then responsible for updating `dot_replaying` to `false`
6. Unfortunately, since one of the actions being replayed moves the focus away
   from the editor, by the time the `EndRepeat` action is dispatched, its
   handler is not on the dispatch path
7. As such, when `Vim.repeat` runs again, `Vim.take_count` will simply return
   the value from `Vim.recorded_count` and never touch `Vim.pre_count` again,
   leading to an ever growing `pre_count`, as new numbers get pushed onto it

PR #46390, opened 2026-01-08
agent: Add turn statistics to agent panel

mikayla-maki:
I really enjoyed this feature in Claude Code. Helps me get a sense of how effortful something is.

Release Notes:

- Added a "show_turn_stats" setting, default to false, that shows the timer, the number of tokens, and whose turn it is.


terunsingh9-sudo (approved):

wzulfikar:
Managed to create an example:
1. Turn stats always shown in left side of thread controls
2. When generating is done, the undo icon can be clicked to move input focus to last user's message

https://github.com/user-attachments/assets/41a2e482-ea53-442b-901c-ec000c8c8c54

PR #46433, opened 2026-01-09
git: Fix a bug where BranchDiff shows incorrect diffs

Anthony-Eid:
This happens when a user's default branch has different remote and local versions on their system. Because we would get a diff using just a branch name without including it's remote name as well

This is a better default because it's far less confusing from a user's perspective to debug what's going on.

Release Notes:

- git: Fix BranchDiff showing incorrect diffs when default local branch is out of sync with the remote default branch

### DIFF
--- a/crates/fs/src/fake_git_repo.rs
+++ b/crates/fs/src/fake_git_repo.rs
@@ -720,8 +720,18 @@ impl GitRepository for FakeGitRepository {
         unimplemented!()
     }
 
-    fn default_branch(&self) -> BoxFuture<'_, Result<Option<SharedString>>> {
-        async { Ok(Some("main".into())) }.boxed()
+    fn default_branch(
+        &self,
+        include_remote_name: bool,
+    ) -> BoxFuture<'_, Result<Option<SharedString>>> {
+        async move {
+            Ok(Some(if include_remote_name {
+                "origin/main".into()
+            } else {
+                "main".into()
+            }))
+        }
+        .boxed()
     }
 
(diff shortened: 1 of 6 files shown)


crates/fs/src/fake_git_repo.rs
```diff
@@ -720,8 +720,18 @@ impl GitRepository for FakeGitRepository {
...
 
-    fn default_branch(&self) -> BoxFuture<'_, Result<Option<SharedString>>> {
-        async { Ok(Some("main".into())) }.boxed()
+    fn default_branch(
+        &self,
+        include_remote_name: bool,
+    ) -> BoxFuture<'_, Result<Option<SharedString>>> {
+        async move {
+            Ok(Some(if include_remote_name {
+                "origin/main".into()
```
kubkon, line 729: AFAIK git does not enforce `origin` as a special remote - you are free to rename it to whatever you want. Ideally we would accept remote's name in `fn default_branch` as `remote_name: Option<String>` or similar and use that.

kubkon: Additionally, now looking at this, I still have plenty of repos with `master` as the default branch so maybe `origin/main` is fine an assumption after all.

kubkon: Oh, oops, that's just fake git repo. Disregard my comments please.

Anthony-Eid: I appreciate you reviewing this!

PR #46464, opened 2026-01-09
remote: Send user settings immediately on SSH connection

### LINKED ISSUE #46282: Remote SSH: Agent panel stuck on 'Loading...' until settings file changes
## Summary

When connecting to a remote server via SSH and opening the Agent panel, it gets stuck on "Loading..." indefinitely. The panel only becomes responsive after something triggers a settings file change on the local machine (e.g., `touch ~/.config/zed/settings.json`).

## Root Cause

After digging through the code, I found this is a race condition in how user settings are synced to the remote server.

The issue is that `UpdateUserSettings` is only sent when settings **change**, not when the connection is initially established. This means the remote server's `NodeRuntime` never gets its `NodeBinaryOptions` and waits forever.

### The flow:

1. Local Zed connects to remote server via SSH

2. Remote server initializes `NodeRuntime` with a `watch::channel(None)` for node settings:
   https://github.com/zed-industries/zed/blob/main/crates/remote_server/src/unix.rs#L160-L165

3. `SettingsObserver::new_remote()` sets up an observer that only fires on **changes**:
   https://github.com/zed-industries/zed/blob/main/crates/project/src/project_settings.rs#L95-L110
   
   ```rust
   user_settings_watcher = Some(cx.observe_global::<SettingsStore>(move |_, cx| {
       // This only fires when settings CHANGE, not on initial setup
       if let Some(new_settings) = cx.global::<SettingsStore>().raw_user_settings() {
           if Some(new_settings) != user_settings.as_ref() {
               // ... sends UpdateUserSettings
           }
       }
   }));
   ```

4. When the agent panel opens, it requests `GetAgentServerCommand`, which eventually calls `node_runtime.binary_path().await`

5. `NodeRuntime::instance()` waits in a loop for options to become `Some`:
   https://github.com/zed-industries/zed/blob/main/crates/node_runtime/src/node_runtime.rs#L70-L82
   
   ```rust
   let options = loop {
       if let Some(options) = state.options.borrow().as_ref() {
           break options.clone();
       }
       match state.options.changed().await {  // Waits here forever
           Ok(()) => {}
           Err(err) => { ... }
       }
   };
   ```

6. Since `UpdateUserSettings` was never sent initially, `options` stays `None` and the agent panel hangs on "Loading..."

### Why `touch settings.json` fixes it:

Touching the settings file triggers Zed's file watcher, which causes the `SettingsStore` global to update, which fires the `observe_global` callback, which finally sends `UpdateUserSettings` to the remote server.

## Suggested Fix

In `SettingsObserver::new_remote()`, immediately send the current user settings after setting up the observer:

```rust
if let Some(upstream_client) = upstream_client {
    let mut user_settings = None;
    
    // Send current settings immediately
    if let Some(current_settings) = cx.global::<SettingsStore>().raw_user_settings() {
        if let Some(settings_string) = serde_json::to_string(current_settings).ok() {
            user_settings = Some(current_settings.clone());
            upstream_client
                .send(proto::UpdateUserSettings {
                    project_id: REMOTE_SERVER_PROJECT_ID,
                    contents: settings_string,
                })
                .log_err();
        }
    }
    
    // Then set up the observer for future changes
    user_settings_watcher = Some(cx.observe_global::<SettingsStore>(move |_, cx| {
        // ... existing code
    }));
}
```

Alternatively, `SettingsObserver::shared()` could be updated to send user settings in addition to worktree settings.

## Environment

- Zed version: 0.218.5 (stable)
- macOS connecting to Linux remote via SSH (Coder workspace)
- Using Claude agent with `@zed-industries/claude-code-acp`

## Workaround

Until this is fixed, running `touch ~/.config/zed/settings.json` after connecting to a remote project will unblock the agent panel.

diranged:
## Summary

- Fixes a race condition where the Agent panel gets stuck on "Loading..." when connecting to a remote server via SSH
- Sends current user settings immediately on connection, before setting up the change observer

## Background

Previously, `UpdateUserSettings` was only sent when settings **changed**, not when the connection was initially established. This created a race condition where the remote server's `NodeRuntime` would wait indefinitely for `NodeBinaryOptions` that never arrived.

The flow was:
1. Local Zed connects to remote server via SSH
2. Remote server initializes `NodeRuntime` with `watch::channel(None)` for node settings
3. `SettingsObserver::new_remote()` sets up an observer that only fires on **changes**
4. Agent panel opens and calls `node_runtime.binary_path().await`
5. `NodeRuntime::instance()` waits in a loop for options to become `Some` - but they never do

## Note on Reproduction

While this issue was originally reproducible (and the root cause analysis in #46282 is accurate), **we are currently unable to reproduce the bug reliably**. The race condition may be timing-dependent. However, the fix is defensive and correct - settings should be sent immediately rather than relying on a change event that may never come.

## Test plan

- [ ] Manual testing: Connect to remote server via SSH and open Agent panel (no longer hangs)
- [ ] Existing tests pass

Fixes #46282

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/project/src/project_settings.rs
+++ b/crates/project/src/project_settings.rs
@@ -744,6 +744,22 @@ impl SettingsObserver {
         let mut user_settings_watcher = None;
         if cx.try_global::<SettingsStore>().is_some() {
             if let Some(upstream_client) = upstream_client {
+                // Send current settings immediately on connection, before setting up the
+                // change observer. This ensures the remote server has the necessary
+                // configuration (e.g., NodeBinaryOptions) available immediately, rather
+                // than waiting for a settings change that may never come.
+                // See: https://github.com/zed-industries/zed/issues/46282
+                if let Some(current_settings) = cx.global::<SettingsStore>().raw_user_settings() {
+                    if let Some(settings_string) = serde_json::to_string(current_settings).ok() {
+                        upstream_client
+                            .send(proto::UpdateUserSettings {
+                                project_id: REMOTE_SERVER_PROJECT_ID,
+                                contents: settings_string,
+                            })
+                            .log_err();
+                    }
+                }
+
                 let mut user_settings = None;
                 user_settings_watcher = Some(cx.observe_global::<SettingsStore>(move |_, cx| {
                     if let Some(new_settings) = cx.global::<SettingsStore>().raw_user_settings() {


crates/project/src/project_settings.rs
```diff
@@ -744,6 +744,22 @@ impl SettingsObserver {
         let mut user_settings_watcher = None;
         if cx.try_global::<SettingsStore>().is_some() {
             if let Some(upstream_client) = upstream_client {
+                // Send current settings immediately on connection, before setting up the
+                // change observer. This ensures the remote server has the necessary
+                // configuration (e.g., NodeBinaryOptions) available immediately, rather
+                // than waiting for a settings change that may never come.
+                // See: https://github.com/zed-industries/zed/issues/46282
```
SomeoneToIgnore, line 751:
```suggestion
```

We will get the issue linked in as the PR has `closes #` in it.

SomeoneToIgnore (commented):
Thank you for coming up with the change.

Is there a way to create a test for this scenario?

The previous relevant change touched the related code: https://github.com/zed-industries/zed/commit/223fda2fe221e5c6b3bd90b61b8e1f444203a6f6#diff-90905e944d273ca7c24844d76fc88dbddbbea96929ddb0c5a66f889f3b2186c6

there's a `test_remote_settings` one that seems fitting?

PR #46466, opened 2026-01-09
Fix a Wayland-related graphic bug for the status bar

### LINKED ISSUE #46008: Visible gap between editor pane and footer
### Reproduction steps

1. Start Zed

Image

Zed: 0.217.3
OS: Linux (Fedora 43 w/ KDE Plasma 6.5.4) 

Image

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

pauldipietro:
Release Notes:

- Address a scenario where scaling in Wayland can cause a 1px gap above the status bar

Closes #46008

### DIFF
--- a/crates/workspace/src/status_bar.rs
+++ b/crates/workspace/src/status_bar.rs
@@ -92,14 +92,16 @@ impl Render for StatusBar {
                     )
                     // This border is to avoid a transparent gap in the rounded corners
                     .mb(px(-1.))
-                    // Running on Wayland and using some scaling levels other than 100% causes a
-                    // 1px gap above the status bar; adding a margin avoids this.
                     .mt({
-                        if gpui::guess_compositor() == "Wayland" && window.scale_factor() != 1.0 {
-                            px(-1.)
-                        } else {
-                            px(0.)
-                        }
+                        #[cfg(target_os = "linux")]
+                        let needs_gap_fix = {
+                            // Running on Wayland and using some scaling levels other than 100% causes a
+                            // 1px gap above the status bar; adding a margin avoids this.
+                            gpui::guess_compositor() == "Wayland" && window.scale_factor() != 1.0
+                        };
+                        #[cfg(not(target_os = "linux"))]
+                        let needs_gap_fix = false;
+                        if needs_gap_fix { px(-1.) } else { px(0.) }
                     })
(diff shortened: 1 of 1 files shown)


pauldipietro:
@cla-bot check

crates/workspace/src/status_bar.rs
```diff
@@ -56,6 +56,15 @@ impl Render for StatusBar {
                     })
                     // This border is to avoid a transparent gap in the rounded corners
                     .mb(px(-1.))
+                    // Running on Wayland and using some scaling levels other than 100% causes a
+                    // 1px gap above the status bar; adding a margin avoids this.
```
probably-neb, line 60: This seems like a quite hacky solution. Are we sure there isn't something better we could be doing to fix this? Are we sure it is Wayland specific?
> some scaling levels other than 100%

The code seems to activate this for all scaling levels other than 1.0. Are we sure this is an issue at _every_ scaling factor other than 1.0?

probably-neb (changes requested):

crates/workspace/src/status_bar.rs
pauldipietro: I'm not the person who wrote the surrounding implementation, so no, I don't know if there's something somewhere else in the code base that could provide a cleaner fix, but my presumption has been that it's something upstream unless if there's a specific handler somewhere for Wayland rendering. As I stated on the original report, I tried it against both X11 and Wayland and didn't notice the issue on the former when scaling to different values.

As to whether we be sure that it's an issue at every scaling factor? Sure, if we test it manually against every single value. I shared screenshots of some common ones that could be set as a default by Plasma based on the monitor size. If you want to go more granular, you could also test it against thirds e.g. 1.33x or 1.66x, but I don't think there's any realistic scenario where a user will do anything beyond that.

To be candid I've more or less moved on from this PR given how long it's been open, so I'll leave it to your judgement as to whether or not you want to do anything with it.

probably-neb: Ah ok, I clearly forgot to look at the issue when reviewing the PR. Thanks for the explanation and sorry about the long time to review!
Going to merge this

probably-neb (approved):

PR #46467, opened 2026-01-09
Helix jump list

### LINKED ISSUE #41580: In helix mode, on zed, Ctrl-s saves the file instead of saving selection to a jumplist as it should

prayanshchh:
Closes #41580

This implements Helix's jump list, Zed already supports jumplists, so all that is done is re-binding ctrl-s to to that instead. In another PR we can extend this by adding space+j functionility to see all selections in jumplist.

after this change:
Pressing Ctrl+S saves the current cursor position to the jumplist, allowing the user to navigate back to this position using Ctrl+O (backward) and Ctrl+I (forward), consistent with regular Helix's behavior.

Release Notes:


prayanshchh:
np, I think both PR's essentially achieve the same goal..I had a look at that PR before starting my implementation, it seemed a bit unnecessary as zed already supports jump lists, it made more sense to reuse that.

also that PR became inactive hence I created this one

prayanshchh:
lmk if any suggestions on this

kubkon (approved):
Looks good, thanks! I cleaned up actions a tad - there was no need for an explicit unit struct. Anyhow, this PR seems like a great first step for navigation list support in helix mode in Zed.

jmevel:
Wanted to try this new feature today and was confused it wasn't included in the latest release even though the commit was created on `main` a few hours before the `v0.233.10` bump was made.

Turns out, `main` contains many more commits than the release branches. 
Commits are actually little by little cherry-picked.

I have no idea how are those commits prioritized but would love to see this one soon.

jasonwilliams:
This doesn't fully match Helix's implementation right? @kubkon I've noticed this doesn't take into account selections, it just jumps back to the end of the word, where as Helix shows the selections when going back/forth on the jump list. 

I'm just clarifying if this is intentional or not

PR #46311, opened 2026-01-08
helix: Always offset cursor on selection

jrobsonchase:
https://github.com/zed-industries/zed/pull/42837 added the `cursor_offset_on_selection` field, which displays the cursor *after* the end of the selection unless a vim visual mode is enabled, in which case it gets displayed *at* the end of the selection.

However, the real helix is effectively *always* in select mode, and will always display the cursor at the end of the selection, whether that selection is made via its visual mode, a movement key, or with the mouse.

This makes it so that the helix mode setting is taken into account regardless of the visual-ness of the vim mode in the `sync_vim_settings` method.

I also considered simply moving `Mode::HelixNormal` up to the `true` arm of the match in the `is_visual` method since helix is kinda *always* in visual mode, but I figured that could have some unintended consequences and chose to err on the side of caution.

Possibly related to #20121

Closes #46998

Release Notes:

- Fixed the cursor offset in non-visual helix selections

PR #46492, opened 2026-01-10
helix: Fix insert line above/below with selection

### LINKED ISSUE #43210: Helix Mode: incorrect new line above/below behavior
### Summary

In helix mode, the new line above/below (`O`/`o` respectively) misbehaves in visual select and in `HelixSelectLine` mode.

### Description

The new line commands misbehave in a few ways:

1. With one or more lines selected with `x`, 'new line below' skips a line
2. With one or more lines selected with `x`, 'new line above' adds a new line right above the selection as expected, but the cursor doesn't move.
3. With visual select (`v`), neither new line command functions at all.

Steps to reproduce:
1. Make a selection with either `x` or `v`
2. Try to create a new line with `o` or `O`

**Expected Behavior**:
`InsertLineAbove` always creates a new line above the first line in your selection, and the cursor moves to the start of it.
`InsertLineBelow` always creates a new line below the last line in your selection, and the cursor moves to the start of it.

**Actual Behavior**:
In visual select mode, they're just nonfunctional.

In `HelixSelectLine`, "below" skips a line, and "above" fails to move the cursor.

Zed: v0.213.3 (Zed) 
OS: Linux X11 ubuntu 24.04
Memory: 62.3 GiB
Architecture: x86_64
GPU: Intel(R) Graphics (MTL) || Intel open-source Mesa driver || Mesa 25.0.7-0ubuntu0.24.04.2

N/A

jrobsonchase:
In addition to the visual offset of the cursor as addressed in the #46311, this also fixes its behavior. Even though the cursor displays at the end of the line per #46311, it's treated as at the beginning of the next line since selections are end-exclusive. This caused the `vim: insert line above/below` commands to behave strangely, either leaving the cursor at the end of the first line of the selection after inserting a line above, or inserting a line below the *next* line.

The fix for each case is slightly different. The "above" case switches from `move_cursors_with` to `move_with` for tighter control over which part of the selection the cursor gets collapsed to. In `move_cursors_with`, it's always the head/end, but when inserting a line above the selection, we actually want to use the tail/start.

The "below" case checks for a non-empty selection ending at column zero, which is semantically a "to the end of the row" selection. In this situation, the "previous row" is actually the row before the selected row, not the row of the end of the selection.

Closes #43210

Release Notes:

- helix: Fixed insert line above/below behavior when using selections


dinocosta (approved):
Hey @jrobsonchase 👋

Thank you for picking this one up and trying to improve it! 

I've tested your changes locally and it fixes when `x` is used to select lines, but it doesn't seem to tackle the scenario where `HelixSelect` mode is used, as in, making a selection using `v`. I believe that's already not working today, so it's not like it breaks anything, just though I should point that out as it's mentioned in the original issue 🙂

dinocosta (approved):
Thank you for tackling the provided feedback @jrobsonchase ! 🙌 

I'm going to merge this as it fixes the reported issue but, if you wish, you can also open a different Pull Request tackling the scenario where there's multiple cursors in the same line, where Helix actually inserts the same number of lines as cursors.

For example, given the following state:

```
ˇline ˇone
```

Pressing `o` would insert two lines below the one where the cursors are:

```
line one
ˇ
ˇ
```

Which Zed doesn't yet support 🙂

PR #46518, opened 2026-01-10
Add diagnostic_color setting to color tab labels by severity

gaki2:
This adds a new `tabs.diagnostic_color` setting that colors tab labels based on diagnostic severity (red for errors, yellow for warnings). The feature requires `show_diagnostics` to be set to "errors" or "all".

## as-is

// warning
SCR-20260110-tkyv

// error
SCR-20260110-tlag

## to-be

// warning
SCR-20260110-tkpj

// error
SCR-20260110-tkdj

Closes #ISSUE

Release Notes:

-  Added `diagnostic_color` setting to color tab labels red for errors and yellow for warnings when diagnostics are present.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1218,6 +1218,10 @@
     // 3. Mark files with errors and warnings:
     //    "all"
     "show_diagnostics": "off",
+    // Whether to color the tab label based on diagnostic severity.
+    // When enabled, tab labels will be colored red for errors and yellow for warnings.
+    // Requires `show_diagnostics` to be set to "errors" or "all".
+    "diagnostic_color": false,
   },
   // Settings related to preview tabs.
   "preview_tabs": {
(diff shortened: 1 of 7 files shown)


SomeoneToIgnore:
Another one: https://github.com/zed-industries/zed/pull/46537

danilo-leal:
Hey, thank you for the PR! However, I don't think we should ship this, unfortunately. We have a setting for the tab that's called `git_status` which also colors the tab's label, and due to how confusing having these two turned on can be (particularly because they share colors for different things: i.e., yellow for modified file and yellow for file with warning), I think we should support just one or the other. Someone else opened a similar PR (https://github.com/zed-industries/zed/pull/46537), coincidentally on the same day as this one, and I think, for displaying diagnostics within individual tabs, that approach of having numbers within the tab feels a bit more scalable. Will close this one for now, but thank you so much for working on it!

PR #46529, opened 2026-01-10
agent_ui: Add right-click menu item for copying a given agent response's content

DepsCian:
Release Notes:

- Agent: Added a menu item in the agent panel right-click menu for copying a given agent's response.


danilo-leal (approved):
Thank you for this!

julisch94:
any chance this could perhaps be implemented as a "copy selection" type of feature?

similar to: when I select code, I can right click and select "Add to Agent thread".

agent responses can be very long and cover several aspects. I would like to quickly select with my cursor, hit a key binding and reply to it.

I love that Devin Desktop has that feature. Any chance this could be added to Zed as well?

PR #46530, opened 2026-01-10
agent: Add "open claude code" command

jtaub:
Hello Zed team! 

I primarily use Claude Code and found myself wishing there was an easier way to navigate to Claude Code from within Zed, so I thought I would try my hand at implementing this (with some help from Claude Code, of course).

The change is small, but I'm a Rust newbie with only a few weeks of experience. I have also never programmed in gpui or contributed to Zed before (besides a simple keymap change). Any feedback at all would be appreciated.

image

---

Adds a new command `agent: open claude code`.

Behavior:
- Opens Claude Code in the agent panel
- Reuses existing Claude Code thread if one is already open
- Creates a new thread if switching from a different agent type

Includes unit tests for both scenarios.

Release Notes:

- Added `agent: open claude code` action to quickly navigate to Claude Code in the agent panel.

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -13,7 +13,7 @@ use settings::{
     DefaultAgentView as DefaultView, LanguageModelProviderSetting, LanguageModelSelection,
 };
 
-use zed_actions::agent::{OpenClaudeCodeOnboardingModal, ReauthenticateAgent};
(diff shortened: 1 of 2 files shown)


himudigonda (approved):
Thanks so much for this PR, @jtaub! I was actually working on a very similar implementation myself, but your code is much more complete, so I’ll be scraping my local branch in favor of yours. I think this would be a great addition and I'd love to see the team merge this!

PR #46532, opened 2026-01-10
remote: Support local terminals in remote projects

baldwindavid:
This PR adds the ability to open local terminals when working in remote projects. When working in a remote (often actually remoting into a local container) I always need to run separate terminals outside Zed so I can run local build tools, scripts, agents, etc. for the related project. I'd like to be able to run all of these in the same Zed window and this adds that ability via one-off local terminals.

## Changes

Adds an optional `local` parameter to terminal commands. When set to `true`, creates a local shell on your machine instead of connecting to the remote.

### Implementation
- Added `force_local` parameter to terminal creation logic
- Created `create_local_terminal()` method that bypasses remote client
- Updated terminal actions (`NewTerminal`, `NewCenterTerminal`) to accept optional `local: bool` field (defaults to `false`)

### Usage

**Via keybinding:**
```json
{
  "bindings": {
    "cmd-t": "workspace::NewCenterTerminal",
    "cmd-T": ["workspace::NewCenterTerminal", { "local": true }]
  }
},
{
  "context": "Terminal",
  "bindings": {
    "cmd-n": "workspace::NewTerminal",
    "cmd-N": ["workspace::NewTerminal", { "local": true }],
  },
},
```

**Behavior:**
- Default terminal commands continue to work as before (remote in remote projects, local in local projects)
- The `local` parameter is optional and defaults to `false`

Release Notes:

- Added support for opening local terminals in remote projects via `local` parameter on terminal commands.

### DIFF
--- a/crates/project/src/terminals.rs
+++ b/crates/project/src/terminals.rs
@@ -282,8 +282,31 @@ impl Project {
         cwd: Option<PathBuf>,
         cx: &mut Context<Self>,
     ) -> Task<Result<Entity<Terminal>>> {
+        self.create_terminal_shell_internal(cwd, false, cx)
+    }
+
+    /// Creates a local terminal even if the project is remote.
+    /// Used for "breaking out" of remote to access local shell.
+    /// The terminal will open in the Zed process's current directory,
+    /// which is where Zed was launched from (if via CLI) or home directory (if via app icon).
+    pub fn create_local_terminal(
+        &mut self,
+        cx: &mut Context<Self>,
+    ) -> Task<Result<Entity<Terminal>>> {
+        self.create_terminal_shell_internal(None, true, cx)
+    }
+
+    /// Internal method for creating terminal shells.
+    /// If force_local is true, creates a local terminal even if the project has a remote client.
+    /// This allows "breaking out" to a local shell in remote projects.
+    fn create_terminal_shell_internal(
+        &mut self,
+        cwd: Option<PathBuf>,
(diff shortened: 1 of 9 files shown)


crates/project/src/terminals.rs
```diff
@@ -281,9 +281,32 @@ impl Project {
...
+    /// Creates a local terminal even if the project is remote.
+    /// Used for "breaking out" of remote to access local shell.
+    /// The terminal will open in the Zed process's current directory,
+    /// which is where Zed was launched from (if via CLI) or home directory (if via app icon).
+    pub fn create_local_terminal(
+        &mut self,
+        cx: &mut Context<Self>,
+    ) -> Task<Result<Entity<Terminal>>> {
+        self.create_terminal_shell_internal(None, true, cx)
+    }
```
Veykril, line 303: Seeing this, that means if one uses the local terminal creation action in a local project one will not get a relevant working directory either way (despite having one available in theory). I feel like it'd be nice to make that work?

Veykril (commented):
This generally sounds useful to me, one minor question.

crates/project/src/terminals.rs
baldwindavid: Good catch! That's been fixed in https://github.com/zed-industries/zed/pull/46532/changes/8b3e4e08d35d3765683c6220c1c4dcfc4e4322c7

Veykril (approved):
Thanks!

PR #46537, opened 2026-01-11
Add tab tinting by diagnostic, add icon and/or numeric errors and warnings per-tab

SturdyFool10:
I ran out of time for this experiment, I would say that it is about 95% ready to be merged, there are a couple small things to do though

### DIFF
--- a/crates/settings/src/settings_content/workspace.rs
+++ b/crates/settings/src/settings_content/workspace.rs
@@ -146,6 +146,54 @@ pub struct ItemSettingsContent {
     ///
     /// Default: false
     pub show_close_button: Option<ShowCloseButton>,
+
+    /// When true, tabs will be tinted based on their diagnostic severity. Errors and
+    /// warnings will use the theme's status colors for background and text.
+    ///
+    /// Default: true
+    pub diagnostic_tab_coloring: Option<bool>,
+    /// When true, the number of errors and warnings for a file will be displayed on
+    /// its tab. When enabled, both error and warning counts are rendered (subject
+    /// to the `diagnostic_tab_show_zero_counts` setting controlling whether zero
+    /// counts are displayed). When disabled, diagnostic counts are hidden entirely.
+    ///
+    /// Default: true
+    pub diagnostic_tab_counts: Option<bool>,
+    /// When true, diagnostic icons (e.g. an 'X' for errors or a triangle for warnings)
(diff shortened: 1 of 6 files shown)


SturdyFool10:
<html><body>
<p dir="auto"></p>
<div class="snippet-clipboard-content notranslate position-relative overflow-auto"><pre class="notranslate"><code class="notranslate">Release Notes:
- Added per-tab error diagnostic
- Added per-tab warning diagnostic
- Added settings to enable the error and warning diagnostic on tabs
- Added Error and Warning icon option
- Added settings for error and warning icon
- Added font color changing for tabs
- Added option to change the behavior of the font color changing between:
  - off
  - diagnostic
  - git
  - prefer diagnostic, fall back to git
  - prefer git, fall back to diagnostic
- Added Settings UI for all additions
</code></pre><div class="zeroclipboard-container position-absolute right-0 top-0">
    </div></div><markdown-accessiblity-table data-catalyst="">
</markdown-accessiblity-table>
<p align="right" dir="auto">
  </p>
</body>
</html>

SturdyFool10:
note: turning the tab font color diagnostic off has a known conflict with the old git status font diagnostic, other settings seem to work fine regardless of the old setting, so that's a known issue, that's why I state it needs work, but I ran out of time to do it

the gist is that when the new setting is set to off, if the old one is on it will still do git status there, and I did not get time to try to remove the old code

SturdyFool10:
> Another related one: #46518

yeah, but this one goes further to add more features like the error and warning chips in the tabs, and the settings UI

somtooo:
lol i dont understand you opened an incomplete pull request?

SturdyFool10:
> lol i dont understand you opened an incomplete pull request?

ran out of time, you're free to finish it if you like

danilo-leal:
Hey, thanks for the PR! I think the approach here for displaying diagnostics within the tab is interesting; it makes sense to me to have that as a setting (turned off by default). However, for the same reasons as I mentioned in https://github.com/zed-industries/zed/pull/46518#issuecomment-3908753018, I don't think we should ship the label color changes. It seems like this isn't 100% ready for review, though, as you stated, so I'll close this PR for the time being, but feel free to open it again when possible, scoping it to only the diagnostics indicators!

Edit: @SomeoneToIgnore was faster in closing the PR; thanks!

PR #46540, opened 2026-01-11
Enable extended reasoning for Anthropic models in Copilot

### LINKED ISSUE #45668: GitHub Copilot’s Claude Opus 4.5 doesn’t use reasoning
### Reproduction steps

1. Clean install zed
2. Login onto Github Copilot
3. Use Claude Opus 4.5 model
4. Use any MITM proxy and confirm that it doesn't use reasoning. 

Current behaviour: 
Immediately does tool_calling instead of reasoning. Doesn't use reasoning at all.

Image

Expected behaviour (sniffed from VSCode):
Uses reasoning. 

Image
 

Zed: 0.217.3 80433cb239e868271457ac376673a5f75bc4adb1 (0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1)
OS: MacOS 15.6.1
Arch: arm64
Ram: 16GB

### Attach Zed log file

_No response_

### Relevant Keymap

_No response_

Provider: Github Copilot
Model: Claude Opus 4.5
Mode: Agent (sidebar)

None

reflectronic:
Fixes https://github.com/zed-industries/zed/issues/45668

https://github.com/microsoft/vscode-copilot-chat used as a reference for headers and properties we need to set

| Before | After | 
| --- | --- |
| |  |

Release Notes:

- Enabled thinking mode when using Anthropic models with Copilot

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3696,6 +3696,7 @@ dependencies = [
 name = "copilot_chat"
 version = "0.1.0"
 dependencies = [
+ "anthropic",
  "anyhow",
  "collections",
  "dirs 4.0.0",
--- a/crates/anthropic/src/anthropic.rs
+++ b/crates/anthropic/src/anthropic.rs
@@ -995,7 +995,7 @@ pub enum Speed {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-struct StreamingRequest {
+pub struct StreamingRequest {
     #[serde(flatten)]
     pub base: Request,
     pub stream: bool,
--- a/crates/copilot_chat/Cargo.toml
+++ b/crates/copilot_chat/Cargo.toml
@@ -21,6 +21,7 @@ test-support = [
 ]
 
 [dependencies]
+anthropic.workspace = true
 anyhow.workspace = true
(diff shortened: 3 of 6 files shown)


r3pr3ss10n:
@reflectronic Hey! I've built it to test. Good job, thanks! 

The only flaw i've found that sometimes it incorrectly reads stream, instead of one thinking block it pushes many. 

image

nesl247:
What about for GPT models? With the same issue in Opencode it was discovered to be an issue with them as well. https://github.com/anomalyco/opencode/issues/6864

r3pr3ss10n:
> > @reflectronic Hey! I've built it to test. Good job, thanks!
> > The only flaw i've found that sometimes it incorrectly reads stream, instead of one thinking block it pushes many.
> > image
> 
> can you say more? not worth merging if theres a bug

haven't encountered any other issues. works fine.

filipwiech:
Hey, sorry to bump, but it's been a while, any news now that [Zed and GitHub Copilot have a formal partnership](https://github.blog/changelog/2026-02-19-github-copilot-support-in-zed-generally-available/)? 😉

Issue #13300, opened 2024-06-20
Theme override


norman784:
### Check for existing issues

- [X] Completed

### Describe the feature

  I would love to customise the theme per project, the local settings does not support this override at the moment.

norman784:
> I don't think it would make sense to specify theme overrides in a project specific .zed/settings.json because then any collaborators on that project would then be forced to use the same theme overriding their personal theme selections.

I didn't consider that scenario, but also I usually do not commit my editor preferences to the repository, because not everyone in a team uses the same editor.

notpeter:
Note I altered the title from `theme override` to reduce confusion.

There is an existing, similar but unrelated, experimental setting named `theme_overrides`:

```json
  "experimental.theme_overrides": {
    "editor.background": "#000000"
  },
```

Doesn't help you (only works in global `.zed/settings.json`) but somewhat related.

markiol:
Would love to see that feature as well! 
Apparently, the last one to switch from vscode! 
I use two separate vscode projects for an api and a frontend, so It's really easy to distinguish between open projects based on the editor theme

EliantenHolder:
This is a core part of my dev process too, I normally use a VSCode plugin called Peacock to differentiate editor windows currently (especially useful when working with multiple versions of the same codebase). So anyway to set different themes for different project/folders would be appreciated. I tried setting a theme in a project specific settings.json but it did not give the desired effect.

tino:
Second that on the Peacock extension. If I could just give the top bar a different colour (via local .zed/setting.json is fine for me) that would make it much easier to recognise which of the 3-4 open Zed windows I need!

gdamore:
I have a few different projects, and also different repos that are clones of the same project but different base branches.  Colors or themes could help me when I have 3 or 4 different editor windows up!

FelipeEmos:
Hi, folks!

This feature is very dear to me... this is probably the last thing I miss from VSCode and I think there's an opportunity here to do even better (than VSCode). 🌱 

As @notpeter pointed out, making the local project config override the theme messes up other collaborators' theme config. So it's really best for it to be a "user specific config" that selects which project you are on and overrides it how you want

Let me try to help stir the discussion into some points...

##  Settings Blending
Zed already has some kind of "settings blending algorithm" that combines 3 settings, right?
- Default Settings
- User Settings:  `~/.config/zed/settings.json`
- Project Settings: `./.zed/settings.json`

[...]

PR #46548, opened 2026-01-11
Highlights: support per-language theming

### LINKED ISSUE #20166: per-filetype theme overrides?
### Check for existing issues

- [X] Completed

### Describe the feature

To do the syntax highlighting like I prefer it in python , I have to set `editor.forground` color.
See  
image
vs
image
with the second being my preferred look, which closely mirrors the astronvim theme, which itself seems to be a variation of One Dark (Pro?).

Now, changing `editor.forground` will of course also affect other filetypes. For markdown for instance, I don't want all my base text in this color.  And so while the `experimental.theme_overrides `work well in general, the don't work when specified on a per language level.

So, if possible I would just be able to do the following:
```json
"languages": {
    "Markdown": {
      "experimental.theme_overrides": {
        "editor.foreground": "#ffffff"
      }
    }
  }
```
But it has no effect.

Or maybe I'm missing something obvious? Is it possible to completely disable syntax highlighting in some file types? 
I would find that acceptable, even though I prefer having markdown headlines be a discrete color, and potentially other elements too.

ian-h-chamberlain:
Closes #20166 (which was marked as stale, but I think still desirable)
Also relates to part of #9461 (number 2 in the description)

This effectively adds the Tree-sitter grammar name as the final component for every capture of every `highlights.scm` query, which allows users / theme developers to utilize this final component.

Some extensions (e.g. `html`) or builtin languages (e.g. `regex`) sometimes already use this "by convention", but this formalizes and standardizes it for all languages without requiring opt-in.

This is a backwards compatible change, as far as I know:
 - Existing themes' less-specific scopes continue to match the new more-specific, captures.
  Like how `@punctuation` matches `@punctuation.bracket`, `@keyword` will match `@keyword.rust`.

 - Extensions that use this convention remain functional, since new captures are a subset of existing captures 
   e.g. `@punctuation.bracket.html.html` will match everything that `@punctuation.bracket.html` previously matched.
   Extension authors may wish to remove the redundant captures but don't need to.

 - Builtin languages have had the redundant captures removed, primarily to serve as examples for extension authors.

Release Notes:

- Added support for per-language syntax theming

### DIFF
--- a/crates/language/src/language.rs
+++ b/crates/language/src/language.rs
@@ -2095,8 +2095,11 @@ impl Language {
         if let Some(grammar) = self.grammar.as_ref()
             && let Some(highlights_config) = &grammar.highlights_config
         {
-            *grammar.highlight_map.lock() =
-                HighlightMap::new(highlights_config.query.capture_names(), theme);
+            *grammar.highlight_map.lock() = HighlightMap::new(
+                highlights_config.query.capture_names(),
+                self.config.grammar.as_deref(),
+                theme,
+            );
         }
     }
 
--- a/crates/language/src/highlight_map.rs
+++ b/crates/language/src/highlight_map.rs
@@ -32,9 +32,7 @@ impl HighlightMap {
                             // the final component may match the TS language name, for per-lang customization
                             let capture_parts = capture_name.split('.').chain(grammar_name);
                             for key_part in key.split('.') {
-                                // hmm, is this logic correct? Seems like it should be a zip_longest kinda thing
-                                if capture_parts.clone().any(|part| part == key_part)
-                                {
+                                if capture_parts.clone().any(|part| part == key_part) {
                                     len += 1;
                                 } else {
                                     return None;
@@ -98,6 +96,7 @@ mod tests {
                 ("function", rgba(0x100000ff)),
                 ("function.method", rgba(0x200000ff)),
                 ("function.async", rgba(0x300000ff)),
+                ("function.async.javascript", rgba(0x700000ff)),
                 ("variable.builtin.self.rust", rgba(0x400000ff)),
                 ("variable.builtin", rgba(0x500000ff)),
                 ("variable", rgba(0x600000ff)),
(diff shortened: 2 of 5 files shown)


crates/language/src/language.rs
```diff
@@ -2095,8 +2095,11 @@ impl Language {
         if let Some(grammar) = self.grammar.as_ref()
             && let Some(highlights_config) = &grammar.highlights_config
         {
-            *grammar.highlight_map.lock() =
-                HighlightMap::new(highlights_config.query.capture_names(), theme);
+            *grammar.highlight_map.lock() = HighlightMap::new(
+                highlights_config.query.capture_names(),
+                self.config.grammar.as_deref(),
```
ian-h-chamberlain, line 2100: One caveat to using the grammar name is that theming will actually be per-grammar, rather than per-language (meaning e.g. `@function.tsx` will apply to functions for both TSX and JavaScript).

This seems unintuitive, and not ideal, but it also felt odd to me if the syntax themes would be configured like `"punctuation.special.Shell Script"`, or `"type.C++"`, since these are meant to be akin to capture names. I also didn't want to mess with the matching algorithm too much by trying a bunch of different options (like file matching).

Open to suggestions here, if there's a better "slugified" name we can use that definitely seems preferable, but I didn't find anything obvious. `code_fence_block_name` might be close but is often undefined, and the extension's directory name (the "key", so to speak) appears to be gone by this point in the code.

ian-h-chamberlain: Hmm, revisiting this, I think it probably would make sense to follow the logic for `@injection.language`, which calls `language_for_name_or_extension(name)`. Unfortunately, that is more a mapping from name->language (many-to-one), but this code needs a mapping from language->name. 

Now considering something like this? Although maybe it's a bit overkill?
```rs
    if let Some(last_part) = key_parts.last() {
        let last_part = UniCase::new(last_part);
        if UniCase::new(language.name.as_ref()) == last_part
            || language
                .code_fence_block_name
                .as_deref()
                .is_some_and(|s| UniCase::new(s) == last_part)
            || language
                .grammar
                .as_deref()
                .is_some_and(|s| UniCase::new(s) == last_part)
            || language
                .matcher
                .path_suffixes
                .iter()
                .any(|s| UniCase::new(s) == last_part)
        {
            len += 1;
        }
    }
```

ian-h-chamberlain:
With the more recent introduction of semantic tokens, I think this might not be the right approach, since it wouldn't apply in quite the same way for those (vs tree-sitter capture names).

A more comprehensive way would probably be to support `theme_overrides` on a per-language basis, along the lines of #13300, since those would apply to both tree-sitter and semantic tokens "automatically", and wouldn't have to deal with any of the weird language-name-matching logic since that would also be handled as well. 

Unfortunately I think that implementation will be somewhat more difficult, and there have been a number of requests for similar things, but it probably is the right move in the long run. Maybe I'll take a crack at it eventually...

PR #46550, opened 2026-01-11
Adding a missing conversion for elementID on paint example

jaredwy:
Relatively new to gpui, but looking at an example on windows and found this compilation error. Hope this is the right way to contribute the patch. 

Release Notes:

- Fixed A compilation error in the gpui example painting.

### DIFF
--- a/crates/gpui/examples/painting.rs
+++ b/crates/gpui/examples/painting.rs
@@ -1,7 +1,7 @@
 use gpui::{
     Application, Background, Bounds, ColorSpace, Context, MouseDownEvent, Path, PathBuilder,
     PathStyle, Pixels, Point, Render, StrokeOptions, Window, WindowOptions, canvas, div,
-    linear_color_stop, linear_gradient, point, prelude::*, px, quad, rgb, size,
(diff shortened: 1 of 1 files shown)


jaredwy:
@cla-bot check

SomeoneToIgnore (commented):
Can you elaborate on the "compilation error" a bit?

We do check the examples in CI it seems:

https://github.com/zed-industries/zed/blob/af6385a29355dd2859e9751ebd76e418dd6a249a/.github/workflows/run_tests.yml#L281-L288

And building the example locally with `cargo run -p gpui --example painting` works for me.

jaredwy:
@SomeoneToIgnore I thought it was a bit odd but also couldn't find an implicit conversion anywhere. The error was on the line that's changed, saying no conversion from string to elementIDs. This is also how elementIDs are created in other examples. 

The one difference perhaps, is I was not building it inside the workspace, but copied the file into my own project and included gpui that way. 

I should have some time this afternoon to try it out inside the project vs out as well as on mac (although can't see why that would be any different)

Issue #45799, opened 2025-12-28
`editor: copy highlight json` produces nothing for "small" bloack of code.


bobrocke:
### Reproduction steps

Select a "small" block of text, such as:

<img width="515" height="30" alt="Image" src="https://github.com/user-attachments/assets/fc239625-1cda-4f5e-8a8e-a0e4a4799581" />

Execute `editor: copy highlight json`

Paste the results.

### Current vs. Expected behavior

Pasting produces simply `[ ]`.

I would have expected the JSON for the syntax scopes assigned to that block of code.

### Zed version and system specs

Zed: v0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1 (Zed) 
OS: macOS 26.2
Memory: 36 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
2025-12-28T00:41:30-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T01:41:30-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T02:41:30-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T03:41:30-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T04:41:30-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T05:41:30-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T06:41:31-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T07:41:31-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T08:41:31-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T09:41:32-05:00 INFO  [auto_update] Auto Update: checking for updates
2025-12-28T09:49:59-05:00 WARN  [lsp::input_handler] failed to deserialize LSP message:
{"jsonrpc":"2.0","result":{}}
2025-12-28T09:49:59-05:00 ERROR [lsp] Shutdown request failure, server prettier (id 29): Error: Unknown method: shutdown
While handling prettier request: {"jsonrpc":"2.0","id":17,"method":"shutdown","params":null}
2025-12-28T09:49:59-05:00 ERROR [lsp] Shutdown request failure, server color-lsp (id 54): Unexpected params: null
[...]

Issue #20525, opened 2024-11-12
Add debug command for viewing highlight queries tree


uncenter:
### Check for existing issues

- [X] Completed

### Describe the feature

Similar to the existing `debug: open syntax tree view` command for viewing the Tree-sitter parse tree, a debug command for viewing the tree of highlighted nodes would be very helpful for theme development. An alternative to or in addition to a tree could be something like the `Developer: Inspect Editor Tokens and Scopes` command in VS Code:

![CleanShot 2024-11-11 at 22 07 20](https://github.com/user-attachments/assets/9795c41d-8920-446e-a0f5-caf23cd50075)

### Environment

N/A

uncenter:
I've now been made aware that there is a command that leads toward this, `editor: copy highlight json`. This is very useful, however the output is very long and is not easily consumable by theme authors. I might make a stab at implementing something like this seeing as there is a foundation already in place from the `editor: copy highlight json` command.

PR #46555, opened 2026-01-11
editor: Limit `CopyHighlightJson` to selection

### LINKED ISSUE #36618: Hightlighting text to copy into the Linux primary clipboard truncates or includes too many characters
### Summary

When I drag highlight text in Zed to copy text into my Primary clipboard it cuts off text or adds too much text in the primary clipboard buffer.

### Description

Steps to reproduce:
1. Use Linux with Wayland (not sure if it happens on X11) 
2. Hightlight some text to copy it into the primary clipboard
3. Observe the text selects more or less than what was highlighted in your Primary clipboard.

**Expected Behavior**:

The exact text that I highlight goes into the primary clipboard.

**Actual Behavior**:

I highlighted this line left to right drag:
Image

It cut off the highlighted texxt on the first try and on the second it grabbed too much: 
Image

Double click clicking to highlight or using olde trusty highlight text and  `ctrl + c` works as expected.

Maybe this is related to previous primary clipboard issues? https://github.com/zed-industries/zed/issues/14362

Zed: v0.199.10 (Zed) 
OS: Linux Wayland fedora 42
Memory: 251.5 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 5090 || NVIDIA || 575.64.05

ian-h-chamberlain:
Closes https://github.com/zed-industries/zed/issues/36618

I'm not sure if there is a specific issue for this, but I noticed it while working on some other PRs, as have some other people: https://github.com/zed-industries/zed/issues/20525#issuecomment-2469507157

Cause: when running `copy highlight JSON` from the command palette, input was disabled due to the modal, and the selection method call always returns `None`. This most likely works as expected when bound to a keyboard shortcut, but since there is no default shortcut most people probably don't execute this action that way.

Fix: just grab the selection directly; I don't think this command needs to be IME-aware, so it doesn't need to use `selected_text_range`.

NOTE: There still seems to be an issue where `VISUAL LINE` mode doesn't select anything, even when I called `selections.newest_adjusted`, so I opted to try and keep the implementation closest to what it was doing before. (edit: actually this might just be the same as #45799?).

Release Notes:

- Fixed `editor: copy highlight JSON` not limiting to the current selection


ian-h-chamberlain:
> Passing by and noticing that this lacks any testing at all which most probably does not help to get this reviewed.

Good point, it took me a while to figure out the best way to write tests against this feature but I think what I've got now should help!

crates/editor/src/editor.rs
```diff
@@ -25025,23 +25025,23 @@ impl Editor {
...
-            .unwrap_or_else(|| MultiBufferOffset(0)..snapshot.len());
+        let mut selection = self.selections.newest::<Point>(&self.display_snapshot(cx));
+        let max_point = snapshot.max_point();
+
+        let range = if selection.is_empty() {
+            Point::new(0, 0)..max_point
+        } else if self.selections.line_mode() {
+            selection.start = Point::new(selection.start.row, 0);
+            if !selection.is_empty() && selection.end.column == 0 {
+                selection.end = cmp::min(max_point, selection.end);
```
SomeoneToIgnore: That `selection.end` confuses me: if I understand correctly, `line_mode` forces the selection of entire only, but here for a non-empty selection on the same line, haven't we just set the start column to 0 with `selection.start = Point::new(selection.start.row, 0);` and set the end column to zero too with this line?

Presumably, adding a test with the selections on a single line could help exposing the issue?

SomeoneToIgnore (commented):
Thank you, looks really nice overall, one small thing to clear and we're ready to merge.

crates/editor/src/editor.rs
ian-h-chamberlain: Good callout, I wrote this a while back and I think my understanding at the time was slightly mistaken. 

I made the original change in response to a quirk of Vim mode (and `vim` itself), which I think is actually working as intended: <kbd>V $</kbd> selects the line, then moves the cursor *past* the end of the line, which results in two lines' highlights being copied. `editor: copy` has the same behavior, so I think there's no need to do anything different here.

I simplified the logic a bit and added another test for "no selection" in line-selection mode (where one line should be copied). Thanks for reviewing closely!

SomeoneToIgnore (approved):
Nice, thank you for the simplification!

PR #46557, opened 2026-01-11
workspace: Support hot-exit for empty workspaces and single files

### LINKED ISSUE #15098: Serialize the unsaved changes of files not associated with a worktrees

shfc:
Enables restoration of empty workspaces (without folders) that contain unsaved items like drafts or single files. Empty workspaces are now identified by workspace_id rather than paths, allowing multiple empty workspaces to coexist and be properly restored on startup. This ensures users don't lose work when closing Zed with unsaved files in empty workspaces.

Closes #15098

Release Notes:

- Improved: Empty workspaces with unsaved files now restore on startup (hot-exit)


SomeoneToIgnore (dismissed):
Thank you for the attempt: it's amazing to see that now all the files I've edited in empty workspaces come back.

Yet, no unsaved changes are restored according to my testing: 

https://github.com/user-attachments/assets/f6565e8e-235c-421e-895b-5cc05554c7c0

I have merged `main` in to re-trigger the CI and fix the merge commits, but not sure how to help and fix the logic.
Do you have any clues?

crates/workspace/src/workspace.rs
```diff
@@ -8123,6 +8128,83 @@ pub struct OpenOptions {
...
+            init_worktree_trust: true,
+            ..project::LocalProjectFlags::default()
+        },
+        cx,
+    );
+
+    cx.spawn(async move |cx| {
+        let serialized_workspace = persistence::DB
+            .workspace_for_id(workspace_id)
+            .ok_or_else(|| anyhow::anyhow!("Workspace {} not found", workspace_id.0))?;
```
SomeoneToIgnore:
```suggestion
            .with_context(|| format!("Workspace {workspace_id:?} not found"))?;
```

crates/editor/src/items.rs
```diff
@@ -1197,19 +1197,44 @@ impl SerializableItem for Editor {
...
+                                        visible: Some(OpenVisible::None),
+                                        ..Default::default()
+                                    },
+                                    window,
+                                    cx,
+                                )
+                            })?;
+                            let editor = open_by_abs_path.await?.downcast::<Editor>().with_context(|| format!("Failed to downcast to Editor after opening abs path {abs_path:?}"))?;
+
+                            // Restore serialized contents if present (hot-exit for files in empty workspaces)
```
SomeoneToIgnore: There seems to be exactly the same block above, with `// This is a bit wasteful: we're loading the whole buffer from` comment.

Overall, we're in the place with quite complicated and somewhat dirty code, so, let's keep it clean if we can?
I think if we move the same text + that large comment above into its own function/method, we can make its name readable enough to remove this comment line here too?

crates/editor/src/items.rs
```diff
@@ -1197,19 +1197,44 @@ impl SerializableItem for Editor {
...
+                                    abs_path.clone(),
+                                    OpenOptions {
+                                        visible: Some(OpenVisible::None),
+                                        ..Default::default()
+                                    },
+                                    window,
+                                    cx,
+                                )
+                            })?;
+                            let editor = open_by_abs_path.await?.downcast::<Editor>().with_context(|| format!("Failed to downcast to Editor after opening abs path {abs_path:?}"))?;
```
SomeoneToIgnore:
```suggestion
                            let editor = open_by_abs_path.await?.downcast::<Editor>().with_context(|| format!("path {abs_path:?} cannot be opened as an Editor"))?;
```

Given that we need to do some changes still, why not land a drive-by style change.

crates/zed/src/zed/open_listener.rs
```diff
@@ -467,7 +467,9 @@ async fn open_workspaces(
                 .unwrap_or_default()
         }
     } else {
+        // For new workspaces opened via CLI, use default WorkspaceId (will be assigned by open_paths)
```
SomeoneToIgnore: Let's bend this the other way around, instead of creating fake IDs: those are totally ignored in the loop below,  so we can throw them away out of the only code path it comes of (so, 3 lines above, we do `.unwrap_or_default().into_iter()...`) and there, a meaningful comment explaining why do we ignore the DB values, is viable.

SomeoneToIgnore (commented):
Works like a charm now, amazing.

Thank you for finding a way to adjust the code, let's make it a bit more easy to approach for future changes and merge.

crates/zed/src/zed/open_listener.rs
SomeoneToIgnore: We still use some fake WorkspaceId::default() only to skip it with `for (_workspace_id, location, workspace_paths) in grouped_locations`.

Let's not do that.

SomeoneToIgnore (commented):
Almost there.

crates/zed/src/zed/open_listener.rs
shfc: Done. How's it look now?

SomeoneToIgnore (approved):
Great change, thank you a lot!

tdkl:
> > Does this new functionality mean I can have Zed save non-persisted files (e.g. newly created, and not existing edited) without even having a workspace = akin to default on VSCode?
> 
> If you mean unsaved files not in a workspace, the answer is yes. I just tested it. My prayers have been answered.

How can you get it to work? I currently have it set to Restore On Startup: Last Session, but when I create new files and close the Zed editor, it prompts to save the files or discard them all.

jankkm:
I noticed too that it usually works well but there is at least one situation where the unsaved files get lost:
If I close the current empty workspace but not the application (red button on macOS) the window closes without a warning. This might be technically similar to what happens on update like @tdkl described.

The same is true for non-empty workspaces but in that case I can reopen the workspace and the unsaved work comes back.

Issue #19333, opened 2024-10-17
Add filter options for global_search in json settings


bhbmaster:
### Check for existing issues

- [X] Completed

### Describe the feature

I often have to search a subset of my entire directory structure using global search using filters. My work flow is such that I open my work directory as my root directory for Zed and other code/text editors and work from there.
I have multiple directories in my work directory but why I search I only need to look thru `_notes` and `_cases`.

Example:
root directory is `~/work`
Which has the following dirs and each with thousands of files:
```
_notes
_cases
_bin
docx
HR
etc
```

I only want to be able to search thru _notes and _cases. I can do this with the UI global search filter option. However, if I close out of the global search those settings are lost.

Into include I put:
`_notes/*, _cases/*`

Side-Question: is the syntax that I have for the include filter correct to allow searching recursively thru _notes and _cases? There are dozens to hundreds of sub-dirs inside _notes and _cases.

The only editable options I found for the search / global search is this:

```
{
  "search": {
    "whole_word": false,
    "case_sensitive": false,
    "include_ignored": false,
    "regex": true
  }
}
```

Here is a screenshot of my settings mid search. The results ended up coming out:

<img width="1031" alt="Zed 2024-10-16 17 58 52" src="https://github.com/user-attachments/assets/954d90a7-0424-49d9-b394-2a604b78bae6">

Note: i checked and currently this appears to be the first request for this feature. There are mentions of updating file_scan_exclusions or .gitignore. However those have additional affects that I would not want to affect my search. For example I want to be able to still see the other directories. I just want global search to always be bound by my include and exclude parameters if I am working in my work directory.

### If applicable, add mockups / screenshots to help present your vision of the feature

Just add the "filter" json key into the "search" json key & make it understood by the editor.

```
{
  "search": {
    "whole_word": false,
    "case_sensitive": false,
    "include_ignored": false,
    "regex": true
    "filters": {
        "include": "_notes/*, _cases/*",
        "exclude": "" 
      }
  }
}
```

In my case I would put these in `~/work/.zed/settings.json`. However, if someone wants settings for any zed env then they should be able to put it in `~/.config/zed/settings.json`

notpeter:
> `_notes/*, _cases/*`
> Side-Question: is the syntax that I have for the include filter correct to allow searching recursively thru _notes and _cases? There are dozens to hundreds of sub-dirs inside _notes and _cases.

[...]

PR #46558, opened 2026-01-11
Add project search presets

Peter554:
https://github.com/user-attachments/assets/65eb51ce-0b6f-4a80-bcea-ef64c60ea623

Adds support for saved project search presets. Users can define presets in `.zed/settings.json` with include/exclude glob patterns:

```json
"project_search_presets": {
  "source_only": {
    "include": "src/**",
    "exclude": "**/*.test.ts"
  }
}
```

A new `SearchWithPreset` action opens a picker to select a preset, which then opens project search with the filters populated.

I would use this feature when working on large codebases, I'd find it really useful - it's the main thing keeping me in pycharm at the moment.

This PR intentionally keeps the scope minimal:
- Presets are project-only (defined in `.zed/settings.json`)
- Only include/exclude filters are supported
- No UI for creating/saving presets

Possible extensions:
- Save current filters as preset from UI
- Additional preset options (regex mode, case sensitivity, etc.)

---

From the contribution guidelines:

> Make sure the change is desired...

I haven't previously opened an issue to discuss this, so I understand if we don't want to pursue this - I don't mind the wasted effort, it was interesting either way. A related issue I found in the backlog https://github.com/zed-industries/zed/issues/19333.


implementation-plan.md
```diff
@@ -0,0 +1,80 @@
+# Project Search Presets - Implementation Plan
```
Peter554: As the commit message says:

> I used Claude Opus 4.5 to assist me with these changes. As part of my work I had Claude generate some artifacts. I'm including those in this commit for transparency. The artifacts are removed in the next commit.

crates/search/src/project_search.rs
```diff
@@ -189,6 +193,28 @@ pub fn init(cx: &mut App) {
             ProjectSearchView::new_search(workspace, action, window, cx);
             cx.notify();
         });
+        workspace.register_action(move |workspace, _: &SearchWithPreset, window, cx| {
+            if workspace.has_active_modal(window, cx) {
+                cx.propagate();
+                return;
+            }
+
+            // Collect presets from all project settings files across open worktrees.
```
Peter554, line 202: ❓ I don't fully understand worktrees. Typically I'm just using Zed with a single project open, so I assume that's just a single worktree. I'm not sure how one even opens a second worktree 🤔 Is it correct to consider all worktrees here?

crates/search/src/project_search.rs
```diff
@@ -2440,6 +2466,169 @@ fn register_workspace_action_for_present_search<A: Action>(
...
+            picker,
+            presets,
+            filtered_presets,
+            selected_index: 0,
+            workspace,
+        }
+    }
+}
+
+impl PickerDelegate for SearchPresetPickerDelegate {
```
Peter554, line 2531: ❓ This trait implementation feels a bit like boilerplate to me. I wonder if there is any way to DRY this up? Zed must have many other similar implementations of pickers?

Peter554:
Thanks for looking at this @SomeoneToIgnore 

> it seems more viable to implement search persistence?

This sounds like a useful addition. I'm not sure it fully solves my usecase, but it's something.

For context I'm working in a large (>10 million LOC) modular monorepo, so would typically have quite a few different search presets I'd be interested in saving and switching between. I understand that this is not the most typical usecase though, so I understand the decision.

PR #46560, opened 2026-01-11
linux: Fix duplicate IME input on Wayland and Sunshine remotes

hebo6:
## Summary
This PR fixes a issue where Chinese characters (and other IME-based inputs) would duplicate in the editor on Linux Wayland after remote desktop connected (e.g., via Sunshine) .

## Steps to Reproduce
requirement: linux with wayland, and Sunshine(this is remote desktop server) installed
1. opening zed
2. using other device connect to Sunshine
3. input Chinese characters such as "的"

### Expected Result
"的"

### Actual Result
"的的"

### bug recording
https://github.com/user-attachments/assets/d18961ec-48cb-4c06-a396-9fc604591769


reflectronic (approved):
Thank you

Issue #5005, opened 2023-02-15
Option to wrap open tabs in the available window instead of just the current scroll on overflow


rranjan14:
### Check for existing issues

- [X] Completed

### Describe the feature

VS Code currently has this feature to enable tab wrapping where the open file tabs wrap and are shown. In Zed, currently, there's no support for it. Tab wrapping helps to find the open file quicker than the scrolling option. Having this feature will be pretty amazing.

### If applicable, add mockups / screenshots to help present your vision of the feature

Zed:
<img width="1440" alt="ScreenShot 2023-02-15 at 20 09 30" src="https://user-images.githubusercontent.com/47914095/219058770-32cfad22-82ab-4c5c-9994-7e1acf7f9602.png">

VS Code:
<img width="1392" alt="ScreenShot 2023-02-15 at 20 06 38" src="https://user-images.githubusercontent.com/47914095/219058784-49b06ad8-e17a-434f-8c69-338ada900f3d.png">

williamdes:
That would be nice to have. Here is the setting on VScode for reference.

https://code.visualstudio.com/docs/getstarted/settings

```json

  // Controls whether tabs should be wrapped over multiple lines when exceeding available space or whether a scrollbar should appear instead. This value is ignored when `workbench.editor.showTabs` is not set to `multiple`.
  "workbench.editor.wrapTabs": false,
```

morgankrey:
Hi all, closing this issue under the umbrella of "won't do". Our design direction does not agree with the premise, and would create complexity for other UI elements as the number of lines expand. 
You're welcome to shift this to a discussion, but it's very unlikely we'd reconsider.

PR #46573, opened 2026-01-11
Add option to show pinned tabs in a separate row

yaroslavrick:
## Problem:

  When working with many `pinned tabs`, they consume significant horizontal space in the tab bar, leaving less room for unpinned (active working) tabs, especially on small screens. This creates a poor user experience as users must constantly scroll to find their working tabs, or the tabs become too narrow to read file names.

8_tabs  

##  Solution:

  Added a new opt-in setting `tab_bar.show_pinned_tabs_in_separate_row` that displays pinned and unpinned tabs in two separate rows:
  - Top row: Pinned tabs with navigation buttons and tab bar controls
  - Bottom row: Unpinned (working) tabs with full horizontal space
 
- when pinned and not pinned tabs we show 2 rows:  

2_rows

  This is a well-established UX pattern used in many popular IDEs including JetBrains products (IntelliJ IDEA, WebStorm, PyCharm, RubyMine), Visual Studio, and Eclipse.

  The two-row layout only appears when both pinned AND unpinned tabs exist. If only one type is present, a single row is displayed:
  
- when only not pinned tabs we show one row:

when only not pinned

- when only pinned tabs we show one row:

when only pinned tabs

  **Important** - Default Behavior Preserved:

  - The setting defaults to false - existing users will see no change whatsoever
  - Original tab bar logic is preserved in a dedicated `render_single_row_tab_bar` method, while the new two-row layout lives in `render_two_row_tab_bar`. Both methods share common components (`configure_tab_bar_start`, `configure_tab_bar_end`, `render_unpinned_tabs_container`, `render_tab_bar_drop_target`) ensuring DRY code and consistent behavior.
  - Purely opt-in - users must explicitly enable this feature via Settings UI or settings.json
  - No impact on default UX - Zed's default tab bar appearance and behavior remains identical

  This implementation ensures zero risk to the existing user experience while providing an optional enhancement for users who prefer separated tab rows.

  Configuration

  Via Settings UI:
  - Open Settings (Cmd+,) → Editor section → "Pinned Tabs Layout" toggle
  
  
settings_ui

  Via settings.json:
  
  ```json
  {
    "tab_bar": {
      "show_pinned_tabs_in_separate_row": true
    }
  }
  ```
  
manual_settings

## Video of implemented feature:

https://github.com/user-attachments/assets/b1e52074-a5a1-4c4e-ad59-e55e2213509d

## Tested:

  - Verify default behavior unchanged (single row with pinned + unpinned tabs inline)
  - Enable setting → verify two rows appear when both pinned and unpinned tabs exist
  - Enable setting → verify single row when only pinned tabs exist
  - Enable setting → verify single row when only unpinned tabs exist
  - Verify drag & drop works in both layouts
  - Verify navigation buttons and tab bar buttons render correctly
  - Verify setting toggle works in Settings UI
  - Verify setting works via settings.json

  Tested on: MacBook Pro M4, macOS 26.2
  
   ## Release Notes:

  - Added `tab_bar.show_pinned_tabs_in_separate_row` setting to display pinned and unpinned tabs in separate rows, giving unpinned tabs full horizontal space


williamdes:
Thank you so much, I will enable this as soon as tomorrow!
 Since you have experience with the tab bars, would you make them wrap tabs ?

See issue https://github.com/zed-industries/zed/issues/5005

yaroslavrick:
> This may have broken tab reordering when not enabled - please see #46864

I will take a look at this

yaroslavrick:
> This may have broken tab reordering when not enabled - please see #46864

Seems that I've found the roots of issue: here is this line: https://github.com/zed-industries/zed/pull/46573/commits/10b2072f540fd9ca992ee6b939668ccde8ca0488#diff-f3e6f80bf80e98b614b1fba64a700fc347b10286412e005f363e52279a550244R2768

`crates/workspace/src/pane.rs:2768`

cc @mattermill

crates/workspace/src/pane.rs
```diff
@@ -2765,7 +2765,7 @@ impl Pane {
             .on_drop(
                 cx.listener(move |this, dragged_tab: &DraggedTab, window, cx| {
                     this.drag_split_direction = None;
-                    this.handle_tab_drop(dragged_tab, ix, window, cx)
+                    this.handle_tab_drop(dragged_tab, this.items.len(), window, cx)
```
yaroslavrick, line 2768: Seems that I've found the roots of issue: here is this line: https://github.com/zed-industries/zed/commit/10b2072f540fd9ca992ee6b939668ccde8ca0488#diff-f3e6f80bf80e98b614b1fba64a700fc347b10286412e005f363e52279a550244R2768

crates/workspace/src/pane.rs:2768

cc @mattermill

yaroslavrick: PR of fix: https://github.com/zed-industries/zed/pull/46872

PR #46576, opened 2026-01-11
editor: Fix fold_recursive to fold descendants with single cursor

pedrobgodoy:
## Summary

Fixes `editor: fold recursive` command to properly fold nested code blocks when used with a single cursor (no selection).

## Problem

When using `editor: fold recursive` with the cursor on an outer block (e.g., `impl Foo {`) without any text selected:
- Only the outer block was folded
- Nested blocks (inner functions, if statements, etc.) were NOT folded
- After unfolding, all code was fully expanded

This was inconsistent with:
1. Multi-line selection behavior (which correctly folds all creases)
2. `editor: fold all` behavior (where unfolding reveals still-folded children)

## Solution

Modified the single-cursor code path in `fold_recursive()` to:
1. Find the outermost enclosing crease by searching backward (existing logic)
2. Iterate **forward** through the crease's range to collect all nested creases
3. Fold both the enclosing crease and all descendants

## Testing

- Added test `test_fold_recursive_single_cursor_folds_descendants`
- All 12 existing fold tests pass
- Clippy passes

Release Notes:

- Fixed `editor: fold recursive` to properly fold nested code blocks when cursor is on outer block without selection

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -19492,12 +19492,30 @@ impl Editor {
                 }
             }
 
+            // Find the outermost enclosing crease by searching backward
+            let mut enclosing_crease = None;
             for row in (0..=range.start.row).rev() {
                 if let Some(crease) = display_map.crease_for_buffer_row(MultiBufferRow(row)) {
                     if crease.range().end.row >= buffer_start_row {
-                        to_fold.push(crease);
+                        enclosing_crease = Some(crease);
                     } else {
                         break;
+                    }
+                }
+            }
+
+            // If we found an enclosing crease, fold it and all descendants
+            if let Some(crease) = enclosing_crease {
+                let crease_start = crease.range().start.row;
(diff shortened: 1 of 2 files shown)


smitbarmase (changes requested):
Thanks for looking into this! However, I think the current `fold_recursive` behavior is actually correct for vim's `zC` semantics, and this PR would regress it.

`zC` closes all fold levels containing the cursor (upward through the hierarchy). It does not fold children/siblings that don't contain the cursor line. The existing code is correct in that sense.

But, the behavior you're describing i.e. fold the enclosing region and all descendants is also one of use case and VSCode's Fold Recursively does exactly that.

I think the right approach is:
1. Keep `FoldRecursive` as-is (correct vim `zC` behavior)
2. Introduce a new action (e.g., `FoldDescendants`) that folds the enclosing region and all nested regions.
3. Bind this new action to `cmd-k cmd-[` / `ctrl-k ctrl-[`

PR #46582, opened 2026-01-11
terminal_view: Fix terminal opening in / when no project is open

### LINKED ISSUE #46574: Terminal opens to `/` when no projects are open
### Reproduction steps

1. Open Zed without a project
2. Open the terminal

Current: The terminal opens in `/`.

Expected: The `terminal.working_directory` setting defaults to `current_project_directory`, and the behaviour when no project is opened is not specified. However, falling back to the home directory seems like a much more reasonable choice than `/`. I don't think it's worth adding a setting for this.

Zed: v0.218.6+stable.114.e8b0ac131bc2bf80056f3769317ef57ebf32d0d1 (Zed) 
OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

maxmalkin:
Closes #46574

The `default_working_directory()` function in `crates/terminal_view/src/terminal_view.rs:1585` returned `None` when no project directory was available. The code comment (now removed) incorrectly claimed "None implies `~` on whichever machine we end up on". However, when `None` is passed to `alacritty_terminal`, it uses the process CWD, not the home directory. On macOS when Zed is launched from a .`app` bundle, the CWD is `/`.
    
Added a fallback at the end of` default_working_directory()` that explicitly returns the home directory when no project directory is found: `directory.or_else(dirs::home_dir)`

  This ensures:
  1. `CurrentProjectDirectory` with no project open → home directory
  2. `FirstProjectDirectory `with no project open → home directory
  3. `AlwaysHome `→ home directory (explicitly, not relying on shell behavior)
  4. Always `{ directory }` with invalid directory → home directory

Release Notes:
- Fixed terminal opening in `/` instead of home directory when no project is open.

### DIFF
--- a/crates/terminal_view/src/terminal_view.rs
+++ b/crates/terminal_view/src/terminal_view.rs
@@ -1593,12 +1593,10 @@ pub(crate) fn default_working_directory(workspace: &Workspace, cx: &App) -> Opti
             .or_else(|| first_project_directory(workspace, cx)),
         WorkingDirectory::FirstProjectDirectory => first_project_directory(workspace, cx),
         WorkingDirectory::AlwaysHome => None,
-        WorkingDirectory::Always { directory } => {
-            shellexpand::full(directory)
-                .ok()
-                .map(|dir| Path::new(&dir.to_string()).to_path_buf())
-                .filter(|dir| dir.is_dir())
-        }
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #46588, opened 2026-01-12
Add workflow to close stale draft PRs

yeskunall:
It doesn’t look like it’s possible to filter by “draft PRs” using [`actions/stale`](https://github.com/actions/stale/issues/1162). There’s an [`exempt-draft-pr`](https://github.com/actions/stale?tab=readme-ov-file#exempt-draft-pr) option, which is the exact opposite of what we want.

---

Other notes/questions:

- Update messaging? Am I being too terse?
- How soon should a draft PR be marked as `stale`?
- `cron` is set to a safe number, but needs to be updated

Release Notes:

- N/A

### DIFF
--- a/.github/workflows/community_close_stale_draft_prs.yml
+++ b/.github/workflows/community_close_stale_draft_prs.yml
@@ -1,0 +1,78 @@
+name: "Close Stale Draft PRs"
+on:
+  schedule:
+    # TODO: Update to regular schedule after testing
+    - cron: "0 0 31 12 *"
+  workflow_dispatch:
+    inputs:
+      debug-only:
+        description: "Run in dry-run mode (no changes made)"
+        type: boolean
+        default: false
+
+jobs:
+  stale-draft-prs:
+    if: github.repository_owner == 'zed-industries'
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      DAYS_BEFORE_STALE: 60
+      DAYS_BEFORE_CLOSE: 14
+      DRY_RUN: ${{ inputs.debug-only || false }}
+    steps:
+      - name: Process stale draft PRs
+        run: |
+          echo "Processing stale draft PRs..."
+          echo "Dry run mode: $DRY_RUN"
+
+          STALE_DATE=$(date -d "-${DAYS_BEFORE_STALE} days" -Iseconds)
+          CLOSE_DATE=$(date -d "-$((DAYS_BEFORE_STALE + DAYS_BEFORE_CLOSE)) days" -Iseconds)
+
+          STALE_MESSAGE="This draft PR has been open for over 60 days without activity. If you're still working on this, please leave a comment or push an update. Otherwise, it will be closed in 14 days.
+
+          If this PR is intentionally a long-running draft, add the \`never stale\` label to exempt it."
+
+          CLOSE_MESSAGE="This draft PR was closed due to inactivity. Feel free to reopen it if you'd like to continue working on it."
+
+          # Find all draft PRs
+          gh pr list \
+            --repo "$GITHUB_REPOSITORY" \
+            --draft \
+            --state open \
+            --json number,updatedAt,labels \
+            --limit 1000 | jq -c '.[]' | while read -r pr; do
+
+            PR_NUMBER=$(echo "$pr" | jq -r '.number')
+            UPDATED_AT=$(echo "$pr" | jq -r '.updatedAt')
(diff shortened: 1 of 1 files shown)


.github/workflows/community_close_stale_draft_prs.yml
```diff
@@ -0,0 +1,78 @@
...
+        type: boolean
+        default: false
+
+jobs:
+  stale-draft-prs:
+    if: github.repository_owner == 'zed-industries'
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      DAYS_BEFORE_STALE: 60
```
zelenenka, line 19: for drafts, i'd suggest 30. the codebase moves fast, and if a PR is still in half-baked state, chances are some things have changed from under it.

.github/workflows/community_close_stale_draft_prs.yml
```diff
@@ -0,0 +1,78 @@
...
+      DAYS_BEFORE_CLOSE: 14
```
zelenenka, line 20: maybe three weeks; some people go on long vacations and i imagine it would be irritating to see your PR being both flagged and closed before you have a chance to react one way or another.

.github/workflows/community_close_stale_draft_prs.yml
```diff
@@ -0,0 +1,78 @@
+name: "Close Stale Draft PRs"
+on:
+  schedule:
+    # TODO: Update to regular schedule after testing
+    - cron: "0 0 31 12 *"
+  workflow_dispatch:
+    inputs:
+      debug-only:
```
zelenenka, line 8: yay for having this option!

.github/workflows/community_close_stale_draft_prs.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - name: Process stale draft PRs
+        run: |
+          echo "Processing stale draft PRs..."
+          echo "Dry run mode: $DRY_RUN"
+
+          STALE_DATE=$(date -d "-${DAYS_BEFORE_STALE} days" -Iseconds)
+          CLOSE_DATE=$(date -d "-$((DAYS_BEFORE_STALE + DAYS_BEFORE_CLOSE)) days" -Iseconds)
+
+          STALE_MESSAGE="This draft PR has been open for over 60 days without activity. If you're still working on this, please leave a comment or push an update. Otherwise, it will be closed in 14 days.
```
zelenenka, line 31: i'd add a short bit about the codebase moving fast to make the motivation clear

.github/workflows/community_close_stale_draft_prs.yml
```diff
@@ -0,0 +1,78 @@
...
+
+          If this PR is intentionally a long-running draft, add the \`never stale\` label to exempt it."
```
zelenenka, line 33: i don't think people can label their own PRs, so this is only true for staff, and tbh staff should also close or update their draft PRs after 30/60/whatever days :)

.github/workflows/community_close_stale_draft_prs.yml
```diff
@@ -0,0 +1,78 @@
...
+
+          CLOSE_MESSAGE="This draft PR was closed due to inactivity. Feel free to reopen it if you'd like to continue working on it."
```
zelenenka, line 35: can people reopen their own PRs if they weren't the ones closing them? i don't think so — in github's data model PRs are issues, and with issues it's impossible to reopen your issue if the maintainers/bot/whoever closed it.

zelenenka (commented):
ugh, a bit frustrating that there's no option for draft PRs only. in a more perfect world we'd review all the incoming PRs and this wouldn't be a problem :)

looking at the code, i'm not sure we want to maintain a bash script in yaml without error handling, rate-limiting-checking, and all that.

have you checked out https://github.com/multitheftauto/stale-action ? it's made for exactly our use-case, seems battle-tested, and uses octokit.

otherwise, i've just checked, and we have 41 draft PRs (including this one), a lot of them from the staff members. this doesn't feel like many, so i wouldn't overinvest in handling this scenario, at least not while we have bigger fish to fry. what do you think?

.github/workflows/community_close_stale_draft_prs.yml
yeskunall: I think they can -- one cannot reopen _merged_ PRs, however. The user will also lose the ability to reopen the PR if the corresponding base branch is deleted, which usually happens from the contributor’s side.

If you have docs that say otherwise, please link them.

.github/workflows/community_close_stale_draft_prs.yml
yeskunall: Yup, correct about this, and I accidentally left it in here. Users need the triage role (or higher) to be able to add labels. Thanks for catching this!

PR #46644, opened 2026-01-12
git: Correct pending ops state handling when git stage fail

### LINKED ISSUE #46479: Git Panel Stage/Unstage toggle doesn't comprehend failure to stage
### Reproduction steps

1. Start Zed in some repo
2. Edit a file
3. create/touch `.git/index.lock`
4. Open Git Panel and click "Stage All"

The files will be unstaged, but the button will now say "Unstage All".

I expect that if nothing at all is staged, "Unstage All" is not an option.

Zed: v0.219.3+preview.116.3d83b3c9c66e9e5ef6f01c8849beede35264ed04 (Zed Preview) 
OS: macOS 26.2
Memory: 18 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

amtoaer:
Closes #46479

> This PR was initially submitted three months ago; I've rebased it onto the latest main branch and made minor adjustments.

Fix incorrect Git panel state after failed stage/unstage operations:

#### 1. Refine `PendingOps::staging()` logic
Previously, it incorrectly returned `true` for failed jobs with `JobStatus::Error` because it used `last.job_status != JobStatus::Finished` as the condition. Fixed by checking `last.running()` explicitly. This resolves the issue where Git panel checkboxes fall out of sync with the actual stage status. (Note: While full Git scans may still reset checkbox states, this fix significantly improves UI responsiveness by reducing perceived lag.)

#### 2. **Improve UI responsiveness on failure**
Added explicit `update_counts()` calls after `stage_all`/`stage_entries` operations fail. This ensures the "Stage All / Unstage All" buttons stay in sync with the stage status.

#### 3. **Avoid unreliable state negations**
Since `PendingOps::staging()` and `PendingOps::staged()` only inspect the `last` job (which could be an `Error`), the expression `!(ops.staging() || ops.staged())` does not reliably imply an "Unstaged" state. This PR refactors several logic paths to avoid this pattern and ensure more deterministic state checks.

---

Before:

https://github.com/user-attachments/assets/5509f68b-c9cc-49e0-959e-2fbafb58356a

After:

https://github.com/user-attachments/assets/86351825-b5c8-476e-82c5-09a6b822e0d4

Release Notes:

- Fixed incorrect Git panel state after failed stage/unstage operations

### DIFF
--- a/crates/project/src/git_store/pending_op.rs
+++ b/crates/project/src/git_store/pending_op.rs
@@ -121,10 +121,20 @@ impl PendingOps {
         false
     }
 
-    /// File is staged if the last job is not finished and has status Staged.
+    /// File is not staged if the last job is finished and has status other than Staged.
+    pub fn not_staged(&self) -> bool {
+        if let Some(last) = self.ops.last() {
+            if last.git_status != GitStatus::Staged && last.job_status == JobStatus::Finished {
+                return true;
+            }
+        }
+        false
+    }
+
+    /// File is staging if the last job is running and has status Staged.
     pub fn staging(&self) -> bool {
         if let Some(last) = self.ops.last() {
-            if last.git_status == GitStatus::Staged && last.job_status != JobStatus::Finished {
+            if last.git_status == GitStatus::Staged && last.running() {
                 return true;
             }
         }
--- a/crates/project/src/git_store/pending_op.rs
+++ b/crates/project/src/git_store/pending_op.rs
@@ -121,16 +121,6 @@ impl PendingOps {
         false
     }
 
-    /// File is not staged if the last job is finished and has status other than Staged.
-    pub fn not_staged(&self) -> bool {
-        if let Some(last) = self.ops.last() {
-            if last.git_status != GitStatus::Staged && last.job_status == JobStatus::Finished {
-                return true;
-            }
-        }
-        false
-    }
-
     /// File is staging if the last job is running and has status Staged.
     pub fn staging(&self) -> bool {
         if let Some(last) = self.ops.last() {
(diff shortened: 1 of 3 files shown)


crates/project/src/git_store/pending_op.rs
```diff
@@ -138,7 +148,7 @@ impl PendingOp {
     }
 
     pub fn finished(&self) -> bool {
-        matches!(self.job_status, JobStatus::Finished | JobStatus::Skipped)
```
kubkon, line 141: Why did you remove `Skipped` here? `Skipped` does not mean that a job was skipped due to an error condition, but because it was deduped by our job queueing system. I wonder, does that not break long-running git ops such as staging very large files?

kubkon (changes requested):

crates/project/src/git_store/pending_op.rs
amtoaer: My apologies. I noticed that `skipped` is returned when a task is canceled, which didn't seem to mean the task actually "finished." Since that method was only used in one place, I assumed the check was a mistake, so I updated the method and refactored other parts to call it for consistency. Sorry if I got that wrong. 🥲

kubkon: No probs at all! Thanks for looking into it!

Veykril (approved):
Thanks!

amtoaer:
Thanks for the review. I believe this fix has already been covered in: 
+ https://github.com/zed-industries/zed/pull/43693

Closing this now.

PR #46663, opened 2026-01-12
agent_ui: Make stopping a running terminal cancel thread generation

danilo-leal:
This makes stopping a terminal tool call in the agent panel also cancel the thread generation:

https://github.com/user-attachments/assets/702ba078-53a8-4816-80f3-b6215e764365

Release Notes:

- Agent: Made the act of stopping a terminal tool call in the agent panel also cancel the thread generation.

### DIFF
--- a/crates/agent_ui/src/acp/thread_view.rs
+++ b/crates/agent_ui/src/acp/thread_view.rs
@@ -3778,10 +3778,11 @@ impl AcpThreadView {
                         })
                         .on_click({
                             let terminal = terminal.clone();
-                            cx.listener(move |_this, _event, _window, cx| {
+                            cx.listener(move |this, _event, _window, cx| {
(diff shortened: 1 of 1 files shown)


MrPointer:
Oh no... That's not the behavior I want at all!  
Sometimes the model runs tools that hang forever, e.g. it tries to write concurrent code but accidentally creates a deadlock, so when it tests this code, via a tool call, it hangs forever until I manually stop it. It then captures the stop, analyzes how much time it took, and realizes that something's not right, and usually is smart enough to detect the deadlock and resolve it. With this change, that would be made impossible! (or require at least another prompt, which in some plans like Copilot, can be highly costly)

So adding to @sarmel's comment, this should be configurable, as surely we don't all agree that this should be the only behavior.

PR #46668, opened 2026-01-13
agents: Add copilot as an external agent via acp

jrobsonchase:
Adds Copilot as an external agent via its new `--acp` mode.

Mostly copied from the Gemini implementation. Seems to be working apart from authentication, which has to be done in advance.

Release Notes:

- Added Copilot as an External Agent


benbrandt:
Hey thanks for this. But we are moving all agent servers to extensions: https://zed.dev/docs/extensions/agent-servers
So that would be the way to go here.

PR #46708, opened 2026-01-13
Enable configurable dismissal of language server notifications that do not require user interaction

### LINKED ISSUE #38769: LS Server warning popup can not be silenced / does not dismis when stale

tomatitito:
Closes #38769

Release Notes:

- Dismiss server notifications automatically with `"global_lsp_settings": { "notifications": { "dismiss_timeout_ms": 5000 } }` settings defaults.


docs/src/configuring-zed.md
```diff
@@ -1,136 +0,0 @@
...
-
-## Example Configuration
-
-```json [settings]
-{
-  "theme": {
-    "mode": "system",
-    "light": "One Light",
-    "dark": "One Dark"
-  },
```
SomeoneToIgnore, line 113: We definitely do not want to remove this file.

Looks like the time had changed a bit and now the file is https://github.com/zed-industries/zed/blob/ccedb24757e35a2a718138883d7e2cc5125ddf6c/docs/src/reference/all-settings.md?plain=1#L1600 that we want to add a few more lines to.

crates/workspace/src/notifications.rs
```diff
@@ -99,6 +100,40 @@ impl Workspace {
...
+                    .as_ref()
+                    .is_some_and(|request| request.actions.is_empty());
+
+                let dismiss_timeout_ms = ProjectSettings::get_global(cx)
+                    .global_lsp_settings
+                    .notifications
+                    .dismiss_timeout_ms;
+
+                if is_prompt_without_actions {
+                    if let Some(dismiss_duration_ms) = dismiss_timeout_ms.filter(|&ms| ms > 0) {
```
SomeoneToIgnore, line 119: Nice addition.

Not sure we need to do anything now, but noticing that with this approach, we're not getting any settings change propagated to existing notifications.
Some transitions in settings (from "no timeouts" to "some quick timeout" and, presumably, back) may be confusing for users — let's see what the feedback comes.

SomeoneToIgnore (approved):
Looks great, I will push the documentation changes in the few hours, thank you for your persistence and work.

PR #46713, opened 2026-01-13
repl: highlight matching python environment

rgbkrk:
Highlight kernel matching python environment.

image


rgbkrk:
Closing this, coming back to it once I can get it to base it off of the active toolchain.

PR #46720, opened 2026-01-13
repl: Show active Python toolchain as recommended kernel

rgbkrk:
Mark the active toolchain as the recommended kernel for the repl.

image

Release Notes:

- List active toolchain/python environment as the recommend kernel for REPL usage

### DIFF
--- a/crates/project/src/project.rs
+++ b/crates/project/src/project.rs
@@ -3906,6 +3906,7 @@ impl Project {
             .read(cx)
             .active_toolchain(path, language_name, cx)
     }
+
     pub fn language_server_statuses<'a>(
         &'a self,
         cx: &'a App,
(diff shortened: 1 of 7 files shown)


crates/project/src/project.rs
```diff
@@ -3706,6 +3706,20 @@ impl Project {
             .read(cx)
             .active_toolchain(path, language_name, cx)
     }
+
+    /// Get the cached active toolchain synchronously (local projects only).
+    /// Returns None for remote projects or if no toolchain is cached.
+    pub fn cached_active_toolchain(
```
osiewicz: Why do we need this op to be synchronous? Could we make REPL initialization itself asynchronous to allow us to fetch the toolchain via other APIs we have in place?

rgbkrk:
Closing in favor of https://github.com/zed-industries/zed/pull/48763

PR #46721, opened 2026-01-13
agent_ui: Add support for Qwen agent in the agent panel and server store 

thanhtunguet:
agent_ui: Add support for Qwen agent in the agent panel and server store

This commit introduces the Qwen agent to the agent panel and server store, allowing it to be selected and managed alongside existing agents. The following changes were made:

Added Qwen to the AgentType and ExternalAgent enums.
Updated the agent panel to handle Qwen-specific actions and menu entries.
Integrated Qwen into the agent server store, including local and remote server implementations.
Added Qwen settings to the configuration structure.
Release Notes:

Added support for the Qwen agent in the agent panel and server store.


versecafe:
@thanhtunguet is there any reason for this to be a direct integration instead of making it an ACP extension like opencode, mistral vibe, etc. see https://zed.dev/docs/extensions/agent-servers

thanhtunguet:
> @thanhtunguet is there any reason for this to be a direct integration instead of making it an ACP extension like opencode, mistral vibe, etc. see https://zed.dev/docs/extensions/agent-servers

Besides Gemini CLI, qwen code provides us generous free quota to use and try the agentic development experience with Zed.
Unlike Google Gemini CLI, Qwen code support OpenAI Compatible API so that users can custom and use their own AI providers, especially local one.

I believe that would give users one more choice of free CLI agent and one more reason to start using and stay with Zed.

PR #46740, opened 2026-01-13
workspace: Implement focus-follows-mouse for panes

jrobsonchase:
Implements basic focus-follows-mouse behavior.

Right now, it's only applied in the `workspace` crate for `Pane`s, so anything that lives outside of that container (panels and such for the most part) won't have this behavior applied. The core logic is implemented as an extension trait, and should be trivial to apply to other elements as it makes sense.

https://github.com/user-attachments/assets/d338fa30-7f9c-439f-8b50-1720e3f509b1

Closes #8167 

Release Notes:

- Added "Focus Follows Mouse" for editor and terminal panes


jrobsonchase:
Alright, added a debounce timer and enabled ffm on all panels in the doc.

Regarding modals, I think it's already implicitly disabled. It seems nonfunctional when the file picker dialog, `zed: about`, and "restricted mode" modals are open at any rate.

ConradIrwin (approved):

tobiastom:
@jrobsonchase Really like this feature, thank you. 
Is it intentional that it does not seem to work with the Project Panel? The Outline Panel seems to work fine, but when I hover over the Project Panel, and try to use the cursor keys it does not seem to work. 

Edit: It seems to work when an item in the tree is selected before hovering the Panel.

gianpaj:
~Was this merged in `Zed 0.232.3`? I don't see this option or can enable it manually~

image

---

image

ConradIrwin:
@gianpaj I see it in your screenshot "Focus Follows Mouse" fourth from the top on the right :D

gianpaj:
@ConradIrwin , you're right :D I'm getting old, and when searching it couldn't see the setting easily :P

I think all the Settings search results should appear on the right-hand side, so there is no need to know which submenu they're in.

image

---

BTW the Focus-follows-mouse doesn't work with the Agent panel

injust:
After enabling ffm, it was easy to feel that 250ms debounce is a bit too slow. 100ms works much better for me; maybe 150ms if you move the mouse cursor slower across the screen.

PR #46743, opened 2026-01-13
editor: Fix clipboard line numbers when copying from multibuffer

kbidlack:
Release Notes:

- When copying and pasting from a multibuffer view (e.g. into the agent panel), the line numbers from the selection corresponded to the line numbers in the multibuffer view (incorrect), instead of the actual line numbers from the file itself.
- I also handled the case in which the selection spans multiple excerpts (different files in the multibuffer) by just returning None for that case, but maybe it should instead be split into two selections?

Left is before (bugged, should be lines 8:16, instead it is 38:46), right is after (fixed).
<div style="display:flex; gap:12px;">
  
  
</div>

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1643,7 +1643,15 @@ impl ClipboardSelection {
             project.absolute_path(&project_path, cx)
         });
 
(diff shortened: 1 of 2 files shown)


Veykril (approved):
Thanks!

PR #46744, opened 2026-01-13
project_panel: Add Menu key support to project panel context menu

### LINKED ISSUE #45783: "app" key does not work in file tree - should behave as right click
### Reproduction steps

1. Start Zed
2. Hit the app key in the editor area - this works fine and opens up the "right click"-menu.
3. Navigate to the file tree
4. Hit the app key again - this does not trigger the "right click"-menu.

Expected:

Expected the same menu to show up as when clicking the right mouse button in the filetree

Actual:

Nothing happens.

Zed 0.217.3
OS: Linux wayland cachyos
Memory: 32GB
Architecture: x86_64
Vim mode: on

CCXLV:
Closes #45783 

`Menu` key or `Shift + F10` shortcut wasn't supported in project panel for opening the context menu, this PR fixes it.

Release Notes:
- Added an action and default keyboard shortcuts to open the context menu in the project panel. You can now bind `project_panel::OpenContextMenu` to deploy the context menu at the currently selected entry in the panel.

### DIFF
--- a/crates/gpui/src/elements/anchored.rs
+++ b/crates/gpui/src/elements/anchored.rs
@@ -178,12 +178,8 @@ impl Element for Anchored {
             );
 
             match self.force_snap {
-                Some(AnchoredForceSnap::Left | AnchoredForceSnap::Right) => {
-                    desired = horizontal_switched;
-                }
-                Some(AnchoredForceSnap::Above | AnchoredForceSnap::Below) => {
-                    desired = vertical_switched;
-                }
+                Some(AnchoredForceSnap::Horizontal) => desired = horizontal_switched,
+                Some(AnchoredForceSnap::Vertical) => desired = vertical_switched,
                 None => {
                     if desired.left() < limits.left() || desired.right() > limits.right() {
                         if !(horizontal_switched.left() < limits.left()
@@ -266,14 +262,10 @@ impl IntoElement for Anchored {
 /// Enum for force snapping the anchored element.
 #[derive(Copy, Clone, PartialEq)]
 pub enum AnchoredForceSnap {
-    /// Force snap the anchored element above.
-    Above,
-    /// Force snap the anchored element below.
-    Below,
-    /// Force snap the anchored element left.
-    Left,
-    /// Force snap the anchored element right.
-    Right,
+    /// Force snap the anchored element horizontally.
+    Horizontal,
+    /// Force snap the anchored element vertically.
+    Vertical,
 }
 
 /// Which algorithm to use when fitting the anchored element to be inside the window.
--- a/crates/gpui/src/elements/anchored.rs
+++ b/crates/gpui/src/elements/anchored.rs
@@ -163,52 +163,41 @@ impl Element for Anchored {
         };
 
         if self.fit_mode == AnchoredFitMode::SwitchAnchor {
-            let mut anchor_corner = self.anchor_corner;
+            let anchor_corner = self.anchor_corner;
+
+            let horizontal_switched = Bounds::from_corner_and_size(
+                anchor_corner.other_side_corner_along(Axis::Horizontal),
+                origin,
+                size,
+            );
+
+            let vertical_switched = Bounds::from_corner_and_size(
+                anchor_corner.other_side_corner_along(Axis::Vertical),
+                origin,
+                size,
+            );
 
             match self.force_snap {
-                Some(force_snap) => {
-                    let horizontal_switched = Bounds::from_corner_and_size(
-                        anchor_corner.other_side_corner_along(Axis::Horizontal),
-                        origin,
-                        size,
-                    );
-
-                    let vertical_switched = Bounds::from_corner_and_size(
-                        anchor_corner.other_side_corner_along(Axis::Vertical),
-                        origin,
-                        size,
-                    );
-
-                    match force_snap {
-                        AnchoredForceSnap::Left | AnchoredForceSnap::Right => {
-                            desired = horizontal_switched
-                        }
-                        AnchoredForceSnap::Above | AnchoredForceSnap::Below => {
-                            desired = vertical_switched
-                        }
-                    }
+                Some(AnchoredForceSnap::Left | AnchoredForceSnap::Right) => {
+                    desired = horizontal_switched;
+                }
+                Some(AnchoredForceSnap::Above | AnchoredForceSnap::Below) => {
+                    desired = vertical_switched;
                 }
                 None => {
                     if desired.left() < limits.left() || desired.right() > limits.right() {
-                        let switched = Bounds::from_corner_and_size(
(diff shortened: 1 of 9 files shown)


CCXLV:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this, but this will need a slighty revised approach, as we cannot add it in the current form. You'll have to 

- Add another action to the project panel or menu crate `OpenContextMenu` (we have a similar one for the editor but I think that does not make sense here).
- Extract this into a method and call this on action with the `cx.listener` stuff as done for other actions.
- Add the appropriate default entries to the keymaps for Linux/MacOS/Windows as done for other actions.

I think then we should be good to go, as this makes the approach more future-proof and also allows other people to open the context menu with a key of their liking. What do you think?

CCXLV:
@MrSubidubi Sounds good, I’ll take a look!

CCXLV:
@MrSubidubi Check it out when you get a chance

crates/project_panel/src/project_panel.rs
```diff
@@ -341,6 +341,8 @@ actions!(
         SelectPrevDirectory,
         /// Opens a diff view to compare two marked files.
         CompareMarkedFiles,
+        /// Opens context menu at the mouse position
```
MrSubidubi:
```suggestion
        /// Opens the project panel context menu at the current mouse position.
```

crates/project_panel/src/project_panel.rs
```diff
@@ -1200,6 +1202,26 @@ impl ProjectPanel {
...
+        _: &OpenContextMenu,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let entry_id = if let Some(selection) = &self.state.selection {
+            selection.entry_id
+        } else if let Some(entry_id) = self.state.last_worktree_root_id {
+            entry_id
+        } else {
+            return; // No entry to show context menu for
```
MrSubidubi:
```suggestion
            return;
```

crates/project_panel/src/project_panel.rs
```diff
@@ -1200,6 +1202,26 @@ impl ProjectPanel {
         cx.notify();
     }
 
+    fn open_conetxt_menu(
```
MrSubidubi:
```suggestion
    fn open_context_menu(
```

crates/project_panel/src/project_panel.rs
```diff
@@ -1200,6 +1202,26 @@ impl ProjectPanel {
...
+        } else if let Some(entry_id) = self.state.last_worktree_root_id {
+            entry_id
+        } else {
+            return; // No entry to show context menu for
+        };
+
+        let mouse_pos = window.mouse_position();
+        let position = point(mouse_pos.x, mouse_pos.y);
+
+        self.deploy_context_menu(position, entry_id, window, cx);
```
MrSubidubi:
```suggestion
        self.deploy_context_menu(window.mouse_position(), entry_id, window, cx);
```

crates/project_panel/src/project_panel.rs
```diff
@@ -5780,6 +5802,7 @@ impl Render for ProjectPanel {
                 .on_action(cx.listener(Self::fold_directory))
                 .on_action(cx.listener(Self::remove_from_project))
                 .on_action(cx.listener(Self::compare_marked_files))
+                .on_action(cx.listener(Self::open_conetxt_menu))
```
MrSubidubi:
```suggestion
                .on_action(cx.listener(Self::open_context_menu))
```

crates/project_panel/src/project_panel.rs
```diff
@@ -6022,6 +6045,7 @@ impl Render for ProjectPanel {
                         .child(
                             div()
                                 .id("project-panel-blank-area")
+                                .track_focus(&self.focus_handle(cx))
```
MrSubidubi: Are you sure this is still needed?

MrSubidubi (changes requested):
Nice, this looks much better, thanks for the quick follow-up!

Left a few final comments and then we should be good to go I think.

CCXLV:
@MrSubidubi I think this should be it

MrSubidubi (commented):
Nice, much better, thanks for the quick follow-up!

One last thing: Right now, the context menu will pop up at the current mouse position, which obviously works, but makes this less ideal in my opionion. 

How would you feel abot having the context menu pop up somewhere close to the currently selected item? Happy to guide you there a bit if needed.

crates/project_panel/src/project_panel.rs
```diff
@@ -1219,6 +1230,73 @@ impl ProjectPanel {
         self.deploy_context_menu(window.mouse_position(), entry_id, window, cx);
```
CCXLV: The mouse position is unnecessary here. It could be improved

crates/gpui/src/elements/anchored.rs
```diff
@@ -179,6 +187,19 @@ impl Element for Anchored {
...
+
+            if let Some(force_snap) = self.force_snap {
+                let vertical_switched = Bounds::from_corner_and_size(
+                    anchor_corner.other_side_corner_along(Axis::Vertical),
+                    origin,
+                    size,
+                );
+
+                match force_snap {
+                    AnchoredForceSnap::Above => desired = vertical_switched,
```
CCXLV: I only used `Above` here but this could be adjusted to check for all `AnchoredForceSnap` values as well as add new values such as `Left`, `Right`

MrSubidubi (changes requested):
Thanks a lot for looking into this that deeply! 

Unfortunately though, I think we have to slightly change the approch here and take a small step back as to not overshoot things here. That is
- There should be no need for us to calculate the bounds of the project panel - you can find these on the scroll_handle already, so let's just use these to remove the dependency on the title bar. 
- Similarly, I do not think we should force the position of the context menu - although I do like the approach, it reaches pretty deeply into GPUI and we usually lay the element out like this for a reason. Layouting the element is pretty cheap, so I think we can instead compute and layout a context menu and then see whether it was layouted like we expected it to and adjust otherwise? I think that is much easier and also much less prone to errors in the future probably.

Could you look into these two things and see how things proceed there?

CCXLV:
Removed the local `anchor_corner` mutation in `prepaint`. It was never read or used, so the assignment had no effect. This simplifies the SwitchAnchor logic without changing behavior.

CCXLV:
Any updates? @MrSubidubi

MrSubidubi (approved):
I quickly hopped on and got this into a better shape, that might have slightly changed the placement but I believe it to be overall better and more maintanable going forward.

Thanks for getting the ball rolling on this one, if you want to do any follow-ups, please just go ahead! And apologies for letting this sit for so long!

PR #46745, opened 2026-01-13
Make auto-signature help respect popover delay

TheSalarKhan:
Closes #46191

Release Notes:

- Make the auto signature popover —  `"auto_signature_help": true` —  respect `hover_popover_delay`.

### DIFF
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -48,7 +48,7 @@ use project::{
 };
 use serde_json::{self, json};
 use settings::{
-    AllLanguageSettingsContent, EditorSettingsContent, IndentGuideBackgroundColoring,
+    AllLanguageSettingsContent, DelayMs, EditorSettingsContent, IndentGuideBackgroundColoring,
     IndentGuideColoring, InlayHintSettingsContent, ProjectSettingsContent, SearchSettingsContent,
     SettingsStore,
 };
@@ -13466,6 +13466,7 @@ async fn test_handle_input_for_show_signature_help_auto_signature_help_true(
         cx.update_global::<SettingsStore, _>(|settings, cx| {
             settings.update_user_settings(cx, |settings| {
                 settings.editor.auto_signature_help = Some(true);
+                settings.editor.hover_popover_delay = Some(DelayMs(300));
             });
         });
     });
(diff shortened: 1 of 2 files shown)


crates/editor/src/signature_help.rs
```diff
@@ -170,6 +170,8 @@ impl Editor {
             return;
         }
 
+        self.signature_help_state.task = None;
```
TheSalarKhan, line 176: Drop the task incase the previous one is waiting on timer or LSP and we're starting a new one.

Also important when the current iteration will exit early —  incase the cursor has exited the region inside the parenthesis of a function.

crates/editor/src/signature_help.rs
```diff
@@ -184,8 +186,18 @@ impl Editor {
...
+        let signature_help_delay_ms = EditorSettings::get_global(cx).hover_popover_delay.0;
+
         self.signature_help_state
             .set_task(cx.spawn_in(window, async move |editor, cx| {
+                // wait for debounce delay
+                if signature_help_delay_ms > 0 {
+                    cx.background_executor()
+                        .timer(Duration::from_millis(signature_help_delay_ms))
+                        .await;
+                }
```
TheSalarKhan: read the global setting, and before awaiting on the lsp task, await on the timer if non-zero.

TheSalarKhan:
Hi @maxdeviant, can you help me take this forward? Seems like I'll need a maintainer to trigger workflows.

Thanks!

SomeoneToIgnore (changes requested):
Thank you.

Can we please remove all the useless comments that repeat the code below verbatim?
And, please, can we add a test for this? `editor_tests.rs` should contain the signature help-related tests as an example.

TheSalarKhan:
Hi @SomeoneToIgnore

Thanks for guiding me on the next steps.

I've removed the useless comments, and I've added coverage for the new code to an existing test for signature help.

Could you please review this now.

crates/editor/src/editor_tests.rs
```diff
@@ -13466,6 +13466,7 @@ async fn test_handle_input_for_show_signature_help_auto_signature_help_true(
         cx.update_global::<SettingsStore, _>(|settings, cx| {
             settings.update_user_settings(cx, |settings| {
                 settings.editor.auto_signature_help = Some(true);
+                settings.editor.hover_popover_delay = Some(DelayMs(300));
```
TheSalarKhan, line 13469: Add `hover_popover_delay` for coverage.

SomeoneToIgnore (approved):
Thank you.

PR #46750, opened 2026-01-13
project_panel: Improve file/folder creation behavior in folded paths

### LINKED ISSUE #45550: File creation is inconsistent with result
### Reproduction steps

1. Have the following structure:
```
dir/
- sub/
- - file
``` 
2. Open Zed in root
3. Select `dir` with and call new file creation (`%` in vim mode)

### Current

https://github.com/user-attachments/assets/8b918846-c9b9-40ac-8bf4-b24e838f667f

(stuff in square brackets is in active state):
While creating file it looks like this
```
[dir]/sub/
- [___]
- file
```

And after file is created it unfolds all directories and it looks correct:
```
dir/
- sub/
- - file
- test
```

### Expected
It should show where new file will be created when it shows input field, not after it's creation, like this:
```
[dir]/
- [___]
- sub/
- - file
```
Like VSCode does:
<details><summary>Details</summary>
<p>

Image

</p>
</details> 

OS: Linux Wayland fedora 41
Memory: 27.2 GiB
Architecture: x86_64
GPU: AMD Radeon 780M (RADV PHOENIX) || radv || Mesa 25.0.7

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

andre-eriksson:
This fix, adds a new field to `EditState` to track and temporarily unfold the given directory when creating files or directories in collapsed paths.

Closes #45550  

Release Notes:

- Fixed the file and folder creation input appearing in the wrong location when creating inside a collapsed folder path..

Recording:

https://github.com/user-attachments/assets/15e48863-26b9-4dc9-888a-b31fb3208d98

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -95,6 +95,7 @@ struct State {
     visible_entries: Vec<VisibleEntriesForWorktree>,
     max_width_item_index: Option<usize>,
     edit_state: Option<EditState>,
+    temporarily_unfolded_pending_state: Option<TemporaryUnfoldedPendingState>,
     unfolded_dir_ids: HashSet<ProjectEntryId>,
     expanded_dir_ids: HashMap<WorktreeId, Vec<ProjectEntryId>>,
 }
@@ -114,6 +115,7 @@ impl State {
             visible_entries: Default::default(),
             max_width_item_index: None,
             edit_state: old.edit_state.clone(),
+            temporarily_unfolded_pending_state: None,
             unfolded_dir_ids: old.unfolded_dir_ids.clone(),
             expanded_dir_ids: old.expanded_dir_ids.clone(),
         }
@@ -157,6 +159,12 @@ struct UpdateVisibleEntriesTask {
     _visible_entries_task: Task<()>,
     focus_filename_editor: bool,
     autoscroll: bool,
+}
+
+#[derive(Debug)]
+struct TemporaryUnfoldedPendingState {
+    previously_focused_leaf_entry: SelectedEntry,
+    temporarily_unfolded_active_entry_id: ProjectEntryId,
 }
 
 impl Default for UpdateVisibleEntriesTask {
@@ -837,6 +845,7 @@ impl ProjectPanel {
                 state: State {
                     max_width_item_index: None,
                     edit_state: None,
+                    temporarily_unfolded_pending_state: None,
                     last_worktree_root_id: Default::default(),
                     visible_entries: Default::default(),
                     ancestors: Default::default(),
@@ -1938,22 +1947,31 @@ impl ProjectPanel {
     }
 
     fn discard_edit_state(&mut self, window: &mut Window, cx: &mut Context<Self>) {
-        let previously_focused = self
-            .state
-            .edit_state
-            .take()
-            .and_then(|edit_state| edit_state.previously_focused)
-            .map(|entry| (entry.worktree_id, entry.entry_id));
-        self.marked_entries.clear();
-        cx.notify();
-        window.focus(&self.focus_handle, cx);
-        self.update_visible_entries(
-            previously_focused,
-            false,
-            previously_focused.is_some(),
-            window,
-            cx,
-        );
+        if let Some(edit_state) = self.state.edit_state.take() {
+            self.state.temporarily_unfolded_pending_state = edit_state
+                .temporarily_unfolded
+                .and_then(|temporarily_unfolded_entry_id| {
+                    let previously_focused_leaf_entry = edit_state.previously_focused?;
+                    let folded_ancestors =
+                        self.state.ancestors.get(&temporarily_unfolded_entry_id)?;
+                    Some(TemporaryUnfoldedPendingState {
+                        previously_focused_leaf_entry,
+                        temporarily_unfolded_active_entry_id: folded_ancestors
+                            .active_ancestor()
+                            .unwrap_or(temporarily_unfolded_entry_id),
+                    })
+                });
+            let previously_focused = edit_state
+                .previously_focused
+                .map(|entry| (entry.worktree_id, entry.entry_id));
+            self.update_visible_entries(
+                previously_focused,
+                false,
+                previously_focused.is_some(),
+                window,
+                cx,
+            );
+        }
     }
 
     fn cancel(&mut self, _: &menu::Cancel, window: &mut Window, cx: &mut Context<Self>) {
@@ -1962,7 +1980,10 @@ impl ProjectPanel {
             self.hover_expand_task.take();
             return;
         }
+        self.marked_entries.clear();
+        cx.notify();
         self.discard_edit_state(window, cx);
+        window.focus(&self.focus_handle, cx);
     }
 
     fn open_entry(
@@ -3751,6 +3772,7 @@ impl ProjectPanel {
         let repo_snapshots = project.git_store().read(cx).repo_snapshots(cx);
 
         let old_ancestors = self.state.ancestors.clone();
+        let temporary_unfolded_pending_state = self.state.temporarily_unfolded_pending_state.take();
         let mut new_state = State::derive(&self.state);
         new_state.last_worktree_root_id = project
             .visible_worktrees(cx)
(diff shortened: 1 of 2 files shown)


crates/project_panel/src/project_panel.rs
```diff
@@ -6409,6 +6442,16 @@ fn cmp_files_first(a: &Entry, b: &Entry) -> cmp::Ordering {
...
+    if a.id == NEW_ENTRY_ID || b.id == NEW_ENTRY_ID {
+        if a.path.parent() == b.path.parent() {
+            match (a.id == NEW_ENTRY_ID, b.id == NEW_ENTRY_ID) {
+                (true, false) => return cmp::Ordering::Less,
+                (false, true) => return cmp::Ordering::Greater,
+                _ => {}
+            };
+        }
+    }
+
```
andre-eriksson: Without this, we ended up in an inconsistent situation where directories would show up under the parent directory and files at the very bottom.

Example without this:

**Folders**
```
dir/[subdir]
- - [folder]
- - /nested
- - - file.txt

dir/[subdir]
- - /nested
- - - file.txt
- - /test
```

**Files**
```
dir/[subdir]
- - /nested
- - - file.txt
- - [file]

dir/[subdir]
- - /nested
- - - file.txt
- - text.txt
```

This could easily be changed or removed if not desired.

crates/project_panel/src/project_panel.rs
```diff
@@ -752,6 +753,13 @@ impl ProjectPanel {
                                     task.detach_and_notify_err(window, cx);
                                 }
                                 None => {
+                                    if let Some(edit_state) = project_panel.state.edit_state.take()
```
smitbarmase: I think we can get rid of these manual cleanup checks by making `unfolded_dir_ids` private and accessing it through methods on `State`.

Instead of storing the temporary unfold in both `EditState.temporarily_unfolded` and inserting into `unfolded_dir_ids` (which requires manual cleanup in multiple places), we could make `is_unfolded()` compute the result:

I'm thinking something like this:
```rs
impl State {
    fn is_unfolded(&self, entry_id: &ProjectEntryId) -> bool {
            self.edit_state
                .as_ref()
                .and_then(|e| e.temporarily_unfolded)
                .is_some_and(|id| id == *entry_id) || self.unfolded_dir_ids.contains(entry_id)
    }
    fn unfold(&mut self, entry_id: ProjectEntryId) { ... }
    fn fold(&mut self, entry_id: &ProjectEntryId) { ... }
}
```

This way, the temporary unfold only lives in `EditState`, and when `edit_state` becomes `None`, it automatically stops appearing in `is_unfolded()` checks.

crates/project_panel/src/project_panel.rs
```diff
@@ -1941,6 +1975,16 @@ impl ProjectPanel {
...
         let new_entry_id = self.resolve_entry(entry_id);
+
+        let mut temporarily_unfolded = None;
+
+        if new_entry_id != entry_id {
+            if !self.state.unfolded_dir_ids.contains(&new_entry_id) {
+                self.state.unfolded_dir_ids.insert(new_entry_id);
+                temporarily_unfolded = Some(new_entry_id);
+            }
+        }
```
smitbarmase: Similarly, now we only check `new_entry_id != entry_id` and set `temporarily_unfolded`, no need to deal with `unfolded_dir_ids` at all here.

smitbarmase (changes requested):
Thanks for the PR! This is a great UX improvement. I've left a suggestion about the `temporarily_unfolded` cleanup pattern. Let me know what you think.

crates/project_panel/src/project_panel.rs
andre-eriksson: Looking a bit more on the code I think we can make it possible simpler, let me know if this would work!

In the `update_visible_entries`
https://github.com/andre-eriksson/zed/blob/f51798c4c966594e7c1460aa1b81565121ed08da/crates/project_panel/src/project_panel.rs#L3519-L3520

We can create a variable to check the temporary_unfolded, and check in the if statement:
```rust

let temporarily_unfolded = new_state.edit_state.as_ref().and_then(|e| e.temporarily_unfolded);

if !new_state.unfolded_dir_ids.contains(&entry.id) 
    && temporarily_unfolded != Some(entry.id) 
    && let Some(root_path) = worktree_snapshot.root_entry() { 
    // ... 
}
```

And then getting rid off every single cleanup function, from testing it seems to work just fine, so I'll commit it for now!

crates/project_panel/src/project_panel.rs
```diff
@@ -3556,6 +3570,7 @@ impl ProjectPanel {
                             if auto_collapse_dirs && entry.kind.is_dir() {
                                 auto_folded_ancestors.push(entry.id);
                                 if !new_state.unfolded_dir_ids.contains(&entry.id)
+                                    && temporarily_unfolded != Some(entry.id)
```
smitbarmase: This approach works well. My only concern is maintainability. If someone later checks just `unfolded_dir_ids.contains()` to determine if a directory is unfolded (as `is_unfoldable` already does), they might miss the `temporarily_unfolded` case.

In practice this isn't an issue today since `is_unfoldable` is only used for the context menu, which wouldn't be open during file creation. But for better encapsulation, having helper method on `State` would provide a single source of truth and prevent subtle bugs in the future.

crates/project_panel/src/project_panel.rs
```diff
@@ -1941,6 +1942,12 @@ impl ProjectPanel {
 
         let directory_id;
         let new_entry_id = self.resolve_entry(entry_id);
+        let mut temporarily_unfolded = None;
+
+        if new_entry_id != entry_id {
+            temporarily_unfolded = Some(new_entry_id);
```
smitbarmase: Can we only compute this if we're actually constructing `edit_state`? If we return before creating edit state, we don't need it, right? 

Maybe:

```rs
temporarily_unfolded: (new_entry_id != entry_id).then_some(new_entry_id),
```

crates/project_panel/src/project_panel.rs
```diff
@@ -6409,6 +6424,20 @@ fn cmp_files_first(a: &Entry, b: &Entry) -> cmp::Ordering {
 
 #[inline]
 fn cmp_with_mode(a: &Entry, b: &Entry, mode: &settings::ProjectPanelSortMode) -> cmp::Ordering {
```
smitbarmase, line 6952: I noticed that VSCode has more nuanced behavior where the new entry placement respects the current sort mode:

| Mode | Creating | Placement In Parent|
|------|----------|-----------|
| DirectoriesFirst | dir | top |
| DirectoriesFirst | file | top of files (after dirs) |
| Mixed | dir | top |
| Mixed | file | top |
| FilesFirst | dir | top of dirs (after files) |
| FilesFirst | file | top |

Currently, it always places the new entry at the absolute top in parent regardless of sort mode or entry type. Could we update this logic? This would require checking sort mode and whether the new entry is a file or directory.

smitbarmase (changes requested):
I have provided a few more suggestions: 

Also, could you add tests for:
1. `temporarily_unfolded`: Creating a file/directory inside a folded directory and verify the editor appears correctly and the directory re-folds after confirm/cancel.
2. New entry sorting: Creating both files and directories in each sort mode (`DirectoriesFirst`, `Mixed`, `FilesFirst`) to verify placement matches the expected behavior.

crates/project_panel/src/project_panel.rs
andre-eriksson: Yeah that would work and it is nicer!

crates/project_panel/src/project_panel.rs
andre-eriksson: The reason I went for this is due to `expanded_dir_ids` using `new_state` as mutable before and using it after, so, more or less avoiding unnecessary refactors, but I do agree with the maintainability concerns.

crates/project_panel/src/project_panel.rs
andre-eriksson: Would it be good to have that table as docs for the function?
Example:
```rust
/// Compare two entries when creating a new entry.
///
/// ## Calculation
/// | Mode             | Creating | Placement In Parent       |
/// |------------------|----------|---------------------------|
/// | DirectoriesFirst | dir      | top                       |
/// | DirectoriesFirst | file     | top of files (after dirs) |
/// | Mixed            | dir      | top                       |
/// | Mixed            | file     | top                       |
/// | FilesFirst       | dir      | top of dirs (after files) |
/// | FilesFirst       | file     | top                       |
#[inline]
fn cmp_new_entry(
    a: &Entry,
    b: &Entry,
    mode: &settings::ProjectPanelSortMode,
)
```

smitbarmase: Yeah, I believe that would be okay.

andre-eriksson:
An edge-case that I haven't been able to solve yet is the following scenario:
```
<root>
- dir/[subdir]/nested
```
```
<root>
- dir/[subdir]
- - - nested/
- - - [input]
```

Now cancel the input right click dir and start creating there:
```
<root>
- [dir]/subdir/nested
- - - - [input]
```
Finishing that input puts the file under dir/subdir which is incorrect, so both where the file is meant to go and where it is display is wrong. It is also quite annoying to debug since it requires multi-step interactions, wondering if you could potentially give some directions or advice here. 

Thanks for your feedback!

smitbarmase (changes requested):
Thanks! The code changes look good. I have a few suggestions before we can merge, sorry for dragging this out a bit longer:

1. I feel this should be two separate PRs. One for the folding fix with its tests, and one for the sort mode placement. Would you mind splitting these up? We can keep this PR focused on the folding file creation fix.

2. For the tests, could we reorganize them a bit? The test you added `test_ensure_folding_when_creating` is basically testing that files end up in the correct location when creating them in a folded directory after confirming, which is good to have since we don't already have these tests. But we should add a separate test specifically for the `temporarily_unfolded` behavior. That is, assert that the directory temporarily unfolds to show the editor in the correct location, and that after cancel, it re-folds back to its original state. Also, let's drop `test_ensure_folding_when_deleting` since I don't think we need this test, at least not for the changes made in this PR.

3. I tried to reproduce the edge case you mentioned but couldn't. I think adding the above test will make it visible if there's a bug. But if you can share a video with me, maybe I can suggest what's going on.

andre-eriksson:
Sure here is the edge case example video: 

https://github.com/user-attachments/assets/1066be71-4b90-4fc1-8740-2eee3027660a

Ill work on those fixes and split the PR, thanks!

smitbarmase (approved):
I added those test cases and handle the context menu case that was breaking. This is a great improvement. Thank you so much!

Issue #44814, opened 2025-12-14
Rendering Freeze/Visual Lockup on Wayland with Tiling WM (Niri)


RatDotRaw:
### Reproduction steps

The bug is on stock settings, a freshly installed version.

1. Start Niri, a Wayland **tiling** window manager.
2. Start Zed

That's it.
If anything else is needed, even testing help, I'll see what I can do.

### Current Behavior: 
On launch, Zed's visual content is completely frozen and does not update (e.g., blinking text cursor, new text, menu changes). But the underlying application remains responsive, clicking and typing actions register correctly in the background. The rendering only resumes if the window is forced into a free-floating state and then back into a tiled state, or resized in any way.

### Expected Behavior: 
Rendering the window contents immediately.

### Zed version and system specs

Zed: Zed 0.216.1+stable (flatpak & pacman package)
OS: Arch Linux 
Window manager:  Niri 25.11-1
RAM: 32GB
Architecture: x86_64

### Attach Zed log file

<details><summary>Zed.log</summary>

I have picked out the most important logs:
 There are thousands of lines of `WARN  [blade_graphics::hal::surface] Acquire failed because the surface is out of date`

```log
2025-12-14T14:26:38+01:00 INFO  [zed] ========== starting zed version 0.216.1+stable.95.81870d5cf77d61d40ec526a37141ebe3a74318ca, sha 81870d5 ==========
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::handshake::client] perform;
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::handshake::client] authenticate;
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::handshake::common] read_command;
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::handshake::client] send_secondary_commands;
2025-12-14T14:26:38+01:00 INFO  [zbus::proxy] new;
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::socket_reader] socket reader;
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::socket_reader] read_socket;
2025-12-14T14:26:38+01:00 INFO  [crashes] spawning crash handler process
2025-12-14T14:26:38+01:00 INFO  [zbus::connection::socket_reader] read_socket;
2025-12-14T14:26:38+01:00 INFO  [crashes] connected to crash handler process after 0ns
2025-12-14T14:26:38+01:00 INFO  [crashes] crash handler registered
[...]

Issue #45836, opened 2025-12-29
Zed freezes at startup


cacilhas:
### Reproduction steps

Just start Zed, that’s all it takes. It freezes on Arch Linux (using niri/Wayland) during the startup.

### Zed version and system specs

I cannot open the command palette, Zed is freezing at startup.

```shell
❯ zeditor --version
Zed 0.217.3 – /usr/lib/zed/zed-editor
❯ uname -a
Linux dream 6.12.63-1-lts #1 SMP PREEMPT_DYNAMIC Thu, 18 Dec 2025 14:48:15 +0000 x86_64 GNU/Linux
```

Whatelse can I do to supply more info?

### Attach Zed log file

How  am I supposed to run a command in the command palette if Zed simply freezes?

<details><summary>Zed.log</summary>

```log

```

</details>

miguelraz:
Thanks for reporting!

Please let us know what happens after running `vkcube`.

camspiers:
@MostlyKIGuess It's not a Niri bug (I mean you could class it as one, but ultimately zed+blade need to handle the surface being outdated, not just ignore it), and it is unlikely to be fixed there as far as I can tell. I've provided a writeup in this comment:

https://github.com/zed-industries/zed/issues/40481#issuecomment-3688093740

You can see example fixes of the same issue in two separate systems:

https://github.com/linebender/xilem/pull/1550

and

https://github.com/iced-rs/iced/pull/3067

The blade code that doesn't handle the outdated surface correctly is:

https://github.com/kvark/blade/blob/8cd905be95d9bcf476f69790fcfdbeb615c5154c/blade-graphics/src/vulkan/surface.rs#L44-L51

miguelraz:
Thank you so much for the sleuthing @camspiers ! Really appreciate it!

I've redirected the duplicates to the oldest issue.

Issue #44252, opened 2025-12-05
Zed is not openning in CachyOS with LXDE


igotfr:
### Reproduction steps

1. Start Zed
2. Zed doesn't open

### Current vs. Expected behavior

I have already installed the drivers

```
$ lsmod | grep nvidia

nvidia_drm 81920 3

nvidia_uvm 1703936 0

nvidia_modeset 1519616 5 nvidia_drm

video 81920 2 asus_wmi,nvidia_modeset

nvidia 40759296 167 nvidia_uvm,nvidia_modeset
```
```
$ grep "(EE)" /var/log/Xorg.0.log

(WW) warning, (EE) error, (NI) not implemented, (??) unknown.

[ 33.528] (EE) NVIDIA(GPU-1): Failed to initialize the NVIDIA graphics device!

[ 33.528] (EE) NVIDIA(G0): Failing initialization of X screen

[ 33.528] (EE) NVIDIA(GPU-1): Failed to initialize the NVIDIA graphics device!
```

### Zed version and system specs

Zed: 0.214.7 16179728a298ba215be25b09660277f8872c471f
OS: CachyOS
Memory: 8GiB
Architecture: x86_64
GPU: GTX 650Ti

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
$ cat .local/share/zed/logs/Zed.log
2025-12-04T09:53:09-03:00 INFO [gpui::platform::linux::x11::client] x11: compositor present: true, gtk_frame_extents_supported: false

2025-12-04T09:53:09-03:00 WARN [blade_graphics::hal::init] Rejected for device extension "VK_KHR_dynamic_rendering" not supported. Please update the driver!
```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

miguelraz:
What do you get when you run `vkcube` on your machine?

igotfr:
@miguelraz the cube and:
```
$ vkcube
Selected WSI platform: xcb
Selected GPU 0: NVIDIA GeForce GTX 650 Ti, type: DiscreteGpu
```

miguelraz:
Are you using the most up to date GPU drivers?

igotfr:
@miguelraz I installed with `yay -S nvidia-470xx-dkms nvidia-470xx-utils lib32-nvidia-470xx-utils nvidia-settings`, I don't know how to check drivers versions

[...]

Issue #40361, opened 2025-10-16
Endeavour OS: Settings window on new zed release are not clickable


donl07:
### Summary

This issue can be quit easy to reproduce on my EndeavourOS. I found that Zed Settings, the child window of Zed is opened with floating on my Hyprland Window Manager, instead of tiling, which led to the unclickable of Zed Settings Window and I only encounter this issue on the latest release of Zed. 

When I change the hyprland scale from 1.0 to 2.0(actually 1.25), it is clickable again. But that's ugly.

### Zed Version and System Specs

zed v0.208.4
os: linux wayland EndeavourOS
memory: 7.4GiB
Architecture: x86_64
GPU: intel i5-8250U  Intel UHD Graphics 620

katie-z-geer:
Yikes, thanks for letting us know. We'll dig into this and see if we can repro this week.

probably-neb:
Hey @donl07, I'm a bit confused that changing the scaling fixed this for you, when you say "clickable" do you mean that you can click on the window with your mouse and use it? Or are you referring to the window being floating instead of tiling so you can't navigate to it with the keyboard?

donl07:
Thanks for you@probably-neb reply. What I mean is that the window becomes interactive after adjusting the scaling, and it remains in a floating state. That's strange.

AlvaroParker:
So when you open the settings window for the first time and try to click on any button it is non-responsive? Which Hyprland version are you on?

donl07:
Certainly non-response. I use v49.0

probably-neb:
Hey @donl07, would you mind providing a screen recording of what you're seeing? I'm still a bit confused on what sequence of events is going wrong here

jilen:
For me, the window (both main window and settings window), won't respond to click event unless resize window first.

probably-neb:
Can you paste the output of `/path/to/zed --system-specs` @jilen? And are you also on hyprland?

jilen:
I'm using [niri](https://github.com/YaLTeR/niri) 


Zed System Specs (from CLI):
Zed: v0.213.4 (Zed)
OS: Linux Wayland arch unknown
Memory: 31.3 GiB
Architecture: x86_64
GPU: <details><summary>`vulkaninfo --summary` output</summary>
```
==========
VULKANINFO
==========

Vulkan Instance Version: 1.4.328


Instance Extensions: count = 24
-------------------------------
VK_EXT_acquire_drm_display             : extension revision 1
[...]

PR #46758, opened 2026-01-14
gpui: Remove blade, reimplement linux renderer with wgpu

### LINKED ISSUE #2335: Zed Editor is broken in Niri
This may not be a niri bug specifically, but the zed editor is broken in niri.  The reason is that when the application starts up, something invalidates the vulkan surface, and you have to resize the window to get it to work again.

Maybe there is something with the way that niri starts up windows that it's skipping a resize check or something.

### System Information

* niri version: niri 25.08 (01be0e6)

* Distro: Arch

* GPU: Nvidia 3090

* CPU: AMD Ryzen 9 3900XT

zortax:
The blade graphics library is a mess and causes several issues for both Zed users as well as other 3rd party apps using GPUI. This PR removes blade and implements the linux platform using `wgpu` which is the de-facto standard in the rust UI and graphics ecosystem. This will not just fix [issues that Zed users have today](https://github.com/YaLTeR/niri/issues/2335), but also profit from wgpu improvements in the futures, from other projects contributing (such as the bevy game engine, Iced, or pretty much every other relevant project).

This will close several related issues on the zed repo as well. See https://github.com/zed-industries/zed/issues?q=frozen%20nvidia%20linux (probably not all of them, have only tested the freeze on nvidia and Smithay-based wayland compositors).

Some related issues:
https://github.com/zed-industries/zed/issues/44814
https://github.com/zed-industries/zed/issues/40481
https://github.com/YaLTeR/niri/issues/2335
https://github.com/zortax/zlaunch/issues/15

Would appreciate feedback if this is something the zed maintainers would be interested in.

Release Notes:

- N/A


crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1277 @@
...
+
+        let surface_caps = surface.get_capabilities(&context.adapter);
+        // Prefer non-sRGB format to avoid sRGB blending issues
+        // The shader outputs linear values which will be displayed directly
+        let surface_format = surface_caps
+            .formats
+            .iter()
+            .find(|f| !f.is_srgb())
+            .copied()
+            .unwrap_or(surface_caps.formats[0]);
```
reflectronic: Not convinced this is right--I think this should error out instead of picking a format we don't want

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1277 @@
...
+                .map_err(|e| anyhow::anyhow!("Failed to create surface: {e}"))?
+        };
+
+        let surface_caps = surface.get_capabilities(&context.adapter);
+        // Prefer non-sRGB format to avoid sRGB blending issues
+        // The shader outputs linear values which will be displayed directly
+        let surface_format = surface_caps
+            .formats
+            .iter()
+            .find(|f| !f.is_srgb())
```
andre-eriksson: On my AMD 6800 XT this surface format selection can pick a 16-bit normalized format:
```
surface formats: [Bgra8UnormSrgb, Rgba8UnormSrgb, Rgba16Unorm, Rgb10a2Unorm, Bgra8Unorm, Rgba8Unorm, Rgba16Float]
chosen surface_format: Rgba16Unorm
```
Which causes:
```
wgpu error: Validation Error

Caused by:
  In Device::create_render_pipeline, label = 'quads'
    Features Features { features_wgpu: FeaturesWGPU(TEXTURE_FORMAT_16BIT_NORM), features_webgpu: FeaturesWebGPU(0x0) } are required but not enabled on the device
```

`Rgba16Unorm` requires `Features::TEXTURE_FORMAT_16BIT_NORM`, but the device is requested without enabling that feature: https://docs.rs/wgpu/latest/wgpu/struct.Features.html#associatedconstant.TEXTURE_FORMAT_16BIT_NORM

Maybe we could prefer `Bgra8Unorm` / `Rgba8Unorm` or conditionally allow `Rgba16Unorm` based on the enabled device features?

zortax: I tried enabling the feature and conditionally allowing `Rgba16Unorm` if supported. Seems to work on my end, but I only have a nvidia system at hand right now, so I think it currently selects Bgra8Unorm anyway.

reflectronic:
Thank you for the pull request. Yes, we are interested in this approach. From a brief look at the patch, things look good at a high level. I will make time for a deeper review soon.

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
zortax: On a second thought, maybe it would be better to just always use `Bgra8Unorm` / `Rgba8Unorm`...

crates/gpui/src/platform/mac.rs
```diff
@@ -10,18 +10,12 @@ mod pasteboard;
...
 mod metal_atlas;
-#[cfg(not(feature = "macos-blade"))]
 pub mod metal_renderer;
 
 use core_video::image_buffer::CVImageBuffer;
-#[cfg(not(feature = "macos-blade"))]
 use metal_renderer as renderer;
 
-#[cfg(feature = "macos-blade")]
-use crate::platform::blade as renderer;
```
maddythewisp, line 23: Since WGPU is a cross-platform API, I think it would be nice to be able to use it on Mac and Windows builds via a feature flag, similar to how `macos-blade` previously enabled the Blade renderer on Mac.

zortax: Agreed, but I can't test that, don't have the necessary hardware

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
andre-eriksson: You can do something like, to avoid `TextureFormat::Rgba16Unorm` it if it isn't enabled:
```rust
let can_16bit_norm = context
            .adapter
            .features()
            .contains(wgpu::Features::TEXTURE_FORMAT_16BIT_NORM);

let surface_caps = surface.get_capabilities(&context.adapter);

let surface_format = surface_caps
            .formats
            .iter()
            .find(|f| !f.is_srgb() && (!can_16bit_norm || **f != wgpu::TextureFormat::Rgba16Unorm))
            .copied()
            .unwrap_or(surface_caps.formats[0]);
```

Output (no validation errors):
```
surface formats: [Bgra8UnormSrgb, Rgba8UnormSrgb, Rgba16Unorm, Rgb10a2Unorm, Bgra8Unorm, Rgba8Unorm, Rgba16Float]
chosen surface_format: Rgb10a2Unorm
```

---

> On a second thought, maybe it would be better to just always use  Bgra8Unorm / Rgba8Unorm

Perhaps, but maybe we should prefer them rather than only using them. I am no expert in wgpu/webgpu or shaders so my input is limited here.

zortax: I pretty much tried that, but apparently it still panics on AMD: https://github.com/zortax/zlaunch/issues/15#issuecomment-3750880015
I can't really debug that since it doesn't happen on my system.. :/

andre-eriksson: I tested the commit [81794ad](https://github.com/zed-industries/zed/pull/46758/commits/81794adeed8fb8938e852991963699c519b0acdf) and it resulted in the same errors as that comment however, adding the `can_16bit_norm` check does make it work for my AMD system on Wayland, since the linked commit [81794ad](https://github.com/zed-industries/zed/pull/46758/commits/81794adeed8fb8938e852991963699c519b0acdf) didn't change what `surface_format` was being selected.

reflectronic: I agree that we should avoid the high-precision formats. We are not interested in WCG right now, so these formats will just consume more memory for no benefit.

zortax: I think you logic is wrong way around, isn't it?
```rs
(!can_16bit_norm || **f != wgpu::TextureFormat::Rgba16Unorm)
```
shouldn't this be
```rs
(can_16bit_norm || **f != wgpu::TextureFormat::Rgba16Unorm)
```
(this is exactly what I pushed in [81794ad](https://github.com/zed-industries/zed/pull/46758/commits/81794adeed8fb8938e852991963699c519b0acdf))

I think with your version it will filter out the 16bit format even though the adapter reports the feature. I thought the point of this is to allow the 16bit format if that feature is available?

andre-eriksson: I think you are right, good catch!

Also the features check is wrong therefore my tests were wrong, correction below:
```rust
let can_16bit_norm = context
            .device // Using device rather than adapter
            .features()
            .contains(wgpu::Features::TEXTURE_FORMAT_16BIT_NORM);
            
let surface_format = surface_caps
            .formats
            .iter()
            .find(|f| !f.is_srgb() && (can_16bit_norm || **f != wgpu::TextureFormat::Rgba16Unorm))
            .copied()
            .unwrap_or(surface_caps.formats[0]);
```

zortax: Ah `device` makes sense, thank you. But I guess we'll just prefer 8bit anyway now.

crates/gpui/src/platform/wgpu/wgpu_context.rs
```diff
@@ -0,0 +1,168 @@
...
+        if let Some(device_id) = device_id_filter {
+            for adapter in &adapters {
+                let info = adapter.get_info();
+                if info.device == device_id {
+                    log::info!(
+                        "Found GPU matching ZED_DEVICE_ID={:#06x}: {}",
+                        device_id,
+                        info.name
+                    );
+                    return instance
```
camspiers: I'm not super familiar with wgpu, but I would have thought that (assuming an adapter for the device exists) this wouldn't guarantee the return of an adapter for the device id specified because the device specified might actually be a different power type to what the request options specified, or the user has more than one discrete/high performance GPU installed.

For example the user specifies a device id in their env that is for an integrated gpu while having a discrete gpu installed (meaning that when we subsequently call `.request_adapter` with `HighPerformance` it's going to select an adapter for the discrete GPU).

andre-eriksson: Good point, it could return `Ok(adapter.clone())` instead of `.request_adapter(...)`.

camspiers: I managed to get a repo of what I mean. In the following `0x164e` is my integrated GPU on my CPU (AMD 7950x3d), it finds it but ends up selecting the 4080 (discrete GPU).

```bash
~/d/zed (wgpu)> ZED_DEVICE_ID=0x164e cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.58s
     Running `target/debug/zed`
2026-01-15T17:35:26+13:00 INFO  [zed] ========== starting zed version 0.220.0+dev.324945dea43acd3fda7e8ba1f0a7e1f091f658a0, sha 324945d ==========
2026-01-15T17:35:26+13:00 INFO  [gpui::platform::wgpu::wgpu_context] Found GPU matching ZED_DEVICE_ID=0x164e: AMD Radeon Graphics (RADV RAPHAEL_MENDOCINO)
2026-01-15T17:35:26+13:00 INFO  [gpui::platform::wgpu::wgpu_context] Selected GPU adapter: "NVIDIA GeForce RTX 4080" (Vulkan)
```

camspiers:
```rust
    async fn select_adapter(
        instance: &wgpu::Instance,
        device_id_filter: Option<u32>,
    ) -> anyhow::Result<wgpu::Adapter> {
        if let Some(device_id) = device_id_filter {
            let adapters: Vec<_> = instance.enumerate_adapters(wgpu::Backends::all()).await;

            if adapters.is_empty() {
                anyhow::bail!("No GPU adapters found");
            }

            let mut non_matching_adapter_infos: Vec<wgpu::AdapterInfo> = Vec::new();

            for adapter in adapters.into_iter() {
                let info = adapter.get_info();
                if info.device == device_id {
                    log::info!(
                        "Found GPU matching ZED_DEVICE_ID={:#06x}: {}",
                        device_id,
                        info.name
                    );
                    return Ok(adapter);
                } else {
                    non_matching_adapter_infos.push(info);
                }
            }

            log::warn!(
                "No GPU found matching ZED_DEVICE_ID={:#06x}. Available devices:",
                device_id
            );

            for info in &non_matching_adapter_infos {
                log::warn!(
                    "  - {} (device_id={:#06x}, backend={})",
                    info.name,
                    info.device,
                    info.backend
                );
            }
        }

        instance
            .request_adapter(&wgpu::RequestAdapterOptions {
                power_preference: wgpu::PowerPreference::HighPerformance,
                compatible_surface: None,
                force_fallback_adapter: false,
            })
            .await
            .map_err(|e| anyhow::anyhow!("Failed to request GPU adapter: {e}"))
    }
```

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1282 @@
...
+
+        let target = wgpu::SurfaceTargetUnsafe::RawHandle {
+            raw_display_handle: display_handle.as_raw(),
+            raw_window_handle: window_handle.as_raw(),
+        };
+
+        // Safety: The caller guarantees that the window handle is valid for the
+        // lifetime of this renderer. In practice, the RawWindow struct is created
+        // from the native window handles and the surface is dropped before the window.
+        let surface = unsafe {
```
camspiers, line 146: Is it worth using `SurfaceTargetUnsafe::from_window`?

```rust
        let surface = unsafe {
            context
                .instance
                .create_surface_unsafe(
                    wgpu::SurfaceTargetUnsafe::from_window(window)
                        .map_err(|e| anyhow::anyhow!("Failed to create surface target: {e}"))?,
                )
                .map_err(|e| anyhow::anyhow!("Failed to create surface: {e}"))?
        };
```

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1282 @@
...
+    }
+
+    pub fn gpu_specs(&self) -> GpuSpecs {
+        GpuSpecs {
+            is_software_emulated: false,
+            device_name: "wgpu".to_string(),
+            driver_name: "wgpu".to_string(),
+            driver_info: "wgpu renderer".to_string(),
+        }
+    }
```
probably-neb, line 823: Low priority issue, but it would be great if we could wire `wgpu::Adapter` through to this point and have accurate info here. We use this as metadata in crash dumps, so important for fixing any future issues :)

crates/gpui/src/platform/wgpu/wgpu_context.rs
zortax: yeah that didn't make a lot of sense, don't know what I thought when writing this here

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
zortax: fair enough, skipped that in the beginning to get stuff working, kinda forgot about it

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
zortax: To be honest I think this is "unnecessary complexity", since, as far as I understand, in the current implementation we would still wrap the underlying error. The only difference with this is the separate wrapper message, but the underlying error should make it very clear which error path was hit anyway, but please correct me if I'm wrong.

camspiers: We wouldn't gain much by using their function over manually doing it ourselves, and it's a fair argument that the error messages in the more manual case are less layered. It's probably not important in practice to see directly in the error that we failed to get a window or display handle while creating a surface target, because you are going to see that context by virtue of where the error occurs (in the creation of the renderer).

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
probably-neb: No worries!

zortax:
I believe this PR closes the following issues:
https://github.com/zed-industries/zed/issues/39097
https://github.com/zed-industries/zed/issues/44814
https://github.com/zed-industries/zed/issues/45836
https://github.com/zed-industries/zed/issues/40481
https://github.com/zed-industries/zed/issues/38497

There are some other issues this might close, but haven't tested/reproduced:
https://github.com/zed-industries/zed/issues/40361
https://github.com/zed-industries/zed/issues/44252

crates/gpui/src/platform/mac.rs
pampas9000: > Since WGPU is a cross-platform API, I think it would be nice to be able to use it on Mac and Windows builds via a feature flag, similar to how `macos-blade` previously enabled the Blade renderer on Mac.

With Zed moving to wgpu (WebGPU), does this mean a web-based client is now technically possible, similar to Flutter Web?

maddythewisp: @bamboo512 There is significant work beyond the renderer that would need to happen to run Zed in a browser - notably background tasks and filesystem/input APIs would need web/wasm-compatible implementations.

zeux:
This PR causes a significant regression in VRAM usage for me.

Here's `nvtop` comparing current production builds vs a local build off this branch; I verified that a local debug build off the commit on this branch before blade removal has the same VRAM requirement as the production build. The window sizes of two Zed instances are the same (2540x2076). This is running on a 4K screen (3840x2160), with DPI scaling 1.5x, using niri compositor on Ubuntu 25.10.

image

Here's `nvtop` with both windows maximized:

image

zeux:
@zortax Re: 30 MB difference - you are looking at the wrong column in my screenshots; the rightmost column is "host memory" which I'm not interested in; the 6th column accounts for GPU memory allocation and that is where the difference is significant for me.

zeux:
That seems to fix the problem for me; in fact, after this change, the VRAM consumption in this PR is better than it is in the production build. Here's a snapshot with both windows maximized:

image

zortax:
@zeux yes that shows that in your specific case it doesn't make a difference, my point was more that it is bad practice to reason about memory or VRAM usage based on debug build data. There many possible scenarios that could lead to significant differences, even in VRAM usage, not just performance, even in the sense that release mode could make it worse. Now that I think about this, it may even be a problem in my implementation currently: I use Fifo mode (V-Sync locked in wgpu), but rely on present() for synchronization. I also create new buffers on every render (blade used a pool), so if the GPU falls behind, GPU buffers could potentially pile up. This would be worse in release mode, because in debug mode the CPU part of the pipeline could be significantly slower. 

So again, the point of my comments is only that it is bad practice to base arguments on debug builds (especially when comparing against release builds), especially if the underlying "issue" is not even reproducible for everyone involved.

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1286 @@
...
+    pub fn gpu_specs(&self) -> GpuSpecs {
+        GpuSpecs {
+            is_software_emulated: self.adapter_info.device_type == wgpu::DeviceType::Cpu,
+            device_name: self.adapter_info.name.clone(),
+            driver_name: self.adapter_info.driver.clone(),
+            driver_info: self.adapter_info.driver_info.clone(),
+        }
+    }
+
+    fn create_storage_buffer(&self, data: &[u8]) -> wgpu::Buffer {
```
jlucaso1: this allocates a new GPU buffer for every single batch every frame - quads, shadows, sprites, underlines, etc. that's easily 10-20+ buffer allocations per frame, so at 60fps you're looking at ~1000 GPU allocs/sec.

the old blade renderer used `BufferBelt` from blade-util for exactly this reason, and the metal renderer on macOS uses a custom `InstanceBufferPool` (see `crates/gpui/src/platform/mac/metal_renderer.rs:56`). both pool and reuse buffers across frames

wgpu's `write_buffer` uses internal staging so it's safe to reuse a buffer from a previous frame even if the GPU hasn't finished reading it yet. a simple pool with two vecs (available + in_flight) swapped at the start of each frame would eliminate all these allocations after the first few frames stabilize

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1286 @@
...
+        self.queue.write_buffer(&self.gamma_buffer, 0, bytemuck::bytes_of(&gamma_params));
+
+        let globals_bind_group = self.device.create_bind_group(&wgpu::BindGroupDescriptor {
+            label: Some("globals_bind_group"),
+            layout: &self.bind_group_layouts.globals,
+            entries: &[wgpu::BindGroupEntry {
+                binding: 0,
+                resource: self.globals_buffer.as_entire_binding(),
+            }],
+        });
```
jlucaso1: `globals_bind_group` and `globals_with_gamma_bind_group` are recreated every frame but the underlying buffer handles never change (only the contents via `write_buffer`). bind groups reference buffer handles not their contents, so these can be created once in `new()` and stored as fields. small win but it's free

crates/gpui/src/platform/wgpu/wgpu_renderer.rs
```diff
@@ -0,0 +1,1286 @@
...
+            premultiplied_alpha: 0,
+            pad: 0,
+        };
+        let globals_buffer = self.device.create_buffer(&wgpu::BufferDescriptor {
+            label: Some("path_globals"),
+            size: std::mem::size_of::<GlobalParams>() as u64,
+            usage: wgpu::BufferUsages::UNIFORM | wgpu::BufferUsages::COPY_DST,
+            mapped_at_creation: false,
+        });
+        self.queue.write_buffer(&globals_buffer, 0, bytemuck::bytes_of(&globals));
```
jlucaso1: this creates a brand new uniform buffer + bind group every time paths are rendered, just to set `premultiplied_alpha: 0` instead of the main globals value. could just have a persistent `path_globals_buffer` field created once in `new()` and updated with `write_buffer` each frame, same as you do with `self.globals_buffer`. saves a buffer allocation + bind group creation per path batch

gohryt:
No problem with your concrete implementation and yes it will be zed maintainers decision. I am just hinting that this medal has two sides and while wgpu is good for libraries and apps with many different target OSes and architectures it has twice dependencies which means slow compilation and twice rust-analyzer ram usage at dev time, it works worser for vram and perfomance in run time and sometimes it has long-time problems without fixes as for example you can check its github issues.

You personally like wgpu, i personally dislike it especially after development and usage of apps based on it. My point was that wgpu is not solution for everybody and looks like something that can be made selectable by cfg flags based on checking  this pr.

zortax:
> it works worse for vram and perfomance

have you read the thread above? This statement is simply not true when comparing to blade. The implementation already has a lower VRAM and memory footprint than the blade implementation had, and that is before some performance improvements addressing the comments above. Performance will not change with this PR significantly either, the wgpu implementation does essentially the same thing as the blade implementation did. Feel free to run benchmarks on this...

>  slow compilation and twice rust-analyzer ram usage at dev time

Have you tried compiling the zed repo? Wgpu build times are not the issue or bottleneck here. On my system the difference was barely noticable, and since we have incremental builds for dependencies most of it will be one-time only anyway.

> sometimes it has long-time problems without fixes

Sure, wgpu has issues too. The difference is that it is actively maintained and a core dependency of many big projects. There is generally a big interest in getting significant things fixed. Blade on the other hand has open PRs (not even just issues) with small fixes that have a big impact (breaking issues) for downstream apps that haven't been merged in half a year. I don't want to attack the author, I know maintaining OSS is a lot of work and usually doesn't pay the bills, but zed and gpui are projects people rely on, so I don't think this is acceptable for a core dependency.

> You personally like wgpu, i personally dislike it especially after development and usage of apps based on it

I could also turn that argument around, but more importantly I think community support and stability are plainly more important that slight differences in API ergonomics.

> looks like something that can be made selectable by cfg flags based on checking this pr

Yes of course it could, but that has certain implications that should be considered. If you add it  as a new backend, it will also have to be maintained, so more work for further improvements. This is already a bit of an issue with GPUI given the different renderer implementation. I respect the decision of the zed maintainers to support those implementation, but personally I think we should thrive to consolidate the rendering infrastructure, not further fragment it. There are exciting things for the future that could be enabled (or significantly simplified for end-users) by consolidation, like APIs for custom shaders, direct buffer access for custom elements, embedding into other wgpu apps (or being embeddable by them), like bevy.

reflectronic:
I've merged some optimizations and simplifications. In particular, we're reusing the same buffer object for every frame now.

reflectronic:
> Maybe make this optional like gles instead of replacing blade?

The Blade backend is a lot of code that we (Zed) don't have a use for anymore; I don't think we're interested in maintaining it.

reflectronic:
f988a34 (the latest commit) is still a memory usage improvement over Zed Stable (listed as `zed-editor`):

image

<br>

Additionally, f988a34 is a ~20% CPU time improvement over b64007c40ef1b93fc841f78fef6d1b6d6bb28a3d (before I started optimizing) with approximately the same GPU time: 

```
frames 1320-1439: CPU draw median=238.82µs, GPU median=12.303ms
frames 1440-1559: CPU draw median=229.59µs, GPU median=12.304ms
frames 1560-1679: CPU draw median=240.19µs, GPU median=12.272ms
frames 1680-1799: CPU draw median=236.18µs, GPU median=12.466ms
```
vs.

```
frames 5400-5519: CPU draw median=301.45µs, GPU median=12.328ms
frames 5520-5639: CPU draw median=305.02µs, GPU median=12.269ms
frames 5640-5759: CPU draw median=295.21µs, GPU median=12.022ms
frames 5760-5879: CPU draw median=291.69µs, GPU median=11.812ms
```

The GPU time seems rather high, but this doesn't seem like it was caused by the pull request. (I have not properly benchmarked Blade, but the FPS does not feel higher for me in Zed Stable).

reflectronic (approved):
I think we're good to go with this. Thank you for your work on this, @zortax.

Akiyamka:
How to disable this? after update editor constantly freezes on Radeon 780M Graphic  (Framework Laptop)

jlucaso1:
> How to disable this? after update editor constantly freezes on Radeon 780M Graphic (Framework Laptop)

Can you provide some debug logs? I've tested here in a rx 6600 and it's running fine. Also check if you are using the latest mesa version

crates/gpui/src/platform/mac.rs
TimTheBig: True, I do think that [`WASI`](https://wasi.dev/) as a minimum feature-set would ease that work

Issue #45928, opened 2026-01-01
SSH Remote: LSP requests fail for language servers registered via `register_available_lsp_adapter` (ty, pylsp, pyright)


lingyaochu:
### Reproduction steps

1. Configure Zed to use `ty`, `pylsp`, or `pyright` as the Python language server in settings:
```json
{
  "languages": {
    "Python": {
      "language_servers": ["ty", "!basedpyright", "..."]
    }
  }
}
```
2. Open a Python project via SSH remote connection
3. Wait for the language server to start (diagnostics should appear)
4. Attempt to use "Go to Definition"
5. Result: Nothing happens - no navigation occurs

### Current vs. Expected behavior

Feature | Expected | Actual |
|---------|----------|--------|
| Diagnostics | ✅ Working | ✅ Working |
| Go to Definition | ✅ Working | ❌ **Fails silently** |
| Find References | ✅ Working | ❌ **Fails silently** |
| Hover | ✅ Working | ❌ **Fails silently** |
| Completions | ✅ Working | ❌ **Fails silently** |
| Signature Help | ✅ Working | ❌ **Fails silently** |

### Zed version and system specs

Zed: v0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1 (Zed) 
OS: Windows 10.0.26200
Memory: 47.8 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 770 || Intel Corporation || 32.0.101.7082

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

lingyaochu:
Here is the code analysis:

The bug is in `crates/project/src/lsp_store.rs` in the methods `check_if_capable_for_proto_request()` (line 4656) and `all_capable_for_proto_request()` (line 4685).

These methods determine whether the client should forward an LSP request to the SSH server by checking if any "relevant" language servers have the required capability. The problem is how "relevant" is determined:
```rust
let relevant_language_servers = self
    .languages
    .lsp_adapters(&language.name())
    .into_iter()
    .map(|lsp_adapter| lsp_adapter.name())
    .collect::<HashSet<_>>();
self.language_server_statuses
    .iter()
[...]

PR #43104, opened 2025-11-19
Add autocomplete for initialization_options

Nereuxofficial:
Closes #18287

Release Notes:

- Added autocomplete for lsp initialization_options

## Description
This MR adds the following code-changes:
- `initialization_options_schema` to the `LspAdapter` to get JSON Schema's from the language server
- Adds a post-processing step to inject schema request paths  into the settings schema in `SettingsStore::json_schema`
- Adds an implementation for fetching the schema for rust-analyzer which fetches it from the binary it is provided with
- Similarly for ruff
image

## Open Questions(Would be nice to get some advice here)
- Binary Fetching:
   - I'm pretty sure the binary fetching is suboptimal. The main problem here was getting access to the delegate but i figured that out eventually in a way that i _hope_ should be fine.
   - The toolchain and binary options can differ from what the user has configured potentially leading to mismatches in the autocomplete values returned(these are probably rarely changed though). I could not really find a way to fetch these in this context so the provided ones are for now just `default` values.
   - For the trait API it is just provided a binary, since i wanted to use the potentially cached binary from the CachedLspAdapter. Is that fine our should the arguments be passed to the LspAdapter such that it can potentially download the LSP?
- As for those LSPs with JSON schema files in their repositories i can add the files to zed manually e.g. in languages/language/initialization_options_schema.json, which could cause mismatches with the actual binary. Is  there a preferred approach for Zed here also with regards to updating them?

PR #46766, opened 2026-01-14
json_schema_store: Include available LSP adapters in settings schema

### LINKED ISSUE #46556: LSP property “ty" is not recognized in `settings.json` schema
### Reproduction steps

1. Add `ty` settings to settings.json, for example 
```json
{
  "lsp": {
    "ty": {}
  }
}
```
2. the editor gives message "Property `ty` is not allowed"

Current behavior: A warning appears for `ty` settings in `settings.json` ("Property ty is not allowed"). However, this does not affect the actual functionality of the `ty` language server.

Image

Expected behavior: Since `ty` is a built-in language server for Python, the settings.json schema should recognize it as a valid property so that users can configure it without seeing warnings.

Zed: v0.219.3+preview.116.3d83b3c9c66e9e5ef6f01c8849beede35264ed04 (Zed Preview) 
OS: Windows 10.0.26200
Memory: 47.8 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 770 || Intel Corporation || 32.0.101.7082

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json
  "lsp": {
    "basedpyright": {
      "settings": {
        "analysis": {
          "diagnosticMode": "workspace",
          "typeCheckingMode": "standard",
          "inlayHints.callArgumentNames": true,
          "inlayHints.variableTypes": true,
        },
      },
    },
    "ty": {
      "settings": {
        "diagnosticMode": "workspace",
        "inlayHints": {
          "variableTypes": true,
          "callArgumentNames": true,
        },
      },
    },
  },
```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

baeseokjae:
Closes #46556

  ## Summary

  - Fix "Property `ty` is not allowed" warning in `settings.json` for LSP adapters registered via `register_available_lsp_adapter()`
  - Add `available_lsp_adapter_names()` method to include these adapters in schema generation
  - Support `initialization_options` schema lookup for available adapters

  ## Problem

  LSP adapters registered via `register_available_lsp_adapter()` were not included in the settings JSON schema. This caused validation warnings like:

  Property ty is not allowed

  Even though `ty` is a built-in Python language server that works correctly.

  **Affected adapters:**
  - `ty`, `py`, `python-lsp-server`
  - `eslint`, `vtsls`, `typescript-language-server`
  - `tailwindcss-language-server`, `tailwindcss-intellisense-css`

  ## Solution

  Schema generation now queries both:
  1. `all_lsp_adapters()` - adapters bound to specific languages
  2. `available_lsp_adapter_names()` - adapters enabled via settings (new)

  Related: #43104, #45928 
  
  Release Notes:

- Fixed an issue where not all LSP adapters would be suggested for completion, or recognized as valid in `settings.json`

### DIFF
--- a/crates/json_schema_store/src/json_schema_store.rs
+++ b/crates/json_schema_store/src/json_schema_store.rs
@@ -264,14 +264,22 @@ async fn resolve_dynamic_schema(
             let mut lsp_adapter_names: Vec<String> = languages
                 .all_lsp_adapters()
                 .into_iter()
-                .map(|adapter| adapter.name().to_string())
+                .map(|adapter| adapter.name())
+                .chain(languages.available_lsp_adapter_names().into_iter())
+                .map(|name| name.to_string())
                 .collect();
 
-            for name in languages.available_lsp_adapter_names() {
-                let name_str = name.to_string();
-                if !lsp_adapter_names.contains(&name_str) {
-                    lsp_adapter_names.push(name_str);
+            let mut i = 0;
+            while i < lsp_adapter_names.len() {
+                let mut j = i + 1;
+                while j < lsp_adapter_names.len() {
+                    if lsp_adapter_names[i] == lsp_adapter_names[j] {
+                        lsp_adapter_names.swap_remove(j);
+                    } else {
+                        j += 1;
+                    }
                 }
+                i += 1;
             }
 
             cx.update(|cx| {
--- a/crates/json_schema_store/src/json_schema_store.rs
+++ b/crates/json_schema_store/src/json_schema_store.rs
@@ -5,7 +5,7 @@ use collections::HashMap;
 use gpui::{App, AsyncApp, BorrowAppContext as _, Entity, Task, WeakEntity};
 use language::{
     LanguageRegistry, LanguageServerName, LspAdapterDelegate,
-    language_settings::{AllLanguageSettings, all_language_settings},
+    language_settings::AllLanguageSettings,
 };
 use parking_lot::RwLock;
 use project::{LspStore, lsp_store::LocalLspAdapterDelegate};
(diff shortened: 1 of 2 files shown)


crates/json_schema_store/src/json_schema_store.rs
```diff
@@ -224,6 +225,9 @@ async fn resolve_dynamic_schema(
                 .all_lsp_adapters()
                 .into_iter()
                 .find(|adapter| adapter.name().as_ref() as &str == lsp_name)
+                .or_else(|| {
+                    languages.load_available_lsp_adapter(&LanguageServerName::from(lsp_name))
```
lingyaochu, line 229: I'm concerned about calling `load_available_lsp_adapter` here. It is not a pure "getter" for information; it has significant side effects and directly mutates the state of the language server module.

In my view, the settings schema resolution should be decoupled from the LSP lifecycle. We should only be querying the existing state rather than triggering a "load" action, which could lead to unintended consequences.

baeseokjae:
@cla-bot check

crates/json_schema_store/src/json_schema_store.rs
baeseokjae: @lingyaochu thanks for the review.

Just to clarify, `load_available_lsp_adapter` here doesn’t really act as a pure accessor, but it also doesn’t mutate shared state. It only takes a read lock and creates a temporary instance for schema lookup, then drops it.

For available adapters like `ty`, `eslint`, or `pylsp`, this is basically the only way to access their schemas today since they’re lazy-loaded and not part of `all_lsp_adapters`.

I agree that having schema resolution go through a “load” path isn’t ideal. That feels like something better handled as a separate issue later on. For this PR though, sticking with the current approach seems reasonable to me and I don’t see this causing issues in practice.

lingyaochu: Thanks for the clarification! I misunderstood how `load_available_lsp_adapter` worked under the hood. Given that it doesn't mutate shared state and is necessary for lazy-loaded adapters, this approach makes sense for now. I'm happy with the fix!

MrSubidubi:
@baeseokjae Can I ask you to re-sign the CLA? We track this only based on the current username, so would be nice if you could just sign it again. Thanks!

PR #46798, opened 2026-01-14
ui: Only show diff stat numbers if they're bigger than zero

danilo-leal:
No need to show, for example, "+25 -0" when there's been zero removed lines; the absence of that number communicates the message I think. :)

Release Notes:

- N/A

### DIFF
--- a/crates/ui/src/components/diff_stat.rs
+++ b/crates/ui/src/components/diff_stat.rs
@@ -29,34 +29,38 @@ impl RenderOnce for DiffStat {
         h_flex()
             .id(self.id)
             .gap_1()
-            .child(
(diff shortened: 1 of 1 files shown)


danilo-leal:
Uhm... interesting, I'm unsure how that's useful, though; maybe it's just a psychological thing (satisfaction?) of _seeing_ the 0, particularly when it's zero _added_ lines? It's an additional visual element for which its absence already communicates the thing, I think.

Screenshot 2026-01-14 at 11  37@2x

danilo-leal:
Okay, I'll back off from doing this for a bit then 😄 The consistency with Git and GitHub are definitely good points!

PR #46806, opened 2026-01-14
debugger: Fix crash when dragging pane items to split view

### LINKED ISSUE #46784: Zed crashes during debug when a debug pane (terminal, console, etc.) is moved to another one
### Reproduction steps

1. Start debug
2. Move (using mouse) terminal/console/variables/frames pane from it's original position inside to another pane as split view (vertical or horizontal)
3. Zed crashes

https://github.com/user-attachments/assets/78432fdd-7169-4baf-a070-39b8039f0886

Zed: 0.218.7 
OS: MacOS 17 15.7.3
Memory: 32GiB
Architecture: arm (M1 Max)

### Attach Zed log file

```log

```

</details>

xcb3d:
Closes #46784

This panic was caused by a double borrow when the pane an item was moved too is the same pane the item originate from. The fix was using a `window.defer` to avoid the double lease, instead of updating in a `Context<Pane>`

Release Notes:

- debugger: Fix crash when splitting a pane by dragging a pane item that originated from that pane


crates/workspace/src/pane_group.rs
```diff
@@ -877,9 +877,10 @@ impl PaneAxis {
     }
 
     fn pane_at_pixel_position(&self, coordinate: Point<Pixels>) -> Option<&Entity<Pane>> {
-        debug_assert!(self.members.len() == self.bounding_boxes.lock().len());
-
         let bounding_boxes = self.bounding_boxes.lock();
+        if self.members.len() != bounding_boxes.len() {
+            return None;
+        }
```
Anthony-Eid: Let's leave the `debug_assert` here.

If this assertion fails in a debug build that means there's another bug we have to fix

xcb3d:
Hi @Anthony-Eid. Added debug_assert! back per review feedback. Ready for re-review.

PR #46809, opened 2026-01-14
workspace: Move panel telemetry to workspace level

katie-z-geer:
This refactors the "Panel Button Clicked" telemetry event to fire from `workspace.rs` instead of `dock.rs`. 

Previously, the event was emitted in the `PanelButtons` render method's click handler. Now it fires at the workspace level in two places:
- `toggle_dock()` - captures panel toggles via dock buttons
- `focus_panel<T>()` - captures panel focus via keyboard shortcuts

This ensures the telemetry fires once per user action regardless of input method, and retrieves the panel name dynamically via `persistent_name()` rather than relying on a pre-computed local variable.

Release Notes:

- N/A


mikayla-maki (approved):
Looks good!

PR #46814, opened 2026-01-14
Agent Skills

rtfeldman:
This is behind a feature flag, so not including in release notes yet.

Release Notes:

- N/A


crates/paths/src/paths.rs
```diff
@@ -419,6 +419,13 @@ pub fn local_settings_folder_name() -> &'static str {
     ".zed"
 }
 
+/// Returns the path to the global skills directory.
+/// This is where user-defined agent skills are stored.
+pub fn skills_dir() -> &'static PathBuf {
+    static SKILLS_DIR: OnceLock<PathBuf> = OnceLock::new();
+    SKILLS_DIR.get_or_init(|| config_dir().join("skills"))
```
KuSh, line 426: [Agent Skills](https://agentskills.io/) specification [says](https://agentskills.io/integrate-skills#skill-discovery)
> Skills are folders containing a SKILL.md file. Your agent should scan configured directories for valid skills.

Actually, claude code scans for skills in .claude folder, opencode in .claude folder atm, vscode in .claude and .vscode folders.
I think we should allow multiple skills dirs to allow teams with people using different editors to work without supplementary shenanigans.

Perhaps default to .zed + .claude (historical folder) + .vscode (most used ide)
Otherwise search for SKILL.md files in the workspace + (those in the config dir if specified), and use that as a result

PR #46822, opened 2026-01-14
lsp: Add `container_name` to `lsp::Symbol`

aviatesk:
Some language servers include local symbols (e.g., local variables, parameters) in workspace symbol results. Without the `containerName` information, these symbols lack context information, making it difficult to distinguish them from top-level definitions and hindering efficient symbol lookup.

This change exposes the `container_name` field from LSP [`SymbolInformation`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolInformation) to the extension API, allowing language server extensions to access `symbol.container_name` in `label_for_symbol` and provide meaningful context when rendering symbol labels.

Note: The `container_name `field is added to all extension API versions because they seem to share the same underlying Rust types via wasmtime bindgen. The field is optional, so existing extensions would remain compatible as far as I understand.

Closes #ISSUE

Release Notes:

- Added `container_name` field to `lsp::Symbol`, accessible via the extension API's `label_for_symbol` function

### DIFF
--- a/crates/extension/src/types/lsp.rs
+++ b/crates/extension/src/types/lsp.rs
@@ -61,6 +61,7 @@ pub enum InsertTextFormat {
 pub struct Symbol {
     pub kind: SymbolKind,
     pub name: String,
+    pub container_name: Option<String>,
 }
 
 /// The kind of an LSP symbol.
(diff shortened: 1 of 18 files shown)


MrSubidubi (changes requested):
Thanks for this! 

If we want to add this, you'll need to follow the changes as outlined in https://github.com/zed-industries/zed/pull/46269 that does something similar. We cannot change the old WIT definitions. Existing extensions will break with your current set of changes, so we need to update this for v0.8.0 and keep the previous versions as is.

MrSubidubi (approved):
Thanks!

Issue #8316, opened 2024-02-24
better active selection coloring or indicator


tecandrew:
### Check for existing issues

- [X] Completed

### Describe the feature

Without a border around the current search selection or other active background indicator it's hard to tell at a glance where I'm at in a search. Maybe provide one or both customizable or default behavior option?

### Current Behavior

Ayu Dark
![image](https://github.com/zed-industries/zed/assets/21319730/0653b59a-1a45-4a65-ba47-721383387309)

Custom Light Theme
![image](https://github.com/zed-industries/zed/assets/21319730/ac81dc1d-fe85-42e5-8005-777d86f81362)

### VS Code 

![image](https://github.com/zed-industries/zed/assets/21319730/ce9ddd52-02e3-4344-9f0f-c73cd92c9446)

### Neovim
![image](https://github.com/zed-industries/zed/assets/21319730/bd7d3105-3cb4-402d-a30f-12ab508985de)

### Jetbrains

![image](https://github.com/zed-industries/zed/assets/21319730/89d4400c-8905-4522-a9e5-f6b8b71024c4)

thataboy:
I would like to echo this request. I have somewhat poor vision and need the contrast. Note that there is a tradeoff to simply making the selection background lighter since it decreases contrast. In sublimetext, I can have a border around selections and search results, making them very easy to see. Obviously, most people would not choose such bright colors, but I think this is an accessibility feature that will help everyone. And it can be turned off if not wanted.
 
<img width="1085" alt="Screenshot 2024-08-14 at 9 50 07 AM" src="https://github.com/user-attachments/assets/2fda57a0-ec60-421e-b29d-ee7ee254de34">
<img width="467" alt="Screenshot 2024-08-14 at 9 50 28 AM" src="https://github.com/user-attachments/assets/950baded-83f4-48de-9a10-a6c4b4255a37">

machin3io:
Bumping. With no easymotion or similar extension, jumping to specific places is difficult. 
If you have multiple search hits in a single line, it's possible to know where you are.

dargor:
Highlighting the current search result in a bright color would be nice indeed.

At the very least, highlighting the current line would make it easier to see where we are in the file.

Just checked vscode (I start it only to compare to zed nowadays), it does both and the result is way better.

zed: 
<img width="734" height="120" alt="Image" src="https://github.com/user-attachments/assets/463d5d94-81bf-41f6-a0eb-220c6683461a" />

vscode: 
[...]

PR #46825, opened 2026-01-14
gpui: Add hardware-accelerated blend modes for Metal

zambetti:
This PR adds support for hardware-accelerated blend modes for GPUI quads to the Metal renderer.

Currently, many UI elements in Zed rely on opacity/alpha blending, which can lead to low-contrast visuals in themes (e.g. selection highlight, search matches, scrollbar thumb and indicators, gutter indicators). Support for blend modes opens up possibilities for UI visuals to feel lighter and more vibrant, while offering better visibility and legibility in some scenarios.

### Benefit: Offloading Procedural Styling to the GPU

For text selection highlighting and block cursors, Zed handles low-contrast scenarios via procedural CPU logic. The CPU iterates through text to identify and override problematic color combinations. With hardware-accelerated blend modes for Quads, this kind of logic can be removed and text can remain unaltered, allowing the GPU to handle these cases automatically based on the theme element's configured blend mode (such as Invert, Multiply, or Screen).

For example in `element.rs`:
- `split_runs_by_bg_segments()`: Iterates through text fragments based on where a selection background exists. Calculates new colors using `ensure_minimum_contrast` to override the font color.
- `layout_visible_cursors()`: Procedurally determines if a block cursor should have black or white text inside it. This function has a comment: "TODO: In the future we should ensure themes have a text_inverse color". Instead, this could be handed by specifying a blend mode on the block cursor in the theme.

### Benefit: Opaque Scrollbar Thumb

Today, the [scrollbar thumb must be semi-transparent](https://github.com/zed-industries/zed/discussions/9490) to avoid obscuring underlying markers (such as search results or diagnostics). With support for blend modes, the scrollbar thumb would appear more saturated, while improving marker visibility.

### Benefit: Search Match Visibility

Differentiating the active match from other matches relies on transparency, which makes it difficult to maintain contrast. Support for blend modes on the selection highlight would allow for better contrast and visibility when it is applied to matches.

### Demo
To run the draggable rectangle example (screenshot below) that demonstrates these blend modes...
```
git clone -b gpui-blend-modes https://github.com/zambetti/zed.git
cd zed
cargo run --package gpui --example blend_modes
```
Screenshot 2026-01-14 at 13 14 15

### Changes
- Added dedicated MTLRenderPipelineState objects to the Metal renderer for Multiply, Add, Subtract, Invert, and Screen blend modes. Updated Metal renderer's Quad render pass to batch primitives by their blend mode to minimize pipeline state changes during a frame
- Updated PaintQuad and Scene for blend_mode property plumbing
- Added blend_mode to the Style struct and exposed it via the Styled trait. This allows any element to set a blend mode with, for example: .blend_mode(BlendMode::Multiply)
- Added blend_modes to GPUI examples

### API Design: `blend_mode` vs. `bg_blend_mode`

The property is currently named `blend_mode`. Since this implementation stays within the fast path of direct Quad rendering, only the background Quad is blended (drawing it directly to the main frame buffer). Children of the element are not included in the blend since group blending requires expensive offscreen rendering, which is unnecessary for Zed UI scenarios.

For API clarity, a thought is to rename this property to `background_blend_mode` or `bg_blend_mode` to make it explicit that only the background is affected, and reserve the more generic `blend_mode` name for possible future support if desired.

---
Related Issues & Disccussions:

- [Better active selection coloring or indicator](https://github.com/zed-industries/zed/issues/8316)

- [Allow custom active search background colour](https://github.com/zed-industries/zed/discussions/24640)

---
Release Notes:

- Added support for hardware-accelerated blend modes (Multiply, Add, Subtract, Invert, Screen) for Quads in the Metal renderer


ChristopherBiscardi:
Thanks for the PR! I think that we aren't going to pull this in at this time, as we don't really have a clear use for it in zed and it only implements support for the metal backend.

zambetti:
> it only implements support for the metal backend

I understand this concern. Cross-backend support makes this significantly more involved.

> we don't really have a clear use for it in zed

I disagree with this. There are several concrete uses for blend modes in zed, some of which I mentioned in the PR description, but the value isn’t purely functional. Blend modes enable richer visual layering, depth, contrast, and more refined visual interactions between UI elements, effects that aren’t possible with alpha compositing alone. And since gpui is used [beyond](https://github.com/zed-industries/awesome-gpui) zed itself, there’s value in supporting blend modes as a general rendering capability, even if zed doesn’t make immediate use of it.

If backend parity is the main blocker, would either full cross-backend support or a staged rollout with a backend-agnostic API be acceptable?

PR #46845, opened 2026-01-14
Split settings content into its own crate

probably-neb:
Closes #ISSUE

Moves the settings content definitions into their own crate, so that they are compiled+cached separately from settings, primarily to avoid recompiles due to changes in gpui. In that vain many gpui types such as font weight/features, and `SharedString` were replaced in the content crate, either with `*Content` types for font/modifier things, or `String`/`Arc<str>` for `SharedString`. To make the conversions easy a new trait method in the settings crate named `IntoGpui::into_gpui` allows for `into()` like conversions to the gpui types in `from_settings` impls.

Release Notes:


crates/terminal_view/src/terminal_element.rs
```diff
@@ -972,7 +969,7 @@ impl Element for TerminalElement {
                     let rem_size = window.rem_size();
                     let font_pixels = text_style.font_size.to_pixels(rem_size);
                     // TODO: line_height should be an f32 not an AbsoluteLength.
```
probably-neb:
```suggestion
```

crates/settings_content/src/merge_from.rs
```diff
@@ -87,18 +84,17 @@ impl<T: MergeFrom> MergeFrom for Box<T> {
     }
 }
```
probably-neb, line 86:
```suggestion

// Implementations for collections that extend/merge their contents
```

crates/settings_content/src/theme.rs
```diff
@@ -1,14 +1,117 @@
...
 
+/// OpenType font features as a map of feature tag to value.
+/// This is a content type that mirrors `gpui::FontFeatures` but without the Arc wrapper.
+/// Values can be specified as booleans (true=1, false=0) or integers.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, MergeFrom)]
+#[serde(transparent)]
+pub struct FontFeaturesContent(pub IndexMap<String, u32>);
+
+impl FontFeaturesContent {
+    /// Creates an empty font features map.
```
probably-neb:
```suggestion
```

osiewicz (commented):
This makes `settings` go from `2379500` llvm lines to `1821677` lines, whereas the new `settings_content` is at `570623`. I've honestly expected a bit more out of this and thus I'm wondering if there's a stray monomorphization somewhere.

osiewicz:
Ok, that's more like it.. :) With a small tweak I've went up to `1632100` in `settings_content` and down to `758589` in `settings`
[review.patch](https://github.com/user-attachments/files/24629381/review.patch)

osiewicz:
admittedly, I'm still seeing a bunch of serde-adjacent functions get codegened in `settings`, but that's a yak to shave for tomorrow. (attached is the output of `cargo llvm-lines -p settings` with my patch applied)
[new2.txt](https://github.com/user-attachments/files/24629442/new2.txt)

PR #46848, opened 2026-01-14
Add "Agent Panel Error Shown" telemetry with ACP error details

katie-z-geer:
Adds a new "Agent Panel Error Shown" telemetry event that fires when users see errors in the agent panel. For errors from external ACP agents (like Claude Code), we capture additional details.

Previously, we had no visibility into what errors users were encountering in the agent panel. This made it difficult to diagnose issues, especially with external agents.

The new telemetry event includes:
- `agent` — The agent telemetry ID
- `session_id` — The session ID
- `kind` — Error category (payment_required, authentication_required, refusal, other, etc.)
- `acp_error_code` — The ACP error code when available (e.g., "InternalError")
- `acp_error_message` — The ACP error message when available

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/acp/thread_view.rs
+++ b/crates/agent_ui/src/acp/thread_view.rs
@@ -1883,24 +1883,6 @@ impl AcpThreadView {
                     "You reached your free usage limit. Upgrade to Zed Pro for more prompts."
                         .into(),
                 ),
-                ThreadError::ModelRequestLimitReached(plan) => {
-                    let message = match plan {
-                        cloud_llm_client::Plan::V1(PlanV1::ZedPro) => {
-                            "Upgrade to usage-based billing for more prompts."
-                        }
-                        cloud_llm_client::Plan::V1(PlanV1::ZedProTrial)
-                        | cloud_llm_client::Plan::V1(PlanV1::ZedFree) => {
-                            "Upgrade to Zed Pro for more prompts."
-                        }
-                        cloud_llm_client::Plan::V2(_) => "Model prompt limit reached.",
-                    };
-                    ("model_request_limit_reached", None, message.into())
-                }
-                ThreadError::ToolUseLimitReached => (
-                    "tool_use_limit_reached",
-                    None,
-                    "Consecutive tool use limit reached.".into(),
-                ),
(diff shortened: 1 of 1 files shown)


crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -1831,10 +1851,47 @@ impl AcpThreadView {
...
+        let (error_kind, acp_error_code, acp_error_message) = match error {
+            ThreadError::PaymentRequired => ("payment_required", None, None),
+            ThreadError::ModelRequestLimitReached(_) => ("model_request_limit_reached", None, None),
+            ThreadError::ToolUseLimitReached => ("tool_use_limit_reached", None, None),
+            ThreadError::Refusal => ("refusal", None, None),
+            ThreadError::AuthenticationRequired(_) => ("authentication_required", None, None),
+            ThreadError::Other {
+                acp_error_code,
+                acp_error_message,
+                ..
```
mikebenfield: `Other` also has a `message` member, and we likely want to put that into the telemetry too.

crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -96,7 +96,11 @@ enum ThreadError {
     ToolUseLimitReached,
     Refusal,
     AuthenticationRequired(SharedString),
-    Other(SharedString),
+    Other {
+        message: SharedString,
+        acp_error_code: Option<SharedString>,
+        acp_error_message: Option<SharedString>,
```
mikebenfield: Is there a particular reason why we want separate members for `acp_error_code` and `acp_error_message`, as opposed to just letting them appear readable in the `message`?

katie-z-geer: I'd like to keep `acp_error_code` since that's giving us information that doesn't go into the message, but I'll remove `acp_error_message` since that should show in `message` regardless.

crates/agent_ui/src/acp/thread_view.rs
katie-z-geer: Good call, I updated this so all errors show the message.

PR #46854, opened 2026-01-15
Show active scope range on scrollbar

DomanskiFilip:
Closes Discussion Feature Request Show active scope range in scrollbar/minimap (similar to VS Code) #46472

I wanted to have a current scope markers/indicators and figured it is a small enough change that I can implement it myself. I believe it will be widely appreciated change since the feature is present in vs code. I tried to make it similar to other logic already in the codebase but if there need to be tweaks let me know and I will change it.

Release Notes:

- Added/Fixed/Improved ...
- New `active_scope_markers` setting for the scrollbar (true/enabled by default)
- Two small square markers on the scrollbar:
  - **Top marker**: shows the start of your current scope
  - **Bottom marker**: shows the end of your current scope
- Markers update automatically as you navigate between functions, methods, and classes
- Can be easily disabled via `settings.json`

## How to use

The feature is true/enabled by default, If you want to turn it off:
```json
{
  "scrollbar": {
    "active_scope_markers": false
  }
}
```

## Screenshots
turning the feature on

turnon

turning the feature off

turnoff

showcase demo in a vue file
showcase

## Testing

- [x] Markers appear at the correct positions
- [x] Markers update when moving between scopes
- [x] Setting toggle works properly
- [x] Works with files of any size
- [x] No performance issues -> I made sure to make the scope lookup once outside of any loops but if there could be some issues I missed let me know


MrSubidubi (changes requested):
Sorry for the delay, few notes:
- We should not by any chance be doing this on every frame. This should be part of the refresh in `refresh_slow_scrollbar_markers`
- We should disable it by default.
- We should add an entry to the settings UI for this.

Other than that, nothing against it and looks reasonable to have.

crates/editor/src/element.rs
```diff
@@ -7239,7 +7239,14 @@ impl EditorElement {
             let snapshot = layout.position_map.snapshot.clone();
```
MrSubidubi, line 7256: Is there any reason tat we cannot use the snapshot provided here in `current_scope_boundary`?

crates/editor/src/editor_settings.rs
```diff
@@ -228,6 +229,7 @@ impl Settings for EditorSettings {
                     horizontal: axes.horizontal.unwrap(),
                     vertical: axes.vertical.unwrap(),
                 },
+                active_scope_markers: scrollbar.active_scope_markers.unwrap_or(false),
```
MrSubidubi: You should be able to just `unwrap` here by adding an entry for this in the `default.json`

crates/editor/src/element.rs
```diff
@@ -6978,7 +6978,7 @@ impl EditorElement {
             return;
         };
         let any_scrollbar_dragged = self.editor.read(cx).scroll_manager.any_scrollbar_dragged();
-
+
```
MrSubidubi:
```suggestion

```

crates/editor/src/element.rs
```diff
@@ -7386,7 +7393,49 @@ impl EditorElement {
...
+                                    let marker_y = track_bounds.top() + y;
+                                    marker_quads.push(gpui::fill(
+                                        Bounds::new(
+                                            point(marker_x, (marker_y - marker_h / 2.0).round()),
+                                            size(marker_w, marker_h),
+                                        ),
+                                        color,
+                                    ));
+                                }
+                            }
```
MrSubidubi: This shoud still be using `markers_for_ranges`, is there any reason it should not uese that?

crates/editor/src/element.rs
```diff
@@ -7386,7 +7393,49 @@ impl EditorElement {
...
+                                let end_y = end_display_row.as_f64() as f32 * text_unit_size;
+                                
+                                let track_bounds = scrollbar_layout.hitbox.bounds;
+                                // size
+                                let marker_w = px(4.0);
+                                let marker_h = px(4.0);
+                                // position
+                                let marker_x = track_bounds.right() - marker_w - px(2.0);
+                                //color 
+                                let color = gpui::white().opacity(0.6);
```
MrSubidubi: We should not be using a hardcoded color here but instead one that makes sense coming from themes.

crates/editor/src/editor.rs
```diff
@@ -3747,6 +3747,25 @@ impl Editor {
         }
         result
     }
+    
+    // Returns the start and end row of the current scope the cursor is in, if any
+    pub fn current_scope_boundary(&self, window: &Window, cx: &mut App) -> Option<(u32, u32)> {
```
MrSubidubi: Not strictly necessary, but it would be good if you could add a test for this method somewhere so that we can ensure it does not regress.

crates/settings_ui/src/page_data.rs
```diff
@@ -2069,6 +2069,30 @@ fn editor_page() -> SettingsPage {
                 metadata: None,
                 files: USER,
             }),
+            SettingsPageItem::SettingItem(SettingItem {
+                title: "Active Scope Markers",
+                description: "Show markers in the scrollbar indicating the start and end of the current scope.",
+                field: Box::new(SettingField {
+                    json_path: Some("scrollbar.active_scope_markers"),
```
MrSubidubi, line 2076: Does a similar setting in VSCode exists?

MrSubidubi (changes requested):
Left a few notes. There is no need to ping us, we will get back to your PR if we find the time. That said, it would definitely help if you were able to test this yourself somehow. Thanks.

crates/settings_ui/src/page_data.rs
DomanskiFilip: in vs code its always on there is no way to disable it that I am aware of

crates/editor/src/element.rs
DomanskiFilip: I tried using markers for ranges and markers for quad ranges and it was misaligned in one case and invisible in the other

MrSubidubi: Hm but then I would prefer if we were to dig why that happened as opposed to adding additional logic here. Scrollbar markers in their current state are not set in stone, so I'd like this to use the API provided there if possible.

DomanskiFilip: ok I downgraded my env and freed space on disk and got testing working. Sadly there is bad news: after moving the logic to async (now also tested with separate dirty markers) the current scope markers don't show up probably because the latency of the slow render. The markers do show up when they are in the synchronous rendering I didn't test with special refreshing on click cos that would defeat the purpose of async anyway. The calculation isn't heavy and making it synchronous also ensures it is responsive and snappy like in vs code. There is GOOD news tho too I got the markers working using ```marker_quads_for_ranges``` -> previously I probably overcomplicated it when trying to utilize it looking at the initial commit

crates/editor/src/editor_settings.rs
DomanskiFilip: done

crates/editor/src/element.rs
DomanskiFilip: as said here: https://github.com/zed-industries/zed/pull/46854#discussion_r2756502029 when the painting logic is in the
```refresh_slow_scrollbar_markers``` the markers don't show up

File: crates/editor/src/element.rs (hunk 5 above)
Comment (DomanskiFilip) [reply crates/editor/src/element.rs]: done

File: crates/editor/src/element.rs (hunk 4 above)
Comment (MrSubidubi) [reply crates/editor/src/element.rs]: >  The calculation isn't heavy

It is, and thus, we'll have to do it in the background.

File: crates/editor/src/element.rs (hunk 1 above)
Comment (MrSubidubi) [reply crates/editor/src/element.rs:7256]: But then we should rather check as to why this happens and not just live with it.

File: crates/editor/src/editor.rs (hunk 6 above)
Comment (MrSubidubi) [reply crates/editor/src/editor.rs]: Thinking about this more, we should definitely test this here a bit to define the behavior.

File: crates/editor/src/element.rs [hunk 8]
```diff
@@ -7020,6 +7028,66 @@ impl EditorElement {
...
+                            
+                                window.paint_quad(gpui::fill(
+                                    Bounds::new(
+                                        point(marker_x, center_y - (marker_size / 2.0)),
+                                        size(marker_size, marker_size),
+                                    ),
+                                    color,
+                                ));
+                            }
+                        }
```
MrSubidubi: Again, we should not need this - we should rely on the existing APIs here and have these live in the scrollbar marker state and figure out the issue as to why these are not rendered as opposed to duplicating code.

MrSubidubi (changes requested):
Sorry for the slow reply, was sick and am still catching up across repos. Left some more comments.

crates/editor/src/element.rs
DomanskiFilip: the reason is that slow render refreshes on dirty and moving cursor marker(clicking somewhere) doesn't make state dirty so I would need to change dirty on click and as said in the linked comment I think that would defeat the purpose of having slow render cos every click it would re render. the natural place for this feature is in collect fast scrollbar render otherwise it would show after  you start typing not when you just click. In vs code it changes when you click on a different place and i wanted to recreate that but its ultimately up to you. I can make slow render re render on click or scope show up after you start typing.

MrSubidubi: Well but that changes nothing about the method that you call being potentially slow and thus it better belonging to be computed in the background.

Also, we could also update the `dirty` flag from a bool to an enum and that way ensure that we only update as needed (although practically, I think we already update on every click due to changes to the background highlights).

DomanskiFilip: ok 2 things then. when I tested, as is with panting in slow render it didn't show the markers  I can test it again maybe refactor ``` current_scope_boundry ``` or make dirty marker an enum as you said ( I think I would prefer that)  and try with dirty as enum if that fixes it. 

But that leads me to the second thing. If the slow render re renders on click anyway what is the difference between it and fast render just a delay of 500 or whatever? and if its just a delay doesn't adding a delay to a scope indicator make it a bad ux since its not as responsive?  just a genuine question cos I am kinda lost

crates/editor/src/element.rs
DomanskiFilip: > Again, we should not need this - we should rely on the existing APIs here and have these live in the scrollbar marker state and figure out the issue as to why these are not rendered as opposed to duplicating code.


do you mean that its in paint scrollbar or that its a custom defined square? It does use ```marker_quads_for_ranges``` as you wanted

DomanskiFilip: out of the variants I think the 2nd one is the closest to my custom one so we can go with it if you dont want the custom squares. its kinda longer/bigger bit close enough

image

MrSubidubi:
Sorry for the long delay here, I got pulled into another project and thus was limited on time. I'll throw some thoughts:
- the delay most likely comes from the fact that all the slow markers are resolved in one task as opposed to multiple/"streaming" them in of sorts. Yes, it is not good UX if stuff is delayed, it is worse UX if the UI freezes during render - we will not get around having this run on a task that is not blocking the main thread. That will most likely fix the issue you are seeing
- It seems the state management is a bit all over the place currently and there should be a bit more thought/refactoring around this - updating two fields on the struct all times should really be within a method so that in the future, this will not regress. 

Due to my imperfect response time, this also gained a lot of conflicts - sincerely sorry for this. Either way, I think at this point it would probably best if you were to revisit the approach chosen here - while many issues are now fixed (it is in far better shape compared to where we started), plenty are not and there is another set of big changes that needs to be made aside from conflict resolution.

Given this, I am gonna close this for the time being - I'll gladly look at another PR of yours that has everything tackled and the conflicts resolved should you still be interested in adding this (we'd definitely like to have it!). We decided recently that we need to close PRs more eagerly in some cases, and I do think this applies here. However, please do not let this discourage you from opening another one! Once you adressed this, resolved the conflicts and gave it another round of self-review, I'll gladly take another look. Thanks!

DomanskiFilip:
ok thanks I will get to it from the up to date commit sometime in the future then

PR #46856, opened 2026-01-15
file_finder: Fix "Reveal in File Manager" for WSL remote connections on Windows

maxmalkin:
Closes #46767

Release Notes:
- Fixed "Reveal in File Manager" not working for WSL remote connections on Windows.

 When using WSL remote connections on Windows, the "Reveal in File Manager" command fail without panicking.This happened because the worktree stores POSIX paths. Adds a `to_unc_path` method to `WslConnectionOptions` that converts POSIX paths to Windows UNC format. The `reveal_in_finder` handlers in the project panel, editor, and outline panel now check for WSL connections on Windows and convert paths accordingly before passing them to Explorer.


crates/remote/src/transport/wsl.rs
```diff
@@ -42,6 +42,22 @@ impl From<settings::WslConnection> for WslConnectionOptions {
...
+        let path_str = posix_path.to_string_lossy();
+        let path_without_leading_slash = path_str.strip_prefix('/').unwrap_or(&path_str);
+        PathBuf::from(format!(
+            "\\\\wsl.localhost\\{}\\{}",
+            self.distro_name,
+            path_without_leading_slash.replace('/', "\\")
+        ))
+    }
+}
+
```
Veykril: This the wrong way to do this, we should be using `wslpath` to convert between host <-> wsl paths, we already have host to wsl via `windows_path_to_wsl_path` so we should expose the reverse by using that same command with the corresponding different flag

crates/editor/src/editor.rs
```diff
@@ -22144,8 +20651,23 @@ impl Editor {
...
+                        wsl_options.to_unc_path(&path)
+                    } else {
+                        path
+                    }
+                } else {
+                    path
+                }
+            };
+
+            cx.reveal_path(&path);
```
Veykril: Given the duplication here, it feels like it would make more senes to expose a `reveal_path` function on the project itself

crates/editor/Cargo.toml
```diff
@@ -62,6 +62,7 @@ ordered-float.workspace = true
 parking_lot.workspace = true
 pretty_assertions.workspace = true
 project.workspace = true
+remote.workspace = true
```
maxdeviant: Please insert this dependency in the right spot alphabetically.

maxmalkin: This dependency is no longer needed based on changes from other review comments, removed it from `Cargo.toml`.

crates/editor/src/editor.rs
maxmalkin: Thanks for the suggestion, I didn't see that until now. Implemented in changes!

crates/remote/src/transport/wsl.rs
maxmalkin: Thanks! Implemented in fixes.

Veykril (approved):

SomeoneToIgnore:
To note, there's https://github.com/zed-industries/zed/discussions/47236 that seems like a good part of the same issue: project panel could use a context menu item same as other OSes have it.

Veykril (approved):
Thanks!

PR #46863, opened 2026-01-15
Smart focus management for tool permission requests

Brandtweary:
When a tool call requires user approval, the keyboard shortcuts (ctrl-y to allow, ctrl-alt-z to deny) don't work until the user manually clicks on the permission box. This is because the ThreadView never receives focus.

This fix implements smart focus management:
- Only steals focus if the message editor is empty (indicating user is waiting, not composing)
- Saves the previous focus before stealing
- Restores focus after any authorization action (allow/deny)

This is a UX improvement over other coding agents which interrupt users mid-typing to show permission dialogs. Zed's approach respects active composition - if you're typing, the permission notification appears but doesn't yank focus. Finish your thought, then ctrl-y when ready.

Related: Discussion #37784 (user requesting keyboard shortcuts - they exist but don't work without this fix)

## Release Notes

N/A

### DIFF
--- a/crates/agent_ui/src/acp/thread_view.rs
+++ b/crates/agent_ui/src/acp/thread_view.rs
@@ -357,6 +357,9 @@ pub struct AcpThreadView {
     skip_queue_processing_count: usize,
     user_interrupted_generation: bool,
     can_fast_track_queue: bool,
+    /// Saved focus handle to restore after tool authorization is complete.
+    /// Set when we steal focus for a permission request, cleared after auth action.
+    saved_focus_for_tool_auth: Option<FocusHandle>,
     turn_tokens: Option<u64>,
     last_turn_tokens: Option<u64>,
     turn_started_at: Option<Instant>,
@@ -549,6 +552,7 @@ impl AcpThreadView {
             skip_queue_processing_count: 0,
             user_interrupted_generation: false,
             can_fast_track_queue: false,
+            saved_focus_for_tool_auth: None,
             turn_tokens: None,
             last_turn_tokens: None,
             turn_started_at: None,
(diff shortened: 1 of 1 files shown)


Brandtweary:
This PR is a pragmatic workaround for the immediate UX problem, but after further reflection I wanted to document what a more complete solution might look like.

**Current limitation:** The `AllowOnce`/`RejectOnce` actions are registered at the `AcpThreadView` level (lines 8094-8096 in thread_view.rs), so Gpui's action dispatch requires ThreadView to be in the focus chain. This PR manages focus intelligently to enable that, but it's still fundamentally reactive - we're stealing and restoring focus rather than avoiding the problem entirely.

**Ideal architecture:** Register these actions at the `AgentPanel` level, delegating to the active ThreadView:
```rust
// In AgentPanel::render()
.on_action(cx.listener(|this, _: &AllowOnce, window, cx| {
    if let Some(thread_view) = this.active_thread_view() {
        thread_view.update(cx, |view, cx| view.allow_once(window, cx));
    }
}))
```
This would allow users to approve/deny tool calls even while actively typing in the message editor - no focus management needed at all, no interruption to flow.

**Future consideration for multiple queued permissions:** When the agent makes parallel tool calls requiring authorization, a "pseudo-focus" system could highlight which permission is next in the queue (distinct from actual keyboard focus). The highlighted permission receives the keybinding action; arrow keys could navigate the queue. This keeps authorization keyboard-driven without ever touching the message editor's focus.

I'd be happy to submit a follow-up PR for the panel-level action registration if there's interest. The current fix solves the immediate "shortcuts don't work" problem, but the deeper architectural change would provide a smoother experience.

PR #46866, opened 2026-01-15
ui: Dismiss context menus when window loses focus

notnotjake:
Context menus close when the window becomes inactive (e.g., clicking to another window or app). Previously, menus would remain visible until the window regained focus.

This aligns Zed's context menu behavior with native macOS apps, where NSMenu automatically dismisses when the window loses focus. Users expect menus to close when switching windows or apps.

Uses `observe_window_activation` to detect when the window becomes inactive and calls `cancel` to dismiss the menu.

Release Notes:

- Fixed context menus dismiss/cancel when the window loses focus

### DIFF
--- a/crates/ui/src/components/context_menu.rs
+++ b/crates/ui/src/components/context_menu.rs
@@ -296,14 +296,12 @@ impl ContextMenu {
                 this.cancel(&menu::Cancel, window, cx)
             },
         );
-        let _on_window_deactivate_subscription = cx.observe_window_activation(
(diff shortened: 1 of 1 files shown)


danilo-leal (approved):
Thanks!

ConradIrwin:
I'm not sure why yet, but this seems to have caused two problems: the terminal lost focus when window switching, and the macOS emoji picker stopped working. I'm reverting in #47835

PR #46867, opened 2026-01-15
workspace: Improve external file open to respect project boundaries

notnotjake:
Files opened from external sources (Finder, CLI, etc.) now open in a new window unless the file's path is inside an already-open project's worktree, or --wait flag is used.

Previously, Zed would handle directories correctly but would fall back to opening files in any available workspace window even if the file wasn't part of that project.

Added `test_external_file_opens_in_new_window` to verify files not belonging to any open project open in a new window. 

Closes #44355 

Release Notes:

- Improved external file opening to respect project boundaries, opening in a new window unless they belong to an already-open project

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -8097,6 +8097,7 @@ pub fn open_paths(
             });
 
             if open_options.open_new_workspace.is_none()
+                && (best_match.is_some() || open_options.prefer_focused_window)
                 && (existing.is_none() || open_options.prefer_focused_window)
                 && all_metadatas.iter().all(|file| !file.is_dir)
             {
--- a/crates/zed/src/zed.rs
+++ b/crates/zed/src/zed.rs
@@ -5185,4 +5185,55 @@ mod tests {
             });
         }
     }
+
+    #[gpui::test]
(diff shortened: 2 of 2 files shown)


mryall-mawson:
@dinocosta - can you please review this decision based on the [continued community feedback](https://github.com/zed-industries/zed/discussions/44355)? I suspect the team misunderstood the issue, and I was surprised to see what appears to be a working solution here ready to merge.

The problem is when Zed is mapped as the default text editor in your OS. At the moment, if Zed is your default editor and you double-click a text file you just downloaded, it opens in whatever current project context you have open in Zed. This is unexpected and different to other editors like VS Code, which [defaults to opening a new window](https://code.visualstudio.com/docs/editing/getting-started/userinterface#_window-management) unless the file is in an already-open project folder.

The only workaround in Zed currently is to close the file, manually open a new Zed window, and then open the file again. This is really clunky, and becomes a major usability issue for users with multiple projects who want to use Zed as their default OS editor for text files.

The `-n` command line flag is not a suitable workaround, because you cannot use a command line flag when an editor is opened by the operating system as a default file handler. For Zed to work properly as a default editor, the open-file behaviour needs to change. The implementation provided by @notnotjake here looks like a good start.

Thanks for all the great work on Zed!

Issue #16961, opened 2024-08-27
"Run All" command in REPL


srush:
### Check for existing issues

- [X] Completed

### Describe the feature

It will be very useful to have a "Run All" feature in the REPL that runs all the cell blocks individually. You can currently highlight everything but that only runs as if it was a big cell. In my use case I would like to see all the results aligned with the cells of the code.

haznai:
I'd also love a restart + run all option. It's easy to forget in which state your application/cell is, so just cleaning the state and re-running everything is killer ux.

i'd argue that things like [marimo](https://github.com/marimo-team/marimo) exist just because jupyter kernels have this ambigous state problem.

awesome job so far on the repl though, i really like my new workflow :)

ermaolaoye:
Agreed, very indeed wanting a Run All feature for the REPL environment.

gepcel:
Seems that not many people use zed with python. Are you still intrested in this feature? Would you mind leaving some comments on [Proposal 46594](https://github.com/zed-industries/zed/discussions/46594),thank you. @srush @zihang-huang @haznai

PR #46868, opened 2026-01-15
repl: Change default run actions to run line and add four run related actions

gepcel:
**edited in 2026-02-02**:

Should be able to close #46594, #19031, #16961

This pr changes the default behavior of two already existed actions: `Run` and `Run in place` to run lines. And add four other actions: `Run cell` and `Run cell in palce` to run cells, `Run above` to run cells above, `Run all` to run all cells.

* repl: run
* repl: run in place
* repl: run cell
* repl: run cell in place
* repl: run all
* repl: run above

Previous this pr there were two actions in repl: Run (included in repl menu, as shown in the following pic) and RunInPlace (not included in repl menu). When there are no cells (as of no `# %%` mark in python script file), these actions run current line or selected lines. When there are cells in file, these actions run current cell or selected cells, therefore no way to execute line inside a cell.

image

Release Notes:

- Fixed that default run actions of repl to run lines.
- Added 4 actions  `RunCell` , `RunCellInPlace`, `RunAbove` and `RunAll`.


rgbkrk:
At the very least, I like the change in naming if someone is "in" a cell, on a line, or selecting contents. I don't want to overload the menu with too many options though so I prefer to make it more context dependent. Now... actions wise it probably makes sense to have multiple, just not all shown in one menu.

gepcel:
Based on the feedback from @rgbkrk, the following changes have been made. The command palette now displays two original actions and four newly added actions:

* repl: run
* repl: run in place
* repl: run cell
* repl: run cell in place
* repl: run all
* repl: run above

The names of these actions are self-explanatory, but a few details are worth noting:

1. `run` and `run in place` execute the current line or selected lines now, different as previous.
2. `run cell` and `run cell in place` execute the current cell or the cells to which the selected lines belong. A special case occurs when the cursor (or selected lines) is positioned before the first cell block (i.e., before the first # %%). For continuity, the code preceding the first cell is treated as a cell and executed. This avoids the need to go to the top line and insert a new line and new # %%.
3. There are no changes to the shortcut keys; users may assign them as they prefer.
4. There are no changes to the REPL menu (located at the top right of the editor). However, it seems that @rgbkrk wishes to add only some of these actions to that menu, but not all. This can be addressed in a separate pull request.

crates/repl/src/repl_editor.rs
```diff
@@ -428,11 +436,29 @@ fn runnable_ranges(
...
+                return (Vec::new(), None);
+            }
+        }
+        ReplRunMode::Above => {
+            snippet_range = cell_range(buffer, 0, range.end.row);
+        }
+        ReplRunMode::All => {
+            snippet_range = cell_range(buffer, 0, buffer.max_point().row);
+        }
+    }
```
rgbkrk, line 626: It's been so long since I've looked at this. It's going to run it all as one batch rather than running each little block section?

gepcel: Yes, this means `repl: run all`. Execute all lines, no matter where the cursor is and selected lines. `ReplRunMode::Cell` will run each little block section (aka cell)

rgbkrk: I mean more like... if I have a bunch of jupytext cell blocks, and I submit run all, I'd expect each to be run individually with their output next to them.

gepcel: Oh, I get your point now. Makes sense, I'll try figuring it out.

gepcel:
@rgbkrk Now this is ready for review, please have a look.

rgbkrk:
Thank you for the contribution. I'll get this folded in somehow with the other ongoing changes to this and give you commit attribution.

Issue #39507, opened 2025-10-04
AI: Tailwind CSS IntelliSense not working in Angular inline templates and HTML files detected as "Angular" language


npajaro:
### Summary

Tailwind CSS IntelliSense not working in Angular inline templates and HTML files detected as "Angular" language

### Description
I'm experiencing issues with Tailwind CSS IntelliSense when working with Angular 20 projects in Zed. The `tailwindcss-language-server` doesn't provide completions in specific Angular contexts, even with extensive configuration attempts.

Steps to trigger the problem:
1. Install the Angular extension (v0.0.4 by nathansbradshaw) in Zed
2. Open an Angular project with Tailwind CSS v4 configured
3. Observe IntelliSense behavior in three scenarios:
   - **Scenario A**: Open an `.html` template file → Zed detects it as "Angular" language → No Tailwind IntelliSense
   - **Scenario B**: Open a `.component.ts` file with inline `template:` → Try to autocomplete classes inside `class="..."` within the template → No IntelliSense
   - **Scenario C**: In the same `.component.ts`, try autocomplete in `host: { class: '...' }` → IntelliSense works (with custom regex)

**Expected Behavior**:
- Tailwind CSS IntelliSense should work in files detected as "Angular" language type
- IntelliSense should work inside inline templates in TypeScript files (within `template: \`...\``)
- The `host` property should have IntelliSense by default without custom regex configuration

**Actual Behavior**:
- **Scenario A**: No IntelliSense when file is detected as "Angular". Works only when manually switching language to "HTML"
- **Scenario B**: No IntelliSense inside inline templates regardless of regex configuration
- **Scenario C**: Works in `host` property only with custom regex like `"class:\\s*[\"'`]([^\"'`]*)[\"'`]"`

### Environment Details
- **Zed Version**: Latest stable
- **OS**: [macOS/Linux/Windows]
- **Angular Version**: 20
- **Tailwind CSS**: v4.1.14 with `@tailwindcss/postcss`
- **Extensions**: Angular v0.0.4 (by nathansbradshaw)
- **LSP**: `tailwindcss-language-server` (installed locally in `node_modules/.bin/`)

### Configuration Attempted

**File**: `~/.config/zed/settings.json`

```json
{
  "lsp": {
    "tailwindcss-language-server": {
      "binary": {
        "path": "node_modules/.bin/tailwindcss-language-server",
        "arguments": ["--stdio"]
      },
      "settings": {
        "includeLanguages": {
[...]

PR #46870, opened 2026-01-15
Add injection-aware LSP support for embedded languages

sheikalthaf:
This adds support for language servers to work with embedded/injected languages like HTML templates inside Angular components. Key changes:

- Add `enable_injection_language_servers` setting (default: false)
- Add `delegate_languages` trait method for language LSP delegation
- Create virtual URIs for injection regions (e.g., file.ts.__virtual__.html_0.html)
- Implement position mapping between parent and injection coordinates
- Support injection-aware completions, hover, document highlights, and colors
- Register buffers with injection servers using extracted content
- Handle didOpen/didChange/didClose for virtual documents

The feature is gated behind a settings flag and when disabled, behavior is identical to before these changes.

Closes #ISSUE

https://github.com/user-attachments/assets/bd1e4823-6bea-45ea-a430-7c8654cf24fd

Angular:

https://github.com/user-attachments/assets/acef2530-565e-48bc-b474-a52654b9e685

Release Notes:

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1127,6 +1127,10 @@
   "show_call_status_icon": true,
   // Whether to use language servers to provide code intelligence.
   "enable_language_server": true,
+  // Whether to enable language servers for injection regions (embedded languages).
+  // When enabled, language servers will be started for embedded language content,
+  // such as HTML templates inside Angular components or CSS in style blocks.
+  "enable_injection_language_servers": false,
   // Whether to perform linked edits of associated ranges, if the language server supports it.
   // For example, when editing opening <html> tag, the contents of the closing </html> tag will be edited as well.
   "linked_edits": true,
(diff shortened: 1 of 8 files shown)


sheikalthaf:
closes. #39507

MrSubidubi (changes requested):
Thanks for this!

Can't help but wonder how much of this we could already solve with the exisiting `scope_opt_in_language_servers`

https://github.com/zed-industries/zed/blob/dc77f690e3c1343e0617cc1623f002096ec245ce/crates/language_core/src/language_config.rs#L107-L109

Curious whether we can not just work around the already existing infrastructure there and fix/discuss how we could adapt this to work for this use case, if it not already works for that?

MrSubidubi:
image

Will close as stale since this is getting conflict-heavy and in need of reply, happy to reopen with the above concern answered.

PR #46920, opened 2026-01-15
linux: Add support for building `.deb` packages

OrelSokolov:
Release Notes:

- Added script to build `.deb` packages, so every Ubuntu\Debian (33% of Linux market) user can install it via `.deb` 

Before:

- Project has snap builder, but it has issues for example on my Ubuntu 25.04 
- `cargo build --release ` requires more than 32GB of memory without CARGO_BUILD_JOBS=4
- With ENV var it takes half of my life to build release. ( ~ 25 minutes )
- Almost nobody can build editor by itself without performance issues of `debug version` ( too much asserts )
- PR allowing policy is so strict so I am unable to push updates directly to the core repo
- Sharing with my friend is almost impossible without `.deb` package. 

After:
- `Zed` now allows to build `.deb` package so it can beat popularity of Cursor

### DIFF
--- a/crates/zed/Cargo.toml
+++ b/crates/zed/Cargo.toml
@@ -284,3 +284,23 @@ osx_url_schemes = ["zed"]
 
 [package.metadata.cargo-machete]
 ignored = ["profiling", "zstd", "tracing"]
+
+[package.metadata.deb]
+name = "zed"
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore:
I've asked more teammates and no one got interested in this addition, sorry but I will have to close this one too.

PR #46932, opened 2026-01-15
Add optional relative line jumps to go-to-line action

KalaayPT:
Implements relative line jumping, a common feature i missed from Vim/Neovim where you can jump to lines relative to your current cursor position.

- New `{"relative": true}` flag for `go_to_line::Toggle` action 
- Supports both `-5` (5 lines up) and `b5` (backward 5 lines) syntax
- Full lines only (no column support in relative mode, wasnt sure if that would be necessary)
- Unbound by default - users can add e.g. `"ctrl-j": ["go_to_line::Toggle", {"relative": true}]` to their keymap

Example usage:
- `5` → jump 5 lines down
- `-3` or `b3` → jump 3 lines up  
- `0` → stay on current line

[Screencast_20260115_191312.webm](https://github.com/user-attachments/assets/395c0e7b-8ac1-48c8-a39e-5ade4f9206ec)

Release Notes:

- Added relative line jump support to go-to-line action via `+/-/f/b/F/B`

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -380,14 +380,17 @@ actions!(
     ]
 );
 
-actions!(
-    go_to_line,
-    [
-        /// Toggles the go to line dialog.
-        #[action(name = "Toggle")]
-        ToggleGoToLine
-    ]
-);
+/// Toggles the go to line dialog.
+/// When `relative` is true, the line number entered is interpreted as an offset
+/// from the current cursor position (positive for down, negative for up).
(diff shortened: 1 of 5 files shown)


KalaayPT:
updated to implement what we discussed. + and - are now built into go-to-line directly and support +/-/f/b/F/B. also only takes the last direction so "+b2" still jumps backwards, accounting for custom keybinds inserting a '+'. should i update the description? or only release notes?

demo:
[Screencast_20260116_111238.webm](https://github.com/user-attachments/assets/e1306e90-fa01-4101-80f3-b5cf20c291c8)

PR #46957, opened 2026-01-15
file_finder: Remove project's root name from the file finder history

### LINKED ISSUE #45135: Zed still includes project directory in file search results after hiding root
### Reproduction steps

Hit `cmd-p` to trigger file search, and look at search results.

Prior to `project_panel.hide_root` setting being added, the directories of file results didn't include the project directory. Disabling that setting removed the root directory in the project panel, but it still remains in the search results.

In the screenshot below, my project root directory is called "core", so I would be great if `core/*` was excluded from the directory names.

Image

Zed: v0.217.1+stable.98.56049ac3b58f2da9e171c3fee318f7d4944b0159 (Zed) 
OS: macOS 26.2
Memory: 8 GiB
Architecture: aarch64

CCXLV:
Closes #45135

Since neither the project search nor file finder search was showing project's root name, including it in the history was unnecessary, especially when the user had `project_panel.hide_root` set to `true`. 

Release Notes:

- Made file finder to respect `project_panel.hide_root` settings


SomeoneToIgnore (changes requested):
I have tried adding a test (as every such logical thing needs one!) but then realized this is a wrong fix?

If I have 2 worktrees (directories) added, I want to have worktree prefix shown for history items either.
Otherwise, the search will show the prefix but the history items won't which looks very odd:

https://github.com/user-attachments/assets/d2dafd0e-deaf-4b21-a6fe-4270a847c535

SomeoneToIgnore:
No, I think that's a valid case where we have to hide the prefixes.
In my video though, I have the defaults for this setting, `false`, and it works wrong.

crates/file_finder/src/file_finder.rs
```diff
@@ -1049,8 +1050,12 @@ impl FileFinderDelegate {
                     if let Some(panel_match) = panel_match {
                         self.labels_for_path_match(&panel_match.0, path_style)
                     } else if let Some(worktree) = worktree {
-                        let full_path =
+                        let project_panel_settings = ProjectPanelSettings::get_global(cx);
+                        let mut full_path =
```
SomeoneToIgnore: Let's restructure to have this code like `let full_path = if ProjectPanelSettings::get_global(cx).hide_root {....`-like form, so we do 1 allocation instead of 2 here, per each history entry.

SomeoneToIgnore (commented):
Right, good point, I also start to notice that after looking at 

> // Whether to hide the root entry when only one folder is open in the window.

documentation bit on the `hide_root`.

Fully agree: we should behave consistently (and have a test for it) + mention that file finder is also affected by this setting, in the default.json and wherever else (rustdocs?) so people are aware of all side-effects.

SomeoneToIgnore (approved):
Splendid, thank you so much.

PR #46965, opened 2026-01-16
Fix open path prompt not showing hidden files

austincummings:
Closes #39036 

The open path prompt will now show hidden files when "." is entered. Also fixes an issue with "open this directory" showing twice when used by the "toolchain: add toolchain" prompt.

With a tree of 
```
zed-industries
├── .hidden
├── .hidden-file
├── zed
├── zed-working
├── zeta
└── zeta-dataset
```
**Before:**
image

**After (current directory view without inputting "."):**
image

**After (when inputting "." to see hidden entries):**
image

Release Notes:
- Made Zed's built in file picker to show all hidden files by default

### DIFF
--- a/crates/open_path_prompt/src/open_path_prompt.rs
+++ b/crates/open_path_prompt/src/open_path_prompt.rs
@@ -41,7 +41,6 @@ pub struct OpenPathDelegate {
     replace_prompt: Task<()>,
     render_footer:
         Arc<dyn Fn(&mut Window, &mut Context<Picker<Self>>) -> Option<AnyElement> + 'static>,
-    hidden_entries: bool,
 }
 
 impl OpenPathDelegate {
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (commented):
My main concern with the approach is the absence of transparency: how a user is supposed to guess they have to input `.` first?

What if we return back all `show_hidden` functionality and always enable that for the path prompt instead?

austincummings:
Good point, it isn't very discoverable like this. I did remember at one point Zed had similar behavior so I went back to find the behavior. It looks like it used to in v0.190.0 behave similarly, however it required a `.` and then at least one other character to show the dotfiles. This video is from just prior to #32014.

I'm happy to add back the `show_hidden` and enable it though, just let me know!

https://github.com/user-attachments/assets/5b2c6270-88fb-4996-ab5b-68b94a1051c8

Commit SHA: `8c46a4f594b496fd666b8357a205ff228846d136`

SomeoneToIgnore (approved):
Thank you, let's see how this will be recepted, but the change itself looks nice.

Issue #46974, opened 2026-01-16
Zed is crashing when I select the option `Unstage All`


shettydev:
### Reproduction steps

1. Start zed
2. Add a few files in staging and make changes to the same files
3. Unstage all files

### Zed version and system specs

Zed: v0.219.4+stable.119.5595f7ce23d51a58eb66dd025a487bbd6c75dad1 (Zed) 
OS: Linux X11 ubuntu 24.04
Memory: 47 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3060 || NVIDIA || 580.95.05

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
2026-01-16T13:10:15+05:30 INFO  [lsp] Language server with id 2 sent unhandled notification @/tailwindCSS/clearColors:
[
  null
]
[...]

PR #47019, opened 2026-01-16
rope: Fix an unintentional panic

marcocondrache:
Avoids the panic in #46974

But should be solved by using `saturating_sub` on the `start_overshoot` calculation, ref: https://github.com/zed-industries/zed/pull/47046#issuecomment-3762855786

https://github.com/zed-industries/zed/blob/37715d5a50666baaa824eafec2131ee6d80f8e0b/crates/rope/src/chunk.rs#L403C5-L415C6

This function gets called with PANIC = false from `slice`. But if the offset is outside the text, `log_err_char_boundary` will still panic, because the first `if` doesn’t bail out early.

Release Notes:

- N/A


Veykril (approved):
Thanks!

PR #47024, opened 2026-01-16
helix: Fix normal-mode selection expansion during `vim::Scroll` commands

jrobsonchase:
Fixes the `vim::ScrollUp/Down` commands when a normal-mode vim or helix selection exists.

Pretty straightforward: make the `scroll_editor` function aware of the current vim/helix mode, and only expand the selection if a visual mode is active.

Closes #47022

Release Notes:

- Fixed bug causing normal-mode vim/helix selections to get expanded during `vim::Scroll` commands

### DIFF
--- a/crates/vim/src/helix.rs
+++ b/crates/vim/src/helix.rs
@@ -1758,4 +1758,81 @@ mod test {
             assert_eq!(vim_mode, Some(Mode::HelixNormal));
         });
     }
+
+    #[gpui::test]
+    async fn test_scroll_with_selection(cx: &mut gpui::TestAppContext) {
+        let mut cx = VimTestContext::new(cx, true).await;
+        cx.enable_helix();
+
+       // Start with a selection
+        cx.set_state(
+            indoc! {"
+            «lineˇ» one
+            line two
+            line three
+            line four
+            line five"},
+            Mode::HelixNormal,
+        );
+
+        // Scroll down, selection should collapse
+        cx.simulate_keystrokes("ctrl-d");
+        cx.assert_state(
+            indoc! {"
+            line one
+            line two
+            line three
+            line four
+            line fiveˇ"},
+            Mode::HelixNormal,
+        );
+
+        // Make a new selection
+        cx.simulate_keystroke("b");
+        cx.assert_state(
+            indoc! {"
+            line one
+            line two
+            line three
+            line four
+            line «ˇfive»"},
+            Mode::HelixNormal,
+        );
+
+        // And scroll up, once again collapsing the selection.
+        cx.simulate_keystroke("ctrl-u");
+        cx.assert_state(
+            indoc! {"
+            line one
+            line two
+            line three
+            line ˇfour
+            line five"},
+            Mode::HelixNormal,
+        );
+
+        // Enter select mode
+        cx.simulate_keystroke("v");
+        cx.assert_state(
+            indoc! {"
+            line one
+            line two
+            line three
+            line «fˇ»our
+            line five"},
+            Mode::HelixSelect,
+        );
+
+        // And now the selection should be kept/expanded.
+        cx.simulate_keystroke("ctrl-d");
+        cx.assert_state(
+            indoc! {"
+            line one
+            line two
+            line three
+            line «four
+            line fiveˇ»"},
+            Mode::HelixSelect,
+        );
+    }
 }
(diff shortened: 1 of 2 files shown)


jrobsonchase:
That isn't quite correct - I expect it to expand the selection if you're in select mode, which is helix's visual mode. I think that would miss vim's normal mode as well, and you'd still see selection expansion if you make it with the mouse and then use the scroll commands.

What situation are you still seeing the incorrect behavior in?

jrobsonchase:
Updated the test to ensure that the selection *does* get extended in helix select mode.

crates/vim/src/helix.rs
```diff
@@ -1694,4 +1694,81 @@ mod test {
...
+        cx.simulate_keystroke("ctrl-u");
+        cx.set_state(
+            indoc! {"
+            ˇline one
+            line two
+            line three
+            line four
+            line five"},
+            Mode::HelixNormal,
+        );
```
dinocosta: Should this be a `cx.assert_state` instead of `cx.set_state`? Changing it makes the test fail but I believe it's just because the cursor ends up in a different position ▼

```
Diff < left / right > :
<line one
>ˇline one
 line two
 line three
<line ˇfour
>line four
 line five
```

dinocosta (commented):
Hey @jrobsonchase 👋 

Thank you so much for fixing this and including tests for it too! I've left a single comment regarding the test, as I believe there's a place where state is being set where the intent might have been to actually assert state. 

Let me know if you have any questions! 🙂

crates/vim/src/helix.rs
jrobsonchase: Good catch!

I'm not sure why it's landing on line four though. It behaves that way before my fix, so I'm pretty sure I didn't inadvertently change that.

i.e. this tests against the current scroll-with-normal-selection behavior passes:

```rust
    #[gpui::test]
    async fn test_scroll_with_selection(cx: &mut gpui::TestAppContext) {
        cx.enable_helix();

        // Start with a selection
        cx.set_state(
            indoc! {"
            «lineˇ» one
            line two
            line three
            line four
            line five"},
            Mode::HelixNormal,
        );

        // Scroll down, selection should (not) collapse
        cx.simulate_keystrokes("ctrl-d");
        cx.assert_state(
            indoc! {"
            «line one
            line two
            line three
            line four
            line fiveˇ»"},
            Mode::HelixNormal,
        );

        // Make a new selection
        cx.simulate_keystroke("b");
        cx.assert_state(
            indoc! {"
            line one
            line two
            line three
            line four
            line «ˇfive»"},
            Mode::HelixNormal,
        );

        // And scroll up, once again (not) collapsing the selection.
        cx.simulate_keystroke("ctrl-u");
        cx.assert_state(
            indoc! {"
            line one
            line two
            line three
            line «ˇfour
            line five»"},
            Mode::HelixNormal,
        );

        // Enter select mode
        cx.simulate_keystroke(";");
        cx.simulate_keystroke("v");
        cx.assert_state(
            indoc! {"
            line one
            line two
            line three
            line «fˇ»our
            line five"},
            Mode::HelixSelect,
        );

        // And now the selection should be kept/expanded.
        cx.simulate_keystroke("ctrl-d");
        cx.assert_state(
            indoc! {"
            line one
            line two
            line three
            line «four
            line fiveˇ»"},
            Mode::HelixSelect,
        );
    }
```

I can't seem to reproduce this in practice though, so it might just be a quirk of the test harness.

dinocosta (approved):
Thank you for taking another look @jrobsonchase ! I've pushed a single commit fixing indentation in one comment, otherwise everything's good to go.

Will merge this shortly. Thanks! 🙂

crates/vim/src/helix.rs
jrobsonchase: Did a bit more digging and uncovered why `ctrl-u` was only going up by one line. The default "half page" scroll also takes the visual scroll state into account, not just cursor position. We could set the binding to scroll by a static number of lines for this test, but then we run the risk of not having the default half-page behavior tested. But if we keep it as-is, the number of lines scrolled could be altered by an unrelated to the test harness and its virtual window size.

Going to do a bit more reading of other tests to see if this situation is already addressed somewhere.

dinocosta: @jrobsonchase Thank you for taking a look! Seems like something minor we might not want to bother with right now? 🤔 

If you do end up wanting to go this route and finding something, let me know!

PR #47037, opened 2026-01-17
Show language server path in language server menu

JosephTLyons:
SCR-20260116-qrif

In most languages, Zed will check 2-3 locations for language servers. For instance, when opening a python file, zed looks for ruff in:

- project venv
- user's PATH
- location that zed installs language servers

The language server menu surfaces information that makes it easier to gather data about your language server, I think offering a path for the one being ran makes things more clear to the user.

That being said, Idk the best way to surface this. (@danilo-leal?)

Release Notes:

- Add language server path to a tooltip that is produced when hovering on its status in the language server menu

### DIFF
--- a/crates/language_tools/src/lsp_button.rs
+++ b/crates/language_tools/src/lsp_button.rs
@@ -20,7 +20,7 @@ use ui::{
     ContextMenu, ContextMenuEntry, Indicator, PopoverMenu, PopoverMenuHandle, Tooltip, prelude::*,
 };
 
-use util::{ResultExt, rel_path::RelPath};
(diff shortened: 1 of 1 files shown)


niekdomi:
Not sure if this is unrelated, but e.g., for `clangd` the  version is quite verbose:
image

I think truncation could improve the ui.

Sorry, if this is noise

JosephTLyons:
I'm going to go ahead and ship this, as we currently have tooltips with paths in them in Zed (see below), and this UI is fairly hidden. We can iterate in future PRs if this isn't ideal.

SCR-20260118-pksw

PR #47046, opened 2026-01-17
buffer_diff: Avoid usize underflow when computing start overshoot

cppcoffee:
Closes https://github.com/zed-industries/zed/issues/47045

Release Notes:

- buffer_diff: Avoid usize underflow when computing start overshoot


cole-miller:
I'm going to close this, because we don't have a reproducer for a problem caused by this non-saturating subtraction that would give us confidence that using saturating subtraction is the right fix.

PR #47079, opened 2026-01-17
Add tab switcher mode similar to vim/helix buffer picker

baldwindavid:
Implements the enhancement requested in #24655 - enables tab switcher to show all open files across all panes deduped. Whichever is the active pane is ALWAYS where the file, terminal, etc will be opened from the tab switcher.

Adds `tab_switcher::OpenInActivePane`. When enabled:

- **Shows all tabs across panes** - Same file in multiple panes appears once (deduped by path)
- **Opens in active pane** - Selecting a file opens it in the active/current pane, creating independent editor instances
- **Moves terminals** - Non-file items (terminals, multibuffers, etc.) move to current pane instead of duplicating

Intended for vim/helix `:buffers` command workflow while respecting Zed's tab-based paradigm.

## Q&A

**Q: Why not tie this to vim mode?**

A: This feature is different enough from vim/helix buffer behavior that I opted not to couple them. It is an approximation using Zed's tab switcher/picker and would take architectural changes to implement real vim/helix buffers. Additionally, both this and regular tab switcher mode (`tab_switcher::ToggleAll`) are useful window/tab management workflows that vim or non-vim users may want regardless of their editing mode preference.

**Q: How to use this?**

A: For a vim/helix-like workflow:

1. **Bind the tab switcher to a key** - Add to your keymap:
   ```json
   {
     "context": "Workspace",
     "bindings": {
       "cmd-i": "tab_switcher::OpenInActivePane"
     }
   }
   ```

2. **Hide the tab bar (optional)** - Add to settings:
   ```json
   {
     "tab_bar": {
       "show": false
     }
   }
   ```

3. **Move tabs to other pane on close (optional)** - Add to your keymap to join tabs into next pane when closing:
   ```json
   {
     "context": "Pane",
     "bindings": {
       "cmd-w": "pane::JoinIntoNext"
     }
   }
   ```

This makes closing a tab behave more like vim's split closure - items move to the adjacent pane instead of being closed.

If you find that you'd rather have the tab bar visible, there's a pretty good chance you don't need/want this feature. Vim/helix don't have visible tab bars, so this feature is primarily for users who want that sort of experience where buffers are not tied to a specific split/pane and closing a split does not impact the available buffers.

## Future Work

A few additional features could make for more complete vim/helix-like buffer workflow:

**`:bd` (buffer delete) improvements:**
- Should/could close the buffer from ALL panes, not just the current one
- Current implementation: `workspace::CloseActiveItem` only closes in current pane
- Needed: New action like `workspace::CloseActiveItemAcrossPanes` and/or update vim command mapping

**`:q` (quit) improvements:**
- Should/could close the split/pane
- Current implementation: workspace::CloseActiveItem closes active item, not the pane
- Needed: Perhaps just call `pane::CloseAllItems` to actually close the pane. This would be closer to what vim/helix does although those only close the "window" into buffers rather than all buffers within. Users of the behavior in this PR would likely want to have a keybinding to `pane::JoinIntoNext` like mentioned above. I don't think it's completely clear what the best course of action is, but something to consider.

**Previews:**
When first opened, this PR had preview functionality. However, that added a whole lot of complexity for a nice feature, but not the critical part of this workflow. Perhaps that could be approached in the future.

Release Notes:

- Add `tab_switcher::OpenInActivePane` to allow a more vim-like tab switching experience.


mkcode:
Thank you for creating this @baldwindavid. It's much appreciated.

To share some perspective, I tried getting a [similar feature merged into vscode 4 years ago](https://github.com/microsoft/vscode/pull/134525) and there were [follow ups](https://github.com/microsoft/vscode/pull/205442).

I bet this PR will go a lot better than the vscode folks. :)

alexispurslane:
Just chiming in to say that I am another user that would very much appreciate this functionality, as it would also allow emulating Emacs better as well. I found the existing behavior extremely confusing and opaque lol

baldwindavid:
Thanks, @ConradIrwin 
My first run at this was without previews and I really should have stuck with that since it got out of hand and with oh so many tests. I've pulled out the preview functionality which simplifies things quite a lot and cleaned out some tests.

This has also been changed to a standalone command named `tab_switcher::ToggleUnique`. Not sure that's better than `Merged` or there's also `Deduped`.

That it is a standalone command makes me think _maybe_ its okay to diverge from the other modes in lacking the previews, but also understandable if that feels like an incomplete/inconsistent feature. I know for my own usage, the critical part is not the previews, but it would also be nice to have. If this gets to a shippable state perhaps I'd take another run at previews later too.

The other thing I suspect is that all the modals will eventually end up going in a direction similar to #44530 with a preview decoupled from tabs, but that might be a big assumption.

> When testing the implementation I noticed that items were not removed from the pane

One thing I want to make sure is clear is that the intent is very much that you can easily end up with a file open in multiple panes. Regular files/buffers should "activate or clone" (never removed) in whichever pane you're in while terminals, multibuffers and such should "move" to whichever pane you're in. You wouldn't typically necessarily know or care which tabs are open in which panes because, if you're using this feature, you probably also hide the tab bar altogether, similar to vim/helix.

baldwindavid:
I've pushed the `OpenInActivePane` name change only as an option that's better than `ToggleUnique`, but it's of course real simple to change to anything from here.

Issue #45564, opened 2025-12-23
Incorrect file history link


t4m4k0ty4:
### Reproduction steps

1) git pull of remote branch
2) right click on file that was changed in last commit
3) choose "File History"
4) choose first commit in list
5) open window with diffs of other file that was changed in last commit too 

### Current vs. Expected behavior

**AS IS**

1) Open file history for needed file

<img width="1890" height="1084" alt="Image" src="https://github.com/user-attachments/assets/2af66070-fce8-49e3-8729-92ed077a4d1a" />

2)  Choose last commit for viewing changes

<img width="1453" height="254" alt="Image" src="https://github.com/user-attachments/assets/19188a96-ad82-48d9-b612-ee27a07c7152" />

3) See changes for different file

<img width="1448" height="934" alt="Image" src="https://github.com/user-attachments/assets/7e93d725-1d2a-40bb-894e-c91c2f2641c1" />

**TO BE**

1) Open file history for needed file

<img width="1890" height="1084" alt="Image" src="https://github.com/user-attachments/assets/2af66070-fce8-49e3-8729-92ed077a4d1a" />

2)  Choose last commit for viewing changes

<img width="1453" height="254" alt="Image" src="https://github.com/user-attachments/assets/19188a96-ad82-48d9-b612-ee27a07c7152" />

3) See correct diffs of chosen file

### Zed version and system specs

Zed: v0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1 (Zed) 
OS: macOS 15.5
Memory: 32 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

ian-h-chamberlain:
I'm not sure if this is related at all, but when I run `git: file history`, it sometimes shows me history for a different file than I had focused in the editor. Selecting from the sidebar seems to "reset it" to the right file.

I would just file as a separate issue, but it makes me think there's some kind of cache being used to show the file history and/or the commit links, so perhaps it can be "reset" as a workaround by loading history for a different file, then going back to the one you want to look at. 

[...]

PR #47088, opened 2026-01-18
git: Fix file history action for multibuffer + multi-select

ian-h-chamberlain:
There were a few pieces missing here:

- Project Panel used selection only, instead of "effective entries", meaning that right-click > File History would sometimes show history for the selected file or current editor, instead of the right-clicked file.

   Since the fix uses a set of selections, now it also supports multiple files from the project panel at once!

- In multibuffers, nothing would be picked, sometimes leading to the last used history view reappearing instead of the focused file. Use cursor position to pick a file instead when the editor is not a singleton.

This may be a fix for #45564 but I am not sure. I'm not able to reproduce it  with these changes, but I also couldn't reproduce before, so maybe I can ask OP if they want to try with these changes...

Release Notes:

- Fixed `git: file history` opening the wrong file's history from multibuffers.
- Added the ability to open file history for multiple selections in project panel.

### DIFF
--- a/crates/git_ui/src/git_ui.rs
+++ b/crates/git_ui/src/git_ui.rs
@@ -244,10 +244,15 @@ pub fn init(cx: &mut App) {
             let Some(active_item) = workspace.active_item(cx) else {
                 return;
             };
-            let Some(editor) = active_item.downcast::<Editor>() else {
+            let Some(editor) = active_item.act_as::<Editor>(cx) else {
                 return;
             };
-            let Some(buffer) = editor.read(cx).buffer().read(cx).as_singleton() else {
+            let buffer = editor.read(cx).buffer();
+            let Some(buffer) = buffer.read(cx).as_singleton().or_else(|| {
+                // if this is a multibuffer, use the cursor pos to pick a buffer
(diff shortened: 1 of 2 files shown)


crates/git_ui/src/git_ui.rs
```diff
@@ -244,10 +244,15 @@ pub fn init(cx: &mut App) {
...
+            let Some(editor) = active_item.act_as::<Editor>(cx) else {
                 return;
             };
-            let Some(buffer) = editor.read(cx).buffer().read(cx).as_singleton() else {
+            let buffer = editor.read(cx).buffer();
+            let Some(buffer) = buffer.read(cx).as_singleton().or_else(|| {
+                // if this is a multibuffer, use the cursor pos to pick a buffer
+                let anchor = editor.read(cx).selections.newest_anchor().head();
+                buffer.read(cx).buffer_for_anchor(anchor, cx)
+            }) else {
```
ian-h-chamberlain, line 255: I couldn't actually find where in the app this action handler is used, any view I tried never hit my breakpoints in this file (and used the project panel instead). Still, I figured it would make sense to keep the implementation consistent with the project panel impl

PR #45008, opened 2025-12-16
Add undo/redo in project panel

HactarCE:
Closes #5039

Release Notes:

- Added undo/redo for file creation, renaming, moving, and deleting in project panel

PR #47091, opened 2026-01-18
Add basic undo in project panel

### LINKED ISSUE #5039: Cannot undo actions in the project panel
### Summary

Can't undo renames, can't undo moved files or files that have been copied and pasted, deletes, etc.

### Environment

Zed: 0.70.0 (preview)
OS: macOS 13.1.0
Memory: 64 GiB
Architecture: aarch64

marcocondrache:
Initial part for #5039

Continues work from #45008. The author is no longer assigned to the issue, so I’m assuming it’s unclaimed. Please close this if that’s not correct.

- [x] Undo rename
- [x] Undo move
- [x] Undo paste
- [x] Undo drag
- [x] Enough tests

Release Notes:

- N/A


dinocosta:
> I looked at `trash-rs` when starting the PR, but I didn’t really like the implementation. It requires passing around handles to trashed items, which gets inconvenient, especially for remote worktrees since they’d need proto serialization. I’m currently working on a separate trash crate that abstracts the platform implementation like `trash-rs` does and allows restoration by only knowing the old path (I can explain how this works).

Yep, I believe we don't actually require all the information that is kept in `TrashItem` in order to be able to restore a file from trash. 

However, in Linux, are we actually able to restore a file from trash only with it's old path? As far as I can tell, not only is the file moved to trash, but a corresponding `.trashinfo` file is also created so, when restoring, we probably need to both move the file from trash back to its old path while also deleting the `.trashinfo` file, which is what the [current `trash::freedesktop::restore_all`](https://github.com/Byron/trash-rs/blob/3ca5818fac0f4eac0b9244f0bacf16c557683c2d/src/freedesktop.rs#L344) implementation seems to do 🤔

crates/project_panel/src/project_panel.rs
```diff
@@ -3096,26 +3215,51 @@ impl ProjectPanel {
...
                 }
+
+                if !operations.is_empty() {
+                    project_panel
+                        .update(cx, |this, _| {
+                            if operations.len() == 1 {
+                                this.record_operation(operations.pop().unwrap());
+                            } else {
+                                this.record_operation(ProjectPanelOperation::Batch(operations));
+                            }
```
dinocosta: It seems there's multiple places where we are performing this check, so we could likely extract this logic and avoid calling `unwrap` on the popped operation, even though we're checking the vector's length .

Maybe even updating `record_operation` to `record_operations(&mut self, operations: Vec<ProjectPanelOperation>)`? We can then match on the vector's length and also match on `Vec::pop` instead of unwrapping 🧹

crates/project_panel/src/project_panel.rs
```diff
@@ -2172,6 +2210,73 @@ impl ProjectPanel {
         }
     }
 
+    pub fn undo(&mut self, _: &Undo, _window: &mut Window, cx: &mut Context<Self>) {
+        if let Some(operation) = self.undo_stack.pop() {
+            let task = self.revert_operation(operation, cx);
+            cx.spawn(async |_, _| task.await).detach_and_log_err(cx);
```
dinocosta: I wonder if we actually need the `cx.spawn` here. Can't we just call `detach_and_log_err(cx)` on the result of `self.revert_operation(operation, cx)`? 🤔

crates/project_panel/src/project_panel.rs
```diff
@@ -2172,6 +2210,73 @@ impl ProjectPanel {
...
+                else {
+                    return Task::ready(Err(anyhow!("no entry for path")));
+                };
+                let task = self.project.update(cx, |project, cx| {
+                    project.rename_entry(entry_id, old_path.clone(), cx)
+                });
+                cx.spawn(async move |_, _| task.await.map(|_| ()))
+            }
+            ProjectPanelOperation::Create {
+                is_directory: _,
```
dinocosta: Seems like we actually end up not using the `is_directory` field in the current implementation, so we can probably remove it altogether. We can always add it back later, if necessary 🙂

crates/project_panel/src/project_panel.rs
```diff
@@ -156,6 +156,7 @@ pub struct ProjectPanel {
     sticky_items_count: usize,
     last_reported_update: Instant,
     update_visible_entries_task: UpdateVisibleEntriesTask,
+    undo_stack: Vec<ProjectPanelOperation>,
```
dinocosta: I'm now thinking that we might want to already have a limit to the number of `ProjectPanelOperations` we keep in the undo stack, otherwise this could lead to an unbounded growth of the stack, for example, if an user keeps Zed open for a very long time and performs a lot of operations in the Project Panel 🤔

crates/project_panel/src/project_panel.rs
```diff
@@ -2172,6 +2210,73 @@ impl ProjectPanel {
         }
     }
 
+    pub fn undo(&mut self, _: &Undo, _window: &mut Window, cx: &mut Context<Self>) {
+        if let Some(operation) = self.undo_stack.pop() {
+            let task = self.revert_operation(operation, cx);
+            cx.spawn(async |_, _| task.await).detach_and_log_err(cx);
+        }
```
dinocosta: Not sure if strictly necessary, but we might want to call `cx.notify()` here? I believe the filesystem watcher we setup eventually picks up the file changes from undoing an operation, but it might be worth explicitly notifying the UI to update 🤔

dinocosta (commented):
Hey @marcocondrache 👋 

Thank you so much for continuing the work on this! I've had a chance to review the current state of these changes and leave some comments in the code, where appropriate. 

Some feedback isn't particular to the file or code, so I'll leave it here instead:

* We'll probably need to think a little bit more on how we undo the `ProjectPanelOperation::Batch` operation. The current implementation seems to simply pop the operation off the stack and then attempts to undo all the operations within it. However, it seems that, if one of the operations within the batch fails to undo, we halt execution and silently "drop" the remaining operations. I wonder if we'll want to, at least, notify the user of which operations failed.
* We still need to explore how we're going to deal with conflicts. For example, right now, if I have a `src/main.rs` file and, through the Project Panel, rename it to `src/bananas.rs` but then, outside of Zed, create a new `src/main.rs` file, attempting to undo the rename will silently fail, while also popping the operation from the `undo_stack`. For the scope of this Pull Request, I believe we only need to care about displaying an error message to the user when undoing an operation fails, possibly with concrete information on what the operation was trying to do, for example, in the case shared above we'd likely want to tell the user that the rename failed because the target file already exists. In a future Pull Request, we might want to explore how to actually provide conflict resolution steps to the user, maybe something as simple as allowing the user to skip the operation or keeping track of the failed operations so we can retry, after the user manually resolves the conflict.
* Having an "Undo" option option to the right-click menu used in the project panel might also make it easier to discover that this feature exists.
    * Should be disable in case `undo_stack.is_empty()` is `true`
    * I might also, in the future, discuss with our design team if there's any icon/indicator/button we want to add for this, to make it easier for folks to know it exists
* I feel like maybe creating a new `crates/project_panel/src/undo.rs` module might be a good idea, so we can centralize all this undo/redo system logic and implementation there, especially since  `crates/project_panel/src/project_panel.rs` is already ~6500 lines long.

Hope this helps! Let me know if you have any question or suggestion regarding any of the points or if there's something you'd prefer we pair on. Thanks! 🙂

crates/project_panel/src/project_panel.rs
marcocondrache: Yes right, I've extracted the logic so this is handled better now

crates/project_panel/src/project_panel.rs
marcocondrache: Right, it was a leftover from the previous implementation

crates/project_panel/src/project_panel.rs
marcocondrache: Should be fixed

crates/project_panel/src/project_panel.rs
marcocondrache: I've added a limit of 10_000 operations which with the current enum (56 bytes) should use around 0.56MB of heap

crates/project_panel/src/project_panel.rs
marcocondrache: I've added it :)

marcocondrache:
@dinocosta thanks a lot for reviewing this!

- I extracted the logic into `undo.rs` to make it easier to manage and reason about. I've got inspired by https://developer.apple.com/documentation/foundation/undomanager
- I looked at how VSCode handles conflicts during undo operations. It simply shows a toast notification with the raw error message, I think we can do something similar.
- I updated the revert logic for ProjectPanelOperation::Batch so it attempts to undo all operations and then collects any errors that occur. We still need to decide how we want to surface or handle errors in this particular case.

dinocosta (commented):
Hey @marcocondrache ! 👋 

Thank you so much for tackling the provided feedback, and sorry for the late reply! I ended up pushing some changes, should be pretty easy for you to review what I changed based on each commit's message but here's a quick overview:

* Update the revert implementation for `Create` in order to actually trash and not delete the file
* Implemented error notifications when undoing a project panel operation fails
* Refactored `UndoManager::stack` to use `std::colllections::VecDeque` instead of `circular_buffer::CircularBuffer`
    * Since Rust's standard library already provides a structure to implement a ring buffer, I though we could use that instead, only downside is that we need to be the ones ensuring that we respect the operations limit, but that seems okay
    * I believe this will also make it easier to implement redo support, as we could easily keep track of a cursor/index in the vector, such that, undoing an operation moves the decreases the cursor and redoing simply increases the cursor. Any new actions would simply be inserted at the current cursor position, discarding any operations that existed from the cursor to the stack's end

Feel free to take a look at the changes! I'm also organizing next steps and will update you as soon as possible 😀

PR #47093, opened 2026-01-18
Fix Zed CLI column index broken on non-ASCII

### LINKED ISSUE #43329: zed cli column index broken on non ascii
### Summary

`zed /some/path/to/a/file.cpp:line:column` opens on a wrong column if file text is non ASCII. With ASCII content offset takes correctly.

### Description

Steps to reproduce:
1. `zed /some/path/to/a/файл.cpp:1:10`

**Expected Behavior**:
Zed opens correct file under path on the first line with column offset == 9.

**Actual Behavior**:
Zed opens correct file under path on the correct line, offset in column will be kinda twice lower than expected (it's 4 when column==10).

Zed: v0.213.3 (Zed) 
OS: macOS 26.1
Memory: 16 GiB
Architecture: aarch64

Test file content:

```
Докинуть картинки диаграммы в git lfs -- pdf, добавить в docker-ignore.

еще строка



еще одна строка просто длинная.
```

pigletfly:
Closes #43329 

before 
image
 after
image

Release Notes:

- Fix Zed CLI column index broken on non-ASCII


crates/file_finder/src/file_finder.rs
```diff
@@ -1575,7 +1574,14 @@ impl PickerDelegate for FileFinderDelegate {
                     active_editor
                         .downgrade()
                         .update_in(cx, |editor, window, cx| {
-                            editor.go_to_singleton_buffer_point(Point::new(row, col), window, cx);
```
SomeoneToIgnore, line 1601: Interesting, I'd expect file_finder to cover this one way or another.
Can we add a test on the topic into `file_finder_tests.rs` that fails before the change and passes after?

SomeoneToIgnore (commented):
Thank you, looks nice overall and great that many places were changed to the new approach.

One extra place that needs a similar change is https://github.com/zed-industries/zed/blob/e8cfd1941da1d3b92f7c23e7d54d47e6d018b7f2/crates/go_to_line/src/go_to_line.rs#L210 (it needs some modification though, but the problem here is the same multi-char UTF-8 approach).

PR #47095, opened 2026-01-18
agent: Check is_error flag in MCP tool responses

cppcoffee:
Previously, when an MCP server returned a tool response with
`is_error: true`, the error content was incorrectly treated as
a successful result. This could mislead the LLM into thinking
the tool call succeeded when it actually failed.

Now we check the `is_error` flag and propagate the error message
properly, allowing the agent to handle failures appropriately.

Release Notes:

- N/A

### DIFF
--- a/crates/agent/src/tools/context_server_registry.rs
+++ b/crates/agent/src/tools/context_server_registry.rs
@@ -382,6 +382,12 @@ impl AnyAgentTool for ContextServerTool {
                 }
             };
 
+            if response.is_error == Some(true) {
+                let error_message: String =
+                    response.content.iter().filter_map(|c| c.text()).collect();
+                bail!("MCP tool error: {}", error_message);
+            }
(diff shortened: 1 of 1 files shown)


crates/agent/src/tools/context_server_registry.rs
```diff
@@ -366,21 +351,21 @@ impl AnyAgentTool for ContextServerTool {
...
+                        arguments,
+                        meta: None,
+                    },
+                )
+                .await?;
+
+            if response.is_error == Some(true) {
+                let error_message: String =
+                    response.content.iter().filter_map(|c| c.text()).collect();
+                bail!("MCP tool error: {}", error_message);
```
benbrandt:
```suggestion
                bail!(error_message);
```

benbrandt (approved):
Thanks!

PR #47100, opened 2026-01-18
languages: Fix `poetry` environment discovery on Linux

lingyaochu:
Closes #47098 

The root cause of this issue is related to how `Poetry` (and the upstream `pet-poetry` library) handles path hashing. While perhaps it's an upstream behavior, we can easily fix it on the Zed side.

The related code are
https://github.com/zed-industries/zed/blob/7ce845210d3af82a57a7518e0abe8c167d60cc6a/crates/languages/src/python.rs#L1181-L1211

In my debugging, I found that `worktree_root` takes the form `/home/user/project`, but `config.workspace_directories` often ends up as `/home/user/project/`(with a trailing slash). Normally this wouldn't be an issue, but `Poetry` generates environment names based on the hash of the absolute path. Since the hashes for `/home/user/project` and `/home/user/project/` are different, `pet-poetry` fails to find the environment.

The fix is straightforward: we just need to ensure the trailing `/` is removed so the hashes match.

Release Notes:

- Fixed poetry environment not discovered on linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11603,7 +11603,7 @@ dependencies = [
 [[package]]
 name = "pet"
 version = "0.1.0"
-source = "git+https://github.com/microsoft/python-environment-tools.git?rev=1e86914c3ce2f3a08c0cedbcb0615a7f9fa7a5da#1e86914c3ce2f3a08c0cedbcb0615a7f9fa7a5da"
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Thanks!

PR #47101, opened 2026-01-18
Swap arrayvec crate for heapless to use LenT optimization

GnomedDev:
Swaps the `arrayvec` dependency for `heapless`, as the `heapless` library allows changing the type used for the `len` field, which `arrayvec` hard-codes to `usize`. This means that, for all the `ArrayVec`s in Zed, we can save 7 bytes on 64 bit platforms by just storing the length as a `u8`.

I have not benchmarked this change locally, as I don't know what benchmarking tools are in this project.

As a small bit of context, I wrote the PR to `heapless` to add this `LenT` generic after seeing a PR on the `arrayvec` crate that seems to be dead now. Once I saw some of Zed's blog posts about the `rope` crate and noticed the usage of `arrayvec`, I thought this might be a welcome change.

Release Notes:

- N/A


osiewicz:
Yes! I literally longed for this for several years now, never quite finding time to fork and what not. This is a very welcome change.
I can run some benchmarks for you.

osiewicz (approved):
The benches are quite noisy (seeing mixed results, that aren't even consistent between runs), but this feels like a substantial improvement to me.

Issue #18287, opened 2024-09-24
lsp settings should support autocomplete


notpeter:
### Check for existing issues

- [X] Completed

### Describe the feature

Currently we do not support autocomplete/suggestions for keys under the `lsp` object in settings so users are forced to operate blindly

Additionally, we should also support JSON schema validation for settings when available:

- `lsp.vtsls.settings` can be validated against the [vtsls configuration schema](https://raw.githubusercontent.com/yioneko/vtsls/refs/heads/main/packages/service/configuration.schema.json)
- `lsp.haskell-language-server.settings` a schema can be extracted from hls with `~/.ghcup/bin/haskell-language-server-9.10.1 vscode-extension-schema`:
<details><summary>haskell-language-server-9.10.1.settings.schema.json</summary><pre>
{
    "haskell.plugin.alternateNumberFormat.globalOn": {
        "default": true,
        "description": "Enables alternateNumberFormat plugin",
        "scope": "resource",
        "type": "boolean"
    },
    "haskell.plugin.cabal-fmt.config.path": {
        "default": "cabal-fmt",
        "markdownDescription": "Set path to 'cabal-fmt' executable",
        "scope": "resource",
        "type": "string"
    },
    "haskell.plugin.cabal-gild.config.path": {
        "default": "cabal-gild",
        "markdownDescription": "Set path to 'cabal-gild' executable",
        "scope": "resource",
        "type": "string"
    },
    "haskell.plugin.cabal.codeActionsOn": {
        "default": true,
        "description": "Enables cabal code actions",
        "scope": "resource",
        "type": "boolean"
    },
    "haskell.plugin.cabal.completionOn": {
        "default": true,
        "description": "Enables cabal completions",
        "scope": "resource",
        "type": "boolean"
    },
    "haskell.plugin.cabal.diagnosticsOn": {
        "default": true,
        "description": "Enables cabal diagnostics",
        "scope": "resource",
        "type": "boolean"
    },
    "haskell.plugin.callHierarchy.globalOn": {
        "default": true,
        "description": "Enables callHierarchy plugin",
        "scope": "resource",
        "type": "boolean"
    },
    "haskell.plugin.changeTypeSignature.globalOn": {
        "default": true,
        "description": "Enables changeTypeSignature plugin",
        "scope": "resource",
        "type": "boolean"
    },
    "haskell.plugin.class.codeActionsOn": {
        "default": true,
        "description": "Enables class code actions",
        "scope": "resource",
[...]

PR #47110, opened 2026-01-18
Add `initialization_options_schema` support for extensions

Nereuxofficial:
Follow-up for #18287

Release Notes:

- Added initialization_options_schema support for extensions

Things left TODO:
- [ ] Settings schema support
- [ ] Spawn task for extension initialization_options_schema fetching to not block async executor


MrSubidubi:
FYI I have also been looking into this recently and have a WIP setup locally for this which also adds settings schema support. With that said, please be informed that I might not go with your approach here, happy to discuss though.

PR #47112, opened 2026-01-18
Fix inverted char boundary check causing invalid offsets

jrdi:
The logic in `anchor_at_offset` and `to_offset` seems inverted when checking character boundaries. `assert_char_boundary` returns `true` when the offset IS valid, but the code was adjusting offsets when the function returned `true` (valid) instead of `false` (invalid).

Release Notes:

- N/A

### DIFF
--- a/crates/text/src/text.rs
+++ b/crates/text/src/text.rs
@@ -2417,7 +2417,7 @@ impl BufferSnapshot {
         {
             Anchor::max_for_buffer(self.remote_id)
         } else {
-            if self
+            if !self
                 .visible_text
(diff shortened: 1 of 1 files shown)


jrdi:
@Veykril, mentioning you since I've seen you make some changes to this over the past months. It's the first time I've checked Zed source code (just trying to understand how Ropes and SumTree work), and I've found these two IFs a bit weird.

I'm probably completely wrong since I don't understand a thing about the code base, just submitting the PR in case it's worth taking a look 🙏

jrdi:
@cla-bot check

Veykril (approved):
Right, thanks for catching that!

PR #47114, opened 2026-01-18
Add configurable REPL output size limits

cvanelteren:
Closes #47113

Adds configurable REPL output size limits with two new settings, `repl.output_max_height_lines` and `repl.output_max_width_columns`, so large outputs scroll instead of expanding and images scale down to fit the available space. The output containers in both inline REPL blocks and notebook cells now respect these bounds, and image sizing uses the same text metrics as the terminal output for consistent column-based width calculations.

Release Notes:

- REPL output now supports configurable max height and width limits, with large outputs scrolling and images scaling to stay within the viewport.


crates/repl/src/outputs/plain.rs
```diff
@@ -114,6 +114,27 @@ pub fn terminal_size(window: &mut Window, cx: &mut App) -> terminal::TerminalBou
...
+        return None;
+    }
+
+    let text_style = text_style(window, cx);
+    let text_system = window.text_system();
+    let font_pixels = text_style.font_size.to_pixels(window.rem_size());
+    let font_id = text_system.resolve_font(&text_style.font());
+    let cell_width = text_system
+        .advance(font_id, font_pixels, 'w')
+        .unwrap()
```
rgbkrk: Have a default come out here instead of unwrapping, if it seems sensible. That or return `None` as bailing would be worse than falling back to `None`.

docs/src/configuring-zed.md
```diff
@@ -21,7 +21,4730 @@ As you type in the search box, matching settings appear with descriptions and co
 
 ### User Settings
 
-Your user settings apply globally across all projects. Open the file with {#kb zed::OpenSettingsFile} or run `zed: open settings file` from the command palette.
```
rgbkrk, line 24: Did you mean to make all these changes to the configuring zed document? I have no opinions on the content but this should not be in this PR.

rgbkrk (changes requested):
This is sensible. Thanks for the contribution. Only two things to address - the `unwrap` in `max_width_for_columns` and removing the changes from `docs/src/configuring-zed.md`.

docs/src/configuring-zed.md
cvanelteren: Done

crates/repl/src/outputs/plain.rs
cvanelteren: Implemented, thanks.

docs/src/configuring-zed.md
```diff
@@ -1,136 +1,4998 @@
 # Configuring Zed
```
rgbkrk, line 1: Can you do a git checkout from main for this file so that we don't have any changes in this file?

cvanelteren: Not sure what went wrong there -- it is restored now

rgbkrk (approved):

rgbkrk (approved):

crates/repl/src/notebook/cell.rs
```diff
@@ -1157,6 +1169,78 @@ impl Render for CodeCell {
...
+                        div().py_1p5().w_full().child(
+                            div()
+                                .flex()
+                                .size_full()
+                                .flex_1()
+                                .py_3()
+                                .px_5()
+                                .rounded_lg()
+                                .border_1()
+                                // .border_color(cx.theme().colors().border)
```
miguelraz: These comments should be deleted now.

crates/repl/src/notebook/cell.rs
```diff
@@ -1157,6 +1169,78 @@ impl Render for CodeCell {
...
+                                                    Some(content.clone().into_any_element())
+                                                }
+                                                Output::ErrorOutput(error_view) => {
+                                                    error_view.render(window, cx)
+                                                }
+                                                Output::ClearOutputWaitMarker => None,
+                                            };
+
+                                            div()
+                                                // .w_full()
```
miguelraz: These comments also should be deleted if they're not being used

cvanelteren: Thanks.  Done.

crates/repl/src/notebook/cell.rs
cvanelteren: Thanks done.

cvanelteren:
@miguelraz is there anything else that needs to change?

miguelraz:
@cvanelteren Yes, I think some basic unit tests are indispensable here. The `repl` crate doesn't have many tests to start with and it would be healthy for test coverage to grow, IMO.

@rgbkrk has approved the implementation changes and I see nothing to change there.

miguelraz:
LGTM - I don't see any of the float checks going haywire in our pixel-calculating conditions, so they unit tests seem sound.

There are some clippy tests you should address @cvanelteren.

cvanelteren:
Thanks @miguelraz; I am not following the release schedule closely on zed, when would this be released on preview?

PR #47152, opened 2026-01-19
worktree: Use public-api-only tests

osiewicz:
- **settings: Do not leak type serialization into downstream crates**
- **worktree: Separate integration tests into separate target**

Release Notes:

- N/A


crates/worktree/Cargo.toml
```diff
@@ -40,6 +47,7 @@ log.workspace = true
 parking_lot.workspace = true
 paths.workspace = true
 postage.workspace = true
+pretty_assertions = { workspace = true, optional = true }
```
maxdeviant, line 50: I think this is also a pretty big downside of integration tests if we have to pull test dependencies up into `dependencies`.

Feels very strange to see `pretty_assertions` outside of `dev-dependencies`.

osiewicz: Agreed it is not ideal. Is it not worth the build time gains in your opinion? (Granted we're not yet sure how big the gains are outside of the crates I've ported so far).

maxdeviant: > Is it not worth the build time gains in your opinion?

My opinion is based on how large the gains are compared to how we have to contort our code to achieve them. Right now I'm pessimistic about the change to integration tests.

I still think that a lot of our build time issues come down to the way we've chosen to put a chain of "god crates" near the root of the tree (`project`, `editor`, etc.), and could be fixed with some good old software engineering/inversion of control.

And with us investigating using buck2/better build caching, it makes me wonder if the integration switch will be moot compared to the gains we can get elsewhere (without having to make a bunch of changes that, IMO, make tests worse).

osiewicz: > I still think that a lot of our build time issues come down to the way we've chosen to put a chain of "god crates" near the root of the tree (project, editor, etc.), and could be fixed with some good old software engineering/inversion of control.

Agreed, though I'm not sure how realistic that is given the API surface of these objects. 

> And with us investigating using buck2/better build caching, it makes me wonder if the integration switch will be moot compared to the gains we can get elsewhere (without having to make a bunch of changes that, IMO, make tests worse).

I am actually positive that integration tests change can augment both buck2 and better caching -  going with integration tests means that we only need to rebuild/cache a small binary with just tests (and nothing else); the build system simply cannot save us from building a crate twice that is required when using non-integration tests.

crates/worktree/tests/integration/main.rs
```diff
@@ -1,4 +1,5 @@
-use crate::{Entry, EntryKind, Event, PathChange, Worktree, WorktreeModelHandle};
+mod worktree_settings;
+
```
maxbrunsfeld, line 2: Can we still call this file `worktree_tests.rs` instead of `main.rs`?

crates/worktree/Cargo.toml
maxbrunsfeld: I actually think this is an improvement in some ways. Before, the correct way to have dependencies was, generally:
* add them to `dev-dependencies`
* **also** add them as optional dependencies, enabled by the `test-support` feature flag

Now you just do the second one.

maxbrunsfeld: > Agreed, though I'm not sure how realistic that is given the API surface of these objects.

I do think it's realistic; agents should be able to substantially speed up some of the rote work in extracting out sub-crates / traits / etc, from `project`, and `editor`.

osiewicz: Dyn-ifying `Project`/`Editor` is one part, but even that `dyn` interface would have a pretty large potential for necessitating rebuilds. Take a signature of `Project::create_buffer`:
```
    #[inline]
    pub fn create_buffer(
        &mut self,
        language: Option<Arc<Language>>,
        project_searchable: bool,
        cx: &mut Context<Self>,
    ) -> Task<Result<Entity<Buffer>>> {
```
Sure, we can make that a part of a trait (with a bit of wrangling around use of `Context<Self>`), but you're still left depending on a `language::Buffer`. Thus any change at *that* layer would still necessitate a transitive rebuild.

PR #47154, opened 2026-01-19
gpui: Improve Anchored to support center position

huacnlee:
Release Notes:

- N/A

Ref https://github.com/longbridge/gpui-component/pull/1956 extract my fork version of `anchored.rs` to let GPUI to support position Anchored at center.

https://github.com/user-attachments/assets/8d0230ed-4b75-440b-b8c3-9bde3decd141


Veykril (commented):
This seems reasonable to me, but we clearly need to come up with a new name for `Corner` here now, as this is no longer a corner

huacnlee:
I remembered that it was changed from `AnchorCorner` to `Corner` in PR https://github.com/zed-industries/zed/pull/22119

So, what do you think would be a more appropriate change?

- `AnchorPoint` 
- `Anchor`

huacnlee:
I still don't think `BoxAnchor` is good; it's not as elegant and perfect as `Corner` and `Anchor`.

Therefore, I switched to `Anchor`.

Veykril (approved):

Veykril:
sorry for the delay, if you could resolve the conflicts once more then I'll get this merged

huacnlee:
Wait me a moment, now I will do this.

huacnlee:
`cargo fmt` and `clippy` now fixed.

bowenxuuu:
@huacnlee 
I've found that the `Anchor::other_side_along()` method behaves inconsistently when handling `Axis::Vertical` and `Axis::Horizontal`. Is this intentional or an oversight?

anchor.other_side_along

huacnlee:
> @huacnlee I've found that the `Anchor::other_side_along()` method behaves inconsistently when handling `Axis::Vertical` and `Axis::Horizontal`. Is this intentional or an oversight?

This is right. I know that.

See the `TopLeft,` `TopRight`, `BottomLeft` ... in `Vertical`, they are not change the x way.

bowenxuuu:
Haha, I spaced out.

PR #47156, opened 2026-01-19
languages: Add attribute highlighting in C

ozacod:
Closes #46125.

Before:
before

After:
after

Release Notes:

- Updated tree-sitter-c for highlighting attribute specifier in C


ozacod:
@niekdomi Removed bracket highlighting, this is better, thank you.

after

PR #47158, opened 2026-01-19
diagnostics: Info level counter

### LINKED ISSUE #12817: [Dart] include info level diagnostics
### Check for existing issues

- [X] Completed

### Describe the feature

Currently only warning and error are shown inside diagnostics, as well as the diagnostics counter.
Although Zed get the info diagnostics from LSP Server and can display the message (as you can see in the image below), it doesn't include this inside diagnostics:deploy.
Because in dart there are linter warning that are at the info-level, it would be great to have this be included in the diagnostics.

Screenshot 2024-06-07 at 13 58 42

_No response_

oliverbarnes:
Closes [**zed-industries/zed#12817**](https://github.com/zed-industries/zed/issues/12817)

Implements info-level diagnostic counting and display throughout Zed’s project diagnostics UI and collaboration/proto layers.

## How to test:

1) Install https://github.com/zed-extensions/bash

2) Create a project with the following bash script that should trigger info-level notifications: https://gist.github.com/oliverbarnes/9cb5014974010dcc8720401aa5809fe4

2) Build and open target Zed:

```bash
cargo build --release
./target/release/zed
```

3) Open the bash script project on target Zed:

```
./target/release/zed  ../bash-project
```

(or just open it from the editor, of course)

Info-level diagnostics should now show together with warnings and errors.

For languages with no info-level diagnostics available, only warnings and errors will show, as before.

Release Notes:

- Added info-level diagnostics when include_warnings is true


crates/proto/proto/lsp.proto
```diff
@@ -1,971 +1,970 @@
 syntax = "proto3";
 package zed.messages;
 
+import "buffer.proto";
 import "core.proto";
 import "worktree.proto";
-import "buffer.proto";
 
 message GetDefinition {
-     uint64 project_id = 1;
```
oliverbarnes: Sorry for the large re-formatting diff, it was done automatically on saving the file. I can submit the reformatting as a separate PR, if preferred - let me know

l1qu1d:
This is really needed and one of the main reasons Zed isn't my daily code editor. I'd love to see this get merged sooner than later.

ConradIrwin:
The primary concern I have is that this PR changes the behavior of the toggle in the diagnostics pane. Today it "hides warnings", but with infos in the mix it will "show infos" (or for people like me in Rust, do nothing) until I click it twice.

I think we can fix that by changing the cycle order (first click should hide warnings). I also think it'd be nice to skip infos if there are none to avoid the double click being required to re-enable warnings.

One thing I'm not sure about is do users actually want the ability to see "errors and warnings, but not infos". If this isn't a common need, then we can keep the existing two-state toggle. you either see "just errors", or "everything". If people don't need this, then we can keep the button a two-state (but I don't know if people are happy because we don't show infos right now; or happy because they use language servers that don't report infos).

VsCode's defaults seem to be to show the counts for all three in the bottom, and show "errors and warnings" in the diagnostics panel; and that seems like a reasonable place to start.

Settings only updating on restart is unusual (and not ideal) - is there a reason that doesn't apply immediately?

ConradIrwin:
Sorry for the slow merge here. We need to internally apply the migration before deploying this change, so we can't add a not null field because older code won't be able to write to the column.

I think we could use a default (or just accept nulls and coalesce them to 0s)

oliverbarnes:
No worries - good catch. Defaulted to 0

PR #47159, opened 2026-01-19
Limit Helix surround edits to pairs containing cursor

mchisolm0:
Limit edits to only pairs which contain the cursor.
- Fixes #47155
- Only seemed to affect pairs after the cursor
- Fix should protect pairs before and after cursor

Closes #47155

Release Notes:

- Fixed helix surround delete/replace affecting pairs not containing the cursor


SomeoneToIgnore:
Given that this is a draft since January, is there any point to keep this open?

PR #44317, opened 2025-12-07
Add helix match surround operations

leonqadirie:
Partially addresses #38151

Release Notes:

- Added Helix match surround operations

---
Implements Helix-style surround operations under the `m` (match) menu:

- `ms<char>` - Add surrounding characters around selection
- `mr<from><to>` - Replace surrounding characters
- `md<char>` - Delete surrounding characters

It supports Helix' `m` as a special target character meaning "match nearest surrounding pair", allowing operations like `mdm` (delete nearest surround) and `mrm<char>` (replace nearest surround).

It also adds the concept of a `SurroundPair` as a single source of truth for all supported surround characters (brackets and quotes), used by both Vim and Helix flows.
- Vim helpers (`surround_pair_for_char_vim`, `bracket_pair_for_str_vim`) keep aliasing (`b/B/r/a`) while rejecting unknown characters.
- Helix helpers (`surround_pair_for_char_helix`, `bracket_pair_for_str_helix`) avoid aliases and treat unknown characters as symmetric pairs.

---
> [!NOTE]
> - I'm unsure about the `helix/surround.rs` as `surround.rs` exists already.
> - The `SurroundPair` abstraction seemed like a good idea but is definitely not required.

PR #47163, opened 2026-01-19
vim: Ensure surrounding markers contain cursor for Helix surround operations

leonqadirie:
Closes #47155, introduced by #44317

I naively trusted `crates/vim/src/object.rs:1624`'s `surrounding_markers` to check whether the cursor is actually surrounded, but it has nearest match fallback behavior, apparently by design as indicated by `test_anyquotes_object` in the same file.

Release Notes:

- Fixed Helix surround operations falling back to closest match if cursor not surrounded


kubkon (approved):
Looks great, thanks!

PR #47177, opened 2026-01-19
agent_ui: Parse @fetch URLs containing '@' correctly

### LINKED ISSUE #47111: `@fetch <url with a @ in it>` doesn't detect it as an URL
### Reproduction steps

1. Use the Zed Agent
2. Enter `@fetch <url with a @ in it>` (e.g. `@fetch https://www.npmjs.com/package/@matterport/sdk`)
3. See that it doesn't detect the URL

Currently, it doesn't detect an URL with a @ in it as an URL
I expect an URL with a @ in it to be recognize as an URL

Also, it's related but not the same issue:
When pasting a link after `@fetch`, it is often required to delete a character and write it back to detect it as URL. But this workaround doesn't work with the URL i tested (`@fetch https://www.npmjs.com/package/@matterport/sdk`) UNTIL i remove the `@` from the URL

Zed: v0.217.4+stable (Zed) 
OS: Linux Wayland nixos 25.11
Memory: 13.5 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 25.2.6

### Attach Zed log file

irrelevant

### Relevant Zed settings

irrelevant

### Relevant Keymap

irrelevant

oliverbarnes:
Closes https://github.com/zed-industries/zed/issues/47111

Fix `@fetch` mention parsing in the Agent Panel when the URL contains `@` (e.g. npm scoped package URLs). We now avoid treating `@` inside the URL as the start of a new mention, and for `@fetch` we consume the rest of the line as the URL argument. Added coverage for URLs with one or more `@` segments.

Screenshot 2026-01-19 at 23 28 00

Release Notes:

- Fixed `@fetch` not recognizing URLs that include `@` (such as npm scoped package links) in the Agent Panel.


oliverbarnes:
Just noticed that typing '@' still hides the pill, until we start typing again (then the pill comes back). Ideally the pill should remain throughout. Having another look

PR #47180, opened 2026-01-20
gpui: Add action_schema_by_name for O(1) action schema lookup

cppcoffee:
Release Notes:

- N/A

Avoids iterating through all registered actions when looking up a single action's schema in json_schema_store.


SomeoneToIgnore (approved):
Thank you.

PR #47195, opened 2026-01-20
 terminal: Reduce flicker on resize

pigletfly:
Skip PTY resizes for pixel-only changes and coalesce pending resize events.
  Snap standalone terminal layout to whole device pixels to avoid subpixel jitter.

before:

https://github.com/user-attachments/assets/0ad0db83-0099-44c8-b8d1-3dc8146b25ef

after:

https://github.com/user-attachments/assets/86278014-1c87-4263-a9e5-b58bcc1fa2ea

Release Notes:

- Fixed: Reduce terminal flicker on resize


probably-neb:
If you could please fix the compile errors, I think this is about ready to merge

pigletfly:
> If you could please fix the compile errors, I think this is about ready to merge

PTAL

probably-neb (approved):

PR #47196, opened 2026-01-20
gpui_linux: Add name/description to display for wayland

happenslol:
Not sure if I'm happy with having 2 wayland-only fields on `PlatformDisplay`, open to any suggestions. The only other idea I have is allowing access to the underlying `WaylandDisplay` through a wayland-only function.

The fact that we have to wait for a bit after startup for displays to be present is also unfortunate, but inherent to wayland, I suppose. I've updated the example to show that, could remove it again if not desirable.

Release Notes:

- N/A


happenslol:
I'm already using a fork of gpui for my own projects, so I'm gonna close it since it's probably not going to land.

PR #47202, opened 2026-01-20
remote: Add configurable remote server cache retention (default 5)

cppcoffee:
Release Notes:

- Improved add configurable remote server cache retention (default 5)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1350,6 +1350,7 @@ dependencies = [
  "ctor",
  "db",
  "futures 0.3.31",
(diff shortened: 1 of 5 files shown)


Veykril (commented):
I don't think we need to have this configurable

cppcoffee:
> I don't think we need to have this configurable

Downloaded files will be stored locally and will not be deleted. Keep only 5 files(default value), delete the older ones, or is there a better way?

Veykril (approved):
Thanks!

PR #47214, opened 2026-01-20
agent: Reuse ongoing project searches

osiewicz:
Co-authored-by: Smit Barmase <heysmitbarmase@gmail.com>

Release Notes:

- Improved performance of large agent searches.

### DIFF
--- a/crates/agent/src/tools/grep_tool.rs
+++ b/crates/agent/src/tools/grep_tool.rs
@@ -66,17 +66,12 @@ pub struct GrepToolInput {
 const RESULTS_PER_PAGE: usize = 20;
 
 pub(crate) struct GrepTool {
-    project: Entity<Project>,
     ongoing_project_searches: Entity<ProjectSearchStore>,
 }
 
 impl GrepTool {
-    pub(crate) fn new(
-        project: Entity<Project>,
-        ongoing_project_searches: Entity<ProjectSearchStore>,
-    ) -> Self {
+    pub(crate) fn new(ongoing_project_searches: Entity<ProjectSearchStore>) -> Self {
         Self {
-            project,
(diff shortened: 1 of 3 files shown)


crates/agent/src/tools/grep_tool.rs
```diff
@@ -48,31 +53,27 @@ pub struct GrepToolInput {
     /// Use "**/*.rs" to search Rust files across all root directories.
     /// </example>
     pub include_pattern: Option<String>,
-    /// Optional starting position for paginated results (0-based).
-    /// When not provided, starts from the beginning.
+    /// Optional results token from a previous search to continue from. Tokens can only be reused within a single turn.
+    /// They must also not be reused after a codebase is modified by the agent. Is it more efficient to reuse an existing search (if possible) than starting a new one, but you should not shy away from starting a new search if needed.
+    /// When not provided, starts from the beginning. The token is provided with the page of results.
```
benbrandt, line 58: I wonder if we should call this a more conventional name for pagination like `cursor`?
And what do we mean when we tell the agent "can only be reused within a single turn?

These instructions get passed to the agent directly, so I wonder if we can write this more as we would for api documentation or something.

I'll give it a shot, feel free to tweak:

```
    /// Optional cursor from a previous search to retrieve the next page of results.
    /// It is more efficient to reuse an existing search, but a new search must always be used if you've made edits to the codebase in the meantime. Also, cursors expire after the user sends a new message.
    /// When not provided, the search results start from the beginning. The cursor is provided with the page of results when more results are available.
```

What I am not sure how to convey the concept of a "turn". Hopefully the model is able to distinguish between user messages and tool results (they are both of type "user" usually).. but i wonder how well it will be able to distinguish this concept regardless

PR #47228, opened 2026-01-20
agent: Add initial support for debugger tool calls

Anthony-Eid:
This PR adds the initial support for debugging tool calling from the agent panel. It allows an agent to look at active debug sessions and examine the stack trace, variables, toggle standard breakpoints, and step through the debuggee. This isn't ready for the end user yet because: 

1) The agent isn't able to create a debug config or start a session.
- The first step is hooking up the adapter schema trait method to the agent.
- The second is adding finite permissions for allowing an agent to start a session.
- The third would be allowing the agent to use locators, but this isn't necessary.
2) Make the debugger followable.
- This will allow users to see what the agent is looking at in the debugger.
- This is also needed for collab debugger support. 
3) Extras.
- Allow agents to set other types of breakpoints. 
- Access to the console. 
- Access to a module list if it exists. 
- Give agents access to stop state history.

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,6 +167,7 @@ dependencies = [
  "collections",
  "context_server",
  "ctor",
+ "dap",
  "db",
  "derive_more 0.99.20",
  "editor",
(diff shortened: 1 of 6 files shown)


Anthony-Eid:
Closing this because we're planning on coming back to this in the future, and I don't have the bandwidth to work on this in my free time

PR #47245, opened 2026-01-20
onboarding: Fix theme picker always updating theme based on system appearance

HackAttack:
The settings store separate themes for Light and Dark modes. When the user clicks a theme, the code uses the system appearance to decide which to update, rather than the mode the user selected in the UI. This causes the wrong theme to be saved when the system appearance differs from the selected mode.

image

Release Notes:

- Fixed onboarding page saving the selected theme based on the system appearance and not the currently selected one.

### DIFF
--- a/crates/onboarding/src/basics_page.rs
+++ b/crates/onboarding/src/basics_page.rs
@@ -215,7 +215,7 @@ fn render_theme_section(tab_index: &mut isize, cx: &mut App) -> impl IntoElement
                     get_theme_family_themes(&theme).unwrap_or((theme.as_ref(), theme.as_ref()));
 
                 settings.theme.theme = Some(settings::ThemeSelection::Dynamic {
-                    mode: theme_mode,
+                    mode: ThemeAppearanceMode::System,
                     light: ThemeName(light_theme.into()),
                     dark: ThemeName(dark_theme.into()),
                 });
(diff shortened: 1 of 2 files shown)


crates/onboarding/src/basics_page.rs
MrSubidubi: Can we just make this if-check a match arm instead and match once on all theme mode variants? That way, we don't need the unreachable and I think I would slightly prefer that for readabiity

crates/onboarding/src/basics_page.rs
```diff
@@ -220,8 +220,12 @@ fn render_theme_section(tab_index: &mut isize, cx: &mut App) -> impl IntoElement
                     dark: ThemeName(dark_theme.into()),
                 });
             } else {
-                let appearance = *SystemAppearance::global(cx);
-                theme::set_theme(settings, theme, appearance, appearance);
+                let appearance = match theme_mode {
```
MrSubidubi: So that we basically have this match on theme_mode instead of the `if-else` stuff

MrSubidubi (changes requested):
Sincerely appreciate the updated description, with that, it became a lot easier to repro, thank you and nice find!

One comment and I think we are good to go!

crates/onboarding/src/basics_page.rs
HackAttack: Heh, I actually did that at first then changed it to make the diff smaller; but I agree that’s more idiomatic. Done.

MrSubidubi (approved):
This was what I meant so that the match here is more explicit (whilst also more verbose, but I think this is more readable).

Still, thank you for looking into and raising this and congratulations to your first contribution! 🎉

PR #47247, opened 2026-01-20
Support Insider variants of Code and VSCodium for settings.json import

maj113:
Release Notes: 

- Import VS Code Insiders and VSCodium Insiders user settings (`settings.json`).

Note:
- If multiple VS Code-family installs have a `settings.json`, Zed selects the last matching path in its search order (so Insiders is preferred over stable when both exist).


SomeoneToIgnore (approved):
Thank you.

PR #47261, opened 2026-01-21
Don't try to spawn conda if it's not there

ConradIrwin:
Closes #ISSUE

Release Notes:

- Fixed conda showing up in terminal windows

### DIFF
--- a/crates/language/src/toolchain.rs
+++ b/crates/language/src/toolchain.rs
@@ -12,8 +12,8 @@ use std::{
 use async_trait::async_trait;
 use collections::HashMap;
 use fs::Fs;
-use gpui::{App, AsyncApp, SharedString};
-use settings::WorktreeId;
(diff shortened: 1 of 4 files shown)


crates/language/src/toolchain.rs
```diff
@@ -113,7 +113,12 @@ pub trait ToolchainLister: Send + Sync + 'static {
         fs: &dyn Fs,
     ) -> anyhow::Result<Toolchain>;
 
-    fn activation_script(&self, toolchain: &Toolchain, shell: ShellKind, cx: &App) -> Vec<String>;
+    async fn activation_script(
+        &self,
+        toolchain: &Toolchain,
+        shell: ShellKind,
+        conda_manager: CondaManager,
```
osiewicz: This is not too pretty. Why can't we look that up within `languages/src/python.rs` directly?

PR #47292, opened 2026-01-21
ui: Improve performance in the `CopyButton` component

danilo-leal:
This PR improves the performance of the `CopyButton` component by tracking the copied state locally through a `CopyButtonState` struct instead of making an OS call every time the component re-renders. Also pushing a slight improvement here by resetting the state after two seconds so as to make the check mark go away after you clicked to copy.

Release Notes:

- N/A

### DIFF
--- a/crates/ui/src/components/button/copy_button.rs
+++ b/crates/ui/src/components/button/copy_button.rs
@@ -116,16 +116,14 @@ impl RenderOnce for CopyButton {
                     cx.write_to_clipboard(ClipboardItem::new_string(message.to_string()));
                 }
 
-                let state = state.clone();
-                window
-                    .spawn(cx, async move |cx| {
-                        cx.background_executor().timer(COPIED_STATE_DURATION).await;
-                        cx.update(|_window, cx| {
-                            cx.notify(state.entity_id());
-                        })
-                        .ok();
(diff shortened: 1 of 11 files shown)


crates/ui/src/components/button/copy_button.rs
```diff
@@ -56,38 +83,49 @@ impl CopyButton {
...
+                let state = state.clone();
+                window
+                    .spawn(cx, async move |cx| {
+                        cx.background_executor().timer(COPIED_STATE_DURATION).await;
+                        cx.update(|_window, cx| {
+                            cx.notify(state.entity_id());
+                        })
+                        .ok();
+                    })
+                    .detach();
```
Anthony-Eid: I think you could avoid cloning state here if you call state.entity_id() before moving it into the spawned closure.

Also, you might be able to get away with using cx.spawn instead of window.spawn

danilo-leal: Nice! Just pushed. Mind checking again?

maxbrunsfeld (approved):

crates/ui/src/components/button/copy_button.rs
Anthony-Eid: I'm late, but great work!

PR #47296, opened 2026-01-21
git_panel: Add error treatment when trying to push from an outdated branch

Aladiah96:
When I uploaded to a remote branch with newer commits on my work the push failed and I had no idea why until I used Git through CLI, so I enhanced the UX so no one needs to go through this anymore.

Release Notes:
- Improved git panel treatment error, previously if you pushed from an outdated branch, it logged a generic message and offered no solution, now it shows an specific message with a button that allows you to solve your issue by fetching, pulling and pushing.


cole-miller:
Thanks! I really like the idea of better handling the error here. However, I think the UX needs some more iteration. In particular, I'd like to give the user more control over the action they take to reconcile the remote and local branches--they might want to `pull --rebase` for example. I'm not sure exactly what the UI should look like, but I think we should find a way for the error condition to direct the user's attention to the normal remote button (the one in the panel that you click to push/pull/fetch), and suggest that they either `pull` or `pull --rebase`. Then they can push themself once that's been done.

PR #47344, opened 2026-01-22
project_panel: Add right-click download option for folders/files stored in remote development server

leoliu0605:
Closes #24431 , closes #42501

https://github.com/user-attachments/assets/13ace1d7-7699-4f2b-aa97-86235008adb3

Release Notes:

- Added right click download option for folders/files stored in remote development server

### DIFF
--- a/crates/project/src/project.rs
+++ b/crates/project/src/project.rs
@@ -2914,20 +2914,29 @@ impl Project {
         };
 
         let proto_client = remote_client.read(cx).proto_client();
-        let project_id = proto::REMOTE_SERVER_PROJECT_ID;
+        // For SSH remote projects, use REMOTE_SERVER_PROJECT_ID instead of remote_id()
+        // because SSH projects have client_state: Local but still need to communicate with remote server
+        let project_id = self.remote_id().unwrap_or(REMOTE_SERVER_PROJECT_ID);
         let downloading_files = self.downloading_files.clone();
         let path_str = path.as_ref().as_unix_str().to_string();
 
+        static NEXT_FILE_ID: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
+        let file_id = NEXT_FILE_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+
         // Register BEFORE sending request to avoid race condition
         let key = (worktree_id, path_str.clone());
-        log::debug!("download_file: pre-registering download with key={:?}", key);
+        log::debug!(
+            "download_file: pre-registering download with key={:?}, file_id={}",
+            key,
+            file_id
+        );
         downloading_files.lock().insert(
             key.clone(),
             DownloadingFile {
                 destination_path: destination_path.clone(),
                 chunks: Vec::new(),
                 total_size: 0,
-                file_id: None,
+                file_id: Some(file_id),
             },
         );
         log::debug!(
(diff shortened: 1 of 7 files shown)


crates/project/src/project.rs
```diff
@@ -2848,6 +2861,63 @@ impl Project {
...
+            destination_path
+        );
+
+        let Some(remote_client) = &self.remote_client else {
+            log::error!("download_file: not a remote project");
+            return Task::ready(Err(anyhow!("not a remote project")));
+        };
+
+        let proto_client = remote_client.read(cx).proto_client();
+        let project_id = proto::REMOTE_SERVER_PROJECT_ID;
```
Veykril:
```suggestion
        let Some(project_id) = self.remote_id() else {
            log::error!("download_file: not a remote project");
            return Task::ready(Err(anyhow!("not a remote project")));
        };
```

crates/project_panel/src/project_panel.rs
```diff
@@ -3010,6 +3017,182 @@ impl ProjectPanel {
...
+            if let Ok(Ok(Some(mut paths))) = destination_dir.await {
+                if let Some(dest_dir) = paths.pop() {
+                    // Show initial toast
+                    workspace
+                        .update(cx, |workspace, cx| {
+                            workspace.show_toast(
+                                workspace::Toast::new(
+                                    workspace::notifications::NotificationId::Named(
+                                        "download-progress".into(),
+                                    ),
```
Veykril: Let's extract this into a variable to be cloned from for the other toasts? Or a closure that produces this to dedup the string as we ought to make sure this doesn't differ between the toasts

crates/proto/proto/zed.proto
```diff
@@ -455,7 +456,11 @@ message Envelope {
         FindSearchCandidatesChunk find_search_candidates_chunk = 409;
         FindSearchCandidatesCancelled find_search_candidates_cancelled = 410;
         GetContextServerCommand get_context_server_command = 411;
-        ContextServerCommand context_server_command = 412; // current max
+
+        DownloadFileByPath download_file_by_path = 412;
+        DownloadFileResponse download_file_response = 413;
+        CreateFileForPeer create_file_for_peer = 414;
+        ContextServerCommand context_server_command = 415; // current max
```
Veykril: This one should stay at 412

crates/remote_server/src/headless_project.rs
```diff
@@ -681,6 +682,95 @@ impl HeadlessProject {
         Ok(proto::Ack {})
     }
 
+    pub async fn handle_download_file_by_path(
+        this: Entity<Self>,
+        message: TypedEnvelope<proto::DownloadFileByPath>,
+        mut cx: AsyncApp,
+    ) -> Result<proto::DownloadFileResponse> {
+        static NEXT_FILE_ID: AtomicU64 = AtomicU64::new(1);
```
Veykril: I feel like it would make more sense for the client to assign an identifier and hand that off in the `DownloadFileByPath` payload? instead of the server having this counter here

crates/project/src/project.rs
```diff
@@ -2881,6 +2894,73 @@ impl Project {
...
+            log::error!("download_file: not a remote project");
+            return Task::ready(Err(anyhow!("not a remote project")));
+        };
+
+        let proto_client = remote_client.read(cx).proto_client();
+        // For SSH remote projects, use REMOTE_SERVER_PROJECT_ID instead of remote_id()
+        // because SSH projects have client_state: Local but still need to communicate with remote server
+        let project_id = self.remote_id().unwrap_or(REMOTE_SERVER_PROJECT_ID);
+        let downloading_files = self.downloading_files.clone();
+        let path_str = path.as_ref().as_unix_str().to_string();
```
Veykril:
```suggestion
        let path_str = path.to_proto();
```

Veykril (commented):
One more tiny thing + clippy fixes

PR #47351, opened 2026-01-22
image_viewer: Add pinch event support

MostlyKIGuess:
This change implements pinch / magnification gesture handling.

This uses the following wayland [protocol](https://wayland.app/protocols/pointer-gestures-unstable-v1). And the following [API](https://developer.apple.com/documentation/appkit/nsevent/magnification) for mac.

- Original: https://github.com/gpui-ce/gpui-ce/pull/11

Release Notes:

- Zooming works with pinching in and out inside Image Viewer


agu-z (dismissed):
@MostlyKIGuess Thank you for your contribution. This is working well in my testing.

Before merging, I’d like to change the new GPUI methods to use `[cfg(any(target_os = "linux", target_os = "macos"))]` so consumers of the API are aware that this event isn’t available on Windows.

In the doc comments, we should also clarify that the event only works on macOS and Wayland, and explain how Windows simulates scroll wheel + <kbd>Ctrl</kbd> events on pinch.

In the future we may want to handle this complexity at the GPUI level, but I don’t think it’s a blocker as long as the API doesn’t suggest it should work when it won’t.

MostlyKIGuess:
@MrSubidubi Sorry for the tag, is there anything that should be done here?

MrSubidubi (approved):
Yes - I should hit merge.

Thank you!

PR #47376, opened 2026-01-22
lsp: Exclude dynamic settings from LanguageServerSeed identity

aviatesk:
`LanguageServerSeed` is used as a key to identify language servers. Previously, it included the entire `LspSettings`, which meant that changing `lsp.<server>.settings` (dynamic configuration) would cause the server to restart unnecessarily.

Dynamic settings can be updated via LSP's `workspace/didChangeConfiguration` notification without requiring a server restart. Only `binary` and `initialization_options` should be part of the server identity, as changes to these genuinely require restarting the server.

This is a follow-up fix to #35270 which introduced `LanguageServerSeed` but inadvertently included dynamic settings in the server identity (although I remember that this dynamic settings reflection stopped working pretty recently, so there might be other commits besides #35270 that changed the behavior of `LanguageServerSeed`)

Closes #ISSUE

Release Notes:

- Fixed language servers unnecessarily restarting when changing `lsp.<server>.settings` configuration. Dynamic settings are now properly updated via `workspace/didChangeConfiguration` without requiring a server restart.

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -36,7 +36,7 @@ use crate::{
         ManifestTree,
     },
     prettier_store::{self, PrettierStore, PrettierStoreEvent},
-    project_settings::{LspSettings, ProjectSettings},
+    project_settings::{BinarySettings, LspSettings, ProjectSettings},
     toolchain_store::{LocalToolchainStore, ToolchainStoreEvent},
     trusted_worktrees::{PathTrust, TrustedWorktrees, TrustedWorktreesEvent},
     worktree_store::{WorktreeStore, WorktreeStoreEvent},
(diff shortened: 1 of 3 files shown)


crates/project/src/lsp_store.rs
```diff
@@ -225,14 +225,42 @@ struct UnifiedLanguageServer {
...
+impl PartialEq for LanguageServerSeed {
+    fn eq(&self, other: &Self) -> bool {
+        self.worktree_id == other.worktree_id
+            && self.name == other.name
+            && self.toolchain == other.toolchain
+            // Only compare settings that require server restart.
+            // Dynamic settings (settings.settings) can be updated via DidChangeConfiguration
+            // without restarting the server.
+            && self.settings.binary == other.settings.binary
+            && self.settings.initialization_options == other.settings.initialization_options
```
osiewicz: I believe you could wrap these members into tuples and then compare the touples, which would make it a bit less mouthful. This is just a nit though.
So `self.foo == other.foo && self.bar == other.bar` should be equal with `(self.foo, self.bar) == (other.foo, other.bar)`

osiewicz (changes requested):
I wonder if instead of jumping hoops around comparison operators, we could instead create a newtype that would contain `settings.binary` and `settings.initialization_options`. I think that'd be less error-prone. What do you think?

crates/project/src/lsp_store.rs
aviatesk: Addressed. I left server_tree.rs as is though, since it only compares 3 fields and the current state might be easier imo. Also happy to change it if you want.

osiewicz:
Yeah, I feel somewhat strongly about it; for one, you don't need to provide `From` impl because it'd be the responsibility of whoever is using that newtype to figure out what they want to do with the `settings` part.

Right now `settings.settings` has no meaning whatsoever when it is extracted from the tree, because we quietly ignore it's contents. I'd rather not have `settings.settings` within the tree at all and force the user to figure it out. Your PR is a great contribution because it solves a real issue, but in face of that I don't think we should be using `LspSettings` anymore - we just want to know what settings the language was spawned with.

crates/project/src/lsp_store.rs
osiewicz: Thanks! Yeah I don't mind either way, it's a nit anyways. :)

osiewicz (approved):
Thanks. Let's address your concern on `start_language_server` in a follow-up PR. Also, the CI does not pass currently.

PR #47377, opened 2026-01-22
Add  "Add to .gitignore" action in project_panel

korbindeman:
This PR adds a "Add to .gitignore" action to the project panel's right-click context menu. Similar to the "Restore File" action that I previously added, I frequently find myself wanting this in the project panel.

image

With the restore file option:
image

Notes:
- **Implementation**: The `add_to_gitignore` function is essentially copy-pasted from `git_panel.rs`.

- **Error handling**: Added toast notification on error, which is consistent with `restore_file` in project_panel and `perform_checkout` in git_panel.
Note that `add_to_gitignore` in git_panel does NOT show a toast (just uses `detach_and_log_err`). I don't know if this is on purpose.
To follow up, I can either: match the project_panel implementation to the git_panel one (no toast), or update the git_panel implementation to also show a toast on error.

- **Menu grouping**: Previously "Restore File" and "View File History" were in separate sections, but both relate to git. With this third git action, I grouped all three together under a single separator (see screenshot).
We could also keep "View File History" separate and only group "Restore File" + "Add to .gitignore" together (both modify the working tree state in some way), if we don't want to alter the existing UI too much.

Release Notes:

- Added "Add to .gitignore" option to the project panel context menu for files in git repositories.


cole-miller:
Thanks! Could we extract the code to add a path to `.gitignore` to a crate (maybe `git`) where both `project_panel` and `git_ui` can refer to it? Other than that this seems good.

ChristopherBiscardi (approved):
This is looking good. I've merged main and made some updates that were required from other refactors in the meantime.

This will create a new .gitignore if none exists, and add to if it does, which is super nice, thanks!

Image

aresstokrat:
image
would be nice if it will work with Remote Repo at Ubuntu (i'm using Zed at Windows 11 as client to host Ubuntu)

korbindeman:
Good point. Although, this PR uses the existing `.gitignore` logic that was already in `git_panel.rs` to add a button to the project panel as well. Remote support is likely a deeper issue worth filing separately rather than something specific to this PR.

PR #47423, opened 2026-01-22
editor: Fix Shift+Enter inserting newline at end instead of cursor position with JetBrains keymap

### LINKED ISSUE #47269: Shift+Enter inserts new line at end instead of cursor position
## Steps to Reproduce

1. Open a new project in Zed
2. Go to the agent panel
3. Type two words, e.g. `one two`
4. Set the cursor after the first word: `one| two` (cursor between "one" and "two")
5. Press Shift+Enter

## Expected Behavior

A new line should be inserted at the cursor position.

The new line is inserted at the end (after "two") instead of at the cursor position

Zed: v0.219.5+stable.122.afc07d9fb20c3b7451071e2494d75f8837cb28db (Zed) 
OS: macOS 26.2
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

mason-palmer:
Closes #47269 

Release Notes:

- Fixed shift-enter inserting newline at end of text instead of cursor position in the agent panel when using JetBrains keymap.

# Before:

(tested using shift-enter then enter in both editors)

https://github.com/user-attachments/assets/9e0a128b-eaff-4944-bc50-22bcb60d338a

# After: 

(tested using shift-enter then enter in both editors)

https://github.com/user-attachments/assets/5ebc1530-bbb2-458c-96d8-6fdc652f7c7d

### DIFF
--- a/assets/keymaps/linux/jetbrains.json
+++ b/assets/keymaps/linux/jetbrains.json
@@ -29,7 +29,6 @@
       "ctrl-pageup": "editor::MovePageUp",
       // "ctrl-alt-shift-b": "editor::SelectToPreviousWordStart",
       "ctrl-alt-enter": "editor::NewlineAbove",
-      "shift-enter": "editor::NewlineBelow",
       // "ctrl--": "editor::Fold", // TODO: `ctrl-numpad--` (numpad not implemented)
       // "ctrl-+": "editor::UnfoldLines", // TODO: `ctrl-numpad+` (numpad not implemented)
       "alt-shift-g": "editor::SplitSelectionIntoLines",
@@ -68,6 +67,7 @@
   {
     "context": "Editor && mode == full",
     "bindings": {
+      "shift-enter": "editor::NewlineBelow",
       "ctrl-f12": "outline::Toggle",
       "ctrl-r": ["buffer_search::Deploy", { "replace_enabled": true }],
       "ctrl-e": "file_finder::Toggle",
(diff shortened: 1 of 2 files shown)


assets/keymaps/linux/jetbrains.json
```diff
@@ -29,7 +29,6 @@
       "ctrl-pageup": "editor::MovePageUp",
       // "ctrl-alt-shift-b": "editor::SelectToPreviousWordStart",
       "ctrl-alt-enter": "editor::NewlineAbove",
```
Veykril, line 31: Shouldn't we also move this as well? I do not know the jetbrains keymap but I doubt the two related actions being split makes sense?

mason-palmer: Honestly, I think it could go either way. The reason I only moved shift-enter is that the default keymap explicitly binds it to Newline in auto_height mode to preserve the standard chat input behavior, so JetBrains was       
  overriding that. cmd-alt-enter doesn't have an equivalent default binding, so there's no conflict there currently. That said, moving both for consistency makes sense too. What do you think?

Veykril: I see, thanks for the input, I think https://github.com/zed-industries/zed/pull/47595 might be the more appropriate way to fix this then, thoughts? (assuming the PR will also change this for the linux keymap then which it right now does not)

PR #47435, opened 2026-01-22
Improve TypeScript keyword syntax highlighting

ahcurrier:
Improve syntax highlighting of keywords by creating groups for declarations and import/export.

### Release Notes:
- Highlight declarations with `keyword.declaration`:
  - const
  - let
  - var
  - function
  - class
  - enum
  - interface
  - type

- Highlight imports and exports with `keyword.import`:
  - export
  - from
  - import

### DIFF
--- a/crates/languages/src/javascript/highlights.scm
+++ b/crates/languages/src/javascript/highlights.scm
@@ -246,20 +246,13 @@
   "as"
   "async"
   "await"
-  "class"
-  "const"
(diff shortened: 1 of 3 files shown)


ahcurrier:
@cla-bot check

compeso-mgruenwald:
Same goes for `TSX`. Seems like this only works reliably in plain `TypeScript`

PR #47437, opened 2026-01-22
git_ui: Automatically load more commits on file history

dinocosta:
Update the `FileHistoryView` implementation so as to automatically load more commits when the user scrolls to within 5 items of the bottom of the list.

Release Notes:

- Added automatic loading of more commits to the `git: file history` view when scrolling near the bottom of the list


Anthony-Eid:
Make me wonder if we could use the git graph someone for the file history view

PR #47447, opened 2026-01-23
languages: Fix C++ indentation for multi-line declarations and assignments

ozacod:
Before:
indentation-problem

After:
after

Release Notes:

- Fixed indentation problem with multi-line declarations and assignments in C++


smitbarmase (changes requested):
Thanks for the PR! I think this can be simplified. Just adding `(init_declarator)` to the existing `@indent` list is sufficient, no need to restructure the `assignment_expression` rule:

```diff
 [
     (field_expression)
     (assignment_expression)
+    (init_declarator)
     (if_statement)
     (for_statement)
     (while_statement)
     (do_statement)
     (else_clause)
 ] @indent
-
-(declaration
-    declarator: (init_declarator) @indent
-    ";" @end) @indent
-
-(expression_statement
-    (assignment_expression) @indent
-    ";" @end) @indent
```

Could you also apply the same fix to `crates/languages/src/c/indents.scm`? It has identical rules.

smitbarmase (approved):
Thanks!

PR #47452, opened 2026-01-23
git_ui: Add 'Add to .git/info/exclude' action to git panel

dastrobu:
This adds a new context menu option to add untracked files to .git/info/exclude, which works like .gitignore but is local-only and not committed to the repository.

Implementation:
- Added AddToGitExclude action to git crate
- Refactored common logic into add_entry_to_ignore_file helper
- Both add_to_gitignore and add_to_git_exclude use the shared helper
- Added context menu entry after 'Add to .gitignore'
- Only enabled for untracked (created) files

This allows users to ignore files locally without modifying the shared .gitignore file, which is useful for personal development preferences.

Closes  #29507
  
Release Notes:

- Added: Git panel now has an 'Add to .git/info/exclude' option for untracked files, allowing local-only ignoring without modifying .gitignore

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7282,6 +7282,7 @@ dependencies = [
  "watch",
  "windows 0.61.3",
  "workspace",
+ "worktree",
  "zed_actions",
  "zeroize",
  "zlog",
--- a/crates/git/src/git.rs
+++ b/crates/git/src/git.rs
@@ -101,6 +101,8 @@ actions!(
         Clone,
         /// Adds a file to .gitignore.
         AddToGitignore,
+        /// Adds a file to .git/info/exclude.
+        AddToGitExclude,
     ]
 );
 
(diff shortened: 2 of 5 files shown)


cole-miller:
Thanks for opening a PR--however, compared to `.gitignore` I think this is a bit too niche for us to include in the UI.

dastrobu:
@cole-miller Thanks for reviewing the PR. I respectfully disagree with the "too niche" assessment.

I'd suggest continuing the discussion in [#29507](https://github.com/zed-industries/zed/discussions/29507), where I've left a detailed comment with additional context explaining why this is actually a standard Git workflow feature (built into JetBrains IDEs, 10k+ VSCode extension installs, officially documented in Git docs, etc.).

I'd prefer to keep the discussion there to avoid commenting on a closed PR, and because there's already an active community request for this exact feature.

PR #22644, opened 2025-01-04
Delay hiding git blame tooltip

mgsloan:
It's easy to overshoot the bottom of the tooltip when cursoring to a button, such as opening the commit from a blame tooltip. Before this change the tooltip would immediately disappear, and now it sticks around for a bit.

Also:

* Shares the implementation with `elements/text.rs`. This will particularly be handy when it makes use of hoverable tooltips.

* Improves the fix to #21657.

    - Now the element will no longer think it has an active tooltip that it registers with the window.

    - It will instead display the next available tooltip, whereas I believe before the next available tooltip would be suppressed.

* Fixes bug where `cx.refresh()` wasn't called when text tooltip is hidden due to a mouse down event.

* Ports over fix in https://github.com/zed-industries/zed/pull/14832 to `elements/text.rs`

Release Notes:

- The tooltip for inline git blame now waits a bit before disappearing when the mouse leaves it.

PR #47471, opened 2026-01-23
editor: Persist multi-line diagnostic hovers in whitespace areas

### LINKED ISSUE #46841: Diagnostic hover popover disappears when moving mouse cursor to the popover
### Reproduction steps

1. Add this to your `pyproject.toml`:

```toml
[tool.ruff.lint]
extend-select = ["I"]
```

2. Save this Python snippet to a file:

```python
#! /usr/bin/env -S uv run --script
#
# /// script
# dependencies = []
# ///

from datetime import UTC
import datetime as dt
from typing import TYPE_CHECKING, Any, Self

print(dt, UTC, TYPE_CHECKING, Any, Self)
```

3. Hover the mouse cursor over the end of line 9. The diagnostic hover popup appears above_line 7.
4. Move your mouse cursor to the hover popup.

Ruff warns about [unsorted-imports](https://docs.astral.sh/ruff/rules/unsorted-imports/) on lines 7-9. Line 9 and the hover popup are both long. But lines 7-8 are short, so there is some empty space, which Zed doesn't consider to be part of the warning.

If you put the mouse cursor near the end of line 9 to see the hover popup, you will probably move it straight upward to the hover popup, which causes the popup to disappear.

https://github.com/user-attachments/assets/310e0dbf-6282-4f7b-bcf9-8ca7486c7baa

Zed: v0.219.4+stable.119.5595f7ce23d51a58eb66dd025a487bbd6c75dad1 (Zed) 
OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

daydalek:
When the mouse cursor moves into the whitespace of a line within a multi-line diagnostic range, the hover popover would previously disappear. This change adds a check to keep the diagnostic hover visible if the mouse row intersects with the active diagnostic's range.

Fixes #46841

Release Notes:

- Improved hover behavior for multi-line diagnostics to persist when hovering over whitespace.

https://github.com/user-attachments/assets/0965cb25-6207-4d4a-9165-0d51157fc6e4

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -8360,6 +8360,7 @@ impl Editor {
 
         self.update_hovered_link(
             position_map.point_for_position(mouse_position),
+            Some(mouse_position),
             &position_map.snapshot,
             modifiers,
             window,
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -1458,6 +1458,7 @@ impl EditorElement {
         if text_hovered {
             editor.update_hovered_link(
                 point_for_position,
+                Some(event.position),
                 &position_map.snapshot,
                 modifiers,
                 window,
@@ -1469,12 +1470,13 @@ impl EditorElement {
                     .snapshot
                     .buffer_snapshot()
                     .anchor_before(point.to_offset(&position_map.snapshot, Bias::Left));
-                hover_at(editor, Some(anchor), window, cx);
+                hover_at(editor, Some(anchor), Some(event.position), window, cx);
                 Self::update_visible_cursor(editor, point, position_map, window, cx);
             } else {
                 editor.update_inlay_link_and_hover_points(
                     &position_map.snapshot,
                     point_for_position,
+                    Some(event.position),
                     modifiers.secondary(),
                     modifiers.shift,
                     window,
@@ -1483,7 +1485,7 @@ impl EditorElement {
             }
         } else {
             editor.hide_hovered_link(cx);
-            hover_at(editor, None, window, cx);
+            hover_at(editor, None, Some(event.position), window, cx);
         }
     }
 
--- a/crates/editor/src/hover_links.rs
+++ b/crates/editor/src/hover_links.rs
@@ -4,7 +4,7 @@ use crate::{
     HighlightKey, Navigated, PointForPosition, SelectPhase,
     editor_settings::GoToDefinitionFallback, scroll::ScrollAmount,
 };
-use gpui::{App, AsyncWindowContext, Context, Entity, Modifiers, Task, Window, px};
+use gpui::{App, AsyncWindowContext, Context, Entity, Modifiers, Pixels, Task, Window, px};
 use language::{Bias, ToOffset};
 use linkify::{LinkFinder, LinkKind};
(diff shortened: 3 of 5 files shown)


daydalek:
>  a possible problem with this is how it interacts with inactive regions in clangd which are implemented using diagnostics. since its between `#ifdef`s the diagnostic can often be hundreds or even thousands of lines which means the hover will basically never go away since the whole screen is a diagnostic

True that. And I think maybe a special judge for conditional compilation macros or hardcoded line-limit wouldn't be very elegant and would likely lead to more edge cases.I've been looking into how VS Code handles this, and their approach is quite clever. They focus on the user's movement intent using a distance-tracking mechanism. Basically, once the hover appears, the editor starts tracking the Euclidean distance between the mouse and the hover widget itself. As long as the mouse is "getting closer" to the widget (within a tiny 4px tolerance), the hover stays alive.

I think this intent-based logic is much more robust than simple range checks. Still mulling over the best way to fit this into Zed's model. And It seems some work to do, I'll try to figure it out.

Ahmad-Alsaleh:
> > I wonder if an alternative approach could be something like #22644 ? Surprised to see that not all popovers are delayed by default — and wondering if this solves most of the issues like #46841
> 
> Thanks for pointing that out! I think a delay is a viable alternative. My only concern, echoing @xdBronch's point, is that a fixed duration might feel brittle for large diagnostic ranges. If we decide to pursue this path, we should probably explore an adaptive delay based on the block's size to make the solution more robust.

@daydalek how about displaying the popover near the user's cursor instead of at the beginning of the range? It might not be the best user experience if someone hovers over a line and then has to scroll up to read the error/warning message. What do you think?

I made PR #49257 implementing that, feel free to let me know your thoughts.

SomeoneToIgnore (approved):
Sorry for the pause, revisiting it now, I think this is quite nicely blends both the timeout idea proposed and the heuristics on top.

* One wish would be to have settings like

Image

in Zed's settings (+ an import in `vscode_import.rs`), smth. like `hover_popover_hide_delay`?

that apply instead of the hardcoded `let delay = 300u64` one.

* I wonder if the test is possible somehow to create for this, but not sure we have the ability to move mouse over random parts, hence not suggesting anything.

----------------------

I'll merge the PR after Wednesday's release to have more time to test it internally, but overall things worked way better than before, thank you for the fix!

daydalek:
> Sorry for the pause, revisiting it now, I think this is quite nicely blends both the timeout idea proposed and the heuristics on top.
> 
> * One wish would be to have settings like
> 
> Image
> in Zed's settings (+ an import in `vscode_import.rs`), smth. like `hover_popover_hide_delay`?
> 
> that apply instead of the hardcoded `let delay = 300u64` one.
> 
> * I wonder if the test is possible somehow to create for this, but not sure we have the ability to move mouse over random parts, hence not suggesting anything.
> 
> I'll merge the PR after Wednesday's release to have more time to test it internally, but overall things worked way better than before, thank you for the fix!

Thanks for the review and the suggestion!

Making the delay configurable definitely makes sense. I’m currently preparing for my graduate school entrance interview, so my time is a bit limited these weeks. I’ll look into this improvement in April and follow up with a separate PR if needed.

Thanks again for the testing and feedback!

PR #47472, opened 2026-01-23
Show the model provider in the favorited model list

zapp88:
When a user selects a model as favorite, the model name displays correctly, but the provider information is missing.
For example, a favorited "Gemini 3 Flash" model shows no indication of the provider (e.g., OpenRouter, Google API, GitHub Copilot). While the provider is clearly shown at the top of each model list, this context is lost once the model is favorited.

obraz

Proposed solution would be to display the provider name when hovering over favorite model like below:

obraz

This PR provides said modification.

Release Notes:

- Adds model provider name visibility when hovering over favorite model.

### DIFF
--- a/crates/acp_thread/src/connection.rs
+++ b/crates/acp_thread/src/connection.rs
@@ -350,6 +350,7 @@ pub enum AgentModelIcon {
 pub struct AgentModelInfo {
     pub id: acp::ModelId,
     pub name: SharedString,
+    pub provider_name: Option<SharedString>,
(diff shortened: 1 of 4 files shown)


danilo-leal (commented):
Heya, thanks for the PR! I don't we want to include the provider name in there because this would clutter the list item unnecessarily given we already have the icon to communicate from which provider a given model is from. However, I might be open to add that information in a docs aside. If you want to explore this option, let me know! Otherwise, I'll unfortunately close this issue!

PR #47505, opened 2026-01-23
editor: Fix bracket color change when system theme changes

smitchaudhary:
Closes #47503 

I added an observer for `GlobalTheme` changes that check if the accent colors have changed or not. I thought this was closest to the patterns and style in the codebase. I also considered triggering a `SettingsStore` notification on theme reload but that seemed incorrect since the settings didn't really change, even if it would solve the problem (as the editor already observes `SettingsStore` to detect theme changes.

Release Notes:

- Fixed Bracket color not updating when system theme changes


SomeoneToIgnore (approved):
Thanks!

PR #47506, opened 2026-01-23
title_bar: Respect Linux titlebar config

mufeedali:
Currently, Zed always places three fixed window buttons (Minimize, Maximize and Close) on the right side of the window in a fixed order ignoring any user configuration or desktop environment preference (like elementary).

This PR adds support for GNOME-style layout strings (`minimize:close`). By default, we pull it from the gsettings portal, but we also allow manual configuration via `title_bar.button_layout` config key.

image

Closes #46512

I know it's a relatively large PR for my first one and I'm new to Rust. So, sorry if I've made any huge mistakes. I had just made it for personal use and then decided to try to clean it up and submit it.

I've tested with different configs on Linux. Untested on other platforms, but should have no impact.

If it's not up to par, it's okay, feel free to close :)

Release Notes:

- Added support for GNOME's window buttons configuration on Linux.


crates/settings_content/src/settings_content.rs
```diff
@@ -351,6 +351,15 @@ pub struct TitleBarSettingsContent {
...
     pub show_menus: Option<bool>,
+    /// The layout of window control buttons in the title bar (Linux only).
+    ///
+    /// This can be set to "auto" to follow the system configuration,
+    /// or a custom string in the format "left:right".
+    ///
+    /// Example: "close:minimize,maximize"
+    ///
+    /// Default: "auto"
+    pub button_layout: Option<String>,
```
smitbarmase: How about instead of this, we turn it into an Enum?

```rs
 pub enum WindowButtonLayoutContent {
      #[default]
      Auto,
      Default,
      Custom(WindowButtonLayout),
  }
```

crates/title_bar/src/title_bar_settings.rs
```diff
@@ -24,6 +26,10 @@ impl Settings for TitleBarSettings {
             show_sign_in: content.show_sign_in.unwrap(),
             show_user_menu: content.show_user_menu.unwrap(),
             show_menus: content.show_menus.unwrap(),
+            button_layout: content
```
smitbarmase: And then you can implement `Deserialize` for that enum we introduced, so here instead of this, we can simply have:

```rs
button_layout: content.button_layout.unwrap().into_layout()
```

where `into_layout` can contain `WindowButtonLayoutContent` to `WindowButtonLayout` mapping.

Something like:

```rs
pub fn into_layout(self) -> Option<WindowButtonLayout> {
          match self {
              Self::Auto => None,
              Self::Custom(layout) => Some(layout),
          }
      }
```

Note: I haven't added `Default` variant here, but I think we should do that to preserve current Zed behavior, for users who don't want Gnome configuration, and also don't really want to specify string manually to opt out.

smitbarmase (changes requested):
Thanks! I have added few suggestions regarding how we are handling settings.

smitbarmase (approved):
Thanks! I have made few API changes to make it more generic.

PR #47517, opened 2026-01-23
git: Remove --follow from file history pagination

dinocosta:
The `--follow` flag in `git log` causes `--skip` to count all commits walked in the graph, not just commits that touch the target file. This results in incorrect pagination when other commits are interleaved in the repository history, causing duplicate entries or missing commits.

This was discovered because `git log --follow --skip N` would return duplicates when paginating file history for some files in Zed's repository, for example, `crates/vim/src/normal/paste.rs`.

The tradeoff is that file history will no longer track files across renames but, right now, having correct pagination feels more important for the file history view's functionality.

I believe we can eventually continue using `--follow`, as long as we re-implement the file history pagination to rely on the last commit's hash to find older commits, instead of relying in the `--skip` flag.

Closes #47487 

Release Notes:

- Fixed issue with `git: file history` where duplicated entries could be returned and the wrong number of commits loaded


GoldStrikeArch:
@dinocosta wdyt about the approach I did in [this one](https://github.com/zed-industries/zed/pull/47551) which is a "client side filtering"? IMO the case when user will click more than a few times on "load more" is a very extreme and rare (why would you do that?), so it should work fine.

PR #47522, opened 2026-01-23
ztracing: Annotate more callsites that invoke TreeSitter

kubkon:
Release Notes:

- N/A


crates/ztracing/src/lib.rs
```diff
@@ -4,13 +4,106 @@ pub use tracing::{Level, field};
...
+#[cfg(all(ztracing, ztracing_with_memory))]
+unsafe impl Send for Allocation {}
+#[cfg(all(ztracing, ztracing_with_memory))]
+unsafe impl Sync for Allocation {}
+
+#[cfg(all(ztracing, ztracing_with_memory))]
+#[derive(Default)]
+struct AllocRecorder {
+    outstanding: std::sync::Mutex<std::collections::HashMap<Allocation, std::alloc::Layout>>,
+}
```
maxbrunsfeld: Why have this allocation recorder? Does this affect how tracy treats the allocations?

kubkon: Initially Cole and I tried simply wrapping `libc` system alloc in tracy bookkeeping calls. However, this led to a weird problem of tracy crashing with supposed double malloc reported for the same address. It seems to have originated from the tracy-wrapped global Rust system allocator. So to circumvent this I figured that using Rust's global alloc (which is already wrapping tracy bookkeeping calls) ought to do it. Now, to do that, we need to track memory layout between the calls since otherwise Rust's allocator may panic if unenforced (not sure that's true actually judging by the implementation in Rust's libstd). So that's what the recorder is for.

kubkon: Thinking about this some more, I am going to hold off merging this until I re-confirm that setting the allocator for TreeSitter is actually necessary at all. Thanks for catching this @maxbrunsfeld!

kubkon: Sorry it took me this long to re-investigate this - good news, locally on Linux I don't need to overwrite TreeSitter's alloc hooks and still get tracy traces. This would also explain why I was seeing double alloc errors in tracy when using a different annotated allocator before. tl;dr, lemme remove all of this tracking code as it is clearly not needed to get valid alloc traces in tracy.

PR #47533, opened 2026-01-24
repl: Add quality of life changes in Jupyter view

MostlyKIGuess:
- Keyboard navigation where you can traverse through cells using up and down arrow
- Jupyter Logo added 
- Initialize kernel as shutdown for more predictable behavior
- Ability to create .ipynb files with bare essential metadata.
- Optimize editor initialization to avoid cloning the entire notebook and shortcuts 

Release Notes:

- N/A


rgbkrk (approved):

maxdeviant:
> Since notebook is still behind a feature flag and has much to finish, we probably don't want release notes for this right @maxdeviant?

Yea, if it's still feature-flagged then we should "- N/A" the release notes.

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -243,7 +245,7 @@ impl NotebookEditor {
             cell_order: cell_order.clone(),
             original_cell_order: cell_order.clone(),
             cell_map: cell_map.clone(),
-            kernel: Kernel::StartingKernel(Task::ready(()).shared()),
+            kernel: Kernel::Shutdown, // TODO: use recommended kernel after the implementation is done in repl
```
rgbkrk, line 243: Makes sense for now. I trust you'll be hacking on that anyways as you're moving along with #47891.

rgbkrk (approved):
Looking good, especially since we're feature flagged anyhow.

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -1628,6 +1693,20 @@ impl KernelSession for NotebookEditor {
...
+            if let Ok(language_info) = serde_json::from_value::<nbformat::v4::LanguageInfo>(
+                serde_json::to_value(&reply.language_info).unwrap(),
+            ) {
+                self.notebook_item.update(cx, |item, cx| {
+                    item.notebook.metadata.language_info = Some(language_info);
+                    cx.emit(());
+                });
+            }
+            cx.notify();
+        }
```
rgbkrk, line 1708: 👍 

There's more metadata I think we're supposed to set to have parity but I can't recall at this moment. Good to push forward with for the moment though.

MostlyKIGuess: Yeah we need to add signature for security.. and later on other optional metadata as well like author, title etc.

rgbkrk: The security one is kind of interesting since Jupyter devs put that in place for html outputs, which Zed/GPUI doesn't support and if we did they'd be in isolated web views anyhow.

rgbkrk (approved):

PR #47542, opened 2026-01-24
git: Add commit hooks toggle

tdgroot:
With the Hooks toggle, found in both the commit modal and panel, users can choose to commit their changes without running the pre-commit hook.

The hooks option is enabled by default, which is what you want out of the box.

Closes #47124

Release Notes:

- Added toggle for to enable/disable git commit hooks in commit panel and modal.

**Design questions**

- When committing with hooks disabled, should it be enabled back after commit?
- Any suggestion for better naming of the action which I for now called `git::Hooks`?

**Screenshots**

Hooks option in commit panel
image

Hooks option in commit modal
image

### DIFF
--- a/crates/git/src/git.rs
+++ b/crates/git/src/git.rs
@@ -87,6 +87,8 @@ actions!(
         Amend,
         /// Enable the --signoff option.
         Signoff,
+        /// Enable the commit hooks option.
+        Hooks,
         /// Cancels the current git operation.
         Cancel,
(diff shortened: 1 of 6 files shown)


cole-miller:
Thanks for opening a PR. I unfortunately missed that this was open when I approved #61529, so we now have UI for `--no-verify` on main. I'll close this, but we'll add an acknowledgement of your contribution to our release notes for this feature.

PR #47556, opened 2026-01-24
languages: Add highlighting for document comments in C/C++

ozacod:
Before:
before

After:
after

Release Notes:

- Added highlighting for document comments in C/C++


MrSubidubi (changes requested):
Should this not be using doxygen as an injection language instead? https://github.com/zed-industries/zed/pull/43581

MrSubidubi (approved):

osiewicz (approved):

PR #47557, opened 2026-01-24
copilot_chat: Return true context window size

### LINKED ISSUE #44909: Incorrectly reported token limits in agent panel on many GitHub Copilot models
### Reproduction steps

Creating a new issue following up on #44632.

Token size limits are often misreported in GitHub Copilot models. The marked up screenshot should be enough context to understand the issue.

Image

Are here is where it finally errors. The error log is included below.

Image

In this case, using Claude Open 4.5 via GitHub Copilot, the tokens go way over the stated limit.

It's my believe that there are actually two issues here:

1. The token counter doesn't match what copilot / models use.
2. Zed is not looking up what GitHub reports the context size to be.

In the case of the screenshots above with Opus 4.5, given that the error says the context size is 128k, I suspect the error is related to incorrect token counting.

I have also seen either ChatGPT 5 or Sonnet 4.5 go well above 200k tokens to around 265k tokens used reported in the Zed agent panel before erroring. In that case, I believe it is the second issue. The agent panel always shows a 128k token limit for GitHub Copilot models.

Luckily, continuing to use the model beyond the Zed reported token limit does work, so this is not as severe of an issue.

Unfortunately, it doesn't look like GitHub Copilot publishes any up-to-date information on it's context size limits.

I expect the token counter in the agent panel to report accurate values.

OS: macOS 15.6
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

```log
2025-12-15T14:34:02-05:00 ERROR [agent] Error in model response stream: Failed to connect to API: 400 Bad Request {"error":{"message":"prompt token count of 131835 exceeds the limit of 128000","code":"model_max_prompt_tokens_exceeded"}}
```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

Claude Opus 4.5 via GitHub Copilot

The scope of issue may exist on all models served via GitHub Copilot

None

anilpai:
## Fix incorrect context size limits for GitHub Copilot Chat models

Fixes #44909

### Problem

The agent panel was displaying incorrect token limits for GitHub Copilot models. Users reported that:
- The agent panel always showed a **128K token limit** for all GitHub Copilot models, regardless of their actual context window size
- Claude models (e.g., Claude 3.7 Sonnet, Claude Opus 4.5) were showing ~90K instead of their actual 200K context window
- GPT-4o was showing 110K instead of its actual 128K context window
- Users could continue using models beyond the displayed limit, which worked but was confusing

### Root Cause

The `max_token_count()` method in `copilot_chat.rs` was returning `max_prompt_tokens` instead of `max_context_window_tokens`:

```rust
// Before (incorrect)
pub fn max_token_count(&self) -> u64 {
    self.capabilities.limits.max_prompt_tokens
}
```

GitHub's API returns three different token-related fields:
- `max_context_window_tokens`: The **full context window size** (e.g., 200K for Claude 3.7)
- `max_prompt_tokens`: GitHub's limit for prompt input (e.g., 90K for Claude 3.7)
- `max_output_tokens`: Maximum output tokens (e.g., 16K)

The `max_token_count()` method in the `LanguageModel` trait is expected to return the **full context window size** — this is consistent with all other providers (Anthropic returns 200K for Claude, OpenAI returns 128K for GPT-4o, etc.).

### Solution

Screenshot 2026-01-25 at 1 07 53 AM

Changed `max_token_count()` to return `max_context_window_tokens`:

```rust
// After (correct)
pub fn max_token_count(&self) -> u64 {
    self.capabilities.limits.max_context_window_tokens as u64
}
```

### Impact

| Model | Before | After |
|-------|--------|-------|
| Claude 3.7 Sonnet | 90,000 | **200,000** |
| Claude Opus 4.5 | 90,000 | **200,000** |
| GPT-4o | 110,000 | **128,000** |

### Testing

Added a new test `test_max_token_count_returns_context_window_not_prompt_tokens` that:

1. Deserializes model JSON with distinct `max_context_window_tokens` and `max_prompt_tokens` values
2. Verifies Claude 3.7 Sonnet returns 200,000 (context window), not 90,000 (prompt tokens)
3. Verifies GPT-4o returns 128,000 (context window), not 110,000 (prompt tokens)

All existing tests continue to pass:
```
running 4 tests
test tests::test_unknown_vendor_resilience ... ok
test tests::test_max_token_count_returns_context_window_not_prompt_tokens ... ok
test tests::test_resilient_model_schema_deserialize ... ok
test result: ok. 4 passed; 0 failed
```

Release Notes:

- copilot: Fixed incorrect context window size displayed for GitHub Copilot Chat models in the agent panel.


anilpai:
@yeskunall Added Release Notes.

crates/copilot_chat/src/copilot_chat.rs
```diff
@@ -1014,6 +1014,11 @@ mod tests {
     }
 
     #[test]
+    fn test_max_token_count_returns_context_window_not_prompt_tokens() {
```
benbrandt, line 1017: this seems like invalid syntax

anilpai: Thanks. Just fixed it.

PR #47571, opened 2026-01-25
editor: Add select to {start,end} of larger syntax node commands

alexispurslane:
I'm used to having the Emacs mark-sexp command, which selects/extends the selection to the start or end of a sexp/larger syntax node, so when I found out the `move to {start,end} of larger syntax node` commands were added, these seemed like a natural addition; especially since most other commands in Zed seem to have movement and selection pairs, but those didn't really feel like they did.

I did a really *lot* of tests (brainstormed ideas with GLM 4.7, then wrote the actual test case strings myself, then had it convert them to Rust, then made sure the tests made sense and passed), I'm not sure if it's too much or too little, but I looked at the PR that added the `move to` versions of these commands and it was also +~550 lines, so this seemed in the ballpark 😅 

I factored out the core syntax node finding logic from those commands into a common function, so that both sets of commands could use it, just with different code for modifying the editor state wrapped around — mine just moves each selection's head, instead of totally resetting it.

Release Notes:

- Added commands for extending selections to syntax node boundaries. i.e. `editor: select to start of larger syntax node` and `editor: select to end of larger syntax node`.

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -737,6 +737,10 @@ actions!(
         SelectDown,
         /// Selects the enclosing symbol.
         SelectEnclosingSymbol,
+        /// Selects to the start of the next larger syntax node.
+        SelectToStartOfLargerSyntaxNode,
+        /// Selects to the end of the next larger syntax node.
+        SelectToEndOfLargerSyntaxNode,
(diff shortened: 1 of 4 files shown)


smitbarmase (approved):
Thank you!

PR #47578, opened 2026-01-25
Ensure array tool schemas include items for OpenAI

cacaosteve:
### Summary

Fixes an OpenAI `invalid_function_parameters` error caused by tool/function schemas that define `"type": "array"` without an `items` field.

OpenAI’s schema validator requires all array schemas to explicitly define `items`, even though this is optional in standard JSON Schema.

### Changes

- Ensure array schemas in the OpenAI JSON Schema path include an `items` field
- Normalize boolean `items` schemas (`true` / `false`) to `{}` for OpenAI compatibility
- Leave `JsonSchemaSubset` adaptation unchanged for non-OpenAI providers
- Add tests covering both OpenAI preprocessing and `JsonSchemaSubset` behavior

### Result

Tool/function schemas generated for OpenAI are now compatible with OpenAI’s stricter validation rules without broadening the change to other providers.

---

Closes #47576

Release Notes:
- Fixed OpenAI tool/function schema validation errors caused by array parameters missing an `items` definition

### DIFF
--- a/crates/language_model/src/tool_schema.rs
+++ b/crates/language_model/src/tool_schema.rs
@@ -85,6 +85,36 @@ fn preprocess_json_schema(json: &mut Value) -> Result<()> {
             obj.insert("properties".to_string(), Value::Object(Default::default()));
         }
     }
+
+    // OpenAI API requires non-missing `items` for arrays
+    if let Value::Object(obj) = json
+        && matches!(obj.get("type"), Some(Value::String(s)) if s == "array")
+    {
+        if !obj.contains_key("items") {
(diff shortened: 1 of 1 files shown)


benbrandt (changes requested):
Hi usually for these sorts of schema changes we only apply them when necessary (see what we do for google provider)

cacaosteve:
Updated this to only apply the array `items` insertion/normalization on the OpenAI JSON Schema path, rather than in the broader `JsonSchemaSubset` adaptation. I also updated the tests accordingly.

Issue #39211, opened 2025-09-30
AI: Mistral models unable to use `edit_file`


EpigeneMax:
### Summary

Mistral models fail to use `edit_file` tool from the Agent Panel (Inline Assist works fine).

### Description

Steps to trigger the problem:
1. In a new directory, open Zed
2. In the agent panel, prompt "create a Hello world C file" to codestral-latest
3. 

**Expected Behavior**:
agent creates and writes code to a C file
**Actual Behavior**:
agent creates an empty C file and answers with "I'm sorry, but I'm unable to assist with that request at the moment. If you have any other questions or need help with something else, please let me know!"

### Model Provider Details
- Provider: Mistral
- Model Name: codestral-latest, devstral-medium-latest, devstral-small-latest
- Mode: Agent Panel
- Other Details (MCPs, other settings, etc): all other tools work fine (tested `diagnostics`, `list_directory`, `now`, `read_file`, `terminal`). Unlikely to be a permission issues, since the file is created (although empty).
OpenAI models work flawlessly, haven't tried other providers.

For what it's worth, here the markdown version of the conversation:

    ## User

    create a "Hello world" C file

    ## Assistant

    **Tool Call: hello.c**
    Status: Failed

    Diff: hello.c
    ```

    ```

    ## Assistant

    **Tool Call: hello.c**
    Status: Failed

    Diff: hello.c
    ```

    ```

    ## Assistant

    I'm sorry, but I'm unable to assist with that request at the moment. If you have any other questions or need help with something else, please let me know!

### Zed Version and System Specs

Zed: v0.205.8 (Zed) 
OS: macOS 15.4
Memory: 16 GiB
Architecture: aarch64

silent-sixth-cargo:
likely model's cant comprehend / in compatible , 

Not a zed issue .
No changes shall be made as it might affect perfectly functional for advance model like claude .

silent-sixth-cargo:
use quality model.

MrSubidubi:
Thanks for the report! This seems to be very similar to #39031, could you check? As you mentioned, that seems to be an issue specific with mistral models and the other issue also mentiones basically the same issue.

EpigeneMax:
> Thanks for the report! This seems to be very similar to https://github.com/zed-industries/zed/issues/39031, could you check? As you mentioned, that seems to be an issue specific with mistral models and the other issue also mentiones basically the same issue.

[...]

Issue #39031, opened 2025-09-27
AI: Agent & tools just stop working no matter the model/provider


kinoute:
### Summary

The Agent panel works for a few time then it either goes in an infinite loop doing something, either it says it will do something (like listing or searching) and it doesnt do it. The "loader" disappears after a few seconds just like that, without any errors.

### Description

Steps to trigger the problem:
1. Run Zed
2. Open Agent Panel, ask something to do
3. The Agent sometimes starts answering but it might have to use one tool, it just stops (see video).

It seems to have getting worse in the latest versions. Earlier (one week or two weeks ago?) I got the infinite loop thing a lot but not this "no action done" thing.

**Expected Behavior**: The agent does its thing
**Actual Behavior**: The agent stops just like it thinks it has done its job, no matter the model picked. I have no problem running VSCode+Cline with these models.

https://github.com/user-attachments/assets/42d3b891-494d-40a0-a540-c48ea014d1fa

### Model Provider Details
- Provider: Mistral API or custom OpenAI provider (with Mistral Small behind)
- Model Name: Codestral, devestral, magistral...
- Mode: Agent Panel
- Other Details (MCPs, other settings, etc):

Zed Settings:

```json
// Zed settings
//
// For information on how to configure Zed, see the Zed
// documentation: https://zed.dev/docs/configuring-zed
//
// To see all of Zed's default settings without changing your
// custom settings, run `zed: open default settings` from the
// command palette (cmd-shift-p / ctrl-shift-p)
{
  "context_servers": {
    "mcp-server-gitlab": {
      "source": "extension",
      "enabled": true,
      "settings": {
        "gitlab_personal_access_token": "xxx",
        "gitlab_api_url": "xxx"
      }
    },
    "mcp-server-git": {
      "source": "custom",
      "enabled": true,
      "command": "uvx",
      "args": ["mcp-server-git"]
    }
  },
  "icon_theme": "Material Icon Theme",
  "journal": {
    "hour_format": "hour24"
  },
  "collaboration_panel": {
    // Whether to show the collaboration panel button in the status bar.
    "button": false
  },
  "linked_edits": false,
  "git": {
    "inline_blame": {
      "enabled": false
    }
  },
  "tabs": {
    "activate_on_close": "neighbour"
  },
  "languages": {
    "Shell Script": {
      "format_on_save": "on",
      "formatter": {
        "external": {
          "command": "shfmt",
          "arguments": [
            "--filename",
            "{buffer_path}",
[...]

Issue #37021, opened 2025-08-27
Mistral: 400 Error: "Unexpected role 'tool' after role 'user'"


notpeter:
### Summary

While using certain Mistral models via the Agent panel frequently there are errors:
> Failed to connect to Mistral API: 400 Bad Request
> ("object":"error","message":"Unexpected role 'user' after role 'tool'","type":"invalid_request_message_order","param":null,"code":"32500")

### Description

<img width="620" height="376" alt="Image" src="https://github.com/user-attachments/assets/f4496ff0-90fa-4d59-a75a-b7293994badb" />

Previously:
- https://github.com/zed-industries/zed/issues/32675

### Model Provider Details

- Provider: Mistral
- Model Name: `devstral-medium-latest`, `mistral-medium`
- Mode: Agent Panel
- Other Details: N/A

### Zed Version and System Specs

Zed: v0.200.5 (Zed)
OS: macOS 15.6.1
Memory: 32 GiB
Architecture: aarch64

vlebert:
tools call seems totally disabled on mistral models ?

<details><summary>screenshots</summary>

<img width="871" height="886" alt="Image" src="https://github.com/user-attachments/assets/fbb658e9-edf8-464c-acb9-32ec2ff128dc" />

<img width="866" height="227" alt="Image" src="https://github.com/user-attachments/assets/1b302926-afe6-4fa3-bfdd-d8fd7597f9d2" />

</details>

notpeter:
@vlebert:
1. Which models you're using?
2. Which Zed provider (Mistral, LMStudio, Ollama, etc)

vlebert:
mistral-medium-latest, same result with Mistral and Openrouter

but devstral seem to work as expected (tool).

I am almost sure mistral-medium did work before

julius-frenzel:
I have been encountering the same issue for some weeks now. Currently OpenHands is unusable with Mistral (both mistral provider and openai-compatible provider). The root cause may be in litellm though...

pleibers:
Still having this issue with most mistral models

KintsugiUwU:
Still having the same issue, version 1.0.0

dargor:
Just tried mistral small latest (alias for mistral small 4) in Zed 1.0.0 and everything seems to work as expected.

KintsugiUwU:
It happens at random. Maybe you have configured it differently? I you login with my api key from the built in llm providers setting page

dargor:
[...]

Issue #35025, opened 2025-07-24
AI: Mistral Tool not supported


confact:
### Summary

Special tool commands from Mistral seem not to be supported

An important thing to know is that I am running the latest version of Zed (0.196.5), which should have partly fixed this. I could be wrong though, but PR #34692 was part of the changelog.

### Description
When asking mistral (codestral, devstral-small) to do some lookups, it does the `thinking` and then returns this message:
```
[{"name": "find_path", "arguments": {"glob": "*.txt"}}]
```
I did get similar for `fetch` as well. 

**Expected Behavior**:
Run the tool and send to the model

**Actual Behavior**:
prints out the json array and just stops.

### Model Provider Details
- Provider: Mistral
- Model Name: codestral, devstral-small-latest (tried mistral-medium put it doesn't want to do any commands)
- Mode: Agent Panel
- Other Details (MCPs, other settings, etc):

### Zed Version and System Specs

Zed 0.196.5 
cc82f1eacd880a8730db24c448e922898601c849

wmnnd:
This issue seems to persist in 0.211.6. I tried creating a simple file with both Devstral and Codestral via the built-in Mistral integration:

<img width="1002" height="693" alt="Image" src="https://github.com/user-attachments/assets/9449ec3b-df32-4eb7-bb34-4a985d6b4411" />

<img width="909" height="693" alt="Image" src="https://github.com/user-attachments/assets/556e678c-252c-4cb2-99c7-6f5ba9469370" />

PR #47579, opened 2026-01-25
agent: Skip empty assistant messages for Mistral provider

ian-h-chamberlain:
This small change *may* help with #39211, #35025 or #39031 although I'm not positive it will fix it entirely. At a minimum, it should prevent sending invalid requests to the Mistral API, which return a 400 error and can prevent the thread from progressing if it gets stuck in a retry loop.

This should prevent sending a message like `{ "role": "assistant", "content": "" }` if for some reason the thread contains a message with an empty text content and no tool calls.

Some users (including myself) have encountered this with the Devstral 2 models, although it doesn't seem trivial to reproduce: https://github.com/zed-industries/zed/issues/37021#issuecomment-3728902562

When it occurs, error logs look something like this:
```
2026-01-24T19:53:35-08:00 ERROR [crates/agent/src/thread.rs:984] invalid type: string "Failed to connect to Mistral API: 400 Bad Request {\"object\":\"error\",\"message\":\"Assistant message must have either content or tool_calls, but not none.\",\"type\":\"invalid_request_assistant_message\",\"param\":null,\"code\":\"3240\"}", expected struct EditFileToolOutput
2026-01-24T19:53:35-08:00 ERROR [crates/agent/src/thread.rs:984] invalid type: string "Failed to connect to Mistral API: 400 Bad Request {\"object\":\"error\",\"message\":\"Assistant message must have either content or tool_calls, but not none.\",\"type\":\"invalid_request_assistant_message\",\"param\":null,\"code\":\"3240\"}", expected struct EditFileToolOutput
2026-01-24T19:53:35-08:00 ERROR [crates/agent/src/thread.rs:984] invalid type: string "Failed to connect to Mistral API: 400 Bad Request {\"object\":\"error\",\"message\":\"Assistant message must have either content or tool_calls, but not none.\",\"type\":\"invalid_request_assistant_message\",\"param\":null,\"code\":\"3240\"}", expected struct EditFileToolOutput
```

Release Notes:

- Fixed agent sometimes sending invalid messages to Mistral API

### DIFF
--- a/crates/language_models/src/provider/mistral.rs
+++ b/crates/language_models/src/provider/mistral.rs
@@ -405,6 +405,9 @@ pub fn into_mistral(
             Role::Assistant => {
                 for content in &message.content {
                     match content {
+                        MessageContent::Text(text) if text.is_empty() => {
+                            // Mistral API returns a 400 if there's neither content nor tool_calls
+                        }
                         MessageContent::Text(text) => {
                             messages.push(mistral::RequestMessage::Assistant {
                                 content: Some(mistral::MessageContent::Plain {
@@ -850,6 +853,13 @@ mod tests {
                 LanguageModelRequestMessage {
                     role: Role::User,
                     content: vec![MessageContent::Text("Hello".into())],
+                    cache: false,
+                    reasoning_details: None,
+                },
+                // should skip empty assistant messages
+                LanguageModelRequestMessage {
+                    role: Role::Assistant,
+                    content: vec![MessageContent::Text("".into())],
                     cache: false,
                     reasoning_details: None,
                 },


vianney:
I can confirm that this PR does help with the Mistral API.

The issue appears mostly when the agent calls the `edit_file` tool without any explanation before, e.g., when editing two files one after the other. It seems closely related to the way the `edit_file` tool works: when the model answers with such a tool call, Zed makes its next request by appending the model's answer without tool call in an "agent" block, followed by a "user" block asking to provide the content. If the model's answer consisted of only the tool call, we get an empty "agent" block, which is forbidden by the Mistral API.

Note however that this patch does not fix file editing completely for Mistral models. Sometimes, the Mistral model would answer to the content request with the file content correctly enclosed with backticks, but with some extra text appended afterwards. Zed happily writes the additional text to the file instead of showing it in the agent panel. As far as I can tell, this is an unrelated bug though.

benbrandt (approved):
thanks!

PR #47586, opened 2026-01-25
build: Re-enable incremental compilation

marcocondrache:
#47358 bumped Rust to the latest stable, which includes the fix for rust-lang/cargo#16104 (rust-lang/cargo#16262). We can now re-enable incremental compilation in the codebase.

Release Notes:

- N/A


SomeoneToIgnore (approved):
🎉

PR #47616, opened 2026-01-26
Update nim entry with additional file suffixes 

ire4ever1190:
Expanded the nim entry to include `nims` and `nimble` files

- [nims](https://nim-lang.org/docs/nims.html) is a subscript of Nim that runs in a VM
- [nimble](https://github.com/nim-lang/nimble) files are project files for Nim that are written in Nim

Release Notes:

- Updated Nim file aliases so icon is shown for nims/nimble files


SomeoneToIgnore (approved):
Thank you.

PR #47618, opened 2026-01-26
extensions_ui: Add support for offline viewing of extensions

### LINKED ISSUE #12163: Cannot see list of installed extensions when offline / disconnected
### Summary

Open the Extensions panel shows nothing when offline.

Steps to reproduce:
1. Be offline
2. Open the extensions panel

Actual behavior: "No extensions installed" shown

Expected behavior: List of installed extensions is shown.

When you come back online, you need to reopen the extensions panel before it is usable.

### 

<details><summary>screenshot</summary>

Screenshot 2024-05-22 alle 23 59 18

</details>

### Environment

```
Zed: v0.136.2 (Zed Preview)
OS: macOS 14.1.1
Memory: 16 GiB
Architecture: x86_64
```

andre-eriksson:
This only shows the currently installed extensions if you are offline however, you can still uninstall extensions, search and filter via category.

Closes #12163

Release Notes:

- Added an offline view for extensions, showing your currently installed extensions when offline.

Examples:

Offline, all
offline

Offline, all with no extensions
offline-no

Online (as usual)
online

Notes:
* Not sure how "permissive" the search/filtering function should be, I used `.contains(...)` and included the option to filter via the description (I prefer this, but it can easily be changed if not desired/is inconsistent).
* I also included the button to view the repository, for an edge-case where the API is down but you have an internet connection.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5979,6 +5979,7 @@ dependencies = [
  "parking_lot",
  "pretty_assertions",
  "proto",
+ "rpc",
  "semver",
  "serde",
  "serde_json",
--- a/crates/extension/Cargo.toml
+++ b/crates/extension/Cargo.toml
@@ -26,6 +26,7 @@ log.workspace = true
 lsp.workspace = true
 parking_lot.workspace = true
 proto.workspace = true
+rpc.workspace = true
 semver.workspace = true
 serde.workspace = true
 serde_json.workspace = true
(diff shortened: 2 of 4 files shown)


maxdeviant:
Thank you for the PR!

While I do think this would be a nice thing to have, I think the implementation as-written is adding a lot of additional complexity. There's now an additional separate path for rendering extensions that will have to be kept consistent.

There probably needs to be a deeper rework of this view to make it more maintainable.

PR #47620, opened 2026-01-26
project: Add spinner while search is underway

zertosh:
When there are no search results, it's clear that a search is still
ongoing because the landing page says "Searching...". However, once
there's at least one result found, it becomes completely unclear when
search is actually finished. This adds a spinner immediately to the
right of the result counter that stops once the search is finished.

https://github.com/user-attachments/assets/a0ca4e2a-c506-42a4-bc4b-c1eb32d69e79

Release Notes:

- Improved project search loading UI.

### DIFF
--- a/crates/search/src/project_search.rs
+++ b/crates/search/src/project_search.rs
@@ -41,7 +41,7 @@ use std::{
     sync::Arc,
 };
 use ui::{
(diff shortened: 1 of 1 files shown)


danilo-leal (approved):
Thank you for this one!

PR #47621, opened 2026-01-26
languages: Add highlighting for Go directives

### LINKED ISSUE #4381: Keywords in comments highlighting for Golang
### Check for existing issues

- [X] Completed

### Describe the feature

In Golang, comments that start with certain strings may imply special usage for functions or variables.
For examples (screenshots from JetBrains GoLand):

image

image

I suppose there is more, but I'm not familiar with those.

_No response_

ozacod:
Closes #4381
Before:
before

After:
after

Release Notes:

- Added highlighting for Go directives

### DIFF
--- a/crates/languages/src/go/highlights.scm
+++ b/crates/languages/src/go/highlights.scm
@@ -138,3 +138,11 @@
 ] @constant.builtin
 
 (comment) @comment
+
+; Go directives
+((comment) @attribute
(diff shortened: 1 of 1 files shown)


ConradIrwin:
"preproc" sounds more right to me. Are there other languages with a similar feature, what do they use?.

 Does preproc look better or worse in the default themes? Happy to ship something and adjust later as needed.

PR #47631, opened 2026-01-26
project_panel: Add `bold_folder_labels` setting

zertosh:
This setting brings Zed in parity with Sublime's `bold_folder_labels`.

The settings does just that, it makes directory labels bold. This setting is
particularly useful for those who turn icons off, but do need a visual
queue (besides the chevron) to quickly tell apart files and folders.

Note: This PR depends on https://github.com/zed-industries/zed/pull/47629.
Otherwise, the setting will appear to have no uneffect (unless you're
using a custom UI font). ZedSans has "bold" today, but that's too thick
for the project panel.

zed-project-panel

Release Notes:

- Added `project_panel.bold_folder_labels` to show folder names with bold text in the project panel (defaults to `false`).

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -731,6 +731,8 @@
     // Whether to fold directories automatically and show compact folders
     // (e.g. "a/b/c" ) when a directory has only one subdirectory inside.
     "auto_fold_dirs": true,
+    // Whether to show folder names with bold text in the project panel.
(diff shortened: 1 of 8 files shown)


smitbarmase (approved):
Thanks, this looks great! It looks like we depend on https://github.com/zed-industries/zed/pull/47725, so I’ll wait for that to merge before merging this.

PR #47637, opened 2026-01-26
agent_ui: Add ability to add terminal output to agent context

rutgercap:
Summary

This PR adds the ability to quote terminal selections into Agent threads, similar to how editor selections can be quoted. Users can now select text in the terminal and add it as context to their agent conversation.

### Features

- **Context menu**: Added "Add to Agent Thread" option to the terminal's right-click menu (only shows when text is selected)
- **Keyboard shortcut**: The existing `cmd->` / `ctrl->` shortcut now works in the terminal
- **Collapsed display**: Terminal output appears as a collapsed "Terminal" crease in the chat, just like code references
- **Works with both thread types**: Supports both ACP-style threads (Claude Code) and text threads

### Implementation

- Extended `quote_selection` handler to check for terminal panel focus after checking for editor selections
- Terminal content is formatted as ` ```terminal` code blocks via new `insert_terminal_crease` method
- Refactored crease insertion into `insert_crease_impl` helper to support both code snippets (with `TextSnippet` icon) and terminal output (with `Terminal` icon)
- Added automatic detection and folding of ` ```terminal` blocks in `set_message` so terminal content stays collapsed in sent message history
- Terminal creases use a trailing space instead of newline so cursor behavior matches user expectations

### Demo

1. Select text in terminal
2. Right-click → "Add to Agent Thread" (or press `cmd->`)
3. Terminal content appears as collapsed "Terminal" block in chat
4. Send message - AI receives full terminal output, but chat history shows it collapsed

### General note

I'm new to open source and the codebase in general so let me know if anything should be different! 
I tried to stay in line with other patterns I saw.
I didn't see a way to setup an end to end test with a real terminal so I just added a unit test for the adding of the text to the chat, if it exists and I missed it then I think the test coverage could be improved with a test that uses a real terminal + running the command

image

Release Notes:

- agent: Added the ability to send selections from terminals into the Agent thread

### DIFF
--- a/crates/agent_ui/src/acp/message_editor.rs
+++ b/crates/agent_ui/src/acp/message_editor.rs
@@ -1018,7 +1018,14 @@ impl MessageEditor {
         cx: &mut Context<Self>,
     ) {
         let formatted_text = format!("```terminal\n{}\n```", text);
-        self.insert_crease_impl(formatted_text, "Terminal".to_string(), IconName::Terminal, false, window, cx);
+        self.insert_crease_impl(
+            formatted_text,
+            "Terminal".to_string(),
+            IconName::Terminal,
+            false,
+            window,
+            cx,
+        );
         self.editor.update(cx, |editor, cx| {
             editor.insert(" ", window, cx);
         });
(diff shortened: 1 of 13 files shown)


crates/agent_ui/src/acp/message_editor.rs
```diff
@@ -1097,11 +1126,30 @@ impl MessageEditor {
...
 
         for chunk in message {
             match chunk {
                 acp::ContentBlock::Text(text_content) => {
-                    text.push_str(&text_content.text);
+                    // Track terminal code blocks for folding
+                    let chunk_text = &text_content.text;
+                    let chunk_start = text.len();
+
+                    // Find all ```terminal blocks in this chunk
```
benbrandt: @danilo-leal any ideas for how we can better track these creases without parsing the string again?

rutgercap: I considered using resource links, but it feels hacky and probably has unintended side effects.

The string parsing happens in `set_message`, which is only called when loading messages from the server/database, restoring edited messages, or error recovery. It's not in a hot path, it doesn't run during typing or streaming. I figured it would be fine for now, but if anybody disagrees I'll gladly implement something more sophisticated.

This seems like a popular use case though (I've seen it in the other editors too), so terminal blocks might be something that's currently missing in ACP.

PR #47663, opened 2026-01-26
git_graph: Add remote repository support

feeiyu:
Closes #ISSUE

This PR introduces two new RPCs, GetInitialGraphCommit and GetGraphCommitData, to fetch commit information from the remote service.
Both RPCs currently use a pull-based approach. I’m not fully sure whether this is the best design, so feedback on this is very welcome.

It also optimizes CommitDataReader with batch processing to improve remote read performance, and fixes an issue where the first commit was updated late because no query was sent when graph_commit_data_handler was in the CLOSE state.

[2026-01-27 20-38-55.webm](https://github.com/user-attachments/assets/c2bf4530-e007-45be-b8f6-1a2cd4d9d638)

Release Notes:

- Added remote repository support


Anthony-Eid:
Thank you for putting this together, I really appreciate the time and thought that went into it. I’m going to close this PR because I’m moving forward with a different implementation on my end.

The main difference is that I’d prefer to treat remote commit loading as an extension of the existing commit data/repository path, rather than introducing Git-graph-specific remote fetching paths and retry/polling behavior. That feels simpler to me, easier to reuse outside the graph, and closer to the direction I want this API to evolve.

PR #28138, opened 2025-04-05
Update default vim substitute command behavior and add support for 'g' flag

dinocosta:
This Pull Request updates the default behavior of the substitute (`s`) command in vim mode to only replace the next match by default, instead of all, and replace all matches only when the `g` flag is provided, making it more similar to NeoVim's behavior.

In order to achieve this, the following changes were introduced:

- Update `BufferSearchBar::replace_next` to be a public method, so it can be called from `Vim::replace_command` .
- Update the `Replacement::parse` to set the `should_replace_all` field to `false` by default, and only set it to `true` if the `'g'` flag is present in the query.
- Add support for when the `Replacement.should_replace_all` is set to `false` in `Vim::replace_command`, so as to have it only replace the next occurrence instead of all occurrences in the line.
- Introduce `BufferSearchBar::select_first_match` so as to activate the first match on the line under the cursor.

Closes #24450 

Release Notes:

- vim: The `:s//` command now defaults to replacing the first match per line (like vim). Use `/g` to replace all matches.

PR #47664, opened 2026-01-26
vim: Add `gdefault` setting to set `/g` as a default substitution flag

bluetech:
vim has a `:set gdefault` flag, which is described as follows:

    'gdefault' 'gd'         boolean (default off)
                        global
        When on, the ":substitute" flag 'g' is default on.  This means that
        all matches in a line are substituted instead of one.  When a 'g' flag
        is given to a ":substitute" command, this will toggle the substitution
        of all or one match.  See complex-change.

                command         'gdefault' on   'gdefault' off
                :s///             subst. all      subst. one
                :s///g            subst. one      subst. all
                :s///gg           subst. all      subst. one

After using vim for years with this flag it is hard to remember to add `/g` all the time, and I think this setting is convenient. So add support for this flag under `vim.gdefault`.

The name `gdefault` comes from vim, but if a less obscure name is better, I can change it to `substitute_g_default` or `substitute_all_in_line_default` or another suggestion.

Was requested here: https://github.com/zed-industries/zed/discussions/36209
Also mentioned as a desired feature here (`g` was the default before this PR): https://github.com/zed-industries/zed/pull/28138#issuecomment-2790132019

Release Notes:

- vim: Add `vim.gdefault` setting to make `/g` (replace all matches in a line) the default for substitutions, along with `:set gdefault` and `:set nogdefault` commands (short forms: `gd`, `nogd`)


dinocosta (approved):
@bluetech Thank you for picking this one up and including tests! 🙌 

I've pushed two smaller commits that add:

- Support for `:set [gd]default` and `:set [nogd]efault` in Zed's command palette
- Documentation in `VimSettingsContent` so that it shows while user's are editing the settings file
- Update the test to use `NeoVimBackedTestContext` so that we actually test against Neovim's behavior

Will just run this over someone else in the team to ensure that we're ok with using `gdefault` for the setting name, but it's looking good. Thank you so much! 🙂

PR #47698, opened 2026-01-26
editor: Store folds by file path for persistence across tab close

Brandtweary:
## Summary

Extends #46011 to make folds survive tab close and workspace cleanup.

- Adds `file_folds` table keyed by `(workspace_id, path)` instead of `editor_id`
- Follows the `breakpoints` table pattern in `workspace/src/persistence.rs`
- Includes backwards-compatible migration from `editor_folds` on first read

cc @Veykril - you reviewed the original fold persistence PR, this extends it to handle the tab-close case.

## Problem

Folds stored by `editor_id` get deleted when:
1. User closes a tab
2. Tab is removed from workspace serialization
3. On next Zed start, `cleanup()` deletes editor rows not in `loaded_items`
4. `ON DELETE CASCADE` wipes associated folds
5. User manually reopens file → folds gone

This is especially painful for PKM/notes workflows where folds represent permanent document structure (collapsed sections, reference blocks, etc).

## Solution

Store folds by file path, not editor ID. The `breakpoints` table already proves this pattern works in Zed - breakpoints persist across tab close because they're keyed by path.

## Migration

The PR includes backwards-compatible migration: reads from `file_folds` first, falls back to `get_editor_folds()`, and migrates old data on first read. Happy to remove this if you'd prefer a clean break - I'd delete `get_editor_folds()`, remove the fallback logic in `read_metadata_from_db()`, and add a `DROP TABLE IF EXISTS editor_folds` migration. Users would lose any folds saved before the update.

## Test Plan

- [x] Unit test for `file_folds` queries in `persistence.rs`
- [x] Manual test: Open file → fold → close tab → quit Zed → reopen → manually open file → folds restored

Release Notes:

- N/A

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -122,7 +122,7 @@ use language::{
     AutoindentMode, BlockCommentConfig, BracketMatch, BracketPair, Buffer, BufferRow,
     BufferSnapshot, Capability, CharClassifier, CharKind, CharScopeContext, CodeLabel, CursorShape,
     DiagnosticEntryRef, DiffOptions, EditPredictionsMode, EditPreview, HighlightedText, IndentKind,
-    IndentSize, Language, LanguageName, LanguageRegistry, LanguageScope, OffsetRangeExt,
+    IndentSize, Language, LanguageName, LanguageRegistry, LanguageScope, LocalFile, OffsetRangeExt,
     OutlineItem, Point, Runnable, Selection, SelectionGoal, TextObject, TransactionId,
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (commented):
Given https://github.com/zed-industries/zed/blob/f1f88b8a169708df40d293d295dd8beda413052f/crates/editor/src/items.rs#L1324-L1334 , is there any point, functionality-wise, in this PR?

Veykril (approved):

PR #47702, opened 2026-01-26
Fix tool calling with empty call_id on OpenAI-compatible APIs

### LINKED ISSUE #47696: Agent tool calling fails with empty call_id when using enterprise OpenAI-compatible API
### Summary

When using Zed agent with an enterprise OpenAI-compatible API, tool calls fail with an error about empty \ parameter.

### Description

I've configured Zed to use our enterprise OpenAI-compatible API endpoint. Regular message exchange works correctly - I can send messages and receive responses. However, when the agent attempts to call tools, the request fails with the following error:

\
The error indicates that Zed is sending an empty string for the \ field in position 8 of the input array, but the API requires this to be a non-empty string.

### Model Provider Details

- **Provider**: Enterprise OpenAI-compatible API
- **Mode**: Agent Panel
- **Tool Calling**: Failing

### Expected Behavior

When the agent needs to call a tool:
1. Zed should generate a valid non-empty \ for each tool call
2. The tool call should execute successfully
3. Results should be returned to the agent

### Actual Behavior

1. Zed sends a tool call request with an empty 2. The OpenAI-compatible API rejects the request with a 400 error
3. Tool calling fails completely

### Reproduction Steps

1. Configure Zed to use an enterprise OpenAI-compatible API endpoint
2. Start a conversation with the agent
3. Request the agent to perform an action that requires tool calling
4. Observe the error in the agent panel

### Additional Context

This appears to be specific to tool calling - regular conversation works fine. The issue may be related to how Zed constructs the tool call payload for OpenAI-compatible endpoints.

(Please add your version information from Help > About Zed or \)

ThePlenkov:
Fixes #47696

## Problem

When using Zed agent with enterprise OpenAI-compatible API endpoints (particularly AWS Bedrock-based), tool calls fail with validation errors:
```
Invalid length for parameter messages[3].content[1].toolUse.toolUseId, value: 0, valid min length: 1
Invalid length for parameter messages[3].content[1].toolUse.name, value: 0, valid min length: 1
```

### Root Cause

Some OpenAI-compatible APIs return tool calls with empty `id` and `name` fields. When Zed sends these back to the API as conversation history in subsequent requests, validation fails because these APIs require non-empty strings.

### Evidence from #47696

Zed was **sending** malformed tool calls in conversation history:
```json
"tool_calls": [
  {
    "id": "",  // Empty string causes validation error
    "type": "function",
    "function": {
      "name": "",  // Empty string causes validation error
      "arguments": "{\"path\":\"test\"}"
    }
  }
]
```

## Solution

Generate valid tool call IDs and names when serializing conversation history for API requests:

### Changes
- **Added UUID dependency** - Added `uuid` crate with v4 feature to language_models crate
- **Fixed request serialization** (`into_open_ai`) - Generate IDs/names when tool calls are sent **back to the API** as conversation history (the actual bug)
- **Fixed streaming responses** - Defense-in-depth: also fix empty IDs when receiving streaming responses
- **Fixed non-streaming responses** - Defense-in-depth: also fix empty IDs when receiving non-streaming responses
- **Added logging** - Warning logs when tool names are empty
- **Added tests** - Both response parsing and round-trip serialization tests

### Generated IDs Format
Uses OpenAI's convention: `call_{uuid}` (e.g., `call_123e4567-e89b-12d3-a456-426614174000`)

## Files Changed
- `crates/language_models/Cargo.toml` - Added uuid dependency
- `crates/language_models/src/provider/open_ai.rs` - Fixed tool call ID/name generation and added tests

---

## ⚠️ Remaining Issue: Empty Tool Names

The current PR fixes empty IDs but **only warns** about empty names without fixing them (lines 456-458):

```rust
let mut tool_name = tool_use.name.to_string();
if tool_name.is_empty() {
    log::warn\!("Tool call has empty name when sending to API, this may cause issues");
}
```

This still causes validation failures:
```
Invalid length for parameter messages[1].content[1].toolUse.name, value: 0, valid min length: 1
```

### Suggested Fix for Empty Names

**Option 1: Infer from arguments (best)**
Match the tool call arguments against the `tools` parameter to infer which tool was called:

```rust
if tool_name.is_empty() {
    // Try to infer from arguments by matching against available tools
    if let Ok(args) = serde_json::from_str::<serde_json::Value>(&tool_use.raw_input) {
        tool_name = infer_tool_name_from_args(&args, &request.tools)
            .unwrap_or_else(|| "unknown_tool".to_string());
    } else {
        tool_name = "unknown_tool".to_string();
    }
    log::warn\!("Generated tool name '{}' for empty name", tool_name);
}
```

Helper function:
```rust
fn infer_tool_name_from_args(args: &serde_json::Value, tools: &[LanguageModelTool]) -> Option<String> {
    let arg_keys: HashSet<&str> = args
        .as_object()?
        .keys()
        .map(|s| s.as_str())
        .collect();
    
    let mut best_match = None;
    let mut best_score = 0;
    
    for tool in tools {
        let required: HashSet<&str> = tool.input_schema.required
            .iter()
            .map(|s| s.as_str())
            .collect();
        let all_params: HashSet<&str> = tool.input_schema.properties
            .keys()
            .map(|s| s.as_str())
            .collect();
        
        let common = arg_keys.intersection(&all_params).count();
        let required_satisfied = required.is_subset(&arg_keys);
        
        if required_satisfied && common > best_score {
            best_score = common;
            best_match = Some(tool.name.to_string());
        }
    }
    
    best_match
}
```

**Option 2: Use placeholder (simpler)**
```rust
if tool_name.is_empty() {
    tool_name = "unknown_tool".to_string();
    log::warn\!("Using placeholder 'unknown_tool' for empty tool name");
}
```

### Testing for Empty Names

The test `into_open_ai_generates_ids_for_empty_tool_calls` currently only validates empty IDs with a valid name (`"create_directory"`). Please add a case that tests **both** empty IDs and empty names:

```rust
MessageContent::ToolUse(LanguageModelToolUse {
    id: LanguageModelToolUseId::from(""),  // Empty ID
    name: "".into(),                         // Empty name
    raw_input: r#"{"path":"test"}"#.to_string(),
    input: serde_json::json\!({"path": "test"}),
    is_input_complete: true,
    thought_signature: None,
})
```

Then verify the serialized request has both valid ID **and** name.

---

## Testing Context

Tested with LiteLLM proxy + custom hook that works around these issues. The hook successfully:
- Generates UUIDs for empty IDs
- Infers tool names by matching arguments against tool schemas
- Maintains ID consistency across conversation history

This workaround validates that the approach works end-to-end.

### DIFF
--- a/crates/language_models/Cargo.toml
+++ b/crates/language_models/Cargo.toml
@@ -63,6 +63,7 @@ tokio = { workspace = true, features = ["rt", "rt-multi-thread"] }
 ui.workspace = true
 ui_input.workspace = true
 util.workspace = true
+uuid = { workspace = true, features = ["v4"] }
 vercel = { workspace = true, features = ["schemars"] }
 x_ai = { workspace = true, features = ["schemars"] }
 
--- a/crates/language_models/src/provider/open_ai.rs
+++ b/crates/language_models/src/provider/open_ai.rs
@@ -32,6 +32,7 @@ use std::str::FromStr as _;
 use std::sync::{Arc, LazyLock};
 use strum::IntoEnumIterator;
 use ui::{ButtonLink, ConfiguredApiCard, List, ListBulletItem, prelude::*};
+use uuid;
 use ui_input::InputField;
 use util::ResultExt;
 
@@ -829,7 +830,15 @@ impl OpenAiEventMapper {
                 events.push(Ok(LanguageModelCompletionEvent::Stop(StopReason::EndTurn)));
             }
             Some("tool_calls") => {
-                events.extend(self.tool_calls_by_index.drain().map(|(_, tool_call)| {
+                events.extend(self.tool_calls_by_index.drain().map(|(_, mut tool_call)| {
+                    // Generate valid IDs and names if they're empty (some OpenAI-compatible APIs return empty values)
+                    if tool_call.id.is_empty() {
+                        tool_call.id = format!("call_{}", uuid::Uuid::new_v4());
+                    }
+                    if tool_call.name.is_empty() {
+                        log::warn!("Tool call has empty name, this may cause issues with the API");
+                    }
+
                     match serde_json::Value::from_str(&tool_call.arguments) {
                         Ok(input) => Ok(LanguageModelCompletionEvent::ToolUse(
                             LanguageModelToolUse {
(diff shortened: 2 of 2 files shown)


ThePlenkov:
@ConradIrwin model is using tool name ( from the list provided by zed ) and next zed call ( I can easily track that with litellm ) has a call without tool name! And the next message is not from user or application but from tool with a message like tool doesn't exist

ConradIrwin:
Wierd! Seems like instead of generating new ones, we should be remembering the old ones.

PR #47713, opened 2026-01-27
repl: Streamline Markdown output usage

rgbkrk:
Brought the Markdown output up to date with how Markdown is used in the Agent panel. This fixed an issue with outputs that were too large for the execution view as well as made sure that markdown would wrap.

image

Release Notes:

- Fixed REPL output wrapping for large content
- Fixed markdown wrapping in REPL outputs


crates/repl/src/outputs/markdown.rs
```diff
@@ -1,44 +1,27 @@
...
-        let task = cx.spawn(async move |markdown_view, cx| {
-            let content = parsed.await;
-
-            markdown_view.update(cx, |markdown, cx| {
-                markdown.parsing_markdown_task.take();
-                markdown.contents = Some(content);
-                cx.notify();
-            })
-        });
+        let markdown = cx.new(|cx| Markdown::new(text.clone().into(), None, None, cx));
```
MrSubidubi, line 13: I think we can just make this text a `SharedString` here and also store it as such to avoid some cloning, right?

crates/repl/src/outputs/markdown.rs
```diff
@@ -70,24 +52,112 @@ impl OutputContent for MarkdownView {
...
-            .py_4()
-            .children(parsed.children.iter().map(|child| {
-                div().relative().child(
-                    div()
-                        .relative()
-                        .child(render_markdown_block(child, &mut markdown_render_context)),
-                )
-            }))
-            .into_any_element()
+    MarkdownStyle {
```
MrSubidubi: I think instead of duplicating this, I'd probably prefer if we were to move the `default_markdown_style` from the `agent_ui` into the markdown crate perhaps (and probably make it an associated function of `MarkdownStyle` even), given that we want to style these the same way anyway. What do you think?

MrSubidubi (commented):
Thank you again for pushing this forward and helping with this. Left two minor notes

crates/repl/src/outputs/markdown.rs
rgbkrk: There is some slight difference as `default_markdown_style` in the agent_ui crate uses `agent_buffer_font_size` and `agent_ui_font_size` from the theme settings. I generally agree with you though as I'd love to have markdown text content be consistent.

crates/repl/src/outputs/markdown.rs
rgbkrk: Made this change, though I still end up having to convert to a String when making a ClipboardItem. I think though that `Markdown` can get its `source` so I can simplify this further and not have to carry around the raw text for that purpose. Seeing if I can make this better now.

rgbkrk: Ok way better now, thanks for noting this.

crates/repl/src/outputs/markdown.rs
MrSubidubi: Didn't notice that, thanks for mentioning it.

I took another look just now, and it might be worth to traitify fonts across the code base at some point, although that is very much a chore work I neither require nor want you to do (after all, thanks for helping here).

However, as an intermediate solution: How would you feel about moving it and have the `default_markdown_style` take an enum instead that covers the current 4 font types, like

```rust
enum MarkdownFont {
    AgentBuffer, AgentUi, Buffer, UI }
```

for the time being?

crates/repl/src/outputs/markdown.rs
MrSubidubi: Yep indeed, thank you!

crates/repl/src/outputs/markdown.rs
rgbkrk: Yeah I can do that.

MrSubidubi (approved):
This is awesome, I love every piece of it! Reads so much better at a glance and love that we do not have to duplicate the style anymore

Looks good to me, thanks again and thank you for the quick turnarounds!

rgbkrk:
Thank you @JosephTLyons! I've updated it.

PR #47720, opened 2026-01-27
markdown: Fix markdown table selection hit testing

cppcoffee:
Release Notes:

- Fixed markdown table selection hit testing

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -1911,19 +1911,33 @@ struct RenderedLink {
 impl RenderedText {
     fn source_index_for_position(&self, position: Point<Pixels>) -> Result<usize, usize> {
         let mut lines = self.lines.iter().peekable();
+        let mut fallback_line: Option<&RenderedLine> = None;
 
         while let Some(line) = lines.next() {
             let line_bounds = line.layout.bounds();
+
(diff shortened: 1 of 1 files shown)


crates/markdown/src/markdown.rs
```diff
@@ -1911,22 +1911,37 @@ struct RenderedLink {
...
-            if position.y > line_bounds.bottom() {
-                if let Some(next_line) = lines.peek()
-                    && position.y < next_line.layout.bounds().top()
-                {
-                    return Err(line.source_end);
-                }
 
-                continue;
+            if line_containing_position.is_none() && line_bounds.contains(&position) {
+                line_containing_position = Some(line);
```
ConradIrwin: @cppcoffee we can just return here, or do you want the last such matching line?

cppcoffee: Logic simplification done

ConradIrwin:
Now we're iterating over the list twice again? Maybe I'm missing something...

cppcoffee:
> Now we're iterating over the list twice again? Maybe I'm missing something...

Changed to a single loop

Handle both exact match and Y-coordinate fallback match in a single loop
Use the `fallback_line` variable to record the first line matching the Y-coordinate

PR #47732, opened 2026-01-27
vim: Restore cursor position when dismissing buffer search

### LINKED ISSUE #8048: vim: dismissing buffer search can change cursor position
### Check for existing issues

- [X] Completed

1. Search for an existing text on the current buffer.
2. Press esc to dismiss the search dialog.
3. The cursor position changes to the first match.

I'd expect the cursor position to not change.

### Environment

Zed: v0.122.2 (Zed)
OS: macOS 14.3.1
Memory: 32 GiB
Architecture: aarch64

_No response_

If you only need the most recent lines, you can run the `zed: open log` command palette action to see the last 1000.

_No response_

lex00:
Fixes #8048

## Summary

In vim mode, pressing Escape to dismiss the buffer search now correctly restores the cursor to its original position, rather than leaving it at the first match.

## Problem

When using vim's `/` command to search:
1. User positions cursor at line X
2. User presses `/` to open search, types a query
3. Matches are highlighted, cursor may visually jump to first match
4. User presses Escape to dismiss without navigating
5. **Bug:** Cursor ends up at first match instead of line X

This breaks vim parity where Escape should cancel the search and restore cursor position.

## Solution

The fix leverages the `focused()` callback in `vim.rs`, which is called when the editor regains focus after the search bar is dismissed.

**Key insight:** When search starts via `/`, the cursor position is saved in `SearchState.prior_selections`. When search is *submitted* with Enter, `search_submit()` drains these selections. But when search is *dismissed* with Escape, they remain.

So in `focused()`, if:
- `prior_selections` is non-empty, AND
- The search bar's `is_dismissed()` returns true

...then we know the user dismissed the search (Escape) rather than submitted it (Enter), and we restore the cursor.

## Why not handle `buffer_search::Dismiss` directly?

The initial approach tried to register a vim handler for the `Dismiss` action. This didn't work because when Escape is pressed, the search bar (which has focus) handles the `Cancel` action internally and calls its `dismiss()` method directly—it doesn't dispatch `Dismiss` through the action system. The vim handler registered on the editor was never invoked.

## Test Plan

- Added `test_search_dismiss_restores_cursor` — verifies cursor restoration when search is dismissed
- Added `test_search_dismiss_restores_cursor_no_matches` — verifies behavior when query has no matches
- All 455 vim tests pass
- Manual testing confirms fix works with both `/` and `cmd-f`

## Release Notes

- Fixed vim mode: cursor now returns to original position when dismissing buffer search with Escape (#8048)

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -94,6 +94,7 @@ impl Deploy {
 
 pub enum Event {
     UpdateLocation,
+    Dismissed,
 }
 
 pub fn init(cx: &mut App) {
@@ -826,6 +827,7 @@ impl BufferSearchBar {
 
     pub fn dismiss(&mut self, _: &Dismiss, window: &mut Window, cx: &mut Context<Self>) {
         self.dismissed = true;
+        cx.emit(Event::Dismissed);
         self.query_error = None;
         self.sync_select_next_case_sensitivity(cx);
 
--- a/crates/vim/src/helix.rs
+++ b/crates/vim/src/helix.rs
@@ -583,6 +583,7 @@ impl Vim {
                         prior_operator: self.operator_stack.last().cloned(),
                         prior_mode: self.mode,
                         helix_select: true,
+                        _dismiss_subscription: None,
                     }
                 });
             }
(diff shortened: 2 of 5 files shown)


ConradIrwin:
Thanks

PR #47734, opened 2026-01-27
vim: Ensure paragraph motions use empty and not blank lines

lex00:
## Summary

Fixes #36171

The `}` and `{` paragraph motions now correctly treat only truly empty lines (zero characters) as paragraph boundaries, matching vim's documented behavior. Whitespace-only lines are no longer treated as boundaries.

### The problem

In vim mode, pressing `}` would stop at lines containing only whitespace (spaces, tabs). According to vim's `:help paragraph`:

> A paragraph begins after each empty line

An "empty" line in vim terminology has zero characters, not whitespace-only.

### The fix

Changed `start_of_paragraph()` and `end_of_paragraph()` in `editor/src/movement.rs` to check `line_len() == 0` instead of `is_line_blank()`.

Note: This change does NOT affect the `ap`/`ip` text objects. Per vim's `:help ap`, those DO treat whitespace-only lines as boundaries, which is the existing (correct) behavior in `vim/src/object.rs`.

## Test plan

- Added `test_paragraph_motion_with_whitespace_lines` using `NeovimBackedTestContext`
- Existing `test_start_end_of_paragraph` still passes

Release Notes:

- Fixed vim mode paragraph motions (`}` and `{`) to correctly ignore whitespace-only lines

### DIFF
--- a/crates/editor/src/movement.rs
+++ b/crates/editor/src/movement.rs
@@ -523,7 +523,7 @@ fn is_subword_boundary_end(left: char, right: char, classifier: &CharClassifier)
 }
 
 /// Returns a position of the start of the current paragraph, where a paragraph
-/// is defined as a run of non-blank lines.
+/// is defined as a run of non-empty lines.
 pub fn start_of_paragraph(
     map: &DisplaySnapshot,
     display_point: DisplayPoint,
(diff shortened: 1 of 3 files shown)


dinocosta (approved):
Thank you @lex00 ! I've pushed a small commit updating the test you added (thanks!) to leverage our NeoVim backed tests, that way we can ensure that this matches NeoVim's behavior too ✌️

aviatesk:
Although I understand the intent to match vim's documented behavior for `{` and `}` motions, I'd like to point out that this change affects the general editor actions (`MoveToStartOfParagraph` / `MoveToEndOfParagraph`), not just vim mode.
For vim users, this is indeed a fix, but for users who don't use vim mode (like myself), this is a breaking change.  The previous behavior of treating whitespace-only lines as paragraph boundaries was useful for navigating code blocks separated by indented blank lines.

It's also worth noting that even within the vim community, this behavior has been https://github.com/vim/vim/issues/222, with plugins like
https://github.com/dbakker/vim-paragraph-motion existing specifically to restore the "blank line as boundary" behavior.

Would it be possible to either:
1. Apply this stricter "empty line only" behavior exclusively within vim mode, or
2. Make this configurable via a setting?

This would preserve vim compatibility for those who need it while avoiding regression for other users.

PR #47745, opened 2026-01-27
extensions_ui: Fix duplicated button element IDs preventing uninstall

fury0928:
Closes #47656

Release Notes:

- Fixed extension uninstall button not working when an upgrade is available.

### DIFF
--- a/crates/extensions_ui/src/extensions_ui.rs
+++ b/crates/extensions_ui/src/extensions_ui.rs
@@ -1041,7 +1041,7 @@ impl ExtensionsPage {
             },
             ExtensionStatus::Upgrading => ExtensionCardButtons {
                 install_or_uninstall: Button::new(
-                    SharedString::from(extension.id.clone()),
+                    SharedString::from(format!("uninstall-{}", extension.id)),
                     "Uninstall",
                 )
                 .style(ButtonStyle::OutlinedGhost)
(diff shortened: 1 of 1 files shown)


crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -1041,7 +1041,7 @@ impl ExtensionsPage {
             },
             ExtensionStatus::Upgrading => ExtensionCardButtons {
                 install_or_uninstall: Button::new(
-                    SharedString::from(extension.id.clone()),
+                    SharedString::from(format!("uninstall-{}", extension.id)),
```
MrSubidubi: Lets go for something like 

```suggestion
                    (SharedString::from(extension.id.clone()), ExtensionOperation::Uninstall as usize),
```

and vice versa for the others as this will prevent us from allocating a string on every render and should have the same effect.

MrSubidubi (dismissed):
Nice find, thank you for looking into this! One comment and we are good to merge.

crates/extensions_ui/src/extensions_ui.rs
fury0928: Pushed the fix

maxdeviant (approved):
Thanks!

PR #47751, opened 2026-01-27
editor: Add support for no auto-indent on enter

oliverbarnes:
Closes #47550

Changes the `auto_indent` setting from a boolean to an enum with three modes:

- **`syntax_aware`** (default): Adjusts indentation based on syntax context when typing (previous `true` behavior)
- **`preserve_indent`**: Preserves the current line's indentation on new lines, but doesn't adjust based on syntax
- **`none`**: No automatic indentation - new lines start at column 0 (previous `false` behavior)

This gives users more control over indentation behavior. Previously, setting `auto_indent: false` would still preserve indentation on new lines, which was unexpected.

Includes:
- Settings migration from boolean to enum values
- Settings UI dropdown renderer

Release Notes:

- Changed `auto_indent` setting from boolean to enum with `syntax_aware`, `preserve_indent`, and `none` options

Screenshot 2026-01-27 at 16 32 10


MrSubidubi (changes requested):
Thanks for this!

I see the use-case for this, however, this is the first complaint that we have had for this feature and I do not think this should be an on-off swich.

Can we instead refactor this setting to be an enum that holds these three variants, add/modify a settings migration for that and then go with that route?

crates/editor/src/editor.rs
```diff
@@ -5124,29 +5124,44 @@ impl Editor {
                                 extra_line_additional_indent,
                                 prevent_auto_indent,
                             } => {
+                                let auto_indent_mode =
+                                    buffer.language_settings_at(start, cx).auto_indent;
+                                let preserve_indent =
+                                    auto_indent_mode != language::AutoIndentMode::None;
+                                let use_autoindent =
+                                    auto_indent_mode == language::AutoIndentMode::Full;
```
MrSubidubi: Can you elaborate as to why this is needed here? At a glance, it seems to be unrelated.

crates/settings_content/src/language.rs
```diff
@@ -260,6 +260,32 @@ pub enum EditPredictionsMode {
...
+    JsonSchema,
+    MergeFrom,
+    strum::VariantArray,
+    strum::VariantNames,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum AutoIndentMode {
+    /// Full auto-indent based on syntax context.
+    /// Adjusts indentation based on the code structure when typing.
+    Full,
```
MrSubidubi: Not too sure I like the naming here, what about `ContextAware` or something similar? I feel like `Full` does not immediately make you understand as to what actually happens here

MrSubidubi (changes requested):
Two minor questions, otherwise looks good!

crates/settings_content/src/language.rs
oliverbarnes: Hmm, agree - ended up going with SyntaxAware, to be more specific - wdyt?

crates/editor/src/editor.rs
oliverbarnes: Yeah, this is a little confusing, isn't it 😅 I now renamed `use_autoindent` to `apply_syntax_indent` as part of the refactor, so that:

```
let (edit_start, new_text, prevent_auto_indent) = match &newline_config {
...
        (start, new_text, *prevent_auto_indent || !apply_syntax_indent)
```

As far as I could tell, we need this to avoid adding the tree-sitter adjustment when on `preserve_indent` mode.

Does it make sense?

crates/settings_content/src/language.rs
MrSubidubi: Yup, that sounds good, thanks!

crates/editor/src/editor.rs
MrSubidubi: Well, but we did not need this previously, did we? This was more what my question was about.

That said, I think when I added this this guard here did not exist, so I think it should be fine to add (and be perhaps a bit more efficient). Just wanted to check because strictly speaking, this change is not related to the feature this PR wants to add, right?

oliverbarnes: Ok I see what you mean now. But we didn't need this because the preserve_indent scenario didn't exist yet, right? 

We only had the no-indent and syntax-indent scenarios. If I'm understanding correctly, of course - I might be missing something.

But if this is correct, then to introduce preserve-indent (which is dumber and only follows the previous line's indentation), we need this here to ensure it works?

I'm pushing a couple of tests to demonstrate.

oliverbarnes: (re-requested review just in case this fell-off the radar - no rush)

MrSubidubi (approved):
Sorry for the delay, was out sick. Looks good, thanks!

crates/editor/src/editor.rs
MrSubidubi: Strictly speaking, I'd argue we would not need this, as we do already also handle it downstream (e.g. the functionality to have no auto-indentation existed prior already). That's what I meant here. We could remove `use_autoindent` and everything would still be working here as expected. Decided to keep this around though, so all good.

injust:
> Not sure how this is a breaking change though?

Oh, I guess it isn't. But it's in that section at the bottom, anyway.

crates/editor/src/editor.rs
bbaseemfawzeelhware-dotcom: لن افهم أي شيء

PR #47752, opened 2026-01-27
git_ui: Align repo and branch pickers

tartarughina:
When working in a workspace with multiple repositories, the git panel provides a repository picker to switch between them. However, there was no visual indication of which repositories have uncommitted changes: users had to either select each repository individually or check the project panel where modified directories are highlighted.

This change adds git status icons to the repository picker, allowing users to see at a glance which repositories contain changes (modified, added, deleted, or conflicted files). The icons use the same visual language already established for file status throughout the git panel.

Additionally, the repository picker now matches the branch picker's styling for visual consistency:
- Added "Repositories" header
- Aligned popover width and positioning
- Added scrollbar
- Added check icon next to currently selected repo
- Added selected branch under repo list item
- Sort by display name is now case insensitive

Before:

Screenshot 2026-01-27 at 11 43 55

After:

Screenshot 2026-01-27 at 14 07 52

Branch picker for style reference:

Screenshot 2026-01-27 at 11 44 03

Release Notes:

- Git: Improved the project picker in the panel by also displaying the GIt status icon on them, to clearly indicate which repos have changes.

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -5775,7 +5775,7 @@ impl RenderOnce for PanelRepoFooter {
                 let project = project;
                 move |window, cx| {
                     let project = project.clone()?;
-                    Some(cx.new(|cx| RepositorySelector::new(project, rems(16.), window, cx)))
+                    Some(cx.new(|cx| RepositorySelector::new(project, rems(20.), window, cx)))
                 }
             })
             .trigger_with_tooltip(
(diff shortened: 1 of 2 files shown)


danilo-leal (approved):
Thanks for the PR & congrats on your first contribution! I pushed a few commits cleaning the design up a bit:
- Don't think we need the branch name here because that's displayed right to the side of the project name already, no need to repeat it. Having it in the project list item was also messing up the height; particularly given this is a uniform list (i.e., all items must have the same height), the fact that some had it and some didn't meant a lot of layout shift
- Improved the icon alignment
- Made the currently selected list item when you open the picker be the active repo, instead of just resetting the active index to 0 always
- Made the button trigger have a default color when available (it was muted before). Also removed the "Switch Active Repository" tooltip when in a single repo scenario
- Removed the header because in this case, it's always repositories, so the header becomes unnecessary. We have one in the branch picker because the header changes depending on whether you have the filter for remotes toggled on

Image

PR #47754, opened 2026-01-27
Fix VSCode tasks.json parsing for tasks without explicit labels

### LINKED ISSUE #47749: VSCode tasks.json parser fails on valid tasks without explicit `label` field
### Reproduction steps

1. Create a project with a `.vscode/tasks.json` file containing tasks without explicit `label` fields (valid in VSCode)
2. Open the project in Zed
3. Observe error in log file

Example `tasks.json` that reproduces the issue:
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "type": "npm",
      "script": "start",
      "isBackground": true,
      "problemMatcher": {
        "owner": "typescript",
        "pattern": "$tsc",
        "background": {
          "activeOnStart": true,
          "beginsPattern": {
            "regexp": "(.*?)"
          },
          "endsPattern": {
            "regexp": "bundle generation complete"
          }
        }
      }
    },
    {
      "type": "npm",
      "script": "test"
    }
  ]
}
```

**Current behavior:** Zed fails to parse the tasks.json file with the following error:
```
ERROR [crates/project/src/project_settings.rs:1159] parsing VSCode tasks, file "<project>/.vscode/tasks.json"

Caused by:
    tasks[0]: missing field `label` at line 22 column 5
```

**Expected behavior:** Zed should parse tasks without explicit `label` fields and auto-generate labels like VSCode does:
- For `"type": "npm"` with `"script": "start"` → label should be `"npm: start"`
- For `"type": "npm"` with `"script": "test"` → label should be `"npm: test"`
- For `"type": "gulp"` with `"task": "build"` → label should be `"gulp: build"`
- For `"type": "shell"` → use the command as the label

This is valid VSCode behavior per the [official tasks documentation](https://code.visualstudio.com/docs/editor/tasks).

Zed: v0.221.0+dev.6574fdacb995c49c1c8f4bb00a85f22640139869 (Zed Dev 6574fdacb995c49c1c8f4bb00a85f22640139869) (Taylor's Version)
OS: macOS 26.2
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

dastrobu:
Implements automatic label generation for VSCode tasks that don't have explicit 'label' fields, matching VSCode's behavior.

Changes:
- Made label field optional in VsCodeTaskDefinition deserialization
- Implemented custom deserializer to auto-generate labels:
  - npm tasks: 'npm: {script}' (e.g., 'npm: start')
  - shell tasks: first word of command (e.g., 'echo')
  - gulp tasks: 'gulp: {task}' (e.g., 'gulp: build')
  - fallback: 'Untitled Task'
- Added test data file with tasks without labels
- Added test cases

Closes #47749

Release Notes:

- Fixed: VSCode tasks.json files with tasks missing explicit `label` fields now parse correctly. Labels are auto-generated matching VSCode's behavior (e.g., "npm: start").

### DIFF
--- a/crates/task/test_data/tasks-without-labels.json
+++ b/crates/task/test_data/tasks-without-labels.json
@@ -1,0 +1,22 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "type": "npm",
+      "script": "start"
+    },
+    {
+      "label": "Explicit Label",
+      "type": "npm",
+      "script": "test"
+    },
+    {
+      "type": "gulp",
+      "task": "build"
+    },
+    {
+      "type": "shell",
+      "command": "echo hello"
+    }
+  ]
+}
--- a/crates/task/src/vscode_format.rs
+++ b/crates/task/src/vscode_format.rs
@@ -13,15 +13,44 @@ struct TaskOptions {
     env: HashMap<String, String>,
 }
 
-#[derive(Clone, Debug, Deserialize, PartialEq)]
-#[serde(rename_all = "camelCase")]
+#[derive(Clone, Debug, PartialEq)]
 struct VsCodeTaskDefinition {
     label: String,
-    #[serde(flatten)]
     command: Option<Command>,
-    #[serde(flatten)]
     other_attributes: HashMap<String, serde_json_lenient::Value>,
     options: Option<TaskOptions>,
+}
+
+impl<'de> serde::Deserialize<'de> for VsCodeTaskDefinition {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct TaskHelper {
+            #[serde(default)]
+            label: Option<String>,
+            #[serde(flatten)]
+            command: Option<Command>,
+            #[serde(flatten)]
+            other_attributes: HashMap<String, serde_json_lenient::Value>,
+            options: Option<TaskOptions>,
+        }
+
+        let helper = TaskHelper::deserialize(deserializer)?;
+
+        let label = helper
+            .label
+            .unwrap_or_else(|| generate_label(&helper.command));
+
+        Ok(VsCodeTaskDefinition {
+            label,
+            command: helper.command,
+            other_attributes: helper.other_attributes,
+            options: helper.options,
+        })
+    }
 }
 
 #[derive(Clone, Deserialize, PartialEq, Debug)]
@@ -39,6 +68,21 @@ enum Command {
     Gulp {
         task: String,
     },
+}
+
+fn generate_label(command: &Option<Command>) -> String {
+    match command {
+        Some(Command::Npm { script }) => format!("npm: {}", script),
+        Some(Command::Gulp { task }) => format!("gulp: {}", task),
+        Some(Command::Shell { command, .. }) => {
+            if command.trim().is_empty() {
+                "shell".to_string()
+            } else {
+                command.clone()
+            }
+        }
+        None => "Untitled Task".to_string(),
+    }
 }
 
 impl VsCodeTaskDefinition {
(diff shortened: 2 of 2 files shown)


SomeoneToIgnore (commented):
Thank you, this looks good but https://github.com/zed-industries/zed/issues/47749 mentions more cases that this PR does not implement.

Yet, in the documentation link I see no such rules too, so quite confused: should we actually do what's proposed in the issue, what's in the PR, something else?

Can we have a link to the docs/sources with these rules?

Lastly, we want to test more conversion cases for coverage, even at the current PR's state.

dastrobu:
Thanks for the feedback!

> Yet, in the documentation link I see no such rules too, so quite confused: should we actually do what's proposed in the issue, what's in the PR, something else?

The issue mentions grunt and typescript task types, but these aren't currently supported by Zed's parser. Adding label generation for them would require first implementing full parsing support. Here's what Zed currently handles:

```rs
enum Command {
    Npm {
        script: String,
    },
    Shell {
        command: String,
        #[serde(default)]
        args: Vec<String>,
    },
    Gulp {
        task: String,
    },
}
```

> Can we have a link to the docs/sources with these rules?

Sure! Here's where VSCode implements this:

The core logic is in taskConfiguration.ts lines 1533-1547:
https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/tasks/common/taskConfiguration.ts#L1533-L1547

```typescript
if (result.configurationProperties.name) {
    result._label = result.configurationProperties.name;
} else {
    let label = result.configures.type;
    if (typeDeclaration.required && typeDeclaration.required.length > 0) {
        for (const required of typeDeclaration.required) {
            const value = result.configures[required];
            if (value) {
                label = label + ': ' + value;
                break;
            }
        }
    }
    result._label = label;
}
```

Basically: if there's no label field, VSCode takes the task type and appends the first required property value.

The task type definitions show what's required:

https://github.com/microsoft/vscode/blob/main/extensions/npm/package.json#L344-L360
https://github.com/microsoft/vscode/blob/main/extensions/gulp/package.json#L49-L69
https://github.com/microsoft/vscode/blob/main/extensions/grunt/package.json#L50-L71
https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/package.json#L1894-L1910
https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/tasks/common/taskConfiguration.ts#L1574-L1576

```typescript
if (!taskName) {
    context.problemReporter.error(nls.localize('ConfigurationParser.noTaskName',
        'Error: a task must provide a label property. The task will be ignored.\n{0}\n',
        JSON.stringify(external, null, 4)));
    return undefined;
}
```

This brings up the shell task question - the PR currently generates labels for shell tasks (using the first word of the command), but VSCode actually errors on these. 

We have a few options:
1. Match VSCode and reject shell tasks without labels
2. Keep the current behavior (more permissive)
3. Accept them but log a warning

I'd lean toward option 2 since it's more user-friendly and Zed already diverges from VSCode on which task types are supported. But what's your preference?

> Lastly, we want to test more conversion cases for coverage, even at the current PR's state.

Done! I refactored the label generation into a standalone generate_label() function that can be tested independently.

crates/task/test_data/tasks-without-labels.json
```diff
@@ -0,0 +1,22 @@
...
+    {
+      "type": "gulp",
+      "task": "build"
+    },
+    {
+      "type": "shell",
+      "command": "echo hello"
+    }
+  ]
+}
```
SomeoneToIgnore: Would be good to have a trailing newline, same as the rest of the files in this repo.

SomeoneToIgnore (commented):
Thank you for sharing the context.

Seems reasonable to keep things as they are then, one last concern about the `shell` command: 

Image

is it really ok to keep just the first word of the command?
I imagine there could be multiple `echo` or `cargo` or `npm` or whatnot and we'll have a conflict — instead, does it make sense to have the entire command as a label?

SomeoneToIgnore (approved):
Thanks!

PR #47770, opened 2026-01-27
gpui: Fix nested deferred support

huacnlee:
Release Notes:

- N/A

To fix crash when use `deferred` in a `deferred`.

```
thread 'main' (4024343) panicked at /Users/jason/.cargo/git/checkouts/zed-a70e2ad075855582/83ca310/crates/gpui/src/window.rs:2411:9:
assertion `left == right` failed: cannot call defer_draw during deferred drawing
  left: 1
 right: 0
```

#### Case 1

- A Popover (used `deferred`) with a form.
- An Input element in the Popover form.
- Input have ContextMenu that used `deferred`.
- User right click Input to open menu.

image

#### Case 2

- A Popover (`deferred`) with some content.
- There have a `info` icon in the content to show some tip by open another Popover.

### Test by `examples/popover.rs`

image


Veykril (approved):
Thanks!

PR #47795, opened 2026-01-27
New multi workspace

mikayla-maki:
It's happeningggggg

Release Notes:

- Changed the Agent Panel so that the Active Thread is restored on restart.


crates/dev_container/src/devcontainer_api.rs
```diff
@@ -631,17 +543,16 @@ fn parse_json_from_cli<T: serde::de::DeserializeOwned>(raw: &str) -> Result<T, D
 }
 
 fn devcontainer_cli_command(
-    path_to_cli: &PathBuf,
-    found_in_path: bool,
+    cli: &DevContainerCli,
     node_runtime_path: &PathBuf,
```
KyleBarton: We could probably add an Option<NodeRuntime> to the `DevContainerCli` struct, do the optional path extraction logic here; would reduce duplication in all the calling sites. That's really the intent of this function anyway: to wrap the "do I call `node devcontainer.js` or just call `devcontainer`" logic, depending on what's installed

PR #47800, opened 2026-01-27
git: Fix stage/unstage failure with a large number of files

marcocondrache:
Git commands can fail when staging or unstaging a very large number of files because each file path is passed as a separate CLI argument. Once the argument list grows beyond the OS limit, git errors with Argument list too long (os error 7). 

Since git 2.26, pathspecs can be read from stdin, which lets us avoid passing thousands of paths as arguments and bypass this limitation.

I looked for existing issues related to this but couldn’t find any.

Repro:

- Clone rust-lang repo
- Delete root tests folder
- Try to stage or unstage (40,000+ files)
- Git fails with Argument list too long (os error 7)

Release Notes:

- Fixed an issue where stage/unstage could fail when operating on a very large number of files

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -1954,35 +1954,15 @@ impl GitRepository for RealGitRepository {
         let git_binary_path = self.any_git_binary_path.clone();
         self.executor
             .spawn(async move {
-                if paths.is_empty() {
-                    return Ok(());
-                }
-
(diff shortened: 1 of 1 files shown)


marcocondrache:
@cole-miller thanks! I added the fallback to the old command. Since `pathspec-from-file` is available from git 2.26 (2020), most users should hit the first path anyway. The stage method uses `update-index`, which has supported --stdin since the earliest releases, so this should be safe.

PR #47806, opened 2026-01-27
git_ui: Avoid blocking main thread during stage/unstage

marcocondrache:
With a large number of files in the git status, pressing stage/unstage could cause a noticeable freeze. The async performance profiler showed that the task spawned inside `change_all_files_stage` blocked the main thread for 300ms+, resulting in a clear and visible UI stall. The main cause was the need to traverse all entries to determine what to stage or unstage, which becomes expensive for large change sets.

Same repro as: #47800

Release Notes:

- Improved latency of stage/unstage operations


Veykril (approved):
Thanks!

PR #47824, opened 2026-01-28
project: Increase search detect yield threshold to 20KB

cppcoffee:
Release Notes:

- Improved CPU switching frequency for project search

### DIFF
--- a/crates/project/src/search.rs
+++ b/crates/project/src/search.rs
@@ -345,16 +345,16 @@ impl SearchQuery {
 
         let mut text = String::new();
         let mut bytes_read = 0;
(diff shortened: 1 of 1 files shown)


cppcoffee:
For modern CPU and I/O speeds, the 128-byte threshold is too small. This means the program will extremely frequently pause the current task and yield control back to the scheduler. This excessive Context Switching causes huge CPU overhead, severely slowing down the scanning speed of large batches of files.


Partially solve https://github.com/zed-industries/zed/issues/47767

osiewicz (approved):
Makes sense :+1: We'll probably merge it tomorrow, as today's a release day. Let's give it some baking time on Nightly.

PR #47829, opened 2026-01-28
cli: Prevent duplicate workspaces on macOS restoration

faustbrian:
When macOS restores workspace windows on launch (with "Close windows when quitting an application" disabled in system preferences), and then `zed .` is invoked via CLI for the same workspace, Zed was opening duplicate windows - one from macOS restoration and another from the CLI handler.

This fix adds a check in the `open_workspaces` function to detect if workspace windows already exist before attempting to restore from settings. If windows are already open (e.g., from macOS restoration), the function returns an empty vector instead of calling `restorable_workspace_locations()`, preventing duplicate window creation.

Side effects: None. The fix only affects the CLI path when no paths are provided and `open_new_workspace != Some(true)`. All other behaviors (welcome screen, empty window creation, explicit path opening) remain unchanged.

Fixes #40588

Release Notes:

- N/A


faustbrian:
@MrSubidubi resolved conflicts

PR #47830, opened 2026-01-28
Reapply titlebar settings on window restoration on macOS

### LINKED ISSUE #33709: [macOS] Restored Zed window gets native window border
### Summary

Zed shows up with a native macOS window title bar, drawing over its custom one.

### Description

There's a macOS sort of.. bug (feature?), where when an app has both a full-screen space, and a window in a desktop space, and then you reboot, they are both restored into the same space, with the desktop window drawing over the full-screen space. This is something I've seen in other apps, notably Mail.

So I had one project open in full-screen, and one open on the desktop. When I rebooted, the small window was drawn floating atop the full-screen instance of Zed, and the small window had a native Mac titlebar, as can be seen in the image.

Image

I have to apologise for the poor repro image - it happened on a work machine so I've had to redact code/filenames, and unfortunately I closed the full-screen instance before I thought to make a bug report. If I can reproduce consistently I'll try to get a better report/screenshot.

Steps to reproduce:
1. Open one Zed project and full-screen that window.
2. Swipe back to the desktop and open another project, don't full-screen it.
3. Restart the Mac with "Restore window positions" enabled.
4. Observe the child window having incorrect window decoration.

I am running the macOS 26 Tahoe Beta. I'm not sure if that's related to the window border bug, but it definitely isn't what causes child windows to be restored in the wrong space - I've seen that behaviour on macOS since Sierra.

Zed: v0.193.3 (Zed Preview) 
OS: macOS 16.0.0
Memory: 16 GiB
Architecture: aarch64

faustbrian:
When macOS restores windows after reboot (with "Restore windows when quitting an app" enabled), the transparent titlebar settings were not being reapplied, causing the native macOS titlebar to appear instead of Zed's custom transparent titlebar.

This fix adds a helper method that reapplies titlebar settings when a window becomes key. The method checks the transparent_titlebar flag and reapplies setTitlebarAppearsTransparent and setTitleVisibility calls as needed.

Side effects: None. This only reapplies existing settings that should have been preserved by macOS but weren't.

Fixes #33709

Release Notes:

- N/A


MrSubidubi:
Thanks for this! I also see this ocasionally fail, however, most of the time it does work for me. Could you perhaps elaborate how this will fix this issue for good and why `window_did_change_key_status` is the proper place for this? Thanks!

faustbrian:
@MrSubidubi resolved conflicts

PR #47838, opened 2026-01-28
docs: Document default for include_warnings

fernandoperezlara:
## Summary
Add documentation for the default value of `include_warnings` in diagnostics settings. This makes it consistent with other settings that document their default values.

## Test plan
- Documentation-only change, no functional changes

Release Notes:

- N/A


SomeoneToIgnore (approved):
Thanks.

PR #47847, opened 2026-01-28
anthropic: Support alternative provider SSE formatting

raphaelluethy:
Closes #ISSUE

The issue I ran into was that responses from anthropic compatible providers, like Kimi for Coding, have no space after `data:`. This change just adds a quick check to also allow for those providers to work.

Before it just resolved but did not show any output:
CleanShot 2026-01-28 at 12 50 31@2x

Now it returns the proper result:
CleanShot 2026-01-28 at 12 56 30@2x

Normal Anthropic models still work as expected:
CleanShot 2026-01-28 at 12 58 37@2x

Config to test
```json
 "language_models": {
    "anthropic": {
      "api_url": "https://api.kimi.com/coding",
      "available_models": [
          {
            "name": "kimi-for-coding",
            "display_name": "Kimi 2.5 Coding",
            "max_tokens": 262144,
            "max_output_tokens": 32768,
          },
      ],
    },
}
```

TLDR:
- Accepts SSE data:{...} lines (no space) emitted by some alternative Anthropic providers, in addition to the standard data: {...} format.

Release Notes:

- Fixed Anthropic streaming for alternative providers by accepting SSE data:{...} (no space) lines.

### DIFF
--- a/crates/anthropic/src/anthropic.rs
+++ b/crates/anthropic/src/anthropic.rs
@@ -703,7 +703,16 @@ pub async fn stream_completion_with_rate_limit_info(
             .filter_map(|line| async move {
                 match line {
                     Ok(line) => {
-                        let line = line.strip_prefix("data: ")?;
+                        // Some Anthropic-compatible endpoints omit the space after `data:`.
+                        // Accept both `data: {json}` and `data:{json}`.
+                        let line = match line.strip_prefix("data:") {
+                            Some(rest) => rest.trim_start(),
+                            None => return None,
+                        };
+
+                        if line.is_empty() {
+                            return None;
+                        }
                         match serde_json::from_str(line) {
                             Ok(response) => Some(Ok(response)),
                             Err(error) => Some(Err(AnthropicError::DeserializeResponse(error))),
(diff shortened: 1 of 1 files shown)


crates/anthropic/src/anthropic.rs
```diff
@@ -703,7 +703,11 @@ pub async fn stream_completion_with_rate_limit_info(
             .filter_map(|line| async move {
                 match line {
                     Ok(line) => {
-                        let line = line.strip_prefix("data: ")?;
+                        let line = match line.strip_prefix("data:") {
+                            Some(rest) => rest.trim_start(),
```
benbrandt: Hmm I think we'd want to only trim up to one space?

raphaelluethy: I added that to prevent any occurrence of the issue, one space or more should not impact the happy path. But yes, for my issue 1 space would be the original fix.

raphaelluethy: As an addition: I am not sure if I got your comment correctly. The implementation for Antropic itself would leave no space, as it just directly matches the single space that exists.

My solution accounts for 0, or many spaces (the bug originally was having **no** space after `:`.

The following logic does not expect a space to be available, so there should be no issue.

Did I understand correctly? Or have you meant something different?

benbrandt: The issue is that trim_start will now strip more than one space

benbrandt: Updated to the same behavior we have for openai

benbrandt (approved):

PR #47851, opened 2026-01-28
editor: Use integration tests

osiewicz:
- **Update naming of existing tests for fs, project and worktree**
- **WIP: Use integration tests for editor**

Release Notes:

- N/A


osiewicz:
Yeah, there's just one integration test crate in editor_tests.rs; integration tests can use modules tho, which is why what we're doing here works. Cargo would create a binary-per-source (as you've said)!if we hadn't used a subdirectory here (`tests/editor_tests.rs` instead of what's proposed in this PR:  `tests/integration/editor_tests.rs`)

PR #47868, opened 2026-01-28
client: Store `ReleaseChannel` instead of `&'static str`

tidely:
Closes #ISSUE

Store `ReleaseChannel` instead of `&'static str`. This allows displaying other aspects about the channel than just the display name. Makes it more concrete what is stored within the field, as it has limited variants, as opposed to something like `session_id`

Release Notes:


MrSubidubi (approved):
Thanks

PR #47871, opened 2026-01-28
dap: Make `ProtoConversion::to_proto` consume `self`

tidely:
Closes #ISSUE

Make `ProtoConversion::to_proto` consume `self`. We never create more than one `proto` type out of a single `dap` type, meaning having the conversation be non-consuming has no benefits, while consuming allows us to remove a bunch of clones.

Release Notes:

- N/A


MrSubidubi (approved):
Thanks!

PR #47888, opened 2026-01-28
Add timeout to open task to prevent crash loops

namanag97:
When Zed restores workspaces on startup, it attempts to open files from the previous session. If a file triggers a parser failure (e.g., a large XES/XML file causing a tree-sitter-html OOM), the open_task in restore_or_create_workspace hangs indefinitely.

This causes the app to consume massive amounts of memory (400GB+) and prevents the user from launching Zed, creating an unrecoverable crash loop.

Solution

This PR adds a safety wrapper using tokio::time::timeout around the workspace::open_paths task.

Duration: 5 seconds.
Behavior: If the file cannot be opened within 5 seconds, the task is canceled, an error is logged, and the startup sequence continues (skipping the problematic file).
Impact

Prevents the app from freezing on startup due to problematic parser states.
Allows users to launch Zed and remove the problematic file manually instead of being stuck in a crash loop.

### DIFF
--- a/crates/zed/src/main.rs
+++ b/crates/zed/src/main.rs
@@ -50,6 +50,8 @@ use std::{
     sync::{Arc, OnceLock},
     time::Instant,
 };
+use std::time::Duration;
+use tokio::time::timeout;
 use theme::{ActiveTheme, GlobalTheme, ThemeRegistry};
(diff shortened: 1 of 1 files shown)


namanag97:
@cla-bot check

namanag97:
@cla-bot check

Issue #46918, opened 2026-01-15
Fail to detect new kernels in remote machine


Eddudos:
### Reproduction steps

1. Connect via ssh to the machine
2. Create python uv project and  env in .venv
3. Add ipykernel and create kernel as described in doc (https://zed.dev/docs/repl#repl)
4. See if there are kernels:
```
✗ jupyter kernelspec list
Available kernels:
  some-venv    /home/user/main_projects/free-text/.venv/share/jupyter/kernels/some-venv
  python            /home/user/main_projects/free-text/.venv/share/jupyter/kernels/python
  te         /home/user/.local/share/jupyter/kernels/te
  python3           /home/user/.local/share/jupyter/kernels/python3
  venv              /home/user/.local/share/jupyter/kernels/venv
```
5. workspace: reload, repl: refresh kernelspec, repl: sessions doesn't show any kernels for some reason.

### Current vs. Expected behavior

Current behavior: doesn't show kernels even if they exist in proper places.
I expected it to detect them. Am I doing something wrong? 

### Zed version and system specs

Zed 0.219.4 0.219.4+stable.119.5595f7ce23d51a58eb66dd025a487bbd6c75dad1
macOS sonoma 14.5 (local), ubuntu 24.04 (remote)

### If you are using WSL on Windows, what flavor of Linux are you using?

None

rgbkrk:
> I still experience exactly the same issue on Zed 0.225.13, can't use repl on remote machine I'm new to Zed so currently have no idea how to debug this/collect any useful logs

The fix for this hasn't landed on a stable release yet. However it is in Zed preview if you can test it out.

14Si:
I also seem to be having this issue.

tejasunku:
Not sure this will help, but I had this problem and I managed to fix it.

I am running Ubuntu 24.04 and I needed to install the python packages globally for Zed to be able to find the kernels. In particular, I ran the following:

```
#Should be enough, running this at the end fixed a kernel loading issue I had, was getting errors about missing ipykernel_launcher before installing this
sudo apt-get install jupyter-client

#Things I installed before that, unsure if they'll make a difference. May be installed if you install jupyter-client
sudo apt-get install python3-ipykernel
sudo apt-get install jupyter-notebook
```

[...]

PR #47891, opened 2026-01-28
repl: Add WSL and SSH remote kernel support

### LINKED ISSUE #15196: Remote REPL
### Check for existing issues

- [X] Completed

### Describe the feature

Run jupyter repl on remote machines

From what I could understand from code:
* Currently `repl_store` is global, and not updated when changing projects (i.e. connecting to remote project)
* `JupyterSettings` are global too, and project settings are ignored (there is a big change that remote machine needs different kernel configs)

It seems that current code _could_ work on remote, if it weren't global, and it would update to changing workspaces and connecting to remote.

_No response_

MostlyKIGuess:
Closes #15196, #46918 

- fix: notebook_ui, use buffer so that notebooks open in remote/WSL settings.
- fix: add musl in nix for cross-compilation, without this remote server doesn't build inside NixOS

Release Notes:

- Implement WSL and SSH remote kernels (crates/repl/src/kernels/*) and wire up spawn/kill kernel proto messages and client requests.


crates/project/src/toolchain_store.rs
```diff
@@ -683,8 +683,17 @@ impl RemoteToolchainStore {
                     language_name: language_name.clone().into(),
                     path: Some(path.path.to_proto()),
                 })
-                .await
-                .log_err()?;
+                .await;
```
rgbkrk: This could all be

```rust
            .await
            .inspect_err(|e| {
                log::warn!(
                    "RemoteToolchainStore::list_toolchains: RPC failed for language {}: {:?}",
                    language_name,
                    e
                );
            })
            .ok()?;
```

crates/remote_server/src/headless_project.rs
```diff
@@ -788,6 +795,92 @@ impl HeadlessProject {
...
+
+        let working_directory = if envelope.payload.working_directory.is_empty() {
+            std::env::current_dir()
+                .ok()
+                .map(|p| p.to_string_lossy().into_owned())
+        } else {
+            Some(envelope.payload.working_directory)
+        };
+
+        // Spawn kernel (Assuming python for now, or we'd need to parse kernelspec logic here or pass the command)
```
rgbkrk, line 951: True, we're probably going to have to provide the listing based on what's available on the remote as well as be able to launch using those as well.

crates/repl/src/kernels/mod.rs
```diff
@@ -116,6 +186,35 @@ pub fn python_env_kernel_specifications(
             .chain(toolchains.toolchains)
             .map(|toolchain| {
                 background_executor.spawn(async move {
+                    // For remote projects, we assume python is available assuming toolchain is reported.
+                    // We can skip the `ipykernel` check or run it remotely.
```
rgbkrk, line 388: We probably do still need to do some kind of ipykernel check. I think in some cases people want the same flow as VS Code where they can install ipykernel. That doesn't always work in every env (oddly even in my `uv sync` created `venv`s it has trouble).

crates/repl/src/kernels/ssh_kernel.rs
```diff
@@ -0,0 +1,288 @@
...
+                    while let Ok(message) = iopub_socket.read().await {
+                        session
+                            .update_in(cx, |session, window, cx| {
+                                session.route(&message, window, cx);
+                            })
+                            .log_err();
+                    }
+                }
+            })
+            .detach();
```
rgbkrk: This again reminds me we need `.split()` on zmq.rs so we don't have to create so many background processes and can send while receiving on all the channels.

crates/remote_server/src/headless_project.rs
MostlyKIGuess: I was hesitant to wire up full remote kernel discovery for now, felt like it might bloat this PR too much. 
Do you think we should block this until we have proper spec listing, or is the python3 fallback okay for an initial merge?

crates/repl/src/kernels/mod.rs
MostlyKIGuess: Should we add a prompt to install ipykernel if the selected kernel is missing it? that sort of reflects the notification workflow from VSCode.. I would want to investigate why it's not working on uv venv..

crates/repl/src/kernels/ssh_kernel.rs
MostlyKIGuess: Yeah, had to fight socket ownership here, once zmq adds split this would allow us to drop multiple spawners

crates/project/src/toolchain_store.rs
```diff
@@ -684,7 +684,14 @@ impl RemoteToolchainStore {
...
                 .await
-                .log_err()?;
+                .inspect_err(|e| {
+                    log::warn!(
+                        "RemoteToolchainStore::list_toolchains: RPC failed for language {}: {:?}",
+                        language_name,
+                        e
+                    );
+                })
+                .ok()?;
```
Veykril: This should remain a log_err imo

crates/remote_server/src/headless_project.rs
Veykril: Note that `python3` won't work on windows by default as its just `python` there

crates/remote_server/src/headless_project.rs
```diff
@@ -788,6 +795,92 @@ impl HeadlessProject {
...
+    async fn handle_kill_kernel(
+        this: Entity<Self>,
+        envelope: TypedEnvelope<proto::KillKernel>,
+        mut cx: AsyncApp,
+    ) -> Result<proto::Ack> {
+        let kernel_id = envelope.payload.kernel_id;
+        let child = this.update(&mut cx, |this, _| this.kernels.remove(&kernel_id));
+        if let Some(mut child) = child {
+            child.kill().log_err();
+            let _ = child.status().await; // Perhaps kill should be enough, should we wait?
```
Veykril: `kill` should suffice, no need to wait since we don't make any use of the return value here

crates/repl/src/components/kernel_options.rs
```diff
@@ -161,6 +161,8 @@ impl PickerDelegate for KernelPickerDelegate {
                 "Remote",
                 Some(truncate_path(&kernelspec.path(), 42)),
             ),
+            KernelSpecification::SshRemote(_) => (kernelspec.name(), "SSH Remote", None),
+            KernelSpecification::WslRemote(_) => (kernelspec.name(), "WSL Remote", None),
```
Veykril: Why are these separate from `Remote`?

Veykril (commented):
This does not seem to support docker and also I don't quite understand why we have to special case between different remote types here. Ideally all the specifics of ssh and wsl should be hidden away behind our `RemoteConnectionOptions` and such. (I have only skimmed over the ssh/wsl modules here for now, so do tell the reasons for the split here)

I am also confused by the fact that we now have a `Remote` variant but also an ssh and wsl remote variant each

crates/remote_server/src/headless_project.rs
MostlyKIGuess: Alright, I will remove this

Veykril:
Yea lets rename that variant then, and yes i the dev-containers. Absolutely fine to keep that out of this PR for now though

MostlyKIGuess:
Changed the original  body to not reflect the NixOS dev build fix and Jupyter View additions as they are dev and behind feature flag respectively.

crates/repl/src/kernels/mod.rs
rgbkrk: I'm sorry I meant that _VS Code_ runs into a problem with a fresh `uv sync`'ed repo and is unable to install into it. Maybe VS Code has to detect that it's a `uv` backed repo that needs to use `uv pip install ipykernel` (or `uv add` if intended to be kept).

crates/remote_server/src/headless_project.rs
MostlyKIGuess: Handled the python3 by having both python and python3 and removed this

crates/repl/src/kernels/ssh_kernel.rs
rgbkrk: I started on that here: https://github.com/zeromq/zmq.rs/pull/194

Probably easier to maintain in runtimed, zmq.rs, and even Zed to just use the Tokio setup that Zed now has inside (my work predates that).

rgbkrk (approved):

crates/repl/src/kernels/ssh_kernel.rs
rgbkrk: `.split()` is now supported in here as of https://github.com/zed-industries/zed/pull/48823 so this should nicely clean up.

crates/repl/src/kernels/ssh_kernel.rs
```diff
@@ -140,94 +143,99 @@ impl SshRunningKernel {
...
+                        if let Err(err) = result {
+                            session.update(cx, |session, cx| {
+                                session.kernel_errored(
+                                    format!("handling failed for {name}: {err}"),
+                                    cx,
+                                );
+                                cx.notify();
+                            });
+                        }
+                    }
```
rgbkrk: Looking over all this repeated code from the native kernels makes me wonder where this can be shared so we're not tripping over the networking bits multiple times.

MostlyKIGuess: Should I add like a start kernel tasks function inside mod.rs so instead of task spawning in all the kernels we can just use that, and it can output JupyterMessage and a supervisor task that keeps it running?

rgbkrk: I'm not sure what the interface should be but I'm happy to look tomorrow. Feel free to experiment and I'll read it again later.

crates/remote/src/transport/ssh.rs
```diff
@@ -1223,6 +1223,10 @@ impl SshSocket {
                 "ControlMaster=no".to_string(),
                 "-o".to_string(),
                 format!("ControlPath={}", self.socket_path.display()),
+                "-o".to_string(),
+                "BatchMode=yes".to_string(),
+                "-o".to_string(),
+                "ConnectTimeout=10".to_string(),
```
Veykril: I worry about this change, we haven't needed this until now for anything else in the ssh layer so I fear this may cause unexpected issues. Do we need this?

crates/zed/src/zed/quick_action_bar/repl_menu.rs
```diff
@@ -38,13 +38,16 @@ impl QuickActionBar {
...
         let editor = self.active_editor()?;
 
-        let is_local_project = editor
+        let is_valid_project = editor
             .read(cx)
             .workspace()
-            .map(|workspace| workspace.read(cx).project().read(cx).is_local())
+            .map(|workspace| {
+                let project = workspace.read(cx).project().read(cx);
+                project.is_local() || project.is_remote()
```
Veykril: This is always true as `is_remote` is just `!is_local`, I think you want `!project.is_via_collab()` (the methods here need some better documentation ....)

Veykril (commented):
one more thing but other lgtm

crates/remote/src/transport/ssh.rs
MostlyKIGuess: hmm, I kept this as my connection was unstable, not sure if we need it. Anyhow, if the kernel hangs, we can handle timeouts at the `SshRunningKernel` level later. I was able to make it work without it anyways

crates/zed/src/zed/quick_action_bar/repl_menu.rs
MostlyKIGuess: Yeah, thank you for this! Merged laxly when repl menu was added to upstream,  didn't realize this was A or not A by making this.

Veykril (approved):
Thanks!

crates/zed/src/zed/quick_action_bar/repl_menu.rs
MostlyKIGuess: will make the same changes on repl session UI as well

PR #47902, opened 2026-01-28
agent: Add project-level `disable_ai` setting

oliverbarnes:
Closes #47854

Move `disable_ai` from root settings to `ProjectSettingsContent` to enable per-project AI configuration via `.zed/settings.json`.

- Update settings UI to allow viewing/editing at both user and project level
- Update editor to check project-level settings for edit predictions and context menus
- Prevent MCP servers from starting when AI is disabled at project level

Note: SaturatingBool ensures that if user globally disables AI, projects cannot re-enable it. Projects can only further restrict AI, not grant it.

Release Notes:

- Updated settings to allow disabling AI for a specific project


probably-neb:
Great work @oliverbarnes! I've pushed up a commit that adds a second helper to the `DisableAiSettings`, just because I realized we were doing the `buffer.read(cx).file()` thing everywhere. It all seems to work on my end, once you give me the word I'd be happy to merge this

davc0n:
This is clear, my question was about the behavior, not the implementation.

Let's consider three possible scenarios:
**a)** globally off, but can enable on specific projects
**b)** globally on, but can disable on specifc projects
**c)** forced off, cannot enable (global)

The current implementation covers scenarios **b** and **c**.
I think the behavior of being allowed to override the option per-project only to turn **off** the feature, but not turn **on**, is not intuitive.

Using different values such as `on | off | force_off` all the scenarios would be covered.
Alternatively a seperate setting could be used to force off.

I was wondering if those options have been considered and then discarded for some reasons.

chidniu:
> @davc0n those are also valid possibilities, I hadn't considered those when implementing this.
> 
> But like I explained to Garth above, I think expanding the functionality needs some discussion and a decision from the Zed team, and the best way is to kick it off is as a [feature request](https://github.com/zed-industries/zed/discussions/categories/feature-requests). Another option is to post on Discord to see what the community thinks there.
> 
> Comments on merged PRs aren't very likely to get noticed.

I opened the discussion under #56340.

PR #47918, opened 2026-01-29
rust: Highlight enum variants as variant

my4ng:
Improve #45546, by changing the capture to be variant instead of type.

Release Notes:

- Improved Rust enum variant highlighting as variant instead of type.

Before (note variants have the same style as `Foo`):
image

After:
image

### DIFF
--- a/crates/languages/src/rust.rs
+++ b/crates/languages/src/rust.rs
@@ -1807,6 +1807,7 @@ mod tests {
             ("keyword", Hsla::default()),
             ("function", Hsla::default()),
             ("property", Hsla::default()),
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (approved):
Thank you.

SomeoneToIgnore (approved):
Thanks a lot for a good fix.

maxbrunsfeld:
I don't think we want this change, because variants typically can't be distinguished from types based on syntax alone. For example, `Some` is a variant. This change will introduce inconsistent coloring between the *definition* of a variant, and usages of that variant.

PR #47967, opened 2026-01-29
command_palette: Add option to set language

zertosh:
Add a new `language_selector.show_in_command_palette` setting that, when
enabled, populates the command palette with "language: {name}" entries
for every available language. This behaves the same as the Language Selector
but rather than being its own action with its own picker, it's part
of the command palette.

This is useful for those who prefer to go through the command palette
for everything. Although the Language Selector does a similar thing,
it requires knowing and calling a separate action.

command-palette-1
command-palette-2
settings

Release Notes:

- Added `language_selector.show_in_command_palette` setting to add the language selector entries to the command palette.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9166,6 +9166,7 @@ name = "language_selector"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "command_palette_hooks",
  "editor",
  "file_icons",
  "fuzzy",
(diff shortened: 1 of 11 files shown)


SomeoneToIgnore:
Sorry for coming that late: I hoped someone else comes by and voices their opinion.

I have polled a few people internally and seems that we do not want this now: overall, I agree that a combined panel is good, but for that we need a bit more UI amenities (some folding lists with sublists in that panel, etc.).

With the current impl, it gets quite crowded in this panel and given that we have the language selector already that can be used with keyboard only, we'll stick to the current state of things for some time.

PR #47971, opened 2026-01-29
agent_ui: Add additional telemetry for thread errors

katie-z-geer:
This PR extends error telemetry coverage in the agent panel, building on #46836 which added telemetry for thread errors.

`"Agent Panel Error Shown"` (extended)

Now also fires for:

- **Load errors** - when an agent fails to start 
- **Configuration errors** - when LLM provider setup is incomplete 

`"Agent Token Limit Warning"` (new event)

Fires when a thread approaches or exceeds its token/context limit. This is tracked separately from errors because it's an informational warning, not a failure.

## Implementation Notes

- All telemetry fires **once per error occurrence**, not on every render
- Load error telemetry fires in `handle_load_error()` when the error is first received
- Token limit telemetry uses a flag to track state transitions (Normal → Warning → Exceeded)
- Configuration error telemetry tracks the last emitted error kind to avoid duplicates

Release Notes:


crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -4818,6 +4820,83 @@ impl AcpThreadView {
...
+
+        let agent_name = self.agent.name();
+
+        telemetry::event!(
+            "Agent Panel Error Shown",
+            agent = agent_name,
+            kind = error_kind,
+            message = message,
+        );
+    }
```
mikebenfield: I think the match statement above is basically just duplicating the Display implementation of LoadError that already exists. The body of this function can be replaced with:
```suggestion
        let error_kind = match error {
            LoadError::Unsupported { .. } => "unsupported",
            LoadError::FailedToInstall(_) => "failed_to_install",
            LoadError::Exited { .. } => "exited",
            LoadError::Other(_) => "other",
        };

        let agent_name = self.agent.name();

        telemetry::event!(
            "Agent Panel Error Shown",
            agent = agent_name,
            kind = error_kind,
            message = error.to_string(),
        );
```

crates/agent_ui/src/acp/thread_view.rs
```diff
@@ -4818,6 +4820,83 @@ impl AcpThreadView {
...
+            }
+            acp_thread::TokenUsageRatio::Warning => "warning",
+            acp_thread::TokenUsageRatio::Exceeded => "exceeded",
+        };
+
+        if active.token_limit_telemetry_emitted {
+            return;
+        }
+
+        active.token_limit_telemetry_emitted = true;
```
mikebenfield: Are we sure we don't want to emit token limit telemetry for a thread if we've already done so once? For instance might we want to emit once for a warning and later for exceeded? (I don't know; this might be the right thing to do. Just confirming.)

.gitignore
```diff
@@ -53,3 +53,5 @@ crates/docs_preprocessor/actions.json
 # Local documentation audit files
 /december-2025-releases.md
 /docs/december-2025-documentation-gaps.md
+docs/src/agent-error-context
+docs/src/agent-errors
```
mikebenfield: What are these two paths for?

mikebenfield (commented):
Overall looks fine; left some comments.

.gitignore
katie-z-geer: Theyre my notes for the agent, I'll actually trash these entirely.

katie-z-geer (commented):
all good

PR #40616, opened 2025-10-19
git_ui: Allow opening a file with the diff hunks expanded

FloppyDisco:
So i just discovered `editor::ExpandAllDiffHunks`

I have been really missing the ability to look at changes NOT in a multi buffer so i was very pleased to finally figure out that this is already possible in Zed.

i have seen alot of discussion/issues requesting this feature so i think it is safe to say i'm not the only one that is not aware it exists.

i think the wording in the docs could better communicate what this feature actually is, however, i think an even better way to show users that this feature exists would be to just put it in front of them.

In the `GitPanel`:
- `menu::Confirm` opens the project diff
- `menu::SecondaryConfirm` opens the selected file in a new editor.

I think it would be REALLY nice if opening a file with `SecondaryConfirm` opened the file with the diff hunks already expanded and scrolled the editor to the first hunk.

ideally i see this being toggle-able in settings something like `GitPanel - Open File with Diffs Expanded` or something. so the user could turn this off if they preferred.

I tried creating a new keybinding using the new `actions::Sequence`
it was something like:
```json
{
  "context": "GitPanel && ChangesList",
  "bindings": {
    "cmd-enter" : [ "actions::Sequence", ["menu:SecondaryConfirm", "editor::ToggleFocus", "editor::ExpandAllDiffHunks", "editor::GoToHunk"]]
  }
}
```
but the action sequence does not work. i think because opening the file is an async task.

i have a first attempt here, of just trying to get the diff hunks to expand after opening the file.
i tried to copy and paste the logic/structure as best i could from the confirm method in file_finder.rs:1432 

it compiles, but it does not work, and i do not have enough experience in rust or in this project to figure out anything further.

if anyone was interested in working on this with me i would enjoy learning more and i think this would be a nice way to showcase this tool!

PR #47989, opened 2026-01-30
git_ui: Open file on double-click in status list

cppcoffee:
Release Notes:

- Improved open file on double-click in status list

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -5066,7 +5066,7 @@ impl GitPanel {
                 cx.listener(move |this, event: &ClickEvent, window, cx| {
                     this.selected_entry = Some(ix);
                     cx.notify();
-                    if event.modifiers().secondary() {
+                    if event.click_count() > 1 || event.modifiers().secondary() {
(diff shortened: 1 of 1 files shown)


cppcoffee:
This change improves the user experience in the Git Panel by allowing users to open a file in the editor by double-clicking its entry in the status list.

Previously, opening a file required a secondary click (e.g., Cmd/Ctrl+Click). Single clicking continues to open the file in the diff view.

cppcoffee:
> Small feedback here: Since #40616 was shipped, I am missing a lot the old behaviour that is: secondary click just open the file **but without expanded diff hunks**. I also saw that in the linked PR, it was said that no configuration should be needed, but I personally find this very annoying. I extensively use this when working to navigate through changed files.
> 
> Anyhow, I am willing to PR to add a configuration, but one idea could also be to distinguish the 2 actions like so:
> 
> * double-click: just open the file
> * secondary click: open the file with expanded diff
> 
> wdyt?

Hi @AlexandreBonaventure , The original functionality remains unchanged; only the double-click behavior was modified to just open the file. This change has not been released yet.

PR #47997, opened 2026-01-30
agent_ui: Disable "Copy Selection" when no text is selected

cppcoffee:
Release Notes:

- Agent: Improved the "Copy Selection" right-click menu item by disabling it when there are no selections.


danilo-leal (approved):
Good catch, thank you!

PR #48029, opened 2026-01-30
gpui: Add support for NSWindow's representedFilename

MrMage:
This sets the accessibility document property (AXDocument) of the window, which other apps can use to understand what file the current window represents. Document-based apps on macOS are generally expected to set this property.

The document path is set via `Window::set_document_path()` which calls `setRepresentedFilename:` on the underlying NSWindow. The workspace updates this path in `update_window_title()` whenever the active item or project structure changes. For tests, instead of trying to somehow assemble a proper NSWindow, we store the document path on the window mock used for testing and check its value.

Motivation: I am the developer of Timing, an automatic time tracking app for Mac. Timing uses the `AXDocument` property (a standard property of most document-based app windows on macOS) to understand what document the user is working on. With this change, Timing is able to understand which directory the user is working in. Without this, Timing would only record the window title, i.e. the filename without information about the containing directory. I've had several users ask for better support for Zed.

Here's a screenshot of the macOS Accessibility Inspector showing the `AXDocument` property with this change. The UI of Zed itself does not change. However, in my dev build of Zed, the traffic lights are a bit too large and misaligned. However, this happens even when building `main`, so I assume it’s unrelated to my changes.

Screenshot 2026-01-30 at 16 18 25

Release Notes:

- Set the represented filename property of windows on macOS


MrMage:
@cla-bot check

MrMage:
Friendly ping :-)

MrMage:
@ChristopherBiscardi Were you able to access Zed debug builds in Accessibility Inspector using the steps I provided?

ChristopherBiscardi (approved):
@MrMage I was able to confirm the behavior using the release scripts in the repo, yes. I've also merged main in here and kicked off the tests.

ChristopherBiscardi:
Thank you for the work here @MrMage 🙏🏼

PR #48035, opened 2026-01-30
editor: Fix JoinLines with selection to end-of-line

### LINKED ISSUE #48030: JoinLines includes the next line if the selection is to the end of a line
### Reproduction steps

This is easiest to illustrate in vim/helix mode. ~Not entirely sure if it even affects non-modal editing.~ Also affects non-modal editing.

1. Make a multi-full-line selection. Visual line mode in vim, multiple `x`'s in helix, or invocations of `editor::SelectLine` (`ctrl-l` by default) will work.
2. invoke `editor::JoinLines` (`shift-j` in both vim and helix, `ctrl-shift-j` otherwise)

Current behavior: All selected lines plus the next non-selected line are joined.

Expected behavior: All selected lines are joined and the next non-selected line is left where it is.
Note: If the selection only includes a single line, the next line should still be joined. This bug is only related to multi-line selections.
Note2: If the selection doesn't include the full line, e.g. is made in non-line visual mode or helix select mode, things appear to work as expected.

My guess is that this is another instance of the selection line off-by-one errors that I seem to be fixing a lot of recently 😂 With full-line selections, the cursor is technically at the start of the *next* line, which is probably why it's getting erroneously included in the join. Since this is actually correct for the single-full-line-selection case, the fix will need to take that into account and will thus be slightly trickier, but (hopefully) only slightly.

Zed: v0.223.0+nightly (Zed Nightly) 
OS: Linux X11 nixos 26.05
Memory: 30.5 GiB
Architecture: x86_64
GPU: AMD Radeon RX 6800 XT (RADV NAVI21) || radv || Mesa 25.3.4

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

jrobsonchase:
Full line selections should only include the next line if there's a single line selected. Otherwise, if multiple lines are selected, only the selected lines should be joined.

Also updates the editor tests to include both of these cases.

Closes #48030 

Release Notes:

- Fixed `editor::JoinLines` when a multi-line selection includes the end of the final line

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -11304,6 +11304,15 @@ impl Editor {
             // would do nothing for single line selections individual cursors.
             let end = if selection.start.row == selection.end.row {
                 MultiBufferRow(selection.start.row + 1)
+            } else if selection.end.column == 0 {
+                // If the selection ends at the start of a line, it's logically at the end of the
+                // previous line (plus its newline).
+                // Don't include the end line unless there's only one line selected.
+                if selection.start.row + 1 == selection.end.row {
+                    MultiBufferRow(selection.end.row)
+                } else {
+                    MultiBufferRow(selection.end.row - 1)
+                }
             } else {
                 MultiBufferRow(selection.end.row)
             };
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -4837,6 +4837,32 @@ fn test_join_lines_with_single_selection(cx: &mut TestAppContext) {
             &[Point::new(0, 3)..Point::new(0, 3)]
         );
 
+        editor.undo(&Undo, window, cx);
+        assert_eq!(buffer.read(cx).text(), "aaa\nbbb\nccc\nddd\n\n");
+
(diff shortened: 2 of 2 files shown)


crates/editor/src/editor.rs
```diff
@@ -11110,6 +11110,15 @@ impl Editor {
             // would do nothing for single line selections individual cursors.
             let end = if selection.start.row == selection.end.row {
                 MultiBufferRow(selection.start.row + 1)
+            } else if selection.end.column == 0 {
+                // If the selection ends at the start of a line, it's logically at the end of the
+                // previous line (plus its newline).
+                // Don't include the end line unless there's only one line selected.
+                if selection.start.row + 1 == selection.end.row {
```
jrobsonchase, line 11311: This check might not be correct in the case of reverse selections.

Edit: Nope! looks like the selection logic will always flip start/end such that they're in the proper order if it's reversed, e.g. https://github.com/zed-industries/zed/blob/fa68aa770493fe82e90ba057671a3c9fd6391561/crates/editor/src/selections_collection.rs#L802-L805

SomeoneToIgnore (approved):
Nice, thank you!

PR #48063, opened 2026-01-31
languages: Support double quoted keys in YAML highlights

ozacod:
Before:
before

After:
after

Release Notes:

- Improved YAML highlights to support quoted keys


veganstraightedge:
thank you [ozacod](https://github.com/ozacod)!

PR #48064, opened 2026-01-31
languages: Add shebang highlight support for Bash

ozacod:
Highlights the shebang line as a directive instead of a comment.

Before:
before

After:
after

Release Notes:

- Added shebang highlight support for Bash


SomeoneToIgnore (approved):
Thank you!

PR #48067, opened 2026-01-31
languages: Improve function parameter highlighting for Bash

ozacod:
Before:
before

After: 
after

Release Notes:

- Improves function parameter highlighting for Bash


SomeoneToIgnore (approved):
Nice, thanks!

PR #48057, opened 2026-01-30
Introduce GPUI inline element

pkondzior:
Hi folks, I’d love a sanity check on whether this direction is a good fit for GPUI and worth taking to merge. I’ve built a full inline element that’s layout/styling‑equivalent to `div()`, but supports true inline flow (text + elements in one run, proper wrapping around inline boxes). This will let a lot of new futures to be built into the Markdown, Agent Panel text rendering or anything other that needs custom inline flow system.

Feature highlight:
- **Inline content model:** text, hard breaks, and inline child elements (with logical lengths for cursor/selection semantics).  
- **Style + interactivity parity:** `inline()` implements `Styled` + `InteractiveElement`, so it supports the same style refinements, hover/click, and event wiring as `div()`.  
- **Per‑run text styling:** explicit `TextRun` support for syntax‑like styling via `text_runs`/`runs`.  
- **Precise layout + hit‑testing:** `InlineLayout` exposes text/plain_text, word ranges, bounds, and index↔position mapping (caret positioning and selection are supported).  
- **Truncation and ellipsis:** `truncate`, `text_overflow`, and `line_clamp` are implemented end‑to‑end, including ellipsis styling and clipping for inline boxes.  
- **Performance plumbing:** inline layout is cached (`InlineLayoutCache`) and there’s a benchmark comparing inline vs div for mixed text/box content.  
- **Inspector/debug support:** inline layout carries line/box counts and truncation info.

Anything `div()` can do, `inline()` can do as well. The key difference is that `div()` can’t mix text and other div elements on the same line with wrapping, while `inline()` can.

```rust
div()
      .border_1()
      .border_color(gpui::black())
      .p(px(8.0))
      .flex()
      .flex_col()
      .gap(px(4.0))
      .child("Test 8: Interactive inline with hover and click")
      .child({
          inline()
              .text("Here is some text with a ")
              .border_1()
              .border_color(gpui::black())
              .child(
                  inline()
                      .h(px(40.0))
                      .border_1()
                      .border_color(gpui::blue())
                      .text_color(gpui::blue())
                      .text("clickable link")
                      .truncate()
                      .into_element()
                      .id("interactive-link")
                      .hover(|style| {
                          style
                              .bg(gpui::blue().opacity(0.1))
                              .border_color(gpui::red())
                              .cursor_pointer()
                      })
                      .on_click(|_event, _window, _cx| {
                          println!("Interactive inline clicked!");
                      })
              )
              .text(" embedded in the flow. Try hovering and clicking!")
[...]

PR #48054, opened 2026-01-30
Make mention creases actionable

pkondzior:
Summary:

Mention Crease buttons are clickable with proper tooltip being used for all kind of mentions. 

https://github.com/user-attachments/assets/d4b8ca17-d8b1-4654-902d-1d9c156cadcb

Changes:

### 1) Centralized mention/link navigation
- Added a dedicated navigation module:
  - `open_workspace_link(...)`
  - `open_workspace_mention(...)`
- Removed duplicated link-opening logic from thread view internals.

### 2) ACP thread views now use shared navigation
- ACP markdown/link click paths now call the shared navigation helper.
- Cleaned up unused imports in ACP thread view files.

### 3) Mention URI is now preserved through crease creation
- Mention insertion APIs now carry `MentionUri` explicitly.
- Updated all callsites (message editor, completion provider, mention set).

### 4) `MentionCrease` refactor (UI-only + callback-based)
- `MentionCrease` now stores `mention` and optional `icon`/`label`.
- Added callback-style `.on_click(...)` API.
- Default icon/label/tooltip are derived from `MentionUri`.

### 5) MentionSet render path updates
- Fold/button rendering now uses the new `MentionCrease` API.
- Extracted duplicated click handler into one helper:
  - `open_mention(...)`

### 6) Inline prompt editor keeps mention identity
- `MessageCrease` now stores `MentionUri`.
- During extraction/reinsertion of creases, mention mapping is preserved.
- This keeps creases actionable after editor unlink/rebuild.

### 7) Behavioral outcome
- Mention creases are actionable using the stored `MentionUri`.
- Navigation behavior is centralized and consistent.
- Non-openable mention kinds still safely no-op in navigation.
- Tooltips for actionable mentions are presented

Related to #37563

Release Notes:
- Added clicable mention creases with centralized link navigation, so mention-derived creases now retain `MentionUri` metadata and reliably open their targets (file/symbol/thread/rule/link) across ACP and inline prompt editors.

Issue #37563, opened 2025-09-04
Make mentions of project files clickable


pcholakov:
When Claude makes plans, it mentions file locations within the project, e.g.:

> Files to modify:
> 1. `lib/core.ts:94` - Change foo from `X` to `Y`

It would be great if these locations were clickable links that opened the relevant file (and line or symbol) in the Zed editor.

stefanosandes:
Yes, please!! Links for project files generated in agent panel thrown an error when clicked: The application cannot be opened.

deadnews:
`[@filename](file:///absolute/path/to/filename#L42)` works for me

willlma:
Sure, but the LLMs often just write `relative/path/to/filename` instead. I agree that they should be rendered clickable. Thanks

rubin-alldev:
My Codex integration had the problem of broken links as well, and I fixed it by adding this instruction to AGENTS.md (they render not as chips but in default HTML link style, but at least they aren't broken anymore).:

```markdown
<ZED AGENT ONLY>: If you mention a file in your response, use [%filename%.%extension%](file://%filepath%.%extension%) syntax to create a link to it, where it makes sense.
If you mention a symbol (function, variable etc.), use this format: [%symbol_name%](file://%filepath%.%extension%#L%line_number%)
```

Beat them with their weapon, as they say

PR #48074, opened 2026-01-31
Markdown link renderer and Thread View Mention links 

pkondzior:
This PR adds native-looking Mentions in the Agent Panel Thread view by introducing a custom inline Markdown link renderer.

**Dependency**

*   This work **depends heavily on** [https://github.com/zed-industries/zed/pull/48057](https://github.com/zed-industries/zed/pull/48057), because Mentions require the new **inline layout system** that can wrap arbitrary UI elements inside flowing text.
*   Because that dependency does not live in the Zed repo, I can’t reasonably open this PR with a base branch outside of `zed-industries/zed`

Demo:  

https://github.com/user-attachments/assets/cfafe38f-0684-466a-867d-0330ed573db5

**Also included**

*   This PR also includes [https://github.com/zed-industries/zed/pull/48054](https://github.com/zed-industries/zed/pull/48054). I extracted the `MentionCrease` changes into this MVP PR.
*   The `MentionCrease` changes are **independent**: they **do not depend on** the Markdown link renderer or the gpui inline layout work.

List of changes:
- Added a **custom markdown link rendering pipeline** via `LinkRenderer` and `LinkRenderContext`, allowing callers to render links as arbitrary elements instead of plain inline text.  
- Introduced `LinkLabel` to preserve mixed inline content (text + child elements) inside links, so custom renderers can reproduce the full link label.  
- Reworked the markdown inline builder to track nested inline content using `PendingInline`/`PendingItem` and a `link_stack`, then render links at close time via `pop_link`.  
- `MarkdownElement::link_renderer(...)` now lets the Agent thread view provide a custom renderer; mention links are rendered as `MentionCrease` with active/selected styling.  
- `AcpThreadView::link_renderer(...)` parses links into `MentionUri` and renders mentions as clickable buttons; non‑mention links fall back to the default link element.  

Related #37563

Release Notes:

- Improved Agent Thread mention links render as Zed-style clickable buttons with tooltips, powered by a custom markdown link renderer for more consistent visuals and smoother navigation.


xyt-dev:
This is a very useful feature. Looking forward to seeing it implemented soon!

PR #48082, opened 2026-01-31
docs: Add section on disabling `autoUseWorkspaceTsdk` per-project

stromseng:
Add section on configuring workspace TypeScript version in vtsls.

This is important for libraries like [effect](https://effect.website/) where they ship their own LSP patches.

Release Notes:

- N/A

### DIFF
--- a/docs/src/languages/typescript.md
+++ b/docs/src/languages/typescript.md
@@ -184,6 +184,23 @@ When using `vtsls`:
 }
 ```
 
+## Workspace TypeScript version
+
+```json [settings]
+{
+  "lsp": {
+    "vtsls": {
+      "initialization_options": {
+        "vtsls": {
+          "autoUseWorkspaceTsdk": true,
+        },
+      },
+    },
+  },
+}
+
+```
+
 ## Debugging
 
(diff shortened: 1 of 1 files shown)


stromseng:
@cla-bot check

yeskunall:
We already set [`autoUseWorkspaceTsdk`](https://github.com/zed-industries/zed/blob/00ea07cfcdd883f23f3f813f2152aadaa8d55916/crates/languages/src/vtsls.rs#L287-L288).

image

Are you not seeing different behavior? If yes, can you open an issue for that?

yeskunall (commented):
I don't believe this is the right way. Have you tried setting it this way and gotten it to work? I couldn't. I believe this is what you're looking for:

```jsonc
{
  "lsp": {
    "vtsls": {
      "settings": {
        "vtsls": {
          "autoUseWorkspaceTsdk": true
        }
      }
    }
  }
}

```

stromseng:
Ahh I'm terribly sorry.  

I assume i must've messed up somewhere along the way, and it worked after i had added the config. Although it was probably a different reason for the LSP not showing effect errors. 

Could i suggest documenting this default behavior in the docs then?

PR #48084, opened 2026-01-31
terminal: Prompt before closing tabs with running processes

### LINKED ISSUE #20051: confirmation on closing terminal that has running process
### Description

i often press the shortcut to close tab while thinking i am focus on the workspace but find myself closing a terminal tab instead.

it would be good to have an option to ask for confirmation to close a terminal tab when there is a running process (ssh, build command running, etc.) to avoid any issue when moving a little too fast

VSCode has
```json
{
    "terminal.integrated.confirmOnExit": "hasChildProcesses"
}
```
settings for this.

imfing:
Adds a `terminal.confirm_on_kill` setting that prompts before closing terminal tabs that have child processes

- `never` (default) – no confirmation
- `always` – prompt when a non-shell process is running

Closes #20051

Equivalent to VS Code's `terminal.integrated.confirmOnKill`, (without the `editor`/`panel` options in this attempt), the setting can be imported from VS Code directly.

Screenshots:

image

image

https://github.com/user-attachments/assets/c6224939-b91a-4374-b7c1-c85f5589898c

---

Note: this PR does not implement the editor level quit confirmation (like VS Code `terminal.integrated.confirmOnExit`) which detects terminal processes when quiting the entire editor.

Would appreciate the team's thoughts on this.

Release Notes:

- Added `terminal.confirm_on_kill` setting to prompt before closing terminal tabs with running processes


crates/workspace/src/pane.rs
```diff
@@ -1944,6 +1944,29 @@ impl Pane {
...
+                        let buttons = [
+                            PromptButton::ok(confirm_button),
+                            PromptButton::cancel(cancel_button),
+                        ];
+                        let detail = detail.as_ref().map(|detail| detail.as_ref());
+                        window.prompt(level, message.as_ref(), detail, &buttons, cx)
+                    })?;
+                    match answer.await {
+                        Ok(0) => {}
+                        Ok(1..) | Err(_) => return Ok(()),
```
SomeoneToIgnore: Is it correct, given that there could be more items to close?
It seems better to close as much as possible, hence here should be `should_close = false; should_save = false;` or similar?

crates/terminal_view/src/terminal_view.rs
```diff
@@ -1304,6 +1305,29 @@ impl Item for TerminalView {
         }))))
     }
 
+    fn close_confirmation(&self, cx: &App) -> Option<CloseConfirmation> {
+        let confirm_on_kill = TerminalSettings::get_global(cx).confirm_on_kill;
+        let terminal = self.terminal().read(cx);
```
SomeoneToIgnore: NIT: could be moved into `Always` match arm block, as not needed anywhere else.

SomeoneToIgnore (commented):
Sorry for taking so long for the review, this looks really nice.

Would you be able to rebase and discuss and maybe fix the small issue with saving/closing?

crates/workspace/src/pane.rs
```diff
@@ -1957,6 +1957,32 @@ impl Pane {
...
+                        confirm_button,
+                        cancel_button,
+                    } = confirmation;
+                    let answer = pane.update_in(cx, |_, window, cx| {
+                        let buttons = [
+                            PromptButton::ok(confirm_button),
+                            PromptButton::cancel(cancel_button),
+                        ];
+                        let detail = detail.as_ref().map(|detail| detail.as_ref());
+                        window.prompt(level, message.as_ref(), detail, &buttons, cx)
```
SomeoneToIgnore: Thinking of it more: we will prompt 10 times for 10 tabs with subprocesses now with this code, won't we?

Flashing that many prompts seems wrong, esp. given the fact that we deduplicate the saving: 

Image

To replicate the behavior here, we have to find some basic key, e.g. `level`, or `(level, message)` and show al list of all details, it seems?

crates/terminal_view/src/terminal_view.rs
```diff
@@ -1317,6 +1318,29 @@ impl Item for TerminalView {
...
+
+        match confirm_on_kill {
+            TerminalConfirmOnKill::Never => None,
+            TerminalConfirmOnKill::Always => {
+                let terminal = self.terminal().read(cx);
+                if !terminal.has_running_process() {
+                    return None;
+                }
+                let title = terminal.title(true);
+                let message = "Do you want to terminate the active terminal session?";
```
SomeoneToIgnore:
```suggestion
                let message = "Terminate the active terminal session?";
```

crates/terminal/src/terminal.rs
```diff
@@ -2207,6 +2207,48 @@ impl Terminal {
...
+        };
+
+        !self.is_shell_process(process_info)
+    }
+
+    fn is_shell_process(&self, process_info: &ProcessInfo) -> bool {
+        let Some(shell_name) = self.shell_process_name() else {
+            return false;
+        };
+        Self::normalize_process_name(&process_info.name)
```
SomeoneToIgnore, line 2233: Considering this place further: 

* does not both shell process name and process info name have the same notation?
Seems odd that we have to normalize them, can you share more details on that?

* is it really what VSCode does, semantically?
I'd expect a `sub process` to mean also things like `for i in $(seq 0 10); do sleep 1; echo "hey $i"; done &` , but this code won't warn about them, right?

Is there a way to use `PtyProcessInfo` or something adjacent about the process?
I would expect so, given we it provides a `pid` to us.

Curious how this all works for Windows too.

SomeoneToIgnore (commented):
Thank you, a few more general notes.

crates/workspace/src/pane.rs
imfing: that makes sense, I was thinking users might have just a few buffers open, which makes prompting for each of them felt acceptable.

I agree it is much better to stay consistent with the grouped file saving prompt. 

will push an update to align with that.

imfing: pushed an update to group the close confirmations, now it looks like this:

https://github.com/user-attachments/assets/f3d90d06-197a-42ff-90d7-a8a053b89bda

SomeoneToIgnore:
Thank you for the PR, I think we have most of the code lined up, and the task handling is what's left to do.
Seems not that simple though, given that we have to deal with multiple platforms here and consider what to do in the remote cases.

Given that there had not been any activity in the last 3 weeks, will close this PR as stale to keep the PR queue cleaner.

imfing:
> Thank you for the PR, I think we have most of the code lined up, and the task handling is what's left to do.
> 
> Seems not that simple though, given that we have to deal with multiple platforms here and consider what to do in the remote cases.
> 
> 
> 
> Given that there had not been any activity in the last 3 weeks, will close this PR as stale to keep the PR queue cleaner.


Thanks for the follow-up, and no worries at all about closing the PR.

As I dug into the task handling, I found that reliable foreground detection across platforms requires some non-trivial changes.

Since this has grown beyond my current bandwidth, I'm happy for the maintainers or another contributor to use this PR as a starting point.

Thanks again for the review and the guidance!

PR #48107, opened 2026-01-31
workspace: Add copy button to ErrorMessagePrompt popup

deralaxo:
My own uncreated issue. Super simple QOL change.
Just needed to copy errors and the only other place I'm aware of is console. 

The text inside the error popup is not selectable (at least on macos)

before:
image

after:
image

Release Notes:

- Added `CopyButton` to `ErrorMessagePrompt`


danilo-leal (approved):
Thank you!

PR #48118, opened 2026-02-01
linux: Forward env vars starting with ZED_ to flatpak-spawn

refi64:
Otherwise, it's not easily possible to set *any* of these variables when running Zed.

Release Notes:

- Added forwarding of `ZED_*` environment variables when using the Flatpak


refi64:
It's not really about it being disabled..,explicitly at least. Consider that the launch flow, when not staying within the sandbox, is:

flatpak -> zed (inside sandbox) -> `flatpak-spawn --host` -> zed (outside sandbox)

Flatpak passes most environment variables into the sandbox, so `zed (inside sandbox)` gets them normally. However, flatpak-spawn works by giving the command, along with the *explicitly given environment variables*, to an external service that runs the outer zed. There's no normal environment inheritance at that point, and thus anything you want to forward needs to be done manually.

PR #48119, opened 2026-02-01
ollama: Don't override model's default stop tokens

littleKitchen:
## Summary

When no stop tokens are provided, Zed was sending an empty array (`"stop": []`) to Ollama. This caused Ollama to override the model's default stop tokens (defined in its Modelfile) with nothing, resulting in models like rnj-1:8b generating infinitely with literal stop tokens appearing in the output.

## Problem

Models with custom stop tokens in their Modelfile (like `<|eot_id|>` for rnj-1:8b) would generate forever because:

1. Agent thread creates request with `stop: Vec::new()` (empty)
2. Ollama provider converts this to `stop: Some(vec![])`
3. Serializes as `"stop": []` in JSON
4. Ollama interprets this as "override default stop tokens with nothing"
5. Model generates forever, outputting stop tokens as literal text

## Solution

1. In `crates/language_models/src/provider/ollama.rs`:
   - Only send `stop` when explicitly provided
   - When empty, use `None` so the field is omitted from JSON
   
2. In `crates/ollama/src/ollama.rs`:
   - Add `#[serde(skip_serializing_if = "Option::is_none")]` to all `ChatOptions` fields
   - Ensures `None` values are omitted, not serialized as `null`

## Testing

Added 4 new tests in `crates/ollama/src/ollama.rs`:
- `test_chat_options_serialization`: Verifies None fields are omitted
- `test_chat_request_with_stop_tokens`: Verifies stop tokens are serialized when provided
- `test_chat_request_without_stop_tokens_omits_field`: Verifies empty stop is omitted

All 11 ollama tests pass, plus 1 language_models ollama test.

Fixes #47798

Release Notes:

- Fixed Ollama models with custom stop tokens generating infinitely by not overriding model defaults when no stop tokens are specified.

### DIFF
--- a/crates/language_models/src/provider/ollama.rs
+++ b/crates/language_models/src/provider/ollama.rs
@@ -400,7 +400,14 @@ impl OllamaLanguageModel {
             stream: true,
             options: Some(ChatOptions {
                 num_ctx: Some(self.model.max_tokens),
-                stop: Some(request.stop),
+                // Only send stop tokens if explicitly provided. When empty/None,
+                // Ollama will use the model's default stop tokens from its Modelfile.
+                // Sending an empty array would override and disable the defaults.
+                stop: if request.stop.is_empty() {
+                    None
+                } else {
+                    Some(request.stop)
+                },
                 temperature: request.temperature.or(Some(1.0)),
                 ..Default::default()
             }),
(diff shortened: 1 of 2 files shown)


benbrandt (approved):
Makes sense, good find!

PR #48143, opened 2026-02-01
Add LaTeX rendering to Markdown preview

MostlyKIGuess:
Closes #40813 

Continuing from #44525 

- add the `latex_render` crate using ReX to render LaTeX.
- supports inline and display styles, bundles the Latin Modern Math font, and implements a cache.

Working based on https://github.com/zed-industries/zed/pull/44525#issuecomment-3739372864

There are few issues which need help right now.. I am not exactly sure how we would fix the font tearing and relative sizes, I have tried to keep all the magic numbers on top of all the files to speed up work on this issue.

Release Notes:
- Ability to  render LaTeX in markdown preview.


pkondzior:
Awesome work! 

Having inline() from #48057 would significantly improve the LaTeX renderer. It would let us render inline paragraphs properly, so blocks of text between equations would wrap coherently. Right now, a lot of content overflows especially div lines that exceed their container width and switching to inline rendering should fix that.

Screenshot 2026-02-01 at 18 23 05

Arcitec:
Continued in #54668

Also rewritten by someone in #57339

PR #48157, opened 2026-02-01
repl: Use wry to render HTML outputs

rgbkrk:
WIP. In support of #9778

Builds will require more system dependencies in order for this to work. I've only built it on MacOS.

image

Major caveats:

* Cannot use keybindings to copy text. Can right click to copy text.
* The webview will overlay any other GPUI element, like if the agent panel is open and stretched out over the webview.
* No interactive plotly plot, though I bet it's looking for JupyterJavascript. I'll have to set up an inspector.

image

Closes #ISSUE

Release Notes:

- Added HTML Rendering support to REPL outputs


marcocondrache:
Just my two cents: the `wry` crate is fairly heavy to compile, and it would affect everyone working on Zed. Maybe it could live behind a feature? Or, since HTML views are something the team is already planning to support, it might make sense to wait for that and reuse whatever comes out of it

rgbkrk:
> Just my two cents: the wry crate is fairly heavy to compile, and it would affect everyone working on Zed. Maybe it could live behind a feature? Or, since HTML views are something the team is already planning to support, it might make sense to wait for that and reuse whatever comes out of it

Luckily this is just a draft and I'm not pushing it hard. I'd love to see core Zed folks implement a webview for Zed. Until then I'm happy with showing people where we could take things and what it unlocks.

Issue #16727, opened 2024-08-23
Select text in markdown preview


atmelmicro:
### Check for existing issues

- [X] Completed

### Describe the feature

I want to be able to highlight/select text in the markdown preview to copy it.

bitfield:
This is a really useful feature in VS Code: you can preview your Markdown document, and then copy and paste the formatted preview into Google Docs. It would be great to be able to do this in Zed too.

im-0:
Just encountered this while trying to copy the markdown doc as a plain text (without the MD syntax, links, etc).

It’s frustrating that such a basic feature is missing.

schallm:
I am unable to select and copy formatted text/images from the Markdown Preview window.  I use this a lot from VS Code when so I can type in Markdown, but send an email in formatted html.

bobby-huang:
Yes please, going to stay on cursor for now.

KuSh:
> Would fixing this enable searching in Markdown previews as well, or is that a separate issue?
> 
> Ctrl+F does nothing and there's no context menu or button to search. This makes searching for bug fixes in long release notes impossible.

Not magically, there's no search in the agent panel neither, but having it implemented in it would simplify having it here too (if markdown_preview is migrated to the markdown crate).

olejorgenb:
This is great, but it lack the double-click to select words from https://github.com/zed-industries/zed/issues/43354

AnPod:
Hi,

To add to the current discussion regarding issue #16727, I've noticed that
while copy-paste now works in the new version, it does not carry over the
formatting.

In VS Code, the formatting is preserved when pasting into documentation
tools like Confluence. It would be helpful if Zed could match this behavior.

Best regards,

Andrej Podgoršek

mobob:
Confirmed what @AnPod, says, this should be reopened - for me this is all moot if selecting the markdown preview text doesn't retain the formatting. Might as well just copy the raw text?

The use case is to see it rendered, then dump into gmail / notes / etc..

KaKi87:
Hello,
Any news on double-click ?
Thanks

PR #48162, opened 2026-02-01
Streamline Markdown notification rendering

KuSh:
Brought the Markdown notification rendering up to date with how Markdown is used in the Agent panel and REPL.
Removed async parsing overhead and simplified the code by using the markdown crate API directly.

This is preparatory work, my ultimate goal is to make markdown_preview use markdown to parse and render md to [fix not being able to select and copy content](https://github.com/zed-industries/zed/issues/16727)

|Current|Now|
|-|-|
|Capture d’écran du 2026-02-01 22-39-39|Capture d’écran du 2026-02-01 22-37-16|

Release Notes:

- N/A


danilo-leal (approved):
Nice, thank you so much!

PR #48164, opened 2026-02-01
title_bar: Add per-project background color setting

Vany:
I have multiple Zed windows open, and I want to be able to tell them apart.

This adds a `background` option to the `title_bar` settings that can be configured per-project in `.zed/settings.json`.
I often work on multiple projects and wanted a way to quickly identify which window belongs to which project. Now I can set a different title bar color for each:

```json
{
  "title_bar": {
    "background": "#FF00FF"
  }
}
```

### DIFF
--- a/crates/platform_title_bar/src/platform_title_bar.rs
+++ b/crates/platform_title_bar/src/platform_title_bar.rs
@@ -31,6 +31,7 @@ pub struct PlatformTitleBar {
     system_window_tabs: Entity<SystemWindowTabs>,
     workspace_sidebar_open: bool,
     sidebar_has_notifications: bool,
+    override_background: Option<Hsla>,
 }
 
 impl PlatformTitleBar {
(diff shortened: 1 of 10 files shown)


probably-neb:
As a general principle we like to avoid allowing theme/appearance customization in project settings. Regardless, it would be far better to reuse the existing theme overrides than add another setting.

I think the real way to make this work would be something similar to our profile, os,  and release "profile" settings, where you could define project specific settings in your user settings file in such a way that it doesn't control how other people's Zed will appear when viewing your project 
```jsonc
{
  "projects": {
    "absolute/path/to/project": {
      // arbitrary user settings, including theme customization 
   }
}
```

I'd have to think about whether this would take precedence over actual project settings, but by adding this separate feature, the custom titlebar will be fully supported using existing settings, and won't control how it appears for other people looking at your project.

dlight:
@probably-ned can you take a look at [this](https://github.com/zed-industries/zed/issues/13300#issuecomment-3816258667)? Various projects use this approach - two config files in the repo layered on top of each other, one is committed to git, the other is put to `.gitignore`

PR #48174, opened 2026-02-02
editor: Prevent comment extension on REPL separators

lingyaochu:
Closes #47691

Release Notes:

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -25271,19 +25271,20 @@ fn comment_delimiter_for_newline(
     let comment_candidate = snapshot
         .chars_for_range(range.clone())
         .skip(num_of_whitespaces)
-        .take(max_len_of_delimiter)
+        .take(max_len_of_delimiter + 2)
         .collect::<String>();
-    let (delimiter, trimmed_len) = delimiters
+    let (delimiter, trimmed_len, is_repl) = delimiters
         .iter()
         .filter_map(|delimiter| {
             let prefix = delimiter.trim_end();
             if comment_candidate.starts_with(prefix) {
-                Some((delimiter, prefix.len()))
+                let is_repl = comment_candidate[prefix.len()..].starts_with(" %%");
+                Some((delimiter, prefix.len(), is_repl))
             } else {
                 None
             }
         })
-        .max_by_key(|(_, len)| *len)?;
+        .max_by_key(|(_, len, _)| *len)?;
 
     if let Some(BlockCommentConfig {
         start: block_start, ..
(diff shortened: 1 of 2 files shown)


crates/editor/src/editor.rs
```diff
@@ -25323,19 +25271,20 @@ fn comment_delimiter_for_newline(
     let comment_candidate = snapshot
         .chars_for_range(range.clone())
         .skip(num_of_whitespaces)
-        .take(max_len_of_delimiter)
+        .take(max_len_of_delimiter + 2)
```
maxdeviant, line 25274: What is this magic `2`? Is this for the comment characters (`//`)?

lingyaochu: No, the `2` is to account for the possible addition of `%%`.

For example, in a Python file where the comment delimiter is `#`, a line might start with `# %%`. To keep it performant, the old code `.take(max_len_of_delimiter)` would only read the `# ` and stop. By adding 2, we can now include the `%%` in the check.

crates/editor/src/editor.rs
```diff
@@ -25358,7 +25307,20 @@ fn comment_delimiter_for_newline(
...
+        let line_content_after_cursor: String = snapshot
+            .chars_for_range(range)
+            .skip(start_point.column as usize)
+            .collect();
+
+        if line_content_after_cursor.trim().is_empty() {
+            return None;
+        } else {
+            return Some(delimiter.clone());
+        }
```
rgbkrk, line 25328: 👍

rgbkrk (approved):
Thank you so much. This is a fantastic quality of life improvement.

crates/editor/src/editor.rs
```diff
@@ -25323,19 +25271,20 @@ fn comment_delimiter_for_newline(
...
+        .take(max_len_of_delimiter + 2)
         .collect::<String>();
-    let (delimiter, trimmed_len) = delimiters
+    let (delimiter, trimmed_len, is_repl) = delimiters
         .iter()
         .filter_map(|delimiter| {
             let prefix = delimiter.trim_end();
             if comment_candidate.starts_with(prefix) {
-                Some((delimiter, prefix.len()))
+                let is_repl = comment_candidate[prefix.len()..].starts_with(" %%");
```
maxdeviant: It doesn't seem right that we should have to be aware of the specifics of the REPL inside of the editor code (which is otherwise generic).

lingyaochu: I completely understand the concern about leaking REPL-specific logic into the generic editor core. My rationale for this approach was based on a few trade-offs:

- The `# %%` pattern is a de-facto standard for REPL cell division (Jupyter, VS Code, etc.). Since Zed's own [documentation](https://zed.dev/docs/repl?highlight=repl#cell-mode) also references this specific pattern, I treated it as a stable "comment protocol" rather than a transient feature.
- The `comment_delimiter_for_newline` function is on a critical path (triggered on every newline). I chose a direct string match to ensure minimal overhead and avoid the complexity of querying language-specific configurations or the REPL implementations.
- Simplicity: I wanted to keep the fix self-contained within the comment logic.

I agree that the editor logic should remain generic. The current implementation was a trade-off to balance performance on this critical path with the `# %%` standard.

I'm currently unsure how to refactor this into a generic solution while preserving the performance and simplicity of this direct check. What would you suggest as the preferred approach?

rgbkrk (approved):

crates/editor/src/editor.rs
Veykril: Did not see this review for some reason, I do agree this being hard coded in editor is a bit weird so I've been thinking. It feels like notebooks/REPL kind of form a dialect on the original language with this here, so it might make sense to model this as such on the underlying language buffer?

rgbkrk: Yeah there's some component of "runnables" for languages. As mentioned in https://github.com/zed-industries/zed/pull/49636, maybe we need a `runnables.scm`? Not sure. It works so well in that PR that I know I'm probably not going to use the jupytext markers as much. Tinkering on models is extremely fast and fluid.

lingyaochu: I completely agree that modeling this at the syntax/Tree-sitter layer is the more elegant and fundamental solution. Intuitively, it feels like the right level of abstraction to solve this problem thoroughly.

However, to be honest, I'm not very familiar with the Tree-sitter implementation details. Given this constraint, the best generic approach I can come up with right now is to introduce a new language setting that allows users to specify patterns where comment auto-continuation should be disabled.

We could then add the REPL separator (# %%) to the default settings. This effectively addresses the 'hard-coding' concern by decoupling the REPL logic from the editor core and moving it to user configuration.

Does this sound like a reasonable compromise?

PR #48188, opened 2026-02-02
settings_ui: Fix renderer lookup for font weight settings

cppcoffee:
Release Notes:

- Fixed renderer lookup for font weight settings


MrSubidubi (changes requested):
Was this not already fixed by https://github.com/zed-industries/zed/pull/47384?

PR #48193, opened 2026-02-02
Commit message editor history

vittorius:
Implements https://github.com/zed-industries/zed/discussions/29084

Release Notes:

- Add the commit message editor history holding 50 last commit messages
- Enable commit message history navigation with actions `git_panel::PrevCommitEditorHistoryEntry` and `git_panel::NextCommitEditorHistoryEntry` (like [in VS Code](https://code.visualstudio.com/docs/sourcecontrol/staging-commits#:~:text=To%20cycle%20through%20your%20previous%20commit%20messages%2C%20press%20%E2%86%91%20and%20%E2%86%93%20while%20focused%20in%20the%20commit%20message%20input%20box.))

Demo:

https://github.com/user-attachments/assets/c85614fd-247f-414c-a239-c3abe8e1f602

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -1001,6 +1001,8 @@
       "ctrl-shift-enter": "git::Amend",
       "alt-up": "git_panel::FocusChanges",
       "alt-l": "git::GenerateCommitMessage",
+      "up": "git_panel::PrevCommitEditorHistoryItem",
+      "down": "git_panel::NextCommitEditorHistoryItem",
     },
   },
   {
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -1001,8 +1001,8 @@
       "ctrl-shift-enter": "git::Amend",
       "alt-up": "git_panel::FocusChanges",
       "alt-l": "git::GenerateCommitMessage",
-      "up": "git_panel::PrevCommitEditorHistoryItem",
-      "down": "git_panel::NextCommitEditorHistoryItem",
+      "up": "git_panel::PrevCommitEditorHistoryEntry",
+      "down": "git_panel::NextCommitEditorHistoryEntry",
     },
   },
   {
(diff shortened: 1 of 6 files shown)


EugeneTorap:
Hey @dinocosta @danilo-leal
Can you review this PR?

dinocosta (commented):
Hey @vittorius! Thank you for working on this 🙂

I took a look at the changes and tested them locally. I think we need some additional product discussion before settling on the implementation:

1. The `up` and `down` keys cannot unconditionally navigate history because they already move the cursor within the commit editor. With these changes, normal vertical cursor movement no longer works, particularly for multiline commit messages. We should preserve regular editor movement and only navigate history when the cursor is at the appropriate boundary. For example, `up` at the beginning of the input and `down` at the end, similar to how the Agent Panel prioritizes cursor movement before scrolling the thread.
2. History is currently scoped to the workspace, but one workspace can contain multiple repositories and the Git panel can switch between them. As such, history should be kept per repository, similar to how commit-message drafts are currently handled.
3. Before changing the implementation, I think we should clarify the intended source of the history. The current implementation records messages from successful commits made through Zed. I believe this is also how VS Code implements the feature, but another option would be to derive it from the repository’s commit history.

The [existing discussion](https://github.com/zed-industries/zed/discussions/29084) has not yet converged on the expected behavior for the feature. Given that the direction is still unresolved and could substantially affect the implementation, I’m going to close this for now to avoid asking you to invest more time before we agree on the expected behavior. To be clear, this is not a rejection of the feature, I would like to see it in Zed, and we can reopen this pull request once the approach has been clarified, in case it makes sense to do so.

Thanks again for putting this together and providing a concrete implementation for us to evaluate!

vittorius:
@dinocosta Thanks for the feedback! I'm happy to rework the implementation according to whatever the implementation path eventually is finalized in the [discussion](https://github.com/zed-industries/zed/discussions/29084). Will wait for that.

PR #48197, opened 2026-02-02
language: Avoid cloning of the previous tree_sitter::Tree

marcocondrache:
The parsing text function used the old tree only as a ref so it doesn't make sense to clone it.

Release Notes:

- N/A


SomeoneToIgnore (approved):
Thanks!

PR #46722, opened 2026-01-13
languages: Separate control flow keywords for Shell Scripts

colemcanelly:
Part of https://github.com/zed-industries/zed/issues/9461. Similar PRs:

- https://github.com/zed-industries/zed/pull/39683
- https://github.com/zed-industries/zed/pull/39801

## Release Notes:

- Added the ability to separately highlight control flow keywords for Shell Scripts for users and theme authors via the `keyword.control` syntax property.

## Screenshots

| **Before** | **After** |
|:--:|:--:|
| image | image |

PR #39801, opened 2025-10-08
languages: Separate control flow keywords for JS/TS/TSX

azeier:
Related to #9461, inspired by #39683

`await` and `yield` both seem somewhat debatable on whether they should be considered the be control flow keywords. 

For now I went with:
- `await`: no – The control flow effect of `await` is at a level does not seem relevant for syntax highlighting. 
- `yield`: yes – `yield` directly affects the output of a generator, and is also included for consistency with Rust (#39683).
 
 Happy to change these either direction.

SCR-20251008-izus

--- 

Release Notes:

- Improved granularity of keyword highlighting for JS/TS/TSX: Themes can now specify `keyword.control` for control flow keywords like `if`, `else`, `return`, etc.

PR #39683, opened 2025-10-07
Highlight control flow in Rust/C/C++

Hmikihiro:
part of https://github.com/zed-industries/zed/issues/9461

Release Notes:
- Added the ability to seperately highlight control flow keywords for Rust, C and C++ for users and theme authors via the `keyword.control` syntax property
スクリーンショット 2025-10-07 22 21 59

PR #48200, opened 2026-02-02
languages: Highlight case statements and special variables for Shell Scripts

colemcanelly:
Part of https://github.com/zed-industries/zed/issues/9461. Similar PRs:
- https://github.com/zed-industries/zed/pull/46722
- https://github.com/zed-industries/zed/pull/39683
- https://github.com/zed-industries/zed/pull/39801

## Release Notes:

- Allowed highlighting for case statement items for users and theme authors via the `string.regex` syntax property.
- Allowed highlighting for special variables via `variable.special`

## Screenshots

| **Before** | **After** |
|:--:|:--:|
| image | image |

### DIFF
--- a/crates/languages/src/bash/highlights.scm
+++ b/crates/languages/src/bash/highlights.scm
@@ -104,3 +104,6 @@
   (command (_) @constant)
   (#match? @constant "^-")
 )
+
+(case_item value: (_) @string.regex)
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (approved):
Thank you!

PR #48203, opened 2026-02-02
Fix subdirectory `.editorconfig` files being ignored in certain directory structures

amtoaer:
Closes #48187

The bug occurred when iterating internal_configs (a BTreeMap sorted by path): the code would `break` on the first non-matching path, causing configs with lexicographically later paths to be skipped.

For example, when querying "d/d.rs" with configs ["", "b", "d"], iteration would break at "b" (since "d/d.rs" doesn't start with "b"), preventing "d"'s config from being applied.

This PR replaces `break` with `continue` to skip non-ancestors, and adds a minor early-exit optimization when `config_path > for_path` since later paths can't be ancestors.

Release Notes:

- Fixed subdirectory `.editorconfig` files being ignored in certain directory structures


smitbarmase (approved):
Thanks for fixing this! It looks like the bug predates #46332, and the added test is great.

PR #48207, opened 2026-02-02
csv_preview: #1 Add CSV preview with live table view and interactive features

HalavicH:
## **Description:**

**Context:**
This PR introduces an initial CSV preview feature for Zed, building upon two previously merged infrastructure PRs:
- [#46341](https://github.com/zed-industries/zed/pull/46341) - Data table dynamic column support (removed const generics)
- [#46190](https://github.com/zed-industries/zed/pull/46190) - Variable row height mode for data tables

This implementation is based on the [original draft PR #44344](https://github.com/zed-industries/zed/pull/44344), which has been carefully decomposed into smaller, reviewable pieces.

---

#### **Features Included:**

**Core Infrastructure:**
- Live CSV parsing with smart debouncing (200ms cooldown)
- Performance monitoring with built-in timing metrics (not displayed in UI yet)
- Automatic file change detection and re-parsing
- Support for quoted fields, multiline cells, and escaped characters

**Table Display:**
- Variable row height rendering with fallback to uniform mode (switchable via settings)
- Draggable column resizing (reusing existing data table infrastructure)
- Row identifiers supporting both source line numbers and sequential row numbers
- Configurable font rendering (UI font vs monospace)
- Tooltips showing full cell content on hover

**Interactive Features:**
- Column sorting (ascending/descending) with visual indicators

**Settings Panel:**
- Toggle between variable/uniform row rendering
- Font type selection (UI/monospace)
- Row identifier type configuration
- Debug information display
- Multiline cell rendering options

---

#### **Features Intentionally Removed for This PR:**

To reduce complexity and review scope, the following features were temporarily reverted and will be reintroduced in subsequent PRs:

- ❌ Settings pannel with performance metrics overlay
- ❌ Cell selection (single, multiple, and range selections)
- ❌ Keyboard navigation with arrow keys and selection extension
- ❌ Copy functionality supporting CSV, TSV, and Markdown table formats
- ❌ Inline cell editing with file persistence
- ❌ Viewport following for large datasets
- ❌ Column filtering and search capabilities

These removals were done via "time-machine" commits that cleanly nuked vertical slices of functionality from the complete implementation.

---

**Technical Implementation:**

The feature is organized into a dedicated `csv_preview` crate with the following structure:

```
crates/csv_preview/
├── src/
│   ├── csv_preview.rs          # Main view and coordination logic
│   ├── parser.rs               # CSV parsing and editor integration
│   ├── settings.rs             # Configuration types and defaults
│   ├── table_data_engine.rs    # Data transformation logic
│   ├── renderer/               # UI rendering modules
│   │   ├── preview_view.rs     # Main render implementation
│   │   ├── render_table.rs     # Table component assembly
│   │   ├── table_cell.rs       # Individual cell rendering
│   │   ├── table_header.rs     # Header with sorting controls
│   │   └── row_identifiers.rs  # Line number column
│   └── types/                  # Core data structures
│       ├── table_like_content.rs
│       ├── coordinates.rs      # Display vs data coordinate systems
│       └── table_cell.rs
```

**Key architectural decisions:**
- **Dual coordinate system**: Separates data indices from display indices to support sorting/filtering
- **Component reuse**: Leverages existing `data_table` infrastructure from the keymap editor

---

**Integration:**
- Registers `csv::OpenPreview` action (currently without default keybindings)
- Follows the same workspace integration pattern as `markdown_preview` and `svg_preview`
- Automatically detects `.csv` file extensions
- Tab integration with appropriate icons and naming

---

**Code Structure Note:**
Some code structures, types, and documentation may appear redundant or over-engineered in this initial implementation. This is intentional - the feature was developed as a complete system and then decomposed by functionality rather than being built incrementally. The "extra" infrastructure supports features that were removed for this PR but will be reintroduced in subsequent ones.

This approach was chosen over extensive refactoring because:
1. The complete feature took 200+ commits to develop with significant rewrites
2. Clean extraction of vertical slices was more feasible than rebuilding incrementally
3. The end state will utilize all these components, making current "redundancy" temporary

I apologize for any inconvenience this may cause during review, but the alternative would have required significant refactoring effort just to make intermediate states "prettier," which seemed counterproductive.

---

**Future Work:**
This lays the groundwork for upcoming PRs that will reintroduce the removed features:
- Cell selection and keyboard navigation
- Copy functionality with multiple output formats
- Inline editing capabilities with undo/redo
- Column filtering and search
- TSV and other delimiter support
- Improved horizontal scrolling behavior
- Settings persistence

**Testing:**
Includes test fixtures demonstrating multiline cell handling, various column counts, and edge cases.

---

**Release Notes:**

- N/A This is feature flagged


crates/csv_preview/src/csv_preview.rs
```diff
@@ -0,0 +1,228 @@
...
+        workspace.register_action(|workspace, _: &OpenPreview, window, cx| {
+            if let Some(editor) = workspace
+                .active_item(cx)
+                .and_then(|item| item.act_as::<Editor>(cx))
+                .filter(|editor| Self::is_csv_file(editor, cx))
+            {
+                let csv_preview = Self::new(&editor, cx);
+                workspace.add_item_to_active_pane(Box::new(csv_preview), None, true, window, cx);
+            }
+        });
```
Anthony-Eid: `OpenPreview` action should also check to make sure there are no pre existing item that represent the same csv file in the active pane. In the case that there is a pre existing item Zed should just refocus it

crates/csv_preview/src/csv_preview.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+impl CsvPreviewView {
+    pub(crate) fn editor_state(&self) -> &EditorState {
+        self.active_editor_state
+            .as_ref()
+            .expect("Expected main editor to be initialized")
```
Anthony-Eid: I would rather have this function return a Result instead of panicking where there's not active editor state.

crates/csv_preview/src/csv_preview.rs
```diff
@@ -0,0 +1,228 @@
...
+                last_parse_end_time: None,
+                engine: TableDataEngine::default(),
+            };
+
+            // No need to trigger any filtering / sorting here, as it's retrigered on parsing
+
+            view.set_editor(editor.clone(), cx);
+            view
+        })
+    }
```
Anthony-Eid: Two things that could be improved here

1. `new` methods are usually the first in a implementation block in the Zed codebase. So moving this up would be more inline with how we do things
2. `view.set_editor(editor.clone(), cx);` seems to be infallible, so couldn't we get rid of the Option wrapper around `active_editor_state` struct field? If we can, we should

crates/csv_preview/src/parser.rs
```diff
@@ -0,0 +1,546 @@
...
+            raw_row.push(TableCell::Virtual);
+        }
+    }
+
+    TableRow::from_vec(raw_row, max_number_of_cols)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
Anthony-Eid, line 320: Not a blocker, but I wonder how hard it would be to add property testing to this parser

crates/csv_preview/src/settings.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Clone, Default)]
+pub(crate) struct CsvPreviewSettings {
+    pub(crate) rendering_with: RowRenderMechanism,
+    pub(crate) vertical_alignment: VerticalAlignment,
+    pub(crate) font_type: FontType,
+    pub(crate) numbering_type: RowIdentifiers,
+    pub(crate) show_debug_info: bool,
+    pub(crate) multiline_cells_enabled: bool,
+}
```
Anthony-Eid, line 46: Not a blocker, but we would need to choose what settings we actually want to add to Zed. We prefer adding as little as possible and having defaults that are as right as possible.

For example, `font_type` should be `UI` at all times, unless a user is editing a cell.

crates/csv_preview/src/table_data_engine.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.d2d_mapping
+            .apply_sorting(self.applied_sorting, &self.contents.rows);
+        self.d2d_mapping.merge_mappings();
+    }
+
+    /// Applies sorting and filtering to the data and produces display to data mapping
+    pub(crate) fn calculate_d2d_mapping(&mut self) {
+        self.d2d_mapping
+            .apply_sorting(self.applied_sorting, &self.contents.rows);
+        // self.calculate_filters_with_availability();
```
Anthony-Eid: Should this line be commented out?

crates/csv_preview/src/table_data_engine.rs
```diff
@@ -0,0 +1,95 @@
+//! This module defines core operations and config of tabular data view (CSV table)
+//! It operates in 2 coordinate systems:
+//! - `DataCellId` - indices of src data cells
+//! - `DisplayCellId` - indices of data after applied transformations like sorting/filtering, which is used to render cell on the screen
+//!
+//! It's designed to contain core logic of operations without relying on `CsvPreviewView`, context or window handles.
```
Anthony-Eid, line 6: Love the separation of UI and data layers you have, plus the comments! Good job

crates/csv_preview/src/table_data_engine.rs
```diff
@@ -0,0 +1,95 @@
...
+            log::debug!("Sorting data rows by {sorting:?}");
+            sort_data_rows(&rows, data_rows, sorting)
+        } else {
+            log::debug!("Disabling sorting");
+            data_rows
+        };
+
+        self.sorted_rows = sorted_rows;
+        log::trace!("Sorted mapping: {:?}", self.sorted_rows);
+    }
```
Anthony-Eid, line 78: We should remove most of the log statements here, if not all

crates/csv_preview/src/renderer/preview_view.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        let theme = cx.theme();
+
+        let render_prep_start = Instant::now();
+        let table_with_settings = v_flex()
+            .size_full()
+            .p_4()
+            .bg(theme.colors().editor_background)
+            // Apparently, this should make newly created CSV preview to get focus automatically
+            .track_focus(&self.focus_handle)
```
Anthony-Eid: I'm not sure this is true, from my understanding it just attaches a focus handle to a div, we should double check this together

crates/csv_preview/src/table_data_engine/sorting_by_column.rs
```diff
@@ -0,0 +1,49 @@
...
+pub fn sort_data_rows(
+    content_rows: &[TableRow<TableCell>],
+    mut data_row_ids: Vec<DataRow>,
+    sorting: AppliedSorting,
+) -> Vec<DataRow> {
+    data_row_ids.sort_by(|&a, &b| {
+        let row_a = &content_rows[*a];
+        let row_b = &content_rows[*b];
+
+        // TODO: Decide how to handle nulls (on top or on bottom)
```
Anthony-Eid, line 29: We could show Zed's null symbol in this case

Anthony-Eid (changes requested):
This is looking good for an MVP, there are some small changes we should make before merging 

Also, this will have to be featured flagged

crates/csv_preview/src/csv_preview.rs
HalavicH: Done

crates/csv_preview/src/csv_preview.rs
HalavicH: done

crates/csv_preview/src/csv_preview.rs
HalavicH: done

crates/csv_preview/src/table_data_engine.rs
HalavicH: removed

crates/csv_preview/src/settings.rs
HalavicH: will be refined at later stages.

> For example, font_type should be UI at all times, unless a user is editing a cell.

I disagree. monospace font is usually prefered in such editors (we'll decide on that later)

crates/csv_preview/src/table_data_engine.rs
HalavicH: Removed some, but the latter are emited on event (on button click), not every frame. So it's not poluting the logs

crates/csv_preview/src/table_data_engine/sorting_by_column.rs
HalavicH: for not I will keep empty string. I'll introduce null with editor feature

crates/csv_preview/src/renderer/preview_view.rs
HalavicH: removed comment

Anthony-Eid (approved):
This is ready for its first iteration in Zed and thus is mergable! Congrats 

It's featured flag for now, so we can keep polishing the cvs preview before it's released

kczx3:
Does this effort being in core necessitate closing https://github.com/zed-industries/extensions/issues/476?

kushvinth:
Elite PR

PR #48222, opened 2026-02-02
Support path pasting in terminal

dunkmann00:
This adds the functionality to support pasting the file path of an item when the copied item supports it. This mirrors the behavior of `Terminal.app` on macOS.

This only implements the functionality on macOS but could be extended to other platforms.

I find this convenient when I'm using Finder to navigate around the file system and I want to copy a directory or file path and put it in the terminal. You can copy the item from Finder and paste it into the terminal and it will write out the full path of the item, making it easy to change directories or provide path parameters to commands.

Release Notes:

- Added path pasting functionality in terminal

### DIFF
--- a/assets/keymaps/default-macos.json
+++ b/assets/keymaps/default-macos.json
@@ -1264,6 +1264,7 @@
       "ctrl-cmd-space": "terminal::ShowCharacterPalette",
       "cmd-c": "terminal::Copy",
       "cmd-v": "terminal::Paste",
+      "ctrl-cmd-v": "terminal::PasteText",
       "cmd-f": "buffer_search::Deploy",
       "cmd-a": "editor::SelectAll",
       "cmd-k": "terminal::Clear",
--- a/assets/keymaps/default-windows.json
+++ b/assets/keymaps/default-windows.json
@@ -1221,6 +1221,7 @@
       "shift-insert": "terminal::Paste",
       "ctrl-v": "terminal::Paste",
       "ctrl-shift-v": "terminal::Paste",
+      "ctrl-alt-v": "terminal::PasteText",
       "ctrl-i": "assistant::InlineAssist",
       "alt-b": ["terminal::SendText", "\u001bb"],
       "alt-f": ["terminal::SendText", "\u001bf"],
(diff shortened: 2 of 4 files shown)


dunkmann00:
I think that should address the error. My bad on that, it was on Windows so I didn't catch it locally.

dunkmann00:
Okay ran clippy locally so the test error should be fixed and I think windows should be good now.

ChristopherBiscardi:
I pulled this locally and confirmed that it does succeed at implementing pasting filepaths from finder, so we should be able to merge this.

It will need to have main merged since there are a couple conflicts, and pass tests, but feel free to ping me and I'll make sure they're running so we can get this in.

dunkmann00:
Hey @ChristopherBiscardi, sorry it took me so long to get back to this, but I've updated the PR.

There were a bunch of changes since I first made this, it actually is much simpler to do now since path handling has been added to the macOS clipboard. It also took advantage of `ExternalPaths`, which at the time I wasn't sure if that was what I should use to implement this, but looks like it probably was!

I haven't tested this on Windows, but I think it should work. If the tests show a problem or you see anything you'd like me to change I'd be happy to do so!

dunkmann00:
I was able to test this out on Windows and it works 👍 

I think the error from CI doesn't have to do with this PR?

ChristopherBiscardi:
CI failure seemed like a networking issue. I kicked it off again.

ChristopherBiscardi:
Looks like the new implementation has some extra space and is wrapped in quotes:

image

vs Terminal behavior:

image

dunkmann00:
This is how drag and drop was handled already so I thought it would be better to have it be consistent. This way uses `self.add_paths_to_terminal` which is where the spaces and quotes are getting added. This also handles if you copy multiple files. And this is what drag and drop uses, so the behavior is consistent in zed, but different from the Terminal.app. But it does enable the same thing, the ability to copy a file and paste its path in zed's terminal.

ChristopherBiscardi (approved):
yep that makes sense to me, thanks!

dunkmann00:
Thanks!

m-ar-c:
Works on GNU/Linux too (debian) 👌

PR #48233, opened 2026-02-03
Add initial support for edit predictions via Ollama

### LINKED ISSUE #15968: Support using ollama as an inline_completion_provider
### Description

I am successfully using my local ollama models using assistant panel.

I would love to be able to use them as well as an `inline_completion_provider`. 

Currently, only `none`, `copilot` or `supermaven` values are supported.

maxbrunsfeld:
Closes https://github.com/zed-industries/zed/issues/15968

Release Notes:

- Added the ability to use Ollama as an edit prediction provider.


maxbrunsfeld:
This is actually already out; we fast tracked it.

PR #48245, opened 2026-02-03
Fix incorrect memory display in the language servers menu on Linux

feeiyu:
While trying out the new feature introduced in #48226, I noticed an issue on Linux where the reported memory usage is incorrect.

On Linux, even when using ProcessRefreshKind::nothing, thread information is still collected. To fix this, the memory calculation needs to explicitly exclude task/thread data by using without_tasks.

Before

After

Release Notes:

- N/A


reflectronic (approved):
Thank you!

PR #48248, opened 2026-02-03
Add git untracked changes config

iHildy:
# Git panel untracked changes modes

## Problem
Zed's git panel treats untracked files differently from VS Code's default behavior, which makes staging and committing new files feel inconsistent for users coming from VS Code. This PR adds opt-in settings for VS Code-style behavior while preserving the current default behavior for existing users.

## Summary
- Add `git_panel.untracked_changes` with modes `classic`, `mixed`, `separate`, and `hidden` (default `classic`).
- `classic` preserves pre-change behavior: separate untracked section, and Stage All includes untracked.
- `mixed` merges untracked into the main changes list and includes them in stage/commit-all.
- `separate` keeps untracked in their own section, stages/commits tracked only, and adds a "Stage Untracked" action.
- `hidden` hides untracked entirely and limits stage/commit-all to tracked.
- Stage/Unstage labels and actions now reflect the active scope (tracked-only in `separate`/`hidden`).
- Settings UI, defaults, docs, and VS Code import mapping updated.

## Design decisions
- Defaulted to `classic` to avoid changing behavior for existing users.
- Implemented `mixed` as a single list (no "Untracked" header) to better match VS Code default.
- Kept staging of untracked explicit in `separate` via a menu action instead of auto-staging.
- Scoped Unstage in `separate`/`hidden` to tracked entries to preserve separation semantics.

## Testing
  - cargo test -p git_ui untracked
      - `test_hidden_mode_excludes_untracked` ok
      - `test_separate_mode_stage_all_excludes_untracked` ok
      - `test_mixed_mode_commit_with_untracked` ok
  - cargo test -p git_ui test_classic_mode_preserves_stage_all_behavior ok

## Screenshots
[Google Slideshow](https://docs.google.com/presentation/d/1o-J6ISE3rMZ4v_7gWx0Eu05Q2f0k_gTCaF45NvBnAeQ/edit?usp=sharing)

---

Release Notes:
- Added `git_panel.untracked_changes` to control how untracked files are displayed and whether stage/commit-all include them (`classic`, `mixed`, `separate`, `hidden`).

Issue:
Closes no issue - didn't create a discussion for this feature. I apologize for skipping the procedure and am happy to create a discussion now or take suggestions here.

This change was assisted by AI but reviewed and tested manually


cole-miller:
Thanks! This is a real issue, however I don't think we want to approach it by adding a setting (there are a lot of knobs on the git panel status list already). Note that we already have a flat/sort by status distinction that allows mixing untracked files into the list with the rest. For stage all, I think we should remove that button from the top of the panel and instead have one button for each section header (tracked/untracked/etc.), that only acts on the files in that section.

PR #48273, opened 2026-02-03
Add support for absolute paths in file finder

### LINKED ISSUE #23740: Let me type an absolute filepath in the file finder
### Description

VS code has this feature in its file finder: if I type and absolute path, something that starts with `~` or `/`, and the file exists it'll show a listing for the file and let me tap enter to open it.

This's super useful if I'm editing in a project workspace but quickly want to access my `~/.zshrc` or something.

Image

suryapandian:
Closes https://github.com/zed-industries/zed/issues/23740

Release Notes:
  -   The file finder now supports opening files using absolute paths, including home directory shortcuts (~). Previously, you could only search for files within your project workspace.   

Bug fix:
- The file finder checks for absolute path `&&` and then checks if it looks like absolute path

### DIFF
--- a/crates/file_finder/src/file_finder.rs
+++ b/crates/file_finder/src/file_finder.rs
@@ -39,7 +39,7 @@ use ui::{
 };
 use util::{
     ResultExt, maybe,
-    paths::{PathStyle, PathWithPosition, looks_like_absolute},
+    paths::{PathWithPosition},
     post_inc,
     rel_path::RelPath,
 };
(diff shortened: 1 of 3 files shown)


crates/file_finder/src/file_finder.rs
```diff
@@ -1424,19 +1424,20 @@ impl PickerDelegate for FileFinderDelegate {
                 path_position,
             };
 
+            let path_str_for_check = path_str.unwrap_or("");
+            let is_looks_absolute_path = path.is_absolute() || path_str_for_check.starts_with('~');
```
SomeoneToIgnore: > is_looks_absolute_path

This is not even a proper English, can we have a better name?

Besides, what's the point of converting the path to string? 
Cannot we check the path for `.starts_with` directly or remove the whole check entirely, as `lookup_absolute_path` does those all for us?

SomeoneToIgnore (changes requested):
Thank you, but `crates/file_finder/src/file_finder_tests.rs` badly needs a test for this.

crates/file_finder/src/file_finder.rs
suryapandian: addressed [here](https://github.com/zed-industries/zed/pull/48273/changes/3b216163253cab0a11b17bdd2e121b713aee963c) by removing the check as `lookup_absolute_path` does that

SomeoneToIgnore (commented):
I cannot check this properly as it does not compile, but the code generally looks nice now, thank you.

PR #48287, opened 2026-02-03
editor: Ensure that spacer blocks are visible in light themes

dinocosta:
The color used for the slash pattern when rendering `Block::Spacer` was set to a fixed color, `0xFFFFFF10`, which is almost white, making it super hard to view in light themes, where the editor's background is almost white.

As such, this commit updates that color so as to use something that is more theme-specific, ensuring that it is easily visible in both light and dark themes.

Not entirely sure if the `border` color is the best one for this, so we might need to revisit that, depending on themes.

Release Notes:

- N/A

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -4006,7 +4006,7 @@ impl EditorElement {
                 .w_full()
                 .h((*height as f32) * line_height)
                 .bg(pattern_slash(
-                    cx.theme().colors().border.opacity(0.5),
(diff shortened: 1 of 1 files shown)


dinocosta:
Here's a quick before and after comparison using One Light and One Dark ▼

| Theme | Before | After |
|--------|--------|------|
| One Light | one_light_before | one_light_after |
| One Dark | one_dark_before | one_dark_after |

PR #48310, opened 2026-02-04
remote: Reuse existing window when opening zed://ssh/ URLs for the same host

zertosh:
When opening a `zed://ssh/$AUTHORITY/$PATH` url, zed always opened a new
window, even if there is already a window connected to that host. Now it
finds an existing workspace window with a matching remote connection and
opens the file there instead. This matches the behavior of the `zed` cli
when opening a local path.

Release Notes:

- Opening `zed://ssh/` urls will now reuse an existing window if possible.


reflectronic:
Thanks for the pull request. I've been working on https://github.com/zed-industries/zed/pull/48891 that implements the same functionality while also cleaning up some duplication in this part of the codebase. I'll be closing this PR in favor of that one.

PR #48317, opened 2026-02-04
agent: Remove duplicate `line_hint` assignment in `StreamingFuzzyMatcher::push`

Albab-Hasan:
Release Notes:

- N/A


crates/agent/src/edit_agent/streaming_fuzzy_matcher.rs
MrSubidubi: Am I missing something here or is this assignment here save to remove as opposed to the other one? We pass `None` here quite a few times and it would make a difference if we were to only set this if it is `Some`, would it not?

Albab-Hasan: you're right. I removed the wrong one. the unconditional assignment was the one that determined the actual behavior making the conditional check above it dead code. i should remove the conditional if
line_hint.is_some() block instead and keep the unconditional assignment. ill update the pr

MrSubidubi (approved):
This looks good now, thanks!

PR #48319, opened 2026-02-04
remote: Fix build_remote_server_from_source compression on unix

zertosh:
When using gzip, the output path's extension becomes `.gz`.

Release Notes:

- N/A


Veykril (approved):
Thanks!

PR #48332, opened 2026-02-04
lsp: Add schema support for LSP settings field

aviatesk:
This extends the LSP settings schema system to also provide autocomplete for the `settings` field (used for `workspace/configuration` responses), in addition to the existing `initialization_options` support (#).

Changes:
- Add `settings_schema` method to `LspAdapter` trait and `CachedLspAdapter`
- Update schema URL paths to be more explicit:
  - `lsp/{adapter}/initialization_options` for init options schema
  - `lsp/{adapter}/settings` for settings schema
- Add schema resolution logic for the new settings path
- Update tests to verify both schema references

Release Notes:

- Added autocomplete support for the `settings` field in LSP configuration, complementing the existing `initialization_options` autocomplete.

PR #48334, opened 2026-02-04
extension_api: Add language server schema methods

aviatesk:
(This should be merged after #48332)

This PR exposes the LSP settings schema functionality to extensions, allowing them to provide JSON schema for `initialization_options` and `settings` fields to enable autocomplete in settings files.

New extension API methods (v0.8.0+):
- `language_server_initialization_options_schema`
- `language_server_settings_schema`

Both methods return an optional JSON string conforming to JSON schema.
Older extension versions gracefully return `None`.

Release Notes:

- Added support for settings schemas for the next version of the extension API so that settings autocompletion can be provided for language server settings.


crates/extension_host/src/wasm_host/wit.rs
```diff
@@ -465,6 +465,42 @@ impl Extension {
...
+            Extension::V0_8_0(ext) => {
+                ext.call_language_server_initialization_options_schema(
+                    store,
+                    &language_server_id.0,
+                    resource,
+                )
+                .await
+            }
+            // Schema methods are only available in v0.8.0+
+            _ => Ok(Ok(None)),
```
MrSubidubi: Let's remove the comment and make the match exhaustive here please

crates/extension_host/src/wasm_host/wit.rs
```diff
@@ -465,6 +465,42 @@ impl Extension {
...
+        language_server_id: &LanguageServerName,
+        resource: Resource<Arc<dyn WorktreeDelegate>>,
+    ) -> Result<Result<Option<String>, String>> {
+        match self {
+            Extension::V0_8_0(ext) => {
+                ext.call_language_server_settings_schema(store, &language_server_id.0, resource)
+                    .await
+            }
+            // Schema methods are only available in v0.8.0+
+            _ => Ok(Ok(None)),
```
MrSubidubi: Same here

crates/extension_api/src/extension_api.rs
```diff
@@ -100,6 +100,30 @@ pub trait Extension: Send + Sync {
...
 
+    /// Returns the JSON schema for the initialization options.
+    ///
+    /// The schema should conform to JSON Schema and will be used to provide
+    /// autocomplete in settings files.
+    fn language_server_initialization_options_schema(
+        &mut self,
+        _language_server_id: &LanguageServerId,
+        _worktree: &Worktree,
+    ) -> Result<Option<serde_json::Value>> {
```
MrSubidubi: I think I would prefer for the return type to just be `Option<serde_json::Value`, as we do for the builtin one (applies to some other code paths here too). Haven't chatted about this internally yet, but I see no big advatage in having this failable in the context of extensions when we don't allow it to fail, and I also don't really think this should fail much if at all. Happy to discuss this though

MrSubidubi (changes requested):
Another topic to discuss: I have been thinking about supporting an additional way to provide a schema, which would be via an URI. Strictly speaking, this might already be possible with the API at hand, but I am wondering as to whether we would want to make the option more explicit, as not many people might know about this.

crates/extension_api/src/extension_api.rs
aviatesk: That's a fair point, but I think keeping `Result` is worth it for extension APIs. For example, some extension may fetch the schema from a remote URL, which involves HTTP requests and JSON parsing and these are inherently fallible operations. If the return type were `Option<Value>`, extension authors would be forced to silently discard errors and return `None`, losing the distinction between "no schema available" and "schema retrieval failed." I think the host side should probably have the opportunity to log or surface these errors.

aviatesk:
Friendly ping.

crates/extension_api/src/extension_api.rs
MrSubidubi: I think this is better solved once we have logging available in extensions and thus decided against that. At the end, it is not an error we will surface to the user, and for most cases, extensions will not need the fetching (best case, they should not do it at all).

MrSubidubi (approved):
Added some changes and fixes on top of this myself. Thanks for the PR and getting the ball rolling on this!

aviatesk:
Hi! Just wanted to follow up on the `zed_extension_api` v0.8.0 release.

As far as I can tell, crates.io still only has `zed_extension_api` v0.7.0, while several extension API changes have already landed for v0.8.0, including:

- `lsp::Symbol.container_name` from #46822
- language server schema methods from this PR

I'm currently using these APIs in an external Julia extension, but it requires a local `[patch.crates-io]` against the Zed repository and can't be published cleanly yet.

Is the v0.8.0 release still blocked on agent provider extensions, or would it be possible to publish the current v0.8.0 API soon? If a PR similar to #32945 for `zed_extension_api: Release v0.8.0` would be welcome at this point, I'm willing to contribute.

MrSubidubi:
We are currently still landing some incremental improvements and yes, so far we are still blocked on a bigger change for this to be released. I can understand that is frustrating (there are some other benefits extensions could already use with 0.8.0). 

Apologies for stalling as long here, I do have this in the back of my mind and will at some point push for this, as I do have some other things I want to add/remove. But that will still take some time, sorry for that.

In the meantime, if there is anything else you feel would make sense to add, happy to discuss this and perhaps go ahead with it.

PR #48365, opened 2026-02-04
search: Clone only first match instead of entire match ranges vector

cppcoffee:
Release Notes:

- N/A


MrSubidubi (changes requested):
Can we not just clone the first element and then refactor the entire `if !has_matches` into something like `if let Some(first_match)  = first_match {...} else {...}`?

PR #48447, opened 2026-02-05
terminal: Fix hyperlinks not being detected correctly when preceded by box-drawing chars

Charlie-XIAO:
Closes #46795

This PR aims to correctly detect the hyperlinks in cases like the following, where box-drawing characters immediately precede the path (without spaces in between).

```
╭─[Cargo.toml:55:1]
...
╰────
```

The only false negative with the fix in this PR would be that a file path really contains leading box drawing characters which I think is very unlikely.

---

Release Notes:

- Fixed an issue where the hyperlinks would not be properly detected in the terminal if they were preceded by box-drawing characters (like ─, │, ┌, ┐)


probably-neb:
Makes sense that these shouldn't be expected to be part of a link. Nice work! Thanks!

PR #48449, opened 2026-02-05
gpui: Fix image format comment for `RenderImage`

bowenxuuu:
Release Notes:


rtfeldman:
@bowenxuuu We actually already [convert from RGBA to BGRA](https://github.com/zed-industries/zed/blob/3b81feb7c339b56f6f1b5908e89a860a81f29a1a/crates/gpui/src/color.rs#L26) by the time we store it in this data structure, so the current comment is correct!

Issue #16120, opened 2024-08-12
Large files without newlines (all on one line) cause Zed to hang/crash


giovannicimolin:
Summary: Opening a big JSON file without linebreaks causes Zed to freeze.

**To reproduce:**

1. Open a big JSON file with no linebreaks, example: https://formulae.brew.sh/api/formula.jws.json
2. Zed will freeze/crash.

Also reproducible with:
```sh
dd if=/dev/zero of=null-file-1mb.txt bs=1M count=1
dd if=/dev/zero of=null-file-10mb.txt bs=10M count=1
zed null-file-10mb.txt
```

Workaround: Format the JSON onto multiple lines

```sh
jq . filename > filename-formatted.json
```

### Environment

Zed: v0.146.5 (Zed)
OS: Linux Wayland manjaro unknown
Memory: 15.3 GiB
Architecture: x86_64
GPU: Intel(R) Xe Graphics (TGL GT2) || Intel open-source Mesa driver || Mesa 24.1.5-manjaro1.1

failable:
> (with highlighting disabled).

Have a setting like [`large-file-warning-threshold` in Emacs](https://www.gnu.org/software/emacs/manual/html_node/emacs/Visiting.html#:~:text=If%20you%20try%20to%20visit,file%20literally%20(see%20below).) would be great. If the file is larger than the threshold, the file will be opened literally without any latency and issue.

notpeter:
For me Zed is able to open a 1.2GB text file without crashing, so I wonder if this is JSON specific.  Does Zed still crash if you rename the file from `.json` to `.txt`?

Zed performing poorly with buffers >16M lines long is a known issue:
- https://github.com/zed-industries/zed/issues/5355

wlinna:
For me Zed freezes and crashes for a 166 MiB JSON file just the same regardless of the file extension, so it's not JSON specific. But all that text is on one line, so it might have something to do with that

kanielrkirby:
My guess is that this is to do with line length, and (unless one already exists), I think it simply needs a cap (like VSCode's `editor.stopRenderingLineAt`, and potentially `editor.maxTokenizationLineLength`).

---

I have a `.log` (20MB) file, and around 200-300 lines down, there's a very long line (1,541,369 characters long), so as soon as I open the file it hangs and usually crashes outright (no extensions or custom settings).

I've done a couple tests in some other editors I frequently use:

[...]

PR #48450, opened 2026-02-05
editor: Put more syntax walking features on the background

Veykril:
Currently we always compute breadcrumbs and sticky headers on every editor paint which is not cheap to do especially in bigger files, moving this off to be computed on event handling where they change and then caching them can save serveral milliseconds per render in bigger files. 

This also puts matching brackets refreshing and document highlights on a background task, as this tends to block the main task for prolonged time as well.

Release Notes:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2348,7 +2348,6 @@ version = "0.1.0"
 dependencies = [
  "editor",
  "gpui",
- "theme",
  "ui",
  "workspace",
 ]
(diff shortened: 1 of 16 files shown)


marcocondrache:
This should help the hang in: https://github.com/zed-industries/zed/issues/16120. At least from the profiling I've observed yesterday the main thread is stuck in tree traversal when the sticky headers are enabled

Veykril:
That still does not help enough it seems unfortunately

PR #48452, opened 2026-02-05
agent: Fix missing signatures

rutgercap:
Problem

When using thinking models (like Claude with extended thinking), cancelling a response while the model is still in the thinking phase could cause all subsequent messages in that thread to fail with:

```
Second message failed with error: invalid request format to Anthropic's API: messages.1.content.0: Invalid 'signature' in 'thinking' block
```

This happened because:
1. The model starts streaming thinking content before the signature arrives
2. If the user cancels during this window, the thinking block is saved without a signature
3. On the next request, this incomplete thinking block is sent to the API, which rejects it

## Solution

1. Added `requires_thinking_signature()` method to the `LanguageModel` trait (with no default implementation, forcing explicit consideration for new providers)
2. Filter out incomplete thinking blocks (those without a valid signature) in `flush_pending_message`, but only for providers that require signatures
3. Made Bedrock's `is_anthropic_model()` exhaustive to ensure new model variants are explicitly categorized

### Provider signature requirements:
| Provider | Requires Signature | Reason |
|----------|-------------------|--------|
| Anthropic | ✅ | API rejects thinking blocks with invalid signatures |
| Google | ✅ | Requires `thought_signature` in thinking blocks |
| Bedrock | ✅ (Claude only) | Only Anthropic models on Bedrock use signatures |
| Cloud (Zed.dev) | ✅ (Anthropic/Google) | Routes to providers that require signatures |
| DeepSeek, OpenAI, etc. | ❌ | Don't use thinking signatures |

## Testing
- Added unit test `test_cancelled_thinking_without_signature_excluded_from_request` that verifies incomplete thinking blocks are filtered from API requests
- Added ignored integration test `test_cancelled_thinking_with_real_api` for manual verification with a real API key

## Notes
- The incomplete thinking content will still be visible in the UI during streaming, but won't be persisted or sent to the API after cancellation
- Providers that don't use signatures (DeepSeek, OpenAI, etc.) keep their thinking blocks intact

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -2280,6 +2280,22 @@ impl Thread {
             return;
         };
 
+        // Remove incomplete thinking blocks (those without a valid signature).
+        // This can happen when the user cancels a response mid-stream before
+        // the signature arrives. Without a valid signature, the thinking block
+        // cannot be sent back to the API for providers that require signatures
+        // (e.g., Anthropic, Google). Other providers (e.g., DeepSeek, OpenAI)
+        // ignore signatures entirely.
+        if self
+            .model
+            .as_ref()
+            .is_some_and(|model| model.requires_thinking_signature())
+        {
+            message.content.retain(|content| {
+                !matches!(content, AgentMessageContent::Thinking { signature, .. } if signature.is_none())
+            });
+        }
+
         if message.content.is_empty() {
             return;
         }
--- a/crates/bedrock/src/models.rs
+++ b/crates/bedrock/src/models.rs
@@ -584,6 +584,13 @@ impl Model {
         }
     }
 
+    pub fn is_anthropic_model(&self) -> bool {
+        match self {
+            Model::Custom { name, .. } => name.starts_with("anthropic."),
+            _ => self.id().starts_with("claude"),
+        }
+    }
+
     pub fn cross_region_inference_id(
         &self,
         region: &str,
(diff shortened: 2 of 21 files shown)


crates/agent/src/thread.rs
```diff
@@ -2235,6 +2189,14 @@ impl Thread {
             return;
         };
 
+        // Remove incomplete thinking blocks (those without a valid signature).
+        // This can happen when the user cancels a response mid-stream before
+        // the signature arrives. Without a valid signature, the thinking block
+        // cannot be sent back to the API.
+        // message.content.retain(|content| {
+        //     !matches!(content, AgentMessageContent::Thinking { signature, .. } if signature.is_none())
```
benbrandt: Hmm I wonder if there are providers that may not use signatures and this might filter them out?
I wonder if we need some sort of provider flag for whether or not they use signatures.

crates/bedrock/src/models.rs
```diff
@@ -584,6 +584,89 @@ impl Model {
         }
     }
 
+    /// Returns whether this model is an Anthropic model.
+    /// Anthropic models require thinking signatures when sending thinking blocks back to the API.
+    pub fn is_anthropic_model(&self) -> bool {
```
rutgercap, line 587: I intentionally made this an exhaustive match so when new models are added to the enum people are forced to check this. The downside ofcourse is this huge match arm.

Alternatively we could just have:
matches!(self, anthropic1,anthropic2, ...)
But if a new anthropic model is added, people could forget adding it to this function and then we'll have a hidden bug. 

I'm not sure what is preferred, so let me know!

crates/agent/src/thread.rs
rutgercap: Yes you're right. I only use claude so I didn't try enough others. I updated it with a flag.

PR #48455, opened 2026-02-05
Add multibuffer file-based filtering

DanielMSchmidt:
We already have this capability in the project search view and especially in big projects with lots of tests (like Terraform) it would be super useful if I could just filter out the test cases when trying to understand a code path.

Screenshot 2026-02-05 at 11 05 29

Release Notes:

- Added file filter to multi-buffer views (e.g. Find all References)


DanielMSchmidt:
@MrSubidubi Could you take another look at this PR? 😇

PR #48475, opened 2026-02-05
language_selector: Auto-select current language when opening

cppcoffee:
Release Notes:

- Added Auto-select current language when opening language selector

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9270,6 +9270,7 @@ dependencies = [
  "open_path_prompt",
  "picker",
  "project",
+ "serde_json",
  "settings",
  "ui",
  "util",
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (commented):
Looks like a nice improvement, thank you.

Can we have a test for this one to ensure this does not break in the future?
I think we could create a FakeFS project with 2 files of a different language + an empty buffer (so, 3 different files in general), and we can open the language modal on each such file to assert what's selected.

cppcoffee:
Hi @SomeoneToIgnore , The test unit has been added. Please review it.

SomeoneToIgnore (approved):
Thank you a lot!

PR #48495, opened 2026-02-05
copilot: Fix Copilot not respecting `disable_ai` setting

oliverbarnes:
Closes #48274

Previously, the Copilot language server would continue running even when `disable_ai: true` was set in settings. This change ensures Copilot properly responds to the `disable_ai` setting:

- Add `disable_ai` check in `start_copilot()` to prevent starting when AI is disabled
- Modify the `SettingsStore` observer to shut down the running language server when `disable_ai` changes from false to true
- Add tests for all scenarios:
  - Copilot doesn't start when `disable_ai` is true
  - Copilot stops when `disable_ai` becomes true
  - Copilot can start again when `disable_ai` becomes false

Release Notes:

- Fixed Copilot starting when disabled_ai: true


osiewicz:
Thank you!

PR #48502, opened 2026-02-05
gpui: Allow targeting output for layer shell windows on Wayland

### LINKED ISSUE #46378: GPUI (Wayland): displays() not available during startup, breaks layer-shell output selection
### Reproduction steps

1. Build a minimal GPUI application on Wayland using layer-shell windows.
2. In the app startup / initialization phase, call `AppContext::displays()`.
3. Observe that `displays()` returns an empty array.
4. Create a window.
5. On the next frame / render, call `AppContext::displays()` again.
6. Observe that `displays()` is now populated.

## Current behavior:

On Wayland, `AppContext::displays()` returns an empty list until after at least one window has been created and the render loop has run.

This makes it impossible to select a specific output when creating a layer-shell window, because on Wayland the output must be chosen at creation time and cannot be changed afterwards.

Attempting to work around this by creating and destroying windows during initialization can also lead to arena invalidation panics.

## Expected behavior:

`AppContext::displays()` should return the available Wayland outputs before any window is created, or GPUI should provide a supported mechanism to:

initialize Wayland output discovery independently of window creation, or

delay / stage layer-shell window creation until outputs are known, or

otherwise allow correct output selection for layer-shell use cases.

Zed: N/A (issue reproduced in standalone GPUI application)
OS: NixOS
Window system: Wayland (Hyprland)
Architecture: x86_64

### Attach Zed log file

Not applicable

### Relevant Zed settings

Not applicable.

### Relevant Keymap

Not applicable.

Not applicable.

None

andre-brandao:
Closes #46378

Release Notes:

- Fixed: allow targeting a specific output when creating Wayland layer shell windows.

### DIFF
--- a/crates/gpui/src/platform/linux/wayland/client.rs
+++ b/crates/gpui/src/platform/linux/wayland/client.rs
@@ -201,6 +201,12 @@ pub struct Output {
     pub bounds: Bounds<DevicePixels>,
 }
 
+#[derive(Clone)]
+struct OutputHandle {
+    global_name: u32,
+    wl_output: wl_output::WlOutput,
+}
+
 pub(crate) struct WaylandClientState {
     serial_tracker: SerialTracker,
     globals: Globals,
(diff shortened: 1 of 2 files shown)


andre-brandao:
I’m building a Wayland shell with GPUI (`https://github.com/andre-brandao/gpui-shell`) and ran into this while trying to create a bar per monitor. Without being able to target a specific output, layer shell surfaces end up on the wrong display. This change unblocks that use case.

I’m also considering adding an event for display add/remove, but I’m not sure what the best approach is in GPUI: implement it across all platforms, or add a Linux-only trait extension? I’m still getting comfortable with Rust, so I’d really appreciate guidance on the preferred design here.

PR #48510, opened 2026-02-05
editor: Prevent underlines from appearing in minimap

amtoaer:
I noticed that the minimap seems to render underlines with the same thickness as the main editor, which looks a bit off. This becomes much more noticeable when enabling `semantic_token_rules` (due to the increased number of underlines):

```json
"global_lsp_settings": {
  "semantic_token_rules": [
    {
      "token_modifiers": ["mutable"],
      "underline": true,
    },
  ],
}
```
Looking at the existing code, I found that diagnostic underlines already check `editor_style.show_underlines` to ensure they are only displayed in the main editor. To maintain consistency, I applied the same filtering logic to `chunk_highlight` so that these underlines are no longer rendered in the minimap.

Before：
CleanShot 2026-02-06 at 02 28 31@2x
After：
CleanShot 2026-02-06 at 02 31 36@2x

Release Notes:

- Fixed underlines showing in the minimap.


MrSubidubi (approved):
Sorry for the delay, looks good. Thank you!

MrSubidubi:
Thanks again!

PR #48522, opened 2026-02-05
language: Respect combined injection sub-ranges for language queries

vitallium:
Follow-up to https://github.com/zed-industries/zed/pull/41111.

PR #41111 introduced combined injection handling, but cursor language queries still relied on layer range selection alone. For combined injections this can surface a language from the outer combined layer even when the cursor is outside that language's actual included sub-range. So, we just need to add sub-range filtering based on anchor-aware boundary checks in the same way we did in the previous PR. That means: apply it in `Buffer::language_at`,
`Buffer::languages_at`, and `BufferSnapshot::language_scope_at`. All places that rely on the described behavior.

I also added some additional test cases for the `HTML+ERB` lang to verify language resolution for HTML and Ruby positions. 
Thank you!

Closes https://github.com/zed-industries/zed/issues/48358

Release Notes:

- Respect combined injection sub-ranges for language queries

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -1708,28 +1708,14 @@ impl Buffer {
     /// Returns the [`Language`] at the given location.
     pub fn language_at<D: ToOffset>(&self, position: D) -> Option<Arc<Language>> {
         let offset = position.to_offset(self);
-        let mut is_first = true;
-        let start_anchor = self.anchor_before(offset);
-        let end_anchor = self.anchor_after(offset);
+        let text: &TextBufferSnapshot = &self.text;
         self.syntax_map
             .lock()
-            .layers_for_range(offset..offset, &self.text, false)
+            .layers_for_range(offset..offset, text, false)
             .filter(|layer| {
-                if is_first {
-                    is_first = false;
-                    return true;
-                }
-
                 layer
                     .included_sub_ranges
-                    .map(|sub_ranges| {
-                        sub_ranges.iter().any(|sub_range| {
-                            let is_before_start = sub_range.end.cmp(&start_anchor, self).is_lt();
-                            let is_after_end = sub_range.start.cmp(&end_anchor, self).is_gt();
-                            !is_before_start && !is_after_end
-                        })
-                    })
-                    .unwrap_or(true)
+                    .is_none_or(|ranges| offset_in_sub_ranges(ranges, offset, text))
             })
             .last()
             .map(|info| info.language.clone())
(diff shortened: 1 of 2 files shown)


crates/language/src/buffer.rs
```diff
@@ -1714,28 +1708,14 @@ impl Buffer {
...
             .filter(|layer| {
-                if is_first {
-                    is_first = false;
-                    return true;
-                }
-
                 layer
                     .included_sub_ranges
-                    .map(|sub_ranges| {
-                        sub_ranges.iter().any(|sub_range| {
```
vitallium, line 1732: Extracted that as `offset_in_sub_ranges` since I need it in multiple places.

crates/language/src/buffer.rs
```diff
@@ -1745,10 +1725,17 @@ impl Buffer {
...
     pub fn languages_at<D: ToOffset>(&self, position: D) -> Vec<Arc<Language>> {
         let offset = position.to_offset(self);
+        let text: &TextBufferSnapshot = &self.text;
         let mut languages: Vec<Arc<Language>> = self
             .syntax_map
             .lock()
-            .layers_for_range(offset..offset, &self.text, false)
+            .layers_for_range(offset..offset, text, false)
+            .filter(|layer| {
+                // For combined injections, check if offset is within the actual sub-ranges.
```
vitallium, line 1734: HTML+ERB is quite tricky language since it:

- Defines itself as `HTML+ERB` language
- Injects HTML
- Injects Ruby

Let's take the following example:

```erb
<div>Hello</div>
<%= yield %>
<p>World</p>
```

- HTML+ERB lang is active in `<%|`
- HTML lang is active in `<div>Hello</div>` and `<p>World</p>`.
- Ruby lang is active only inside `<%= yield %>`.

So, when a file mixes languages, HTML+ERB for example, the injected language (Ruby or HTML) is only valid in specific spans, so we check those real sub-ranges at the cursor
  position to avoid reporting the wrong lang and breaking language-specific actions.

crates/language/src/buffer_tests.rs
```diff
@@ -2894,15 +2891,80 @@ fn test_language_at_for_markdown_code_block(cx: &mut App) {
...
+            html_after_ruby,
+            language.name()
+        );
+
+        buffer
+    });
+}
+
+#[gpui::test]
+fn test_languages_at_for_combined_injections(cx: &mut App) {
```
vitallium, line 2959: Copy&pasted both tests, I hope that's OK? :D

Veykril (approved):
Thanks!

PR #48525, opened 2026-02-05
bedrock: Add Claude Opus 4.6

5herlocked:
Release Notes:

- Added Claude Opus 4.6 and 4.6 Thinking with Cross region inference for US, EU, and Global endpoints.

### DIFF
--- a/crates/bedrock/src/models.rs
+++ b/crates/bedrock/src/models.rs
@@ -2,6 +2,27 @@ use serde::{Deserialize, Serialize};
 use strum::EnumIter;
 
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, PartialEq)]
+pub enum BedrockAdaptiveThinkingEffort {
+    Low,
+    Medium,
+    #[default]
+    High,
+    Max,
+}
+
+impl BedrockAdaptiveThinkingEffort {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Low => "low",
+            Self::Medium => "medium",
+            Self::High => "high",
+            Self::Max => "max",
+        }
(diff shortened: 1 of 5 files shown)


rgbkrk (approved):

crates/bedrock/src/models.rs
```diff
@@ -407,7 +447,9 @@ impl Model {
             | Self::ClaudeOpus4Thinking
             | Self::ClaudeOpus4_1Thinking
             | Self::ClaudeOpus4_5
-            | Self::ClaudeOpus4_5Thinking => 200_000,
+            | Self::ClaudeOpus4_5Thinking
+            | Self::ClaudeOpus4_6
+            | Self::ClaudeOpus4_6Thinking => 200_000,
```
rgbkrk, line 452: ~~Max tokens on Opus 4.6 is much higher now right?~~

Nevermind, the 1M token limit is in beta.

5herlocked: only with the Beta header

5herlocked: I can throw that in, but I would also do opt-in on that, the same way we've done for "allow_global"

5herlocked: welp -- #48542

benbrandt (approved):
thank you!

crates/zed/Cargo.toml
```diff
@@ -259,7 +259,7 @@ agent_ui_v2 = { workspace = true, features = ["test-support"] }
 search = { workspace = true, features = ["test-support"] }
 repl = { workspace = true, features = ["test-support"] }
 
-[package.metadata.bundle-dev]
```
5herlocked: @benbrandt -- This is an unintended change -- i don't know how it breaks Zed's Dev / Preview / Stable releases -- feel free to revert this.

PR #48530, opened 2026-02-05
settings_ui: Fix tab navigation in edit predictions settings

### LINKED ISSUE #48391: Settings UI focus moves to text field after using dropdown menu
### Reproduction steps

1. Go to "AI > Edit Predictions > Configure Providers" in the settings UI
2. Select something from the Provider dropdown menu

After using the Provider dropdown menu, focus shifts to the first text field (in this case, the API key for Mercury).

This is bad UI design. The focus should stay on the element I just interacted with.

OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

austincummings:
The provider dropdown and GitHub Copilot sign-in button were not tab-navigable because they lacked tab_index. The copilot button conditionally sets tab_index only when edit_prediction is true, since it's also used in the agent configuration panel where tab navigation isn't used.

Closes #48391

Release Notes:

- Fixed focus skipping provider dropdown and GitHub Copilot button in edit prediction settings

### DIFF
--- a/crates/copilot_ui/src/sign_in.rs
+++ b/crates/copilot_ui/src/sign_in.rs
@@ -568,6 +568,7 @@ impl ConfigurationView {
             .icon_color(Color::Muted)
             .icon_position(IconPosition::Start)
             .icon_size(IconSize::Small)
+            .when(edit_prediction, |this| this.tab_index(0isize))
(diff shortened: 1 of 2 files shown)


danilo-leal (approved):
Thank you so much! 🙌

PR #48542, opened 2026-02-06
bedrock: Enable 1M context window

5herlocked:
Release Notes:

- Added `allow_extended_context` to the Bedrock settings which enables 1M context windows on models that support it


rgbkrk (approved):

bennetbo (approved):
Thank you!

PR #48550, opened 2026-02-06
Fix settings migrations for nested platform/channel/profile keys

rtfeldman:
Previously, some settings migrations only operated on root-level keys and missed settings nested under platform keys (`macos`, `linux`, etc.), channel keys (`nightly`, `stable`, etc.), or profile blocks. This fixes migrations to recurse into those nested locations.

Also fixes `m_2026_02_02` to gracefully skip when `edit_predictions` is not an object (e.g. `true`) instead of bailing and aborting the entire migration chain.

Release Notes:

- Fixed settings migrations to correctly handle settings nested under platform, channel, or profile keys.


crates/migrator/src/patterns/settings.rs
```diff
@@ -108,24 +108,3 @@ pub const SETTINGS_DUPLICATED_AGENT_PATTERN: &str = r#"(document
     (#eq? @agent1 "agent")
     (#eq? @agent2 "agent")
 )"#;
-
-/// Migrate language settings,
-/// calls `migrate_fn` with the top level object as well as all language settings under the "languages" key
-/// Fails early if `migrate_fn` returns an error at any point
-pub fn migrate_language_setting(
```
rtfeldman, line 115: This function is now in migrator.rs (and is revised)

PR #48551, opened 2026-02-06
git_ui: Clear commit message before generating AI text

cppcoffee:
Release Notes:

- Improved clear commit message before generating AI text.

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -2612,6 +2612,12 @@ impl GitPanel {
             return;
         };
 
+        self.commit_message_buffer(cx).update(cx, |buffer, cx| {
+            let start = buffer.anchor_before(0);
+            let end = buffer.anchor_after(buffer.len());
+            buffer.edit([(start..end, "")], None, cx);
+        });
+
         telemetry::event!("Git Commit Message Generated");
 
         let diff = repo.update(cx, |repo, cx| {


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR clears the commit message buffer before generating an AI commit message. This prevents accumulation of previous commit messages when the AI commit message generation is triggered multiple times.

## Architecture

The change is minimal and targeted:
- In `git_panel.rs`, before the AI commit message generation process begins
- Clears all content from the commit message buffer by editing it to an empty string
- Uses the buffer's `edit` method with the full range (0 to buffer length)

This ensures a clean slate for each AI-generated commit message.

## Issues

### 1. Potential UX Issue - User Input Loss (Moderate)

If a user has manually typed or edited the commit message buffer and then triggers AI generation, their work will be **completely lost**. This could be frustrating if:

- User partially writes a commit message
- User decides to use AI to complete/generate the message
- All their manual input is cleared before the AI generation starts

**Potential solutions:**
1. Add a confirmation dialog if the buffer is not empty
2. Only clear the buffer if it contains AI-generated content (not user input)
3. Keep the user's input and have the AI append/modify it instead

Consider the user workflow: is it common to trigger AI generation after manual edits? If so, this change could be disruptive.

### 2. No Test Coverage (Minor)

No tests were added to verify:
- The buffer is properly cleared before AI generation
- Subsequent AI generations start with an empty buffer
- The edit operation doesn't cause any side effects

**Action**: Consider adding a test case that:
- Populates the commit message buffer
- Triggers AI generation
- Verifies the buffer is cleared

### 3. Telemetry Event Timing (Minor)

The telemetry event `Git Commit Message Generated` (line 2700) fires **before** the buffer is cleared in this PR's diff. However, looking at the diff more carefully, it seems the clear operation is inserted **before** the telemetry event.

Wait - re-reading the diff, the clear operation (lines 9-13) is indeed inserted **before** the telemetry event (line 15 in diff). This is correct - the buffer is cleared first, then telemetry fires.

However, you might want to consider adding a separate telemetry event to track when buffers are cleared, which could help understand how often this situation occurs.

## Cross-file Impact

- **git_panel.rs**: 6 lines added, no breaking changes
- The change is localized to the `generate_commit_message` function
- No changes to public APIs or data structures

## Assessment

**✅ APPROVE with suggestions**

This is a simple, well-intentioned fix that addresses a real issue (accumulation of commit messages). The implementation is clean and follows the existing code patterns.

However, I recommend addressing the UX concern about clearing user input:

**Before merging:**
1. Consider if this UX pattern is acceptable (clearing all buffer content)
2. If clearing is desired, consider adding a confirmation when buffer has user content

**Optional improvements:**
3. Add test coverage for buffer clearing behavior
4. Consider adding telemetry for buffer clearing events

The code is safe to merge as-is if the team has decided that clearing all buffer content before AI generation is the desired behavior. If there's concern about losing user input, that should be addressed before merge.

ChristopherBiscardi:
I don't think it makes sense to remove/delete user data just because input is being generated.

PR #48570, opened 2026-02-06
auto_update: Strip pre-release and build metadata from release notes URL

RubenFricke:
Without this fix, the zed.dev server doesn't recognize the version with metadata and redirects to /releases, which
  defaults to the Stable channel, so Preview users end up seeing Stable release notes.

Closes #47539

  Changes:
  - Strip pre and build metadata from the semver version in release_notes_url before constructing the path, so the URL
  becomes /releases/preview/0.218.0 instead of /releases/preview/0.218.0-beta.1+preview.131.68e98a53.
  - Standardized the ordering of pre/build clearing to pre-first across all 4 call sites (2 were build-first).

  Open question: With this fix, the URL now points to the specific version you're running (e.g.
  /releases/preview/0.218.0). An alternative would be to link to /releases/preview (no version) so users can see all
  Preview releases, including ones newer than what they have installed. I went with per-version linking since it matches
  the existing intent of the code, but happy to change it if the team prefers the other approach.

  Release Notes:

  - Fixed "View Release Notes" in Preview/Stable opening the wrong release channel page due to build metadata in the URL.

### DIFF
--- a/crates/auto_update/src/auto_update.rs
+++ b/crates/auto_update/src/auto_update.rs
@@ -260,7 +260,9 @@ pub fn release_notes_url(cx: &mut App) -> Option<String> {
         ReleaseChannel::Stable | ReleaseChannel::Preview => {
             let auto_updater = AutoUpdater::get(cx)?;
             let auto_updater = auto_updater.read(cx);
-            let current_version = &auto_updater.current_version;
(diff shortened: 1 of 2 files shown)


osiewicz (approved):
Sorry, I've simply been underperforming as a reviewer recently. Your PR looks good to me. Thank you!

PR #48295, opened 2026-02-03
Motion system to make Zed feel fluid and intentional

srbsingh3:
## Problem

Zed feels snappy but mechanical - panels and modals appear instantly, losing spatial context.

## Demo

All five surfaces with reduce_motion toggle:

https://github.com/user-attachments/assets/a6b37599-371e-4eac-bee5-22567c095209

## Changes

| Surface | Animation |
|---------|-----------|
| Panels (docks & utility) | Slide in from edge |
| Modals | Fade + subtle slide |
| Popover menus | Fade + subtle slide |
| Picker list | Smooth selection tracking |

Plus a `reduce_motion` setting that respects macOS accessibility preferences by default.

## Motion principles

- **150ms open, 100ms close** - entrances can breathe; exits stay snappy
- **ease_out_cubic for panels/modals** - responsive start, gentle settle
- **ease_in_out for picker** - selection travels between items, needs acceleration and deceleration
- Consistent timing across all surfaces

## Accessibility

New setting: `reduce_motion: "system" | "on" | "off"`

Defaults to `system`, which reads the macOS preference. When enabled, animations skip gracefully.

## Implementation notes

- Built on GPUI's `Animation` and easing primitives
- Timing centralized as constants for easy tuning
- Animations cancel cleanly on rapid toggle - no stacking or glitches
- 29 unit tests covering state machines and edge cases

> **Note:** In practice, I'd ship this as smaller, incremental PRs (one surface at a time). This is bundled as a single prototype to show the complete motion system working together.

PR #48577, opened 2026-02-06
Add reduce motion accessibility setting

srbsingh3:
Follows up on #48295, scoped down per feedback to ship one primitive at a time.

## Summary

- Adds a `reduce_motion` setting (`"system"` | `"on"` | `"off"`, default: `"system"`)
- Queries macOS `accessibilityDisplayShouldReduceMotion` when set to `"system"`
- Applies the setting to toast slide-in animations as a first consumer
- Adds the setting to the Appearance page in Settings UI

## Next steps

- [ ] Add platform support for Linux (`gtk-enable-animations` / portal) and Windows (`SPI_GETCLIENTAREAANIMATION`)
- [ ] Gate more animation call sites (`with_animation` is used in ~15 other places: cursor blink, edit predictions, agent panel, spinners, etc.)
- [ ] Consider lifting the check into the `animate_in` trait itself so all callers get it for free

## Test plan

- [x] Unit tests for serialization, round-trip, aliases, and settings store integration
- [x] GPUI tests for `should_reduce_motion` with each variant
- [ ] Manual: set `"reduce_motion": "on"` and verify toast appears without animation
- [ ] Manual: set `"reduce_motion": "system"` and toggle macOS System Settings > Accessibility > Display > Reduce motion

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -190,6 +190,16 @@
   //
   // Default: "client"
   "window_decorations": "client",
+  // Whether to reduce motion in UI animations.
+  // 1. Follow the OS accessibility setting
+  //    "system"
+  // 2. Always reduce motion (skip animations)
+  //    "on"
+  // 3. Never reduce motion (always animate)
+  //    "off"
+  //
+  // Default: "system"
+  "reduce_motion": "system",
   // Whether to use the system provided dialogs for Open and Save As.
   // When set to false, Zed will use the built-in keyboard-first pickers.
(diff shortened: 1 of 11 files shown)


SomeoneToIgnore:
Thank you for the PR: seems that it got a bunch of conflicts meanwhile, and I've decided to use a part of the idea to reduce all spinners' motion in https://github.com/zed-industries/zed/pull/60614

That leaves out system detection and presumably something else, which can be followed-up if you're interested.

PR #48584, opened 2026-02-06
mistral: Add x-affinity header

vianney:
Mistral Vibe, the official coding agent from Mistral, sets the x-affinity header to the session ID to enable prompt caching. This patch implements the same mechanism, resulting in a faster agent loop.

Release Notes:

- Added prompt caching for Mistral AI.

### DIFF
--- a/crates/language_models/src/provider/mistral.rs
+++ b/crates/language_models/src/provider/mistral.rs
@@ -225,6 +225,7 @@ impl MistralLanguageModel {
     fn stream_completion(
         &self,
         request: mistral::Request,
+        affinity: Option<String>,
         cx: &AsyncApp,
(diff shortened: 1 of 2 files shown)


crates/language_models/src/provider/mistral.rs
```diff
@@ -225,6 +225,7 @@ impl MistralLanguageModel {
         &self,
         request: mistral::Request,
         cx: &AsyncApp,
+        affinity: Option<String>,
```
maxdeviant, line 228: `cx’ should be the last parameter to the function.

vianney: Fixed.

benbrandt (approved):
thanks!

PR #48585, opened 2026-02-06
gpui: Impl `Cow<'static, str>` to `IntoElement`

huacnlee:
Release Notes:

- N/A

Make this change to let `div().child(...)` support any types that have impl `Into<SharedString>`, for example: `Cow<'static, str>`.

```rs
// From somewhere get a `Cow<'static, str>` type.
let label = Cow::Borrowed("Hello");

div()
  .child(label);
```

### DIFF
--- a/crates/gpui/src/elements/text.rs
+++ b/crates/gpui/src/elements/text.rs
@@ -84,6 +84,14 @@ impl IntoElement for String {
     }
 }
 
+impl IntoElement for Cow<'static, str> {
+    type Element = SharedString;
+
+    fn into_element(self) -> Self::Element {
+        self.into()
+    }
+}
+
 impl Element for SharedString {
     type RequestLayoutState = TextLayout;
     type PrepaintState = ();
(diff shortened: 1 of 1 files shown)


crates/gpui/src/elements/text.rs
```diff
@@ -134,11 +118,11 @@ impl Element for SharedString {
     }
 }
 
-impl IntoElement for SharedString {
-    type Element = Self;
+impl<T: Into<SharedString>> IntoElement for T {
```
osiewicz: I'm a bit uneasy about having a blanket impl like that, because we can have at most one (due to overlap rules for blanket impls).

huacnlee: So, how about I change to just add the `Cow<'_, str>` for IntoElement?

I have a library [rust-i18n](https://crates.io/crates/rust-i18n), the `t!` method returns `Cow<'_, str>` type. GPUI have this supports will let easy to use.

```diff
use rust_i18n::t;

div()
-  .child(SharedString::from(t!("hello_world")))
+  .child(t!("hello_world"))
```

osiewicz: Yeah, adding an impl for Cow<'_, str> seems fine.

huacnlee:
Updated.

PR #48588, opened 2026-02-06
agent: Fix edit_file tool description

vianney:
This patch updates the description of the `edit_file` tool to recommend models to use the `move_path` tool instead of the `terminal` tool to move or rename files.  The `move_path` tool is more specific and shall be preferred as it provides a better UI and dedicated permissions.

Release Notes:

- N/A


benbrandt (approved):
Thanks!

PR #48592, opened 2026-02-06
vim: Make `:bdelete` use new `workspace::CloseItemInAllPanes` command 

baldwindavid:
## Summary

Makes `:bd` (`:bdelete`) close a buffer across all panes where it's open, rather than just the current pane.

This was mentioned in the "Future Work" section of the recently merged #47079 as a supporting feature. 

This matches Helix's `:bc` (`:bclose`) behavior. In vanilla Vim, `:bdelete` closes all windows displaying the buffer, which many users find disruptive—hence popular plugins and keybinding workarounds that preserve window layout. This implementation aligns with this behavior: the buffer is removed from all panes, but panes remain open (showing another tab, or closing only if empty). This is also a better fit for Zed since buffers are inextricably tied to panes so closing the entire pane may mistakenly close a bunch of buffers.

Also refactors the tab switcher's close button to leverage this command.

Release Notes:

- Vim: `:bd` (`:bdelete`) now closes the file in all panes where it's open
- Added `workspace::CloseItemInAllPanes` action to close a file across all panes


dinocosta (approved):
Thank you @baldwindavid ! 🙂

dinocosta (approved):

Issue #31504, opened 2025-05-27
Terminal spawned as a separate entity


roworu:
### Summary

When I start a debugging session, the `console` setting (whether "externalTerminal" or "integratedTerminal") has no effect. The debugger always spawns a new terminal that appears and behaves like a normal terminal, but is isolated from my existing terminal tabs.

### Description

Steps to trigger the problem:
1. Start a debugging session that prints something to the terminal.
2. Try to open a new tab in debugging terminal

Actual Behavior:
----
The debugging session opens a new terminal, separate from any existing terminal group.
This terminal looks and behaves just like a normal one.
When I try to create a new tab from this debugger terminal, it opens in the normal terminal group, not within the debugger's.

Expected Behavior:
----
The debugger terminal should open in the same terminal group as normal terminals.
It could also include a small bug icon or some kind of visual indicator to distinguish it from regular terminal sessions.

I've also attached a video demonstrating the behavior. Please, let me know if any additional info may be provided to help resolve that issue. Thanks!

https://github.com/user-attachments/assets/c8b84d91-bbed-4b9e-8ec2-3eea68053e80

### Zed Version and System Specs

Zed: v0.188.1 (Zed Preview) 
OS: Linux Wayland fedora 42
Memory: 31.3 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3070 || NVIDIA || 570.153.02

P1n3appl3:
Just to confirm: you _can_ still move the terminal from the debug dock to the terminal dock right? I think we'd actually prefer to keep it in the debug panel by default, I find that it's helpful to keep it associated with the other debug panes in there. It's also worth noting that you can set the terminal dock and the debug dock to the same side (`bottom` for example) and then it'd be really unfortunate if starting a debug session switched you _away_ from the debug dock to show the new terminal there.

Zed currently doesn't support the `externalConsole` setting, so the program will always use an integrated terminal.

Regarding the behavior you showed at the end where opening a new terminal from the toolbar dropdown caused it to open in the terminal dock: that's currently the same behavior as if you opened a terminal in the main editor pane (for example by moving one from the terminal dock there, or by using `workspace::NewCenterTerminal`). I can see how you'd find that surprising, and feel free to open a separate issue for that (perhaps we could make the dropdown button spawn a new center terminal when clicked outside of the dock?).

P1n3appl3:
[...]

PR #48609, opened 2026-02-06
terminal_panel: Cmd+N opens new terminal where you are already working

srbsingh3:
Related to #31504

## Problem

If you drag a terminal into the center editor area as a tab, pressing `Cmd+N` still opens the new terminal in the terminal panel instead of next to the tab you're looking at.

## Solution

`Cmd+N` now checks whether the center pane is focused and its active tab is a terminal. If both are true, the new terminal opens there. All other behavior is unchanged.

https://github.com/user-attachments/assets/01036cb4-5f1f-456e-900e-eed2e53d61f5

## Test plan

- [ ] Drag a terminal into the center editor area. Focus it. Press Cmd+N. New terminal opens as a center tab.
- [ ] Toggle terminal panel with Ctrl+`. Press Cmd+N. Opens in panel as before.
- [ ] Focus a regular code file. Press Cmd+N. Opens in panel as before.
- [ ] Have a terminal in the center, but focus the terminal panel. Press Cmd+N. Opens in panel as before.

Release Notes:

- Improved `terminal: New Terminal` (Cmd+N) to open in the center pane when a terminal tab there is focused, instead of always opening in the terminal panel.

### DIFF
--- a/crates/terminal_view/src/terminal_panel.rs
+++ b/crates/terminal_view/src/terminal_panel.rs
@@ -653,6 +653,27 @@ impl TerminalPanel {
         window: &mut Window,
         cx: &mut Context<Workspace>,
     ) {
+        let center_pane = workspace.active_pane();
+        let center_pane_has_focus = center_pane.focus_handle(cx).contains_focused(window, cx);
+        let active_center_item_is_terminal = center_pane
+            .read(cx)
+            .active_item()
+            .is_some_and(|item| item.downcast::<TerminalView>().is_some());
+
+        if center_pane_has_focus && active_center_item_is_terminal {
+            let working_directory = default_working_directory(workspace, cx);
+            let local = action.local;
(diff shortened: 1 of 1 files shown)


crates/terminal_view/src/terminal_panel.rs
```diff
@@ -659,6 +659,29 @@ impl TerminalPanel {
             return;
```
SomeoneToIgnore: Should the new approach also open the center terminal in this case?

SomeoneToIgnore (commented):
Nice change, thank you.

Let's do another small update of the logic while we're here and merge this.

crates/terminal_view/src/terminal_panel.rs
srbsingh3: Moved it above the guard, thanks for catching this.

SomeoneToIgnore (approved):
Thanks!

PR #48616, opened 2026-02-06
repl: Fix duplicate output inside notebooks

MostlyKIGuess:
- Render the output only when needed, fixes the duplicate output that can happen after opening a saved notebook.
- Vim in Jupyter View with j/k navigation across notebook cells

Release Notes:

- N/A


rgbkrk (approved):

rgbkrk (approved):

MrSubidubi (approved):
Thanks!

PR #48622, opened 2026-02-06
Improve performance when detecting JSX auto close

marcocondrache:
Helps #48601

image

`syntax_layers` does some offset conversion that might require getting some chunks from the rope, which is quite expensive. For detecting autoclose, we only use the language from those syntax layers, so having a short path that skips all the conversion should skip some sum_tree traversals. 
   
I'm pretty sure other places would benefit from this as well, but I haven't searched them yet.

Release Notes:

- N/A


smitbarmase (approved):
Thanks!

PR #48629, opened 2026-02-06
languages: Add syntax highlighting for HTML character references

br-schneider:
HTML character references like `&middot;`, `&apos;`, and `&#123;` are correctly parsed by tree-sitter as named nodes (`html_character_reference` in TSX/JavaScript, `entity` in HTML), but no highlight query captures them. This means they render as plain, unhighlighted text in the editor.

This PR adds one-line highlight captures for each:

- **TSX** (`crates/languages/src/tsx/highlights.scm`): `(html_character_reference) @string.special`
- **JavaScript** (`crates/languages/src/javascript/highlights.scm`): `(html_character_reference) @string.special`
- **HTML** (`extensions/html/languages/html/highlights.scm`): `(entity) @string.special`

`@string.special` is already styled by all built-in themes (One Dark, Ayu, Gruvbox, etc.), so no theme changes are needed.

Release Notes:

- Added syntax highlighting for HTML character references (`&middot;`, `&apos;`, `&#123;`, etc.) in TSX, JavaScript, and HTML files.

### DIFF
--- a/crates/languages/src/javascript/highlights.scm
+++ b/crates/languages/src/javascript/highlights.scm
@@ -360,3 +360,4 @@
 (jsx_self_closing_element (["<" "/>"]) @punctuation.bracket.jsx)
 (jsx_attribute "=" @punctuation.delimiter.jsx)
 (jsx_text) @text.jsx
+(html_character_reference) @string.special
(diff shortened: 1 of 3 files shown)


yeskunall (commented):
Did you test this with an HTML file? I get syntax highlighting for JSX/TSX files, but not HTML files. I tested with all the default themes Zed ships with, so the ones in your list should have worked. Maybe I’m missing something?

yeskunall (approved):

yeskunall (approved):

PR #48652, opened 2026-02-06
editor: Use buffer_font for folds and change foreground color

xdBronch:
re: https://github.com/zed-industries/zed/pull/48624
using the UI font could cause a visual bug when the cursor was over the folded text
before:
image
after:
image

changing the color is of course just opinionated but i think it looks better and makes more sense as a placeholder

Release Notes:

- N/A


SomeoneToIgnore (approved):
Nice catch, thank you: this means all previous `…` were having the same issue?

xdBronch:
i guess so, i never noticed before because normally the ellipses are at/past the end of the line so the cursor never ends up over it just from moving to the right

PR #48664, opened 2026-02-07
feedback: Add action for listing installed extensions

mchisolm0:
Closes #48037

- [X] Tests or screenshots needed? <--- Let me know if tests are needed
- [ ] Code Reviewed
- [X] Manual QA

Release Notes:

- Added action to copy installed extensions to clipboard

<details><summary>Picture of Action</summary>Screenshot from 2026-03-03 01-00-22</details>
<details><summary>Picture of Modal</summary>
Screenshot from 2026-03-03 01-00-35
</details>
<details><summary>Picture of Docs</summary>
Screenshot from 2026-03-03 01-19-53
</details>


crates/feedback/src/feedback.rs
```diff
@@ -1,3 +1,4 @@
+use extension_host::{CopyInstalledExtensionsIntoClipboard, installed_extensions_for_clipboard};
```
MrSubidubi: This action should live within this crate (so just below) given that the listener for it also is here.

crates/extension_host/src/extension_host.rs
MrSubidubi, line 1: I think all of this should live in the feedback crate, as this is not at all directly related to the extension host.

crates/extension_host/src/extension_host.rs
```diff
@@ -429,6 +431,28 @@ impl ExtensionStore {
         &self.extension_index.extensions
     }
 
+    fn format_installed_extensions_for_clipboard(&self) -> String {
+        let mut lines = Vec::with_capacity(self.extension_index.extensions.len());
+
+        for (extension_id, entry) in self.extension_index.extensions.iter() {
+            let dev_suffix = if entry.dev { " (dev)" } else { "" };
+            lines.push(format!(
+                "{} ({}) v{}{}",
```
MrSubidubi: I think we should make this list sorted alphabetically here (perhaps with the dev extensions at the top, not sure though) and then have a `-` at the start of every line so this is more pleasant to read and to take a glimpse at.

crates/extension_host/src/extension_host.rs
```diff
@@ -1846,6 +1870,12 @@ impl ExtensionStore {
     }
 }
 
+pub fn installed_extensions_for_clipboard(cx: &App) -> String {
+    ExtensionStore::try_global(cx)
```
MrSubidubi: Should be fine to just `ExtensionStore::global(cx)` it here.

MrSubidubi (changes requested):
Thanks for taking this up, left some notes.

crates/extension_host/src/extension_host.rs
mchisolm0: I love all of these suggestions! 😁 I just used the `.sort()` method. I think that will mean it is case-sensitive, but I felt that was worth it to avoid unnecessarily complicating the implementation.

I think this is still reasonable sorting from a user's perspective, since I think all the extensions start capitalized. Do you agree?

<details><summary>Screenshot of new, sorted extensions</summary>CleanShot 2026-03-12 at 22 46 08@2x</details>

crates/feedback/src/feedback.rs
```diff
@@ -96,3 +110,43 @@ pub fn init(cx: &mut App) {
     })
     .detach();
 }
+
+pub fn installed_extensions_for_clipboard(cx: &App) -> String {
+    let store = ExtensionStore::global(cx);
+    format_installed_extensions_for_clipboard(store.read(cx))
+}
+
+fn format_installed_extensions_for_clipboard(store: &ExtensionStore) -> String {
```
MrSubidubi: Let's just make this one method, please

crates/feedback/src/feedback.rs
```diff
@@ -96,3 +110,43 @@ pub fn init(cx: &mut App) {
...
+            entry.manifest.name,
+            extension_id,
+            entry.manifest.version,
+            if entry.dev { " (dev)" } else { "" }
+        );
+        if entry.dev {
+            top_lines.push(line);
+        } else {
+            bottom_lines.push(line);
+        }
```
MrSubidubi: Apologies for the back and forth here, but let's just sort the list alphabetically and not consider dev extensions separately. My bad here.

crates/feedback/src/feedback.rs
```diff
@@ -96,3 +110,43 @@ pub fn init(cx: &mut App) {
...
+    }
+
+    top_lines.sort();
+    bottom_lines.sort();
+
+    top_lines.extend(bottom_lines);
+
+    if top_lines.is_empty() {
+        return "Installed extensions: none".to_string();
```
MrSubidubi:
```suggestion
        return "No extensions installed.".to_string();
```

MrSubidubi (changes requested):

crates/feedback/src/feedback.rs
mchisolm0: No worries at all. I would rather get it the way we want it. Happy to make the changes. 😁

MrSubidubi (approved):
Sorry for the long delay, very nice! Thank you so much!

mchisolm0:
No worries at all! I knew you would get to it when you had the time 😁

PR #48669, opened 2026-02-07
Add a `default_diff_view` setting

JosephTLyons:
- [X] Tests
- [ ] Code Reviewed
- [X] Manual QA

For the person reviewing:

- Opinions on the setting / option names?

I didn't add the setting to settings editor in this PR, figured I'd do that in a follow up.
I kept the default as stacked, but we might want to consider the results [here](https://x.com/zeddotdev/status/2019994065039667311?s=20).

Release Notes:

- N/A


yeskunall:
While you’re at it @JosephTLyons do you think you can update the docs as well? If not, you can ping me and I’ll do it.

Issue #22926, opened 2025-01-09
vim mode: Incorrect ligature splitting


niekdomi:
Although I don't think this is an issue exclusive to vim-mode, apparently, based on the comments I received, it's more vim related...

So if the cursor is placed under a ligature, the ligature should split, this makes it easier to edit the character under the ligature e.g.,

<img width="154" height="58" alt="Image" src="https://github.com/user-attachments/assets/cc647e3c-c9c8-4fd2-8700-9359169be41c" />

So as you can see, the spaceship operator is split into its characters if the cursor is placed under any character of that ligature. 

Though, in zed, only the character that is **under** the cursor gets de-rendered, the remaining part of the ligature is contained:

<img width="171" height="63" alt="Image" src="https://github.com/user-attachments/assets/4b603c71-614a-4d17-9dc9-81cf496c1727" />

A better example may be "less-equal":

<img width="171" height="63" alt="Image" src="https://github.com/user-attachments/assets/0de70268-3ead-4474-a2b9-0cd30888d622" />

As you can see, the `=` is still shown as the right part of `<=`

ThatOneCalculator:
+1 on this, VSCode handles this pretty well imo.

https://github.com/user-attachments/assets/67a66930-86a8-494b-8051-53110281c241

niekdomi:
not really, vscode has exactly the same issue

![Image](https://github.com/user-attachments/assets/6fae87b8-381b-4388-ae53-f14c9f1ffe9c)

ThatOneCalculator:
That's with the block cursor, not the line cursor

niekdomi:
I think you misunderstand the issue. If I watch your clip the ligatures do **not** split at all. This is (as I just tested) something that only appears using block cursor (vim-mode).

As you can see in this image (and the initial issue...) the ligature is split in its characters.

![Image](https://github.com/user-attachments/assets/7fd8cb1b-b4ab-4c10-b0b1-96006e245e9c)

or 

![Image](https://github.com/user-attachments/assets/e5ce70bc-9a1c-42ff-9ec2-63214d22c3cf)

ThatOneCalculator:
Ahh, gotcha. Out rn, but I'll file a separate issue later today for cursors going inside ligatures.

niekdomi:
btw what you're requesting, seems to work for me 

https://github.com/user-attachments/assets/03d576c0-bdb9-4e9e-a8e9-499e7ef64bbd

[...]

PR #48677, opened 2026-02-07
gpui: Improve handling of x position for index

### LINKED ISSUE #31332: Cursor appears to the right of ligature instead of in the middle
### Summary

When you position the cursor in between characters that create a ligature, Zed shows the cursor to the right of the ligature.

### Description

Steps to reproduce:

1. Install Monaspace (`brew install font-monaspace`)
2. Use the font (`"buffer_font_family": "Monaspace Argon"`) and enable some stylistic sets (e.g. `"buffer_font_features": { "ss01": true }`)
3. Open a buffer and type characters that form a ligature, e.g. `...` or `==`
4. Move the cursor in between the characters

Expected Behavior: The cursor appears in the middle of the ligature, and ideally the ligature breaks (shows its individual characters)
Actual Behavior: The cursor appears on the right of the ligature

Zed: v1.4.4+stable.297.1d0b482638588e5507184d91631e30596339ee1b (Zed) 
OS: macOS 15.7.7
Memory: 16 GiB
Architecture: x86_64

garthdw:
Ligatures and text run calculations can end up with glyph indices with the same value.

Now we will keep track of potential matches and count across the same indices.

Appears to be an issue in how we split into text runs and the underlying CoreText on MacOS at least.

Partially closes #31332

https://github.com/user-attachments/assets/987306c2-3334-446e-b7e5-97db4c628ac0

- [x] Code Reviewed
- [x] Manual QA

Release Notes:

- Fixed cursor position in ligatures

### DIFF
--- a/crates/gpui/src/text_system/line_layout.rs
+++ b/crates/gpui/src/text_system/line_layout.rs
@@ -103,13 +103,56 @@ impl LineLayout {
 
     /// The x position of the character at the given index
     pub fn x_for_index(&self, index: usize) -> Pixels {
+        // When ligatures are enabled, the index of the glyphs can
+        // change and become repeating values.
+        // e.g. 5, 5, 7, for `==` in a line.
+        // We want to be able to track the offset and the potential match
+        // if the passed index is inside a ligature.
+        // Only happens for keywords/end of lines and not inside text.
+        let mut potential_glyph: Option<&ShapedGlyph> = None;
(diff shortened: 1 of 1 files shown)


niekdomi:
This is a fix for  https://github.com/zed-industries/zed/issues/22926#issuecomment-2973042622, though it doesn't close the actual issue. I will update/simplify the issue description since there seems to be some misunderstanding.

PR #48685, opened 2026-02-07
language: Return early if no grammars are added

marcocondrache:
Helps #48601

Whenever an extension is installed, we call `register_grammars` even when the grammar list is empty. This unnecessarily increments reload_count and notifies the LSP store, which clears all languages and triggers a full reparse.

Clearing languages also emits `LanguageChanged` events for buffers, causing the editor to perform expensive recomputations (like https://github.com/zed-industries/zed/pull/48622) which can block the main thread for large multibuffers.

This PR addresses the empty-grammar case. If an extension actually adds a grammar, the underlying issue still exists and will require additional fixes to fully resolve.

- [ ] Tests or screenshots needed?
- [x] Code Reviewed
- [ ] Manual QA

Release Notes:

- Fixed an issue where installing theme extensions could block the main thread

### DIFF
--- a/crates/language/src/language_registry.rs
+++ b/crates/language/src/language_registry.rs
@@ -552,15 +552,16 @@ impl LanguageRegistry {
     }
 
     /// Adds paths to WASM grammar files, which can be loaded if needed.
-    pub fn register_wasm_grammars(
-        &self,
-        grammars: impl IntoIterator<Item = (impl Into<Arc<str>>, PathBuf)>,
-    ) {
+    pub fn register_wasm_grammars(&self, grammars: Vec<(Arc<str>, PathBuf)>) {
+        if grammars.is_empty() {
+            return;
+        }
+
         let mut state = self.state.write();
         state.grammars.extend(
             grammars
                 .into_iter()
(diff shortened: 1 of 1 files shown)


MrSubidubi (commented):
As so often, nice catch and nice digging!

I just did a quick check and it seems that https://github.com/zed-industries/zed/blob/e8030aea4a22dc3521b6b5911a87d986406935f3/crates/language_extension/src/language_extension.rs#L41-L43 is the only caller of this currently. As I see it, we could perhaps even get rid of acquiring the lock if we instead check upfront if the Vec is empty by either handling that at the call site or chaning the argument type, can we not?

MrSubidubi (commented):
Just checked again, sorry for the back and forth and the nit here. I think it would be better if we were to add this check at the level of the language registry - https://github.com/zed-industries/zed/blob/e8030aea4a22dc3521b6b5911a87d986406935f3/crates/language/src/language_registry.rs#L1251-L1258 does this also within the registry state  and I think we should not handle checking this differently for something that is similar here. So let this just take a Vec or alternatively a Peekable so that we handle this the same for all cases.

MrSubidubi (approved):
Again, nice find, thank you so much and thanks for the follow-ups!

PR #48696, opened 2026-02-07
Fix Thai character rendering in terminal

oatsaysai:
- [x] Manual QA
- [ ] Code Reviewed

## Summary

Fix incorrect rendering of Thai text in the integrated terminal. Thai combining characters (vowels, tone marks, and diacritics) were being treated as regular characters occupying their own cells, causing misaligned and broken text output.

## Problem

Thai uses combining marks (e.g. ิ ี ึ ื ุ ู ่ ้ ๊ ๋) that must be rendered on top of or below a preceding consonant, not in a separate cell. Previously, the terminal treated every character including these combining marks as occupying one full cell width. This caused Thai text like `สวัสดี` to appear garbled, with vowels and tone marks displaced from their consonants.

## Solution

Two changes in `terminal_element.rs`:

1. **Skip cell count increment for Thai combining marks:** Added `is_thai_combining_mark()` to identify Unicode code points that are Thai combining characters (U+0E31, U+0E34–U+0E3A, U+0E47–U+0E4E). When one of these characters is encountered during batching, it is appended to the current text run *without* incrementing `cell_count`, so it overlays the preceding consonant's cell instead of occupying its own.

2. **Disable fixed-width cell shaping for Thai text:** When a `BatchedTextRun` contains Thai characters (U+0E01–U+0E5B), the `cell_width` parameter passed to `shape_line` is set to `None` instead of `Some(dimensions.cell_width)`. This allows the text shaper to freely position combining marks relative to their base consonants rather than forcing every glyph into a fixed-width grid.

## Test

Tested by running Thai text in the integrated terminal:

```
echo "สวัสดีครับ ภาษาไทย ก้อนหิน น้ำตก"
```

### before
image

### after
image

## Release Notes
- Fixed Thai character rendering in the integrated terminal, where combining vowels and tone marks were incorrectly occupying separate cells instead of overlaying their consonants.

### DIFF
--- a/crates/terminal_view/src/terminal_element.rs
+++ b/crates/terminal_view/src/terminal_element.rs
@@ -145,13 +145,24 @@ impl BatchedTextRun {
             origin.y + self.start_point.line as f32 * dimensions.line_height,
         );
 
+        // Check whether this batch contains any Thai characters
+        let contains_thai = self
+            .text
+            .chars()
+            .any(|c| (0x0E01..=0x0E5B).contains(&(c as u32)));
+
         let _ = window
             .text_system()
             .shape_line(
                 self.text.clone().into(),
                 self.font_size.to_pixels(window.rem_size()),
                 std::slice::from_ref(&self.style),
-                Some(dimensions.cell_width),
+                // If Thai text is present, pass None so the shaper can freely position combining vowels and tone marks
+                if contains_thai {
+                    None
+                } else {
+                    Some(dimensions.cell_width)
+                },
             )
             .paint(
                 pos,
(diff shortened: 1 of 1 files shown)


oatsaysai:
@cla-bot check

ConradIrwin:
Thanks for this!

I think we need to use a less Thai-specific approach here; maybe using the unicode-segmentation library to bundle code-points into cells, or the unicode-width crate to not allocate cells for combining chars? Otherwise I'd look at how Ghostty implements this as I know they have spent a lot of time on Unicode rendering.

PR #48719, opened 2026-02-08
Use split token display for xAI models

anilpai:
### Split token display for xAI

Extends the split input/output token display (introduced in #46829 for OpenAI) to all xAI models.

Instead of the combined `48k / 1M` token counter, xAI models now show:
- **↑** input tokens used / input token limit
- **↓** output tokens used / output token limit

#### Before

Screenshot 2026-02-08 at 11 07 13 AM

#### After

Screenshot 2026-02-08 at 11 05 36 AM

#### Changes

- **x_ai.rs** — Override `supports_split_token_display()` to return `true` on `XAiLanguageModel`. All built-in Grok models already implement `max_output_tokens()`, so no additional plumbing was needed.
- **cloud.rs** — Add `XAi` to the `matches!` pattern in `CloudLanguageModel::supports_split_token_display()` so cloud-routed xAI models also get the split display.

#### Tests

- `test_xai_supports_split_token_display` — Verifies all built-in Grok model variants return `true` for split token display.
- `test_xai_models_have_max_output_tokens` — Validates all built-in Grok models report `max_output_tokens` that is `Some`, positive, and less than `max_token_count` (required for the UI to compute the input token limit).
- `test_split_token_display_supported_providers` — Confirms the cloud provider match pattern includes `OpenAi` and `XAi` while excluding `Anthropic` and `Google`.

Release Notes:

- Changed the display of tokens for xAI models to reflect the input/output limits.


crates/language_models/src/provider/cloud.rs
```diff
@@ -1214,6 +1214,29 @@ mod tests {
     use http_client::http::{HeaderMap, StatusCode};
     use language_model::LanguageModelCompletionError;
 
+    #[test]
```
maxdeviant: These tests don't seem like they're really testing anything.

I think we can get rid of them.

PR #48723, opened 2026-02-08
Fix Agent Panel infinite loading when OpenCode ACP startup stalls on Windows

Spandan7724:
Closes #46890

This fixes a Windows bug where starting an OpenCode ACP thread could leave the Agent Panel stuck on “Loading…” indefinitely.

## What changed

- Added a timeout around ACP `initialize` in `crates/agent_servers/src/acp.rs` so startup cannot hang forever.
- On timeout, the spawned agent process is terminated and a `LoadError::Other` is returned.
- Updated `crates/agent_ui/src/acp/thread_view.rs` to map non-`LoadError` connection failures into `LoadError::Other`, so the panel transitions to an error state instead of remaining in loading.

## Why this fixes it

The infinite loading happened because ACP startup did not always reach a terminal state.

- In `crates/agent_servers/src/acp.rs`, Zed waited for ACP `initialize` without a timeout. If the agent handshake stalled, the future never resolved and the Agent Panel stayed in loading indefinitely.
- In `crates/agent_ui/src/acp/thread_view.rs`, non-`LoadError` failures during `agent.connect` were not consistently converted into a load-state error, so some failures did not transition the UI out of loading.

This change fixes both paths:

- Adds a 30s timeout around ACP `initialize`; on timeout, Zed kills the child process and returns `LoadError::Other`.
- Maps non-`LoadError` connect failures to `LoadError::Other` and routes them through `handle_load_error`.

As a result, ACP startup is now deterministic: it either connects successfully or shows an explicit error state, instead of spinning forever.

## Testing

- Built Zed on Windows successfully with `cargo run --release`.
- Verified OpenCode ACP now starts normally and the Agent Panel no longer gets stuck on infinite loading in this environment.

Release Notes:

- Fixed the Agent Panel on Windows getting stuck on loading when starting ACP-based agents (e.g., OpenCode) if initialization stalls or fails.


benbrandt:
The opencode team let us know they fixed the issue on their end. We looked into having a timeout here before, but this can cause issues for some agents depending on people's connection speeds, so I would prefer not to add it again. Thanks for taking a look though!

PR #48727, opened 2026-02-08
gpui: Reset `external_files_dragged` after successful drag-drop on macOS

Albab-Hasan:
after a successful file drag-drop, conclude_drag_operation did not reset external_files_dragged to false. since dragging_exited (the only place that resets this flag) is never called for successful drops (only for cancelled ones), the flag stayed true permanently. this caused synthetic drags (used for text selection during buffer scrolling) to be suppressed for the lifetime of the window.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/platform/mac/window.rs
+++ b/crates/gpui/src/platform/mac/window.rs
@@ -2454,11 +2454,9 @@ extern "C" fn dragging_entered(this: &Object, _: Sel, dragging_info: id) -> NSDr
     let window_state = unsafe { get_window_state(this) };
     let position = drag_event_position(&window_state, dragging_info);
     let paths = external_paths_from_event(dragging_info);
-    if let Some(event) =
-        paths.map(|paths| PlatformInput::FileDrop(FileDropEvent::Entered { position, paths }))
-        && send_new_event(&window_state, event)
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Nice digging, thanks for this!

Whilst looking into this, I noticed we do some redundant stuff here and can centralize the logic here a bit, hence I quickly hopped on to add that refactor. Didn't want to annoy you with that.

That said, the fix itself looks very nice! Thank you for the PR and congratulations to your first contribution! 🎉

Albab-Hasan:
@MrSubidubi also i would appreciate it if you could check out this pr if u have free time, its a minimal code cleanup. (im not sure about this organizations policy about one liners, feel free to ignore if thats an issue)

https://github.com/zed-industries/zed/pull/48317

MrSubidubi:
> if there are any UI or AI related tasks (or honestly anything), i'd be happy to take them on. thanks again!

Appreciated! Feel free to search through our issues, we try labelling these to be suitable for external contributors. You could probaby look into either UI labeled issues or good first issues there and take whatever you like.

> im not sure about this organizations policy about one liners
We are generally fine with these as long as they provide some value, however, context often helps. I already saw that PR but couldn't come to an easy decision there because of that. Will get back there

PR #48733, opened 2026-02-08
Add generic auto_preview setting for Markdown and SVG

### LINKED ISSUE #10966: Configure markdown files to open in preview mode by default
### Check for existing issues

- [X] Completed

### Describe the feature

I know you can open a file at a specific line with the `zed` cli with:

```
zed README.md:111
```

It would be great if you could open a markdown file could be opened (by default) in "preview" mode instead of the editable document? In VSCode this can be accomplished with `workspace.editorAssociations`:

```json
  "workbench.editorAssociations": {
    "*.ipynb": "jupyter-notebook",
    "*.md": "vscode.markdown.preview.editor"
  },
```

_No response_

matanshavit:
Closes #10966

## Summary

Adds a generic `auto_preview` setting that automatically opens a preview when opening previewable files (Markdown and SVG). The setting accepts a simple boolean or an object form for additional configuration.

## Approach

Completely overhauled based on review feedback to use a generic, extensible approach instead of the previous markdown-specific `ProjectItem` pattern.

**How it works:** Each preview crate (`markdown_preview`, `svg_preview`) subscribes to `workspace::Event::ItemAdded`. When a new editor is added to a pane, the crate checks whether the file is previewable using its existing detection function (`is_markdown_file`, `is_svg_file`). If the `auto_preview` setting is enabled, the corresponding `OpenPreview` or `OpenPreviewToTheSide` action is dispatched — reusing the existing action handlers that already handle deduplication.

**Setting format** (hybrid, follows `ThemeSelection` pattern):
```jsonc
// Simple boolean — opens preview in the same pane
"auto_preview": true

// Object form — opens preview in an adjacent pane
"auto_preview": {"mode": "preview_to_side"}

// Disabled (default)
"auto_preview": false
```

**Key behaviors:**
- Files always open as Editor first, then preview is dispatched
- Works for both Markdown and SVG files
- Extensible to future preview types (each preview crate adds its own subscription)
- No duplicate previews — existing action handlers already check for this
- Tab switching to an already-open file does not re-trigger preview (`ItemAdded` only fires for new tabs)
- Defaults to `false` to preserve existing behavior

## Changes from previous revision

- Replaced `auto_preview_markdown` (bool, markdown-specific) with `auto_preview` (generic, hybrid bool/object)
- Removed `MarkdownItem`, `ProjectItem` impl, and all related code — no longer intercepts file opening
- Added SVG auto-preview support using the same pattern
- Uses `workspace::Event::ItemAdded` subscription instead of `ProjectItem` registration

Release Notes:

- Added `auto_preview` setting to automatically open a preview when opening Markdown or SVG files


mvanhorn:
Just want to voice strong support for this. I use Zed daily and work with markdown files constantly - Claude Code generates .md plans, notes, and drafts that I need to read, not edit. Right now every .md file opens as raw source and I have to Cmd+Shift+M every time.

This PR's approach (ImageViewer pattern, ~160 lines, `auto_preview_markdown` setting) looks clean and well-scoped. The three-state idea from #10966 (`code`, `preview`, `preview_to_side`) would be ideal long-term, but starting with preview-only default-off is the right call.

Happy to help test, review, or iterate on this if it would help move it forward. The issue has 33 thumbs-up and has been open for nearly 2 years - there's clear demand.

crates/settings/src/vscode_import.rs
```diff
@@ -986,6 +873,7 @@ impl VsCodeSettings {
                 }
             }),
             zoomed_padding: None,
+            auto_preview_markdown: None,
```
SomeoneToIgnore: There's definitely something related to auto preview in VSCode, which we'd better import?

Image

SomeoneToIgnore (changes requested):
Thank you for the idea.

There are 2 large things I'd love to see changed:

* this new functionality should not be bound to `markdown` or something else.
Instead, it should be the generic "preview", similar to what VSCode example posted shows — this way we'll be able to extend the functionality later to e.g. `latex` preview

* speaking of extending, hardcoding a new, special handler for markdown files does not seem too compatible with whatever future extensions API or other automatic previews?

We have a more or less generic code for previews in https://github.com/zed-industries/zed/blob/809e70116390801b878b031e3df18e24fd43f263/crates/zed/src/zed/quick_action_bar/preview.rs#L55-L57 which is not reused here for some reason which is strange.

Seems wrong to have 2 totally separate tracks of logic to determine this is a markdown file, and, later, for svg case we would be forced to do a similar thing.

Can we extract the `PreviewType`-based logic and use it also somewhere near the item opening, to fire the corresponding action to open the preview for us?

The code around

https://github.com/zed-industries/zed/blob/809e70116390801b878b031e3df18e24fd43f263/crates/workspace/src/workspace.rs#L822-L834

is generic and returns the closure to open the item.
Somewhere where it's used, presumably all usages of `fn load_path` https://github.com/zed-industries/zed/blob/809e70116390801b878b031e3df18e24fd43f263/crates/workspace/src/workspace.rs#L4513-L4522 should be changed somehow.

* last but not least, is seems that this implementation only considers the case when the file is open for the first time (that's what `for_project_item` is called for), but should we also auto preview markdown when we open the file for the first time, close the preview, then switch to another markdown/svg/etc. file?

Interesting to know what other editors do.

assets/settings/default.json
```diff
@@ -230,6 +230,16 @@
     "enabled": false,
     "debounce_ms": 250,
   },
+  // Whether to automatically open a preview when opening previewable
+  // files (e.g., Markdown, SVG).
+  //
+  // Accepts a boolean or an object:
+  //   false - disabled (default)
+  //   true - open preview in the same pane
+  //   {"mode": "preview_to_side"} - open preview in an adjacent pane
```
SomeoneToIgnore: This mixture of objects and primitives is quite bad, as gives zero intel when editing the settings:

Image

We have quite many examples around the settings that show how to use an `enum` for all 3 states defined.

Can we have some 

```rs
/// Add a comment here
enum AutoPreviewMode {
    /// And here!
    Disabled,
    /// Same here
    SamePane,
    /// And here!
    ToSide,
}
```

or similar (whatever naming I quickly came up with is somewhat wacky and could be improved).

Having comments on this struct in crucial, as this is what will be shown as additional detail during settings completion.

assets/settings/default.json
```diff
@@ -230,6 +230,16 @@
...
+  //
+  // Default: false
+  "auto_preview": false,
```
SomeoneToIgnore: Besides the settings file, we need two more places adjusted:

* docs/src/reference/all-settings.md so the users can see the information in https://zed.dev/docs/

* settings_ui crate so people can avoid the json file and configure things via the UI:

Image

crates/markdown_preview/src/markdown_preview.rs
```diff
@@ -37,6 +39,35 @@ pub fn init(cx: &mut App) {
             return;
         };
         markdown_preview_view::MarkdownPreviewView::register(workspace, window, cx);
+
+        let workspace_handle = cx.entity();
+        cx.subscribe_in(
```
SomeoneToIgnore, line 44: Can that be `cx.subscribe_self`?
If we can, `svg_preview` is worth changing too.

SomeoneToIgnore (commented):
Nice!

This looks so much simpler than before, I have left a few comments that I hope are the last ones (given that you test both settings changes after implementing) before we merge this.

crates/settings/src/vscode_import.rs
SomeoneToIgnore: Curios still: `markdown.extension.preview.autoShowPreviewToSide` seems like a good setting to auto import (sure, we'll apply the same to svg previews that but that seems fine), are there any issues with this?

matanshavit: Seems good to me in general. I want to point out that the setting is from the "Markdown All in One" extension, not the core VSCode. Is that ok? I looked at `vscode_import.rs` and I think there aren't any extension settings imported yet, so I wanted to check, but I think it could definitely make sense to start with this setting. Users may expect it to work and I don't think it hurts to have it.
The reason to leave it out would be if Zed intentionally only imports settings from VSCode and never from extensions. This seems to just be what has happened so far, and not a strict decision, but please correct me if that is wrong.

crates/markdown_preview/src/markdown_preview.rs
matanshavit: `subscribe_self` doesn't pass a `Window` to its callback. We need `window.dispatch_action` for am `OpenPreview` action here.

Do we want to create a new `subscribe_self_in` counterpart that also passes the window to the callback? That increases the scope of this PR a bit, but Zed can also start using it other places.

precedent for `subscribe_in` in `initialize_workspace`: [crates/zed/src/zed.rs:449](https://github.com/zed-industries/zed/blob/ec9be5c332f79986b046a2985e003766591ec82f/crates/zed/src/zed.rs#L449)

SomeoneToIgnore: Oh, never mind then, I guess, if we need `window`, thank you for checking.

crates/settings/src/vscode_import.rs
SomeoneToIgnore: Oh, that's why it is so confusing.
If we can look for that setting by id and detect it, seems worth a try.

assets/settings/default.json
matanshavit: Thanks! I updated 
- `docs/src/reference/all-settings.md`
- `crates/settings_ui/page_data.rs`
- `crates/settings_ui/settings_ui.rs`

assets/settings/default.json
matanshavit: I see having the comments is important. I updated to the suggestion.

matanshavit:
@SomeoneToIgnore Great, I think that is everything. Please let me know if the updates cover the needed changes. Thanks!

crates/settings_content/src/workspace.rs
```diff
@@ -130,6 +130,38 @@ pub struct WorkspaceSettingsContent {
...
+    Eq,
+    strum::VariantArray,
+    strum::VariantNames,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum AutoPreviewSetting {
+    /// Do not automatically open a preview.
+    #[default]
+    Disabled,
+    /// Open the preview in the same pane, replacing the editor tab.
```
SomeoneToIgnore, line 161: > replacing the editor tab

Is it really the case?
Currently, even if I open a "preview tab" that gets closed by other tabs opened, the `OpenPreview` invocation always opening the preview tab next, without replacing.

There's another place that needs the same adjustment.

SomeoneToIgnore (changes requested):
Sorry, but seems like nothing works at all in the branch when I test it?

https://github.com/user-attachments/assets/adf29b3f-7329-4d3e-8812-d1cf02bb7275

From what my quick `dbg!`s had shown, `window.dispatch_action` is reached but seems that the action is not getting anywhere.

I was about to test the scenarios and merge the PR, but seems like we'd better have a test for this to be sure it does not regress, would you be able to do this?

Also, I've just discovered we have a csv preview added over a month ago!
https://github.com/zed-industries/zed/blob/9fec902c8ae48353cc3c57f2c380c6c8c4430136/crates/csv_preview/src/csv_preview.rs#L223

Would be an overlook to miss it, let's add that too.

SomeoneToIgnore:
Given first conflicts, no work happening in a month and no meaningfully working change here yet, closing as stale.

image

PR #48735, opened 2026-02-08
docs: Suggest alternative to Vim tag navigation feature

sourcefrog:
Closes N/A

- [ ] Tests or screenshots needed? -- no
- [x] Code Reviewed 
- [x] Manual QA -- I ran `mdbook serve` and checked the output looks OK

Release Notes:

- N/A

### DIFF
--- a/docs/src/vim.md
+++ b/docs/src/vim.md
@@ -539,6 +539,7 @@ If you're using vim mode on Linux or Windows, you may find it overrides keybindi
     "ctrl-f": "buffer_search::Deploy",      // vim default: page down
     "ctrl-o": "workspace::Open",            // vim default: go back
     "ctrl-s": "workspace::Save",            // vim default: show signature
+    "ctrl-t": "project_symbols::Toggle",    // vim default: go to older tag
     "ctrl-a": "editor::SelectAll",          // vim default: increment
     "ctrl-b": "workspace::ToggleLeftDock"   // vim default: down
   }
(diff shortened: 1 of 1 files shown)


docs/src/vim.md
```diff
@@ -539,6 +539,7 @@ If you're using vim mode on Linux or Windows, you may find it overrides keybindi
     "ctrl-f": "buffer_search::Deploy",      // vim default: page down
     "ctrl-o": "workspace::Open",            // vim default: go back
     "ctrl-s": "workspace::Save",            // vim default: show signature
+    "ctrl-t": "project_symbols::Toggle",    // vim default: go to older tag
```
yeskunall, line 541: Thanks, but shouldn’t this be [`pane::GoToOlderTag`](https://github.com/zed-industries/zed/blob/70feabf659466607628f93e9348c6e251841df3c/assets/keymaps/vim.json#L70)? Your comment suggests that. Or is that exactly what you are trying to override here?

sourcefrog: Yes, the comment shows the default vim binding. The uncommented active code restores it to what it normally does in Zed and VS Code.

sourcefrog: Yes, the comment shows the default vim binding. The uncommented active code restores it to what it normally does in Zed and VS Code.

yeskunall: Ah, got it -- your change makes much more sense now. Thanks for elaborating. I need to update the PR title, then.

PR #48738, opened 2026-02-08
Allow specifying different fonts for inlays

### LINKED ISSUE #15716: Custom font family for inlay hints
### Check for existing issues

- [X] Completed

### Describe the feature

You should be able to set a custom font family for inlay hints.

This should probably include setting font features, weight, and size.

_No response_

garthdw:
Rework `is_inlay` on chunks to specify the kind of chunk instead as an enum.

Mark all existing inlay hint kinds as InlayHint except for edit predictions which are marked separately.

This allows us to handle rendering differently at the element level, currently this is used to adjust the font.

Closes #15716

- [x] Code Reviewed
- [x] Manual QA

Release Notes:

- Add support for specifying different fonts for inlay hints and edit predictions

https://github.com/user-attachments/assets/af0bb920-6d4a-4f56-9142-d6a03900e530

image

Fonts in use:
- Monaspace Neon (buffer font)
- Monaspace Krypton (edit prediction font)
- OpenDyslexic Nerd Font Propo for inlay hints

Please note I have only dabbled in rust, so could be holding it wrong.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -566,7 +566,9 @@ pub struct EditorStyle {
     pub syntax: Arc<SyntaxTheme>,
     pub status: StatusColors,
     pub inlay_hints_style: HighlightStyle,
+    pub inlay_hints_font: Option<SharedString>,
     pub edit_prediction_styles: EditPredictionStyles,
+    pub edit_prediction_font: Option<SharedString>,
     pub unnecessary_code_fade: f32,
     pub show_underlines: bool,
 }
@@ -585,10 +587,12 @@ impl Default for EditorStyle {
             // style and retrieve them directly from the theme.
             status: StatusColors::dark(),
             inlay_hints_style: HighlightStyle::default(),
+            inlay_hints_font: None,
             edit_prediction_styles: EditPredictionStyles {
                 insertion: HighlightStyle::default(),
                 whitespace: HighlightStyle::default(),
             },
+            edit_prediction_font: None,
             unnecessary_code_fade: Default::default(),
             show_underlines: true,
         }
@@ -25776,7 +25780,9 @@ impl Editor {
             syntax: cx.theme().syntax().clone(),
             status: cx.theme().status().clone(),
             inlay_hints_style: make_inlay_hints_style(cx),
+            inlay_hints_font: Some("Monaspace Krypton".into()),
             edit_prediction_styles: make_suggestion_styles(cx),
+            edit_prediction_font: Some("Monaspace Krypton".into()),
             unnecessary_code_fade: settings.unnecessary_code_fade,
             show_underlines: self.diagnostics_enabled(),
         }
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -566,9 +566,9 @@ pub struct EditorStyle {
     pub syntax: Arc<SyntaxTheme>,
     pub status: StatusColors,
     pub inlay_hints_style: HighlightStyle,
-    pub inlay_hints_font: Option<SharedString>,
+    pub inlay_hints_font: gpui::Font,
     pub edit_prediction_styles: EditPredictionStyles,
-    pub edit_prediction_font: Option<SharedString>,
+    pub edit_predictions_font: gpui::Font,
     pub unnecessary_code_fade: f32,
     pub show_underlines: bool,
 }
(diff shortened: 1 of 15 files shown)


kelvincesar:
It is possible to change the font size also?

crates/theme/src/settings.rs
```diff
@@ -135,6 +135,10 @@ pub struct ThemeSettings {
     pub ui_density: UiDensity,
     /// The amount of fading applied to unnecessary code.
     pub unnecessary_code_fade: f32,
+    /// The font used for inlay hints.
+    pub inlay_hints_font: Font,
```
SomeoneToIgnore, line 139: 1. It's very hard for me to understand this particular separation (same for `ChunkKind`), as, technically, Zed has more than 2 kinds of inlays: https://github.com/zed-industries/zed/blob/44015e0379b6c868258318d9c159134adbf801d9/crates/editor/src/inlays.rs#L6-L12

Would not it be enough to have a uniform, single, set of knobs to configure all inlays at once?

2. In the end, before merging the PR but after we establish the settings scope, we should have an entry in default.json + all-settings.md files.

crates/editor/src/editor.rs
```diff
@@ -25347,7 +25351,9 @@ impl Editor {
             syntax: cx.theme().syntax().clone(),
             status: cx.theme().status().clone(),
             inlay_hints_style: make_inlay_hints_style(cx),
+            inlay_hints_font: settings.inlay_hints_font.clone(),
```
SomeoneToIgnore, line 25783: Seems that there's no tracking of these settings changed?
We would need to invalidate editor's inlay map's data and re-render the hints if the font-related settings change.

SomeoneToIgnore (commented):
1. I would agree with the comment here, as https://github.com/zed-industries/zed/issues/15716 clearly states 

>  This should probably include setting font features, weight, and size.

I think it's fair to get them all on board in the PR at once — I think it's enough of `inlay_font_something` set of settings added?
Or we could move the away into a separate struct in the editor settings?

2. I do not think we should have any fine-grained settings for various kinds of inlays: it's all or nothing at this point

SomeoneToIgnore:
Then, we can scope down to https://github.com/zed-industries/zed/blob/44015e0379b6c868258318d9c159134adbf801d9/crates/settings_content/src/language.rs#L637 and add the settings here — seems viable to me too, if we ignore the edit predictions' part this way too.

crates/editor/src/editor.rs
garthdw: With the current approach it _just works™_ I assume because it's done in the same place as theming changes.


https://github.com/user-attachments/assets/3035aa2c-5307-4bbf-9a00-381cfa5ac0f4

crates/theme/src/settings.rs
garthdw: My goal was to be able to mimic the following settings from VSCode and the Monaspace font (https://monaspace.githubnext.com/#:~:text=What%20if)

image

`ChunkKind` was a way for us to be able to act on some additional concern about the chunk at the "top" render level.

Currently the only toggles in place are, edit predictions, (all other inlay hints), not an inlay hint.

Since we already allow styling edit predictions vs "inlay hints", I feel that allow different fonts at that level is a natural extension of that separation.

Currently I made it default all back to the buffer font but we could also make it default back to inlay, which would default back to buffer.


We set all other inlay types to just [inlay hint here](https://github.com/zed-industries/zed/pull/48738/changes#diff-0d4f1ed762a5f976a8d9797148c06990c516d44fd9c011ad3bde56bd92321b08R328), which should be "easy" to extend `ChunkKind` with the other types if there's a need for identifying more directly in the future. I do think `kind` is probably not the right choice of word, maybe `inlay_type` might be more self documenting choice.

SomeoneToIgnore: Ok, but then — do you expect the color and debug inlays to have a different font when you set the inlay font size to something else?
Because now the code will make them so, if I'm not mistaken?

Great point on the settings — we need to have a proper `vscode_import.rs` adjustments to get their related settings migrated into Zed's.

garthdw: In this case I'd expect any inlay that hasn't been "specialised" (e.g edit predictions) to "inherit" the inlay hints font options. So as the code has been written.

Basically buffer > inlay hints > x

Where only edit predictions have x.

I will expand the font options to include the other options.

I will also update the migrate correctly.

I did originally try to have the font family nested under inlay hints settings (ideally then edit predictions font family can be under that section) but ran into some clone/copy stuff that given my rust inexperience I decided to opt for a simpler solution.

garthdw: I finally made time to finish up this.

I know the all-settings.md still needs work to expand on what the settings mean.

I have included font-weight and font features for inlay and edit predictions in particular.

wesguirra:
please do font-size too

PR #48750, opened 2026-02-09
node_runtime: Support Bun as a drop-in replacement for Node/npm

### LINKED ISSUE #32042: Zed errors with node-compatible runtimes such as bun
### Summary

Zed assumes Node.js runtime and raises errors when running on Node.js-compatible runtimes, such as bun.

### Description

Steps to reproduce:
1. 	// set bun instead of node
        "node": {
		"path": "/opt/homebrew/bin/bun",
		"npm_path": "/opt/homebrew/bin/bun"
	},
2. open any project that triggers language servers

Expected Behavior: Zed should work with node-compatible runtimes similarly as e.g. vscode does.
Actual Behavior: Zed checks for node-compatible runtime and fails if version doesn't match

error examples:
```
Language server error: json-language-server

node at /opt/homebrew/bin/bun is too old. want: 20.0.0, got: 1.2.15
-- stderr--
```
```
Language server error: biome

node at /opt/homebrew/bin/bun is too old. want: 20.0.0, got: 1.2.15
-- stderr--
```

Zed 0.188.6 – /Applications/Zed.app

agene0001:
## Summary

Adds support for using [Bun](https://bun.sh) as a drop-in replacement for Node.js and npm in Zed. When a user configures their `node.path` and `npm_path` settings to point to a Bun binary, Zed now automatically detects it and translates package manager commands accordingly.

Closes #32042

## Problem

When a user sets `"node": {"path": "/opt/homebrew/bin/bun"}` in their Zed settings, Zed:
1. Runs `bun --version` and gets something like `1.2.15`
2. Compares this against `MIN_VERSION = 22.0.0` (intended for Node.js)
3. Rejects it with "node at ... is too old"

Even if the version check were bypassed, npm-specific CLI flags (`--cache`, `--userconfig`, `--globalconfig`, `--fetch-retry-mintimeout`, etc.) would be passed to bun where they have no effect or cause errors.

## Implementation

All changes are contained within the `node_runtime` crate — **zero changes** to the 70+ callers across language servers, prettier, copilot, extensions, etc.

### Detection
- Added a `PackageManager` enum (`Npm`/`Bun`) to `SystemNodeRuntime`
- Bun is detected by checking if the binary filename is `bun` or `bunx`
- When Bun is detected, the Node.js `MIN_VERSION` check is skipped
- The `npm root -g` initialization call is also skipped for Bun

### Command Translation (`build_bun_command_args`)
| npm command | bun equivalent |
|---|---|
| `npm install pkg@ver --save-exact` | `bun add --exact pkg@ver` |
| `npm install` (bare) | `bun install` |
| `npm run-script compile` | `bun run compile` |
| `npm exec` | `bun x` |
| `--prefix <dir>` | `--cwd <dir>` |

npm-specific flags that bun doesn't understand are stripped:
`--cache`, `--userconfig`, `--globalconfig`, `--fetch-retry-mintimeout`, `--fetch-retry-maxtimeout`, `--fetch-timeout`

### Package Version Queries
Since bun doesn't support `npm info <pkg> --json`, when Bun is detected, `npm_package_latest_version()` queries the npm registry HTTP API directly (`https://registry.npmjs.org/<pkg>`) using Zed's existing HTTP client.

### Settings Documentation
Updated `default.json` and `NodeBinarySettings` doc comments to mention Bun support.

## Usage

```json
{
  "node": {
    "path": "/opt/homebrew/bin/bun",
    "npm_path": "/opt/homebrew/bin/bun"
  }
}
```

## Testing

- `cargo test -p node_runtime` — all existing tests pass
- Manual verification: language servers install and start correctly with Bun configured
- No regression for existing Node.js users — all Bun-specific code paths are behind detection

## Release Notes:

- Added support for using Bun as a drop-in replacement for Node.js/npm (#32042). Set `node.path` and `npm_path` in settings to your Bun binary path.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1856,8 +1856,9 @@
     "Markdown": [".rules", ".cursorrules", ".windsurfrules", ".clinerules"],
     "Shell Script": [".env.*"],
   },
-  // Settings for which version of Node.js and NPM to use when installing
-  // language servers and Copilot.
+  // Settings for which version of Node.js/Bun and NPM to use when installing
+  // language servers and Copilot. Bun can be used as a drop-in replacement
+  // by setting `path` and `npm_path` to your Bun binary.
   //
   // Note: changing this setting currently requires restarting Zed.
   "node": {
@@ -1866,9 +1867,10 @@
     // to prevent this, and force Zed to always download and install its own
     // version of Node.
     "ignore_system_version": false,
-    // You can also specify alternative paths to Node and NPM. If you specify
-    // `path`, but not `npm_path`, Zed will assume that `npm` is located at
-    // `${path}/../npm`.
+    // You can also specify alternative paths to Node (or Bun) and NPM.
+    // If you specify `path`, but not `npm_path`, Zed will assume that `npm`
+    // is located at `${path}/../npm`. When using Bun, set both `path` and
+    // `npm_path` to your Bun binary path.
     "path": null,
     "npm_path": null,
   },
--- a/crates/node_runtime/src/node_runtime.rs
+++ b/crates/node_runtime/src/node_runtime.rs
@@ -1050,16 +1050,18 @@ fn build_bun_command_args(_proxy: Option<&Url>, subcommand: &str, args: &[&str])
             continue;
         }
 
-        // Skip npm-specific flags that bun doesn't understand
-        if arg.starts_with("--cache=")
-            || arg.starts_with("--fetch-retry-")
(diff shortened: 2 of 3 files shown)


osiewicz (commented):
What does that get us, exactly? I'm not sure if we need to dabble with supporting different runtimes, when all we care about is that we can install packages via it in a reliable way. Obviously that may leave a bit of performance on the table, but I still feel that we should err on the side of compatibility.

PR #48752, opened 2026-02-09
editor: Add action to toggle block comments

### LINKED ISSUE #4751: Block comment command
### Check for existing issues

- [X] Completed

### Describe the feature

similar to `editor: Toggle Comment` there should be `editor: Toggle Block Comment`. In python this creates `""" hello """`, or `/* */` in JavaScript etc.

Related: https://github.com/zed-industries/extensions/issues/522

ozacod:
Closes #4751

## Testing
- Manually tested by comparing the behaviors with vscode.
- Those requirements are added to unit tests.

Release Notes:

- Added action to toggle block comments


niekdomi:
I tested this PR and it works quite well 😄.

One thing I realized is, that it may be more intuitive to exit the visual mode (quit selection) after running the block-comment command

Also, do you plan to add keybindings in this PR? E.g., in NeoVim, [this](https://github.com/numToStr/Comment.nvim) one is likely the most popular plugin for block-comments which uses `gb..` motion to toggle block comments

ozacod:
@niekdomi Thanks a lot, fixed it for gb$, gbc and gbe.

ozacod:
> There's another bug (I think this wasn't an issue before). running `gbj` should comment out the current and next line. Though, if the next line is a newline/empty line, zed crashes.
> 
> This seems also to happen with `gbk` which comments out the current line and the one above, though, here zed crashes, if the current line (where the cursor is positioned) is a new line.

Fixed it. I definitely should have run more comprehensive stress tests. Until then, I am switching this PR to draft. I really appreciate your help. Thanks,

ozacod:
Tested following keybindings in vim mode. "gbc, 5gbc, gbe, gbw, gbj, gbk, gb5j, gb$, gb0, gbG, gbgg, gb%", some edge cases too. 
Ready for review.

crates/vim/src/normal.rs
```diff
@@ -976,6 +985,26 @@ impl Vim {
...
+        self.record_current_action(cx);
+        self.store_visual_marks(window, cx);
+        self.update_editor(cx, |vim, editor, cx| {
+            editor.transact(window, cx, |editor, window, cx| {
+                let original_positions = vim.save_selection_starts(editor, cx);
+                editor.toggle_block_comments(&Default::default(), window, cx);
+                vim.restore_selection_cursors(editor, window, cx, original_positions);
+            });
+        });
+        if self.mode.is_visual() {
```
SomeoneToIgnore, line 1024: Zed has nvim-backed tests, can we have one for this new case?

crates/languages/src/rust/config.toml
```diff
@@ -16,6 +16,7 @@ brackets = [
 ]
 collapsed_placeholder = " /* ... */ "
 debuggers = ["CodeLLDB", "GDB"]
+block_comment = { start = "/*", end = "*/", prefix = "", tab_size = 0 }
```
SomeoneToIgnore: Curious why here and other similarly changed cases the numbers are different from the ones in `documentation_comment`?

assets/keymaps/default-macos.json
```diff
@@ -551,6 +551,7 @@
       "cmd-k cmd-i": "editor::Hover",
       "cmd-k cmd-b": "editor::BlameHover",
       "cmd-/": ["editor::ToggleComments", { "advance_downwards": false }],
+      "shift-alt-a": "editor::ToggleBlockComments",
```
SomeoneToIgnore, line 586: I know this is how VSCode has this, but in our case seems that we should use something else as alt-letter would be expected to type letters in macOS?

cc @ConradIrwin for this and the Vim-related part to check out.

crates/editor/src/editor.rs
```diff
@@ -16001,6 +16003,236 @@ impl Editor {
...
+                    end_point.row,
+                    snapshot.line_len(MultiBufferRow(end_point.row)),
+                );
+                let region_bytes: Vec<u8> = snapshot
+                    .bytes_in_range(region_start..region_end)
+                    .flatten()
+                    .copied()
+                    .collect();
+
+                let to_byte = |target: Point| -> usize {
```
SomeoneToIgnore: We have the `snapshot` and can call `to_offset` , similarly the reverse is possible + it should be line-ending aware (current verison seems to disregard Windows?).

What do I miss and why is this all needed?

crates/editor/src/editor.rs
```diff
@@ -16001,6 +16003,236 @@ impl Editor {
...
+
+                if is_commented {
+                    // Also remove the space after /* and before */
+                    if snapshot
+                        .bytes_in_range(prefix_range.end..snapshot.max_point())
+                        .flatten()
+                        .next()
+                        == Some(&b' ')
+                    {
+                        prefix_range.end.column += 1;
```
SomeoneToIgnore, line 16526: This might panic for multi-byte characters e.g. `√√√` and might need clipping (the snapshot has the method).
Let's add tests that cover all those branches and ensure we do not panic.

Same applies to the rest of the arithmetic below, including -1 and the selection-related column manipulations.

crates/editor/src/editor.rs
```diff
@@ -16001,6 +16003,236 @@ impl Editor {
...
+                // on an empty line is the actual end-of-line, not a trailing newline.
+                if end_point.column == 0
+                    && end_point.row > start_point.row
+                    && snapshot.line_len(MultiBufferRow(end_point.row)) > 0
+                {
+                    end_point.row -= 1;
+                    end_point.column = snapshot.line_len(MultiBufferRow(end_point.row));
+                }
+
+                let language = if let Some(language) =
```
SomeoneToIgnore: NIT: looks that both `continue` do not depend on `end_point`, so we could move them both above the `end_point` adjustment.

crates/editor/src/editor.rs
```diff
@@ -16001,6 +16003,236 @@ impl Editor {
...
+            let mut edits = Vec::new();
+            let snapshot = this.buffer.read(cx).read(cx);
+            let empty_str: Arc<str> = Arc::default();
+            let mut markers_inserted = Vec::new();
+
+            for selection in &mut selections {
+                let start_point = selection.start;
+                let mut end_point = selection.end;
+
+                // Visual line mode: move end back to end of previous line.
```
SomeoneToIgnore: Is this Vim-related?
Curious why do we do that always, not hiding under a condition of vim being enabled.

SomeoneToIgnore (changes requested):
Thank you, seems that we need to test it a bit more, I've also left a few more code-related comments.

crates/languages/src/rust/config.toml
ozacod: fixed.

crates/editor/src/editor.rs
ozacod: fixed.

crates/editor/src/editor.rs
ozacod: fixed and added tests.

crates/editor/src/editor.rs
ozacod: fixed.

crates/editor/src/editor.rs
ozacod: it was deadcode, so removed it.

crates/vim/src/normal.rs
ozacod: I have added related tests to vim crate. Block comment support is available via plugin for Neovim, and CI runs neovim with --clean flag.

ozacod:
Hi @ConradIrwin,
@SomeoneToIgnore pointed this out in code review. What should be our action here?

> I know this is how VSCode has this, but in our case seems that we should use something else as alt-letter would be expected to type letters in macOS?
> 
> cc @ConradIrwin for this and the Vim-related part to check out.

crates/languages/src/rust/config.toml
SomeoneToIgnore: Not really, some of the properties are different still?
And, again, curious why is it different.

ozacod: I mainly write C++ and it is a convention to use * as a prefix for doc comments, not for block comments. Is not it the same in Rust?

SomeoneToIgnore: I believe it's not settled this strict in Rust and there's been some discussions on the necessity of having both forms supported.

At this stage though, nothing is improved there and for now it seems the safest bet is to keep things the same in both cases?

SomeoneToIgnore (approved):
Checked with Conrad and seems we're good to move on, thank you for fixing this long-awaited issue!

PR #48773, opened 2026-02-09
search: Fix collapse/expand all button sync

### LINKED ISSUE #48734: Collapse/Expand All button state desyncs when toggling individual files
### Reproduction steps

  1. Use "Find All References" on a symbol that has references across multiple files
  2. Note the "Collapse All" / "Expand All" button
  3. Manually collapse or expand individual file excerpts by clicking their chevron headers

Current behavior: Clicking "Collapse All" expands all excerpts instead of collapsing them. The label says one thing but the behavior does the opposite.

  Expected behavior: The label should stay consistent with its behavior.

Zed: v0.222.4+stable.147.b385025df963c9e8c3f74cc4dadb1c4b29b3c6f0 (Zed) 
OS: macOS 26.1
Memory: 48 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

bnjjj:
Closes #48734 

- [X] Tests or screenshots needed?
- [X] Code Reviewed
- [X] Manual QA

Release Notes:

- Fixed collapse/expand all button in buffer search and project search not syncing correctly when toggling individual file sections


crates/editor/src/editor.rs
```diff
@@ -20321,6 +20321,17 @@ impl Editor {
...
 
+    pub fn has_any_buffer_folded(&self, cx: &App) -> bool {
+        if self.buffer().read(cx).is_singleton() {
+            return false;
+        }
+        self.buffer()
+            .read(cx)
+            .excerpt_buffer_ids()
+            .iter()
+            .any(|buffer_id| self.is_buffer_folded(*buffer_id, cx))
```
dinocosta: I believe we can leverage `Editor.folded_buffers` here and simply check if that's empty:


```suggestion
        !self.folded_buffers(cx).is_empty()
```

That way we avoid iterating over all buffer ids and allocating a new vector 👍

dinocosta (changes requested):
Thank you for tackling this @bnjjj ! 🙂

Changes are looking good but I suspect that, if we're updating this, we can probably also remove `BufferSearchBar.is_collapsed` and `ProjectSearchView.results_collapsed` . I also noticed that we still have a bug in case the button is used when focused on the search query editor, as that ends up calling the `BufferSearchBar.toggle_fold_all_in_item` method which still relies on `BufferSearchBar.is_collapsed` ▼

https://github.com/user-attachments/assets/578bb16e-ae4f-4705-8200-88428ec95215

Notice how, even though the button says, "Expand All Files", as it was updated, after collapsing a single file, clicking it actually collapses all files .

Let me know if you have any questions. Thanks!

bnjjj:
Thanks for the review @dinocosta I pushed a new commit which fix the project search bar. I tested manually and it worked properly. Let me know if you're able to reproduce your bug.

dinocosta (changes requested):
@bnjjj Thank you for tackling that change to `BufferSearchBar.toggle_fold_all_in_item` ✌️

Since we're now relying on the `Editor.has_any_buffer_folded` method as the source of truth for which icon and tooltip to show, I believe we can still remove both the `BufferSearchBar.is_collapsed` and `ProjectSearchView.results_collapsed` fields . 

In the case of the `Render` implementation for `BufferSearchBar`, I believe we can either:

* Opt for a simpler approach and just use `.unwrap_or(false)` where we're currently doing `.unwrap_or(self.is_collapsed)`
* Update the `SearchableItemHandle` trait with a `has_results_collapsed` method and avoid downcasting

Let me know what you think. Thanks! 🙂

bnjjj:
thanks @dinocosta ! I think I addressed your comments

dinocosta (approved):
@bnjjj Thank you for tackling the comments! I've also removed the `ProjectSearchView.results_collapsed` field and the `SearchEvent::ResultsCollapsedChanged` event, as it's no longer needed for the UI to re-render with the correct icon and tooltip 🙂

PR #48777, opened 2026-02-09
extension_host: Add DAP methods dispatch for v0.8.0

Brad-Fullwood:
## Summary

When the v0.8.0 extension API was forked in #44025, the five DAP dispatcher methods in `wit.rs` were not updated to handle `Extension::V0_8_0`. Because `V0_8_0` is listed before `V0_6_0` in the enum, the wildcard `_ =>` catch-all fires first, causing all DAP calls to bail with `"not available prior to v0.6.0"` for any extension targeting the v0.8.0 API.

The DAP WIT interface is identical between v0.6.0 and v0.8.0, so the handler code is the same — this just adds the missing match arms for:

- `call_get_dap_binary`
- `call_dap_request_kind`
- `call_dap_config_to_scenario`
- `call_dap_locator_create_scenario`
- `call_run_dap_locator`

This follows the same pattern used by every other method in the `Extension` impl block, which already handles both `V0_8_0` and `V0_6_0`.

## Test plan

- Verified that an extension targeting `zed_extension_api` v0.8.0 with DAP support can successfully start a debug session (previously failed with `"dap_request_kind not available prior to v0.6.0"`)

Release Notes:

- Fixed DAP (Debug Adapter Protocol) methods failing for extensions targeting the v0.8.0 extension API.

### DIFF
--- a/crates/extension_host/src/wasm_host/wit.rs
+++ b/crates/extension_host/src/wasm_host/wit.rs
@@ -1007,6 +1007,20 @@ impl Extension {
         resource: Resource<Arc<dyn WorktreeDelegate>>,
     ) -> Result<Result<DebugAdapterBinary, String>> {
         match self {
+            Extension::V0_8_0(ext) => {
+                let dap_binary = ext
+                    .call_get_dap_binary(
+                        store,
+                        &adapter_name,
+                        &task.try_into()?,
+                        user_installed_path.as_ref().and_then(|p| p.to_str()),
+                        resource,
+                    )
+                    .await?
+                    .map_err(|e| anyhow!("{e:?}"))?;
+
+                Ok(Ok(dap_binary))
+            }
             Extension::V0_6_0(ext) => {
                 let dap_binary = ext
                     .call_get_dap_binary(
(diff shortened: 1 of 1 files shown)


Brad-Fullwood:
@cla-bot check

MrSubidubi (changes requested):
Very very nice catch and thank you for opening the PR here. 

To prevent this from happening in the future, could you please make the matches exhaustive and remove the wildcard patterns everywhere?

Brad-Fullwood:
> Very very nice catch and thank you for opening the PR here.
> 
> To prevent this from happening in the future, could you please make the matches exhaustive and remove the wildcard patterns everywhere?

Hiya @MrSubidubi , no problem at all.
Updated, built and tested with the requested change.

crates/extension_host/src/wasm_host/wit.rs
```diff
@@ -1108,7 +1200,16 @@ impl Extension {
...
-            _ => anyhow::bail!("`dap_locator_create_scenario` not available prior to v0.6.0"),
+            Extension::V0_5_0(_)
+            | Extension::V0_4_0(_)
+            | Extension::V0_3_0(_)
+            | Extension::V0_2_0(_)
+            | Extension::V0_1_0(_)
+            | Extension::V0_0_6(_)
+            | Extension::V0_0_4(_)
+            | Extension::V0_0_1(_) => {
+                anyhow::bail!("`run_dap_locator` not available prior to v0.6.0");
```
MrSubidubi, line 1211: Very nice additional catch 🙂

MrSubidubi (approved):
Awesome, thanks for the quick follow-up! 

Thank you so much for this PR and congratulations to such an awesome first contribution! 🎉

PR #48789, opened 2026-02-09
gpui: Fix services menu passing NSMenuItem instead of NSMenu on macOS

Albab-Hasan:
`setServicesMenu:` expects an `NSMenu` parameter, but was being passed `item` (an `NSMenuItem`) instead of `submenu` (the actual `NSMenu`).

Release Notes:

- N/A


Albab-Hasan:
@SomeoneToIgnore hey, can i get someone assigned to this?

Albab-Hasan:
@ChristopherBiscardi it was more like "the code looked wrong and this might be an issue" type of thing in this pr. Like most of my other merged prs in zed. I just read through the code and fix whatever that seems wrong to be.

I was wrong in this case I guess.

PR #46541, opened 2026-01-11
Git partially staged changes

drbh:
This PR explores the addition of a new feature and UI to improve visibility into partially staged commits.

Currently, the Git panel shows tracked and untracked changes, but it does not clearly distinguish between staged and unstaged changes.

As a result, it’s difficult to quickly see which changes are not staged in the current UI. Both staged and unstaged changes are combined into the `Uncommitted Changes` multibuffer. This developer experience differs from other editors, most notably VS Code; which presents separate Staged Changes and Changes lists.

### Staged and unstaged diffs in multibuffers

This PR introduces an alternative UI for unstaged changes that aligns with the overall Zed experience. Instead of showing changes on a per-file basis, staged and unstaged diffs are each displayed in their own multibuffers, similar to how `Uncommitted Changes` currently works.

For example the following screenshot shows the current `Uncommitted Changes` on the left, the `Staged Changes` in the middle and the `Unstaged Changes` buffer on the right for comparison

### Indicators/interactions

The new multibuffers can be opened in two ways:

1. Via a new `U` chip, which appears when a file has unstaged changes
2. Via new menu options

(See screenshots below for both interaction paths.)

<table>
  <tr>
    <td style="text-align: center; vertical-align: top;">
      <p>via the chip</p>
      Via the chip
    </td>
    <td style="text-align: center; vertical-align: top;">
      <p>via the menu</p>
      Via the menu
    </td>
  </tr>
</table>

### Design goals
- minimally intrusive UI changes (small new badge and menu items)
- adhere by Zed'ism (use multibuffer where possible)
- avoid disabling any current interactions (Uncommitted Changes ui is unchanged)
- avoid introducing an app level view mode (no new settings needed)

### Experience goals
- make it easy to see what changes are not staged
- make it easy to see that a file has unstaged changes (avoid developers accidently leaving out changes in a commit; a personal issue that I have when using Zed)
- elegantly handle large file's unstaged changes (follows the same collapse and expanding seen in `Uncommitted Changes`)

### How to try

- Clone the repo and run `cargo run`
- Make a change to a file and stage it
- Make another change to the file (the `U` indicator will appear)
- Click the `U` to see the unstaged view

### Open questions/rough edges
- [ ] determine if this user experience is useful for others
- [ ] ensure all interactions work as expected (response to all update cases)

In general I'm really interested in hearing the community's feedback about this interface, more than happy to make any changes or explore a different solution!

### Related issue:
- https://github.com/zed-industries/zed/pull/36646
[...]

PR #48792, opened 2026-02-09
Add support for staged and unstaged changes in project diff view

### LINKED ISSUE #36703: The Git panel & Uncomitted Changes view is confusing for files with staged (but since undone) changes.
### Summary

The Git panel is confusing for files with staged (but since undone) changes.

### Description
When you undo changes to a file after you staged it, Zed is deeply confusing in how it surfaces that fact to users of the Git Pane, and shows an **Open Diff** option that does nothing.

I do understand what is happening under the hood here. This is a case of technical correctness still being deeply confusing at first glance due to Zed UX decisions.  The `git status` is actually quite clear: I have staged changes to file X, and I have un-staged changes to file X too. 

How Zed displays it feels like a bug.  I can also trivially `git diff` to see what is not yet staged, but Zed's `Open Diff` does nothing, and is unhelpful.

DESCRIPTION_HERE

Steps to reproduce:
1. Open a tracked file in a git repo
2. Edit it
3. Go to Git Panel
4. View Uncommitted Changes Pane
5. Stage Change in Uncommitted Changes Pane
6. Go back to file (g space)
7. Manually revert change (or just undo) + Save/Write File
8. Switch back to Uncommitted Changes Pane
9. UNEXPECTED > File remains visible in Git Pane, but not in Uncommitted Changes pane.
10. UNEXPECTED > "Open Diff" option doesn't work

This is fixed by staging the file (which should drop it off both lists completely, no changes to it from HEAD).

**Expected Behavior**:
 - Open Diff works (showing diff of current state to staged).
 - Some information is show to help me determine why I can see the file on the left but not the right.

**Actual Behavior**:
File appears in one list and the only clue something is up is an indeterminate checkbox state on it, UI is shown and that does not work.

Image

Zed: v0.200.4 (Zed) 
OS: macOS 15.5.0
Memory: 16 GiB
Architecture: aarch64

bnjjj:
Closes #36703 

I asked several times on #36703 to know what kind of design we would like to have for this feature. I tried to add a non intrusive new select box to select the different diff views we would like to have. cc @cole-miller, @mattermill or @esthertrapadoux if you can give me feedback that would be really nice.

- [X] Code Reviewed
- [X] Manual QA

Screencast:

https://github.com/user-attachments/assets/6c2afbf8-2cab-4a0b-b9eb-2b7261587902

Release Notes:

- Add support of git staged and unstaged changes in project diff view

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -2,7 +2,7 @@ use crate::askpass_modal::AskPassModal;
 use crate::commit_modal::CommitModal;
 use crate::commit_tooltip::CommitTooltip;
 use crate::commit_view::CommitView;
-use crate::project_diff::{self, Diff, ProjectDiff};
+use crate::project_diff::{self, Diff, ProjectDiff, StagedDiff, UnstagedDiff};
 use crate::remote_output::{self, RemoteAction, SuccessMessage};
 use crate::{branch_picker, picker_prompt, render_remote_button};
 use crate::{
@@ -4852,6 +4852,8 @@ impl GitPanel {
                 )
                 .separator()
                 .action("Open Diff", menu::Confirm.boxed_clone())
+                .action("Open Staged Diff", StagedDiff.boxed_clone())
+                .action("Open Unstaged Diff", UnstagedDiff.boxed_clone())
                 .action("Open File", menu::SecondaryConfirm.boxed_clone())
                 .separator()
                 .action_disabled_when(is_created, "View File History", Box::new(git::FileHistory))
--- a/crates/git_ui/src/project_diff.rs
+++ b/crates/git_ui/src/project_diff.rs
@@ -38,7 +38,9 @@ use smol::future::yield_now;
 use std::any::{Any, TypeId};
 use std::sync::Arc;
 use theme::ActiveTheme;
-use ui::{KeyBinding, Tooltip, prelude::*, vertical_divider};
+use ui::{
+    ContextMenu, DropdownMenu, DropdownStyle, KeyBinding, Tooltip, prelude::*, vertical_divider,
+};
 use util::{ResultExt as _, rel_path::RelPath};
 use workspace::{
     CloseActiveItem, ItemNavHistory, SerializableItem, ToolbarItemEvent, ToolbarItemLocation,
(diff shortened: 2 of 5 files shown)


nicojerome:
I need this so badly. Praying this gets merged sooner rather than later.

The select box is a perfectly clean, non-intrusive UI solution, IMO.

bnjjj:
I'm open to update it if needed, I just need feedback from zed team, cc @esthertrapadoux

pastr:
To review agents work this is incredibly useful

sonny-campbell:
Agreeing that this is a really useful feature, but one question/piece of feedback on the design side. In most git tools we see a separation like this for staged vs unstaged, where all the changes and items are visible in a single pane, but logically separated. This makes it really easy to switch between staged/unstaged versions of the code and selectively stage sections for committing. 

Is there a specific reason for using a different approach here? I can see that swapping being a little tedious when there are lots of changes to look through

Gitnuro
image

VS Code
image

Sourcetree:
image

mateusduraes:
Hey,

Are there any plans in moving forward with this feature?

maxbrunsfeld (changes requested):
Hi @bnjjj, thanks for this — and sorry for the delayed review.

Since you opened this, we ended up implementing staged/unstaged diffs on `main` ourselves. The bulk of it landed in #58266 (a `BufferDiff` refactor by @cole-miller) and #46541 ("Git partially staged changes"), so the backend part of your PR now overlaps heavily with code that's already merged.

The main difference is in the data model for the index text. Your PR keeps the index content for the staged diff in a **separate, standalone buffer** that's manually kept in sync with the index, which is why the staged view has to be read-only. The refactor on `main` instead lets multiple diffs **share the same base-text buffer entity**: there's a single canonical index buffer that serves as the base of the *unstaged* diff and simultaneously the content of the *staged* diff, and a single head buffer shared by the staged and uncommitted diffs. Because there's one source of truth for the index, the staged view becomes interactive (you can unstage individual hunks straight from it), there's no risk of the two copies drifting, and it avoids a second full copy + foreground reparse of the file.

### UI Work Needed

We are still open to contributions that add new UI around these diffs. Right now, the staged and unstaged diffs are only accessible by right-clicking a file in the git panel and choosing `Staged Changes` and `Unstaged Changes` - and we'd like to add a more clear way of accessing these views.

My concern about the dropdown that you added is that it might not interact well with the *search bar* that appears when you type `command-f` in the project diff view. It seems like that header could get very crowded.

If you're interested, you could rebase this PR onto main, removing the changes to the model layer (since main already has the needed core functionality) leaving only the UI parts of your PR, and share a video of what it looks like on a small screen when you have the search active.

SomeoneToIgnore:

PR #48814, opened 2026-02-09
devcontainer: Support `.devcontainer.json` in project root

oliverbarnes:
Closes #48683

Per the devcontainer spec, `.devcontainer.json` in the project root is
a valid config location. It is only used when no configurations are
found inside `.devcontainer/`.

Extract `find_configs_in_snapshot` for testability and add tests.

Release Notes:

- Added support for `.devcontainer.json` in project root

### DIFF
--- a/crates/dev_container/src/devcontainer_api.rs
+++ b/crates/dev_container/src/devcontainer_api.rs
@@ -35,7 +35,7 @@ impl DevContainerConfig {
 
     pub fn root_config() -> Self {
         Self {
-            name: "default".to_string(),
+            name: "root".to_string(),
             config_path: PathBuf::from(".devcontainer.json"),
         }
     }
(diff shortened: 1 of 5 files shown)


KyleBarton:
@oliverbarnes thanks for this! Just to check - is this ready for review? I see you changed the title, but the PR is still in draft.

KyleBarton:
Awesome, take your time, just wanted to make sure you weren't waiting on us. Thanks again!

crates/dev_container/src/devcontainer_api.rs
```diff
@@ -153,9 +161,14 @@ fn use_podman(cx: &mut AsyncWindowContext) -> bool {
...
 ///
-/// This function scans for:
+/// Per the devcontainer spec, configurations are searched in this order of precedence:
 /// 1. `.devcontainer/devcontainer.json` (the default location)
-/// 2. `.devcontainer/<subfolder>/devcontainer.json` (named configurations)
+/// 2. `.devcontainer.json` in the project root (used as default when #1 is absent)
+/// 3. `.devcontainer/<subfolder>/devcontainer.json` (named configurations)
+///
+/// `.devcontainer/devcontainer.json` and `.devcontainer.json` both serve as the
+/// "default" config — only the higher-precedence one is included. Subfolder
```
KyleBarton: This comment doesn't sound right - both root and default are included. I think that's the correct behavior, but the wording here should change I think

crates/dev_container/src/devcontainer_api.rs
```diff
@@ -177,82 +190,102 @@ pub fn find_devcontainer_configs(cx: &mut AsyncWindowContext) -> Vec<DevContaine
...
-                entry.is_dir()
-            );
+/// Scans a worktree snapshot for devcontainer configurations.
+///
+/// Per the devcontainer spec, configurations are searched in this order of precedence:
+/// 1. `.devcontainer/devcontainer.json` (the default location)
+/// 2. `.devcontainer.json` in the project root (used as default when #1 is absent)
+/// 3. `.devcontainer/<subfolder>/devcontainer.json` (named configurations)
+///
+/// `.devcontainer/devcontainer.json` and `.devcontainer.json` both serve as the
```
KyleBarton: We can probably keep either doc comments here or those above `find_devcontainer_configs` but they are a bit redundant

KyleBarton (commented):
Looks great! Had a couple of nits on the comments, if you don't mind taking a quick look. Otherwise, thanks for a very clean and well-tested change!

crates/dev_container/src/devcontainer_api.rs
oliverbarnes: Good catch, thanks - fixed

crates/dev_container/src/devcontainer_api.rs
oliverbarnes: Agreed - done.

KyleBarton (approved):

PR #48819, opened 2026-02-09
gpui: Make `on_drag` drag-start threshold configurable

Chriss4123:
## Problem
`StatefulInteractiveElement::on_drag(...)` used a hardcoded 2px pointer-movement threshold before starting a drag. On high-DPI displays / trackpads this can be too small and can lead to accidental drags when the user intended a click. There was no public API to adjust this per element.

## Solution
Make the drag-start threshold configurable on a per-element basis while preserving the existing default behavior.

## User-Facing API
- Fluent (per-element):

```rust
use gpui::{div, px, StatefulInteractiveElement};

div()
    .id("row")
    .drag_threshold(px(6.))
    .on_drag(MyDrag, |_, _, _, cx| cx.new(|_| Preview));
```

- Imperative:
  - `Interactivity::drag_threshold(...)`

Defaults remain unchanged (2px) unless an override is set.

## Tests
- Added GPUI tests validating:
  - Default behavior: drag starts only after moving more than 2px.
  - Override behavior: drag starts only after moving more than the configured threshold.

Closes #ISSUE

- [x] Code Reviewed
- [x] Manual QA

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/elements/div.rs
+++ b/crates/gpui/src/elements/div.rs
@@ -44,7 +44,7 @@ use util::ResultExt;
 
 use super::ImageCacheProvider;
 
-const DRAG_THRESHOLD: f64 = 2.;
+const DEFAULT_DRAG_THRESHOLD: f64 = 2.;
 const TOOLTIP_SHOW_DELAY: Duration = Duration::from_millis(500);
 const HOVERABLE_TOOLTIP_HIDE_DELAY: Duration = Duration::from_millis(500);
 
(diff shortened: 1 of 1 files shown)


Veykril (changes requested):
I fail to see the usefulness of having this configurable per element. This feels like it should be configurable per app or per window.

SomeoneToIgnore:
Will close this as first conflicts start to appear and we did not agree on the design for a few months.

Chriss4123:
> Will close this as first conflicts start to appear and we did not agree on the design for a few months.

@SomeoneToIgnore

Sorry, completely forgot about this.

The reason I wanted per-element is customizability. The code paths already exist with a hardcoded value, and this PR simply makes that value configurable.

Suppose you have an application with a small reorderable list and a larger card based system where each card can be interacted with. Naturally, you'd want the two to have different thresholds for drag triggering: a smaller one for the list and bigger for the cards.

In practice with the many apps I have created with gpui, I've had to make my own bespoke logic for extending the threshold in certain drag and drop flows, and I have always ended up with different values to my taste.

I believe a global config is ultimately worse. If you look at most drag and drop libraries for web applications, they allow a case-by-case configuration of the threshold.

Perhaps we can do a middle ground: a global configuration AND also a per-element config, with the per-element superseding the global configuration.

Does this work for you? Once again apologies for the idling.

SomeoneToIgnore:
No worries, we just close entries in this large list to keep it more actionable.

I think that compromise and maybe an example would work?
I'm not way working with `gpui` myself to claim that the new PR will be reviewed timely, though...

PR #48837, opened 2026-02-10
repl: Be more resilient with message parsing

rgbkrk:
Closes #21529

Show a toast when a message from a Jupyter kernel doesn't parse properly.

Discovered that some kernels don't include their execution count in an `ExecuteReply` which is required in the jupyter protocol. Upstream I'll go make that field either be an `Option` or just stick `#[serde(default)]` on it so it's not a change in the interface.

Release Notes:

- Show error when parsing a message from a Jupyter kernel fails

### DIFF
--- a/crates/repl/src/kernels/native_kernel.rs
+++ b/crates/repl/src/kernels/native_kernel.rs
@@ -5,13 +5,13 @@ use futures::{
     io::BufReader,
     stream::FuturesUnordered,
 };
-use gpui::{App, AppContext as _, Entity, EntityId, Task, Window};
+use gpui::{App, AppContext as _, Entity, EntityId, Task, Window, WindowHandle};
 use jupyter_protocol::{
     ExecutionState, JupyterKernelspec, JupyterMessage, JupyterMessageContent, KernelInfoReply,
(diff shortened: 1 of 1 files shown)


crates/repl/src/kernels/native_kernel.rs
```diff
@@ -173,16 +173,65 @@ impl NativeRunningKernel {
...
+                            Err(
+                                ref err @ (RuntimeError::ParseError { .. }
+                                | RuntimeError::SerdeError(_)),
+                            ) => {
+                                let error_detail =
+                                    format!("Kernel issue on {channel} channel\n\n{err}");
+                                log::warn!("kernel: {error_detail}");
+                                let workspace_window: Option<
+                                    WindowHandle<workspace::Workspace>,
+                                > = session
```
miguelraz: Nit: This breaking line on the `Option<...>` looks awful, don't you think?

Not a blocker if it compiles though.

rgbkrk: I just took off the type assertion and let the compiler infer it.

maxdeviant (approved):

PR #48844, opened 2026-02-10
repl: Support Jupyter widgets

rgbkrk:
This PR adds basic support for IPython/Jupyter widgets.

At the very least, I wanted to get `tqdm` supported which as every ML engineer knows is the most used library in machine learning. The progress bar.

https://github.com/user-attachments/assets/91a95b47-0b32-4d88-ba72-f742d4ddc33f

What took way more work though is supporting the Jupyter widgets which handle more generic components. I did a _lot_ of them with what we have in GPUI/Zed here. They could be better but it's a start.

widgets

Release Notes:

- Added Jupyter/IPython widgets support to REPL

### DIFF
--- a/crates/repl/src/outputs.rs
+++ b/crates/repl/src/outputs.rs
@@ -34,7 +34,7 @@
 //! interpreting and displaying various types of Jupyter output.
 
 use editor::{Editor, MultiBuffer};
-use gpui::{AnyElement, ClipboardItem, Entity, EventEmitter, Render, WeakEntity};
+use gpui::{AnyElement, ClipboardItem, Entity, EventEmitter, Render, Subscription, WeakEntity};
 use language::Buffer;
 use menu;
 use runtimelib::{ExecutionState, JupyterMessage, JupyterMessageContent, MimeBundle, MimeType};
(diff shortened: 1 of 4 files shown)


rgbkrk:
Ok, made sure the styling works decently for the huggingface hub login widget. 

image

crates/repl/src/outputs.rs
```diff
@@ -752,6 +797,11 @@ impl ExecutionView {
 
 impl Render for ExecutionView {
     fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        self.widget_store.update(cx, |store, cx| {
+            store.create_missing_text_editors(window, cx);
+            store.create_missing_markdown_views(cx);
+        });
```
Veykril, line 803: Can we avoid doing this in render? That feels a bit out of place

crates/repl/src/outputs/widget.rs
```diff
@@ -0,0 +1,975 @@
...
+                        return;
+                    }
+                    let mut state_patch = serde_json::Map::new();
+                    state_patch.insert("value".into(), serde_json::Value::String(new_text));
+                    this.send_update(&model_id, state_patch, cx);
+                }
+            });
+
+            self.text_editors.insert(model_id, editor);
+            self._editor_subscriptions.push(subscription);
```
Veykril, line 137: we never remove the pushed subscriptions

rgbkrk:
I think for now I'll just close this and someone else can revisit this if it's of interest.

PR #48850, opened 2026-02-10
gpui: Fix double-close of fd in read_fd on linux

Albab-Hasan:
`read_fd` calls `File::from_raw_fd(fd.as_raw_fd())`, which creates a new `File` that assumes ownership of the raw fd, while the original `FileDescriptor` also retains ownership. when both drop at the end of the function, the same fd is closed twice. between the two closes another thread can receive that fd number from the kernel for an unrelated file/socket, and the second close silently kills it. this can cause hard to diagnose data corruption or i/o failures.

fix: use `fd.into_raw_fd()` instead of `fd.as_raw_fd()` to transfer ownership to the `File` without running `FileDescriptor`'s drop.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/platform/linux/platform.rs
+++ b/crates/gpui/src/platform/linux/platform.rs
@@ -9,7 +9,7 @@ use std::{
     ffi::OsString,
     fs::File,
     io::Read as _,
-    os::fd::{AsFd, AsRawFd, FromRawFd},
+    os::fd::{AsFd, FromRawFd, IntoRawFd},
     time::Duration,
 };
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

Albab-Hasan:
@Veykril you are most welcome! I would really appreciate it if u could check out this pr of mine too.

https://github.com/zed-industries/zed/pull/48789

also is there anyway i can talk to you? I would like to ask a few questions if you dont mind.

PR #48330, opened 2026-02-04
gpui: Allow OS caption/buttons for custom Windows titlebar

momota1029:
## Summary
Fixes an issue where GPUI's handling of `WM_NCLBUTTONDOWN` prevented Windows from processing default titlebar interactions (dragging, caption buttons, and border resize).

## Changes
- Allow OS to handle `WM_NCLBUTTONDOWN` events for `HTCAPTION`, caption button areas, and resize border areas (`HTLEFT`, `HTRIGHT`, `HTTOP`, `HTBOTTOM`, `HTTOPLEFT`, `HTTOPRIGHT`, `HTBOTTOMLEFT`, `HTBOTTOMRIGHT`)
- Use current cursor position in `WM_NCHITTEST` for accurate hit-testing

## Testing
- Manual testing on Windows: titlebar dragging works
- Manual testing on Windows: minimize/maximize/close buttons work
- Manual testing on Windows: window border resize works in all directions

Release Notes:
- N/A

PR #48858, opened 2026-02-10
gpui: Prioritize topmost hitbox for window control hit testing

momota1029:
Closes #ISSUE

## Summary
Follow-up to #48330.

Fix a Windows regression where interactive child elements inside a
`WindowControlArea::Drag` region do not receive clicks because hit-testing can
resolve to the ancestor drag area (`HTCAPTION`).

## Changes
- Update `on_hit_test_window_control` in `crates/gpui/src/window.rs`
- Evaluate hovered hitboxes in topmost-first order (`mouse_hit_test.ids`)
- Return a window control area only if the topmost hovered hitbox is a control hitbox
- If the topmost hovered hitbox is not a control hitbox, return `None`

## Testing
- Reproduced using the report snippet from the PR comment (custom titlebar with a button inside `WindowControlArea::Drag`)
- Verified that the child button click handler fires after this change on Windows
- Verified that drag/caption hit-testing still works for intended titlebar controls

Release Notes:

- Fixed Windows custom titlebar hit-testing so clickable child UI is not overridden by ancestor `WindowControlArea::Drag`.

### DIFF
--- a/crates/gpui/src/window.rs
+++ b/crates/gpui/src/window.rs
@@ -1310,11 +1310,19 @@ impl Window {
             Box::new(move || {
                 handle
                     .update(&mut cx, |_, window, _cx| {
-                        for (area, hitbox) in &window.rendered_frame.window_control_hitboxes {
-                            if window.mouse_hit_test.ids.contains(&hitbox.id) {
-                                return Some(*area);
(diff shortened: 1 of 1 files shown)


Veykril:
Superseded by https://github.com/zed-industries/zed/pull/49781

PR #48861, opened 2026-02-10
Save edited buffers before running a task

### LINKED ISSUE #10251: Optionally save dirty buffers before running a task
### Check for existing issues

- [X] Completed

### Describe the feature

Add a setting that forces Zed to save dirty buffers before running a task.

VSCode offers the "always" (default), "never" and "prompt" options, which seems to cover almost all needs.

Related Issues:

- Besides being a useful feature per se, it will be another step to a comfortable tabless mode (#6424)

N/A

abenea:
Save edited buffers before running a task

Introduces a new task field for configuring which buffers are saved. For now, this defaults to saving all buffers, but in the future we could have a global task template to configure this setting for dynamically created tasks.

Closes #10251.

Release Notes:

- Edited buffers are now saved before running a task. This can be configured with the new "save" field in `tasks.json`.

### DIFF
--- a/assets/settings/initial_tasks.json
+++ b/assets/settings/initial_tasks.json
@@ -48,6 +48,11 @@
     "show_summary": true,
     // Whether to show the command line in the output of the spawned task, defaults to `true`.
     "show_command": true,
+    // Which edited buffers to save before running the task:
+    // * `all` — save all edited buffers
+    // * `current` — save current buffer only
+    // * `none` — don't save any buffers
+    "save": "all",
     // Represents the tags for inline runnable indicators, or spawning multiple tasks at once.
     // "tags": []
   },
--- a/crates/project/tests/integration/debugger.rs
+++ b/crates/project/tests/integration/debugger.rs
@@ -3,7 +3,7 @@ mod go_locator {
     use dap::{DapLocator, adapters::DebugAdapterName};
     use gpui::TestAppContext;
     use project::debugger::locators::go::{DelveLaunchRequest, GoLocator};
-    use task::{HideStrategy, RevealStrategy, RevealTarget, Shell, TaskTemplate};
+    use task::{HideStrategy, RevealStrategy, RevealTarget, SaveStrategy, Shell, TaskTemplate};
     #[gpui::test]
     async fn test_create_scenario_for_go_build(_: &mut TestAppContext) {
         let locator = GoLocator;
@@ -22,6 +22,7 @@ mod go_locator {
             tags: vec![],
             show_summary: true,
             show_command: true,
+            save: SaveStrategy::default(),
         };
 
         let scenario = locator
@@ -49,6 +50,7 @@ mod go_locator {
             tags: vec![],
             show_summary: true,
             show_command: true,
+            save: SaveStrategy::default(),
         };
 
         let scenario = locator
@@ -187,6 +189,7 @@ mod go_locator {
             tags: vec![],
             show_summary: true,
             show_command: true,
+            save: SaveStrategy::default(),
         };
 
         let scenario = locator
@@ -221,6 +224,7 @@ mod python_locator {
             shell: task::Shell::System,
             show_summary: false,
             show_command: false,
+            save: task::SaveStrategy::default(),
         };
 
         let expected_scenario = DebugScenario {
--- a/crates/workspace/src/tasks.rs
+++ b/crates/workspace/src/tasks.rs
@@ -6,11 +6,13 @@ use language::Buffer;
 use project::{TaskSourceKind, WorktreeId};
 use remote::ConnectionState;
 use task::{
-    DebugScenario, ResolvedTask, SharedTaskContext, SpawnInTerminal, TaskContext, TaskTemplate,
+    DebugScenario, ResolvedTask, SaveStrategy, SharedTaskContext, SpawnInTerminal, TaskContext,
+    TaskTemplate,
 };
 use ui::Window;
+use util::TryFutureExt;
 
-use crate::{Toast, Workspace, notifications::NotificationId};
+use crate::{SaveIntent, Toast, Workspace, notifications::NotificationId};
 
 impl Workspace {
     pub fn schedule_task(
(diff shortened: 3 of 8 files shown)


assets/settings/default.json
```diff
@@ -1163,6 +1163,11 @@
   // 4. Save when idle for a certain amount of time:
   //     "autosave": { "after_delay": {"milliseconds": 500} },
   "autosave": "off",
+  // When to save all dirty buffers before running a task. This setting can be either:
+  //
+  // - "always": save without asking
+  // - "never": don't save
+  "save_before_task_run": "always",
```
SomeoneToIgnore: This seems not the approach we want to go in general: not every task might need saving currently active file, and blocking them all on such saves is not flexible.

It seems that the task template could get a field instead.
We would need to update the initial template (https://github.com/zed-industries/zed/blob/3bc4b584b17b1b2858021ceef52dd27cfeb9cd83/assets/settings/initial_tasks.json#L1) , the docs in an md file, the code.

In addition to that, I wonder if "save all" is needed for all the tasks, and should we introduce "save current" (default) and "save all" as options to "never".
That might require some names to change, both enum variants' and new field's.

crates/workspace/src/tasks.rs
```diff
@@ -73,11 +75,25 @@ impl Workspace {
             });
         }
 
-        if let Some(terminal_provider) = self.terminal_provider.as_ref() {
-            let task_status = terminal_provider.spawn(spawn_in_terminal, window, cx);
-
-            let task = cx.spawn(async |w, cx| {
-                let res = cx.background_spawn(task_status).await;
+        if let Some(terminal_provider) = self.terminal_provider.clone() {
```
SomeoneToIgnore: Let's rework this code to avoid `terminal_provider` being moved to some closures: it may lead to subtle leaks.
This means that `Box` -> `Arc` change should also be undone.

We have `w` (which ideally should be renamed into `workspace`), that one could be always `workspace.update(cx, |workspace, cx| {...})` later — it will return an `Err` if the entire window is closed and removed, so the usual reaction on this error would be to exit before.

Both `save_all_internal` and `terminal_provider.spawn` return a `Task` that can be moved outside of those `update`/`read_with` closures and then `.await`ed.

SomeoneToIgnore (commented):
Thank you, after a few tweaks this should be a nice way to add the functionality in general.

assets/settings/default.json
abenea: > I wonder if "save all" is needed for all the tasks, and should we introduce "save current" (default) and "save all" as options to "never".

I'm assuming by "save current" you mean save current file. The editors I'm familiar with (Emacs and VSCode) don't even offer a "save current" setting. If this would be the default, it would cause confusion for people coming from other editors. But it's also something that woudn't work for me because I don't want go through the hassle of defining custom tasks and use default tasks the language provides (run or debug current project). If the default was "save current" I wouldn't be able to use the default tasks.

crates/workspace/src/tasks.rs
abenea: Thanks for the detailed suggestions! I am not sure this is what you meant by moving the `Task` outside of the `update_in` but I was able to eliminate some nesting and the code looks nicer now.

assets/settings/default.json
abenea: I implemented the task field and renamed the options, but also kept the global setting.

SomeoneToIgnore: Thank you, but let's not have any related global settings and whatever extra code for this.
The right way to solve it seems to be a "global" task template, not separate settings per each of the template's field.

SomeoneToIgnore (changes requested):
Great to see entities not moved into tasks now, we're almost there.

To add to the settings discussion, let's add the "save the current file" when making the rest of the changes.
It makes sense to have the "save all" by default for now, and work on the "global template" to override global task defaults later.

assets/settings/default.json
abenea: The global task template is indeed a much better design.

I have reverted the global setting. I also removed the SaveStrategy::Default enum variant because it's no longer needed. I assume a more generic way of representing a missing field will have to be implemented for the global task template feature.

assets/settings/initial_tasks.json
```diff
@@ -48,6 +48,11 @@
     "show_summary": true,
     // Whether to show the command line in the output of the spawned task, defaults to `true`.
     "show_command": true,
+    // Which edited buffers to save before running the task:
+    // * `all` — save all edited buffers
+    // * `current` — save current buffer only
+    // * `none` — don't save any buffers
+    "save": "all",
```
SomeoneToIgnore, line 55: Great, let's also mention it in the `task.md`'s template.

crates/project/tests/integration/debugger.rs
```diff
@@ -22,6 +22,7 @@ mod go_locator {
             tags: vec![],
             show_summary: true,
             show_command: true,
+            save: Default::default(),
```
SomeoneToIgnore: It's quite hard to read and comprehend, can we at least use `SaveStrategy::default()` here and in other places?

crates/workspace/src/tasks.rs
```diff
@@ -73,11 +75,39 @@ impl Workspace {
...
-            let task = cx.spawn(async |w, cx| {
-                let res = cx.background_spawn(task_status).await;
+        if self.terminal_provider.is_some() {
+            let task = cx.spawn_in(window, async move |w, cx| {
+                let save_action = match spawn_in_terminal.save {
+                    SaveStrategy::All => {
+                        let save_all = w.update_in(cx, |workspace, win, cx| {
+                            let task = workspace.save_all_internal(SaveIntent::SaveAll, win, cx);
+                            // Match the type of the other arm by ignoring the bool value returned
+                            win.spawn(cx, async |_| task.await.map(|_| ()))
```
SomeoneToIgnore: Wonder if it's possible to do `cx.background_spawn` here instead?

crates/workspace/src/tasks.rs
```diff
@@ -73,11 +75,39 @@ impl Workspace {
...
-        if let Some(terminal_provider) = self.terminal_provider.as_ref() {
-            let task_status = terminal_provider.spawn(spawn_in_terminal, window, cx);
-
-            let task = cx.spawn(async |w, cx| {
-                let res = cx.background_spawn(task_status).await;
+        if self.terminal_provider.is_some() {
+            let task = cx.spawn_in(window, async move |w, cx| {
+                let save_action = match spawn_in_terminal.save {
+                    SaveStrategy::All => {
+                        let save_all = w.update_in(cx, |workspace, win, cx| {
```
SomeoneToIgnore: Can we use full names here, please?
`window`, `workspace`, etc.

crates/workspace/src/tasks.rs
```diff
@@ -73,11 +75,39 @@ impl Workspace {
...
+                        });
+                        save_current.ok()
+                    }
+                    SaveStrategy::None => None,
+                };
+                if let Some(save_action) = save_action {
+                    save_action.log_err().await;
+                }
+                let task_status = w.update_in(cx, |workspace, win, cx| {
+                    if let Some(terminal_provider) = workspace.terminal_provider.as_ref() {
```
SomeoneToIgnore: Looks like here we could do `workspace.terminal_provider.as_ref().map(..` and simplify the branching a bit, with a single `if let` after?

SomeoneToIgnore (commented):
Great, thank you!

All that's left are style nits and one extra doc entry and we're good to merge.

crates/workspace/src/tasks.rs
abenea: Done, but I don't know if it makes a big difference. `save_all_internal` does little before calling `cx.spawn_in(window, ...` to do most of its work, which goes to the foreground executor if my understanding is correct.

crates/workspace/src/tasks.rs
abenea: Done, but I kept `cx` because it seems to be a convention in gpui.

SomeoneToIgnore (approved):
Thank you so much!

docs/src/tasks.md
```diff
@@ -51,6 +51,11 @@ Zed supports ways to spawn (and rerun) commands using its integrated [terminal](
     "show_summary": true,
     // Whether to show the command line in the output of the spawned task, defaults to `true`.
     "show_command": true
```
SomeoneToIgnore:
```suggestion
    "show_command": true,
```

injust:
> seems that we cannot use the "save a single file" mode either

This is what I observed when testing the Python tasks:

- The play button appears in the gutter for all Python buffers, but the button only works in saved files. So "save a single file" works fine here.
- If you run the "task: spawn" command, the Python tasks only appear when the currently focused file is a saved Python buffer. _However_, they also appear as a recently run task. I haven't tested this, but I assume if you run that task while an unsaved buffer is focused, then it will try to save that file?

So defaulting to "save the current file" is significantly less breaking of a change, but there is still an edge case.

PR #48862, opened 2026-02-10
rope: Fix `point_utf16_to_offset` returning relative offset instead of absolute

jrdi:
`ChunkSlice::point_utf16_to_offset` returned `line.len()` (a relative line byte length) instead of `row_offset_range.end` (the absolute byte offset within the chunk) when clipping an out-of-range UTF-16 column on a non-first row.

This caused incorrect position calculations in LSP coordinate conversions whenever a UTF-16 point had a column beyond the line's extent on any row after the first in a chunk.

Added a regression test covering clipping on both row 0 and row 1. Not sure if it's the best place for a test like this one though.

Release Notes:

- N/A


Veykril (approved):
Thanks!

PR #33174, opened 2025-06-21
bedrock: Fix subsequent bedrock tool calls fail

kuzaxak:
Closes #30714

Bedrock converse api expect to see tool options if at least one tool was used in conversation in the past messages.

Right now if `LanguageModelToolChoice::None` isn't supported edit agent [remove][1] tools from request. That point breaks Converse API of Bedrock. As was proposed in [the issue][2] we won't drop tool choose but instead will deny any of them if model will respond with a tool choose.

[1]: https://github.com/x-qdo/zed/blob/fceba6c79540677c2504d2c22191963b6170591a/crates/assistant_tools/src/edit_agent.rs#L703
[2]: https://github.com/zed-industries/zed/issues/30714#issuecomment-2886422716

Release Notes:

- Fixed bedrock tool calls in edit mode

PR #48863, opened 2026-02-10
bedrock: Add placeholder tool when summarising threads with tool history

dsturnbull:
Bedrock Converse API requires toolConfig whenever messages have tool use
    or result blocks. PR #33174 handled the case where tools are defined but
    tool_choice is None. This change addresses another case: the tools array
    is empty (e.g. thread summarisation) but messages still have tool blocks
    from conversation history. A placeholder tool satisfies the requirement.

Testing:

Tested manually by:
1. Starting a conversation with Bedrock that uses tools
2. Triggering thread summarisation
3. Confirming summarisation now succeeds instead of failing with an API error

Release Notes:

- Fixed Bedrock thread summarization failing when conversation had tools


benbrandt (approved):
thanks!

PR #48893, opened 2026-02-10
Add configurable autoscroll strategy for go to definition

torarnv:
When navigating to a definition (e.g., via editor::GoToDefinition or Cmd+Click), the target is now positioned according to a new configurable setting rather than always using the minimal "fit" behavior.

The new `go_to_definition_autoscroll` setting supports:
- Simple strategies: "fit", "center", "focused", "top", "bottom"
- Parameterized strategies with offsets:
  - {"strategy": "center_relative", "offset": -10} for positioning relative to center (negative=above, positive=below)
  - {"strategy": "top_relative", "offset": N} for N lines from top
  - {"strategy": "bottom_relative", "offset": N} for N lines from bottom

The default remains "fit" to preserve the original behavior.

Implementation details:
- Added AutoscrollStrategy::CenterRelative(isize) variant
- Created GoToDefinitionAutoscroll enum with Simple and WithOffset variants
- Updated go_to_singleton_buffer_range to use the setting
- Added default value and documentation to default.json

Release Notes:

- Add configurable autoscroll strategy for go to definition

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -318,6 +318,13 @@
   // 1. Do nothing: `none`
   // 2. Find references for the same symbol: `find_all_references` (default)
   "go_to_definition_fallback": "find_all_references",
+  // How to position the cursor when navigating to a definition.
+  // Options: "fit" (default), "center", "focused", "top", "bottom"
+  // Or use an object for relative positioning:
+  //   {"strategy": "center_relative", "offset": -10}  // 10 lines above center
+  //   {"strategy": "top_relative", "offset": 5}       // 5 lines from top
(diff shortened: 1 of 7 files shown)


torarnv:
My personal preference for this is centered, but with a slight negative (-5) offset, hence the addition of a new scroll strategy in addition to the setting :)

torarnv:
I'm coming from Sublime Text, where I believe the behavior is `center`. Basically Zed's default `fit` behavior was so jarring that I had to implement this patch. I'm okey keeping it in my local tree and rebasing it, as without it Zed is kind-of unusable for me, so I hope that a discussion of some kind of default change or configurable option would bring something down the line.

PR #48936, opened 2026-02-11
Fix discard not reloading buffer from disk

TomPlanche:
Closes #48308.

Before:
zed-#48308-before

After:
zed-#48308-after

Release Notes:
- Fixed file changes not being discarded when closing a tab with "Don't Save" or "Discard all".


crates/workspace/src/pane.rs
```diff
@@ -2184,6 +2184,12 @@ impl Pane {
 
         let path_style = project.read_with(cx, |project, cx| project.path_style(cx));
         if save_intent == SaveIntent::Skip {
+            let can_save = cx.update(|_window, cx| item.can_save(cx))?;
```
SomeoneToIgnore: This logic seems to be wrong?

1. `can_save` returns `true` for a multi buffer: https://github.com/zed-industries/zed/blob/44015e0379b6c868258318d9c159134adbf801d9/crates/editor/src/items.rs#L847-L854
2. MutliBuffer's reload will reload all its contents: https://github.com/zed-industries/zed/blob/44015e0379b6c868258318d9c159134adbf801d9/crates/editor/src/items.rs#L934-L958

We do not want to reload from FS files that were not involved into the discard dialogue, do we?
Sounds like we need a multi buffer test for this and a bit more convoluted logic for a fix?

SomeoneToIgnore (approved):
I would expect multi buffer to discard the related changes too, but sure we can start off this — ready to merge it after CI is green.

PR #48953, opened 2026-02-11
git: Disable some more expensive things for the split diff's left-hand side

cole-miller:
- Conflict UI
- LSP data
- Runnables

Release Notes:

- Improved performance when opening the split diff view.

### DIFF
--- a/crates/editor/src/split.rs
+++ b/crates/editor/src/split.rs
@@ -528,6 +528,7 @@ impl SplittableEditor {
             editor.set_show_vertical_scrollbar(false, cx);
             editor.disable_lsp_data();
             editor.disable_runnables();
+            editor.disable_diagnostics(cx);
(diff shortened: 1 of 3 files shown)


crates/editor/src/split.rs
```diff
@@ -526,6 +526,8 @@ impl SplittableEditor {
             editor.set_delegate_stage_and_restore(true);
             editor.set_delegate_open_excerpts(true);
             editor.set_show_vertical_scrollbar(false, cx);
+            editor.disable_lsp_data();
```
SomeoneToIgnore, line 529: I wonder if `disable_diagnostics` is also worth adding?
Unfortunately, despite also being an "lsp_data", it has its own lifecycle hence managed separately.

cole-miller: Sounds good to me!

PR #48954, opened 2026-02-11
extensions_ui: Apply category filter to dev extensions

chenwuji2000-cyber:
Following up on #49183 which fixed the category filter for remote extensions.

This PR applies the same category filter logic to **dev extensions**. Previously, dev extensions were always shown regardless of the selected category filter (e.g., "Themes", "Languages").

Changes:
- Add `filtered_dev_extension_indices` to track which dev extensions match the active `provides_filter`
- Add `dev_extension_matches_provides()` helper to map `ExtensionManifest` fields to `ExtensionProvides` variants
- Update `render_extensions()` and list count to use filtered dev extension indices

Release Notes:

- Fixed extension category filter not applying to dev extensions in the extensions panel.


chenwuji2000-cyber (commented):
learned from it!

maxdeviant (approved):
Thank you!

PR #48970, opened 2026-02-11
editor: Fix removing indentation while converting selected lines

CCXLV:
Closes #48945 

Release Notes:

- Fixed removing line indentations when formatting the lines with:
     - `editor: convert to title case` 
     - `editor: convert to lower camel case`
     - `editor: convert to upper camel case`

As well as the commands:
  -  `editor: convert to kebab case`
  -  `editor: convert to sentence case`
  - `editor: convert to snake case`

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -12442,11 +12442,7 @@ impl Editor {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
-        self.manipulate_text(window, cx, |text| {
-            text.split('\n')
-                .map(|line| line.to_case(Case::Title))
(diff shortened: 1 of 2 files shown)


crates/editor/src/editor.rs
```diff
@@ -12564,6 +12558,16 @@ impl Editor {
         })
     }
 
+    fn convert_text(text: &str, case: Case) -> String {
```
Ahmad-Alsaleh: I recommend renaming this function to `convert_text_case` for better readability. `convert_text` is not clear IMO. Convert it to what exactly, to bytes? UTF-16? to upper case? etc.

PR #48978, opened 2026-02-11
Use document symbols' ranges to derive their outline labels

SomeoneToIgnore:
Follow-up of https://github.com/zed-industries/zed/pull/48780

Before:
before

After:
after

Release Notes:

- N/A


xdBronch:
is it a bug that `super` and `crate` arent being highlighted there or is it just something with the theme?

PR #48904, opened 2026-02-11
docs: Update the formatting section for Python

lingyaochu:
I have added the `code_actions_on_format` setting. I also temporarily removed the mention of `"formatter": null`. Based on our findings in #48600, its current behavior is inconsistent, so I've omitted it for now to avoid confusing users. I will add it back once the underlying issue is fixed (likely in a subsequent PR).

Closes #48600.

- [ ] Tests or screenshots needed?
- [x] Code Reviewed
- [x] Manual QA

Release Notes:

- N/A

Issue #48600, opened 2026-02-06
Can't turn off or change Python formatter


csk-ableton:
### Reproduction steps

1. Open the attached Python project which disables the Python formatter
2. Trust its content
3. Open script.py
4. Save the file

[python-project.zip](https://github.com/user-attachments/files/25132075/python-project.zip)

### Current vs. Expected behavior

Current behaviour: Zed auto-formats the file and removes blank lines between imports. Turning the ruff language server off avoids the additional formatting. The same can be observed when changing the external formatter to `black`. `ruff` is still formatting for me.

Expected behaviour: Zed does not change the formatting

I feel like I'm missing something, given it's such a basic functionality.

### Zed version and system specs

Zed: v0.222.4+stable.147.b385025df963c9e8c3f74cc4dadb1c4b29b3c6f0 (Zed) 
OS: macOS 15.7.3
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json
{
  "languages": {
    "Python": {
      "formatter": null,
    },
  },
}
```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

lingyaochu:
While the issue with `"formatter": null` persists, I’ve identified an additional setting affecting import sorting logic:
```json
{
  "languages": {
    "Python": {
      "code_actions_on_format": {
        "source.organizeImports.ruff": true,
      },
    },
  },
}
```
This setting appears to default to true and is triggered before the formatter runs. However, I couldn't find any mention of this in the Python-specific documentation; the only reference to `code_actions_on_format` I could find is on the [ESLint](https://zed.dev/docs/languages/javascript?highlight=code_actions_on_format#eslint) page.

lingyaochu:
I am a bit confused by the current documentation regarding how to disable formatting.
In [this](https://zed.dev/docs/configuring-languages?highlight=formatter%20null#configuring-formatters) section, it suggests: 

> To disable formatting for a specific language:
> 
> ```json
> "languages": {
>   "Markdown": {
>     "format_on_save": "off"
>   }
> }
> ```

However, in the [Python](http://zed.dev/docs/languages/python#configuring-formatting) section, it provides two different ways to disable formatting:

> You can disable format-on-save for Python files in your settings.json:
[...]

PR #48991, opened 2026-02-12
editor: Add an option to disable the formatter

lingyaochu:
Related issue #48600, following #48904

Setting `"formatter": null` does not actually disable the formatter, as it is functionally equivalent to leaving the setting undefined. In this case, Zed falls back to the default behavior—either "auto" or a specific language server, depending on the language.

This PR adds a "none" option to the formatter setting, allowing users to explicitly disable formatting while keeping `code_actions_on_format` enabled separately.
```json
"formatter": "none"
```

I still have some doubts about the necessity of this setting itself, but if it is a desired addition, I will update the documentation accordingly. If not, please feel free to close this PR.

- [x] Code Reviewed
- [x] Manual QA

Release Notes:

- Added `"formatter": "none"` in settings to explicitly disable the formatter.

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -1703,6 +1703,10 @@ impl LocalLspStore {
                 formatter
             };
             match formatter {
+                Formatter::None => {
+                    zlog::trace!(logger => "skipping formatter 'none'");
+                    continue;
+                }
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Thanks!

Veykril:
yea looks like CI got stuck, will merge after your changes then

PR #48994, opened 2026-02-12
editor: Add command to toggle wrap guide visibility

### LINKED ISSUE #32284: show_wrap_guides setting not toggle'able
### Summary

`show_wrap_guides` not toggle'able

### Description
The `show_wrap_guides` setting controls wether configured wrap guides are visible or not. However, this setting does not seem to have an equivalent `editor: toggle wrap guides` entry in the command menu, which also means you cannot assign a keyboard shortcut to it (afaik).

Steps to reproduce:
1. Set `"wrap_guides": [80, 120]` and `"show_wrap_guides": true`
2. Observe that you see two wrap guides in your editor
3. Set `"show_wrap_guides": false`
4. Observe the wrap guides disappear
5. Search for a command-pallette (shift-ctrl-p) equivalent like `editor: toggle wrap guides` and observe there isn't one
6. Realize you then can't assign a key binding to it and write a bug report

Expected Behavior:
Find a command-pallette equivalent for this setting, set a keybind for it.

Actual Behavior:
Can't find a command-pallette equivalent for this setting, can't set a keybinding for it.

Zed: v0.189.5 (Zed) 
Memory: 125.4 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 4090 || NVIDIA || 575.57.08

menkaru:
Added a command to toggle the 'show wrap guides' setting, to match similar commands available for related settings (for example, the 'soft wrap' and 'show indent guides' settings both have toggles).

The command uses the preexisting override setting in the editor, which looks to be currently used to turn off guides in certain windows. I did a bit of testing and toggling guides on in those windows doesn't seem to break anything.

My most recent commit also modifies the visibility check in the wrap_guides function to make it a little clearer, but also removes the early return. If you'd prefer to keep changes to wrap_guides as minimal as possible, then feel free to omit that last commit.

Relevant discussion #45613
Closes #32284 (bot closed it already a while ago, but it was never actually resolved)


menkaru:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Overall, this looks fine-ish but I think this will cause issues all around in the current state, as we currently use that field to disable wrap guides for certain editors altogether. We'd need to fix this so that this actually toggles wrap guides in the indended editor as well as ensure that deactivation of wrap guides for a certain editor cannot be overwritten by users. What do you think?

MrSubidubi (changes requested):
> Might I suggest renaming show_wrap_guides to something like wrap_guides_override to better reflect its purpose?

I think it would be better if we could merge this into one field on the struct - I think I once did something similar for the minimap, but haven't looked in ages, so would need to check whether we could make that more generic/reuse this here.
Also, could you please add some tests for this behavior? 

Lastly, apologies for the slow reply, got pulled into some other stuff

MrSubidubi:
Will close for the time being, but happy to reopen once you pick this back up!

PR #49015, opened 2026-02-12
settings: Add ability to select audio input/output devices for collab

kubkon:
This PR adds ability to select and test audio input/output devices for use in collaboration setting (which is what the team at Zed relies heavily on). Currently, we only ever used whatever the system default is and it worked well until it didn't - for some reason, when I am on my Linux laptop, I am unable to force Zed to use my external mic + headphones via external USB audio interface. With this PR, now I can list all available devices and select the one I want.

There are still a couple of caveats that we should be aware of:
* I've decided to list *all* available devices meaning on Linux it is quite possible that you may discover that what your desktop environment is reporting to you is a significantly shorter list than what your sound framework/hw is actually exposing. I think this makes sense given my inexperience with audio drivers/devices and frameworks on various OSes so that we get full control over what is available with the goal of being able to come up with some filtering heuristic as we go along.
* We currently populate the list of available audio devices only once at startup meaning if you unplug your device while you have Zed running this will not register until you restart Zed which is a PITA. However, in order to keep the changes manageable I thought it would be best to do minimal work in this regard now, and iterate on this some more in the near future. After all, we don't really monitor device changes on any platform except macOS anyhow, so it might be the case that when I get round to implementing this I will have the opportunity to tackle both at the same time.
* In order to get a valid list of all audio devices using `cpal` crate (which is the building block of `rodio`), I had to bump `cpal` to 0.17, and pin `rodio` to a more recent commit sha as a result, so if you see any regressions, lemme know and/or feel free to revert this PR.
* Finally, I've done my best to integrate this with the settings UI, but I am sure more could be done in terms of styling, etc.

Some screenshots:

Screenshot From 2026-02-12 11-40-04
Screenshot From 2026-02-12 11-40-16

Release Notes:

- Added ability to select audio input/output devices as part of Collaboration page in Settings. Added ability to test selected devices with a simple playback loop routing input directly into output for easier debugging of your audio devices.


crates/audio/src/audio.rs
```diff
@@ -108,51 +120,58 @@ impl Default for Audio {
...
+                    });
+                output_handle.mixer().add(source);
+            }
+
+            #[cfg(any(
+                any(all(target_os = "windows", target_env = "gnu")),
+                target_os = "freebsd"
+            ))]
+            {
+                let source = rodio::source::Zero::<f32>::new(CHANNEL_COUNT, SAMPLE_RATE);
```
ognevny, line 164:
```suggestion
                let source = rodio::source::Zero::new(CHANNEL_COUNT, SAMPLE_RATE);
```
this change broke build for MinGW. I'll send a follow-up PR after small testing

PR #49027, opened 2026-02-12
Add hotkeys and actions for toggle light and dark theme

tommyming:
Mentioned in #47258 

Release Notes:

- Added hotkey options and actions for toggling light and dark theme.
- Add default keymap as `cmd/ctrl+k cmd/ctrl+shift+t`


SomeoneToIgnore (commented):
Overall, looking at the PR, I am not sure about the amount of actions to add: would it be better if we could have an action that toggles theme modes (first, if needed, switches into the mode that allows toggling first)?
If we go this way, it might also better be tested in `editor_tests.rs` instead of the current settings-related change?

SomeoneToIgnore:
From the user's perspective:
* there has to be an action that toggles between bright and dark, persistently
* ??? nothing else, right? 

So, if we can add an action that gets us into this switching loop out of whatever related settings state + have it tested, we're good.
The rest is details interesting to know at the impl review time, no more (but no less).

crates/settings/src/settings_store.rs
```diff
@@ -1518,6 +1519,60 @@ mod tests {
...
+                ..
+            } = &mut selection
+            {
+                *mode_to_update = mode;
+            }
+            theme.theme = Some(selection);
+        }
+    }
+
+    fn toggle_theme_mode(content: &mut SettingsContent, active_appearance: ThemeAppearanceMode) {
```
SomeoneToIgnore: What's going on here?
This method (and `apply_theme_mode`, respectively) is not used anywhere but in the tests?
Why do we need it then?

docs/src/themes.md
```diff
@@ -44,6 +44,33 @@ You can set the mode to `"dark"` or `"light"` to ignore the current system mode.
...
+  "theme": {
+    "mode": "light",
+    "light": "One Dark",
+    "dark": "One Dark"
+  }
+}
+```
+
+After that, toggling updates only `theme.mode`.
+If `light` and `dark` are the same theme, the first toggle may not produce a visible UI change until you set different values for `light` and `dark`.
```
SomeoneToIgnore, line 74: I might not understand something, but right above the docs say that we do set `light` and `dark` values to the same theme?

This is odd, we need to set the light one to `One Light` it seems?

SomeoneToIgnore (changes requested):
Great, we have some progress shaping, but now the changes are very odd: `crates/settings/src/settings_store.rs`'s new code is not used anywhere but its new tests.

So, we still miss the tests on the `crates/workspace/src/workspace.rs` level (you can check out `crates/editor/src/editor_tests.rs` or tests in the `workspace.rs` module on how those are created).

crates/workspace/src/workspace.rs
```diff
@@ -7036,6 +7003,27 @@ impl Workspace {
...
+            Some(theme::ThemeAppearanceMode::System) | None => match cx.theme().appearance() {
+                theme::Appearance::Light => theme::ThemeAppearanceMode::Dark,
+                theme::Appearance::Dark => theme::ThemeAppearanceMode::Light,
+            },
+        };
+
+        self.set_theme_mode(next_mode, cx);
+    }
+
+    fn set_theme_mode(&mut self, mode: theme::ThemeAppearanceMode, cx: &mut Context<Self>) {
```
SomeoneToIgnore: Let's inline this method, it's tiny and used once only.

docs/src/themes.md
tommyming: From the static mode, I have set it to "One Dark" as my current theme.  
Therefore, when switching to dynamic mode, it will keep the "One Dark" and fill up both light and dark values.

This is what I am trying to explain here.

crates/settings/src/settings_store.rs
tommyming: This method is not needed anymore, maybe I forget to remove it. Lemme remove it on the next commit

docs/src/themes.md
SomeoneToIgnore: Yeah, that seems somewhat odd, I'd expect `One Light` value for the `light` key added in this case — given that we have the control over the code anyway and can do that?

Either way, I guess I'll understand it better when build and test it myself, after the PR gets tidied up to remove all redundant changes.

tommyming: Ignore my prev comment, just deleted. Lemme have a look at how themes are constructed first, thanks for the reminder.

SomeoneToIgnore: One Dark and One Light are Zed's bundled themes that you cannot un-bundle, which should be pretty much visible from the code in https://github.com/zed-industries/zed/blob/465e2b5ffd9ac8553341bd3989ce960b4793f917/assets/themes/one/one.json

SomeoneToIgnore: It's an interesting idea whether we can try and use person's added themes instead, but I think that's not worth it at all, as it brings complexity for the only use case of the "system -> dynamic" transition, unless I miss something.

Given that we can describe this behavior in the docs, we may as well suggest that custom themes can be adjusted manually for each light/dark variant and then, the toggle will do a switch between those custom ones instead of the default ones.

tommyming: But the user may change to a custom theme even when in static mode. 

When the user toggles the shortcut, if the user is using a custom theme, instead of one dark/light, I think my current behavior is a better approach?

One Dark/Light is bundled by default, but people may use custom themes.

Lemme know what you think.

About the documentation of custom themes, I agree that we can add that. Lemme do this in the next commit.

SomeoneToIgnore: Definitely, using a custom theme is a better approach if you can detect which (dark/light) that variant is easily.
If not — falling back to bundled themes seems ok (esp. if documented properly), as it's done once only and then the users can later manually adjust their variants to whatever themes.

Using a single theme for both variants is a very odd and bad approach that breaks the whole idea of toggling.

tommyming: yeah, I am also thinking about that.

How about we just change it to:
when the user performs "static -> dynamic", it resets the theme to one dark/light, then let users to manually setup their light/dark theme?

and toggle will just works on other cases.

SomeoneToIgnore: Glad we're both arriving to my initial proposal now.

tommyming: nah it is just me keep looking for better handlings, and turns out this is the best one atm.

crates/workspace/src/workspace.rs
tommyming: should be updated, please review when you are free.

crates/settings/src/settings_store.rs
tommyming: updated tests and remove the useless methods.

docs/src/themes.md
tommyming: updated the whole approach with the mentioned conversation here, please review when free.

crates/settings/src/settings_store.rs
```diff
@@ -1669,6 +1670,118 @@ mod tests {
         );
     }
 
+    #[gpui::test]
+    fn test_toggle_theme_mode_in_dynamic_theme_selection(cx: &mut App) {
+        let mut store = SettingsStore::new(cx, &test_settings());
+
+        check_settings_update(
```
SomeoneToIgnore: What are these tests checking?
That a `|settings| ..` closure's changes are persisted?

Looks redundant, I'd expect that to be covered already, and even if not, what's the point?
This PR brings the action, and this is what we need to test.

The new functionality is not tested still, as we miss the editor tests, despite me bringing this up multiple times, e.g in https://github.com/zed-industries/zed/pull/49027#pullrequestreview-3860376677

crates/zed_actions/src/lib.rs
```diff
@@ -325,6 +325,12 @@ pub mod feedback {
     );
 }
 
+pub mod theme_mode {
+    use gpui::actions;
+
+    actions!(theme_mode, [Toggle]);
```
SomeoneToIgnore: This should be more generic in case we want to add more theme-related actions later.
Let's have a `theme` namespace and `ToggleMode` instead?

SomeoneToIgnore (commented):
Nice, we're getting there.

Let's finally write a proper test that uses the editor's action, and we're good to merge.

crates/settings/src/settings_store.rs
tommyming: **About Editor Tests**

Apologies, I definitely missed that.

For `editior_tests.rs` I should be writing tests about if the theme changes, calling the `update_editor` method, to make sure themes are changing correctly?

---

**About settings test**

These 2 test aims to check if the theme column inside settings.json changed accordingly, after the action is triggered.
The 1st test checks if theme flipping (light <-> dark) works when action is triggered.
The 2nd test checks if static -> dynamic works.
In this case, I guess they are not redundant...?

The only problem of these 2 test is the existance.
At this point I think they are optional.

tommyming: but yeah you are right, I should do them inside `editor_tests.rs` instead of `settings_store.rs`, since they will affect editor layout.

SomeoneToIgnore: No, the approach is actually totally different and more like "integration testing" in this regard: the idea is to start from firing off the action that switches the themes and ensure all side effects are taking place.

The way I imagine this could be structured is the same initialization each of those editor tests do: create a fake worktree, open a file in the editor + ensure the theme-related settings in the beginning are something.

Then, dispatch the new command to toggle the theme and ensure that the settings change and, maybe, the theme can be asserted too, as there's `cx.theme()` at least?

To note, this may be placed anywhere else not `editor_tests` necessarily, it's just the approach used there that's the same.

tommyming: Thanks for the guide, I might need to take a bit of time to build the tests, lemme update you next week.

crates/editor/src/editor_tests.rs
```diff
@@ -19330,6 +19330,107 @@ async fn test_language_server_restart_due_to_settings_change(cx: &mut TestAppCon
...
+    let settings_text = SettingsStore::load_settings(&settings_fs)
+        .await
+        .expect("failed to load persisted theme settings");
+    workspace.update_in(cx, |_workspace, _window, cx| {
+        SettingsStore::update_global(cx, |store, cx| {
+            store
+                .set_user_settings(&settings_text, cx)
+                .expect("failed to reload persisted theme settings");
+        });
+        GlobalTheme::reload_theme(cx);
```
SomeoneToIgnore: This does not look right?

Toggling a theme with the `ToggleMode` action should trigger the mechanisms that reload everything, yet here we cannot test it because manually force the reload in the test.

tommyming: oopsie I accdentally pushed unreviewed tests in this commit, my bad.
I have used a bit of AI assisting me here for the test, I am still checking the code.

Lemme update this, and make sure the tests passes on local first.

SomeoneToIgnore: Ok, and what's up with the rest of the manipulations like setting the user settings?
Why cannot just the theme be toggled and the state asserted instead???

tommyming: Suppose the flow of the action is:
user triggers the action -> updates the theme on `settings.json` -> `SettingsStore` detects the changes, updates the internal state -> observe changes in `SettingStore` and update the theme on display, right?

My understanding is the test should test the whole flow, thats why I have the "manuplicate settings" part to the testing method.

tommyming: But yes, if just testing theme changes, I can just toggle the theme, observe the state asserted is correct or not. 

The case is, the `toggle_theme_mode` method is calling `update_settings_file` method, which means I need to, first, create a fake settings file, update the settings file, and check if the theme of the workspace is changed?

I might missed something, so please feel free to let me know, just wanna clarify things.

SomeoneToIgnore: I think it's very important to concentrate and formulate what are you testing.

The idea is to test, however close possible, user's interactions with the new feature or whatever the code adjustments were relevant to.

Did we change or add the way settings.json changes are propagated into the theme?
Not really — that's why manually settings changes and checking something is somewhat odd, that is supposed to work and that is not changed in the PR — that includes the current test and the ones from `crates/settings/src/settings_store.rs`

What did we change?
We've added an action that toggles the theme variant and, in the special case of the `system` variant, it also writes extra settings.json

So, we have to toggle the settings using that action and ensure that settings.json change expectedly for each case + that in-memory theme changes in this case.
With `set_user_settings` done after the toggle, we cannot test this, as cannot ensure that __the action itself__ changed the state — it can be the `set_user_settings` that did that instead (which was not changed and was not intended to be tested).

tommyming: I have to admit I am not good at capturing the scope of writing tests for this PR, but I wil try my best. Maybe it is just me overthinking on the whole time.

---

> So, we have to toggle the settings using that action and ensure that settings.json change expectedly for each case + that in-memory theme changes in this case.

Yes I agree on this, and this is my intention from the very beginning when crafting those tests (include this one).

> as cannot ensure that the action itself changed the state — it can be the set_user_settings that did that instead (which was not changed and was not intended to be tested).

So the tests should be, simply:
make a fake `settings.json`, assign to the workspace, call `toggle_theme_mode`, check `settings,json` has the asserted theme or not, make sure is `toggle_theme_mode` changing the json.
And probably repeat the flow for all the possible cases?

SomeoneToIgnore: Yes, pretty much the way any user will interact with the new feature.

And also check the editor's in-memory state (`cx.theme().`?) that it has a correct value.

tommyming: good! at least I am on the right direction.

for in-memory state, probably is `cx.theme()` but will double check.

crates/editor/src/editor_tests.rs
```diff
@@ -19674,6 +19674,7 @@ async fn test_language_server_restart_due_to_settings_change(cx: &mut TestAppCon
     );
 }
 
+#[gpui::test]
```
SomeoneToIgnore: One small thing left to remove.

crates/workspace/src/workspace.rs
```diff
@@ -13549,6 +13568,89 @@ mod tests {
...
+        });
+        cx.executor().advance_clock(Duration::from_millis(200));
+        cx.run_until_parked();
+
+        // Verify the first toggle persisted the expected split theme
+        // configuration plus the explicit system mode marker.
+        let settings_text = SettingsStore::load_settings(&settings_fs).await.unwrap();
+        assert!(settings_text.contains(r#""mode": "system""#));
+        assert!(settings_text.contains(r#""light": "One Light""#));
+        assert!(settings_text.contains(r#""dark": "One Dark""#));
```
SomeoneToIgnore: How big of a snippet is there, can we do `assert_eq!` normally instead of all this?

Pretty much every new `assert!` is bad in this regards.

crates/workspace/src/workspace.rs
```diff
@@ -13549,6 +13568,89 @@ mod tests {
...
+
+        // Toggle again. This should leave the split light/dark themes in
+        // place while clearing the explicit system-mode persistence.
+        workspace.update_in(cx, |workspace, window, cx| {
+            workspace.toggle_theme_mode(&ToggleMode, window, cx);
+        });
+        cx.executor().advance_clock(Duration::from_millis(200));
+        cx.run_until_parked();
+
+        // Verify the system mode entry was removed, while the light/dark
```
SomeoneToIgnore: I find it suspicious that just toggling does not only add variants, but changes the json multiple times, differently, but hard to tell what's up without seeing the proper JSONs.

crates/workspace/src/workspace.rs
```diff
@@ -13549,6 +13568,112 @@ mod tests {
...
+        // configuration plus the explicit system mode marker, and that
+        // repeated reads through the same test context report the same
+        // active theme.
+        let settings_text = SettingsStore::load_settings(&settings_fs).await.unwrap();
+        assert!(settings_text.contains(r#""mode": "system""#));
+        assert!(settings_text.contains(r#""light": "One Light""#));
+        assert!(settings_text.contains(r#""dark": "One Dark""#));
+        let theme_name =
+            workspace.update_in(cx, |_workspace, _window, cx| cx.theme().name.to_string());
+        let repeated_theme_name =
```
SomeoneToIgnore: What is the point of loading the very same data twice and assert its equality???

SomeoneToIgnore (changes requested):
We're getting there, but the test code is too LLM-esque and needs improvement.

crates/workspace/src/workspace.rs
tommyming: I might need to go deeper to see what is going on to the JSONs, currently the toggle method just call `update_settings_file` in `workspace.rs:7175`

crates/workspace/src/workspace.rs
tommyming: I am cleaning the `settings_text` up, seems repeating, and checking that part multiple times is useless.

crates/workspace/src/workspace.rs
tommyming: apologies, cleaning the code in my computer but got personal stuff banged up, still in progress.

SomeoneToIgnore (approved):
I see, so you cannot really verify the in-memory state then, sorry for confusing, and thank you for the context.
That does not seem good and we'd better keep this listener in tests, but for now let's move on, as indeed this works when tested manually, and the json values are tested now.

Thank you a lot for bearing to this point!

tommyming:
@SomeoneToIgnore Thank you too, I do learn a lot during a whole process.

PR #49031, opened 2026-02-12
languages: Fix C/C++ preprocessor directive highlight regression

chenwuji2000-cyber:
## Summary

- PR #48109 changed the capture name for C/C++ preprocessor directives from `@keyword.directive` to `@preproc`. While semantically correct, the builtin themes had `preproc` defined with colors nearly indistinguishable from plain text (e.g. One Dark `#dce0e5ff`, Ayu Dark `#bfbdb6ff`), making `#include`, `#define`, etc. appear unhighlighted.
- This PR updates the `preproc` color in all builtin themes (and the fallback theme) to match their respective `keyword` color, restoring visible highlighting for preprocessor directives.

Fixes #49024

## Side effects

- Go uses `@preproc` for `//go:` and `// +build` compiler directives. These will also change from the previous muted gray to the keyword color. This is arguably an improvement — compiler directives are special constructs that benefit from visible highlighting, consistent with how other editors (CLion, VS Code) handle them.

## Test plan

- [x] `cargo test -p language highlight_map` passes
- [x] Open a C/C++ file and verify `#include`, `#define`, `#ifdef`, etc. are highlighted with the keyword color
- [x] Verify across multiple builtin themes (One Dark, Ayu Dark, Gruvbox Dark, etc.)
- [x] Open a Go file and verify `//go:` directives are highlighted reasonably

Release Notes:

- Fixed C/C++ preprocessor directives (`#include`, `#define`, etc.) appearing unhighlighted in builtin themes.


MrSubidubi (changes requested):
I think we should instead fix the highlights for the builtin themes here. Eventually (hopefully soon) we can add a fallback if `preproc` is not defined to `keyword.preproc` as defined here, but for now I think just fixing builtin themes is the better path. What do you think?

MrSubidubi (approved):
Sorry for the delay, thanks!

SomeoneToIgnore (approved):

PR #49041, opened 2026-02-12
Add small alloc and lookup optimizations

dybucc:
A vector was being instantiated when the callsite only required an iterable.

Another part of the code was performing multiple `contains()` lookups on a vector, and now it does it on a hashed set. This change has required some extra modifications across the codebase, but the affected sites are minimal and have been adjusted without major issues.

The changes include some `Hash` derived implementations, which were proposed in the original `lsp-types` in [[1]], and maybe could be merged into Zed's fork. I went ahead and used a newtype with a custom `Hash` implementation that simply called on the structure's public members' implementations of `Hash`.

The next change includes the removal of a check of the request capabilities after having already checked the same thing in the call to `to_lsp_params_or_response()` right before. The result of the `match` expression should already have returned a `Task::ready()` if the above mentioned function failed in performing the check that was later repeated and now removed.

Finally, in the `edits_from_lsp()` method, stable sorting was being performed when only unstable sorting would suffice. The method can only sort with respect to the key data, and not the satellite data, as the latter are the literal strings of the edit. It matters not which one of a sequence of overlapping edits (with same ranges that thus resolve the edits for equivalence) should come before the other.

[1]: https://github.com/gluon-lang/lsp-types/pull/295/changes#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R540

- [ ] Tests or screenshots needed?
- [ ] Code Reviewed
- [ ] Manual QA

Release Notes:

- Removed a vector allocation where the callsite only required an iterable.
- Improved multiple lookup operations when deserializing LSP edit operations.
- Removed a double-check of capabilities after requesting and thus determining LSP capabilities.
- Replaced stable sorting with unstable sorting of edits returned by the LSP.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -15203,7 +15203,7 @@ impl Editor {
     pub fn select_all(&mut self, _: &SelectAll, window: &mut Window, cx: &mut Context<Self>) {
         self.hide_mouse_cursor(HideMouseCursorOrigin::MovementAction, cx);
         self.change_selections(SelectionEffects::no_scroll(), window, cx, |s| {
-            s.select_ranges(vec![Anchor::min()..Anchor::max()]);
+            s.select_ranges([Anchor::min()..Anchor::max()]);
         });
     }
 
--- a/crates/project/src/lsp_command.rs
+++ b/crates/project/src/lsp_command.rs
@@ -533,7 +533,7 @@ impl LspCommand for PerformRename {
             .rename_provider
             .is_some_and(|capability| match capability {
                 OneOf::Left(enabled) => enabled,
-                OneOf::Right(_options) => true,
+                OneOf::Right(_) => true,
             })
     }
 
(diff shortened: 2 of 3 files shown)


dybucc:
> @dybucc thanks for this! Do you have benchmarks to show how much faster this is? The simpler changes (like &[] vs Vec) seem fine, but implementing a custom Hash function for the edits doesn't pass the sniff test for "complexity is worth it".
> 
> Leaning towards closing this for now, but open to merging if it's demonstrably significantly faster (though even then with the caveat that we should derive Hash and not implement it)

None whatsoever. I was looking into solving another issue as I read through the code and realized there was a possibility for a (possible, but not sure) performance improvement.

Would you mind keeping the PR open or making it a draft as I look into benchmarking?

If it turns out to not be worth it, I'll just go ahead and ammend the commit without that particular change, in case you do want to merge the other changes.

ConradIrwin:
https://github.com/zed-industries/zed/pull/49041/changes#diff-5b9d51b75b74266b5c651ef3c6988fe2e62d1680ac138d806cdaa75900bc8b5fL4993-L4995

Looks like some other code got removed erroneously

PR #49065, opened 2026-02-12
lsp: Add clangd readonly token modifier to semantic token rules to highlight constant variables as constant

ozacod:
Clangd uses the "readonly" token modifier instead of "constant". Therefore, "readonly" should be mapped to highlight constant variables as constants.

Before:
before

After:
after

- [x] Code Reviewed
- [x] Manual QA

Release Notes:

- Added clangd readonly token modifier to semantic token rules to highlight constant variables as constant.

### DIFF
--- a/crates/languages/src/cpp.rs
+++ b/crates/languages/src/cpp.rs
@@ -1,3 +1,15 @@
+use settings::SemanticTokenRules;
+
+use crate::LanguageDir;
+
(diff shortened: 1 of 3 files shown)


ozacod:
> Sorry for the late reply, for builtin languages we already have that, you can take a look at how rust does it, it has its own json rules file. So we can do the same clangd here

I am planning to open PR to add support for clangd, closing this for now.

Veykril (approved):
Thanks!

PR #49092, opened 2026-02-13
lsp: Sanitize newlines in document and workspace symbol names

aviatesk:
LSP servers may return symbol names containing newlines. Since outline panel, breadcrumbs, and project search modal all expect single-line items, collapse newlines before display.

- Document symbols: collapse newlines to spaces in name
- Workspace symbols: collapse newlines with `↵ ` separator for name and container_name, preserving visual hint of original structure

Closes #ISSUE

Release Notes:

- N/A

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -14276,14 +14276,10 @@ async fn populate_labels_for_symbols(
     }
 }
 
-fn collapse_newlines(text: &str, separator: &str) -> String {
-    if !text.contains('\n') {
-        return text.to_string();
-    }
-    text.split('\n')
+pub(crate) fn collapse_newlines(text: &str, separator: &str) -> String {
+    text.lines()
         .map(|line| line.trim())
         .filter(|line| !line.is_empty())
-        .collect::<Vec<_>>()
         .join(separator)
 }
 
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -7668,9 +7668,10 @@ impl LspStore {
                                         source_worktree_id,
                                         path,
                                         kind: symbol_kind,
-                                        name: symbol_name,
+                                        name: collapse_newlines(&symbol_name, "↵ "),
                                         range: range_from_lsp(symbol_location.range),
-                                        container_name,
+                                        container_name: container_name
+                                            .map(|c| collapse_newlines(&c, "↵ ")),
                                     })
                                 },
                             )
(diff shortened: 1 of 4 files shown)


crates/project/src/lsp_store/document_symbols.rs
```diff
@@ -454,4 +450,44 @@ mod tests {
         flatten_document_symbols(&symbols, &snapshot, 0, &mut items);
         assert!(items.is_empty());
     }
+
+    #[test]
+    fn test_collapse_newlines() {
```
SomeoneToIgnore: Can we not test some private function, but rather add a few examples right into LSP response of any of the outline/outline panel tests to see the whole pipeline passing through and sanitizing those LSP responses?

crates/project/src/lsp_store.rs
```diff
@@ -14275,6 +14276,17 @@ async fn populate_labels_for_symbols(
...
 }
 
+fn collapse_newlines(text: &str, separator: &str) -> String {
+    if !text.contains('\n') {
+        return text.to_string();
+    }
+    text.split('\n')
+        .map(|line| line.trim())
+        .filter(|line| !line.is_empty())
+        .collect::<Vec<_>>()
```
SomeoneToIgnore: We're supposed to have `Itertools` around, which can skip this `collect` and do `join` directly on iterators, let's do that instead.

crates/project/src/lsp_store.rs
```diff
@@ -14275,6 +14276,17 @@ async fn populate_labels_for_symbols(
     }
 }
 
+fn collapse_newlines(text: &str, separator: &str) -> String {
+    if !text.contains('\n') {
+        return text.to_string();
+    }
+    text.split('\n')
```
SomeoneToIgnore: We have Windows, so this is not correct, `.lines()` is the way to go.
The `if` above it is not needed that much as if we do not have a newline, the `.lines()` split will give us an iterator with a single element.

SomeoneToIgnore (commented):
Great catch, thank you.

Let's start unifying this with the rest of the LSP sanitation attempts, specifically https://github.com/zed-industries/zed/blob/abaabce3b701406643210e173f19dcfdab0a7af6/crates/project/src/lsp_command.rs#L4802-L4812

SomeoneToIgnore (approved):
Thank you!

aviatesk:
> Oh, but one test for the folds fails.

Ops. Fixed it.

PR #49112, opened 2026-02-13
search: Fix buffer/project empty search query with triple click line selection

ozacod:
Triple-clicking is typically used to select a line and it includes "\n" or "\r\n" characters with the line as well. If we triple-click and press "cmd+f" or "cmd+shift+f" it creates a new empty string for search query. This PR fixes this behavior.

Before:
https://github.com/user-attachments/assets/ab63d908-5059-4173-8321-dc656f376139

After:
https://github.com/user-attachments/assets/16b5c77c-972b-4789-a6d4-7c6e248d12fb

Release Notes:

- Fixed buffer/project empty search query with triple click line selection


crates/editor/src/items.rs
```diff
@@ -1610,10 +1610,11 @@ impl SearchableItem for Editor {
                 let text: String = buffer_snapshot
                     .text_for_range(selection.start..selection.end)
                     .collect();
-                if text.contains('\n') {
+                let text = text.trim_end_matches(&['\n', '\r'][..]);
```
SomeoneToIgnore: Cannot this all be just `text.trim().is_empty()` check?

SomeoneToIgnore (commented):
Nice spot, one question before we merge.

SomeoneToIgnore (changes requested):
To think of it more, we can add a test into `crates/search/src/project_search.rs` for this, cannot we?

Triple click can be emulated with 
```rs
editor.select(
    SelectPhase::Extend {
        position,
        click_count: 3,
    },
    window,
    cx,
);
```

crates/editor/src/items.rs
ozacod: fixed. did not remove '\n', '\r' so that double-triple line selection will be disabled as before.

ozacod:
#50783 introduced multiline search capability, therefore I changed the code slightly and added tests.

SomeoneToIgnore (commented):
Thank you for the tests, I have tried a few things with it and also tried more editors that have a similar selection suggest feature (VSCode, Intellij, Fleet) and they all keep the trailing whitespace when those are selected (e.g. on triple click).
The ones that do not auto fill queries (by default, at least, Sublime Text 2 and XCode) preserve the trailing whitespaces when the triple click selection result is copied.

Zed seems to have improved its query input now, and shows the triple click selection result this way now:

image

---------------

All that makes me wonder what to do here — does not seem good to have this change as a default behavior for sure, and if we want to add it, would be good to see which editors allow doing the same.

PR #49114, opened 2026-02-13
editor: Fix multi-cursor editing breaking indentation

### LINKED ISSUE #21334: Multi-line edit is wrecking havoc with indention
### Check for existing issues

- [X] Completed

When using multi-line edit ([Alt]+[Shift]+[arrow keys] to select where and which line) and inserting a char into indented space auto indention is executed which wreck havoc with indention already defined. If multi-line cursor is released there is no way later to multi-line delete the change. This is extremely important in cases of indention sensitive formats like Python or YAML. Example below shows the issue (I know I can use [Ctrl]+[/] for toggle comment and it works but commenting is not the only case when multiple lines are edited and I have selected this example just for clarity.) 

Example is shown on YAML file.

Steps shown below:
1. Create multi-line cursor at the beginning of multiple lines
2. Insert [#]
3. Cursor is released
4. Same as step 1
5. Deletion of one character [Del]
It removed only some of the "#" and Indention is broken even in places where they were removed
Indention Broken

Now even if "#" is removed manually deleting just this one character in each line the indention is broken and differs from the initial one as shown below.
After manual removal

### Environment

System Specs:
Zed: v0.163.2 (Zed)
OS: Linux X11 rocky 9.5
Memory: 125.5 GiB
Architecture: x86_64
GPU: NVIDIA RTX 5000 Ada Generation Laptop GPU || NVIDIA || 565.57.01

Zed configuration (~/.config/zed/settings.json):
{
  "base_keymap": "Atom",
  "telemetry": {
    "diagnostics": false,
    "metrics": false
  },
  "ui_font_size": 16,
  "buffer_font_size": 16,
  "theme": {
    "mode": "system",
    "light": "Solarized Dark",
    "dark": "One Dark",
    "tabWidth": 2,
    "useTabs": false
  },
  "project_panel": {
    "dock": "left",
    "auto_fold_dirs": false
  },
  "languages": {
    "YAML": {
      "prettier": {
        "bracketSpacing": false,
        "auto_indent_on_paste": false,
        "set auto_indent_using_last_non_empty_line": false
      }
    }
  }
}

_No response_

```

```
</details>

laibe:
Closes #21334 

- [x] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) -> no UI changes

When using multi-cursor to insert characters at the beginning of indented lines, autoindent would re-evaluate each line's indentation based on the combined post-edit state, causing adjacent edits to interfere with each other's indent suggestions and wreaking havoc on indentation—especially in indent-sensitive formats like YAML and Python.

Release Notes:

Disable `AutoindentMode::EachLine` when inserting non-newline text with multiple cursors. Single-cursor auto-outdent is preserved. When multiple cursors are on different rows, skip suggested_indents and insert a uniform tab instead of jumping each cursor to a language-computed indent level. This makes Tab behave predictably with multi-cursor, preserving relative indentation.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -4915,11 +4915,16 @@ impl Editor {
             let initial_buffer_versions =
                 jsx_tag_auto_close::construct_initial_buffer_versions_map(this, &edits, cx);
 
+            let autoindent_mode = if !text.contains('\n') && edits.len() > 1 {
+                None
+            } else {
+                this.autoindent_mode.clone()
+            };
             this.buffer.update(cx, |buffer, cx| {
                 if has_adjacent_edits {
-                    buffer.edit_non_coalesce(edits, this.autoindent_mode.clone(), cx);
+                    buffer.edit_non_coalesce(edits, autoindent_mode.clone(), cx);
                 } else {
-                    buffer.edit(edits, this.autoindent_mode.clone(), cx);
+                    buffer.edit(edits, autoindent_mode, cx);
                 }
             });
             for (buffer, edits) in linked_edits {
@@ -10843,8 +10848,24 @@ impl Editor {
         let mut selections = self.selections.all_adjusted(&self.display_snapshot(cx));
         let buffer = self.buffer.read(cx);
         let snapshot = buffer.snapshot(cx);
-        let rows_iter = selections.iter().map(|s| s.head().row);
-        let suggested_indents = snapshot.suggested_indents(rows_iter, cx);
+
+        // When there are multiple empty cursors on different rows, skip
+        // suggested-indent logic so Tab uniformly adds one indent level
+        // instead of jumping each cursor to a language-suggested position.
+        let has_multiple_cursor_rows = {
+            let mut rows = selections
+                .iter()
+                .filter(|s| s.is_empty())
+                .map(|s| s.head().row);
+            let first = rows.next();
+            first.is_some() && rows.any(|row| Some(row) != first)
+        };
+        let suggested_indents = if has_multiple_cursor_rows {
+            BTreeMap::new()
+        } else {
+            let rows_iter = selections.iter().map(|s| s.head().row);
+            snapshot.suggested_indents(rows_iter, cx)
+        };
 
         let has_some_cursor_in_whitespace = selections
             .iter()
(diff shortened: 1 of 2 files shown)


laibe:
@maxdeviant could you please kick of the test workflow?

crates/editor/src/editor_tests.rs
```diff
@@ -4286,11 +4276,62 @@ async fn test_indent_yaml_non_comments_with_multiple_cursors(cx: &mut TestAppCon
...
+        r#"    ˇingress:
+    ˇ  api:
+    ˇ    enabled: false
+    ˇ    pathType: Prefix
+"#,
+    );
+}
+
+#[gpui::test]
+async fn test_multicursor_input_does_not_break_yaml_indentation(cx: &mut TestAppContext) {
```
SomeoneToIgnore, line 4289: The test passes on current `main`, so we do not test the scenario from the issue and the non-tab-related code from the PR.
It seems important to test this logic and capture some specific yaml-related benefit of the PR, so let's rework this test.

Yaml seems to break on `main` for me when from

Image

I append any symbol, e.g. `#` comment:

Image

or `2` as a random symbol:

Image

and the branch fixes it, so the test should be hopefully simple to write?

crates/editor/src/editor.rs
```diff
@@ -10843,8 +10848,24 @@ impl Editor {
         let mut selections = self.selections.all_adjusted(&self.display_snapshot(cx));
         let buffer = self.buffer.read(cx);
         let snapshot = buffer.snapshot(cx);
-        let rows_iter = selections.iter().map(|s| s.head().row);
-        let suggested_indents = snapshot.suggested_indents(rows_iter, cx);
+
+        // When there are multiple empty cursors on different rows, skip
+        // suggested-indent logic so Tab uniformly adds one indent level
+        // instead of jumping each cursor to a language-suggested position.
+        let has_multiple_cursor_rows = {
```
SomeoneToIgnore, line 10855: Is this the right place to do the change?
We have `backtab`, also newline insertion (above and below) may trigger reformatting — should any of this ones be covered?

Maybe, something else, e.g. are other users of `suggested_indents`?

SomeoneToIgnore (changes requested):
Sorry for the late review.

Seems that there are 2 different mechanisms adjusted and tested at once, we need to be a bit more vigilant and systematic to capture, fix and test all related code branches, left a few comments in that direction.

SomeoneToIgnore:
Thank you for the idea, this seems like a good direction to go on with.
For now, given the activity history, will close the PR as stale to keep the PR queue cleaner.

PR #49120, opened 2026-02-13
go: split definition and call highlights for functions, methods, and types

tvi:
This allows definitions to use a different highlight than function calls.

Release Notes:

- go: Add definition highlights for functions, methods, and types

### DIFF
--- a/crates/languages/src/go/highlights.scm
+++ b/crates/languages/src/go/highlights.scm
@@ -1,6 +1,9 @@
 (identifier) @variable
 
 (type_identifier) @type
+(type_spec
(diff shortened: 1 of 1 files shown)


ConradIrwin:
Thanks! Can you please share screenshots of before and after in the default theme?

tvi:
Hi @ConradIrwin ; I switched it from `.defintion` to make calls `.call` to be more consistent with python highlights.

Default themes only specify `"function":{` so nothing changes for any of those 3 themes. But I am attaching screenshots for one dark.

Original:
original

Updated:
updated

PR #49121, opened 2026-02-13
cloud_llm_client: Add StreamEnded and Unknown variants to CompletionRequestStatus

tomhoule:
Add StreamEnded variant so the client can distinguish between a stream that the cloud ran to completion versus one that was interrupted (see CLO-258). **That logic is to be added in a follow up PR**.

Add an Unknown fallback with #[serde(other)] for forward-compatible deserialization of future variants.

The client advertises support via a new
x-zed-client-supports-stream-ended-request-completion-status header. The server will only send the new variant if that header is passed. Both StreamEnded and Unknown are silently ignored at the event mapping layer (from_completion_request_status returns Ok(None)).

Part of CLO-264 and CLO-266; cloud-side changes to follow.

Release Notes:

- N/A

### DIFF
--- a/crates/language_model/src/language_model.rs
+++ b/crates/language_model/src/language_model.rs
@@ -213,7 +213,7 @@ pub enum LanguageModelCompletionError {
         error: serde_json::Error,
     },
 
-    #[error("stream from {provider} over cloud ended unexpectedly")]
+    #[error("stream from {provider} ended unexpectedly")]
     StreamEndedUnexpectedly { provider: LanguageModelProviderName },
 
     // TODO: Ideally this would be removed in favor of having a comprehensive list of errors.
--- a/crates/language_model/src/language_model.rs
+++ b/crates/language_model/src/language_model.rs
@@ -104,12 +104,13 @@ impl LanguageModelCompletionEvent {
     pub fn from_completion_request_status(
         status: CompletionRequestStatus,
         upstream_provider: LanguageModelProviderName,
-    ) -> Result<Self, LanguageModelCompletionError> {
+    ) -> Result<Option<Self>, LanguageModelCompletionError> {
         match status {
             CompletionRequestStatus::Queued { position } => {
-                Ok(LanguageModelCompletionEvent::Queued { position })
+                Ok(Some(LanguageModelCompletionEvent::Queued { position }))
             }
-            CompletionRequestStatus::Started => Ok(LanguageModelCompletionEvent::Started),
+            CompletionRequestStatus::Started => Ok(Some(LanguageModelCompletionEvent::Started)),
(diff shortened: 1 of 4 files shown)


crates/language_model/src/language_model.rs
```diff
@@ -104,12 +104,13 @@ impl LanguageModelCompletionEvent {
...
-    ) -> Result<Self, LanguageModelCompletionError> {
+    ) -> Result<Option<Self>, LanguageModelCompletionError> {
         match status {
             CompletionRequestStatus::Queued { position } => {
-                Ok(LanguageModelCompletionEvent::Queued { position })
+                Ok(Some(LanguageModelCompletionEvent::Queued { position }))
             }
-            CompletionRequestStatus::Started => Ok(LanguageModelCompletionEvent::Started),
+            CompletionRequestStatus::Started => Ok(Some(LanguageModelCompletionEvent::Started)),
+            CompletionRequestStatus::Unknown | CompletionRequestStatus::StreamEnded => Ok(None),
```
maxdeviant, line 113: Are we planning to do more later with the `StreamEnded` event? Or this the extent of the change?

tomhoule: Yes, we also have to report an error if the stream is closed without the cloud emitting a StreamEnded event first (that’s how we can detect if the stream was closed abnormally), I’m syncing with Bennet on that today.

tomhoule: Ok I looked into it and I think there’s a straightforward way to handle this. So I pushed that. This PR should be mergeable as is. We’ll still need a follow up PR to turn on enforcement of StreamEnded only once it’s supported in cloud (there’s a catch 22 here, we need the crate with the shared type from this PR before we can merge the cloud PR).

tomhoule: @bennetbo if you want to have a look. Just keeping track of whether we’ve received StreamEnded, and returning a known error. With a single retry.

tomhoule: I confirmed the behavior: one retry, and the error is reported.

image

crates/language_model/src/language_model.rs
```diff
@@ -212,6 +213,9 @@ pub enum LanguageModelCompletionError {
         error: serde_json::Error,
     },
 
+    #[error("stream from {provider} over cloud ended unexpectedly")]
```
maxdeviant: We shouldn't use "Cloud" in our public-facing messaging, as it's our internal name and users aren't expected to know what it is.

I think omitting that part is fine:

```suggestion
    #[error("stream from {provider} ended unexpectedly")]
```

maxdeviant (approved):

PR #49142, opened 2026-02-13
Add regression tests for ZED-4VS UTF-8 char boundary crash

morgankrey:
## Summary

Adds regression tests for the UTF-8 character boundary panic in `edit_prediction::parse_edits` (Sentry issue ZED-4VS), which was fixed in #48960.

## Context: Background Agent Crash Investigation Experiment

This PR is an output of testing the **Sentry crash → regression test** workflow as part of our testing of background agents. The goal is to validate whether AI agents can effectively:

1. Query Sentry for high-impact, fixable crashes
2. Follow a structured investigation prompt to identify root cause
3. Write meaningful regression tests (or fixes, if not already merged)

### What we did

1. **Queried Sentry** for `edit_prediction` crashes sorted by frequency
2. **Selected ZED-4VS** as a good candidate (clear panic message, reproducible pattern)
3. **Followed the crash investigation prompt** to:
   - Analyze the stack trace and identify crash site (`zeta1.rs:328`)
   - Identify root cause: `+1`/`-1` byte adjustments for newline skipping land inside multi-byte UTF-8 characters
   - Discover the fix was already merged in #48960
4. **Wrote regression tests** that exercise the exact failure mode

### Key learning

The investigation workflow successfully identified the root cause and produced actionable tests. Tests should exercise user-visible operations, not directly construct failure states. These tests use realistic model output patterns (multi-byte chars near markers) rather than manually crafting invalid byte indices.

### Tests added

- `test_parse_edits_with_multibyte_utf8` - generic multi-byte content in editable region
- `test_parse_edits_multibyte_before_end_marker` - multi-byte char immediately before end marker
- `test_parse_edits_multibyte_after_start_marker_no_newline` - multi-byte char after start marker without newline (the actual crash case)

## Verification

All 69 tests in edit_prediction pass.

Release Notes:

- N/A

### DIFF
--- a/crates/zeta_prompt/src/zeta_prompt.rs
+++ b/crates/zeta_prompt/src/zeta_prompt.rs
@@ -1677,4 +1677,25 @@ mod tests {
         let cleaned = zeta1::clean_zeta1_model_output(output).unwrap();
         assert_eq!(cleaned, "");
     }
+
+    #[test]
+    fn test_clean_zeta1_model_output_multibyte_utf8() {
+        let output = "<|editable_region_start|>\n┌───┐\n│box│\n└───┘\n<|editable_region_end|>";
+        let cleaned = zeta1::clean_zeta1_model_output(output).unwrap();
+        assert_eq!(cleaned, "┌───┐\n│box│\n└───┘");
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


eholk (approved):
This seems pretty reasonable to me. It adds coverage for a few more corner cases where we've seen related crashes.

PR #49156, opened 2026-02-14
recent_projects: Check for WSL paths in OpenFolderInWsl

### LINKED ISSUE #42842: Windows: "projects: open folder in wsl" doesn't work but is equivalent to other ways that do work
## This way doesn't work

- Ctrl Shift P
- projects: open folder in WSL
- select a folder
- "Invalid path": "Invalid path specified when try to open a folder inside WSL. Please note that
Zed currently does not support opening network share folders inside wsl."

## Other equivalent ways do work

- Left side bar: "Open Project"
- I then select a folder in wsl2.
- "`/path/to/folder` is inside a WSL filesystem, some features may not work unless
you open it with WSL remote. "Open in WSL". It works as I would expect.

or

- Ctrl Shift P
- projects: open wsl
- open folder... etc. that way works.

these are all equivalent (?) and should have the same result

Zed 0.212.6 
74924bc91865f111542ac6b1554231a0d3f66f89

-

maximxlss:
Closes #42842

Previously, OpenFolderInWsl action assumed a valid path is on the host and failed to open paths inside the WSL filesystem (accessed through UNC)

Now the action checks for a WSL path and opens it with the appropriate distro (skipping the choice modal)

This was manually confirmed to not regress as well as work with WSL paths.

Code was written by hand, with extra checking by OpenAI Codex.

Release Notes:

- Fixed paths inside the WSL filesystem failing to open when using "open folder in wsl"

### DIFF
--- a/crates/recent_projects/src/recent_projects.rs
+++ b/crates/recent_projects/src/recent_projects.rs
@@ -230,12 +230,41 @@ pub fn init(cx: &mut App) {
                 cx,
             );
 
+            let app_state = workspace.app_state().clone();
+            let window_handle = window.window_handle().downcast::<MultiWorkspace>();
+
             cx.spawn_in(window, async move |workspace, cx| {
                 use util::paths::SanitizedPath;
 
                 let Some(paths) = paths.await.log_err().flatten() else {
                     return;
                 };
+
+                let wsl_path = paths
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #49160, opened 2026-02-14
python: Prevent shell command injection in conda environment activation

cppcoffee:
Release Notes:

- Fixed prevent shell command injection in conda environment activation

The conda environment name was directly interpolated into the shell
command without proper escaping, which could allow command injection
if the environment name contained malicious shell metacharacters.


Veykril (approved):
Thanks!

PR #49185, opened 2026-02-14
gpui: Fix swapped '<' and '>' in `guess_ascii` for Linux

Albab-Hasan:
the guess_ascii fallback (used for non-latin keyboard layouts) had the shifted characters for comma (keycode 59) and period (keycode 60) swapped. shift+comma should produce '<' and shift+period should produce '>', not the other way around.

Release Notes:

- N/A


SomeoneToIgnore:
The reviews are our best effort and may drag behind the schedule, unfortunately.

This particular one now has conflicts as the entire file was overwritten in the PR I've mentioned above and now the same PR seems to have the codes made rights — so the fix itself was made already.

PR #49086, opened 2026-02-13
gpui: Fix thermal state notifications on intel macOS

marcocondrache:
Fixes https://github.com/zed-industries/zed/pull/45638#issuecomment-3893732034
Closes #49005

The issue was similar to the one we already had with the quit method https://github.com/zed-industries/zed/blob/8249ef56187b966c33f6667d0d3a35d88d8f2dc0/crates/gpui/src/platform/mac/platform.rs#L491

@notpeter Could you please test this branch to confirm it resolves the issue?

- [ ] Tests or screenshots needed?
- [x] Code Reviewed
- [ ] Manual QA

Release Notes:

- Fixed an issue where Zed would randomly crash on macOS intel

PR #49189, opened 2026-02-14
gpui: Defer thermal/keyboard state updates when app is borrowed

morgankrey:
Follow-up to #49187.

Instead of silently skipping updates when `try_borrow_mut` fails, this PR defers the update by spawning it on the foreground executor. This ensures the state change is eventually processed after the current borrow completes.

This addresses the feedback: "if we pull something out of the event queue asynchronously and the state we want to update is unavailable, either we should put it back in the queue or if that doesn't work, retry in a loop."

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/.rules
+++ b/crates/gpui/.rules
@@ -1,0 +1,9 @@
+# GPUI crate-specific rules
+#
+# This file is non-exhaustive. Check the root .rules for general guidelines.
+
+# Platform callbacks
+
+* Platform callbacks (e.g., `on_keyboard_layout_change`, `on_thermal_state_change`) can fire
+  asynchronously from macOS while `AppCell` is already borrowed. Defer work via
+  `ForegroundExecutor::spawn` rather than borrowing `AppCell` directly in the callback body.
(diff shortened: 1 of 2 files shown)


marcocondrache:
@notpeter sorry for the direct ping, could you please check this fixes the issue on your Intel machine? So I can eventually close the fix I've already had: https://github.com/zed-industries/zed/pull/49086

notpeter:
Update: new crash for the code which was introduced in this PR.
https://github.com/zed-industries/zed/blob/99a558f3689e6920eb3c3ad5e78854c7068c5d75/crates/gpui/src/app.rs#L757-L762

```
2026-02-15T12:52:23-05:00 ERROR [crashes] thread 'main' panicked at /Users/runner/work/zed/zed/crates/scheduler/src/executor.rs:388:13:
local task polled by a thread that didn't spawn it. Task spawned at crates/gpui/src/app.rs:757:26...
2026-02-15T12:52:23-05:00 ERROR [crashes] triggering a crash to generate a minidump...
```

PR #49210, opened 2026-02-15
explorer_command_injector: Avoid COM out-pointer overwrite in class factory exports

cppcoffee:
Release Notes:

- N/A


reflectronic (approved):
Thank you

PR #49232, opened 2026-02-15
editor: Autoscroll to initial selection on select all matches edit

Ben-Vollrath:
Closes #32894

## Summary
Adjust **Select All Matches** autoscroll behavior so that when a fit-based strategy can’t keep all selections on screen and we fall back to `newest`, we scroll to the **initial selection** that triggered `select_all_matches` instead of the very last one.

## Design decision
The most intuitive behavior is to keep the user oriented around the **initial selection**:

- **User is still viewing the initial selection** → no autoscroll
- **User has scrolled away** → autoscroll back to the initial selection (indicate that something happened)

This takes into account the feedback on [PR #42009](https://github.com/zed-industries/zed/pull/42009#pullrequestreview-3650051705). This implements a fixed version of the first idea.

## Implementation details
- Push the **initial selection** into `new_selections` **last**, so it receives the **highest selection id**.
- The `newest` fallback uses the highest id (see [`autoscroll.rs`](https://github.com/zed-industries/zed/blob/main/crates/editor/src/scroll/autoscroll.rs#L176)), so when fit fails and the logic falls back to `newest`, it scrolls to the **initial selection**, not the very last match.

Release Notes:

- Fixed an issue where editing selections after `editor: select all matches` would scroll to the last match


dinocosta (approved):
Thank you for suggesting this fix @Ben-Vollrath ! 🙂

I've pushed a smaller commit refactoring a small part of the code and update an existing test to actually catch the bug that is being fixed. Shipping! 🚀

PR #49236, opened 2026-02-15
gpui: Take advantage of unified memory on macOS

marcocondrache:
Third attempt to land this improvement: (#45577, #44273)

The previous PR didn’t work on Intel MacBooks because I made a wrong assumption about the unified memory check. `has_unified_memory` only tells us that the CPU and GPU share memory. It does not mean we’re running on an Apple GPU family.

Memoryless textures are only supported on Apple GPUs. Some Intel Macs report unified memory, but they don’t support memoryless textures, which is why the previous change failed there.

So instead of relying on unified memory, we now explicitly check that we’re running on an Apple GPU family before enabling memoryless textures.

Release Notes:

- Reduced memory usage on macOS


notpeter:
This worked for me on my Intel Macbook (16inch 2019) as long as I commented out the breakage introduced in https://github.com/zed-industries/zed/pull/49189

image

Anthony-Eid:
> This worked for me on my Intel Macbook (16inch 2019) as long as I commented out the breakage introduced in #49189
> 
> image

Is this still broken for you and are you able to use a preview version of Zed?

marcocondrache:
@Anthony-Eid sorry, I forgot to update `Cargo.lock`

PR #49243, opened 2026-02-16
collab: Compare API token in constant time to reduce timing leak risk

cppcoffee:
Release Notes:

- N/A


maxdeviant:
Thank you for the PR!

These comparisons have both been deleted in #49312 and #49314, so no further changes are necessary.

Issue #49181, opened 2026-02-14
Zed is crashing consistently on Mac Intel


arrrrny:
### Reproduction steps

Based on the logs, Zed is crashing due to a **Rust borrow checker panic** in the GPUI framework:

```
ERROR [crashes] thread '<unnamed>' panicked at crates/gpui/src/app.rs:749:39:
RefCell already borrowed...
```

This same crash occurred **3 times**:
- Feb 13, 21:48:41
- Feb 14, 17:48:23  
- Feb 14, 18:12:37

This is a **Zed bug** (not configuration-related). The error indicates a double-borrow issue in Zed's UI framework where code tries to borrow a `RefCell` while it's already borrowed.

### Zed version and system specs

Zed: v0.224.1+preview.154.76f187618b338834e6cc414b121d2228cf1fbe78 (Zed Preview) 
OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
2026-02-14T17:48:23+03:00 ERROR [crashes] thread '<unnamed>' panicked at crates/gpui/src/app.rs:749:39:
RefCell already borrowed...
2026-02-14T17:48:23+03:00 ERROR [crashes] triggering a crash to generate a minidump...
2026-02-14T17:58:09+03:00 INFO  [zed] ========== starting zed version 0.224.1+preview.154.76f187618b338834e6cc414b121d2228cf1fbe78, sha 76f1876 ==========
2026-02-14T17:58:09+03:00 INFO  [crashes] spawning crash handler process
2026-02-14T17:58:09+03:00 INFO  [crashes] connected to crash handler process after 100ms
2026-02-14T17:58:09+03:00 INFO  [crashes] crash handler registered
2026-02-14T17:58:10+03:00 INFO  [zed] Using git binary path: "/Applications/Zed Preview.app/Contents/MacOS/git"
2026-02-14T17:58:10+03:00 INFO  [extension_host] extensions updated. loading 10, reloading 0, unloading 0
2026-02-14T17:58:10+03:00 INFO  [auto_update] Auto Update: checking for updates
2026-02-14T17:58:10+03:00 WARN  [project::agent_server_store] Registry agent 'github-copilot' not found in ACP registry
2026-02-14T17:58:10+03:00 WARN  [project::agent_server_store] Registry agent 'factory-droid' not found in ACP registry
[...]

PR #49251, opened 2026-02-16
Revert "gpui: Defer thermal/keyboard state updates when app is borrowed"

marcocondrache:
Reverts #49189
Reverts #49187

#49189 introduces a panic (https://github.com/zed-industries/zed/pull/49189#issuecomment-3904914597, https://github.com/zed-industries/zed/issues/49181#issuecomment-3906639392)
#49187 It's wrong since it leads to missing updates

The original crash should be fixed with #49086

cc: @bennetbo 

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/.rules
+++ b/crates/gpui/.rules
@@ -1,9 +1,0 @@
-# GPUI crate-specific rules
-#
-# This file is non-exhaustive. Check the root .rules for general guidelines.
-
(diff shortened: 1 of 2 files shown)


marcocondrache:
@bennetbo thank you so much for handling this and sorry for the confusion it created!

PR #49261, opened 2026-02-16
Add automatic syntax highlighting for GeoJSON files

flother:
GeoJSON is a popular geospatial data interchange format, standardised as [RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946). Because all GeoJSON files are valid JSON files, they can be highlighted using Zed's existing JSON language support.

This change adds `geojson` as a recognised path suffix, so GeoJSON files automatically open as JSON and get the standard JSON syntax highlighting.

Before and after screenshots:

geojson

Release Notes:

- Added automatic syntax highlighting for GeoJSON files

### DIFF
--- a/crates/languages/src/json/config.toml
+++ b/crates/languages/src/json/config.toml
@@ -1,6 +1,6 @@
 name = "JSON"
 grammar = "json"
-path_suffixes = ["json", "flake.lock"]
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (approved):
Thank you.

PR #49262, opened 2026-02-16
ruby: Add Kanayago as an available language server

S-H-GAMELINKS:
Add Kanayago as an available Ruby language server in default settings. Disabled by default (prefixed with `!`), users can opt-in via their settings.

This corresponds to the extension-side support added in zed-extensions/ruby#224.

Release Notes:

- ruby: Added Kanayago as an available language server.


MrSubidubi (approved):
Thanks!

PR #49265, opened 2026-02-16
Fix language injections containing interpolations

jeffbrennan:
Release Notes:

-  Fixed injected language highlighting for strings containing interpolations

current             |  updated
:-------------------------:|:-------------------------:
 | 

<details>

<summary> test snippet  </summary>

```python
my_regex = r"[0-9]"
spark.sql(  # sql
    f"""
        SELECT
            col1,
            CASE WHEN col2 RLIKE '{my_regex}' THEN '1' ELSE '2' END col4,
            ROW_NUMBER() OVER(PARTITION BY col1, col2 ORDER BY col3) as rn
        FROM tbl
    """
)
```

</details>

## Summary
Strings with interpolations like `f"SELECT col1 FROM tbl WHERE col2 = '{my_var}'"` in Python produce multiple `string_content` nodes. Without grouping, each fragment gets its own injection layer which breaks syntax highlighting.

This PR adds an `@injection.host` capture to group `@injection.content` ranges by their host node. This was implemented by extending the existing `combined_injection_ranges` hash map with a node key per host. Added tests for this behavior covering simple strings, f-strings with single/multiple interpolations, and multi-string isolation.


osiewicz (approved):
Thanks!

kczx3:
Will similar changes need to be made for other language grammars?

PR #49282, opened 2026-02-16
extension_host: Allow extensions to define semantic highlighting rules

Veykril:
for their given language via a `semantic_token_rules.json` file

Release Notes:


ozacod:
> 1. **User settings** — rules from `semantic_token_rules` in your settings file.
> 2. **Default rules** — Zed's built-in rules for standard LSP token types.
> 3. **Extension rules** — rules from `semantic_token_rules.json` in extension language directories.
      
With the current ordering, extension author can't reliably style language tokens I think it defeats the purpose of allowing defining language-specific rules in themes. In my opinion, swapping 2 and 3 makes more sense.

PR #49283, opened 2026-02-16
git_ui: Fix tree view next selection out of bounds

### LINKED ISSUE #49259: Zed Editor Crashes When Navigating Between Files in Git Panel Tree View Using Keyboard
### Reproduction steps

Steps to Reproduce:

Open Zed editor
Open the git panel
Navigate to the tree view of files
Use keyboard navigation (arrow keys) to move between files, fold folders
Editor crashes

Zed 0.223.3 
59dfdea1697030fd39de6c92172ff192d013a984

0.223.3+stable.152.59dfdea1697030fd39de6c92172ff192d013a984

Apple m3 pro
OS: 15.7.4 (24G517)

### Attach Zed log file

```log

```

</details>

errmayank:
Closes #49259

Release Notes:

- This change ensures that when the last visible collapsed directory is selected, the selection remains on that directory.

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -1112,12 +1112,14 @@ impl GitPanel {
             return;
         };
 
-        if selected_entry == item_count - 1 {
-            return;
-        }
-
         let new_index = match &self.view_mode {
-            GitPanelViewMode::Flat => selected_entry.saturating_add(1),
+            GitPanelViewMode::Flat => {
+                if selected_entry >= item_count.saturating_sub(1) {
+                    return;
+                }
+
+                selected_entry.saturating_add(1)
+            }
             GitPanelViewMode::Tree(state) => {
                 let Some(current_logical_index) = state
                     .logical_indices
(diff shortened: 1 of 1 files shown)


crates/git_ui/src/git_panel.rs
```diff
@@ -1127,12 +1129,24 @@ impl GitPanel {
...
             }
         };
 
         if matches!(self.entries.get(new_index), Some(GitListEntry::Header(..))) {
-            self.selected_entry = Some(new_index.saturating_add(1));
+            let Some(new_selected_entry) = new_index.checked_add(1).filter(|ix| *ix < item_count)
+            else {
+                return;
+            };
+            self.selected_entry = Some(new_selected_entry);
```
kubkon: This check seems redundant:
* in `GitPanelViewMode::Flat`, we already check for this
* in `GitPanelViewMode::Tree`, next entry is bound by `logical_indices` precludes this case
Lemme know if I missed something though! Also, your test does not account for this which strengthens my belief that this check is redundant.

crates/git_ui/src/git_panel.rs
```diff
@@ -7211,6 +7225,127 @@ mod tests {
...
+                ".git": {},
+                "src": {
+                    "a": {
+                        "foo.rs": "fn foo() {}",
+                    },
+                    "b": {
+                        "bar.rs": "fn bar() {}",
+                        "baz.rs": "fn baz() {}",
+                    },
+                },
```
kubkon, line 7242: Looks good! I would suggest we strive to replicate @mchisolm0's repro with one twist - let's add a `GitHeader` entry into the mix, so something like this:

```
Tracked
bar/
  bar1.py
  bar2.py
foo/
  foo1.py
  foo2.py
Untracked
foobar.py
```

kubkon (changes requested):
Looks great! I've got a couple of nits though that hopefully are straightforward to address.

crates/git_ui/src/git_panel.rs
errmayank: Right, i've updated it now

kubkon (approved):

PR #49287, opened 2026-02-16
bedrock: Model streamlining and cleanup

5herlocked:
Release Notes:

- Improved Bedrock error messages: region-locked models ask the user to try a different region, rate limits and access errors are reported cleanly instead of as raw API responses
- Streamlined Bedrock model list to 39 curated models
- Fixed API errors when using non-tool models in agent threads

### DIFF
--- a/crates/bedrock/src/models.rs
+++ b/crates/bedrock/src/models.rs
@@ -45,123 +45,91 @@ pub struct BedrockModelCacheConfiguration {
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, EnumIter)]
 pub enum Model {
-    // Anthropic models (already included)
+    // Anthropic Claude 4+ models
+    #[serde(rename = "claude-haiku-4-5", alias = "claude-haiku-4-5-latest")]
+    ClaudeHaiku4_5,
     #[serde(rename = "claude-sonnet-4", alias = "claude-sonnet-4-latest")]
     ClaudeSonnet4,
-    #[serde(
-        rename = "claude-sonnet-4-thinking",
-        alias = "claude-sonnet-4-thinking-latest"
-    )]
+    #[serde(rename = "claude-sonnet-4-thinking", alias = "claude-sonnet-4-thinking-latest")]
     ClaudeSonnet4Thinking,
     #[default]
     #[serde(rename = "claude-sonnet-4-5", alias = "claude-sonnet-4-5-latest")]
     ClaudeSonnet4_5,
-    #[serde(
-        rename = "claude-sonnet-4-5-thinking",
-        alias = "claude-sonnet-4-5-thinking-latest"
-    )]
+    #[serde(rename = "claude-sonnet-4-5-thinking", alias = "claude-sonnet-4-5-thinking-latest")]
     ClaudeSonnet4_5Thinking,
-    #[serde(rename = "claude-opus-4", alias = "claude-opus-4-latest")]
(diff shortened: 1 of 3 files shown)


crates/bedrock/src/models.rs
```diff
@@ -89,79 +84,83 @@ pub enum Model {
...
+    Nova2Lite,
+
+    // OpenAI GPT OSS models
+    #[serde(rename = "gpt-oss-20b")]
+    GptOss20B,
+    #[serde(rename = "gpt-oss-120b")]
+    GptOss120B,
+
+    // MiniMax models
+    #[serde(rename = "minimax-m2")]
```
iggy, line 143: `aws bedrock list-foundation-models` lists `minimax.minimax-m2.1`, but [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) don't.

I expect minimax m2.5 will be available soon too.

crates/bedrock/src/models.rs
```diff
@@ -89,79 +84,83 @@ pub enum Model {
...
+    #[serde(rename = "qwen3-vl-235b")]
+    Qwen3VL235B,
+    #[serde(rename = "qwen3-235b")]
+    Qwen3_235B,
+    #[serde(rename = "qwen3-next-80b")]
+    Qwen3Next80B,
+    #[serde(rename = "qwen3-coder-30b")]
+    Qwen3Coder30B,
+    #[serde(rename = "qwen3-coder-next")]
+    Qwen3CoderNext,
```
iggy, line 122: This one doesn't show up in `list-foundation-models` or docs, but it is mentioned in a [recent post](https://aws.amazon.com/about-aws/whats-new/2026/02/amazon-bedrock-adds-support-six-open-weights-models/) about new models that mentions DeepSeek V3.2, MiniMax M2.1, GLM 4.7, GLM 4.7 Flash, Kimi K2.5, and Qwen3 Coder Next.

It does show up in the openapi compatible (bedrock mantle) model list endpoint.

Should the docs be updated to point people at using that to access other models that aren't explicitly handled in this provider?

crates/bedrock/src/models.rs
```diff
@@ -89,79 +84,83 @@ pub enum Model {
...
+    KimiK2_5,
+
+    // DeepSeek models
+    #[serde(rename = "deepseek-r1")]
+    DeepSeekR1,
+    #[serde(rename = "deepseek-v3")]
+    DeepSeekV3_1,
+    #[serde(rename = "deepseek-v3-2")]
+    DeepSeekV3_2,
+
```
iggy, line 159: Thoughts on adding GLM models? Another case of them showing up in the list-foundation-models output, but not docs.

benbrandt (approved):
thanks!

crates/bedrock/src/models.rs
5herlocked: List foundation models responses actually varies per region -- not to criticize too aggressively -- but the docs suck right now, we have several conversations already underway.

These are aggregated across `us-east-1` and `us-west-2`.

crates/bedrock/src/models.rs
5herlocked: agreed -- will keep an eye open for the new models.

crates/bedrock/src/models.rs
5herlocked: +100 i personally love the GLM models -- feel free to add those in through a PR / I'll get to it eventually.

crates/bedrock/src/models.rs
iggy: I was only looking at us-west-2

PR #49294, opened 2026-02-16
lsp: Add rust-analyzer semantic token modifier support

bbqsrc:
Register all 17 `rust-analyzer` custom semantic token modifiers and add default visual styling rules for Rust:

- `unsafe` → bold
- `mutable` → underline
- `consuming` → italic
- `deprecated` → strikethrough
- `keyword` + `controlFlow` → `keyword.control` (forward-compatible scope for themes that distinguish control-flow keywords)

The modifier registration in `SEMANTIC_TOKEN_MODIFIERS` tells rust-analyzer that Zed understands these modifiers, enabling richer token data.

An example of porting the [Pale Fire](https://github.com/matklad/pale-fire) theme to Zed:

image

Release Notes:

- Improved semantic token support for Rust

### DIFF
--- a/crates/languages/src/rust/semantic_token_rules.json
+++ b/crates/languages/src/rust/semantic_token_rules.json
@@ -1,5 +1,26 @@
 [
   {
+    "token_modifiers": ["unsafe"],
+    "font_weight": "bold"
+  },
+  {
+    "token_modifiers": ["mutable"],
+    "underline": true
+  },
+  {
+    "token_modifiers": ["consuming"],
+    "font_style": "italic"
+  },
+  {
+    "token_modifiers": ["deprecated"],
+    "strikethrough": true
+  },
+  {
+    "token_type": "keyword",
+    "token_modifiers": ["controlFlow"],
+    "style": ["keyword.control", "keyword"]
+  },
+  {
     "token_type": "angle",
     "style": ["punctuation.bracket"]
   },
--- a/crates/lsp/src/lsp.rs
+++ b/crates/lsp/src/lsp.rs
@@ -389,6 +389,22 @@ pub const SEMANTIC_TOKEN_MODIFIERS: &[SemanticTokenModifier] = &[
     // Language specific things below.
     // Rust
     SemanticTokenModifier::new("constant"),
+    SemanticTokenModifier::new("unsafe"),
+    SemanticTokenModifier::new("mutable"),
+    SemanticTokenModifier::new("consuming"),
+    SemanticTokenModifier::new("callable"),
+    SemanticTokenModifier::new("trait"),
+    SemanticTokenModifier::new("library"),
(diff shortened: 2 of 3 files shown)


crates/lsp/src/lsp.rs
```diff
@@ -389,6 +389,22 @@ pub const SEMANTIC_TOKEN_MODIFIERS: &[SemanticTokenModifier] = &[
...
+    SemanticTokenModifier::new("public"),
+    SemanticTokenModifier::new("reference"),
+    SemanticTokenModifier::new("controlFlow"),
+    SemanticTokenModifier::new("crateRoot"),
+    SemanticTokenModifier::new("injected"),
+    SemanticTokenModifier::new("intraDocLink"),
+    SemanticTokenModifier::new("associated"),
+    SemanticTokenModifier::new("attribute"),
+    SemanticTokenModifier::new("macro"),
+    SemanticTokenModifier::new("procMacro"),
```
Veykril, line 407: All of these should be unnecessary as we will register them when rust-analyzer informs us about them

crates/languages/src/rust/semantic_token_rules.json
```diff
@@ -1,4 +1,25 @@
...
+    "underline": true
+  },
+  {
+    "token_modifiers": ["consuming"],
+    "font_style": "italic"
+  },
+  {
+    "token_modifiers": ["deprecated"],
+    "strikethrough": true
+  },
```
Veykril, line 17: This should be a default rule, not rust specific

crates/languages/src/rust/semantic_token_rules.json
```diff
@@ -1,4 +1,25 @@
...
+    "font_weight": "bold"
+  },
+  {
+    "token_modifiers": ["mutable"],
+    "underline": true
+  },
+  {
+    "token_modifiers": ["consuming"],
+    "font_style": "italic"
+  },
```
Veykril, line 13: I don't think this one pulls its weight as a default

crates/languages/src/rust/semantic_token_rules.json
```diff
@@ -1,4 +1,25 @@
 [
+  {
+    "token_modifiers": ["unsafe"],
+    "font_weight": "bold"
+  },
```
Veykril, line 5: This as well. We should probably talk about some interesting setups in the rust docs though for people to customize things here.

crates/languages/src/rust/semantic_token_rules.json
```diff
@@ -1,4 +1,25 @@
...
+  {
+    "token_modifiers": ["mutable"],
+    "underline": true
+  },
```
Veykril, line 9: This would match with the VSCode default of rust-analyzer, I am generally not opposed to this but we need to double check if we can opt out of this with a user rule or not.

PR #49304, opened 2026-02-16
fs: Avoid panic when metadata tasks are cancelled (ZED-4GH)

morgankrey:
Fixes a crash where RealFs::metadata could panic with 'Task polled after completion' when the executor is closed and a spawned runnable is cancelled.

This switches metadata task awaits to fallible awaits and returns structured errors on cancellation, plus adds a regression test for metadata calls after executor shutdown.

Release Notes:

- Fixed a panic in filesystem metadata scanning when background executor tasks are cancelled during shutdown.

### DIFF
--- a/crates/fs/tests/integration/fs.rs
+++ b/crates/fs/tests/integration/fs.rs
@@ -450,6 +450,16 @@ async fn test_realfs_atomic_write_non_existing_file(executor: BackgroundExecutor
 }
 
 #[gpui::test]
+async fn test_realfs_metadata_after_executor_close(executor: BackgroundExecutor) {
+    let fs = RealFs::new(None, executor.clone());
+    let temp_dir = TempDir::new().unwrap();
+    executor.close();
+
+    let result = fs.metadata(temp_dir.path()).await;
+    assert!(result.is_err());
+}
+
+#[gpui::test]
 #[cfg(target_os = "windows")]
 async fn test_realfs_canonicalize(executor: BackgroundExecutor) {
     use util::paths::SanitizedPath;
--- a/crates/fs/src/fs.rs
+++ b/crates/fs/src/fs.rs
@@ -882,14 +882,20 @@ impl Fs for RealFs {
         let symlink_metadata = match self
             .executor
             .spawn(async move { std::fs::symlink_metadata(&path_buf) })
+            .fallible()
             .await
         {
-            Ok(metadata) => metadata,
-            Err(err) => {
+            Some(Ok(metadata)) => metadata,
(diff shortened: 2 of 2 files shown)


crates/fs/tests/integration/fs.rs
```diff
@@ -449,6 +449,16 @@ async fn test_realfs_atomic_write_non_existing_file(executor: BackgroundExecutor
     assert_eq!(content, "Hello");
 }
 
+#[gpui::test]
+async fn test_realfs_metadata_after_executor_close(executor: BackgroundExecutor) {
```
eholk, line 453: This test is getting at a design question. I'd argue the test is misusing our `RealFs` API by starting an operation and then closing the executor before completing the operation. I think both panicking or returning an error are reasonable here, but I'm torn about what's better in this case. My first thought is we should panic, since crashing loudly makes it more likely we'll fix the underlying cause. On the other hand, if we're doing something like shutting down, but makes more sense to close the executor and shutdown all the pending operations than to panic.

So maybe returning an error is the way to go and then the caller can decide whether to crash or to swallow the error.

eholk: _(Replied by Claude on @eholk's behalf)_

I think returning an error is the right call here, and I'd push back a bit on the "API misuse" framing.

The cancellation scenario is inherently racy — there's no way for a caller to atomically check "is the executor open?" before calling `metadata`. The executor can close between any two operations during shutdown. This isn't a programming error like an out-of-bounds index; it's a timing issue callers can't reliably avoid. Panicking on a shutdown race condition is a poor contract.

Returning an error also gives callers the right flexibility automatically. Looking at how `metadata` is called across the codebase:
- Callers using `?` (e.g. `prettier_store.rs`, `remote_server.rs`) will propagate the error and abort their work — exactly right during shutdown.
- Callers using `.ok().flatten()` or `is_ok_and(...)` (e.g. `project.rs`, `settings_file.rs`) will silently treat it as "no metadata" — also harmless.

Both outcomes fall out naturally from each caller's existing error-handling intent. That's the composable behavior you described.

One inconsistency worth fixing while we're here: `RealFs::is_dir` still uses a bare `.await` with no `.fallible()`, so it would still panic on cancellation. Since it returns `bool` it can't propagate an error, but `.fallible().await.unwrap_or(false)` would be a safe fix.

ConradIrwin: I don't understand why executor open/close races with task creation/polling.

It seems ideal to hide this behind the executor (it should not poll a task after it's closed)

PR #49333, opened 2026-02-17
Fix stale diagnostic error markers in file tree

### LINKED ISSUE #48289: Error file markers linger in file tree after errors have been fixed
### Reproduction steps

An easy way to trigger it seems to be to open a TypeScript project and then nuke node_modules. This will lead to a bunch of type errors. Then run yarn install and restart the LSP. This will fix the type errors in the editor, but not the project tree

Image

In the editor pane there are no more errors in that file, but in the file tree the files are still marked as having errors.

Like I hinted in the repro steps above, I often have to restart the LSP server to clear errors in the file view, but the error markers remain in the tree view

OS: macOS 15.2
Memory: 36 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

arthjean:
## Summary

Fixes #48289
Closes https://github.com/zed-industries/zed/issues/52021

Diagnostic error markers (red/yellow dots) in the project panel file tree persisted after errors were resolved. Three root causes:

- **WorktreeUpdatedEntries ignored** — when files changed on disk (e.g. `yarn install`), stale diagnostic summaries were never cleared. Added `invalidate_diagnostic_summaries_for_updated_entries()` to clear summaries for `Removed`/`Updated`/`AddedOrUpdated` paths.
- **Missing DiagnosticsUpdated emission on server stop** — `stop_local_language_server()` cleared summaries and sent proto messages but never emitted `LspStoreEvent::DiagnosticsUpdated`, so the project panel never refreshed.
- **Buffer reload not handled** — reloading a buffer from disk did not clear stale summaries. Added `BufferEvent::Reloaded` handler.

All three paths also send zeroed `UpdateDiagnosticSummary` proto messages to downstream collab clients.

## Test plan

- [x] `./script/clippy` passes
- [x] `cargo test -p project -p project_panel -p worktree` passes (319 tests, 0 failures)
- [x] 4 new tests added:
  - `test_diagnostic_summaries_cleared_on_worktree_entry_removal`
  - `test_diagnostic_summaries_cleared_on_worktree_entry_update`
  - `test_diagnostic_summaries_cleared_on_server_restart`
  - `test_diagnostic_summaries_cleared_on_buffer_reload`
- [x] Manual testing: error markers clear when files change on disk
- [x] Manual testing: error markers clear on LSP restart

Release Notes:

- Fixed stale diagnostic data persisting after file reloads, server restarts and FS entry removals


crates/project/src/lsp_store.rs
```diff
@@ -8089,6 +8141,61 @@ impl LspStore {
...
+        };
+
+        let mut cleared_paths: Vec<ProjectPath> = Vec::new();
+        let mut cleared_server_ids: HashSet<LanguageServerId> = HashSet::default();
+
+        for (path, _, change) in changes.iter() {
+            match change {
+                PathChange::Removed | PathChange::Updated | PathChange::AddedOrUpdated => {
+                    if let Some(summaries_by_server_id) = summaries_for_tree.remove(path) {
+                        for (server_id, _) in &summaries_by_server_id {
```
SomeoneToIgnore: Here and in the new method above, we should be able to move `if let Some(..) = self.downstream_client.clone()`, as, logically, we're looping here only for sending the donwstream updates.

crates/project/src/lsp_store.rs
```diff
@@ -8089,6 +8141,61 @@ impl LspStore {
...
+        let Some(summaries_for_tree) = self.diagnostic_summaries.get_mut(&worktree_id) else {
+            return;
+        };
+
+        let mut cleared_paths: Vec<ProjectPath> = Vec::new();
+        let mut cleared_server_ids: HashSet<LanguageServerId> = HashSet::default();
+
+        for (path, _, change) in changes.iter() {
+            match change {
+                PathChange::Removed | PathChange::Updated | PathChange::AddedOrUpdated => {
```
SomeoneToIgnore: Do we really need to process Updated and AddedOrUpdated worktree events here?

If the entry is open in an editor, we will pull its diagnostics based on its edit events, or language server(s) will push the diagnostics themselves.

We do the workspace diagnostics pull in the background also, on each buffer edit, in `LspStore::on_buffer_edited` + presumably more things.

So, the edit part is either covered there, or has a bug with coverage and has to be fixed there instead?

crates/project/src/lsp_store.rs
```diff
@@ -7921,6 +7927,52 @@ impl LspStore {
         None
     }
 
+    fn on_buffer_reloaded(&mut self, buffer: Entity<Buffer>, cx: &mut Context<Self>) {
```
SomeoneToIgnore, line 7967: > reloading a buffer from disk did not clear stale summaries

Can you elaborate what was made to reload a buffer?
If that's an interaction with the open editor, seems that it should query for more diagnostics and invalidate its own cached data, should not it?

SomeoneToIgnore (commented):
Sorry for coming late, the diagnostics part is relatively hard at places and somehow I failed to express what I found odd in the PR for some time.

https://github.com/zed-industries/zed/issues/48289 mentions removing files and language server restarts, those parts are covered good in the PR.

The rest, buffer reload and edited, seem a bit out of place, I've left some comments.

Overall, seems that there's indeed some invalidation gist here that needs solving, it's the way to solve it has to be slightly different.

crates/project/src/lsp_store.rs
```diff
@@ -8122,6 +8106,67 @@ impl LspStore {
...
+        let Some(summaries_for_tree) = self.diagnostic_summaries.get_mut(&worktree_id) else {
+            return;
+        };
+
+        let mut cleared_paths: Vec<ProjectPath> = Vec::new();
+        let mut cleared_server_ids: HashSet<LanguageServerId> = HashSet::default();
+        let downstream = self.downstream_client.clone();
+
+        for (path, _, change) in changes.iter() {
+            match change {
```
SomeoneToIgnore: NIT: could it be rewritten into `changes.iter().filter(|(_, _, change)| change == PathChange::Removed)` ?

crates/project/src/lsp_store.rs
```diff
@@ -8122,6 +8106,67 @@ impl LspStore {
...
+                                        worktree_id: worktree_id.to_proto(),
+                                        summary: Some(proto::DiagnosticSummary {
+                                            path: path.as_ref().to_proto(),
+                                            language_server_id: server_id.0 as u64,
+                                            error_count: 0,
+                                            warning_count: 0,
+                                        }),
+                                        more_summaries: Vec::new(),
+                                    })
+                                    .log_err();
```
```suggestion
                                    .ok();
```

`log_err` prints out a backtrace and is triggered only when there's nothing to send to, so seems pointless to have here, it seems?

------------

Now it seems that we iterate `summaries_by_server_id` twice?
I wonder if we could merge that together and do the `if let Some((client, project_id)) = &downstream` check there — sorry if this is how it was before, I think those closures around are relatively large and confusing (hence posted another comment about the refactor proposal).

crates/project/src/lsp_store.rs
```diff
@@ -8122,6 +8106,67 @@ impl LspStore {
         }
     }
 
+    fn invalidate_diagnostic_summaries_for_updated_entries(
```
```suggestion
    fn invalidate_diagnostic_summaries_for_removed_entries(
```

SomeoneToIgnore (commented):
Thank you, I now have mostly style nits and one large question here: it seems more reasonable to update buffer's diagnostics, not the workspace (many buffers) when a single buffer is reloaded?

`LspStore` has `pub fn pull_diagnostics_for_buffer` which should suffice for such cases?
Workspace diagnostics refresh can be, again, by the server's request or other, open editors' logic.

That should also help with the notes of the regular buffer's pull prevalence over workspace diagnostics: hopefully the new code structure removes such shenanigans out of this particular test case.
As a downside, we'd be better storing some `HashMap<BufferId, Task<()>>` and removing that on buffer invalidation — `.detach()` is bad in this (and many more) cases as we loose the cancellation control.

SomeoneToIgnore (approved):
Thank you, this seems quite clean compared to the original version.

PR #49337, opened 2026-02-17
language_models: Activate StreamEnded support in completion streams for Cloud provider

tomhoule:
Now that the cloud-side StreamEnded support has landed, enable it on the Zed client:

- Send the `x-zed-client-supports-stream-ended-request-completion-status` header so cloud knows to send `StreamEnded` status events.
- Track whether `StreamEnded` was received during the stream.
- Emit `StreamEndedUnexpectedly` error when the stream closes without a `StreamEnded` status, indicating the response was truncated.

Concretely, this means that the agent panel will now display an abnormally terminated response stream for model calls through a Zed subscription as an error instead of just stopping.

Release Notes:

- The agent panel now displays an abnormally terminated response stream for model calls through a Zed subscription as an error instead of just stopping.

### DIFF
--- a/crates/language_models/src/provider/cloud.rs
+++ b/crates/language_models/src/provider/cloud.rs
@@ -5,9 +5,10 @@ use chrono::{DateTime, Utc};
 use client::{Client, UserStore, zed_urls};
 use cloud_api_types::Plan;
 use cloud_llm_client::{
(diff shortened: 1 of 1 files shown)


tomhoule:
> Concretely, this means that the agent panel will now display an abnormally terminated response stream for model calls through a Zed subscription as an error instead of just stopping.

Something for the release notes? Or irrelevant internal implementation detail?

edit: feels like release notes relevant, adding it.

PR #49366, opened 2026-02-17
worktree: Close single-file worktrees when file is deleted

### LINKED ISSUE #34864: Zed goes crazy when a buffer points to a nonexistent file in /tmp/.
### Summary

To reproduce: I think _any_ sequence causing you to have an open buffer to a nonexistent path in /tmp/ will do, but here's one:

- Open zed with a workspace root somewhere outside of /tmp/. (E.g. `zed ~`.) [I don't know if this is required.]
- Create a file directly in /tmp/ (not in a subdirectory). (E.g. `touch /tmp/foobarbaz`.) [I don't know if the restriction is required.]
- Open that file in Zed.
- Delete the file while leaving the buffer open. (E.g. `rm /tmp/foobarbaz`.)

I think this may have to do with workspaces in a way where it matters that, before doing this, your current zed workspace root is somewhere outside of /tmp/. HOWEVER, I suspect putting a workspace root at /tmp/ might by itself cause Zed to go crazy in the same way (or other ways.)

Doing this results in approximately the following sequence (seen in strace), which I find very suspicious:
- `statx` on `/tmp/`
- `readlink` on `/tmp/foobarbaz` (doesn't exist)*
- make a directory in `/tmp/` (this _only_ happens if the readlink fails)
- make some files in that directory that appear to be testing the filesystem for case sensitivity
- make various filesystem calls related to those files and that directory, then remove all of them
- look for git repos in various places (/tmp/foobarbaz/.git, /tmp/.git, /.git), several times, as well as .gitignore files
- try to readlink /tmp/ and the file again
- try to open the file
- log this failure
- ... do this all again from the top.

For files that exist, including in /tmp/, this stops at the second line above (the readlink), never does the case sensitivity check, and there's no problem.

For files that don't exist outside of /tmp/, this whole sequence happens once, but does not repeat.

My _guess_ is that you are putting an inotify or something on /tmp/ (as soon as /tmp/foobarbaz becomes a project root or something), and then you are self-triggering your own inotify with the case check stuff, and once that happens you go to recheck everything nearby, which causes you to notice the file isn't there, which causes you to do the case check... etc. forever until you die.

(I think this process on its own would not end the world, BUT I think there's some leak or leaks in Zed that get hit once every time through this loop, which rapidly ends the world if the loop is allowed to continue.)

Zed: v0.195.5 (Zed) 
OS: Linux X11 pop 22.04
Memory: 39 GiB
Architecture: x86_64
GPU: Intel(R) Graphics (RPL-U) || Intel open-source Mesa driver || Mesa 24.0.3-1pop1~1711635559~22.04~7a9f319

dastrobu:
When a single-file worktree's root file no longer exists, the background scanner would previously enter an infinite retry loop attempting to canonicalize the path. This caused continuous error logging and resource waste.

This fix detects when a single-file worktree root cannot be canonicalized (after attempting the file handle fallback) and emits a new Deleted event, allowing the worktree to be properly closed.

This is most commonly encountered with temporary files, logs, and similar files that are opened in Zed and then deleted externally, but persist in the workspace database across sessions.

Closes #34864

## Test
**Logs** from manual testing:

```
2026-02-17T16:16:11+01:00 INFO  [worktree] inserting parent git repo for this worktree: "tmp.md"
2026-02-17T16:16:17+01:00 ERROR [worktree] root path could not be canonicalized: canonicalizing "/Users/***/tmp/tmp.md": No such file or directory (os error 2)
2026-02-17T16:16:17+01:00 INFO  [worktree] single-file worktree root "/Users/***/tmp/tmp.md" no longer exists, marking as deleted
2026-02-17T16:16:17+01:00 INFO  [worktree] worktree root /Users/***/tmp/tmp.md no longer exists, closing worktree
```
Release Notes:

- Fixed an issue where Zed would enter an infinite retry loop when the backing file for a single-file worktree was deleted

### DIFF
--- a/crates/edit_prediction/src/license_detection.rs
+++ b/crates/edit_prediction/src/license_detection.rs
@@ -308,7 +308,9 @@ impl LicenseDetectionWatcher {
                         }
                     }
                 }
-                worktree::Event::DeletedEntry(_) | worktree::Event::UpdatedGitRepositories(_) => {}
+                worktree::Event::DeletedEntry(_)
+                | worktree::Event::UpdatedGitRepositories(_)
+                | worktree::Event::Deleted => {}
             });
 
         let worktree_snapshot = worktree.read(cx).snapshot();
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -4359,7 +4359,8 @@ impl LspStore {
                         this.update_local_worktree_language_servers(&worktree, changes, cx);
                     }
                     worktree::Event::UpdatedGitRepositories(_)
-                    | worktree::Event::DeletedEntry(_) => {}
+                    | worktree::Event::DeletedEntry(_)
+                    | worktree::Event::Deleted => {}
                 })
                 .detach()
             }
--- a/crates/project/src/manifest_tree.rs
+++ b/crates/project/src/manifest_tree.rs
@@ -59,6 +59,7 @@ impl WorktreeRoots {
                         let path = TriePath::from(entry.path.as_ref());
                         this.roots.remove(&path);
                     }
+                    WorktreeEvent::Deleted => {}
                 }
             }),
         })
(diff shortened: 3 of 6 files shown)


probably-neb:
@zed-zippy approved

PR #49367, opened 2026-02-17
project_panel: Add support for pasting external files on macOS

davidgatea:
Part of #29026

## Summary
https://github.com/user-attachments/assets/35b4f969-1fcf-45f4-88cd-cbc27ad9696e

macOS Finder places file paths on the system pasteboard using `NSFilenamesPboardType` when files are copied. Previously, the project panel only supported its own internal clipboard for copy/cut/paste operations and ignored system clipboard content entirely. This meant that copying files in Finder and pasting them in Zed's project panel did nothing.

This PR adds support for reading file paths from the macOS system pasteboard, enabling a natural workflow where users can copy files in Finder (or other file managers) and paste them directly into Zed's project panel.

> **Note:** Pasting files from a system file manager currently only works on macOS. The project panel changes are cross-platform, but the clipboard reading of file paths (`ExternalPaths`) is only implemented in the macOS pasteboard. Windows and Linux would need equivalent changes in their respective platform clipboard implementations to support this. Copying/cutting files from the project panel to the system clipboard as plain text works on all platforms.

### Changes

**`crates/gpui/src/platform/mac/pasteboard.rs`**
- Read `NSFilenamesPboardType` from the system pasteboard and surface file paths as `ClipboardEntry::ExternalPaths`
- Check for file paths before plain text, since Finder puts both types on the pasteboard (without this priority, file paths would be returned as plain text strings)
- Extract the string-reading logic into `read_string_from_pasteboard()` to allow reuse

**`crates/project_panel/src/project_panel.rs`**
- On paste, check the system clipboard for external file paths and use the existing `drop_external_files` mechanism to copy them into the project
- On copy/cut, write the selected entries' absolute paths to the system clipboard so other apps can consume them
- Update the "Paste" context menu item to also be enabled when the system clipboard contains file paths, not just when the internal clipboard has entries

## Test plan

- [ ] Copy one or more files in Finder, paste in the project panel — files should be copied into the selected directory
- [ ] Copy files within the project panel, paste — existing internal copy/paste behavior is preserved
- [ ] Cut files within the project panel, paste — existing internal cut/paste behavior is preserved
- [ ] Copy files in the project panel, paste in Finder or another app — paths are available as plain text
- [ ] Right-click context menu shows "Paste" enabled when system clipboard has file paths
- [ ] Right-click context menu shows "Paste" disabled when both internal and system clipboards are empty

Release Notes:

- Added support for pasting files from Finder (and other file managers) into the project panel via the system clipboard (macOS only). Copying or cutting files in the project panel now also writes their paths to the system clipboard for use in other apps.

### DIFF
--- a/crates/gpui_macos/src/pasteboard.rs
+++ b/crates/gpui_macos/src/pasteboard.rs
@@ -1,16 +1,23 @@
 use core::slice;
-use std::ffi::c_void;
+use std::ffi::{CStr, c_void};
+use std::path::PathBuf;
 
 use cocoa::{
-    appkit::{NSPasteboard, NSPasteboardTypePNG, NSPasteboardTypeString, NSPasteboardTypeTIFF},
+    appkit::{
+        NSFilenamesPboardType, NSPasteboard, NSPasteboardTypePNG, NSPasteboardTypeString,
+        NSPasteboardTypeTIFF,
+    },
     base::{id, nil},
-    foundation::NSData,
+    foundation::{NSData, NSFastEnumeration, NSString},
 };
 use objc::{msg_send, runtime::Object, sel, sel_impl};
+use smallvec::SmallVec;
 use strum::IntoEnumIterator as _;
 
 use crate::ns_string;
-use gpui::{ClipboardEntry, ClipboardItem, ClipboardString, Image, ImageFormat, hash};
+use gpui::{
+    ClipboardEntry, ClipboardItem, ClipboardString, ExternalPaths, Image, ImageFormat, hash,
+};
 
 pub struct Pasteboard {
     inner: id,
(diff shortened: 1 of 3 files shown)


davidg0022:
bump 🥹

smitbarmase (changes requested):
Thanks! This is a great feature that we internally wanted as well. I've added the necessary tests for `pasteboard` and `project_panel`.

While doing that, I realized that we still don't handle the case of copying and pasting from the Project Panel into Finder. Is that intentional? What do the current `write_to_system_clipboard` method calls in the Copy and Cut codepaths do?

smitbarmase (approved):
Thank you! I will merge this once today's releases go out, so that we get more time to test this on Nightly.

davidg0022:
> Thank you! I will merge this once today's releases go out, so that we get more time to test this on Nightly.

Hi, looks like the force push replaced my earlier commits on this branch. Could we add co-author trailers to the commits so the contribution is reflected in the history?

Something like: Co-authored-by: Gatea David <davidgatea21@gmail.com>

Thanks!

PR #49373, opened 2026-02-17
ci: Harden background-agent MVP git authentication loop

morgankrey:
## Summary
- refresh git auth headers before each per-crash iteration in `background_agent_mvp`
- add guarded `git fetch`/`git checkout` handling so one candidate auth failure doesn’t fail the entire run

## Context
- fixes the auth failure seen in workflow run 22106378719 (`fatal: could not read Username for 'https://github.com'`)
- related Linear issue: BIZOPS-853

## Testing
- ruby -e "require 'yaml'; YAML.load_file('.github/workflows/background_agent_mvp.yml')"

Release Notes:

- N/A

### DIFF
--- a/.github/workflows/background_agent_mvp.yml
+++ b/.github/workflows/background_agent_mvp.yml
@@ -179,7 +179,7 @@ jobs:
           2. Follow .factory/prompts/crash/investigate.md and write ANALYSIS.md
           3. Follow .factory/prompts/crash/link-issues.md and write LINKED_ISSUES.md
           4. Follow .factory/prompts/crash/fix.md to implement a minimal fix with tests
-          5. Run validators required by the fix prompt for the affected crate(s)
+          5. Run validators required by the fix prompt for the affected code paths
           6. Write PR_BODY.md with sections:
              - Crash Summary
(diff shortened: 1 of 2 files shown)


crates/gpui/src/platform/mac/pasteboard.rs
MrSubidubi, line 1: This seems to be pulled in from another PR and unrelated to the proposed changes here.

.github/workflows/background_agent_mvp.yml
```diff
@@ -148,8 +148,17 @@ jobs:
               continue
             fi
 
-            git fetch origin main
-            git checkout -B "$BRANCH" origin/main
+            git config --local http.https://github.com/.extraheader "AUTHORIZATION: bearer ${GH_TOKEN}"
+
+            if ! git fetch origin main; then
```
MrSubidubi, line 153: Is this fetch even needed? We use the checkout action just above which should ensure that the changes are based on the lates main already. If we are concerned that main is outdated during the n-th consecutive agent run, should we not instead throw this into a matrix job instead?

PR #49383, opened 2026-02-17
Dedup chatty diagnostic jump requests

probably-neb:
Closes #ISSUE

Release Notes:


crates/edit_prediction/src/edit_prediction.rs
```diff
@@ -1486,6 +1499,28 @@ impl EditPredictionStore {
...
+                DiagnosticSeverity::ERROR => 1u8,
+                DiagnosticSeverity::WARNING => 2,
+                DiagnosticSeverity::INFORMATION => 3,
+                DiagnosticSeverity::HINT => 4,
+                _ => 0,
+            }
+            .hash(&mut hasher);
+            range.start.hash(&mut hasher);
+            range.end.hash(&mut hasher);
+            primary.diagnostic.message.hash(&mut hasher);
```
osyvokon, line 1519: This won't help with the case I'm seeing with basedpyright, unfortunately. Imagine I'm typing an identifier in the wrong place ("f", "fo", "foo"). Every diagnostic will have a different message and a different range.

Ideally, we'd solve the root cause: diagnostic-based predictions and user-initiated predictions sharing a single throttle point (Max knows more about it :)

maxbrunsfeld: @probably-neb I think one of the root causes of the problem is the way we throttle requests - when we rapidly alternate between buffer edits and diagnostic updates, the throttle timeout seems to repeatedly be skipped. I think we may need to restructure the throttling so that we maintain a separate timeout task for the two sources of updates.

If we addressed that part of the problem, I'm wondering if we would still need the hashing-based thing. Thoughts?

probably-neb: That makes sense. I was seeing it work for zig but I think my test case would have the same message each time. I'll scrap this and take a look at splitting the throttles tomorrow morning

probably-neb: That makes sense. I was seeing it work for zig but I think my test case would have the same message each time. I'll scrap this and take a look at splitting the throttles tomorrow morning

PR #49393, opened 2026-02-17
html: Fix JavaScript completions in <script> tags

Dnreikronos:
Closes #49340 

Screenshots:
**Before**
image
**After**
image

**Explanation**
The vscode-html-language-server loads TypeScript type definitions (lib.dom.d.ts, lib.es5.d.ts) via a hardcoded relative path that resolves incorrectly for the extracted package layout, resulting in only 46 keyword completions instead of 976+ with full DOM/browser APIs.

Add a wrapper script that patches fs.readFileSync to redirect reads from the broken path to the actual TypeScript lib location found via require.resolve. Also enable embeddedLanguages for JavaScript in initialization options and add JavaScript to the language_ids mapping.

**Motivation/Impacts/Trade-Offs**
Motivation: JavaScript inside HTML <script> tags has no autocomplete, hover info, or DOM API suggestions. The root cause is a hardcoded relative path in vscode-html-language-server's javascriptLibs.js that resolves to a nonexistent directory in the extracted package layout, so TypeScript type definitions never load. Users get 46 keyword
completions instead of 976+ (no document, fetch, addEventListener, etc.).

Impacts:
- Full JavaScript IntelliSense in <script> blocks (DOM APIs, browser globals, type info on hover)
- HTML and CSS completions are unaffected
- Users with a system-installed server (via PATH) bypass the wrapper and get no benefit
- Slightly increased startup time/memory from loading TypeScript lib definitions
- Replaces an unwrap() with proper error propagation

Trade-offs:
- Monkey-patching fs.readFileSync is a global side effect scoped by a path prefix check — functional but fragile if the upstream package restructures its directories
- Coupled to package internals — if @zed-industries/vscode-langservers-extracted changes its layout, the wrapper path computation will need updating
- Crash vs. degrade — if require.resolve('typescript') fails, the server crashes at startup instead of silently falling back to keyword-only completions
- The proper long-term fix is an upstream patch to vscode-langservers-extracted itself; this is a pragmatic workaround until that happens

Release Notes:

- Fixed JavaScript completions in <script> tags


MrSubidubi (changes requested):
Thanks for this! 

I am rather unsure whether we should fix this here and not do it at https://github.com/zed-industries/vscode-langservers-extracted - would it not make more sense to fix this at the root instead?

MrSubidubi (approved):
Stuff here looks fine, reassigning for the needed change and bump in https://github.com/zed-industries/vscode-langservers-extracted/pull/6

msub2:
Companion PR has been merged, any additional work left to do here?

Dnreikronos:
> Companion PR has been merged, any additional work left to do here?

For my part I believe that this is all 100%

9ithu6-c0m:
It seems all done? It would be really nice if this PR could be reviewed again and merged, I've been waiting for this bug to be fixed since I started using Zed 🙏

Dnreikronos:
Guys, can you, please, guide me which approach I have to take for this PR be re-reviewed and approved?

cc// @esthertrapadoux or @JosephTLyons

ChristopherBiscardi:
The issue this references has been fixed for awhile: https://github.com/zed-industries/zed/issues/49340

msub2:
I am on Zed 1.9.0 and a fresh HTML file still does not give me any JS code completions inside a script tag.

image

ChristopherBiscardi:
I'd suggest filing a new issue if you're still seeing this.

image

PR #49416, opened 2026-02-17
editor: Add semantic highlighting to outline panel and modal

ozacod:
This PR adds the semantic highlighting to outline panel and modal. All configuration options("Off", "Combined", "Full")  are respected by the panel.

| Before (syntax only) | After: Off | After: Combined | After: Full |
|---|---|---|---|
| before | after-off | after-combined | after-full |

Release Notes:

- Added semantic token highlighting to outline panel and modal.

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -1936,14 +1936,17 @@ impl DisplaySnapshot {
         })
     }
 
-    /// Returns combined highlight styles (tree-sitter syntax + semantic tokens)
-    /// for a byte range within the specified buffer.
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (commented):
This seems to work odd when there are highlights with fallbacks?
For example, for this snippets, vtsls does not send much semantic tokens

```
return (
    KEYWORD_LIST.some((term) =>
      new RegExp(`\\b${term}\\b`, 'i').test(inputString)
    ) || /[?？؟՞¿‽]/.test(inputString)
  )
```

but I have the `combined` semantic tokens mode and see the highlights thanks to tree-sitter:

image

seems that we should do the same in both outlines.

PR #49422, opened 2026-02-17
Fix Bedrock API validation errors from malformed message history

dsturnbull:
### Problem

Long agent chat threads can accumulate malformed message data that causes Bedrock `ValidationException` errors. Two distinct bugs were identified, both reproduced against Bedrock. The Anthropic API likely has the same constraints but was not the source of the original errors.

**Bug 1 — Thinking block ordering after message merge**

`into_bedrock()` and `into_anthropic()` merge consecutive same-role messages by extending the last message's content array. When an incomplete/interrupted assistant response (text only, no thinking block) is followed by a complete response (thinking + text), the merge produces:

```
[Text("The"), Thinking("..."), Text("Full response")]
```

Both APIs require thinking blocks to come before all other content. Bedrock rejects this with:

> *"messages.N.content.0: If an assistant message contains any thinking blocks, the first block must be thinking or redacted_thinking. Found text."*

**Bug 2 — Orphaned tool results with no matching tool use**

`AgentMessage::to_request()` conditionally emits `ToolUse` blocks (only if they have a matching `tool_result`), but unconditionally emits all `tool_result` entries. When a `ToolUse` content block is lost during cancellation or an interrupted stream, its orphaned `tool_result` creates a count mismatch. Bedrock rejects this with:

> *"number of toolResult blocks at messages.N.content exceeds the number of toolUse blocks of previous turn."*

### Investigation

Thread data is stored in `~/Library/Application Support/Zed/threads/threads.db` as zstd-compressed JSON blobs. Extracting and inspecting with `jq` revealed the root causes.

**Bug 1:** Thread "Zod Infer vs Template Literal Types Tradeoffs" (`0ccd9860-...`). Message at index 53 was an interrupted 3-character response saved without its thinking block:

```
$ jq '.messages[53]' broken_thread.json
{
  "Agent": {
    "content": [{"Text": "The"}],
    "tool_results": {},
    "reasoning_details": null
  }
}
```

When `into_bedrock()` merges this with the next assistant message (which has thinking blocks), the text ends up before the thinking block, violating the API constraint.

**Bug 2:** Thread "Identifying Who Commits to Git Repo" (`7b8a52c2-...`). Message at index 135 had 4 `ToolUse` entries in `content` but 5 entries in `tool_results`:

```
$ jq '{
  content_tool_use_ids: [.messages[135].Agent.content[] |
    select(type == "object" and has("ToolUse")) | .ToolUse.id],
  tool_result_ids: (.messages[135].Agent.tool_results | keys)
}' broken_thread.json
{
  "content_tool_use_ids": [
    "tooluse_zi6Be3NTg26FfSoRiPiVs8",
    "tooluse_JEwViln0DjQwow6LfDMuE3",
    "tooluse_MwiJSm5kjiXcWLSkphe4nP",
    "tooluse_gGUQc6OVZwDhUDq88lC48j"
  ],
  "tool_result_ids": [
    "tooluse_JEwViln0DjQwow6LfDMuE3",
    "tooluse_MwiJSm5kjiXcWLSkphe4nP",
    "tooluse_gGUQc6OVZwDhUDq88lC48j",
    "tooluse_v26EzBOJAYNq55u3e74iuz",
    "tooluse_zi6Be3NTg26FfSoRiPiVs8"
  ]
}
```

The orphan `tooluse_v26EzBOJAYNq55u3e74iuz` exists in `tool_results` but has no matching `ToolUse` in `content` — likely lost during cancellation or an interrupted stream.

#### Manual rectification:

```shell
cat "$TMPDIR/broken_thread3.json" \
  | jq 'del(.messages[135].Agent.tool_results.tooluse_v26EzBOJAYNq55u3e74iuz)' \
  | zstd -3 -o "$TMPDIR/broken_thread3_fixed.zst" \
&& sqlite3 ~/Library/Application\ Support/Zed/threads/threads.db \
     "UPDATE threads SET data = readfile('$TMPDIR/broken_thread3_fixed.zst') WHERE id = '7b8a52c2-ae22-4946-a45e-45278ee17a21';"
```

### Fix

**Bug 1:** Added a stable sort post-processing step after the message collection loop in `into_bedrock()`, `into_anthropic()`, and `into_anthropic_count_tokens_request()`. This moves thinking/reasoning blocks to the front of assistant messages while preserving relative order within each group.

**Bug 2:** Track which `tool_use` IDs are actually emitted into the assistant message in `AgentMessage::to_request()`, then skip `tool_result` entries that don't have a matching `ToolUse` in the content.

### Relationship to #48002

PR #48002 (`d83d9d3e8c`) fixed one specific root cause of orphaned tool_results: `handle_tool_use_json_parse_error_event()` was creating a `tool_result` without adding the corresponding `ToolUse` to content. That fix addresses the **write path** for that one scenario.

This PR's Bug 2 fix is complementary — it's a **read path** defense in `to_request()` that catches orphans regardless of how they were created (cancellation, stream interruption, or any future code path). It also handles already-corrupted threads that were saved before #48002 landed.

### Testing

- `test_thinking_blocks_reordered_before_text_after_merge` (bedrock + anthropic)
- `test_thinking_block_order_preserved_when_already_correct` (bedrock)
- `test_redacted_thinking_blocks_also_reordered_before_text` (anthropic)
- `test_to_request_skips_orphaned_tool_results` (agent)

### Notes

- Both bugs were reproduced and diagnosed against **Bedrock**. The Anthropic fix is defensive — the same merging code path exists in `into_anthropic()` but the error was not observed there firsthand.
- Both bugs manifest only in long-running threads where interrupted responses or cancelled tool uses accumulate in the message history.

--

Release Notes:
- Fixed Bedrock API errors in long agent threads caused by thinking blocks appearing out of order after message merging
- Fixed API errors from orphaned tool results that no longer have a matching tool use in the conversation history

### DIFF
--- a/crates/language_models/src/provider/anthropic.rs
+++ b/crates/language_models/src/provider/anthropic.rs
@@ -348,6 +348,17 @@ pub fn into_anthropic_count_tokens_request(
         }
     }
 
+    // Same thinking-block-first reordering as in into_anthropic — see comment there.
+    for message in &mut new_messages {
+        if message.role == anthropic::Role::Assistant {
+            message.content.sort_by_key(|block| match block {
+                anthropic::RequestContent::Thinking { .. }
+                | anthropic::RequestContent::RedactedThinking { .. } => 0,
+                _ => 1,
+            });
+        }
+    }
+
     CountTokensRequest {
         model,
         messages: new_messages,
@@ -688,6 +699,21 @@ pub fn into_anthropic(
                 }
                 system_message.push_str(&message.string_contents());
             }
+        }
+    }
+
+    // Anthropic requires that thinking blocks come before all other content in
+    // assistant messages. When consecutive assistant messages are merged (e.g.,
+    // an incomplete response followed by a complete one), text blocks can end
+    // up before thinking blocks, violating this constraint. Use a stable sort
+    // to move thinking blocks to the front while preserving relative order.
+    for message in &mut new_messages {
+        if message.role == anthropic::Role::Assistant {
+            message.content.sort_by_key(|block| match block {
+                anthropic::RequestContent::Thinking { .. }
+                | anthropic::RequestContent::RedactedThinking { .. } => 0,
+                _ => 1,
+            });
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/language_models/src/provider/anthropic.rs
```diff
@@ -690,6 +701,21 @@ pub fn into_anthropic(
         }
     }
 
+    // Anthropic requires that thinking blocks come before all other content in
```
dsturnbull, line 705: fwiw I don't have an actual source for this assertion. It just seems to work. But Anthropic docs might be consulted in detail for this.

PR #49479, opened 2026-02-18
feature_flags: Add note about `acp-beta` feature flag

maxdeviant:
This PR adds a note to the `acp-beta` feature flag about how it is used, and that it should not be deleted if unreferenced.

Release Notes:

- N/A


benbrandt (approved):
thanks! 🙏🏻

PR #49481, opened 2026-02-18
agent_ui: Route agent thread feedback comments through cloud

tomhoule:
The agent thread feedback comments event was still going through the standard telemetry pipeline. This routes it through cloud instead, as was just done for agent thread rated and edit prediction rated, so we can enforce organization data and privacy settings.

Part of CLO-297

### DIFF
--- a/crates/agent_ui/src/acp/thread_view/active_thread.rs
+++ b/crates/agent_ui/src/acp/thread_view/active_thread.rs
@@ -1,4 +1,4 @@
-use cloud_api_types::SubmitAgentThreadFeedbackBody;
+use cloud_api_types::{SubmitAgentThreadFeedbackBody, SubmitAgentThreadFeedbackCommentsBody};
 use gpui::{Corner, List};
 use language_model::LanguageModelEffortLevel;
 use settings::update_settings_file;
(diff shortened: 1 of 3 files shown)


crates/agent_ui/src/acp/thread_view/active_thread.rs
```diff
@@ -84,18 +85,28 @@ impl ThreadFeedbackState {
...
 
-        let session_id = thread.read(cx).session_id().clone();
-        let agent_telemetry_id = thread.read(cx).connection().telemetry_id();
+        let project = thread.project().read(cx);
+        let client = project.client();
+        let user_store = project.user_store();
+        let organization = user_store.read(cx).current_organization();
+
+        let session_id = thread.session_id().clone();
+        let agent_telemetry_id = thread.connection().telemetry_id();
```
tomhoule: I refactored. the thread.read(cx) repeated calls to keep the thread in a variable, because it (at least visually and superficially) felt like. duplicated logic / work. LMK if that’s a bad idea or not consistent with the conventions.

tomhoule: Seems like there is a pattern of repeated calls to `.read()`, that it’s cheap and keeping these references alive longer than necessary is not a good idea (borrowing of cx).  Walking that back.

PR #49504, opened 2026-02-18
Revert "Added support for dismissing the toolchain path selector via …

polyesterswing:
…`esc` (#48201)"

This reverts commit 174fc5115243fda191da6a526fa1f12b627c1fcd.

Select toolchain path dialog immediately dismisses upon selecting a path, preventing the addition of new toolchains

- OpenPathDelegate emits DismissEvents in the `confirm` function too
- when the selection is confirmed, DismissEvent is emitted
- ToolchainSelector calls cancel because it receives the DismissEvent, and it goes to the Search State
- This prevents the selected toolchain from being resolved and the ScopePicker being presented

cc: @JosephTLyons 

Release Notes:

- Fixed a regression where the toolchain path selector would immediately dismiss upon selecting a path, preventing the addition of new toolchains.

### DIFF
--- a/crates/toolchain_selector/src/toolchain_selector.rs
+++ b/crates/toolchain_selector/src/toolchain_selector.rs
@@ -119,20 +119,10 @@ impl AddToolchainState {
                 picker
             });
 
-            let weak_toolchain_selector = weak.clone();
-            let window_handle = window.window_handle();
             Self {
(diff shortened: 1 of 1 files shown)


JosephTLyons:
Nice catch @polyesterswing - reverting for now is probably the best thing to do here.

PR #49519, opened 2026-02-18
git: Add diff stats in git_panel

bobbymannino:
This PR adds the small UI change of `git diff --numstat` to the git panel so you can see the number of additions/deletions per file. There is an option in the settings UI for this under `git_panel`.`diff_stats`. This option is set to `false` by default. 

Screenshot 2026-02-18 at 21 25 02

Release Notes:

- Added git diff stats to git panel entries


bobbymannino:
thanks @danilo-leal, i will make sure to make the descriptions consistent and punctual in the future, my bad

danilo-leal (commented):
I think this is pretty much ready to go! I'd just like @Anthony-Eid to take a look at it before we merge 👍

crates/git_ui/src/git_panel.rs
```diff
@@ -644,6 +645,8 @@ pub struct GitPanel {
     local_committer_task: Option<Task<()>>,
     bulk_staging: Option<BulkStaging>,
     stash_entries: GitStash,
+    diff_stats: HashMap<RepoPath, DiffStat>,
+    diff_stats_task: Task<()>,
```
Anthony-Eid, line 649: If we want to allow requesting diff_stats for a single file the task should be stored in the `diff_stats` hash map.

crates/git_ui/src/git_panel.rs
```diff
@@ -770,6 +785,33 @@ impl GitPanel {
...
+                if let BufferStoreEvent::BufferAdded(buffer) = event {
+                    cx.subscribe(buffer, |this, _buffer, event, cx| {
+                        if matches!(event, BufferEvent::Saved) {
+                            if GitPanelSettings::get_global(cx).diff_stats {
+                                this.fetch_diff_stats(cx);
+                            }
+                        }
+                    })
+                    .detach();
+                }
```
Anthony-Eid, line 811: This could be improved by allowing diff stats to fetch for singular files. IO is slow, and when one buffer is saved, we shouldn't invalidate the diff_stats we have for every other buffer. 

Also, starting a process has overhead, which we run into here if a user saves multiple files back-to-back. Because fetch diff stats will be called rapidly in succession.

Finally, you're essentially using the same closure for both Buffer subscriptions. I think this could be a function or inline closure passed to them instead.

crates/git_ui/src/git_panel.rs
```diff
@@ -5126,6 +5223,19 @@ impl GitPanel {
...
+                    move |this, stat| {
+                        let id = format!("diff-stat-{}", id_for_diff_stat);
+                        this.child(ui::DiffStat::new(
+                            id,
+                            stat.added as usize,
+                            stat.deleted as usize,
+                        ))
+                    },
+                )
+            })
```
Anthony-Eid, line 5234: If self.diff_stat doesn't exist for a visible file in the git panel should we fetch its stats?

Anthony-Eid (changes requested):
While not fully blocking, I think we could improve with an architecture that focuses on catching specific file diff stats when single buffers are saved. 

We could also have a queue that refreshes every 50 ms for multiple buffer saves that happen rapidly; that way, we don't have to spawn more than one process when a user saves multiple files at once. 

Finally, we might not need to call diff_stat at all; there's a chance `BufferDiff` has the information we want only stored.

Anthony-Eid (approved):

danilo-leal:
@secondl1ght changed in here → https://github.com/zed-industries/zed/pull/51215

ion-elgreco:
I noticed the diff stats get stale pretty quickly, only staging them updates the correct value, but unstaging untracked changes makes it disappear

PR #49521, opened 2026-02-18
project_panel: Fix selection not updating for already-visible gitignored files

### LINKED ISSUE #49515: Relocating to a git ignored file doesn't mark the file as selected in the project panel
### Reproduction steps

1. Start Zed
2. Open a file
3. Open another file which is included in `.gitignore`
4. Move between those two files  with `ctrl-tab`
5. Ignored file doesn't get marked as selected in the project panel

## Current behavior

When moving between two files with `ctrl-tab` where one of the file is included in `.gitignore`, the ignored file doesn't get the background color as it normally does when moving to an excluded file. Even moving between the two ignored files the same thing happens.

## Expected behavior

Even though the file is git ignored moving there with `ctrl-tab`, the file should be separated from others as it normally does when using mouse click to open it.

Zed: 0.220.7 (e5b3d51046a85424108f41c3fc6e5e4da68ea706)
OS: Arch Linux
Memory: 16GiB
Architecture: x86-64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

skyline69:
## Summary
- Keep auto-reveal behavior for ignored files unchanged (no implicit reveal).
- When an ignored file is already visible in the project panel, mark it as selected on `ActiveEntryChanged`.
- Add regression coverage for switching back to a visible gitignored file.

## Testing
- `project_panel_tests::test_autoreveal_and_gitignored_files`
- `project_panel_tests::test_gitignored_and_always_included`
- `project_panel_tests::test_explicit_reveal`

Closes #49515

Release Notes:

- Fixed project panel not updating selection when switching to a gitignored file that was already visible.

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -5966,12 +5966,26 @@ impl ProjectPanel {
             .worktree_for_entry(entry_id, cx)
             .context("can't reveal a non-existent entry in the project panel")?;
         let worktree = worktree.read(cx);
-        if skip_ignored
-            && worktree
-                .entry_for_id(entry_id)
-                .is_none_or(|entry| entry.is_ignored && !entry.is_always_included)
-        {
-            anyhow::bail!("can't reveal an ignored entry in the project panel");
+        let worktree_id = worktree.id();
+        let is_ignored = worktree
+            .entry_for_id(entry_id)
+            .is_none_or(|entry| entry.is_ignored && !entry.is_always_included);
+        if skip_ignored && is_ignored {
+            if self.index_for_entry(entry_id, worktree_id).is_none() {
+                anyhow::bail!("can't reveal an ignored entry in the project panel");
+            }
+
+            self.selection = Some(SelectedEntry {
+                worktree_id,
+                entry_id,
+            });
+            self.marked_entries.clear();
+            self.marked_entries.push(SelectedEntry {
+                worktree_id,
+                entry_id,
+            });
+            cx.notify();
+            return Ok(());
         }
         let is_active_item_file_diff_view = self
             .workspace
@@ -5983,7 +5997,6 @@ impl ProjectPanel {
             return Ok(());
         }
 
-        let worktree_id = worktree.id();
         self.expand_entry(worktree_id, entry_id, cx);
         self.update_visible_entries(Some((worktree_id, entry_id)), false, true, window, cx);
         self.marked_entries.clear();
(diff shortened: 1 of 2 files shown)


crates/project_panel/src/project_panel.rs
```diff
@@ -5966,12 +5966,26 @@ impl ProjectPanel {
...
+
+            self.selection = Some(SelectedEntry {
+                worktree_id,
+                entry_id,
+            });
+            self.marked_entries.clear();
+            self.marked_entries.push(SelectedEntry {
+                worktree_id,
+                entry_id,
+            });
```
smitbarmase, line 5986: Can we add `self.autoscroll(cx);` here too?

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -4843,6 +4843,43 @@ async fn test_autoreveal_and_gitignored_files(cx: &mut gpui::TestAppContext) {
...
         "When a gitignored entry is explicitly revealed, it should be shown in the project tree"
     );
+
+    panel.update(cx, |panel, cx| {
+        panel.project.update(cx, |_, cx| {
+            cx.emit(project::Event::ActiveEntryChanged(Some(dir_2_file)))
+        })
+    });
+    cx.run_until_parked();
+
```
smitbarmase, line 4874: I think we should assert the immediate state here. It would be more thorough to assert that `dir_2_file` is selected and marked. That way we can confirm that switch actually happened.

smitbarmase (changes requested):

crates/project_panel/src/project_panel.rs
```diff
@@ -5983,14 +5998,14 @@ impl ProjectPanel {
...
 
-        let worktree_id = worktree.id();
         self.expand_entry(worktree_id, entry_id, cx);
         self.update_visible_entries(Some((worktree_id, entry_id)), false, true, window, cx);
         self.marked_entries.clear();
         self.marked_entries.push(SelectedEntry {
             worktree_id,
             entry_id,
         });
+        self.autoscroll(cx);
```
smitbarmase: Actually, we don't need auto scroll here because `self.update_visible_entries` handles that for us internally. It calls `auto scroll` since we pass `true`.

The reason we don't use `update_visible_entries` in our code path is because there we are not changing anything about entries visually, which, in this case, `expand_entry` does.

smitbarmase (changes requested):
Rest looks good to me, just a minor nit.

smitbarmase (approved):
Thanks!

skyline69:
> Thanks!

You're welcome!

PR #49527, opened 2026-02-18
Allow the keymap and settings UI to be opened without an active window

### LINKED ISSUE #49155: Settings window won't open without an open project window
### Reproduction steps

1. Start Zed
2. Close any/all open Zed windows
3. Press `⌘,`
4. Go to menu bar menus: `Zed > Settings > Open Settings (⌘,)`

# Current

1. Keyboard shortcut does nothing
2. Menubar item is grayed out/disabled, does nothing 

Image

# Expected

1. Keyboard shortcut opens Settings window
2. Menubar item  opens Settings window

Importantly: both of these work as expected, if _any_ window is open

Image

OS: macOS 26.2
Memory: 16 GiB
Architecture: aarch64 

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

transitoryangel:
Closes #49155. 

image

Release Notes:

- UI: Allow the keymap and settings UI to be opened without an active window


crates/settings_ui/src/settings_ui.rs
```diff
@@ -562,7 +595,6 @@ fn init_renderers(cx: &mut App) {
 }
 
 pub fn open_settings_editor(
-    _workspace: &mut Workspace,
```
MrSubidubi, line 567: Good catch!

crates/settings_ui/src/settings_ui.rs
```diff
@@ -387,28 +387,61 @@ struct SettingsFieldMetadata {
...
+            cx.update(|cx| {
+                workspace_window
+                    .update(cx, |_, window, _| {
+                        open_settings_on_next_frame(window);
+                    })
+                    .log_err();
+            });
+            anyhow::Ok(())
+        })
+        .detach_and_log_err(cx);
```
MrSubidubi: I see that the `window_handle` is optional in the settings window code - can we not just make the parameter optional too and then prevent this hack here? Then just a settings window should pop up where you cannot configure project-settings, which would make sense to me

crates/settings_ui/src/settings_ui.rs
```diff
@@ -387,28 +387,61 @@ struct SettingsFieldMetadata {
...
+        open_settings_editor(None, None, window_handle, cx);
+    });
+}
+
 pub fn init(cx: &mut App) {
     init_renderers(cx);
     let queue = ProjectSettingsUpdateQueue::new(cx);
     cx.set_global(queue);
 
+    #[cfg(target_os = "macos")]
```
MrSubidubi: Is there a specific reason that we gate this to MacOS only?

crates/keymap_editor/src/keymap_editor.rs
```diff
@@ -126,6 +126,13 @@ pub fn init(cx: &mut App) {
         }
     }
 
+    #[cfg(target_os = "macos")]
+    cx.on_action(|_: &OpenKeymap, cx| {
+        with_active_or_new_workspace(cx, |workspace, window, cx| {
+            open_keymap_editor(None, workspace, window, cx);
+        });
+    });
+
```
MrSubidubi: I think this makes sense, but can we not remove the other listener then perhaps?

MrSubidubi (changes requested):
Thanks for this! Overall looks sensible, left a few comments

crates/settings_ui/src/settings_ui.rs
transitoryangel: Gated it to MacOS only because I'm not aware of any other case where you can have the program open (and interact with the toolbar as such) with no primary window open - open to changing this though

MrSubidubi: Yeah makes sense, let's remove it though - should not be your concern here and if we ever change stuff related to this (which is not unlikely, this has been a thing of debate), we don't have to consider this case. In theory, the behavior here should be platform-agnostic, so let's pretend it is.

MrSubidubi: Don't understand me wrong, please - I appreciate that you thought about it, but just in the context of what you want to fix, let's pretend this can happen on all platforms.

transitoryangel: No worries! Completely understandable

crates/settings_ui/src/settings_ui.rs
transitoryangel: good catch! ty

crates/settings_ui/src/settings_ui.rs
```diff
@@ -392,23 +392,25 @@ pub fn init(cx: &mut App) {
...
-                    open_settings_editor(workspace, Some(&path), None, window_handle, cx);
-                },
-            )
-            .register_action(|workspace, _: &OpenSettings, window, cx| {
+            .register_action(|_, OpenSettingsAt { path }: &OpenSettingsAt, window, cx| {
+                let window_handle = window
+                    .window_handle()
+                    .downcast::<MultiWorkspace>()
+                    .expect("Workspaces are root Windows");
+                open_settings_editor(Some(&path), None, Some(window_handle), cx);
```
MrSubidubi:
```suggestion
                let window_handle = window
                    .window_handle()
                    .downcast::<MultiWorkspace>();
                open_settings_editor(Some(&path), None, window_handle, cx);
```

MrSubidubi (approved):
Sorry for the delay, looks indeed good now, thanks!

A slight thing I noticed: When I use the now fixed dropdown and have a window on a secondary screen open, that window will be hidden. So I have something fullscreen on my top monitor, then open patched Zed on the lower, close the window and open the settings with your fix, the window on my upper screen will go offscreen.

Not blocking here, but if you feel like it, feel free to look into that.

With that said, thank you for the PR here and your persistence, congratulations on your first contribution and have a great weekend! 🎉

PR #49538, opened 2026-02-18
editor: Fix semantic tokens being fetched when LSP data is disabled

ozacod:
`disable_lsp_data` is currently only used for lhs of split diff view. However, disabling the semantic highlighting is useful in for tasks where a multi-buffer is rendered too(for example, fetching semantic tokens for every buffer is heavyweight operation in project search). 

Issue:
`refresh_semantic_tokens` is called directly by `project::Event::RefreshSemanticTokens`, `theme_changed`, and `settings_changed`. Since the `enable_lsp_data` check is bypassed in `refresh_semantic_tokens`, semantic tokens are still requested for the buffer.  

Release Notes:

- Fixed semantic tokens being fetched with LSP data is disabled.

### DIFF
--- a/crates/editor/src/semantic_tokens.rs
+++ b/crates/editor/src/semantic_tokens.rs
@@ -119,7 +119,7 @@ impl Editor {
         for_server: Option<RefreshForServer>,
         cx: &mut Context<Self>,
     ) {
-        if !self.mode().is_full() || !self.semantic_token_state.enabled() {
+        if !self.mode().is_full() || !self.semantic_token_state.enabled() || !self.enable_lsp_data {
             self.invalidate_semantic_tokens(None);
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (commented):
Very nice catch, thank you.

Seems that we'd better also cover `project::Event::RefreshInlayHints` then, and other places like `register_visible_buffers`, `register_buffer`, `refresh_folding_ranges`, etc., etc. , as those are also able to fetch something LSP-related on e.g. settings change.

What seems to be more viable is to check all `self.mode().is_full()` usages (8 hits) + `fn update_lsp_data` code and extract it all into a method on the `Editor`, that will check both `is_full` and this field.

PR #49539, opened 2026-02-18
Fix font weight in multiple elements

dinocosta:
* Update the font weight used in the Agent Panel's editor to respect the `buffer_font_weight` setting.
* Update the font weight used in the Command Palette to use the `ui_font_weight` setting value, seeing as the command palette is using the `ui_font` settings for the font family and features
* Update the font weight used in the popovers to match the `buffer_font_weight`, as it's also using the buffer font family and features
* Update the `LabelLike.buffer_font` method in order to correctly set the font weight, which fixes the font weight used in the file header when displaying a multibuffer editor

Release Notes:

- Fixed incorrect font weight in the Command Pallete input
- Fixed missing font weight the Agent Panel's buffer
- Fixed missing font weight the Hover Popover
- Fixed missing font weight in Markdown's code block and inline code

### DIFF
--- a/crates/agent_ui/src/acp/message_editor.rs
+++ b/crates/agent_ui/src/acp/message_editor.rs
@@ -1334,7 +1334,7 @@ impl Render for MessageEditor {
                     font_fallbacks: settings.buffer_font.fallbacks.clone(),
                     font_features: settings.buffer_font.features.clone(),
                     font_size: settings.agent_buffer_font_size(cx).into(),
-                    font_weight: settings.buffer_font.weight.clone(),
+                    font_weight: settings.buffer_font.weight,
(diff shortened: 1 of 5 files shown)


dinocosta:
For the screenshots below, a `buffer_font_weight` of `300` was used to exaggerate the effect of these changes and make it easier to see the difference ▼

| Context | Before | After |
|-|-|-|
| Agent Panel Buffer | agent_panel_before | agent_panel_after |
| Hover Popover | hover_before | hover_after |
| Markdown Code | markdown_code_before | markdown_code_after |
| Comand Palette | command_palette_before | command_palette_after |
| Buffer File Header | file_header_before | file_header_after |

PR #49547, opened 2026-02-19
Terminal permissions: Per-command pipeline UI

eholk:
## Summary

Adds a new permission UI for terminal pipeline commands (e.g. `cargo test | tail`) that lets users selectively always-allow individual commands in the pipeline, rather than only offering a blanket always-allow for the first command.

## Screenshot

Screenshot 2026-03-18 at 3 27 48 PM

Release notes:
- The terminal permissions UI now allows you to select individual subcommands independently.


crates/agent_ui/src/conversation_view/thread_view.rs
```diff
@@ -1602,20 +1722,77 @@ impl ThreadView {
...
+            let checked_patterns: Vec<_> = patterns
+                .iter()
+                .enumerate()
+                .filter(|(index, _)| checked.contains(index))
+                .map(|(_, cp)| cp.pattern.clone())
+                .collect();
+
+            if !checked_patterns.is_empty() {
+                let (option_id_str, kind) = if is_allow {
+                    (
```
benbrandt, line 1773: It's a bit of a smell to me that we are doing this string formatting in the UI layer... Maybe we need some additional data structure to encode this?

benbrandt (approved):
One small note. In general, any of this string manipulation feels like it should be in `agent` crate, not the ui. In practice, it won't matter since our agent is the only one using this variant, but it is a small that we have an abstraction leak that could cause weirdness at some point down the line

tredondo:
Related:
- #56817
- #56492

PR #49575, opened 2026-02-19
image_viewer: Use checkerboard from GPUI

MostlyKIGuess:
As we have checkerboard, used same pattern as git diff inside Image Viewer.
Now that CPU instructions aren't an issue, perhaps size can be lowered to 16/24 for base boxes. 

Release Notes:

- Made the checkerboard pattern in the image viewer more visually pleasant.


MrSubidubi (commented):
Have to agree, it indeed does ^^ 

Only thing I'd want to discuss: Do we want to change the colors like this? I'll ping our designer internally, but would like your opinion on it as well.

MrSubidubi (commented):
Got greenlit by design, we like it much more. Thank you for jumping on this!

PR #49583, opened 2026-02-19
docs: Update keybindings in `webstorm.md`

tredondo:
- add Linux/Windows keybindings
- use consistent spacing around the `+` in key combos

Release Notes:

- I don't have access to a Mac, so I haven't verified any of the macOS keybindings. I suspect some may be out of date. Please verify.
- All other migration guides should be updated with Linux/Windows shortcuts, and should use consistent spacing around the `+`:
  - https://zed.dev/docs/migrate/vs-code#differences-in-keybindings
  - https://zed.dev/docs/migrate/intellij#differences-in-keybindings
  - https://zed.dev/docs/migrate/pycharm#differences-in-keybindings
  - https://zed.dev/docs/migrate/rustrover#differences-in-keybindings


MrSubidubi (changes requested):
Your work here is awesome, thank you very much!

Small question: In theory, we have a docs preprocessor where you can use certain commands, see

https://github.com/zed-industries/zed/blob/c6b57b05e2620e2baa3b7a156e6a7aac222343d9/docs/.conventions/CONVENTIONS.md?plain=1#L147-L148

The advantage of these is that
1. You don't have to look up the keybinds anymore, as they stay up to date always
2. We catch errors when actions are renamed, as this will err in CI
3. Keybinds are replaced based on the OS you are viewing this on, which takes the burden from you and others to distinguish between these while writing.

I do not know why these were not used upfront when creating these, but would you be interested in adding them here so that this stays up to date? Not a blocker to get this in, appreciate the effort either way, just curious. Thanks!

MrSubidubi (approved):
Fixed this quickly myself. Thanks for getting the ball rolling here!

PR #49584, opened 2026-02-19
project: Block less in `Respository::paths_changed`

Veykril:
Release Notes:

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -303,7 +303,7 @@ pub struct Repository {
     git_store: WeakEntity<GitStore>,
     // For a local repository, holds paths that have had worktree events since the last status scan completed,
     // and that should be examined during the next status scan.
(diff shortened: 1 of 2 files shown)


crates/project/src/git_store.rs
```diff
@@ -6197,24 +6190,16 @@ impl Repository {
         updates_tx: Option<mpsc::UnboundedSender<DownstreamUpdate>>,
         cx: &mut Context<Self>,
     ) {
-        self.paths_needing_status_update.extend(paths);
-
```
kubkon, line 6201: I seem to remember discussing this with @cole-miller a while back and our conclusion was that this was needed to dedup `git-status` calls. Is that no longer the case?

Veykril: Ah I actually misunderstand how the job key stuff dedupes work, yea then I'll have to change this a bit

PR #49605, opened 2026-02-19
Tail/animated cursor

VincentBurke:
Inspired by Neovide's animated cursor and [ghostty-cursor-shaders](https://github.com/sahaj-b/ghostty-cursor-shaders/tree/main), I've built an implementation for Zed. While not as feature complete, it feels very smooth and performant. I've tried to keep the code changes as unobtrusive as possible. Have tested it for a while and have fixed all bugs I've found. Take a look and let me know what you think, appreciate any feedback:)

https://github.com/user-attachments/assets/1eab825e-bef3-4189-ad62-ba56571234eb

SCR-20260219-gfam
SCR-20260219-gfdx


niekdomi:
this is AWESOME :)

Does this PR fix (doesn't have) the issue that this smooth cursor PR has https://github.com/zed-industries/zed/pull/44770#issuecomment-3679462002?

Since you mentioned ghostty shaders; Do you think its feasible to allow custom shaders so user can write their own `glsl` shaders? This would allow to create all kind of crazy animation, though I have no idea how complex/realistic this is for a whole IDE

PR #49615, opened 2026-02-19
Support MCP image content in tool responses

dastrobu:
Previously, MCP tool responses containing images were simply ignored with a warning log. This change properly handles image content by:

1. Extracting image data from MCP ToolResponseContent::Image
2. Creating LanguageModelImage for the LLM to process
3. Updating the tool call UI via event_stream.update_fields()
4. Returning image content as LanguageModelToolResultContent::Image

The change also:
- Properly handles text content by updating the UI with TextContent blocks
- Logs resource content URIs for debugging instead of silently ignoring
- Adds supports_images field to FakeLanguageModel for testing

## Testing

Added unit tests for MCP tool returning only image content and mixed text/image content.

### Manual Testing with MCP Test Server
The [MCP Test Server](https://github.com/dastrobu/mcp-test-server) was used to verify image handling.

image

with image previews in tool responses:

image

#### Setup
1. Install the test server: `cargo install --git https://github.com/dastrobu/mcp-test-server`
2. Add to Zed `settings.json`: ```json "context_servers": { "mcp-test-server": { "command": "mcp-test-server" } } ```
3. Tested PNG, GIF, JPEG, and WebP formats - all work correctly

Release Notes:

- Added support for image content in MCP tool responses

### DIFF
--- a/crates/agent/src/tests/mod.rs
+++ b/crates/agent/src/tests/mod.rs
@@ -2082,6 +2082,333 @@ async fn test_mcp_tool_truncation(cx: &mut TestAppContext) {
 }
 
 #[gpui::test]
+async fn test_mcp_tool_image_response(cx: &mut TestAppContext) {
+    struct TestCase {
+        name: &'static str,
+        response_content: Vec<context_server::types::ToolResponseContent>,
+        expected_ui_image_count: usize,
+        expected_ui_text_count: usize,
+        expect_llm_image: bool,
+        expected_llm_text: Option<&'static str>,
+    }
+
+    let test_image_base64 = test_png_base64();
+    let test_gif_base64 = test_gif_base64();
+
+    let test_cases = vec![
+        TestCase {
+            name: "text only",
+            response_content: vec![context_server::types::ToolResponseContent::Text {
+                text: "Here is the result.".into(),
+            }],
+            expected_ui_image_count: 0,
+            expected_ui_text_count: 1,
+            expect_llm_image: false,
+            expected_llm_text: Some("Here is the result."),
+        },
+        TestCase {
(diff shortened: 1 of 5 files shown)


bennetbo:
Hey, thanks for working on this and sorry for not responding for so long. I think before landing this change I'd like to check how feasible it is to allow `LanguageModelToolResult` to include multiple results (`content`), so that we don't have to do workarounds of only including one output (like we do here for MCP tools). I'll get back to this soon.

dastrobu:
@bennetbo thanks for the feedback. 

I think there are two related topics: 

 - LanguageModelToolResult include multiple results
 - storing mime-type properly in LanguageModelImage
 
 I'd be willing to contribute to both topics. But I guess this PR is already large, so let me know if I can help out with additional PRs on these topics.

bennetbo:
Main now has `Vec<LanguageModelToolResult>`, so we should be able to make use of this here. Also seems like a bunch of the changes in this PR are not needed/wondering if we can make the changes less invasive.

dastrobu:
> Main now has `Vec<LanguageModelToolResult>`, so we should be able to make use of this here. Also seems like a bunch of the changes in this PR are not needed/wondering if we can make the changes less invasive.

thanks, this is great. I went over the PR and it should be ready now. Tried to make it as least invasive as I can. 
I also tested with mixed content now: 

image

There is a minor UI bug in displaying the mixed content, but I guess fixing this would blow up the PR too much: 

image


(the three collapse buttons are not working as intended). 

I'd appreciate if you could take a look at the PR.

bennetbo:
Hey, thanks for working on this again. I did not fully understand all of the changes on this branch, so decided to implement them a bit differently and bit by bit on other PRs. PR is here: #57134

PR #49624, opened 2026-02-19
git_ui: Show uncommitted change count badge on git panel icon

IVainqueur:
## Summary

- Implements `icon_label` on `GitPanel` to return the total count of uncommitted changes (`new_count + changes_count`) when non-zero, capped at `"99+"` for large repos.
- Updates `PanelButtons::render()` to render that label as a small green badge overlaid on the panel's sidebar icon, using absolute positioning within a `div().relative()` wrapper.
- The badge uses `version_control_added` theme color and `LabelSize::XSmall` text with `LineHeightStyle::UiLabel` for accurate vertical centering, positioned at the top-right corner of the icon button.

The `icon_label` method already existed on the `Panel`/`PanelHandle` traits with a default `None` impl, and was already implemented by `NotificationPanel` (unread notification count) and `TerminalPanel` (open terminal count) — but was never rendered. This wires it up for all three panels at once.

## Notes

- Badge is positioned with non-negative offsets (`top(0)`, `right(0)`) to stay within the parent container's bounds. The status bar's `render_left_tools()` uses `.overflow_x_hidden()`, which in GPUI clips both axes (the `overflow_mask` returns a full content mask whenever any axis is non-`Visible`), so negative offsets would be clipped.
- `LineHeightStyle::UiLabel` collapses line height to `relative(1.)` so flex centering aligns the visual glyph rather than a taller-than-necessary line box.
- No new data tracking logic — `GitPanel` already maintains `new_count` and `changes_count` reactively.
- No feature flag or settings added per YAGNI.

## Suggested .rules additions

The following pattern came up repeatedly and would prevent future sessions from hitting the same issue:

```
## GPUI overflow clipping

`overflow_x_hidden()` (and any single-axis overflow setter) clips **both** axes in GPUI.
The `overflow_mask()` implementation in `style.rs` returns a full `ContentMask` (bounding box)
whenever any axis is non-`Visible`. Absolute-positioned children that extend outside the element
bounds will be clipped even if only the X axis is set to Hidden.
Avoid negative `top`/`right`/`bottom`/`left` offsets on absolute children of containers
that have any overflow hidden — keep badge/overlay elements within the parent's bounds instead.
```

Release Notes:

- Added a numeric badge to the git panel sidebar icon showing the count of uncommitted changes.


Dima-369:
Hey, thanks. Mind sharing some screenshots?

IVainqueur:
Sure thing.
(peep the bottom right corner)

image

Lemme know if it's clear enough or..?

danilo-leal (changes requested):
Heya, thanks for the PR! This is a cool change, but can we make it be under a setting that's turned off by default? I don't think we should have this at all times. And then, even with the setting, I think we should remove the badge if the panel is open.

IVainqueur:
Good suggestion @danilo-leal, added.

danilo-leal (approved):
Thank you! I pushed several changes here, most notably:

- There were what I think are changes from a different branch leaking on on this one (regarding an activity bar implementation, which is very unrelated to this PR). So I removed all of that.
- Had to add settings to the terminal and notification panels given both also implement `icon_label` and I didn't want any badges appearing on them as a result of this PR.
- Created a dedicated component for the count badge instead of creating it directly in the dock file.

@simonfelding, I hear and appreciate your take here, but I think for now we'll have to agree on disagreeing; we really want to preserve Zed's UI minimalism and this could be unnecessary noise. This doesn't mean I'm not open to change it, though, in case we get enough feedback pointing that the default should have it turned on; looking forward to seeing what people will think. But I think whenever not fully sure if something should be turned on by default, a good rule of thumb is to have it turned off.

PR #49632, opened 2026-02-19
Add a cargo wrapper to report build times

ConradIrwin:
This adds `./script/cargo` which is a transparent wrapper around `cargo`
intended for use by internal staff members at Zed.

It:
- adds `--timings` to `cargo build`/`cargo test`/`cargo run`
- saves the output to ZED_DATA_DIR/build-timings

When Zed starts (and you're logged in as a staff member) the build-timings
directory is uploaded to our existing telemetry stack.

Co-authored-by: Marshall <git@maxdeviant.com>

Release Notes:

- N/A

### DIFF
--- a/script/cargo-timing-info.js
+++ b/script/cargo-timing-info.js
@@ -73,7 +73,7 @@ function parseTimestampFromFilename(filePath) {
 }
 
 function writeBuildTimingJson(filePath, durationMs, firstCrate, target, lockWaitMs, command) {
-  const buildTimingsDir = path.join(getZedDataDir(), "build-timings");
+  const buildTimingsDir = path.join(getZedDataDir(), "build_timings");
 
   // Create directory if it doesn't exist
   if (!fs.existsSync(buildTimingsDir)) {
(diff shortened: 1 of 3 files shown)


osiewicz (commented):
I wonder if that'll work with any cargo invokation (e.g. those from RA); that'd be neat.

script/cargo-timing-info.js
```diff
@@ -0,0 +1,183 @@
...
+      lastFinishingUnit = unit;
+    }
+  }
+
+  // Find the first crate that had to be rebuilt (earliest start time)
+  // Sort by start time to find the first one
+  const sortedByStart = [...unitData].sort((a, b) => a.start - b.start);
+  const firstRebuilt = sortedByStart[0];
+
+  // The minimum start time indicates time spent waiting for the cargo lock
```
osiewicz: This is not necessarily true (although that time includes time spent waiting on Cargo.lock). In our repo's case, this figure will be no smaller than 0.6s, which is the time cargo needs to check all of our dependencies and parse the manifests in the workspace.

ConradIrwin: Going to rename to `blocked_ms` - we'll just have to assume that the first second is roughly just cargo being slow, and that beyond that we're likely hitting the lock.

PR #49636, opened 2026-02-19
repl: Add Tree-sitter block detection to Run action

### LINKED ISSUE #15812: REPL: send whole "block" to REPL without having to select it first ("block" can be identified by treesitter)
### Check for existing issues

- [X] Completed

### Describe the feature

Usually, when I have a top-level function, I want to send the whole function to the REPL and not individual lines. This can be achieved with cell mode, but I don't necessarily want to have cell markers in all my files.

So, maybe it's possible to recognize that the cursor is in a top-level function (I think treesitter can give such information) and then to send the whole top-level function to the REPL.

```python
def times_two(x):
    print(x*2)  #         <-- if cursor is here, send the whole function to REPL
times_two(3)  # (<- this line is not part of the block anymore)
```

Similarly, if I have a `for`-loop:

```python
for i in range(3):
    times_two(i)  #         <-- if cursor is here, send the whole for-loop to REPL
```

In Python, these blocks can be identified by identation.

_No response_

rgbkrk:
When the cursor has no selection, the `Run` action now uses treesitter to detect the enclosing top-level block (function, class, loop, etc.) and sends it to the REPL. This lets people send whole blocks without manually selecting them or using Jupytext cell markers.

The implementation:

- Walks up the syntax tree to find the nearest node whose parent is the root
- Respects explicit user selections (selection overrides block detection)
- Integrates with existing cell detection (Jupytext markers, markdown code blocks) as a fallback

Closes #15812

Check out how fluid this is:

treesitterrunner

Release Notes:

- Improved REPL Run to automatically detect and execute the enclosing code block when no text is selected

### DIFF
--- a/crates/repl/src/repl_editor.rs
+++ b/crates/repl/src/repl_editor.rs
@@ -1281,4 +1281,38 @@ mod tests {
         let text: String = snapshot.text_for_range(range.unwrap()).collect();
         assert_eq!(text, "console.log(\"done\");");
     }
+
+    #[gpui::test]
+    fn test_block_range_tsx_multiline(cx: &mut App) {
+        let tsx = languages::language("tsx", tree_sitter_typescript::LANGUAGE_TSX.into());
+
+        // Test multi-line JSX expression
+        let buffer = cx.new(|cx| {
+            let mut buffer = Buffer::local(
+                "renderToJupyter(\n  <div>\n    <h1>Hello</h1>\n  </div>,\n);\n",
+                cx,
+            );
+            buffer.set_language(Some(tsx.clone()), cx);
+            buffer
+        });
+        let snapshot = buffer.read(cx).snapshot();
+
+        // Cursor on closing </div> tag should select entire call expression
+        let range = block_range(&snapshot, Point::new(3, 2)..Point::new(3, 2));
+        assert!(range.is_some());
+        let text: String = snapshot.text_for_range(range.unwrap()).collect();
+        assert_eq!(
+            text,
+            "renderToJupyter(\n  <div>\n    <h1>Hello</h1>\n  </div>,\n);"
+        );
+
+        // Cursor inside nested h1 should also select entire call
+        let range = block_range(&snapshot, Point::new(2, 8)..Point::new(2, 8));
+        assert!(range.is_some());
+        let text: String = snapshot.text_for_range(range.unwrap()).collect();
+        assert_eq!(
+            text,
+            "renderToJupyter(\n  <div>\n    <h1>Hello</h1>\n  </div>,\n);"
+        );
+    }
 }
--- a/crates/repl/src/repl_editor.rs
+++ b/crates/repl/src/repl_editor.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use anyhow::{Context as _, Result};
 use editor::{Editor, MultiBufferOffset};
 use gpui::{App, Entity, WeakEntity, Window, prelude::*};
-use language::{BufferSnapshot, Language, LanguageName, Point};
+use language::{BufferSnapshot, Language, LanguageName, Point, ToOffset};
 use project::{ProjectItem as _, WorktreeId};
 use workspace::{Workspace, notifications::NotificationId};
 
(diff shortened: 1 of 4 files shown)


crates/repl/src/repl_editor.rs
```diff
@@ -282,6 +282,48 @@ pub fn run(
...
+    loop {
+        let node = cursor.node();
+        if let Some(parent) = node.parent() {
+            let parent_kind = parent.kind();
+            // Common root node kinds across languages:
+            // Python: module, JavaScript/TypeScript: program, Rust/Go: source_file, Lua: chunk
+            if matches!(
+                parent_kind,
+                "module" | "program" | "source_file" | "chunk" | "translation_unit"
+            ) {
```
MrSubidubi, line 315: I think we very much would need a better system for this, as well, this should really be living in language configs somewhere. Not too sure it should be another .scm file though, but would love to discuss what advantages different approaches here have.

rgbkrk: Fairly similar to the approach taken in the editor -> terminal repl suggestion in https://github.com/zed-industries/zed/pull/42467/changes#diff-e2b6d1ee5f0d6966b49dcc174ee9c078a428c114abafbd2ce0f468dea8c7ed07R41-R56

Not sure what it should look like but yeah I want some kind of "definitely an expression" that can be run. If possible it would include preceding comments too I guess? Otherwise this kind of "runs" comments (which are no op obviously).

tmke8: I don't know whether it's helpful, but when I implemented a similar functionality in a neovim plugin, I used the following tree-sitter query to get all toplevel blocks for Python:

```scm
; inherit: python

(module
  [
    (function_definition)
    (class_definition)
    (import_statement)
    (import_from_statement)
    (expression_statement)
    (decorated_definition)
    (if_statement)
    (for_statement)
    (while_statement)
    (match_statement)
    (try_statement)
    (with_statement)
    (raise_statement)
    (assert_statement)
    (type_alias_statement)
  ] @toplevel)
```

rgbkrk: Would I need to make a `runnables.scm` for each language in order to make this work? I'm happy with the current approach for now as it makes things much better for folks?

rgbkrk:
I've also tried out making this look at previous and next siblings separated by whitespace. 

image

That feels pretty good too. Kind of opinionated but also pretty easy and fun. It does mean there's no more "run this line" with the REPL. I have not pushed these changes to the branch.

crates/repl/src/repl_editor.rs
MrSubidubi: We would then probably need either that for all languages that should support it or some knob in the `config.toml`. I think as of JSX autoclose, we also host some Tree-sitter configs there.

Queries would give us the advantage that we could in theory capture the same line with two different queries, wondering though whether that'd be a good advantage to have. Personally, I'm also not too enthusiastic about another set of queries, but if we see value in it, why not.

rgbkrk: TSX was a good recommendation, because I found out that `.tsx` files are actually vanilla `TypeScript` by default. This is the result of running some Deno:

image

If you switch it to TSX via the language drop down (and also do some kernel finagling, which will be fixed by https://github.com/zed-industries/zed/pull/49762) it then works really cleanly:

image

rgbkrk:
I truly love this way of operating a repl within Zed but this PR is old enough of mine that I'll just close it. Happy if anyone else takes it on.

gepcel:
I love this PR! Sad to see it closed. I actually thought you were on the Zed dev team. Kind of hopping this will be merged, and wondering why this takes so long time.

MrSubidubi:
I reviewed this partially in a private conversation and will share what was my concern here:

> Yeah I think we now have three ideas floating around there and probably need to pin it to one we commit on, right?
> 
> As mentioned, I am a bit reluctant on adding another set of queries, that said, could we have any advantages with doing that? The upside is not all languages would actually need it, but do we have cases where we might benefit from overlapping captures?
> 
> The list of capture names for a given language might suffice though perhaps? Do you have any strong feeling towards one of the solutions?

gepcel:
I got it that you have other factors to consider, though I haven't got your concern  without the context. Hope you will figure out a better solution. Not long ago, I submitted another proposal #56836, but looking back now, I might have been making a fuss about nothing.

rgbkrk:
> I actually thought you were on the Zed dev team.

I was sometime ago just not as full staff. I primarily do this for fun and love of Zed now. Sadly I don't think I'll be able to contribute at the level necessary and don't wish for people to wait on my PRs.

One follow on that I had in another branch also grouped statements that were together (separated by runs of newlines). It was incredible to use inside of Zed. At the time I wasn't sure what the best bet was to address @MrSubidubi's concerns (which are valid).

PR #49638, opened 2026-02-19
workspace: Add window title customization

mitchcapper:
new window_title_template setting

- [x ] Added a solid test coverage and/or screenshots from doing manual testing
- [x ] Done a self-review taking into account security and performance aspects
- [ x] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Added Window title customization through window_title_template setting

This adds window title customization using similar templating to the existing task code with ZED_* variables.   This should be in line with Zed’s contribution guidelines of “Small extra features, like keybindings or actions you miss from other editors or extensions.”  This allows for similar functionality to vscodes window.title (` "window.title": "${rootName}${separator}${dirty}${activeEditorShort}${separator}${appName}”` ) and allow for sublime style titles where untitled files still can change the window title.  This is useful for those who might also use zed as something of a multi-window scratch pad so it shows something rather than just “empty project”.

The biggest area for potential improvement is probably performance related.   I don’t think the current code does any large amounts of work and I tried to add some minor early pruning by only computing template variables that are contained in the string.   This customization shouldn’t change the default behavior as the default string `` should emulate the current window title fairly well.  

Note there are a good 50+ lines of this PR just around the ZED_SEPERATOR support.   If we were OK with a slightly degraded from current experience where when there wasn't a file open for example it would show "project - " with nothing after, then we could remove it.   The separator implementation here is implemented differently from vscode is a bit of a simpler fashion using sentinels rather than keeping the entire thing segments.   We are also abusing the bash default shorthand to allow the user to override the separator value (basically a reverse of the default option).  This is a bit more dynamic that VSCode that just allows setting the separator as a global setting vs here where you could have different separators for different parts.

image

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -194,6 +194,12 @@
   //
   // Default: "client"
   "window_decorations": "client",
+  // A template string for the window title.
+  // Available variables: $ZED_TAB_TITLE, $ZED_PROJECTS, $ZED_COLLAB,
+  // $ZED_FILE, $ZED_FILENAME, $ZED_STEM, $ZED_DIRNAME,
+  // $ZED_RELATIVE_FILE, $ZED_RELATIVE_DIR, $ZED_WORKTREE_ROOT
+  // Use ${VAR:default} for fallback values.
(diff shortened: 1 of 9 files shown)


SomeoneToIgnore:
Thank you, will continue with https://github.com/zed-industries/zed/pull/54379

PR #49664, opened 2026-02-19
editor: Fix crash from stale state after prepaint depth exhaustion

### LINKED ISSUE #49662: Crash: index out of bounds in EditorElement::prepaint (multiple unfixed sites)
### Summary

Zed Preview 0.225.2 crashes with `panic_bounds_check` (index out of bounds) inside `EditorElement::prepaint` during rendering. This is the same class of bug as #30191, which was partially fixed by #32953, but several other unsafe indexing sites remain.

### Crash Report

- **Version**: Zed Preview 0.225.2 (20260219.190736)
- **OS**: macOS 26.3 (25D5112c), ARM-64
- **Exception**: EXC_CRASH (SIGABRT) — `abort() called` via `core::panicking::panic_bounds_check`
- **Thread**: 0 (main thread), during `Window::draw` → `EditorElement::prepaint`

Key frames from the crash:
```
core::panicking::panic_bounds_check
<editor::element::EditorElement as gpui::element::Element>::prepaint::{{closure}}::{{closure}}::{{closure}}
```

The stack shows ~18 levels of recursion through `Div::prepaint`, indicating this happens during normal editor rendering.

### Root Cause

PR #32953 fixed ONE unsafe indexing site (inline blame layout), but there are at least 6 more sites in `element.rs` that use unchecked `[]` indexing into `line_layouts`, `crease_trailers`, `row_infos`, or `per_row_map` with computed offsets like `row.minus(start_row) as usize`. These can panic when the index exceeds the slice length, which occurs when editor state (selections, diagnostics, hover positions, diff hunks) references rows that are technically in the visible range but whose computed index exceeds the actual slice length due to timing/state inconsistencies.

Remaining unsafe sites:
1. `layout_inline_diagnostics` — `crease_trailers[window_ix]` / `line_layouts[window_ix]`
2. `layout_visible_cursors` — `line_layouts[cursor_position.row().minus(...)]`
3. `layout_cursor_popovers` — `line_layouts[cursor.row().minus(start_row)]`
4. `layout_hover_popovers` — `line_layouts[popover_position.row().minus(...)]`
5. `bg_segments_per_row` — `per_row_map[row.minus(rows.start)]` (only `debug_assert`, panics in release)
6. `layout_diff_hunk_controls` — `row_infos[row_ix]`
7. `paint_highlighted_range` — `line_layouts[row.minus(start_row)]`

Zed: v0.225.2 (Zed Preview)
OS: macOS 26.3 (25D5112c)
Memory: 36 GB
Architecture: aarch64 (Mac15,6 / Apple M3 Pro)

jean-humann:
Closes #49662

## Summary

Fixes an index-out-of-bounds crash in `EditorElement::prepaint` that occurs when block decorations oscillate in size, exhausting the recursive prepaint budget (`MAX_PREPAINT_DEPTH = 5`).

## Root cause

When block decorations resize during `render_blocks`, prepaint recurses to rebuild layout state. Previously, both `resize_blocks()` and `update_renderer_widths()` mutated the display map **before** checking `can_prepaint()`. At max depth:

1. `resize_blocks()` mutated the block map, changing the display row mapping
2. `can_prepaint()` returned false, skipping the recursive rebuild
3. All local state (`snapshot`, `start_row`, `end_row`, `line_layouts`, `row_infos`) remained stale
4. Downstream code — particularly `layout_inline_diagnostics` which takes a **fresh** `editor.snapshot()` — would see the new row mapping but index into the stale `line_layouts`
5. `row.minus(start_row)` produced an index exceeding `line_layouts.len()` → panic

The crash is most likely to occur when many block decorations appear or resize simultaneously (e.g., unfolding a folder containing many git repos, causing a burst of diff hunk blocks), or when `place_near` blocks oscillate between inline (height=0) and block (height>=1) mode, burning through all 5 recursion levels without converging.

## Fix

Guard both mutation sites with `can_prepaint()` so the display map is NOT mutated when we cannot recurse to rebuild state:

- **Block resize**: move `resize_blocks()` inside the `can_prepaint()` guard. At max depth, defer the resize to the next frame via `cx.notify()`. The resize will be re-detected because the stored height still mismatches the rendered height.

https://github.com/jean-humann/zed/blob/021978ecf939723a6ba4ab9843572b6bcefe7cb7/crates/editor/src/element.rs#L10207-L10239

- **Renderer widths**: short-circuit `update_renderer_widths()` with `can_prepaint()` so fold widths are not updated at max depth.

https://github.com/jean-humann/zed/blob/021978ecf939723a6ba4ab9843572b6bcefe7cb7/crates/editor/src/element.rs#L10097-L10115

In both cases the next frame starts with a fresh recursion budget (`EditorRequestLayoutState::default()` resets `prepaint_depth` to 0) and applies the deferred changes normally. The worst case is one frame of slightly wrong-sized blocks — vastly preferable to a crash.

## Test plan

- [x] `cargo check -p editor` passes
- [ ] Manual testing with many open files, inline diagnostics, hover popovers, and git diff hunks

Release Notes:

- Fixed a crash (index out of bounds) during editor rendering when block decorations repeatedly resize, exhausting the recursive prepaint budget.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -10236,29 +10236,24 @@ impl Element for EditorElement {
                                 }
                             })
                     });
-                    if new_renderer_widths.is_some_and(|new_renderer_widths| {
-                        self.editor.update(cx, |editor, cx| {
-                            editor.update_renderer_widths(new_renderer_widths, cx)
-                        })
-                    }) {
-                        // If the fold widths have changed, we need to prepaint
-                        // the element again to account for any changes in
-                        // wrapping.
-                        if request_layout.can_prepaint() {
-                            return self.prepaint(
-                                None,
-                                _inspector_id,
-                                bounds,
-                                request_layout,
-                                window,
-                                cx,
-                            );
-                        } else {
-                            debug_panic!(concat!(
-                                "skipping recursive prepaint at max depth. ",
-                                "renderer widths may be stale."
-                            ));
-                        }
+                    // Only update renderer widths if we can recurse to rebuild
+                    // line_layouts afterwards. Mutating the fold map at max depth
+                    // would leave stale state that can cause index panics.
+                    let renderer_widths_changed = request_layout.can_prepaint()
+                        && new_renderer_widths.is_some_and(|new_renderer_widths| {
+                            self.editor.update(cx, |editor, cx| {
+                                editor.update_renderer_widths(new_renderer_widths, cx)
+                            })
+                        });
+                    if renderer_widths_changed {
+                        return self.prepaint(
+                            None,
+                            _inspector_id,
+                            bounds,
+                            request_layout,
+                            window,
+                            cx,
+                        );
                     }
 
                     let longest_line_blame_width = self
@@ -10374,14 +10369,14 @@ impl Element for EditorElement {
                         resized_blocks,
                     } = blocks;
                     if let Some(resized_blocks) = resized_blocks {
-                        self.editor.update(cx, |editor, cx| {
-                            editor.resize_blocks(
-                                resized_blocks,
-                                autoscroll_request.map(|(autoscroll, _)| autoscroll),
-                                cx,
-                            )
-                        });
                         if request_layout.can_prepaint() {
+                            self.editor.update(cx, |editor, cx| {
+                                editor.resize_blocks(
+                                    resized_blocks,
+                                    autoscroll_request.map(|(autoscroll, _)| autoscroll),
+                                    cx,
+                                )
+                            });
                             return self.prepaint(
                                 None,
                                 _inspector_id,
@@ -10391,10 +10386,17 @@ impl Element for EditorElement {
                                 cx,
                             );
                         } else {
-                            debug_panic!(concat!(
-                                "skipping recursive prepaint at max depth. ",
-                                "block layout may be stale."
-                            ));
+                            // Don't mutate the block map at max depth — doing so
+                            // would leave stale line_layouts/row_infos that don't
+                            // match the new row mapping, causing index panics
+                            // downstream (e.g. in layout_inline_diagnostics which
+                            // takes a fresh snapshot). The resize will be
+                            // re-detected on the next frame.
+                            debug_panic!(
+                                "deferring block resize to next frame \
+                                 (prepaint depth limit reached)"
+                            );
+                            self.editor.update(cx, |_, cx| cx.notify());
                         }
                     }
 
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -9523,7 +9523,7 @@ impl EditorRequestLayoutState {
         }
     }
 
-    fn can_prepaint(&self) -> bool {
+    fn has_remaining_prepaint_depth(&self) -> bool {
         self.prepaint_depth.get() < Self::MAX_PREPAINT_DEPTH
     }
 }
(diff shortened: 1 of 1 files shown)


jean-humann:
@cla-bot check

Veykril:
This doesn't feel right, any panic here is a bug. Replacing these with fallible fetches will likely lead to hard to debug visual bugs instead

jean-humann:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jean-humann:
@Veykril You're right — the original `.get()` approach was papering over the bug. I dug deeper and found the actual root cause.

**The crash path:**

At `MAX_PREPAINT_DEPTH` (5), `resize_blocks()` was called **before** `can_prepaint()`, mutating the block map (changing the display row mapping) while all local state (`line_layouts`, `row_infos`, `start_row`/`end_row`) remained stale. The code then fell through past the `debug_panic!` and continued rendering with this inconsistent state. In particular, `layout_inline_diagnostics` takes a **fresh** `editor.snapshot()` that sees the new row mapping, but indexes into the stale `line_layouts` — producing an out-of-bounds index.

**The fix (updated):**

I removed all the `.get()` band-aids and instead moved both `resize_blocks()` and `update_renderer_widths()` inside their respective `can_prepaint()` guards, so the display map is never mutated when we can't recurse to rebuild state. At max depth for block resize, `cx.notify()` defers the resize to the next frame (which starts with a fresh depth budget and re-detects the mismatch). The `debug_panic!` is preserved for the block resize case.

**What triggers it:**

Block decorations that oscillate in size — particularly `place_near` blocks that flip between inline (height=0) and block mode depending on available width — can burn through all 5 recursion levels without converging. The reporter's scenario (unfolding folders with many git repos, causing a burst of diff hunk blocks) is a plausible trigger.

crates/editor/src/element.rs
```diff
@@ -10255,10 +10222,17 @@ impl Element for EditorElement {
...
+                            // would leave stale line_layouts/row_infos that don't
+                            // match the new row mapping, causing index panics
+                            // downstream (e.g. in layout_inline_diagnostics which
+                            // takes a fresh snapshot). The resize will be
+                            // re-detected on the next frame.
+                            debug_panic!(
+                                "deferring block resize to next frame \
+                                 (prepaint depth limit reached)"
+                            );
+                            self.editor.update(cx, |_, cx| cx.notify());
```
smitbarmase: Actually, this is unnecessary. Calling `cx.notify` in prepaint will have no effect. We should simply remove it and allow this frame to render stale. At least after these changes, we won't panic because either the line layout and row mapping are fresh, or both are stale. The staleness will self-correct on the next repaint.

nit: We should fix comment too. Better just not have it, code is clear.

What I'm more interested in is when this recursion reaches a depth of five levels? However, that's something we should investigate apart from this.

crates/editor/src/element.rs
```diff
@@ -10122,29 +10094,24 @@ impl Element for EditorElement {
...
-                                window,
-                                cx,
-                            );
-                        } else {
-                            debug_panic!(concat!(
-                                "skipping recursive prepaint at max depth. ",
-                                "renderer widths may be stale."
-                            ));
-                        }
+                    // Only update renderer widths if we can recurse to rebuild
```
smitbarmase: nit: Can get rid of comment which address past panic, I think current code is clear.

smitbarmase (changes requested):

smitbarmase (approved):
I updated it. Thanks!

PR #49673, opened 2026-02-20
Sign out upon receiving an Unauthorized response when acquiring an LLM token

maxdeviant:
This PR makes it so the user gets signed out upon receiving an Unauthorized response when acquiring an LLM token.

This is a re-landing of #49661.

Closes CLO-324.

Release Notes:

- N/A

### DIFF
--- a/crates/zed/src/zed.rs
+++ b/crates/zed/src/zed.rs
@@ -2833,7 +2833,7 @@ mod tests {
         assert!(window_is_edited(window, cx));
 
         // Advance the clock to make sure the workspace is serialized
-        cx.executor().advance_clock(Duration::from_secs(1));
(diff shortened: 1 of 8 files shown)


crates/zed/src/zed.rs
```diff
@@ -2833,7 +2833,7 @@ mod tests {
         assert!(window_is_edited(window, cx));
 
         // Advance the clock to make sure the workspace is serialized
-        cx.executor().advance_clock(Duration::from_secs(1));
+        cx.executor().advance_clock(Duration::from_secs(2));
```
maxdeviant: This change was needed to get the test passing with the rest of the changes.

This test seems incredibly brittle, as currently written.

PR #49679, opened 2026-02-20
extension: Stream wasi-sdk download

cppcoffee:
Release Notes:

- N/A

The WASI-SDK is several hundred megabytes in size; using streaming downloads is network-friendly and can reduce memory usage.


Veykril (approved):
Thanks!

PR #49714, opened 2026-02-20
Add fast mode toggle in agent panel

tomhoule:
This is a staff only toggle for now, since the consequences of activating it are not obvious and quite dire (tokens costs 6 times more).

Also, persist thinking, thinking effort and fast mode in DbThread so the thinking mode toggle and thinking effort are persisted.

Release Notes:

- Agent: The thinking mode toggle and thinking effort are now persisted when selecting a thread from history.


assets/keymaps/default-windows.json
```diff
@@ -335,6 +335,7 @@
       "ctrl-alt-k": "agent::ToggleThinkingMode",
       "ctrl-alt-'": "agent::ToggleThinkingEffortMenu",
       "ctrl-'": "agent::CycleThinkingEffort",
+      "ctrl-alt-.": "agent::ToggleFastMode",
```
tomhoule, line 338: I see we render the tooltip for that shortcut, but it may not be wired up, right? I’ll look into that.

tomhoule: I checked. It is.

crates/language_models/src/provider/cloud.rs
```diff
@@ -571,6 +571,10 @@ impl LanguageModel for CloudLanguageModel {
         self.model.supports_thinking
     }
 
+    fn supports_fast_mode(&self) -> bool {
+        self.model.supports_fast_mode
```
tomhoule, line 575: This is for cloud. Do we also want this for opus 4.6 in BYOK?

maxdeviant: Don't think it's necessary right now, as I think most individual user's accounts won't have fast mode available.

maxdeviant (approved):

gb-jos:
@tomhoule @maxdeviant is there plans to make this not staff only at some point ? asking because i use fast mode with codex, and would love to move to zed and still use fast mode

tomhoule:
The feature works. I think the only thing we are missing is a mechanism to make sure that people know what they are opting into when they turn on fast mode, since it is much more expensive. So only this and prioritization basically. Do you have ideas?

gb-jos:
@tomhoule well, honestly i don't see this being any different than some models being more expensive than others. 

But that being said a good way to make sure people know what they get into, is to make it a dropdown than a toggle, which explicitly then states the cost as well, like codex does 

Screenshot 2026-05-15 at 13 09 31

morgankrey:
Fast mode is only available in the codex app and CLI, we can't offer it in the zed agent yet https://developers.openai.com/codex/speed

gb-jos:
@morgankrey my assumption was that when using the zed agent with chatgpt subscription, fast mode should work ?

morgankrey:
I'll certainly ask but my understanding is it's reserved for their first party surfaces

gb-jos:
@morgankrey okay thanks, i have been trying this in a local branch, adding `supports_fast_mode` and removing the staff check. Seemed to be working (well did not show any errors, had no other way to validate). But i completely agree that it makes sense to make sure its fine in terms of ToC. In this case i will also hold off from creating a PR.

Issue #38927, opened 2025-09-26
Find & Replace memory leak on large files


kclonts:
### Summary

Find & Replace on 1 million occurrences works, but when finished still has memory usage at over 3.2GB (and though nothing crashed, my computer was barely responsive)

open a large csv file, for example 200k rows & ~500MiB and lots of \" instead of "" for quoting. search & replace via zed for \" with ""

Being slow is fine (though it could certainly be much faster. Maybe over some size, stream changes to disk and write back?), it did manage to accomplish the task. 

However; the expectation after execution should see memory fall to pre search & replace

### Zed Version and System Specs

Zed: v0.205.5 (Zed) 
OS: macOS 26.0
Memory: 16 GiB
Architecture: aarch64

nia-e:
Hi @kclonts, thanks for the issue! Do you have a reproduction for the leak that you could pass along?

kclonts:
@nia-e I do not, my file cannot be shared. But I imagine that recreating a csv of this size should have the same issue.

nia-e:
That's understandable, I'll see if I can repro it. In the meantime could you attach your Zed logs from an execution which does reproduce this? From the command panel, `zed: open log`

kclonts:
Yeah, I'll try it out soon

kclonts:
Probably related, I just had it open for the past 30 minutes and it started eating more ram and became unresponsive. I had the original file open (no modifications being made) and another file which was the transformation of that initial file which was constantly being truncated and rewritten. Couldn't get a log, had to force close by killing the process.

kclonts:
Here's the relevant log after opening after the incident

<details><summary>Log</summary> 2025-09-26T12:11:21-04:00 INFO  [zed] ========== starting zed version 0.205.5, sha 39de188 ==========
2025-09-26T12:11:21-04:00 INFO  [crashes] spawning crash handler process
2025-09-26T12:11:21-04:00 INFO  [zed] Using git binary path: Some("/Applications/Zed.app/Contents/MacOS/git")
[...]

PR #49726, opened 2026-02-20
text: Store fragment's deletions on stack

marcocondrache:
Helps #38927

- In the common case, each fragment has at most one deletion, so we can skip the heap allocation
- SmallVec is smaller than an empty HashSet
- Makes cloning cheaper
- From my analysis of the code it's not possible to have duplicate ticks (I may be wrong tho)

Release Notes:

- N/A

### DIFF
--- a/crates/text/src/text.rs
+++ b/crates/text/src/text.rs
@@ -542,7 +542,7 @@ pub struct Fragment {
     pub insertion_offset: usize,
     pub len: usize,
     pub visible: bool,
-    pub deletions: SmallVec<[clock::Lamport; 1]>,
+    pub deletions: SmallVec<[clock::Lamport; 2]>,
     pub max_undos: clock::Global,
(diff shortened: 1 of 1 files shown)


crates/text/src/text.rs
```diff
@@ -541,7 +542,7 @@ pub struct Fragment {
     pub insertion_offset: usize,
     pub len: usize,
     pub visible: bool,
-    pub deletions: HashSet<clock::Lamport>,
+    pub deletions: SmallVec<[clock::Lamport; 1]>,
```
Veykril: Lamport is small enough that we can actually squeeze in 2 on the stack without increasing the smallvec size

marcocondrache: Nice!

Veykril (approved):
Nice, thanks!

PR #49731, opened 2026-02-20
git: Prevent panic when updating excerpts in split diff (cherry-pick #49725)

cole-miller:
When using the split view, if the project diff inserts excerpts for the same buffer using two different `PathKey`s, we panic inside `LhsEditor::update_path_excerpts_from_rhs` because of an implicit assumption that `excerpts_for_path()` and `excerpts_for_buffer()` return the same number of values. This PR addresses that by ignoring any excerpts for the given buffer that have a different path key from the provided one.

Release Notes:

- Fixed a crash when using the split diff view.


Alexander-vecos (approved):
Нихера не понял, я что сейчас поделился проектом который у кого-то списал и мне типа помогают его вместе с капилотом?)

PR #49738, opened 2026-02-20
Add `initial_directory` for save as action

### LINKED ISSUE #4734: Use current's file directory as default for the new file in Save As
### Description

When trying to duplicate a file is common that the new file goes in the same directory or very close to the current one.
As a fast way to duplicate it would be easier if when you press save as, the File name picker starts at the current's file directory instead of the root project directory.
In big projects sometimes it's hard/slow to find the current directory using the tree.

bobbymannino:
Closes #4734

Use the file path as the save as file path by default. This will save you searching for the current directory of the file you are already in.

**Testing**

New file save location remains root of project:

Screenshot 2026-02-20 at 18 20 08

If file is in root of project uses that anyway:

Screenshot 2026-02-20 at 18 19 57

If file is in sub folder will use that as the default save as path:

Screenshot 2026-02-20 at 18 19 46

Release Notes:

- Added default save as path to be the path of the file your saving


crates/open_path_prompt/src/open_path_prompt.rs
```diff
@@ -241,12 +251,42 @@ impl OpenPathPrompt {
...
+fn build_initial_query(
+    default_query: impl FnOnce() -> String,
+    suggested_name: Option<String>,
+    initial_directory: Option<PathBuf>,
+) -> String {
+    let mut query = if let Some(initial_dir) = initial_directory {
+        let mut dir_str = initial_dir.to_string_lossy().into_owned();
+        if !dir_str.ends_with(std::path::MAIN_SEPARATOR) {
+            dir_str.push(std::path::MAIN_SEPARATOR);
+        }
```
Veykril, line 281: This is wrong for remote projects where the host and guest differ in path separators, you need to source this from the `workspace.path_style()`

Veykril:
There is still an outstanding review

mikayla-maki:
Given that it's been over a month since the last interaction in here, I'm going to close this PR.

PR #49786, opened 2026-02-21
agent_ui: Fix send failure for pasted file#line mention links

cppcoffee:
Release Notes:

- Fixed send failure for pasted file#line mention links

Handle `MentionUri::Selection` with a file path in `confirm_mention_for_uri`
by resolving it to text content via existing line-range loading logic, instead
of returning "Unsupported mention URI type for paste".

Keep untitled-buffer selections (`abs_path: None`) as an explicit unsupported
error for now.

Add a regression test for pasting selection mentions to ensure they resolve to
`Mention::Text` with the expected line range.


cppcoffee:
before:

https://github.com/user-attachments/assets/9cdf7ec3-6cba-45d1-bd5e-14a0983ec07b

benbrandt (approved):
thanks!

PR #49796, opened 2026-02-21
agent_ui: Set history once we get connection not for each thread view

rabi:
Moves loading of the history connection to once we have a new connection, not on every thread view.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/connection_view.rs
+++ b/crates/agent_ui/src/connection_view.rs
@@ -694,6 +694,7 @@ impl ConnectionView {
                         }
 
                         let id = current.read(cx).thread.read(cx).session_id().clone();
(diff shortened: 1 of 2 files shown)


benbrandt (commented):
There were a fair number of unnecessary changes here, so I cleaned that up.

Also, I am not sure how this would help, as initial_state already called new_thread_view, so your original change would have only called the same thing again. However, I think it is more correct to initialize the history on initial state when we get the new connection rather than on every thread view, so that is a helpful change.

You also removed the feature gate for loading history, I was checking if the agent supports loading or resuming before showing the list, as a list of sessions isn't helpful without one of those. 

Any chance goose doesn't express a capability for one of those, causing the guard to fail?

PR #49802, opened 2026-02-21
project_panel: Add diagnostic count badges and color priority

davidescaccia:
Discussed in #6668 specifically this comment from @zackangelo:

> The biggest thing keeping me from using Zed as a daily driver is error indication in the project panel. When I'm making big project-wide changes I can't clearly see which files have errors (in editors like VSCode the filenames turn red).

> VSCode seems to use a letter on the right gutter to indicate git status and a number next to it to indicate diagnostic status. The color indicates either.

This PR implements that, I added an opt-in `diagnostic_badges` setting (default is false) that shows error and warning counts as colored labels on the right side of each project panel entry. Counts bubble up to parent directories.

When `diagnostic_badges` is enabled, diagnostic severity takes priority over git status for entry text color.
Since warnings and git-modified share the same yellow, git status with this option on is readable through the file icon decoration and the absence of a number badge on the right.

Example:
image

image

Release Notes:
- Added diagnostic count badges to the project panel, displaying error and warning counts next to file names. You can modify this setting using the `diagnostic_badges` option, which is enabled by default.


crates/project_panel/src/project_panel.rs
```diff
@@ -1000,35 +1003,66 @@ impl ProjectPanel {
         let mut diagnostics: HashMap<(WorktreeId, Arc<RelPath>), DiagnosticSeverity> =
             Default::default();
         let show_diagnostics_setting = ProjectPanelSettings::get_global(cx).show_diagnostics;
```
smitbarmase, line 1029: I don't think we need to accumulate diagnostic counts on parent directories. The existing icon decoration already tells you "something in here has diagnostics," and that's sufficient. The actual count on a directory doesn't help much because you still need to drill into files to fix them. I think its just adds visual noise, especially in deep trees where every ancestor ends up with a big number. 

I'd scope `diagnostic_counts` to files only and skip the ancestor propagation.

crates/project_panel/src/project_panel.rs
```diff
@@ -5482,22 +5517,55 @@ impl ProjectPanel {
...
+                                            .size(IconSize::Indicator)
+                                            .color(filename_text_color),
+                                    )
+                            });
+                            this.end_slot::<AnyElement>(
+                                h_flex()
+                                    .gap_1()
+                                    .flex_none()
+                                    .pr_3()
+                                    .when_some(diagnostic_counts, |this, (errors, warnings)| {
```
smitbarmase: We should cap the displayed count so it doesn't take up too much horizontal space. Something like 99+ for anything over 99.

crates/project_panel/src/project_panel.rs
```diff
@@ -5907,8 +5975,21 @@ impl ProjectPanel {
...
 
-        let filename_text_color =
-            entry_git_aware_label_color(git_status, entry.is_ignored, is_marked);
+        let diagnostic_counts = self
+            .diagnostic_counts
+            .get(&(worktree_id, entry.path.clone()))
+            .copied();
+
+        let settings = ProjectPanelSettings::get_global(cx);
+        let filename_text_color = if settings.diagnostic_badges {
```
smitbarmase: I think we should drop the filename color override from this PR. Right now git status is communicated solely through filename text color, there's no other indicator for it. If we override that with diagnostic severity, the user loses git status information entirely.

The right path would be: a separate PR adds git status letter indicators (M, U, A, etc.) to the project panel, and then a follow-up can add the color override since git status would still be visible through the letters. But until that exists, we shouldn't remove the only git status signal we have.

smitbarmase (changes requested):
Thanks, this is a great addition. I left some comments.

smitbarmase (approved):
Thanks! I pushed a commit on top with some cleanup and defaulted it to on, since this is nice addition. 

One behavior change I made is badge counts now respect the `show_diagnostics` setting, so warnings won't show when it's set to "errors".

No pressure on the git status indicators, take your time whenever you get to it.

PR #49812, opened 2026-02-21
elixir: Fix `HEEx` name and add `EEx` defaults

AlternateRT:
- zed-extensions/elixir#61 renames `HEEX` to `HEEx`; to prevent any disruptions from that change, this adds the correct name alongside a settings migration
- zed-extensions/elixir#101 adds support for `EEx` templates; this sets the default language server to be the same as that of `Elixir` and `HEEx`

This also adds a few more extensions that should be recognized as `HEEx` files

Release Notes:

- N/A


MrSubidubi (approved):
Nice, thanks and thank you for your continued efforts!

PR #49825, opened 2026-02-22
terminal: Fix drag-and-drop in vertical terminal panels

### LINKED ISSUE #49800: Drag & drop file in terminal to print path partially works
### Reproduction steps

1. Open a terminal in Zed.
2. Drag and drop a file onto the terminal — the file path is printed as expected.
3. Move the terminal to a vertical split panel (e.g., drag the terminal tab to the right side).
4. Drag and drop a file onto the terminal again — instead of printing the path, Zed opens the file in the editor.

**Expected:** File path is printed in the terminal regardless of which panel it's in.

**Actual:** After moving the terminal to a vertical panel, drag-and-drop opens the file in Zed instead of pasting the path into the terminal.

Zed 0.224.11

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

transitoryangel:
Closes #49800

Adds `handle_drop` to Item & ItemHandle, which allows an active item in a pane to consume drop events before the pane does. 

Release Notes:

- terminal: Fix drag-and-drop not working in vertical terminal panels


crates/workspace/src/pane.rs
```diff
@@ -3938,11 +3960,18 @@ impl Pane {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
+        if let Some(active_item) = self.active_item()
```
SomeoneToIgnore, line 3934: Custom drop handle was introduced for terminal pane needs, so seems that now it should be removed as this branch is executed always instead?

crates/workspace/src/item.rs
```diff
@@ -1098,6 +1105,12 @@ impl<T: Item> ItemHandle for Entity<T> {
         })
     }
 
+    /// Called when the containing pane receives a drop.
```
SomeoneToIgnore:
```suggestion
    /// Called when the containing pane receives a drop on the item.
```

Seems that there should be more conditions?
Same above.

SomeoneToIgnore (changes requested):
Seems that we do not swap items when drag and drop on their tab:

https://github.com/user-attachments/assets/d608965b-9b11-43da-9114-6d95a6804e71

IIRC that tab swapping code is in the `pane.rs` and the terminal tab as to skip the drop processing.

It would be good to get all such things tested beforehand, as definitely saves time on iterating.

--------------

Overall works, nice, we need a few more tweaks to make it good and merge.

crates/workspace/src/pane.rs
transitoryangel: Custom drop handles are used in debugger panes [here](https://github.com/zed-industries/zed/blob/main/crates/debugger_ui/src/session/running.rs#L314) as well. There is also the remaining use for custom drop handles in the terminal panel for handling dragged tabs - my concern with this is that moving this code is trivial, but it would hit GPUI's double lease panic when reading pane state [here](https://github.com/zed-industries/zed/blob/main/crates/terminal_view/src/terminal_panel.rs#L1266-L1282). How would you suggest I go about dealing with the terminal state?

SomeoneToIgnore: We could add an `active_pane: &Pane` argument into `handle_drop`'s signature, would that help?

transitoryangel: Seems reasonable enough!

crates/workspace/src/pane.rs
```diff
@@ -6412,7 +6405,7 @@ mod tests {
                 detail: 0,
                 is_active: true,
             };
-            pane.handle_tab_drop(&dragged_tab, 0, window, cx);
+            pane.handle_tab_drop(&dragged_tab, 0, true, window, cx);
```
SomeoneToIgnore: This new parameter is super peculiar:

* no tests added for `false`
* there are ~5 places where the thing is `false`: all non-test `.handle_tab_drop` invocations instead of 
```rs
.on_drop(cx.listener(move |this, dragged_tab, window, cx| {
    this.handle_tab_drop(
        dragged_tab,
        this.active_item_index(),
        true,
        window,
        cx,
    )
}))
```
one in the `pane.rs`

Can we somehow remove this parameter entirely?

We do know when we drag the tab around, and we do know what are we dropping it onto, don't we?
What info do we miss to make a decision before `handle_tab_drop`/another method that we're dealing with tabs and ready to swap them?

If not possible, would be great to clarify why this particular bit does not need to use `false` as the rest + get the tests on it all with the `false` parameter, but I somehow feel this is not the right way?

SomeoneToIgnore (commented):
Great, I see things start to work for the terminal and its tabs!

I have not checked the debugger code changes well enough yet, but let's deal with the general architecture first and then I can come back there — have posted a question below.

--------------

One remaining "bug" to consider is the fact that we can drag a tab/file around the terminal tab and it will show different dimming, as if that thing could split the terminal — yet all that happens is that the terminal gets that tab/file's path pasted.

What feels more appropriate is either do a split, or avoid drawing that "split possibility" entirely, now it's quite odd:

https://github.com/user-attachments/assets/6e28d9dc-1edc-451b-9f80-9f0fcc5fabef

Before, we did not have such a mixture of types in the same pane, but now all the edge cases jump at us at once and we'd better cover them all.

crates/workspace/src/pane.rs
transitoryangel: Yeah, that's fair - after more looking around, it seems like tabs dropped onto other tabs can be handled exclusively in handle_tab_drop without calling out to the item-specific drop methods?

SomeoneToIgnore: I hope so, but did not dig into the code to know for sure.

transitoryangel: So, disregard my previous comment. from my understanding:
The current behaviour is that when you drop a file tab into the terminal pane, the file's path gets put into the terminal.

if the item-specific drop method is not called in `handle_tab_drop` at all, then that behaviour stops. The current problem is that `handle_tab_drop` is called from both the drag overlay in Pane's rendering and the drag overlay in tab rendering, with no way to tell which is which without the parameter.

Regarding why there is only one call outside of tests where that parameter is `true` - everywhere else it is called is in a drag overlay for tab rendering. The one call site where it is `true` is in the `Pane` drag overlay rendering, hence the parameter name `is_pane_target`. The wording could be made more clear on that if necessary. All of the tests being `true` was a mistake made on behalf of me being very tired so I went ahead and updated those to be correct (to my understanding from reading the tests themselves).

I think passing the argument to `handle_drop` itself is unnecessary - so I will remove it. if the target is not the pane itself, we can skip calling it entirely to maintain the current tab-swap behaviour. 

If you're in agreement with keeping the parameter, I can go ahead and write some tests that validate the behaviour.

claiwe:
> One remaining "bug" to consider is the fact that we can drag a tab/file around the terminal tab and it will show different dimming, as if that thing could split the terminal — yet all that happens is that the terminal gets that tab/file's path pasted.

Regarding this - I think keeping the regular terminal pane's splitting behaviour (only split for other terminal items under certain circumstances) is favourable - but this runs into the same issue originally causing the dropping inconsistencies where panes created by splits don't have the same characteristics (in this case, the `can_split_predicate`).

And again, it seems like the `can_split_predicate` is only used by the debugger UI and the terminal view. So, my instinct is to go ahead and make another per-Item method to handle splits. Does this sound fine?

crates/workspace/src/pane.rs
SomeoneToIgnore: Nice, thank you for finding a way to remove those `bool` out of the public, items' API.

Would be superb to add a test or two indeed, thank you for considering this.

SomeoneToIgnore:
> make another per-Item method to handle splits

I guess I might miss something, but from the bird's view, this adjustment seems odd: it's the pane/pane overlay that knows whether we drop on the whole item or on its part?
If the former, then we're supposed to call item's drop handler; if the latter — we're not supposed to call anything like that at all but rather create another element?
If this is true, no extra methods is needed at all, just the pane has to get more logic added?

If I'm mistaken somehow, I wonder whether we can consider adding a parameter like "dropping into full item/dropping into part of the item" that will be handled differently by each drop handler.
Yet, I am not fond of the direction this "new parameter"/"new method" approach moves to, so if this indeed the case, it would be great to just understand what's causing issues here — and maybe it's fine to leave things as is now, as this all is definitely a very nice improvement already.
Later, we can work on a follow-up PR to fix this bug separately, to avoid bloating the current PR more.

SomeoneToIgnore (approved):
Thanks a lot for all this work!

PR #49834, opened 2026-02-22
recent_projects: Fix remote reconnect when server is not running

mangas:
Closes https://github.com/zed-industries/zed/issues/49363

When an existing remote workspace's connection is dead (e.g. the server died and reconnect failed, leaving the client in `ServerNotRunning` state), `remote_connection()` returns `None`. Previously this caused an error dialog, blocking reconnection — the user had to manually switch to another project and back to recover.

Now the code falls through to establish a fresh connection instead, matching the previous behavior where clicking "Reconnect" would just work.

Release Notes:

- Fixed remote reconnect failing with an error when the server is not running, now establishes a fresh connection instead.

### DIFF
--- a/crates/recent_projects/src/remote_connections.rs
+++ b/crates/recent_projects/src/remote_connections.rs
@@ -144,57 +144,62 @@ pub async fn open_remote_project(
     .await;
 
     if let Some((existing_window, existing_workspace)) = existing {
-        let remote_connection = cx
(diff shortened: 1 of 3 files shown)


mangas:
@cla-bot check

mangas:
@cla-bot check

Veykril (approved):
Thanks!

PR #49841, opened 2026-02-22
git_ui: Disable remote buttons while push/pull/fetch is in progress

virajbhartiya:
## Summary

- Track pending remote operations (push, pull, fetch) with a `pending_remote_operation` task field on `GitPanel`
- While an operation is running, the remote button (Push/Pull/Fetch/Publish/Republish) is rendered in a disabled state
- New operations are rejected while one is already in progress, preventing duplicate operations from accidental repeated clicks

Release Notes:

- Fixed git panel push/pull/fetch buttons not showing a disabled state while the operation is in progress, which could lead to accidentally triggering duplicate operations.

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -623,6 +623,7 @@ pub struct GitPanel {
     changes_count: usize,
     new_staged_count: usize,
     pending_commit: Option<Task<()>>,
+    pending_remote_operation: Option<Task<anyhow::Result<()>>>,
     amend_pending: bool,
     original_commit_message: Option<String>,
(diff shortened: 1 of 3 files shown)


ChristopherBiscardi:
Heya, thanks for the PR!

For this one the concern seems to be duplicate operations. However fetching and pushing are idempotent, and the merge after a pull would recognize there was say, a merge conflict, and be handled on git's side, so this doesn't seem like an issue we need to fix. Thus the overhead to track state to disable the UI doesn't provide enough value to justify tracking and maintaining that state, which can add more complexity for future features.

As a result I'm going to close this one.

PR #49843, opened 2026-02-22
editor: Add gutter checkboxes for split diff and restore hunk to context menu

Dnreikronos:
Closes #49562

**Screenshots:**
- Before
image

- After
image

**Motivation**

In split diff view, the floating Stage/Unstage/Restore buttons overlay the code on the hovered line, hiding the content you're trying to compare. This defeats the purpose of a side-by-side diff (GitHub #49562).

**Changes**

Gutter checkboxes in split diff mode:
- Small checkboxes in the gutter next to each diff hunk in split diff view
- Checked = staged, unchecked = unstaged, dash = partially staged
- Clicking toggles staging state without covering any code
- Disabled during pending operations to prevent double-clicks

Floating toolbar skipped in split diff mode:
- The `layout_diff_hunk_controls` loop skips calling `render_diff_hunk_controls` entirely when `split_side` is set
- Navigation arrows are not needed because split diff always expands all hunks (`set_all_diff_hunks_expanded` in `split.rs`)
- Non-split editor behavior is unchanged

"Restore Hunk" added to right-click context menu:
- Available when a git repo is present and the cursor is inside a diff hunk
- Grayed out when the cursor is outside a hunk
- Works in both split and regular editor modes

**Files changed**

- `crates/editor/src/element.rs`: new `layout_staging_checkboxes` method, painting, checkbox import
- `crates/editor/src/element.rs`: skip `render_diff_hunk_controls` call when in split mode
- `crates/editor/src/mouse_context_menu.rs`: "Restore Hunk" menu item
- `crates/agent_ui/src/agent_diff.rs`: updated closure param count
- `crates/git_ui/src/{text_diff_view,project_diff,multi_diff_view,file_diff_view}.rs`: updated closure param count

**Risk**

Low. All changes gated behind `split_side.is_some()`. No behavioral change to regular editor, unified diff view, or agent diff view. The `RenderDiffHunkControlsFn` signature change is breaking but all call sites are updated.

Release Notes:

- Added gutter checkboxes for staging hunks in split diff view and "Restore Hunk" to the right-click context menu

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -29607,6 +29607,8 @@ fn render_diff_hunk_controls(
     _window: &mut Window,
     cx: &mut App,
 ) -> AnyElement {
+    let all_hunks_expanded = editor.read(cx).buffer().read(cx).all_diff_hunks_expanded();
+
     h_flex()
         .h(line_height)
         .mr_1()
@@ -29618,7 +29620,6 @@ fn render_diff_hunk_controls(
         .border_color(cx.theme().colors().border_variant)
         .rounded_b_lg()
         .bg(cx.theme().colors().editor_background)
-        .gap_1()
         .block_mouse_except_scroll()
         .shadow_md()
         .child(if status.has_secondary_hunk() {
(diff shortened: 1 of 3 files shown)


ChristopherBiscardi:
The buttons on main are rendered on-hover, which covers the original request here.

PR #49846, opened 2026-02-22
Fix OpenAI-compatible setting description

Marcuss2:
While testing the OpenAI compatible endpoint, it didn't work, so after a bit of digging I found out it actually expects the completions endpoint instead. So I changed the description accordingly.

Maybe this is a bug in the implementation instead and the description should be correct.

---

Release Notes:

- AI: Fixed the OpenAI-compatible setting description to point to the server's completion API URL.


danilo-leal (approved):
Good catch, thank you!

PR #49852, opened 2026-02-22
agent: Add ability to pin threads

RostiMelk:
Closes https://github.com/zed-industries/zed/discussions/43860

It can be hard to find your way back to threads you want to revisit. Generated titles don't always help, and ongoing work gets buried in history over time. This adds pin/unpin support so you can mark threads you care about and quickly return to them with the existing conversation context.

You can pin a thread from:
- The options menu (⋯) while viewing a thread ("Pin Thread" / "Unpin Thread")
- Hovering over entries in the history view or the new thread page

Pinned threads appear at the top of the history list under a "Pinned" section, and are blended into the "Recent" list on the new thread page with a small pin icon. Pin state is stored in the threads database and preserved across thread saves.

### New thread page
Pinned threads sort to the top of the "Recent" list with a pin icon.

New thread page

### History view
Pinned threads get their own "Pinned" section at the top. Hover to pin/unpin or delete.

History view

### A note on the `save_thread` query change

The existing `save_thread` used `INSERT OR REPLACE`, which in SQLite means "delete the row, then insert a new one." That was fine when every column was included in the INSERT, but `pinned` is managed separately from thread content, so a plain save would delete the row and reset `pinned` back to false.

The fix is `INSERT ... ON CONFLICT(id) DO UPDATE SET`, which updates only the listed columns and leaves everything else untouched. This pattern is already used in `component_preview/src/persistence.rs`.

An alternative would be to add `pinned` to the `DbThread` struct and include it in the save path, but pinning is user-level metadata, not thread content. Keeping it separate means future metadata columns (archived, color labels, etc.) would just work without touching the save path.

### Open questions

- **Branch/repo metadata on threads:** Another approach to thread discoverability would be attaching metadata like the current branch or repo to a thread, so you could filter or find threads by what you were working on. That feels complementary to pinning rather than a replacement. Pinning is manual and intentional, branch metadata would be automatic and contextual.
- **Pin/unpin icon size:** The pin and unpin SVGs in the icon library feel noticeably smaller than peers like the trash icon. Not something this PR can fix, but worth looking at on the design side.

---

Release Notes:

- Added ability to pin agent threads so they stay at the top of history and the new thread page

### DIFF
--- a/crates/acp_thread/src/connection.rs
+++ b/crates/acp_thread/src/connection.rs
@@ -241,6 +241,7 @@ pub struct AgentSessionInfo {
     pub title: Option<SharedString>,
     pub updated_at: Option<DateTime<Utc>>,
     pub meta: Option<acp::Meta>,
+    pub pinned: bool,
 }
 
 impl AgentSessionInfo {
@@ -251,6 +252,7 @@ impl AgentSessionInfo {
             title: None,
             updated_at: None,
             meta: None,
+            pinned: false,
         }
     }
 }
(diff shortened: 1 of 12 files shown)


RostiMelk:
@cla-bot check

RostiMelk:
@danilo-leal is this something you'd be interrested in helping me land? If so I can go ahead and rebase

angristan:
Would love to be able to pin threads!

PR #49906, opened 2026-02-23
gpui: Call `ack_configure` on throttled Wayland resize events

Albab-Hasan:
when a resizing configure event is throttled (skipped to limit redraws to once per vblank) the xdg_surface serial was never acknowledged. the early return bypassed the ack_configure call that follows the throttle check, violating the xdg-shell protocol requirement that every configure serial be acknowledged before the next surface commit.

fix by calling ack_configure before the early return so the compositor always receives the acknowledgment, dosent matter whether the resize itself is applied.

Release Notes:

- N/A


smitbarmase (approved):
Thanks!

PR #49923, opened 2026-02-23
vim: Bring highlight behavior inline with editor

### LINKED ISSUE #34793: vim: Search highlights should persist after pressing Escape

feitreim:
Closes #34793

I implemented the behavior described in #48114 
 https://github.com/zed-industries/zed/pull/48114#pullrequestreview-3739224174

## Summary
- Added a new option to the vim keymap, 
```
{
    "context": "(vim_mode == normal || vim_mode == helix_normal) && !menu",
    "bindings": {
      "escape": "editor::Cancel", --> "escape": "vim::Cancel",
      ...,
      }
}
```
This is not enabled by default however if you enable it you get:
  - Keep highlights after first `escape` in normal mode
  - Clear retained highlights on second `escape`

## Testing
I manually verified all behaviors described, that is:
default behavior is left unchanged, which is described as a bug in #34793 but later decided to be the correct UX decision due to users already being used to how it functions.
With the new vim::Cancel option set, pressing esc w/ text highlighted will only close the search and replace bar, leaving text highlighted, then a second esc press will clear the highlight.

Also added a Vim test (Not Neovim backend) that tests the new functionality.

Video: 

https://github.com/user-attachments/assets/8718d531-4f78-4c08-9c39-b29f2fcc8a07

description: I type 'hi' on 3 lines, then select with '*', highlighting all 3 'hi''s and bringing up the find and replace menu. Next, I press 'esc' which is bound to `vim::Cancel` thus removing the find and replace menu, while highlights remain. I press 'esc' again removing the highlight.

---

Release Notes:

- added `vim::Cancel` option that allows users to preserve highlights until triggered a second time, when it will clear the highlights. This brings the UX more inline with vim and Zed's normal editor.

### DIFF
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -390,7 +390,7 @@
   {
     "context": "(vim_mode == normal || vim_mode == helix_normal) && !menu",
     "bindings": {
-      "escape": "vim::Cancel",
+      "escape": "editor::Cancel",
       "shift-d": "vim::DeleteToEndOfLine",
       "shift-j": "vim::JoinLines",
       "shift-y": "vim::YankLine",
--- a/assets/keymaps/vim.json
+++ b/assets/keymaps/vim.json
@@ -390,7 +390,7 @@
   {
     "context": "(vim_mode == normal || vim_mode == helix_normal) && !menu",
     "bindings": {
-      "escape": "editor::Cancel",
+      "escape": "vim::Cancel",
       "shift-d": "vim::DeleteToEndOfLine",
       "shift-j": "vim::JoinLines",
       "shift-y": "vim::YankLine",
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -86,7 +86,6 @@ pub struct BufferSearchBar {
     configured_options: SearchOptions,
     query_error: Option<String>,
     dismissed: bool,
-    cancel_suppressed: bool,
     search_history: SearchHistory,
     search_history_cursor: SearchHistoryCursor,
     replace_enabled: bool,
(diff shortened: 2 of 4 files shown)


dinocosta (changes requested):
Hey @feitreim 👋 

Thank you for your efforts on this. The approach in this Pull Request aligns more closely with the original comment, and I've had the chance to review the code and confirm that it works as intended. 🙂

I believe we can still enhance the solution, so it would be great to explore the following:

* I don't think we need a new action (`vim::Cancel`). Instead, we can have Vim intercept `editor::Cancel` to dismiss the search bar or clear the highlights. 
    * This change will make keeping the highlights the default behavior, which is acceptable as it aligns with Neovim's functionality. I'll seek a second opinion on this.
* Since the `BufferSearchBar.dismiss` method dismisses the search bar and clears highlights, we should consider introducing a `BufferSearchBar.dismiss_preserving_highlights` method.
    * This new method will likely share much of the logic with the `dismiss` method, so any duplicate code should be moved to a shared method or function.
    * We could implement a `dismiss_inner` method that takes a `dismiss_highlights: bool` parameter to determine whether to clear the highlights.
* If we introduce `dismiss_preserving_highlights`, I believe we will also need a `BufferSearchBar.clear_all_matches` method, which will be called the second time `escape` is pressed.

Let me know if you need help or want to collaborate on this. I'll check with the team to see if we actually want to implement this change and will inform you if that's not the case. Thanks!

PR #49930, opened 2026-02-23
Fix panic in breadcrumbs

ConradIrwin:
The hypothesis is that some language's tree-sitter queries can return tokens containing newlines; though we weren't able to reproduce easily in a test.

Fixes https://zed-dev.sentry.io/issues/7286512714/?project=4509715135987712&query=is%3Aunresolved&referrer=issue-stream

Release Notes:

- Fix (rare) panic when breadcrumbs contain newlines (hopefully)

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -7835,7 +7835,7 @@ pub fn render_breadcrumb_text(
             return styled_element;
         }
 
-        StyledText::new(segment.text.replace('\n', "⏎"))
+        StyledText::new(segment.text.replace('\n', " "))
             .with_default_highlights(&text_style, segment.highlights.unwrap_or_default())
             .into_any()
     });


crates/editor/src/element.rs
```diff
@@ -7835,7 +7835,7 @@ pub fn render_breadcrumb_text(
             return styled_element;
         }
 
-        StyledText::new(segment.text.replace('\n', "⏎"))
+        StyledText::new(segment.text.replace('\n', " "))
```
Veykril, line 7838: Isnt the cause for the panic here the fact that we switch out a 1 byte character with a 3 byte character, causing the `highlights` if they exist to be incorrect as their offsets are now wrong?

ConradIrwin: Yes, so using a single byte character instead fixes it.

It seems like newlines in breadcrumbs are rare, so it didn’t seem worth rebasing the highlights to keep the fancy newline

Veykril: Ah yea makes sense

PR #49932, opened 2026-02-23
repl: Dynamically adapt terminal to viewport size

### LINKED ISSUE #16934: REPL output is truncated
### Check for existing issues

- [X] Completed

Beginning of the REPL output is truncated. The issue remains if I make the Zed window wider and re-evaluate.

Left: Kitty Terminal Right: Zed
Screenshot 2024-08-27 at 14 40 54

### Environment

Zed: v0.150.3 (Zed Preview)
OS: macOS 14.6.1
Memory: 32 GiB
Architecture: x86_64

_No response_

<details><summary>Zed.log</summary><pre>

</pre></details>

rgbkrk:
This brings a full blown scrollable terminal to REPL output in Zed. We've already got a terminal here. It just wasn't being exposed.

https://github.com/user-attachments/assets/6a2e8b59-dbc8-467f-88d5-f7266ad344e2

This now renders all the grid lines starting from the topmost history line instead of using `display_iter`. In addition, the terminal columns are computed from the window viewport width. Users can still set a fixed setting for the column if they desire. However, adapting to the viewport width allows content to just reflow. It's pretty nice.

This changes `max_columns` default to 0, which is interpreted as "auto". Positive values act as a fixed override.

The terminal component now has scroll behavior when there are more than `max_lines`. To eliminate nested scroll problems (with the editor window), you have to click in the terminal view to scroll it. You can click outside to unfocus that terminal component or hit escape.

tracebackpretty

Closes #16934

Release Notes:

- repl: Improved terminal rendering for big outputs


crates/repl/src/outputs/plain.rs
```diff
@@ -52,6 +57,11 @@ pub struct TerminalOutput {
     parser: Processor,
     /// Alacritty terminal instance that manages the terminal state and content.
     handler: alacritty_terminal::Term<VoidListener>,
+    scroll_handle: ScrollHandle,
+    /// Created lazily on first render (requires `Context<Self>`).
+    focus_handle: Option<FocusHandle>,
```
MrSubidubi: You can create this as well as the subscription already on creating the TerminalOutput, since `cx.new` already provides you the context to your entity. You'd need to change https://github.com/zed-industries/zed/blob/2d0bd7cd04ab7d03428b9cf03a476cefe5f7e08e/crates/repl/src/outputs/plain.rs#L170 to take the context then instead, as currently the context derefs to `App`

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +353,298 @@ mod tests {
...
+    fn test_focus_handle_not_created_until_render(cx: &mut TestAppContext) {
+        let cx = init_test(cx);
+        cx.update(|window, cx| {
+            let output = TerminalOutput::new(window, cx);
+            assert!(
+                output.focus_handle.is_none(),
+                "focus_handle should be None before first render"
+            );
+        });
+    }
```
MrSubidubi: This then should not be needed as per the comment above

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +353,298 @@ mod tests {
...
+
+        // If no max height is configured or the content fits, render the canvas directly.
+        let content_overflows = output_max_height
+            .map(|max_h| content_height > max_h)
+            .unwrap_or(false);
+        if !content_overflows {
+            return canvas_element.into_any_element();
+        }
+
+        let max_height = output_max_height.expect("checked above");
```
MrSubidubi: Let`s change this to a if-let-else to avoid the expect.

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +353,298 @@ mod tests {
...
+        if self.focus_handle.is_none() {
+            let handle = cx.focus_handle();
+            self._focus_subscription = Some(cx.on_focus_out(&handle, window, |this, _, _, cx| {
+                this.scroll_active = false;
+                cx.notify();
+            }));
+            self.focus_handle = Some(handle);
+        }
+        let focus_handle = self.focus_handle.clone().expect("initialized above");
```
MrSubidubi: This should be moved to `TerminalOutput::new`

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +353,298 @@ mod tests {
...
+                            cx.notify();
+                        } else {
+                            cx.propagate();
+                        }
+                    }))
+                    .max_h(max_height)
+                    .track_scroll(&self.scroll_handle)
+                    .when(scroll_active, |this| this.overflow_y_scroll())
+                    .when(!scroll_active, |this| this.overflow_y_hidden())
+                    .on_scroll_wheel(cx.listener(|this, _: &ScrollWheelEvent, _, cx| {
```
MrSubidubi, line 608: Nit: This could use `when_else`

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +353,298 @@ mod tests {
...
+                        .top_0()
+                        .left_0()
+                        .w_full()
+                        .h(fade_height)
+                        .bg(linear_gradient(
+                            0.,
+                            linear_color_stop(bg_color.opacity(0.), 0.),
+                            linear_color_stop(bg_color, 1.),
+                        )),
+                )
```
MrSubidubi: I think I'd prefer if we have a function for this one and the one below like `gradient_overlay` so it is more visible what these do here which could have an enum parameter `enum GradientStart { Top, Bottom }` for both of these here.

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +353,298 @@ mod tests {
...
+            self.focus_handle = Some(handle);
+        }
+        let focus_handle = self.focus_handle.clone().expect("initialized above");
+
+        let max_offset = self.scroll_handle.max_offset();
+        let offset = self.scroll_handle.offset();
+        let has_overflow = max_offset.y > text_line_height;
+        let not_at_top = has_overflow && offset.y < -px(1.);
+        let not_at_bottom = has_overflow && offset.y > -max_offset.y + px(1.);
+
```
MrSubidubi: Could you elaborate why we don't just check for `offset.y.is_zero()` and the offset == max_offset here? I guess we'd like some kind of tolerance, but is it that useful in practice?

crates/repl/src/outputs/plain.rs
```diff
@@ -82,27 +92,47 @@ pub fn text_style(window: &mut Window, cx: &App) -> TextStyle {
...
+    let cell_width = cell_width(window, cx);
+    if cell_width == Pixels::ZERO {
+        return 80;
+    }
+
+    let viewport_width = window.viewport_size().width;
+    let gutter_estimate = cell_width * 8.0;
+    let available_width = (viewport_width - gutter_estimate).max(cell_width * 20.0);
+    (available_width / cell_width).floor() as usize
```
MrSubidubi, line 159: I don't think the estimate here is good and thus I think we need to change things here a bit: 

We render this as a block and thus have access to the `BlockContext` which holds the `EditorMargins`. I think we should wire this through so that we can use these here, as they will always contain the actual margins as configured by the user.

crates/repl/src/outputs/plain.rs
```diff
@@ -200,6 +234,10 @@ impl TerminalOutput {
     ///
     /// * `text` - A string slice containing the text to be appended.
     pub fn append_text(&mut self, text: &str, cx: &mut App) {
+        let max_offset = self.scroll_handle.max_offset();
+        let offset = self.scroll_handle.offset();
+        let at_bottom = max_offset.y <= Pixels::ZERO || offset.y <= -max_offset.y + px(1.);
```
MrSubidubi: If the `px(1.)` is tolerance, can we make this a const variable that indicates just that? As I think this is what it tries to do. That said, is this strictly necessary as in did you run into this a lot during testing?

MrSubidubi (changes requested):

crates/repl/src/outputs/plain.rs
```diff
@@ -311,81 +397,262 @@ mod tests {
...
+            };
+            self.handler.resize(bounds);
+        }
+
+        // Iterate ALL grid lines (history + visible) so output is never truncated.
+        let grid = self.handler.grid();
+        let start = Point::new(Line(grid.topmost_line().0 - 1), grid.last_column());
+        let grid = grid.iter_from(start).map(|ic| terminal::IndexedCell {
+            point: ic.point,
+            cell: ic.cell.clone(),
```
MrSubidubi, line 527: Sorry once again for a delay here. Last thing that bothers me here a bit is that we iterate over everything here - can we not instead just iterate over what we will end up rendering by any chance?

MrSubidubi (changes requested):
Looks good now, one final commend and we should be good to go!

rgbkrk:
@MrSubidubi thanks for the review. I think someone else can take this on if they want it.

@magvag if you want to take it on and add the setting feel free. I'm not likely going to spend cycles on this any time soon.

PR #49945, opened 2026-02-24
bedrock: Show increased context size when `allow_extended_content` set

### LINKED ISSUE #49617: Bedrock: `allow_extended_context` does not update UI context window limit (still shows 200k)
### Reproduction steps

1. Configure Bedrock with `"allow_extended_context": true` in settings.json
2. Open the Agent Panel and select a supported model (e.g., Claude Sonnet 4.5 or Claude Opus 4.6)
3. Have a long conversation that approaches 200k tokens
4. Observe that Zed warns about approaching the context limit and prompts to start a new thread, even though the 1M extended context window is active

**Current behavior:** When `allow_extended_context` is enabled in Bedrock settings, the API correctly receives the `anthropic_beta` header and accepts up to 1M tokens. However, Zed's UI still reports a 200k context window limit. It shows a warning and prompts the user to start a new thread at ~200k tokens.

**Expected behavior:** When `allow_extended_context` is `true` and the model supports extended context, Zed should report the context window as 1,000,000 tokens instead of 200,000. The `max_token_count()` in the `LanguageModel` trait implementation for `BedrockModel` ([here](https://github.com/zed-industries/zed/blob/main/crates/language_models/src/provider/bedrock.rs)) delegates directly to `Model::max_token_count()` ([here](https://github.com/zed-industries/zed/blob/main/crates/bedrock/src/models.rs)), which is hardcoded to `200_000` for all Claude models. PR #48542 updated the API request path to send the extended context header, but did not update the model metadata that the UI uses for context window warnings.

A possible fix would be to have `BedrockModel::max_token_count()` return `1_000_000` when both `allow_extended_context` is enabled and the model's `supports_extended_context()` returns true. Since the `LanguageModel` trait's `max_token_count(&self)` doesn't take a `cx` parameter, the `use_extended_context` flag could be stored directly on `BedrockModel` at construction time (models are already recreated when settings change via `provided_models()`).

Zed Preview 0.225.1 

### Attach Zed log file

N/A

### Relevant Zed settings
```
{
  "language_models": {
    "bedrock": {
      "authentication_method": "named_profile",
      "region": "us-west-2",
      "profile": "<redacted>",
      "allow_extended_context": true
    }
  }
}
```

### Relevant Keymap

N/A

- Provider: Amazon Bedrock (direct AWS credentials)
- Model Name: Claude Opus 4.6 (also affects Claude Sonnet 4, Sonnet 4.5, Opus 4.5, and their Thinking variants)
- Mode: Agent Panel
- Other details: `allow_extended_context: true` is enabled per PR #48542. The API-level 1M context works correctly (the `anthropic_beta` header with `context-1m-2025-08-07` is sent), but the UI-facing `max_token_count()` is not updated to reflect this, causing premature context window warnings.

None

rgbkrk:
Closes #49617

image

Release Notes:

- Fixed Bedrock context window display showing 200k instead of 1M when `allow_extended_context` is enabled for supported Claude models

### DIFF
--- a/crates/language_models/src/provider/bedrock.rs
+++ b/crates/language_models/src/provider/bedrock.rs
@@ -417,7 +417,12 @@ impl BedrockLanguageModelProvider {
         }
     }
 
-    fn create_language_model(&self, model: bedrock::Model) -> Arc<dyn LanguageModel> {
+    fn create_language_model(
+        &self,
+        model: bedrock::Model,
+        allow_extended_context: bool,
+    ) -> Arc<dyn LanguageModel> {
+        let use_extended_context = allow_extended_context && model.supports_extended_context();
         Arc::new(BedrockModel {
             id: LanguageModelId::from(model.id().to_string()),
             model,
@@ -426,6 +431,7 @@ impl BedrockLanguageModelProvider {
             state: self.state.clone(),
             client: OnceCell::new(),
             request_limiter: RateLimiter::new(4),
+            use_extended_context,
         })
     }
 }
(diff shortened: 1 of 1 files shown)


rgbkrk:
@5herlocked ^^

crates/language_models/src/provider/bedrock.rs
```diff
@@ -663,7 +677,11 @@ impl LanguageModel for BedrockModel {
     }
 
     fn max_token_count(&self) -> u64 {
-        self.model.max_token_count()
+        if self.use_extended_context {
+            1_000_000
```
rtfeldman, line 681: Hm, so this would hardcode it to 1M for *all* models?

@rgbkrk could we tie that count to the individual model instead? 😅

rgbkrk: I mean... that was what @5herlocked didn't want to have to do per model because it's _only_ on if they set their settings to allow use extended context. But if it's.... here then we can make sure it applies to the _two_ models that have it (Opus 4.6, Sonnet 4.5).

rtfeldman: Right, I'm just concerned about a new model coming out and silently (incorrectly) getting this applied to it by default. 😅 

Is there a particular reason to not want it to be per-model?

cal-gooo: or we can just have simple option to disable this context window icon and the alert for meeting the context window limit..

cal-gooo: Let's get this merged soon, 1m models are used pretty popularly in companies, and this UI is pretty annoying

rgbkrk: I don't know a clean way to do this so I think I'll just close my PR as I have other work to do.

Issue #25796, opened 2025-02-28
FormatSelections Command does not work


Rocco-Gossmann:
### Summary

When choosing `editor: format selections` from the command-palette, the ENTIRE DOCUMENT is formatted.

Steps to trigger the problem:
1. Open the `settings.json` by pressing `cmd+,`
2. Add some additional Tabs to the start of a line of your choice.
3. Add some additional Tabs to the start of a different line of your choice.
4. Now select the first Line
5. Open the Command-Palette via `cmd-shift-p`
6. Choose `editor: format selections`

Actual Behavior:
The entire document is formatted.

Expected Behavior:
Only the selected Area should be formatted.

If no Formatting of a selection is NOT possible. (Maybe because the LSP does not support selection formatting) I expect the Editor to format nothing and just give me an info. Otherwise this can be very harmful to older projects with big files.

### Zed Version and System Specs

Zed: v0.175.6 (Zed) 
OS: macOS 15.1.1
Memory: 8 GiB
Architecture: aarch64

notpeter:
This is a little confusing. Default formatter for JSON/JSONC is prettier, which cannot do range formatting.

https://github.com/zed-industries/zed/blob/19232c035a51b2801e5bb5be5b9e5e1b1f345395/assets/settings/default.json#L1156-L1165

We should probably remove/disable this action unless you are using the language server as your formatter.  If you want this functionality you can switch your formatter for JSON/JSONC to json-language-server with:

```
  "languages": {
    "JSONC": {
      "formatter": "language_server"
    },
    "JSON": {
      "formatter": "language_server"
    }
  },```

Rocco-Gossmann:
I see the point of having a formatter, but I don't like to have to meddle with configs just get one running. 
So, I usually just use what ever NeoVim, VSCode or, in this case, Zed sets up.
- VSCode wouldn't offer the option for formatting the selection if the Language-Server does not support the function.
- NeoVim just gives me a Warning, that the feature is not available.

- Zed is the only Editor so far, that would format everything. Letting the User think it "worked", without them noticing, that maybe other parts of the document are now ruined.

If I could ask for a wish, it would be, that Zed just does nothing, if the function is called via keyboard shortcut (and maybe also hides the option completely from the menu)

terziele:
It is possible to do the selection formattion via prettier, btw.
But yes, it seem reasonable not to show the button, or perhaps show a warning, if it's not possible due to the external formatter or  unsupported feature in LSP.

[...]

PR #49964, opened 2026-02-24
Add support for `format_on_save` only changed lines

### LINKED ISSUE #16509: format_on_save: save only changed lines
### Description

Currently `format_on_save` has two options: "on" and "off".

It would be useful to have the option "only changed lines", so that only lines that you just edited will be formatted.
This way your coworkers will not be confronted with all indentation changes in a pull request.

It would be nice if the same logic could also suppress jitter from `remove_trailing_whitespace_on_save` / `ensure_final_newline_on_save` changes.

G36maid:
## Summary

Introduces `format_on_save` variants that format only modified lines, limiting formatting to lines changed since the last commit. This prevents massive diffs when editing legacy codebases. Aligns with VS Code's `editor.formatOnSaveMode` naming:

- `"modifications"` — formats only git-diffed lines. Requires source control; skips formatting if no diff is available.
- `"modifications_if_available"` — formats only git-diffed lines, falling back to full-file formatting for untracked files, when source control is unavailable, or when the LSP does not support range formatting.

Also supports importing equivalent VS Code settings (`formatOnSaveMode`).

This PR uses the range-based whitespace/newline infrastructure from the dependency PR above.

## Changes

<details>
<summary><b>New settings, modified ranges computation, VS Code import</b></summary>

### New settings (`language.rs`, `default.json`, `all-settings.md`)

Two new `FormatOnSave` variants: `Modifications` and `ModificationsIfAvailable`.

### Modified ranges computation (`items.rs`)

- `compute_format_decision()` — reads `format_on_save` setting across all buffers, determines whether to use ranged formatting (`Ranges`), full formatting (`Full`), or skip (`Skip`).
- `compute_modified_ranges()` — extracts modified line ranges from git diff hunks via `BufferDiffSnapshot`.
- `is_empty_range()` — helper to detect deletion-only hunks that produce no formatable content.

The `save()` method calls `compute_format_decision()` and dispatches accordingly.

### Modifications mode in `lsp_store.rs`

- Adds `FormatOnSave::Modifications | FormatOnSave::ModificationsIfAvailable` to the formatter selection match arm.
- `ModificationsIfAvailable` falls back to full-file formatting when ranged formatting produces no results.

### VS Code settings import (`vscode_import.rs`, `settings_store.rs`)

Imports `editor.formatOnSaveMode` mapping:
- `"modifications"` → `FormatOnSave::Modifications`
- `"modificationsIfAvailable"` → `FormatOnSave::ModificationsIfAvailable`
- `"file"` → `FormatOnSave::On`

</details>

## Tests

- `test_modifications_format_on_save` — basic modifications mode formatting with dirty buffer
- `test_modifications_format_no_changes` — clean buffer triggers no formatting
- `test_modifications_format_lsp_no_range_support` — LSP without range formatting skips entirely for `Modifications`
- `test_modifications_format_lsp_returns_empty_edits` — empty edits handled gracefully
- `test_modifications_format_multiple_hunks` — two non-adjacent edits produce two separate range formatting requests

---

Self-Review Checklist:

Closes #16509
Depends on #53942

Release Notes:

- Added `modifications` and `modifications_if_available` options to `format_on_save`, which format only git-changed lines instead of the entire file
- When using modifications mode, `remove_trailing_whitespace_on_save` and `ensure_final_newline_on_save` are also scoped to changed lines, preventing unwanted diff jitter in legacy codebases
- Added support for importing VS Code's `editor.formatOnSaveMode` setting


UnknownDK:
I get this, when trying to use it with ty for Python. Maybe this should be handled, instead of an error? 
```
2026-02-24T19:48:05+01:00 ERROR [project::lsp_store] Formatting failed: Failed to format ranges via language server: Unknown request: textDocument/rangeFormatting
2026-02-24T19:48:05+01:00 ERROR [crates/editor/src/editor.rs:19222] Failed to format ranges via language server

Caused by:
    Unknown request: textDocument/rangeFormatting
```

When trying with Ruff, it just formats like normal without throwing any errors, so that also doesnt work :/

G36maid:
### 🧪  Testing: LSP Range Formatting Support Status

Since `textDocument/rangeFormatting` implementation varies wildly across the ecosystem, let's track how different Language Servers and Formatters behave with `format_on_save: "modifiedlines"`. 

If you test a language, please help update this checklist!

**Legend:**
- [x] **Works perfectly:** Only modifies changed lines.
- [~] **Graceful Skip:** Doesn't support ranges, safely skips without errors.
- [F] **Full-file Fallback:** Ignores the ranges and formats the whole file anyway.
- [!] **Errors out:** Throws an `Unknown request` or breaks the save pipeline.

#### 📦 Built-in Languages
- [ ] **C / C++**
  - [ ] `clangd`
- [ ] **Python**
  - [ ] `Ruff` *(Expected: Ruff doesn't support range formatting yet)*
  - [ ] `ty` *(Throws `Unknown request` - fix for graceful error handling incoming)*
  - [ ] `Pyright`
- [ ] **Rust**
  - [ ] `rust-analyzer` / `rustfmt` *(Note: requires nightly/unstable config)*
- [ ] **TypeScript / JavaScript**
  - [ ] `vtsls`
  - [ ] `typescript-language-server`
  - [ ] `Prettier` (External)
  - [ ] `Biome`
- [ ] **Go**
  - [ ] `gopls` *(Expected to reject/fallback as Go prefers strict full-file formatting)*
- [ ] **Web (JSON / HTML / CSS)**
  - [ ] `json-language-server`
  - [ ] `vscode-html-language-server`

#### 🧩 Popular Extensions
- [ ] **Vue** (`volar`)
- [ ] **Svelte** (`svelte-language-server`)
- [ ] **Ruby** (`ruby-lsp` / `rubocop`)
- [ ] **PHP** (`intelephense` / `phpactor`)
- [ ] **Java** (`jdtls`)
- [ ] **C#** (`omnisharp` / `roslyn`)
- [ ] **Elixir** (`elixir-ls` / `lexical`)
- [ ] **Zig** (`zls`)
- [ ] **Nix** (`nil` / `nixd`)
- [ ] **Gleam** (`gleam`)

---
**How to test:**
1. Pull the branch: `gh pr checkout 49964`
2. Set `"format_on_save": "modifiedlines"` in your `settings.json`
3. Edit a file, save, and observe the behavior (and check the Zed logs for errors!).

G36maid:
> I get this, when trying to use it with ty for Python. Maybe this should be handled, instead of an error?
> 
> ```
> 2026-02-24T19:48:05+01:00 ERROR [project::lsp_store] Formatting failed: Failed to format ranges via language server: Unknown request: textDocument/rangeFormatting
> 2026-02-24T19:48:05+01:00 ERROR [crates/editor/src/editor.rs:19222] Failed to format ranges via language server
> 
> Caused by:
>     Unknown request: textDocument/rangeFormatting
> ```
> 
> When trying with Ruff, it just formats like normal without throwing any errors, so that also doesnt work :/
> 
> Seems to work for C++ though.

Maybe because ty isn't a formater?

According to Astral's official docs, `ty` is strictly a type checker, while `Ruff` is the formatter.

crates/settings_content/src/language.rs
```diff
@@ -886,12 +895,21 @@ pub struct PrettierSettingsContent {
     strum::VariantArray,
     strum::VariantNames,
 )]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "camelCase")]
```
SomeoneToIgnore:
```suggestion
#[serde(rename_all = "snake_case")]
```

In our settings at least, that seems to be the baseline.

G36maid:
here are some recent related issues and PRs

- #25796 — `FormatSelections` command formats entire document instead of selection (Open)
- #53178  — Gate Format Selections on whether the active formatter can actually format ranges

SomeoneToIgnore (commented):
FWIW I've even started with the PR at some point, if you're so eager to get _some_ review...

G36maid:
Got it, thanks for the update.

crates/settings_content/src/language.rs
```diff
@@ -1205,4 +1214,34 @@ mod test {
...
+        let json = serde_json::to_string(&format_on_save)
+            .expect("Failed to serialize FormatOnSave::Modifications");
+        assert_eq!(json, "\"modifications\"");
+        let deserialized: FormatOnSave =
+            serde_json::from_str(&json).expect("Failed to deserialize FormatOnSave::Modifications");
+        assert_eq!(deserialized, FormatOnSave::Modifications);
+    }
+
+    #[test]
+    fn test_format_on_save_modifications_if_available_roundtrip() {
```
SomeoneToIgnore: These tests seem to test `serde`'s serialization, not Zed's code which is odd and redundant — let's remove them.

crates/migrator/src/migrations/m_2025_10_02/settings.rs
```diff
@@ -14,9 +14,9 @@ fn remove_formatters_on_save_inner(value: &mut Value, path: &[&str]) -> Result<(
     let Some(format_on_save) = obj.get("format_on_save").cloned() else {
         return Ok(());
     };
-    let is_format_on_save_set_to_formatter = format_on_save
-        .as_str()
-        .map_or(true, |s| s != "on" && s != "off");
+    let is_format_on_save_set_to_formatter = format_on_save.as_str().map_or(true, |s| {
+        s != "on" && s != "off" && s != "modifications" && s != "modifications_if_available"
```
SomeoneToIgnore, line 18: That does not look right?

On `2025_10_02` we did not have any new options, so we could not accept them as valid, hence no need to migrate?

crates/project/src/lsp_store.rs
```diff
@@ -1605,20 +1605,73 @@ impl LocalLspStore {
...
                 buffer.apply_diff(diff, cx);
             })?;
         }
 
         if settings.ensure_final_newline_on_save {
             zlog::trace!(logger => "ensuring final newline");
-            extend_formatting_transaction(buffer, formatting_transaction_id, cx, |buffer, cx| {
-                buffer.ensure_final_newline(cx);
-            })?;
+            if let Some(ranges) = buffer.ranges.as_ref() {
```
SomeoneToIgnore: Nice find, so we need to limit Zed's whitespace "formatting" to given ranges only.
Yet, changes in this file look quite unrelated to git and "only changed lines" functionality, am I right?
If so, we'd better extract it into a separate PR with tests.

crates/language/src/buffer.rs
```diff
@@ -2272,6 +2272,88 @@ impl Buffer {
...
+                        return Diff {
+                            base_version,
+                            line_ending,
+                            edits: Vec::new(),
+                        };
+                    }
+                    break;
+                }
+            }
+            let newline: Arc<str> = Arc::from("\n");
```
SomeoneToIgnore: What about Windows and their line separators, should that be considered either?

crates/editor/src/items.rs
```diff
@@ -888,29 +891,116 @@ impl Item for Editor {
...
+                .filter_map(|buffer_entity| {
+                    let buffer = buffer_entity.read(cx);
+                    let settings = LanguageSettings::for_buffer(buffer, cx);
+                    match settings.format_on_save {
+                        FormatOnSave::Modifications => Some(false),
+                        FormatOnSave::ModificationsIfAvailable => Some(true),
+                        _ => None,
+                    }
+                })
+                .reduce(|a, b| a && b);
```
SomeoneToIgnore: Can this `.filter_map(..).reduce(..)` chain be replaced with `all(..)`?

crates/editor/src/items.rs
```diff
@@ -888,29 +891,116 @@ impl Item for Editor {
...
+
+            if let Some(fallback_on_no_diff) = modifications_mode {
+                let mut mb_ranges: Vec<Range<Point>> = Vec::new();
```
SomeoneToIgnore: `mb` — let's use proper names instead of abbreviations, as most of the code around does.
That also has to be `m_b` at least, given the naming conventions.

crates/editor/src/items.rs
```diff
@@ -888,29 +891,116 @@ impl Item for Editor {
...
+
+                for buffer_entity in &buffers_to_save {
+                    let buffer = buffer_entity.read(cx);
+                    let buffer_id = buffer.remote_id();
+                    let diff_snapshot = git_store
+                        .get_unstaged_diff(buffer_id, cx)
+                        .map(|diff| diff.read(cx).snapshot(cx));
+
+                    match compute_modified_ranges(buffer, diff_snapshot.as_ref()) {
```
SomeoneToIgnore: If I'm not mistaken, this all is being computed synchronously on the main thread, which seems dangerous.
Just below, there's a task being spawned, can we do that?

crates/editor/src/items.rs
```diff
@@ -888,29 +891,116 @@ impl Item for Editor {
                 .filter(|buffer| buffer.read(cx).is_dirty())
                 .collect()
         };
+        let (ranged_format_target, fallback_buffers, modifications_mode) = {
+            let multi_buffer = self.buffer.read(cx);
+            let mb_snapshot = multi_buffer.read(cx);
+            let git_store = project.read(cx).git_store().read(cx);
+
+            // Determine if all buffers use a modifications-only format mode.
```
SomeoneToIgnore: Any chance we can rename a variable or extract this into a method to eliminate the comment that describes the code below verbatim?

SomeoneToIgnore: Even without that comment, the readability would definitely benefit if we could extract the new code into methods/functions.

SomeoneToIgnore (commented):
Overall, looks rather nice, thank you for coming up with this.

One major question I have is whether we can split that into two PRs instead: one for the whitespaces + ranges and another for the git diff-related logic.

crates/language/src/buffer.rs
G36maid: Good point for raising this. I think no changes needed here.

Zed's internal rope always stores `\n` exclusively. When `TextBuffer::new` loads text, it calls `LineEnding::detect` to record the original line ending as metadata, then `LineEnding::normalize` converts all `\r\n` → `\n`.

The `Diff.line_ending` field is just that metadata — it gets applied via `set_line_ending` in `apply_diff` and is used only when writing back to disk to restore `\r\n` if needed.

The original `ensure_final_newline` also hardcodes `"\n"` .

crates/editor/src/items.rs
G36maid: You're right that this runs synchronously on the main thread, but the work is lightweight:
- `get_unstaged_diff` is a `HashMap::get` + `WeakEntity::upgrade` — O(1), no I/O
- Diff snapshots are maintained incrementally as the user types; we only read them, never compute from scratch
- `compute_modified_ranges` is O(hunks) pure memory traversal, bounded by edit count not file size
Moving this into the spawned task via `cx.update(...)` wouldn't change the execution model

G36maid: Actually, on second thought — the subsequent `compute_modified_ranges` iterates over all hunks (expand to full lines, sort, merge adjacent ranges), followed by `text_anchors_to_visible_anchors` and `to_point` conversions. For a file with hundreds of changes, this is non-trivial.

I can move the heavy part into a `background_spawn` inside the existing `cx.spawn_in` — collect snapshots synchronously first (lightweight), then do the pure computation on a background thread. This requires some refactoring to split the function into a "collect snapshots" phase and a "compute from snapshots" phase, since all the snapshot types (`BufferSnapshot`, `BufferDiffSnapshot`, `MultiBufferSnapshot`) are `Clone`.

Would you like me to do that in this PR, or should we follow up separately?

crates/editor/src/items.rs
G36maid: Sure, i will make this part more self-documenting.

crates/project/src/lsp_store.rs
G36maid: Good point.
I'll split them into two PRs after a rebase.

G36maid:
Just split for the whitespaces + ranges into #53942

crates/migrator/src/migrations/m_2025_10_02/settings.rs
G36maid: I had the same thought initially. 
But the migration system doesn't track versions: `migrate_settings()` runs all migrations sequentially every time.
So even though `"modifications"` didn't exist on 2025_10_02, migration will treats it as a formatter name.

SomeoneToIgnore: Yes, but on that year, there's been no such settings value, so it could have not appeared in the sources to migrate from?

I've also asked the person who wrote most of the migration-related code, they agree this bit does not belong and has to be left as is.

crates/editor/src/editor_tests.rs
```diff
@@ -14236,106 +14239,204 @@ async fn setup_range_format_test_with_capabilities(
...
-        settings.defaults.formatter = Some(FormatterList::Single(Formatter::External {
-            command: "awk".into(),
-            arguments: Some(vec!["{ print }".to_string()]),
-        }));
+async fn test_modifications_format_no_changes(cx: &mut TestAppContext) {
+    let (project, editor, cx, fake_server) = setup_range_format_test(cx).await;
+
+    update_test_language_settings(cx, &|settings| {
+        settings.defaults.format_on_save = Some(FormatOnSave::Modifications);
+        settings
```
SomeoneToIgnore: Sorry, I do not understand this test a bit — this does not format because originally, the editor is not dirty?
Maybe would be a bit more readable if we assert that too (as the previous test did) and add some messages into `assert!` macro as the last argument.
 
Also, here and in some other tests, why do we set a special, Rust-related setting that is the same as the defaults (hence, noop?)?

crates/editor/src/items.rs
```diff
@@ -895,18 +898,49 @@ impl Item for Editor {
...
-                        FormatTarget::Buffers(buffers_to_save.clone()),
-                        window,
-                        cx,
-                    )
-                })?
-                .await?;
+                match format_decision {
+                    FormatDecision::Ranges(ranges) => {
+                        this.update_in(cx, |editor, window, cx| {
+                            editor.perform_format(
```
SomeoneToIgnore: Looks like those two formatting blocks differ with the `FormatTarget` only?

Wondering if it makes any sense to emit `Option<FormatTarget>` from `compute_format_decision`-esque function instead, and consider removing that `FormatTarget` enum entirely?

Then, another idea to evaluate is whether we can remove the `collect_format_decision_inputs` along with the inputs type and get both `collect_format_decision_inputs` and `compute_format_decision` merged together?
We could structure it as 
```rs
let formatting_task = compute_format_target(&buffers_to_save, self.buffer.clone(), project.read(cx).git_store().clone(), cx);`
```

and then await on the task later in the `spawn_in` where we can move it.

crates/editor/src/items.rs
```diff
@@ -2088,6 +2119,183 @@ fn deserialize_path_key(path_key: proto::PathKey) -> Option<PathKey> {
...
+}
+
+fn is_empty_range(range: &Range<text::Anchor>, snapshot: &text::BufferSnapshot) -> bool {
+    range.start.cmp(&range.end, snapshot).is_eq()
+}
+
+/// Computes the buffer ranges that have been modified, for use with format-on-save.
+/// Returns `None` when there is no diff or edit history to determine changed regions,
+/// which signals the caller should fall back to full-document formatting.
+pub fn compute_modified_ranges(
```
SomeoneToIgnore: Any reason to have it `pub`?

SomeoneToIgnore: Also, wanted to mention, whilst the current code seems ok, we need to be really careful and not move any snapshot past `.await` points — otherwise we risk getting panics due to whatever data mismatch; if we need anything snapshot-like after the `.await`, we need to procure fresh ones.

crates/editor/src/items.rs
SomeoneToIgnore: Also, it seems that `compute_format_target` is not really async?
What we do there are coordinate transformations it seems, what is our reason to get the async wrapper around this?

crates/editor/src/items.rs
```diff
@@ -2088,6 +2119,183 @@ fn deserialize_path_key(path_key: proto::PathKey) -> Option<PathKey> {
...
+            } else {
+                end_pt.row
+            };
+            let line_start = text::Point::new(start_row, 0);
+            let line_end = text::Point::new(end_row, buffer_snapshot.line_len(end_row));
+            buffer_snapshot.anchor_before(line_start)..buffer_snapshot.anchor_after(line_end)
+        })
+        .collect();
+
+    expanded.sort_by(|a, b| a.start.cmp(&b.start, buffer_snapshot));
```
SomeoneToIgnore: I'm quite certain [git] hunks are ordered and I do not see reasons why `anchored_edits_since` would differ in that, as we traverse the B+Tree for that anyway.

So, seems that we do not need sorting and can do the merging when processing the expansions?

crates/project/src/lsp_store.rs
```diff
@@ -1867,8 +1871,35 @@ impl LocalLspStore {
...
+                    .await;
+
+                    match range_result {
+                        Ok(edits) => edits,
+                        Err(err) => {
+                            if matches!(
+                                settings.format_on_save,
+                                FormatOnSave::ModificationsIfAvailable
+                            ) {
+                                zlog::debug!(
```
SomeoneToIgnore, line 1967: Had raised a similar discussion in a split-off PR, just to repeat: would it make sense to use `info` for all such skips, to let people know somehow if things do not work silently.

I guess, it could be spammy if we execute this branch on each save, but at least for the manual formatting case this would be useful?

SomeoneToIgnore (commented):
Thank you, this looks much cleaner and now I could focus on the impl details better.

Had left a few comments, this seems to need a good pass on `lsp_store.rs` new code at least.

crates/migrator/src/migrations/m_2025_10_02/settings.rs
G36maid: Sure, i will remove this, thanks!

crates/project/src/lsp_store.rs
G36maid: Thanks for the reminder!

crates/editor/src/items.rs
G36maid: Just remove the pub,
I will double-check the logic.

crates/editor/src/editor_tests.rs
G36maid: I removed the migration change per your suggestion, but the tests now fail (they pass with the entries present).

I'm not sure how to make both the removal and the tests work — any ideas?

crates/editor/src/items.rs
G36maid: Done with  merged both functions into a single synchronous compute_format_target that returns Option<FormatTarget> directly, and the async wrapper is removed — it's just coordinate transformations on the main thread now.
 Thanks.

crates/editor/src/items.rs
G36maid: Done. Removed the sort step and now merge during expansion in a single pass.
 Thanks!

crates/editor/src/editor_tests.rs
SomeoneToIgnore: The tests seem to fail on a logical assertion, though?

image


I have no good ideas but to come and investigate what's wrong, but I find it extremely odd that the migration entries influence that at all — IIRC we do not run the migration in regular tests at all?

G36maid: Here's what I found with LLM assistance :
update_test_language_settings() (editor_tests.rs:31745) → 
set_user_settings() (settings_store.rs:916) → 
parse_and_migrate_zed_settings() (settings_store.rs:775 ) → 
migrate_settings() (migrator.rs:159)

If we remove `modifications` / `modifications_if_available` from the whitelist, we'd need to change SettingsStore to only run migrations on actual version upgrades 
but that's a larger change outside the scope of this PR

SomeoneToIgnore: I have checked on that closer myself now and seems that you're right, sorry for confusing, it is a new code area for me.

Luckily, seems that this is the only place that removes the values, so nothing is broken fundamentally, just that particular place being odd.

In this case, we indeed better whitelist instead of changing it too much.

G36maid: Thanks! 
I will add it back.

G36maid:
It's been a while since I touched this PR. I'm planning to rebase it and see how the build/tests go.

G36maid:
I found some bug only appears in real-world use cases. 

Since the current tests lack a Git environment, I will improve the test coverage.

> Update:notice that VS Code's `format-modified-lines` feature diffs against the Git `index`, not `HEAD`. Therefore, using `get_unstaged_diff` is correct.

G36maid:
note: behavioral divergence from VS Code

While adding test coverage I audited modifications_if_available against VS Code's editor.formatOnSaveMode: "modificationsIfAvailable" and found two intentional divergences in sad-path behavior:

1. SCM exists but no uncommitted changes: VS Code skips formatting; Zed falls back to full-file format.
2. LSP lacks range formatting support: VS Code skips formatting; Zed falls back to full-file format.

Both are deliberate ("if_available should always format something"), now documented in all-settings.md and covered by tests. Wanted to flag in case maintainer/community prefer strict VS Code parity. Happy to switch to the VS Code semantics if that's the desired direction.

SomeoneToIgnore (approved):
Thanks!

PR #50019, opened 2026-02-24
image_viewer: Add image smoothing setting for nearest neighbor based rendering

MostlyKIGuess:
Allows users to turn on/off linear interpolation in Linux.
This is especially useful when dealing with similarity matrices, confusion matrices, many other ML visualizations where linear interpolation would create harm to the actual values.

Currently only implemented for Linux (Wayland/X11) via the wgpu renderer. macOS and Windows use hardcoded linear samplers in their respective shaders and are unaffected. I am not that familiar with metal and don't have a Mac to test out by adding a shader than regardless. I could work on DirectX implementation after this for Windows.

Release Notes:

- Added `image_smoothing` setting for the image viewer, supporting `"linear"` (default) and `"nearest"` interpolation on Linux


crates/gpui/src/elements/img.rs
```diff
@@ -127,6 +127,7 @@ where
 /// The style of an image element.
 pub struct ImageStyle {
     grayscale: bool,
+    nearest_neighbor: bool,
```
mikayla-maki, line 130: Let's prefer enums to bools

mikayla-maki (changes requested):
Great PR, like the simple code. I'd like to know at least what kind of effort it would take to implement this for Windows and macOS, I'd prefer not to ship settings that should work cross platform, but only work for one platform.

DavidDudson:
Just ran into this myself (webp pixel art for Bevy), keen to see this in at some point.

PR #50027, opened 2026-02-24
git_ui: Improve connection between the graph and commit views

danilo-leal:
- Enabled opening the Git Graph, with the corresponding commit detail drawer open, from the commit view
- Redesigned the commit view's header and toolbar to allow addition of the Git Graph icon button
- Redesigned icons for the Git Graph and commit view

https://github.com/user-attachments/assets/8efef60a-0893-4752-9b40-838da21ceb54

---

Release Notes:

- N/A (_Git Graph is still feature flagged, so no release notes for now_)

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -39,7 +39,7 @@ use ui::{
 };
 use workspace::{
     Workspace,
-    item::{Item, ItemEvent, SerializableItem},
+    item::{Item, ItemEvent, SerializableItem, TabTooltipContent},
 };
 
 const COMMIT_CIRCLE_RADIUS: Pixels = px(3.5);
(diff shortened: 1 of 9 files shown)


maxbrunsfeld:
Love this! Is there a way to navigate the other way - opening the commit view from the git graph?

maxbrunsfeld:
Nice!

crates/git_graph/src/git_graph.rs
```diff
@@ -1153,6 +1189,37 @@ impl GitGraph {
...
+        let Ok(oid) = sha.parse::<Oid>() else {
+            return;
+        };
+        for (idx, commit) in self.graph_data.commits.iter().enumerate() {
+            if commit.data.sha == oid {
+                self.pending_select_sha = None;
+                self.select_entry(idx, cx);
+                return;
+            }
+        }
```
Anthony-Eid, line 1202: We should probably make a hash map of commit shas to their row in the graph, it will be needed for git search as well. This isn't a blocker, just something to add to our todo list in the future

crates/git_graph/src/git_graph.rs
```diff
@@ -944,8 +978,10 @@ impl GitGraph {
                     self.graph_data.add_commits(commits);
                 });
                 cx.notify();
+                self.retry_pending_select(cx);
```
Anthony-Eid, line 981: I think we can avoid having `PENDING_SELECT_MAX_RETRIES` in retry_pending select. 

`repository.graph_data(...).1` is true if the graph is still loading commits, we can clear the pending selection if it's false and we haven't found the commit a user wants to see

Anthony-Eid: It might be better if we make another repository event that is emitted once the graph is loaded, because there's a small race condition with `repository.graph_data(...).1`, so it might be `true` when we hit the last block of loaded commits because the task hasn't been finish yet.

Anthony-Eid (approved):
None of my feedback is blocking, if you want to merge this I can make the changes I want too later today

PR #50035, opened 2026-02-24
Add pane join actions for vim-like split close workflows

baldwindavid:
Follow-up to #47079 which added `tab_switcher::OpenInActivePane` for vim/helix-like buffer management (i.e. tabs hidden, buffers available in every pane). That PR noted the need for better pane close behavior — in vim/helix, closing a split (`:q`) closes the window but buffers remain available. Zed's tab model ties items to panes, so the closest equivalent is joining items into an adjacent pane without disrupting the destination's active tab.

The existing `pane::JoinIntoNext` gets partway there, but it always activates the source pane's tab in the destination, and only searches in the "next" direction (Right → Down → Left → Up). These new actions fill the gaps:

**Actions:**
- `pane::JoinIntoNext` *(existing)* — Joins this pane into the next pane. The source pane's active tab remains active.
- `pane::JoinIntoNextAsInactive` *(new)* — Joins this pane into the next pane. The destination pane's active tab remains active.
- `pane::JoinIntoPrevious` *(new)* — Joins this pane into the previous pane. The source pane's active tab remains active.
- `pane::JoinIntoPreviousAsInactive` *(new)* — Joins this pane into the previous pane. The destination pane's active tab remains active.

**Example keymap for vim-like `:q` behavior:**
```json
{
  "context": "Pane",
  "bindings": {
    "cmd-w": "pane::JoinIntoPreviousAsInactive"
  }
}
```

This closes the current split and moves its items to the previous pane (typically left/above), without disrupting the destination pane's active tab — approximating how vim/helix close windows.

**Future work:**
- Map vim's `:q` to `pane::JoinIntoPreviousAsInactive` so closing a split behaves like vim/helix out of the box.

- [x] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) (N/A — no UI changes)

Release Notes:

- Added `pane::JoinIntoNextAsInactive`, `pane::JoinIntoPrevious`, and `pane::JoinIntoPreviousAsInactive` actions for more flexible pane join behavior, enabling vim/helix-like split close workflows.

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -249,8 +249,14 @@ actions!(
         GoToOlderTag,
         /// Navigates forward in the tag stack.
         GoToNewerTag,
-        /// Joins this pane into the next pane.
+        /// Joins this pane into the next pane. The source pane's active tab remains active.
         JoinIntoNext,
+        /// Joins this pane into the next pane. The destination pane's active tab remains active.
+        JoinIntoNextAsInactive,
+        /// Joins this pane into the previous pane. The source pane's active tab remains active.
+        JoinIntoPrevious,
+        /// Joins this pane into the previous pane. The destination pane's active tab remains active.
+        JoinIntoPreviousAsInactive,
         /// Joins all panes into one.
         JoinAll,
         /// Reopens the most recently closed item.
@@ -310,6 +316,9 @@ pub enum Event {
     ItemUnpinned,
     JoinAll,
     JoinIntoNext,
+    JoinIntoNextAsInactive,
+    JoinIntoPrevious,
+    JoinIntoPreviousAsInactive,
     ChangeItemTitle,
     Focus,
     ZoomIn,
@@ -343,6 +352,9 @@ impl fmt::Debug for Event {
                 .finish(),
             Event::JoinAll => f.write_str("JoinAll"),
             Event::JoinIntoNext => f.write_str("JoinIntoNext"),
+            Event::JoinIntoNextAsInactive => f.write_str("JoinIntoNextAsInactive"),
+            Event::JoinIntoPrevious => f.write_str("JoinIntoPrevious"),
+            Event::JoinIntoPreviousAsInactive => f.write_str("JoinIntoPreviousAsInactive"),
             Event::ChangeItemTitle => f.write_str("ChangeItemTitle"),
             Event::Focus => f.write_str("Focus"),
             Event::ZoomIn => f.write_str("ZoomIn"),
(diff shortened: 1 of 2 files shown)


dinocosta (commented):
Hey @baldwindavid 👋 

Thank you so much for suggesting this change! I can see how introducing something like this can be useful for users that are trying to get a setup as close as possible to Neovim/Vim with [`tab_bar`](https://zed.dev/docs/reference/all-settings?highlight=tab_bar#editor-tab-bar) hidden 🙂

Am I missing a specific use case or would simply supporting the `JoinIntoNextAsInactive` (or similar) be enough to support Neovim's behavior? I tested the `:q` behavior a bit with 3 splits in Neovim and it seemed to always focus on the next (right) split, when closing the middle split, regardless of whether the previously focused split was the left or right one.

dinocosta:
> the key thing is just having _some_ way to approximate keeping buffers around and either of the `*AsInactive` actions would do that.

Makes total sense to me! 🙂

> Also, I'm not really enamored with `JoinIntoNextAsInactive`, but didn't land on anything better. A few others considered:
> 
> * JoinIntoNextPreserved
> * JoinIntoNextUnfocused
> * JoinIntoNextAfter
> * JoinAfterNext
> 
> And then there's also the option of keeping just `JoinIntoNext/Previous` and being able to provide config options to it (e.g. `preserve: true`), but it can be nice to be able to call via the command palette rather than only via configured keybinding.

Yes, was also thinking whether we could simply update/extend `JoinIntoNext` . Will dwell a little bit more on this and get back to you.

Thanks!

dinocosta:
I've also just noticed a couple of things that might be relevant:

* With the provided mapping, that is, mapping `cmd-w` to `JoinIntoNext` or any of its variants, we stop being able to close buffers/files when there's only a single pane remaining
* The approach also doesn't seem to directly tackle the `:q` behavior, which is probably what we'd actually want
    * Having this in mind, I'm not entirely sure if we can achieve the proposed goal through updating keybindings, which is not possible for commands like `:q[uit]` and I'm not sure if we can split the single-pane vs multi-pane scenario in keymap contexts
    * Wonder if we'll need an entirely new action (maybe `vim::Quit`) for this or if we can repurpose `workspace::CloseActiveItem`, as if we're trying to chase this behavior, it would probably be nice to have it work on `:q`?
    
Still thinking about this! 😅

baldwindavid:
Related: I apparently also set this probably a couple years ago in my Zed settings and forgot about it...

```json
"pane_split_direction_horizontal": "down",
"pane_split_direction_vertical": "right",
```

So if we were to end up going with a standalone action (e.g. `vim::Quit`) we'd likely want it to look at this setting to determine which way to try to join first.

BCastilloAT3W:
I just want to add another user voice here: this workflow would be genuinely useful. For people using hidden tabs and a more vim buffer workflow, closing a split without disrupting the destination pane’s active item while keeping the buffers open is a big quality of life improvement.

Issue #26900, opened 2025-03-17
Many frames of latency between input and render


trbabb:
### Summary

Zed shows multiple frames of latency (I count at least four) between input and render. See the attached video for an example:

https://github.com/user-attachments/assets/09cae02e-3bdf-4f85-9090-2cfc9981a5f0

## Steps to trigger the problem:

See video.

Any user input (selection, typing, etc) is not reflected for several frames.

If it turns out to be relevant, I'm viewing on a 4K monitor. Other apps do not seem to show this level of latency.

## Actual Behavior:

Lag can be seen when sweep selecting text; observe the difference in pointer position and the end of the selection in the document. You can count the frames between when the pointer reaches an extreme of its oscillation to when the selection reaches the same extreme.

There is also a similar delay in typing, though this is harder to show with video or measure concretely. It manifests as a strong experience of sluggishness when typing.

## Expected Behavior:

I would expect no more than one frame of input latency. The demonstration video on Zed's main page advertises 120fps rendering.

### Zed Version and System Specs

Zed: v0.177.9 (Zed) 
OS: macOS 15.3.2
Memory: 32 GiB
Architecture: x86_64

probably-neb:
That is definitely an excess delay, and Is very difficult to reproduce with the suped-up m-series macs most of us use. Thanks for providing such a clear visualization!

javiersalcedopuyo:
I'm having the same problem, it's particularly noticeable when using vim mode since you're typing constantly.

### Zed Version and System Specs
Zed: v0.183.12 (Zed)
OS: macOS 15.3
Memory: 16 GiB
Architecture: x86_64 (integrated GPU)

HETAOSAMA:
https://github.com/user-attachments/assets/12057c25-8f6a-43e9-a457-24f890145eef

I have the same problem.It is very obvious when using the mouse to click.

Zed Version and System Specs
Zed: v0.187.9
OS: macOS 15.5
Memory: 32 GB
CPU: 2 GHz 4Core Intel Core i5 
integrated GPU：Intel Iris Plus Graphics 1536 MB

carpelli:
I have the same problem. There is noticeably more delay typing than in other programs.

Zed 0.193.3
1.2GHz 4-core Intel Core i7, Iris Plus
16 GB memory
macOS 14.6.1

tmbs:
Same.

- Zed 0.198.5 on macOS Version 15.6 (24G84)
[...]

PR #50040, opened 2026-02-25
Add MacOS and Linux settings for immediate rendering

shakeelmohamed:
This PR adds a user setting for immediate rendering, on MacOS and Linux. I saw noticeable improvements on MacOS 14.8.3 with Apple Silicon. I haven’t been able to manually test on Linux due to GPU virtualization issues. On my Windows VM I didn’t notice any rendering latency to begin with.

As a longtime user of Sublime Text, I love the extremely low latency from keystroke to render. The current Zed rendering latency on MacOS is the reason I haven’t used it more, but making immediate rendering the default could be a breaking change impacting more than the typing experience. 

I’m happy to refine as needed.

Possibly related to #26900.

### DIFF
--- a/crates/gpui_macos/src/metal_renderer.rs
+++ b/crates/gpui_macos/src/metal_renderer.rs
@@ -8,8 +8,8 @@ use cocoa::{
 };
 use gpui::{
     AtlasTextureId, Background, Bounds, ContentMask, DevicePixels, MonochromeSprite, PaintSurface,
-    Path, Point, PolychromeSprite, PrimitiveBatch, Quad, ScaledPixels, Scene, Shadow, Size,
-    Surface, Underline, point, size,
+    Path, Point, PolychromeSprite, PrimitiveBatch, Quad, RenderingMode, ScaledPixels, Scene,
+    Shadow, Size, Surface, Underline, point, size,
 };
 #[cfg(any(test, feature = "test-support"))]
 use image::RgbaImage;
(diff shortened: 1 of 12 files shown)


xdBronch:
https://docs.rs/wgpu/latest/wgpu/enum.PresentMode.html#variant.Immediate
`Immediate` doesnt seem to be supported on wayland, maybe allow `Mailbox` too? im not very familiar with any of this stuff so i dont know what drawbacks there may be

xdBronch:
im not a maintainer here or anything so i dont have any say there, just pointing it out :smile:

crates/gpui_macos/src/metal_renderer.rs
```diff
@@ -169,7 +177,11 @@ impl MetalRenderer {
         // Support direct-to-display rendering if the window is not transparent
         // https://developer.apple.com/documentation/metal/managing-your-game-window-for-metal-in-macos
         layer.set_opaque(!transparent);
-        layer.set_maximum_drawable_count(3);
+        let drawable_count = RENDERING_MODE.with(|m| m.get());
+        layer.set_maximum_drawable_count(match drawable_count {
+            RenderingMode::PlatformDefault => 3,
+            RenderingMode::Immediate => 1,
+        });
```
reflectronic, line 184: The documentation for `setMaximumDrawableCount` says: https://developer.apple.com/documentation/quartzcore/cametallayer/maximumdrawablecount

> You can set this value to 2 or 3 only; if you pass a different value, Core Animation ignores the value and throws an exception.

So it's not clear to me how this does anything.

reflectronic (changes requested):

crates/gpui_macos/src/metal_renderer.rs
shakeelmohamed: @reflectronic thanks for this, I guess any improvements I saw were placebo effect. I ran some profiling with [typometer](https://github.com/pavelfatin/typometer) using all 3 values in local builds and official version 0.226.5 but didn’t see a noticeable difference (~80ms latency). I’ll close this PR accordingly.

PR #50046, opened 2026-02-25
Add side-by-side split diff view to FileDiffView

NaritaHARUKI:
## Summary
- Enable VSCode-like side-by-side file comparison in FileDiffView using SplittableEditor
- Leverages existing SplittableEditor infrastructure that ProjectDiff already uses
- Controlled by the user's `diff_view_style` setting (split/unified)

## Changes
- `crates/git_ui/src/file_diff_view.rs`: Refactored to use SplittableEditor instead of plain Editor (+66/-30 lines)

## Test plan
- [ ] Open a file with git changes and verify the diff view renders correctly
- [ ] Toggle between split and unified diff view styles via settings
- [ ] Verify side-by-side comparison shows old/new content correctly

Release Notes:

- Enabled side by side diffs in the file diff view

### DIFF
--- a/crates/git_ui/src/file_diff_view.rs
+++ b/crates/git_ui/src/file_diff_view.rs
@@ -2,7 +2,7 @@
 
 use anyhow::Result;
 use buffer_diff::BufferDiff;
-use editor::{Editor, EditorEvent, MultiBuffer};
+use editor::{Editor, EditorEvent, EditorSettings, MultiBuffer, SplittableEditor};
 use futures::{FutureExt, select_biased};
 use gpui::{
     AnyElement, App, AppContext as _, AsyncApp, Context, Entity, EventEmitter, FocusHandle,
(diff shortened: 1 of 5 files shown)


Veykril (approved):
Thanks!

crates/git_ui/src/file_diff_view.rs
```diff
@@ -316,16 +338,19 @@ impl Item for FileDiffView {
...
+        })
     }
 
     fn breadcrumb_location(&self, _: &App) -> ToolbarItemLocation {
         ToolbarItemLocation::PrimaryLeft
     }
 
     fn breadcrumbs(&self, cx: &App) -> Option<(Vec<HighlightedText>, Option<Font>)> {
-        self.editor.breadcrumbs(cx)
+        self.editor.read(cx).rhs_editor().breadcrumbs(cx
```
Veykril:
```suggestion
        self.editor.read(cx).rhs_editor().breadcrumbs(cx)
```

Killusions:
Could we get a separate setting for this? I'd prefer unified in one place and split in the other.

PR #50049, opened 2026-02-25
editor: Fix folding for unindented multiline strings and comments

ryan-m-walker:
Closes #5057

## Summary

The indent-based code fold detection does not account for lines inside multiline strings or block comments that have less indentation than the surrounding code. This causes the fold scanner to think the enclosing block ends prematurely.

For example, folding `fn main()` here would fail because the raw string content at indent level 0 gets treated as the end of the function body:

``` rust
fn main() {
    let s = r#"
unindented content
"#;
}  
```

This PR checks whether a low-indent line falls inside a string or comment override region and skips it if so. This works across all languages that define `@string` or `@comment` overrides in their `overrides.scm`.

## Before

https://github.com/user-attachments/assets/a08e6bf8-4f25-4211-8a46-8f6da7e49247

## After

https://github.com/user-attachments/assets/cd5b36db-6d4d-420b-9d60-79f9fad8638e

## Test Plan

- Added `test_fold_with_unindented_multiline_raw_string`
- Added `test_fold_with_unindented_multiline_block_comment`
- All existing fold tests pass
- Manually tested both Rust and Python examples

Release Notes:

- Fixed code folding incorrectly collapsing when multiline strings or block comments contained unindented content

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -2320,6 +2320,19 @@ impl DisplaySnapshot {
                 if !line_indent.is_line_blank()
                     && line_indent.raw_len() <= start_line_indent.raw_len()
                 {
+                    if self
+                        .buffer_snapshot()
+                        .language_scope_at(Point::new(row, 0))
+                        .is_some_and(|scope| {
+                            matches!(
+                                scope.override_name(),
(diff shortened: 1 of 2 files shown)


ryan-m-walker:
@cla-bot check

smitbarmase (approved):
Thanks!

PR #50085, opened 2026-02-25
Fix "add custom context server" modal hanging indefinitely

tomhoule:
When adding a new MCP server through the UI modal, the "waiting for context server..." state can hang forever without surfacing any error. This PR addresses:

In `maintain_servers` we resolve the configuration for every enabled server (including extension servers) through a `join_all`. Extension servers call into the extension host to resolve their command, and that call had no timeout. A single slow or stuck extension blocks the entire `join_all`, which means `maintain_servers` never returns, `populate_server_ids` never runs, and the new server never even appears in the panel. This is now bounded by a 30-second per-extension timeout.

The `servers_to_start` loop in `maintain_servers` used `?` on `create_context_server`. If any server in the loop failed to create (e.g. the process could not be spawned), the error propagated out to `available_context_servers_changed`, which only logged it. No `ServerStatusChangedEvent` was ever emitted for the failed server, so the modal subscription waiting for Running/Stopped/Error never resolved. Other servers later in the loop were also skipped. The loop now handles each failure individually: it logs the error, emits an Error status event so the modal gets notified, and continues to the next server.

And `wait_for_context_server` (the future the modal awaits) had no timeout of its own. If no status event was ever emitted for any reason, it would wait forever. It now has a 120-second timeout that surfaces an actionable error message.

grafik

Release Notes:

- Fixed the loading state never terminating after clicking "Add Server" when adding a custom MCP servers. There are now timeouts, logs and user facing errors.

### DIFF
--- a/crates/project/src/context_server_store.rs
+++ b/crates/project/src/context_server_store.rs
@@ -8,7 +8,7 @@ use std::time::Duration;
 use anyhow::{Context as _, Result};
 use collections::{HashMap, HashSet};
 use context_server::{ContextServer, ContextServerCommand, ContextServerId};
-use futures::{FutureExt as _, future::join_all};
+use futures::{FutureExt as _, future::Either, future::join_all};
 use gpui::{App, AsyncApp, Context, Entity, EventEmitter, Subscription, Task, WeakEntity, actions};
 use itertools::Itertools;
 use registry::ContextServerDescriptorRegistry;
@@ -141,6 +141,8 @@ impl ContextServerConfiguration {
         worktree_store: Entity<WorktreeStore>,
         cx: &AsyncApp,
     ) -> Option<Self> {
+        const EXTENSION_COMMAND_TIMEOUT: Duration = Duration::from_secs(30);
+
         match settings {
             ContextServerSettings::Stdio {
                 enabled: _,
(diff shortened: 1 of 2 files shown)


crates/project/src/context_server_store.rs
```diff
@@ -141,6 +141,8 @@ impl ContextServerConfiguration {
         worktree_store: Entity<WorktreeStore>,
         cx: &AsyncApp,
     ) -> Option<Self> {
+        const EXTENSION_COMMAND_TIMEOUT: Duration = Duration::from_secs(10);
```
tomhoule: Is my understanding correct that the extension could be attempting an npm install during that command execution? If that’s the case, we have to bump this timeout (maybe 30 seconds?).

bennetbo: The extension gives us a command that we run ourselves. I am not aware of any Extensions that try to install something when we resolve the command, so 10 seconds should be fine

tomhoule: I got that impression from here: https://github.com/akbxr/zed-mcp-server-context7/blob/c6b427cff4bc5c36df15533ea64608cb47ab81ff/src/mcp_server_context7.rs#L32 — (it’s the one that is problematic in my setup). But I’m not familiar, so I probably misinterpreted.

bennetbo: No, it looks like you're right. Let's increase the timeout, as you suggested

tomhoule: Bumped to 30 seconds.

PR #50090, opened 2026-02-25
editor: Include closing delimiter on same line when folding

### LINKED ISSUE #50002: collapsed line in go (and other languages) should be on one line
### Reproduction steps

The lines are on two lines rather than on one line, as expected, to easily move a group of lines in the file.

Image

current behavior: collapsed line on two lines
expected behavior: collapsed line on one line, for better copy and movement of the lines.

Zed: v0.224.11+stable.175.e4cabf49a18ed03969d84ee15643e9ec81857e97 (Zed) 
OS: Linux Wayland tuxedo 24.04
Memory: 61.9 GiB
Architecture: x86_64
GPU: AMD Radeon 890M Graphics (RADV GFX1150) || radv || Mesa 25.2.8-1~24.04-tux1

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Dnreikronos:
CLOSES: #50002 

When using indent-based folding (the default, `document_folding_ranges = Off`),
collapsed folds in brace-delimited languages displayed the closing delimiter on
a separate line:

fn b() {⋯
}

This extends the fold range in `crease_for_buffer_row` to include the trailing
newline and leading whitespace before a closing `}`, `)`, or `]`, producing
the expected single-line display:

fn b() {⋯}

Whitespace-sensitive languages like Python are unaffected — their terminating
lines don't start with a closing delimiter, so the existing behavior is
preserved.

Screenshots:
Before:
image

After:
image

Release Notes:

- Fixed indent-based code folding to display the closing delimiter (`}`, `)`, `]`) on the same line as the fold placeholder instead of on a separate line ([#50002](https://github.com/zed-industries/zed/issues/50002)).


crates/editor/src/display_map.rs
```diff
@@ -2351,39 +2323,72 @@ impl DisplaySnapshot {
...
-                row_before_line_breaks.row,
-                self.buffer_snapshot()
-                    .line_len(MultiBufferRow(row_before_line_breaks.row)),
-            );
+            let end = if let Some(row) = closing_row {
+                let mut col = 0u32;
+                let mut found_closing_delimiter = false;
+                let mut chars = self.buffer_snapshot().chars_at(Point::new(row, 0));
+                while let Some(ch) = chars.next() {
+                    if ch == '\n' {
```
SomeoneToIgnore: That has to support Windows and \r\n too, right?
We have `LineEnding` that might help a bit, definitely worth a test!

crates/editor/src/display_map.rs
```diff
@@ -2351,39 +2323,72 @@ impl DisplaySnapshot {
...
+                        break;
+                    }
+                    if !ch.is_whitespace() {
+                        found_closing_delimiter = matches!(ch, '}' | ')' | ']')
```
SomeoneToIgnore: That looks quite hardcoded and odd: at worst, we have a way to understand which bracket pairs we have in scope with https://github.com/zed-industries/zed/blob/e5150abe6f6ed1f7c8a5420ce66f20921ba19110/crates/language/src/language.rs#L2295 , but seems that there has to be a better heuristic applied?

crates/editor/src/display_map.rs
```diff
@@ -2351,39 +2323,72 @@ impl DisplaySnapshot {
...
+                    // Include newline and whitespace before closing delimiter,
+                    // so it appears on the same display line as the fold placeholder
+                    Point::new(row, col)
+                } else {
+                    let prev_row = row - 1;
+                    let mut end = Point::new(
+                        prev_row,
+                        self.buffer_snapshot().line_len(MultiBufferRow(prev_row)),
+                    );
+                    while end.row > start.row
```
SomeoneToIgnore: Can we this common code into a well-named `fn`/closure to improve the readability?

SomeoneToIgnore (changes requested):
Overall, it's not very clear why brackets have to be a part of all the heuristic code — if we can solve that and the rest of the comments, would be great.

crates/editor/src/display_map.rs
Dnreikronos: Switched to `scope.brackets()` from the language's bracket pair config. It picks up whatever closing brackets the language defines.

Closing tags like `</div>` are a different story though. HTML's config has `<`/`>` as a pair, not `</`/`>`, so those get their own check. When there's no language set at all (`language_scope_at` returns `None`), it falls back to `}`, `)`, `]`.

Moved all of this into a `line_starts_with_closing_bracket` helper to keep it out of the main fold logic

crates/editor/src/display_map.rs
Dnreikronos: I removed the `chars_at` loop . Now it uses `line_indent_for_buffer_row` for the offset and `take_while(|ch| *ch != '\n')` for the line content. The buffer normalizes line endings to `\n` internally so `\r\n` isn't an issue.

crates/editor/src/display_map.rs
Dnreikronos: Yep, pulled it into a `last_non_blank_row` closure. Both branches use it.

crates/editor/src/editor_tests.rs
```diff
@@ -1469,21 +1463,20 @@ async fn test_fold_with_unindented_multiline_block_comment(cx: &mut TestAppConte
 
     cx.update_buffer(|buffer, cx| buffer.set_language(Some(language), cx));
     cx.set_state(indoc! {"
-        fn main() {
+        ˇfn main() {
```
SomeoneToIgnore, line 1618: This change is repeated twice and both times it seems odd?

We should be able to fold this function out of the original position too, right?
So we should preserve the previous test and add another case nearby (instead of editing), or fix whatever breaks this case instead.

crates/editor/src/display_map.rs
```diff
@@ -2269,6 +2269,36 @@ impl DisplaySnapshot {
...
+
+        if let Some(scope) = snapshot.language_scope_at(Point::new(row, 0)) {
+            let mut brackets = scope.brackets().peekable();
+            if brackets.peek().is_some() {
+                return brackets.any(|(pair, _)| line_text.starts_with(&pair.end));
+            }
+        }
+
+        // Fallback when no language is available or brackets are not configured
+        line_text.starts_with('}') || line_text.starts_with(')') || line_text.starts_with(']')
```
SomeoneToIgnore: That is a hardcode again, and when there's "no language available or brackets or not configured", currently Zed does nothing — inventing something on the fly when fixing a bug might be a bad start, so we should not do that.

crates/editor/src/display_map.rs
```diff
@@ -2269,6 +2269,36 @@ impl DisplaySnapshot {
...
+            .raw_len();
+        let content_start = Point::new(row, indent_len);
+        let line_text: String = snapshot
+            .chars_at(content_start)
+            .take_while(|ch| *ch != '\n')
+            .collect();
+
+        // Closing tags (e.g. `</div>`) are not represented in bracket pair
+        // configs, so check for them separately.
+        if line_text.starts_with("</") {
```
SomeoneToIgnore: Have to repeat again that we do not want the hardcoded things in the generic code.

SomeoneToIgnore (changes requested):
Thank you for the style fixes, we still use hardcoded values it seems, so things have not changed too much, globally.

crates/editor/src/editor_tests.rs
SomeoneToIgnore: Now, this is formally fixed as many other tests seem to work as before, but it means the current version does not always include the closing delimiter on the same line when folding?

E.g.

https://github.com/zed-industries/zed/blob/42e7811d4536ea284078d5461bbd45fd0b589dd8/crates/editor/src/editor_tests.rs#L1199-L1206

and many others — that bit was correct to adjust, and the initial version had it right IMO.

crates/editor/src/editor_tests.rs
```diff
@@ -1390,6 +1390,62 @@ fn test_fold_action_multiple_line_breaks(cx: &mut TestAppContext) {
     });
 }
 
+fn rust_language_with_brackets_and_overrides() -> Arc<Language> {
```
SomeoneToIgnore: We have `language::rust_lang()`, does that not work?

SomeoneToIgnore (commented):
Great, looks like the remaining questions are about the test code only.

crates/editor/src/editor_tests.rs
Dnreikronos: `rust_lang()` doesn't work here as-is. It has no `BracketPairConfig` in its `LanguageConfig`, and `scope.brackets()` reads from `LanguageConfig.brackets.pairs`, not from `brackets.scm`.

So even though `rust_lang()` loads the real `brackets.scm` and `overrides.scm` queries, without bracket pairs configured, `closing_bracket_indent_len()` never finds a match and the closing delimiter doesn't get included in the fold.

I see two ways to go:

1. Keep `rust_language_with_brackets_and_overrides()` I created for these tests, since it sets up both the bracket pairs and the queries.

2. Add `BracketPairConfig` with at least `{}`, `()`, and `[]` to `rust_lang()`'s `LanguageConfig`, matching what Rust has in `config.toml`. Then `rust_lang()` would just work here.

What do you think?

SomeoneToIgnore: I would love to try and keep `rust_lang()` as close to the real Rust-related definitions we have, so would be great to try the 2nd approach.

crates/editor/src/editor_tests.rs
Dnreikronos: > Now, this is formally fixed as many other tests seem to work as before, but it means the current version does not always include the closing delimiter on the same line when folding?
> 
> E.g.
> 
> https://github.com/zed-industries/zed/blob/42e7811d4536ea284078d5461bbd45fd0b589dd8/crates/editor/src/editor_tests.rs#L1199-L1206
> 
> and many others — that bit was correct to adjust, and the initial version had it right IMO.

I created` test_fold_action_without_language` to test the case where there's no language that would pull the closing bracket onto the same line, so it behaves like HTML, where the closing bracket stays on its own line. That way we also cover the scenario where no language is configured. I also adjusted `test_fold_action` to use the `rust_language_with_brackets_and_overrides` helper so it validates that the closing bracket gets folded onto the same line. I'm waiting on confirmation whether we'll go with  `rust_lang()` directly or the helper function I created, once that's decided I'll update the tests and push.

SomeoneToIgnore (approved):
Thank you!

Dnreikronos:
> @Dnreikronos How can it be configured to display like this?

Hi, mate!
That will be the normal behavior. It's not necessary any specific configuration.
I just think that it's not deployed on production yet.

aaxper:
Is there a way to disable this behavior? It also collapses `if (condition) { ... } else {` onto one line, meaning I don't have an arrow to collapse the `else` block (because the line that would have it instead has the one for the `if` block). And, for some reason I'm not sure of, my keybinds also won't work to collapse the `else` block in cases like that. I would much prefer to push the `else` onto the next line and get back the normal functionality.

PR #50097, opened 2026-02-25
agent: Add parent to mention file/directory when available

jmg-duarte:
When mentioning files with the same name (`mod.rs`, `index.ts`, etc) it gets really confusing. This PR aims to mitigate that by showing the first parent.

Dark Mode
Screenshot 2026-02-25 at 15 39 13
Light Mode
Screenshot 2026-02-25 at 15 39 26

Release Notes:

- Agent: Disambiguated mentioned files with the same name by displaying their parent folder's name in the mention chip.


danilo-leal (changes requested):
Hey, thanks for the PR! I think we should follow the tabs pattern here: only display the parent folder's name _if_ there are mentions with the same file name, instead of all the time.

crates/agent_ui/src/mention_set.rs
```diff
@@ -156,8 +163,28 @@ impl MentionSet {
...
+        self.mentions
+            .values()
+            .filter(|(mention_uri, _)| match mention_uri {
+                MentionUri::File { abs_path } | MentionUri::Directory { abs_path } => {
+                    abs_path.file_name() == target_name
+                }
+                _ => false,
+            })
+            .nth(1)
+            .is_some()
```
jmg-duarte: This isn't great algorithmically but for a small number of mentions it should be ok.

I don't have any metrics on how many mentions users use so knowing whether this is ok or must be addressed will be helpful.

Thanks!

jmg-duarte: Fixed it

bennetbo:
Thanks for working on this. This was fixed on main by another PR: 
image

jmg-duarte:
@bennetbo could you at least link the PR? It's the least you guys could do after scoping my PR down, having me fixing it and then ignoring me for months...

bennetbo:
Hey, sorry for the oversight, this was not intentional. We get a lot of PRs and sometimes we lose track of which ones are stale and which ones are not. We implemented it here #56926. I'll make sure to link to this PR when this lands on Stable next week

JosephTLyons:
I've included a link to this PR in [today's stable release](https://github.com/zed-industries/zed/releases/tag/v1.4.2).

PR #50121, opened 2026-02-25
debugger: Fetch and display sources with sourceReference from DAP

### LINKED ISSUE #49934: Debugging Deno test scripts ends up in blank test script
### Reproduction steps

1. Open the sample Deno test file below in Zed
2. Set a breakpoint
3. Start a debug session
4. Step over until execution exits the test file

**main_test.ts**
```ts
/**
 * Basic arithmetic tests.
 */
import { assertEquals } from '@std/assert';
import { describe, test } from '@std/testing/bdd';

describe('basic arithmetic', () => {
   test('adding 1 + 1 returns 2', () => {
      assertEquals(1 + 1, 2, '1 + 1 should equal 2');
   });

   test('concatenating "hello" and "world" returns "hello world"', () => {
      assertEquals(
         'hello' + ' ' + 'world',
         'hello world',
         'string concatenation should join with spaces',
      );
   });
});
```

**deno.jsonc**:

```json
{
   "imports": {
      "@std/assert": "jsr:@std/assert@^1",
      "@std/testing": "jsr:@std/testing@^1"
   }
}
```

Execution ends up in a blank test script from the @std standard library `/@std/testing/1.0.17/https:/jsr.io/@std/testing/1.0.17/_test_suite.ts`

https://github.com/user-attachments/assets/3705559a-f8a4-422e-b074-8072e88cbb82

Debugging should at least show the code from the library file, but ideally stepping into library code could be prevented altogether by adding scripts to an ignore list (see the Chrome Dev Tools [Ignore List](https://developer.chrome.com/docs/devtools/settings/ignore-list)).

Zed: v0.224.11+stable.175.e4cabf49a18ed03969d84ee15643e9ec81857e97 (Zed) 
OS: Linux Wayland Ubuntu 24.04
Memory: 62.5 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3050 Ti Laptop GPU || NVIDIA || 580.126.09

### Attach Zed log file

_No response_

### Relevant Zed settings

<details><a href="https://zed.dev/docs/languages/deno#dap-support">standard</a> .zed/debug.json for Deno</summary>

```json
[
  {
    "adapter": "JavaScript",
    "label": "Deno",
    "request": "launch",
    "type": "pwa-node",
    "cwd": "$ZED_WORKTREE_ROOT",
    "program": "$ZED_FILE",
    "runtimeExecutable": "deno",
    "runtimeArgs": ["test", "--allow-all", "--inspect-wait"],
    "attachSimplePort": 9229
  }
]
```

</details>

Dnreikronos:
Closes #49934

When debugging Deno tests, stepping into standard library code shows a blank file because the debug adapter returns a `sourceReference` instead of a real file path. Per the DAP spec: "If sourceReference > 0, the contents of the source
must be retrieved through the source request (even if a path is specified)."

This PR implements the DAP `source` request so Zed can fetch and display these sources.

### Changes

- **`dap_command.rs`**: Add `SourceCommand` implementing `LocalDapCommand` for `dap::requests::Source`
- **`session.rs`**: Add `fetch_source()` method on `Session` to send the source request
- **`stack_frame_list.rs`**: When `sourceReference > 0`, fetch content, create a read-only buffer, detect language from the source name for syntax highlighting, cache buffers by source reference, set the active debug line, and display in
the editor with the source name as the tab title
- **`stack_trace_view.rs`**: Handle sourceReference frames in the stack trace multibuffer view with the same caching and language detection

### Screenshot
image

image

Release Notes:

- Fixed debugger showing blank file when stepping into sources with `sourceReference` (e.g. Deno standard library code).


crates/debugger_ui/src/session/running/stack_frame_list.rs
```diff
@@ -144,6 +146,7 @@ impl StackFrameList {
             list_filter,
             list_state,
             _refresh_task: Task::ready(()),
+            source_reference_buffers: Default::default(),
```
Anthony-Eid: This request caching should likely be apart of project::debugger::Session layer instead of the UI layer

crates/project/src/debugger/session.rs
```diff
@@ -2106,6 +2106,22 @@ impl Session {
...
+        let request = self.state.request_dap(SourceCommand {
+            source: Some(source),
+            source_reference,
+        });
+        cx.spawn(async move |_this, _cx| {
+            let response = request.await?;
+            Ok(response.content)
+        })
+    }
+
```
Anthony-Eid: We should be caching the String return type here. It could even return a SharedFuture<Entity<Buffer>> to make things more streamline.

crates/debugger_ui/src/session/running/stack_frame_list.rs
```diff
@@ -154,6 +157,25 @@ impl StackFrameList {
...
+    }
+
+    pub(crate) fn cache_source_reference_buffer(
+        &mut self,
+        source_reference: u64,
+        buffer: gpui::Entity<language::Buffer>,
+    ) {
+        self.source_reference_buffers.insert(source_reference, buffer);
+    }
+
```
Anthony-Eid: These function should be remove once we store this state in the Data layer

crates/debugger_ui/src/session/running/stack_frame_list.rs
```diff
@@ -429,78 +462,202 @@ impl StackFrameList {
...
+            })?
+        })
+    }
+
+    fn go_to_source_reference_frame(
+        &mut self,
+        stack_frame: dap::StackFrame,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) -> Task<Result<()>> {
```
Anthony-Eid: I wonder if we could unify both this and go to stack frame, there's a lot of repeated code between the two

Anthony-Eid (changes requested):
Before this is mergable we need to store the responses from source requests onto the data layer instead of the UI layer. 

Two other things this PR is missing
1. An integration test
2. Clearing the loaded buffers, which moving the response to the data layer will do

Dnreikronos:
@Anthony-Eid, first of all, thanks for the feedback.
I adjusted the changes you asked. Can you analyze again?

Dnreikronos:
I tested on stable and the same problem is still there, so I reworked the changes based on the review feedback.

https://github.com/user-attachments/assets/ba7300f5-7988-4175-83ce-b34f251a08a3



Here's what changed:

**Buffer caching moved to Session**

Previously the buffer cache lived in `StackFrameList` (UI layer). Now `Session` owns it via `source_buffers: HashMap<u64, Shared<Task<Result<Entity<Buffer>>>>>`. `fetch_source_buffer` does the DAP request, language detection, read-only setup, and caching all in one place. The UI layer just creates/activates editors. `Shared<Task>` means if multiple callers hit the same source reference, they share one DAP request.

**Unified go_to methods**

`go_to_file_frame` and `go_to_source_reference_frame` were about 70% identical. Merged into `go_to_stack_frame_inner` with a branch on `source_reference > 0`. The tail logic (debug pane tracking + active position) lives in a shared `set_active_debug_position` helper now.

**Editor dedup**

Before, clicking the same source reference frame opened a new tab every time. Now it checks existing panes for an editor with the same buffer and activates that one instead.

**Debug pane tracking fix**

The source reference branch was missing `set_active_debug_pane_id`. The shared helper takes care of it.

**ActiveStackFrame::path -> Option**

Was using `Arc::from(Path::new(""))` as a placeholder for source reference frames. Changed to `Option<Arc<Path>>` since these frames don't have a real path. Nothing downstream reads this field yet.

**debug_panic! guard**

For the case where `source_reference > 0` but `stack_frame.source` is None (shouldn't happen). Panics in debug builds, returns an error in release.

**Test**

Added `test_go_to_source_reference_frame`: checks content is correct, buffer is read-only, and re-clicking the same frame doesn't open a second tab.

**Cleanup**

`source_buffers` gets cleared on disconnect, same as the other caches.

After changes:


https://github.com/user-attachments/assets/e23ef29a-4b8b-4420-8b71-80b0d165b531

PR #50128, opened 2026-02-25
Fix Open tool ignoring permission settings

rtfeldman:
The Open tool was unconditionally prompting for user confirmation on every invocation, ignoring both the `tool_permissions.tools.open.default` setting and the "Always Allow" button. Other tools (copy_path, delete_path, create_directory, etc.) check `ToolPermissionDecision` from settings before deciding whether to call `authorize()`, but the Open tool was missing this check entirely.

This adds the same permission-checking pattern used by the other tools: consult `decide_permission_for_path()` and only prompt when the decision is `Confirm`. If the decision is `Allow`, skip the prompt; if `Deny`, block the call.

Closes AI-60

Release Notes:

- Fixed the Open tool always prompting for confirmation even when "Always Allow" was selected or permissions were configured in settings.

### DIFF
--- a/crates/agent/src/tools/open_tool.rs
+++ b/crates/agent/src/tools/open_tool.rs
@@ -2,13 +2,15 @@ use super::tool_permissions::{
     ResolvedProjectPath, authorize_symlink_access, canonicalize_worktree_roots,
     resolve_project_path,
 };
-use crate::{AgentTool, ToolInput};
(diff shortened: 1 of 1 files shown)


james-elicx:
Is this bug the reason I'm constantly getting asked to confirm a tool after updating today? Any news on when this fix will land in preview?

PR #50133, opened 2026-02-25
project_panel: Add file nesting

quantumsheep:
## Description

This PR adds file nesting support in the project panel.

It supports the exact same features and formats as [VSCode's file nesting](https://code.visualstudio.com/updates/v1_67#_explorer-file-nesting).

It closes discussion #25156.

## Notes

It was quite difficult to understand how the project panel works. I've finally managed to get it working, but please review my changes carefully.

### DIFF
--- a/crates/file_finder/src/file_finder_tests.rs
+++ b/crates/file_finder/src/file_finder_tests.rs
@@ -1827,7 +1827,7 @@ async fn test_history_labels_do_not_include_worktree_root_name(cx: &mut gpui::Te
     let app_state = init_test(cx);
 
     cx.update(|cx| {
-        let settings = *ProjectPanelSettings::get_global(cx);
+        let settings = ProjectPanelSettings::get_global(cx).clone();
         ProjectPanelSettings::override_global(
             ProjectPanelSettings {
                 hide_root: true,
@@ -1927,7 +1927,7 @@ async fn test_history_labels_include_worktree_root_name_when_hide_root_false(
     let app_state = init_test(cx);
 
     cx.update(|cx| {
-        let settings = *ProjectPanelSettings::get_global(cx);
+        let settings = ProjectPanelSettings::get_global(cx).clone();
         ProjectPanelSettings::override_global(
             ProjectPanelSettings {
                 hide_root: false,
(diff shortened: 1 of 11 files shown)


quantumsheep:
@cla-bot check

crates/file_finder/src/file_finder_tests.rs
```diff
@@ -1807,7 +1807,7 @@ async fn test_history_labels_include_worktree_root_name_when_hide_root_false(
     let app_state = init_test(cx);
 
     cx.update(|cx| {
-        let settings = *ProjectPanelSettings::get_global(cx);
+        let settings = ProjectPanelSettings::get_global(cx).clone();
```
smitbarmase, line 1930: I've noticed a lot of these changes. Is it necessary? Could we limit this PR to just file nesting?

crates/project/src/file_nesting.rs
```diff
@@ -0,0 +1,136 @@
+use regex::Regex;
+
```
smitbarmase, line 2: Looks like file nesting is purely Project Panel construct. We can move all this to `project_panel` crate, right?

crates/project_panel/src/project_panel.rs
```diff
@@ -828,9 +845,9 @@ impl ProjectPanel {
             })
             .detach();
 
-            let mut project_panel_settings = *ProjectPanelSettings::get_global(cx);
+            let mut project_panel_settings = ProjectPanelSettings::get_global(cx).clone();
```
smitbarmase, line 848: Same as above comment.

smitbarmase (changes requested):
Thanks. I made an initial pass and added a few suggestions before I could delve into the logic.

crates/file_finder/src/file_finder_tests.rs
quantumsheep: Sadly it **is** required because `HashMap<String, String>` doesn't implement `Copy`

Chandraprakash-Darji:
@smitbarmase Anyone working on this feature, I would like to work this feature

PR #50173, opened 2026-02-26
gpui: Fix `send_file_drop_event` holding mutex during event callback

Albab-Hasan:
`send_file_drop_event` was the only event dispatch site in window.rs that called the event callback while still holding the `MacWindowState` mutex. every other callback site in the file uses the established `take() / drop(lock) / callback()` pattern which releases the lock before invoking user code.

this is a latent code quality issue: no deadlock occurs today because the current file-drop callback chain does not reenter `MacWindowState`'s mutex (`Window::mouse_position()` returns a cached field `Window::refresh()` only marks a dirty flag and `cx.activate()` uses a separate `MacPlatform` mutex). but `parking_lot::Mutex` is not entering so any future platform call added inside a filedrop handler would deadlock immediately and silently.

the fix brings `send_file_drop_event` in line with the rest of the file by taking the callback out of the state dropping the lock invoking the callback then reacquiring the lock to restore the callback and update state.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui_macos/src/window.rs
+++ b/crates/gpui_macos/src/window.rs
@@ -2551,19 +2551,20 @@ fn send_file_drop_event(
     window_state: Arc<Mutex<MacWindowState>>,
     file_drop_event: FileDropEvent,
 ) -> bool {
-    let mut window_state = window_state.lock();
-    let window_event_callback = window_state.event_callback.as_mut();
-    if let Some(callback) = window_event_callback {
-        let external_files_dragged = match file_drop_event {
-            FileDropEvent::Entered { .. } => Some(true),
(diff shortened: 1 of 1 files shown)


Albab-Hasan:
@MrSubidubi hey, can you check this out? only if u have some free time

MrSubidubi (approved):
Nice catch and follow-up, thank you!

MrSubidubi:
You are free to send them, as mentioned above, given that they also aligh with https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md. So no need to hesistate, probably just don't send 5 PRs at one 😅

PR #50207, opened 2026-02-26
Add Vercel AI gateway provider integration

dancer:
## Summary
- add a new `ai_gateway` language model provider in zed using the openai-compatible chat completions path
- register ai gateway in provider wiring, settings schema, and default settings
- add vercel ai gateway icon and provider list entry
- parse ai gateway model capabilities from tags (`tool-use`, `vision`) and supported parameters
- clean up ai gateway oidc auth failures into a concise actionable error message

Release Notes:

- Added Vercel AI Gateway as a new LLM provider in Zed.


crates/language_models/src/provider/vercel_ai_gateway.rs
```diff
@@ -0,0 +1,704 @@
...
+use serde::Deserialize;
+pub use settings::OpenAiCompatibleModelCapabilities as ModelCapabilities;
+pub use settings::VercelAiGatewayAvailableModel as AvailableModel;
+use settings::{Settings, SettingsStore};
+use std::sync::{Arc, LazyLock};
+use ui::{ButtonLink, ConfiguredApiCard, List, ListBulletItem, prelude::*};
+use ui_input::InputField;
+use util::ResultExt;
+
+const PROVIDER_ID: LanguageModelProviderId = LanguageModelProviderId::new("ai_gateway");
```
morgankrey: Think these should be `vercel_ai_gateway` and `Vercel AI Gateway` to match the settings key, env var (`VERCEL_AI_GATEWAY_API_KEY`), and icon (`AiVercel`). Right now users see "AI Gateway" in the model picker but configure `vercel_ai_gateway` in settings JSON.

dancer: true true wanted to be at the top of the list for providers 🤣 updating now!

crates/language_models/src/provider/vercel_ai_gateway.rs
```diff
@@ -0,0 +1,704 @@
...
+    api_key: Option<&str>,
+) -> Result<Vec<AvailableModel>, LanguageModelCompletionError> {
+    let uri = format!("{api_url}/models?include_mappings=true");
+    let mut request_builder = HttpRequest::builder()
+        .method(Method::GET)
+        .uri(uri)
+        .header("Accept", "application/json");
+    if let Some(api_key) = api_key {
+        request_builder = request_builder.header("Authorization", format!("Bearer {}", api_key));
+    }
```
rtfeldman, line 491: Just double-checking: the `/models` endpoint works even if you're not auth'd, yeah?

crates/language_models/src/provider/vercel_ai_gateway.rs
```diff
@@ -0,0 +1,704 @@
...
+        if api_url.is_empty() {
+            API_URL.into()
+        } else {
+            SharedString::new(api_url.as_str())
+        }
+    }
+
+    fn default_available_model() -> AvailableModel {
+        AvailableModel {
+            name: "openai/gpt-4o-mini".to_string(),
```
rtfeldman: Just checking - this is being used for both `default_model` and `default_fast_model`, and usually `default_fast_model` is a smaller/faster model than `default_fast_model`. Is it intentional to have them both be 4o mini?

Also, that's a surprisingly old model to have as the default, but maybe that's intentional too!

crates/language_models/src/provider/vercel_ai_gateway.rs
dancer: yes it does!

crates/language_models/src/provider/vercel_ai_gateway.rs
dancer: oops i forgot to update i always just test with 4o - will change this now thank you for seeing this!

crates/language_models/src/provider/vercel_ai_gateway.rs
```diff
@@ -0,0 +1,704 @@
...
+        .map(ToString::to_string)
+        .unwrap_or_else(|| body.to_string());
+
+    clean_error_message(&message)
+}
+
+fn clean_error_message(message: &str) -> String {
+    let lower = message.to_lowercase();
+
+    if lower.contains("vercel_oidc_token") && lower.contains("oidc token") {
```
rtfeldman, line 330: Is there anything less brittle we can do for this than checking for the presence of hardcoded strings in the error message? 😅

dancer: team is working on this currently... so for now is this okay? and i can send a follow-up pr?

dancer: i can remove it and just let the default show and we are updating the current default error which is this currently 
image
but don't want to confuse users yk?

rtfeldman (commented):
@dancer Looks good overall! A few minor questions, and it looks like it needs a `cargo fmt` for CI to pass. 😄

crates/language_models/src/provider/vercel_ai_gateway.rs
rtfeldman: Yeah follow-up would be great, since it won't affect end user UX for now. 😄

rtfeldman:
Thanks @dancer!

tosage05:
reasoning effort doesn't show up

PR #50216, opened 2026-02-26
project_panel: Add Git status indicators

davidescaccia:
This PR adds Git status badges next to file names in the Project Panel, following my older PR #49802
These are enabled by having "git_status" true.

Screenshot
image

I'd love to hear feedback about any of this :)
Especially feedback on these:

- File name colour is determined only by Git status, the diagnostic badges remain separate. Should diagnostics also affect the filename colour?
-  (Unstaged) Modified files and staged modifications share the same colour, in vscode staged modifications use a brownish colour by default which I could not find the colours. I think differentiating them is definetely something to add.

Release Notes

- Added git status indicators in Project Panel. It can be enabled by setting `git_status_indicator` to `true` in `project_panel` settings.

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -5425,6 +5425,7 @@ impl ProjectPanel {
                 false
             }
         };
+        let git_indicator = git_status_indicator(details.git_status);
 
         let id: ElementId = if is_sticky {
             SharedString::from(format!("project_panel_sticky_item_{}", entry_id.to_usize())).into()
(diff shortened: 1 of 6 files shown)


smitbarmase (changes requested):
Thanks. Aren't we introducing a setting to turn off these indicators, just like we have a setting for the diagnostic count?

smitbarmase (approved):
Thanks! This looks great.

crates/project_panel/src/project_panel.rs
```diff
@@ -7368,5 +7388,30 @@ pub fn par_sort_worktree_entries_with_mode(
...
+    }
+    if git_status.untracked > 0 {
+        return Some(("U", Color::Created));
+    }
+    if git_status.worktree.deleted > 0 {
+        return Some(("D", Color::Deleted));
+    }
+    if git_status.worktree.modified > 0 {
+        return Some(("M", Color::Warning));
+    }
```
colemcanelly, line 7403: Looks like this is a bug, causing modified files to show up with warning colors.

Image

PR #50221, opened 2026-02-26
project_panel: Add `sort_order` settings 

Steffen70:
_(Feature Requests #24962)_

_"Before you mark this PR as ready for review, make sure that you have:"_

* [x] Added a solid test coverage and/or screenshots from doing manual testing
* [x] Done a self-review taking into account security and performance aspects
* [x] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Added a `sort_order` to `project_panel` settings which dictates how files and directories are sorted relative to each other in a `sort_mode`.

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -2492,7 +2492,8 @@ impl ProjectPanel {
                 .collect();
 
         let mode = ProjectPanelSettings::get_global(cx).sort_mode;
-        sort_worktree_entries_with_mode(&mut siblings, mode);
+        let case_sensitive = ProjectPanelSettings::get_global(cx).case_sensitive;
+        sort_worktree_entries_with_mode(&mut siblings, mode, case_sensitive);
         let sibling_entry_index = siblings
             .iter()
             .position(|sibling| sibling.id == latest_entry.id)?;
@@ -3921,6 +3922,7 @@ impl ProjectPanel {
         let auto_collapse_dirs = settings.auto_fold_dirs;
         let hide_gitignore = settings.hide_gitignore;
         let sort_mode = settings.sort_mode;
+        let case_sensitive = settings.case_sensitive;
         let project = self.project.read(cx);
         let repo_snapshots = project.git_store().read(cx).repo_snapshots(cx);
 
@@ -4155,6 +4157,7 @@ impl ProjectPanel {
                         par_sort_worktree_entries_with_mode(
                             &mut visible_worktree_entries,
                             sort_mode,
+                            case_sensitive,
                         );
                         new_state.visible_entries.push(VisibleEntriesForWorktree {
                             worktree_id,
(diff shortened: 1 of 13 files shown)


docs/src/reference/all-settings.md
```diff
@@ -5031,6 +5030,34 @@ Run the {#action theme_selector::Toggle} action in the command palette to see a
...
+### Case-Sensitive
+
+- Description: Whether to sort files and directories case-sensitive in the project panel.
+- Setting: `case_sensitive`
+- Default: `false`
+
+**Options**
+
+1. Natural soring
+
```
Palanikannan1437, line 5033: Typo: 'Natural soring' should be 'Natural sorting'

Why: Documentation quality — visible typo in settings reference

Suggested fix: Change 'Natural soring' to 'Natural sorting'

Just was trying agentic reviews on this PR and found this :p

Steffen70: Thx ... I'll fix that

crates/settings/src/vscode_import.rs
```diff
@@ -805,6 +805,16 @@ impl VsCodeSettings {
                 .read_bool("problems.decorations.enabled")
                 .and_then(|b| if b { Some(ShowDiagnostics::Off) } else { None }),
             sort_mode: None,
+            sort_order_lexicographic: self.read_enum(
+                "search.quickOpen.history.filterSortOrder",
```
dinocosta: Not super familiar with VS Code's settings, but I wonder if this should be the setting that used to be referenced in the PR's description, `explorer.sortOrderLexicographicOption`? 🤔

crates/project_panel/src/project_panel.rs
```diff
@@ -847,6 +847,11 @@ impl ProjectPanel {
                     if project_panel_settings.sort_mode != new_settings.sort_mode {
                         this.update_visible_entries(None, false, false, window, cx);
                     }
+                    if project_panel_settings.sort_order_lexicographic
+                        != new_settings.sort_order_lexicographic
+                    {
+                        this.update_visible_entries(None, false, false, window, cx);
+                    }
```
dinocosta, line 852: Not sure if you want to tackle it on this Pull Request, but I believe we could probably just check if one of these conditions is true and then call `this.update_visible_entries` once, as it's always called with the same arguments, regardless of which setting changed 👀

dinocosta (approved):

crates/settings/src/vscode_import.rs
Steffen70: That's the correct VS Code key: "explorer.sortOrderLexicographicOptions"

crates/project_panel/src/project_panel.rs
smitbarmase: Fair, optimization. I don't think it can be triggered through the Settings UI, but maybe replacing settings.json might trigger this. Can we tackle it in a separate PR.

PR #50246, opened 2026-02-26
acp: Fix stale ACP reasoning-effort options when model switch keeps same config IDs

nikhil-pandey:
## Problem
Zed's ACP config-options UI was only rebuilding selectors when the set of config option IDs changed.

For many model switches, the IDs stay the same (`mode`, `model`, `reasoning_effort`) while the valid values for `reasoning_effort` change by model. As a result, the picker could show stale values (for example, missing `xhigh` on `gpt-5.3-codex`, or showing extra `xhigh` after switching away).

This is especially problematic for providers like Copilot that expose multiple agents/models with different reasoning-level capabilities.

## Fix
Rebuild ACP config-option selectors on every `config_option_update`, not only when config IDs change.

This refreshes cached picker entries whenever model-specific option values change, even if option IDs are unchanged.

## User Impact
- Reasoning-effort picker now reflects the selected model immediately on Claude <-> GPT transitions.
- Prevents stale or invalid effort choices in ACP sessions.

## Validation
- Manual validation: switch across models with different reasoning-effort sets and confirm picker updates immediately.
- Local `cargo check -p agent_ui` remains blocked by unrelated pre-existing `livekit-protocol` compile errors in this checkout.

## Files Changed
- `crates/agent_ui/src/acp/config_options.rs`

Release Notes:

- acp: Fix for config selectors not always being refreshed


benbrandt (approved):
Thanks!

PR #50248, opened 2026-02-26
New "Filter Selection Through Command"

### LINKED ISSUE #12598: Run shell command on selected region

kanocz:
Closes #12598

implementing pipe selection to command and replace original selection or copy to clipboard or create new document with output

image
image
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5415,6 +5415,7 @@ dependencies = [
  "multi_buffer",
  "ordered-float 2.10.1",
  "parking_lot",
+ "picker",
  "pretty_assertions",
  "project",
  "rand 0.9.2",
--- a/crates/editor/Cargo.toml
+++ b/crates/editor/Cargo.toml
@@ -57,6 +57,7 @@ linkify.workspace = true
 log.workspace = true
 lsp.workspace = true
 markdown.workspace = true
+picker.workspace = true
 menu.workspace = true
 multi_buffer.workspace = true
 ordered-float.workspace = true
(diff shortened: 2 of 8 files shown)


SomeoneToIgnore:
Thank you, this is something we aim to solve with extensions one day, and nothing we want to have in Zed's core editor code.

PR #50250, opened 2026-02-26
remote: Add CLI file-opening support for remote terminal sessions

raghavpillai:
Adds `zed <file>` support from within remote terminal sessions — the equivalent of VS Code's `code` command that people have been asking about in #32214.

I built [zed-remote-open](https://github.com/raghavpillai/zed-remote-open) as a stopgap for this a while back, but it relied on a separate TCP bridge + SSH reverse tunneling which was clunky to set up. This implements the same idea natively using @xJonathanLEI's suggested approach from that discussion: reuse the `remote_server` binary in a CLI mode and communicate over a Unix socket.

### How it works

The remote server spins up a Unix socket listener (`cli.sock`) alongside the existing session. A small wrapper script gets installed at `~/.zed/bin/zed` on the remote host, and the terminal session gets `PATH` + `ZED_REMOTE_CLI_SOCKET` injected so the command is discoverable out of the box.

When you run `zed file.rs:42` in a remote terminal:

1. Wrapper invokes the remote server binary in `cli` mode
2. CLI client connects to the Unix socket, sends the path
3. Listener parses the request (supports `path:line:column` and `--wait`)
4. Sends an `OpenPathOnClient` RPC over the existing SSH channel
5. Project handler on the client emits an event, workspace opens the file

No extra binaries, no version mismatch concerns, no SSH reverse tunneling needed.

### What changed

- **Proto layer** (`app.proto`, `zed.proto`, `proto.rs`): New `OpenPathOnClient`/`OpenPathOnClientResponse` message pair at fields 431-432
- **`cli_client.rs`** (new): CLI client that connects to the socket, resolves it via env var / identifier / auto-discovery
- **`cli_listener.rs`** (new): Unix socket listener on the server side, parses requests, forwards RPC. Includes tests for path parsing
- **`server.rs`**: New `Cli` subcommand, socket lifecycle, `install_cli_wrapper()` that writes the shell script
- **`project.rs`**: RPC handler + `OpenPathOnClient` event emission
- **`workspace.rs`**: Handles the event, opens the file via `open_abs_path`
- **`terminals.rs`**: Injects env vars into remote shell sessions
- **`remote_client.rs`**: Exposes `unique_identifier()` getter

### Limitations / future work

- Line/column navigation: the workspace currently just opens the file. Wiring up `row`/`column` to scroll to a specific position needs the `editor` crate which `workspace` doesn't depend on. Could be added as a follow-up.
- `--wait` flag: parsed and forwarded but not fully wired yet (would need to hold the CLI process open until the buffer is closed)
- Only tested on Linux so far

Would appreciate any feedback on the approach. Happy to adjust based on how the team wants this to fit into the broader remote architecture.

Release Notes:

- Added `zed` CLI command support in remote terminal sessions, allowing `zed <file>` to open files on the local Zed client from a remote terminal.


raghavpillai:
@cla-bot check

raghavpillai:
@cla-bot check

crates/project/src/terminals.rs
```diff
@@ -613,6 +613,24 @@ fn create_remote_shell(
 ) -> Result<(Shell, HashMap<String, String>)> {
     insert_zed_terminal_env(&mut env, &release_channel::AppVersion::global(cx));
 
+    // Inject CLI environment for `zed` command in remote terminals.
+    // The socket path follows the same convention as the remote server state dir.
+    let identifier = remote_client.read(cx).unique_identifier().to_string();
+    let cli_socket_path = format!(
+        "$HOME/.local/share/zed/server_state/{}/cli.sock",
```
Dabsunter, line 620:
```suggestion
        "$XDG_DATA_HOME/zed/server_state/{}/cli.sock",
```
(or any proper handling for [XDG Base Directories](https://specifications.freedesktop.org/basedir/latest/))

crates/project/src/terminals.rs
```diff
@@ -613,6 +613,24 @@ fn create_remote_shell(
...
+    );
+    env.insert("ZED_REMOTE_CLI_SOCKET".to_string(), cli_socket_path);
+
+    // Prepend ~/.zed/bin to PATH so the `zed` wrapper script is discoverable
+    let existing_path = env.get("PATH").cloned().unwrap_or_default();
+    let new_path = if existing_path.is_empty() {
+        "$HOME/.zed/bin:$PATH".to_string()
+    } else {
+        format!("$HOME/.zed/bin:{existing_path}")
+    };
```
Dabsunter, line 631: Since the wrapper script is specific to the server, wouldn't `~/.zed_server/bin` be more appropriate?

benjaminhr:
@raghavpillai do you have a chance to finish this? it would be super useful!

anikinmd:
@raghavpillai I miss this feature a lot. Are you planning to continue working on this PR? If not, I’d be happy to pick it up.

PR #50260, opened 2026-02-26
agent_ui: Add more improvements to the subagent UI

danilo-leal:
Follow-up to https://github.com/zed-industries/zed/pull/50032:

- Add divider in the full screen view to expose what part of a subagent's output is actually sent to the parent model
- Auto-expand the subagent card if there's only one running on a turn
- Adjust errors display within the subagent card

Release Notes:

- N/A


benbrandt (approved):
thanks, let's give it a try!

PR #50264, opened 2026-02-27
git_panel: Respect sort by path in tree view

oomathias:
https://github.com/user-attachments/assets/11a340e4-de00-4dca-ba97-047e86289377

Sort By Path had no effect when Tree View is enabled. 
Sort By Path is now respected in Tree View. 
As per the setting description Sort By Path is expected to disable the sort by status (aka disable grouping by statuses)
Now when Sort By Path and Tree View are enabled, grouping is disabled. (see last state in video)

Release Notes:

- Fix Tree View mode in Git Panel to supports no grouping

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -211,17 +211,15 @@ fn git_panel_context_menu(
                 Some(Box::new(ToggleTreeView)),
                 move |window, cx| window.dispatch_action(Box::new(ToggleTreeView), cx),
             )
-            .when(!state.tree_view, |this| {
-                this.entry(
-                    if state.sort_by_path {
-                        "Sort by Status"
-                    } else {
-                        "Sort by Path"
-                    },
-                    Some(Box::new(ToggleSortByPath)),
-                    move |window, cx| window.dispatch_action(Box::new(ToggleSortByPath), cx),
-                )
-            })
+            .entry(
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Thanks so much for your work here!

Unfortunately I don't think sorting by status makes sense in combination with the tree view, since it is not especially clear what should happen for nested files in directories. As a result I'm going to close this.

If you feel that this should still be accepted, please include more examples of the behavior in more complex scenarios.

oomathias:
@ChristopherBiscardi

Thanks for reviewing.

I agree that sorting/grouping by status in Tree View is ambiguous for nested directories. That ambiguity is the issue I’m trying to address.

With a flat view, the Tracked/Untracked grouping is manageable because files are listed individually. But with Tree View enabled, status grouping can duplicate the same directory across sections, which makes reviewing changes harder because you have to mentally merge multiple partial trees.

What we currently have:
```text
Tracked
├── src/
│   └── utils.js           modified
└── app.js                 modified

Untracked
├── src/
│   └── notes.md           untracked
└── notes.md               untracked
```

What I would expect when `Sort by Path` is enabled:
```text
.
├── src/
│   ├── utils.js           modified
│   └── notes.md           untracked
├── app.js                 modified
└── notes.md               untracked
```

I guess a better named option would be "disable status grouping".

PR #50327, opened 2026-02-27
git: Fix conflicted paths not getting cleared

cole-miller:

Release Notes:

- Fixed a bug where files would still be marked as having git conflicts after resolving them.

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -1024,7 +1024,7 @@ impl GitGraph {
                     }
                 }
             }
-            RepositoryEvent::BranchChanged | RepositoryEvent::MergeHeadsChanged => {
+            RepositoryEvent::BranchChanged => {
                 self.pending_select_sha = None;
                 // Only invalidate if we scanned atleast once,
                 // meaning we are not inside the initial repo loading state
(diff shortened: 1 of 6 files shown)


crates/git_graph/src/git_graph.rs
```diff
@@ -1024,7 +1024,7 @@ impl GitGraph {
                     }
                 }
             }
-            RepositoryEvent::BranchChanged | RepositoryEvent::MergeHeadsChanged => {
```
RemcoSmitsDev, line 1027: Not sure if this is intended, but removing this event will break the auto update of the graph for new commits. Just wanted to point it out.

cole-miller: I think BranchChanged is also emitted when the HEAD commit moves, right? Was MergeHeadsChanged picking up any situation where HEAD hasn't changed but we do want to update the graph?

RemcoSmitsDev: Yeah it should be emitted! But i had a case where it wasn't but the graph should been updated. But I can't remember the exact case, tried testing what I tested back then but can't reproduce it anymore. So you can probably ignore my comment :)

cole-miller: :+1: one thing that I'm sure we don't catch with the current setup is when you do a fetch and remote branches update, I would be down to emit an event in that case so the git graph knows to rebuild (if it shows remote branches). And happy to add another event to replace this one if it turns out the graph needs it

RemcoSmitsDev: Yeah your right we don't update on remote change yet. Happy to look into that if you haven't done it already.

cole-miller: That'd be great!

PR #50372, opened 2026-02-28
Fix incorrect error response

OmChillure:
### Fixes #50254

This change clarifies a misleading Agent error when generation stops due to per-response output limits.

Previously, Zed surfaced Max tokens reached, which implied the full context window (max_tokens) was exhausted. In affected cases, the actual limit was max_output_tokens for a single response.
The error now explicitly states that the output-token cap for the current response was reached, includes the limit value when available, and provides actionable guidance to send continue to resume.

### Updated:

Runtime error text in ACP thread handling

Release Notes

- Agent: Improve max token error


benbrandt (approved):
Cleaned it up a bit, but thanks!

PR #50379, opened 2026-02-28
agent_ui: Mask API key input in Add LLM provider modal

cppcoffee:
Release Notes:

- Added Mask API key input in Add LLM provider modal

截屏2026-02-28 17 35 22


danilo-leal (changes requested):
Hey, thanks for the PR! I don't think we should mask this by default, though. If we did that, how could you double check whether the API key is correct, like, free from typos or any other mistake? I think the pattern we'd like to follow here, if anything, is having the typical input "eye" icon button that allows you to toggle the mask on and off.

danilo-leal (approved):
Thank you so much! Pushed a change here embedding the hide/show icon button directly in the input field. It would be great later on to add the ability to toggle it with the keyboard somehow!

PR #50381, opened 2026-02-28
language_models: Add Anthropic-compatible provider support in settings

cppcoffee:
Release Notes:

- Added Anthropic-compatible provider support in settings

1. add anthropic-compatible provider via menu

https://github.com/user-attachments/assets/14b33d59-ae08-4215-b05f-9f3896a4a6f2

2. use anthropic-compatible

https://github.com/user-attachments/assets/8bc7f1c5-644e-4528-942e-29150c35fed0


anantdgoel:
Thank you for putting up this change! I cleaned up some of the code because the anthropic specific code had moved folders since you opened this PR, and I added some tests + a new logo.

PR #50393, opened 2026-02-28
agent: add parallel_agents tool for multi-agent collaboration

ghost:
Adds a new `parallel_agents` tool that lets the root orchestrator agent
fan out independent subtasks to multiple workers simultaneously and
collect all results in one shot.

- New `ParallelAgentsTool` using `join_all` to drive workers concurrently
- New `MultiAgentCollaborationFeatureFlag` (staff-enabled, dark for everyone else)
- Bumps max subagent depth to 2 when flag is active
- Tool only registered at depth 0 — workers cannot trigger a second fan-out

Release Notes:
- Added: parallel_agents tool for multi-agent collaboration

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -9,7 +9,8 @@ use crate::{
 use acp_thread::{MentionUri, UserMessageId};
 use action_log::ActionLog;
 use feature_flags::{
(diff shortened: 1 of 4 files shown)


benbrandt:
Hi thanks for this but we'll be releasing sub agents soon which already allow for this behavior if the model does parallel tool calls. 

We had an approach like this at first but with parallel tool calls we can also stream them in and start them as they arrive, and also better track cancellation + tool status because they are all separate, which is why we moved away from this interface

PR #50401, opened 2026-02-28
git_graph: Allow having multiple git graphs open for different repositories

RemcoSmitsDev:
Before this PR you could only have the git graph open for the active repository, this PR changes that. 
So you can have 1 git graph open per repository, allowing you to open multiple different graph at the same time.

**Example**:

https://github.com/user-attachments/assets/9775108f-826a-476f-95de-46abcc1777a6

**Example Persistence**

https://github.com/user-attachments/assets/dbcf6692-7a67-46d9-a7ae-43a7c9a35818

cc @Anthony-Eid I removed the switch repo test since it's no longer a thing.

Release Notes:

- N/A


RemcoSmitsDev:
@Anthony-Eid I tried using the repo base path as that is stable enough, but the issue is that there is a race condition with loading of the repositories and deserializing the git graph. Because most of the time there are no repositories loaded yet, so we would error out resulting in no git graph for initial loading. 

One thing we could do is store the repo base path and make the repoid optional, and check when we receive a repo event if we not already have a repoid. if not determine it based on the base path, set the id and handle the event. This way we would do kinda lazy loading of the graph based on events.

Anthony-Eid (approved):
I paired heavily with Remco on this, changes are good with me

PR #50411, opened 2026-02-28
markdown: Fix markdown copy range resolution

### LINKED ISSUE #42958: AI: copying and pasting from AI window sometimes gives incomplete markdown
### Summary

When I copy and paste a paragraph that begins with bolded text from the AI window, I do not get the `**` at the beginning.

### Description

Steps to trigger the problem:
1. Convince the AI to give you a list of things with bolded items at the front.
2. Select as shown in the screenshot below.
3. Paste into the "prompt" window.

Here is a screen shot

Image

**Expected Behavior**:

I expected to see `**How should the new flow work?** ...`

**Actual Behavior**:

I see `How should the new flow work?** ...`

### Model Provider Details
N/A

Zed: v0.212.6 (Zed) 
OS: macOS 15.7.1
Memory: 36 GiB
Architecture: aarch64

```log

```

</details>

pkondzior:
## Summary

Markdown copy could break formatting when selection edges landed inside wrapper syntax, producing output like `bold**` instead of `**bold**`. This PR makes copy range resolution span-aware, so copied markdown keeps valid boundaries.

## Why

The old flow used raw source slicing. That works for plain text but is fragile for formatted markdown. Using parse-time span metadata makes selection-to-source mapping consistent for wrappers like bold, links, and inline code.

## What changed

The markdown parser now keeps wrapper bounds for inline code, and markdown copy resolution now runs through source-span normalization and wrapper expansion logic. Copy still falls back to clamped raw selection when parsed source is stale. Tests were expanded across parser, resolver, and markdown integration paths.

## Cases covered

| Case | Description | Before | After |
|---|---|---|---|
| 1. Full formatted token selection | Selecting visible text inside a full formatted token now copies the full markdown wrapper. | Case 1 before | Case 1 after |
| 2. Partial formatted token selection | Selecting only part of formatted content now copies only that part, without leaking wrapper syntax. | Case 2 before | Case 2 after |
| 3a. Link boundary handling | Selections near link edges no longer leak broken suffixes like `](...)`. | Case 3a before | Case 3a after |
| 3b. Link boundary handling (additional) | Additional boundary scenario for links now resolves cleanly. | Case 3b before | Case 3b after |
| 4. Multiline list boundary handling | Selections spanning list items now start cleanly and do not inherit suffix fragments from prior items. | Case 4 before | Case 4 after |
| 5. List boundary with inline style wrapper | If a selection begins at a list-item boundary inside visible styled text, copy now preserves complete inline style wrappers. | Case 5 before | Case 5 after |

Closes #42958

- [x] Added solid test coverage and/or manual testing screenshots
- [x] Completed self-review, including security and performance considerations
- [x] Aligned UI changes with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Fixed markdown copy behavior for formatted selections.

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -1083,7 +1083,7 @@ mod tests {
                         MarkdownEvent::SubstitutedText(parsed) => {
                             rendered_text.push_str(parsed.as_str())
                         }
-                        MarkdownEvent::Text | MarkdownEvent::Code => {
+                        MarkdownEvent::Text | MarkdownEvent::Code { .. } => {
                             rendered_text.push_str(&text[range.clone()])
                         }
                         _ => {}
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -1,5 +1,6 @@
 pub mod parser;
 mod path_range;
+mod source_spans;
 
 use base64::Engine as _;
 use futures::FutureExt as _;
@@ -10,6 +11,7 @@ use language::LanguageName;
 use log::Level;
 pub use path_range::{LineCol, PathWithRange};
 use settings::Settings as _;
+pub use source_spans::MarkdownSpan;
 use theme::ThemeSettings;
 use ui::Checkbox;
 use ui::CopyButton;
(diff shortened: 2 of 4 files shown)


bennetbo:
Hey, sorry for taking so long to look at this. I think in general this would be nice to have, but I'm wondering if we really need all the complexity introduced here, wondering what we can do to make the code simpler. Closing for now because of conflicts.

Issue #46709, opened 2026-01-13
Zed fails to reflect changes made to directory in side panel


bcheidemann:
### Reproduction steps

1. Run `zed <my-repo>`
2. Expand a folder in the repo
3. Run a command which makes a lot of changes to that folder

Observe that changes are not reflected in side panel.

### Current vs. Expected behavior

<img width="1530" height="858" alt="Image" src="https://github.com/user-attachments/assets/59645738-cc49-4638-a8b9-0a32434db64e" />

> [!NOTE]
> The screenshot is from a previous version of Zed but I have updated and re-tested. The bug still exists on the latest version.

Files listed by `ls` should display in the side panel.

### Zed version and system specs

Zed 0.218.7 
41bab92dac34777ae4256e5041f7b7236fd16cdd

0.218.7+stable.118.41bab92dac34777ae4256e5041f7b7236fd16cdd

OS: Linux (Ubuntu)

### Attach Zed log file

<details><summary>Zed.log</summary>

No logs are produced.

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json
{
  "ui_font_size": 16,
  "buffer_font_size": 16,
  "theme": {
    "mode": "system",
    "light": "Gruvbox Dark Soft",
    "dark": "One Dark"
  },
  "vim_mode": true,
  "vim": {
    "use_system_clipboard": "always"
  },
  "wrap_guides": [80, 120],
  "double_click_in_multibuffer": "open",
  "buffer_font_family": "Hack Nerd Font",
  "cursor_blink": false,
  "hide_mouse": "never",
  "lsp": {
    "rust-analyzer": {
      "initialization_options": {
        "rust": {
          "analyzerTargetDir": true
        }
      }
    }
  }
}
```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

n/a

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

Zachiah:
@bcheidemann Thanks for the report!

I wonder if this is a dupe of #46580

bcheidemann:
@Zachiah this was fixed but appears to have regressed recently.

```sh
$ zed --version
Zed 1.12.0 f96212f2c50f54d93712fa130d6226b1ce7d76b5  – /home/ben/.local/zed.app/libexec/zed-editor
```

PR #50412, opened 2026-02-28
Fix Linux watcher cleanup for recreated directories

Chriss4123:
## Problem
- On Linux, non-recursive watcher registrations remained path-cached after deleting and recreating a directory in the same session.
- The recreated directory was not re-watched, so newly created child entries under that path could be missing.

## Summary
- Remove directory watcher registrations when worktree paths are removed from snapshot state.
- Ensure recreated directories can be watched again on Linux by allowing `scan_dir` to re-add fresh watches.
- Add a Linux integration regression test for directory delete/recreate path reuse and child file creation.

## Testing
- `cargo test -p project --features test-support --test integration test_recreated_directory_receives_child_events -- --exact`
- `cargo test -p project --features test-support --test integration test_rescan_and_remote_updates -- --exact`

## Related
- #46709

Release Notes:

- Fixed Linux worktree file watching so child entries appear after deleting and recreating a directory at the same path.

### DIFF
--- a/crates/project/tests/integration/project_tests.rs
+++ b/crates/project/tests/integration/project_tests.rs
@@ -5359,6 +5359,52 @@ async fn test_rescan_and_remote_updates(cx: &mut gpui::TestAppContext) {
     });
 }
 
+#[cfg(target_os = "linux")]
+#[gpui::test(retries = 5)]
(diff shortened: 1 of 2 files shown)


Chriss4123:
@codex review

PR #50413, opened 2026-02-28
Format proto files

Anthony-Eid:
This fixes a problem where editing a proto file would make the proto LSP show a ton of warnings and avoids the giant diff that formatting a .proto file would cause. 

I also added a lint check when our clippy script is ran locally

Release Notes:

### DIFF
--- a/script/clippy
+++ b/script/clippy
@@ -16,4 +16,8 @@ if [[ -z "${GITHUB_ACTIONS+x}" ]]; then
 
     which typos >/dev/null 2>&1 || exit 0
     typos --config typos.toml
+
+    which buf >/dev/null 2>&1 || exit 0
+    buf lint crates/proto/proto
+    buf format --diff --exit-code crates/proto/proto
 fi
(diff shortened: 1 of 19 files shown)


script/clippy
```diff
@@ -16,4 +16,8 @@ if [[ -z "${GITHUB_ACTIONS+x}" ]]; then
 
     which typos >/dev/null 2>&1 || exit 0
     typos --config typos.toml
+
+    which buf >/dev/null 2>&1 || exit 0
+    buf lint crates/proto/proto
+    buf format --diff --exit-code crates/proto/proto
```
MrSubidubi: Should we not add a step for this to the protobuf migrations check? We install the buf CLI there already, so why not also use it there for linting and formatting checks?

MrSubidubi (dismissed):

script/clippy
Anthony-Eid: We should! I was planning on asking for your help with that on Monday, felt bad pinging you on a weekend

PR #50430, opened 2026-03-01
Add `recent_changes` tool to surface unsaved buffer diffs as agent context

ghost:
The agent currently has no way to know what files the user is actively 
editing unless they explicitly @-mention them. This means the agent can 
easily clobber unsaved work or miss important in-progress changes when 
making edits.

This PR adds a `recent_changes` tool that lets the agent call a single 
tool and immediately see every dirty buffer in the project as a unified 
diff — what changed, in which file, since the last save.

The tool is read-only, always-on, and follows the exact same pattern as 
the existing `diagnostics` and `grep` tools. No new architecture, no 
feature flag needed.

Internally it snapshots dirty buffers synchronously via `BufferStore`, 
then loads the saved-on-disk version asynchronously via `Fs` and runs 
`diffy::create_patch` to produce the diff. Three cases are handled: 
files modified since last save, files never saved yet, and files whose 
saved version can't be read from disk.

Also promotes `diffy` to a workspace dependency since it was already 
pinned inline in the `language` crate.

Release Notes:
- Added: `recent_changes` tool for the agent to discover unsaved edits 
  across the project without requiring explicit file mentions


BitStrider:
Thanks for the label, @SomeoneToIgnore! I was wondering if I could get an assignee for this one, or if there's a specific protocol I should follow to move it toward a review? Happy to make any adjustments needed. Thanks!

PR #50437, opened 2026-03-01
gpui(web): Pass events to input handlers and handle IME composition

someone13574:
Currently the web backend for gpui doesn't send any events to the `InputHandler`'s like `EntityInputHandler` which are needed for the input example and the editor crate, among others. This PR makes it pass those events, in addition to also dealing with composition events so that IME works. It adds an invisible input element to listen for composition events, since canvases don't receive them.

Release Notes:

- N/A


Veykril (approved):
Thanks!

PR #50440, opened 2026-03-01
markdown_preview: Enable parser options individually

### LINKED ISSUE #50170: Markdown not rendering content in preview mode
### Reproduction steps

Image

Image

It should show up the content in preview mode

Zed: v0.225.9+stable.181.cbc2b428b9003b31077f91f06f8912880b1dbcf4 (Zed) 
OS: macOS 26.3
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

yeskunall:
Closes #50170.

Release Notes:

- Enable parser options individually to avoid rendering issues


crates/markdown_preview/src/markdown_parser.rs
```diff
@@ -14,15 +14,24 @@ use std::{
 };
 use ui::SharedString;
 
+const PARSE_OPTIONS: Options = Options::ENABLE_TABLES
```
MrSubidubi: Since this matches with what is found in the markdown crate, can we not instead re-export the options from the markdown crate and use that here instead? We want to migrate to that crate anyway at some point, so perhaps better to introduce the dependency here then?

yeskunall: > We want to migrate to that crate anyway at some point

Ah, OK that’s good to know.

> Since this matches with what is found in the markdown crate, can we not instead re-export the options from the markdown crate and use that here instead?

Perfectly sound and should be possible -- I’ll do that. Thanks Finn!

crates/markdown_preview/src/markdown_parser.rs
```diff
@@ -3076,6 +3074,26 @@ More text
         );
     }
 
+    #[gpui::test]
```
yeskunall, line 3077: Left the regression tests I added intact since they test the preview-specific parsing behavior.

LMK if there is a better way/place to write these tests.

MrSubidubi: Think that's fine to keep them here, especially with the prospect of the migration 🔜 ™️

MrSubidubi (approved):

PR #50443, opened 2026-03-01
agent_ui: Open local file links in thread view

cppcoffee:
Release Notes:

- Added open local file links in thread view

open link like: `/Users/foobar/Desktop/zed/crates/git_ui/src/commit_view.rs:193`

https://github.com/user-attachments/assets/086a24fc-e877-480e-9edb-397b40a80042

### DIFF
--- a/crates/agent_ui/src/conversation_view/thread_view.rs
+++ b/crates/agent_ui/src/conversation_view/thread_view.rs
@@ -9,14 +9,24 @@ use acp_thread::{ContentBlock, PlanEntry};
 use cloud_api_types::{SubmitAgentThreadFeedbackBody, SubmitAgentThreadFeedbackCommentsBody};
 use editor::actions::OpenExcerpts;
 use feature_flags::AcpBetaFeatureFlag;
+use project::ProjectPath;
 
 use crate::message_editor::SharedSessionCapabilities;
+<<<<<<< HEAD
 
 use gpui::List;
 use heapless::Vec as ArrayVec;
 use language_model::{LanguageModelEffortLevel, Speed};
 use settings::{SidebarSide, update_settings_file};
 use ui::{ButtonLike, SpinnerLabel, SpinnerVariant, SplitButton, SplitButtonStyle, Tab};
+=======
+use gpui::{AsyncWindowContext, Corner, List};
+use heapless::Vec as ArrayVec;
+use language_model::{LanguageModelEffortLevel, Speed};
+use settings::update_settings_file;
+use ui::{ButtonLike, SplitButton, SplitButtonStyle, Tab};
+use workspace::ItemHandle;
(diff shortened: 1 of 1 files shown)


crates/agent_ui/src/connection_view/thread_view.rs
```diff
@@ -7755,7 +7755,91 @@ pub(crate) fn open_link(
...
         })
+    } else if let Some((abs_path, row)) = parse_local_file_link(url.as_ref()) {
+        workspace.update(cx, |workspace, cx| {
+            let project = workspace.project();
+            let item = if let Some(project_path) =
+                project.update(cx, |project, cx| project.find_project_path(&abs_path, cx))
+            {
+                workspace.open_path(project_path, None, true, window, cx)
+            } else {
+                workspace.open_abs_path(abs_path, workspace::OpenOptions::default(), window, cx)
```
benbrandt: I think we only want to do this if we have a project path

crates/agent_ui/src/connection_view/thread_view.rs
```diff
@@ -7755,7 +7755,91 @@ pub(crate) fn open_link(
...
 }
+
+fn parse_local_file_link(url: &str) -> Option<(std::path::PathBuf, u32)> {
+    let path_with_position = util::paths::PathWithPosition::parse_str(url.trim());
+    if !path_with_position.path.is_absolute() {
+        return None;
+    }
+
+    let line = path_with_position.row?;
+    if path_with_position.column.is_some() {
```
benbrandt: Why do we skip if we have a column? Could we at least still go to the line?

benbrandt (changes requested):
This seems nice, just a new notes

crates/agent_ui/src/connection_view/thread_view.rs
cppcoffee: It skips `:line:column` links because that code path was written as line-only, not because the editor can’t handle columns. In `thread_view.rs:8132`, `parse_local_file_link` only returns `(PathBuf, row)`, and the caller later hardcodes column `0` in `thread_view.rs:8112`. The guard in `thread_view.rs:8139` is basically saying: “if the link includes a column, reject it rather than pretending we handled it.”

benbrandt: Right, but if we have a row, can't we get close?

cppcoffee: Modified it to support absolute file links containing optional line and column positions, preventing unsaved buffers from being opened for missing local paths.

cppcoffee:
Rebase `main` branch done

cppcoffee:
Fixed https://github.com/zed-industries/zed/issues/55273

yeskunall:
See https://github.com/zed-industries/zed/pull/53229.

PR #50445, opened 2026-03-01
agent: Enable @diagnostics for acp agents

raphaelluethy:
This change allows the user to send the error messages from the diagnostics panel to the acp agent as text. Since the editor already has all the information, I think this would be a nice addition to the acp workflow.

Whats actually happening is that we just copy the contents of `mentionUri::Diagnostics` to text and send that to the acp instead of linking up the LSP errors directly.

Screenshots:
CleanShot 2026-03-01 at 16 57 28@2x
CleanShot 2026-03-01 at 16 58 09@2x
CleanShot 2026-03-01 at 16 58 28@2x

Screenshot 1: Diagnostics show up since there is an error
Screenshot 2: Diagnostics error is solved by agent
Screenshot 3: Diagnistics does not show up anymore.

- [x] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) // re-using zeds component

Release Notes:

- Added `@diagnostics` to acp agent chat.

### DIFF
--- a/crates/agent_ui/src/message_editor.rs
+++ b/crates/agent_ui/src/message_editor.rs
@@ -70,16 +70,16 @@ impl PromptCompletionProviderDelegate for Entity<MessageEditor> {
     }
 
     fn supported_modes(&self, cx: &App) -> Vec<PromptContextType> {
-        let mut supported = vec![PromptContextType::File, PromptContextType::Symbol];
+        let mut supported = vec![
+            PromptContextType::File,
+            PromptContextType::Symbol,
+            PromptContextType::Diagnostics,
+        ];
         if self.read(cx).prompt_capabilities.borrow().embedded_context {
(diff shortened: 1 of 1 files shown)


raphaelluethy:
> Hey, thanks for this PR! I'm a bit confused, though: isn't this already possible? Here's a screenshot from `main` of Claude Code through ACP where diagnostics is already available as a mentionable context:
> 
> Screenshot 2026-03-02 at 9  34@2x
> Do we need these changes?

Hey, thank you for having a quick look...

...and sorry, I messed up big time. 
I had an installation of stable that was not up to date. 
I should have used the main branch before my change to test the `before` vs `after` instead of my installation. 

Sorry about that, also for wasting your time, wont happen again :(

PR #50457, opened 2026-03-02
editor: Make "diff clipboard with selection" do exact diff

### LINKED ISSUE #49204: "diff clipboard with selection" command adds newline to clipboard content
### Reproduction steps

1. Type something into a buffer, with no trailing newline
2. Copy the buffer contents
3. Run the "editor: diff clipboard with selection" command
4. Add a trailing newline to the buffer but don't re-copy
5. Run the "editor: diff clipboard with selection" command again

When the clipboard content does not have a trailing newline, Zed adds one when performing the diff.

If your buffer does not have a trailing newline, the diff will never be clean (when comparing the entire buffer).

OS: macOS 15.7.4
Memory: 16 GiB
Architecture: x86_64

### LINKED ISSUE #51812: Word diff is not displayed on the last line if the file lacks a trailing newline
### Reproduction steps

1. Open a project managed by Git in Zed
2. Open a file that does not have a trailing newline at the end
3. Edit the very last line of this file
4. Observe the diff: while the gutter shows the line is modified, the word-level diff (inline highlighting of the specific change) is missing
5. Add a trailing newline to the file and repeat step 3. The word diff now appears correctly

Here are two screen recordings for comparison:
with trailing newline:

https://github.com/user-attachments/assets/fff4817a-9cea-4b60-bdfe-ede2672797d8

without trailing newline:

https://github.com/user-attachments/assets/8ef701e1-a46e-47f4-b4a3-77a10daa208d

Current: Word diff (inline highlighting) fails to render on the last line if there is no trailing newline.

Expected: Word diff should be displayed consistently for all lines, regardless of whether the file ends with a newline.

Zed: v0.228.0+preview.201.7d41a5350a09e4e0ef25194b99e5a3f279bcc230 (Zed Preview) 
OS: Windows 10.0.26200
Memory: 47.8 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 770 || Intel Corporation || 32.0.101.6129

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

lingyaochu:
Closes #49204 
Closes #51812

As analyzed in https://github.com/zed-industries/zed/issues/49204#issuecomment-3905303481, the issue stems from an inconsistency: while selection extension (from #34999) fails at EOF, the logic still unconditionally appends a `\n` to the clipboard. This PR ensures the clipboard newline is only added if the selection was actually extended.

- [x] Test added
- [x] No UI change

Release Notes:

- Improved `editor: diff clipboard with selection` to make it do exact diff.
- Fixed the word diff highlight don't show on the last line if the file lacks a trailing newline.


crates/git_ui/src/text_diff_view.rs
```diff
@@ -88,7 +88,11 @@ impl TextDiffView {
         let source_buffer_snapshot = source_buffer.read(cx).snapshot();
         let mut clipboard_text = diff_data.clipboard_text.clone();
 
-        if !clipboard_text.ends_with("\n") {
+        let range_end_offset = source_buffer_snapshot.point_to_offset(expanded_selection_range.end);
+        let source_text_ends_with_newline = range_end_offset > 0
+            && source_buffer_snapshot.chars_at(range_end_offset - 1).next() == Some('\n');
```
SomeoneToIgnore: > range_end_offset - 1

This will panic if you try to test with the buffer with a single `√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√√` line without newlines.

We'd better reflect this in the tests.

SomeoneToIgnore (changes requested):
The fix is a bit confusing to me: if I understand the root of the issue correctly, we always add a trailing newline to the text copied off selection, when comparing?

I see that we do insert the copied text just fine, without the newline though, so the issue seems to be around the comparison layer somewhere.

In the issue's description, we compare that clipboard against 2 different texts: one without the trailing newline (1), and one with (2).

With the selection that has no trailing newline, I would expect the (1) case won't have a diff and (2) will.
Current `main` does the reverse which is a bug.

Yet, the new code seems to show now that both (1) and (2) have no diff???
This is not right at all?

crates/git_ui/src/text_diff_view.rs
lingyaochu: Oh, utf-8, my Python brain always falls into this trap.

lingyaochu:
> In the issue's description, we compare that clipboard against 2 different texts: one without the trailing newline (1), and one with (2).
> 
> With the selection that has no trailing newline, I would expect the (1) case won't have a diff and (2) will. Current `main` does the reverse which is a bug.
> 
> Yet, the new code seems to show now that both (1) and (2) have no diff??? This is not right at all?

I think I see where our perspectives diverge.

### The context:
Before this PR, Zed’s existing logic (#34999) already implements "selection extension." In any position except the very end of a file, selecting to the end of line1 (1) or the start of line2 (2) results in the same expanded selection. To match this, Zed also appends a `\n` to the text from clipboard.

My understanding was that this was an intentional design to prioritize "valuable diffs" over invisible trailing characters, ensuring (1) and (2) always behave identically.

### The issue(#49204):
The bug reported in #49204 occurs because this extension logic fails at the EOF (when there's no trailing newline to extend into). In this edge case, the buffer selection cannot be extended, but the clipboard is still padded with `\n`, creating an inconsistent diff.

### The Decision:
My PR was designed to fix this "EOF inconsistency" and maintain the existing behavior where (1) and (2) are treated as equivalent.

---

Based on your comment, it sounds like you prefer a stricter approach where (1) and (2) should be distinguished (I guess the git diff in Zed's other place also behave like this? So this is consistent). If that's the case, we are actually looking at a fundamental change to the "selection extension" logic for this action, rather than just fixing an edge case.

While I personally prefer this approach for the UX it provides, I agree with your reasoning from a design standpoint. I'm honestly a bit torn between the two—prioritizing a "clean" diff by ignoring invisible characters versus maintaining strict correctness. I'm happy to update the logic to distinguish between (1) and (2); I’ll just need to dive back into the extension code to figure out the cleanest way to do it (and hopefully not get lost in there for too long.).

I’d like your final word on which philosophy fits Zed's goals better before I proceed.

SomeoneToIgnore (commented):
Thank you, this works quite nice when testing the original issue.

I cannot comment well the changes in the `git_ui`, hence asking @cole-miller to chime in when he has some time.

lingyaochu:
I've lost some context due to the long delay, and the merge conflicts are quite large now, so I am closing this PR.

Issue #46673, opened 2026-01-13
Space button doesn't work on non-English(US) keyboard layout


kolakdd:
### Reproduction steps

1. Start Zed
2. Switch keyboard layout on non-English(US) (like German or French) 
3. Try typing space with space bar

### Current vs. Expected behavior

After updating between 12/28/2025 and 12/01/2026, the spacebar stopped working on non-English(US) keyboard layouts. Switching to English(US) works fine.

The same issue occurs when switching to any other keyboard layout besides English(US) (German, Arabic, English(Australia)).

### Zed version and system specs

Linux Mint 22.1 Cinnamon v6.4.8

Zed 0.218.7 
41bab92dac34777ae4256e5041f7b7236fd16cdd

0.218.7+stable.118.41bab92dac34777ae4256e5041f7b7236fd16cdd

teryble09:
having issues too, linux mint cinnamon

DoctorCat92:
The spacebar doesn't work in Russian. Ubuntu 24.04. Honestly, I don't really want to change the keyboard layout just for the sake of IDE. For now, I'll roll back to an older version of Zed. If anyone has found another solution, please let me know.

Issue #20465, opened 2024-11-09
Spacebar not recognized using Cinnamon desktop and Spanish keyboard layout


bibanez:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

When writing code I can insert a space character just fine when using the US keyboard layout. Nonetheless, when I type using the Spanish keyboard layout the spacebar doesn't register at all.

Steps to reproduce:
1. Use Cinnamon desktop (my distribution is Linux Mint and I installed Zed stable using the website's installation script)
2. Change keyboard layout to Spanish
3. Type spacebar into any text field (code editor, terminal, commands)

This only happens with this application.

### Environment

Zed: v0.160.7 (Zed)
OS: Linux X11 linuxmint 22
Memory: 15.5 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics (CML GT2) || Intel open-source Mesa driver || Mesa 24.0.9-0ubuntu0.2

### If applicable, attach your Zed.log file to this issue.

[Zed.log](https://github.com/user-attachments/files/17689417/Zed.log)

Oscarnordstrom:
I have the same issue. Freshly installed Zed and Linut mint 22, spacebar is not recognised in UK-ISO keyboard layout but works in swedish.

notpeter:
Can you open `debug open key context view` from the command palette and press `space`?
I'm curious what key shows up under Last Keystroke (mine shows "Typed: "space"`)

beniaminzagan:
@He-Is-HaZaRdOuS @Oscarnordstrom @bibanez any if you that can reproduce this with notpeter his suggestion? And let us know what shows up there? Appreciate it! 

> Can you open `debug open key context view` from the command palette and press `space`? I'm curious what key shows up under Last Keystroke (mine shows "Typed: "space"`)

He-Is-HaZaRdOuS:
> [@He-Is-HaZaRdOuS](https://github.com/He-Is-HaZaRdOuS) [@Oscarnordstrom](https://github.com/Oscarnordstrom) [@bibanez](https://github.com/bibanez) any if you that can reproduce this with notpeter his suggestion? And let us know what shows up there? Appreciate it!
> 
> > Can you open `debug open key context view` from the command palette and press `space`? I'm curious what key shows up under Last Keystroke (mine shows "Typed: "space"`)

I did as notpeter suggested and the spacebar just isn't registering as a keystroke at all. 
[...]

PR #50516, opened 2026-03-02
 linux: Fix space key ignored on X11 due to update_key/update_mask conflict

Dnreikronos:
Closes #49329

## Summary

On X11, the space key (and potentially other keys) could be silently ignored because `update_key()` was called after `from_xkb()` resolved the keysym. This created a conflict with
`update_mask()` — when both were used, `update_key()` could advance the XKB state in a way that made subsequent `update_mask()` calls produce stale or incorrect modifier state, causing keys
like space to resolve to `NoSymbol`.

The fix:
- **Removes `update_key()` calls** from both KeyPress and KeyRelease handlers. The X11 event's own `state` field (via `update_mask()`) is the authoritative source for modifier state and is
sufficient.
- **Fixes `update_xkb_mask_from_event_state()`** to correctly mask off mouse button bits and preserve depressed/latched layout groups, avoiding breakage with multi-layout keyboard
configurations.

This also fixes the related issue where non-US keyboard layouts (Spanish, German, etc.) had broken spacebar input on X11.

Also related to #20465 and #46673.

## Test plan

- [x] Added XKB unit tests verifying space, letters, Shift+bracket, and sequential key presses all resolve correctly using only `update_mask` (no `update_key`)
- [x] Manual testing on X11 with US keyboard layout — space key works
- [x] Manual testing on X11 with non-US keyboard layout — space key works

Release Notes:

- Fixed space key (and other keys) being ignored on X11, particularly with non-US keyboard layouts (#49329).

### DIFF
--- a/crates/gpui_linux/src/linux/platform.rs
+++ b/crates/gpui_linux/src/linux/platform.rs
@@ -1060,4 +1060,116 @@ mod tests {
             Point::new(px(5.0), px(5.1))
         ),);
     }
+
+    #[cfg(any(feature = "wayland", feature = "x11"))]
+    mod xkb_tests {
+        use crate::{Keystroke, Modifiers};
+        use xkbcommon::xkb;
+
+        fn create_us_xkb_state() -> xkb::State {
+            let context = xkb::Context::new(xkb::CONTEXT_NO_FLAGS);
+            let keymap = xkb::Keymap::new_from_names(
+                &context,
+                &"evdev",
+                &"pc105",
+                &"us",
+                &"",
+                None,
+                xkb::KEYMAP_COMPILE_NO_FLAGS,
+            )
+            .expect("Failed to create US keymap");
+            xkb::State::new(&keymap)
+        }
+
+        // Evdev keycodes + 8 = XKB keycodes
+        const KEYCODE_SPACE: u32 = 65;
+        const KEYCODE_A: u32 = 38;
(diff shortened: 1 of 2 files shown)


smitbarmase:
Thanks for digging into this. I’m going to close this in favor of a follow-up, but your PR helped narrow the issue down.

The main part I agree with is removing the remaining `update_key()` calls. For X11, the libxkbcommon docs are explicit (https://xkbcommon.org/doc/current/group__x11.html):

> However, XkbStateNotify is more accurate. There is no need to call xkb_state_update_key(); the state is already synchronized.

The part I’m not taking is continuing to mutate the shared `state.xkb` from KeyPress / KeyRelease state. The `state.xkb` is our client-side mirror of the server’s keyboard state, so I want that long-lived state to stay driven by XkbStateNotify and keymap notifications only.
  
The `update_mask()` docs also warn:

> Warning: The serialization is lossy and will not survive round trips; it must only be used to feed client state objects, and must not be used to update the server state.

That is why I don’t want to keep rewriting the shared `state.xkb` from per-event KeyPress / KeyRelease state. Instead, I’m keeping `state.xkb` notify-driven and using a temporary event-local XKB state for per-event lookup.

Thanks again for working through this.

smitbarmase:
Follow up PR: https://github.com/zed-industries/zed/pull/53903

PR #50523, opened 2026-03-02
Fix title/camelCase commands stripping leading indentation  Fixes

OmChillure:
Fixes: #48945

Description:

The convert:to-title-case, convert:to-upper-camel-case, and convert:to-lower-camel-case editor commands were stripping leading whitespace from each line of a multi-line selection.

Root cause: The conversion functions split on whitespace using .split_whitespace() and then joined the resulting words, discarding any leading spaces/tabs before the first word on each line.

Fix: Each line now preserves its leading whitespace by capturing and re-prepending it before applying the case conversion.

Tests: Added test cases covering multi-line selections with indentation for all three commands.

Video : 

[bug1fix.webm](https://github.com/user-attachments/assets/f4d25c55-bc6d-44e6-a989-7d9b4bc59ac9)

Release Notes:

- Fixed trailing whitespace handling on text case changes

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -12549,7 +12549,9 @@ impl Editor {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
-        self.manipulate_text(window, cx, |text| Self::convert_text_case(text, Case::Title))
+        self.manipulate_text(window, cx, |text| {
+            Self::convert_text_case(text, Case::Title)
+        })
     }
 
     pub fn convert_to_snake_case(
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (commented):
Nice catch, but the fix seems somewhat odd given it's the library that removes the whitespaces, should we fix the rest of the invocations? 
There are 6 total in this file but only 3 are being fixed and tested here which seems odd — seems that all of them would benefit from a common method that fixes things via `trim_start`.

Should we also use `trim_end` for similar reasons?

crates/editor/src/editor.rs
```diff
@@ -12668,6 +12644,18 @@ impl Editor {
         })
     }
 
+    fn convert_text_case(text: &str, case: Case) -> String {
+        text.split('\n')
```
```suggestion
        text.lines()
```

Sorry, missed that, but we do support Windows hence have to do this too.

SomeoneToIgnore (approved):
Thank you!

crates/editor/src/editor.rs
OmChillure: done

PR #50528, opened 2026-03-02
wGPU: Select more specifically

ConradIrwin:
This uses the compositor hints if available to pick the best GPU. If none is
available, it tries each GPU in turn, and the first that actually works is chosen

Release Notes:

- Linux: Select a more appropriate GPU

### DIFF
--- a/crates/zed/src/main.rs
+++ b/crates/zed/src/main.rs
@@ -276,7 +276,7 @@ fn main() {
 
     zlog::init();
 
-    if stdout_is_a_pty() {
+    if true {
         zlog::init_output_stdout();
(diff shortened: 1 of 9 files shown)


crates/zed/src/main.rs
```diff
@@ -276,7 +276,7 @@ fn main() {
 
     zlog::init();
 
-    if stdout_is_a_pty() {
+    if true {
```
Fingel, line 279: Debugging left in?

ConradIrwin: Fixed in #50543

PR #50532, opened 2026-03-02
Fix incorrect shell level on reload

### LINKED ISSUE #33958: $SHLVL starts at 2 and increases by 1 every time Zed is reloaded

prayanshchh:
Closes #33958

Release Notes:


prayanshchh:
so I have two options which we could take

1. run zed as detached helper - we spawn the zed binary as detached helper we can adjust the main function in `src/main.rs` to recognise this restart (we can do this as we will start this binary with some restart args like the pid of the process we have to wait to finish to start zed again and the path of the zed binary), this newly started zed binary will only poll and check if the process with pid exists (the old zed instance) if it doesn't it opens zed again

2. create a dedicated binary that does exactly 1 this way restart helper won't need to be as heavy as zed.

none of the above I feel is the way to go, I have a feeling they might be too heavy for the problem we are trying to solve...bash is the lightest of them all.

if we really need to solve the issue we can just do `.env_remove("SHLVL")` when we spawn the bash.

do u have any ideas?

prayanshchh:
@tomhoule any ideas?

tomhoule:
Sorry for the delay. We could explore option 1. from your message, I think. Introducing an extra binary would impact the installation process, CI and probably other places.

prayanshchh:
@codex

PR #50538, opened 2026-03-02
workspace: Fix scrolling to active tab when pinned tabs are present

zharinov:
When pinned tabs are present, activating an unpinned tab passed the absolute tab index to the scroll handle, which only contains unpinned tabs. This caused the scroll-into-view to silently fail.

Subtract `pinned_tab_count` from the index so it maps to the correct child in the unpinned tabs scroll container.

Release Notes:

- Fixed tab bar not reliably scrolling to the active tab when pinned tabs are present.

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -7996,7 +7996,7 @@ mod tests {
         let scroll_bounds = scroll_handle.bounds();
 
         assert!(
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Nice find and good fix, very much appreciate you adding a test here. Thank you very much!

PR #50540, opened 2026-03-02
gpui: Hide XF86 keybindings from menus and keybinding hints

Dnreikronos:
Closes #50436

Release Notes:

- Fixed XF86 multimedia key names ("New", "Save", "Open") being shown as keybinding hints in menus instead of the actual keyboard shortcuts.

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -224,8 +224,8 @@
     "context": "ContextEditor > Editor",
     "bindings": {
       "ctrl-enter": "assistant::Assist",
-      "ctrl-s": "workspace::Save",
       "save": "workspace::Save",
+      "ctrl-s": "workspace::Save",
       "ctrl-<": "assistant::InsertIntoEditor",
       "shift-enter": "assistant::Split",
       "ctrl-r": "assistant::CycleMessageRole",
--- a/crates/gpui/src/key_dispatch.rs
+++ b/crates/gpui/src/key_dispatch.rs
@@ -427,7 +427,12 @@ impl DispatchTree {
             .bindings_for_action(action)
             .rev()
             .find(|binding| {
-                Self::binding_matches_predicate_and_not_shadowed(&keymap, binding, context_stack)
+                !binding.has_xf86_keystroke()
+                    && Self::binding_matches_predicate_and_not_shadowed(
+                        &keymap,
+                        binding,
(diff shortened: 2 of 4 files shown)


Dnreikronos:
### Summary

XF86 multimedia keysyms (XF86New, XF86Save, XF86Open, etc.) are mapped to key names like "New", "Save", "Open" on the Linux platform layer. When Zed
picks the highest-precedence binding to display in menus and keybinding hints, it can select these XF86 bindings over useful ones like Ctrl+N, resulting in confusing labels  "New" as a shortcut looks like the UI is broken, not like a keybinding.

### What changed

- Added is_xf86_key() in keystroke.rs to identify XF86 multimedia key names (new, open, save, cut, copy, paste).
- Added has_xf86_keystroke() on KeyBinding to check if any keystroke in a binding is an XF86 key.
- Filtered XF86 bindings out of highest_precedence_binding_for_action in key_dispatch.rs — this is the method used by menus and keybinding hints to decide what to display.
- Added two unit tests covering both the filtering and dispatch behavior.

### Impact

- Menus and keybinding hints no longer show XF86 key names. They fall through to the next non-XF86 binding (e.g. Ctrl+N) or show nothing if no
alternative exists.
- Key dispatch is untouched — if a user actually presses an XF86 key, it still triggers the bound action.
- bindings_for_action (the full list, used by the keymap editor) still includes XF86 bindings.

### Trade-offs

- Users with rare XF86 keyboards will no longer see those bindings in menus, so they can't discover them through the UI. This is acceptable because:
(1) these keyboards are essentially extinct, (2) the keys still work when pressed, and (3) the bindings are still visible in the keymap settings.
- The is_xf86_key check uses a hardcoded list of key names. If new XF86 keysyms are added to the platform layer in the future, the list would need to
be updated. An alternative would be tagging keystrokes as XF86 at the platform layer, but that's a larger refactor for minimal gain.

tredondo:
OpenRecent was fixed, but I still see these weird shortcuts in 1.0.0+stable.257.5ec84a926ef83865afb92d2a3d1ca3b419572cf9

image

PS: I'm curious how these keybindings even ended up in the source code, given:

> Users with rare XF86 keyboards will no longer see those bindings in menus, so they can't discover them through the UI. This is acceptable because: **(1) these keyboards are essentially extinct**

PR #50545, opened 2026-03-03
agent_ui: Allow selection of commands from tool calls

### LINKED ISSUE #50427: Can't copy text from shell commands in the Agent Panel
### Reproduction steps

1. In the Agent Panel, run a prompt that outputs a shell command, for example `Obtain this Linux machine's OS version`
2. Try to select some of the text of that shell command

Notice the cursor is an arrow rather than the text selection.

Image

 OpenCode had [the same bug](https://github.com/anomalyco/opencode/issues/15358).

Zed: v0.225.10+stable.184.7c950f7b7c11ed485b0e7d72216ca36ebc86b58b (Zed) 
OS: Linux Wayland tuxedo 24.04
Memory: 62.5 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3050 Ti Laptop GPU || NVIDIA || 580.126.09

yeskunall:
Closes #50427.

The changes respect your current theme + mimic the existing UI when it comes to font size+weight.

image

https://github.com/user-attachments/assets/a5fb6c82-fffd-494f-a374-9296d1690736

Release Notes:

- Allow selection of commands from tool calls

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -297,6 +297,15 @@ impl MarkdownStyle {
         }
     }
 
+    pub fn with_buffer_font(mut self, cx: &App) -> Self {
+        let theme_settings = ThemeSettings::get_global(cx);
+        self.base_text_style.font_family = theme_settings.buffer_font.family.clone();
+        self.base_text_style.font_fallbacks = theme_settings.buffer_font.fallbacks.clone();
+        self.base_text_style.font_features = theme_settings.buffer_font.features.clone();
+        self.base_text_style.font_weight = theme_settings.buffer_font.weight;
+        self
+    }
+
     pub fn with_muted_text(mut self, cx: &App) -> Self {
         let colors = cx.theme().colors();
         self.base_text_style.color = colors.text_muted;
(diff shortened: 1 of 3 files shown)


crates/markdown/src/markdown.rs
```diff
@@ -226,6 +226,15 @@ impl MarkdownStyle {
         }
     }
 
+    pub fn with_buffer_font(mut self, cx: &App) -> Self {
```
yeskunall, line 300: Didn’t notice an existing helper for this combo and also didn’t like writing isolated styles like that. The `MarkdownStyle::themed()` always uses the UI font for base text and the buffer font settings in `themed()` are only applied to the `code_block` style, not to paragraph text.

The following is from Opus 4.6:

The manual code is needed because of a GPUI quirk: font family/weight/features propagate through `TextRun`s (so they go on `base_text_style`), but font size and line height only cascade through the div tree (so they must go on `container_style`). That's why the overrides are split across two places.

tomhoule:
I rebased this and did a round of manual testing. As far as I can tell it's working great!

tredondo:
Thanks @tomhoule ! Did this make it into v0.233.10+stable.250.350f33871e84d5ed83a403523e9490b9bf014a94? I still can't select text from agent tool call blocks.

MrSubidubi:
@zed-industries/approved 

@tredondo It did not make the stable cut

PR #50554, opened 2026-03-03
debugger: Reverse Python repr escaping

ConradIrwin:
Closes #37168

Authored-By: @ngauder

Release Notes:

- debugger: Unescape Python strings


Anthony-Eid:
The fix makes sense, the only thing I'm not a fan of is that it's adapter specific. There doesn't seem to be a clean way around that though. 

I wonder if there's an arg we can pass into Debugpy to fix this, or if we should submit a PR.

PR #50593, opened 2026-03-03
Fix external file drops in terminal tabs outside terminal panel

### LINKED ISSUE #49800: Drag & drop file in terminal to print path partially works

mangeshraut712:
Fixes #49800

## Problem
Dropping files from the OS into a terminal tab only pasted paths when that terminal lived inside the terminal panel. If the same terminal tab was moved into a regular workspace pane, pane-level drop handling opened files/directories instead of pasting paths into the terminal.

## Repro
1. Open a terminal tab from the terminal panel.
2. Move/split that terminal tab into a regular pane.
3. Drag a file from Finder/Explorer into that terminal tab.
4. Observe file-open behavior instead of path paste.

## Fix
- Added an item-level external-path drop hook on `workspace::Item` and `ItemHandle`.
- Updated generic pane external-path drop handling to first let the active item consume the drop.
- Implemented this hook in `TerminalView` to paste quoted paths into the terminal input (local projects only).
- Updated terminal panel path-drop helper to use the same item-level hook so terminal drop behavior is consistent in both panel and regular panes.

## Tests
- Added `test_external_paths_drop_can_be_handled_by_active_item` in `crates/workspace/src/pane.rs`.
- Ran: `cargo check -p workspace --lib`
- Ran: `cargo test -p terminal_view test_custom_title_initially_none -- --nocapture`
- Note: `cargo test -p workspace ...` currently fails on an unrelated pre-existing compile error in `crates/workspace/src/persistence.rs` (`RemoteConnectionOptions::Mock(_)` match arm missing).

## Risk
Low. The pane behavior change only applies to external-path drops and only when the active item explicitly opts in by returning `true` from the new hook. Existing items keep previous behavior via default `false`.

Release Notes:

- Fixed dropping external files into terminal tabs so paths are pasted even when the terminal tab is outside the terminal panel.

### DIFF
--- a/crates/terminal_view/src/terminal_panel.rs
+++ b/crates/terminal_view/src/terminal_panel.rs
@@ -1382,18 +1382,8 @@ fn add_paths_to_terminal(
     window: &mut Window,
     cx: &mut Context<Pane>,
 ) {
-    if let Some(terminal_view) = pane
-        .active_item()
-        .and_then(|item| item.downcast::<TerminalView>())
-    {
-        window.focus(&terminal_view.focus_handle(cx), cx);
-        let mut new_text = paths.iter().map(|path| format!(" {path:?}")).join("");
-        new_text.push(' ');
-        terminal_view.update(cx, |terminal_view, cx| {
-            terminal_view.terminal().update(cx, |terminal, _| {
-                terminal.paste(&new_text);
-            });
-        });
+    if let Some(active_item) = pane.active_item() {
+        active_item.handle_external_paths_drop(paths, window, cx);
     }
 }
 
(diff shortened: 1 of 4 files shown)


mangeshraut712:
@cla-bot check

mangeshraut712:
@cla-bot check

PR #38757, opened 2025-09-24
When searching outline, select nearest match to cursor if multiple are found

bprize15:
Closes https://github.com/zed-industries/zed/issues/29774

**Background**
There was a bug when multiple matches were found when searching an outline. If multiple matches were found, it would simply compare by match score, which is calculated using longest subsequence. If there were multiple matches with the same score, it would always return the last match (see [max_by_key docs)](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by_key)).

**Solution**
Now, the match closest to the cursor when the search takes place is selected. This is based on a distance calculation.

Release Notes:
- Fixed bug when searching outline, if multiple matches were found with the same match score, it would always select the last match. Now it selects the match closest to the cursor in this when multiple matches with the same match score are found.

PR #50594, opened 2026-03-03
Fix outline filtering always selecting last match

### LINKED ISSUE #29774: Outline filtering always select the last match at first
### Summary

No matter where the cursor is, filtering in outline always select the last result first.
### Description

Steps to reproduce:
1. Paste the code in a Python buffer
```python
class A:
    def f(self):
        pass
        
    def g(self):
        pass
        
        
class B:
    def f(self):
        pass
        
    def g(self):
        pass
        
        
class C:
    def f(self):
        pass
        
    def g(self):
        pass
```
2. Open outline.
3. Input `f`.
4. Move the cursor around
5. Open outline.
6. Input `f`.

Expected Behavior:

`C::f` should not be always selected by default. Maybe the first match after cursor should be selected by default.

Actual Behavior:

`C::f` is always selected by default. Even the cursor is at the beginning of the buffer.

Image

Zed: v0.185.2 (Zed Preview) 
OS: macOS 15.4.1
Memory: 64 GiB
Architecture: aarch64

OmChillure:
Fixes #29774

When filtering the outline panel, matches with equal fuzzy scores previously defaulted to the last item due to iterator `max_by_key` semantics. This caused the bottommost match (e.g. `C::f`) to always be pre-selected regardless of cursor position.

Changes:
- Select the match nearest to the cursor when scores are tied, using a multi-criteria comparison: score -> cursor containment depth -> proximity to cursor -> earlier index
- Move outline search off the UI thread (`smol::block_on` -> async `cx.spawn_in`) to avoid blocking during filtering
- Wrap `Outline<Anchor>` in `Arc` for cheap cloning into the async task
- Add `match_update_count` to discard results from stale queries

Tests : 
Adds a regression test: `test_outline_filtered_selection_prefers_cursor_proximity_over_last_tie`  which passes

Video : 
[Screencast from 2026-03-03 17-01-32.webm](https://github.com/user-attachments/assets/7a27eaed-82a0-4990-85af-08c5a781f269)

Release Notes:
Fixed the outline filtering always select last match


SomeoneToIgnore (commented):
I need time to do a context switch and focus, that's mainly it — pinging does not help much with this, if not hurts.
My plans to do this next week, maybe on Monday even.

Without understanding the context much: `match_update_count` usage seems peculiar as it's not clear why this is needed; re-sorting all outline items each time seems odd too — imagine we deal with `editor.rs`'s outlines, and you do that on sorting on the main thread.

crates/outline/src/outline.rs
```diff
@@ -322,25 +326,85 @@ impl PickerDelegate for OutlineViewDelegate {
...
-        self.set_selected_index(selected_index, !self.last_query.is_empty(), cx);
-        Task::ready(())
+
+        let outline = self.outline.clone();
+        cx.spawn_in(window, async move |this, cx| {
+            let matches = outline
+                .search(&query, cx.background_executor().clone())
+                .await;
+            let _ = this.update(cx, |this, cx| {
+                if this.delegate.match_update_count != match_update_count {
```
SomeoneToIgnore: This seems to be the only place that handles with the delegate's method and the related task: https://github.com/zed-industries/zed/blob/1fd5cbafe66cb575ebe4789d69bda5fe1304d215/crates/picker/src/picker.rs#L686-L713

The previous one gets dropped after replacement, stopping its tasks' tree execution, so do we really need to preserve like that?

crates/outline/src/outline.rs
```diff
@@ -322,25 +326,85 @@ impl PickerDelegate for OutlineViewDelegate {
...
+                                        );
+                                        distance_2.cmp(&distance_1)
+                                    })
+                                    .then_with(|| ix_2.cmp(ix_1))
+                            })
+                    })
+                    .map(|(ix, _)| ix)
+                    .unwrap_or(0);
+
+                this.delegate.last_query = query.clone();
```
SomeoneToIgnore:
```suggestion
                this.delegate.last_query = query;
```

crates/outline/src/outline.rs
```diff
@@ -322,25 +326,85 @@ impl PickerDelegate for OutlineViewDelegate {
...
+                            .head();
+                        (buffer, cursor_offset)
+                    });
+
+                let selected_index = this
+                    .delegate
+                    .matches
+                    .iter()
+                    .enumerate()
+                    .max_by(|(ix_1, m_1), (ix_2, m_2)| {
```
SomeoneToIgnore: What's confusing me here are 2 different ways to compare the same data?

Above, for the empty query, `outline.items` is used and ordered via one code.
Here, something that looks quite different despite `candidate_id` should've been enough to allow reusing the same approach to sorting?

Could we reorder the code to reuse the same sorting approach to both logic branches?

SomeoneToIgnore (commented):
* Thank you for the fixes to `block_on` and related.
I think we'd better rework the code so that all outline item (after matching the fuzzy match or not) are run as `cx.executor().background_spawn(async move {... })` task, or at least on the task on the main thread, but asynchronously still, not like what the empty query code path does — this way we'll improve the rendering-related code and stop risking to block it for too long.

* I have been testing in https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide/src/inlay_hints.rs by placing a caret to the beginning and the end of the file, and have noticed that the implementation selects something that's not on the edge for both cases (as I'd expected based on the code changes)

https://github.com/user-attachments/assets/2b7f7778-fc82-4fee-a0b0-8b4c0ae1855f

Is it something expected or should we have more tests and fixes to the comparison code?

crates/outline/src/outline.rs
OmChillure: I tested the cases shown in the video and the current behavior is intentional.

The ranking prioritizes:

1. Match score from the fuzzy matcher
2. Cursor containment (symbols containing the cursor)
3. Proximity to the cursor
4. Earlier index as a tiebreaker

Because match score is evaluated first, a symbol with a significantly better fuzzy match can be selected even if it is farther from the cursor. Cursor proximity mainly affects ordering when the match scores are similar.

If you'd prefer cursor proximity to have stronger priority than match score, I can adjust the comparison logic or add additional tests.

SomeoneToIgnore: ~To note, the original comment in this thread is about removing the `match_update_count` field entirely.~

------------

Ideally, I prefer an explanation why this works as is, and comparison with other editors and IDEs.
We have a good chance to make things ergonomic and predictable from the first try, so seems worth investing the time.


The PR does not justify the choice much, and https://github.com/zed-industries/zed/issues/29774 does not specify this particular approach, rather suggesting selecting the first match?

So, whatever happens here is relatively arbitrary and might be not intuitive for certain people?

For comparison, Helix has a symbol picker triggered with `space-s` and that always selects the first item, which seems to be just fine?

image

As a bonus, such approach requires no extra sorting.

OmChillure: Actually the bug is in the empty-query path specifically, all items score 0.0, so without tiebreaking, `max_by` falls back to iterator order and picks the last item. Selecting the first item would just move the problem (last -> top of file), not fix it. Pre-selecting the symbol nearest to/containing the cursor is the meaningful fix, and matches how VSCode's outline panel behaves.

The issue reporter themselves suggested "the first match after cursor should be selected by default" cursor-proximity is exactly that. There was also a previous attempt at this in #38757 with the same approach ("select nearest match to cursor"), so this seems to be the intuitive fix that multiple people independently arrived at. It also matches how VSCode's outline panel behaves.

Happy to simplify the filtered (non-empty query) path to just `matches[0]` if you prefer, proximity that I added was just consistency, not the core fix.

SomeoneToIgnore: > It also matches how VSCode's outline panel behaves.

It does not?
Or do I miss something?


https://github.com/user-attachments/assets/adff6101-fe4e-4fa6-8da5-18f2d65115f7

OmChillure: Ahh I overstated the VSCode comparison apologize for that.

For the filtered-query path, I can simplify to just matches[0] since outline.search() already returns results sorted by score no extra cursor-proximity logic needed there.

For the empty-query path, all scores are 0.0 so matches[0] would always be the top-of-file symbol regardless of cursor position which is what the original issue complained about (just in reverse). Cursor-proximity selection makes sense there since there's no score signal to rely on.

Happy to go with whichever approach you prefer - I can either keep proximity only for the empty-query case, or simplify both paths to matches[0] which seems to be cleaner and better.

SomeoneToIgnore: My preference stays the same, as mentioned above:

image

I would love to see a few short videos similar to mine, comparing different editors (so, combined with mine, we get 4 or more editors/IDEs to reflect on) — the whole point of contributions is helping this research and context gathering, is not it?
LLMs can deal with the rest, including overstatements and writing the code.

Based on what we see in those 4 or more editors', we can decide how to proceed.

OmChillure: Sure will get you that !!

SomeoneToIgnore: Great, thank you for the videos found.
Now, we need to check the behavior in each video to see what editor does when

* the outline/symbol/etc. panel is just open
* the filtering had been applied on top of it

cases happen in different places in the file, does it change anything or not?

From what I see in the videos (correct me here if I'm wrong): 

* all editors but Helix try to open the latest selection's item, if we are in its range.
No proximity or anything like that is considered, just the direct "selection is within the symbol's range" match that alters the selected item — otherwise all of them ball back to Helix's behavior to select the first item

And from what I see in Zed right now, this is what is done already?
So, all that's needed from the PR here is to move the original code into a task?

* when filtering, many editors seem to always select the first item, even if on initial open some specific item was selected?

Intellij seems to do something better based on heuristics, but seems not too much worth it given the amount of code and extra filtering task needed?

So, seems that we need to select the first item here — what's odd is that any other modals in Zed (e.g. command one cmd/ctrl-shift-p) where the first match is selected always — so ideally we should find a bug in the outline's code and fix that, not just do `set_selected(0)`-like band-aid.

-----------

Please, also consider rechecking the scenarios for various caret placement and, If I'm wrong and some editors demonstrate a behavior worth implementing, let's do that instead.

If we have those 2 scenarios tested then, we're good to merge.

OmChillure: > * all editors but Helix try to open the latest selection's item, if we are in its range.
>   No proximity or anything like that is considered, just the direct "selection is within the symbol's range" match that alters the selected item — otherwise all of them ball back to Helix's behavior to select the first item

The current code already does cursor-in-range -> deepest containing symbol -> else first item. The proximity tiebreaker I added is unnecessary - other editors don't use it, and it overcomplicates things. I'll simplify to just: cursor-in-range check -> deepest containing symbol -> else fall back to index 0.

OmChillure: > * when filtering, many editors seem to always select the first item, even if on initial open some specific item was selected?
> 
> Intellij seems to do something better based on heuristics, but seems not too much worth it given the amount of code and extra filtering task needed?

the bug in my code is  `max_by_key(|(_, m)|` OrderedFloat(m.score)). Rust's `max_by_key` returns the last element on ties, not the first. That's why it always jumped to the bottommost match. The fix is changing the comparator to tiebreak by earlier index:


.max_by(|(ix_a, a), (ix_b, b)| {
    OrderedFloat(a.score)
        .cmp(&OrderedFloat(b.score))
        .then(ix_b.cmp(ix_a)) // earlier index wins on tie
})

No set_selected(0) band-aid, no cursor proximity during filtering — just fixing the actual tie-breaking bug. Consistent with Zed's command palette behavior and what other editors do.

OmChillure: So as fix will simplify the PR, remove proximity from filtering, and update the tests for both scenarios.

crates/outline/src/outline.rs
```diff
@@ -280,67 +275,73 @@ impl PickerDelegate for OutlineViewDelegate {
...
+                    .search(&query, cx.background_executor().clone())
+                    .await
+            };
+
+            let _ = this.update(cx, |this, cx| {
+                this.delegate.matches = matches;
+                let selected_index = if is_query_empty {
+                    let (buffer, cursor_offset) =
+                        this.delegate.active_editor.update(cx, |editor, cx| {
+                            let buffer = editor.buffer().read(cx).snapshot(cx);
```
SomeoneToIgnore: Note: taking a snapshot has some cost, and `display_snapshot` has `buffer()` to access the multi buffer one.
So, we can construct one less snapshot which would be good.

SomeoneToIgnore (approved):
Thank you for pushing it forward!

Seems that now we do not select the latest element that aggressively, yet something is a bit odd to me in how we do it that I fail to formulate — this is still quite an increment so let's see how more people find this.

crates/outline/src/outline.rs
OmChillure: Ohh yeah , thanks for the fix !

PR #50609, opened 2026-03-03
Fix showing unsupported window controls on Linux Wayland

Moulberry:
Release Notes:

- Fixed showing unsupported window controls on Linux Wayland

Before:
image

After:
image


Moulberry:
@probably-neb Done

PR #50636, opened 2026-03-03
agent: Auto-compact when reaching token limit

G36maid:
Closes #34515

### Context & Motivation
Users heavily relying on agentic editing or local LLMs (which often have smaller context windows) frequently hit the token limit. Currently, users must manually babysit the token bar and click "New from summary" to rotate the session. This interrupts the agentic workflow and breaks the development flow.

Inspired by the context management in [OpenCode](https://github.com/sst/opencode) and [OpenCode-ai](https://github.com/opencode-ai/opencode), this PR introduces an automated background context rotation mechanism.

### Implementation
Instead of mutating the current thread (which causes irreversible context loss of inline files and terminal outputs), this implementation reuses the existing `NewNativeAgentThreadFromSummary` action for a non-destructive rotation.

1. **Settings**: Added `auto_compact` (bool, default `false`) and `auto_compact_threshold` (float, default `0.95`) to `AgentSettings`.
2. **Event Interception**: Intercepted `AcpThreadEvent::TokenUsageUpdated` in the UI layer (`ConnectionView`).
3. **Execution**: Once the threshold is crossed, it programmatically dispatches the "New from summary" action. This seamlessly spawns a new session while leaving the old session safely archived in the history.

### Alternatives Considered
**In-place truncation:** Clearing the message history and replacing it with a summary block inside the same thread.
*Rejected because:* If the LLM generates a poor summary, the user permanently loses their file context and chat history. Spawning a new session is a non-destructive and much safer approach. *(Side note: I considered auto-triggering a `git commit` during auto-compact for extra safety, but kept it out of scope for this initial implementation to avoid feature creep.)*

### How to Test
1. In your `settings.json`, configure:
```json
   "agent": {
     "auto_compact": true, // default=false
     "auto_compact_threshold": 0.5 // default=0.95
   }
```
2. Open the Agent panel and start chatting/adding files.
3. Once the token usage exceeds 50%, observe that Zed automatically summarizes the thread and transitions you into a new session.
4. Verify the previous session is intact in the History menu.

### Open Questions

I'd love to discuss a few UX details:

1. **The Continue Prompt:** Right now, "New from summary" leaves the new thread waiting for user input. If the agent was mid-task, how should we prompt it to automatically continue working based on the new summary?
2. **The Trigger Rate:** Is `0.95` (95%) the right default threshold? I'm concerned that sudden, massive tool-call responses (e.g., large terminal outputs or MCP server responses) might overflow the context before the 95% threshold event is even caught.

Release Notes:

- Added an `auto_compact` setting to automatically summarize and rotate Agent sessions when token usage approaches the context window limit.


G36maid:
Just logging some thoughts on the next steps for this PR. 

The main piece missing right now is injecting a proper system prompt when triggering the compact action. I'm planning to use `opencode`'s implementation as a reference: [compaction.ts#L173](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/session/compaction.ts#L173).

Also considering how we should handle the prompt text: should we hardcode a default, or allow users to define a custom prompt in their settings (similar to how compaction behavior is configured in `opencode.json`)? 

I'll start experimenting with the logic, but any early feedback on the config approach is welcome!

osyvokon:
We're planning a new handoff system, so I'm going to close this PR. Thanks anyway!

PR #50639, opened 2026-03-03
editor: Add git diff-hunk signs to editor gutter

seanstrom:
Closes #46817

**Summary**:

* I've added a Zed setting to the Git Gutter Visibility dropdown to allow for choosing to display the git-gutter UI with git-diff characters.
* When the setting is enabled, the git-diff characters are displayed next to the line-numbers inside the editor gutter, and the characters are only displayed when the git-diff hunk is expanded.
  * Note, the editor gutter will display the plus `+` and minus `-` characters next to the line-number.
  * At the moment, the git-diff characters are rendered as text and colored based on the editor theme.

Release Notes:

- Add support for displaying Git diff-signs characters inside the Editor gutter when viewing an expanded diff-hunk.

Screenshots:

#### Expanded Diff Hunk
Screenshot 2026-03-03 at 13 26 56

#### Source Control View
Screenshot 2026-03-03 at 13 23 18

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1658,10 +1658,12 @@
     "enable_status": true,
     // Whether to enable git diff display.
     "enable_diff": true,
-    // Control whether the git gutter is shown. May take 2 values:
+    // Control whether the git gutter is shown. May take 3 values:
     // 1. Show the gutter
     //      "git_gutter": "tracked_files"
-    // 2. Hide the gutter
+    // 2. Show the gutter with per-line signs
+    //      "git_gutter": "tracked_files_with_signs"
+    // 3. Hide the gutter
     //      "git_gutter": "hide"
     "git_gutter": "tracked_files",
     /// Sets the debounce threshold (in milliseconds) after which changes are reflected in the git gutter.
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1666,6 +1666,8 @@
     // 3. Hide the gutter
     //      "git_gutter": "hide"
     "git_gutter": "tracked_files",
+    // Minimum contrast used when rendering diff hunk signs in the gutter.
+    "git_gutter_sign_minimum_contrast": 70.0,
     /// Sets the debounce threshold (in milliseconds) after which changes are reflected in the git gutter.
     ///
     /// Default: 0
--- a/crates/editor/src/config.rs
+++ b/crates/editor/src/config.rs
@@ -208,6 +208,15 @@ impl Editor {
         }
     }
 
+    pub fn set_show_git_diff_hunk_signs_gutter(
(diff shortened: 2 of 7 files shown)


Swivelgames:
@seanstrom Can we make sure the colors of the symbols `-`/`+` meet WCAG accessibility guidelines for contrast?

Or is it completely up to the Theme what color they end up being?

As of right now, it's very difficult to see the symbols in your screenshots :disappointed: 

:link: [WebAIM Contrast Checker](https://webaim.org/resources/contrastchecker/)

WebAIM Contrast Results Screenshot

GitLab solves this problem by providing the ability to override the theme colors and choose your own for additions/subtractions:

GitLab Diff Color Preferences Screenshot

seanstrom:
I've shared my change to increase the color contrast of the diff characters. I used the color contrast tool that you mentioned to help me adjust the code to choose a minimum contrast level for these changes.

I imagine we could also make the minimum contrast for these diff-characters a configurable setting too (seems like we already have some contrast settings for the terminal and highlights).

Here's some screenshots of the color tweaks:

## Light Theme (original example)

Screenshot 2026-03-04 at 17 29 06

## Dark Theme (for reference)
Screenshot 2026-03-04 at 17 31 27

seanstrom:
I noticed the git diff-hunk characters being affected by changes to the editor font-family setting, so I've updated the implementation to use SVG icons. Alternatively, we could also use the Zed UI font-family for displaying these git diff-hunk characters.

Here's some screenshots of the svg icon version:

##### Zed Mono Font
Screenshot 2026-03-07 at 19 56 39

##### Iosevka Font
Screenshot 2026-03-07 at 19 56 59

seanstrom:
Okay I've updated the PR based on some feedback during a Zed office hours.

Some of the feedback included:
* Re-positioning the diff-hunk signs in the visual lane as the "fold" buttons
* Also the color of the diff-hunk signs were adapted to the line-number color
* And there was a preference to use the line-number font for rendering the diff-hunk signs

One thing to note is that some of the themes use a very low contrast line-number color, which means the diff-hunk signs will not always be perfectly visible across themes (unless we mix colors to force the diff-hunk-signs to a have a minimum contrast).

I've made the layout and color changes, and I've left the code for integrating a minimum contrast to the diff-hunk signs (though I'm open to another solution).

cc @ConradIrwin @danilo-leal 

---

Here's some screenshots of the current implementation:

##### Lower Contrast Example

Screenshot 2026-06-11 at 15 20 52


##### Higher Contrast Example
Screenshot 2026-06-11 at 15 20 38

##### Aligning with fold buttons

Screenshot 2026-06-11 at 15 26 48


##### Diff View

Screenshot 2026-06-11 at 15 27 01

PR #50641, opened 2026-03-03
lsp: Add semantic token rules for Go

### LINKED ISSUE #50642: Go const variables not styled as constants (gopls readonly modifier)
## Summary

Go `const` declarations are not styled with the `constant` theme color in Zed. They render as regular variables instead.

## Expected behavior

In VS Code, Go `const` variables are styled with the `constant` color, distinct from regular variables.

## Actual behavior

In Zed, Go `const` variables render with the same color as regular variables.

## Root cause

`gopls` reports Go `const` declarations as semantic token type `variable` with the `readonly` modifier. There is no Go-specific semantic token rule to map `variable` + `readonly` to the `constant` theme style, so they fall back to plain `variable`.

Rust already has a language-specific `semantic_token_rules.json` (for rust-analyzer's custom tokens). The same pattern can be used for Go to map `variable` + `["readonly"]` → `["constant"]`.

## Steps to reproduce

1. Open a Go file with `const` declarations in Zed
2. Observe that `const` variables have the same color as regular variables
3. Compare with VS Code where they are styled differently

## Environment

- Zed: 0.225.12 
- OS: macOS

nianyush:
gopls reports Go `const` declarations as `variable` with the `readonly` modifier. Add a Go-specific semantic token rule to map this to the `constant` theme style, matching VS Code's behavior.

image

Closes #50642

Release Notes:

### DIFF
--- a/crates/languages/src/go.rs
+++ b/crates/languages/src/go.rs
@@ -11,6 +11,7 @@ use lsp::{LanguageServerBinary, LanguageServerName};
 use project::lsp_store::language_server_settings;
 use regex::Regex;
 use serde_json::{Value, json};
+use settings::SemanticTokenRules;
 use smol::fs;
 use std::{
     borrow::Cow,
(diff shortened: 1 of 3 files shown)


nianyush:
@cla-bot check

nianyush:
Btw in vscode doc, they have this mapping from textmate rules.https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#predefined-textmate-scope-mappings
i am not sure if we should add this in custom rules per lanaguage or just modify the global default settings. 
image

PR #50649, opened 2026-03-03
git: Add trusted worktree support to git integrations 

Anthony-Eid:
This PR cleans up the git command spawning by wrapping everything in GitBinary instead to follow a builder/factory pattern. It also extends trusted workspace support to git commands.

I also added a `clippy.toml` configuration to our git crate that warns against using `Command` struct to spawn git commands instead of going through `GitBinary`. This should help us maintain the factory pattern in the future

- [x] Added solid test coverage and/or screenshots from doing manual testing

Release Notes:

- git: Add trusted workspace support for Zed's git integration

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -1022,6 +1022,15 @@ impl RealGitRepository {
             .map(Path::to_path_buf)
     }
 
+    fn git_binary(&self) -> Result<GitBinary> {
+        Ok(GitBinary::new(
+            self.any_git_binary_path.clone(),
+            self.working_directory()?,
+            self.executor.clone(),
+            self.is_trusted(),
+        ))
+    }
+
     async fn any_git_binary_help_output(&self) -> SharedString {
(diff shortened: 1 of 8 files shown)


crates/git/src/repository.rs
```diff
@@ -3241,12 +3122,20 @@ impl GitBinary {
...
+    pub(crate) fn build_command<S>(
+        &self,
+        args: impl IntoIterator<Item = S>,
+    ) -> util::command::Command
     where
         S: AsRef<OsStr>,
     {
         let mut command = new_command(&self.git_binary_path);
         command.current_dir(&self.working_directory);
+        command.args(["-c", "core.fsmonitor=false"]);
```
lgeiger, line 3135: @Anthony-Eid On large repos disabling `fsmonitor` can make git status checks quite slow. Is it intentional that we also do that for trusted repos? Intuitively I would've expected that we should disable `fsmonitor` only for untrusted repos, or am I missing some security risks that would also apply to trusted repos?

PR #50659, opened 2026-03-04
editor: Fix selection above and below for multibyte strings

zharinov:
- Fixes: #50820

When `skip_soft_wrap` is true, columnar selection alignment used byte offsets (buffer columns) to place cursors across buffer rows. This breaks with multi-byte UTF-8 characters.

For example, consider file:

```
aaaaaa
áááááá
```

If you put the cursor in the middle and then add selection via keystroke to the another line, it jumps instead of landing on the same position:

```
aaa|aaa
á|ááááá
```

```
aaaaaa|
ááá|ááá
```

This PR switches to absolute pixel positions measured from the buffer line start, so alignment is visual rather than byte-based.

Release Notes:

- Fixed multi-cursor column selection misaligning on lines with multi-byte characters.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -15544,12 +15544,11 @@ impl Editor {
                     && let Some(oldest_selection) =
                         columnar_selections.iter().find(|s| s.id == *oldest_id)
                 {
-                    let absolute_x_range =
-                        selections_collection::absolute_x_range_for_selection(
-                            &display_map,
-                            oldest_selection,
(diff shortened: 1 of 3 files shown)


Veykril (commented):
Hm, I don't think we need to add all these new pixel based methods in selections collections, we should have all we need in the `movement` module for handling this kind of stuff I think to map columns across rows

Veykril:
Superseded by https://github.com/zed-industries/zed/pull/51780

PR #50667, opened 2026-03-04
Add .cppm (C++20 module interface) to C++ file extensions

moleium:
`.cppm` is the widely used extension for C++20 module interface units, supported by MSVC, Clang, and GCC. Currently Zed doesn't recognize it as C++, so users get no syntax highlighting or LSP support.

Changes:
`crates/languages/src/cpp/config.toml`: add cppm to path_suffixes
`crates/theme/src/icon_theme.rs`: add cppm to the C++ icon matcher

https://github.com/search?q=path%3A*.cppm&type=code

Release Notes:

- N/A


moleium:
@cla-bot check

SomeoneToIgnore (approved):
Thank you.

PR #50673, opened 2026-03-04
bedrock: Make thinking toggle toggle thinking

5herlocked:
Release Notes:

- Support for Native Thinking toggle instead of model variants


bennetbo (approved):
Thank you!

PR #50676, opened 2026-03-04
editor: Fix Accessibility Keyboard word completion corrupting text

### LINKED ISSUE #38052: Accessibility: macOS On-screen keyboard word completion issues
### Summary

On macOS, On-screen keyboard completion bug 

### Description
Steps to trigger the problem:
1. Start typing in agent writing box with the On-screen keyboard
2. Choose a word in the On-screen keyboard completion

**Expected Behavior**: Type the full word
**Actual Behavior**: Last character missing

Zed: v0.203.4 (Zed) 
OS: macOS 15.6.1
Memory: 36 GiB
Architecture: aarch64

criticic:
When typing with the macOS Accessibility Keyboard and clicking a word completion suggestion, text was corrupted in several ways.

macOS sends `insertText:replacementRange:` when a completion is selected. The replacement range was passed to `selection_replacement_ranges` using the wrong offset type, causing incorrect delta calculation for multi-cursor scenarios. Additionally, `backspace` was called unconditionally even when the replacement range was empty (as with the trailing space the Accessibility Keyboard appends), deleting the last character of every completed word. Finally, an empty replacement range in a multi-cursor context carries a stale cursor position from macOS's single-cursor view of the buffer, so it is now ignored and text is inserted at each cursor's actual position instead.

Closes #38052

Release Notes:

- Fixed text corruption when using macOS Accessibility Keyboard word completion.


Veykril (approved):
Thanks!

PR #50681, opened 2026-03-04
editor: Skip `stop_at_indent` for single-line editors

### LINKED ISSUE #50634: `editor::MoveToBeginningOfLine` breaks in Find bar on multiline searches
### Reproduction steps

1. Open the Default Settings
2. Copy more than one line (but no trailing newline) from the Default Settings into the find bar
3. Press cmd-left (`editor::MoveToBeginningOfLine`)

In a normal buffer, `editor::MoveToBeginningOfLine` stops after the indentation at the beginning of the line. You have to press cmd-left a second time to actually move the cursor to the beginning of the line.

This behaviour should not happen in the Find bar, because it only supports a single "line". This results in cmd-left stopping in a very unexpected position.

Zed: v0.225.12+stable.191.8f51e075bdf969725ce8fa2a032a19cdb821b9fb (Zed) 
OS: macOS 15.7.4
Memory: 16 GiB
Architecture: x86_64

virajbhartiya:
In single-line editors like the Find bar, MoveToBeginningOfLine with stop_at_indent should go directly to column 0 instead of stopping at the indentation level.

Closes #50634

- [ ] Aligned any UI changes with the [UI checklist](https://github.com/zedindustries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Fixed `MoveToBeginningOfLine` stopping at indentation in single-line editors like the Find bar instead of moving to column 0.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -14824,6 +14824,7 @@ impl Editor {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
+        let stop_at_indent = action.stop_at_indent && !self.mode.is_single_line();
         self.hide_mouse_cursor(HideMouseCursorOrigin::MovementAction, cx);
         self.change_selections(Default::default(), window, cx, |s| {
             s.move_cursors_with(&mut |map, head, _| {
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (approved):
Nice, thank you.

PR #50697, opened 2026-03-04
lsp: Use correct LSP adapter for completion labels in remote development

### LINKED ISSUE #47917: Python ty can't show autocompletion symbol package name.
### Reproduction steps

1. Enable ty and disable basedpyright in Zed.
2. Open a python file and type something to trigger autocomplete menu.

Auto-import works correctly. With ty, selecting some autocompletion item will trigger auto-import. The problem is that there's no way to distinguish shared names from different packages.

The autocompele menu looks like: 
Image

If disable ty and enable basedpyright, the menu look like:

Image

Zed: v0.221.4+stable.137.128478628237a6d03a6890c810d21b55d3250a70 (Zed) 
OS: macOS 26.2
Memory: 36 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json
{
  "languages": {
    "Python": {
      "language_servers": [
        "ty",
        "ruff",
        "!basedpyright",
      ]
    }
  }
}
```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

lingyaochu:
Closes #47917

Currently, during remote development, Zed uses the first available local LSP adapter to handle all label computing. This isn't ideal and causes bugs because different adapters may expect different label formats. By leveraging the `all_capable_for_proto_request` method, we can now retrieve the specific language server name and use it as a key to find the correct LSP adapter for label population. Even if this lookup fails, we can still fall back to the first adapter, so this PR should provide more accurate label population than before.

This addresses the root cause of #47917, which stems from two main issues. For example, in remote Python development, the `basedpyright` adapter might incorrectly handle labels even when the remote server is actually `ty`. The completion items returned by `ty` are slightly different from `basedpyright`: `ty` stores labels in `labelDetails.detail`, while basedpyright uses `labelDetails.description`. By matching the correct adapter, we ensure labels are populated properly for completion items.
```json
// RPC message returned by `ty`, label is in `labelDetails.detail`
{
  ...
  "labelDetails": { "detail": " (import pathlib)" },
  ...
}

// RPC message returned by `basedpyright`, label is in `labelDetails.description`
{
  ...
  "labelDetails": { "description": "pathlib" },
  ...
}
```
Additionally, adapters registered via `register_available_lsp_adapter` are lazy-loaded into the `LanguageRegistry` (which is the case for `ty` before). In remote scenarios, the adapter might be loaded on the remote server but not on the local host, making it hard to find in `lsp_adapters`. This is partially resolved in #50662, and combined with this PR, we can fully address #47917.

There is still more to do, however. In some cases, we still can't find the correct local LSP adapter if the local host lacks the registry that the remote server has; this typically happens when the adapter is registered via `register_available_lsp_adapter`. I've opened a feature discussion #49178 to track this. If it's decided that this needs further refinement, I'm happy to continue working on it.

Release Notes:

- Fixed missing labels for `ty` completion items in remote development.

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -4895,7 +4895,7 @@ impl LspStore {
         buffer: &Entity<Buffer>,
         mut check: F,
         cx: &App,
-    ) -> Vec<lsp::LanguageServerId>
+    ) -> Vec<(lsp::LanguageServerId, lsp::LanguageServerName)>
     where
         F: FnMut(&lsp::LanguageServerName, &lsp::ServerCapabilities) -> bool,
     {
@@ -4925,7 +4925,7 @@ impl LspStore {
                     .map(|c| (server_id, server_name, c))
             })
             .filter(|(_, server_name, capabilities)| check(server_name, capabilities))
-            .map(|(server_id, _, _)| *server_id)
+            .map(|(server_id, server_name, _)| (*server_id, server_name.clone()))
             .collect()
     }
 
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #50705, opened 2026-03-04
picker: Prevent clicking non-selectable entries from confirming selection

### LINKED ISSUE #50627: "Recent Projects" picker title is weird
### Reproduction steps

1. Press cmd-opt-o (`projects::OpenRecent`)

Image

There are a few things wrong here:

- The "Recent Projects" title is almost the _smallest_ UI element, so it looks off
- Clicking the title counts as clicking the first item in the list

I'm confused why the design differs from the other pickers (e.g. the command palette and file finder). Just mention "recent projects" in the placeholder text and get rid of the title entirely.

Zed: v0.225.12+stable.191.8f51e075bdf969725ce8fa2a032a19cdb821b9fb (Zed) 
OS: macOS 15.7.4
Memory: 16 GiB
Architecture: x86_64

Dnreikronos:
Closes #50627

### Videos:
Before:
https://drive.google.com/file/d/1PahGhfx-wq9cyNqqMlctb9Np7M1Meo71/view?usp=sharing
After:
https://drive.google.com/file/d/135W6MQ9hKBurw5Z7YpHQQhoQwkh1NvUb/view?usp=sharing

Release Notes:

- Fixed clicking on non-selectable picker entries (e.g. section headers) confirming the currently selected item.

### DIFF
--- a/crates/picker/src/picker.rs
+++ b/crates/picker/src/picker.rs
@@ -750,16 +750,17 @@ impl<D: PickerDelegate> Picker<D> {
     }
 
     fn render_element(
-        &self,
+        &mut self,
         window: &mut Window,
         cx: &mut Context<Self>,
         ix: usize,
     ) -> impl IntoElement + use<D> {
         let item_bounds = self.item_bounds.clone();
+        let selectable = self.delegate.can_select(ix, window, cx);
 
         div()
             .id(("item", ix))
-            .cursor_pointer()
+            .when(selectable, |this| this.cursor_pointer())
             .child(
                 canvas(
                     move |bounds, _window, _cx| {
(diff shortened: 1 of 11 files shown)


MrSubidubi (approved):
Thanks, this looks like a good fix for the issue at hand. 

I'd like if we could add a test for this behavior to ensure that this does not regress. Could you look into that? 

Also, one thing that popped up while I looked at this is that we actually show a pointer cursor when hovering the element despite it not being clickable. Could I interest you into looking to fix that as well perhaps? Not insisting, just curious.

Edit: Hiding because wrong button.

MrSubidubi (changes requested):
Thanks, this looks like a good fix for the issue at hand.

I'd like if we could add a test for this behavior to ensure that this does not regress. Could you look into that?

Also, one thing that popped up while I looked at this is that we actually show a pointer cursor when hovering the element despite it not being clickable. Could I interest you into looking to fix that as well perhaps? Not insisting, just curious.

crates/picker/src/picker.rs
```diff
@@ -747,16 +750,17 @@ impl<D: PickerDelegate> Picker<D> {
     }
 
     fn render_element(
-        &self,
+        &mut self,
```
MrSubidubi: Not your fault here, but let's revert this and change the `can_select` method on the trait to only take `&self` instead of the mutable reference

crates/picker/src/picker.rs
```diff
@@ -850,6 +854,184 @@ impl<D: PickerDelegate> Picker<D> {
...
+    }
+
+    fn init_test(cx: &mut TestAppContext) {
+        cx.update(|cx| {
+            let store = settings::SettingsStore::test(cx);
+            cx.set_global(store);
+            theme::init(theme::LoadThemes::JustBase, cx);
+            editor::init(cx);
+        });
+    }
```
MrSubidubi, line 955: Have you checked whether we actually need all of this? I'd suspect we might actually not, but might be proven wrong (and happy to). Would like to avoid pulling in the new dependency if we can.

MrSubidubi (changes requested):
Very very nice, thanks for looking into this and the quick follow-up!

Two minor comments and we can 🚢

crates/picker/src/picker.rs
Dnreikronos: Done! Changed **can_select** on the **PickerDelegate** trait to take **&self** instead of **&mut self**, and updated all 9 implementations across the codebase to match. This lets **render_element** stay as **&self** since **can_select** no longer requires a mutable borrow. Since **can_select** is defined on the trait, changing its signature requires updating every implementation, otherwise the compiler rejects the mismatch. This affected 8 other files across **agent_ui, rules_library, recent_projects, and sidebar**. None of them actually needed **&mut self**, they were just following the original trait signature.

crates/picker/src/picker.rs
Dnreikronos: Checked and removed ctor, env_logger, and serde_json from dev-dependencies — none were actually needed. ctor and env_logger were only used for an init_logger helper which isn't required for the tests to run. serde_json wasn't used at all. The remaining dev-dependencies (editor, gpui, settings) are necessary since editor::init registers the ErasedEditor factory that the picker needs, and settings/theme::init are required by the editor.

MrSubidubi (approved):
Awesome work here and very much appreciate the work that went into the test. Thank you!

Issue #35616, opened 2025-08-04
Helix mode, problems near empty line at the end of the file


romaninsh:
### Summary

Second-to-last line behaves weirdly if **inlay-hints** are enabled and **last line is empty**.

1. Use this minimal config:

```json
{
  "ui_font_size": 16,
  "buffer_font_size": 15,
  "vim_mode": true,
  "inlay_hints": {
    "enabled": true // <-- IMPORTANT FOR THIS BUG
  },
  "theme": {
    "mode": "system",
    "light": "One Light",
    "dark": "One Dark"
  }
}
```

2. create `test.rs` file:
```rust
mod test {

    #[test]
    fn test_new() {
        // repeat enough times to cover entire screen vertically
        assert!(true);
        assert!(true);
        assert!(true);
        assert!(true);
        assert!(true);
        assert!(true);
        assert!(true);
        assert!(true);
    }
}
```

IMPORTANT: label `mod test` should be visible on the last line

<img width="205" height="89" alt="Image" src="https://github.com/user-attachments/assets/6e3b9c1f-88a4-45bc-9d57-b3d1b49edc56" />

3. Position cursor like so:

```rust
        });
    ˇ}
}

```

4. press `j`

Expected: cursor should appear on next line `}ˇ`, just before the inlay text.
Actual: cursor jumps over next line, goes to the last empty line instead

Probably related problem. This one requires Helix mode on:

5. position cursor here:

```rust
        });
    }ˇ mod test <-- NON-SELECTABLE INLAY
}

```
6. press `V` or `x`

Expected: line with `}` should be highlighted
Actual: line with `}` and next empty line is highlighted

Weirdly, if you disable `inlay_hints`, problem is not occurring.

### Zed Version and System Specs

Zed: v0.198.1 (Zed Preview) 
OS: macOS 15.5.0
Memory: 32 GiB
Architecture: aarch64

PR #50709, opened 2026-03-04
vim: Fix visual mode entry at line end near trailing newline

SkandaBhat:
There's an edge case while entering Visual mode where movement::right jumps to the next line on the second last line, and when the last line is a trailing newline.

- Added to the vim visual tests to cover this case
- to fix this, if the cursor is at the line end, we snap it to the last character on that line. if not, we calculate selection.end as usual

before fix: 
zed_repro_before

after fix:
zed_repro_after

Release Notes:

- Fixed `shift-a` in Helix select mode placing the cursor on the wrong line after selecting with `x`


dinocosta (commented):
Hey @SkandaBhat 👋

Thank you so much for picking this one up! I can reproduce the bug you’re working on, but after looking at the changes, I’m not sure ‎`Vim::switch_mode` is the right place to handle this 🤔

From what I can tell, this bug only appears when using Helix mode, not Vim mode itself. It seems tied to how the line selection is calculated when the cursor is on the newline character, which isn’t possible in Vim mode unless you’re already in a visual mode.

With that in mind, I wonder if the fix should instead live in how we handle the ‎`vim::HelixSelectLine` action and its motion. Testing in Helix, if the cursor is at the last position of the line and you press ‎`x`, the whole line is selected but the cursor stays put. In Zed’s implementation, ‎`x` selects the whole line but then moves the cursor to the first column of the next line, which suggests that’s where the fix needs to go.

One simple way to see this mismatch is to select multiple lines with ‎`x` and then press ‎`A` to start inserting at the end of the last selected line. In Helix, if you select three lines, the cursor ends up at the end of the last selected line; in Zed’s implementation, it lands on the line below 🫠

Let me know if you have any questions or would like more guidance. Happy to help!

dinocosta (approved):
Hey @SkandaBhat 👋 

I've updated your branch to be up to date with `main`, fixing an existing conflict, and ended up cleaning it up a bit, as it was now concerned with `shift-a` behavior after using `x` to select a whole line, then the issue it was associated with. 

As such, I'm going to update the description so it doesn't actually mention that it closes https://github.com/zed-industries/zed/issues/35616 , as that seems to be a little bit more tricky that what initially thought .

Thanks! 🙂

PR #50733, opened 2026-03-04
collab_ui: Add an incoming call sound effect when a collaborator calls you

### LINKED ISSUE #5139: Add a ring tone a new call is pending for a user.
### Summary

When somebody invites you to a pair coding session, there's no audio indicator that makes it hard to know there's a call.

A nice audio notification that is not too obnoxious.

nathan-fiscaletti:
Fixes #5139 

**Relates to discussion**: https://github.com/zed-industries/zed/discussions/50714

I'd like to preface by saying that I'm a first time contributor to this project. I realize that the normal channel for pull-requests is to wait for an issue to officially be opened based on a discussion, but with how simple this was to implement and the value I believe it adds I thought it was worth just creating a pull-request directly.

This pull-request adds an "incoming call" sound effect (with a setting for enabling/disabling it) when receiving a call from a collaborator.

The `agent_done.wav` file was copied into the `incoming_call.wav` file. I don't know how to create a "new" sound effect, so if it's a problem that it's being re-used I'm happy to accept a different sound effect. (I personally think that this sound effect sounds good as a "ring tone" style sound effect.)

The new setting defaults to an `enabled` state. If it would be preferable for it to default to a `disabled` state, I'm happy to make that change as well.

---

Release Notes:

- Added a sound effect when receiving an incoming call from a collaborator


nathan-fiscaletti:
> I have polled for this feature internally and looks like no one in the team uses 1-1 calls and wants to support this.

Other issues with this:
- When using channels as opposed to direct calls, screen sharing seems to be at a much lower bit rate. One of the major reasons we moved to Zed for collaboration / calls is that the screen sharing was immensely more clear than most other platforms. Using channels removes this benefit.

- When using channels, screen sharing and sharing a workspace are separate. This means if I'm following another collaborator in the workspace, when they move to an application/window that is outside of the workspace, the "follow" feature does not automatically move me to their screen share.

Overall, this makes collaboration much less seamless when working with channels.

The only feature we really felt was missing from solo-collaboration was an actual ring tone when a call was incoming (which is what this PR was intended to fix.)

PR #50738, opened 2026-03-04
Ollama: Add support for DMR and response serialization

sbe-arg:
- Extends Tool Support assuming tools are supported as fallback unless told by the model that tools are unsuported. Makes Docker Model Runner llmstools work.

- Serialize responses to avoid missing expected values, Currently Docker Model Runner responses are field limited and loops trough triying to get context such as role and others.

- Fixes error:

```
Unable to parse chat response: unknown variant ``, expected one of `assistant`, `user`, `system`, `tool` at line 1 column 108
```

With this you can use https://docs.docker.com/ai/model-runner/ as ollama endpoint.

### DIFF
--- a/crates/ollama/src/ollama.rs
+++ b/crates/ollama/src/ollama.rs
@@ -134,11 +134,61 @@ pub struct ChatOptions {
 pub struct ChatResponseDelta {
     pub model: String,
     pub created_at: String,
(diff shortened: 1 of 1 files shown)


sbe-arg:
@benbrandt ? can you tag someone to check on this pr hahaha thanks in advance

PR #50769, opened 2026-03-05
Fix blame hover popover not showing on first trigger when inline blame is disabled

### LINKED ISSUE #50285: Git: Blame popover out of sync with GitBlame state
### Reproduction steps

1. Have a project with git repo
2. Have settings with disabled inline blame (settings.json provided below)
3. Open some file
4. Trigger `editor::BlameHover` command via palette or <kbd>⌘K ⌘B</kbd>

https://github.com/user-attachments/assets/298b197f-9c3e-4679-990f-254866a88e02

Current behavior: After the first trigger the blame popover does not appear and on the second trigger - it appears
Expected behavior: Blame popover appears after first trigger

There was an issue for that already #42936 and fix in #42992 but it is not completely fixed as can be seen in the video
It happens due to early return in `blame_hover` here https://github.com/zed-industries/zed/blob/365e4157b71e1faaf63f6a4d87db1116ccd30f2d/crates/editor/src/editor.rs#L7287
After `start_git_blame` was called it continues synchronously and GitBlame buffers has no entries at that point of time.
I was able to fix it locally, but the solution is a bit awkward since I don't know Rust very well.
It's better to leave this to more experienced contributors

Zed: v0.225.10+stable.184.7c950f7b7c11ed485b0e7d72216ca36ebc86b58b (Zed) 
OS: macOS 15.7.3
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

```json
"git": {
  "inline_blame": {
    "enabled": false
  }
}
```

</details>

### Relevant Keymap

_No response_

_No response_

None

Dnreikronos:
● Closes #50285

## Summary

When `inline_blame` is disabled in settings and `editor::BlameHover` is triggered,
the blame popover fails to appear on the first invocation because `start_git_blame`
kicks off an async task to fetch blame data, but `blame_hover` immediately tries to
read that data synchronously before it's available.

This fix registers a one-shot observation on the blame entity when it's freshly
created, so the popover is shown once blame data has finished generating.

- [ ] ~Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)~ (No UI changes)

### Before:
https://github.com/user-attachments/assets/d2ca4f8e-8186-49af-9908-5a82bfca0de2

Ps: pressed the keybind two times

### After:
https://github.com/user-attachments/assets/08575cd9-2bbc-462d-92fa-f0e7ef23d8d6

## Release Notes:

- Fixed blame hover popover not appearing on first trigger when inline blame is disabled (#50285)


cole-miller:
Thanks! The core change here makes sense, but it seems like there is also an unrelated set of changes to the LSP store--would you mind removing that part of the patch to keep this focused?

Dnreikronos:
Done. I removed the LSP store changes, so lsp_store.rs and project.rs now match main and the PR only touches the blame hover code. I also rebased on main to clear the merge conflict. Happy to adjust if you spot anything else.

ChristopherBiscardi (approved):
thanks!

PR #50770, opened 2026-03-05
Fix parsing of filenames like main (1).log

hagz0r:
## Summary
Fixes Windows file-open parsing for names like `main (1).log`.

`PathWithPosition::parse_str` could treat `(1)` in a normal filename as a position suffix and drop the extension/path tail. The regex is now anchored so parenthesized row/column parsing only applies at the end of the filename (with optional trailing `:` and optional range suffix).

## Testing
- `cargo test -p util path_with_position_parse_`

Closes #50597

Release Notes:

- Fixed opening files with names like `main (1).log` on Windows.

### DIFF
--- a/crates/util/src/paths.rs
+++ b/crates/util/src/paths.rs
@@ -595,10 +595,6 @@ const ROW_COL_CAPTURE_REGEX: &str = r"(?xs)
         |
         \((\d+)\)()     # filename:(row)
     )
-    (?:
-        -\(\d+(?:[,:]\d+)?\) # filename:(row,column)-(row,column), filename:(row,column)-(row)
-    )?
-    \:*$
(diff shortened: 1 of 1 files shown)


crates/util/src/paths.rs
```diff
@@ -595,12 +595,17 @@ const ROW_COL_CAPTURE_REGEX: &str = r"(?xs)
         |
         \((\d+)\)()     # filename:(row)
     )
+    (?:
+        -\(\d+(?:[,:]\d+)?\) # filename:(row,column)-(row,column), filename:(row,column)-(row)
+    )?
```
Veykril: This bit seems unnecessary / should not be part of this bug fix?

PR #50777, opened 2026-03-05
editor: Go to previous and next symbol actions

nishanthkarthik:
Closes discussion #34890 

This is similar to the vim prev/next method/section motion, but more flexible because this follows the items in editor's outline (Tree sitter or LSP provided).

Release Notes:

- Added actions `editor::GoToPreviousSymbol` and `editor::GoToNextSymbol` actions to go to the previous and next outline symbol. This is either the tree sitter outline, or the LSP provided outline depending on the configuration.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -18434,23 +18434,30 @@ impl Editor {
         offset: i8,
     ) -> Task<Result<()>> {
         let editor_snapshot = self.snapshot(window, cx);
-        let Some((excerpt_id, _, buffer_snapshot)) = editor_snapshot.as_singleton() else {
+
+        // We don't care about multi-buffer symbols
+        let Some((excerpt_id, _, _)) = editor_snapshot.as_singleton() else {
             return Task::ready(Ok(()));
         };
 
-        let task = self.buffer_outline_items(buffer_snapshot.remote_id(), cx);
-        let multi_snapshot = self.buffer().read(cx).snapshot(cx);
         let cursor_offset = self
             .selections
-            .newest::<MultiBufferOffset>(&self.display_snapshot(cx))
+            .newest::<MultiBufferOffset>(&editor_snapshot.display_snapshot)
             .head();
 
         cx.spawn_in(window, async move |editor, wcx| -> Result<()> {
-            let Some(editor) = editor.upgrade() else {
+            let Ok(Some(remote_id)) = editor.update(wcx, |ed, cx| {
+                let buffer = ed.buffer.read(cx).as_singleton()?;
+                Some(buffer.read(cx).remote_id())
+            }) else {
                 return Ok(());
             };
 
+            let task = editor.update(wcx, |ed, cx| ed.buffer_outline_items(remote_id, cx))?;
             let outline_items: Vec<OutlineItem<text::Anchor>> = task.await;
+
+            let multi_snapshot = editor_snapshot.buffer();
+            let buffer_range = |range: &Range<_>| Anchor::range_in_buffer(excerpt_id, range.clone()).to_offset(multi_snapshot);
 
             wcx.update_window(wcx.window_handle(), |_, window, acx| {
                 let current_idx = outline_items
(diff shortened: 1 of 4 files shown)


crates/editor/src/editor.rs
```diff
@@ -18464,6 +18427,115 @@ impl Editor {
...
+        let editor_snapshot = self.snapshot(window, cx);
+        let Some((excerpt_id, _, buffer_snapshot)) = editor_snapshot.as_singleton() else {
+            return Task::ready(Ok(()));
+        };
+
+        let task = self.buffer_outline_items(buffer_snapshot.remote_id(), cx);
+        let multi_snapshot = self.buffer().read(cx).snapshot(cx);
+        let cursor_offset = self
+            .selections
+            .newest::<MultiBufferOffset>(&self.display_snapshot(cx))
```
RemcoSmitsDev: Small note: `self.display_snapshot(cx)` is quite expensive to call and is not needed in this case, you can use the `display_snapshot` from the `editor_snapshot` instead. So we skip an extra display_snapshot creation which keeps it performant :)

crates/editor/src/editor.rs
```diff
@@ -18464,6 +18427,115 @@ impl Editor {
...
+        window: &mut Window,
+        cx: &mut Context<Self>,
+        offset: i8,
+    ) -> Task<Result<()>> {
+        let editor_snapshot = self.snapshot(window, cx);
+        let Some((excerpt_id, _, buffer_snapshot)) = editor_snapshot.as_singleton() else {
+            return Task::ready(Ok(()));
+        };
+
+        let task = self.buffer_outline_items(buffer_snapshot.remote_id(), cx);
```
SomeoneToIgnore: Instead of taking any snapshots here, seems that we could go `self.buffer().as_singleton()` as less resource-intensive.

crates/editor/src/editor.rs
```diff
@@ -18464,6 +18427,115 @@ impl Editor {
...
+        let multi_snapshot = self.buffer().read(cx).snapshot(cx);
+        let cursor_offset = self
+            .selections
+            .newest::<MultiBufferOffset>(&self.display_snapshot(cx))
+            .head();
+
+        cx.spawn_in(window, async move |editor, wcx| -> Result<()> {
+            let Some(editor) = editor.upgrade() else {
```
SomeoneToIgnore: We'd better do that after `.await` or not do at all, just `.ok()` the selections' update instead below.

SomeoneToIgnore (commented):
As commented by @\RemcoSmitsDev too, new code definitely benefits from less operations invoked: besides snapshots, there's something suspicious in how we want to list all outlines and sort them, that might take a while for large files like `editor.rs` itself.

Another concern is that we don't use `symbol` that much and seem to have somewhat similar methods already?

Image

I'm particularly interested in `SelectNextSyntaxNode` and its `Prev` counterpart — how far it is from what's needed?
It seems that we can either rework it, or use a `SyntaxNode`-based name for the new action at least?
Ideally, we could try to reuse `syntax_next_sibling` and related methods to simplify the calculations.

---------------------

Naming and approaches aside, the current implementation seems to loose track of its neighbours when placed on a blank line?

For example, here:


```py
class A:
    def f(self):
        pass

    def g(self):
        pass
<CARET>

class B:
    def f(self):
        pass

    def g(self):
        pass


class C:
    def f(self):
        pass

    def g(self):
        pass
```

I cannot move anywhere using new actions, which looks like a bug worthy of a fix and a test.

Should we also allow wrapping over the editor's edge when we reach the first/last symbol?

nishanthkarthik:
Thanks for the feedback!

> As commented by @\RemcoSmitsDev too, new code definitely benefits from less operations invoked: besides snapshots, there's something suspicious in how we want to list all outlines and sort them, that might take a while for large files like editor.rs itself.

I profiled large files (about thousand outline items) in my first version. The performance was not quite great, in the order of few hundred milliseconds. There are two ideas I want to implement in the future -

- We already have primitives to navigate to previous and next nodes. Outline scm should provide us the same structure from `buffer_outline_items`. We should be able to do it in constant time instead of linear time. I'm thinking something built on top of `BufferSnapshot::next_outline_item`.
- For really complex and large files, LSPs are likely better at incremental updates. Currently the outline panel refreshes itself (and fetches all outline items) on editor updates. I have not tested the performance of LSP provided outlines for large files.

> Another concern is that we don't use symbol that much and seem to have somewhat similar methods already?

I did not want to use `syntax node` because that's tree-sitter specific. It would be nicer to not tie this implementation to tree-sitter or LSP. LSPs have their own notion of syntax nodes, but `syntax node` actions specifically refer to tree sitter ones in Zed. I picked the name `Symbols` because the outline panel also uses the term "Search buffer symbols..." to filter outline items.

crates/editor/src/editor.rs
nishanthkarthik: I looked at buffer methods but I could not understand your suggestion. Could you please explain if there is a way to avoid snapshotting? :)

SomeoneToIgnore (approved):
Thank you, looks like the snapshot situation had become better.

Regarding the wrap around and large file implications, let's merge and see how things work for now.
Overall, both are good to improve: for the wrap around, we could add a parameter into actions.

PR #50798, opened 2026-03-05
git_ui: Disable Push/Fetch/Pull buttons while operation is in progress

virajbhartiya:
Closes #50779

- Track pending remote operations in `GitPanel` via a `pending_remote_operation` field so Push/Fetch/Pull buttons are disabled during execution
- Buttons show a spinning icon and muted label while an operation is running, preventing duplicate concurrent git operations
- Added early-return guards in `fetch()`, `pull()`, and `push()` to prevent duplicate invocations via keyboard shortcuts

https://github.com/user-attachments/assets/8858e807-6aae-434f-8c27-1a43653708c3

Release Notes:

- Fixed Push/Fetch/Pull buttons remaining clickable while their operation is in progress

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -653,7 +653,7 @@ pub struct GitPanel {
     local_committer_task: Option<Task<()>>,
     bulk_staging: Option<BulkStaging>,
     stash_entries: GitStash,
-
+    pending_remote_operation: Option<Task<()>>,
     _settings_subscription: Subscription,
 }
 
@@ -826,6 +826,7 @@ impl GitPanel {
                 entry_count: 0,
                 bulk_staging: None,
                 stash_entries: Default::default(),
+                pending_remote_operation: None,
                 _settings_subscription,
             };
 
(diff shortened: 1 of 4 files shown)


cole-miller:
Thanks, and sorry for the delay in reviewing!

Disabling the button when a remote operation is in progress makes sense. There are a couple of things we can improve in the UX:

- We shouldn't have a spinner on the button, since this duplicates the spinner that already appears in the activity bar (at the bottom of the window) when a remote operation is in progress.
- Instead, we should just show the button as disabled (the way it is now), with the same text that it had (including the counts). It should have a tooltip that says "Push in progress" or similar.

I also think that we should read the information about whether a remote operation is in progress out of the `GitStore`, rather than keeping it in the panel. This will allow passing the right `disabled` value in the empty project diff element, for example. We can achieve this by adding an `is_remote_operation` boolean field to the git job system, and then checking that field for the current job.

ChristopherBiscardi:
This seems to be a duplicate PR of #49841 , so I'm going to close it for the same reasons. Having two pushes active is not actually a problem and there is already a progress indicator for the active action. Disabling buttons is typically to prevent duplicate submissions of a sensitive form, or similar action that needs to be deduplicated, which pulling, etc do not need to be. Running pull multiple times is fine.

prior reasoning: https://github.com/zed-industries/zed/pull/49841#issuecomment-4325629048

PR #50800, opened 2026-03-05
languages: Support more JSON configuration files

rodrii-dp:
Several widely used configuration files are JSON but have no `.json` extension,
such as `.babelrc` or `.eslintrc`.

Editors like VS Code detect these as JSON automatically, but Zed currently
treats them as plain text.

This change adds common JSON config suffixes so they get proper JSON syntax
highlighting.

Added:
- [`.babelrc`](https://babeljs.io/docs/config-files)
- [`.eslintrc`](https://archive.eslint.org/docs/user-guide/configuring/configuration-files)
- [`.stylelintrc`](https://stylelint.io/user-guide/configure/)
- [`.swcrc`](https://github.com/swc-project/swc/pull/4236)

Release Notes:

- Support more JSON configuration files

### DIFF
--- a/crates/languages/src/json/config.toml
+++ b/crates/languages/src/json/config.toml
@@ -1,6 +1,6 @@
 name = "JSON"
 grammar = "json"
-path_suffixes = ["json", "flake.lock", "geojson", "prettierrc"]
+path_suffixes = ["json", "flake.lock", "geojson", "prettierrc", "babelrc", "eslintrc", "stylelintrc", "swcrc"]
 line_comments = ["// "]
 autoclose_before = ",]}"
 brackets = [
--- a/crates/languages/src/json/config.toml
+++ b/crates/languages/src/json/config.toml
@@ -1,6 +1,6 @@
 name = "JSON"
 grammar = "json"
-path_suffixes = ["json", "flake.lock", "geojson", "prettierrc", "babelrc", "eslintrc", "stylelintrc", "swcrc"]
+path_suffixes = ["json", "flake.lock", "geojson", "prettierrc"]
 line_comments = ["// "]
 autoclose_before = ",]}"
 brackets = [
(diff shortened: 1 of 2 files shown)


crates/languages/src/json/config.toml
```diff
@@ -1,6 +1,6 @@
 name = "JSON"
 grammar = "json"
-path_suffixes = ["json", "flake.lock", "geojson", "prettierrc"]
+path_suffixes = ["json", "flake.lock", "geojson", "prettierrc", "babelrc", "eslintrc", "stylelintrc", "swcrc"]
```
yeskunall: @rodrii-dp IIRC `.swcrc` [supports](https://github.com/swc-project/swc/pull/4236) comments in JSON ([JSONC](https://jsonc.org/)), and `.babelrc` is based off the [JSON5 spec](https://json5.org/). So at the very least, for `.swcrc`, the `path_suffixes` should be modified [here](https://github.com/zed-industries/zed/blob/main/crates/languages/src/jsonc/config.toml#L3) instead.

Still looking into how Babel config files should be handled. LMK if you find a good resource before I do.

yeskunall (changes requested):

crates/languages/src/json/config.toml
rodrii-dp: Good catch, thanks. I didn’t realize .swcrc uses JSONC. I’ll move it to the JSONC grammar once I get home.

Regarding .babelrc, I see it’s based on JSON5

yeskunall: Yeah -- and we have no dedicated JSON5 handler. So, either a new language crate (I don’t want that, if anything I want to reduce the amount of code Zed ships with) -- or we handle `.babelrc` as JSONC as well and be done with it.

Being vague here intentionally as I still don’t have a concrete answer for Babel.

yeskunall (approved):

PR #50816, opened 2026-03-05
Fix whitespace marker ranges for non-ASCII spaces

### LINKED ISSUE #49186: "Show Whitespace" does not show trailing space if line contains non-breaking space
### Reproduction steps

1. Set "Show Whitespace" to "All". (You'll see essentially the same bug with "Boundary" or "Trailing", or "Selection" if you then select text.)
2. Copy this into a new buffer:

```
start
nbsp: regular space: trailing space: 
regular space: regular space: trailing space: 
end
```

The trailing space does not show up on the line with the non-breaking space, but it should:

Image

Nor does the one after "regular space:" on that line, curiously.

I've confirmed this happens with all the built-in themes, not just my custom one.

OS: macOS 26.2
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

No log output if I just make a new file and paste the repro.

### Relevant Zed settings

-

### Relevant Keymap

-

-

None

siewcapital:
## Summary

- Store both start and end byte offsets for  instead of assuming every whitespace is one byte
- Use the real UTF-8 span when creating whitespace invisibles so downstream adjacency checks stay correct  
- Add a regression test that covers non-breaking space (U+00A0) followed by ASCII whitespace

## Why

Previously, the code used  for whitespace ranges, which is incorrect for multibyte whitespace characters like non-breaking space (2 bytes in UTF-8). This could break boundary/trailing whitespace rendering logic once a line contains non-ASCII whitespace.

Fixes #49186

## Testing

- Added regression test: 
- Verified existing invisibles tests still pass
- Attempted full test suite (environment limitations prevented full run)

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -8970,7 +8970,8 @@ impl LineWithInvisibles {
                                             && (non_whitespace_added || !is_soft_wrapped)
                                         {
                                             Some(Invisible::Whitespace {
-                                                line_offset: line.len() + index,
+                                                line_start_offset: line.len() + index,
+                                                line_end_offset: line.len() + index + c.len_utf8(),
                                             })
                                         } else {
                                             None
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
Without the CLA we cannot deal with PRs much, and CLA had not been dealt with for 5 days, hence will close this as stale for now.

image

PR #50817, opened 2026-03-05
markdown: Fix code block scrollbars flashing on vertical scroll

cppcoffee:
Release Notes:

- Fixed code block scrollbars flashing on vertical scroll

before:

When there are many code blocks, scrolling through markdown will display a horizontal scrollbar (when the mouse is not inside a code block).

https://github.com/user-attachments/assets/1fae36ec-5a3f-4283-b54f-e5cb4f45646b

after:

When scrolling markdown, do not display the horizontal scrollbar when the mouse is not in a code block.

https://github.com/user-attachments/assets/0c0f2016-9b18-4055-87a6-4f508dbfd193

### DIFF
--- a/crates/ui/src/components/scrollbar.rs
+++ b/crates/ui/src/components/scrollbar.rs
@@ -1053,8 +1053,6 @@ impl PartialEq for ScrollbarLayout {
 
         thumb_offset == other_thumb_offset
             && self.thumb_bounds.size.along(axis) == other.thumb_bounds.size.along(axis)
-            && self.track_bounds.size.along(axis) == other.track_bounds.size.along(axis)
-            && self.reserved_space == other.reserved_space
     }
 }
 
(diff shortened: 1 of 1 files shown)


MrSubidubi (changes requested):
Thank you for this, but I do not think this is the best solution for the issue at hand. I think we should rather check whether we can modify the PartialEq impl for the `ScrollbarLayoutState` so that this does not happen. I'd suspect we can instead of comparing the entire bounds compare just the size of the bounds, as that should not change. But I think this might be a more sensible fix - the caller should not be bothered with this. What do you think?

cppcoffee:
Hi @MrSubidubi , You are right. Now, modify the ScrollbarLayout, and the effect is as follows:


https://github.com/user-attachments/assets/cc899acb-ff4b-477d-a5c7-2a8990050817

MrSubidubi (changes requested):
This looks much better, thank you!

I am not sure as to whether we need the track bounds or the reserved space here too though - both were not present before. Can we not remove these? IIRC we do not really update these anyway for a thumb layout between frames at least.

MrSubidubi (approved):
Nice find and fix, thanks for the follow-ups!

PR #50819, opened 2026-03-05
agent_ui: Group together pending permissions above AI chat input

MathiasWP:
I have no idea how many times i have thought that my Claude Code agent was buggy and hanging, before i realised that it was because i hadn't provided an answer to a permission that was drowned in the chat. This PR improves the interaction IMO by having a "Pending permissions" section right above the chat input, so that you can easily see the permissions you have to interact it.

Here's how it looks like in a normal chat. Note that the existing pending approvals are not removed inline in the chat, i think that they should also be there:

image

And here's how it looks like when you have a pending message:

image

### Changes

- Added `all_pending_tool_calls()` to `Conversation` to collect all pending permission requests across threads
- Added `render_pending_permissions()` to `ThreadView` that groups all pending tool call permission requests in a section between the chat and the input area
- Styled to match native tool cards with editor background and border colors, showing a "Pending Permissions (N)" header with the total count and each tool call's label

### Release Notes:

- Improved permission prompts in the agent panel to be grouped above the chat input instead of inline

### DIFF
--- a/crates/agent_ui/src/connection_view.rs
+++ b/crates/agent_ui/src/connection_view.rs
@@ -161,14 +161,6 @@ pub(crate) struct Conversation {
 }
 
 impl Conversation {
-    pub fn thread(
-        &self,
-        session_id: &acp::SessionId,
-        _cx: &App,
-    ) -> Option<&Entity<AcpThread>> {
-        self.threads.get(session_id)
-    }
-
     pub fn register_thread(&mut self, thread: Entity<AcpThread>, cx: &mut Context<Self>) {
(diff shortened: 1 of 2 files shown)


MathiasWP:
@cla-bot check

MathiasWP:
> Thank you for this PR. As per our [contrubution guidelines](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#sending-changes), could you please add some screenshot comparisons here as to what was changed/added by your PR?

Yes! I just added them! @MrSubidubi

PR #50822, opened 2026-03-05
vim: Fix visual block Shift-I jumping to original cursor after Ctrl-D scroll

cppcoffee:
Release Notes:

- Fixed visual block `shift-i` jumping to the initial cursor position after scrolling with `ctrl-d`/`ctrl-u`

In Vim mode, press Ctrl+V to enter line selection, press Ctrl+D to start scrolling, then press Shift+I to insert into the newly selected multiple lines.

Current: Jump to the beginning
Expected: Enter multi-line insert mode

before:

https://github.com/user-attachments/assets/3cf6489c-828d-44d8-b575-239b68e50c2b

after:

https://github.com/user-attachments/assets/a086f6d1-abeb-464d-8e09-5a3c1173840c

vim behaviour:

https://github.com/user-attachments/assets/900cb540-9cca-4e90-9796-dcc56874c8b7

### DIFF
--- a/crates/vim/src/normal/scroll.rs
+++ b/crates/vim/src/normal/scroll.rs
@@ -97,13 +97,14 @@ impl Vim {
         let mode = self.mode;
         Vim::take_forced_motion(cx);
         self.exit_temporary_normal(window, cx);
-        self.update_editor(cx, |_, editor, cx| {
-            scroll_editor(editor, mode, move_cursor, amount, window, cx)
+        self.update_editor(cx, |vim, editor, cx| {
+            scroll_editor(vim, editor, mode, move_cursor, amount, window, cx)
         });
     }
 }
 
 fn scroll_editor(
+    vim: &mut Vim,
     editor: &mut Editor,
     mode: Mode,
     preserve_cursor_position: bool,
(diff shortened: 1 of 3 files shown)


crates/vim/src/normal/scroll.rs
```diff
@@ -97,13 +97,14 @@ impl Vim {
         let mode = self.mode;
         Vim::take_forced_motion(cx);
         self.exit_temporary_normal(window, cx);
-        self.update_editor(cx, |_, editor, cx| {
-            scroll_editor(editor, mode, move_cursor, amount, window, cx)
+        self.update_editor(cx, |vim, editor, cx| {
+            scroll_editor(vim, editor, mode, move_cursor, amount, window, cx)
         });
```
dinocosta: Since `scroll_editor` is now taking a `&mut Vim` parameter, we can probably just update it to be a method in `Vim` 👍

crates/vim/src/visual.rs
```diff
@@ -1561,6 +1561,41 @@ mod test {
         });
     }
 
+    #[gpui::test]
+    async fn test_visual_block_insert_after_ctrl_d_scroll(cx: &mut gpui::TestAppContext) {
```
dinocosta, line 1565: I think we're missing the JSON test data file for this test. After you run `cargo t -p vim test_visual_block_insert_after_ctrl_d_scroll --features=neovim` it should generate the corresponding JSON file, which also needs to be included in these changes 🙂

crates/vim/src/normal/scroll.rs
```diff
@@ -152,124 +153,133 @@ fn scroll_editor(
...
+        let goal = match amount {
+            ScrollAmount::Column(_) | ScrollAmount::PageWidth(_) => SelectionGoal::None,
+            _ => goal,
+        };
+
+        Some((new_head, goal))
+    };
+
+    if mode == Mode::VisualBlock {
+        vim.visual_block_motion(false, editor, window, cx, &mut move_cursor);
```
dinocosta: I'm not entirely sure if we want the `preserve_goal` argument to be `false` as this breaks how selections are updated when moving to a shorter line.

Here's what it currently looks like when you enter Visual Block mode, select to the end of the line and then navigate downwards to shorter lines in both Zed and NeoVim ▼

**Zed**

https://github.com/user-attachments/assets/724abb2d-efb4-4d4b-b226-27e75647b763 

**NeoVim**

https://github.com/user-attachments/assets/98d2c10d-6295-42e7-81bf-40f7da4e6e29

And here's what it looks like with the changes in this Pull Request ▼

https://github.com/user-attachments/assets/443bba10-72f7-49d5-8c7e-af980353c3b3

Notice how not only did it not scroll the editor's content, as well as clipping the selections to the smallest line length .

dinocosta (changes requested):
Hey @cppcoffee 👋

Thank you so much for finding this issue and attempting to fix it! Although the changes do fix the outlined issue, it seems to accidentally introduce a different issue, so I've left some feedback 🙂

dinocosta (approved):
Thank you @cppcoffee 🙂

I've pushed a commit with some minor refactoring but everything looks good!

PR #50825, opened 2026-03-05
agent_ui: Add rules token counts to agent thread tooltip

tsoernes:
## Summary
- add async token counting for user and project rules in the token usage tooltip
- show per-ruleset token counts while counting is in progress

## Screenshot
Rules token counts tooltip

Release Notes:

- Agent: Improved agent token usage display by also showing tokens for rules files.


bennetbo:
Hey, thanks for working on this. We recently removed the token counting code in all other places since it did not feel worth it to maintain, since we only used it for rules files. Also, it felt like this code was quite messy, e.g. re-checking if the rules changed on every frame is not the right approach, that logic should probably live in the thread. I think the maintenance burden of this feature is not worth it right now.

PR #50827, opened 2026-03-05
picker: Fix mouse overrides selected item highlight in searchable dropdown

### LINKED ISSUE #31865: UI: Mouse hover highlight overrides selected item highlight in searchable dropdown
### Summary

In a searchable dropdown (e.g. command palette or file finder), the highlight of a selected item is overridden by a mouse hover.

### Description

Steps to reproduce:
1. Open the command palette or file finder
2. Use the mouse to hover over the selected item

Expected Behavior: The selected item is still highlighted like before
Actual Behavior: The selected item's highlight becomes dimmer because the mouse hover highlight took precedence

https://github.com/user-attachments/assets/f3c96d1e-a904-455e-a7ba-0a189b9ce505

Zed: v0.188.5 (Zed) 
OS: macOS 15.5.0
Memory: 16 GiB
Architecture: x86_64

OmChillure:
### Problem
When navigating a picker (command palette, file finder, etc.) with the keyboard and then hovering with the mouse, two items would appear highlighted at the same time - one from the keyboard selection ghost_element_selected background and one from the CSS style ghost_element_hover background. This happened because GPUI's CSS hover is applied at paint time independently from the Rust selection state.

### Solution
Added a keyboard_navigated: bool flag to [Picker] to track whether the last navigation was via keyboard or mouse.

When a keyboard navigation action fires (↑, ↓, Home, End, etc.), `keyboard_navigated` is set to true. A transparent `block_mouse_except_scroll` overlay div is added on top of each list item, preventing the CSS style from firing on `ListItem` children.

When the mouse moves over any item, the overlay's `on_mouse_move` handler fires, clears keyboard_navigated, moves the selection to that item, and re-renders - removing all overlays and restoring normal mouse hover behavior.
The `on_hover` handler also moves the selection to whichever item the mouse enters, so mouse navigation is fully functional.
Clicks work through the overlay since `block_mouse_except_scroll` only affects hover hit-testing, not click events on the parent div.

### Behavior
Matches VS Code / IntelliJ — keyboard navigation suppresses hover highlights until the mouse moves, at which point normal hover resumes. Only one item is ever highlighted at a time.

### Video

[Screencast from 2026-03-05 19-20-16.webm](https://github.com/user-attachments/assets/7a8a543a-95f3-481f-b59d-171604e5f883)


OmChillure:
Hmm we will have to wait for the team to decide on this, lets see when weekends are over !!

nathansobo:
I'm going to go ahead and merge this just to get the benefits, but please offer any feedback you have, and we can always fix forward.

PR #50842, opened 2026-03-05
Fix settings file restored to both panes after restart

### LINKED ISSUE #35947: Settings file is restored to both panes after restart, closing in 1st pane closes in 2nd pane
### Summary

After moving the settings file to the 2nd pane with cmd-2 and restarting Zed, it opens in both panes. Closing it in the 1st pane actually closes it in the 2nd pane.

### Description

<details>
<summary>Screen recording</summary>

https://github.com/user-attachments/assets/8c0d2282-450f-441c-9b41-2a1436768fc7

</details>

Steps to reproduce:
1. Open the settings file
2. <kbd>cmd</kbd><kbd>2</kbd> to duplicate it to the right pane
3. Close it in the left pane
4. Restart Zed -- the settings file is now open in both left and right panes
5. Close the settings file in the left pane with a keyboard shortcut -- the right pane closes instead

**Expected Behavior**: Panes are restored correctly
**Actual Behavior**: Right pane is restored to left+right pane and closing the left pane actually closes the right pane

Zed: v0.198.5 (Zed) 
OS: macOS 15.6.0
Memory: 16 GiB
Architecture: x86_64

iam-liam:
Closes #35947

## Summary

When a settings file was moved to a second pane and Zed restarted, the file appeared in both panes. Closing it in one pane would close it in the other.

## Root cause

`Editor::deserialize` restored standalone files (like settings.json) by calling `workspace.open_abs_path()`, which routes through `open_path_preview` → `pane.open_item()`, adding the editor to the workspace's default pane. The caller (`SerializedPane::deserialize_to`) then also adds the item to the target pane, so it ends up in two panes.

This also caused the SQL constraint violations @MrSubidubi noted: the `items` table has `PRIMARY KEY(item_id, workspace_id)`, so the duplicate triggers errors on the next serialisation cycle.

## Fix

Replace `workspace.open_abs_path()` with `project.open_local_buffer()`, which opens the buffer without touching any pane. Pane placement is left to `deserialize_to`.

https://github.com/user-attachments/assets/68d3c5b4-d002-429f-b907-ec21cb0019ec

## Test plan

- [x] Reproduced the original bug (settings file duplicated across panes after restart)
- [x] Verified the fix: file restores only to the correct pane
- [x] Added regression test (`test_deserialize_non_worktree_file_does_not_add_to_pane`)
- [x] Existing `items::tests::test_deserialize` passes (all 6 cases)
- [x] `cargo clippy -p editor` clean

Release Notes:

- Fixed settings file being restored to multiple panes after restart
  ([#35947](https://github.com/zed-industries/zed/issues/35947)).

### DIFF
--- a/crates/editor/src/items.rs
+++ b/crates/editor/src/items.rs
@@ -41,6 +41,7 @@ use std::{
 use text::{BufferId, BufferSnapshot, Selection};
 use ui::{IconDecorationKind, prelude::*};
 use util::{ResultExt, TryFutureExt, paths::PathExt};
+use workspace::item::{Dedup, ItemSettings, SerializableItem, TabContentParams};
 use workspace::{
     CollaboratorId, ItemId, ItemNavHistory, ToolbarItemLocation, ViewId, Workspace, WorkspaceId,
     invalid_item_view::InvalidItemView,
@@ -51,11 +52,7 @@ use workspace::{
     },
 };
 use workspace::{
-    OpenOptions,
-    item::{Dedup, ItemSettings, SerializableItem, TabContentParams},
-};
-use workspace::{
-    OpenVisible, Pane, WorkspaceSettings,
+    Pane, WorkspaceSettings,
     item::{FollowEvent, ProjectItemKind},
     searchable::SearchOptions,
 };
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #50852, opened 2026-03-05
search: Fix replace all being silently dropped

OmChillure:
Fixes #50848

### Problem
When Replace All was triggered with a stale search query (i.e., the query text had changed since the last completed search), the old code ould restart the search and immediately return, silently discarding the replace-all intent. This caused the Replace All action to appear to do nothing on the first try, only working on subsequent attempts once results were already loaded.

### Fix :
Fix this by introducing a `pending_replace_all` flag on ProjectSearchView. When Replace All is invoked while a search is in
flight or the query is stale, the flag is set and the action is deferred.
Once the search completes and `entity_changed` is called, the flag is checked and `replace_all` is automatically dispatched.

Also disable the Replace Next button in the UI while a search is underway, since it cannot meaningfully act without up-to-date results.

### Release Notes:

- Fixed "Replace All" in project search not working on the first attempt when the search query was changed or results hadn't loaded yet.

### DIFF
--- a/crates/search/src/project_search.rs
+++ b/crates/search/src/project_search.rs
@@ -264,6 +264,7 @@ pub struct ProjectSearchView {
     excluded_files_editor: Entity<Editor>,
     filters_enabled: bool,
     replace_enabled: bool,
+    pending_replace_all: bool,
     included_opened_only: bool,
     regex_language: Option<Arc<Language>>,
     _subscriptions: Vec<Subscription>,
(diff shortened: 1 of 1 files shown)


crates/search/src/project_search.rs
```diff
@@ -762,17 +766,28 @@ impl ProjectSearchView {
...
+            if self.entity.read(cx).pending_search.is_none() {
+                self.pending_replace_all = false;
+            }
             return;
         }
         if self.active_match_index.is_none() {
+            self.pending_replace_all = false;
             return;
         }
+        self.pending_replace_all = false;
```
osiewicz: We can hoist this assignment out in front of `if` statement to reduce duplication.

osiewicz (approved):
LGTM, sans a small nit

crates/search/src/project_search.rs
OmChillure: hey, I have fixed it and pushed.

thank you

PR #50861, opened 2026-03-05
lsp: Implement didFocus RPC message when a file gets focused

0PandaDEV:
Closes https://github.com/zed-industries/zed/discussions/40999

Release Notes:

- [feat: implement didFocus RPC message when a file gets focused](https://github.com/zed-industries/zed/pull/50861/commits/05d1076a6ae019f2829507a318ddeb9c6ba7c0f7)

For some reason it doesn't trigger the didFocus for some file. I'm not sure what is causing this currently it applies to `README.md` and `.gitignore`.


SomeoneToIgnore (commented):
Is this some extension of the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#version_3_17_0) that's not used anywhere but a single case?

I do not think we're interested in having anything like this before refactoring our LSP extensions-related code.

osiewicz (changes requested):
I think OP is referring to discussion over at https://github.com/zed-industries/zed/discussions/40999, which refers to extension API's `didOpen`, `didClose`, `didChange`.. Except that that's LSP, not Zed's extension API. These are not sufficient to draw conclusions from, because LSP's explicit goal is to not couple itself to the presentation layer (which files are visible and what not).

I feel strongly about not merging this in the current form. `didFocus` is not a part of the spec and while we do have `didFocus` support for copilot, that is because copilot documents that as a necessity for it's edit prediction features + it's a core service provider for Zed. Adding a protocol extension just because feels quite wrong.

PR #50900, opened 2026-03-06
Fix sidebar selections

mikayla-maki:
Release Notes:

- N/A


crates/sidebar/src/sidebar.rs
```diff
@@ -157,34 +177,21 @@ fn workspace_path_list_and_label(
...
     contents: SidebarContents,
+    /// The index of the list item that currently has the keyboard focus
+    ///
+    /// Note: This is NOT the same as the active item.
     selection: Option<usize>,
+    focused_thread: Option<acp::SessionId>,
+    active_entry_index: Option<usize>,
     collapsed_groups: HashSet<PathList>,
     expanded_groups: HashSet<PathList>,
-    _subscriptions: Vec<Subscription>,
```
mikayla-maki, line 184: I removed all of these subscriptions by moving the logic to follow a simple rule: Only create a single observation callback, that is bound on initialization, and lives for the length of the object. By doing that, we can `.detach()` them all.

PR #50901, opened 2026-03-06
docs: Allow navigating search results with arrow keys

### LINKED ISSUE #50604: docs: Search bar results view doesn't scroll with arrow keys
### Reproduction steps

1. Go to https://zed.dev/docs/
2. Use the search bar on top
3. Use the down Arrow Key to see more search results

https://github.com/user-attachments/assets/d7855c2c-e0c0-493f-a83b-98da53049687

Current behavior: 
focus entry does iterate with arrow keys but pane stays still
this creates more friction for keyboard driver users as they now have to manually use the scrollbar

Expected behavior: 
pane scrolls after the focused entries is not in the current view
caveat: pane doesn't scroll on every arrow key movement.

Web: Chromium and Firefox

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

implabinash:
Closes #50604

https://github.com/user-attachments/assets/8a85b39e-e463-4fee-bc1f-2a03fe193ce3

https://github.com/user-attachments/assets/8290bb06-1eaf-4852-9568-97654e30211e

### Results: 
Now, you can scroll the search results with the arrow keys.

### Suggestion:
While this is working, I like to propose some better UX.

When you scroll with the arrow keys, the whole search results are scrolling along with the search bar. But, I think it would be better if we keep the search bar fixed on top and only the results scroll. Here is an example:

https://github.com/user-attachments/assets/af9dce73-27c6-407b-8180-2f771a85303b

If you think this will be better, please let me know, and I will update this PR.

Thank you.

Release Notes:

- Fixed docs search results scrolling with arrow keys

### DIFF
--- a/docs/theme/css/chrome.css
+++ b/docs/theme/css/chrome.css
@@ -368,7 +368,10 @@ mark.fade-out {
 .searchbar-outer {
   margin-inline-start: auto;
   margin-inline-end: auto;
+  width: 100%;
   max-width: var(--content-max-width);
+  box-sizing: border-box;
+  padding: 16px;
 }
 
 #searchbar {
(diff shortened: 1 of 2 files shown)


implabinash:
@cla-bot check

implabinash:
I updated the search bar style to fix it at the top while scrolling.

PR #50903, opened 2026-03-06
Preserve built-in Bedrock model routing for metadata overrides

cal-gooo:
## Summary
- preserve built-in Bedrock model routing when `language_models.bedrock.available_models` overrides a built-in model
- let Bedrock metadata overrides such as `max_tokens` and `display_name` affect the token meter and warning UI without breaking region routing or extended-context behavior
- keep custom Bedrock models working as before when the model id is not built in

## Why
Before this change, adding a built-in Bedrock model like `claude-sonnet-4-5` to `language_models.bedrock.available_models` replaced the built-in model with `Model::Custom`.

That broke Bedrock-specific behavior for those models because `Model::Custom` bypasses the built-in cross-region inference id mapping and uses the raw model name directly. For Bedrock-hosted Claude models, that meant overrides intended only for UI metadata such as `max_tokens` could cause requests to fail with region support / invalid model identifier errors. It also meant `allow_extended_context` no longer followed the built-in model behavior.

## How
This change keeps built-in Bedrock models as their original enum variants and applies matching `available_models` entries as metadata overrides instead of converting them into `Custom` models.

That preserves:
- Bedrock cross-region inference id mapping
- region support behavior
- extended-context support for built-in models

while still allowing overridden metadata to flow through to the UI and request configuration where appropriate.

## Testing
- cargo fmt --package language_models
- cargo test -p language_models test_available_model_override_preserves_builtin_bedrock_model
- cargo test -p language_models test_available_model_override_adds_custom_bedrock_model_when_missing

### DIFF
--- a/crates/language_models/src/provider/bedrock.rs
+++ b/crates/language_models/src/provider/bedrock.rs
@@ -417,10 +417,57 @@ impl BedrockLanguageModelProvider {
         }
     }
 
-    fn create_language_model(&self, model: bedrock::Model) -> Arc<dyn LanguageModel> {
+    fn apply_available_model_overrides(
+        models: &mut BTreeMap<String, bedrock::Model>,
+        available_models: &[AvailableModel],
+    ) {
+        for model in available_models {
+            if models.contains_key(&model.name) {
+                continue;
+            }
+
+            models.insert(
+                model.name.clone(),
+                bedrock::Model::Custom {
+                    name: model.name.clone(),
+                    display_name: model.display_name.clone(),
+                    max_tokens: model.max_tokens,
+                    max_output_tokens: model.max_output_tokens,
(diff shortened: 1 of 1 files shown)


cal-gooo:
@cla-bot check

cal-gooo:
@cla-bot check

bennetbo:
Hey, thanks for working on this. Can you clarify what specifically you are trying to fix? Sounds like the issue is that `cross_region_inference_id` is not working how you'd like it to for custom models?

> That broke Bedrock-specific behavior for those models because Model::Custom bypasses the built-in cross-region inference id mapping and uses the raw model name directly.

Can't you just specify the region specific name directly?

Let me know if I am missing something. For now I'll close this.

PR #50918, opened 2026-03-06
vim: Reset cursor after deleting keys in a multi-key keybinding

### LINKED ISSUE #41744: Helix: Some weird behaviour
### Summary

I’m not sure why this happens — it’s quite strange....
When my cursor is positioned at the final `EMPTY` character, I press `j j` to switch to Helix_Normal mode and then press `o` to insert a new line below. However, the cursor doesn’t move to the new line. If I use Esc instead of `j j`, this issue doesn’t occur.

### Description
Custom KeyBind, like this:

```json
[
  {
    "bindings": {
      "j j": "vim::SwitchToHelixNormalMode"
    }
  }
]
```

This causes unexpected behavior.

DESCRIPTION_HERE

Steps to reproduce:
1. I wrote the following test:
```Rust
    #[gpui::test]
    async fn test_insert_line(cx: &mut gpui::TestAppContext) {
        cx.enable_helix();

        cx.update(|_, cx| {
            cx.bind_keys([KeyBinding::new("j j", SwitchToHelixNormalMode, None)]);
        });

        cx.set_state(
            indoc! {
            "hello world!ˇ"},
            Mode::Insert,
        );
        cx.simulate_keystrokes("j j");
        cx.simulate_keystrokes("o");
        let state = cx.editor_state();

        cx.set_state(
            indoc! {
            "hello world!ˇ"},
            Mode::HelixNormal,
        );
        cx.simulate_keystrokes("o");
        assert_eq!(state, cx.editor_state());
    }
```
2. run it, and it failed with this: 
```
---- helix::test::test_insert_line stdout ----

thread 'helix::test::test_insert_line' panicked at crates\vim\src\helix.rs:1479:9:
assertion `left == right` failed
  left: "hello world!ˇ\n"
 right: "hello world!\nˇ"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    helix::test::test_insert_line
```
3. This is unexpected...

**Expected Behavior**:
The test should pass.
**Actual Behavior**:
The test fails unexpectedly.

Zed: v0.210.4 (Zed) 
OS: Windows 10.0.22621
Memory: 31.8 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 4080 Laptop GPU || NVIDIA Corporation || 581.15 r581_07

polyesterswing:
Closes #41744 

Need some advice here, but it looks like the cursor was not being properly reset after keys are deleted in a keybinding involving multiple keys

The test from @HuaGu-Dragon  passes with this fix

- [X] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Fixed an issue where the cursor was improperly reset after deleting keys in a multi-key keybinding


crates/vim/src/helix.rs
```diff
@@ -860,6 +860,7 @@ impl Vim {
 #[cfg(test)]
 mod test {
     use gpui::{UpdateGlobal, VisualTestContext};
+    use gpui::KeyBinding;
```
dinocosta: I believe we usually try to keep these grouped 🙂


```suggestion
use gpui::{KeyBinding, UpdateGlobal, VisualTestContext};
```

crates/vim/src/helix.rs
```diff
@@ -869,6 +870,34 @@ mod test {
     use workspace::{DeploySearch, MultiWorkspace};
 
     use crate::{VimAddon, state::Mode, test::VimTestContext};
+    use crate::SwitchToHelixNormalMode;
```
dinocosta: Same thing about grouping imports 🙂


```suggestion
use crate::{SwitchToHelixNormalMode, VimAddon, state::Mode, test::VimTestContext};
```

crates/vim/src/helix.rs
```diff
@@ -869,6 +870,34 @@ mod test {
...
+        cx.simulate_keystrokes("o");
+        let state = cx.editor_state();
+
+        cx.set_state(
+            indoc! {
+            "hello world!ˇ"},
+            Mode::HelixNormal,
+        );
+        cx.simulate_keystrokes("o");
+        assert_eq!(state, cx.editor_state());
```
dinocosta: Although not a practical difference, I guess this might read better if we first obtain the state of the "normal" path, where the editor is in normal mode and `o` is used to insert the newline, and then assert that using `j j` followed by `o` leads to the same result, like so:

```
cx.set_state("hello world!ˇ", Mode::Normal);
cx.simulate_keystrokes("o");
let state = cx.editor_state();

cx.set_state("hello world!ˇ", Mode::Insert);
cx.simulate_keystrokes("j j");
cx.simulate_keystrokes("o");
assert_eq!(cx.editor_state(), state);
```

dinocosta (commented):
Hey @polyesterswing 👋 

Thank you so much picking this one up! I've confirmed that your changes fix the reported issue, will still take a look at the change and understand if it does make sense to be there, but I've already left some comments 🙂

crates/vim/src/helix.rs
polyesterswing: fixed!

dinocosta (approved):
Thank you for tackling the suggested changes @polyesterswing ! 😀

ConradIrwin:
Thanks for this; and sorry for the very slow response.

Although I can imagine that this does fix the problem, it seems to me to be at the wrong layer of abstraction.

I'm guessing this issue is that when `jj` happens, the switch to normal mode happens concurrently with the pending `j` being removed from the document.

I think we should handle this case explicitly in the switch to normal mode, not implicitly in the selections code.

PR #50925, opened 2026-03-06
fs: Retry FS ops with elevation on PermissionDenied

### LINKED ISSUE #22179: Remote development as sudo user
### Check for existing issues

- [X] Completed

### Describe the feature

When I open a remote folder that can be modified only with sudo rights, any remote development ends with `Permission denied (os error 13)`. I want to somehow specify in ssh connection that user has sudo rights and hence zed should use is. I'm thinking of something like this:
```json
{
  "ssh_connections": [
    {
      "host": "00.00.00.00",
      "username": "my_username",
      "projects": [
        {
          "paths": [
            "~/protected_folder"
          ]
        }
      ],
      "sudo": true
    },
  ]
}
```

Zed: v0.165.4 (Zed)
OS: macOS 15.2.0
Memory: 48 GiB
Architecture: aarch64

_No response_

MostlyKIGuess:
On Linux/FreeBSD, FS operations that return PermissionDenied are retried by invoking the command via pkexec, sudo -A (askpass), or plain sudo as a last resort.

The pipeline ideation is taken from https://github.com/microsoft/vscode-sudo-prompt 

Closes #22179 

Release Notes:

- Allows user with sudo access to modify file without permission error.


MostlyKIGuess:
Wasn't sure if this aligns with the UI change, happy to iterate more on that.

MostlyKIGuess:
> @MostlyKIGuess Thanks for this!

 
Hey! Thanks for the review. I wanted to ask few things before moving into refactoring.






> The overall feature makes sense, but it needs to be at the UI layer, not the FS layer. We should not do this on auto_save, or in cases that zed is doing stuff like reading/writing settings (or like any other use of the fs apis except user hitting ctrl-s) to avoid randomly showing the dialogue.

Yeah, I read the save_item in pane and the conflict messages, it makes sense to have that here as well.

> What I'd suggest is in `fn save_item` - if the save fails with a permission error, prompt the user to explain the problem and ask if they want to retry with elevated permissions. If they want to, then prompt for sudo.

Should elevation also cover create_dir, remove_file and rename? I added it because I am assuming that if a user wants to edit an elevated file, they might as well want to create/delete it as well. So would it make sense to have a single save_with_elevation path that handles the full sequence (create dirs + write), and also handle elevation for explicit create/delete actions triggered by the user.

> Looking briefly at the sudo code; I'm not super comfortable requiring all callers to remember to escape. Maybe the run_cmd_with_sudo could take a command and args vec and do the escaping itself.

yeah, injection bugs could happen, I will follow the codebase practice here.


> Is there a similar mechanism we can use on macOS (if not, happy to merge just for linux to start with, but I'd like this feature too!)

I asked an LLM and looked up a bit, what I found is that `osascript -e 'do shell script "..." with administrator privileges'` should work and gives a native auth dialog. However I don't have a Mac to test it, so I'd prefer to keep macOS in a follow-up PR rather than merge untested code. Happy to leave a stub or a TODO comment for it here.

ConradIrwin:
Sorry for the slow response here. 

I spent a bit of time trying to hammer this into shape, but I think it needs solving in a way that integrates better with the existing code paths (i.e. it should fail gracefully on remote projects; re-use the existing encoding logic etc.)

I think the right path is probably to add save_with_elevated_permissions to the Item trait and implement it; re-use the existing save logic's encoding handling when writing the tmp file.

At the very least, the fs methods need to be part of the trait so they can be stubbed out in tests, and not free floating functions.

Closing for now, but open to reviewing future similar PRs

PR #50931, opened 2026-03-06
Refresh LLM API token on organization change

NeelChotai:
Emit client-side organization changed events through `RefreshLlmTokenListener` so it produces the same `RefreshLlmTokenEvent` used for server-pushed `UserUpdated` messages.

This keeps token refresh fan-out in one place.

Closes CLO-383.

Release Notes:

- N/A


crates/client/src/user.rs
```diff
@@ -694,8 +695,21 @@ impl UserStore {
...
     }
 
-    pub fn set_current_organization(&mut self, organization: Arc<Organization>) {
-        self.current_organization.replace(organization);
+    pub fn set_current_organization(
+        &mut self,
+        organization: Arc<Organization>,
+        cx: &mut Context<Self>,
+    ) {
+        if !self
```
maxdeviant: Think this might be a bit more intelligible if we introduced a local:

```rs
    let is_same_organization = self
        .current_organization
        .as_ref()
        .is_some_and(|current| current.id == organization.id);
    if !is_same_organization {
        // 
    }
```

maxdeviant (approved):

PR #50941, opened 2026-03-06
workspace: Hide "View AI Settings" when AI is disabled

OmChillure:
Fixes #50835

### Problem :
The "View AI Settings" button on the Welcome page was always rendered regardless of the disable_ai setting. This made it visible (and non-functional) for users who had AI disabled, which was confusing.

### Fix :
- Adds an optional visibility: Option<fn(&App) -> bool> predicate field to SectionEntry
- At render time, Section::render uses filter_map to skip entries whose predicate returns false. 
- The "View AI Settings" entry is given a predicate that checks !DisableAiSettings::get_global(cx).disable_ai, matching the same pattern used in `title_bar.rs` and `quick_action_bar.rs`.
- All other entries have visibility: None, meaning they are always shown — no behaviour change for them.

### Video :
[Screencast from 2026-03-06 20-18-43.webm](https://github.com/user-attachments/assets/cbfab423-3ef3-41dd-a9ab-cbae055eef6e)

Release Notes:

- Fixed the "View AI Settings" button being visible on the Welcome page despite AI features being disabled in settings.


crates/workspace/src/welcome.rs
```diff
@@ -125,17 +127,26 @@ struct SectionEntry {
     icon: IconName,
     title: &'static str,
     action: &'static dyn Action,
+    visibility: Option<fn(&App) -> bool>,
```
MrSubidubi: I think something like `visibility_guard` might be more fitting here? Otherwise, I find the name `visibility` a bit confusing in the context otherwise.

We could also intruduce an enum like
```rust
enum SectionVisibility {
    Always,
   Conditional(fn(&App) -> bool)
}
```

crates/workspace/src/welcome.rs
```diff
@@ -125,17 +127,26 @@ struct SectionEntry {
...
+    fn render(
+        &self,
+        button_index: usize,
+        focus: &FocusHandle,
+        cx: &App,
+    ) -> Option<impl IntoElement> {
+        if self.visibility.is_some_and(|is_visible| !is_visible(cx)) {
+            return None;
+        }
+        Some(SectionButton::new(
```
MrSubidubi: Let's not return early here and instead do something like 
```suggestion
        self.visibility.is_none_or(|is_visible| is_visible(cx))
            .then(|| SectionButton::new(
```

obviously this somewhat depends on the changes above

MrSubidubi (changes requested):
Makes sense and looks good overall! Two minor nits and we should be good

MrSubidubi (approved):
Yup, this looks much better now. Thanks!

PR #50978, opened 2026-03-07
repl: Correctly discover user's kernels

raideno:
Closes #50977

Before:
image

After:
image

Release Notes:

- Fixed kernels discovery in python notebook editor.


crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -259,6 +262,31 @@ impl NotebookEditor {
...
+
+        cx.spawn_in(window, async move |this, cx| {
+            refresh_task.await.ok();
+
+            this.update_in(cx, |this, window, cx| {
+                // Only re-launch if the user hasn't explicitly chosen a kernel yet,
+                // so we can upgrade from the initial python3 fallback to the
+                // discovered .venv or recommended kernel.
+                if !this.kernel_explicitly_chosen {
+                    this.launch_kernel(window, cx);
```
rgbkrk: Needing to pull the kernelspecs based on the environments in the toolchain makes sense.

However, I don't think we want launching to be a side effect of this function. The name of the function doesn't really tell that story.

raideno: Thank you for the review, so I should just remove the kernel launching part from the `refresh_kernelspecs` which makes sense and not handle the kernel launching ?

rgbkrk: I feel like the user should explicitly launch for the moment.

crates/repl/src/notebook/notebook_ui.rs
MostlyKIGuess: While you are at it, we should keep:
```rs
            .worktree_for_id(self.worktree_id, cx)
```

rather than 
```rs
            .worktrees(cx)
            .next()
```

raideno: Okey, done in the last commit.Okey, done in the last commit fe80742f9ca08c3e6128736f7937d17f5225aa2c.

crates/repl/src/notebook/notebook_ui.rs
raideno: Okey, done in the last commit fe80742f9ca08c3e6128736f7937d17f5225aa2c.

raideno:
@rgbkrk please, could you review the changes whenever you have time ? thanks a lot.

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -494,6 +516,7 @@ impl NotebookEditor {
         }
 
         self.execution_requests.clear();
+        self.kernel_explicitly_chosen = true;
```
rgbkrk: We can delete `kernel_explicitly_chosen` entirely now right?

Other than that, looks good to go.

raideno: Forgot it yes mb, just removed it in c6ee56f.

MrSubidubi (approved):
Nice, thanks! And congratulations to your first contribution!

PR #51000, opened 2026-03-07
git_ui: Add file and folder icons to the Git panel

kickbelldev:
Closes https://github.com/zed-industries/zed/discussions/49740

Adds optional file and folder icons to the Git panel so its file list is easier to scan, especially in larger repositories.

- add folder icons for Git panel entries
- add Git panel settings for file and folder icon visibility

---

Release Notes:

  - Made the Git Panel aware of icon themes.
  - Added the ability to render file type icons in the Git panel.


danilo-leal (approved):
Thank you! Note that I turned file icons off by default as I think it gets pretty noisy with it over there. But you can always turn it on if you want it 👍

PR #51019, opened 2026-03-07
agent: Fix external agents starting when AI is disabled

oliverbarnes:
Closes #48274

Slimmed-down version of https://github.com/zed-industries/zed/pull/48637 cc @bennetbo 

When `disable_ai: true` is set, external agent servers (Claude Code ACP, Codex, Gemini, etc.) were continuing to run because the Agent Panel remained accessible and active threads weren't terminated.

This change:

- Adds `enabled()` guards to `Workspace::toggle_panel_focus` and `Workspace::open_panel` so disabled panels cannot be opened via keyboard shortcuts or programmatically
- Adds a settings observer in `AgentPanel` that reacts to `AgentSettings::enabled(cx)` changes: when AI becomes disabled, it drops active thread views (triggering process termination via `AcpConnection::drop()`) and emits `PanelEvent::Close` to close the dock

The panel's existing `enabled()` method already incorporates the `disable_ai` check via `AgentSettings::enabled(cx)`, and `icon()` already returns `None` when disabled (hiding the dock button). This change closes the remaining gaps: the keyboard shortcut and active thread cleanup.

Release Notes:
- Fixed `disable_ai` not stopping external agent servers


crates/agent_ui/src/agent_connection_store.rs
```diff
@@ -134,13 +144,27 @@ impl AgentConnectionStore {
...
+    }
+
     fn handle_agent_servers_updated(
         &mut self,
         store: Entity<AgentServerStore>,
         _: &AgentServersUpdated,
         cx: &mut Context<Self>,
     ) {
         let store = store.read(cx);
+        // This retain is sound because the tasks spawned in `request_connection` hold only
```
bennetbo: This comments seems like noise

crates/agent_ui/src/agent_connection_store.rs
```diff
@@ -88,41 +88,51 @@ impl AgentConnectionStore {
 
             self.entries.insert(key.clone(), entry.clone());
 
+            // Both tasks below capture only a WeakEntity, so the entry's strong-ref count
```
bennetbo: This comments seems like noise

crates/agent_ui/src/agent_connection_store.rs
```diff
@@ -134,13 +144,27 @@ impl AgentConnectionStore {
         })
     }
 
+    /// Removes all external (non-native) agent entries from the store.
+    ///
+    /// Because the tasks spawned in `request_connection` hold only `WeakEntity` refs,
```
bennetbo: All of these comments are just the internal monologue of the agent. They do not actually provide any value for the person trying to understand the function.

Okay with keeping this portion:
> Removes all external (non-native) agent entries from the store.

bennetbo (changes requested):

crates/agent_ui/src/agent_connection_store.rs
oliverbarnes: They were helpful to me at the time of implementation to make sense of what was happening, but I see what you mean - they don't really document anything for the future. Removed most comments and tried to keep just the ones that might be useful for someone else.

PR #51020, opened 2026-03-07
 editor: Add Ctrl+scroll wheel zoom for buffer font size

dmitry-soluyanov:
When Ctrl (or Cmd on macOS) is held, scrolling over the editor changes the buffer font size instead of scrolling: scroll up to increase, scroll
  down to decrease. Uses the same non-persistent zoom as the Ctrl+/Ctrl- shortcuts.

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -7644,6 +7644,31 @@ impl EditorElement {
                 .max(0.01);
 
             move |event: &ScrollWheelEvent, phase, window, cx| {
+                // Ctrl+scroll (Cmd+scroll on macOS) zooms the buffer font size.
+                if event.modifiers.control || event.modifiers.platform {
+                    if phase == DispatchPhase::Bubble && hitbox.should_handle_scroll(window) {
+                        let delta_y = match event.delta {
+                            ScrollDelta::Pixels(pixels) => pixels.y.into(),
+                            ScrollDelta::Lines(lines) => lines.y,
+                        };
+                        if delta_y > 0.0 {
+                            window.dispatch_action(
+                                Box::new(zed_actions::IncreaseBufferFontSize { persist: false }),
+                                cx,
+                            );
+                        } else if delta_y < 0.0 {
+                            window.dispatch_action(
+                                Box::new(zed_actions::DecreaseBufferFontSize { persist: false }),
+                                cx,
+                            );
+                        }
+                        if delta_y != 0.0 {
+                            cx.stop_propagation();
+                            return;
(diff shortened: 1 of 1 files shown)


crates/editor/src/element.rs
```diff
@@ -7644,6 +7644,31 @@ impl EditorElement {
                 .max(0.01);
 
             move |event: &ScrollWheelEvent, phase, window, cx| {
+                // Ctrl+scroll (Cmd+scroll on macOS) zooms the buffer font size.
+                if event.modifiers.control || event.modifiers.platform {
```
SomeoneToIgnore, line 7648: We have to be a bit more precise here, as `platform` may mean `Win` or `Super` key on other platforms — we do not want this.

There's a good `pub fn secondary` method in the same `Modifiers` struct that does something similar as an example.

This hopefully allows to remove the new comment line from above too, as it's rather redundant.

crates/editor/src/element.rs
```diff
@@ -7644,6 +7644,31 @@ impl EditorElement {
...
+                                Box::new(zed_actions::IncreaseBufferFontSize { persist: false }),
+                                cx,
+                            );
+                        } else if delta_y < 0.0 {
+                            window.dispatch_action(
+                                Box::new(zed_actions::DecreaseBufferFontSize { persist: false }),
+                                cx,
+                            );
+                        }
+                        if delta_y != 0.0 {
```
SomeoneToIgnore, line 7665: Can you elaborate on this branch?
Why do we stop the even propagation on this case only?

crates/editor/src/element.rs
```diff
@@ -7644,6 +7644,31 @@ impl EditorElement {
                 .max(0.01);
 
             move |event: &ScrollWheelEvent, phase, window, cx| {
+                // Ctrl+scroll (Cmd+scroll on macOS) zooms the buffer font size.
+                if event.modifiers.control || event.modifiers.platform {
+                    if phase == DispatchPhase::Bubble && hitbox.should_handle_scroll(window) {
```
SomeoneToIgnore, line 7649: Seems that we can merge this block with another right below with the same condition?

```rs
if phase == DispatchPhase::Bubble && hitbox.should_handle_scroll(window) {
    if conditions_for_zooming {
//....
    } else {
//.... maybe even worth extracting both into their own fns?
    }
}
```

SomeoneToIgnore (changes requested):
Nice idea, have left a few comments.

When fixed further, can we get a video of how this works attached here, please?

PR #51026, opened 2026-03-07
Lazily initialize kernelspecs

reflectronic:
On Windows, the WSL VM always boots up when Zed is opened, because we eagerly discover which jupyter kernels are installed inside WSL on startup. This is not desirable if the REPL feature is not being used. Defer this work to the point where we actually need to know what kernels are installed.

Release Notes:

- N/A

### DIFF
--- a/crates/repl/src/components/kernel_options.rs
+++ b/crates/repl/src/components/kernel_options.rs
@@ -448,7 +448,9 @@ where
     TT: Fn(&mut Window, &mut App) -> AnyView + 'static,
 {
     fn render(self, window: &mut Window, cx: &mut App) -> impl IntoElement {
-        let store = ReplStore::global(cx).read(cx);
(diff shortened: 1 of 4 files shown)


crates/repl/src/components/kernel_options.rs
```diff
@@ -448,7 +448,9 @@ where
     TT: Fn(&mut Window, &mut App) -> AnyView + 'static,
 {
     fn render(self, window: &mut Window, cx: &mut App) -> impl IntoElement {
-        let store = ReplStore::global(cx).read(cx);
+        let store = ReplStore::global(cx);
+        store.update(cx, |store, cx| store.ensure_kernelspecs(cx));
```
rgbkrk, line 452: 👍

rgbkrk (approved):

PR #51034, opened 2026-03-08
languages: Validate pylsp binary before returning from check_if_user_installed

rgbkrk:
Run `pylsp --version` via `delegate.try_exec()` in both branches of `PyLspAdapter::check_if_user_installed` before returning the binary. If execution fails (broken shebang, missing interpreter, etc.), log a warning and return None so the system falls through gracefully instead of surfacing an error dialog.

This matches the existing validation pattern used by TyLspAdapter and RuffLspAdapter in their `fetch_server_binary` implementations.

No idea if this closes an issue but it sure was annoying on a system where I deleted the pylsp environment I had. It surprised me too since I had pylsp disabled in settings.

Release Notes:

- Fixed detection of when `pylsp` is not installed properly on a user's system so that it doesn't get launched as an LSP when it doesn't exist.


Veykril (approved):

Veykril:
Thanks!

Issue #38109, opened 2025-09-13
Zed out of sync with changes made outside of editor


binarypie:
### Summary

Adding, removing, or renaming files and folders outside of Zed will not automatically update the project panel.

### Description
When working with complex projects where files will change outside of Zed the Zed editor falls out of sync with the system state. Specifically, open buffers and the project panel, and the git panel. Likely anything inside of Zed that relies on a cache of the current filesystem state. 

### Steps to reproduce:
1. Start with a deep folder structure

> 
>  * Apps
>    * app1
>    * app2
>    * app3
>  * Packages
>    * package 1
>    * package 2
>      * protos
>        * service/v1/
>      * src
>    * package 3

2. Have AI create some protos in `package2/protos/service/v1`
3. Open some of these protos in Zed
4. Download some actual protos that replace some of those files
5. Download some actual protos that add new protos to that folder
6. Open a terminal and cd into the directory to verify they are there
7. Notice the open buffer did not update
8. Notice the project file list did not update

**Expected Behavior**:
* Open buffers should update when files are replaced on the system.
* Project folder structure should be in sync when making changes outside of Zed
* Git panel should always show the current state as if one typed git status in their terminal.

**Actual Behavior**:
* Open buffers display the previous version of the file downloaded.
* Git panel is completely ignorant of any changes even after doing `git add .`
* Project panel still show the old state

### Errors
<details><summary>LSP Failed to load workspaces. (click to expand)</summary>

```
2025-09-12T22:07:22-07:00 INFO  [project.lsp_format] Formatting via LSP
2025-09-12T22:07:22-07:00 ERROR [project::lsp_store] Formatting failed: failed to format via language server: Failed to spawn cd "/Code/packages/rust/services" && "rustfmt"
2025-09-12T22:07:22-07:00 ERROR [acp_thread] failed to format via language server

Caused by:
    Failed to spawn cd "/Code/packages/rust/services" && "rustfmt"
2025-09-12T22:07:37-07:00 INFO  [project.lsp_format] Formatting via LSP
2025-09-12T22:07:37-07:00 ERROR [project::lsp_store] Formatting failed: failed to format via language server: Failed to spawn cd "/Code/packages/rust/services/src/openfga" && "rustfmt"
2025-09-12T22:07:37-07:00 ERROR [acp_thread] failed to format via language server

```
</details>

### Zed Version and System Specs

Zed: v0.202.8 (Zed) 
OS: Linux Wayland bluefin 42
[...]

PR #51036, opened 2026-03-08
fs: Map EventKind::Other to Changed instead of None

lex00:
Closes #48694
Supersedes #48695
Related to #38109

## Problem

When the OS event queue overflows, the `notify` crate emits `EventKind::Other` as a rescan signal (e.g. macOS `MustScanSubDirs`, Linux `IN_Q_OVERFLOW`). The catch-all in `fs_watcher.rs` maps this to `None`.

Three downstream consumers silently drop events with `kind: None`:
- BackgroundScanner initial scan filters them out (worktree.rs)
- Settings file watcher ignores them (settings_file.rs)
- LSP `didChangeWatchedFiles` notifications skip them (lsp_store.rs)

The original `_ => None` predates these consumers. When it was written, `process_events()` stripped the kind and only used paths, so `None` was harmless. The settings watcher and LSP path were added later and assumed `kind` would always be set.

This is an uncommon trigger (requires event queue overflow from heavy filesystem activity) and the failure is silent, which makes it hard to link to specific user reports. But the fix is simple and the current behavior is clearly wrong.

## Fix

Map `EventKind::Other` and `EventKind::Any` to `Some(PathEventKind::Changed)`. "Something changed, re-read from disk" is the correct semantic for both variants. Other unknown variants (like `EventKind::Access`) remain mapped to `None`.

Extracted the mapping into a standalone `path_event_kind()` function with unit tests.

## Test plan

- New: `test_path_event_kind_mapping` covers all `EventKind` variants
- All existing fs tests pass
- clippy clean

Release Notes:

- Fixed an issue where filesystem events could be silently dropped after OS event queue overflow, potentially causing stale settings and missing LSP notifications.

### DIFF
--- a/crates/fs/src/fs_watcher.rs
+++ b/crates/fs/src/fs_watcher.rs
@@ -91,12 +91,7 @@ impl Watcher for FsWatcher {
             |g| {
                 g.add(path, mode, move |event: &notify::Event| {
                     log::trace!("watcher received event: {event:?}");
-                    let kind = match event.kind {
-                        EventKind::Create(_) => Some(PathEventKind::Created),
-                        EventKind::Modify(_) => Some(PathEventKind::Changed),
-                        EventKind::Remove(_) => Some(PathEventKind::Removed),
-                        _ => None,
-                    };
+                    let kind = path_event_kind(&event.kind);
                     let mut path_events = event
                         .paths
                         .iter()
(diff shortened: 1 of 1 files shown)


probably-neb:
Hey @lex00, thanks for recognizing this and following up on it. I'm going to close this in favor of https://github.com/zed-industries/zed/pull/51208, naively mapping the rescan events to changed events is not a complete solution as far as I can tell, and from my testing may introduce new issues.

PR #51042, opened 2026-03-08
Restore untitled buffers after closing with red button

### LINKED ISSUE #48468: "Restore Unsaved Buffers" is not respected when closing the window on macOS
### Reproduction steps

1. Open Zed with empty session
2. Type something into the editor
3. Close the window using the red button on macOS
4. Reopen Zed app

I am grateful that [long awaited feature](https://github.com/zed-industries/zed/pull/46557) has been released, but i found a little annoying bug that probably should be easy to fix.

## Current behavior

Zed reopens with empty editor — no changes are saved

## Expected behavior

Zed reopens with last changes. 
This works as expected when quitting the App directly (CMD + Q on Mac)

OS: macOS 26.2
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

```log
2026-02-05T12:31:06+01:00 INFO  [zed] ========== starting zed version 0.222.2+stable.142.5baf53428483a23e8e253ebf65fff4717a4b5464, sha 5baf534 ==========
2026-02-05T12:31:06+01:00 INFO  [crashes] spawning crash handler process
2026-02-05T12:31:06+01:00 INFO  [zed] Using git binary path: "/Applications/Zed.app/Contents/MacOS/git"
2026-02-05T12:31:07+01:00 INFO  [crashes] connected to crash handler process after 100ms
2026-02-05T12:31:07+01:00 INFO  [crashes] crash handler registered
2026-02-05T12:31:07+01:00 INFO  [extension_host] extensions updated. loading 4, reloading 0, unloading 0
2026-02-05T12:31:07+01:00 INFO  [auto_update] Auto Update: checking for updates
2026-02-05T12:31:07+01:00 INFO  [workspace] Rendered first frame
2026-02-05T12:31:07+01:00 INFO  [util] set environment variables
2026-02-05T12:31:58+01:00 INFO  [zed] ========== starting zed version 0.222.2+stable.142.5baf53428483a23e8e253ebf65fff4717a4b5464, sha 5baf534 ==========
2026-02-05T12:31:58+01:00` INFO  [crashes] spawning crash handler process
2026-02-05T12:31:58+01:00 INFO  [zed] Using git binary path: "/Applications/Zed.app/Contents/MacOS/git"
2026-02-05T12:31:58+01:00 INFO  [crashes] connected to crash handler process after 100ms
2026-02-05T12:31:58+01:00 INFO  [crashes] crash handler registered
2026-02-05T12:31:58+01:00 INFO  [extension_host] extensions updated. loading 4, reloading 0, unloading 0
2026-02-05T12:31:58+01:00 INFO  [auto_update] Auto Update: checking for updates
2026-02-05T12:31:58+01:00 INFO  [workspace] Rendered first frame
2026-02-05T12:31:59+01:00 INFO  [util] set environment variables
2026-02-05T12:32:52+01:00 INFO  [zed] ========== starting zed version 0.222.2+stable.142.5baf53428483a23e8e253ebf65fff4717a4b5464, sha 5baf534 ==========
2026-02-05T12:32:52+01:00 INFO  [crashes] spawning crash handler process
2026-02-05T12:32:52+01:00 INFO  [zed] Using git binary path: "/Applications/Zed.app/Contents/MacOS/git"
2026-02-05T12:32:52+01:00 INFO  [crashes] connected to crash handler process after 100ms
2026-02-05T12:32:52+01:00 INFO  [crashes] crash handler registered
2026-02-05T12:32:52+01:00 INFO  [extension_host] extensions updated. loading 4, reloading 0, unloading 0
2026-02-05T12:32:53+01:00 INFO  [auto_update] Auto Update: checking for updates
2026-02-05T12:32:53+01:00 INFO  [workspace] Rendered first frame
2026-02-05T12:32:53+01:00 INFO  [util] set environment variables
```

### Relevant Zed settings

```json
{
  "when_closing_with_no_tabs": "keep_window_open",
  "disable_ai": true,
  "buffer_font_family": "PT Mono",
  "use_system_path_prompts": true,
  "use_system_prompts": true,
  "inlay_hints": {
    "enabled": true,
  },
  "status_bar": {
    "active_language_button": true,
  },
  "autosave": "on_window_change",
  "theme": "Gruvbox Dark Hard",
  "buffer_font_features": {
    "calt": false,
  },
  "base_keymap": "VSCode",
  "telemetry": {
    "metrics": true,
  },
  "confirm_quit": false,
  "ui_font_size": 16,
  "buffer_font_size": 16,
  "session": {
    "restore_unsaved_buffers": true,
  },
  "restore_on_startup": "last_session",
  "on_last_window_closed": "platform_default",
}
```

_No response_

None

VictorCano:
## Summary

Fixes session restoration failing for untitled/unsaved buffers when closing Zed with the red button.

### Root cause

`history_manager::init()` runs on every startup and calls `recent_workspaces_on_disk()` to populate the "Open Recent" list. That function deletes any workspace where the paths no longer exist on disk. For untitled-buffer workspaces, `paths` is empty — `all_paths_exist_with_a_directory([])` returns `false` (the loop never runs, `any_dir` stays `false`), so the workspace row gets deleted before session restoration can find it.

### Fixes

**Bug 4 (root cause)** — `recent_workspaces_on_disk` deletes empty-paths workspaces:
- Skip workspaces with no paths instead of deleting them. They are not useful in the "Open Recent" list but must be preserved for session restoration.
- Added regression test `test_recent_workspaces_on_disk_does_not_delete_empty_paths_workspace`.

**Bug 3** — `flush_serialization` not called on `close_window` path:
- `close_window()` (red button) never called `flush_serialization()`, so the DB write hadn't completed before the process exited.
- Fix: await `flush_serialization()` for all workspaces before calling `remove_window()`.
- Added regression test `test_close_window_flushes_serialization_before_removal`.

**Bug 2** — `session_id` cleared by a race in `serialize_workspace_internal`:
- The `DetachFromSession` branch unconditionally called `set_session_id(None)`, which could land *after* a subsequent `save_workspace(session_id = Some(...))`, leaving `session_id = NULL`.
- Fix: only call `set_session_id(None)` when `self.session_id` is already `None` (i.e. `remove_from_session()` was explicitly called).
- Added regression test `test_detach_from_session_does_not_clear_session_id`.

**Bug 1** (macOS) — red button always clears session even when `on_last_window_closed = "quit_app"`:
- On macOS, `save_last_workspace` was hardcoded to `false`, causing `remove_from_session()` to be called on close even when closing the last window should behave like quit.
- Fix: use the same logic as Linux/Windows, but additionally require `on_last_window_closed.is_quit_app()`.

**Bonus improvements:**
- Log an error when SQLite file open fails and falls back to in-memory DB.
- `save_workspace` now explicitly deletes `center_panes` before `panes` to avoid a UNIQUE constraint failure caused by SQLite recycling pane IDs (no `AUTOINCREMENT`).
- Promote "Opening database" log from `trace` to `debug`.

## Test plan

- [ ] Open Zed dev build, create an untitled buffer with content
- [ ] Click the red × button (with `on_last_window_closed: "quit_app"` in settings)
- [ ] Relaunch — untitled buffer with content is restored ✅
- [ ] `cargo test -p workspace -- test_recent_workspaces_on_disk_does_not_delete_empty_paths_workspace` — passes
- [ ] `cargo test -p workspace -- test_detach_from_session_does_not_clear_session_id` — passes
- [ ] `cargo test -p workspace -- test_close_window_flushes_serialization_before_removal` — passes
- [ ] Full workspace + sqlez test suite: 143 passed, 0 failed

Release Notes:

- Fixed unsaved/untitled buffers not being restored after closing Zed with the red button when `restore_on_startup` is set to `last_session`.

Closes #48468.

### DIFF
--- a/crates/sqlez/src/connection.rs
+++ b/crates/sqlez/src/connection.rs
@@ -54,7 +54,14 @@ impl Connection {
     /// Attempts to open the database at uri. If it fails, a shared memory db will be opened
     /// instead.
     pub fn open_file(uri: &str) -> Self {
-        Self::open(uri, true).unwrap_or_else(|_| Self::open_memory(Some(uri)))
+        Self::open(uri, true).unwrap_or_else(|err| {
+            log::error!(
+                "open_file: failed to open '{}': {}. Falling back to in-memory DB!",
+                uri,
+                err
+            );
+            Self::open_memory(Some(uri))
+        })
     }
 
     pub fn open_memory(uri: Option<&str>) -> Self {
--- a/crates/sqlez/src/connection.rs
+++ b/crates/sqlez/src/connection.rs
@@ -55,7 +55,7 @@ impl Connection {
     /// instead.
     pub fn open_file(uri: &str) -> Self {
         Self::open(uri, true).unwrap_or_else(|err| {
-            log::error!(
+            log::info!(
                 "open_file: failed to open '{}': {}. Falling back to in-memory DB!",
                 uri,
                 err
--- a/crates/workspace/src/persistence.rs
+++ b/crates/workspace/src/persistence.rs
@@ -4697,6 +4697,136 @@ mod tests {
         assert_eq!(result[2].0, WorkspaceId(4));
     }
 
+    /// Regression test for https://github.com/zed-industries/zed/issues/48468 (Bug 2).
+    ///
+    /// When `serialize_workspace_internal` takes the `DetachFromSession` path (empty workspace,
+    /// no open items), the old code unconditionally called `set_session_id(None)`. This could
+    /// race against a later `save_workspace(session_id = Some(...))` and leave the DB with
+    /// `session_id = NULL`. The fix: only clear session_id when `self.session_id` is already
+    /// `None` (i.e. `remove_from_session` was explicitly called).
+    #[gpui::test]
+    async fn test_detach_from_session_does_not_clear_session_id(cx: &mut gpui::TestAppContext) {
+        use crate::multi_workspace::MultiWorkspace;
+        use feature_flags::FeatureFlagAppExt;
+        use project::Project;
+
+        crate::tests::init_test(cx);
+
+        cx.update(|cx| {
+            cx.set_staff(true);
+            cx.update_flags(true, vec!["agent-v2".to_string()]);
+        });
+
+        let fs = fs::FakeFs::new(cx.executor());
+        // Empty project (no paths) → workspace_location returns DetachFromSession
+        let project = Project::test(fs.clone(), [], cx).await;
+
+        let (multi_workspace, cx) =
+            cx.add_window_view(|window, cx| MultiWorkspace::test_new(project.clone(), window, cx));
+
+        // Read the session_id the Workspace was constructed with.
+        let session_id = multi_workspace.read_with(cx, |mw, cx| {
+            mw.workspace().read(cx).session_id()
+        });
+        let session_id = session_id.expect("workspace should have a session_id from Session::test");
+
+        let db = cx.update(|_, cx| WorkspaceDb::global(cx));
+
+        let workspace_id = db.next_id().await.unwrap();
+        multi_workspace.update_in(cx, |mw, _, cx| {
+            mw.workspace().update(cx, |ws, _cx| {
+                ws.set_database_id(workspace_id);
+            });
+        });
+
+        // Persist the workspace row with session_id so there is something to potentially clear.
+        db.set_session_binding(workspace_id, Some(session_id.clone()), None)
+            .await
+            .unwrap();
+
+        // flush_serialization triggers serialize_workspace_internal. Because the project has no
+        // paths and no items are open, it takes the DetachFromSession branch.
+        let task = multi_workspace.update_in(cx, |mw, window, cx| {
+            mw.workspace()
+                .update(cx, |ws, cx| ws.flush_serialization(window, cx))
+        });
+        task.await;
+
+        // The session_id must still be set. The workspace must still be discoverable for
+        // session restoration on next launch.
+        let locations = db
+            .last_session_workspace_locations(&session_id, None, fs.as_ref())
+            .await
+            .unwrap();
+        assert!(
+            locations
+                .iter()
+                .any(|sw| sw.workspace_id == workspace_id),
+            "session_id must not be cleared by DetachFromSession when not explicitly detached; \
+             workspace must remain findable for session restoration"
+        );
+    }
+
+    /// Regression test for https://github.com/zed-industries/zed/issues/48468 (Bug 3).
+    ///
+    /// `close_window` (red button) never called `flush_serialization`, so session data was
+    /// not persisted before the window was removed. The fix: await `flush_serialization` for
+    /// all workspaces before `remove_window`.
+    #[gpui::test]
+    async fn test_close_window_flushes_serialization_before_removal(
+        cx: &mut gpui::TestAppContext,
+    ) {
+        use crate::multi_workspace::MultiWorkspace;
+        use crate::CloseWindow;
+        use feature_flags::FeatureFlagAppExt;
+        use project::Project;
+
+        crate::tests::init_test(cx);
+
+        cx.update(|cx| {
+            cx.set_staff(true);
+            cx.update_flags(true, vec!["agent-v2".to_string()]);
+        });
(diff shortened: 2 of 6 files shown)


VictorCano:
> Warnings
> ⚠️	
> This PR is missing release notes.
> 
> Please add a "Release Notes" section that describes the change:
> 
> ```
> Release Notes:
> 
> - Added/Fixed/Improved ...
> ```
> 
> If your change is not user-facing, you can use "N/A" for the entry:
> 
> ```
> Release Notes:
> 
> - N/A
> ```
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [c76b5c9](https://github.com/zed-industries/zed/commit/c76b5c92b61cacb086d32f431e1a908eeebd0c54)

Done

crates/sqlez/src/connection.rs
```diff
@@ -47,7 +47,14 @@ impl Connection {
     /// Attempts to open the database at uri. If it fails, a shared memory db will be opened
     /// instead.
     pub fn open_file(uri: &str) -> Self {
-        Self::open(uri, true).unwrap_or_else(|_| Self::open_memory(Some(uri)))
+        Self::open(uri, true).unwrap_or_else(|err| {
+            log::error!(
```
eholk:
```suggestion
            log::info!(
```

I think `info` or even `debug` is better here because it's not really an error; it's a case we expect and handle. But it is nice to know when it's happening so logging makes sense to me.

crates/workspace/src/persistence.rs
```diff
@@ -1335,8 +1339,13 @@ impl WorkspaceDb {
                     }
                 };
 
-                // Clear out panes and pane_groups
+                // Clear out panes and pane_groups.
+                // We must delete center_panes explicitly before deleting panes because
+                // SQLite recycles pane_ids (no AUTOINCREMENT), causing a UNIQUE constraint
+                // failure on center_panes when the same id is reused after deletion.
+                // FK cascade from panes -> center_panes is not reliably triggered.
```
eholk: > FK cascade from panes -> center_panes is not reliably triggered.

Do we know why that's the case? It seems like we'd expect sqlite to be able to do the cascades well.

VictorCano: The schema does have `ON DELETE CASCADE` on `center_panes(pane_id)` referencing `panes(pane_id)`, and `PRAGMA foreign_keys=TRUE` is set in `CONNECTION_INITIALIZE_QUERY` (`crates/db/src/db.rs`). So the cascade should work. I've updated the comment — the explicit delete is a defensive measure: `pane_id` is `INTEGER PRIMARY KEY` without `AUTOINCREMENT`, so SQLite can reuse deleted rowids within the same transaction, and the explicit delete eliminates any risk of a UNIQUE constraint failure from a recycled pane_id collision.

echuber2:
I wonder if https://github.com/zed-industries/zed/pull/54224 might have already changed the status of Bugs 2-4 on your list. However, Bug 1 (the red button issue on macOS) is still a problem, as far as I can tell.

echuber2:
If you have time to rebase (or trim) the PR and see it through, you might want to put "Closes #48468" as the first line of your PR description and rename it to mention macOS specifically. Thanks for doing a deep dive on it also.

echuber2:
I tried splicing _just_ this section that had targeted macOS into the current main branch:

https://github.com/VictorCano/zed/blob/d9240dee8636b8b2526961afd7e0a33a763a5bde/crates/workspace/src/workspace.rs#L2864

Then, it seems to fix one case but not others:

- If I choose the setting "On Last Window Closed": "Quit App", then if I type some text into a buffer and click the red button, after relaunching Zed, the buffer with text reappears.
- However, if I have any other Zed windows open, then when I click the red button, the window with the unsaved buffer is just discarded without warning.
- If I don't choose the setting "On Last Window Closed": "Quit App", then closing that last window will lose the buffer no matter what.

Personally, I would expect that closing a window with any unsaved buffers, where the window cannot be restored as a "recent workspace", would then warn and prompt to save the buffers instead of instantly discarding them. This is the behavior in Sublime Text, for example.

SomeoneToIgnore:
Thank you for this and sorry for the slow progress: we will continue with https://github.com/zed-industries/zed/pull/60477

SomeoneToIgnore:
If you do not mind testing that other PR, that would be superb and would help to merge that in.

PR #51054, opened 2026-03-08
git_ui: Auto-fold file diff on stage in Uncommitted Changes view

ZibanPirate:
## Summary

When reviewing uncommitted changes, clicking the checkmark to stage a file leaves its diff expanded, requiring a manual collapse. This adds automatic fold/unfold behavior:

- Staging a file via the checkmark in the diff header automatically collapses its diff
- Unstaging unfolds it back so it can be reviewed again

This is controlled by a new `stage_fold_behavior` setting with three options:

| Value | Stage | Unstage |
|---|---|---|
| `collapse_and_expand` (default) | folds | unfolds |
| `collapse` | folds | no action |
| `none` | disabled | disabled |

Screenshot of the setting:

settings sc

Demo of the behavior:

auto collapse-and-expand

Settings panel:

image

Related discussion: https://github.com/zed-industries/zed/discussions/51052

## Test plan

- [ ] Open a repo with multiple changed files in the Uncommitted Changes view
- [ ] Click the checkmark to stage a file - its diff should collapse
- [ ] Click the checkmark again to unstage - its diff should expand
- [ ] Set `"stage_fold_behavior": "collapse"` - staging collapses, unstaging does nothing
- [ ] Set `"stage_fold_behavior": "none"` - no automatic fold/unfold (original behavior)
- [ ] Verify branch diff view is unaffected

Release Notes:

- Added `stage_fold_behavior` setting to auto-collapse file diffs when staging in the Uncommitted Changes view

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1000,6 +1000,12 @@
     //
     // Default: 72
     "commit_title_max_length": 72,
+    // Controls whether file diffs automatically collapse/expand when
+    // staging or unstaging via the checkmark in the diff view.
+    //
+    // Choices: collapse_and_expand, collapse, none
+    // Default: collapse_and_expand
+    "stage_fold_behavior": "collapse_and_expand",
   },
   "message_editor": {
     // Whether to automatically replace emoji shortcodes with emoji characters.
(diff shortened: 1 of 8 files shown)


ZibanPirate:
hey @cole-miller , is there anything i could do to help get this accross the line? or should i just close it in favour of other ongoing work?

cole-miller:
@ZibanPirate Thanks for opening a PR, and sorry for the delay in reviewing.

I don't think we want to add configurability in this particular area; we're working on adding support for a separate "Unstaged changes" multibuffer, which I think would serve the same need.

Grafikart:
I would love to have this feature it will mirror the behaviour of github and I think even with a "unstaged changes" multibuffer, being able to see both stage / unstaged change in the same place is interesting (especially since the work is already done in this MR).

PR #51060, opened 2026-03-08
data_table: #4 Rework column/table width API

HalavicH:
data_table: Replace column width builder API with `ColumnWidthConfig` enum

This PR consolidates the data table width configuration API from three separate builder methods (`.column_widths()`, `.resizable_columns()`, `.width()`) into a single `.width_config(ColumnWidthConfig)` call. This makes invalid state combinations unrepresentable and clarifies the two distinct width management modes.

**What changed:**

- Introduces `ColumnWidthConfig` enum with two variants:
  - `Static`: Fixed column widths, no resize handles
  - `Redistributable`: Drag-to-resize columns that redistribute space within a fixed table width
- Introduces `TableResizeBehavior` enum (`None`, `Resizable`, `MinSize(f32)`) for per-column resize policy
- Renames `TableColumnWidths` → `RedistributableColumnsState` to better reflect its purpose
- Extracts all width management logic into a new `width_management.rs` module
- Updates all callers: `csv_preview`, `git_graph`, `keymap_editor`, `edit_prediction_context_view`

```rust
pub enum ColumnWidthConfig {
    /// Static column widths (no resize handles).
    Static {
        widths: StaticColumnWidths,
        /// Controls widths of the whole table.
        table_width: Option<DefiniteLength>,
    },
    /// Redistributable columns — dragging redistributes the fixed available space
    /// among columns without changing the overall table width.
    Redistributable {
        entity: Entity<RedistributableColumnsState>,
        table_width: Option<DefiniteLength>,
    },
}
```

**Why:**

The old API allowed callers to combine methods incorrectly. The new enum-based design enforces correct usage at compile time and provides a clearer path for adding independently resizable columns in PR #3.

**Context:**

This is part 2 of a 3-PR series improving data table column width handling:
1. [#51059](https://github.com/zed-industries/zed/pull/51059) - Extract modules into separate files (mechanical change)
2. **This PR**: Introduce width config enum for redistributable column widths (API rework)
3. Implement independently resizable column widths (new feature)

The series builds on previously merged infrastructure:
- [#46341](https://github.com/zed-industries/zed/pull/46341) - Data table dynamic column support
- [#46190](https://github.com/zed-industries/zed/pull/46190) - Variable row height mode for data tables

Primary beneficiary: CSV preview feature ([#48207](https://github.com/zed-industries/zed/pull/48207))

### Anthony's note

This PR also fixes the table dividers being a couple pixels off, and the csv preview from having double line rendering for a single column in some cases. 

Release Notes:

- N/A


crates/ui/src/components/data_table/width_management.rs
Anthony-Eid, line 1: I would rather prevent separating `data_table.rs` into two files, it's inconsistent with the rest of the Zed codebase for the most part, and it would make finding the resizable columns stuff harder for devs that are used to our current organization patterns

Cargo.toml
Anthony-Eid, line 869: What's the motivation behind this change? `Cargo.toml` changes like this should be in their own PR

crates/ui/src/components/data_table.rs
```diff
@@ -1147,7 +676,9 @@ impl Component for Table {
                             single_example(
                                 "Simple Table",
                                 Table::new(3)
-                                    .width(px(400.))
+                                    .width_config(ColumnWidthConfig::auto_with_table_width(px(
+                                        400.,
+                                    )))
```
Anthony-Eid: Not sure I like the new Table interface. We try to keep UI APIs like this as simple as possible, maybe we could just say `width(px(400.))` defaults to auth_with_table_width?

Anthony-Eid (commented):
If we end up splitting up some of `data_table.rs` into `width_managements.rs` that should be it's own PR, just to make this one as easy to review as possible.

crates/ui/src/components/data_table/width_management.rs
HalavicH: Okay i'll merge it back in

crates/ui/src/components/data_table.rs
HalavicH: I removed it to avoid ambiguity, but sure, i can rollback it

Cargo.toml
HalavicH: it seems i accidentally commited redundant commit

crates/ui/src/components/data_table.rs
Anthony-Eid: maybe we could change the name to `width_auto`, we could also look at web mdm and tailwind to see if there already a similar component that has an interface.

PR #51061, opened 2026-03-08
search: Fix focus replacement field when opening replace (Ctrl+H)

CCXLV:
Previously, focus stayed on the search bar because a pre-focus check `handle.is_focused(window)` was always false at deploy time.

Release Notes:

- Fixed: When opening find-and-replace with `Ctrl+H`, the replacement input is now focused instead of the search bar.

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -977,10 +977,17 @@ impl BufferSearchBar {
                 let mut handle = self.query_editor.focus_handle(cx);
                 let mut select_query = true;
 
-                let has_seed_text = self.query_suggestion(window, cx).is_some();
-                if deploy.replace_enabled && has_seed_text {
-                    handle = self.replacement_editor.focus_handle(cx);
-                    select_query = false;
+                if deploy.replace_enabled {
+                    let has_selection = self
+                        .active_searchable_item
+                        .as_ref()
+                        .map(|item| item.has_non_empty_selection(window, cx))
+                        .unwrap_or(false);
+
+                    if has_selection {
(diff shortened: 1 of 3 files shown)


crates/search/src/buffer_search.rs
```diff
@@ -973,9 +973,18 @@ impl BufferSearchBar {
...
-                if deploy.replace_enabled && handle.is_focused(window) {
-                    handle = self.replacement_editor.focus_handle(cx);
-                    select_query = false;
+
+                if deploy.replace_enabled {
+                    let has_selection = self
+                        .active_searchable_item
+                        .as_ref()
+                        .map(|item| item.has_non_empty_selection(window, cx))
+                        .unwrap_or(false);
```
Veykril: No need for a new method here, we should tie this to what the default search population uses

```suggestion
                    let has_selection = self.query_suggestion(window, cx)
                        .is_some();
```

CCXLV: Just to confirm: this means it'll also focus the replace field when the query is seeded from the cursor word (no explicit selection). Is that the behavior we want here? That new method makes sure that the buffer search is being opened with an actual selection.

Veykril: Yea I think that feels right to me personally, the reason at least for me disliking yet another new method here is that `SearchableItem` has already a lot of methods (and I do not want this to devolve into something like `Item`)

Veykril (approved):
Thanks!

Veykril:
can you rebase on top of main once more? I think the CI step is bugged on the branch base you are on

PR #35157, opened 2025-07-27
Change behavior of search with vim mode enabled

viveksjain:
When vim mode is enabled, previously if `Cmd-F` (or platform equivalent) was pressed, enter will go to the editor's first match, and then hitting enter again goes to the next line rather than next match. This PR changes it to make enter go to the next match, which matches the convention in most other programs. The behavior when search is initiated with `/` is left unchanged.

Closes #7692

Release Notes:

- Improved: In vim mode, if search is not triggered with `/` and instead a keyboard shortcut, hitting enter goes to the next match

PR #51073, opened 2026-03-09
Change behavior of search with vim mode enabled

### LINKED ISSUE #7692: Vim mode search should act like non-Vim mode when not triggered with `/`
### Summary

When searching with `/` in Vim mode, hitting `enter` moves the cursor to the next line in the buffer being searched, matching Vim's behavior. This is at odds with how `enter` behaves in a non-Vim-mode buffer search, jumping to the next result. We should refine the action here to behave like non-Vim mode unless `/` is used to start the search.

<details>
image
</details>

viveksjain:
When vim mode is enabled, previously if Cmd-F (or platform equivalent) was pressed, enter will go to the editor's first match, and then hitting enter again goes to the next line rather than next match. This PR changes it to make enter go to the next match, which matches the convention in most other programs. The behavior when search is initiated with / is left unchanged.

This is a reopen of #35157, rebased and fixed.
Closes #7692

Release Notes:

- In vim mode, when search is triggered by the non-vim mode shortcut (cmd-f by default) enter will now behave as it does outside of vim mode.

### DIFF
--- a/crates/vim/src/helix.rs
+++ b/crates/vim/src/helix.rs
@@ -648,6 +648,8 @@ impl Vim {
                     self.search = SearchState {
                         direction: searchable::Direction::Next,
                         count: 1,
+                        vim_mode_search: true,
+                        show_match_highlights: false,
                         prior_selections,
                         prior_operator: self.operator_stack.last().cloned(),
                         prior_mode: self.mode,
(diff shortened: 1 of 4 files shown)


ConradIrwin:
@viveksjain can we use `!vim_mode_search` instead of having a second variable to track? I would have thought they would be the same condition

viveksjain:
I think I've fixed the tests now!

ConradIrwin (approved):

PR #18513, opened 2024-09-29
Project panel horizontal scrollbar

SomeoneToIgnore:
image

https://github.com/user-attachments/assets/734f1f52-70d9-4308-b1fc-36c7cfd4dd76

Closes https://github.com/zed-industries/zed/issues/7001
Closes https://github.com/zed-industries/zed/issues/4427
Part of https://github.com/zed-industries/zed/issues/15324
Part of https://github.com/zed-industries/zed/issues/14551

* Adjusts a `UniformList` to have a horizontal sizing behavior: the old mode forced all items to have the size of the list exactly.
A new mode (with corresponding `ListItems` having `overflow_x` enabled) lays out the uniform list elements with width of its widest element, setting the same width to the list itself too.

* Using the new behavior, adds a new scrollbar into the project panel and enhances its file name editor to scroll it during editing of long file names

* Also restyles the scrollbar a bit, making it narrower and removing its background

* Changes the project_panel.scrollbar.show settings to accept `null` and be `null` by default, to inherit `editor`'s scrollbar settings. All editor scrollbar settings are supported now.

Release Notes:

- Added a horizontal scrollbar to project panel ([#7001](https://github.com/zed-industries/zed/issues/7001)) ([#4427](https://github.com/zed-industries/zed/issues/4427))

Issue #7001, opened 2024-01-29
Project panel horizontal scroll


NicolayD:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

The left dock doesn't seem to have a horizontal scroll and the names of nested files are hidden. Resizing manually works, but it would be a welcome addition to have a horizontal scroll.

### Environment

Zed: v0.119.20 (Zed)
OS: macOS Sonoma 14.3
Memory: 8 GiB
Architecture: aarch64, Apple M1

### If applicable, attach your `~/Library/Logs/Zed/Zed.log` file to this issue.
If you only need the most recent lines, you can run the `zed: open log` command palette action to see the last 1000.

_No response_

timfjord:
As per https://github.com/zed-industries/zed/issues/5550 it is a design decision to make it unscrollable

NicolayD:
> As per #5550 it is a design decision to make it unscrollable

It's still very strange to use - even at only 5 or 6 levels deeply nested directories filenames start to get hidden and it's not very comfortable to have to manually expand it and shrink it. For example, in Sublime it is scrollable and in my experience has worked without issues.

timfjord:
I think it has to do with how renaming items works in Zed.
In Zed, you rename files directly in the Project Panel whereas in ST there is a popup located outside of the sidebar.

tausifcreates:
> I think it has to do with how renaming items works in Zed.

It's not just related items tho. if the file names are long or file structure is deeply nested, I need to resize the panel often to view the files.

tausifcreates:
To tackle the problem of renaming the file in place (not a popup like Intellij), we could possibly lock the x-scroll when a file is being edited in Zed. As code is read more than it's written, it would make some sense cause files are renamed miniscule amount of times than it is read 🙂

belianu:
I have the same problem too. 
When you have many folders and the file names are long it is really frustrating not to be able to scroll the file view horizontally. 
Zed is really good at many things, keeping the file view unscrollable would be a really bad lapse in style. 
I hope that the slider will be implemented in the future, or at least give the possibility to activate it from the settings if someone wants to activate it.

notpeter:
A potential workaround: Zed has a settings option for the project panel to folder/collapse nested folders which only contain a single directory:
```
  "project_panel": {
    "auto_fold_dirs": true
  },
```

[...]

Issue #5550, opened 2022-07-06
Allow project panel to be scrolled horizontally


maraisr:
With a very deeply nested file tree, and a somewhat cozy viewport, file names cannot be read easily.

![image](https://user-images.githubusercontent.com/599459/177655736-4735e4a2-57c0-487c-ba17-586eddcd7862.png)

PR #51143, opened 2026-03-10
project_panel: Add horizontal scroll setting

k4yt3x:
This PR introduces the `project_panel.scrollbar.horizontal_scroll` setting to allow users to toggle the horizontal scroll bar in the project panel. This was Zed's design before PR #18513, and the default behavior of VSCode (`workbench.list.horizontalScrolling`).

https://github.com/user-attachments/assets/f633f4e4-a585-4494-8f48-df77c6aca418

## Rationale

Zed's design used to be the same as the default behavior of VSCode. I.e., no horizontal scrolling, and the view is always snapped to the left, with long file names clipped of. If you want to see the content that is out-of-frame, you'll need to drag the handle and expand the project panel. This could be problematic, especially for large repos with multiple levels of nested directories, as pointed out by issues #5550 and #7001.

image\
*VSCode's default setup, for reference.*

Then came PR #18513, which added horizontal scroll and addressed this pain point, but users didn't have a choice. They're stuck with horizontal scrolling always turned on. I, for instance, personally prefer the old, VSCode-default behavior, for most projects I open are small and don't need horizontal scrolling in the project panel. With horizontal scrolling always turned on, I find it annoying to have my project panel view accidentally scrolled to the middle, and I'll have to grab my mouse and scroll it back. It's also visually redundant.

Thus, why not add an option like VSCode's `workbench.list.horizontalScrolling` and let users choose? I'd love to be able to, say, set a per-project override for the projects that need horizontal scrolling, while having it disabled by default.

## Extra Notes

- I was originally thinking about using `ScrollbarAxes` from `src/editor_settings.rs` and make the option `project_panel.scrollbar.axes.horizontal` similar to the global editor scrollbar settings, but this option is specific to the project panel and it doesn't quite make sense to allow disabling vertical scrolling on the project panel, so I added a standalone option for it instead, similar to VSCode's `workbench.list.horizontalScrolling`.

- I went the conservative route and set horizontal scrolling to enabled (current behavior) by default. Imo it might make more sense to disable it by default instead, similar to VSCode, but I'll leave this for the Zed team to decide.

- I named it `horizontal_scroll` instead of `horizontal_scrolling` to be consistent with the adjacent setting `sticky_scroll`.

- As for tests, I don't see tests for the scrollbar, so I didn't add any.

I'd be glad to update the PR if anything is not inline with the project's requirements or conventions.

---

Release Notes:

- Added `project_panel.scrollbar.horizontal_scroll` setting to allow toggling horizontal scrolling in the project panel

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -768,6 +768,9 @@
       // 5. Never show the scrollbar:
       //    "never"
       "show": null,
+      // Whether to allow horizontal scrolling in the project panel.
+      // When false, the view is locked to the leftmost position and long file names are clipped.
+      "horizontal_scroll": true,
     },
     // Which files containing diagnostic errors/warnings to mark in the project panel.
     // This setting can take the following three values:
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -6309,6 +6309,7 @@ impl Render for ProjectPanel {
         let panel_settings = ProjectPanelSettings::get_global(cx);
         let indent_size = panel_settings.indent_size;
         let show_indent_guides = panel_settings.indent_guides.show == ShowIndentGuides::Always;
+        let horizontal_scroll = panel_settings.scrollbar.horizontal_scroll;
         let show_sticky_entries = {
             if panel_settings.sticky_scroll {
                 let is_scrollable = self.scroll_handle.is_scrollable();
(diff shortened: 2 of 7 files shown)


SomeoneToIgnore (commented):
Thank you, can we adjust `crates/settings/src/vscode_import.rs` also to have the settings properly imported?

SomeoneToIgnore (commented):
Great, thank you so much for adding this all.

Sorry, there's one last thing I've forgotten to mention which we need to adjust: https://github.com/zed-industries/zed/blob/3bc4b584b17b1b2858021ceef52dd27cfeb9cd83/docs/src/reference/all-settings.md?plain=1#L4944-L4959

There, we describe all our settings so that they appear on the website and can be indexed better: https://zed.dev/docs/reference/all-settings?highlight=search%3A#search

Happy to merge the PR right after that.

SomeoneToIgnore (approved):
Great, thank you!

PR #51161, opened 2026-03-10
git_ui: Fix trashing untracked files

### LINKED ISSUE #35417: Trashing paths from git panel doesn't work when a subdirectory of the repo is open

nihalxkumar:
<details>
<summary>Problem</summary>

When a workspace is opened on a subdirectory of a git repo (e.g. zed/crates), trashing untracked files that live outside that subdirectory (e.g. foo at the repo root) silently fails. The git panel shows these files via the repository's status, but when you try to trash them:

`repo_path_to_project_path` returns `None` because the file's path doesn't fall within any opened worktree.
The old code in `revert_entry` early-returns via `maybe!` on that `None`, swallowing the failure.
The old code in `clean_all` skips the file in its `filter_map`, so nothing happens either.
</details>

Solution

Extract `GitPanel::trash_entries` — a shared helper that both `revert_entry` (single file) and `clean_all` (bulk "Trash Untracked Files") now delegate to. The helper uses a two-tier strategy for each entry:

Project-level deletion: If `repo_path_to_project_path` resolves to a `ProjectPath` and `Project::delete_file` finds a matching entry, delete through the Project (the existing path — handles worktree events, buffer cleanup, etc.).
Repository-level fallback: If the project path can't be resolved, or if `delete_file` returns `None` (no entry in the project for that path), collect the repo path and batch-delete via the new `Repository::trash_paths`.

This gracefully handles the case where `delete_file` returns `None` without aborting — the path simply falls through to the repo-level fallback instead of becoming a fatal error.

Add `Repository::trash_paths` on the git store side. This method:

- Resolves each `RepoPath` to an absolute path via the repository snapshot.
- Schedules a tracked job (`spawn_job_with_tracking`) so the UI shows pending status.
- Trashes files/directories on disk using the `Fs` trait (supports both files and recursive directory removal).
- Calls `paths_changed` after completion so the git status refreshes and the entry disappears from the panel.
- Errors on remote repositories with a clear message, since direct filesystem access isn't available.

https://github.com/user-attachments/assets/f02af06e-7a7f-4ae5-b325-df20ec5fe0f7

Closes #35417

Release Notes:

- Fixed trashing untracked files from the git panel when a subdirectory of the repository is opened as the workspace root.


dinocosta (commented):
Hey @nihalxkumar 👋 

Thank you for taking a stab at this one and including tests! I've tested out your changes and confirmed that it does seem to fix the issue reported in https://github.com/zed-industries/zed/issues/35417 but one thing I did notice, which you might want to take a look at is that, even though I can delete the file, if I create it again outside of Zed, it no longer shows up in the Git Panel.

For example, using Zed's repo:

1. Create `zed/foo.txt`
2. Open Zed on `zed/crates`
3. Open the Git Panel and confirm `foo.txt` is shown
4. Trash the `foo.txt` file
5. Repeat step 1 and confirm how the file does not show up in the Git Panel

Assigning this to Cole as he's more familiar with our Git integration and created the original issue. 

Small request though, the original issue – https://github.com/zed-industries/zed/issues/35417 – was already assigned to someone else, so it's always best to first leave a comment in the issue asking if the assignee is ok with you taking over the issue or even pairing on it 🙂

nihalxkumar:
Hey @dinocosta @cole-miller

PR #37123, opened 2025-08-28
Use nucleo for fuzzy string matching

P1n3appl3:
Closes #7586

Release Notes:

- Fuzzy pickers now use the [nucleo](https://docs.rs/nucleo/latest/nucleo) library from the [Helix editor](https://helix-editor.com) TODO: add a note about new syntax and/or some examples

PR #51164, opened 2026-03-10
Add fuzzy_nucleo crate for order independent file finder search

### LINKED ISSUE #14428: Ordering of search tokens in file finder fuzzy match
### Check for existing issues

- [x] Completed

### Describe the feature

Firstly, zed looks, feels, and works great! Thank you. I've been playing with it it since the Linux release and stumbled upon this muscle-memory quality-of-life enhancement.

Universally, fuzzy search in file finder widgets across editors are agnostic of the ordering of search tokens. For instance, the search inputs `internal auth` and `auth internal` should fuzzy-match the path `/internal/auth`. However, in the current version of zed (v0.143.7), the search is not ordering agnostic. In the above example, only `internal auth`  produces a hit. Screenshots below.

Possibly related: https://github.com/zed-industries/zed/issues/7586

Actual path: `/internal/auth`

Results for `auth`
image

Results for `internal auth`
image

Results for `auth internal`. Ideally, this should also produce the same result.
image

## Comparison
vscode
image

Sublime
image

11happy:
Closes #14428 

https://github.com/user-attachments/assets/7e0d67ff-cc4e-4609-880d-5c1794c64dcf

Release Notes:

- Adds a new `fuzzy_nucleo` crate that implements order independent path 
matching using the `nucleo` library. currently integrated for file finder.


crates/fuzzy_nucleo/LICENSE_GPL
MrSubidubi, line 1: CI will also tell you this, but this should just be a softlink to the license file living at the root.

11happy: done, have fixed the clippy & addded a symlink of License

crates/file_finder/src/file_finder.rs
```diff
@@ -765,6 +757,17 @@ fn matching_history_items<'a>(
...
+                if let Some(filename) = path_match.path.file_name() {
+                    let filename_start = path_match.path.as_unix_str().len() - filename.len();
+                    path_match
+                        .positions
+                        .iter()
+                        .any(|&pos| pos >= filename_start)
+                } else {
+                    true
+                }
+            })
```
11happy, line 750: The earlier fuzzy matcher scored directory matches differntly  however for example nucleo scores `colab_ui` against `collab_ui/first.rs` high, while working a previous test was failing (`test_history_items_vs_very_good_external_match`)  so filtering out matches where query does not match filename.

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -0,0 +1,312 @@
...
+        } else {
+            candidate_buf.push_str(candidate.path.as_unix_str());
+        }
+
+        let mut indices = Vec::new();
+        let mut buf = Vec::new();
+        let haystack = Utf32Str::new(&candidate_buf, &mut buf);
+
+        if let Some(score) = pattern.indices(haystack, matcher, &mut indices) {
+            let length_penalty = candidate_buf.len() as f64 * 0.001;
```
11happy: added a length penalty as nucleo was scoring `maaa.rs` & `moo.rs` same (in this case when I debugged it was 80 for both ) causing some test failures ex.(`test_keep_opened_file_on_top_of_search_results_and_select_next_one`)

11happy (commented):
I have kept lot of data structures & methods similar to what already exist in current fuzzy crate & also took reference from this PR https://github.com/zed-industries/zed/pull/37123 to implement `path_match_helper`

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -0,0 +1,312 @@
...
+        self.score
+            .partial_cmp(&other.score)
+            .unwrap_or(Ordering::Equal)
+            .then_with(|| self.worktree_id.cmp(&other.worktree_id))
+            .then_with(|| {
+                other
+                    .distance_to_relative_ancestor
+                    .cmp(&self.distance_to_relative_ancestor)
+            })
+            .then_with(|| self.path.cmp(&other.path))
```
yara-blue, line 73: in our original attempt we had another tie breaker based on path length: 

```rust
            .then_with(|| {
                self.path
                    .as_unix_str()
                    .chars()
                    .count()
                    .cmp(&other.path.as_unix_str().chars().count())
            })
```

We had this to prevent this test (also added in the unmerged #37123) from failing:
```rust
    #[gpui::test]
    async fn shorter_over_lexicographical(cx: &mut TestAppContext) {
        const CANDIDATES: &'static [&'static str] = &["qr", "qqqqqqqqqqqq"];
        assert_eq!(
            path_matches(cx, CANDIDATES, "q", None).await,
            ["qr", "qqqqqqqqqqqq"]
        );
    }
```

Do you think that makes sense to add here?

yara-blue: nevermind, I think you've taken care of that wit the length penalty in the `pattern.indices` call right?

yara-blue: just saw your comment below :)

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -0,0 +1,312 @@
...
+        .scoped(|scope| {
+            for (segment_idx, results) in segment_results.iter_mut().enumerate() {
+                let pattern = pattern.clone();
+                let relative_to = relative_to.clone();
+                scope.spawn(async move {
+                    let segment_start = segment_idx * segment_size;
+                    let segment_end = segment_start + segment_size;
+                    let mut config = nucleo::Config::DEFAULT;
+                    config.set_match_paths();
+                    let mut matcher = nucleo::Matcher::new(config);
```
yara-blue: We will be calling this a lot (on every character typed). Creating the Matcher can take a while since it has to allocate a fair bit. In the org implementation (https://github.com/zed-industries/zed/blob/1caafceac004ca8de576a80657395b5eb1f6ce21/crates/fuzzy/src/matcher.rs) we here caching those matchers between calls. Could be worth doing here, what do you think?

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -0,0 +1,312 @@
...
+
+                    let mut tree_start = 0;
+                    for candidate_set in candidate_sets {
+                        if cancel_flag.load(atomic::Ordering::Acquire) {
```
yara-blue: This is not needed, its taken care off by `path_match_helper` returning Err when cancelled.

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -0,0 +1,312 @@
...
+#[derive(Clone, Debug)]
+pub struct PathMatchCandidate<'a> {
+    pub is_dir: bool,
+    pub path: &'a RelPath,
+}
+
+#[derive(Clone, Debug)]
+pub struct PathMatch {
+    pub score: f64,
+    pub positions: Vec<usize>,
```
yara-blue, line 25: I remember these needed to be sorted otherwise the rendering would crash sometimes. I do not think that display logic has changed (though I have not checked) . Maybe that was for another picker, then we'll run into it later when we expand the scope for Nucleo based search.

Shall we sort them or leave a note in the code in case someone has to debug a panic?

crates/file_finder/src/file_finder.rs
yara-blue: it's not completely clear to me how filtering out some matches helps reordering. I feel this could use a comment in the code. What do you think?

yara-blue (changes requested):
A few small concerns otherwise looks great. Let me know when it's ready for another review :)

yara-blue:
> ### 📏 PR Size: **547 lines changed** (Size L)
> 
> Please note: this PR exceeds the 400 LOC soft limit.
> 
>     * Consider **splitting** into separate PRs if the changes are separable
> 
>     * Ensure the PR description includes a **guided tour** in the "How to Review" section so reviewers know where to start

you may ignore this :)

crates/fuzzy_nucleo/src/paths.rs
yara-blue: are these always sorted though? I do not see them getting sorted anywhere now.

yara-blue (changes requested):
Code is looking good, Unfortunately I found regression compared to the current ranking:

this PR
image

mainline Zed:
image

As fuzzy finding is a method of navigation for a large user base we must be sure not to regress there (they would get pretty annoyed). 

We can still get this over the line though! I think what we need is to extend the test-suite for the file finder comparing the current fuzzy finder and this one. With the above (and the audio settings example) we have at least two test cases already. Maybe you can find a few more?

Let me know what you think.

crates/fuzzy_nucleo/src/paths.rs
11happy: sure I have them sorted now in `path_macth_helper`

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -0,0 +1,329 @@
...
+                .iter()
+                .filter(|&&i| i as usize >= filename_start)
+                .count();
+
+            let all_in_filename = filename_match_count == indices.len();
+            let density = filename_match_count as f64 / filename_len as f64;
+            let filename_bonus = (all_in_filename && density > 0.5)
+                .then_some(FILENAME_BONUS)
+                .unwrap_or(0.0);
+            let adjusted_score = score as f64 + filename_bonus - length_penalty;
```
11happy: a filename density score, it cheeks if all the matched indices are in the filename part & also checks the density wrt to length ,

crates/fuzzy_nucleo/src/paths.rs
yara-blue: ahh thanks! I missed them.

crates/fuzzy_nucleo/src/paths.rs
yara-blue: nice solution! Could be extracted to its own function to help with readability, or not, your call. I'm gonna try out a few more searches by hand tomorrow morning unless that turns up anything I expect this to be merged shortly after! Really nice work.

11happy: but still its not working correct fully if we search `ai onboarding` the correct result is at 4th index should be at top. I will debug & update once I have it fixed.
image

yara-blue: its surprising how much refinement is needed to get the best results out of Nucleo.

Thank you for keeping at it! I'm sure you'll get it perfect.

11happy: its working now, thank you for having patience with this one.

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -87,11 +99,32 @@ pub(crate) fn distance_between_paths(path: &RelPath, relative_to: &RelPath) -> u
     path_components.count() + relative_components.count() + 1
 }
 
+fn get_filename_match_bonus(
+    candidate_buf: &str,
+    query_atoms: &[Atom],
+    matcher: &mut nucleo::Matcher,
+) -> f64 {
+    let filename_start = candidate_buf.rfind(['/', '\\']).map_or(0, |i| i + 1);
```
yara-blue: maybe std::path::Path::new(candidate_buf) then you can just use .filename()
in the Utf32Str::new you can the use the buffer_candidate + filename.len()

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -109,8 +142,12 @@ fn path_match_helper<'a>(
         String::new()
     };
     let path_prefix_len = candidate_buf.len();
-
+    let mut buf = Vec::new();
+    let mut all_indices: Vec<u32> = Vec::new();
```
yara-blue: I was wondering if you know a better name for indices? Or maybe a short doc comment explaining what they are used for. This code has always been hard to get into would be nice to make it more accessible.

yara-blue (changes requested):
Really elegant solution with the Atoms! 

I'm gonna daily drive this for two days and see if I can find any other regression. If you could do that minor file_name refactor then we should be good. 

Oh and see my note on indices. You can absolutely leave them as is however there are many more pickers to make fuzzy so it might help us in the future to have this code as readable as possible :)

crates/fuzzy_nucleo/src/paths.rs
11happy: done

crates/fuzzy_nucleo/src/paths.rs
11happy: done, I have the names changed.

yara-blue (approved):

gee-forr:
I saw the release notes and was compelled to post a HUGE THANK YOU!!!

No matter how I tried, I could not rewire my muscle memory around this limitation compared to VSCode's search algo. THANK YOU THANK YOU THANK YOU!!!!!!!

PR #51174, opened 2026-03-10
workspace: Avoid environment contamination when opening local settings from remote

cppcoffee:
Release Notes:

- Fixed settings file cannot be opened when accessing a remote directory

before:

https://github.com/user-attachments/assets/10595772-f085-424a-9e5e-b1630ed6396c

after:

https://github.com/user-attachments/assets/42e56254-d07a-4c20-946d-1d2e2ded2a2d

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -2693,6 +2693,7 @@ impl Workspace {
             cx.spawn_in(window, async move |_vh, cx| {
                 let (multi_workspace_window, _) = task.await?;
                 multi_workspace_window.update(cx, |multi_workspace, window, cx| {
+                    window.activate_window();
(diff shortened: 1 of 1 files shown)


dinocosta (commented):
Hey @cppcoffee 👋 

I've assigned this to a team member I suspect is more familiar with how this works but, while testing, I noticed there's changes in  both `workspace::Workspace::with_local_workspace` and `workspace::Workspace::with_local_or_wsl_workspace`. Seems like `settings_ui::SettingsWindow::open_current_settings_file` and `zed::zed::open_settings_file` both end up using `workspace::Workspace::with_local_or_wsl_workspace` . 

Reverting the changes to `workspace::Workspace::with_local_workspace` doesn't seem to break anything in terms of being able to open the settings file while in a remote project, so might be worth taking another look 🙂

Issue #48682, opened 2026-02-07
No protocol to open project in Dev Container mode right from CLI


ai:
### Reproduction steps

```sh
zed devcontainer://home/user/my-project-with-devcontainer
```

Maybe I need to use different protocol. Let’s put it to docs: https://zed.dev/docs/dev-containers

### Current vs. Expected behavior

Current: Open non-existed file

Expected: Start project in Dev Container mode

### Zed version and system specs

Zed: v0.222.4+stable.147.b385025df963c9e8c3f74cc4dadb1c4b29b3c6f0 (Zed) OS: Linux Wayland fedora 43 Memory: 27.2 GiB Architecture: x86_64 GPU: AMD Radeon 780M Graphics (RADV PHOENIX) || radv || Mesa 25.3.4

### If you are using WSL on Windows, what flavor of Linux are you using?

None

ai:
> I’m not sure if it exists in VS Code either.

I little hacky, but possible:

```
code --folder-uri vscode-remote://attached-container+<container id or name>/workspace/path
```

People [asked](https://stackoverflow.com/questions/60379221/how-to-attach-a-remote-container-using-vscode-command-line) for this, so it is useful.

> If you open a project with a valid Dev Container setup, Zed should automatically detect it as one for you and prompt you accordingly. Are you not seeing that?

There are two smalls problems with that:

1. Extra click which I can miss
2. https://github.com/zed-industries/zed/issues/48683

And one big (for me): `--wait` does not track project reopening in Dev Container mode. `--wait` will keep terminal busy only until I click `Reopen in Dev Container` (since we are recreating Zed session). I have special options for Dev Container, so I start in manually, and I am [using `--wait` to close Dev Container](https://github.com/ai/environment/blob/main/bin/zed-isolate#L59-L66) when I finish editing project.

ai:
> You can answer if you’d like but I’ll check myself when I get the chance

Sorry, I am so big fan of Zed, so I deleted VS Code :D.

I installed it and was not able to find such setting.

> a setting to auto-detect a Dev Container proejct and start it up in Docker? If so, I don’t mind suggesting this be added as a settings option.

My idea is to add protocol which will open a project and call `projects: open dev container` automatically. It contrast with detecting Docker contain it is implicit and could be easier to implement.

But detecting running Docker image for current project and start Dev Container mode will also work for me (but from UX perspective it could be a little less explicit).

> Hmm, this sounds like another bug.

I am going to document it right now

[...]

PR #51175, opened 2026-03-10
cli: Add --dev-container flag to open project in dev container

antont:
## Before you mark this PR as ready for review, make sure that you have:

## Summary

Adds a `--dev-container` CLI flag that automatically triggers "Reopen in Dev Container" when a `.devcontainer/` configuration is found in the project directory.

```sh
zed --dev-container /path/to/project
```

## Motivation

This enables fully scripted dev container workflows — for example, creating a git worktree and immediately opening it in a dev container without any manual UI interaction:

```sh
git worktree add ../feature-branch
zed --dev-container ../feature-branch
```

The dev container modal fires automatically once the workspace finishes initializing, so the environment is ready by the time you look at the window. This is useful for automation scripts that prepare environments and kick off agent runs for tasks like bug report triage.

Here's an [example script](https://github.com/antont/todo-rs-ts/blob/main/scripts/devcontainer-new.sh) that creates a worktree and opens it as a dev container in one step.

Related: #48682 requests a `devcontainer://` protocol for connecting to already-running dev containers — a complementary but different use case. This PR covers the "open project and trigger dev container setup" path.

## How it works

- The `--dev-container` flag flows through the CLI IPC protocol to the workspace as an `open_in_dev_container` option.
- On the first worktree scan completion, if devcontainer configs are detected, the dev container modal opens automatically.
- If no `.devcontainer/` config is found, the flag is cleared and a warning is logged.

## Notable changes

- **`Workspace::worktree_scans_complete`** promoted from `#[cfg(test)]` to production. It was only test-gated because it had no production callers — it's a pure read-only future with no side effects.
- **`suggest_on_worktree_updated`** now takes `&mut Workspace` to read and clear the CLI flag.
- Extracted **`open_dev_container_modal`** helper shared between the CLI code path and the suggest notification.

## Test plan

- [x] `cargo test -p zed open_listener` — includes `test_dev_container_flag_opens_modal` and `test_dev_container_flag_cleared_without_config`
- [x] `cargo test -p recent_projects` — existing suggest tests still pass
- [x] Manual: `cargo run -- --dev-container /path/to/project-with-devcontainer` opens the modal

Release Notes:

- Added `--dev-container` CLI flag to automatically open a project in a dev container when `.devcontainer/` configuration is present.

### DIFF
--- a/crates/recent_projects/src/dev_container_suggest.rs
+++ b/crates/recent_projects/src/dev_container_suggest.rs
@@ -62,10 +62,8 @@ pub fn suggest_on_worktree_updated(
     if cli_auto_open {
         workspace.set_open_in_dev_container(false);
         let task = cx.spawn_in(window, async move |workspace, cx| {
-            let scans_complete = workspace
-                .update(cx, |workspace, cx| {
-                    workspace.worktree_scans_complete(cx)
-                })?;
+            let scans_complete =
+                workspace.update(cx, |workspace, cx| workspace.worktree_scans_complete(cx))?;
             scans_complete.await;
 
             workspace.update_in(cx, |workspace, window, cx| {
@@ -79,9 +77,7 @@ pub fn suggest_on_worktree_updated(
                         window.dispatch_action(Box::new(zed_actions::OpenDevContainer), cx);
                     });
                 } else {
-                    log::warn!(
-                        "--dev-container: no devcontainer configuration found in project"
-                    );
+                    log::warn!("--dev-container: no devcontainer configuration found in project");
                 }
             })
         });
--- a/crates/recent_projects/src/dev_container_suggest.rs
+++ b/crates/recent_projects/src/dev_container_suggest.rs
@@ -75,7 +75,9 @@ pub fn suggest_on_worktree_updated(
                     .worktrees(cx)
                     .any(|wt| !find_configs_in_snapshot(wt.read(cx)).is_empty());
                 if has_configs {
-                    window.dispatch_action(Box::new(zed_actions::OpenDevContainer), cx);
+                    cx.on_next_frame(window, move |_workspace, window, cx| {
+                        window.dispatch_action(Box::new(zed_actions::OpenDevContainer), cx);
+                    });
                 } else {
                     log::warn!(
                         "--dev-container: no devcontainer configuration found in project"
--- a/crates/recent_projects/src/dev_container_suggest.rs
+++ b/crates/recent_projects/src/dev_container_suggest.rs
@@ -1,5 +1,6 @@
+use crate::RemoteServerProjects;
 use db::kvp::KeyValueStore;
-use dev_container::find_configs_in_snapshot;
+use dev_container::{DevContainerContext, find_configs_in_snapshot, find_devcontainer_configs};
 use gpui::{SharedString, Window};
 use project::{Project, WorktreeId};
 use std::sync::LazyLock;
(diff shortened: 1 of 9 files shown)


crates/recent_projects/src/dev_container_suggest.rs
```diff
@@ -12,6 +13,33 @@ use worktree::UpdatedEntriesSet;
 
 const DEV_CONTAINER_SUGGEST_KEY: &str = "dev_container_suggest_dismissed";
 
+fn open_dev_container_modal(
```
KyleBarton: @antont thanks for the contribution. I'm unsure why this function needs to exist? Can't the caller dispatch the `OpenDevContainer` action, the way the modal suggester does later on in this file?

antont: @KyleBarton 

Thanks a lot of the quick and insightful review!

> I'm unsure why this function needs to exist? Can't the caller dispatch the `OpenDevContainer` action, the way the modal suggester does later on in this file?

I think it can, but it was difficult to cover in tests.  That's why I left the todo note, anticipating that Docker could be mocked soon, or some other such solution would make devcontainer tests for this easier to write. (// TODO: Remove once dev container commands are mockable in tests.)

I went a bit back & forth with the approach, the version without the extra function is in my original commit history in my fork.

**Now I have a candidate solution though!** Add `AppState::set_global(Arc::downgrade(&app_state), cx);` in `init_test_with_state`.  My earlier difficulty was that the OpenDevContainer action handler uses `with_active_or_new_workspace` which requires `GlobalAppState`, making it unusable from the spawn context in tests. 

Without that, the test errs:

>   called `Result::unwrap()` on an `Err` value: no state of type workspace::GlobalAppState exists

at workspace.rs:9988, which is `AppState::global(cx)` line in the `None` branch of `with_active_or_new_workspace`.

Claude's explanation of the chain:
 
>   1. suggest_on_worktree_updated dispatches OpenDevContainer via window.dispatch_action()
>   2. The action handler (registered in recent_projects::init) calls with_active_or_new_workspace(cx,
>   ...)
>   3. with_active_or_new_workspace (line 9969) has two branches:
>     - Active window exists (line 9977): uses cx.active_window() directly — no GlobalAppState needed
>     - No active window (line 9987): falls back to AppState::global(cx) (line 9988), which reads from
>   cx.global::<GlobalAppState>() to create a new workspace
> 
>   In the test, the action is dispatched from within a window, so it actually hits the first branch
>   (active window exists). But the dispatch_action call goes through the global action dispatch on App,
>   which means the action handler receives &mut App — not a window context. So cx.active_window()
>   returns None, and it falls into the second branch that needs GlobalAppState.
> 
>   That's why just adding AppState::set_global(Arc::downgrade(&app_state), cx) to init_test_with_state
>   fixes it — it makes the fallback path work.

I'm testing this new version now - I had a stacked PR for it first but it's now a new commit here.

antont: Note @KyleBarton - I was wrong actually, the problem with the action dispatching was there in normal GUI run too, not only tests:

The first new commit here, which removed the extra function and switched to action dispatch, hits error:

> 2026-03-13T11:33:21+02:00 ERROR [recent_projects::remote_servers] No active project directory for Dev Container

I think it may have worked for me first, that's why I pushed it. Might be timing related. Certainly it was not reliable, now it was failing to me, while the original version was reliable.

The 2nd new commit restores the defer to `cx.on_next_frame` while keeping the action dispatching. This has been reliable, I made a minimal dev container and launched with it maybe 15 times total in both release and dev modes.

I'll keep using the current version here in the PR now for my work so it gets tested a lot, I'm working on a semi-automated dev task execution thing which is now triggering this kind of a Zed run always when I'm testing it.

PR #51194, opened 2026-03-10
repl: Add notebook command/edit modes and keybindings

MostlyKIGuess:
- Introducing NotebookMode state and handlers (EnterEditMode, EnterCommandMode, RunAndAdvance).
- Wire up UI to switch modes,focus editors appropriately, and advance selection while in command mode.
- Update default and vim keymaps to use the new bindings (shift-enter runs+advances, escape enters command mode,
and enter/up/down work in command mode).
- Reveal and Focus the new Cell when inserted 

Release Notes:

- N/A


crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -614,8 +603,84 @@ impl NotebookEditor {
...
+        if self.selected_cell_index < count - 1 {
+            self.selected_cell_index += 1;
+            self.cell_list
+                .scroll_to_reveal_item(self.selected_cell_index);
+        }
+        self.notebook_mode = NotebookMode::Command;
+        self.focus_handle.focus(window, cx);
+        cx.notify();
+    }
+
```
rgbkrk, line 661: Beautiful. I look forward to trying this.

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -212,6 +200,7 @@ impl NotebookEditor {
                                 .position(|id| id == &cell_id_for_editor)
                             {
                                 this.selected_cell_index = index;
+                                this.notebook_mode = NotebookMode::Edit;
                                 cx.notify();
                             }
```
MrSubidubi: Since this seems to be the same code everywhere here, how about we move this into a method instead?

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -614,8 +603,84 @@ impl NotebookEditor {
         }
     }
 
+    fn run_and_advance(&mut self, _: &RunAndAdvance, window: &mut Window, cx: &mut Context<Self>) {
+        if let Some(cell_id) = self.cell_order.get(self.selected_cell_index).cloned() {
+            if let Some(cell) = self.cell_map.get(&cell_id) {
```
MrSubidubi, line 581: Let's collapse this here again to remove the unneeded indentation.

crates/repl/src/notebook/notebook_ui.rs
MrSubidubi: Naively looking at this, it seems we are missing the change of mode here? Stuff like this is why I'd really like for some method here.

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -774,13 +848,18 @@ impl NotebookEditor {
...
             }
         })
         .detach();
 
-        self.cell_list.reset(self.cell_order.len());
+        self.cell_list.splice(insert_index..insert_index, 1);
+        self.cell_list.scroll_to_reveal_item(insert_index);
+        let editor = code_cell.read(cx).editor().clone();
+        window.focus(&editor.focus_handle(cx), cx);
+        self.notebook_mode = NotebookMode::Edit;
```
MrSubidubi, line 812: This also looks very very similar to tthe code above. Can we move this into a method too and have an enum that changes the editing state if needed?

crates/zed_actions/src/lib.rs
MrSubidubi, line 1: Curious, is there a good reason why we move these here in the first place?

MrSubidubi (changes requested):

crates/zed_actions/src/lib.rs
MostlyKIGuess: Yeah, while adding these on vim json, the test doesn't get these actions and it would fail all the vim tests.

I started doing this when I added vim support

MrSubidubi: Makes sense 👍 Thanks for the context and sorry for missing this.

MostlyKIGuess:
> This will close #51272 
> 
> Just a suggestion from my work on this before you pointed out to me you had already done it (thanks for catching me on this already being a WIP). it was suggested to me to have run and advance create a new cell if you run+advance the last cell, which is what other notebook editors do. 
> 
> I took a look at your code and I don't *think* your code currently does that, but with all the new machinery you added here I think it will be a pretty trivial addition.


That does make sense yes, I will add this. Thank you!

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -1216,9 +1252,14 @@ impl NotebookEditor {
 
 impl Render for NotebookEditor {
     fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        let mode_context = match self.notebook_mode {
+            NotebookMode::Command => "NotebookEditor notebook_mode=command",
+            NotebookMode::Edit => "NotebookEditor notebook_mode=edit",
+        };
+
```
MrSubidubi: Can we build ourselves a context here using the https://github.com/zed-industries/zed/blob/a17a1c10c87519676a65de7a46106520f13f22f3/crates/gpui/src/keymap/context.rs#L116-L134 APIs please?

crates/repl/src/notebook/notebook_ui.rs
MrSubidubi: This is better, however, the focusing still seems to be rather duplicate and similarly could perhaps even be moved into the same method?

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -624,7 +668,7 @@ impl NotebookEditor {
                 if let Some(cell) = self.cell_map.get(cell_id) {
                     match cell {
                         Cell::Code(code_cell) => {
-                            let editor = code_cell.read(cx).editor();
+                            let editor = code_cell.read(cx).editor().clone();
```
MrSubidubi: Do we need to clone here?

MrSubidubi (changes requested):

crates/repl/src/notebook/notebook_ui.rs
MostlyKIGuess: Thanks for this!

MrSubidubi (approved):
Nope, looks good, thank you and sorry for the delay!

PR #51195, opened 2026-03-10
editor: Account for horizontal scrollbar height in `SizeByContent` layout

### LINKED ISSUE #51177: Horizontal scroll overlaps last line in agent diff display block
### Reproduction steps

Ask agent edit an very long string, so you can see horizontal scroll when chat shows diff for changes changes

Changed string must be the last one it the diff

Current behaviour:
Image

Expected:
I can see last string, or at least I can scroll vertically to see it somehow.
Ideally - scroll height must be taken into account when height of block calculated

Zed: v0.226.4+stable.195.175cfdc40fa202d8baa9ff457ffd1733cd2b5f12 (Zed)
OS: Linux X11 fedora 42
Memory: 27.2 GiB
Architecture: x86_64
GPU: AMD Radeon 780M Graphics (RADV PHOENIX) || radv || Mesa 25.1.9

### Attach Zed log file

_No response_

Claude code agent

Fedora

Dnreikronos:
Closes #51177

### Before:
image

### After:
image

Release Notes:

- Fixed horizontal scrollbar overlapping the last line in agent diff display blocks


MrSubidubi (changes requested):
Hm I do not think that this is really the right approach - just because the scrollbar is enabled, we should not be unconditonally be reserving space for it. What we actually want here I'd argue is that we reserve the space when we need it but don't if we end up not needing it. This will make this harder to fix I believe, but would be the more proper fix.

Dnreikronos:
> Hm I do not think that this is really the right approach - just because the scrollbar is enabled, we should not be unconditonally be reserving space for it. What we actually want here I'd argue is that we reserve the space when we need it but don't if we end up not needing it. This will make this harder to fix I believe, but would be the more proper fix.

### Long line edited:
image


### Short line edited:
image


### If the Agent box is increased:
image



MrDubiDubi, as you can see, the horizontal scrollbar will be added just if the line edited is long enough that don't fit on the screen

MrSubidubi (changes requested):
Thanks for the follow-up! However, I sadly still do not think this is sustainable for the long term - after all, the longest line is not the singular factor to determine whether we have scrolling in an editor or whether we do not. There goes more into determining when to layout a scrollbar in the prepaint rendering code, and I think the path here is to find a sustainable way to determine properly, efficiently and reliably when we need a scrollbar and when we do not. The current approach will not at all work for most full-sized editors and might even break for the agent panel quickly, so I am not convinced this is a good path forward as-is

Dnreikronos:
> Thanks for the follow-up! However, I sadly still do not think this is sustainable for the long term - after all, the longest line is not the singular factor to determine whether we have scrolling in an editor or whether we do not. There goes more into determining when to layout a scrollbar in the prepaint rendering code, and I think the path here is to find a sustainable way to determine properly, efficiently and reliably when we need a scrollbar and when we do not. The current approach will not at all work for most full-sized editors and might even break for the agent panel quickly, so I am not convinced this is a good path forward as-is

Wassup! Thanks for the feedback on the previous approach.

I reworked this to stop duplicating scrollbar visibility logic in the layout phase. Instead of computing longest line width during layout and checking if it overflows, this now reads `last_position_map` that prepaint already stores on the `Editor` struct.

During prepaint, `scroll_max.x` comes from the full scroll width calculation (longest line, block widths, blame column, overscroll). That value lives in `PositionMap`, and the editor caches it as `last_position_map` after every prepaint. Other parts of the editor already use this field.

The layout function went from this:

```rust
let font_id = window.text_system().resolve_font(&style.text.font());
let font_size = style.text.font_size.to_pixels(window.rem_size());
let em_width = window.text_system().em_width(font_id, font_size).ok()?;
let gutter_dimensions = snapshot.gutter_dimensions(font_id, font_size, style, window, cx);
let editor_width = width - gutter_dimensions.width - gutter_dimensions.margin - 2. * em_width;

if horizontal_scrollbar_enabled {
    let longest_line_width = layout_line(snapshot.longest_row(), ...).width;
    if longest_line_width > editor_width {
        scroll_height += scrollbar_width;
    }
}
```

To this:

```rust
let had_horizontal_overflow = editor
    .last_position_map
    .as_ref()
    .is_some_and(|pm| pm.scroll_max.x > 0.01);

let scrollbar_enabled = editor.show_scrollbars.horizontal
    && EditorSettings::get_global(cx).scrollbar.axes.horizontal
    && EditorSettings::get_global(cx).scrollbar.show != ShowScrollbar::Never
    && !scrollbar_width.is_zero();

if had_horizontal_overflow && scrollbar_enabled {
    scroll_height += scrollbar_width;
}
```

For convergence on first render (when `last_position_map` is `None`), prepaint now checks if horizontal overflow state changed for `SizeByContent` editors:

```rust
self.editor.update(cx, |editor, cx| {
    let old_position_map = editor.last_position_map.replace(position_map.clone());
    if matches!(
        editor.mode(),
        EditorMode::Full { sizing_behavior: SizingBehavior::SizeByContent, .. }
    ) {
        let had_overflow = old_position_map
            .as_ref()
            .is_some_and(|pm| pm.scroll_max.x > 0.01);
        let has_overflow = position_map.scroll_max.x > 0.01;
        if had_overflow != has_overflow {
            cx.notify();
        }
    }
});
```

I would like to  use `> 0.01` instead of `> 0.` as a small epsilon to avoid triggering on floating point noise.

It only fires when the overflow state actually flips, so there's no loop. The codebase already does this with `last_bounds` for other cross-phase decisions.

So: removed the `layout_line()` call and all the font/gutter dimension calculations from layout, replaced with a boolean check against existing cached data. Future changes to how prepaint computes scroll width get picked up automatically.


Does you think that it's still not good enough? 


_Ps: the changes are just local on my machine. If you think that the approach looks good I will send them to remote._

MrSubidubi:
Thanks for the detailed explanation. If I am to understand you correctly, would this not result in the size computation always lagging one frame behind, especially during resizing operations? Since you use the `last_position_map`, I assume this would always be based on the last frame and not the current one, which I then assume would cause flickering, would it not?

I think at the core, the issue here is that we do way too much in the editor during prepaint, whereas we would probably need to do the actual layout computations during layout and not prepaint. Note that this is a short statement but a very big task/refactor and more a thought of mine where I'd appreciate your view on it, it might be that we need to stick with the current approach for now and just live with the tradeoff.

Other note, feel free to push changes up as a new commit, as this makes discussing this easier IMO. After all, you can always revert your commits if we decide it was not the right approach, but I think your comments above would just be as valuable as self-review comments here on the actual code.

Lastly, looking for a reviewer on the script PR right now, haven't found one yet but will nudge people again.

Dnreikronos:
> Thanks for the detailed explanation. If I am to understand you correctly, would this not result in the size computation always lagging one frame behind, especially during resizing operations? Since you use the `last_position_map`, I assume this would always be based on the last frame and not the current one, which I then assume would cause flickering, would it not?
> 
> I think at the core, the issue here is that we do way too much in the editor during prepaint, whereas we would probably need to do the actual layout computations during layout and not prepaint. Note that this is a short statement but a very big task/refactor and more a thought of mine where I'd appreciate your view on it, it might be that we need to stick with the current approach for now and just live with the tradeoff.
> 
> Other note, feel free to push changes up as a new commit, as this makes discussing this easier IMO. After all, you can always revert your commits if we decide it was not the right approach, but I think your comments above would just be as valuable as self-review comments here on the actual code.
> 
> Lastly, looking for a reviewer on the script PR right now, haven't found one yet but will nudge people again.

Hey, you're right about the one-frame lag. Since we read `last_position_map` from the previous frame, the first render after a resize will have stale scroll_max data. Prepaint then updates it and calls `cx.notify()`, so the second frame corrects itself. I tested resizing and didn't notice any flickering, but the theoretical lag is there:


I pushed the new approach as a commit so you can review the actual diff.

On the bigger question of moving layout computations out of prepaint and into the layout phase, yeah I think you're right that's where this should go long-term. Happy to take a crack at it for this PR if you think it's worth it, or we can land this and do the refactor separately. Up to you.

https://github.com/user-attachments/assets/e4faf6a0-7bc5-4083-b923-e8bac3a22ea8

_Ps: sorry for the lag on the video_

MrSubidubi:
Thanks for taking a stab at this and sorry for letting this sit in the queue for this long - I thought abut this from time to time and came to the conclusion that we first need to cleanup layout and such so we know during layout whether we need the space for a scrollbar as opposed to adding the hack this proposes. There will be flickering in a release build and for every editor that is added, so ultimately, I do not think this is a good approach for the issue at hand.

As for moving stuff to layout, thanks for offering, but I'd be unsure how feasible that is as an external contribution, since we are talking about the very core of the product.

In any case, thank you for looking into the issue and giving it a shot!

PR #51197, opened 2026-03-10
Add fuzzy search on remote projects

rohaquinlop:
Closes #35588

Added fuzzy search to the Remote Projects modal to filter servers and projects by name

Release Notes:

- N/A


crates/recent_projects/src/remote_servers.rs
```diff
@@ -2593,6 +2637,20 @@ impl RemoteServerProjects {
...
+        let query = self.search_editor.read(cx).text(cx);
+        let query_lower = query.to_lowercase();
+
+        let filtered_servers: Vec<RemoteEntry> = if query_lower.is_empty() {
+            state.servers.clone()
+        } else {
+            state
+                .servers
+                .iter()
+                .filter(|entry| matches_fuzzy_query(&entry.search_text(), &query))
```
SomeoneToIgnore, line 2649: We need to keep in mind that this is a part of the render flow, which may be run 120 times/second and we'd better try better.

* we have a `fuzzy` crate which allows looking up things asynchronously, on a background executor/thread — let's use that instead, as it also returns the indices matched which we'd better highlight (as e.g. outline panel's filter does)

Also, `matches_fuzzy_query` also allocates badly per each entry (due to `to_lowercase`) and we should not do that.

* we have to re-trigger this fuzzy matching as a separate task after each search buffer edit — that task fills and filters the resulting set to render, and, last step, replaces the old set with the new one.

The rendering code should always use this set instead and never calculate anything itself.

SomeoneToIgnore (changes requested):
Thank you, this looks rather accurately added.
I will ask for @\danilo-leal 's opinion later after we polish this, but it's fine overall.

I've commented that we need to redo the filtering part, but also what's important is to improve the usability:

* Currently, the panel does not seem to focus the input field when open initially — that seems rather confusing as I'd expect I can query right away, whenever the focus is

* Current filtering seems somewhat limiting as does not match project paths — can we change that to include them too?
Presumably, we want to show host name header too if any project path underneath got matched.

I've tried to capture those issues on the video:

https://github.com/user-attachments/assets/25bf9a62-6faa-4faa-b773-a741fa16c17d

PR #51219, opened 2026-03-10
editor: Fix inlay hint highlighting

### LINKED ISSUE #48141: Selecting `(` in function call also selects first inlay hint
### Reproduction steps

Use basedpyright with this snippet and select the `(` on the last line:

```python
def foo(num: int) -> None:
    pass


foo(1)
```

The inlay hint is associated with the argument and not the function call, so it should not be selected until the argument is.

Image

OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

### Relevant Zed settings

```json
{
    "inlay_hints": { "enabled": true, "show_background": true }
}
```

ayushk-1801:
Closes #48141

Release Notes:

- Fixed an issue where selecting text immediately before an inlay hint (e.g. the opening `(` in a function call) would incorrectly select the first inlay hint. Now, inlay hints are correctly selected only when the token they are attached to is selected.

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -1752,15 +1752,14 @@ impl DisplaySnapshot {
     }
 
     /// Converts a buffer offset range into one or more `DisplayPoint` ranges
-    /// that cover only actual buffer text, excluding any inlay hint text that
-    /// falls within the range.
-    pub fn isomorphic_display_point_ranges_for_buffer_range(
+    /// that cover the actual buffer text, including any logically selected inlay hint text.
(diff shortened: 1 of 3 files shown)


Veykril (commented):
I am not sure how this PR fixes thing? The issue is that the LSP only describes insert locations in the document, so hints effectively are not "attached" to anything in the document. In other words we can't fix this without fixing the atrocious LSP specification.

From what I can tell the changes here effectively make things work "as expected" for hints that logically precede text, but by doing so it now breaks the current setup which is hints succeeding text?

PR #51227, opened 2026-03-10
Add guix container transport

daym:
Guix shell containers are much more fine-grained than other devcontainers, and otherwise pretty similar to support in zed. So I implemented them.

Release Notes:


maxdeviant:
Thanks for taking the time to put this PR together.

This isn't something that we want to maintain in Zed core, at the moment. At some point in the future I think it would make sense to support devcontainer-like functionality through the extension system, which would allow for adding Guix support via an extension.

Going to close, for now.

PR #51236, opened 2026-03-10
editor: Add persistent undo/redo history

SpaceDux:
Hello,

I hope it's okay to raise this PR.

I love Zed, but one thing that really bugs me is that undo/redo history
doesn't survive closing a tab and editing another file. Zed already persists dirty buffer contents, but
the undo stack is wiped, so when you close a file and reopen it later, you
can't undo your previous edits. As someone who likes to keep a tidy workspace
and not leave dozens of tabs open, this means I regularly lose my edit
history.

_I have relied on Claude Code (along with GSD) to help get this over the line as Rust is not my language, although I have been learning it, I'm still not that confident syntactically or on best practices. I have purposely left the "with Claude Code" git author, as it deserves it's credit._

**Solution**
A new opt-in setting persist_history that serializes the undo/redo stack to
disk and restores it when a file is reopened.

**How it works**
- Write: On save, the buffer's undo/redo stack is serialized and written to a
blob file (keyed by a SHA-256 hash of the absolute path). A content hash and
mtime are stored in the DB to validate freshness on restore.
- Restore: When an editor is opened (via added_to_workspace or deserialize),
the blob is loaded and replayed into the buffer, but only if the file's
content hash still matches, preventing stale history from being applied to a
file that changed externally.
- Prune: During workspace cleanup, blobs for files that no longer exist on
disk are removed.
- Setting guard: All three paths (write, restore, prune) are gated behind
persist_history.enabled, which defaults to false.

Configuration

// settings.json
{
  "persist_history": {
    "enabled": true,
    "max_entries": 10000
  }
}

What's included

- DB migration adding persist_history table
- Blob serialization/deserialization of undo/redo stacks
- Content hash validation to avoid applying stale history
- max_entries cap to limit disk usage per file
- Pruning of orphaned history on workspace cleanup
- Settings UI integration
- Tests for the DB layer and editor integration

Thank you for reading, and I look forward to your reviews 😄

Release Notes:

- Added persist_history setting to preserve undo/redo history across tab close
 and restart


tommyming:
@SpaceDux are there any screenshots/demo videos that I can have a look to see what is happening here? Probably a comparsion, e.g. before-after photo set would be better.

Will help to have a look in the code later.

ChristopherBiscardi:
Thank you for the PR!

Considering a number of factors, such as familiarity with the language/codebase, the lack of requested screen recordings, etc on this PR: If we move forward with this feature I believe we'll go with the alternative PR #51357 instead, which is a significantly more concise PR for the feature.

PR #51238, opened 2026-03-10
Fix non-ASCII path:line:column navigation

### LINKED ISSUE #43329: zed cli column index broken on non ascii

loadingalias:
Closes #43329

## Summary
This fixes `path:line:column` navigation for files containing non-ASCII text.

Before this change, open path flows were passing the external column directly into `go_to_singleton_buffer_point`. That happened to work for ASCII, but it was wrong for Unicode because external columns are user-visible character positions while the editor buffer stores columns as UTF-8 byte offsets.

This PR adds a shared text layer conversion for external row/column coordinates and uses it in the affected open-path flows:
 - file finder navigation
 - recent project remote connection navigation
 - recent project remote server navigation

It also adds regression coverage for the Unicode case that originally failed.

As a small - necessary - prerequisite, this also adds `remote_connection` test support to `file_finder`'s dev-deps so the local regression test can build and run on this branch. That follows the same feature mismatch pattern previously fixed in #48280. I wasn't able to locally verify the tests/etc. w/o the addition... so I've rolled it into this PR. Tests are green.

The earlier attempt in #47093 was headed in the right direction, but it did not land and did not include the final regression coverage requested in review.

## Verification
 - `cargo fmt --all -- --check`
 - `./script/clippy -p text`
 - `./script/clippy -p file_finder`
 - `./script/clippy -p recent_projects`
 - `cargo test -p file_finder --lib --no-run`
 - `cargo test -p file_finder file_finder_tests::test_row_column_numbers_query_inside_file -- --exact`
 - `cargo test -p file_finder file_finder_tests::test_row_column_numbers_query_inside_unicode_file -- --exact`
 - `cargo test -p text tests::test_point_for_row_and_column_from_external_source -- --exact`

## Manual

I reproduced locally on my machine (macOS) w/ a stateless launch using a Unicode file (Cyrillic)

Before:
 - `:1:5` landed too far left
 - `:1:10` landed around the 4th visible Cyrillic character

After:
 - `:1:5` lands after 4 visible characters / before the 5th
 - `:1:10` lands after 9 visible characters / before the 10th

Release Notes:
 - Fixed `path:line:column` navigation so non-ASCII columns land on the correct character.

### DIFF
--- a/crates/file_finder/Cargo.toml
+++ b/crates/file_finder/Cargo.toml
@@ -28,7 +28,6 @@ picker.workspace = true
 project.workspace = true
 settings.workspace = true
 serde.workspace = true
-text.workspace = true
 theme.workspace = true
 ui.workspace = true
 util.workspace = true
--- a/crates/file_finder/Cargo.toml
+++ b/crates/file_finder/Cargo.toml
@@ -43,6 +43,7 @@ gpui = { workspace = true, features = ["test-support"] }
 
 picker = { workspace = true, features = ["test-support"] }
 pretty_assertions.workspace = true
+remote_connection = { workspace = true, features = ["test-support"] }
 serde_json.workspace = true
 theme = { workspace = true, features = ["test-support"] }
 workspace = { workspace = true, features = ["test-support"] }
--- a/crates/file_finder/Cargo.toml
+++ b/crates/file_finder/Cargo.toml
@@ -42,7 +42,6 @@ gpui = { workspace = true, features = ["test-support"] }
 
 picker = { workspace = true, features = ["test-support"] }
 pretty_assertions.workspace = true
-remote_connection = { workspace = true, features = ["test-support"] }
 serde_json.workspace = true
 theme = { workspace = true, features = ["test-support"] }
 workspace = { workspace = true, features = ["test-support"] }
--- a/crates/multi_buffer/src/multi_buffer.rs
+++ b/crates/multi_buffer/src/multi_buffer.rs
@@ -2141,7 +2141,7 @@ impl MultiBuffer {
             if point < start {
                 found = Some((start, excerpt_id));
             }
-            if point > end {
+            if point >= end {
                 found = Some((end, excerpt_id));
             }
         }
--- a/crates/multi_buffer/src/multi_buffer.rs
+++ b/crates/multi_buffer/src/multi_buffer.rs
@@ -2134,7 +2134,7 @@ impl MultiBuffer {
         for (excerpt_id, _, range) in self.excerpts_for_buffer(snapshot.remote_id(), cx) {
             let start = range.context.start.to_point(&snapshot);
             let end = range.context.end.to_point(&snapshot);
-            if start <= point && point <= end {
+            if start <= point && point < end {
                 found = Some((snapshot.clip_point(point, Bias::Left), excerpt_id));
                 break;
             }
--- a/crates/multi_buffer/src/multi_buffer.rs
+++ b/crates/multi_buffer/src/multi_buffer.rs
@@ -5188,6 +5188,16 @@ impl MultiBufferSnapshot {
         }
     }
 
+    pub fn point_for_row_and_column_from_external_source(&self, row: u32, column: u32) -> Point {
+        let row = row.min(self.max_point().row);
+        let range = Point::new(row, 0)..Point::new(row, self.line_len(MultiBufferRow(row)));
+        text::BufferSnapshot::point_for_column_in_range_from_external_source(
+            range.clone(),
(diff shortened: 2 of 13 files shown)


crates/file_finder/Cargo.toml
```diff
@@ -38,9 +38,10 @@ project_panel.workspace = true
 ctor.workspace = true
 editor = { workspace = true, features = ["test-support"] }
 gpui = { workspace = true, features = ["test-support"] }
-
+language = { workspace = true, features = ["test-support"] }
 picker = { workspace = true, features = ["test-support"] }
 pretty_assertions.workspace = true
+remote_connection = { workspace = true, features = ["test-support"] }
```
SomeoneToIgnore: Why is this needed?

SomeoneToIgnore (changes requested):
Any chance we can address the https://github.com/zed-industries/zed/pull/47093#pullrequestreview-3789958300 comment?

crates/file_finder/Cargo.toml
loadingalias: It’s only there to enable `remote_connection/test-support` in the `file_finder` test build. `file_finder`’s tests already pull in `workspace/project` test-support, which enables `remote/test-support` and adds `RemoteConnectionOptions::Mock()`. 

If I don't enable `remote_connection/test-support`, the matching `Mock()` arm in `remote_connection.rs` is cfg’d out, and `cargo test -p file_finder --lib --no-run` fails with a non-exhaustive match before the regression test can run.

crates/recent_projects/src/remote_connections.rs
```diff
@@ -458,7 +457,14 @@ pub fn navigate_to_positions(
                     active_editor.update(cx, |editor, cx| {
                         let row = row.saturating_sub(1);
                         let col = path.column.unwrap_or(0).saturating_sub(1);
-                        editor.go_to_singleton_buffer_point(Point::new(row, col), window, cx);
+                        let Some(buffer) = editor.buffer().read(cx).as_singleton() else {
+                            return;
+                        };
+                        let buffer_snapshot = buffer.read(cx).snapshot();
+                        let point =
```
SomeoneToIgnore: Curious: we seem to always have the `point` from a singleton buffer already, can we use `go_to_singleton_buffer_point` instead and remove `go_to_singleton_buffer_text_anchor` entirely?

crates/multi_buffer/src/multi_buffer.rs
```diff
@@ -5188,6 +5176,16 @@ impl MultiBufferSnapshot {
         }
     }
 
+    pub fn point_for_row_and_column_from_external_source(&self, row: u32, column: u32) -> Point {
```
SomeoneToIgnore:
```suggestion
    pub fn point_for_row_and_column_from_external_source(&self, row: MultiBufferRow, column: u32) -> Point {
```

Seems odd to have another public API exposure for a single `go_to_line.rs` case, can we inline this logic there and remove the corresponding tests?

crates/text/src/text.rs
```diff
@@ -2254,6 +2254,37 @@ impl BufferSnapshot {
...
+                point.column += character.len_utf8() as u32;
+            }
+        }
+
+        point
+    }
+
+    pub fn point_for_row_and_column_from_external_source(&self, row: u32, column: u32) -> Point {
+        let row = row.min(self.max_point().row);
+        let range = Point::new(row, 0)..Point::new(row, self.line_len(row));
```
SomeoneToIgnore: It feels somewhat odd to take the entire row, given how long strings can be.
We know `MAX_BYTES_IN_UTF_8` and can estimate that a certain `column` may span `[a, b]` search range only, which we can clip both sides and use, right?

SomeoneToIgnore (commented):
Great, thank you for covering all places.

A few minor things left to fix and we're ready to merge.

SomeoneToIgnore (approved):
Thank you!

PR #51246, opened 2026-03-11
Fix pane::RevealInProjectPanel to focus/open project panel for non-project buffers

### LINKED ISSUE #23967: pane::RevealInProjectPanel should focus/open project panel for non-project buffers
### Summary

Currently cmd-shift-e will *usually* open the project panel (either pane::RevealInProjectPanel or project_panel::ToggleFocus), but does not for non-project buffers

Steps to trigger the problem:
1. Open Settings (cmd-,) or an Untitled buffer (cmd-n)
2. Hit cmd-shift-e (bound to `pane::RevealInProjectPanel` in Editor)

Actual Behavior: Nothing happens

Expected Behavior: Project Panel is revealed.

https://github.com/zed-industries/zed/blob/fc6ee7b62ded4ff1b5e99b3654e4b046b0f003ab/assets/keymaps/default-macos.json#L626
https://github.com/zed-industries/zed/blob/fc6ee7b62ded4ff1b5e99b3654e4b046b0f003ab/assets/keymaps/default-macos.json#L505

Zed: v0.172.0 (Zed Nightly 9bf5e552338893a6ed6fff0cecb817ff5d2991b3)
OS: macOS 15.2.0
Memory: 64 GiB
Architecture: aarch64

prertik:
Closes #23967 

image

Release Notes:

- N/A


crates/workspace/src/pane.rs
```diff
@@ -4335,6 +4335,16 @@ impl Render for Pane {
...
+                    } else {
+                        let project = pane.project.clone();
+                        cx.spawn(async move |_, cx| {
+                            let _ = cx.update(|cx| {
+                                let _ = project.update(cx, |_, cx| {
+                                    cx.emit(project::Event::ActivateProjectPanel)
+                                });
+                            });
+                        })
+                        .detach();
```
dinocosta: I believe we can simply do a `pane.project.update` call here, just like the `if` block above 🤔

dinocosta (commented):
Hey @prertik 👋 

Thank you for taking a stab at this one! I'm not entirely sure if this is the correct fix, or if we even should update `RevealInProjectPanel` to simply open and focus on the Project Panel.

Right now, even though it's not perfect, at least we have this behavior where if the file is not part of the project, `pane: reveal in project panel` will have no effect, making it distinct from the scenario where the file is not part of the project. I wonder if, in case we actually want it to always focus on the project panel, we should at least shown a toast stating that the file is not part of the project so `RevealInProjectPanel` can't actually reveal the file in the project panel 🤔

crates/workspace/src/pane.rs
prertik: Will iterate on this based on the comment below: https://github.com/zed-industries/zed/pull/51246#issuecomment-4039374374

dinocosta (commented):
Hey @prertik !

Looks like your changes weren't actually working on my end, as even a file that doesn't belong to one of the open projects still gets a worktree and an entry. I've updated the code to actually display the toast if the worktree is not visible, when `pane: reveal in project panel` is used . I'm now checking with Zed's designer if this makes sense and asking for better copy 🙂

P.S.: I've just noticed I was thinking of existing files that are not in any of the open projects while you were thinking of unsaved buffers ! Will fix this tomorrow, sorry... 🫠

prertik:
> P.S.: I've just noticed I was thinking of existing files that are not in any of the open projects while you were thinking of unsaved buffers ! Will fix this tomorrow, sorry... 🫠

Thanks, @dinocosta.

Yes, I've been tackling the case of unsaved buffers in this PR, but if it can handle both cases, then I think it's better. 👍🏾

dinocosta (approved):
Hey @prertik ! 👋 

After chatting with the team, we've decided it makes more sense to simply show a notification when the buffer/file that the user is trying to reveal in the project panel doesn't actually belong to any of the open projects, instead of simply opening the project panel.

I've pushed some commits updating the behavior and extending the existing tests to cover this scenario. Will merge soon! 🚢

prertik:
> Hey @prertik ! There was some internal feedback that having `cmd-shift-e` not open the project panel but simply show a notification to be a little bit of an unexpected behavior because folks expect that binding to always open the project panel, in almost all contexts.
> 
> As such, I've opened #53539 but I've already updated the Preview Release Notes in order to ensure that you still get tagged as having helped with this. Thanks! 🙂



> Hey @prertik ! There was some internal feedback that having `cmd-shift-e` not open the project panel but simply show a notification to be a little bit of an unexpected behavior because folks expect that binding to always open the project panel, in almost all contexts.
> 
> As such, I've opened #53539 but I've already updated the Preview Release Notes in order to ensure that you still get tagged as having helped with this. Thanks! 🙂

Thanks, @dinocosta!

PR #51247, opened 2026-03-11
debugger: Fix restart only working once per session

nelsoncampos-cloudwalk:
`Session::restart_task` is set to `Some` when a restart is initiated but never cleared back to `None`. The guard at the top of `restart()` checks `self.restart_task.is_some()` and returns early, so only the first restart attempt succeeds.

This primarily affects debug adapters that advertise `supportsRestartRequest` dynamically via a `CapabilitiesEvent` after launch, such as the Flutter debug adapter.

Related: https://github.com/zed-extensions/dart/issues/45

- [ ] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) (N/A — no UI changes)

Release Notes:

- debugger: Fixed debug session restart only working once when the adapter supports DAP restart requests.

### DIFF
--- a/crates/project/src/debugger/session.rs
+++ b/crates/project/src/debugger/session.rs
@@ -2201,6 +2201,7 @@ impl Session {
                 } else {
                     cx.emit(SessionStateEvent::Restart);
                 }
+                session.restart_task = None;
             });
         }));
     }
--- a/crates/project/src/debugger/session.rs
+++ b/crates/project/src/debugger/session.rs
@@ -2187,29 +2187,27 @@ impl Session {
             self.capabilities.supports_restart_request.unwrap_or(false) && !self.is_terminated();
 
         self.restart_task = Some(cx.spawn(async move |this, cx| {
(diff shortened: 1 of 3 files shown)


crates/project/src/debugger/session.rs
Anthony-Eid: We should await this now that we're setting the `session.restart_task` to None. Otherwise it would be possible for us to send multiple restart requests before a response

Anthony-Eid (changes requested):

crates/project/src/debugger/session.rs
nelsoncampos-cloudwalk: Good call. Restructured to await the request before clearing restart_task.

Anthony-Eid (approved):
Thanks for catching this bug and making a fix! I added a regression test to prevent this from happening in the future. 

I'm going to enable auto merge 😀

PR #51253, opened 2026-03-11
workspace: Allow disabling tab cycles when using keyboard navigation

winzamark123:
Hi I've wanted this feature and looked through most open issues and pull request and couldn't find an implementation! Currently keymap of  `pane::ActivatePreviousItem` or `pane::ActivateNextItem` will cycle through the tab when it reaches the end (either the start or the beginning). This PR allows the user to set whether they want to "clamp" it so that it doesn't cycle. 

Heres a video of before and after when the keyboard short cut has been set. Let me know if this needs more clarifying.

https://github.com/user-attachments/assets/883b49b3-8721-419c-94dd-8268a450a7d8

## Summary

- Adds an optional `wrap_around` field to `ActivatePreviousItem` and `ActivateNextItem` pane actions
- When `wrap_around: false`, tab navigation stops at the first/last tab instead of wrapping circularly
- Defaults to `true`, fully preserving existing behavior

### Usage

Taking the current default macOS keymap, one could update the current behavior with the following bindings:

```json
{
    "context": "Pane",
    "use_key_equivalents": true,
    "bindings": {
      "alt-cmd-left": ["pane::ActivatePreviousItem", { "wrap_around": false }],
      "cmd-{": ["pane::ActivatePreviousItem", { "wrap_around": false }],
      "alt-cmd-right": ["pane::ActivateNextItem", { "wrap_around": false }],
      "cmd-}": ["pane::ActivateNextItem", { "wrap_around": false }],
    },
  }
```

## Test plan

- [x] Added `test_activate_item_with_clamp` in `crates/workspace/src/pane.rs` covering:
  - `clamp: true` at last tab stays on last tab
  - `clamp: true` at first tab stays on first tab
  - Default behavior (wrapping) unchanged
  - Normal mid-list navigation works with clamp
- [x] Existing tests pass (`cargo test -p workspace`, `cargo test -p tab_switcher`)
- [x] Tested locally: default keybindings still wrap, custom `clamp: true` bindings stop at boundaries

Release Notes:

- Added a `wrap_around` option to both `pane::ActivatePreviousItem` and `pane::ActivateNextItem` actions to optionally disable wrapping when cycling past the first or last tab.

### DIFF
--- a/crates/tab_switcher/src/tab_switcher_tests.rs
+++ b/crates/tab_switcher/src/tab_switcher_tests.rs
@@ -223,8 +223,8 @@ async fn test_close_selected_item(cx: &mut gpui::TestAppContext) {
     // 1.txt | [3.txt] | 2.txt | 4.txt
     //
     // With 3.txt being the active item in the pane.
-    cx.dispatch_action(ActivatePreviousItem);
-    cx.dispatch_action(ActivatePreviousItem);
+    cx.dispatch_action(ActivatePreviousItem::default());
+    cx.dispatch_action(ActivatePreviousItem::default());
     cx.run_until_parked();
 
     cx.simulate_modifiers_change(Modifiers::control());
--- a/crates/vim/src/command.rs
+++ b/crates/vim/src/command.rs
@@ -1659,9 +1659,9 @@ fn generate_commands(_: &App) -> Vec<VimCommand> {
             action.range.replace(range.clone());
             Some(Box::new(action))
         }),
-        VimCommand::new(("bn", "ext"), workspace::ActivateNextItem).count(),
-        VimCommand::new(("bN", "ext"), workspace::ActivatePreviousItem).count(),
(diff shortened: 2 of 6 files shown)


dinocosta (commented):
Hey @winzamark123 👋 

I've checked out your changes and everything seems to be working, implementation also seems fine and appreciate you including tests for the introduced changes! 

Will share with the team to see if there's an interest in introducing this. Will keep you posted 🙂

dinocosta (approved):
Hey @winzamark123! 👋

Sorry for the slow response, I’ve been busy with other pull requests. I didn’t get much feedback from the team, but since this is a fairly small change with a limited surface area for errors, I think we’re fine to go ahead and introduce it.

I’ve renamed the ‎`clamp` field to ‎`wrap_around`, as I think it’s a bit clearer from a user-facing perspective. 🙂

PR #51263, opened 2026-03-11
Revert "livekit_client: Apply correct priority for audio threads"

kubkon:
Reverts zed-industries/zed#51178


osiewicz (changes requested):
Just for the context: it does not seem like that PR caused issues Jakub saw with audio on his end. Hence this revert might not be warranted.

PR #51269, opened 2026-03-11
repl: Add notebook cell delete button

raideno:
Closes #51267

Before:
image

After:
image

Release Notes:

- Added a delete current cell button.

### DIFF
--- a/crates/repl/src/notebook/notebook_ui.rs
+++ b/crates/repl/src/notebook/notebook_ui.rs
@@ -62,6 +62,8 @@ actions!(
         AddMarkdownBlock,
         /// Adds a new code cell.
         AddCodeBlock,
+        /// Delete current cell.
+        DeleteCurrentCell,
         /// Restarts the kernel.
         RestartKernel,
         /// Interrupts the current execution.
@@ -784,6 +786,25 @@ impl NotebookEditor {
         cx.notify();
     }
 
+    fn delete_current_cell(&mut self, _window: &mut Window, cx: &mut Context<Self>) {
+        if self.cell_order.is_empty() {
+            return;
+        }
+
+        let cell_id = self.cell_order.remove(self.selected_cell_index);
+        self.cell_map.remove(&cell_id);
+
+        if !self.cell_order.is_empty() {
+            self.selected_cell_index =
+                self.selected_cell_index.min(self.cell_order.len() - 1);
+        } else {
+            self.selected_cell_index = 0;
+        }
+
+        self.cell_list.reset(self.cell_order.len());
+        cx.notify();
+    }
+
     fn cell_count(&self) -> usize {
         self.cell_map.len()
     }
@@ -1006,6 +1027,21 @@ impl NotebookEditor {
                                 })
                                 .on_click(|_, window, cx| {
                                     window.dispatch_action(Box::new(AddCodeBlock), cx);
+                                }),
+                            )
+                            .child(
+                                Self::render_notebook_control(
+                                    "delete-current-cell",
+                                    IconName::Trash,
+                                    window,
+                                    cx,
+                                )
+                                .disabled(self.cell_order.is_empty())
+                                .tooltip(move |window, cx| {
+                                    Tooltip::for_action("Delete current cell", &DeleteCurrentCell, cx)
+                                })
+                                .on_click(|_, window, cx| {
+                                    window.dispatch_action(Box::new(DeleteCurrentCell), cx);
                                 }),
                             ),
                     ),
@@ -1240,6 +1276,9 @@ impl Render for NotebookEditor {
             }))
             .on_action(
                 cx.listener(|this, _: &AddCodeBlock, window, cx| this.add_code_block(window, cx)),
+            )
+            .on_action(
+                cx.listener(|this, _: &DeleteCurrentCell, window, cx| this.delete_current_cell(window, cx)),
             )
             .on_action(cx.listener(|this, _: &MoveUp, window, cx| {
                 this.select_previous(&menu::SelectPrevious, window, cx);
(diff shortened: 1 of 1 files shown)


miguelraz:
@raideno there's a failing style check here.

raideno:
I'll try to add the confirmation button and then we can rerun the pipeline then. For the delete/hide cell output it should be an issue/pull request of its own I guess ?

raideno:
Should be all good now @miguelraz, I added a confirmation prompt and spaced the delete button from the rest and putted it in its own bar for destructive actions.

image
image

raideno:
@rgbkrk (sorry for the ping, let me know if it is annoying so I won't do it again), please let me know if everything's good whenever you have time, thanks!

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -784,6 +787,54 @@ impl NotebookEditor {
...
+
+            Ok(())
+        })
+        .detach_and_prompt_err("Failed to delete cell", window, cx, |e, _, _| {
+            Some(format!("{e}"))
+        });
+    }
+
+    fn delete_current_cell_confirmed(&mut self, _window: &mut Window, cx: &mut Context<Self>) {
+        if self.cell_order.is_empty() {
```
miguelraz, line 821: At first I thought this was a bit too defensive since we're duplicating the logic from just above is

```rust
        if self.cell_order.is_empty() {
            return;
        }
```

but I think it's actually a good idea in case we later scaffold functionality to delete several cells at once and we want to have this check fire without the prompt confirmation dialog every time. 

Nice thinking!

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -784,6 +787,54 @@ impl NotebookEditor {
...
+        let cell_id = self.cell_order.remove(self.selected_cell_index);
+        self.cell_map.remove(&cell_id);
+
+        if !self.cell_order.is_empty() {
+            self.selected_cell_index = self.selected_cell_index.min(self.cell_order.len() - 1);
+        } else {
+            self.selected_cell_index = 0;
+        }
+
+        self.cell_list.reset(self.cell_order.len());
```
miguelraz, line 834: I've noticed a present ~~bug~~ opportunity for future work because the scrollbar resets to the top of the document on every delete.

AFAIUI (and I'm likely wrong), since `.reset()` 

```rust
    pub fn reset(&self, element_count: usize) {
        let old_count = {
            let state = &mut *self.0.borrow_mut();
            state.reset = true;
            state.measuring_behavior.reset();
            state.logical_scroll_top = None;
            state.scrollbar_drag_start_height = None;
            state.items.summary().count
        };

        self.splice(0..old_count, element_count);
    }
```

sets

```rust
state.scrollbar_drag_start_heigh = None;
```

That then shoots up the scroll bar to the very top of the [document because of](https://github.com/zed-industries/zed/blob/680145064010e03be916683f302c59503734bdf3/crates/gpui/src/elements/list.rs#L516) 

```rust
            // if dragging the scrollbar, we want to offset the point if the height changed
            content_height - state.scrollbar_drag_start_height.unwrap_or(content_height);
```

The end result would be very annoying if you can only delete cells sequentially and they are at the very bottom of the document since you'd have to scroll to the bottom on every cell delete. 

Once the next functionality lands (😉) of selecting multiple cells and deleting them we may revisit this codepath because it's not a blocker for this PR.

crates/repl/src/notebook/notebook_ui.rs
```diff
@@ -1008,6 +1059,23 @@ impl NotebookEditor {
                                     window.dispatch_action(Box::new(AddCodeBlock), cx);
                                 }),
                             ),
+                    )
+                    .child(
+                        Self::button_group(window, cx).child(
```
miguelraz, line 1064: Color commentary:

I'm surprised the formatter doesn't pick up this `.child(` at the right as something to re-indent.

crates/repl/src/notebook/notebook_ui.rs
miguelraz: The above problem of resetting the scroll position seems to also happen when adding new cells.

miguelraz:
@raideno Overall I'm happy with the structure of the PR.

I pulled the changes down and confirmed that the deletion action happened correctly when confirming the dialog and was aborted when clicking "Cancel", both via the Command Palette `notebook: delete current cell` action and when selecting the 🗑️ icon on the top right.

Thank you for your patience with this review!

If you resolve the conflicts I'll be happy to merge.

spolisar:
@raideno thanks for opening this PR. It's good to see notebook support moving forward in Zed, it's the main thing I keep VS Code around for.

Currently in this PR, if a cell is deleted while it is running it will continue running in the background until it finishes, requiring a separate interrupt command. An interrupt should be sent with the deletion in that scenario. Here's my initial attempt at that using this branch and the changes from main as the base.

https://github.com/spolisar/zed/commit/579c86ed08623f382293af12c8b18fdbbe7eb5ae

An issue with my implementation is that it doesn't handle execution queueing properly. If 3 cells have been sent to the kernel for execution and the second cell is deleted, an interrupt and deletion will occur. If you test that same thing in VS Code, the second cell will just be removed from the execution queue. I haven't dug deep enough to know if that should be handled with an execution queue internal to Zed or if an execution request sent to a Jupyter kernel can be canceled.

I've noticed the `execution_requests` HashMap in notebook_ui.rs, but it doesn't look like an execution queue at first glance.

AlexElizard:
Similar functionality is also present in PR #58179

nobkd:
If I saw that correctly, this was superseded by #61295

MrSubidubi:
> If I saw that correctly, this was superseded by https://github.com/zed-industries/zed/pull/61295

That indeed seems to be the case, thanks @nobkd .

Apologies @raideno for skipping over this and instead going with the other PR.

That said, we'd be happy to incorporate any changes into the functionality that were discussed here but are not implemented in the change that landed. If someone wants to look into that, we'd appreciate it - and will definitely do our best to get back to you there then.

Thanks and sorrry for dropping the ball here!

PR #51280, opened 2026-03-11
acp: Update to `0.10.2`

bennetbo:
Updates to `0.10.2`, most notable change is implementing `session/close`.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/connection_view.rs
+++ b/crates/agent_ui/src/connection_view.rs
@@ -1498,14 +1498,12 @@ impl ConnectionView {
         let auth_method = connection.auth_methods().iter().find(|m| m.id() == &method);
 
         if let Some(terminal_auth) = auth_method
-            .and_then(|a| {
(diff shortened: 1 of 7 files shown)


crates/agent_ui/src/connection_view.rs
```diff
@@ -1495,9 +1495,16 @@ impl ConnectionView {
         let agent_telemetry_id = connection.telemetry_id();
 
         // Check for the experimental "terminal-auth" _meta field
-        let auth_method = connection.auth_methods().iter().find(|m| m.id == method);
+        let auth_method = connection.auth_methods().iter().find(|m| m.id() == &method);
 
         if let Some(terminal_auth) = auth_method
+            .and_then(|a| {
+                if let acp::AuthMethod::Agent(a) = a {
```
benbrandt: change to a match and return Some(a.meta) for all branches (because it is inexhaustive there will be a wildcard branch that returns None)

benbrandt (approved):

PR #51281, opened 2026-03-11
agent_ui: Defer thread restore until panel becomes active

vampik33:
Summary

  Fixes a race condition where focus_panel was called before updating the panel state,
  triggering set_active which could create a new default thread before the intended thread
   restore or action had a chance to run. Introduces a ThreadInitState state machine to
  defer thread restoration until the panel is actually activated.

  Changes

  Core mechanism — ThreadInitState state machine:
  - New ThreadInitState enum (Idle, PendingRestore, Creating) in agent_panel.rs tracks the
   panel's thread initialization lifecycle
  - During deserialization, thread restore is deferred as PendingRestore(AgentSessionInfo)
   instead of calling load_agent_thread_inner immediately
  - set_active now dispatches based on state: restores pending thread, creates new thread,
   or skips if creation is in-flight
  - Serialization preserves PendingRestore info so that panels serialized before
  activation still round-trip correctly

  Action handler reordering:
  - All action handlers in init() now call workspace.focus_panel::<AgentPanel>() after
  updating the panel, preventing set_active from racing with the intended action
  - Affected handlers: OpenHistory, OpenSettings, NewTextThread, NewExternalAgentThread,
  LoadThreadFromClipboard
  - Same reordering applied in acp_onboarding_modal.rs, claude_agent_onboarding_modal.rs,
  and main.rs

  State reset on all thread-creating paths:
  - new_text_thread, new_agent_thread_inner, external_thread, create_external_thread,
  load_thread_from_clipboard all reset state to Idle
  - external_thread sets state to Creating on entry and resets to Idle on error in the
  async path

  Tests (4 new):
  - test_new_text_thread_action_handler — verifies new text thread clears pending restore
  - test_new_external_agent_thread_action_handler — verifies new external thread clears
  pending restore
  - test_open_history_bypasses_pending_restore — verifies history view preserves pending
  restore for serialization
  - test_set_active_skips_restore_for_non_thread_views — verifies set_active doesn't
  restore when view is already initialized

  Impact

  - Thread restore is now lazy — the panel no longer restores the last active thread until
   the user actually opens the panel
  - Existing serialized panel test updated to verify two-phase restore: first
  PendingRestore state, then actual restore after set_active
  - No breaking API changes

  Testing

  - Verify agent panel restores the last active thread when opened (not before)
  - Verify creating a new thread (text or external) while a pending restore exists
  discards the pending restore
  - Verify opening history/settings before the panel is activated doesn't trigger thread
  restoration
  - Verify the panel correctly serializes pending restore state (close and reopen Zed
  before opening the panel)

  Release Notes:

  - Fixed a race condition where the agent panel could create a duplicate default thread
  instead of restoring the previous session's thread

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -9,7 +9,7 @@ use std::{
     time::Duration,
 };
 
-use acp_thread::{AcpThread, MentionUri, ThreadStatus};
+use acp_thread::{AcpThread, AgentSessionInfo, MentionUri, ThreadStatus};
 use agent::{ContextServerRegistry, SharedThread, ThreadStore};
 use agent_client_protocol as acp;
 use agent_servers::AgentServer;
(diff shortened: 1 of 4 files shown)


benbrandt:
Given we vastly changed the agent panel with parallel agents, goign to close. If you are still experiencing this and want to, feel free to open a new pr

PR #51321, opened 2026-03-11
gpui_linux: Add global menu

### LINKED ISSUE #14580: Please add global menu in Linux
### Check for existing issues

- [X] Completed

### Describe the feature

Please add global menu in Linux version, like MacOS in KDE, its very cool feature, i vote for add check box for make this menu, and thx for cool app!

Screenshot_20240716_190810

nihalxkumar:
This PR implements native Global Menu support via D-Bus (`com.canonical.dbusmenu`) on Linux. It allows the editor's menu bar to be exported and rendered natively by desktop environments like KDE Plasma, Unity, or standalone panels (e.g., Waybar) across both X11 and Wayland.

### How to Review This PR

Given the size of this feature, I recommend reviewing the files in the following order to best understand the architecture:

**1. The Core Server**: `crates/gpui_linux/src/linux/dbusmenu.rs` is the heart of the PR. It implements the `com.canonical.dbusmenu` D-Bus interface.
*   **Understand the separation between the internal MenuState** L159-162 (protected by a `Mutex`) and the asynchronous communication channels.
*   **Focus on Threading spawn_dbus_menu_thread L206-316**: Notice how the D-Bus connection is pushed onto its own background thread. The main thread sends commands (`LayoutUpdated`, `Shutdown`) to this thread to emit signals without blocking the GPUI loop.
*   **Focus on Reentrancy setup_global_menu_sources L21-122**: Review the `app_menu_action` callback. Pay close attention to the `take-invoke-restore` pattern used. This prevents a `RefCell` reborrow panic if resolving an action synchronously triggers another menu mutation.
*   **Focus on Error Handling**: Review the about_to_show L625-656 and about_to_show_group L658-694 methods. Notice the specific `log::debug!()` statements for channel timeouts and lock poisoning, which are critical for diagnosing missing menu drop-downs in edge cases.

**2. Wayland Lifecycle:** `crates/gpui_linux/src/linux/wayland/client.rs`
*   **Focus on Initialization**: Look at how the `appmenu` registry is bound and how a unique DBusMenuServer L165-172 is instantiated and attached to Wayland surfaces via KDE's unofficial `appmenu.xml` protocols.
*   **Focus on `Drop` Semantics**: Review `impl Drop for WaylandClient`. It's critical that we send a `Shutdown` command and mark_disconnected L399-407 so we cleanly terminate the background D-Bus thread and avoid leaking Zombie DBus objects. Notice the fallback warning when state cannot be borrowed.

**3. X11 Lifecycle**: crates/gpui_linux/src/linux/x11/client.rs
*   **Focus on Window Properties**: Review how we set the `_KDE_NET_WM_APPMENU_SERVICE_NAME` and `_KDE_NET_WM_APPMENU_OBJECT_PATH` X11 properties to broadcast the menu's D-Bus address to the desktop environment.
*   **Focus on `Drop`**: Similar to Wayland, verify the strict shutdown routines to ensure X11 doesn't hang onto the properties after the window is destroyed.

**4. The Integration**: `crates/gpui_linux/src/linux/platform.rs`
*   Review `is_global_menu_active()` and how GPUI queries the connected state to decide whether to render the internal fallback hamburger menu or hide it in favor of the OS panel.

---

### Key Implementation Details

*   **Spec Compliance**: The `GetGroupProperties` handler explicitly honors the D-Bus spec by returning all item properties when passed an empty list. Error types correctly differentiate between `InvalidArgs` (unknown item IDs) and `UnknownProperty` (property doesn't exist).
*   **Strict Synchronization**: AboutToShow L142-145 is fully synchronized. It calculates the exact intersection of requested ids L599-620 vs actually changed layout IDs to avoid unnecessarily blasting full layout invalidations over D-Bus, which causes flickering on KDE Plasma.
*   **Simplified Topology**: Instead of managing complex per-window DBus object exports, we now export a single `/MenuBar` object per session and safely route events to the currently focused context.
*   **Toggle Checkmarks**: `checkable` menu items seamlessly translate into D-Bus `toggle-type=checkmark` properties with `toggle-state=1/0`.

---

### How to Test

1. Run a Linux build with the feature toggled/enabled (e.g. `ZED_GLOBAL_MENU=1 cargo run -p zed --features global-menu`).
2. **KDE Plasma/GNOME (with AppMenu extension):** Add a global menu widget to your top panel.
3. Verify that the window's menus populate dynamically in the native panel.
4. Verify sub-menu navigation, toggleable items, and shortcuts trigger the corresponding Zed commands.
5. Quickly open/close the app/windows multiple times to ensure `Drop` correctly unregisters the D-Bus service without memory leaks or stray window coordinates.
6. Open complex modals (like settings or new windows) directly from the global menu to ensure no `RefCell` reborrow panics occur.

Closes #14580 

Release Notes:

- Add Global menu in Linux on X11 and Wayland.


nihalxkumar:
gentle reminder @probably-neb

Cargo.toml
```diff
@@ -915,9 +914,6 @@ debug = "limited"
 lto = "thin"
 codegen-units = 1
 
-[profile.release.package]
-zed = { codegen-units = 16 }
-
```
probably-neb, line 960: ?
This seems accidental.

crates/gpui_linux/Cargo.toml
```diff
@@ -14,6 +14,7 @@ path = "src/gpui_linux.rs"
 [features]
 default = ["wayland", "x11"]
 test-support = ["gpui/test-support"]
+global-menu = ["dep:async-channel", "dep:zbus", "dep:wayland-scanner"]
```
probably-neb: I don't think we should have the cargo feature flag. Keep things simple and always compile it in, and trust our detection/configuration of when to use this when appropriate

crates/gpui_linux/src/linux/wayland/appmenu.xml
probably-neb, line 1: we should use our existing `wayland_protocols_plasma` dep instead of `wayland_scanner` + xml files

probably-neb (changes requested):
Great work! Thanks for taking the time to do this. 

A couple notes I'm seeing so far
1. Need to fix the merge conflicts (that's my bad, sorry for taking so long to review)
2. we should use `wayland_protocols_plasma` instead of `wayland_scanner` + xml files
3. I don't think we should have the cargo feature flag. Keep things simple and always compile it in, and trust our detection/configuration of when to use th

Cargo.toml
nihalxkumar: yeah

crates/gpui_linux/Cargo.toml
nihalxkumar: I thought it helped in the WIP process.. Removed

crates/gpui_linux/src/linux/wayland/appmenu.xml
nihalxkumar: Done

nihalxkumar:
@probably-neb ready for review

crates/gpui_linux/src/linux/dbusmenu.rs
```diff
@@ -0,0 +1,1249 @@
...
+                    Some((*id, action, previous_enabled))
+                })
+                .collect()
+        };
+
+        let mut updated_props: Vec<(i32, HashMap<String, OwnedValue>)> = Vec::new();
+        let mut updates: Vec<(i32, bool, OwnedValue)> = Vec::new();
+
+        for (id, action, previous_enabled) in items_to_check {
+            let enabled = validate(action.as_ref());
```
probably-neb, line 514: Will this reverse the disabled state? If the server tells us an item is not enabled, should we re-enable it because it validates?

crates/gpui/src/platform/app_menu.rs
```diff
@@ -269,6 +276,9 @@ pub enum OwnedMenuItem {
         /// See [`OsAction`] for more information
         os_action: Option<OsAction>,
 
+        /// Whether this action is checkable
+        checkable: bool,
```
probably-neb, line 280: I don't see where this is ever set to true, and It looks like the menu only writes checks when it's true?

crates/gpui_linux/src/linux/platform.rs
```diff
@@ -215,6 +217,15 @@ impl<P: LinuxClient + 'static> Platform for LinuxPlatform<P> {
...
 
+    fn is_global_menu_active(&self) -> bool {
+        self.inner.with_common(|common| {
+            common
+                .dbus_menu_server
+                .as_ref()
+                .is_some_and(|server| server.is_connected())
+        })
+    }
+
```
probably-neb, line 228: A bit concerned that it doesn't seem like we are responsive to changes in the server being active. I'm having a hard time tracking how we ensure the local menu dissapears when the dbus menu server connects, and how we ensure the local menu returns and dbus menu goes away when the server disconnects

crates/gpui_linux/src/linux/dbusmenu.rs
```diff
@@ -0,0 +1,1249 @@
...
+        .and_then(|value| value.split('-').next())
+        .unwrap_or_default()
+        .trim()
+        .to_ascii_lowercase();
+
+    matches!(
+        language.as_str(),
+        "ar" | "he" | "iw" | "fa" | "ur" | "ps" | "dv" | "ku" | "sd" | "ug" | "yi"
+    )
+}
```
probably-neb, line 1249: Do we have a better way to do this? Is this really the first time in GPUI we're detecting RTL locales? I suspect not

crates/gpui_linux/Cargo.toml
```diff
@@ -72,6 +72,8 @@ strum.workspace = true
 url.workspace = true
 util.workspace = true
 uuid.workspace = true
+async-channel.workspace = true
+zbus = "5"
```
probably-neb, line 76: we should make these optional dependencies like the other wayland/x11 specific deps below, and enable them in the x11 and wayland features. Otherwise the headless remote server will need to compile these for no reason

probably-neb (changes requested):
Sorry for the delayed review (again) 
I hope you understand the size of the PR means it necessitates a decent chunk of time to sit down and begin to wrap my head around.

Besides the comments below (most of which are genuine questions, so feel free to disagree or point out what I'm missing), I'm a bit concerned by the lack of tests. I think we should take the time to setup a sort of fake server abstraction for testing, so that we can test different (async!) sequences of messages across the protocol. We do similar things for our tests elsewhere in Zed that involve http requests. Besides making sure the system is working as expected in it's current state, having a testing system setup will also come in handy in the future if we find any issues, because we'll be able to create re-producer tests for message sequences we aren't handling correctly

crates/gpui_linux/Cargo.toml
nihalxkumar: I have been busy. Will address the reviews 🙏

PR #51327, opened 2026-03-11
agent_ui: Add timestamp to thread item in the sidebar

danilo-leal:
Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/sidebar.rs
+++ b/crates/agent_ui/src/sidebar.rs
@@ -1180,11 +1180,37 @@ impl Sidebar {
 
         let id = SharedString::from(format!("thread-entry-{}", ix));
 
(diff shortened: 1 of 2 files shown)


crates/agent_ui/src/sidebar.rs
```diff
@@ -1180,11 +1180,37 @@ impl Sidebar {
 
         let id = SharedString::from(format!("thread-entry-{}", ix));
 
+        let timestamp = thread
```
mikayla-maki, line 1183: Q: should this timestamp be updated_at or created_at? I'm of two minds here. I think right now we sort by updated at, which means threads jump around. I kind of want to sort by created at but show updated at. Not 100% sure. Thoughts?

PR #51355, opened 2026-03-12
python: Add `label_for_symbol` for ty adapter

lingyaochu:
Ported `label_for_symbol` logic directly from the basedpyright adapter without adjustments.

Given Python's dynamic nature, the current implementation provides sufficient coverage. No further modifications are needed for now.

Release Notes:

- Fixed missing syntax highlighting in symbol search when using the ty language server.

### DIFF
--- a/crates/languages/src/python.rs
+++ b/crates/languages/src/python.rs
@@ -255,6 +255,41 @@ impl LspAdapter for TyLspAdapter {
         ))
     }
 
+    async fn label_for_symbol(
+        &self,
+        symbol: &language::Symbol,
+        language: &Arc<language::Language>,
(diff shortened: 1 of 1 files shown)


MrSubidubi (changes requested):
Thank you!

> Ported label_for_symbol logic directly from the basedpyright adapter without adjustments.

That sounds somewhat reasonable, but can we not move this into a common function then that we call for both of the adapters? Otherwise we have to keep the code in sync and that does not sound too nice

MrSubidubi:
That sounds awesome, thank you!

MrSubidubi (approved):
Nice find, nice cleanup and very nice PR. Thank you for submitting this and the quick follow-up!

PR #51358, opened 2026-03-12
remote: Reuse existing WSL connections across workspaces

JuanIPlanes:
## Problem

Every time a user runs `zed .` from a WSL distro, Zed performs a full reconnect sequence — shell detection, platform detection, binary version check — showing a "Connecting remote…" modal. This happens even when another Zed workspace is already open against the same WSL instance.

## Solution

Introduce a `Persistent` state in `ConnectionPool` that holds a strong `Arc<dyn RemoteConnection>` for WSL connections. Second and subsequent `connect()` calls for the same host reuse the live connection directly, skipping the reconnect sequence entirely.

## Changes

- `ConnectionPoolEntry::Persistent { connection, workspace_count }` — new enum variant holding a ref-counted live connection
- `should_persist()` — returns `true` for WSL connections only; SSH and Docker follow the existing non-persistent path
- Fast-path in `connect()` — reuses a `Persistent` entry and increments `workspace_count`
- Ref-counted `release()` — drops the strong `Arc` only when the last workspace closes
- `evict_if_persistent()` — removes stale entries when the WSL proxy exits with exit code 90
- `on_app_quit` hook — releases all `Persistent` entries on shutdown
- UI: "Disconnect Persistent Session" action in the Remote Servers panel, visible for WSL entries
- 21 unit tests covering all new behaviors

## Out of Scope (Phase 2)

Pre-warming the WSL server before first use ("shadow server" mode) is not included in this PR.

Release Notes:

- Improved WSL remote connection performance by reusing existing connections across workspaces

### DIFF
--- a/crates/recent_projects/src/remote_servers.rs
+++ b/crates/recent_projects/src/remote_servers.rs
@@ -499,11 +499,30 @@ impl ProjectPicker {
                         })
                         .log_err()?;
 
+                    let connection_opts_for_release = connection.clone();
                     let items = open_remote_project_with_existing_connection(
                         connection, project, paths, app_state, window, cx,
                     )
                     .await
                     .log_err();
+
+                    // Register an on_release hook so that closing the workspace window
+                    // decrements the ConnectionPool workspace_count (or removes the
+                    // persistent entry when the count reaches zero).  Without this,
+                    // workspace_count grows monotonically because release() is only
+                    // called from the "Disconnect" UI button or on_app_quit.
+                    window
+                        .update(cx, |multi_workspace, _, cx| {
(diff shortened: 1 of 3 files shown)


JuanIPlanes:
@cla-bot check

Veykril (commented):
Do we need to introduce a new `ConnectionPoolEntry`? I am not a fan of the persistence handling here, imo it should be possible to implement this in a simpler manner by just iterating the current `connected` instances, looking for a wsl instance for the given distro / host (having connections expose such a lookup "key"), and if that connection is still alive derive things from there.

Veykril:
Closing this as I do not believe this to be the right approach

PR #51370, opened 2026-03-12
Fix Tree-sitter link in documentation

hferreiro:
Release Notes:

- N/A


MrSubidubi (approved):
Nice find, thank you for seeing and fixing this!

PR #51389, opened 2026-03-12
edit_prediction: Expose `allow_data_collection` in settings

### LINKED ISSUE #48394: Zed AI "Training Data Collection" isn't exposed in settings.json
### Reproduction steps

1. Grep for `training` or `collection` in the default settings.json

Zed exposes basically all settings in settings.json, including settings for 3rd-party edit prediction providers.

However, the "Training Data Collection" setting is only exposed in the Edit Prediction menu. It should be exposed in settings.json too.

OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

oliverbarnes:
Closes #48394

Moves the data collection preference for Zed's Edit Predictions out of the internal KV store and into `settings.json` as a proper `allow_data_collection` setting under `edit_predictions`.

**Migration:** Existing users' choices are preserved. When `allow_data_collection` is absent from `settings.json`, the resolved value falls back to the legacy KV entry (`zed_predict_data_collection_choice`). Once the user toggles the setting or sets it explicitly, the new setting takes precedence and the KV entry is ignored.

**Bug fixed:** The original implementation of `toggle_data_collection` read the raw (unresolved) settings content to determine the current state. When `allow_data_collection` was absent from `settings.json` but the KV store held `"true"`, the raw read returned `None → false`, causing the first toggle click to write `Some(true)` (re-enabling) instead of `Some(false)` (disabling). The fix reads the resolved `is_data_collection_enabled()` value before entering the `update_settings_file` closure.

## Manual testing

**Setting takes effect:**
1. Open settings (`cmd+,`) and add `"allow_data_collection": true` under `edit_predictions`. Save.
2. Open a file — the data collection indicator in the editor should reflect the enabled state.
3. Flip to `false` and confirm it updates.

**Toggle correctly disables from KV-enabled state (migration bug fix):**
1. Remove `allow_data_collection` from `settings.json`.
2. Write the legacy KV entry directly:
   ```
   sqlite3 ~/Library/Application\ Support/Zed/db/0-dev/db.sqlite \
     "INSERT OR REPLACE INTO kv_store(key,value) VALUES('zed_predict_data_collection_choice','true');"
   ```
3. Restart Zed. The data collection toggle should show as **enabled** (reading from KV store).
4. Click the toggle once to disable. `allow_data_collection` should appear as `false` in `settings.json` — not `true`, which was the pre-fix behaviour.

**Upsell modal still appears for new users:**
1. Clear both KV keys and restart:
   ```
   sqlite3 ~/Library/Application\ Support/Zed/db/0-dev/db.sqlite \
     "DELETE FROM kv_store WHERE key IN ('zed_predict_data_collection_choice','dismissed-edit-predict-upsell');"
   ```
2. Open any file so the status bar is visible.
3. Click the edit prediction button (bottom-right status bar) — it should have a muted dot indicator.
4. The upsell modal should appear. Dismissing it should prevent it from reappearing.

## Release Notes:

- `allow_data_collection` for Zed's Edit Predictions can now be set explicitly in `settings.json` under `edit_predictions`. Existing preferences stored in the internal database are preserved as a fallback.


crates/language/src/language_settings.rs
```diff
@@ -476,6 +476,14 @@ pub struct EditPredictionSettings {
...
     pub enabled_in_text_threads: bool,
     pub examples_dir: Option<Arc<Path>>,
+    /// Whether training data collection is enabled in user settings.
+    ///
+    /// Unlike most resolved settings fields, this is kept as `Option<bool>` rather than
+    /// being unwrapped to a plain `bool`. The distinction between `None` (absent from the
+    /// user's settings file) and `Some(false)` (explicitly disabled) is needed by
+    /// `is_data_collection_enabled`: when the field is `None` it falls back to the legacy
+    /// KV store entry so that existing users' choices are preserved without a migration.
+    pub allow_data_collection: Option<bool>,
```
probably-neb: We do not want to do this this way. Settings are supposed to be `Option<T>` in  the content types to disambiguate between unset and set explicitly to `null` both for the purposes of merging, and displaying settings in the settings editor.

Instead we should have an enum for this value, for example

```
pub enum EditPredictionDataCollectionChoice {
  /// Default: No data collection, unless you have previously configured it to true in the kv store
  Default,
  /// No data collection or training
  No,
  /// Allow data collection and training on collected data
  Yes,
}
```

And add an entry to `default.json` for the default value

assets/settings/default.json
```diff
@@ -1600,6 +1600,10 @@
     // Whether edit predictions are enabled when editing text threads in the agent panel.
     // This setting has no effect if globally disabled.
     "enabled_in_text_threads": true,
+    // Whether to allow Zed to collect training data when using Zed's Edit Predictions.
+    // Data is only captured when the project is detected as open source.
+    //
+    // "allow_data_collection": false
```
probably-neb: This should not be commented out. All settings should have a default value for displaying in the settings UI

probably-neb (changes requested):
The problem I outline in my review comments is quite subtle and poorly documented, so I'm not surprised you missed it. But all settings should have a default value for displaying in the settings UI. Relatedly, I think this PR should add this setting to the settings UI under the edit predictions section. We don't need to handle showing the true value based on the KV store in the settings UI, using the solution I recommend in my review comments we just need to document the default enum value as meaning "infer from KV, else false"

assets/settings/default.json
```diff
@@ -1620,6 +1620,14 @@
       "prompt_format": "infer",
       "max_output_tokens": 64,
     },
+    // Controls whether Zed may collect training data when using Zed's Edit Predictions.
+    // Data is only captured when the project is detected as open source.
+    // Possible values:
+    //   - "default": use the preference previously set via the status-bar toggle,
+    //     or false if no preference has been stored.
+    //   - "yes": always allow data collection.
```
probably-neb: probably worth re-iterating on this line that "yes" doesn't mean _always_ collect data. It means collect data when in open source repos, potentially with a link to the policy in the bigger comment above

crates/edit_prediction/src/edit_prediction_tests.rs
```diff
@@ -3318,6 +3322,244 @@ async fn test_edit_prediction_settled(cx: &mut TestAppContext) {
...
+        .unwrap();
+
+    cx.update(|cx| {
+        assert!(ep_store.read(cx).is_data_collection_enabled(cx));
+    });
+
+    cx.update(|cx| KeyValueStore::global(cx))
+        .delete_kvp(ZED_PREDICT_DATA_COLLECTION_CHOICE.into())
+        .await
+        .unwrap();
```
probably-neb: why do we delete and not assert anything? Is the `unwrap` here load bearing?

crates/edit_prediction/src/edit_prediction_tests.rs
```diff
@@ -3318,6 +3322,244 @@ async fn test_edit_prediction_settled(cx: &mut TestAppContext) {
...
+    cx.update(|cx| {
+        cx.set_staff(true);
+        assert!(ep_store.read(cx).is_data_collection_enabled(cx));
+    });
+}
+
+// When a user had data collection enabled via the legacy KV store (with no explicit
+// setting in settings.json), toggle_data_collection must read the *resolved* state
+// (true) and write Some(false). The old code read the raw settings content
+// (None → false) and would have incorrectly written Some(true) instead.
```
probably-neb: I assume the 'old code' here is the code previously in this PR? No use keeping the comment

crates/edit_prediction/src/edit_prediction_tests.rs
```diff
@@ -3318,6 +3322,244 @@ async fn test_edit_prediction_settled(cx: &mut TestAppContext) {
...
+    cx.update(|cx| {
+        assert!(
+            ep_store.read(cx).is_data_collection_enabled(cx),
+            "data collection should be enabled via KV store before toggle"
+        );
+    });
+
+    // Simulate what toggle_data_collection does: capture the resolved current
+    // state, then write its inverse. With the fix, is_currently_enabled = true,
+    // so allow_data_collection becomes Some(No).
```
probably-neb: Same problem as the above comment. "the fix" is in this PR. No need to document it

crates/edit_prediction/src/edit_prediction_tests.rs
```diff
@@ -3318,6 +3322,244 @@ async fn test_edit_prediction_settled(cx: &mut TestAppContext) {
...
+}
+
+#[gpui::test]
+async fn test_upsell_shown_by_default(cx: &mut TestAppContext) {
+    init_test(cx);
+    let kvp = cx.update(|cx| KeyValueStore::global(cx));
+    kvp.delete_kvp(ZED_PREDICT_DATA_COLLECTION_CHOICE.into())
+        .await
+        .ok();
+    kvp.delete_kvp("dismissed-edit-predict-upsell".into())
```
probably-neb: We should either use the `ZedPredictUpsell::KEY` here or have a different constant like `ZED_PREDICT_DATA_COLLECTION_CHOICE`

crates/settings_content/src/language.rs
```diff
@@ -182,6 +182,14 @@ pub struct EditPredictionSettingsContent {
     pub open_ai_compatible_api: Option<CustomEditPredictionProviderSettingsContent>,
     /// The directory where manually captured edit prediction examples are stored.
     pub examples_dir: Option<Arc<Path>>,
+    /// Controls whether Zed may collect training data when using Zed's Edit Predictions.
+    /// Data is only ever captured for files in projects that are detected as open source.
+    ///
+    /// - `"default"`: use the preference previously set via the status-bar toggle,
+    ///   or false if no preference has been stored.
+    /// - `"yes"`: always allow data collection.
```
probably-neb: should have same changes as the description in `default.json`

crates/settings_ui/src/page_data.rs
```diff
@@ -9265,6 +9265,32 @@ fn edit_prediction_language_settings_section() -> [SettingsPageItem; 4] {
             files: USER,
             render: render_edit_prediction_setup_page
         }),
+        SettingsPageItem::SettingItem(SettingItem {
+            title: "Data Collection",
+            description: "Controls whether Zed may collect training data when using Zed's Edit Predictions. The default value uses the preference previously set via the status-bar toggle, or false if no preference has been stored.",
```
probably-neb: should state that data collection only happens in open source repos here. Users may only ever see the description in settings UI

probably-neb (changes requested):

oliverbarnes:
Thanks @probably-neb, latest comments addressed.

probably-neb (approved):

probably-neb:
Thanks for all your work on this @oliverbarnes!
I've made a couple cleanup/fixes, the first being making sure we check the organization filter in the ep store (added since this PR was opened), and caching the KV read so we aren't blocking edit prediction requests on it.

probably-neb (approved):

PR #51403, opened 2026-03-12
outline_panel: Add breadcrumbs when searching for symbols

jmg-duarte:
The symbol/outline panel is very confusing when searching for symbols across multiple files (for in search buffers for example) this PR aims to add breadcrumb-like entries to the list to provide more context on where each symbol is located.

Before:
Screenshot 2026-03-12 at 17 19 06

After:
Screenshot 2026-03-12 at 17 18 34

I searched for different symbols hence the difference in results 😅 

Release Notes:

- Added file breadcrumbs when searching for symbols in the outline panel

### DIFF
--- a/crates/outline_panel/src/outline_panel.rs
+++ b/crates/outline_panel/src/outline_panel.rs
@@ -2409,7 +2409,12 @@ impl OutlinePanel {
             FsEntry::File(FsEntryFile {
                 worktree_id, entry, ..
             }) => {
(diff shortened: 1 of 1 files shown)


jmg-duarte:
> "already in main", the current filtering query applies to directories, file names and search results simultaneously:

That's interesting, that wasn't my experience at all; with this clarifying experience I now agree that there's something to be improved on ergonomics / how to show the info

I'll keep thinking, thanks for the replies!

Issue #16828, opened 2024-08-25
Bookmarks / Manual, long-term multi-bufs


elkowar:
### Check for existing issues

- [X] Completed

### Describe the feature

When trying to understand larger codebases, you often come across a wide range of places in the code that are relevant to the problem you're trying to solve. It often happens that you loose track of where which function was, or that relevant functions are spread out across a single file, making it very hard to get an overview over the code.

I propose the following solution, uniquely well suited to Zed's multibuffer system:
**Bookmarks**.

Imagine the following workflow:
You find a function you want to remember and go back to later. You select the function, and hit a keybind, which adds it to your _bookmarks_. The bookmarks are a special, manually populated multibuffer. Optimally, the excerpts are Treesitter-aware, so you _actually_ just bookmark the function, rather than a specific range of line-numbers.
Now, you continue reading code and searching for different parts of the logic you want to dive into, bookmarking all relevant bits of code on the way.

These bookmarks are stored in a file inside the project directory (i.e. some `.zed/bookmarks.json`).

Now, once you think you found everything, you open the bookmarks buffer. You then see all the functions you care about for this project, nicely listed one after another. You can then step into more details of each function, or start changing stuff, directly inside the list of bookmarks. You can also discard bookmarks when you realize they don't actually matter.

This workflow would _massively_ improve the experience when working with large codebases you don't fully understand, or even just when working on any problem that's spread out far across a codebase.
It's something I haven't really seen implemented anywhere else, but I think it'd be _very_ well suited to the multibuf workflow, and would probably interact quite well with a variety of other UX aspects of multibufs.

aplatypus:
I came across this looking for a bookmark plugin for Zed.  Similar to the Bookmark feature in IDE-s/editors like InteliJ, Netbeans, Visual Studio, SCiTE and the VS Code Bookmarks plug-in.

This description sounds more like a code index or cross-reference, perhaps some checkpoint support??  It sounds useful, though I'd not recognise it as "bookmark" functionality.

NotNullDev:
Isn't that something like mnemonic bookmark in IntelliJ?

PR #51404, opened 2026-03-12
editor: Bookmarks MVP

### LINKED ISSUE #4526: Bookmarks
### Check for existing issues

- [X] Completed

### Describe the feature

Setting a bookmark to a line of code saves a lot of time while working on unfamiliar codebase.

Screenshot 2023-08-21 at 10 49 14
Screenshot 2023-08-21 at 10 49 46

austincummings:
Closes #4526

Adds basic bookmark functionality to the editor, allowing users to mark lines and later navigate between them.

### What's new

**Toggling bookmarks**
Users can toggle a bookmark on the current line(s) via the `editor: toggle bookmark` action. A bookmark icon appears in the gutter for each bookmarked line.

**Navigation**
Two new actions, `editor: go to next bookmark` and `editor: go to previous bookmark`, navigate between bookmarks in the current buffer, wrapping around at the ends of the buffer.

**Viewing all bookmarks**
`editor: view bookmarks` opens all bookmarks across the project in a multibuffer, similar to how references and diagnostics are surfaced.

**Clearing bookmarks**
`workspace: clear bookmarks` removes all bookmarks in the current project.

**Persistence**
Bookmarks are persisted to the workspace database and restored when the workspace is reopened. They are stored as `(path, row)` pairs and resolved back to text anchors. Out of range or unresolvable bookmarks are skipped with a logged warning.

**Gutter rendering**
Bookmark icons are rendered in the gutter using the existing gutter button layout system, consistent with breakpoints. They are suppressed on lines that already show a breakpoint or phantom breakpoint indicator. A new `gutter.bookmarks` setting (defaulting to `true`) controls their visibility.

### What's left

- [x] Lazily load buffers that have bookmarks
- [x] Clean up test boilerplate
- [ ] Assign default keybindings
- [ ] Compare line of saved bookmarks with current buffer (gray out the "stale" bookmarks)

### What's next (and nice to haves)
- [ ] Resilience against external edits
- [ ] Save column position with the bookmark
- [ ] Bookmarks attached to syntactic structures?
- [ ] Labeled bookmarks?

---

Release Notes:

- Added bookmarks: toggle bookmarks on lines with `editor: toggle bookmark`, navigate with `editor: go to next bookmark` / `editor: go to previous bookmark`, view all bookmarks with `editor: view bookmarks`, and clear with `workspace: clear bookmarks`. Bookmarks are shown in the gutter and persisted across sessions.


banesbyte:
@austincummings This looks awesome and I believe many people agree if I said: Thank you for finally getting this moving toward a solid concept!

I have some feedback, please consider this more just an opinion. While reading through many discussions, I noticed Zed community is trying to reevaluate each feature and ensure it works with its new unique eco-system. People just can't seem to agree on how it should actually work, especially when it comes to the specifics of Zed: real-time collaboration, making AI Agent understand them.

Minor note on **Bookmark**, this term is kind of outdated and semantically incorrect to the modern IDEs. **Vim Mode** already implements **Marks** which are essentially the same. To maintain consistency and familiarity for all users, it may be beneficial to keep the name the same – **Marks**.

Major point of argument (pointed out in #16828) is **Treesitter-Aware** marks could be more natural for Zed and advanced rather than storing them as _path/row_ pairs. Otherwise, this approach essentially replicates every other editor, which is also why they can be annoying sometimes. For example, if you add or delete lines the row numbers get messed up and the bookmark might end up pointing to a completely different line.
The **Treesitter** would anchor to the syntax node: the function, struct, or method. It already parses the entire buffer into a syntax tree, would store something like `path, node_kind="function_def", node_name="handle_request"` and dynamically resolve to the current position of that node regardless of how many lines moved around it.

Thanks again! I will try to give it a try during.

assets/icons/bookmark.svg
yara-blue, line 1: TODO(yara), this is just for myself: 

talk to a designer to get an icon we can merge (needs to be GPL etc, easiest is to use one owned by zed-industries)

crates/editor/src/editor.rs
```diff
@@ -12119,6 +12193,185 @@ impl Editor {
         }
     }
 
+    pub fn toggle_bookmark(
```
yara-blue: There is enough bookmark code here that it's probably worth lifting it to it's own module. Then just impl Editor in that module have these there.

crates/editor/src/editor.rs
```diff
@@ -12119,6 +12193,185 @@ impl Editor {
...
+        let Some(bookmark_store) = &self.bookmark_store else {
+            return;
+        };
+
+        let selection = self
+            .selections
+            .newest::<MultiBufferOffset>(&self.display_snapshot(cx));
+        let multi_buffer_snapshot = self.buffer.read(cx).snapshot(cx);
+
+        let bookmarks_in_range =
```
yara-blue: feels like this should be its own function not just a closure. LSP convert to fn works for me, (then just put it somewhere close).

crates/editor/src/editor.rs
```diff
@@ -12119,6 +12193,185 @@ impl Editor {
...
+                    })
+                    .collect()
+            };
+
+        let mut before = bookmarks_in_range(MultiBufferOffset(0)..selection.head(), cx);
+        let mut after = bookmarks_in_range(selection.head()..multi_buffer_snapshot.len(), cx);
+        before.sort_by_key(|a| a.to_offset(&multi_buffer_snapshot));
+        after.sort_by_key(|a| a.to_offset(&multi_buffer_snapshot));
+
+        let anchor = if direction == Direction::Next {
```
yara-blue: I do not think this wraps around in both cases (for example: call prev on first bookmark)? Can you see if you can find a clearer solution? Let me know if you need help!

yara-blue: nit: I'd prefer a match statement here

crates/editor/src/editor.rs
```diff
@@ -12119,6 +12193,185 @@ impl Editor {
...
+            after
+                .into_iter()
+                .chain(before)
+                .find(|anchor| anchor.to_offset(&multi_buffer_snapshot) != selection.head())
+        } else {
+            [before, after]
```
yara-blue: This is not clear to me, why do we search through both?

crates/editor/src/editor.rs
```diff
@@ -12119,6 +12193,185 @@ impl Editor {
...
+
+        let resolve_task = bookmark_store.update(cx, |store, cx| store.resolve_all(cx));
+
+        cx.spawn_in(window, async move |workspace, cx| {
+            resolve_task.await.log_err();
+
+            workspace
+                .update_in(cx, |workspace, window, cx| {
+                    let bookmark_store = workspace.project().read(cx).bookmark_store();
+                    let locations = bookmark_store.read(cx).all_bookmark_locations(cx);
```
yara-blue: should this not be an async fn that resolves and then lists the bookmarks instead of having the separate resolve task?

crates/editor/src/editor.rs
```diff
@@ -27804,7 +28063,7 @@ impl EditorSnapshot {
             let mut left_padding = git_blame_entries_width.unwrap_or(Pixels::ZERO);
```
yara-blue: Might be worth a rewrite now that there is so much in here. Is it also correct that a bookmark is 3 characters wide?

crates/editor/src/element.rs
```diff
@@ -3125,6 +3128,67 @@ impl EditorElement {
...
+        bookmarks: HashSet<DisplayRow>,
+        row_infos: &[RowInfo],
+        window: &mut Window,
+        cx: &mut App,
+    ) -> Vec<AnyElement> {
+        if self.split_side == Some(SplitSide::Left) {
+            return Vec::new();
+        }
+
+        self.editor.update(cx, |editor, cx| {
```
yara-blue, line 3128: This is almost exactly layout_breakpoints. The only change seems to be the HashSet item and the render_method called, can we deduplicate them? One approach would be to have one method layout_gutter_item which takes a: 
 - Fn(&T) -> display_row; where T is the hashset content
 - Fn(T, cx) -> AnyElement;  renders bookmark Icon or breakpoint ButtonIcon

crates/editor/src/element.rs
```diff
@@ -10747,6 +10822,25 @@ impl Element for EditorElement {
                         Vec::new()
                     };
 
+                    let show_bookmarks =
+                        snapshot.show_bookmarks.unwrap_or(gutter_settings.bookmarks);
+                    let bookmarks = if show_bookmarks {
```
yara-blue, line 10764: since layout_bookmarks and layout_breakpoints share so much it may be worth extracting the common arguments to some Struct. May even be able to use that for layout_run_indicators

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+pub struct BookmarkAnchor(text::Anchor);
+
+impl BookmarkAnchor {
+    pub fn anchor(&self) -> text::Anchor {
+        self.0
+    }
+}
+
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
yara-blue, line 22: Could use a comment explaining how this is the serialized representation

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+            BookmarkEntry::Loaded(buffer_bookmarks) => buffer_bookmarks.bookmarks.is_empty(),
+            BookmarkEntry::Unloaded(rows) => rows.is_empty(),
+        }
+    }
+}
+
+pub struct BookmarkStore {
+    buffer_store: Entity<BufferStore>,
+    worktree_store: Entity<WorktreeStore>,
+    bookmarks: BTreeMap<Arc<Path>, BookmarkEntry>,
```
yara-blue, line 85: What if we move a file around in Zed (like a rename)? We lose the bookmarks then. Not something we should fir in this MVP (already huge at 2.2kL) but we do need to track it somewhere.

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    pub fn with_serialized_bookmarks(
+        &mut self,
+        bookmark_rows: BTreeMap<Arc<Path>, Vec<BookmarkRow>>,
+        cx: &mut Context<Self>,
+    ) -> Task<Result<()>> {
+        self.bookmarks.clear();
+
+        for (path, rows) in bookmark_rows {
+            if rows.is_empty() {
```
yara-blue, line 105: Q: this is where finding the new location for moved source would go?
(the right choice not to do that in this MVP btw)

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+                continue;
+            }
+
+            let count = rows.len();
```
yara-blue, line 109: nit: bit overkill, I think we can live with: "Stored 8 unloaded bookmark(s) at some place"

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+            return;
+        };
+
+        let snapshot = buffer.read(cx).snapshot();
+        let max_point = snapshot.max_point();
+
+        let anchors: Vec<BookmarkAnchor> = rows
+            .iter()
+            .filter_map(|bookmark_row| {
+                let point = Point::new(bookmark_row.0, 0);
```
yara-blue, line 135: Something to consider, should bookmarks bring you to a specific character? On the one hand this is more work to implement now but it feels relatively small. It would save us a DB migration later.

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+            self.bookmarks.remove(abs_path);
+        } else {
+            let mut buffer_bookmarks = BufferBookmarks::new(buffer.clone(), cx);
+            buffer_bookmarks.bookmarks = anchors;
+            self.bookmarks
+                .insert(abs_path.clone(), BookmarkEntry::Loaded(buffer_bookmarks));
+        }
+    }
+
+    /// Opens buffers for all unloaded bookmark entries and resolves them to anchors.
```
yara-blue: Maybe add: "This is used to show all bookmarks in a large multi-buffer"

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+        &self.bookmarks
+    }
+
+    /// Returns the bookmarks for a given buffer within an optional range.
+    /// Only returns bookmarks that have been resolved to anchors (loaded).
+    /// Unloaded bookmarks for the given buffer will be resolved first.
+    pub fn bookmarks_for_buffer(
+        &mut self,
+        buffer: Entity<Buffer>,
+        range: Option<Range<text::Anchor>>,
```
yara-blue: When I read this first the Option surprised me a little. This is only used in the test, maybe the test should pass in not None but an empty range?

crates/project/src/bookmark_store.rs
```diff
@@ -0,0 +1,436 @@
...
+                if rows.is_empty() {
+                    None
+                } else {
+                    Some((path.clone(), rows))
+                }
+            })
+            .collect()
+    }
+
+    pub fn all_bookmark_locations(&self, cx: &App) -> HashMap<Entity<Buffer>, Vec<Range<Point>>> {
```
yara-blue: Could use a comment that the bookmarks are already resolved when this is called

crates/workspace/src/persistence.rs
```diff
@@ -970,6 +1004,16 @@ impl Domain for WorkspaceDb {
         sql!(
             ALTER TABLE remote_connections ADD COLUMN use_podman BOOLEAN;
         ),
+        sql!(
```
yara-blue, line 1017: Since bookmarks mirrors breakpoints might help readability to keep them close to each-other. Unless the order of the array here matters but I think not...?

crates/workspace/src/persistence.rs
```diff
@@ -1194,12 +1239,47 @@ impl WorkspaceDb {
...
+            })
+            .and_then(|mut prepared_statement| (prepared_statement)(workspace_id));
+
+        match bookmarks {
+            Ok(bookmarks) => {
+                if bookmarks.is_empty() {
+                    log::debug!("Bookmarks are empty after querying database for them");
+                }
+
+                let mut map: BTreeMap<Arc<Path>, Vec<BookmarkRow>> = BTreeMap::default();
```
yara-blue: nit: type is not needed

yara-blue (changes requested):
Overall very nice! Could use some refactoring work and a clarification here and there (especially the next and prev bookmarks impl).

Please feel free to let me know if you think differently on any of the points below. The nits do not block merging btw.

crates/editor/src/editor.rs
austincummings: Done, broke that out into an associated method: `bookmarks_in_range`

crates/editor/src/editor.rs
austincummings: Switched to a match

crates/editor/src/editor.rs
austincummings: That was used to allow looping back around to the front of the file. I rewrote this now though to hopefully be clearer.

crates/project/src/bookmark_store.rs
austincummings: I went ahead and just renamed it to `SerializedBookmark`

crates/project/src/bookmark_store.rs
austincummings: Done, got rid of the option, if needed the caller can always just pass the full range of the buffer.

crates/workspace/src/persistence.rs
austincummings: I believe the order of this does matter. I _think_ (could be wrong) that if it were up by breakpoints it might not run since that set of migrations has already run.

austincummings:
> @austincummings I'd love to merge this before we get more conflicts. If you don't have time now that's no problem. Hope you don't mind me fixing the issues and merging in that case.

@yara-blue Understood! I addressed some of your comments and will continue to do so today as I get the chance. I also resolved any conflicts that had accumulated so far. I'll leave comments on the areas I haven't addressed yet at the end of the day.

crates/workspace/src/persistence.rs
yara-blue: on it, ill see what it is :)

yara-blue: yes you are right! https://github.com/zed-industries/zed/blob/c372f246a0790da012dddbe3935ee2758a3a9c3b/crates/sqlez/src/migrations.rs#L75 we assume the migrations and completed_migrations (a possibly pre-existing table) are in the same order. Good catch.

crates/project/src/bookmark_store.rs
austincummings: Agreed and done.

crates/project/src/bookmark_store.rs
austincummings: Done!

crates/editor/src/editor.rs
yara-blue: the `or_else` is really readable nice refactor!!

crates/project/src/bookmark_store.rs
yara-blue: reworking this atm

crates/editor/src/editor.rs
yara-blue: refactoring atm will do this

crates/editor/src/element.rs
yara-blue: Done

crates/workspace/src/persistence.rs
yara-blue: I was wrong :)

yara-blue:
few TODOs for myself:

- [x] bookmark icon overlays runnables 
  image
- [x] no Set bookmarks in context menu
  image
- [x] no setting to hide bookmarks
  image
- [x] cannot click bookmark to remove (breakpoint takes priority)
  imageimage

crates/editor/src/editor.rs
yara-blue: nvmd needed docs otherwise good

yara-blue:
more todos, 

- [x] I broke this: 
  image 
  context: the context menu should not show there is a "play" (runs test) button under there.

yara-blue:
Hey @austincummings I unfortunately could not get this polished enough to merge today. There are four things still needed:
- the bookmarks icon needs to become a button, so you can click it to remove the current bookmark.
- the context menu in the gutter needs an extra line for add/remove bookmark. This context menu is called `breakpoint_context_menu`. It should probably be renamed to `gutter_context_menu` at that point. Most work here will be in detecting whether or not the current line has a bookmark set. 
- the setting on whether to show or hide bookmarks needs to be added/fixed.
- the tests are all messed up (that's really my fault). I disabled these for now so they will not bother you.
- the commits I added needs to be reviewed.

Feel free to take some of these on over the weekend.

crates/editor/src/editor.rs
```diff
@@ -12159,6 +12173,66 @@ impl Editor {
...
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) -> Option<Anchor> {
+        let snapshot = self.snapshot(window, cx);
+        let bookmark_position = snapshot.buffer_snapshot().anchor_before(Point::new(row, 0));
+
+        self.bookmark_at_anchor(bookmark_position, &snapshot, cx)
+    }
+
+    // TODO(austincummings): This is largely the same as breakpoint_at_anchor, maybe an opportunity to refactor this
```
yara-blue: Thanks for the note :)

I've given it a short try but it gets ugly quick. You'd need to make bookmarks_for_buffer return an iterator. Doing so would make its code kinda ugly in my opinion. For a second I considered if it was worth it for perf. but this is far from a hot path so no.

yara-blue:
TODO: 
- [x] refactor the phantom breakpoint into the "gutter button" it now has become
- [x] collab notes have the breakpoint hover/gutter button (regression)
- [x] uncomment the line tests and fix their API

yara-blue:
@cameron1024 regarding the review feedback on the Path type, we only get/set paths on zed API's that return normal paths not RelPaths. I double checked and those API's live above the remoting/local switch. So we are good there! No remoting gotcha's :)

cameron1024:
Nice, thanks for checking 😁

yara-blue (approved):

yara-blue:
@austincummings I messed up merging (forgot that I needed to set a proper commit message for squashing). So I reverted this and re-opened the PR separately: https://github.com/zed-industries/zed/pull/54174

Sorry for all the notifications/emails you undoubtedly got.

Basic bookmarks should be in now! Still a lot of work to be done namely:
- add a picker.
- integrate with vim marks.
- add that cool syntax tree based bookmark positions tracking you where working on.
- maybe make it possible to give bookmarks names or a description?

These should all be in separate PR's. 

I thought you might want to take on some of these yourselves (it is your feature after all). If you would not mind to have someone else work on some let me know then I'll put them up as feature issues for the guild.

cameron1024:
@zed-industries/approved

yara-blue:
Hi everyone. Lets make Zed's bookmarks system the best there is! I've just opened a tracking issue (https://github.com/zed-industries/zed/issues/55871). If you are looking to work on something please open an issue to discuss it and let us know with a comment in the tracking issue.

PR #51414, opened 2026-03-12
gpui: Recalculate list layout after the window has been resized

feitreim:
Closes #51417

I noticed this bug in the settings menu where when I opened the settings menu, I could not scroll down through all the available options, eg. on the initial page I wasn't able to scroll down to privacy. When I saw that no one else had reported this issue, I figured it may be due to my setup, and it turns out that using Aerospace, the window manager I use, was what made this bug visible to me. Because aerospace resizes the window right after it launches, the originally computed heights for the list are incorrect, meaning the scroll bar is the wrong size as well.

in the relevant code there was a comment that says "If the width of the list has changed, invalidate all cached item heights" which wasn't incorrect per-se, but it just invalidated them without triggering any re-computation, causing incorrect scroll bars.

My intuition is that window resizes/events that change the width of the list bounds are fairly rare, so there shouldn't be a large performance hit from the change.

Also implemented a test that directly showcases the behavior, if you run the test without the change it fails, as the max_offset_for_scrollbar will be wrong.

Videos:

Before 

https://github.com/user-attachments/assets/2b680222-7071-4098-863f-519361f0756a

After:

https://github.com/user-attachments/assets/1222a299-23d7-4007-8e88-55d2daccce64

[x] Tests
[x] Video of behavior

Release Notes:

- gpui: fixed list height re-computation when the list width changes.

### DIFF
--- a/crates/gpui/src/elements/list.rs
+++ b/crates/gpui/src/elements/list.rs
@@ -1350,6 +1350,41 @@ mod test {
     }
 
     #[gpui::test]
+    fn test_measure_all_after_width_change(cx: &mut TestAppContext) {
+        let cx = cx.add_empty_window();
+
+        let state = ListState::new(10, crate::ListAlignment::Top, px(0.)).measure_all();
+
+        struct TestView(ListState);
+        impl Render for TestView {
+            fn render(&mut self, _: &mut Window, _: &mut Context<Self>) -> impl IntoElement {
+                list(self.0.clone(), |_, _, _| {
+                    div().h(px(50.)).w_full().into_any()
+                })
+                .w_full()
+                .h_full()
+            }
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi (approved):
I've reproduced this with a similar window management software (yabai) and confirmed that the fix works.

before:

https://github.com/user-attachments/assets/367dfb1e-83a4-457a-bcc7-f13ea32f7e33

after:

https://github.com/user-attachments/assets/e90725b7-df83-4e9d-b7b8-fe7e95fcc9dc

Test fails without the fix, and passes with the fix.

I think this fix is correct, since the width is being invalidated and this is *also* what is done when height changes in `remeasure`

PR #51420, opened 2026-03-12
Use `-e` instead of `-c` when getting environment from nushell

### LINKED ISSUE #38200: Zed fails to find dart lsp with nix, nushell and direnv
### Summary

Zed fails to expand `~` or environment variables like `$FLUTTER_ROOT` in LSP configuration paths, requiring absolute paths instead.

**Problem**: This breaks cross-platform dotfiles since absolute paths differ between systems (e.g., `/Users/username` vs `/home/username`).

**Expected behavior**: Zed should expand `~` to the user's home directory and resolve environment variables in configuration  paths, similar to how other editors like Helix handle this.

**What I tried**:
- Using `"dart"` (relying on PATH) - failed, Zed doesn't inherit my shell's PATH
- Using `"~/.local/share/flutter/bin/dart"` - failed, Zed doesn't expand tilde
- Using `"$FLUTTER_ROOT/bin/dart"` - failed, Zed doesn't expand environment variables

**What works**:
- Only absolute paths like `"/Users/username/.local/share/flutter/bin/dart"` work

```
Zed: v0.203.4 (Zed) 
OS: macOS 15.6.1
Memory: 32 GiB
Architecture: aarch64
```

**System info**:
- macOS with Nix package manager
- Using Nushell as shell
- Flutter installed at `~/.local/share/flutter` via manual install (not Nix due to known issues)
- `FLUTTER_ROOT` environment variable properly set in Nix config

NotLebedev:
Closes #38200

Applied [suggestion](https://github.com/zed-industries/zed/issues/38200#issuecomment-3354159899) from issue to use `-l -e` instead of `-l -i -c` when running on nushell because of how it treats `-l` as implying interactive session. Using `-e` also means that command needs to end with `exit` to terminate shell manually. With this changes everything now works fine in my testing.

Before, zed fails to load environment variables and there is error in logs:

image

image

With this patch everything works fine and all language servers and stuff loads fine:

image

Tested on nixos unstable. Nushell version 0.110.0 and 0.111.0. Zed version 0.223.3+stable and compiled from main fail. Zed from this branch works.

Release Notes:

- Fixed loading environment variables when nushell is used as shell

### DIFF
--- a/crates/util/src/shell_env.rs
+++ b/crates/util/src/shell_env.rs
@@ -73,13 +73,27 @@ async fn capture_unix(
             command.arg("-l");
         }
     }
+
+    match shell_kind {
+        // Nushell does not allow non-interactive login shells.
+        // Instead of doing "-l -i -c '<command>'"
+        // use "-l -e '<command>; exit'" instead
+        ShellKind::Nushell => command.arg("-e"),
+        _ => command.args(["-i", "-c"]),
+    };
+
     // cd into the directory, triggering directory specific side-effects (asdf, direnv, etc)
     command_string.push_str(&format!("cd '{}';", directory.display()));
     if let Some(prefix) = shell_kind.command_prefix() {
         command_string.push(prefix);
     }
     command_string.push_str(&format!("{} --printenv {}", zed_path, redir));
-    command.args(["-i", "-c", &command_string]);
+
+    if let ShellKind::Nushell = shell_kind {
+        command_string.push_str("; exit");
+    }
+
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

Veykril:
Okay, I misunderstood the core issue. Windows or not shouldn't matter, as this is a shell expansion / config loading issue, so this only fails on certain setups I believe. Nu being weird stays the norm I guess.

PR #51435, opened 2026-03-12
Add popover click to go to definition

eholk:
This makes the type words in popover hints clickable to go to their definition.

There's a key limitation right now where we can't disambiguate if the same name occurs more than once to refer to multiple types.

Release Notes:


crates/editor/src/hover_popover.rs
```diff
@@ -2155,4 +2081,513 @@ mod tests {
...
+
+        cx.set_state(indoc! {"
+            mod foo {
+                pub struct Baz;
+            }
+            mod bar {
+                pub struct Baz;
+            }
+            struct Wrapper<T, U>(T, U);
+            let vaˇriable = Wrapper(foo::Baz, bar::Baz);
```
eholk, line 2351: Currently this case gives us incorrect information. The popover will show `let variable: Wrapper<Baz, Baz>`, but if you click either `Baz` it will take you to `bar::Baz` when the first one should take you to `foo::Baz`.

PR #51446, opened 2026-03-13
Implement cross window clipboard

mxvsh:

Demo:

https://github.com/user-attachments/assets/03356711-c36b-4336-acd1-80496db0e1a1

Release Notes:

- Added cross-window file/folder copy-paste support

### DIFF
--- a/crates/project_panel/src/project_panel.rs
+++ b/crates/project_panel/src/project_panel.rs
@@ -478,8 +478,6 @@ impl FoldedAncestors {
 }
 
 pub fn init(cx: &mut App) {
-    ProjectPanelClipboardStore::global(cx);
-
     cx.observe_new(|workspace: &mut Workspace, _, _| {
         workspace.register_action(|workspace, _: &ToggleFocus, window, cx| {
             workspace.toggle_panel_focus::<ProjectPanel>(window, cx);
@@ -862,11 +860,6 @@ impl ProjectPanel {
             .detach();
 
             cx.observe_global::<FileIcons>(|_, cx| {
-                cx.notify();
-            })
-            .detach();
-
-            cx.observe(&clipboard_store, |_, _, cx| {
                 cx.notify();
             })
             .detach();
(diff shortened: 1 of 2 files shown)


crates/project_panel/src/project_panel.rs
```diff
@@ -444,6 +478,8 @@ impl FoldedAncestors {
 }
 
 pub fn init(cx: &mut App) {
+    ProjectPanelClipboardStore::global(cx);
```
Veykril: Given we init this lazily anyways, no need to call this here

crates/project_panel/src/project_panel.rs
```diff
@@ -829,6 +866,11 @@ impl ProjectPanel {
             })
             .detach();
 
+            cx.observe(&clipboard_store, |_, _, cx| {
+                cx.notify();
+            })
+            .detach();
```
Veykril: This does not do anything.

crates/project_panel/src/project_panel.rs
```diff
@@ -2997,39 +3041,76 @@ impl ProjectPanel {
 
     fn cut(&mut self, _: &Cut, _: &mut Window, cx: &mut Context<Self>) {
         let entries = self.disjoint_effective_entries(cx);
-        if !entries.is_empty() {
-            self.clipboard = Some(ClipboardEntry::Cut(entries));
-            cx.notify();
+        if let Some(clipboard_data) = self.build_clipboard_data(entries, cx) {
+            self.clipboard_store.update(cx, |clipboard_store, cx| {
+                clipboard_store.clipboard = Some(ClipboardEntry::Cut(clipboard_data));
+                cx.notify();
```
Veykril: This notify needs to be called on the outer `cx`, we want to notify observes of the project panel here

crates/project_panel/src/project_panel.rs
```diff
@@ -2997,39 +3041,76 @@ impl ProjectPanel {
...
 
     fn copy(&mut self, _: &Copy, _: &mut Window, cx: &mut Context<Self>) {
         let entries = self.disjoint_effective_entries(cx);
-        if !entries.is_empty() {
-            self.clipboard = Some(ClipboardEntry::Copied(entries));
-            cx.notify();
+        if let Some(clipboard_data) = self.build_clipboard_data(entries, cx) {
+            self.clipboard_store.update(cx, |clipboard_store, cx| {
+                clipboard_store.clipboard = Some(ClipboardEntry::Copied(clipboard_data));
+                cx.notify();
```
Veykril: This notify needs to be called on the outer `cx`, we want to notify observes of the project panel here

crates/project_panel/src/project_panel.rs
```diff
@@ -230,8 +231,41 @@ impl EditState {
...
+struct ClipboardData {
+    source_project: WeakEntity<Project>,
+    source_project_id: gpui::EntityId,
+    items: BTreeSet<SelectedEntry>,
+    source_paths: Vec<ClipboardSourcePath>,
+}
+
+#[derive(Clone, Debug)]
+struct ClipboardSourcePath {
+    abs_path: Arc<Path>,
```
Veykril, line 248: Can we consider storing something else here that is not yet resolved, like the entry id from the foreign project? Keeping the abs path here means if the entry changes in some way in the foreign project the operation will fail as the abs path is no longer valid

Veykril (commented):
There is a fair amount of duplication between the same project and cross project paste logic, would be nice if we could deduplicate that a bit

Veykril:
Any update on this?

Veykril:
Closing this due to inactivity for now, feel free to come back with this when you got the time

PR #51451, opened 2026-03-13
workspace: Add default project folder in settings

### LINKED ISSUE #15198: Default open folder
### Check for existing issues

- [X] Completed

### Describe the feature

Hey all, 

I am loving the editor.  I get a lot of delight from the amount of thought that went into quality of life features.  A quality of life feature that would be nice for my workflow is if I could set the default folder which opens when I press "Open a new project".  

I place all of my software projects in a single folder in my home directory called programs.  I would like to be able to tell the editor to look there first.

I also understand that this feature could already exist, I looked through the issue tracker and could not find a similar request, but I also admit that I didn't do an exhaustive search, there were lots of work items in there.  I also understand that this is request would be low in the priority list and might never get tackled.  Thank you so much anyways!

_No response_

Verdenroz:
Closes #15198

Users can now configure what directory opens on opening new projects/folders.

A really old issue, but one that had been bugging me for days after migrating from vscode since I had all my projects in a specific directory.

https://github.com/user-attachments/assets/100b270b-d494-496c-b1a3-14e657cc7510

Release Notes:

- Added a `default_project_folder` workspace setting that controls the starting directory shown when opening folders.


crates/gpui_windows/src/platform.rs
```diff
@@ -1106,6 +1106,21 @@ fn file_open_dialog(
     unsafe {
         folder_dialog.SetOptions(dialog_options)?;
 
+        if let Some(initial_directory) = options.initial_directory.as_ref()
+            && !initial_directory.to_string_lossy().is_empty()
+            && let Some(full_path) = initial_directory
+                .canonicalize()
+                .context("failed to canonicalize initial directory")
```
SomeoneToIgnore, line 1113: * That's odd to do the canonicalization and sanitation that late, as usually great to do somewhere closer to startup after the config is parsed?
* At least, would be use `Fs` trait for this as this is a blocking OS operation?
* And would be definitely good to use `with_context`, to remove parasitic `failed to` and to add more context (e.g. file path).

crates/open_path_prompt/src/open_path_prompt.rs
```diff
@@ -228,7 +229,13 @@ impl OpenPathPrompt {
             let delegate =
                 OpenPathDelegate::new(tx, lister.clone(), creating_path, cx).show_hidden();
             let picker = Picker::uniform_list(delegate, window, cx).width(rems(34.));
-            let mut query = lister.default_query(cx);
+            let mut query = options
```
SomeoneToIgnore, line 232: We can test this at least, right?
Let's add one below that uses the settings (correct and incorrect path test cases).

crates/util/src/paths.rs
```diff
@@ -18,6 +18,17 @@ use std::{
 use crate::rel_path::RelPath;
 use crate::rel_path::RelPathBuf;
 
+/// Expands a leading `~` in a path string to the user's home directory.
+pub fn expand_tilde(path: &str) -> PathBuf {
```
SomeoneToIgnore, line 22: This is odd to see as the issue had been definitely dealt with in Zed before, and besides expanding we need canonicalization and `SanitizedPath` wrapping.

Let's remove this and unify the approach with the rest of the code, maybe checking for `shellexpand` also helps.

crates/settings_ui/src/page_data.rs
```diff
@@ -155,6 +155,24 @@ fn general_page() -> SettingsPage {
...
+                    json_path: Some("default_project_folder"),
+                    pick: |settings_content| {
+                        settings_content.workspace.default_project_folder.as_ref().or(DEFAULT_EMPTY_STRING)
+                    },
+                    write: |settings_content, value| {
+                        settings_content.workspace.default_project_folder = value;
+                    },
+                }),
+                metadata: Some(Box::new(SettingsFieldMetadata {
+                    placeholder: Some("~/programs"),
```
SomeoneToIgnore, line 171:
```suggestion
                    placeholder: Some("~/"),
```

Let's not use too custom names here, in the docs and the rest of the PR.

crates/git_ui/src/clone.rs
```diff
@@ -18,6 +18,7 @@ pub fn clone_and_open(
         files: false,
         directories: true,
         multiple: false,
+        initial_directory: None,
```
SomeoneToIgnore, line 21: I wonder if it makes sense to introduce breaking changes to the public API (as gpui is a library) and touch that many files for 3 (?) places exactly where we read that field — we seem to have `cx` near all those places and I wonder what it takes to have a method to get + pass this path specifically?

Maybe, this is a good place to do all the async FS operations and return a `Task` from this new method too.

SomeoneToIgnore (changes requested):
Thank you, and sorry for the late review.

Looking at the surface of the changes, I fear that's quite a bunch of moving pieces and at least 2 separate tracks (`platform` and `open_path_prompt`) for manipulating and, worst, determining the default path.

Can we somehow have a common, single place that determines a default path for all kinds of prompts?
`pub fn default_query` or its analogue seems like a good start?

SomeoneToIgnore:
Thank you for the update, and no rush.

I will close this PR for now just to keep the queue cleaner, as the conflicts start to pop-up:
image

Verdenroz:
Still in the process for approval! I want to see if I can look at this within 1-2 weeks, hopefully this time for real 😅

PR #51457, opened 2026-03-13
git_ui: Fix diff clipboard against selection

### LINKED ISSUE #50912: "diff clipboard with selection" doesn't support split view
### Reproduction steps

1. write something into a buffer
2. copy to clipboard
3. make some small changes
4. trigger "editor: diff clipboard with selection

No split view:

Image

Zed: v0.227.0+preview.196.170485782a1d0ffb3a67f3382681b50deea10b6e (Zed Preview) 
OS: Linux Wayland fedora 44
Memory: 62.7 GiB
Architecture: x86_64
GPU: AMD Radeon RX 6800 XT (RADV NAVI21) || radv || Mesa 25.3.6

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

OmChillure:
fixes #50912 

### What changed

- Mapped selection points from multibuffer space to buffer-local points for non-singleton editors.
- Switched the view wiring to SplittableEditor-compatible behavior.
- Preserved previous empty-selection behavior for singleton buffers (full-buffer range behavior remains unchanged).
- Kept diff hunk controls behavior consistent with prior UX.
- Added a regression test for the non-singleton multibuffer selection path.
- Stabilized tests by pinning diff view style in test init so snapshots are deterministic.

### Testing

- cargo test -p git_ui text_diff_view
- cargo test -p git_ui text_diff_view::tests::test_diffing_clipboard_against_selection
- cargo check -p git_ui

### Result

- Existing text_diff_view tests pass.
- New regression test passes.
- Behavior now works for split/multi-excerpt editors while preserving prior singleton behavior.

### Release Notes

- Fixed “Diff Clipboard with Selection” to work correctly from split view / non-singleton multibuffer editors.
- Preserved existing singleton behavior for empty selections.
- Added regression coverage for non-singleton selection mapping.

### DIFF
--- a/crates/git_ui/src/text_diff_view.rs
+++ b/crates/git_ui/src/text_diff_view.rs
@@ -2,7 +2,10 @@
 
 use anyhow::Result;
 use buffer_diff::BufferDiff;
-use editor::{Editor, EditorEvent, MultiBuffer, ToPoint, actions::DiffClipboardWithSelectionData};
+use editor::{
+    Editor, EditorEvent, EditorSettings, MultiBuffer, SplittableEditor, ToPoint,
+    actions::DiffClipboardWithSelectionData,
+};
 use futures::{FutureExt, select_biased};
 use gpui::{
     AnyElement, App, AppContext as _, AsyncApp, Context, Entity, EventEmitter, FocusHandle,
@@ -10,6 +13,7 @@ use gpui::{
 };
 use language::{self, Buffer, Point};
 use project::Project;
+use settings::Settings;
 use std::{
     any::{Any, TypeId},
     cmp,
@@ -22,13 +26,13 @@ use ui::{Color, Icon, IconName, Label, LabelCommon as _, SharedString};
 use util::paths::PathExt;
 
 use workspace::{
-    Item, ItemHandle as _, ItemNavHistory, Workspace,
+    Item, ItemNavHistory, Workspace,
     item::{ItemEvent, SaveOptions, TabContentParams},
     searchable::SearchableItemHandle,
 };
 
 pub struct TextDiffView {
-    diff_editor: Entity<Editor>,
+    diff_editor: Entity<SplittableEditor>,
     title: SharedString,
     path: Option<SharedString>,
     buffer_changes_tx: watch::Sender<()>,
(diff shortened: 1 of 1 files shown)


dinocosta (commented):
Hey @OmChillure 👋 

Thank you for picking this one up! Can you confirm whether these changes are ready for review? Testing it out locally seems to work fine but the description states `Draft adding tests !!` even though tests have also been added.

Also, are you able to add a release note for this? Seems like the template's content was completely removed from the pull request's description 🤔

P.S.: I actually even wonder if it's worth adding this, as it seems that `editor: diff clipboard with selection` is a bit wonky? If you have, say, 10 lines of text and copy that, then remove 4 or 5 words from the text, `editor: diff clipboard with selection` doesn't really result in a pleasant way to view that diff...

OmChillure:
> Hi, thanks for the PR 😁
> 
> It looks like there are two separate things this PR is fixing:
> 
> * you cannot currently open the selection/clipboard diff from a non-singleton multibuffer
> * the side-by-side diff view is not supported in the selection/clipboard diff
> 
> The code changes for the side-by-side diff view look broadly good, but I think there are some misunderstanding w.r.t. how multibuffer co-ordinates work that are complicating this PR. Could you split this out into two PRs please? I think we should be good to merge the side-by-side change without too many changes, which will already be a nice improvement😁
> 
> Some context on the multibuffer coordinate issue:
> 
> You have a comment: `For a singleton multibuffer the multibuffer Points are identical to buffer-local Points`. This isn't always true, because even a singleton multibuffer can have expanded deleted diff hunks, which are included in multibuffer co-ordinates but not in buffer co-ordinates. So for example, if you delete a line at the top of a file, then expand the deleted hunk, then do a clipboard/selection diff lower down in the file, the selected region will be incorrect.
> 
> Ideally, we'd have one code path that handles both singleton and non-singleton multibuffers and working on the multibuffer coordinate fix

Thanks for the detailed feedback @cameron1024! I've split this into two PRs as suggested:

PR 1 [Link](https://github.com/zed-industries/zed/pull/51966) : Side-by-side diff view support via SplittableEditor - no multibuffer coordinate changes.

PR 2 (follow-up): Multibuffer coordinate fix with a unified code path for both singleton and non-singleton multibuffers, accounting for expanded deleted diff hunks as you pointed out.

Working on PR 2 now. Will update this once done

PR #51461, opened 2026-03-13
editor: Open a singleton buffer when go to action results fit into a single excerpt

### LINKED ISSUE #44203: Go to definition and friends should not open a multibuffer if the multiple results fit into a single excerpt
https://github.com/zed-industries/zed/issues/6730#issuecomment-3402039073
> 2025-12-05.17.07.59.mov
> 
> Same behavior in default typescript project with `vtsls`

https://github.com/zed-industries/zed/issues/6730#issuecomment-3616173903
> having this issue in a giant TypeScript repo.
> SS.2025-10-14.at.15.57.17.mov</div>

Veykril:
Closes https://github.com/zed-industries/zed/issues/44203

Release Notes:

- Go to definition (and similar actions) will now no longer open a multi buffer if there are multiple results that all fit into a single excerpt

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -18119,6 +18119,7 @@ impl Editor {
 
         let workspace = self.workspace();
 
+        let excerpt_context_lines = multi_buffer::excerpt_context_lines(cx);
         cx.spawn_in(window, async move |editor, cx| {
             let locations: Vec<Location> = future::join_all(definitions)
                 .await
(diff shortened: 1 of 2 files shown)


osiewicz (approved):

crates/editor/src/editor.rs
```diff
@@ -17872,7 +17873,11 @@ impl Editor {
             for ranges in locations.values_mut() {
                 ranges.sort_by_key(|range| (range.start, Reverse(range.end)));
                 ranges.dedup();
-                num_locations += ranges.len();
+                let fits_in_one_excerpt = ranges
+                    .iter()
+                    .tuple_windows()
+                    .all(|(a, b)| b.start.row - a.end.row <= excerpt_context_lines);
```
cole-miller: 2 * excerpt_context_lines? Since both `a` and `b` will get expanded

Veykril: good catch

cole-miller (approved):

PR #51471, opened 2026-03-13
extension: Add support for local grammars during development

magecnion:
Closes #51352

- Add a new `[dev.grammars]` manifest section so dev extensions can build grammars from local parser instead of fetching remote repositories.
- Update extension compilation and grammar registration to use the active grammar set from the manifest, while keeping `[dev.grammars]` restricted to dev extensions only.
- Add manifest validation and regression coverage for invalid grammar configurations.

Release Notes:

- N/A


magecnion:
@cla-bot check

crates/extension/src/extension_manifest.rs
```diff
@@ -298,13 +298,101 @@ pub enum ExtensionLibraryKind {
...
 }
 
-#[derive(Clone, Default, PartialEq, Eq, Debug, Deserialize, Serialize)]
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum GrammarRepositorySource {
+    Local { repository: String },
+    Remote { repository: String, rev: String },
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
```
MrSubidubi: It does not really make sense to me that we can have a local repository and still a path - it should either be fully local or remote with the repo, rev and optional path. I think this and the GrammarManifestEntry should probably be combined.

crates/extension/src/extension_manifest.rs
```diff
@@ -298,13 +298,101 @@ pub enum ExtensionLibraryKind {
...
+
+impl<'de> Deserialize<'de> for GrammarManifestEntry {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let raw = RawGrammarManifestEntry::deserialize(deserializer)?;
+        Self::from_raw(raw).map_err(serde::de::Error::custom)
+    }
 }
```
MrSubidubi: I think we can simplify this by a lot with serde attributes and using an enum perhaps

crates/extension/src/extension_manifest.rs
```diff
@@ -599,4 +687,94 @@ args = ["--serve"]
...
     }
+
+    #[test]
+    fn parse_remote_grammar_manifest_entry() {
+        let entry: GrammarManifestEntry = toml::from_str(
+            r#"
+repository = "https://github.com/tree-sitter/tree-sitter-rust"
+rev = "abc123"
+path = "grammar"
+"#,
```
MrSubidubi: Nit but we can use `indoc` here to remove the indentation

MrSubidubi (changes requested):
Left a few nits, thanks for picking this up yourself.

Importantly though, we need to ensure that this only happens for dev extensions - we cannot compile the extensions otherwise for release. So we need to have a test and validation that this will only work for local environments.

crates/extension/src/extension_manifest.rs
```diff
@@ -298,13 +298,106 @@ pub enum ExtensionLibraryKind {
     Rust,
 }
 
-#[derive(Clone, Default, PartialEq, Eq, Debug, Deserialize, Serialize)]
-pub struct GrammarManifestEntry {
```
maxdeviant, line 302: I don’t think changing the grammar manifest entry is the right direction to go here.

Since the majority of the time we only care about the remote case, it introduces an additional impossible state that needs to be accounted for by the rest of the code.

I think we need to explore alternative directions for how to model this.

Perhaps a separate `dev_grammars` key in the manifest.

magecnion: Fair point. I'm not sure creating a new key for `dev_grammars` is an approach that will scale well, though. Every time something new is needed for the dev extension, a new key would have to be added to the manifest, if I understand correctly.

What about creating a `DevExtensionManifest` that holds the dev grammars and could potentially include additional parameters for dev extensions in the future? What do you think?

magecnion: Finally I discarded `DevExtensionManifest` way and created a new key in the manifest for dev extensions `dev`. Also included `grammars` there. Let me know what do you think @maxdeviant @MrSubidubi

PR #51491, opened 2026-03-13
agent_ui: Compact thread context in place

Esubaalew:
## Summary
Zed currently forces to open a new thread when the context limit reachs but that is little bad becuase chat history will get lost so this PR  adds compacting the thread so that  chat flows smoothly. the compaction is also hidden meaning the summary is is internaly managned to keep the chat feel coool

Release Notes:

- Improved agent chat compaction so long conversations continue in the same thread more smoothly.


benbrandt:
We''ll be working on this soon, thanks though!

PR #51502, opened 2026-03-13
editor: Fix indent-based folding breaking on multiline string literals

### LINKED ISSUE #27839: Folding of code gets weird on multiline text containing brackets
### Summary

It seems that folding functions gets some hick-up when my function contains multi-line text containing brackets. Here's an example:

```js
describe("1", () => {
  describe("-1", () => {
    it("tests something", async () => {
      await expect(
        new Promise((resolve, reject) => {
          reject(new Error("My error message"));
        }),
      ).rejects.toThrow(`[
  {
    "code": "custom",
    "message": "My error message.",
    "path": [
      "flowDiagram"
    ]
  }
]`);
    });
  });
});
```

Steps to trigger the problem:
1. Copy the code above into a javascript file.
2. Try to fold the 3rd line - the `it` statement.

Actual Behavior:

It evaluates the text, which contains brackets, as code it seems and does not properly collapse the function.

Image

Expected Behavior:

It should fold the function with all its content.

I removed the multi-line string and got the expected result:
Image

Zed: v0.179.2 (Zed) 
OS: macOS 14.4.1
Memory: 16 GiB
Architecture: aarch64

notJoon:
Closes #27839

Fix indent-based folding prematurely terminating when multiline string literals, comments kinds contain lines with lower indentation than surrounding code.

**Before Fix**
스크린샷 2026-03-14 오전 1 17 49

**After Fix**
스크린샷 2026-03-13 오전 9 05 58

스크린샷 2026-03-13 오전 9 06 13

Release Notes:

- Fixed indent-based code folding breaking when multiline strings or comments contain lines with less indentation than the surrounding code

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -2465,12 +2465,16 @@ impl std::ops::Deref for DisplaySnapshot {
     }
 }
 
-// Matches tree-sitter node kinds that represent multiline literals whose interior
-// indentation is content, not code structure. We match by substring because
-// grammars use varied names across languages (e.g. `string_literal`, `raw_string`,
-// `block_comment`, `heredoc_body`).
+// Matched by substring because tree-sitter grammars use varied node kind names
+// across languages (e.g. `string_literal`, `raw_string`, `block_comment`).
+// "heredoc" is separate because those nodes don't contain "string" or "comment"
+// in their names (e.g. `heredoc_body` in Bash, Ruby, Perl).
+const MULTILINE_LITERAL_NODE_PATTERNS: &[&str] = &["string", "comment", "heredoc"];
+
 fn is_multiline_literal_node(kind: &str) -> bool {
-    kind.contains("string") || kind.contains("comment") || kind.contains("heredoc")
+    MULTILINE_LITERAL_NODE_PATTERNS
+        .iter()
+        .any(|pattern| kind.contains(pattern))
 }
 
 /// A zero-indexed point in a text buffer consisting of a row and column adjusted for inserted blocks.
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -2362,15 +2362,17 @@ impl DisplaySnapshot {
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/editor/src/display_map.rs
```diff
@@ -2356,6 +2362,26 @@ impl DisplaySnapshot {
...
 
+    fn multiline_literal_end_row(&self, row: u32) -> Option<u32> {
+        let point = Point::new(row, 0);
+        let (node, _) = self.buffer_snapshot().syntax_ancestor(point..point)?;
+
+        let mut current = node;
+        loop {
+            let kind = current.kind();
+            if current.start_position().row != current.end_position().row
+                && (kind.contains("string") || kind.contains("comment") || kind.contains("heredoc"))
```
SomeoneToIgnore: Unfortunately, I do not have a good context on this part, but this seems suspicious: we have not that many hardcoded entities used, and adding more (esp. such "non-generic" as `heredoc`) seems like adding a code debt.

Can we extract those into constants and maybe a separate function?


Also, some comment explaining why these 3 is good to add (e.g. I'm confused by the absence of `string_literal` but that is ok, it seems?).


Do we really need `heredoc`(this is bash-related only, is it not?) and why this is not a `comment` kind?

SomeoneToIgnore (commented):
Thank you, let's make the new heuristics a bit more self-documented in the code and merge this.

crates/editor/src/display_map.rs
notJoon: I had exactly the same thought when adding this part. Since `Node.kind` is of type `&str`, I wrote it with hardcoding as it is now. As you suggested, I extracted it into a separate function and added comments. I also changed from exact match to substring match to handle cases where names differ by language as much as possible.

https://github.com/zed-industries/zed/pull/51502/commits/91c3f6f50f033242550dec7422734b5b00984ecc

For `heredoc`, I added it because fold didn't work correctly in bash. Also this node kinds (like `heredoc_body`) don't contain "string" or "cooment" in their names, so they need a separate pattern. It's also used in Ruby, Perl, and PHP grammars.

PR #51507, opened 2026-03-13
languages: Update J/TSX component highlighting to use semantic `tag.component` scope

yeskunall:
Changes JSX component identifiers from generic `@type` to the more specific `@tag.component` scope in both JavaScript and TypeScript syntax highlighting. This allows themes to style React/JSX components distinctly from regular types.

The only thing to be aware of is the **visual change for JSX/TSX users**: components that were previously teal (via `@type`) will now inherit whatever their theme's `tag` color is. For themes like Gruvbox, and One (perhaps other built-in themes too?) that define `tag` with a color, components will pick that up. For themes that leave `tag` unstyled (including the fallback theme), components will lose their teal color. That’s a deliberate trade-off with these changes, something I also mention in https://github.com/zed-industries/zed/issues/51478#issuecomment-4056409919.

Closes #51478.

| Before | After |
|--------|--------|
| image | image |

Ayu Light

| Before | After |
|--------|--------|
| image | image | 

One Dark

| Before | After |
|--------|--------|
| image | image | 

Gruvbox Light

| Before | After |
|--------|--------|
| image | image | 

Release Notes:

- Updated J/TSX component highlighting to use semantic `tag.component` scope

### DIFF
--- a/crates/languages/src/javascript/highlights.scm
+++ b/crates/languages/src/javascript/highlights.scm
@@ -328,26 +328,26 @@
 ; JSX elements
 (jsx_opening_element
   [
-    (identifier) @type
+    (identifier) @tag.component
     (member_expression
-      object: (identifier) @type
-      property: (property_identifier) @type)
+      object: (identifier) @tag.component
+      property: (property_identifier) @tag.component)
   ])
 
 (jsx_closing_element
   [
-    (identifier) @type
+    (identifier) @tag.component
     (member_expression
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
I do not think we should proceed with this before we have a proper fallback mechanism in place here. https://github.com/zed-industries/zed/tree/highlight-map-better-resolution is working towards that direction and will include such a mechanism

PR #51511, opened 2026-03-13
markdown_preview: Add preview font size multiplier

dive:
Markdown preview currently always follows `buffer_font_size`, which makes it hard to make preview content more readable without also changing editor text size.

This adds a dedicated `markdown_preview_font_size_multiplier` setting so users can increase or decrease markdown preview text size independently while keeping the existing preview rendering behaviour unchanged.

> [!NOTE]
> Why not `markdown_preview_font_size`? I thought that a multiplier is better because markdown preview should stay relative to `buffer_font_size`, not become a separate typography system.
>
> - stays in sync when users change editor font size
> - expresses the real intent: "make preview a bit bigger/smaller than editor text"
> - avoids another absolute font-size setting to manage
>
> But, perhaps, I am wrong with my assumptions.

**What changed:**

- added `markdown_preview_font_size_multiplier` to theme settings
- documented the setting in default settings
- applied the multiplier to markdown preview font sizing
- kept the implementation minimal & avoided changing the preview rendering model

Release Notes:

- Improved markdown preview font size customisation via `markdown_preview_font_size_multiplier` setting

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -74,6 +74,10 @@
   "agent_ui_font_size": null,
   // The default font size for user messages in the agent panel.
   "agent_buffer_font_size": 12,
+  // Scales markdown preview typography relative to the buffer font size.
+  // 1.0 uses the buffer font size unchanged, 1.2 makes it 20% larger,
+  // and 0.8 makes it 20% smaller.
+  "markdown_preview_font_size_multiplier": 1.0,
   // How much to fade out unused code.
   "unnecessary_code_fade": 0.3,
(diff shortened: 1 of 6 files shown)


dinocosta (commented):
I’ve assigned Danilo, as I trust his input more on this one, but I do wonder if it wouldn’t be enough to update the Markdown preview to actually use ‎`ui_font_size`. I’m not sure why we went with ‎`buffer_font_size`, as using the same font size as the UI makes more sense to me 🤔

PR #51528, opened 2026-03-13
Filter common git hosting providers from SSH config hosts

### LINKED ISSUE #50218: Remote Dev: github.com listed as SSH host, fails with confusing error
## Summary

Zed's Remote Development feature lists `github.com` as an available SSH host (presumably picked up from `~/.ssh/known_hosts` or `~/.ssh/config`). Selecting it and clicking "Open Folder" fails with a confusing error because GitHub's SSH endpoint only supports Git protocol, not shell access.

## Steps to Reproduce

1. Open Zed → Remote Development
2. `github.com` appears in the SSH host list
3. Select it and click "Open Folder"

## Error

```
Failed to run 'uname -sm' to determine platform: failed to run command
Command { program: "ssh", args: [..., "github.com", "-T", "cd; uname -sm"], ... }:
Invalid command: cd; uname -sm
You appear to be using ssh to clone a git:// URL.
Make sure your core.gitProxy config option and the
GIT_PROXY_COMMAND environment variable are NOT set.
```

## Expected Behavior

Either:
- **Filter out** hosts that are known Git-only services (github.com, gitlab.com, bitbucket.org, etc.) from the remote SSH host list, or
- **Gracefully handle** the connection failure with a clear message like "This host does not support shell access" instead of showing the raw error

## Environment

- Zed: latest stable (macOS)
- macOS Sequoia
- `github.com` present in `~/.ssh/known_hosts` from normal Git-over-SSH usage

sarmadgulzar:
Fixes https://github.com/zed-industries/zed/issues/50218

- For every `Host` block: collect its host patterns (handling continuations and filtering `!`/`*/\`), remember any `Hostname`.
- When the block ends: if `Hostname` is a git provider → discard everything.
- Else if `Hostname` exists → keep all host aliases.
- Else → keep only the host aliases that are not git providers.

Added more extensive test coverage as well to cover the above!

Release Notes:

- Fixed SSH host picker showing git provider domains (e.g. github.com, gitlab.com) from SSH config.

### DIFF
--- a/crates/recent_projects/src/ssh_config.rs
+++ b/crates/recent_projects/src/ssh_config.rs
@@ -1,5 +1,25 @@
 use std::collections::BTreeSet;
 
+const FILTERED_GIT_PROVIDER_HOSTS: &[&str] = &[
+    "dev.azure.com",
+    "bitbucket",
+    "bitbucket.org",
+    "chromium.googlesource.com",
+    "codeberg",
+    "codeberg.org",
+    "gitea",
+    "gitea.com",
+    "gitee",
+    "gitee.com",
+    "github",
+    "github.com",
+    "gitlab",
+    "gitlab.com",
+    "sourcehut",
+    "sourcehut.org",
+    "git.sr.ht",
+];
+
 pub fn parse_ssh_config_hosts(config: &str) -> BTreeSet<String> {
     let mut hosts = BTreeSet::new();
     let mut needs_another_line = false;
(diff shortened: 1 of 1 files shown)


crates/recent_projects/src/ssh_config.rs
```diff
@@ -1,5 +1,25 @@
...
+    "bitbucket.org",
+    "chromium.googlesource.com",
+    "codeberg",
+    "codeberg.org",
+    "gitea",
+    "gitea.com",
+    "gitee",
+    "gitee.com",
+    "github",
+    "github.com",
```
injust, line 10: If hard-coding, then you probably also want `gist.github.com`.

nathansobo:
Thanks for this, @sarmadgulzar! I pushed a revision with some structural changes:

- **Two-pass design**: `parse_host_blocks` handles all the SSH config parsing (keywords, backslash continuations, block boundaries) and returns a `Vec<HostBlock>`. The public function then just filters and collects.

- **`HostBlock::non_git_provider_hosts`**: Moved the filtering logic onto the block as a method.

- **Simplified the three-branch filter to one predicate**: The three cases are all equivalent to checking `is_git_provider_domain(hostname.unwrap_or(alias))` per alias.

- **Renamed `parse_hosts_from` → `parse_hosts`**.

All existing tests pass unchanged. Let me know what you think!

nathansobo:
I'm gonna go ahead and merge this, but feel free to leave any feedback on my additions, and we can fix forward if there's anything I did that we want to change.

JosephTLyons:
Thank you @sarmadgulzar!

PR #51547, opened 2026-03-14
git_ui: Move commit messages out of split commit diffs

cppcoffee:
Release Notes:

- Added Render commit messages in a separate read-only editor and keep the SplittableEditor focused on file diffs.

https://github.com/user-attachments/assets/dc98851d-7f9a-48a4-ad71-2982d978844f


cole-miller:
Thanks for opening a PR! We ended up going with a different design in #58274, and the split diff is now supported in the commit view.

PR #51550, opened 2026-03-14
Fix multi-character bracket auto-close in language injection contexts

RomainDW:
## Summary

- Fixes multi-character bracket pairs (e.g., `{{`/`}}`, `{%`/`%}`) not auto-closing in files with language injections
- Currently, only the deepest (injected) language's brackets are checked. For example, in Liquid files where `template_content` is injected as HTML, typing `{` triggers HTML's `{`/`}` pair, and the subsequent `{` or `%` never forms the intended `{{`/`}}` or `{%`/`%}` pair
- This affects any language that uses multi-character brackets and has injected content (Liquid, potentially Handlebars, Jinja, etc.)

### Changes

1. **`crates/language/src/buffer.rs`**: Added `language_scopes_at()` method that returns all language scopes at a position (deepest first), complementing the existing `language_scope_at()` which returns only the deepest
2. **`crates/multi_buffer/src/multi_buffer.rs`**: Added delegation for the new method
3. **`crates/editor/src/editor.rs`**: Modified bracket matching in `handle_input()` to:
   - Collect bracket pairs from all scopes at the cursor position
   - Prefer longer bracket pairs from outer scopes over shorter ones from the deepest scope (so Liquid's `{{` wins over HTML's `{`)
   - Only check outer scopes when the deepest scope already has a match, preventing bracket leaking (HTML's `<`/`>` won't auto-close inside `<script>` JavaScript)
   - Replace stale auto-closed brackets when a multi-character pair supersedes a previous single-character auto-close

## Test plan

- [x] All 28 existing `autoclose` tests pass
- [x] All 18 existing `bracket` tests pass  
- [x] `test_autoclose_with_embedded_language` specifically validates that injected language bracket scoping still works correctly (HTML `<`/`>` does not leak into JavaScript)
- [ ] Manual testing with Liquid files: `{{`, `{%`, `{{-`, `{%-` should auto-close correctly inside `template_content` (HTML injection zones)

### Repro steps (before fix)
1. Open a `.liquid` file with a Zed extension that injects HTML into `template_content`
2. Type `{{` — get `{{}` instead of `{{ | }}`
3. Type `{%` — get `{%}` instead of `{% | %}`

### Expected (after fix)
1. `{{` → `{{|}}` (cursor between)
2. `{%` → `{%|%}` (cursor between)

## Release Notes

- Fixed multi-character bracket pairs (e.g., `{{`/`}}`) not auto-closing correctly in files with language injections, such as Liquid templates with HTML injection

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -4591,48 +4591,96 @@ impl Editor {
             }
             all_selections_read_only = false;
 
-            if let Some(scope) = snapshot.language_scope_at(selection.head()) {
+            {
+                let all_scopes = snapshot.language_scopes_at(selection.head());
+                // Use the deepest scope for autoclose-before checks
+                let scope = snapshot.language_scope_at(selection.head());
+
                 // Determine if the inserted text matches the opening or closing
                 // bracket of any of this language's bracket pairs.
+                // We collect pairs from ALL scopes and prefer the longest match.
                 let mut bracket_pair = None;
                 let mut is_bracket_pair_start = false;
                 let mut is_bracket_pair_end = false;
                 if !text.is_empty() {
+                    // First, find the best match from the deepest scope only.
+                    let mut deepest_start_match: Option<BracketPair> = None;
+                    let mut deepest_start_len: usize = 0;
                     let mut bracket_pair_matching_end = None;
                     // `text` can be empty when a user is using IME (e.g. Chinese Wubi Simplified)
                     //  and they are removing the character that triggered IME popup.
-                    for (pair, enabled) in scope.brackets() {
-                        if !pair.close && !pair.surround {
-                            continue;
-                        }
+                    if let Some(deepest_scope) = all_scopes.first() {
+                        for (pair, enabled) in deepest_scope.brackets() {
+                            if !pair.close && !pair.surround {
+                                continue;
+                            }
 
-                        if enabled && pair.start.ends_with(text.as_ref()) {
-                            let prefix_len = pair.start.len() - text.len();
-                            let preceding_text_matches_prefix = prefix_len == 0
-                                || (selection.start.column >= (prefix_len as u32)
-                                    && snapshot.contains_str_at(
-                                        Point::new(
-                                            selection.start.row,
-                                            selection.start.column - (prefix_len as u32),
-                                        ),
-                                        &pair.start[..prefix_len],
-                                    ));
-                            if preceding_text_matches_prefix {
-                                bracket_pair = Some(pair.clone());
-                                is_bracket_pair_start = true;
-                                break;
+                            if enabled && pair.start.ends_with(text.as_ref()) {
+                                let prefix_len = pair.start.len() - text.len();
+                                let preceding_text_matches_prefix = prefix_len == 0
+                                    || (selection.start.column >= (prefix_len as u32)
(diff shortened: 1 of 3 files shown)


crates/editor/src/editor.rs
```diff
@@ -4591,48 +4591,96 @@ impl Editor {
             }
             all_selections_read_only = false;
 
-            if let Some(scope) = snapshot.language_scope_at(selection.head()) {
+            {
+                let all_scopes = snapshot.language_scopes_at(selection.head());
+                // Use the deepest scope for autoclose-before checks
+                let scope = snapshot.language_scope_at(selection.head());
```
osiewicz, line 4597: Can we grab the innermost scope from `all_scopes`? This function is not free by any means and you're on a hot path there.

crates/editor/src/editor.rs
```diff
@@ -4591,48 +4591,96 @@ impl Editor {
...
+                    let mut deepest_start_match: Option<BracketPair> = None;
+                    let mut deepest_start_len: usize = 0;
                     let mut bracket_pair_matching_end = None;
                     // `text` can be empty when a user is using IME (e.g. Chinese Wubi Simplified)
                     //  and they are removing the character that triggered IME popup.
-                    for (pair, enabled) in scope.brackets() {
-                        if !pair.close && !pair.surround {
-                            continue;
-                        }
+                    if let Some(deepest_scope) = all_scopes.first() {
```
osiewicz, line 4612: Maybe we don't even need a separate `scope` variable, if we're intermixing it with `all_scopes.first()` anyways

crates/editor/src/editor.rs
```diff
@@ -4648,10 +4696,12 @@ impl Editor {
                             // If the inserted text is a suffix of an opening bracket and the
                             // selection is preceded by the rest of the opening bracket, then
                             // insert the closing bracket.
-                            let following_text_allows_autoclose = snapshot
-                                .chars_at(selection.start)
-                                .next()
-                                .is_none_or(|c| scope.should_autoclose_before(c));
+                            let following_text_allows_autoclose = scope.as_ref().is_none_or(|s| {
```
osiewicz, line 4699: Should you allow auto-closing when there's no scope? Not that it's a possible scenario, but this line reads a bit weird to me

crates/language/src/buffer.rs
```diff
@@ -3958,6 +3958,74 @@ impl BufferSnapshot {
...
+        if scopes.is_empty() {
+            if let Some(language) = self.language.clone() {
+                scopes.push(LanguageScope {
+                    language,
+                    override_id: None,
+                });
+            }
+        }
+
+        scopes
```
osiewicz, line 4026: This whole function duplicates `language_scope_at`; I think it'd be nice to have an inner function that returns `impl Iterator<>` that is then `.collect::<Vec<_>>()`'d by `language_scopes_at` or collected into an Option by `language_scope_at`. Besides, we should also rename `language_scope_at` to `innermost_language_scope_at`.

osiewicz (changes requested):
It'd be nice to get some test coverage on the new behaviour.

smitbarmase:
Hey, since it's been stale for a while and I haven't received any updates on this, I'm going to close it. Feel free to open a new one with the suggested changes if you still want to get this in. Thanks.

PR #51553, opened 2026-03-14
Add deleted color to project panel

tidefield:
When a file is deleted, the parent folder isn't colored which makes it hard to know unless we look at the Git Panel.

This PR adds `Color::Deleted` to fix that.

Screenshot 2026-03-14 at 7 56 04 PM

Release Notes:

- Improved project panel to color folder with deleted file

### DIFF
--- a/crates/editor/src/items.rs
+++ b/crates/editor/src/items.rs
@@ -1864,6 +1864,8 @@ pub fn entry_git_aware_label_color(git_status: GitSummary, ignored: bool, select
         Color::Ignored
     } else if git_status.conflict > 0 {
         Color::Conflict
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (approved):
Thank you.

lilly-lizard:
> I don't know if adding a setting is reasonable unless other people (and Zed team) agree.

@tidefield evidence of people agreeing with @ghost can be seen in the upvotes of their discussion here: https://github.com/zed-industries/zed/discussions/52619

The key reasoning is that seeing folders highlighted red usually implies there are **errors** all over the project, especially for people moving over from vscode.

Issue #50269, opened 2026-02-27
Zed editor freezes and is not clickable


jehof:
### Reproduction steps

Since the latest update (The thing with the `wgpu` update) the zed editor freezes and is not clickable or usable anymore. Even with shortcuts it is not closable. I have to use htop / pkill to exit zed.

I have not figured out when it exactly happens.

- After update, i want to change configuration of AI things (Settings->Rules). Open up dialog => freeze
- Yesterday doing normal work and opening multiple windows --> freeze
- Today opening zed and it restores my sessions (3 Windows). Dragging windows to other monitor -> freeze

How can i help you to improve the zed editor in that case? I am not a linux poweruser so provide me some guide. I like the zed editor very well, but currently it is buggy and not usable.

### Zed version and system specs

Zed: v0.225.9/0.225.10
OS: Pop!_OS 22.04 LTS (Ubuntu) 20Linux 6.17.9-76061709-generic #202511241048~1764704751~22.04~b24b425 SMP PREEMPT_DYNAMIC Tue D x86_64 x86_64 x86_64 GNU/Linux
Memory: 32 GB
3 Display-Setup (X11)

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

jehof:
I updated to v0.225.10 today and check how this works.

yeskunall:
Last 50 lines of the trace file (after pretty-printing it):

<details><summary>Details</summary>
<p>

```jsonc
[
  {
    "location": {
      "file": "crates/editor/src/blink_manager.rs",
      "line": 73,
      "column": 20
    },
    "start": 27239023458,
    "duration": 100469
  },
  {
    "location": {
      "file": "crates/settings_ui/src/settings_ui.rs",
      "line": 1475,
      "column": 12
    },
    "start": 27503298531,
    "duration": 7374
  },
  {
    "location": {
      "file": "crates/editor/src/editor.rs",
      "line": 7129,
      "column": 42
    },
    "start": 27503306175,
    "duration": 7254
  },
  {
    "location": {
      "file": "crates/editor/src/highlight_matching_bracket.rs",
      "line": 41,
      "column": 60
    },
    "start": 27503313669,
    "duration": 3677
  },
  {
    "location": {
      "file": "crates/editor/src/blink_manager.rs",
      "line": 51,
      "column": 12
    },
    "start": 27503317517,
    "duration": 2935
  },
  {
    "location": {
[...]

PR #51561, opened 2026-03-14
gpui_wgpu: Release surface resources during destroy

AxXxB:
## Summary

Implements the fix proposed by @israelrios in https://github.com/zed-industries/zed/issues/50269#issuecomment-4049006746.

The issue comment narrowed the hang down to X11 window teardown:

- when the Settings window is closed on X11, the UI thread can stall while `WgpuRenderer` is dropped during final `X11WindowState` teardown
- at that point, the X11 window has already been destroyed
- the proposed fix is to make `WgpuRenderer::destroy()` release surface-bound GPU resources eagerly so the native window is destroyed only after the `wgpu::Surface` and related resources are gone

This PR applies that change in `crates/gpui_wgpu/src/wgpu_renderer.rs` by calling `self.resources.take()` inside `destroy()`.

## Validation

I confirmed this fix locally on Linux Mint 21.3 Cinnamon by applying it to the `v0.227.1` build:

- opening and closing Settings no longer causes the main window to hang

## References

- Refs #50269
- Original analysis and fix proposal by @israelrios: https://github.com/zed-industries/zed/issues/50269#issuecomment-4049006746

Release Notes:

- Fixed a Linux/X11 issue where closing the Settings window could cause Zed to hang.

### DIFF
--- a/crates/gpui_wgpu/src/wgpu_renderer.rs
+++ b/crates/gpui_wgpu/src/wgpu_renderer.rs
@@ -1609,7 +1609,9 @@ impl WgpuRenderer {
     }
 
     pub fn destroy(&mut self) {
-        // wgpu resources are automatically cleaned up when dropped
+        // Release surface-bound GPU resources eagerly so the underlying native
+        // window can be destroyed before the renderer itself is dropped.
+        self.resources.take();
     }
 
     /// Returns true if the GPU device was lost and recovery is needed.


AxXxB:
I've added Release Notes, but I'm not sure this PR fully resolves the referenced issue, since users seem to be reporting multiple similar problems there.

Veykril (approved):
Thanks!

PR #51566, opened 2026-03-14
git_ui: Improve repository clone suggestions

itsfuad:
## Summary

- Add provider-neutral repository search to `GitHostingProvider`, implemented for public GitHub and GitLab.
- Keep repository search disabled by default and require explicit per-provider opt-in before making any search request. Enabled providers persist through `git.repository_search_providers`.
- Search enabled providers concurrently, debounce subsequent queries, cache and deduplicate results, and ignore stale async responses when the query changes.
- Surface provider identity, loading, and error states while preserving direct URL, SSH, and local-path cloning.

https://github.com/user-attachments/assets/87762db0-03cc-41bc-ab0c-6e4558d9c491

Settings UI:
image

## Testing

- `cargo test -p git_hosting_providers providers::gitlab::tests -- --nocapture`
- `cargo test -p git_ui clone_suggestions -- --nocapture`
- `cargo check -p git_ui`
- Formatting checks
- `git diff --check`

Release Notes:

- Improved repository cloning with opt-in hosting-provider suggestions.


itsfuad:
Addressed issues are resolved

mealicecooper:
Ah.. Why isnt that added to zed yet! I am too much lazy to remember the repo url or copy the url from browser. 
@danilo-leal @cole-miller  Do you have plan to add this feature?

itsfuad:
@mealicecooper Thank you for the concern. As they are a big team. It may take time before it gets merged. Or they might implement this feature in a more better way than me and close this PR. 
So we can only wait for their review.

fujiiai1:
why this pr has no update? it is a simple basic feature all would love to have 😞

itsfuad:
@Anthony-Eid Sorry to mention, is there any update for the PR? 
Merge conflicts will get more complicated as time goes on.

itsfuad:
I did some improvements on the code and resolved the merge conflicts. Hope this is good to go now.

crates/git_ui/src/clone_suggestions.rs
```diff
@@ -0,0 +1,347 @@
...
+        return vec![
+            CloneSuggestion {
+                title: "Clone GitHub repository".into(),
+                detail: github_url.clone().into(),
+                repo_url: github_url.into(),
+            },
+            CloneSuggestion {
+                title: "Clone entered location".into(),
+                detail: input.to_owned().into(),
+                repo_url: input.to_owned().into(),
```
tidely: title should use `SharedString::new_static`. This allows us to reference the static string instead of allocating a new one even though the string is always the same.

detail and repo_url should call `.into()` directly without `to_owned()`, or use `SharedString::new`. This removes an unnecessary intermediate allocation.

These changes should apply to all instances of `CloneSuggestion`

crates/git_ui/src/clone_suggestions.rs
```diff
@@ -0,0 +1,347 @@
...
+) -> anyhow::Result<Vec<CloneSuggestion>> {
+    let request = github_search_request(query, github_token().as_deref())?;
+    let mut response = http_client
+        .send(request)
+        .await
+        .context("requesting GitHub repository suggestions")?;
+    let status = response.status();
+    if !status.is_success() {
+        anyhow::bail!("GitHub repository search returned HTTP {}", status.as_u16());
+    }
```
tidely:
```suggestion
    anyhow::ensure!(
        status.is_success(),
        "GitHub repository search returned HTTP {}",
        status,
    );
```

tidely (commented):
Just had some nits to keep code quality high, nothing blocking

crates/git_ui/src/clone_suggestions.rs
itsfuad: Addressed in c2248d9708 by using static SharedString titles and avoiding intermediate owned strings across CloneSuggestion construction.

crates/git_ui/src/clone_suggestions.rs
itsfuad: Applied the suggested anyhow::ensure! form in c2248d9708.

MrSubidubi (changes requested):
Sorry, we have too much to look at, but I think this is a worthwhile improvement. 

That said, I have some big concerns here currently
- there is not at all an indicator we are currently loading results
- the implementation currently is too hardcoded to GitHub


Now, I do not want you to add support to this for all hosting providers, but we have to wire this through the supported hosting providers and resolve this through a method there. What we then could do is have some context menu somewhere in the UI with config controls so the user can select which providers to query. 


Then, I do admit this will blow scope here a lot. Curious about whether and how we could perhaps chunk this? After all, interested in your thougths on this either way.

tidely:
During my earlier review I thought about suggesting to make this a method on the `GitHostingProvider` trait (`crates/git/src/hosting_provider.rs`). This could have a default implementation and such providers can be added incrementally. This would also let us move networking code out of the `git_ui` crate.

I agree that this feature should not make any network requests without the user opting into it first.

itsfuad:
Addressed the latest hosting-provider and opt-in feedback in `70fdc6bcee`:

- Repository search is now a defaulted capability on `GitHostingProvider`, so providers can adopt it incrementally.
- GitHub request construction, authentication, parsing, and errors moved out of `git_ui` into `git_hosting_providers`.
- Provider search is disabled by default and makes no request until the user enables a provider from the clone picker menu.
- The picker now exposes loading and error states and retains debouncing, per-provider caching, and deduplication.
- The placeholder no longer advertises ambiguous `owner/repo` input, and the loading label no longer duplicates its animated ellipsis.

Validated with the focused GitHub and clone-suggestion tests, the full affected-crate test suites, clippy for all affected crates, and `git diff --check`.

The PR description is updated.

crates/git_hosting_providers/src/providers/github.rs
```diff
@@ -163,6 +199,64 @@ impl Github {
...
+            .append_pair("per_page", &MAX_REPOSITORY_SEARCH_RESULTS.to_string());
+
+        let mut request = Request::get(url.as_str())
+            .header("Accept", "application/vnd.github+json")
+            .header("X-GitHub-Api-Version", "2022-11-28")
+            .follow_redirects(http_client::RedirectPolicy::FollowAll)
+            .timeout(REPOSITORY_SEARCH_TIMEOUT);
+        if let Some(token) = github_token() {
+            request = request.header("Authorization", format!("Bearer {token}"));
+        }
```
Anthony-Eid: Is it safe from a security perspective to be seeing someone's github token in a request that could contain redirects?

crates/git_ui/src/git_ui.rs
```diff
@@ -1267,90 +1282,434 @@ impl Component for GitStatusIcon {
...
+        let has_enabled_provider = providers.iter().any(|provider| {
+            self.enabled_provider_urls
+                .contains(provider.base_url().as_str())
+        });
+
+        Some(
+            PopoverMenu::new("repository-search-providers")
+                .with_handle(self.provider_menu_handle.clone())
+                .trigger_with_tooltip(
+                    IconButton::new("repository-search-providers-trigger", IconName::Settings)
```
Anthony-Eid, line 1425: I think we should use a different icon here, perhaps the search glass. @danilo-leal Is there an icon you would recommend that control what hosting providers a git clone picker would use?

crates/git_ui/src/git_ui.rs
```diff
@@ -1267,90 +1282,434 @@ impl Component for GitStatusIcon {
...
+                                                        .delegate
+                                                        .enabled_provider_urls
+                                                        .remove(&provider_url);
+                                                } else {
+                                                    picker
+                                                        .delegate
+                                                        .enabled_provider_urls
+                                                        .insert(provider_url.clone());
+                                                }
+                                                cx.default_global::<
```
Anthony-Eid: I'm not sure how I feel about this being a global. Could the picker just get this information whenever it's created? Maybe there's a different owner that would be better as well

crates/git_ui/src/git_ui.rs
```diff
@@ -1267,90 +1282,434 @@ impl Component for GitStatusIcon {
...
+            return Task::ready(());
+        }
+
+        self.is_searching = true;
+        cx.notify();
+        let http_client = self.http_client.clone();
+        cx.spawn_in(window, async move |picker, cx| {
+            cx.background_executor()
+                .timer(REPOSITORY_SEARCH_DEBOUNCE)
+                .await;
```
Anthony-Eid: nit: For an initial search maybe we should avoid this debounce. Not a merge blocker but an idea

crates/git_ui/src/git_ui.rs
```diff
@@ -1267,90 +1282,434 @@ impl Component for GitStatusIcon {
...
+                .child(status)
+                .child(
+                    Button::new("learn-more", "Learn More")
+                        .label_size(LabelSize::Small)
+                        .end_icon(Icon::new(IconName::ArrowUpRight).size(IconSize::XSmall))
+                        .on_click(|_, _, cx| {
+                            cx.open_url("https://github.com/git-guides/git-clone");
+                        }),
+                )
+                .into_any_element(),
```
Anthony-Eid, line 1689: I don't think we need this learn more button about git clone. Willing to have my mind change tho!

Anthony-Eid (changes requested):
The main thing I would like to talk about is the token being part of the GitHub request. Besides that, there's nothing blocking this from being merged soon 😃

crates/git_ui/src/git_ui.rs
itsfuad: It was already part of the existing clone UI and is still present in current versions, so I kept it for consistency with the existing design. I’m happy to remove it if you think the picker works better without it.

Screenshot From 2026-09-14 19-11-09

Anthony-Eid: oh if that's the case we should definitely leave it! I do wonder if we should right align it tho

Thank you for pointing that out

crates/git_ui/src/git_ui.rs
itsfuad: I guess it doesn't look that well. I kept current gear icon. Will change if you ask.

image

image

itsfuad:
Addressed the latest review feedback and updated the PR recording to demonstrate the current flow.

- Merged the latest `main`.
- Prevented authenticated GitHub search requests from following redirects.
- Made enabled-provider state local to each clone picker.
- Made the initial provider search immediate while retaining debounce for subsequent input.
- Kept the gear icon because it more clearly represents provider configuration.
- Retained the existing Learn More action and kept it right-aligned based on the follow-up discussion.

Validation completed with focused GitHub provider and clone-suggestion tests, formatting checks, `git diff --check`, and clippy for the affected crates.

This should be ready for another review. Thank you!

itsfuad:
Follow-up on persistence and provider scope: enabled repository-search providers now persist through the existing settings system (`git.repository_search_providers`) instead of resetting when the clone picker is recreated. I kept this as a user setting rather than internal picker key-value state because it represents explicit network-search opt-in and may eventually need to coexist with other non-secret provider configuration.

I also extended this PR to include public GitLab repository search in addition to GitHub. Both providers go through the same `GitHostingProvider` repository-search capability and remain disabled until explicitly enabled by the user. GitLab `namespace/repository` queries opt into namespace matching only when the query contains `/`, and self-hosted instances remain unsupported for search without making a request.

The multi-provider path now also guards against stale async search responses and shows the provider alongside each result. Bitbucket can still be added incrementally through the same provider interface as follow-up work.

Additional validation for the GitLab follow-up:
- `cargo test -p git_hosting_providers providers::gitlab::tests -- --nocapture` — 21 tests
- `cargo check -p git_ui`
- `cargo test -p git_ui clone_suggestions -- --nocapture` — 3 tests
- formatting checks and `git diff --check`

PR #51569, opened 2026-03-14
gpui: Fix emoji rendering in SVG preview

alanpjohn:
Closes #50483 

## Findings

As reported in the original issue, emojis in SVG preview were not rendering consistently with the editor. 

The SVG renderer uses `usvg`/`resvg` for parsing and rendering SVG files. The first problem was that emoji fonts were not rendering at all, which was fixed by enabling the `raster_images` on `resvg`.

Beyond that it was observed that the default font fallback mechanism in `usvg` searches through the font database alphabetically without prioritizing emoji fonts. This caused emojis to sometimes render in non-emoji fonts that happened to contain glyph mappings for those characters.

For example, on Linux systems with the default `uvsg::FontResolver::default_fallback_selector()`:
- The character ✅ would fall back to `FreeSerif` (monochrome)
- Instead of `Noto Color Emoji` (full color)

Log output showed the inconsistent behavior:
```
WARN  [usvg::text] Fallback from FreeSans to Noto Color Emoji.
WARN  [usvg::text] Fallback from FreeSans to FreeSerif.
WARN  [usvg::text] Fallback from FreeSans to Noto Color Emoji.
```

Image

This created a jarring inconsistency where the same emoji character would render differently in:
- The editor (correct, using platform emoji fonts)
- SVG preview (incorrect, using arbitrary fallback fonts)
## Solution

If the specified font in SVG is available on the system, we should show that. If not, we should fallback to what editors show today for that emoji.

This PR implements emoji-aware font fallback that:

1. **Enabled `raster_images` build feature** to render emojis in SVG.
2. **Detects emoji characters** using Unicode emoji properties (via `\p{Emoji}` regex pattern), consistent with how we check for emoji in the Editor as well.
3. **Preserves user-specified fonts** by only intervening when the default font resolver would use a non-emoji font for emoji characters

### Font Family Selection

I avoided completely reusing/rebuilding the logic for emoji font selection used by the editor as `uvsg` internally does quite a bit of the job and it felt like overcomplicating the solution. Instead using hard coded platform specific font family names.
The hardcoded emoji font families are sourced from Zed's existing platform-specific text rendering systems:

- **macOS**: `Apple Color Emoji`, `.AppleColorEmojiUI`  
  Source: https://github.com/zed-industries/zed/blob/db622edc8b26bd138c91027a02792a84c083acbf/crates/gpui_macos/src/text_system.rs#L353-L359
- **Linux/FreeBSD**: `Noto Color Emoji`, `Emoji One`  
  Source: https://github.com/zed-industries/zed/blob/db622edc8b26bd138c91027a02792a84c083acbf/crates/gpui_wgpu/src/cosmic_text_system.rs#L642-L646
- **Windows**: `Segoe UI Emoji`, `Segoe UI Symbol`  
  Source: Standard Windows emoji font stack

These match the fonts the editor uses for emoji rendering on each platform.
To break down further into the similarity and differences in the emoji font resolution:
**Similarities:**
- Both now use the regex based emoji detection logic
- Both prioritize the same platform-specific emoji font families
- Both support color emoji rendering
**Differences:**
- **Editor**: Uses platform-native text shaping (CoreText on macOS, DirectWrite on Windows, cosmic-text on Linux) which handles fallback automatically
- **SVG**: Uses custom fallback selector that explicitly queries emoji fonts first, then falls back to default usvg behavior
## Testing

- Added unit tests for `is_emoji_character` in `util` crate
- Tested emoji detection for various Unicode characters
- [ ] Verified platform-specific font lists compile correctly (Only linux done)
- [ ] Manual testing with SVG files containing emojis on all platforms (Only linux done)

Release Notes:

- Fixed SVG preview to render emojis consistently with the editor by prioritizing platform-specific color emoji fonts

### DIFF
--- a/crates/util/src/util.rs
+++ b/crates/util/src/util.rs
@@ -727,6 +727,13 @@ pub fn word_consists_of_emojis(s: &str) -> bool {
     prev_end == s.len()
 }
 
+/// Check if a character is an emoji using Unicode properties.
+pub fn is_emoji_character(c: char) -> bool {
+    let mut buf = [0u8; 4];
+    let s = c.encode_utf8(&mut buf);
+    emoji_regex().is_match(s)
+}
+
 /// Similar to `str::split`, but also provides byte-offset ranges of the results. Unlike
 /// `str::split`, this is not generic on pattern types and does not return an `Iterator`.
 pub fn split_str_with_ranges<'s>(
@@ -966,6 +973,27 @@ mod tests {
 
         for (text, expected_result) in words_to_test {
             assert_eq!(word_consists_of_emojis(text), expected_result);
+        }
+    }
+
+    #[test]
+    fn test_char_is_emoji() {
+        let chars_to_test = vec![
+            ('👋', true),
+            ('✅', true),
+            ('©', true),
+            ('♥', true),
+            ('a', false),
+            ('Z', false),
+            ('1', false),
+            (' ', false),
+            ('漢', false),
+            ('中', false),
+            ('カ', false),
+        ];
+
+        for (text, expected_result) in chars_to_test {
+            assert_eq!(is_emoji_character(text), expected_result);
         }
     }
 
--- a/crates/util/src/util.rs
+++ b/crates/util/src/util.rs
@@ -727,13 +727,6 @@ pub fn word_consists_of_emojis(s: &str) -> bool {
     prev_end == s.len()
 }
 
-/// Check if a character is an emoji using Unicode properties.
-pub fn is_emoji_character(c: char) -> bool {
-    let mut buf = [0u8; 4];
-    let s = c.encode_utf8(&mut buf);
-    emoji_regex().is_match(s)
-}
-
 /// Similar to `str::split`, but also provides byte-offset ranges of the results. Unlike
 /// `str::split`, this is not generic on pattern types and does not return an `Iterator`.
 pub fn split_str_with_ranges<'s>(
@@ -973,27 +966,6 @@ mod tests {
 
         for (text, expected_result) in words_to_test {
             assert_eq!(word_consists_of_emojis(text), expected_result);
-        }
-    }
-
-    #[test]
-    fn test_char_is_emoji() {
-        let chars_to_test = vec![
-            ('👋', true),
-            ('✅', true),
-            ('©', true),
-            ('♥', true),
-            ('a', false),
-            ('Z', false),
-            ('1', false),
-            (' ', false),
-            ('漢', false),
-            ('中', false),
-            ('カ', false),
-        ];
-
-        for (text, expected_result) in chars_to_test {
-            assert_eq!(is_emoji_character(text), expected_result);
         }
     }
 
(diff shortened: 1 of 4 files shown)


crates/util/src/util.rs
```diff
@@ -969,6 +976,27 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_char_is_emoji() {
```
smitbarmase: This test is failing for me. Can you check: `cargo test -p util test_char_is_emoji`?

smitbarmase (commented):
Thanks, this is good improvement overall.

I have two concerns:

1. `util::is_emoji_character` does not seem like the right predicate for font fallback. The added unit test also fails locally for me. I think it fails on `1`. 

We should verify these too in addition to existing ones:

```rs
  [
      ("©", false),
      ("©️", true),
      ("♥", false),
      ("♥️", true),
      ("1", false),
      ("1️⃣", true),
      ("🇺🇸", true),
      ("👨‍👩‍👧‍👧", true),
      ("a", false),
  ]
```

Can you try with `Emoji_Presentation` for regex directly, which might be a bit conservative but will do that job in most cases? Let's keep it local to `svg_renderer` and not in `util`.

2. One more thing I noticed: `fontdb::Database::query` returns the best face from the first installed family in `families`. It does not keep walking the rest of `EMOJI_FONT_FAMILIES` looking for one that also supports `ch`.

That means if the first installed emoji family is already in `fonts`, or if it exists but does not have this glyph, we fall back to the default `usvg` resolver without ever trying the remaining emoji families.

I think this should iterate the families manually and check both:
- `!fonts.contains(&id)`
- `db.has_char(id, ch)`

smitbarmase (changes requested):

crates/util/src/util.rs
alanpjohn: Moved the test to `crates/gpui/src/svg_renderer.rs`

```bash
 cargo test -p gpui test_is_emoji_presentation
   Compiling gpui v0.2.2 (/home/alan/Documents/rust/zed/crates/gpui)
   Compiling gpui_linux v0.1.0 (/home/alan/Documents/rust/zed/crates/gpui_linux)
   Compiling gpui_platform v0.1.0 (/home/alan/Documents/rust/zed/crates/gpui_platform)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 8.27s
     Running unittests src/gpui.rs (target/debug/deps/gpui-3a3d3cdb14e35427)

running 1 test
test svg_renderer::tests::test_is_emoji_presentation ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 86 filtered out; finished in 0.00s

     Running tests/action_macros.rs (target/debug/deps/action_macros-89a1d21cd98d02a6)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

```

smitbarmase (approved):
Thanks! I added a has_char check and test for that.

Also, I updated the emoji presentation test with an expected shortcoming for certain emojis like 1️⃣ and ♥️, which are cluster-based rather than scalar. The fix for that will be in upstream, giving us the right API to work with for fallback rather than just `char`.

But this is fine for now, and the current form is a huge improvement anyways. Thanks again for working on this.

PR #51575, opened 2026-03-14
message_editor: Fix image file copied from external files paste inserting filepath text alongside image

monkey-mode:
What
---
Fix pasting image files copied from Finder in the Agent Panel on macOS. Previously, pasting an image file copied with `Cmd+C` in Finder would insert the file path as plain text instead of attaching the image as context.

Why
---
Two bugs combined to cause this:

1. **Wrong clipboard type priority in `pasteboard.rs` (fixed in #49367):** ~~The macOS pasteboard reader checked `public.utf8-plain-text` first. When Finder copies a file it places the filename as a string, the file path in `NSFilenamesPboardType`, and a TIFF of the file icon — all in the same clipboard event. Because strings were checked first, Zed read the filename string and ignored the image data entirely.~~

2. **Missing `cx.stop_propagation()` in `message_editor.rs`:** The `paste()` handler is registered as a `capture_action`. In GPUI's capture phase, `propagate_event` defaults to `true` — simply `return`ing does not stop propagation. Without an explicit `cx.stop_propagation()`, the inner `Editor`'s bubble-phase paste handler also fired, read `ExternalPaths` from the clipboard, converted it to text via `ClipboardItem::text()` (which returns the file path string), and inserted it alongside the image.

Fix
---
gpui_macos/src/pasteboard.rs`: Change clipboard read priority to **file paths → image data → string**. Added `read_external_paths()` which reads `NSFilenamesPboardType` and returns a `ClipboardEntry::ExternalPaths`. This lets the existing `paste_images_as_context` path load the actual file content.

- `agent_ui/src/message_editor.rs`: Add `cx.stop_propagation()` before `task.detach()` when the image paste task is accepted, preventing the inner editor from also handling the event.

Closes #51574

Test Plan
---

- [x] `cargo fmt --all -- --check`
- [x] `cargo build -p gpui_macos agent_ui` compiles clean
- [x] Manual verification of:
   - [x] Copy an image file in Finder (`Cmd+C`), paste into Agent Panel — image attaches correctly
   - [x] Copy image from browser ("Copy Image"), paste into Agent Panel — image attaches correctly  
   - [x] `Cmd+Shift+4` screenshot paste still works
   - [x] Regular text paste still works
   
Release Notes
- Fixed pasting image files copied from Finder inserting the file path instead of attaching the image in the Agent Panel

Screenshots
---
https://github.com/user-attachments/assets/edf6ba5a-6ff7-478c-a9ed-7cb5e889ccb3

image


monkey-mode:
This PR fixed the clipboard type priority bug (related: #49367), but there's a remaining issue: missing `cx.stop_propagation()` in the paste handler.

Two bugs combined to cause the original problem:

**2. Missing `cx.stop_propagation()` in `message_editor.rs`**
The `paste()` handler is registered as a `capture_action`. In GPUI's capture phase, `propagate_event` defaults to `true` — simply returning does not stop propagation. Without an explicit `cx.stop_propagation()`, the inner `Editor`'s bubble-phase paste handler also fired, read `ExternalPaths` from the clipboard, converted it to text via `ClipboardItem::text()` (which returns the file path string), and inserted it alongside the image.

Veykril (approved):
Thanks!

PR #33513, opened 2025-06-27
Fix blend alpha colors with editor background in inline preview

ddoemonn:
Closes #33505

## Before

Screenshot 2025-06-27 at 12 22 57

## After

Screenshot 2025-06-27 at 12 22 47

Release Notes:

- Fixed inline color previews not correctly blending alpha/transparency values with the editor background

PR #51580, opened 2026-03-15
editor: Add intelligent bracket colorization

kaikozlov:
Closes #50210

Bracket colorization currently runs theme accents through `ensure_minimum_contrast`, which can significantly alter authored palette colors in order to satisfy a single foreground-vs-background target. In practice, that makes rainbow brackets drift away from the theme colors they are supposed to reflect.

This PR keeps bracket colorization as a single automatic behavior, but changes how the palette is derived:

- preserve authored accent colors when they already work
- adjust only the specific accents that are too weak against the editor background
- only reorder the palette when adjacent bracket levels are still too similar after that

The implementation is staged because bracket colorization has to handle two separate problems:

- **Bracket-vs-Background Readability**: is an individual bracket visible against the editor background?
- **Adjacent Bracket Separation**: can you still distinguish one nesting level from the next?

Those two constraints are also why the implementation cannot simply defer to whatever accent colors the theme provides. Some themes, including the bundled One and Ayu, do not define explicit accents at all, so bracket colorization falls back to Zed's shared default accent palette. That palette was not designed with any particular theme's background in mind, which makes the background contrast check especially important for those cases.

That is why the previous `ensure_minimum_contrast(..., 55.0)` path is not enough on its own. It addresses foreground-vs-background readability, but if lightness-only adjustment cannot satisfy that target it escalates to saturation reduction and then black/white fallback.

This PR instead keeps the bracket-specific background fix bounded and lightness-only in OKLCH, so the resulting color stays much closer to the authored accent.

### How it works

- **Background Contrast Correction**: accents that fall below the APCA floor against the editor background are adjusted by changing OKLCH lightness only, preserving hue and chroma
- **Adjacency Correction**: if the resulting palette is still too weak between neighboring nesting levels, the palette is reordered conservatively to improve adjacent separation

That ordering matters. Doing the reorder first can easily be undone by the background-fix step.

### Color measurement

- background contrast uses APCA, via the existing `apca_contrast` path
- adjacent bracket separation uses OKLab Euclidean distance

APCA is the right tool for bracket-vs-background readability, but it is not designed to compare two foreground colors against each other. OKLab distance is a better fit for detecting when neighboring bracket levels are perceptually too close.

### Thresholds

- background APCA floor: `30` on dark themes, `35` on light themes
- adjacent OKLab target: `0.08` on dark themes, `0.10` on light themes
- light-theme reorder tolerance band: reorder only triggers below `0.095`, avoiding threshold-artifact reordering on palettes that are visually fine but narrowly miss `0.10`

### Validation

- `cargo test -p editor bracket_colorization`
- manual comparison on bundled One, Ayu, and Gruvbox theme families

Release Notes:

- Improved bracket colorization by preserving theme accent colors when possible and applying targeted contrast fixes only when needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5740,6 +5740,7 @@ dependencies = [
  "menu",
  "multi_buffer",
  "ordered-float 2.10.1",
+ "palette",
  "parking_lot",
  "pretty_assertions",
  "project",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5740,7 +5740,6 @@ dependencies = [
  "menu",
  "multi_buffer",
  "ordered-float 2.10.1",
- "palette",
  "parking_lot",
  "pretty_assertions",
  "project",
--- a/crates/editor/Cargo.toml
+++ b/crates/editor/Cargo.toml
@@ -63,7 +63,6 @@ markdown.workspace = true
 menu.workspace = true
 multi_buffer.workspace = true
 ordered-float.workspace = true
-palette.workspace = true
 parking_lot.workspace = true
 pretty_assertions.workspace = true
 project.workspace = true
--- a/crates/editor/Cargo.toml
+++ b/crates/editor/Cargo.toml
@@ -63,6 +63,7 @@ markdown.workspace = true
 menu.workspace = true
 multi_buffer.workspace = true
 ordered-float.workspace = true
+palette.workspace = true
 parking_lot.workspace = true
 pretty_assertions.workspace = true
 project.workspace = true
--- a/crates/editor/src/bracket_colorization.rs
+++ b/crates/editor/src/bracket_colorization.rs
@@ -8,15 +8,11 @@ use std::sync::Arc;
 
 use crate::{Editor, HighlightKey};
 use collections::{HashMap, HashSet};
-use gpui::{AppContext as _, Context, HighlightStyle, Hsla, Rgba};
+use gpui::{AppContext as _, Context, HighlightStyle, Hsla};
 use language::{BufferRow, BufferSnapshot, language_settings::LanguageSettings};
 use multi_buffer::{Anchor, BufferOffset, ExcerptRange, MultiBufferSnapshot};
-use palette::{
-    FromColor, Oklab, Oklch,
-    rgb::{LinSrgba, Srgba},
-};
 use text::OffsetRangeExt as _;
-use theme::{ActiveTheme, Appearance};
+use theme::{Appearance, Oklab, Oklch, hsla_to_oklab, hsla_to_oklch, oklch_to_hsla};
 use ui::utils::apca_contrast;
 
 impl Editor {
@@ -29,18 +25,10 @@ impl Editor {
             self.bracket_fetched_tree_sitter_chunks.clear();
         }
 
-        let accents = self
-            .accent_data
-            .as_ref()
-            .map(|accent_data| accent_data.colors.0.clone())
-            .unwrap_or_else(|| {
-                let theme = cx.theme();
-                bracket_colorization_accents(
-                    &theme.accents().0,
-                    theme.appearance,
-                    theme.colors().editor_background,
-                )
-            });
+        let Some(accent_data) = self.accent_data.as_ref() else {
+            return;
+        };
+        let accents = accent_data.colors.0.clone();
         let multi_buffer_snapshot = self.buffer().read(cx).snapshot(cx);
 
         let visible_excerpts = self.visible_buffer_ranges(cx);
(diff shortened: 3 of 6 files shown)


SomeoneToIgnore (changes requested):
https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#sending-changes

We would need some screenshots with before-after comparison, at least for both standard themes.

SomeoneToIgnore (commented):
Thank you, this is a very good set of themes to test.
Could you try and demonstrate it better?
E.g. a sequence of brackets (given that we change the rainbow brackets feature) to evaluate the contrast between them.

For example, here things have degraded

image

which does not seem what we want.

kaikozlov:
Yes, I can do that. Here are some better demo pictures:

## One Dark
Screenshot 2026-03-15 at 2 30 08 PM

## One Light
Screenshot 2026-03-15 at 2 30 13 PM

## Ayu Dark
Screenshot 2026-03-15 at 2 30 21 PM

## Ayu Light
Screenshot 2026-03-15 at 2 30 28 PM

## Gruvbox Dark
Screenshot 2026-03-15 at 2 30 38 PM

## Gruvbox Light
Screenshot 2026-03-15 at 2 30 45 PM

SomeoneToIgnore (commented):
Yeah, to me some versions look worse, some better — seems that this change alone is not enough and we need to change themes?
At least, does not seem right to degrade Light themes just because.

Maybe @danilo-leal has any ideas on this all.

SomeoneToIgnore (commented):
Frankly, the amount of new code added seems a bit more than I've expected for a situational fix for the rainbow brackets.
A somewhat related issue with the CSS color display was fixed way more elegant with https://github.com/zed-industries/zed/pull/33513 , for example.

Something I'd want to avoid for sure is the extra configuration knob: we should find a way to display those accents well enough in all cases, even if it comes to tweaking the colors?

Yet I definitely like the way right (new) color set looks like, thank you a lot for putting up the screenshots of the result!
So, at least we have some sort of a goal (the last image and a few of the bright themes are a bit dim, though?) which we can try to achieve the different way, if it's possible?

kaikozlov:
Thanks for the feedback. I reworked this pretty substantially in response to the footprint concern.

The previous version was around +1440 / -104. The current branch is down to +487 / -92 across 4 files.

Compared to the earlier version, I removed:

- the configuration knob and all related settings/schema/UI/import plumbing
- duplicate palette-acquisition paths
- tests that were overlapping rather than adding distinct coverage

What remains is the automatic bracket-palette derivation itself and the tests around it. The updated PR description walks through the reasoning in more detail, but the short version is that the remaining logic is staged because bracket colorization still has to satisfy two separate constraints at once:

- bracket colors need to stay readable against the editor background
- adjacent nesting levels still need to remain distinguishable from each other

Those are not the same problem and do not respond to the same treatment, which is why the implementation rescues only the accents that fail the background check first, then reorders only if the adjacent separation is still too weak after that.

This also still needs to behave correctly for themes that rely on Zed's shared default accent palette rather than authored theme accents, including the bundled One and Ayu families.

At this point, I think this is much closer to the minimum implementation required to produce the results shown in the screenshots. I'd appreciate a fresh look at the current version.

SomeoneToIgnore:
Sorry for the long wait: I was taking a long hiatus and _maybe_ next month is when I'll be able to review this.

SomeoneToIgnore (approved):
Thank you a lot for the patience and the rebases!

Pushed a few fixes:

* removed potential panic spots
* re-adjusted the constants to be symmetric, as it looked odd otherwise
* removed the extra dependency and exposed some public API from `theme` crate instead

I will merge the PR after the next release on Wednesday, would appreciate if you double-check that I did not spoil things with my constant fiddling.
Based on my visual assertions, the updated colors look mostly good, my only concern are the colors in OneDark that are somewhat close to each other:

(left — current; right — the branch, brackets #4 and #6 look almost identical to me, unlike the original colors, yet seems tolerable due to another color breaking the sequence)
Image

kaikozlov:
Hi! Thanks for the fixes and for working with me to get this right!

I got a chance to look at your changes this morning, and from what I can see, everything looks good. Nothing is spoiled 😅

On the close pair in One Dark, the behavior we're seeing is currently by design. The adjacency correction only reorders when immediately adjacent nesting levels are too close, resulting in the high-contrast orange bracket separating its neighboring blue and purple brackets. I landed on this behavior during implementation, thinking the main goal is to keep boundaries legible.

Also worth noting: One and Ayu don't define their own accent palettes, so the fallback is Zed's shared default palette being adjusted against each theme's background. If that ever needs tightening, the cleaner approach would be to add theme-specific accent palettes for the bundled themes, but that's probably out of scope here.

Excited to see this land!

PR #51592, opened 2026-03-15
workspace: Handle double-click on pinned tab row empty space

mgabor3141:
When `tab_bar.show_pinned_tabs_in_separate_row` is enabled, double-clicking the empty space on the unpinned tab row creates a new tab, but double-clicking the empty space on the pinned tab row does nothing.

Add the same `on_click` double-click handler to the pinned tab bar drop target so both rows behave consistently.

Release Notes:

- Fixed double-clicking empty space in the pinned tab row not opening a new tab when `show_pinned_tabs_in_separate_row` is enabled.


JosephTLyons (approved):

mgabor3141:
> I went ahead and added in a test so we could merge this in. Thanks @mgabor3141 

Thanks. I've been busy with other things so I couldn't get to this one. Thank you for taking it over the finish line.

PR #51593, opened 2026-03-15
Fix FormatSelections to only format selected ranges, not entire document

prertik:
When `editor: format selections` get invoked, the Prettier and external formatter branches in `format_buffer_locally` ignored the selection ranges entirely, causing the whole document to be formatted.

- Thread selection ranges as UTF-16 offsets through to Prettier via `rangeStart/rangeEnd` options in the format request.
- Skip external formatters when ranges are present, since they have no mechanism for range formatting.
-  Create diff edits and apply them for JSON-like languages. For single-expression languages like JSON, it wasn't respecting the range commands from Prettier. So, filter the diff edits returned by Prettier to retain only those overlapping with the user's selection byte ranges, ensuring changes outside the selection are never applied.

Part of #25796 

Release Notes:

- Fixed FormatSelections to only format selected ranges, not the entire document where prettier is supported.

Current Behaviour:

[original behaviour.webm](https://github.com/user-attachments/assets/d5f0cb48-4c3f-44aa-89a9-975f31fce92d)

New Behaviour:

[new behaviour.webm](https://github.com/user-attachments/assets/41e04b90-f37f-43e1-b8ed-4622684454b1)

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -74,8 +74,7 @@ use language::{
     CodeLabelExt, Diagnostic, DiagnosticEntry, DiagnosticSet, DiagnosticSourceKind, Diff,
     File as _, Language, LanguageName, LanguageRegistry, LocalFile, LspAdapter, LspAdapterDelegate,
     LspInstaller, ManifestDelegate, ManifestName, ModelineSettings, OffsetUtf16, Patch, PointUtf16,
-    TextBufferSnapshot, ToOffset, ToOffsetUtf16,
-    ToPointUtf16, Toolchain, Transaction, Unclipped,
+    TextBufferSnapshot, ToOffset, ToOffsetUtf16, ToPointUtf16, Toolchain, Transaction, Unclipped,
     language_settings::{
         AllLanguageSettings, FormatOnSave, Formatter, LanguageSettings, all_language_settings,
     },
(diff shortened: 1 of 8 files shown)


crates/project/src/lsp_store.rs
```diff
@@ -1731,6 +1774,12 @@ impl LocalLspStore {
                 }
                 Formatter::External { command, arguments } => {
                     let logger = zlog::scoped!(logger => "command");
+
+                    if buffer.ranges.is_some() {
+                        zlog::trace!(logger => "External formatter does not support range formatting; skipping");
```
```suggestion
                        zlog::debug!(logger => "External formatter does not support range formatting; skipping");
```

crates/project/src/lsp_store.rs
```diff
@@ -1709,17 +1709,60 @@ impl LocalLspStore {
...
+                        cx,
+                    )
+                    .await
+                    .transpose()?;
+                    let Some(mut diff) = diff else {
                         zlog::trace!(logger => "No changes");
                         continue;
                     };
 
+                    if let Some(byte_ranges) = byte_ranges {
```
SomeoneToIgnore: Ideally, such logic changes (same for the one below and the main one) are great to test, as otherwise the may degrade in the future.

The the very least, we need to test now that editor's range formatting applies it correctly for Prettier.
We test parsers already with
https://github.com/zed-industries/zed/blob/6b91dfbd04f60fee298e170082755082dcb295a5/crates/prettier/src/prettier.rs#L520-L522
so I guess it's another mock output line we can add here, and then write another test like https://github.com/zed-industries/zed/blob/6b91dfbd04f60fee298e170082755082dcb295a5/crates/editor/src/editor_tests.rs#L21064

2 other cases, hopefully, can be tested quite similarly, with the same code that initializes the editor, selects a range in a snippet and tries to format it.
We could add all new tests into a prettier module if, it helps, and extract the initialization code there more specifically for such kinds of tests.

crates/project/src/lsp_store.rs
```diff
@@ -1709,17 +1709,60 @@ impl LocalLspStore {
...
+                        diff.edits.retain(|(edit_range, _)| {
```
SomeoneToIgnore: What is the reason to restrict Prettier output to the selection byte ranges only?
Zed know almost nothing about languages and _has to_ trust the formatter server to accept my range and output whatever appropriate set of edits.

As a guessing example, I could expect my selection (`«»`) that touches the unindented struct to get formatted?

```rs
«struct Foo {
  normal: u32,»
    bad_1: u32,
       bad_2 u32,
}
```

crates/project/src/prettier_store.rs
```diff
@@ -736,6 +736,7 @@ pub fn prettier_plugins_for_language(
 pub(super) async fn format_with_prettier(
     prettier_store: &WeakEntity<PrettierStore>,
     buffer: &Entity<Buffer>,
+    range_utf16: Option<Range<usize>>,
```
SomeoneToIgnore: > usize

It's quite important to preserve the specific coordinate knowledge, let's use `OffsetUtf16` here and around instead.

SomeoneToIgnore (commented):
Nice to see this working well.

One curious thing to test manually (as due to the way Prettier is mocked now we cannot test the real formatting, alas) is the multi-char treatment, e.g. what if there's a bunch of √ before and after and within the selection for Prettier, does Zed behave normally?

crates/project/src/lsp_store.rs
prertik: Will add and iterate, thanks @SomeoneToIgnore!

crates/project/src/lsp_store.rs
prertik: I'll add tests for this. Thanks, @SomeoneToIgnore

crates/project/src/lsp_store.rs
prertik: Yeah, so, when we apply formatting to a selection, we just want it to format that selection only rather than the whole file. That was the issue we were trying to solve in this PR. We have two ways to format a file: a complete format buffer (or file) and format selection on every editor.

Say, if you provide only a selection in which the below code is obviously not formatted properly:
```
«struct Foo {
  normal: u32,»
    bad_1: u32,
       bad_2 u32,
}
```

Then, it should format only the lines that you selected for formatting and not touch other lines, even if they are badly formatted elsewhere; otherwise, it loses the point of only selective formatting.

LSP formatting does it the eager way, like, say, if we set config to format only edited lines for LSP based server.
Then, I might be wrong, but LSP knows the structure and hence can accurately format only the areas. What I mean by this is, in the same example:

```
 struct Foo {
  normal: u32,
    bad_1: u32,
       bad_2 u32,
}
```

Suppose I just change this:

```
 struct Foo {
  nmal: u32,
    bad_1: u32,
       bad_2 u32,
}  
```
If I have LSP format only edited lines (set on some config), then LSP understands I made some change on the struct, but the struct isn't properly formatted and makes the change on the whole struct but not on the whole file. If some other struct in the file is badly formatted, it leaves them. (This is what my understanding is, based on LSP usage; please correct me if I'm wrong).

Such kind of logic, we cannot have on basic prettier formatting because we would have to either have the complex LSP logic extracted (which might not work). Otherwise, we have to resort to this kind of solution, where we only do selective formatting using the Prettier server. Only format a selection that the user has specifically asked for.

crates/project/src/prettier_store.rs
prertik: Will iterate on this.

crates/project/src/lsp_store.rs
SomeoneToIgnore: Yeah, I see, makes sense then if tested, maybe a comment to note this particular intent (to format only what's selected) is good to have too.

Thank you for the explanation.

SomeoneToIgnore (approved):
Great, thank you so much for the fix, this should solve most of the format on select -related issues it seems.

Congratulations with your first contribution!

PR #20995, opened 2024-11-21
Fix keybindings on a Spanish ISO keyboard

ConradIrwin:
Co-Authored-By: Peter <peter@zed.dev>

Also reformatted the mappings to be easier to read/edit by hand.

Release Notes:

- Fixed keyboard shortcuts on Spanish ISO keyboards

PR #51600, opened 2026-03-15
gpui_macos: Fix cmd+backtick window cycling on Spanish keyboard layouts

jamarju:
Follows up on #51394.

## Problem

On Spanish keyboard layouts, cmd+backtick does not cycle windows because Zed's key_equivalents system remaps `cmd-'` (ToggleSelectedDiffHunks) onto the backtick key, consuming the event before macOS can handle window cycling.

## How key_equivalents remapping works

Zed replicates macOS's menu shortcut localization: when a US-defined shortcut like `cmd-'` can't be typed on another layout, macOS relocates it to a reachable key. These relocations form chains that terminate at layout-specific characters (like ñ, ¡, ç, ...) that have no US shortcut to displace.

On Spanish-ISO, the chain containing backtick is:

```
" -> ^ -> & -> / -> ' -> ` -> < -> ; -> ¡ (open exclamation)
```

The `' -> backtick` entry causes `cmd-'` (ToggleSelectedDiffHunks) to listen on the backtick key, shadowing macOS window cycling:

https://github.com/zed-industries/zed/blob/66d7452b78494699909e876d505222497e288a0e/crates/gpui_macos/src/keyboard.rs#L1225

## Fix

Bypass backtick in the chain by connecting its predecessor directly to its successor:

```
Before: ... -> ' -> ` -> < -> ; -> ¡
After:  ... -> ' -> < -> ; -> ¡       (backtick removed from chain)
```

This frees the backtick key for macOS window cycling.

The side effect is that ToggleSelectedDiffHunks moves from backtick to `<`.

This PR also removes a bogus `} -> backtick` entry in Spanish-ISO that looked like a transcription error from #20995. The original [keyboard-inspector](https://github.com/ConradIrwin/keyboard-inspector) data reported "not found" for `}`, which was incorrectly converted to a backtick mapping. Nothing maps TO `}` in the table, so removing it has no chain side effects.

## Scope

This PR only targets the two Spanish layouts (Spanish and Spanish-ISO). The same backtick collision affects other layouts via different US keys (`"`, `=`, `\`), but those are left for follow-up since they require testing with those physical keyboards.

| US key | Zed binding                              | Layouts affected (count) |
|--------|------------------------------------------|--------------------------|
| '      | cmd-' editor::ToggleSelectedDiffHunks    | Spanish, Spanish-ISO (2) |
| "      | cmd-" editor::ExpandAllDiffHunks         | German, Austrian, Brazilian, Portuguese, Swiss... (8) |
| =      | cmd-= zed::IncreaseBufferFontSize        | Danish, Finnish, Norwegian, Swedish... (13) |
| \      | cmd-\ pane::SplitRight                   | French, Belgian, ABC-AZERTY (4) |

Release Notes:

- Fixed cmd+backtick window cycling not working on Spanish keyboard layouts.

### DIFF
--- a/crates/gpui_macos/src/gpui_macos.rs
+++ b/crates/gpui_macos/src/gpui_macos.rs
@@ -28,6 +28,7 @@ mod text_system;
 mod platform;
 mod window;
 mod window_appearance;
+mod window_cycle;
 
 use cocoa::{
     base::{id, nil},
@@ -46,6 +47,7 @@ pub(crate) use display_link::*;
 pub(crate) use keyboard::*;
 pub(crate) use platform::*;
 pub(crate) use window::*;
+pub(crate) use window_cycle::*;
 
 #[cfg(feature = "font-kit")]
 pub(crate) use text_system::*;
--- a/crates/gpui_macos/src/keyboard.rs
+++ b/crates/gpui_macos/src/keyboard.rs
@@ -1235,28 +1235,28 @@ fn get_key_equivalents(layout_id: &str) -> Option<HashMap<char, char>> {
             ('~', '>'),
         ],
         "com.apple.keylayout.Spanish-ISO" => &[
-            ('"', '¨'),
+            ('"', '^'),
             ('#', '·'),
             ('&', '/'),
+            ('\'', '`'),
             ('(', ')'),
             (')', '='),
             ('*', '('),
-            ('.', 'ç'),
-            ('/', '.'),
-            (':', 'º'),
-            (';', '´'),
-            ('<', '¿'),
-            ('>', 'Ç'),
+            ('/', '\''),
+            (':', '¿'),
+            (';', '¡'),
+            ('<', ';'),
+            ('=', '*'),
+            ('>', ':'),
             ('@', '"'),
             ('[', 'ñ'),
-            ('\'', '`'),
-            ('\\', '\''),
-            (']', ';'),
+            ('\\', 'ç'),
+            (']', '´'),
             ('^', '&'),
             ('`', '<'),
             ('{', 'Ñ'),
-            ('|', '"'),
-            ('}', '`'),
+            ('|', 'Ç'),
+            ('}', '¨'),
             ('~', '>'),
         ],
         "com.apple.keylayout.Swedish" => &[
@@ -1499,4 +1499,41 @@ fn get_key_equivalents(layout_id: &str) -> Option<HashMap<char, char>> {
     };
 
     Some(HashMap::from_iter(mappings.iter().cloned()))
+}
+
+#[cfg(test)]
+mod tests {
+    use gpui::Modifiers;
+
+    use super::*;
+
+    #[test]
+    fn spanish_iso_maps_apostrophe_to_backtick() {
+        let mapper = MacKeyboardMapper::new("com.apple.keylayout.Spanish-ISO");
+        let mapped = mapper.map_key_equivalent(
+            Keystroke {
+                modifiers: Modifiers::none(),
+                key: "'".into(),
+                key_char: None,
+            },
+            true,
+        );
+
+        assert_eq!(mapped.inner().key, "`");
+    }
+
+    #[test]
+    fn spanish_iso_mapping_still_applies_with_command_pressed() {
+        let mapper = MacKeyboardMapper::new("com.apple.keylayout.Spanish-ISO");
+        let mapped = mapper.map_key_equivalent(
+            Keystroke {
+                modifiers: Modifiers::command(),
+                key: "'".into(),
+                key_char: None,
+            },
+            true,
+        );
+
+        assert_eq!(mapped.inner().key, "`");
+    }
 }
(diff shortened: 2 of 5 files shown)


jamarju:
> What happens if you "reset default shortcuts" on the Spanish locale?

It resets to cmd+backtick. I also tried right after coming from a US config in case macOS decides to remap cmd+` to cmd+< when switching to ES:

1. Set primary language to English, region US, layout US
2. "reset default shortcuts" -> focus to next window is cmd+` at this point
3. Set primary language to Spanish, region Spain, layout ES
4. still cmd+`
5. "reset default shortcuts" -> still cmd+`

So I was not able to reproduce your observation that cmd+< is the Spanish shortcut for "move focus to next window". As far as I can remember this shortcut has always been cmd+` in Spanish Macbooks at least since ~2010.

> when we tested a Portuguese laptop, the default was cmd-´ (not cmd-`).

I was able to reproduce this: by switching locale from Spanish to Portuguese, the shortcut was still cmd-` but after I pressed "reset default shortcuts", then the shortcut became cmd-´.

> I think we'll need to figure out how to query for this shortcut

I went this route instead of hardcoding cmd+`. Take a look at the latest changes. The fix is now a bit more involved, as it:

- queries the active macOS window-cycle shortcut directly and prevents Zed from handling it.
- watches that plist for changes while Zed is running, so if the user changes the shortcut later we pick it up without restart.

I’ve tested the new behavior locally and the reserved shortcut updates correctly when changed at runtime.

Looking forward to your feedback on this one.

ConradIrwin:
Thanks for taking the time to explore this. I like the idea of reading it, but I notice I don't have `"27"` in that plist at all...

If you are sure that Spanish users will always have cmd-` as window switching, then ensuring we exclude that from mappings might be the right trade off for now, and we can deal with any further complaints that come in.

jamarju:
> but I notice I don't have "27" in that plist at all...

That makes sense as I think that plist is an override file. English and Spain-Spanish templates ship empty for entry 27, so users in those locales fall through to the SkyLight default (cmd+backtick). Latin American Spanish (es_419) and  Portuguese (pt_PT) templates ship overrides (cmd+forward tick).

If you still want to go the "read macOS config + observe" (this PR's) way:

I have recently committed a new change, now `CGSGetSymbolicHotKeyValue` is used instead of reading the plist directly. This is a private API but seems stable and widely used and does the override lookup, else falls back to cmd+backtick. So the right value comes back per-locale without us hardcoding anything or parsing the plist.

I haven't found an API to subscribe to notifications of changes to this shortcut, though, so we're stuck with the file watcher for that...

> then ensuring we exclude that from mappings might be the right trade off for now, 

That's probably simpler than this PR, however I'm not sure how you want to do this. I can see a few options:

1. Unbind the problematic shortcuts (for all users or only those with non US layouts)
2. Rebind them to something else
3. Keep the bindings as they are but catch cmd+backtick and let OS handle
4. ...?

All have shortcomings IMO:

1 and 2: let's define "problematic shortcuts" here as those that are bound to a Zed action and get remapped to the default "focus next window" shortcut in non-US layouts. The "focus next window" itself depends on the locale (cmd+backtick generally, cmd+´ on some regions). For example:

*  German layout, German locale (macOS default = cmd+´)
    - German layout maps US ' -> ´
    - Zed's cmd-' (ToggleSelectedDiffHunks) relocates to that key
    - Pressing cmd+(that key) maps to Zed's cmd-' action -> ToggleSelectedDiffHunks shadows the cycler

*  German layout, US locale (macOS default = cmd+`)
    - German layout maps US " -> backtick
    - Zed's cmd-" (ExpandAllDiffHunks) relocates to that key
    - Pressing cmd+(that key) maps to Zed's cmd-" -> ExpandAllDiffHunks shadows the cycler

3: cmd+backtick is the default for many regions, but it is not for many others.

Honestly, this is a much deeper rabbit hole than I anticipated when I raised the PR 😅

jamarju:
Update: I just simplified this further by removing the file watcher.

I benchmarked CGSGetSymbolicHotKeyValue and it's ~10µs per call. At realistic typing rates this is insignificant, so I just call it on every matching attempt instead of caching the value and watching the plist for changes. 

This results in much simpler code and a general solution that works regardless of the locale, layout, or even customization of the shortcut.

ConradIrwin:
Nice find on that API!

So to clarify this branch does two things:
* Tweaks the default Spanish mapping
* Ensures that we never override the window cycling shortcut.

I get that this shortcut could (in theory) change at any time, and that 10µs is only 0.1% of our 8ms time budget, but should we just cache it when Zed opens, given that it doesn't change except when you force reconfigure it?

Otherwise, this is looking good to me – I'm so excited to have figured out a way to solve this. Thank you!

jamarju:
> So to clarify this branch does two things:
> - Tweaks the default Spanish mapping
> - Ensures that we never override the window cycling shortcut.

That's right.

> but should we just cache it when Zed opens, given that it doesn't change except when you force reconfigure it?

I wasn't aware there was an 8 ms budget, good to know! 10 µs is negligible over 8 ms, but I do understand your concern that we are introducing a time dependency on an API that we have no control of, may change in the future or have unusual code paths that take longer, so in a way I agree with caching it at the beginning. But that has the drawback that if the user changes the shortcut it needs to restart Zed to pick the new one up... unless we watch the plist. I can bring back the plist watcher but IMO it was super kludgey.

If you ask me I'd keep the PR as it is, trust the private API is going to work and take 10µs for years to come, and deal with it if / when it changes. But that's ultimately your call... I can keep iterating if you don't feel 100% happy with the change.

10µs was measured on a 2021 M1 Macbook, btw! Should be even faster on more modern hardware...

PR #51601, opened 2026-03-15
git_ui: Add starts_open setting to git panel

### LINKED ISSUE #51542: Add starts_open setting to git_panel (and other dock panels) for parity with project_panel
## Problem

When opening a new directory that hasn't been visited before (e.g., via `zed --reuse <dir>` from a terminal multiplexer), the right dock closes even if the user had it open with the git panel. The left dock stays open because `project_panel` has a `starts_open` setting (defaulting to `true`), but no other panel has this option.

This is particularly noticeable in workflows where an external tool switches Zed's workspace frequently (e.g., terminal multiplexers like cmux that sync the editor to the active terminal directory). Every directory switch closes the right panel.

## Proposed Solutions

I see two possible approaches and would be happy with either:

### Option A: Add `starts_open` to panels that are missing it (additive, no behavior change)

Add the same `starts_open` setting that `project_panel` already has to other panels — starting with `git_panel`. The infrastructure already exists via the `Panel` trait's `starts_open()` method and the dock's `add_panel()` logic.

This would let users configure:
```json
{
  "project_panel": { "starts_open": true },
  "git_panel": { "dock": "right", "starts_open": true }
}
```

Changes required per panel (~15 lines across 4 files):
- `settings_content.rs`: Add `starts_open: Option<bool>` to the panel's settings content struct
- Panel settings file: Add `starts_open: bool` field + deserialization
- Panel impl: Add `fn starts_open()` to the `Panel` impl
- `default.json`: Add `"starts_open": false` default

### Option B: Restore default dock state for all new workspaces (behavior change, smaller diff)

Currently, when no per-workspace serialized dock state exists, the "default dock state" (saved from the user's last session) is only restored for empty workspaces (`is_empty_workspace && !serialized_workspace_has_paths` in `workspace.rs`). Removing the `is_empty_workspace` guard would carry forward the user's dock layout to any new workspace — a 3-line change.

This approach is arguably more correct ("remember my layout") but changes default behavior for everyone.

## Context

- `project_panel` already has `starts_open` with a settings-backed implementation
- The `Panel` trait defines `starts_open()` with a default of `false` — panels just need to override it
- `git_panel`, `agent` panel, `notification_panel`, `terminal_panel`, and others all use the default (`false`) with no way to change it
- Issue #46589 is related (project_panel's `starts_open` not being respected)

Happy to discuss which approach fits better, and can submit a PR if invited to do so. Would rather contribute upstream than maintain a fork.

mvanhorn:
Fixes #51542

Adds a `git_panel.starts_open` setting for parity with `project_panel`. When set to `true`, the git panel opens automatically in new workspaces that don't have saved panel state.

The default is `false`, matching current behavior (non-breaking).

### Usage

```json
{
  "git_panel": {
    "starts_open": true
  }
}
```

### Implementation

Follows the same pattern as `project_panel.starts_open`:

1. `GitPanelSettingsContent` - added `starts_open: Option<bool>` to the settings schema
2. `GitPanelSettings` - added `starts_open: bool` field
3. `GitPanel` Panel impl - overrides `starts_open()` to read from settings

Release Notes:

- Added `git_panel.starts_open` setting to control whether the git panel opens automatically in new workspaces (default: false)

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -922,6 +922,10 @@
     ///
     /// Default: false
     "tree_view": false,
+    // Whether the git panel should open on startup.
+    //
+    // Default: false
+    "starts_open": false,
     // Whether to show a badge on the git panel icon with the count of uncommitted changes.
     //
     // Default: false
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -5824,6 +5824,10 @@ impl Panel for GitPanel {
         Box::new(ToggleFocus)
     }
 
+    fn starts_open(&self, _: &Window, cx: &App) -> bool {
+        GitPanelSettings::get_global(cx).starts_open
+    }
+
     fn activation_priority(&self) -> u32 {
         2
     }
--- a/crates/git_ui/src/git_panel_settings.rs
+++ b/crates/git_ui/src/git_panel_settings.rs
@@ -29,6 +29,7 @@ pub struct GitPanelSettings {
     pub tree_view: bool,
     pub diff_stats: bool,
     pub show_count_badge: bool,
+    pub starts_open: bool,
 }
 
 impl ScrollbarVisibility for GitPanelSettings {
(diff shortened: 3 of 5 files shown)


SomeoneToIgnore (commented):
Before merging this needs a rebase and a change in all-settings.md , otherwise looks good.

SomeoneToIgnore (approved):
Thank you!

SomeoneToIgnore (commented):
Oh, there's no new entry in `default.json` (would be good to have the comment on it, similar as all other entries have there) which causes the tests to fail.

SomeoneToIgnore (approved):

PR #51605, opened 2026-03-15
project: Fix code_actions_on_format edits being reverted when no formatter available

mvanhorn:
Fixes #51490

## Problem

When `code_actions_on_format` is configured (e.g., `source.fixAll.phpcs`) and `formatter` is not explicitly set, code action edits are applied but then reverted during format-on-save.

The root cause is in the formatting pipeline's error handling. Code actions and formatters share a single transaction via `.chain()`. When `Formatter::Auto` resolves to `Formatter::Prettier` (PHP defaults to `prettier.allowed: true`), and prettier is not installed or fails, the error propagates via `?` out of the formatter loop, aborting the entire `format_buffer_locally` function. This causes the formatting transaction - which already contains the successfully applied code action edits - to be lost.

The workaround of setting `"formatter": []` works because it removes all formatters from the chain, so no formatter can fail and trigger the abort.

## Fix

Change error handling for individual formatters (Prettier, External command, Language Server) from propagating errors via `?` to logging the error and continuing to the next formatter. This treats "formatter failed" as a no-op for that specific formatter rather than aborting the entire pipeline.

This matches the existing pattern used when `LanguageServer(Current)` finds no server that supports formatting - it already logs and `continue`s rather than erroring.

## Testing

- `cargo check -p project` passes
- `cargo fmt -p project -- --check` passes

Release Notes:

- Fixed `code_actions_on_format` edits being reverted when `formatter: "auto"` resolves to an unavailable formatter.


mvanhorn:
@cla-bot check

SomeoneToIgnore (changes requested):
That's a great investigation under the link posted.

It seems that the main idea is that the loop `for formatter in formatters` must be always carried up to the end, and here we try to fix some of the `?` that forbid it.

Yet, things like `extend_formatting_transaction` are still `?`-marked and would need a similar treatment?

Overall, the whole approach seems not very sustainable, we would be better off extracting the entire loop's body into its own function/method.
We could return an `anyhow::Result<std::ops::ControlFlow>` there and cover all various `?` explicitly, once, by logging an error with the context and moving on.

-------------

Besides that, the usual ask are the tests: formatting is rather intricate and would definitely benefit from covering this scenario in e.g. `editor_tests.rs`

SomeoneToIgnore (approved):
Thank you!

PR #51608, opened 2026-03-15
Add terminal.scroll_on_output setting

### LINKED ISSUE #51585: Add terminal.scroll_on_output setting to control auto-scroll behavior
Hey team — long-time fan, first-time issue filer. Switched to Zed from VS Code a while back and honestly can't imagine going back. The performance alone sold me, but the restraint in the design is what keeps me here. You all clearly care about getting the details right, which is why I think this one's worth flagging.

## What's happening

I run Claude Code in the integrated terminal pretty much all day. When it streams back a response, the terminal auto-scrolls to the bottom as output comes in. So every time I get a response, I'm staring at the last line and have to scroll back up to find where it actually started. Multiply that by a few dozen times per session and it adds up fast.

It's honestly the only thing that breaks the flow in what's otherwise the smoothest dev environment I've used.

## The ask

A `terminal.scroll_on_output` boolean (default `true` so nothing changes for anyone). When set to `false`, the terminal just holds your scroll position while new output writes below. You scroll down when you're ready.

```json
{
  "terminal": {
    "scroll_on_output": true
  }
}
```

Small surface area, non-breaking, fits right into the existing terminal settings block.

## Why it matters more now than it used to

This probably wasn't a big deal when terminal output was mostly build logs and test runners. But with Claude Code, Aider, and other AI tools becoming a core part of the workflow, the terminal is now producing long, structured text that you actually need to *read* top-down. The auto-scroll fights that pattern hard.

Zed is already positioning itself as the best editor for AI-assisted dev — this would remove the one rough edge in that experience.

## Prior art

Most terminals have had this for a while:

- **iTerm2**: "Scroll to bottom on output" toggle (off by default)
- **Windows Terminal**: `"snapOnOutput": false`
- **GNOME Terminal**: "Scroll on output" checkbox

Would love to see Zed's clean take on it.

---

Keep up the great work. Genuinely one of the best pieces of software I use daily.

zerone0x:
## Summary
Adds a `terminal.scroll_on_output` boolean setting (default: `true`) to control whether the terminal auto-scrolls when new output arrives.

When set to `false`, the terminal holds the current scroll position while new output writes below, allowing users to scroll down when ready. This is particularly useful for AI coding tools (Claude Code, Aider, etc.) that produce long structured output in the terminal.

### Implementation
- Captures scroll position before `sync()` processes terminal events
- When `scroll_on_output` is `false` and the user was scrolled up, restores the previous offset after alacritty resets it

Fixes #51585

## Release Notes

- Added `terminal.scroll_on_output` setting to control auto-scroll behavior when new terminal output arrives


zerone0x:
'@cla-bot check

PR #51615, opened 2026-03-15
pane: Add "Reveal in Finder" to tab context menu

mvanhorn:
The tab context menu has "Copy Path", "Open in Terminal", and "Reveal In Project Panel" but no way to reveal the file in the system file manager. This action already exists in three other context menus (editor right-click, project panel, outline panel) but was missing from tab right-click.

## Changes

Adds a platform-specific entry to the tab context menu:
- **macOS:** "Reveal in Finder"
- **Windows:** "Reveal in File Explorer"
- **Linux:** "Reveal in File Manager"

Placed after "Copy Relative Path" and before "Pin Tab". Gated behind `is_local` (including WSL with host interop) to match the project panel's behavior. Uses the existing `project.reveal_path()` infrastructure, which handles platform-specific file manager invocation and WSL path conversion.

## Prior art

Every major editor has this in the tab context menu:
- VS Code: "Reveal in Finder" (macOS) / "Reveal in File Explorer" (Windows)
- JetBrains IDEs: Right-click tab -> "Open in" -> "Finder"
- Sublime Text: Right-click tab -> "Reveal in Finder"

Zed already has this in the editor body right-click menu (`Cmd-K R`), project panel (`Alt-Cmd-R`), and outline panel. The tab context menu was the only place it was missing.

This contribution was developed with AI assistance (Claude Code).

Release Notes:

- Added "Reveal in Finder" to the tab context menu

## Screenshot

Reveal in Finder in tab context menu

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -3192,6 +3192,7 @@ impl Pane {
                             });
 
                             let entry_abs_path = pane.read(cx).entry_abs_path(entry, cx);
+                            let reveal_path = entry_abs_path.clone();
                             let parent_abs_path = entry_abs_path
                                 .as_deref()
                                 .and_then(|abs_path| Some(abs_path.parent()?.to_path_buf()));
@@ -3201,6 +3202,15 @@ impl Pane {
 
                             let visible_in_project_panel = relative_path.is_some()
                                 && worktree.is_some_and(|worktree| worktree.read(cx).is_visible());
+                            let is_local = pane.read(cx).project.upgrade().is_some_and(|project| {
+                                let project = project.read(cx);
+                                project.is_local() || project.is_via_wsl_with_host_interop(cx)
+                            });
+                            let is_remote = pane
+                                .read(cx)
+                                .project
+                                .upgrade()
+                                .is_some_and(|project| project.read(cx).is_remote());
 
                             let entry_id = entry.to_proto();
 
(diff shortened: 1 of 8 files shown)


SomeoneToIgnore (changes requested):
This needs a screenshot (https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#sending-changes)

crates/workspace/src/pane.rs
```diff
@@ -3233,6 +3243,29 @@ impl Pane {
...
                                 })
+                                .when(is_local, |menu| {
+                                    menu.when_some(reveal_path, |menu, reveal_path| {
+                                        menu.entry(
+                                            if cfg!(target_os = "macos") && !is_remote {
+                                                "Reveal in Finder"
+                                            } else if cfg!(target_os = "windows") && !is_remote {
+                                                "Reveal in File Explorer"
+                                            } else {
+                                                "Reveal in File Manager"
```
SomeoneToIgnore: This is the 3rd time we repeat the same cfg-based incarnation with string outputs, let's extract it anywhere into the common function and reuse.

SomeoneToIgnore (changes requested):
Thank you for the screenshot, this works.

What's odd is that there's no binding for that action is displayed — that looks wrong given that editor clearly has the default on macOS:

https://github.com/zed-industries/zed/blob/b707f975fa40dce7fb318f652502e0d3696910b3/assets/keymaps/default-macos.json#L599

Image

Apart from that, let's try to do that VSCode and Intellij do, they have separators around "open in" actions, to make them look fully standalone in the menu.

SomeoneToIgnore (commented):
Thank you for the update.

For sharing an action we can use zed_actions module, and `editor` here is a namespace that can be configured, similar to  https://github.com/zed-industries/zed/blob/89c118e6c20f29d495ca0beb39077500f57f67f3/crates/zed_actions/src/lib.rs#L15-L21

that should be enough to break the indirection, hopefully?

Would be also great to have an updated screenshot when this last issue is fixed.

SomeoneToIgnore (commented):
Great, thank you so much — one last thing would be to fix that morbid double separator thing (whilst keeping the new entry surrounded by both) and we're good to go.

crates/workspace/src/pane.rs
```diff
@@ -3233,8 +3243,25 @@ impl Pane {
                                         }),
                                     )
                                 })
+                                .when(is_local, |menu| {
```
SomeoneToIgnore, line 3246: To note, it's a bit odd to see `is_remote` within `is_local` here.

SomeoneToIgnore (approved):
Thank you.

mvanhorn:
Thanks\! Zed's been my daily driver for a while now, felt good to contribute back. Appreciate the warm welcome.

PR #51638, opened 2026-03-16
rope: Handle invalid `Cursor::slice` more gracefully

Veykril:
Release Notes:


Veykril:
Found an actual backtrace on sentry now, so fixing the faulty caller instead

PR #51640, opened 2026-03-16
debugger: Show hover values for stopped sessions

Tortes:
This implements debugger hover values for stopped debug sessions.

Current expression support:
- local identifiers, e.g. `value`
- member expressions, e.g. `obj.field` and `a.b.c`
- index expressions, e.g. `values[1]`

When a debug session is stopped on the same source file, hovering one of these expressions now evaluates it in the active stack frame and shows the runtime value before the LSP hover content.

> This PR is finished with codex;

## Release Notes:

- Improved debugger hovers to show runtime values for stopped sessions, including identifiers, member expressions, and index expressions.

## Screenshots

> Build under Macos26.1

Snipaste_2026-03-18_22-38-49
Snipaste_2026-03-18_22-39-22

## TODO

- [ ] ~~debug the failure with codelldb c++ scenario; Removed for next PR~~

### DIFF
--- a/crates/debugger_ui/src/tests/hover_values.rs
+++ b/crates/debugger_ui/src/tests/hover_values.rs
@@ -1,0 +1,167 @@
+use dap::{
+    StackFrame,
+    requests::{Evaluate, Scopes, StackTrace, Threads},
+};
+use editor::{
+    Editor, EditorMode, MultiBuffer, MultiBufferOffset, SelectionEffects,
+    actions::Hover as HoverAction,
+};
+use gpui::{BackgroundExecutor, TestAppContext, VisualTestContext};
+use language::rust_lang;
+use project::{FakeFs, Project};
+use serde_json::json;
+use util::path;
+
+use crate::{
+    debugger_panel::DebugPanel,
+    tests::{init_test, init_test_workspace, start_debug_session},
+};
+
+const SOURCE: &str =
+    "fn main() {\n    let value = 42;\n    let x = value + 1;\n    println!(\"{}\", x);\n}\n";
+
+fn hover_source<C: gpui::AppContext>(editor: &gpui::Entity<Editor>, cx: &C) -> Option<String> {
+    editor.read_with(cx, |editor, cx| {
+        editor
+            .hover_state
+            .info_popovers
+            .first()
+            .and_then(|popover| {
+                popover.parsed_content.as_ref().map(|markdown| {
+                    markdown.read_with(cx, |markdown, _| markdown.source().to_string())
+                })
+            })
+    })
+}
+
+fn trigger_hover(editor: &gpui::Entity<Editor>, cx: &mut VisualTestContext, offset: usize) {
+    editor.update_in(cx, |editor, window, cx| {
+        editor.change_selections(SelectionEffects::no_scroll(), window, cx, |selection| {
+            selection.select_ranges([MultiBufferOffset(offset)..MultiBufferOffset(offset)])
+        });
+        editor::hover_popover::hover(editor, &HoverAction, window, cx);
+    });
+}
+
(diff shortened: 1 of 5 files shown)


Tortes:
> We require contributors to sign our Contributor License Agreement, and we don't have @Tortes on file. You can sign our CLA at https://zed.dev/cla. Once you've signed, post a comment here that says '@cla-bot check'.

I re-assigned it again, please check it!

crates/project/src/debugger/dap_store.rs
```diff
@@ -617,6 +617,45 @@ impl DapStore {
...
+                            context: Some(EvaluateArgumentsContext::Variables),
+                        })
+                    });
+
+                    variables_eval_task.await.log_err()
+                }
+            }
+        })
+    }
+
```
Anthony-Eid: This function should be in apart of the session type instead of dap store

crates/languages/src/c/debugger.scm
Anthony-Eid, line 1: This should be it's own PR

crates/languages/src/cpp/debugger.scm
Anthony-Eid, line 1: Apart of the c debugger.scm PR

crates/debugger_ui/src/session/running/stack_frame_list.rs
```diff
@@ -496,14 +496,16 @@ impl StackFrameList {
...
+                        BreakpointStore::abs_path_from_buffer(&buffer, cx).unwrap_or(abs_path.clone());
 
                     breakpoint_store.update(cx, |store, cx| {
                         store.set_active_position(
                             ActiveStackFrame {
                                 session_id: this.session.read(cx).session_id(),
                                 thread_id,
                                 stack_frame_id,
-                                path: abs_path,
+                                path: active_buffer_path,
```
Anthony-Eid: Curious about the context of this change, why is it needed?

crates/debugger_ui/src/tests/hover_values.rs
Anthony-Eid, line 1: We should try finding a way to shorten this test file. This can be the last thing we review though

crates/debugger_ui/Cargo.toml
Anthony-Eid, line 1: I would hope we can avoid adding language and lsp to this.

Anthony-Eid (changes requested):
Some changes are needed for this but thank you for contributing so far

crates/project/src/debugger/dap_store.rs
Tortes: Done, moved to session

crates/languages/src/c/debugger.scm
Tortes: Agree, my target is to support c/cpp hover value finally; but first I should support hover value only in this PR

crates/languages/src/cpp/debugger.scm
Tortes: Removed

crates/debugger_ui/src/session/running/stack_frame_list.rs
Tortes: U r right, I  removed it; codex think here should use `Buffer` loc other than `DAP` to get right pos under remote env, but it should always be local for Zed;

crates/debugger_ui/Cargo.toml
Tortes: Removed, it's just used in the test

crates/debugger_ui/src/tests/hover_values.rs
Tortes: Removed some duplication part

Tortes:
Closed for a better solution: https://github.com/zed-industries/zed/pull/53198

PR #50928, opened 2026-03-06
debugger: Fall back to cached adapter binaries when offline

nihalxkumar:
The Python (debugpy) and Go (delve-shim-dap) debug adapters unconditionally require network access on every debug session start — even when the adapter binary is already cached locally. This makes the debugger completely unusable offline.

Python: fetch_debugpy_whl → maybe_fetch_new_wheel hits pypi; failure propagates as a fatal error.
Go: install_shim → fetch_latest_adapter_version hits the GitHub API; failure propagates as a fatal error.
CodeLLDB and JavaScript adapters already handle this correctly.

Fix for Python and Go:

Wrap the network fetch in each adapter with a fallback: if the request fails and a previously downloaded adapter exists on disk, log a warning and use the cached version. If no cache exists, the original error propagates unchanged.

Logs after the fix:

```
2026-03-06T16:31:51+05:30 WARN  [dap_adapters::python] Failed to fetch latest debugpy, using cached version: getting latest release
2026-03-06T16:31:51+05:30 INFO  [dap::transport] Debug adapter has connected to TCP server 127.0.0.1:45533
```

Limitations: 

The debugger must be run at least once with internet connectivity to populate the cache on disk. This PR does not change that requirement.

Closes #45781

Before you mark this PR as ready for review, make sure that you have:
- [ ] Added a solid test coverage and/or screenshots from doing manual testing
  - No automated tests are included. The existing MockDelegate stubs http_client() and fs() as unimplemented!(), so testing the fallback path would require new mock infrastructure. The fix was verified manually by running the debug build offline with and without cached adapters. The CodeLLDB adapter's equivalent fallback (lines 357-374) also has no dedicated test.
- [x] Done a self-review taking into account security and performance aspects

Release Notes:

- Fixed debugger failing to start when offline if a debug adapter was previously downloaded.

Issue #45781, opened 2025-12-28
Debugger doesn't work offline


nihalxkumar:
### Reproduction steps

1. Reboot system into no network
2. Start Debugger in a python file
3. Check Debugger `Console`

### Current vs. Expected behavior

Current behavior: 

when ran debugger without internet

<img width="577" height="100" alt="Console error getting new release " src="https://github.com/user-attachments/assets/0c6bada7-1e2e-4c5f-b9ff-2e762f7fae33" />

Expected behavior:

when ran debugger with internet

<img width="696" height="273" alt="Console rocking" src="https://github.com/user-attachments/assets/0d78ec79-7ca9-4a04-8889-b8ff1776087a" />

### Zed version and system specs

Zed 0.217.3 

0.217.3+stable

### If you are using WSL on Windows, what flavor of Linux are you using?

None

yeskunall:
### Reproduction steps

1. Configure Zed to use Python debugger (Debugpy)
2. Ensure network cannot access pypi.org (e.g., corporate firewall, GFW in China)
3. Start a Python debugging session
4. Wait ~30 seconds before debugging actually starts
5. Stop debugging and restart - same delay occurs again

### Current vs. Expected behavior


## Current behavior

Every time I start a Python debugging session, Zed checks for debugpy updates by requesting `https://pypi.org/pypi/debugpy/json`.

When the network is slow or blocked, this request times out after ~30 seconds, causing significant delay before debugging starts.

From Zed.log:
```
2026-05-29T18:45:10+08:00 INFO  [dap::transport] Debug adapter has connected...
2026-05-29T18:45:39+08:00 ERROR error sending request for url (https://pypi.org/pypi/debugpy/json): client error (Connect): operation timed out
```

Key issues:
1. The HTTP timeout is too long (~30s per request)
2. The update check happens **every time** debugging starts, not just once per session
3. There's no configuration option to disable the update check

The code already falls back to cached debugpy when the request fails (issue #45781, PR #50928), but users still suffer the timeout delay.

## Expected behavior

Option 1: Add a setting to disable debugpy auto-update check:
```json
{
  "debugger": {
    "auto_update_debug_adapters": false
  }
}
```

Option 2: Reduce HTTP timeout to a reasonable value (e.g., 5 seconds)

Option 3: Cache the "update check failed" status so it doesn't retry on every debug session


### Zed version and system specs

- Zed: v1.4.4
- OS: macOS 15.5 (Sequoia)
[...]

PR #51696, opened 2026-03-16
debugger: Fall back to local debugpy when offline

nihalxkumar:
When the debugpy wheel cannot be fetched from PyPI and no cached version exists on disk, check if debugpy is available in the user's Python environment. If so, use `python -m debugpy.adapter` instead of failing.

Log:
```
2026-03-16T23:43:49+05:30 INFO  [dap_adapters::python] Using system-installed debugpy as fallback (fetch failed: getting latest release)
```

---

In https://github.com/zed-industries/zed/pull/50928 we added support for offline debugging which leveraged zed generated cache in `~/.local/share/zed/debug_adapters` or ` ~/Library/Application Support/Zed/debug_adapters/`

Now we have total three layers for debugging.

1. Network fetch
2. Cached adapter
3. System debugpy

---

Related: #45781 

Release Notes:

- Added support for offline python debugging leveraging locally installed debugpy


nihalxkumar:
@Anthony-Eid What do you think about this?

Anthony-Eid:
This change seems reasonable, but I would rather have something that works on the adapter trait itself.

Also, I don't have much bandwidth right now so it's going to be a while until I can review. If you end up going with the adapter agnostic approach I can ask someone on the team to review it

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been inactive for some time without the new impl.

PR #51703, opened 2026-03-16
helix: Fix pasting from the system clipboard

feitreim:
Closes #51693

Helix was unable to paste from the system clipboard, ex: `vim: paste` would work but `helix: paste` would not work. Helix paste was silently requiring the content it was going to paste to have selection metadata exist, and just silently fail if it didn't, and the system clipboard doesn't have that metadata. note: this is not necessarily for parity with helix, as helix didn't seem to support this either in my testing, but rather parity with the other parts of zed, editor mode and vim mode.

single-line paste:

https://github.com/user-attachments/assets/c8696032-d265-4025-9c4c-a8c35dfd2529

multi-line paste:

https://github.com/user-attachments/assets/4bf96033-e13d-4ec1-8a7e-8c56bbc12b94

I also added a new test verifying the behavior.

Release Notes:

- helix: fixed helix paste not pasting from system clipboard.

### DIFF
--- a/crates/vim/src/helix/paste.rs
+++ b/crates/vim/src/helix/paste.rs
@@ -102,13 +102,16 @@ impl Vim {
                     } else if action.before {
                         sel.start
                     } else if sel.start == sel.end {
-                        // Helix and Zed differ in how they understand
-                        // single-point cursors. In Helix, a single-point cursor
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
Thanks for the PR! We did a round of manual testing and it seems to be working as expected.

PR #51706, opened 2026-03-16
Fix switching between helix and vim modes to not require returning to non-modal in between

feitreim:
Closes #51704 

Previously it took three inputs of `workspace: toggle vim mode` to switch from helix mode back to vim mode and three inputs of `workspace: toggle helix mode` to switch to helix mode from vim mode.

Now it takes only one, when the setting is updated, the editor needs to be deactivated and reactivated, this won't cause performance issues because this is not a code path that people hit frequently, if it all. The only reason I hit it so much is contributing to zed and fixing helix mode issues as a vim mode user.

Before:

https://github.com/user-attachments/assets/b7c2f0e6-666e-4bc6-b273-424578d67dc7

After:

https://github.com/user-attachments/assets/00d5e602-8791-4260-b48f-d3055fed8982

Release Notes:

- Fixed switching between vim and helix mode needing multiple calls to have an effect while an editor is already open
- Update Settings UI such that, enabling vim or helix mode, now disables the other if it was previously active

### DIFF
--- a/crates/settings_ui/src/page_data.rs
+++ b/crates/settings_ui/src/page_data.rs
@@ -1294,17 +1294,13 @@ fn keymap_page() -> SettingsPage {
     fn modal_editing_section() -> [SettingsPageItem; 3] {
         [
             SettingsPageItem::SectionHeader("Modal Editing"),
-            // todo(settings_ui): Vim/Helix Mode should be apart of one type because it's undefined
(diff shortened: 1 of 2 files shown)


dinocosta (approved):
Thanks @feitreim! 🙂

I've also pushed a commit updating the Settings UI such that, enabling one of the modes disables the other, in case it is set.

PR #51708, opened 2026-03-16
Fix .git/info/exclude patterns matching against wrong root

mdippery:
Entries in `.git/info/exclude` were not being properly shown as ignored files in the project panel. This was due to the fact that `build_gitignore` was matching against the wrong root when loading entries from `.git/info/exclude`: it was using `.git/info` as the root, but the patterns in that file are relative to the repository root. Entries in `.git/info/exclude` will now be relative to the repository root, and those entries will be shown as ignored in the project panel.

- [X] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Fixed a bug in which entries in `.git/info/exclude` where not properly marked as ignored in the project panel.

### DIFF
--- a/crates/worktree/src/worktree.rs
+++ b/crates/worktree/src/worktree.rs
@@ -3233,12 +3233,20 @@ async fn is_git_dir(path: &Path, fs: &dyn Fs) -> bool {
 }
 
 async fn build_gitignore(abs_path: &Path, fs: &dyn Fs) -> Result<Gitignore> {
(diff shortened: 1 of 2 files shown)


cole-miller:
Thanks!

cole-miller:
Thanks--without realizing it I merged a different fix PR (#57779) for the same issue; sorry for not following up and letting this become stale.

PR #51717, opened 2026-03-16
editor: Re-serialize on file handle change to fix stale path after rename

saberoueslati:
Closes #51629

When a file is renamed via the project panel while open in an editor, Zed would restore the old (no longer existing) path on reload .

On rename, `language::Buffer` emits `BufferEvent::FileHandleChanged`, which propagates to `multi_buffer::Event::FileHandleChanged` → `EditorEvent::TitleChanged`. However, `should_serialize()` only matched `Saved | DirtyChanged | BufferEdited`, so no re-serialization was triggered and the `editors` DB table retained the stale `abs_path`.

The fix adds a dedicated `EditorEvent::FileHandleChanged` variant, emits it alongside `TitleChanged` when the buffer's file handle changes, and adds it to `should_serialize()`. Since `Editor::serialize()` already reads the current path from the buffer at call time, this naturally writes the new path to the DB.

Release Notes:

- Fixed renamed files being restored with their old name after reload

Manual test video below :

[Screencast from 2026-03-16 23-28-46.webm](https://github.com/user-attachments/assets/76a9af97-a9e8-49a3-9712-60e4efcb4753)

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -24283,9 +24283,13 @@ impl Editor {
             }
             multi_buffer::Event::DirtyChanged => cx.emit(EditorEvent::DirtyChanged),
             multi_buffer::Event::Saved => cx.emit(EditorEvent::Saved),
-            multi_buffer::Event::FileHandleChanged
-            | multi_buffer::Event::Reloaded
-            | multi_buffer::Event::BufferDiffChanged => cx.emit(EditorEvent::TitleChanged),
+            multi_buffer::Event::FileHandleChanged => {
+                cx.emit(EditorEvent::TitleChanged);
+                cx.emit(EditorEvent::FileHandleChanged);
+            }
+            multi_buffer::Event::Reloaded | multi_buffer::Event::BufferDiffChanged => {
+                cx.emit(EditorEvent::TitleChanged)
+            }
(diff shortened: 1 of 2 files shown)


crates/editor/src/editor.rs
```diff
@@ -27826,6 +28106,7 @@ pub enum EditorEvent {
     DirtyChanged,
     Saved,
     TitleChanged,
+    FileHandleChanged,
```
SomeoneToIgnore, line 28109: Would it make any sense to reuse `DirtyChanged`?
Seems that the semantics is the same here, "file has changed, reparse"?

SomeoneToIgnore (commented):
Thank you, wondering if there's a possibility to create a test for this?

crates/editor/src/editor.rs
saberoueslati: @SomeoneToIgnore FileHandleChanged is more clear than DirtyChanged I think but in theory yes DirtyChanged can be used

saberoueslati: @SomeoneToIgnore
I looked further, DirtyChanged fires when the buffer's dirty flag flips (clean→dirty or dirty→clean), which means the user started editing or the buffer was saved/reverted.

FileHandleChanged was made to indicate when the name of the file changes, name of the file indicated by file handle in the name

PR #51726, opened 2026-03-17
Fix ctrl-delete keybind in the terminal

injust:
showkey in Ghostty, iTerm2, and Terminal.app reports ctrl-delete as `<ESC>[3;5~`.

I tested this keybind on macOS, where fish_key_reader correctly interprets it as ctrl-delete.

Closes #51725

Release Notes:

- Fixed ctrl-delete keybind in the terminal


kubkon (approved):
Looks great, thanks!

notpeter:
@injust. Thanks for the fix, it works for me as expected in Fish.

In ZSH I still see `5~` with change, but this actually matches the behavior of ZSH in Terminal.app and GhosTTY on macOS, so this is likely the correct behavior for the Terminal (Alacritty) inside Zed.

Apologies to you and @hron for getting this wrong a year ago.  By default ZSH doesn't bind anything to `ctrl-delete` (check with `bindkey | grep '3;5~'`) -- but it can easily bound with `bindkey '^[[3;5~' kill-word`.

Thanks for the follow-through.

injust:
> In ZSH I still see `5~` with change, but this actually matches the behavior of ZSH in Terminal.app and GhosTTY on macOS, so this is likely the correct behavior for the Terminal (Alacritty) inside Zed.

Sounds like the expected behaviour for Zed then. Thanks for verifying!

PR #51727, opened 2026-03-17
git_ui: Disable commit actions during commit message generation

cppcoffee:
Release Notes:

- Add disable commit actions during commit message generation


cole-miller (changes requested):
Thanks! Makes sense, but I don't think we should change the button text to `Generating...`, it's redundant with the existing UI that shows that a commit message is being generated

cole-miller:
Thanks!

cole-miller (approved):

PR #51756, opened 2026-03-17
Reduce agent spinner GPU usage

AndPuQing:
Closes #39532

## Summary

Move the long-running agent "generating" spinner into a dedicated `GeneratingSpinner` view.

Previously, the generating state rendered `SpinnerLabel` inline as part of the full thread view. This change keeps the same spinner styling, but isolates it into its own small view so the animation no longer needs to live directly inside the larger thread UI subtree while the agent is running.

before:

image

after:

image

## Testing

- `cargo run -j 4 ~/Downloads/WorkSpace/zed/`
- `cargo check -p agent_ui`

Release Notes:

- Fixed high GPU usage from the agent panel's generating spinner while an agent is running.


crates/agent_ui/src/conversation_view/thread_view.rs
```diff
@@ -4553,7 +4570,8 @@ impl ThreadView {
                     this.child(
                         h_flex()
                             .w_2()
-                            .child(SpinnerLabel::sand().size(LabelSize::Small)),
+                            .justify_center()
+                            .child(self.generating_spinner.clone()),
```
Anthony-Eid: This would change the confirmation spinner from rendering with the sand animation to the default. I don't think this is what we want here

cc: @danilo-leal For additional feedback if you have any

Anthony-Eid (changes requested):
UI is slightly changed here

crates/agent_ui/src/conversation_view/thread_view.rs
AndPuQing: image

fix

Anthony-Eid:
@AndPuQing Sorry, is there anyway you could give me permission to make an edit directly on your branch? I want to fix the merge conflicts. Also, my patch was incomplete, we shouldn't be passing any of the entities for the spinners around. 

`window.use_state` is a lot cleaner in this case, so i would prefer if we did that

AndPuQing:
> @AndPuQing Sorry, is there anyway you could give me permission to make an edit directly on your branch? I want to fix the merge conflicts. Also, my patch was incomplete, we shouldn't be passing any of the entities for the spinners around.
> 
> `window.use_state` is a lot cleaner in this case, so i would prefer if we did that

I have enabled "Allow edits by maintainer" and I have also invited you to join my forked collaborator group.

Anthony-Eid (approved):

steckes:
It is released! TY! :)

AndPuQing:
image

Seems like it worked well?

Anthony-Eid:
> Seems like it worked well?

Yes, but we can still do better. Next week I'm going to see what else I can do to improve gpu usage in the agent panel because that's still high

steckes:
Awesome! Yes on Linux I still have up to 30% GPU usage just from the agent panel.

PR #51761, opened 2026-03-17
editor: Accumulate consecutive KillRingCut line kills

### LINKED ISSUE #22490: Editor::KillRingCut should support for multiple lines (emacs)
### Check for existing issues

- [X] Completed

I noticed that emacs-like "kill-ring" cut/yank was added recently. My base keymap is Jetbrains and in my custom `~/.config/zed/keymap.json` I have this:

```json
 {
    "context": "Editor",
    "use_key_equivalents": true,    
    "bindings": {
      "ctrl-y": "editor::KillRingYank", 
      "ctrl-k": "editor::KillRingCut"      
    }
  },
```
However I'm finding that `Ctrl-k` only adds a single like to the kill ring, i.e. when I consecutively kill multiple lines, and then yank with `Ctrl-y`, I just get the last line killed.

Not sure if I need to set up anything else for this to work.

Zed: v0.166.1 (Zed)
OS: macOS 15.0.0
Memory: 64 GiB
Architecture: aarch64

### If applicable, add screenshots or screencasts of the incorrect state / behavior

_No response_

```

```
</details>

loadingalias:
Closes #22490

## Summary
 • Keep kill-ring state as a dedicated global structure (`text`, `metadata`, `row`, `column`, `buffer_id`) instead of a raw clipboard wrapper.
 • Update `editor::KillRingCut` to append consecutive empty-line kills only when all are true:
   - one selection only
   - previous selection was empty
   - same row/column
   - same buffer

 • Preserve existing behavior for non-empty and multi-selection cuts by replacing the ring payload instead of appending.
 • Preserve yank metadata behavior through the existing clipboard selection metadata path.
 • Add regression coverage in `editor_tests::test_kill_ring_cut_accumulates_multi_line_kills`.

## Verification
 • `cargo fmt --all -- --check`
 • `./script/check-keymaps`
 • `./script/clippy -p editor`
 • `cargo test -p editor -- --nocapture`
 • `cargo test -p editor editor_tests::test_kill_ring_cut_accumulates_multi_line_kills -- --exact`
 • `cargo test -p editor test_cut_line_ends -- --exact`

## Manual
 • `Ctrl+K`, `Ctrl+K`, `Ctrl+Y` with repeated line-end kills at same caret location pastes concatenated text.
 • Non-empty selection cut and multi-selection still replace as before (no unintended append).
 • Cut in one buffer then cut in another does not cross-append.
 • Moving caret between kills does not append.
 • Undo/redo around cut+yank path works cleanly.

Release Notes:

- Fixed Emacs-style kill-ring behavior so successive `KillRingCut` operations at the same caret context are accumulated and pasted together via one `KillRingYank`, while preserving existing non-empty selection and multi-selection semantics.

### DIFF
--- a/crates/editor/src/clipboard.rs
+++ b/crates/editor/src/clipboard.rs
@@ -354,6 +354,22 @@ impl Editor {
         if self.read_only(cx) {
             return;
         }
+        let selection_count = self.selections.count();
+        let first_selection = self.selections.first_anchor();
+        let snapshot = self.buffer.read(cx).snapshot(cx);
+        let first_selection_is_empty = first_selection.start == first_selection.end;
+        let selection_start_point = first_selection.start.to_point(&snapshot);
+        let selection_start_row = selection_start_point.row;
+        let selection_start_column = selection_start_point.column;
+        let Some((_, text_anchor)) = self
+            .buffer
+            .read(cx)
+            .text_anchor_for_position(first_selection.start, cx)
+        else {
+            return;
+        };
+        let buffer_id = text_anchor.buffer_id;
+
         self.change_selections(SelectionEffects::no_scroll(), window, cx, |s| {
             s.move_with(&mut |snapshot, sel| {
                 if sel.is_empty() {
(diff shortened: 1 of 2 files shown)


loadingalias:
> recording_20260331_211012.mp4

Yeah, this is fixed. Pushing now.

ConradIrwin:
Sorry for the slow response here; I'm happy to merge this, but it seems like the underlying system has changed a bit since you wrote the code.

Would you be able to rebase? Otherwise we can hold out for @Ziqi-Yang's PR.

loadingalias:
Yeah, I'll rebase tonight and run the local checks, too.

ConradIrwin:
@loadingalias I'm still seeing a bunch of CI failures

loadingalias:
Whoa. I don't know how this has gone unnoticed until now. I just got an alert about it not being stale.

I will close the CI failures. When I pushed - it was clean. A lot has changed between my PR and the review. I will clean it up. Give me 48-hours. 

I'm sorry for the delay - I'd assumed this was merged a long time ago.

loadingalias:
Rebased to `main` and ported the fix to the current editor layout.

The original branch added the kill-ring accumulation logic in `editor.rs`, but current `main` has moved the cut/yank clipboard behavior into `crates/editor/src/clipboard.rs`. I kept the current layout and moved the fix there instead of restoring the old `editor.rs` impl.

This still addresses #22490: repeated `KillRingCut` calls at the same empty-selection caret context accumulate into one kill-ring payload, and `KillRingYank` pastes that accumulated payload. Yank now also breaks the append chain, so later kills do not append to stale yanked content. The EOF/no-trailing-newline case is covered by regression tests too.

I also added a guard for accumulated kill-ring metadata so one accumulated kill-ring entry is pasted as a whole at each cursor instead of being split across multiple cursors.

This should close the issue. Verification was identical.

ChristopherBiscardi (approved):
Thanks! I've confirmed that this improves the kill-ring behavior and cleaned it up a bit.

PR #51762, opened 2026-03-17
agent: Allow all settings in the thread to apply immediately

benbrandt:
Before we kind of had it mixed: you could change thinking but not model etc. We brute forced a solution by disabling, but it seems better to just allow all of them to be updated whenever we would build the request.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/model_selector_popover.rs
+++ b/crates/agent_ui/src/model_selector_popover.rs
@@ -13,7 +13,6 @@ use crate::{ModelSelector, model_selector::acp_model_selector};
 pub struct ModelSelectorPopover {
     selector: Entity<ModelSelector>,
     menu_handle: PopoverMenuHandle<ModelSelector>,
-    disabled: bool,
 }
 
 impl ModelSelectorPopover {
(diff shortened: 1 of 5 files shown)


crates/agent_ui/src/model_selector_popover.rs
```diff
@@ -73,33 +61,25 @@ impl Render for ModelSelectorPopover {
...
-                }
-            }))
-        };
+        let tooltip: Box<dyn Fn(&mut Window, &mut App) -> AnyView> = Box::new(Tooltip::element({
+            move |_, _cx| {
+                ModelSelectorTooltip::new()
+                    .show_cycle_row(show_cycle_row)
+                    .into_any_element()
+            }
+        }));
```
agu-z: nit: I think boxing this function is no longer necessary since `PickerPopoverMenu` is generic over it

agu-z (approved):
Looks good! Left a minor nit comment

crates/agent_ui/src/model_selector_popover.rs
benbrandt: good catch! was there because of the if before but not needed

PR #51766, opened 2026-03-17
markdown: Fix visible escape characters in LSP diagnostics

### LINKED ISSUE #51622: Markdown escapes visible in LSP errors
### Reproduction steps

1. Start Zed
2. Make sure the luau extension is installed
3. Use settings below
5. Make a `.luau` file
6. Put:

```luau
--!strict
type foo = { a: string } | { b: number }
local foo: foo = { c = {} }
```

### Current behavior

Image

The error message contains spaces that become code block in Markdown, but those don't need the escape characters Zed inserts and Zed does not take this into account when doing markdown escape.

### Expected behavior

No markdown escapes, like in the image below taken in vscode.

Image

OS: macOS 26.1
Memory: 18 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json
{
  "lsp": {
    "luau-lsp": {
      "settings": {
        "fflags": {
          "enable_new_solver": true,
        },
      },
    },
  },
}
```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Dnreikronos:
Closes #51622

Release Notes:

- Fixed markdown escape characters being visible in LSP diagnostic messages when leading whitespace caused indented code blocks

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -3145,62 +3145,85 @@ mod tests {
         );
     }
 
-    #[test]
-    fn test_escape() {
-        assert_eq!(Markdown::escape("hello `world`"), "hello \\`world\\`");
-        assert_eq!(
-            Markdown::escape("hello\n    cool world"),
-            "hello\n\n\u{00A0}\u{00A0}\u{00A0}\u{00A0}cool world"
-        );
-
-        // Leading whitespace on the first line is replaced with non-breaking
-        // spaces to preserve indentation without triggering code blocks.
-        assert_eq!(
-            Markdown::escape("    | { a: string }"),
-            "\u{00A0}\u{00A0}\u{00A0}\u{00A0}\\| \\{ a\\: string \\}"
-        );
-        assert_eq!(Markdown::escape("    hello"), "\u{00A0}\u{00A0}\u{00A0}\u{00A0}hello");
-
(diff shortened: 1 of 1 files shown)


smitbarmase (changes requested):
Thanks, I think we should not strip the leading whitespace since it might lead to stylistic changes that LSP didn't advertise. Can we replace these whitespaces with `&nbsp;` so that Markdown renderer still adds those intended spacing, without turning it into a codeblock?

Dnreikronos:
> Thanks, I think we should not strip the leading whitespace since it might lead to stylistic changes that LSP didn't advertise. Can we replace these whitespaces with `&nbsp;` so that Markdown renderer still adds those intended spacing, without turning it into a codeblock?

Just adjusted as you requested!

smitbarmase (approved):
Thanks, I made a couple of fixes here:

1. Tabs in leading whitespace were getting replaced with a single non-breaking space instead of 4. This means tab-indented diagnostics would look collapsed. Fixed it by expanding each tab to 4 NBSPs.

2. `String::with_capacity` was underestimating. It accounted for doubled newlines and backslash prefixes but missed the NBSP expansion (each `\u{00A0}` is 2 bytes in UTF-8 vs 1 for the original space/tab). The capacity is now computed exactly upfront.

I think now it is in right shape! Thanks again.

PR #51776, opened 2026-03-17
vim/helix: Use grapheme count on replace

feitreim:
Closes #51772 

Multibyte characters are not handled properly with vim/helix replace, they replace based on bytes, instead of characters, so you will end up with too many new characters.

Vim Behavior Before:

https://github.com/user-attachments/assets/55b22f43-e403-4d09-9dcf-53ce9395c2a4

Vim Behavior After:

https://github.com/user-attachments/assets/d08b3f18-23e6-4c59-b2b3-a938fe4a27eb

Helix Before:

https://github.com/user-attachments/assets/3d4d4227-d6d6-4309-9620-852d6778bd57

Helix After:

https://github.com/user-attachments/assets/47051062-75fb-496e-9063-8079eea2e6c1

I also added tests for the functionality.

Release Notes:

- Fixed vim/helix's replace action to take into consideration grapheme count

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10735,6 +10735,7 @@ dependencies = [
  "theme",
  "tracing",
  "tree-sitter",
+ "unicode-segmentation",
  "util",
  "zlog",
  "ztracing",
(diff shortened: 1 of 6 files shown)


dinocosta (approved):
Thanks @feitreim ! 🙂

Pushed a smaller commit updating this to take into consideration graphemes instead so as to tackle even more cases like certain emoji and unicode decomposed characters.

PR #51780, opened 2026-03-17
editor: Fix multi-line cursor expansion when multi-byte characters are involved

feitreim:
Closes #51740 

The multi-line cursor expansion operates off of byte offsets, instead of character offsets, so multi-byte characters like the umlaut cause the multi-line cursors to be weirdly offset. To fix we just convert the expansion logic to rely on utf16 characters instead of bytes.

before behavior:

https://github.com/user-attachments/assets/320e24e9-0fdd-4d16-a9e8-ca17c9e21ff2

after behavior: 

https://github.com/user-attachments/assets/c4f0334b-dffc-4530-91ee-577b4fab75dd

+ test to verify functionality.

Release Notes:

- editor: fixed multi-line cursor expansion dealing with multi-byte characters.

### DIFF
--- a/crates/editor/src/selections_collection.rs
+++ b/crates/editor/src/selections_collection.rs
@@ -7,7 +7,7 @@ use std::{
 use collections::HashMap;
 use gpui::Pixels;
 use itertools::Itertools as _;
-use language::{Bias, Point, Selection, SelectionGoal};
+use language::{Bias, Point, PointUtf16, Selection, SelectionGoal};
 use multi_buffer::{MultiBufferDimension, MultiBufferOffset};
 use util::post_inc;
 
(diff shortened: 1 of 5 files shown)


crates/editor/src/selections_collection.rs
```diff
@@ -408,35 +408,39 @@ impl SelectionsCollection {
     }
 
     /// Attempts to build a selection in the provided buffer row using the
-    /// same buffer column range as specified.
+    /// same UTF-16 column range as specified.
     /// Returns `None` if the range is not empty but it starts past the line's
     /// length, meaning that the line isn't long enough to be contained within
     /// part of the provided range.
-    pub fn build_columnar_selection_from_buffer_columns(
+    pub fn build_columnar_selection_from_utf16_columns(
```
SomeoneToIgnore:
```suggestion
    fn build_columnar_selection_from_utf16_columns(
```

Looks like we can be on a safe side and not expose something that's not used externally.

SomeoneToIgnore (approved):
Thank you!

PR #51784, opened 2026-03-17
theme: Expose editor diff hunk colors

vitallium:
This PR makes editor diff hunk colors configurable from themes, instead of hardcoded values.

It introduces 6 new optional theme keys:

- `editor.diff_hunk.added.background`
- `editor.diff_hunk.added.hollow_background`
- `editor.diff_hunk.added.hollow_border`
- `editor.diff_hunk.deleted.background`
- `editor.diff_hunk.deleted.hollow_background`
- `editor.diff_hunk.deleted.hollow_border`

When a theme omits these keys, each color falls back to the existing version-control color with the previous hardcoded opacity values:

- Light defaults:
  - background_opacity = 0.16
  - hollow_background_opacity = 0.08
  - hollow_border_opacity = 0.48
- Dark defaults:
  - background_opacity = 0.12
  - hollow_background_opacity = 0.06
  - hollow_border_opacity = 0.36

There is an existing feature request https://github.com/zed-industries/zed/discussions/51667

## Screenshots

I used `Modus Themes` (Modus Vivendi) since these themes provide highly accessible themes.

Original version:

CleanShot 2026-03-17 at 20 26 41@2x

This version:

CleanShot 2026-03-17 at 20 23 09@2x

Release Notes:

- Added theme keys for configuring editor diff hunk colors.


MrSubidubi (changes requested):
Thanks for this!

Is there a particular reason as to why we just don't make the entire color configurable and have the current colors with the hardcoded opacities as a fallback? Also, do we have floats in the themes anywhere else?

vitallium:
>Is there a particular reason as to why we just don't make the entire color configurable and have the current colors with the hardcoded opacities as a fallback?

@MrSubidubi The base background color for diff hunks is already configured via `version_control` colors, so I opted to keep them as base colors but expose opacity configurations. Is your suggestion about definining new keys for `diff_hunk.` keys like `editor.diff_hunk.added.background`, `editor.diff_hunk.added.hollow_background` and so on?

crates/theme_settings/src/schema.rs
```diff
@@ -576,6 +576,30 @@ pub fn theme_colors_refinement(
...
+            .as_ref()
+            .and_then(|color| try_parse_color(color).ok()),
+        editor_diff_hunk_deleted_hollow_background: this
+            .editor_diff_hunk_deleted_hollow_background
+            .as_ref()
+            .and_then(|color| try_parse_color(color).ok()),
+        editor_diff_hunk_deleted_hollow_border: this
+            .editor_diff_hunk_deleted_hollow_border
+            .as_ref()
+            .and_then(|color| try_parse_color(color).ok()),
```
MrSubidubi: If we want to go ahead and add this, we should also be adding fallbacks here on a per-theme basis so that they fallback to the colors that would previously have been in place.

crates/editor/src/element.rs
```diff
@@ -9916,37 +9914,33 @@ impl Element for EditorElement {
...
-                            }
-                        };
-
-                        let hunk_opacity = if is_light { 0.16 } else { 0.12 };
+                        let (filled_background_color, hollow_background_color, hollow_border_color) =
+                            match diff_status.kind {
+                                DiffHunkStatusKind::Added => (
+                                    cx.theme().colors().editor_diff_hunk_added_background,
+                                    cx.theme().colors().editor_diff_hunk_added_hollow_background,
+                                    cx.theme().colors().editor_diff_hunk_added_hollow_border,
```
MrSubidubi: Let`s move for this and the one below the `cx.theme().colors()` into a variable, please.

MrSubidubi (changes requested):

crates/editor/src/element.rs
vitallium: Done, thanks!

crates/theme_settings/src/schema.rs
vitallium: Done, thanks!

MrSubidubi (approved):
Sorry for letting this sit for so long, looks good! Thanks!

PR #51801, opened 2026-03-18
Allow modals to opt out of blocking autosave

### LINKED ISSUE #47968: Inconsistent save behavior with vim mode and autosave on_focus_change
### Reproduction steps

Save behavior in vim mode with auto-save set to { "autosave": "on_focus_change" } is not consistent anymore Whether autosave is performed when switching from one edited file to another now depends on whether the switch is performed using a modal dialog or directly. In the first case, the first file is not saved automatically, in the second case it is. 

Steps to reproduce:

1. Edit file1
2. Switch to file2 with tab switch (`ctrl + ⇥`) -> file1 not saved

1. Edit file1
2. Switch to file2 with vim shortcut (`g t`) -> file1 is saved

I think the problem is that since PR #45166 autosave is now disabled for all modal dialogs, which include all file switching dialogs.

https://github.com/user-attachments/assets/1cd42a08-f5c9-4631-9b54-a1496f07869d

The video shows switching from one edited file to another. When using tab switcher or file finder the file is not save (not expected), when clicking on the tab of the other file or using vim motion `g t` is file is saved (expected).

Zed 0.221.4 
OS: macOS 15.7.3
Memory: 128GB
Architecture: aarch64

Zed 0.221.4 
OS: Ubuntu 22.04
Memory: 96GB
Architecture: x86-64

mchisolm0:
Currently, autosave behavior is guarded by 

1. if the item has lost focus and 
2. an unconditional check if a modal is active. 

This allows the TabSwitcher to block autosave when it gains focus, preventing autosave from ever triggering for the file which lost focus. 

This PR allows modals to opt out of blocking autosave, however, I am concerned having autosave trigger immediately on TabSwitcher getting focus may cause the `unwanted formatting and cursor jumps` mentioned in the comment above the `item.rs` change. I have a PR built on this one which only triggers autosave when the TabSwitcher focuses the *new* tab (#51802)

- Add ModalView::blocks_focus_change_autosave with a default of true so existing modals continue to block autosave.
- Thread the check through ModalViewHandle, ModalLayer, and Workspace and use it when deciding whether to autosave items on focus change.
- Implement TabSwitcher to return false.

Closes #47968

- N/A Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

<details><summary>Recording of fix</summary><video src="https://github.com/user-attachments/assets/3902c8bc-d050-4116-a731-c9af748aeb15">Unable to load video</video></details>

Release Notes:

- Fixed tab switcher not triggering autosave

### DIFF
--- a/crates/tab_switcher/src/tab_switcher.rs
+++ b/crates/tab_switcher/src/tab_switcher.rs
@@ -56,7 +56,11 @@ pub struct TabSwitcher {
     init_modifiers: Option<Modifiers>,
 }
 
-impl ModalView for TabSwitcher {}
+impl ModalView for TabSwitcher {
+    fn blocks_focus_change_autosave(&self) -> bool {
+        false
+    }
+}
 
 pub fn init(cx: &mut App) {
     cx.observe_new(TabSwitcher::register).detach();
--- a/crates/workspace/src/item.rs
+++ b/crates/workspace/src/item.rs
@@ -950,7 +950,7 @@ impl<T: Item> ItemHandle for Entity<T> {
                         // since the user is still interacting with the workspace.
                         let focus_handle = item.item_focus_handle(cx);
                         if !focus_handle.contains_focused(window, cx)
-                            && !workspace.has_active_modal(window, cx)
+                            && !workspace.active_modal_blocks_focus_change_autosave(cx)
                         {
                             Pane::autosave_item(&item, workspace.project.clone(), window, cx)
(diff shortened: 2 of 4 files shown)


maxdeviant:
Thanks for taking a look!

We have an alternative solution open in https://github.com/zed-industries/zed/pull/51949 that we are discussing that would supersede this one.

PR #51808, opened 2026-03-18
languages: Add outline for diff files

nishanthkarthik:

The diff file in the unit test maps to this outline below. It covers most of the format variations on the [tree-sitter-diff corpus](https://github.com/tree-sitter-grammars/tree-sitter-diff/blob/main/test/corpus/text.txt).

image

Also added a missing `tree-sitter-rust.workspace = true`. The tests did not build without this declaration.

Release Notes:

- Added outline definition for diff files


crates/languages/Cargo.toml
```diff
@@ -103,8 +103,10 @@ tree-sitter-bash.workspace = true
 tree-sitter-c.workspace = true
 tree-sitter-cpp.workspace = true
 tree-sitter-css.workspace = true
+tree-sitter-diff.workspace = true
 tree-sitter-go.workspace = true
 tree-sitter-python.workspace = true
+tree-sitter-rust.workspace = true
```
SomeoneToIgnore, line 109: Why is this needed?

SomeoneToIgnore (commented):
Thank you, this looks like a nice idea but the implementation is a bit useless?

Here, for example, for `trie.rs` I can navigate to 2 neighbor `---` and `+++` lines, and that's it, despite the diff having 5 hunks (or whatever I may call regions like `@@ -3,7 +3,7 @@`).

Image

This does not seem too useful, any way we can use hunks instead of/along with the file names too?

CI also wants the tree-sitter file to be reformatted.

PR #48338, opened 2026-02-04
git_ui: Fix branch picker deleting remote instead of remote branch

amtoaer:
Closes #48256

It appears that the current git remotes support was implemented in #42819, following the design described in https://github.com/zed-industries/zed/pull/42486#issuecomment-3524092306. That design does not include an interaction for deleting remotes, but the delete remote branch action was mistakenly implemented as deleting the remote itself.

After this PR, there should be no code paths that use `remove_remote` anymore. I've kept it for now though, as it may be useful when we introduce the corresponding interaction in the future.

Release Notes:

- Fixed a bug where deleting a remote branch from the branch picker would incorrectly remove the entire remote configuration

PR #51825, opened 2026-03-18
git_ui: Improve delete branch tests to verify underlying repository state

amtoaer:
While fixing the tests in the previous PR (#48338), I noticed that the `delete_branch` tests only verified the UI state changes rather than the actual state of the underlying repository.

Currently, if `delete_branch` were a no-op that returned Ok(), the tests would still pass incorrectly. This PR addresses that gap by ensuring the branch is actually removed from the repository.

|Before|After|
|--|--|
|图片|CleanShot 2026-03-18 at 17 51 36@2x|

Release Notes:


Veykril (approved):
Thanks!

PR #51826, opened 2026-03-18
language_model: Clear the LlmApiToken first on org switch

tomhoule:
When we switch organizations, we try and refresh the token. If the token
refresh fails, we are left with the old LlmApiToken, which is for the
wrong organization. In this commit, we make sure to clear the old token
before trying a refresh on organization switch.

Release Notes:

- N/A

### DIFF
--- a/crates/language_model/src/model/cloud_model.rs
+++ b/crates/language_model/src/model/cloud_model.rs
@@ -96,7 +96,9 @@ impl LlmApiToken {
                 *lock = Some(response.token.0.clone());
                 Ok(response.token.0)
             }
-            Err(err) => match err {
+            Err(err) => {
+                *lock = None;
+                match err {
                 ClientApiError::Unauthorized => {
                     client.request_sign_out();
                     Err(err).context("Failed to create LLM token")
--- a/crates/language_model/src/model/cloud_model.rs
+++ b/crates/language_model/src/model/cloud_model.rs
@@ -63,6 +63,20 @@ impl LlmApiToken {
         Self::fetch(self.0.write().await, client, organization_id).await
     }
 
+    /// Clears the existing token before attempting to fetch a new one.
+    ///
+    /// Used when switching organizations so that a failed refresh doesn't
+    /// leave a token for the wrong organization.
+    pub async fn clear_and_refresh(
+        &self,
+        client: &Arc<Client>,
+        organization_id: Option<OrganizationId>,
+    ) -> Result<String> {
+        let mut lock = self.0.write().await;
+        *lock = None;
+        Self::fetch(lock, client, organization_id).await
+    }
+
     async fn fetch(
         mut lock: RwLockWriteGuard<'_, Option<String>>,
         client: &Arc<Client>,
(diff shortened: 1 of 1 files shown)


crates/language_model/src/model/cloud_model.rs
```diff
@@ -63,6 +63,20 @@ impl LlmApiToken {
...
+    /// leave a token for the wrong organization.
+    pub async fn clear_and_refresh(
+        &self,
+        client: &Arc<Client>,
+        organization_id: Option<OrganizationId>,
+    ) -> Result<String> {
+        let mut lock = self.0.write().await;
+        *lock = None;
+        Self::fetch(lock, client, organization_id).await
+    }
```
tomhoule, line 78: Has to be a separate method to take advantage of having that mutex guard available, so we can do the clear and the fetch atomically.

crates/language_model/src/model/cloud_model.rs
```diff
@@ -63,6 +63,20 @@ impl LlmApiToken {
...
 
+    /// Clears the existing token before attempting to fetch a new one.
+    ///
+    /// Used when switching organizations so that a failed refresh doesn't
+    /// leave a token for the wrong organization.
+    pub async fn clear_and_refresh(
+        &self,
+        client: &Arc<Client>,
+        organization_id: Option<OrganizationId>,
+    ) -> Result<String> {
```
tomhoule, line 74: Should we add a TokenRefreshMode argument to refresh() instead? I'm thinking maybe not because that would force us to change "many" call sites.

crates/language_model/src/model/cloud_model.rs
NeelChotai: Would it be possible to have the clearing logic on this match arm instead? It would fire for any errors fetching, organization or not. That could be a potential downside on `Other` but the token will be regenerated on the next LLM call anyway

tomhoule: That would completely avoid changing the API of LlmApiToken, so that sounds good. There's one case I see where clearing before trying to refresh could still be better: when the refresh request takes forever due to network or server side delays. For the whole time the request is in flight, LLM API requests would still we issued with the old token, right? We could also do both (proactive clearing, and add the clearing here too, but that would only affect the behaviour of regular refresh calls, not clear_and_refresh). WDYT?

maxdeviant: I think we want to be proactively clearing it on organization switch, rather than waiting for an error response.

NeelChotai: In that case, my preference would be to add `clear` to `LlmApiToken` and call it on the org switch path + the error path

tomhoule: Added this in the error branch:

```rust
                ClientApiError::Other(err) => {
                    *lock = None;

                    Err(err)
                }
```

NeelChotai: > In that case, my preference would be to add `clear` to `LlmApiToken` and call it on the org switch path + the error path

I just pulled the code down and realised why this isn't really possible, my suggestion was instead of having a clear and refresh at all we would just clear it on the `Err` branch always. Since we already overwrite it in the `Ok` branch, there shouldn't be old tokens around.

But the thing about using the old token whilst the request is still in-flight makes sense also.

NeelChotai: think the lock clearing should be here

```suggestion
            Err(err) => {
                *lock = None;
                match err {
```

not sure if sign out clears it

tomhoule: Indeed, it doesn't. Applying your change.

NeelChotai (approved):

PR #51833, opened 2026-03-18
ci: Add merge queue trigger

MrSubidubi:
This adds a merge queue trigger to our main CI job to allow for a merge queue which runs a subset of our tests before merging to main. 

With this setup, merging to main would probably take around one to two minutes and would not catch all but definitely most issues. We could extend it here though as we'd like. 

Release Notes:

- N/A


ConradIrwin (approved):
Thanks.

Instead of trying to re-use the existing run tests workflow, which is already very complicated. Let's generate a separate workflow that does this. We can use the xtask workflows stuff so we don't need to duplicate step definitions, but I think it'd be easier to maintain going forward if we have a workflow that lists a small number of checks, not a workflow that lists a large number of checks most of which are disabled

PR #51849, opened 2026-03-18
gpui: Add MagnifyEvent for trackpad pinch-to-zoom gestures

scott-wyatt:
## Summary

Adds macOS trackpad magnification (pinch-to-zoom) gesture support to GPUI's event pipeline. macOS delivers these via `magnifyWithEvent:` / `NSEventTypeMagnify = 30`, and the `cocoa` crate 0.26 already exposes both. GPUI's `from_native()` previously had a `_ => None` catch-all that silently swallowed these events.

- **`MagnifyEvent`** struct with `position`, `magnification` (incremental delta), `touch_phase`, and `modifiers`
- **`PlatformInput::Magnify`** variant with proper `mouse_event()`/`keyboard_event()` classification
- **`NSEventTypeMagnify`** → `PlatformInput::Magnify` conversion in `platform/mac/events.rs`
- **`magnifyWithEvent:`** selector registration in `platform/mac/window.rs`
- **`dispatch_event`** handling in `window.rs` (position + modifier tracking)
- **`on_magnify` / `capture_magnify`** fluent + imperative APIs in `div.rs`
- **`simulate_magnify`** test helper on `VisualTestContext`
- **4 tests**: handler dispatch, delta accumulation, default values, platform input classification

### Cross-platform safety

The `MagnifyEvent` struct exists in cross-platform code but is only constructed on macOS — same pattern as `MousePressureEvent`. Linux/Windows/test platforms are unaffected.

### Design decisions

- `on_magnify` uses `should_handle_scroll()` (not `is_hovered()`) for the hitbox check, matching `on_scroll_wheel` behavior — gesture events carry their own position and shouldn't require prior mouse movement
- `magnification` is an incremental delta (positive = zoom in, negative = zoom out), matching the native `NSEvent.magnification()` API

## Test plan

- [x] `cargo test -p gpui --features test-support -- magnify` — 4/4 pass
- [x] `cargo test -p gpui --features test-support` — 82/82 pass, 0 regressions
- [ ] Manual: run app, open canvas, pinch trackpad — zoom should work


MrSubidubi (changes requested):
Thanks for this! I think the conflicts here might very well come from the fact that we recently added the `on_pinch` event handlers here. Can you elaborate how that differs from that? I think that both basically handle the same and you should have support for this by just updating GPUI in your project

PR #51867, opened 2026-03-18
Add Grok 4.20 models to xAI

vdawger:
Adds support for the latest Grok 4.20 model variants from xAI (multi-agent, reasoning, and non-reasoning).

These are now available in the model selector when using an xAI API key.

Release Notes:

- Added Grok 4.20 model variants


archerallstars:
The commit is the beta version. But now, the stable version of Grok 4.2 has come out. I hope it will be included soon.

benbrandt:
Thanks this got addressed in #55931

PR #51886, opened 2026-03-19
git_graph: Implement basic search functionality

Anthony-Eid:
## Context

This uses `git log` to get a basic search working in the git graph. This is one of the last blockers until a full release, the others being improvements to the graph canvas UI. 

## Self-Review Checklist

Release Notes:

- N/A


crates/git/src/repository.rs
```diff
@@ -95,6 +190,21 @@ pub struct InitialGraphCommitData {
...
+    if is_regex {
+        let regex = regex::RegexBuilder::new(query)
+            .case_insensitive(true)
+            .build()?;
+        Ok(regex.is_match(message))
+    } else {
+        Ok(message.to_lowercase().contains(&query.to_lowercase()))
+    }
+}
+
```
cole-miller: Unused?

crates/git_graph/src/git_graph.rs
```diff
@@ -1128,18 +1183,69 @@ impl GitGraph {
...
                         .truncate()
                         .into_any_element()
                 };
 
+                let subject_label = if is_matched {
+                    let query = match &self.search_state.state {
+                        QueryState::Confirmed((query, _)) => Some(query.clone()),
+                        _ => None,
+                    };
+                    let highlight_indices = query
```
cole-miller: Could we add a constructor for HighlightedLabel that takes ranges directly, since that seems to be what it wants internally? Then we could use `match_indices` from the stdlib or something similar instead of having to deal in detail with indices like this

crates/git_graph/src/git_graph.rs
```diff
@@ -532,6 +563,7 @@ struct GraphData {
     lines: Vec<Rc<CommitLine>>,
     active_commit_lines: HashMap<CommitLineKey, usize>,
     active_commit_lines_by_parent: HashMap<Oid, SmallVec<[usize; 1]>>,
+    oid_to_line: HashMap<Oid, usize>,
```
cole-miller: Unused?

crates/git_graph/src/git_graph.rs
```diff
@@ -1197,14 +1308,68 @@ impl GitGraph {
...
+        let search_task = cx.spawn(async move |this, cx| {
+            while let Ok(oid) = request_rx.recv().await {
+                this.update(cx, |this, cx| {
+                    if this.search_state.selected_index.is_none() {
+                        this.search_state.selected_index = Some(0);
+                        this.select_commit_by_sha(oid, cx);
+                    }
+
+                    this.search_state.matches.insert(oid);
+                    cx.notify();
```
cole-miller, line 1345: Opus had what I thought was a good suggestion: add a `while let Some() = try_recv` loop here so we insert all the matches we can every time this task wakes up. That way we can notify less often

cole-miller (changes requested):

cole-miller (approved):

PR #51897, opened 2026-03-19
repl: Use uv to install ipykernel for uv-managed venv

feitreim:
## Context

Closes #51874

the repl is able to recognize that the venv is managed by uv, but still runs `python -m pip install ipykernel`, despite this not working. this PR fixes that behavior and uses uv to install ipkernel.

## How to Review

Added a path that uses uv to install ipykernel in repl_editor.rs
Added a function to repl_store.rs that allows updating the venv as having ipykernel installed after installing it.

## Videos

Old Behavior:

https://github.com/user-attachments/assets/9de81cc9-cd78-4570-ad57-550f5ecabffa

New Behavior:

https://github.com/user-attachments/assets/391f54c7-ae67-4d85-8f4f-9d87ddc8db63

## Self-Review Checklist

Release Notes:

- repl: Fixed installing ipykernel with uv managed environements

### DIFF
--- a/crates/repl/src/repl_editor.rs
+++ b/crates/repl/src/repl_editor.rs
@@ -87,6 +87,10 @@ pub fn install_ipykernel_and_assign(
 
     let python_path = env_spec.path.clone();
     let env_name = env_spec.name.clone();
+    let is_uv = env_spec
+        .environment_kind
+        .as_deref()
+        .is_some_and(|kind| kind.starts_with("uv"));
     let env_spec = env_spec.clone();
 
     struct IpykernelInstall;
@@ -109,11 +113,25 @@ pub fn install_ipykernel_and_assign(
     let window_handle = window.window_handle();
 
     let install_task = cx.background_spawn(async move {
-        let output = util::command::new_command(python_path.to_string_lossy().as_ref())
-            .args(&["-m", "pip", "install", "ipykernel"])
-            .output()
-            .await
-            .context("failed to run pip install ipykernel")?;
+        let output = if is_uv {
+            util::command::new_command("uv")
+                .args(&[
+                    "pip",
+                    "install",
+                    "ipykernel",
+                    "--python",
+                    &python_path.to_string_lossy(),
+                ])
+                .output()
+                .await
+                .context("failed to run uv pip install ipykernel")?
+        } else {
+            util::command::new_command(python_path.to_string_lossy().as_ref())
+                .args(&["-m", "pip", "install", "ipykernel"])
+                .output()
+                .await
+                .context("failed to run pip install ipykernel")?
+        };
 
         if output.status.success() {
             anyhow::Ok(())
@@ -146,6 +164,11 @@ pub fn install_ipykernel_and_assign(
 
                 window_handle
                     .update(cx, |_, window, cx| {
+                        let store = ReplStore::global(cx);
+                        store.update(cx, |store, _cx| {
+                            store.mark_ipykernel_installed(&env_spec);
+                        });
+
                         let updated_spec =
                             KernelSpecification::PythonEnv(PythonEnvKernelSpecification {
                                 has_ipykernel: true,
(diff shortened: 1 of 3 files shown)


crates/repl/src/repl_store.rs
```diff
@@ -136,6 +136,19 @@ impl ReplStore {
         cx.notify();
     }
 
+    pub fn mark_ipykernel_installed(&mut self, cx: &mut Context<Self>, spec: &PythonEnvKernelSpecification) {
```
miguelraz: What is the problem this function is solving, and why did adding the `uv` install path make this necessary?

crates/repl/src/repl_editor.rs
```diff
@@ -109,11 +110,25 @@ pub fn install_ipykernel_and_assign(
...
-            .output()
-            .await
-            .context("failed to run pip install ipykernel")?;
+        let output = if is_uv {
+            util::command::new_command("uv")
+                .args(&[
+                    "pip",
+                    "install",
+                    "ipykernel",
+                    "--python",
```
miguelraz, line 119: Why is the `--python` flag here necessary? I'm sorry, but I'm not too familiar with `uv` tooling flag details.

feitreim: I'm not *100%* sure its necessary but its a nice extra way to ensure you're calling it on the correct python installation and not like a the global python install or something. 

this should make it slightly more robust to things like trying to use a kernel with a uv managed venv that isn't the one in your project directory.

crates/repl/src/repl_store.rs
feitreim: Oops I should have documented this better. This is admittedly a secondary bug that I just found while I was working on the uv thing, heres demo of the behavior before and after this change:

the actual behavior it fixes it that after installing the ipykernel, the repl isnt made aware of that, and it doesnt update the display, so you can just click on it again and it will 'reinstall' the kernel. worth noting that it doesn't actually effect the functionality, after the first install the kernel will work and you can run cells, it just isnt updating the UI properly.

Behavior Before:
https://github.com/user-attachments/assets/ec1134e5-637b-42cf-8e93-81fc4512de7f

Behavior After:
https://github.com/user-attachments/assets/7359f35e-0fb8-4fee-8160-165673ec219f

crates/repl/src/repl_store.rs
```diff
@@ -136,6 +136,23 @@ impl ReplStore {
...
+                if let KernelSpecification::PythonEnv(env_spec) = kernel_spec {
+                    if env_spec == spec {
+                        env_spec.has_ipykernel = true;
+                    }
+                }
+            }
+        }
+        cx.notify();
+    }
+
```
probably-neb, line 155: Why is this necessary? I'm having a hard time understanding what this code is doing - how did this work before?. On the surface it seems like an unrelated bug-fix?

feitreim: @probably-neb Yes sorry, miguel also commented on this and I replied to him but that maybe isn't super visible? 

> Oops I should have documented this better. This is admittedly a secondary bug that I just found while I was working on the uv thing, heres demo of the behavior before and after this change:
> 
> the actual behavior it fixes it that after installing the ipykernel, the repl isnt made aware of that, and it doesnt update the display, so you can just click on it again and it will 'reinstall' the kernel. worth noting that it doesn't actually effect the functionality, after the first install the kernel will work and you can run cells, it just isnt updating the UI properly.
> 
> Behavior Before:
> https://github.com/user-attachments/assets/ec1134e5-637b-42cf-8e93-81fc4512de7f
> 
> Behavior After:
> https://github.com/user-attachments/assets/7359f35e-0fb8-4fee-8160-165673ec219f

I could totally make that into a seperate PR though.

probably-neb: No need. Makes sense

PR #51898, opened 2026-03-19
title_bar: Prefer most specific repository for nested git repos

### LINKED ISSUE #7566: Show submodule branch on the GUI
### Summary

When we edit a file under a submodule, it should show the submodule branch at the top (or some visual way to identify it).

Currently you can only see the branch of the current project root (in the branch picker control in the menu bar) not of any submodule you may happen to inside when viewing files.

Caio-Ze:
## Context

I have a project with a nested git repo inside a parent repo (separate `.git` directory, not a submodule). The title bar shows the parent's branch instead of the nested repo's own branch.

The issue is in `get_repository_for_worktree()` — it iterates `git_store.repositories()` (an `FxHashMap`) and returns the first `starts_with` match. Both the parent and nested repo match, and whichever one FxHashMap iterates first wins. There's no reason it should be one or the other.

The fix already exists elsewhere in the codebase — `GitStore::repository_and_path_for_project_path()` at `git_store.rs:1826` uses `.max_by_key()` to pick the most specific (longest path) match. This PR applies the same approach to three functions that have the same problem:

- `TitleBar::get_repository_for_worktree()` — branch display in title bar
- `resolve_active_repository()` in `git_ui` — repository selection for the git panel
- `get_branch_for_worktree()` in `recent_projects` — branch display in the project switcher

Two other locations use a similar `starts_with` pattern (`effective_active_worktree()` in `title_bar.rs` and worktree selection in `recent_projects.rs`) but those iterate worktrees against a single known repo, not repos against a worktree — so first-match and longest-match give the same result. Left those unchanged.

Closes #7566

## How to Review

All three changes are the same transformation: first-match loop (or `.find()`) → `.filter().max_by_key()` on path length. The reference is at `crates/project/src/git_store.rs:1826`.

The primary fix is `get_repository_for_worktree()` in `title_bar.rs`. The other two are the same pattern.

One difference from the reference: I used `.as_os_str().len()` instead of `.clone()` for the `max_by_key` key — avoids cloning an `Arc<Path>` per comparison. Among prefix-related paths (which is all that passes the filter), the longer path is always the more specific match, so length comparison is equivalent.

`title_bar` has no existing test infrastructure. Happy to add a test if you'd like — the setup would follow the pattern in `test_git_traversal_with_nested_repos` (`crates/project/tests/integration/git_store.rs`).

```
cargo test -p title_bar -p git_ui -p recent_projects
./script/clippy
```

## Self-Review Checklist

Release Notes:

- Fixed branch picker showing parent repository's branch instead of the nested repository's branch when working in submodules or nested git repos.


osiewicz (approved):

smitbarmase (approved):
Thanks!

PR #51911, opened 2026-03-19
editor: Place cursor at clicked column when clicking sticky headers

Veykril:
When clicking on a sticky header line, the cursor is now placed at the column corresponding to the click position, rather than always at the start of the outline item's range. The hover cursor is also changed from a pointing hand to an I-beam to reflect that clicking behaves like clicking in normal editor text.

Release Notes:

- Clicking a sticky header now puts the cursor at the clicked column


cole-miller (approved):
Using this feels a little bit weird to me because you can't click and drag to select even though there's an I-beam, but also seems reasonable and implementation looks good

PR #51919, opened 2026-03-19
Add automatic-volume to legacy

yara-blue:
## Context

This allows using of Rodio's effects library within our home brewn audio pipeline. The alternative would be inlining Rodio's effects which is problematic from a legal stance. We would then have to make clear that code is not owned by zed-industries while the code would be surrounded by zed-industries owned code.

This adaptor does incur a slight performance penalty (copying into a pre-allocated vec and back) however the impact will be immeasurably low.

There is no latency impact.

## How to Review

- Adds an adapter for Rodio effects
- Enables the adapter and effects only when the setting is enabled
-Makes the setting pub(crate) so we can use it from livekit playback

## Self-Review Checklist

Release Notes:

- Added Automatic volume control to calls

### DIFF
--- a/crates/audio/src/audio.rs
+++ b/crates/audio/src/audio.rs
@@ -8,6 +8,7 @@ pub const CHANNEL_COUNT: ChannelCount = nz!(2);
 
 mod audio_settings;
 pub use audio_settings::AudioSettings;
+pub use audio_settings::LIVE_SETTINGS;
 
 mod audio_pipeline;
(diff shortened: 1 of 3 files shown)


yara-blue:
Note to tests:
No good automated testing infrastructure exists within the legacy pipeline. The effect can be disabled with a setting. This is trivial and has been manually tested and found to work.

kubkon (approved):
Looks great!

PR #51945, opened 2026-03-19
audio: Remove rodio audio backend

osiewicz:
Co-authored-by: Smit Barmase <heysmitbarmase@gmail.com>

We've decided to remove rodio backend for audio as we didn't have time to get it to a nice spot. I personally believe we should eventually re-land it (the composable pipelines are quite nice!), but given that we need audio to work, this seems like the best way forward. We won't have to worry about all the ways in which the legacy pipeline interoped with rodio's.

## Self-Review Checklist

Release Notes:

- audio: Experimental rodio audio backend has been removed.


crates/livekit_client/src/livekit_client/playback.rs
```diff
@@ -170,33 +136,18 @@ impl AudioStack {
...
-                AudioSourceOptions::default(),
-                SAMPLE_RATE.get(),
-                CHANNEL_COUNT.get().into(),
-                10,
-            )
-        };
+        let source = NativeAudioSource::new(
+            // n.b. this struct's options are always ignored, noise cancellation is provided by apm.
+            AudioSourceOptions::default(),
+            SAMPLE_RATE.get(), // TODO(audio): this was legacy params,
```
yara-blue: This `TODO` is no longer relevant. With the removal of the experimental audio pipeline we'll not be able to switch to more sensible audio params (currently we are sending two channel audio while nearly all microphones are mono). So we can take out this `TODO` and any mention of legacy params.

crates/livekit_client/src/livekit_client/playback.rs
```diff
@@ -170,33 +136,18 @@ impl AudioStack {
...
+            // removed for now for simplicity
```
yara-blue: same for this comment

crates/livekit_client/src/livekit_client/playback.rs
```diff
@@ -546,40 +483,6 @@ impl rodio::Source for RodioEffectsAdaptor {
 pub struct Speaker {
```
yara-blue, line 482: In principle we can take this out as well. It was used to add metadata to replay's, and to prevent recording non-staff users. Now that we have removed the replay system we do not need this info anymore.

yara-blue: On the other hand it can be nice for future debugging if the live-kit tracks have sensible names. Ok with leaving it in.

yara-blue (approved):
Looks good to me
(took out two todo's that where left from the experimental audio pipeline effort)

PR #51959, opened 2026-03-19
Auto-resize panel to fit content on border double-click

jessedelira:
## Summary

- Double-clicking the project panel's resize border now auto-sizes the panel to fit the widest visible entry, eliminating horizontal scrolling
- Adds a `content_width` method to the `Panel` trait so any panel can opt into auto-resize behavior
- Adds an `auto_resize_on_double_click` setting (default: `true`) to the project panel; when set to `false`, the previous static-reset behavior is preserved

## Test plan

- [ ] Open a project with deeply nested files so the project panel shows a horizontal scrollbar
- [ ] Double-click the panel border — panel should resize to exactly fit the widest entry (no horizontal scrollbar)
- [ ] Manually drag the panel wider than the content, then double-click — panel should shrink back to fit
- [ ] Set `"project_panel": { "auto_resize_on_double_click": false }` in settings and double-click — panel should reset to `default_width` (old behavior)
- [ ] Verify other dock panels (outline, git, etc.) still reset to default on double-click

Release Notes:

- Added auto-resize on double-click for the project panel border, fitting the panel width to the longest visible file entry. Controlled by the `auto_resize_on_double_click` setting (default: `true`).


dinocosta (commented):
Hey @jessedelira 👋 

Thank you for suggesting these changes. Seems like a nice thing to support fit to size behavior in some of Zed's panels!

It appears that the base branch's state has shifted quite significantly from this one when I started reviewing the changes, as in the meantime we added support for panels to opt into flexible width. As such, I merged `main` into this branch but refactored things a little bit.

I'm going to ask Danilo for input on this since I believe he worked on the flexible width support and might have some ideas or opinions on whether this approach is the best for what we're trying to achieve.

Thanks! 🙂

jessedelira:
@dinocosta bumping to see if there is still a review in the works, didn't see required reviewers update.

Thank you

dinocosta:
Hey @jessedelira! Thank you for your patience, and apologies for the delayed response! 🙇

While fitting a panel to its content could be useful, we’re not currently convinced that it should replace the existing double-click behavior of restoring the panel’s default size. Fitting to content can also be somewhat ambiguous for a dynamic tree, it could be based on the currently visible entries or the entire worktree, and a single deeply nested or unusually long entry could result in an impractically wide panel. With panels also supporting flexible widths, we're not sure that the additional behavior and implementation complexity are justified at this point.

Given that, I’m going to close this PR for now. If you feel strongly that fit-to-content is something Zed should support, you’re welcome to open a [GitHub Discussion](https://github.com/zed-industries/zed/discussions) describing the use cases and proposed interaction. It could, for example, make more sense as a command in the command palette rather than replacing the double-click behavior on the panel’s resize border. A discussion would be the best place to gather feedback, see if there's enough user interest, validate the use cases, and establish whether this is a direction we want to pursue before revisiting the implementation.

We really appreciate the work you put into exploring this. Thank you again! 🙂

PR #51974, opened 2026-03-19
editor: Preserve initiating selection in select all matches

loadingalias:
Closes #32894

## Summary
 • preserve the initiating selection as the newest selection in `select_all_matches`
 • keep `Autoscroll::fit()` semantics unchanged and fix the bug at the selection ordering source instead
 • avoid unnecessary API expansion, editor state, or broader autoscroll changes
 • add regression coverage for both forward and reversed initiating selections when editing after `Select All Matches`

## Verification
 • `cargo fmt --all -- --check`
 • `./script/clippy -p editor`
 • `cargo test -p editor select_all_matches_does_not_scroll`
 • `cargo test -p editor select_all_matches_edit_keeps_scroll_on_initiating_selection`
 • `cargo test -p editor select_all_matches_edit_keeps_scroll_on_reversed_initiating_selection`

## Manual
 • Repro'd the issue with the caret on the middle match, ran `Select All Matches`, typed, and confirmed the viewport stayed anchored to the initiating match instead of jumping to the last occurrence.
 • Repeated with a reversed selection over the initiating match and confirmed the same behavior.
 • Confirmed the existing caret only `wordwise = true` behavior is unchanged: substring matches like `efabc` are excluded unless `abc` is explicitly selected first.

Release Notes:
 • Fixed `Select All Matches` so editing stays anchored to the initiating match instead of scrolling to the last occurrence.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -16088,10 +16088,13 @@ impl Editor {
 
         let mut new_selections = Vec::new();
 
-        let reversed = self
-            .selections
-            .oldest::<MultiBufferOffset>(&display_map)
-            .reversed;
+        let initial_selection = self.selections.oldest::<MultiBufferOffset>(&display_map);
+        let initial_range = if initial_selection.reversed {
+            initial_selection.end..initial_selection.start
+        } else {
+            initial_selection.start..initial_selection.end
+        };
+        let reversed = initial_selection.reversed;
         let buffer = display_map.buffer_snapshot();
         let query_matches = select_next_state
             .query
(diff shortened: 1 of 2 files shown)


dinocosta (commented):
Hey @loadingalias ! Thank you for attempting to fix this issue 🙌 

I Was taking a look at the original issue and just noticed that https://github.com/zed-industries/zed/pull/49232 was opened about a month earlier with pretty much the same solution, so I'll have to close this one and merge that one instead, as I feel it would be a bit unfair for the author of that Pull Request for us to merge this one.

Sorry it took me a while to get to this and spot the overlap!

Issue #51905, opened 2026-03-19
Inline Assistant Not Working Correctly, inserts spurious `\`


90aca:
### Reproduction steps

1. Open Inline Assistant (CTRL+Enter)
2. Try to execute it
3. Wrong/No results

### Current vs. Expected behavior

Error with GitHub Copilot (Gemini 3 Pro (Preview)):

<img width="968" height="286" alt="Image" src="https://github.com/user-attachments/assets/bbc5f90c-0843-4104-aee7-b998e14b2b29" />

Error with GitHub Copilot Claude Sonnet 4.5: It adds "\\" characters

<img width="970" height="199" alt="Image" src="https://github.com/user-attachments/assets/0e18ceee-c475-41fd-bf73-ba5b7582a2ea" />

Error with OpenRouter Gemini 3 Flash Preview: It does not change anything whatsoever

<img width="915" height="328" alt="Image" src="https://github.com/user-attachments/assets/9fb9ff5a-a3e8-4819-a73b-038056f735e5" />

---

It should work like in previous versions, like it always did.

### Zed version and system specs

Zed: v0.229.0+preview.204.433d9299a3fad19eeaa023d6a9d33f551797f56e (Zed Preview) 
OS: macOS 26.4 (Build 25E5233c)
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
2026-03-19T08:30:43+01:00 WARN  [agent_ui::buffer_codegen] Unexpected event: ReasoningDetails(Array [Object {"type": String("reasoning.encrypted"), "data": String("CiEBjz1rX9H2pIcSupqGrcW/CqSFUupWMEyIWvqYtN1yDGw="), "format": String("google-gemini-v1"), "id": String("tool_rewrite_section_n2dZt6yNGZV3Uo3TTzQK"), "index": Number(0)}])
2026-03-19T08:31:55+01:00 WARN  [agent_ui::buffer_codegen] Unexpected event: ReasoningDetails(Array [Object {"type": String("reasoning.encrypted"), "data": String("CiEBjz1rX1QHaSrSMEGVFOZmrRxxUSNdxDn+ADm8rMxSZ34="), "format": String("google-gemini-v1"), "id": String("tool_failure_message_8YfoQmWOJxuJ1UJ1C4xq"), "index": Number(0)}])
```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json
{
  "agent": {
[...]

PR #51976, opened 2026-03-19
language_models: Fix the partial json streaming to not blast `\` everywhere

feitreim:
## Context

This PR fixes one of the issues in #51905, where model outputs are full of errant `\` characters. 

heres the problem: As the response is streamed back to zed, we accumulate the message chunks and and need to convert those chunks to valid json, to do that we use `partial_json_fixer::fix_json`, when the last character of a chunk is `\`, the `fix_json` has to escape that backslash, because its inside of a string (if it isn't, its invalid json and the tool call will crash) and other wise you would end up escaping the end `"` and everything would be messed up.

why is this a problem for zed:
T_0 is the output at some step.
T_1 is the output at the next step.

the `fix_json` system is meant to be used by replacing T_0 with T_1, however in the editor, replacing the entirety of T_0 with T_1 would be slow/cause flickering/etc.. so we calculate the difference between T_0 and T_1 and just add it to the current buffer state. So when a chunk ends on `\`, we end up with something like  `... end of line\\"}` at the end of T_0,
in T_1, this becomes `... end of line\n ...`. then when we add the new chunk from T_1, it just picks up after the \n because its tracking the length to manage the deltas.

## How to Review

utils.rs:
fix_streamed_json => remove trailing backslashes from incoming json streams so that `partial_json_fixer::fix_json` doesn't try to escape them.
other files: call fix_streamed_json before passing to `serde_json`

I had claude write a bunch of tests while I was working on the fix, which I have kept in for now, but the end functionality of fix_streamed_json is really simple now, so maybe these arent really needed.

## Videos
Behavior Before:

https://github.com/user-attachments/assets/f23f5579-b2e1-4d71-9e24-f15ea831de52

Behavior After:

https://github.com/user-attachments/assets/40acdc23-4522-4621-be28-895965f4f262

## Self-Review Checklist

Release Notes:

- language_models: fixed partial json streaming

### DIFF
--- a/crates/language_models/Cargo.toml
+++ b/crates/language_models/Cargo.toml
@@ -67,7 +67,6 @@ vercel = { workspace = true, features = ["schemars"] }
 x_ai = { workspace = true, features = ["schemars"] }
 
 [dev-dependencies]
-
 language_model = { workspace = true, features = ["test-support"] }
 pretty_assertions.workspace = true
 
--- a/crates/language_models/src/provider/anthropic.rs
+++ b/crates/language_models/src/provider/anthropic.rs
@@ -24,7 +24,7 @@ use ui::{ButtonLink, ConfiguredApiCard, List, ListBulletItem, prelude::*};
 use ui_input::InputField;
 use util::ResultExt;
 
-use crate::provider::util::parse_tool_arguments;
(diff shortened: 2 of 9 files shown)


Veykril (approved):
Thanks!

PR #51983, opened 2026-03-20
extensions_ui: Remove upsell banners for Bash

lingyaochu:
## Context

Closes #51917

Zed's built-in support only covers Tree-sitter highlighting for Bash, with no language server included. An extension is still required to get full language support, so we should not show banners to users who want to install an extension for it, because this is truly needed.

## How to Review

Removed the `LanguageBash` related code blocks in `crates/extensions_ui/src/extensions_ui.rs`.

## Self-Review Checklist

Release Notes:

### DIFF
--- a/crates/extensions_ui/src/extensions_ui.rs
+++ b/crates/extensions_ui/src/extensions_ui.rs
@@ -232,7 +232,6 @@ enum Feature {
     ExtensionTailwind,
     ExtensionTy,
     Git,
-    LanguageBash,
     LanguageC,
     LanguageCpp,
     LanguageGo,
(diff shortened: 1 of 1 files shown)


MrSubidubi (changes requested):
Thanks for this!

While I see the linked discussion, we could also discuss the possibility of deprecating the extension in favor of also having builtin language server support for Bash - this might just be the more intuitive approach here, given that we support the highlighting out of the box (and the extension is fairly minimalistic at that).

Curios about @ConradIrwin s opinion on this here though too, might also ask some more folks internally on this.

PR #51985, opened 2026-03-20
git_ui: Fix multibuffer coordinate conversion in clipboard diff

OmChillure:
## Context

Fixes the visual selection update in `TextDiffView::open` to properly convert between buffer-local and multibuffer coordinates using `buffer_point_to_anchor`. Previously, raw multibuffer Points were used directly for line expansion, which produced incorrect regions when expanded deleted diff hunks shifted multibuffer row numbers.

This provides a single unified code path for both singleton and non-singleton multibuffers, as suggested in [#51457 review feedback](https://github.com/zed-industries/zed/pull/51457#issuecomment-4091134303).

Follow-up to #51457.

## How to Review

Small PR - all changes are in `crates/git_ui/src/text_diff_view.rs`. Focus on:

- `open()`: The visual selection update block now uses `buffer_point_to_anchor` + `to_point` instead of raw multibuffer coordinate math
- No more assumption that multibuffer Points == buffer-local Points
- Existing tests validate both singleton and non-singleton multibuffer paths

## Self-Review Checklist

Release Notes:

- Fixed incorrect diff region when using "Diff Clipboard with Selection" with expanded diff hunks in the editor.

### DIFF
--- a/crates/git_ui/src/text_diff_view.rs
+++ b/crates/git_ui/src/text_diff_view.rs
@@ -47,11 +47,24 @@ impl TextDiffView {
         let source_editor = diff_data.editor.clone();
 
         let selection_data = source_editor.update(cx, |editor, cx| {
-            let multibuffer = editor.buffer().read(cx);
-            let source_buffer = multibuffer.as_singleton()?;
+            let multibuffer = editor.buffer();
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #51991, opened 2026-03-20
Reload remote buffers on external file changes with unchanged metadata

cam-br0wn:
When an external tool modifies a file over SSH, the file watcher reports PathChange::Updated but filesystem timestamp granularity can cause the file metadata (mtime, size) to appear unchanged. Previously, the buffer store would short-circuit and never reload the buffer.

Emit BufferEvent::ReloadNeeded when the watcher reports an update for a clean buffer whose metadata is unchanged. Handle this event in BufferStore::on_buffer_event for non-collab downstream clients to trigger an async reload from disk.

## Context

On SSH/NFS filesystems with coarse mtime granularity (often 1-second resolution), an external tool like a formatter or code generator can write new content where the mtime lands in the same tick and the byte length doesn't change. The file watcher still picks up the inode change and reports `PathChange::Updated`, but `local_worktree_entry_changed` compares the old and new `File` structs, finds them equal, and bails out without reloading. The buffer just shows stale content.

This PR threads `PathChange` into `local_worktree_entry_changed` so it can tell the difference between a real watcher update and a no-op. When the path change is `Updated` or `AddedOrUpdated`, the metadata matches, and the buffer is clean, it emits `BufferEvent::ReloadNeeded`. The handler in `BufferStore::on_buffer_event` picks this up for non-collab downstream clients and triggers an async reload from disk. Dirty buffers are left alone so we don't blow away user edits.

## How to Review

Two files:

1. **`crates/project/src/buffer_store.rs`** - The core change. The `force_reload_needed_for_updated_path` flag in `local_worktree_entry_changed` (lines 555-561) and the `ReloadNeeded` emission (lines 606-612) are the interesting parts. Also check the new `BufferEvent::ReloadNeeded` arm in `on_buffer_event` (lines 1163-1169), specifically the `!downstream_client.is_via_collab()` guard.
2. **`crates/project/tests/integration/project_tests.rs`** - Two new tests. `test_clean_buffer_reloads_on_unchanged_metadata` covers the happy path (clean buffer reloads), `test_dirty_buffer_skips_reload_on_unchanged_metadata` checks that dirty buffers are left alone. Both freeze the mtime with `set_next_mtime` and use `atomic_write` to simulate an external tool.

## Self-Review Checklist

Release Notes:

- Fixed remote buffers not refreshing when an external tool modifies a file over SSH and the filesystem metadata (mtime, size) appears unchanged.

### DIFF
--- a/crates/project/src/buffer_store.rs
+++ b/crates/project/src/buffer_store.rs
@@ -464,12 +464,13 @@ impl LocalBufferStore {
         cx: &mut Context<BufferStore>,
     ) {
         let snapshot = worktree_handle.read(cx).snapshot();
-        for (path, entry_id, _) in changes {
+        for (path, entry_id, path_change) in changes {
             Self::local_worktree_entry_changed(
                 this,
                 *entry_id,
                 path,
                 worktree_handle,
+                *path_change,
                 &snapshot,
                 cx,
             );
@@ -481,6 +482,7 @@ impl LocalBufferStore {
         entry_id: ProjectEntryId,
         path: &Arc<RelPath>,
         worktree: &Entity<worktree::Worktree>,
+        path_change: PathChange,
         snapshot: &worktree::Snapshot,
         cx: &mut Context<BufferStore>,
     ) -> Option<()> {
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore:

image

PR #51992, opened 2026-03-20
gpui_macos: Guard deferred AppKit calls against closed windows

Veykril:
MacWindow methods like activate(), zoom(), toggle_fullscreen(), resize(), prompt(), and titlebar_double_click() capture the raw native_window pointer and spawn detached async tasks on the foreground executor. If the window is closed between the spawn and execution (e.g. the user clicks the close button), the task sends AppKit messages to a closed NSWindow, which throws an ObjC exception in _changeKeyAndMainLimitedOK: and aborts the process.

Add an isVisible check inside each deferred task before sending messages to the native window. After [NSWindow close], isVisible returns NO, so the guard prevents the crash. Both the check and the subsequent AppKit call execute synchronously within the same main-thread task, so there is no TOCTOU race.

For prompt(), the else branch releases the alert object to avoid leaking it; the oneshot sender inside the completion block is dropped, which cancels the channel and propagates as an error to the caller.

Closes ZED-5TN

Release Notes:

### DIFF
--- a/crates/gpui_macos/src/window.rs
+++ b/crates/gpui_macos/src/window.rs
@@ -55,7 +55,10 @@ use std::{
     path::PathBuf,
     ptr::{self, NonNull},
     rc::Rc,
-    sync::{Arc, Weak},
+    sync::{
+        Arc, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
     time::Duration,
 };
 use util::ResultExt;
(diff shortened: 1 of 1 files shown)


osiewicz (approved):
It'd be nice to have some kind of helper for it (right now it's gonna be pretty easy to forget to check the closed flag and while the helper doesn't help with that, it at least codifies the pattern + lets us document the "why").

Sans that, this looks good to me.

PR #51996, opened 2026-03-20
gpui_macos: Fix stale atlas key causing crash on None texture lookup

Veykril:
MetalAtlas::remove() used tiles_by_key.get(key) to look up the texture ID but only called tiles_by_key.remove(key) when the texture became fully unreferenced. When a tile was removed while other tiles remained on the same texture, the key stayed in tiles_by_key as a stale entry.

Once subsequent removals deleted the texture, get_or_insert_with could return the stale tile referencing a now-deleted texture slot, causing an unwrap panic in MetalAtlasState::texture().

Fix: change .get(key) to .remove(key) unconditionally, matching the WGPU and DirectX atlas implementations which already do this correctly.

Closes ZED-5KV

Release Notes:

### DIFF
--- a/crates/gpui_macos/src/metal_atlas.rs
+++ b/crates/gpui_macos/src/metal_atlas.rs
@@ -61,7 +61,7 @@ impl PlatformAtlas for MetalAtlas {
 
     fn remove(&self, key: &AtlasKey) {
         let mut lock = self.0.lock();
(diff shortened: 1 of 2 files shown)


gobijan:
Just stumbled upon this as I was using that https://crates.io/crates/gpui 0.2.2 that doesn't have this already patched in :(

PR #52002, opened 2026-03-20
acp: Add agent websites to the registry page

benbrandt:
## Context

The registry now distinguishes between websites and repos, so we can show either or both if available.

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_registry_ui.rs
+++ b/crates/agent_ui/src/agent_registry_ui.rs
@@ -404,16 +404,15 @@ impl AgentRegistryPage {
         });
 
         let website_button = agent.website().map(|website| {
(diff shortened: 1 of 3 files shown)


crates/agent_ui/src/agent_registry_ui.rs
```diff
@@ -403,6 +403,23 @@ impl AgentRegistryPage {
             })
         });
 
+        let website_button = agent.website().map(|website| {
+            let website_for_tooltip: SharedString = website.to_string().into();
+            let website_for_click = website.to_string();
```
bennetbo: NIT: We could probably reuse the shared string from above, and move these clones into the closures instead and use shadowing here

bennetbo (approved):

PR #52003, opened 2026-03-20
Make paste a separate undo transaction from preceding edits

Veykril:
When undoing a paste, it is really confusing when that actually also removes what was type right before the paste if the paste happened  fast enough after.

Release Notes:

- Fixed undoing a paste sometimes also undoing edits right before the paste

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -14018,6 +14018,8 @@ impl Editor {
             return;
         }
 
+        self.finalize_last_transaction(cx);
+
         let clipboard_text = Cow::Borrowed(text.as_str());
 
         self.transact(window, cx, |this, window, cx| {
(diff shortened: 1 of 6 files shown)


dinocosta (commented):
Hey @Veykril, sorry for being late reviewing this! I’d gotten used to being assigned to ‎`staff` PRs without having to explicitly approve them 😅

The change makes sense to me, but I’ve confirmed that the test still passes if the ‎`self.finalize_last_transaction` call is commented out. Looking into this a bit, it seems related to ‎`text::History::group_interval`, which is set to ‎`0` in tests, meaning that no edits are ever grouped. As a result, it seems that we never end up, in tests, in a situation where undoing would remove both the paste and the inserted text. We probably need to call ‎`Buffer::set_group_interval` with a sensible value that will actually exercise this test.

Let me know what you think. Thanks!

dinocosta (approved):

PR #52005, opened 2026-03-20
acp: Set agent server cwd from project paths

benbrandt:
## Context

Add Project::default_path_list and reuse it for agent session work
directories and ACP server startup so agent processes start in the
project context by default

We previously removed all cwd from this hoping to have a global process
shared, but it doesn't work for remote projects. Since we're at the
project boundary anyway, we might as well start it up in a similar spot
as a new thread.

## Self-Review Checklist

Release Notes:

- acp: Make sure the agent server is started in a project directory

### DIFF
--- a/crates/agent_servers/src/acp.rs
+++ b/crates/agent_servers/src/acp.rs
@@ -166,6 +166,7 @@ impl AgentSessionList for AcpSessionList {
 
 pub async fn connect(
     agent_id: AgentId,
+    project: Entity<Project>,
     display_name: SharedString,
     command: AgentServerCommand,
     default_mode: Option<acp::SessionModeId>,
@@ -175,6 +176,7 @@ pub async fn connect(
 ) -> Result<Rc<dyn AgentConnection>> {
     let conn = AcpConnection::stdio(
         agent_id,
+        project,
         display_name,
         command.clone(),
         default_mode,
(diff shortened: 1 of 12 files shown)


crates/agent_servers/src/acp.rs
```diff
@@ -203,6 +206,15 @@ impl AcpConnection {
         let mut child =
             builder.build_std_command(Some(command.path.display().to_string()), &command.args);
         child.envs(command.env.iter().flatten());
+        if let Some(cwd) = project.update(cx, |project, cx| {
+            project
+                .default_path_list(cx)
+                .ordered_paths()
+                .next()
+                .cloned()
```
benbrandt, line 214: Same as we do for new session for now

crates/project/src/project.rs
```diff
@@ -2286,6 +2287,32 @@ impl Project {
...
+                }
+            })
+            .collect::<Vec<_>>();
+
+        if worktree_roots.is_empty() {
+            PathList::new(&[paths::home_dir().as_path()])
+        } else {
+            PathList::new(&worktree_roots)
+        }
+    }
```
benbrandt, line 2314: Pulled the central logic here

bennetbo (approved):

crates/agent_servers/src/acp.rs
```diff
@@ -203,6 +206,15 @@ impl AcpConnection {
...
+        }) {
+            child.current_dir(cwd);
```
maan2003, line 216: @benbrandt this might have broken acp on remote projects.


```
failed to spawn command cd "/home/maan2003/src/codex" && "fish" "-c" "ssh -o 'ControlMaster=no' -o 'ControlPath=/tmp/zed-ssh-sessionGIul5x/ssh.sock' -q -T devbox \"cd \\\"/home/maan2003\\\" && exec env 'SSH_AUTH_SOCK=/..<bunch of env variables>' codex-acp\"" 
```
it tries to cd locally for a remote project, which doesn't work because path doesn't exist locally.

PR #52008, opened 2026-03-20
markdown_preview: Refactor to use shared markdown crate

smitbarmase:
We now use the parser and renderer from the `markdown` crate for Markdown Preview, instead of maintaining two separate code paths.

How it works:

`markdown_preview_view.rs` is now a consumer of `MarkdownElement`. It acts as a thin wrapper, handling things like resolving URL clicks and image URLs, which can vary between consumers. It also handles syncing the editor selection with the active block in the preview. The APIs for this are provided by `MarkdownElement`.

All the heavy lifting like parsing HTML, rendering block markers on hover, handling the active block, etc. is done by `MarkdownElement`. Everything is opt-in. For example, markdown in the Agent Panel can choose not to enable block marker rendering or HTML parsing, while Markdown Preview opts into those features.

Final outcome:

For Markdown Preview View:

- Added:
  - Selection support in the preview

- Stays:
  - Syncing between editor and preview
  - Autoscroll
  - Hover and active block markers
  - Checkbox toggling
  - Image rendering
  - Mermaid rendering

For the `markdown` crate:

- No changes for existing consumers like the Agent Panel
- Consumers can now opt into:
  - HTML rendering
  - Block marker rendering
  - Click event handling
  - Custom image resolvers
  - Mermaid rendering

Release Notes:

- N/A


osiewicz (approved):

luapmartin:
@smitbarmase thanks a lot, also wondering how hard it would be adding markdown preview mode search (Find) support too?

PR #52012, opened 2026-03-20
fs: Return trashed file location

### LINKED ISSUE #5039: Cannot undo actions in the project panel

dinocosta:
## Context

These changes are meant to eventually be used by the Project Panel's Undo/Redo system, which was started on https://github.com/zed-industries/zed/pull/47091 .

In order to be able to undo file trashing and file creation operations, we'll want to be able to track where the file is trashed to such that, reverting these operations is simply a matter of moving the file from the OS' trash back to the original location.

Note that this Pull Request will remain in Draft until the corresponding Pull Request in the our `trash-rs` fork, https://github.com/zed-industries/trash-rs/pull/1 and https://github.com/zed-industries/trash-rs/pull/2, are reviewed and merged.

## How to Review

Since most of the work is handled by the `trash-rs` crate, the most relevant changes in this Pull Request are:

* Add `fs::Fs::TrashedEntry` struct, which allows us to track the original file path and the new path in the OS' trash
* Update the `fs::Fs::trash_dir` and `fs::Fs::trash_file` signatures to now return `Result<TrashedEntry>` instead of `Result<()>`
    * The `options` argument was removed because it was never used by implementations other than the default one, and with this change to the signature type, we no longer have a default implementation, so the `options` argument would no longer make sense
* Update `fs::RealFs::trash_dir` and `fs::RealFs::trash_file` implementations to simply delegate to `trash-rs` and convert the result to a `TrashedEntry`
* Add `fs::FakeFs::trash` so we can simulate the OS' trash during tests that touch the filesystem
* Add `fs::FakeFs::trash_file` implementation to leverage `fs::FakeFs::trash`
* Add `fs::FakeFs::trash_dir` implementation to leverage `fs::FakeFs::trash`

## Self-Review Checklist

Release Notes:

- N/A


crates/fs/src/fs.rs
```diff
@@ -164,6 +176,31 @@ pub trait Fs: Send + Sync {
     }
 }
 
+/// Represents a file or directory that was moved to the system's trash.
```
dinocosta: Document why we want this instead of just using `trash::TrashItem` as a return type in `trash_file`/`trash_dir` .

dinocosta: Updated with https://github.com/zed-industries/zed/pull/52012/commits/ebfb3552462c4ce8f95a3e14b8cb69a5f8cbc044 ✅

crates/fs/Cargo.toml
```diff
@@ -41,10 +41,7 @@ time.workspace = true
 util.workspace = true
 is_executable = "1.0.5"
 notify = "8.2.0"
-
-[target.'cfg(target_os = "macos")'.dependencies]
-objc.workspace = true
-cocoa = "0.26"
+trash = { git = "https://github.com/dinocosta/trash-rs", branch = "5039-macos-restore" }
```
yara-blue: should be a zed-owned fork.

yara-blue (changes requested):

crates/fs/Cargo.toml
dinocosta: Updated on https://github.com/zed-industries/zed/pull/52012/commits/500e08c2fb6394dcb8ed09da907d92cacd5eaaf3 ✅

yara-blue (approved):

Koshkaj:
this is merged but has it been released? i am hitting cmd+z expecting vscode like behavior where the file gets restored if its deleted but nothing happens. Im on version 1.6.3

CC: @JosephTLyons

dinocosta:
> this is merged but has it been released? i am hitting cmd+z expecting vscode like behavior where the file gets restored if its deleted but nothing happens. Im on version 1.6.3
> 
> CC: @JosephTLyons

Sorry for the confusion here! Even though these changes have already been merged, support for undo/redo is still behind a feature flag while we finish wrapping up support for remote sessions.

I’ve also just noticed that merging these changes ended up closing this issue – https://github.com/zed-industries/zed/issues/5039 – which it shouldn’t have 🤦

I'll be re-opening that issue and doing a better job of communicating there. Once again, sorry for the confusion!

Koshkaj:
> Sorry for the confusion here! Even though these changes have already been merged, support for undo/redo is still behind a feature flag while we finish wrapping up support for remote sessions.

Is there an ETA on this feature? Perhaps you could release it for the local FS first and then iterate over it? I am sure that a lot of people are waiting for this change since it's super important for everyone's workflow

PR #52014, opened 2026-03-20
fs: Add support for restoring trashed files

dinocosta:
## Context

With #52012 adding support for tracking the trashed file's location, we can now easily leverage that information in order to restore the file back to its previous location, as `fs::TrashedEntry` contains all the necessary information to perform this.

As such, the changes in this Pull Request update the `fs::Fs` trait with a new method, `restore`, which attempts to restore the trashed file back to its original location.

Note that this Pull Request will remain in Draft until the corresponding base Pull Request, https://github.com/zed-industries/zed/pull/52012, is merged.

## How to Review

1. Review the `fs::Fs::restore` method signature
2. Review the `fs::Fs::TrashRestoreError` enum, listing all errors we currently identify when attempting to restore a trashed file
3. Review the `fs::Fs::restore` implementations in both `fs::RealFs` and `fs::FakeFs`
    * In the case of `fs::RealFs` it just delegates to the `trash-rs` implementation
    * In the case of `fs::FakeFs` it leverages the `fs::FakeFsState` struct in order to simulate an actual move from the trash to the original location

## Self-Review Checklist

Release Notes:

- N/A


crates/fs/Cargo.toml
```diff
@@ -43,10 +43,6 @@ is_executable = "1.0.5"
 notify = "8.2.0"
 trash = { git = "https://github.com/dinocosta/trash-rs", branch = "5039-macos-restore" }
 
-[target.'cfg(target_os = "macos")'.dependencies]
-objc.workspace = true
-cocoa = "0.26"
-
```
dinocosta: I believe this should be in #52012 .

dinocosta: Done 🧹

yara-blue (approved):
Assuming the actual testing of the trash impl happens in trash-rs this is great :)

yara-blue (approved):
Good call on removing the `RemoveOptions`, cleans things up nicely :).

PR #52016, opened 2026-03-20
agent_ui: Smarter logic for when we drop background threads

benbrandt:
## Context

Previously we were very eagerly dropping threads from memory. However, reloading threads has a cost:
- For our agent, we have to replay events, which is lossy
- It's slower for external agents
- Some agents don't support reloading

So we keep the most recent 5 idle threads (or all threads who can't be reloaded) and clean up our list after every insertion.

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -4965,7 +4965,10 @@ impl AgentPanel {
 mod tests {
     use super::*;
     use crate::conversation_view::tests::{StubAgentServer, init_test};
-    use crate::test_support::{active_session_id, open_thread_with_connection, send_message};
+    use crate::test_support::{
+        active_session_id, open_thread_with_connection, open_thread_with_custom_connection,
+        send_message,
+    };
     use acp_thread::{StubAgentConnection, ThreadStatus};
     use agent_servers::CODEX_ID;
     use assistant_text_thread::TextThreadStore;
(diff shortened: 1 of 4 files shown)


crates/agent_ui/src/agent_panel.rs
```diff
@@ -1997,22 +1997,44 @@ impl AgentPanel {
...
+        self.background_threads
+            .insert(thread_view.read(cx).id.clone(), conversation_view);
+        self.cleanup_background_threads(cx);
+    }
 
-        if status != ThreadStatus::Generating {
-            return;
-        }
+    /// We keep at most 5 idle background threads.
+    /// We keep threads who have had the most recent events, or don't support reloading the full session.
```
bennetbo: We could also note here that we will always retain running threads

crates/agent_ui/src/conversation_view.rs
```diff
@@ -2620,6 +2629,16 @@ impl ConversationView {
             cx.notify();
         }
     }
+
+    pub fn set_updated_at_for_tests(&mut self, updated_at: Instant, cx: &mut Context<Self>) {
```
bennetbo: We should drop the `for_tests` and hide this behind `#[cfg(any(test, feature="test-support")]`

bennetbo (commented):
Looks good, we should guard that one function with a #cfg(test)... though

bennetbo (approved):

PR #52030, opened 2026-03-20
Fix close_session final save being a guaranteed no-op

rtfeldman:
`close_session` removed the session from `self.sessions` before calling `save_thread`, but `save_thread` immediately looks up the session in `self.sessions` and returns early if it's not found. This meant the explicit final save was always a no-op.

Additionally, when the local `session` variable was dropped at the end of the closure, it cancelled any in-flight save from a prior observe callback by dropping `session.pending_save`.

So on session close:
- The explicit final save was a guaranteed no-op
- Any in-flight save from an earlier observation was cancelled

**Fix:** Call `save_thread` while the session is still in the map, then extract the `pending_save` task and return it so the caller can actually await the save completing.

**Test:** Added `test_close_session_saves_thread` which sets a draft prompt without a `run_until_parked()` before `close_session`, so the only way the data gets persisted is if `close_session` itself performs the save. Also strengthened the existing `test_save_load_thread` similarly.

Closes AI-95

Release Notes:

- Fixed a bug where closing an agent thread could lose unsaved changes (e.g. draft prompts) made in the same frame.

### DIFF
--- a/crates/agent/src/agent.rs
+++ b/crates/agent/src/agent.rs
@@ -82,7 +82,7 @@ struct Session {
     /// The ACP thread that handles protocol communication
     acp_thread: Entity<acp_thread::AcpThread>,
     project_id: EntityId,
-    pending_save: Task<()>,
+    pending_save: Task<Result<()>>,
     _subscriptions: Vec<Subscription>,
 }
 
@@ -387,7 +387,7 @@ impl NativeAgent {
                 acp_thread: acp_thread.clone(),
                 project_id,
                 _subscriptions: subscriptions,
-                pending_save: Task::ready(()),
+                pending_save: Task::ready(Ok(())),
             },
         );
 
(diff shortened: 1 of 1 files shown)


bennetbo (approved):

crates/agent/src/agent.rs
```diff
@@ -1434,18 +1434,27 @@ impl acp_thread::AgentConnection for NativeAgentConnection {
...
+            let pending_save = session.pending_save;
 
             let has_remaining = agent.sessions.values().any(|s| s.project_id == project_id);
             if !has_remaining {
                 agent.projects.remove(&project_id);
             }
-        });
-        Task::ready(Ok(()))
+
+            cx.spawn(async move |_, _| {
```
bennetbo: Couldn't we just return the task here?

PR #52047, opened 2026-03-20
Feat unbind

probably-neb:
## Context

This PR adds an `Unbind` action, as well as syntax sugar in the keymaps for declaring it
```
{
  "unbind": {
    "tab: "editor::AcceptEditPrediction"
  }
}
```
Is equivalent to
```
{
  "bindings": {
    "tab: ["zed::Unbind", "editor::AcceptEditPrediction"]
  }
}
```
In the keymap, unbind is always parsed first, so that you can unbind and rebind something in the same block.

The semantics of `Unbind` differ from `NoAction` in that `NoAction` is treated _as an action_, `Unbind` is treated as a filter. In practice this means that when resolving bindings, we stop searching when we hit a `NoAction` (because we found a matching binding), but we keep looking when we hit an `Unbind` and filter out keystroke:action pairs that match previous unbindings. In essence `Unbind` is only an action so that it fits cleanly in the existing logic. It is really just a storage of deleted bindings.

The plan is to rework the edit predictions key bindings on top of this, as well as use `Unbind` rather than `NoAction` in the keymap UI. Both will be done in follow up PRs.

Additionally, in this initial implementation unbound actions are matched by name only. The assumption is that actions with arguments are bound to different keys in general. However, the current syntax allows providing arguments to the unbound actions. Both so that copy-paste works, and so that in the future if this functionality is added, keymaps will not break.

## How to Review

- The dispatch logic in GPUI
- The parsing logic in `keymap_file.rs`

## Self-Review Checklist

Release Notes:

- Added support for unbinding key bindings from the default keymaps. You can now remove default bindings you don't want, without having to re-declare default bindings that use the same keys. For instance, to unbind `tab` from `editor::AcceptEditPrediction`, you can put the following in your `keymap.json`
```
[
  {
    "context": "Editor && edit_prediction",
    "unbind": {
      "tab": "editor::AcceptEditPrediction"
    }
  }
]
```

### DIFF
--- a/crates/gpui/src/action.rs
+++ b/crates/gpui/src/action.rs
@@ -432,27 +432,35 @@ pub fn generate_list_of_all_registered_actions() -> impl Iterator<Item = MacroAc
 
 mod no_action {
     use crate as gpui;
-    use std::any::Any as _;
+    use schemars::JsonSchema;
+    use serde::Deserialize;
 
     actions!(
         zed,
         [
             /// Action with special handling which unbinds the keybinding this is associated with,
             /// if it is the highest precedence match.
-            NoAction,
-            /// Action with special handling which unbinds only later bindings with the exact same
-            /// keystrokes and context predicate, while allowing other bindings to fall through.
-            Unbind
+            NoAction
         ]
     );
 
+    /// Action with special handling which unbinds later bindings for the same keystrokes when they
+    /// dispatch the named action, regardless of that action's context.
+    ///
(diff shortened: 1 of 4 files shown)


nooblyf:
@probably-neb This makes GPUI's TestAppContext `cx.all_action_names();` leak "zed" in the names.

probably-neb:
@nooblyf this is an existing limitation with built-in GPUI actions

PR #52057, opened 2026-03-21
Make editor::LineUp & editor::LineDown honor vertical_scroll_margin like vim::LineUp & vim::LineDown

TwoClocks:
## Context

This is an implementation of this feature:
https://github.com/zed-industries/zed/discussions/49821

Although, I'd argue it's a bug fix, not a feature.

Either way : This copies the window scroll logic from the `vim` mode versions with out all the extra logic for visual mode.

Could refactor the common logic out of the vim code and make it common. But that seems like a bigger PR. Happy to take a stab at it if that's what you prefer.
Could also add new commands for the new behavior if you prefer. I didn't do that, because it seems like more clutter in the commands, and my belief that the existing behavior is a bug. But happy to do that if you prefer.

## How to Review

creates new function `scroll_screen_with_cursor_margin` in `scroll.rs`
wires up editor::LineUp/LineDown to new function in `elements.rs`
adds a test in  `editor_tests.rs`

## Self-Review Checklist

Release Notes:

- editor::LineUp/LineDown commands now honor vertical_scroll_margin (same as vim::LineUp/LineDown)


Vastargazing:
Currently, vim::LineUp respects per-instance margin. Your PR makes editor::LineUp respect margin too, but from a global source. This introduces a divergence where the same setting behaves differently depending on which command is used. Since set_vertical_scroll_margin exists as an instance method, this function should use it too.

SomeoneToIgnore (approved):
Thank you, adjusted more actions that seemed to need it.

PR #52066, opened 2026-03-21
agent_ui: Show estimated cost per agent thread

mvanhorn:
## Context

Zed tracks token usage per agent thread and knows model pricing, but these two pieces of data are never combined. Users see "45.2k / 1M" tokens and have no way to know what that cost in dollars without checking their provider dashboard.

This adds estimated cost to the token usage display in agent threads. When a model provides pricing data (`LanguageModelCostInfo::TokenCost`), the cost is calculated from cumulative input/output token counts and displayed alongside the existing token information.

**In the circular progress tooltip** (non-split display): cost appears as a new "Estimated cost" section below the context percentage.

**In the split token display**: cost appears as an inline label after the output token count.

Models without pricing data (local models, credit-based Zed-hosted models using `RequestCost`) gracefully omit the cost label.

## Why this matters

Users choosing between models for a task (Opus vs Haiku, GPT-4o vs GPT-4o-mini) currently can't see real-time cost. The subagent model discussion ([#52042](https://github.com/zed-industries/zed/issues/52042)) exists specifically because cost is opaque. Users reporting unexpected token consumption in [#50254](https://github.com/zed-industries/zed/issues/50254) and [#51646](https://github.com/zed-industries/zed/issues/51646) would benefit from seeing cost accumulate in real-time.

## Changes

- `crates/language_model/src/language_model.rs`: Added `estimate_cost()` and `format_cost()` to `LanguageModelCostInfo`. Cost is calculated as `(input_tokens * input_rate / 1M) + (output_tokens * output_rate / 1M)`. Formatting handles sub-cent amounts ("< $0.01"), normal range ("$0.47"), and large amounts ("$16").
- `crates/agent_ui/src/conversation_view/thread_view.rs`: Added cost display to both the split token display (inline label) and circular progress tooltip (new section) in `render_token_usage()`.
- 4 unit tests for cost calculation and formatting.

## Testing

- `cargo test -p language_model -- test_estimate_cost test_format_cost` - all 4 tests pass
- `cargo check -p agent_ui` - compiles cleanly
- `cargo clippy -p language_model -p agent_ui` - no warnings
- `cargo fmt -- --check` - clean

This contribution was developed with AI assistance (Claude Code).

Release Notes:

- Added estimated cost display in agent thread token usage tooltip

### DIFF
--- a/crates/language_model/src/language_model.rs
+++ b/crates/language_model/src/language_model.rs
@@ -932,6 +932,30 @@ impl LanguageModelCostInfo {
             SharedString::from(format!("{:.2}", cost))
         }
     }
+
+    pub fn estimate_cost(&self, input_tokens: u64, output_tokens: u64) -> Option<f64> {
+        match self {
+            LanguageModelCostInfo::TokenCost {
+                input_token_cost_per_1m,
+                output_token_cost_per_1m,
+            } => {
+                let cost = (input_tokens as f64 * input_token_cost_per_1m / 1_000_000.0)
+                    + (output_tokens as f64 * output_token_cost_per_1m / 1_000_000.0);
+                Some(cost)
+            }
+            LanguageModelCostInfo::RequestCost { .. } => None,
+        }
+    }
+
+    pub fn format_cost(cost: f64) -> SharedString {
+        if cost < 0.005 {
+            SharedString::from("< $0.01")
+        } else if cost < 10.0 {
+            SharedString::from(format!("${:.2}", cost))
+        } else {
+            SharedString::from(format!("${:.0}", cost))
+        }
+    }
 }
 
 impl LanguageModelProviderId {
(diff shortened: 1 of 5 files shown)


crates/language_model/src/language_model.rs
```diff
@@ -932,6 +932,30 @@ impl LanguageModelCostInfo {
             SharedString::from(format!("{:.2}", cost))
         }
     }
+
+    pub fn estimate_cost(&self, input_tokens: u64, output_tokens: u64) -> Option<f64> {
```
osyvokon: The cost of an agentic thread is dominated by cache reads followed by cache writes. We absolutely need to account for these if we want to show a reasonable estimate.

Then there's different prices depending on context size and things like fast mode. Pricing pages give some ideas for what to look for: [Anthropic](https://platform.claude.com/docs/en/about-claude/pricing#model-pricing), [OpenAI](https://developers.openai.com/api/docs/pricing), [Google](https://ai.google.dev/gemini-api/docs/pricing)

osyvokon (changes requested):

crates/language_model/src/language_model.rs
mvanhorn: Good point. Fixed in af77e4b — `estimate_cost()` now accounts for cache read and cache creation tokens via optional rates on `TokenCost`. The `TokenUsage` struct carries all four token counts (input, output, cache_read, cache_creation) through the pipeline.

Context-size and fast-mode pricing differences are provider-level concerns that would need changes to how models report their `CostInfo` — happy to look at that as a follow-up if you'd like.

osyvokon:
Thanks for opening the PR. Unfortunately we decided not to merge it. Estimating costs is hard (there are many factors), and it's a moving target, so older client will stuck with outdated estimates. We will think about how to do this differently in the future.

PR #52081, opened 2026-03-21
settings: Use fallible parsing in `edits_for_update` to preserve settings

### LINKED ISSUE #41344: `update_settings_file` deletes settings if settings file needs to be migrated

OmChillure:
## Context

`update_settings_file` deletes unrelated settings when the settings file contains deprecated keys that need migration. For example, changing the model from the Agent Panel overwrites the entire `agent` block instead of just updating `default_model`.

The root cause is that `edits_for_update` used `parse_json_with_comments` (strict parser), which returns `Err` on deprecated/unknown fields. The error is swallowed by `log_err()`, falling back to `Default::default()` (empty settings). The diff then sees everything as new and replaces the entire block.

The fix switches to `parse_json` (the fallible/lenient parser), which returns `Some(parsed_value)` even when deprecated fields are present  - the same pattern already used by `parse_and_migrate_zed_settings`.

## How to Review

Single-file change in `settings_store.rs`, focus on `edits_for_update` . Compare with `parse_and_migrate_zed_settings` (line 702) which already uses the same `parse_json` approach.

## Self-Review Checklist

## Video
[Screencast from 2026-03-21 00-36-12.webm](https://github.com/user-attachments/assets/31bd584a-2674-4c91-bdb2-69ed8fa35e88)

Release Notes:

- Fixed settings being overwritten when updating a single setting via UI while the settings file contains deprecated keys.

### DIFF
--- a/crates/agent_ui/src/agent_configuration.rs
+++ b/crates/agent_ui/src/agent_configuration.rs
@@ -1348,38 +1348,45 @@ async fn open_new_agent_servers_entry_in_settings_editor(
             let settings = cx.global::<SettingsStore>();
 
             let mut unique_server_name = None;
-            let edits = settings.edits_for_update(&text, |settings| {
-                let server_name: Option<String> = (0..u8::MAX)
-                    .map(|i| {
-                        if i == 0 {
-                            "your_agent".to_string()
-                        } else {
-                            format!("your_agent_{}", i)
-                        }
-                    })
-                    .find(|name| {
-                        !settings
-                            .agent_servers
-                            .as_ref()
-                            .is_some_and(|agent_servers| agent_servers.contains_key(name.as_str()))
-                    });
-                if let Some(server_name) = server_name {
-                    unique_server_name = Some(SharedString::from(server_name.clone()));
-                    settings.agent_servers.get_or_insert_default().insert(
-                        server_name,
-                        settings::CustomAgentServerSettings::Custom {
-                            path: "path_to_executable".into(),
-                            args: vec![],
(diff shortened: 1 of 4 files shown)


MrSubidubi (changes requested):
Passing by and am fairly sure that we want a test case for this so that we both know the fix is good and that it will not regress in the future. 

Similarly, I think we should probably at least discuss what to do when the parsing failed as oppposed to just falling back to defaults here and then messing around with the user settings anyway despite them having obvious errors.

PR #52082, opened 2026-03-21
agent_ui: Fix pasted image context showing Image instead of actual filename

monkey-mode:
## What

Fix image context mentions always showing the generic label `Image` instead of the actual filename when pasting from Finder or picking via the `+` → Image button in the Agent Panel.

## Why

`insert_images_as_context` hardcoded the crease label to `MentionUri::PastedImage.name()` (`"Image"`) for every image, regardless of whether it originated from a named file. Both code paths that load images from file paths — `paste_images_as_context` and `add_images_from_picker` — discarded the filename before passing images to the shared insert function.

## Fix

- `agent_ui/src/mention_set.rs`: Changed `insert_images_as_context` to accept `Vec<(gpui::Image, SharedString)>` instead of `Vec<gpui::Image>`, using the provided name as the crease label. In `paste_images_as_context`, extract `file_name()` from each path and pair it with the loaded image. Raw clipboard images (screenshots, copy from image editors) continue to use `"Image"` as there is no filename.

- `agent_ui/src/message_editor.rs`: Same fix for `add_images_from_picker` — extract `file_name()` from each selected path and pass it alongside the image.

Closes #52079

## Test Plan

- [x] `cargo build -p agent_ui` compiles clean
- [x]  `cargo fmt --all -- --check` format check
- [x] Manual verification of:
   - [x] Copy an image file in Finder (`Cmd+C`), paste into Agent Panel — mention shows actual filename
   - [x] `+` → Image → pick a file — mention shows actual filename
   - [x] Screenshot paste (`Cmd+Shift+4`) still shows `Image`
   - [x] Regular text paste still works

## Screenshots

image

---
Release Notes:

- Fixed image context mentions always showing `Image` instead of the actual filename when pasting from Finder or using the image picker in the Agent Panel

### DIFF
--- a/crates/agent_ui/src/mention_set.rs
+++ b/crates/agent_ui/src/mention_set.rs
@@ -739,7 +739,7 @@ mod tests {
 /// Inserts a list of images into the editor as context mentions.
 /// This is the shared implementation used by both paste and file picker operations.
 pub(crate) async fn insert_images_as_context(
-    images: Vec<gpui::Image>,
+    images: Vec<(gpui::Image, SharedString)>,
     editor: Entity<Editor>,
     mention_set: Entity<MentionSet>,
     workspace: WeakEntity<Workspace>,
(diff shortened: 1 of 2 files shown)


bennetbo (approved):
Thank you!

PR #52104, opened 2026-03-21
Focus buffer/multibuffer when clicking a file in the Git panel

### LINKED ISSUE #51712: Single vs double clicking files in project/Git panels don't focus the buffer/diff consistently
### Reproduction steps

1. Double-click a file in the Git UI

In the Project panel:

- Single-clicking a file does not focus the buffer (happens whether preview tabs are enabled/disabled, also see https://github.com/zed-industries/zed/issues/34357)
- Double-clicking a file opens it in a normal tab and focuses the buffer

In the Git panel:

- Single-clicking a file jumps to it in the Uncommitted Changes view, but does not focus the multibuffer
- Double-clicking a file opens it in a normal tab, but does not focus the buffer (inconsistent with the Project panel)

There is a case to be made that the buffer shouldn't be focused when opening a preview tab. Because currently, you can use the arrow keys to navigate files in the project panel and spacebar to open it in a preview tab.

Following that logic, you can also use the arrow keys to navigate files in the Git panel and spacebar to jump to the respective diff in the multibuffer. I don't think this is extremely useful, but not sure if this should be a supported workflow.

This might need some design work.

OS: macOS 15.7.4
Memory: 16 GiB
Architecture: x86_64

Dnreikronos:
Context

Fixes #51712

Clicking a file in the Git panel did not focus the opened buffer/multibuffer, which was inconsistent with the Project panel behavior. Single-clicking a file jumped to it in the Uncommitted Changes view but kept focus on the Git panel. Double-clicking opened the file in a normal tab but also did not focus the buffer.

Two changes were made in `crates/git_ui/src/git_panel.rs`:
- Single-click: After deploying the ProjectDiff, focus the ProjectDiff editor instead of the Git panel's own focus handle.
- Double-click: Pass `focus_item: true` to `open_path_preview` so the opened buffer receives focus.

## How to Review

Small PR — focus on the two changes in `git_panel.rs`:
- `open_diff()` (~line 1301-1308): After `deploy_at`, we now focus the ProjectDiff editor via `project_diff.focus_handle(cx).focus(window, cx)`
- `open_file()` (~line 1351): Changed `focus_item` from `false` to `true` in `open_path_preview`
- Click handler (~line 5267): Removed redundant `self.focus_handle.focus(window, cx)` since `open_diff` now handles focus internally

## Video
https://github.com/user-attachments/assets/fea4e6d0-a6e9-41ba-8601-9753525f6855

## Self-Review Checklist

Release Notes:

- Fixed clicking a file in the Git panel not focusing the buffer/multibuffer (#51712)


Dnreikronos:
Hi, @cole-miller!
Probably here we will have to have a design opinion about that changes

SomeoneToIgnore:
Thanks, seems that this PR is no longer relevant: the code had shifted a lot, and `main` has a deliberately opposing behavior; the issue is closed either.

PR #52107, opened 2026-03-21
Fix incorrect rainbow bracket matching in Markdown

kaikozlov:
## Context

Fixes #52022.

Rainbow bracket matching could become incorrect when tree-sitter returned ambiguous bracket pairs for the same opening delimiter. The repair path rebuilt pairs using a shared stack across all bracket query patterns, which let excluded delimiters like Markdown single quotes interfere with parenthesis matching.

This change scopes that repair logic to each bracket query pattern so ambiguous matches are rebuilt without mixing unrelated delimiter types. It also adds a regression test for the Markdown repro from the issue.

image

image

## How to Review

Review `crates/language/src/buffer.rs` first, especially the fallback repair path for bogus tree-sitter bracket matches.

Then review `crates/editor/src/bracket_colorization.rs`, which adds regression coverage for the issue repro.

## Self-Review Checklist

Release Notes:

- Fixed rainbow brackets in Markdown when quotes caused parentheses to match incorrectly

### DIFF
--- a/crates/editor/src/bracket_colorization.rs
+++ b/crates/editor/src/bracket_colorization.rs
@@ -392,6 +392,20 @@ where
             &bracket_colors_markup(&mut cx),
             "All markdown brackets should be colored based on their depth, again"
         );
+
+        cx.set_state(indoc! {r#"ˇ('')('')
+
+((''))('')
+
+('')((''))"#});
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore (approved):
Thank you!

PR #52111, opened 2026-03-22
Fix terminal block missing first line via f32 tolerance

Dnreikronos:
## Context

`TerminalBounds::num_lines()` uses `floor(height / line_height)` to compute the terminal grid row count. When the height is derived from `N * line_height` (as it is for inline/embedded terminals in the Agent Panel), IEEE 754 float32 arithmetic can produce `N - epsilon` instead of `N`, causing `floor()` to return `N - 1`. This makes the terminal grid one row too small, leaving the first line of output in invisible scrollback (since `display_offset = 0`). The same issue applies to `num_columns()`.

The fix adds a small tolerance (`0.01`) before flooring, which absorbs float precision errors without affecting genuine fractional results.

Closes #51609

## How to Review

Small PR — focus on the tolerance value (`0.01`) in `num_lines()` and `num_columns()` in `crates/terminal/src/terminal.rs`. The two new tests (`test_num_lines_float_precision`, `test_num_columns_float_precision`) verify the fix across 1,000+ float combinations that previously triggered the bug.

## Self-Review Checklist

Release Notes:

- Fixed the first line of terminal output sometimes missing in Agent Panel terminal blocks.

### DIFF
--- a/crates/terminal/src/terminal.rs
+++ b/crates/terminal/src/terminal.rs
@@ -211,12 +211,12 @@ impl TerminalBounds {
         // `N * line_height / line_height` can be N-epsilon, which floor()
         // would round down, pushing the first line into invisible scrollback.
         let raw = f32::from(self.bounds.size.height / self.line_height);
-        (raw + 0.01).floor() as usize
+        raw.next_up().floor() as usize
     }
 
     pub fn num_columns(&self) -> usize {
         let raw = f32::from(self.bounds.size.width / self.cell_width);
-        (raw + 0.01).floor() as usize
+        raw.next_up().floor() as usize
     }
 
     pub fn height(&self) -> Pixels {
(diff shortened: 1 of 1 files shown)


crates/terminal/src/terminal.rs
```diff
@@ -207,11 +207,16 @@ impl TerminalBounds {
     }
 
     pub fn num_lines(&self) -> usize {
-        (self.bounds.size.height / self.line_height).floor() as usize
+        // Tolerance to prevent f32 precision from losing a row:
+        // `N * line_height / line_height` can be N-epsilon, which floor()
+        // would round down, pushing the first line into invisible scrollback.
+        let raw = f32::from(self.bounds.size.height / self.line_height);
+        (raw + 0.01).floor() as usize
```
Veykril:
```suggestion
        raw.next_up().floor() as usize
```
Might be better to use this here?

Dnreikronos: Yeah, definitely, thanks for your attention and time invested here.
I made the changes following your recommendation.

PR #52119, opened 2026-03-22
workspace: Always open projects in a new window

vorbei:
Currently, opening a project replaces the current workspace window when certain conditions are met (no dirty items, no remote session, no existing worktree). This can be surprising — users expect their current window to stay open.

This change makes Zed always open a new window when opening a project, regardless of the current window state.

**Changes:**

- `workspace/src/workspace.rs`: `open_workspace_for_paths` always sets `window_to_replace = None`, ignoring the `replace_current_window` parameter.
- `workspace/src/multi_workspace.rs`: Simplify `open_paths` to always pass `replace_current_window: false`, removing the conditional branch that previously closed the current window.

Release Notes:

- Changed project opening to always create a new window instead of replacing the current workspace.

### DIFF
--- a/crates/workspace/src/multi_workspace.rs
+++ b/crates/workspace/src/multi_workspace.rs
@@ -652,28 +652,9 @@ impl MultiWorkspace {
     ) -> Task<Result<Entity<Workspace>>> {
         let workspace = self.workspace().clone();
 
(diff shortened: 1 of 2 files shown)


vorbei:
CLA has been signed. @dinocosta @reflectronic — could you re-review when you get a chance? Thanks!

PR #52123, opened 2026-03-22
agent: Add a new `subagent_model` setting to use specific model in sub agents

GoldStrikeArch:
Closes #52042

## Overview

Adds a `subagent_model` setting and chat UI control for choosing which model subagents spawned via `spawn_agent` should use. Users can leave subagents on `Automatic`, or choose a custom tool-capable model for subagent work while keeping a different model for the main agent thread.

Custom subagent model choices are stored on the current thread and written back to `settings.json`, matching the "sticky behavior" of the main agent model selector for future threads.

## Context

- **Request:** Users want to use fast/cheap models for subagents while keeping strong reasoning models for the main agent thread.
- **UI:** The chat footer now includes a `Subagent model` selector for native parent threads.
- **Automatic mode:** When no custom subagent model is configured, subagents use `default_model` if it supports tools, otherwise they fall back to the parent thread's model.
- **Custom mode:** The dropdown only shows tool-capable models.
- **Sticky behavior:** When a custom subagent model is active, parent model/thinking/speed/profile changes are not propagated to subagents. This prevents later parent-thread changes from overriding the intentionally different subagent generation settings.
- **Storage model:** The selected subagent model is persisted per thread via `subagent_model_selection`, while settings still remember the latest custom choice as `agent.subagent_model`.

## Configuration

```json
"agent": {
  "default_model": {
    "provider": "zed.dev",
    "model": "claude-opus-4-7"
  },
  "subagent_model": {
    "provider": "zed.dev",
    "model": "claude-sonnet-4-5",
    "enable_thinking": false
  }
}
```

Omitting `subagent_model` is equivalent to `Automatic`.

## How to Review

1. Start with `thread.rs` around `new_subagent()` — this contains the model resolution chain, the tool-capability guard, per-thread `subagent_model_selection`, and the dedicated subagent thinking override.
2. Review the propagation guards in `set_model` / `set_thinking_enabled` / `set_thinking_effort` / `set_speed` / `set_profile` — these keep custom subagent settings stable.
3. Review `agent_model_selector.rs` and `language_model_selector.rs` — these add the `SubagentModelSelector`, the `Automatic` option, and tool-capable filtering.
4. Review `conversation_view.rs` / `thread_view.rs` — these mount the selector in the chat footer for parent native-agent threads.
5. Review the settings and persistence plumbing in `agent_settings.rs`, `settings_content/src/agent.rs`, `db.rs`, and thread metadata helpers.
6. Most of the remaining `thread.rs` diff is test infrastructure and new tests for automatic/custom subagent model behavior.

## Self-Review Checklist

Release Notes:

- Added a Subagent model setting and chat control for choosing Automatic or a custom tool-capable model for subagents spawned via `spawn_agent`.


GoldStrikeArch:
@benbrandt @maxbrunsfeld  Hi! Any updates on this one?

buzzvolt:
should the user be asked to figure out from a category of Small/Medium/Large models to choose from? (owing to choices based on speed, economics and flexibility)

GoldStrikeArch:
> should the user be asked to figure out from a category of Small/Medium/Large models to choose from? (owing to choices based on speed, economics and flexibility)

IMO there should be an ability to explicitly set the model you want (the exact version, maybe even from openai and not anthropic, etc)

buzzvolt:
hi @GoldStrikeArch thanks for working on this feature

- ideally, the llm/provider setting for/as a sub-agent need not be static, and it can change across the chat sessions and depending on the projects/tasks involved. 
- I would prefer if there is a hierarchial dropdown navigation at the agent selection in the chat settings itself that allows for selecting (choices of) sub-agents similar to the current option of selecting the llm provider/model, and more importantly, if the sub-agents are even allowed per se in the current session (toggle on/off) kind of behavior so user/developer is still in full control of the agentic session

what do you think of this?

GoldStrikeArch:
> I would prefer if there is a hierarchial dropdown navigation at the agent selection in the chat settings itself that allows for selecting (choices of) sub-agents similar to the current option of selecting the llm provider/model, and more importantly, if the sub-agents are even allowed per se in the current session (toggle on/off) kind of behavior so user/developer is still in full control of the agentic session

Yeah, this looks useful to have and would improve the UX compared to the current solution for sure. But that might be what Zed team will implement on their own I guess, see [this comment](https://github.com/zed-industries/zed/issues/52042#issuecomment-4353446283)

I will try to check how this can be implemented and will do it in this branch but the question (product question actually) is what models we should allow to chose from? Like the same models we already have to chose from (even OPUS 4.7, etc)?

After some thinking:
- It looks like the easiest way is indeed to allow to select any model for the subagents for now. I will try this approach

GoldStrikeArch:
@buzzvolt Alright, I updated this PR and branch and now it has the following:
1) UI settings for the subagent model (which behaves exactly the same as the current dropdown for the "main model", which is when you select it it "stores the current selection" for the thread and also updates your "global settings" for the default model)
2) You can choose the models the same way you choose the "main model". But I am filtering out the models which are not "tool capable"

All of that looks like this now


https://github.com/user-attachments/assets/322b9c5a-a350-484a-9898-65b53829a1c5

buzzvolt:
awesome!! thank you, that was a very quick response. appreciate it! I will try this 

@zed-guild @zed-intelligence hopefully take notes of this.

benbrandt:
Thanks for taking a stab at this! @bennetbo happened to also address this in #56203 and it seems to be a simpler approach so I'm going to close this one. Hopefully it helps!

GoldStrikeArch:
@benbrandt Sure, I don't mind. The only thing I want to highlight is that these are the things #56203 is missing which are present in this PR
- it doesn't filter models without `supports_tools()` (otherwise it is kinda useless to use a model without tools support as subagent)
- it doesn't "persist sub_agent model per thread, similar to the main model"

There is also no UI picker but I guess this was intended (I am also fine with that)

buzzvolt:
@benbrandt @bennetbo  @franciskafyi please do take a look at the comments from @GoldStrikeArch they are really worth it, most importantly those two points and the UI picker. It is worth pulling them in, if not make appropriate changes. @GoldStrikeArch has been very considerate of making this work. thanks!

benbrandt:
- Regarding point 1, I think if the user wants to pick a model that doesn't support tools, they are configuring a custom setting and that is I guess there prerogative to choose
- We are working on migrating AI settings to the main settings and can address it once there
- it does persist, as we don't update the subagent model if this is set

GoldStrikeArch:
@benbrandt 

> Regarding point 1, I think if the user wants to pick a model that doesn't support tools, they are configuring a custom setting and that is I guess there prerogative to choose

Thanks, that makes sense on `supports_tools()` as a product choice.

> it does persist, as we don't update the subagent model if this is set

I think we may be using “persist” differently in this context. #56203 persists `agent.subagent_model` as a global/user setting, and it keeps an already-created subagent from being overwritten by later parent model/thinking/speed changes when that setting is present.

What I meant by “persist per thread, similar to the main model” is different: storing the parent thread’s chosen subagent model in the thread metadata itself. The main model is persisted via `DbThread.model`, but #56203 doesn’t add a corresponding `subagent_model_selection` field to `Thread`/`DbThread` or serialize it in `to_db`/`from_db`. `new_subagent()` resolves from the current global `AgentSettings::subagent_model` each time.

So with #56203, if `agent.subagent_model` changes later, an old thread does not retain the subagent model choice that was active for that thread.

PR #52130, opened 2026-03-22
languages: Improve semantic token highlighting for parameters and Python

lingyaochu:
## Context

Zed's semantic token highlighting does not cover all token types returned by language servers, so the highlighting looks fairly primitive compared with tree-sitter highlighting, especially for Python language servers. This PR adds some global and Python-specific rules for better highlighting.

I need to admit that the built-in Python language servers currently have weak semantic highlighting implementations. Pylance, the closed-source Python language server from Microsoft, provides the best highlighting for now, but I think ty will do better, even though it still has a long way to go.
## How to Review

Basically, this is a rule-adding change. Some rules are made global, and some are made Python-specific.

## Self-Review Checklist

Release Notes:

- Improved semantic token highlighting for parameters and Python

### DIFF
--- a/assets/settings/default_semantic_token_rules.json
+++ b/assets/settings/default_semantic_token_rules.json
@@ -121,6 +121,11 @@
   // References
   {
     "token_type": "parameter",
+    "token_modifiers": ["declaration"],
+    "style": ["variable.parameter"]
+  },
+  {
+    "token_type": "parameter",
     "token_modifiers": [],
     "style": ["variable"],
   },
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore (changes requested):
Given that something visual is changed in this PR and we have https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#sending-changes , please add before/after screenshots.

SomeoneToIgnore (approved):
Thank you a lot!

PR #52149, opened 2026-03-22
Fix fuzzy matcher to use platform-aware path separators instead of hardcoded "/" 

anaslimem:
## Context

The fuzzy matcher in `crates/fuzzy/src/matcher.rs` had hardcoded `'/'` characters when checking for path separators. This was not platform-aware - on Windows, the path separator is `\` while the code only checked for `/`.

This change replaces all hardcoded path separator checks with a platform-aware helper function that uses `std::path::MAIN_SEPARATOR`. The `RelPath` type already normalizes all paths to use `/` internally, so this change ensures the fuzzy matcher works correctly across all platforms while maintaining backward compatibility.

## How to Review

- Focus on the `is_path_separator()` helper function in `crates/fuzzy/src/matcher.rs:27-29`
- Verify all 3 hardcoded `'/'` checks are replaced with the new helper
- Run `cargo test -p fuzzy` to confirm all tests pass

## Self-Review Checklist

- [x] Unsafe blocks (if any) have justifying comments — N/A
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) — N/A

Release Notes:

- Fixed fuzzy matcher to use platform-aware path separators instead of hardcoded '/'

### DIFF
--- a/crates/fuzzy/src/matcher.rs
+++ b/crates/fuzzy/src/matcher.rs
@@ -1,5 +1,6 @@
 use std::{
     borrow::Borrow,
+    path::MAIN_SEPARATOR,
     sync::atomic::{self, AtomicBool},
 };
 
@@ -9,8 +10,6 @@ const BASE_DISTANCE_PENALTY: f64 = 0.6;
 const ADDITIONAL_DISTANCE_PENALTY: f64 = 0.05;
 const MIN_DISTANCE_PENALTY: f64 = 0.2;
 
-// TODO:
-// Use `Path` instead of `&str` for paths.
 pub struct Matcher<'a> {
     query: &'a [char],
     lowercase_query: &'a [char],
@@ -22,6 +21,10 @@ pub struct Matcher<'a> {
     last_positions: Vec<usize>,
     score_matrix: Vec<Option<f64>>,
     best_position_matrix: Vec<usize>,
+}
+
+fn is_path_separator(c: char) -> bool {
+    c == MAIN_SEPARATOR
 }
 
 pub trait MatchCandidate {
@@ -226,7 +229,7 @@ impl<'a> Matcher<'a> {
                     None => continue,
                 }
             };
-            let is_path_sep = path_char == '/';
+            let is_path_sep = is_path_separator(path_char);
 
             if query_idx == 0 && is_path_sep {
                 last_slash = j;
@@ -245,7 +248,7 @@ impl<'a> Matcher<'a> {
                         None => path[j - 1 - prefix.len()],
                     };
 
-                    if last == '/' {
+                    if is_path_separator(last) {
                         char_score = 0.9;
                     } else if (last == '-' || last == '_' || last == ' ' || last.is_numeric())
                         || (last.is_lowercase() && curr.is_uppercase())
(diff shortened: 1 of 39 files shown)


themavik (commented):
Matcher: MAIN_SEPARATOR instead of hardcoded slash fixes Windows scoring; strings that still use only forward slashes (URIs, some remotes) may need both separators in is_path_separator.

anaslimem:
True, so you think i even need to resolve how the data flows through the matcher not only replace the hardcoded "/" with a more dynamic separator function to work on both os (because they do not use the same separator)?

anaslimem:
@dinocosta  Can u clarify the intent here ?

anaslimem:
> > True, so you think i even need to resolve how the data flows through the matcher not only replace the hardcoded "/" with a more dynamic separator function to work on both os (because they do not use the same separator)?
> 
> Yes, though i think those are two separate things:
> 
> Your PR fixes the immediate bug: the separator check now works on both OS. That's correct and useful on its own.
> 
> The TODO was pointing at a deeper refactor: instead of passing paths as `&[char]` and manually checking separators, the idea was to use `std::path::Path` directly and rely on `.components()` which is already OS-aware by design.
> 
> But doing that in a fuzzy matcher is tricky because of performance; raw slice scanning is usually much faster than iterating over path components. Since i'm still exploring the codebase, pls correct me if i’m misinterpreting the original design goals here.

I think what you have said looks right. i will juste wait for the clarification of the comment from one of the maintaners so i know either i change the implementation or i keep my own

crates/fuzzy/src/matcher.rs
```diff
@@ -25,6 +24,10 @@ pub struct Matcher<'a> {
     best_position_matrix: Vec<usize>,
 }
 
+fn is_path_separator(c: char) -> bool {
+    c == '/' || c == MAIN_SEPARATOR
```
Veykril: This will be wrong due to the existence of remoting, if the remote uses a different path separator than the client this will break. We need to thread a `PathStyle` through here (supplied by users of the fuzzy matcher)

anaslimem: True i will change it right away 
thank you

crates/fuzzy/src/strings.rs
```diff
@@ -167,6 +168,7 @@ where
                         query_char_bag,
                         smart_case,
                         penalize_length,
+                        PathStyle::local(),
```
Veykril: We unfortunately need to push this to callers too. I know this has a lot of call sites, some of which can be local probably. If you do not want to spend the time on that thats fine though! In that case put a TODO comment here

anaslimem: No all good i will check it

anaslimem: I Found that some call sites don't match paths (commands, themes, etc.),
so made path_style optional. For those I pass None (defaults to local).
For path-related call sites (file picker, recent projects) I pass
the project's PathStyle so separator detection works correctly when
remoting to different platforms.

anaslimem:
@Veykril  is this good enough for now ? as it should be ?

SomeoneToIgnore:
Will close to keep the queue cleaner: CI fails for over 3 weeks and there are conflicts.

PR #52210, opened 2026-03-23
agent_ui: Don't reset cursor position when editing queued messages

danilo-leal:
## Context

When typing on the read-only queued message editor, we push the content to the main, editable message editor. However, the cursor position was being reset when that happened. This PR fixes that by tracking the cursor offset when doing that, also accounting for the case where there could be pre-existing content in the message editor.

## How to Review

- Queue a message in the agent panel
- Place your cursor somewhere in the middle of your queued message and type something
- See how the cursor position is preserved once it goes down to the main message editor

## Self-Review Checklist

Release Notes:

- Agent: Fix cursor position being reset when editing queued messages.

### DIFF
--- a/crates/agent_ui/src/message_editor.rs
+++ b/crates/agent_ui/src/message_editor.rs
@@ -1599,9 +1599,9 @@ impl MessageEditor {
     ) {
         self.editor.update(cx, |editor, cx| {
             let snapshot = editor.buffer().read(cx).snapshot(cx);
-            let clamped = MultiBufferOffset(offset.min(snapshot.len().0));
+            let offset = snapshot.clip_offset(MultiBufferOffset(offset), text::Bias::Left);
             editor.change_selections(Default::default(), window, cx, |selections| {
-                selections.select_ranges([clamped..clamped]);
+                selections.select_ranges([offset..offset]);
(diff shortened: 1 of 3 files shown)


crates/agent_ui/src/message_editor.rs
```diff
@@ -1580,6 +1591,21 @@ impl MessageEditor {
...
 
+    pub fn set_cursor_offset(
+        &mut self,
+        offset: usize,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        self.editor.update(cx, |editor, cx| {
+            let snapshot = editor.buffer().read(cx).snapshot(cx);
+            let clamped = MultiBufferOffset(offset.min(snapshot.len().0));
```
maxbrunsfeld: I think we may want to use `MultiBufferSnapshot::clip_offset` here, rather than `offset.min(snapshot.len().0)`, because that will also ensure that the offset is not in the middle of a multi-byte character. It's also slightly shorter, since it takes and a `MultiBufferOffset`.

maxbrunsfeld (commented):
This looks good - just left one suggestion.

PR #52217, opened 2026-03-23
Fix git panel context menu keybinding jitter

Dnreikronos:
Context

The git panel's context menu caused visual jitter (flickering/jumping) when opened via right-click on a tracked file. The root cause was in `dispatch_context()`: it used `self.focus_handle == focused` to check if the panel itself was directly focused, but when a context menu opened, focus moved to the menu (a child element), causing the `"menu"` and `"ChangesList"` key contexts to be dropped. This triggered a re-render with different keybindings, which re-added them, creating a loop of jitter.

The fix replaces the direct focus equality check with `self.focus_handle.contains_focused(window, cx)`, which returns `true` when any child element (including the context menu) holds focus within the panel's focus tree. This is consistent with how other panels (project panel, outline panel, collab panel) handle focus-based dispatch contexts.

Closes #51813

## Demo

**Before fix:**

https://github.com/user-attachments/assets/e18d49b2-72a6-4411-8ec5-519e36628f29

**After fix:**

https://github.com/user-attachments/assets/94c936d2-1e81-4d28-a86a-8b1ed76ddde1

## How to Review

This is a small, focused change in a single file: `crates/git_ui/src/git_panel.rs`.

1. **The fix** (~line 974): `dispatch_context()` method — the old code checked direct focus equality (`self.focus_handle == focused`), the new code uses `self.focus_handle.contains_focused(window, cx)` and restructures the conditionals so `CommitEditor` is checked first via `if/else if`.
2. **The test** (~line 7871): `test_dispatch_context_with_focus_states` — verifies 4 focus state transitions: commit editor focused, changes list focused, back to commit editor, and back to changes list. Each case asserts the correct key contexts are present/absent.

## Self-Review Checklist

Release Notes:

- Fixed git panel context menu jitter caused by keybinding dispatch context flickering when right-clicking on tracked files (#51813)

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -975,16 +975,11 @@ impl GitPanel {
         let mut dispatch_context = KeyContext::new_with_defaults();
         dispatch_context.add("GitPanel");
 
-        if window
-            .focused(cx)
-            .is_some_and(|focused| self.focus_handle == focused)
-        {
-            dispatch_context.add("menu");
-            dispatch_context.add("ChangesList");
-        }
-
         if self.commit_editor.read(cx).is_focused(window) {
             dispatch_context.add("CommitEditor");
+        } else if self.focus_handle.contains_focused(window, cx) {
+            dispatch_context.add("menu");
+            dispatch_context.add("ChangesList");
         }
 
         dispatch_context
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

Issue #23774, opened 2025-01-28
Doesn't support Node.js stack trace file links


Qard:
### Describe the feature

Node.js stack trace lines are of the form `at Object.method (/path/to/file.js:12:34)`. The contents extracted from the parentheses works as a clickable file location on its own, thanks to #22559, but _within_ the parentheses it is not recognized so Node.js stack trace file locations remain unclickable. It would be nice if it could recognize those to make the file paths clickable.

A full error trace looks something like:

```
Error: my message
      at Object.method (/path/to/file.js:12:34)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async Test.run (node:internal/test_runner/test:935:9)
      at async startSubtestAfterBootstrap (node:internal/test_runner/harness:296:3) 
```

The `node:` prefixed paths might also be nice to support by doing a jump to definition or something like that, but the local file paths are _almost_ there for support, just need recognition of the parentheses wrap.

Qard:
Wow, that was fast! Thanks @SomeoneToIgnore. 👏🏻

Issue #18556, opened 2024-09-30
Support opening files URLs surrounded by parenthesis


silvenon:
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

In the integrated terminal's stack trace if a file URL is surrounded by parenthesis (this is common with Vite/Node.js). Observe the examples below:

```
      at trim_prefix (/Users/me/Code/my-project/node_modules/express/lib/router/index.js:321:13)
      at /Users/me/Code/my-project/node_modules/express/lib/router/index.js:284:7
```

The URL below is clickable, but the one above is not.

### Environment

Zed: v0.154.3 (Zed)
OS: macOS 15.0.0
Memory: 16 GiB
Architecture: aarch64

PR #52222, opened 2026-03-23
Fix terminal path detection inside parentheses

butvinm:
## Summary

Paths inside parentheses without a preceding space (e.g. `Update(.claude/skills/sandbox/SKILL.md)` or `Write(/test/cool.rs)` from Claude Code output) were not clickable in the terminal. The `(` character was allowed as a middle character in the default path hyperlink regex, causing the entire `Update(.path/here` to be matched as a single invalid path.

**Changes:**

- Remove `(` from the middle-chars alternation (`[:(]` → `:`) in the default path hyperlink regex, so `(` always acts as a path boundary — consistent with it already being excluded from first and last character sets. Preserves upstream's space exclusion after `:`.
- Iterate all regex matches in the line instead of only the first, so the correct path (which may be the second match after a prefix like `Update`) is found. This also simplifies the code by removing the separate hovered-word search logic.

**Known tradeoff:**

Filenames with parentheses in the middle (e.g. `docker-compose.prod(copy).yml`) are no longer matched as a single path. This is uncommon in terminal output contexts (compiler errors, stack traces, tool output) and is documented with a `should_panic` test.

**What doesn't break:**

- `(/path/file.js:321:13)` — `(` at word start is excluded by the first-char rule (unchanged)
- Node.js stack traces like `at fn (/path/file.js:10:5)` — space before `(` makes it a separate word
- All 38 terminal hyperlink tests pass

Related: #18556, #23774

Release Notes:

- Fixed terminal path detection for paths inside parentheses without preceding space (e.g. `Update(path)` or `Write(path)` patterns from CLI tool output)


crates/terminal/src/terminal_hyperlinks.rs
```diff
@@ -1003,6 +969,17 @@ mod tests {
...
+            #[test]
+            #[should_panic(expected = "Path = «copy).yml»")]
+            // Filenames with parentheses in the middle (e.g. `file(copy).txt`) are not
+            // matched as a single path because `(` is treated as a path boundary. This is
+            // a tradeoff to support the more common `Tool(path)` pattern (e.g. Claude Code
+            // output like `Update(.claude/SKILL.md)`) where `(` is a delimiter, not part
+            // of the filename.
+            fn parens_in_filename_not_matched() {
+                test_path!("‹«docker-compose.prod(👉copy).yml»›");
+            }
```
Veykril, line 1015: This isn't really an acceptable tradeoff. Some folks tend to have usernames with parentheses inside which breaks especially on Windows systems

butvinm: I overlooked that it might be a common use case.

I implemented detection of such file paths following sanitize_url_punctuation logic. Now paths like /tmp/test(copy).txt are clickable.

Veykril (approved):
Thanks!

Veykril (approved):
Thanks!

PR #52228, opened 2026-03-23
sidebar: Adjust the sidebar UI slightly

mikayla-maki:
This PR:
- Moves the "remove workspace" option into the context menu, so that it's not swapping places with the + button in some circumstances. This matches codexs behavior. Another solution to the problem could be having the add button remain even when the "new thread" option exists.
- Adds a "go to workspace" button to the header

## Self-Review Checklist

Release Notes:

- N/A


danilo-leal (approved):
:shipit:

PR #52229, opened 2026-03-23
project: Use path-scoped language server selection for buffer registration

trevordixon:
## Summary

(Might be related to #21990.) I want to enable the Deno language server for one directory in my monorepo and turn off all other TS language servers. With this change, language server routing respects directory-level settings so irrelevant language servers don't contribute to directories where they aren't welcome.

This change solves the problem for my use case, but if it isn't the way to go, feel free to close and consider this a vote for support in some other way! The change seems clean and in line with existing abstractions though to me.

- make language-server adapter selection path-scoped (buffer path) while preserving manifest/worktree-rooted server identity
- rebind buffer-to-server registrations when settings-driven adapter selection changes
- reconcile stale buffer registrations so deselected servers receive buffer unregister without forcing process churn
- add integration coverage for path-scoped TypeScript/Deno selection and dynamic rebind behavior

## Why
Zed already supports directory-scoped `.zed/settings.json`, but server selection in this flow was effectively resolved from manifest/worktree root. In monorepos that mix Deno and TypeScript by directory, this can cause both servers to attach to files where only one should run.

This change keeps server process rooting unchanged (good workspace semantics), but makes adapter selection and buffer registration path-aware so each file is attached only to the servers selected for that file path.

Release Notes:

- Fixed language server selection in monorepos to respect path-scoped `.zed/settings.json` so files can attach to different servers by directory.

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -2808,6 +2808,22 @@ impl LocalLspStore {
                 }
             })
             .collect::<Vec<_>>();
+        // Partial registration (e.g. restart selected servers) must not drop
+        // existing bindings for non-targeted servers.
+        if only_register_servers.is_empty() {
+            let desired_server_ids = servers_and_adapters
+                .iter()
+                .map(|(server, _)| server.server_id())
+                .collect::<HashSet<_>>();
+            self.unregister_buffer_from_obsolete_language_servers(
+                buffer_handle,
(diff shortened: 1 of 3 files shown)


osiewicz (commented):
2-centing: I'm a bit uneasy about that change, because IIRC our manifest handling exploits the assumption that a language server enabled in the parent directory is also implicitly enabled for children.

PR #52233, opened 2026-03-23
sidebar: Make sure workspace has git state loaded when opening a new one

Anthony-Eid:
## Context

This fixes a bug where switching to a worktree workspace sometimes briefly flashes it as a top-level project in the sidebar. This happened because the sidebar would rebuild its entries whenever the multi workspace emits the added workspace event. 

The fix is checking from the root workspace git linked worktree list as well to see if newly added workspaces are linked.

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/crates/sidebar/src/sidebar.rs
+++ b/crates/sidebar/src/sidebar.rs
@@ -669,6 +669,19 @@ impl Sidebar {
         let mut absorbed: HashMap<usize, (usize, SharedString)> = HashMap::new();
         let mut pending: HashMap<Arc<Path>, Vec<(usize, SharedString, Arc<Path>)>> = HashMap::new();
         let mut absorbed_workspace_by_path: HashMap<Arc<Path>, usize> = HashMap::new();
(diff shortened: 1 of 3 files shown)


maxbrunsfeld (approved):

crates/sidebar/src/sidebar.rs
```diff
@@ -6074,6 +6081,100 @@ mod tests {
...
+
+        let window = cx.windows()[0];
+        cx.update_window(window, |_, window, cx| {
+            window.dispatch_action(Confirm.boxed_clone(), cx);
+        })
+        .unwrap();
+
+        let mut saw_opened_worktree_workspace = false;
+        for _ in 0..50 {
+            if !cx.dispatcher.tick(false) {
```
maxbrunsfeld: not blocking, but could we observe the sidebar instead of using this 50-iteration loop

maxbrunsfeld (approved):

crates/sidebar/src/sidebar.rs
Anthony-Eid: resolved

PR #52234, opened 2026-03-23
Fix blank git diff view on Windows when working in a subfolder

### LINKED ISSUE #51758: Diff panel don`t work when you open not a root project folder
### Reproduction steps

open folder (Open Folder... Ctrl-K Ctrl-O) that is not a root folder of your project
make some changes
go to git panel and try to view diffs

Image

OS: Windows 10.0.26100
Memory: 31.7 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics 630 || Intel Corporation || 31.0.101.2125

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

HiteshRohira:
## Context

Fixes a Windows path handling bug in `crates/util/src/paths.rs` when computing a relative path via `strip_prefix`. (Closes #51758 )

Previously, Windows prefix matching only handled drive-letter case-insensitivity. It could still fail when the parent and child paths referred to the same location but used different separator styles (`\` vs `/`) or different casing in later path segments. That caused valid Windows paths to return `None` instead of a relative path, which caused the diff to break.

This change normalizes Windows paths for prefix comparison by lowercasing any ASCII characters and converting backslashes to forward slashes before checking the prefix.

Tests were added for mixed-case and mixed-separator Windows paths to cover the bug and prevent further regressions.

## How to Review

1. Check the `strip_prefix` Windows branch to confirm the new normalization logic only affects prefix comparison and still rejects partial-segment matches.
2. Check the added tests for mixed separator and mixed casing cases on Windows paths.

## Self-Review Checklist

Release Notes:

- Fix blank git diff view on Windows when working in a subfolder


crates/util/src/paths.rs
```diff
@@ -443,13 +443,17 @@ impl PathStyle {
...
+            let normalized_child = normalize_windows_path_for_prefix(child);
+            let normalized_parent = normalize_windows_path_for_prefix(parent);
+            let stripped = normalized_child.strip_prefix(normalized_parent.as_ref())?;
+            if stripped.is_empty() {
+                ""
+            } else if stripped.starts_with('/') {
+                &child[normalized_parent.len()..]
+            } else {
+                return None;
+            }
```
cole-miller: As an alternative way to structure this, could we introduce `PathStyle::decompose(&self, abs_path: &Path)` that gives you an optional drive letter and a `RelPath` for the part after the leading separator? Then we could delegate to `RelPath::strip_prefix`, if we could figure out a way for `RelPath` to have the same case-insensitivity properties on Windows (which admittedly might be tricky)

devoper2022:
can we know the status of this PR

HiteshRohira:
@devoper2022 just waiting for approval

devoper2022:
@Anthony-Eid can you review this PR

crates/util/src/paths.rs
Veykril: `PathStyle::strip_prefix` is definitely bugged here in that it doesn't consider the two separator variations on windows. It is very easy to end up with mixed style separator paths on windows through the std library APIs.

Veykril: Why is this function not delegating to `std::path::Path::strip_prefix` in the first place?

HiteshRohira: @Veykril have removed this function altogether and made a change in a different place. Marking this as resolved to avoid confusion

reflectronic (approved):
Thank you

PR #52236, opened 2026-03-23
Always open microphone track on call join

JosephTLyons:
## Context

This PR works around a 1-2 second audio silence that occurs when unmuting for the first time during a call for Bluetooth headphone users with `mute_on_join` enabled. Bluetooth headphones must switch from A2DP (listening-only) to HFP (microphone-enabled) mode when the mic is first activated, which interrupts audio output. Now the microphone track opens on join (still muted at the publication level), moving this switch to when users already expect a connection delay / audio instability. It should be noted that I expect that the OS microphone indicator will now appear on channel join rather than on first unmute, but I feel this is a justified side effect:

1. This is consistent with other apps—joining a Slack huddle with "Mute my microphone" enabled also shows the indicator immediately on join while muted. 
1. Zed users without `mute_on_join` already see the OS-level microphone indicator for the entire call (even after muting) and already experience the Bluetooth profile switch at join time—this change simply brings `mute_on_join` users in line with that existing behavior.

## Self-Review Checklist

Release Notes:

- Fixed 1-2 seconds of audio silence when unmuting for the first time during a call with Bluetooth headphones when `mute_on_join` is enabled.

### DIFF
--- a/crates/call/src/call_impl/room.rs
+++ b/crates/call/src/call_impl/room.rs
@@ -1773,7 +1773,12 @@ fn spawn_room_connection(
                 });
                 this.diagnostics = Some(cx.new(|cx| CallDiagnostics::new(weak_room, cx)));
 
-                if !muted_by_user && this.can_use_microphone() {
+                // Always open the microphone track on join, even when
+                // `muted_by_user` is set. `share_microphone` will still mute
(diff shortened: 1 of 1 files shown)


kubkon (approved):
Looks great to me!

PR #52238, opened 2026-03-23
Extract `language_core` and `grammars` crates from `language`

nathansobo:
This extracts a `language_core` crate from the existing `language` crate, and creates a `grammars` data crate. The goal is to separate tree-sitter grammar infrastructure, language configuration, and LSP adapter types from the heavier buffer/editor integration layer in `language`.

## Motivation

The `language` crate pulls in `text`, `theme`, `settings`, `rpc`, `task`, `fs`, `clock`, `sum_tree`, and `fuzzy` — all of which are needed for buffer integration (`Buffer`, `SyntaxMap`, `Outline`, `DiagnosticSet`) but not for grammar parsing or language configuration. Extracting the core types lets downstream consumers depend on `language_core` without pulling in the full integration stack.

## Dependency graph after extraction

```
language_core   ← gpui, lsp, tree-sitter, util, collections
grammars        ← language_core, rust_embed, tree-sitter-{rust,python,...}
language        ← language_core, text, theme, settings, rpc, task, fs, ...
languages       ← language, grammars
```

## What moved to `language_core`

- `Grammar`, `GrammarId`, and all query config/builder types
- `LanguageConfig`, `LanguageMatcher`, bracket/comment/indent config types
- `HighlightMap`, `HighlightId` (theme-dependent free functions `highlight_style` and `highlight_name` stay in `language`)
- `LanguageName`, `LanguageId`
- `LanguageQueries`, `QUERY_FILENAME_PREFIXES`
- `CodeLabel`, `CodeLabelBuilder`, `Symbol`
- `Diagnostic`, `DiagnosticSourceKind`
- `Toolchain`, `ToolchainScope`, `ToolchainList`, `ToolchainMetadata`
- `ManifestName`
- `SoftWrap`
- LSP data types: `BinaryStatus`, `ServerHealth`, `LanguageServerStatusUpdate`, `PromptResponseContext`, `ToLspPosition`

## What stays in `language`

- `Buffer`, `BufferSnapshot`, `SyntaxMap`, `Outline`, `DiagnosticSet`, `LanguageScope`
- `LspAdapter`, `CachedLspAdapter`, `LspAdapterDelegate` (reference `Arc<Language>` and `WorktreeId`)
- `ToolchainLister`, `LanguageToolchainStore` (reference `task` and `settings` types)
- `ManifestQuery`, `ManifestProvider`, `ManifestDelegate` (reference `WorktreeId`)
- Parser/query cursor pools, `PLAIN_TEXT`, point conversion functions

## What the `grammars` crate provides

- Embedded `.scm` query files and `config.toml` files for all built-in languages (via `rust_embed`)
- `load_queries(name)`, `load_config(name)`, `load_config_for_feature(name, grammars_loaded)`, and `get_file(path)` functions
- `native_grammars()` for tree-sitter grammar registration (behind `load-grammars` feature)

## Pre-cleanup (also in this PR)

- Removed unused `Option<&Buffer>` from `LspAdapter::process_diagnostics`
- Removed unused `&App` from `LspAdapter::retain_old_diagnostic`
- Removed `fs: &dyn Fs` from `ToolchainLister` trait methods (`PythonToolchainProvider` captures `fs` at construction time instead)
- Moved `Diagnostic`/`DiagnosticSourceKind` out of `buffer.rs` into their own module

## Backward compatibility

The `language` crate re-exports everything from `language_core`, so existing `use language::Grammar` (etc.) continues to work unchanged. The only downstream change required is importing `CodeLabelExt` where `.fallback_for_completion()` is called on the now-foreign `CodeLabel` type.

Release Notes:

- N/A


crates/grammars/src/bash/brackets.scm
MrSubidubi, line 1: Is there a reason that we need to duplicate all the language configurations here as opposed to e.g. symlinking them or moving them from the old to the new crate? 

I think that otherwise we'll see these files get out of sync rather fast and while we could add some CI checks to ensure these are always updated in sync, I'd be curious if there is a good reason as to why we need the duplication here.

nathansobo: I think you're right and we should work harder to dedupe them.

agu-z: Done

crates/language/src/highlight_map.rs
```diff
@@ -1,67 +1,27 @@
...
-        theme.get_capture_name(self.0 as usize)
-    }
+pub fn build_highlight_map(capture_names: &[&str], theme: &SyntaxTheme) -> HighlightMap {
+    HighlightMap::from_ids(
+        capture_names
+            .iter()
+            .map(|capture_name| {
+                theme
+                    .highlight_id(capture_name)
+                    .map_or_else(HighlightId::default, HighlightId)
```
MrSubidubi: Since `HighlightId` only has a size of 4, I think we could consider just keeping the `map_or` that we had prior.

crates/language/src/highlight_map.rs
```diff
@@ -1,67 +1,27 @@
...
+pub fn highlight_style(id: HighlightId, theme: &SyntaxTheme) -> Option<HighlightStyle> {
+    theme.get(id.index() as usize).cloned()
 }
 
-impl Default for HighlightId {
-    fn default() -> Self {
-        DEFAULT_SYNTAX_HIGHLIGHT_ID
-    }
+pub fn highlight_name(id: HighlightId, theme: &SyntaxTheme) -> Option<&str> {
+    theme.get_capture_name(id.index() as usize)
```
MrSubidubi: Not sure whether that's worth the effort here (or better), but given that these are basically just wrapping methods on `SyntaxTheme` now, we could perhaps just use these directly and get rid of the methods here (and maybe also just have a deref for `HighlightId` to `usize`). 
That would also alllow us to get rid of the castings here.

MrSubidubi (approved):
Thanks for taking care of the merge. Left two thoughts but consider neither blocking.

crates/language/src/highlight_map.rs
agu-z: Totally. That didn't make sense to change.

crates/language/src/highlight_map.rs
agu-z: Sure! I'll see what that looks like

PR #52244, opened 2026-03-23
dap: Support IPv6 addresses in TCP transport

moktamd:
The DAP TCP transport layer was hardcoded to `Ipv4Addr`, so IPv6 addresses like `fd00::a` in a debug config's `connect.host` always failed with `hostname must be IPv4: invalid IPv4 address syntax`.

Replaced `Ipv4Addr` with `IpAddr` and `SocketAddrV4` with `SocketAddr` across the `task`, `dap`, `dap_adapters`, and `project` crates. The WASM extension API still uses `u32` for the host field to avoid a breaking WIT interface change; IPv4 round-trips through extensions as before.

Fixes #52237

Release Notes:

- Fixed DAP TCP transport rejecting IPv6 addresses when connecting to remote debug adapters.

### DIFF
--- a/crates/dap/src/adapters.rs
+++ b/crates/dap/src/adapters.rs
@@ -18,7 +18,7 @@ use std::{
     borrow::Borrow,
     ffi::OsStr,
     fmt::Debug,
-    net::Ipv4Addr,
+    net::IpAddr,
     ops::Deref,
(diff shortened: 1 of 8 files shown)


tomhoule (approved):
Thank you for the contribution!

PR #52249, opened 2026-03-23
extension_api: Add support for setting the working dir for `Command`

vitallium:
Adds an optional `working-dir` field to the `Command` type across the extension API.

## Context

Extensions that use the `Command` sometimes need to run commands in a specific directory, for example, in a worktree root. Without this, the CWD was uncontrolled/implicit.

This makes it explicit and overridable.
If `working_dir` is not set, it defaults to the extension's work directory. This is the current behavior.

## How to Review

I think the most important bit is extension API compatibility.  
I tried to keep the existing extensions that rely on the `Command` API
working in the same way as they do currently.

Not sure if a bump is requires but happy to do that.

## Self-Review Checklist

Release Notes:

- Add support for setting the working dir for `Command` in Extension API

### DIFF
--- a/crates/extension/src/types.rs
+++ b/crates/extension/src/types.rs
@@ -23,6 +23,8 @@ pub struct Command {
     pub args: Vec<String>,
     /// The environment variables to set for the command.
     pub env: EnvVars,
+    /// The working directory for the command
+    pub working_dir: Option<PathBuf>,
 }
 
 impl std::fmt::Debug for Command {
@@ -37,6 +39,7 @@ impl std::fmt::Debug for Command {
             .field("command", &self.command)
             .field("args", &self.args)
             .field("env", &filtered_env)
+            .field("working_dir", &self.working_dir)
             .finish()
     }
 }
--- a/crates/extension_api/src/process.rs
+++ b/crates/extension_api/src/process.rs
@@ -9,6 +9,7 @@ impl Command {
             command: program.into(),
             args: Vec::new(),
             env: Vec::new(),
+            working_dir: None,
         }
     }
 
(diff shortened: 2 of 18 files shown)


MrSubidubi:
Thank you for this!

Unfortunately, aside from having issues with the WIT modifications, I think this would be better handled by properly introducing the manifest concept to extensions, which I think would handle your use case here. 

Furthermore, I think this PR could be misused in certain circumstances.

Thus, am gonna close for now, but still, thanks for the attempt!

vitallium:
> Unfortunately, aside from having issues with the WIT modifications

Do you mean compatibility with the released API?

>I think this would be better handled by properly introducing the manifest concept to extensions, which I think would handle your use case here.

Is there a discussion or issue where I can learn more about the manifest idea?
 
> Furthermore, I think this PR could be misused in certain circumstances.

Yeah, I totally agree. Coming back to this PR, I clearly see this potential. For example, an approved `./tool --version` could be turned into execution of any attacker-chosen `./tool` by switching the cwd to a different directory.

> Thus, am gonna close for now, but still, thanks for the attempt!

No problem at all. My biggest worry is that the `cwd` of the Command API is not deterministic. Would it be possible to take a different approach and have the Command API always set the `cwd` to the current worktree root?

Thanks!

MrSubidubi:
> Do you mean compatibility with the released API?

Yep.

> Is there a discussion or issue where I can learn more about the manifest idea?

I do not think so, however, that is already implemented for Python in core (and one of the reasons Basedpyright was lifted into core).

> Would it be possible to take a different approach and have the Command API always set the cwd to the current worktree root?

I think it would definitely be interesting to find out why this seems to be nondeterministic, would depend on what you'd find there. But if you find an approach there that would make sense, we could definitely take another look.

PR #52257, opened 2026-03-23
git_hosting_providers: Add Hugging Face support

julien-c:
## Summary

- Add a `HuggingFace` Git hosting provider so that commit hashes and file references become clickable links to `huggingface.co`
- Supports both `huggingface.co` and `hf.co` (SSH shorthand) hosts, HTTPS and SSH remote URLs, and model/dataset/space repository types
- Add Hugging Face to the docs list of supported Git hosting services

## Test plan

- [x] All 11 unit tests pass (`cargo test -p git_hosting_providers --lib providers::huggingface`)
- [x] `cargo clippy -p git_hosting_providers` passes with no warnings
- [ ] Manual: clone a Hugging Face repo and verify commit hashes render as clickable links

Release Notes:

- Added support for Hugging Face as a Git hosting provider, making commit hashes and file references clickable links to huggingface.co.

### DIFF
--- a/crates/git_hosting_providers/src/git_hosting_providers.rs
+++ b/crates/git_hosting_providers/src/git_hosting_providers.rs
@@ -26,6 +26,7 @@ pub fn init(cx: &mut App) {
     provider_registry.register_hosting_provider(Arc::new(Gitea::public_instance()));
     provider_registry.register_hosting_provider(Arc::new(Gitee));
     provider_registry.register_hosting_provider(Arc::new(Github::public_instance()));
+    provider_registry.register_hosting_provider(Arc::new(HuggingFace));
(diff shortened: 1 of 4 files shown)


cole-miller:
Thanks, but this seems a bit too niche for us to have in Zed core. At some point we'd like to allow extensions to define Git hosting providers so that we can support arbitrary providers without this maintenance tradeoff, but for now we unfortunately have to be selective about what providers we merge support for.

julien-c:
well Zed has an active account on HF (https://huggingface.co/zed-industries) whereas you don't seem to have one on SourceHut, Codeberg, etc

but nevermind! Thanks for reading anyways

PR #52267, opened 2026-03-23
tasks: Prioritize user-defined tasks and add auto_detected_tasks setting

### LINKED ISSUE #36759: Tasks: Manually defined tasks should supersede other project tasks
### Summary

When tasks are manually defined in a `.zed/tasks.json` I'd expect them to appear at the top of the list, rather than below other project tasks (like those in a `package.json`)

### Description

(cc @osiewicz!)

In my case, I've used tasks to narrow down the list of runnables to those that I actually care about, and having to scroll to the bottom of the list somewhat defeats the purpose of the `tasks.json`

Image

In the image above, the green is what I'd expect to see at the top, rather than the purple.

Steps to reproduce:
1. Create a `tasks.json` in any project with runnables
2. Open `task:spawn`
3. 💥

(also, 👋👋👋)

Zed: v0.200.4 (Zed) 
OS: macOS 15.5.0
Memory: 24 GiB
Architecture: aarch64

micthiesen:
## Summary

- Reorder task source preferences so user-defined tasks (`.zed/tasks.json`, global config, one-off commands) appear above auto-detected ones (LSP, language extensions) in the task picker
- Add a project-level `auto_detected_tasks` setting (`"all"` or `"hidden"`) to hide auto-detected tasks entirely

CleanShot 2026-03-23 at 15 23 27

Closes https://github.com/zed-industries/zed/issues/36759

## Design notes

The new ordering for `task_source_kind_preference` is: Worktree > Global > UserInput > LSP > Language. The previous ordering had LSP and Language first, which meant auto-detected tasks (e.g. npm scripts from package.json) appeared above manually curated tasks in `.zed/tasks.json`, defeating the purpose of defining them.

For the `auto_detected_tasks` setting, I considered a few shapes:

- **Per-source toggles** (separate booleans for LSP vs language extension tasks): More granular, but the existing per-language `tasks.enabled` and `prefer_lsp` settings already cover that use case. A project-wide toggle felt more appropriate for "I just want my own tasks."
- **A filter/allowlist approach**: More powerful but higher complexity for a setting most people would use as a simple on/off.
- **An enum with more variants** (e.g. `"lsp_only"`, `"language_only"`): Premature without concrete demand. The enum is easy to extend later if needed.

The setting is not yet in the visual settings UI (same as other project-level settings like `load_direnv`). It works via `settings.json`.

## Test coverage

- Ordering test validates the full preference chain with assertion messages
- Settings deserialization test for the new enum
- Integration test verifies worktree/global tasks survive the `"hidden"` setting
- Full integration testing of Language/LSP filtering would require a heavier test setup with a language context provider. That path is covered by the early-return in `editor::lsp_tasks()` and the inventory filter, but not exercised end-to-end in a test.

CLA signed by @micthiesen.

Release Notes:

- Improved task picker ordering to show manually defined tasks above auto-detected ones
- Added `auto_detected_tasks` project setting to hide auto-detected tasks from language servers and extensions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17387,6 +17387,7 @@ dependencies = [
  "project",
  "serde",
  "serde_json",
+ "settings",
  "task",
  "tree-sitter-rust",
  "tree-sitter-typescript",
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1559,6 +1559,14 @@
   // 3. Don't load direnv configuration at all.
   //      "load_direnv": "disabled"
   "load_direnv": "direct",
+  // Controls the visibility of auto-detected tasks (e.g. npm scripts from
+  // package.json, tasks from language servers and extensions).
+  // Manually defined tasks from `.zed/tasks.json` and global tasks are
+  // always shown regardless of this setting.
+  //
+  // "all": Show all auto-detected tasks.
+  // "hidden": Hide all auto-detected tasks.
+  "auto_detected_tasks": "all",
   "edit_predictions": {
     // Which edit prediction provider to use.
     "provider": "zed",
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1563,6 +1563,7 @@
   // package.json, tasks from language servers and extensions).
   // Manually defined tasks from `.zed/tasks.json` and global tasks are
   // always shown regardless of this setting.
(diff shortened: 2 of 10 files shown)


osiewicz (changes requested):
I don't think we should change the sort order of tasks based on preference of one person alone. If anything, we could expose it as a setting and let people decide for themselves.

PR #52269, opened 2026-03-23
agent_ui: Fix full screen behavior on worktree creation

danilo-leal:
## Context

This PR removes the behavior in which we turn the agent panel full screen on worktree creation, even if there aren't any files open. Changing the width behavior of the panel will ultimately solve this, and auto-full screening it turned out to be surprising. Also, in reaction to this, I've added an icon button in the agent panel's toolbar when in full screen so it's easier to get out of it. Eventually though, we can probably remove it (also because of the flexible width work).

## How to Review

- Create a new worktree from a thread and see the panel with the same width
- See the "disable full screen" icon button in the toolbar when you full screen it

--- 

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -2994,8 +2994,6 @@ impl AgentPanel {
                 }
 
                 // If we had an active buffer, remap its path and reopen it.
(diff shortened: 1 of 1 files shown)


Anthony-Eid (approved):
looks good

PR #52272, opened 2026-03-23
extension_cli: Ensure extension that provide themes do not provide further features

MrSubidubi:
We generally enforce this policy already and also want this to be this way, so better to be error than sorry here. 

Release Notes:

- N/A


maxbrunsfeld (commented):
Why do we have this restriction?

maxbrunsfeld (approved):

PR #52275, opened 2026-03-24
gpui: Ensure global keybinds do not have a too high precedence in some cases

MrSubidubi:
## Context

This fixes an issue we were seeing where global keybindings would take precedence over keybinds with a context that would match a context higher up in the key context stack. 

## Self-Review Checklist

Release Notes:

- Fixed an issue where global keybindings would take precedence over keybinds with a context that matched a key context higher up in the key context stack. E.g., if your keybind had the context `Pane` and the current key context would be `Pane > Editor`, any global keybind with the same keys would take precedence. This is now fixed.

### DIFF
--- a/crates/gpui/src/keymap.rs
+++ b/crates/gpui/src/keymap.rs
@@ -242,12 +242,13 @@ impl Keymap {
         (bindings, !pending.is_empty())
     }
     /// Check if the given binding is enabled, given a certain key context.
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
Autoclosing all draft PRs with no activity for 2 weeks or over, this one also has CI failures: 

image

PR #52285, opened 2026-03-24
git_ui: Fix Git Graph button flicker in Git panel

cppcoffee:
Release Notes:

- Fixed Git Graph button flicker in Git panel

Replace the Git Graph button visibility check with a stable feature-flag gate instead of querying action availability from the current window focus path, which could briefly remove the button during initial panel render and after opening Git Graph.

before:

https://github.com/user-attachments/assets/0af38a49-b28d-4b50-9d63-a4676fd229f7

after:

https://github.com/user-attachments/assets/fcdf8245-bd34-45d0-9ea3-8671fc99a7c2


kubkon (approved):
Looks great, thanks!

PR #52297, opened 2026-03-24
remote_connection: Add visibility toggle to password prompt

cppcoffee:
This adds an eye icon button to the right of the password input field when connecting to a remote device, allowing users to toggle the visibility of the typed password. The toggle defaults to masked and only appears for actual password prompts, avoiding yes/no confirmations.

After modification, an eye button has been added:

https://github.com/user-attachments/assets/76fb5a89-0aa3-4017-b050-d6bb9c2ad4a8

--- 

Release Notes:

- Added a toggle to control password visibility when connecting to a remote project.


danilo-leal (approved):
Thank you!

PR #52303, opened 2026-03-24
agent: Allow model to respond with stringified array when it edits

bennetbo:
## Context

Sometimes the model likes to respond with `"[...]"` for the edits array, instead of just `[...]`, which causes deserialisation to fail:
```
Failed to receive tool input: invalid type: string "[{"old_text": ..., "new_text": ...}]", expected a sequence
```

## Self-Review Checklist

Release Notes:

- N/A


rtfeldman (approved):
I'm surprised models are doing this, but this does seem like the right way to deal with it! 👍

PR #52308, opened 2026-03-24
acp: Fix the spawn directory for agent servers in remote environments

benbrandt:
We can't add a current_dir when in remote scenarios in the same place, it gets added elsewhere. 

Reused the same worktree selection logic, and make sure this gets passed to the RPC call instead.

This fixes something that hasn't made it to preview yet.

Release Notes:

- N/A


bennetbo (approved):

maan2003:
This is still happening btw but only on "new worktree" path. cd is done on local side for newly created worktrees.

PR #52313, opened 2026-03-24
agent: Reduce DB load in ThreadMetadataStore

bennetbo:
## Context

We now batch multiple db operations together if they happen very frequently (similar to how we handle it in `thread_history.rs`)

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/thread_metadata_store.rs
+++ b/crates/agent_ui/src/thread_metadata_store.rs
@@ -3,7 +3,7 @@ use std::{path::Path, sync::Arc};
 use acp_thread::AgentSessionInfo;
 use agent::{ThreadStore, ZED_AGENT_ID};
 use agent_client_protocol as acp;
-use anyhow::{Context as _, Result};
+use anyhow::Context as _;
 use chrono::{DateTime, Utc};
 use collections::HashMap;
 use db::{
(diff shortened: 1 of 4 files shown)


crates/agent_ui/src/thread_metadata_store.rs
```diff
@@ -338,17 +343,56 @@ impl SidebarThreadMetadataStore {
...
         };
         let _ = this.reload(cx);
         this
     }
 
+    fn dedup_db_operations(operations: Vec<DbOperation>) -> Vec<DbOperation> {
+        let mut ops = HashMap::default();
+        for operation in operations.into_iter().rev() {
+            if ops.contains_key(operation.id()) {
+                continue;
```
benbrandt, line 389: If we get an insert and then a delete: should we prioritize the delete? Or have some other sort of deduping?

benbrandt: I mean I guess it is reversed so the last one wins. I think that makes sense

benbrandt (approved):

crates/zed/src/zed.rs
```diff
@@ -5867,8 +5890,9 @@ mod tests {
             })
             .unwrap();
 
-        // --- Flush serialization ---
-        cx.executor().advance_clock(SERIALIZATION_THROTTLE_TIME);
+        cx.run_until_parked();
+        flush_workspace_serialization(&window_a, cx).await;
```
bennetbo, line 5894: This test broke because it relied on ordering of some async tasks, so we explicitly flush serialisation here and await it

benbrandt (approved):

PR #52333, opened 2026-03-24
Don't block thread creation when PromptStore fails to initialize

rtfeldman:
NativeAgentServer::connect used `prompt_store.await?` which turned any PromptStore initialization failure (permission denied, disk issues, etc.) into a hard connection error, putting the ConversationView into LoadError state and preventing all new native agent threads from being created.

Changed to `.log_err()` so the error is logged at ERROR level but thread creation proceeds with `prompt_store: None`. `NativeAgent::new` already accepts `Option<Entity<PromptStore>>` and handles `None` gracefully — user custom prompts won't load but threads still work.

Added a regression test that serializes a stale thread ID, loads the panel (triggering the "last active thread not found in database" warning), then dispatches NewThread through the real NativeAgentServer path and verifies it produces a connected thread.

Release Notes:

- Agent panel now handles filesystem errors more gracefully

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -6248,4 +6248,42 @@ mod tests {
             "the new worktree workspace should use the same agent (Codex) that was selected in the original panel",
         );
     }
+
+    /// Regression test: NewThread must produce a connected thread even when
+    /// the PromptStore fails to initialize (e.g. LMDB permission error).
+    /// Before the fix, `NativeAgentServer::connect` propagated the
+    /// PromptStore error with `?`, which put every new ConversationView
+    /// into LoadError and made it impossible to start any native-agent
+    /// thread.
+    #[gpui::test]
+    async fn test_new_thread_with_prompt_store_error(cx: &mut TestAppContext) {
+        let (panel, mut cx) = setup_panel(cx).await;
+
+        // NativeAgentServer::connect needs a global Fs.
+        let fs = FakeFs::new(cx.executor());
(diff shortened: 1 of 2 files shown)


crates/agent_ui/src/agent_panel.rs
```diff
@@ -6293,4 +6248,91 @@ mod tests {
...
+
+        // Write serialized panel state with a native-agent thread whose
+        // session_id does NOT exist in the threads database.
+        let workspace_id = workspace.read_with(cx, |ws, _| ws.database_id().unwrap());
+        let kvp = cx.update(|_, cx| KeyValueStore::global(cx));
+        let serialized = SerializedAgentPanel {
+            width: Some(px(300.0)),
+            selected_agent: Some(AgentType::NativeAgent),
+            last_active_thread: Some(SerializedActiveThread {
+                session_id: "stale-thread-00000000-0000-0000-0000-000000000000".to_string(),
```
mikayla-maki: Why does this stale thread ID cause the prompt store to fail initializing at all? Seems like the prompt store shouldn't depend on the active thread ID.

rtfeldman: Good catch — it doesn't! The stale thread ID and the PromptStore failure are completely independent. I've updated the PR to remove the stale-thread ceremony entirely. The test now just calls `NewThread` through the real `NativeAgentServer` path and verifies it produces a connected thread despite the PromptStore LMDB open failing.

mikayla-maki (approved):

PR #52342, opened 2026-03-24
editor: Go to file from multibuffer in multibuffer pane, not in the active pane

timvermeulen:
Fixes a bug that caused multibuffer go-to-file actions (clicking a file name in an excerpt header, (double) clicking a line number) to open the file in the active pane, not necessarily in the multibuffer pane itself. The after behavior is more consistent with e.g. cmd-click to Go to Definition in a pane that isn't the active one.

Before:

https://github.com/user-attachments/assets/0f9a0166-b318-4bdd-9b30-0f77fb150172

After:

https://github.com/user-attachments/assets/edf9c80e-c668-4afe-a343-80a6d2d03552

I went with the minimal change, but it might make sense to merge the `pane_item_id: Option<EntityId>` and `split: bool` parameters, since the former is only relevant when `split` is `false`.

## Self-Review Checklist

Release Notes:

- Fixed a bug that caused multibuffer go-to-file actions to open the file in the active pane, not necessarily in the multibuffer pane itself

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -24757,6 +24757,7 @@ impl Editor {
         Self::open_buffers_in_workspace(
             workspace.downgrade(),
             new_selections_by_buffer,
+            Some(cx.entity().entity_id()),
             split,
             window,
             cx,
@@ -24769,6 +24770,7 @@ impl Editor {
             Entity<language::Buffer>,
             (Vec<Range<BufferOffset>>, Option<u32>),
         >,
+        pane_item_id: Option<EntityId>,
         split: bool,
         window: &mut Window,
         cx: &mut App,
@@ -24782,7 +24784,9 @@ impl Editor {
                     let pane = if split {
                         workspace.adjacent_pane(window, cx)
                     } else {
-                        workspace.active_pane().clone()
+                        pane_item_id
+                            .and_then(|id| workspace.pane_for_item_id(id))
+                            .unwrap_or_else(|| workspace.active_pane().clone())
                     };
 
                     for (buffer, (ranges, scroll_offset)) in new_selections_by_buffer {
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -67,7 +67,7 @@ use util::{
 };
 use workspace::{
     CloseActiveItem, CloseAllItems, CloseOtherItems, MultiWorkspace, NavigationEntry, OpenOptions,
-    ViewId,
+    SplitDirection, ViewId,
     item::{FollowEvent, FollowableItem, Item, ItemHandle, SaveOptions},
     register_project_item,
 };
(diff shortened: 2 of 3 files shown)


dinocosta (commented):
Thank you for the suggestion @timvermeulen ! 🙂

Although the changes look good to me, I'm going to reach out to the team and see if they're interested in merging this as I personally prefere the current implementation, as it allows the user to pick either:

1. Keep cursor in multibuffer, file opens in the same pane
2. Keep cursor in different pane, file opens in the focused pane but multibuffer is still visible

With these changes that would no longer be possible, so I'm unsure whether we want this, but I'll see what the team has to say. Thanks!

timvermeulen:
@dinocosta Totally! Thanks for that perspective, it didn't really occur to me before that the current behavior could actually be preferable to some, that's a valid point. Either way, the inconsistency with cmd-click is probably worth addressing one way or another 🙂

dinocosta:
@timvermeulen Thank you for the understanding! 🙂

I did chat with the team and we've reached the conclusion that we'd rather keep the behavior as is, thank you for suggesting this anyway!

> Either way, the inconsistency with `cmd-click` is probably worth addressing one way or another 🙂

Good point! Just to double-check, do you mean this following behavior?

* Using `cmd-click` on a file header will open a split pane for the file
* Using `cmd-click` on the line number will open the file in the active pane
* Using `cmd-click` on the symbol, to go to definition, will always open the file in the multibuffer's pane

timvermeulen:
> Good point! Just to double-check, do you mean this following behavior?

Yeah the third one, the `Go to Definition` action, though I didn't mean specifically in a multibuffer. It was the only other action that I could think of besides multibuffer navigation where clicking within a pane causes you to navigate somewhere, although I suppose there's a few, depending on how you count:

- opening a file from a multibuffer using either the line number or excerpt header, i.e. what this PR is about
- cmd-click on a symbol for `Go to Definition` (or cmd-shift-click for `Go to Type Definition`)
- clicking on a tab to open it (not in the editor, but technically still in the pane)
- clicking on a sticky scroll header to jump to that line

These are the ones that come to mind but maybe there are more, and maybe it's debatable if these should all count as the same thing 😄 Anyway, the last three actions all do the navigation in the pane where the click happens, not necessarily in the pane that's currently active, so that's why I was a bit quick to assume that multibuffer navigation not working this way would be considered a bug.

I suspect Zed users would find it counterintuitive if any of the last three actions did the navigation in the currently active pane in case the click was done in a pane other than the currently active one, but frankly I don't know, I haven't asked anyone 🙂 To me these four actions are roughly the same in terms of in which pane I'd personally expect the navigation to happen, hence the comment about making them more consistent with each other. But if you consider multibuffer navigation to be meaningfully different from the others such that it's preferable for the four actions to not behave consistently in this regard, that would be totally fair too!

Special mention for opening a file by clicking on it in the Project Panel: that's also a click that causes the active pane to navigate somewhere, rather than the navigation happening where the click was. I didn't include it in the list because the Project Panel isn't an editor pane so the navigation couldn't possibly happen there, but still it's a noteworthy example of a navigation action triggered by a mouse click that always uses the active pane.

timvermeulen:
Closing this for now, but feel free to let me know your thoughts on the above if you have any!

dinocosta:
@timvermeulen Thank you for the write-up and sorry for the late reply, was out off office for the past month! 🙂

Reviewing the scenarios you shared, I think there's only two that I think are worth trying to update, namely, the difference in behavior between clicking and cmd-clicking a file header and a line number in a multibuffer.

Right now, this is the behavior I'm seeing:

* Using a `click` in the file header opens the file in the active pane
* Using a `cmd-click` in the file header opens the file in the pane to the right of the active one, creating a new pane if already at the right-most pane
* Using a `click` in the line number opens the line in the active pane
* Using a `cmd-click` in the line number opens the line in the active pane, so `cmd` has no effect on this one

If you're up for it, I believe it might be interesting to explore updating the latter case, when a user is holding `cmd` when the line number is clicked, in order to behave the same as the file header case. That is, if an user is holding `cmd` when they click a line number in a multibuffer, we'd open the file at that line in a pane to the right of the active one, just like in the case of a file header.

As for the other scenarios, here's my thoughts:

* The `editor: go to definition` action already requires the user hold `cmd` in order to be able to click, so there's no an alternative click version to consider
* For both tabs and sticky scroll headers, I don't think we want to consider even introducing a different behavior when the user is holding `cmd`

Hope that helps and let me know if you'd be interested in opening that Pull Request. Thanks! 🙂

PR #52347, opened 2026-03-24
Add hidden model filtering to agent model selectors

ljt019:
## Context

This adds support for hiding models from the agent model selectors. 

Zed already supports favoriting models, but there is not currently a way to reduce noise in the model list that largely shows old/less frequently used models. This change adds a 'hidden_models' setting for both built-in language models and ACP/custom agent servers, and filters those models out of the corresponding selectors.

This behavior is intentionally simple, this pr does not add any new ui and must be configured within the settings file, I'm open to adding UI for it if you guys want though :)

## Self-Review Checklist

## Release Notes:

- Added support for hiding models from agent model selectors via `hidden_models` settings.


benbrandt:
Hey thanks for reaching out but I think this adds some complexity I'd need to think about and the PR is older and we don't have a ton of people asking for this right now, so going to close

ljt019:
> Hey thanks for reaching out but I think this adds some complexity I'd need to think about and the PR is older and we don't have a ton of people asking for this right now, so going to close

Well let me know, I did my best to keep it up to date so it wasn't stale. I kept it out of adding any ui with the feature so it wouldn't complicate things much, it's almost an exact copy of the favorite code just hides instead, and I'm happy to make any adjustments if you change your mind in the future.

It's really frustrating seeing a huge wall of models I never use, and I know I've seen issues and discussions, albeit not super popular like you said asking for this feature.

PR #52350, opened 2026-03-24
agent: Mark subagent completions with Subagent intent

benbrandt:
## Context

Ensure subagent threads build requests with the Subagent intent instead of UserPrompt. This allows us to properly
attribute this as a tool call for certain providers instead of a user request.

## Self-Review Checklist

Release Notes:

- copilot_chat: Fix subagent requests being marked as user requests.

### DIFF
--- a/crates/cloud_llm_client/src/cloud_llm_client.rs
+++ b/crates/cloud_llm_client/src/cloud_llm_client.rs
@@ -197,6 +197,7 @@ pub enum EditPredictionRejectReason {
 #[serde(rename_all = "snake_case")]
 pub enum CompletionIntent {
     UserPrompt,
+    Subagent,
     ToolResults,
     ThreadSummarization,
     ThreadContextSummarization,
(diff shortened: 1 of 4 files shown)


crates/cloud_llm_client/src/cloud_llm_client.rs
```diff
@@ -197,6 +197,7 @@ pub enum EditPredictionRejectReason {
 #[serde(rename_all = "snake_case")]
 pub enum CompletionIntent {
     UserPrompt,
+    Subagent,
```
benbrandt, line 200: @maxdeviant do we need to change anything on your end?

maxdeviant: Yea, this will require changes in Cloud as well.

But it's rather incorrect that the Copilot Chat code has come to depend on this enum for its behavior.

We don't even use the `CompletionIntent` anymore (and it will go away at some point), so I think we should try to fix the Copilot Chat behavior away from relying on this.

benbrandt: Ok we'll need to thread _something_ through. Will look tomorrow with fresh eyes

maxdeviant (approved):

crates/cloud_llm_client/src/cloud_llm_client.rs
maxdeviant: I ended up removing this on the Cloud side, so we just ignore it now, if it is sent up.

PR #52359, opened 2026-03-24
cloud_llm_client: Move `CompletionIntent` to `language_model`

maxdeviant:
This PR moves the `CompletionIntent` enum from the `cloud_llm_client` crate to the `language_model` crate, as it is no longer part of the Cloud interface.

Release Notes:

- N/A


benbrandt (approved):
thanks!

PR #52365, opened 2026-03-24
theme_selector: Fix theme preview not updating on filter changes

### LINKED ISSUE #52118: Theme Selector: Search filtering triggers unintended preview theme switch
### Reproduction steps

1. Press `Ctrl-K Ctrl-T` to open the Theme Selector.
2. Select and confirm **Ayu Dark** as the active theme.
3. Reopen the Theme Selector via `Ctrl-K Ctrl-T` (with Ayu Dark currently active).
4. Type `Ayu l` into the search input field.
5. Observe the editor's theme preview immediately after the `l` keystroke filters the list.

**Expected Behavior**:
The theme preview should remain on the currently active theme (**Ayu Dark**) until the user explicitly selects a different item via arrow keys or mouse click. Search filtering alone should not trigger a preview switch.

**Actual Behavior**:
When typing `l` filters the list (and the current theme *Ayu Dark* no longer matches the query), the selector automatically highlights the first filtered result (e.g., **Ayu Light**), causing the editor's theme preview to switch prematurely—requiring the user to press `Esc` to revert to the original theme.

Zed: v0.224.0+dev.395.77df6bfef2d43f27a4202dfcb699e0e83f03c137 (Zed Dev 77df6bfef2d43f27a4202dfcb699e0e83f03c137) 
OS: Windows 10.0.26200
Memory: 15.7 GiB
Architecture: x86_64
GPU: Intel(R) UHD Graphics || Intel Corporation || 31.0.101.5333

Dnreikronos:
Fixes #52118

When a user filters themes in the theme selector and then confirms or clears the filter, the preview could show the wrong theme. Two issues were at play:

1. Confirming a selection after filtering (without navigating with arrow keys) could persist a stale theme because `show_selected_theme` was never called in `confirm`.
2. When a filter query matched no results (e.g., "Ayu 1"), `selected_theme` was overwritten with `None`. Clearing the filter then lost track of the previously selected theme and defaulted to index 0 (typically the first dark theme).

Both issues are fixed in `ThemeSelectorDelegate` and `IconThemeSelectorDelegate`.

## How to Review

Small PR — two files with symmetrical changes. Focus on the `update_matches` and `confirm` methods in:

- `crates/theme_selector/src/theme_selector.rs`
- `crates/theme_selector/src/icon_theme_selector.rs`

Key change: `selected_theme` is now only updated when `show_selected_theme` returns `Some`, preserving the previous selection when the filter yields no results.

## Self-Review Checklist

## Before:

https://github.com/user-attachments/assets/55da1c74-e2a9-4f64-9d35-3b46b3f973f0

## After:

https://github.com/user-attachments/assets/cf39933c-f119-4d9a-817d-d1af559666da

Release Notes:

- Fixed theme selector preview not updating correctly when filtering themes without navigating with arrow keys.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17585,12 +17585,15 @@ dependencies = [
 name = "theme_selector"
 version = "0.1.0"
 dependencies = [
+ "editor",
  "fs",
  "fuzzy",
  "gpui",
  "log",
  "picker",
+ "project",
  "serde",
+ "serde_json",
  "settings",
  "telemetry",
  "theme",
--- a/crates/theme/src/registry.rs
+++ b/crates/theme/src/registry.rs
@@ -124,6 +124,12 @@ impl ThemeRegistry {
         }
     }
 
+    /// Registers theme families for use in tests.
+    #[cfg(any(test, feature = "test-support"))]
+    pub fn register_test_themes(&self, families: impl IntoIterator<Item = ThemeFamily>) {
+        self.insert_theme_families(families);
+    }
+
     fn insert_themes(&self, themes: impl IntoIterator<Item = Theme>) {
         let mut state = self.state.write();
         for theme in themes.into_iter() {
--- a/crates/theme/src/registry.rs
+++ b/crates/theme/src/registry.rs
@@ -124,7 +124,7 @@ impl ThemeRegistry {
         }
     }
 
-    /// Registers theme families for use in tests.
+    /// Inserts theme families directly, bypassing extension loading.
     #[cfg(any(test, feature = "test-support"))]
     pub fn register_test_themes(&self, families: impl IntoIterator<Item = ThemeFamily>) {
         self.insert_theme_families(families);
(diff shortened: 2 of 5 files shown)


SomeoneToIgnore (changes requested):
https://github.com/zed-industries/zed/issues/52118#issuecomment-4106682241 says

> When the triage of that is done, I will send a new PR with the changes.

The triage is not done yet, what did I miss?

Overall, the issue seems to be about something else entirely, as it wants exactly the opposite: the filter change should not update the preview.

So, the PR definitely does not fix the issue and is wrong to label anything related as "fixed".

-----------------

I'm also not sure about part of the fix `test_theme_selector_updates_preview_on_filter` passes on `main` right now and I failed to break anything by filtering.

What's well spotted is the bug with the empty query: happy to take that bit only (and a corresponding test) out of the PR and move on, the rest (including the closes mention) should be stripped off then.

PR #52369, opened 2026-03-25
Fix interleaved subscription operations

mikayla-maki:
## Context

Since https://github.com/zed-industries/zed/pull/52232/, there have been a lot of problems related to panels randomly hiding and showing themselves. However, the code in that PR is pretty innocuous. It shouldn't be having that kind of effect. However, it turns out there's a bug in how GPUI subscriptions work that caused there to be orphaned observe callbacks firing spuriously. The core sequence is captured in `test_unsubscribe_during_callback_with_insert`, essentially you have to have two observers on the same item (`SettingsStore`, in this case), that both drop themselves. If the first observer _adds_ a callback onto the _same item_, then the second observer will never successfully drop itself. 

While in there, I also fixed an unrelated bug that @maxbrunsfeld noticed where if you have two callbacks on the same item, and an earlier callback drops a later one, then the second callback will spuriously fire.

I also added a few extra smoke tests to the subscription code, and a test capturing the observed bug at the workspace level.

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/subscription.rs
+++ b/crates/gpui/src/subscription.rs
@@ -75,16 +75,14 @@ where
                 };
 
                 if let Some(subscribers) = subscribers {
-                    subscribers.remove(&subscriber_id);
-                    if subscribers.is_empty() {
-                        lock.subscribers.remove(&emitter_key);
+                    if subscribers.remove(&subscriber_id).is_some() {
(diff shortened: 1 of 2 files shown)


maxbrunsfeld (commented):
Nice! Thanks for figuring this out.

PR #52374, opened 2026-03-25
Simplify sidebar internals

mikayla-maki:
## Context

Commit 1:  This commit simplifies the sidebar internals by removing state that could be derived live, and unifying the two separate concepts of "active" (draft or focused) with a new enum. I also renamed "focused" to "active", to match "active workspace" more than "keyboard focus".

Commit 2: I simplified the sidebar a bit more, by doing the thread metadata sorting in sqlite rather than on the sidebar.  Then I also simplified the implementation of the "archive" action by:
- Adjusting our derivation to cache the "parent" workspaces in the thread entries
- Caching a `Map<Session ID, ThreadEntryIndex>` so that we can look up the visual entries quickly
- Simplify the group traversal to only check the thread immediately below, then immediately above the current thread entry. This relies on the fact that all groups are visually bounded by project headers, so we we're guaranteed to either get a valid thread entry OR a non-entry.

Finally, I fixed the bug I set out to start with: once a thread deletion goes through, we drop the workspace it's associated with from the multiworkspace. I also added a test demonstrating this.

Subsequent Commits: Getting the PR ready for merging.

## Self-Review Checklist

- [ ] I've sufficient done manual testing

Release Notes:

- N/A

### DIFF
--- a/crates/sidebar/src/sidebar.rs
+++ b/crates/sidebar/src/sidebar.rs
@@ -654,12 +654,9 @@ impl Sidebar {
 
         let query = self.filter_editor.read(cx).text(cx);
 
-        // Derive active_entry from the active workspace's agent panel.
-        // Thread when the panel has a non-draft thread, Draft when the
-        // workspace has no active thread (or no panel at all).
-        // Only overwrite with Thread on a positive signal — if the panel
-        // transiently returns None while loading, keep the previous value
-        // so eager writes from user actions survive.
+        // Update active_entry to reflect the focused thread in the
+        // agent panel. If the panel has no focused thread, leave
+        // active_entry alone to preserve optimistic UI set by user actions.
         let panel_focused = active_workspace
             .as_ref()
             .and_then(|ws| ws.read(cx).panel::<AgentPanel>(cx))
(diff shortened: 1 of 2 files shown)


crates/sidebar/src/sidebar.rs
```diff
@@ -5134,12 +5200,11 @@ mod tests {
...
-        // ── 1. Initial state: focused thread derived from active panel ─────
+        // ── 1. Initial state: workspace_b is active (no threads) so Draft ────
         sidebar.read_with(cx, |sidebar, _cx| {
-            assert_eq!(
-                sidebar.focused_thread.as_ref(),
-                Some(&session_id_a),
-                "The active panel's thread should be focused on startup"
+            assert!(
+                sidebar.active_entry.as_ref().is_some_and(|e| e.is_draft()),
+                "Workspace B is active with no threads, so active_entry should be Draft"
```
mikayla-maki, line 5206: This assertion is a good example of why this refactor was needed. If the "focused thread" was for a non-active workspace, why is this value hanging around in the sidebar at all? We should always have perfect, up-to-date values after an update_entries.

PR #52381, opened 2026-03-25
workspace: Show file path in bottom bar

cameron1024:
Context: if the toolbar and tab bar are both disabled, the current filename is not visible. This adds it to the bottom bar, similar to vim. Behind a setting, disabled by default

Release Notes:

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1620,6 +1620,8 @@
   "status_bar": {
     // Whether to show the status bar.
     "experimental.show": true,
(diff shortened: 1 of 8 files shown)


danilo-leal:
I think it's simpler for now if they're all independent settings as there's maybe a use case for this outside from the one in which the toolbar and tab bar are disabled. We could also create next a setting that's a bit more Vim-branded, as removing these UI elements seem to usually be done as an effort to approximate Zed to a Vim UI-sort of experience?!

danilo-leal (approved):

PR #52382, opened 2026-03-25
editor: Replace distance-based hover persistence with safe triangle

### LINKED ISSUE #20355: Hover tooltip disappears too easily
### Check for existing issues

- [X] Completed

The overlay tooltip disappears much too easily, which makes it very hard (and sometimes impossible) to click or select within it.

See this video for an example of this problem:

https://github.com/user-attachments/assets/f36abba6-6e2d-43b0-aee2-7c2571a129f4

### Environment

Zed: v0.160.7 (Zed)
OS: macOS 15.1.0
Memory: 32 GiB
Architecture: aarch64

For inspiration on how to fix this, safe triangles are a nice way to go about it:

https://www.smashingmagazine.com/2023/08/better-context-menus-safe-triangles/

```

```
</details>

alikimovich:
## Context

Hover tooltips disappear too easily when moving the cursor toward them, making it hard or impossible to click or select within them.

The previous approach tracked the scalar distance between the cursor and the popover center. If the distance increased by more than 4px, it was treated as "moving away" and a 300ms hide timer started. This was unreliable — diagonal movement toward the popover could register as moving away depending on the angle.

This replaces it with a [safe triangle pattern](https://www.smashingmagazine.com/2023/08/better-context-menus-safe-triangles/). When the cursor leaves the trigger, a safe zone is defined as the convex hull of the last mouse position over the trigger (the apex) and the popover bounds, expanded by a 10px margin. If the cursor stays within this zone, the popover remains visible. This lets users move diagonally toward the popover without it disappearing.

Closes #20355

**before:**
https://github.com/user-attachments/assets/7d95e7c9-81ac-4542-b260-d930b19d5485

**after:**
https://github.com/user-attachments/assets/82c10c62-f31e-475a-b4cf-434f7fd96857

## How to Review

Focus on `crates/editor/src/hover_popover.rs`:

1. **`hover_at` function** — the control flow changed: safe zone check → cancel timer and return early; outside safe zone with existing timer → let it count down; outside safe zone with no timer → start 300ms timer
2. **`HoverState`** — `closest_mouse_distance: Option<Pixels>` replaced with `safe_triangle_apex: Option<gpui::Point<Pixels>>`
3. **`is_mouse_in_safe_zone`** — now `&self` instead of `&mut self` (no longer mutates state in a query method)
4. **`point_in_safe_zone` / `point_in_triangle` / `cross_2d`** — geometry helpers using barycentric sign test
5. **`InfoPopover` / `DiagnosticPopover` mouse_move handlers** — removed `closest_mouse_distance = Some(px(0.0))`, only `hiding_delay_task = None` is needed now

## Self-Review Checklist

Release Notes:

- Fixed hover tooltips disappearing when moving the cursor diagonally toward them

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -2236,7 +2236,6 @@ mod tests {
             size: gpui::size(px(300.0), px(100.0)),
         };
 
-        // Point on the diagonal path from apex toward the popover
         assert!(HoverState::point_in_safe_zone(
             gpui::point(px(120.0), px(80.0)),
             apex,
@@ -2252,7 +2251,6 @@ mod tests {
             size: gpui::size(px(300.0), px(100.0)),
         };
 
-        // Point far away from both apex and popover
         assert!(!HoverState::point_in_safe_zone(
             gpui::point(px(50.0), px(300.0)),
             apex,
@@ -2268,7 +2266,6 @@ mod tests {
             size: gpui::size(px(300.0), px(100.0)),
         };
 
-        // Point on the opposite side of the apex from the popover
         assert!(!HoverState::point_in_safe_zone(
             gpui::point(px(10.0), px(100.0)),
             apex,
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -2182,4 +2182,97 @@ mod tests {
             InlayOffset(MultiBufferOffset(104))..InlayOffset(MultiBufferOffset(108))
         );
     }
+
+    #[test]
+    fn test_point_in_triangle() {
+        let a = gpui::point(px(0.0), px(0.0));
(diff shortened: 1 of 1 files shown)


dinocosta (commented):
Hey @alikimovich 👋

Thank you for suggesting this change. Although I’ve mostly seen the safe triangle approach used in context menus, I can see how it could still be applicable in this scenario.

That said, it appears that https://github.com/zed-industries/zed/pull/47471￼ was recently merged, and it attempts to address the same issue. I don’t think we want two different systems trying to solve the same problem, so I’ll go ahead and close this pull request.

If users continue to report trouble interacting with diagnostic hover popovers, we could reconsider this approach as an alternative to the one introduced in ￼https://github.com/zed-industries/zed/pull/47471 🙂

PR #52393, opened 2026-03-25
Agent speech to text dictation feature

cxnmai:
I've added a speech to text dictation tool for the agent tool. I believe this belongs in zed for a few reasons:
1. speed. talking is way faster than typing. it's also very natural for prompting since LLMs are relatively conversational.
2. accessibility + convenience. vibe coding has made software development more accessible by closing knowledge gaps, but voice typing can assist users who struggle to use keyboards.
3. competitiveness. both cursor and claude code now have voice modes, so adding it to zed will assist in growing the community of zed users.
I have implemented it in a way that minimizes bloat for the regular user but keeps feature parity for those who choose to use it.
How it works:
By default, voice mode is disabled. I've added a setting in the agent panel menu:
image
Once enabled, the agent panel gains this minimal mic icon with the dropdown menu:
image
However right now it's grayed out as it's not configured yet:
image
Now in settings under "agent":
    "speech_to_text": {
          "enabled": true,
          "whisper_cpp_executable_path": "{your binary path}",
          "whisper_cpp_model_path": "{your model path}",
          "trigger_mode": "hold",
    },
The transcription powered by [whisper.cpp](ttps://github.com/ggml-org/whisper.cpp) running locally. By having the user install it and set the path themselves, it ensures that users who don't want the feature don't have to install a whole voice dictation model along with Zed.

Once that's in, it stops being gray and will work when used. you can hit the dropdown menu to select a mic (by default set to System Default):
image

The user can set between hold and toggle options in settings for the mouse. There is also a keybind for toggling by default set to ctrl + alt + v.
image

While active, there's this flashing dot indicator on the mic icon:
image

While whisper is processing it, it has this loading buffer icon:
image

Then when you're done speaking, the text is inserted wherever your cursor is in the text box.

The flow of processing is
toggle on = start recording -> toggle off = stop recording -> saves .wav file to an audio folder in zed's set caching directory -> whispr processes it into text -> text inserts into agent panel -> audio file is deleted.

I encourage you to try it out yourself in the fork, it's pretty fun to talk to the LLM! This is also my first time contributing to zed so I would appreciate feedback.
Also - I understand for features it's recommended I start a discussion first. I opted to just do this since I had the free time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -418,6 +418,7 @@ dependencies = [
  "util",
  "uuid",
  "watch",
+ "which 6.0.3",
  "workspace",
  "zed_actions",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -372,6 +372,7 @@ dependencies = [
  "language_model",
  "language_models",
  "languages",
+ "livekit_client",
  "log",
  "lsp",
  "markdown",
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -332,6 +332,7 @@
       // `alt-l` is provided as an alternative to `alt-tab` as the latter breaks on Linux under the `AcpThread > Editor` context
       "alt-l": "agent::CycleFavoriteModels",
       "ctrl-;": "agent::OpenAddContextMenu",
+      "ctrl-alt-v": "agent::ToggleVoiceRecording",
       "ctrl-alt-k": "agent::ToggleThinkingMode",
       "ctrl-alt-'": "agent::ToggleThinkingEffortMenu",
       "ctrl-'": "agent::CycleThinkingEffort",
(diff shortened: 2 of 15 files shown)


benbrandt:
thanks for this, we've discussed this feature a lot internally, I think we will handle this in a different way though. thanks!

chimbori:
I think someone who took the time to provide a full PR deserves at least a little bit more justification than “we will handle this in a different way” kthxbye.

Zed is a great editor, and glad to see the community enthusiasm behind it, but I hope the maintainers and the company are able to sustain and harness this better.

theguy000:
> thanks for this, we've discussed this feature a lot internally, I think we will handle this in a different way though. thanks!

what are your plans and when is the voice feature will be expected to be added? this is a very much needed feature so idk why this was rejected?

PR #52394, opened 2026-03-25
go: Add semantic token rule for format string interpolations

mfattakhov:
## Context

Closes #52391

gopls sends semantic tokens for format args with [format] modifier. Without a simple support for there is no special highlighting for format args like `%s`. Its a simple change inside `go/semantic_token_rules.json` that makes this a default behavior.

Currently (on 0.229 preview):
Screenshot 2026-03-25 at 11 34 57

After fix:

Screenshot 2026-03-25 at 11 58 37

## Self-Review Checklist

Release Notes:

- gopls: Add semantic token rule for format string interpolations


Veykril (approved):
Thanks!

PR #52401, opened 2026-03-25
agent_ui: Add agent connection restart controls

benbrandt:
Track agent connection status in the configuration UI, show a
restart action for connected custom agents, and only render the
External Agents menu section when entries exist.

image

Release Notes:

- acp: Allow for restarting agent servers from the Agent Settings panel.


danilo-leal (approved):
Increased the scope of the PR just a bit by adding a component that we can share between the external agents and MCP sections of the settings view!

Image

PR #52404, opened 2026-03-25
Hide cursor in embedded terminal when not focused

anaslimem:
## Context
Fixes #52063
This change hides the cursor in embedded terminal mode when not focused. Embedded mode is used for read-only terminal output (like the Agent panel). Showing a cursor in a read-only context when unfocused is confusing, so we suppress it.
Screenshot 2026-03-25 at 12 03 15

## How to Review

The change is in a single file: `crates/terminal_view/src/terminal_view.rs:754-761`. 
Focus on the `should_show_cursor()` method 
Verify the logic correctly hides the cursor only when both conditions are met (Embedded mode AND not focused).

## Self-Review Checklist

- [x] Unsafe blocks (if any) have justifying comments (N/A - no unsafe code)
- [x] Tests cover the new/changed behavior (behavior is minimal UI fix, existing tests should cover)
- [x] Performance impact has been considered and is acceptable (negligible)

Release Notes:

- Fixed cursor visibility issue in embedded terminal panels

### DIFF
--- a/crates/terminal_view/src/terminal_view.rs
+++ b/crates/terminal_view/src/terminal_view.rs
@@ -761,7 +761,7 @@ impl TerminalView {
             }
         }
 
-        // Always show cursor when not focused or in special modes
+        // For Standalone mode: always show cursor when not focused or in special modes
         if !focused
             || self
                 .terminal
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #52411, opened 2026-03-25
workspace: Fix empty workspaces deleted by recent_workspaces_on_disk

masak1yu:
Fixes #52038

## Problem

Empty workspaces (a Zed window with no folder open, containing only unsaved buffers) are saved to the database with `paths = ""`. On the next startup, `recent_workspaces_on_disk` is called early — via `HistoryManager::init` — and unconditionally **deletes** these workspaces.

The bug is in this check:

```rust
if !has_wsl_path
    && Self::all_paths_exist_with_a_directory(paths.paths(), fs, Some(timestamp)).await
{
    result.push(...);
} else {
    delete_tasks.push(self.delete_workspace_by_id(id)); // ← empty workspace deleted here
}
```

`all_paths_exist_with_a_directory` receives an empty slice, its loop never runs, `any_dir` stays `false`, and the workspace is queued for deletion.

This runs concurrently with session restoration. By the time restoration calls `workspace_for_id(N)`, the row is gone, producing:

```
ERROR [zed] Failed to restore workspace: Workspace WorkspaceId(N) not found
ERROR [zed] All workspace restorations failed. Opening fallback empty workspace.
```

The user's unsaved buffers are lost even though `restore_unsaved_buffers` is enabled.

## Fix

Skip empty-path workspaces in `recent_workspaces_on_disk`. They are identified by workspace ID rather than paths and are restored via `session_workspaces` — they must not appear in the "recent projects" list, and must not be deleted by this function.

```rust
// Empty workspaces (no folder paths) are identified by workspace_id, not paths.
// They must not be deleted here; they are restored separately via session_workspaces.
if paths.is_empty() {
    continue;
}
```

A regression test is included that saves an empty workspace with a session ID, calls `recent_workspaces_on_disk`, and asserts the workspace row and its session entry still exist afterwards.

Release Notes:

- Fixed `restore_unsaved_buffers` not restoring unsaved file contents when Zed was closed with an empty window (no folder open). The workspace was being incorrectly deleted during startup, causing a "Workspace not found" error and falling back to a blank window.

### DIFF
--- a/crates/workspace/src/persistence.rs
+++ b/crates/workspace/src/persistence.rs
@@ -1853,6 +1853,12 @@ impl WorkspaceDb {
                 continue;
             }
 
+            // Empty workspaces (no folder paths) are identified by workspace_id, not paths.
+            // They must not be deleted here; they are restored separately via session_workspaces.
+            if paths.is_empty() {
+                continue;
+            }
+
             // Delete the workspace if any of the paths are WSL paths. If a
             // local workspace points to WSL, attempting to read its metadata
             // will wait for the WSL VM and file server to boot up. This can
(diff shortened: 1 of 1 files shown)


echuber2:
Would https://github.com/zed-industries/zed/pull/52810 conflict with this in any way?

masak1yu:
@echuber2 Thanks for the heads up! I checked #52810 and I don't think they conflict:

- **Git conflict**: No — this PR modifies `recent_workspaces_on_disk` while #52810 modifies `save_workspace`, so they touch different functions.
- **Logic conflict**: No — the two fixes are complementary. #52810 ensures empty workspaces are saved with `NULL` (instead of an empty string) in `paths`/`paths_order`, while this PR prevents `recent_workspaces_on_disk` from deleting empty workspaces during its cleanup. Both together would make empty workspace handling more robust.

echuber2:
Thanks for your comments. Can you rebase on the latest main and check the tests locally? (I don't have the ability to start a workflow run.) On my own PR, it was merged successfully but later reverted because new tests failed, despite it previously passing; you'll want to make sure you're rebased and passing on the latest master. I'd also like to resubmit my PR but judging from the new tests they added, it might be best if your PR was merged first.

maxbrunsfeld:
I think this has already been fixed by https://github.com/zed-industries/zed/pull/54224

masak1yu:
Thank you @maxbrunsfeld! I confirmed that #54224 includes the same fix — `recent_workspaces_for_ui` now skips workspaces with empty paths via the `paths.paths().is_empty()` check, which addresses the root cause described here. Closing this PR as a duplicate.

PR #52423, opened 2026-03-25
sidebar: Switch with `ctrl-tab`

cameron1024:
Overrides `ctrl-tab` when the sidebar/agent panel is focused to switch between recently viewed threads

Release Notes:


crates/sidebar/src/sidebar.rs
```diff
@@ -118,6 +122,8 @@ struct ThreadEntry {
     worktree_full_path: Option<SharedString>,
     worktree_highlight_positions: Vec<usize>,
     diff_stats: DiffStats,
+    last_accessed_at: Option<DateTime<Utc>>,
+    last_message_sent_or_queued: Option<DateTime<Utc>>,
```
maxbrunsfeld: I see we *also* need to store the timestamps outside of the `ThreadEntry`, in separate hash maps on the `Sidebar`. Since that's the case, do we need to *also* store them here, or should we just use the hash maps exclusively?

maxbrunsfeld (commented):
Would it be worth adding a unit test asserting the new thread ordering? In particular, you mentioned that subtlety about enqueuing messages. It might be worth asserting that the threads stay in the expected order in that case.

I left one other implementation comment. Neither of these are necessarily blocking - just suggestions.

crates/sidebar/src/sidebar.rs
cameron1024: Oh yeah this was just old implementation code I forgot to clean up - removing it

maxbrunsfeld (approved):

PR #52431, opened 2026-03-25
Sidebar and git cleanups

eholk:
A few minor cleanups in sidebar.rs and some related areas.

Release notes:

- N/A

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -3706,6 +3706,23 @@ impl RepositorySnapshot {
         }
     }
 
+    /// The main worktree is the original checkout that other worktrees were
+    /// created from.
+    ///
+    /// For example, if you had both `~/code/zed` and `~/code/worktrees/zed-2`,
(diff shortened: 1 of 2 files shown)


crates/project/src/git_store.rs
```diff
@@ -3706,6 +3706,23 @@ impl RepositorySnapshot {
...
+        self.work_directory_abs_path == self.original_repo_abs_path
+    }
+
+    /// Returns true if this repository is a linked worktree, that is, one that
+    /// was created from another worktree.
+    ///
+    /// This is by definition the opposite of [`Self::is_main_worktree`].
+    pub fn is_linked_worktree(&self) -> bool {
+        !self.is_main_worktree()
+    }
```
Anthony-Eid, line 3724: We should double check if this is true for sub modules as well. I have a feeling `is_main_worktree` might have some false negatives in that case. Because abs_path could be different from the original repo abs path

Anthony-Eid (approved):
Nice clean up!

PR #52440, opened 2026-03-25
agent_ui: Make thread generation top-down

danilo-leal:
Ever since we introduced the agent panel the way it is right now back in May 2025, we wanted to make the thread generation be top-down. Now, with the changes we're planning to launch very soon revolving around parallel agents, doing this become even more important for a better experience. Particularly because in the agent panel's new empty state, the message editor is full screen. We want to minimize as much as possible layout shift between writing your first prompt and actually submitting it. So this means that content will stream down from your first prompt and auto-scroll you if it goes beyond the viewport.

To pull this off, we added a `follow_tail` feature directly to the GPUI list so that we could only call it in the thread view layer as opposed to doing it all there.

https://github.com/user-attachments/assets/99961819-6a79-40e0-b482-dca68c829161

Release Notes:

- Agent: Made the thread generation be top-down instead of bottom-up. Agent content now streams from the top and auto-scroll as they go beyond the viewport.


crates/gpui/src/elements/list.rs
```diff
@@ -958,6 +1004,8 @@ impl StateInner {
             content_height - self.scrollbar_drag_start_height.unwrap_or(content_height);
         let new_scroll_top = (point.y - drag_offset).abs().max(px(0.)).min(scroll_max);
 
+        self.follow_tail = false;
```
rtfeldman, line 1007: @danilo-leal I found a couple of edge cases this was missing and I just pushed fixes instead of suggestions 😄 

This one was for the case where you're dragging the scroll bar around instead of using a `ScrollWheel` event; I assume we want to disengage `follow_tail` there too!

rtfeldman (approved):
Excited to use this! 😃

nsxdavid:
Yeah this is a pretty awful change for anyone who considers this the sort of thing you fight against in bad implementations of a scrolling stream.  (sigh)

PR #52441, opened 2026-03-25
call: Update call location when active multi workspace changes

Anthony-Eid:
## Context

This fixes a participant location out of sync bug when the active workspace changes in a multi workspace. It wouldn't update the project id, which breaks following.

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/crates/call/src/call_impl/mod.rs
+++ b/crates/call/src/call_impl/mod.rs
@@ -17,8 +17,8 @@ use room::Event;
 use settings::Settings;
 use std::sync::Arc;
 use workspace::{
-    ActiveCallEvent, AnyActiveCall, GlobalAnyActiveCall, Pane, RemoteCollaborator, SharedScreen,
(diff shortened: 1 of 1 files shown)


crates/call/src/call_impl/mod.rs
```diff
@@ -28,6 +28,41 @@ use crate::call_settings::CallSettings;
...
+                }
+
+                let project = multi_workspace
+                    .read(cx)
+                    .workspace()
+                    .read(cx)
+                    .project()
+                    .clone();
+                if let Ok(task) = active_call_handle.update(cx, |active_call, cx| {
+                    active_call.set_location(Some(&project), cx)
```
mikayla-maki, line 51: Does this need to be from the set of shared workspaces? I think if you move to a non-shared workspace you should set the location to None.

mikayla-maki (approved):

PR #52447, opened 2026-03-25
Add browser

nihalxkumar:
## Context

This PR proposes a lightweight browser tab using the system webview via https://github.com/tauri-apps/wry

The webview is composited by the OS as a native child view inside GPUI's window.

Related: https://github.com/zed-industries/zed/discussions/39885 https://github.com/zed-industries/zed/discussions/51251 Ialsopersonallyvote

---

Demo:

https://github.com/user-attachments/assets/4377f8e0-1a4a-4885-8cf5-964c9e35bff5

The experience is a bit jittery atm and it has to be improved.

The global menu at the start(cherry-picked for demo from https://github.com/zed-industries/zed/pull/51321) is important because with the editor-menu some of the dropdown options hide behind the webpage. As a workaround, I experimented with hiding the webview when dropdowns were opened, but this approach is not ideal. AFAIU there is no way to have both of them co-exist and give a higher z-index to dropdown options.

The demo was recorded on EndeavourOS KDE X11. It should work on MacOS and Windows as well. 

Wayland support would require GPUI to expose subsurface support so the webview's surface could be placed within Zed's window bounds.

Audio also works on the browser tab but was intentionally stripped out from the recording.

The lines of code may appear large, ~1k is just boilerplate: auto generated cargo.toml config only because this is adding a new feature

---

## Self-Review Checklist

Release Notes:

- Added web view

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -166,7 +166,7 @@ dependencies = [
  "context_server",
  "ctor",
  "db",
- "derive_more",
+ "derive_more 2.1.1",
  "editor",
  "env_logger 0.11.8",
  "eval_utils",
@@ -235,7 +235,7 @@ dependencies = [
  "anyhow",
  "async-broadcast",
  "async-trait",
- "derive_more",
+ "derive_more 2.1.1",
  "futures 0.3.31",
  "log",
  "serde",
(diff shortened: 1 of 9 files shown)


esthertrapadoux:
Closing this PR since it requires more discussion. But really appreciate you wanting to lean in here @nihalxkumar

PR #52454, opened 2026-03-25
terminal: Resolve path hyperlinks using per-line cwd

### LINKED ISSUE #52325: cmd-click on filename in terminal panel doesn't account for terminal cwd
### Reproduction steps

1. Start zed
2. create multiple files with the same name but in different folders from the terminal

```bash
mkdir -p folder1/
touch test.txt
touch folder1/test.txt
```

3. navigate to folder1/

```
cd folder1/
ls    # this outputs "test.txt"
```

now cmd+click on `test.txt` (opens `ROOT/test.txt` instead of `ROOT/folder1/test.txt`)

related to https://github.com/zed-industries/zed/issues/33498

cmd+click on `test.txt` opens `ROOT/test.txt` instead of `ROOT/folder1/test.txt` when terminal is in `folder1/` folder

Zed 0.228.0 
8421009ef8a022df1196d54bb42fd94366ec0988
0.228.0+stable.203.8421009ef8a022df1196d54bb42fd94366ec0988

MacOS

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

saberoueslati:
## Context

When cmd/ctrl-clicking a relative path in the terminal, Zed was resolving it against the *current* working directory at click time, not the directory that was active when the line was printed. So if you run command `cd folder1` after running command `ls`, clicking `test.txt` in the old `ls` output would open `ROOT/test.txt` instead of `ROOT/folder1/test.txt`.

The approach is similar to the VSCode approach for hyperlinks in the terminal

**Video below of the test on VSCode:**

[Screencast from 2026-03-25 23-28-26.webm](https://github.com/user-attachments/assets/d1c3b163-7b18-43db-9981-71c1cb09a801)

**Video below of the test on Zed after the fix:**

[Screencast from 2026-03-25 23-31-49.webm](https://github.com/user-attachments/assets/fc23da34-dd16-4cce-afed-53f681762152)

Closes #52325

## How to Review

Two files changed — read in this order:

1. **`crates/terminal/src/pty_info.rs`** Bug fix: `emit_title_changed_if_changed` was comparing cwd with itself because `load()` writes to `self.current` *before* the comparison. Fixed by capturing `prev_cwd`/`prev_name` before calling `load()`. Also extended the background task to return the new cwd when it changes, so the foreground task can call `terminal.record_cwd_change(cwd)`.

2. **`crates/terminal/src/terminal.rs`** 
   - `cwd_history: Vec<(i32, PathBuf)>` field on `Terminal`, seeded with `(i32::MIN, initial_cwd)` for PTY terminals so all pre-history lines resolve correctly.
   - `record_cwd_change` called when the PTY cwd changes; records the current scrollback position alongside the new path.
   - `cwd_at_line` given a scrollback line, binary-searches the history to find the most recent cwd active at or before that line.
   - `process_hyperlink` updated to use `cwd_at_line` for both plain paths and `file://` URLs.

## Self-Review Checklist

Release Notes:

- Fixed terminal path hyperlinks resolving against the current working directory instead of the one active when the line was printed


saberoueslati:
@ChristopherBiscardi can you run the CI/CD pipeline for this PR ? Thank you

saberoueslati:
@ChristopherBiscardi I'll fix the style issue this evening

saberoueslati:
@ChristopherBiscardi it appears the style test failed because error

`Run crate-ci/typos@2d0ce569feab1f8752f1dde43cc2f2aa53236e06
Run $GITHUB_ACTION_PATH/action/entrypoint.sh
Downloading 'typos' v1.40.0
--2026-04-28 07:00:28--  https://github.com/crate-ci/typos/releases/download/v1.40.0/typos-v1.40.0-x86_64-unknown-linux-musl.tar.gz
Resolving github.com (github.com)... 140.82.112.4
Connecting to github.com (github.com)|140.82.112.4|:443... connected.
HTTP request sent, awaiting response... 502 Bad Gateway or Proxy Error
2026-04-28 07:00:28 ERROR 502: Bad Gateway or Proxy Error.`

So it's not connected to the code at all, I'm sure running the test again will probably fix it, thank you

ChristopherBiscardi:
yep, it passed after an extra two runs. Was a gateway issue fetching a tar.

crates/terminal/src/terminal.rs
```diff
@@ -883,6 +888,7 @@ pub struct Terminal {
     event_loop_task: Task<Result<(), anyhow::Error>>,
     background_executor: BackgroundExecutor,
     path_style: PathStyle,
+    cwd_history: Vec<(i32, PathBuf)>,
```
Veykril: comment as to what the i32 here stands for would be nice

crates/terminal/src/terminal.rs
```diff
@@ -1235,15 +1248,15 @@ impl Terminal {
 
                 MaybeNavigationTarget::PathLike(PathLikeTarget {
                     maybe_path: decoded_path,
-                    terminal_dir: self.working_directory(),
+                    terminal_dir,
```
Veykril: Probably makes sense to rename the field to working_directory?

crates/terminal/src/terminal.rs
```diff
@@ -2152,6 +2165,26 @@ impl Terminal {
         }
     }
 
+    pub(crate) fn record_cwd_change(&mut self, new_cwd: PathBuf) {
+        let term = self.term.lock_unfair();
+        let pos = term.history_size() as i32 + term.grid().cursor.point.line.0;
+        drop(term);
+        self.cwd_history.push((pos, new_cwd));
+    }
```
Veykril: Is `history_size` guaranteed to be always increasing?

crates/terminal/src/terminal.rs
saberoueslati: I replaced the tuple with a `CwdHistoryEntry` struct, renamed the `i32` field to `scrollback_position`, and added a field comment explaining that it is the line offset in the retained scrollback buffer.

crates/terminal/src/terminal.rs
saberoueslati: Renamed `terminal_dir` to `working_directory` on `PathLikeTarget` and updated the terminal_view call sites and tests accordingly.

crates/terminal/src/terminal.rs
saberoueslati: It isn't. `history_size` grows until it hits the scrollback cap, then stays constant while old lines are evicted. I extracted the coordinate computation into a `scrollback_position` helper with a comment noting that this is a coordinate in the retained buffer rather than a global line number, and made the arithmetic overflow-safe.

One thing worth flagging: once the cap is hit, recorded positions can drift relative to the lines they were tagged with (`line.0` keeps decreasing as retained lines move upward, while `history_size` stays capped). So in very long sessions, lookup can return the wrong cwd for older lines that are still in scrollback. Fixing that would mean either tracking a monotonic "lines-ever-written" counter and translating on lookup, or hooking the trim path to shift/prune `cwd_history` entries. Happy to do either if you'd prefer stricter bookkeeping here, otherwise, I think the current scoped version is reasonable.

Veykril: I think im fine with having that behavior as long as we note it down in source somewhere.

saberoueslati: I'll add a comment and take care of the CI/CD issues later today, thank you

saberoueslati: @Veykril Documented the scrollback-cap drift limitation directly on the `scrollback_position` helper, explaining that recorded positions can drift once the cap is reached and how a stricter version (monotonic lines-ever-written counter, or pruning/shifting `cwd_history` on trim) would work.

PR #52455, opened 2026-03-25
Gemini CLI sandbox startup

### LINKED ISSUE #51333: Gemini CLI external agent does not start when using Gemini's sandbox
Note: This is a reposting of #43490 which was previously closed but is still not resolved (I don't have permission to reopen that existing issue)

### Reproduction steps

Edit gemini settings file (.gemini/settings.json) to configure its sandbox. tools.sandbox can be set to true, docker, or sandbox-exec (macOS only for this option), they all will fail in the subsequent steps.
Start a "New Gemini CLI" external agent thread in Zed

Expected behavior: A new Gemini CLI agent thread starts

Current behavior: The new thread is stuck in the "Loading..." state. ACP logs show an initial initialize request, but no activity after that.

OS: macOS 15.0
Memory: 18 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>ACP logs</summary>

```log
Initialize:
{
  "protocolVersion": 1,
  "clientCapabilities": {
    "fs": {
      "readTextFile": true,
      "writeTextFile": true
    },
    "terminal": true,
    "_meta": {
      "terminal_output": true,
      "terminal-auth": true
    }
  },
  "clientInfo": {
    "name": "zed",
    "title": "Zed",
    "version": "0.221.5+stable.141.03bfbf242c57f7fdb45708d63d74182898edf2c5"
  }
}

(nothing is logged after the above)
```

</details>

Gemini CLI (via ACP), version 0.33

None

loadingalias:
Closes #51333

## Summary
• Add a timeout for external ACP agent initialization so stalled startups fail with an error instead of hanging forever.
• Kill the child process if ACP init times out.
• Surface a targeted Gemini hint when startup stalls with sandboxing enabled.
• Propagate `test-support` through `acp_tools` so the `agent_servers` validation graph compiles cleanly - this was required to run the tests/checks locally, unless I'm missing something?

## Root Cause
Gemini CLI hangs before ACP initialization when its own sandboxing is enabled and stdin is piped.

Local repro showed:
• with `tools.sandbox = false`, Gemini responds to ACP initialize over piped stdio
• with `tools.sandbox = "sandbox-exec"`, Gemini never responds to ACP initialize
• this reproduces both with Zed’s shell-wrapped launch shape and with a direct piped launch

So, Zed’s shell wrapper isn't the root cause here, IMO.

This is an upstream issue in the Gemini CLI startup. When sandboxing is enabled and stdin is not a TTY, Gemini reads stdin before entering ACP mode. In ACP mode stdin is the protocol transport, so that waits for EOF forever and Gemini never starts its ACP server.

Zed still owns the UX here, so this change makes the failure explicit instead of leaving the thread in `Loading...`.

## Verification
• `./script/check-keymaps`
• `./script/clippy -p agent_servers`
• `cargo test -p agent_servers -- --nocapture`

## Manual Testing
• Repro’d Gemini ACP startup locally with Gemini CLI `0.35.0`
• Confirmed `initialize` succeeds with `tools.sandbox = false`
• Confirmed `initialize` times out with `tools.sandbox = "sandbox-exec"`
• Confirmed the timeout reproduces both with a direct piped launch and a Zed-like shell-wrapped piped launch

Release Notes:

 - Fixed external agent startup to fail with an error instead of hanging forever when ACP init stalls.

### DIFF
--- a/crates/acp_tools/Cargo.toml
+++ b/crates/acp_tools/Cargo.toml
@@ -5,6 +5,8 @@ edition.workspace = true
 publish.workspace = true
 license = "GPL-3.0-or-later"
 
+[features]
+test-support = ["workspace/test-support"]
 
 [lints]
 workspace = true
--- a/crates/agent_servers/Cargo.toml
+++ b/crates/agent_servers/Cargo.toml
@@ -6,7 +6,7 @@ publish.workspace = true
 license = "GPL-3.0-or-later"
 
 [features]
-test-support = ["acp_thread/test-support", "gpui/test-support", "project/test-support", "dep:env_logger", "client/test-support", "dep:gpui_tokio", "reqwest_client/test-support"]
+test-support = ["acp_thread/test-support", "acp_tools/test-support", "gpui/test-support", "project/test-support", "dep:env_logger", "client/test-support", "dep:gpui_tokio", "reqwest_client/test-support"]
 e2e = []
 
 [lints]
@@ -59,6 +59,7 @@ libc.workspace = true
 nix.workspace = true
 
 [dev-dependencies]
+acp_tools = { workspace = true, features = ["test-support"] }
 client = { workspace = true, features = ["test-support"] }
 env_logger.workspace = true
 fs.workspace = true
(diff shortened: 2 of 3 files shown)


benbrandt:
thanks for this, but I think we'll need a different solution. I tried this at one point, and for npm agents specifically on some people's networks or folks with mcp servers we may kill a perfectly valid agent with this approach

PR #52456, opened 2026-03-25
Settings profile base option

JosephTLyons:
## Context

This PR introduces a `base` field for settings profiles to allow profiles to either overlay `user` settings or to overlay `default`, which is simply zed's defaults (user settings are skipped). I'm not entirely sure I love `default` because it's a bit confusing (there's a setting called `default` but the default is `user`). Another idea I had was `factory` (`user` (default) or `factory`) - curious to hear from the reviewers. This will be useful for those of us who need to quickly flip to a default state, or a default state with some customizations on top. Additionally, from what I can tell, VS Code's profile system is more in line with what this PR is offering in Zed - profiles overlay the default settings, not the user's customization layer. So this will be familiar for those users.

I've had no issue with the migrator, code is pretty simple there, but would love for @smitbarmase to review the migration to make sure I'm not missing something.

## Self-Review Checklist

Release Notes:

- Improved the flexibility of settings profiles by offering a way for profiles to lay atop of zed defaults, skipping user settings all together. Settings Profiles now take the following form.

```json5
"Your Profile": {
  "base": "user" // or "default"
  "settings": {
    // ...
  },
},
```


crates/settings/src/settings_store.rs
```diff
@@ -1199,10 +1199,20 @@ impl SettingsStore {
...
             merged.merge_from_option(self.global_settings.as_deref());
             if let Some(user_settings) = self.user_settings.as_ref() {
-                merged.merge_from(&user_settings.content);
-                merged.merge_from_option(user_settings.for_release_channel());
-                merged.merge_from_option(user_settings.for_os());
-                merged.merge_from_option(user_settings.for_profile(cx));
+                let active_profile = user_settings.for_profile(cx);
+                let include_user_settings = active_profile
+                    .map(|p| p.base == ProfileBase::User)
+                    .unwrap_or(true);
```
MrSubidubi:
```suggestion
                    .is_none_or(|profile| profile.base == ProfileBase::User);
```

Nit but I think this would make this a bit more understandable at a glance.

MrSubidubi (approved):
As for the naming of `base`: An idea that came to mind is that we could instead name that enum `UserSettings`, `DefaultSettings` for it to be more clear. Another idea would be something like `"application"` which could then be `accumulative` or `exclusive`, naming very much to be discussed (and still think I like your current approach more).

crates/settings/src/settings_store.rs
JosephTLyons: I agree, that is a readability improvement. Thank you!

PR #52458, opened 2026-03-26
agent_ui: Fix panic when multiple rewrite tool uses share a char offset counter

mre:
## Context

The process_tool_use closure used a single chars_read_so_far counter shared across all REWRITE_SECTION_TOOL_NAME tool use events. When the LLM produced multiple separate rewrite tool uses (different IDs), the counter from the first tool use carried over into the second. If the second tool use had a shorter replacement_text, the slice indexing panicked with index out of bounds.

(This caused my first ever Zed panic today.)

Fix by tracking bytes-read per tool use ID using a HashMap keyed on LanguageModelToolUseId. Also replace the direct slice index with .get() so any remaining out-of-bounds case is handled gracefully instead of panicking.

Looks like it hasn't been reported yet. At least I couldn't find any related issues.

## How to Review

I hope the fix is pretty self-explanatory. We keep track of multiple rewrite tools, which explains the extra ceremony around getting the right character count, but apart from that it's pretty straightforward.

The performance impact should be neglible because:

1. Once the completion stream ends (or the task is dropped/cancelled), the entire Arc<Mutex<HashMap>> is freed.
2. The number of entries is tiny. In a single completion request the LLM will produce at most a handful of 
REWRITE_SECTION_TOOL_NAME tool uses. Each entry is just an Arc<str> key + usize value.

So the HashMap grows for the duration of one handle_completion call and is then dropped wholesale.

## Self-Review Checklist

Release Notes:

- Fixed an out-of-bounds panic when the AI produced multiple inline rewrites in a single completion.

### DIFF
--- a/crates/agent_ui/src/buffer_codegen.rs
+++ b/crates/agent_ui/src/buffer_codegen.rs
@@ -1,7 +1,7 @@
 use crate::{context::LoadedContext, inline_prompt_editor::CodegenStatus};
 use agent_settings::AgentSettings;
 use anyhow::{Context as _, Result};
-use collections::HashSet;
+use collections::{HashMap, HashSet};
 use editor::{Anchor, AnchorRangeExt, MultiBuffer, MultiBufferSnapshot, ToOffset as _, ToPoint};
 use futures::FutureExt;
 use futures::{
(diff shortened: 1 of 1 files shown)


benbrandt (approved):
Looks great, thank you!

mre:
Sorry for letting this branch go stale. 😅 Can you nudge the pipeline once more?

benbrandt (approved):

PR #52460, opened 2026-03-26
agent_ui: Display an activity bar for subagents waiting for permission

danilo-leal:
Closes https://github.com/zed-industries/zed/issues/52346

Given the parallel nature of subagents calls, it's possible that there is a subagent way out of view that's waiting for the user to give permissions. Right now, it's kind of hard to know this and you may think something wrong is happening given the thread generation isn't making any progress.

This PR adds an "activity bar" to the thread view that displays subagents on a "waiting for confirmation" status. We display the subagent's summary label as well as allow clicking on it to quickly scrolling to that subagent.

Screenshot 2026-03-25 at 10  09@2x

Release Notes:

- Agent: Improved the experience of interacting with subagents waiting for confirmation.


mikayla-maki (approved):

benbrandt (approved):
Any chance we can make this generic to any tool call that might be pending and scroll to it?

PR #52462, opened 2026-03-26
agent_ui: Move fully complete plan to the thread view

danilo-leal:
When a plan generate by the plan tool fully completes, there's no need for that to be in the activity bar anymore. It's complete and in the next turn, the agent may come up with another plan and the cycle restarts. So, this PR moves a fully complete plan to the thread view, so that it stays as part of a given turn:

image

The way this PR does this is by adding a new entry to `AgentThreadEntry` and snapshotting the completed plan so we can display it properly in the thread.

Release Notes:

- N/A


benbrandt (approved):
For our agent we could potentially do this based on the tool call output itself. but for other agents this is maybe more generic (it just won't persist on reload)

PR #52464, opened 2026-03-26
windows: Preserve launch environment for forwarded CLI opens

### LINKED ISSUE #47306: Regression: Zed CLI instance shares environment variables between windows again in 0.219.5
### Reproduction steps

1. Launch Zed from CLI/terminal with customized environment variables.
2. Launch a second instance of Zed (new project) from a different terminal, or from the Start menu/desktop shortcut with a different set of environment variables.
3. Check which environment variables each Zed process inherits (e.g., using an integrated terminal or an extension that prints the environment).

I noticed that the first launched Zed instance determines which environment variables get loaded and shares them with all subsequently opened Zed windows, even when launched from shells with different environments. According to the official documentation (https://zed.dev/docs/environment?highlight=environment#environment-variables), this was a bug that was fixed several versions ago. However, it seems to be back in 0.219.5.

**Current behavior:**
- When launching Zed from different shells with different environment variables, all windows share the environment of the first opened instance. 
- This contradicts the documentation, which says each window should use the environment context it was launched from, without sharing state between them.

**Expected behavior:**
- Each Zed window should respect the environment of the shell it was launched from, and not share shell-specific environment variables across instances—especially when launched from separate terminals or locations.

This is an issue because it leads to inconsistencies when working with projects that rely on shell/environment-specific variables.

Zed: 0.219.5 afc07d9fb20c3b7451071e2494d75f8837cb28db
OS: Windows 11

### Attach Zed log file

```log
2026-01-20T12:33:11-06:00 INFO  [zed] ========== starting zed version 0.219.5+stable. 122.afc07d9fb20c3b7451071e2494d75f8837cb28db, sha afc07d9 ==========
2026-01-20T12:33:11-06:00 INFO  [gpui::platform::windows::directx_devices] Using GPU: Intel(R) Graphics
2026-01-20T12:33:11-06:00 INFO  [gpui::platform::windows::directx_devices] Created device with Direct3D 11.1 feature level.
2026-01-20T12:33:11-06:00 INFO  [gpui::platform::windows::direct_write] Use Segoe UI as UI font. 
2026-01-20T12:33:11-06:00 INFO  [crashes] spawning crash handler process
2026-01-20T12:33:11-06:00 INFO  [crashes] connected to crash handler process after 100ms
2026-01-20T12:33:11-06:00 INFO  [zed::zed] Using GPU: 2026-01-20T12:33:11-06:00 INFO  [crashes] crash handler registered
GpuSpecs { is_software_emulated: false, device_name: "Intel(R) Graphics", driver_name: "Intel Corporation", driver_info: "32.0.101.6651" }
```

</details>

### Relevant Zed settings

```json
```
</details>

### Relevant Keymap

```json

```

</details>

_No response_

_No response_

loadingalias:
Closes #47306

## Summary
 • preserve the launching process env when Windows forwards a CLI open into an already running Zed instance
 • make the Windows CLI path serialize the launch environment explicitly, so both request producers stay consistent
 • keep the downstream environment plumbing unchanged, since `CliRequest::Open`, `open_listener`, and `ProjectEnvironment` already handle per-open CLI env correctly

### Why
The bug was in the Windows request producers, not in the receiver side. On Windows, both `crates/zed/src/zed/windows_only_instance.rs` and `crates/cli/src/main.rs` were constructing `CliRequest::Open` with `env: None`. That assumption only works when a fresh child process is doing the work. It breaks when a later launch forwards into an already running Zed process, which is the exact single instance path this issue describes.

The only way for the later window to see the later launch environment is to serialize it in the request.

## Verification
 • `cargo fmt --all --check`
 • `./script/check-keymaps`
 • `./script/clippy -p cli -p zed`
 • `cargo test -p cli -- --nocapture`
 • `cargo test -p cli -p zed -- --nocapture`

Release Notes:

 - Fixed Windows CLI-opened workspaces to preserve the launch envs when forwarding to an already running Zed instance.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -611,9 +611,12 @@ fn run() -> Result<()> {
 
         #[cfg(target_os = "windows")]
         {
-            // On Windows, by default, a child process inherits a copy of the environment block of the parent process.
-            // So we don't need to pass env vars explicitly.
-            None
+            use collections::HashMap;
+
+            // Child-process inheritance does not help when the request is forwarded into an
+            // already-running Zed instance. In that case we need to serialize the launching
+            // process environment explicitly.
+            Some(std::env::vars().collect::<HashMap<_, _>>())
         }
 
         #[cfg(not(any(target_os = "linux", target_os = "freebsd", target_os = "windows")))]
--- a/crates/zed/src/zed/windows_only_instance.rs
+++ b/crates/zed/src/zed/windows_only_instance.rs
@@ -159,7 +159,7 @@ fn send_args_to_instance(args: &Args) -> anyhow::Result<()> {
             wait: false,
             wsl: args.wsl.clone(),
             open_behavior: Default::default(),
-            env: None,
+            env: Some(std::env::vars().collect()),
             user_data_dir: args.user_data_dir.clone(),
             dev_container: args.dev_container,
             cwd: std::env::current_dir().ok(),


loadingalias:
This was a code first trace/fix from my local macOS box, not a local Windows repro.

I validated the full local suite I could here; a quick Windows repro/confirm before merge would still be useful, IMO.

loadingalias:
Rebased onto `main`.

The conflict was in the Windows single-instance request path. I kept the current `open_behavior` request shape from `main` and reapplied the actual fix: 

Windows forwarded opens now send the launching process environment instead of `env: None`.

The diff is still just the two request producers:
• `crates/cli/src/main.rs`
• `crates/zed/src/zed/windows_only_instance.rs`

I still don't have a local Windows repro machine, so we still need a check before merge.

ChristopherBiscardi:
Thanks for the PR!

I don't think its wise to write platform-specific fixes without testing on the platform that has the issue. I'm not sure how you would've verified that this actually fixed it which makes me extra wary of how this fix was created, especially since the PR doesn't add any tests that would run in CI.

loadingalias:
With all the respect in the world - this is annoying to see closed.

I clearly agree that the fix should be validated on Windows before merge - I've said that explicitly in the PR and again after rebasing.

However, I completely disagree w/ treating that as a reason to close the PR, or as some kind of evidence that the PR was created carelessly... or it's like a shit PR because I don't have a Windows box?

The linked issue already has a Windows repro. The code path is also VERY narrow: both Windows request producers were sending `env: None`, while the downstream `CliRequest::Open` path already supports carrying a per-open env.

The patch changes those producers to serialize the launching process env instead of relying on child process inheritance, which doesn't apply once the request is forwarded into an already running Zed instance.

If the bar is “add a CI test for this request” - that is a clear review request and I’m happy to do it. If the bar is “the author/contributor must personally own the affected platform,” then external contributors can't practically help with platform-specific bugs unless they have the exact hardware/OS locally, even when the bug is clearly source traceable and already reproduced by other users.... presumably on the target.

I’m fine w/ you reviewing it and telling me “needs Windows validation before merge.” I’m not fine with closing a minimal, pretty well scoped fix on the basis that I was transparent about needing that validation.

Why through away the fix? It's not my first PR. It's not a junk PR. I don't know. Ultimately, this is your house and I'm visiting. If this is what you want... keep it closed. If you decide that you'd like to actually look at the fix and have someone on the team run the manual checks on a Windows box + merge to actually close the issue - let me know. If there is a flaw or weakness in the impl - let me know.

ChristopherBiscardi:
I can empathize with being annoyed. Having a PR closed can feel very intense but it is not a reflection of you personally or the effort you put in.

To expand on why this was closed: We can not entertain PRs that are not validated by their authors. Even though the linked issue has *other* people who have reproduced the issue, it is clear that you did not at any point run that reproduction. Since you didn't run the reproduction, you also can't have validated whether this actually fixes the issue. Fundamentally this means *you do not know if this code actually fixes the issue* and are relying on reviewers to figure that out and deal with the ramifications. This disproportionately places the burden on reviewers and is the reason for the close. We appreciate your effort, but reviewers are not your personal CI system for platforms you don't have access to.

> external contributors can't practically help with platform-specific bugs unless they have the exact hardware/OS locally,

Yes. We do not want PRs fixing platform issues if the fix can not be validated by the author who wrote it. It is expected that you have run the code on the platform the fix is intended for and that you have validated that it actually fixes the issue.

There are plenty of other issues to work on that *can* be validated by an author without a specific platform.

The clear request you are asking for is this: Please work on fixes you can validate for platforms you can execute code for. Please do not submit code you haven't executed or verified.

PR #52475, opened 2026-03-26
Fix slow Tree-sitter parsing on broken code

### LINKED ISSUE #52390: Slow tree-sitter parsing on broken code
### Reproduction steps

1. Generate a file with broken sql 
```sh
$ (echo ""; for i in $(seq 1 9999); do echo "('$i'),"; done; echo "('10000')"; echo ') as x(ix);') > test_broken.sql
```
2. `zed test_broken.sql`

Try scrolling down or editing the `test_broken.sql`, it's really slow because of the main thread being stalled.

I've fixed this same issue in gpui-component, the problem there was that tree-sitter parses the broken file into
```
[program][1, 0] - [10003, 0]
  [ERROR] [1, 0] - [10001, 11]
    [ERROR] [1, 1] - [1, 6]
    [ERROR] [2, 1] - [10001, 6]
      [keyword_as] [10001, 2] - [10001, 4]
```

So tree-sitter will try to iterate over all the nodes for highlighting on every render, even though they are outside of the visible range. I haven't looked at how Zed sets things up, but I solved it with trying to skip subtrees that are outside of the visible range.
I'm no tree-sitter expert though, perhaps there's a better way to detect this and bail early?

The broken sql thing is quite common for me, e.g. copying a big insert statement, and then editing the insert part into like `SELECT * FROM (VALUES`. Zed probably runs into the same issue with other languages when a large part of code becomes "broken", like a large array of any language. This is just a way to easily reproduce the issue

Zed: v0.228.0+stable.203.8421009ef8a022df1196d54bb42fd94366ec0988 (Zed) 
Memory: 60.7 GiB
Architecture: x86_64
GPU: AMD Radeon RX 5700 XT (RADV NAVI10) || radv || Mesa 25.2.8-0ubuntu0.25.10.1

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

OmChillure:
#### Context

When tree-sitter parses a file with broken syntax (e.g. a large partial SQL `VALUES` clause, or any language where a large chunk becomes invalid), it can produce a single `ERROR` node spanning thousands of lines. On every render frame, Zed queries this tree for syntax highlights via `SyntaxMapCaptures`. Previously, only `set_byte_range` was applied to the query cursor - this limits which captures are *returned*, but tree-sitter still had to *traverse* the entire ERROR subtree to find them, causing O(file size) work per frame and making scrolling/editing visibly laggy.

The fix applies `set_containing_byte_range` to the highlight query cursor, mirroring what `SyntaxMapMatches` already does for indentation and bracket queries. This tells tree-sitter to skip subtrees that extend far beyond the visible window, reducing traversal to the visible range only.

**Note:** This fix eliminates the main freeze/stall caused by full-tree traversal.  A small amount of lag may still occur on very large broken files, as tree-sitter still needs to parse the error-recovery structure. Further improvements would require deeper changes to tree-sitter's query execution or incremental parsing.

#### How to Review

Small change — focus on [syntax_map.rs:1119-1123](crates/language/src/syntax_map.rs#L1119) (the fix) and the `containing_byte_range_for_captures` helper below it. Compare with the existing `SyntaxMapMatches::new` path (line ~1255) which uses the same pattern.

#### Self-Review Checklist

#### Video 
[Screencast from 2026-03-26 14-19-19.webm](https://github.com/user-attachments/assets/6628492a-f013-438a-836a-2740f6e2f266)

Release Notes:

- Fixed laggy scrolling and editing in files with large broken syntax regions (e.g. incomplete SQL `VALUES` clauses or large invalid blocks in any language)

### DIFF
--- a/crates/language/src/syntax_map.rs
+++ b/crates/language/src/syntax_map.rs
@@ -25,6 +25,7 @@ use tree_sitter::{
 };
 
 pub const MAX_BYTES_TO_QUERY: usize = 16 * 1024;
+const CAPTURE_CONTAINING_CONTEXT_BYTES: usize = 3 * 1024;
 
 pub struct SyntaxMap {
     snapshot: SyntaxSnapshot,
@@ -1117,6 +1118,7 @@ impl<'a> SyntaxMapCaptures<'a> {
             };
 
             cursor.set_byte_range(range.clone());
+            cursor.set_containing_byte_range(containing_byte_range_for_captures(&range));
             let captures = cursor.captures(query, layer.node(), TextProvider(text));
             let grammar_index = result
                 .grammars
@@ -1206,6 +1208,19 @@ impl<'a> SyntaxMapCaptures<'a> {
             .position(|layer| layer.next_capture.is_none())
             .unwrap_or(self.layers.len());
     }
+}
+
+fn containing_byte_range_for_captures(query_range: &Range<usize>) -> Range<usize> {
+    let desired_window = query_range
+        .len()
+        .saturating_add(CAPTURE_CONTAINING_CONTEXT_BYTES.saturating_mul(2));
+    let containing_window_len = desired_window.min(MAX_BYTES_TO_QUERY);
+    let midpoint = query_range
+        .start
+        .saturating_add(query_range.len().saturating_div(2));
+    let containing_start = midpoint.saturating_sub(containing_window_len / 2);
+    let containing_end = containing_start.saturating_add(containing_window_len);
+    containing_start..containing_end
 }
 
 #[derive(Default)]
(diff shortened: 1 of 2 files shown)


OmChillure:
> `set_containing_byte_range` doesn't quite fix it in my experience, I think you can see that when you scroll further down in the file. I had to collect nodes from `layer.node()` to be able to skip problematic nodes, and then run `.matches` on the nodes collected.

Thanks for the suggestion! I went with `set_containing_byte_range` for now since it follows the same pattern already used in SyntaxMapMatches for indentation/bracket queries and fixes the main freeze/larger lags. It still has a tiny bit of lag on really large broken files when scrolling far down, but no more stalling.

I considered the node-collection approach you described - walking children with goto_first_child_for_byte and running captures per-node but it touches the captures iterator architecture quite a bit. Felt like that'd be better as a follow-up if this doesn't hold up well enough in practice. Lets wait for reviwers to chime in and suggest some stuff !

PR #52486, opened 2026-03-26
collab_ui: Fix "lost session" visual bug in Collab Panel

ruxwez:
## Context

This PR fixes a UX issue ("visual bug") in the collaboration panel documented in issue [#51800](https://github.com/zed-industries/zed/issues/51800), where users who had already signed in were still seeing the "Sign In" screen after restarting the editor. As I mentioned in my response there ([Link to comment #4132366441](https://github.com/zed-industries/zed/issues/51800#issuecomment-4132366441)), I have investigated the problem thoroughly and found that the session is not actually lost.

What I discovered is that in Zed, only "staff" users automatically connect to the collaboration servers when opening the editor (by design, this logic is in `crates/client/src/client.rs` starting at line `962`). Therefore, regular users keep their saved session and `Authenticated` status, but since they don't automatically connect upon startup, the UI didn't detect this correctly. It erroneously showed the GitHub account request and the "Sign in to enable collaboration" text, giving the false impression that the user had been logged out.

### Screenshots

Before (Bug)
image

After (Fix)
image

**Note:** This PR specifically addresses the visual issue in the **Collab Panel**. Similar behaviors might exist in other parts of the editor, but this change focuses on correcting the collaboration interface.

This current PR:
1. Improves the `render_signed_out` function in `crates/collab_ui/src/collab_panel.rs`.
2. Simplifies the connection check using `self.client.user_id().is_some()`, which is more robust against volatile network states and perfectly covers connection transitions.
3. During rendering, it detects existing credentials and shows the correct message "Connect" / "Connecting...", replacing the GitHub icon with the appropriate network icon (`SignalHigh`).

## How to Review

- Review the cleaner and simplified code in `crates/collab_ui/src/collab_panel.rs:render_signed_out`.
- Verify that instead of verbose validations on the `Status` enum, simply checking the user ID correctly captures any subsequent subtype, properly differentiating between account authorization and a simple network reconnection.

## Self-Review Checklist

Release Notes:

- Fixed an issue (#51800) in the Collab Panel where the UI appeared to log users out. Implemented improvements to properly differentiate between "Sign In" and "Connect," avoiding false authentication prompts when users are already logged in but not automatically connected to the servers.

### DIFF
--- a/crates/collab_ui/src/collab_panel.rs
+++ b/crates/collab_ui/src/collab_panel.rs
@@ -2351,7 +2351,7 @@ impl CollabPanel {
         let (button_id, button_label, button_icon, helper_text) = if is_authenticated {
             (
                 "connect",
-                if is_busy { "Connecting…" } else { "Connect to Collaboration" },
+                if is_busy { "Connecting…" } else { "Connect" },
                 Icon::new(IconName::SignalHigh).color(Color::Muted),
                 "Connect to collaboration servers.",
             )
--- a/crates/collab_ui/src/collab_panel.rs
+++ b/crates/collab_ui/src/collab_panel.rs
@@ -2340,11 +2340,28 @@ impl CollabPanel {
 
     fn render_signed_out(&mut self, cx: &mut Context<Self>) -> Div {
         let collab_blurb = "Work with your team in realtime with collaborative editing, voice, shared notes and more.";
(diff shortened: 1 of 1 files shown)


danilo-leal (approved):
Thank you, good catch!

PR #52490, opened 2026-03-26
gpui: Fix scrollbar drag position inversion in bottom-aligned lists

KlausUllrich:
## Context

The scrollbar drag handler in `list.rs` uses `.abs()` to normalize the scroll
position, which causes scroll position inversion when content is actively growing
(e.g., in a streaming chat view or terminal output).

The current code at line 952:

    let new_scroll_top = (point.y - drag_offset).abs().max(px(0.)).min(scroll_max);

The scrollbar always passes negative `point.y` values (representing downward offset
from the top of the content area). The `drag_offset` compensates for content height
changes during drag by subtracting `(current_height - frozen_height)`.

The problem: when content grows during a drag, `drag_offset` becomes positive and
can exceed `-point.y`. This makes `(point.y - drag_offset)` a large negative number,
and `.abs()` flips it to a large positive number -- mapping to a scroll position near
the *top* of the list instead of the intended position. The user drags the scrollbar
thumb down, but the viewport jumps upward.

The fix replaces the `.abs()` approach with direct negation:

    let new_scroll_top = (-point.y).max(px(0.)).min(scroll_max);

Since `point.y` is always negative (downward offset), `-point.y` gives the correct
positive scroll-top value. The `.max(px(0.)).min(scroll_max)` clamp handles edge
cases.

The `drag_offset` compensation is also removed because during a drag operation, both
the scrollbar position mapping and `scroll_max` should use the same frozen content
height (`scrollbar_drag_start_height`). The `drag_offset` was double-compensating:
the scrollbar already maps the thumb position relative to the content height at drag
start, so adjusting the result for content growth introduces drift.

I discovered this while building a bottom-aligned list view with streaming content
(new items appending while the user interacts with the scrollbar). The inversion is
reproducible in any `ListAlignment::Bottom` list where content height increases during
a scrollbar drag.

## How to Review

Focus on `set_offset_from_scrollbar` in `crates/gpui/src/elements/list.rs` (line ~940).
The change is small:

1. `content_height` now uses `scrollbar_drag_start_height` (frozen at drag start)
   instead of live `items.summary().height`
2. `drag_offset` calculation removed entirely
3. `.abs()` replaced with `-point.y`

The corresponding read path `scroll_px_offset_for_scrollbar` also removes the
`drag_offset` for consistency -- both sides of the scrollbar mapping now use the
same offset basis.

## Self-Review Checklist

Release Notes:

- gpui: Fixed scrollbar drag position inversion in bottom-aligned lists when content height changes during drag

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2429,6 +2429,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bzip2"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3a53fac24f34a81bc9954b5d6cfce0c21e18ec6959f44f56e8e90e4bb7c346c"
+dependencies = [
+ "libbz2-rs-sys",
+]
+
+[[package]]
 name = "bzip2-sys"
 version = "0.1.13+1.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3467,6 +3476,7 @@ version = "0.4.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef8a506ec4b81c460798f572caead636d57d3d7e940f998160f52bd254bf2d23"
 dependencies = [
+ "bzip2 0.6.1",
  "compression-core",
  "deflate64",
  "flate2",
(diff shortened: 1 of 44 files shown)


dinocosta (commented):
Hey @KlausUllrich  👋

Thank you for wanting to contribute to Zed. I took a quick look at the changes in this pull request, and they seem to touch more than just the issue you’re reporting. It seems they also include changes to code related to edit prediction, the default keymap, and Elixir’s default language servers.

As such, I’m going to go ahead and close this pull request. If this is an issue you’re actually facing, I’d suggest opening an issue first at ￼https://github.com/zed-industries/zed/issues , so that someone else can try to reproduce it and we can then compare any proposed changes directly against the reported problem. 

You might also find our [contributing guide](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md) helpful for next steps and for structuring future Pull Requests. Thanks! 🙂

Issue #29048, opened 2025-04-18
Git: zed fails to ssh-askpass on commit


dnbln:
### Summary

Zed fails to ssh-askpass on commit

### Description

Steps to trigger the problem:
1. Open zed in a project set up with git
2. Make some changes.
3. Follow the steps in https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key
4.
```bash
git config --global commit.gpgsign true
git config --global tag.gpgSign true
```
5. Attempt to create a commit in zed.

Actual Behavior:

```
Failed to commit:
error: ssh_askpass: exec(/usr/X11R6/bin/ssh-askpass): No such file or directory?
Load key "/Users/dinu/.ssh/id_ed25519": incorrect passphrase supplied to decrypt private key?

fatal: failed to write commit object
```

Expected Behavior:
It working correctly

### Zed Version and System Specs

Zed: v0.182.11 (Zed) 
OS: macOS 13.7.4
Memory: 16 GiB
Architecture: x86_64

markusstephanides:
Still getting this issue on git push:

```
ssh_askpass: exec(powershell.exe -ExecutionPolicy Bypass -File "C:\\Users\\<USER>\\AppData\\Local\\Temp\\zed-askpasslbB2Hn\\askpass.ps1"): No such file or directory
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

No prompt for the password, just this error directly.

Zaenalos:
Any updates?

PR #52491, opened 2026-03-26
Fix SSH askpass on Windows by invoking `cli.exe` directly

Zaenalos:
On Windows, Zed generated a .ps1 askpass script and set **SSH_ASKPASS** to 'powershell.exe -ExecutionPolicy Bypass -File ...'. SSH calls exec() on SSH_ASKPASS which cannot exec a command string, causing:

  error: ssh_askpass: exec(powershell.exe ...): No such file or directory

Fix by pointing **SSH_ASKPASS** directly to cli.exe and passing the socket path via **ZED_ASKPASS_SOCKET** env var, which cli.exe reads before clap parses arguments.

---
> Compiling Zed almost blew up my computer.
---

Related #29048 

Release Notes:

- Fixed some ssh issues on windows that prevented from connecting to a remote

### DIFF
--- a/crates/askpass/src/askpass.rs
+++ b/crates/askpass/src/askpass.rs
@@ -4,6 +4,7 @@ pub use encrypted_password::{EncryptedPassword, IKnowWhatIAmDoingAndIHaveReadThe
 
 use net::async_net::UnixListener;
 use smol::lock::Mutex;
+#[cfg(not(target_os = "windows"))]
 use util::fs::make_file_executable;
 
 use std::ffi::OsStr;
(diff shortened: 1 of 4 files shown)


Zaenalos:
Greetings, can someone review my PR? Thank you.

Veykril (commented):
Generally I am happy with this, but I think this might cause the askpass helper to be invoked with the prompt passed as an argument to the helper instead of being passed via stdin which may cause issues

Veykril (approved):

Zaenalos:
Thank you, sir🫡

PR #52494, opened 2026-03-26
sidebar: Ensure the projects menu is dismissed

danilo-leal:
Specifically the "Add Local Project" button was missing emitting a dismiss event. Regular list items within the picker match entries are covered because the picker already handles that, but using that standalone button would sometimes not dismiss the menu properly.

Release Notes:

- N/A


cameron1024:
@zed-industries/approved

PR #52499, opened 2026-03-26
Always pass `--no-optional-lock` to git

Veykril:
Release Notes:

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -1415,10 +1415,7 @@ impl GitRepository for RealGitRepository {
             .spawn(async move {
                 let git = git_binary?;
                 let mut process = git
-                    .build_command(&[
-                        "cat-file",
-                        "--batch-check=%(objectname)",
-                    ])
+                    .build_command(&["cat-file", "--batch-check=%(objectname)"])
                     .stdin(Stdio::piped())
                     .stdout(Stdio::piped())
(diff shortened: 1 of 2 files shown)


crates/worktree/src/worktree.rs
```diff
@@ -4059,6 +4061,7 @@ impl BackgroundScanner {
     }
 
     async fn process_events(&self, mut events: Vec<PathEvent>) {
+        dbg!(&events);
```
mikayla-maki: dbg!

mikayla-maki (dismissed):

crates/git/src/repository.rs
```diff
@@ -3039,6 +3025,7 @@ impl GitBinary {
         let mut command = new_command(&self.git_binary_path);
         command.current_dir(&self.working_directory);
         command.args(["-c", "core.fsmonitor=false"]);
+        command.arg("--no-optional-locks");
```
mikayla-maki, line 3028: https://git-scm.com/docs/git-status#_background_refresh

Do we need to run some git status calls _with_ optional locks to make sure we have up to date info?

Veykril: I don't believe so. We basically do not want to refresh the index ourselves when querying for information from what i can gather, so we want to put this on pretty much everything.

cole-miller (approved):

PR #52505, opened 2026-03-26
client: Persist last used organization

NeelChotai:
## Context

Persists last used organization through restart. Opted to do this via `kvp` instead of `settings.json` since the value could change often, and we would have to persist an ID rather than a friendly name.

Closes CLO-568.

## Self-Review Checklist

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2972,6 +2972,7 @@ dependencies = [
  "cloud_llm_client",
  "collections",
  "credentials_provider",
+ "db",
  "derive_more",
  "feature_flags",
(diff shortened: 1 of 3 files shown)


tomhoule (approved):
Nicely self contained. Looks good to me 👍

maxdeviant (approved):
I have some minor concerns about this given that the key-value store isn't easily inspectable by users for debugging, but we can always revisit if it becomes problematic.

I'm on board with using the KV store as the MVP for this 👍🏻

PR #52506, opened 2026-03-26
Prevent dev container modal dismissal during creation

antont:
## Context

When the dev container creation modal is showing "Creating Dev Container", clicking anywhere on the workspace backdrop dismisses the dialog. The container creation continues in the background, but the user loses visual feedback and the subsequent `open_remote_project` call may fail because the modal entity is gone.

This adds an `allow_dismissal` flag to `RemoteServerProjects` that blocks accidental dismissal (backdrop clicks, focus loss) while a dev container is being created, but allows explicit dismissal on success or error.

## How to Review

Small PR — two files changed:

1. **`remote_servers.rs`** (the fix): `allow_dismissal` bool field added, set to `false` when entering Creating state, set to `true` before emitting `DismissEvent` on success/error. `on_before_dismiss` override checks the flag.
2. **`recent_projects.rs`** (the test): Regression test that opens a dev container modal, simulates a backdrop click, and asserts the modal stays open.

## Self-Review Checklist

Release Notes:

- Fixed dev container creation modal being dismissed when clicking outside it

### DIFF
--- a/crates/recent_projects/src/recent_projects.rs
+++ b/crates/recent_projects/src/recent_projects.rs
@@ -2003,7 +2003,7 @@ mod tests {
     use std::path::PathBuf;
 
     use editor::Editor;
-    use gpui::{TestAppContext, UpdateGlobal, WindowHandle};
+    use gpui::{TestAppContext, UpdateGlobal, VisualTestContext, WindowHandle};
 
     use serde_json::json;
     use settings::SettingsStore;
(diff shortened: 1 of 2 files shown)


KyleBarton:
Hi @antont thanks for this change - mind merging main and addressing the failing test here? Otherwise, this approach looks sound and works as expected when I test. We should be good to merge once CI is green. Thanks!

KyleBarton (approved):

PR #52507, opened 2026-03-26
Fix submodules being incorrectly classified as linked worktrees

rtfeldman:
For submodules, `common_dir_abs_path` equals `repository_dir_abs_path` (since submodules don't have a `commondir` file). The previous code passed this path to `original_repo_path_from_common_dir` unconditionally, which returned the `.git/modules/<name>` path as the `original_repo_abs_path` — causing `linked_worktree_path()` to return a false positive for submodules.

Now we detect linked worktrees by checking whether `common_dir` differs from `repository_dir` (only true for actual linked worktrees that have a `commondir` file). For normal repos and submodules, `original_repo_abs_path` is simply `work_directory_abs_path`.

Also fixes the misleading doc comment on `common_dir_abs_path` in `LocalRepositoryEntry` and adds test assertions for `original_repo_abs_path` and `linked_worktree_path()` on both worktrees and submodules.

Closes AI-102

Release Notes:

- Fixed git submodules being incorrectly classified as linked worktrees, which could cause issues with worktree-related operations.

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -1532,13 +1532,24 @@ impl GitStore {
             } else if let UpdatedGitRepository {
                 new_work_directory_abs_path: Some(work_directory_abs_path),
                 dot_git_abs_path: Some(dot_git_abs_path),
-                repository_dir_abs_path: Some(_repository_dir_abs_path),
+                repository_dir_abs_path: Some(repository_dir_abs_path),
                 common_dir_abs_path: Some(common_dir_abs_path),
                 ..
             } = update
             {
-                let original_repo_abs_path: Arc<Path> =
-                    git::repository::original_repo_path_from_common_dir(common_dir_abs_path).into();
+                // For linked worktrees, common_dir points to the main repo's
+                // .git directory (different from the worktree-specific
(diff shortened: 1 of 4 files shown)


crates/project/src/git_store.rs
```diff
@@ -1532,13 +1532,24 @@ impl GitStore {
...
             {
-                let original_repo_abs_path: Arc<Path> =
-                    git::repository::original_repo_path_from_common_dir(common_dir_abs_path).into();
+                // For linked worktrees, common_dir points to the main repo's
+                // .git directory (different from the worktree-specific
+                // repository_dir), so we can derive the original repo's working
+                // directory from it. For normal repos and submodules,
+                // common_dir equals repository_dir, and the original repo is
+                // simply the work directory itself.
+                let is_linked_worktree =
```
rtfeldman: Note to self: we now have a helper function called `is_linked_worktree` which we can use for this instead.

Anthony-Eid: https://github.com/zed-industries/zed/blob/a16b74024469059823a2ad5f33a01a853297f53b/crates/project/src/git_store.rs#L3709-L3725

Anthony-Eid (approved):

PR #52524, opened 2026-03-26
Group threads by canonical path lists

eholk:
## Context

With the new sidebar, we are having some bugs around multi-root projects combined with git work trees that can cause threads to be visible in the agent panel but not have an entry in the sidebar.

## How to Review

This PR takes a step towards resolving these issue by adding a `ProjectGroupBuilder` which is responsible for gathering the set of projects groups from the open workspaces and then helping to discover threads and map them into this set.

## Self-Review Checklist

Release Notes:

- N/A


crates/sidebar/src/sidebar.rs
```diff
@@ -607,257 +640,191 @@ impl Sidebar {
...
+                            icon,
+                            icon_from_external_svg,
+                            status: AgentThreadStatus::default(),
+                            workspace: ThreadEntryWorkspace::Open(workspace.clone()),
+                            is_live: false,
+                            is_background: false,
+                            is_title_generating: false,
+                            highlight_positions: Vec::new(),
+                            worktree_name: worktree_info.as_ref().map(|(name, _)| name.clone()),
+                            worktree_full_path: worktree_info
```
eholk, line 740: Since a thread can have multiple paths, the name and full_path fields need to be a vector so we can reflect this in the UI. Right now we just arbitrarily pick one to show. I'm fixing this in a follow-up PR.

eholk (commented):
I'm working on a follow-up PR that better handles multi-root workspaces and different corner cases with work trees there. It was getting quite a bit bigger, so I wanted to get this PR ready to go first to commit something incremental. This should be strictly better than what we had before.

crates/sidebar/src/project_group_builder.rs
```diff
@@ -0,0 +1,325 @@
...
+            .first()
+            .expect("groups always have at least one workspace")
+    }
+}
+
+pub struct ProjectGroupBuilder {
+    /// Maps git repositories' work_directory_abs_path to their original_repo_abs_path
+    directory_mappings: HashMap<PathBuf, PathBuf>,
+    project_group_names: Vec<ProjectGroupName>,
+    project_groups: Vec<ProjectGroup>,
```
eholk, line 82: In a followup PR we should replace this with an IndexMap or something similar. Maybe add VecMap to our collections crate.

mikayla-maki (approved):

PR #52526, opened 2026-03-26
open_router: Fix incorrect rate limit enforcement after reset had already expired

tidely:
## Context

The `extract_retry_after` function in the `open_router` crate falsely returned `None` for its timeout when the reset had already passed before the function was called. This would make the caller assume the header was missing, and issue a default 60 second timeout.

The new behavior returns `Duration::ZERO` when the timeout has already passed.

## How to Review

The only logic change happens in the `extract_retry_after` function. I have included other nits such as updating documentation links, fixing typos, removing dead code and combining fragmented builder method calls. Going through the commits one by one keeps changes in logical groups.

## Self-Review Checklist

Release Notes:

- Fix rare incorrect rate limit handling when using OpenRouter

### DIFF
--- a/crates/deepseek/src/deepseek.rs
+++ b/crates/deepseek/src/deepseek.rs
@@ -7,7 +7,6 @@ use futures::{
 use http_client::{AsyncBody, HttpClient, Method, Request as HttpRequest};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use std::convert::TryFrom;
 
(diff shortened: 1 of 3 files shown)


benbrandt:
Thanks but this seems to have a lot of unrelated changes that are making this harder to review regarding the bug in question. Given there are also conflicts now, I am going to close, but feel free to re-open a more scoped one if you like

PR #52527, opened 2026-03-26
agent_ui: Rework ACP mode selector

tartarughina:
## Context

The modes/agent selector in ACP mode is not consistent with the model selector.
As soon as I started creating more than 10 modes/agents I really missed the option to set favorites, quickly cycle through them, as well as fuzzy finding for those instances where the mode/agent is not among the fav ones.

No issue exists for this.

Add favorite modes support to the mode selector, allowing users to star modes and cycle through them with a keyboard shortcut, mirroring the existing favorite models feature.

### Changes

**Favorite modes infrastructure:**
- Add `favorite_mode_ids()` and `toggle_favorite_mode()` to the `AgentServer` trait, with implementation in `CustomAgentServer` that reads/writes favorite mode IDs from settings.
- Add `favorite_modes: Vec<String>` field to all three `CustomAgentServerSettings` variants (`Custom`, `Extension`, `Registry`) in both `settings_content` and `project` crates, persisted via `serde` with `skip_serializing_if = "Vec::is_empty"`.

**Mode selector rewrite:**
- Replace the old `ContextMenu`-based `ModeSelector` with a `Picker<ModePickerDelegate>` that supports fuzzy search, grouped favorites/all sections, and inline favorite toggling.
- Extract the popover wrapper into a new `ModeSelectorPopover` component (similar to the existing `ModelSelectorPopover` pattern), keeping the picker and popover concerns separate.

**New actions and keybindings:**
- Bind `shift-tab` → `CycleFavoriteModes` replacing `CycleModeSelector`
- The mode selector tooltip conditionally hides the "Cycle Favorited Modes" row when fewer than 2 modes are favorited.

## How to Review

1. Start with the `AgentServer` trait additions in `agent_servers.rs` and the settings schema in `settings_content/src/agent.rs` — these define the data model.
2. Read the new `mode_selector.rs` picker implementation (`ModePickerDelegate`) and `mode_selector_popover.rs` wrapper.
3. Check `thread_view.rs` for the `CycleFavoriteModes` action handler wiring.
4. Skim the keymap changes in `default-*.json` for the new bindings.

## Self-Review Checklist

Release Notes:

- Added support for favoriting agent modes in the mode selector, with `shift-tab` to cycle through favorited modes.

BEFORE
image

AFTER
image


benbrandt (changes requested):
Hi I think there are some bugs here that were introduced by registering double handlers and I'm nto sure we added favorite modes everywhre, but also I'd like to find a route that helps us share more UI code here if possible.

PR #52537, opened 2026-03-26
terminal_view: Resolve remote terminal paths outside worktrees

loadingalias:
Closes #39159

## Summary
 • resolve remote terminal path candidates outside known worktrees through `Project::resolve_abs_path`
 • keep the local filesystem backed resolution path unchanged
 • preserve the existing preference for direct in-worktree matches when the terminal cwd is already inside a worktree
 • add regression coverage for both absolute and cwd-relative remote external paths

### Why
The hyperlink parser wasn't the bug. `terminal_path_like_target` already knew how to parse these paths. The actual problem was that remote projects had no equivalent of the local background existence check. After worktree lookup failed, remote projects only fell back to worktree traversal, so cmd-click on `/tmp/a.txt` in an SSH remote terminal no-op'd.

The project layer already has remote absolute-path resolution via `Project::resolve_abs_path`, so the right fix is to use that existing path for remote candidates instead of pretending remote projects cannot resolve paths outside their worktrees.

## Verification
 • `cargo test -p terminal_view terminal_path_like_target --quiet`
 • `cargo test -p terminal_view --quiet`

## Manual
 • Verified locally on my MBP w/ a stateless Zed build using SSH to self (`localhost`)
 • Opened `/tmp/zed-39159-project` as a remote project
 • Confirmed cmd-click opens `/tmp/zed-39159-external/a.txt`
 • Confirmed cmd-click opens `zed-39159-external/b.txt` from `cwd=/tmp`
 • Confirmed in-worktree paths still open correctly
 • Confirmed nonexistent external paths do not open bogus targets

Release Notes:

  - Fixed remote terminal path opening so cmd-click can open files outside the project worktree.

### DIFF
--- a/crates/terminal_view/src/terminal_path_like_target.rs
+++ b/crates/terminal_view/src/terminal_path_like_target.rs
@@ -9,7 +9,7 @@ use util::{ResultExt, debug_panic};
 use workspace::path_link::possible_open_target;
 #[cfg(test)]
 use workspace::path_link::{
-    BackgroundFsChecks, OpenTargetFoundBy, possible_open_target_with_fs_checks,
+    BackgroundPathChecks, OpenTargetFoundBy, possible_open_target_with_fs_checks,
 };
 use workspace::{OpenOptions, OpenVisible, Workspace, path_link::OpenTarget};
 
@@ -30,7 +30,7 @@ pub(super) fn hover_path_like_target(
             hovered_word,
             path_like_target,
             cx,
-            BackgroundFsChecks::Enabled,
+            BackgroundPathChecks::LocalFileSystem,
         )
     }
 }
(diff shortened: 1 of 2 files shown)


shahamran:
Hi, any chance to get this reviewed? It seems to solve my only blocker for using Zed.
Thanks!

loadingalias:
If "Release Notes" are actually blocking this - someone tell me and I'll fix it.

I don't know what the hold up it @shahamran.

Veykril (commented):
The reason for us not reviewing this timely is just a lack of capacity for reviewing, sorry! It seems #57303 moved most of this around so this will need a rebase there. Once that's done I'll try to review this ASAP

loadingalias:
Rebased onto current `main`.

#57303 moved the path-linking code into `workspace::path_link`, so I moved this fix there too instead of trying to keep the old terminal-local impl alive.

Local projects still use the normal filesystem metadata path; remote projects now ask the project layer to resolve absolute / cwd-based terminal paths w/ `Project::resolve_abs_path`.

Veykril (approved):
Thanks!

loadingalias:
> Thanks!

You're welcome!

PR #52540, opened 2026-03-26
sidebar: Add property tests

Anthony-Eid:
## Self-Review Checklist

Release Notes:

- N/A


crates/sidebar/src/sidebar.rs
```diff
@@ -3290,9 +3290,297 @@ mod tests {
...
+        }
+    }
+    /// Verifies the sidebar's materialized contents against the invariants that
+    /// `rebuild_contents` is expected to uphold.
+    ///
+    /// Invariants:
+    /// - Visible thread rows must exactly match the rows that should currently be
+    ///   materialized for each workspace given the current collapse state,
+    ///   expansion state, query text, saved thread metadata, and live agent panel
+    ///   overlay semantics.
```
mikayla-maki: I think an easier property to mention here is:
(Threads for collapsed groups ∪ Threads in the sidebar) = (Active threads).

We should ignore search.

Anthony-Eid: One problem I ran into is that the sidebar won't always show all threads per project header, because the default visible thread limit is 5 until we show a "view more" entry

mikayla-maki (approved):

eholk (approved):
Excited to have property tests now!

PR #52543, opened 2026-03-26
Restore language query watcher in dev builds

MrSubidubi:
The watcher had been broken for some time, but became even more broken after the recent move of the queries. 

This PR restores the reloading behavior for debug builds so that languages are reloaded once a scheme file is changed. 

Release Notes:

- N/A

### DIFF
--- a/crates/edit_prediction_cli/Cargo.toml
+++ b/crates/edit_prediction_cli/Cargo.toml
@@ -65,7 +65,7 @@ rand.workspace = true
 similar = "2.7.0"
 flate2 = "1.1.8"
 toml.workspace = true
-rust-embed = { workspace = true, features = ["debug-embed"] }
+rust-embed.workspace = true
 gaoya = "0.2.0"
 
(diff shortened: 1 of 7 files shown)


crates/edit_prediction_cli/Cargo.toml
```diff
@@ -65,7 +65,7 @@ rand.workspace = true
 similar = "2.7.0"
 flate2 = "1.1.8"
 toml.workspace = true
-rust-embed = { workspace = true, features = ["debug-embed"] }
+rust-embed.workspace = true
```
osiewicz, line 68: This could've been a separate PR, Finn. But aight

osiewicz (approved):

crates/edit_prediction_cli/Cargo.toml
MrSubidubi: This is actually needed here as otherwise the queries will not actually refresh.

PR #52565, opened 2026-03-27
gpui: Fix `scroll_to_reveal_item` scrolling for visible items

Skxxtz:
### Description

This PR refactors `scroll_to_reveal_item` in `gpui` to prevent unnecessary "jumping" when navigating through a list. 

Previously, moving "Up" would aggressively snap the selected item to the top of the viewport even if it was already fully visible.

 #### Previous behavior:

https://github.com/user-attachments/assets/bb3e646c-f2b4-4c4a-b384-77d5b9a3ac58

#### New behavior:

https://github.com/user-attachments/assets/4b144e43-ad3c-4982-8964-de2a09c7ab90

#### How:

The new implementation uses pixel-coordinate math (via `SumTree` cursors) to implement a **Nearest** scroll strategy:
- **If the item is already fully visible:** The scroll position remains unchanged.
- **If the item is above the viewport:** The list scrolls just enough to align the item's top with the viewport's top.
- **If the item is below the viewport:** The list scrolls just enough to align the item's bottom with the viewport's bottom.

This creates a much smoother experience in the Launcher and other list-heavy interfaces where users frequently toggle between adjacent items.

### Self-Review Checklist:

- [x] Tested locally on a project

### Release Notes:

- Fixed "jumpy" list scrolling by ensuring items only trigger a scroll when they are actually outside the visible area.

### DIFF
--- a/crates/gpui/src/elements/list.rs
+++ b/crates/gpui/src/elements/list.rs
@@ -442,31 +442,39 @@ impl ListState {
     /// Scroll the list to the given item, such that the item is fully visible.
     pub fn scroll_to_reveal_item(&self, ix: usize) {
         let state = &mut *self.0.borrow_mut();
-
         let mut scroll_top = state.logical_scroll_top();
-        let height = state
+        let viewport_height = state
             .last_layout_bounds
             .map_or(px(0.), |bounds| bounds.size.height);
         let padding = state.last_padding.unwrap_or_default();
 
-        if ix <= scroll_top.item_ix {
+        // calculate top coordinate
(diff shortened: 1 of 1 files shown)


smitbarmase:
Thanks, but could you provide a reproduction example first (as Github Issue) that I can test against? I’m not able to reproduce this in Zed at the moment, though a GPUI example would also work for investigating this further.

For a change like this, I also think we should have test coverage for the different cases involved, such as when the item is above the viewport, partially visible from the top, partially visible from the bottom, and so on.

Closing it for now.

PR #52568, opened 2026-03-27
Fix unformatted error contexts

jakcharvat:
When debugging a remote SSH connection, I came across an unformatted format string in the output log. I changed the raw `.context(fmt)` call to a `.with_context(|| format!(fmt))`. I ran a quick sweep through the codebase to identify and fix two other instances of the same issue.

Self-Review Checklist:

Release Notes:

- N/A


SomeoneToIgnore (approved):
Nice catch, thank you!

PR #52575, opened 2026-03-27
language_core: Introduce fallback highlights

MrSubidubi:
Self-Review Checklist:

Release Notes:

- Added the option for highlights from languages to specify fallbacks. That means that if you have a pattern with the captures `@second.capture @first.capture`, Zed will first try resolving a highlight from your theme for the code fragment using the first capture, then look for the second capture if no match for the first capture could be found.


MrSubidubi:
> Don't you mean "... then look for the second capture if no match for the first capture could be found"?

Yes, I very much do 😅 Thank you!

SomeoneToIgnore (approved):

PR #52594, opened 2026-03-27
vim: Fix insert above auto-indentation

143mailliw:
Fixes InsertLineAbove (Shift + O) auto-indent handling by deciding whether to trim the first character or the last character based on direction. Previously, inserting a line above would auto format the line the cursor was originally on and not the new line (instead of the correct behavior which is to do the opposite).

No tests because I couldn't find any Vim-specific auto-indent tests and this only affects Vim mode. If I'm missing something there LMK and I'll take a closer look.

Closes #52588.

Release Notes:

- Vim: Fixed auto-indentation for insert above action.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -3275,7 +3275,7 @@ impl Editor {
         });
     }
 
-    pub fn edit_bottom_up_with_autoindent<I, S, T>(&mut self, edits: I, cx: &mut Context<Self>)
+    pub fn edit_before_with_autoindent<I, S, T>(&mut self, edits: I, cx: &mut Context<Self>)
     where
         I: IntoIterator<Item = (Range<S>, T)>,
         S: ToOffset,
@@ -3286,7 +3286,7 @@ impl Editor {
         }
 
         self.buffer.update(cx, |buffer, cx| {
-            buffer.edit_bottom_up(edits, self.autoindent_mode.clone(), cx)
+            buffer.edit_before(edits, self.autoindent_mode.clone(), cx)
         });
     }
 
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -3275,6 +3275,21 @@ impl Editor {
         });
     }
 
+    pub fn edit_bottom_up_with_autoindent<I, S, T>(&mut self, edits: I, cx: &mut Context<Self>)
+    where
+        I: IntoIterator<Item = (Range<S>, T)>,
+        S: ToOffset,
+        T: Into<Arc<str>>,
+    {
+        if self.read_only(cx) {
+            return;
+        }
+
+        self.buffer.update(cx, |buffer, cx| {
+            buffer.edit_bottom_up(edits, self.autoindent_mode.clone(), cx)
+        });
+    }
+
     pub fn edit_with_block_indent<I, S, T>(
         &mut self,
         edits: I,
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -950,6 +950,12 @@ impl EditPreview {
     }
 }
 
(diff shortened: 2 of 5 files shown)


numbereddev:
I can confirm against my issue #52588 that the behavior now works consistently and provides the expected result each time under all cases I could think of.

liitx (approved):

numbereddev (approved):

phdavis1027:
I have changes which fix #53570 but they obviously touch the same code as this PR. For that reason, for my own debugging, I cherry-picked e04ef2a6462204cb4245fb28b684dd4982c8a56f and made my changes against it, but I just wanted to double check how y'all would prefer I submit them -- just wait to submit if / until this one gets merged?

ConradIrwin:
Thanks for sending this, and sorry it sat in the pile for so long.

I think this may have been fixed by https://github.com/zed-industries/zed/pull/53620? Sorry for missing this PR, and if there are more fixes we should land; please open a new PR and @-me so I can take a look.

143mailliw:
> Thanks for sending this, and sorry it sat in the pile for so long.
> 
> I think this may have been fixed by #53620? Sorry for missing this PR, and if there are more fixes we should land; please open a new PR and @-me so I can take a look.

This doesn't seem to be the case: #53620 fixes auto-indentation occurring when it shouldn't at all, but this PR is an unrelated fix that resolves auto-indentation occurring on the wrong line when inserting above.

I am still able to reproduce the issue this PR resolves on the latest nightly (1.3.0+nightly.2494.367db070)

ConradIrwin:
sorry about that! Would you mind rebasing, and I'll take a look

143mailliw:
@ConradIrwin Rebased and tests were added to verify the new behavior.

Sorry it took me a while to get around to this, it's been a rough few months for me personally. The issue was closed as fixed in #55459 but it's a narrow fix that doesn't fully resolve the issue.

The fix in #55459 checked if the insert position was } or ) (theoretically the end of the block) and if so moves the indent row. This isn't a correct fix because it doesn't actually solve the problem, it just introduces a hack that solves one specific case. The test cases I added now fail in Zed today and don't in Vim.

The fix that I implemented flips the insertion range upside down when the insertion starts opposite normal reading direction, so this results in correct auto-indentation regardless of the content of the line the cursor is on.

dinocosta (changes requested):
@143mailliw Thank you for working on this! 🙂

The overall approach makes sense to me but I would like to tighten the abstraction before merging, namely:

- I feel that the `edit_bottom_up` name can be misleading because the edits are not applied bottom-to-top or reversed, it only changes which newline boundary is excluded from auto-indentation. As such, I'm wondering if a name like `edit_before_with_autoindent` would be clearer at the `Editor`/`Vim` layer while, at the `Buffer` layer having an explicit policy like `AutoIndentBoundary::{LeadingNewLine, TrailingNewLine}` would describe the behavior more accurately.
- Now that the auto-indent request targets the correct line, I think we should be able to remove the existing `}`/`)` special case in `insert_line_above`.
- I think we'll also want to ensure this works correctly with the different `auto_indent` values. For example, with `preserve_indent`, in the following snippet, using `O` in the `}` should place the cursor just above `}` with no space added, which isn't yet the case:

  ```rust
  // Before
  fn test() {
      println!("");
  ˇ}
  ```

  ```rust
  // After
  fn test() {
      println!("");
  ˇ
  }
  ```

Thank you and let me know if you have any more questions or need help!

143mailliw:
Sounds reasonable, I'll address these over the weekend.

143mailliw:
@dinocosta Should be good now. I think the issue you found was because of the previous fix attempt; after removing that I couldn't reproduce it, but I added the test anyways.

All the concerns you mentioned have been addressed.

dinocosta (approved):
Thank you @143mailliw ! 🚢

PR #52611, opened 2026-03-28
editor: Fix selection occurrence highlights during regex buffer search

saberoueslati:
## Context

When using regex buffer search (e.g. `^something`), Zed correctly navigates only to actual matches. However, navigating to a match selects the matched text, which triggers the **selection occurrence highlight** feature. That feature performs a *plain literal* search for the selected word and highlights all occurrences, including ones that don't satisfy the regex. The user would see a 4th mid-line occurrence highlighted, even though `^something` never matched it, this behavior is quite confusing.

The fix tracks whether the current selection was set by search navigation via a new `from_search: bool` on `SelectionEffects`. When `last_selection_from_search` is set and `BufferSearchHighlights` are active, selection occurrence highlights are suppressed. Making a manual text selection during an active search clears the flag, restoring normal occurrence-highlight behavior. 

The behavior now matches how VSCode handles this case.

The video below demonstrates the behavior after the fix and shows that it matches the VSCode behavior now

[Screencast from 2026-03-28 01-33-46.webm](https://github.com/user-attachments/assets/07a005b8-53b1-4abf-93d2-96406f0b6a11)

Closes #52589

## How to Review

Three files changed — read in this order:

1. **`crates/editor/src/editor.rs`**  Adds `from_search: bool` to `SelectionEffects` (with a builder method) and `last_selection_from_search: bool` to `Editor`. `selections_did_change()` records the flag from effects. `prepare_highlight_query_from_selection()` returns `None` early when both `last_selection_from_search` and active `BufferSearchHighlights` are set.

2. **`crates/editor/src/items.rs`**  `activate_match()` now calls `.from_search(true)` on its `SelectionEffects` so search-driven selections are marked at the source.

3. **`crates/search/src/buffer_search.rs`**  Regression test `test_regex_search_does_not_highlight_non_matching_occurrences`: verifies that after search navigation, `SelectedTextHighlight` is suppressed and exactly 3 `BufferSearchHighlights` exist; and that after a manual selection, `SelectedTextHighlight` is restored.

## Self-Review Checklist

Release Notes:

- Fixed regex buffer search highlighting non-matching word occurrences via the selection occurrence highlight feature

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1216,6 +1216,7 @@ pub struct Editor {
     quick_selection_highlight_task: Option<(Range<Anchor>, Task<()>)>,
     debounced_selection_highlight_task: Option<(Range<Anchor>, Task<()>)>,
     debounced_selection_highlight_complete: bool,
+    last_selection_from_search: bool,
     document_highlights_task: Option<Task<()>>,
     linked_editing_range_task: Option<Task<Option<()>>>,
     linked_edit_ranges: linked_editing_ranges::LinkedEditingRanges,
@@ -1504,6 +1505,7 @@ pub struct SelectionEffects {
     nav_history: Option<bool>,
     completions: bool,
     scroll: Option<Autoscroll>,
+    from_search: bool,
 }
 
 impl Default for SelectionEffects {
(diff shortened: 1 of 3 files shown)


ChristopherBiscardi:
Thank you! This PR *does* fix the issue, and has comments and tests, so I'm going to move forward with merging it. Showing text that is selected by a user *in addition* to a search is also useful and uninhibited by this PR, so fully disabling text selection feels like the wrong approach. However, if we want to take a different approach in the future we can do so.

ChristopherBiscardi:
@zed-industries/approved

PR #52545, opened 2026-03-27
agent_ui: Fix message editor not expanding after sending a message

Dnreikronos:
Context

Tracks the previous `v2_empty_state` so `set_mode()` only fires on actual state transitions, not every render frame.

Closes #52424

### Before:

https://github.com/user-attachments/assets/76b61861-cebc-44ce-b483-596eeed19bb1

### After:

https://github.com/user-attachments/assets/9da9f3bc-2fc0-4182-8712-4f42d108650b

## How to review

1. `crates/agent_ui/src/conversation_view/thread_view.rs` — adds a `was_v2_empty_state` field to gate the `set_mode()` call in `render_message_editor()` so it only runs on transitions
2. `crates/agent_ui/src/agent_panel.rs` — test verifying that manually setting the editor to Full mode survives a render cycle without being reset back to AutoHeight

## Self-review checklist

- [x] I've reviewed my own diff for quality, security, and reliability
- [ ] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Release Notes:

- Fixed agent chat input box not expanding after sending a message (#52424)

PR #52612, opened 2026-03-28
agent_ui: Fix expanded message editor jitters while typing

Charlie-XIAO:
Self-Review Checklist:

Closes #52132.

*This PR previously also fixes the "expanded message editor not taking up full height" part, but #52545 has already fixed that. Yet it seems to leave a new issue (that was previously not revealed) behind, as follows.*

This PR fixes the `Full + ExcludeOverscrollMargin` editor mode (which the agent panel message editor uses in expanded mode), which could jitter while typing because render-time layout and scroll-position updates were clamping against different effective `scroll_beyond_last_line` policies. This PR fixes that inconsistency so the expanded editor stays stable while typing, and adds a regression test covering `ExcludeOverscrollMargin` scroll clamping.

https://github.com/user-attachments/assets/86abf04d-c1a1-419b-96d0-8ca097c0acb0

https://github.com/user-attachments/assets/03dbdc3c-f58e-4378-8c6a-4bda1ae425c8

Release Notes:

- Fixed the expanded Agent Panel message editor so it no longer jitters while typing.


Charlie-XIAO:
Re-scoped the PR.

MrSubidubi (approved):
Nice find and fix, thanks for this and adding a test upfront!

PR #52622, opened 2026-03-28
project: Fix CSS nested selector document highlights

MoonMao42:
Closes #51483

When the cursor is on a nested tag selector inside a selector chain like:

```css
div {
  div p {
    color: red;
  }
}
```

Zed currently highlights the parent `div` as if it were part of the same symbol.


MoonMao42:
Don't contrubute for this repo.

PR #52641, opened 2026-03-28
gpui: Fix BGRA conversion for SVG rendering

nakashima-hikaru:
### Description
Fixes swapped red/blue channels when rendering SVG images.

#### Describe the bug

When rendering a full-color SVG into an Image object using Image::from_bytes(ImageFormat::Svg, ...) on macOS, the resulting bitmap has its Red and Blue channels swapped. For example, a color specified as #38BDF8 (Light Blue) in the SVG source appears as yellowish in the rendered GPUI view. 

#### Steps to reproduce
1. Create a GPUI application. 
1. Generate or load an SVG string containing a specific color, for example:
```xml
   <svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
     <rect width="100" height="100" fill="#38BDF8"/>
   </svg>
```
3. Load this SVG into an Image object:
```rust
   let image = Arc::new(Image::from_bytes(
       ImageFormat::Svg,
       svg_string.into_bytes(),
   ));
```
4. Display this image in a view using an img() element. 

#### Expected behavior
The rectangle should be rendered in **Light Blue (#38BDF8)**. 

#### Actual behavior
The rectangle is rendered in **Yellowish Color (#F8BD38)**.

### Self-Review Checklist:

### Release Notes:

- Fixed swapped color channels when pasting SVG images from the clipboard.

### DIFF
--- a/crates/gpui/src/elements/img.rs
+++ b/crates/gpui/src/elements/img.rs
@@ -697,7 +697,7 @@ impl Asset for ImageAssetLoader {
                 Ok(Arc::new(RenderImage::new(data)))
             } else {
                 svg_renderer
-                    .render_single_frame(&bytes, 1.0, true)
+                    .render_single_frame(&bytes, 1.0)
                     .map_err(Into::into)
             }
         }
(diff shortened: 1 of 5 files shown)


nakashima-hikaru:
@cla-bot check

MrSubidubi (approved):
Nice find and very much appreciate the test you added for this! 

I went ahead and removed the parameter altogether, given that it is now `true` throughout the codebase and I think that other consumers would probably expect the image to always be in the proper color space.

In any case, thank you for this and congratulations to your first contribution! 🎉

Issue #22473, opened 2024-12-28
Editor doesn't show errors that are past the end of a line


injust:
### Describe the bug / provide steps to reproduce it

```python
foo = []
for bar in foo
    pass

foo = []
for bar in foo^
    pass
```

The overview in the scrollbar shows the errors, even though they do not appear in the editor:

<img width="567" alt="image" src="https://github.com/user-attachments/assets/6712c4a8-bf29-4124-848a-644d3641782f" />

Both language servers correctly report the errors:

<img width="316" alt="image" src="https://github.com/user-attachments/assets/827567e6-fe3a-4b7f-8cf1-4f3a317b92c4" />

VS Code shows the errors correctly with the same language servers. Note how the errors are reported with the location _past_ the end of the line:

<img width="177" alt="image" src="https://github.com/user-attachments/assets/77f13692-d3d6-40e2-ae60-2e0240be5ca9" />

### Zed Version and System Specs

Zed: v0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1 (Zed) 
OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

CoderParth:
Hi @injust!
Could you please try the following?
```
cd ~/.local/share/zed/languages
rm -rf pylsp
```
Please restart your Zed after this. The above command removes the pylsp, and it will be reinstalled once you reboot your zed again. This may help you solve your issue. Please let us know how it comes out. 

Also, you may want to look at the official [docs](https://zed.dev/docs/languages/python) to tweak your language server for python. 

Below is my screenshot of lsp showing the error correctly. 
I have a vscode theme installed, which is why it may look similar to vs code, but it is actually Zed. 

![Screenshot from 2024-12-29 12-57-53](https://github.com/user-attachments/assets/a23608e1-3849-44a9-a256-89b80fe5e7f1)

Please let us know how you go, and I would be happy to provide further help.

injust:
@CoderParth I don't have a ~/.local/share/zed/` directory. I'm also using pyright and Ruff as my language servers, not pylsp.

My language servers are correctly configured, as the errors are correctly reported. Please see the screenshots I included.

You are seeing different errors because you are using pylsp.

injust:
Bad bot

notpeter:
For reference the `textDocument/diagnostic` response from the LSP RPC logs has the following:

`result.items`:
```json
[
  { 
[...]

PR #52651, opened 2026-03-29
editor: Fix LSP extension commands failing at end of file

florian-trehaut:
## Context

Closes #51330

When the cursor is at the very end of a file, `find_specific_language_server_in_selection` in `crates/editor/src/lsp_ext.rs` silently skips the selection because the anchor's `buffer_id` is `None`. This causes `editor: switch source header` (clangd) and rust-analyzer extension commands (`expand macro`, `open docs`, `open playground`) to do nothing.

The fix falls back to the singleton buffer's ID when the anchor has no `buffer_id`.

## How to Review

Single file change in `crates/editor/src/lsp_ext.rs`. The diff is small — pre-compute the singleton buffer ID, then use it as fallback in the `filter_map` closure. An integration test verifies the fix.

## Self-Review Checklist

- [x] Reviewed own diff for quality, security, and reliability
- [x] No unsafe blocks
- [x] Tests pass (639 editor tests, 0 failures)
- [x] Manual testing: switch source header works at beginning, middle, and end of C++ file

## Test Plan

- [x] New integration test: `test_find_language_server_at_end_of_file` — verifies `find_specific_language_server_in_selection` returns `Some` at both beginning and end of file
- [x] Confirmed the test fails without the fix (assertion on "should find language server at end of file")
- [x] Manual: open a C++ file with clangd, place cursor at very end, run `switch source header` — now correctly opens the header
- [x] Manual: verify it still works at beginning and middle of file (non-regression)
- [x] `cargo test -p editor` — 639 passed, 0 failed

Release Notes:

- Fixed `editor: switch source header` and other LSP extension commands not working when the cursor is at the very end of a file.

### DIFF
--- a/crates/editor/src/lsp_ext.rs
+++ b/crates/editor/src/lsp_ext.rs
@@ -33,11 +33,20 @@ where
     F: Fn(&Language) -> bool,
 {
     let project = editor.project.clone()?;
+    let singleton_buffer_id = editor
+        .buffer()
+        .read(cx)
+        .as_singleton()
+        .map(|buffer| buffer.read(cx).remote_id());
     editor
         .selections
         .disjoint_anchors_arc()
         .iter()
-        .filter_map(|selection| Some((selection.head(), selection.head().text_anchor.buffer_id?)))
+        .filter_map(|selection| {
+            let head = selection.head();
+            let buffer_id = head.text_anchor.buffer_id.or(singleton_buffer_id);
+            Some((head, buffer_id?))
+        })
         .unique_by(|(_, buffer_id)| *buffer_id)
         .find_map(|(trigger_anchor, buffer_id)| {
             let buffer = editor.buffer().read(cx).buffer(buffer_id)?;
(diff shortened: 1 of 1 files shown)


crates/editor/src/lsp_ext.rs
```diff
@@ -173,3 +182,90 @@ pub fn lsp_tasks(
...
+        // Cursor at end of file — should still find server.
+        editor.update(cx, |editor, cx| {
+            let result = find_specific_language_server_in_selection(
+                editor,
+                cx,
+                filter,
+                language_server_name.clone(),
+            );
+            assert!(
+                result.is_some(),
```
SomeoneToIgnore: Here and above: let's assert that this is a particular server we've got, not just something.
Presumably, can be done in another closure to keep the assertion logic shared?

We have a `_fake_server` from where we can get an id at least, maybe something more; we can also be sure that the buffer has a path and a language that we expect.

crates/editor/src/lsp_ext.rs
```diff
@@ -33,11 +33,20 @@ where
...
+        .as_singleton()
+        .map(|buffer| buffer.read(cx).remote_id());
     editor
         .selections
         .disjoint_anchors_arc()
         .iter()
-        .filter_map(|selection| Some((selection.head(), selection.head().text_anchor.buffer_id?)))
+        .filter_map(|selection| {
+            let head = selection.head();
+            let buffer_id = head.text_anchor.buffer_id.or(singleton_buffer_id);
```
SomeoneToIgnore: This fix seems a bit odd, as the method deals with `multi_buffer::Anchor` and may be used to get some data in a multi buffer's context (end selection there presumably produces the same result).

What seems to be more appropriate to do is `or_else(|| editor.buffer().read(cx).excerpt_containing(selection.head(), cx))`-ish change.
That will cause a bunch of borrow checker issues though, so presumably `find_map` below has to be merged into this closure and `unique_by` accommodated differently — this way all filtering logic stay in the same single block with `cx` and no borrowing issues should be left.

Besides this, it seems reasonable to do this all only if `head.is_max()`, and, maybe, falling back to `tail` is not that bad either.

SomeoneToIgnore (changes requested):
Thank you for the fix, this is a very nice catch overall meaning there's a lot of places like this elsewhere.
The most important one most probably is the diagnostics-related code as there were reports of diagnostics not shown on the very edge of the file?

Either way, let's adjust the code for a proper multi buffer coordinate contract `find_specific_language_server_in_selection` exposes and merge this.

crates/editor/src/lsp_ext.rs
florian-trehaut: Fixed. The test now uses a shared `assert_result` closure that checks server ID, language name, buffer identity, and that the buffer has an associated file.

crates/editor/src/lsp_ext.rs
florian-trehaut: Reworked. Replaced `resolve_anchor_to_buffer` with `MultiBufferSnapshot::buffer_id_for_anchor` which handles the `excerpt_containing` fallback internally. Everything is in a single `find_map` now, with a scoped block to drop the multi-buffer borrow before `buffer.update`. Tail fallback kept.

SomeoneToIgnore (approved):
Thank you, I have pushed another small fix on top to avoid extra snapshot-related creation costs.

--------------

I've rechecked the diagnostics-related idea of mine and that does not seem related to this particular issue: https://github.com/zed-industries/zed/issues/22473

Yet, I think if you check for `text_anchor.buffer_id` field usages, some of them may come out wrong, e.g. 

https://github.com/zed-industries/zed/blob/63326552b6528cabf557173107740c0d5500b293/crates/editor/src/document_symbols.rs#L85-L88

or even

https://github.com/zed-industries/zed/blob/63326552b6528cabf557173107740c0d5500b293/crates/multi_buffer/src/multi_buffer.rs#L2386-L2399

are quite suspicious already.

PR #52655, opened 2026-03-29
git_ui: Follow the active repository in branch picker

cppcoffee:
git_ui: Follow the active repository in branch picker

Reload branch picker data when it is created without an explicit
repository and the workspace later resolves an active repository.
Subscribe to git store repository changes, reset picker state on repo
switches, and ignore stale async branch loads.

Add coverage for recovering when the picker starts without an initial
repository.

Release Notes:

- Add follow the active repository in branch picker

before:

截屏2026-03-29 17 04 30

after:

https://github.com/user-attachments/assets/3cf48f0e-0f82-4c28-a029-534f28f94fbe

### DIFF
--- a/crates/git_ui/src/branch_picker.rs
+++ b/crates/git_ui/src/branch_picker.rs
@@ -1,4 +1,3 @@
-use anyhow::Context as _;
 use editor::Editor;
 use fuzzy::StringMatchCandidate;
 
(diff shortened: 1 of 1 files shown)


kubkon:
Hey @cppcoffee , thanks for the patch! I've cherry-picked your test into `main` to see if it passes given recent changes to the branch picker but it seems to still fail. However, before fixing conflicts with `main`, could you double check this patch is still needed? What are the actual repro conditions for the bug? Reading over the test it seems that if I somehow force the branch picker to get constructed without active repository which I cannot seem to do.

cppcoffee:
Hi @kubkon ， I'll close this PR for now, and reopen it if it happens again.

PR #52669, opened 2026-03-29
gpui: Add wgpu Surface element and GPU device requirements API for Linux

zortax:
Adds support for embedding external wgpu GPU textures in gpui's scene on Linux, mirroring the existing CVPixelBuffer-based Surface element on macOS. Also introduces a WgpuDeviceRequirements API that lets applications request additional wgpu features and limits on top of gpui's baseline, without changing the defaults.

All new APIs are gated to cfg(any(target_os = "linux", target_os = "freebsd")).

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/elements/surface.rs
+++ b/crates/gpui/src/elements/surface.rs
@@ -2,16 +2,55 @@ use crate::{
     App, Bounds, Element, ElementId, GlobalElementId, InspectorElementId, IntoElement, LayoutId,
     ObjectFit, Pixels, Style, StyleRefinement, Styled, Window,
 };
+#[cfg(any(target_os = "linux", target_os = "freebsd"))]
+use crate::{DevicePixels, Size};
 #[cfg(target_os = "macos")]
 use core_video::pixel_buffer::CVPixelBuffer;
 use refineable::Refineable;
+#[cfg(any(target_os = "linux", target_os = "freebsd"))]
+use std::sync::Arc;
 
 /// A source of a surface's content.
-#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum SurfaceSource {
     /// A macOS image buffer from CoreVideo
     #[cfg(target_os = "macos")]
     Surface(CVPixelBuffer),
+    /// A GPU texture handle (type-erased to avoid depending on wgpu)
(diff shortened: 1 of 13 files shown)


crates/gpui/src/elements/surface.rs
```diff
@@ -2,16 +2,55 @@ use crate::{
...
+use std::sync::Arc;
 
 /// A source of a surface's content.
-#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum SurfaceSource {
     /// A macOS image buffer from CoreVideo
     #[cfg(target_os = "macos")]
     Surface(CVPixelBuffer),
+    /// A GPU texture handle (type-erased to avoid depending on wgpu)
+    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
```
ForsakenHarmony, line 19: Would it make sense to have this available on all platforms?

Seems like a way to have this work with the same API everywhere (linux, macos, windows)

zortax: yeah honestly I'm not sure about this dyn Any stuff.. Maybe it would be better to have this all properly typed and adding optional wgpu dependency is probably fine, since you are compiling against it anyway at the end of the day... 

Would appreciate more feedback on this :)

ForsakenHarmony: yeah probably fine

SomeoneToIgnore:

image

zortax:
@SomeoneToIgnore hey, is the zed team still interested in this? I think this is quite a relevant PR to zed as well, it will enable proper hardware-accelerated video playback (e.g. for the screensharing stuff) on linux. I can take the time to update the branch and resolve the conflicts, if someone from the zed team is willing to give feedback or collab on the API I proposed (the gpu spec/feature stuff is required to actually make this useful for third-party GPUI users).
I also already have a hacky reference implementation for how to use this for hw decoded video [here](https://github.com/zortax/gpui-video).

SomeoneToIgnore:
I'm afraid I have no clue, yet if there would be a person qualified and interested enough, it could be @reflectronic — pinging them in case they want to chime in.

smirzaei:
Just wanted to add that as someone who is playing around with GPUI on side projects, this would be a very useful addition and I will help in any way I can. (I have almost no experience in that area though)

PR #52671, opened 2026-03-29
Fix tab switcher random jumps

### LINKED ISSUE #52579: ctrl+tab jumps to random documents when the mouse is moved
### Reproduction steps

1. Open multiple files, so that the tab switcher is not empty.
2. Click and immediately release <kbd>⌃</kbd><kbd>⇥</kbd>, while simultaneously moving the mouse over the area where tab switcher appears.

**Current:** I'm thrown into a random file that happened to be under the cursor.

https://github.com/user-attachments/assets/be6bdcbc-0357-4098-94de-666ad035e0d0

This is my main method for navigation between open files, and on the nightly, I'm thrown to the wrong file 90% of the time.

**Expected:** I'm switched to the recently focused file, regardless of what I'm doing with the mouse.

There should probably be some delay before the tab switcher appears, so that quick <kbd>⌃</kbd><kbd>⇥</kbd> does not show the UI at all. Here's an example of how it works in my window switcher:

https://github.com/user-attachments/assets/8cfaa232-c83d-4db3-8280-927098409424

You can see that if I hold <kbd>⌘</kbd><kbd>⇥</kbd>, the window switcher appears and starts reacting to my cursor. However, if I press and release it fast enough, the UI does not even appear, allowing me to freely move the mouse without causing random items to be selected.

Zed: v0.230.0+preview.205.9437a84390a396d666f04b38db87d89bb07284c1 (Zed Preview) 
OS: macOS 26.2
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

OmChillure:
#### Fixes #52579

When quickly pressing and releasing Ctrl+Tab, the tab switcher popover would briefly appear, and any mouse movement over it would select a random tab instead of the intended recently-focused file.

### Fix : 
This adds a 300ms delay before showing the tab switcher UI when opened via modifier keys. If the modifier is released before the delay elapses, the selected item is confirmed directly without ever showing the popover. This matches the behavior of system window switchers (e.g. macOS Cmd+Tab), where quick presses never flash the UI.

- `visible` and `_show_task` fields track whether the popover should be rendered
- When not yet visible, the picker is mounted in a zero-size overflow-hidden container to preserve focus and keyboard state
- When modifiers are released before the popover is visible, `confirm()` is called directly instead of dispatching `menu::Confirm`, avoiding any mouse-hover-based selection

### Self-Review Checklist

- [x] Unsafe blocks (if any) have justifying comments — *N/A, no unsafe blocks*
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) — *Matches system window switcher behavior (e.g. macOS Cmd+Tab); no visual change for slow/held switches*
- [x] Tests cover the new/changed behavior — *Added `test_quick_switch_before_popover_visible`*
- [x] Performance impact has been considered and is acceptable — *Only adds a lightweight 300ms timer task; no extra allocations or rendering cost*

### Video 

[Screencast from 2026-03-29 11-52-07.webm](https://github.com/user-attachments/assets/8e148a74-75ae-4c36-8d44-594c3c9097e6)

### Note : Reopens previous work from closed PR #52648 (fork was deleted by mistake)

Release Notes:

- Fixed Ctrl+Tab jumping to random documents when the mouse is moved during a quick tab switch


jrmajor:
After 3 months I still haven't gotten used to it, I keep jumping to random files all the time. To be honest, I still believe that we should restore the separation between keyboard focus and mouse hover, like in other editors, but thanks for keeping this PR up to date — this will make it a bit more tolerable.

crates/tab_switcher/src/tab_switcher.rs
```diff
@@ -169,6 +173,19 @@ impl TabSwitcher {
...
+                cx.background_executor().timer(POPOVER_DELAY).await;
+                this.update_in(cx, |this, _window, cx| {
+                    this.visible = true;
+                    cx.notify();
+                })
+                .ok();
+            }))
+        } else {
+            None
+        };
```
osiewicz: Nit: we can use `has_modifiers.then(|| ...)`.

osiewicz (approved):
LGTM, sans a minor nit.

osiewicz (approved):

jerrygreen:
Side question – is there any way to tell the exact version when this feature gets landed? Anywhere in GitHub UI? Like maybe in one of the referenced tasks? GitHub actions? Or any other way?

This can be useful in package managers, because they aren’t updated immediately and it would be useful to know which exact version is worth updating.

osiewicz:
@jerrygreen sadly, no, not at the moment. I've always wanted such a feature for our GH automation. :/
It looks like we've cut a Preview of 1.20 today. So.. the version worth updating to is 1.20. :)

hazicy:
@OmChillure @osiewicz I think the delay starts from the wrong event. If Cmd has already been held for 300ms, pressing Tab should show the switcher immediately. Currently, the timer starts after the first Tab press, forcing an initial switch followed by another 300ms wait, which feels inconsistent with the system Cmd+Tab behavior.

PR #52677, opened 2026-03-29
Fix numbering for indented list

### LINKED ISSUE #48244: Zed incorrectly continues parent numbering for indented/nested list items instead of restarting at 1

OmChillure:
Fixes #48244 

Fix ordered list renumbering on indent and outdent in Markdown

### Summary

When tabbing (indenting) or shift-tabbing (outdenting) ordered list items in a Markdown file, Zed previously left the list numbers unchanged — resulting in broken sequences like `1. item` → `  2. item` after indenting, or `  2. child` → `2. child` after outdenting (skipping the correct continuation number).

This PR fixes both indent and outdent to correctly renumber ordered list markers:

**Indent (`Tab`)**
- Single-line indent resets the marker to `1.` (start of a new nested list)
- Multi-row indent renumbers selected items sequentially starting from `1.`
- After indenting, trailing siblings at the original indent level are renumbered to close the gap

**Outdent (`Shift+Tab`)**
- Outdented items pick up the correct next number by scanning backward for the last item at the target indent level
- Multiple selected rows share a running counter, incrementing from where the parent list left off
- Trailing siblings at the deeper (original) indent are renumbered to fill the hole left behind
- Trailing siblings at the shallower (target) indent are renumbered to account for the newly inserted items

### Implementation

Added four new helper functions in `crates/editor/src/editor.rs`:
- `ordered_list_marker` — parses an ordered list marker at a given row/indent using the language's `ordered_list` config
- `ordered_list_indent_renumber` — thin wrapper used during indent to get `(start_col, end_col, new_text)` for a marker
- `previous_ordered_list_number_at_indent` — scans backward (up to 256 lines) to find the last list number at a given indent level
- `renumber_following_ordered_list_siblings_after_indent` — scans forward from a row, renumbering sibling items at the same indent level

The outdent path was also refactored from building a separate `deletion_ranges` vec to a unified `edits: Vec<(Range<Point>, String)>` so that marker-renumber edits can be emitted alongside the deletion edits in a single `buffer.edit()` call.

### Tests

Two new async tests added to `crates/editor/src/editor_tests.rs`:

**`test_tab_ordered_list_single_line_renumbering`** (7 cases)
- Single-line indent resets to `1.`
- Single-line indent of empty item resets to `1.`
- Multi-row indent renumbers `2., 3.` → `1., 2.`
- Multi-row indent after an existing nested list resets correctly
- Trailing sibling at original indent is renumbered to close the gap
- Selection starting before the moved ordered rows still renumbers trailing sibling
- Blank lines above the moved block don't reset trailing list continuation

**`test_tab_list_outdent`** (6 cases)
- Outdenting a sub-item continues parent list numbering
- Outdenting after two parent items picks up the next number in sequence
- Outdenting with no preceding parent at the target indent starts at `1.`
- Outdenting a top-level item with no indentation is a no-op
- Outdenting an empty ordered list item renumbers correctly
- Outdenting multiple rows uses a shared running counter

#### Release Notes:

- Fixed ordered list markers not being renumbered when indenting or outdenting list items in Markdown

#### Video

https://github.com/user-attachments/assets/eb64aa51-60aa-47a7-ae44-ac54f30429de

### Note : Reopens previous work from closed PR #51294 (fork was deleted)


dinocosta (changes requested):
Hey @OmChillure 👋 

Thank you for taking a stab at improving this, I checked out your branch and was able to confirm that these changes work as expected. However, I believe we might need to revisit the approach.

I believe we'll want to try and avoid having this Markdown-specific logic live in both `Editor::indent_selection` and `Editor::outdent`. Since the keymap functionality already supports scoping by extension, for example:

```
{
  "context": "Editor && extension == md",
  "use_key_equivalents": true,
  "bindings": {
    "cmd-k v": "markdown::OpenPreviewToTheSide",
    "cmd-shift-v": "markdown::OpenPreview",
  },
}
```

The cleaner approach would likely be to introduce two new `markdown`-specific actions, `markdown::Indent` and `markdown::Outdent`, that are fired when `tab`, `shift-tab`, `ctrl/cmd-]` and `ctrl/cmd-[` are used while in a Markdown buffer.

We could then have the markdown-specific logic live in the handlers for those actions, dispatching to the existing implementation in case the user is not indenting or outdenting a list item.

Let me know what you think or if you need more pointers! 🙂

dinocosta (commented):
Thank you for tackling the suggested changes @OmChillure ! 🙂

I'm taking another look at these changes and doing some minor changes to it, will see if I can get it ready to merge tomorrow. Thanks!

dinocosta (commented):
Hey @OmChillure ,

After taking some more time looking at these changes, I feel like we might be taking the wrong approach here.

The current implementation is a little confusing and brittle, as anytime we indent or outdent, we'll need to look at the previous rows in the buffer, see if we find a list item marker number and then use that to insert new edits that should be applied in the buffer. At the same time, that also means we need to take into consideration that we might have list item markers further down the buffer that belong to the same list that was just edited and edit those too.

Unfortunately this leads to a lot of complicated logic that makes it hard to follow what's actually happening, especially because, at any point in time, the content being indented/outdented might not even be a list.

I've also been able to find a subtle bug where, if you're indenting a selection that spans multiple lists at different levels, the renumbering breaks. For example, consider the following content:

```
1. A
2. B
  1. C
  2. D
3. E
```

If you select the second and third line and then indent, you'll end up with:

```
1. A
  1. B
    2. C
  2. D
2. E
```

Notice `C` becomes `2.` instead of `1.`, it's at a completely different nesting level from `B` but shares its sequential counter. This happens because the existing functionality assumes that the list being updated, if any, is all at the same indent level.

Thinking about this a little bit more, I believe the correct approach is to actually leverage existing tree-sitter functionality to determine whether the cursor (or selections) are inside a `list` or `list_item` node and leverage that information to be able to easily jump to all of the `list_item` nodes and renumber accordingly. I believe that will lead to a much cleaner and straightforward solution.

As such, I'm going to go ahead and close this. I hope this makes sense and thank you for the time you took with this! :)

OmChillure:
> Hey @OmChillure ,
> 
> After taking some more time looking at these changes, I feel like we might be taking the wrong approach here.
> 
> The current implementation is a little confusing and brittle, as anytime we indent or outdent, we'll need to look at the previous rows in the buffer, see if we find a list item marker number and then use that to insert new edits that should be applied in the buffer. At the same time, that also means we need to take into consideration that we might have list item markers further down the buffer that belong to the same list that was just edited and edit those too.
> 
> Unfortunately this leads to a lot of complicated logic that makes it hard to follow what's actually happening, especially because, at any point in time, the content being indented/outdented might not even be a list.
> 
> I've also been able to find a subtle bug where, if you're indenting a selection that spans multiple lists at different levels, the renumbering breaks. For example, consider the following content:
> 
> ```
> 1. A
> 2. B
>   1. C
>   2. D
> 3. E
> ```
> 
> If you select the second and third line and then indent, you'll end up with:
> 
> ```
> 1. A
>   1. B
>     2. C
>   2. D
> 2. E
> ```
> 
> Notice `C` becomes `2.` instead of `1.`, it's at a completely different nesting level from `B` but shares its sequential counter. This happens because the existing functionality assumes that the list being updated, if any, is all at the same indent level.
> 
> Thinking about this a little bit more, I believe the correct approach is to actually leverage existing tree-sitter functionality to determine whether the cursor (or selections) are inside a `list` or `list_item` node and leverage that information to be able to easily jump to all of the `list_item` nodes and renumber accordingly. I believe that will lead to a much cleaner and straightforward solution.
> 
> As such, I'm going to go ahead and close this. I hope this makes sense and thank you for the time you took with this! :)

Hey @dinocosta, thanks for the detailed feedback and the bug example — you're right that the current code mixes counters across nesting levels, that's a real issue.

Quick context on why the current approach looks the way it does: this work originally started in #49017, where **@/smitbarmase** raised the same concern about row-by-row scanning being expensive on large documents. We discussed tree-sitter there, and I checked the grammar — the `r` nodes don't carry the number value (more on that below), so we landed on a bounded backward scan that stops at the first blank line, the first item at a lower indent, or after a hard cap of 256 rows. Smit was good with that approach, so I carried it forward into this PR.

On the tree-sitter angle here — I rechecked node-types.json for the pinned tree-sitter-md grammar:

- list_item has no named fields ("fields": {})
- list_marker_dot and list_marker_parenthesis are leaf nodes with empty fields — they don't expose the integer
So tree-sitter can identify which nodes are ordered list items, but the actual number still has to be read from the buffer text at the marker's byte range.

That said, tree-sitter would cleanly fix the bug you found: walking `list_item` children of a single list node guarantees we never share counters across nesting levels which is exactly the broken assumption right now.

One caveat — at edit-planning time the parse tree is pre-edit, so for the new indent level after a tab we'd still need a small text scan to find the continuation number. Tree-sitter alone can't tell us that.

The markdown::Indent / markdown::Outdent action plumbing from your earlier review is already in place, so the rework would be inside those handlers only. Could I take a shot at this hybrid approach (tree-sitter for list grouping, text read for numbers) before and open a fresh PR?

PR #52686, opened 2026-03-29
eval_cli: Simplify build setup for more datasets

benbrandt:
Cleans up build setup and simplifies it a bunch so that we can use the binary in more eval docker containers

Release Notes:

- N/A


bennetbo:
@zed-zippy approved

PR #52720, opened 2026-03-30
sidebar: Clarify which projects are remote in the UI

danilo-leal:
This PR adds a little server icon to the side of the project label to clarify that's a remote project. This is particularly useful when having the remote and non-remote version of a project open in the same multi-workspace/window.

Screenshot 2026-03-30 at 7  59@2x

Release Notes:

- N/A


benbrandt (approved):
❤️

PR #52724, opened 2026-03-30
acp_thread: Clarify max token limit error message

OmChillure:
When generation stops due to the per-response output limit, Zed was surfacing "Max tokens reached", which implies the full context window was exhausted. In reality, `max_output_tokens` (the per-response cap) may have been hit a different condition.

This change distinguishes between the two cases: if `output_tokens >= max_output_tokens`, it surfaces "Maximum output tokens reached"; otherwise it falls back to "Maximum tokens reached".

Self-Review Checklist:

Closes #50254

Note : Reopens #50372 suggested by @bennetbo 

Release Notes:

- Fixed misleading "Max tokens reached" error by distinguishing between per-response output token limit and total context window limit.


bennetbo (approved):
Pushed a change that also adjusts the logging message. Thank you!

PR #52730, opened 2026-03-30
agent_ui: Set max-width for thread view content

danilo-leal:
This PR adds a configurable max-width to the agent panel. This will be particularly useful when opting into an agentic-first layout where the thread will be at the center of the UI (with the panel most likely full-screen'ed, which is why I'm also adding here the button to make it full screen in the toolbar). The default max-width is 850, which is a bit bigger than the one generally considered as a standard (~66 characters wide, which usually sums up to 750 pixels).

Release Notes:

- Agent: Added a max-width to the thread view for better readability, particularly when the panel is zoomed in.


eholk (approved):

PedroHLC:
You probably dislike Ultra-Wide and 4K displays to set 850 as the default. 🤣

Jokes aside, woke up today to see 1600px of paddings, might not have been the best idea to change it to 850 for those who already resized theirs.

tredondo:
A more serious downside of this is crippling tables to small widths. Run this prompt to repro, and note that clicking the full screen button doesn't increase the width.

```
We're testing Markdown tables in Zed, where you're being run from. Output a fictitious code
review for a TODO app in a table, with several columns:
- Severity
- Type
- Location
- Issue Description (content-rich columsn with detailed description)
- Suggested Fix (content-rich too, fill both with details to show table behavior at different widths)
```

image

Those who land here from the release notes and want to change the 850px width, go to Settings -> Agent panel, and scroll all the way down:

zed://settings/agent.max_content_width

danilo-leal:
There's a setting you can use to customize the width cap: [zed://settings/agent.max_content_width](zed://settings/agent.max_content_width).

PR #36802, opened 2025-08-23
language: Increase cursor match limit

Spissable:
With my [last PR](https://github.com/zed-industries/zed/pull/35657) adding support to run individual Go table sub-tests, I and some other people noticed that it hasn't been working quite as expected, especially for larger table tests. Basically in larger table tests, the detection would only work bottom up, up to a certain point.

After debugging for a while and also confirming that the query seems to work just fine (via [tree-sitter playground](https://tree-sitter.github.io/tree-sitter/7-playground.html)) I realized that the cursor limit is causing it to only find a limited amount of sub-tests.

I made my test example a bit longer to provoke a failure and then bumped the limit to 256 fixed it. Trying it out with some of my real world tests also yielded correct results.

I'm not sure what the golden number could possibly be, but 64 was definitely too low for my day to day use-cases.

If we're worried about performance, I suppose we could also play with `ts_query_cursor_set_timeout_micros` rather than capping the cursor limit at a fixed number.

Release Notes:

- Increased tree-sitter cursor match limit to support more complex queries.

PR #52741, opened 2026-03-30
go: Fix "not all subtests are localized"

### LINKED ISSUE #46881: bug: Go subtests/table-tests are often not detected
### Reproduction steps

Create a `whatever_test.go` with the following content:

```go
func TestSomething(t *testing.T) {
	tests := []struct {
		name string
	}{
		{name: "test_1"},
		{name: "test_2"},
		{name: "test_3"},
		{name: "test_4"},
		{name: "test_5"},
		{name: "test_6"},
		{name: "test_7"},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			fmt.Println(tt.name)
		})
	}
}

```

You should see the "play" buttons against all test cases, but I typically only see them against several (or none at all), e.g.

I can't find the reason(s) for this. Some tests don't have any cases detected, some have all detected, some have several, seemingly at random.

Current:
Image

Expected:
All test cases should be runnable.

Zed: Zed 0.219.4 
OS: macOS 14.4 (arm64)

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

mfattakhov:
Performance impact: increases the memory+cpu occupation for up to 16 times but only for `fn runnable_ranges`, which runs only on scroll/edit. And 16 is the upper limit, usually go table tests have up to 10-15 cases which should be just 2-3x from current state.

Self-Review Checklist:

Closes #46881

Screenshots:

Before:
image

After (locally):
image

Catches:

The proposed limit of 1024 is still finite, thus files with more then ~80 subtests will still fail to show all of the run buttons. But this is a small change that will help in the majority of situations.

Release Notes:

- go: fix a bug with not all of the subtests being runnable

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -3434,7 +3434,7 @@ impl BufferSnapshot {
             &self.text,
             TreeSitterOptions {
                 max_bytes_to_query: Some(MAX_BYTES_TO_QUERY),
-                max_start_depth: None,
+                ..Default::default()
             },
             |grammar| Some(&grammar.indents_config.as_ref()?.query),
         );
@@ -4590,7 +4590,7 @@ impl BufferSnapshot {
                 &self.text,
                 TreeSitterOptions {
                     max_bytes_to_query: Some(MAX_BYTES_TO_QUERY),
-                    max_start_depth: None,
+                    ..Default::default()
                 },
                 |grammar| grammar.brackets_config.as_ref().map(|c| &c.query),
             );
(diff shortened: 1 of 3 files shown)


smitbarmase:
Thanks for picking this up. We actually had #36802 for the same issue, which we closed because bumping `match_limit` (even when scoped to one cursor) weakens the global bound on how much work tree-sitter queries can do during matching, and we want to keep that bound strict. 

The in-progress direction is rewriting the query with repetition and post-processing the multi-capture match into multiple `TaskVariables` downstream. @osiewicz and @dinocosta were paired on that piece, so let me sync with them on where it stands. Closing this for now since we know this approach isn't the right one. Thanks again.

PR #52755, opened 2026-03-30
Skip stale workspace file restores

rtfeldman:
## Summary
Treat restored project file paths as stale when they are no longer present in the worktree snapshot.
Reopen the file as a local buffer if it still exists on disk, and otherwise skip deserializing the editor without logging a startup error.
Add a regression test for missing worktree files and keep the existing non-worktree restore behavior covered.

Release Notes:

- Improved workspace restore to quietly skip stale saved file tabs that no longer exist.

### DIFF
--- a/crates/editor/src/items.rs
+++ b/crates/editor/src/items.rs
@@ -52,7 +52,7 @@ use workspace::{
     },
 };
 use workspace::{
-    Pane, WorkspaceSettings,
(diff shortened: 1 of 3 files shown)


echuber2:
I'm not sure, but this might have already been handled by https://github.com/zed-industries/zed/pull/54224

PR #52756, opened 2026-03-30
Swap KVP store used for collab favorites

mikayla-maki:
We should generally prefer using the normal key value store

Self-Review Checklist:

Release Notes:

- N/A


danilo-leal (approved):
Thank you!

PR #52757, opened 2026-03-30
Remove text thread and slash command crates

benbrandt:
🫡

Self-Review Checklist:

Release Notes:

- Removed legacy Text Threads feature to help streamline the new agentic workflows in Zed. Thanks to all of you who were enthusiastic Text Thread users over the years ❤️!


crates/proto/proto/zed.proto
```diff
@@ -502,6 +493,7 @@ message Envelope {
   reserved 332 to 333;
   reserved 394 to 396;
   reserved 429 to 430;
+  reserved 211 to 216, 232 to 233;
```
bennetbo, line 496: Where does 233 come from?

crates/proto/proto/call.proto
```diff
@@ -370,9 +370,10 @@ message View {
   oneof variant {
     Editor editor = 3;
     ChannelView channel_view = 4;
-    ContextEditor context_editor = 5;
   }
 
+  reserved 5;
```
bennetbo, line 375: Why is this outside of the variant block?

crates/migrator/src/migrations/m_2025_05_05/settings.rs
```diff
@@ -1,18 +1,9 @@
 use std::ops::Range;
```
bennetbo, line 1: Can we just change existing migrations? Wondering if this breaks something because they get applied in order.

bennetbo: Also do we need to write new migrations to remove obsolete keys (e.g. enabled_in_text_threads)

crates/agent_ui/src/conversation_view.rs
```diff
@@ -2524,22 +2524,6 @@ impl ConversationView {
         }
     }
 
-    /// Inserts terminal text as a crease into the message editor.
-    pub(crate) fn insert_terminal_text(
```
bennetbo, line 2528: Why did we remove this?

bennetbo: Never mind, handled in https://github.com/zed-industries/zed/blob/97a139cbc1278ee4df7b05b86416e94bdf724a06/crates/agent_ui/src/completion_provider.rs#L632

bennetbo (approved):

feraxhp:
What a terrible decision. text threads where the best of its kind... this breaks my workflow... I am terribly sad.

JavaGamer:
It be nice to have a non-agentic interface to an llm.

bsutton:
so this change seems to have inadvertently broken slash commands.  As far a I can tell (global) slash commands where only available in the text threads.

The slash commands that I have been developing, as part of a dart language extension , no longer show up in any part of the ui (that I can find).

zvodd:
I have to weigh in here and say this decision or at least the way it has been executed is **kind of insane**, feels
**objectively bad** and **possibly malicious**.

Zed is a text editor right? And inside the Text Threads feature is the storage of a bucket load of user generated textual data: 
- This USER GENERATED TEXTUAL DATA is otherwise inaccessible - as it is stored inside a database file and serialized with custom data scheme using rust's Serde library.
- Removing this feature has either deleted / made-practically-inaccessible untold quantities of USER DATA.
No prompting - no escape button - no option given to export that data.

Seems pretty wild to me.

benbrandt:
Hey all, your conversations are still available. 
Depending on when you did your first text thread, it will either be in 

`~/.config/zed/conversations`

Or 
`~/.local/state/zed/conversations`

(Windows will be in  `C:\Users\Alice\AppData\Local\` or `C:\Users\Alice\AppData\Roaming\`)

These are just json files that should be parseable.

We know it isn't convenient, but we did not delete the files that powered the conversations if you really need to grab something from there.

benbrandt:
@bsutton the slash commands from extensions were only ever used in text threads, that is correct.

bsutton:
@benbrandt so two features deleted in one go...

Is there any intent to resurface the feature?

benbrandt:
Well these slash commands were only available in text threads anyway so they were tied together. 

This use case will likely be handled via skills in the future for the main agent panel

bennetbo:
@bsutton If you want to add custom slash commands for "normal" threads you can provide an MCP server that exposes [Prompts](https://modelcontextprotocol.io/specification/2025-11-25/server/prompts)

thezetrax:
I loved the text thread feature, that was the main reason I used zed.dev from the beginning

17twenty:
Thanks for removing - it was confusing to have the legacy behaviour as the agentic stuff rolled out 🎉

ksdev-pl:
Text threads were the main reason I used Zed.

ak64th:
Today I opened zed and got upgrade notification then lost the text threads. I can't find either ~/.config/zed/conversations or ~/.local/state/zed/conversations. Well I know it’s not a good idea to rely on chat history for saving data, but losing data is still frustrating.

benbrandt:
Are you on windows? then it is in the AppState directory I believe

bardsrealms:
Text threads were the main reason for me to use Zed. I can't see why this change was necessary; you destroyed my whole workflow in an instant by also not providing any alternatives.

I am speechless.

ak64th:
> Are you on windows? then it is in the AppState directory I believe

On Linux. Ubuntu 22.04 actually. I finally found the data at .local/share/zed/conversations

taoeffect:
Hey folks, apropos of nothing in particular, I've started work on a new app.

Introducing **Dez** 😎 

https://github.com/user-attachments/assets/99a4b112-49e5-44eb-9579-00db6623d2fb

Grab it here: https://github.com/taoeffect/dez

Vanuan:
It looks like Zed has reached the enshittification stage.

PR #52762, opened 2026-03-30
markdown: Skip linkify inside code blocks

smitbarmase:
Profiling streamed agent responses showed `parse_markdown_with_options` consuming ~50% of CPU cycles. The hot path is `LinkFinder` scanning every `Text` event, including code block content, where detected URLs are never rendered as links.

In the profiled case, the markdown had 235k+ URLs inside code blocks, making linkify extremely costly per-parse. This PR short-circuits the text-merging + linkify path for code block content.

Release Notes:

- N/A


ConradIrwin (approved):
Thanks!

I have a vague memory that linkify used to re-parse the entire block when new content is appended? is that still true; if so in addition to this we might need to keep track of a "scanned until" location so that even a large number of URLs in a text block cause no problems.

PR #52763, opened 2026-03-30
Fix agent servers loading environment from home dir instead of project dir

derekparker:
All local agent server types were calling `local_directory_environment`
with `paths::home_dir()`, causing direnv and shell environment to be
loaded from `~` rather than the project's worktree directory. This meant
project-specific `.envrc` variables (e.g. Google Vertex credentials)
were never picked up by external agents like Claude.

Added `default_environment()` on `ProjectEnvironment` that resolves the
default visible worktree path and uses it for environment loading,
falling back to home_dir() only when no worktree is available.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Fixed default environment variable context for external agents


crates/project/src/environment.rs
```diff
@@ -194,6 +194,39 @@ impl ProjectEnvironment {
...
+        cx: &mut App,
+    ) -> Shared<Task<Option<HashMap<String, String>>>> {
+        let abs_path = self.default_worktree_path(cx);
+        self.local_directory_environment(&Shell::System, abs_path, cx)
+    }
+
+    /// Resolves the path used for default environment loading.
+    /// Prefers visible directory worktrees over single-file worktrees,
+    /// falling back to the home directory if no worktrees are present.
+    pub fn default_worktree_path(&self, cx: &App) -> Arc<Path> {
```
benbrandt: I think we should use `Project::default_visible_worktree_paths` to keep the logic the same

benbrandt (changes requested):

crates/project/src/environment.rs
derekparker: Done.

benbrandt (approved):

PR #52765, opened 2026-03-30
Return image content from MCP tool response

eliaslundgren:
Currently Zed Agent MCP tool calls only support text responses, this PR adds support for images as well. When using Zed Agent with a locally ran llm like Qwen3.5 that supports images, this makes tools like grabbing screenshots possible. 

Main change is simply doing an early return in `crates/agent/src/tools/context_server_registry.rs` when parsing content in case an image is encountered. I recon this solution isn't perfect, in case the tool call returns both a string and an image, or multiple images for that matter, but it does feel like a step up from the current behavior of just ignoring it.

Self-Review Checklist:

Release Notes:

- Parse image in MCP Tool Response


benbrandt:
Closing in favor of #49615

PR #52767, opened 2026-03-30
Remove deprecated 1M context beta header for Sonnet 4.5

rtfeldman:
The `CONTEXT_1M_BETA_HEADER` (`context-1m-2025-08-07`) is deprecated for Sonnet 4 and 4.5. This removes the constant from the anthropic crate and the match arm in `beta_headers()` that sent it for `ClaudeSonnet4_5_1mContext`.

Note: The bedrock crate still has its own copy of this constant, used when the user-configurable `allow_extended_context` setting is enabled. That may warrant a separate cleanup.

Closes AI-114

Release Notes:

- N/A

### DIFF
--- a/crates/anthropic/src/anthropic.rs
+++ b/crates/anthropic/src/anthropic.rs
@@ -83,13 +83,6 @@ pub enum Model {
         alias = "claude-sonnet-4-5-thinking-latest"
     )]
     ClaudeSonnet4_5,
-    #[serde(
-        rename = "claude-sonnet-4-5-1m-context",
-        alias = "claude-sonnet-4-5-1m-context-latest",
-        alias = "claude-sonnet-4-5-1m-context-thinking",
-        alias = "claude-sonnet-4-5-1m-context-thinking-latest"
-    )]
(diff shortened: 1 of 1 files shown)


mikayla-maki (approved):

crates/anthropic/src/anthropic.rs
```diff
@@ -341,9 +339,6 @@ impl Model {
         let mut headers = vec![];
 
         match self {
-            Self::ClaudeSonnet4_5_1mContext => {
```
maxdeviant, line 344: I think we need to remove the `ClaudeSonnet4_5_1mContext` variant entirely, as with the end of the beta the 1M context window is only supported for Claude 4.6.

mikayla-maki: Oh, I misread the instructions. Thanks for the catch.

rtfeldman: Oh I missed that too! I'll push a fix.

benbrandt (approved):

PR #52769, opened 2026-03-30
elixir: Disable Emmet by default

AlternateRT:
The Emmet extension is enabled for Elixir files (`.ex`) because of the `~H` sigil, but this is a bit unexpected if you work exclusively with HEEx files (`.heex`) or you don't actually use Phoenix in any capacity

With this change, the user must explicitly opt into Emmet for Elixir files, just like Tailwind

Release Notes:

- N/A


MrSubidubi (approved):
Thank you!

PR #52773, opened 2026-03-30
theme_selector: Revert theme when modal is dismissed by clicking outside

Dnreikronos:
Context

Related to #52118

Found this while looking into the theme selector filtering bug: clicking outside the theme selector (or icon theme selector) keeps the previewed theme applied instead of reverting. `hide_modal()` doesn't go through the Picker's `dismissed()` callback, so the in-memory theme override sticks around without being written to `settings.json`.

This PR only fixes the click-outside behavior. The filtering issue from #52118 is separate.

## Demo

**Before (bug):**

https://github.com/user-attachments/assets/f8be49c4-7e1d-483c-bf23-346022ac83aa

**After (fix):**

https://github.com/user-attachments/assets/29cc1740-c869-42d1-8aee-fe46a091a949

## How to review

Two files, symmetrical changes. Look at:
- `on_before_dismiss` in `theme_selector.rs` and `icon_theme_selector.rs`
- The extracted `revert_theme()` method on both delegates, now shared between `dismissed()` and `on_before_dismiss`

## Self-Review Checklist

Release Notes:

- Fixed theme selector keeping the previewed theme when clicking outside instead of reverting to the original.

### DIFF
--- a/crates/theme_selector/src/icon_theme_selector.rs
+++ b/crates/theme_selector/src/icon_theme_selector.rs
@@ -26,7 +26,18 @@ impl Focusable for IconThemeSelector {
     }
 }
 
-impl ModalView for IconThemeSelector {}
+impl ModalView for IconThemeSelector {
+    fn on_before_dismiss(
+        &mut self,
+        _window: &mut Window,
+        cx: &mut Context<Self>,
+    ) -> workspace::DismissDecision {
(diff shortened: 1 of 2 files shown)


dinocosta (commented):
Thank you @Dnreikronos ! 🎉

dinocosta (approved):

PR #52776, opened 2026-03-30
ui: Adjust tree view item alignment

danilo-leal:
Adjusting the alignment of things in the tree view item given it got affected after I took out the square shape from the disclosure button in https://github.com/zed-industries/zed/pull/52322. Now, the indent line and the labels are back to being fully aligned.

Release Notes:

- N/A


ConradIrwin:
@zed-zippy approved

PR #52781, opened 2026-03-30
Automatically switch to unified diffs when diff view is narrower than a configurable "minimum split diff width"

maxbrunsfeld:
Release Notes:

- The git diff diff view now automatically switches from split mode to unified mode when the pane is narrower than a configurable minimum column count. You can configure this via the new `minimum_split_diff_width` setting.


cole-miller (commented):
Looks good!

PR #52792, opened 2026-03-31
Remove Agent V2 feature flag

mikayla-maki:
It's happening.

Self-Review Checklist:

Release Notes:

- N/A


danilo-leal (approved):

RomanPro100:
All panels are now on the opposite sides. Is that expected?
image

Also the doc comments are not up to date. The new default is `right`, but the tooltip still says it's `left`. And if I remove that setting, the panel IS on the right.
image

PR #52793, opened 2026-03-31
Implement work dir updating for active and background threads

mikayla-maki:
This PR changes things so that when the user mutates their current project (e.g. adding a folder), the thread entries now update the currently running threads to be pointing at this new set of folders.

Self-Review Checklist:

TODO:
- [x] Manually test this case

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -1996,17 +1996,12 @@ impl AgentPanel {
     fn update_thread_work_dirs(&self, cx: &mut Context<Self>) {
         let new_work_dirs = self.project.read(cx).default_path_list(cx);
 
-        let conversation_views: Vec<Entity<ConversationView>> = self
-            .active_conversation_view()
-            .cloned()
-            .into_iter()
-            .chain(self.background_threads.values().cloned())
-            .collect();
-
-        // Only update root threads. Subagent threads don't own their
-        // work_dirs
+        // Only update the active thread and still-running background threads.
+        // Idle background threads have finished their work against the old
+        // worktree set and shouldn't have their metadata rewritten.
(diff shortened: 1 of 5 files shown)


crates/agent_ui/src/agent_panel.rs
```diff
@@ -1981,6 +1993,68 @@ impl AgentPanel {
...
+        // Only update the active thread and still-running background threads.
+        // Idle background threads have finished their work against the old
+        // worktree set and shouldn't have their metadata rewritten.
+        let mut root_threads: Vec<Entity<AcpThread>> = Vec::new();
+
+        if let Some(conversation_view) = self.active_conversation_view() {
+            if let Some(connected) = conversation_view.read(cx).as_connected() {
+                for thread_view in connected.threads.values() {
+                    let thread = &thread_view.read(cx).thread;
+                    if thread.read(cx).parent_session_id().is_none() {
```
benbrandt, line 2008: Hmm why wouldn't we also propagate down to subagents? 
we try to keep values in sync wherever possible

crates/agent_ui/src/agent_panel.rs
```diff
@@ -1981,6 +1993,68 @@ impl AgentPanel {
...
+            let Some(connected) = conversation_view.read(cx).as_connected() else {
+                continue;
+            };
+            for thread_view in connected.threads.values() {
+                let thread = &thread_view.read(cx).thread;
+                let thread_ref = thread.read(cx);
+                if thread_ref.parent_session_id().is_some() {
+                    continue;
+                }
+                if thread_ref.status() != acp_thread::ThreadStatus::Generating {
```
benbrandt, line 2025: Why do we also skip non-running threads? If I add a worktree between prompts, wouldn't I expect it to be updated?

PR #52803, opened 2026-03-31
Fix resizing bottom-docked terminal panel

maxbrunsfeld:
The terminal panel is now flexible by default, but only when in a left or right panel. This PR fixes a bug where resizing the terminal panel was incorrectly setting the flex value instead of the fixed size when the terminal panel was in the bottom dock and flexible.

Release Notes:

- N/A


mikayla-maki (approved):
Nice DRYing

PR #52806, opened 2026-03-31
Retain draft threads in the sidebar

mikayla-maki:
This PR adjusts our thread serialization to store threads with nothing but draft prompt text, allowing us to deserialize them and not lose any user text.

Self-Review Checklist:

TODO:
  - [ ] Don't archive retained drafts

Release Notes:

- N/A

### DIFF
--- a/crates/agent/src/native_agent_server.rs
+++ b/crates/agent/src/native_agent_server.rs
@@ -1,4 +1,5 @@
 use std::{any::Any, rc::Rc, sync::Arc};
+use util::ResultExt as _;
 
 use agent_client_protocol as acp;
 use agent_servers::{AgentServer, AgentServerDelegate};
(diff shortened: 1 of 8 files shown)


crates/agent/src/native_agent_server.rs
```diff
@@ -45,17 +46,13 @@ impl AgentServer for NativeAgentServer {
         let thread_store = self.thread_store.clone();
         let prompt_store = PromptStore::global(cx);
         cx.spawn(async move |cx| {
-            log::debug!("Creating templates for native agent");
             let templates = Templates::new();
-            let prompt_store = prompt_store.await?;
+            let prompt_store = prompt_store.await.log_err();
```
mikayla-maki, line 50: This await errors in tests, causing obscure bugs revealed in the property test. Silently hiding it is bad but not being able to use the test is worse.

PR #52808, opened 2026-03-31
markdown: Render `details` and `summary` blocks 

Vastargazing:
## Summary
Fix `<details>`/`<summary>` tags not rendering in markdown preview

Closes #52387

## Problem
`pulldown_cmark` passes `<details>`/`<summary>` as a raw `HtmlBlock`. The HTML parser had no handler for the `<details>` tag - `consume_children` flattened it into plain text.

Second issue: `<details>` placed directly inside `<ul>`/`<ol>` without a `<li>` wrapper was silently dropped by `extract_html_list`.

## Fix

**html_parser.rs**

- Added ParsedHtmlDetails struct (source_range, summary, children)
- Added Details variant to ParsedHtmlElement
- Added "details" branch in parse_html_node - parses `<summary>` separately from body children
- Added non-`<li>` children pass in `<ul>`/`<ol>` branch - surfaces nested `<details>`

**html_rendering.rs**

- Added render_html_details - renders summary text + indented children
- Added Details arm in render_html_element match

## Screenshots

Before:
image

After:
Screenshot from 2026-03-31 10-14-28

Screenshot from 2026-03-31 10-14-31

Screenshot from 2026-03-31 10-14-34

## Tests
```
cargo nextest run -p markdown
Summary [0.189s] 46 tests run: 46 passed, 0 skipped
```

**Release notes**:
- Fixed `<details>`/`<summary>` tags not rendering in markdown preview


crates/markdown/src/html/html_parser.rs
```diff
@@ -173,11 +186,15 @@ impl HtmlImage {
 #[derive(Debug)]
 struct ParseHtmlNodeContext {
     list_item_depth: u16,
+    details_counter: Cell<usize>,
```
mikayla-maki: We'd like to avoid using Cells directly in the codebase wherever possible, and I think we should remove this because....

crates/markdown/src/html/html_rendering.rs
```diff
@@ -119,6 +119,74 @@ impl MarkdownElement {
...
+        builder: &mut MarkdownElementBuilder,
+        markdown_end: usize,
+        cx: &mut App,
+    ) {
+        let details_id = details.id;
+        let is_open = self
+            .markdown
+            .read(cx)
+            .details_open_states
+            .get(&details_id)
```
mikayla-maki: .... Using IDs like this is buggy. We're essentially storing the index of the detail item here, but if the user edits the markdown and adds a list item before this one, the opened/closed item will change.

However, fixing this will require us to make a much larger change, essentially making markdown rendering stateful so that we can track diffs and do the right thing in certain cases. 

As such, I think the simplest way forward would be to remove the mouse interactivity entirely and just show these always opened.

mikayla-maki (changes requested):

mikayla-maki:
Please mark this PR as ready, once it's been changed :)

Vastargazing:
Done - dropped the Cell counter, id field, and all the click/toggle logic. 
`details` just renders always-open now, which makes more sense anyway since the index would silently shift on any edit. 

46/46 nextest passed

alvarorichard:
---
  Following up on the state-keying concern — the index-based approach does shift on edits, but the failure mode goes away if the disclosure identity is derived from a
  hash of the open-tag text only (not body, not source offset).

  Concretely:

  - anchor = hash(trim(open_tag_text)) — e.g. &lt;details&gt; vs &lt;details open&gt; vs &lt;details id="x"&gt; each hash independently.
  - Editing the body never changes the anchor.
  - Inserting a sibling &lt;details&gt; or list item above doesn't change the anchor of existing widgets either, since each open tag is hashed in isolation.
  - State lives in a FxHashMap<SharedString, bool> on the Markdown entity. Toggle calls cx.notify() only — no reparse, no diff-aware rendering machinery needed.

  The other piece is that CommonMark splits &lt;details&gt;…blank line…&lt;/details&gt; into separate HTML blocks, so pairing must happen on the event stream, not inside
  a single html5ever parse. I prototyped both halves (event-level pairing in parser.rs via synthetic DisclosureStart / SummaryStart / SummaryEnd / DisclosureEnd events,
  plus a GPUI summary header with click toggle) on a fork:

  https://github.com/alvarorichard/zed/commit/2352619f385199ff8218cd255bc19ad3042befd5

  Lib + lib-clippy clean. Unit tests cover pair / open attr / nesting / stray close / feature gate / EOF auto-close / anchor stability.

  Happy to leave #52808 to land always-open as planned and extract this into a small follow-up PR if the team wants toggle after — or to pair on folding it in here,
  whichever is preferred.

  ---

mikayla-maki:
Hi y'all

PR is looking better! Can we actually use a hybrid approach to ID disambiguation? Essentially, use the text as you have here, but introduce a disambiguating counter for items with identical contents. We'll still have ID collisions, but I think this shrinks it to people almost never seeing it. I believe we're doing something similar in "build_heading_slugs" rn, according to Claude.

Thanks for the work on this :)

SomeoneToIgnore:

PR #52819, opened 2026-03-31
agent_ui: Prevent race conditions inside thread metadata store

bennetbo:
Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/thread_metadata_store.rs
+++ b/crates/agent_ui/src/thread_metadata_store.rs
@@ -180,7 +180,7 @@ impl ThreadMetadata {
 pub struct ThreadMetadataStore {
     db: ThreadMetadataDb,
     threads: HashMap<acp::SessionId, ThreadMetadata>,
-    threads_by_paths: HashMap<PathList, Vec<ThreadMetadata>>,
+    threads_by_paths: HashMap<PathList, HashSet<acp::SessionId>>,
     reload_task: Option<Shared<Task<()>>>,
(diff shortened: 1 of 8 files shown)


crates/sidebar/src/sidebar_tests.rs
```diff
@@ -4585,7 +4552,23 @@ async fn test_thread_switcher_ordering(cx: &mut TestAppContext) {
         );
     });
 
-    // Re-open switcher: Thread C is now most-recently-accessed.
+    sidebar.update(cx, |sidebar, _cx| {
```
bennetbo, line 4555: Turns out this test was not actually exercising the last_accessed code path, because we were not confirming (opening) the threads. This test just "accidentally" worked, because it relied on a bug in thread metadata store, where we would not respect the `created_at` date specified above (year 2024), instead it would use Utc::now(), which meant that the threads were accidentally in the expected order @cameron1024

crates/agent_ui/src/thread_metadata_store.rs
```diff
@@ -345,13 +335,10 @@ impl ThreadMetadataStore {
         }
 
         if let Some(thread) = self.threads.get(session_id) {
```
bennetbo, line 337: This was the core issue previously, if you inserted a new thread entry and reloading was not finished, this would return `None`. We now always optimistically update our in memory state, meaning that we don't have to hit the DB after every save, which is a good thing.

benbrandt (approved):

PR #52841, opened 2026-03-31
git: Forbid main git worktree deletion from worktree picker

Anthony-Eid:
This operation would always fail in the UI because the git binary already disallowed it. This PR just makes the UI more inline with what's actually valid

Self-Review Checklist:

Release Notes:

- git: Forbid main git worktree deletion from git worktree picker

### DIFF
--- a/crates/git_ui/src/worktree_picker.rs
+++ b/crates/git_ui/src/worktree_picker.rs
@@ -470,8 +470,7 @@ impl WorktreeListDelegate {
         let Some(entry) = self.matches.get(idx).cloned() else {
             return;
         };
-        if !entry.can_delete(self.forbidden_deletion_path.as_ref())
-        {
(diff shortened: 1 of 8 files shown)


crates/git_ui/src/worktree_picker.rs
```diff
@@ -462,7 +470,10 @@ impl WorktreeListDelegate {
         let Some(entry) = self.matches.get(idx).cloned() else {
             return;
         };
-        if entry.is_new || self.forbidden_deletion_path.as_ref() == Some(&entry.worktree.path) {
+        if entry.is_new
+            || entry.worktree.is_main
+            || self.forbidden_deletion_path.as_ref() == Some(&entry.worktree.path)
```
eholk:
```suggestion
        if !entry.can_delete(self.forbidden_deletion_path.as_ref())
```

eholk (approved):

PR #52848, opened 2026-03-31
Disambiguate project names

rtfeldman:
Disambiguate project names in the sidebar (and project picker) so that we don't show e.g. `zed, zed` but rather `foo/zed, bar/zed` if the last path component is the same but they are different absolute paths.

Release Notes:

- N/A


mikayla-maki:
@zed-industries/approved

PR #52851, opened 2026-03-31
Update links to troubleshooting images

JosephTLyons:
Now that I've started migrating blog content into the `zeddotdev-images` bucket, I've added a bit of structure.

`/blog`
`/docs`

This PR just updates the links so we can stay organized going forward. Shouldn't be seeing much thrash like this after this PR.

Release Notes:

- N/A


miguelraz:
@zed-zippy approved

PR #52855, opened 2026-03-31
Avoid reloading a file when reversing its unsaved changes due to an auto-format

maxbrunsfeld:
This fixes an issue that was introduced in https://github.com/zed-industries/zed/pull/51037, where a buffer might be incorrectly reloaded right as it is being saved, because of a format-on-save bringing it back into a clean state.

Release Notes:

- N/A


SomeoneToIgnore:

image

PR #52856, opened 2026-03-31
Bump tree-sitter for fix to wasm loading of grammars w/ reserved words

maxbrunsfeld:
This PR bumps Tree-sitter for this crash fix https://github.com/tree-sitter/tree-sitter/pull/5475

Release Notes:

- Fixed a crash that could occasionally occur when parsing files using certain language extensions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -113,20 +113,11 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


ChristopherBiscardi (approved):
lgtm

ChristopherBiscardi:
couple changes to the api usage need to happen though. looks like maybe multiple crate versions

ChristopherBiscardi (changes requested):
looks like the wasmtime bump is going to be a decent amount of changes

maxbrunsfeld:
Ok, upgrading Wasmtime was not complicated, and I think that has other benefits.

ChristopherBiscardi (approved):
sweet, compiles and runs without issue on my end.

PR #52857, opened 2026-03-31
language_model: Refactor crate structure and dependencies

kubkon:
A couple of things that this PR wants to accomplish:
* remove dependency on `settings` crate from `language_model`
* refactor provider-specific code into submodules - to be honest, I would go one step further and put all provider-specific bits in `language_models` instead but I realise we have cloud logic in `language_model` which uses those too making it tricky
* move anthropic-specific telemetry into `language_models` crate - I think it makes more sense for it to be there

Anyhow, I would very appreciate if you could have a look @mikayla-maki and @maxdeviant and lemme know what you think, if you would tweak something, etc.

Release Notes:

- N/A

### DIFF
--- a/crates/agent/src/native_agent_server.rs
+++ b/crates/agent/src/native_agent_server.rs
@@ -112,7 +112,7 @@ mod tests {
                 prompt_store::init(cx);
                 let registry = language_model::LanguageModelRegistry::read_global(cx);
                 let auth = registry
-                    .provider(&language_model::ANTHROPIC_PROVIDER_ID)
+                    .provider(&language_model::provider::ANTHROPIC_PROVIDER_ID)
                     .unwrap()
                     .authenticate(cx);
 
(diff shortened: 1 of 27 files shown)


crates/agent/src/native_agent_server.rs
```diff
@@ -112,7 +112,7 @@ mod tests {
                 prompt_store::init(cx);
                 let registry = language_model::LanguageModelRegistry::read_global(cx);
                 let auth = registry
-                    .provider(&language_model::ANTHROPIC_PROVIDER_ID)
+                    .provider(&language_model::provider::ANTHROPIC_PROVIDER_ID)
```
maxdeviant: Is there a reason we need this extra level of module nesting outside of the `language_model` crate?

I can understand wanting to have the code broken up into provider-specific modules, but seems like we could `pub use provider::*` to have the external interface remain the same.

kubkon: Ah yeah, good point. Thanks!

kubkon: Done!

maxdeviant (approved):

PR #17614, opened 2024-09-09
docs: Split vim mode documentation into two pages, edit for clarity

NathanLovato:
Closes #17215

Release Notes:

- N/A

---

This PR builds upon the vim mode documentation page and aims bring the following improvements:

- Separate vim mode-specific configuration from introducing vim mode.
- Reformat some lists of provided commands and keymaps from code blocks to sub-sections containing tables.
- Flesh out the text a little bit to make it more explicit in some parts.
- Generally format notes and a couple of other things closer to some other docs pages.

Checking the diff doesn't give a good idea of the changes, so here are some before after images for quick examples of the kinds of changes brought by this PR.

**Introducing the key differences of Zed's vim mode**

Before

After

---

**Zed-specific vim key bindings**

Before

After

PR #52892, opened 2026-04-01
docs: Improve vim mode's command mnemonics section

sinder38:
The "Command menonics" section in Zed's vim mode documentation made it seem that the mentioned aliases were available, by default, which isn't the case. These changes update the documentation to make it clearer that this is meant to be set on the `command_aliases` setting and provides and example configuration that users can easily copy and paste into their settings file.

Self-Review Checklist:

Closes #52894

Release Notes:

- N/A


dinocosta (commented):
Hey @sinder38 👋 

Thank you so much for opening this Pull Request! Even though the changes look good to me, I'm checking with the original author whether it makes sense for us to have defaults, as those were never in place in Zed, seems they were accidentally added when the documentation for `Command Mnemonics` was added by another contributor.

At the same time I'll double-check whether we want the `WorkspaceSettingsContent::command_aliases` field to make the command optional, I suspect we'll simply want the user's settings to override the defaults, in case we end up having any, as users can view the defaults simply by using `zed: open default settings` .

Will keep you posted! 🙂

P.S.: Related Pull Requests:

* https://github.com/zed-industries/zed/pull/14826 – Adds `command_aliases`
* https://github.com/zed-industries/zed/pull/17614 – Adds documentation for "Command Mnemonics"

sinder38:
Nullability for mnemonics is a must if there are any defaults. 

If someone doesn't like particular mnemonic or all of them they would have to update it every time Zed's command list affects it's placement.

For example a person decides to remove `zlog`mnemonic and types out a command it would trigger if it didn't exist
```json
    "zlog": "zed::OpenLog",
```

But some time later zed team adds a new command "zed::logger toggle"

So, if there are no mnemonics typing `zlog` should trigger logger toggle, but it doesn't since an old mnemonic still exists and points to command for the older list of commands

dinocosta:
Thanks for elaborating!

> Nullability for mnemonics is a must if there are any defaults.

Just to clarify, the behavior I was proposing is that `command_aliases` in user settings is treated as a complete replacement for any defaults, not merged with them. Right now, when `command_aliases` is set, it’s merged with the default map (even though we don’t currently define any defaults), but we can instead have it override that map entirely.

In that model, something as simple as `command_aliases: {}` would clear all defaults, so we wouldn’t need nullability to “turn off” specific default mnemonics. Hope that helps! 

P.S.: I even wonder if this is not the preferred approach, as it completely avoids the situation where, if a new default is added, an user could start seeing different results because of it, which would then require the user to override it with a `null` value just to be able to disable it.

sinder38:
Yea, that makes sense. I can implement that If you want.

I went with the "merge and remove" instead of "override" approach because it felt more appropriate in case many mnemonics were added. For the same reason zed handles keybinds with merge, too many to do override

dinocosta:
@sinder38 Thank you for getting back on that! 🙂

I've chatted with the original author of the feature and it seems we want to continue not having any defaults, but I do think we can improve documentation on this, making it clear what setting this relates to and providing the existing examples as a simple copy paste alternative, like we already have for [`Any Bracket Functionality > Example Configuration`](https://zed.dev/docs/vim#example-configuration).

Do you want to take a stab at updating the documentation with this information? You can find it under `docs/src/vim.md` and should be able to build and view Zed's documentation using `mdbook serve docs` at the root of the repository. Let me know if you need help!

dinocosta (approved):
Thank you @sinder38 ! 🎉

PR #52893, opened 2026-04-01
keymaps: Add Ctrl+R open recent binding for macOS and Linux

davidalecrim1:
Closes #52879

## Summary

VS Code binds `Ctrl+R` to open recent workspaces/folders on all platforms (Windows, macOS, and Linux). Zed already had this binding in `default-windows.json`, but it was missing from `default-macos.json` and `default-linux.json`.

Since `BaseKeymap::VSCode` returns no supplemental keymap file and relies entirely on the platform default keymaps, users who selected VS Code keybindings on macOS or Linux would not get the expected `Ctrl+R` behavior — instead getting nothing, or having to use the non-VSCode binding (`Alt+Cmd+O` / `Alt+Ctrl+O`).

This adds the missing binding to both platform defaults, consistent with what Windows already had.

## Screenshot
The default keybinding in VS Code:
Screenshot 2026-04-01 at 07 38 09

Release Notes:

- Added `Ctrl+R` keybinding for opening recent projects on macOS and Linux, matching VS Code's default behavior on all platforms.

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -574,6 +574,7 @@
       // Change the default action on `menu::Confirm` by setting the parameter
       // "alt-ctrl-o": ["projects::OpenRecent", { "create_new_window": true }],
       "alt-ctrl-o": ["projects::OpenRecent", { "create_new_window": false }],
+      "ctrl-r": ["projects::OpenRecent", { "create_new_window": false }],
       "alt-shift-open": ["projects::OpenRemote", { "from_existing_connection": false, "create_new_window": false }],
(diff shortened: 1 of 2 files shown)


dinocosta:
Hey folks 👋 

Seems like these changes break the ability to use `ctrl-r` while in the terminal in order to activate history search in certain shells, for example, zsh and bash. I've opened a Pull Request that tightens the context for these new bindings, https://github.com/zed-industries/zed/pull/53301 , feel free to take a look ! 🙂

ConradIrwin:
@zed-zippy approved (with the follow up PR from Dino!)

PR #52900, opened 2026-04-01
Implement MCP OAuth client preregistration

tomhoule:
In the interactive MCP OAuth flow, the MCP client registers itself with the authorization in one of three ways:

- Client ID Metadata Document aka CIMD (recommended default). This is already implemented: https://zed.dev/oauth/client-metadata.json.
- Dynamic Client Registration (DCR). This is the traditional method. Also already implemented in Zed.
- Pre-registration: the client is registered out of band, typically in the IdP or SaaS provider's UI. You get a client id and maybe a client secret, that have to be provided by the MCP client when it wants to exchange an access token. This is what this pull request is about.

This PR has two main parts:

- Allow users to configure a client id and optional client secret for an MCP server in their configuration, under a new `oauth` key, and take it into account 
- Make the MCP server state and the configuration modal aware of the intermediate states (client secret missing) and error cases stemming from client pre-registration.

The client secret can be stored either in the system keychain or in plain text in the MCP server configuration. The UI tries to steer user towards the more secure option: the keychain.

Screenshot 2026-04-10 at 16 48 06

Screenshot 2026-04-10 at 16 47 07

Screenshot 2026-04-10 at 16 47 23

Self-Review Checklist:

Closes https://github.com/issues/assigned?issue=zed-industries%7Czed%7C52198

**Note for the reviewer: I know how busy the AI team is at the moment so please treat this as low priority, we don't have signal that this is a highly desired feature. It's a rather large PR, so I'm happy to pair review / walk through it.**

Release Notes:

- Added support for OAuth client pre-registration (client id, client secret) to the built-in MCP client.

### DIFF
--- a/crates/agent_ui/src/agent_configuration/configure_context_server_modal.rs
+++ b/crates/agent_ui/src/agent_configuration/configure_context_server_modal.rs
@@ -17,7 +17,7 @@ use project::{
         ContextServerStatus, ContextServerStore, ServerStatusChangedEvent,
         registry::ContextServerDescriptorRegistry,
     },
-    project_settings::{ContextServerSettings, ProjectSettings},
+    project_settings::{ContextServerSettings, OAuthClientSettings, ProjectSettings},
     worktree_store::WorktreeStore,
 };
 use serde::Deserialize;
@@ -43,7 +43,9 @@ enum ConfigurationTarget {
         id: ContextServerId,
         url: String,
         headers: HashMap<String, String>,
+        oauth: Option<OAuthClientSettings>,
     },
+
     Extension {
         id: ContextServerId,
         repository_url: Option<SharedString>,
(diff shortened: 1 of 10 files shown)


crates/agent_ui/src/agent_configuration/configure_context_server_modal.rs
```diff
@@ -274,24 +284,58 @@ fn context_server_http_input(
...
         r#"{{
   /// Configure an MCP server that you connect to over HTTP
   ///
   /// The name of your remote MCP server
   "{name}": {{
     /// The URL of the remote MCP server
-    "url": "{url}",
+    "url": "{url}",{oauth}
     "headers": {{
+
```
bennetbo: I don't think we want this newline here: 

Image

bennetbo (approved):
Looks good! We should probably remove that newline though

crates/agent_ui/src/agent_configuration/configure_context_server_modal.rs
tomhoule: Oh yes, good catch! Thanks for the review!

Issue #38252, opened 2025-09-16
Custom MCP Servers Timeout on Zed Startup


khollingworth:
### Summary

Custom MCP Servers Timeout on Zed Startup

### Description

I run the following MCP servers.

<img width="622" height="357" alt="Image" src="https://github.com/user-attachments/assets/f49fd3b6-f3b4-4858-b6a5-c82b386d0565" />

On startup map-server-ref, 
{
  /// The name of your MCP server
  "mcp-server-ref": {
    /// The command which runs the MCP server
    "command": "/usr/local/bin/npx",
    /// The arguments to pass to the MCP server
    "args": ["ref-tools-mcp@latest"],
    /// The environment variables to set
    "env": {"REF_API_KEY":"ref-116b4149ad3b1126df54"}
  }
}

and map-server-browser,
{
  /// The name of your MCP server
  "mcp-server-browser": {
    /// The command which runs the MCP server
    "command": "/usr/local/bin/npx",
    /// The arguments to pass to the MCP server
    "args": ["@browsermcp/mcp@latest"],
    /// The environment variables to set
    "env": {}
  }
}

fail to load as they time out.  They are both npx and the only npx in the list.  The others are installed as MCP Extensions.  If I switch the failed  mcp servers off and back on, they then load correctly without timeout. 

Steps to reproduce:
1. Add the MCP extensions as per the screenshot
2. Add the two npx mcp servers using the config above.
3. restart Zed and see they fail due to timeout
4. toggle the failed mcp servers off then on 

**Expected Behavior**:
All MCP servers load and don't timeout on start up
**Actual Behavior**:
The npx based mcp servers timeout and fail to load

### Zed Version and System Specs

Zed: v0.203.5 (Zed) 
OS: macOS 26.0.0
Memory: 64 GiB
Architecture: aarch64

PR #52908, opened 2026-04-01
Stagger MCP server initialization and retry on timeout

yassineMouttalib:
When multiple MCP context servers start simultaneously during project load, the GPUI executor gets overwhelmed by concurrent process spawns and JSON-RPC handshakes. This is especially visible when using NPX to launch servers, which adds its own startup overhead — often causing timeouts as described in #38252.

While configurable timeouts were added in #45378, the root cause is that spawning multiple servers concurrently starves the executor of IO resources. This PR addresses that directly with two changes:

### Staggered initialization
Wait 3 seconds after project load before starting any MCP servers, then start them sequentially with 500ms between each. This gives the executor time to settle and prevents IO starvation during the critical startup window.

### Retry on timeout
When a server fails to start with a timeout error, retry up to 3 times with a 2-second delay between attempts. Non-timeout errors (including OAuth `401` responses handled by `resolve_start_failure`) skip the retry loop and are handled immediately.

Release Notes:

- Improved MCP context server startup reliability by staggering server initialization and retrying on timeout

### DIFF
--- a/crates/project/src/context_server_store.rs
+++ b/crates/project/src/context_server_store.rs
@@ -640,26 +640,82 @@ impl ContextServerStore {
         ) {
             self.stop_server(&id, cx).log_err();
         }
+
         let task = cx.spawn({
             let id = server.id();
             let server = server.clone();
             let configuration = configuration.clone();
 
             async move |this, cx| {
-                let new_state = match server.clone().start(cx).await {
-                    Ok(_) => {
-                        debug_assert!(server.client().is_some());
-                        ContextServerState::Running {
-                            server,
-                            configuration,
+                const MAX_START_RETRIES: usize = 3;
+                const RETRY_DELAY: Duration = Duration::from_secs(2);
+
+                for attempt in 0..=MAX_START_RETRIES {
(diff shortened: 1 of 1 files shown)


smitbarmase:
Thanks for taking the time on this.

I don't think this is the right direction. The `context_server_timeout` setting from #45378 already covers global default and per-server overrides (up to 10 minutes), and an affected user on #38252 confirmed it resolved the issue for them. The real cause in the `npx` cases is just cold-cache install exceeding the 60s default, not executor starvation, I couldn't reproduce that on a typical machine.

The stagger and 3s startup delay don't shorten install time, and the startup delay fires on every settings or registry change, not just startup. I think the right lever is the timeout setting itself, the default is probably just too low for cold `npx`.

The retry-on-timeout idea has some independent value (cold attempt fails, warm cache succeeds), so that could be worth pulling out separately, which is likely why it's been working for you.

I'll close this one for now. Feel free to follow up or open a new PR with just the retry-on-timeout idea, I think that's worth surfacing to users. Thanks again.

PR #52910, opened 2026-04-01
Update futures to 0.3.32

dinocosta:
As part of the work that is being developed for the Project Panel's Undo & Redo system, in https://github.com/zed-industries/zed/tree/5039-create-redo , we're implementing an asynchronous task queue which simply receives a message with the operation/change that is meant to be carried out, in order to ensure these run in a sequential fashion.

While trying to use `futures_channel::mpsc::Receiver`, it was noted that `recv` method was not available so this Pull Request updates the `futures` crate to `0.3.32`, where it is available.

This version also deprecates `try_next` in favor of `try_recv` so this Pull Request updates existing callers of `try_next` to use `try_recv`, which was mostly updating the expected return type from `Result<Option<T>>` to `Result<T>`.

Self-Review Checklist:

Release Notes:

- N/A


benbrandt (approved):
I would like this for some ACP reasons :D looks good

benbrandt (approved):

PR #52948, opened 2026-04-02
vim: Preserve system clipboard when pasting over visual selection

### LINKED ISSUE #52352: Paste in Vim mode replaces system clipboard with overwritten selection
### Reproduction steps

1. Enable Vim mode
2. Cmd-C to copy something from the buffer
3. Select something to overwrite
4. Cmd-V to paste the copied text over the selection
5. Repeat steps 3 and 4 on a different selection

Current behavior: the second paste inserts the text that was overwritten in step 4
Expected behavior: the second paste inserts the same text that was copied in step 2

This used to be the case, at some point in the last few months it changed. Even if the default doesn't change because That's How Vim Does It, some guidance as to how to enable this functionality with a custom keymap would be greatly appreciated :)

Zed: v0.228.0+stable.203.8421009ef8a022df1196d54bb42fd94366ec0988 (Zed) 
OS: macOS 26.3.1 (a) (Build 25D771280a)
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

N/A, nothing logged in log

### Relevant Zed settings

```json
Vim mode enabled, no other relevant changes are made
```

</details>

### Relevant Keymap

```json
I haven't changed anything from the default keymap (and `zed: open keymap file` doesn't do anything)
```

</details>

_No response_

None

Dnreikronos:
Self-Review Checklist:

Closes #52352

Release Notes:

- Fixed system clipboard being overwritten when pasting over a visual selection with Cmd-V/Ctrl-V in vim mode

### DIFF
--- a/crates/vim/src/vim.rs
+++ b/crates/vim/src/vim.rs
@@ -956,7 +956,9 @@ impl Vim {
                     Mode::Replace => vim.paste_replace(window, cx),
                     Mode::Visual | Mode::VisualLine | Mode::VisualBlock => {
                         vim.selected_register.replace('+');
-                        vim.paste(&VimPaste::default(), window, cx);
+                        let mut action = VimPaste::default();
+                        action.preserve_clipboard = true;
+                        vim.paste(&action, window, cx);
                     }
                     _ => {
                         vim.update_editor(cx, |_, editor, cx| editor.paste(&Paste, window, cx));
--- a/crates/vim/src/normal/paste.rs
+++ b/crates/vim/src/normal/paste.rs
@@ -25,7 +25,7 @@ pub struct Paste {
     #[serde(default)]
     before: bool,
     #[serde(default)]
-    preserve_clipboard: bool,
+    pub(crate) preserve_clipboard: bool,
 }
 
 impl Vim {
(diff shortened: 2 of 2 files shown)


crates/vim/src/vim.rs
```diff
@@ -956,7 +956,9 @@ impl Vim {
                     Mode::Replace => vim.paste_replace(window, cx),
                     Mode::Visual | Mode::VisualLine | Mode::VisualBlock => {
                         vim.selected_register.replace('+');
-                        vim.paste(&VimPaste::default(), window, cx);
+                        let mut action = VimPaste::default();
+                        action.preserve_clipboard = true;
+                        vim.paste(&action, window, cx);
```
seanstrom, line 961: Does this change mean that Vim-Mode will no longer "swap" text when pasting over a visual-selection?

Dnreikronos: Yes!
Before the fix, copying text with CMD+C and pasting it over a word worked fine. But on the second paste, instead of getting the original copied text again, it would paste the word that was overwritten in the first paste.

seanstrom: Okay, thanks for clarifying!

I think I understand the desire to avoid replacing/adding clipboard entries when using `CMD+V` during a vim-visual selection.

But I'm concerned that changing the vim-put command (paste with `p` key) to not swap text would diverge from some standard vim-mode behaviour.

Does this change mimic the behaviour of another editor with a vim-mode extension?

Dnreikronos: All good mate!
This only touches the `editor::actions::Paste` handler (line 952-964), the Cmd-V/Ctrl-V path. The vim `p` key dispatches `vim::Paste` through the normal keybinding system and doesn't go through here at all. So `p` in visual mode still swaps the selection into the default register, same as before.

Mechanically: `preserve_clipboard` in `paste.rs:62` defaults to `false`. Only the Cmd-V path sets it to `true`, which skips `copy_selections_content`. The `p` path uses `Paste::default()` (preserve_clipboard = false), so the register swap still happens.

And yes, this matches what other editors do. vscodevim and IdeaVim both preserve the clipboard on Cmd-V/Ctrl-V over a visual selection while keeping `p` as the standard swap. Neovim's system clipboard paste doesn't swap registers either. People who want "paste without swap" on `p` typically remap to `"_dP`.

seanstrom: Nice! I ran it locally and can confirm that it works just as you say, thanks again

Dnreikronos: You're welcome, mate!
Any problem, please reach me out!!!!

ConradIrwin (approved):
Interesting. Hopefully there aren't too many people relying on this (and if they are we can ask them to use p instead).

PR #52957, opened 2026-04-02
diagnostics: Prefer activating diagnostic under cursor

nullstalgia:
Update the behavior of both `editor: go to diagnostic` and `editor: go
to previous diagnostic` in order to ensure that, if there's a diagnostic
under the user's cursor that isn't active, it is first activated, with a
subsequent call jumping to the next or previous diagnostic,
respectively.

These changes also update how diagnostic activation handles the
situation when the global diagnostic renderer is not registered, as we
used to not update the active diagnostic group in that situation.
However, we now rely on it to determine whether the user's cursor is
already in the active diagnostic, with some tests now failing, so we now
default to an empty set of blocks for the active diagnostic group when
no global renderer is registered.

Release Notes:

- Update both `editor: go to diagnostic` and `editor: go to previous diagnostic` to prefer activating the diagnostic under the cursor before jumping to the next or previous diagnostic, respectively


nullstalgia:
I'm not super familiar with the test setup for Zed, but I think I was able to make relevant test cases for the new toggle being on/off!

Gently pinging @dinocosta as the tests have to be run manually again and they're the assignee according to the codeowners bot, apologies if this is uncouth!

dinocosta (commented):
Hey @nullstalgia ! 

Thank you so much for suggesting these changes. I believe the proposed behavior is sensible enough that it doesn't warrant a setting, so I went ahead and removed the setting, updating the default behavior for both `editor: go to diagnostic` and `editor: go to previous diagnostic` such that, if there's a non-active diagnostic under the cursor, we'll first activate that one, otherwise we'll navigate away to either the next or previous diagnostic, accordingly .

Lastly, I've brought this branch up to date with `main` and refactored the approach a bit, splitting the navigation from the activation so we could get rid of the `Option<Direction>`. I'm just wrapping up some new tests and will merge afterwards 🙂

P.S.: Sorry for the late reply, was out of office for the whole month.

dinocosta:
Updated the PR's description as that will be used as the commit's body. Merging now 🚢

PR #52970, opened 2026-04-02
gpui: Throttle framerate to 30 for unfocused windows

Veykril:
This should reduce energy consumption when having agents running in background windows, as the spinner that is being animated won't refresh at the display framerate anymore. 

Release Notes:

- Reduced energy consumption for background windows by throttling their update speed

### DIFF
--- a/crates/gpui/src/window.rs
+++ b/crates/gpui/src/window.rs
@@ -1211,18 +1211,31 @@ impl Window {
                     .update(&mut cx, |_, _, cx| cx.thermal_state())
                     .log_err();
 
-                if thermal_state == Some(ThermalState::Serious)
-                    || thermal_state == Some(ThermalState::Critical)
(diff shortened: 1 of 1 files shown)


cameron1024 (approved):

Anthony-Eid:
I'm not sure this was the right solution for this problem. We should have each animate be behind an AnyView element so we only re render the animated element instead of it's parent (The full agent panel) each frame

Edit: I also feel like this is an API that should be configured by APP, because other applications that are gpui based may not want this behavior

Veykril:
> I'm not sure this was the right solution for this problem. We should have each animate be behind an AnyView element so we only re render the animated element instead of it's parent (The full agent panel) each frame

We do not have damage tracking, so any change to the frame will cause us to redraw the entire frame. You are correct that we do not recompute the full scene graph, but thats not what this change is about.

PR #52975, opened 2026-04-02
anthropic: Preserve custom model thinking mode after thinking-toggle refactor

enoch-robinson:
PR #51946 broke `Model::Custom` thinking behavior: `mode()`, `supports_thinking()`, and `supports_adaptive_thinking()` all inferred capabilities from hardcoded built-in model lists, so any `Custom` variant always fell back to `Default` regardless of its configured `mode` field.

### Fixes

- **`Model::mode()`** — `Custom` now short-circuits to `mode.clone()` before the built-in inference logic
- **`Model::supports_thinking()`** — `Custom` returns `true` when `mode` is `Thinking { .. }` or `AdaptiveThinking`
- **`Model::supports_adaptive_thinking()`** — `Custom` returns `true` when `mode` is `AdaptiveThinking`

Built-in model behavior is unchanged.

### Tests

Three regression tests added covering the three `Custom` mode cases: explicit `Thinking`, `AdaptiveThinking`, and `Default` (which must disable both flags).

Self-Review Checklist:

Release Notes:

- Fixed custom Anthropic models losing their configured thinking/adaptive-thinking mode after the thinking-toggle refactor (#51946)

### DIFF
--- a/crates/anthropic/src/anthropic.rs
+++ b/crates/anthropic/src/anthropic.rs
@@ -287,33 +287,43 @@ impl Model {
     }
 
     pub fn mode(&self) -> AnthropicModelMode {
-        if self.supports_adaptive_thinking() {
-            AnthropicModelMode::AdaptiveThinking
-        } else if self.supports_thinking() {
-            AnthropicModelMode::Thinking {
+        match self {
+            Self::Custom { mode, .. } => mode.clone(),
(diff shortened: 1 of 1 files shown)


osyvokon (approved):

osyvokon:
Thank you!

PR #52996, opened 2026-04-02
recent_projects: Fix worktree path resolving to bare repo in recent projects modal

saberoueslati:
## Context

When using a bare-repo-based git worktree layout (e.g. `foo/.bare` as the bare repository and `foo/my-feature` as a linked worktree), the "recent projects" modal was showing `foo/.bare` instead of `foo/my-feature`.

The root cause was in `original_repo_path_from_common_dir` — when resolving a linked worktree back to its "main" repo, the function reads the `commondir` file which, for bare repos, points to the bare directory itself (e.g. `foo/.bare`). Since that path doesn't end in `.git`, the old code fell back to returning it as-is. This bare repo path was then substituted into the workspace's stored paths, causing the modal to show the bare directory instead of the actual worktree the user had opened.

The fix makes `original_repo_path_from_common_dir` return `Option<PathBuf>`, returning `None` when `common_dir` doesn't end with `.git`. `resolve_git_worktree_to_main_repo` propagates this `None`, meaning the original worktree path is preserved in recent projects rather than being replaced with the bare repo path.

Closes #52931

Video of the manual test after the fix is below :

[Screencast from 2026-04-02 16-05-48.webm](https://github.com/user-attachments/assets/9659c7a7-c095-4c23-af59-17715f84ce3e)

## How to Review

- **`crates/git/src/repository.rs`** : `original_repo_path_from_common_dir` changed to return `Option<PathBuf>`. Returns `None` for bare repos (no `.git` suffix). The existing `original_repo_path` call site falls back to `work_directory` when `None` is returned, preserving its prior behaviour. Unit test expectations updated accordingly, with the bare-repo case now asserting `None`.

- **`crates/project/src/git_store.rs`** : `resolve_git_worktree_to_main_repo` now simply forwards the `Option<PathBuf>` returned by `original_repo_path_from_common_dir` directly, propagating `None` for bare repos so the caller keeps the original worktree path.

- **`crates/workspace/src/persistence.rs`** : New test `test_resolve_worktree_workspaces_bare_repo` exercises the exact scenario from the issue: a workspace entry pointing to a linked worktree whose `commondir` resolves to a bare repo. Asserts the path is left unchanged.

## Self-Review Checklist

Release Notes:

- Fixed recent projects modal showing `.bare` folder instead of the worktree directory for bare-repo-based git worktree setups


osyvokon (approved):

osyvokon:
Thanks!

PR #53008, opened 2026-04-02
editor: Fix diagnostic rendering when semantic tokens set to full

### LINKED ISSUE #50212: No red squiggly underlines when `semantic_tokens` is set to `full`
### Reproduction steps

1. Set `semantic_tokens` to `full`
2. Write some bad code

Image

The error is reported as a diagnostic, but the red squiggly underline doesn't appear.

Zed: v0.225.9+stable.181.cbc2b428b9003b31077f91f06f8912880b1dbcf4 (Zed) 
OS: macOS 15.7.4
Memory: 16 GiB
Architecture: x86_64

lingyaochu:
Self-Review Checklist:

Closes #50212

There are two unreasonable coupling account for this issue, the coupling of `use_tree_sitter` with `languge_aware` in 
https://github.com/zed-industries/zed/blob/7892b932795911516f26f3c1c1c72249ed181ba8/crates/editor/src/element.rs#L3820-L3822
and the coupling of `language_aware` with `diagnostics` in https://github.com/zed-industries/zed/blob/7892b932795911516f26f3c1c1c72249ed181ba8/crates/language/src/buffer.rs#L3736-L3746

Because of these couplings, when the editor stops using Tree-sitter highlighting when `"semantic_tokens"` set to `"full"`, it also accidentally stops fetching diagnostic information. This is why error and warning underlines disappear.

I’ve fixed this by adding a separate `use_tree_sitter` parameter to `highlighted_chunks`. This way, we can keep `language_aware` true to get the diagnostic data we need, but still decide whether or not to apply Tree-sitter highlights. I chose to fix this at the `highlighted_chunks` level because I’m worried that changing the logic in the deeper layers of the DisplayMap or Buffer might have too many side effects that are hard to predict. This approach feels like a safer way to solve the problem.

Release Notes:

- Fixed a bug where diagnostic underlines disappeared when "semantic_tokens" set to "full"

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -98,7 +98,7 @@ use gpui::{
     WeakEntity,
 };
 use language::{
-    LanguageAware, Point, Subscription as BufferSubscription,
+    LanguageAwareStyling, Point, Subscription as BufferSubscription,
     language_settings::{AllLanguageSettings, LanguageSettings},
 };
 
@@ -1769,7 +1769,10 @@ impl DisplaySnapshot {
         self.block_snapshot
             .chunks(
                 BlockRow(display_row.0)..BlockRow(self.max_point().row().next_row().0),
-                false.into(),
+                LanguageAwareStyling {
+                    tree_sitter: false,
+                    diagnostics: false,
+                },
                 self.masked,
                 Highlights::default(),
             )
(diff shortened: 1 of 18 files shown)


SomeoneToIgnore (dismissed):
Great catch, thank you.

One question: why skip the most important part?

Image

SomeoneToIgnore (commented):
Oh, I see now, sorry: it's the rendering code that we're hitting here purely, so it's hard to test — would be amazing if we can get it tested somehow still, but not insisting now.

What's confusing though is this bit:

https://github.com/zed-industries/zed/blob/cb99ab4ac7ba4c8e60c437491ff2891e039fde26/crates/language/src/buffer.rs#L3736-L3746

it seems wasteful to get language highlights for the `full` case here — which we will do due to the new code, won't we?
Can we rework the code somehow to avoid getting the highlights but do get the diagnostics?

SomeoneToIgnore (approved):
Thank you a lot for the fix and the test!

PR #53029, opened 2026-04-02
Add persistence to the `Show Occupied Channels` collab toggle

JosephTLyons:
This PR adds persistence to the toggle state for the `Show Occupied Channels` filter.

Also, while driving by, I:
- renamed a few variables to use `occupied` over `active`, which should've happened in https://github.com/zed-industries/zed/pull/52531.
- extracted `"favorite_channels"` into a global

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Added persistence to the `Show Occupied Channels` collab toggle.

### DIFF
--- a/crates/collab_ui/src/collab_panel.rs
+++ b/crates/collab_ui/src/collab_panel.rs
@@ -483,8 +483,12 @@ impl CollabPanel {
                 .flatten()
                 .is_some();
 
-            panel.update(cx, |panel, _cx| {
+            panel.update(cx, |panel, cx| {
                 panel.filter_occupied_channels = filter_occupied_channels;
+
+                if filter_occupied_channels {
+                    panel.update_entries(false, cx);
+                }
             });
 
             let favorites: Vec<ChannelId> = KeyValueStore::global(cx)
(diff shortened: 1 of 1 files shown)


swannysec (approved):
A couple of near-nit things to consider (subject to your better historical context):

#### Suggestion: Call `update_entries` after restoring filter state on load

In `load()`, the PR sets `filter_occupied_channels` but doesn't call `update_entries()`. The entry list was already built by `CollabPanel::new()` with the default (`false`), and `update_entries` is the only path that applies the filter (line 790) and resets `list_state` (line 1030). The result: the toggle button appears "on" but the channel list shows all channels unfiltered until the next `channel_store` notification (typically a server-pushed `UpdateChannels` after the first `render_signed_in` triggers `SubscribeToChannels`).

```rust
panel.update(cx, |panel, cx| {
    panel.filter_occupied_channels = filter_occupied_channels;
    if filter_occupied_channels {
        panel.update_entries(false, cx);
    }
});
```

**Minor pre-existing finding:** The existing `pending_serialization` single-task pattern means the three writers (`serialize`, `persist_favorites`, `persist_filter_occupied_channels`) can theoretically cancel each other's in-flight writes. Pre-existing; not introduced here.

Neither is blocking from my perspective.

PR #53031, opened 2026-04-02
workspace: Skip preview tab paths when choosing default save location

### LINKED ISSUE #42787: Saving a new file defaults to the last file's directory, even if it's a preview tab in a venv
### Description

"Go to Definition" will often take you to a library's code in your venv. If you try to create a new file after that, it will default the save location inside the venv. This is almost never what you want, especially when that file opened in a preview tab.

Steps to reproduce:
1. Use "Go to Definition" for any library method
2. Open a new buffer and save it

**Expected Behavior**: The save location should default to something reasonable, e.g. the directory of my most recent file in a worktree, or the root of the most recently used worktree
**Actual Behavior**: The save location defaults to a directory inside my venv

Zed: v0.212.6 (Zed) 
OS: macOS 15.7.2
Memory: 16 GiB
Architecture: x86_64

_No response_

pdrgds:
Fixes #42787

## Summary

When "Go to Definition" navigates into a dependency (e.g., a file inside `.venv/` or `node_modules/`), the file opens as a preview tab. If the user then creates a new buffer and saves it, the save dialog defaulted to the dependency directory — almost never the intended location.

This PR makes `most_recent_active_path` skip paths belonging to current preview tabs, causing the save dialog to fall back to the next non-preview path, the worktree root, or the home directory.

## Design tradeoffs

We considered three approaches:

1. **Filter by `is_ignored`/`is_hidden`/`is_external` on worktree entries** — check if the path or any ancestor directory is gitignored, hidden (dotfile), or external (symlinked outside worktree). This catches `.venv` when it's in `.gitignore` (via `is_ignored`) or when it starts with `.` (via `is_hidden` matching the default `"**/.*"` pattern). However, it also false-positives on `.github/workflows/` — a directory users intentionally edit but that matches `is_hidden`. It would also miss `node_modules` when there's no `.gitignore`.

2. **Hardcode directory names** like `.venv`, `node_modules`, `site-packages` — fragile, opinionated, and would be a whack-a-mole game as ecosystems evolve.

3. **Use preview tab status** (this PR) — "Go to Definition" opens files as preview tabs by default. Skipping preview tab paths directly targets the user intent: the file was navigated to transitively, not opened deliberately. If the user double-clicks the tab (unpreviewing it), the path becomes eligible again — correctly reflecting a deliberate choice to work there.

The one gap in approach 3: if the user has disabled preview tabs entirely, "Go to Definition" opens files as regular tabs and they won't be filtered. We consider this acceptable since disabling preview tabs is an explicit opt-in to treating every opened file as permanent.

## Test plan

- [x] Manual test: open project with `.venv`, Go to Definition into a dependency, create new file, verify save dialog defaults to project root instead of `.venv/...` path
- [x] Added `test_most_recent_active_path_skips_preview_items` covering: normal active path returned, preview path skipped, unpreviewed path returned
- [x] All existing workspace tests pass (141/143 pass; 2 pre-existing flaky persistence tests)

Release Notes:

- Fixed save dialog defaulting to dependency directories (e.g. `.venv/`, `node_modules/`) after using Go to Definition.

### DIFF
--- a/crates/workspace/src/item.rs
+++ b/crates/workspace/src/item.rs
@@ -1470,9 +1470,18 @@ pub mod test {
 
     impl TestProjectItem {
         pub fn new(id: u64, path: &str, cx: &mut App) -> Entity<Self> {
+            Self::new_in_worktree(id, path, WorktreeId::from_usize(0), cx)
+        }
+
+        pub fn new_in_worktree(
+            id: u64,
+            path: &str,
+            worktree_id: WorktreeId,
+            cx: &mut App,
+        ) -> Entity<Self> {
             let entry_id = Some(ProjectEntryId::from_proto(id));
             let project_path = Some(ProjectPath {
-                worktree_id: WorktreeId::from_usize(0),
+                worktree_id,
                 path: rel_path(path).into(),
             });
             cx.new(|_| Self {
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -3731,8 +3731,17 @@ impl Workspace {
     }
 
     pub fn most_recent_active_path(&self, cx: &App) -> Option<PathBuf> {
+        let preview_paths: HashSet<ProjectPath> = self
+            .panes
+            .iter()
+            .filter_map(|pane| pane.read(cx).preview_item()?.project_path(cx))
+            .collect();
+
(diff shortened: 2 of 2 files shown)


pdrgds:
Closing in favor of a simpler approach using the `read_only_files` setting — see new PR.

PR #53033, opened 2026-04-03
Display agent-powered merge conflict resolution in the status bar

danilo-leal:
Follow up to https://github.com/zed-industries/zed/pull/49807

Previously, when there were multiple conflicts across the codebase, we would pop a toast at the bottom right corner of the UI. A toast seemed like a functional idea because it'd be visible from any state of the app and thus it'd be a good place to expose the button that allows you to quickly prompt the agent to resolve all conflicts, as opposed to creating a thread for each individual one. However, the toast was met with some negative (and correct) feedback, mostly because it is interruptive, and thus can  sometimes block very relevant surfaces, like either the agent panel itself or the Git commit area. 

Therefore, in this PR, I'm removing the toast and adding a button in the status bar instead; a bit more minimal, not interruptive, and a common place for other items that might require your attention. The status bar can be quite busy these days, though; we can display diagnostics, LSP status, and file names in there; conscious of that. But it felt like it could work given this button is such a transient one that you can either easily manually dismiss or wait for it to be auto-dismissed as you or the agent resolves the merge conflicts.

Screenshot 2026-04-02 at 9  15@2x

Release Notes:

- Git: Improved how we surface the affordance to resolve codebase-wide merge conflicts with the agent in the UI.
- Agent: Added a setting to control whether or not the button to resolve merge conflicts with the agent should be displayed.


crates/git_ui/src/conflict_view.rs
```diff
@@ -573,3 +505,162 @@ pub(crate) fn resolve_conflict(
...
+        self.dismissed = true;
+        cx.notify();
+    }
+}
+
+impl Render for MergeConflictIndicator {
+    fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
+        if self.conflicted_paths.is_empty() || self.dismissed {
+            return Empty.into_any_element();
+        }
```
Anthony-Eid: If disable ai is enabled we shouldn't show anything here

Anthony-Eid: I also wonder if we should have a setting to disable/enable this as well, that could be another PR tho

Anthony-Eid (approved):

crates/git_ui/src/conflict_view.rs
mikayla-maki: Agreed on a setting for this... feels a bit pushy to have it always show every time. If it's a 1 time upsell, it feels better.

PR #53036, opened 2026-04-03
Save Settings text inputs on blur

ChristopherBiscardi:
Previously, when writing in a Settings input field and tabbing between text inputs, the value the user wrote would be lost. This was particularly an issue when searching for settings meant that a text value was at the *end* of the settings page, and a tab out of the input would result in cycling to the top and losing a view of the input.

Settings inputs that don't have confirm buttons should confirm when blurring focus, since a user has written something into the text box.

This is complicated by the fact that there is another piece of code here that fires events during render, which is an anti-pattern since it means that events relating to inputs will fire out-of-order (such as: on_blur:confirm -> render:set_state back to original value -> render value from Settings file). However this code's behavior is meant to prevent the override of a user's active field if the settings file is changed by another process. It is unclear to me that this is a good behavior, since a user can lose that data without realizing it, but it was intentionally implemented.

The test for this render logic currently uses the user's written data in the input field for comparison, which isn't actually indicating whether the initial value has changed. So in this PR we store the *original* initial value, and compare future initial values against it, updating the initial value if necessary.

## Previous Behavior

https://github.com/user-attachments/assets/26f69809-787b-497c-ace5-c959c57a68bd

## This PR

https://github.com/user-attachments/assets/cb55d4c9-3553-4337-9f02-a8fb4c81aea7

Self-Review Checklist:

Release Notes:

- Settings text inputs no longer lose user data when blurring focus

### DIFF
--- a/crates/settings_ui/src/components/input_field.rs
+++ b/crates/settings_ui/src/components/input_field.rs
@@ -124,7 +124,11 @@ impl RenderOnce for SettingsInputField {
                         editor.set_text(text, window, cx);
                     }
 
-                    if let Some(confirm) = confirm.take() {
+                    if let Some(confirm) = confirm.take()
+                        && !self.display_confirm_button
+                        && !self.display_clear_button
+                        && !self.clear_on_confirm
+                    {
                         cx.on_focus_out(
                             &editor_focus_handle,
                             window,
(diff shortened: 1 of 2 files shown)


Anthony-Eid (approved):
Great work!

PR #53044, opened 2026-04-03
Honor annotations.audience on MCP tool response content blocks

dsturnbull:
MCP content blocks in tool results can carry `annotations.audience` to indicate whether the content is intended for the user, the model, or both. This is part of the [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations) and the [ACP content spec](https://agentclientprotocol.com/protocol/content#param-annotations). Zed was ignoring this field entirely — every tool result went to both the UI and the model regardless.

This matters for MCP servers that return large display-only content (sixel image previews, verbose logs, binary-ish data). Without audience filtering, that content consumes context budget on every subsequent turn without providing value to the model.

### How it works

| `audience` value       | Displayed to user | Sent to model |
|------------------------|-------------------|---------------|
| `None` (default)       | ✓                 | ✓             |
| `["user"]`             | ✓                 | ✗             |
| `["assistant"]`        | ✗                 | ✓             |
| `["user","assistant"]` | ✓                 | ✓             |

When all content blocks in a tool result are user-only, the model receives a placeholder `"[output displayed to user]"` so it still gets a valid tool result.

### Changes

**`context_server/src/types.rs`**
- Added `annotations: Option<MessageAnnotations>` to each `ToolResponseContent` variant, matching the existing `MessageContent` pattern
- Added `audience()` and `is_user_only()` helper methods
- 16 unit tests covering audience logic, serde round-tripping, and MCP wire format deserialisation

**`agent/src/tools/context_server_registry.rs`**
- `ContextServerTool::run()` partitions response content blocks using `is_user_only()`
- User-only text blocks are pushed to the event stream as display content (rendered in the tool call card but excluded from model context)
- When all blocks are user-only, the placeholder is sent to the model

**`context_server/src/listener.rs`** and **`agent/src/tests/mod.rs`**
- Updated `ToolResponseContent::Text` construction sites with the new `annotations` field
- Two integration tests: `test_mcp_tool_audience_filtering` (mixed audience) and `test_mcp_tool_audience_all_user_only` (all user-only with placeholder)

### Testing

22 tests total, all passing:
- 16 unit tests in `context_server::types::tests`
- 6 integration tests in `agent::tests` (4 pre-existing MCP tests still pass, 2 new)

Release Notes:

- Improved MCP tool integration to honour `annotations.audience` on content blocks, allowing MCP servers to mark output as display-only so it doesn't consume model context

### DIFF
--- a/crates/agent/src/tests/mod.rs
+++ b/crates/agent/src/tests/mod.rs
@@ -54,6 +54,7 @@ use std::{
 use util::path;
 
 mod edit_file_thread_test;
+mod test_mcp_audience;
 mod test_tools;
 use test_tools::*;
 
@@ -1493,6 +1494,7 @@ async fn test_mcp_tools(cx: &mut TestAppContext) {
         .send(context_server::types::CallToolResponse {
             content: vec![context_server::types::ToolResponseContent::Text {
                 text: "test".into(),
+                annotations: None,
             }],
             is_error: None,
             meta: None,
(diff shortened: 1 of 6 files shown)


bennetbo:
Closing because of conflicts. I would not merge the PR in the current state, but I think we would accept the PR if it was more scoped (remove the test script, remove the separate test module and add a single test to tests.rs)

PR #53066, opened 2026-04-03
Fix stale clangd buffer bookkeeping

ghost:
Release Notes:

- Fixed stale clangd requests and noisy watcher errors after buffers are unregistered.


SomeoneToIgnore:
image

Sorry for closing as stale, seems that the CI had never passed since creation months ago hence no one looked at it.
I think having a set of instructions on how to reproduce this in real workflow would be helpful too — ideally in a form of an issue.

PR #53071, opened 2026-04-03
terminal_view: Don't try home_dir when working locally

pepyakin:
When opening a remote session, sometimes you get a message:

    /bin/bash: line 1: cd: /Users/pep: No such file or directory

which suggests that that the local home dir (/Users/pep) is used for the remote terminal session, where it should be something like /home/ubuntu.

Release Notes:

- Fixed remote terminals incorrectly trying to change to a local home directory.


Veykril (approved):
Thanks!

PR #53074, opened 2026-04-03
Fix freeze when binary content written to open empty file

deadcode-walker:
Opening an empty file then writing binary content to it on disk causes a permanent freeze (not a crash — requires force kill).

`reload_impl` loads raw bytes via `load_bytes` and decodes with `encoding_rs` but never checks if the content is binary. The existing binary check in `decode_file_text` only runs on fresh opens, not reloads. So binary content enters the buffer as lossy UTF-8.

Binary data has almost no newlines, producing a single enormous row. The wrap map's sync fast path in `flush_edits` checks row count (< 100 rows) but not line length, so it runs `wrap_line` on a multi-MB single line synchronously on the main thread via `smol::block_on`. Font shaping millions of non-ASCII replacement characters blocks the UI thread indefinitely.

Two fixes, both one-condition guards:
- Null-byte check in `reload_impl` before decoding, same heuristic (first 8000 bytes) used by `decode_file_text` on fresh opens. Binary content never enters the buffer.
- Column-length guard (`MAX_SYNC_WRAP_COLUMNS = 10_000`) on the wrap map sync fast path so absurdly long lines fall through to the async background path. Defense in depth for any single-line content that's too long to shape synchronously.

## Test plan
- [ ] Open an empty file in Zed, write binary content to it externally (e.g. `cp /bin/ls /path/to/open-file`) — should show "Binary files are not supported" instead of freezing
- [ ] Open a compressed file (zip, gz) that starts empty and gets filled — same behavior
- [ ] Normal text file reload still works (no regression)
- [ ] Very long single-line text files (>10k columns) don't freeze the editor

### DIFF
--- a/crates/editor/src/display_map/wrap_map.rs
+++ b/crates/editor/src/display_map/wrap_map.rs
@@ -20,9 +20,6 @@ pub type WrapPatch = text::Patch<WrapRow>;
 pub struct WrapRow(pub u32);
 
 const WRAP_YIELD_ROW_INTERVAL: usize = 100;
-// max column span for the sync wrap fast path; lines longer than this go async
-// to avoid blocking the main thread (e.g. binary files decoded as a single huge line)
-const MAX_SYNC_WRAP_COLUMNS: u32 = 10_000;
 
 impl_for_row_types! {
     WrapRow => RowDelta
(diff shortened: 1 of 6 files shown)


smitbarmase (approved):
Thanks. I made a few improvements.

- Instead of using a naive binary content check, I extracted this logic from the open file flow, and we now use it.
- I also removed the wrap map change, since I don't think we should be handling it in this PR. If Zed never reveals binary file content, we might just not need it. Feel free to open a new one for that if you think it's necessary.

One thing that could be improved on top is that, instead of silently ignoring binary content, we might as well change the view to what the open file path shows. But that can be an addition on top.

PR #53076, opened 2026-04-03
Fix LSP status tooltip showing node binary path instead of server script

### LINKED ISSUE #52822: LSP status tooltip shows the `node` binary path instead of the language server's script path.
### Reproduction steps

1. Start Zed and open a project that uses an LSP managed by Zed/npm (e.g., `basedpyright` or `pyright`).
2. Ensure the language server is downloaded and managed by Zed.
3. Click the LSP icon in the status bar to open the language servers list.
4. Hover over the "running" status for the active server.
5. The tooltip displays the path to the node executable instead of the path to the language server script itself.

Here is a screenshot:

Image

Current:
The tooltip displays the path to the node binary.

Expected:
The tooltip should display the actual path to the language server.

Zed: v0.230.0+preview.205.9437a84390a396d666f04b38db87d89bb07284c1 (Zed Preview) 
OS: Windows 10.0.26200
Memory: 47.8 GiB
Architecture: x86_64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Sudarsh1010:
## Summary

- The LSP status tooltip was displaying the `node` binary path (e.g.
  `~/.local/share/nvm/v24.14.0/bin/node`) instead of the actual language
  server script path for node-based servers
- For node-based language servers (pyright, basedpyright, json, yaml, css,
  tailwind, vtsls, eslint), `LanguageServerBinary.path` is the node
  executable while the actual server script lives in `arguments[0]`
- Fixed by using `arguments[0]` for the tooltip when it is a non-flag
  argument (a path), falling back to `binary.path` for standalone and
  user-installed servers

## Before

Tooltip showed: `/home/user/.local/share/nvm/v24.14.0/bin/node`

## After

Tooltip shows: `/home/user/.local/share/zed/node_modules/vscode-langservers-extracted/bin/vscode-json-language-server`

Closes #52822

Release Notes:

- Fixed LSP status tooltip displaying the node binary path instead of the language server script path for node-based servers

### DIFF
--- a/crates/language_tools/src/lsp_button.rs
+++ b/crates/language_tools/src/lsp_button.rs
@@ -231,7 +231,10 @@ impl LanguageServerState {
                             server_id,
                             (
                                 status.server_readable_version.clone(),
-                                status.binary.as_ref().map(|b| b.path.clone()),
+                                status
+                                    .binary
+                                    .as_ref()
+                                    .map(|b| b.display_path().to_path_buf()),
                                 status.process_id,
                             ),
                         )
--- a/crates/lsp/src/lsp.rs
+++ b/crates/lsp/src/lsp.rs
@@ -73,6 +73,16 @@ pub enum IoKind {
     StdErr,
 }
 
+/// Describes how a language server binary is launched, determining what path
+/// to display to the user.
+#[derive(Clone, Debug, Serialize)]
+pub enum ServerBinaryKind {
+    NodeRuntime(usize),
+    PythonRuntime(usize),
+    Standalone,
+    Extension(Option<usize>),
+}
+
 /// Represents a launchable language server. This can either be a standalone binary or the path
 /// to a runtime with arguments to instruct it to launch the actual language server file.
 #[derive(Clone, Serialize)]
@@ -80,6 +90,26 @@ pub struct LanguageServerBinary {
     pub path: PathBuf,
     pub arguments: Vec<OsString>,
     pub env: Option<HashMap<String, String>>,
+    pub kind: ServerBinaryKind,
+}
+
+impl ServerBinaryKind {
+    fn script_argument_index(&self) -> Option<usize> {
+        match self {
+            Self::NodeRuntime(index) | Self::PythonRuntime(index) => Some(*index),
+            Self::Extension(Some(index)) => Some(*index),
+            Self::Standalone | Self::Extension(None) => None,
+        }
+    }
+}
+
+impl LanguageServerBinary {
+    pub fn display_path(&self) -> &Path {
+        self.kind
+            .script_argument_index()
+            .and_then(|index| self.arguments.get(index).map(|arg| Path::new(arg)))
+            .unwrap_or(&self.path)
+    }
 }
 
 /// Configures the search (and installation) of language servers.
@@ -1770,6 +1800,7 @@ impl fmt::Debug for LanguageServerBinary {
         let mut debug = f.debug_struct("LanguageServerBinary");
         debug.field("path", &self.path);
         debug.field("arguments", &self.arguments);
+        debug.field("kind", &self.kind);
 
         if let Some(env) = &self.env {
             let redacted_env: BTreeMap<String, String> = env
@@ -2095,6 +2126,7 @@ mod tests {
                 path: "path/to/language-server".into(),
                 arguments: vec![],
                 env: None,
+                kind: ServerBinaryKind::Standalone,
             },
             "the-lsp".to_string(),
             Default::default(),
@@ -2231,6 +2263,7 @@ mod tests {
                 path: "path/to/language-server".into(),
                 arguments: vec![],
                 env: None,
+                kind: ServerBinaryKind::Standalone,
             },
             "test-lsp".to_string(),
             Default::default(),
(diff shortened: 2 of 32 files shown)


crates/language_tools/src/lsp_button.rs
```diff
@@ -231,7 +231,13 @@ impl LanguageServerState {
                             server_id,
                             (
                                 status.server_readable_version.clone(),
-                                status.binary.as_ref().map(|b| b.path.clone()),
+                                status.binary.as_ref().map(|b| {
+                                    b.arguments
+                                        .first()
+                                        .filter(|arg| !arg.to_string_lossy().starts_with('-'))
```
lingyaochu: for ty or ruff installed by zed, they use `ty server` command to start, the path shown is not right:

Image

lingyaochu (changes requested):

crates/language_tools/src/lsp_button.rs
Sudarsh1010: ah got it, I will fix it

crates/language_tools/src/lsp_button.rs
```diff
@@ -231,7 +231,13 @@ impl LanguageServerState {
...
+                                        .map(|arg| PathBuf::from(arg))
+                                        .unwrap_or_else(|| b.path.clone())
+                                }),
```
lingyaochu: Would it make sense to encapsulate these changes within the `LanguageServerBinary` struct? It seems more cohesive there, and it would be easier to add independent unit tests.

Sudarsh1010: yes, I think so, I'd get back with my approach first before implementation

MrSubidubi (changes requested):
Thanks for this!

Aside from this needing some tests, I very much appreciate the notes from @lingyaochu and think they are very valid points, we should look into making this testable methods where the behavior is well explained - we have ways to see whether this was invoked using the node runtime, so we should use those and test the display stuff we want to fix here.

Will put this in a draft state so it is more clear this still needs some work, please feel free to move to review again once you tackled this. Thanks!

Sudarsh1010:
@MrSubidubi @lingyaochu 
got your point, was caught up with some other stuffs, will get back to this pr tomorrow 🫡

crates/lsp/src/lsp.rs
```diff
@@ -80,6 +80,16 @@ pub struct LanguageServerBinary {
     pub path: PathBuf,
     pub arguments: Vec<OsString>,
     pub env: Option<HashMap<String, String>>,
+    /// The path to the language server script or binary that should be displayed to the user.
+    /// Set when `path` is a runtime (e.g. node) and the actual server artifact is in `arguments`.
+    /// When `None`, `path` is the server binary itself.
+    pub server_path: Option<PathBuf>,
```
lingyaochu: For better semantic readability, would it be better to make this field an enum? For example:
```rust
pub server_binary_type: ServerBinaryType

enum ServerBinaryType {
    Node(PathBuf),
    Binary
}
```
When reading the code elsewhere, it's hard to understand what `server_path: None` stands for.

Sudarsh1010: yes you are right, I will get on it

lingyaochu:
Thanks for the updates! No further comments from my side.

MrSubidubi (changes requested):
How would you expect this to work for extension-provided language servers? If I am not missing something here, this would with the current approach not work at all, right?

Also, we are allocating the same string multiple times for something rather niece, could we not instead have something like `ServerBinaryKind::KnownArgument(usize)` that instead let's us index into the argument list? That way, we could entirely get rid of the duplicate string, could we not?

MrSubidubi:
Forgot to say above, please do not consider the feedback above as a discouragement - appreciate you looking into this!

Sudarsh1010:
`ServerBinaryKind::KnownArgument(usize)` I didn't thought this, but you are right, this would be correct way to do it, point to server path index.

and about extension provided lsp, yep it will not work, I will go through the extension lsp flow, will ask for help here If I got stuck 🫡

MrSubidubi:
Sounds good, quick hint for you for the extension-ones: We expose a [`node_binary_path`](https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.node_binary_path.html) for those which extensions then use as a path. You may assume that if the path matches the node binary path, the actual path will be somewhere in the argument list. If not, it's probably the path. After all, will be best effort, but we should have an enum variant that indicates we are aiming for just that.

MrSubidubi (changes requested):
Okay, at this point, the surface area of this has gotton huge for a fix I'd deem nice to have, but I think we have to think and discuss the approach chosen here more to find a better balance between the fix impact and change blast radius. Sorry for asking for this just now and not earlier.

Can we perhaps try resolving this just before we show the tooltip? Currently, I see us having three cases:
- language server is at the actual path
- language server is a node script and somewhere in the argument list, most likely first position, but perhaps second
- language server is invoked using Python, otherwise same rules apply

I think we can check for invoked through node -> invoked through python -> most likely standalone binary and add checks for this as needed (first one using the binary path most likely). Then, we can perhaps also show something like `(node)` after the path. 

With that, we keep the outside impact minimal (after all, we have no benefit from adding this besides the small case here, at least from what I can tell) whilst still fixing the issue. Sorry to change direction again here, what do you think?

Sudarsh1010:
yeah agreed, blast radius got big.
I will go through your approach and get back 🫡

MrSubidubi:
Will mark as draft for the time being, feel free to take the time you need here!

SomeoneToIgnore:

image

PR #51433, opened 2026-03-12
livekit: Use our build of libwebrtc.a

kubkon:
Closes #51339

This should address issues with too new libstdc++.so on older/more conservative distros such as RHEL9.

Release Notes:

- Relaxed requirement for libstdc++.so available on Linux distros.

PR #53078, opened 2026-04-03
Removal of mold/wild scripts and mentions in docs

NewtonChutney:
Self-Review Checklist:

Release Notes:

- N/A


script/install-wild
```diff
@@ -4,18 +4,40 @@
...
     echo "Error: This script is intended for Linux systems only."
     exit 1
-elif [ -z "$WILD_VERSION" ]; then
-    echo "Usage: $0 [version]"
-    exit 1
-elif command -v wild >/dev/null 2>&1 && wild --version | grep -Fq "$WILD_VERSION" ; then
-    echo "Warning: existing wild $WILD_VERSION found at $(command -v wild). Skipping installation."
+fi
+
+WILD_REPO="${WILD_REPO:-https://github.com/davidlattimore/wild}"
```
lapla-cogito: The repository has been moved to https://github.com/wild-linker/wild.

SomeoneToIgnore (changes requested):
Thank you, yet I am not sure what is the point of all this:

* Zed project does not use either `mold` or `wild` at all nowadays due to linking bugs with webrtc, see https://github.com/zed-industries/zed/pull/51433

* As you can clearly see, it's quite a churn to maintain: links get moved, versions change, etc. etc. 
People who know what a linker is should be capable of looking up the information on how to install and configure one for the project.

-------------

Would you be down to removing all the scripts and mold/wild mentions in the docs instead?

SomeoneToIgnore (approved):
Thank you.

Looks like I was a bit wrong and Zed did use `mold` for the headless server builds.
Those are not the Zed editor itself and do not deal with webrtc hence could be fine to use.

Still, seems ok to remove and re-approach if we need it later, but cc @kubkon just in case for a 2nd opinion.

I will merge this after Wednesday's release, to get more time to spot potential regressions and avoid a slightest chance to break the release process tomorrow.

kubkon (approved):
Looks good, thanks! I removed mold as it was mislinking on x86_64 with CREL sections. Given that since then I've patched out webrtc to never emit compact relocations, perhaps it would be OK to bring back mold, but then seems like nobody complained about link times being longer etc since I removed it so I'd just kept it out and if anyone wants to use mold/wild they can just plug it in.

PR #53085, opened 2026-04-03
bedrock: Rework Bedrock extended context management

### LINKED ISSUE #49617: Bedrock: `allow_extended_context` does not update UI context window limit (still shows 200k)

5herlocked:
Self-Review Checklist:

Closes #49617
Related #49945

Release Notes:

- BREAKING: `allow_extended_context` is deprecated and removed, that behaviour is dynamically adopted by the provider.
- Improved Context and Cost management for models with Extended Context capabilities


5herlocked:
Note to the maintainers, I have a CP-able commit 5039aaa that reports tokens as a feature of the Zed threads.

Probably useful for the other providers with "extended context" capabilities.

5herlocked:
~~There's also a panic in there somewhere which I'm struggling to find.~~

Fixed

bennetbo:
Thank you for this. Removing the explicit opt-in makes sense. I'm wondering if this is still accurate: https://github.com/zed-industries/zed/blob/00fb56d5a3462dfd617b954e924bcb2d46d5b08f/crates/language_models/src/provider/bedrock.rs#L762-L771
Anthropic [dropped the cost multiplier](https://claude.com/blog/1m-context-ga), so my understanding is that we can just always use it. Or is that different on Bedrock?

5herlocked:
Depends on the model -- the older (Sonnet 4, Sonnet 4.5) have the price multiplier still though it's 2x rather than a 5x; which I can make a comment change to that.

https://aws.amazon.com/bedrock/pricing/

bennetbo:
I think the simplest solution would be to just enable it for 4.6/7 then? And not bother with the 1 mi context window for 4/4.5 (why would people use it anyway if it's older but more expensive). Seems like that would make the code much simpler too

5herlocked:
Good shout. I'll do that today.

5herlocked:
On a sidenote @bennetbo could you take a look at #53177 -- I need a maintainers opinion on what the better approach is.

bennetbo:
@5herlocked removed the 1M beta header in https://github.com/zed-industries/zed/pull/56195. Seems like the pricing was updated for all models now, so we can just always use 1M context window size for all Anthropic models.

PR #53088, opened 2026-04-03
Fix crash in WgpuAtlas when viewing a screen share

osyvokon:
When atlas tiles are rapidly allocated and freed (e.g. watching a shared screen in Collab), a texture can become unreferenced and be removed while GPU uploads for it are still pending. On the next frame, `flush_uploads` indexes into the now-empty texture slot and panics:

```
  thread 'main' panicked at crates/gpui_wgpu/src/wgpu_atlas.rs:231:40:
  texture must exist...

  #11 core::option::expect_failed
  #12 gpui_wgpu::wgpu_atlas::WgpuAtlas::before_frame
  #13 gpui_wgpu::wgpu_renderer::WgpuRenderer::draw
```

This change drains pending uploads for a texture when it becomes unreferenced in `remove`, and skips uploads for missing textures in `flush_uploads` as a safety net.

Self-Review Checklist:

Release Notes:

- Fixed occasional crashes when viewing a screen share

### DIFF
--- a/crates/gpui_wgpu/src/wgpu_atlas.rs
+++ b/crates/gpui_wgpu/src/wgpu_atlas.rs
@@ -115,6 +115,8 @@ impl PlatformAtlas for WgpuAtlas {
         if let Some(mut texture) = texture_slot.take() {
             texture.decrement_ref_count();
             if texture.is_unreferenced() {
+                lock.pending_uploads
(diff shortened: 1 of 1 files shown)


ConradIrwin (approved):
Sounds good to me, see also #51996

PR #53089, opened 2026-04-03
git: Add git auto fetch

andriygm:
## Summary

Adds an opt-in periodic background `git fetch --all`, controlled by two new settings under `git`:

- `auto_fetch` (bool, default `false`) — on/off toggle
- `auto_fetch_interval_secs` (u64, default `60`, clamped to a minimum of `15`) — fetch cadence

Both are exposed in the settings UI under **Version Control > Git Integration**, where the interval field is only shown while auto-fetch is enabled. Unrelated settings changes don't restart the timer (guarded by comparing the cached `enabled`/`interval_secs`).

Because this fetch runs unattended, it differs from a user-initiated fetch in a few ways:

- **Trusted repositories only.** Re-checked on every tick, since a worktree can be trusted or restricted at any point in a session.
- **Never prompts.** It uses a no-op `AskPassDelegate`, and a new `AskPassDelegate::is_interactive` lets the git layer also pass `credential.interactive=false` and `GIT_TERMINAL_PROMPT=0` so git can't raise a prompt of its own. Credentials that need no interaction (SSH agent, a helper answering from a keychain) still work; anything requiring input fails instead of blocking on a dialog nothing will answer.
- **No repacking, no orphaned processes.** `gc.auto=0` keeps a background fetch from triggering auto-maintenance on the user's behalf, and `kill_on_drop` ensures a cancelled fetch doesn't outlive its task while holding ref locks.
- **Runs off the repository's serial job queue.** A fetch is network-bound, so occupying the queue for its duration would stall status refreshes and index writes behind it once per interval. This is safe because auto-fetch mutates no `Repository` state — it only updates remote-tracking refs, and the results reach the UI via the resulting git directory rescan. Manual fetch stays on the queue, since it also writes snapshot state.
- **No activity indicator.** A spinner appearing every minute is noise; only user-initiated fetches report progress in the status bar.
- **At most one fetch per repository**, via a semaphore. Auto-fetch skips its tick when a fetch is already in flight, while a manual fetch waits so an explicit user action is never dropped. Overlapping fetches would otherwise contend for the same remote-tracking ref locks.

## Test plan

`test_auto_fetch` covers:

- [x] An untrusted repository is never auto-fetched, and fetching begins once its worktree is trusted
- [x] Timer fires on schedule, and again on the following interval
- [x] Unrelated settings change mid-interval doesn't reset the timer
- [x] Disabling `auto_fetch` stops all fetches
- [x] Re-enabling with a longer interval respects the new cadence

## Review
Self-Review Checklist:

Addresses #24611(https://github.com/zed-industries/zed/discussions/24611)

Release Notes:

- Added git auto fetch option


levrik:
Oh. I would love to see this merged. Immediately noticed this as a downgrade after coming from VS Code/Cursor.

luizfneves404:
This PR shows merge conflicts against `main` — I opened andriygm#1 against this branch (`feat/git-auto-fetch`) that resolves them (mostly reconciling with an unrelated refactor of `Repository::local`/`remote` in `git_store.rs`). `cargo check` passes clean on the affected crates after the merge. Feel free to pull it in if useful!

LuckyPotatoe:
Any ideas/plans on when will this get merged?

assets/settings/default.json
```diff
@@ -1554,6 +1554,11 @@
     //
     // Trailing slashes are ignored.
     "worktree_directory": "../worktrees",
+    // Whether to automatically fetch from all remotes in the background.
+    "auto_fetch": false,
+    // How often to automatically fetch from all remotes, in seconds.
+    // Only applies when auto_fetch is enabled. (min: 15)
+    "auto_fetch_interval_secs": 60,
```
cameron1024, line 1766: We document all settings in `docs/src/reference/all-settings.md` - could we add an entry for these new ones?

crates/askpass/src/askpass.rs
```diff
@@ -61,6 +61,12 @@ impl AskPassDelegate {
         }
     }
 
+    pub fn no_op(cx: &mut AsyncApp) -> Self {
```
cameron1024, line 99: Nice :)

BTW, I think we also need a similar mitigation for the `git` binary itself, since that can also create prompts. Probably something along the lines of clearing out `credential.helper` and setting `GIT_TERMINAL_PROMPT=0`

crates/git_ui/src/git_panel.rs
```diff
@@ -7863,6 +7863,152 @@ mod tests {
         assert_eq!(message, Some("Update tracked".to_string()));
     }
 
+    #[gpui::test]
```
cameron1024: Only one please ;-p

cameron1024 (changes requested):
Looking good, just a few small changes needed.

BTW, we should also make sure this doesn't run on untrusted repos, since it's a bit of a security risk otherwise. You can find the trust mechanism in `crates/project/src/trusted_worktrees.rs`

0xdea:
Can't wait for this to get merged, it would be a great improvement.

crates/askpass/src/askpass.rs
andriygm: https://github.com/zed-industries/zed/pull/53089/commits/86ac91b8ce3e39850f871d26da42739fef7c5dfe handles this, but leans on `ask_pass.is_interactive`, meaning that it is somewhat coupled to a couple levers that are scoped outside of askpass itself. in the git crate we can instead do something like:
```rust
pub enum GitCaller {
    /// A user initiated this and is waiting on it.
    User(AskPassDelegate),
    /// Background work, no user attached: never prompts, never does unrequested
    /// maintenance, and may be cancelled at any time.
    Background,
}
```
which I think would be a little cleaner separation, but would expand scope a bit, is this acceptable?

tahayvr:
Patiently (kinda) waiting for this feature!

PR #53091, opened 2026-04-03
terminal: Bind Alt+F4 to Close Window in the terminal keymap

SAKETH11111:
Closes #52774

## Summary
- Bind Windows `Alt+F4` to `workspace::CloseWindow` in the `Terminal` keymap context
- Add a regression test covering the built-in Windows terminal keymap entry

## Why
When the integrated terminal is focused, `Alt+F4` should close the window instead of falling through to terminal keystroke handling. Handling this in the Windows `Terminal` keymap keeps the fix aligned with the rest of the terminal shortcut overrides.

## Validation
- `cargo test -p settings windows_terminal_keymap_closes_window_on_alt_f4`

Release Notes:

- Fixed Alt+F4 on Windows so Zed closes even when the integrated terminal is focused.

### DIFF
--- a/assets/keymaps/default-windows.json
+++ b/assets/keymaps/default-windows.json
@@ -1197,6 +1197,7 @@
       "ctrl-delete": ["terminal::SendText", "\u001bd"],
       "ctrl-n": "workspace::NewTerminal",
       // Overrides for conflicting keybindings
+      "alt-f4": "workspace::CloseWindow",
       "ctrl-b": ["terminal::SendKeystroke", "ctrl-b"],
       "ctrl-c": ["terminal::SendKeystroke", "ctrl-c"],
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore:
Also passing by and noting that `crates/settings/src/keymap_file.rs` is somewhat odd, we do not have other tests like these and do not seem to need one.

I'll leave the decision about the binding to @\reflectronic .

kubkon:
> Also passing by and noting that `crates/settings/src/keymap_file.rs` is somewhat odd, we do not have other tests like these and do not seem to need one.
> 
> I'll leave the decision about the binding to @\reflectronic .

Agreed about the test - removed. The binding looks fine to me tho.

kubkon (approved):

PR #53098, opened 2026-04-03
editor: Fix split diff spacer calculation for non-row-aligned patch groups

timvermeulen:
Fixes a bug in the split diff spacer calculation when a patch group starts mid-row, sometimes causing extra spacers to be inserted.

`spacer_blocks` already explicitly handles the case where `first_point` isn't at the start of `edit_for_first_point.old`, but the `while let Some(source_point) = source_points.next()` loop that follows implicitly assumes that `source_point` is at the start of `current_range`, which in turn seems to be based on the assumption that `current_range` starts at the beginning of a row. As it turns out, `current_range` isn't guaranteed to start at the beginning of a row, which can sometimes lead to incorrect spacer blocks being inserted.

This addresses that by moving the existing `if edit_for_first_point.old.start < first_point` logic into the loop body as `if current_edit.old.start < current_boundary` in order to handle any non-row-aligned patch groups, not just the first one.

Here's an example of how this bug could manifest:

https://github.com/user-attachments/assets/1d3a5b4c-e4ad-4d87-804b-c4390d25f408

After:

https://github.com/user-attachments/assets/b15acc62-33fe-4154-82e5-5cdf1806ffa7

Self-Review Checklist:

Release Notes:

- Fixed incorrect spacer blocks sometimes appearing in the split diff view when editing the file.

### DIFF
--- a/crates/editor/src/display_map/block_map.rs
+++ b/crates/editor/src/display_map/block_map.rs
@@ -1368,48 +1368,47 @@ impl BlockMap {
 
             let mut delta = their_baseline.0 as i32 - our_baseline.0 as i32;
 
-            // If we started out in the middle of a hunk/group, work up to the end of that group to set up the main loop below.
-            if edit_for_first_point.old.start < first_point {
-                let mut current_boundary = first_point;
-                let current_range = edit_for_first_point.new;
-                while let Some(next_point) = source_points.peek().cloned() {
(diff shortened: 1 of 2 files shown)


cole-miller:
Thanks @timvermeulen!

MrSubidubi:
@zed-industries/approved

PR #53100, opened 2026-04-03
workspace: Skip read-only paths when choosing default save location

### LINKED ISSUE #42787: Saving a new file defaults to the last file's directory, even if it's a preview tab in a venv

pdrgds:
Fixes #42787

## Summary

- When "Go to Definition" navigates into a dependency (e.g. `.venv/`, `node_modules/`), the save dialog for new files defaulted to that directory
- `most_recent_active_path` now checks the `read_only_files` setting and skips matching paths, falling back to the next eligible path, the worktree root, or the home directory

## Design tradeoffs

We considered three approaches:

1. **Filter by `is_ignored`/`is_hidden`/`is_external` on worktree entries** — catches `.venv` when gitignored or when it's a dotfile, but also false-positives on directories like `.github/workflows/` that users intentionally edit.

2. **Use preview tab status** — "Go to Definition" opens files as preview tabs, so skipping preview paths targets the right intent. But it doesn't work when preview tabs are disabled, and the signal is transient (preview status changes as you interact with tabs).

3. **Use `read_only_files` setting** (this PR) — an explicit user declaration of "I never want to edit files here." If you can't edit them, you don't want to save new files next to them either. This is the clearest signal of intent and respects user configuration. The tradeoff is that `read_only_files` is empty by default, so users need to configure it. But the kind of user bothered by the save dialog defaulting to a dependency directory is the same kind of user who already configures `read_only_files` (see [#46827](https://github.com/zed-industries/zed/discussions/46827) for an example).

## Test plan

- [x] Manual test: configured `read_only_files: ["**/.venv/**"]`, opened project, Go to Definition into `.venv`, created new file — save dialog defaults to project root
- [x] Added `test_most_recent_active_path_skips_read_only_paths`
- [x] All existing workspace tests pass

Release Notes:

- Fixed save dialog defaulting to dependency directories (e.g. `.venv/`, `node_modules/`) after using Go to Definition, when those directories are configured as `read_only_files`.


Veykril (approved):
Thanks!

PR #53124, opened 2026-04-04
workspace: Keep restricted mode modal actions visible

SAKETH11111:
Closes #52586

## Summary
- cap the restricted project list height inside the security modal and make it scroll
- cap the modal body content height so the action buttons stay reachable on smaller screens
- add a regression test that reproduces the overflow scenario with many restricted projects in a constrained window

## Validation
- manually reproduced the overflow by opening 60 untrusted projects in a 720x620 window before the fix
- cargo test -p workspace test_security_modal_project_list_scrolls_when_many_projects_are_restricted
- cargo check -p workspace

Release Notes:

- Fixed restricted mode dialogs overflowing past the window when many unrecognized projects are open.


danilo-leal (approved):
Thank you for tackling this one! I pushed a couple of commits simplifying the solution; this as an arguably easy UI problem, not sure it warrants a whole test for it. Here's the final product:

https://github.com/user-attachments/assets/ea75a478-72e4-4817-b5c2-bbcb5b7702b3

PR #53125, opened 2026-04-04
Add browser settings for configuring which browser opens URLs

dsturnbull:
Add `browser` setting for configuring which browser opens URLs

Zed currently always opens URLs with the system default browser via `cx.open_url()`. This adds a `browser` workspace setting that lets users choose which browser to use, supporting three forms:

```json
// Use the system default (current behaviour, and the new default)
"browser": "system"

// Use a specific program
"browser": "Safari"

// Use a program with arguments
"browser": { "program": "firefox", "args": ["--private-window"] }
```

### Changes

**Core:**
- `crates/settings_content/src/workspace.rs` — defines `BrowserSettings` enum with `System`, `Program(String)`, and `WithArguments { program, args }` variants
- `crates/workspace/src/workspace.rs` — adds `open_url_with_browser_settings()` which reads the setting and dispatches via `cx.open_url()`, `open::with()`, or `std::process::Command` respectively
- `crates/workspace/src/workspace_settings.rs` — adds `browser: BrowserSettings` field
- `crates/workspace/Cargo.toml` — adds `open` crate dependency
- `assets/settings/default.json` — adds default `"browser": "system"` entry
- `crates/settings/src/vscode_import.rs` — adds `browser: None` to VS Code import

**Wiring:**
- `crates/agent_ui/` — replaces all `cx.open_url()` calls with `open_url_with_browser_settings()` across `thread_view.rs`, `configure_context_server_modal.rs`, `agent_registry_ui.rs`, `end_trial_upsell.rs`, and `mention_crease.rs`
- `crates/zed/src/zed.rs` — replaces `cx.open_url()` in the `OpenUrl` action handler

### Not yet wired

There are remaining `cx.open_url()` calls in `zed.rs` that were intentionally left alone:
- `OpenDocs` action and account URL — could be wired up in a follow-up
- Error-then-quit paths (inotify warning, Windows warning, software emulation) — the app exits immediately after, so the browser setting is less relevant
- Other crates (`editor`, `collab`, `extensions`, etc.) may also have `cx.open_url()` calls worth converting

### Testing

The `workspace` crate compiles cleanly. Full build verification of downstream crates (`agent_ui`, `zed`) is pending.

Release Notes:

- Added a `browser` setting to control which browser Zed uses to open URLs, with support for specifying a program name and custom arguments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21463,6 +21463,7 @@ dependencies = [
  "markdown",
  "menu",
  "node_runtime",
+ "open",
  "parking_lot",
  "postage",
  "pretty_assertions",
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -194,6 +194,15 @@
   //
   // Default: "client"
   "window_decorations": "client",
+  // The browser to use when opening URLs.
+  // Can be:
+  //   - "system" to use the system default browser
+  //   - A path to a browser executable (e.g., "/usr/bin/firefox")
+  //   - An object with "program" and optional "args" fields:
+  //     { "program": "firefox", "args": ["--private-window"] }
+  //
+  // Default: "system"
+  "browser": "system",
   // Whether to use the system provided dialogs for Open and Save As.
   // When set to false, Zed will use the built-in keyboard-first pickers.
   "use_system_path_prompts": true,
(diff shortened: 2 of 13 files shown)


SomeoneToIgnore:

image

PR #50360, opened 2026-02-28
acp_thread: Handle UsageUpdate in ACP session notifications

rohan-patra:
The `UsageUpdate` variant from external ACP agent servers (e.g. `claude-agent-acp`) was silently dropped by the catch-all `_ => {}` match arm in `handle_session_update()`. This adds a match arm that routes it to the existing `update_token_usage()` method.

This enables the context window progress ring and token limit warning banners that already exist in Zed's UI for ACP-based agent servers.

**What changed:**
- Added a match arm for `acp::SessionUpdate::UsageUpdate` in `AcpThread::handle_session_update()` that maps `usage.size` → `max_tokens` and `usage.used` → `used_tokens`, then calls `self.update_token_usage()`

**Why:**
- The `UsageUpdate` type is already defined in `agent-client-protocol` (behind `unstable_session_usage`, which Zed enables)
- The upstream `claude-agent-acp` server already sends `usage_update` notifications with `used`, `size`, and `cost`
- `update_token_usage()`, `TokenUsage`, the circular progress widget, and the warning banners all already exist — they just weren't wired up for ACP sessions

Before you mark this PR as ready for review, make sure that you have:
- [x] Done a self-review taking into account security and performance aspects

Release Notes:

- Fixed context window usage display (progress ring and warning banners) not appearing for external ACP agent servers

PR #53130, opened 2026-04-04
gpui: Prevent idle sleep during AI response streaming

cppcoffee:
Release Notes:

- Added button tu prevent idle sleep during AI response streaming

**Context / Motivation:**
When an AI agent takes a long time to generate a response, the operating system might enter an idle sleep (or display sleep) state if there is no user interaction (mouse or keyboard). This can interrupt the workflow. This PR aims to prevent the system from sleeping during an active AI turn by invoking native OS APIs, and provides a UI toggle for users to control this behavior.

**Key Changes:**

1. **Cross-Platform Idle Sleep Prevention in GPUI (`gpui`, `gpui_*`)**
   - Added a new `prevent_idle_sleep` method to the `Platform` trait. It returns a RAII-based `PreventIdleSleepToken`, which automatically restores the original sleep policy when dropped.
   - **macOS**: Utilizes `NSProcessInfo`'s `beginActivityWithOptions` with `NSActivityIdleDisplaySleepDisabled`.
   - **Windows**: Uses `SetThreadExecutionState` to prevent display and system sleep, incorporating a reference counter to handle concurrent requests safely.
   - **Linux**: Uses `ashpd` to request `org.freedesktop.portal.Inhibit` (inhibiting `Idle` and `Suspend` states) via D-Bus.

2. **AI Agent State Integration (`crates/acp_thread`)**
   - Introduced idle sleep prevention logic in `AcpThread`: Acquires the token at the beginning of `run_turn`, and automatically releases it when the turn completes normally, errors out, or is explicitly canceled by the user.
   - Added comprehensive unit tests to ensure the token is correctly acquired and released under various cancellation and concurrency scenarios.

3. **Settings and UI (`crates/agent_settings`, `crates/agent_ui`)**
   - Added `AgentOutputIdleSleepControl` to global state.
   - Added a new toggle button (screen icon) to the toolbar at the bottom of the `ThreadView` (AI conversation panel). Users can hover to view the tooltip and click to toggle the feature on or off.

截屏2026-04-04 16 09 08


hahahuy:
Hi, I'd like to take a stab at this.

After reading the code, the root cause is clear: `PromptResponse` in `agent-client-protocol` already has a `usage: Option<Usage>` field (gated on the `unstable_session_usage` feature, which Zed already enables via the `unstable` feature flag). But the `Ok(r)` branch in `run_turn` (acp_thread.rs) never reads it — so `token_usage` stays `None` and `render_token_usage()` in `thread_view.rs` returns early and renders nothing.

My planned fix:
- In `crates/acp_thread/src/acp_thread.rs`, extract `r.usage` from the `PromptResponse` at the end of each turn and call `update_token_usage()` with the data.
- `acp::Usage` has `total_tokens`, `input_tokens`, `output_tokens` but no `max_tokens` (the ACP protocol doesn't expose context window size). I'll set `max_tokens = 0` for now — `render_token_usage` already handles this gracefully (progress ratio defaults to 0.0, counts are still displayed in the tooltip).

The rendering side already works — it just never fires for external ACP agents today. This is a ~15-line change.

Before I open a draft PR: is there anything already in-flight here I should be aware of? Also, should I build on PR #50360 or start fresh from main?

maxbrunsfeld (dismissed):
This seems like a good change

We do already have some similar logic, for macOS only, here: https://github.com/zed-industries/zed/blob/590aaafcabf67a055b7a448bd87ebaf34711c968/crates/livekit_client/src/livekit_client/playback.rs#L880-L915.

In that case, we're telling the OS not to throttle the app because we are playing audio.

As part of this PR, could you replace that `PreventAppNapGuard` with your new cross-platform abstraction?

cppcoffee:
> This seems like a good change
> 
> We do already have some similar logic, for macOS only, here:
> 
> https://github.com/zed-industries/zed/blob/590aaafcabf67a055b7a448bd87ebaf34711c968/crates/livekit_client/src/livekit_client/playback.rs#L880-L915
> 
> .
> In that case, we're telling the OS not to throttle the app because we are playing audio.
> 
> As part of this PR, could you replace that `PreventAppNapGuard` with your new cross-platform abstraction?

rebase `main` branch. use `PreventAppNapGuard` done

SomeoneToIgnore (approved):
Thank you for the idea — I've polished it based on the feedback, seems good enough to merge after this Wednesday's release and test.

gianpaj:
Why am I getting this error if it's supposed to be a boolean?
image

```
Zed: v1.21.0+stable.362.33c95853ed2b6956f339733c63a8220964ecbeb6 (Zed) 
OS: macOS 27.0
Memory: 24 GiB
Architecture: aarch64
```
https://zed.dev/docs/reference/all-settings#prevent-idle-sleep

SomeoneToIgnore:
Seems like you use the wrong key: 
image

You can also try to flip it in the settings ui, as it's simpler than editing the json.

PR #53137, opened 2026-04-04
dev_container: Make volumes key and source key in volume mounts optional

yeetypete:
Fixes some issues concerning volume mounts in the `dev_container` integration:
1. Docker Compose services that don't define a volumes key cause deserialization to fail because the field was required. This field is not strictly necessary, i.e. for other services in a docker compose devcontainer configuration which the editor is not attached to.
1. Volume mounts where source or target is absent (e.g. `tmpfs` mounts that only need a target) also fail to parse because both fields were required. This makes the source key optional, matching the Docker Compose spec.

Self-Review Checklist:

Release Notes:

- Fixed devcontainer initialization erroneously requiring each service to have a volumes key.
- Fixed devcontainer initialization erroneously requiring source keys for all volume mounts.

### DIFF
--- a/crates/dev_container/src/devcontainer_json.rs
+++ b/crates/dev_container/src/devcontainer_json.rs
@@ -424,8 +424,8 @@ fn deserialize_mount_definition<'de, D>(
 where
     D: serde::Deserializer<'de>,
 {
-    use serde::de::Error;
     use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


KyleBarton (approved):

KyleBarton:
@zed-zippy approved

PR #53138, opened 2026-04-04
Add AVIF support across image handling

jackkkonggg:
## Summary

This PR adds AVIF support to the image handling paths that previously accepted the format at the type level but did not consistently enable or validate decoding end to end. That gap meant AVIF images could fail in places like project image loading, pasted image handling, or REPL-related image flows depending on which crate and platform path handled the bytes.

The root cause was that AVIF support was only partially wired through the stack. This change enables the native AVIF decoding feature for the affected Unix-like crates, adds AVIF coverage for GPUI MIME and macOS pasteboard type mappings, and adds focused tests around project image creation and metadata decoding so the supported behavior is exercised directly.

## Validation

- `cargo test -p project test_create_gpui_image_supports_avif --lib`
- `cargo test -p project test_compute_metadata_from_bytes_supports_avif --lib`
- `cargo check -p agent_ui -p repl`
- `cargo build --release --package zed`

image

Release Notes:

- Added AVIF image support across clipboard, project, and REPL image flows.


SomeoneToIgnore:
Sorry, have talked about adding an extra dependency in macOs builds and for now did not get too much excitement.
I think this can be revisited when avif adoption grows, but for now what seems to be a good start is an issue/discussion.

Issue #52928, opened 2026-04-01
Regression: Devcontainer - ARG variables in FROM directive not resolved during base image pull on 0.231.0 preview


albinkc:
### Reproduction steps

After the latest Zed Preview update, devcontainers with ARG-parameterized FROM lines fail to start.
1. Create a `.devcontainer/Dockerfile` that uses `ARG` interpolation in `FROM`:
 
```dockerfile
ARG ELIXIR_VERSION=1.20.0-rc.4
ARG OTP_VERSION=28.4.1
ARG DEBIAN_VERSION=trixie-20260316-slim
ARG IMAGE="docker.io/hexpm/elixir:${ELIXIR_VERSION}-erlang-${OTP_VERSION}-debian-${DEBIAN_VERSION}"
 
FROM ${IMAGE} AS devcontainer
# ...
```
 
2. Create a `.devcontainer/devcontainer.json` referencing this Dockerfile:
 
```json
{
  "name": "Elixir Dev",
  "build": {
    "dockerfile": "Dockerfile"
  }
}
```
 
3. Open the project in Zed and start the devcontainer.

### Current vs. Expected behavior

Zed attempts to pull the literal string `${IMAGE}` without expanding the `ARG` values:
 
```
ERROR [dev_container::docker] Non-success result from docker pull: Error: parsing reference "${IMAGE}": invalid reference format
ERROR [dev_container::command_json] Error running command Command { program: "docker", args: ["inspect", "--format={{json . }}", "${IMAGE}"], ... }: Sent non-successful output; cannot deserialize. StdErr: Error: no such object: "${IMAGE}"
ERROR [recent_projects::remote_servers] Failed to start dev container: DevContainerUpFailed("Failed with nested error: CommandFailed(\"docker\")")
```

### Zed version and system specs

Zed: v0.231.0+preview.209.f14453809b501cf8cea5e9e76d2f9ab3515936aa (Zed Preview) 
OS: macOS 26.4
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
INFO  [dev_container::devcontainer_api] find_configs_in_snapshot: Found 1 configurations
INFO  [recent_projects::remote_servers] SSH: Watching User Config at: "/Users/albin/.ssh/config"
INFO  [recent_projects::remote_servers] SSH: Watching Global Config at: "/etc/ssh/ssh_config"
[...]

PR #53141, opened 2026-04-04
dev_container: Fix alias injection for multi-stage Dockerfiles

yeetypete:
In multi-stage Dockerfiles, the alias injection logic looked at the first FROM line to detect and inject stage aliases. For Dockerfiles with an aliased intermediate stage (e.g. `FROM ubuntu:24.04 AS builder`) followed by an unaliased final stage (e.g. `FROM ${BASE_IMAGE}`), the code would see the first stage's alias and skip injection entirely. This left the final stage without the `dev_container_auto_added_stage_label` alias that the feature injection Dockerfile depends on. The fix targets the last FROM line instead, which is always the stage that produces the output image.

Self-Review Checklist:

Release Notes:

- Fixed dev container builds failing when the Dockerfile uses multi-stage builds with aliased intermediate stages.

### DIFF
--- a/crates/dev_container/src/devcontainer_manifest.rs
+++ b/crates/dev_container/src/devcontainer_manifest.rs
@@ -2249,7 +2249,8 @@ chmod +x ./install.sh
 fn dockerfile_alias(dockerfile_content: &str) -> Option<String> {
     dockerfile_content
         .lines()
-        .find(|line| line.starts_with("FROM"))
+        .filter(|line| line.starts_with("FROM"))
+        .last()
         .and_then(|line| {
             let words: Vec<&str> = line.split(" ").collect();
             if words.len() > 2 && words[words.len() - 2].to_lowercase() == "as" {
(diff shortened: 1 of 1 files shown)


yeetypete:
@KyleBarton sure, happy to. Although after some more thinking I think this may break again if the [`build target`](https://docs.docker.com/reference/cli/docker/buildx/build/#target) is specified.

It's also exposed in the devcontainer specification: https://containers.dev/implementors/json_reference/#image-specific

Should we try to figure out a better way to handle this?

KyleBarton:
@yeetypete yes, I think a more holistic approach is worthwhile here. This looks tangentially related to https://github.com/zed-industries/zed/issues/52928 as well. I think we probably should retrofit `dockerfile_alias`, `dockerfile_inject_alias`, and `image_from_dockerfile` in `devcontainer_manifest.rs` all so that they take in the Dockerfile, the devcontainer.json args, and create a more semantic representation of the contents of the Dockerfile.

PR #53146, opened 2026-04-04
Fix duplicating a file in the project panel triggers rename, but only selects the copy suffix

OmChillure:
Self-Review Checklist:

Closes #53116

Video 

[Screencast from 2026-04-04 19-14-55.webm](https://github.com/user-attachments/assets/d17945d6-b17c-435d-8155-648cd7ba574b)

Release Notes:

- Fixed: File duplication rename now selects the entire filename stem instead of just the " copy" suffix, allowing users to type a new name without manually clearing text


smitbarmase (changes requested):
Have you run tests on this? It's weird to me that we don't have tests for this?

smitbarmase (approved):
Hey, thanks for updating the test. I don't think we need an extra redundant test, so I removed it. I think the current state is fine for this change.

ChristopherBiscardi (approved):

OmChillure:
> Hey, thanks for updating the test. I don't think we need an extra redundant test, so I removed it. I think the current state is fine for this change.

ahh thought soo, but addeed just in case. 

Thank you

PR #53150, opened 2026-04-04
markdown: Add LaTeX math rendering support

### LINKED ISSUE #10899: Markdown math support

n-WN:
## Summary

- Parse `InlineMath`/`DisplayMath` events from pulldown-cmark instead of silently discarding them (add `ENABLE_MATH` to parser options)
- Implement a custom GPUI backend for the [ReX](https://github.com/KenyC/ReX) LaTeX typesetting engine, converting glyph outlines to `Path` primitives and fraction bars/rules to `Quad` primitives
- Bundle XITS Math font (OFL-licensed) for OpenType MATH table support
- Add async rendering with `OnceLock` caching following the existing mermaid diagram pattern
- Display math (`$$...$$`) renders as centered block elements via ReX vector paths
- Inline math (`$...$`) converts LaTeX to Unicode text and renders as italic styled text, preserving natural text flow

### Screenshot

LaTeX math rendering in Zed markdown preview

### Architecture

```
Inline math ($...$):
  LaTeX source → latex_to_unicode() → italic styled text via push_text()

Display math ($$...$$):
  LaTeX source → pulldown-cmark → DisplayMath event
    → ReX parser → ReX layout engine → GpuiMathBackend
    → gpui::Path (glyph outlines) + gpui::Quad (rules/bars)
    → canvas element with paint_path() / paint_quad()
```

Display math rendering is cached per unique expression and runs on a background thread to avoid blocking the UI. Text color automatically adapts to the current theme (light/dark).

### Known Limitations

- Inline math uses Unicode approximation (superscripts ²³, Greek α β, operators ∩⊥·×) rather than full typeset rendering, since GPUI lacks inline replaced element support
- Uses [lnay's ReX fork](https://github.com/lnay/ReX/tree/personal) with `ttfparser-fontparser` feature — ideally this would be upstreamed or vendored
- Some LaTeX commands not supported by ReX (e.g. `\begin{matrix}` with escaped brackets) fall back to showing raw source

Closes #10899

## Test plan

- [ ] Verify display math (`$$E = mc^2$$`) renders as centered formula
- [ ] Verify inline math (`$x^2 + y^2$`) renders within paragraph text flow
- [ ] Verify complex expressions (fractions, subscripts, superscripts, radicals, matrices)
- [ ] Verify error fallback shows raw LaTeX source when parsing fails
- [ ] Verify rendering in both light and dark themes
- [ ] Verify caching works (same expression not re-rendered on scroll)
- [ ] Verify markdown preview in assistant panel with LLM-generated math

Release Notes:

- Added LaTeX math rendering in markdown preview (`$inline$` and `$$display$$` math)


BrannonKing:
Thanks for working on this! I'm really looking forward to real math equations in the AI chat window. Did you have some list of latex functions that you were working from? Or how did you decide which latex functions to support for the inline unicode translation? The unicode fraction slash character might be an interesting next step, as "frac" takes two parameters not one.

MrSubidubi:
Thank you very much for this and the time and effort that went into this! Furthermore, sorry for no reply yet - we are always cautious with such big changes and thus takes us some more time to get back here sometimes, more than we'd like.

Unfortunately, I am gonna close this out in favor of https://github.com/zed-industries/zed/pull/54668 - that PR brings some Windows fixes already as well as an overall less invasive approach for the issue at hand. Thus, I decided we'll push that PR to completion. However, I already informed the contributor there that we should also take learnings from this Pr so that your effort here does not go to waste.

Hope you can understand that, in any case, thank you very much for this PR!

n-WN:
> Thank you very much for this and the time and effort that went into this! Furthermore, sorry for no reply yet - we are always cautious with such big changes and thus takes us some more time to get back here sometimes, more than we'd like.
> 
> Unfortunately, I am gonna close this out in favor of #54668 - that PR brings some Windows fixes already as well as an overall less invasive approach for the issue at hand. Thus, I decided we'll push that PR to completion. However, I already informed the contributor there that we should also take learnings from this Pr so that your effort here does not go to waste.
> 
> Hope you can understand that, in any case, thank you very much for this PR!

I'm happy to see his effect, but do I have a chance to be a co-author, as some experimental attempts.

PR #53155, opened 2026-04-04
context_server: Add support for `roots/list` request

Vaiz:
proof of concept for https://github.com/zed-industries/zed/discussions/53156

### DIFF
--- a/crates/context_server/src/client.rs
+++ b/crates/context_server/src/client.rs
@@ -38,6 +38,7 @@ pub const INTERNAL_ERROR: i32 = -32603;
 type ResponseHandler = Box<dyn Send + FnOnce(String)>;
 type NotificationHandler = Box<dyn Send + FnMut(Value, AsyncApp)>;
 type RequestHandler = Box<dyn Send + FnMut(RequestId, &RawValue, AsyncApp)>;
+type RequestHandlerMap = HashMap<&'static str, (u64, RequestHandler)>;
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[serde(untagged)]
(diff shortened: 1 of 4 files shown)


osyvokon:
@benbrandt Could you take a look please? Do we want this MCP feature?

bennetbo:
My understanding is that [roots might get removed from MCP](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577) soon. So I don't think it's worth it to support this. If you have a specific MCP server in mind where this is really useful I am open to re-visit this decision.

Vaiz:
> My understanding is that [roots might get removed from MCP](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577) soon. So I don't think it's worth it to support this. If you have a specific MCP server in mind where this is really useful I am open to re-visit this decision.

thanks for the link, I didn't expect it would get removed

Aeron:
@bennetbo I don’t have a specific MCP server in mind, but this capability is the only way to reliably get a list of roots in the multi-project workspace when using a custom MCP server. Because of how Zed runs MCP commands. 

Here are the scenarios:
- If you quit Zed, the next time you start it, an MCP command will be invoked from the plain `/`. It doesn’t matter if it was the command-line tool or an app.
- If you activate a tab with a file from a different project in your workspace and cycle a relevant MCP server toggle in the agents settings panel, Zed will run it from a current file’s project directory. It looks quite similar to the `terminal.working_directory: current_project_directory` behavior, but Zed does not rely on that setting in this case.

Thus, CWD is not an option here. And I see no other way around, except for the `roots/list` capability or explicitly feeding the correct path into a command through its arguments. The latter would be a good option if Zed would have variables in settings, like `${workspaceFolder}` as VSCode has. But it does not.

So, we either need a `working_directory` setting for MCPs, so we can specify the `first_project_directory` behavior if I got the meaning of “first” right, or the soon-to-be-deprecated MCP capability.

Although, there are no MCP guidelines that would guarantee a particular order of such roots in a list, but it’s still better than nothing.

WDYT? Maybe I’m missing something. I’m willing to re-format it as a separate discussion or a bug, but at this point this PR seemed as the most relevant place to write it.

PR #53166, opened 2026-04-04
Add ChatGPT subscription provider via OAuth 2.0 PKCE

morgankrey:
Adds a new language model provider that lets users authenticate with their ChatGPT Plus/Pro subscription and use OpenAI models (codex-mini-latest, o4-mini, o3) directly in the Zed agent — without needing a separate API key.

## How it works

1. **OAuth 2.0 + PKCE sign-in**: Uses OpenAI's official Codex CLI client ID to run an authorization code flow. A local HTTP server on `127.0.0.1:1455` captures the callback, exchanges the code for tokens, and stores them in the system keychain.

2. **Token refresh**: Access tokens are automatically refreshed when they're within 5 minutes of expiry, using the stored refresh token.

3. **Responses API**: Requests go to `https://chatgpt.com/backend-api/codex/responses` using the existing `open_ai::responses` client (Responses API format, not Chat Completions which was deprecated for this endpoint in Feb 2026).

4. **Required headers**: `originator: zed`, `OpenAI-Beta: responses=experimental`, `ChatGPT-Account-Id` (extracted from JWT), `store: false` in the body.

## Files changed

- `crates/open_ai/src/responses.rs`: Add `store: Option<bool>` field to `Request`; add `extra_headers` param to `stream_response` for per-provider header injection
- `crates/language_models/src/provider/openai_subscribed.rs`: New provider (sign-in UI, OAuth flow, token storage/refresh, model list)
- `crates/language_models/src/provider/open_ai.rs`, `open_ai_compatible.rs`, `opencode.rs`: Pass `vec![]` for new `extra_headers` param
- `crates/language_models/src/language_models.rs`: Register the new provider
- `crates/language_models/Cargo.toml`: Add `rand` and `sha2` deps for PKCE

## Open questions / known gaps

- [ ] **Terms of service**: Usage appears to be within OpenAI's ToS (interactive use via their official CLI client ID), but needs legal sign-off before shipping
- [ ] **Redirect URI**: Currently `http://localhost:1455/auth/callback` — may need to match exactly what OpenAI's Codex CLI uses
- [ ] **UI polish**: The sign-in card is functional but minimal; needs design review
- [ ] **Error messages**: OAuth error responses from the callback URL aren't surfaced to the user yet
- [ ] **`o3` availability**: o3 may require a higher subscription tier; consider gating it

## Testing

Sign-in flow was designed to match the Copilot Chat provider pattern. Manual testing against the live OAuth endpoint is needed.

Release Notes:

- Added ChatGPT subscription provider, allowing users to use their ChatGPT Plus/Pro subscription with the Zed agent


crates/http_client/src/http_client.rs
```diff
@@ -443,3 +562,184 @@ impl HttpClient for FakeHttpClient {
         self
     }
 }
+
+// ---------------------------------------------------------------------------
+// Shared OAuth callback server (non-wasm only)
+// ---------------------------------------------------------------------------
+
+#[cfg(not(target_family = "wasm"))]
+mod oauth_callback_server {
```
agu-z: This should probably live in its own crate, not `http_client`. Same for the hardcoded html above.

crates/context_server/src/oauth.rs
```diff
@@ -1057,103 +1011,26 @@ const CALLBACK_TIMEOUT: Duration = Duration::from_secs(2 * 60);
...
+                code: params.code,
+                state: params.state,
+            }),
+            Ok(Err(e)) => Err(e),
+            Err(_) => Err(anyhow!("OAuth callback channel was cancelled")),
+        };
+        let _ = tx.send(result);
+    })
+    .detach();
+    Ok((redirect_uri, mapped_rx))
```
agu-z: Instead of detaching and returning this `mapped_rx`. I think we should just return the `Task` with the desired result type. Also, is there a reason we use `smol::spawn`? Can we use our own executor?

crates/language_models/src/provider/openai_subscribed.rs
```diff
@@ -0,0 +1,1478 @@
...
+    Gpt51Codex,
+    Gpt51CodexMax,
+    Gpt51CodexMini,
+    Gpt52,
+    Gpt52Codex,
+    Gpt53Codex,
+    Gpt53CodexSpark,
+    Gpt54,
+    Gpt54Mini,
+}
```
agu-z, line 277: It looks like codex uses `/backend-api/codex/models` to get the final list of models. Could we use that too so that we don't have to release a new version when they add a model? 

https://github.com/openai/codex/blob/main/codex-rs/codex-api/src/endpoint/models.rs

crates/language_models/src/provider/openai_subscribed.rs
```diff
@@ -0,0 +1,1478 @@
...
+    let weak_state = state.clone();
+    // Clear credentials and cancel in-flight work immediately so the UI
+    // reflects the sign-out right away.
+    weak_state
+        .update(cx, |s, cx| {
+            s.auth_generation += 1;
+            s.credentials = None;
+            s.sign_in_task = None;
+            s.refresh_task = None;
+            s.last_auth_error = None;
```
agu-z, line 957: Should these be combined under one `AuthState` enum? There may be a reason this isn't practical, I haven't dug into it yet

4ad:
@morgankrey are you still working on this?

morgankrey:
Yes

infused-kim:
@morgankrey, thank you so much for implementing this. 

This feature is more important than ever since agent change review with APC providers is essentially dead (See https://github.com/agentclientprotocol/claude-agent-acp/issues/333; https://github.com/zed-industries/codex-acp/issues/226).

## My testing results

I tried it briefly today, but had 2 issues:

### 1. Auth issue due to the callback url being different than Codex

Authentication didn't work. Instead of the login page, OpenAI showed this error message:
> An error occurred during authentication (unknown_error). Please try again.
> You can contact us through our help center at help.openai.com if you keep seeing this error. (Please include the request ID d51c42c9-3d25-4a71-a9be-c9eef2153c43 in your email.)

This was caused by the `redirect_url` parameter.  This PR was using `&redirect_uri=http://127.0.0.1:57423/callback`, but Codex seems to require `/auth/callback`. 

### 2. Unsupported request parameter `max_output_tokens`

Once I had the auth working, actual chat requests showed this error:
```json
invalid request format to ChatGPT Subscription's API: {"detail":"Unsupported parameter:
max_output_tokens"}
```

## Workaround

I added 2 commits that resolve these issues on this branch:
https://github.com/infused-kim/zed/tree/pr-orig/53166-chatgpt-oauth

## Cherry-pick on top of version 1.0.0

I wanted to integrate this feature into my personal fork. So, I cherry picked the commits on top of the v1.0.0 tag.

It can be found here:
https://github.com/infused-kim/zed/tree/pr-testing/53166-chatgpt-oauth

Please keep in mind that there were quiet a few merge conflicts since the original PR was developed against main. So, the code may not accurately represent OPs work and may not be 100% equivalent.

@morgankrey I hope you don't mind me sharing this. But if you do, please let me know and I will remove this portion of the comment.

gb-jos:
was curious if this is something that will come eventually? Have been tracking this for a while because i use codex and wanted to try out the zed agent.

crates/language_models/src/provider/openai_subscribed.rs
rtfeldman: I think we should do that long-term but shouldn't block this PR on it!

crates/language_models/src/provider/openai_subscribed.rs
rtfeldman: I looked into it and I think it's more invasive than it's worth for this PR.

crates/language_models/src/provider/openai_subscribed.rs
quinncomendant: Looks like it is missing `Gpt55`?

rtfeldman (approved):

PR #53168, opened 2026-04-04
editor: Immediately cancel popover task if the hover ends before the popover is shown

timvermeulen:
Hover popovers aren't immediately dismissed when the hover ends, but this delay mechanism is currently causing popovers to sometimes still briefly appear when the hover lasted too short to have triggered it. Fixed by immediately cancelling the popover task when the hover ends before the popover is displayed.

Most easily reproduced when setting the hover popover delay to a value lower than the default 300ms (but it still happens sometimes with the default value), because the popover hide delay is also 300ms. Here's it with `"hover_popover_delay": 200`:

https://github.com/user-attachments/assets/6415d112-d8e0-4a87-9a79-a7ab559f20f2

After the fix:

https://github.com/user-attachments/assets/34782389-de4c-4a25-bd6e-4858b55028de

Self-Review Checklist:

Release Notes:

- Fixed hover popovers sometimes briefly appearing after the hover already ended.


Veykril (approved):
Thanks!

PR #53170, opened 2026-04-04
dev_container: Fix various startup issues

zdeneklapes:
# Summary

  This merge request fixes multiple dev-container startup issues.

  It improves Docker Compose path resolution, preserves compose build configuration when generating
  feature builds, hardens remote user detection, and fixes Docker inspect parsing for images that omit
  Config.Labels. Together, these changes make dev containers more reliable across a wider range of
  setups, including plain base images such as debian:bookworm-slim.

# Per-Commit Breakdown

1. 43dc3cbcd1 devcontainer fixed
     This commit contains the main dev-container fixes.
    - Fixes Docker Compose dockerfile and build-context resolution, including cases where the Dockerfile lives outside the .devcontainer directory.
    - Preserves compose build args, volumes, and additional contexts when Zed generates feature-build overrides.
    - Improves remote user resolution by falling back through remoteUser, containerUser, image user, and finally root.
    - Fixes Docker chown invocation to use the resolved remote user correctly.
    - Updates remote server build logic to use the active rustup toolchain and stable workspace-root binary paths.
    - Adds regression coverage for compose configs, remote user fallback, and remote server path
        handling.
2. 0220310b3d dont require labels
  This commit fixes startup for images whose docker inspect output does not include Config.Labels.
    - Makes Docker inspect deserialization tolerant of missing Labels.
    - Adds a regression test based on a real debian:bookworm-slim inspect payload.
    - Prevents dev-container startup from failing with missing field Labels on plain images.

# Why These Changes Were Needed

Dev container startup was failing in several valid configurations:

- some Docker Compose setups used dockerfiles or build contexts that were not resolved correctly
- some images did not provide enough metadata for strict remote-user resolution
- some base images omitted Config.Labels entirely, which caused Docker inspect deserialization to fail

These changes make the implementation more tolerant of real-world Docker and Compose outputs while keeping the existing dev-container flow intact.

# Release Notes
- Fixed dev container startup for Docker Compose setups with non-standard dockerfile/build-context
    layouts and for images whose docker inspect output omits Config.Labels

### DIFF
--- a/crates/dev_container/src/devcontainer_manifest.rs
+++ b/crates/dev_container/src/devcontainer_manifest.rs
@@ -8,7 +8,7 @@ use std::{
 
 use fs::Fs;
 use http_client::HttpClient;
-use util::{ResultExt, command::Command};
+use util::{ResultExt, command::Command, paths::normalize_lexically};
 
 use crate::{
     DevContainerConfig, DevContainerContext,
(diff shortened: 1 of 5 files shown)


antont:
Hey @zdeneklapes thanks for the fixes! I'm just a user here, who encountered similar problems, and I'm now taking the Dockerfile path resolution fixes to use in the build I use for daily work.

However, I extracted the fix to only that problem in a dedicated branch, as I did not have time to check the other things and I don't want to have extra risks. I opened a PR with just the Dockerfile path fix too, https://github.com/zed-industries/zed/pull/53860

The zed devs will decide what to do with the fixes, I'm just mentioning for info.

PR #53177, opened 2026-04-05
bedrock: Refactor bedrock model capabilities

5herlocked:
Introduce a data-driven ModelCapabilities struct that centralizes all numeric and boolean model metadata (max tokens, output limits, tool/image support, thinking modes, cache config, etc.) into per-provider const definitions. This replaces 10 match-heavy methods on Model that each enumerated all 34 variants.

Adding a new model is now: define one const in the provider file, add one match arm to capabilities(). Previously it required touching 13 separate methods.

Changes:
- Add ModelCapabilities struct (no Default derive for compile-time safety)
- Break models/ into per-provider files (anthropic, meta, google, mistral, qwen, amazon, openai, minimax, moonshot, deepseek)
- Add Copy to BedrockModelCacheConfiguration and BedrockModelMode
- Add FromStr to BedrockAdaptiveThinkingEffort
- Widen from_id() to resolve all model variants, not just Claude
- Add tests for capabilities, from_id coverage, and FromStr

Self-Review Checklist:

Release Notes:

- Bedrock background changes to make model updates faster


bennetbo:
I am not sure if this is better than what we have right now. On one hand it makes adding a new model more concise, but it makes introducing new fields/parameters harder because macros are hard to read/write. Also we won't get rid of all match statements,`cross_region_inference_id` is still there and might potentially need to be updated when adding a new model.
Since we use the `match` approach for all other provider crates, I don't think it provides much value to use a different approach just for Bedrock, so I'd prefer to keep it as is for now. That said, you are adding new bedrock models all the time, so please let me now if you disagree.

5herlocked:
@bennetbo sorry also check out the commit `8ac793785527ee6210899f5fe43a266d36d0bdb9` on this branch/ PR it breaks all the models apart into their downstream providers.

Which was my initial question? Many files or a macro?

I don't hate the approach but I can also see that it isn't perfect.

But yeah totally makes sense, pass on the macro.

bennetbo:
Ah I misunderstood. Breaking them into individual files like in `8ac793785527ee6210899f5fe43a266d36d0bdb9` makes sense, happy to go with that approach.

bennetbo:
Closing for now, but feel free to open a new PR to make this change, thank you.

PR #53178, opened 2026-04-05
Gate `Format Selections` on whether the active formatter can actually format ranges

### LINKED ISSUE #25796: FormatSelections Command does not work

prertik:
## What Changed

- compute range-format support from formatter configuration and language-server capabilities
- show `Format Selections` only when at least one selected buffer has a range-capable formatter
- keep Prettier-supported range formatting available without depending on LSP support
- hide the action for formatter modes that cannot honor selections, such as external-command and code-action formatters
- keep the action handler safe by rechecking support and returning early when range formatting is unavailable
- restrict capability checks to the servers actually registered for the current buffer
- add regression coverage for supported, unsupported, and mixed-server cases
- document when `Format Selections` is available in both action docs and user docs

Self-Review Checklist:

Closes #25796

Release Notes:
- Improved `Format Selections` so it is only shown when the active formatter supports range formatting.

Testing I did:

Here are the formatters I used:
Screenshot_20260405_102047

Note: I've chosen three formatters:
1. Prettier, which supports range formatting
2. clangd (LSP) which also supports range formatting
3. gopls which do not support range formatting

[Screencast_20260405_102431.webm](https://github.com/user-attachments/assets/69ce3f95-0b73-46a0-853d-3b5be6329dde)

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -19605,11 +19605,11 @@ impl Editor {
         let snapshot = multi_buffer.snapshot(cx);
 
         self.selections
-            .disjoint_anchors_arc()
-            .iter()
-            .filter(|selection| selection.start != selection.end)
-            .flat_map(|selection| snapshot.buffer_ids_for_range(selection.range()))
-            .filter_map(|buffer_id| multi_buffer.buffer(buffer_id))
+            .disjoint_anchor_ranges()
+            .filter(|range| range.start != range.end)
+            .flat_map(|range| [range.start, range.end])
+            .filter_map(|anchor| snapshot.anchor_to_buffer_anchor(anchor))
+            .filter_map(|(_, buffer_snapshot)| multi_buffer.buffer(buffer_snapshot.remote_id()))
             .any(|buffer| project.supports_range_formatting(&buffer, cx))
     }
 
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -19610,8 +19610,6 @@ impl Editor {
             .filter(|selection| selection.start != selection.end)
             .flat_map(|selection| snapshot.buffer_ids_for_range(selection.range()))
             .filter_map(|buffer_id| multi_buffer.buffer(buffer_id))
-            .collect::<HashSet<_>>()
-            .into_iter()
             .any(|buffer| project.supports_range_formatting(&buffer, cx))
     }
 
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -253,6 +253,7 @@ pub(crate) const CURSORS_VISIBLE_FOR: Duration = Duration::from_millis(2000);
 #[doc(hidden)]
 pub const CODE_ACTIONS_DEBOUNCE_TIMEOUT: Duration = Duration::from_millis(250);
 pub const SELECTION_HIGHLIGHT_DEBOUNCE_TIMEOUT: Duration = Duration::from_millis(100);
+pub const FORMAT_SELECTIONS_AVAILABILITY_DEBOUNCE_TIMEOUT: Duration = Duration::from_millis(100);
 
 pub(crate) const CODE_ACTION_TIMEOUT: Duration = Duration::from_secs(5);
 pub(crate) const FORMAT_TIMEOUT: Duration = Duration::from_secs(5);
@@ -1216,6 +1217,8 @@ pub struct Editor {
     quick_selection_highlight_task: Option<(Range<Anchor>, Task<()>)>,
     debounced_selection_highlight_task: Option<(Range<Anchor>, Task<()>)>,
     debounced_selection_highlight_complete: bool,
+    format_selections_available: bool,
+    format_selections_availability_task: Task<()>,
     document_highlights_task: Option<Task<()>>,
     linked_editing_range_task: Option<Task<Option<()>>>,
     linked_edit_ranges: linked_editing_ranges::LinkedEditingRanges,
@@ -2146,6 +2149,9 @@ impl Editor {
                 project,
                 window,
                 |editor, _, event, window, cx| match event {
+                    project::Event::LanguageServerAdded(..) => {
+                        editor.refresh_format_selections_availability(false, cx);
+                    }
                     project::Event::RefreshCodeLens => {
                         // we always query lens with actions, without storing them, always refreshing them
                     }
@@ -2179,6 +2185,7 @@ impl Editor {
                         editor.register_visible_buffers(cx);
                         editor.invalidate_semantic_tokens(None);
                         editor.refresh_runnables(None, window, cx);
+                        editor.refresh_format_selections_availability(false, cx);
                         editor.update_lsp_data(None, window, cx);
                         editor.refresh_inlay_hints(InlayHintRefreshReason::ServerRemoved, cx);
                     }
@@ -2209,6 +2216,7 @@ impl Editor {
                         if editor.buffer().read(cx).buffer(buffer_id).is_some() {
                             editor.register_buffer(buffer_id, cx);
                             editor.refresh_runnables(Some(buffer_id), window, cx);
+                            editor.refresh_format_selections_availability(false, cx);
                             editor.update_lsp_data(Some(buffer_id), window, cx);
                             editor.refresh_inlay_hints(InlayHintRefreshReason::NewLinesShown, cx);
                             refresh_linked_ranges(editor, window, cx);
@@ -2271,6 +2279,11 @@ impl Editor {
                                 cx,
                             );
                         }
+                    }
+
+                    project::Event::DisconnectedFromHost
+                    | project::Event::DisconnectedFromRemote { .. } => {
+                        editor.refresh_format_selections_availability(false, cx);
                     }
 
                     _ => {}
@@ -2458,6 +2471,8 @@ impl Editor {
             quick_selection_highlight_task: None,
             debounced_selection_highlight_task: None,
             debounced_selection_highlight_complete: false,
+            format_selections_available: false,
+            format_selections_availability_task: Task::ready(()),
             document_highlights_task: None,
             linked_editing_range_task: None,
             pending_rename: None,
@@ -2754,6 +2769,8 @@ impl Editor {
             }
             editor.report_editor_event(ReportEditorEvent::EditorOpened, None, cx);
         }
+
+        editor.refresh_format_selections_availability(false, cx);
 
         editor
     }
@@ -3681,6 +3698,7 @@ impl Editor {
         }
 
         let selection_anchors = self.selections.disjoint_anchors_arc();
+        self.refresh_format_selections_availability(true, cx);
 
         if self.focus_handle.is_focused(window) && self.leader_id.is_none() {
             self.buffer.update(cx, |buffer, cx| {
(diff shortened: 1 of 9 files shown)


crates/editor/src/editor.rs
```diff
@@ -2142,6 +2145,9 @@ impl Editor {
                 project,
                 window,
                 |editor, _, event, window, cx| match event {
+                    project::Event::LanguageServerAdded(..) => {
```
SomeoneToIgnore: This is not needed, as will react on each and every langserver added, not the ones that are related to the editor.

`LanguageServerBufferRegistered` should be enough, any issues with that part?

crates/editor/src/editor.rs
```diff
@@ -2269,6 +2277,11 @@ impl Editor {
                         }
                     }
 
+                    project::Event::DisconnectedFromHost
```
SomeoneToIgnore: Why do we care about these?
Do they add or remove any language servers to change the formatting situation?

crates/editor/src/editor.rs
```diff
@@ -19586,6 +19604,66 @@ impl Editor {
         self.pending_rename.as_ref()
     }
 
+    fn enable_format_selections(&self) -> bool {
```
```suggestion
    fn format_selections_available(&self) -> bool {
```

crates/editor/src/editor.rs
```diff
@@ -19586,6 +19604,66 @@ impl Editor {
...
+                .disjoint_anchors_arc()
+                .iter()
+                .filter(|selection| selection.start != selection.end)
+                .flat_map(|selection| snapshot.buffer_ids_for_range(selection.range()))
+                .collect::<HashSet<_>>();
+
+            selected_buffer_ids
+                .into_iter()
+                .filter_map(|buffer_id| multi_buffer.buffer(buffer_id))
+                .collect::<Vec<_>>()
```
SomeoneToIgnore: Cannot we collect to a `HashSet` here, once only instead of 2 re-collections?

crates/editor/src/editor.rs
```diff
@@ -19614,6 +19692,10 @@ impl Editor {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) -> Option<Task<Result<()>>> {
+        if !self.can_format_selections(cx) {
```
SomeoneToIgnore: Seems a bit odd: we are working on hiding the action from the command panel.
If it's hidden, the formatting won't be run already; if it's not — the formatting can be run.

Why do we need this?

crates/editor/src/element.rs
```diff
@@ -552,13 +552,15 @@ impl EditorElement {
...
             }
         });
-        register_action(editor, window, |editor, action, window, cx| {
-            if let Some(task) = editor.format_selections(action, window, cx) {
-                editor.detach_and_notify_err(task, window, cx);
-            } else {
-                cx.propagate();
-            }
-        });
+        if editor.read(cx).enable_format_selections() {
```
SomeoneToIgnore: This is way better than I've proposed, nice!

But then, we do not need any debounce at all, no tasks too: when I was posting the description, I have been thinking of a __global__ way to disable the action, and here we do that per editor, so seems totally fine.

I also think we can remove the extra field (and all the setters) and call `can_format_selections` here straight away: all we do is access a few `Entity`'s data and iterate over <100 language servers or a single prettier property.
That does not seem heavy at all?

crates/project/src/lsp_store.rs
```diff
@@ -3965,6 +3964,7 @@ pub struct LspStore {
     pub lsp_server_capabilities: HashMap<LanguageServerId, lsp::ServerCapabilities>,
     semantic_token_config: SemanticTokenConfig,
     lsp_data: HashMap<BufferId, BufferLspData>,
+    registered_language_servers_by_buffer: HashMap<BufferId, HashSet<LanguageServerId>>,
```
SomeoneToIgnore: I really don't like this extra state, as it adds quite a complexity on top + as the current changes to the editor code show, updating such states can be error-prone.

Why do we need it and why the old way did not work well?

SomeoneToIgnore (changes requested):
Thank you, the way to disable the action seems way better than my proposal, but the implementation is quite odd.

Let's clean it up, as it looks we can do it much simpler.

crates/editor/src/editor.rs
prertik: Makes sense, I'll update to use that.

crates/editor/src/editor.rs
prertik: Addressed.

crates/editor/src/editor.rs
prertik: Addressed.

prertik:
> Let's clean it up, as it looks we can do it much simpler.

Hi @SomeoneToIgnore, I've cleaned it up based on your suggestions.

crates/project/src/lsp_store.rs
```diff
@@ -5003,37 +5002,142 @@ impl LspStore {
...
-                // available to be loaded (for SSH remote mode where adapters like
-                // ty/pylsp/pyright are registered via register_available_lsp_adapter
-                // but only loaded on the server side)
-                let is_relevant = registered_language_servers.contains(&server_status.name)
-                    || self.languages.is_lsp_adapter_available(&server_status.name);
-                is_relevant.then_some(server_id)
+            .collect::<Vec<_>>();
+
+        for configured_language_server in &settings.language_servers {
+            if configured_language_server == "..." || configured_language_server.starts_with('!') {
```
SomeoneToIgnore: Whatever happens here is very odd — we used to work just fine before without any magical strings and language server name parsing, why do we need it now and what is the issue we try to solve here?

SomeoneToIgnore (commented):
I've pushed a few fixes as there was a lot of unneeded allocations all over the place.
There's one very odd place left that we need to discuss.

crates/project/src/lsp_store.rs
prertik: This was added to keep `supports_range_formatting` aligned with language servers that are actually enabled for the current buffer, rather than any server status that happens to exist in the project.

The availability check for `FormatSelections` now flows through `Editor::can_format_selections` and `Project::suports_range_formatting`. That check needs to be buffer-specific; otherwise a buffer can pick up range-formatting support from an unrelated server.

In local mode, we can answer this precisely from `buffers_opened_in_servers`. The awkward part is the fallback path as, on the proxy side (or remote guest collab) we do not have the same exact buffer -> server ids map, so, we reconstruct the effective server set from the buffer's language_servers settings instead. Due to which the behaviour to check `...` and `starts_with(!)` were added which are official Zed's document language server ways to configure user-based disabling of certain language servers.

So, the intent here was: if a server is disabled via settings, or only included via `...`, the range-formatting availability check should follow the same effective server selection that the rest of Zed uses.

Having said this, I agree that the current shape is more ad hoc than ideal, and I can simplify this around the existing settings resolver/shared mapping if that is preferable.

SomeoneToIgnore: I see finally, thank you for elaborating.

So we're trying to solve 2 different, each with its own complexity, tasks at the same time.
That does not really work well and I do not see the point.
Cannot we keep https://github.com/zed-industries/zed/blob/6af5ca3f8036b937e9d84bdf2283e89986caca67/crates/project/src/lsp_store.rs#L5027-L5033 -ish idea for the remote (as the entirety of the remote code works this way anyway) and not improve it further in this area?

This might lead to more false-positives on the remote side, sure, but the proper fix should be more global than a drive-by one along this PR, as at least needs its own set of tests.

prertik: Yes, this makes sense. I'll make this change and push.

crates/project/src/lsp_store.rs
```diff
@@ -5003,37 +5002,130 @@ impl LspStore {
...
+        cx: &App,
+    ) -> HashSet<LanguageServerName> {
         let Some(language) = buffer.read(cx).language().cloned() else {
-            return false;
+            return HashSet::default();
         };
-        let registered_language_servers = self
+
+        let settings = LanguageSettings::for_buffer(buffer.read(cx), cx);
+        if !settings.enable_language_server {
```
SomeoneToIgnore: Sorry, I still do not understand much about the new changes: why are they needed now, given that we've agreed to keep the code similar to what's been before?

What are those settings, why do we deal with that on an `lsp_store` level, when we're supposed to __load__ all those data by this time already??

crates/project/src/lsp_store.rs
```diff
@@ -5003,37 +5002,130 @@ impl LspStore {
...
-            .collect::<HashSet<_>>();
-        self.language_server_statuses
-            .iter()
-            .filter_map(|(server_id, server_status)| {
-                // Include servers that are either registered for this language OR
-                // available to be loaded (for SSH remote mode where adapters like
-                // ty/pylsp/pyright are registered via register_available_lsp_adapter
-                // but only loaded on the server side)
-                let is_relevant = registered_language_servers.contains(&server_status.name)
-                    || self.languages.is_lsp_adapter_available(&server_status.name);
```
SomeoneToIgnore, line 5032: `is_lsp_adapter_available` is not used anywhere else, if we do not need it (why??) , we need to remove it.

crates/project/src/lsp_store.rs
```diff
@@ -5003,37 +5002,130 @@ impl LspStore {
...
+                .copied()
+                .collect();
+        }
+
+        let relevant_language_servers = self.relevant_server_names_for_capability_check(buffer, cx);
+        self.language_server_statuses
+            .iter()
+            .filter_map(|(server_id, server_status)| {
+                relevant_language_servers
+                    .contains(&server_status.name)
```
SomeoneToIgnore: There's a discrepancy in what a `name` is in the code, and I'm not 100% sure whether this `status.name` will work well in all cases.

The biggest problem are `node`-based servers, that might have a language server with name `node`, and the language server adapter having the right name, e.g. `typescript-language-server`.

Just to be on a safe side, I'd try to use the adapter — both the local and remote variants are supposed to have one, aren't they?

Which raises, yet another time, my confusion — why do we need to do that now this way, what was not working with the old code?

SomeoneToIgnore:
Sorry, `lsp_store` is a very important module and I want to understand what those changes are meant for: I would appreciate an elaboration on what's going on and the purpose of it all, given the https://github.com/zed-industries/zed/pull/53178#discussion_r3068838630 agreement that the old code was working ok for our needs.

prertik:
> Sorry, `lsp_store` is a very important module and I want to understand what those changes are meant for: I would appreciate an elaboration on what's going on and the purpose of all it, given the [#53178 (comment)](https://github.com/zed-industries/zed/pull/53178#discussion_r3068838630) agreement that the old code was working ok for our needs.

Apologies, I agreed to the change, but my commits during rebase, I added wrong changes and committed them. I'll amend and push again.

crates/project/src/lsp_store.rs
prertik: Corrected them, @SomeoneToIgnore. Apologies on wrong changes being made as commit.

crates/project/src/lsp_store.rs
prertik: Corrected them, @SomeoneToIgnore. Apologies on wrong changes being made as commit.

crates/project/src/lsp_store.rs
prertik: Corrected them, @SomeoneToIgnore. Apologies on wrong changes being made as commit.

SomeoneToIgnore (approved):
Thank you.

Given all the back-and-forth, I'll merge this after the next release this Wednesday, to be on a safe side.

prertik:
> Thank you.
> 
> Given all the back-and-forth, I'll merge this after the next release this Wednesday, to be on a safe side.

Sure, @SomeoneToIgnore. Thanks for reviewing this and giving me feedback.

PR #53184, opened 2026-04-05
markdown_preview: Support anchor link for headings

### LINKED ISSUE #18699: Markdown Preview clickable links to headings (ToC links, for example)
### Check for existing issues

- [X] Completed

### Describe the feature

In the Markdown preview, links to external files are clickable and open that file.  I would like to have a similar capability for links to headings using the [label](#go-to-this-heading) format.  This is possible in other editors (the obvious one being vscode with suitable extensions).

Ideally, clicking the link in the preview would scroll up/down to the matching heading/anchor in the preview.  Even better would be if the editor and preview panes were synced so the editor also scrolled to the relevant point (probably another feature request).

_No response_

dongdong867:
## What does this PR did

- Generate [GitHub-flavored heading slugs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links) for markdown headings
- Handle `[label](#heading)` same-document anchor links that scroll the preview and editor to the target heading
- Handle `[label](./file.md#heading)` cross-file anchor links that open the file, scroll the preview, and move the editor cursor to the heading

https://github.com/user-attachments/assets/ecc468bf-bed0-4543-a988-703025a61bf8

## What to test

- [ ] Create a markdown file with `[Go to section](#section-name)` links, verify clicking scrolls preview and editor
- [ ] Create two markdown files with cross-file links like `[See other](./other.md#heading)`, verify file opens and preview scrolls to heading
- [ ] Verify duplicate headings produce correct slugs (`heading`, `heading-1`)
- [ ] Verify external URLs (`https://...`) are unaffected

Self-Review Checklist:

Closes #18699

Release Notes:

- Added support for anchor links for headings in Markdown Preview.


crates/markdown/src/markdown.rs
```diff
@@ -261,6 +261,8 @@ pub struct Markdown {
     copied_code_blocks: HashSet<ElementId>,
     code_block_scroll_handles: BTreeMap<usize, ScrollHandle>,
     context_menu_selected_text: Option<String>,
+    heading_slugs: Vec<(SharedString, usize)>,
```
smitbarmase: This should be the part of the parsing code instead of the rendering code, as computation of the slug is the responsibility of the parsing step, not the actual place when we render. And also, instead of this being a vector where we'll have to make O(N) search for this slug to find the source offset, I think we should use a HashMap here.

crates/markdown/src/markdown.rs
```diff
@@ -492,6 +496,33 @@ impl Markdown {
...
+            .iter()
+            .find(|(s, _)| s.as_ref() == slug)
+            .map(|(_, index)| *index)
+    }
+
+    pub fn scroll_to_heading(&mut self, slug: SharedString) {
+        self.pending_heading_scroll = Some(slug);
+    }
+
+    fn sync_headings(&mut self, headings: &[RenderedHeading], cx: &mut Context<Self>) {
```
smitbarmase: We won't be needing this method anymore after you move this step to the parsing step.

crates/markdown/src/markdown.rs
```diff
@@ -1741,6 +1793,7 @@ impl Element for MarkdownElement {
                         builder.pop_div();
                     }
                     MarkdownTagEnd::Heading(_) => {
+                        builder.finalize_heading(range);
```
smitbarmase: Same as above.

crates/markdown/src/markdown.rs
```diff
@@ -2123,6 +2179,10 @@ struct MarkdownElementBuilder {
     rendered_lines: Vec<RenderedLine>,
     pending_line: PendingLine,
     rendered_links: Vec<RenderedLink>,
+    rendered_headings: Vec<RenderedHeading>,
```
smitbarmase: These four variables will go into the parsing step and become variables.

crates/markdown/src/markdown.rs
```diff
@@ -3254,6 +3375,61 @@ mod tests {
         assert!(!has_code_block(&Markdown::escape(diagnostic)));
     }
 
+    #[test]
+    fn test_generate_heading_slug() {
```
smitbarmase: These tests can go into the parsing file.

smitbarmase (changes requested):
Thanks, I added a few suggestions.

smitbarmase (approved):
Thanks, I made a few fixes on top of your commits before merging:

Have you tested it? I don't think the cross-file anchor scroll (`./file.md#heading`) was ever working. It was trying to scroll the current preview while opening a target editor, so the heading would never resolve. There was a bunch of dead code around this that just added complexity, so I cleaned it up. We can introduce cross-file scroll later.

I also removed the scroll logic from render. We should not handle state mutations there, it's unreliable and wrong. Render should purely be used for rendering. I moved it to a `window.defer` call in `handle_url_click` instead, which is a more deterministic way to handle it.

For separation of concern, I removed the `#` fragment handling from `MarkdownElement`. Any URL can have a `#` in it, so stripping it and treating it as an anchor at the shared markdown level is a bug. I fixed that handling and moved into markdown preview itself.

We now also do parsing for heading slug only for markdown preview.

Thanks again for working on this!

SomeoneToIgnore (approved):

mendeza:
Thanks @dongdong867

PR #53200, opened 2026-04-05
Fix sign-in popup appearing above all other applications 

### LINKED ISSUE #51043: GitHub Copilot Sign-In Popup in Zed Stays Always on Top Across All Applications Instead of Being Scoped to Zed Window
### Reproduction steps

1. Open zed where it's not signed in to Github Copilot.
2. Click on Github Copilot icon from the bottom bar and select Sign In to Copilot.
3. This opens up a popup which says `Use Github Copilot in Zed` and has a code which can be copied. 

Image

Current Behaviour - The popup stays on top of all windows and cannot be minimized. The only way to remove it is to click on Cancel button

Image

Here as you can see I'm on a chrome window but still the popup stays on  top.

Expected Behaviour- The popup acts as an overlay only inside Zed. When switching to other windows it shouldn't be on top.

Zed: v0.226.5
OS: Linux X11 fedora 42
Memory: 15 GiB
Architecture: x86_64
GPU: AMD Radeon Graphics (RADV RENOIR) || radv || Mesa 25.1.9

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

boobachad:
Self-Review Checklist:

Closes https://github.com/zed-industries/zed/issues/51043

Release Notes:

- Fixed Copilot sign-in popup appearing above all other applications instead of being scoped to the Zed window

https://github.com/user-attachments/assets/c0666d36-39b3-495c-a9a1-929b3e8410f1

### DIFF
--- a/crates/sidebar/src/sidebar.rs
+++ b/crates/sidebar/src/sidebar.rs
@@ -4727,6 +4727,36 @@ pub fn dump_workspace_info(
         )
         .ok();
 
+        // project_group_key_for_workspace internally reads the workspace,
+        // so we can only call it for workspaces other than this_entity
+        // (which is already being updated).
+        if let Some(mw) = &multi_workspace {
+            if *ws == this_entity {
+                let workspace_key = workspace.project_group_key(cx);
+                writeln!(output, "ProjectGroupKey: {workspace_key:?}").ok();
+            } else {
+                let effective_key = mw.read(cx).project_group_key_for_workspace(ws, cx);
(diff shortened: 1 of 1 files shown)


mikayla-maki:
I'm not sure if I agree this is desirable. We should at least keep it as a seperate window. Perhaps we should just change the window type to not be "pop up", but you're in the middle of an oauth flow? Seems like it should take some precedence..

PR #53216, opened 2026-04-06
Fix repeated prompts in opencode acp

OmChillure:
### Summary
Fixes duplicated Prompts/context in ACP threads after sending a message, as reported in #53201.

### Root Cause
The thread already inserts the user prompt optimistically at send time.  
If an ACP server also echoes UserMessageChunk updates for the same prompt, the same content is appended again, which can duplicate rendered context sections.

### Fix
Ignore echoed UserMessageChunk updates while a turn is actively running, so user prompt content is not appended twice.

### Validation
- Reproduced with OpenCode ACP flow from the issue.
- Confirmed duplication appears after send (not in the input box).
- Confirmed duplicate Prompts/context no longer appears with the fix.

### Video 

[Screencast from 2026-04-06 08-21-32.webm](https://github.com/user-attachments/assets/33075312-9af7-4dd5-a2a3-5e1169b80243)

### Self-Review Checklist

Closes #53201

Release Notes:

- Fixed duplicated Prompts/context in ACP conversations when servers echo user message chunks after send.

### DIFF
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -3452,11 +3452,7 @@ mod tests {
         let connection = Rc::new(FakeAgentConnection::new().on_user_message(
             |request, thread, mut cx| {
                 async move {
-                    let prompt = request
-                        .prompt
-                        .first()
-                        .cloned()
-                        .unwrap_or_else(|| "".into());
+                    let prompt = request.prompt.first().cloned().unwrap_or_else(|| "".into());
 
                     thread.update(&mut cx, |thread, cx| {
                         thread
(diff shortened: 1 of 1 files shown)


crates/acp_thread/src/acp_thread.rs
```diff
@@ -1387,7 +1387,12 @@ impl AcpThread {
     ) -> Result<(), acp::Error> {
         match update {
             acp::SessionUpdate::UserMessageChunk(acp::ContentChunk { content, .. }) => {
-                self.push_user_content_block(None, content, cx);
+                // We optimistically add the full user prompt before calling `prompt`.
+                // Some ACP servers echo user chunks back over updates, which would duplicate
+                // context chips/message text unless ignored while a turn is active.
+                if self.running_turn.is_none() {
+                    self.push_user_content_block(None, content, cx);
```
benbrandt, line 1399: I think it would be better to do some check of if we are currently in a user mesage and if the content would match the current message?

OmChillure: updated the logic and ran the same test already present

benbrandt: Potentially too "fuzzy" a match.. but given that they will probably replay the whole thing probably ok

benbrandt (approved):

PR #53242, opened 2026-04-06
sidebar: Fix behavior of "Remove Project" button

danilo-leal:
- Fix an issue where the "remove project" button, available in the header's ellipsis menu, wouldn't do anything if the sidebar contained only one project
- Fix another issue where attempting to remove a project when the sidebar has more than one project wouldn't actually remove it. This is fixed by cleaning up the project group keys after its been already removed.

Release Notes:

- N/A

### DIFF
--- a/crates/workspace/src/multi_workspace.rs
+++ b/crates/workspace/src/multi_workspace.rs
@@ -5,9 +5,7 @@ use gpui::{
     ManagedView, MouseButton, Pixels, Render, Subscription, Task, Tiling, Window, WindowId,
     actions, deferred, px,
 };
-#[cfg(any(test, feature = "test-support"))]
-use project::Project;
-use project::{DisableAiSettings, ProjectGroupKey};
+use project::{DisableAiSettings, Project, ProjectGroupKey};
 use settings::Settings;
 pub use settings::SidebarSide;
 use std::future::Future;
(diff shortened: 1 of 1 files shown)


crates/workspace/src/multi_workspace.rs
```diff
@@ -1040,26 +859,54 @@ impl MultiWorkspace {
...
+                return false;
+            }
+
+            workspace.update(cx, |workspace, cx| {
+                for id in worktree_ids {
+                    workspace.project().update(cx, |project, cx| {
+                        project.remove_worktree(id, cx);
+                    });
+                }
+            });
```
maxbrunsfeld, line 1083: If we remove the root folders from the current project, will that unexpectedly update multiple threads to have no root folders (the active thread, plus the last N open threads, which all stay tied to the workspace)?

I'm wondering if it would be better to actually remove the whole workspace (leaving no workspace behind), and add back an empty workspace. That would have the effect of replacing the active thread with a new empty draft, but I think that might be more expected when clicking "Remove Project", vs having the thread be updated.

maxbrunsfeld (approved):
I left one question about the desired behavior when there are threads associated with the current workspace. It also might be good to have a test in sidebar_tests that shows what happens when you do this.

But neither of those need to block merging this. We could also revisit afterward.

crates/workspace/src/multi_workspace.rs
danilo-leal: Follow-up: am I unassociating threads with folders by doing this? If so, that's not my intent... just really wanted to prevent the scenario in which I hit "Remove Project" and I am still left with that project in the sidebar, but ideally, without interfering with the project <> thread relationship...

danilo-leal: Pushed up the fix that doesn't mutate any state!

mikayla-maki (approved):

PR #53250, opened 2026-04-06
ai: Preserve `draft_prompt` when creating a new agent threads

### LINKED ISSUE #38028: AI: Input cleared when starting new thread in agent panel
### Summary

When creating a new thread in the agent panel, the input box content is cleared.

### Description

Steps to trigger the problem:
1. Open the agent panel
2. Type some text in the input box
3. Open a new thread (Via shortcut or via UI buttons) (Reproducible both for Zed or external agents (ACP) threads)
4. 💥 The input box is cleared, your message is lost. (Even if you go back to the previous thread the message will not be restored as a "draft").

Recording:
Image

**Expected Behavior**:
You shouldn't lose your previously typed message.

Ideally the message should remain in the input box, if that's not possible for technical reasons consider saving the draft of the unsent message on the previous thread to be able to recover it if needed.

**Actual Behavior**:
The input box is cleared/reset.

### Model Provider Details
- Mode: Agent Panel

**Remark:**

This might be a regression that occurred around the time the Gemini ACP integration was released

Zed: v0.204.0 (Zed Preview) 
OS: macOS 15.6.1
Memory: 64 GiB
Architecture: aarch64

AmaanBilwar:
- added a shared helper `take_active_draft_initial_content` to resolve initial content 
- `draft_prompt` is copied over to a new thread and is persisted in old thread with this behavior 

Self-Review Checklist:
- [x] no Unsafe blocks 

Closes #38028 

Release Notes:

- persist draft messages in ai thread when a new thread is created so we dont use any messages or draft prompts mistakenly 

Question: 
need some guidance on how to persist a thread when the `draft_prompt` is present but the message is unsent because currently having a draft prompt in the agent input box  and thenmaking a new agent thread copies the prompt over to the new thread(changes from this pr) but on going back to the old thread new thread is destroyed

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -207,12 +207,13 @@ pub fn init(cx: &mut App) {
                     if let Some(panel) = workspace.panel::<AgentPanel>(cx) {
                         workspace.focus_panel::<AgentPanel>(window, cx);
                         panel.update(cx, |panel, cx| {
+                            let initial_content = panel.take_active_draft_initial_content(cx);
                             panel.external_thread(
                                 action.agent.clone(),
                                 None,
                                 None,
                                 None,
-                                None,
+                                initial_content,
                                 true,
                                 window,
                                 cx,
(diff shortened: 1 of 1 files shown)


bennetbo (approved):
Thank you!

PR #53277, opened 2026-04-06
Fix vim replace not escaping `$`

phdavis1027:
Self-Review Checklist:

Closes ##42292

The text inserted in the search ('\$SEARCH') and replace ('$$OTHER') inputs of the top-panel is a little anti-aesthetic, but that seems out of scope for this issue.

Release Notes:

- '$' in the second clause of vim-style '%s/find/replace/g' actions is correctly escaped.


phdavis1027:
This passed my initial desk-check, but after sitting down to make some improvements it no longer seems to be behaving. I pulled down the latest from main, but I couldn't say for sure that that's what broke it. Marking as draft until I can make it work again.

EDIT: Re-opening because it turn out I was test from the Zed I was writing Zed in (preview), not the Zed I was writing. Oops! Fix still works as intended.

I looked into making the text populated by the search bar a little more aesthetic, but the existing `tree-sitter-regex` crate does not have anything concerned with the replacement clause. Alternatively, we could parse and manually apply highlights. `regex_automata` etc have a bunch of utilities which *almost* parse, but nothing that quite works as far as I can see. The "grammar" of the replacement clause is pretty straightforward, so manual parsing is an option, but it felt a little smelly replicating `regex` crate internals.

Altogether, since the replacement bar is not highlighted at all in today's Zed, I leave the design choice of how/if to do that for another commit, modulo other opinions.

ConradIrwin:
Thanks for this!

ConradIrwin:
@zed-zippy approved

PR #53285, opened 2026-04-07
lsp: Fix `vtsls` `completeFunctionCalls` setting being ignored

### LINKED ISSUE #53275: vtsls setting "completeFunctionCalls" does nothing
### Reproduction steps

1. Enable "completeFunctionCalls" in settings
2. Trigger the autocompletion of a function in a ts file.

Current behavior: function is autocompleted without parentheses
Expected behavior: function is autocompleted with parentheses

Zed: v0.230.2
OS: linux

### Attach Zed log file

_No response_

### Relevant Zed settings

```"lsp": {
    "vtsls": {
      "settings": {
        "typescript": {
          "suggest": {
            "completeFunctionCalls": true,
            "enabled": true,
          },
        },
      },
    },
  }
```

</details>

### Relevant Keymap

_No response_

_No response_

None

yeskunall:
Some language servers (notably `vtsls` with `completeFunctionCalls`) update `insertText/textEdit` during `completionItem/resolve` to add snippet content like function call parentheses. Previously, `Completion.new_text` was only set from the initial completion response and never re-derived after resolve, so the snippet text was silently discarded.

Re-derive `new_text` from the resolved `lsp_completion`. Only the text content is updated -- replace/insert ranges are left as anchors from the original response since the LSP ranges in the resolved text_edit are stale when completions are cached across keystrokes (re: #34094).

Self-Review Checklist:

Closes #53275.

Release Notes:

- Fixed the `vtsls` `completeFunctionCalls` setting being ignored

### DIFF
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -6587,9 +6587,16 @@ impl LspStore {
             **lsp_completion = resolved_completion;
             *resolved = true;
 
-            // Re-derive new_text from the resolved completion. Some servers
-            // (e.g. vtsls with completeFunctionCalls) update insertText/textEdit
-            // during resolve to add snippet content like function call parentheses.
+            // We must not use any data such as sortText, filterText, insertText and textEdit to edit `Completion` since they are not supposed to change during resolve.
+            // Refer: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion
+            //
+            // We still re-derive new_text here as a workaround for the specific
+            // VS Code TypeScript completion resolve flow that vtsls wraps:
+            // https://github.com/microsoft/vscode/blob/838b48504cd9a2338e2ca9e854da9cec990c4d57/extensions/typescript-language-features/src/languageFeatures/completions.ts#L218
+            //
+            // Some servers (e.g. vtsls with completeFunctionCalls) update
+            // insertText/textEdit during resolve to add snippet content like
+            // function call parentheses.
             //
             // vtsls resolve flow:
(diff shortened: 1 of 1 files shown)


crates/project/src/lsp_store.rs
```diff
@@ -6568,9 +6568,6 @@ impl LspStore {
             .into_response()
             .context("resolve completion")?;
 
-        // We must not use any data such as sortText, filterText, insertText and textEdit to edit `Completion` since they are not suppose change during resolve.
```
yeskunall, line 6571: Hmm -- on second thought. Smit’s comment should probably remain.

yeskunall: If all else looks good, just ping me and I’ll reintroduce it.

smitbarmase (changes requested):
Looks good overall. Since this still isn’t LSP-spec-correct, we should probably keep that comment.

> // We must not use any data such as sortText, filterText, insertText and textEdit to edit `Completion` since they are not suppose change during resolve.
> // Refer: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion


Also, let’s add a comment referencing the underlying resolve request in [typescript-language-features#L218](https://github.com/microsoft/vscode/blob/838b48504cd9a2338e2ca9e854da9cec990c4d57/extensions/typescript-language-features/src/languageFeatures/completions.ts#L218,) which `vtsls` simply wraps.

This fix should be treated as a workaround for this specific VSCode TypeScript extension case.

smitbarmase (approved):

PR #53300, opened 2026-04-07
Fix direnv not loading in Claude Code

### LINKED ISSUE #38988: AI: Of all the agent options Claude Code still has a different environment as the rest of zed.
### Summary

Zed actions, zed agent and gemini cli in agent panel have access to the right environment within a project, claude code in agent panel does not. relates to #37469 

### Description
I have a project that uses direnv to set the development environment. almost all of zed now has consistent access to this environment. Actions, the editor itself, zed agent, and the gemini cli agent from the agent panel have consistent access to the correct environment. Claude code agent from the agent panel does not.

Steps to trigger the problem:
1. Launch zed
2. run an action that is dependent of the dir environment watch it succeed.
3. ask zed agent in agent panel to run the same command and watch it succeed.
4. ask gemini cli in agent panel to run the same command and watch it succeed.
5. ask claude code in agent panel to run the same command and watch it fail due to having a different environment than the rest of zed.

**Expected Behavior**:

all of zed shares the same environment or at least correctly inherits their envs.

**Actual Behavior**:

Almost all of zed shares the same environment with claude code in the agent panel being the only exception.

### Model Provider Details
- Provider: Claude code
- Mode: Agent panel external agent claude code.

Zed: v0.205.5 (Zed) 
OS: macOS 15.7
Memory: 32 GiB
Architecture: aarch64

eskimor:
The previous fix added the environment in get_command, duplicating the logic for each implementation.

This fix restores get_command to only be concerned about the agent and not the project.

Instead we extend the local env handling already present in stdio() to also take care of direnv environment setup and not just setting of the current working directory.

This way we have a clean separation of concerns and have a solution that should work for all agents. get_command gets you the command for running the agent, setting up a local project specific environment is the duty of the caller.

This refactor also makes testing much easier, as we can now test in isolation and be reasonably sure it will work for all agents. 

Self-Review Checklist:

Closes https://github.com/zed-industries/zed/issues/38988

Release Notes:

- Fixed Claude Code not loading the direnv environment. It should now work for all agents.

### DIFF
--- a/crates/agent_servers/src/acp.rs
+++ b/crates/agent_servers/src/acp.rs
@@ -187,6 +187,67 @@ pub async fn connect(
 
 const MINIMUM_SUPPORTED_VERSION: acp::ProtocolVersion = acp::ProtocolVersion::V1;
 
+/// Resolved project context for spawning an agent server subprocess.
+///
+/// Contains the merged environment (project env as base, command-specific on top)
+/// and the working directory. Produced by [`resolve_agent_context`] and consumed
+/// by [`AcpConnection::stdio`] when building the actual process command.
+pub struct AgentContext {
+    pub path: PathBuf,
+    pub args: Vec<String>,
+    pub env: HashMap<String, String>,
+    pub cwd: Option<PathBuf>,
+}
+
+impl AgentContext {
+    pub fn into_command(self, cx: &mut AsyncApp) -> std::process::Command {
+        let shell = cx.update(|cx| TerminalSettings::get(None, cx).shell.clone());
+        let builder = ShellBuilder::new(&shell, cfg!(windows)).non_interactive();
+        let mut command =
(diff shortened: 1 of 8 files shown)


crates/agent_servers/src/acp.rs
```diff
@@ -187,6 +187,67 @@ pub async fn connect(
 
 const MINIMUM_SUPPORTED_VERSION: acp::ProtocolVersion = acp::ProtocolVersion::V1;
 
+/// Resolved project context for spawning an agent server subprocess.
+///
+/// Contains the merged environment (project env as base, command-specific on top)
+/// and the working directory. Produced by [`resolve_agent_context`] and consumed
+/// by [`AcpConnection::stdio`] when building the actual process command.
+pub struct AgentContext {
```
eskimor, line 195: Necessary to keep tests pure. Otherwise we would need to spawn a real process for inspection, including creating a real temporary working directory.

PR #53301, opened 2026-04-07
keymaps: Avoid capturing ctrl-r in the terminal

dinocosta:
These changes update the bindings introduced in https://github.com/zed-industries/zed/pull/52893 so as to avoid capturing `ctrl-r` in case the user is focused on the terminal. 

A lot of shells, for example, zsh and bash, use `ctrl-r` as their binding for history search, so we need to ensure that the terminal can still receive that key combination instead of capturing it and showing the open recent projects modal.

Self-Review Checklist:

Release Notes:

- N/A


ConradIrwin (approved):
Thanks for catching this

PR #53307, opened 2026-04-07
Stop highlighting selection matches in the search inputs

SomeoneToIgnore:
Follow-up of https://github.com/zed-industries/zed/pull/52553

Restores previous search inputs' behavior where no extra highlights were applied.

Before:

https://github.com/user-attachments/assets/38b6e70c-d5d5-4e06-abec-97d20af44f39

After: 

https://github.com/user-attachments/assets/6e4b3931-adf0-4c2a-afc3-f3c839fc9add

Release Notes:

- N/A


ConradIrwin (approved):
Thanks for adding this

PR #53311, opened 2026-04-07
project_panel: Add redo and restore support

dinocosta:
- Introduce `project_panel::Redo` action
- Update all platform keymaps in order to map `redo`/`ctrl-shift-z`/`cmd-shift-z` to the `project_panel::Redo` action

### Restore Entry Support

- Update both `Project::delete_entry` and `Worktree::delete_entry` to return the resulting `fs::TrashedEntry`
- Introduce both `Project::restore_entry` and `Worktree::restore_entry` to allow restoring an entry in a worktree, given the `fs::TrashedEntry`
  - Worth pointing out that support for restoring is not yet implemented for remote worktrees, as that will be dealt with in a separate pull request
  
### Undo Manager

- Split `ProjectPanelOperation` into two different enums, `Change` and `Operation`
    -  While thinking through this, we noticed that simply recording the operation that user was performing was not enough, specifically in the case where undoing would restore the file, as in that specific case, we needed the `trash::TrashedEntry` in order to be able to restore, so we actually needed the result of executing the operation.
    - Having that in mind, we decided to separate the operation (intent) from the change (result), and record the change instead. With the change being recorded, we can easily building the operation that needs to be executed in order to invert that change. 
    - For example, if an user creates a new file, we record the `ProjectPath` where the file was created, so that undoing can be a matter of trashing that file. When undoing, we keep track of the `trash::TrashedEntry` resulting from trashing the originally created file, such that, redoing is a matter of restoring the `trash::TrashedEntry`.
    - Refer to the documentation in the `project_panel::undo` module for a better breakdown on how this is implemented/handled.

- Introduce a task queue for dealing with recording changes, as well as undo and redo requests in a sequential manner
  - This meant moving some of the details in `UndoManager` to a `project_panel::undo::Inner` implementation, and `UndoManager` now serves as a simple wrapper/client around the inner implementation, simply communicating with it to record changes and handle undo/redo requests
  - Callers that depend on the `UndoManager` now simply record which changes they wish to track, which are then sent to the undo manager's inner implementation
  - Same for the undo and redo requests, those are simply sent to the undo manager's inner implementation, which then deals with picking the correct change from the history and executing its inverse operation
- Introduce support for tracking restore changes and operations
  - `project_panel::undo::Change::Restored` – Keeps track that the file/directory associated with the `ProjectPath` was a result of restoring a trashed entry, for which we now that reverting is simply a matter of trashing the path again
  - `project_panel::undo::Operation::Restore` – Keeps track of both the worktree id and the `TrashedEntry`, from which we can build the original `ProjectPath` where the trashed entry needs to be restored
- Move project panel's undo tests to a separate module `project_panel::tests::undo` to avoid growing the `project::project_panel_tests` module into a monolithic test module
  - Some of the functions in `project::project_panel_tests` were made `pub(crate)` in order for us to be able to call those from `project_panel::tests::undo`
  
### FS Changes

- Refactored the `Fs::trash_file` and `Fs::trash_dir` methods into a single `Fs::trash` method
    - This can now be done because `RealFs::trash_dir` and `RealFs::trash_file` were simply calling `trash::delete_with_info`, so we can simplify the trait
    - Tests have also been simplified to reflect this new change, so we no longer need a separate test for trashing a file and trashing a directory
- Update `Fs::trash` and `Fs::restore` to be async
  - On the `RealFs` implementation we're now spawning a thread to perform the trash/restore operation

Self-Review Checklist:

Relates to #5039

Release Notes:

- N/A


yara-blue (approved):
Paired most of this, reviewed other changes outside of pairing.

crates/fs/Cargo.toml
```diff
@@ -41,10 +42,7 @@ time.workspace = true
 util.workspace = true
 is_executable = "1.0.5"
 notify = "8.2.0"
-
-[target.'cfg(target_os = "macos")'.dependencies]
-objc.workspace = true
-cocoa = "0.26"
+trash = { git = "https://github.com/dinocosta/trash-rs", branch = "5039-macos-restore" }
```
yara-blue: update to zed-industries fork

dinocosta: Updated ✅

yara-blue (approved):

PR #53313, opened 2026-04-07
copilot: Wire up reasoning tokens for GPT models

TwistingTwists:
Fix two issues with reasoning support in the Copilot provider:

- Responses API path: use the user's thinking_effort setting instead of hardcoding Medium effort
- Chat Completions path: compute and pass thinking_budget when thinking is enabled, instead of unconditionally setting it to None

Self-Review Checklist:

Closes #52140

Release Notes:

- Fixed a bug where copilot wouldn't use the thinking level the user's have set


mikayla-maki (approved):

Kryan90:
Any chance of this getting merged soon? Having to use VSCode for high thinking for GPT models is big sad.

bennetbo:
Hey, this was not actually working for me. I pushed a commit matching VSCode's behaviour and it seems to work. Will test it a bit more and merge soon.

crates/copilot_chat/src/copilot_chat.rs
```diff
@@ -289,13 +289,8 @@ impl Model {
...
 
     pub fn supports_response(&self) -> bool {
-        self.supported_endpoints.len() > 0
-            && !self
-                .supported_endpoints
-                .contains(&ModelSupportedEndpoint::ChatCompletions)
-            && self
-                .supported_endpoints
-                .contains(&ModelSupportedEndpoint::Responses)
+        self.supported_endpoints
```
bennetbo, line 292: Changed this so that we prefer the responses API when it's available, since the old API does not seem to support reasoning_effort

PR #53314, opened 2026-04-07
Disable code actions button when buffer search bar is focused

Dnreikronos:
Self-Review Checklist:

Closes #53210

Release Notes:

- Fixed code actions toolbar button flickering when using Find to jump between search results

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -745,6 +745,10 @@ impl BufferSearchBar {
         self.query_editor_focused
     }
 
+    pub fn has_focused_editor(&self) -> bool {
+        self.query_editor_focused || self.replacement_editor_focused
+    }
+
     pub fn register(registrar: &mut impl SearchActionsRegistrar) {
         registrar.register_handler(ForDeployed(|this, _: &FocusSearch, window, cx| {
             this.query_editor.focus_handle(cx).focus(window, cx);
--- a/crates/zed/src/zed/quick_action_bar.rs
+++ b/crates/zed/src/zed/quick_action_bar.rs
@@ -131,7 +131,12 @@ impl Render for QuickActionBar {
             editor_value.edit_predictions_enabled_at_cursor(cx);
         let supports_minimap = editor_value.supports_minimap(cx);
         let minimap_enabled = supports_minimap && editor_value.minimap().is_some();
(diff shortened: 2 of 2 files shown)


Dnreikronos:
> Besides `enter` there are other ways to scroll through results, `cmd-g` in macOS in particular, both are bound to `search::SelectNextMatch` That, pane-scoped, action may be invoked without any search bar at all...
> 
> The video tries to showcase 2 problems related to it:
> 
> * in the start of the video after showing that this is the correct version I run off, I try to click around, first clicking the titlebar and the editor: this seem to reproduce the bug as the icon starts to flicker at all
> * later, the video hides the search bar at all, only to display how things start to flicker again
> 
>  blinks.still.mov 
> I do not understand generally why do we disable this button at all, based on some focus — we do place the caret on a certain code region, and if it's applicable, we should allow a person to click the button, no matter where the focus is — this is how it works now and it actually applies the actions just fine.
> 
> So, this does not look like a fix — a proper approach would be to explain **why** do we blink instead of blindly trying to patch things on a top level.

Thanks for the video and the feedback. You're right that my fix was surface-level.

I dug into the actual cause. The flickering comes from a timing mismatch in `selections_did_change` (`editor.rs:3760-3765`):

```rust
if old_cursor_position.to_display_point(&display_map).row()
    != new_cursor_position.to_display_point(&display_map).row()
{
    self.available_code_actions.take();  // clears state instantly
}
self.refresh_code_actions(window, cx);   // re-fetches with 250ms debounce
```

When the cursor moves to a different row, `available_code_actions` gets cleared immediately. But the refresh has a 250ms debounce before it sends the LSP request. During that window the toolbar re-renders, sees `None`, and disables the button. Once the LSP responds, the button re-enables. That's the flicker.

This explains why it happens with Find navigation, `cmd-g`, and even clicking around: they all trigger selection changes that hit this same path.

I agree with you that the button probably shouldn't be disabled based on focus. The fix should be keeping the stale actions visible until the refresh completes with new ones, instead of clearing them eagerly on row change. That way there's no gap where the toolbar sees empty state.

Can I reopen that PR here with that changes, or you prefer for me to open another PR?

SomeoneToIgnore:
I think it's better to open another one and actually test things thoroughly this time.

Would be amazing to learn something from this too, as it's not the first or the second or even the third time we're in this situation...

PR #53324, opened 2026-04-07
agent_panel: Refine git pickers design in the empty thread view

danilo-leal:
Follow-up to https://github.com/zed-industries/zed/pull/52979 with some design refinements.

Release Notes:

- N/A


Anthony-Eid (approved):
Nice work!

Issue #20340, opened 2024-11-07
Require documentation when new settings are implemented.


tzabbi:
### Check for existing issues

- [X] Completed

### Describe the feature

Yesterday I studied the release notes for Zed [0.160.7](https://github.com/zed-industries/zed/releases/tag/v0.160.7) and see a new feature of [git blame](https://github.com/zed-industries/zed/pull/19759), which is disabled by default. I'm missing the setting in the zed documentation. For new user or user which does not read the release notes, it is impossible to know about this feature and maybe open an issue. 

To avoid this, my proposal is, that when a new settings change will be implemented it is required to update the zed documentation.  

### Environment

Zed.dev

maxdeviant:
It is documented here: https://zed.dev/docs/configuring-zed#inline-git-blame

<img width="793" alt="Screenshot 2024-11-15 at 12 51 21 PM" src="https://github.com/user-attachments/assets/4e5a1f40-cd9b-4443-9348-8a21a8dd4810">



Documentation was added in https://github.com/zed-industries/zed/pull/20006.

tzabbi:
Okay I missed it sorry.

But nevertheless I think a process like a merge template, with a checklist, would be a great improvment to my mind.

PR #53325, opened 2026-04-07
settings_content: Add more documentation and warn on missing docs

sinder38:
Self-Review Checklist:

Closes #ISSUE
https://github.com/zed-industries/zed/issues/20340
Partially https://github.com/zed-industries/zed/issues/53323

Release Notes:

- Fix settings documentation

### DIFF
--- a/crates/settings_content/src/agent.rs
+++ b/crates/settings_content/src/agent.rs
@@ -1,3 +1,5 @@
+#![allow(missing_docs)]
+
 use collections::{HashMap, IndexMap};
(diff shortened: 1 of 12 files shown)


probably-neb:
thanks for the PR. I think this a good direction, and gave me an idea for how we could generate `default.json` instead of trying to keep the multiple sources of truth in sync. Will hopefully push a branch for that soon. In the meantime I'm going to close this due to conflicts, and because I don't think we want to do this only part way (e.g. having the `#[allow(missing_docs)]`).

PR #53327, opened 2026-04-07
ci: Add workflow dispatch trigger for compliance workflow

MrSubidubi:
Also refines the reporting a bit.

Release Notes:

- N/A


ConradIrwin (approved):
Thanks

PR #53333, opened 2026-04-07
sidebar: Rename next/prev actions and small fix

cameron1024:
Correctness fix + renaming the actions
Release Notes:

### DIFF
--- a/crates/workspace/src/multi_workspace.rs
+++ b/crates/workspace/src/multi_workspace.rs
@@ -130,13 +130,7 @@ pub trait Sidebar: Focusable + Render + EventEmitter<SidebarEvent> + Sized {
     }
 
     /// Activates the next or previous project group.
-    fn cycle_project(
-        &mut self,
-        _forward: bool,
-        _window: &mut Window,
-        _cx: &mut Context<Self>,
-    ) {
-    }
+    fn cycle_project(&mut self, _forward: bool, _window: &mut Window, _cx: &mut Context<Self>) {}
 
     /// Activates the next or previous thread in sidebar order.
     fn cycle_thread(&mut self, _forward: bool, _window: &mut Window, _cx: &mut Context<Self>) {}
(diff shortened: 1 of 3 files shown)


crates/workspace/src/multi_workspace.rs
```diff
@@ -130,13 +130,7 @@ pub trait Sidebar: Focusable + Render + EventEmitter<SidebarEvent> + Sized {
...
 
     /// Activates the next or previous project group.
-    fn cycle_project_group(
-        &mut self,
-        _forward: bool,
-        _window: &mut Window,
-        _cx: &mut Context<Self>,
-    ) {
-    }
+    fn cycle_project(&mut self, _forward: bool, _window: &mut Window, _cx: &mut Context<Self>) {}
```
mikayla-maki: Can we remove this from the sidebar entirely?

crates/workspace/src/multi_workspace.rs
```diff
@@ -498,6 +434,20 @@ impl MultiWorkspace {
...
+    fn dispatch_to_sidebar(
+        &self,
+        action: Box<dyn gpui::Action>,
+        window: &mut Window,
+        cx: &mut App,
+    ) {
+        if let Some(sidebar) = &self.sidebar {
+            let focus_handle = sidebar.focus_handle(cx);
+            window.defer(cx, move |window, cx| {
+                focus_handle.dispatch_action(&*action, window, cx);
```
mikayla-maki, line 446: Rather than dispatching to the sidebar, we should do this more like the tab switcher. That way, we keep sidebar related logic in the sidebar, but still can have global actions:

https://github.com/zed-industries/zed/blob/main/crates/tab_switcher/src/tab_switcher.rs#L66

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been inactive for some time and has conflicts now.

PR #53340, opened 2026-04-07
multibuffer: Speed up anchor resolution by avoiding an excerpts seek

cole-miller:
Before, every creation of `AnchorSeekTarget` called `MultiBufferSnapshot::buffer_for_path`, which seeks the excerpts sum tree. Now we avoid that by

- looking up the buffer snapshot for the anchor's buffer in the `buffers` map (keyed by `BufferId`)
- handling the case where the anchor's buffer doesn't exist at the original path key with a separate `AnchorSeekTarget::Missing` variant

We also added a separate optimization to `AnchorSeekTarget::cmp`, to skip the full `PathKey` comparison when the cursor's `PathKeyIndex` is the same as the anchor's. This should help with singleton multibuffers.

Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/crates/multi_buffer/src/anchor.rs
+++ b/crates/multi_buffer/src/anchor.rs
@@ -34,17 +34,22 @@ pub enum Anchor {
     Max,
 }
 
-pub(crate) enum AnchorSeekTarget {
-    Excerpt {
-        path_key: PathKey,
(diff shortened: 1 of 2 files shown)


ConradIrwin (approved):
Makes sense to me, thanks!

PR #53342, opened 2026-04-07
Remove old Buffer branching APIs

maxbrunsfeld:
`Buffer::branch` was one API of many standing in the way of adopting a new CRDT as the basis for our buffer. This PR removes it, and adjusts its one legitimate user, `Buffer::preview_edits` to use a different, simpler API, `BufferSnapshot::edit`, that lets us apply edits a buffer snapshot that will never be merged into the main buffer.

Release Notes:

- N/A


SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been inactive for some time and has conflicts now.

PR #53359, opened 2026-04-08
settings_ui: Fix hover popover delay setting's json_path

daydalek:
The "Hover Popover → Delay" setting item had its json_path set to "hover_popover_enabled" (copy-paste error from the item above it), introduced in #40739. This caused:

- "Copy Link" generating a wrong zed://settings/hover_popover_enabled URL
- #hover_popover_delay filter returning no results in settings search
- Telemetry reporting delay changes as hover_popover_enabled

The pick/write closures were always correct, so the setting itself functioned normally.

after the fix, filter works correctly

image

and the copy link button return the correct result
image

Release Notes:

- Fixed an issue where "Hover Popover Delay" setting could not be searched or linked correctly in the settings UI.


SomeoneToIgnore (approved):
Thank you!

PR #53378, opened 2026-04-08
gpui: Fix scrollbar drag position calculation in list

KlausUllrich:
## Summary

Fixes scroll position inversion when content height changes during a scrollbar drag (e.g. in the agent panel while streaming responses).

### Root cause

`set_offset_from_scrollbar` used the **live** `items.summary().height` for `content_height` and computed a `drag_offset` correction (`live − frozen`). However, `max_offset_for_scrollbar()` → `max_scroll_offset()` already uses the **frozen** `scrollbar_drag_start_height` during drag. This mismatch means:

1. The scrollbar computes thumb position using the frozen height range
2. `set_offset_from_scrollbar` converts that position using the live height range
3. As content grows during drag, `drag_offset` increases
4. `(point.y - drag_offset).abs()` produces an incorrectly large value, overshooting `scroll_max`

The `.abs()` call also masked the sign convention: `compute_click_offset` (in `scrollbar.rs`) returns `-max_offset * percentage` — a negative value — but `.abs()` converted it regardless of sign, hiding the mismatch.

### Fix

- Use `scrollbar_drag_start_height` (frozen during drag) for `content_height` in `set_offset_from_scrollbar`, matching `max_scroll_offset()`. Both sides of the scrollbar mapping now use the same height reference.
- Replace `(point.y - drag_offset).abs()` with `(-point.y)` to correctly convert the negative scroll offset.
- Remove the symmetric `drag_offset` from `scroll_px_offset_for_scrollbar`, which reported position back to the scrollbar and suffered from the same inconsistency.
- Update existing tests to use negative offsets, matching the actual values produced by the scrollbar component.
- Add a regression test that freezes height, grows content, drags the scrollbar, and asserts the position is correct.

## Test plan

- [x] Existing tests updated to use correct sign convention (negative offsets)
- [x] New regression test `test_scrollbar_drag_with_growing_content`: verifies scroll position remains correct when content grows during a scrollbar drag
- [ ] Manual: open agent panel, start a long generation, drag scrollbar while content is streaming — position should track the thumb correctly

Release Notes:

- Fixed scrollbar position jumping or inverting when content height changes during a scrollbar drag (e.g. in the agent panel while streaming).

### DIFF
--- a/crates/gpui/src/elements/list.rs
+++ b/crates/gpui/src/elements/list.rs
@@ -709,8 +709,7 @@ impl ListState {
     /// Returns the current scroll offset adjusted for the scrollbar.
     ///
     /// The returned offset has a negative `y` component representing
-    /// how far the content has scrolled, consistent with the GPUI
-    /// scroll convention used by [`ScrollableHandle::offset`].
+    /// how far the content has scrolled.
     pub fn scroll_px_offset_for_scrollbar(&self) -> Point<Pixels> {
         let state = &self.0.borrow();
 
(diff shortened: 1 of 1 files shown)


smitbarmase (approved):

smitbarmase:
Thanks, I noticed that tail follow was still broken in the Agent Panel even after the rebase following your changes, so I made a few fixes to address it. We now have an additional test covering it alongside the one you already added. Thanks again, this is a great improvement.

MrSubidubi (approved):

PR #53381, opened 2026-04-08
gpui: Lazy-init font DB in SvgRenderer to avoid per-test overhead

Veykril:
Commit eaf14d028a6c changed SvgRenderer::new() to eagerly deep-clone the system font database and enrich it with bundled fonts at construction time. Since every #[gpui::test] creates a TestAppContext → App::new_app() → SvgRenderer::new(), and nextest runs each test in its own process, this added ~2-3s of overhead to every GPUI-based test (~132 minutes total across the full suite).

Move the expensive work (deep-clone + load_bundled_fonts + fix_generic_font_families) into a OnceLock inside the font resolver closure, so it only executes on the first actual SVG render. Tests that never render SVGs thus do not need to load the fonts which in itself can be fairly expensive.

This also bumps the opt-level for crane lift and some other wasmtime crates, as only wasmtime isn't really sufficient

Release Notes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -902,6 +902,9 @@ proc-macro2 = { opt-level = 3 }
 taffy = { opt-level = 3 }
 resvg = { opt-level = 3 }
 wasmtime = { opt-level = 3 }
+cranelift-codegen = { opt-level = 3 }
+wasmtime-environ = { opt-level = 3 }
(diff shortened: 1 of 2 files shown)


mikayla-maki:
@zed-zippy approve

mikayla-maki:
@zed-industries/approved

PR #53387, opened 2026-04-08
language_selector: Center selected item in picker window

cppcoffee:
Release Notes:

- N/A

After opening the picker window, the language item is displayed in the middle of the list.

### DIFF
--- a/crates/language_selector/Cargo.toml
+++ b/crates/language_selector/Cargo.toml
@@ -29,4 +29,5 @@ workspace.workspace = true
 
 [dev-dependencies]
 editor = { workspace = true, features = ["test-support"] }
+picker = { workspace = true, features = ["test-support"] }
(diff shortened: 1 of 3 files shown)


smitbarmase:
Hey, thanks for this PR, but looking at the UX perspective of this, this feels a bit weird to see the selected item in the center. At least I don't see that in any other app and it's not very predictable behavior personally for me atleast. Also might feel a bit inconsistent with all the pickers we have across.

I think the current behavior is fine where if the item is in the visible area, the list doesn't scroll, and if the item is off screen, we just scroll it for it to be the last item.

If the reason for you introducing this change is the visibility of the off screen items, maybe we should think about introducing a scrollbar for pickers. I will leave that to our design team to handle, so I'm closing this for now. Feel free to comment here if you feel otherwise.

PR #53401, opened 2026-04-08
language_models: Add API key helper for OpenAI-compatible providers

fflorensa-dd:
Self-Review Checklist:

Release Notes:

- Added support for api_key_helper on OpenAI-compatible providers, allowing API keys to be fetched from external commands like secret managers

### DIFF
--- a/crates/language_models/src/provider/open_ai_compatible.rs
+++ b/crates/language_models/src/provider/open_ai_compatible.rs
@@ -1,7 +1,7 @@
-use anyhow::Result;
+use anyhow::{Context as _, Result};
 use convert_case::{Case, Casing};
 use credentials_provider::CredentialsProvider;
-use futures::{FutureExt, StreamExt, future::BoxFuture};
+use futures::{FutureExt, StreamExt, future::BoxFuture, future::Shared};
 use gpui::{AnyView, App, AsyncApp, Context, Entity, SharedString, Task, Window};
 use http_client::HttpClient;
 use language_model::{
@@ -18,9 +18,11 @@ use open_ai::{
 };
 use settings::{Settings, SettingsStore};
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 use ui::{ElevationIndex, Tooltip, prelude::*};
 use ui_input::InputField;
 use util::ResultExt;
+use util::shell::ShellKind;
 
 use crate::provider::open_ai::{
     OpenAiEventMapper, OpenAiResponseEventMapper, into_open_ai, into_open_ai_response,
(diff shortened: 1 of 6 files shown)


crates/settings_content/src/language_model.rs
```diff
@@ -250,6 +250,20 @@ impl MergeFrom for OpenAiReasoningEffort {
...
     pub api_url: String,
     pub available_models: Vec<OpenAiCompatibleAvailableModel>,
+    /// Configuration for an external command that provides the API key.
+    pub api_key_helper: Option<ApiKeyHelperConfig>,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema, MergeFrom)]
+pub struct ApiKeyHelperConfig {
+    /// The shell command to execute via `sh -c`. Should output the API key on stdout.
+    pub command: String,
```
probably-neb, line 260: This should be `Option<String>`

crates/settings_content/src/language_model.rs
```diff
@@ -250,6 +250,20 @@ impl MergeFrom for OpenAiReasoningEffort {
...
+    /// Cache lifetime in seconds for the key returned by the command.
+    /// Values above 604800 (one week) are clamped to that maximum.
+    /// When omitted, the key is fetched once per Zed session and only refreshed
+    /// when the helper settings change or Zed is restarted.
+    #[schemars(range(min = 0, max = 604800))]
+    pub ttl_seconds: Option<u64>,
 }
```
probably-neb, line 267: We should have defaults for this type in the `default.json` file.  Each field needs a default value so the settings UI has something to display. For strings we generally use the empty string as the default value, and then when parsing the field we do a `.filter(|s| !s.is_empty())` followed by a `.map(...)` into the real type

crates/settings_content/src/language_model.rs
```diff
@@ -250,6 +250,20 @@ impl MergeFrom for OpenAiReasoningEffort {
 pub struct OpenAiCompatibleSettingsContent {
     pub api_url: String,
     pub available_models: Vec<OpenAiCompatibleAvailableModel>,
+    /// Configuration for an external command that provides the API key.
+    pub api_key_helper: Option<ApiKeyHelperConfig>,
```
probably-neb, line 254: We don't have to worry about settings UI for now because I believe these settings are in the agent panel settings rather than the settings UI (hopefully will change soon). However, if there's an obvious way to add the settings to the agent panel settings we should do it for visibility

crates/language_models/src/provider/open_ai_compatible.rs
```diff
@@ -32,6 +34,24 @@ pub use settings::OpenAiCompatibleModelCapabilities as ModelCapabilities;
...
+    pub api_key_helper: Option<ApiKeyHelperConfig>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct ApiKeyHelperConfig {
+    pub command: String,
+    pub ttl_seconds: Option<u64>,
+}
+
+const MAX_TTL_SECONDS: u64 = 7 * 24 * 60 * 60;
```
probably-neb, line 46: why?

probably-neb (changes requested):
I feel like this solution is a bit too localized to the OpenAI compatible keys.
I think much of the code that lives in the OpenAI crate currently, should actually live in the ApiKey type, rather than just leaving the implementation of running and using the helper up to the consuemer of the ApiKey type.

crates/language_models/src/provider/open_ai_compatible.rs
fflorensa-dd: Wanted to wrap the TTL to something that is not crazy high and could overflow the `expires_at: Option<Instant>` and since this was though as a way to have a helper for "short" lived key, a week seemed to be okay

fflorensa-dd:
> We don't have to worry about settings UI for now because I believe these settings are in the agent panel settings rather than the settings UI (hopefully will change soon). However, if there's an obvious way to add the settings to the agent panel settings we should do it for visibility

Thanks for the review, will propose changes

bennetbo:
Closing due to inactivity. Also, I'm not sure if we actually need this feature, and if we want it it should probably work for other providers too.

PR #53412, opened 2026-04-08
Fetch refs in compliance tools

ConradIrwin:
v0.232.x

Release Notes:

- N/A

### DIFF
--- a/tooling/compliance/src/git.rs
+++ b/tooling/compliance/src/git.rs
@@ -343,6 +343,32 @@ impl FromStr for NoOutput {
     }
 }
 
+pub struct FetchRefs {
+    branch: String,
+}
+
+impl FetchRefs {
+    pub fn new(branch: String) -> Self {
+        Self { branch }
+    }
+}
+
+impl Subcommand for FetchRefs {
+    type ParsedOutput = NoOutput;
+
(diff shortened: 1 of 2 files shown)


crates/zed/RELEASE_CHANNEL
```diff
@@ -1 +1 @@
-dev
+preview
```
MrSubidubi, line 1:
```suggestion
dev
```

dinocosta: Yep, I changed the base to `v0.232.x` as I suspect we might want to land this in the preview branch, but will double-check with Conrad! 🫡

dinocosta (approved):

MrSubidubi (commented):
I do think they would be available in the env already as per https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches, alternatively, we could/should also `fetch-tags` on the checkout action instead: https://github.com/actions/checkout?tab=readme-ov-file#usage

crates/zed/RELEASE_CHANNEL
dinocosta: Confirmed with Conrad we want to try and land this in `v0.232.x` for now and, if it fixes the issue, make sure it's also available in `main` ✅

PR #53415, opened 2026-04-08
Improve MCP OAuth resilience for buggy servers

tomhoule:
Two fixes for MCP OAuth discovery against servers like Pydantic Logfire that deviate from the spec in minor ways.

First, when the resource_metadata URL from the WWW-Authenticate header is same-origin but points to a broken endpoint (Logfire doubles the path component, producing /mcp/mcp), fall back to the RFC 9728 well-known URIs instead of failing outright. The header URL is still tried first per the MCP spec requirement.

Second, mirror the authorization server's grant_types_supported in the DCR registration body instead of hardcoding just authorization_code. Logfire's auth server requires both authorization_code and refresh_token in grant_types, and Zed already uses refresh tokens, so the only issue was not advertising the capability during registration. The DCR body now intersects our supported grant types with what the server advertises, or sends all of ours when the server metadata omits grant_types_supported.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

### DIFF
--- a/crates/context_server/src/oauth.rs
+++ b/crates/context_server/src/oauth.rs
@@ -146,6 +146,7 @@ pub struct AuthServerMetadata {
     pub token_endpoint: Url,
     pub registration_endpoint: Option<Url>,
     pub scopes_supported: Option<Vec<String>>,
+    pub grant_types_supported: Option<Vec<String>>,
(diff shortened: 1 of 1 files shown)


tomhoule:
Splitting this into two separate PRs instead.

PR #53420, opened 2026-04-08
extensions_ui: Make extension tab appearance consistent with other tabs

AccMoment:
Self-Review Checklist:

Release Notes:

- make extension tab appearance consistent with other tabs

### before extensions tab appearance
1f62b630-a414-468c-8450-a4e899d4b699

### after extensions tab appearance
QQ_1775666191571


MrSubidubi (changes requested):
Thanks! Can't help but wonder if we could not make the style more standardized by default across the trait. Do you have more context on that? E.g., if we can make some broader changes to the `tab_content` method in the trait to prevent this from happening in the future, I think I'd like for this here to explore into that direction

PR #53453, opened 2026-04-08
Remove git2 (libgit2) dependency

### LINKED ISSUE #46747: Git status indicators not working with reftable storage backend
# Git status indicators not working with reftable storage backend

## Summary

Git status indicators (file colors in project panel, gutter diff markers, branch name) do not appear when opening repositories that use Git's `reftable` reference storage format (`extensions.refstorage=reftable`).

This is a regression report / reopen request for #14414, which was closed on Dec 26, 2025 but the issue persists in Zed 0.218.7.

## Environment

- **Zed version:** 0.218.7
- **OS:** macOS (Darwin 25.2.0)
- **Git version:** 2.x with reftable support

## Steps to Reproduce

```bash
# Create a test repo with reftable format
mkdir zed-reftable-test && cd zed-reftable-test
git init --ref-format=reftable
git config user.email "test@example.com"
git config user.name "Test"
echo "hello" > test.txt
git add . && git commit -m "Initial"
echo "modified" >> test.txt

# Open in Zed
zed .
```

## Expected Behavior

- Branch name should appear in the git panel
- `test.txt` should be highlighted in the project panel

## Actual Behavior

- No branch name displayed
- No file status colors in project panel

## Potential Cause

Zed uses `libgit2` (via the `git2` Rust crate) for git operations. libgit2 does not yet support the reftable format. See: https://github.com/libgit2/libgit2/pull/7117

## Workaround Suggestion

As mentioned in #14414 comments, a potential workaround would be to shell out to the `git` CLI for repositories where libgit2 fails to initialize, falling back gracefully until libgit2 gains reftable support.

## Impact

This affects users of:

- Large monorepos that use reftable for performance (reftable is faster for repos with many refs)
- Any repository cloned with `--ref-format=reftable`
- Repositories where git is configured with `init.defaultRefFormat=reftable`

## Related Issues

- #14414 (original reftable issue, closed without fix)
- libgit2/libgit2#7117 (reftable support PR)

kfreitag1:
## Summary

- Remove the `git2` crate and its C dependencies (`libgit2-sys`, `libz-sys`), replacing all remaining usage with the git CLI and other previously used rust crates
- Replace `git2::Patch` in `buffer_diff` with `imara-diff` (used elsewhere in the codebase for word diffing)
- Replace `git2::Oid` with `[u8; 20]` and the `hex` crate
- Replace `git2::Repository` in `RealGitRepository` with stored paths and git CLI calls via the existing `GitBinary`
- Fixes a bug where linked worktree git dir events could cause the repository entry to be dropped

### Motivation

libgit2 does not support the [reftable](https://github.blog/2024-04-29-highlights-from-git-2-45/#preliminary-reftable-support) storage format that was introduced in git 2.45 (see  [libgit2#7117](https://github.com/libgit2/libgit2/pull/7117)), meaning that git operations in the app using these (like git status, branch names, diffs, e.g.) appeared broken for any repository using `--ref-format=reftable`. [Git 3 will use reftables by default](https://www.deployhq.com/blog/git-3-0-on-the-horizon-what-git-users-need-to-know-about-the-next-major-release), so this needs to be supported eventually.

Most operations used the git binary already, but there were still a handfull of stragglers using libgit2. By swapping out the remaining uses, we can remove the dependancy of libgit2 and eliminate ~30k lines of vendored C and the associated build complexity (cmake, libz, pkg-config), as well as ensure that all git operations go through similar codepaths.

Closes #46747
Closes https://github.com/zed-industries/zed/discussions/45702

fixes ZED-76X
fixes ZED-73N

### Notes

- `reload_index` is removed from the `GitRepository` trait since both implementations were no-ops (the CLI always reads from disk)
- `change_branch` is simplified to `git checkout <name>`, which natively handles local/remote branch resolution
- `load_index_text` and `load_committed_text` no longer explicitly filter symlinks (the old code returned `None` for symlinked entries)

## Test plan

- [x] `cargo test -p git` (34 tests)
- [x] `cargo test -p buffer_diff` (14 tests)
- [x] `cargo test -p worktree -- test_linked_worktree_git_dir_events_do_not_panic`
- [x] `cargo test -p util` (108 tests)
- [x] `cargo test -p project -- test_file_status test_repository_subfolder_git_status test_update_gitignore`
- [ ] Manual testing of diff gutter, staging, branch switching, remote operations
- [ ] Manual testing with a reftable repository

Release Notes:

- Fixed git integration not working with repositories using the reftable reference storage format


maxbrunsfeld (commented):
I haven't reviewed the code in detail, but I strongly approve of the overall goal here.

crates/git/src/repository.rs
```diff
@@ -1786,43 +1799,10 @@ impl GitRepository for RealGitRepository {
...
-            Ok(branch
-                .name()?
-                .context("cannot checkout anonymous branch")?
-                .to_string())
-        });
-
         self.executor
             .spawn(async move {
-                let branch = branch.await?;
-                git_binary?.run(&["checkout", &branch]).await?;
```
cole-miller, line 1825: This changes the behavior of `change_branch` in the case of a remote-tracking branch like `origin/feature`--we want to create a local branch named `feature` and set the upstream, but `git checkout origin/feature` will put the repo into a detached HEAD state. `git checkout feature` almost has the right behavior, except that it will successfully check out an existing branch named `feature` if one exists, instead of failing as we want it to. Maybe there is some git invocation that does what we want?

crates/worktree/tests/integration/main.rs
cole-miller, line 2886: Just confirming, this is an existing bug in our handling of worktrees when the commondir is outside the project root?

crates/worktree/src/worktree.rs
```diff
@@ -5235,6 +5235,8 @@ impl BackgroundScanner {
                         if SanitizedPath::new(repo.common_dir_abs_path.as_ref()) == dot_git_dir
                             || SanitizedPath::new(repo.repository_dir_abs_path.as_ref())
                                 == dot_git_dir
+                            || SanitizedPath::new(repo.dot_git_abs_path.as_ref())
+                                == dot_git_dir
```
cole-miller, line 5239: Can you say more about why this is needed? I think the intention at this point in the code is that `dot_git_dir` is a directory, so if `repository_dir_abs_path != dot_git_abs_path` because `dot_git_abs_path` is a file, we shouldn't be able to have `dot_git_abs_path == dot_git_dir`. But it sounds like there might be a real-life case where that's not true?

crates/worktree/src/worktree.rs
```diff
@@ -5294,7 +5289,7 @@ impl BackgroundScanner {
 
             if exists_in_snapshot
                 || matches!(
-                    self.fs.metadata(&entry.common_dir_abs_path).await,
+                    self.fs.metadata(&entry.dot_git_abs_path).await,
```
cole-miller, line 5292: This change makes sense to me, thanks for fixing and adding a regression test!

crates/git/src/repository.rs
```diff
@@ -1345,75 +1367,53 @@ impl GitRepository for RealGitRepository {
     }
 
     fn load_index_text(&self, path: RepoPath) -> BoxFuture<'_, Option<String>> {
-        // https://git-scm.com/book/en/v2/Git-Internals-Git-Objects
-        const GIT_MODE_SYMLINK: u32 = 0o120000;
-
-        let repo = self.repository.clone();
+        let git_binary = self.git_binary();
+        let path_str = format!(":{}", path.as_unix_str());
```
cole-miller, line 1371: For this and for `load_committed_text`, I think we should preserve the old behavior that essentially doesn't attach a diff to buffers that were opened from symlinks. But I think we can do it at a higher layer--when we are opening the uncommitted diff for a buffer, we should be able to look up a `worktree::Entry` for the buffer's `File` (if there is one) and determine from that whether we're looking at a symlink

cole-miller (dismissed):

crates/worktree/src/worktree.rs
kfreitag1: You're right I think the intention here is for `dot_git_dir` to be a directory, but it seems like `is_git_dir` returns true for any path with `.git` as its filename. So if there's an event for the .git file itself then we want to capture that, so should also match on `dot_git_abs_paths`. I remember running into this while testing in our monorepo, where I would get a panic from errors in the None branch being taken. It seems that the .git file was receiving a change event, but I'm not too sure entirely what was behind this. Adding in this other match resolves the panic, which is masked in the release

Perhaps the local var name `dot_git_dir` is a bit confusing then since this could be a file! I can change this to be a bit clearer.

crates/worktree/tests/integration/main.rs
kfreitag1: Yes it seems to be preexisting 👍  I can see "update_git_repositories called with .git directory outside the worktree root" errors in my Zed logs going back to March, from opening linked worktrees in our monorepo. I've also updated the regression test so it actually fails without the fix (the previous one didn't seem to work quite right)

Anthony-Eid (approved):

Issue #22373, opened 2024-12-23
Allow to hide the thought process of "thinking" models in the assistant panel


bnn1:
Summary: Allow to hide the thought process of "thinking" models like o1 in the assistant panel

Currently, when using thinking models like `gemini-2.0-flash-thinking-exp-1219`, the whole thought process is dumped on user along with the actual response to the user's query. 

Hiding thought process into collapsible section would be preferred when dealing with thinking models along with the indication that the model is "thinking". 

### Zed Version and System Specs

Zed: v0.166.2 (Zed)
OS: Linux X11 arcolinux unknown
Memory: 30.9 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics (ADL GT2) || Intel open-source Mesa driver || Mesa 24.3.2-arch1.1

### If applicable, add mockups / screenshots to help present your vision of the feature

Current behavior:
![image](https://github.com/user-attachments/assets/47fc9e47-5ad5-4c91-9962-7b7eaa6c5624)
![image](https://github.com/user-attachments/assets/08cf33cf-2b07-4f09-a8ec-8c02635fa5a7)

Desired behavior:
![image](https://github.com/user-attachments/assets/b62d72a0-975e-4282-ba5f-66c99627c0da)

bnn1:
@gyzerok The "Thinking process" collapsible section you observe on the "Desired behavior" screenshot was photoshoped, sorry for the confusion 😅

gyzerok:
@bnn1 yeah, I know. I've just reported my use case as well, so maintainers have more info to word with. It just so happened that your section name and mine appears to be very close

nathom:
This would be great since deepseek just released the `deepseek-reasoner` API!

RustoMCSpit:
> I am not using a "thinking" model, however am currently working on a prompt that would force model to think. Was expecting to be able to hide this thought process into collapsible section like on the screenshot, but it didn't work out.
> 
> Would love to see this use case supported as well.
> 
> ![image](https://github.com/user-attachments/assets/f7ecf2e3-efa0-4caf-a9a7-58366e1c357f)

it looks like Zed will need config options as there is no standardised tag. youre using <summary>, deepseek has <think>, etc.

lucascherzer:
FYI: The chat title also does not strip the thinking process.
It might also be worth considering to fall back on small models for title generation, as this task is rather simple in most cases and choosing small models for this task is more efficient.

[...]

PR #53488, opened 2026-04-09
Strip think tags from text stream

gh-keystr0ke:
Self-Review Checklist:

Closes #50989
Also related to #22373

Release Notes:

- Fixed models that emit `<think>` reasoning tags (e.g. DeepSeek, Qwen) leaking internal chain-of-thought into user-facing responses


gh-keystr0ke:
Thanks @osyvokon, you're right, I should have taken a deeper look into the problem before bringing that MR. Thanks for your time!

PR #53494, opened 2026-04-09
git_graph: Use ui_font_size to determine the row height

### LINKED ISSUE #53492: Git graph row height is fixed and does not scale with `ui_font_size`, leading to text clipping
### Reproduction steps

1. Open Zed settings and set "ui_font_size": 24 (or any large value).
2. Open git graph
3. Look at the rows in the "Description", "Date", and "Author" columns.

Image

**Current**:
The row height appears to be fixed. When the font size is increased, the text is vertically clipped at the bottom because the row container is too short.

**Expected behavior**:
The height of each row in the Git Graph should automatically adjust to accommodate the `ui_font_size` so that all text is fully visible.

Zed: v0.233.0+nightly.2382.754451598ec98b77bb5b02b35cd51efe60e2f82c (Zed Nightly 754451598ec98b77bb5b02b35cd51efe60e2f82c) 
OS: Windows 10.0.26200
Memory: 47.8 GiB
Architecture: x86_64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

lingyaochu:
Self-Review Checklist:

Closes #53492

After fix:
ui_font_size: 16
image

ui_font_size: 20
image

ui_font_size: 24
image

Release Notes:

- Fixed git graph row height not scaling with ui_font_size setting

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -930,7 +930,7 @@ impl GitGraph {
     fn row_height(cx: &App) -> Pixels {
         let settings = ThemeSettings::get_global(cx);
         let font_size = settings.ui_font_size(cx);
-        font_size + px(12.0)
+        font_size * (1.0 + 0.75)
     }
 
     fn graph_canvas_content_width(&self) -> Pixels {
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -929,7 +929,7 @@ impl GitGraph {
 
     fn row_height(cx: &App) -> Pixels {
         let settings = ThemeSettings::get_global(cx);
-        let font_size = settings.buffer_font_size(cx);
+        let font_size = settings.ui_font_size(cx);
         font_size + px(12.0)
     }
 
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -58,6 +58,7 @@ const LEFT_PADDING: Pixels = px(12.0);
 const LINE_WIDTH: Pixels = px(1.5);
 const RESIZE_HANDLE_WIDTH: f32 = 8.0;
 const COPIED_STATE_DURATION: Duration = Duration::from_secs(2);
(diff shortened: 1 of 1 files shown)


lingyaochu:
>Well, seems this fix also addresses https://github.com/zed-industries/zed/issues/53469, I need a futher check to see why they are related.

So silly of me, I have changed the code to use `ui_font_size`, of cause changing `buffer_font_size` won't affect the graph 😅. #53496 is related to other parts of code.

crates/git_graph/src/git_graph.rs
```diff
@@ -929,8 +929,8 @@ impl GitGraph {
 
     fn row_height(cx: &App) -> Pixels {
         let settings = ThemeSettings::get_global(cx);
-        let font_size = settings.buffer_font_size(cx);
-        font_size + px(12.0)
+        let font_size = settings.ui_font_size(cx);
+        font_size * (1.0 + 0.75)
```
yeskunall: Regardless of the fix, I’m not sure with the use of [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) in the codebase. Why was it 12px before? And now it’s using a scaling factor? And the scaling factor introduces _more_ magic numbers? 🤨

lingyaochu: I should have explained this earlier: these values—whether the original px(12.0) or the new 0.75 scaling factor—are intended to provide enough vertical breathing room for the git graph lines. Please see the illustration below:
compare

Without this extra spacing,  the characters are no longer fully visible: 
image
Which is definitely not good.

It is necessary to introduce a number here, but yes, we should avoid to make it as a magic number, maybe introduce a variable or constant with meaningful name is better.

Anthony-Eid:
I'm closing this because we're using magic numbers as @yeskunall pointed out (originally caused by yours truly, sorry about that)

A better solution would be to get the line height directly from the window text style 

```rust
let rem_size = window.rem_size();
let font_id = window.text_system().resolve_font(&style.text.font());
let font_size = style.text.font_size.to_pixels(rem_size);
let line_height = style.text.line_height_in_pixels(rem_size);
```

PR #53496, opened 2026-04-09
csv_preview: #2 Add settings UI panel with debug tools during development

HalavicH:
Self-Review Checklist:

------

This PR adds a settings panel above the CSV table with dropdown menus to control rendering behavior, and a performance metrics overlay for debugging.

Currently it's mostly used for me during dev phase, but before release of CSV preview feature all dev-only options will be cleaned up, and future features like "copy selected" will have their settings in this bar

image_2026-04-09_11-52-24

**What changed:**

- Adds `render_settings_panel()` method with dropdown menus for:
  - **Rendering Mode**: Variable Height (multiline support) vs Uniform Height (better performance)
  - **Text Alignment**: Top vs Center vertical alignment within cells
  - **Font Type**: UI Font vs Monospace for better readability (will be exposed to user settings)
  - **Experimental**: Popover menu with toggles for debug features
- Adds `render_performance_metrics_overlay()` method showing:
  - CSV parsing duration
  - Rendered row count and indices
  - Positioned in bottom-right corner with semi-transparent styling

Context:
Will iterate on it with @Anthony-Eid 

Release Notes:

- N/A


Anthony-Eid:
I'm not sure how I feel about exposing some of these things as a setting. e.g. variable list vs uniform list, we should always use variable list instead of supporting two separate code paths. If performance becomes a concern we could make a uniform_variance_list as the engine (I want this anyway)

Font Type, we should stick with UI font for labels to be inline with Zed,

text alignment and debug features should probably be in memory values? We should talk about this

Anthony-Eid (approved):

PR #18571, opened 2024-10-01
windows: Revert "Fix `hide`, `activate` method on Windows to hide/show application"

JunkuiZhang:
This PR reverts the changes introduced via #18164. As shown in the video below, once you `hide` the app, there is essentially no way to bring it back. I must emphasize that the window logic on Windows is entirely different from macOS. On macOS, when you `hide` an app, its icon always remains visible in the dock, and you can always bring the hidden app back by clicking that icon. However, on Windows, there is no such mechanism—the app is literally hidden. 

I think the `hide` feature should be macOS-only.

https://github.com/user-attachments/assets/65c8a007-eedb-4444-9499-787b50f2d1e9

Release Notes:

- N/A

PR #18164, opened 2024-09-20
gpui: Fix `hide`, `activate` method on Windows to hide/show application

huacnlee:
Release Notes:

- N/A

Continue #18161 to fix `cx.hide`, `cx.activate` method on Windows to hide/show application.

## After

https://github.com/user-attachments/assets/fe0070f9-7844-4c2a-b859-3e22ee4b8d22

PR #53499, opened 2026-04-09
gpui: Add `hide()` and `activate()` implementations for Windows

bowenxuuu:
Self-Review Checklist:

fixed: 
#52828 
#48428 

**Before:**
The methods `cx.hide()` and `cx.activate()` do nothing.

**After:**
The method `cx.hide()` will hide the window.The method `cx.activate()` will show the window if it has been hidden.


AzureZee:
PR (#18164) already implemented it, but PR (#18571) reverted it. What will happen this time? :(

bowenxuuu:
ref to #18571
> On macOS, when you hide an app, its icon always remains visible in the dock, and you can always bring the hidden app back by clicking that icon. However, on Windows, there is no such mechanism—the app is literally hidden.

On Windows, if you want to hide an app while keeping its icon visible in the taskbar, you should use the `window.minimize()` method instead of the `cx.hide()` method. Therefore, I think implementing the `cx.hide()` method on Windows is reasonable.

unsecretised:
It seems that `cx.activate(true)` also does nothing in macos .. unsure if that could also be added

unsecretised:
Whats the status on this. Would say its been quite a while since this was created.

PR #53504, opened 2026-04-09
editor: Add configurable hover delay 

daydalek:
follow up #47471 

As described in #47471, we introduced a direction-aware strategy to improve user experience when interacting with hover popovers. In this follow-up, we are adding `hover_popover_sticky` and `hover_popover_hiding_delay` to control whether the feature introduced in 47471 enabled, and to let users configure the delay to balance responsiveness . Also `hover_popover_sticky` can now be imported from `editor.hover.sticky`, as well as `hover_popover_hiding_delay` from `editor.hover.hidingDelay` in VSCode.

Also this PR adds several tests:
- `test_hover_popover_cancel_hide_on_rehover`: when the cursor returns to the hover after leaving once within the hiding delay, the hover should persist while canceling the existing hiding timer.
- `test_hover_popover_enabled_false_ignores_sticky` : when `hover_popover_enabled` is false, the `hover_popover_sticky` and `hover_popover_hiding_delay` have no effect(since no hover is shown).
- `test_hover_popover_sticky_delay_restarts_when_mouse_gets_closer`: when mouse gets closer to hover popover, we expect the timer to reset and the hover remains visible.
- `test_hover_popover_hiding_delay`: check if the delay(in test, that's 500ms) works.
- `test_hover_popover_sticky_disabled`: when hover_popover_sticky is false, the hover popover disappears immediately after the cursor leaving the codes.
- VSCode import test in `settings_store.rs`

Release Notes:

- Added `hover_popover_sticky` and `hover_popover_hiding_delay` settings to balance responsiveness of hover popovers.


SomeoneToIgnore (approved):
Great follow-up, thank you a lot!

PR #53515, opened 2026-04-09
compliance: Reduce noisiness when checks were successful

MrSubidubi:
This will ensure we do not post a second Slack message in case the first check was successful. We still _run_ the second check, but do not notify

Release Notes:

- N/A


ConradIrwin (approved):
nice

PR #53535, opened 2026-04-09
Add Parallel Agents documentation

macraig:
Adds documentation for the Parallel Agents feature.

## Changes

- **New page** `docs/src/ai/parallel-agents.md`: covers the Threads Sidebar component, switching threads, the archive and search, importing ACP threads, running multiple threads, multiple projects, worktree isolation, and the default layout change
- **`docs/src/ai/overview.md`**: rewrites the Agentic editing section to lead with the Threads Sidebar, adds Parallel Agents to the Getting started links
- **`docs/src/ai/agent-panel.md`**: adds a cross-link to the Threads Sidebar in Creating New Threads, removes the Navigating History section (now covered in the Threads Sidebar docs)
- **`docs/src/SUMMARY.md`**: adds Parallel Agents between External Agents and Inline Assistant

Release Notes:

- N/A


docs/src/ai/parallel-agents.md
```diff
@@ -0,0 +1,66 @@
...
+
+In the Agent Panel toolbar, click the worktree selector and choose **New Git Worktree**. You can also cycle between options with {#kb agent::CycleStartThreadIn}. When you send the first message, Zed creates a new worktree from the current branch.
+
+After the agent finishes, review the diff, merge the changes through your normal git workflow, and delete the worktree when done.
+
+> **Note:** Starting a thread in a new worktree requires the project to be in a git repository.
+
+## Default Layout {#layout}
+
+New installs place the Agent Panel and Threads Sidebar on the left. The Project Panel, Git Panel, and other panels move to the right, keeping the thread list and conversation next to each other. To rearrange panels, right-click any panel icon in the status bar.
```
katie-z-geer: @macraig let's be more specific here like "On versions 0.232.0 and beyond, this is what happens". The word new ages so fast haha

docs/src/ai/agent-panel.md
```diff
@@ -36,6 +36,8 @@ By default, the Agent Panel uses Zed's first-party agent.
 
 To choose another agent, go to the plus button in the top-right of the Agent Panel and pick one of the [external agents](./external-agents.md) installed out of the box.
 
+Manage multiple threads at once with the [Threads Sidebar](./parallel-agents.md#threads-sidebar).
```
danilo-leal: I think we can afford a section in the agent panel page talking about multiple threads and managing them; something that gives a bit of an introduction to the threads sidebar. It's okay from my pov to have a bit of duplicated content here between this page and the sidebar page.

docs/src/ai/overview.md
```diff
@@ -18,7 +18,9 @@ Zed's AI features run inside a native, GPU-accelerated application built in Rust
 
 ## Agentic editing
 
-The [Agent Panel](./agent-panel.md) is where you work with AI agents. Agents can read files, edit code, run terminal commands, search the web, and access diagnostics through [built-in tools](./tools.md).
+The [Threads Sidebar](./parallel-agents.md#threads-sidebar) is where you organize agent work. Start a thread, give it a task, and the agent reads, edits, and runs code through [built-in tools](./tools.md). You can run multiple threads at once, each using a different agent and working against different projects.
```
danilo-leal: Maybe we can drop the "built-in tools" bit because that's only valid for the Zed agent. Or re-phrase this somehow so as to consider external agents as well.

docs/src/ai/parallel-agents.md
```diff
@@ -0,0 +1,74 @@
...
+
+The archive holds all your threads. Toggle the archive with {#kb agents_sidebar::ToggleArchive} or by clicking the archive icon in the sidebar bottom bar.
+
+To archive a thread, hover over it in the sidebar and click the archive icon that appears. You can also select a thread and press {#kb agent::RemoveSelectedThread}. Running threads cannot be archived until they finish.
+
+To restore an archived thread, open the archive and click the thread you want to bring back. Zed unarchives it, moves it back to the thread list, and opens it in the Agent Panel. If the thread was running in a git worktree that was deleted when archived, Zed restores the worktree automatically.
+
+To permanently delete a thread, open the archive, hover over the thread, and click the trash icon. This removes the thread's conversation history and cleans up any associated worktree data. Deleted threads cannot be recovered.
+
+You can search your archived threads; search will fuzzy match on thread titles.
```
danilo-leal: This will need some updating as we're changing this view to be more like a "history" view. Also, little nit, but we should capitalize Git.

docs/src/ai/parallel-agents.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Importing External Agent Threads {#importing-threads}
+
+When the archive is open, an **Import ACP Threads** button appears in the sidebar bottom bar. Click it to open the import modal, which lists your installed external agents. Select the agents whose threads you want to import and confirm. Zed finds threads from those agents, whether they were started in Zed or in another client, and adds them to your thread archive.
```
danilo-leal: I think we could to update this a bit, saying something like:

> If you have [external agents](link) installed, Zed will detect whether you have existing threads and invite you to import them into Zed. Every time you open the history view, you should see an icon button that allows you to import threads at any time. Clicking on it will open a modal, where you can select the agents whose threads you want to import.

docs/src/ai/parallel-agents.md
```diff
@@ -0,0 +1,74 @@
...
+
+If two threads might edit the same files, start one in a new git worktree to give it an isolated checkout.
+
+In the Agent Panel toolbar, click the worktree selector to choose which worktree you want the agent to run in, or create a new one. You can also cycle between the current worktree and a new worktree with {#kb agent::CycleStartThreadIn}.
+
+To change the default for new threads, set `new_thread_location` in your agent settings to `"local_project"` or `"new_worktree"`.
+
+After the agent finishes, review the diff and merge the changes through your normal git workflow. If the thread was running in a linked worktree and no other active threads use it, archiving the thread saves the worktree's git state and removes it from disk. Restoring the thread from the archive restores the worktree.
+
+> **Note:** Starting a thread in a new worktree requires the project to be in a git repository.
```
danilo-leal: This section will also need some updating as per the latest changes in the worktree creation UX. The cycling thing, for example, doesn't exist anymore, and neither does the `new_thread_location` setting. We need to mention the detached HEAD state in here, too. And lastly, we could maybe skip the note about worktrees requiring a Git repository; I think that's implied!

docs/src/ai/parallel-agents.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Default Layout {#layout}
+
+From version 0.233.0 onward, the Agent Panel and Threads Sidebar are on the left. The Project Panel, Git Panel, and other panels move to the right, keeping the thread list and conversation next to each other. To rearrange panels, right-click any panel icon.
```
danilo-leal: We should probably look for other places in the docs where we mention the new default layout, too. Being at the bottom of the parallel agents page makes it feel like a little detail, I feel like...

danilo-leal (commented):
Just some quick notes!

docs/src/ai/overview.md
macraig: good catch!

docs/src/ai/parallel-agents.md
macraig: I really like that version, will update

docs/src/ai/parallel-agents.md
macraig: Just to confirm, will we refer to the archive as History View moving forward?

docs/src/ai/parallel-agents.md
macraig: Updating! Do you think it makes sense to have this in the parallel agents page instead of the Agent panel one? If I think of it from a value prop perspective I think the answer is yes, although we could make a case for the Agent panel since the worktree dropdown lives in it 🤔

docs/src/ai/parallel-agents.md
macraig: Do you think it's something we should call out in the docs? I've gone back and forth on it and can't decide if it's worth calling out or not

docs/src/ai/parallel-agents.md
danilo-leal: I think so; "Threads History" is the label we'll have in the app.

docs/src/ai/parallel-agents.md
danilo-leal: I think so as well! We could have a brief section in the agent panel about it, or at least a mention, and the cross link it! I don't think it's bad to touch on similar topics on multiple pages 😀

docs/src/ai/parallel-agents.md
danilo-leal: Yeah, that's a good question... maybe a note like you have in some other place, too, will be good, because it's a change people will notice one way or another and we should probably have a line that refers to it, formalizing it a bit.

macraig: A note feels much better than a section, great idea thanks!

maxbrunsfeld (approved):

docs/src/tasks.md
```diff
@@ -229,6 +230,31 @@ This could be useful for launching a terminal application that you want to use i
 }
 ```
 
+## Hooks
+
+In addition to being spawned manually, tasks can be configured to run automatically in response to certain Zed events by adding one or more hooks to the `hooks` field on a task template. A task with a matching hook will be resolved and spawned when that event fires.
```
Anthony-Eid: We currently only support one hook so this isn't fully accurate (yet). Not sure if we should change the wording or leave it

docs/src/ai/parallel-agents.md
```diff
@@ -0,0 +1,75 @@
...
+Open the Threads Sidebar with {#kb multi_workspace::ToggleWorkspaceSidebar}.
+
+> **Note:** From version 0.233.0 onward, the Agent Panel and Threads Sidebar are on the left by default. The Project Panel, Git Panel, and other panels move to the right, keeping the thread list and conversation next to each other. To rearrange panels, right-click any panel icon.
+
+## Threads Sidebar {#threads-sidebar}
+
+The sidebar shows your threads grouped by project. Each project gets its own section with a header. Threads appear below with their title, status indicator, and which agent is running them.
+
+To focus the sidebar without toggling it, use {#kb multi_workspace::FocusWorkspaceSidebar}. To search your threads, press {#kb agents_sidebar::FocusSidebarFilter} while the sidebar is focused.
+
```
Anthony-Eid, line 19: I wonder if we should explicitly say that linked git worktrees are considered part of the same project (are  under the same project header) in the sidebar here as well.

docs/src/tasks.md
macraig:
```suggestion
In addition to being spawned manually, tasks can be configured to run automatically in response to certain Zed events by adding a hook to the `hooks` field on a task template. A task with a matching hook will be resolved and spawned when that event fires.
```

Updated to only refer to one for now

docs/src/ai/parallel-agents.md
macraig: Added a sentence to clarify it

docs/src/ai/agent-panel.md
```diff
@@ -34,7 +34,27 @@ The sections below cover what you can do from here.
...
 By default, the Agent Panel uses Zed's first-party agent.
 
-To choose another agent, go to the plus button in the top-right of the Agent Panel and pick one of the [external agents](./external-agents.md) installed out of the box.
+Start a new thread with {#kb agent::NewThread}, or open the "New Thread…" menu via the `+` icon in the top-right of the panel toolbar (in the empty state, this menu is exposed as the agent selector button on the left). You can also open that menu with {#kb agent::ToggleNewThreadMenu}.
+
+From the "New Thread…" menu you can:
+
+- Pick **Zed Agent** or any installed [external agent](./external-agents.md) to start a new thread with that agent.
+- Choose **New From Summary** to start a fresh Zed Agent thread seeded with a summary of the current conversation — useful for compacting long threads as you approach the context window limit.
+- Click **Add More Agents** to install additional external agents.
```
danilo-leal:
```suggestion

```

We could potentially remove this bit here; it does more faithfully describe the UI, but it's also self-explanatory 😄

docs/src/ai/agent-panel.md
```diff
@@ -34,7 +34,27 @@ The sections below cover what you can do from here.
...
+
+When an external-agent thread is focused, {#action agent::NewExternalAgentThread} creates another thread with the same external agent.
```
danilo-leal: We should be able to say here that hitting this keybinding will always create a new thread with the currently selected agent. Not a particular thing for external agents-only.

docs/src/ai/agent-panel.md
```diff
@@ -34,7 +34,27 @@ The sections below cover what you can do from here.
...
+
+You can also start a new thread from the [Threads Sidebar](./parallel-agents.md#threads-sidebar), scoped to a specific project — see [Running Multiple Threads](./parallel-agents.md#running-multiple-threads).
+
+### Managing Multiple Threads {#multiple-threads}
+
+You can run multiple agent threads at once, each working independently with its own agent, context window, and conversation history. Open the Threads Sidebar with {#kb multi_workspace::ToggleWorkspaceSidebar} to see all your threads grouped by project. Click any thread to switch to it, or use the thread switcher ({#kb agents_sidebar::ToggleThreadSwitcher}) to cycle between recent threads without opening the sidebar.
+
+Threads you're no longer working on can be archived by hovering over them in the sidebar and clicking the archive icon, or selecting them and pressing {#kb agent::ArchiveSelectedThread}. Thread History holds all your threads and you can restore them at any time.
```
danilo-leal:
```suggestion
Threads you're no longer working on can be archived by hovering over them in the sidebar and clicking the archive icon, or selecting them and pressing {#kb agent::ArchiveSelectedThread}. The Thread History holds all your threads across all projects, sorted chronologically, and you can restore them at any time.
```

Just enhancing this a little bit

docs/src/tasks.md
```diff
@@ -229,6 +230,31 @@ This could be useful for launching a terminal application that you want to use i
 }
 ```
 
+## Hooks
+
+In addition to being spawned manually, tasks can be configured to run automatically in response to certain Zed events by adding a hook to the `hooks` field on a task template. A task with a matching hook will be resolved and spawned when that event fires.
+
+The following hooks are currently supported:
+
+- `create_worktree` — runs after Zed creates a new linked git worktree (for example, via the agent's worktree workflow). The task is spawned with `ZED_WORKTREE_ROOT` pointing at the newly created worktree and `ZED_MAIN_GIT_WORKTREE` pointing at the original repository's working directory, which makes these hooks well suited to copying untracked files (such as `.env` files) or running per-worktree setup commands.
```
danilo-leal:
```suggestion
- `create_worktree` — runs after Zed creates a new linked Git worktree, either directly through the CLI or through the UI with the worktree modal. The task is spawned with `ZED_WORKTREE_ROOT` pointing at the newly created worktree and `ZED_MAIN_GIT_WORKTREE` pointing at the original repository's working directory, which makes these hooks well-suited to copying untracked files (such as `.env` files) or running per-worktree setup commands.
```

Quick tweaks here as there isn't anymore a "agent's worktree workflow"; it's a general Zed feature now.

danilo-leal (approved):
Little tweaks only but looking great! :shipit:

Anthony-Eid (approved):

docs/src/ai/agent-panel.md
macraig: Good catch, updating!

PR #53539, opened 2026-04-09
project_panel: Always open panel on pane's reveal in project panel

dinocosta:
Update the way `pane::RevealInProjectPanel` is handled to ensure that, regardless of whether the file belongs to any open project, the project panel is always activated and focused.

This refactor is a result of some internal feedback after changing its handling so as to show a notification stating that the item that the user was trying to reveal didn't belong to an open project – https://github.com/zed-industries/zed/pull/51246 .

We feel users are probably already used to relying on `cmd-shift-e` (on macOS), in pretty much every context, in order to open the project panel, and so having situations where it doesn't actually happen seems like a bad user experience.

Self-Review Checklist:

Relates to #23967 

Release Notes:

- Improved `pane: reveal in project panel` to open the project panel, even if working with an unsaved buffer.

### DIFF
--- a/crates/project_panel/src/project_panel_tests.rs
+++ b/crates/project_panel/src/project_panel_tests.rs
@@ -6016,7 +6016,7 @@ async fn test_explicit_reveal(cx: &mut gpui::TestAppContext) {
 }
 
 #[gpui::test]
-async fn test_reveal_in_project_panel_notifications(cx: &mut gpui::TestAppContext) {
+async fn test_reveal_in_project_panel_fallback(cx: &mut gpui::TestAppContext) {
     init_test_with_editor(cx);
     let fs = FakeFs::new(cx.background_executor.clone());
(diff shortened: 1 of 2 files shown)


maxbrunsfeld (approved):

injust:
Does this need to be cherry-picked to stable? The stable release notes mention this PR, but it isn't part of the release.

Ideally, this should've landed in the same release as https://github.com/zed-industries/zed/pull/51246.

dinocosta:
> Does this need to be cherry-picked to stable? The stable release notes mention this PR, but it isn't part of the release.
> 
> Ideally, this should've landed in the same release as #51246.

Definitely! Thank you for calling that out, I had a note to cherry pick to preview right after merging but somehow forgot to do that 🤦

Will coordinate with the team to see if it makes sense to do a patch release including this, and likely other changes. Thanks!

PR #53547, opened 2026-04-09
agent_ui: Simplify thread scrolling with keyboard

osyvokon:
This change allows using simple navigation keys (PageDown / PageUp / Ctrl-Home / Ctrl-End) when the message editor is focused, but only if the cursor is at the beginning/end of the message, where pressing these keys would normally result in no-op.

One important corollary is that when the cursor is in an empty message, navigation keys scroll the thread.

We already have this behavior for Up/Down and this change just expands it for other navigation keys.

Demo:

[Demo](https://github.com/user-attachments/assets/ff540c8c-a223-417b-b16a-b0d08599b1ae)

Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -80,7 +80,7 @@ jobs:
 
           # If assets/ changed, add crates that depend on those assets
           if echo "$CHANGED_FILES" | grep -qP '^assets/'; then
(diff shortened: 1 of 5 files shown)


danilo-leal:
@zed-zippy approve

PR #53548, opened 2026-04-09
Scope search filters to specific directories with glob patterns

boobachad:
**context:**
When using "Find in Folder" in a single-worktree workspace, the include filter was set to a bare folder name (e.g. src). This caused the search to match any directory with that name anywhere in the project tree(eg node_modules/.../src/) because the path matcher's ends_with check would match ancestor directories with the same name when walking up the file tree.
Image

**the fix**
Appending  /** to the filter path, anchors it to the exact selected directory so only files under that specific folder are included.
and also added the tests.
image

Self-Review Checklist:

Closes #53495.

Release Notes:

- Fixed "Find in Folder" incorrectly including results from directories with the same name elsewhere in the project (e.g. node_modules/.../src when searching src/)

### DIFF
--- a/crates/search/src/project_search.rs
+++ b/crates/search/src/project_search.rs
@@ -3839,8 +3839,8 @@ pub mod tests {
                     search_view.included_files_editor.update(cx, |editor, cx| {
                         assert_eq!(
                             editor.display_text(cx),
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
Autoclosing all draft PRs with no activity for 2 weeks or over.

PR #53550, opened 2026-04-09
agent: Support remote connection args in thread metadata database

Anthony-Eid:
This PR adds remote connection data to the threads metadata database. This fixes an issue where threads ran on separate projects with the same remote/local path list would show up in the sidebar in both workspaces, instead of only the workspace they were originally created in.

I added a migrator that uses the workspace persistence database to add remote connection argument to threads that only have path list matches with a remote project. If a path list matches with both local/remote workspaces, we default to setting it as local. 

Self-Review Checklist:

Release Notes:

### DIFF
--- a/crates/agent_ui/src/thread_import.rs
+++ b/crates/agent_ui/src/thread_import.rs
@@ -12,6 +12,7 @@ use gpui::{
 };
 use notifications::status_toast::{StatusToast, ToastIcon};
 use project::{AgentId, AgentRegistryStore, AgentServerStore};
+use remote::RemoteConnectionOptions;
 use ui::{
     Checkbox, KeyBinding, ListItem, ListItemSpacing, Modal, ModalFooter, ModalHeader, Section,
     prelude::*,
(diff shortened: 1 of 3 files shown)


crates/agent_ui/src/thread_import.rs
```diff
@@ -479,11 +492,15 @@ fn find_threads_to_import(
 }
 
 fn collect_importable_threads(
-    sessions_by_agent: Vec<(AgentId, Vec<acp_thread::AgentSessionInfo>)>,
+    sessions_by_agent: Vec<(
+        AgentId,
+        Option<RemoteConnectionOptions>,
+        Vec<acp_thread::AgentSessionInfo>,
+    )>,
```
mikayla-maki: Can we introduce a struct for this?

Anthony-Eid: Yes!

mikayla-maki (approved):

PR #53551, opened 2026-04-09
project: Add fuzzy search in remote projects dialogue

feitreim:
This PR primarily implements fuzzy search in the remote projects dialogue, I was encouraged to use the new `fuzzy_nucleo` crate for that so I also switched the non-remote open recent project modal, which already used a fuzzy search, over to the `fuzzy_nucleo` version. This also involved implementing the string matching for fuzzy search in `fuzzy_nucleo`, as before it was only working on specific matching paths.

## fuzzy_nucleo

Also in this PR is a performance improvement to the `fuzzy_nucleo` matching, borrowed from the `fuzzy` implementation we can use the same CharBag prefilter to entirely avoid matching on strings that lack the characters in the query. Secondary to that we can use a binary search instead of contains inside the matching hotloop, where we convert utf32 char offsets back to their original coordinate system (`‎crates/fuzzy_nucleo/src/strings.rs‎` line 290)

effect of the CharBag + binary_search on path matching:

| Benchmark | Size | Nucleo (before) | Nucleo (after) | Fuzzy | Before/Fuzzy | After/Fuzzy |
  |-----------|-----:|----------------:|---------------:|------:|-------------:|------------:|
  | 1-word | 100 | 14.14 µs | 9.12 µs | 9.06 µs | 1.56x | 1.01x |
  | 1-word | 1,000 | 164.37 µs | 114.11 µs | 110.43 µs | 1.49x | 1.03x |
  | 1-word | 10,000 | 1.83 ms | 1.39 ms | 1.41 ms | 1.30x | 0.99x |
  | 2-word | 100 | 12.83 µs | 3.51 µs | 979 ns | 13.10x | 3.59x |
  | 2-word | 1,000 | 131.65 µs | 33.46 µs | 6.37 µs | 20.67x | 5.25x |
  | 2-word | 10,000 | 1.24 ms | 338.84 µs | 52.46 µs | 23.64x | 6.46x |o

## remote project search

The reviews of https://github.com/zed-industries/zed/pull/51197 seemed to indicate that performance was important here so it runs async and cancels the previous stale request when a new character is typed.

other features:
text highlighting of what was being typed, similar to other fuzzy searches.
reorder the projects on the list based on the maximum score from its projects.

video of it in action:

https://github.com/user-attachments/assets/86b3ed85-fc29-46fb-b3c4-c0cbc3bf8edd

Self-Review Checklist:

Closes #53318 

Release Notes:

- recent_projects: Added fuzzy search to the remote projects modal.


feitreim:
@yara-blue can you check out my changes to fuzzy_nucleo :)

feitreim:
huh. i didn't know there was a typo checking hook, fixing that rn

feitreim:
okay I don't think that failure is related to my code? as best me and claude can tell at least.

yara-blue:
> okay I don't think that failure is related to my code? as best me and claude can tell at least.

correct :+1:, don't worry about it

crates/fuzzy_nucleo/src/strings.rs
```diff
@@ -0,0 +1,495 @@
...
+    if cancel_flag.load(atomic::Ordering::Acquire) {
+        return Vec::new();
+    }
+
+    let mut results = segment_results.concat();
+    util::truncate_to_bottom_n_sorted_by(&mut results, max_results, &|a, b| b.cmp(a));
+    results
+}
+
+pub fn match_strings_sync<T>(
```
yara-blue: Nit: I'd flip the names (match_strings_async and match_strings).

crates/fuzzy_nucleo/src/strings.rs
```diff
@@ -0,0 +1,495 @@
...
+        .map(|candidate| StringMatch {
+            candidate_id: candidate.borrow().id,
+            score: 0.,
+            positions: Default::default(),
+            string: candidate.borrow().string.clone(),
+        })
+        .collect()
+}
+
+fn match_string_candidates<T>(
```
yara-blue: lets use the word helper in the name like in `fuzzy_nucleo/src/paths`, that clearly signals the role of the function. So this would be match_strings_helper.

yara-blue (changes requested):
I'd love to merge it as it looks really thorough unfortunately there are two concerns preventing a merge:
- the PR does too many separate things.
- `fuzzy_nucleo::strings` is not tested with the existing pickers.

#### PR doing too many things is a problem
Let me explain. Imagine in a few weeks we find a pretty bad bug in the fuzzy search of the remote projects dialogue. We will then have to revert this entire PR until we've fixed the bug. But that means we also revert the optimizations you did, the benchmarks and `fuzzy_nucleo::strings` if anyone in the mean time has build upon either of those all that work also needs to be reverted. You can imagine this can get out of hand pretty quickly. 

Could you therefore split the PRs into three?

- [A] the optimization for the current `fuzzy_nucleo` and its benchmarks (amazing work btw!).
- [B] the introduction of `fuzzy_nucleo::strings`.
- [C] the use of `fuzzy_nucleo::strings` in the remote project search.

#### Not tested with existed pickers
When we tried to add fuzzy nucleo in november (see https://github.com/zed-industries/zed/pull/37123/changes) a lot of picker tests broke (most tests for fuzzy finding turn out to live in the pickers).

Could you after splitting the PR into A, B and C extend B (the introduction of `fuzzy_nucleo::strings`) to include the use of `fuzzy_nucleo::strings` in a few pickers? That should really increase the testing area.

feitreim:
@yara-blue okay that makes sense, the potential need to split it up into multiple PRs was definitely on my mind. It isn't super clear to me how to actually do that though? i know true stacked PRs are a bit of a missing feature from github but I assume there is some way that it is generally done. I'll experiment with it and see where that gets me.

yara-blue:
Closing as this has been successfully split into 3 independent PRs and is backlinked for context.

PR #53433, opened 2026-04-08
ci: Fix artifact naming conflict

MrSubidubi:
This fixes the issue we saw in https://github.com/zed-industries/zed/actions/runs/24147949240

Release Notes:

- N/A

PR #53559, opened 2026-04-09
ci: Use GitHub context for artifact name

MrSubidubi:
Missed this yesterday in https://github.com/zed-industries/zed/pull/53433, because in the `with` level, we need to use the proper context syntax.

Release Notes:

- N/A


ConradIrwin (approved):
🎫

PR #53560, opened 2026-04-09
acp: Use npm --prefix for registry npx agents

benbrandt:
Because we weren't going through the normal npm subcommand route, we
weren't getting a prefix flag applied. Which meant some users were
seeing errors of incorrect package managers when used with a JS project.

We still want the agent to run in the project dir if we can, we just don't want it resolving packages from that project. (This actually means I can run it in our claude-agent-acp repo again :D )

I refactored the node_runtime functions a bit to make these subcommand functions a bit more inline wiht each other and actually fixed an issue where --prefix might have been added after `--` previously which wouldn't be correct anyway.

Self-Review Checklist:

Release Notes:

- acp: Fix npm-based agents resolving based on current project.


crates/node_runtime/src/node_runtime.rs
```diff
@@ -667,17 +676,11 @@ impl SystemNodeRuntime {
...
+        Ok(Self {
             node,
             npm,
-            global_node_modules: PathBuf::default(),
             scratch_dir,
-        };
-        let output = this.run_npm_subcommand(None, None, "root", &["-g"]).await?;
-        this.global_node_modules =
-            PathBuf::from(String::from_utf8_lossy(&output.stdout).to_string());
-
```
benbrandt, line 679: Realized this was dead code, so we can skip a command run here

osyvokon (approved):
LGTM

crates/project/src/agent_server_store.rs
```diff
@@ -1542,11 +1545,18 @@ impl ExternalAgentServer for LocalRegistryNpxAgent {
                 .await
                 .unwrap_or_default();
 
+            let prefix_dir = paths::external_agents_dir()
```
morgankrey, line 1551: The package string comes from a remote registry and is used directly in a path. If a malicious registry served a package name like ../../../tmp/evil, it could create directories outside external_agents_dir. Consider validating that package doesn't contain .., /, or \ before using it in a path, or canonicalize and verify the result stays within the expected directory.

Low severity since this requires compromising the HTTPS registry, but worth hardening for defense-in-depth.

`if package.contains("..") || package.contains('/') || package.contains('\\') {
      bail!("invalid package name");
  }`

PR #53571, opened 2026-04-09
Stop eagerly clearing available code actions on row change

Dnreikronos:
Self-Review Checklist:

Part of #53210

## Root cause

In `selections_did_change()` (`editor.rs`), when the cursor moved to a different row, `available_code_actions` was cleared instantly via `.take()`. But `refresh_code_actions()`, called right after, debounces for 250ms before querying the LSP. During that gap the toolbar re-rendered, saw `None`, and disabled the button. Once the LSP responded, the button re-enabled. That back-and-forth is the visible flicker.

```rust
// before (editor.rs, selections_did_change):
if old_cursor_position.to_display_point(&display_map).row()
    != new_cursor_position.to_display_point(&display_map).row()
{
    self.available_code_actions.take();  // instantly clears
}
self.refresh_code_actions(window, cx);   // 250ms debounce before LSP request
```

## Fix

Removed the eager `.take()`. Now the previous code actions stay visible until the debounced refresh completes and replaces them with fresh results (or `None` if no actions apply at the new position). `refresh_code_actions()` already does a full replacement at the end, so there's no stale accumulation.

### Before:

https://github.com/user-attachments/assets/33d796d1-f33d-473e-bcd4-14b848e655db

### After:

https://github.com/user-attachments/assets/94b0b1bc-2386-4116-aed0-9180a881c123

## Tested scenarios

- Find + Enter to jump between search results
- F3 / Ctrl+Alt+G (SelectNextMatch) with and without search bar visible
- Clicking between titlebar and editor
- Rapid Enter spam on Find matches
- Arrow keys held down across rows
- Moving from a line with code actions to one without

Release Notes:

- Fixed code actions toolbar button flickering on cursor movement


SomeoneToIgnore (commented):
Thank you, I am somewhat concerned by this change: given that there's a number of code places that try to access the actions, we might (at least in theory) end up in a situation when stale actions are fetched for new selections?

That is not right and might break a thing or two, so I wonder if we can add some enum instead of `Option` here.

So, instead of `Option<(Location, Rc<[AvailableCodeAction]>)>` we would have a 

```rs
enum FindAGoodName {
    AndGoodNameHere {
        foo: Location,
        bar: Rc<[AvailableCodeAction]>,
    },
    Fetching(Task<Result<()>>),
    None, // Is it needed? I guess so, if we get no actions
}
```

-esque structure to make such situation impossible, remove the `code_actions_task` field and have an only `available_code_actions: FindAGoodName` field instead.

Then, the renderer should never look disabled in case if we fetch still?

Dnreikronos:
> Thank you, I am somewhat concerned by this change: given that there's a number of code places that try to access the actions, we might (at least in theory) end up in a situation when stale actions are fetched for new selections?
> 
> That is not right and might break a thing or two, so I wonder if we can add some enum instead of `Option` here.
> 
> So, instead of `Option<(Location, Rc<[AvailableCodeAction]>)>` we would have a
> 
> ```rust
> enum FindAGoodName {
>     AndGoodNameHere {
>         foo: Location,
>         bar: Rc<[AvailableCodeAction]>,
>     },
>     Fetching(Task<Result<()>>),
>     None, // Is it needed? I guess so, if we get no actions
> }
> ```
> 
> -esque structure to make such situation impossible, remove the `code_actions_task` field and have an only `available_code_actions: FindAGoodName` field instead.
> 
> Then, the renderer should never look disabled in case if we fetch still?

Thanks for the review and the suggestion, really appreciate it.

You were right that the two separate fields were asking for trouble. `available_code_actions` as an `Option` and `code_actions_task` off to the side meant nothing stopped stale actions from leaking into new selections.

So I went with your enum idea. `AvailableCodeActions` has three variants: `None`, `Fetching` (keeps previous actions around for display while the new fetch runs), and `Ready`. The toolbar button no longer flickers since `Fetching` still has something to show.

I pulled the "steal the task out of Fetching" logic into a `take_task()` helper since it came up twice.

One thing I added beyond your sketch: a `code_actions_generation` counter. I hit a race where moving the cursor mid-fetch let the old task's completion handler overwrite the newer `Fetching` and drop its task. The counter lets each handler check whether it's still current before writing.

crates/editor/src/editor.rs
```diff
@@ -7355,9 +7390,15 @@ impl Editor {
...
 
     pub fn has_available_code_actions(&self) -> bool {
-        self.available_code_actions
-            .as_ref()
-            .is_some_and(|(_, actions)| !actions.is_empty())
+        match &self.available_code_actions {
+            AvailableCodeActions::Ready { actions, .. } => !actions.is_empty(),
+            AvailableCodeActions::Fetching {
+                previous_actions, ..
+            } => previous_actions
```
SomeoneToIgnore, line 7397: This seems to be the only reason we use `previous_actions`, but for our case (rapidly moving the caret) and for other action update, seems enough to know that we have a task fetch pending?
We might need to change it to `Option<Task<()>>`?

We have to store that task anyway, so removing `previous_actions` helps us with removing a lot of logic and memory usage.

crates/editor/src/editor.rs
```diff
@@ -7317,7 +7352,7 @@ impl Editor {
 
     pub fn clear_code_action_providers(&mut self) {
         self.code_action_providers.clear();
-        self.available_code_actions.take();
+        self.available_code_actions = AvailableCodeActions::None;
```
SomeoneToIgnore, line 7355: That's not used anywhere, let's remove the method instead please.

crates/editor/src/editor.rs
```diff
@@ -7076,30 +7104,37 @@ impl Editor {
...
         cx.spawn_in(window, async move |editor, cx| {
-            while let Some(prev_task) = task {
-                prev_task.await.log_err();
-                task = editor
-                    .update(cx, |this, _| this.code_actions_task.take())
+            if let Some(task) = task {
+                task.await.log_err();
+            }
+
+            // Drain any chained fetches that started while we waited.
```
SomeoneToIgnore, line 7114: I know this code was here before but given these lines are rewritten anyway, let's try to be better?

LSP code actions are rather ephemeral items, as depend on the selection placement: if we move from a line with the comment to a line with the struct definition, we have to refresh the code actions.

* `fn code_actions` path follows the "deployed at a particular row" action, and the sources can be https://github.com/zed-industries/zed/blob/634e91bdace3946ed6959491b0ed46e35d877f76/crates/editor/src/actions.rs#L109-L113 IIRC we cannot have 2 concurrent sources there, can we?
I see not much of a necessity to wait for any previous rows' tasks, as "caching" (`available_code_actions`'s task's side effect) is done using a separate code path.

* another logic line updates the `available_code_actions` based off selections through `refresh_code_actions`: it updates data based on the newest selection's range?

Both of the code paths try to drop the task in this state, only one controls what to update on.

----------------

That all seems overly complicated and maybe one of the reasons is naming?
Nothing tells that this is a selection-based data (also a bit of settings toggling and LSP adapters added events), so maybe changing that is a good step?

Then, there seems to be no point to manipulate the task in `fn code_actions` and preserve/await on previous tasks at all?

I might be missing something, reading related commits' metadata (such as PRs and issues) is good to try.
If I do not miss anything, removing this all leads to a drastic simplification of all related code: all UI elements need to know if there's a finished or pending task along with the action list?

SomeoneToIgnore (commented):
Thank you, things start to shape, but given how things are involved, we need to go deeper.

crates/editor/src/editor.rs
Dnreikronos: Yeah, you're not missing anything.

**Concurrent sources**: no, `toggle_code_actions` dismisses any existing menu before starting a new one, so only one toggle runs at a time:

```rust
// editor.rs:6917-6928
if let Some(CodeContextMenu::CodeActions(code_actions)) = context_menu.as_ref() {
    if code_actions.deployed_from == action.deployed_from {
        *context_menu = None;
        cx.notify();
        return;
    } else {
        *context_menu = None;
        cx.notify();
    }
}
```

**Selection-based refresh**: yes, `refresh_code_actions` always uses the newest selection:

```rust
let newest_selection = this.selections.newest_anchor().clone();
// ...
let newest_selection_adjusted = this.selections.newest_adjusted(&display_snapshot);
```

**Task manipulation in `fn code_actions`**:  agreed, no point. `refresh_code_actions` already writes `Ready` when its task completes:

```rust
this.code_actions_for_selection = if actions.is_empty() {
    AvailableCodeActions::None
} else {
    AvailableCodeActions::Ready {
        location: Location { buffer: start_buffer, range: start..end },
        actions: actions.into(),
    }
};
cx.notify();
```

So `code_actions()` can just read current state directly. Since the `Fetching` variant carries `previous_actions`, the UI still gets the last known actions during a fetch instead of nothing:

```rust
fn code_actions_for_buffer_row(
    &self,
    buffer_row: u32,
    cx: &App,
) -> Option<Rc<[AvailableCodeAction]>> {
    let (location, actions) = match &self.code_actions_for_selection {
        AvailableCodeActions::Ready { location, actions } => (location, actions),
        AvailableCodeActions::Fetching { previous_actions, .. } => {
            let (location, actions) = previous_actions.as_ref()?;
            (location, actions)
        }
        AvailableCodeActions::None => return None,
    };
    let snapshot = location.buffer.read(cx).snapshot();
    let point_range = location.range.to_point(&snapshot);
    if (point_range.start.row..=point_range.end.row).contains(&buffer_row) {
        Some(actions.clone())
    } else {
        None
    }
}
```

This drops `take_task` entirely. The generation counter in `refresh_code_actions` already handles races between overlapping refreshes.

SomeoneToIgnore: Amazing, so our end data structure for "actions under the caret" is more like an enum with `Fetching(Task<()>)` and `Actions { actions: Vec<_> }`-like options, that simplifes all the code around a lot.
Potentially, some disabled state has to be expressed somehow, maybe another enum variant; maybe that task has to be `Shared<Task<Vec<ActionType>>>`, if the task management is supposed to wait for the latest action set code-wise?
That could be achieved from any task (`cx.spawn` and `cx.spawn_background`) with `.shared()` and a corresponding import.

We should also influence the state in one of the logical branches described above only, the one that reacts to selection changes/action provider addition and the rest action provider-related.

Another place, under this discussion branch, that is used for UI purposes, is odd to see manipulating the state, esp. given the related debounce-related bits.

Among with some renaming, that would be quite a good change to do, since we're rewriting at least 60% of the related lines anyway.

Dnreikronos:
Pushed the changes from your review, @SomeoneToIgnore. 
Happy to adjust if anything's off. 
Thanks for walking me through this one!

SomeoneToIgnore (approved):
I have pushed a few improvements on top:

* code formatting fixes
* renames, mentioned previously in reviews, to make it more clear we deal with selections now
* `detach_and_log_err` removed from the main toggle actions task, so we can cancel previous tasks, as menu popovers for those we hide anyway
* `generation_id` is removed due to the shared task redesign and the previous bullet: we now one have a single task, cancelled periodically, that triggers the code actions refresh and the readers get the shared task and its results for that particular refresh operation now

---------------

Besides the code architecture issues fixed, I have revisited the original issue and tested the branch more, which resulted in a second commit.

I've found that flipping the default "show no tasks when task fetch is pending" might result in even worse flickering in the files with no language servers or with the ones that have no tasks to provide.

Instead of this, I've removed the code that made the toggle code actions task a no-op with no focus on the editor.
https://github.com/zed-industries/zed/commit/2a63c5f9512c076be88fc33af9e8f34f7b0e044b had introduced this along with another enum value and seems to work fine still, without the focus-related check removed.

SomeoneToIgnore:
To be on the safe side, I'll merge this after today's release, and had changed the release notes to not to close the original issue yet without a discussion.

Dnreikronos:
Makes sense, @SomeoneToIgnore. I didn't test that with the focus check on files without a language server or code actions.

Going to run some more tests locally to dig into what you found. Thanks for the help on this!

PR #53566, opened 2026-04-09
Handle changing root paths without splitting in the sidebar

mikayla-maki:
Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #ISSUE

Release Notes:

- N/A

PR #53579, opened 2026-04-10
Revert "Handle changing root paths without splitting in the sidebar"

mikayla-maki:
Was a little hasty with this one

Reverts zed-industries/zed#53566


ConradIrwin:
@zed-zippy approved

PR #53585, opened 2026-04-10
Fix more folder mutation things

mikayla-maki:
Continuation of https://github.com/zed-industries/zed/pull/53566, now with proper thread root mutation.

Self-Review Checklist:

Release Notes:

- N/A


crates/agent_ui/src/thread_metadata_store.rs
```diff
@@ -195,6 +194,135 @@ fn migrate_thread_remote_connections(cx: &mut App, migration_task: Task<anyhow::
...
+                    .root_repo_common_dir()
+                    .and_then(|dir| Some(dir.parent()?.to_path_buf()))
+                    .unwrap_or_else(|| folder_path.clone());
+                (main_path, folder_path)
+            })
+            .unzip();
+        Self {
+            folder_paths: PathList::new(&folders),
+            main_worktree_paths: PathList::new(&mains),
+        }
```
eholk, line 240: `PathList::new` sorts the paths. It looks like this means we're sorting the `folder_paths` and the `main_worktree_paths` separately, so could this cause problems?

I'm guessing we're okay because we use `ordered_paths` where it matters.

crates/agent_ui/src/thread_metadata_store.rs
```diff
@@ -195,6 +194,135 @@ fn migrate_thread_remote_connections(cx: &mut App, migration_task: Task<anyhow::
...
+    }
+
+    /// Build from two parallel `PathList`s that already share the same
+    /// insertion order. Used for deserialization from DB.
+    ///
+    /// Returns an error if the two lists have different lengths, which
+    /// indicates corrupted data from a prior migration bug.
```
eholk, line 247: Is this worth adding a `debug_assert!` about too? I guess not because we don't want to crash if your database is corrupted in a way that we can recover from.

crates/agent_ui/src/thread_metadata_store.rs
```diff
@@ -2816,4 +2978,136 @@ mod tests {
         assert!(paths.contains(&Path::new("/projects/worktree-a")));
         assert!(paths.contains(&Path::new("/projects/worktree-b")));
     }
+
+    // ── ThreadWorktreePaths tests ──────────────────────────────────────
+
+    /// Helper to build a `ThreadWorktreePaths` from (main, folder) pairs.
+    fn make_worktree_paths(pairs: &[(&str, &str)]) -> ThreadWorktreePaths {
```
eholk, line 2985: Minor nit, but I'm a fan of taking `impl IntoIterator<Item = (&str, &str)>` in this cases.

eholk (commented):
@zed-zippy approved

PR #53452, opened 2026-04-08
editor: Add Ctrl+scroll wheel zoom for buffer font size

seanstrom:
Closes https://github.com/zed-industries/zed/issues/53319

Release Notes:

- Add event handling on editor to increase/decrease font-size when using the scroll-wheel and holding the secondary modifier (Ctrl on Linux/Windows, and Cmd on macOS). Disabled by default, use `mouse_wheel_zoom` to enable.

Screen Capture:

https://github.com/user-attachments/assets/bf298be4-e2c9-470c-afef-b7e79c2d3ae6

PR #53598, opened 2026-04-10
Restrict mouse wheel zoom for certain editors

SomeoneToIgnore:
Follow-up of https://github.com/zed-industries/zed/pull/53452

* disables mouse wheel zooming in agent, debugger, keymap editor, dev inspector and repl-related editors
* adjusts the code to call for theme changes directly instead of sending the events, so that agent following does not capture the events and changes its font size

Release Notes:

- N/A

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1183,6 +1183,7 @@ pub struct Editor {
     delegate_open_excerpts: bool,
     enable_lsp_data: bool,
     enable_runnables: bool,
+    enable_mouse_wheel_zoom: bool,
     show_line_numbers: Option<bool>,
     use_relative_line_numbers: Option<bool>,
     show_git_diff_gutter: Option<bool>,
@@ -1972,6 +1973,9 @@ impl Editor {
         clone.read_only = self.read_only;
         clone.buffers_with_disabled_indent_guides =
             self.buffers_with_disabled_indent_guides.clone();
+        clone.enable_mouse_wheel_zoom = self.enable_mouse_wheel_zoom;
+        clone.enable_lsp_data = self.enable_lsp_data;
+        clone.enable_runnables = self.enable_runnables;
         clone
     }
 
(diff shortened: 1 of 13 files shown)


crates/editor/src/editor.rs
```diff
@@ -3351,6 +3356,15 @@ impl Editor {
     }
 
     pub fn set_mode(&mut self, mode: EditorMode) {
+        if self.enable_lsp_data {
+            self.enable_lsp_data = mode.is_full();
+        }
+        if self.enable_runnables {
+            self.enable_runnables = mode.is_full();
+        }
```
MrSubidubi: Are you sure about this? I think that for the current call sites this would not be desirable? Might again be blocked on an overall better approach for all this gating.

MrSubidubi (changes requested):

crates/editor/src/editor.rs
SomeoneToIgnore: I can revert that bit, but not sure what's not desirable here?

We do disable the runnables (and the rest) for non-full mode on creation

https://github.com/zed-industries/zed/blob/dcdecc872f47a42bd2ed0bb4d8c6a83ee1969811/crates/editor/src/editor.rs#L2427

and if we then set the mode that's not full, that's not supposed to be enabled?

MrSubidubi: From what I can see, we currently set the mode in https://github.com/zed-industries/zed/blob/50856c9e74426fb85de356f3d447cad8dca7e579/crates/agent_ui/src/message_editor.rs#L1545 and https://github.com/zed-industries/zed/blob/50856c9e74426fb85de356f3d447cad8dca7e579/crates/debugger_ui/src/session/running/console.rs#L66, both of which I'd think would neither benefit nor want this to be toggled. 

This is obviously only what currently applies and not saying we would not want this at some other point, but until then, I'd think it would probably be best if we just do not do it perhaps?

SomeoneToIgnore: Yeah, so we create an editor with all 3 things disabled and then we do not re-enable them with the new code when setting the new mode.

Seems harmful in this context but sure, removed the code.

MrSubidubi (approved):

PR #53607, opened 2026-04-10
ci: Do not install cargo machete by building it

Veykril:
Re-building the tool on CI every time is pointless when can just install the binary itself

Release Notes:


MrSubidubi (approved):

MrSubidubi:
At some point I'll probably revisit whether we can just provide this in the base-image for the job upfront, but for now this is definitely a good solution as-is. Thanks!

PR #53610, opened 2026-04-10
ci: Create cherry-picks using the create-pull-request action

MrSubidubi:
I ran into some issues in https://github.com/zed-industries/zed/actions/runs/24153812289/job/70487669958 because the permissions for github-actions do not suffice for creating the cherry-picks when workflows are modified. 

With this, I hope to fix this, as all creations will go through a more "battle-tested" routine for the workflows, as this worked already for the extension org and Zed Zippy permissions are sufficient for creating the branch, whereas the for github-actions are not.

Release Notes:

- N/A


MrSubidubi:
Closing for now as https://github.com/zed-industries/zed/pull/53640 seems to have taken care of said issue.

PR #53637, opened 2026-04-10
workspace: Fix missing icons on non-terminal tabs during drag

shreyashsaitwal:
Previously, only terminal tabs displayed their icon while being dragged, all other tab types showed just the label, making the drag preview inconsistent with the actual tab appearance.

https://github.com/user-attachments/assets/7c055945-79d6-4f6e-8969-279a4fb1ff58

This PR fixes that by rendering each tab's icon alongside its label, so the dragged preview now accurately reflects the tab regardless of type.

https://github.com/user-attachments/assets/3ad5a99b-5ba7-41d2-a6b3-51101b1f9650

Release Notes:

- Fixed missing icons on non-terminal tabs when dragging

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -4927,8 +4927,14 @@ impl Render for DraggedTab {
             window,
             cx,
         );
+        let icon = self.item.tab_icon(window, cx).map(|icon| {
+            icon.size(IconSize::Small)
+                .color(Color::Muted)
+                .into_any_element()
+        });
         Tab::new("")
             .toggle_state(self.is_active)
+            .children(icon)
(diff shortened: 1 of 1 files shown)


shreyashsaitwal:
@cla-bot check

shreyashsaitwal:
@danilo-leal, please let me know if there's anything on my end I can do to move this forward. No rush if it's not a priority right now, happy to wait.

smitbarmase (approved):
Thanks! I noticed we are still not handling error/warning indicator cases that the actual tab shows for files with diagnostics. I made it such as that logic is reused so real tab and the dragged tab stay consistent.

Issue #46469, opened 2026-01-09
Project Search includes files in symlinked directories


injust:
### Reproduction steps

1. Check out my [dotfiles repo](https://github.com/injust/dotfiles), including submodules
2. Search for something like `DOCTYPE`, which will find results in `bat/themes/`

### Current vs. Expected behavior

`catppuccin/bat` is a submodule, and `bat/themes/` is a symlink to `../catppuccin/bat/themes/`.

I gitignore `catppuccin/*/*`, which ignores the _contents_ of the submodules, but not the submodules themselves. I don't want them to be indexed or searched, whether directly or through the symlinked directory.

This is basically the opposite of https://github.com/zed-industries/zed/issues/41887.

### Zed version and system specs

Zed: v0.218.6+stable.114.e8b0ac131bc2bf80056f3769317ef57ebf32d0d1 (Zed) 
OS: macOS 15.7.3
Memory: 16 GiB
Architecture: x86_64

injust:
I generalized this issue to include File Finder, since it really boils down to "don't follow symlinked directories by default".

Not following symlinks is the default (but configurable) behaviour in tools like fd and ripgrep, which is a good source of precedent (although they are not editors). I'm unsure what default would be better for Zed.

wezm:
Still an issue on:

Zed 0.233.8 
Commit: aa7f9dc9f87ca27252eea8f6ef0a46122af043f7
Version: 0.233.8+stable.247.aa7f9dc9f87ca27252eea8f6ef0a46122af043f7

Also this issue is labelled `state:needs repro`, but `injust` included reproduction instructions when the issue was opened. I also encounter this issue when working in https://github.com/chimera-linux/cports where [the `.parent` symlink](https://github.com/chimera-linux/cports/blob/master/user/.parent) results in huge amounts of duplicated results.

mmal3k:
I was able to reproduce the File Finder side of this on macOS Apple Silicon with Zed 1.16.2 stable.

Zed version: 1.16.2 (CFBundleVersion 20260824.140013)
OS: macOS 26.2 (25C56)
Architecture: Apple Silicon / ARM64

Steps:
1. Opened a tiny git repo with `themes` → `real/themes` symlink and `real/themes/catppuccin.conf` containing `main_bg=...`.
2. Project Search for `main_bg`.
3. File Finder (`Cmd-P`) for `catppuccin`.

Expected:
[...]

PR #53646, opened 2026-04-10
Symlink scan option

### LINKED ISSUE #48890: Search should never follow symlinks that point back into the project
There is in general no point ever following symlinks pointing within the root for any operation

### Reproduction steps

1. Have symlinks in project pointing elsewhere in project. e.g. src/linked_submodule -> src/submodules/original_submodule

1. File search for “source.cpp”
2. Result contains both src/linked_submodule/source.cpp and src/submodules/original_submodule/source.cpp
3. Pretty much the same outcome for searching for code inside all files, both paths to source.cpp will show in the multibuffer

Zed: 0.222.4
Ubuntu 24.04

auwi-nordic:
Adds option to always search/scan symlinks for more thorough coverage of #41887 issue

This is a significant rewrite and more thoroughly reviewed and tested version of the old PR here:
https://github.com/zed-industries/zed/pull/46344

The "never" option was removed from the old PR, since "expanded" is now the default. Perhaps "never" can be added as an option later if there's demand for it. Adding that option may resolve #48890 for instance.

Self-Review Checklist:

- [x] Performance impact has been considered and is acceptable (significant impact with "always" option, but "expanded" is default)

Release Notes:

- Add option choose between including expanded symlinks or include all symlinks in project search (#41887)


auwi-nordic:
@probably-neb FYI, pinging you since you were involved in the related issues/PRs

auwi-nordic:
I was reading #48890 and I think I may agree that symlinks within the project should be ignored. I explicitly tested for symlinks within the project for this PR (in addition to external symlinks). But it could perhaps be updated to completely disable the inclusion of symlinks within the project.

pyronaur:
@eholk or anyone else from Zed team:

Is there anything I can do to help move this forward? Being able to open files from symlinked directories is mission critical for me, having to expand trees manually just to be able to "cmd+P" to a file kind of defeats the purpose of "quick open".

I really don't want to go back to VSCode, but the lack of ability to configure multiple roots in Zed has really been hurting  the entire time I've used Zed. I was hoping symlinks to be a reasonable workaround, but that turned out to be unusable too.

I would really appreciate a way to be able to add multiple directories from my fs to a single project that I can jump through/search through easily. I don't much care if it's symlinks or configuration of some sort, just as long as it can be done in some way.

ercanozer:
@eholk @smitbarmase +1 here, critical for our team to be able to continue using Zed - 

Possible to have some eyes on here?

24samj:
No updates on this? I feel this is such a critical feature that's missing rn. I too don't wanna go back to VS Code tbf. I can't just keep opening the project panel to literally find the file amongst hundreds of folders.

injust:
See https://github.com/zed-industries/zed/issues/46469 for rationale behind a "never" option.

pleibers:
Please merge this, waiting for it forever already

eholk:
@pyronaur:

> I would really appreciate a way to be able to add multiple directories from my fs to a single project that I can jump through/search through easily. I don't much care if it's symlinks or configuration of some sort, just as long as it can be done in some way.

You are able to add multiple folders to a project already, by right clicking in a blank area in the project panel and choosing "Add Folders to Project..." Does this work for you?

I'm looking over this PR now...

crates/worktree/src/worktree.rs
```diff
@@ -2948,12 +2890,28 @@ impl LocalSnapshot {
 }
 
 impl BackgroundScannerState {
-    fn should_scan_directory(&self, entry: &Entry) -> bool {
+    fn should_scan_directory(
+        &self,
+        entry: &Entry,
+        scan_symlinks: settings::ScanSymlinksSetting,
```
eholk: As a small cleanup/refactor, I think we should move this to BackgroundScanner, and then it can directly access the settings rather than passing them in here.

crates/worktree/src/worktree.rs
```diff
@@ -2948,12 +2890,28 @@ impl LocalSnapshot {
...
         (self.scanning_enabled && !entry.is_external && (!entry.is_ignored || entry.is_always_included))
+            || scan_external
             || entry.path.file_name() == Some(DOT_GIT)
             || entry.path.file_name() == Some(local_settings_folder_name())
             || entry.path.file_name() == Some(local_vscode_folder_name())
-            || self.scanned_dirs.contains(&entry.id) // If we've ever scanned it, keep scanning
+            // In Always mode, external directories are already covered by `scan_external` above,
+            // so we exclude them from the scanned_dirs check to avoid re-scanning them a second
+            // time via a different code path with potentially different conditions.
+            || ((!entry.is_external || scan_symlinks != ScanSymlinksSetting::Always)
```
eholk: Do we need to check scan_symlinks twice? It seems like it should be possible to simplify this conditional somewhat.

eholk (commented):
Thanks for the PR! Overall I think it looks like a nice improvement and directionally good. I was hoping to get this reviewed and merged before I head out for the weekend but unfortunately that didn't happen. Instead, here are some thoughts and comments.

There are a couple inline suggestions.

I think it's fine not to include a Never mode in this PR. It adds a lot of the plumbing needed to add one easily as a follow-up, so if there's enough interest we can do that.

It seems worth doing the work to skip within-project symlinks (or land it as a separate PR since that issue already exists without this PR).

Anyway, I'll finish the review next week. I'm happy to help address some of the suggestions I made too.

auwi-nordic:
> It seems worth doing the work to skip within-project symlinks (or land it as a separate PR since that issue already exists without this PR).

Yeah, that seems like the correct order to fix it to me.

1. Add option to skip within-project symlinks in a new PR
2. If there's still a demand to never include symlinks in index, even for stuff you've explored in project panel, then add "Never" option

> Anyway, I'll finish the review next week. I'm happy to help address some of the suggestions I made too.

I'll hold off on making any changes until then. If you want to just push suggestions feel free

pyronaur:
> @pyronaur:
> 
> > I would really appreciate a way to be able to add multiple directories from my fs to a single project that I can jump through/search through easily. I don't much care if it's symlinks or configuration of some sort, just as long as it can be done in some way.
> 
> You are able to add multiple folders to a project already, by right clicking in a blank area in the project panel and choosing "Add Folders to Project..." Does this work for you?
> 
> I'm looking over this PR now...

@eholk  That works even better!  But where's the project configuration saved? I never use UI to open projects, I just do `zed .` when I'm in the right dir, and that doesn't re-open the project.

Symlinks was always a hack because last I tried, Zed didn't have a multi-root project config. I faintly remember that one variant worked, but it expanded all directories on open and I had to manually toggle them every time, and I jump around a lot, so doing that every time is also cumbersome.

eholk:
@pyronaur - You can also use `zed -a some_path` to add `some_path` as a folder in your currently open workspace. You could probably script this to do something like

```
zed dir1
zed -a dir2
zed -a dir3
```

It might be worth considering a `.zedproject` file at some point that does this for you though.

eholk:
I've been a little swamped this week but I should have time to get back to this starting tomorrow.

pyronaur:
> It might be worth considering a .zedproject file at some point that does this for you though.

🙏  Please this!

eholk (approved):

eholk:
I'm doing a Linux smoke test and assuming that goes well I'll merge this.

eholk:
Hmm, unfortunately my Linux build doesn't seem to be following symbolic links at all. I think this will need some more investigation. Linux isn't my daily driver, so if someone else wants to dive in here and confirm whether it works for them, that would be super helpful!

pleibers:
I tried as well and running this zed version does not follow symlinks in the project search even when folders are expanded. (There shold not be a never option right?)

eholk:
@pleibers - Thanks for the confirmation. Out of curiosity, if you run a build off main (or just run the latest Stable/Preview/Nightly release), does it follow symlinks?

eholk:
After some more investigation, the Linux issue I'm seeing seems independent of this PR, so I think we're ready to merge this one and we can fix the other issues separately.

prigaux:
NB: I wrongly expected this `scan_symlinks` option to have an impact on symlinked files, but it currently only impacts symlinked directories (as expected from the release notes, but not expected from this PR title and from the name of the option)

PR #53654, opened 2026-04-10
Open new remote projects in the same window via sidebar

maxbrunsfeld:
Previously, when opening a remote project via the sidebar, and picking a new project path via the path picker, it would incorrectly open in a new window.

Release Notes:

- N/A


mikayla-maki (approved):

eholk:
@zed-zippy approved

PR #53663, opened 2026-04-11
cli: Add first-run prompt for default open behavior and abstract IPC transport

eholk:
This PR adds the `cli_default_open_behavior` setting and a first-run TUI prompt
that appears when `zed <path>` is invoked without flags while existing windows are
open and the setting hasn't been configured yet.

### Setting and prompt

- Adds a new `cli_default_open_behavior` workspace setting with two values:
  `existing_window` (default) and `new_window`.
- When the user runs `zed <path>` for the first time with existing Zed windows
  open, a `dialoguer::Select` prompt in the CLI asks them to choose their
  preferred behavior. The choice is persisted to `settings.json`.
- The prompt is skipped when:
  - An explicit flag (`-n`, `-e`, `-a`) is given
  - No existing Zed windows are open
  - The setting is already configured in `settings.json`
  - The paths being opened are already contained in an existing workspace

### IPC transport abstraction

- Introduces a `CliResponseSink` trait in the `cli` crate that abstracts
  `IpcSender<CliResponse>`, with an implementation for the real IPC sender.
- Replaces `IpcSender<CliResponse>` with `Box<dyn CliResponseSink>` /
  `&dyn CliResponseSink` across all signatures in `open_listener.rs`:
  `OpenRequestKind::CliConnection`, `handle_cli_connection`,
  `maybe_prompt_open_behavior`, `open_workspaces`, `open_local_workspace`.
- Extracts the inline CLI response loop from `main.rs` into a testable
  `cli::run_cli_response_loop` function.
- Switches the request channel from bounded `mpsc::channel(16)` to
  `mpsc::unbounded()`, eliminating `smol::block_on` in the bridge thread.

### End-to-end tests

Seven new tests exercise both the CLI-side response loop and the Zed-side
handler connected through in-memory channels, using `allow_parking()` so the
real `cli::run_cli_response_loop` runs on an OS thread while the GPUI executor
drives the Zed handler:

- No flags, no windows → no prompt, opens new window
- No flags, existing windows, user picks "existing window" → prompt, setting persisted
- No flags, existing windows, user picks "new window" → prompt, setting persisted
- Setting already configured → no prompt
- Paths already in existing workspace → no prompt
- Explicit `-e` flag → no prompt
- Explicit `-n` flag → no prompt

Existing tests that previously used `ipc::channel()` now use a
`DiscardResponseSink`, removing OS-level IPC from all tests.

Release Notes:

- Added a first-run prompt when using `zed <path>` to choose between opening
  in an existing window or a new window. The choice is saved to settings and
  can be changed later via the `cli_default_open_behavior` setting.


injust:
Just to clarify here, by paths you mean files?

Because currently (in 0.231.x), `zed <file>` opens in an already open window. But `zed <directory>` doesn't do that, instead opening the directory as a new workspace (in a new window).

nathansobo (approved):

nathansobo:
@injust The plan is to let you choose the behavior you want. I happen to want to add folders to my active window. This depends on the sidebar feature we haven't rolled out yet.

@eholk Is there anything we need to do to feature flag this or full speed ahead?

eholk:
> @eholk Is there anything we need to do to feature flag this or full speed ahead?

Given that we ask users what to do the first time it matters, I think full speed ahead is fine. I'm happy to flag it if we want a little more time to try it out though.

eholk:
@injust - The file behavior will be the same, but with this change for directories you'll have the option of adding directories as a new workspace to an existing window or opening them in a new menu.

PR #53671, opened 2026-04-11
Clean up empty parent directory after archiving worktree

rtfeldman:
Zed creates worktrees at `<worktrees_dir>/<branch_name>/<project_name>/`. When `git worktree remove` deletes the inner project directory, the intermediate branch-name directory is left behind as an empty folder on disk.

After a successful `git worktree remove`, this now checks if the parent directory is empty AND is inside Zed's managed worktrees directory (from the `git.worktree_directory` setting). If both conditions are met, the empty parent is removed. Directories outside the managed worktrees directory are never touched, so user-created worktrees at custom paths won't have their parent directories deleted.

Release Notes:

- Fixed empty parent directories being left behind after archiving a thread that used a linked git worktree.

### DIFF
--- a/crates/agent_ui/src/thread_worktree_archive.rs
+++ b/crates/agent_ui/src/thread_worktree_archive.rs
@@ -5,7 +5,6 @@ use std::{
 
 use agent_client_protocol as acp;
 use anyhow::{Context as _, Result, anyhow};
-
 use gpui::{App, AsyncApp, Entity, Task};
 use project::{
     LocalProjectFlags, Project, WorktreeId,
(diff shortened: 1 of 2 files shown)


crates/agent_ui/src/thread_worktree_archive.rs
```diff
@@ -5,11 +5,13 @@ use std::{
 
 use agent_client_protocol as acp;
 use anyhow::{Context as _, Result, anyhow};
+
```
rtfeldman:
```suggestion
```

Delete unnecessary blank line.

rtfeldman:
Closing in favor of #53672

PR #53680, opened 2026-04-11
languages: Fix Debug Test for Go subtests

yeskunall:
The `go-subtest` task template wrapped the `-run` arg in single quotes for shell safety. This works for Run Test (terminal strips quotes), but Debug Test sends the arg through Delve’s DAP protocol with no shell involved, so the literal quote characters ended up in the regex and prevented any match.

All other Go task templates (`go-test`, `go-testify-suite`, `go-table-test-case`) use the backslash-escaped format which `GoLocator` already knows how to handle. Align the `go-subtest` template to the same format.

Self-Review Checklist:

Closes #53230.

Release Notes:

- Fixed Debug Test for Go subtests

### DIFF
--- a/crates/languages/src/go.rs
+++ b/crates/languages/src/go.rs
@@ -706,7 +706,7 @@ impl ContextProvider for GoContextProvider {
                     "-v".into(),
                     "-run".into(),
                     format!(
-                        "'^{}$/^{}$'",
+                        "\\^{}\\$/\\^{}\\$",
                         VariableName::Symbol.template_value(),
                         GO_SUBTEST_NAME_TASK_VARIABLE.template_value(),
                     ),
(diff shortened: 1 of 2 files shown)


rft03:
Hi @yeskunall, has there been any update on this?

aexvir:
also hitting this constantly, making hard to use zed for serious work 😕

rft03:
> also hitting this constantly, making hard to use zed for serious work 😕

@aexvir yep. Most of the tests in the projects that Im working has lots of subtests and table tests. And my workflow needs to debug every once in a while. I might as well keep using goland than work a bit in zed and a bit in goland. I hoped they would fix this before v1 😅

smitbarmase (approved):

PR #53692, opened 2026-04-11
git_graph: Hide commits that jj considers hidden

timvermeulen:
Not sure if this is something the team is interested in right now given the current lack of jj-specific features, but I'm opening this anyway because the git graph is the main git feature in Zed that isn't very usable currently in git-colocated jj repos. Jujutsu creates a lot of "hidden" commits that don't show up in the `jj log` output, but they do currently show up in the git graph in Zed 🙂 

This is a simple proof of concept that uses the `jj` binary to obtain the `visible_heads()` revset to pass to `git log`. We'll probably want to use jj-lib for proper jj support (though I don't know how feature complete it currently is).

I went with the `visible_heads()` revset because it most closely matches the current `git log --all` output, but we could also use the `revsets.log` config value (wrapped in `heads(...)`) to more closely match the output of `jj log`. Or we could add a Zed setting for this.

Self-Review Checklist:

Release Notes:

- Improved the git graph to hide commits that `jj` considers hidden.

### DIFF
--- a/crates/git/src/git.rs
+++ b/crates/git/src/git.rs
@@ -1,6 +1,7 @@
 pub mod blame;
 pub mod commit;
 mod hosting_provider;
+mod jj;
 mod remote;
 pub mod repository;
 pub mod stash;
--- a/crates/git/src/jj.rs
+++ b/crates/git/src/jj.rs
@@ -1,0 +1,46 @@
+use anyhow::{Context as _, Result};
+
+use std::path::Path;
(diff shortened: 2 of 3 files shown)


Anthony-Eid:
Thank you for contributing! I'm going to close this because we don't want to be running one off JJ commands, unless there's a JJRepository structure that implements the GitRepository trait. Which the team doesn't have the bandwidth to prioritize right now.

PR #53693, opened 2026-04-11
acp_thread: Forward token usage from PromptResponse to thread UI

hahahuy:
Fixes #53130

## Summary

External ACP agents (e.g. Claude Code CLI) return token usage in `PromptResponse` via the `unstable_session_usage` feature, which Zed already enables. Previously this data was discarded — `token_usage` stayed `None` on the `AcpThread`, so `render_token_usage()` returned early and showed nothing.

**Changes:**

- `acp_thread.rs`: Extract `r.usage` in the `Ok(r)` branch of `run_turn` and call `update_token_usage()` each turn.
- `thread_view.rs`: Add a `max_tokens == 0` branch in `render_token_usage` — since the ACP protocol does not expose context window size, render a plain `"N tokens"` label instead of a circular progress ring (which would show meaningless `"0% • N / 0"`).
- Two new `#[gpui::test]` tests: one asserting usage is populated after a turn with usage data, one asserting `token_usage` stays `None` when `PromptResponse` has no usage.

## Test plan

- [ ] Connect an external ACP agent (e.g. `claude` CLI) and send a message
- [ ] Confirm a token count label (e.g. `"1.5k tokens"`) appears in the thread header after the first response
- [ ] Confirm built-in Zed agent token display (circular ring with percentage) is unchanged
- [ ] Run `cargo test -p acp_thread test_token_usage`

Release Notes:

- Fixed missing token usage display for external ACP agents (e.g. Claude Code CLI) in the agent panel


hahahuy:
@cla-bot check

hahahuy:
@cla-bot check

hahahuy:
@cla-bot check

hahahuy:
Inspired by my Claude CLI's terminal output, which displays per-turn token usage, context window consumption, model name, and relevant context percentage:

Screenshot_20260412_021356

Before:

Screenshot_20260412_021212

After:
                                                                 
Screenshot_20260412_021307

A small change — but knowing you're at 30K/200K context without leaving the editor is genuinely useful when working with long conversations.

  Once this lands, I plan to extend it with the remaining indicators:
   - Tokens consumed this turn
  - Context window fill (e.g. 30K / 200K)
  - Model name
  - Relevant context %

benbrandt:
I think what we need is instead the context update notifications? I dont think we should be keying the ui off of 0 max tokens

PR #53696, opened 2026-04-11
agent_ui: Add some general UI fixes

danilo-leal:
A note-worthy thing I'm adding in this PR to fix an agent panel problem is a trait method allowing to set a min-width in panels. I think some of them—if not all—could benefit from it, because there is a certain width these panels can get in that we  just can't ensure they will work; things will break. So we either accept that and let the user's common sense understand that, or we don't allow them to reach that state... Surely, a small enough _window_ size can still break things, but then it's out of the realms of solution...

Release Notes:

- Agent: Fixed multi-line queued messages getting cut-off when the agent panel is in full screen.
- Agent: Fixed agent panel getting auto-closed after submitting a queued message when the panel is in full screen.
- Agent: Added a min-width to the agent panel, preventing it from reaching a small enough width where it would be essentially unusable.

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -90,6 +90,7 @@ use zed_actions::{
 };
 
 const AGENT_PANEL_KEY: &str = "agent_panel";
+const MIN_PANEL_WIDTH: Pixels = px(300.);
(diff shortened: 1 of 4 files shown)


eholk:
@zed-zippy approved

mikayla-maki:
@zed-industries/approved

PR #53699, opened 2026-04-11
markdown_preview: Fix markdown tasklist not being shown

AdamJedl:
Fixes a regression introduced in #52008

It's because [pulldown-cmark](https://crates.io/crates/pulldown-cmark) may insert Paragraph between Item and TaskListMarker

example markdown:
```md
- [x] a

- [ ] b

```

before:
image

after:
image

<details>
<summary>ParsedMarkdown</summary>

```rust
ParsedMarkdown {
    source: "- [x] a\n\n- [ ] b\n",
    events: [
        (
            0..17,
            RootStart,
        ),
        (
            0..17,
            Start(
                List(
                    None,
                ),
            ),
        ),
        (
            0..9,
            Start(
                Item,
            ),
        ),
        (
            6..8,
            Start(
                Paragraph,
            ),
        ),
        (
            2..5,
            TaskListMarker(
                true,
            ),
        ),
        (
            6..7,
            Text,
        ),
        (
            6..8,
            End(
                Paragraph,
            ),
        ),
        (
            0..9,
            End(
                Item,
            ),
        ),
        (
            9..17,
            Start(
                Item,
            ),
        ),
        (
            15..17,
            Start(
                Paragraph,
            ),
        ),
        (
            11..14,
            TaskListMarker(
                false,
            ),
        ),
        (
            15..16,
            Text,
        ),
        (
            15..17,
            End(
                Paragraph,
            ),
        ),
        (
            9..17,
             q
        (
            0..17,
            End(
                List(
                    false,
                ),
            ),
        ),
        (
            0..17,
            RootEnd(
                0,
            ),
        ),
    ],
    languages_by_name: {},
    languages_by_path: {},
    root_block_starts: [
        0,
    ],
    html_blocks: {},
    mermaid_diagrams: {},
}
```

</details>

Self-Review Checklist:

Release Notes:

- Fixed markdown tasklist not being shown


MrSubidubi (changes requested):
Thanks for this! 

Given that you seem to have an example for how to repro this, could you please add a test for this very case to make sure this does not regress again?

crates/markdown/src/markdown.rs
```diff
@@ -1801,12 +1801,18 @@ impl Element for MarkdownElement {
                             builder.push_div(div().pl_2p5(), range, markdown_end);
                         }
                         MarkdownTag::Item => {
-                            let bullet =
-                                if let Some((task_range, MarkdownEvent::TaskListMarker(checked))) =
-                                    parsed_markdown.events.get(index.saturating_add(1))
-                                {
+                            let task_list_marker = (1..=2).find_map(|offset| match parsed_markdown
```
osiewicz, line 1804: These numbers are a bit magical; would you mind adding a comment to explain why we're alternating the offsets?

AdamJedl:
@MrSubidubi I added a test, but I don't know if this is the best way because I needed to make some structs and fields public.

MrSubidubi:
Thanks for looking into the tests! However, we - should not make GPUI structs pub for this under any circumstances. 

Could we instead try to test the `request_layout` method return here only? I think we have some examples for this somewhere already, but might be wrong. Happy to provide guidance if needed.

AdamJedl:
I tried `request_layout` already and I was not able to get Checkbox, only `gpui::elements::div::Div`, `gpui::elements::text::StyledText`, `gpui::element::AnyElement`, probably because Checkbox is a component, not an element. I couldn't find any similar test.

MrSubidubi:
Ah sorry for sending you down the wrong path. 

Could you see whether you can wire through something using the debug selector method? See https://github.com/zed-industries/zed/blob/7a076aea470dd26bb3aec30e88de3d5abca7bfae/crates/gpui/src/elements/div.rs#L785-L792

This might require changes for the checkbox component but might be the best way foward

crates/markdown/src/markdown.rs
```diff
@@ -3968,4 +3976,52 @@ mod tests {
...
+                        cx,
+                    )
+                });
+
+                TestWindow { markdown }
+            });
+
+            cx.run_until_parked();
+
+            assert!(cx.debug_bounds("checkbox").is_some())
```
osiewicz: Could we incorporate the id of a checkbox into the debug selector? That way your assertions could be more fine-grained

PR #53706, opened 2026-04-11
agent_ui: Fix creating a thread from the toolbar

### LINKED ISSUE #53705: ACP Agent Icon can lie
### Reproduction steps

Had a pretty nasty bug in Nightly. I assumed I was using Claude ACP because it showed the Claude icon in the corner. 

Only found out that I was really using Zed's agent when I got git by a $20 charge for todays work. There were signs. The input box said Zed.  I'm just used to checking with the corner icon and assumed it was trustworthy.

Agent icon **must always** match what agent is being used. I don't know if this is common or not. But can end up with a user being charged a lot of money.

Image

Zed: v0.233.0+nightly.2395.93e36aa55ef9b43efc59ebe7a3f036149f88db47 (Zed Nightly 93e36aa55ef9b43efc59ebe7a3f036149f88db47) 
OS: macOS 26.3
Memory: 24 GiB
Architecture: aarch64

danilo-leal:
Follow up to https://github.com/zed-industries/zed/pull/53574
Closes https://github.com/zed-industries/zed/issues/53705
Closes #53439

In the draft feature PR, we didn't update the new thread creation from the agent panel's toolbar to _also_ create a corresponding draft item in the sidebar. Additionally, the draft feature exposed a previously-existing bug where the toolbar wouldn't get synced with the currently selected agent; and as you navigated between them, we would fall back to the native agent, as well as when you hit cmd-n.

cc @mikayla-maki just a quick FYI here for the revamp! Will push this forward in the meantime given it's a bit of a bad bug.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -7396,7 +7396,7 @@ mod tests {
 
         // Open a different thread (stub agent) — this navigates away from the draft.
         let connection = StubAgentConnection::new();
-        let stub_agent = Agent::Custom { id: "stub".into() };
+        let stub_agent = Agent::Custom { id: "Test".into() };
         open_thread_with_connection(&panel, connection.clone(), &mut cx);
         let other_session_id = active_session_id(&panel, &cx);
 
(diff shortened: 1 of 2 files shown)


eholk:
@zed-zippy approved

mikayla-maki:
@zed-industries/approved

PR #53710, opened 2026-04-11
workspace: Add separate "format and save" action

cameron1024:
Release Notes:

- Added `workspace: format and save` action which always formats, regardless of settings

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -302,6 +302,8 @@ actions!(
         ResetOpenDocksSize,
         /// Reloads the application
         Reload,
+        /// Formats and saves the current file, regardless of the format_on_save setting.
+        FormatAndSave,
         /// Saves the current file with a new name.
         SaveAs,
         /// Saves without formatting.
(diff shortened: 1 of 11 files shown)


crates/workspace/src/workspace.rs
```diff
@@ -6938,6 +6963,11 @@ impl Workspace {
                     .save_active_item(action.save_intent.unwrap_or(SaveIntent::Save), window, cx)
                     .detach_and_prompt_err("Failed to save", window, cx, |_, _, _| None);
             }))
+            .on_action(cx.listener(|workspace, _: &FormatAndSave, window, cx| {
+                workspace
+                    .save_active_item(SaveIntent::FormatAndSave, window, cx)
+                    .detach_and_prompt_err("Failed to save", window, cx, |_, _, _| None);
```
Anthony-Eid, line 6969: nit: Maybe we should say "Failed to save: {item_label}" or something like that

Anthony-Eid (approved):

crates/workspace/src/workspace.rs
cameron1024: Just went to fix but seems like these are being registered at the workspace, so not necessarily clear where the item_label would come from (since I guess the target item is computed in `save_active_item`). There's a few methods near this one that do the same thing 🤷

PR #53712, opened 2026-04-11
editor: Fix semantic tokens missing when opening buffer from multibuffer

### LINKED ISSUE #53051: Semantic tokens are not highlighted when a buffer is opened from a multibuffer
### Reproduction steps

1. Create a project with two Python files, `foo.py` and `bar.py`, each containing a single line:
```python
print("Hello world")
```
2. In settings.json, configure: "semantic_tokens": "full". Ensure a Python LSP (`ty` here) is active and providing semantic tokens.
3. Open a multibuffer containing these two files (e.g., by performing a global search for "print").
4. Observe that the contents in the multibuffer are highlighted correctly.
5. Open one of the files from the multibuffer (e.g., by clicking the file header or the line number). Make sure this file was not already open in another tab.
6. Observe that the semantic highlighting is missing in the newly opened buffer.
7. Close that buffer and open the same file from the Project Panel; this time, semantic highlighting works correctly.

Here is a screen recording:

https://github.com/user-attachments/assets/51219441-be83-4943-8a2c-7398d6792bea

Current Behavior:
Semantic tokens are not provided when the buffer is opened from a multibuffer

Expected Behavior:
Semantic tokens should be requested and applied when a buffer is opened from a multibuffer.

Zed: v0.232.0+nightly.2367.a7248e8c006af22263d97e4eb234cb613a9bf605 (Zed Nightly a7248e8c006af22263d97e4eb234cb613a9bf605) 
OS: Windows 10.0.26200
Memory: 47.8 GiB
Architecture: x86_64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```jsonc
{
  "languages": {
    "Python": {
      "semantic_tokens": "full",
      "language_servers": ["ty", "ruff"],
    },
  }
}
```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Dnreikronos:
Summary

Semantic token highlighting was missing when opening a file from multibuffer search results (Ctrl+Shift+F). Which file got hit depended on window size and scroll offset.

## Root cause

Two async tasks race to write `post_scroll_update`:

1. `set_visible_line_count` (scroll.rs:682) fires on first render and spawns a task that calls `register_visible_buffers` + `update_lsp_data` (requests semantic tokens).

2. `open_buffers_in_workspace` (editor.rs:25049) calls `change_selections` with autoscroll right after creating the editor. This emits `ScrollPositionChanged`, whose handler (editor.rs:2655) replaces `post_scroll_update` with a task calling `update_data_on_scroll`.

3. `update_data_on_scroll` (editor.rs:26099) has a singleton guard: `if !self.buffer().read(cx).is_singleton()` that skips `update_lsp_data` for single-file buffers. This is a scroll optimization, singleton buffers don't change their visible buffer set on scroll.

4. The initial task gets dropped, the replacement skips `update_lsp_data`, semantic tokens are never requested.

## Fix

Added a `needs_initial_lsp_data` flag to the Editor struct, set to `true` on creation. `update_data_on_scroll` checks this flag alongside the singleton guard, so `update_lsp_data` runs at least once even for singletons. The flag flips to `false` right after, so subsequent scrolls behave exactly as before. No perf impact after the first render.

## Self-review checklist

Closes #53051

## Demo

Before:
https://github.com/user-attachments/assets/77d07d95-cb4a-44ff-842d-1f7a46653ca9
After:

https://github.com/user-attachments/assets/2c942f52-4ec3-459f-a97b-93919e4bfb3d

## Release notes

- Fixed semantic token highlighting missing when opening a buffer from multibuffer search results

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1182,7 +1182,7 @@ pub struct Editor {
     delegate_stage_and_restore: bool,
     delegate_open_excerpts: bool,
     enable_lsp_data: bool,
-    needs_initial_lsp_data: bool,
+    needs_initial_data_update: bool,
     enable_runnables: bool,
     enable_mouse_wheel_zoom: bool,
     show_line_numbers: Option<bool>,
@@ -1976,7 +1976,7 @@ impl Editor {
             self.buffers_with_disabled_indent_guides.clone();
         clone.enable_mouse_wheel_zoom = self.enable_mouse_wheel_zoom;
         clone.enable_lsp_data = self.enable_lsp_data;
-        clone.needs_initial_lsp_data = self.enable_lsp_data;
+        clone.needs_initial_data_update = self.enable_lsp_data;
         clone.enable_runnables = self.enable_runnables;
         clone
     }
@@ -2426,7 +2426,7 @@ impl Editor {
             delegate_stage_and_restore: false,
             delegate_open_excerpts: false,
             enable_lsp_data: full_mode,
-            needs_initial_lsp_data: full_mode,
+            needs_initial_data_update: full_mode,
             enable_runnables: full_mode,
             enable_mouse_wheel_zoom: full_mode,
             show_git_diff_gutter: None,
@@ -2655,16 +2655,7 @@ impl Editor {
                             );
                         });
 
-                        editor.post_scroll_update = cx.spawn_in(window, async move |editor, cx| {
-                            cx.background_executor()
-                                .timer(Duration::from_millis(50))
-                                .await;
-                            editor
-                                .update_in(cx, |editor, window, cx| {
-                                    editor.update_data_on_scroll(window, cx)
-                                })
-                                .ok();
-                        });
+                        editor.update_visible_data(true, window, cx);
                     }
                     editor.refresh_sticky_headers(&editor.snapshot(window, cx), cx);
                 }
@@ -20863,7 +20854,7 @@ impl Editor {
         cx.notify();
 
         self.scrollbar_marker_state.dirty = true;
-        self.update_data_on_scroll(window, cx);
+        self.update_visible_data(false, window, cx);
         self.folds_did_change(cx);
     }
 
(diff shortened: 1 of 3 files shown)


crates/editor/src/editor.rs
```diff
@@ -1182,6 +1182,7 @@ pub struct Editor {
     delegate_stage_and_restore: bool,
     delegate_open_excerpts: bool,
     enable_lsp_data: bool,
+    needs_initial_lsp_data: bool,
```
```suggestion
    needs_initial_data_update: bool,
```

This is not about LSP only, so the name is a bit odd.
The new one is also bad, so other options are interesting to consider (but then, `update_data_on_scroll` has to be renamed either).

SomeoneToIgnore (dismissed):
Thank you for this very nice find.

Before it gets too late, a few notes:

* the new test passes on `main` hence does not test anything related really
* we'd better rework the code to use `update_data_on_scroll` in both cases then
* `fold_creases` better re-set the `post_scroll_update` task then, as it's now important to make all of them sequential

crates/editor/src/editor.rs
Dnreikronos: My proposal about the variable names:
- `update_data_on_scroll` → `refresh_visible_data`  
- `needs_initial_data_update` → `needs_first_data_refresh`


What you think, @SomeoneToIgnore?

Dnreikronos:
> Thank you for this very nice find.
> 
> Before it gets too late, a few notes:
> 
> * the new test passes on `main` hence does not test anything related really
> * we'd better rework the code to use `update_data_on_scroll` in both cases then
> * `fold_creases` better re-set the `post_scroll_update` task then, as it's now important to make all of them sequential

Thanks for the review! Here's what I changed:

 1. `set_visible_line_count` now calls `update_data_on_scroll` instead of duplicating the logic.
 2. `fold_creases` resets `post_scroll_update` before calling `update_data_on_scroll` synchronously.
 3. Removed the now-unused `InlayHintRefreshReason` import from scroll.rs.

On the test: it passes on `main` because `set_visible_line_count` used to call `update_lsp_data` directly, bypassing the singleton guard. After the refactor it goes through `update_data_on_scroll`, so the flag is what keeps the test green. That said, it's a heavy test, happy to drop it if you prefer.

SomeoneToIgnore:
Just to mention, I do not see those changes described.
Nonetheless, a few comments:

> `fold_creases` resets `post_scroll_update` before calling `update_data_on_scroll` synchronously.

The reset seems somewhat odd? I would expect all code paths to replace the old task with the new task, and, depending on the conditions, use the timeout or not.
Either way, all places where we update the data would be to call one method without any extra code around.
May be good to refactor the `update_data_on_scroll` method to actually set the task itself?

> On the test: it passes on `main` because

That does not make any sense to me: I have taken the test only and applied to `main`, no other code changes — it passes on the old code which means the test is wrong and exposes no bugs.

crates/editor/src/editor.rs
SomeoneToIgnore: Yeah, no, let's stick with [the proposed](https://github.com/zed-industries/zed/pull/53712#discussion_r3069218960) options then please.

Dnreikronos:
> Just to mention, I do not see those changes described. Nonetheless, a few comments:
> 
> > `fold_creases` resets `post_scroll_update` before calling `update_data_on_scroll` synchronously.
> 
> The reset seems somewhat odd? I would expect all code paths to replace the old task with the new task, and, depending on the conditions, use the timeout or not. Either way, all places where we update the data would be to call one method without any extra code around. May be good to refactor the `update_data_on_scroll` method to actually set the task itself?
> 
> > On the test: it passes on `main` because
> 
> That does not make any sense to me: I have taken the test only and applied to `main`, no other code changes — it passes on the old code which means the test is wrong and exposes no bugs.

You're right, the test can't fail on main.
On main, set_visible_line_count calls update_lsp_data unconditionally for every editor, so singletons always get tokens on first render through that path. The test can't fail there because that unconditional call covers it.

This was wrong, and I apologize for the confusion. I rewrote the test and it now fails on main with no other code changes.

The previous test used a 1-line file. Autoscroll had nothing to scroll, so ScrollPositionChanged never fired, and the initial set_visible_line_count task ran fine. The race never triggered.

The new test uses a 100-line file with the multibuffer excerpt at lines 95-100. When the singleton opens and autoscrolls to that position, it actually changes the scroll position. That fires ScrollPositionChanged, whose handler replaces post_scroll_update with a debounced task that goes through update_data_on_scroll, which skips update_lsp_data for singletons. The initial task from set_visible_line_count gets dropped, tokens are never requested.

Also pushed the other changes you asked for:
- update_data_on_scroll renamed to update_visible_data, now sets the task itself
- Non-debounce path calls the method directly instead of spawning a task
- needs_initial_lsp_data renamed to needs_initial_data_update


If you find any other thing that you think it's not good, please let me know

Dnreikronos:
Once everything is validated I'll squash everything into a single commit

SomeoneToIgnore (commented):
Thank you, this looks quite good now — will wait until it turns into a ready PR, not a draft.

SomeoneToIgnore (approved):
Thank you.

PR #53713, opened 2026-04-11
terminal_view: Show hollow cursor when bar/underline is unfocused

Dnreikronos:
Self-Review Checklist:**

## What

When terminal cursor shape is set via escape sequences (e.g. `\e[6 q` for bar, `\e[4 q` for underline), the cursor looks identical whether the terminal is focused or unfocused. Block cursors already convert to a hollow outline when unfocused, but bar and underline were missing this treatment.

## How

Added unfocused guards for `Beam` and `Underline` in the cursor shape match in `terminal_element.rs`, converting them to `Hollow` when the terminal loses focus. This follows the same pattern already used for `Block`.

## Why this approach

A hollow block outline is the clearest unfocused signal and stays consistent with the existing Block behavior. Alternative approaches like dimming opacity or drawing outlined versions of bar/underline were considered, but a 2px-wide outlined bar would be nearly invisible at normal font sizes.

## Demo:
Before: 

https://github.com/user-attachments/assets/81d49899-0837-42fe-a68f-4eb745892af6

After:

https://github.com/user-attachments/assets/a4444e28-5835-4c9f-872f-e9ce8c4805a0

Closes #52716

Release Notes:

- Fixed terminal bar and underline cursors set via escape sequences not visually distinguishing between focused and unfocused states


ConradIrwin (approved):
Thanks for fixing this

PR #53718, opened 2026-04-12
git_graph: Refresh graph when git refs change

### LINKED ISSUE #53594: Git Graph not auto refreshing
### Reproduction steps

1. Go to the Git Graph Window
2. Add a tag in the terminal
3. See that no tag appears in the Git Graph window

Current behavior:
I add a tag via terminal and the tag doesn't appear in the Git Graph window. Only after full workspace reload.

Image

Expected behavior:
I add a tag via terminal and a tag appears in the git graph. Or at least add a refresh button.

Image

OS: macOS 26.3.1 (a) (Build 25D771280a)
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

lingyaochu:
Self-Review Checklist:

Closes #53594

Previously, the Git Graph UI only automatically refreshed when `HEAD` or branches (local/remote) were modified. This meant that creating or deleting tags, or updating other non-branch refs, did not trigger a UI update until another action forced a refresh.

This PR adds a comprehensive `refs()` scan to the repository snapshot (via `git show-ref`), allowing the system to track the exact state of all refs. By emitting a new `RefsChanged` event, the Git Graph now immediately reflects new tag decorations when they are created, deleted, or fetched.

Release Notes:

- Fixed an issue where the Git Graph UI would not automatically refresh when Git tags were created, deleted, or updated.

### DIFF
--- a/crates/project/src/git_store.rs
+++ b/crates/project/src/git_store.rs
@@ -33,7 +33,7 @@ use git::{
     parse_git_remote_url,
     repository::{
         Branch, CommitDetails, CommitDiff, CommitFile, CommitOptions, CreateWorktreeTarget,
-        DiffType, FetchOptions, GitRepository, GitRepositoryCheckpoint, GraphCommitData,
+        DiffType, FetchOptions, GitRef, GitRepository, GitRepositoryCheckpoint, GraphCommitData,
         InitialGraphCommitData, LogOrder, LogSource, PushOptions, Remote, RemoteCommandOutput,
         RepoPath, ResetMode, SearchCommitArgs, UpstreamTrackingStatus, Worktree as GitWorktree,
     },
@@ -295,7 +295,7 @@ pub struct RepositorySnapshot {
     pub remote_upstream_url: Option<String>,
     pub stash_entries: GitStash,
     pub linked_worktrees: Arc<[GitWorktree]>,
-    pub refs: Arc<[SharedString]>,
+    pub refs: Arc<[GitRef]>,
 }
 
 type JobId = u64;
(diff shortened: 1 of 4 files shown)


lingyaochu:
Is `RefsChanged` duplicative with `BranchListChanged`? Branches are refs, so both fire for branch changes. Would `TagListChanged` be better? It avoids duplication and addresses the tag refresh issue directly.

Anthony-Eid:
Hey @lingyaochu, thank you for catching this bug and providing a fix, there's some things I would like to change before merging this PR. 

Namely, we are adding branches to the repo snapshot which is a subset of refs, I think we can solely track refs and get rid of branches. Would you like to pair next week on fixing this together: https://cal.com/anthony-eid/community-pairing

We could also check async via discord, my username is Anthony.Eid

crates/project/src/git_store.rs
```diff
@@ -7501,14 +7506,15 @@ async fn compute_snapshot(
...
+    let (branches, head_commit, all_worktrees, refs) = cx
         .background_spawn({
             let backend = backend.clone();
             async move {
-                futures::future::try_join3(
+                futures::future::try_join4(
                     backend.branches(),
                     head_commit_future,
                     backend.worktrees(),
+                    backend.refs(),
```
Anthony-Eid, line 7517: This is what I'm talking about, we call both backend.refs() and backend.branches() instead of just `backend.refs()` (which should be able to track branches as well)

Also, I think we need to add collab support to update remote repos with refs as well

Anthony-Eid: We also should compare the performance of backend.refs() with backend.branches() and make sure there's no regression, because `compute_snapshot` blocks the git UI state from updating. We want to make sure that function stays fast for large repos

PR #53725, opened 2026-04-12
Add MCP task-augmented tool calls with model continuation

dsturnbull:
Depends on #53044.

Implements the [MCP Tasks protocol](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) (spec 2025-11-25, experimental). MCP servers can now run long-lived tool calls. Zed polls for status, shows live progress in the tool card, and optionally lets the model keep working while a task runs in the background.

Also adds a `background` mode to the built-in terminal tool using the same infrastructure.

## What changed

**Wire types** (`context_server/src/types.rs`): `Task`, `TaskStatus`, `CreateTaskResult`, capability structs, per-tool `execution.taskSupport` negotiation (`required`/`optional`/`forbidden`), and all the `tasks/*` RPC request types. 16 serde tests.

**Protocol** (`context_server/src/protocol.rs`, `test.rs`): `CallToolAsTask` shares the `tools/call` method but expects `CreateTaskResult` back. `FakeTransport` gained `send_notification()` for injecting progress/status notifications in tests.

**Tool call flow** (`tools/context_server_registry.rs`): When a server advertises task support and the tool isn't `forbidden`, Zed sends `CallToolAsTask` with a progress token and TTL. It subscribes to `notifications/progress` and `notifications/tasks/status`, then either:

- **Blocks**: polls `tasks/get` in a select loop (timer vs cancellation), forwarding status messages to the tool card, and fetches `tasks/result` when done.
- **Continues**: if the server sends `model-immediate-response` in `_meta`, the model gets that provisional result immediately. A detached background poller keeps polling and updates the tool card when the task finishes. Handles both the string form (per spec) and structured `CallToolResponse` (with audience annotations).

**Provisional results** (`thread.rs`, `language_model_core/request.rs`): `is_provisional` flag on `LanguageModelToolResult` and `AgentToolOutput`. When set, `process_tool_result` feeds the result to the model but doesn't flip the tool card to Completed. The background poller does that later. `ToolCallEventStream::set_provisional()` exposes this to built-in tools too (used by the terminal).

**Background event drain** (`agent.rs`): `handle_thread_events` normally returns on the `Stop` event. Background tasks that outlive the turn still need to push tool card updates. A detached `drain_background_events` task keeps consuming the event channel after Stop so those late updates reach the AcpThread.

**Per-tool cancellation** (`thread.rs`, `acp_thread.rs`, `connection.rs`): Each tool call gets its own `watch::channel`. `cancel_tool_call()` fires the individual channel without stopping the turn. `cancelled_by_user()` races both the turn-level and per-tool channels.

**Running tasks bar** (`agent_ui/thread_view.rs`): Shows all in-progress tool calls (not just MCP) with per-task stop buttons. Ticks down as each completes.

**Background terminal** (`tools/terminal_tool.rs`): `background: bool` parameter (default false). When true, starts the command, shows live output in the tool card, returns "Running \`cmd\`..." to the model immediately, and spawns a detached waiter that updates the card on exit. The model doesn't get the command output; the tool description says so explicitly.

## Tests

5 integration tests in `tests/test_mcp_tasks.rs` using FakeTransport: basic lifecycle, failure, progress notifications, forbidden-tool fallback, and model-immediate-response with background poller verification.

Not committed: `script/test-tasks-mcp.py` (12-tool FastMCP test server) and `script/test-immediate-response-wire.py` (stdio wire format validator). Both useful for manual testing but not part of the PR.

## Not done

- **`tasks/list`**: wire types exist but nothing calls them. Could be used for reconnection (picking up tasks from a previous session).
- **`input_required` + elicitation**: spec allows tasks to pause and ask the client for input. Zed doesn't have elicitation yet. The test server has a placeholder tool for when it lands.
- **Client capabilities advertisement**: `ClientCapabilities.tasks` is left as `None`. Should be set once we're confident.
- **Feature flag**: the spec is experimental and could change. No flag gates it currently.
- **Final result delivery to model**: with model continuation, the model only sees the provisional response. A "deferred result" mechanism to inject the final output into a later turn would be useful but doesn't exist.

Release Notes:

- Added experimental support for the MCP Tasks protocol, enabling long-running MCP tool calls with live progress updates, per-task cancellation, and model continuation via `model-immediate-response`.
- Added `background` mode to the terminal tool for commands that don't need to block the model.

### DIFF
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -2357,6 +2357,24 @@ impl AcpThread {
         .boxed()
     }
 
+    pub fn cancel_tool_call(&mut self, tool_call_id: &acp::ToolCallId, cx: &mut Context<Self>) {
+        self.connection
+            .cancel_tool_call(&self.session_id, tool_call_id, cx);
+
+        if let Some((ix, call)) = self.tool_call_mut(tool_call_id) {
+            if matches!(
+                call.status,
+                ToolCallStatus::Pending
+                    | ToolCallStatus::WaitingForConfirmation { .. }
+                    | ToolCallStatus::InProgress
+            ) {
+                call.status = ToolCallStatus::Canceled;
+                cx.emit(AcpThreadEvent::EntryUpdated(ix));
+                cx.notify();
+            }
+        }
+    }
+
     pub fn cancel(&mut self, cx: &mut Context<Self>) -> Task<()> {
         let Some(turn) = self.running_turn.take() else {
             return Task::ready(());
--- a/crates/acp_thread/src/connection.rs
+++ b/crates/acp_thread/src/connection.rs
@@ -136,6 +136,16 @@ pub trait AgentConnection {
 
     fn cancel(&self, session_id: &acp::SessionId, cx: &mut App);
 
+    /// Cancel a single tool call without stopping the entire turn.
+    /// Default implementation is a no-op for connections that don't support it.
+    fn cancel_tool_call(
+        &self,
+        _session_id: &acp::SessionId,
+        _tool_call_id: &acp::ToolCallId,
+        _cx: &mut App,
+    ) {
+    }
+
     fn truncate(
         &self,
         _session_id: &acp::SessionId,
--- a/crates/agent/src/agent.rs
+++ b/crates/agent/src/agent.rs
@@ -1249,6 +1249,19 @@ impl NativeAgentConnection {
                             }
                             ThreadEvent::Stop(stop_reason) => {
                                 log::debug!("Assistant message complete: {:?}", stop_reason);
+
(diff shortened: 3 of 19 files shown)


dsturnbull:
Demo screencasts

https://github.com/user-attachments/assets/69d2fd85-f1ac-499a-901b-dac4f5c84429


https://github.com/user-attachments/assets/d43539d0-3816-4d71-be97-60ebda8b20d6

bennetbo:
Hey, thanks for working on this. We don't have a lot of users asking for MCP features, so right now it's not feasible for us to maintain this surface area going forward (the MCP feature is still experimental, so I expect more changes to come).
Also it seems like this PR includes other changes which I don't entirely understand by briefly looking over the diff.

PR #53728, opened 2026-04-12
Fix devcontainer localEnv/containerEnv substitution when variable is undefined

im-0:
Existing code replaces only the variables that are present in the environment. It ignores references to undefined variables and leaves them as is.

Consider following devcontainer.json:

```json
{
  ...
  "remoteEnv": {
    "RUSTFLAGS": "${localEnv:RUSTFLAGS}"
  },
  ...
}
```

Existing code will set RUSTFLAGS inside the dev container to a correct value only if RUSTFLAGS is defined outside the dev container. If RUSTFLAGS is not defined outside, Zed will erroneously set RUSTFLAGS to `${localEnv:RUSTFLAGS}`.

This patch fixes this by replacing such references with either empty strings or provided default values (`${localEnv:VARNAME:default}`).

This patch also removes replacement of `\` with `/` for environment variables. Existing code makes no sense as simple replacement will not magically make Windows paths valid on *nix. It is also guaranteed to break things: think about passing passwords via env vars, for example.

I also noticed that, for some reason, existing code serializes k-v maps to JSON, then tries to replace the references, and after that deserializes modified JSON back. I believe that this is horribly wrong and may have some security implications. This is out of the scope of this patch. And I hope that somebody from the Zed team will provide the reasoning behind this code.

Self-Review Checklist:

~~Closes #ISSUE~~ This pull request is also a bug report.

Release Notes:

- Fixed substitution of `${localEnv:VARNAME}` and `${containerEnv:VARNAME}` in devcontainer.json when variable is not defined

### DIFF
--- a/crates/dev_container/src/devcontainer_json.rs
+++ b/crates/dev_container/src/devcontainer_json.rs
@@ -237,14 +237,26 @@ pub(crate) struct DevContainer {
     host_requirements: Option<HostRequirements>,
 }
 
+pub(crate) fn deserialize_devcontainer_json_to_value(
+    json: &str,
+) -> Result<serde_json_lenient::Value, DevContainerError> {
+    serde_json_lenient::from_str(json).map_err(|e| {
+        log::error!("Unable to deserialize json values: {e}");
+        DevContainerError::DevContainerParseFailed
+    })
+}
+
+pub(crate) fn deserialize_devcontainer_json_from_value(
+    json: serde_json_lenient::Value,
+) -> Result<DevContainer, DevContainerError> {
+    serde_json_lenient::from_value(json).map_err(|e| {
+        log::error!("Unable to deserialize devcontainer from json values: {e}");
+        DevContainerError::DevContainerParseFailed
+    })
+}
+
 pub(crate) fn deserialize_devcontainer_json(json: &str) -> Result<DevContainer, DevContainerError> {
-    match serde_json_lenient::from_str(json) {
-        Ok(devcontainer) => Ok(devcontainer),
-        Err(e) => {
-            log::error!("Unable to deserialize devcontainer from json: {e}");
-            Err(DevContainerError::DevContainerParseFailed)
-        }
-    }
+    deserialize_devcontainer_json_to_value(json).and_then(deserialize_devcontainer_json_from_value)
 }
 
 impl DevContainer {
--- a/crates/dev_container/src/devcontainer_manifest.rs
+++ b/crates/dev_container/src/devcontainer_manifest.rs
@@ -132,7 +132,7 @@ impl DevContainerManifest {
     }
 
     fn parse_nonremote_vars_for_content(&self, content: &str) -> Result<String, DevContainerError> {
-        let mut replaced_content = content
+        let replaced_content = content
             .replace("${devcontainerId}", &self.devcontainer_id())
             .replace(
                 "${containerWorkspaceFolderBasename}",
@@ -154,11 +154,11 @@ impl DevContainerManifest {
                 "${localWorkspaceFolder}",
                 &self.local_workspace_folder().replace('\\', "/"),
             );
-        for (k, v) in &self.local_environment {
-            let find = format!("${{localEnv:{k}}}");
-            replaced_content = replaced_content.replace(&find, &v.replace('\\', "/"));
-        }
-
+        let replaced_content = Self::replace_environment_variables(
+            &replaced_content,
+            "localEnv",
+            &self.local_environment,
+        );
         Ok(replaced_content)
     }
 
@@ -179,16 +179,14 @@ impl DevContainerManifest {
         // HOME is user-specific, and we will often not run as the image user
         merged_remote_env.remove("HOME");
         if let Some(remote_env) = self.dev_container().remote_env.clone() {
-            let mut raw = serde_json_lenient::to_string(&remote_env).map_err(|e| {
+            let raw = serde_json_lenient::to_string(&remote_env).map_err(|e| {
                 log::error!(
                     "Unexpected error serializing dev container remote_env: {e} - {:?}",
                     remote_env
                 );
                 DevContainerError::DevContainerParseFailed
             })?;
-            for (k, v) in container_env {
-                raw = raw.replace(&format!("${{containerEnv:{k}}}"), v);
-            }
+            let raw = Self::replace_environment_variables(&raw, "containerEnv", &container_env);
             let reserialized: HashMap<String, String> = serde_json_lenient::from_str(&raw)
                 .map_err(|e| {
                     log::error!(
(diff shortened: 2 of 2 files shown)


im-0:
v0.232.2 is still affected.

crates/dev_container/src/devcontainer_manifest.rs
```diff
@@ -205,6 +203,50 @@ impl DevContainerManifest {
         Ok(merged_remote_env)
     }
 
+    fn replace_environment_variables(
```
KyleBarton, line 213: I think this might read better if we used a regex capture - something with named parts like `\$\{localEnv:(?P<variable>[^:}]+)(?::(?P<default>[^}]*))?\}`. What do you think?

crates/dev_container/src/devcontainer_manifest.rs
```diff
@@ -155,11 +155,11 @@ impl DevContainerManifest {
                 "${localWorkspaceFolder}",
                 &self.local_workspace_folder().replace('\\', "/"),
             );
-        for (k, v) in &self.local_environment {
-            let find = format!("${{localEnv:{k}}}");
-            replaced_content = replaced_content.replace(&find, &v.replace('\\', "/"));
```
KyleBarton, line 159: So I think we still need to handle the presence of `\` characters, because it is an escape character in JSON. I agree that replacing with forward slash is probably the wrong approach here. But I think replacing with `\\` would work here

KyleBarton (commented):
Hey, thanks for this! Left a few comments on implementation.

Also re:

> I also noticed that, for some reason, existing code serializes k-v maps to JSON, then > tries to replace the references, and after that deserializes modified JSON back. I 
> believe that this is horribly wrong and may have some security implications.

Can you expand on your security concerns here? I'm not sure I understand.

> And I hope that somebody from the Zed team will provide the reasoning behind this code.

Since these variables can be referenced anywhere in the devcontainer.json object, replacing and re-deserializing seems expeditious. The alternatives would be to perform replacements field-by-field (which I think would amount to the same work, just more verbose) or create a custom "printing" layer that performs the replacement just-in-time, which was complexity that didn't seem necessary. Open to feedback if you think there's a better approach though.

im-0:
> Can you expand on your security concerns here? I'm not sure I understand.

I have not reviewed your code thoroughly, but it looks like local environment variables referenced in devcontainer.json may affect how Zed parses devcontainer.json and add arbitrary new structures there. Like, sneak in a new `onCreateCommand` that does not exist in the original devcontainer.json. This is not catastrophic on its own, but pretty bad.

Also, environment values may contain backslashes and quotes and new lines that will break JSON structure and lead to an unexpected errors.

> Since these variables can be referenced anywhere in the devcontainer.json object, replacing and re-deserializing seems expeditious.

Not anywhere, but only on the value side. And only inside the existing JSON strings.

> Open to feedback if you think there's a better approach though.

I think it is possible to deserialize into `serde_json::Value`, then traverse it and replace all localEnv (and localWorkspaceFolder etc.) references inside any `Value::String`. Then deserialize modified `serde_json::Value` into the "final" struct. containerEnv replacement is trivial as you just need to do something like `remote_env.values_mut().for_each(...)`.

Another option is to create a simple String wrapper struct with custom deserializer that deserializes the inner String and then substitutes localEnv there. But this will probably require adding unwrapping all over the place.

crates/dev_container/src/devcontainer_manifest.rs
im-0: This seems much less readable tbh

crates/dev_container/src/devcontainer_manifest.rs
im-0: Also, `"` and `\n` are problematic, as I explained.

im-0:
> I think it is possible to deserialize into serde_json::Value, then traverse it and replace all localEnv (and localWorkspaceFolder etc.) references inside any Value::String. Then deserialize modified serde_json::Value into the "final" struct. containerEnv replacement is trivial as you just need to do something like remote_env.values_mut().for_each(...).

Tried to implement this ^, please see new commit.

crates/dev_container/src/devcontainer_manifest.rs
im-0: Thought about this a bit more. The proper solution is to implement some kind of generic parser for all `${}` references, I believe. It would catch `${}` and then branch into different cases (localEnv, containerEnv, other simpler references). Will require more code though.

KyleBarton: I support going in that direction. What actually worries me about the current approach you have here is the syntax on the caller side:
```
Self::replace_environment_variables(
                        string,
                        "${localEnv:",
                        &self.local_environment,
                    );
```

Because `"${localEnv:"` is a bit of a strange variable to pass in, and in the future, it wouldn't be obvious to devs how it works without inspecting the underlying implementation.

crates/dev_container/src/devcontainer_json.rs
```diff
@@ -237,16 +237,34 @@ pub(crate) struct DevContainer {
...
+            log::error!("Unable to deserialize json values: {e}");
             Err(DevContainerError::DevContainerParseFailed)
         }
     }
 }
 
+pub(crate) fn deserialize_devcontainer_json_from_value(
+    json: serde_json_lenient::Value,
+) -> Result<DevContainer, DevContainerError> {
+    match serde_json_lenient::from_value(json) {
```
KyleBarton: Nit-picking: Mind moving to `map_err()` syntax here in `deserialize_devcontainer_json_to_value`? It was an oversight that `deserialize_devcontainer_json` didn't do that before.

KyleBarton (commented):
Awesome, yes I really like this direction, thanks for the contribution. My one blocking concern is the syntax around `replace_environment_variables`, as mentioned in my comment below. A more generic `${}` parser would work, as you've suggested. But an alternative smaller change could be to just make the third argument a `prefix` string which the function handles, so that the caller can pass in:
```
Self::replace_environment_variables(
                        string,
                        "localEnv", // <-- a prefix which the function will then appropriately wrap in `${...`
                        &self.local_environment,
                    );
```

Let me know what you think!

crates/dev_container/src/devcontainer_json.rs
im-0: Changed both `deserialize_devcontainer_json_to_value()` and `deserialize_devcontainer_json_from_value()` to use `map_err()`.

im-0:
> But an alternative smaller change could be to just make the third argument a prefix string which the function handles, so that the caller can pass in:

Done. Now `replace_environment_variables()` accepts `environment_source` string and constructs the prefix internally:

```rust
        let prefix = format!("${{{environment_source}:");
```

im-0:
Rebased on current `main` to resolve a merge conflict.

KyleBarton (approved):

PR #53735, opened 2026-04-12
agent: Handle prompts without user message IDs

nathansobo:
## Summary

Fix a production crash (`core::option::expect_failed`) in `NativeAgentConnection::prompt` caused by a missing `UserMessageId`.

Three changes, one per commit:

1. **Stop panicking on missing ID** — the native prompt path now returns an error instead of calling `.expect()`.
2. **Always assign user message IDs** — `AcpThread::send` unconditionally assigns a `UserMessageId`. UI editability/rewindability is decoupled from `message.id.is_some()` and gated on truncate-support instead, since message identity and edit capability were incorrectly conflated.
3. **Tighten the trait** — `AgentConnection::prompt` now takes `UserMessageId` instead of `Option<UserMessageId>`, making the invariant a compile-time guarantee. The runtime rejection test is removed since `None` is no longer representable.

## ACP protocol impact

This is **not** an ACP protocol or schema change.

- `UserMessageId` is a Zed-client-side concept; it never crosses the wire.
- `acp_thread` is `publish = false` — no external consumers exist.
- The `AcpConnection` implementation in `agent_servers` ignores the ID parameter entirely (`_id`).
- No serialized request/response shapes changed.

## Testing

- `cargo test -p acp_thread test_send_assigns_message_id_without_truncate_support`
- `cargo test -p agent_ui test_rewind_views`
- `cargo test -p agent test_agent_connection`

CC @bennetbo

Release Notes:

- Fixed a crash when native agent prompts were triggered without a user message ID.


benbrandt:
@bennetbo let's make sure that we aren't using presence of message id anywhere... I vaguely remember we key "editability of messages" on this, but we should be using a capability for that anyway instead

benbrandt:
Thanks. Should have fixed this last time I came across it 🤦🏻

PR #53746, opened 2026-04-12
gpui_linux: Replace kde-blur protocol with ext-background-effect protocol

Suya1671:
Replaces the `kde-blur` protocol (which is now deprecated and being [removed from kde](https://invent.kde.org/plasma/kwin/-/commit/7f9e0aa1cd2c2ab4fea1097800893b6f780089eb)) with the new [ext-background-effect](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/ext-background-effect/ext-background-effect-v1.xml) protocol (which is supported by niri, new KDE releases, and more compositors over time)

# Current questions
- Currently, the PR completely removes the old KDE blur and replaces it. Should we keep the old kde-blur implementation and use it as a fallback?
- Since the protocol technically supports adding new capabilities and types of background effects, should zed track if the compositor supports the protocol, and only try blurring then? The protocol seems to imply that you can still use [set_blur_region](https://wayland.app/protocols/ext-background-effect-v1#ext_background_effect_surface_v1:request:set_blur_region), but the compositor will do nothing
- Since the protocol is generic, should it be checked when resizing to see if the window should be blurred before updating the blur region? Currently it doesn't do it, since it's implied (a note has been added to the code)
- I wonder if this could be expanded to where popovers (such as diagnostic popovers) can become their own wayland surfaces, and thus also be blurred (so a transparent warning popover/diagnostic can properly blur the text underneath for readability, rather than it having to be opaque). That sounds like a major refactor/out of scope though, not even sure if it's possible (I'd love to work on it though, if the team has any pointers on how to get started)

I'm keeping this PR as a draft until those points are figured out/commented on

Self-Review Checklist:

- [/] Tests cover the new/changed behavior
	- I do not know how to test this, and the old implementation wasn't tested
- [/] Performance impact has been considered and is acceptable
	- I don't see how this would affect performance

Release Notes:

- wayland: replace kde-blur protocol with ext-background-effect protocol


TheRedXD:
I thought I'd chime in with an opinion here as replies to the current questions:
1. kde-blur should probably be kept as fallback until it is clear ext-background-effect is adopted basically everywhere and there is absolutely no need for kde-blur anymore. I believe we're quite close to that being the case
2. If I understood correctly, no, because the whole point seems to be that ext-background-effect should be the one worrying about compositors not supporting it. gpui shouldn't have to worry about compositors not supporting it
3. Unsure, although I'm leaning towards yes
4. Should be possible, although as you said, a major refactor will probably be necessary to achieve that

jansol:
Bumping this as Plasma 6.7 is set to come out on 2026-06-16, about 4 weeks from now and it removes support kde-blur entirely in favour of ext-background-effect. Would be nice to land this before that happens.

I agree that keeping kde-blur as a fallback would be great, since many people will be stuck on older Plasma versions for a while (e.g. it's unclear if Ubuntu 26.04 LTS will update to 6.7 or stick with 6.6). But keeping the complexity of the code down is a very good argument for only keeping the general option. The Zed team will have to make a judgement call on this.

For now I think it's fine to just do the absolute minimum of changes to achieve parity with the kde-blur implementation, since ext-background-effect does not currently support more anyway. More detailed handling can be done once the protocol gets updated with more features.

Popovers are currently rendered directly into the main window's surface so it's not possible to apply the background effect protocol to them in the first place. Changing popovers to have their own wayland subsurfaces would make this possible but is a much bigger endeavour that's definitely out of scope for this change, and would require in depth discussion with the Zed team as it should probably be done for all platforms if it is done at all.

canmom:
CachyOS just updated to the latest KDE and blur mysteriously went away; glad to know the reason is known and a fix is in the works.

ognevny:
> glad to know the reason is known and a fix is in the works.

last comment was 1 month ago. unfortunately I don't think so

Suya1671:
> > glad to know the reason is known and a fix is in the works.
> 
> last comment was 1 month ago. unfortunately I don't think so

Hello. I'm waiting for a review from @probably-neb before continuing working on this. This fix does work:tm: but I want to implement it properly before marking as ready for review

canmom:
> > > glad to know the reason is known and a fix is in the works.
> > 
> > 
> > last comment was 1 month ago. unfortunately I don't think so
> 
> Hello. I'm waiting for a review from @probably-neb before continuing working on this. This fix does work:tm: but I want to implement it properly before marking as ready for review

Thank you for the update, and your work patching it!

SturdyFool10:
well, unfortunate news, I'm on CachyOS and I have decided to be the bell ringer: KDE6.7 is here and on it window blur is just completely busted and effectively NO-OPs turning a blurred window into a transparent one... at least it doesn't crash tho, I agree with the above that we should check for ext_background_effect_v1, and if it does not exist we should check with kde_blur, and if **that** doesn't work then we should either no-op as it currently does or ideally we should probably give the user a notification about it

SomeoneToIgnore:
To keep the PR queue cleaner, will close this PR for now, as it's a draft with no changes over 3 weeks.

Suya1671:
I kept this as a draft because I was waiting for a review from @probably-neb. Can a maintainer please re-open this as a draft and/or review this?

tomhoule:
> Currently, the PR completely removes the old KDE blur and replaces it. Should we keep the old kde-blur implementation and use it as a fallback?

Yes, we should keep the old one as a fallback since I assume it will take time for distros with older versions of KDE to update. We should have a comment with a link to this PR / https://invent.kde.org/plasma/kwin/-/commit/7f9e0aa1cd2c2ab4fea1097800893b6f780089eb so we have an idea of the time frame when we want to remove the old path in the future.

tomhoule:
> Since the protocol technically supports adding new capabilities and types of background effects, should zed track if the compositor supports the protocol, and only try blurring then? The protocol seems to imply that you can still use [set_blur_region](https://wayland.app/protocols/ext-background-effect-v1#ext_background_effect_surface_v1:request:set_blur_region), but the compositor will do nothing

No I think it's fine as-is. We already only create the object when the protocol global is available, and setting a blur region while the capability is inactive is explicitly allowed. KWin stops and resumes applying the effect as its Blur desktop effect is toggled, in my testing.

tomhoule:
It took me time to reproduce but I can now, on a recent KDE Plasma version (6.7.4). Can confirm we don't have a blur on main, and this PR restores it. It was hard to confirm because the blur effect was too strong on my machine (as configured in kwin Desktop Effects), so the background looked completely opaque when blur was turned on.

crates/gpui_linux/src/linux/wayland/window.rs
```diff
@@ -1289,6 +1289,29 @@ impl WaylandWindowStatePtr {
...
+            // If we have a background effects surface, update the blur region
+            // Currently, zed only uses background effects if blur is enabled,
+            // so it can always be updated
+            if let Some(background_effects) = &state.background_effects {
+                let region = state
+                    .globals
+                    .compositor
+                    .create_region(&state.globals.qh, ());
+
+                let bounds = state.window_bounds.map(|v| f32::from(v) as i32);
```
tomhoule, line 1302: Updating here is the correct thing to do, but these are the wrong coordinates. We want region-local coordinates. It should be possible to use `state.bounds.size`. Window bounds don't always start at `(0,0)`. And when you maximize the window, `window_bounds` retains its pre-maximization geometry. Dt's easy to see the visual bug if you start in a relatively small window, then you maximize, and you'll see a region with the size of the original window blurred, but the rest of the maximized window background isn't blurred.

tomhoule:
With the inline comment fixed and a fallback to the older protocol, I think we're ready to merge.

Suya1671:
Hey! Sorry for a bit of a late reply, but I'm currently busy with exams so I might not be able to finish this soon. I'll try though

tomhoule:
No pressure, it's open source work, so you don't owe anybody anything. I'll try to be responsive.

jansol:
Note that there is #59842 which is another attempt at solving this, which purports to keep the kde protocol as a fallback, and to fix the artifacts around rounded window corners as well.

abachmann:
> Hey! Sorry for a bit of a late reply, but I'm currently busy with exams so I might not be able to finish this soon. I'll try though

If it’s okay with you, I would take over. I’ve been waiting for this feature for quite a while. I would just add another commit to address the changes mentioned by @tomhoule , since it looks like this feature is pretty close to being finished.

Suya1671:
I'm fine with that :D I think maybe PR my branch and then I'll merge it in? No clue if that keeps both of us as authors though/for the blame

jansol:
> PR my branch and then I'll merge it in?

That is one good option, the changes will then be directly reflected in this PR. Forking this PR branch and opening a new PR with additional changes works too, but of course loses the discussion/context here.

Commit authorship is independent of PRs, that should Just Work as long as you manage the commits with git instead of, say, manually copypasting code between branches.

crates/gpui_linux/src/linux/wayland/window.rs
abachmann, line 1998: While testing, I discovered another issue. if I open Zed, maximize it, close it, and then start it again, it opens maximized, the blur effect is no longer applied and the background is completely blank.

After changing the region calculations from `state.window_bounds` to `state.bounds`, I could no longer reproduce the issue.

I provided a fix for this here:
https://github.com/Suya1671/zed/pull/22/changes/e97d63db3fe566f271b282dfd7fadb08d600a6c2#diff-921fb8ebd90ac15dc9426ecd3ada9a30865b95704e4c19455e8a23bae6370126L1998-R1998

Worth mentioning that this part is already on `main`, so I hope changing it doesn't have any unintended side effects.

abachmann:
> I'm fine with that :D I think maybe PR my branch and then I'll merge it in? No clue if that keeps both of us as authors though/for the blame

I create a PR: https://github.com/Suya1671/zed/pull/22.

Suya1671:
I'm not entirely sure if that merge conflict is correct, and I'm currently on windows. Can someone else test it?

abachmann:
@cla-bot check

crates/gpui_linux/src/linux/wayland/window.rs
abachmann, line 1159: I guess it should be like this:
```rust
let initial_configure = self.frame_loop.get() == FrameLoop::Unconfigured;
drop(state);

if initial_configure {
    self.frame();
} else {
    self.request_redraw();
    update_window(self.state.borrow_mut());
}
```

I’ll do a quick test today, just to make sure everything still works.

retrixe:
On GNOME 51.0, it seems like the blur implementation in this PR causes the shadow region around the window border to be blurred as well

image

It doesn't seem to happen with other applications I see on the GNOME bug tracker, so likely a bug in this PR

Suya1671:
> On GNOME 51.0, it seems like the blur implementation in this PR causes the shadow region around the window border to be blurred as well
> image
> 
> It doesn't seem to happen with other applications I see on the GNOME bug tracker, so likely a bug in this PR

oh this seems similar to ghostty, which has a [rant over here](https://github.com/pluiedev/ghostty/blob/27fd1c7788a3ebdf347b54c5bc9c1c329f175397/src/apprt/gtk/winproto/BlurRegion.zig#L36-L40) and actual implementation of removing the shadow region [here](https://github.com/pluiedev/ghostty/blob/27fd1c7788a3ebdf347b54c5bc9c1c329f175397/src/apprt/gtk/winproto/BlurRegion.zig#L58-L80). I'm currently on windows sadly, but I'll check this code out after finals if nobody else is giving it a shot.

retrixe:
Yeah, that seems to be it, the ghostty implementation does work correctly for me on GNOME 51.0 as well

Issue #40826, opened 2025-10-21
AI: Notification sound played whenever file is edited


michaelaflores:
### Summary

When `Play sound when agent done` option is on, there is also a sound played whenever the agent has edited a file.

### Description

Steps to reproduce:
1. Turn on the `Play sound when agent done` option
2. Direct an agent to edit a file. This is easiest to observe if the request involves editing multiple files.
3. Observe that the macOS notification chime plays when it has edited each file
4. Lastly, a different notification sound plays and a notification appears when it's actually finished its work

**Expected Behavior**: one single chime plays and notification displays
**Actual Behavior**: many chimes may play during editing

### Zed Version and System Specs

Zed: v0.208.6 (Zed) 
OS: macOS 26.0.1
Memory: 128 GiB
Architecture: aarch64

smitbarmase:
I was not able to reproduce it with "always_allow_tool_actions" set to false and "play_sound_when_agent_done" set to true. Can you share your settings.json? And what model and provider you are using?

michaelaflores:
@smitbarmase In my case `always_allow_tool_actions` is `true`, however it was unexpected that there's a implicit behavior at play based on Play sound when agent **done** that is about playing the sound when agent is "making progress" (aka editing a file, which may happen many times during a piece of work).

PR #53752, opened 2026-04-12
Add integrated terminal bell + settings

### LINKED ISSUE #5303: Support OS notifications long running tasks completing
**Is your feature request related to a problem? Please describe.**
Crazy pitch. Sometimes you need some _disposable_ terminals windows that are one shot and they complete fairly quickly. But there are situations in which you still have a one-shot command, but that action will actually take minutes (e.g. apply a terraform configuration). 

**Describe the solution you'd like**
Would be great if I could open a terminal modal, with the ability to minimize it. When running a long task, I would get a notification on its completion. Then I can run a new task or close the modal.

**Describe alternatives you've considered**
I could use the integrated terminal and just have a spare terminal there, but I got inspired by the current terminal modal. 😬

ian-h-chamberlain:
Self-Review Checklist:

- [x] ~Unsafe blocks (if any) have justifying comments~

Followup to #47531 to use the gpui feature in Zed. This just plumbs the "system bell" feature into the terminal, behind a new setting (enabled by default, like most other terminal applications).

Closes #5303
Relates to https://github.com/zed-industries/zed/issues/40826#issuecomment-3684556858

Release Notes:

- Added audible BEL to Terminal; can be enabled by setting `terminal.bell` to `"system"`.


assets/settings/default.json
```diff
@@ -1908,6 +1908,10 @@
     "path_hyperlink_timeout_ms": 1,
     // Whether to show a badge on the terminal panel icon with the count of open terminals.
     "show_count_badge": false,
+    "bell": {
+      // Whether to invoke the OS-default alert sound when the terminal bell (BEL character) is printed.
+      "system": true,
```
probably-neb: This settings structure doesn't really make sense as it is.
If we were to add different sound options the settings would look like
```json
"bell": {
  "system": false,
  "other-sound": true,
}
```
This should be
```jsonc
"bell": "off" // default
"bell": "system" // use system alert sound
```
or similar

probably-neb (changes requested):

assets/settings/default.json
ian-h-chamberlain: That's fair, I was thinking a struct/object made sense to allow for the possibility of visual bells or other indicators (like desktop notifications, toasts, whatever), as suggested in https://github.com/zed-industries/zed/pull/47531#issuecomment-4122687027

Also was looking at VSCode for inspiration here, they have an accessibility setting for announcing the terminal bell (probably via some text-to-speech or something): https://github.com/microsoft/vscode/blob/1faf9f4d1bab92b16b16cc6306f828a8f32567bf/src/vs/workbench/contrib/accessibility/browser/accessibilityConfiguration.ts#L497-L510

What do you think about something like this?
```jsonc
"bell": {
  "audio": "system", // or "off", etc.

  // possible future enhancements:
  "visual": "off",
  "notification": "on",
  "announce": "off"
}
```

probably-neb: Interesting. I think we should keep it simple, with no nested object for now. It's fairly easy to do a settings migration and add the extra structure if we ever add more capabilities. Until then it's just extra typing in the settings JSON, making the setting less discoverable.

assets/settings/default.json
```diff
@@ -1916,6 +1916,8 @@
     "path_hyperlink_timeout_ms": 1,
     // Whether to show a badge on the terminal panel icon with the count of open terminals.
     "show_count_badge": false,
+    // Whether to invoke the OS-specific alert sound when the terminal bell (BEL character) is printed.
+    "bell": "system",
```
probably-neb, line 1920: I think we should have the default be "off" to preserve behavior

probably-neb:
```suggestion
    "bell": "off",
```

probably-neb (approved):

crates/settings_content/src/terminal.rs
```diff
@@ -179,6 +179,10 @@ pub struct TerminalSettingsContent {
     ///
     /// Default: false
     pub show_count_badge: Option<bool>,
+    /// What to do when the `BEL` character (`\a`) is printed to terminal.
+    ///
+    /// Default: "system"
```
injust, line 184: This was missed when changing the default.

PR #53781, opened 2026-04-13
languages: Fix Python LSP workspace folder detection in uv workspaces

### LINKED ISSUE #47926: Pyright LSP using wrong pyproject.toml when using a uv workspace
### Reproduction steps

1. Create a UV workspace with multiple subproject. 
2. Open the workspace root in Zed
3. Stop the lsp.
4. Open a Python file from one of the subprojects first to let Pyright initialize.

Current behavior: Pyright is initialized with the subproject as workspace folder
Expected behavior:  Initialize Pyright with uv workspace root as the workspace folder.

The issue is probably in the `search` function in `PyprojectTomlManifestProvider`. It searches for the first `pyproject.toml` file recursively, so it ends up finding the subproject's `pyproject.toml` instead of the workspace `pyproject.toml`. Adding a check to see if that `pyproject.toml` is a member of a workspace can solve the issue. :

```rust
// zed/crates/languages/src/python.rs

impl ManifestProvider for PyprojectTomlManifestProvider {
    fn name(&self) -> ManifestName {
        SharedString::new_static("pyproject.toml").into()
    }

    fn search(
        &self,
        ManifestQuery {
            path,
            depth,
            delegate,
        }: ManifestQuery,
    ) -> Option<Arc<RelPath>> {
        for path in path.ancestors().take(depth) {
            let p = path.join(RelPath::unix("pyproject.toml").unwrap());
            if delegate.exists(&p, Some(false)) {
                return Some(path.into());                                 // ⬅️ HERE
            }
        }

        None
    }
}
```

Zed: 0.221.4
OS: macOS 26.2
Memory: 32 GB
Architecture: aarch64

### Attach Zed log file

relevant pyright lsp logs

```log
* Server: pyright (id 19)

* Version: Unknown

...

* Registered workspace folders:
/Users/username/Developer/my-workspace-project/my-subproject

...
```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

space192:
Closes #47926

  ## Summary

  When using a uv (or Poetry/PDM) workspace with multiple subprojects, Python LSP servers (Pyright, Ruff, ty, etc.) are initialized with the subproject directory as their workspace folder instead of the workspace root. This happens because `PyprojectTomlManifestProvider::search()` returns the first (innermost)
  `pyproject.toml` found walking up the directory tree.

  For example, in a uv workspace like:

  main-project/
  ├── pyproject.toml       # workspace root with [tool.uv.workspace]
  ├── uv.lock
  ├── packages/
  │   └── project-api/
  │       └── pyproject.toml   # subpackage
  └── projects/
      └── project-a/
          └── pyproject.toml   # subpackage

  Opening a file in `packages/project-api/` would register `packages/project-api/` as the LSP workspace folder instead of `main-project/`.

  ## Approach

  The fix uses lockfile existence as a heuristic to detect workspace roots. The updated `search()` method walks all ancestors (similar to `CargoManifestProvider`) and:

  - Tracks the **innermost** `pyproject.toml` as a fallback
  - Tracks the **outermost** `pyproject.toml` that has a sibling lockfile (`uv.lock`, `poetry.lock`, `pdm.lock`, or `Pipfile.lock`)
  - Returns the outermost workspace root if found, otherwise falls back to the innermost

  This works within the existing `ManifestDelegate` interface (existence checks only, no file content reading).

  | Scenario | Result |
  |---|---|
  | uv workspace (root `pyproject.toml` + `uv.lock`) | Returns workspace root |
  | Poetry workspace (root `pyproject.toml` + `poetry.lock`) | Returns workspace root |
  | Simple project (single `pyproject.toml`, no lockfile) | Returns project dir (unchanged) |
  | Independent subprojects (no lockfile at any level) | Returns each project's own dir (unchanged) |

  Since the manifest provider is set at the Python **language** level, this fix applies to all Python LSP servers (Pyright, Ruff, ty, etc.).

  ## Test plan

  - [x] Added unit tests for `PyprojectTomlManifestProvider` covering all scenarios above
  - [x] Existing integration test `test_running_multiple_instances_of_a_single_server_in_one_worktree` passes (independent subprojects without lockfiles)
  - [x] `cargo check -p languages` compiles cleanly
  - [x] Manual testing with a real uv workspace (Pyright and Ruff both receive correct workspace root)

Release Notes:

- Fixed Python LSP servers (Pyright, Ruff, etc.) using the wrong workspace folder in uv/Poetry/PDM workspaces with multiple subprojects.


osiewicz (approved):
Makes sense to me. Thanks!

PR #53782, opened 2026-04-13
editor: Align split buffer headers with scrollbar layout

cppcoffee:
Clip split buffer headers to the editor's content bounds so they no longer paint into the horizontal scrollbar area.

Reuse the editor's computed right margin and horizontal scrollbar visibility instead of reimplementing scrollbar detection in split_editor_view. This keeps split headers aligned when minimap, inline blame, or scrollbar visibility change the available content area.

Release Notes:

- Fixed horizontal scrollbars not being rendered behind buffer headers in split editors


dinocosta:
Hey @cppcoffee 👋 

Are you able to share some screenshots/recordings of what the before and after look like? Would probably make it easier to test these changes and see exactly what's changing in the UI. Thanks!

cppcoffee:
Horizontal scroll bar for operation.

before:

Uploading 录屏2026-04-14 09.24.56.mov…



after:


Uploading 录屏2026-04-14 10.21.59.mov…

cppcoffee:
The Editor horizontal scroll bar for operation.

before:

https://github.com/user-attachments/assets/a0b3e2ff-edf3-4ef5-b974-a8cd4fa53f19



after:

https://github.com/user-attachments/assets/b2f49c8e-195f-4915-b593-5a8a90c64219

dinocosta (approved):
Thank you @cppcoffee ! 🙂

I've merged `main` into your changes and updated accordingly, seeing as `Bounds::from_corner_and_size` no longer exists. There was also a case where the bug could still be triggered, in case the horizontal scrollbar was only visible on the left hand side editor, as only the right hand side editor was provided to `SplitBufferHeadersElement` and that editor's horizontal scrollbar visibility was driving the fix.

PR #53792, opened 2026-04-13
sidebar: Fix non-empty draft threads getting discarded on load

bennetbo:
When switching to another thread on the same project group a draft thread would just be discarded

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- N/A


crates/agent_ui/src/agent_panel.rs
eholk, line 440: Not a blocker for this PR, but we should consider refactoring this so there aren't so many parameters to this method.

eholk (approved):

PR #53797, opened 2026-04-13
agent_ui: Clarify disabled profile selector copy

itsfuad:
Self-Review Checklist:

After fix:

https://github.com/user-attachments/assets/7a9b7947-e222-4839-a46d-078b6aec8f18

Closes #53749

Release Notes:

- N/A


danilo-leal (approved):
Thank you! I just pushed a tweak there to not render the profile button when the model hasn't even been selected yet. We totally should keep displaying the "tools unsupported" disabled button because that's how we communicate that aspect of the selected model, but while you haven't yet selected one, I think we can streamline the UI. Thanks again :)

PR #53800, opened 2026-04-13
ci: Build nightly more frequently

MrSubidubi:
This updates the nightly workflow to run hourly as opposed to once daily. 

Notably, it also 
- changes the poll duration for the auto updater - we now poll 4 times an hour on the Nightly channel and continue with the old rate for all other channels
- moves the pre-checks to the linux runners, since we have more resources available there
- only ever allow one concurrent nightly run

Release Notes:

- N/A


osiewicz:
Mm, I'm not sure whether:
> only ever allow one concurrent nightly run - push nightly runs take precedence over scheduled ones, if there is a push for a new nightly, we cancel the in-progress last-push and any in-progress scheduled ones over that push nigthly. However, if we determine the current pushed tag sha is equal to the currently scheduled one, we let the scheduled workflow continue to run.

is the way to go; I'd err towards letting the current build finish and ignoring a dispatched one.

Also, with builds happening so frequently I feel like we should be less incentivized to trigger builds manually. Just wait a bit and bam, it's there anyways.

MrSubidubi:
Yeah that makes sense. Generally for me, I tend to bump nighly only ever if I really want to catch some issues, thus the idea here was to basically keep that urgency and allow people to skip the hourly wait if they e.g. want something in urgently after release day. Was basically for that scenario where one would probably want to dogfood this quicker than what the cronjob would allow. 

Happy to remove it, but then I'd also argue we prolly should remove the manual bumping altogether in favor of automated bumps, as I would then not see much value in keeping those around. Not sure though whether I'd like that though

SomeoneToIgnore:
While we're at it, can we add some automatic retries of the jobs on failure?
Those network issues during dependency installation are real still on Linux runners.

MrSubidubi:
> While we're at it, can we add some automatic retries of the jobs on failure?
Those network issues during dependency installation are real still on Linux runners.


I thought about that already today and will most likely add that to the GitHub Zippy bot, as we have more granular control there. I generally want to move more logic out of the actions, as all the scripting here is slowly getting out of hand IMO. But yeah, we should and made a note of it

ConradIrwin (approved):
Let's do it

osiewicz (approved):
just do it

Thank you!

Issue #9318, opened 2024-03-14
enable type on search by default


kzhui125:
Summary: Other editors search as you type, whereas in Zed project search only happens after you press enter.

VSCode: will search project when type every char

<img width="327" alt="image" src="https://github.com/zed-industries/zed/assets/11925432/2c3d4456-57c8-47a9-9578-2787bbba36a2">

Zed: have to press enter to search.

<img width="835" alt="image" src="https://github.com/zed-industries/zed/assets/11925432/959e9961-8ec4-44f9-8afd-10ce25173530">

failable:
Yes, I miss this interactive search feature in Emacs so much.

Moshyfawn:
This is already a feature request, but I'm going to close it in favour of this one due to the number of upvotes it's already received.

danielzsh:
Do we know if vscode implements any kind of string algorithm/optimization when a single character is typed or deleted (an obvious one that comes to mind is when adding a character to our search, we need only to filter through the results we already have rather than rerun on the entire codebase) and whether such optimizations would be significant enough to implement in zed as well?

stanislav-tkach:
For me personally the search is unusable without this feature. Both vscode and IDEs from jetbrains have it.

mgsloan:
This would be a really nice enhancement.  I think there should be a setting for this to

Tebro:
I am giving this a try, have something working already.

mgsloan:
On reviewing #24086 (thanks @Tebro!), I realized that adding this in a way that works well in all cases is not so straightforward - I probably shouldn't have marked this as "good first issue".  However, happy to help with adding this feature.

Specifically:

* Old results should stay visible until `N` lines of new results are loaded, to avoid them disappearing and reappearing.  `N` should be chosen so that it's typically at least the whole pane.

* Results on these live searches should be limited to `M` results for efficiency. For example, VS Code limits to 20000 results, but I think we might want a lower limit than this.

![image](https://github.com/user-attachments/assets/5dad1f80-75e8-4c4d-9713-55755a94969f)

Some UI choices to be made:

* What limit for live searches (`M`) to choose?

    - Option A: Same as N - relatively low - so that these live searches are just an efficient preview.  I prefer this option.

    - Option B: Large number like VS Code's choice of 20000

* When is the full search performed?

    - Option A: When focus enters the search results, and `enter` just causes focus to switch to the results, same as the behavior on search within a file.  I prefer this option.

[...]

PR #53801, opened 2026-04-13
Enable type on search by default for the project search

SomeoneToIgnore:
Another attempt to deal with https://github.com/zed-industries/zed/issues/9318

Release Notes:

- Enabled type on search by default for the project search

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -680,6 +680,8 @@
     "regex": false,
     // Whether to center the cursor on each search match when navigating.
     "center_on_match": false,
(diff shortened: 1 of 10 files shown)


osiewicz (approved):

osiewicz:
I test-drove this a fair bit and other than too lazy clearing of multi-buffer (which has since been addressed), I don't have any objections.

SomeoneToIgnore:
Based on internal feedback, need to reduce the flicker more, gutter resizes and excerpt preservation in particular.

PR #53808, opened 2026-04-13
gpui_wgpu: Fix WGPU panic from unsupported atlas texture formats

osyvokon:
We were assuming that `Bgra8Unorm` always works, which was causing the panic on configurations that don't support it ("wgpu-hal invariant was violated: Requested feature is not available on this device")

This change makes a proper fallback to `Rgba8Unorm`. In the worst case, if a device doesn't support that as well (highly unlikely), we return a proper error instead of a panic.

Self-Review Checklist:

Closes ZED-69S

Release Notes:

- Fixed panic on devices that don't support BGRA8


rtfeldman:
@zed-industries/approved

PR #48146, opened 2026-02-01
Fix duplicate window when opening from CLI on macOS

koh-sh:
Closes #47140
Closes #44691

When launching Zed from the CLI (`zed .`), macOS delivers the path as a `kAEGetURL` Apple Event via Launch Services. Zed relied on the `application:openURLs:` delegate method to receive this, but its timing relative to `applicationDidFinishLaunching:` is not guaranteed by macOS. In release builds, the app reaches `didFinishLaunching` before the URL is delivered, causing it to be missed and a duplicate window to be opened. This does not reproduce in debug builds due to slower initialization, which is why the issue was hard to reproduce from source.

This replaces `application:openURLs:` with a custom `kAEGetURL` Apple Event handler registered in `applicationWillFinishLaunching:`. Apple Events are guaranteed to be delivered synchronously between `willFinishLaunching` and `didFinishLaunching`, ensuring the URL is always available before startup logic runs.

 Release Notes:

 - Fixed duplicate window creation when opening files/directories from the CLI on macOS.

PR #53814, opened 2026-04-13
Revert "Fix duplicate window when opening from CLI on macOS (#48146)"

ConradIrwin:
This reverts commit 5be9dc1781ef6d2cbfbdbcd417edd0935cbb83a8.

This change breaks dragging directories onto the App icon (or using the
directories listed in the menu on the app icon in the doc)

Release Notes:

- N/A


cole-miller (approved):
Thanks!

PR #53822, opened 2026-04-13
gpui: Record input-to-frame latency histogram

eholk:
Adds instrumentation to track input-to-frame latency in GPUI windows, helping diagnose input responsiveness issues.

## What this does

- Records the time between when an input event is dispatched and when the resulting frame is presented, capturing worst-case latency when multiple events are coalesced into a single frame.
- Tracks how many input events get coalesced per rendered frame.
- Both metrics are stored in [HdrHistogram](https://docs.rs/hdrhistogram) instances with 3 significant digits of precision.
- Latency is only recorded when the input event actually causes a redraw (i.e. marks the window dirty), so idle mouse moves and other no-op events don't skew the data.
- Adds a `Dump Input Latency Histogram` command that opens a buffer with a formatted report including percentile breakdowns and visual distribution bars.

## Example output

The report shows percentile latencies, a bucketed distribution with bar charts, and a per-frame event coalescing breakdown.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/Cargo.toml
+++ b/crates/gpui/Cargo.toml
@@ -96,6 +96,7 @@ sum_tree.workspace = true
 taffy = "=0.9.0"
 thiserror.workspace = true
 gpui_util.workspace = true
+hdrhistogram.workspace = true
 waker-fn = "1.2.0"
 lyon = "1.0"
 pin-project = "1.1.10"
--- a/crates/gpui/Cargo.toml
+++ b/crates/gpui/Cargo.toml
@@ -38,6 +38,7 @@ screen-capture = [
     "scap",
 ]
 windows-manifest = ["dep:embed-resource"]
+input-latency-histogram = ["dep:hdrhistogram"]
 
 [lib]
 path = "src/gpui.rs"
@@ -96,7 +97,7 @@ sum_tree.workspace = true
 taffy = "=0.9.0"
 thiserror.workspace = true
 gpui_util.workspace = true
-hdrhistogram.workspace = true
+hdrhistogram = { workspace = true, optional = true }
 waker-fn = "1.2.0"
 lyon = "1.0"
 pin-project = "1.1.10"
--- a/crates/gpui/src/window.rs
+++ b/crates/gpui/src/window.rs
@@ -959,11 +959,15 @@ pub struct Window {
     /// Tracks recent input event timestamps to determine if input is arriving at a high rate.
     /// Used to selectively enable VRR optimization only when input rate exceeds 60fps.
     pub(crate) input_rate_tracker: Rc<RefCell<InputRateTracker>>,
-    /// Timestamp recorded at the start of dispatch_event; cleared when a frame is presented.
-    /// Used to compute input-to-frame latency.
-    last_input_at: Option<Instant>,
+    /// Timestamp of the first unrendered input event in the current frame; cleared when a frame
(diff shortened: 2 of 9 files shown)


crates/gpui/src/window.rs
```diff
@@ -958,6 +959,15 @@ pub struct Window {
...
     pub(crate) input_rate_tracker: Rc<RefCell<InputRateTracker>>,
+    /// Timestamp of the first unrendered input event in the current frame; cleared when a frame
+    /// is presented. Captures worst-case latency when multiple events are coalesced per frame.
+    first_input_at: Option<Instant>,
+    /// Count of input events received since the last frame was presented.
+    pending_input_count: u64,
+    /// Histogram of input-to-frame latency samples, in nanoseconds.
+    input_latency_histogram: Histogram<u64>,
+    /// Histogram of input events coalesced per rendered frame.
+    input_events_per_frame_histogram: Histogram<u64>,
```
Anthony-Eid: Should these fields be a singular struct?

crates/gpui/src/window.rs
```diff
@@ -2361,12 +2377,133 @@ impl Window {
...
+        if let Some(first_input_at) = self.first_input_at.take() {
+            let latency_nanos = first_input_at.elapsed().as_nanos() as u64;
+            self.input_latency_histogram.record(latency_nanos).ok();
+        }
+        if self.pending_input_count > 0 {
+            self.input_events_per_frame_histogram
+                .record(self.pending_input_count)
+                .ok();
+            self.pending_input_count = 0;
+        }
```
Anthony-Eid: Seems like this and `format_input_latency_report` could be methods if you make a `HistogramRecorder` struct

crates/gpui/src/window.rs
```diff
@@ -4113,6 +4250,7 @@ impl Window {
     /// Dispatch a mouse or keyboard event on the window.
     #[profiling::function]
     pub fn dispatch_event(&mut self, event: PlatformInput, cx: &mut App) -> DispatchEventResult {
+        let dispatch_time = Instant::now();
```
Anthony-Eid, line 4224: nit: You can make dispatch_time and optional and only get it if self.first_input_at.is_none()

crates/gpui/Cargo.toml
```diff
@@ -96,6 +96,7 @@ sum_tree.workspace = true
 taffy = "=0.9.0"
 thiserror.workspace = true
 gpui_util.workspace = true
+hdrhistogram.workspace = true
```
Anthony-Eid: Should this be a feature on gpui?

crates/gpui/src/window.rs
Anthony-Eid: I'm also wondering if we could combine some of this with our profiler task window

eholk: Good call. I combined the fields into a struct.

crates/gpui/src/window.rs
eholk: This should be done now too.

crates/gpui/Cargo.toml
eholk: Good idea. @cole-miller and I made this change during all hands pairing today.

crates/gpui/src/window.rs
eholk: > I'm also wondering if we could combine some of this with our profiler task window

I think I'll hold off on that for this PR but I'd like to learn more so let's talk about it.

PR #53824, opened 2026-04-13
Make UserMessageId non-optional to fix agent prompt crash

rtfeldman:
The `AgentConnection::prompt` trait method accepted `Option<UserMessageId>`, but `NativeAgentConnection::prompt` called `.expect()` on it, causing a crash (SIGABRT) when the id was `None`.

The id was `None` when `AcpThread::send()` found that `truncate()` returned `None` (session not found in `agent.sessions`), which caused it to skip generating a `UserMessageId`. This `None` then flowed through to `prompt()` and hit the `.expect()`.

This makes `UserMessageId` non-optional where it matters, eliminating the crash at the type level:

- `UserMessage.id` is now `UserMessageId` (not `Option<UserMessageId>`)
- `AgentConnection::prompt` takes `UserMessageId` (not `Option<UserMessageId>`)
- `AcpThread::send` always generates a `UserMessageId` — the only case where it previously didn't was the crash case (session missing), which now gracefully errors via the "Session not found" check in `prompt()` instead of panicking

`push_user_content_block` keeps taking `Option<UserMessageId>` because it's used to incrementally append content blocks, where callers either provide a specific ID (e.g. forwarding from Thread to AcpThread) or pass `None` to mean "keep the existing ID / auto-generate one". When creating a new message with `None`, it auto-generates via `UserMessageId::new()`.

Release Notes:

- Fixed a crash that could occur when sending a prompt to the AI agent

### DIFF
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -1414,7 +1414,7 @@ impl AcpThread {
                     .and_then(|entry| entry.user_message())
                     .is_some_and(|message| message.chunks.contains(&content));
                 if !already_in_user_message {
-                    self.push_user_content_block(UserMessageId::new(), content, cx);
+                    self.push_user_content_block(None, content, cx);
                 }
             }
             acp::SessionUpdate::AgentMessageChunk(acp::ContentChunk { content, .. }) => {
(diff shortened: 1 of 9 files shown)


bennetbo:
I think Eric fixed it in #53850

PR #53825, opened 2026-04-13
Fix cross-window file copy/paste in project panel

### LINKED ISSUE #11757: Copy paste of files between instances…+
### Check for existing issues

- [X] Completed

### Describe the feature

I would like to be able to copy paste files from the file tree between instances. 

Related to https://github.com/zed-industries/zed/issues/4983 and probably doesn't make much sense before that one is implemented first.

I guess this needs to tie in to os native features…?

This is doable in vscode:

https://github.com/zed-industries/zed/assets/116520/59128bb4-c7fa-4b42-b06d-bba9559106d1

kathbigra:
Self-Review Checklist:

Closes #11757

## Problem

Copying files from one Zed window's project panel and pasting into another window's project panel would fail silently. The files would appear to copy but wouldn't paste in the destination window.

## Root Cause

There were two bugs working together:

**1. project_panel wrote plain text instead of ExternalPaths**

`write_entries_to_system_clipboard` was serializing file paths as a newline-joined string. But `paste()` reads the clipboard looking for a `ClipboardEntry::ExternalPaths` entry — it never found one, fell through to the panel's internal `self.clipboard` state, which is `None` in the destination window (a different panel instance). Paste silently did nothing on all platforms.

**2. macOS and Windows had no-op ExternalPaths clipboard writers**

Even after fixing the write side, the macOS and Windows platform clipboard implementations had empty handlers for `ClipboardEntry::ExternalPaths`, so the data was never serialized to the system clipboard.

## Fix

**project_panel (all platforms):**
`write_entries_to_system_clipboard` now writes a `ClipboardEntry::ExternalPaths` item to the clipboard instead of a plain text string.

**macOS:**
- Implement `write_external_paths()` using `NSFilenamesPboardType` to properly serialize file paths to the system pasteboard.
- Safety fix: `NSString::UTF8String()` can return null if a path cannot be represented as UTF-8. The prior code passed this directly to `CStr::from_ptr()`, which is undefined behaviour. Added a null check to skip such paths.

**Windows:**
- Implement `write_paths()` using the `CF_HDROP` format with a `DROPFILES` structure so file paths are written to the system clipboard.
- Bug fix: `read_files()` was passing `locked.ptr` (the `GlobalLock`'d memory pointer) as the `HDROP` handle to `DragQueryFileW`. `DragQueryFileW` expects the `HGLOBAL` handle before locking, so reading files copied from external apps (e.g. Explorer) was always broken. Fixed to use `locked.global.0`.
- Bug fix: `write_paths()` was encoding paths with `to_string_lossy().encode_utf16()`, which replaces non-Unicode bytes with U+FFFD and can corrupt valid Windows paths. Fixed to use `path.as_os_str().encode_wide()` which encodes natively without loss.

**Linux (Wayland):**
Cross-window copy had an additional compositor-level obstacle. Most Wayland compositors do not send a `wl_data_device.selection` event back to the client that called `set_selection`. This meant `current_offer` was never updated after a self-write, causing `read()` to return `None` immediately even though the data was stored internally.

Fixed by adding an `is_self_owner` flag to the `Clipboard` struct. When Zed writes to the clipboard, `read()` returns `contents` directly without requiring `current_offer` to be set by the compositor. `is_self_owner` is cleared when the compositor notifies Zed that another app has taken clipboard ownership.

Additionally:
- `write_to_clipboard` now advertises `text/uri-list` in the Wayland data source when the item contains `ExternalPaths`, so external apps (file managers, etc.) can receive the copied paths.
- `send()` now serializes `ExternalPaths` as a RFC 2483 URI list when an external app requests `text/uri-list`.
- Added `read_uri_list()` to `DataOffer` so Zed can receive file paths copied from external apps via the clipboard (previously only possible via drag-and-drop).

**Linux (X11):**
No platform layer change was needed. X11's `read_from_clipboard()` already short-circuits via `is_owner()` and returns the cached `clipboard_item` for same-process reads. With the project_panel fix writing `ExternalPaths` instead of a string, cross-window paste works correctly.

## Release Notes:

- Fixed copying files from one Zed window's project panel and pasting them into another window's project panel on macOS, Windows, and Linux


crates/gpui_windows/src/clipboard.rs
```diff
@@ -268,6 +268,46 @@ fn read_files() -> Option<ClipboardEntry> {
...
+        let global = Owned::new(GlobalAlloc(GMEM_MOVEABLE, total_size)?);
+        let ptr = GlobalLock(*global) as *mut u8;
+        anyhow::ensure!(!ptr.is_null(), "GlobalLock returned null");
+
+        // Write DROPFILES header
+        std::ptr::write_unaligned(ptr as *mut u32, header_size as u32); // pFiles offset
+        std::ptr::write_unaligned(ptr.add(4) as *mut u32, 0); // pt.x
+        std::ptr::write_unaligned(ptr.add(8) as *mut u32, 0); // pt.y
+        std::ptr::write_unaligned(ptr.add(12) as *mut u32, 0); // fNC
+        std::ptr::write_unaligned(ptr.add(16) as *mut u32, 1); // fWide (Unicode)
```
cameron1024: These consecutive writes seem suspicious. Could we use a struct instead? Note that it would have to be `#[repr(C)]` to guarantee correct layout;
```rust
#[repr(C)]
struct Header {
  header_size: u32,
  // ...
}
```
Then you can use `write_unaligned` with an instance of this struct. You can also replace the `header_size` variable with `std::mem::size_of::<Header>()`.

crates/gpui_windows/src/clipboard.rs
```diff
@@ -268,6 +268,46 @@ fn read_files() -> Option<ClipboardEntry> {
...
+    let header_size = 20usize;
+    let total_size = header_size + wide_paths.len() * std::mem::size_of::<u16>();
+
+    unsafe {
+        let global = Owned::new(GlobalAlloc(GMEM_MOVEABLE, total_size)?);
+        let ptr = GlobalLock(*global) as *mut u8;
+        anyhow::ensure!(!ptr.is_null(), "GlobalLock returned null");
+
+        // Write DROPFILES header
+        std::ptr::write_unaligned(ptr as *mut u32, header_size as u32); // pFiles offset
```
cameron1024: A small comment explaining why unaligned writes are needed would also be helpful, even if only for reviewers

crates/gpui_windows/src/clipboard.rs
```diff
@@ -268,6 +268,46 @@ fn read_files() -> Option<ClipboardEntry> {
     )))
 }
 
+fn write_files(paths: &ExternalPaths) -> Result<()> {
```
cameron1024: The name of this function threw me off - I was expecting it to do something analagous to `std::fs::write`. Maybe a name like `copy_paths_to_clipboard` would be clearer? Unless I'm misunderstanding what the function does 😅

cameron1024 (changes requested):
Hi, thanks for the contribution. The overall goals of the PR seem solid.

I have some questions about the windows implementation though. I've left specific comments, but I don't have a mac so I can't review that part.

crates/gpui_macos/src/pasteboard.rs
```diff
@@ -253,6 +256,57 @@ impl Pasteboard {
                 .setData_forType(bytes, Into::<UTType>::into(image.format).inner_mut());
         }
     }
+
+    unsafe fn write_external_paths(&self, paths: &ExternalPaths) {
```
rtfeldman, line 263: @kathbigra FYI, I revised this to handle invalid paths and document why we should use this Cocoa API even though it's deprecated (namely, because the modern alternative requires specifying whether the paths are directories, and we don't know the answer to that in our case).

kathbigra:
@rtfeldman , i will be able to go through the comments and incorporate them after 2 days.

crates/gpui_windows/src/clipboard.rs
kathbigra: Done

crates/gpui_windows/src/clipboard.rs
kathbigra: Done

crates/gpui_windows/src/clipboard.rs
kathbigra: @cameron1024 
We have write_string, write_image already. To keep the naming convention as it is i am renaming it to "write_paths" not "copy_paths_to_clipboard".

Hope thats a little bit more clear to understand and still follows whats happening already in the file.

crates/gpui_macos/src/pasteboard.rs
kathbigra: Thanks @rtfeldman

kathbigra:
@cameron1024 I have incorporated your change requests. Can you review this again?

aizigao:
Looking forward to this feature

kathbigra:
@rtfeldman @cameron1024 

Is there anything remaining in this PR?
Can you guys help merge this?

@cameron1024 I have incorporated the changes you recommended. Can you review this again and close your change request?

mirahost:
Any updates on this basic feature working?

pcfreak30:
@kathbigra is there any reason Linux isn't included in this PR. Im on linux mint, powered by xorg, and this impacts nix as well.

aizigao:
need this too

kathbigra:
@pcfreak. based on this: https://github.com/kathbigra/zed/blob/139954071524f166762db38bcfb5381931f62b86/crates/gpui_linux/src/linux/wayland/client.rs#L964

I was assuming that linux would work fine. BUt looking at in detail. The fix would have only worked fro x11 and now wayland. Have added the fix for linux as well now. 

Lets wait for it to be approved now

kathbigra:
@cameron1024 @rtfeldman 
A gentle reminder to see if this can be merged. We have lots of requests fo rthis feature. Both here and in the original ticket. And lots of other related tickets

nm-z:
I need this as well, this seems trivial.

PR #53833, opened 2026-04-13
sidebar: Propagate workspace trust when creating new worktrees

cameron1024:
When creating a new git worktree, if the workspace is currently trusted, trust the newly-created worktrees

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -3145,7 +3145,7 @@ impl AgentPanel {
                         .has_restricted_worktrees(&source_worktree_store, cx)
                 })
                 .unwrap_or(false);
-            
+
             if !source_is_trusted {
                 return;
             }
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -64,6 +64,7 @@ use language::LanguageRegistry;
 use language_model::LanguageModelRegistry;
 use project::git_store::{GitStoreEvent, RepositoryEvent};
 use project::project_settings::ProjectSettings;
+use project::trusted_worktrees::{PathTrust, TrustedWorktrees};
 use project::{Project, ProjectPath, Worktree, WorktreePaths, linked_worktree_short_name};
 use prompt_store::{PromptStore, UserPromptId};
 use remote::RemoteConnectionOptions;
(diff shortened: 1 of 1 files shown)


crates/agent_ui/src/agent_panel.rs
```diff
@@ -3641,6 +3675,10 @@ impl AgentPanel {
             })
             .await;
 
+        if propagate_trust {
+            Self::trust_worktree_paths(&new_workspace, &all_paths, cx);
```
SomeoneToIgnore: NIT: instead of passing `propagate_trust` a bunch of times, we can check for that inside `trust_worktree_paths` it seems, as both `Self` and `&self` in the new code are for the same struct?

crates/agent_ui/src/agent_panel.rs
```diff
@@ -3125,6 +3122,40 @@ impl AgentPanel {
         }
     }
 
+    fn should_propagate_worktree_trust(&self, cx: &App) -> bool {
+        if ProjectSettings::get_global(cx).session.trust_all_worktrees {
```
SomeoneToIgnore: Nice idea to use `false` for the autotrust, as the user might trust something very odd that they would not want persisted when switched back to safe mode.

Yet, the overall code seems somewhat tricky to repeat in each module separately?
Seems safer to put it into the worktree trust's `fn trust` instead?

crates/agent_ui/src/agent_panel.rs
```diff
@@ -3125,6 +3122,40 @@ impl AgentPanel {
...
+        cx: &mut AsyncWindowContext,
+    ) {
+        cx.update(|_, cx| {
+            let Some(trusted_store) = TrustedWorktrees::try_get_global(cx) else {
+                return;
+            };
+            let worktree_store = new_workspace.read(cx).project().read(cx).worktree_store();
+            let paths_to_trust: collections::HashSet<PathTrust> = paths
+                .iter()
+                .map(|p| PathTrust::AbsPath(p.clone()))
```
SomeoneToIgnore: Can those `all_paths` be somehow resolved against the worktree store?
Overall, there was a push to keep the usage of raw paths to minimum (we could add some docs on that if it helps) to avoid potential collisions and vulnerabilities.

The original intent of this variant was to hold user's "trust all projects in the parent directory" trust action, hence we autotrust all child directories too in this case.

`PathTrust::Worktree` limits the trust to the worktree itself, which seems overall much safer for autogenerated LLM-based things?

SomeoneToIgnore (commented):
I have a concern over `PathTrust::AbsPath` usage, and a couple of optional nits.

crates/agent_ui/src/agent_panel.rs
cameron1024: Double checking: you're suggesting that `fn trust` should have this check for `trust_all_worktrees`, and so if we ever attempt to trust something in this mode, it's a silent no-op?

SomeoneToIgnore: Yes, no be sure we do not add something odd into database automatically.

crates/agent_ui/src/agent_panel.rs
SomeoneToIgnore: Thanks for updating it: just to double check, is there a situation possible when we have a valid path but no worktree found?
Maybe then we can fall back to the abs path, but I do not know the logic well enough to know how possible (and needed) this situation is.

SomeoneToIgnore (approved):
Thank you, the rest of the comments are closer to optional ideas than to issues.

PR #53839, opened 2026-04-13
Add feature flag for CLI first-run prompt

eholk:
Gates the CLI first-run prompt (from #53663) on the `agent-v2` feature flag, giving us an escape hatch if something goes wrong.

## What this does

**Commit 1: Gate on agent-v2 flag**
- Adds an `AgentV2FeatureFlag` check at the top of the open behavior resolution. When the flag is off, the prompt is skipped.
- Refactors `maybe_prompt_open_behavior` → `resolve_open_behavior`: moves the explicit-flags guard (`-n`, `-e`, `--reuse`) to the caller and drops three parameters from the function signature, since the function was only inspecting those values to decide whether to bail early.

**Commit 2: Default to new-window when flag is disabled**
- When the flag is off, falls back to the pre-#53254 behavior of opening projects in a new window, rather than silently adding to the sidebar.

Release Notes:

- N/A

### DIFF
--- a/crates/zed/src/zed/open_listener.rs
+++ b/crates/zed/src/zed/open_listener.rs
@@ -12,6 +12,7 @@ use futures::channel::mpsc::{UnboundedReceiver, UnboundedSender};
 use futures::channel::{mpsc, oneshot};
 use futures::future;
 
+use feature_flags::FeatureFlagAppExt as _;
 use futures::{FutureExt, StreamExt};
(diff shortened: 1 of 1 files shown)


nathansobo:
Should this just be tied to the same feature flag as the rest of the parallel agent features?

nathansobo (approved):
LGTM except my question about whether we need an independent flag.

PR #53859, opened 2026-04-14
Route open_thread through load_agent_thread to prevent duplicate sessions

rtfeldman:
`AgentPanel::open_thread` (called when clicking `@thread` mention links, pasting thread links, or opening threads via the CLI) was calling `external_thread` directly, bypassing the de-duplication guards in `load_agent_thread`. This meant clicking a thread link to the currently-active thread would create a redundant `ConversationView` sharing the same `NativeAgent` session. When the old CV was later evicted from `background_threads`, its `on_release` called `close_all_sessions`, removing the session and breaking the surviving CV with "Session not found".

Fix this by routing `open_thread` through `load_agent_thread`, which checks the active view, retained threads, and background threads before creating a new `ConversationView`. This prevents the duplicate from being created in the first place.

Release Notes:

- Fixed a "Session not found" error that could occur after clicking a thread mention link.

### DIFF
--- a/crates/agent/src/agent.rs
+++ b/crates/agent/src/agent.rs
@@ -3082,6 +3082,97 @@ mod internal_tests {
         })
     }
 
+    #[gpui::test]
+    async fn test_close_session_does_not_remove_reused_session(cx: &mut TestAppContext) {
+        // Regression test: when two ConversationViews share the same session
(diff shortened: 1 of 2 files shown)


mikayla-maki:
@zed-industries/approved

PR #53861, opened 2026-04-14
sidebar: Click project header row to toggle fold instead of switching worktrees

nathansobo:
Previously, clicking a project header in the agent sidebar would activate/switch to that worktree's workspace, resuming whatever thread was most recently active. This felt jarring — activating a random thread based on recency isn't that useful when the user can just click the specific thread they want. There's also no clear semantic for clicking the overall group header since each thread maps to a specific workspace and there's no single workspace to activate for the group.

This changes the click target for folding from just the chevron to the entire header row. The worktree-switching behavior is removed — clicking anywhere on the row now toggles the fold, which is a more predictable and useful interaction.

The hover-only buttons on the right side (ellipsis menu, collapse threads, new thread) continue to work as before, with their own click handlers that stop propagation.

Release Notes:

- Improved sidebar project headers to toggle fold on click instead of switching worktrees.

### DIFF
--- a/crates/sidebar/src/sidebar.rs
+++ b/crates/sidebar/src/sidebar.rs
@@ -1806,9 +1806,6 @@ impl Sidebar {
                             move |this, _, window, cx| {
                                 this.set_group_expanded(&key, true, cx);
                                 this.selection = None;
-                                // If the active workspace belongs to this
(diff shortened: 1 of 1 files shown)


danilo-leal:
@zed-zippy approve

PR #53862, opened 2026-04-14
Stop automatically deleting agent drafts

nathansobo:
This was leftover from trying to hold the invariant of one empty draft, which was dropped in #53852.

Removes all automatic draft deletion:

- `retain_running_thread` discarded empty drafts when navigating away
- `remove_empty_draft` cleaned up all retained drafts when loading from history
- `open_external_thread_with_server` removed empty drafts when opening external threads

Release Notes:

- Fixed agent drafts being silently discarded when navigating between threads


bennetbo:
@zed-zippy approved

PR #53867, opened 2026-04-14
Move agent threads to destination workspace on worktree switch

nathansobo:
The active agent thread now follows you when you switch worktrees. Previously the conversation stayed on the old workspace and you'd start fresh in the new one.

Screenshot 2026-04-14 at 12 21 17 AM

## How it works

**Workspace layer:** `MultiWorkspace` now emits an `ActiveWorkspaceChanged` event that includes why the switch happened (worktree switch vs. other). `Workspace` gains an `activate_with_cause` method so the worktree picker can signal that intent. The sidebar worktree picker uses this when switching.

**Agent panel:** `AgentPanel` subscribes to `MultiWorkspaceEvent`. On a worktree switch, it takes the active conversation from the source panel, replaces it with a fresh draft, and spawns an async task to adopt it in the destination panel. The destination panel's agent panel is loaded if it doesn't exist yet. `reveal_panel` is called outside the panel update to avoid a re-entrant entity update.

**Session rebinding:** `ConversationView::set_workspace` cascades through the thread view, entry view state, message editors, and mention set to rebind them to the new workspace and project. It calls `connection.set_project()` which flows through to `NativeAgent::rebind_session_to_project`, updating the session's project context, action log, and tools.

**Tool rebinding:** All stateful tools store their `Entity<Project>` behind a `Mutex` so `set_project()` can swap it after a workspace switch. Without this, tools like read_file, edit_file, terminal, etc. would keep operating against the old project.

CC @mikayla-maki @maxbrunsfeld

Release Notes:

- N/A


crates/acp_thread/src/connection.rs
```diff
@@ -134,6 +134,15 @@ pub trait AgentConnection {
         None
     }
 
+    fn set_project(
+        &self,
+        _session_id: &acp::SessionId,
+        _project: Entity<Project>,
+        _cx: &mut App,
+    ) -> Result<()> {
+        Ok(())
```
benbrandt: I am not sure we can silently `Ok(())` here. 
If an agent doesn't support moving directories, and we no-op here, then it will still be happily running in the _old_ directory.

We should either gate this behind a capability, or think a bit more about how we want to model this for zed agent vs others

benbrandt: Hmm even more interesting.. this is on the "connection" level not the session level... 

_Maybe_ this is ok then... but we need to take care not to update external agent metadata projects, they will just inherit the new project for the next session?

nathansobo:
Closing this for now — moving agent threads across workspaces on worktree switch isn't viable because ACP agents don't support this, so it's not a path we can pursue at this point.

PR #53887, opened 2026-04-14
sidebar: Consistent thread display time

cameron1024:
Standardizes various places in the sidebar to use the same time values (i.e. the last time the thread received user interaction, or else the last-updated time in the metadata if that's `None`)

Release Notes:

### DIFF
--- a/crates/sidebar/src/sidebar.rs
+++ b/crates/sidebar/src/sidebar.rs
@@ -3493,8 +3493,7 @@ impl Sidebar {
                     }?;
                     let notified = self.contents.is_thread_notified(&thread.metadata.thread_id);
                     let timestamp: SharedString =
-                        format_history_entry_timestamp(self.display_time(&thread.metadata))
-                            .into();
+                        format_history_entry_timestamp(self.display_time(&thread.metadata)).into();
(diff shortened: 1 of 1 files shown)


crates/sidebar/src/sidebar.rs
```diff
@@ -3475,6 +3455,13 @@ impl Sidebar {
             .insert(*thread_id, Utc::now());
     }
 
+    fn display_time(&self, metadata: &ThreadMetadata) -> DateTime<Utc> {
+        self.thread_last_message_sent_or_queued
+            .get(&metadata.thread_id)
+            .copied()
+            .unwrap_or(metadata.updated_at)
```
bennetbo, line 3462: Previously we were using `created_at` > `updated_at`. Is this right? Will this cause ordering issues?

cameron1024: We were using a weird mix of both - the old logic first checked `thread_last_message_sent_or_queued`, which is only set for threads which have been focused during this session (as part of the tab switcher). So you'd end up with threads switching order when you reload the workspace

bennetbo (approved):

PR #53889, opened 2026-04-14
Add support for user configurable mouse bindings

### LINKED ISSUE #10647: User configurable mouse bindings (like keymap for key+mouse)
Support for user-defined mappings of (for example):

- `mouse1` - left click
- `mouse2` - right click
- `mouse3` - toggle scroll, paste, jump to definition
- `mouse3 drag` - multi-cursor
- `alt-mouse1` - open url (match vscode linux)
- `alt-mouse1` - columnular selection (see: [`COLUMNAR_SELECTION_MODIFIERS`](https://github.com/zed-industries/zed/blob/fc262feb3ccdb7cfb987b83e75f0eae8b726adb9/crates/editor/src/editor.rs#L246-L252))
- `alt-shift-mouse1` - multi-cursor
- `shift-mouse1` - close tab
- `ctrl-mouse1` - right click (macos default)
- `ctrl-mousescrolldown` - Font size increase 
- `ctrl-mousescrollup` - Font size decrease
- `alt-mousescrolldown` - Fast scroll down
- `alt-mousescrollup` - Fast scroll up
- `force-click` - GotoDefinition (Jetbrains default)
etc.

Related issues:
- https://github.com/zed-industries/zed/issues/13295
- https://github.com/zed-industries/zed/issues/11990
- https://github.com/zed-industries/zed/issues/11786
- https://github.com/zed-industries/zed/issues/9745
- https://github.com/zed-industries/zed/issues/14612
- https://github.com/zed-industries/zed/issues/4644

OmChillure:
### Summary

- Added `MouseStroke` and `ScrollStroke` as first-class binding input types alongside keystrokes, with full modifier + click-count support
- Extended `KeyBinding` / `BindingInputKind` so the keymap can hold mouse and scroll bindings, and wired `Keymap` + `DispatchTree` to look them up by context
- `Window` now probes mouse and scroll bindings before the normal event chain: `MouseDown` fires matching actions, `MouseUp` is suppressed only when the same button+modifiers would have triggered a binding (preserving drag), and `ScrollWheel` fires only when a modifier is held so unmodified scrolling is unaffected
- `keymap_file.rs` auto-detects whether a binding string is a mouse stroke, scroll stroke, or keystroke and routes it to the right loader — both `bindings` and `unbind` entries work
- Added five editor actions (`AddCursorAtMouse`, `ExtendSelectionToMouse`, `GoToDefinitionAtMouse`, `GoToTypeDefinitionAtMouse`, `StartColumnarSelectionAtMouse`) so users can bind mouse buttons to common editor operations

### Example keymap entries

```json
[
  {
    "bindings": {
      "alt-mouse1":       "editor::GoToDefinitionAtMouse",
      "shift-mouse1":     "editor::ExtendSelectionToMouse",
      "ctrl-mouse1":      "editor::AddCursorAtMouse",
      "double-mouse1":    "editor::SelectToBeginningOfLine",
      "ctrl-mousescrolldown": "zed::IncreaseBufferFontSize",
      "ctrl-mousescrollup": "zed::DecreaseBufferFontSize",
    }
  }
]
```

### Self-Review Checklist

### Video:

[Screencast from 2026-04-14 17-02-06.webm](https://github.com/user-attachments/assets/53ac895d-fe75-4cca-af1b-6cc1b4ef5956)

Release Notes
Added support for user-configurable mouse button and scroll wheel keybindings in keymap.json

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -946,3 +946,28 @@ pub struct ConfirmEditReviewComment {
 pub struct CancelEditReviewComment {
     pub id: usize,
 }
+
+/// Adds a cursor at the current mouse position (for use in mouse keybindings).
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct AddCursorAtMouse;
+
+/// Extends the current selection to the mouse position (for use in mouse keybindings).
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct ExtendSelectionToMouse;
+
+/// Goes to the definition of the symbol at the current mouse position.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct GoToDefinitionAtMouse;
+
+/// Goes to the type definition of the symbol at the current mouse position.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct GoToTypeDefinitionAtMouse;
+
+/// Starts a columnar (block) selection at the current mouse position.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct StartColumnarSelectionAtMouse;
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -18293,6 +18293,116 @@ impl Editor {
         self.go_to_definition_of_kind(GotoDefinitionKind::Type, true, window, cx)
     }
 
+    fn mouse_position_map(&self, window: &Window) -> Option<(Rc<PositionMap>, PointForPosition)> {
+        let position_map = self.last_position_map.clone()?;
+        let mouse_position = window.mouse_position();
+        if !position_map.text_hitbox.contains(&mouse_position) {
+            return None;
+        }
+        let point = position_map.point_for_position(mouse_position);
+        Some((position_map, point))
+    }
+
+    fn goto_at_mouse(
+        &mut self,
+        kind: GotoDefinitionKind,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let Some((position_map, point_for_position)) = self.mouse_position_map(window) else {
+            return;
+        };
+        let display_snapshot = &position_map.snapshot.display_snapshot;
+        let point = point_for_position.previous_valid.to_point(display_snapshot);
+        let anchor = display_snapshot.buffer_snapshot().anchor_before(point);
+        self.change_selections(SelectionEffects::default(), window, cx, |s| {
+            s.select_anchor_ranges([anchor..anchor]);
+        });
+        self.select(SelectPhase::End, window, cx);
+        self.go_to_definition_of_kind(kind, false, window, cx)
+            .detach_and_log_err(cx);
+    }
+
+    pub fn add_cursor_at_mouse(
+        &mut self,
+        _: &AddCursorAtMouse,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let Some((_, point_for_position)) = self.mouse_position_map(window) else {
+            return;
+        };
+        self.select(
+            SelectPhase::Begin {
+                position: point_for_position.previous_valid,
+                add: true,
+                click_count: 1,
+            },
+            window,
+            cx,
+        );
+        self.select(SelectPhase::End, window, cx);
+    }
+
+    pub fn extend_selection_to_mouse(
+        &mut self,
+        _: &ExtendSelectionToMouse,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let Some((_, point_for_position)) = self.mouse_position_map(window) else {
+            return;
+        };
+        self.select(
+            SelectPhase::Extend {
+                position: point_for_position.previous_valid,
+                click_count: 1,
+            },
+            window,
+            cx,
+        );
+        self.select(SelectPhase::End, window, cx);
+    }
+
+    pub fn go_to_definition_at_mouse(
+        &mut self,
+        _: &GoToDefinitionAtMouse,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        self.goto_at_mouse(GotoDefinitionKind::Symbol, window, cx);
+    }
+
+    pub fn go_to_type_definition_at_mouse(
+        &mut self,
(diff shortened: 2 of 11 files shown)


OmChillure:
## Supported Mouse & Scroll Bindings

### Mouse buttons
| Token | Button |
|-------|--------|
| `mouse1` | Left click |
| `mouse2` | Right click |
| `mouse3` | Middle click |
| `mouse4` | Navigate back |
| `mouse5` | Navigate forward |

### Click counts
| Token | Count |
|-------|-------|
| *(none)* | Single click |
| `double` | Double click |
| `triple` | Triple click |

### Scroll directions
| Token | Direction |
|-------|-----------|
| `scroll-up` | Scroll up |
| `scroll-down` | Scroll down |

### Modifiers (all platforms)
`ctrl`, `alt`, `shift`, `fn`, `cmd` / `super` / `win`

---

### Syntax


### Examples

```jsonr
"alt-mouse1":        "editor::GoToDefinitionAtMouse"
"shift-mouse1":      "editor::ExtendSelectionToMouse"
"ctrl-mouse1":       "editor::AddCursorAtMouse"
"double-mouse1":     "editor::StartColumnarSelectionAtMouse"
"ctrl-scroll-up":    "zed::IncreaseBufferFontSize"
"ctrl-mousescrollup": "zed::IncreaseBufferFontSize"
```

If this pr seems to get traction form the devs to merge can add mroe keybindings to it as follow up

ConradIrwin:
Thanks for this.

It looks like this is using the keyboard focus for dispatching mouse events, but I if we're going to support mouse actions, they should be based on the dispatch tree at the cursor position instead.

If we are going to add this, are we able to rebuild any of our existing mouse handling to use actions, or is this a strictly additive thing? A lot of mouse handlers today just do the same as a keyboard event; and it'd be more compelling (for me at least) to consider this if we could use it to tidy up our code.

I'm on the fence about whether we should have this at all (it makes sense for people with a large number of mouse buttons), but I think overall it doesn't quite feel right like this.

OmChillure:
> Thanks for this.
> 
> It looks like this is using the keyboard focus for dispatching mouse events, but I if we're going to support mouse actions, they should be based on the dispatch tree at the cursor position instead.
> 
> If we are going to add this, are we able to rebuild any of our existing mouse handling to use actions, or is this a strictly additive thing? A lot of mouse handlers today just do the same as a keyboard event; and it'd be more compelling (for me at least) to consider this if we could use it to tidy up our code.
> 
> I'm on the fence about whether we should have this at all (it makes sense for people with a large number of mouse buttons), but I think overall it doesn't quite feel right like this.

Thanks for the thoughtful review.

**On dispatch source**: You're right, routing through the keyboard-focus dispatch tree is wrong. Mouse actions should resolve against the dispatch tree at the cursor's hit-tested element, so a binding like alt-mouse1 on the editor fires when the cursor is over editor content regardless of what's focused. I'll rework the probe in Window to walk the dispatch tree from the element under the cursor.

**On replacing existing handlers**: Agreed this is more compelling as a cleanup than as a parallel system. From a quick scan, the obvious candidates are the modifier-aware click paths in the editor (alt/shift/ctrl + click for go-to-definition, extend-selection, add-cursor, columnar-selection) — those are exactly the five actions I added, and today they live as bespoke mouse handlers. Migrating them onto the action dispatch would let the keymap own the mappings and remove the hardcoded modifier checks. Ctrl+scroll for font size is another good candidate. I'd want to scope it to handlers that genuinely mirror keyboard semantics and leave drag/hover behavior alone.

**On whether to have it at all**: Totally fair to be on the fence. If the refactor angle makes it worth pursuing, I'd be happy to open a new PR built around that framing => cursor-position dispatch + migrating the existing modifier-click handlers in the editor as the first concrete win, with the additive "extra mouse buttons" use case falling out for free. Want me to go ahead and open that, or any other suggestions on this?

ConradIrwin:
I think it'd be interesting to explore, but I can't guarantee that we'd merge the end result.

To me the questions that always come to mind are:
* Does it add significant maintenance burden (yes)
* Does it add more kinds of codepaths to test (yes in this PR, but fewer if we're using this functionality for click handling generally)

As I suspect most people won't use the ability to customize mouse shortcuts, we'd have to find a way to do this that feels well encapsulated (and relatively small).

One thing I think might help with that is making the buttons use the existing key based types (right now this PR introduces a number of parallel types for handling mouse events that are very similar to the key events - maybe we could unify those?) but it's hard to know without spending time with the code. (because it might make things more confusing..)

OmChillure:
> I think it'd be interesting to explore, but I can't guarantee that we'd merge the end result.
> 
> To me the questions that always come to mind are:
> 
> * Does it add significant maintenance burden (yes)
> * Does it add more kinds of codepaths to test (yes in this PR, but fewer if we're using this functionality for click handling generally)
> 
> As I suspect most people won't use the ability to customize mouse shortcuts, we'd have to find a way to do this that feels well encapsulated (and relatively small).
> 
> One thing I think might help with that is making the buttons use the existing key based types (right now this PR introduces a number of parallel types for handling mouse events that are very similar to the key events - maybe we could unify those?) but it's hard to know without spending time with the code. (because it might make things more confusing..)

Worth grounding this in prior art. Vim has done unified mouse-as-keymap — <LeftMouse>, <C-LeftMouse>, <2-LeftMouse>, <ScrollWheelUp> all flow through the same :map machinery as keyboard bindings, with bindings overriding default handling. 

My proposal: extend Keystroke.key to accept reserved values like "mouse1" / "mousescrolldown", with click count as a prefix ("double-mouse1" — already in this PR, same pattern as Vim's <2-LeftMouse>). Deletes the parallel MouseStroke / ScrollStroke types and their dispatch paths.

My approach would be:
- Unify into Keystroke.
- Cursor hit-test dispatch instead of keyboard focus.

- On migrating existing modifier-click handlers — I dug in and there are two real constraints (multi_cursor_modifier also gates hover-link behavior which can't live in a keymap, and the inline code carries click_count through to extend-by-word/line which my actions hardcode to 1). Worth treating that cleanup as a follow-up once the foundation lands and we've designed the click-count plumbing. Happy to go deeper on either if useful.

Want me to spike the 2-step version ? or what do you suggest?

ConradIrwin:
Sounds good! Also happy if you figure out other learnings along the way.

Currently in the codebase we use left/right/middle/forward/back rather than the numbers (and don't treat scroll buttons as click); so "leftclick" might be a bit more evocative than "mouse1",2,3,6,7; but no strong opinion on what to call these things.

OmChillure:
> Sounds good! Also happy if you figure out other learnings along the way.
> 
> Currently in the codebase we use left/right/middle/forward/back rather than the numbers (and don't treat scroll buttons as click); so "leftclick" might be a bit more evocative than "mouse1",2,3,6,7; but no strong opinion on what to call these things.

Hey , opened a PR at https://github.com/zed-industries/zed/pull/55216 with these changes and some design questions, hope you can check it.

Thank you

PR #53894, opened 2026-04-14
acp: Initial support for ACP usage

benbrandt:
Adds initial beta test of ACP usage stats. Behind a flag for now while we work on standardizing the usage values.

Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,6 +14,7 @@ dependencies = [
  "chrono",
  "collections",
  "env_logger 0.11.8",
+ "feature_flags",
  "file_icons",
  "futures 0.3.32",
  "gpui",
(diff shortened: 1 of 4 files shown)


bennetbo (approved):

gb-jos:
@bennetbo was interesting in using this. Any timeline for when it may be available ?

bennetbo:
@gb-jos It should already be available in Stable as of 1.7.2, but the underlying ACP agent actually needs to emit those usage updates. What agent are you using? E.g. Claude Agent supports it already:
image

gb-jos:
@bennetbo well yes, usage_update emit works and it shows in UI. However zed agent currently also takes detailed data from usage in prompt response, and show per turn usage and also separate input/output tracking. This is still gated behind the feature flag (https://github.com/gb-jos/zed/blob/main/crates/acp_thread/src/acp_thread.rs#L2979) and what i was interested in. I am using a custom acp adapter that i am writing for pi.

benbrandt:
@gb-jos thanks for trying it out! But I think that one in particular is going to need a rework at the RFD level which is why we only stabilized the context window part.

PR #53915, opened 2026-04-14
cli: Defer app activation until after open-behavior prompt

eholk:
On macOS, the CLI uses `LSOpenFromURLSpec` to deliver the `zed-cli://` URL to the running Zed app. With `kLSLaunchDefaults`, Launch Services activates (brings to foreground) the Zed app as a side effect of URL delivery. This happens before the IPC handshake completes and before the user can answer the terminal prompt asking whether to open in a new window or existing workspace, forcing them to manually switch back to the terminal.

## Fix

**CLI side** (`crates/cli/src/main.rs`): Add `kLSLaunchDontSwitch` to the launch flags so macOS delivers the URL without activating the app.

**App side** (`crates/zed/src/zed/open_listener.rs`): Add `cx.activate(true)` in `handle_cli_connection` after the prompt resolves (or immediately for the URL-only path), so the app comes to the foreground at the right time.

Linux and Windows are unaffected — they use Unix sockets and named pipes respectively, which don't have activation side effects.

Release Notes:

- Fixed the Zed CLI activating the app window before the user answers the open-behavior prompt in the terminal.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1217,7 +1217,9 @@ mod mac_os {
         string::kCFStringEncodingUTF8,
         url::{CFURL, CFURLCreateWithBytes},
     };
-    use core_services::{LSLaunchURLSpec, LSOpenFromURLSpec, kLSLaunchDefaults};
+    use core_services::{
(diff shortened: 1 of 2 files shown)


mikayla-maki:
@zed-zippy approve

PR #53924, opened 2026-04-14
cli: Hide -r/--reuse from --help in favor of -e/--existing

eholk:
Soft-deprecates the `-r`/`--reuse` CLI flag by hiding it from `--help` output. The flag remains fully functional for backward compatibility, but new users will discover `-e`/`--existing` instead.

The change is a single `hide = true` addition to the clap arg attribute.

Release Notes:

- N/A


bennetbo:
@zed-industries/approved

PR #53927, opened 2026-04-14
Improve error handling in worktree restore and archive paths

rtfeldman:
Improves the worktree restore and archive code paths:

- In `restore_worktree_via_git`, the old code blindly attempted `change_branch` and fell back to `create_branch` on failure, which could clobber a branch that had moved to a different SHA during archival. Now resolves the branch ref first and only checks out the branch if it still points at the original commit; if the branch has moved, the worktree stays in detached HEAD. If the branch doesn't exist, tries to recreate it.
- In `change_worktree_paths` and `change_worktree_paths_by_main`, archived threads are now excluded from path mutations so their folder paths are preserved for restore.
- In the sidebar, `ProjectGroupKey` construction for archived thread activation now uses `from_worktree_paths` (which resolves to main worktree paths) instead of `new` with the linked worktree's folder paths. This ensures that restoring a linked worktree thread matches the main repo workspace rather than creating a spurious new one.
- Added `Repository::resolve_ref` for resolving a git ref to its SHA.
- Added tests for `resolve_ref`, branch-moved/not-moved/deleted restore scenarios, and workspace-matching on restore.

Release Notes:

- N/A


eholk:
@zed-industries/approved

Issue #15166, opened 2024-07-25
Excessive display server repaints from missing damage/present regions


kennylevinsen:
When zed draws/presents a window, it uses full damage/no present regions, causing the display server to repaint the entire window on every single paint. This wastes significant GPU resources and power, as this is not just a copy.

Damage/present regions specify which portion of an application surface/layer/window is invalidated and needs to be redrawn by the display server from their latest presented image/buffer. How to do this depends on the platform API used:

- Vulkan: The [VK_KHR_incremental_present](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_incremental_present.html) extension provides [VkPresentRegionsKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPresentRegionsKHR.html), an extension to `VkPresentInfoKHR`
- EGL: [EXT_swap_buffers_with_damage](https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_swap_buffers_with_damage.txt) provides eglSwapBuffersWithDamageEXT as replacement for `eglSwapBuffers`.
- Metal: [CALayer::setNeedsDisplay](https://developer.apple.com/documentation/quartzcore/calayer/1410800-setneedsdisplay) (setNeedsDisplayInRect for Obj-C) as a separate path for communicating the need to redraw.

For Wayland on Linux, the Mesa Vulkan WSI is used through a vendored blade crate, which would rely on `VkPresentRegionsKHR` to submit Wayland damage.

Zed will also need to track which regions it has changed. The most important area for fine-grained damage reporting is of course cursor and buffer updates which affect very few pixels at a time, and are also relatively easy to track. When fine-grained tracking is not possible, coarse tracking erring on the side of too much damage should be the next goal, with full surface damage being the fallback for scenarios where it's too hard to track (or too rare a scenario to matter).

Damage reporting is relatively easy to inspect on Wayland by looking at the protocol messages:

```
WAYLAND_DEBUG=1 ./path/to/zed-editor 2>&1 | grep damage
```

However, all platforms should benefit from damage tracking, especially if the display server performs color management, HDR mapping or high detail blending in FP16 intermediate buffers.

### Environment

Zed: v0.145.1 (Zed)
OS: Linux Wayland arch unknown
Memory: 61.5 GiB
Architecture: x86_64

fatihaziz:
anything we could do to reduce or make this issue gone without doing any PR or changes on the zed code?

[...]

PR #53932, opened 2026-04-14
gpui: Track and forward damage regions

loadingalias:
Part 2/2 for #15166 fix.
The companion PR is in WGPU (v29): https://github.com/zed-industries/wgpu/pull/2

## Summary
• add coarse damage tracking to GPUI w/ `DamageRegion` plus `Context::notify_with_damage(bounds)`, while keeping `notify()` as the conservative full damage fallback
• forward `damage: Option<Bounds<DevicePixels>>` through `PlatformWindow::draw` so each backend can either consume damage explicitly or safely ignore it
• make the Linux / `wgpu` path call `SurfaceTexture::present_with_damage()` and update the editor hot paths that matter here: caret blink, caret movement, selections, hover popovers, and coarse editor area updates for buffer and display map changes
• teach the macOS Metal renderer to keep a persistent render target and use scissoring for partial redraws, reducing local fragment work without depending on compositor present region support
• preserve GPUI's existing invalidation semantics for accessed entities; damage remains an explicit optimization hint and does not narrow redraw scheduling rules
• add regression coverage for observer driven non-view invalidation, present only frames retaining computed damage, and the editor behaviors that regressed when invalidation semantics were narrowed

### Why
The issue here was not just that Zed repainted too much. It was that the editor had no end-to-end path to express small visual changes as damage and carry that information down to presentation.

This patch fixes that in the minimal places that matter. Zed now computes meaningful damage for cursor blink, cursor movement, selections, hover state, and edits. GPUI can carry that damage without changing its redraw contract. Linux can forward it into the companion `wgpu` present-with-damage path. macOS can reduce local render work with scissoring even though Metal does not have compositor facing present regions.

## Testing
Local (MacOS):
- `cargo fmt --all --check`
- `./script/clippy -p gpui`
- `./script/clippy -p gpui_macos`
- `./script/clippy -p editor`
- `cargo test -p gpui window::tests::test_observed_non_view_notify_reaches_the_view -- --exact --nocapture`
- `cargo test -p gpui window::tests::test_present_keeps_pending_damage_for_presentation_only_frames -- --exact --nocapture`
- `cargo test -p agent_ui message_editor::tests::test_autoscroll_after_insert_selections -- --exact --nocapture`
- `cargo test -p vim change_list::test::test_change_list_insert -- --exact --nocapture`

Manual verification still **required** before claiming full closure of `#15166`:
- Linux / Wayland protocol proof showing narrowed `damage_buffer` output during idle caret blink
- Windows no-regression verification

## Notes
• This PR depends on the companion `wgpu` PR and temporarily pins to WGPU commit `bb070fec2ca45625855581ba894a6cb0b55e2105` on `loadingalias/wgpu` through `[patch.'https://github.com/zed-industries/wgpu.git']`.
• Zed now computes and forwards meaningful damage correctly; however, it really should have the Linux/Wayland tests done and Windows regressions checks run.

Release Notes:

- Improved rendering efficiency by tracking and forwarding damage regions for small editor updates such as caret blink, caret movement, selections, hover popovers, and edits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10790,7 +10790,7 @@ dependencies = [
 [[package]]
 name = "naga"
 version = "29.0.0"
-source = "git+https://github.com/zed-industries/wgpu.git?branch=v29#a466bc382ea747f8e1ac810efdb6dcd49a514575"
+source = "git+https://github.com/loadingalias/wgpu.git?rev=bb070fec2ca45625855581ba894a6cb0b55e2105#bb070fec2ca45625855581ba894a6cb0b55e2105"
 dependencies = [
  "arrayvec",
  "bit-set 0.9.1",
@@ -20252,7 +20252,7 @@ checksum = "a751b3277700db47d3e574514de2eced5e54dc8a5436a3bf7a0b248b2cee16f3"
 [[package]]
 name = "wgpu"
 version = "29.0.0"
-source = "git+https://github.com/zed-industries/wgpu.git?branch=v29#a466bc382ea747f8e1ac810efdb6dcd49a514575"
+source = "git+https://github.com/loadingalias/wgpu.git?rev=bb070fec2ca45625855581ba894a6cb0b55e2105#bb070fec2ca45625855581ba894a6cb0b55e2105"
 dependencies = [
  "arrayvec",
  "bitflags 2.10.0",
@@ -20263,7 +20263,7 @@ dependencies = [
  "hashbrown 0.16.1",
  "js-sys",
  "log",
- "naga 29.0.0 (git+https://github.com/zed-industries/wgpu.git?branch=v29)",
+ "naga 29.0.0 (git+https://github.com/loadingalias/wgpu.git?rev=bb070fec2ca45625855581ba894a6cb0b55e2105)",
  "parking_lot",
  "portable-atomic",
  "profiling",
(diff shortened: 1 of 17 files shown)


kennylevinsen:
Did a quick test of this with the wgpu changes on Wayland. Two observations:
1. All testing I did always lead to `damage=None` by the time we reached `WaylandWindow::draw`, reporting full damage on every frame.
2. If I make `WindowInvalidator::invalidate_view` inert rather than always committing full damage, I do see damage propagating through the entire stack, which does seem close but not entirely right (inserting text into the middle of a string or scrolling the sidebar does not carry correct damage for example) 

I did not dig further into it, just poked the PR with a stick a bit.

loadingalias:
> Did a quick test of this with the wgpu changes on Wayland. Two observations:
> 
> 1. All testing I did always lead to `damage=None` by the time we reached `WaylandWindow::draw`, reporting full damage on every frame.
> 2. If I make `WindowInvalidator::invalidate_view` inert rather than always committing full damage, I do see damage propagating through the entire stack, which does seem close but not entirely right (inserting text into the middle of a string or scrolling the sidebar does not carry correct damage for example)
> 
> I did not dig further into it, just poked the PR with a stick a bit.

Thank you so much. Let me have a look. Anything else you can share is really appreciated!

ChristopherBiscardi:
Thanks for looking into this, however since the PR relies on patches to wgpu and we're trying to return to upstream (as indicated in https://github.com/zed-industries/wgpu/pull/2#issuecomment-4864349037 ) I'm going to close it. There is some damage tracking related work going on in https://github.com/zed-industries/zed/pull/59461 as well

PR #53939, opened 2026-04-14
cli: Fix -n behavior and refactor open options

eholk:
This fixes a regression where `zed -n .` in a subdirectory of an already-open
project would redirect to the parent window instead of creating a new one.
The root cause was that commit 66d2cb20c9 ("Adjust `zed -n` behavior") made
`-n` run the worktree matching loop with subdirectory matching enabled, when
previously `-n` skipped matching entirely.

### Bug fix
- **Restore `-n` to always create a new window.** No worktree matching, no
  exceptions. This matches the behavior from when `-n` was first introduced.

### New `--classic` flag
- Adds a hidden `--classic` CLI flag that explicitly selects the pre-sidebar
  default behavior: new window for directories, reuse existing window for
  files already in an open worktree.
- The `cli_default_open_behavior` setting now toggles between `-e` (add to
  sidebar) and `--classic` behavior. When set to `new_window`, the classic
  logic is used instead of unconditionally opening a new window.

### Refactor CLI open options
Replaces the old grab-bag of `open_new_workspace: Option<bool>`,
`force_existing_window: bool`, `classic: bool`, and `reuse: bool` with:

- **`cli::CliOpenBehavior` enum** — a single enum on the IPC boundary with
  variants `Default`, `AlwaysNew`, `Add`, `ExistingWindow`, `Classic`, and
  `Reuse`.
- **`workspace::WorkspaceMatching` enum** — describes how to match paths
  against existing worktrees (`None`, `MatchExact`, `MatchSubdirectory`).
- **`workspace::OpenOptions`** — uses `WorkspaceMatching` plus a simple
  `add_dirs_to_sidebar: bool` instead of overlapping boolean flags.

The translation from CLI enum to workspace options happens in
`open_listener.rs`, keeping both layers clean and independent.

Release Notes:

- N/A

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -145,9 +145,10 @@
   // an explicit `-e` (existing window) or `-n` (new window) flag.
   //
   // May take 2 values:
-  //  1. Add to the existing Zed window
+  //  1. Open directories as a new workspace in the current Zed window's sidebar
   //         "cli_default_open_behavior": "existing_window"
-  //  2. Open a new Zed window
+  //  2. Open directories in a new window (reuse existing windows for files
+  //     that are already part of an open project)
   //         "cli_default_open_behavior": "new_window"
   "cli_default_open_behavior": "existing_window",
   // Whether to attempt to restore previous file's state when opening it again.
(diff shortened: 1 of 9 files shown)


benbrandt:
@zed-zippy approved

PR #53941, opened 2026-04-14
agent_ui: Improve the new thread worktree UX

danilo-leal:
Closes https://github.com/zed-industries/zed/issues/53262

- Remove the ability to pick a branch from the agent panel; delegate this to the title bar picker
- Make the worktree creation earger, just as you selected whether you want to create it from main or current branch
- Remove flicker when creating a new worktree and switching to a previously existing one
- Improve some UI stuff: how we display that a worktree is creating/loading, the branch and worktree icons, etc.
- Fixed a bug where worktrees in a detached HEAD state wouldn't show up in the worktree pickers

A big part of the diff of this PR is the removal of everything involved with the `StartThreadIn` enum/the set up involved in only creating the worktree by the time of the first prompt send.

Release Notes:

- Agent: Improved and simplified the UX of creating threads in Git worktrees.
- Git: Fixed a bug where worktrees in a detached HEAD state wouldn't show up in the worktree picker.


nathansobo (approved):
I pushed a cleanup commit (8b18822) with the following changes:

- **Renamed actions**: `CreateWorktreeImmediately` → `CreateWorktree`, `SwitchToLinkedWorktree` → `SwitchWorktree`. The "Immediately" suffix was a vestige of the old deferred creation path, and "ToLinked" is an implementation detail the action name doesn't need.
- **Removed dead code**: The `FirstSendRequested` event variant and its handler were no longer emitted after this PR removed the send interception. Also removed the no-op git store subscription and the unnecessary `#[allow(dead_code)]` on `WorktreeCreationStatus`.
- **Fixed disabled tooltips**: `create_new_list_item` had a hardcoded "Requires a Git repository" tooltip for all disabled entries. `CreateNamed` entries now show their specific reason ("A worktree with this name already exists" or "Cannot create in multi-repo project").
- **Consistent create entry visibility**: When typing a query that exactly matches an existing worktree name, the create entry now always appears (disabled) rather than being hidden. This is consistent with the space-normalized case that was already showing a disabled entry.
- **Explicit `OpenMode` parameter**: `find_or_create_workspace` / `find_or_create_local_workspace` now take an explicit `open_mode: OpenMode` parameter instead of inferring `Add` vs `Activate` from `init.is_some()`. Those are independent concerns.
- **Added picker tests**: 3 tests for `ThreadWorktreePickerDelegate` covering empty query entry building, fuzzy filtering with exact/space-normalized disabled entries, and multi-repo restrictions.

PR #53942, opened 2026-04-15
Add range-based whitespace and newline removal to buffer formatting

G36maid:
## Summary

When formatting specific ranges (e.g. via Format Selections), trailing whitespace removal and final newline enforcement currently operate on the entire file. This PR makes these operations range-aware, so they only affect lines within the target ranges.

As noted in #16509, users want `remove_trailing_whitespace_on_save` and `ensure_final_newline_on_save` to suppress jitter by only touching changed lines. This PR provides the foundational infrastructure for that — range-based whitespace and newline operations — which a follow-up PR will wire into the `format_on_save` modifications mode.

## Changes

- Added `remove_trailing_whitespace_in_ranges` and `ensure_final_newline_in_range` to `Buffer`
- When a `FormattableBuffer` has ranges set, whitespace and newline operations now use the range-based variants
- `format_ranges_via_lsp` returns `Ok(None)` instead of an error when the LSP doesn't support range formatting
- Extracted `anchor_ranges_to_row_ranges` helper to deduplicate conversion logic

<details><summary>Tests (4)</summary>

- `test_trailing_whitespace_in_ranges` — only modified lines have whitespace removed
- `test_trailing_whitespace_empty_ranges` — empty ranges → no changes
- `test_final_newline_modified_last_line` — newline added when last line is in range
- `test_final_newline_unmodified_last_line` — newline not added when last line is outside range

</details>

---

Self-Review Checklist:

Release Notes:

- Improved Format Selection to scope trailing whitespace removal and final newline enforcement to the selected range instead of the entire file
- Improved handling of LSP servers that do not support range formatting by gracefully skipping instead of producing an error

### DIFF
--- a/crates/language/src/buffer.rs
+++ b/crates/language/src/buffer.rs
@@ -2285,6 +2285,88 @@ impl Buffer {
         self.edit([(offset..len, "\n")], None, cx);
     }
 
+    /// Like [`remove_trailing_whitespace`](Self::remove_trailing_whitespace), but only removes
+    /// trailing whitespace on lines whose row falls within one of the given `modified_rows` ranges.
+    pub fn remove_trailing_whitespace_in_ranges(
+        &self,
+        modified_rows: &[Range<u32>],
+        cx: &App,
+    ) -> Task<Diff> {
+        let old_text = self.as_rope().clone();
+        let line_ending = self.line_ending();
+        let base_version = self.version();
+        let modified_rows = modified_rows.to_vec();
+        cx.background_spawn(async move {
+            let all_ranges = trailing_whitespace_ranges(&old_text);
+            let empty = Arc::<str>::from("");
+            let edits = all_ranges
+                .into_iter()
+                .filter(|range| {
+                    let row = old_text.offset_to_point(range.start).row;
+                    modified_rows.iter().any(|modified| modified.contains(&row))
+                })
+                .map(|range| (range, empty.clone()))
+                .collect();
+            Diff {
+                base_version,
+                line_ending,
+                edits,
+            }
+        })
+    }
+
+    /// Like [`ensure_final_newline`](Self::ensure_final_newline), but only applies the fix when
+    /// the last line's row falls within one of the given `modified_rows` ranges. Returns an empty
+    /// [`Diff`] when the last line is not in any modified range.
+    pub fn ensure_final_newline_in_range(
+        &self,
+        modified_rows: &[Range<u32>],
+        cx: &App,
+    ) -> Task<Diff> {
+        let len = self.len();
+        let line_ending = self.line_ending();
+        let base_version = self.version();
+        let last_row = self.max_point().row;
+
+        let last_row_is_modified = modified_rows.iter().any(|r| r.contains(&last_row));
+        if len == 0 || !last_row_is_modified {
+            return Task::ready(Diff {
+                base_version,
+                line_ending,
+                edits: Vec::new(),
+            });
+        }
+
+        let old_text = self.as_rope().clone();
+        cx.background_spawn(async move {
+            let mut offset = len;
+            for chunk in old_text.reversed_chunks_in_range(0..len) {
+                let non_whitespace_len = chunk
+                    .trim_end_matches(|c: char| c.is_ascii_whitespace())
+                    .len();
+                offset -= chunk.len();
+                offset += non_whitespace_len;
+                if non_whitespace_len != 0 {
+                    if offset == len - 1 && chunk.get(non_whitespace_len..) == Some("\n") {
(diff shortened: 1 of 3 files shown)


crates/project/src/lsp_store.rs
```diff
@@ -2265,15 +2304,20 @@ impl LocalLspStore {
...
         let capabilities = &language_server.capabilities();
         let range_formatting_provider = capabilities.document_range_formatting_provider.as_ref();
-        if range_formatting_provider == Some(&OneOf::Left(false)) {
-            anyhow::bail!(
-                "{} language server does not support range formatting",
+        if !matches!(
+            range_formatting_provider,
+            Some(OneOf::Left(true) | OneOf::Right(_))
+        ) {
+            zlog::debug!(
```
SomeoneToIgnore: Could that be just `log::debug!` instead so we do not have to bend the code and add another parameter just for the debug logging?

Also, presumably, this is not a frequent thing so maybe `info` level is ok to let people know why things are silently not working?

crates/language/src/buffer.rs
```diff
@@ -2280,6 +2280,88 @@ impl Buffer {
...
+        let base_version = self.version();
+        let modified_rows = modified_rows.to_vec();
+        cx.background_spawn(async move {
+            let all_ranges = trailing_whitespace_ranges(&old_text);
+            let empty = Arc::<str>::from("");
+            let edits = all_ranges
+                .into_iter()
+                .filter(|range| {
+                    let row = old_text.offset_to_point(range.start).row;
+                    modified_rows.iter().any(|modified| modified.contains(&row))
```
SomeoneToIgnore: I'm concerned by somewhat quadratic nature of this `filter` + the `trailing_whitespace_ranges` going over the entire text already.

What if we somehow amend the `trailing_whitespace_ranges` or make another function nearby that does exactly one pass over this all?

crates/language/src/buffer.rs
```diff
@@ -2280,6 +2280,88 @@ impl Buffer {
...
+                        return Diff {
+                            base_version,
+                            line_ending,
+                            edits: Vec::new(),
+                        };
+                    }
+                    break;
+                }
+            }
+            let newline: Arc<str> = Arc::from("\n");
```
SomeoneToIgnore: This is wrong for the CRLF case we support, is not it?
Would be cool to have tests on this case too.

SomeoneToIgnore (commented):
Sorry for the delay, I'm finally dealing with my review debts.

This looks nice overall, two main concerns I have:

* the perf of `remove_trailing_whitespace_in_ranges` in a large file with lots of selections
* the CRLF case untested

Ready to merge after we improve them.

crates/language/src/buffer.rs
G36maid: Thanks for flagging the CRLF case. I've added two tests that cover it:
- test_trailing_whitespace_in_ranges_crlf
- test_final_newline_in_range_crlf

The CI needs to rerun, we can confirm if the Windows test matrix passes as well.

crates/project/src/lsp_store.rs
G36maid: Thanks!
Removed the logger parameter and switched to log::info! .

crates/language/src/buffer.rs
G36maid: Thanks for catching that. 
Replaced the two-pass approach with trailing_whitespace_ranges_in_rows that does a single pass over the rope, checking row membership inline.

G36maid:
All three test platforms pass including Windows — the CRLF tests are green. 
The two failed checks (tests_pass, check_workspace_binaries) appear to be infra flakes unrelated to this PR.

G36maid:
Hi @SomeoneToIgnore , just checking in on the status of this(and another) PR.

SomeoneToIgnore:
Sorry, I've totally dropped the ball here: somehow, these PRs are hard for me to review, and I am leaving the company at the end of the month.

I expect another teammate to eventually come and review these.

SomeoneToIgnore (approved):
Thank you a lot for the wait.
I will merge this PR after this Wednesday's release.

I've finally sat down and tidied it all up, as finally realized that my issues with reviewing it were caused by severe duplication of the old code.

During that, turned out that there were bugs (fixed now):
* `anchor_ranges_to_row_ranges` over-included a row (selection ending at column 0)
* `ensure_final_newline_in_range` deleted unselected content at EOF

Instead of [irritating and pinging people](https://github.com/zed-industries/zed/issues/16509#issuecomment-4826749688), please strongly consider improving your work during the wait, as a way more productive thing.
Building a binary and testing it seems like a good idea too, as it should have uncovered those latent bugs in no time...

PR #53943, opened 2026-04-15
vim: Restore normal search jumps after dismissing cmd-f

### LINKED ISSUE #53896: vim mode: search by pressing ‎`n` selects everything between matches instead of jumping to next match
### Reproduction steps

1. Open a file that contains several occurrences of the same word.
2. Enable Vim mode in Zed.
3. Place the cursor on one of those words.
4. Press ‎`*` or ‎`#` to search for the current word.
5. Press ‎`n` or ‎`N` to jump to the next/previous match.

When pressing ‎`n` or ‎`N`, Zed duplicates the selection: it selects both the current match highlight and the previous one, including all text between them, instead of just moving the cursor to the next/previous match.

If, instead of ‎`n` / ‎`N`, you repeatedly press ‎`*` or ‎`#`, they behave correctly and only move to the next/previous match without selecting the whole range.

In an older Zed version, ‎`n` / ‎`N` also behaved correctly, so this looks like a regression.

This might be related to the recent change in search behavior with Vim mode enabled (#51073).

https://github.com/user-attachments/assets/ceb1522a-6b93-4268-8565-a949dcd0d6ae

## Expected Behavior

Pressing ‎`n` or ‎`N` in Vim mode should behave like in Vim and like repeated ‎`*` / ‎`#` in Zed:
just move the cursor to the next/previous match highlight, without selecting the entire region between the previous and current match.

OS: macOS 26.3.1
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

hamza-paracha:
This fixes #53896.

The editor was holding on to `cmd_f_search` state after the search UI was dismissed, so the next Vim `n` or `N` jump could behave like a Visual-mode selection instead of a normal search move.

This clears that state before native Vim search navigation, clears it again on submit and dismissal, and adds regression coverage for the reported `cmd-f` -> dismiss -> `n` flow.

## Validation
- `cargo fmt --all`
- manually verified the reported `cmd-f` dismissal + `n` flow works as expected
- added a regression test for the reported `cmd-f` dismissal + `n` flow

Release Notes:

- Fixed Vim search jumps after dismissing `cmd-f` search so `n`/`N` stay in Normal mode instead of selecting into Visual mode.

### DIFF
--- a/crates/vim/src/normal/search.rs
+++ b/crates/vim/src/normal/search.rs
@@ -132,6 +132,13 @@ pub(crate) fn register(editor: &mut Editor, cx: &mut Context<Vim>) {
 }
 
 impl Vim {
+    fn reset_cmd_f_search(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        if self.search.cmd_f_search {
+            self.search.cmd_f_search = false;
+            self.sync_vim_settings(window, cx);
+        }
+    }
+
     fn search_under_cursor(
         &mut self,
         action: &SearchUnderCursor,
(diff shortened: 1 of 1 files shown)


DevDonzo:
@cla-bot check

osiewicz:
Hey, thank you for submitting a PR! We've looked at your PR with @osyvokon and realized that we've received another submission in the meantime for that issue, and thus I'll leave it up to my colleagues (@dinocosta and @cameron1024) to decide whether #54027 or #53943 is a better approach to fix it, as I'm not that well-versed with Vim mode myself. That said, I appreciate your contribution nonetheless!

One nit/note for Cameron and Dino: I think we may want to avoid adding multiple fields just for `cmd-f` handling? That seems a bit brittle to me.

DevDonzo:
Pushed a compile fix for the async window context issue. The task now uses the window context provided by spawn_in via update_in instead of capturing the outer mutable Window reference across the async boundary.

dinocosta (commented):
Hey @DevDonzo 👋

Sorry for the late reply here, I was out of office for the past month. 

Unfortunately it seems that #54027 has already been merged and it fixed the linked issue. Sorry about that, we probably need to do a better job of surfacing existing pull requests for issues to avoid multiple contributors opening PRs for the same thing 😔

Seeing as those changes have already been merged, I’m going to go ahead and close this pull request, but do let me know if there’s any scenario your changes take into consideration that the merged changes do not.

Thank you for trying to tackle this, and I hope you manage to find another issue to contribute to! 🙂⁠￼

PR #53920, opened 2026-04-14
Fix autosave on focus change error when switching git diff via command palette

OmChillure:
Summary of whats fixed:

- Fixed the autosave regression where focus change to the command palette could trigger autosave at the wrong time in Vim/Helix workflows.
- The behavior is now: defer autosave while command palette is active, then decide on dismiss:
- Skip autosave if focus returns to the same item.
- Run autosave if focus moved elsewhere. 

Files changed:

- Added deferred autosave queueing when focus leaves an item to command palette in crates/workspace/src/item.rs.
- Added a modal dismissed event so workspace can react after modal close in crates/workspace/src/modal_layer.rs.
- Wired workspace to process deferred autosaves on modal dismiss and added handling logic in crates/workspace/src/workspace.rs.

Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #53863

Video 

[Screencast from 2026-04-14 22-57-12.webm](https://github.com/user-attachments/assets/8403403a-3e2e-4e46-a46f-c30e827c1c63)

Release Notes:

- Fixed autosave-on-focus-change behavior around command palette dismissal in Vim/Helix mode by covering deferred autosave semantics (skip when focus returns to same item, save when focus moves elsewhere).

PR #53948, opened 2026-04-15
workspace: Autosave when command palette moves focus away

### LINKED ISSUE #53863: Autosave on focus change doesn't trigger when switching to git diff via command palette
### Reproduction steps

1. enable vim mode
2. change a file in zed
3. open git diff via command pallette (ctrl +p then search for git:diff then open the diff)

Image

as you can see git diff panel is open but the app.vue file has "not saved" indicator.

the problem is only present for me when vim mode is enabled

Zed 0.231.2 
Os. Arch + hyprland(wayland)

### Attach Zed log file

```log
nothing relevant in the log
```

</details>

### Relevant Zed settings

```json

{
  "diff_view_style": "unified",
  "session": {
    "trust_all_worktrees": true,
  },
  "autosave": "on_focus_change",
  "vim_mode": true,
  "git": {
    "inline_blame": {
      "enabled": false,
    },
  }
}


```

</details>

### Relevant Keymap

```json
[
  {
    "bindings": {
      "ctrl-s": "workspace::Save",
    },
  },
  {
    "context": "Editor && vim_mode == normal && !VimWaiting && !menu",
    "bindings": {
      "ctrl-b": "workspace::ToggleLeftDock",
    },
  },
  {
    "context": "Workspace && vim_mode == normal",
    "bindings": {
      "ctrl-b": "workspace::ToggleLeftDock",
    },
  },
  {
    "context": "!Terminal",
    "bindings": {
      "ctrl-w": "pane::CloseActiveItem",
    },
  },
]

```

</details>

_No response_

None

DevDonzo:
This fixes #53863.

Autosave was being skipped when Vim or Helix mode routed focus through the command palette on the way to a different item, like opening Git Diff. The existing guard was correct for commands that return to the same editor, but too broad for commands that actually move focus somewhere else.

This defers that check by one turn so we can tell the difference between those two cases, and adds regression coverage for both paths.

Release Notes:

- Fixed on-focus-change autosave when command palette actions open another item in Vim and Helix modes

### DIFF
--- a/crates/workspace/src/item.rs
+++ b/crates/workspace/src/item.rs
@@ -965,6 +965,25 @@ impl<T: Item> ItemHandle for Entity<T> {
                             // palette (such as by opening up the tab switcher from it and then switching tabs that
                             // way).
                             if workspace.is_active_modal_command_palette(cx) {
+                                let item = item.clone();
+                                let project = workspace.project.clone();
+                                cx.defer_in(window, move |workspace, window, cx| {
+                                    if item.workspace_settings(cx).autosave
+                                        != AutosaveSetting::OnFocusChange
+                                    {
+                                        return;
+                                    }
+
+                                    let focus_handle = item.item_focus_handle(cx);
+                                    if focus_handle.contains_focused(window, cx)
+                                        || workspace.is_active_modal_command_palette(cx)
+                                    {
+                                        return;
+                                    }
+
+                                    Pane::autosave_item(&item, project.clone(), window, cx)
+                                        .detach_and_log_err(cx);
+                                });
                                 return;
                             }
                         }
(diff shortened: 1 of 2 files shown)


dinocosta (commented):
Hey @DevDonzo 👋 

Thank you for trying to solve this issue. It looks like @OmChillure had already opened a Pull Request for the same issue ~14 hours before – https://github.com/zed-industries/zed/pull/53920 . 

In order to avoid having maintainers review different fixes for the same issue I'm going to go ahead and close this one. However, feel free to take a look at the other solution and check if it's missing some use-case you might be contemplating here and engaging with the author in order to improve the solution. Thanks! 🙂

PR #53951, opened 2026-04-15
remote: Set current_dir for cargo zigbuild cross-compilation

### LINKED ISSUE #53950: Dev container cross-compile build fails: cargo zigbuild missing current_dir
## Reproduction steps

1. Build Zed from source on macOS (aarch64)
2. Launch Zed from a directory that is not the Zed source tree (e.g. `cd /tmp && /path/to/zed some-project --dev-container`)
3. Open a project with a `.devcontainer` targeting linux (aarch64-unknown-linux-musl)
4. Dev container creation triggers cross-compilation of `remote_server` via `cargo zigbuild`

## Current vs. Expected behavior

Current: Build fails with `error: cannot specify features for packages outside of workspace` (or `could not find Cargo.toml`) because `cargo zigbuild` runs without a working directory set, inheriting whatever the process cwd happens to be.

```
Failed to run command: Command { program: "cargo", args: ["zigbuild", "--package", "remote_server", "--features", "debug-embed", "--target-dir", "target/remote_server", "--target", "aarch64-unknown-linux-musl"], envs: {"RUSTFLAGS": Some(" -C target-feature=+crt-static")}, env_clear: false, current_dir: None, ...}: output: error: cannot specify features for packages outside of workspace
```

Expected: The cross-compile path at `crates/remote/src/transport.rs:317` should set `.current_dir(concat!(env!("CARGO_MANIFEST_DIR"), "/../.."))` like the native build path already does at line 271.

This works by accident when Zed's cwd happens to be the Zed source tree, but fails in normal use.

## Zed version and system specs

Built from source, branch `main` at commit ff2aca4597. macOS 15.5, arm64.

antont:
## Summary

- The cross-compile path for building `remote_server` via `cargo zigbuild` was missing `.current_dir()`, causing it to inherit the process cwd
- This fails with "cannot specify features for packages outside of workspace" when Zed is launched from outside its source tree
- The native build path at the same location already sets `.current_dir()` correctly — this aligns the cross-compile path

Closes #53950

## Test plan

- [x] Build Zed from source
- [x] Launch from `/tmp`: `cd /tmp && /path/to/zed some-project --dev-container`
- [x] Verify dev container cross-compilation succeeds

Release Notes:

- Fixed dev container cross-compilation failing when Zed is launched from outside its source directory

### DIFF
--- a/crates/remote/src/transport.rs
+++ b/crates/remote/src/transport.rs
@@ -316,6 +316,7 @@ async fn build_remote_server_from_source(
         log::info!("building remote binary from source for {triple} with Zig");
         run_cmd(
             new_command("cargo")
+                .current_dir(concat!(env!("CARGO_MANIFEST_DIR"), "/../.."))
                 .args([
                     "zigbuild",
                     "--package",
--- a/crates/remote/src/transport.rs
+++ b/crates/remote/src/transport.rs
@@ -332,7 +332,8 @@ async fn build_remote_server_from_source(
         )
         .await?;
     };
-    let bin_path = Path::new("target")
+    let bin_path = Path::new(concat!(env!("CARGO_MANIFEST_DIR"), "/../.."))
+        .join("target")
         .join("remote_server")
         .join(&triple)
         .join("debug")


SomeoneToIgnore (approved):
Thank you.

PR #53952, opened 2026-04-15
lsp: Enforce stricter CSS selector highlights

JacobNWolf:
**Self-Review Checklist:**

**Closes #51483.**

Proof: 

https://github.com/user-attachments/assets/afd5d656-b93d-4144-99a9-874b38baa19e

Release Notes:

- Fixed document highlights in CSS (and SCSS/Less) so a nested tag in a compound selector (for example `div p` inside an outer `div` rule) is not grouped with the parent rule’s tag name.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5414,6 +5414,7 @@ dependencies = [
  "tracing",
  "tree-sitter-bash",
  "tree-sitter-c",
+ "tree-sitter-css",
  "tree-sitter-html",
  "tree-sitter-md",
  "tree-sitter-python",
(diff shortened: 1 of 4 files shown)


osiewicz:
Can you repro this issue within VSC as well? I'm skeptical of slapping a fix on top of Zed for what seems to be a language server issue.

ChristopherBiscardi:
This is also the way VSCode handles this, and this PR in particular modifies *all* lsp responses which wouldn't be a mergable solution. (more context left on the original issue as well)

PR #53953, opened 2026-04-15
Show both local and remote checkouts in Recent Projects

### LINKED ISSUE #53917: Remote and local projects with same name shadow each other in Recent Projects
### Reproduction steps

1. Clone a git repo locally and on some remote machine
2. Open both the remote and local clone as independent workspaces in Zed once
3. Switch Zed to an unrelated workspace
4. Restart Zed
5. Activate "projects: open recent" via the title bar button, command palette, File menu or hotkey

Current: The popup only shows one of the workspaces from (1.) 
(if one of the two workspaces is already open, neither one is shown in the popup)

Expected: the Recent Projects popups show both the remote and the local workspaces

Zed: v0.230.1+stable (Zed) 
OS: Linux Wayland ubuntu 26.04
Memory: 31.3 GiB
Architecture: x86_64
GPU: AMD Radeon Vega Frontier Edition (RADV VEGA10) || radv || Mesa 26.0.3

hamza-paracha:
This fixes #53917.

Recent Projects was treating local and remote workspaces as the same entry whenever they resolved to the same checkout path. That meant one could hide the other, and if one of them was already open, the picker could end up hiding both.

This change keeps the deduplication and picker filtering keyed on both the workspace location and the path list, so local and remote copies of the same repo can both appear when they should.

## Test plan
- `rustfmt --edition 2024 crates/recent_projects/src/recent_projects.rs crates/workspace/src/persistence.rs`
- `git diff --check -- crates/recent_projects/src/recent_projects.rs crates/workspace/src/persistence.rs`

Release Notes:

- Fixed Recent Projects hiding either the local or remote workspace when both used the same checkout path.

### DIFF
--- a/crates/recent_projects/src/recent_projects.rs
+++ b/crates/recent_projects/src/recent_projects.rs
@@ -17,7 +17,7 @@ use fs::Fs;
 #[cfg(target_os = "windows")]
 mod wsl_picker;
 
-use remote::RemoteConnectionOptions;
+use remote::{RemoteConnectionOptions, same_remote_connection_identity};
 pub use remote_connection::{RemoteConnectionModal, connect, connect_with_modal};
 pub use remote_connections::{navigate_to_positions, open_remote_project};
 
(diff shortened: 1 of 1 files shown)


hamza-paracha:
Pushed a follow-up for the review comments:

- Replaced the O(n²) recent-workspace dedupe pass with a hash map keyed by stable remote identity plus paths.
- Switched open remote matching to `same_remote_connection_identity` so runtime-only SSH fields like nickname/password/args do not affect matching.
- Removed the unused `VisualTestContext` import and added coverage for matching persisted SSH options against live options with runtime-only fields.

Verified locally:
- cargo fmt --check
- cargo test -p workspace workspace_location
- ./script/clippy -p workspace

I also attempted `cargo test -p recent_projects --lib test_location_matches_remote_host`, but this machine is missing the Xcode metal compiler and fails before tests with `xcrun: error: unable to find utility "metal"` while building `gpui_macos`.

mikayla-maki (approved):
Thanks for the contribution!

We totally rewrote the code this was based on, so I took the liberty of recreating the intent on our current codebase :)

PR #53954, opened 2026-04-15
Respect proxy settings when fetching JSON schemas

### LINKED ISSUE #53819: Fetching of JSON schema fails regularly (not respecting proxy setting) when behind corporate network
### Reproduction steps

1. Start Zed
2. open a package.json file from a JavaScript project

`Unable to load schema from 'https://www.schemastore.org/semantic-release.json':  . (json 65536)`
`Unable to load schema from 'https://www.schemastore.org/jscpd.json': . (json 65536)`

If i go to those end points in a browser directly they work, so its not like they're down or 500ing.

I am on a corporate network and I am assuming this is failing the firewall. I do have proxy information set via the `proxy` setting, which other parts of the editor work for but not this. My assumption is whatever handles fetching JSON schemas isn't respecting the proxy settings.

If i set the `http_proxy=` env var before starting Zed it works fine.

Also semi-related:
https://github.com/zed-industries/zed/discussions/30516

I see this warning message at the top of ever JSON file

I expect not to see this message, and requests to schema work (or i can turn it off so it doesn't attempt)

Zed 0.231.2 
cc335b70f85a17974a4c61f852dbebff8c4b1db7

0.231.2+stable.221.cc335b70f85a17974a4c61f852dbebff8c4b1db7

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

DevDonzo:
This fixes #53819.

JSON schema downloads were still going through the JSON language server without picking up Zed's configured proxy. On networks that require the proxy, that made schema fetches fail even though the rest of the app could connect normally.

This passes the proxy setting through to the JSON language server configuration so schema requests follow the same network path as the rest of Zed.

## Test plan
- `rustfmt --edition 2024 crates/languages/src/json.rs`
- `git diff --check -- crates/languages/src/json.rs`

Release Notes:

- Fixed JSON schema downloads ignoring Zed's proxy setting.

### DIFF
--- a/crates/languages/src/json.rs
+++ b/crates/languages/src/json.rs
@@ -5,7 +5,10 @@ use async_trait::async_trait;
 use collections::HashMap;
 use futures::StreamExt;
 use gpui::{App, AsyncApp, Entity, Task};
-use http_client::github::{GitHubLspBinaryVersion, latest_github_release};
+use http_client::{
+    HttpClient as _,
+    github::{GitHubLspBinaryVersion, latest_github_release},
+};
 use language::{
     Buffer, ContextProvider, LanguageName, LanguageRegistry, LocalFile as _, LspAdapter,
     LspAdapterDelegate, LspInstaller, Toolchain,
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi (approved):
I've confirmed that this does result in redirecting the relevant schema requests; which in the case of semantic-release, etc are caused by a non-standard vscode json lsp action since while package.json's schema is local to zed, the semantic-release url has to then be requested.

Image

PR #53958, opened 2026-04-15
file_finder: Support line range queries with path:start-end syntax

cppcoffee:
Release Notes:

- Added support for opening a file with a line range selected in the file finder using the `path:start-end` syntax (e.g. `file.rs:10-20`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6184,6 +6184,7 @@ dependencies = [
  "futures 0.3.32",
  "fuzzy",
  "gpui",
+ "language",
  "menu",
  "open_path_prompt",
  "picker",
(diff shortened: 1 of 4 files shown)


smitbarmase (approved):
Thanks!

1. I noticed one case where the query `foo.rs:1-` would drop the result for `foo.rs`. Fixed that in https://github.com/zed-industries/zed/pull/53958/commits/9f856aa63b90342c5b66fa0d8f5e9fcc6060c373.

2. After that, I also saw that `foo.rs:1-2` would show only foo.rs in the create path. I think we should follow what happens for `foo.rs:2:3`, i.e. let the user create a file with that exact name. If they don't want it, they can clear the trailing characters. At least this way we stay consistent and predictable about creating a file from the typed string. Updated it in https://github.com/zed-industries/zed/pull/53958/commits/96336910f1e5c7d01f6a363b9b1c0a5545eefe1b.

One thing I noticed but didn't fix: for a query like `foo.rs:2:`, the create path shows only `foo.rs:2`. That isn't something introduced by this PR though, so I think it can be addressed separately.

Thanks again.

PR #53959, opened 2026-04-15
Delete worktree directory before git metadata cleanup during archive

rtfeldman:
When archiving a thread's linked worktree, `git worktree remove` can fail to delete the working directory while still removing the admin metadata in `.git/worktrees/<name>`. This leaves an orphaned folder on disk that no longer appears in `git worktree list`.

## Fix

Delete the directory ourselves first with `std::fs::remove_dir_all`, then call `git worktree remove --force` to clean up only the admin entry. Git already handles the case where the working tree is already gone (since [git/git@ee6763a](https://github.com/git/git/commit/ee6763af0a3b97225803c6c908a29de40336cf38)).

Also fixes `FakeGitRepository::remove_worktree` to handle a missing working tree directory (matching real git behavior) by scanning `.git/worktrees/` entries when the `.git` pointer file can't be read.

Release Notes:

- Fixed archiving an agent thread sometimes leaving the worktree folder on disk even though the git worktree was removed.

### DIFF
--- a/crates/agent_ui/src/thread_worktree_archive.rs
+++ b/crates/agent_ui/src/thread_worktree_archive.rs
@@ -189,20 +189,40 @@ async fn remove_root_after_worktree_removal(
         }
     }
 
+    // Delete the directory ourselves first, then tell git to clean up the
+    // metadata. This avoids a problem where `git worktree remove` can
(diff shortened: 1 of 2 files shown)


bennetbo:
@zed-industries/approved

PR #53961, opened 2026-04-15
cli: Fix -n behavior and refactor open options (v0.232.x)

eholk:
Cherry-pick of #53939 to v0.232.x.

Fixes a regression where `zed -n .` in a subdirectory of an already-open
project would redirect to the parent window instead of creating a new one.

See the original PR for full details. This is a manual cherry-pick because
the patch didn't apply cleanly due to divergence on the v0.232.x branch.

Release Notes:

- N/A


benbrandt:
@zed-zippy approved

PR #53964, opened 2026-04-15
settings_ui: Gate CLI open behavior setting on feature flag

eholk:
The `cli_default_open_behavior` setting controls how `zed <path>` opens directories, but the underlying behavior is gated behind the `AgentV2FeatureFlag` in `open_listener.rs`. Without the flag enabled, the setting has no effect — Zed always falls back to `NewWindow` behavior.

This change gates the settings UI entry behind the same feature flag so it only appears when the feature is actually active.

**Changes:**
- Converted `general_settings_section()` from a fixed-size array to a `Vec` so the CLI setting can be conditionally appended
- Added `feature_flags` as a dependency to `settings_ui`

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15896,6 +15896,7 @@ dependencies = [
  "edit_prediction",
  "edit_prediction_ui",
  "editor",
+ "feature_flags",
  "fs",
  "futures 0.3.32",
  "fuzzy",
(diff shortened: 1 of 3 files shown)


benbrandt:
@zed-zippy approved

PR #53970, opened 2026-04-15
ci: Add workflow for bumping Zed versions

MrSubidubi:
This PR refactors the way we do Zed version bumps to instead have these done by the Zed Zippy bot. 

Since we before were doing all on a singular push, I had to split it up into two phases: Creating the actual changes to the files and then tagging the releases appropriately after. I consider especially the workflow for the latter a _very_ temporary solution and think we are better off with moving this into the cloud worker for the Zed Zippy bot.

For bumping the version, the process is now as follows:
1. An authorized actor runs `script/trigger-zed-version-bump` with the appropriate parameters (`major`, `minor` or `patch`, where the latter also takes `stable`, `preview` or `both` for the patch bump. The former always bumps all versions across channels).
2. This triggers a workflow that first determines a matrix for the changes needed to make across the three channels (bumping main, creating preview and moving stable to preview).
3. The changes are then performed in a Matrix workflow for all three channels and PRs created.

After these are then sucessfully merged, we trigger the actual releases with the tagging workflow. 

I think this is the easiest way going forward to stay compliant, as this way, we can ensure that the responsible actor has actually only bumped the version and nothing else. Open for feedback nonetheless

Release Notes:

- N/A


MrSubidubi:
Closing since we will revisit a more contained approach for this

PR #53971, opened 2026-04-15
language_models: Fix misleading copy when hosted models are disabled

tomhoule:
The agent settings UI unconditionally showed "You have access to Zed's hosted models through your Organization" for all Business plan users, even when the org admin had turned off the Zed AI model provider. Now the copy reads "Zed's hosted models are disabled by your organization's configuration" when is_zed_model_provider_enabled is false.

Also added component preview entries for both Business plan states.

Closes CLO-667

Release Notes:

- N/A

### DIFF
--- a/crates/language_models/src/provider/cloud.rs
+++ b/crates/language_models/src/provider/cloud.rs
@@ -276,6 +276,7 @@ impl LanguageModelProvider for CloudLanguageModelProvider {
 struct ZedAiConfiguration {
     is_connected: bool,
     plan: Option<Plan>,
+    is_zed_model_provider_enabled: bool,
     eligible_for_trial: bool,
     account_too_young: bool,
     sign_in_callback: Arc<dyn Fn(&mut Window, &mut App) + Send + Sync>,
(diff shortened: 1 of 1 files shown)


crates/language_models/src/provider/cloud.rs
```diff
@@ -297,7 +298,11 @@ impl RenderOnce for ZedAiConfiguration {
                 true,
             ),
             Some(Plan::ZedBusiness) => (
-                "You have access to Zed's hosted models through your Organization.",
+                if self.is_zed_model_provider_enabled {
+                    "You have access to Zed's hosted models through your Organization."
```
NeelChotai:
```suggestion
                    "You have access to Zed's hosted models through your organization."
```

NeelChotai (approved):

crates/language_models/src/provider/cloud.rs
NeelChotai: think I added this, committed the suggestion to be consistent

PR #53974, opened 2026-04-15
sidebar: Fix removed projects not immediately disappearing from the UI

danilo-leal:
Release Notes:

- N/A


bennetbo:
@zed-industries/approved

PR #52699, opened 2026-03-30
editor: Merge additional completion edits into primary undo transaction

HuaGu-Dragon:
Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #45986

Release Notes:

- Fixed postfix snippets now creating 1 `undo` actions instead of 2.

PR #53980, opened 2026-04-15
Vim/Helix: Prevent undo grouping when any LSP completion

HuaGu-Dragon:
Self-Review Checklist:

## Summary  
Prevent undo grouping when an LSP completion includes extra edits so that the completion and its extra edits are applied and reverted atomically.

## Problem
When applying an LSP completion that also applies extra edits, the editor may merge that completion into the surrounding undo group or split the transaction while waiting for edits, causing undo to leave the buffer in an inconsistent state.

## Solution
* Always block undo merging for completions

Closes #ISSUE

Release Notes:

- Prevent undo grouping when any LSP completion

### DIFF
--- a/crates/editor/src/completions.rs
+++ b/crates/editor/src/completions.rs
@@ -975,6 +975,11 @@ impl Editor {
             self.show_signature_help(&ShowSignatureHelp, window, cx);
         }
 
+        // After the code completion is finished, we should finalize the last transaction.
+        // This ensure vim/helix not group the edits together.
+        self.buffer
+            .update(cx, |buffer, cx| buffer.finalize_last_transaction(cx));
+
         Some(cx.spawn_in(window, async move |editor, cx| {
             let additional_edits_tx = apply_edits.await?;
 
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1068,6 +1068,7 @@ pub struct Editor {
     context_menu_options: Option<ContextMenuOptions>,
     mouse_context_menu: Option<MouseContextMenu>,
     completion_tasks: Vec<(CompletionId, Task<()>)>,
+    completion_additional_edits_pending: bool,
     inline_blame_popover: Option<InlineBlamePopover>,
     inline_blame_popover_show_task: Option<Task<()>>,
     signature_help_state: SignatureHelpState,
@@ -2321,6 +2322,7 @@ impl Editor {
             context_menu_options: None,
             mouse_context_menu: None,
             completion_tasks: Vec::new(),
+            completion_additional_edits_pending: false,
             inline_blame_popover: None,
             inline_blame_popover_show_task: None,
             signature_help_state: SignatureHelpState::default(),
@@ -5696,6 +5698,10 @@ impl Editor {
                 .borrow()
                 .as_ref()
                 .is_some_and(|menu| menu.visible())
+    }
+
+    pub fn completion_additional_edits_pending(&self) -> bool {
+        self.completion_additional_edits_pending
     }
 
     pub fn context_menu_origin(&self) -> Option<ContextMenuOrigin> {
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -1068,7 +1068,6 @@ pub struct Editor {
     context_menu_options: Option<ContextMenuOptions>,
     mouse_context_menu: Option<MouseContextMenu>,
     completion_tasks: Vec<(CompletionId, Task<()>)>,
-    completion_additional_edits_pending: bool,
     inline_blame_popover: Option<InlineBlamePopover>,
     inline_blame_popover_show_task: Option<Task<()>>,
     signature_help_state: SignatureHelpState,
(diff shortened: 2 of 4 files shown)


SomeoneToIgnore:
Unassigned myself as apparently do not understand Vim specific related to all this.

I have assumed we need a generic fix like https://github.com/zed-industries/zed/pull/52699 or similar and got quite confused on the current impl.

HuaGu-Dragon:
> Unassigned myself as apparently do not understand Vim specific related to all this.
> 
> I have assumed we need a generic fix like #52699 or similar and got quite confused on the current impl.

Yeah....
Perhaps I should change the logic for `group transactions` rather than using a hack here.

HuaGu-Dragon:
Maybe I need some help here. After digging deeper, the core issue is very specific to Vim/Helix mode:

Vim tracks an entire **Insert mode session** as a single undo group via `self.current_tx`. When confirming an LSP completion, the editor inserts the main text (moving the cursor) and then `await` additional text edits. During this `await`, `vim.rs` sees the cursor move (`current_anchor != newest`), assumes the user manually moved it (e.g., via arrow keys), and prematurely drops `current_tx`, breaking the undo group.

While it might be tempting to add metadata to the generic `Transaction` struct to tell Vim "this is a completion, don't break the group", doing so for a Vim-specific behavior would require a huge refactor and break the crate layering?

HuaGu-Dragon:
Maybe explicitly calls `buffer.finalize_last_transaction()` is better, since `project/src/lsp_store.rs` also do that before applying the additional text.
https://github.com/zed-industries/zed/blob/ff8fa053ff0d0c02aec95def5b796dbe5cd541e1/crates/project/src/lsp_store.rs#L7014-L7015

we can actually solve this elegantly without adding any Vim-specific boolean flags (`completion_additional_edits_pending`) to the `Editor` struct.

SomeoneToIgnore (commented):
This definitely looks like a proper direction!

2 small questions I have left:

* pending LSP completion edits are the additional text edits, if I understand the issue right?

If so, then can we move the fix under `if let Some` checks in 

https://github.com/zed-industries/zed/blob/d4f06ee3740b0da1bd55476a5ff19d7c2ac1e57f/crates/editor/src/editor.rs#L6981-L6991

Or do you think this fix should apply to all, "synchronous" also, edits?

* The only fix-related comment I've found is

https://github.com/zed-industries/zed/blob/d4f06ee3740b0da1bd55476a5ff19d7c2ac1e57f/crates/language/src/buffer.rs#L2479-L2481

and that does not explain much what happens here.
Can we add a small comment on why this is needed and how the rest of the code interacts with it?

If it helps, a good example of such comment is a few lines above: https://github.com/zed-industries/zed/blob/d4f06ee3740b0da1bd55476a5ff19d7c2ac1e57f/crates/editor/src/editor.rs#L6953-L6954

HuaGu-Dragon:
> This definitely looks like a proper direction!
> 
> 2 small questions I have left:
> 
> * pending LSP completion edits are the additional text edits, if I understand the issue right?
> 
> If so, then can we move the fix under `if let Some` checks in
> 
> https://github.com/zed-industries/zed/blob/d4f06ee3740b0da1bd55476a5ff19d7c2ac1e57f/crates/editor/src/editor.rs#L6981-L6991
> 
> Or do you think this fix should apply to all, "synchronous" also, edits?
> 
> * The only fix-related comment I've found is
> 
> https://github.com/zed-industries/zed/blob/d4f06ee3740b0da1bd55476a5ff19d7c2ac1e57f/crates/language/src/buffer.rs#L2479-L2481
> 
> and that does not explain much what happens here. Can we add a small comment on why this is needed and how the rest of the code interacts with it?
> 
> If it helps, a good example of such comment is a few lines above: https://github.com/zed-industries/zed/blob/d4f06ee3740b0da1bd55476a5ff19d7c2ac1e57f/crates/editor/src/editor.rs#L6953-L6954

Yes — the pending LSP completion edits are the additional text edits.

Regarding your second question: I intentionally placed the fix (`finalize_last_transaction`) outside the `if let block` so that it applies to all completions, including synchronous ones. Here is my reasoning:

* After we enter completion, the cursor moves, then `vim.rs` detects this in `local_selections_changed` And Vim immediately issues a `group_until_transaction`. Since `apply_edits.await` is asynchronous, If we wait to finalize until inside the `if let Some` block, the `merge_transaction` will be incorrect.

https://github.com/zed-industries/zed/blob/8a8293b3bd658fbca3a5415d6663346d1a8078a4/crates/vim/src/vim.rs#L1941-L1949

* You raised a great point about whether this should apply to all synchronous edits as well. I tested the native behavior in Neovim and Helix, When you type `let a = ` and accept the completion `Some`, pressing `u` should only revert `Some`, not the `let a = ` typed beforehand. (This is also why the existing test case expects the state to revert to `2.` instead of `2`). I also add a test to cover this.

Here is the video about their behaviour. Sorry for that I do not have neovim locally, so I asked my friend to have a test in his machine, but the video have a bit blurry.

<details>
  <summary>Neovim</summary>
  

https://github.com/user-attachments/assets/8280d78c-ec77-4391-8b73-0a6fe4fb8791


</details>

<details>
  <summary>Helix</summary>
  

https://github.com/user-attachments/assets/ee3a126c-5e49-43c7-9e49-42fc31ee681f


</details>


Does this solve your concerns?

HuaGu-Dragon:
Sorry for the delay—I've been under academic pressure
I updated the PR title and description and rebased the branch to resolve conflicts.

SomeoneToIgnore (approved):
Thank you!

PR #53992, opened 2026-04-15
agent_ui: Fix session equality check for reloading retained threads

nathansobo:
## Summary

Adds three tests that exercise session disassociation ("Session not found" after reopening a thread). The third test **fails**, reproducing the bug.

### Root cause

A retained `ConversationView` with a `thread_error` receives `AgentServersUpdated`, which triggers `reset()` → `close_all_sessions` → state transitions to `Loading`. When the user then reopens the same session via `open_thread`, `load_agent_thread`'s `has_session` check returns `false` (because the retained view is in `Loading` state, not `Connected`), so a **second** `ConversationView` is created for the same session. Both views' async load tasks complete and both transition to `Connected` with the same session ID. When the retained view is later cleaned up, its `on_release` → `close_all_sessions` removes the session from the connection, leaving the active view stranded with a dead session.

### Tests

| Test | Status | What it verifies |
|------|--------|-----------------|
| `test_open_thread_on_visible_session_does_not_disassociate_it` | ✅ Pass | `load_agent_thread` short-circuits when session is already active |
| `test_retained_thread_cleanup_does_not_disassociate_reopened_session` | ✅ Pass | Eviction + reopen cycle works (close-then-reload) |
| `test_retained_thread_reset_race_disassociates_session` | ❌ **Fail** | Reproduces the race described above |

cc @bennetbo @ConradIrwin

Release Notes:

- N/A


crates/agent_ui/src/agent_panel.rs
```diff
@@ -2520,8 +2520,9 @@ impl AgentPanel {
 
         let has_session = |cv: &Entity<ConversationView>| -> bool {
             cv.read(cx)
-                .active_thread()
-                .is_some_and(|tv| tv.read(cx).thread.read(cx).session_id() == &session_id)
```
benbrandt, line 2524: The issue was Active thread session id could be a subagent or we could be in a loading/error state and the active thread isn't set yet. But we already know the session id, so we sholuld be using that instead

benbrandt (approved):

bennetbo (approved):

PR #53996, opened 2026-04-15
Persist agent follow toggle

nathansobo:
## Summary
- persist the agent follow toggle in settings as agent.auto_follow_agent
- re-engage follow on send and tool authorization when that setting is enabled
- keep the current implementation in draft because there are known issues to address before merge

## Known concerns
- The toggle currently depends on a live project to persist, so it may become a no-op for projectless threads.
- The toggle no longer updates ThreadView-local state, and ThreadView does not observe SettingsStore, so the button may not rerender immediately after clicking.

Release Notes:

- N/A


SomeoneToIgnore:
Autoclosing all draft PRs with no activity for 2 weeks or over, this still has a lot of CI failures: 

image

PR #54003, opened 2026-04-15
markdown_preview: Independent font family and theme

robinojw:
## Summary

Adds two user-facing settings that let the markdown preview render with a font family and theme independent from the editor. Both are optional and fall back to the editor defaults when unset.

```json
{
  "markdown_preview_font_family": "Georgia",
  "markdown_preview_theme": "Solarized Light"
}
```

## Details

- New `MarkdownFont::Preview` variant, used only by the preview surface.
- New `MarkdownStyle::themed_with_overrides()` that accepts explicit `ThemeColors` and `SyntaxTheme` so the preview can render with a theme other than the active editor theme. Existing `themed()` callers are unchanged.
- The preview pane background adopts the chosen theme's `editor_background`.
- Follows the pattern already used by `agent_ui_font_size` / `agent_buffer_font_size`.

## Scope

Sizing-related changes (`markdown_preview_font_size`, `markdown_preview_line_height`, typography tweaks, responsive max-width container) were in an earlier revision of this PR and have been removed per review feedback — they interact with the Cmd+/- zoom behavior in ways that still need design work, and will land in a follow-up.

## Files Changed

- `crates/settings_content/src/theme.rs` — schema
- `crates/theme_settings/src/settings.rs` — runtime + accessor
- `crates/markdown/src/markdown.rs` — `MarkdownFont::Preview` + `themed_with_overrides()`
- `crates/markdown_preview/src/markdown_preview_view.rs` — theme resolution, style selection, background
- `crates/markdown_preview/Cargo.toml` — `theme` dependency
- `crates/settings/src/vscode_import.rs` — new fields in struct literal

Release Notes:

- Added `markdown_preview_font_family` and `markdown_preview_theme` settings to customize the markdown preview independently from the editor.


robinojw:
> We require contributors to sign our Contributor License Agreement, and we don't have @robinojw on file. You can sign our CLA at https://zed.dev/cla. Once you've signed, post a comment here that says '@cla-bot check'.

Signed 🖊️

robinojw:
@cla-bot check

robinojw:
## Screenshots

**Before:** 
Screenshot 2026-04-16 at 3 54 32 PM


**After:**
Screenshot 2026-04-16 at 3 48 14 PM

Custom font etc:
Screenshot 2026-04-16 at 3 35 01 PM

dshap474:
Please push this through ASAP

robinojw:
@Anthony-Eid 🙏🏼

ChristopherBiscardi (changes requested):
This seems interesting as a feature, and is a decent set of starting configurations.

`markdown_preview_font_family` and `markdown_preview_theme` work as expected. In an ideal world we probably want the ability to set the font family differently for headings and paragraph text, but I don't think we support that *now*, so its fine that this PR doesn't address that.

However a couple options seem to apply to selective elements, or no elements, as seen in this screenshot:

Image

- `markdown_preview_font_size` seems to only apply to code blocks
- `markdown_preview_line_height` doesn't seem to apply when viewing preview

robinojw:
@ChristopherBiscardi thanks for the review:

- **`markdown_preview_font_size`**: preview heading sizes are now computed as multiples of the configured font size (h1 = 2.0×, h2 = 1.6×, h3 = 1.3×, h4 = 1.1×, h5 = 1.0×, h6 = 0.9×) instead of being pinned to GPUI's default rem base, so the setting scales the whole preview (headings + body + code) rather than only code blocks. Paragraph body text already picked up the setting via `base_text_style`, but the visual effect was dominated by the unchanged headings.
- **`markdown_preview_line_height`**: preview paragraphs now use the configured line-height instead of a fixed `1.6`. List items also pick it up (floored at `1.5` so tight rows don't clip).

ChristopherBiscardi (changes requested):
Currently:

- `markdown_preview_font_size` isn't applying to paragraph text, links, list elements, and such. Screenshot here shows a small font size set.
- similarly `markdown_preview_line_height` doesn't apply to code blocks.

Image

I also noticed the markdown preview is not responsive to Cmd+/- at all, which means you can't zoom in or out. That is notably different than how other text works (like regular .md files, etc). It turns out that on `main` Cmd+/- only affects codeblocks in the preview, so there seems to be some buggy interactions here, but this PR removes the ability that current exists to zoom those codeblocks. I feel fairly strongy that since this PR touches font sizing and makes it theme-able that zooming should work with that theming.

Also we should remove the "responsive container size" for the markdown document. The pane is already sizeable by users and the responsive container doesn't have a large impact on the display.

---

Since the sizing has gone through a couple of rounds now, I think we could move the size-related changes to a different PR and keep the font family/theme selection changes here. The theme/font family changes look good, but the sizing changes seem like they still need some work.

robinojw:
@ChristopherBiscardi agreed, dropped all the sizing/line-height/typography/responsive-container changes in 606cc04 and kept just font family + theme selection as you suggested.

Just wanted to say thank you for being so kind and bearing with me here!
I know how time consuming these tests/reviews can be. I've tested on my local and everything looks good.

The scope now is:

- `markdown_preview_font_family` (falls back to UI font)
- `markdown_preview_theme` (falls back to active editor theme)

Size-related changes will come back in a follow-up PR so the Cmd+/- zoom interaction can be thought through properly. PR title + body updated to reflect the reduced scope.

ChristopherBiscardi (approved):
Awesome, thanks for taking the feedback so well! This looks good to me now.

Image

diviatrix:
For those who will desperately try to remove these insane useless boders, here is solution. 

{
  "markdown_preview": {
    "limit_content_width": false
  }
}

image

PR #54004, opened 2026-04-15
gpui: Make property tests deterministic by default

Anthony-Eid:
### Summary

`#[gpui::property_test]` was non-deterministic because it always used a random seed with proptest. So it generated different test cases and scheduler seeds on every run, causing flaky failures. `#[gpui::test]` doesn't have this problem because it defaults to fixed seeds.

This changes `#[gpui::property_test]` to match: proptest's RNG seed now defaults to `0` instead of random, so the same cases run every time. The`$SEED` env var overrides both the scheduler seed and case generation seed.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

### DIFF
--- a/crates/gpui/src/test.rs
+++ b/crates/gpui/src/test.rs
@@ -44,6 +44,28 @@ pub fn seed_strategy() -> impl Strategy<Value = u64> {
     }
 }
 
+/// Returns a proptest config that is deterministic by default, matching
+/// `#[gpui::test]`'s behavior of using fixed seeds. When `$SEED` is set,
+/// uses that value; otherwise defaults to `0`.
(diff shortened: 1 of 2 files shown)


cameron1024:
Commentary on the commits I left:

`syn` itself provides facilities for parsing the `#[macro(foo = 123, bar = "hello")]` syntax that is common among attribute macros, using `Punctuated::<MetaNameValue, Token![,]>`. Doing this requires a `ParseStream`, which is not user-constructable, so we implement the `Parse` trait on the `Args` struct. Then, we can use `parse2` (or any of the other standard parse functions) to extract out the structured syntax we care about, and reconstruct the correct attribute macro in `fn render()`.

cameron1024 (approved):

PR #54011, opened 2026-04-15
project: Use runnable kind to disambiguate between cargo and shell commands

sunshowers:
Found this bug while investigating why configuring nextest based on the instructions at https://github.com/rust-lang/rust-analyzer/issues/21137#issuecomment-4254611341 wasn't working within Zed.

Previously, we'd use `serde(untagged)`, preferring cargo over shell commands. The problem is that every instance of a shell command is a valid instance of a cargo command. For example, the shell command:

```json
{
    "label": "test my_test",
    "kind": "shell",
    "args": {
        "environment": {"RUSTC_TOOLCHAIN": "/path/to/toolchain"},
        "cwd": "/project",
        "program": "cargo",
        "args": ["nextest", "run", "--package", "my-crate", "--lib", "--", "my_test", "--exact", "--include-ignored"]
    }
}
```

would end up getting deserialized as a Cargo command, silently dropping `program` and `args`.

With this fix, we now use the provided `kind` as a tag. We do have to introduce a `#[serde(flatten)]` unfortunately, which has a few side effects due to internal buffering, but `#[serde(untagged)]` also does internal buffering so this doesn't make things worse.

I've manually tested this by configuring:

```json
{
    "lsp": {
        "rust-analyzer": {
            "initialization_options": {
                "runnables": {
                    "test": {
                        "overrideCommand": [
                            "cargo",
                            "nextest",
                            "run",
                            "--package",
                            "${package}",
                            "${target_arg}",
                            "${target}",
                            "--",
                            "${test_name}",
                            "${exact}",
                            "${include_ignored}"
                        ]
                    }
                }
            }
        }
    }
}
```

and ensuring that nextest is correctly invoked.

Self-Review Checklist:

Release Notes:

- Fixed deserialization of rust-analyzer shell runnables.

### DIFF
--- a/crates/editor/src/runnables.rs
+++ b/crates/editor/src/runnables.rs
@@ -715,7 +715,9 @@ mod tests {
     use multi_buffer::{MultiBuffer, PathKey};
     use project::{
         FakeFs, Project, ProjectPath,
-        lsp_store::lsp_ext_command::{CargoRunnableArgs, Runnable, RunnableArgs, RunnableKind},
+        lsp_store::lsp_ext_command::{
+            CargoRunnableArgs, Runnable, RunnableArgs, ShellRunnableArgs,
+        },
     };
     use serde_json::json;
     use task::{TaskTemplate, TaskTemplates};
@@ -1028,7 +1030,6 @@ mod tests {
                             lsp::Position::new(3, 1),
                         ),
                     }),
-                    kind: RunnableKind::Cargo,
                     args: RunnableArgs::Cargo(CargoRunnableArgs {
                         environment: Default::default(),
                         cwd: path!("/project").into(),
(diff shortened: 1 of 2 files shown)


sunshowers:
@cla-bot check

SomeoneToIgnore (approved):
Thank you for digging into this.

I guess then https://github.com/zed-industries/zed/blob/5a589c4b1d3ebbd07ac69be43017c1fa149c6b4d/crates/project/src/lsp_store/lsp_ext_command.rs#L503-L504 deserves some update/rewrite, can you do that?

I think CI fails will go away with a rebase and are not related.

sunshowers:
Thanks @SomeoneToIgnore. I've rebased this on top of main and added a commit with an expanded comment on top. (Happy to squash this in if desired as well.)

To be honest I'd consider r-a to not be quite correct here, though I think it only does serializing so it doesn't get hit by this.

SomeoneToIgnore (approved):
Thank you!
No need to squash, we do that in the end anyway.

Fully agree with you that r-a has to get this fix ideally.

Veykril:
> To be honest I'd consider r-a to not be quite correct here, though I think it only does serializing so it doesn't get hit by this.

Curious to hear what the issue is on the r-a side (in your opinion). Note I haven't checked this PR yet to detail yet to catch up with this.

Veykril:
Ah, the fact that we have the enums defined as untagged despite clearly having a tag field I suppose?

sunshowers:
> Ah, the fact that we have the enums defined as untagged despite clearly having a tag field I suppose?

Yeah, exactly -- I think the functional change in this PR can be directly transferred over to r-a. Do you want to do that or should I?

Veykril:
If you could do that I'd really appreciate it, sounds like a reasonable change

sunshowers:
> If you could do that I'd really appreciate it, sounds like a reasonable change

Did this at https://github.com/rust-lang/rust-analyzer/pull/22295.

PR #54012, opened 2026-04-15
vim: Don't steal focus from non-pane panels for search commands

rtfeldman:
When the Agent panel (or any dock panel without its own pane) is focused and a file is open in the center editor, pressing `/` in vim mode would steal focus to the buffer's search bar instead of staying on the panel.

## Root cause

`Vim::pane()` calls `workspace.focused_pane()`, which falls back to the center pane when a dock panel without its own `pane()` method is focused. Vim search commands then open the `BufferSearchBar` on the center pane and focus it, stealing focus from the panel.

## Fix

Add a guard in `Vim::pane()` that returns `None` when the resolved pane doesn't actually contain focus. This prevents all vim search/match commands (`/`, `?`, `n`, `N`, `*`, `#`, etc.) from stealing focus from non-pane panels.

All 497 vim tests and 253 agent_ui tests pass.

Release Notes:

- Fixed vim search (`/`) stealing focus from the Agent panel when a file is open in the editor.


ConradIrwin (approved):
This seems reasonable. 

Not sure if we want to fix cmd-f etc at the same time? It might be better to add something to the key context when search is available and only register those commands then

PR #54014, opened 2026-04-15
Avoid showing draft threads in the sidebar or storing them in the database

maxbrunsfeld:
Now that we don't intend to show drafts in the sidebar, we just need this to protect against drafts that were stored by an older version of Zed.

Release Notes:

- N/A


mikayla-maki:
@zed-industries/approved

PR #54018, opened 2026-04-15
Add `git.discover_nested_repositories` setting

Mjoyufull:
## Summary
When a workspace folder contains many nested Git repositories (e.g. kept
for study/reference), Zed discovers all of them and creates active
repository state for each a worker threads, status scans, diff base
loads, branch queries. This causes significant CPU usage with no benefit
when the user doesn't actively develop in those repos.
This PR adds `git.discover_nested_repositories` (boolean, default `true`)
to control this behavior. When set to `false`, only the root repository
(or ancestor repository above the project root) is monitored. Nested
`.git` directories are skipped during scanning but remain visible in the
project panel.
The check is centralized in `BackgroundScannerState::insert_git_repository()`,
which all discovery paths flow through (initial scan, FS event entry
insertion, and FS event repository updates i think).
## Usage
```json
{
  "git": {
    "discover_nested_repositories": false
  }
}
```
Release Notes:
- Added a git.discover_nested_repositories setting to control whether
  nested Git repositories inside a workspace are actively monitored. Set
  to false to reduce CPU usage in workspaces containing many reference
  repos.


Mjoyufull:
@cla-bot check

eholk:
@Mjoyufull - Thanks for this PR. I'd like to understand a little more about your use case for this. For example, how many nested repos does your workspace folder have? How much CPU usage are you seeing? I'm wondering if we can find a way to handle that use case better rather than turning off nested git scanning.

Mjoyufull:
@eholk 
i think a .zedignore would help 100%
this is my dir
```sh
┌─[chris@Kyokai] - [~/projects/kyokai] - [12839]
└─[$] tree -L 1                                                             16:21:30
.
├── 0304-structured-concurrency.md
├── 0321842685.pdf
├── 13 Joe Duffy - The Error Model_files
├── 13 Joe Duffy - The Error Model.html
├── ada_language_server
├── api-guidelines
├── austral
├── austral.github.io
├── austral.rs
├── australspec
├── bfetch
├── cargo
├── cephes
├── codingprocess.md
├── compiler-builtins
├── crystal
├── c_standard_lib
├── c-ward
├── dmd
├── eudoxia0.github.io
├── fdlibm
├── files.zip
├── futhark
├── gingerBill.org
├── glibc
├── hashbrown
├── hylo
├── inko
├── julia
├── koika
├── kyokailang
├── kyokaiplan.md
├── langtest
├── language-server-protocol
├── ldc
├── libc
├── log
├── musl
├── nextstd
├── Nim
├── Odin
├── odin-lang.org
├── odinstudykyokai.md
├── paper.pdf
├── pkg.odin-lang.org
├── polonius
├── ponyc
├── POWER-KI
├── reason
├── rust
├── rust-analyzer
├── rust-clippy
├── rustfmt
├── rust-reference
├── rust-regex
├── rustup
├── scripts
├── specverifyinfo.md
├── standard
├── superfile
├── swift-evolution
├── v
├── wg-async
├── zig
└── zls

56 directories, 9 files
```

because currently the amount of spawned git instances were running wild on my system every time i entered the directories to read code.
and if i dare used any Ai features my entire cpu would get to 100% cpu usage
i think both this and a dedicated .zedignore would make sense because the current option to also remove the git scans hides the directories in the panel.
the zed log file is about 31md so i am unable to send it over here but from my scans it spawns about 3,341 git instances straight on my poor cpu killing any modicum of performance i have rn oom just takes em all out cuz i can't be having that.

PR #54025, opened 2026-04-16
sidebar: Add some UI adjustments

danilo-leal:
- Don't ever swap to the ellipsis menu with the close icon button; we now always have it
- Promote the "focus the last workspace" feature through the ellipsis menu
- Add a unified tooltip for the thread item to show relevant thread metadata
- Use a different icon for accessing the now "all threads" view
- Simplifies how we display archived threads
- Bonus: Don't display the "open in new window" button in currently active worktrees (in dedicated picker)
- Bonus: Use the "main worktree" label for whenever we're mentioning the original worktree

Release Notes:

- N/A

PR #54030, opened 2026-04-16
sidebar: Fix ellipsis menu options

danilo-leal:
Just a quick follow up to https://github.com/zed-industries/zed/pull/54025 where I incorrectly wrapped the "Remove Project" button in a conditional, when it should be available at all times.

Screenshot 2026-04-16 at 12  00@2x

Release Notes:

- N/A


eholk (approved):

eholk:
I rebased your PR onto main because there was a wasm compiler error that I fixed as part of #54023.

PR #54031, opened 2026-04-16
agent_ui: Serialize copied selection mentions as links

### LINKED ISSUE #53981: Copying prompt with code selections pastes only the word “selection” and discards original input in the Agent Panel
### Reproduction steps

1. Open a project in Zed and open the Agent panel (Codex CLI in my case).

2. In one or more files, select code ranges and use the agent chat input to do a kind of code review:
 ▫ select some code
 ▫ type a comment for that selection in the agent input
 ▫ repeat across multiple files so the input contains multiple selection+comment entries (like ‎`showWizard.ts (25:30) …`, ‎`index.ts (6:6) …`, etc.).
Image

3. After composing a longer review like this, select all text in the agent input and copy it (‎`cmd+A`, ‎`cmd+C`).

4. Start a new agent thread (e.g. click “New thread”).

5. Paste into the new thread’s input (‎`cmd+V`).

When I paste into the new thread, all the code selections I had referenced are gone. Instead of the actual selected code (or any useful representation of it), Zed pastes the literal word ‎`selection` in place of each snippet:
Image

On top of that, once I’ve started the new thread, there seems to be no way to get back to the previous thread’s unsent input. The long, carefully crafted review I wrote (with real selections visible in the UI) is effectively lost; only this degraded “selection + comment” form survives in the new thread.

So in practice:

- Copying from the agent input loses the actual code context.

- Starting a new thread discards the original input with no undo / way to recover, which in my case meant losing roughly an hour of work.

### Expected
I expect copying from the agent input to preserve the actual selected code (or a meaningful textual representation of it) instead of replacing it with the word ‎`selection`, and starting a new thread should not silently destroy the original draft or should at least provide an obvious way to recover it.

OS: macOS 26.3.1
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

hamza-paracha:
Fixes #53981

Selection mentions inserted from the agent panel are stored as folded placeholder text in the message editor. That meant a normal copy would put the literal `selection` token on the clipboard, and pasting into a new thread would lose the original mention context.

This teaches `MessageEditor` to serialize selected folded mentions as their mention links before writing to the clipboard. The surrounding text stays intact, and the existing paste path can rebuild the mentions in a fresh thread.

Release Notes:

- Fixed copied agent prompts replacing selection mentions with the literal word `selection`.


NeelChotai:
Nice! Thank you for the contribution.

NeelChotai (approved):

PR #54039, opened 2026-04-16
buffer_diff: Relax line count checks for word diffs

CnsMaple:
- relax line count checks in word-diff computation to permit unequal line counts while both counts must be within max_word_diff_line_count
- remove strict equality requirement: base_line_count == buffer_row_range.len()
- add test coverage for different line counts in word diffs: test_word_diff_modified_lines_with_different_line_counts

Self-Review Checklist:

Release Notes:

- fix(buffer_diff): Relax line count checks for word diffs

before:
Image_2026-04-16_10-10-42_uksitut4 2z2

after:
Image_2026-04-16_10-36-51_rgyifcku gct

Image_2026-04-16_10-45-30_crviekbg dmr


cameron1024:
Hi, thanks for the contribution :grin: 

In your screenshot, seeing the `, sr` get highlighted as deleted looks great, but I'm a bit concerned about the end of line `,;` being on the wrong line, since it's pretty misleading (it took me a few seconds to realise what was happening, at first I assumed it was a bug in our diffing algorithm :sweat_smile: ).

I think it's worth investigating why the deleted line is only being partially highlighted (and adding a test that captures this behaviour). Other than that, I think this is a good direction to take though.

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been waiting for a reply for over 3 weeks and has conflicts now.

PR #54051, opened 2026-04-16
editor: Fix soft-wrap in auto-height editors

osiewicz:
We had an internal report of soft wrap not working in git panel's commit editor. Given the following settings:
```json
{
  "languages": {
    "Git Commit": {
      "preferred_line_length": 80,
      "soft_wrap": "preferred_line_length",
    },
  },
}
```
We would not soft-wrap in narrow viewports. As it turned out, the problem was that we were always prefering a `preferred_line_length` as our soft wrap boundary over the actual width of the editor.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Fixed git commits editor not respecting soft wrap boundaries.
- settings: Removed `"soft_wrap": "preferred_line_length"` in favour of `"soft_wrap": "bounded"`. Soft wrap now always respects editor width when it's enabled.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1391,9 +1391,7 @@
   //      "soft_wrap": "prefer_line", // (deprecated, same as "none")
   // 2. Soft wrap lines that overflow the editor.
   //      "soft_wrap": "editor_width",
-  // 3. Soft wrap lines at the preferred line length.
-  //      "soft_wrap": "preferred_line_length",
-  // 4. Soft wrap lines at the preferred line length or the editor width (whichever is smaller).
+  // 3. Soft wrap lines at the preferred line length or the editor width (whichever is smaller).
   //      "soft_wrap": "bounded",
   "soft_wrap": "none",
   // The column at which to soft-wrap lines, for buffers where soft-wrap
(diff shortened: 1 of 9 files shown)


cole-miller:
Thanks! Since `SoftWrap::Bounded` already does the min internally, should we use that for the commit language, and perhaps delete `SoftWrap::Column`? (It seems like you would never want to soft wrap at a finite width that's greater than the editor width, to avoid this kind of misbehavior)

osiewicz:
Yup, that makes sense to me.

osiewicz:
@cole-miller I removed `Column` variant, but I refrained from tweaking the default for Git commit language (since Antonio's settings were non-default: our default right now is `editor_width`). Hope that's ok with you?

cole-miller:
That makes sense, thanks!

cole-miller (approved):

tariqkb:
This breaks an existing use case: 

```json
"preferred_line_length": 120000,
"soft_wrap": "preferred_line_length",
```

This is currently the only way in Zed to view very long lines without wrapping, AFAIK. When soft wrap is off, there is a hardcoded max.

It's not clear to me why `preferred_line_length` needed to be removed to address the initial bug report - that user incorrectly used `preferred_line_length` when they really wanted `bounded`. I'm not saying it makes sense that soft wrapping is the only way to view very long lines, but it has been like this in Zed for a long time.

PR #54055, opened 2026-04-16
sidebar: Move to new workspace non-Wayland

cameron1024:
We were just creating a window without showing it. However, wayland does not respect this setting, so it seemed to work.

Now we actually activate the window :)

Rather than changing the single call-site, we always do this when calling `Workspace::new_local`, since there were no code paths in the repo that pass `NewWindow` without immediately activating it

Release Notes:

### DIFF
--- a/crates/workspace/src/multi_workspace.rs
+++ b/crates/workspace/src/multi_workspace.rs
@@ -930,10 +930,16 @@ impl MultiWorkspace {
                 return Ok(());
             }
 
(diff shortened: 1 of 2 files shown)


nathansobo (changes requested):
I think we should bake this deeper in. Don't we always want to activate the window when we create a new local workspace.

maxbrunsfeld (approved):
Tested on macOS, and this fixes the issue for me.

PR #54061, opened 2026-04-16
Swap Kotlin LSPs in documentation to reflect actual default config

lixiaoyan:
## Summary

- The default settings in `assets/settings/default.json` disable `kotlin-language-server` (prefixed with `!`) and enable `kotlin-lsp` as the primary LSP, but the Kotlin documentation listed them the other way around. This swaps the order so the docs match reality.

## Test plan

- [x] Verified `assets/settings/default.json` has `["!kotlin-language-server", "kotlin-lsp", "..."]` for Kotlin
- [x] Confirmed the documentation now lists `kotlin-lsp` as the primary Language Server and `kotlin-language-server` as the alternate

Release Notes:

- N/A


smitbarmase (approved):
Thanks!

PR #54068, opened 2026-04-16
dev_container: Detect error state when docker ps returns duplicate containers

antont:
Self-Review Checklist:

- [x] Unsafe blocks (if any) have justifying comments — N/A, no unsafe blocks
- [x] Performance impact has been considered and is acceptable — N/A, no performance-sensitive changes

## Summary

Per the Dev Containers spec, the identifying labels `devcontainer.local_folder` + `devcontainer.config_file` should be unique per project. In practice two tools can end up with containers that share those labels under different Docker Compose projects — e.g. Zed and the reference [`@devcontainers/cli`](https://github.com/devcontainers/cli) / VS Code against the same folder, or Zed across impl/version boundaries (the Rust-native impl landed in v0.232.2; older Zed shelled out to the CLI and used its derivation). Root-cause write-up with fixture and captured output: https://github.com/antont/zed/issues/5.

When that happens, `docker ps --format={{ json . }}` returns newline-delimited JSON — one object per line — and Zed's previous lookup crashed on it. Silently picking the first value would have been worse: it'd connect Zed to an arbitrary one of the stale containers without the user knowing.

## What this PR does

Detects the multi-match state and surfaces it as a clear, user-facing error naming the duplicate IDs, instead of either crashing or silently choosing. The fix is scoped to detection only; no attempt here to resolve the duplicate state (see [antont/zed#6](https://github.com/antont/zed/pull/6) for the follow-up that eliminates the root cause and adds graceful recovery).

- `DevContainerError::MultipleMatchingContainers(Vec<String>)` — new variant with a `Display` that names the IDs and suggests `docker stop` / `docker rm`.
- `find_process_by_filters` now parses output through `parse_find_process_output`:
  - 0 rows → `Ok(None)`
  - 1 row  → `Ok(Some(_))`
  - ≥2 rows → `Err(MultipleMatchingContainers(ids))`
- The generic `evaluate_json_command` / `deserialize_json_output` helper in `command_json.rs` stays strict (single JSON value per call). NDJSON awareness is confined to `find_process_by_filters`, so the generic helper can't drift into silently-picking-first behavior elsewhere.
- `start_dev_container_with_config` passes the new error variant through unchanged, so its crafted `Display` reaches the UI prompt rather than being swallowed by `DevContainerUpFailed`.

## Why detect rather than resolve

Detection alone is defensive: it prevents connecting to the wrong container and tells the user exactly what's in Docker. Resolution (picking a canonical container, migrating namespaces, etc.) is a larger change that depends on fixing the derivation divergence upstream — tracked in the follow-up PR above.

## Test plan

- [x] `cargo test -p dev_container --lib` — all tests pass
- [x] New unit test `parse_find_process_output_returns_multiple_matching_containers` exercises the ≥2-rows branch and checks both the variant and the error message
- [x] Existing single-match / empty-output cases covered by `parse_find_process_output_parses_single_object` and the empty-output test
- [x] End-to-end with the compose fixture at https://github.com/antont/zed-devcontainer-compose-test: step §5 of its VERIFY.md reproduces the dual-label `docker ps` that returns two rows — with this PR built, opening the fixture in Zed surfaces the named error instead of crashing or silently connecting

Release Notes:

- Fixed dev container start silently connecting to a stale container when multiple containers matched the project's identifying labels; Zed now surfaces a clear error naming the duplicate IDs.

### DIFF
--- a/crates/dev_container/src/command_json.rs
+++ b/crates/dev_container/src/command_json.rs
@@ -52,7 +52,8 @@ where
         if raw.is_empty() || raw.trim() == "[]" || raw.trim() == "{}" {
             return Ok(None);
         }
-        serde_json_lenient::from_str(&raw).map_err(|e| format!("Error deserializing from raw json: {e}"))
+        serde_json_lenient::from_str(&raw)
+            .map_err(|e| format!("Error deserializing from raw json: {e}"))
     } else {
         let std_err = String::from_utf8_lossy(&output.stderr);
         Err(format!(
@@ -84,7 +85,12 @@ mod tests {
     fn test_deserialize_single_json_object() {
         let output = success_output(r#"{"id":"abc123"}"#);
         let result: Option<TestItem> = deserialize_json_output(output).unwrap();
-        assert_eq!(result, Some(TestItem { id: "abc123".into() }));
+        assert_eq!(
+            result,
+            Some(TestItem {
+                id: "abc123".into()
+            })
+        );
     }
 
     #[test]
--- a/crates/dev_container/src/command_json.rs
+++ b/crates/dev_container/src/command_json.rs
@@ -52,13 +52,62 @@ where
         if raw.is_empty() || raw.trim() == "[]" || raw.trim() == "{}" {
             return Ok(None);
         }
-        let value = serde_json_lenient::from_str(&raw)
-            .map_err(|e| format!("Error deserializing from raw json: {e}"));
-        value
+        serde_json_lenient::from_str(&raw).map_err(|e| format!("Error deserializing from raw json: {e}"))
     } else {
         let std_err = String::from_utf8_lossy(&output.stderr);
         Err(format!(
             "Sent non-successful output; cannot deserialize. StdErr: {std_err}"
         ))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::process::ExitStatus;
+
+    use super::*;
+
+    fn success_output(stdout: &str) -> Output {
+        Output {
+            status: ExitStatus::default(),
+            stdout: stdout.as_bytes().to_vec(),
+            stderr: Vec::new(),
+        }
+    }
+
+    #[derive(Debug, Deserialize, PartialEq)]
+    struct TestItem {
+        id: String,
+    }
+
+    #[test]
+    fn test_deserialize_single_json_object() {
+        let output = success_output(r#"{"id":"abc123"}"#);
+        let result: Option<TestItem> = deserialize_json_output(output).unwrap();
+        assert_eq!(result, Some(TestItem { id: "abc123".into() }));
+    }
+
+    #[test]
+    fn test_deserialize_newline_delimited_json_rejected() {
+        // Strict single-value contract: NDJSON must be rejected. Commands that
+        // may legitimately return multiple rows (e.g. `docker ps`) parse their
+        // output themselves rather than routing through this helper.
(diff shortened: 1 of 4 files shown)


KyleBarton:
@antont can you show an example of how you ended up in that state?

> since all containers in the compose project get the same devcontainer.local_folder and devcontainer.config_file labels.

I don't think that's the intended implementation, in Zed or in the reference implementation. The labels should apply only to the "main" service, as defined in the `service` field of devcontainer.json. It's the same reason why:

> find_process_by_filters returns Option<DockerPs> (singular) — it only needs one container ID to confirm a container exists. The caller (check_for_existing_container) then passes that ID to docker inspect for full details.

That's because Zed (or VSCode/whatever) needs to know which specific container to connect to. At that point, the broader docker compose infrastructure is opaque to the tool.

antont:
@KyleBarton oh you are right to ask, apparently my container config is a mess!

I did not look as it was working earlier and also now with official Zed 0.227.1 

I don't have time to check more now but an agent is telling this at least:

> ⏺ From the docker ps output, there are two app containers for the same project
  folder:
  1. xxx-devcontainer-app-1 — created by VS Code / devcontainer CLI (project name from the folder name)
  2. xxx-app-1 — created by Zed (project name xxx from the devcontainer.json name field)

>  Both have the same devcontainer.local_folder and devcontainer.config_file labels
  pointing to the same project, but they were created by different tools that
  generated different compose project names. So the filter matches both.

KyleBarton:
@antont got it - definitely a strange state to be in, but I don't think it's one that we should handle this way. These labels are supposed to be uniquely-identifying. The algorithm basically goes:
- Look for a container with these identifying labels
 - If one is present, use it (start it if it's in a stopped state)
 - If none are present, create it
 - If >1 are present, ??? <- we are here

I am all for better error messaging in that last case, so that the user can identify what's going on. But I think it's correctly classed as an error case. Zed doesn't know which container is valid to connect to in this case, and shouldn't proceed by just taking the first one. Do you have interest in transforming this PR into an improvement for error handling in this case? Otherwise, I think we should close

antont:
@KyleBarton 

> But I think it's correctly classed as an error case. Zed doesn't know which container is valid to connect to in this case, and shouldn't proceed by just taking the first one. Do you have interest in transforming this PR into an improvement for error handling in this case? Otherwise, I think we should close

Yes. I don't know the details of the dev container specs or anything, but based on your explanation I agree.

I am interested in doing that error handling -- it would have been helpful in my case, now I just didn't know that it had happened. It would be a nice exercise for me to make that change, and it would be useful if the same happens later. I don't know when I'll have the time but hopefully in say 3-4 days. If I fail to arrange that, I'll just close. I'll change this to a draft in the meantime.

antont:
@KyleBarton

I made a thorough investigation into how this happened for me, reported in https://github.com/zed-industries/zed/issues/54255 with repro instructions, using fixture repos I put up with docs with evidence from my repro runs.

I also have a fix which I'm using locally in my work build now, I'll submit a PR for consideration soon.

Before the investigation and the fix for the root cause, I also added the error dialog to this PR.

antont:
Okie the follow-up is there, with a somewhat crazy amount of things to ensure matching the original dev containers implementation, https://github.com/zed-industries/zed/pull/54302

antont:
Folded into #54302 per review — closing unmerged. The `MultipleMatchingContainers` detection functionality lands there.

Issue #37081, opened 2025-08-28
vtsls LSP suggestion sorting - classes have higher priority than local variables


smitbarmase:
### Summary

From: https://github.com/zed-industries/zed/issues/27993#issuecomment-3233115612

<img width="1594" height="956" alt="Image" src="https://github.com/user-attachments/assets/17583040-d493-49e8-811e-2fb8fda669e6" />

CakeWithDivinity:
I have taken a look at this myself, because it bothered me a lot. Here are my findings:
In this example:
```ts
import { Subscription } from 'rxjs';

const subscription = null;

console.log(s);
```
When putting the cursor at the `s` in the console log and requesting completion from the LSP, it returns the following items as the first two results:
```json
{
  "label": "subscription",
  "kind": 6,
  "sortText": "11",
  "data": { "cacheId": 1 },
  "textEdit": {
    "newText": "subscription",
    "insert": {...},
    "replace": {...}
  }
},
{
  "label": "Subscription",
  "kind": 6,
  "sortText": "11",
  "data": { "cacheId": 2 },
  "textEdit": {
    "newText": "Subscription",
    "insert": {...},
    "replace": {...}
  }
},
```

So, even though `Subscription` is a class, that gets imported from a dependency, the typescript LSP reports both as `Variable` according to the spec. The same thing happens for `typescript-language-server` (as `vtsls` does not add much functionality on top of it)

In Zed these two completions will be sorted by their label, because they differ in no other property: https://github.com/zed-industries/zed/blob/9ac511e47cf6dc313641cba7364de60ce78b737c/crates/editor/src/code_context_menus.rs#L1202

That's why `Subscription` ends up before `subscription`. When using std::cmp::Reverse on the `sort_label`, `subscription` comes first, but I don't think this is a feasible "fix" for this, as there might be languages where variables with uppercase letters should have higher priority?

foretoo:
@CakeWithDivinity hello! i think the right way to sort it is — most letter-matching should come first. For example, if LSP returns `subscription` and `Subscription` for letter `s` `subscription` should come first because of first letter matching.

foretoo:
hey y'all, I vibecoded a fix, PR #54072

foretoo:
@smitbarmase the issue can be closed, the fix is released in v1.0.0 🎉

Issue #27993, opened 2025-04-03
Poor LSP suggestion sorting (meta tracking issue)


probably-neb:
There are many instances currently where the top LSP auto-complete suggestions are either not relevant, miss very obvious completions (i.e. local variable names), etc. 

The goal of this issue is to track the various cases where the sorting/filtering of LSP provided auto-complete suggestions is not as good as it could be.

MrSubidubi:
@smitbarmase Thanks for starting to work on this. I would argue that https://github.com/zed-industries/zed/issues/25231 should also be added to the list, given that the error also emerges in the suggestion sorting code you just started working on. See the second part of https://github.com/zed-industries/zed/issues/29050#issuecomment-2818279428 as to why this happens.

smitbarmase:
Thanks @MrSubidubi, you're right, it should be handled as part of the fix. I'm working on writing comprehensive tests for this so that we don't accidentally break it again. It will be part of this PR: https://github.com/zed-industries/zed/pull/29159

SomeoneToIgnore:
https://github.com/zed-industries/zed/issues/25231 is the only one left, stale, so this meta issue seems to be done.

MihaelBercic:
Has something happened lately that has reverted the solution? Suggestions are worse off again.

![Image](https://github.com/user-attachments/assets/da7d4e4e-641a-48a4-8306-106d7a39c103)

smitbarmase:
Not a regression, but looks like a case that was never accounted for. Thanks for reporting.

With dash:
<img width="475" alt="Image" src="https://github.com/user-attachments/assets/01f1a0f9-6303-4e5d-b04b-dbacdb29297e" />

Without dash:
<img width="479" alt="Image" src="https://github.com/user-attachments/assets/5eeee2b8-7f0f-43c1-bbde-65db8ae0dce1" />

MihaelBercic:
> Not a regression, but looks like a case that was never accounted for. Thanks for reporting.
> 
[...]

PR #54072, opened 2026-04-16
Prefer exact case matches when breaking completion ties

foretoo:
## Summary

Fix completion ordering when two items are otherwise tied and only differ by letter case.

In cases like `abc` vs `ABC`, if the user types `a`, Zed should prefer `abc`. If the user types `A`, Zed should prefer `ABC`. This matches the expectation described in #37081 and #27993, where `subscription` should rank above `Subscription` for query `s`.

## What changed

In `CompletionsMenu::sort_string_matches`, I added a tie-breaker that prefers completions with more exact case-sensitive matches at the fuzzy-match positions.

This only applies after the existing higher-priority sort keys, so it does not replace fuzzy score, match positions, snippet ordering, or LSP `sortText`. It only resolves ambiguous ties more intuitively.

## Tests

Added regression coverage in `code_completion_tests` for abstract case-only examples:

- `a` prefers `abc` over `ABC`
- `A` prefers `ABC` over `abc`
- `ab` prefers `abc` over `ABC`
- `AB` prefers `ABC` over `abc`
- mixed-case multi-letter queries like `Ab` and `aB`

## Verification

Ran:

```bash
cargo test -p editor code_completion_tests
```

PS: all code and description is generated by Codex

Release Notes:

- Fixed completions not tie braking by case

### DIFF
--- a/crates/editor/src/code_completion_tests.rs
+++ b/crates/editor/src/code_completion_tests.rs
@@ -218,6 +218,77 @@ async fn test_sort_positions(cx: &mut TestAppContext) {
 }
 
 #[gpui::test]
+async fn test_case_sensitive_match_tie_breaker(cx: &mut TestAppContext) {
+    let completions = vec![
+        CompletionBuilder::variable("abc", None, "11"),
+        CompletionBuilder::variable("ABC", None, "11"),
+    ];
+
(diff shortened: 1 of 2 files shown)


foretoo:
@cla-bot check

Veykril (approved):
Thanks!

PR #45764, opened 2025-12-28
Enable edit prediction and Copilot for untitled buffers

tacshi:
Closes [#45631](https://github.com/zed-industries/zed/issues/45631)

Recording:

https://github.com/user-attachments/assets/a5143eb4-fae3-42a7-9d64-fb7c42ee97c2

Release Notes:

- copilot: Edit predictions now work in temporary files

Issue #45631, opened 2025-12-24
Fix that edit predictions do not work for buffers without files, like ones started from `workspace: new file`


Dima-369:
### Reproduction steps

1. Start Zed
2. Trigger the `workspace: new file` action
3. Type in `1+1=` to trigger the edit prediction

### Current vs. Expected behavior

# Current behavior

No edit predictions appear.

The status bar lists this:

<img width="342" height="188" alt="Image" src="https://github.com/user-attachments/assets/3cce25ce-8e08-4b6b-a3b6-d10ee6591a83" />

# Expected behavior

I expect edit predictions to also work in those buffers without a file association. The reason is that I sometimes paste in certain temporary content and then edit it in Zed. Saving it first to a file is a bit annoying, just to get edit predictions.

I suppose, the context for the edit prediction model should then be scoped to just this file, or can maybe still include other edit locations?

### Zed version and system specs

Zed: v0.217.3+stable.105.80433cb239e868271457ac376673a5f75bc4adb1 (Zed) 
OS: macOS 15.6.1
Memory: 16 GiB
Architecture: aarch64

### Relevant Zed settings

Happens with every edit prediction provider, I think.

### If you are using WSL on Windows, what flavor of Linux are you using?

None

Issue #20839, opened 2024-11-18
Emmet should trigger based on Language Selector


albedoa:
### Check for existing issues

- [X] Completed

### Describe the feature

I want to be able to do the following with the Emmet extension installed:

1. Open a new document.
2. Change the language selector to "HTML".
3. Type `html:5`.
4. Type `<tab>`.
5. Observe that my text has expanded into some HTML 5 boilerplate.

Currently, as far as I can tell, I have to save the file with an `.html` extension. So this feature would be useful both for unsaved files and for files that are not strictly of the HTML type or extension.

My apologies if I am using it wrong or if I missed a duplicate issue.

### Environment

Zed: v0.161.2 (Zed)
OS: macOS 14.4.1
Memory: 8 GiB
Architecture: aarch64

osiewicz:
For that we'd have to support running LSP on unsaved buffers first.

Issue #17098, opened 2024-08-29
JSON diagnostics/formatting - support unsaved buffers


krstp:
### Check for existing issues

- [X] Completed

### Describe the feature

When using JSON formatter I would like to see in the interface a pop-up window or a bubble that will communicate to the user what is wrong with the user provided snippet/code. Currently the formatter will just not proceed not providing really any user feedback.

### If applicable, add mockups / screenshots to help present your vision of the feature

![Screenshot 2024-08-29 at 11 38 15 AM](https://github.com/user-attachments/assets/7f11d253-ecb0-4851-b9ec-39303c8a41c6)

notpeter:
Not sure exactly what you're asking for.  Currently if there's a JSON error, it shows a squiggly underline which shows the error on hover.  Additionally you can open the Project Diagnostics Panel (cmd-shift-m or the little Red X in the lower left) and see all the errors at once.  You can also use the `editor: go to diagnostic` / `editor: go to previous diagnostic` action (default bind of F8/shift-F8) to have them pop up inline.

Is your request that we put errors in a big red box? Or that it should be more persistent and show the error while you're editing until you've fixed it?  Just want to be clear what you're looking for.

<img width="496" alt="Screenshot 2024-09-06 at 14 43 38" src="https://github.com/user-attachments/assets/125350b9-9cba-4478-b0a7-9302f4bbcf12">

krstp:
Sorry for a late reply, other things took priority; just getting to it now.

In summary, after retesting, in latest release the issue boils down to the input MUST be saved into a file - such as in your example these are persistent/saved `json` files.

If the JSON paste is not saved but only pasted into `untitled` buffer, regardless of specifying the code type `CMD+K M -> JSON`, the code will not format. Only after saving the file the code "action" shows up as in your screenshot and I am able to format the code.

There were other tickets in the same fashion, I would need to look for them.

Core issue: often times I do not want to be saving the JSON input, I only want to format it ad-hoc for clarity of reading. In such cases paste of the output JSON buffer into ZED and ad-hoc formatting would really help. In current state, per latest ZED iterations, in order to prettify the code the `json` input needs to be saved.

I guess, this would make this ticket a request to be able to format an unsaved JSON buffer.

I hope this clarifies. Thanks for your follow up.

MichaReiser:
[...]

PR #54082, opened 2026-04-16
lsp: Enable LSP features for unsaved buffers

aviatesk:
## Rationale

Currently Zed does not start language servers for unsaved buffers, so even after the user explicitly selects a language via the language selector, the buffer gets none of the LSP-driven features: no diagnostics, completions, hover, formatting, inlay hints, semantic tokens, document symbols, etc.

This is a long-standing pain point that surfaces whenever a user wants to ad-hoc work on a snippet without committing to a file path:

- #17098 — JSON diagnostics/formatting requested for unsaved buffers
- #20839 — Emmet expansion requested based on language selector for unsaved files
- See also [this comment on #17098](https://github.com/zed-industries/zed/issues/17098#issuecomment-2461977893) noting the same root cause affects the Ruff and ty extensions

These issues were closed as `NOT_PLANNED` during the Feb 2026 cleanup of older feature requests, but the underlying demand is real and broad (it affects every language server, not just any one extension).

Related prior work in the same direction:
- #45631 / #45764 — enabling edit predictions / Copilot for untitled buffers

## Summary

This PR makes Zed support the LSP `untitled:` URI scheme so that buffers without a backing file participate in the same LSP machinery as file-backed buffers. The trigger is the user selecting a language for the untitled buffer (via the language selector). The buffer is registered with the matching language server using the first visible worktree as the host; on save it is unregistered under `untitled:` and re-registered under `file://`.

## Self-Review Checklist

- [-] Unsafe blocks (if any) have justifying comments

Addresses #17098, #20839

Release Notes:

- Added LSP support for unsaved buffers — selecting a language for an untitled buffer now starts the matching language server and enables completions, diagnostics, hover, formatting, inlay hints, and other LSP features

### DIFF
--- a/crates/editor/src/completions.rs
+++ b/crates/editor/src/completions.rs
@@ -136,17 +136,20 @@ impl Editor {
         }
     }
 
-    pub(super) fn is_lsp_relevant(&self, file: Option<&Arc<dyn language::File>>, cx: &App) -> bool {
+    pub(crate) fn is_lsp_relevant(&self, buffer: &impl LspBufferContext, cx: &App) -> bool {
         let Some(project) = self.project() else {
             return false;
         };
-        let Some(buffer_file) = project::File::from_dyn(file) else {
-            return false;
+        let project = project.read(cx);
+        let Some(buffer_file) = project::File::from_dyn(buffer.file()) else {
+            return project
+                .lsp_store()
+                .read(cx)
+                .has_buffer_uri(buffer.remote_id());
         };
         let Some(entry_id) = buffer_file.project_entry_id() else {
             return false;
         };
-        let project = project.read(cx);
         let Some(buffer_worktree) = project.worktree_for_id(buffer_file.worktree_id(cx), cx) else {
             return false;
         };
--- a/crates/language_tools/src/lsp_button.rs
+++ b/crates/language_tools/src/lsp_button.rs
@@ -981,11 +981,18 @@ impl LspButton {
                 ..
             } => {
                 self.server_state.update(cx, |state, cx| {
+                    // For unsaved buffers `buffer_abs_path` carries an LSP URI
+                    // (e.g. "untitled:Untitled-1") instead of a filesystem path.
+                    // They have no real worktree, so skip them from this map.
+                    if update.buffer_abs_path.starts_with("untitled:") {
+                        return;
+                    }
+                    let path = Path::new(&update.buffer_abs_path);
                     let Ok(worktree) = state.workspace.update(cx, |workspace, cx| {
                         workspace
                             .project()
                             .read(cx)
-                            .find_worktree(Path::new(&update.buffer_abs_path), cx)
+                            .find_worktree(path, cx)
                             .map(|(worktree, _)| worktree.downgrade())
                     }) else {
                         return;
--- a/crates/language_tools/src/lsp_button.rs
+++ b/crates/language_tools/src/lsp_button.rs
@@ -1031,11 +1031,18 @@ impl LspButton {
                 ..
             } => {
                 self.server_state.update(cx, |state, cx| {
+                    // For unsaved buffers `buffer_abs_path` carries an LSP URI
+                    // (e.g. "untitled:Untitled-1") instead of a filesystem path.
+                    // They have no real worktree, so skip them from this map.
+                    if update.buffer_abs_path.starts_with("untitled:") {
+                        return;
+                    }
+                    let path = Path::new(&update.buffer_abs_path);
                     let Ok(worktree) = state.workspace.update(cx, |workspace, cx| {
                         workspace
                             .project()
                             .read(cx)
-                            .find_worktree(Path::new(&update.buffer_abs_path), cx)
+                            .find_worktree(path, cx)
                             .map(|(worktree, _)| worktree.downgrade())
                     }) else {
                         return;
--- a/crates/project/src/lsp_store.rs
+++ b/crates/project/src/lsp_store.rs
@@ -309,6 +309,16 @@ struct DocumentSelectorContext {
     scheme: &'static str,
 }
 
+/// Per-untitled-buffer LSP state. File-backed buffers don't need this — their
+/// URI and worktree are derivable from `file.abs_path()` / `file.worktree_id()`.
+#[derive(Debug, Clone)]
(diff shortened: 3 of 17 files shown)


OmerFI:
Any chance of a status update on this one? I ran into the same limitation from the JSON side: in an unsaved buffer `editor: format` falls back to Prettier, so pasted JSON keeps its blank lines and short objects stay collapsed. Setting `"formatter": "language_server"` fixes it, but only once the buffer is saved to disk, which defeats the point of a scratch tab.

The branch seems to conflict with `main` now and there's no review on it yet — is this still on the table, and is there a rough idea of when it might land?

crates/project/src/lsp_store.rs
```diff
@@ -13406,68 +13689,83 @@ impl LspStore {
...
+                    uri = match lsp::Uri::from_file_path(&abs_path) {
                         Ok(uri) => uri,
                         Err(()) => {
                             log::error!("failed to convert path to URI: {:?}", abs_path);
                             continue;
                         }
                     };
+                    uri_display = abs_path.to_string_lossy().into_owned();
+                } else {
+                    let Some(stored_uri) = local.buffer_uris.get(&buffer_id).map(|s| s.uri.clone())
```
SomeoneToIgnore, line 13727: This branch also needs to check the stored host worktree against `worktrees_using_server`.
If an untitled buffer is registered under visible root A and a matching server then starts in visible root B, this branch sends the buffer to B as well.
Unlike the file-backed branch, it never checks the host worktree, and adding B to `buffers_opened_in_servers` also routes subsequent requests to it.
Please cover both server-startup orders with two roots.

crates/project/src/lsp_store.rs
```diff
@@ -4986,13 +5104,17 @@ impl LspStore {
...
+                        local.reset_buffer(buffer, old_file, cx);
 
-                    if local.registered_buffers.contains_key(&buffer_id) {
-                        local.unregister_old_buffer_from_language_servers(buffer, old_file, cx);
+                        if local.registered_buffers.contains_key(&buffer_id) {
+                            local.unregister_old_buffer_from_language_servers(buffer, old_file, cx);
+                        }
+                    } else if local.registered_buffers.contains_key(&buffer_id) {
+                        if let Some(state) = local.buffer_uris.remove(&buffer_id) {
+                            local.unregister_buffer_from_language_servers(buffer, &state.uri, cx);
```
SomeoneToIgnore, line 5116: Please unregister against the servers that actually opened the untitled URI, rather than resolving them from the current file.
Start both roots’ servers before creating the untitled buffer under A, then save it into B: by this event the buffer already has B’s file, so `unregister_buffer_from_language_servers` searches B’s servers while discarding A’s snapshots.
A never receives `didClose`, leaving the old untitled document open.
The save test needs a cross-worktree case.

crates/project/src/lsp_store.rs
```diff
@@ -5574,13 +5715,26 @@ impl LspStore {
...
+                    );
+                    detached_abs_path = Some(abs_path);
+                }
+            } else if let Some(uri) = local_store
+                .buffer_uris
+                .get(&buffer_id)
+                .map(|state| state.uri.clone())
+            {
+                detached_servers =
+                    local_store.unregister_buffer_from_language_servers(buffer_entity, &uri, cx);
```
SomeoneToIgnore, line 5736: Please clear the old server’s buffer diagnostics when it is no longer attached.
For an untitled buffer, `detached_abs_path` remains `None`, so the diagnostic cleanup below is skipped, and unregistering does not clear diagnostics.
If A has published errors, remains running for another document, and the buffer switches to a language served by B, A’s errors remain even after B publishes an empty result.

crates/editor/src/completions.rs
```diff
@@ -136,17 +136,20 @@ impl Editor {
...
             return false;
         };
-        let Some(buffer_file) = project::File::from_dyn(file) else {
-            return false;
+        let project = project.read(cx);
+        let Some(buffer_file) = project::File::from_dyn(buffer.file()) else {
+            return project
+                .lsp_store()
+                .read(cx)
+                .has_buffer_uri(buffer.remote_id());
```
SomeoneToIgnore, line 148: This check needs registration state that is available on remote stores too.
`has_buffer_uri` returns false for every remote store because it only consults `as_local()`, but remote projects and collaboration guests use that mode.
Their untitled buffers are therefore excluded from the inlay-hint and document-symbol queries even when the host has registered them.

crates/project/src/lsp_store.rs
```diff
@@ -1424,6 +1435,11 @@ impl LocalLspStore {
                 .unwrap_or_else(|| file.path().clone());
             let worktree_path = ProjectPath { worktree_id, path };
             self.language_server_ids_for_project_path(worktree_path, language, cx)
+        } else if buffer.file().is_none() {
+            self.buffers_opened_in_servers
+                .get(&buffer.remote_id())
+                .map(|server_ids| server_ids.iter().copied().collect())
```
SomeoneToIgnore, line 1441: Please preserve the configured `language_servers` order rather than returning `HashSet` iteration order.
The file-backed path gets an ordered result from the server tree, while formatting with `"formatter": "language_server"` selects the first capable server from this list.
With two capable servers, an untitled buffer can therefore use the secondary formatter instead of the configured first choice.

crates/language_tools/src/lsp_button.rs
```diff
@@ -1031,11 +1031,18 @@ impl LspButton {
                 ..
             } => {
                 self.server_state.update(cx, |state, cx| {
+                    // For unsaved buffers `buffer_abs_path` carries an LSP URI
+                    // (e.g. "untitled:Untitled-1") instead of a filesystem path.
+                    // They have no real worktree, so skip them from this map.
+                    if update.buffer_abs_path.starts_with("untitled:") {
+                        return;
```
SomeoneToIgnore, line 1038: Skipping these registrations makes `Restart Server` a no-op when only untitled buffers are open.
The action and its fallback both collect candidates from `servers_per_buffer_abs_path`, so they find no buffers and never call `restart_language_servers_for_buffers`.
The server still appears in the menu through the language-server status fallback.
Please track buffer IDs for this action instead of requiring a filesystem path.

SomeoneToIgnore (changes requested):

PR #54092, opened 2026-04-16
Avoid popping up `conhost.exe` window on `Windows`

emmettlu:
Self-Review Checklist:

Closes #ISSUE

Release Notes:

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -7,6 +7,8 @@
     allow(dead_code)
 )]
 
+#![windows_subsystem = "windows"]
+
 use anyhow::{Context as _, Result};
 use clap::Parser;
 use cli::{CliRequest, CliResponse, IpcHandshake, ipc::IpcOneShotServer};


emmettlu:
It would not reach this issue if you are opening zed in a terminal shell.

It only accurs `Win` + `R` or any other third party launchers to exec programes in `PATH`.

reflectronic (changes requested):
Thanks for the pull request, but this isn't the right approach. The CLI needs to be a `console` subsystem executable so that it can print to the console when needed (e.g. `zed --foreground`).

The right solution is to use [Console Allocation Policy](https://learn.microsoft.com/en-us/windows/console/console-allocation-policy) to suppress the allocation of a new console when running the CLI outside of a shell. This will fix the Win + R etc. scenarios.

smitbarmase:
Thanks for the PR. It's been a while with no response, so I'll close this out as stale. Feel free to reopen if you'd like to pick it back up.

PR #54100, opened 2026-04-16
Support code lens in the editor

### LINKED ISSUE #11565: CodeLens with # of references, implementations etc.
### Check for existing issues

- [X] Completed

### Describe the feature

Code lens listing and allowing quick navigation to references and, in the case of a trait, to the implementation are very useful features and must-haves for an IDE. It would be great to have something like this.

As seen in VS Code for a trait

image

SomeoneToIgnore:
Closes https://github.com/zed-industries/zed/issues/11565

Adds the `editor: toggle code lens` action and the `"code_lens"` settings (disabled by default) which, when enabled, will make Zed to query the language server for code lens.

A language server might need more configuration (Zed has `"lsp"` settings section for this) first to start sending meaningful code lens responses back.

https://github.com/user-attachments/assets/eccd7ac4-b587-4241-a1bd-ff545a4cc602

Release Notes:

- Added LSP code lens support, disabled by default, use `"code_lens": "on"` settings to enable or the action to toggle them on

### DIFF
--- a/crates/editor/src/code_lens.rs
+++ b/crates/editor/src/code_lens.rs
@@ -1,0 +1,1084 @@
+use std::{collections::HashMap as StdHashMap, iter, ops::Range, sync::Arc};
+
+use collections::{HashMap, HashSet};
+use futures::future::join_all;
+use gpui::{MouseButton, SharedString, Task, WeakEntity};
+use itertools::Itertools;
+use language::{BufferId, ClientCommand};
+use multi_buffer::{Anchor, MultiBufferRow, MultiBufferSnapshot, ToPoint as _};
+use project::{CodeAction, TaskSourceKind};
+use settings::Settings as _;
+use task::TaskContext;
+use text::Point;
+
+use ui::{Context, Window, div, prelude::*};
+use workspace::PreviewTabsSettings;
+
+use crate::{
+    Editor, LSP_REQUEST_DEBOUNCE_TIMEOUT, MultibufferSelectionMode, SelectionEffects,
+    actions::ToggleCodeLens,
+    display_map::{BlockPlacement, BlockProperties, BlockStyle, CustomBlockId},
+};
+
+#[derive(Clone, Debug)]
+struct CodeLensLine {
+    position: Anchor,
+    indent_column: u32,
+    items: Vec<CodeLensItem>,
+}
+
+#[derive(Clone, Debug)]
+struct CodeLensItem {
+    title: SharedString,
+    action: CodeAction,
+}
+
+pub(super) struct CodeLensState {
+    pub(super) block_ids: HashMap<BufferId, Vec<CustomBlockId>>,
+    resolve_task: Task<()>,
+}
+
+impl Default for CodeLensState {
+    fn default() -> Self {
+        Self {
+            block_ids: HashMap::default(),
+            resolve_task: Task::ready(()),
+        }
+    }
+}
+
+impl CodeLensState {
+    fn all_block_ids(&self) -> HashSet<CustomBlockId> {
+        self.block_ids.values().flatten().copied().collect()
+    }
+}
+
+fn group_lenses_by_row(
+    lenses: Vec<(Anchor, CodeLensItem)>,
+    snapshot: &MultiBufferSnapshot,
+) -> impl Iterator<Item = CodeLensLine> {
+    let mut grouped: HashMap<MultiBufferRow, (Anchor, Vec<CodeLensItem>)> = HashMap::default();
+
+    for (position, item) in lenses {
+        let row = position.to_point(snapshot).row;
+        grouped
+            .entry(MultiBufferRow(row))
+            .or_insert_with(|| (position, Vec::new()))
+            .1
+            .push(item);
+    }
+
+    grouped
+        .into_iter()
+        .map(|(_, (position, items))| {
+            let row = position.to_point(snapshot).row;
+            let indent_column = snapshot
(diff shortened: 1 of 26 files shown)


crates/project/src/lsp_store/code_lens.rs
```diff
@@ -36,7 +37,50 @@ impl CodeLensData {
...
+            lsp_data.code_lens = None;
+        }
+    }
+
+    /// Fetches all code lenses for the buffer and resolves the ones within
+    /// `resolve_range`. Returns ALL lenses for the buffer (not filtered).
     pub fn code_lens_actions(
+        &mut self,
+        buffer: &Entity<Buffer>,
+        resolve_range: Range<Anchor>,
```
Veykril: nit, would it make sense to do the resolve outside of this part? In theory, a low language server could take a while until it responds with the initial lenses, at which point the user might've scrolled away by quite a bit, causing us to unnecessarily try to resolve the out of view lenses now

crates/editor/src/code_lens.rs
```diff
@@ -0,0 +1,1084 @@
...
+            let indent_column = snapshot
+                .indent_size_for_line(multi_buffer::MultiBufferRow(row))
+                .len;
+            CodeLensLine {
+                position,
+                indent_column,
+                items,
+            }
+        })
+        .sorted_by_key(|lens| lens.position.to_point(snapshot).row)
```
Veykril: Can be simplified, also reduces the amount of `to_point` calls from 3 to 1 per entry
```suggestion
    lenses
        .into_iter()
        .into_group_map_by(|(position, _)| {
            let row = position.to_point(snapshot).row;
            MultiBufferRow(row)
        })
        .into_iter()
        .sorted_by_key(|&(row, _)| row)
        .map(|(row, (position, items))| {
            let indent_column = snapshot
                .indent_size_for_line(row)
                .len;
            CodeLensLine {
                position,
                indent_column,
                items,
            }
        })
```

crates/editor/src/code_lens.rs
```diff
@@ -0,0 +1,1084 @@
...
+}
+
+fn render_code_lens_line(
+    line_number: usize,
+    lens: CodeLensLine,
+    editor: WeakEntity<Editor>,
+) -> impl Fn(&mut crate::display_map::BlockContext) -> gpui::AnyElement {
+    move |cx| {
+        let mut children: Vec<gpui::AnyElement> = Vec::new();
```
Veykril:
```suggestion
        let mut children: Vec<gpui::AnyElement> = Vec::with_capacity((1 * lens.items.len()).saturating_sub(1));
```

crates/editor/src/code_lens.rs
```diff
@@ -0,0 +1,1084 @@
+use std::{collections::HashMap as StdHashMap, iter, ops::Range, sync::Arc};
```
Veykril:
```suggestion
use std::{iter, ops::Range, sync::Arc};
```
There should be no reason to use std hash map here

Veykril (approved):

mustakimali:
Thanks 👏🏽  - can't wait for the release.

nicojerome:
Thanks for adding this; it’s really helpful.

I noticed that when opening and scrolling through a file, lines jump around. It seems the "n references" annotations are loaded async and inserted above functions, causing layout shifts.

Not sure if this is worth opening an issue for, but it’s a bit distracting. Maybe showing the "n references" annotations inline (at the end of the same line as the function name, similar to git blames) instead of above could avoid the jumping.

What do you think, @SomeoneToIgnore?

SomeoneToIgnore:
I think we should understand why the jumps happen before proposing any solutions at all.

[The spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens) clearly states that code lens are sent per document, so whatever jumps occur, occur due to the resolving approach?

We cannot show a lens before we get its `command` and presumably some servers omit that field until a separate resolve request is issued — we could enlarge the area we request the resolves for; or check whether the resolving is not written buggy which somehow makes it slow.
Maybe, something else is in play, hard to say without checking for more, but worth an issue created if e.g. VSCode handles them better.

Also note that we have a way to hide lens into action list already, see the related settings.

nicojerome:
I made a screen recording below. These layout shifts happen on every file for me. Its actually so distracting that I had to turn off code lens completely.

https://github.com/user-attachments/assets/72c7819c-2cb8-4d19-b739-6c3f52e7605c

SomeoneToIgnore:
Would you be able to share your project (and a language server, if nonstandard)?

Alternatively, try https://github.com/zed-industries/zed/tree/kb/lens-placeholders branch to see if it helps?

nicojerome:
> Would you be able to share your project (and a language server, if nonstandard)?

I can’t share the full project, but I can share the sample file ([repro.js](https://github.com/user-attachments/files/27472315/repro.js)). I verified that the same issue happens when I open just that file in a new Zed instance instead of the entire project. It’s plain JavaScript, and I don’t use any custom LSPs or extensions, just standard Zed. I’ve also attached my Zed settings in case that helps: [zed-settings.json](https://github.com/user-attachments/files/27472365/zed-settings.json)

> Alternatively, try https://github.com/zed-industries/zed/tree/kb/lens-placeholders branch to see if it helps?

That one definitely helps. The "n references" annotations still appear delayed, but no more vertical layout jumping:

https://github.com/user-attachments/assets/05951f12-9b16-4cb4-bd8f-98225ba88ff9

PR #54102, opened 2026-04-16
editor: Make the multiline comment folding more robust

### LINKED ISSUE #53606: Regression: Zero-indentation comments consumed by code folds in Python (since #50049)
## Description

Since v0.230.0, standalone comments at zero indentation in Python files are swallowed by the fold of the code block above them. Previously, these comments remained visible when code was folded, which is essential for using them as section separators between classes/functions.

## Steps to reproduce

1. Open a Python file with this structure:

```python
class Foo:
    def bar(self):
        pass


# SECTION SEPARATOR

class Baz:
    def qux(self):
        pass
```

2. Fold class Foo
3. The # SECTION SEPARATOR comment disappears into the fold

## Root cause

PR #50049 added logic in the indent-based fold scanner to skip lines that fall inside a @comment, @comment.inclusive, or @string tree-sitter scope when determining fold boundaries. This was intended to fix folding for unindented content within multiline strings and block comments.

However, Python’s tree-sitter overrides mark every # comment as @comment.inclusive, including standalone top-level separator comments. The scanner can’t distinguish between a comment that’s part of an indented block and a standalone section separator at column 0, so it extends the fold past all of them.

## Suggested fix

The fold scanner should check whether the comment node is actually a *descendant* of the syntax node being folded, rather than blanket-skipping all lines with a comment scope. A standalone top-level comment between two class definitions is not part of either class and should terminate the fold.

#### Current

Image

#### Expected

The zero-indentation comment should remain visible after folding the class above it, as it did before v0.230.0. These comments are not part of the class, they sit between top-level definitions and serve as navigational landmarks.

Image

Zed: v0.231.1+stable.213.e71b6aa44ffaac9bdaf986288289359e6f1b3572 (Zed) 
OS: Linux X11 pop 22.04
Memory: 62.6 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3080 Ti Laptop GPU || NVIDIA || 580.126.18

feitreim:
Basically just applied the suggested fix as described in #53606 , I also experimented with a hackier soln. that used the same logic as the code this is replacing, but it seemed solidly worse than this version.

The original issue extremely clear with what is going on / what needed to be fixed.

heres a video: 
https://github.com/user-attachments/assets/e7fc60f7-0a8c-403c-877f-93f58f370b00

Self-Review Checklist:

Closes #53606 

Release Notes:

- editor: Fixed incorrect inclusion of comments into folds

### DIFF
--- a/crates/editor/src/display_map.rs
+++ b/crates/editor/src/display_map.rs
@@ -2281,12 +2281,25 @@ impl DisplaySnapshot {
             let max_point = self.buffer_snapshot().max_point();
             let mut closing_row = None;
 
+            // End byte of the smallest syntactic node enclosing `buffer_row`.
+            // Used to tell standalone top-level comments (which terminate the
+            // fold) apart from unindented content inside a multi-line string
+            // or block comment belonging to the folded node (which does not).
+            let foldable_node_end = {
+                let row_start = Point::new(buffer_row.0, 0);
+                let row_end =
+                    Point::new(buffer_row.0, self.buffer_snapshot().line_len(buffer_row));
+                self.buffer_snapshot()
+                    .syntax_ancestor(row_start..row_end)
+                    .map(|(_, range)| range.end)
+            };
+
             for row in (buffer_row.0 + 1)..=max_point.row {
                 let line_indent = self.line_indent_for_buffer_row(MultiBufferRow(row));
                 if !line_indent.is_line_blank()
                     && line_indent.raw_len() <= start_line_indent.raw_len()
                 {
-                    if self
+                    let in_string_or_comment_scope = self
                         .buffer_snapshot()
                         .language_scope_at(Point::new(row, 0))
                         .is_some_and(|scope| {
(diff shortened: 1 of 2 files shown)


tomhoule:
@zed-zippy autofix

tomhoule:
There is a rustfmt check failure, but the substance of the PR is good. Once that's fixed I think we can merge. Thank you for the contribution!

feitreim:
should be fixed now. (well I ran `cargo fmt --all` and the clippy script atleast). I'll keep an eye on the test runner, but im not sure if that'll run on its own.

machin3io:
Thank you so much, I really appreciate it :)

feitreim:
@machin3io of course! your bug report made it so easy I had to do it.

bennetbo:
@zed-industries/approved

tomhoule:
@bennetbo thanks, I forgot to approve before merging 🙈

PR #54115, opened 2026-04-16
Preserve focused tool calls in agent panel when scrolling them out of view

maxbrunsfeld:
Release Notes:

- Fixed a bug where the agent panel would sometimes close if you scrolled up or down while it was zoomed.


bennetbo:
@zed-industries/approved

PR #54117, opened 2026-04-16
branch_picker: Ensure create buttons have unique IDs

danilo-leal:
Little bug I noticed that was preventing the buttons in the branch picker's footer from working correctly.

Release Notes:

- Fixed a bug in the branch picker where the buttons wouldn't properly get activated.


Anthony-Eid:
@zed-industries/approved

PR #54112, opened 2026-04-16
fuzzy_nucleo: Optimize path matching with CharBag prefilter and add benchmarks

feitreim:
This PR was originally a part of https://github.com/zed-industries/zed/pull/53551 so theres more info about its motivation there.

- Add a CharBag prefilter on path candidates to skip irrelevant entries before invoking nucleo's matcher.
- Use binary_search on sorted matched char indices when reconstructing byte positions (perf improvement).
- Add a criterion benchmark comparing `fuzzy_nucleo` path matching against the existing fuzzy crate.

Performance Chart:

| Benchmark | Size | Nucleo (before) | Nucleo (after) | Fuzzy | Before/Fuzzy | After/Fuzzy |
  |-----------|-----:|----------------:|---------------:|------:|-------------:|------------:|
  | 1-word | 100 | 14.14 µs | 9.12 µs | 9.06 µs | 1.56x | 1.01x |
  | 1-word | 1,000 | 164.37 µs | 114.11 µs | 110.43 µs | 1.49x | 1.03x |
  | 1-word | 10,000 | 1.83 ms | 1.39 ms | 1.41 ms | 1.30x | 0.99x |
  | 2-word | 100 | 12.83 µs | 3.51 µs | 979 ns | 13.10x | 3.59x |
  | 2-word | 1,000 | 131.65 µs | 33.46 µs | 6.37 µs | 20.67x | 5.25x |
  | 2-word | 10,000 | 1.24 ms | 338.84 µs | 52.46 µs | 23.64x | 6.46x |o

Exact Current State:
| query | size | nucleo | fuzzy | nucleo/fuzzy |
  |---|---:|---:|---:|---:|
  | 1-word | 100 | 8.62 µs | 9.22 µs | 0.93× |
  | 1-word | 1000 | 102 µs | 111 µs | 0.92× |
  | 1-word | 10000 | 1.13 ms | 1.28 ms | 0.88× |
  | 2-word | 100 | 3.48 µs | 0.98 µs | 3.55× |
  | 2-word | 1000 | 29.9 µs | 6.39 µs | 4.68× |
  | 2-word | 10000 | 271 µs | 53.4 µs | 5.08× |
  | 4-word | 100 | 0.85 µs | 0.53 µs | 1.60× |
  | 4-word | 1000 | 2.99 µs | 1.66 µs | 1.80× |
  | 4-word | 10000 | 20.1 µs | 9.14 µs | 2.20× |

Self-Review Checklist:

[...]

PR #54123, opened 2026-04-16
fuzzy_nucleo: Add strings module and route several pickers through it

feitreim:
Stacked on top of #54112
This is part 2 of 3 towards #51197
More details from the original PR #53551

This PR includes the changes from #54112 , im not sure how to avoid that, my understanding is that after that one is merged, this PR can be rebased onto main and everything will be correct. You can also view the version of this that does reflect the changes more directly here: https://github.com/feitreim/zed/pull/1

## Changes

In this PR I added a more general string matching functionality to `fuzzy_nucleo`, in order to have proper testing for this, I also changed the command palette, tab switching picker, branch picker, and recent projects picker to use this new implementation. I think the command palette change in particular is awesome, just super nice to vaguely gesture at the command i want and have it pop right up.

The main change here and departure from https://github.com/zed-industries/zed/pull/37123 is realizing that the primary reason for the regressions is actually how nucleo handles smart case, the old `fuzzy` crate only uses the smart case argument to score things differently, while nucleo actually filters on the case, eg. with smart case query "Apple" wouldnt match "apple". To get around this we always pass `CaseMatching::Ignore` to nucleo and implement the same score modifications from fuzzy in our code.

There is a performance cost to that, of course, but from my testing it is fairly static, not growing as the size increases, so maybe a query takes 35 µs instead of 25 µs, but a query that takes 800 µs will only take 820 µs.

Benchmark:
| kind | query | size | nucleo | fuzzy | nucleo/fuzzy |
  |---|---|---:|---:|---:|---:|
  | string | 1-word | 100 | 9.15 µs | 24.6 µs | 0.37× |
  | string | 1-word | 1000 | 150.2 µs | 207.2 µs | 0.72× |
  | string | 1-word | 10000 | 1.34 ms | 2.07 ms | 0.65× |
  | string | 2-word | 100 | 5.16 µs | 2.94 µs | 1.75× |
  | string | 2-word | 1000 | 29.0 µs | 11.0 µs | 2.63× |
  | string | 2-word | 10000 | 210.6 µs | 55.5 µs | 3.79× |
  | string | 4-word | 100 | 2.57 µs | 2.33 µs | 1.10× |
  | string | 4-word | 1000 | 6.98 µs | 5.85 µs | 1.19× |
  | string | 4-word | 10000 | 20.0 µs | 12.0 µs | 1.66× |

When I added the 4-word queries to the benchmarks I was actually really concerned that the performance would be awful, making it unsuitable for the command palette especially. However, I think due to the CharBag pre-filtering when the query is longer, the performance is actually way better than the 2 word case.

Video:

https://github.com/user-attachments/assets/3cd7221b-424f-4fd3-8df1-5543dcc340a3

Self-Review Checklist:

Release Notes:

- Improved fuzzy matching in the command palette, branch picker, tab switcher, and recent projects picker to support multi-word queries.


feitreim:
@yara-blue here is Part B, it seems like I will need to make somekind of rebase change in the future once part A is done, so ill be sure to keep my eye on that.

yara-blue:
> @yara-blue here is Part B, it seems like I will need to make somekind of rebase change in the future once part A is done, so ill be sure to keep my eye on that.

No worries I can do that (just did). Reviewing now, first one is in already lets see if we can get this one in today as well. Will be lovely to have nucleo's order independent fuzzy in more matchers.

yara-blue (approved):

dinocosta (approved):

redactedscribe:
Always appreciate improvements to fuzzy finding, however each time some land it seems like the Outline misses out. I wonder why.

feitreim:
@redactedscribe right now only like 4-5 of the ~50 pickers in zed have been switched over to this new fuzzy finder! if these go well I'm not sure exactly the plan but I imagine most of the rest could be switched over fairly easily.

yara-blue:
@redactedscribe implementation is now tracked in https://github.com/zed-industries/zed/issues/55872. 

If you want to help by adding a few more pickers let us know in the tracking issue!

PR #54126, opened 2026-04-16
agent_ui: Adjust thread item component and fix thread switcher

danilo-leal:
This PR adds some adjustments to the thread item component so metadata labels (project, worktree, and branch names) properly show up and truncate in different places, as well as fixes the thread switcher by making hover auto-select the item, making it more consistent with the regular tab switcher. We still have to figure out how to make the modal dismiss on click, though; that's pending.

Ended up also cleaning up the thread item's tooltip a bit, and tweaking the preview examples we have for it.

Release Notes:

- Agent: Fixed worktree and branch labels not showing up in the thread switcher.
- Agent: Fixed the thread switcher not selecting on hover.

### DIFF
--- a/crates/sidebar/src/thread_switcher.rs
+++ b/crates/sidebar/src/thread_switcher.rs
@@ -249,9 +249,9 @@ impl Render for ThreadSwitcher {
                             this.select_index(ix, cx);
                         }
                     }))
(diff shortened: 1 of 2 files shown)


eholk:
@zed-industries/approved

PR #54128, opened 2026-04-16
sidebar: Fix cmd-click in the header not taking to the last workspace

danilo-leal:
Fixes a bug where we weren't properly recording the last active workspace to power the cmd-click interaction in the sidebar's header. This PR introduces a field in the sidebar struct for that, allowing to store that value and to come back to it when clicking on the header, which is the single caller of the function introduced here.

Release Notes:

- Agent: Fixed a bug where cmd-clicking on the project header wouldn't actually take you to the last active workspace.


Anthony-Eid (approved):
This looks good, but I think it would be better if the multi workspace was able to track it's last active workspace

PR #54135, opened 2026-04-17
Fix double borrow panic when ACP process dies

maxbrunsfeld:
Release Notes:

- Fixed a crash that would occur when an ACP server process exited unexpectedly.


crates/agent_servers/Cargo.toml
```diff
@@ -40,6 +40,7 @@ log.workspace = true
 project.workspace = true
 release_channel.workspace = true
 remote.workspace = true
+piper = { version = "0.2", optional = true }
```
eholk: Should this be a workspace dependency?

eholk (approved):

crates/agent_servers/Cargo.toml
maxbrunsfeld: Yes, thanks for catching that.

maxbrunsfeld: Actually, I realized we already have a different crate for this purpose in the graph (`async-pipe`). Glad you pointed this out.

PR #54143, opened 2026-04-17
project: Add fuzzy search in remote projects dialogue

feitreim:
Stacked on top of #54123
This is part 3 of 3 towards #51197
More details from the original PR https://github.com/zed-industries/zed/pull/53551

This PR implements fuzzy search in the remote projects dialogue, using the `fuzzy_nucleo::match_strings` introduced in #54123.

## remote project search

The reviews of https://github.com/zed-industries/zed/pull/51197 seemed to indicate that performance was important here so it runs async and cancels the previous stale request when a new character is typed.

other features:
text highlighting of what was being typed, similar to other fuzzy searches.
reorder the projects on the list based on the maximum score from its projects.

video of it in action:

https://github.com/user-attachments/assets/86b3ed85-fc29-46fb-b3c4-c0cbc3bf8edd

Self-Review Checklist:

Closes #53318

Release Notes:

- recent_projects: Added fuzzy search to the remote projects modal.

### DIFF
--- a/crates/recent_projects/src/remote_servers.rs
+++ b/crates/recent_projects/src/remote_servers.rs
@@ -14,7 +14,7 @@ use extension_host::ExtensionStore;
 use futures::{FutureExt, StreamExt as _, channel::oneshot, future::Shared};
 mod filter;
 
-use filter::FilterData;
+use filter::{FilterData, FilteredServer};
 use gpui::{
     Action, AnyElement, App, ClickEvent, ClipboardItem, Context, DismissEvent, Entity,
     EventEmitter, FocusHandle, Focusable, PromptLevel, ScrollHandle, Subscription, Task,
@@ -631,7 +631,6 @@ impl From<WslServerIndex> for ServerIndex {
 struct ProjectEntry {
     navigation: NavigableEntry,
     project: RemoteProject,
-    highlight_positions: Vec<usize>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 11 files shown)


IOAyman:
Is there an ETA for this getting merged?

yara-blue:
A bit over a week :)

crates/recent_projects/src/remote_servers.rs
```diff
@@ -732,18 +775,210 @@ impl DefaultState {
...
+
+        let entry = server_map[meta.server_index].get_or_insert_with(|| FilteredServer {
+            server_index: meta.server_index,
+            host_positions: Vec::new(),
+            project_matches: Vec::new(),
+            score: f64::NEG_INFINITY,
+        });
+
+        entry.score = entry.score.max(m.score);
+
```
yara-blue: could you add something explaining why we are mutating the vector while iterating over it? I do not see why it's needed and it makes the whole process quiet complex.

yara-blue (changes requested):
The `build_filter_results` part is hard to understand, could you explain that a little or refactor it so it's easier to reason about?

It makes senses to at least move all the filter related stuff into a sub-module. That would also be a good place to document how it works. As I understand it we:
- cache a string representation of the server options (FilterData) 
- run fuzzy search on that 
- then "unpack" that back into the server list
- have two server lists, one filtered and one unfiltered


Otherwise it looks really nice btw!

yara-blue:
Also thank you for your patience @feitreim! I'm sorry it took a while.

feitreim:
@yara-blue thanks for looking at it! moved the filter stuff into a filter submodule that has some more comments explaining things. Split the result building logic so that mutating the vector while iterating over it bit makes more sense. basically just building buckets of results because of the host -> project structure. that logic is now group_matches_by_server, build_server_result and build_filter_results

yara-blue (approved):

PR #54165, opened 2026-04-17
markdown: Render inline <br> tags as line breaks

davidalecrim1:
## Summary
Inline `<br>` tags in markdown (e.g. `<br>`, `<br/>`, `<br />`) were rendered as literal text instead of line breaks when images are around. VS Code and GitHub both treat them as hard line breaks.

The fix has two layers:

**Parser (`parser.rs`):** The text-merge loop was consuming `InlineHtml` events as empty text. `<br>` variants are now excluded from that loop and converted to `HardBreak` events instead, matching the semantics of a trailing two-space or backslash line break in CommonMark.

**Renderer (`markdown.rs`):** Image paragraphs use a flex row layout, so text `\n` has no effect. `HardBreak` inside an image paragraph now inserts a `div().w_full().h_0()` child, which forces a flex-wrap row break. Images also get `mr_2()`/`mb_2()` margins for spacing, replacing the previous `SoftBreak`-as-space hack. `SoftBreak` is suppressed inside image paragraphs to avoid phantom whitespace elements disrupting the flex layout.

### Before

How Zed WAS rendering this:
Screenshot 2026-04-17 at 07 53 26

### Comparison

This is how Github renders this:
Screenshot 2026-04-17 at 07 06 49

This is how VS Code renders this:
Screenshot 2026-04-17 at 07 03 36

### After

This is the end result after the changes:
Screenshot 2026-04-17 at 07 45 32

[Created this Github repo for testing
](https://github.com/davidalecrim1/zed-md-test)

Self-Review Checklist:

Release Notes:

- Fixed inline `<br>` tags in markdown previews rendering as literal text instead of line breaks

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -1350,6 +1350,8 @@ impl MarkdownElement {
                 .min_w_0()
                 .max_w_full()
                 .rounded_md()
+                .mr_2()
+                .mb_2()
                 .when_some(height, |this, height| this.h(height))
                 .when_some(width, |this, width| this.w(width))
                 .with_fallback(move || image_fallback_element(dest_url.clone(), alt_text.clone())),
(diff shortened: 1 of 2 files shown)


smitbarmase (approved):
Thanks! I made a few architectural changes that sit better with the direction we want to go.

davidalecrim1:
@smitbarmase thanks for the review and the changes!

PR #54168, opened 2026-04-17
editor: introduce Bookmarks

yara-blue:
Adds basic bookmark functionality to the editor, allowing users to mark lines and later navigate between them. This is an MVP and will later be expanded with a picker, vim marks integration and syntax tree based bookmark positions. In this MVP bookmarks shift under external edits.

# UI
## Adding/Removing bookmarks
To add a bookmark:
- run the toggle bookmark action 
- hold secondary and click in the gutter
- open the context menu by right clicking in the gutter and select add bookmark
To remove a bookmark:
- run the toggle bookmark action 
- click on the bookmarks icon in the gutter
- open the context menu by right clicking in the gutter and select remove bookmark

remove all bookmarks with `workspace: clear bookmarks`

# Implementation
This mirrors the implementation of breakpoints. The rendering of the gutter was refactored to make place for bookmark icons and buttons:
- Code was extracted to a `Gutter` struct
- Runnables, breakpoints and bookmarks are now collected ahead of layouting. Just before layouting we remove the items that collide and do not have priority.
- The `phantom_breakpoint` is replaced by a `gutter_hover_button`

## In depth phantom breakpoint discussion:
This was phantom_breakpoint. It worked as follows:
 - A fake breakpoint was added to the list of breakpoints.
 - While rendering the breakpoints it a breakpoint turned out to be fake
      it would get a different description and look.
 - The breakpoint list was edited run_indicators ("play buttons")
      rendering to removes the fake breakpoint if it collided.

This would not scale to more functionality. Now we only render
breakpoints, bookmarks and run indicators. Then we render a button if
there is not breakpoint, bookmark or run indicator already present. We
can do so since the rendering of such "gutter indicators" has been
refactored into two phases:
 - collect the items.
 - render them if no higher priority item collides.

This is far easier and more readable which enabled me to easily take the
phantom_breakpoint system and use it for placing bookmarks as well :)

Note: this was previously merged but it needed a better squashed commit message. For the actual PR see: https://github.com/zed-industries/zed/pull/51404 

Release Notes:

- Added Bookmarks


yara-blue:
DO NOT MERGE, co-author not set etc

yara-blue:
replaced with: https://github.com/zed-industries/zed/pull/54174

PR #54170, opened 2026-04-17
agent: Parse reasoning tags

RedBaron1914:
### Description
This PR adds native support for parsing tags like `<think>`, `<thinking>`, and `<thoughts>` directly from streaming LLM responses. This allows custom proxy providers and locally hosted models (e.g., DeepSeek-R1 via OpenAI-compatible endpoints) to seamlessly render their reasoning process inside Zed's native `Thinking` UI blocks, preventing the reasoning from mixing with code or normal text.

### Changes
1. Added `parse_reasoning_tags` and `drop_reasoning_blocks` toggles to the Add Custom Provider UI and `OpenAiCompatibleAvailableModel` settings.
2. Implemented a robust `extract_thinking_from_stream` parser in `language_model.rs` that works across network chunk boundaries without breaking normal text blocks.
3. Updated `open_ai_compatible` provider to apply the stream mapping conditionally.
4. Dropped `MessageContent::Thinking` from the outgoing context history when `drop_reasoning_blocks` is enabled, saving massive amounts of context window tokens.
5. Added unit tests for the stream parser.

### Motivation
With reasoning models becoming the industry standard, many users rely on proxy APIs (like Gemini-2.5-pro or DeepSeek-R1) which return thoughts directly in the stream text. This PR ensures that Zed's native thinking UI works universally for any OpenAI-compatible endpoint that emits reasoning tags, without hardcoding this behavior for all models globally.

Self-Review Checklist:

Release Notes:

- Added `parse_reasoning_tags` and `drop_reasoning_blocks` options for OpenAI-compatible providers, enabling native thinking UI support for streaming reasoning tags.

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -1933,10 +1933,12 @@ impl Thread {
 
             log::debug!("Calling model.stream_completion, attempt {}", attempt);
 
+
             let (mut events, mut error) = match model.stream_completion(request, cx).await {
-                Ok(events) => (events.fuse(), None),
-                Err(err) => (stream::empty().boxed().fuse(), Some(err)),
+                Ok(events) => (language_model::extract_thinking_from_stream(events).fuse(), None),
+                Err(err) => (futures::stream::empty().boxed().fuse(), Some(err)),
             };
+
             let mut tool_results: FuturesUnordered<Task<LanguageModelToolResult>> =
                 FuturesUnordered::new();
             let mut early_tool_results: Vec<LanguageModelToolResult> = Vec::new();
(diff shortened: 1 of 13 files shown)


RedBaron1914:
@cla-bot check

benbrandt:
GIven this is a provider issue usually, I am not sure we want to add the parsing overhead for _every_ provider, especially since the main ones most people are using don't have this issue

RedBaron1914:
> GIven this is a provider issue usually, I am not sure we want to add the parsing overhead for _every_ provider, especially since the main ones most people are using don't have this issue

Hi, thanks for review, I totally understand the concern about performance and overhead, which is exactly why it's designed to be strictly opt-in and isolated. There is no overhead, and it's not for "every" provider. It's literally opt-in feature(so zero overhead by default) in UI when creating new provider, which creates no overhead when not in use. It's also in isolated scope, implemented only in open_ai_compatible file. It doesn't touch any other provider as far as I understand the code.
This opt-in simply gives those who want a native way to utilize Zed's already implemented reasoning UI + another opt-in feature to stop sending unnecessary tags which will be cutoff anyway by provider. Especially if it's something locally hosted.

RedBaron1914:
Great code review(or lack of thereof more like)

PR #54183, opened 2026-04-17
Move the worktree picker to the title bar + make it always visible

danilo-leal:
This PR makes Zed only have one worktree picker, as opposed to a flavor of it in the title bar and another in the agent panel. It then moves it to the title bar, making it always present, so that its trigger is separate from the branch picker (which now contains only two views: branches and stashes). For the worktree picker, I'm mostly favoring the behavior we've introduced in the agent-panel-flavored version.

It also updates the title bar settings migration to use the JSON `migrate_settings` helper instead of a shallow Tree-sitter rewrite, so old `show_branch_icon = true` values are promoted to `show_branch_status_icon = true` across root, platform, release-channel, and profile settings scopes.

- [x] Move worktree creation logic to the `git_ui` crate to make this more generic and less agent-specific
- [x] Double-check the remote use case and ensure nothing broke there
- [x] Improve the UX for the detached HEAD state; better invite people to create a branch
- [x] Migrate `show_branch_icon = true` to `show_branch_status_icon = true` across nested settings scopes

Suggested .rules additions

When migrating renamed settings keys that can appear in platform overrides, release-channel overrides, or profiles, prefer the JSON `migrations::migrate_settings` helper over shallow Tree-sitter key rewrites unless tests explicitly cover every nested scope that can contain the key.

Release Notes:

- Improved migration of the title bar branch status icon setting.


nathansobo:
A couple questions and observations from review:

1. In crates/title_bar/src/title_bar.rs, the branch button still computes a status-aware icon based on conflict / modified / added / deleted state, but the rendered button now always shows a muted GitBranch icon. Was that change in intent deliberate? I wanted to check whether losing the at-a-glance repo status indicator in the title bar was expected here.

2. In crates/git_ui/src/worktree_picker.rs, the create-worktree entries appear disabled in unsupported contexts, but confirm still dispatches CreateWorktree for the fixed create entries CreateFromCurrentBranch and CreateFromDefaultBranch. Then agent_panel rejects the action later for no-git or collab cases. From the user side, that seems like it can still look like a silent no-op when the picker dismisses without clear feedback. Also, the disabled tooltip text looks a bit misleading for collab projects since it always says Requires a Git repository in the project.

Wanted to flag both in case they were unintentional.

nathansobo:
One scope question from review: why are the bookmark changes part of this PR? I noticed the diff/history here pulls in the new bookmark store/editor/persistence work as well, but the rest of the PR reads as title bar / git picker / worktree UX. Are the bookmark pieces intentionally coupled to this change, or are they carryover from branch history that should be split out?

benbrandt (approved):

PR #54185, opened 2026-04-17
i18n: Localize remaining core UI strings

chan-yuu:
Self-Review Checklist:

Closes #ISSUE

Release Notes:


danilo-leal:
Heya, thank you for the PR but even though localization is certainly something of interest and ideal to do, we won't engage with it right now. Zed moves fast and things change quickly, and so having to update every single string in the app to different languages would significantly 1) slow us down and 2) create the possibility for massive out-of-sync content between certain languages. We don't have the bandwidth to support anything but English at the moment. 

Appreciate it again, though, as a non-native English speaker myself I know having it would be very helpful, but it's not the right moment for introducing this in the app.

chan-yuu:
Understand, thank you

PR #54213, opened 2026-04-17
Add create_thread and list_agents_and_models agent tools

eholk:
Adds two new built-in agent tools that let the agent spawn independent
sibling threads which appear in the agent sidebar and run in parallel with
the current conversation:

- **`create_thread`** — creates a new thread with a user-supplied title and
  initial prompt. The new thread runs on the user's behalf, persists like any
  other thread, and does **not** report results back to the parent (use
  `spawn_agent` for that). The parent thread's active view is never
  interrupted.
- **`list_agents_and_models`** — lets the agent enumerate configured agents
  and, for the native agent, the models that are available. Useful for
  choosing a cheaper model for bulk work (the motivating stress test:
  "spawn 100 threads that each write a haiku, using Haiku rather than Opus").

## Behavior

- The new thread shares the parent's project and worktree.
- It's inserted into `AgentPanel.retained_threads` just like a user-created
  thread, subject to the same `MaxIdleRetainedThreads` eviction and
  `ThreadMetadataStore` persistence. Running threads are never evicted;
  finished threads are rehydrated from disk on click.
- `create_thread` supports optional `agent` and `model` overrides. An agent
  override does **not** mutate the panel's `selected_agent` or the
  last-used-agent preference — those are snapshotted and restored around the
  call.
- A `use_new_worktree` field is on the schema but currently returns an error;
  the flag is there so follow-up work can wire it through to the existing
  `handle_worktree_requested` flow without a schema change.

## Architecture

- `agent::ThreadEnvironment` grows two new methods (`create_sibling_thread`,
  `list_available_agents`) with default impls that return an error, so
  existing implementors are unaffected.
- A new `agent::SiblingThreadHost` trait bridges the `agent` crate (which
  cannot depend on UI) to `agent_ui`. `NativeAgent` stores an optional host,
  `NativeThreadEnvironment` forwards calls to it.
- `AgentPanel` implements the host (`AgentPanelSiblingHost`) and installs it
  on the `NativeAgent` whenever a native-agent conversation view connects.
- `AgentPanel::create_thread_with_options` parameterizes the existing
  `create_thread` with an initial content block, agent, and model override.
  The plain `create_thread` becomes a thin wrapper that passes defaults, so
  no existing call sites change.

The tool goes through the same `create_agent_thread` → `ConversationView`
path as the user-facing "new thread" action, so eviction, persistence, and
sidebar rendering are identical to user-created threads.

## What's not in this PR

- `use_new_worktree: true` currently returns an error. The existing worktree
  flow (`handle_worktree_requested` / `open_worktree_workspace_and_start_thread`)
  already creates a second `Workspace` inside the current `MultiWorkspace`
  via `find_or_create_workspace`; a follow-up will thread a "background"
  placement parameter through that flow so it can skip
  `multi_workspace.activate` and install the new thread without taking focus.
- No tool for operating on an already-created sibling thread by ID. The
  initial success payload deliberately does not include a session ID because
  the session is populated asynchronously and we have no downstream consumer
  for it today.

## Testing

- `cargo check --lib` and `cargo clippy --no-deps` are clean on `agent`,
  `agent_ui`, and `settings_ui`.
- Manually exercised in a dev build: the agent can create a sibling thread
  with a title and prompt, the user's active thread stays focused, the new
  thread appears in the sidebar and begins generating.
- `list_agents_and_models` correctly enumerates the native agent with its
  default model flagged, plus any configured external agents.

## Suggested .rules additions

None required — a `.rules` file for the `agent` crate was added alongside
this change documenting the non-obvious fact that a new built-in tool must
be listed in `assets/settings/default.json`'s built-in profiles and either
added to `TOOLS` or `EXCLUDED_TOOLS` in
`settings_ui/src/pages/tool_permissions_setup.rs`. I hit both of those gates
during this work.

Release Notes:

- N/A

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1079,10 +1079,12 @@
         "tools": {
           "copy_path": true,
           "create_directory": true,
+          "create_thread": true,
           "delete_path": true,
           "diagnostics": true,
           "edit_file": true,
           "fetch": true,
+          "list_agents_and_models": true,
           "list_directory": true,
           "project_notifications": false,
           "move_path": true,
@@ -1105,8 +1107,10 @@
         // We don't know which of the context server tools are safe for the "Ask" profile, so we don't enable them by default.
         // "enable_all_context_servers": true,
         "tools": {
+          "create_thread": true,
           "diagnostics": true,
           "fetch": true,
+          "list_agents_and_models": true,
           "list_directory": true,
           "project_notifications": false,
           "now": true,
--- a/crates/agent/.rules
+++ b/crates/agent/.rules
@@ -1,0 +1,25 @@
+# Registering a new built-in agent tool
+
+Registering a tool in `Thread::add_default_tools()` and the `tools!` macro in
+`crates/agent/src/tools.rs` is **not enough** for the model to actually receive
+it. There are two additional gates that must be updated or the tool will be
+silently filtered out:
+
+1. **Built-in profiles in `assets/settings/default.json`.** The `write` and
+   `ask` profiles under `agent.profiles` each contain an explicit `tools` map
+   that acts as a per-profile allowlist. `Thread::enabled_tools()` calls
+   `profile.is_tool_enabled(name)` and excludes any tool not present with value
+   `true`. Add the new tool name to both profiles (or whichever profiles it
+   should be available in).
+
+2. **`test_all_tools_are_in_tool_info_or_excluded` in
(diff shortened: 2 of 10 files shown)


SomeoneToIgnore:

image

eholk:
Reopened as #57987 — GitHub wouldn't let me reopen this one after a force-push rewrote the branch.

PR #54226, opened 2026-04-18
Add note about login shell sourcing to environment docs

technocidal:
## Summary

- Adds a note to the "Launched via window manager, Dock, or launcher" section of the environment docs explaining that login shells do not source `.zshrc` or `.bashrc`, which is a common source of confusion when environment variables like `OPENAI_API_KEY` are missing.
- Lists solutions: moving variables to `.zshenv`/`.zprofile`, configuring keys in the Settings Editor, or launching Zed from the terminal.

Release Notes:

- N/A


technocidal:
@cla-bot check

PR #54233, opened 2026-04-18
git_graph: fix the misalignment between the canvas and table

lingyaochu:
Self-Review Checklist:

Closes #53469

The git graph panel consists of two separate components: the graph canvas and the commit info table. While they are rendered using different methods, their vertical layout both depend on the `row_height` property.

https://github.com/zed-industries/zed/blob/ec9be5c332f79986b046a2985e003766591ec82f/crates/git_graph/src/git_graph.rs#L930-L934

The discrepancy arises because the canvas rendering adopts the floating-point `row_height` directly, whereas the commit info table is implemented via GPUI's `uniform_list`. The underlying layout engine for the table intrinsically snaps elements to physical device pixels. When `display_scale * buffer_font_size` results in a fractional logical pixel value, the table's snapped row height slightly differs from the raw float used by the canvas. Over many rows, this compounding subpixel difference creates a noticeable visual mismatch.

This change fixes the issue by explicitly snapping the computed `row_height` property to the nearest physical pixel boundary beforehand. Because both the canvas math and the table's layout engine now operate on the exact same rounded device-pixel height, they remain perfectly aligned regardless of fractional font scaling.

Release Notes:

- Fixed an issue where the git graph lines would become vertically misaligned with the commit list alongside it when `buffer_font_size` set to non-integer.

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -927,10 +927,13 @@ impl GitGraph {
         cx.notify();
     }
 
-    fn row_height(cx: &App) -> Pixels {
+    fn row_height(window: &Window, cx: &App) -> Pixels {
         let settings = ThemeSettings::get_global(cx);
         let font_size = settings.buffer_font_size(cx);
-        font_size + px(12.0)
+        let raw = font_size + px(12.0);
+        let scale = window.scale_factor();
+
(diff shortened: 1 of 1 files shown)


lingyaochu:
Changes has been contained in #54429, closing this.

PR #54242, opened 2026-04-18
Import largest VS Code settings file instead of last found

### LINKED ISSUE #54241: Import VS Code Settings Doesn't Detect Correct File
### Reproduction steps

1. Install and open a VS Code variant
2. Install and open another VS Code variant
3. Open Zed
4. Go to Onboarding
5. Choose VS Code button under Import Settings

Current behavior:

Image

I tried Code OSS briefly before switching back to VS Code, but its the Code OSS settings that would be imported.

VS Code variant is chosen based on reverse order from `paths::vscode_user_data_paths()`, ranking insiders and dev builds above release builds and OSS and Codium above the official MS version.

Expected behavior:

Several behaviors would be superior:
1. Use the unreversed order so that release builds are prioritized rather than dev/insiders
2. Offer a choice from any detected build
3. Rank variants off some other criterion, like most recently used (mtime?) or size of preferences file.

Zed: v0.232.2+stable.229.34f1025a77a17ae5401c9395b4f73536c6f76d6c (Zed) 
OS: Linux Wayland manjaro unknown
Memory: 30.8 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics (ADL GT2) || Intel open-source Mesa driver || Mesa 26.0.2-arch1.1

### Relevant Keymap

_No response_

_No response_

None

pydsigner:
Zed onboarding offers to import settings from VS Code or Cursor. But there are a plethora of VS Code variants available, and settings from several may be present. Currently, they are checked sequentially but prioritized in reverse order. This patch adjusts that behavior to prioritize the largest settings file found, which is more likely to be associated with the user's primary editor.

Self-Review Checklist:

Closes #54241

Release Notes:

- Improved settings migration detection when multiple settings files exist


pydsigner:
@cla-bot check

pydsigner:
Updated behavior (compare with screenshot in #54241)
Screenshot_20260418_130301

maxdeviant:
Thank you for the PR, but I don't think simply using the largest settings file is the behavior we want here.

I think a better way to handle the ambiguity would be to have the UI present the user with a list of settings files that were found and allow them to select which one they want to import.

Going to close this PR, but we would likely accept one that goes in that direction.

PR #54247, opened 2026-04-19
vim: Add noremap support to SendKeystrokes

tsukinoko-kun:
This feature is for people like me, coming from Neovim.

In Neovim/Vim, there is an option for keymaps being recursive or not.  
This is called noremap in Vim and was required for bringing my keymaps I am used to on Neovim to Zed.
Since I am not alone with this problem, I implemented it.

In my case, I wanted to change the clipboard behavior. `d` for deleting without writing to clipboard. `x` for cutting to clipboard. etc. This requires a non-recursive keymap. Zed's keymaps being recursive, makes this impossible.

For your information, I am inexperienced in Rust.

What changed:
- SendKeystrokes changed from a plain string action to an enum that accepts either:
  - a string, preserving old behavior
  - an object with keystrokes and noremap
- dispatch logic now respects noremap per queued synthetic keystroke
- tests and docs were added for the new behavior

Why:
- allow Vim-style non-recursive remaps
- keep existing keymaps working unchanged
- avoid queued keystrokes accidentally inheriting wrong remap mode

### DIFF
--- a/crates/gpui/src/key_dispatch.rs
+++ b/crates/gpui/src/key_dispatch.rs
@@ -449,16 +449,20 @@ impl DispatchTree {
         &self,
         input: &[Keystroke],
         dispatch_path: &SmallVec<[DispatchNodeId; 32]>,
+        skip_send_keystrokes_remap: bool,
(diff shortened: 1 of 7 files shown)


tsukinoko-kun:
@cla-bot check

PR #54248, opened 2026-04-19
markdown: Fix rendering error of the selection box when text is fully selected by mouse events in Markdown

Classpi:
In Markdown, four or more consecutive left-clicks cause Zed to select all rendered text. Yet, despite the text being fully selected (copyable via Command+C), the visual selection highlight fails to render highlight.

图片

As illustrated above, when all rendered text in the floating Markdown window is selected, Command+C copies it successfully, yet the selection highlight fails to render. The fixed version is shown below.         

图片

Self-Review Checklist:

- [ x ] I've reviewed my own diff for quality, security, and reliability
- [ x ] Unsafe blocks (if any) have justifying comments
- [ x ] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [ x ] Tests cover the new/changed behavior
- [ x ] Performance impact has been considered and is acceptable

Release Notes:

- Fixed markdown all selection highlight render error

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -2848,15 +2848,17 @@ impl RenderedText {
         for line in self.lines.iter() {
             let line_source_start = line.source_mappings.first().unwrap().source_index;
             if source_index < line_source_start {
-                break;
-            } else if source_index > line.source_end {
-                continue;
-            } else {
                 let line_height = line.layout.line_height();
-                let rendered_index_within_line = line.rendered_index_for_source_index(source_index);
-                let position = line.layout.position_for_index(rendered_index_within_line)?;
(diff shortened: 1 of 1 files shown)


smitbarmase:
Hey, I may be missing something, but I’m not sure we’re seeing the same behavior. From what I can tell, after four clicks or when selecting all - the current behavior on `main` seems correct (which is also how the editor shows selection) and looks better than the change in this PR.

Current main:
image

This PR:
image

I’m going to close this for now since I don’t think this change is needed. Please feel free to comment if I’ve misunderstood anything, or open a new PR if there’s another case I’m not accounting for.

PR #54254, opened 2026-04-19
Update hover tracking logic to use frame-by-frame distance comparison…

daydalek:
This is a follow-up refinement to the direction-aware hover logic introduced in #47471 based on users' feedback (and my daily drive experience). Previously in #47471 , `closest_mouse_distance` tracked the absolute minimum distance the cursor ever reached. This meant that any slight overshoot or horizontal movement (like moving through trailing whitespace) would unintentionally trigger a dismissal countdown. This seems to be too strict, users feel it unintuitive and the popover is too easy to be dismissed.

By shifting `closest_mouse_distance` to `last_mouse_distance` and comparing the current distance against the previous frame (with a 2px tolerance instead of the original 4px, since we are now tracking per frame.), the popover now reliably persists as long as the cursor is generally moving towards it or moving horizontally within its bounds. This improves the stability of hovers and makes them much easier to interact with.

Release Notes:

- Improved hover popover stability when moving the cursor towards it or moving horizontally.

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -60,7 +60,7 @@ pub fn hover_at(
 
         if let Some(anchor) = anchor {
             editor.hover_state.hiding_delay_task = None;
-            editor.hover_state.closest_mouse_distance = None;
+            editor.hover_state.last_mouse_distance = None;
             show_hover(editor, anchor, false, window, cx);
         } else {
             let settings = EditorSettings::get_global(cx);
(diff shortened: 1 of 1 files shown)


daydalek:
Seems I missed the point. originally I thought users feel the popover easy to disappear because the logic is too strict ,but it seems to be a language-specific issue mentioned in https://github.com/zed-industries/zed/issues/46841#issuecomment-4283571008, It's related to godot lsp and a fallback in `hover_at` when symbol _range is set to zero(seems `range` is not provided by the godot lsp), so the hiding delay is not applied at all. gonna make another pr once I implement the fix.

PR #54264, opened 2026-04-19
ci: Verify commits and releases created with Zed Zippy

MrSubidubi:
This changes our approach for creating Zed version bumps with the Zed Zippy identity: 

With these changes, the Zippy version bumps as well as the tag creations will be verified as per https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#signature-verification-for-bots. This adds an extra layer of security for version bumps and release triggers done by the Zippy identity.

Eventually, we can move code more around for this so it happens in the bot itself, but for now, I think its a good and solid interim solution to have this here.

Release Notes:

- N/A


ConradIrwin (approved):
Neat!

PR #54270, opened 2026-04-19
dev_container: Resolve compose service build args in Dockerfile expansion

fechu:
For docker-compose-based dev containers, build args live on the primary compose service's `build.args`, not on `dev_container.build` (which is `None` in the compose case). `expanded_dockerfile_content` only consulted the latter, so `${…}` references in the service's Dockerfile — e.g. `FROM ${BASE_IMAGE}` — were never substituted. Downstream callers then invoked `docker inspect "${BASE_IMAGE}"` and `docker pull "${BASE_IMAGE}"`, both of which fail and the dev container fails to start.

A new unit test `test_expands_compose_service_args_in_dockerfile` mounts a Dockerfile with `FROM ${BASE_IMAGE}` backed by a compose service whose `build.args` define `BASE_IMAGE=test_image:latest`, and asserts both `expanded_dockerfile_content` and `image_from_dockerfile` produce the resolved reference.

Self-Review Checklist:

Release Notes:

- Fixed dev container startup failing for docker-compose configs whose service Dockerfile uses build-arg substitution in the FROM line (for example, `FROM ${BASE_IMAGE}`).


KyleBarton (approved):
Thanks for this!

PR #54271, opened 2026-04-19
dev_container: Preserve post-FROM directives in generated stage alias

fechu:
## Self-Review Checklist:

## Release Notes:

- Fixed dev container startup for Dockerfiles that perform RUN/COPY after the FROM line 

## Details

#53538 rewrote `dockerfile_inject_alias` to append a fresh `FROM <base> AS dev_container_auto_added_stage_label` at the end of the user's Dockerfile instead of aliasing the existing stage in place. For a single-FROM Dockerfile that does user setup after `FROM` (e.g. pre-creating `/home/<user>/.local/...` with correct ownership, fixing up UID/GID), that setup was silently dropped from the aliased stage. The downstream `dev_containers_target_stage` inherits from `dev_container_auto_added_stage_label`, which was now a pristine copy of the base image, so the final container was missing every RUN/COPY the user wrote after `FROM`.

In practice this caused the Zed remote server (running as the non-root user inside the container) to get `Permission denied` when trying to write `/home/<user>/.local/share/zed/server_state/dev-workspace-N`, because `/home/<user>/.local/share` had the wrong ownership.

## Fix

Rewrite `dockerfile_inject_alias` to:

- Locate the targeted `FROM` line (matching `build_target` if set, otherwise the last `FROM`).
- If that line has no `AS <name>` clause, attach `AS dev_container_auto_added_stage_label` to it in place.
- If the targeted line already has an alias (multi-stage builds with an explicit `target`), append `FROM <existing-alias> AS dev_container_auto_added_stage_label` at the end so the new stage inherits from the named one.

Two new focused unit tests:
- `test_dockerfile_inject_alias_preserves_post_from_directives` — single-FROM Dockerfile with a RUN after it; asserts the alias attaches to the original FROM in place.
- `test_dockerfile_inject_alias_with_existing_stage_alias` — multi-stage Dockerfile where the targeted FROM has `AS development`; asserts we append `FROM development AS dev_container_auto_added_stage_label`.

Existing feature-dockerfile golden tests were updated to reflect the restored in-place alias format.

## Minimal reproduction

```
repro/
└── .devcontainer/
    ├── devcontainer.json
    ├── docker-compose.yml
    └── Dockerfile
```

`.devcontainer/devcontainer.json`:
```json
{
  "name": "repro",
  "dockerComposeFile": "docker-compose.yml",
  "service": "app",
  "workspaceFolder": "/workspace"
}
```

`.devcontainer/docker-compose.yml`:
```yaml
services:
  app:
    build:
      context: .
      dockerfile: Dockerfile
    user: ubuntu
    command: [sleep, infinity]
```

`.devcontainer/Dockerfile`:
```dockerfile
FROM ubuntu:24.04
RUN useradd -m -u 1000 ubuntu \\
 && install -o ubuntu -g ubuntu -d /home/ubuntu/.local /home/ubuntu/.local/share
USER ubuntu
```

On `main`, Zed's remote server fails to start with:
```
(remote) Failed to init server paths: Failed to create server_dir \`/home/ubuntu/.local/share/zed/server_state/dev-workspace-N\`: Permission denied (os error 13)
```
because the `install -o ubuntu ...` step never runs in the aliased stage. With this PR the alias is attached to the existing `FROM` line, the RUN becomes part of the stage, and the remote server starts normally.

### DIFF
--- a/crates/dev_container/src/devcontainer_manifest.rs
+++ b/crates/dev_container/src/devcontainer_manifest.rs
@@ -2376,39 +2376,102 @@ fn dockerfile_inject_alias(
     alias: &str,
     build_target: Option<String>,
 ) -> String {
-    match image_from_dockerfile(dockerfile_content.to_string(), &build_target) {
-        Some(target) => format!(
-            r#"{dockerfile_content}
-FROM {target} AS {alias}"#
-        ),
-        None => dockerfile_content.to_string(),
+    let from_lines = from_lines_from_dockerfile(dockerfile_content);
+    let mut lines: Vec<&str> = dockerfile_content.lines().collect();
+
+    if let Some(build_target) = build_target {
+        format!("{dockerfile_content}\nFROM {build_target} AS {alias}")
+    } else {
+        let Some(last_from_line) = from_lines.last() else {
+            return dockerfile_content.to_string();
+        };
+        if let Some(this_alias) = last_from_line.line.alias.as_deref() {
+            format!("{dockerfile_content}\nFROM {this_alias} AS {alias}")
+        } else {
+            let image = &last_from_line.line.image;
+            let line = format!("FROM {image} AS {alias}");
+            lines[last_from_line.line_number] = line.as_str();
+            lines.join("\n")
+        }
     }
 }
 
-fn image_from_dockerfile(dockerfile_contents: String, target: &Option<String>) -> Option<String> {
+// Represents a FROM line in a Dockerfile
+struct FromLine {
+    image: String,
+    alias: Option<String>,
+}
+
+struct IndexedFromLine {
+    line_number: usize,
+    line: FromLine,
+}
+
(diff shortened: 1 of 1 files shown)


KyleBarton:
@fechu thanks for the contribution - I actually think #54210 (just merged) solves this discrepancy already, so I believe this is duplicate work. Let me know if you see any reason to continue here; otherwise I think this can be closed.

fechu:
> @fechu thanks for the contribution - I actually think #54210 (just merged) solves this discrepancy already, so I believe this is duplicate work. Let me know if you see any reason to continue here; otherwise I think this can be closed.

Even better! I didn't see that one. I'll give it a try with the dockerfiles I have at hand and report back.

KyleBarton:
Going to close for now as I believe this is duplicated. Let's revisit if needed

PR #54273, opened 2026-04-20
Highlight CSS font shorthand line-height values

### LINKED ISSUE #54235: CSS syntax highlighting is incorrect for `font` properties
### Reproduction steps

The `font` shorthand accepts a `line-height` value right after the `font-size` (separated by a `/`), but the height is not being highlighted properly. 

The active theme needs to be one that uses a different color for numerals in order to spot the issue. Try: One Dark or Gruvbox Dark Hard.

To reproduce, add this to a CSS file:
```css
.foo{
  font: 700 16px/32px serif;
}
```

### Current behavior: 
Image

Image

### Expected behavior:
The digits in `32px` after the `/` should be highlighted as a number rather than being formatted as a generic value.

Zed: 0.230.2 
OS: macOS 15.7.4
Memory: 64 GB
Architecture: aarch64

### Attach Zed log file

N/A

### Relevant Zed settings

N/A

### Relevant Keymap

N/A

N/A

None

hamza-paracha:
Fixes #54235

Summary:
- Highlights slash-prefixed numeric CSS values like /32px as numbers.
- Keeps the query scoped to numeric slash values used by font shorthand line-height syntax.

Release Notes:

- Fixed CSS highlighting for line-height values in font shorthand properties.

### DIFF
--- a/crates/grammars/src/css/highlights.scm
+++ b/crates/grammars/src/css/highlights.scm
@@ -96,6 +96,9 @@
   (float_value)
 ] @number
 
+((plain_value) @number
+  (#match? @number "^/[0-9]+(\\.[0-9]+)?[a-zA-Z%]*$"))
+
 (unit) @type.unit
 
 [


yeskunall:
Thanks @DevDonzo -- heads-up: a reviewer will likely request some screenshots of your changes (before / after).

DevDonzo:
Thanks for the heads-up. I will add real before/after screenshots once I can capture the rendered highlight in Zed; I do not want to use a synthetic mock image for this.

smitbarmase:
Thanks, but I don’t think this fixes it. In the grammar, `16px/32px` gets lexed as `16` plus a single `plain_value` of `px/32px`, so a node starting with `/` never exists.

Moreover, the root cause is tokenization in `tree-sitter-css`, so this needs an upstream fix first. Closing this one.

PR #54274, opened 2026-04-20
Show added line counts for untracked files

### LINKED ISSUE #54266: Show "+N lines" for new files
For new files, in the changes dialog, we should add "+N" lines for new files, instead of leaving blank 

Image

hamza-paracha:
Fixes #54266

Summary:
- Includes untracked files when building git diff stats.
- Counts untracked file lines so new files can show +N in git UI instead of a blank stat.

Release Notes:

- Fixed new files missing added-line counts in git change lists.

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -43,6 +43,20 @@ pub use askpass::{AskPassDelegate, AskPassResult, AskPassSession};
 
 pub const REMOTE_CANCELLED_BY_USER: &str = "Operation cancelled by user";
 
+fn line_count_for_diff_stat(contents: &[u8]) -> u32 {
+    if contents.is_empty() {
+        return 0;
+    }
+
+    let trailing_line = if contents.last() == Some(&b'\n') {
+        0
+    } else {
+        1
+    };
+    let line_count = contents.iter().filter(|byte| **byte == b'\n').count() + trailing_line;
+    u32::try_from(line_count).unwrap_or(u32::MAX)
+}
+
 /// Format string used in graph log to get initial data for the git graph
 /// %H - Full commit hash
 /// %P - Parent hashes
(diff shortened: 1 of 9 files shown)


MaxStaszkiewicz:
Would running `git add -N` for newly created files? That would make them show up in `git diff` without actually staging the files 🤔

crates/git/src/repository.rs
```diff
@@ -3744,6 +3811,15 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_line_count_for_diff_stat() {
```
yara-blue: this unit test is not needed as the function it tests is a trivial implementation detail (so it just adds noise).

Please read through the other tests, then make a similar one or modify a suitable existing test to see if the result from repo.diff_stat() is correct.

crates/git/src/repository.rs
```diff
@@ -2197,7 +2211,60 @@ impl GitRepository for RealGitRepository {
...
+                        }
+                        Err(err) => {
+                            log::debug!("failed to read untracked file {full_path:?}: {err}");
+                        }
+                    }
+                }
+
+                entries.sort_by(|(a, _), (b, _)| a.cmp(b));
+                entries.dedup_by(|(a, _), (b, _)| a == b);
+                diff_stat.entries = entries.into();
```
yara-blue, line 2306: function is getting quite long, please extract some of this, like with `crate::status::parse_numstat`.

yara-blue (dismissed):
Thanks for figuring out what needs to happen here. Its not there yet it needs a more "integration style" test and a little code cleanup, otherwise good to go! Please ping me (@yara-blue ) when it's ready.

DevDonzo:
@yara-blue Updated this. I removed the narrow line-count unit test, extracted the untracked-file diff stat collection out of diff_stat, and added a RealGitRepository::diff_stat test that verifies an untracked file reports its added line count. Thanks for the pointers.

DevDonzo:
@yara-blue I pushed another update after the bot review. This keeps the integration-style diff_stat test and extracted helper from your review, and also hardens the helper so untracked stats stream regular files, handle symlinks as link-target text, and skip special files instead of reading them.

yara-blue:
hi @DevDonzo thanks for the work, it turns out this gets rather nuanced so I'll be taking it from here. Don't worry will get it merged! :)

yara-blue:
Okay so update:
I still need to handle: 
- binary files, 
- symlinks escaping the repo
- edits modifying the line count

Worries:
This (counting untracked file lines) is called concurrently with the git status call, if this takes long it will block git actions. Adding a huge untracked dir by accident is a relatively common occurrence. To resolve this delay the line number counting must be decoupled from the status call etc. That is complex since the results could be invalidated by edits both internal and external to zed.

That latter concern is def out of scope for this since resolving it would require rewriting the git jobsystem @Anthony-Eid has some great ideas for that. We need to discuss with the team if the effort of doing so is worth it right now, I suspect our time is better spend elsewhere.

As a crutch we can limit the counting time to some maximum time, maybe 2 seconds? That would allow merging this without a jobsystem refactor.

hamza-paracha:
@yara-blue I pushed a follow-up for the latest concerns. Untracked diff-stat collection is now bounded to a 2 second budget, binary files are skipped, symlinks are counted from the link target text without following them, and files that change while being read are skipped. I also kept the integration-style diff_stat test and added a binary-file case.

Verified locally:
- cargo fmt --check
- cargo test -p git test_diff_stat_includes_untracked_files
- ./script/clippy -p git

hamza-paracha:
@yara-blue I rebased this on current upstream `main` to clear the merge conflict and force-pushed the branch.

Re-verified after rebase:
- cargo fmt --check
- cargo test -p git test_diff_stat_includes_untracked_files
- ./script/clippy -p git

cameron1024 (approved):

yara-blue:
@DevDonzo Do you have a windows machine you have access too? We could use some help to get the tests to pass. The windows fakefs diff_stat implementation was wrong but I do not think it's good yet.

hamza-paracha:
I don’t have access to a Windows machine, unfortunately. I’m happy to keep helping from CI logs if that’s useful, but I won’t be able to run the Windows fakefs/diff_stat tests locally.

yara-blue:
> I don’t have access to a Windows machine, unfortunately. I’m happy to keep helping from CI logs if that’s useful, but I won’t be able to run the Windows fakefs/diff_stat tests locally.

That would be great! Also feel free to ask any other contributor for help on the discord!

SomeoneToIgnore:
To keep the PR queue cleaner, will close this one for now, as it had accumulated quite a number of conflicts with no activity

image

PR #54275, opened 2026-04-20
Use buffer line numbers when copying file locations

### LINKED ISSUE #54250: Copy file location in multi-buffer copies wrong line
### Reproduction steps

1. Do a global search to find something
2. Click on a line in the results
3. Run `editor: copy file location`

### Current behavior

The line number copied is the line number relative to the start of the multi-buffer. Say that the first result is in a file call `foo` on line 27. You'd expect it to copy `foo:27` to your clipboard. It copies `foo:3` instead.

### Expected behavior

The line number should reflect which line of the _file_ it's on, not which line of the multi-buffer. In this case, I expect `foo:27`.

Zed: v0.232.2+stable.229.34f1025a77a17ae5401c9395b4f73536c6f76d6c (Zed) 
OS: macOS 15.7.4
Memory: 64 GiB
Architecture: aarch64

hamza-paracha:
Fixes #54250

Summary:
- Converts copied file-location selections through the underlying buffer range before formatting line numbers.
- Adds regression coverage for copying a location from a multibuffer excerpt.

Release Notes:

- Fixed copy file location from multibuffer excerpts reporting multibuffer-relative lines.

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -23918,38 +23918,46 @@ impl Editor {
 
         let multi_buffer = self.buffer().read(cx);
         let multi_buffer_snapshot = multi_buffer.snapshot(cx);
-        let buffer_ranges = multi_buffer_snapshot
-            .range_to_buffer_ranges(selection_range.start..selection_range.end);
-        let selected_buffer_range = if selection.reversed {
-            buffer_ranges.first()
-        } else {
-            buffer_ranges.last()
+        let format_location = |buffer_snapshot, buffer_range: Range<Point>| {
+            let start_line = buffer_range.start.row + 1;
+            let end_line = buffer_range.end.row + 1;
+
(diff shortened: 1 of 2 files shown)


DevDonzo:
Pushed a follow-up for the CI failure. The copy action now handles collapsed cursor selections with point_to_buffer_point, and non-empty selections still use range_to_buffer_ranges for buffer-relative line numbers.

ConradIrwin:
@DevDonzo thanks for this! Looks like there are some unexpected changes in the diff around the tests/empty modal view. Are you able to update the PR to just have the relevant change?

vgpopov:
@ConradIrwin Could you merge the PR please. 
It's a real annoying bug that breaks review flow for the terminal threads. I always have to fix line number manually...

SomeoneToIgnore:
Will close as stale as no requested fixes were made and now first conflicts appear:

image

PR #54281, opened 2026-04-20
Add starts_open setting to terminal panel

jknlsn:
## Summary

- Add `terminal.starts_open` setting so the terminal panel can open automatically in new workspaces, like project and git panels already can
- Expose the setting through terminal settings docs, default settings, Settings Editor metadata, and the terminal panel implementation
- I also added a matching settings page item for the existing `git_panel.start_open` setting

Self-Review Checklist:

- [ ] Tests cover the new/changed behaviour
  - I decided not to add tests as the behaviour seems covered by the existing settings tests, and similar areas don't seem to have their own explicit tests.

Related to #51542 (issue mentions possibly adding for all panels, closed with implementation for `git_panel`)

Release Notes:

- Added `terminal.starts_open` to control whether the terminal panel opens automatically in new workspaces


jknlsn:
@cla-bot check

Issue #53922, opened 2026-04-14
Dev Containers don't support podman-compose


KyleBarton:
### Reproduction steps

1. Download podman
2. Ensure podman's compose provider is `podman-compose` (most podman distros will use docker-compose as the provider if available)
3. Start Zed, open a project
4. Ensure `use_podman: true` is set in your settings.json file
5. Cmd-Shift-P to for `projects: initialize dev container`
6. Create a dev container with a docker-compose variant - `Rust & PostgreSQL` will work
7. Attempt to open dev container

Error is described in https://github.com/zed-industries/zed/issues/53478#issuecomment-4241214247 - in short, podman-compose has no `--format=json` option, which means that the Dev Container module cannot deserialize the compose files.

The solution here is to take a dependency on a YAML deserialization provider - presumably [rust-yaml2](https://github.com/Ethiraric/yaml-rust2) since it is a transitive dependency already - and move away from json deserialization for docker/podman-compose cases.

### Current vs. Expected behavior

Current behavior: from the issue linked above, an error of the form:
```
2026-04-13T23:16:29-04:00 ERROR [dev_container::command_json] Error running command Command("podman" "compose" "-f" "/home/pp/Projects/test-repo/.devcontainer/docker-compose.yml" "-f" "/home/pp/Projects/test-repo/.devcontainer/docker-compose.runtime.yml" "config" "--format" "json"): Sent non-successful output; cannot deserialize. StdErr: �[4m>>>> Executing external compose provider "podman-compose". Please see podman-compose(1) for how to disable this message.
```

Expected behavior: Dev Container opens without issue

### Zed version and system specs

Zed: v0.231.2+stable.221.cc335b70f85a17974a4c61f852dbebff8c4b1db7 (Zed) 
OS: macOS 15.7.4
Memory: 18 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

KyleBarton:
FYI, this was fixed as part of #55117  -- actually merged by way of #55225. Closing manually

PR #52338, opened 2026-03-24
Dev containers native implementation

KyleBarton:
## Context

Closes #11473

In-house Zed implementation of devcontainers. Replaces the dependency on the [reference implementation](https://github.com/devcontainers/cli) via Node.

This enables additional features with this implementation:
1. Zed extensions can be specified in the `customizations` block, via this syntax in `devcontainer.json:
```
...
  "customizations": {
    "zed": {
      "extensions": ["vue", "ruby"],
    },
  },

```
2. [forwardPorts](https://containers.dev/implementors/json_reference/#general-properties) are supported for multiple ports proxied to the host

## Self-Review Checklist

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Release Notes:

- Improved devcontainer implementation by moving initialization and creation in-house

PR #54302, opened 2026-04-20
dev_container: Align compose project name with reference CLI

### LINKED ISSUE #54255: Dev container: compose project name derived from `name` field diverges from devcontainer CLI / VS Code
#### Summary

When a `devcontainer.json` uses `dockerComposeFile`, Zed derives the Docker Compose project name from the `devcontainer.json` `name` field (safe-id-lowered). The [`@devcontainers/cli`](https://github.com/devcontainers/cli) reference implementation (and by extension VS Code) instead follows a five-step precedence in [`getProjectName` in `src/spec-node/dockerCompose.ts`](https://github.com/devcontainers/cli/blob/main/src/spec-node/dockerCompose.ts):

1. `COMPOSE_PROJECT_NAME` from the environment.
2. `COMPOSE_PROJECT_NAME=` in the workspace `.env` file.
3. Top-level `name:` on the merged compose config, when at least one fragment declared it explicitly.
4. `${workspaceFolderBasename}_devcontainer`, only when the first compose file's directory is `<workspace>/.devcontainer/`.
5. Otherwise, the plain basename of the first compose file's directory (no suffix).

Zed's old derivation doesn't match any of these inputs, so the two tools operate on *different* Docker Compose projects for the same folder and same `devcontainer.json`, creating parallel container sets that both claim the same Dev Containers spec identifying labels (`devcontainer.local_folder`, `devcontainer.config_file`). The identifying labels are supposed to uniquely identify a project's container across tools; this divergence breaks that, and [zed-industries/zed#54068](https://github.com/zed-industries/zed/pull/54068) surfaces the resulting ambiguity as a hard error on lookup.

**When this matters.** The duplicate appears when new Zed (v0.232+, name-field derivation) creates a container for a folder and a CLI-derivation tool — `@devcontainers/cli`, VS Code, or old Zed ≤ v0.231.x that shelled out to the CLI — then runs against the same folder: the second tool doesn't check the spec's identifying labels before creating its own project, so a parallel container set appears. The reverse order (CLI-derivation tool first, new Zed second) does not duplicate — Zed's `check_for_existing_container` is label-first and reuses the existing container — so a v0.231.x → v0.232+ upgrade by itself is not a reproducer; a mixed-version rhythm where the name-field container was created first is. Users who only ever open a folder with a single Zed version still hit the downstream consequences: the project namespace is tied to user-editable display text, and external tooling that assumes CLI-style naming no-ops against Zed-created containers (failure mode §2 below).

#### Reproduction

Fixture: minimal compose-based devcontainer with a `name` field, exercising rule 4 of the CLI precedence (compose file under `.devcontainer/`, no `COMPOSE_PROJECT_NAME`, no `.env`, no explicit top-level `name:`). Clone into a folder named `devcontainer-compose-test`, or follow the inline files below.

Fixture repo with copy-pasteable runbook and captured outputs: [antont/zed-devcontainer-compose-test](https://github.com/antont/zed-devcontainer-compose-test) ([`VERIFY.md`](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md)).

`devcontainer-compose-test/.devcontainer/devcontainer.json`:
```json
{
  "name": "Compose Duplicate Repro",
  "dockerComposeFile": "docker-compose.yml",
  "service": "app",
  "workspaceFolder": "/workspace"
}
```

`devcontainer-compose-test/.devcontainer/docker-compose.yml`:
```yaml
services:
  app:
    image: alpine:latest
    command: sleep infinity
    volumes:
      - ..:/workspace:cached
```

Steps (see [`VERIFY.md`](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md) for captured outputs from a real run):

1. [Clean slate](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md#1-clean-slate):
   ```sh
   docker ps -a --filter "label=devcontainer.local_folder=$PWD" -q | xargs -r docker rm -f
   ```
2. [Open the folder in Zed with `--dev-container`](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md#2-claim-zed-derives-compose-project-from-safe_id_lowerdevcontainerjsons-name). Zed creates compose project `compose_duplicate_repro`, container `compose_duplicate_repro-app-1`.
3. [Without quitting Zed, run `devcontainer up --workspace-folder .`](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md#3-claim-devcontainerscli-derives-project-from-folderbasename_devcontainer). The CLI creates compose project `devcontainer-compose-test_devcontainer` (rule 4), container `devcontainer-compose-test_devcontainer-app-1`.
4. [The dual-label `docker ps` query](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md#5-claim-spec-style-dual-label-query-returns-1-row) now returns two containers with different compose projects but identical `devcontainer.*` labels.

**Order matters**: Zed first, CLI second. [Reversed order](https://github.com/antont/zed-devcontainer-compose-test/blob/main/VERIFY.md#6-claim-order-matters--cli-first-then-zed-does-not-duplicate), Zed's `check_for_existing_container` reuses the CLI's container (since it correctly finds it by label) and the collision doesn't manifest.

The fixture only exercises rule 4, but the same divergence occurs whenever any CLI-precedence rule resolves differently from Zed's `name`-field derivation — users who set `COMPOSE_PROJECT_NAME`, ship a `.env` with one, declare a top-level compose `name:`, or point `dockerComposeFile` outside `.devcontainer/` (e.g. `"../docker-compose.yml"`) all land in the same duplicate-project state.

#### Root cause

`crates/dev_container/src/devcontainer_manifest.rs` — `project_name()`:
```rust
fn project_name(&self) -> String {
    if let Some(name) = &self.dev_container().name {
        safe_id_lower(name)
    } else {
        let alternate_name = &self
            .local_workspace_base_name()
            .unwrap_or(self.local_workspace_folder());
        safe_id_lower(alternate_name)
    }
}
```

The `name` field in `devcontainer.json` is meant as a **display name** for the dev container (per the [spec](https://containers.dev/implementors/json_reference/#general-properties)). Using it as the compose project namespace couples the namespace to user-editable display text and never consults any of the five inputs the reference CLI's [`getProjectName`](https://github.com/devcontainers/cli/blob/main/src/spec-node/dockerCompose.ts) walks. VS Code uses the same reference implementation.

#### Failure modes observed in practice

Three distinct failure modes, all from the same root cause, observed on a real project where multiple git worktrees share one `devcontainer.json` (`"name": "(project name redacted)"`):

1. **Silent db/volume sharing across worktrees.** Compose identifies services by `(project, service_name)` and uses the `com.docker.compose.config-hash` label to decide whether to recreate. When Zed opens worktree B under the same `name_redacted` project that was previously used for worktree A, Compose keeps the existing `name_redacted-db-1` and `name_redacted-localstack-1` containers (their config hashes match across worktrees — same image, same env, no per-worktree differences) and only replaces the `app` service (whose bind-mount path differs per worktree, so its hash changes). So worktree B silently inherits worktree A's database. This is invisible until two worktrees have diverging migrations — at which point whichever ran last wins, and earlier workers' data is clobbered.

   Standalone fixture with a two-worktree runbook proving this live (db container ID and sentinel file survive across the worktree switch): [antont/zed-devcontainer-db-share-repro](https://github.com/antont/zed-devcontainer-db-share-repro) ([`VERIFY.md`](https://github.com/antont/zed-devcontainer-db-share-repro/blob/main/VERIFY.md)).

2. **External tooling that assumes reference-CLI project naming silently no-ops.** E.g. a "close this worker" CLI we use runs `docker compose -p ${folder}_devcontainer down`. Against old-Zed-/CLI-created containers that works. Against new-Zed-created containers it targets a project that doesn't exist and exits cleanly — containers keep running. Any third-party orchestration (scripts, CI, test harnesses) that picks the CLI-style name is affected the same way.

3. **Mixed-version Zed sessions produce parallel container sets.** The Rust dev-container implementation landed in stable [v0.232.2](https://github.com/zed-industries/zed/pull/52338) on 2026-04-15. Older Zed (≤ v0.231.x) shelled out to `@devcontainers/cli` and so used the full reference precedence — matching whichever rule the CLI resolves. Opening a folder first with new Zed (name-field derivation) and then with old Zed or the CLI duplicates — the same shape the reproduction's §2→§3 sequence shows; the reverse order (old Zed / CLI first, new Zed second) does not, because new Zed's label-first lookup reuses the existing container (reproduction §6). So a v0.231.x → v0.232+ upgrade doesn't duplicate by itself; the collision requires a rhythm where a name-field container was created first and a CLI-derivation tool runs against the same folder afterwards — including transitional cases like a local dev-build rebuild that drops back to stable. How we first noticed this: `docker ps --filter label=devcontainer.local_folder=$pwd --filter label=devcontainer.config_file=$path` returned two JSON rows for one workspace (the exact ambiguity #54068 addresses on the lookup side).

These are not separate bugs — they're what "two tools don't agree on the project namespace" looks like in day-to-day use. The project-name fix below resolves all three.

#### Proposed fix

Replace `project_name()` with the full `getProjectName` precedence from `@devcontainers/cli` (see source linked above): `COMPOSE_PROJECT_NAME` in the environment → `COMPOSE_PROJECT_NAME=` in the workspace `.env` file → top-level `name:` on the merged compose config (only when a fragment explicitly declared it — `docker compose config` injects a default `name: devcontainer` that must be distinguished from a user-declared one) → `${workspaceFolderBasename}_devcontainer` when the first compose file's directory is `<workspace>/.devcontainer/` → otherwise the plain basename of that directory. All branches go through the CLI's final normalization step (lowercase + strip `[^-_a-z0-9]`, preserving hyphens).

Effects:
- `devcontainer.json`'s `name` is no longer used for the compose project namespace (it stays in use as the features image-tag prefix, per the spec's display-field role).
- For every input the CLI consults, Zed produces the same project name it does, so container sets are interchangeable between Zed, the devcontainer CLI, and VS Code for the same folder.
- Migration is handled via label-first lookup plus a canonical-project tiebreak (see below).

##### Migration / backwards compatibility

Existing Zed-created containers (under the old `safe_id_lower(name)` project) remain discoverable on upgrade because `check_for_existing_container` already queries by `devcontainer.local_folder` + `devcontainer.config_file` labels, not by project name. Zed reuses the existing container on first run after upgrade and no user action is required.

The only case the label-based lookup can't resolve on its own is when **both** a legacy-named Zed container and a CLI-style container already exist for the same folder — the label query returns two candidates. To keep the upgrade transparent in that case, `check_for_existing_container` prefers the container whose `com.docker.compose.project` label matches the value the new derivation produces, logging the ignored legacy orphan. Two non-canonical candidates still surface [#54068](https://github.com/zed-industries/zed/pull/54068)'s `MultipleMatchingContainers` error as the safety net for genuine ambiguity.

##### Out of scope / follow-ups

- Actual duplicate-state detection / error messaging is being handled in #54068.
- The new derivation normalizes per the CLI's rules (lowercase + strip `[^-_a-z0-9]`, preserving hyphens), which isn't what `safe_id_lower` does — so `safe_id_lower` drops out of the compose-project path under this proposal. It should stay in the features image-tag path, which continues to use the `name` field.

#### References

- Spec: https://containers.dev/implementors/spec/
- Reference CLI — `getProjectName`: https://github.com/devcontainers/cli/blob/main/src/spec-node/dockerCompose.ts
- Related Zed PR (duplicate detection): https://github.com/zed-industries/zed/pull/54068
- Implementing PR (draft in fork): https://github.com/antont/zed/pull/6
- Reviewer's point on identifying-labels uniqueness: https://github.com/zed-industries/zed/pull/54068#issuecomment-4261904446
- Fixture — compose-project-name divergence (VERIFY.md): https://github.com/antont/zed-devcontainer-compose-test
- Fixture — cross-worktree db reuse (VERIFY.md): https://github.com/antont/zed-devcontainer-db-share-repro

Current behavior: Zed's `name`-field derivation diverges from the reference CLI's five-step `getProjectName` precedence, so the two tools operate on different compose projects for identical `devcontainer.*` labels (full details in Reproduction steps above).

Expected behavior: Zed matches the same five-step `getProjectName` precedence, so containers created by Zed, the devcontainer CLI, and VS Code are interchangeable for the same folder regardless of which precedence rule applies. The `name` field stays a display field only, per the spec.

Zed: v0.234.0+dev.303ebca92112433e6bbc65abf60257cc97acc4a4 (Zed Dev 303ebca92112433e6bbc65abf60257cc97acc4a4) 
OS: macOS 26.3
Memory: 32 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

antont:
> **Draft / open question for maintainers.** The failure mode this fixes is narrow — a new-Zed-created container exists under the `name`-field project while a CLI-derivation tool (`@devcontainers/cli`, VS Code) operates on the same folder (the container persists in Docker, so the originating Zed session doesn't need to still be open). See issue #54255 failure mode 3 and the fixture's step 6.
>
> I'd like to pose this as a question rather than a claim: is matching `@devcontainers/cli`'s `getProjectName` precedence something the project wants to take on, given the narrowness of the bug? I wrote this implementation mostly as a way to explore what parity would actually cost — happy to close it if you'd rather leave it as-is, or pare it down (e.g. just rule 4) if a partial match is preferable.
>
> The broader value beyond this specific bug: devcontainer impls agreeing on the same project name means containers created by Zed, the devcontainer CLI, and VS Code are interchangeable for the same folder, which feels worth it to me — but you know the project's priorities better.
>
> Folds in #54068 (detection) — closing that PR unmerged; its `MultipleMatchingContainers` error lands here.

Self-Review Checklist:

Closes #54255

## Summary

**Match `@devcontainers/cli`'s full `getProjectName` precedence.** Replaces `safe_id_lower(devcontainer.json's name)` with the five-step chain the reference CLI walks (see [`src/spec-node/dockerCompose.ts` in devcontainers/cli](https://github.com/devcontainers/cli/blob/main/src/spec-node/dockerCompose.ts)):

1. `COMPOSE_PROJECT_NAME` from the local environment.
2. `COMPOSE_PROJECT_NAME=` in the workspace `.env` file.
3. Top-level `name:` on the merged compose config, when at least one fragment declared it explicitly.
4. `${workspaceFolderBasename}_devcontainer` — only when the first compose file's directory is `<workspace>/.devcontainer/`.
5. Otherwise, the plain basename of the first compose file's directory (no suffix).

The old Zed implementation diverged at every one of those inputs: any user setting `COMPOSE_PROJECT_NAME`, shipping a `.env` with one, declaring a top-level compose `name:`, or pointing `dockerComposeFile` outside `.devcontainer/` (e.g. `"../docker-compose.yml"`) got a different project namespace than the CLI and VS Code, producing two compose projects for the same folder.

Adds a small `sanitize_compose_project_name()` helper implementing the CLI's rules (lowercase + strip `[^-_a-z0-9]`) — notably preserving hyphens, which `safe_id_lower` would have replaced with underscores.

Adds two helpers used by the precedence walk:
- `parse_dotenv_compose_project_name` — line scan extracting `COMPOSE_PROJECT_NAME=…` from the workspace `.env`, matching the subset the CLI's regex dotenv reader recognizes.
- `compose_fragment_declares_name` — parses each compose fragment with `yaml-rust2` (already a transitive workspace dep; slated to become a direct dep via #53922) and checks for a `name` key on the root mapping (block, quoted, or flow style all work), matching the CLI's own `yaml.load`. `docker compose config` always injects `name: devcontainer` into its merged output when no fragment declared one, so rule 3 needs to distinguish the user-provided case from the injected default — this helper supplies that signal. On YAML parse failure it returns "not declared" (rule 4 applies), matching the CLI's fallback.

`project_name()` becomes async and fallible (`async fn project_name(&self) -> Result<String, DevContainerError>`) so it can load the `.env` file and each compose fragment via `self.fs.load`. Four call sites now `.await?` the derivation. Real I/O errors on the `.env` read propagate as `FilesystemError` (matching the CLI's narrow `ENOENT`/`EISDIR` swallow); fragment-rescan read errors are logged and skipped (matching the CLI's broader try/catch over its fragment read + parse).

The `name` field is still used as the features image-tag prefix (`generate_features_image_tag`); only the compose project namespace is decoupled from it.

**Duplicate-container detection (from #54068).** When `check_for_existing_container`'s label-based lookup returns more than one match, propagate `MultipleMatchingContainers(ids)` with instructions to clean up the stale one(s). This covers the mixed-version upgrade edge case where a pre-fix Zed left a container under the legacy project name alongside a CLI-style one — transparent to users in the common case (one tool, one container), explicit error when two legacy siblings need manual cleanup.

## Why

Full write-up with verified fixtures and captured output: #54255.

Three failure modes from the same root cause, all resolved by this change:

1. **Interop** — opening a folder in both Zed and `devcontainer up` (or Zed and VS Code) creates two compose projects with identical `devcontainer.local_folder` + `devcontainer.config_file` labels, breaking the spec's uniqueness invariant.
2. **Cross-worktree silent db/volume reuse** — if multiple git worktrees share a `devcontainer.json` with the same `name`, Zed uses the same compose project for all of them; Compose reuses stateful siblings (db, cache, localstack) by config-hash, so worktree B silently inherits worktree A's database. Fixture + captured output: [antont/zed-devcontainer-db-share-repro](https://github.com/antont/zed-devcontainer-db-share-repro).
3. **Mixed-version Zed sessions** — the Rust impl landed in stable v0.232.2 (2026-04-15, #52338). Older Zed (≤v0.231.x) shelled out to `@devcontainers/cli` so it used the reference derivation. The collision shows up when a new-Zed-created container exists under the name-field project while a CLI-derivation tool (old Zed, `devcontainer up`, VS Code) operates on the same folder.

## Migration / compatibility

Existing Zed-created containers (under the old `safe_id_lower(name)` project) continue to be found via `check_for_existing_container`'s label-based lookup — they're looked up by `devcontainer.local_folder` + `devcontainer.config_file`, not by project name. A user with duplicate legacy containers from a prior Zed session sees `MultipleMatchingContainers` with cleanup instructions.

## Revision — 2026-04-22

Revised per @KyleBarton review on the prior version:
- Swapped the YAML parser from `serde_yaml_ng` to `yaml-rust2` (already transitive via `tree-sitter-yaml`; net reduction of one direct workspace dep; also what #53922 will pull in).
- Dropped the mixed-version tiebreak (`pick_canonical_container`) and its `com.docker.compose.project` serde label. The edge case it covered is transient enough to address via the explicit `MultipleMatchingContainers` error rather than permanent tiebreaking code.
- Folded #54068's detection commit into this PR; #54068 closed unmerged.
- Rebased onto `main`.

## Test plan

- [x] `cargo test -p dev_container --lib` — 89 passed, including:
  - `sanitize_compose_project_name_matches_cli_rules`
  - `--project-name` assertion added to `test_spawns_devcontainer_with_docker_compose`
  - `check_for_existing_container_errors_when_multiple_match`
  - `derive_project_name_env_wins_over_everything`
  - `derive_project_name_dotenv_wins_over_compose_and_fallback`
  - `derive_project_name_compose_name_wins_over_fallback`
  - `derive_project_name_skips_compose_name_when_not_explicitly_declared`
  - `derive_project_name_omits_suffix_when_compose_file_outside_devcontainer_dir`
  - `derive_project_name_normalizes_compose_path_for_rule_4`
  - `compose_fragment_declares_name_detects_top_level_name_key` (covers block, quoted-key, and flow-style roots, plus parse failure → not-declared)
  - `is_missing_file_error_only_accepts_notfound_and_isadirectory`
- [x] `cargo fmt --all` — clean
- [x] `./script/clippy -p dev_container` — clean
- [x] **End-to-end with fixture** [antont/zed-devcontainer-compose-test](https://github.com/antont/zed-devcontainer-compose-test):
  - Build `zed` from this branch.
  - Clean slate: `docker ps -a --filter "label=devcontainer.local_folder=$PWD" -q | xargs -r docker rm -f`
  - `zed --dev-container /path/to/devcontainer-compose-test` → Zed creates container under project `devcontainer-compose-test_devcontainer` (was `compose_duplicate_repro` before the fix).
  - `devcontainer up --workspace-folder $PWD` → CLI reports the same `containerId` Zed created; no second compose project is introduced.
  - Captured: `devcontainer-compose-test_devcontainer-app-1`, `composeProjectName: "devcontainer-compose-test_devcontainer"` reported by both tools.

Release Notes:

- Fixed dev container Docker Compose project name now matches the full `getProjectName` precedence from the reference devcontainer CLI (`COMPOSE_PROJECT_NAME` in the environment, then in the workspace `.env`, then an explicit top-level `name:` on the merged compose config, then the basename of the first compose file's directory — with the `_devcontainer` suffix only when that directory is `<workspace>/.devcontainer`). This prevents duplicate containers when the same folder is opened with both Zed and the devcontainer CLI / VS Code.


crates/dev_container/src/devcontainer_manifest.rs
```diff
@@ -2086,27 +2089,144 @@ RUN sed -i -E 's/((^|\s)PATH=)([^\$]*)$/\1\${PATH:-\3}/g' /etc/profile || true
...
+        let first_compose_file = compose_resources
+            .as_ref()
+            .and_then(|r| r.files.first())
+            .map(PathBuf::as_path);
+        let compose_config_name = compose_resources
+            .as_ref()
+            .and_then(|r| r.config.name.as_deref());
+        let mut compose_name_explicitly_declared = false;
+        if let Some(resources) = &compose_resources {
+            for file in &resources.files {
```
KyleBarton, line 2130: So, the `DockerComposeConfig` returned by `self.docker_compose_manifest()` is supposed to be the json representation of all of these resource files, produced by `docker compose config -f {file1} -f {file2}... --format=json`. Do we need to parse these files further like we do here? Wouldn't this be covered by that json representation?

crates/dev_container/src/devcontainer_manifest.rs
```diff
@@ -2086,27 +2089,144 @@ RUN sed -i -E 's/((^|\s)PATH=)([^\$]*)$/\1\${PATH:-\3}/g' /etc/profile || true
...
+    /// `self.project_name()`. Falls through to `MultipleMatchingContainers`
+    /// when zero or ≥2 candidates claim the canonical project — users then
+    /// have to clean up the duplicate state themselves, per #54068.
+    ///
+    /// Covers the narrow case where a pre-fix Zed left a container under the
+    /// legacy `safe_id_lower(devcontainer.name)` project and a CLI-style
+    /// container for the same folder now coexists — the identifying-labels
+    /// query returns both, and the canonical-project filter picks the one
+    /// matching the current derivation.
+    async fn pick_canonical_container(
```
KyleBarton: I'm a little confused by this - this is a scenario where 
1) Zed creates a legacy safe_id_lower dev container
2) The CLI creates a second dev container
3) The post-merge-of-this-PR version of Zed then tries to open the container?

I'm inclined to error out in this case and asking the user to clean up, a la your `MultipleMatchingContainers` error message. Dev containers should be fairly lightweight, so housing this logic feels a bit heavy for a disappearing use-case. Open to pushback if you feel otherwise.

KyleBarton (changes requested):
Thanks for all the deep diving here! Some feedback:

- based on my comment in `devcontainer_manifest`, do we need do be parsing yaml here, or can we rely on the json representation of the docker compose manifest in order to decide whether a name is explicitly specified? And if this isn't needed, can we remove the yaml parsing dependency?
- If we _do_ need to pull in a yaml dependency here, then we should probably pull in `rust-yaml2`, since it is a transitive dependency in Zed already. This is discussed a bit here: https://github.com/zed-industries/zed/issues/53922 (the work of which will requires us to eventually take this dependency, I think)
- I'm content to close #54068 and just focus on getting this more holistic change over the line -- any objections?

crates/dev_container/src/devcontainer_manifest.rs
antont: @KyleBarton 

> I'm inclined to error out in this case and asking the user to clean up, a la your `MultipleMatchingContainers` error message. Dev containers should be fairly lightweight, so housing this logic feels a bit heavy for a disappearing use-case. Open to pushback if you feel otherwise.

I think we are aligned here. We should not put permanent cruft in to deal with a fringe situation which is just a temporary like during a transition anyway.

It was just practical for me to get a working version of Zed now that just deals with the issue and cleans up automatically, to use for my daily work now.

crates/dev_container/src/devcontainer_manifest.rs
antont: @KyleBarton 

> So, the `DockerComposeConfig` returned by `self.docker_compose_manifest()` is supposed to be the json representation of all of these resource files

Yes, but apparently the JSON does not tell us if that name is declared by the user in YAML, or autogenerated by Docker, and the Typescript devcontainer implementation naming differs in those cases.

If we trust what my dev agent says here - I did see it verify that by creating a test devcontainer and running those docker and devcontainer CLI commands:
   
>   The JSON field unfortunately can't carry the signal we need. docker compose config always emits a name:, synthesizing a default when no  fragment declared one — and Docker's default is different from the devcontainer CLI's getProjectName default, so using the JSON directly would re-introduce the divergence this PR is fixing.                                                                                                                                                              

>   Concrete: workspace `myproject/` with `.devcontainer/docker-compose.yml` and no user `name`:.                                                  

   
```
  $ docker compose -f .devcontainer/docker-compose.yml config --format=json | jq .name                                                     
  "devcontainer"                      # Docker's default: basename of compose file's parent dir                                            
```                                                                                                                                           

> But @devcontainers/cli's getProjectName rule 4 produces myproject_devcontainer for that same layout (the `<workspaceFolderBasename>_devcontainer` rule, since the compose file is under .devcontainer/). If Zed trusts the JSON, we match docker compose up but diverge from the CLI / VS Code — same class of bug as before, different surface.                             

> And rule 3 needs a signal the JSON destroys: "did the user explicitly declare name:?" — because if yes, we honor it (e.g. name: devcontainer on purpose); if no, we skip rule 3 and fall to rule 4. Both cases produce "name": "devcontainer" in the JSON output, byte-identical. The only way to distinguish user-declared from Compose-synthesized is to re-parse the raw YAML fragments for a top-level name: key. That's what `dockerCompose.ts`:663-673 in the reference CLI does (the `yaml.load` inside the try/catch), and what `compose_fragment_declares_name` mirrors here.

> So the YAML parse is load-bearing for parity — dropping it would quietly break rule 3 whenever the user's folder basename happens to match the injected default, and break rule 4 universally for the standard .devcontainer/ layout.

Maybe it's stupid that Docker and the Devcontainer impl diverge there? But that's apparently how it is now.

I'll switch this branch to use `rust-yaml2` in case it's useful, but I don't know really what the best course of action here would be overall.

antont:
@KyleBarton

**YAML parser.** Accepted. Swapped to `yaml-rust2`. It was already transitive in the workspace (via `tree-sitter-yaml` and, previously, `serde_yaml_ng → unsafe-libyaml`), so this is a net reduction of one direct workspace dep — and it'll line up with #53922 when that lands.

**tiebreak + #54068.** Accepted. Dropped the `pick_canonical_container` stack and its `com.docker.compose.project` serde label. The mixed-legacy-container edge is transient enough to handle via the explicit `MultipleMatchingContainers` error (from #54068) rather than permanent tiebreaking code. Folded #54068 in and closed it unmerged — the detection commit is on this branch now.

Also rebased onto `main`, targeted that and updated the PR description accordingly.

antont:
Re-verified end-to-end against the revised branch with [antont/zed-devcontainer-compose-test](https://github.com/antont/zed-devcontainer-compose-test):

- Zed opens the fixture and creates container under project `devcontainer-compose-test_devcontainer` (CLI rule 4).
- `devcontainer up --workspace-folder $PWD` from a separate shell reports `"containerId": "369feedbbcec..."` — the exact container Zed is using — and `"composeProjectName": "devcontainer-compose-test_devcontainer"`. No second compose project spawned.

crates/dev_container/src/devcontainer_manifest.rs
KyleBarton: Cool, ok I got a bit caught up and this makes sense. 
> Maybe it's stupid that Docker and the Devcontainer impl diverge there? But that's apparently how it is now.

Unfortunately I think it has to be the case. If the user uses an explicit `name` field, it makes intuitive sense to use it. But the default `name` is just the containing directory name, which means that if devcontainer implementations used it, 99% of all docker-compose containers would just be called `devcontainer`.

I'm convinced, and this will set the stage for #53922 nicely anyway

crates/dev_container/src/devcontainer_manifest.rs
```diff
@@ -2341,6 +2409,140 @@ fn escape_regex_chars(input: &str) -> String {
...
+    }
+    // `docker_compose_manifest()` joins `self.config_directory` with each
+    // raw `dockerComposeFile` entry verbatim, so paths can carry `..`
+    // components. Normalize before comparing against `<workspace>/.devcontainer`
+    // and before taking `file_name()`, otherwise `subdir/..` under
+    // `.devcontainer` misses rule 4 and a trailing `..` produces `None`
+    // from `file_name()`.
+    let compose_dir = first_compose_file
+        .and_then(Path::parent)
+        .map(normalize_path);
```
KyleBarton: How would you feel about just doing this for all files upfront in `docker_compose_manifest` instead? I think it would be better and lead to fewer surprises in the future

KyleBarton (commented):
Looking good - I had one comment about normalizing filepaths earlier in the call path. Really appreciate you digging in here!

KyleBarton:
Oh and mind the checkstyle error on spelling - parsable vs parseable

crates/dev_container/src/devcontainer_manifest.rs
antont: @KyleBarton seems like a good idea, thanks, am giving it a shot!

antont: Done in 1baf48dd7c. Moved the normalization into `docker_compose_manifest` so every downstream consumer of `DockerComposeResources.files` (fragment reads, project-name derivation, `docker compose -f` invocations) sees resolved paths. `derive_project_name` now just reads `first_compose_file.and_then(Path::parent)`; dropped the redundant comment too.

Existing `derive_project_name_handles_resolved_paths_from_docker_compose_manifest` test reframed to pin rule-4/rule-5 behavior on the already-normalized paths the helper now receives.

KyleBarton (approved):

PR #54306, opened 2026-04-20
ep: Change kept_rate definition to a more intuitive one

osyvokon:
This change contains a number of fixes to make kept_rate more intuitive. It also adds a CLI utility to print debug info on how the metric is computed.

Release Notes:

- N/A


eholk (commented):
@zed-industries/approved

PR #54316, opened 2026-04-20
agent_ui: Add setting for turning off content max-width

danilo-leal:
Follow up to https://github.com/zed-industries/zed/pull/52730. 

This PR adds a boolean setting `limit_content_width` in the Agent Panel settings that allows turning off the content max-width entirely, which was added for better readability. We had a handful of requests for it, so it feels fair.

Screenshot 2026-04-20 at 8  57@2x

Release Notes:

- Agent: Added a new `limit_content_width` setting in the agent panel that allows turning off the content max-width limit.

### DIFF
--- a/crates/agent_settings/src/agent_settings.rs
+++ b/crates/agent_settings/src/agent_settings.rs
@@ -142,7 +142,7 @@ pub struct AgentSettings {
     pub sidebar_side: SidebarDockPosition,
     pub default_width: Pixels,
     pub default_height: Pixels,
-    pub max_content_width: Pixels,
+    pub max_content_width: Option<Pixels>,
     pub default_model: Option<LanguageModelSelection>,
     pub inline_assistant_model: Option<LanguageModelSelection>,
     pub inline_assistant_use_streaming_tools: bool,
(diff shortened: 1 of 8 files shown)


crates/agent_settings/src/agent_settings.rs
```diff
@@ -593,7 +593,11 @@ impl Settings for AgentSettings {
             sidebar_side: agent.sidebar_side.unwrap(),
             default_width: px(agent.default_width.unwrap()),
             default_height: px(agent.default_height.unwrap()),
-            max_content_width: px(agent.max_content_width.unwrap()),
+            max_content_width: if agent.limit_content_width.unwrap() {
```
eholk, line 596: This is a broader cleanup and not needed for this PR, but could we refactor this so we don't need all these `unwrap`s?

eholk:
@zed-industries/approved

crates/agent_settings/src/agent_settings.rs
danilo-leal: I think so! Probably valid for the entirety of agent settings, I guess 😄

maxbrunsfeld: FWIW, I think the unwrap is effectively asserting that we have a default value for the setting in `default.json`

PR #54317, opened 2026-04-20
sidebar: Open project header ellipsis menu on right-click

danilo-leal:
This is just a refinement given it's pretty common for menus under ellipsis icon buttons to also open as a context-menu through the mouse's right-button click. This should make it slightly more convenient to interact with this menu.

Release Notes:

- N/A


maxbrunsfeld:
@zed-industries/approved

PR #54206, opened 2026-04-17
Feature flag overrides

mikayla-maki:
This PR revamps our feature flag system, to enable richer iteration. 

Feature flags can now:
- Support enum values, for richer configuration
- Be manually set via the settings file
- Be manually set via the settings UI

This PR also adds a feature flag to demonstrate this behavior, a `agent-thread-worktree-label`, which controls which how the worktree tag UI displays.

Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Screenshot 2026-04-17 at 11 14 54 PM

Release Notes:

- N/A

PR #54332, opened 2026-04-20
feature_flags: Fix issue where staff is not automatically signed into collab

bennetbo:
Follow up to #54206

`on_flags_ready` relied on the fact that the `FeatureFlagStore` was only set once the flags had been received from the server.
However, after #54206 the global gets instantiated earlier, without the flags being resolved.

Self-Review Checklist:

Release Notes:

- N/A


MrSubidubi:
@zed-zippy approved

PR #54342, opened 2026-04-20
compliance: Allow Zippy version bumps

MrSubidubi:
This adds checks for the Zed Zippy version bumps. These are rather strict, but since we should also be strict with the bumps themselves, I consider this the better approach.

Note this intentionally only includes the version bumps here for now.

Release Notes:

- N/A


ConradIrwin (approved):
Thanks!

PR #54347, opened 2026-04-20
Add new settings to improve split-pane behavior

p-avital:
Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Added a `new_tab_on_gotos` option that swaps behavior between `Ctrl+Click` and `Ctrl+Alt+Click`, allowing users to use split-pane behavior by default when performing lookups.
- Added a `prefer_left_pane_to_new_panes` option which, when set to true, makes Zed prefer swapping to a pane on the left rather than create a new pane on the right when a request for an adjacent pane is made; making the navigation experience more consistent and avoiding pane cascades.

# Author comments

Hi there!

This is my first PR to Zed, so I tried to keep the embedded model above as intact as possible.
Here's some more details on what I did and why.

This PR just proposes additional settings that tweak behavior slightly, and are off-by-default to avoid disrupting other users. I'll highlight caveats when relevant.

## `new_tab_on_gotos`

This one is kinda self-explanatory, I like to use `Ctrl+Click` in other editors to peek at code and dislike how doing so in Zed hides the current editor entirely behind the referenced one.

While I'd love to come with a PR that adds VSCode's Peek Editors to Zed eventually, the most straight-forward path to me was to at least solve the "steals focus" issue, which I feel allowing users to chose to open definitions in an adjacent tab with the default lookup action does for cheap.

Caveats:
- I'd like this to also work for reference search, but it doesn't yet. I simply haven't pushed my luck further yet, but I might edit this PR (or create a new one) to add support for this.
- The option might be a bit restrictive, especially if other `Ctrl+Click` behaviors (such as Peek Editors) eventually turn up. Maybe a better name for the option would be `"go_to_symbols": "in_current_pane"|"in_adjacent_pane"` which could then be expanded with `"in_peek_editor"` in the future.

## `prefer_left_pane_to_new_panes`

This setting tweaks the `adjacent_pane` function behavior so that it looks for a pane on the left first before trying to create a new pane.

This makes adjacent-pane-tied behavior much neater, as it behaves more consistently between tabs and avoids the cascade effect that this workflow leads to with current Zed.

If anything, I feel like this behavior would be a sound default, but I still left it to default to `false` to avoid breaking unsuspecting users' workflows.

### DIFF
--- a/crates/editor/src/hover_links.rs
+++ b/crates/editor/src/hover_links.rs
@@ -15,6 +15,7 @@ use std::{ops::Range, sync::LazyLock};
 use text::OffsetRangeExt;
 use theme::ActiveTheme as _;
 use util::{ResultExt, TryFutureExt as _, maybe};
+use workspace::WorkspaceSettings;
 
 #[derive(Debug)]
 pub struct HoveredLinkState {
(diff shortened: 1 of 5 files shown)


p-avital:
@cla-bot check

PR #54349, opened 2026-04-20
Try to recover even harder from linux GPU errors

ConradIrwin:
Release Notes:

- N/A


ConradIrwin:
To be honest, I'm not sure this retry loop is actually doing much. Looking at sentry we have a lot of `TextureView with label '' is invalid`. We added it because it seemed likely that after restore from sleep we'd hit transient errors; but it's hard to tell how effective it is given that hypothetical successful cases we don't get any events.

Veykril:
I think the retry loop was copied from windows, no clue how effective it is (the one in windows is also not super reliable, unsure if that one even works).

Veykril (approved):

PR #54354, opened 2026-04-20
Add 'protobuf' to path_suffixes for proto

tdanner:
I have encountered markdown files that fence protobuf source blocks with a `protobuf` tag instead of `proto`. [GitHub accepts both](https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml#L6042).

Self-Review Checklist:

~Closes #ISSUE~ (no issue)

Release Notes:

- `protobuf` accepted as markdown code fence tag for Protocol Buffers

### DIFF
--- a/extensions/proto/languages/proto/config.toml
+++ b/extensions/proto/languages/proto/config.toml
@@ -1,6 +1,6 @@
 name = "Proto"
 grammar = "proto"
-path_suffixes = ["proto"]
+path_suffixes = ["proto", "protobuf"]
 line_comments = ["// "]
 autoclose_before = ";:.,=}])>"
 brackets = [


MrSubidubi (changes requested):
Thanks for this!

The link you posted hints that GitHub this as a code fence name as opposed to a path suffix.

Could we not do the same using https://github.com/zed-industries/zed/blob/99df369cf859ed05a68fd36a91b8d903aab24f8a/crates/language_core/src/language_config.rs#L33-L34?

tdanner:
I overlooked `code_fence_block_name`. I see this takes a single value, not a list of options. I'm not sure if `proto` would still work when I set `code_fence_block_name = "protobuf"`. Maybe I should get a local build going so I can test it.

Also, Claude (possibly incorrectly) tells me that `code_fence_block_name` is _not_ used for looking up languages for markdown fence parsing. It claims:

> Today it’s used in three real places:
> 
> 1. Generating fenced markdown snippets for diagnostics/UI, so Zed writes ```<name> with that value (crates/agent_ui/src/diagnostics.rs:222).
> 2. Matching Jupyter kernel languages, as one of the accepted names for a language (crates/language/src/language.rs:979).
> 3. Looking up legacy kernel selections in REPL/Jupyter settings by language key (crates/repl/src/repl_store.rs:372).

If I'm following this right, the lookup happens at https://github.com/zed-industries/zed/blob/main/crates/language/src/syntax_map.rs#L1635 and uses `language_registry.language_for_name_or_extension` which appears to use `.path_suffixes` but not `code_fence_block_name `.

MrSubidubi:
Claude might be right, I think this might very well be a regression. 

Can I interest you to look into this? Namely,
- get this wired up again for the markdown crate
- add some tests for that
- use it in the extension
?

Similarly, we could also ensure that there is proper fallback behavior. Not insisting, but I think this would be the best way forward

tdanner:
I can try! That seems like a better fix than just adding it as a path suffix. Might be the weekend before I get around to it.

MrSubidubi:
Sure, no rush! I'll be here in case you have any questions

MrSubidubi:
Will go ahead and close this for the time being here, as we agreed upon a different approach and I'd like my backlog to be a bit more manageable at glance. Hope that does not cause you any inconvenience. 

If you have any questions once you start picking this up, please do feel free to reach out and I can answer and also reopen if you'd like me to! Alternative, feel free to open another PR and ping me there, I'll gladly take a look either way. Hope you understand, thanks!

tdanner:
@MrSubidubi - I finally got around to making a more robust version here: #56309

PR #54355, opened 2026-04-20
extension_host: Make extension updates more resilient

MrSubidubi:
Following our incident earlier, I noticed a report that mentioned their extension was uninstalled upon update which made them stumble across the issue for our outage.

The reason for this was that currently, whenever we install extension updates, we _first_ removed the old directory and _then_ went ahead and downloaded the new version. This would in turn obviousy lead to issus where something malfunctioned during the update/install download - whether that is an issue on our side with the servers or an issue on the users side when they have a poor network connection or lose that.

This PR changes this - we now do the entirety of the download on a background task, unpack the archive into a temporary directory if possible and then proceed with removing the old extension contents and moving the new contents only _after_ everything prior succeeded.

I also moved this off of the main thread since there is no reason to do this there as well as the `Drop` impl of a `TempDir` doing some blocking work on the main thread otherwise.

Release Notes:

- Made extension updates more resilient to network and upstream failures

### DIFF
--- a/crates/extension_host/src/extension_host.rs
+++ b/crates/extension_host/src/extension_host.rs
@@ -19,7 +19,7 @@ use extension::{
     ExtensionGrammarProxy, ExtensionHostProxy, ExtensionLanguageProxy,
     ExtensionLanguageServerProxy, ExtensionSnippetProxy, ExtensionThemeProxy,
 };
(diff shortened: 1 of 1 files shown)


osiewicz (approved):

osiewicz (approved):
Thank you.

PR #54357, opened 2026-04-20
language_model: Fix `ImageSize` storing pre-downscale dimensions

dastrobu:
When images are resized to meet provider size constraints (Anthropic's 1568px limit or the 5MB encoded-PNG cap), the stored ImageSize was still recording the original width/height rather than the final post-downscale dimensions. This caused incorrect token estimation via estimate_tokens() since it uses width * height / 750.

Use processed_image.dimensions() after all downscale passes so that ImageSize reflects the actual image sent to the provider.

Release Notes:

- Fixed an issue where token estimation would be incorrect in case where the thread contained downscaled images.

### DIFF
--- a/crates/language_model/src/request.rs
+++ b/crates/language_model/src/request.rs
@@ -118,10 +118,12 @@ impl LanguageModelImageExt for LanguageModelImage {
             // SAFETY: The base64 encoder should not produce non-UTF8.
             let source = unsafe { String::from_utf8_unchecked(base64_image) };
 
+            let (final_width, final_height) = processed_image.dimensions();
+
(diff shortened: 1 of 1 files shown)


MrSubidubi (changes requested):
Thanks for this!

This seems like a good fix, can we add a test for this to make sure this does not regress?

dastrobu:
> This seems like a good fix, can we add a test for this to make sure this does not regress?

@MrSubidubi absolutely, thanks for the quick feedback. Test cases is updated, please review again.

MrSubidubi (approved):
Nice find, fix and test update. Thank you for this contribution!

PR #54360, opened 2026-04-20
sidebar: Adjust display of workspaces in header's ellipsis menu

danilo-leal:
This PR only displays the X icon button to close an workspace for non-active ones, and does some light UI tweaks to how we display the project labels inside the menu item.

Release Notes:

- N/A


maxbrunsfeld (commented):
What do you think about hiding the `x` only if the workspace is the *active* one, rather than making it conditional on the number of workspaces?

maxbrunsfeld (approved):

PR #54361, opened 2026-04-20
Cleanup open thread connections on close

alphathekiwi:
Self-Review Checklist:

- [x] Unsafe blocks - None present

Closes #NA

Release Notes:

- Added cleanup logic for open thread connections using `session/close`

image

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -2162,7 +2162,7 @@ impl AgentPanel {
         self.cleanup_retained_threads(cx);
     }
 
-    fn cleanup_retained_threads(&mut self, cx: &App) {
+    fn cleanup_retained_threads(&mut self, cx: &mut App) {
         let mut potential_removals = self
             .retained_threads
             .iter()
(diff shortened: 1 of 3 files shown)


alphathekiwi:
@benbrandt 

The context around the need for this is that we have started building some internal tooling for usage of claude-acp in docker volumes and on the cloud, for these runners they have an associated cost, so its valuable to us that we are able to receive the `session/close` message from Zed at the time in which these sessions stop being actively communicated on.

As I understand it even when archiving the session its connection is held open in a small pool with a few other recently closed connections and not cleaned up immediately which while likely helpful for 90% of use-cases our desire would be to close of connections firmly with a `session/close` signal.

crates/agent_ui/src/agent_panel.rs
```diff
@@ -2240,7 +2185,61 @@ impl AgentPanel {
             .take(n)
             .collect::<Vec<_>>();
         for id in to_remove {
-            self.retained_threads.remove(&id);
+            let Some(conversation_view) = self.retained_threads.remove(&id) else {
+                continue;
+            };
+            Self::close_conversation_view_session(&conversation_view, cx);
```
alphathekiwi, line 2191: This is the main place where calling the close should be needed

benbrandt: Hmm my confusion is we are already callign session/close...when the conversation view is dropped:

https://github.com/zed-industries/zed/blob/58e2b7ecdd9170cf3c23cf04da903201a739f52f/crates/agent_ui/src/conversation_view.rs#L643-L655

benbrandt: However, I am not seeing it consistently called... Looking into it

alphathekiwi: > Hmm my confusion is we are already callign session/close...when the conversation view is dropped:
> 
> https://github.com/zed-industries/zed/blob/58e2b7ecdd9170cf3c23cf04da903201a739f52f/crates/agent_ui/src/conversation_view.rs#L643-L655

```rs
        connection
            .close_session(&session_id, cx)
            .detach_and_log_err(cx);
```

I think its also a different method being used `close_all_sessions` vs `close_session` and i assume close all sessions would drop all ACP connections?

benbrandt: It is used because of subagents. We found the bug though and have a fix we are testing. Will be out soon!

We have a way of retaining a number of threads so we don't drop them all the time when you switch around, but they weren't getting dropped from there on archive right away. 

We'll make sure archived threads get dropped from that list immediately and not just cleaned up eventually

benbrandt: This will be fixed in #54516

PR #54364, opened 2026-04-20
copilot_chat: Respect prompt token limits

Milo123459:
## Summary

Fix Copilot-backed thread token budgeting to use the provider's prompt token limit instead of the larger context window when available.

This keeps Zed's token usage UI and warning thresholds aligned with the effective backend request cap for Copilot models that expose `max_prompt_tokens`.

## Changes

- add `max_prompt_token_count()` to the `LanguageModel` trait with a default `None`
- expose `max_prompt_tokens` from Copilot model metadata
- wire the Copilot prompt token limit through the Copilot language model provider
- update thread token usage to prefer `max_prompt_token_count()` over `max_token_count()`
- extend Copilot model tests to verify both context window and prompt token limits are preserved separately

## Why

Copilot model metadata includes both:

- `max_context_window_tokens`
- `max_prompt_tokens`

Zed was using the context window as the thread token budget, which can overestimate the amount of prompt that the Copilot backend will actually accept. This leads to misleading token-limit UI and avoidable `model_max_prompt_tokens_exceeded` failures.

This change keeps `max_token_count()` as the context window and introduces a separate prompt-limit accessor so budgeting can use the more accurate value without changing the meaning of the existing API.

Release Notes:

- Fixed Copilot-backed agent threads using the model context window instead of the prompt token limit for token budgeting and limit warnings.

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -1691,7 +1691,9 @@ impl Thread {
         let usage = self.latest_request_token_usage()?;
         let model = self.model.clone()?;
         Some(acp_thread::TokenUsage {
-            max_tokens: model.max_token_count(),
+            max_tokens: model
+                .max_prompt_token_count()
+                .unwrap_or_else(|| model.max_token_count()),
             max_output_tokens: model.max_output_tokens(),
             used_tokens: usage.total_tokens(),
             input_tokens: usage.input_tokens,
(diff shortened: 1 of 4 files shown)


Milo123459:
@cla-bot check

benbrandt:
We only use this number for usage anyway, so wouldn't the context window be correct?
Going to close

PR #54367, opened 2026-04-20
cli: Make `zed --wait --diff` not wait until window close

timvermeulen:
`zed --wait --diff <left> <right>` currently waits for the entire window to close instead of just the diff itself, due to the combination of
- `zed --diff` adds the cwd to the paths to open (see #45131), and
- passing a directory to `zed --wait` blocks until the Zed window is closed (see #44936), which doesn't distinguish between explicit paths and the one `zed --diff` adds

Fixed by only running the `zed --wait <dir>` logic to block until the window is closed in case the user actually passed any non-diff paths.

I've confirmed locally that it works, though I wasn't really sure how to write a good test for this.

Self-Review Checklist:

Release Notes:

- `zed --wait --diff <left> <right>` now blocks until the diff is closed, and not until the entire window is closed.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -642,14 +642,6 @@ fn main() -> Result<()> {
         }
     }
 
-    // When only diff paths are provided (no regular paths), add the current
-    // working directory so the workspace opens with the right context.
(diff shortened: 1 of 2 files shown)


ChristopherBiscardi (approved):
Thanks :) I've verified that this works as expected

PR #54369, opened 2026-04-20
compliance: Add support for checking singular commit

MrSubidubi:
This helps with two things: Testing changes locally against real commits as well as laying the groundwork for making things less convoluted for the GitHub worker.

Also removes some useless wrapping left from an earlier direction.

Release Notes:

- N/A


ConradIrwin (approved):
makes sense to me!

PR #54371, opened 2026-04-21
fuzzy_nucleo: Fix out of range panic

MrSubidubi:
Closes ZED-6PK

The issue here was that we could hit cases where the amount of segments < amount of CPUs, e.g. for 5 candidates and 4 CPUs, we would have 2 candidates per matcher, so for the fourth matcher, we would start slicing at 3 * 2 = 6 > 5, which is out of bounds.

Instead, make it so that we distribute the candidates across all matchers so that all matchers have either n or n + 1 candidates.

No release notes since this is only on Nightly.

Release Notes:

- N/A


yara-blue (approved):
Thanks for the save!

PR #54379, opened 2026-04-21
workspace: Add configurable window title formatting

### LINKED ISSUE #14534: Configure window title pattern
### Check for existing issues

- [X] Completed

### Describe the feature

Currently, Zed's windows titles take the form of `{filename} — {project}`, e.g. `channel.rs — app`.

I would like to configure this to be the inverse: `{project} — {filename}`, e.g. `app — channel.rs`.

I tend to have one window for each project, with multiple projects open at once. Having the project name prefix the filename ensures that in my window switcher, the project name is always visible even when the title is truncated.

As-is, I can wind up with situations like:
- `Cargo.toml — project_a`
- `Cargo.toml — project_b`
- `Cargo.toml — project_c`

Truncated to:
- `Cargo.toml — pr...`
- `Cargo.toml — pr...`
- `Cargo.toml — pr...`

Searching for this, I only turned up #5283 — which should possibly be closed as complete anyway since I _am_ seeing details in the window title.

Example of the situation I can currently find myself in when switching windows: (this is the full width of my switcher)

Screenshot 2024-07-16 at 4 06 55 PM

Vision of the feature: my prior VSCode setting:

```
  "window.title": "${rootName}${separator}${activeEditorShort}",
```

jknlsn:
## Summary
- Add `window_title_format` and `window_title_separator` settings for customizing the window title
- Expose `${projectName}`, `${fileName}`, `${filePath}`, `${relativePath}`, `${fileStem}`, `${remoteName}`, `${remoteHost}`, `${appName}`, and `${separator}` as the native Zed window-title variables
- Keep the collab `↗/↙` indicator hard-appended
- Translate VS Code `window.title` / `window.titleSeparator` aliases into Zed's canonical variables during settings import, including `${appName}` passthrough support and mapping VS Code's `${activeEditorMedium}` onto Zed's `${relativePath}`
- Gate path-, remote-, and release-channel-derived lookups on whether the active template references the corresponding variable, so the default template avoids the extra work and still uses the existing `last_window_title` cache to skip redundant platform updates

An alternative approach is at #49638. I went with `${variable}` names over `$ZED_*`, canonical Zed variable names with VS Code aliases handled at import time, kept `${separator}` as the only structural token, and left `${dirty}` out of scope.

### Variables considered but not included
- **collab/shared indicator variable**: kept hard-appended instead. Folding it into the template would need whitespace/separator logic the other variables don't need, and I'd rather not bake those semantics in without a concrete need for it.
- **dirty variable**: The glyph choice and interaction with OS-level edited indicators (like macOS's titlebar dot via `set_window_edited`) felt like a separate design call I wasn't confident making here.

A separator-style config for each of collab/shared/dirty would solve the above, but I didn't want to add too many additional settings initially.

### Performance
`update_window_title` is on a hot path, so two short-circuits apply:
- Path-, remote-, and release-channel-derived lookups (`absolute_path`, relative-path string, `file_stem`, `remote_connection_options`, `ReleaseChannel::try_global`) are gated on `template.contains(...)` via `WindowTitleNeeds::from_template`, so the default template avoids the extra work.
- The existing `last_window_title` cache still skips the platform call when the rendered title is unchanged.

## Screenshots

Settings Editor entries for `window_title_format` and `window_title_separator`:
Settings Editor screenshot

Customized Zed window titles in AltTab app switcher (remote, local, empty):
AltTab app switcher screenshot

## Self-Review Checklist:

Closes #14534

Release Notes:

- Added configurable window title formatting with `window_title_format` / `window_title_separator`, including `${appName}` / `${remoteName}` / `${remoteHost}` variables and VS Code `window.title` import support


mwja:
I'd be interested if this could also include the worktree name? 

I often have two worktrees open at a time (`main` and my current feature branch), so it could be super useful to distinguish

jknlsn:
> I'd be interested if this could also include the worktree name?
> 
> I often have two worktrees open at a time (`main` and my current feature branch), so it could be super useful to distinguish

@mwja agreed, this would be useful. I've added it in now! You can use the `${branch}` variable.

If anyone from the Zed team has any feedback or direction on the implementation, I'd love to hear it :)

crates/settings/src/vscode_import.rs
```diff
@@ -1141,6 +1151,46 @@ impl VsCodeSettings {
...
+
+        let content_start = variable_start + 2;
+        let Some(content_end_offset) = template[content_start..].find('}') else {
+            translated.push_str(&template[variable_start..]);
+            return translated;
+        };
+
+        let content_end = content_start + content_end_offset;
+        let variable = &template[content_start..content_end];
+        match variable {
```
SomeoneToIgnore, line 1186: Please translate VS Code’s `${activeRepositoryBranchName}` to `${branch}` because this PR already computes and subscribes for `${branch}`, while the importer currently removes the documented VS Code branch token.

crates/settings/src/vscode_import.rs
```diff
@@ -99,6 +99,14 @@ impl VsCodeSettings {
             .map(|s| s.to_owned())
     }
 
+    fn read_window_title_format(&self) -> Option<String> {
```
SomeoneToIgnore, line 102: Please reject translated formats whose trimmed value is empty because `${unsupportedVariable} ` currently imports as one space and bypasses the normal-title fallback.

crates/workspace/src/workspace.rs
```diff
@@ -12008,6 +12444,107 @@ mod tests {
...
+        cx.update(|_, cx| {
+            SettingsStore::update_global(cx, |settings, cx| {
+                settings.update_user_settings(cx, |settings| {
+                    settings.workspace.window_title_format =
+                        Some("${projectName}${separator}${relativePath}".to_string());
+                })
+            });
+        });
+        cx.executor().run_until_parked();
+        assert_eq!(cx.window_title().as_deref(), Some("root1 — src/one.txt"));
```
SomeoneToIgnore: Please derive this expected relative path using the local project path style because Windows correctly renders `src\one.txt` while this assertion hardcodes `src/one.txt`.

crates/settings/src/vscode_import.rs
```diff
@@ -1141,6 +1151,46 @@ impl VsCodeSettings {
...
+        let content_start = variable_start + 2;
+        let Some(content_end_offset) = template[content_start..].find('}') else {
+            translated.push_str(&template[variable_start..]);
+            return translated;
+        };
+
+        let content_end = content_start + content_end_offset;
+        let variable = &template[content_start..content_end];
+        match variable {
+            "projectName" | "fileName" | "filePath" | "relativePath" | "fileStem"
```
SomeoneToIgnore, line 1187: Please translate `${remoteName}` to matching provider semantics or decline to import it because VS Code uses a provider label such as `SSH`, while this implementation resolves the same token to a nickname or hostname.

crates/workspace/src/workspace.rs
```diff
@@ -1796,6 +1967,24 @@ impl Workspace {
 
         let subscriptions = vec![
             cx.observe_window_activation(window, Self::on_window_activation_changed),
+            cx.observe_global_in::<SettingsStore>(window, |this, window, cx| {
```
SomeoneToIgnore, line 2011: Please filter settings and Git notifications by the effective title settings and `${branch}` requirement because every matching notification currently recomputes the title and reapplies the represented document path before the title cache can return.

crates/settings_ui/src/page_data.rs
```diff
@@ -4953,9 +4953,48 @@ fn window_and_layout_page() -> SettingsPage {
...
+                    },
+                    write: |settings_content, value, _| {
+                        settings_content.workspace.window_title_separator = value;
+                    },
+                }),
+                metadata: Some(Box::new(SettingsFieldMetadata {
+                    placeholder: Some(" — "),
+                    ..Default::default()
+                })),
+                files: USER,
```
SomeoneToIgnore, line 5096: I wonder if it makes sense to support this in project settings too, as something projects would prefer doing?

SomeoneToIgnore (commented):
Looks nice, thank you.

Let's fix a bunch of VSCode import-related things before merging.

crates/settings/src/vscode_import.rs
jknlsn: Good pick up — I've added the mapping so `${activeRepositoryBranchName}` imports as `${branch}`, reusing the subscription the PR already sets up. Covered in `test_import_window_title_format`. Fixed in [a8f1cdba68](https://github.com/zed-industries/zed/commit/a8f1cdba68).

crates/settings/src/vscode_import.rs
jknlsn: Got it, the importer now rejects translated formats that trim to empty, so `${unsupportedVariable} ` falls back to the normal title. Literal text still imports: `${activeFolderShort} — literal` comes through as ` — literal`. Both cases are pinned in tests. Fixed in [a8f1cdba68](https://github.com/zed-industries/zed/commit/a8f1cdba68).

crates/workspace/src/workspace.rs
jknlsn: Have done now in [a8f1cdba68](https://github.com/zed-industries/zed/commit/a8f1cdba68)! The assertions build the expected path with `RelPath::display(project.path_style(cx))`, so they assert `src\one.txt` on Windows.

crates/settings/src/vscode_import.rs
jknlsn: Went with declining it — Zed's `${remoteName}` resolves to the connection's nickname/host, so importing VS Code's token would mistranslate it. If a provider-label variable is wanted (Zed knows the connection type — ssh/wsl/docker/podman), I'm happy to add one as a follow-up. Templates that end up empty after the drop fall back to the default title per the trim change above. Fixed in [a8f1cdba68](https://github.com/zed-industries/zed/commit/a8f1cdba68).

crates/workspace/src/workspace.rs
jknlsn: Yep — the `SettingsStore` observer now only recomputes when the effective `(window_title_format, window_title_separator)` pair changed (cached in `last_window_title_settings`), and the Git subscription only fires when the active template references `${branch}`. Template changes still flow through the settings observer either way, and `refresh_window_state` bypasses the caches on workspace switches. Fixed in [a8f1cdba68](https://github.com/zed-industries/zed/commit/a8f1cdba68).

crates/settings_ui/src/page_data.rs
jknlsn: I think it'd make sense, but it needs a call on resolution semantics first — a window can host multiple workspaces (which is why this PR has the window-chrome ownership handling), so it's not obvious which project's setting should win. My inclination would be the active workspace's settings, re-resolving as the active workspace changes. That's a separate piece of work if someone wants to integrate that functionality, but I could implement it if considered essential for this PR.

SomeoneToIgnore (commented):
Thanks.

PR #54397, opened 2026-04-21
Fix agent default model not picking up after authentication resolve

smitbarmase:
Regression in https://github.com/zed-industries/zed/pull/54125

Release Notes:

- agent: Fixed an issue where the default Zed model would not get selected after sign-in completed

### DIFF
--- a/crates/language_models/src/language_models.rs
+++ b/crates/language_models/src/language_models.rs
@@ -119,19 +119,6 @@ pub fn init(user_store: Entity<UserStore>, client: Arc<Client>, cx: &mut App) {
         );
     });
 
-    cx.subscribe(
-        &registry,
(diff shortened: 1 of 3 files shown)


crates/language_models/src/language_models.rs
```diff
@@ -119,19 +119,6 @@ pub fn init(user_store: Entity<UserStore>, client: Arc<Client>, cx: &mut App) {
         );
     });
 
-    cx.subscribe(
```
bennetbo, line 122: We removed this because it causes flickering issues, e.g. if multiple providers authenticate one after another. Instead we only want to update the fallback once after all authenticate tasks completed. And because the Zed provider now actually returns a task we can rely on only calling `update_environment_fallback_model` once all have authenticated

bennetbo (approved):

Issue #52993, opened 2026-04-02
`AgentPanel` thinks `last_active_thread` doesn't exist and overwrites it


phdavis1027:
### Reproduction steps

1. Start Zed, open a project
2. Start a Claude thread, send a few messages
3. Close Zed
4. Open Zed again, such that the same project is opened on start

Best-effort red test [here](https://github.com/phdavis1027/zed/commit/820a8a8f373374ed9510cdd8f834bfc41a8682fb)

### Current vs. Expected behavior

# Expected
In step 4, the same Claude thread from step 2 is open

# Current
A new thread is opened

# Miscellaneous note
- Besides just personal observation, I have personally verified that the sessions in question exist. However, the recent 'skip if missing metadata' sees them as non-existent. More details in my comments on the possibly-related #52253 

### Zed version and system specs

Zed 0.229.0 
7c07887d9555953bca7fe78417602114092ef8c8

0.229.0+stable.206.7c07887d9555953bca7fe78417602114092ef8c8

Fedora 43, Hyprland 0.54.2

### Attach Zed log file

<details><summary>Zed.log (apologies for the spam, relevant greps are 'LoadError' and 'skipping restoration')</summary>

2026-04-02T10:20:51-04:00 INFO  [zed::zed] Using GPU: GpuSpecs { is_software_emulated: false, device_name: "AMD Radeon Graphics (RADV RENOIR)", driver_name: "radv", driver_info: "Mesa 25.3.6" }
2026-04-02T10:20:51-04:00 INFO  [client] set status on client 183186: Authenticated
2026-04-02T10:20:52-04:00 INFO  [agent_ui::agent_panel] Global last used agent: Some(Custom { id: AgentId("claude-acp") })
2026-04-02T10:20:52-04:00 WARN  [agent_ui::agent_panel] last active thread 6940d66f-e599-4872-844b-11c7204550a0 has no metadata, skipping restoration
[crates/agent_ui/src/agent_panel.rs:783:13] &last_active_thread = None
2026-04-02T10:20:52-04:00 INFO  [agent_ui::conversation_view] ConversationView::set_server_state: LoadError(session=None, err=Custom agent server `claude-acp` is not registered)
2026-04-02T10:20:52-04:00 INFO  [agent_ui::agent_panel] agent_panel::serialize: workspace_id=19, session_id=None
2026-04-02T10:20:52-04:00 INFO  [agent_ui::agent_panel] agent_panel::serialize: workspace_id=19, session_id=None
[...]

PR #54411, opened 2026-04-21
agent_ui: Preserve session resume state after load errors

benbrandt:
Use stored thread metadata during reset when no root thread view exists,
so retries keep the original session id and title instead of starting a
new session.

Add a regression test for reconnecting after an initial custom agent
load failure.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -723,18 +723,43 @@ impl AgentPanel {
         let selected_agent = self.selected_agent.clone();
 
         let is_draft_active = self.active_thread_is_draft(cx);
-        let last_active_thread = self.active_agent_thread(cx).map(|thread| {
-            let thread = thread.read(cx);
+        let last_active_thread = self
+            .active_agent_thread(cx)
+            .map(|thread| {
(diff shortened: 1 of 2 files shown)


benbrandt:
Related to #52993

crates/agent_ui/src/agent_panel.rs
```diff
@@ -3755,6 +3780,35 @@ impl AgentPanel {
         self.set_base_view(thread.into(), true, window, cx);
     }
 
+    /// Opens a restored external thread with an arbitrary AgentServer and
+    /// a specific `resume_session_id` — as if we just restored from the KVP.
+    ///
+    /// Test-only helper. Not compiled into production builds.
+    pub fn open_restored_thread_with_server(
```
bennetbo, line 3787: Let's put this behind cfg(test) if it's not used in production

bennetbo (approved):

crates/agent_ui/src/agent_panel.rs
benbrandt: I got confused by the same thing... but it is in an impl block that is behind a cfg flag :D

PR #28752, opened 2025-04-15
Case Insensitive Unicode Text Search: Fallback To Regex

mockersf:
Closes #9980

Release Notes:

- Fixed: case insensitive text search with unicode characters

PR #54413, opened 2026-04-21
project: Fix infinite search loop for literal queries with non-ASCII and regex-specific characters

### LINKED ISSUE #54331: Project search: Replace all hangs in infinite loop when search/replace text shares a common prefix with `##`
### Reproduction steps

1. Start Zed and open project folder (for me, it's obsidian vault)
2. Hit Ctrl + Shift + H to open replace text input on project search
3. Type `## この日に作成したノート` (Notes created on the day) as original text
4. Hit Enter to execute the search before entering the replace text.
5. Type `## この日に関連するノート` (Notes related to the day) as replace text
6. Hit Ctrl + Alt + Enter to replace all
7. The GUI starts blinking and repeats the replacement indefinitely.

\*1. This issue continues until I close the project search window.
\*2. This issue also occurs when replacing text one by one, not just with replace all.

Current behavior:

As the following video:

https://youtu.be/WBAKK5YGRBc

What I expected the behavior to be:

I expected the replacement to complete normally, just as it does when the search/replace text does not include `##`.

Zed: v0.232.3+stable (Zed)
OS: Arch Linux Wayland Hyprland
Memory: 15.5 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3060 || NVIDIA || 595.58.03

### Attach Zed log file

No relevant log entries were observed during reproduction.

### Relevant Zed settings

```json
{
    "base_keymap": "VSCode",
}
```

\*Trimmed irrelevant settings

</details>

### Relevant Keymap

No custom bindings relevant to this issue.

_No response_

None

lingyaochu:
Self-Review Checklist:

Closes #54331

In #28752, in order to get case-insensitive search for non-ASCII queries, we introduced a regex fallback:
https://github.com/zed-industries/zed/blob/cfebe3a85af0fab9688c275ecf22aa30c3501cc6/crates/project/src/search.rs#L103-L118

Inside the `regex()` build method, we store the passed query in the `SearchInputs` as `initial_query`, which is intended to represent the exact user input. However, when these two parts are combined during the Unicode fallback, this logic breaks.

If a user input a "# 😀", the code detects non-ASCII characters and passes the query to the regex constructor. To ensure it still behaves like a literal search, we call `regex::escape()`, passing "\\# 😀" instead of the raw input. This causes the internal identity of the search to become the escaped string.

The "Replace All" logic in project search then performs a staleness check to see if the query has been updated:
https://github.com/zed-industries/zed/blob/cfebe3a85af0fab9688c275ecf22aa30c3501cc6/crates/search/src/project_search.rs#L839-L849
It compares the raw text in the editor ("# 😀") against the last search query text ("\\# 😀"). Because they never match, Zed assumes the query has changed and  triggers a refresh search to update the results. This leads to an endless cycle of mismatches and retries.

This PR ensures that the original, non-escaped query is stored correctly in the search identity, even when the logic requires an escaped regex fallback. It's also worth noting that the regression test included here won't produce a standard error on the current main branch; instead, it will run endlessly, which is the exact behavior being fixed.

Release Notes:

- Fixed an infinite loop in Project Search & Replace when using non-ASCII characters and regex-specific symbols with Case Sensitivity disabled.

### DIFF
--- a/crates/project/src/search.rs
+++ b/crates/project/src/search.rs
@@ -115,7 +115,13 @@ impl SearchQuery {
                 files_to_exclude,
                 false,
                 buffers,
-            );
+            )
+            .map(|mut search_query| {
+                if let Self::Regex { ref mut inner, .. } = search_query {
+                    inner.query = Arc::from(query.as_str());
+                }
+                search_query
+            });
         }
         let search = AhoCorasickBuilder::new()
             .ascii_case_insensitive(!case_sensitive)
--- a/crates/search/src/project_search.rs
+++ b/crates/search/src/project_search.rs
@@ -5466,4 +5466,71 @@ pub mod tests {
         );
         cx.background_executor.run_until_parked();
     }
+
+    #[gpui::test]
+    async fn test_project_search_replace_all_with_unicode_and_regex_chars(cx: &mut TestAppContext) {
+        init_test(cx);
+
+        let fs = FakeFs::new(cx.background_executor.clone());
+        fs.insert_tree(
+            path!("/dir"),
+            json!({
+                "file.rs": "😀#12345\n😀#12345\n",
+            }),
+        )
+        .await;
+        let project = Project::test(fs.clone(), [path!("/dir").as_ref()], cx).await;
+        let window =
+            cx.add_window(|window, cx| MultiWorkspace::test_new(project.clone(), window, cx));
+        let workspace = window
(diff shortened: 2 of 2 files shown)


ConradIrwin:
@lingyaochu nice, thanks for this. I think we should go with the version in the other PR to keep the API a bit harder to misuse in the future.

PR #54433, opened 2026-04-21
gpui: Add WebView element backed by wry + WebKitGTK

MostlyKIGuess:
Self-Review Checklist:

## Motivation

- REPL output (Plotly/Bokeh), Jupyter notebook rendering, rich agent UI -- anywhere we need an interactive HTML/JS surface inside a Zed window. This will stand as the final stone required to finish Jupyter Notebook integration.

## Summary

- Add a `WebView` GPUI element gated behind the `webview` feature flag, backed by [`wry`](https://crates.io/crates/wry) (WebKitGTK on Linux, WebView2 on Windows, WKWebView on macOS).

Example:

```sh
cargo run -p gpui --example webview --features webview
````

## Screenshots:

image
image
image

image

## What works

* [x] Windows - `build_as_child()` creates a child `HWND`. Inline, follows layout.
* [x] Linux / X11 - `build_as_child()` bridges GPUI's XCB window to GDK via
  `gdk_x11_window_foreign_new_for_display`. GTK event loop runs on a dedicated thread (`x11_child_thread`) to avoid
  blocking GPUI's render loop; the main thread extracts the raw `Xlib` window ID and ships it over a channel.
* [x] macOS - `wry::WebViewBuilder::build_as_child()` attaches the `WKWebView` as an `NSView` subview of GPUI's
  Metal-backed `NSView`. Inline, follows layout.

### What doesn't work inline: Linux / Wayland

- Because gtk is a deprecated crate, it will no longer fix the glib dependency and wry doesn't support gtk4 yet. Until then we can perhaps fork gtk-rs-core like [this](https://github.com/MostlyKIGuess/gtk-rs-core/tree/fix-variantstr-iter-0.18) and fix the vulnerability and use it as a patched crate for glib.
- Regarding the LICENCE test failure, https://cdla.dev/permissive-2-0/ not sure what Zed stands on this.

On Wayland the webview opens in its own GTK window on a dedicated GTK thread. It works. It is not inside the Zed window, which is the part I want input on.

The blocker is `wry::WebViewBuilder::build_as_child`. In wry 0.55 that code path is X11 only. The code takes the parent's Xlib window, puts it in a GDK window via `gdk_x11_window_foreign_new_for_display` and then adds a GTK widget tree to it. There is no Wayland equivalent in wry. 

The wry docs confirm this and point to `WebViewBuilderExtUnix::new_gtk`, but that function needs a `gtk::Container` to embed into. Zed's Wayland window is `raw wl_surface` with an `xdg_toplevel` role which doesn't ace like a GTK widget.

On Wayland, no `XEmbed` equivalent exists. Although `wl_subsurface` exists, but GTK3 and WebKitGTK do not expose a public way to give up their `xdg_toplevel` role and become a subsurface of a foreign compositor connection.

How other projects did this:

- Tauri's window itself is a GTK window, so wry embeds into it with no extra work.
- Bevy, iced, Slint, and egui do not ship a native inline webview. When they need one, they open a separate window,
  which is the same fallback I did here.

Some paths we can take:

- Add `xdg_foreign_v2` support so the webview window can call set_parent_of on the Zed window. The compositor keeps
  them grouped and moves them together. Still a separate window, but feels attached.
- Offscreen WebKit rendering into a texture that Zed composites through wgpu. True inline. Large scope, no prior art in wry.
- Upstream a Wayland child path into wry. Long timeline.

I would like the maintainers to chime in before investing more time here. Open to other directions.

Release Notes:

- N/A


MostlyKIGuess:
edit: moved screenshots to main body

ConradIrwin:
Thanks for this @MostlyKIGuess!

Right now this seems disproportionate (adding a web view is a comparatively giant feature) for better support for workbooks (which we do want, but not necessarily this much yet).

Particularly given the problems you outline above with Linux, I think we should close this.

Not saying we should never ever add a web view, but if we do, it needs to be done with an eye to keeping the dependency tree and binary requirements minimal as the majority of our users are not going to benefit from it.

MostlyKIGuess:
> Thanks for this @MostlyKIGuess!
> 
> Right now this seems disproportionate (adding a web view is a comparatively giant feature) for better support for workbooks (which we do want, but not necessarily this much yet).
> 
> Particularly given the problems you outline above with Linux, I think we should close this.
> 
> Not saying we should never ever add a web view, but if we do, it needs to be done with an eye to keeping the dependency tree and binary requirements minimal as the majority of our users are not going to benefit from it.

Makes sense, thanks.
I'll focus on work closer to the roadmap, especially around notebooks. Will keep an eye on ongoing efforts there.

PR #54444, opened 2026-04-21
Make guild member check case-insensitive in PR labeler

JosephTLyons:
Fixes the casing of the remaining names and adjust the script to allow any casing of handles.

Release Notes:

- N/A


ConradIrwin (approved):
Seems reasonable to me. Maybe we could list them all in lowercase to save having to downcast every time (but not a blocker)

PR #54224, opened 2026-04-18
workspace: Fix recent-projects cleanup wiping active workspaces

transitoryangel:
The recent-projects picker, sidebar, welcome screen, and agent thread store all called `recent_workspaces_on_disk`, which combined listing with deleting stale rows. Its retention predicate rejected workspaces with no on-disk directory, including empty workspaces holding unsaved scratch buffers, and the resulting `delete_workspace_by_id` call cascaded into `items`, `pane_groups`, and the per-editor tables. For clarity, the method has been renamed to `recent_workspaces_for_ui`.

Meanwhile, `last_session_workspace_locations` used a slightly different form of the same predicate. The two disagreeing on what counts as a valid workspace caused #48799, `Workspace WorkspaceId(N) not found` errors on repeated launches (#50409), the `test_window_edit_state_restoring_enabled` flake (#50871), and the foreign key constraint fail on `projects: open recent` with a dirty scratch buffer (#51456). 

Note that for the last issue mentioned (#51456) there is no save prompt for scratch buffers. This seems out of scope for this PR so I'll fix that after this is addressed. 

Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [ ] Unsafe blocks (if any) have justifying comments
- [ ] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #48799 
Closes #50409 
Closes #50871
Closes #51456 

Release Notes:

- Fixed unsaved scratch buffers being lost across restarts and an occasional error when opening a recent project.

PR #54447, opened 2026-04-21
workspace: Prompt to save dirty untitled buffers before swapping projects

transitoryangel:
Swapping the active project (ReplaceWindow, open_workspace_for_paths, MultiWorkspace::open_project, and open_remote_project) previously let dirty untitled buffers get silently serialized by hot-exit or left behind in the outgoing workspace. Follow-up for #54224.

Self-Review Checklist:

Release Notes:

- Fixed unsaved changes in scratch buffers being silently discarded when switching to a different project.

### DIFF
--- a/crates/recent_projects/src/remote_connections.rs
+++ b/crates/recent_projects/src/remote_connections.rs
@@ -202,6 +202,18 @@ pub async fn open_remote_project(
         );
     }
 
+    if let Some(requesting_window) = open_options.requesting_window {
+        let prompt_task = requesting_window.update(cx, |multi_workspace, window, cx| {
+            let workspace = multi_workspace.workspace().clone();
+            workspace.update(cx, |workspace, cx| {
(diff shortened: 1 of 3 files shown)


dinocosta (commented):
Hey @transitoryangel ! Thank you for suggesting these changes 🙂

It looks like #55889 has since updated when Zed prompts about dirty unsaved buffers. In particular, Zed now prompts when replacing an empty workspace (one without a project) because although the buffer can be serialized, there would be no way to navigate back to that workspace and recover it through the UI.

In other cases, such as switching away from a workspace associated with a project, the workspace and its serialized buffers remain recoverable by reopening the project. With multi-workspace enabled, the outgoing workspace may also remain open in the same window. Prompting in these cases would therefore interrupt the user without providing additional data safety.

Given that, I’m going to close this as superseded by #55889. Please let me know if there’s a scenario I’ve overlooked where the outgoing workspace or its buffers would no longer be recoverable, and I’d be happy to revisit it. Thanks!

PR #54449, opened 2026-04-21
git: Revert skipping of events for the `.git` directory itself (#54443) (cherry-pick to stable)

cole-miller:
This reverts #54329 and the part of #52499 that was an earlier attempt at the same thing, which caused us to incorrectly miss git state updates on Windows. cc @Veykril it seems like we need to find a different way to fix the problem of `.git` scanning cycles.

Self-Review Checklist:

Release Notes:

- Fixed a bug causing stale git state on Windows.


bennetbo:
@zed-industries/approved

PR #54453, opened 2026-04-21
ep: Remove code duplication and extend prompt size limit

osyvokon:
V0327 requires a larger prompt limit.

Release Notes:

- N/A


probably-neb:
@zed-industries/approved

PR #54454, opened 2026-04-21
telemetry: Add latency metrics

katie-z-geer:
Self-Review Checklist:

Closes #ISSUE

Release Notes:

### DIFF
--- a/crates/input_latency_ui/src/input_latency_ui.rs
+++ b/crates/input_latency_ui/src/input_latency_ui.rs
@@ -110,7 +110,10 @@ pub fn report_input_latency_telemetry(window: &Window, cx: &mut App) {
     let frames_33to100 = count_frames_in_range(&delta_latency, MS33_NS, MS100_NS);
     // frames > 100 ms are implicitly total_frames - (sub4 + 4to8 + 8to16 + 16to33 + 33to100)
 
-    let events_per_frame_p99 = delta_coalesce.value_at_quantile(0.99);
+    let frames_with_1_event = count_frames_in_range(&delta_coalesce, 1, 2);
+    let frames_with_2_events = count_frames_in_range(&delta_coalesce, 2, 3);
+    let frames_with_3_events = count_frames_in_range(&delta_coalesce, 3, 4);
+    // frames with 4+ events are implicitly total_frames - (1 + 2 + 3)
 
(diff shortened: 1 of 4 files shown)


crates/input_latency_ui/src/input_latency_ui.rs
```diff
@@ -32,6 +34,103 @@ struct ReporterState {
...
+
+    telemetry::event!(
+        "Latency Report",
+        frames_sub4 = frames_sub4,
+        frames_4to8 = frames_4to8,
+        frames_8to16 = frames_8to16,
+        frames_16to33 = frames_16to33,
+        frames_33to100 = frames_33to100,
+        total_frames = total_frames,
+        events_per_frame_p99 = events_per_frame_p99,
```
eholk: It looks like this is just logging the P99 `events_per_frame`. Is this enough or do we want the whole histogram?

crates/zed/src/zed.rs
```diff
@@ -410,6 +410,23 @@ pub fn initialize_workspace(app_state: Arc<AppState>, cx: &mut App) {
             .detach();
         }
 
+        #[cfg(feature = "input-latency-histogram")]
```
eholk: I don't think we want this `cfg(feature)` here... I'll remove it.

crates/input_latency_ui/src/input_latency_ui.rs
eholk: Okay, I just updated this to report 1, 2, and 3 event buckets, and then we can subtract those from the total to get 4+.

```
Events coalesced per frame:
  min  :      1 events
  p50  :      1 events
  p75  :      1 events
  p90  :      2 events
  p95  :      2 events
  p99  :      2 events
  p99.9:      3 events
  max  :     10 events

Distribution:
       1 events:  20349 ( 79.5%) ███████████████████████
       2 events:   5183 ( 20.3%) ██████
       3 events:     59 (  0.2%) 
      10 events:      1 (  0.0%) 
```

That's what my current histogram looks like, so 1, 2, 3, 4+ should give us enough data. Basically, if we start seeing too many things in 4+ then we can add more resolution or think about our strategy here in general.

eholk (approved):

PR #54461, opened 2026-04-21
compliance: Account for Zippy commits without PR in report summary

MrSubidubi:
The report summary previously assumed that every commit that had a review would also have an associated PR - this was no longer the case with Zippy PRs that are allowed without PR. This pull request fixes this wrong assumption.

Release Notes:

- N/A


ConradIrwin (approved):
:+1:

PR #54469, opened 2026-04-22
sidebar: Improve scrollbar overlap with thread item icon buttons

danilo-leal:
This PR removes the overlap between the sidebar scrollbar and the thread item icon buttons. Now, they don't render on top of each other anymore :)

Release Notes:

- Agent: Improved the scrollbar overlap with the thread item icon buttons in the threads sidebar.


bennetbo:
@zed-industries/approved

gianpaj:
Sorry for this comment, but shouldn't UI changes be better understood with screenshots?

PR #54471, opened 2026-04-22
sidebar: Add scrollbar and max-height to the thread switcher

danilo-leal:
Just a little refinement as I experienced the thread switcher modal overflowing beyond the viewport if there were enough threads on it. Additionally, took the opportunity to render a scrollbar inside of it, too.

Release Notes:

- N/A


bennetbo:
@zed-industries/approved

PR #54478, opened 2026-04-22
Cherry pick more sidebar improvements to v0.233.x

maxbrunsfeld:
* #54005
* #54469 
* #54471

Release Notes:

- Agent: Improved the scrollbar overlap with the thread item icon buttons in the threads sidebar.

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -349,7 +349,7 @@ pub fn init(cx: &mut App) {
                                 auto_submit: true,
                             }),
                             true,
-                            "agent_panel",
+                            "git_panel",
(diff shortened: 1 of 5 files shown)


crates/agent_ui/src/agent_panel.rs
```diff
@@ -349,7 +349,7 @@ pub fn init(cx: &mut App) {
                                 auto_submit: true,
                             }),
                             true,
-                            "agent_panel",
+                            "git_panel",
```
Anthony-Eid, line 352: Is this right? The source aka `panel` is the agent panel here not the git panel. I'm not sure if I understand the intent

Anthony-Eid (approved):

PR #54485, opened 2026-04-22
ci: Add workflow for bumping Zed versions

MrSubidubi:
Release Notes:

- N/A


script/get-released-version
```diff
@@ -18,4 +19,4 @@ case $channel in
     ;;
 esac
 
-curl -s "https://cloud.zed.dev/releases/$channel/latest/asset?asset=zed&os=macos&arch=aarch64" | jq -r .version
+curl -s "https://cloud.zed.dev/releases/$channel/$version/asset?asset=zed&os=macos&arch=aarch64" | jq -r .version
```
ConradIrwin: Does this work? I would have thought that this will just echo $version right back at you.

MrSubidubi: It does actually work - e.g., you have 
https://cloud.zed.dev/releases/stable/latest/asset?asset=zed&os=macos&arch=aarch64
https://cloud.zed.dev/releases/preview/0.233.2/asset?asset=zed&os=macos&arch=aarch64

but
https://cloud.zed.dev/releases/stable/0.235.0/asset?asset=zed&os=macos&arch=aarch64

will throw you a 404. Checked that earlier

ConradIrwin: the error you'll get is `jq: parse error: Invalid numeric literal at line 1, column 4`

I think it'd be clearer to leave this script untouched and call `curl --fail` if you want to detect if a release exists.

ConradIrwin (approved):
Thanks. I think the overall shape makes sense.

Just so I understand, the expectation for internal users is

To release a new patch version on the current release branches:

* ./script/trigger-release [stable|preview|nightly] 

To kick off a Wednesday release cycle:

* ./script/bump-zed-version

To reset the current release to the latest version on that branch if it failed:

* ./script/retag-release [stable|preview]

I think we can get rid of the things we don't need:
* bump-zed-version shouldn't take an argument that lets you bump major versions
* bump-zed-patch-version shouldn't exist (use trigger-release for that)

MrSubidubi:
> Just so I understand, the expectation for internal users is
> 
> To release a new patch version on the current release branches:
> 
>     ./script/trigger-release [stable|preview|nightly]
> 
> To kick off a Wednesday release cycle:
> 
>     ./script/bump-zed-version
> 
> To reset the current release to the latest version on that branch if it failed:
> 
>     ./script/retag-release [stable|preview]
> 

Yes, precisely

PR #54496, opened 2026-04-22
settings: Add auto completion to command aliases setting

dinocosta:
Update the JSON schema generated for the settings file in order to be able to provide the list of valid actions when editing the values for the `command_aliases` setting.

While reviewing https://github.com/zed-industries/zed/pull/52892 , I noticed that, even though we already have support for this in the keymap file, we don't support it for the `command_aliases` setting, so went ahead and refactored this a bit such that the existing functionality for the keymap file JSON schema could also be re-used for the `command_aliases` setting.

Here's a quick big-picture breakdown of the relevant changes:

* Add `settings_content::ActionName` newtype, representing a simple named action without arguments. The
  `settings_content::ActionName::build_schema` function can be used to build the schema of all possible action names.
* Add `settings_content::ActionWithArguments` newtype, representing an action with arguments. This was mostly done so as to keep both action without arguments and action with arguments newtypes together,
  even though we don't have `settings_content::ActionWithArguments::build_schema`, as it is only
  used by the keymap schema generation logic and probably doesn't warrant moving it here right now.
* Update both `settings_content::WorkspaceSettingsContent::command_aliases` and `workspace::workspace_settings::WorkspaceSettings::command_aliases` to now be of type `HashMap<String, ActionName>` such that, when the json schema for `command_aliases` is generate, it'll now reference the `#/$defs/ActionName` schema.
* Update `SettingsStore::json_schema` so as to populate the `#/$defs/ActionName` schema at runtime, replacing it with the actual list of valid action names.

Self-Review Checklist:

Release Notes:

- Added support for auto-completing action names on `command_aliases` setting

### DIFF
--- a/crates/command_palette/src/command_palette.rs
+++ b/crates/command_palette/src/command_palette.rs
@@ -442,7 +442,7 @@ impl PickerDelegate for CommandPaletteDelegate {
     ) -> gpui::Task<()> {
         let settings = WorkspaceSettings::get_global(cx);
         if let Some(alias) = settings.command_aliases.get(&query) {
-            query = alias.to_string();
+            query = alias.as_ref().to_owned();
         }
 
         let workspace = self.workspace.clone();
--- a/crates/docs_preprocessor/src/main.rs
+++ b/crates/docs_preprocessor/src/main.rs
@@ -3,7 +3,7 @@ use mdbook::BookItem;
 use mdbook::book::{Book, Chapter};
 use mdbook::preprocess::CmdPreprocessor;
 use regex::Regex;
-use settings::{KeymapFile, SettingsStore};
+use settings::{KeymapFile, SettingsJsonSchemaParams, SettingsStore};
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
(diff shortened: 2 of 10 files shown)


probably-neb (approved):

JonGretar:
This made my existing, vim butter finger, aliases give out a warning.

```JSONC
"command_aliases": {
    "W": "w",
    "Wq": "wq",
    "WQ": "wq",
    "Q": "q",
  },
  ```
  
  It's very common for non english keyboard users to have to add these aliases.

dinocosta:
> This made my existing, vim butter finger, aliases give out a warning.
> 
> ```json-comments
> "command_aliases": {
>     "W": "w",
>     "Wq": "wq",
>     "WQ": "wq",
>     "Q": "q",
>   },
> ```
> 
> It's very common for non english keyboard users to have to add these aliases.

Thank you for sharing! Completely forgot that you can have it set to any string and not necessarily to a specific action. Will look into updating the JSON schema so as not complain if a string is provided, that way we can still provide autocompletion without displaying errors in these cases 🙂

PR #54504, opened 2026-04-22
Revert "markdown: Apply theme syntax color to inline code in rendered markdown"

bennetbo:
Reverts zed-industries/zed#54475

This causes inline code blocks to look strange in some themes

image

Release Notes:

- N/A


danilo-leal:
@zed-industries/approved

PR #36713, opened 2025-08-21
Allow user-defined worktree names in title bar and platform windows

Warpten:
Closes #36637 

Release Notes:
 - Adds the ability  to specify a human-readable project name for each worktree.

https://github.com/user-attachments/assets/ce980fa6-65cf-46d7-9343-d08c800914fd

PR #54511, opened 2026-04-22
settings: Remove the `project_name` project setting

tomhoule:
The `project_name` worktree setting was added in #36713 to let users override the name shown in the window title. Its description ("The displayed name of this project. If left empty, the root directory name will be displayed.") suggests broader coverage, and #46440 reports the reasonable expectation that it should also apply in the project switcher. In practice the setting has only ever affected `Workspace::update_window_title`, so everywhere else (recent projects, the multi-worktree pane, ...) keeps falling back to the worktree root name.

Rather than plumb the setting through each of those surfaces, I'm removing it. Having a project-level setting control how your editor displays the project has downsides. For example it means a checkout can dictate UI in someone else's Zed. The natural home for a custom display name is the workspace DB, set from the UI, which is what we should do if we want this feature back.

If you want this back, the path forward is to store the display name in `WorkspaceDb`, expose a UI affordance to edit it, and read it from `update_window_title`, `recent_projects::get_recent_projects` / `get_open_folders`, and any other places that currently derive a display name from the worktree root.

Closes #46440

Release Notes:

- Removed the `project_name` project setting. It only ever affected the OS window title, and the expectation that it would show up in the project switcher and elsewhere is better served by a future UI-driven, per-workspace setting stored locally.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1,8 +1,5 @@
 {
   "$schema": "zed://schemas/settings",
-  /// The displayed name of this project. If not set or null, the root directory name
-  /// will be displayed.
-  "project_name": null,
   // The name of the Zed theme to use for the UI.
   //
   // `mode` is one of:
--- a/crates/settings/src/vscode_import.rs
+++ b/crates/settings/src/vscode_import.rs
@@ -1036,7 +1036,6 @@ impl VsCodeSettings {
 
     fn worktree_settings_content(&self) -> WorktreeSettingsContent {
         WorktreeSettingsContent {
-            project_name: None,
             prevent_sharing_in_public_channels: false,
             file_scan_exclusions: self
                 .read_value("files.watcherExclude")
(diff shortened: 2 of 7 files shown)


probably-neb (approved):

chances:
I for one liked this feature; it helped me distinguish Zed windows.

Hopefully this comes back with better support across the app.

mashpie:
Miss that setting. Without it all our recent projects and window titles are called "monorepo" or similar (from repo structure `<customer>/<project>`). It was the counter part of vscodes `"window.title"` and convinced the last of us to use zed instead of vs :(

Warpten:
I initially tried to make it work with workspace settings but it got confusing really fast especially if you start loading multiple root folders. I'll try to take another stab at this because I **need** this, but could use some more pointers for some of the finer stuff (also, the code changed quite a lot since I last looked at it)

PR #54512, opened 2026-04-22
Add BlurRect and LensRect rendering primitives

df49b9cd:
## Summary

Adds two new GPUI scene primitives for per-element backdrop compositing, enabling Liquid Glass and frosted-glass UI effects without waiting for `WindowBackgroundAppearance::Blurred`:

- **`Primitive::BlurRect`** — dual-Kawase (Marius Bjorge, SIGGRAPH 2015) backdrop blur with configurable kernel levels, per-corner radii, and tint overlay.
- **`Primitive::LensRect`** — Liquid Glass composite: backdrop blur plus parabolic UV refraction, chromatic aberration, and a directional Fresnel edge glow.

Both primitives are wired through the full rendering pipeline on every backend:

| Backend | Approach |
|---------|----------|
| **WGPU** | `copy_texture_to_texture` snapshot → `blur_io_shaders.wgsl` downsample/upsample → `fs_blur_rect`/`fs_lens_rect` compositing |
| **Metal** | `MTLBlitCommandEncoder::copy_from_texture` → MSL port of both shader modules → cbindgen input-index enums |
| **DirectX** | Exhaustive-match stub returning `Ok(())` (future implementation) |

### Public API (`Window`)

```rust
window.paint_blur_rect(bounds, corner_radii, BlurEffect { radius, kernel_levels, tint });
window.paint_lens_rect(bounds, corner_radii, LensEffect { radius, refraction, depth, dispersion, splay, light_angle_radians, light_intensity, tint });
```

Each call forces a render-pass break so the framebuffer can be sampled, then re-opens the pass with `LoadOp::Load` for subsequent primitives.

### Design notes

- Shaders live in a separate `blur_io_shaders.wgsl` module (not mixed into the monolithic `shaders.wgsl`) for clarity.
- Metal and WGPU backends follow the same structural pattern (snapshot → downsample → upsample → composite) for easy cross-platform parity verification.
- `set_framebuffer_only(false)` is set on the Metal layer to allow blit-reading the drawable.
- A `surface_supports_copy_src` flag gracefully degrades on WGPU platforms that don't support `COPY_SRC` usage on the swapchain texture.

## Test plan

- [x] WGPU pipeline-creation smoke test (`pipeline_creation_smoke`) — compiles all pipelines including the four new ones.
- [x] Scene round-trip tests (`blur_rect_round_trip`, `lens_rect_round_trip`) — push primitive → verify it reaches `Scene::blur_rects`/`lens_rects` via `Scene::batches()`.
- [ ] Manual: `cargo run -p tahoe-gpui --example liquid_glass_gallery` — visual diff against Figma Tahoe UI Kit.
- [ ] Manual: `cargo run -p tahoe-gpui --example liquid_glass_interactive` — sweep Refraction / Depth / Dispersion / Light-angle sliders.
- [ ] Manual: resize window while Liquid Glass elements visible — no artifacts.
- [ ] Manual: toggle Reduce Transparency — opaque fallback renders correctly.

## Known follow-ups

- `kernel_levels` and `blur_radius` are stored in the primitives but not yet consumed by the dual-Kawase passes (hardcoded `offset_multiplier: 1.0`). A follow-up will wire these to control the number of down/up iterations and the Kawase spread.
- Framebuffer snapshot runs separately for `BlurRects` and `LensRects` batches; when both exist in the same frame the snapshot is duplicated. A single snapshot before the first blur/lens batch would eliminate the redundancy.
- No Metal pipeline smoke test yet (Metal requires a live `MTLDevice` which is unavailable in CI).

Release Notes:

- Added `Primitive::BlurRect` and `Primitive::LensRect` scene primitives for per-element backdrop blur and Liquid Glass compositing, with full WGPU and Metal backend support.


df49b9cd:
Closing — moving this PR to our fork at df49b9cd/zed instead.

PR #54513, opened 2026-04-22
Continue project search until worktree scan is complete

### LINKED ISSUE #9858: project search doesn't work correctly while the file index isn't finished
### Check for existing issues

- [X] Completed

When opening Zed, it appears the a file index with the files of the project is created. For big repos (e.g. Chromium), this index takes a while to build. During that time, project search seems to be limited to the files already indexed.

### Environment

Zed: v0.127.5 (Zed)
OS: macOS 14.4.0
Memory: 32 GiB
Architecture: aarch64

_No response_

_No response_

osyvokon:
Previously, when starting a search on a project which is in the middle of the scan, we would miss the results in the files that are not yet scanned.

This change makes two stage search: start searching on whatever files are already scanned (just as we do now), but then wait for the scan to finish and complete the search.

Self-Review Checklist:

- [ ] Tests cover the new/changed behavior (I failed to write a test that is both deterministic and compact; tested manually)

Closes #9858

Release Notes:

- Fixed incomplete search results when the project scan is incomplete

### DIFF
--- a/crates/project/src/project_search.rs
+++ b/crates/project/src/project_search.rs
@@ -1,5 +1,6 @@
 use std::{
     cell::LazyCell,
+    cmp::Ordering,
     collections::BTreeSet,
     io::{BufRead, BufReader},
     ops::Range,
@@ -418,9 +419,13 @@ impl Search {
                 let gitignored_tracker = PathInclusionMatcher::new(query.clone());
                 let include_ignored = query.include_ignored();
                 for worktree in worktrees {
-                    let (mut snapshot, worktree_settings) = worktree
+                    let (mut snapshot, worktree_settings, snapshot_scan_incomplete) = worktree
                         .read_with(cx, |this, _| {
-                            Some((this.snapshot(), this.as_local()?.settings()))
+                            Some((
+                                this.snapshot(),
+                                this.as_local()?.settings(),
+                                this.completed_scan_id() < this.scan_id(),
+                            ))
                         })
                         .context("The worktree is not local")?;
                     if query.include_ignored() {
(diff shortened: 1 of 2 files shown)


crates/search/src/project_search.rs
```diff
@@ -372,6 +372,26 @@ impl ProjectSearch {
         cx.notify();
     }
 
+    fn insert_match_ranges(&mut self, mut new_ranges: Vec<Range<Anchor>>, cx: &App) {
```
osiewicz, line 375: I don't love that we're losing the invariant here (that search results are sorted with regards to path).

crates/project/src/project_search.rs
```diff
@@ -418,9 +419,13 @@ impl Search {
...
                 let include_ignored = query.include_ignored();
                 for worktree in worktrees {
-                    let (mut snapshot, worktree_settings) = worktree
+                    let (mut snapshot, worktree_settings, snapshot_scan_incomplete) = worktree
                         .read_with(cx, |this, _| {
-                            Some((this.snapshot(), this.as_local()?.settings()))
+                            Some((
+                                this.snapshot(),
+                                this.as_local()?.settings(),
+                                this.completed_scan_id() < this.scan_id(),
```
osiewicz, line 427: I believe we're conflating initial scan with any scan with this check. If there's an ongoing search and an ongoing non-initial scan, we'll defer the search until that worktree scan is complete. That feels quite brittle in face of the fact that worktree scans are non-deterministic. 

Put another way: with this code as is, we won't wait beyond one scan. Thus it's still possible for search to miss stuff I believe?

crates/search/src/project_search.rs
osiewicz, line 375: As is, this PR will make project search wait on any ongoing scan, not just initial scan. Thus, you should handle a case where results are removed too; or, ensure it's only holding search back up for initial scan, not any subsequent ones.

osiewicz (changes requested):
This may be a bit unpopular, but.. how about holding a search off until the worktree is ready?

Our current situation (project search potentially missing files) is bad. I'm not sure if this PR would be much better; from users POV there's literally no way to tell that a project search has an obstacle in it's way. I think what we could do instead is to show a "helpful" message up until the worktree scan of all worktrees is complete. Something along the lines of "Please wait patiently while we try to figure out what's on disk". I get that it's not ideal, but from the perspective of throughput it should be only slightly slower (and may even be faster); I'm also pretty positive that the implementation could be vastly simplified that way. What do you think?

crates/project/src/project_search.rs
```diff
@@ -449,37 +454,123 @@ impl Search {
...
+            let maybe_wait_for_snapshot = worktree.update(cx, |this, _| {
+                let scan_id = this.scan_id();
+                (this.completed_scan_id() < scan_id).then(|| this.wait_for_snapshot(scan_id))
+            });
+            let Some(wait_for_snapshot) = maybe_wait_for_snapshot else {
+                return Ok(worktree.read_with(cx, |this, _| this.snapshot()));
+            };
+            wait_for_snapshot.await?;
+        }
+    }
```
osiewicz, line 514: I believe this code will only be called from local callsites (where you're working with LocalSnapshots); you can use `LocalWorktree::scan_complete` and just await the returned future.

osyvokon:
>  from users POV there's literally no way to tell that a project search has an obstacle in it's way. 

Well, in the current implementation there will be an arrow circle spinning until we the scan is complete

osiewicz:
Right, but that's associated with the project search just being underway. You can't tell that it's still spinning (and, in fact, not making any progress on the search itself) because of the worktree scan not being complete.

osyvokon:
>  show a "helpful" message up until the worktree scan of all worktrees is complete.

Well, it would simplify code for sure, but it also makes UX worse. Like, in many searches, I only care about the first 1-2 matched files, and I don't wait for the full scan to complete. 

This is especially true given that our scans are very slow for some reason. But even if we fix that, there will always be cases like slow filesystems, network mounts, etc.

So I'd give this solution another round before reverting to a simpler one.

crates/search/src/project_search.rs
osyvokon: We're not losing it. When "late" results arrive, we sort the list, which sorts by worktree and path first

osyvokon:
On the other hand, this PR doesn't make true incremental search updates, so it doesn't fully help with slow scans either 🤔

osyvokon:
Closing in favor of #54534

PR #54519, opened 2026-04-22
command_palette: Fix showing keybinding in footer actions on zero matches

galuis116:
Self-Review Checklist:

Closes #54517

Release Notes:

- Fixed Command Palette behavior where footer actions could still route to a fallback hidden command when search returned no matches.


ConradIrwin (approved):
Neat. Seems like more sensitible behavior for selected_command.

I don't know that you need the check in confirm as well, but doesn't seem problematic.

joaovictor712:
I think there is no need to display footer hints such as ""Change Keybinding... Ctrl-Enter, Run Enter as there is No matches, so this pull request is fair.
image

PR #54525, opened 2026-04-22
compliance: Catch all Zippy version bumps on stable

MrSubidubi:
I checked the temporary fix against the 0.233.2 bump - however, some other guy fixed (and thus changed) the commit authors from 0.233.3. Or, to put it differently:

image

Anyway, this fixes the temporary fix and also cherry-picks the singular checking support to stable so that should something else come up, we can look at that.

Release Notes:

- N/A


ConradIrwin (approved):
thanks

PR #52997, opened 2026-04-02
acp: Use new Rust SDK

benbrandt:
Testing out Niko's new SDK design

Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Release Notes:

- N/A

PR #54526, opened 2026-04-22
acp_tools: Buffer recent ACP messages so the log pane can show history

eholk:
Today the ACP log pane (`dev::OpenAcpLogs`) subscribes to a connection's live message stream on open, so if you open it after noticing something went wrong, the traffic that led up to the problem is already gone.

This adds a bounded ring buffer (`MessageLog`) on each `AcpConnection` that captures every ACP stream message from the moment the connection is created, independent of whether any UI is observing. When the log pane opens, it pre-populates from the buffer and then tails live updates via a `watch` channel.

## How it works

- `MessageLog` holds a `VecDeque<acp::StreamMessage>` capped at 500 entries, plus a monotonic `total_count` and a `watch::Sender<u64>` used purely as a wakeup signal.
- `AcpConnection::stdio` creates the log immediately after constructing the `ClientSideConnection` and spawns a foreground task that drains `connection.subscribe()` into it — crucially, *before* `initialize`, so the init handshake is captured too.
- `AcpConnectionRegistry` now holds a `Weak<MessageLog>` instead of a `Weak<ClientSideConnection>`; the log's identity tracks the connection 1:1.
- `AcpTools::update_connection` snapshots the log on open, then subscribes to the watch channel and calls `read_since(last_seen_total)` on each notification. The monotonic counter makes the tail idempotent regardless of watch coalescing.

## Known limitations

- If the ring rolls (>500 messages) while the pane is open, those messages silently disappear from the view. `MessageLogSnapshot` carries a `skipped` field so we know when this happens, but nothing surfaces it yet.
- Each live `AcpConnection` now holds up to 500 buffered messages for its lifetime (previously the UI-side buffer only existed while the pane was open). In practice this is ~5–20MB for a busy session, dominated by JSON param payloads.
- When an agent connection is restarted, conversations stay bound to the old connection while the registry points at the new one — so the log pane ends up watching an idle new connection. The ring buffer itself works correctly per-connection, but the pane's 'active connection' concept needs to grow to handle this. Deferred.

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,6 +63,7 @@ dependencies = [
  "theme_settings",
  "ui",
  "util",
+ "watch",
  "workspace",
 ]
 
--- a/crates/acp_tools/Cargo.toml
+++ b/crates/acp_tools/Cargo.toml
@@ -26,4 +26,5 @@ settings.workspace = true
 theme_settings.workspace = true
 ui.workspace = true
 util.workspace = true
+watch.workspace = true
 workspace.workspace = true
--- a/crates/acp_tools/src/acp_tools.rs
+++ b/crates/acp_tools/src/acp_tools.rs
@@ -1,6 +1,6 @@
 use std::{
-    cell::RefCell,
-    collections::HashSet,
+    cell::{Cell, RefCell},
+    collections::{HashSet, VecDeque},
     fmt::Display,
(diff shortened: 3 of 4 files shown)


eholk:
Closing in favor of working directly on top of the new ACP SDK architecture that landed in #52997.

PR #54534, opened 2026-04-22
Wait for worktree scan to complete before starting search

### LINKED ISSUE #9858: project search doesn't work correctly while the file index isn't finished

osyvokon:
When starting a search on a project that is in the middle of a scan, we used to miss results in files that had not yet been scanned.

This change makes the search wait for the scan to complete.

Closes #9858

Release Notes:

- Fixed incomplete search results when the project scan is incomplete

### DIFF
--- a/crates/project/tests/integration/project_tests.rs
+++ b/crates/project/tests/integration/project_tests.rs
@@ -12280,8 +12280,7 @@ async fn search(
             SearchResult::Buffer { buffer, ranges } => {
                 results.entry(buffer).or_insert(ranges);
             }
-            SearchResult::LimitReached => {}
-            SearchResult::WaitingForScan => {}
+            SearchResult::LimitReached | SearchResult::WaitingForScan => {}
         }
     }
     Ok(results
--- a/crates/project/tests/integration/project_tests.rs
+++ b/crates/project/tests/integration/project_tests.rs
@@ -12281,6 +12281,7 @@ async fn search(
                 results.entry(buffer).or_insert(ranges);
             }
             SearchResult::LimitReached => {}
+            SearchResult::WaitingForScan => {}
         }
     }
     Ok(results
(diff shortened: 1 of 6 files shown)


crates/project/tests/integration/project_tests.rs
```diff
@@ -12281,6 +12281,7 @@ async fn search(
                 results.entry(buffer).or_insert(ranges);
             }
             SearchResult::LimitReached => {}
+            SearchResult::WaitingForScan => {}
```
osiewicz:
```suggestion
SearchResult::LimitReached | SearchResult::WaitingForScan => {}
```

crates/search/src/project_search.rs
```diff
@@ -516,8 +533,11 @@ impl Render for ProjectSearchView {
             let model = self.entity.read(cx);
             let has_no_results = model.no_results.unwrap_or(false);
             let is_search_underway = model.pending_search.is_some();
+            let is_waiting_for_scan = model.waiting_for_scan;
 
-            let heading_text = if is_search_underway {
+            let heading_text = if is_waiting_for_scan {
+                "Waiting for project scan to complete…"
```
osiewicz: How about something slightly more generic, like "Loading project..."?

osiewicz (approved):
LGTM, thanks!

hferreiro:
IIUC this will make searching for files in big repos even slower. Are there plans for doing the search in parallel to the scanning?

osyvokon:
> IIUC this will make searching for files in big repos even slower. 

This change doesn't make the overall search slower, but it may increase the time to first result -- under specific conditions. Like, for example, in this scenario:

1. Open Chromium in Zed
2. Start searching _immediately_
3. Observe about 10 seconds delay before the search starts.

However, it's not as bad as it sounds because:

- Subsequent searches won't have this delay
- If you don't search as the first thing after opening the editor, you won't have this delay either.


> Are there plans for doing the search in parallel to the scanning?

Not at the moment, but if you see a good motivational case for it, please open an issue.

Issue #54531, opened 2026-04-22
ACP robustness and debugging


eholk:
I've been noticing issues with ACP threads getting stuck and not always handling failures as gracefully as we could. I want to look into this for quality week and also make sure we have enough debugging tools to be able to diagnose and fix these issues.

PR #54536, opened 2026-04-22
acp_tools: Always capture ACP transport and stderr into the log ring

eholk:
Part of #54531.

Previously the log tap started disabled and only flipped on when someone opened the ACP logs panel for the first time on a given connection. That meant the common *'I noticed something is wrong with my agent, let me go look at the logs'* workflow started with an empty pane \u2014 all the traffic that led up to the problem was already gone.

This removes the `enabled` flag on `AcpLogTap` so capture is unconditional. The backlog ring is still bounded at `MAX_BACKLOG_MESSAGES` (2000), so memory is capped. The steady-state cost is one channel send per transport/stderr line, which is negligible compared to the JSON-RPC serialization that already happened to produce the line.

Net diff is ~40 lines deleted from `acp_tools.rs` \u2014 the flag, its accessors, the early-return in `emit`, and the `active_tap` bookkeeping needed to enable it on first subscribe.

I confirmed with the author of the original opt-in design before making this change.

Release Notes:

- N/A

### DIFF
--- a/crates/acp_tools/src/acp_tools.rs
+++ b/crates/acp_tools/src/acp_tools.rs
@@ -1,10 +1,7 @@
 use std::{
     collections::{HashSet, VecDeque},
     fmt::Display,
-    sync::{
-        Arc,
-        atomic::{AtomicBool, Ordering},
-    },
+    sync::Arc,
 };
 
 use agent_client_protocol::schema as acp;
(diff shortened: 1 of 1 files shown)


benbrandt (approved):
Let's do it!

PR #54560, opened 2026-04-22
agent_ui: Add partial tab completion in agent panel when mentioning files/directories

AmaanBilwar:
Self-Review Checklist:

Closes #37761
 
Release Notes:
- agent: When @mentioning files allow partial completion when hitting `tab`


dinocosta (changes requested):
Hey @AmaanBilwar ! Thank you for taking a look at this one 🙌

I've confirmed that these changes are working on macOS but I suspect we want to explore a different approach. 

* I believe we'll want `compose_new_text` to `Completion` instead of leaving in `CompletionSource::Custom`, as the `CompletionSource` is meant to represent the place from where the completion was built from, along with any data originated by the source, and this new field feels related to the completion's edit, so we can put it next to `new_text. Also, not sure if `new_text_for_compose` would read better. What do you think?
* Not sure if you have a Windows machine, but was wondering whether the changes in `PromptCompletionProvider::completion_for_path` would fail on Windows, seeing as we're matching on `/` as the path separator, while Windows uses `\` so it might need to use something agnostic in that case.

Lastly, having a test for this change would be sweet! In case you're having a hard time writing one, let me know, and I can try to help. Thanks! 🙂

AmaanBilwar:
def agree with the `compose_new_text` to `Completion` approach, makes much more sense. 
i have a linux dual boot machine with windows so i can test both out, its been a while since ive worked on this so i'd need some time to go through this and appreciate any help i can get. 

thank you :)) @dinocosta

dinocosta:
> def agree with the `compose_new_text` to `Completion` approach, makes much more sense. i have a linux dual boot machine with windows so i can test both out, its been a while since ive worked on this so i'd need some time to go through this and appreciate any help i can get.
> 
> thank you :)) @dinocosta

All good! Take your time and let me know if you need any help! ✌️

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been inactive for quite some time and has a record number of conflicts now 

image

PR #54583, opened 2026-04-23
gpui: Avoid redundant SharedString allocations in ShapedLine::split_at

cppcoffee:
Build split text directly from Arc<str> slices instead of allocating intermediate Strings. Reuse the original SharedString for full-line splits, return a static empty string for empty sides, and add a regression test covering boundary splits.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/text_system/line.rs
+++ b/crates/gpui/src/text_system/line.rs
@@ -64,6 +64,18 @@ impl ShapedLine {
         self.layout.width
     }
 
+    fn text_slice(&self, start: usize, end: usize) -> SharedString {
+        if start == end {
+            return SharedString::new_static("");
+        }
+
+        if start == 0 && end == self.text.len() {
+            return self.text.clone();
+        }
+
+        SharedString::new(Arc::<str>::from(&self.text[start..end]))
+    }
+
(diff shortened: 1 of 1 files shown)


crates/gpui/src/text_system/line.rs
```diff
@@ -208,8 +220,8 @@ impl ShapedLine {
         }
 
         // Split text
-        let left_text = SharedString::new(self.text[..byte_index].to_string());
-        let right_text = SharedString::new(self.text[byte_index..].to_string());
+        let left_text = self.text_slice(0, byte_index);
+        let right_text = self.text_slice(byte_index, self.text.len());
```
Veykril:
```suggestion
        let left_text = if byte_index == self.text.len() {
            self.text.clone()
        } else {
            SharedString::new(&self.text[..byte_index])
        };
        let right_text = if byte_index == 0 {
            self.text.clone()
        } else {
            SharedString::new(&self.text[byte_index..])
        };

```
is all we really need to change here to make this slightly more efficient. No need for that function or the test really.

Veykril (approved):

Veykril:
Thanks!

PR #54600, opened 2026-04-23
git_graph: Align CommitDataReader with batched commit processing

feeiyu:
Self-Review Checklist:

Updated `CommitDataReader::read` from single SHA to batched input/output.
`run_commit_data_reader` is designed to process requests in batches, but `CommitDataReader::read` previously only sent one SHA per call.   So in practice, commit data was effectively processed one-by-one from the caller path.

The temporary debug logs show that batching was not performed previously.

before:
image

after:
image

Release Notes:

- N/A


Anthony-Eid:
Thank you for finding this bug! it's 3:30am for me right now but I'm interested in getting this merged tomorrow 

Sorry I missed this in your original remote PR, I thought the batching was happening at a lower level and didn't realize reader.read(she).await was such a big foot gun here

Anthony-Eid (approved):

Anthony-Eid:
Just a heads up, I updated the local commit data reader to keep `CommitDataReader::read` as a single-SHA API and move batching/concurrency into `local_commit_data_reader`. This lets local commit data requests and completed reads be polled concurrently, avoiding blocking on a full batch while still allowing the underlying `cat-file --batch` reader to batch work internally.

Thank you once again for catching that the batch wasn't batching!

PR #54604, opened 2026-04-23
agent: Show MCP tool primary argument in tool-call header

neuodev:
## Problem

Built-in tools like `Fetch` render their primary argument next to the tool name in the collapsed tool-call header (e.g. `Fetch https://example.com`). MCP tools don't — they always show `Run MCP tool <name>` and the argument is hidden inside the expandable "Raw Input" JSON block.

For Playwright, fetch, search, or any MCP tool where one argument carries all the context, this makes every call opaque. You have to click to expand just to see which URL is being opened or which file is being read.

Before:

```
⊙ mcp__playwright__open_url_in_browser
  [click to expand]
  Raw Input:
  {
    "url": "https://linkedin.com/in/neuodev"
  }
```

After:

```
⊙ Run MCP tool `mcp__playwright__open_url_in_browser` https://linkedin.com/in/neuodev
```

## Root cause

`ContextServerTool::initial_title` at `crates/agent/src/tools/context_server_registry.rs` discards the input:

```rust
fn initial_title(&self, _input: serde_json::Value, _cx: &mut App) -> SharedString {
    format!("Run MCP tool `{}`", self.tool.name).into()
}
```

Built-in tools read the input (`fetch_tool.rs`: `format!("Fetch {}", MarkdownEscaped(&input.url))`) but the generic MCP path never did.

## Change

When the MCP tool's input is a JSON object with exactly one string-valued field, inline that value next to the tool name. All other shapes (multi-field objects, non-string values, null/empty input) fall back to the existing `Run MCP tool <name>` label, so behavior is unchanged for them.

- Value is truncated at 120 chars with `…` on overflow (char-safe, not byte-safe — "café" won't panic).
- Value is run through `MarkdownEscaped` so markdown metacharacters render literally.
- Single-string heuristic was chosen because the MCP spec has no standard field for a display template, and the common case (`url`, `path`, `query`, `selector`, `name`) is covered. I considered honoring `Tool.annotations.title` or `_meta["zed/title"]` as a template — happy to follow up with that if the team prefers an opt-in mechanism, but wanted to start with something that works out of the box for every existing MCP server.

## Tests

Added 10 unit tests in the existing `mod tests` block covering: happy path, no args, null input, multi-field fallback, non-string fallback, truncation, markdown escaping, multibyte chars, and empty-string preservation.

`cargo check -p agent --tests` passes locally. I couldn't run the full test binary on my machine because `wasmtime-c-api-impl` (a transitive test-dep) needs `cmake`, but the new code is pure string-manipulation with no wasm/runtime touch — CI should cover.

## Scope

One file, +127/-3. No behavior change for any tool shape other than single-string-input MCP tools.

## Per CONTRIBUTING.md

Happy to convert to a discussion first if the team prefers — didn't see a tracking issue for this. Also filed under CLA (signed).

Release Notes:

- Show MCP tool primary argument in tool header if space allows


neuodev:
@cla-bot check

ChristopherBiscardi (approved):
Thanks! I talked this over with other staff and we think this is worth it.

PR #54615, opened 2026-04-23
ci: Bump script check runner size

MrSubidubi:
Since we build the workflows xtask on this runner, increasing the size here to make that job faster.

Release Notes:

- N/A


SomeoneToIgnore:
@zed-industries/approved

PR #54635, opened 2026-04-23
Fix default.json profile and os settings not applying on startup

SomeoneToIgnore:
When started working on "disable the profiler for all but Nightly and Dev" users, I've tried to do that using the default OS settings and discovered that those are ignored.

Release Notes:

- Fixed default.json profile and os settings not applying on startup


SomeoneToIgnore:
Will merge optimistically as https://github.com/zed-industries/zed/pull/54645 got merged and both are needed to properly function for Nighlty and Dev builds.

MrSubidubi (commented):
@zed-industries approved

MrSubidubi:
@zed-zippy approved

PR #54638, opened 2026-04-23
agent_ui: Enable split diff view in agent diff

Veykril:
Release Notes:

- Enable split diff view in agent diff

### DIFF
--- a/crates/agent_ui/src/agent_diff.rs
+++ b/crates/agent_ui/src/agent_diff.rs
@@ -7,7 +7,7 @@ use buffer_diff::DiffHunkStatus;
 use collections::{HashMap, HashSet};
 use editor::{
     Direction, Editor, EditorEvent, EditorSettings, MultiBuffer, MultiBufferSnapshot,
-    SelectionEffects, ToPoint,
+    SelectionEffects, SplittableEditor, ToPoint,
     actions::{GoToHunk, GoToPreviousHunk},
     multibuffer_context_lines,
     scroll::Autoscroll,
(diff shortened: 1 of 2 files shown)


crates/agent_ui/src/agent_diff.rs
```diff
@@ -91,15 +91,23 @@ impl AgentDiffPane {
...
+                window,
+                cx,
+            );
+            diff_display_editor
                 .set_render_diff_hunk_controls(diff_hunk_controls(&thread, workspace.clone()), cx);
-            editor.register_addon(AgentDiffAddon);
-            editor.disable_mouse_wheel_zoom();
-            editor
+            diff_display_editor.rhs_editor().update(cx, |editor, _cx| {
+                editor.disable_inline_diagnostics();
```
cole-miller: (Minor) I don't think we need this

cole-miller (approved):

crates/agent_ui/src/agent_diff.rs
```diff
@@ -91,15 +91,23 @@ impl AgentDiffPane {
...
+                editor.register_addon(AgentDiffAddon);
+                editor.disable_mouse_wheel_zoom();
```
cole-miller: I think we should add disabling mouse wheel zoom (for both editors) as part of general SplittableEditor construction

crates/agent_ui/src/agent_diff.rs
```diff
@@ -91,15 +91,23 @@ impl AgentDiffPane {
...
+                cx,
+            );
+            diff_display_editor
                 .set_render_diff_hunk_controls(diff_hunk_controls(&thread, workspace.clone()), cx);
-            editor.register_addon(AgentDiffAddon);
-            editor.disable_mouse_wheel_zoom();
-            editor
+            diff_display_editor.rhs_editor().update(cx, |editor, _cx| {
+                editor.disable_inline_diagnostics();
+                editor.register_addon(AgentDiffAddon);
```
cole-miller, line 106: Should probably have `SplittableEditor::register_addon` that adds the addon to both sides since in this case we want the key context change to apply to both

Issue #48968, opened 2026-02-11
Memory leak from rapid FS events in watched directory — 50GB consumption, swap exhaustion, WindowServer crash


mrm007:
## Summary

Zed consumed ~50GB of memory (16GB physical + ~34GB swap) when a directory it had open contained a lock file being rapidly created/destroyed by an external process. The swap exhaustion filled the disk, starved WindowServer of resources, and required a hard reboot.

## Environment

- **Zed**: 0.222.4 (Stable)
- **macOS**: 15.7.3 (24G419)
- **Hardware**: MacBook Air M4, 16GB RAM

## Steps to Reproduce

1. Open `~/.config/opencode/` (or any directory) as a project in Zed
2. Multiple external processes rapidly create/destroy a `.lock` directory via `proper-lockfile` (the standard Node.js file locking library), contending on the same lock

**Realistic reproduction** using the same locking library as the trigger plugin:

Save as `thrash.mjs`, run `bun add proper-lockfile`, then launch 8 instances:

```js
import lockfile from "proper-lockfile";
import { writeFileSync, readFileSync, mkdirSync, existsSync, renameSync } from "node:fs";
import { randomBytes } from "node:crypto";
import { join, dirname } from "node:path";

const TARGET_DIR = "/tmp/zed-leak-test";  // Open this directory in Zed
const ACCOUNTS_FILE = join(TARGET_DIR, "state.json");
const INSTANCE = process.env.INSTANCE || "1";

const LOCK_OPTIONS = {
  stale: 10000,
  retries: { retries: 999, minTimeout: 5, maxTimeout: 50, factor: 1.1 },
};

if (!existsSync(ACCOUNTS_FILE)) {
  mkdirSync(dirname(ACCOUNTS_FILE), { recursive: true });
  writeFileSync(ACCOUNTS_FILE, JSON.stringify({ version: 1, data: [] }, null, 2));
}

async function save() {
  const release = await lockfile.lock(ACCOUNTS_FILE, LOCK_OPTIONS);
  try {
    const existing = JSON.parse(readFileSync(ACCOUNTS_FILE, "utf-8"));
    existing.data = [{ instance: INSTANCE, ts: Date.now(), rand: Math.random() }];
    const tmp = `${ACCOUNTS_FILE}.${randomBytes(6).toString("hex")}.tmp`;
    writeFileSync(tmp, JSON.stringify(existing, null, 2));
    renameSync(tmp, ACCOUNTS_FILE);
  } finally {
[...]

PR #54645, opened 2026-04-23
Disable miniprofiler by default

SomeoneToIgnore:
Needs https://github.com/zed-industries/zed/pull/54635 for the profile overrides added into default settings json to work.
Part of https://github.com/zed-industries/zed/issues/48968
Another part of the fix related seems to be https://github.com/zed-industries/zed/pull/45669 ?

Using the steps from the issue and profiling on macOs had shown that Zed has 2 memory "leaks" in play when a certain file is being rewritten a lot of times.

* First, the thread profiler registers a lot of tasks' data and fills its buffer to the limit:

image

* Second, if the buffer gets open, the undo history fragments start to creep up infinitely:

image

The PR aims to solve the first issue by disabling the profiling by default, yet leaving the way to turn in on quickly with settings.

The memory usage profiling shows that the memory usage is now dynamically affected by the new setting:

image

While the test directory being thrashed with the script from the issue, 
* first, Zed starts with the profiling disabled
* then gets the profiling enabled which results in the memory growth close to 1 minute mark of the screenshot
* last, the profiling gets disabled again, releasing all the memory accumulated

Release Notes:

- Improved Zed's default memory usage

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10790,9 +10790,11 @@ checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
 name = "miniprofiler_ui"
 version = "0.1.0"
 dependencies = [
+ "command_palette_hooks",
  "gpui",
  "rpc",
  "serde_json",
+ "settings",
  "smol",
  "theme_settings",
  "util",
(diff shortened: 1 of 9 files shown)


reflectronic (approved):
Thanks for doing this.

PR #54649, opened 2026-04-23
gpui_shared_string: Implement SharedString via smol_str

Veykril:
This reduces the size of Sharedstring from 32 bytes to 24 while also allowing for small-string optimization, meaning strings with length < 23 bytes will not actually allocate.

Release Notes:


osiewicz (approved):
Looks good. As long as we don't allocate for static str, we're gucci. It seems like we won't:
```
Additionally, a SmolStr can be explicitly created from a &'static str without allocation
```

PR #54660, opened 2026-04-23
Stop showing backtraces in default logging

SomeoneToIgnore:
When using `.log_err()` or `.detach_and_log_err(cx)` or similar.

See also https://github.com/zed-industries/zed/pull/36404, https://github.com/zed-industries/zed/pull/46383, https://github.com/zed-industries/zed/pull/44896, https://github.com/zed-industries/zed/pull/43917 and many more.

Before, a https://github.com/zed-industries/zed/blob/62bd61a6793c5f983fbc1700825208f19d780d52/crates/languages/src/go.rs#L568 line would show a backtrace even for `.context("no cached binary")?;` case, same as many other usages around the code:

before

After:

after

To show a backtrace as before, use `log_err_with_backtrace`.

Release Notes:

- Improved Zed's log output on errors


ConradIrwin (approved):
Thank You! These bother me so much

PR #54515, opened 2026-04-22
Show progress indicator in file finder

osyvokon:
When opening the file finder in a large project, it was impossible to tell whether the scan was complete and if the results were final. Now we show an animated arrow circle while the scan is active.

Screenshot 2026-04-22 at 18 42 01

Closes #48009

Release Notes:

- N/A

PR #54680, opened 2026-04-23
file_finder: Fix progress indicator layout

osyvokon:
Follow up to #54515 that wasn't committed by mistake 🤦 

Release Notes:

- N/A

### DIFF
--- a/crates/file_finder/src/file_finder.rs
+++ b/crates/file_finder/src/file_finder.rs
@@ -1771,18 +1771,13 @@ impl PickerDelegate for FileFinderDelegate {
         };
 
         h_flex()
-            .flex_none()
             .h_9()
             .px_2p5()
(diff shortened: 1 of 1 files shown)


danilo-leal:
We shouldn't need these styles, I think! Is it broken without them?

osyvokon:
@danilo-leal, nice catch! I simplified styles as much as possible. This was an agentic leftover after a last minute move of the progress circle to the editor line, where it feels a bit more natural. So it looks like this now:

Screenshot 2026-04-23 at 18 08 58

danilo-leal:
I think we can probably close this PR 😄 I pushed a few commits to your previous one with styling tweaks and they're all here, essentially. The diff for this PR is only including `items_center` which is already available in `h_flex`.

PR #54681, opened 2026-04-23
Adjust UI for gutter right-click menus

danilo-leal:
- Don't render the tooltip while the menu is open to avoid collision
- Try to maintain the trigger element visible while the menu is open
- Tweak bookmarks icon and tooltip description copy

Release Notes:

- N/A


yara-blue (approved):
Looks great to me!

yara-blue:
Flaky test it seems, I got the exact same one another PR, updating the branch should work.

PR #54694, opened 2026-04-23
agent_ui: Handle Cut for selection mentions 

NeelChotai:
Following on from https://github.com/zed-industries/zed/pull/54031, implement the same but for `Cut`.

Release Notes:

- N/A


probably-neb:
@zed-industries/approved

PR #54322, opened 2026-04-20
OSC8-URIs support in terminal

yara-blue:
Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #39339 

Release Notes:

- Fixed hyperlinks in the terminal on UNIX

Issue #39339, opened 2025-10-02
Ctrl+Click on terminal `file://` hyperlink works randomly


yajo:
### Summary

When navigating ripgrep results, Zed won't react to Ctrl+Click reliably.

### Description

Steps to reproduce:
1. Run `rg "--hyperlink-format=file://{path}:{line}:{column}" your_search_term` in Zed's terminal.
2. Ctrl+click on file numbers or file names hyperlinks.

**Expected Behavior**: Zed always navigates to the link.
**Actual Behavior**: Zed navigates to the link randomly.

<details><summary>Video</summary>

In this video I Ctrl+Click once each result hyperlink, in loop. You'll see that Zed opens the result randomly:

https://github.com/user-attachments/assets/0e686251-f762-4019-baa6-3fb4b35ed71c

</details>

@moduon MT-10900

### Zed Version and System Specs

Zed: v0.204.5 (Zed) 
OS: Linux Wayland nixos 25.05
Memory: 15.3 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics (RPL-P) || Intel open-source Mesa driver || Mesa 25.0.7

ConradIrwin:
@yajo is there any pattern to the  links that work and don't work?

I'm curious if it's something like a specific character (maybe spaces, or non-ascii) in the path. Or if they're printed in some way that's not relative to the current directory?

yajo:
I'd say that Zed opens more often links in the file name than in the file number, but as you see in the video, it's not 100% accurate.

I'd also say that Zed likes more when a path is written directly in the terminal instead of added as a hyperlink.

niedzielski:
I'm on Debian GNOME Wayland and use `rg --hyperlink-format=default`. control-clicking links in Zed doesn't seem to do anything at all on my machine.

PR #54695, opened 2026-04-23
terminal: Fix unreliable hyperlink display 

davewa:
Improves reliability of hyperlink painting

We were missing a `cx.notify()` when updating the hyperlink target.

# TODO
- [ ] More manual testing
- [ ] Add a unit test

Release Notes:

- terminal: Fixed unreliable hyperlink navigation

### DIFF
--- a/crates/terminal_view/src/terminal_path_like_target.rs
+++ b/crates/terminal_view/src/terminal_path_like_target.rs
@@ -106,17 +106,21 @@ fn possible_hover_target(
     cx.spawn(async move |terminal_view, cx| {
         let file_to_open = file_to_open_task.await;
         terminal_view
-            .update(cx, |terminal_view, _| match file_to_open {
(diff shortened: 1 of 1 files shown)


davewa:
Note that this PR addresses unreliable hyperlinks in general, so if #39339 really is specific to OSC8-URIs addressed by #54322, this is still useful.

davewa:
Also, while working the todos (more manual testing) now, I'm having difficulty repo'ing #39339 in main rn, so possibly this is unrelated to that? However, I have been experiencing unreliable hyperlinks in general on released Zed builds for some time, and this change definitely fixes that on one of my development branches, so I'm going to continue with adding tests for this, and if that proves successful, I'll take this out of draft, or close otherwise.

davewa:
So this only affects repainting the hyperlink, and doesn't fix the issue where repainting doesn't happen if the terminal isn't focused. If there isn't already a bug for hyperlinks not displaying if the terminal tab isn't focused, I'll create one and go from there.

PR #54698, opened 2026-04-23
agent_ui: Restore `Ctrl + >` behavior for whole lines

NeelChotai:
Restores current line fallback when using `Ctrl + >` to add context to the agent.

Release Notes:

- N/A


danilo-leal:
What's the line fallback we're restoring here? Not quite sure I know...

NeelChotai:
fyi @danilo-leal this partially reverts some changes made in https://github.com/zed-industries/zed/pull/52860.

This maintains the fix to the terminal side, so additional context from editor doesn't get added to the agent panel when unfocused, however it felt very strange having to select a single character to add a line of context to the agent

gaauwe:
@zed-zippy approved

PR #54701, opened 2026-04-23
gpui: Fix out-of-bounds panic rendering GIFs

smitbarmase:
Clamp the cached `frame_index` of an `img` element when the underlying image changes so a stale index from a previous, longer animation cannot index out of bounds on the new image.

Release Notes:

- Fixed a panic when a GIF was replaced with one that had fewer frames in Markdown Preview.


dinocosta (approved):
🚢🚢🚢

Issue #54697, opened 2026-04-23
The application crashed


bradckeer:
### Reproduction steps

The application crashed and wouldn't open any more files. After force-closing it, I tried to open it again, but it wouldn't open, even though it was throttling processor usage. (The app has no settings, extensions, or plugins installed; it's on its default configuration.)

### Current vs. Expected behavior

Failed to load the database file.

### Zed version and system specs

Zed: v0.233.8+stable.247.aa7f9dc9f87ca27252eea8f6ef0a46122af043f7 (Zed) 
OS: Windows 11 25H2 (Compilation SO 26200.8246)
Memory: 32 GiB
Architecture: x4
GPU: NVIDIA GeForce GTX 770 || NVIDIA Corporation || 475.14 r475_06

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

yeskunall:
@bradckeer -- is this the only error message you’re seeing?:

```
Failed to load the database file.
```

What about your [Zed log](https://zed.dev/docs/troubleshooting#zed-log)?

vickyqt:
Maybe try re-installing zed as it doesn't crashes on my system, I have same Windows version
`
Zed: v0.233.8+stable.247.aa7f9dc9f87ca27252eea8f6ef0a46122af043f7 (Zed) 
OS: Windows 10.0.26200 Version 25H2 (OS Build 26200.8246)
Memory: 23.7 GiB
Architecture: x86_64
GPU: Intel(R) Iris(R) Xe Graphics || Intel Corporation || 32.0.101.7082
`

yeskunall:
Haven’t heard back from OP in quite some time, so I’ll be closing this one out now.

PR #54705, opened 2026-04-23
db: Fix db recovery on failure to load db

feitreim:
https://github.com/zed-industries/zed/issues/54697 Highlights an issue I have also been experiencing, where after a crash (or some kind of other event that screws up the db, every subsequent startup just gives a toast that says "Failed to load the database file". I'd actually been experiencing this on my own, on nightly, for a while now but I had just chocked it up to some kind of nightly weirdness, this issue made me realize its not.

In the code for opening the db, the doc comment stated that on failure, it would move the old db to a backup location, and then create a new db, however, the code didn't actually do this, it just tried once, and then failed and gave up, triggering the unhelpful toast. My changes just fulfill the promise made by this doc comment.

Also added a toast that triggers on the successful rebuild, because there will still likely be some data lost.

Also to be clear this doesn't fix the issue causing the crash in #54697.

New Toast:

Screenshot 2026-04-23 at 3 30 35 PM

Self-Review Checklist:

Release Notes:

- db: Improved recovery on failure to load


yeskunall:
@feitreim I haven’t run into this (in any channel) for quite some time now. Have you?

feitreim:
@yeskunall yeah i haven't seen this in a while either

SomeoneToIgnore:
Given that no one had experienced this, let's close for now.

Thank you for the idea, we might need to revisit it, but later.

PR #54708, opened 2026-04-23
keymap_editor: Fix create keybinding button clipping out of editor

felixschwamm:
## Summary

In the keymap editor, the search input at the top used `size_full()` while the
adjacent button row had `min_w_96()`. On narrow panes (e.g. side-by-side
splits), this caused the action buttons — including "Create keybinding" — to
overflow and clip out of the editor.

This change lets the search input flex and shrink (`flex_1()` + `min_w_0()`),
and makes the button row `flex_none()` so it keeps its natural width and
stays visible at any pane size.

## Before / After

Screenshot 2026-04-23 215511

## Test plan

- [x] Open the keymap editor (`zed: open keymap editor`)
- [x] Confirm the "Create keybinding" and other action buttons remain visible
- [x] Confirm the search input shrinks gracefully instead of pushing buttons off-screen


Veykril (approved):
Thanks!

PR #42817, opened 2025-11-15
Custom headers for OpenAPI compatible providers

aurabindo:
Discussion thread I created for the feature: https://github.com/zed-industries/zed/discussions/42663

Release Notes:

- Add custom header support in settings
- Create a UI to manipulate custom headers in AI settings

PR #54712, opened 2026-04-23
Add custom headers for OpenAI-compatible providers

chindris-mihai-alexandru:
## Summary

This refreshes the earlier custom-header work from #42817 and #45673 with a smaller, settings-first implementation.

It adds `custom_headers` support to `language_models.openai_compatible`, threads those headers through both `/chat/completions` and `/responses`, and documents the setting in the provider docs.

While testing custom OpenAI-compatible gateways, I also ran into streaming responses that emit `data: null` keepalive chunks, so this PR skips those chunks instead of failing deserialization.

## Testing

- `cargo test -p open_ai -q`
- `cargo check -p zed`

Release Notes:

- Improved OpenAI-compatible providers to support custom HTTP headers and ignore null SSE keepalive events.


bennetbo:
Thanks for working on this. Closing in favour of #55246

PR #54717, opened 2026-04-23
editor: Make `editor: expand excerpts` actually expand all excerpts

### LINKED ISSUE #54651: `editor: expand excerpts` doesn't expand all, just one
### Reproduction steps

1. Open a 2 pane diff view, like from any git change.
2. Run the command: `editor: expand excerpts`

Current behaviour: only a single excerpt is expanded
Expected behaviour: all the excerpts are expanded, the full file is visible

Zed: v0.233.7+stable.244.2f844c3cdb1380f992a892f173eb45b6bcaad24c (Zed) 
OS: macOS 15.7.3
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

log doesn't show anything related

### Relevant Zed settings

settings doesn't have anything related

### Relevant Keymap

I'm not using a keymap for this

_No response_

None

feitreim:
Expand all excerpts had a doc comment describing it as expanding all excerpts, but in practice it only expanded the excerpt that was the most relevant.

I fixed that to make it expand all excerpts. 

video:

https://github.com/user-attachments/assets/9858ebda-199c-4f72-8a2f-3cd606b0eff4

Self-Review Checklist:

Closes #54651 

Release Notes:

- editor: `expand excerpts` now has correct documentation explaining its function.

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -192,7 +192,7 @@ pub struct SelectDownByLines {
     pub(super) lines: u32,
 }
 
-/// Expands all excerpts in the editor.
+/// Expands all excerpts with selections.
 #[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
 #[action(namespace = editor)]
 #[serde(deny_unknown_fields)]


SomeoneToIgnore (changes requested):
Thank you, this is quite a stark change of the existing behavior, so let's not ship the PR in this form to avoid breaking someone's workflow for an issue with no upvotes.

* First of all, a very nice spot of the 

https://github.com/zed-industries/zed/blob/5294e16547d9cacba49bbc8968183fb70fe8b58b/crates/editor/src/actions.rs#L195

doc being wrong: let's instead document that it expands all excerpts with selections.

* Is it so that other actions behave differently?
That seems wrong, but would love to get a comparison.

* If selecting all and expanding excerpts (on macOs that's `cmd-a` and repeated `alt-enter`) does not work as a solution still, let's add a standalone action to do what's intended.
That action could have a doc and tests similar to what's done in the PR.

feitreim:
@SomeoneToIgnore Thanks for the feedback, updated it to just change the doc comment, I tested `cmd+a` then `shift+enter` to expand and that accomplished the goal just fine. 

Looking into other actions this seems like the correct and consistent name for the action, the name for the action as originally described would contain the word "All", like `FoldAll`, `UnfoldAll`, `StageAll`.

SomeoneToIgnore (approved):
Amazing, thank you for the small fix!

hyperknot:
Sorry this is still confusing. I'm the author of the original issue https://github.com/zed-industries/zed/issues/54651

Expand all should expand all excerpts, not just the selected ones. The non-all version should expand all the selected ones.

Which behaviour got chosen?

feitreim:
@hyperknot right now there is no `Expand All Excerpts` command, just `Expand Excerpts`, previously, the doc comment for `Expand Excerpts` said that it would effect all, instead just selected. Naming convention wise, this isn't correct. If there were an action that did expand ALL the excerpts, it would have All in the name, like `Expand All Excerpts`. 

so right now the behavior is `Expand Excerpts` expands excerpts with selections in them.
There is no `Expand All Excerpts`.

hyperknot:
I see. Not having a full 2 pane view is the most painful part of Zed for me currently, coming from JetBrains. I actually keep JetBrains open only for this feature.

Ideally it'd be a setting something like "git-diff-initial-view": expanded/collapsed

hyperknot:
Opened feature request here:
https://github.com/zed-industries/zed/discussions/55261

Issue #30829, opened 2025-05-16
Opening the same SSH remote via CLI creates duplicate windows


nrontsis:
### Issue
When opening the same SSH remote location twice through the Zed CLI, two separate windows are created. This behavior is inconsistent with how Zed handles local folders, where opening the same folder twice refocuses the existing window instead of creating a duplicate.

### Expected Behavior
Opening the same SSH remote location multiple times via the CLI should refocus the existing window containing that remote location, consistent with how local folder paths are handled.

### How to Reproduce
Run twice:
```
zed ssh://user@example.com/path/to/directory
```

### Zed Version and System Specs

Zed: v0.185.15 (Zed) 
OS: macOS 15.4.1
Memory: 36 GiB
Architecture: aarch64

paulhammond:
This also happens when running `open zed://ssh/user@example.com/path/to/directory`

notpeter:
I can partially reproduce the behavior you are describing with Zed: v0.201.0 (Zed Nightly 9e8ec72bd5c697edc6b61f4e18542afc4e343a1b) on macOS connecting to Linux.

When I run the same command (e.g. `zed ssh://think21/~/.config/`) again with that workspace open, instead of a second workspace window, I get an error:
"Failed to connect over SSH" / "Timed out pinging remote client".  

<img width="268" height="216" alt="Image" src="https://github.com/user-attachments/assets/c1d0e779-d31e-4ef7-9f5a-ae5d0eec8e68" />

Looking at the logs, I think the second session is killing the first, and then the first is reconnecting, but not sure.

```
2025-08-19T16:48:36-04:00 ERROR [agent_ui::language_model_selector] Failed to authenticate provider: Anthropic: credentials not found
2025-08-19T16:48:36-04:00 ERROR [remote_server::unix] (remote proxy) encountered error while forwarding messages: stderr_task failed

Caused by:
    stderr closed, terminating...
2025-08-19T16:48:36-04:00 ERROR [remote_server] (remote proxy) exiting due to error: stderr_task failed
2025-08-19T16:48:36-04:00 WARN  [remote::ssh_session] ssh heartbeat: connection activity channel has been dropped. stopping.
2025-08-19T16:48:36-04:00 ERROR [remote::ssh_session] proxy process terminated unexpectedly
2025-08-19T16:48:36-04:00 INFO  [remote::ssh_session] setting state to 'reconnecting'
[...]

PR #54733, opened 2026-04-24
workspace: Treat remote roots as directories

jknlsn:
## Summary

- Add a regression test covering reopening an already-open remote workspace root inside the reused workspace path
- Use the resolved worktree entry instead of the local filesystem when `Workspace::open_paths` decides whether a reused path is a directory
- This fixes the follow-on case where reopening an already-open SSH workspace root could show `Error: opening project path ...` because the remote root was being treated like a local file open

I kept this fix in `Workspace::open_paths` rather than adding another remote-specific branch higher up, because the bug was specifically in the generic reused-workspace directory check.

## Testing

- `cargo test -p recent_projects test_reopen_existing_remote_root_treats_root_as_directory -- --nocapture`
- `cargo test -p recent_projects test_reuse_existing_remote_workspace_window_with_tilde_path -- --nocapture`
- `cargo test -p recent_projects test_reuse_existing_remote_workspace_window -- --nocapture`

Self-Review Checklist:

- [x] Tests cover the new/changed behaviour

Related to #30829

Release Notes:

- Fixed reopening an already-open SSH workspace root showing an `opening project path` error instead of treating the root as a directory

### DIFF
--- a/crates/recent_projects/src/remote_connections.rs
+++ b/crates/recent_projects/src/remote_connections.rs
@@ -734,6 +734,7 @@ mod tests {
         );
     }
 
+    #[gpui::test]
     async fn test_reopen_existing_remote_root_treats_root_as_directory(
         cx: &mut TestAppContext,
         server_cx: &mut TestAppContext,
--- a/crates/recent_projects/src/remote_connections.rs
+++ b/crates/recent_projects/src/remote_connections.rs
@@ -734,6 +734,102 @@ mod tests {
         );
     }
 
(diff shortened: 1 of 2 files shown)


jknlsn:
Fixed the missed clippy issue, and failing tests in latest rebase off `main`. I can instead merge main in if someone is reviewing this, otherwise in the meantime I find it easier to just rebase off main to keep the PR current!

Veykril (approved):
Thanks!

jknlsn:
> Thanks!

No problem! Saw that my test had an issue on Windows, just pushed a test only fix for that now.

PR #50595, opened 2026-03-03
markdown: Render checkboxes in markdown table cells

iam-liam:
Render `[x]` and `[ ]` as checkbox widgets when they appear as the sole content of a markdown table cell. Previously these were displayed as raw text. List-item checkboxes were already rendered correctly; this extends the same treatment to table cells.

Fixes #50045.

## What this does

- Table cells containing only `[x]`, `[X]`, or `[ ]` now render as visual checkboxes instead of plain text
- Both markdown rendering paths are covered: the `markdown` crate (agent panel, chat) and the `markdown_preview` crate (file preview)
- Checkboxes are display-only, matching the existing list-item checkbox behavior

## How it works

pulldown-cmark splits `[x]` in table cells into three separate `Text` events (`[`, `x`, `]`) rather than emitting a `TaskListMarker` event (which only fires for list items per the GFM spec). The fix operates at each crate's natural interception point:

- **`markdown` crate**: After all text events for a table cell have been buffered, `replace_pending_checkbox()` checks the accumulated text before the cell div is finalized. If it matches the checkbox pattern, the pending text is replaced with a `Checkbox` widget.
- **`markdown_preview` crate**: In `render_markdown_text()`, text chunks whose trimmed content matches the checkbox pattern are rendered as `MarkdownCheckbox` widgets instead of `InteractiveText`.

## Scope

Three files, purely additive:

- `crates/markdown/src/markdown.rs` — `replace_pending_checkbox()` on builder, called at `TableCell` end
- `crates/markdown_preview/src/markdown_renderer.rs` — checkbox detection in `render_markdown_text()`
- `crates/markdown_preview/src/markdown_parser.rs` — test only

No changes to parser data models, GPUI, or any shared infrastructure.

## What's not in scope

- **HTML `<input type="checkbox">`** — pulldown-cmark strips these as raw HTML. Supporting them requires HTML tag parsing, which is a separate concern.
- **Interactive (click-to-toggle) checkboxes in tables** — table checkboxes are display-only. List-item checkboxes in Zed support Cmd+click toggling, but extending that to table cells would require tracking source ranges across the split parser events, which is a separate enhancement.

## Follow-up

Table checkbox interactivity (Cmd+click toggle) is straightforward to add as a follow-up — the source ranges are already available in `markdown_preview`, and the `markdown` crate would need minor callback plumbing.

## Screenshots

**Markdown checkbox before**

md-checkbox-before-1

md-checkbox-before-2

**Markdown checkbox after**

[...]

PR #54747, opened 2026-04-24
markdown: Make table cell checkboxes clickable

### LINKED ISSUE #53587: Checkboxes in Markdown output can't be clicked
### Reproduction steps

1. Run this agent prompt:

> Output a fictitious code review for a small "TODO app" project, with checkboxes for every "issue". Some issues should be marked as "trivial" (e.g. comment typos) and their checkboxed should be checked. No heading or anything, just 10 brief code review findings.

> The use case for checkboxes in the Agent panel is for the user to manually check them as they go over the code review items. We want to see how this process works here in Zed.

2. Attempt to check some checkboxes as you go through the code review

Image

Zed: v0.231.1+stable.213.e71b6aa44ffaac9bdaf986288289359e6f1b3572 (Zed)

iam-liam:
Closes #53587

Follow-up to #50595.

[#50595](https://github.com/zed-industries/zed/pull/50595) added display-only checkboxes when `[x]` / `[X]` / `[ ]` appears as the sole content of a markdown table cell. That PR called out interactivity as a later step — this PR delivers it.

https://github.com/user-attachments/assets/c666ee4c-7e31-4450-ab33-c07c82949818

## What this does

When a consumer of `MarkdownElement` supplies an `on_checkbox_toggle` callback, table-cell checkboxes now invoke it on click with the source range of the `[x]` / `[ ]` marker and the new checked state. This mirrors the existing list-item checkbox path.

For markdown preview (`markdown_preview_view.rs`) the callback is already wired to edit the source buffer, so clicking a checkbox in a previewed `.md` file now flips `[x]` to `[ ]` and back in the file, the same as it already did for list checkboxes.

## How it works

`replace_pending_checkbox` previously took the cell's outer source range and always built a visualization-only checkbox. It now takes the optional toggle callback instead, and reconstructs the marker's exact source range from `pending_line.source_mappings` — pulldown-cmark emits `[x]` in a table cell as three separate `Text` events, so the per-chunk mappings are needed to recover the 3-character range to rewrite. If a callback was supplied, the checkbox attaches an `on_click` that invokes it with that range and `!checked`; otherwise it falls back to the prior visualization-only rendering.

Two free helpers (`source_range_for_rendered` / `source_index_for_rendered`) were extracted so the mapping logic is unit-testable.

## Scope

One file: `crates/markdown/src/markdown.rs`. No changes to `markdown_preview` — since #52008 it renders through the shared `MarkdownElement`, so its existing `on_checkbox_toggle` wiring picks up table checkboxes for free.

## Relation to #53587

[#53587](https://github.com/zed-industries/zed/issues/53587) reports that markdown checkboxes in the **agent panel** can't be clicked. This PR is a necessary piece of that fix — without it, even if the agent panel wired `on_checkbox_toggle`, its table checkboxes would stay inert. It does not fully close #53587 on its own: the agent panel's `MarkdownElement` instances in `conversation_view.rs` / `thread_view.rs` don't currently wire `on_checkbox_toggle`, and wiring it there requires deciding how clicks should mutate the in-memory agent response (no source file to edit into). That's a follow-up.

## Test plan

**Unit tests** (2 new, both in `crates/markdown/src/markdown.rs`):
- `test_table_checkbox_marker_source_range` — walks parser events for a table with checkboxes, replays what the builder accumulates, and asserts the reconstructed source range slices to exactly `[x]` / `[ ]` in the original markdown (including a padded-whitespace case).
- `test_source_range_for_rendered_handles_split_chunks` — pins the mapping helper against the three-chunk layout pulldown-cmark produces.

**Automated**:
- [x] `cargo test -p markdown` — 46 tests pass (44 existing + 2 new)
- [x] `cargo test -p markdown_preview` — 3 tests pass
- [x] `./script/clippy -p markdown` — clean

**Manual** (against a preview of a markdown file with checkbox tables):
- [x] Click a checked `[x]` table cell — it becomes `[ ]` in the source and rerenders as unchecked
- [x] Click an unchecked `[ ]` — becomes `[x]`
- [x] Uppercase `[X]` toggles on click (replacement writes `[x]` / `[ ]`, matching the list-item behaviour)
- [x] Padded-whitespace cells still target just the three marker characters
- [x] Multiple checkbox columns in one table all independently clickable
- [x] Alignment variants (left/center/right) behave the same
- [x] Non-checkbox table text unaffected
- [x] List-item checkboxes continue to work (regression)

Release Notes:

- Made table-cell markdown checkboxes clickable in markdown preview, matching list-item checkbox behavior

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -2131,7 +2131,7 @@ impl Element for MarkdownElement {
                         builder.table.end_row();
                     }
                     MarkdownTagEnd::TableCell => {
-                        builder.replace_pending_checkbox(range);
+                        builder.replace_pending_checkbox(self.on_checkbox_toggle.clone());
                         builder.pop_div();
                         builder.table.end_cell();
                     }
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

iam-liam:
@Veykril no problem, we have a flaky test

  `extension_host::test_extension_store_with_test_extension` timed out after 60s (crates/extension_host/src/extension_store_test.rs:556). It has zero dependency on markdown. Just slow CI.

tredondo:
Thanks for the fix! Any reason why it doesn't work in the agent panel?

image

Prompt: Output a Markdown table with 5 rows and 2 columns: Manual test task (random string about a UI test item) and Done (checkbox).

PR #53860, opened 2026-04-14
dev_container: Resolve compose dockerfile path relative to build context

antont:
## Context

When a Docker Compose service specifies `context: ..` and `dockerfile: .devcontainer/Dockerfile`, Zed resolves the dockerfile path relative to the compose file's directory instead of the build context. This produces a doubled path like `.devcontainer/.devcontainer/Dockerfile` which doesn't exist.

Per the [Docker Compose spec](https://docs.docker.com/reference/compose-file/build/#dockerfile), the `dockerfile` field is relative to the build context directory.

The fix resolves the context directory first (relative to the compose file), then joins the dockerfile path to that.

Closes #53473

## Prior art

This fix is extracted from #53170 by @zdeneklapes, which addresses this bug among several other dev container startup issues. This PR isolates the dockerfile path resolution fix into a focused change to make it easier to review and merge independently.

Differences from #53170:
- **Scope**: Only the dockerfile-relative-to-context fix, not the other fixes (compose build args preservation, remote user fallback, Docker inspect labels, etc.)
- **Implementation**: Inline resolution in `dockerfile_location()` rather than separate helper methods
- **Absolute path handling**: Handles absolute dockerfile and context paths
- **Tests**: Two test cases — compose file inside `.devcontainer/` with `context: ..`, and compose file at project root with `context: .`

## How to Review

Single file change in `crates/dev_container/src/devcontainer_manifest.rs`:
- **Fix** (line 234-252): Resolve build context relative to compose file directory, then join dockerfile to that, instead of joining dockerfile to `config_directory` directly. Uses `normalize_path` to resolve `..` components.
- **Tests**: Two new `FakeDocker` compose config entries and corresponding tests asserting correct resolved paths.

## Self-Review Checklist

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Release Notes:

- Fixed docker-compose `dockerfile` path being resolved relative to the compose file instead of the build `context` directory.

PR #54749, opened 2026-04-24
dev_container: Resolve compose build context from declaring fragment

### LINKED ISSUE #54662: Dev Container incorrectly resolves Dockerfile path when using Docker Compose
### Reproduction steps

1. Have a project with the following structure:
```
.
├── docker-compose.yml
├── docker-compose.override.yml  ← contains build: with dockerfile: docker/django/Dockerfile
├── docker/
│   └── django/
│       └── Dockerfile
└── .devcontainer/
    ├── devcontainer.json
    └── docker-compose.devcontainer.yml
```

2. Create `.devcontainer/devcontainer.json`:
```json
{
  "name": "My Project",
  "dockerComposeFile": [
    "../docker-compose.yml",
    "docker-compose.devcontainer.yml"
  ],
  "service": "web",
  "workspaceFolder": "/code"
}
```

3. Create `.devcontainer/docker-compose.devcontainer.yml`:
```yaml
services:
  web:
    build:
      context: ..
      dockerfile: docker/django/Dockerfile
      target: development_build
    volumes:
      - ..:/code
```

4. Open the project in Zed
5. Open Command Palette → "Dev Containers: Reopen in Container"

**Current behavior:**
Zed always prepends `.devcontainer/` to the `dockerfile:` value, completely ignoring the `context:` field. Every path combination fails:

| `dockerfile:` value | Zed looks for | Result |
|---|---|---|
| `docker/django/Dockerfile` | `.devcontainer/docker/django/Dockerfile` | ❌ |
| `.devcontainer/Dockerfile` | `.devcontainer/.devcontainer/Dockerfile` | ❌ |
| `Dockerfile` | `.devcontainer/Dockerfile` | ❌ |

Error from Zed logs:
```
Failed to load Dockerfile: Failed to read file /Users/michael/Desktop/Backend/.devcontainer/docker/django/Dockerfile
Failed to start dev container: DevContainerUpFailed("Failed with nested error: FilesystemError")
```

**Expected behavior:**
`dockerfile:` should be resolved relative to `context:`, as described in the Docker Compose spec. With `context: ..` and `dockerfile: docker/django/Dockerfile`, the resolved path should be `<project_root>/docker/django/Dockerfile`.

Note: the VS Code Dev Containers extension handles the same project correctly without any `.devcontainer/` config at all — it picks up `docker-compose.yml` and `docker-compose.override.yml` from the project root automatically and resolves all paths correctly.

Zed: v0.233.7+stable.244.2f844c3cdb1380f992a892f173eb45b6bcaad24c (Zed) 
OS: macOS 14.8.2
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

```log

[Zed.log](https://github.com/user-attachments/files/27013574/Zed.log)

```

</details>

### Relevant Zed settings

No custom settings relevant to this issue.

### Relevant Keymap

No custom settings relevant to this issue.

_No response_

None

OmChillure:
### Fix

Centralize the correction in `docker_compose_manifest`: after fetching the merged config, re-read the raw YAML of the fragment that declares each service's build and resolve the rISSUEaw `context` relative to that fragment's directory. Every downstream consumer (`dockerfile_location`, `build_and_extend_compose_files`, and any future reader of `build.context`) now sees a correctly-anchored absolute path.

Skips normalization when the raw `context` contains `$`, since Compose does variable interpolation at merge time and the expanded value is the one to trust.

### Changes

- `normalize_build_contexts` : new helper that rewrites each service's `build.context` in-place from the declaring fragment's raw YAML
- `compose_fragment_build_context` : new helper that extracts `services.<name>.build.context` as written in a single compose fragment
- `docker_compose_manifest` calls `normalize_build_contexts` once, at the source
- `dockerfile_location`  simplified back to trusting `build.context`
- `find_compose_file_declaring_build`  docstring updated (the old claim that Compose "preserves paths verbatim" was incorrect)
- Test fake `get_docker_compose_config` the two-fragment case now emits an absolute pre-resolved context matching real Compose behavior. `test_dockerfile_location_with_multi_file_compose_override` now exercises the real bug and will fail without the normalizer.

### Self-Review Checklist:

- [ ] Unsafe blocks (if any) have justifying comments  ← N/A, no unsafe blocks

Video :

[Screencast from 2026-04-24 13-46-54.webm](https://github.com/user-attachments/assets/013b45a5-dbd4-4441-b8e7-f875aa068d34)

Release Notes:

- Fixed dev container failing to start when an override compose file uses a relative `build.context` (e.g. `context: ..`) that `docker compose config` rewrites against the project root


KyleBarton:
@OmChillure I'm not sure this is quite right. From the compose spec:

```
When you use multiple Compose files, all paths in the files are relative to the first configuration file specified with -f. You can use the --project-directory option to override this base path.
```

So we shouldn't find the context directory relative to the compose file which declares `context`. We should find the context directory relative to the first compose file specified in the `docker compose -f file1 -f file2... config` call. In other words, in your test, the assertion:

```
        let expected = PathBuf::from(TEST_PROJECT_PATH)
            .join("docker")
            .join("django")
            .join("Dockerfile");
        assert_eq!(
            devcontainer_manifest.dockerfile_location().await,
            Some(expected)
        );
```

Should actually assert that the dockerfile location is `{PARENT_OF_TEST_PROJECT_PATH}/docker/django/Dockerfile`

~I think `resolve_compose_dockerfile` is therefore doing the right thing already. That being said, I'm realizing we can probably simplify further by using the main service.build.context field directly, if specified, and default to the parent of the first file path otherwise. E.g. something like this (except cleaned up):~

```
                main_service.build.and_then(|b| {
                    let context_path = b.context.map(|c| PathBuf::from(c)).unwrap_or(
                        docker_compose_manifest
                            .files
                            .first()
                            .unwrap()
                            .parent()
                            .unwrap()
                            .to_path_buf(),
                    );
                    let Some(dockerfile) = b.dockerfile else {
                        return None;
                    };
                    Some(context_path.join(dockerfile))
                })

```

~to replace `resolve_compose_dockerfile`~

Edit:

Looking further, the above logic is just a reframing of what `resolve_compose_dockerfile` does already:

```
let context_dir = match context {
        Some(ctx) => {
            let ctx = PathBuf::from(ctx);
            if ctx.is_absolute() {
                ctx
            } else {
                normalize_path(&compose_dir.join(ctx))
            }
        }
```

ctx will almost certainly be absolute in this case (because `docker compose config` resolves for us), so it'll essentially do what I wrote poorly above



Sorry for the winding explanation, but in summary, I think this means that as of #53860, behavior is correct.

OmChillure:
> @OmChillure I'm not sure this is quite right. From the compose spec:
> 
> ```
> When you use multiple Compose files, all paths in the files are relative to the first configuration file specified with -f. You can use the --project-directory option to override this base path.
> ```
> 
> So we shouldn't find the context directory relative to the compose file which declares `context`. We should find the context directory relative to the first compose file specified in the `docker compose -f file1 -f file2... config` call. In other words, in your test, the assertion:
> 
> ```
>         let expected = PathBuf::from(TEST_PROJECT_PATH)
>             .join("docker")
>             .join("django")
>             .join("Dockerfile");
>         assert_eq!(
>             devcontainer_manifest.dockerfile_location().await,
>             Some(expected)
>         );
> ```
> 
> Should actually assert that the dockerfile location is `{PARENT_OF_TEST_PROJECT_PATH}/docker/django/Dockerfile`
> 
> ~I think `resolve_compose_dockerfile` is therefore doing the right thing already. That being said, I'm realizing we can probably simplify further by using the main service.build.context field directly, if specified, and default to the parent of the first file path otherwise. E.g. something like this (except cleaned up):~
> 
> ```
>                 main_service.build.and_then(|b| {
>                     let context_path = b.context.map(|c| PathBuf::from(c)).unwrap_or(
>                         docker_compose_manifest
>                             .files
>                             .first()
>                             .unwrap()
>                             .parent()
>                             .unwrap()
>                             .to_path_buf(),
>                     );
>                     let Some(dockerfile) = b.dockerfile else {
>                         return None;
>                     };
>                     Some(context_path.join(dockerfile))
>                 })
> ```
> 
> ~to replace `resolve_compose_dockerfile`~
> 
> Edit:
> 
> Looking further, the above logic is just a reframing of what `resolve_compose_dockerfile` does already:
> 
> ```
> let context_dir = match context {
>         Some(ctx) => {
>             let ctx = PathBuf::from(ctx);
>             if ctx.is_absolute() {
>                 ctx
>             } else {
>                 normalize_path(&compose_dir.join(ctx))
>             }
>         }
> ```
> 
> ctx will almost certainly be absolute in this case (because `docker compose config` resolves for us), so it'll essentially do what I wrote poorly above
> 
> Sorry for the winding explanation, but in summary, I think this means that as of #53860, behavior is correct.

Ahh you are correct @/KyleBarton, that makes sense, and #53860 already handles this correctly/. Thank you for the detailed explanation!

PR #54753, opened 2026-04-24
project_search: Fix project search status text and refactor search state

osyvokon:
This change fixes a small bug where we were showing "Loading project..." even when in fact we had already started the search.

It also refactors three booleans in the `SearchState` enum, so that it's harder to make similar mistakes in the future.

Release Notes:

- N/A

### DIFF
--- a/crates/search/src/project_search.rs
+++ b/crates/search/src/project_search.rs
@@ -235,13 +235,42 @@ pub struct ProjectSearch {
     active_query: Option<SearchQuery>,
     last_search_query_text: Option<String>,
     search_id: usize,
-    no_results: Option<bool>,
-    limit_reached: bool,
-    waiting_for_scan: bool,
+    search_state: SearchState,
     search_history_cursor: SearchHistoryCursor,
     search_included_history_cursor: SearchHistoryCursor,
     search_excluded_history_cursor: SearchHistoryCursor,
     _excerpts_subscription: Subscription,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+enum SearchState {
+    #[default]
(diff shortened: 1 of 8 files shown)


crates/search/src/project_search.rs
```diff
@@ -423,19 +453,17 @@ impl ProjectSearch {
                     project_search
                         .excerpts
                         .update(cx, |excerpts, cx| excerpts.clear(cx));
-                    project_search.no_results = Some(true);
-                    project_search.limit_reached = false;
-                    project_search.waiting_for_scan = false;
+                    project_search.search_state = SearchState::Running(SearchActivity::Searching);
```
SomeoneToIgnore: NIT: Seems redundant, as we've just set it at the top?

Also curious about the state change flow: seems like now we can go Searching (initial) -> WaitingForScan -> Searching path due to the loop below?

Overall, feels that we are not searching yet when setting that Searching one here and initially.

SomeoneToIgnore (approved):
Nice, thank you for refactoring it.

I have one optional questionable NIT that I do not understand.

crates/search/src/project_search.rs
osyvokon: Oops, redundant indeed. Thanks!

The UI state flow is `Idle` --> `Running(Search)` (after user hits Enter) --> `Running(WaitingForScan)` (only if needed) --> `Running(Search)` (only after the scan) --> `Completed(_)`

We're not really doing search when we first set `Running(Search)`, this is just to update the UI. We could change the workflow a to transition from `Idle` to `Running(_)` on the actual signal. I played safe to avoid potential delays 🤷

PR #54763, opened 2026-04-24
agent_ui: Switch to All filter when last archived thread is removed

lixiaoyan:
When the archive view's filter is set to "Archived Only" and the user deletes (or unarchives) the last archived thread, the toggle button becomes disabled because there are no archived threads left. With the filter still pinned to "Archived Only", the list goes empty and there's no way for the user to switch back to the All view.

This fixes it by automatically falling back to `ThreadFilter::All` inside `update_items` whenever the current filter is `ArchivedOnly` but the store has no archived entries. Since `update_items` is invoked from the `ThreadMetadataStore` observer, this covers all paths that mutate the store, not only deletions triggered from this view.

Release Notes:

- Fixed the agent thread archive view getting stuck on an empty "Archived Only" list after the last archived thread was removed.

### DIFF
--- a/crates/agent_ui/src/threads_archive_view.rs
+++ b/crates/agent_ui/src/threads_archive_view.rs
@@ -261,6 +261,19 @@ impl ThreadsArchiveView {
     }
 
     fn update_items(&mut self, cx: &mut Context<Self>) {
+        // If we're filtering to archived threads but none remain (e.g. the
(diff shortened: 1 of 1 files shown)


NeelChotai:
nice! thank you for the fix

NeelChotai (approved):

PR #54791, opened 2026-04-24
Fix when agent-powered merge conflict button shows up

danilo-leal:
This PR fixes the logic to dismiss the "resolve merge conflict with agent" button. Previously, we were just observing `merge_heads_by_conflicted_path`, which seems to be intentionally sticky, preserving the conflicted paths until changes are either committed or aborted. This would make the button to resolve conflicts show up even _after_ the changes get resolved. Now, we're checking whether paths are _currently_ conflicted (`is_conflicted()`), and if they are not, we don't display the button, even though the resolution might have not been committed or aborted yet.

As a bonus, in this PR, I'm also putting the resolve conflict button _before_ the activity indicator, so as to avoid bounciness, and did a quick polish of the activity indicator button itself, by using the `Button` component.

Release Notes:

- Fixed a bug with the merge conflict "resolve with agent" button where it would be displayed even though all conflicts have already been resolved.

### DIFF
--- a/crates/activity_indicator/src/activity_indicator.rs
+++ b/crates/activity_indicator/src/activity_indicator.rs
@@ -3,8 +3,8 @@ use editor::Editor;
 use extension_host::{ExtensionOperation, ExtensionStore};
 use futures::StreamExt;
 use gpui::{
-    App, Context, CursorStyle, Entity, EventEmitter, InteractiveElement as _, ParentElement as _,
-    Render, SharedString, StatefulInteractiveElement, Styled, Window, actions,
+    App, Context, Entity, EventEmitter, InteractiveElement as _, ParentElement as _, Render,
+    SharedString, Styled, Window, actions,
 };
 use language::{
     BinaryStatus, LanguageRegistry, LanguageServerId, LanguageServerName,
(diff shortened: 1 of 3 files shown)


ConradIrwin (approved):
Fabulous, thank you!

aviatesk:
Looks like the activity indicator polish in this PR introduced a regression: the spinner no longer animates. `content_to_render` still produces `Content::icon` values with `.with_rotate_animation(...)` for several cases (LSP `$/progress`, debug session startup, long-running git/fs ops), but the new render path only checks `content.icon.is_some()` and then always renders a static `IconName::LoadCircle`:

```rust
.when(content.icon.is_some(), |this| {
    this.start_icon(
        Icon::new(IconName::LoadCircle)
            .color(Color::Muted)
            .size(IconSize::Small),
    )
})
```

The previous `ButtonLike`-based code rendered `content.icon` directly via `.children(content.icon)`, preserving both the icon kind (Warning / Download / ArrowCircle) and the rotate animation. After this PR everything collapses to a non-animated `LoadCircle`, so LSP progress no longer spins in the status bar.

PR #54806, opened 2026-04-24
editor: Cache code actions per viewport instead of per selection

Dnreikronos:
Summary

This swaps the per-selection code action fetch for a viewport-based cache. Code actions are now pre-fetched for visible buffer ranges (plus a margin), so they're already there when you move the cursor or scroll. This is the caching approach described in [#53210 (comment)](https://github.com/zed-industries/zed/issues/53210#issuecomment-4295663190) as a follow-up to #53571.

Closes #53210

### What changed

Fetches are scoped to the visible buffer range plus a 50% margin above and below, instead of firing on every selection change. A `CodeActionCache` stores results per `BufferId` with the buffer version and fetched row range. If the visible rows fall within the cached range and the version hasn't changed, we skip the re-fetch.

Debounce varies by trigger:
- `BufferEdited` → `CODE_ACTIONS_DEBOUNCE_TIMEOUT`
- `NewLinesShown` (scroll) → 100ms
- `ProvidersChanged` → `Duration::ZERO`

Each cached action carries two row ranges: a broad `row_range` from the LSP request range (used for the bolt indicator) and an optional `diagnostic_row_range` from the LSP diagnostic (used for menu filtering). The broad range keeps the indicator stable during cursor movement. The diagnostic range keeps the menu accurate when you click it.

`fetch_rows` is recomputed from anchors against the post-await buffer snapshot, so row ranges stay correct when edits land during the LSP roundtrip.

### How it differs from the old approach

The [comment](https://github.com/zed-industries/zed/issues/53210#issuecomment-4295663190) noted that per-selection fetching caused blinking and that a viewport approach would need "proper caching and updating" across multi-buffer excerpts. That's what this does:

| Aspect | Before (per-selection) | After (viewport cache) |
|--------|----------------------|----------------------|
| Fetch trigger | Every selection change | Scroll, edit, or provider change only |
| Fetch range | Cursor position | Visible rows + 50% margin |
| Cache granularity | Single `CodeActionsForSelection` | Per-buffer with version + row coverage |
| Bolt indicator | Flickers on cursor move (fetch in-flight) | Stable (cached actions always available) |
| Menu precision | Shows all actions from cursor fetch | Filters by diagnostic row range |
| Multi-buffer | Single selection only | Iterates all visible buffer ranges |

### What this fixes

The bolt icon no longer flickers during cursor movement, Ctrl+F jumps, or arrow key navigation. One viewport fetch replaces N per-cursor-change fetches. Small scrolls are cache hits because of the margin prefetch.

## Self-Review Checklist

> The new cache logic (`CodeActionCache`, `cached_code_actions_for_row`, `has_cached_code_actions_at_cursor`, invalidation/prefetch) doesn't have unit tests yet. Behavior was validated manually across the scenarios in the demo below.

Closes #53210

## Demo

1. Cursor movement through lines with code actions - bolt icon stays stable
2. Ctrl+F search jumping between results - bolt icon doesn't blink
3. Clicking bolt icon - menu shows only row-relevant actions
4. Scrolling - code actions appear for newly visible lines after brief debounce
5. Editing a line (e.g., using an unused variable) - cache invalidates, bolt updates

_Video placeholder_

https://github.com/user-attachments/assets/70fba6a4-98f4-4190-aff4-ba13f496614e

Release Notes:

- Fixed code action indicator (bolt icon) flickering during cursor movement by caching actions per viewport instead of fetching per cursor position.


Dnreikronos:
@SomeoneToIgnore, as we paired on the last PR related with that, I think that would be good to you review the changes, as you have more context.

If you think that something can be improved or it's not right, please let me know!!!

SomeoneToIgnore:
Thanks, I see that the code actions are cached by rows which will never work properly — we'll have to use anchors; some conflicts appeared, quite a lot of other related regressions seem to have appeared.

PR #54808, opened 2026-04-24
Shorten explanation when edit predictions are disabled for a language

injust:
The current text is a bit wordy and puts the actually important information at the end:

image

The button already shows as disabled, so the text just needs to explain why.

I don't think this affects accessibility, because a screen reader should be able to tell that the button is disabled.

Self-Review Checklist:

Release Notes:

- N/A


Veykril (approved):
Thanks!

injust:
Rebased on main and ran `cargo fmt`.

PR #54816, opened 2026-04-24
agent_ui: Add thread search to the agent panel

Tahinli:
Self-Review Checklist:

Closes #ISSUE https://github.com/zed-industries/zed/discussions/39338 partially

Release Notes:

Adds in-thread search to the agent panel: Cmd/Ctrl+F (and `/` in vim mode) deploys a `BufferSearchBar` over the active thread, matches the query against assistant messages, user messages, tool-call labels, plan entries, and terminal commands, and highlights the matches inline. The bar's existing "next / previous / case / whole-word / regex" controls work unchanged; an "Include Hidden Context" toggle extends search into collapsed tool calls and thinking blocks on demand, auto-expanding only what the match lives  inside and collapsing again on dismiss.

https://github.com/user-attachments/assets/5b910cfb-0c92-40e8-aeca-b3866b27aa9f

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -297,6 +297,8 @@
       "ctrl-alt-down": "agent::ScrollOutputLineDown",
       "ctrl-alt-shift-pageup": "agent::ScrollOutputToPreviousMessage",
       "ctrl-alt-shift-pagedown": "agent::ScrollOutputToNextMessage",
+      "find": "buffer_search::Deploy",
+      "ctrl-f": "buffer_search::Deploy",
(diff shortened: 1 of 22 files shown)


Tahinli:
Please review carefully guys. It's a bit big update. I don't want to make any mistake that causes problem for ZED in no case. I will try my best to cooperate and fix each and every request that ZED staff provides.

benbrandt:
Thanks for taking a look, but given we don't have search over historic external agents we'll need to think about how to communicate this feature, and I don't think this is the approach we want to take. Thanks though!

PR #54817, opened 2026-04-24
Show restore checkpoint more

ConradIrwin:
Release Notes:

- Fixed the agent panel to show "restore checkpoint" as soon as the agent starts editing instead of waiting until the turn ends.


JohnXu22786:
> Release Notes:
> 
> * Fixed the agent panel to show "restore checkpoint" as soon as the agent starts editing instead of waiting until the turn ends.

Well, the button also not displays after finished sometimes #56336 , and I haven't figured out how it triggers.

bennetbo:
Closing in favour of #56611

PR #54850, opened 2026-04-25
editor: Add smooth cursor animation and redesigned caret shape

### LINKED ISSUE #4991: Animated/smooth cursor movement

sunwunking990:
# PR — smooth_cursor

 `editor: smooth cursor animation and redesigned bar cursor shape (opt-in)`

---

Adds `smooth_cursor` — an opt-in setting that animates the local bar cursor and gives it a
redesigned shape. Off by default.

```jsonc
// settings.json
{
  "smooth_cursor": true
}
```

## What it does

When the cursor moves, it glides from the previous position to the new one instead of jumping.
The animation uses ease-out-cubic easing — fast off the mark, settles quickly. Duration scales with
distance: 25ms for adjacent characters, up to 70ms for large jumps across the file. It reads as
snappy, not floaty.

Only the local bar cursor is animated. Block, underline, hollow shapes and all collab/remote
cursors behave exactly as before. When `smooth_cursor` is false, there is zero overhead from
this feature on the non-animated path.

## Demo Video

https://github.com/user-attachments/assets/f4fae808-3073-4152-a701-b1dc2dc9b18c

## Cursor shape

With `smooth_cursor` on, the bar cursor gets a subtle redesign. Two styles are available via
`smooth_cursor_style`:

- **`"shaped"`** (default) — The top terminates in a rounded semicircle; the bottom sits flat
  against the baseline. 2.5px wide so the dome reads clearly without the cursor feeling heavier
  than the surrounding text.
- **`"classic"`** — A gradient fade-in from transparent at the top to full color below, standard
  2px bar width with rounded bottom corners.

```jsonc
"smooth_cursor": true                     // "shaped" by default
"smooth_cursor_style": "classic"          // opt into the gradient style
```
## Selection

When `smooth_cursor` is enabled, text selections also pick up more pronounced corner rounding:
0.30 × line height, versus 0.15 for the existing `rounded_selection` setting. The fill color is
the theme's `player_color.selection` untouched — no saturation or lightness manipulation — so it
stays coherent across all themes without any special-casing.

## Implementation

**`CursorAnimationState`** (editor.rs) — A plain struct on `Editor` holding `from`, `to`
(viewport-relative pixel positions), `cursor_position` (the logical `DisplayPoint`), `started_at`,
and `duration`. No heap allocation. `visual_origin(now)` does the eased interpolation;
`is_settled(now)` returns true once the animation is done.

**State update** (element.rs, `layout_visible_cursors`) — Each frame, for the local bar cursor,
we compute the target position and check whether the logical `DisplayPoint` changed. Using
`DisplayPoint` rather than raw pixels is important: a pure scroll event shifts viewport-relative
coordinates without moving the cursor in the document. Triggering animation on pixel deltas alone
caused the cursor to drift upward when holding Ctrl+Down. The `DisplayPoint` gate prevents that.

**Frame scheduling** (element.rs, `paint_cursors`) — After painting, if `!is_settled()` the
element calls `window.request_animation_frame()`. When the animation finishes, no more frames are
requested until the cursor moves again. The cost is bounded: at most 70ms / ~4 frames per cursor
move at 60fps.

**Blink** — `BlinkManager::pause_blinking()` is called when the cursor moves, so the cursor stays
solid during the glide and only resumes blinking after it's been idle.

## Edge cases

**Rapid typing** — Each keypress starts a new animation from the current visual position (not the
logical destination of the previous one). At fast typing speeds the cursor tracks closely; there's
no compounding lag.

**Scroll without cursor move** — `DisplayPoint` comparison means holding Ctrl+Down or scrolling
the viewport doesn't trigger animation.

**Toggle mid-session** — `cursor_animation` persists on `Editor` but is never read when the
setting is off. No visual artifact.

## Files changed

| File | What changed |
|---|---|
| `crates/editor/src/editor.rs` | `CursorAnimationState` struct, `cursor_animation: Option<CursorAnimationState>` field |
| `crates/editor/src/element.rs` | Animation state update in layout, `request_animation_frame` in paint, cursor shape via PathBuilder, selection corner rounding |
| `crates/editor/src/editor_settings.rs` | `smooth_cursor: bool`, `smooth_cursor_style: SmoothCursorStyle` |
| `crates/settings_content/src/editor.rs` | `smooth_cursor: Option<bool>`, `smooth_cursor_style: Option<SmoothCursorStyle>`, `SmoothCursorStyle` enum |
| `assets/settings/default.json` | `"smooth_cursor": false`, `"smooth_cursor_style": "shaped"` |
| `crates/settings/src/vscode_import.rs` | `smooth_cursor: None`, `smooth_cursor_style: None` |

## Testing

- [x] Normal typing — cursor tracks each character
- [x] Arrow key navigation
- [x] Vim `h j k l`, word jumps `w b e`
- [x] Large jumps: `G / gg`, goto definition
- [x] Multi-cursor editing
- [x] Hold Ctrl+Down — cursor should not drift during scroll
- [x] Collab session — remote cursors stay instant
- [x] `smooth_cursor: false` — zero behavioral change from baseline
- [x] Rapid key repeat (hold `j` for a couple seconds)

Closes #4991 
Release Notes:

- Added `smooth_cursor` setting to animate the cursor and apply a redesigned caret shape (opt-in, off by default)


sunwunking990:
I apologize for not being able to sign the CLA, I have contacted Zed billing department through the proper channel. This is just a message to simply note that it'll take some time to get that done, thanks.

MrSubidubi:
Thanks for opening this pul request.

There has been plenty of discussion around this feature and how to implement it as outlined through https://github.com/zed-industries/zed/issues/4991 and related PRs. For what was discussed there, I do not see much of this in here, as there is no optimization done here and the last PR went off-course in that regard as well as far beyond scope. 

As I do not see a path forward for this with the changes here, I am gonna close this eagerly, sorry for that. We'll probably want to have this at some point, but this requires much more thought as well as a more clear path for moving this forward. I hope you understand that, thank you for your time spent on this regardless.

sunwunking990:
Understood.

PR #54857, opened 2026-04-25
gpui_web: Implement clipboard paste/copy/cut

Yatekii:
Hook the browser's native `paste`, `copy`, and `cut` events on the window's hidden capture input element and bridge them to the `PlatformInputHandler`:
- paste: read clipboard text, replace selection via the input handler
- copy: read selected text via the input handler, write to clipboardData
- cut: read selected range+text, write to clipboardData, clear range

Exclude Cmd+V/Ctrl+V, Cmd+C/Ctrl+C, Cmd+X/Ctrl+X from the keydown `preventDefault()` such that the browser's clipboard event fires.

I used Claude to edit this and I tested all of it for my application. It works great :)

I am happy to clean this up a bit more, but I have some questions that I would like some input on how we should handle this before I would like to do that.

### DIFF
--- a/crates/gpui_web/src/events.rs
+++ b/crates/gpui_web/src/events.rs
@@ -64,6 +64,9 @@ impl WebWindowInner {
             self.register_dragleave(),
             self.register_key_down(),
             self.register_key_up(),
+            self.register_paste(),
+            self.register_copy(),
+            self.register_cut(),
             self.register_composition_start(),
             self.register_composition_update(),
             self.register_composition_end(),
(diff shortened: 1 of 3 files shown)


crates/gpui_web/src/events.rs
```diff
@@ -351,7 +354,22 @@ impl WebWindowInner {
...
+            // input handler.
+            let primary_modifier = if this.is_mac {
+                modifiers.platform
+            } else {
+                modifiers.control
+            };
+            let is_clipboard_shortcut =
+                primary_modifier && (key == "v" || key == "c" || key == "x");
+
+            if !is_clipboard_shortcut {
```
Yatekii, line 370: This feels very hacky and wrong.

crates/gpui_web/src/events.rs
```diff
@@ -434,6 +452,65 @@ impl WebWindowInner {
...
     }
 
+    fn register_paste(self: &Rc<Self>) -> Closure<dyn FnMut(JsValue)> {
+        let this = Rc::clone(self);
+        self.listen_input("paste", move |event: JsValue| {
+            let event: web_sys::ClipboardEvent = event.unchecked_into();
+            let Some(data) = event.clipboard_data() else {
+                return;
+            };
+            let Ok(text) = data.get_data("text/plain") else {
```
Yatekii, line 462: Should we handle other data formats or do we not care?

crates/gpui_web/src/events.rs
```diff
@@ -434,6 +452,65 @@ impl WebWindowInner {
         })
     }
 
+    fn register_paste(self: &Rc<Self>) -> Closure<dyn FnMut(JsValue)> {
+        let this = Rc::clone(self);
+        self.listen_input("paste", move |event: JsValue| {
+            let event: web_sys::ClipboardEvent = event.unchecked_into();
+            let Some(data) = event.clipboard_data() else {
+                return;
```
Yatekii, line 460: Should we add more logging if these fail or are we okay to silently continue?

SomeoneToIgnore:
@cla-bot check

PR #54869, opened 2026-04-25
docs: Update actions format

lingyaochu:
Self-Review Checklist:

Change the actions in docs to adopt the right format.

Release Notes:

- N/A


MrSubidubi (changes requested):
Thanks for this, good find! Could we instead please use the [docs preprocessor syntax](https://github.com/zed-industries/zed/blob/166eb1e28fb25e33d25a3790a430d7e943885f08/docs/README.md?plain=1#L47) for this to ensure these are always rendered properly and exist?

lingyaochu:
I noticed that some other actions also don't follow the format. Should I fix them as well while I'm at it? It should be a very straightforward review.

MrSubidubi:
Sure, feel free to!

lingyaochu:
> It should be a very straightforward review.

I take that back—it's not as straightforward as I expected. I found quite a few cases, some of them outdated. I've addressed everything I found via regex, but I'm not entirely sure if anything is left.

Additionally, I noticed that some platform-specific actions, such as `projects: open wsl`, are missing from the [All actions](https://zed.dev/docs/all-actions) table. It's worth another look (an issue #55376 is created for this), but I’ll leave it out of this PR to keep things focused.

MrSubidubi (changes requested):
Still a very nice change and the other stuff is definitely something worth looking into after.

However, could you please look into removing the backticks everywhere, please? The syntax itself will ensure that the keybinds will be rendered in monospace, so we can safely remove the backticks. Thanks!

Also, definitely would like to note that
> but I’ll leave it out of this PR to keep things focused.

is a very good idea and approach and I appreciate it very much, thanks!

lingyaochu:
> However, could you please look into removing the backticks everywhere, please?

Done

MrSubidubi (approved):
Very very very nice, thank you so much!

lingyaochu:
Sorry, there are some style issues, should be resolved now.

MrSubidubi:
All good, no need to apologize, that happens too often to me too.

PR #54884, opened 2026-04-25
Make terminal hyperlinks display correctly with changing content

### LINKED ISSUE #31866: Very minor visual bug when ctrl-hovering over a link in the terminal while the terminal is printing lines
### Summary

The underline of a link travels with the terminal while it's printing lines.

### Description

Steps to reproduce:
1. Run a script that prints something to the terminal with file links in it
2. While it's busy printing, scroll up a bit so the text doesn't move anymore
3. Hover over a link with ctrl pressed.
4. The underline will move down as long as the terminal is still printing stuff

Expected Behavior: The underline stays where the cursor is
Actual Behavior: The underline travels down

<details><summary>Video showing this behaviour</summary>

https://github.com/user-attachments/assets/4dbb26eb-71b5-4c7d-9543-ccaab8726ac0

There are no logs in Zed.log while this is happening.

</details>

---

Reproduction steps from #35352:

1. Open terminal in Zed
2. Create an executable script file named `terminal_hyperlink_bug` with the following contents
```
#!/usr/bin/env bash

for ((i = 0; i < 1000; i++)); do
    echo;
done;

for ((i = 0; i < 10; i++)); do
    echo "terminal_hyperlink_bug";
    sleep 5;
    echo "Noooo not a hyperlink!";
    sleep 5;
done;
```
2. Execute the script, hover the mouse over `terminal_hyperlink_bug` in the output, and hold down Cmd.

```
Zed: v0.190.0 (Zed Dev d3bc561f26e8077a1de8ffb98b7d9e012133980c) 
Memory: 30.5 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3080 || NVIDIA || 570.153.02
```

davewa:
- Closes #31866

Currently, `Terminal::alacritty::make_content()` always carries over any existing hovered word. With this PR we now:
- *Are more selective*: Only carry forward the hovered word if the terminal grid shape and visible lines have not changed
- *Correctly handle a shifting viewport*: If the grid shape and visible lines have not changed, but new lines were added to the terminal we carry the hovered word forward, and adjust the lines of the `word_match` so that the original text remains hyperlinked.

Also, this PR relaxes the test for displaying a hyperlink in `TerminalElement::prepaint`. We now display hyperlinks as long as the `hovered_word.id`s from terminal and terminal_view match. Previously we required all fields to be equal, which resulted in the hyperlink flickering when scrolled back with new lines being added.

Release Notes:

- Terminal: Made hyperlinks display correctly with changing content


crates/terminal/src/terminal.rs
```diff
@@ -3273,70 +3332,519 @@ mod tests {
...
+                        term.bounds_to_string(
+                            AlacPoint::new(Line(line), Column(0)),
+                            AlacPoint::new(Line(line), Column(columns)),
+                        ),
+                    )
+                })
+                .collect()
+        }
+
+        fn unthrottle(terminal: &mut Terminal) {
```
davewa: I'm not too happy with this... I'm tempted to make hyperlink throttling a setting, but that feels like too much scope creep for this PR and may not be something we want. Removing hyperlink throttling entirely for testing also seems undesirable because it should have it's own test coverage. So maybe this is acceptable for now.

davewa: Ended up going with test-only one-time throttle suppression flag.

crates/terminal/src/terminal.rs
```diff
@@ -1826,6 +1847,13 @@ impl Terminal {
         }
     }
 
+    fn clear_hyperlink(&mut self, cx: &mut Context<Self>) {
+        if self.last_content.last_hovered_word.is_some() {
+            self.last_content.last_hovered_word = None;
```
osiewicz:
```suggestion
        if self.last_content.last_hovered_word.take().is_some() {
```

davewa: Nice, thanks!

smitbarmase (approved):
Thanks, I tested the hover behavior and it works as expected.

I only updated the hyperlink tests to use a display-only terminal so they no longer launch a real shell and is more deterministic. Thanks again!

PR #54886, opened 2026-04-25
Implement luminance-based glyph dilation for macOS

reflectronic:
Apple's text rendering stack dilates glyph outlines for text rendered with a light foreground color.  Zed doesn't consider this nuance today; we populate our atlas using glyphs rendered with a dark foreground color. This means that, particularly in dark themes, text in Zed looks thin and blurry, and doesn't match the look of native macOS applications.

This pull request replicates the native behavior of Core Graphics. Some reverse-engineering revealed that CG computes the foreground color luminance using the Rec. 709 formula ($Y=0.2126R + 0.7152B + 0.0722G$) and quantizes it into five levels (0, 0.25, 0.5, 0.75, and 1). Each level uses a different dilation factor. 

With this patch, we calculate this same luminance bucket and supply it as the foreground color during rasterization. The correct dilation will be applied, and we'll store this glyph in the atlas keyed by this luminance bucket. So, we'll generate and use up to 5 different bitmaps for each glyph based on its foreground color.

I've confirmed that text rendered by Zed now exactly matches native applications like Safari, TextEdit, etc.

Release Notes:

- Improved text rendering clarity on macOS, particularly in dark themes.

### DIFF
--- a/crates/gpui/src/platform.rs
+++ b/crates/gpui/src/platform.rs
@@ -31,10 +31,10 @@ pub(crate) type PlatformScreenCaptureFrame = core_video::image_buffer::CVImageBu
 use crate::{
     Action, AnyWindowHandle, App, AsyncWindowContext, BackgroundExecutor, Bounds,
     DEFAULT_WINDOW_SIZE, DevicePixels, DispatchEventResult, Font, FontId, FontMetrics, FontRun,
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore (approved):

mikesun:
This is incredible work! Thank you! Finally don't have to use extra weight for my fonts!

draconacht:
god bless ive waited a year for this.

Issue #52263, opened 2026-03-23
Remote open hard-fails when cloud.zed.dev remote-server metadata lookup fails for known versions


iqdoctor:
### Summary

Opening a remote SSH project hard-fails if the local machine cannot fetch the remote-server metadata endpoint on `cloud.zed.dev`, even when the Zed version is already known and the release asset URL is deterministic.

Related issue: #46236 asks for clearer error messaging when proxy policy blocks the remote-server download URL. This issue is about the functional fallback gap, not just the message.

### Reproduction steps

1. Use `File -> Open Remote...` and connect to an SSH host.
2. Be on a network where the local machine cannot reach:
   `https://cloud.zed.dev/releases/<channel>/<version>/asset?asset=zed-remote-server&os=<os>&arch=<arch>`
   for example because a proxy/CDN rule blocks `cloud.zed.dev` or CONNECT tunneling to that host.
3. Try to open the remote project.

Example error from the dialog:

```text
Failed to connect
error sending request for url
.../cloud.zed.dev/releases/stable/0.228.0/asset?asset=...
client error (Connect): unexpected eof while tunneling.
```

### Actual behavior

Remote open aborts during the local metadata lookup. Zed never reaches a usable fallback path for the remote server binary, even though for stable/preview builds it already knows the exact version and the matching GitHub release asset can be derived.

### Expected behavior

If the remote-server metadata lookup fails for a known stable/preview version, Zed should fall back to the deterministic GitHub release asset URL instead of failing the whole remote-open flow.

For example:

- stable `0.228.0` linux x86_64 -> `https://github.com/zed-industries/zed/releases/download/v0.228.0/zed-remote-server-linux-x86_64.gz`
- preview `0.228.0` windows aarch64 -> `https://github.com/zed-industries/zed/releases/download/v0.228.0-pre/zed-remote-server-windows-aarch64.zip`

### Why this matters

Today remote development has an unnecessary hard dependency on `cloud.zed.dev` for a lookup that is redundant once the app version is known. In locked-down environments this blocks remote development entirely, even when the actual GitHub asset would be reachable.

### Code path

- `crates/remote_connection/src/remote_connection.rs`
- `crates/auto_update/src/auto_update.rs`
- `crates/remote/src/transport/ssh.rs`

The failing request happens in `AutoUpdater::get_release_asset()` via the `.../releases/<channel>/<version>/asset` endpoint.

PR #54898, opened 2026-04-25
remote: Fall back to GitHub remote-server assets

### LINKED ISSUE #54897: Remote SSH open fails when cloud.zed.dev remote-server metadata lookup is blocked
### Summary

Opening an SSH remote project can still fail before the remote server starts if the local Zed client cannot reach the `cloud.zed.dev` remote-server metadata endpoint, even when the Zed channel and version are already known and the matching `zed-remote-server` asset exists in the public GitHub release.

I can reproduce this on Zed Preview `0.234.6`. The failure is in the local client before the remote project opens.

### Reproduction steps

1. Use Zed Preview `0.234.6`.
2. Open `File -> Open Remote...` and connect to an SSH host that is otherwise reachable.
3. Use a network/proxy environment where CONNECT tunneling to `cloud.zed.dev` fails or is blocked, while GitHub release assets are reachable.
4. Try to open the remote project.

### Current behavior

Zed shows a critical connection failure and aborts the remote-open flow:

```text
Failed to connect

error sending request for url
.../cloud.zed.dev/releases/preview/0.234.6/asset?asset=...
client error (Connect): unexpected eof while tunneling
```

The important part is that the remote session fails during the release metadata lookup against this endpoint shape:

```text
https://cloud.zed.dev/releases/preview/0.234.6/asset?asset=zed-remote-server&os=<os>&arch=<arch>
```

For this build, the deterministic GitHub release assets already exist. For example, the `v0.234.6-pre` release contains remote-server assets such as:

```text
https://github.com/zed-industries/zed/releases/download/v0.234.6-pre/zed-remote-server-linux-x86_64.gz
```

So the failure is not that the remote-server artifact is unavailable. The failure is that remote development has a hard local dependency on the `cloud.zed.dev` metadata lookup before it can use an asset URL that can be derived from the already-known stable/preview version.

### Expected behavior

For stock Zed stable/preview builds, when Zed already knows the exact app version and is requesting `zed-remote-server`, a failure or non-success response from the `cloud.zed.dev/releases/.../asset` metadata endpoint should not necessarily abort remote open.

Zed can derive the matching GitHub release asset URL and continue, for example:

```text
preview 0.234.6 + linux x86_64
-> https://github.com/zed-industries/zed/releases/download/v0.234.6-pre/zed-remote-server-linux-x86_64.gz
```

This fallback should be narrow:

- only for `zed-remote-server`, not the normal app updater asset
- only when the version is known, not for `latest`
- only for stable/preview channels, not dev/nightly
- only for stock Zed release hosts, so custom packager overrides keep their current behavior

### Why this matters

In corporate and restricted-network environments, `cloud.zed.dev` and GitHub release assets may not have the same proxy/CDN policy. If `cloud.zed.dev` CONNECT tunneling fails, Zed currently blocks remote development entirely even though the matching remote-server binary may be available through GitHub Releases.

The documented `upload_binary_over_ssh` path does not fully avoid this failure class, because the local download path also starts by asking the same release metadata endpoint for the `zed-remote-server` asset before uploading it over SSH.

### Current code path

In current `main`, the SSH remote setup still reaches the same mandatory metadata lookup path:

- `crates/remote/src/transport/ssh.rs`: `ensure_server_binary()` calls `delegate.get_download_url(...).await?` before trying to download the remote-server binary on the host.
- `crates/remote_connection/src/remote_connection.rs`: the delegate resolves that URL through `AutoUpdater::get_remote_server_release_url()`.
- `crates/auto_update/src/auto_update.rs`: `get_remote_server_release_url()` calls `AutoUpdater::get_release_asset()` with `asset = "zed-remote-server"`.
- `get_release_asset()` builds `/releases/{channel}/{version}/asset` and propagates request/status failures instead of falling back to a deterministic GitHub release asset URL.

That means a local `cloud.zed.dev` metadata lookup failure can abort SSH remote open before Zed reaches a usable fallback path.

Zed: `0.234.6` Preview

Network condition: local client is behind a proxy/restricted network where CONNECT tunneling to `cloud.zed.dev` fails with `unexpected eof while tunneling`.

### Related previous issue

- #52263

iqdoctor:
Fixes #54897.

## Summary

Remote SSH setup currently treats the `cloud.zed.dev/releases/.../asset` metadata lookup as mandatory for `zed-remote-server` downloads. If that local metadata request fails, remote open aborts even when Zed already knows the stable or preview version and the matching GitHub release asset URL can be derived.

This adds a narrow fallback for `zed-remote-server` assets:

- the cloud metadata endpoint remains the primary source
- if that request fails or returns a non-success status
- and the requested asset is `zed-remote-server`
- and the version is known
- and the channel is stable or preview
- and the client is using a stock Zed base URL

then Zed derives the matching GitHub release asset URL and continues.

## Why

In restricted corporate/proxy networks, `cloud.zed.dev` and GitHub Releases can be subject to different allowlists or CDN policies. A CONNECT failure to `cloud.zed.dev` currently blocks SSH remote development entirely, even though the remote-server binary is already present in the public release.

For example, Zed Preview `0.234.6` maps to GitHub release tag `v0.234.6-pre`, which contains deterministic remote-server assets such as `zed-remote-server-linux-x86_64.gz`.

`upload_binary_over_ssh` does not fully avoid this failure class because the local download path also starts by resolving the remote-server asset through the same metadata endpoint.

## Implementation notes

The fallback is intentionally limited to avoid changing the normal app updater or custom packaging behavior:

- no fallback for the `zed` app asset
- no fallback for `latest`
- no fallback for dev/nightly
- no fallback for custom base URLs
- no change to the existing metadata response parsing when the metadata endpoint succeeds

## Testing

- `rustfmt +1.94.1 --edition 2024 crates/auto_update/src/auto_update.rs`
- `cargo fmt -p auto_update -- --check`
- `git diff --check`
- `cargo test -p auto_update github_remote_server_release_asset -- --nocapture`

Release Notes:

- Fixed SSH remote development failing when the remote-server metadata lookup is blocked but the matching GitHub release asset is available.

### DIFF
--- a/crates/auto_update/src/auto_update.rs
+++ b/crates/auto_update/src/auto_update.rs
@@ -172,6 +172,35 @@ pub struct ReleaseAsset {
     pub url: String,
 }
 
+fn github_remote_server_release_asset(
+    release_channel: ReleaseChannel,
+    version: Option<&Version>,
+    asset: &str,
+    os: &str,
+    arch: &str,
+) -> Option<ReleaseAsset> {
+    if asset != "zed-remote-server" {
+        return None;
+    }
+
+    let mut version = version?.clone();
+    version.pre = semver::Prerelease::EMPTY;
+    version.build = semver::BuildMetadata::EMPTY;
+
+    let tag = match release_channel {
+        ReleaseChannel::Stable => format!("v{version}"),
+        ReleaseChannel::Preview => format!("v{version}-pre"),
+        ReleaseChannel::Dev | ReleaseChannel::Nightly => return None,
+    };
+    let extension = if os == "windows" { "zip" } else { "gz" };
+    let asset_name = format!("{asset}-{os}-{arch}.{extension}");
+
+    Some(ReleaseAsset {
+        version: version.to_string(),
+        url: format!("https://github.com/zed-industries/zed/releases/download/{tag}/{asset_name}"),
+    })
+}
+
 struct MacOsUnmounter<'a> {
     mount_path: PathBuf,
     background_executor: &'a BackgroundExecutor,
@@ -589,6 +618,7 @@ impl AutoUpdater {
         cx: &mut AsyncApp,
     ) -> Result<ReleaseAsset> {
         let client = this.read_with(cx, |this, _| this.client.clone());
+        let requested_version = version.clone();
 
         let (system_id, metrics_id, is_staff) = if client.telemetry().metrics_enabled() {
             (
@@ -608,6 +638,10 @@ impl AutoUpdater {
             "latest".to_string()
         };
         let http_client = client.http_client();
+        let allow_github_fallback = matches!(
+            http_client.base_url().as_str(),
+            "https://zed.dev" | "https://staging.zed.dev"
+        );
 
         let path = format!("/releases/{}/{}/asset", release_channel.dev_name(), version,);
         let url = http_client.build_zed_cloud_url_with_query(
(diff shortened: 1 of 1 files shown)


iqdoctor:
I have now signed the Zed CLA. @cla-bot check

iqdoctor:
Signed

iqdoctor:
@cla-bot check

maxdeviant:
Thank you for the PR, but as explained in https://github.com/zed-industries/zed/issues/52263#issuecomment-4189432493, we don't want to add a fallback for downloading assets.

We don't want to embed knowledge about the underlying infrastructure for our updates—which could change in the future—into Zed.

PR #9471, opened 2024-03-18
Allow specifying no base keymap

zaucy:
This PR is a bit of a shot in the dark. I'm not sure if this will be acceptable and I understand if it gets rejected.

I've been trying to integrate Zed as my daily driver and the key bindings have been a major hurdle for me. Mostly due to the windows/linux keybindings being messed up, but also me wanting to have more chained key bindings similar to helix or common in custom neovim configurations.

I think having a `None` base keymap would allow someone to more easily implement a new base keymap (#4642) and would make my daily use of Zed a little nicer 😅.

Also I am aware that there would need to be a little more work done in this PR for the other base keymaps such as 'atom' since they assume the 'default' (vscode) base keymaps are loaded. I'm happy to do that work if a 'none' base keymap is acceptable.

Release Notes:

- Added ability to specify no base keymap which allows for full keybinding customization

PR #54899, opened 2026-04-25
Enable Vim/Helix keybindings when the base keymap is set to None

profojak:
Hi! 👋 When `base_keymap` is set to `None`, it disables all the keybindings, even if `vim_mode` or `helix_mode` is enabled. However, I think the Vim/Helix keybindings should be applied on top of the empty base keymap.

My use case for this is to start with the minimal set of Vim/Helix key shortcuts and add other bindings on top of that, instead of flooding the keymap with hundreds of predefined shortcuts from some base keymap.

### DIFF
--- a/crates/zed/src/zed.rs
+++ b/crates/zed/src/zed.rs
@@ -2074,16 +2074,16 @@ fn reload_keymaps(cx: &mut App, mut user_key_bindings: Vec<KeyBinding>) {
 
 pub fn load_default_keymap(cx: &mut App) {
     let base_keymap = *BaseKeymap::get_global(cx);
-    if base_keymap == BaseKeymap::None {
-        return;
-    }
+    if base_keymap != BaseKeymap::None {
+        cx.bind_keys(
+            KeymapFile::load_asset(DEFAULT_KEYMAP_PATH, Some(KeybindSource::Default), cx).unwrap(),
+        );
 
-    cx.bind_keys(
-        KeymapFile::load_asset(DEFAULT_KEYMAP_PATH, Some(KeybindSource::Default), cx).unwrap(),
-    );
-
-    if let Some(asset_path) = base_keymap.asset_path() {
-        cx.bind_keys(KeymapFile::load_asset(asset_path, Some(KeybindSource::Base), cx).unwrap());
+        if let Some(asset_path) = base_keymap.asset_path() {
+            cx.bind_keys(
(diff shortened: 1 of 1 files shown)


profojak:
@cla-bot check

NeelChotai (approved):
Nice, thanks for the contribution!

xab3r:
I haven't seen this coming. Wouldn't it be more appropriate to add another option like `base_keymap: vim/helix` or something if someone wants vim/helix keymap as default mapping, and keep `base_keymap: None` to behave exactly as it says - "no base map"?

@NeelChotai The "None" map was introduced on purpose in #9471 specifically to clear the map entirely. This change is breaking it. Can it be reverted?
cc @ConradIrwin 

P.S. My situation is this: the Zed setup is based on a clean base map, and all my keybindings are non-standard. This change introduces many unwanted mappings and makes the setup much harder to maintain.

ConradIrwin:
Sorry about that, revert is here: https://github.com/zed-industries/zed/pull/55144

profojak:
Thanks for your feedback @xab3r!

I submitted this pull request because it felt like a bug: you toggle the Vim/Helix mode while having base keymap set to `None`, and it does literally nothing. From the current UI, I would expect that to apply Vim/Helix key bindings *on top* of the emtpy base keymap. I did not realize that, in fact, Vim/Helix mode != Vim/Helix key bindings.

I agree a Vim/Helix base keymap is a good compromise, but I think the UI also has to change to communicate this effectively. Imagine you set the keymap to Vim/Helix, but do not really have Vim/Helix behavior, until you also toggle the Vim/Helix mode? To me, from a UI perspective, it makes much more sense to *apply* Vim/Helix key binds *on top* of empty base keymap.

I would propose something like `Disable all default key binds` override toggle instead of relying on the `None` base keymap to disable key binds, even when Vim/Helix mode is set. Should I open an Issue for this @ConradIrwin?

ConradIrwin:
@profojak vim/helix mode rely on a number of keybindings from the default keymap; so I think that having none mean none is the right approach.

What are you trying to do?

(In hindsight, I think it was a mistake to allow none at all, but seems like that ship has sailed and people are relying on it).

profojak:
@ConradIrwin I wanted to turn off all the key bindings and only enable in-editor Vim/Helix key shortcuts. Setting base keymap to `None` and enabling Vim/Helix mode kind of implied that to me, in my head. But I can work around this, I actually think having `None` base keymap is a good thing!

PR #54907, opened 2026-04-26
svg_preview: Add zoom and pan support

### LINKED ISSUE #54802: Zoom/Pan not possible in SVG Preview (only other images)
### Reproduction steps

Open the SVG Preview of an SVG and try zooming

Same behavior as any other image.

Zed: v0.233.9+stable.248.236e6ddf54e0763d0343f4d3e369d2ae00fe9670 (Zed) 
OS: macOS 26.4.1
Memory: 32 GiB
Architecture: aarch64

OmChillure:
Adds zoom and pan to the SVG preview, matching the behavior of the
image viewer for PNG/JPG/GIF. Previously, opening the SVG preview
showed the rasterized image with no way to interact with it.

#### Changes:
- Added `svg::ZoomIn`, `svg::ZoomOut`, `svg::ResetZoom`, `svg::FitToView`, and `svg::ZoomToActualSize` actions.
- Added zoom/pan state to `SvgPreviewView` (`zoom_level`, `pan_offset`,  `last_mouse_position`, `container_bounds`) along with handlers for scroll wheel (Ctrl/Cmd to zoom toward cursor, plain scroll to pan), trackpad pinch, and click-and-drag pan with cursor swap (open-hand ↔ closed-hand).
- Introduced a `SvgContentElement` that captures container bounds during prepaint, so the rendered SVG can be centered, scaled by `zoom_level`, and offset by `pan_offset`. The first layout auto-fits the SVG to the viewport.
- Added `SvgPreviewToolbarControls` — a toolbar item that appears in the right slot when an SVG preview is active, showing `−`, the current zoom percentage (click to reset), `+`, and a fit-to-view button, all with action-aware tooltips.
- Implemented `breadcrumb_location` and `breadcrumbs` on the `SvgPreviewView` `Item` impl so the file name appears on the left side of the toolbar, mirroring the image viewer.
- Registered the new toolbar item in `zed.rs` next to `ImageViewToolbarControls`.
- Added a `SvgPreview` keymap context with bindings for the new actions in `default-macos.json`, `default-linux.json`, and `default-windows.json` — same shortcuts as the image viewer (`cmd/ctrl + =/+/-/0/1/shift-0`).

#### Self-Review Checklist:

- [x] Performance impact has been considered and is acceptable.

#### Closes #54802

Video :

[Screencast from 2026-04-26 12-00-42.webm](https://github.com/user-attachments/assets/b9238690-8b5b-4621-9d64-3a4f81405208)

Release Notes:

- Fixed zoom and pan not working in the SVG preview.

### DIFF
--- a/assets/keymaps/default-linux.json
+++ b/assets/keymaps/default-linux.json
@@ -1532,6 +1532,17 @@
     },
   },
   {
+    "context": "SvgPreview",
+    "bindings": {
+      "ctrl-=": "svg::ZoomIn",
+      "ctrl-+": "svg::ZoomIn",
+      "ctrl--": "svg::ZoomOut",
+      "ctrl-0": "svg::ResetZoom",
+      "ctrl-1": "svg::ZoomToActualSize",
+      "ctrl-shift-0": "svg::FitToView",
+    },
+  },
+  {
     "context": "RunModal",
     "bindings": {
       "alt-1": "new_process_modal::ActivateTaskTab",
--- a/assets/keymaps/default-macos.json
+++ b/assets/keymaps/default-macos.json
@@ -1594,6 +1594,18 @@
     },
   },
   {
+    "context": "SvgPreview",
+    "use_key_equivalents": true,
+    "bindings": {
+      "cmd-=": "svg::ZoomIn",
+      "cmd-+": "svg::ZoomIn",
+      "cmd--": "svg::ZoomOut",
+      "cmd-0": "svg::ResetZoom",
+      "cmd-1": "svg::ZoomToActualSize",
+      "cmd-shift-0": "svg::FitToView",
+    },
+  },
+  {
     "context": "RunModal",
     "bindings": {
       "cmd-1": "new_process_modal::ActivateTaskTab",
--- a/assets/keymaps/default-windows.json
+++ b/assets/keymaps/default-windows.json
@@ -1512,6 +1512,17 @@
     },
   },
   {
+    "context": "SvgPreview",
+    "bindings": {
+      "ctrl-=": "svg::ZoomIn",
+      "ctrl-+": "svg::ZoomIn",
+      "ctrl--": "svg::ZoomOut",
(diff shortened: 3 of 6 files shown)


OmChillure:
A couple of items from the self-review checklist I'd like guidance on:

1] **Tests:** I didn't add tests because the reference `image_viewer` crate has zero tests for the identical zoom/pan logic (`compute_fit_to_view_zoom`, `set_zoom`, scroll/pinch/drag handlers). Adding tests only for the SVG
copy would set a higher bar here than exists for the original. Happy to add them if you'd like.

2] **Performance / zoom quality:** The current approach scales the already-rasterized bitmap, matching what `image_viewer` does for PNG/JPG. For raster files that's fine, but for SVG it means deep zoom-in is blurry — we lose the vector-quality benefit. The clean fix would be to pass `zoom_level` as the `SCALE_FACTOR` into `cx.svg_renderer().render_single_frame(...)` and re-rasterize on zoom change (debounced so we don't re-render on every wheel tick).

I left it out of this PR to keep the scope tight. Would you prefer I add it here, or open a follow-up issue/PR? 

3] **Duplicated Code:** There's now ~200 LoC of duplication between `image_viewer` and `svg_preview` (zoom constants, state methods, input handlers, custom Element scaffold, toolbar). Both viewers work correctly, but if a third zoomable viewer ever appears (PDF, plot, etc.) it'd be worth extracting a shared `zoom_pan` crate. I held off
because at N=2 it felt premature, but happy to do it now if you'd prefer.

smitbarmase (changes requested):
Thanks. Few suggestions:

- Yes, I think preserving vector quality should be part of adding zoom support to SVG rather than a follow-up.
- No need to extract the shared code between `image_viewer` and `svg_preview`.

smitbarmase:
To keep the PR queue cleaner, will close this PR for now, as there have been no changes for 3 weeks since the review. Happy to review a new PR that addresses the suggestions above. Thank you for working on this!

PR #54910, opened 2026-04-26
open_ai: Use responses API for all models

bennetbo:
From the [docs](https://developers.openai.com/api/docs/guides/migrate-to-responses#responses-benefits):

> Better performance: Using reasoning models, like GPT-5, with Responses will result in better model intelligence when compared to Chat Completions. Our internal evals reveal a 3% improvement in SWE-bench with same prompt and setup.
Agentic by default: The Responses API is an agentic loop, allowing the model to call multiple tools, like web_search, image_generation, file_search, code_interpreter, remote MCP servers, as well as your own custom functions, within the span of one API request.
Lower costs: Results in lower costs due to improved cache utilization (40% to 80% improvement when compared to Chat Completions in internal tests).
Stateful context: Use store: true to maintain state from turn to turn, preserving reasoning and tool context from turn-to-turn.
Flexible inputs: Pass a string with input or a list of messages; use instructions for system-level guidance.
Encrypted reasoning: Opt-out of statefulness while still benefiting from advanced reasoning.
Future-proof: Future-proofed for upcoming models.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Always use Responses API for OpenAI models

### DIFF
--- a/crates/open_ai/src/open_ai.rs
+++ b/crates/open_ai/src/open_ai.rs
@@ -272,18 +272,30 @@ impl Model {
         }
     }
 
-    pub fn supports_chat_completions(&self) -> bool {
+    pub fn uses_responses_api(&self) -> bool {
         match self {
             Self::Custom {
                 supports_chat_completions,
                 ..
-            } => *supports_chat_completions,
-            Self::FiveCodex
+            } => !*supports_chat_completions,
+            Self::ThreePointFiveTurbo | Self::Four | Self::FourTurbo => false,
+            Self::FourOmniMini
+            | Self::FourPointOneNano
+            | Self::O1
+            | Self::O3
+            | Self::O3Mini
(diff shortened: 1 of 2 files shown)


crates/open_ai/src/open_ai.rs
```diff
@@ -272,17 +272,12 @@ impl Model {
         }
     }
 
-    pub fn supports_chat_completions(&self) -> bool {
+    pub fn uses_responses_api(&self) -> bool {
         match self {
             Self::Custom {
                 supports_chat_completions,
```
bennetbo, line 278: We should probably rename this to `use_chat_completions` or something, but we need a settings migration for that, so I left it as is for now so we can cherry-pick this to Preview/Stable

benbrandt (approved):

PR #54620, opened 2026-04-23
Fix Settings UI silently dropping failed writes (no user-facing error)

galuis116:
Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #54617

Release Notes:

- Fixed 
Settings UI now surfaces save failures directly in the UI instead of silently logging them.
When a settings write fails (for example due to file permissions, path issues, or write conflicts), Settings now shows a warning banner with details and a dismiss action.

PR #54923, opened 2026-04-26
settings_ui: Rebuild languages list when extensions change

galuis116:
Self-Review Checklist:

Closes #54922 

Release Notes:

- Settings: The Languages list under Languages & Tools now updates when you install or uninstall a language extension, even if the Settings window was already open. You no longer need to close and reopen Settings to see new per-language Configure entries.


joaovictor712:
waiting to merge as all has been passed.

joaovictor712:
Hello @maxdeviant   and @MrSubidubi ,    wating to merge

SomeoneToIgnore:
~https://github.com/zed-industries/zed/pull/54620#issuecomment-4348166707~
https://github.com/zed-industries/zed/pull/54620#issuecomment-4351268088

MrSubidubi:
Thanks for this! I think if anyone were to pick this up again, a better approach forward would be be a bit more surgical here and perhaps check what could be persisted as opposed to doing a full rebuild. Will close for now given other activity on the repository, but I think this would be a nice fix (and might be the right one after all, would still be better though if we were to double-check whether there is no better way)

PR #54943, opened 2026-04-27
json_schema_store: Match symlinked settings/keymap files for autocomplete

### LINKED ISSUE #54888: JSON autocomplete/suggestions don't work in `settings.json` and `keymap.json` when file is a symlink
When `~/.config/zed/settings.json` is a symlink to a file elsewhere on disk, JSON autocomplete and suggestions stop working in the settings file. Replacing the symlink with the actual file restores them.

The settings themselves still load and apply correctly — only the editor experience in the file is affected.

### Steps to reproduce

1. Close Zed
2. Move `~/.config/zed/settings.json` somewhere else (e.g. `~/zed-settings.json`)
3. Symlink it back: `ln -s ~/zed-settings.json ~/.config/zed/settings.json`
4. Open Zed 
5. zed: open settings file

### Expected

The `zed://schemas/settings` schema is applied and autocomplete works, same as with a regular file.

### Actual

No completions are offered.

### Suspected cause

The schema's `fileMatch` rule for `zed/settings.json` is likely being checked against the resolved/canonical path instead of the opened path, so symlinked settings files no longer match the rule.

### Context

This setup (symlinking settings into `~/.config/zed/` from a Git repo to sync across machines) worked for a long time — the missing autocomplete is a recent regression.

-----

**Edit**: I just found that it’s the same case for `keymap.json`; symlinks also break autocomplete suggestions in that file.

mvanhorn:
## Closes #54888

When `~/.config/zed/settings.json` (or `keymap.json`, `tasks.json`, etc.) is a symlink, JSON autocomplete and schema-aware suggestions stop working. The settings still apply because the file content is read normally, only the editor experience breaks.

## Root cause

`schema_file_match` returns a single suffix string built from the path-as-given. The JSON LSP matches that string against `fileMatch` globs like `**/*settings.json`, but when Zed opens a symlinked path, the buffer's URL ends up canonicalized to the real file. The original path the schema registered against and the URL the LSP sees diverge, so the schema never binds.

## Change

Replace `schema_file_match` (returns one `String`) with `schema_file_match_entries` (returns `Vec<String>`). For non-symlinked paths the vector has one entry, identical to the previous output. When the input path is a symlink, the vector also contains the canonicalized path so either form binds:

```rust
fn schema_file_match_entries(path: &std::path::Path) -> Vec<String> {
    let mut out = Vec::with_capacity(2);
    out.push(stripped_match(path));
    if let Ok(canonical) = path.canonicalize() {
        if canonical != path {
            out.push(stripped_match(&canonical));
        }
    }
    out
}
```

The five call sites that registered schemas (settings, keymap, tasks, debug_scenarios, snippets) now flatten this vector into the schema entry.

## Tests

`cargo test -p json_schema_store` covers three new units:
- `schema_file_match_entries_returns_single_entry_for_regular_file` (regression guard)
- `schema_file_match_entries_returns_canonical_for_symlink` (the fix; canonicalizes the tmp root so it works on macOS where `/var/folders` symlinks to `/private/var/folders`)
- `stripped_match_normalizes_separators` (Windows backslash handling)

## Verification

- [x] `cargo fmt --check`
- [x] `cargo clippy -p json_schema_store` clean
- [x] `cargo build`
- [x] `cargo test -p json_schema_store` (3 passed)

🤖 Built with assistance from Claude Code and Codex.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9268,6 +9268,7 @@ dependencies = [
  "settings",
  "snippet_provider",
  "task",
+ "tempfile",
  "theme",
  "util",
 ]
--- a/crates/json_schema_store/Cargo.toml
+++ b/crates/json_schema_store/Cargo.toml
@@ -17,6 +17,7 @@ default = []
 [dev-dependencies]
 gpui = { workspace = true, features = ["test-support"] }
 settings = { workspace = true, features = ["test-support"] }
+tempfile.workspace = true
 
 [dependencies]
 anyhow.workspace = true
--- a/crates/json_schema_store/src/json_schema_store.rs
+++ b/crates/json_schema_store/src/json_schema_store.rs
@@ -449,45 +449,48 @@ pub fn all_schema_file_associations(
         .flat_map(|(_, glob_strings)| glob_strings)
         .cloned();
     let jsonc_globs = extension_globs.chain(override_globs).collect::<Vec<_>>();
+    let settings_file_matches = schema_file_match_entries(paths::settings_file());
+    let keymap_file_matches = schema_file_match_entries(paths::keymap_file());
+    let mut tasks_file_matches = schema_file_match_entries(paths::tasks_file());
+    tasks_file_matches.push(
+        paths::local_tasks_file_relative_path()
+            .as_unix_str()
+            .to_string(),
+    );
+    let mut debug_file_matches = schema_file_match_entries(paths::debug_scenarios_file());
+    debug_file_matches.push(
+        paths::local_debug_file_relative_path()
+            .as_unix_str()
+            .to_string(),
+    );
+    let snippet_file_matches =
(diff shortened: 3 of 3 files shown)


everdrone:
Hello!
What is blocking this PR? Would love to have autocomplete back on windows.

mvanhorn:
@everdrone nothing's blocking on my side - the PR has been sitting waiting for a maintainer review since April and hasn't picked up any reviewers yet. CI is green and the diff is small (+97/-23). Happy to rebase if it goes stale. If you want to nudge it along, a +1 reaction on the top comment usually helps surface it in the team's queue.

everdrone:
@mvanhorn yeah I already +1'd it, trying to get someone on zed's team attention

nicojerome:
Please Zed team, I want my autocomplete back

PR #54983, opened 2026-04-27
gpui_macos: Fix unintended modifier-only shortcuts when typing with an input method

najeira:
## Summary

By ignoring key events during IME input, unintended modifier-only shortcuts are triggered. This fix will correctly handle key events during IME input to prevent it.

## Details

When an input method is not involved, typing `shift+a shift+a` to type `AA` does not trigger `shift shift`. The `a` key down events reach GPUI, so `ModifierState.saw_keystroke` is set to true and the Shift presses are not treated as standalone modifier keystrokes.

When an input method handles a character key down directly and does not route it back to GPUI as a regular `KeyDownEvent`, `ModifierState.saw_keystroke` can remain false. In that case, the Shift portions of `shift+a shift+a` can be interpreted as `shift shift`.

This change notifies GPUI with `KeyDownHandledByInputMethod` when an input method handles a key down. The event is not dispatched as a keybinding; the receiver only sets `ModifierState.saw_keystroke` to true.

## Testing

Added a GPUI test covering `KeyDownHandledByInputMethodEvent` and modifier-only shortcut detection.

## Extra

Self-Review Checklist:

Release Notes:

- Fixed unintended modifier-only shortcuts when typing with an input method.

### DIFF
--- a/crates/gpui/src/interactive.rs
+++ b/crates/gpui/src/interactive.rs
@@ -57,6 +57,17 @@ impl InputEvent for KeyUpEvent {
 }
 impl KeyEvent for KeyUpEvent {}
 
+/// The key down event handled by the input method equivalent for the platform.
+#[derive(Clone, Debug)]
+pub struct KeyDownHandledByInputMethodEvent;
+
+impl Sealed for KeyDownHandledByInputMethodEvent {}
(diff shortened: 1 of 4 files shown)


najeira:
I rebased this PR to keep up with main.

PR #54984, opened 2026-04-27
extension_api: Add IPv6 support in DAP extensions API

tomhoule:
Fix forward for #52244

Release Notes:

- N/A

### DIFF
--- a/crates/extension_api/wit/since_v0.8.0/dap.wit
+++ b/crates/extension_api/wit/since_v0.8.0/dap.wit
@@ -20,15 +20,23 @@ interface dap {
         attach(attach-request)
     }
 
+    type ipv4-address = tuple<u8, u8, u8, u8>;
+    type ipv6-address = tuple<u16, u16, u16, u16, u16, u16, u16, u16>;
+
+    variant ip-address {
+        ipv4(ipv4-address),
+        ipv6(ipv6-address),
+    }
+
     record tcp-arguments {
         port: u16,
-        host: u32,
+        host: ip-address,
         timeout: option<u64>,
     }
 
     record tcp-arguments-template {
         port: option<u16>,
-        host: option<u32>,
+        host: option<ip-address>,
         timeout: option<u64>,
     }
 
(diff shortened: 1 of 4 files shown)


MrSubidubi (changes requested):
Since v0.8.0 is unreleased, do we not want to look into just adding support here perhaps?

tomhoule:
Will look into that.

tomhoule:
Going with a structured variant for ip-address in the wit. Feedback appreciated (pushing soon).

tomhoule:
Pushing current state. Will do a more thorough self review and ping later.

crates/extension_host/src/wasm_host/wit/since_v0_6_0.rs
```diff
@@ -384,3 +383,324 @@ impl ExtensionImports for WasmState {
         latest::ExtensionImports::make_file_executable(self, path).await
     }
 }
+
+impl From<dap::TcpArguments> for latest::dap::TcpArguments {
```
tomhoule, line 387: Is this the right place for these conversions?

crates/extension_api/wit/since_v0.8.0/dap.wit
```diff
@@ -20,15 +20,23 @@ interface dap {
         attach(attach-request)
     }
 
+    type ipv4-address = tuple<u8, u8, u8, u8>;
```
tomhoule, line 23: Happy to replace this with u32 if that seems better.

crates/extension_host/src/wasm_host/wit/since_v0_6_0.rs
osiewicz: Yep.

crates/extension_api/wit/since_v0.8.0/dap.wit
osiewicz: I think it's fine as is. If we had a fixed-length list in WIT, I'd lean towards using those with both IPv4 and IPv6 instead, but, alas..

osiewicz (approved):

tomhoule:
@MrSubidubi I can't merge without the changes requested being resolved

MrSubidubi (approved):
Apologies!

tomhoule:
No worries, thanksyou!

PR #55008, opened 2026-04-27
markdown_preview: Keep content anchored when search bar opens

### LINKED ISSUE #54987: Search bars should not scroll down content on buffers or Markdown preview
### Reproduction steps

1. Open a Markdown file and open its preview tab (`markdown: open preview`).
2. Scroll partway down the document.
3. Press `Cmd-F` to open buffer search inside the preview.
4. Observe: the line that was at the top of the viewport before the search bar opened is now drawn at a different screen position (or hidden behind the toolbar, depending on toolbar height).

### Current vs. expected behavior

**Current:** When the search bar deploys, the pane's toolbar grows and the preview's viewport shrinks via flex layout. The preview keeps its scroll offset unchanged, so the same content is now drawn at a different screen position, visually sliding as the user toggles search. Navigating to a match whose position falls within the just-shrunk region can land the highlighted match underneath the search bar.

**Expected:** Toggling the search bar should keep the same content visible at the same screen position. The regular code editor already behaves this way: when its search bar deploys, `Editor::search_bar_visibility_changed` records the pre-toggle bounds and `autoscroll_vertically` shifts the scroll offset by the bounds delta on the next layout pass, anchoring visible content.

```
Zed: v0.233.10+stable.250.350f33871e84d5ed83a403523e9490b9bf014a94 (Zed)
OS: macOS 26.5 (Build 25F5058e)
Memory: 32 GiB
Architecture: aarch64
```

### Notes

Cmd-F support was added to the markdown preview in #52502. `MarkdownPreviewView` implements `SearchableItem` but inherits the no-op default for `search_bar_visibility_changed` from `crates/workspace/src/searchable.rs`, so it does not perform the same scroll-offset compensation as the editor (`crates/editor/src/items.rs` + `crates/editor/src/scroll/autoscroll.rs`).

tkilaker:
## Summary

- Adds `SearchableItem::search_bar_visibility_changed` to `MarkdownPreviewView`, mirroring the editor's `expect_bounds_change` pattern.
- Anchors visible content at the same screen y when the buffer search bar deploys or dismisses, instead of letting it slide by the toolbar's growth.
- The compensation is deferred via two chained `window.on_next_frame` callbacks because the first fires before the upcoming draw — the second sees the post-paint bounds. Subtracts the top delta from `ScrollHandle::offset.y` (which is non-positive when scrolled down) to scroll further down by exactly the toolbar's growth, holding content's screen position constant.

Same root cause as the editor's `crates/editor/src/scroll/autoscroll.rs:138` compensation — `ScrollHandle::offset.y` has the opposite sign convention from the editor's `scroll_position.y`, so we subtract instead of add.

Fixes #54987.

## Test plan

- [x] Compiles with `cargo clippy -p markdown_preview -- --deny warnings`.
- [x] Manual: open `Cmd-F` mid-document — content stays at same screen y; close + reopen returns offset to its original value (round-trip exact).
- [x] Manual: at top of doc (`offset.y == 0`), no compensation runs (mirrors the editor's guard).

Release Notes:

- Fixed markdown preview content shifting when toggling buffer search

### DIFF
--- a/crates/markdown_preview/src/markdown_preview_view.rs
+++ b/crates/markdown_preview/src/markdown_preview_view.rs
@@ -50,6 +50,7 @@ pub struct MarkdownPreviewView {
     image_cache: Entity<RetainAllImageCache>,
     base_directory: Option<PathBuf>,
     pending_update_task: Option<Task<Result<()>>>,
+    pending_anchor_top: Option<Pixels>,
     mode: MarkdownPreviewMode,
 }
 
@@ -244,6 +245,7 @@ impl MarkdownPreviewView {
                 image_cache: RetainAllImageCache::new(cx),
                 base_directory: None,
                 pending_update_task: None,
+                pending_anchor_top: None,
                 mode,
             };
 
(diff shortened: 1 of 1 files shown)


tkilaker:
@cla-bot check

PR #55009, opened 2026-04-27
markdown_preview: Shrink preview to fit when pane toolbar grows

tkilaker:
## Summary

- Changes the outer `MarkdownPreviewView` div from `.size_full()` to `.w_full().flex_1().min_h_0()`.
- Before: preview's height stayed at 100% of the pane regardless of toolbar size, so deploying the buffer search bar pushed the preview's bottom 45px past the pane edge — clipped by the pane's `overflow_hidden`.
- After: preview takes only the leftover main-axis space and shrinks as the toolbar grows. Bottom edge stays at the pane bottom.

Verified empirically: `bounds.size.height` round-trips `952px ↔ 907px` with the toolbar at `67 ↔ 112` (Δ = 45px = the search bar's height).

Companion to #55008 (anchor compensation). Both PRs come out of investigating #54987 — they fix two independent bugs that surface together when toggling the search bar.

## Test plan

- [x] Compiles with `cargo clippy -p markdown_preview -- --deny warnings`.
- [x] Manual: `Cmd-F` in the markdown preview no longer clips bottom content past the pane edge.
- [x] Manual: `Esc` restores the preview to full available height.

Release Notes:

- Fixed markdown preview content being clipped past the pane edge when the buffer search bar is open


tkilaker:
@cla-bot check

agu-z (approved):
Looks great! Thank you for your contribution

PR #55034, opened 2026-04-27
workspace: Persist recent navigation history across sessions

### LINKED ISSUE #56271: File finder recent files are lost after restarting a workspace
### Reproduction steps

1. Open a project in Zed.
2. Open a file via `cmd-p` / `ctrl-p`.
3. Close the file tab, or leave it open and continue working.
4. Open `cmd-p` / `ctrl-p` again and observe that the file appears in the recent file finder results.
5. Quit Zed.
6. Reopen the same project.
7. Open `cmd-p` / `ctrl-p` again.

Current behavior: File finder recent results are only remembered for the current session. After quitting and reopening Zed, recently opened files no longer appear in `cmd-p` /`ctrl-p` history. This is especially noticeable for files that are not part of the normal project index, such as gitignored files, because the user has to locate them manually again.

Expected behavior: Files the user explicitly opened recently should remain available in file finder recent results after restarting/reopening the workspace, as long as the file still exists and belongs to the reopened project.

This should not change which files Zed indexes by default or make all gitignored files searchable. It should only preserve the existing recent file finder history across restarts.

OS: macOS 26.4.1
Memory: 36 GiB
Architecture: aarch64

anilzeybek:
Saves the recent navigation history (up to 20 entries) to the workspace DB so it survives restarts. When reopening a workspace, persisted paths that resolve to the current project are merged into the file finder's history, deduplicated against files already opened in the session.

https://github.com/user-attachments/assets/3d8934e1-0da6-4445-8f3d-acc5597d7ac8

Self-Review Checklist:

Closes: https://github.com/zed-industries/zed/issues/56271

Release Notes:

- Improved file finder history to include recently opened files from previous sessions.


crates/workspace/src/persistence.rs
```diff
@@ -1036,6 +1053,7 @@ impl Domain for WorkspaceDb {
         sql!(
             ALTER TABLE workspaces ADD COLUMN identity_paths TEXT;
             ALTER TABLE workspaces ADD COLUMN identity_paths_order TEXT;
+            ALTER TABLE workspaces ADD COLUMN recent_navigation_history TEXT;
```
SomeoneToIgnore: This has to be a new migration, not a change of an existing one.

crates/file_finder/src/file_finder.rs
```diff
@@ -794,7 +794,7 @@ impl FoundPath {
     }
 }
 
-const MAX_RECENT_SELECTIONS: usize = 20;
+use workspace::MAX_RECENT_SELECTIONS;
```
SomeoneToIgnore: Why is this import is here?
We have them on the top.

crates/workspace/src/workspace.rs
```diff
@@ -6903,6 +6925,11 @@ impl Workspace {
                 let docks = build_serialized_docks(self, window, cx);
                 let window_bounds = Some(SerializedWindowBounds(window.window_bounds()));
                 let identity_paths_hint = self.project_group_key(cx).path_list().clone();
+                let recent_navigation_history = self
+                    .recent_navigation_history(Some(MAX_RECENT_SELECTIONS), cx)
```
SomeoneToIgnore: This runs on the main thread on every throttled serialize (tab switch, split, dock change, 200 ms throttle) and walks every pane's nav stacks — up to 3×1024 entries per pane, cloning a `(ProjectPath, Option<PathBuf>)` per entry, plus hash maps and a sort.

Previously that cost was only paid on file-finder open, but now this is quite hard to justify — let's make it more performant, please.

crates/workspace/src/workspace.rs
```diff
@@ -7091,6 +7120,8 @@ impl Workspace {
 
             // Remove old panes from workspace panes list
             workspace.update_in(cx, |workspace, window, cx| {
+                workspace.persisted_recent_navigation_history = recent_navigation_history;
```
SomeoneToIgnore: Is this a race?

The field is set after item restoration starts; restoration emits `AddItem` → serialize, which can overwrite the DB column without the persisted entries.
Killing Zed in that window loses them (wider window for remote projects, since `worktree_store.rs:463-486` resolves by prefix only).

SomeoneToIgnore (changes requested):
Thank you, this looks quite good.

One general question is whether we should use a separate table for this, similar to github.com/zed-industries/zed/pull/60908

Then we can solve all the cleanup issues via `ON DELETE CASCADE` native approach instead of a racy code we have now.

Blocking for now, as the existing migration is spoiled, which is a severe bug.

crates/workspace/src/workspace.rs
anilzeybek: Serialize no longer walks pane nav stacks. We keep a <= 20 path list in memory and only update it when the active project path actually changes (tab switch, breadcrumbs, etc.). Same-path cursor moves are just an equality check and don’t reserialize. Serialize clones that small list.

crates/workspace/src/workspace.rs
anilzeybek: Yeah you were right, that was a race. We now load the persisted paths and set `restoring_workspace` before deserialize, and leave it on through pane swap / dock restore so restore-time item activation can’t wipe the DB list.

anilzeybek:
Thanks for the review @SomeoneToIgnore!

Switched to a separate table as you suggested, in a new migration. Cleanup is now handled by `ON DELETE CASCADE`.

Can you take a look at the PR once again?

SomeoneToIgnore (approved):
Thank you!

Issue #52173, opened 2026-03-23
"Go to Definition" opens with target at the bottom of the viewport


injust:
### Reproduction steps

1. Open any Python file that imports something
2. Use "Go to Definition" (or cmd-click) on something that was imported

### Current vs. Expected behavior

With this code:

```python
from string import Formatter

print(Formatter)
```

cmd-clicking on `Formatter` opens the `__init__.pyi` file with the `Formatter` class 3 rows from the end:

<img width="746" height="343" alt="Image" src="https://github.com/user-attachments/assets/81a93325-bd26-4a5a-b284-ca1c057d1ee5" />

Instead, it should open 3 rows from the top or something.

### Zed version and system specs

Zed: v0.228.0+stable.203.8421009ef8a022df1196d54bb42fd94366ec0988 (Zed) 
OS: macOS 15.7.4
Memory: 16 GiB
Architecture: x86_64

dinocosta:
Thanks for reporting! 🙂

I was able to reproduce this on the latest stable release (0.231.2) on macOS. It's worth noting that this is likely related with the `vertical_scroll_margin` setting, as reported [here](https://github.com/zed-industries/zed/issues/21368#issuecomment-3145438491) and it only seems to manifest itself in case you haven't yet opened the specific file and scrolled during the session, otherwise it seems to avoid scrolling, as long as the target line is within view, as shown below ▼

https://github.com/user-attachments/assets/6a989bf0-1263-4e98-bdbc-1f1e4c302718

tcerqueira:
User feedback: imo it should scroll to the exact height as where you called the action. It minimizes the eye movement and avoids that half a second where I'm not sure where the cursor landed.

dinocosta:
> User feedback: imo it should scroll to the exact height as where you called the action. It minimizes the eye movement and avoids that half a second where I'm not sure where the cursor landed.

Yep, this seems to also make the most sense to me! Might not always be possible, for example, if the cursor is in the middle of the screen when the action is called but the symbol to which it is navigating to is at the start of a file.

AnkanMisra:
[...]

PR #55036, opened 2026-04-27
editor: Add preserve scroll strategy for go to definition

dinocosta:
These changes attempt to expand on the work introduced by https://github.com/zed-industries/zed/pull/54778 by introducing a new `GoToDefinitionScrollStrategy::Preserve` variant that attempts to keep the cursor at the same vertical offset within the viewport when navigating to a definition.

Most of the machinery for this was already in place. To support cases where the user's scroll position isn't snapped to an exact display row, for example, after scrolling with the mmouse, `Autoscroll::TopRelative` and `Autoscroll::BottomRelative` were updated from `usize` to `ScrollOffset`, allowing fractional offsets.

When the cursor is offscreen at the moment the `editor: go to definition` action is invoked, `Preserve` falls back to `Autoscroll::center`, matching the existing default for `go_to_definition_scroll_strategy`. This avoids attempting to preserve an offset where the cursor isn't visible which would lead to the cursor being offscreen when jumping to the definition.

Documentation has also been updated to reflect this new strategy value.

Self-Review Checklist:

Relates to #52173

Release Notes:

- Added a new `preserve` option to `go_to_definition_scroll_strategy` that keeps the cursor at the same vertical position within the viewport when navigating to a definition


dinocosta:
Short screen recording of this strategy in action ▼

https://github.com/user-attachments/assets/0add8f0d-a07b-49a4-8dc2-cd117e6b1be7

SomeoneToIgnore (approved):

glokta1:
IMO, this should be the default scroll strategy instead of center. It's easier on the eyes. Also mentioned by https://github.com/zed-industries/zed/issues/52173#issuecomment-4262335660

PR #55040, opened 2026-04-27
Avoid breaking YAML curly braces

### LINKED ISSUE #46632: Breaking YAML files containing multiple curly braces - report 2
### Reproduction steps

See this bug report: Breaking YAML files containing multiple curly braces #25536.

I hit the same bug today, on a brand new zed installation. This bug is especially easy to address with a documentation workaround IMHO. This bug causes poor usability IMHO especially for users who are encountering this kind of bug for the first time.

## To reproduce

Use any typical editor (not Zed) to create any kind of text yaml file that has curlies and a long line, such as:

```yaml
this_is_any_long_key: {{ get_env(name="THIS_IS_ANY_LONG_ENV", default="This is any long default") }}
```

Open the file with Zed, do any edit, and save; the output is reformatted like this, which is broken/invalid/wrong:

```yaml
this_is_any_long_key:
  {
    {
      get_env(name="THIS_IS_ANY_LONG_ENV",
      default="This is any long default"),
    },
  }
```

The best way I've found to solve this is simply to wrap the value in single quote marks:

this_is_any_long_key: '{{ get_env(name="THIS_IS_ANY_LONG_ENV", default="This is any long default") }}'

Problem: I understand this to be a defect in Prettier. However, it's also IMHO a kind of glitch that's counter-intuitive e.g. I use many editors, and use Prettier with them, and this is the first time I've seen this bug because Zed does format-on-save out of the box, and also has built-in support for YAML. It took an hour to figure out, and seems like it will affect any user-developer using YAML with long lines.

Would you be open to considering this workaround?

1. Edit https://zed.dev/docs/languages/yaml

2. Add a note about this gotcha, to make it easier for zed users to find in the docs and to solve.

3. Add info about how to do workarounds:

   - Option 1: wrap the value in single quotation marks (IMHO this is what most people will want)
   - Option 2: configure zed yaml to omit Prettier (and show the actual config, which is mentioned in the previous bug report but not shown)
   - Option 3: configure the longer word wrap (which is mentioned in the previous bug report, and shown, but still puts the user at risk if lines get longer.

Thanks for your consideration and a great editor! <3

Current: My YAML file gets broken by Zed & Prettier.

Expected: My YAML file continues to work as usual.

Zed: v0.212.6 (Zed) 
OS: macOS 26.1
Memory: 128 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

afonsoferreiraverissimo:
YAML files that contain templating syntax with double curly braces (for example Ansible-style expressions) could be rewritten into invalid output when saved after edits. This was especially visible on long lines, where formatting expanded the expression into broken multi-line braces.

The fix changes YAML to use the language server formatter by default instead of relying on Prettier defaults for this language. This preserves templated expressions and prevents Zed from emitting invalid YAML when users edit and save these files.

Tests were added to confirm that YAML uses language server formatting by default and that users can still configure Prettier formatting in settings.

The fix gives the user the option to revert to the previous Prettier formatting by modifying the settings file.

Self-Review Checklist:

Closes #46632

Release Notes:

- Fixed YAML formatting breaking templated expressions with double curly braces

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -2197,6 +2197,7 @@
       },
     },
     "YAML": {
+      "formatter": "language_server",
       "prettier": {
         "allowed": true,
       },
--- a/crates/editor/src/editor_tests.rs
+++ b/crates/editor/src/editor_tests.rs
@@ -21447,6 +21447,190 @@ fn completion_menu_entries(menu: &CompletionsMenu) -> Vec<String> {
 }
 
 #[gpui::test]
+async fn test_yaml_document_format_defaults_to_language_server(cx: &mut TestAppContext) {
+    init_test(cx, |settings| {
+        settings.defaults.formatter = Some(FormatterList::default())
+    });
+
+    let fs = FakeFs::new(cx.executor());
+    fs.insert_file(path!("/file.yaml"), Default::default()).await;
+
+    let project = Project::test(fs, [path!("/file.yaml").as_ref()], cx).await;
+    let language_registry = project.read_with(cx, |project, _| project.languages().clone());
+    language_registry.add(Arc::new(Language::new(
+        LanguageConfig {
+            name: "YAML".into(),
+            matcher: LanguageMatcher {
+                path_suffixes: vec!["yaml".to_string(), "yml".to_string()],
+                ..Default::default()
+            },
+            ..LanguageConfig::default()
+        },
+        Some(tree_sitter_yaml::LANGUAGE.into()),
+    )));
+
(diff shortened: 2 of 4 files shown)


afonsoferreiraverissimo:
Hello! Just wanted to gently bump this PR whenever someone has time to take a look. I'd be happy to make any changes if needed, thanks!

PR #55051, opened 2026-04-28
git_ui: Add setting to automatically expand diffs on file open

FloppyDisco:
Self-Review Checklist:

from disscussion #53717
this PR adds a setting for the user to be able to control whether or not the diff hunks are automatically expanded when opening a file from the GitPanel.
the name of the setting is `git_panel.open_file_with_diff` and the default setting is `true` which preserves the current behavior.

just to be upfront, claude knocked this one out for me as i have not done any work with the settings system before. it's a pretty simple change and seems to be written the same as the other settings. 

i have successfully built and ran this change and it seems to be functioning as intended.

Screenshot 2026-04-27 at 22 10 48

Release Notes:

- Added a setting for opening files with diffs automatically expanded

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -987,6 +987,11 @@
     //
     // Default: 72
     "commit_title_max_length": 72,
+    // Whether opening a file from the git panel should automatically expand
+    // the diff hunks.
+    //
+    // Default: true
+    "open_file_with_diff": true,
   },
   "message_editor": {
     // Whether to automatically replace emoji shortcodes with emoji characters.
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -988,10 +988,10 @@
     // Default: 72
     "commit_title_max_length": 72,
     // Whether opening a file from the git panel should automatically expand
-    // the diff hunks.
+    // all diff hunks and navigate to the first change.
     //
     // Default: true
-    "open_file_with_diff": true,
+    "expand_diff_hunks_on_open": true,
   },
   "message_editor": {
     // Whether to automatically replace emoji shortcodes with emoji characters.
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -1338,6 +1338,7 @@ impl GitPanel {
                 })
                 .ok()?;
 
+            let open_file_with_diff = GitPanelSettings::get_global(cx).open_file_with_diff;
             let workspace = self.workspace.clone();
             cx.spawn_in(window, async move |_, mut cx| {
                 let item = open_task
@@ -1351,22 +1352,24 @@ impl GitPanel {
                         diff_task.await;
                     }
 
-                    cx.update(|window, cx| {
-                        active_editor.update(cx, |editor, cx| {
-                            editor.expand_all_diff_hunks(&ExpandAllDiffHunks, window, cx);
+                    if open_file_with_diff {
+                        cx.update(|window, cx| {
+                            active_editor.update(cx, |editor, cx| {
+                                editor.expand_all_diff_hunks(&ExpandAllDiffHunks, window, cx);
 
-                            let snapshot = editor.snapshot(window, cx);
-                            editor.go_to_hunk_before_or_after_position(
-                                &snapshot,
-                                language::Point::new(0, 0),
-                                Direction::Next,
-                                true,
-                                window,
-                                cx,
-                            );
+                                let snapshot = editor.snapshot(window, cx);
+                                editor.go_to_hunk_before_or_after_position(
+                                    &snapshot,
+                                    language::Point::new(0, 0),
+                                    Direction::Next,
+                                    true,
+                                    window,
+                                    cx,
+                                );
+                            })
                         })
-                    })
-                    .log_err();
+                        .log_err();
+                    }
                 }
 
                 anyhow::Ok(())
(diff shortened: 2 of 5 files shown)


NeelChotai:
Thank you for the contribution, but we want to move in a slightly different direction with this. I don't think this should be scoped to its own setting, and it feels like the toolbar is the more natural place to surface the "Collapse/Expand All" functionality.

In the meantime, we have the `editor: expand all diff hunks` from the command palette that can be used for this purpose.

auduongtuan:
https://github.com/zed-industries/zed/discussions/53717 `editor: expand all diff hunks` seems not working for my use case 😭 i have to do it every time I open a file to hide the diff?

FloppyDisco:
@NeelChotai 

Hey just want to be sure,
The functionality is already in place that automatically expands the diff hunks. This PR would allow users to NOT have this happen automatically if they don't want it. The linked discussion is to a group of users that would prefer this behavior doesn't happen. How would you solve this with the toolbar? I don't understand.

NeelChotai:
> @NeelChotai
> 
> Hey just want to be sure, The functionality is already in place that automatically expands the diff hunks. This PR would allow users to NOT have this happen automatically if they don't want it. The linked discussion is to a group of users that would prefer this behavior doesn't happen. How would you solve this with the toolbar? I don't understand.

Ah, you want the opposite command: `editor: collapse all diff hunks`.

My point was that this shouldn't be a setting at all. A sticky toolbar button (like the unified/split diff buttons in git view) is the more natural place to surface it. It covers both directions, the state persists so you're not re-clicking on every file open, and it keeps this as a UI toggle rather than a config option.

FloppyDisco:
@NeelChotai 
ok. so if I am understanding your suggestion you are proposing that there be a button here on the toolbar that would toggle the diff hunks being shown. 
Screenshot 2026-04-30 at 09 22 41

I don't think this solves the same issue here.

I like the idea of having the ui button, but that button's state would need to be tied to the state of the individual editor that is visible, otherwise if you clicked the button in one editor it would expand the diff hunks in all editor tabs? that doesn't make any sense. 

adding the toolbar button also would not solve the issue presented here, because the issue here is not about the user not understanding how to expand or collapse the diff hunks, the issue is the preference to not have the diff hunks automatically expanded when a file is opened from the GitPanel. 

the current behavior is if you open a file from the ProjectPanel the file is opened without changes, and if you open a file from the GitPanel the file is opened with the changes already visible. both of these ux paths feel good because of the relationship between the GitPanel and changes, and this also helps surface the single file diff view to users, as being able to expand the changes in a file was implemented for several months before anyone even figured out you could do it. lol. 

I like the idea of a ui button that is tied to the state of the editor and toggles whether or not the diffs are expanded, but it don't see how you could tie "whether or not to open the file with the diffs expanded from the GitPanel" to this UI button. 
if you changed the `open file` command for the GitPanel to check the state of this button when opening a file (something like "are the diffs expanded in the current editor? then open the file with the diffs expanded) then that would fundamentally change the UX for all users, because most people aren't going to have the diffs expanded in the current editor when they open a new file from the GitPanel.

These are solving fundamentally different questions:
- the UI button shows the current state of a visible element of the ui (the editor)
- this setting would allow users to change the default way the application *behaves* while preserving the current behavior for most users

I think buttons should be for UI state, settings should alter behavior

auduongtuan:
Any update?

auduongtuan:
@danilo-leal hello can you check on this design change 😂

NeelChotai:
@FloppyDisco you're right about the toolbar button, in retrospect that is definitely the wrong surface for git-specific control.

The friction this PR addresses is symptomatic, the underlying issue is that "Open File" from the git panel is currently doing two things at once: opening the file and surfacing its changes (via hunk expansion). A boolean setting that toggles the second half doesn't resolve that compound, it just adds a switch between two halves of an overloaded action.

The cleaner direction is for those concerns to live in separate affordances: "open the file" and "view this file's changes" as distinct things. [Better single-file diff support](https://github.com/zed-industries/zed/discussions/33773) is one route that would naturally let those split apart. Adding a setting now locks user config around the compound shape we'd rather decompose.

In the meantime, `editor: collapse all diff hunks` from the command palette is the workaround for users hitting this regularly.

NeelChotai:
@auduongtuan Please refrain from pinging uninvolved staff members, this isn't how design decisions get revisited. 

Further discussion belongs in https://github.com/zed-industries/zed/discussions/53717 rather than on a closed PR.

FloppyDisco:
> @FloppyDisco you're right about the toolbar button, in retrospect that is definitely the wrong surface for git-specific control.
> 
> 
> 
> The friction this PR addresses is symptomatic, the underlying issue is that "Open File" from the git panel is currently doing two things at once: opening the file and surfacing its changes (via hunk expansion). A boolean setting that toggles the second half doesn't resolve that compound, it just adds a switch between two halves of an overloaded action.
> 
> 
> 
> The cleaner direction is for those concerns to live in separate affordances: "open the file" and "view this file's changes" as distinct things. [Better single-file diff support](https://github.com/zed-industries/zed/discussions/33773) is one route that would naturally let those split apart. Adding a setting now locks user config around the compound shape we'd rather decompose.
> 
> 
> 
> In the meantime, `editor: collapse all diff hunks` from the command palette is the workaround for users hitting this regularly.


Cool, that makes sense. Let me know if I can help out working up any of that!

PR #55078, opened 2026-04-28
language_models: Extract shared model merge helper for LLM providers

hans-d:
Extracts "configured models override discovered/default models by exact name" into a shared `merge_models_by_name` helper, and refactors OpenRouter and Ollama to use it. This is a preparatory step for adding `openai_compatible` autodiscovery with the same override semantics in a follow-up PR (requested functionality). Benefit is that the OpenRouter logic branch has now additional test coverage re the merge.

- Separated the needed refactoring (to keep code DRY) from the actual feature implementation
- Kept the initial refactoring small, although other providers can (should) use it as well. Hence also the generic signature for the helper.
- Kept the current behaviour (and made it more explicit in tests) re difference in sorting behaviour 
- Ollama has an explicit context window override, this change makes that more explicit

Self-Review Checklist:

Release Notes:
- N/A


bennetbo:
Hey, we don't accept contributions for code refactoring unless they provide clear value, and to me it's not clear what this is trying to solve. If you want to contribute some OpenAI compatible feature feel free to include the refactorings there

PR #55083, opened 2026-04-28
git: Stage files automatically when no conflict markers left

elmarco:
Self-Review Checklist:

- [-] Unsafe blocks (if any) have justifying comments

Closes NA

Release Notes:

- Zed will automatically stage files that are in conflicts with no markers left (can be configured via ``git.auto_stage_on_conflict_resolution`` option).


cole-miller:
I don't think we want this feature, sorry--it's too opinionated for Zed to be doing this for you, even as an optional behavior.

PR #55096, opened 2026-04-28
git: Force delete local branches from picker

jakeleventhal:
## Summary

There's an annoying issue where many local branches that I want to delete (but not on the remote) require me to view the error log pop up, then see that I must first delete the branch. This will just opt to delete it.

- Force branch picker deletion of local branches with `git branch -D`, so unmerged local branches can be removed.
- Preserve remote-tracking branch deletion with lowercase `-d` and keep user-facing command text in sync.

## Test Plan

- `cargo fmt --check`
- `git diff --check`
- `cargo test -p git test_delete_local_branch_forces_deletion`
- `./script/clippy -p git -p project -p git_ui`

Release Notes:

- Fixed branch picker deletion for local branches that require force deletion.

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -2043,7 +2043,7 @@ impl GitRepository for RealGitRepository {
         self.executor
             .spawn(async move {
                 git_binary?
-                    .run(&["branch", if is_remote { "-dr" } else { "-d" }, &name])
+                    .run(&["branch", if is_remote { "-dr" } else { "-D" }, &name])
                     .await?;
                 anyhow::Ok(())
             })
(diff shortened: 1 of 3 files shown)


jakeleventhal:
@cla-bot check

marcogroot:
Made another PR similar to this with local testing, and adding as a setting instead of the default. Closed it since this was made earlier and will see feedback for this, but can re-open it if its the preferred method.
https://github.com/zed-industries/zed/pull/55138

smitbarmase:
Thanks, closing this as there is an internal PR we are working on which lets you opt in to force delete rather than changing the default delete behavior.

PR #55127, opened 2026-04-29
agent_ui: Insert dropped file links at the cursor

### LINKED ISSUE #55002: Dragging a file into an agent thread inserts the file link at the end of the composer instead of at the cursor
### Reproduction steps

1. Open an agent thread in Zed using an ACP/SCP-style external agent chat.
2. Type a prompt in the composer and place the text cursor in the middle of the existing text.
3. Drag a file from the Project Panel / file tree into the composer.
4. Drop it where the cursor is expected to receive the file reference.

Current behavior:
When the file is dropped into the composer, Zed appends the inserted file reference to the very end of the message instead of inserting it at the current caret position.

Expected behavior:
The dropped file reference should be inserted exactly at the current cursor position in the composer, so drag-and-drop can be used inline while writing a prompt.

### Why this is a problem

This makes drag-and-drop in the agent composer frustrating for prompts that are already partially written, because the inserted file reference has to be cut and pasted back into the intended location.

### Notes

- The issue is specifically about dragging a file from the file tree / Project Panel into the agent chat composer.
- This was observed in the agent chat interface used with external agents.
- I searched existing issues before filing and did not find a matching report for this exact caret-position behavior.

saberoueslati:
## Context

Dragging a file from the Project Panel into the agent composer was appending the generated file mention to the end of the prompt instead of inserting it at the active caret position. This change routes dragged file insertion through the same caret-based mention insertion path already used for pasted external paths, so dropped file and directory mentions are inserted inline where the user is typing.

Closes #55002

Video of the manual test below :

[Screencast from 2026-04-29 01-13-59.webm](https://github.com/user-attachments/assets/9d2196fb-accd-4f4a-a9c0-0ce45b212121)

## How to Review

- [crates/agent_ui/src/message_editor.rs](/home/saber/coding/zed/crates/agent_ui/src/message_editor.rs) removes the separate end-of-buffer insertion mode from `insert_mention_for_project_path`, updates `insert_dragged_files` to use the shared caret-based insertion flow, and adds regression tests covering single-file and multi-file drag insertion order at the cursor. The existing paste-at-cursor behavior remains covered by the pre-existing test in the same module.

## Self-Review Checklist

Release Notes:

- Fixed dragged file mentions in the agent composer being appended to the end of the prompt instead of inserted at the cursor


Veykril (approved):
Thanks!

PR #55131, opened 2026-04-29
worktrees: Allow opening existing branches in new worktrees

AlvaroParker:
Typing an exact local branch name in the worktree picker should let users open that branch directly in a new linked worktree instead of creating a detached worktree from it. This PR adds a new option to directly checkout an existing branch on a new worktree. 

Video (right with new feature, left is current behavior on main):

https://github.com/user-attachments/assets/0490bc7c-3b72-4c33-9c98-0ba69519cfb2

Now when a user types a branch name that already exists, it can choose to either create a worktree with the existing branch checked out, or just create a worktree with the specified name. 

Self-Review Checklist:

- [x] Tests cover the new/changed behavior-

Closes #ISSUE

Release Notes:

- Added option to open existing local branches in new worktrees from the worktree picker.

### DIFF
--- a/crates/git_ui/src/worktree_picker.rs
+++ b/crates/git_ui/src/worktree_picker.rs
@@ -289,7 +289,7 @@ enum WorktreeEntry {
         from_branch: Option<RemoteBranchName>,
         disabled_reason: Option<String>,
     },
-    OpenBranchInNewWorktree {
+    CheckoutBranchInNewWorktree {
         branch_name: String,
         disabled_reason: Option<String>,
     },
@@ -492,7 +492,7 @@ impl WorktreePickerDelegate {
         has_named_worktree: bool,
     ) -> Option<String> {
         if self.has_multiple_repositories {
-            Some("Cannot create a named worktree in a project with multiple repositories".into())
+            Some("Cannot create a branch worktree in a project with multiple repositories".into())
         } else if branch_checked_out_in_worktree(&self.all_worktrees, branch_name) {
             Some("This branch is already checked out in a worktree".into())
         } else if has_named_worktree {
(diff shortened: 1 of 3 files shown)


danilo-leal:
Heya, thanks for the PR! This feels like a good idea (auto-creating a branch for the newly created worktree), but the UX feels a bit non-intuitive to me. The reason is that you're typing in the _worktree picker_ and then you see a "Open branch..." button that 1) creates a new worktree and then 2) creates a branch with the same name. I think there should be a better way to communicate what it really does through the UI... not 100% sure how yet, but will think about it and come back to this PR.

AlvaroParker:
> auto-creating a branch for the newly created worktree

I think there may be a misunderstanding here: the goal of this PR is not to create a new branch for the new worktree.

The goal is to let the user open an existing local branch in a new linked worktree when that branch already exists but is not currently checked out anywhere.

Currently, if the user types the name of an existing branch in the worktree picker, Zed creates the new worktree in a detached HEAD state at the tip of that branch instead. This PR adds an option to check out the existing branch directly in the new worktree, while still allowing the detached-worktree behavior as a separate choice.

I do agree the current UX does not communicate that clearly enough. One idea would be to add a filter/toggle in the worktree picker, similar to the one in the branch picker (on the image, the button on the bottom left), so branch suggestions can be shown or hidden when the typed worktree name matches an existing branch.

image

The second alternative would be moving this feature to the branch picker, adding a `Checkout in new worktree` action. However, in the current UI this new action would not fit.

AlvaroParker:
I changed the implementation to make it more clear the new possible actions. 

Checking out an existing branch on a new worktree:

image

If the user types the branch name and the branch is already checked out in a different worktree:

image

danilo-leal:
Oh I see, yeah, thanks for the clarification and images/video! I think this is a valid problem to solve, but I'm unsure whether the worktree picker should have that ability. Not sure how to quite articulate why I feel this way, but I think it'd be more natural to support this flow/capability in the _branch_ picker instead... like, in there, you'd type the name of the branch you want to create and then you could toggle a switch that creates a new worktree together with it, where it'll be checked out. Ultimately, I think the action/intent here is to manipulate _a branch_, instead of a worktree, so the domain feels more like it's in the branch space. Consequentially, the thing you added "Already checked out in {worktree}..." (which is great, something awesome to support) would also feel more natural, imo, in the branch picker. Does this resonate?

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been waiting on the answer for over 3 weeks and has conflicts now.

PR #55137, opened 2026-04-29
git_ui: Fix stale information in branch picker

### LINKED ISSUE #55133: Remote SSH project: branch indicator and branch picker stay stale after branch switch/create until restart
### Summary

On a remote SSH project, Zed's Git UI can keep stale branch state after branch changes.

The remote Git state is correct:

- `git branch --show-current` shows the new branch
- `cat .git/HEAD` points to the new branch
- `git branch` shows the expected branch list

But Zed's UI remains stale:

- the branch indicator in the top-left/title area still shows the previous branch
- the branch picker still shows stale branch data
- newly created branches may not appear immediately in the branch picker search results

Restarting Zed makes the branch indicator and branch picker pick up the correct branch state.

This happens for me in a remote project on an Ubuntu VM over SSH.

### Reproduction steps

I can reproduce this in two ways.

#### Case 1: switch branches outside Zed

1. Open a Git repository as a remote SSH project in Zed.
2. Confirm the current branch shown in Zed.
3. In a terminal on the remote host, switch branches outside Zed:

```bash
git switch some-other-branch
```

4. Verify on the remote host:

```bash
git branch --show-current
cat .git/HEAD
git branch
```

5. Return to Zed and inspect:
   - the branch label in the top-left/title area
   - the branch picker

#### Case 2: create a branch using Zed's own branch picker

1. Open the branch picker in a remote SSH project.
2. Create a new branch from the picker.
3. Immediately reopen the branch picker and search for the branch that was just created.

Observed:

- Zed can continue to display the old branch in the UI even though the remote repo has already switched.
- The branch picker can remain stale and fail to show the newly current branch.
- If I create a branch using the branch picker itself, I still may not be able to immediately search for or switch to that new branch there.
- Restarting Zed makes the UI catch up.

Expected:

- After a branch switch or branch creation, Zed should refresh the current branch and branch list immediately in remote SSH projects.
- The top-left branch indicator and branch picker should reflect the actual remote Git state without requiring a restart.

### Environment

Local:

- Zed `0.233.10`
- Commit: `350f33871e84d5ed83a403523e9490b9bf014a94`
- Version: `0.233.10+stable.250.350f33871e84d5ed83a403523e9490b9bf014a94`

Remote:

- Ubuntu 25.10
- `Distributor ID: Ubuntu`
- `Description: Ubuntu 25.10`
- `Release: 25.10`

### Why this seems like a stale remote Git snapshot bug

The remote repo itself is already correct when checked in terminal and by inspecting `.git/HEAD`.

The fact that restarting Zed makes the correct branch appear suggests the on-disk Git state is fine, but Zed's in-memory Git state is stale until startup rebuilds it.

This seems especially likely because the problem affects both:

- out-of-band branch switches done in terminal
- branch creation done through Zed's own branch picker

### Related issues

These seem related or tangential, but I could not tell whether they fully cover this exact bug:

- #13176 `Git information can get out of sync`
- #51398 `Git status synchronization failure in Git Panel, Editor, and Project Panel`
- #54581 `No branches appear in branch picker on remote container` (closed)
- #54530 `Git worktree creator and branch picker don't work on WSL2` (closed)

This issue may be a more specific remote-SSH/branch-state instance of one of the broader Git sync issues above. Happy to close this if maintainers consider it a duplicate, but I wanted to file this because the reproduction here is specific and consistent:

- remote project over SSH
- branch/head correct in terminal and `.git/HEAD`
- stale branch indicator / stale branch picker in Zed
- restart fixes it

thomascolden585-svg:
## Summary

The branch picker could show an outdated current branch or branch list (for example after `HEAD` or refs change on disk, including setups like SSH remotes or other workflows where filesystem events under `.git` were not handled the way we expected).

This change:

- Adds `Repository::refresh_branches_in_snapshot`, which reloads branches asynchronously and updates the repository snapshot, emitting `HeadChanged` / `BranchListChanged` when appropriate.
- Triggers that refresh when the branch picker is opened and when the window becomes active (via the title bar path), so the UI catches up without requiring a full restart.
- Adjusts worktree background scanning so events under `.git` are not skipped solely because the path resolves to the `.git` directory root in the “skipped subdirs” logic (the previous condition could suppress updates we still care about).

Also adds `.tmp-astro/` to `.gitignore` so local scratch clones are not accidentally staged as embedded repositories.

## Test plan

- [ ] Open the branch picker on a repo; confirm listed branches and current branch match `git branch -a` / `git status`.
- [ ] Change `HEAD` outside Zed (e.g. `git checkout other-branch` in a terminal); reopen the branch picker or refocus the window; confirm the picker updates.
- [ ] Sanity-check worktree / large-repo scanning still behaves (no unexpected FS watcher noise in logs).
- [ ] `./script/clippy` (or CI green).

Closes #55133

---

## Release Notes:

- Fixed stale or incorrect branch information in the branch picker after the repository’s `HEAD` or branch list changed on disk.


thomascolden585-svg:
> We require contributors to sign our Contributor License Agreement, and we don't have @thomascolden585-svg on file. You can sign our CLA at https://zed.dev/cla. Once you've signed, post a comment here that says '@cla-bot check'.

@cla-bot check

agu-z:
Thanks for your contribution, but we won't be merging this change.

I don't think polling from the UI is the right approach here. Repository snapshots are kept in sync through dedicated paths (`schedule_scan` on the host, `apply_remote_update` on clients), and overwriting the snapshot out-of-band from the branch picker and title bar fights that design: on an SSH client, the next update from a stale server would just overwrite it back, and everything else that reads the snapshot would stay stale anyway. If the staleness in #55133 is real, the fix belongs wherever the server-side snapshot fails to refresh, not in the UI. The `worktree.rs` change also reverts filtering that's there deliberately (see #54548 and #57173).

Also worth noting: this doesn't compile (`cx.spawn(|mut cx| async move { ... })` is a long-removed GPUI API), so the test plan couldn't have been exercised. Please make sure changes build and are actually tested before opening a PR.

PR #55138, opened 2026-04-29
Add force setting for branch deletion

marcogroot:
Made a discussion here: https://github.com/zed-industries/zed/discussions/54594
but essentially, right now there is no way to force delete branches from the UI. This is the default behavior in other editors like Intellij. Added a setting to configure this.

Chose to add this as a setting instead of a new endpoint to make the blast radius as small as possible an opt in.

Tested locally

https://github.com/user-attachments/assets/7b97b16a-9e37-4099-87f7-d743aeaed380

image

Self-Review Checklist:

Release Notes:
- added an option to override the default branch deletion behavior for the branch picker to use `-D` instead of `-d`

### DIFF
--- a/crates/project/src/project_settings.rs
+++ b/crates/project/src/project_settings.rs
@@ -542,15 +542,21 @@ impl GitSettings {
 pub struct BranchPickerSettings {
     /// Whether to show author name as part of the commit information.
     ///
-    /// Default: false
+    /// Default: true
     #[serde(default)]
     pub show_author_name: bool,
+    #[serde(default)]
+    // whether branch deletion uses the soft delete ('-d') or hard delete ('-D') flag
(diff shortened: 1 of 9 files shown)


crates/project/src/project_settings.rs
```diff
@@ -542,15 +542,21 @@ impl GitSettings {
 pub struct BranchPickerSettings {
     /// Whether to show author name as part of the commit information.
     ///
-    /// Default: false
+    /// Default: true
```
marcogroot, line 545: This default value comment was previously incorrect, driveby fixing it

marcogroot:
Closing this since i see a similar PR here https://github.com/zed-industries/zed/pull/55096

PR #55140, opened 2026-04-29
Add "Add to .git/info/exclude" action in project_panel and git_panel

kayw-geek:
This PR adds an "Add to .git/info/exclude" action to the project panel and git panel context menus, immediately below the existing "Add to .gitignore" entry. `.gitignore` is committed and affects every collaborator, while `.git/info/exclude` is per-clone and local-only — useful for paths a developer wants to ignore without forcing the choice on the team (local debug scripts, personal scratchpads, IDE-specific notes).

Reading support for `.git/info/exclude` already landed in #42082; this completes the round-trip by adding the write affordance.

Notes:

- **Implementation**: `Repository::add_path_to_exclude` is essentially copy-pasted from `add_path_to_gitignore` (added in #47377). The append/dedup/newline-handling logic is line-for-line identical, with three differences: the path is computed via `backend.main_repository_path().join(REPO_EXCLUDE)` so it lands in the common gitdir (`<commondir>/info/exclude`), matching the location git itself uses (per `gitrepository-layout(5)`, `info/` in a per-worktree gitdir is ignored when `$GIT_COMMON_DIR` is set) and the location Zed's worktree scanner reads from (`crates/worktree/src/worktree.rs:5207`); `fs.create_dir(parent)` ensures `.git/info/` exists for fresh repos; the error message references `.git/info/exclude`. I did not refactor the shared logic into a helper to keep the change scoped to one feature.

- **Git panel parity**: The git panel entry uses `action_disabled_when(!is_created, ...)` to mirror the existing `Add to .gitignore` semantics — both ignore mechanisms are no-ops for tracked files (would need `git rm --cached` first), so the constraint is the same.

- **Toast inconsistency carried forward**: As noted in #47377, `add_to_gitignore` shows a `StatusToast` in `project_panel` but uses `show_error_toast` in `git_panel`. I mirrored that asymmetry rather than introduce a new one. Happy to follow up to unify in a separate PR.

- **Tests**: Added five integration tests in `project_panel_tests.rs` covering empty file, dedup, missing trailing newline, existing trailing newline, and directory entries. No git_panel test was added because that crate has no test scaffolding today (the existing `add_to_gitignore` git_panel handler is also untested).

Release Notes:

- Added an "Add to .git/info/exclude" action to the project panel and git panel context menus, allowing a path to be ignored locally without modifying `.gitignore`.

Screenshot：
CleanShot 2026-04-29 at 13 30 03

### DIFF
--- a/crates/git/src/git.rs
+++ b/crates/git/src/git.rs
@@ -108,6 +108,8 @@ actions!(
         ViewCommit,
         /// Adds a file to .gitignore.
         AddToGitignore,
+        /// Adds a file to .git/info/exclude (local-only, not committed).
+        AddToGitExclude,
         /// Copies the current branch name to the clipboard.
         CopyBranchName,
     ]
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -1473,6 +1473,43 @@ impl GitPanel {
         });
     }
 
+    fn add_to_git_exclude(
+        &mut self,
+        _: &git::AddToGitExclude,
+        _window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        maybe!({
+            let list_entry = self.entries.get(self.selected_entry?)?.clone();
+            let entry = list_entry.status_entry()?.to_owned();
+
+            if !entry.status.is_created() {
+                return Some(());
+            }
+
+            let active_repository = self.active_repository.clone()?;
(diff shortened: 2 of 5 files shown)


kayw-geek:
@cla-bot check

cole-miller:
Thanks for opening a PR! I'm going to close this in favor of #57044, but appreciate the contribution.

PR #55143, opened 2026-04-29
agent_ui: Fix minimum width for inputs

cppcoffee:
Release Notes:

- Fixed agent ui minimum width for inputs

The width of the input box exceeds the width of the parent container.

before:

before

after:
after


danilo-leal:
Heya, thanks for the PR! How do I repro this issue, though? I can't see it right off the bat, even with a larger UI font size.

cppcoffee:
> Heya, thanks for the PR! How do I repro this issue, though? I can't see it right off the bat, even with a larger UI font size.

You can repro this issue as follow:

https://github.com/user-attachments/assets/9be4581e-4309-4e0b-b5c5-ad8a0d14570c

benbrandt:
given there's a conflict and Danilo had some questions on the implementation, going to close

PR #55160, opened 2026-04-29
extension: Perform compilation tasks in parallel

MrSubidubi:
This will improve local dev installation times as well as CI times by a lot (I measured it once and for some extensions, it might save us up to 60% of CI time. It will similarly help locally).

I also moved the debug adapter schema validation to the other validations, as it felt it makes more sense to check there as opposed to doing this during the compilation step.

Release Notes:

- Improved dev extension installation times for language extensions with multiple grammars or language servers.


crates/extension/src/extension_builder.rs
```diff
@@ -96,58 +114,117 @@ impl ExtensionBuilder {
...
-                .context("failed to compile Rust extension")?;
-            log::info!("compiled Rust extension {}", extension_dir.display());
-        }
+        let (tx, mut rx) = oneshot::channel();
 
-        for (debug_adapter_name, meta) in &mut extension_manifest.debug_adapters {
-            let debug_adapter_schema_path =
-                extension_dir.join(build_debug_adapter_schema_path(debug_adapter_name, meta)?);
+        let clang_path = extension_manifest.grammars.is_empty().not().then(|| {
+            std::iter::repeat_n(
```
maxdeviant, line 116: It's not clear to me what this code is doing.

It seems like it's constructing a call to `self.install_wasi_sdk_if_needed()` for every single grammar, but it's not clear to me why we don't just have a single call to install the WASI SDK up front?

MrSubidubi: We actually only construct it once, then share that task across grammar compilation tasks, hence the `.shared()`. 

The advantage of this is that it allows us to more easily start the Rust compilation in parallel already - the grammar task might have to wait for the WASI SDK to be downloaded, but the Rust compilation does not need to wait for that to already proceed with the compilation. Using it like this allows us to already start the Rust compilation while downloading the SDK whilst also allowing to start the grammar compilation should the Rust compilation not yet be finished.

osiewicz: I agree that this code looks quite confusing.  I think it'd help if we didn't have to pass in the sdk task to the compilation tasks, but instead used something like https://docs.rs/async-once-cell/latest/async_once_cell/.

crates/extension/src/extension_builder.rs
```diff
@@ -88,58 +106,117 @@ impl ExtensionBuilder {
...
+            .chain(grammar_compilation_tasks)
+            .collect::<Vec<_>>();
+
+        match options.max_concurrency {
+            CompilationConcurrency::Unbounded => {
+                futures::future::try_join_all(tasks).await?;
             }
+            CompilationConcurrency::Bounded(max_concurrency) => {
+                let mut stream =
+                    futures::stream::iter(tasks).buffer_unordered(max_concurrency.get());
```
osiewicz: Shouldn't this be ordered? As in, ideally the compilation should follow whatever the order of the input is?

crates/extension/src/extension_builder.rs
MrSubidubi: I do agree it looks rather confusing, but I am unsure whether I'd want to introduce another dependency to the codebase, given that `.shared()` basically adds this very functionality already. I heard that adding new dependencies for this is frowned upon by some people here.

Perhaps adding a comment that explains why this happens this way would be a good path forward here?

crates/extension/src/extension_builder.rs
MrSubidubi: We do not rely on the order of the input at all, after all, we only want for the compilation to be done ASAP. Since ordering could introduce some starvation if the first task is running the longest, I just went for the unordered buffer here. That should make things only faster if anything.

osiewicz: I'd argue that it'd still be better to order the results here, as otherwise if there are multiple failing extensions builds, the result of compilation task is not going surfaced to the user deterministically.

MrSubidubi: That is a fair point, changed that just now. Should not make too much of a difference in practice anyway, let's see what happens.

osiewicz (approved):

PR #55173, opened 2026-04-29
extensions_ui: Add `RebuildDevExtension` action

tahayvr:
For the extension devs:

Adds a `RebuildDevExtension` action that triggers a rebuild of an installed dev extension. 

- If no dev extensions are installed, shows an error notification.
- If exactly one dev extension is installed, rebuilds it immediately.
- If multiple dev extensions are installed, opens a fuzzy picker to select which one to rebuild.

Makes it easy for devs to trigger a rebuild with a keymap once an action is defined with this PR. 

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Added `RebuildDevExtension` action to rebuild installed dev extensions from the command palette

Screenshot 2026-04-29 at 1 49 54 AM

Screenshot 2026-04-29 at 1 50 16 AM


MrSubidubi (changes requested):
Thanks for this, this is something I wanted to have for a long time and am very happy to see someone pick this up and come up with a reasonable approach.

Have two notes on a high level regardless:
- I think that in the case where people have zero dev extensions installed, we should not show that action in the command pallete and instead hide it. You should be able to do this by looking into the `command_pallette_hooks` and perhaps looking into subscribing to extension install events, although I am not entirely sure the latter is the best way for this. This will also let us get rid of that error message for 0 extensions.
- I think it would similarly be nice if we were to add an optional parameter `extension_id` to that action so that people can add a keybinding to their settings for reinstalling just that very extension. I envision that would be helpful to maintainers that maintain a certain extension but have another one installed ocasionally. By having this, they could always have the proper extension rebuild and we could prompt them if that extension is currently not installed with a toast at the bottom.

What do you think?

tahayvr:
Hey @MrSubidubi ! Appreciate you checking this out and supporting the idea. I thought the suggestions were great so I went ahead and implemented them. 

## What's changed:

- If there are no Dev Extensions installed, the `RebuildDevExtension` action will not show up in the command palette (so no error message for 0 extensions).
- Added `extension_id` as an argument for the action so it can be used for the keymap.

Screenshot 2026-05-07 at 11 06 14 AM

-----

- If the specific extension used in the keymap is not installed anymore, we see an error toast.

Screenshot 2026-05-07 at 11 05 39 AM

crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -45,11 +50,44 @@ actions!(
...
+#[serde(deny_unknown_fields)]
+pub struct RebuildDevExtension {
+    /// The ID of the dev extension to rebuild.
+    ///
+    /// Default: opens a picker if multiple dev extensions are installed.
+    #[serde(default)]
+    pub extension_id: Option<String>,
+}
+
+fn update_rebuild_dev_extension_visibility(cx: &mut App) {
```
MrSubidubi: I think we could have this take the `ExtensionStore` as a parameter instead, as 
- we observe using the `ExtensionStore::global` already and could just uplift that call and then pass it to that method
- we get the store already as a parameter for the invoked observe closure

crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -165,6 +203,48 @@ pub fn init(cx: &mut App) {
...
+            })
+            .register_action(move |workspace, action: &RebuildDevExtension, window, cx| {
+                if let Some(target_id) = action.extension_id.as_deref() {
+                    let found = ExtensionStore::global(cx)
+                        .read(cx)
+                        .dev_extensions()
+                        .any(|m| m.id.as_ref() == target_id);
+
+                    if found {
+                        let extension_id = Arc::from(target_id);
```
MrSubidubi: Minor but we could make this a `find_map` instead and then already clone the exising Arc above while searching. That way, we don't have to create the ID ourselves and could just make this an `if let Some(extension_id)`

crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -165,6 +203,48 @@ pub fn init(cx: &mut App) {
...
+                        ExtensionStore::global(cx).update(cx, |store, cx| {
+                            store.rebuild_dev_extension(extension_id, cx);
+                        });
+                    } else {
+                        workspace.show_error(
+                            &format!("Dev extension '{}' is not installed.", target_id),
```
MrSubidubi:
```suggestion
                            &format!("Dev extension '{target_id}' is not installed."),
```

crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -165,6 +203,48 @@ pub fn init(cx: &mut App) {
...
+                }
+
+                let dev_extensions = ExtensionStore::global(cx)
+                    .read(cx)
+                    .dev_extensions()
+                    .cloned()
+                    .collect::<Vec<_>>();
+
+                match dev_extensions.len() {
+                    1 => {
```
MrSubidubi, line 269: I might have been a bit too over-eager with the latest review, apologies for that - I think we should still note an error if there are no extensions, as in theory users could also bind the action for no specific extension and if they invoke it, we should probably show an error still.

Although, we could in theory also just keep it as a noop as currently. What do you think?

crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -1707,6 +1787,167 @@ impl ExtensionsPage {
...
+        &self,
+        ix: usize,
+        selected: bool,
+        _window: &mut Window,
+        _cx: &mut Context<Picker<Self>>,
+    ) -> Option<Self::ListItem> {
+        let mat = self.matches.get(ix)?;
+        let entry = self.entries.get(mat.candidate_id)?;
+
+        let item = ListItem::new(SharedString::from(format!("dev-extension-{}", entry.id)))
```
MrSubidubi: We should avoid allocating a string here on every render for the sharedstring, as we can create the ID here differently.

I think `("dev-extension-list-item", mat.candidate_id)` should be stable and would be much more efficient.

MrSubidubi (changes requested):
Nice, appreciate the quick follow-up! Left a few comments and think we should probably be good to go then!

crates/extensions_ui/src/extensions_ui.rs
tahayvr: Good catch! I'm a fan of giving user feedback when desired action doesn't happen so I brought it back. Thanks

crates/extensions_ui/src/extensions_ui.rs
```diff
@@ -165,6 +200,55 @@ pub fn init(cx: &mut App) {
...
+                                None
+                            }
+                        });
+                    if let Some(extension_id) = extension_id {
+                        ExtensionStore::global(cx).update(cx, |store, cx| {
+                            store.rebuild_dev_extension(extension_id, cx);
+                        });
+                    } else {
+                        workspace.show_error(
+                            &format!("Dev extension '{target_id}' is not installed."),
```
MrSubidubi: Last minor thing: Can you look into using `ShowNotification` here instead and add a button to the notification which allows the user to install a dev extension? 

I am currently working on adding support for errors on the workspace level, so you could in theory also wait for that change to land, at the same time, we can perhaps get this in quicker by using the notification API here. 

However, I do think we should add that button here no matter what, given that a popup on it's own would be very useless (although right) - the information is relevant, but would be much better with the option to act on it IMO.

MrSubidubi (changes requested):
Nice, thanks! One final thought and I think we should be good!

crates/extensions_ui/src/extensions_ui.rs
tahayvr: Makes sense! I'll give it a go. I assume the other "error: no dev extensions are installed" should also be the same?

MrSubidubi: In theory you could look into that, yes! Not insisting there (since we do hide the action), but since it is in theory invokable using a keyboard shortcut, why not?

MrSubidubi:
Hey, sorry for letting this slip for a bit!

I since added a new API to `Workspace::show_error`, so what we can now do is prompt users to install a dev extension instead of silently failing - would you be interested in looking into this and changing it? All else looks good!

MrSubidubi (approved):
Sorry for the slight delay!

Quickly made that use it myself, given you waited long enough for me here.

Thanks!

PR #55174, opened 2026-04-29
Support BGR subpixel layout

reflectronic:
Release Notes:

- Added text rendering support for BGR subpixel layouts.


SomeoneToIgnore (approved):

drujd:
How exactly do I set the BGR layout? I see that on wayland it utilizes [wl_output::subpixel](https://wayland.app/protocols/wayland#wl_output:enum:subpixel), but how do I even set that for a display on e.g. kwin?

Can't this just be set based on system fontconfig, like most apps work?

PR #55179, opened 2026-04-29
git: Add partial committing

AidanV:
Self-Review Checklist:

Closes #45295

Release Notes:

- Adds action for staging/unstaging selected lines `ToggleStagedSelectedLines`
- Adds keybind to Windows for staging/unstaging hunks `ctrl-alt-y`. I am not sure why this wasn't already set. I do not have a windows machine, so I would appreciate some confirmation on this.
- This PR does not make any UX decisions for how this feature will be used besides providing an action for it. This just provides the infrastructure to be able to make the correct UX decision in a follow up PR.

- **Known limitation**: Identical lines without staging additional surrounding context can cause the line not under the cursor to be staged. This is fixable, but I think that that should be its own PR. 

https://github.com/user-attachments/assets/d4afe032-b207-4522-8488-f197170d83fd

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14146,6 +14146,7 @@ dependencies = [
  "regex",
  "release_channel",
  "remote",
+ "rope",
  "rpc",
  "schemars 1.0.4",
  "semver",
--- a/assets/keymaps/default-windows.json
+++ b/assets/keymaps/default-windows.json
@@ -188,6 +188,7 @@
     "use_key_equivalents": true,
     "bindings": {
       "ctrl-k ctrl-r": "git::Restore",
+      "ctrl-alt-y": "git::ToggleStaged",
       "alt-y": "git::StageAndNext",
       "shift-alt-y": "git::UnstageAndNext",
     },
--- a/crates/agent_ui/src/agent_diff.rs
+++ b/crates/agent_ui/src/agent_diff.rs
@@ -783,7 +783,7 @@ impl DiffHunkDelegate for AgentDiffDelegate {
         &self,
         _stage: bool,
         _hunks: Vec<ResolvedDiffHunks>,
-        _ranges: Vec<Range<editor::Anchor>>,
+        _selections: HashMap<BufferId, Vec<(Range<u32>, bool)>>,
         _editor: &mut Editor,
         _window: &mut Window,
         _cx: &mut Context<Editor>,
--- a/crates/agent_ui/src/agent_diff.rs
+++ b/crates/agent_ui/src/agent_diff.rs
@@ -3,7 +3,7 @@ use acp_thread::{AcpThread, AcpThreadEvent};
 use action_log::{ActionLogTelemetry, LastRejectUndo};
 use agent_settings::AgentSettings;
 use anyhow::Result;
-use buffer_diff::DiffHunkStatus;
+use buffer_diff::{DiffHunkStatus, ResolvedLineSelection};
 use collections::{HashMap, HashSet};
 use editor::{
     DiffHunkDelegate, Direction, Editor, EditorEvent, EditorSettings, MultiBuffer,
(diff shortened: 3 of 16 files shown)


AidanV:
Apologies for the size of this PR. Many of the changes are closely connected. Happy to pair/walk through it.

janlarres:
Thanks for working on this! I just have one comment:

Instead of having to memorize yet another keybinding depending on context I think it would make sense to base the behaviour on whether there is a selection: if there is one then the existing keybinding only stages the selection, and if there isn't then it stages the current hunk just as before. This should also obviate the need for a drop-down menu on the Stage button. That would feel a lot more natural to me.

paolobenve:
why isn't this PR merged?

re1sub:
Is this feature abandoned?

paolobenve:
Please implement it!


 Paolo Benvenuto

Il gio 2 lug 2026, 15:23 reisub ***@***.***> ha scritto:

AidanV:
I am close to finishing this, just busy.

paolobenve:
Thank you


don Paolo Benvenuto

Il gio 2 lug 2026, 17:32 AidanV ***@***.***> ha scritto:

AidanV:
@janlarres I took your advice and got rid of the other keybinding by doing partial staging if there is a selection. However, I think getting this behavior to be always intuitive is a little tricky, and I am curious to hear what you and others think. The main point of contention is figuring out when deletions that are collapsed should be staged. The behavior that I went with is that if the selection goes across the top of the hunk, then the deletions should be staged. I think this behavior is okay but not perfect, as it is not super obvious to the user that they did not stage the whole hunk if they select all of the lines (see video below). 

I think this could be resolved with a visual change to the gutter so it doesn't look fully staged when collapsed (but is that enough? And what should that look like, I included a sketch below). Another, imperfect workaround is staging the deleted lines whenever all visible lines of a hunk are selected (but what if a user stages the first half, realizes they have only staged the first half, and then stages the second half? We end up back where we started: the hunk looks fully staged but isn't. Or what if they want to only stage the added lines, then staging the deleted lines is unexpected behavior). 

I kept the drop-down. I like still being able to stage a specific hunk regardless of the current state of my cursor, and there should be a way for the user to partial stage without using a keybinding. 

Current behavior:

https://github.com/user-attachments/assets/4561f694-8a58-4f62-848e-084fd138f686

Example visual gutter change:
2026-07-05-225024_hyprshot

janlarres:
@AidanV That's an interesting question. Intuitively I would think that selecting all of the lines in a hunk should work the same as staging the hunk as a whole, but you're right that that leads to ambiguity if someone stages a hunk piecemeal. 

I'm currently mostly using LazyGit, which doesn't have this issue because modified lines are never collapsed but shown as individual added and deleted lines. LazyGit also has separate "hunk" and "line" modes instead of allowing selection of lines, and depending on the mode `<space>` stages either the current hunk or the current line. Zed already has a command to expand all hunks, maybe this could be integrated into a similar mode so that you're either in the default "hunk" mode or switch to a "line" mode where all the hunks are expanded, and the "stage" commands apply either to the current selection or to the current line if there is no selection? I'm not sure how feasible that is but it seems to me that it would solve the problems you're describing.

If the above idea is not feasible then a differently-coloured gutter like you described would at least make the ambiguities more obvious. It would even be helpful together with the above idea for when you switch back to "hunk" mode.

AidanV:
#46541 Is quite the merge conflict unfortunately. However, this is a complimentary change and should make this feature better in the long run. I am currently working through the necessary rewrite.

AidanV:
I finished resolving the merge conflict, and I removed the new keybinding and the drop-down UI feature. This means that after this PR lands the only way of interacting with this feature will be through dispatching the `ToggleStagedSelectedLines` action. We can improve the UX in follow-up PRs. The core functionality works well and increasing the scope of this PR to include discovering the correct UX decisions would drastically increase the length of review time.

Additionally, this feature is not yet implemented on the new 'view unstaged/staged' multibuffers. I am calling it out of scope, and it should be an easy follow up PR.

jakeleventhal:
@AidanV you should add this to context menu as well like vs code

right click "stage selected lines" and "unstage selected lines"

PR #55182, opened 2026-04-29
editor: Add path to context

stackmystack:
Hello,

This PR enables keybindings to match against the path of the file opened in the editor.

The main motivation is to use `GIT_EDITOR='zed --wait'` and then when doing interactive rebase we can have a set of custom keybindings for the list of actions: p for pick, r for reword, etc. that match on `path == '.git/rebase-merge/git-rebase-todo'`.

Combined with tasks, I can use [`gitu`](https://github.com/altsem/gitu) or [`lazygit`](https://github.com/jesseduffield/lazygit) from within zed (as demoed in your blog), and truly have all my git workflow inside zed (I know you've been working hard on git integration in zed, but some habits refuse to die, specifically [_magit habit_](https://docs.magit.vc/)).

I looked into your issues and discussions to see if it's something already discussed, but I found nothing. You may have some other concerns or plans about it, and if so, I'd like to help.

I should've asked on these channels before maybe, but TBH this so simple that I am seriously considering maintaining a fork of zed just for that, since AFAICT extensions are not expressive enough to achieve this.

This PR is at a bare minimum (the path is not normalized, no tests, etc.), waiting your feedback. I don't want to commit to something if rejected.

Any guidance would be really appreciated.

Thanks for the snappiest, most comfortable editor out there!

# Self-Review Checklist:

# Release Notes:

- N/A

### DIFF
--- a/crates/editor/src/editor.rs
+++ b/crates/editor/src/editor.rs
@@ -2931,15 +2931,15 @@ impl Editor {
         }
 
         if let Some(singleton_buffer) = self.buffer.read(cx).as_singleton() {
-            if let Some(extension) = singleton_buffer.read(cx).file().and_then(|file| {
-                Some(
-                    file.full_path(cx)
-                        .extension()?
-                        .to_string_lossy()
-                        .to_lowercase(),
-                )
+            if let Some((extension, path)) = singleton_buffer.read(cx).file().and_then(|file| {
+                let full = file.full_path(cx);
+                Some((
+                    full.extension()?.to_string_lossy().to_lowercase(),
+                    full.to_string_lossy().into_owned(),
+                ))
             }) {
                 key_context.set("extension", extension);
+                key_context.set("path", path);
             }
         } else {
             key_context.add("multibuffer");
--- a/crates/gpui/src/keymap/context.rs
+++ b/crates/gpui/src/keymap/context.rs
@@ -393,6 +393,18 @@ impl KeyBindingContextPredicate {
                 let (predicate, source) = Self::parse_expr(source, PRECEDENCE_NOT)?;
                 Ok((KeyBindingContextPredicate::Not(Box::new(predicate)), source))
             }
+            '\'' => {
+                let s = &source[1..];
+                let end = s
+                    .find('\'')
+                    .context("unterminated string in context predicate")?;
+                let value = &s[..end];
+                source = skip_whitespace(&s[end + 1..]);
+                Ok((
+                    KeyBindingContextPredicate::Identifier(value.to_string().into()),
+                    source,
+                ))
+            }
             _ if is_identifier_char(next) => {
                 let len = source
                     .find(|c: char| !is_identifier_char(c) && !is_vim_operator_char(c))


stackmystack:
@cla-bot check

mikayla-maki:
Can you show an example keybinding that this enables? Like how it would show up in the json file?

stackmystack:
> Can you show an example keybinding that this enables? Like how it would show up in the json file?

ofc:

```json
  {
    "context": "vim_mode == normal && path == '.git/rebase-merge/git-rebase-todo'",
    "bindings": {
      "p": ["workspace::SendKeystrokes", "^ c w pick escape"],
      "r": ["workspace::SendKeystrokes", "^ c w reword escape"],
      "e": ["workspace::SendKeystrokes", "^ c w edit escape"],
      "s": ["workspace::SendKeystrokes", "^ c w squash escape"],
    },
  },
```

And so on.

EDIT: forgot to add the `gitu` task:

```json
  {
    "label": "gitu",
    "command": "gitu",
    "env": {
      "GIT_EDITOR": "zed --wait",
    },
    "hide": "always",
    "reveal_target": "dock",
    "show_summary": false,
    "show_command": false,
    "allow_concurrent_runs": false,
    "use_new_terminal": true,
  },
```

stackmystack:
This can go a long way: regex, globbing, etc. Depending on your vision. So I don't mind implementing a solid feature if it's to be accepted.

crates/editor/src/editor.rs
```diff
@@ -2931,15 +2931,15 @@ impl Editor {
...
-                Some(
-                    file.full_path(cx)
-                        .extension()?
-                        .to_string_lossy()
-                        .to_lowercase(),
-                )
+            if let Some((extension, path)) = singleton_buffer.read(cx).file().and_then(|file| {
+                let full = file.full_path(cx);
+                Some((
+                    full.extension()?.to_string_lossy().to_lowercase(),
```
SomeoneToIgnore, line 2937: The `?` on `full.extension()` gates both tuple fields, so `path` is only set when the file has an extension, which won't work for the examples in OP: `Path::new(".git/rebase-merge/git-rebase-todo").extension()` returns `None` (same for `COMMIT_EDITMSG`).

So the rebase-todo binding from the PR description can never activate.
`path` should be set unconditionally, independent of `extension`.

crates/editor/src/editor.rs
```diff
@@ -2931,15 +2931,15 @@ impl Editor {
...
+                    full.to_string_lossy().into_owned(),
+                ))
             }) {
                 key_context.set("extension", extension);
+                key_context.set("path", path);
```
SomeoneToIgnore, line 2942: Even with the extension gating fixed, this value can't equal `.git/rebase-merge/git-rebase-todo`.
`File::full_path` prefixes the worktree root name for visible worktrees (`crates/language/src/buffer.rs`), and for invisible single-file worktrees — which is exactly what `GIT_EDITOR='zed --wait'` produces — it returns an absolute path with `~` substitution (`crates/worktree/src/worktree.rs`, `Worktree::full_path`).

To make this usable I'd suggest a normalized project-relative value, and/or a separate `file_name` key (which alone covers the git use case).
Also worth deciding here: on Windows this value contains `\`, so bindings won't be portable, and `==` is case-sensitive while most filesystems aren't (`extension` sidesteps this by lowercasing).

crates/gpui/src/keymap/context.rs
```diff
@@ -393,6 +393,18 @@ impl KeyBindingContextPredicate {
                 let (predicate, source) = Self::parse_expr(source, PRECEDENCE_NOT)?;
                 Ok((KeyBindingContextPredicate::Not(Box::new(predicate)), source))
             }
+            '\'' => {
```
SomeoneToIgnore, line 396: This breaks round-tripping and, through it, the keymap editor.

`Display` prints `Identifier` bare (`context.rs`, `impl fmt::Display`), so this predicate renders as `path == .git/rebase-merge/git-rebase-todo`, which fails to re-parse (`.` is not an identifier char).
The keymap editor renders contexts via `predicate.to_string()` (`crates/keymap_editor/src/keymap_editor.rs`, ~L861) and validates edits by re-parsing (~L2768), so a binding using a quoted string shows an invalid context in the UI and can't be saved from there.

Parsing works, but `parse(parsed.to_string())` fails.

SomeoneToIgnore, line 396: Two smaller design points on parsing strings into `Identifier`:
Bare `'Editor'` becomes a legal predicate meaning "context contains identifier Editor", and a string on the left of `==` silently turns into a key lookup (`new_eq` accepts any two identifiers).
A distinct `Literal` variant that `new_eq` only accepts on the right side would keep the semantics honest and also keep `is_superset` correct.
There's also no escape support, so a path containing `'` is inexpressible right now — guess, it's fine for now but wanted to raise...

SomeoneToIgnore (changes requested):
Thanks for the PR — the concept is welcome, and there's precedent for it (the `extension` context key).
That said, the current implementation can't express its own motivating use case, and the parser change breaks the keymap editor, so this needs another iteration.

Since you asked for guidance, here's the shape I'd suggest:
Set `path` unconditionally, normalized to a project-relative form, and consider adding `file_name` too — that alone covers the git-rebase case portably.
Give string literals their own `Literal` parser variant and make `Display` quote non-identifier values.
Add tests for the parser (including round-trip!) and for the editor context keys.
Update `docs/src/key-bindings.md` with the new key and the quoted-string syntax.
Change the release note from `N/A` to a `- Added ...` entry, since this is user-facing.

Happy to review the next round.

SomeoneToIgnore:
Will close to keep the PR queue cleaner, as 2 weeks had passed with no activity.

PR #55188, opened 2026-04-29
Fix the thread completion notification not dismissing when focusing the panel

SomeoneToIgnore:
Before:

https://github.com/user-attachments/assets/2ea0f1ed-5189-4b27-8b1e-3f47d7684a28

After:

https://github.com/user-attachments/assets/535d47b6-d6dc-41cd-a14e-d2f5d6b5a3a5

Release Notes:

- Fixed the thread completion notification not dismissing when focusing the panel


benbrandt (approved):
Thanks!

Issue #54633, opened 2026-04-23
sub agents list squashed.


waleed1892:
### Reproduction steps

<img width="1227" height="690" alt="Image" src="https://github.com/user-attachments/assets/92b1d2fb-52df-4f98-8c70-5c5093a00a9e" />

### Current vs. Expected behavior

sub agents should be clearly visible

### Zed version and system specs

Zed: v0.233.7+stable.244.2f844c3cdb1380f992a892f173eb45b6bcaad24c (Zed) 
OS: macOS 26.4.1
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

danilo-leal:
Hey! Do you have any settings that might be relevant here? Without any font size-related configurations, I can't repro this.

HalavicH:
Hi. @waleed1892, I just encountered the same issue. 

<img width="1730" height="622" alt="Image" src="https://github.com/user-attachments/assets/b05c5a8a-fa33-466d-8768-609d12ba8b13" />

Easy repro on Zed Preview / Zed Dev: ask your agent (e.g. Claude) to _"create a plan with 20 tasks"_. The plan view immediately shows the issue.

@danilo-leal, I don't understand why you need font settings to reproduce since it's layout issue.

**Root cause**

The problem is that `v_flex()` is a **flexbox column container**. Flexbox children have `flex-shrink: 1` by default, so when the container hits `max_h`, the flex algorithm compresses all children to fit rather than overflowing and scrolling. So we end up with 20 items crammed into 160px instead of a scrollable list.

Separate the scroll boundary from the flex layout container. A plain `div` (non-flex) as the outer scroll container lets its inner `v_flex` size naturally — content overflows the bounded `div` and scrolling works correctly. 

The same fix applies to the edited files list in the same panel.

Fix in PR 👉  #55189 👈 . Works for my Zed Dev build on MacOS

PR #55189, opened 2026-04-29
agent_ui: Adjust edited files and plan lists to fit content height

HalavicH:
When a plan contains many tasks (or the edited-files list is long), items were being visually compressed into the fixed-height container rather than scrolling.

image_2026-04-28_20-17-43

Asking an agent to "create a plan with 20 tasks" is an easy repro.

**Root cause**

Both lists used a `v_flex()` as their scroll container, and flexbox children have `flex-shrink: 1` by default. When the container hits `max_h`, the flex algorithm compresses all children to fit instead of overflowing into the scroll region — resulting in 20 items crammed into 160px.

**Fix**

Separate the scroll boundary from the flex layout. A plain `div` (non-flex) as the outer scroll container lets its inner `v_flex` size naturally — content then overflows the bounded `div` and scrolling works correctly.

Result:
image_2026-04-29_10-09-24

Closes issue #54633.

> What's missing: vertical scrollbar, as it requires large refactoring due to double mutable borrowing of `cx` (the second introduced by scroll handle), which will slowdown review of this pr

Release Notes:

- Fixed plan and edited-files lists in the agent panel being squashed when they contain many items


Anthony-Eid:
In the future this should probably be a `uniform_list`, but this is a good fix for now! Thank you for doing this

Anthony-Eid (approved):

HalavicH:
> In the future this should probably be a `uniform_list`, but this is a good fix for now! Thank you for doing this

didn't want to introduce more changes than needed.

I can rework it to uniform list with next pr (which adds scrollbar too)

PR #55190, opened 2026-04-29
sidebar: Don't do anything on cmd-click if project is already active

danilo-leal:
Prevent flashing the currently selected thread upon cmd-clicking the active project's header.

Release Notes:

- Fixed a bug where a thread within the currently active project would flash upon cmd-clicking the project header.


bennetbo:
@zed-industries/approved

PR #55194, opened 2026-04-29
agent_ui: Delete to start of soft-wrapped line on cmd-backspace

faradayfury:
## Summary

Make `cmd-backspace` in the agent panel's message editor delete to the start of the current *visual* (soft-wrapped) line instead of the start of the logical line. This matches the standard macOS text-input convention, which is the behavior most users expect from muscle memory in browsers, native text fields, Slack, iMessage, etc.

Before: typing a long prompt that wrapped across three rows and pressing cmd-backspace deleted everything back to the previous newline — wiping out all four rows.

image
image

---

After: cmd-backspace deletes only the contents of the current soft-wrapped row up to the caret.

image
image

## Implementation notes

- New `DeleteToBeginningOfMessageEditorVisualLine` action declared in `agent_ui.rs`.
- `cmd-backspace` is bound to it in the `MessageEditor > Editor` context inside `assets/keymaps/default-macos.json`. macOS-only; no change for other platforms.
- The handler in `message_editor.rs` uses `editor::movement::line_beginning(map, head, /* soft */ true)` to compute the deletion target.
- Edge case: if the caret is at the end of the last soft-wrapped row of a multi-row paragraph, an empty replacement would collapse the row and pull the caret up to the previous row. To avoid that jump, the handler substitutes `"\n"` instead of `""` in that one case so the row stays in place.

## Tests

Two new `#[gpui::test]`s in `crates/agent_ui/src/message_editor.rs`:

- `test_delete_to_beginning_of_message_editor_soft_wrapped_line` — caret in the middle of a soft-wrapped row; expects only that row's preceding content to be removed.
- `test_delete_to_beginning_of_message_editor_last_soft_wrapped_line_preserves_row` — caret at end of the last soft-wrapped row; expects the row to be cleared but preserved (caret stays on the empty row).

## Test plan

- [x] `cargo test -p agent_ui delete_to_beginning_of_message_editor`
- [x] Manual: open the agent panel, paste a long prompt that soft-wraps across multiple rows, place the caret partway through the second row, press cmd-backspace — only that row is cleared up to the caret; the rows above are untouched.
- [x] Manual: place the caret at the end of the last soft-wrapped row, press cmd-backspace — the row is cleared but the caret stays on the now-empty row instead of jumping up.
- [x] Manual: caret at column 0 — no-op (existing behavior preserved).

Release Notes:

- Improved cmd-backspace in the agent panel's message editor on macOS to delete to the start of the current soft-wrapped line, matching native text-input behavior.


faradayfury:
@cla-bot check now

faradayfury:
@cla-bot check

benbrandt:
This already works for me. Going to close

faradayfury:
> This already works for me. Going to close

😅 is there a setting that i need to toggle? because it still deletes the whole thing for me

faradayfury:
@benbrandt i still face this issue :(

PR #55198, opened 2026-04-29
ci: Ensure generated workflows do not become stale

MrSubidubi:
This ensures that upon workflow removal in Rust code, the corresponding workflow file will not accidentally stay in the repository.

Release Notes:

- N/A


ConradIrwin (approved):
Thanks! But lets make --cleanup the default

PR #55199, opened 2026-04-29
ci: Remove release retagging support

MrSubidubi:
From now on, we will instead just make another patch version bump. That has the advantage that tags will actually always only be applied to the version bump and not some random change as well as us being able to prohibit updates of tag refs for the Zed Zippy identity as well.

Release Notes:

- N/A


ConradIrwin:
@zed-zippy approved

PR #55203, opened 2026-04-29
agent_ui: Preserve selection mentions when starting a new thread

NeelChotai:
When the "+" button created a fresh draft, `active_initial_content`
fell back to the raw editor text when the async `draft_prompt`
observer had not yet resolved. That raw text contains fold placeholder
strings (e.g. "selection") rather than the mention links, so creases
and their registered URIs were dropped from the carried-over draft.

Related to https://github.com/zed-industries/zed/issues/53981.

Release Notes:

- Fixed a bug where selection mentions would resolve to the literal `selection` rather than the URI in draft threads.


gaauwe:
@zed-zippy approved

Issue #43989, opened 2025-12-02
Right-click context menu triggers format on save with Auto Save on Focus Change, causing cursor jitter


flappyBug:
### Reproduction steps

1. Start from a clean Zed installation.
2. Enable "format on save" and set "Auto Save" to "on focus change" in settings.
3. Open a code file that supports formatting.
4. Edit some content but do not save.
5. Open the right-click context menu in the editor (such as quick actions or code context menu).

### Current vs. Expected behavior

**Current behavior:**
- Triggering the right-click context menu causes Zed to detect a focus change, which activates Auto Save and thereby triggers format on save.
- Formatting can result in the cursor position jumping (cursor jitter), impacting user experience.
- Editor treats the appearance of the context menu as a loss of focus, which initiates saving and formatting.

**Expected behavior:**
- Right-click context menu should act as an internal editor action, NOT resulting in a focus loss, and should NOT trigger save or format actions.
- When invoking the context menu, the current editing state and cursor position should be maintained.

### Zed version and system specs

Zed: v0.214.7 (Zed) 
OS: macOS 15.7.1
Memory: 16 GiB
Architecture: aarch64

### Relevant Zed settings

<details><summary>settings.json</summary>

```json
{
  "format_on_save": "on",
  "autosave": "on_focus_change"
}
```

</details>

PR #55216, opened 2026-04-29
Add user-configurable mouse and scroll wheel keybindings

### LINKED ISSUE #10647: User configurable mouse bindings (like keymap for key+mouse)

OmChillure:
Closes #10647

Replaces #53889 with the approach Conrad and I aligned on: 
unify mouse/scroll into `Keystroke` rather than introducing parallel `MouseStroke`/`ScrollStroke` types, and resolve bindings against the dispatch path under the cursor rather than the keyboard-focus path.

### Design choices worth flagging

- **Probe runs before the mouse listener chain.** A bound mouse action pre-empts the editor's existing modifier-click handlers; if a binding fires, the normal `dispatch_mouse_event` is skipped.
- **MouseUp suppression.** When a `MouseDown` consumed a binding, the matching `MouseUp` (same button + modifiers) is swallowed so click handlers downstream don't double-fire.
- **Scroll bindings only fire with a modifier held.** Unmodified scrolling continues through the normal listener chain, so we don't walk the keymap on every scroll tick.

Are these fine? or do they requries some changes? 

### Other changes

- Mouse buttons and scroll events become `Keystroke` values with special key names: `leftclick`/`rightclick`/`middleclick`/`backclick`/ `forwardclick`, `mousescrollup`/`mousescrolldown`. 
- Click count is encoded in the key (`doubleleftclick`, `tripleleftclick`) so no field  added to `Keystroke`. Naming follows
Conrad's suggestion.
- New `HitboxId → DispatchNodeId` map on `DispatchTree`, populated by `Window::insert_hitbox`, remapped through `reuse_subtree`.
- Five editor actions: `AddCursorAtMouse`, `ExtendSelectionToMouse`, `GoToDefinitionAtMouse`, `GoToTypeDefinitionAtMouse`, `StartColumnarSelectionAtMouse`.

### Example keymap

```json
[
  {
    "context": "Editor",
    "bindings": {
      "alt-leftclick":       "editor::GoToDefinitionAtMouse",
      "shift-leftclick":     "editor::ExtendSelectionToMouse",
      "ctrl-leftclick":      "editor::AddCursorAtMouse",
      "alt-shift-leftclick": "editor::StartColumnarSelectionAtMouse",
      "ctrl-mousescrollup":   "zed::IncreaseBufferFontSize",
      "ctrl-mousescrolldown": "zed::DecreaseBufferFontSize"
    }
  },
]
```

Video 

[Screencast from 2026-04-29 19-48-18.webm](https://github.com/user-attachments/assets/dc835d0f-ff6d-4970-9f8f-1e9c6106254c)

Release Notes:

- Added user-configurable mouse button and scroll wheel keybindings in keymap.json

### DIFF
--- a/crates/editor/src/actions.rs
+++ b/crates/editor/src/actions.rs
@@ -993,3 +993,28 @@ pub struct ConfirmEditReviewComment {
 pub struct CancelEditReviewComment {
     pub id: usize,
 }
+
+/// Adds a cursor at the current mouse position. Intended for keymap-bound mouse buttons.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct AddCursorAtMouse;
+
+/// Extends the current selection to the mouse position. Intended for keymap-bound mouse buttons.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct ExtendSelectionToMouse;
+
+/// Goes to the definition of the symbol at the current mouse position.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct GoToDefinitionAtMouse;
+
+/// Goes to the type definition of the symbol at the current mouse position.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct GoToTypeDefinitionAtMouse;
+
+/// Starts a columnar (block) selection at the current mouse position.
+#[derive(PartialEq, Clone, Deserialize, Default, JsonSchema, Action)]
+#[action(namespace = editor)]
+pub struct StartColumnarSelectionAtMouse;
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -391,6 +391,11 @@ impl EditorElement {
                 .go_to_type_definition_split(action, window, cx)
                 .detach_and_log_err(cx);
         });
+        register_action(editor, window, Editor::add_cursor_at_mouse);
+        register_action(editor, window, Editor::extend_selection_to_mouse);
+        register_action(editor, window, Editor::go_to_definition_at_mouse);
+        register_action(editor, window, Editor::go_to_type_definition_at_mouse);
+        register_action(editor, window, Editor::start_columnar_selection_at_mouse);
         register_action(editor, window, Editor::open_url);
         register_action(editor, window, Editor::open_selected_filename);
         register_action(editor, window, Editor::fold);
--- a/crates/editor/src/navigation.rs
+++ b/crates/editor/src/navigation.rs
@@ -1052,6 +1052,122 @@ impl Editor {
         self.go_to_definition_of_kind(GotoDefinitionKind::Type, true, window, cx)
     }
 
+    fn point_for_mouse(&self, window: &Window) -> Option<(Rc<PositionMap>, PointForPosition)> {
+        let position_map = self.last_position_map.clone()?;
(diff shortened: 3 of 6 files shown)


herohiralal:
does anyone know what's keeping this from getting merged?
i'd really love to be able to use middle-mouse button for "go-to-reference" (like intellij ide's, or like how vscode added it with this: https://github.com/microsoft/vscode/pull/266044 )

solanab:
Thanks for working on this! The `Keystroke`-unified approach is clean.

I have a use case that this infrastructure supports but the current defaults don't yet cover:

**Right-click → both show context menu AND trigger autosave on focus change.**

Currently `rightclick` has no action representation — the editor's `mouse_right_down` calls `deploy_context_menu` directly via the old mouse listener chain. If `rightclick` were exposed as a default binding, users could:

- Keep the default: `rightclick: "editor::ShowContextMenu"`
- Remap: `rightclick: "workspace::Save"`
- Or layer both (if the parser supports duplicate-keys or array values)

### What would make this work

1. **Allow bare `rightclick`** — it appears the code already doesn't enforce a modifier requirement for mouse buttons (unlike scroll). The PR description says modifierless inputs are "intentionally rejected" but the actual `try_dispatch_mouse_button_binding` has no guard. Could this be clarified? It would be great to explicitly support bare `rightclick` so the default context menu can be a binding rather than remaining hardcoded.

2. **Define `ShowContextMenu` as an action** — this would be a small follow-up PR (I'm happy to open one). The editor's `mouse_right_down` handler moves behind an action, and a default keymap entry is added. Then the `rightclick` event either dispatches the user's binding(s) or falls through to the default.

3. **Duplicate key bindings** — does the current keymap parser allow the same key to map to multiple actions (e.g. `rightclick` → both `ShowContextMenu` and `Save`)? If not, array-value syntax like `"rightclick": ["editor::ShowContextMenu", "workspace::Save"]` would be a helpful addition.

### Related

- #43989 — the original fix that stopped context menu from triggering autosave (the motivation behind wanting to bind both actions)
- see also VS Code's `editor.action.showContextMenu` as prior art

Happy to help with the `ShowContextMenu` action-ification PR once this one lands.

solanab:
The branch currently conflicts with `main`. Happy to rebase and push to this PR if that's useful.

My use case is binding right-click to save, e.g. `"rightclick": "workspace::Save"`. Once this lands I can follow up with a small PR that turns the hardcoded editor right-click handler into `editor::ShowContextMenuAtMouse` plus a default `"rightclick"` binding (that's #60849, parked until this is on `main`) so the context menu stays available as an action instead of only existing on the listener chain.

PR #55235, opened 2026-04-29
git_panel: Fix empty state label not align to center

dongdong867:
When the git panel is too narrow for "No changes to commit" or "No Git Repositories" to fit on a single line, the wrapped lines were start-aligned while the button below remained centered, breaking the visual stack.

This wraps each label in a stretched, text-centered `div`, matching the pattern already used by `render_unsafe_repo_ui` in the same file. The bare string is kept (no `Label`) so the parent `v_flex`'s `text_color(Color::Placeholder)` continues to apply unchanged.

### Repository detect

| As-is | To-be |
| --- | --- |
|  |  |

### Repository not detect

| As-is | To-be |
| --- | --- |
|  |  |

Self-Review Checklist:

Closes N/A

Release Notes:

- git_panel: Fixed empty state labels in the git panel becoming left-aligned when wrapped to multiple lines

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -4824,7 +4824,13 @@ impl GitPanel {
     }
 
     fn render_no_changes_ui(&self, cx: &Context<Self>) -> Vec<AnyElement> {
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi (approved):
awesome, thanks!

PR #51288, opened 2026-03-11
editor: Show verbose rust-analyzer diagnostics in hover popover

notJoon:
### Summary 

This is the first of two planned PRs for #21376. the follow-up PR will improve the project diagnostics view to display rendered diagnostics inline.

### Changes

- Store the `rendered` field from rust-analyzer's `diagnostic.data` in the internal `Diagnostic` struct
- Update the diagnostic popover copy button to copy the full verbose compiler output when available
- Add an "Open Full Diagnostic" button that opens the rendered output in a scratch buffer tab

### Result

**Before Fix**
before_fix

**After Fix**
after_fix

after_fix_2

Now we can see the "Open Full Diagnostic" clickable button and rendered compiler error output.

Before you mark this PR as ready for review, make sure that you have:
- [ ] Added a solid test coverage and/or screenshots from doing manual testing
- [ ] Done a self-review taking into account security and performance aspects
- [ ] Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)

Release Notes:

- Added "Open Full Diagnostic" button to the diagnostic popover that opens verbose rust-analyzer compiler output in a scratch buffer
- Improved "Copy Diagnostic" to copy the full rendered compiler output when available

PR #55262, opened 2026-04-30
editor: Surface rust-analyzer's rendered diagnostics

### LINKED ISSUE #21376: Verbose Rust Analyzer Errors
### Check for existing issues

- [X] Completed

### Describe the feature

In VsCode, (I think by default), `rust-analyzer` lsp errors are much more verbose. 
The same thing can be done in Neovim, it should be possible in Zed too because Neovim also renders `rust-analyzer`'s lsp errors as less verbose by default just like zed does:

Like I said Neovim also acts like zed by showing less verbose errors by default. _(1st screenshot)_. 
You actually need to specify to Neovim to show more verbose errors if they exist, you can do something like this to get it to show like the 2nd screenshot by specifying in your global vim config:

```lua
vim.diagnostic.open_float(nil, {
  -- this format function allows neovim to show the full error that
  -- the lsp is providing (right image)
  format = function(diagnostic)
    if diagnostic.source == 'rustc'
      and diagnostic.user_data.lsp.data ~= nil
    then
      return diagnostic.user_data.lsp.data.rendered
    else
      return diagnostic.message
    end
  end,
})
``` 

I think it might be a better idea to give as much lsp info to the user & allow them to turn it off if they wish, because it was hard for me to find out how to do this in neovim as well.

I am on windows right now but I tested it on macOS and it's not any different.

### Environment

Zed: v0.165.0 (Zed Dev 04ff9f060cf9eeb3b848c858f80d6c882bb5cc20)
OS: Windows 10.0.26100
Memory: 31.1 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 4090 || NVIDIA || 561.17

<p float="left">

</p>

tautik:
picks up #51288 (closed for inactivity, thanks @notjoon and @veykril). plumbs rust-analyzer's `diagnostic.data.rendered` through `language_core::diagnostic`, the proto wire format, and `lsp_store`, then surfaces it in the hover popover (copy button + "open full diagnostic" scratch-buffer button, with @veykril's `workspace::for_window` + `color::muted` fixes) and inline as a fenced code block in the project diagnostics view. verified with `cargo check -p language -p project -p editor -p diagnostics` and `cargo test -p diagnostics --lib` (14/14). closes #21376.

Release Notes:

- Surfaced rust-analyzer's full pretty-printed diagnostics in the hover popover (copy + "open full diagnostic" buttons) and inline in the project diagnostics view.


mchisolm0:
Thanks @/tautik for picking this up 😁 Surfacing the verbose diagnostics in the hover popover seems to be working. With them being in the popover, is there still a reason to have the "Open Full Diagnostics" button?

### What I used to test

```rs
fn main() {
    add("a", 1)
}

fn add(a: u32, b: u32) -> u32 {
    a + b
}
```

### What I saw

<details><summary>Main popover</summary>
SCR-20260518-qsqm
</details>

<details><summary>Full diagnostic tab</summary>
SCR-20260518-qgbv
</details>

#### From VS Code for reference

<details><summary>VS Code main popover</summary>SCR-20260518-tbqt</details>

<details><summary>VS Code Full diagnostic tab</summary>
SCR-20260518-tbud
</details>

notJoon:
@mchisolm0 When I initially worked on this, I enabled the "Open Full Diagnostics" button because I thought it would be good to align it with other IDE's functionality (like vscode). I’m leaving this comment in case it’s helpful for the review.

mchisolm0:
Thanks @notJoon 😁 I added the VS Code popover and diagnostic to my first comment. It looks to me like VS Code shows *different* information in the popover than when you open the new tab. That is what I was expecting (and what your original PR did if I remember right).

I am saying the current changes in *this* PR seem to show the same information in the popover and the "Open Full Diagnostic Tab". I hope this clarifies what I was meaning. 😁

crates/project/src/lsp_store.rs
```diff
@@ -13947,6 +13949,18 @@ impl LspStore {
     }
 }
 
+// Pulls the verbose pre-rendered diagnostic text out of a language server's
+// `diagnostic.data` payload. rust-analyzer puts the full pretty-printed compiler
+// output (with ascii-art spans and trait-bound details) under the `rendered`
+// key; other servers may follow the same convention. Returning `None` is the
+// no-op path - the regular `message` is still shown.
+fn extract_rendered(data: Option<&serde_json::Value>) -> Option<String> {
```
SomeoneToIgnore, line 13957: `data` is a very language-specific field which we cannot really have in the generic lsp_store.rs

Let's move it to `rust.rs`, we need something similar to https://github.com/zed-industries/zed/blob/1817b43c80707a3e49c3946a72bd39ee3134c654/crates/languages/src/rust.rs#L319 , near the rest of the rust-analyzer-specific code.

Also we'd better pass the entire diagnostics object by reference in that new method, to be more type-safe that the correct data is passed.

crates/proto/proto/buffer.proto
```diff
@@ -294,6 +294,7 @@ message Diagnostic {
   optional string data = 12;
   optional string code_description = 13;
   optional string markdown = 14;
+  optional string rendered = 18;
```
SomeoneToIgnore, line 297: Don't we pass `data` just above?
Diagnostics can be abundant and we better be reasonable when passing them over the wire, without bloating the data excessively.

crates/diagnostics/src/diagnostic_renderer.rs
```diff
@@ -56,7 +56,12 @@ impl DiagnosticRenderer {
                     initial_range: primary.range.clone(),
                     severity: primary.diagnostic.severity,
                     diagnostics_editor: diagnostics_editor.clone(),
-                    copy_message: primary.diagnostic.message.clone().into(),
+                    copy_message: primary
+                        .diagnostic
+                        .rendered
+                        .clone()
```
SomeoneToIgnore, line 62: Given that we `.clone()` that everywhere, let's use `SharedString` instead of `String` for the `rendered` message.

crates/diagnostics/src/diagnostic_renderer.rs
```diff
@@ -92,6 +102,20 @@ impl DiagnosticRenderer {
...
+        // When the language server provides a verbose pre-rendered diagnostic
+        // (e.g. rust-analyzer's `rendered` field with full ascii-art spans and
+        // trait-bound details), append it inline as a fenced code block so users
+        // see the full error in the project diagnostics view without having to
+        // open a separate buffer.
+        if let Some(rendered) = diagnostic.rendered.as_ref().filter(|s| !s.is_empty()) {
+            markdown.push_str("\n\n```\n");
+            markdown.push_str(rendered);
+            if !rendered.ends_with('\n') {
+                markdown.push('\n');
```
SomeoneToIgnore, line 115: What if it ends with `\r\n`, do we need to cover such case or not?

SomeoneToIgnore (changes requested):
Thank you, as this is a UI change, please add a demo video of the new change.

crates/editor/src/hover_popover.rs
```diff
@@ -1249,10 +1249,78 @@ impl DiagnosticPopover {
...
+                                        .on_click(move |_, window, cx| {
+                                            let rendered = rendered.clone();
+                                            let Some(workspace) =
+                                                Workspace::for_window(window, cx)
+                                            else {
+                                                return;
+                                            };
+                                            workspace.update(cx, |workspace, cx| {
+                                                let project = workspace.project().clone();
+                                                if !project.read(cx).is_local() {
```
SomeoneToIgnore, line 1295: We have to show the diagnostics in remote ssh projects and collab projects, those have all LSP data for that.

crates/diagnostics/src/diagnostic_renderer.rs
```diff
@@ -92,6 +102,20 @@ impl DiagnosticRenderer {
...
             markdown.push_str(&Markdown::escape(&diagnostic.message));
         };
+
+        // When the language server provides a verbose pre-rendered diagnostic
+        // (e.g. rust-analyzer's `rendered` field with full ascii-art spans and
+        // trait-bound details), append it inline as a fenced code block so users
+        // see the full error in the project diagnostics view without having to
+        // open a separate buffer.
+        if let Some(rendered) = diagnostic.rendered.as_ref().filter(|s| !s.is_empty()) {
+            markdown.push_str("\n\n```\n");
```
SomeoneToIgnore, line 112: Is it even correct, I wonder?

First of all, seems quire redundant to carry `rendered` as a new field just to here, as we have `diagnostic.data` here too, from where we can extract the data any time needed?

Then, it's a rendered data already, we do not need to double-render it by putting into another ``` block?

SomeoneToIgnore (changes requested):

SomeoneToIgnore:
To keep the PR queue cleaner, will close it for now, as it's been a bit over 3 weeks without actions, conflicts start to creep in and CI is red still — I think if we fix all the comments, that's quite a great feature to have, unless https://github.com/zed-industries/zed/pull/61030 fixes it?

PR #55263, opened 2026-04-30
Suggest language extensions for untitled buffers

tavaresgmg:
Fixes #53527.

## Summary

- Suggest `untitled.<extension>` when saving an untitled editor buffer with a selected non-Plain Text language.
- Preserve the existing title-based suggestion for existing files, Plain Text buffers, and buffers without a language extension.
- Add a regression test for an untitled Rust buffer suggesting `untitled.rs`.

## Testing

- `mise exec rust@1.95.0 -- cargo fmt --check -p editor`
- `mise exec rust@1.95.0 -- cargo test -p editor test_suggested_filename_uses_language_extension_for_untitled_buffer --lib`

## Suggested .rules additions

None.

Release Notes:

- Fixed Save As suggestions for untitled buffers with a selected language.

### DIFF
--- a/crates/acp_thread/src/diff.rs
+++ b/crates/acp_thread/src/diff.rs
@@ -177,7 +177,7 @@ impl Diff {
         };
         format!(
             "Diff: {}\n```\n{}\n```\n",
-            path.unwrap_or("untitled".into()),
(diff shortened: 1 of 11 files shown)


SomeoneToIgnore (approved):
Thank you.

Issue #55195, opened 2026-04-29
fuzzy matching: double hits are not sorted higher


yara-blue:
### Reproduction steps

Open a code base with paths of the shape `x/x.rs` and `y/x.rs`

Do a fuzzy search for `x` in the file picker.
Note that `x/x.rs` is not above `y/x.rs`.

Example in the Zed codebase:

### Current vs. Expected behavior

<img width="1135" height="951" alt="Image" src="https://github.com/user-attachments/assets/c945467f-e168-414e-98d1-3a99dee46e75" />

Expected: `Cargo.toml zed/crates/zed` at the top

yara-blue:
@feitreim and @11happy you both made fuzzy search soooo much better. Do you have any ideas for a clean fix?

Should we add an additional tie breaker or is there something smart we can do with the search atoms?

feitreim:
One soln I have come up with so far today from a few hours of experimenting is to just scan successful matches for additional occurrences of the word, then give some small reward for multiple occurrences. This does seem to solve this pretty nicely.

Another soln I just came up with, and haven't experimented as much with, is reworking the get_filename_match_bonus function to instead of the current thing it does, (checking the score of each atom against the filename) I think it may work if we just reweight scores to according to some metric related to match density in the tail of the path. The nucleo matcher favors the tail, as you can see in your screenshot the tail-most zed is highlighted in the path. So if the atom appears in the filename the returned indicies represent that, rewarding matches in the tail, for paths specifically seems like intuitive ux to me. further thought: the downside of this approach is that it wouldn't even effect the simple repro case you gave, `x/x.rs` `y/x.rs`, the order wouldn't change. both approaches could be combined.

 
the main constraint, no matter the soln, is that nucleo will only generate a single match char per query char, which makes sense, so no matter what, if you want to do something like match all occurences, you'll have to iterate over the candidate again some way or another. The good news is that doing this only on successfully matched haystacks isn't actually that big of a deal, like a ~~10% hit. That hit was even with adding highlights to matched occurrences, so you could not do that and theres less of a hit.

Ella-horizon:
Hi @yara-blue 
I am so interested in this issue.
I would like to work on your issue.
Thank you!

yara-blue:
@CaptainTimon implementation is now tracked in https://github.com/zed-industries/zed/issues/55872.

[...]

PR #55264, opened 2026-04-30
fuzzy_nucleo: Refactor multi-atom code to use nucleo::Pattern

feitreim:
refactor of the fuzzy_nucleo string and path matching code, instead of handling the multiple atoms ourselves we can just use `nucleo::Pattern` and abstract that all away. this replaces the for loop in the path/string_match_helper functions. all functionality is exactly the same. basically the same / within some tiny margin of the original.

this could enable the use of `nucleo::Pattern::parse` in the future if that was wanted, which allows some extra syntax to activate different matching modes. [more info from deepwiki](https://deepwiki.com/search/how-do-the-different-atom-matc_37e510de-af27-44a1-a52f-3fc367462e6e?mode=fast). I'm pretty sure that enabling that is as simple as switching a `Pattern::new(...)` call with `Pattern::parse(...)`.

Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/crates/fuzzy_nucleo/src/fuzzy_nucleo.rs
+++ b/crates/fuzzy_nucleo/src/fuzzy_nucleo.rs
@@ -79,8 +79,8 @@ impl Query {
             return None;
         }
         let wants_case_penalty = case.is_smart() && query.chars().any(|c| c.is_uppercase());
-        let query_chars = wants_case_penalty
-            .then(|| query.chars().filter(|c| !c.is_whitespace()).collect());
+        let query_chars =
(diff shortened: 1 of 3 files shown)


yara-blue (commented):
First off apologies, took a while to get to this. I'm reviewing it now. One thing I already spotted: we should probably add a test to prevent regressions, could you add one?

yara-blue:
> First off apologies, took a while to get to this. I'm reviewing it now. One thing I already spotted: we should probably add a test to prevent regressions, could you add one?

I thought this addressed #55195 but that is actually not what this does. In that case no test is needed. Merging!

yara-blue:
@zed-zippy approved

PR #55274, opened 2026-04-30
git: Fall back to ls-remote when origin/HEAD symbolic-ref is missing

brunorodmoreira:
## Problem

When a clone's `refs/remotes/origin/HEAD` isn't a symbolic-ref locally — common after older clones or clones made without `-c remote.origin.followRemoteHEAD=always` — `RealGitRepository::default_branch()` falls through to `git config init.defaultBranch`. That config controls *new repo init*, not the clone's actual remote default, so users who have `init.defaultBranch = main` set globally end up with the wrong branch when the actual remote default is something else (e.g. `develop`).

Concretely, in a repo where:

```
$ git symbolic-ref refs/remotes/origin/HEAD
fatal: ref refs/remotes/origin/HEAD is not a symbolic ref
$ git ls-remote --symref origin HEAD | head -1
ref: refs/heads/develop	HEAD
$ git config --get init.defaultBranch
main
```

…"view branch diff" silently diffs against `main` even though the remote default is `develop`. The `git remote set-head origin -a` workaround fixes it locally, but most users never run that command and there's no UX signal that the wrong branch is being picked.

For comparison, VSCode's built-in Git extension (`extensions/git/src/git.ts`) only runs `git symbolic-ref --short refs/remotes/<remote>/HEAD` and throws "No default branch" on failure — no `ls-remote` fallback either. So this also makes Zed strictly better than the closest comparable behavior elsewhere.

## Fix

Insert a new step in the resolution chain: between the existing `refs/remotes/origin/HEAD` symbolic-ref attempt and the `init.defaultBranch` fallback, query the remote with `git ls-remote --symref origin HEAD` and use the symref it returns.

The new step only fires when the local symbolic-refs fail to resolve — the broken-clone case — so healthy clones (where `origin/HEAD` is properly set) hit the existing fast local path and never make a network call. Any failure of the `ls-remote` step (offline, no remote, malformed output) falls through cleanly to the existing `init.defaultBranch` and `master` fallbacks.

The parser is extracted into a small `parse_ls_remote_symref_head` helper for unit testing.

## Test plan

- Added a unit test for `parse_ls_remote_symref_head` covering the typical line, sha-only output, empty input, malformed `refs/heads/` with empty branch, missing trailing `\tHEAD`, and nested branch names like `feature/foo`.
- Added an integration test that sets up a temp "remote" repo with `develop` as its default branch, fetches it into a separate local repo *without* running `git remote set-head` (so `refs/remotes/origin/HEAD` stays unset), and asserts `default_branch()` returns `Some("develop")` for both `include_remote_name=false` and `=true`.
- Manually verified on a real-world repo (`origin/HEAD` unset, remote default `develop`, global `init.defaultBranch = main`): "view branch diff" now opens against `develop` instead of `main` (see screenshots below).

`cargo test -p git --lib` passes (36/36) and `./script/clippy --package git` is clean.

## Screenshots

- Before (origin/HEAD unset, init.defaultBranch=main, remote default=develop):
image

- After
image

Release Notes:

- Fixed "view branch diff" using the wrong base branch when the local `refs/remotes/origin/HEAD` symbolic-ref was missing


brunorodmoreira:
> We require contributors to sign our Contributor License Agreement, and we don't have @brunorodmoreira on file. You can sign our CLA at https://zed.dev/cla. Once you've signed, post a comment here that says '@cla-bot check'.

@cla-bot check

brunorodmoreira:
@cla-bot check

Issue #55128, opened 2026-04-29
Zed hangs when creating a 1 character wide visual block select and moving using either `gj` or `gk` in vim mode.


ZachB-AUT:
### Reproduction steps

1. Install Zed.
2. Enable Vim mode.
3. Create a line long enough to wrap.
4. Start a block select (\<C-v\>).
5. Move either up or down using `gj` or `gk`.

The system then hangs.

This only occurs when the width of the block is 1 character (moving to the side before moving vertically appears to not cause this).

It does not appear to matter if the file is saved or not.

### Zed version and system specs

Zed: v0.233.10+stable.250.350f33871e84d5ed83a403523e9490b9bf014a94 (Zed) 
OS: macOS 15.2
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

<details><summary>Zed.log</summary>

```log
  2983. ['private', 'private', 'char']
  2984. ['private', 'private', 'float']
  2985. ['private', 'private', 'double']
  2986. ['private', 'private', 'boolean']
  2987. ['private', 'private', Identifier]
  2988. ['private', 'private', 'void']
  2989. ['private', 'abstract', '@']
  2990. ['private', 'abstract', 'public']
  2991. ['private', 'abstract', 'protected']
  2992. ['private', 'abstract', 'private']
  2993. ['private', 'abstract', 'abstract']
  2994. ['private', 'abstract', 'static']
  2995. ['private', 'abstract', 'final']
  2996. ['private', 'abstract', 'synchronized']
  2997. ['private', 'abstract', 'native']
  2998. ['private', 'abstract', 'strictfp']
  2999. ['private', 'abstract', '<']
  3000. ['private', 'abstract', 'byte']
  3001. ['private', 'abstract', 'short']
  3002. ['private', 'abstract', 'int']
  3003. ['private', 'abstract', 'long']
  3004. ['private', 'abstract', 'char']
  3005. ['private', 'abstract', 'float']
  3006. ['private', 'abstract', 'double']
  3007. ['private', 'abstract', 'boolean']
  3008. ['private', 'abstract', Identifier]
  3009. ['private', 'abstract', 'void']
  3010. ['private', 'static', '@']
  3011. ['private', 'static', 'public']
  3012. ['private', 'static', 'protected']
  3013. ['private', 'static', 'private']
  3014. ['private', 'static', 'abstract']
  3015. ['private', 'static', 'static']
  3016. ['private', 'static', 'final']
[...]

PR #55285, opened 2026-04-30
vim: Fix visual block display-line movement on wraps

Cloudymap1e:
Summary:

- Fix visual block row traversal so `gj`/`gk` includes soft-wrapped display rows instead of skipping by buffer row.
- Preserve display-row traversal when extending an existing display-line visual block selection horizontally.
- Add a regression for `ctrl-v g j`, follow-up `l`, and `ctrl-v g k` on a single wrapped line.

Tests:

- `cargo test -p vim test_visual_block_display_line_motion_on_wrapped_line -- --nocapture`
- `cargo test -p vim test_visual_block_wrapping_selection -- --nocapture`
- `cargo test -p vim visual_block -- --nocapture`
- `./script/clippy -p vim`

Release Notes:

- Fixed Vim visual block selections hanging when moving by display lines on soft-wrapped lines.

### DIFF
--- a/crates/vim/src/normal/scroll.rs
+++ b/crates/vim/src/normal/scroll.rs
@@ -259,7 +259,7 @@ impl Vim {
             };
 
             if vim.mode == Mode::VisualBlock {
-                vim.visual_block_motion(true, editor, window, cx, &mut move_cursor);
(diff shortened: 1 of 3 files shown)


ChristopherBiscardi:
It looks like this is trying to fix https://github.com/zed-industries/zed/issues/55128

PR #55288, opened 2026-04-30
tools: Make time-zones more case-insensitive

osyvokon:
gpt-5.5 likes to call the `now` tool with upper-case "UTC", leading to this error:

> Failed to receive tool input: tool input was not fully received

Release Notes:

- N/A


benbrandt (approved):
thanks!

PR #55310, opened 2026-04-30
Fix macOS traffic light position when performing a project search

bennetbo:
Regression introduced in #48029, applying the same workaround here that we seem to use for `setDocumentEdited`

Self-Review Checklist:

Release Notes:

- (Preview only) Fixed an issue on macOS where the traffic light position would be wrong when opening the project search


agu-z (approved):
🚥

PR #55328, opened 2026-04-30
gpui: Attribute slow input events to actions

eholk:
When the `input-latency-histogram` feature is enabled, this attaches the action name (if any) to each input event timing the latency tracker records, and buckets per-action counts of slow events at three frame-budget tiers.

## What it does

In `InputLatencyTracker`:

- New `current_action: Option<&'static str>`, set with first-wins semantics matching how `first_input_at` already records only the earliest event in a coalesced frame.
- New `slow_event_counts: FxHashMap<&'static str, SlowEventCounts>` with three tiers:
  - `>= 8ms` — missed a 120fps frame
  - `>= 16ms` — missed a 60fps frame
  - `>= 100ms` — noticeably slow
- Plus `max_latency_ns` per action so the dump can show worst-case.

Only actions that crossed the 8ms tier appear in the map, so it stays bounded to actions that are actually a perf concern.

Events at the `>= 100ms` tier also emit a `log::warn!`, so slow-frame attributions show up in the log file without needing the dump command.

## How attribution stays accurate

- A guard (`inside_input_dispatch`) is set/cleared by `Window::dispatch_event`, so action dispatches from menus, the command palette, or async work don't get attributed to whatever input event happens to fire next.
- Inside `dispatch_action_on_node`, we snapshot the window's update count up front and only call `record_dispatched_action` if state actually changed during the dispatch. This both:
  - Filters out no-op dispatches (action with no listener registered), and
  - Handles the "first action no-ops, second does the work" case correctly — the first call leaves `current_action` unset so the second call gets credit.

## What the dump looks like

`DumpInputLatencyHistogram` gains a "Slow events by action" section sorted by `>=100ms` count desc, then `max_latency_ns`, then `>=16ms` count, capped at 20 rows:

```
Slow events by action:
  action               >=8ms   >=16ms  >=100ms          max
  editor::Backspace        42        8        2     238.4ms
  workspace::Save          12        5        1     150.7ms
  pane::CloseActiveItem     7        3        0      18.2ms
```

## Notes

- All new code is gated behind the existing `input-latency-histogram` cargo feature, so there is zero overhead in default builds.
- Attribution is keyed on `Action::name()`, which is `&'static str`, so per-frame cost is essentially zero (one `Option::get_or_insert` of a string slice).
- Mouse handlers that don't go through the action system (hover, scroll, drag, on_click handlers that do work directly) won't have an attributed action — those slow frames still appear in the latency histogram, just not in the per-action table. This is intentional.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/window.rs
+++ b/crates/gpui/src/window.rs
@@ -1084,7 +1084,36 @@ pub struct InputLatencySnapshot {
     /// Count of input events that arrived mid-draw and were excluded from
     /// latency recording.
     pub mid_draw_events_dropped: u64,
+    /// Per-action counts of slow input events at three frame-budget tiers.
+    /// Only actions that crossed at least the 8ms threshold are present.
+    pub slow_event_counts: FxHashMap<&'static str, SlowEventCounts>,
 }
+
+/// Counts of slow input events for a single action, bucketed by
+/// frame-budget thresholds. An event is counted in every tier whose
+/// threshold its latency meets or exceeds.
+#[cfg(feature = "input-latency-histogram")]
+#[derive(Default, Clone, Debug)]
+pub struct SlowEventCounts {
+    /// Events that missed the 120fps frame budget (>= 8ms).
+    pub slow_8ms: u64,
+    /// Events that missed the 60fps frame budget (>= 16ms).
+    pub slow_16ms: u64,
+    /// Events that crossed the user-noticeable threshold (>= 100ms).
+    pub slow_100ms: u64,
+    /// Highest latency observed for this action across all slow events.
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore:

image

SomeoneToIgnore:
Keeping the PR queue a bit cleaner and will close this one for now, sorry: it's a draft PR with no activity for over a month and merge conflicts.

image

Since it's the 2nd time I do this, maybe I misunderstand the workflow somehow and keeping the PRs is needed in this and other cases?
I hoped that there's always a branch that can be restored when the time comes right hence all this.

PR #55336, opened 2026-04-30
build: Allow pinning of webrtc build artifacts (✈️ edition)

osiewicz:
tl;dr: you can now run `cargo xtask setup_webrtc`, which:
1. Fetches webrtc artifacts into a gitignored directory in Zed repo
2. Adds a [env] section to `~/.cargo/config.toml` on your box which forces
   LK to NOT download webrtc artifacts as a part of its build script

The end result: `cargo clean` is no longer a horrid experience with in
horrid network environments.

Caveats:
1. This does not handle appending to existing cargo config. The setup
   script will fail if there's one in place.
2. You need to redo this thing (fetch env var and whatnot) whenever LK
   version is bumped.
3. This is not mandatory for builds to work. You only really have to do this for your own convenience, but builds will work just fine without it (unless your connection sucks).

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 **/*.proptest-regressions
 **/cargo-target
 **/target
+.webrtc-sys/
 **/venv
 **/.direnv
 *.wasm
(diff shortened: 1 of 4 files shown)


MrSubidubi (commented):
@zed-industries/approved

maxbrunsfeld:
@zed-industries/approved

PR #43703, opened 2025-11-28
Improve support for multiple registrations of  `textDocument/diagnostic`

reflectronic:
Closes https://github.com/zed-industries/zed/issues/41935

The registration ID responsible for generating each diagnostic is now tracked. This allows us to replace only the diagnostics from the same registration ID when a pull diagnostics report is applied.

Additionally, various deficiencies in our support for pull diagnostics have been fixed:
- Document pulls are issued for all open buffers, not just the edited one. A shorter debounce is used for the edited buffer. Workspace diagnostics are also now ignored for open buffers.
- Tracking of `lastResultId` is improved.
- Stored pull diagnostics are discarded when the corresponding buffer is closed.

Release Notes:

- Improved compatibility with language servers that use the "pull diagnostics" feature of Language Server Protocol.

Issue #37838, opened 2025-09-09
LSP: Handling for multiple registrations on the same method is incorrect


reflectronic:
### Summary

Currently, when we receive a dynamic registration, we overwrite any information from previous registrations on the same method.

https://github.com/zed-industries/zed/blob/11d81b95d45f73d277e17b9df53e9ad359937175/crates/project/src/lsp_store.rs#L11977-L11988

The LSP specification is, unfortunately, not clear on how exactly a client should respond to multiple registrations on the same method (https://github.com/microsoft/language-server-protocol/issues/1069). But, the behavior of the VS Code language client suggests that we should be sending separate messages to the server _for each registration_. For example, if a server registers for `textDocument/didChange` twice, VS Code sends _two_ notifications to the server each time a document is changed.

(n.b.: The reason it works this way is that, each time a server sends a registration for `textDocument/didChange`, VS Code adds another set of "text changed" event handlers that result in notifications being sent.)

Some language servers are already making use of these semantics. For example, the C# language server uses multiple registrations for `textDocument/diagnostic`. This allows it to use one request to compute only syntax errors (which is fast), another to compute semantic errors (which is slower), a third to run code analysis plugins, etc. This way, syntax errors do not have to wait for potentially slow code analyzers to complete before being returned to the user. Since Zed only remembers the last registration, diagnostics aren't currently working properly with the C# language server.

### Zed Version and System Specs

Zed: v0.204.0 (Zed Dev 57c6dbd71e483646cd0409894547e97f664ebed3) (Taylor's Version)
OS: Windows 10.0.26100
Memory: 31.1 GiB
Architecture: x86_64
GPU: AMD Radeon RX 6600 || AMD Corporation || Unknown Driver

smitbarmase:
I started work on this, taking a similar approach to VSCode:

The idea is to:
- Store capabilities in a `HashMap` keyed by registration ID with their options. This way, we keep track of all registrations for a given capability, and only remove the options tied to an unregister request, without wiping out the entire capability when a single registration is removed.
- With multiple options per capability, we can then decide the effective capability based on custom logic:
  1. For example like you said, in C#, we can notify server twice when capabilities are registered twice.
  2. See `impl EffectiveCapability for cap::DidChangeTextDocument` and `impl EffectiveCapability for cap::DidSaveTextDocument` it is similar to how VSCode does it.

Not sure when I’ll pick this back up, or if anyone wants to take over. Just wanted to document it here.

[...]

PR #55340, opened 2026-04-30
project: Track inlay hint / code lens / document symbol registrations by ID

tautik:
extends the precedent set by #43703 (thanks @reflectronic) so that `textDocument/inlayHint`, `textDocument/codeLens`, and `textDocument/documentSymbol` correctly track multiple dynamic registrations by id. previously each register call overwrote the slot on `ServerCapabilities`, and unregister either silently did nothing (inlayHint, documentSymbol — no match arm) or wiped the entire capability (codeLens). now each method keeps its own id-keyed map on `DynamicRegistrations`; the field on `ServerCapabilities` is cleared only when the last registration is removed.

textDocument-sync notifications (`didChange`/`didSave`/`willSave`) are deliberately out of scope — their fan-out semantics are ambiguous in the spec and were flagged on @smitbarmase's earlier exploration in #36876.

verified with `cargo check -p project -p lsp -p editor`, `cargo test -p project --test integration -- multi_registration` (3/3 passing), and `cargo fmt -p project`.

closes part of #37838.

Release Notes:

- Improved support for language servers that dynamically register inlay hints, code lens, or document symbols multiple times.


SomeoneToIgnore (approved):
I had to fix the original submission due to various registration bugs (e.g. static capabilities were not restored properly on final unregistration), but then realized I have to fix it all one way or another, so why not now...

Pushed the fixes to properly [un]register all the dynamic capabilities, unified that into mehods, and fixed whatever refresh-related issues.
Also added more tests.

Thank you for motivating, will merge after tomorrow's release.

PR #55341, opened 2026-04-30
javascript, typescript: Inject GLSL and WGSL syntax in comment-labeled template literals

foretoo:
Extends the existing ECMAScript comment-label injection system to support GLSL and WGSL shader languages inside JS/TS strings and template literals.

The pattern `/* language */` before a string is already used for `html`, `sql`, `graphql`, and `css`. This adds the same support for two shader languages commonly embedded as strings in WebGL/WebGPU code:

before:
Screenshot 2026-04-30 at 23 07 05

after:
Screenshot 2026-04-30 at 23 07 39

Syntax highlighting activates when the corresponding extension is installed:
- GLSL: the built-in `glsl` extension
- WGSL: the community extension [`wgsl-wesl-zed`](https://github.com/lucascompython/wgsl-wesl-zed) (language name `WGSL/WESL`)

Both `/*glsl*/` and `/* glsl */` forms are accepted (same behavior as the existing patterns).

Release Notes:

- Added `/*glsl*/` and `/*wgsl*/` comment-label syntax injection for JavaScript and TypeScript template literals

### DIFF
--- a/crates/grammars/src/javascript/injections.scm
+++ b/crates/grammars/src/javascript/injections.scm
@@ -142,3 +142,25 @@
   ])
   (#match? @_ecma_comment "^\\/\\*\\s*(css)\\s*\\*\\/")
   (#set! injection.language "css"))
+
+; '/* glsl */' or '/*glsl*/'
+(((comment) @_ecma_comment
+  [
+    (string
+      (string_fragment) @injection.content)
+    (template_string
+      (string_fragment) @injection.content)
+  ])
+  (#match? @_ecma_comment "^\\/\\*\\s*glsl\\s*\\*\\/")
+  (#set! injection.language "glsl"))
+
+; '/* wgsl */' or '/*wgsl*/'
+(((comment) @_ecma_comment
+  [
+    (string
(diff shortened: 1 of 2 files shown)


versecafe:
What's the reason there's not a standardized optional comment syntax for injecting any arbitrary language highlights into multiline strings the same way jebtrains does, it feels odd to need to manually list every language inside each language for this to work

foretoo:
@versecafe
I agree, it would be far more convenient to have a standardized approach like this. I think the difficulty lies in linking different LSPs that come as third-party extensions, along with the various regex patterns involved. For example, in my PR, the `/*glsl*/` pattern maps to the `glsl` language name, 100% matching, but the `/*wgsl*/` pattern maps to the `WGSL/WESL` language name defined by a third-party extension, completely different naming way. This definitely could be standardized somehow; we need a discussion on this topic.

tomhoule (approved):
I agree that we'll need a more principled solution at some point. In the meantime, this is a straightforward improvement, so let's merge. Thank you for the contribution!

PR #55353, opened 2026-04-30
Fix split diff not activating with DiffViewStyle::Split setting

nandak522:
## Summary

- The `minimum_split_diff_width` default of 100 em-widths (~500-600px at typical font sizes) was too aggressive, causing `width_changed()` to auto-unsplit immediately after construction's deferred `split()` call, silently reverting to unified mode even when `DiffViewStyle` was set to `Split`.
- Lowered the default from 100 to 68 em-widths (~340-400px), allowing split diffs to activate in typical editor panel widths.
- Added a `log::info` message when auto-unsplit happens so users can diagnose why split diff isn't active and learn how to disable auto-switching (`minimum_split_diff_width: 0`).

**Root cause:** When `SplittableEditor` is constructed with `DiffViewStyle::Split`, it defers a `split()` call via `window.defer()`. On the first render, a `canvas` element detects the width and also defers `width_changed()`. Since the constructor's defer runs first, `split()` succeeds, but `width_changed()` then fires and — finding the panel narrower than 100 em-widths — immediately calls `unsplit()`, undoing the split. Most editor panels are narrower than 100 em-widths, so users with `DiffViewStyle: Split` would always see unified diffs.

## Test plan

- [x] `cargo check -p editor` — compiles clean
- [x] `cargo test -p editor -- split` — 40 passed, 0 failed
- [x] `cargo clippy -p editor -- -D warnings` — no warnings
- [ ] Set `Diff View Style` to `Split` in settings, open a git diff (Project Diff), verify split view appears
- [ ] Resize window narrow enough to trigger auto-switch — verify unified mode activates and log message appears
- [ ] Resize back — verify split mode re-activates automatically

Release Notes:

- Fixed split diff not activating when `Diff View Style` is set to `Split` by lowering the `minimum_split_diff_width` default from 100 to 68 columns

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -324,8 +324,8 @@
   // switches to unified mode and switches back when the editor is wide
   // enough. Set to 0 to disable automatic switching.
   //
-  // Default: 100
-  "minimum_split_diff_width": 100,
+  // Default: 68
+  "minimum_split_diff_width": 68,
   // Show method signatures in the editor, when inside parentheses.
   "auto_signature_help": false,
   // Whether to show the signature help after completion or a bracket pair inserted.
--- a/crates/editor/src/split.rs
+++ b/crates/editor/src/split.rs
@@ -4,6 +4,7 @@ use std::{
 };
 
 use buffer_diff::{BufferDiff, BufferDiffSnapshot};
+use log::info;
 use collections::HashMap;
 
 use gpui::{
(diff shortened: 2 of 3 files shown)


nandak522:
@cla-bot check

agu-z:
Thanks for your contribution, but we won't be merging this change.

I tried a few different values, and I think we already have a good default. Anything <100 cols feels too cramped for me. I know this is subjective, but that's why we have a setting that people can change.

Furthermore, the logging added seems unnecessary when we already have a UI affordance:

CleanShot 2026-05-19 at 15 11 03@2x

PR #55361, opened 2026-04-30
editor: Add base64 encode/decode commands

chriskievit:
Self-Review Checklist:

Adds `editor: convert to base64` and `editor: convert from base64` to the command palette. Both commands operate on the current selection, or the word under the cursor when nothing is selected. The decode command silently no-ops on invalid base64 input or non-UTF-8 decoded bytes, consistent with how other convert commands handle untransformable input.

Release Notes:

- Added `editor: convert to base64` and `editor: convert from base64` commands to the command palette


ChristopherBiscardi (approved):
works as expected, thanks!

PR #55365, opened 2026-04-30
helix: Fix line selection after undoing delete

### LINKED ISSUE #55170: Helix: v selects one line too many after d
### Reproduction steps

Imagine some text:

Image

Press `gl` and then`l` to move to the last character (= the line break):

Image

If you'd now press `x`, it would select the whole line (as expected):

Image

But let's clear the selection again by pressing `;` as this was just to illustrate the problem later:

Image

...instead, press `d` to delete that new line char:

Image

...and now, press `u` to undo the previous deletion:

Image

Looks like we're back at the start, right? But not quite.

#### Current behavior

If we press `x` again, it now wrongly selects one line too many:

Image

#### Expected behavior

Like in original Helix, I'd expect only the current line to be selected:

Image

Zed: v0.233.10+stable.250.350f33871e84d5ed83a403523e9490b9bf014a94 (Zed) 
OS: macOS 26.4.1
Memory: 24 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

GoldStrikeArch:
Closes #55170

Fixes Helix mode undo selection history after `d` deletes the character under an empty cursor. Previously, `d` temporarily expanded the cursor to a one-character selection before deleting, and undo restored that temporary selection. When the deleted character was a newline, pressing `x` after undo selected additional line which was wrong.

This preserves the original Helix selection as the undo-restored selection while keeping the existing delete behavior unchanged.

Self-Review Checklist:

Release Notes:

- Fixed Helix mode selecting one extra line after undoing a newline delete


crates/vim/src/helix.rs
```diff
@@ -2829,6 +2829,10 @@ mod test {
             ˇ»line five"},
             Mode::HelixNormal,
         );
+
+        cx.set_state("oneˇ\ntwo\nthree", Mode::HelixNormal);
+        cx.simulate_keystrokes("d u x");
+        cx.assert_state("«one\nˇ»two\nthree", Mode::HelixNormal);
```
tomhoule, line 2835: Isn't the expected behaviour `cx.assert_state("«oneˇ»\ntwo\nthree", Mode::HelixNormal);` (selection and cursor on the same line)? I just checked in Helix. This assertion seems to match what I see in Zed (current main).

GoldStrikeArch: Yes, the cursor is on a newline character but the visible selection is on a character before it. This is how it is right now (somewhat hacky implementation of a helix mode itself). I had some proposal how all of that could be refactored here, if you are interested -> https://github.com/zed-industries/zed/discussions/56768

> This assertion seems to match what I see in Zed (current main).

In the current main branch you will end up selecting both `one` and `two` in this example:
```
oneˇ
two
three
```
I think it was this commit which is responsible for it -> https://github.com/zed-industries/zed/commit/81da953acfdcba1951875abb9664cd371d4c7f86

GoldStrikeArch: To be more precise in our concrete example it would be this code -> https://github.com/zed-industries/zed/blob/main/crates/editor/src/element.rs#L152 but its all triggered because of this:

```
 cursor_offset_on_selection: self.mode.is_visual() || self.mode.is_helix(),
 ```

tomhoule: Indeed, I probably wasn't using the right sequence, I can see the difference with main now.

tomhoule (approved):
This is an improvement, and the implementation makes sense. Thanks for the contribution!

PR #55386, opened 2026-05-01
search: Persist buffer search and replace history across sessions

wilsonglasser:
## Summary

- Adds a SQLite-backed store for the buffer search bar so recent **find** and **replace** queries persist across sessions, exposed via a dropdown next to each input (capped at 50 entries per kind).
- New `search.history_enabled` setting (default `true`) gates persistence; toggling it off also wipes any previously stored entries from disk.
- Find-query writes are debounced (1s) so prefixes of an in-progress query (`"f"`, `"fo"`, `"foo"`) don't pollute the history. Replace entries are only recorded after an actual replacement is performed.
- `ContextMenu` gains an opt-in vertical scrollbar (`with_vertical_scrollbar`) so the new dropdowns stay usable when the list grows past the menu's max height.

## Test plan

- [ ] Open the buffer search bar, run several queries, restart Zed, and verify recent searches show up in the new dropdown.
- [ ] Type a long query slowly and confirm only the final query is persisted (no prefix pollution).
- [ ] Trigger `replace_next` and `replace_all` and verify the replacement dropdown shows what was replaced.
- [ ] Toggle `search.history_enabled` to `false` in settings and verify the in-memory + on-disk history is cleared.
- [ ] `cargo test -p search` (covers new replacement-history and disable-clears-history tests) and `cargo test -p search persistence` (covers the new SQLite store).


wilsonglassertbl:
@cla-bot check

wilsonglassertbl:
image
image

assets/settings/default.json
```diff
@@ -700,6 +700,10 @@
     "regex": false,
     // Whether to center the cursor on each search match when navigating.
     "center_on_match": false,
+    // Whether to remember recent search and replace queries across sessions.
+    // When enabled, queries are stored on disk so they survive restarts.
+    // When disabled, any previously stored history is also cleared from disk.
+    "history_enabled": false,
```
SomeoneToIgnore, line 706: I wonder if there's any reason not to remember those?
We have to edit `all-settings.md` too, if there's a reason and we keep this setting for it, as that Markdown document gets into the website docs.

The name is also somewhat odd, it's more `persistence_enabled` than `history` as the latter works even now.

crates/project/src/search_history.rs
```diff
@@ -51,6 +51,21 @@ impl SearchHistory {
         }
     }
 
+    /// Replaces the history with the given entries, ordered from oldest to
+    /// newest. Used to seed the in-memory history from a persisted store
+    /// without triggering insertion-behavior side effects. Any existing
+    /// `SearchHistoryCursor` should be reset by the caller after this call.
```
SomeoneToIgnore, line 57: >  Any existing `SearchHistoryCursor` should be reset by the caller after this call.

Such contracts look rather binding, and odd that here we seem to initialize the history only?
Can we add another constructor/parameter to the existing one instead and instantiate the data without extra contracts?

crates/project/src/search_history.rs
```diff
@@ -51,6 +51,21 @@ impl SearchHistory {
...
+    pub fn bulk_load(&mut self, entries: Vec<String>) {
+        self.history.clear();
+        let cap = self.max_history_len.unwrap_or(usize::MAX);
+        let skip = entries.len().saturating_sub(cap);
+        self.history.extend(entries.into_iter().skip(skip));
+    }
+
+    pub fn clear(&mut self) {
```
SomeoneToIgnore, line 65: This one would benefit from the new comment above, about the cursor invalidation.
I wonder is there a way to somehow do that without any extra implicit contracts, though?

crates/search/src/buffer_search.rs
```diff
@@ -870,7 +902,23 @@ impl BufferSearchBar {
...
 
-        let search_options = SearchOptions::from_settings(&EditorSettings::get_global(cx).search);
+        let editor_search_settings = EditorSettings::get_global(cx).search;
+        let search_options = SearchOptions::from_settings(&editor_search_settings);
+        let history_persistence_enabled = editor_search_settings.history_enabled;
+
+        let (initial_find_history, initial_replacement_history) = if history_persistence_enabled {
+            let db = BufferSearchHistoryDB::global(cx);
+            (
+                db.list(HistoryKind::Find).unwrap_or_default(),
```
SomeoneToIgnore, line 912: This looks dangerous: is it a synchronous db data read on the rendering thread?
I would initialize the history asynchronously, if we restore it from the database.

crates/search/src/buffer_search.rs
```diff
@@ -1860,8 +1930,130 @@ impl BufferSearchBar {
...
+                this.search_history
+                    .add(&mut this.search_history_cursor, query.clone());
+                this.persist_history_entry(HistoryKind::Find, query, cx);
+                this.pending_search_history_save = None;
+            })
+            .ok();
+        }));
+    }
+
+    fn record_replacement_history(&mut self, value: String, cx: &mut Context<Self>) {
```
SomeoneToIgnore, line 1957: So, the replacement entries are never debounced, only the search ones?
Any reason to have it this way?

crates/search/src/buffer_search.rs
```diff
@@ -1860,8 +1930,130 @@ impl BufferSearchBar {
...
+        self.persist_history_entry(HistoryKind::Replace, value, cx);
+    }
+
+    fn persist_history_entry(&self, kind: HistoryKind, value: String, cx: &mut Context<Self>) {
+        if !self.history_persistence_enabled || value.is_empty() {
+            return;
+        }
+        let db = BufferSearchHistoryDB::global(cx);
+        cx.background_spawn(async move { db.record(kind, value).await })
+            .detach_and_log_err(cx);
```
SomeoneToIgnore, line 1972: Let's avoid `detach` at all: this makes the persisting task never cancellable which is good and bad at the same time, but for now seems better to be explicit and return a task from here.

crates/search/src/buffer_search.rs
```diff
@@ -1860,8 +1930,130 @@ impl BufferSearchBar {
...
+        };
+        let weak_self = cx.entity().downgrade();
+        let trigger = IconButton::new(id, IconName::ChevronDown)
+            .shape(IconButtonShape::Square)
+            .icon_size(IconSize::Small)
+            .icon_color(Color::Muted)
+            .style(ButtonStyle::Subtle)
+            .disabled(entries.is_empty());
+
+        PopoverMenu::new(id)
```
SomeoneToIgnore, line 2005: One very prominent feature request is to be able to remove search results.
With persistence, that seems even more important, can we somehow implement it?

crates/ui/src/components/context_menu.rs
```diff
@@ -519,6 +527,14 @@ impl ContextMenu {
         self.items.push(item.into());
     }
 
+    /// Renders a vertical scrollbar on the right side of the menu, using the
+    /// menu's existing max-height. Useful when the entry list may overflow.
+    pub fn with_vertical_scrollbar(mut self) -> Self {
+        self.show_vertical_scrollbar = true;
+        self.scroll_handle.get_or_insert_with(ScrollHandle::default);
```
SomeoneToIgnore, line 534: Is it so that 1 optional `scroll_handle` field is enough already to encode this "scroll enabled/disabled" state?

If so, let's simplify?

crates/ui/src/components/context_menu.rs
```diff
@@ -233,6 +233,8 @@ pub struct ContextMenu {
     submenu_trigger_bounds: Rc<Cell<Option<Bounds<Pixels>>>>,
     submenu_trigger_mouse_down: bool,
     ignore_blur_until: Option<Instant>,
+    show_vertical_scrollbar: bool,
+    scroll_handle: Option<ScrollHandle>,
```
SomeoneToIgnore, line 237: Can you please record a video of this UI change?

https://github.com/zed-industries/zed/pull/55386#issuecomment-4357922600 seems to have no scrollbars.

crates/search/src/persistence.rs
```diff
@@ -0,0 +1,139 @@
...
+            query TEXT NOT NULL
+        ) STRICT;
+        CREATE INDEX IF NOT EXISTS search_history_kind_id ON search_history(kind, id);
+    )];
+}
+
+db::static_connection!(BufferSearchHistoryDB, []);
+
+impl BufferSearchHistoryDB {
+    pub async fn record(&self, kind: HistoryKind, query: String) -> Result<()> {
```
SomeoneToIgnore, line 42: Right now, the history is stored in memory per-window.

This makes it semi-global, as now windows will get the entire history on restart, and append history from all windows — but never share the new, appended data until next restart.

This seems rather confusing and not the intended way to work?

Also, seems that there's no good deduplication of the queries?
This worth a test at least.

SomeoneToIgnore (changes requested):
Very sorry for the late review: the PR sort of confused me and the topic is relatively complex, both took time to think through.

Let me know if you have any plans to continue with the feedback, I seem to have the context for now to be much faster with the reviews.

wilsonglasser:
Thanks for the feedback, I will look for some time to remember the PR this week and return each point.

SomeoneToIgnore:
It's been a bit over 3 weeks with no changes, conflicts start to creep in — will close it for now to keep the PR queue cleaner.

That feature is definitely something we'd better have, but implemented differently.

PR #50197, opened 2026-02-26
workspace: Add detachable panel windows with WorkspaceSatellite

iam-liam:
# PR: workspace: Add detachable panel windows with WorkspaceSatellite

Adds the ability to detach any panel item (starting with terminals) into a standalone OS window. The underlying entity is shared — same PTY session, same project context, no duplication. Users with multiple monitors can put their terminal on a second screen while keeping the editor focused.

Addresses #9662 and #5229.

## What this does

- **`DetachActiveItem` action** (command palette + tab context menu) pops the active item into a new OS window via `WorkspaceSatellite`
- **`ReattachActiveItem` action** (context menu in satellite) returns the item to its original location (dock panel or center pane)
- Dock panel auto-hides when all terminals are detached, reopens on reattach
- Closing the satellite window triggers cleanup via event subscription
- Satellite windows are isolated from workspace pane routing — clicking files in the project panel stays in the main window
- Non-functional toolbar buttons (new file, split, zoom) are hidden in satellite panes
- "Detach" context menu entry hidden when already in a satellite window
- Satellite window title bar matches main window styling (tab bar background)

## Architecture

- **`WorkspaceSatellite`** (`workspace_satellite.rs`) — lightweight window container with its own `PaneGroup`, subscribed to pane events for cleanup
- **`ItemOrigin`** tracking — records whether an item came from a center pane or dock panel, so reattach puts it back in the right place
- Satellite panes are tracked in `Workspace.panes` but not subscribed to workspace's `handle_pane_event`, keeping them isolated from focus/activation routing
- The satellite pane's `in_satellite` flag controls UI differences (hidden toolbar buttons, "Move to Main Window" context menu entry)

## Scope

Five files, purely additive:

| File                                          | Change                                                                |
| --------------------------------------------- | --------------------------------------------------------------------- |
| `crates/workspace/src/workspace_satellite.rs` | New — satellite window view                                           |
| `crates/workspace/src/workspace.rs`           | Satellite lifecycle, detach/reattach, item origin tracking, 4 tests   |
| `crates/workspace/src/pane.rs`                | `in_satellite` flag, hide toolbar buttons, "Move to Main Window" menu |
| `crates/terminal_view/src/terminal_panel.rs`  | Detach action handler, dock auto-hide/reopen, tests                   |
| `crates/terminal_view/src/terminal_view.rs`   | Context menu entry for detach                                         |

## Current limitations (planned follow-ups)

[...]

Issue #9662, opened 2024-03-21
Add secondary editor windows to support multi-monitor setup


abey79:
### Check for existing issues

- [X] Completed

### Describe the feature

One thing that I greatly enjoy with JetBrain IDEs (and, I believe, was added to VSCode not so long ago) is the ability to open a secondary windows for editor tabs only, to support multi-monitor setup.

FWIW, I typically use a horizontal + vertical monitor setup where:
- the main project window, along with one loooong editor tab and the terminal go to the vertical screen
- a side-by-side, dual editor window goes to the horizontal screen

thorfour:
Really is making/breaking zed for me to be able to have it up on multiple monitors. Want to be able to have an editors on all of my vertical monitors.

0xBYTESHIFT:
Is it even planned?

AvdulRash:
Any update one this ?

ser:
The worrisome thing is that there are no reactions to this ticket at all...

jcelliott:
I use a setup like this as well, but I often put "panels", not just editors, in the second window.

janek:
I found a workaround in Discussions, it helped me: https://github.com/zed-industries/zed/discussions/12074#discussioncomment-12472029

It would be really nice to get an update on this feature, for me it's an important one and it seems to have a lot of support (here and in duplicate https://github.com/zed-industries/zed/issues/17618 and in possible other duplicates)

danielytics:
Now with the new agent workflow, I would really like to pop the agent dock out into its own window so I can display it fullscreen on my second portrait monitor. Even with a large main monitor, I find the agent window too cramped to comfortably read its messages and making it wider eats up too much of my editor space.

This would _really_ help me, since I use quite large font sizes due to poor eyesight.

MarkParker5:
Possible workarounds:
- External terminal. iTerm2 with "[Open with editor](https://stackoverflow.com/questions/38906815/can-i-click-on-a-filename-and-line-number-in-iterm2-to-open-that-file-in-visual)"="Zed" and oh-my-zsh (+ OneDark theme and Zed Plex Mono Font for aesthetic) is pretty decent.
- For editor, panels, and the agent, Zed window can be resized to take both monitors. Add a second panel and move the separator to the edge of the monitor. On macOS, requires "Displays have separate spaces" off and horizontal display arrangement.
[...]

Issue #5229, opened 2022-09-02
Allow tabs to be moved between Zed instances


JosephTLyons:
### Check for existing issues

- [X] Completed

### Is your feature request related to a problem?

-

### Describe the solution you'd like

I think it would be nice if you could drag tabs between instances of Zed.  If the edit history could travel to the new instance, that would be a plus.  It seems like VS Code doesn't support moving the edit history (it is simply lost), but Sublime Text does.

dumski:
Hi, is this planned to implement anytime soon? When working on multiple monitors the feature seems really necessary. Or maybe there is another way to work on one project using multiple windows?

fluffy-critter:
I am really surprised to see that there's no way to drag an editor tab between windows, or to move an editor tab into a new window. It feels like a pretty major UX shortcoming to me.

everdrone:
This is a much needed feature, especially for people working with more than one monitor per project

ZombieHarvester:
Would be great if assistant pane and terminal could be moved to a separate window for multi monitor set up

dlight:
> Would be great if assistant pane and terminal could be moved to a separate window for multi monitor set up

That would be awesome! But I think that a Zed terminal or assistant panel in another window shouldn't be in another instance, but continue to be linked in the same instance. So a Zed instance should be comprised of potentially multiple windows, instead of the current (and simpler) model of one window per instance.

That's different than moving a tab from one instance to another, because each instance is supposed to be a different project, with its own per-project config, its own set of LSP servers, etc

ZombieHarvester:
> continue to be linked in the same instance

Exactly what I had in mind: not a full-featured Zed window, but a single-panel mini window still connected to the project and draggable to another monitor.

krstp:
Any ETA on when we can expect this functionality? Thanks!

PR #55404, opened 2026-05-01
workspace: Add action to move the active item to a new window

iam-liam:
## Problem

Multi-monitor users want to put a single editor tab or terminal on a second display without giving up Zed's project context. Today the only way is to open another full project window, losing the shared `Entity<Project>` and forcing duplicate workspaces.

There's been steady demand for it:

- [#9662 _Open editor tabs in a secondary window_](https://github.com/zed-industries/zed/issues/9662) — opened Mar 2024, **306 👍 / 27 👀**, the staff ranked it among the most-requested feature gaps.
- [#5229](https://github.com/zed-industries/zed/issues/5229) — same shape, longer-running thread.
- The previous attempt ([#50197](https://github.com/zed-industries/zed/pull/50197)) kept drawing comments months after it was closed — 20 👍 / 10 ❤️ / 4 🚀 on the PR, plus unsolicited "any update?" pings from `jmevel`, `marcoafm26`, `leferi99`. It was closed for the wrong architecture (`WorkspaceSatellite`), not the wrong feature. Cameron's close note: _"Multi-window in general is definitely something that we're keen to support."_

This PR is the rewrite against that feedback.

## Prior art

The interaction model exists in every major editor:

- **VS Code** — _Move Editor into New Window_ (and _View → Editor Layout → Move Editor into New Window_).
- **JetBrains IDEs** — _Window → Detach Tab_.
- **Sublime Text / Atom** — drag a tab off the tab bar.

Zed adapts the same UX: a tab (or any item that lives in a `Pane`) can be moved into its own window, and brought back. The new window is a real `Workspace` over the same `Project`, so all the existing workspace machinery (sidebar, dock, project-switcher) works inside it.

## What this does

Adds two actions:

- `workspace::MoveActiveItemToNewWindow` — moves the focused pane's active item into its own window.
- `workspace::ReattachActiveItemToSourceWindow` — sends the item back to its source pane and closes the detached window.

Surfaces:

| Surface                                     | Coverage                                                   |
| ------------------------------------------- | ---------------------------------------------------------- |
| Tab right-click menu — "Move to New Window" | Center pane, dock panels with tab strips (terminal, agent) |
| Command palette                             | Both actions, all surfaces                                 |
| Default keybinding                          | None — users bind their own                                |

The detached window's title suffix is the moved item's tab text (`myproject — main.rs`, `myproject — Terminal: zsh`) so it's distinguishable in the macOS Dock and `cmd+~` switcher.

## Why this approach

Cameron's review of #50197 surfaced three structural critiques. Each is addressed by the architecture here:

1. _"Satellite windows are the wrong model — a second window should be a normal `Workspace`."_
   The detached window is a normal `MultiWorkspace + Workspace` sharing the source's `Entity<Project>`. No new top-level types. `MultiWorkspace::new` is reused unchanged.
2. _"GPUI now lets entities be shared across windows directly — proxy machinery is unnecessary."_
   The item's underlying entity (`Entity<Editor>`, `Entity<TerminalView>`, `Entity<Terminal>`) is referenced from views in both windows. State continuity is by entity sharing, not by serialisation: PTYs keep running, buffers keep their undo history.
3. _"Don't special-case panels — should also work for tabs and integrate with the multi-window infrastructure."_
   `Workspace::focused_pane` resolves the source pane uniformly across the center pane group and any panel that exposes `Panel::pane()`. The action is the same code path for tabs and dock-panel items, and it sits next to `MoveProjectToNewWindow` in the same action family.

No `WorkspaceSatellite`. No sqlite round-trip. No changes to `MultiWorkspace::new`, the `Dock`, or the `Panel` trait.

## What this doesn't do

- **Persistence** — detached windows are ephemeral; closing one destroys it. The follow-up extends the existing sqlite-backed `MultiWorkspace` persistence.
- **Drag-out detach** — drag-a-tab-out-of-the-titlebar isn't wired up. The design supports it via the same `move_active_item_to_new_window` entry point.
- **Chromeless windows** — the detached window has the full normal chrome (sidebar, dock, title bar). The user can drag more items in; it becomes a real second workspace over the same project. This was deliberate: a `chromeless: bool` flag would re-introduce the panel/satellite-style special-casing #50197 was rejected for.

## Cross-cuts

| Concern                        | Status                                                                                                                                                                                                                                      |
| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Vim mode**                   | No vim keybindings overridden. The actions are `workspace::*`-namespaced; vim doesn't intercept. State (mode, register, marks) is held on the editor's entity and follows the move.                                                         |
| **Remote / collab projects**   | The shared `Entity<Project>` is the same regardless of local/remote, so collab/SSH projects should work without changes. I haven't tested on collab — happy to do so before merge if a maintainer can confirm a test setup.                  |
| **Workspace trust**            | The detached `Workspace` shares the same `Entity<Project>`, so trusted-worktree state is inherited automatically. No new trust surface introduced.                                                                                          |
| **Settings / themes / keymap** | No new settings. No theme changes. No default keybinding. The detached window inherits the active theme, so light and dark mode are covered by construction.                                                                                |
| **Accessibility**              | Command palette covers every case. Tab right-click covers center-pane tabs and dock panels with internal tab strips. Panels without a tab strip rely on the command palette — easy to add per-panel menu entries in a follow-up if desired. |
| **Performance**                | Action runs on user input; not in any render or input-event hot path.                                                                                                                                                                       |
| **Platforms**                  | macOS verified end-to-end. The feature uses only `cx.open_window`, which is cross-platform, so Linux and Windows should work — I haven't been able to test them. Happy to do so before merge if it's a blocker.                             |

## Editor coexistence

State that lives on the editor's entity follows the move automatically because the entity is the same across windows:

- Multi-buffer items (project search, find-all-references) — entity-shared, preserved.
- Dirty state, undo history, folds — `MultiBuffer`/`Editor` state, preserved.
- LSP popovers / hover state — re-attached via the editor's normal lifecycle on `added_to_workspace`.
- Inline blocks, edit predictions, gutter elements — held on the editor entity, follow the move.

Manual repro covered: editor with unsaved changes, editor with active find-all-results, terminal running `tail -f`. Everything keeps state. LSP popovers / find-all matrix-testing pending.

## Incidental fixes

Three supporting changes — two required for the feature, one pre-existing latent bug surfaced during testing.

1. **Required — defer window creation.** The action handler runs inside the source workspace's `update`. Building the new window synchronously means the source workspace is borrowed while the new window paints, and any element on the moved item that touches the source workspace would double-lease. The actual panic stack from manual testing fingered `debugger_ui::register_action_renderer` (`crates/debugger_ui/src/debugger_ui.rs:269`) reading `editor.workspace()`. `window.defer(cx, ...)` lets the source's update finish first.
2. **Required — `window.activate_window()` after `cx.open_window`.** Production `build_window_options` returns `show: false, focus: false`, so without an explicit `activate_window` the new window is created but never revealed.
3. **Pre-existing latent bug — editor's window-activation observer never followed the editor between windows** (`crates/editor/src/items.rs`; new `_window_activation_subscription: Option<Subscription>` field on `Editor`). The observer was registered once at construction against the editor's first window. The detach feature was the first thing to actually move editors between windows and exposed it, but the same bug would surface for any future feature that does so (split-window UX, follow-mode revisions, etc.). Replacing the field on `added_to_workspace` rather than detaching keeps exactly one live observer per editor and resets `mouse_cursor_hidden` for the new window.

Happy to split this into a precursor PR if you'd prefer to land it independently.

## Files changed

| File                                | Change                                                                                                                                                                                                                                                                                                            |
| ----------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `crates/workspace/src/workspace.rs` | New `MoveActiveItemToNewWindow` and `ReattachActiveItemToSourceWindow` actions; `Workspace::move_active_item_to_new_window` and `Workspace::reattach_active_item_to_source_window`; `detached_from: Option<DetachedFrom>` field; window-title suffix for detached workspaces; action handler wiring; 2 unit tests |
| `crates/workspace/src/pane.rs`      | "Move to New Window" entry in the tab right-click menu                                                                                                                                                                                                                                                            |
| `crates/editor/src/editor.rs`       | `_window_activation_subscription: Option<Subscription>` field; removed the construction-time observer                                                                                                                                                                                                             |
| `crates/editor/src/items.rs`        | `Editor::added_to_workspace` re-registers the window-activation observer for the current window and resets `mouse_cursor_hidden`                                                                                                                                                                                  |

## Demo

https://github.com/user-attachments/assets/3cabd87a-e369-4f5b-9abf-fd240d26ae36

The recording walks through:

1. Open multiple editor tabs, detach one, reattach.
2. Same flow with a terminal pane.
3. Detach several windows, show the macOS Window menu listing them as `zed — <item>`.
4. Reattach everything.

## Tests

- `crates/workspace/src/workspace.rs::tests::test_move_active_item_to_new_window` — verifies the source pane is empty and the detached window holds the item with `detached_from` set.
- `crates/workspace/src/workspace.rs::tests::test_reattach_active_item_to_source_window` — verifies reattach returns the item to the source pane.
- `cargo nextest run -p workspace --lib` — 196 pass.
- `cargo nextest run -p editor --lib` — 723 pass (no regressions from the editor subscription change).
- `./script/clippy -p workspace -p editor -p gpui` — clean.

Release Notes:

- Added action to move the active tab or dock-panel item into its own window

### DIFF
--- a/crates/editor/src/items.rs
+++ b/crates/editor/src/items.rs
@@ -1027,6 +1027,34 @@ impl Item for Editor {
             .detach();
         }
 
+        // The window-activation observer is bound to a specific window, so it
+        // has to follow the editor when it moves between windows (e.g. via
+        // `MoveActiveItemToNewWindow`). Replacing the field drops any prior
+        // subscription, keeping exactly one live observer per editor.
+        self.show_mouse_cursor(cx);
+        let blink_manager = self.blink_manager.clone();
+        blink_manager.update(cx, |blink_manager, cx| {
+            if window.is_window_active() {
+                blink_manager.enable(cx);
+            } else {
+                blink_manager.disable(cx);
+            }
+        });
+        self._window_activation_subscription =
+            Some(cx.observe_window_activation(window, |editor, window, cx| {
+                let active = window.is_window_active();
+                editor.blink_manager.update(cx, |blink_manager, cx| {
+                    if active {
+                        blink_manager.enable(cx);
+                    } else {
+                        blink_manager.disable(cx);
+                    }
+                });
+                if active {
+                    editor.show_mouse_cursor(cx);
+                }
+            }));
+
         // Load persisted folds if this editor doesn't already have folds.
         // This handles manually-opened files (not workspace restoration).
         let display_snapshot = self
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -283,6 +283,17 @@ actions!(
         FollowNextCollaborator,
         /// Moves the focused panel to the next position.
         MoveFocusedPanelToNextPosition,
+        /// Moves the active item out of its pane and into its own window.
+        ///
+        /// Works for tabs in the center pane group and items inside dock panels
+        /// (any panel whose `Panel::pane()` returns `Some`). The detached window
+        /// is a normal `Workspace` sharing the same `Project`; state continuity
+        /// comes from sharing the underlying `Entity<T>` directly.
+        MoveActiveItemToNewWindow,
+        /// In a window opened via `MoveActiveItemToNewWindow`, returns the
+        /// active item back to the source workspace's active pane and closes
+        /// the detached window.
+        ReattachActiveItemToSourceWindow,
         /// Creates a new file.
         NewFile,
         /// Creates a new file in a vertical split.
@@ -1398,6 +1409,15 @@ pub struct Workspace {
     multi_workspace: Option<WeakEntity<MultiWorkspace>>,
     active_worktree_creation: ActiveWorktreeCreation,
     deferred_save_items: Vec<Box<dyn WeakItemHandle>>,
+    /// Set when this workspace was opened by `MoveActiveItemToNewWindow`.
+    /// Reattach uses this to send the active item back to the source pane.
+    detached_from: Option<DetachedFrom>,
+}
+
+#[derive(Clone)]
+struct DetachedFrom {
+    workspace: WeakEntity<Workspace>,
+    pane: WeakEntity<Pane>,
 }
 
 impl EventEmitter<Event> for Workspace {}
@@ -1830,6 +1850,7 @@ impl Workspace {
             open_in_dev_container: false,
             _dev_container_task: None,
             deferred_save_items: Vec::new(),
+            detached_from: None,
         }
     }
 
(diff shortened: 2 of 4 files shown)


carlfarrington:
Really looking forward to this. If/when it does get merged and released, I suspect a lot of people are not going to realise the change until "Drag-out detach" is wired-in - it's the expected way to do this (all the editors you mentioned, and all/most web-browsers)

SpaskeISO:
Any update on this? Pull request was set to ready 3 weeks ago.
Or does zed team all work on single monitor setups and don't care about multi monitor support?

phonkd:
There have been i believe 3 PRs that stalled similar to this one...

guenthereder:
on Mac OS all tests passed, regarding "tests_pass Expected" in the conflict resolution

franciskafyi:
Hey @iam-liam, I caught up with Cameron on this and I wanted to share what I learned. We've been rewriting the core data structures for the workspace and we're still in the middle of another rewrite that Anthony is working on.

So we can't ship this until Anthony lands his changes and at that point, we'd have to rework your PR too 😫 Sorry about that!

iam-liam:
@franciskafyi sorry, I missed this. no problem.

crates/workspace/src/workspace.rs
```diff
@@ -5144,6 +5165,119 @@ impl Workspace {
...
+            &detached_pane,
+            &source_pane,
+            item_id,
+            destination_index,
+            true,
+            window,
+            cx,
+        );
+
+        window.remove_window();
```
smitbarmase, line 5278: I think we should only close the detached window when it becomes empty. Right now, if I open a few more tabs there and reattach one of them, the whole window closes along with all the other tabs.

crates/workspace/src/workspace.rs
```diff
@@ -5144,6 +5165,119 @@ impl Workspace {
...
+        cx: &mut Context<Self>,
+    ) {
+        let Some(detached_from) = self.detached_from.clone() else {
+            return;
+        };
+
+        let source_workspace = match detached_from.workspace.upgrade() {
+            Some(workspace) => workspace,
+            None => {
+                window.remove_window();
```
smitbarmase, line 5246: If the source window is already gone, reattaching closes the detached window too. I think this should just do nothing and leave the item where it is?

crates/editor/src/items.rs
```diff
@@ -1027,6 +1027,34 @@ impl Item for Editor {
...
+        let blink_manager = self.blink_manager.clone();
+        blink_manager.update(cx, |blink_manager, cx| {
+            if window.is_window_active() {
+                blink_manager.enable(cx);
+            } else {
+                blink_manager.disable(cx);
+            }
+        });
+        self._window_activation_subscription =
+            Some(cx.observe_window_activation(window, |editor, window, cx| {
```
smitbarmase, line 1044: One other case I noticed: after detaching an editor and closing the source window, changing cursor_shape no longer updates that editor.

Maybe moving the same `Entity<Editor>` between windows is the underlying problem. Could we use something similar to the existing editor split/clone path? It shares the underlying buffer while copying selections, folds, and scroll position into a fresh editor.

For in our case, that editor could be constructed using the new window so its window-bound subscriptions are registered there.

smitbarmase (changes requested):
Thanks for working on this. I tested it locally and found few edge cases. Let's try to fix these before I dive deep into reading more code.

iam-liam:
@smitbarmase Thanks for the review. I will be able to see to these in the coming weeks and will tag you in the changes

smitbarmase:
Hey, it's been about a month since the review, and the branch has also picked up conflicts, so I'm going to close this for now to keep the PR queue manageable.

If you get time to pick it back up with the suggested changes, feel free to open a new PR. Thanks again for the effort!

jmevel:
Oh no! Once again my hope of being able to use Zed on all my screens goes away.

@iam-liam hope you could find some time to work on the suggestions and open a new PR.

Thanks for your multiple attempts 🙏

PR #55419, opened 2026-05-01
Escape newline literals in syntax tree view

### LINKED ISSUE #54725: Newline literals break Syntax Tree rendering
### Reproduction steps

1. write a grammar with newline literals, ie.
```js
export default grammar({
  name: "note",
  rules: {
    source_file: ($) => repeat($.line),
    line: ($) => seq(
        /[^\n]*/,
        "\n", // <-- newline literal
    ),
  },
});
```
2. `dev: open syntax tree view` on any file (with multiple lines) using that grammar

Current:
Image

Idk if I would call it "expected", more observed when I use a regex (`/\n/`) to match the newlines instead (since zed doesn't print regex matched tokens). I suspect the ui should either handle the rows being taller, or render these as `"\n"` if it's preferred to preserve the consistent height for each row.
Image

Alternatively the tree view could simply not render literal tokens (ie. similar to how `tree-sitter parse` renders)
```console
$ tree-sitter parse sample.txt --grammar-path .
(source_file [0, 0] - [2, 0]
  (line [0, 0] - [1, 0])
  (line [1, 0] - [2, 0]))
```

Zed: v0.232.3+stable (Zed) 
Memory: N/A
Architecture: x86_64
GPU: N/A

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

theanuragg:
Fix Syntax Tree view rendering when Tree-sitter grammars use literal newline tokens (e.g. `"\n"`). Previously, these tokens were rendered as real newline characters in the UI, causing rows to stretch vertically and misalign.

This change ensures control-character literals are rendered as escaped text (e.g. `"\n"`) instead of inserting actual control characters into the label.

---

### Root cause

In `syntax_tree_view.rs`, `render_node` used `node.kind()` directly for anonymous (literal) tokens:

```rust
Label::new(format!("\"{}\"", node.kind())).color(Color::Created)
```

For literals like `"\n"`, `node.kind()` contained an actual newline character. When interpolated into the label string, this produced a real line break in the Syntax Tree view row, breaking row height and alignment.

---

### Fix

Introduce an `escape_node_kind()` helper that uses Rust’s `str::escape_debug()` to escape control characters before display, and use it for anonymous node kinds:

```rust
fn escape_node_kind(kind: &str) -> String {
    kind.escape_debug().collect()
}
```

`render_node` now uses:

```rust
Label::new(format!("\"{}\"", escape_node_kind(node.kind()))).color(Color::Created)
```

This converts control characters (`\n`, `\r`, `\t`, etc.) into their escaped representation for UI display while leaving named nodes unchanged.

---

### Behavior

- Literal `"\n"` tokens → displayed as the text `"\n"` (no actual newline in the row)
- Literal `"\t"` tokens → displayed as the text `"\t"`
- Literal `"\r"` tokens → displayed as the text `"\r"`
- Regex tokens like `/\n/` continue to behave as before (not displayed in this label)
- Syntax Tree rows maintain consistent height and alignment

---

### Self-review checklist

- [x] **Quality:** Change is minimal and localized to the Syntax Tree view. `escape_node_kind()` is a pure helper built on `str::escape_debug()`, which correctly handles all control characters.
- [x] **Security:** No security impact; this only affects how token kinds are rendered in the UI.
- [x] **Unsafe blocks:** No new `unsafe` code added.
- [x] **UI/UX consistency:** Prevents broken/misaligned rows caused by literal control characters, preserving the tree’s visual structure.
- [x] **Performance:** Token kinds are short strings; `escape_debug().collect()` allocates small, bounded strings. The performance impact is negligible relative to the rest of the view rendering.

Closes #54725

---

### Release notes
- Fixed Syntax Tree view rendering of literal control-character tokens (e.g. `"\n"`), which previously inserted real line breaks and caused row misalignment.


MrSubidubi:
@cla-bot check

PR #55432, opened 2026-05-01
git_ui: Add diff against custom branch action

ozacod:
## Summary

Adds a `git: diff against branch` action that opens a branch picker and shows the diff against the chosen branch instead of only the default. Addresses #43119. Action name is subject to change.

## Test Plan

- cargo fmt --check
- cargo test -p git_ui test_selection_mode_only_selects_existing_branches
- cargo test -p git_ui test_branch_diff_against_custom_branch
- cargo check -p git_ui
- Manually verified `git: diff against branch` opens the branch picker and shows a diff against the
  selected branch.

## Release Notes:

- Added `git: diff against branch` action to diff the working tree against a chosen branch.

### DIFF
--- a/crates/git_ui/src/branch_picker.rs
+++ b/crates/git_ui/src/branch_picker.rs
@@ -22,6 +22,9 @@ use util::ResultExt;
 use workspace::notifications::DetachAndPromptErr;
 use workspace::{ModalView, Workspace};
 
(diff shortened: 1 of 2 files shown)


siakc:
Current diff with the default is almost useless because for example we always diff features with `staging` branch. Thanks for the PR.

PR #55438, opened 2026-05-01
Make error notifications copyable

ozacod:
## Summary

Adds a copy button to simple message notifications marked as error or warning, so error text can be copied. Addresses #55430.

## Test Plan

- Run cargo fmt --check
- Run ./script/clippy
- Manually verified error notifications show a copy button.
- Manually verified info notifications, such as extension suggestions, do not show a copy button.

## Screenshot
image

## Release Notes:

- Added a copy button to error notifications.

### DIFF
--- a/crates/project_panel/src/undo.rs
+++ b/crates/project_panel/src/undo.rs
@@ -138,7 +138,7 @@ use gpui::{AppContext, AsyncApp, SharedString, Task, WeakEntity};
 use project::{ProjectPath, WorktreeId};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::{collections::VecDeque, sync::Arc};
-use ui::App;
+use ui::{App, Severity};
 use workspace::{
     Workspace,
     notifications::{NotificationId, simple_message_notification::MessageNotification},
@@ -550,7 +550,11 @@ impl Inner {
                     NotificationId::Named(SharedString::new_static("project_panel_undo"));
 
                 workspace.show_notification(notification_id, cx, move |cx| {
-                    cx.new(|cx| MessageNotification::new(error, cx).with_title(title))
+                    cx.new(|cx| {
+                        MessageNotification::new(error, cx)
+                            .severity(Severity::Error)
+                            .with_title(title)
+                    })
                 })
             })
             .ok();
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore (dismissed):
This is a visual change which does not fulfill https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#sending-changes , please fix this.

SomeoneToIgnore (commented):
I see 3 confusing things:

* all other components have the copy icon in the top-right corner, e.g. https://github.com/zed-industries/zed/blob/950161ff0ff34d313c795eef6c28b958cb487039/crates/workspace/src/notifications.rs#L362-L368
* `copyable` seems like an odd word, `text_to_copy` feels more natural
* very odd to see a limitation on severity and all extra API related to that — what's the problem to have a button in kind of a popover?

I'll defer to @danilo-leal to check on the design, just in case.

ozacod:
> I see 3 confusing things:
> 
> * all other components have the copy icon in the top-right corner, e.g. https://github.com/zed-industries/zed/blob/950161ff0ff34d313c795eef6c28b958cb487039/crates/workspace/src/notifications.rs#L362-L368
> * `copyable` seems like an odd word, `text_to_copy` feels more natural
> * very odd to see a limitation on severity and all extra API related to that — what's the problem to have a button in kind of a popover?
> 
> I'll defer to @danilo-leal to check on the design, just in case.


* LSP error/warning notifications does not have a close button in the top-right corner, while "MessageNotification" does. For copyable error text, placing the copy button next to close feels risky because the user could accidentally dismiss the notification while trying to copy it. That said, I am happy to defer to @danilo-leal on the best placement.
* Agree, I will rename it.
* Based on what I’ve seen in the code, notifications are used more like toast messages, so they can cover success and info messages as well. In this case, copy button on toast message could look odd. That said, I agree extra API adds some clutter to code. Since it is trade-off, I think either choice is reasonable, so I am happy to leave decision to team.

MrSubidubi:
Thanks for this!

I recently reworked how we handle error notifications on the workspace level and will close with that - I added a copy button for precisely that case.

However, that might not always be properly wired up - if you feel like it, I'd appreciate a follow-up PR of a fresh main that perhaps improves the copy functionality where it makes sense (if you deem that to be the case). I am not sure how that behaves all the time currently, hence would be curious if we can improve things there.

Will close this PR here due to that, still, thanks for taking the time to get this up and into shape!

PR #55442, opened 2026-05-01
helix: Add primary selection rotations and removal

### LINKED ISSUE #53965: Helix Cycling and removing selections
### Reproduction steps

1. Create a test block of text like:
```
How much would would a wouldchuck chuck
if a wouldchuck could chuck would?
```
2. Select both lines with `xx` or `2x`.
3. Type `s` to select, type "would" and enter.
4. Use `(` and `)` to cycle the primary selection and deselect the second "would" with `Alt-,` .
... and doesn't work.

Another example is to use `C` a few time to spawn multiple cursors across lines, now try and remove one; in Helix you can.

It's not just that `(`, `)`, and `alt-,` have no binding, the concept of cycling cursors, with/out selections, and removing them individually is missing (`n`/`N` is a different idea). This is a defining feature of Helix and the only way to manage multi-cursors in Zed in the above example is to reach for the mouse...

Zed Preview 0.232.2
MacOS 26.3

### Relevant Keymap

Tested with a default keymap file.

GoldStrikeArch:
Closes #53965

Adds Helix-style primary selection management:

- `)` rotates the primary selection forward
- `(` rotates the primary selection backward
- `alt-,` removes the primary selection

This also adds a Helix-only visual distinction for multiple local cursors by dimming secondary cursors. Without this, rotated primary cursors are functionally updated, but there is no reliable visual way to tell which cursor is currently primary, especially for collapsed multi-cursor selections. The dimming is scoped to Helix mode via editor state so regular editor and non-Helix Vim multi-cursor behavior remains unchanged.

This follows the existing internal model where the newest selection is treated as the active/primary selection. Happy to revisit the exact visual treatment if reviewers prefer a different primary-cursor indicator, such as highlighting the primary selection range differently.

Self-Review Checklist:

Release Notes:

- Added primary selection rotation and removal in Helix mode.


Amparose:
This is great! Top is Helix in Ghostty and bottom is Zed in this branch:

https://github.com/user-attachments/assets/48eb82da-d454-44c4-8968-169f9ee1705f

One thing I note looking back at the screencap is in Zed, the deselected words retain selection highlighting. This is with a different theme to make it more obvious:

Screenshot 2026-05-04 at 11 54 02@2x

Given that these targets cannot be re-added with cursors, I would prefer this hight-lighting went away like the pure Helix version; it's too distracting and adds more noise that varies theme-to-theme (Rose Pine Moon looks like a stain and is awful).

I did find the highlighting/dimming to be problematic enough with the One Light theme that I initially thought the feature was broken. I quickly tested other themes and discovered these make the active cursor hard enough to distinguish as to make the feature unusable:

- One Dark and Light
- Gruvbox Lights, Darks are on the edge of unusable
- Ayu is on the edge of unusable
- Snazzy is on the edge of unusable

I don't know enough to say if this is an issue of the themes or if there's a way to increase the contrast of an active cursor.

GoldStrikeArch:
@Amparose 

> I don't know enough to say if this is an issue of the themes or if there's a way to increase the contrast of an active cursor.

How much contrast do you want? We can also try to use the different highlighting instead of a "dimmed cursor" to highlight the primary/secondary selection (primary selection = active cursor). Btw does VIM mode differentiate the cursors in any way? how can you tell where is the primary/secondary cursor with Vim?

> Given that these targets cannot be re-added with cursors, I would prefer this hight-lighting went away like the pure Helix version

Yeah, makes sense. This can be adjusted, I will do it. But I think this is the "automatic highlights" from Zed 🤔 can you try to disable this setting and see if it will be the same? `zed://settings/selection_highlight`

Amparose:
> Btw does VIM mode differentiate the cursors in any way? how can you tell where is the primary/secondary cursor with Vim?

Interesting! So at first you tripped me up cause "vim doesn't have multiple cursors" but we are in Zed... so there's actually two UX paths to check in all this and I think we may have been referring to different things and not understanding the other: You can `⌘D` for Zed's native multi-cursor-select and then there is Helix's `%`, `s`, `↵` to multi-cursor based on selection and these have very different visual results. See video:

https://github.com/user-attachments/assets/ad93c6e6-9f3b-40f5-be7d-cc975e15735c

**The first part** shows `⌘D`, which is what I think you have been testing. It looks pretty legible, although I would suggest the letter under a non-selected cursor (letter "d" here) be foreground colour as it's quite hard to read currently. Dark themes look bad too:


Screenshot 2026-05-06 at 19 08 36@2x</br>

**The second part `@13sec`** shows pressing `%` to select all, `s` to search, `enter` creates the multiple cursors on selections. Hopefully, here you can see the contrast issue I was talking about: I can't tell where the active cursor is, and you will notice that removing a selection doesn't remove the "selection highlight" as it does with the `⌘D` example.

---

This is how multiple cursor in Vim mode looks. It's not overly helpful for us because it doesn't have the concept of an unselected cursor but contrast remains high and it is very legible:

Screenshot 2026-05-06 at 19 17 26@2x</br>

> How much contrast do you want? We can also try to use the different highlighting instead of a "dimmed cursor" to highlight the primary/secondary selection (primary selection = active cursor).

I think if the "Helix way" (`%`, `s`, `↵`) behaved like the Zed `⌘D` way, and had the letter under inactive cursors as "foreground colour" (so it would be unchanged black in the video above), then this would be resolved.

> I think this is the "automatic highlights" from Zed 🤔 can you try to disable this setting and see if it will be the same?

_Selection Highlight_ preference made no difference for me.

GoldStrikeArch:
@Amparose I see. So, on your "second part of the video" the search still has the "would" word and therefore the highlights are still applied. I guess we can clear the search after you hit the enter in Helix mode? Like the "s + enter" combo can do it I guess, what other combos you can use for a similar functionality, beside `s` + "word" + `enter`?

GoldStrikeArch:
@Amparose Alright, so I updated the branch again, now we close the search bar in Helix mode after "s + enter" and the dimming is a bit different now (we increase the lightness of a primary cursor selection and decrease the secondary ones, the cursors themselves are not dimmed)

So, in theory we can try to use a different color for a secondary selection (this is how helix is doing it ATM), but I am not sure if it would be better

Amparose:
@GoldStrikeArch 

> what other combos you can use for a similar functionality, beside `s` + "word" + `enter`?

This took a while to go through but here are all the examples I could think of that have subtle differences that are important, if annoying at times. This list is may not be exhaustive and I only tested in *Helix* mode not *Vim* or *off*.

**Using `⌘F`**

- `⌘F` pre-fills search with word under cursor
- First `enter` progresses active selection
- `n` types in and replaces the search term
- Exit by `escape` (bar has focus)
- Clicking doesn't exit

Or, instead of exiting:

- `tab` focuses buffer with *multi-selection, single-cursor mode*
- `n` now cycles selections

Selecting text before `⌘F` pre-fills search with selection

**Using `/`**

- `/` pre-fills search with last searched term
- First `enter` focuses buffer with *multi-selection, single-cursor mode*
- Subsequent `enter`s move cursor down
- `n` cycles selections
- Exit by `/` then `escape`
- Clicking doesn't exit

Selecting text before `/` pre-fills search with last search but ignores selection scope on `enter`

**Using `s`**

- `s` pre-fills search with last searched term
- `enter` dismisses search AND focuses buffer with *multi-selection, multi-cursor mode*
- `⌥(` / `⌥)` rotates active cursor
- `n` changes to *multi-selection, single-cursor mode* where rotate selections `⌥(` / `⌥)` does nothing (this is fine in my opinion just there's complexity here that may need documentation for users)
- Exit with `,` as per Helix's collapse cursors (same as ⌘D selections)
- Clicking also exits

Selecting text before `s` pre-fills search with last search and `enter` focuses buffer with *multi-selection, single-cursor mode* limited to selection.

**Using `*`**

- `*` pre-fills search with word under cursor AND focuses *multi-selection, single-cursor mode*
- `n` cycles selections (`enter` moves cursor down)
- Exit by focusing search input (`s` or `/`) then `escape`
- Clicking doesn't exit

Selecting text before `*` pre-fills search with selection

**Using `⇧⌘F`** (Find in Project)

- `⇧⌘F` pre-fills search with word under cursor and opens new tab
- `enter` is required to actually search and focuses *multi-selection, single-cursor mode* in multi-buffer
- `enter` moves cursor down
- `n` opens *Find in Results* bar and cycles selections.
  - However I had an instance where `n` did nothing and I could only cycle with `⌘G`; video on repro below as it might be a bug.
- Exit by closing tab

---

With all that exposed, closing the search bar on `s` + `enter` is really beneficial because:

1. It hints that you aren't in one of the other search modes and matches the `⌘D` behaviour set.
2. It's fast because I've never wanted the search bar to persist after doing multi-cursor actions and it has been endlessly annoying me.
3. It matches native Helix.

So keep that!

This will diverge from native Helix but I really think the active cursor highlight should be more prominent. Can it be the light yellow highlight in One Light (I assume that will translate to other themes)? Like this:

active highlight

Amparose:
This shows some undesirable behaviour I discovered with *Find in Project*:


https://github.com/user-attachments/assets/dde2d5b2-a7b3-4a33-8583-81eaad231be3



- Cycling terms with `n` brings up a secondary *Find in Results* bar.
- This bar is populated with the last search term used in a normal search NOT the current *Find in Project* term. Maybe this is useful for others but I found it disorientating until I understood wtf was happening. For the video I purposely searched "wouldchuck" before the *Find in Project* for "would".
- When closing the *Find in Project* tab:
  - If you used `n` to cycle and opened the second *Find in Results* bar, the original tab will have a search bar open. I found that unexpected and it requires two keystrokes to dismiss.
  - If you didn't trigger the *Find in Results* bar, the original tab has no search bar. My opinion is this should be default in both cases.
- The bug: If you enter something that doesn't exist into the *Find in Results* bar, it will break cycling selections the next time you *Find in Project*. The only way to "fix" this is perform a normal buffer search for something that exists and then *Find in Project* cycling will work again.

GoldStrikeArch:
@Amparose 

> This will diverge from native Helix but I really think the active cursor highlight should be more prominent. Can it be the light yellow highlight in One Light (I assume that will translate to other themes)? Like this:

Makes sense! I think we should reuse how `n` (and it's highlights of the "primary vs secondary) behaves right now in those scenarios. I will implement it like that, so it is consistent

> The bug: If you enter something that doesn't exist into the Find in Results bar, it will break cycling selections the next time you Find in Project. The only way to "fix" this is perform a normal buffer search for something that exists and then Find in Project cycling will work again.

I didn't get what exactly the bug is, I will try to check that part, it is not related to what we are trying to do though 😅

GoldStrikeArch:
@Amparose so, I updated the branch and now we are using "search color" for the secondary selections

Let me know if we need to implement/fix anything based on the original issue or we can stop here and fix other things in the separate PRs

Amparose:
@GoldStrikeArch I have checked the updates and to confirm, this is what it looks like for me:

Screenshot 2026-05-07 at 22 16 19@2x</br>

This is better in this particular theme but I think we still have issues. For example, in the Zedokai theme it's illegible: the highlight BGs are `#424043` and the active selection is `#414042`. Which is the active cursor selection here:

Screenshot 2026-05-07 at 22 29 22@2x</br>

I think this Helix multi-cursor-selection feature is novel for Zed and may not fit conventions perfectly. For instance, the `n` search behaviour with Zedokai colouring works because there's only one cursor that is bright and obvious where you are:

Screenshot 2026-05-07 at 22 42 38@2x</br>

> we should reuse how n (and it's highlights of the "primary vs secondary) behaves right now in those scenarios. I will implement it like that, so it is consistent

Does it though? The BG colours aren't consistent between `n` and Helix `s` as the active selection BG in the screenshot above is `#555357`, brighter and more differentiated than the Helix version's `#414042`. It doesn't match One Light's `n` version either:

Screenshot 2026-05-07 at 23 10 14@2x</br>

My suggestion is to make the active selection BG match `n` version, and try a "Helix convention" of the cursor adopting a theme's key colour (e.g. the colour of the "s" character in these screenshots). So here's a mockup of what I mean in Zedokai:

Artboard2</br>

> I didn't get what exactly the bug is, I will try to check that part, it is not related to what we are trying to do though 😅

Yep sorry, I came across it towards the end of my investigation, was tired, and wanted to capture the problem. I can open a separate issue and explain it better. It would probably be useful to re-record with an overlay showing key presses as it's tedious/confusing to explain that in text.

Amparose:
I am concerned that variations between themes might mean we don't consistently get a "key colour" that works. So another idea, that may be safer/easier, is to dim all the other cursors' BGs:

Artboard2dim

GoldStrikeArch:
@Amparose 

> So here's a mockup of what I mean in Zedokai

So, currently I was using this exact theme color (it's called `search_match_background`) for the "secondary selection" but you say it is not applied the same way as `n` in Zedokai theme? It does work like this on my theme (Andromeda):

Screenshot 2026-05-13 at 19 26 34


> So another idea, that may be safer/easier, is to dim all the other cursors' BGs:

Well, we already tried that here -> https://github.com/zed-industries/zed/pull/55442#issuecomment-4390524841 And it looked also not that great on some themes (by your own investigation)

IMO it is easier if the colors are different but we need to pick some good color 🤔 So, I think it would be easier to have the primary selection have a different color (like the blue in white theme you are using or yellowish/reddish on dark theme)

Amparose:
@GoldStrikeArch

> IMO it is easier if the colors are different but we need to pick some good color 🤔 So, I think it would be easier to have the **primary selection have a different color**

I agree. Going through [Zed themes](https://zed-themes.com/), I think this might work for the active cursor:

- `terminal.ansi.magenta` for active cursor BG
- `terminal.ansi.bright_white` for active cursor text

GoldStrikeArch:
@Amparose Alright, so now I updated the branch and the current version will use "players colors" (this is the collaboration mode feature, when you have multiple people on the same buffer it will highlight a cursors and selection of other participants differently, so I am reusing it here). In theory we should choose a different color in collaboration mode (so it would not "collide" with existing pallete) but I didn't implement that one (it is a niche case we can solve later of needed)

If that will also not work for you, then let's try those 2 colors you mentioned

> `terminal.ansi.magenta` for active cursor BG
> `terminal.ansi.bright_white` for active cursor text

Amparose:
@GoldStrikeArch I didn't think of multiplayer mode! That would be hell-confusing with all the same colours. I'm still new to Zed and haven't used that yet.

I just tried the updates with the problematic themes I posted here and a few others and it's AWESOME! Well done! So good:

Screenshot 2026-05-13 at 22 24 08@2x

Screenshot 2026-05-13 at 22 22 58@2x

The only issue I found in my testing was with the Ayu theme variants:

Screenshot 2026-05-13 at 22 28 53@2x

For me, I am really happy, but I think we should try and get it working well with the [themes Zed ships with](https://github.com/zed-industries/zed/tree/main/assets/themes). I can't see why it wouldn't work though as Ayu has [player color definitions](https://github.com/zed-industries/zed/blob/7d8fe38b56ca457835c22913208e546d3815162d/assets/themes/ayu/ayu.json#L142).

GoldStrikeArch:
@Amparose 

The thing with Ayu theme is that they use very similar colors as the first player color, so they have a problem in collaboration mode as well...IMO it is not worth it right now to have a more sophisticated solution (like to estimate how "close on pallete" the chosen colors are, and if they are close, try to use next player color and so on). As even those changes might be rejected (as we touch a lot of things in this PR :D)

Amparose:
@GoldStrikeArch 

> The thing with Ayu theme is that they use very similar colors as the first player color, so **they have a problem in collaboration mode as well**

Ok great! That means it's a problem for a different task/scope and doesn't need to block us here.

> As even those changes might be rejected (as we touch a lot of things in this PR :D)

Oh... so there's still a ways before this makes it into the Preview build? I'm pretty excited to daily drive this feature!

GoldStrikeArch:
@Amparose 
> Oh... so there's still a ways before this makes it into the Preview build

Well, I am not working at Zed, so I can't "call the shots" here 😅 I will try to bring this PR to maintainers and we will see their conclusion

taylorplewe:
Any update on this? Would be very nice to have. It's the last missing piece from Helix for me.

GoldStrikeArch:
> Any update on this? Would be very nice to have. It's the last missing piece from Helix for me.

I hope I can push this forward with the Zed team soon. I guess they are too busy working on Delta ATM

PR #55449, opened 2026-05-01
collab: Introduce `UserService`

maxdeviant:
This PR introduces a `UserService` trait to Collab.

This is a step towards moving Collab away from reading user information directly from the database.

We currently have two implementations for the trait:

- The `DatabaseUserService`, which leverages the existing query methods to talk to the database
- The `FakeUserService`, which will be used in tests

Once we're ready, we'll be able to replace the `DatabaseUserService` with a `CloudUserService` to fetch the users from Cloud.

Release Notes:

- N/A

### DIFF
--- a/crates/collab/migrations.sqlite/20221109000000_test_schema.sql
+++ b/crates/collab/migrations.sqlite/20221109000000_test_schema.sql
@@ -21,8 +21,8 @@ CREATE UNIQUE INDEX "index_users_on_github_user_id" ON "users" ("github_user_id"
 
 CREATE TABLE "contacts" (
     "id" INTEGER PRIMARY KEY AUTOINCREMENT,
-    "user_id_a" INTEGER REFERENCES users (id) NOT NULL,
-    "user_id_b" INTEGER REFERENCES users (id) NOT NULL,
+    "user_id_a" INTEGER NOT NULL,
+    "user_id_b" INTEGER NOT NULL,
     "a_to_b" BOOLEAN NOT NULL,
     "should_notify" BOOLEAN NOT NULL,
     "accepted" BOOLEAN NOT NULL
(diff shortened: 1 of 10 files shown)


maxdeviant:
I suspect the tests are going to be failing for this, as the `FakeUserService` hasn't been implemented yet. Will get them passing when I'm not on airplane WiFi.

crates/collab/migrations.sqlite/20221109000000_test_schema.sql
```diff
@@ -21,8 +21,8 @@ CREATE UNIQUE INDEX "index_users_on_github_user_id" ON "users" ("github_user_id"
 
 CREATE TABLE "contacts" (
     "id" INTEGER PRIMARY KEY AUTOINCREMENT,
-    "user_id_a" INTEGER REFERENCES users (id) NOT NULL,
```
maxdeviant, line 24: We've dropped all of the foreign key constraints on the `users` table from the SQLite schema, as we will no longer be creating records in that table in the tests.

We use the in-memory storage on the `FakeUserService` to create the users.

In production, the `users` records will be there (just managed by Cloud, instead).

crates/collab/src/db/queries/channels.rs
```diff
@@ -684,6 +684,26 @@ impl Database {
...
 
+    /// Returns the channel memberships for the users with the specified IDs.
+    #[cfg(feature = "test-support")]
+    pub async fn get_channel_memberships_for_user_ids(
+        &self,
+        channel: &Channel,
+        ids: Vec<UserId>,
+    ) -> Result<Vec<channel_member::Model>> {
+        self.transaction(|tx| async {
+            let tx = tx;
```
tomhoule, line 695: Is this intentional?

tomhoule (approved):
I wonder if the fake user service couldn't insert a row in the database in addition to its in memory representation so we still catch the foreign key constraint errors in development. We're not going to change things very much or very fast in collab so maybe not a concern.

crates/collab/src/db/queries/channels.rs
maxdeviant: Yea, without it there's a borrow checker error. I don't fully understand it (but I cribbed this from another method).

maxdeviant:
> I wonder if the fake user service couldn't insert a row in the database in addition to its in memory representation so we still catch the foreign key constraint errors in development. We're not going to change things very much or very fast in collab so maybe not a concern.

I suppose that could be an option 🤔 

Right now I'm trying to rip out as much knowledge of the `users` table as possible from Collab.

crates/collab/src/db/queries/channels.rs
tomhoule: Ah yes, moving the argument into the async block, I missed that 👍

PR #55492, opened 2026-05-02
vim: Add Helix jump-to-word support to Vim mode

GoldStrikeArch:
Closes #55481

Adds Vim-mode access to the existing Helix jump-to-word overlay via `g z`. We use `g z` because it is currently unassigned in Vim mode, while `g w` is already used for rewrap.

Most of the implementation lives in `helix.rs` because the existing jump overlay, label generation, and Helix/Vim modal behavior are currently intertwined there. This keeps the change small and reuses the existing navigation overlay logic instead of doing a broader refactor.

In Vim normal mode, jump labels behave like a cursor motion: selecting a label moves the cursor to the start of the target word without selecting it. In Vim visual mode, jump labels extend the selection like a Vim word-start motion, preserving Vim’s inclusive visual-selection behavior.

Self-Review Checklist:

Release Notes:

- Added Vim-mode jump-to-word navigation on `g z`.


ksweetie:
I tested this and it works great, thank you!

> In Vim visual mode, the behavior continues to extend the selection like Helix select mode.

I find this behavior surprising, but that might just be me.

GoldStrikeArch:
@ksweetie 

> I find this behavior surprising, but that might just be me.

What you behavior you would expect?

ksweetie:
Caveat that I don't consider myself to be a vim expert. But I generally expect a motion to move the cursor to the same place whether in normal or visual mode. I'm playing around with various motions to try to find any that are different in the two modes. So far, I've only found `$` to be slightly different.

- `$` (cursor ends up on the last character of the line)
- `v$` (cursor ends up one character past the last character, i.e. it includes the line break)

Otherwise, everything I've tried moves the cursor to the same position in both modes.

GoldStrikeArch:
@ksweetie

> a motion to move the cursor to the same place whether in normal or visual mode

I see, that makes sense. But from the practical perspective -> if you would put it to the same position as a "normal mode motion" (which is to the beginning of the word, like to the first character), is it actually what you want? Like usually you do it to delete text, etc, you would select the first letter of a word instead of the whole world (which does seem like a bad thing to me, but again I don't use Vim mode 😅 )

After some thinking:
- commands like `v w` already do that (select the first letter after a movement). So, I guess Vim users are fine with that and this is indeed a desired behavior? 🤔 

So, I updated the branch and it should behave similar to Normal mode now

ksweetie:
Ha, yeah those are interesting questions. I can definitely see the argument for wanting to select the word by default.

> commands like `v w` already do that (select the first letter after a movement). So, I guess Vim users are fine with that and this is indeed a desired behavior?

I guess that's something I wouldn't do in practice. If I want to select the current word, I would do `v e` or `v E`. If I want to select the entire next word as well, then I would do something like `v e e` or `v 2 e`.

The only other relevant data point I can think of is that when you search with `/`, the cursor will land on the start of the search term in both normal and visual modes.

All that said, I think Conrad is the maintainer of vim in Zed. He'll have a much more informed opinion than me. Thanks again for adding this, I'm really excited for it!

ConradIrwin:
@GoldStrikeArch Thanks for taking a pass at this!

We already have sneak support, can we merge this and that?

What other vim plugins exist, and what shortcuts do they use (where did you get the idea for `gz` from?)

GoldStrikeArch:
@ConradIrwin 

> We already have sneak support, can we merge this and that?

I am not sure that I am following, this one and sneak are a different things

> What other vim plugins exist, and what shortcuts do they use (where did you get the idea for gz from?)

I checked out and easymotions, nvim.flash, nvim.leap and they use <Leader><Leader>, `s`, `S`, `f`, `F`, `T`, so they kinda override the default motion with their version. I took `g z` just because it was not taken. It can be anything really, something like `space + q`? or some other "easy combo". Helix does use `g w` for it but it is taken in Vim mode

niekdomi:
Just my two cents:

The afaik major jump plugins are
- [vim-sneak](https://github.com/justinmk/vim-sneak) (already implemented, but isn't really a jump plugin imo, just a "better" `f`)
- [leap.nvim](https://codeberg.org/andyg/leap.nvim)
- [flash.nvim](https://github.com/folke/flash.nvim)


## Differences

### This PR
shows two labels at start of word. is a "vim port" of the helix jump (or whatever its called)

### leap
You type `s` (or whatever key to start the search) and type two characters e.g., `ti` if you want to jump to `t` in `action`.  After typing the characters, all matches of `ti` will receive a label. So if you want to jump to `t` in `action` you likely type something like `stib`. (This is at least how I understand a plugin, correct me if I'm wrong)

### flash
You normally type `s` to go into search mode and start typing the word/position you want to jump to. All matches will get a label

https://github.com/user-attachments/assets/40e17c8a-118e-4b01-a88d-a8ee087f7d99

> they use , s, S, f, F, T, so they kinda override the default motion with their version

`s`/`S` aren't really required, since the same can be achieved with `cl` or  `cc`. though ideally there is no default mapping ,similar to how its done for `vim-sneak`.

From the docs:

> The [Sneak motion](https://github.com/justinmk/vim-sneak) feature allows for quick navigation to any two-character sequence in your text. You can enable it by adding the following keybindings to your keymap. By default, the s key is mapped to vim::Substitute. Adding these bindings will override that behavior, so ensure this change aligns with your workflow preferences.

GoldStrikeArch:
@niekdomi 

I see, so those plugins have a different mechanics and purpose. The "amp jump" is a different meaning -> you get "prepared" labels, so you can jump to any word within the current view. 

And this particular PR is just for Vim mode users to "get amp jump for free" because some users already tried to use it and it "kinda worked" but with Helix semantics (like the selection first) which can be somewhat easily adjusted.

niekdomi:
> this particular PR is just for Vim mode users to "get amp jump for free"

This is a good stopgap, but I think the real win would be enabling proper Flash/Leap navigation (maybe via a plugin). I'd love to see us move toward that rather than just settling for another jump-style alternative.

GoldStrikeArch:
> the real win would be enabling proper Flash/Leap navigation (maybe via a plugin)

IMO it is much easier to do it directly here in the `vim.rs` rather than re-iterate on extensions API and expose the "host editor state", so you can scan the ranges, set/unset `NavigationOverlay` and so on. All of that can't be done right now from the extension and I guess maintainers would not want to expose all of that just for the Flash plugin (maybe I am wrong here)

k-soomro:
Thank you all for working on this. I switched the preview version just to get this feature and eagerly looking forward to this being merged. 

FWIW I've used both flash and leap and I am used to jumping to the start of the word and have it unselected.

GoldStrikeArch:
> FWIW I've used both flash and leap and I am used to jumping to the start of the word and have it unselected.

@k-soomro 
What would be a good keymap for this in Vim? Right now we are using `g z` just because it is not taken, but maybe there are others common keybings people are using for it?

k-soomro:
> > FWIW I've used both flash and leap and I am used to jumping to the start of the word and have it unselected.
> 
> @k-soomro What would be a good keymap for this in Vim? Right now we are using `g z` just because it is not taken, but maybe there are others common keybings people are using for it?

Being a user of leap and flash, I'm used to just using `s` and `S`. But I've currently mapped the Amp Jump action to `g w` in vim mode as well. To be honest I'd say it doesn't matter too much since people can reconfigure it to suit their needs anyway.

arcqiufeng:
Can't wait for this feature. Looking forward to its early merge and release.

ConradIrwin:
@GoldStrikeArch I'm happy to merge this, but:

* Let's not take the `gz` binding by default (it's not used by any plugin, and I really want to keep some bindings free for new features)
* Let's document how to set it up alongside where we document the sneak code.

Sorry for the slow response, and thanks for making that work.

GoldStrikeArch:
@ConradIrwin 

> Let's not take the `gz` binding by default (it's not used by any plugin, and I really want to keep some bindings free for new features)

Do you have any good default binding in mind?

k-soomro:
> @ConradIrwin 
> 
> 
> 
> > Let's not take the `gz` binding by default (it's not used by any plugin, and I really want to keep some bindings free for new features)
> 
> 
> 
> Do you have any good default binding in mind?

I think he's saying that there shouldn't be any default keybinding. Just have instructions for people to set it up themselves.

GoldStrikeArch:
@k-soomro 

> I think he's saying that there shouldn't be any default keybinding. Just have instructions for people to set it up themselves.

Makes sense, thanks! I will then adjust it like that

GoldStrikeArch:
@ConradIrwin Alright, I removed the default keybinding and added the docs as you requested

ConradIrwin:
Yeah; it sounds like people mostly bind this to `<leader><leader>` which would be space space or , , - but I don't want to use those to avoid colliding with other leader shortcuts people add. or `s`, which is already something else

k-soomro:
> Yeah; it sounds like people mostly bind this to `<leader><leader>` which would be space space or , , - but I don't want to use those to avoid colliding with other leader shortcuts people add. or `s`, which is already something else

Is there a specific way to bind <leader> or do you just bind space?

arcqiufeng:
The currently released version is mostly sufficient. However, for Vim users, there is definitely room for improvement.


I’d like to mention a very useful feature of flash.nvim: when displaying jump labels, the syntax highlighting of the rest of the content is completely disabled and turns grayscale, making the jump labels stand out very clearly. Currently, the color of the jump labels blends in with the theme's colors. If the chosen theme is not appropriate, the labels become completely indistinguishable.






---Original---
From: "Mikhail ***@***.***&gt;
Date: Wed, May 13, 2026 05:21 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [zed-industries/zed] vim: Add Helix jump-to-word support to Vim mode (PR #55492)


GoldStrikeArch left a comment (zed-industries/zed#55492)
 
@ConradIrwin Alright, I removed the default keybinding and added the docs as you requested
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

probably-neb:
@zed-zippy approve

trkoch:
For people coming here from the release notes, note that the default binding `g z` was removed. You need to define your own binding.

```json
  {
    "context": "vim_mode == normal || vim_mode == visual",
    "bindings": {
      "s": "vim::HelixJumpToWord",
    },
  },

```

GoldStrikeArch:
Yes, those release notes were not changed according to the latest changes

tannerkovach:
I love that this minor change was added so you aren't auto visually selecting the whole word like the Helix version. However with that said, we are SO CLOSE to feature parity with popular vim/nvim jump plugins, the main thing missing is being able to jump to non-words. Very often I like to jump to specific brackets or other symbols, and this just doesn't support that. Surely it wouldn't be too difficult to expand upon this?

GoldStrikeArch:
@tannerkovach 

> the main thing missing is being able to jump to non-words. Very often I like to jump to specific brackets or other symbols, and this just doesn't support that. Surely it wouldn't be too difficult to expand upon this?

So, basically you want to jump at any "text cell" (whether it is a letter/symbol, etc) in a view? The problem with brackets is that they are "1 character" length and if you add 2 letter symbols to it then it might overlap with the next word, etc

Like if you have something like this:

```js
if (smth) {return}
```
To jump into `{` bracket will require to have only 1 letter label 🤔 which is like changing how we calculate labels 

But isn't it "easy enough" to jump into specific word first (like `return`) in our case and then just use `[{` (native Vim)?

PR #55506, opened 2026-05-02
git_graph: Improve navigation in Vim mode

aprogramq:
Added Ctrl-D/Ctrl-U for Vim navigation in the Git graph

https://github.com/user-attachments/assets/8d3ad67c-829a-4a80-9508-80d48cf0decf

Self-Review Checklist:

Closes #55485 

Release Notes:
- Improved Vim navigation in the git graph with Ctrl-U and Ctrl-D half-page scrolling.

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -11,11 +11,11 @@ use git::{
 };
 use git_ui::{commit_tooltip::CommitAvatar, commit_view::CommitView, git_status_icon};
 use gpui::{
-    Anchor, AnyElement, App, Bounds, ClickEvent, ClipboardItem, DefiniteLength, DragMoveEvent,
-    ElementId, Empty, Entity, EventEmitter, FocusHandle, Focusable, Hsla, PathBuilder, Pixels,
-    Point, ScrollStrategy, ScrollWheelEvent, SharedString, Subscription, Task, TextStyleRefinement,
-    UniformListScrollHandle, WeakEntity, Window, actions, anchored, deferred, point, prelude::*,
-    px, uniform_list,
+    Action, Anchor, AnyElement, App, Bounds, ClickEvent, ClipboardItem, DefiniteLength,
+    DragMoveEvent, ElementId, Empty, Entity, EventEmitter, FocusHandle, Focusable, Hsla,
+    PathBuilder, Pixels, Point, ScrollStrategy, ScrollWheelEvent, SharedString, Subscription, Task,
+    TextStyleRefinement, UniformListScrollHandle, WeakEntity, Window, actions, anchored, deferred,
+    point, prelude::*, px, uniform_list,
 };
 use language::line_diff;
 use menu::{Cancel, SelectFirst, SelectLast, SelectNext, SelectPrevious};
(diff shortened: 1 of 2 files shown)


crates/git_graph/src/git_graph.rs
```diff
@@ -1186,6 +1191,7 @@ impl GitGraph {
             column_widths,
             selected_entry_idx: None,
             hovered_entry_idx: None,
+            rendered_entries_len: 0,
```
Anthony-Eid: We don't need this field, look at how the git graph canvas rendering can get the visible row/entries count.

crates/git_graph/src/git_graph.rs
```diff
@@ -1526,6 +1534,18 @@ impl GitGraph {
...
+            window.dispatch_action(SelectPrevious.boxed_clone(), cx);
+        }
+    }
+
+    fn scroll_down(&mut self, _: &ScrollDown, window: &mut Window, cx: &mut Context<Self>) {
+        for _ in 0..self.rendered_entries_len / 2 {
+            window.dispatch_action(SelectNext.boxed_clone(), cx);
+        }
+    }
+
```
Anthony-Eid, line 1576: This would be cleaner if you called something like 

```rust
        let scroll_strategy = ScrollStrategy::Bottom; // Top when you're scrolling up

        self.table_interaction_state.update(cx, |state, cx| {
            state.scroll_handle.scroll_to_item(to_scroll_index, scroll_strategy);
            cx.notify();
        });
```

Anthony-Eid (changes requested):
I commented some of the changes I would like to see

crates/git_graph/src/git_graph.rs
aprogramq: Thanks for the review. What do you think of this option?

```rust   
fn scroll_up(&mut self, _: &ScrollUp, window: &mut Window, cx: &mut Context<Self>) {
        let step = ( self.visible_row_count(window, cx) / 2 ).max(1);
        let target_idx = self.selected_entry_idx.unwrap_or(0).saturating_sub(step);
        
        //"select_entry" is more appropriate, as it is the handler for selecting a commit from the history
        self.select_entry(target_idx, ScrollStrategy::Nearest, cx);
    }
```
and the addition of the `visible_row_count` helper function
```rust
    fn visible_row_count(&self, window: &Window, cx: &App) -> usize {
        let row_height = Self::row_height(window, cx);
        let viewport_height = self
            .table_interaction_state
            .read(cx)
            .scroll_handle
            .0
            .borrow()
            .last_item_size
            .map_or(window.viewport_size().height, |size| size.item.height);

        (viewport_height / row_height).floor() as usize
    }

```

Anthony-Eid: That looks good! 

`viewport_height / row_height).ceil` is more correct to me, because half a row could still be considered a visible row

aprogramq:
@cla-bot check

Anthony-Eid:
@aprogramq Thank you for contributing, I'm going to turn on auto merge.

I pushed a commit that fixed the clippy error, added bounds checking to `scroll_down`, and made the `render_graph_canvas` call `visible_rows_count` function you added to unify where that calculation is done.

Anthony-Eid (approved):

aprogramq:
@Anthony-Eid Thank you so much for your review. I've never had such a pleasant experience contributing before.

PR #55515, opened 2026-05-02
vim: Add C preprocessor check in matching function

### LINKED ISSUE #24820: vim: % doesn't match #if #endif and /* */
### Summary

Vim mode implements the key % to match pairs of parenthesis, brackets, etc. Compared to vim, it misses #if/#endif preprocessor directives and multiline comments /* */.

Zed: v0.172.11 (Zed)
OS: macOS 15.3.0
Memory: 32 GiB
Architecture: aarch64

tanaxer01:
Closes #24820

This PR fixes the bug specified in issue https://github.com/zed-industries/zed/issues/24820, now the matching function checks if the cursor is above a comment or a directive before defaulting to a bracket range as neovim does.

It also fixes fixes the `line_end` calculations so that when `%` is pressed inside a bracket range 

https://github.com/user-attachments/assets/f59daa6f-9769-45e8-bb8c-2d533470b59d

Release Notes:

- `fn matching()` checks for `preprocessor directives` or `comments` before defaulting to any bracket range.
- In `fn matching()`line_end calculations avoid expanding a blank current line into start..EOF.

### DIFF
--- a/crates/vim/src/motion.rs
+++ b/crates/vim/src/motion.rs
@@ -2452,7 +2452,7 @@ fn find_matching_bracket_text_based(
         .find_map(|(ch, char_offset)| get_bracket_pair(ch).map(|info| (info, char_offset)));
 
     if bracket_info.is_none() {
-        return find_matching_c_preprocessor_directive(map, line_range);
+        return find_matching_c_preprocessor_directive(map, line_range, offset);
     }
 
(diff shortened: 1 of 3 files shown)


ConradIrwin (approved):
Thanks!

tanaxer01:
@ConradIrwin  ups, I messed up the approve

PR #55540, opened 2026-05-03
ollama: Fix thinking being also sent as content

### LINKED ISSUE #55537: Ollama: thinking is also sent as content
### Reproduction steps

1. Start a recent Zed version
2. Use the ollama endpoint with an AI that support thinking (in my case gemma4)
3. log or intercept the HTTP message
4. notice that what should be in the thinking field is also in the content field

Expected behavior: thinking stay in the content section to ollama request

Commit: 6b28db5ef5ea7e90ba6d267b6f1be3e4de00a46a

Zed: v1.0.0+stable (Zed)
OS: Linux Wayland nixos 26.05
Memory: 30.5 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 3090 || NVIDIA || 595.71.05

### Attach Zed log file

Nothing relevant in log (I printed stuff directly to eprintln!)

### Relevant Zed settings

_No response_

- Provider: local ollama
- model: gemma4:26b
- agent panel

None

marius851000:
"msg.string_contents();" return more than just content. It also return tool result (which need special handling, and should be emitted from the tool role) and thinking (which need already implemented special handling).

This resulted in thinking being sent as content, which apparently ollama’s gemma4 parser didn’t handled well, and caused a number of issue that manifested by 1. outputting end of though token where unappropriate and 2. repeating things endlessly

Self-Review Checklist:

- [ ] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) (no UI/UX impact)
- [x] Performance impact has been considered and is acceptable (a quick review, without benchmark, show that this should be at least as fast as the previous code. This is called only exceptionally anyway.)

Closes #55537

Release Notes:

- Fixed "thinking" text being badly formatted when sent to Ollama

### DIFF
--- a/crates/language_models/src/provider/ollama.rs
+++ b/crates/language_models/src/provider/ollama.rs
@@ -381,11 +381,14 @@ impl OllamaLanguageModel {
                     }
                 }
                 Role::Assistant => {
-                    let content = msg.string_contents();
+                    let mut text_content = String::new();
                     let mut thinking = None;
                     let mut tool_calls = Vec::new();
                     for content in msg.content.into_iter() {
                         match content {
+                            MessageContent::Text(text) => {
+                                text_content.push_str(&text);
+                            }
                             MessageContent::Thinking { text, .. } if !text.is_empty() => {
                                 thinking = Some(text)
                             }
@@ -402,7 +405,7 @@ impl OllamaLanguageModel {
                         }
                     }
                     messages.push(ChatMessage::Assistant {
-                        content,
+                        content: text_content,
                         tool_calls: Some(tool_calls),
                         images: if images.is_empty() {
                             None


marius851000:
In the end, I still have issue with gemma4. Less systematic, but I suspect there are still issue somewhere between Zed and Ollama. Changed the release notes in accordance. (this patch is benificial nonetheless)

benbrandt (approved):
good catch, thanks! 
Should we also be appending thinking text?

marius851000:
Appending thinking text? As in, merging ChatMessage::Assistant to contain both the thinking part and the generated part? Maybe. But looking at the generated text feed to the model by ollama, I notice nothing wrong beside what is fixed here.

Issue #47928, opened 2026-01-29
claude code integration swallows commands


dignifiedquire:
### Reproduction steps

After updating sending a command while claude code is executing is not shown in the interface, it is just swallowed, with no sign of it being acknowledged anywhere in the system. 
Before it would interrupt, claude code and properly interact with it, this only happens now when the task it is working on is finished

### Current vs. Expected behavior

see above

### Zed version and system specs

Zed: 0.221.4+stable.137.128478628237a6d03a6890c810d21b55d3250a70
Macos

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

nik9000:
I'm continuing to interact with this chat a little, for science. I can confirm that it is *rendering* one late. I asked it to:
```
   ME: optimize `appendVInt` by making an `appendVIntToTail` and `appendVIntPaged` - kick in the `ToTail` implementation if there are 5 bytes in the tail. call `appendToTail(byte)` there.
SONET: ... thinking ...
SONET: <replied to the prompt before the one above>
```

Then, a few second later, it did the task I asked it to do.

mchisolm0:
**Reproduction steps**
Haven't isolated the repro but so far it seems:

1. Use Claude ACP integration for a while
2. Queue a message in the middle of a task, probably close to a conversation compaction point
3. Press the button to send the message immediately
NOTE: I've encountered this issue on both Windows and macOS.

*Edit: it seems to happen during long-running tools are being executed and possibly cannot be interrupted? Per my note below, if I wait a very long time for the command to finish things eventually re-sync.

**Current behavior**: Message gets dropped (i.e. not displayed in the feed), Zed shows a spinner with "Loading added context..."

<details><summary>Screenshot of loading message</summary><img width="412" height="160" alt="Image" src="https://github.com/user-attachments/assets/d0756579-9924-45c6-bdc2-2ce6793e5cfc" /></details>

Trying to start a new conversation then go back to this old conversation does not clear things and does not show the dropped message. Usually I have to completely restart Zed.

If I wait a VERY, VERY long time it might eventually add. Perhaps a tool gets stuck and can't be interrupted or something? Not sure.

[...]

PR #55562, opened 2026-05-03
acp_thread: Clear running_turn when prompt task drops tx

brunorodmoreira:
`AcpThread::status` is purely `running_turn.is_some()`. The cleanup that takes `running_turn` sat below the early-return guard that fires when the prompt response oneshot resolves to `Err(Cancelled)` (the inner `send_task` was dropped before `tx.send`). Any code path that drops the in-flight `send_task` therefore left the panel stuck in `Generating`. Reordered so cleanup runs before the dropped-tx guard; the same-turn invariant is preserved.

Related to #47928 (partial — that issue also has an upstream `claude-agent-acp` component this PR does not address).

Self-Review Checklist:

Release Notes:

- Fixed agent panel staying in a generating state when the underlying prompt task was cancelled before completing

### DIFF
--- a/crates/acp_thread/src/acp_thread.rs
+++ b/crates/acp_thread/src/acp_thread.rs
@@ -2288,10 +2288,6 @@ impl AcpThread {
                     this.project
                         .update(cx, |project, cx| project.set_agent_location(None, cx));
                 }
-                let Ok(response) = response else {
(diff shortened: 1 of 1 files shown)


benbrandt (approved):
thanks!

PR #55609, opened 2026-05-04
ep: Move scores aggegation to edit_prediction_metrics

osyvokon:
This way, it can be shared with Python bindings.

Release Notes:

- N/A


probably-neb (approved):

probably-neb:
@zed-industries/approved

PR #55621, opened 2026-05-04
language_tools: Open server logs instead of server trace on View Logs

Veykril:
Release Notes:


SomeoneToIgnore (approved):
Arguably, server logs contain less information usually (as certain language server are not configured to emit a thing), so I'd try opening the trace logs as a fallback/instead of logs entirely.

But this is also one way to remove the naming discrepancy.

Veykril:
The rationale for this change from me at least is the fact that when r-a fails to load, we prompt the user to open logs. The traces do not contain the relevant info though, and additionally I'd expect a `View Logs` button to open the logs, not the traces

PR #55649, opened 2026-05-04
ci: Create releases with the Zed Zippy identity

MrSubidubi:
Just a small QoL, the change here will make it so that under https://github.com/zed-industries/zed/releases/tag/v1.0.1 the releases will show as created by zed-zippy and not github-actions.

Release Notes:

- N/A


ConradIrwin (approved):
Thanks!

PR #55660, opened 2026-05-04
extension_builder: Use `cfg_select` for WASI SDK asset name

MrSubidubi:
Non-functional change that just makes use of our most recent version bump and helps with readability IMO. Also helps with showing that we are in fact missing dev extension installation support for Windows ARM at the moment.

Release Notes:

- N/A


osiewicz (approved):
On one hand it bumps our MSRV, but on the other it's not like we're tracking that anyway, so eh.

PR #51165, opened 2026-03-10
agent: Use streaming diff when agent rewrites the entire file

bennetbo:
TODO:
- [ ] Check change in multi-buffer
- [ ] Add tests for compute_hunks

We now use a streaming line based-diff when the model decides to rewrite the entire file.

Main:

https://github.com/user-attachments/assets/fe8723c1-034f-4158-8d5f-d3d6e1abda36

This branch:

https://github.com/user-attachments/assets/55708146-32e0-42f1-8ca8-675d46cfa4c6

Release Notes:

- N/A

PR #55663, opened 2026-05-04
agent: Refactor edit file tool state handling

bennetbo:
Ports some changes introduced in #51165 out to make merge conflicts easier to handle.

Splits the `Pipeline` into two separate types for mode edit/write so we don't need to maintain that invariant inside the pipeline/in the parser

Also moves the parser to be a submodule of `edit_file_tool`

Release Notes:

- N/A


benbrandt:
@zed-industries/approved

PR #39609, opened 2025-10-06
Add line endings indicator in status bar

kitt-cat:
Closes #5294

This PR adds a line ending indicator to the status bar, hidden by default as discussed in https://github.com/zed-industries/zed/issues/5294.

### Changes

- 8b063a22d8700bed9c93989b9e0f6a064b2e86cf add the indicator and `status_bar.line_endings_button` setting.

- ~~9926237b709dd4e25ce58d558fd385d63b405f3b changes `status_bar.line_endings_button` from a boolean to an enum:~~
  <details> <summary> show details </summary>

   - `always`     Always show line endings indicator.
   - `non_native` Indicate when line endings do not match the current platform.
   - `lf_only`    Indicate when using unix-style (LF) line endings only.
   - `crlf_only`  Indicate when using windows-style (CRLF) line endings only.
   - `never`      Do not show line endings indicator.
   
   I know this many options might be overdoing it, but I was torn between the pleasant default of `non_native` and the simplicity of `lf_only` / `crlf_only`.

   My thinking was if one is developing on a project which exclusively uses one line-ending style or the other, it would be nice to be able to configure no-indicator-in-the-happy-case behavior regardless of the platform zed is running on. But I'm not really familiar with any projects that use exclusively CRLF line endings in practice. Is this a scenario worth supporting or just something I dreamed up?

   </details>

- 01174191e4cf337069e7a31b0f0432ae94c52515 rename the action context for `line ending: Toggle` -> `line ending selector: Toggle`.
   When running the action in the command palette with the old name I felt surprised to be greeted with an additional menu, with the new name it feels more predictable (plus now it matches `language_selector::Toggle`!)

### Future work

Hidden status bar items still get padding, creating inconsistent spacing (and it kind of stands out where I placed the line-endings button):

the gap after the indicator is larger than for other buttons

I started a new follow-up PR to address that: https://github.com/zed-industries/zed/pull/39992

Release Notes:

- Added line ending indicator to the status bar (disabled by default; enabled by  setting `status_bar.line_endings_button` to `true`)

PR #55707, opened 2026-05-04
Add show line endings button to Settings UI

dunkmann00:
This is following up from #39609 to add the line endings button setting into settings ui.

Release Notes:

- Added the line endings button setting to settings ui.


ChristopherBiscardi (approved):
awesome, thanks!

PR #55715, opened 2026-05-05
Improve grouping of worktrees by repo in recent projects

maxbrunsfeld:
* Perform grouping even for repositories that have no main worktree
* Enable grouping for remote projects
* Delete entire project groups when deleting via the recent project picker

Release Notes:

- Fixed a bug where each linked worktree appeared as its own entry in recent projects for repositories without main worktrees
- Fixed a bug where deleting projects from the recent projects sometimes appeared to have no effect.


benbrandt (approved):
Nice. seems like a nice simplification in addition to the fix 👍🏻

PR #55720, opened 2026-05-05
Add an initial context menu to the git graph entrees

JosephTLyons:
This PR adds an initial context menu to the git graph entries. There are a bunch of commit-specific actions we will likely want to add here over time (e.g. create a branch at this commit, revert, etc.), but for this PR, it only exposes the actions that were already available from the commit details panel: `Copy SHA` and `View Commit`. We will also need a context menu to land a future want of mine: custom git command support in the git graph.

This was a bit trickier than a normal context menu addition because the git graph uses the selected entry to drive the commit details panel. If right-clicking a row went through the normal selection path, it would also pop open the commit details panel if it was closed, or change the commit currently being shown if it was already open. I don’t think right-clicking to open a context menu should do either of those things. The context menu target and the commit details panel should be independent of one another.

To support that, this PR introduces `GitGraphContextMenu`. Most of this state was already present as a tuple for rendering a context menu, but it wasn’t wired up to graph rows. I pulled that state into its own type and added an `entry_idx` field to track which row the context menu was opened on. This lets the row highlight while the menu is open without changing the selected commit or opening the details panel.

This also suppresses the commit subject tooltip while the context menu is open, matching the pattern used elsewhere to avoid tooltips appearing on top of context menus.

Self-Review Checklist:

Release Notes:

- Added a context menu the git graph entrees

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -1895,15 +1895,17 @@ impl GitGraph {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
-        if self.graph_data.commits.get(index).is_none() {
+        let Some(commit) = self.graph_data.commits.get(index) else {
             return;
-        }
+        };
+        let short_sha = commit.data.sha.display_short();
 
         let focus_handle = self.focus_handle.clone();
         let git_graph = cx.entity();
         let context_menu = ContextMenu::build(window, cx, |context_menu, window, _| {
             context_menu
                 .context(focus_handle)
+                .header(format!("Commit {short_sha}"))
                 .entry(
                     "View Commit",
                     Some(OpenCommitView.boxed_clone()),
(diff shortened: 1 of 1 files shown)


crates/git_graph/src/git_graph.rs
```diff
@@ -1481,7 +1497,9 @@ impl GitGraph {
                     div()
                         .id(ElementId::NamedInteger("commit-subject".into(), idx as u64))
                         .overflow_hidden()
-                        .tooltip(Tooltip::text(subject))
+                        .when(!has_context_menu, |this| {
+                            this.tooltip(Tooltip::text(subject))
+                        })
```
JosephTLyons, line 1502: Prevents the tooltip from covering the context menu

Anthony-Eid:
Your changes look great!

Anthony-Eid (approved):

PR #55744, opened 2026-05-05
agent: LSP tools

cameron1024:
Adds 4 (technically 5) new tools to the zed agent, corresponding to LSP actions:
- `find_references`
- `goto_definition`
- `rename_symbol`
- `get_code_actions` and `apply_code_actions`

Notes:
- `rename_symbol` skips doing a `prepare_rename`. If there is nothing to rename at the position, it will forward the error to the agent
- The code action tools are stateful. The state is stored in the `get_code_actions` tool itself as a `PendingCodeActions`. It is not passed into/out of subagents. Calling `apply_code_actions` without calling `get_code_actions` first is an error, but I've never seen an agent do this

Symbols are identified by:
- file name
- line number
- symbol

If there is no substring match on that line for the symbol text, it is an error. If there are multiple, it chooses the first. This may not be great if you have a line like: `fn convert(x: foo::Something) -> bar::Something` - the second `Something` is a different symbol, but is inacessible to these tools. Probably fine for now, but we can look into improving

Release Notes:

- N/A


bennetbo (approved):
Let's experiment with it and see how it feels

AJenbo:
Made some improvements based on testing https://github.com/zed-industries/zed/pull/55803

AJenbo:
The agent seems surprised that the rename tool doesn't save the files it edits:

> Many unsaved files. The rename tool updated them but they weren't saved. Let me save all of them:

This caused it to have to do grep to find all unsaved files and then save them one by one or simply do the edits again manually. It probably spent as much time trying to recover form this as it would have if it did all the edits traditionally. This should probably be addressed if the tool it to be made generally available, either by having it save, letting it be an option, or making it clear to the agent that the tool doesn't save the changes.

PR #55747, opened 2026-05-05
editor: Extract Diagnostics code out of `editor.rs`

GoldStrikeArch:
cc @SomeoneToIgnore

## Summary

This mechanically extracts diagnostics-related editor code into `crates/editor/src/editor/diagnostics.rs` while preserving the existing public API via re-exports.

## Testing

- `cargo check -p editor --lib`
- `cargo check -p diagnostics --lib`
- `cargo check -p diagnostics --tests`

Self-Review Checklist:

Release Notes:

- N/A


crates/editor/src/editor.rs
```diff
@@ -57,7 +57,15 @@ mod signature_help;
 #[cfg(any(test, feature = "test-support"))]
 pub mod test;
 
+#[path = "editor/diagnostics.rs"]
+mod diagnostics_impl;
```
SomeoneToIgnore: *

```suggestion
mod diagnostics;
```

Could it be simpler?
There's a `diagnostics` crate, but that depends on `editor`, not vice-versa, so should be fine?

* Also, 

`diagnostic_style` is not needed to be `pub`, maybe even could get private and get moved into the `display_map`?

Same for `ActiveDiagnosticGroup`, definitely does not need to be exposed from `editor`.
Curious if you could find more redundancies like that and tidy them up?

crates/editor/src/editor/diagnostics.rs
```diff
@@ -0,0 +1,513 @@
...
+        self.change_selections(Default::default(), window, cx, |s| {
+            s.select_ranges(vec![
+                next_diagnostic.range.start..next_diagnostic.range.start,
+            ])
+        });
+        self.activate_diagnostics(buffer_id, next_diagnostic, window, cx);
+        self.refresh_edit_prediction(false, true, window, cx);
+    }
+
+    pub(super) fn refresh_active_diagnostics(&mut self, cx: &mut Context<Editor>) {
```
SomeoneToIgnore: Optional NIT: not insisting, but sort of odd to see `pub` interleaving with `pub(super)` modifiers, I'd order them from public to private first to see the action handlers simpler (?).

SomeoneToIgnore (commented):
Great, thank you for the example, the idea seems to be well demonstrated and great if we can process each move sequentially, making some notes between steps if necessary.

Here, some overly public members are somewhat odd to expose: let's check out the usages of each entry moved and see if that could be inlined/made private/etc.
In some cases, cargo does not emit warnings for items that are `pub` but not used anywhere, so it's nice to make some tidy up if we're onto it anyway.

GoldStrikeArch:
@SomeoneToIgnore

Thanks, agreed. I tightened this up by:

- removing the public re-exports for `diagnostic_style` and `ActiveDiagnosticGroup`
- moving `diagnostic_style` into `display_map` as a private helper
- moving the `pub(super)` diagnostics helpers below the public methods

I kept the remaining public methods because they still have cross-crate callers from `diagnostics`, `git_ui`, `agent_ui`, etc.

crates/editor/src/editor/diagnostics.rs
GoldStrikeArch: I moved `pub(super)` to the bottom

crates/editor/src/editor.rs
GoldStrikeArch: Yeah, that makes sense, changed it to be `diagnostics` and inlined `diagnostic_style` into `display_map`. 

making `ActiveDiagnosticGroup` private required to patch some tests, so I did it as well

SomeoneToIgnore (approved):
Nice, thank you.

PR #55757, opened 2026-05-05
agent_servers: Include stderr in ACP startup exit errors

benbrandt:
Previously, we weren't waiting on the status future early enough so we
would just hang if we weren't able to start the agent process.

I also added the recent stderr logs in there to help the user debug the
issue, since it is likely relevant in these cases.

image

Self-Review Checklist:

Release Notes:

- acp: Improve error messages if the ACP agent fails to start.


bennetbo (approved):
🚢

PR #55760, opened 2026-05-05
Skip Git tracking for invisible worktrees

Veykril:
Release Notes:


cole-miller (changes requested):
I think we can remove the `is_visible` filtering at the git store level now but otherwise makes sense to me!

cole-miller (approved):

PR #55770, opened 2026-05-05
acp: Allow resolving older npm package versions

### LINKED ISSUE #516: Failed to Launch - Internal error: "server shut down unexpectedly"
Due to the recent increase in supply chain attacks, I have added this line to my ~/.npmrc file:

```
min-release-age=7
```

Since then, every time I try to start a new Claude Agent conversation in Zed, I get this error:

Image

When I remove the min-release-age security setting, the issue disappears and I can start new Claude Agent conversations again.

However, I do not want to lower my security settings just to use Zed Agents.

benbrandt:
Lots of people are using `min-release-age` in their .npmrc files these days.

I saw two options:

1. Force min-release-age=0 so we can always install the latest
2. Be more lenient in what we allow

I opted for 2, which means we convert `package@0.1.2` to `package@<=0.1.2`. This means npm can find the latest version we can that meets the user's requirements. 

The downside is, the registry args/env may or may not work with the resolved version, but that should at least surface better thanks to #55757 

There is also the issue that npm will cache package metadata and an older version it has cached would still resolve. However, once the metadata is updated, npm does use the newer tarball at least, so it will update eventually.

It's a tradeoff, but I'd rather start with this until we have a better solution on the ACP registry, rather than have users be upset becaue we installed packages in a way they didn't want.

Self-Review Checklist:

Closes https://github.com/agentclientprotocol/claude-agent-acp/issues/516

Release Notes:

- acp: Better support min-release-age settings for npx-based agents from the registry

### DIFF
--- a/crates/agent_servers/src/acp.rs
+++ b/crates/agent_servers/src/acp.rs
@@ -225,26 +225,6 @@ fn exited_load_error_with_stderr(status: ExitStatus, debug_log: &AcpDebugLog) ->
     }
 }
 
-fn should_retry_with_fallback_command(error: &anyhow::Error) -> bool {
-    let Some(LoadError::Exited {
-        stderr: Some(stderr),
-        ..
-    }) = error.downcast_ref::<LoadError>()
-    else {
-        return false;
-    };
-
-    is_npm_release_age_resolution_failure(stderr)
-}
-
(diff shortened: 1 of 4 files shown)


bennetbo (approved):

Qard:
@benbrandt I noticed the github copilot plugin also has the same issue. Perhaps the solution could be made more generic to fix that too?

Issue #55186, opened 2026-04-29
edit_file mode enum parameter is getting double-quoted


m-rutter:
### Reproduction steps

I'm running a llama.cpp server for local inference and I'm using the openai compatible intergration to intergrate locally run models into zed. I've suddenly noticed in the last day or so that models that were very capable of making document edit or write tool calls suddenly and consistently cannot, and I'm beginning to suspect that there might be a recently introduced bug in zed. I'm not entirely sure how to reproduce this manually.

```json
{
  "error": "Failed to receive tool input: unknown variant `\"edit\"`, expected `write` or `edit`"
}
```

What I think is happening is somewhere in serialisation escaped double quotes are getting added to the `mode` field value which the LLM is not writing. The model is writing `"mode": "edit"` but its getting serialised somehwere as  `"mode": "\"edit\""` and then failing when its getting read by serde deserialisation code.

### Current vs. Expected behavior

LLMs via the openapi compatiable interface can use the edit_file tool calls without their inputs getting double quoted.

### Zed version and system specs

Zed: v0.233.10+stable.250.350f33871e84d5ed83a403523e9490b9bf014a94 (Zed) 
OS: macOS 26.3.1
Memory: 32 GiB
Architecture: aarch64

### Relevant Zed settings

<details><summary>settings.json</summary>

```json
{
  "agent": {
    "default_model": {
      "provider": "llama-seer",
      "model": "llama.cpp",
      "enable_thinking": false
    },
    "favorite_models": [],
    "model_parameters": []
  },
  "theme": {
    "mode": "system",
    "light": "One Light",
    "dark": "Ayu Mirage",
  },
  "buffer_font_size": 14,
  "ui_font_size": 14,
  "buffer_font_family": "Berkeley Mono",
  "ui_font_family": "Berkeley Mono",
  "vim_mode": true,

  "agent_servers": {
    "claude-acp": {
      "type": "registry",
    },
  },
  "edit_predictions": {
    "provider": "none",
  },
  "show_edit_predictions": false,

  "language_models": {
    "openai_compatible": {
      "llama-seer": {
        "api_url": "http://seer.elk-ruler.ts.net:8080/v1",
        "available_models": [
          {
            "display_name": "llama.cpp",
            "name": "llama.cpp",
            "max_tokens": 262144,
            "capabilities": {
              "tools": true,
              "images": false,
              "parallel_tool_calls": true,
              "prompt_cache_key": true,
            },
          },
        ],
      },
    },
  },

[...]

PR #55776, opened 2026-05-05
agent: Allow model to provide stringified `timezone` in `now` tool

bennetbo:
See https://github.com/zed-industries/zed/issues/55186#issuecomment-4376114420

I think the recent changes to the tool schema in #55763 will make this more unlikely, but does not hurt to allow the model to provide `"utc"`.

Release Notes:

- N/A


benbrandt (approved):

maxbrunsfeld:
@zed-industries/approved

PR #55782, opened 2026-05-05
markdown: Fix escaping non-ASCII chars

### LINKED ISSUE #55704: markdown: MarkdownEscaper escapes non-ASCII characters due to lossy `as u8` cast
### Reproduction steps

1. Use any LSP server that emits Cyrillic text in `Diagnostic.message`.
2. Open a file that triggers a diagnostic whose message contains both:
   - a Cyrillic letter from the ranges `U+043A`–`U+0440` (`к л м н о п р`) or `U+0421`–`U+042F` (`С Т У Ф Х Ц Ч Ш Щ Ъ Ы Ь Э Ю Я`), and
   - any markdown-special character (`'`, `"`, `(`, `)`, `#`, `/`, …).
3. Inspect the diagnostic message in the editor.

### Current behavior

Spurious `\` characters appear before specific Cyrillic letters:

> `Отсутствует \п\р\обе\л с\п\рава \от ','`
>
> `В\мест\о уста\ревшег\о св\ойства "\Эта\Ф\о\р\ма" с\ледует ис\п\о\льз\овать "\Эт\отОбъе\кт"`
>
> `П\р\оцеду\ра \нах\одится в\не \об\ласти (#Об\ласть/#Region). Весь \к\од \м\оду\ля д\о\лже\н быть \о\рга\низ\ова\н в \об\ласти д\ля \лучшей ст\ру\кту\ры.`

Copying the message yields the clean text — the backslashes are inserted by the renderer.

### Expected behavior

Display the message exactly as the LSP server sent it, e.g.:

> `Отсутствует пробел справа от ','`

### Root cause

`crates/markdown/src/markdown.rs`, `pub fn escape` (~line 457):

```rust
for c in s.chars() {
    escaper.next(c as u8).write_to(c, &mut output);
}
```

`MarkdownEscaper::next` decides to prefix a backslash via `byte.is_ascii_punctuation()`. For non-ASCII codepoints, `c as u8` truncates to the low 8 bits, which can land inside ASCII-punctuation ranges — e.g. `р` (U+0440) → `0x40` → `@`, `о` (U+043E) → `0x3E` → `>`, `Э` (U+042D) → `0x2D` → `-`. Per CommonMark §6.1, `\X` is consumed only when `X` is ASCII punctuation, so for non-ASCII characters the backslash is rendered literally.

The earlier fix in #51766 addressed visible escapes inside indented code blocks but did not touch this lossy-cast path. The bug affects any script outside ASCII whose codepoints land in those low-bit ranges — Cyrillic, Greek, Hebrew, Arabic, CJK, etc.

Zed 1.0.1, macOS 15.7.4

### Attach Zed log file

n/a

### Relevant Zed settings / Keymap

n/a

n/a

### WSL

No

alkinun:
Fixes #55704

The `escape` function in `crates/markdown/src/markdown.rs` was calling `c as u8` on the `char`s before passing to `MarkdownEscaper::next()`. This strips non ASCII Unicode codepoints down to just their low 8 bits which might be in the ASCII punctuation range and thus cause an extra backslash to be added in front of these non ASCII chars.
 
Release Notes:

- Fixed a bug where non-ASCII chars in diagnostic messages were incorrectly rendered with spurious `\` characters

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -658,8 +658,22 @@ impl Markdown {
 
         let mut escaper = MarkdownEscaper::new();
         let mut output = String::with_capacity(output_len);
-        for c in s.chars() {
-            escaper.next(c as u8).write_to(c, &mut output);
+        let mut byte_idx = 0;
+
+        while byte_idx < s.len() {
+            let byte = s.as_bytes()[byte_idx];
+            let action = escaper.next(byte);
+
+            if byte.is_ascii() {
+                action.write_to(byte as char, &mut output);
+                byte_idx += 1;
+            } else {
+                // Multi-byte UTF-8 characters never trigger escaping actions,
+                // so just decode and pass them through.
+                let c = s[byte_idx..].chars().next().unwrap();
+                output.push(c);
+                byte_idx += c.len_utf8();
+            }
         }
         output.into()
     }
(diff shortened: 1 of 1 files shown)


agu-z (changes requested):
Instead of dealing with it at the call site, why don't we just make `MarkdownEscaper` operate on `char`?  

```rs
fn next(&mut self, c: char) -> EscapeAction
```

That makes the whole class of "low byte of a non-ASCII char looks like punctuation" bugs unrepresentable, instead of relying on callers to pre-validate.

If we do that, we should remove this comment: https://github.com/zed-industries/zed/blob/3bd9d13b63fc5a5ffa39326597bc4fd91adc82d1/crates/markdown/src/markdown.rs#L434-L435

alkinun:
Sure, that seems to be a better way, just commited the changes.

agu-z (approved):
Nice! Thank you for your contribution

PR #55783, opened 2026-05-05
acp_thread: Render execute tool titles as plain text

benbrandt:
We had a regression where the labels were being rendered as markdown,
which is usually not what you want on a command

Self-Review Checklist:

Release Notes:

- N/A


bennetbo (approved):

lixiaoyan:
I believe we should not treat other tool calls as markdown either. otherwise e.g. `__tests__` will become some bold texts
image

https://github.com/zed-industries/zed/issues/56034

PR #55787, opened 2026-05-05
Strengthen guidance for git commands in the terminal tool

rtfeldman:
Move guidance about pty-blocking commands from the `TerminalToolInput` struct-level doc comment to the `command` field's doc comment so it's surfaced more prominently to the model in the tool's JSON schema, and call out `--no-pager` and `GIT_EDITOR=true` explicitly for git operations.

In practice, the previous wording about `git --no-pager diff` was easy for agents to overlook, leading to terminal calls that block waiting on `less` (for `git log`/`git diff`/`git show`) or on an interactive editor (for `git rebase`/`git commit`/`git merge`).

Unit eval change before/after wording change:

image

Closes AI-154

Release Notes:

- Zed Agent's terminal tool now much more consistently uses `--no-pager` and `GIT_EDITOR` with `git` commands

### DIFF
--- a/crates/agent/src/tools/evals.rs
+++ b/crates/agent/src/tools/evals.rs
@@ -1,2 +1,4 @@
 #[cfg(all(test, feature = "unit-eval"))]
 mod edit_file;
+#[cfg(all(test, feature = "unit-eval"))]
+mod terminal_tool;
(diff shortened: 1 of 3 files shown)


benbrandt (approved):
thanks!

PR #55803, opened 2026-05-05
agent: Improve LSP tool symbol resolution and definition lookup

AJenbo:
This PR improves the `SymbolLocator::resolve()` used by all LSP tools (`go_to_definition`, `find_references`, `rename_symbol`, `get_code_actions`) to be more robust when agents provide slightly inaccurate positions, and improves `go_to_definition` to prefer type definitions.

### Symbol resolution improvements

The existing implementation does a plain substring match on the exact line specified. This fails in several common cases:

**1. False substring matches** — Searching for `Payment` on a line containing `captureReservedPayment(Payment $payment)` would match inside `captureReservedPayment` first, positioning the cursor in the middle of the wrong symbol.

**Fixed with word boundary matching:** Only standalone occurrences are matched. Word characters are alphanumeric and `_`. Special prefixes (`$`, `@`, `#`) skip the left-boundary check since they already act as delimiters.

**2. Off-by-one line numbers** — Agents frequently land on the blank line before code (e.g. the line before a function signature) because selection ranges and outline results sometimes point there.

**Fixed with nearby line search (±4 lines):** If the symbol isn't found on the exact line, nearby lines are searched before giving up.

**3. Wrong casing** — Agents sometimes use `PaymentToken` when the code has `paymentToken`, especially when paraphrasing from memory.

**Fixed with case-insensitive fallback:** Case-sensitive matching is tried first and preferred; case-insensitive is used only when no exact match exists.

### Definition lookup improvement

`go_to_definition` now tries `type_definitions` first, falling back to regular `definitions`. This means navigating to a variable like `$order` takes you to the `Order` class rather than the assignment site, which is usually more useful when exploring code.

### Testing

Unit tests cover word boundary matching, case-insensitive fallback, special prefix handling, and no-match cases.

Release Notes:

- Improved LSP agent tools to be more robust when the agent provides slightly inaccurate symbol positions (word boundary matching, nearby line search, case-insensitive fallback).

### DIFF
--- a/crates/agent/src/tools/go_to_definition_tool.rs
+++ b/crates/agent/src/tools/go_to_definition_tool.rs
@@ -69,11 +69,10 @@ impl AgentTool for GoToDefinitionTool {
 
             let resolved = input.symbol.resolve(&project, cx).await?;
 
-            let definitions_task = project.update(cx, |project, cx| {
-                project.definitions(&resolved.buffer, resolved.position, cx)
-            });
-
-            let definitions = definitions_task
+            let definitions = project
+                .update(cx, |project, cx| {
+                    project.definitions(&resolved.buffer, resolved.position, cx)
+                })
                 .await
                 .map_err(|e| format!("Go to definition failed: {e}"))?
                 .unwrap_or_default();
--- a/crates/agent/src/tools/symbol_locator.rs
+++ b/crates/agent/src/tools/symbol_locator.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::cmp;
 use std::fmt;
 
 use gpui::{App, AsyncApp, Entity};
@@ -8,6 +8,10 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use text::ToPoint as _;
 use text::{Anchor, Point};
+
+/// The number of lines to search above and below when a symbol isn't found on the exact line.
+/// Agents often land on a blank line before the actual code due to selection ranges.
+const NEARBY_LINE_SEARCH_RADIUS: u32 = 4;
 
 /// Identifies a specific symbol (declaration or usage) in the source code.
 ///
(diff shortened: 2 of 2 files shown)


crates/agent/src/tools/go_to_definition_tool.rs
```diff
@@ -69,14 +75,27 @@ impl AgentTool for GoToDefinitionTool {
 
             let resolved = input.symbol.resolve(&project, cx).await?;
 
-            let definitions_task = project.update(cx, |project, cx| {
-                project.definitions(&resolved.buffer, resolved.position, cx)
+            // Try type definition first (e.g. navigates $order to the Order class),
+            // fall back to regular definition (e.g. the assignment site).
+            let type_def_task = project.update(cx, |project, cx| {
```
cameron1024: I'm not sure this is a good default behaviour. Getting type defintions can be useful, definitely, but so can finding the declaration of a variable (or struct/class/etc. field). I don't think one is necessarily more useful than the other.

I think there's scope for an extra tool, or perhaps merging the "goto X" tools into a single tool with a field that allows the agent to specify whether it wants to see the declaration, type definition, references, implementations, etc. For now though, we're just evaluating a limited subset.

crates/agent/src/tools/symbol_locator.rs
```diff
@@ -104,48 +108,116 @@ impl fmt::Display for LocationDisplay {
...
 
-    let mut window: VecDeque<char> = VecDeque::with_capacity(needle_chars.len());
-    let mut byte_offsets: VecDeque<usize> = VecDeque::with_capacity(needle_chars.len());
-    let mut byte_offset = 0usize;
+/// Find all word-bounded occurrences of `needle` in `haystack`, returning byte offsets.
+/// Respects special prefixes like `$`, `@`, `#` that aren't word characters but are part
+/// of the symbol (e.g. PHP variables `$foo`).
+fn find_word_bounded<'a>(haystack: &'a str, needle: &'a str) -> impl Iterator<Item = usize> + 'a {
+    find_word_bounded_impl(haystack, needle, false)
+}
```
cameron1024: This function is quite complex, and encodes some language-specific details that I'm not sure we can rely on in all cases. For example, the string `foo$` can be a full identifier (e.g. in Dart, where `$` is a non-special symbol when used in identifiers), or it could be an identifier and operator (e.g. in Lean4, where `$` is function application). Or `a-b` is an identifier in `Nix`, but in C-style langauges it's `a` minus `b`.

The ideal solution would be using the syntax tree, but that's a relatively complex change. 

The point you make about `void handlePayment(Payment payment)` not being able to resolve the type is a fair point though, and we should handle this better. For this purpose, I actually think ignoring casing is actually not the right approach, since:
- the case can provide useful information (e.g. `const FOO: Foo = ...`)
- it doesn't actually work in languages that use `snake_case` or `kebab-case` heavily

I think a simpler implementation could:
- split the haystack based on some conservative regex for non-identifier characters (whitespace, and symbols excluding `$`, `@`, `#`, `-`, and maybe some others)
- search the current line for an exact symbol match
- if that fails, try the line before and the line after
- repeat, expanding the context each time, failing after some maximum context size

crates/agent/src/tools/symbol_locator.rs
```diff
@@ -220,20 +300,54 @@ impl SymbolLocator {
     }
 }
 
+fn format_line_display(line_text: &str) -> String {
+    let display: String = line_text
+        .chars()
+        .skip_while(|c| c.is_whitespace())
+        .take(MAX_LINE_DISPLAY_LEN)
+        .collect();
+    display.trim_end().to_string()
```
cameron1024: I'm not sure if we need to strip whitespace - seems like a bit of a footgun if we end up concatenating multiple lines together, we lose indentation information.

But if we are going to do it, this method doesn't need to allocate

cameron1024 (changes requested):

crates/agent/src/tools/symbol_locator.rs
AJenbo: Ok, it no longer strips leading whitespace. Just truncates to `MAX_LINE_DISPLAY_LEN` and trims trailing whitespace using `&str` slicing to reduce allocation.

crates/agent/src/tools/symbol_locator.rs
AJenbo: Replaced the complex `find_word_bounded`/`find_word_bounded_impl`/`find_word_bounded_case_insensitive` functions with a single `find_symbol_on_line` that uses `is_identifier_char` (alphanumeric, `_`, `$`, `@`, `#`, `-`) for token boundary checks. Removed case-insensitive fallback entirely.

crates/agent/src/tools/go_to_definition_tool.rs
AJenbo: It now just calls `project.definitions()` directly instead of trying `type_definitions` first. Removed the doc comment paragraph about type definitions.

PR #55805, opened 2026-05-05
Derive data directory names from APP_NAME constant

rtfeldman:
Replace all hardcoded `"Zed"` and `"zed"` directory names in `config_dir()`, `data_dir()`, `state_dir()`, and `temp_dir()` with a single `APP_NAME` constant in the `paths` crate.

- On macOS/Windows (native paths like Application Support, LocalAppData, Caches), `APP_NAME` is used directly (`"Zed"`).
- On Linux/FreeBSD (XDG-style paths), `app_name_lowercase()` is used (`"zed"`).

This ensures forks that change `APP_NAME` automatically get their own data directories, preventing them from stomping on Zed users' databases and config.

Release Notes:

- N/A


crates/zed/src/main.rs
```diff
@@ -4,6 +4,16 @@
...
 
+// Ensure the binary name stays in sync with APP_NAME so that the paths used
+// at runtime (data dir, config dir, etc.) match what the binary is called.
+const _: () = assert!(
+    paths::APP_NAME_LOWERCASE
+        .as_bytes()
+        .eq_ignore_ascii_case(env!("CARGO_BIN_NAME").as_bytes()),
+    "paths::APP_NAME_LOWERCASE must match the binary name. \
+     Forks: update APP_NAME in crates/paths/src/paths.rs when renaming the binary.",
+);
```
Anthony-Eid, line 15: This is awesome!

Anthony-Eid (approved):
Looks good! One consequence of this is forks won't be able to reuse downloaded LSPs, DAPs, extensions, etc. I think this change is still worth making though

PR #55813, opened 2026-05-05
optimization: [editor] skip diff hunk iteration over folded buffers

rnjohn:
I became aware of this issue because of a react native upgrade I was doing in another project. The package-lock.json was very big and had a whole lot of changes, and if you collapsed the file diff, Zed would lag terribly as long as the collapsed file appeared in the screen.

This change computes the gaps between folded buffer regions and queries `diff_hunks_in_range` only for those gaps, so the iterator never enters folded content at all. The post-hoc `folded_buffers.contains()` filter is also removed since exclusion is now handled upstream by range construction.

Release Notes:
Improved editor performance when large files with many diff hunks are collapsed in a multi-buffer view.


rnjohn:
@cla-bot check

PR #55830, opened 2026-05-05
Add NEAR AI Cloud as LLM Provider for Zed Agent

itexpert120:
Self-Review Checklist:

Release Notes:
- Added [NEAR AI Cloud](https://cloud.near.ai) as a LLM provider for Zed Agent. 

Screenshots

In LLM Provider settings - Unconfigured
image

In LLM Provider settings - Configured
image

Working in Chat
image

Model Selection
image


benbrandt:
Hi, thanks for this! for new providers, since each comes wiht some amount of maintenance cost, I think it would be better to start a discussion so we can see how much user demand there is before adding this. thanks

PR #55838, opened 2026-05-06
workspace: Prompt to save unsaved buffers when opening a new project

### LINKED ISSUE #55726: Opening a file or project loses unsaved file
### Reproduction steps

1. Launch Zed
2. In the untitled blank document, type some text
3. Open something (File > Open..., File > Open Recent..., Open Recent Project)
4. The edited blank document is lost
5. The opened file / project shows dirty state (dot in red stoplight / close window button)

Current behavior is outlined in the steps to reproduce. I would expect a confirmation dialog providing me an option to save the unsaved file, or for the open operation to keep the unsaved buffer open in a tab.

OS: macOS 15.7.4
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

saberoueslati:
## Context

Closes #55726

When Zed starts with an untitled buffer, the user types content, then opens a file or project, the typed content disappears silently with no save prompt.

**Root cause:** `prepare_to_close(CloseIntent::ReplaceWindow)` called `save_all_internal(SaveIntent::Close)`. That function has a hot-exit path: for each dirty item it first attempts to serialize the content into the session DB.
For a real `Editor` with `restore_unsaved_buffers = true` (the default), this serialization succeeds and the item is removed from the pending list, so no "Save / Don't Save / Cancel" prompt is ever shown.

Hot-exit is intentional for `CloseWindow` and `Quit`, the same workspace id is restored on next launch, so the content comes back. For `ReplaceWindow` the workspace is replaced with a brand-new id, so the serialized content is never
restored and is effectively lost.

**Fix:** add `SaveIntent::CloseReplace`, used only for `ReplaceWindow`. It bypasses the hot-exit serialization block in `save_all_internal`, so all dirty items fall through to the per-item `save_item` prompt ("Save / Don't Save /
Cancel") unchanged. `Event::Activate` is also emitted before the prompt loop so the workspace is brought into focus first (matching existing `SaveIntent::Close` behaviour).

Video of the manual test below :

[Screencast from 2026-05-06 02-19-09.webm](https://github.com/user-attachments/assets/318798c4-78ce-4610-b74d-20a2c7762954)

## How to Review

- **`crates/workspace/src/pane.rs`** : 
new `SaveIntent::CloseReplace` variant (after `Close`); `save_item` prompt condition extended to include it so the per-item "Save / Don't Save / Cancel" dialog fires.

- **`crates/workspace/src/workspace.rs`** : 
`prepare_to_close` now selects `SaveIntent::CloseReplace` when `close_intent == CloseIntent::ReplaceWindow`
  `save_all_internal` emits `Event::Activate` before the `save_item` loop for `CloseReplace` (the hot-exit block already handled this for `SaveIntent::Close`)
regression test `test_replace_window_prompts_for_serializable_dirty_items` exercises the exact path the bug was on a serializable dirty item plus `CloseIntent::ReplaceWindow`.

## Self-Review Checklist

Release Notes:

- Fixed unsaved content in untitled buffers being silently discarded when opening a new project

### DIFF
--- a/crates/workspace/src/pane.rs
+++ b/crates/workspace/src/pane.rs
@@ -94,6 +94,10 @@ pub enum SaveIntent {
     SaveAs,
     /// prompt "you have unsaved changes" before writing
     Close,
+    /// like `Close`, but skips hot-exit serialization; used when the workspace
+    /// is being replaced by a new project, where session-DB content would not
+    /// be restored under the new workspace id
+    CloseReplace,
     /// write all dirty files, don't prompt on conflict
     Overwrite,
     /// skip all save-related behavior
@@ -2371,7 +2375,7 @@ impl Pane {
                 }
             }
         } else if is_dirty && (can_save || can_save_as) {
-            if save_intent == SaveIntent::Close {
+            if matches!(save_intent, SaveIntent::Close | SaveIntent::CloseReplace) {
                 let will_autosave = cx.update(|_window, cx| {
                     item.can_autosave(cx)
                         && item.workspace_settings(cx).autosave.should_save_on_close()
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -3305,9 +3305,14 @@ impl Workspace {
                 }
             }
 
+            let save_intent = if close_intent == CloseIntent::ReplaceWindow {
+                SaveIntent::CloseReplace
+            } else {
+                SaveIntent::Close
+            };
             let save_result = this
                 .update_in(cx, |this, window, cx| {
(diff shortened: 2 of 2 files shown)


ChristopherBiscardi:
This doesn't feel like the right solution to me. Any existing workspace retains the unsaved buffers, even through a quit/restart, so doesn't require this solution. The issue mentioned is only for "new"/unsaved workspaces.

saberoueslati:
@ChristopherBiscardi I'll work on it more tomorrow and open a new PR with a more relevant solution to the specific unsaved workspace use case.

PR #55841, opened 2026-05-06
agent: Fix edit file always failing when using DeepSeek v4 Pro

### LINKED ISSUE #55398: `edit_file` always fail with normal mode when using DeepSeek V4 Pro
### Reproduction steps

use deepseek-v4-pro official api endpoint
let ai writes codes

Image

it always fail with normal edit mode, ai will try many times and waste the token, then ai will recognize it should use the overwrite mode, to rewrite the whole file

that should not happen, it should just use normal edit mode, to reduce tokens

Zed: v1.0.0+stable (Zed) 
Memory: 62.7 GiB
Architecture: x86_64
GPU: NVIDIA GeForce RTX 4080 || NVIDIA || 595.71.05

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

- provider: deepseek
- model: deepseek v4 pro with max thinking
- mode: write agent

None

Alejandro-HUB:
Fix premature old_text finalization when model streams fields out of order

The streaming parser for the `edit_file` tool used `new_text`'s
presence as a signal that `old_text` was complete, emitting it with
`done: true`. This fails when models stream JSON fields out of schema
order — specifically DeepSeek V4 Pro, which can emit `new_text` before
`old_text` has been fully received.

When this happens, `old_text` gets finalized as `""` (the partial JSON
fixer closing an incomplete string), the real value arriving in later
chunks is silently ignored, and the fuzzy matcher finds no match.

The fix: only finalize `old_text` during streaming if actual content has
been emitted (`old_text_emitted_len > 0`) or the value is non-empty.
Empty `old_text` with `new_text` present is deferred to `finalize_edits`,
which has the complete data.

Release Notes:

- Fixed `edit_file` tool failing with "Could not find matching text"
  when using DeepSeek V4 Pro

Closes #55398

### DIFF
--- a/crates/agent/src/tools/edit_file_tool/streaming_parser.rs
+++ b/crates/agent/src/tools/edit_file_tool/streaming_parser.rs
@@ -80,29 +80,44 @@ impl StreamingParser {
             if let Some(old_text) = &partial.old_text
                 && !state.old_text_done
             {
-                if partial.new_text.is_some() {
-                    // new_text appeared, so old_text is done — emit everything.
-                    let start = state.old_text_emitted_len.min(old_text.len());
-                    let chunk = normalize_done_chunk(old_text[start..].to_string());
-                    state.old_text_done = true;
-                    state.old_text_emitted_len = old_text.len();
+                let safe_end = safe_emit_end_for_edit_text(old_text);
+                let has_new_data = safe_end > state.old_text_emitted_len;
+
+                // old_text is done when new_text is present AND either:
+                // - we've already emitted some old_text (it was streaming), or
+                // - old_text is non-empty (arrived complete in this chunk).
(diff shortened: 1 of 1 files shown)


Alejandro-HUB:
@cla-bot check

bennetbo:
Hey, thanks for working on this and finding the cause of the issue. The code here does not work for the case where we receive both old_text and new_text in the same chunk. I made a PR to handle both cases (new_text appearing first, new_text appearing with old_text) here #55894, closing.

Alejandro-HUB:
@bennetbo I see; I didn't consider the case where both are received in the same chunk. Excited to see your PR merged so that DeepSeek can operate as expected. 👍

PR #55855, opened 2026-05-06
editor: Extract rewrap and config out of `editor.rs`

GoldStrikeArch:
cc @SomeoneToIgnore

## Summary

This mechanically extracts editor config and reflow-related code into `crates/editor/src/config.rs` and `crates/editor/src/rewrap.rs`, while preserving existing behavior and keeping externally-used APIs public where needed.

Self-Review Checklist:

Release Notes:

- N/A


crates/editor/src/editor.rs
```diff
@@ -57,8 +57,12 @@ mod signature_help;
 #[cfg(any(test, feature = "test-support"))]
 pub mod test;
 
+#[path = "editor/config.rs"]
+mod config;
 #[path = "editor/diagnostics.rs"]
 mod diagnostics;
+#[path = "editor/reflow.rs"]
+mod reflow;
```
```suggestion
mod rewrap;
```

We do not use `reflow` anywhere but in Vim, so this word seems wrong.

Also, this starts to look odd: let's avoid `path` directives entirely and put the modules next to `editor.rs` instead.

GoldStrikeArch: Alright, I moved all modules (and diagnostics.rs from yesterday as well) beside `editor.rs`

SomeoneToIgnore (approved):
Great, thank you for pushing for this.

Eventually, a better module hierarchy is a good thing to get, but we'd better deal with the basic layer, splitting, first (same as we do not rush with the `EditorMode`-related changes yet).

crates/editor/src/config.rs
```diff
@@ -0,0 +1,349 @@
...
+        self.show_line_numbers = Some(show_line_numbers);
+        cx.notify();
+    }
+
+    pub fn disable_expand_excerpt_buttons(&mut self, cx: &mut Context<Self>) {
+        self.disable_expand_excerpt_buttons = true;
+        cx.notify();
+    }
+
+    pub fn set_delegate_open_excerpts(&mut self, delegate: bool) {
```
SomeoneToIgnore: `set_delegate_` seems to be a group of special methods that do not need `cx.notify()` based off the logic checks I just did.

Let's move this one to the rest below to be less confusing to read.

crates/editor/src/config.rs
```diff
@@ -0,0 +1,349 @@
+use super::*;
+
+impl Editor {
+    pub fn style(&mut self, cx: &App) -> &EditorStyle {
+        if self.style.is_none() {
+            self.style = Some(self.create_style(cx));
+        }
+        self.style.as_ref().unwrap()
```
SomeoneToIgnore: `Option` has a `.get_or_insert_with`-esque method to deal with this without unwrapping, let's rewrite this if you do not mind.

crates/editor/src/rewrap.rs
```diff
@@ -0,0 +1,794 @@
...
+        }
+        if self.mode.is_single_line() {
+            cx.propagate();
+            return;
+        }
+
+        self.rewrap_impl(RewrapOptions::default(), cx)
+    }
+
+    pub fn rewrap_impl(&mut self, options: RewrapOptions, cx: &mut Context<Self>) {
```
SomeoneToIgnore: This is one of those important places we need to pay attention to when moving method around: odd boundaries and odd checks across them.

* First, both checks better be carried out once, inside `rewrap_impl`. 
Propagating from there does not make too much sense, though, so maybe we can add just another `&&` to if.

* Seems that `rewrap_impl` should be `rewrap` as it's the only truly public method.
I wonder if we could inline the previous `fn rewrap` entirely then, as all its usages are the action handling and tests?

SomeoneToIgnore (commented):
Sorry, one more set of notes I've got, mostly to keep the same, mentioned in the previous PR, "eliminate `pub`s" idea we exercise along the way with the moves — would be cool to pay attention to this more in the future related PRs.

The rest of the notes are attempts to keep similar things together (that absence of `notify` causes instant questions otherwise) and `unwrap` eradication which I think is always welcome as a drive-by change.

crates/editor/src/rewrap.rs
GoldStrikeArch: Alright, I renamed the public method to `rewrap` (and fixed all usages) and now inside the `register_action` we explicitly propagate

crates/editor/src/config.rs
GoldStrikeArch: Changed to use `match` instead and no `unwrap()` cals now

crates/editor/src/config.rs
GoldStrikeArch: Ok, I moved it to be below but it still needs to be `pub fn` because `agent_ui` is using it.

GoldStrikeArch:
@SomeoneToIgnore 


Agreed, those are new changes:

- Renamed the public rewrap API from `rewrap_impl` to `rewrap`.
- Inlined the editor `Rewrap` action adapter in `element.rs`, preserving the old single-line propagation behavior
- Updated editor internals, editor tests, Vim, and git UI call sites to use the renamed API.

crates/editor/src/element.rs
```diff
@@ -570,7 +570,16 @@ impl EditorElement {
             register_action(editor, window, Editor::move_line_up);
             register_action(editor, window, Editor::move_line_down);
             register_action(editor, window, Editor::transpose);
-            register_action(editor, window, Editor::rewrap);
+            register_action(editor, window, |editor, _: &crate::Rewrap, _, cx| {
+                if editor.mode.is_single_line() {
```
SomeoneToIgnore: Let's remove those checks entirely please, I do not think anything is worth propagating in the rewrap case anywhere from singleton editor up the focus stack.

Just the `editor.rewrap` call.

GoldStrikeArch: Done, now we only call `editor.rewrap` there

SomeoneToIgnore (approved):
Thank you a lot.

PR #55863, opened 2026-05-06
editor: Fix missing background highlights in multibuffer find-all-references

lingyaochu:
Self-Review Checklist:

The root cause is in this part of code:
https://github.com/zed-industries/zed/blob/f5945344cc21ebff07eb7ea04dab0555fbe49007/crates/editor/src/editor.rs#L19446-L19464
The ranges across multibuffer are stored in the order of the HashMap iterator, which is not sorted at all. They are then consumed by `background_highlights_in_range()`, which uses a binary search that requires the ranges to be sorted:
https://github.com/zed-industries/zed/blob/f5945344cc21ebff07eb7ea04dab0555fbe49007/crates/editor/src/editor.rs#L24127-L24135
The fix sorts the ranges by multibuffer anchor position before storing them.

Release Notes:

-  Fixed missing background highlights in "Find All References" when results span multiple files.


lingyaochu:
Before:

https://github.com/user-attachments/assets/02f6c4b9-119f-4f3d-b7a0-21f97757ce32

After:


https://github.com/user-attachments/assets/7c66f779-1f7e-498a-b630-687feb56cb2c

GoldStrikeArch (approved):

GoldStrikeArch (approved):

Veykril (approved):
Amazing find, thank you!

PR #55865, opened 2026-05-06
agent: Add `write_file` tool

bennetbo:
Splits the edit tool into two separate tools `write_file` (previously `mode = write`), and `edit_file` (previously `mode = edit`).
This makes the JSON schema for the `edit_tool` much simpler. We've seen models (especially older ones) struggle with providing `mode = edit + edits` and `mode = write + content` fields. This seems to improve eval scores for Sonnet 4.6 slightly.

Also added two unit evals to ensure that the model uses the tool to create new/override existing files

Release Notes:

- N/A


benbrandt (approved):
Let's try it out!

PR #55866, opened 2026-05-06
language: Reset pooled tree-sitter parsers after cancellation

benbrandt:
It seems new cancellation behavior in tree-sitter caused at least one issue for a user.
Attempting to proactively reset before any parser use to make sure things are clean.

Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/crates/language/src/language.rs
+++ b/crates/language/src/language.rs
@@ -129,6 +129,12 @@ where
             .unwrap();
         parser
     });
+    // Tree-sitter auto-resets the parser at the end of a successful parse,
+    // but the cancellation paths (progress callback returning `Break`,
+    // cancelled balancing) leave outstanding state on the parser. The next
+    // call to `parse_with_options` would then *resume* that cancelled parse
+    // instead of starting fresh.
+    parser.reset();
     parser.set_included_ranges(&[]).unwrap();
     let result = func(&mut parser);
     PARSERS.lock().push(parser);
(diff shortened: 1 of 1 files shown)


benbrandt:
@maxbrunsfeld let me know if there are any tradeoffs here I should be aware of as it seems to only happen in the cancellation path and most places call set_language anyway which I think does a reset.

maxbrunsfeld (approved):
This makes sense - I had not seen that Zed had started using Tree-sitter's cancellation. It looks like it was introduced in https://github.com/zed-industries/zed/pull/45347.

@Veykril @benbrandt It makes sense that we need to reset the Parsers after grabbing them from the pool, now that we are cancelling them mid-parse. But because of the changes in #45347, we will now incur some redundant parsing work done every time a parse does not complete within the initial synchronous budget.

Currently, when we exceed the sync budget, it looks like we restart the parse from scratch, rather than resuming the parse using the same `Parser` instance. I think it would be better to do the parsing all in one background task so that we could hold onto a single parser from the pool, and if it exceeds its sync timeout, resume it immediately. Then, on the main thread, we could simply use `block_with_timeout` on that task.

This doesn't need to block *this* PR though, as it seems to fix an active bug. Just noting a perf issue.

benbrandt:
Gotcha. If we fix that case, would this also make the reset a noop since it should clean itself up if we fix the other issue? 

Just trying to see if there is some overhead I'm introducing in the happy path where it was already reset

maxbrunsfeld:
> If we fix that case, would this also make the reset a noop since it should clean itself up if we fix the other issue?

I still think we'd need to do this reset, because you might drop the task after blocking for the initial sync timeout. So this change seems good either way.

PR #55870, opened 2026-05-06
workspace: Rename "Discard" to "Discard Edits" in file conflict prompt

StrongTheDev:
When a file has unsaved edits and has also changed on disk, the conflict prompt offered "Overwrite / Discard / Cancel". Both "Overwrite" and "Discard" were directionally ambiguous, it wasn't clear whether the disk contents or the buffer edits would be lost. "Discard Edits" makes it explicit that the buffer edits are thrown away and the disk version is reloaded.

Release Notes:

- Improved the file conflict prompt by renaming the "Discard" button to "Discard Edits" to clarify that unsaved edits are discarded in favor of the on-disk contents.


StrongTheDev:
I updated tests. Ran them locally successfully.

smitbarmase (approved):
Thanks, that make sense to me.

PR #55875, opened 2026-05-06
file_finder: Fix `path:line` navigation when target file is already open

### LINKED ISSUE #55551: File finder offers to create file instead of navigating to line when file is already open
When pasting a file path with line number in the file finder, if said file is already open, the finder offers to create it:

### Reproduction steps

1. Create a file e.g. `bug1.py` with some contents in it.

2. Open file finder i.e. Cmd + P, and paste `bug1.py:3`:

Image

3. Zed opens `bug1.py` and places the cursor at line 3.

4. Open file finder again, and paste `bug1.py:3`.

5. The file finder offers to create the file:

Image

When pasting a string in the file finder:

1. if the string is a path that exists, it should offer to open the file if it's not open, or focus on it if it is;
2. if in addition to 1 the string also contains a line specifier, it should jump to it;
3. if the string looks like a path and it doesn't exist, the file finder should offer to create it;
4. if in addition to 3 the string also contains a line specifier, the file finder should not do anything imho.

Zed: v1.0.0+stable.257.5ec84a926ef83865afb92d2a3d1ca3b419572cf9 (Zed)

yeskunall:
Three pieces of state combined to break `path:line` queries when the target file was already open:

- The `CreateNew` lookup ran against `query.raw_query.as_str()`, so a path with a row suffix never matched the existing file and a stray `Create file: foo.rs:3` row was always pushed.

- `calculate_selected_index` applied `skip_focus_for_active_in_search` even when the query carried a row/column, jumping past the file the user was asking to navigate inside.

- The preserved-selection path only invalidated on `path_query()` changes. Appending `:3` to a typed-out path therefore kept the previously skipped-to fuzzy neighbor selected, hiding the fix above when the query was built up character by character.

Self-Review Checklist:

Closes #55551

Release Notes:

- Fixed File Finder offering to create a file instead of navigating to a line when the target file is already open


MrSubidubi (approved):
Aside from the conflicts

crates/file_finder/src/file_finder_tests.rs
```diff
@@ -681,6 +687,48 @@ async fn test_row_column_numbers_query_outside_file(cx: &mut TestAppContext) {
         });
 }
 
+// Regression test for https://github.com/zed-industries/zed/issues/55551.
```
yeskunall, line 884: For myself: do we leave comments like this elsewhere in the codebase? If not, drop it to maintain consistency.

yeskunall:
Thanks for the review Finn! Will fix the conflicts and merge after!

crates/file_finder/src/file_finder_tests.rs
MrSubidubi: Yes, we occasionally do, so not an issue. E.g. https://github.com/zed-industries/zed/blob/3d59fbec28ad84adf85441dce6735991b2f36404/crates/fs/tests/integration/fs_tests.rs#L436 and there are plenty others, search right now yielded 114 results

yeskunall (commented):
Just did another round of manual testing after merging the backfilled commits, and everything still works as expected.

PR #55876, opened 2026-05-06
gpui: Anchor IME candidate window to the start of the visual line

WantenMN:
## Summary

This PR fixes horizontal jumping of the IME candidate window during text composition.

On the shared `selected_bounds` path used by Windows, the candidate window no longer follows the preedit caret character-by-character. Instead, it anchors to the start of the current visual line, which keeps the candidate window stable while typing.

Linux Wayland uses a different IME positioning path, so it was not affected by the original implementation in this PR. This PR now also includes a Wayland-specific adjustment so that Wayland uses the same visual-line anchoring behavior.

## Changes

1. **Shared / Windows path**
   Updated `selected_bounds` in `crates/gpui/src/platform.rs` to use a visual-line-aware anchor for preedit text.

   This removes the distracting horizontal movement of the candidate window while typing and keeps the anchor aligned with the active visual line.

2. **Linux Wayland path**
   Updated the Wayland IME area calculation in `crates/gpui_linux/src/linux/wayland/window.rs` to use the same visual-line-start anchoring strategy for preedit text.

   This brings Linux Wayland in line with the Windows behavior while preserving Wayland's platform-specific IME handling.

## Notes

I had previously tried a separate Wayland-specific fix in 5d0c96872bc3aeed7405ab7bee668b8e77ec7430, but later reverted it in 8cfe7a2a54441f89c170fbdd165a49b3e787b3c4 because the behavior was not good enough and it regressed the original positioning behavior.

The current Wayland implementation is a new, simpler approach that keeps the original behavior intact while also removing horizontal candidate-window jumping.

### Windows / shared path

*   **Before the fix:** (The candidate box moves along with the preedit text, which is distracting)
    
before

before_1

*   **After the fix:** (Candidate box stays fixed at the line start during typing, correctly jumps to new lines or follows active segments)

after

after_1

### Linux Wayland

*  **Before / previous behavior**

before

*  **After / current implementation**

after

---

## Self-Review Checklist:

*   [x] I've reviewed my own diff for quality, security, and reliability
*   [x] Unsafe blocks (if any) have justifying comments
*   [x] The content is consistent with the UI/UX checklist
*   [x] Tests cover the new/changed behavior (manual verification only)
*   [x] Performance impact has been considered and is acceptable

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/platform.rs
+++ b/crates/gpui/src/platform.rs
@@ -1225,42 +1225,55 @@ impl PlatformInputHandler {
         self.handler.replace_text_in_range(None, input, window, cx);
     }
 
-    pub fn selected_bounds(&mut self, window: &mut Window, cx: &mut App) -> Option<Bounds<Pixels>> {
-        let marked_range = self.handler.marked_text_range(window, cx);
-        let selection = self.handler.selected_text_range(true, window, cx)?;
-
+    pub fn compute_ime_candidate_bounds(
+        marked_range: Option<Range<usize>>,
+        selection: &UTF16Selection,
+        mut bounds_for_range: impl FnMut(Range<usize>) -> Option<Bounds<Pixels>>,
+    ) -> Option<Bounds<Pixels>> {
         if let Some(marked_range) = marked_range {
-            let (base_start, caret_offset) =
-                if selection.range.is_empty() && selection.range.end == marked_range.end {
-                    (marked_range.start, selection.range.end)
-                } else {
-                    (selection.range.start, selection.range.end)
-                };
+            // Default to the start of the marked (composing) range.
+            let mut line_start = marked_range.start;
 
-            let mut anchor_offset = base_start;
-            if let Some(caret_bounds) =
-                self.handler
-                    .bounds_for_range(caret_offset..caret_offset, window, cx)
-            {
-                for i in (base_start..caret_offset).rev() {
-                    if let Some(b) = self.handler.bounds_for_range(i..i, window, cx) {
+            // Walk backward from the caret looking for a line break. A change in
+            // the Y coordinate means we crossed into the previous visual line, so
+            // the line start is one position after the break point.
+            let caret = selection.range.end;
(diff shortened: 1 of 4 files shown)


crates/gpui_linux/src/linux/wayland/window.rs
```diff
@@ -942,8 +942,42 @@ impl WaylandWindowStatePtr {
         let mut bounds: Option<Bounds<Pixels>> = None;
         if let Some(mut input_handler) = state.input_handler.take() {
             drop(state);
-            if let Some(selection) = input_handler.marked_text_range() {
-                bounds = input_handler.bounds_for_range(selection.start..selection.start);
+            let marked_range = input_handler.marked_text_range();
+            if let Some(selection) = input_handler.selected_text_range(true) {
```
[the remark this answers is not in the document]
WantenMN: I don't think this is a regression introduced by this change.

On Wayland, the previous `get_ime_area()` implementation also depended on `selected_text_range(true)`, so in alt-screen terminal mode it already returned `None` when
  `selected_text_range()` was unavailable. In other words, this edge case was not handled before, and this patch does not change that behavior.

 So this may still be a separate limitation worth addressing, but it is not newly introduced by this change.

crates/gpui/src/platform.rs
```diff
@@ -1226,16 +1226,41 @@ impl PlatformInputHandler {
...
             } else {
-                selection.range.end..selection.range.end
-            },
-            window,
-            cx,
-        )
+                selection.range.end
+            };
+            self.handler.bounds_for_range(offset..offset, window, cx)
+        }
```
Veykril: Sorry for the late review. Mind adding a comment or two to this logic explaining what is being done here? It's a lot more complicated from what was here before for something that should presumably be very simple in nature

WantenMN: Simplified the logic as requested. Let me know if any further changes are needed.

Veykril (approved):
Thanks!

PR #55903, opened 2026-05-06
terminal_view: Show terminal inline assist keybinding in tooltip

di404:
## Summary

- Fixed the terminal inline assist tab bar tooltip so it resolves the keybinding from the active terminal view.
- The bug happened when the initial terminal was closed and a new terminal was opened: the tab bar button kept a cached focus handle for the old terminal, so tooltip keybinding lookup could no longer find the terminal key context.
- The button is now created while rendering the tab bar with the current terminal view's focus handle, avoiding stale focus handles as terminals are closed and recreated.

## Validation

- `cargo fmt --package terminal_view --check`
- `cargo check -p terminal_view --message-format short`

Release Notes:

- Fixed the terminal inline assist toolbar tooltip not showing its keybinding after reopening terminals.

### DIFF
--- a/crates/terminal_view/src/terminal_panel.rs
+++ b/crates/terminal_view/src/terminal_panel.rs
@@ -11,9 +11,9 @@ use collections::HashMap;
 use db::kvp::KeyValueStore;
 use futures::{channel::oneshot, future::join_all};
 use gpui::{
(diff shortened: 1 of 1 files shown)


smitbarmase (approved):
Thanks! I made two changes on top. Clicking the inline assist button did nothing since we were recreating the button entity on every render, so turned that to `RenderOnce` component. Also added a test.

PR #55923, opened 2026-05-06
Fix repeated path fuzzy ranking

### LINKED ISSUE #55195: fuzzy matching: double hits are not sorted higher

Ella-horizon:
Self-Review Checklist:

Closes #55195

Testing:

- `git diff --check`

Local Rust verification could not be run in this environment because `cargo`, `rustfmt`, and `node` are not installed. The repo commands attempted were:

- `./script/cargo test -p fuzzy_nucleo --lib repeated_path`
- `./script/clippy -p fuzzy_nucleo`

Release Notes:

- Fixed file picker fuzzy ranking so paths with repeated query occurrences rank higher.


Ella-horizon:
@cla-bot check

Ella-horizon:
@cla-bot check

crates/fuzzy_nucleo/src/fuzzy_nucleo.rs
```diff
@@ -57,6 +57,7 @@ impl LengthPenalty {
 // are downranked by a per-mismatch penalty rather than dropped.
 pub(crate) struct Query {
     pub(crate) pattern: Pattern,
+    pub(crate) atoms: Vec<String>,
```
feitreim: You don't need to add this field, you can just get the atoms via `pattern.atoms`.

crates/fuzzy_nucleo/src/paths.rs
```diff
@@ -134,6 +136,18 @@ fn get_filename_match_bonus(
     score as f64 / filename.len().max(1) as f64
 }
 
+#[inline]
+fn get_additional_path_match_bonus(candidate_buf: &str, query: &Query) -> f64 {
+    let candidate_buf = candidate_buf.to_lowercase();
+    let additional_match_count = query
```
feitreim: without the atoms field on Query this would just be `query.pattern.atoms...`

feitreim (commented):
Seems like the reasonable solution, I tried some more complex/convoluted ideas but I think keeping it simple is the way to go. If I have time though I will try to benchmark this impl against mine.

Ella-horizon:
@feitreim 
If you have time...could you review my PR, please?
Thank you!

feitreim:
I had been working on this code myself last week, but then I stepped away for a few days, that said here is a comparison between your code and what I had been working on.

The main difference is that my code provides highlighting for the additional matched copies, so in a path like `zed/crates/zed/Cargo.toml`, both instances of zed are highlighted. This is at the cost of quite a bit more code overall, and Im totally not sure if the complexity is worth the tradeoff. That said you can expand this to see my code for populating the match highlights and counting the extra matches. The benefit of all that complexity is effectively negligible performance wise (other than that it does get more work done), as you can see below as well.

  <details>
  <summary>Code</summary>

  ```rust
  const OCCURRENCE_BONUS: f64 = 2.0;

  /// For each atom, push every occurrence's char positions into `matched_chars`
  /// (so the UI highlights all matches, not just nucleo's single alignment) and
  /// return the total extra occurrences across atoms — i.e. matches beyond the
  /// first. The caller scales this into a score bonus that rewards paths where
  /// query atoms recur (e.g. `zed/crates/zed/Cargo.toml` for query `zed`).
  pub(crate) fn populate_match_highlights(
      haystack: &str,
      pattern: &Pattern,
      matched_chars: &mut Vec<u32>,
  ) -> u32 {
      let mut extras: u32 = 0;
      for atom in &pattern.atoms {
          let count = scan_occurrences(haystack, atom.needle_text(), matched_chars);
          extras += count.saturating_sub(1);
      }
      extras
  }

  fn scan_occurrences(haystack: &str, needle: Utf32Str<'_>, matched_chars: &mut Vec<u32>) -> u32
   {
      if needle.is_empty() {
          return 0;
      }
      match needle {
          Utf32Str::Ascii(bytes) => scan_occurrences_ascii(haystack, bytes, matched_chars),
          Utf32Str::Unicode(chars) => scan_occurrences_unicode(haystack, chars, matched_chars),
      }
  }

  fn scan_occurrences_ascii(haystack: &str, n: &[u8], matched_chars: &mut Vec<u32>) -> u32 {
      let h = haystack.as_bytes();
      if n.len() > h.len() {
          return 0;
      }
      let needle_chars = n.len() as u32;
      // The walker maintains the invariant that `byte_pos` is always at a char
      // boundary: it starts at 0 and only advances by full ASCII bytes, full
      // UTF-8 codepoint widths, or the byte length of a confirmed needle match
      // (which spans `needle_chars` whole chars). So we don't need an explicit
      // `is_char_boundary` check at each step.
      let first_lo = n[0].to_ascii_lowercase();
      let first_hi = n[0].to_ascii_uppercase();

      let mut count: u32 = 0;
      let mut byte_pos: usize = 0;
      let mut char_pos: u32 = 0;
      while byte_pos + n.len() <= h.len() {
          let b = h[byte_pos];
          if (b == first_lo || b == first_hi)
              && h[byte_pos..byte_pos + n.len()].eq_ignore_ascii_case(n)
          {
              count += 1;
              for offset in 0..needle_chars {
                  matched_chars.push(char_pos + offset);
              }
              byte_pos += n.len();
              char_pos += needle_chars;
          } else if b < 0x80 {
              byte_pos += 1;
              char_pos += 1;
          } else {
              byte_pos += haystack[byte_pos..]
                  .chars()
                  .next()
                  .map_or(1, |c| c.len_utf8());
              char_pos += 1;
          }
      count
  }

  fn scan_occurrences_unicode(haystack: &str, needle: &[char], matched_chars: &mut Vec<u32>) ->
  u32 {
      let needle_chars = needle.len() as u32;
      let mut count: u32 = 0;
      let mut byte_pos: usize = 0;
      let mut char_pos: u32 = 0;
      while byte_pos < haystack.len() {
          let mut probe = haystack[byte_pos..].chars();
          let head = probe.next().expect("byte_pos < len ⇒ at least one char");
          let mut all_match = head.eq_ignore_ascii_case(&needle[0]);
          if all_match {
              for &want in &needle[1..] {
                  match probe.next() {
                      Some(c) if c.eq_ignore_ascii_case(&want) => {}
                      _ => {
                          all_match = false;
                          break;
                      }
                  }
              }
          }
          if all_match {
              count += 1;
              for offset in 0..needle_chars {
                  matched_chars.push(char_pos + offset);
              }
              let consumed: usize = haystack[byte_pos..]
                  .chars()
                  .take(needle.len())
                  .map(|c| c.len_utf8())
                  .sum();
              byte_pos += consumed;
              char_pos += needle_chars;
          } else {
              byte_pos += head.len_utf8();
              char_pos += 1;
          }
      }
      count
  }
  ```
 

  Wired into path_match_helper immediately before the sort_unstable / dedup of matched_chars:
``` rust
  let occurrence_extras =
      populate_match_highlights(&candidate_buf, &query.pattern, &mut matched_chars);
  matched_chars.sort_unstable();
  matched_chars.dedup();
  // ...
  let occurrence_bonus = occurrence_extras as f64 * OCCURRENCE_BONUS;
  let positive =
      (score as f64 + filename_bonus + occurrence_bonus) * case_penalty(case_mismatches);

```
</details>

  Benchmarks

  crates/fuzzy_nucleo/benches/match_benchmark.rs, filtered to path.*nucleo, criterion 0.5 with 100 samples per cell. Both branches share the same base commit (dacf984596) so this isolates the bonus implementation. Numbers are criterion's median.

  | bench | this PR | my code | Δ |
  |---|---|---|---|
  | path/1-word/100 | 9.87 µs | 9.24 µs | −6.4% |
  | path/1-word/1000 | 114.9 µs | 108.6 µs | −5.5% |
  | path/1-word/10000 | 1.370 ms | 1.292 ms | −5.7% |
  | path/2-word/100 | 4.07 µs | 3.95 µs | −3.0% |
  | path/2-word/1000 | 33.46 µs | 32.68 µs | −2.3% |
  | path/2-word/10000 | 332.6 µs | 319.4 µs | −4.0% |
  | path/4-word/100 | 1.009 µs | 1.037 µs | +2.7% |
  | path/4-word/1000 | 3.18 µs | 3.17 µs | ~0% |
  | path/4-word/10000 | 20.45 µs | 20.19 µs | −1.3% |
  | path_smart/1-word/10000 | 1.414 ms | 1.325 ms | −6.3% |
  | path_smart/2-word/10000 | 338.8 µs | 329.5 µs | −2.7% |
  | path_smart/4-word/10000 | 20.28 µs | 20.42 µs | +0.7% |
  
  
  I would be interested to see if you could add that same highlighting feature with a simpler mechanism, matching the rest of your code, to provide a better comparison as to which is better. If the simpler one was still within the noise floor performance wise that would probably be the ideal solution :)

Ella-horizon:
@yara-blue  Could you please review my PR?
Thanks!

yara-blue:
@www-dev-1 I think you changed your handle and now the cla-bot doesn't trust you anymore :)

Could you try signing again?

PR #55926, opened 2026-05-06
vim: Keep cursor position when using `gw`

### LINKED ISSUE #32521: vim: Formatting text with gw results in incorrect cursor position
### Summary

Formatting text with `gw` results in incorrect cursor position

Steps to reproduce:
1. Let's take the following sentence:
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam
```

2. The cursor is represented as `[ ]`. If we place the cursor somewhere on the sentence and press `gww`. It should keep the cursor position.
```
Lorem ipsum dolor [s]it amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam
```

3. Now we format the line with `gww`
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
[i]ncididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam
```

4. The cursor moved to the first character of the next line.

**Expected Behavior**: The cursor should keep its position.
```
Lorem ipsum dolor [s]it amet, consectetur adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam
```

```
╰─[I] zed -v && uname -rs
Zed preview 0.190.3 857897ee10263fc46580a0a2e28be37237425a59  – /usr/lib/zed/zed-editor
Linux 6.14.10-arch1-1
```

EDIT:

```
gw{motion}		Format the lines that {motion} moves over.  Similar to
			|gq| but puts the cursor back at the same position in
			the text.  However, 'formatprg' and 'formatexpr' are
			not used.

gwgw							*gwgw* *gww*
gww			Format the current line as with "gw".
```

424ever:
Self-Review Checklist:

- [x] Unsafe blocks (if any) have justifying comments (no unsafe)

Closes #32521

Release Notes:

- Fixed incorrect cursor position after formatting text with gw

### DIFF
--- a/crates/vim/src/normal.rs
+++ b/crates/vim/src/normal.rs
@@ -403,7 +403,9 @@ impl Vim {
                 window,
                 cx,
             ),
-            Some(Operator::Rewrap) => self.rewrap_motion(motion, times, forced_motion, window, cx),
+            Some(Operator::Rewrap { keep_cursor }) => {
+                self.rewrap_motion(motion, times, keep_cursor, forced_motion, window, cx)
+            }
             Some(Operator::Outdent) => self.indent_motion(
                 motion,
                 times,
@@ -510,7 +512,9 @@ impl Vim {
                 Some(Operator::ShellCommand) => {
                     self.shell_command_object(object, around, window, cx);
                 }
-                Some(Operator::Rewrap) => self.rewrap_object(object, around, times, window, cx),
+                Some(Operator::Rewrap { keep_cursor }) => {
+                    self.rewrap_object(object, around, keep_cursor, times, window, cx)
+                }
                 Some(Operator::Lowercase) => {
                     self.convert_object(object, around, ConvertTarget::LowerCase, times, window, cx)
                 }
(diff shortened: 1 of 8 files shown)


crates/vim/src/vim.rs
```diff
@@ -163,6 +163,14 @@ struct PushLiteral {
     prefix: Option<String>,
 }
 
+/// Starts a rewrap operation.
+#[derive(Clone, Deserialize, JsonSchema, PartialEq, Action)]
+#[action(namespace = vim)]
+#[serde(deny_unknown_fields)]
+struct PushRewrap {
+    keep_cursor: bool,
```
yara-blue, line 172: we need to migrate users over. I'd suggest adding a ` #[serde(default)]` so the default for anyone without the argument set is false.

crates/vim/src/normal.rs
```diff
@@ -2120,6 +2124,57 @@ mod test {
             .assert_eq("th th\nth th\nth th\nth th\nth th\nˇth th\n");
     }
 
+    #[gpui::test]
+    async fn test_gw(cx: &mut gpui::TestAppContext) {
```
yara-blue, line 2128: how about rewrapping a long commented line?

That adds a `//` (in the case of rust) on each line. How does that interact with the keep cursor? Does it shift the cursor?

crates/vim/src/rewrap.rs
```diff
@@ -1,16 +1,36 @@
...
+    selection: &Selection<DisplayPoint>,
+) -> MultiBufferOffset {
+    let off = selection.head().to_offset(display_map, Bias::Left);
+
+    // When the selection isn't reversed, the offset works out to be after the head position. It
+    // needs to be adjusted so that the cursor ends up in the same position as it does in neovim.
+    if selection.reversed {
+        off
+    } else {
+        off.saturating_sub_usize(1)
```
yara-blue: What happens if this is in the middle of a utf8 multi byte character? Like 🍀. I think the cursor is off by one then right? Could you add a test for that?

crates/vim/src/normal.rs
yara-blue: Might be worth adding a test for

yara-blue (changes requested):
Overall looks very good. Three concerns blocking them one needs a simple fix the other two additional tests and if those fail they need to be addressed. One of those is easy (utf8 stuff), the other could get tricky (rewrap inserting comment prefixes throwing off the cursor position).

424ever:
I successfully confused myself for a few hours now and will have to come back on the weekend. What I currently don't get is:

1. `th ˇth th th th th` + `v e e` produces the same `th «th thˇ» th th th` state in zed and nvim
2. `th ˇth th th th th` + `v e e g w` produces the same `th th\ntˇh th\nth th` in zed and nvim
3. but `th «th thˇ» th th th` + `g w` suddenly differs ??

yara-blue:
> I successfully confused myself for a few hours now and will have to come back on the weekend

Take all the time you need, this kinda logic is soooo hard to get right. Don't let it drive you insane!

PR #55927, opened 2026-05-06
git_ui: Add force delete for unmerged branches

smitbarmase:
Git's `-d` flag deletes a branch only if it's fully merged into its upstream or HEAD - this is what we were using before, which caused the "not fully merged" error. The `-D` flag force deletes a branch even with unmerged changes (equivalent to `--delete --force`).

### Before

Deleting an unmerged branch failed with a "not fully merged" error toast.

### After

- Deleting an unmerged branch prompts for confirmation to force delete
- Delete button tooltip shows "Hold alt to force delete" hint
- Holding **alt** turns the delete icon red and tooltip changes to "Force Delete Branch"
- Force delete keybinding: `cmd-alt-shift-backspace`

Release Notes:

- Added confirmation prompt when deleting unmerged git branches, with option to force delete.
- Added alt+click on delete button to force delete a branch immediately.

### DIFF
--- a/crates/git_ui/src/branch_picker.rs
+++ b/crates/git_ui/src/branch_picker.rs
@@ -271,7 +271,7 @@ impl BranchList {
         self.picker.update(cx, |picker, cx| {
             picker
                 .delegate
-                .delete_at(picker.delegate.selected_index, window, cx, false)
+                .delete_at(picker.delegate.selected_index, false, window, cx)
         })
     }
 
@@ -284,7 +284,7 @@ impl BranchList {
         self.picker.update(cx, |picker, cx| {
             picker
                 .delegate
-                .delete_at(picker.delegate.selected_index, window, cx, true)
+                .delete_at(picker.delegate.selected_index, true, window, cx)
         })
     }
 
@@ -607,9 +607,9 @@ impl BranchListDelegate {
     fn delete_at(
         &self,
         idx: usize,
+        force: bool,
         window: &mut Window,
         cx: &mut Context<Picker<Self>>,
-        force: bool,
     ) {
         let Some(entry) = self.matches.get(idx).cloned() else {
             return;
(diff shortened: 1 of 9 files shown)


crates/git_ui/src/branch_picker.rs
```diff
@@ -267,7 +271,20 @@ impl BranchList {
         self.picker.update(cx, |picker, cx| {
             picker
                 .delegate
-                .delete_at(picker.delegate.selected_index, window, cx)
+                .delete_at(picker.delegate.selected_index, window, cx, false)
```
Anthony-Eid: nit: false should be passed in before window, cx to be inline with gpui's style

crates/git_ui/src/branch_picker.rs
```diff
@@ -509,7 +604,13 @@ impl BranchListDelegate {
         cx.emit(DismissEvent);
     }
 
-    fn delete_at(&self, idx: usize, window: &mut Window, cx: &mut Context<Picker<Self>>) {
+    fn delete_at(
+        &self,
+        idx: usize,
+        window: &mut Window,
+        cx: &mut Context<Picker<Self>>,
+        force: bool,
```
Anthony-Eid: Same nit as above: force: bool should be added between idx: usize and window: &mut Window

Anthony-Eid (approved):
Overall this is a good change, but product wise I'm unsure if we want a separate action for this. 

IMO it's more mainstream if we only prompt this option after the regular delete branch option fails. Still approving this because the code looks good, and it solves a problem. We can always fix forward if that's what we want todo 😀

aslilac:
this is a rad solution, I really appreciate the care that y'all put into fixing even little things like this when I complain about them 💕

smitbarmase:
> but product wise I'm unsure if we want a separate action for this.

@Anthony-Eid Yeah, I initially had just the confirmation + Alt-click since quick branch deletion is my most common use case. But then I realized keyboard-centric users had no equivalent, so I added the action. I think this covers both camps, and the keybind itself is hard to hit accidentally, which helps.

PR #55930, opened 2026-05-06
Revert "acp: Allow resolving older npm package versions"

agu-z:
Trying to fix forward, but opening this just in case

Reverts zed-industries/zed#55770


agu-z:
Fixed forward by #55938

Issue #55902, opened 2026-05-06
Agent panel UX: copy last response, search inside thread, and jump to any previous prompt


L0GYKAL:
Three closely-related gaps in the Agent panel that all hit the same workflow — extracting / navigating content from a long thread. Filing together because they share a fix surface (thread view).

## 1. One-click copy of the last agent response

Today, copying the agent's last response means selecting the message text manually and `Ctrl/Cmd+C`, which is unreliable in the rendered view (selection drifts across markdown blocks, code fences, tool-call sections, etc.).

**Ask:** A dedicated copy button on each agent message — at minimum on the last one — that copies the rendered message as Markdown to the clipboard. Same affordance most chat UIs already have. Keyboard shortcut + Command Palette entry (e.g. `agent: copy last response`) would close the loop.

## 2. Search inside an agent thread

Long agent threads (multi-hour debugging or refactor sessions) routinely run to thousands of lines. There is currently no way to search within a thread from inside Zed. The only workaround I've found is to export/copy the thread to a scratch file and search there — painful, easy to forget, and breaks any link back to the original message.

**Ask:** A `Ctrl/Cmd+F` search affordance inside the agent panel, scoped to the active thread:
- Match-and-jump on user prompts, agent responses, and tool-call output.
- Ideally honor the same regex / case-sensitivity options as buffer search.
- Highlights and `n/N` navigation, like the rest of Zed.

Bonus: cross-thread search across the workspace's thread history.

## 3. Jump to any previous prompt, not just the last one

The agent panel already has a button/affordance that scrolls to the last user prompt — extremely useful when the response is long. But it stops there: there's no way to jump to an earlier prompt without scrolling manually.

**Ask:** Extend the same button/keystroke to cycle through previous prompts (e.g. repeated press = previous prompt, Shift+press = next prompt), or expose a small "prompts" outline / picker for the active thread. Conceptually similar to a Git log of just-the-user-messages.

## Why bundle these

All three are about treating an agent thread like a *first-class document* — searchable, navigable, extractable — rather than a write-only chat log. They're independently small but compound when you actually live in long threads.

Happy to split into separate issues if the team prefers that for tracking.

PR #55945, opened 2026-05-06
agent_ui: Add Copy Last Response button and cycle through prior prompts

L0GYKAL:
Closes part of #55902.

## Summary

Two small additions to the agent thread toolbar:

1. **Copy Last Response button.** A new icon button (`IconName::Copy`) that copies the most recent assistant message in the active thread to the system clipboard, rendered as Markdown via the existing `AssistantMessage::to_markdown` path. No-op when the thread has no assistant message yet. Sits next to the existing "Open Thread as Markdown" button.

2. **Cycle through prior prompts on the existing scroll-to-recent-prompt button.** The button previously always jumped to the latest user prompt. With this change the first click from anywhere below the latest prompt still lands on it (existing behavior preserved), but subsequent clicks step back to earlier prompts. After the oldest prompt it wraps to the latest. Tooltip updated to reflect the new behavior.

The keyboard actions `ScrollOutputToPreviousMessage` / `ScrollOutputToNextMessage` already exist and are unchanged; this only wires equivalent reverse-stepping into the toolbar button so it's discoverable without a keybind.

## Why

The toolbar already exposes navigation/extraction affordances (open as markdown, scroll to top, scroll to recent prompt). Two of the three asks in #55902 fit that pattern with minimal surface change:

- Copying a long, code-fenced agent response by hand is unreliable across rendered Markdown blocks. A dedicated button removes the manual selection step.
- The existing "jump to last prompt" button is great for navigating *one* level up; cycling lets the same affordance reach any prior prompt without scrolling.

The third ask in #55902 (in-thread search) needs more design and is left for a follow-up PR.

## Test plan

- `cargo check -p agent_ui` is clean.
- Manual:
  - Long thread with multiple user prompts and assistant responses.
  - Click the new copy button → paste into a markdown buffer; the last assistant response (including code fences and `<thinking>` wrappers) round-trips correctly.
  - Click the recent-prompt button repeatedly → scroll position steps from latest prompt to previous prompts and wraps after the oldest.
  - On a thread with zero assistant messages, the copy button is a no-op (does not panic, does not clear clipboard).

## Files

- `crates/agent_ui/src/conversation_view/thread_view.rs`: new `copy_last_assistant_response` method, updated `scroll_to_most_recent_user_prompt` to walk backwards on repeated calls, new toolbar button + updated tooltip.

## Out of scope

- In-thread search (#55902 ②) — separate, larger UI surface.
- Per-message copy buttons on each assistant message (this PR only adds a single "copy last" affordance on the toolbar).
- Keyboard shortcut / Command Palette entry for the copy action (happy to add in a follow-up if desired).

CLA: I'll accept the CLA prompt on this PR.

Release Notes:

- Agent Panel: Added a "Copy Last Response" toolbar button and made the existing "Scroll To Recent User Prompt" button cycle through earlier prompts on repeated clicks.

Release Notes:

- Agent: Added a "Copy Last Response" button and made the existing "Scroll To Recent User Prompt" button cycle backwards through earlier prompts on repeated clicks.

### DIFF
--- a/crates/agent_ui/src/conversation_view/thread_view.rs
+++ b/crates/agent_ui/src/conversation_view/thread_view.rs
@@ -4930,10 +4930,21 @@ impl ThreadView {
                 .shape(ui::IconButtonShape::Square)
                 .icon_size(IconSize::Small)
                 .icon_color(Color::Ignored)
-                .tooltip(Tooltip::text("Scroll To Most Recent User Prompt"))
+                .tooltip(Tooltip::text(
+                    "Scroll To Recent User Prompt (click again for previous)",
+                ))
                 .on_click(cx.listener(move |this, _, _, cx| {
                     this.scroll_to_most_recent_user_prompt(cx);
                 }));
+
+        let copy_last_response = IconButton::new("copy_last_response", IconName::Copy)
+            .shape(ui::IconButtonShape::Square)
+            .icon_size(IconSize::Small)
+            .icon_color(Color::Ignored)
+            .tooltip(Tooltip::text("Copy Last Response"))
+            .on_click(cx.listener(move |this, _, _, cx| {
+                this.copy_last_assistant_response(cx);
+            }));
 
         let scroll_to_top = IconButton::new("scroll_to_top", IconName::ArrowUp)
             .shape(ui::IconButtonShape::Square)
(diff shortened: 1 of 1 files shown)


L0GYKAL:
@cla-bot check

L0GYKAL:
@cla-bot check

PR #55973, opened 2026-05-06
editor: Decode escaped characters in hover popover links

mgood:
Decodes url escape sequences in hover preview `file:///` links like escaped
spaces in the file path.

I'm working on an LSP and happened to be working with some files in a directory with spaces. When adding Markdown links with `file:///` the `%20` escape for spaces was being included verbatim in the path that Zed tried to open.

I'm reusing the lines from `markdown_preview_view.rs` for decoding. In the existing tests I don't see coverage for `file:///` links. If you'd like some tests for this can you point me to any examples to start from?

Self-Review Checklist:

Release Notes:

- Fixed decoding spaces and other escaped characters in `file://` links used in hover popovers

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -798,19 +798,36 @@ pub fn diagnostics_markdown_style(window: &Window, cx: &App) -> MarkdownStyle {
     }
 }
 
+fn parse_file_link(link: &str) -> Option<(PathBuf, Option<String>)> {
+    if let Ok(uri) = Url::parse(link)
+        && uri.scheme() == "file"
+    {
+        let path = uri.to_file_path().ok().unwrap_or_else(|| {
+            let path_text = uri.path();
+            let decoded_url = urlencoding::decode(path_text)
+                .map(|decoded| decoded.into_owned())
+                .unwrap_or_else(|_| path_text.to_string());
+            PathBuf::from(&decoded_url)
+        });
+        let fragment = uri.fragment().map(ToOwned::to_owned);
+        Some((path, fragment))
+    } else {
+        None
+    }
+}
+
 pub fn open_markdown_url(
     workspace: Option<Entity<Workspace>>,
     link: SharedString,
     window: &mut Window,
     cx: &mut App,
 ) {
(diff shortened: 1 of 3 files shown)


mgood:
@cla-bot check

crates/editor/src/hover_popover.rs
```diff
@@ -766,8 +766,12 @@ pub fn open_markdown_url(link: SharedString, window: &mut Window, cx: &mut App)
         && let Some(workspace) = Workspace::for_window(window, cx)
     {
         workspace.update(cx, |workspace, cx| {
+            let path_text = uri.path();
+            let decoded_url = urlencoding::decode(path_text)
+                .map(|decoded| decoded.into_owned())
+                .unwrap_or_else(|_| path_text.to_string());
```
dinocosta: Since `uri` is already a `Url`, we can probably just leverage [`Url::to_file_path`](https://docs.rs/url/latest/url/struct.Url.html#method.to_file_path) while falling back on the `urlencoding::decode` call if it fails .

dinocosta (changes requested):
Hey @mgood ! Thank you for taking the time to improve Zed :)

Change looks good to me but I left a small comment as I believe we can refactor the approach and leverage existing implementations for `Url`. 

Regarding the test, I believe writing a test that exercises the changes in `open_markdown_url` method might be more than the necessary for these changes so I think we could simply extract this logic into a `parse_file_link(link: SharedString) -> Option<PathBuf>` function that we could then provide different links to and check that the resulting `PathBuf` has the `%20` bits correctly decoded . 

Let me know if you need help and sorry it took us so long to get to this!

mgood:
Thanks for the feedback, yeah good call on `uri.to_file_path`, does this look better?

mgood:
(I had to move the fragment parsing to the helper function too since it's needed from the parsed URI later for the line numbers)

dinocosta (approved):
Thank you @mgood ! 🙂

mgood:
Thanks for the more idiomatic update, that looks much cleaner. I'm still trying to find the right incantations to get it to compile most of the time ;)

PR #55997, opened 2026-05-07
git: Normalize submodule gitdir reported as worktree path

plutohan:
`git worktree list --porcelain` from inside a submodule's main checkout reports the submodule's gitdir (`<super>/.git/modules/<sub>`) as the worktree path, not the actual `<super>/<sub>` working tree. `parse_worktrees_from_str` doesn't normalize, so the gitdir lands in `Worktree.path`, and the worktree picker's `SwitchWorktree` action then opens the workspace at the gitdir — a directory containing only git internals.

The fix is in `RealGitRepository::worktrees()`: after parsing, check whether each entry's path equals or sits under the queried repo's `commondir` (which for a submodule's main checkout literally is the gitdir). If so, run `git rev-parse --show-toplevel` from there to get the actual working tree, and set `wt.is_main = true` since `original_repo_path_from_common_dir` returns `None` for the `<gitdir>/modules/<sub>` shape. The commondir comparison is layout-independent and covers `.git/`, `.bare/`, and `--separate-git-dir` setups uniformly. Non-submodule repos pay no extra process spawns.

Tests in `crates/git/src/repository.rs` cover the helper plus an integration test with a real `git submodule add` fixture; without the fix it asserts `left: .../.git/modules/erp / right: .../erp` and passes after.

There's a related bug in `crates/project/src/git_store.rs::resolve_git_worktree_to_main_repo` (line 8123) for the rarer case of `git worktree add` from inside a submodule. Different code path; can do as a follow-up.

Self-Review Checklist:

Release Notes:

- Fixed switching workspaces from a submodule's worktree picker landing in the submodule's gitdir.

### DIFF
--- a/crates/git/src/repository.rs
+++ b/crates/git/src/repository.rs
@@ -328,6 +328,17 @@ impl Worktree {
     }
 }
 
+/// Returns true if `path` is `base` or any descendant of it. Used in
+/// `worktrees()` to detect `git worktree list --porcelain` output that points
+/// at the queried repo's own common-dir — git's bookkeeping convention for a
+/// submodule's main checkout, not a real working tree. Layout-independent:
+/// covers `<super>/.git/modules/<sub>`, `<super>/.bare/modules/<sub>`, and
+/// any future layout where a submodule's gitdir lives inside the
+/// superproject's git directory.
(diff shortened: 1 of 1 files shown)


plutohan:
@cla-bot check

PR #56010, opened 2026-05-07
editor:  Extract `fold`, `selection`, `coade_actions` and `completions` out of `editor.rs`

GoldStrikeArch:
cc @SomeoneToIgnore

## Summary

This extracts several editor topics into focused sibling modules:

- `crates/editor/src/fold.rs`
- `crates/editor/src/code_actions.rs`
- `crates/editor/src/completions.rs`
- `crates/editor/src/selection.rs`

While doing the extraction, this also tightens the module boundaries instead of preserving every old access path by inertia:

- Avoids exposing the raw `context_menu` storage and replaces external usage with focused APIs for code actions and completions state.
- Keeps broad editor editing/cancel behavior in `editor.rs`, since those APIs are not selection-specific.
- Moves linked-editing range lookup into `linked_editing_ranges.rs`, closer to its only real caller.

Self-Review Checklist:

Release Notes:

- N/A


SomeoneToIgnore (changes requested):
Image

Hard "nope", sorry.

One of the points is to keep things mechanical enough to review quickly and this is not the case now.
Also, why there's so many `+` in that particular file, seems odd?

GoldStrikeArch:
> Also, why there's so many + in that particular file, seems odd?

Yeah, this is weird. It should be -4k lines. I guess there is some re-ordering I introduced. I will try to fix it

GoldStrikeArch:
@SomeoneToIgnore 

Looks like something is wrong with the diffing here from the Github/git because on my branch `git diff --minimal` showed `12 insertions(+), 4610 deletions(-)`

I can't solve it for some reason, so I decided to break them up into 2 sequential PRs, this is the first one -> https://github.com/zed-industries/zed/pull/56030

PR #56011, opened 2026-05-07
Support LSP document links

SomeoneToIgnore:
Closes https://github.com/zed-industries/zed/issues/33587

https://github.com/user-attachments/assets/bbaea8a9-402e-485b-800e-2f4486142956

Release Notes:

- Supported LSP document links (enabled by default, use `"lsp_document_links": false` to turn it off)


Veykril (approved):

aviatesk:
Dynamic registration seems to need to be fixed. Opened a follow up PR: https://github.com/zed-industries/zed/pull/57749

aviatesk:
One more follow-up I wonder about: should we render document link ranges as underlined by default, similar to how VS Code does?

Right now, `textDocument/documentLink` seems to be surfaced mostly like `textDocument/definition` and it becomes visible when hovering or command-clicking a position. But `documentLink` is document-scoped rather than position-scoped. So unlike `textDocument/definition`, the server can return all jumpable link ranges for the document in one request, and each `DocumentLink` already includes the range it applies to. In that sense it feels closer to `documentColor`, diagnostics, or code lenses, and once we have the document-level data, surfacing it directly in the editor improves discoverability.

Would you be open to that as a follow-up?

SomeoneToIgnore:
I'd ask a designer for that, myself is not very open to that as it adds more visual noise than needed.

aviatesk:
Understood. I think the noise argument has a validity. Since I only mentioned it as a tentative suggestion, I will hold off on upstreaming this idea for now.

PR #56030, opened 2026-05-07
editor: Extract `completions` and `code_actions` out of `editor.rs`

GoldStrikeArch:
cc @SomeoneToIgnore

## Summary

- `crates/editor/src/code_actions.rs`
- `crates/editor/src/completions.rs`

One odd boundary remains: `Editor::context_menu()` is still a general context-menu accessor, but it now lives in `code_actions.rs` because it was part of the moved code actions block and is also used by completions, Vim tests, agent UI, and the quick action bar. Would you prefer that generic context-menu accessor stay in `editor.rs` for now until context-menu code gets its own extraction?

## Testing

- `cargo check -p editor --lib`
- `cargo check -p editor --tests`
- `cargo check -p editor --lib --features test-support`

Self-Review Checklist:

Release Notes:

- N/A


crates/editor/src/code_actions.rs
```diff
@@ -0,0 +1,527 @@
...
+            false
+        }
+    }
+
+    pub fn context_menu(&self) -> &RefCell<Option<CodeContextMenu>> {
+        &self.context_menu
+    }
+}
+
+impl Editor {
```
SomeoneToIgnore: We could squash all into a single `impl Editor` here and maybe in other related places?

crates/editor/src/completions.rs
```diff
@@ -0,0 +1,1495 @@
...
+        if let CodeContextMenu::Completions(menu) = menu.as_ref()? {
+            let completions = menu.completions.borrow();
+            Some(completions.to_vec())
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(test)]
```
SomeoneToIgnore: That can be unified with the one above.
This seems to be a good example where another `impl Editor` works, but the rest could be just merged?
And, even when split, would be nice to group together without any extra functions between.

crates/editor/src/code_actions.rs
GoldStrikeArch: Done. Moved everything to a single `impl Editor`

crates/editor/src/completions.rs
GoldStrikeArch: Done. Moved to the above `impl Editor`

SomeoneToIgnore (changes requested):
`impl Editor` are not grouped still; `fn` and `pub(super) fn` and maybe others seem to be mixed, not ordered.

GoldStrikeArch:
@SomeoneToIgnore 

> `impl Editor` are not grouped still;

I see. I misunderstood your point, now with the latest changes this:

```rs
impl Editor {
// ...
}
``` 
Is next to the "main"`impl Editor` in `completions.rs`. Is this what you were referring to? 

> fn and pub(super) fn and maybe others seem to be mixed, not ordered

Right now both "main" `impl Editor` in `code_actions.rs` and `completions.rs` are ordered like this:
- pub fn
- fn
- pub(super) fn

I tried to use the same pattern as in `config.rs` which we did yesterday. Do you want some other order?

SomeoneToIgnore:
`fn` is more private than its surroundings in this case, so the sequence seems odd.
I'd put `fn` last.

GoldStrikeArch:
@SomeoneToIgnore 

> `fn` is more private than its surroundings in this case, so the sequence seems odd.
I'd put `fn` last.

Agreed. I moved the `fn` to the bottom

SomeoneToIgnore (approved):
Thank you.

PR #56033, opened 2026-05-07
settings_ui: Add `Clear` button to the search field

asgoode:
This PR adds a 'clear' button to the search field in the Settings window. The button becomes visible only when the field contains text. Clicking the button clears the search field, which also resets the navigation list.

Empty:
image

With a search term:
image

There are similar UIs, like the macOS and Windows System Settings windows, and even Zed's 'Help' menu filter (on macOS), where the field acts as a 'live filter', updating results on each keystroke, and the 'clear' action is more convenient than backspacing.

The first commit adds the button and the 'on click' functionality. If the Zed team approves this change, I would also like to use the `escape` key to clear the field. I have done some work on this, but there are several issues and I would like some feedback before committing the appropriate changes:
1. The `escape` binding to clear the field could arguably be built-in (not user-configurable) since it's the standard key for this type of action, and it may not make sense to allow other key combinations since this field accepts keyboard input. Would the team approve a built-in key binding, or must it be bound to a user-configurable `Action`?
2. For the Settings Window, `escape` is currently bound to the `CloseWindow` Action by default. If the search field is focused, pressing the `escape` key should clear the field, but not close the window.
   - If a new `Action` is used to clear the field (containing text), the binding to that `Action` can have a more specific context (`SettingsWindow > NavigationMenu && search`) than the `Action` on the window (`SettingsWindow > NavigationMenu`), so the window won't close … but if the field is *already empty*, should `escape` close the window (and is there a way to re-enable the `CloseWindow` action in that case)?
   - If a `KeyDownEvent` listener for `escape` is used to clear the field, the `CloseWindow` Action still closes the window – is there a way to cancel the `CloseWindow` Action in this case?
3. There is at least one other location with a similar 'live filter' field, namely the Filter field in the Keymap Editor, where this same functionality could be implemented. (I could update that too if this PR is approved). Are there other similar instances in Zed? This 'clear filter' UX should be consistent in all those locations.

Self-Review Checklist:

(I could not find an existing issue or discussion about this feature.)

Release Notes:

- Added a 'clear' button to the search field in the Settings window, which clears the field and resets the navigation list.


crates/settings_ui/src/settings_ui.rs
```diff
@@ -2573,10 +2581,14 @@ impl SettingsWindow {
     //     }
     // }
 
-    fn render_search(&self, _window: &mut Window, cx: &mut App) -> Div {
+    fn render_search(&self, _window: &mut Window, cx: &mut Context<SettingsWindow>) -> Div {
```
asgoode: Changed the `cx` param from `&mut App` to `&mut Context<SettingsWindow>` because the `cx` arg in `on_click` requires that type, and I wasn't sure if it could be obtained another way. Also, the other `render_*` methods on `SettingsWindow` use the  `&mut Context<SettingsWindow>` type for `cx`, so this seems more consistent.

MrSubidubi (approved):
Sorry for the long delay, this looks good to me, thanks! And congrats to your first contribution 🎉 

If you are still interested despite the suboptimal experience here, I also think the direction described here sounds good, so do feel free to open up a follow-up if you feel like it!

PR #56044, opened 2026-05-07
agent: Fix race-condition for LSP tool registration

bennetbo:
This fixes a race condition where the thread would not get the LSP tools at startup if the feature flag was not resolved yet.
We now always add the tools, but filter them out when we start a new turn if the feature flag is not set. 

Release Notes:

- N/A

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -2897,8 +2897,7 @@ impl Thread {
                 cx.has_flag::<LspToolFeatureFlag>()
                     || !matches!(
                         tool_name.as_ref(),
-                        DiagnosticsTool::NAME
-                            | FindReferencesTool::NAME
+                        FindReferencesTool::NAME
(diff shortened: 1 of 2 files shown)


crates/agent/src/thread.rs
```diff
@@ -2894,6 +2893,18 @@ impl Thread {
                     None
                 }
             })
+            .filter(|(tool_name, _)| {
+                cx.has_flag::<LspToolFeatureFlag>()
+                    || !matches!(
+                        tool_name.as_ref(),
+                        DiagnosticsTool::NAME
```
cameron1024: I don't think we wanna filter out the diagnostic tool

cameron1024 (approved):
Looks good, other than the diagnostic tool issue

crates/agent/src/thread.rs
bennetbo: Whoops good catch thank you

PR #51097, opened 2026-03-09
accesskit

cameron1024:
rudimentary accessibility support. Unix-only for now

PR #56065, opened 2026-05-07
gpui: Accesskit support

cameron1024:
GPUI AccessKit integration

This PR is replacing #51097 , and is much more limited in scope. This PR *ONLY* adds AccessKit support to GPUI, and doesn't touch Zed. Once this lands, we can start adding aria attributes to Zed's components.

This PR is the first step to addressing #41138 .

Release Notes:


sportfloh:
I tried to demo app on macOS 26.5 with VoiceOver.
When I start the app only the top bar is reachable via VoiceOver keyboard navigation. If I switch to some other app and then back to the demo app , I'm able to navigate further. I'm able to jump to the window content area and the into the content area. All labels and controls are accessible.

For context: I'm visually impaired, so I have some eyesight left to confirm what labels/controls are on the screen and witch labels/controls get spoken.

Let me know if I can test anything else.
Thank you for your work!
Chears

cameron1024:
Thanks for testing @sportfloh :grin: 

I hit the same issue on Windows with NVDA, and pushed a fix that I think addressed part of the cause, but didn't fully solve it on Windows. 

I've just tested it with @agu-z and it seems to be working. Here's what we observed:
- start VoiceOver
- run `cargo run -p gpui --example a11y`
- VoiceOver draws a box around the entire window
- It's possible to navigate within the window to individual elements and interact with them. VoiceOver boxes are then drawn around individual elements, and I can hear individual elements being announced.

Based on that, it seems to be working on our end, but please feel free to complain if it's not working for you :sweat_smile: .

I'd add that I wouldn't consider this issue to be a blocker for merging, since the PR is already a significant improvement in usability for users of assistive technology, and it sounds like there's a reliable workaround.

cameron1024:
Progress update:

- After testing MacOS with VoiceOver, it seems to be working well
- Windows seems to work "perfectly" (to my knowledge) under NVDA. Haven't tested JAWS
- Wayland: I had trouble setting up Orca on my Nixos/Niri machine, heaven forbid I make things easy for myself. That said, the nodes are visible, and interactable via AT-SPI, so I feel relatively confident about this
- X11: I don't have an X11 machine, so instead set up a Nixos integration test for this. AT-SPI is working here too

sportfloh:
@cameron1024 thx for the fix, works now for me too! :-)

crates/gpui/examples/a11y.rs
```diff
@@ -0,0 +1,260 @@
...
+                            .map(|(i, label)| {
+                                div()
+                                    .id(("task", i))
+                                    .role(Role::ListItem)
+                                    .aria_label(SharedString::from(*label))
+                                    .aria_position_in_set(i + 1)
+                                    .aria_size_of_set(3)
+                                    .py_1()
+                                    .px_2()
+                                    .child(text!(format!("{}. {}", i + 1, label)))
```
agu-z, line 218: Should we add a comment here clarifying that this is ok because each parent div provides a globally unique id?

crates/gpui/src/elements/div.rs
```diff
@@ -1170,6 +1177,141 @@ pub trait InteractiveElement: Sized {
         self.interactivity().focus_visible_style = Some(Box::new(f(StyleRefinement::default())));
         self
     }
+
+    /// Set the accessible role for this element.
+    ///
+    /// See the [accessibility guide](crate::_accessibility) for an overview.
+    fn role(mut self, role: accesskit::Role) -> Self {
```
agu-z, line 1181: Could these move to `StatefulInteractiveElement` instead? That way it would fail to compile instead of debug-panic if the `div` doesn't have an id.

crates/gpui/src/window.rs
```diff
@@ -5296,6 +5350,89 @@ impl Window {
...
+            }
+            accesskit::Action::Focus => {
+                if let Some(focus_id) = self.a11y.focus_ids.get(&request.target_node).copied() {
+                    if self.focus_enabled {
+                        self.focus = Some(focus_id);
+                        self.clear_pending_keystrokes();
+                        self.refresh();
+                    }
+                }
+            }
```
agu-z: Is there a reason this doesn't go through `Window::focus`? Something like:

 ```rust
 accesskit::Action::Focus => {
     if let Some(focus_id) = self.a11y.focus_ids.get(&request.target_node).copied()
         && let Some(handle) = FocusHandle::for_id(focus_id, &cx.focus_handles)
     {
         self.focus(&handle, cx);
     }
 }
 ```

 Setting `self.focus` directly skips the deferred observer notification in `Window::focus`, so focus subscribers wouldn't see AT-driven focus changes.

crates/gpui_linux/src/linux/x11/window.rs
```diff
@@ -1886,4 +1891,58 @@ impl PlatformWindow for X11Window {
...
+
+    fn a11y_tree_update(&self, tree_update: accesskit::TreeUpdate) {
+        let mut state = self.0.state.borrow_mut();
+        if let Some(adapter) = state.accesskit_adapter.as_mut() {
+            adapter.update_if_active(|| tree_update);
+        }
+    }
+
+    fn a11y_update_window_bounds(&self) {
+        // X11 could report window bounds, but for now it's a no-op
```
agu-z: Are you considering this out-of-scope or was it forgotten?

crates/gpui/src/_accessibility.rs
```diff
@@ -0,0 +1,226 @@
...
+//! let todos = vec!["eat lunch", "drink water", "go to gym"];
+//! let todo_divs = todos.into_iter().enumerate().map(|(index, todo)| {
+//!     text!(todo).with_id(index)  // OR `text(id = index, todo)`
+//! });
+//!
+//! div()
+//!     .id("todo-list")
+//!     .role(Role::Document)
+//!     .children(todo_divs);
+//! ```
```
agu-z, line 165: I wonder if we should have an example here of the case where it's actually fine to use `text!` because the global id is unique via the parent.

crates/gpui_linux/src/linux/x11/window.rs
cameron1024: Forgotten, shouldn't be too hard

crates/gpui/src/window.rs
cameron1024: Ah yeah, this was on my list of things to fix, but slipped my mind

crates/gpui/src/elements/div.rs
cameron1024: I'm actually thinking we could do this for *every* aria method...

agu-z (approved):
Thank you so much for your work on this!

Almost-Senseless-Coder:
Finally had the time to test this on my laptop running Windows 11 with `cargo run -p gpui --example a11y`. Observed behaviours:

#### NVDA 2026.1.1

The example app opens, the window title is spoken, and otherwise the window remains inaccessible until focus gets moved away from the window and then returned to it using e.g. ALT + TAB. Afterwards, a toggle button, a numerical button that increases a counter when clicked and an ordinary button that resets the counter value get detected; cycling through tab order using the TAB key works as expected. Arrow keys don't do a.nything, but I suspect that's expected behaviour.

#### Narrator

The behaviour is identical to NVDA, including the necessity to switch focus away and back again once.

#### JAWS 2025

The behaviour is identical to NVDA except the app controls are accessible right away, although no control initially receives focus.

#### Conclusion

I suspect the slight accessibility issues with NVDA and Narrator have less to do with gpui's integration of AccessKit and more with the specific implementation of the demo app. As such, I think it's highly likely the AccessKit integration was a success.

I'll test further on Linux with X11, will report back once that's done.

PR #56070, opened 2026-05-07
editor: Extract `fold` and `selection` out of `editor.rs`

GoldStrikeArch:
cc @SomeoneToIgnore

## Summary

Follow-up to #56030 

- `crates/editor/src/fold.rs`
- `crates/editor/src/selection.rs`

One odd boundary remains: several selection state types still live in `editor.rs`. I didn't move them because those caused that "huge 11k diff" in the previous PR, so I propose to move them later.

Self-Review Checklist:

Release Notes:

- N/A


SomeoneToIgnore (approved):
Thanks!

PR #56080, opened 2026-05-07
remote: Fix incorrect arguments passed to SCP & SFTP

cocool97:
All arguments passed to the initial remote connection command were initially blindly passed to `scp` or `sftp` command line.
This is an issue with some arguments like "-X" or "-Y" that either does not exist, or have a different semantic.
This PR introduces a denylist of arguments before crafting the command to execute.

Reproduce steps:

1. Create a new remote connection, e.g `ssh -X user@192.168.1.1`
2. Expect the upload of the remote server to fail with a message saying that '-X' is not a valid argument for `ssh` / `sftp` depending on your setup.

Self-Review Checklist:

Release Notes:

- Fixed wrong arguments that may be passed to scp or sftp command when pushing remote files.


cocool97:
@cla-bot check

crates/remote/src/transport/ssh.rs
```diff
@@ -1148,17 +1148,22 @@ impl SshRemoteConnection {
         &self,
         src_path: &Path,
         dest_path_str: &str,
-        args: Option<&[&str]>,
+        additional_args: Option<&[&str]>,
     ) -> util::command::Command {
+        // these arguments exist for "ssh" but don't exist / don't have the same semantic for "scp"
```
maxdeviant: It would be better to use a doc comment here so that it attaches itself to the `const`:

```suggestion
        /// These arguments exist for `ssh` but don't exist / don't have the same semantic for `scp`.
```

Veykril (approved):
Thanks!

PR #56087, opened 2026-05-07
Auto watch improvements

JosephTLyons:
Just a few tweaks to [Auto Watch](https://github.com/zed-industries/zed/pull/54839).

- Swapped the positions of the Copy Channel Link button and the Auto Watch button. The UI still needs a design pass, but I think it’s less awkward now that the Auto Watch button is no longer dangling by itself when engaged.

   Before:

   SCR-20260507-obsu

   After:

   SCR-20260507-obzs

- Turned Auto Watch off when following another collaborator. Following is already disengaged when Auto Watch is enabled, and now Auto Watch is disengaged when following starts. This makes the two modes fully mutually exclusive, which I think is what we want.

Self-Review Checklist:

Release Notes:

- N/A

### DIFF
--- a/crates/collab/tests/integration/auto_watch_tests.rs
+++ b/crates/collab/tests/integration/auto_watch_tests.rs
@@ -228,7 +228,7 @@ async fn test_auto_watch_toggle_off_leaves_tabs_open(
 }
 
 #[gpui::test]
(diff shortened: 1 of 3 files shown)


JosephTLyons:
Superseded by: https://github.com/zed-industries/zed/pull/56126

PR #56105, opened 2026-05-07
collab: Route `get_users_by_ids` through Cloud

maxdeviant:
This PR makes it so we route the `UserService::get_users_by_ids` call through Cloud instead of hitting the database.

We've introduced a new `CloudUserService` that will fetch the users from Cloud using the internal API. Note that we've only implemented the `get_users_by_ids` method on this service, as the endpoints for the other methods don't yet exist.

We have also introduced a `TransitionalUserService` for the purposes of gradually transitioning these calls over to Cloud. Right now it uses the `CloudUserService` for the `get_users_by_ids` implementation, but then uses the `DatabaseUserService` for the other methods.

Closes CLO-740.

Release Notes:

- N/A

### DIFF
--- a/crates/cloud_api_types/src/cloud_api_types.rs
+++ b/crates/cloud_api_types/src/cloud_api_types.rs
@@ -1,4 +1,5 @@
 mod extension;
+pub mod internal_api;
 mod known_or_unknown;
 mod plan;
 mod timestamp;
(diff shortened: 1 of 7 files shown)


tomhoule (approved):
Nice!

NeelChotai (approved):

PR #56119, opened 2026-05-08
git_ui: Add Git Panel entry click behavior setting

matthewgrossman:
## Summary
- add a `git_panel.entry_click_behavior` setting so Git Panel clicks can either open the diff view or open the file directly
- keep the existing default behavior by leaving `diff` as the default, while allowing users to opt into `open`
- document the setting and its `diff`/`open` meanings in the settings schema and defaults

I often wish I could just click the file without having to right-click + open, or deal with the somewhat flaky double-click. I think letting users choose what click does by default seems reasonable? If not, happy to open an issue first where we can discuss it. First time contributor; sorry if I missed anything!

EDIT: Looking more at the double-click, I wonder if this should also toggle what double click does? meaning if we set single-click to `open`, then we should set double-click to `diff`?

## Related
- Related to #51712

## Testing
- `cargo fmt --all`
- Could not run targeted `cargo test -p git_ui ...` in this environment because Cargo dependency fetches require GitHub SSH authentication

Release Notes:

- Improved Git Panel settings by adding a click behavior option for opening diffs or files directly.

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -962,9 +962,16 @@
     //
     // Default: false
     "collapse_untracked_diff": false,
-    /// Whether to show entries with tree or flat view in the panel
-    ///
-    /// Default: false
+    // What clicking a file entry in the Git panel should do.
+    // `diff` opens the Git diff view for that file.
+    // `open` opens the file directly in the editor.
+    //
+    // Choices: diff, open
+    // Default: diff
+    "entry_click_behavior": "diff",
+    // Whether to show entries with tree or flat view in the panel
+    //
+    // Default: false
     "tree_view": false,
     // Whether the git panel should open on startup.
     //
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -61,7 +61,7 @@ use project::{
 use prompt_store::{BuiltInPrompt, PromptId, PromptStore, RULES_FILE_NAMES};
 use proto::RpcError;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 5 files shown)


matthewgrossman:
@cla-bot check

ChristopherBiscardi:
Thank you for the PR!

Unfortunately I don't think we want this to be a setting, since there are already options for opening the diff or file in three places, and the context here is the git panel rather than the project panel, so it makes sense for the primary action to be git related:

- click vs double click
- context menu
- keyboard (enter vs enter + ctrl)

This also doesn't solve #51712 which is a focus/navigational issue.

matthewgrossman:
Appreciate the feedback! I agree w/ the perspective that the git panel should trigger git-related actions. However, I believe there’s a distinction between Navigation (editing) and Reviewing.

In tools like magit and vim-fugitive, the primary action (Enter) on a file name navigates the user directly to that file. This supports the "editing phase," where the Git panel serves as a filtered list of files currently being "juggled" for a commit.

The git-diff multibuffer, while great for a final audit, feels like a "Reviewing phase" tool. Forcing this view on a single click introduces friction when the user’s intent is simply to jump into a specific file to continue working.

By making a single click open the file itself (with diffs enabled), we support the high-frequency "edit" workflow while still keeping the action contextually relevant to Git. It boils down to which workflow to prioritize as the default:

    Single Click/Enter: Focused navigation for active development (The Edit phase).
    Secondary Action: Project-wide hunk review (The Review phase).

In my experience, the Edit phase is the more frequent touchpoint, so I'd prefer the git panel better oriented around that. 

Thanks!

PR #56132, opened 2026-05-08
gpui: Remove unsound await_on_background helper

Veykril:
The function is unsound due to the classic fact that one can leak tasks, sidestepping the blocking drop behavior resulting in a use after free.

Release Notes:

### DIFF
--- a/crates/language/src/language.rs
+++ b/crates/language/src/language.rs
@@ -623,8 +623,8 @@ pub trait LspInstaller {
         &self,
         _version: &Self::BinaryVersion,
         _container_dir: &PathBuf,
-        _delegate: &dyn LspAdapterDelegate,
-    ) -> impl Send + Future<Output = Option<LanguageServerBinary>> {
+        _delegate: &Arc<dyn LspAdapterDelegate>,
+    ) -> impl Send + Future<Output = Option<LanguageServerBinary>> + use<Self> {
         async { None }
     }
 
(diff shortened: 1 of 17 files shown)


Veykril (commented):
The language crate changes are merely switching out the async model to be impl Future where the Future is effectively `'static` so we can offload it to the background.

crates/languages/src/bash.rs
```diff
@@ -90,35 +90,41 @@ impl LspInstaller for BashLspAdapter {
...
-            .node
-            .should_install_npm_package(
-                Self::PACKAGE_NAME,
-                &server_path,
-                container_dir,
-                VersionStrategy::Latest(version),
-            )
-            .await;
+        delegate: &Arc<dyn LspAdapterDelegate>,
+    ) -> impl Send + Future<Output = Option<lsp::LanguageServerBinary>> + use<> {
```
SomeoneToIgnore, line 98: I wonder if some generic type alias is ok to have for this?
Maybe not in `gpui` or other public place, but some utils/runtime crate?

crates/language/src/language.rs
```diff
@@ -1421,13 +1413,15 @@ impl LspInstaller for FakeLspAdapter {
...
-    async fn fetch_server_binary(
+    fn fetch_server_binary(
         &self,
         _: (),
         _: PathBuf,
-        _: &dyn LspAdapterDelegate,
-    ) -> Result<LanguageServerBinary> {
-        unreachable!();
+        _: &Arc<dyn LspAdapterDelegate>,
+    ) -> impl Future<Output = Result<LanguageServerBinary>> + use<> {
```
SomeoneToIgnore: NIT: this method is the only `use<>` one that does not have `Send` bound in case it matters.

SomeoneToIgnore (approved):
Would be nice to

* get some elaboration on unsoundness in the PR description as it's always interesting
* have a way to avoid repeating `impl Future<Output = T> + use<>` and use some shorter form

but nothing blocking.

crates/languages/src/bash.rs
Veykril: would need to be a trait alias which is not a stable thing yet

SomeoneToIgnore: Or `impl Foo for Send + ...` and require that instead?
Either way, not too big of a deal.

PR #56139, opened 2026-05-08
gpui: Use `SharedString::new_static` within `From` impls for `ElementId` where possible

MrSubidubi:
Horror of a PR title but could not think of anything better here.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui/src/window.rs
+++ b/crates/gpui/src/window.rs
@@ -5801,7 +5801,7 @@ impl From<Uuid> for ElementId {
 
 impl From<(&'static str, u32)> for ElementId {
     fn from((name, id): (&'static str, u32)) -> Self {
-        ElementId::NamedInteger(SharedString::new_static(name), id as u64)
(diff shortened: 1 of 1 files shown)


crates/gpui/src/window.rs
```diff
@@ -5806,7 +5801,7 @@ impl From<Uuid> for ElementId {
 
 impl From<(&'static str, u32)> for ElementId {
     fn from((name, id): (&'static str, u32)) -> Self {
-        ElementId::NamedInteger(name.into(), id.into())
+        ElementId::NamedInteger(SharedString::new_static(name), id as u64)
```
SomeoneToIgnore:
```suggestion
        ElementId::NamedInteger(SharedString::new_static(name), u64::from(id))
```

NIT: This way we're almost 100% sure nothing is lost on conversion implicitly.

SomeoneToIgnore (approved):

PR #56154, opened 2026-05-08
agent: Use simpler schema for edit tool

bennetbo:
This makes it so that the agent can only specify a single edit. In practice the agent was not specifying multiple edits in the same tool call often. Also the models still seem to struggle with JSON escaping inside an array, so simplifying the schema improves performance (basically brings down the invalid JSON failure modes to zero)

The schema now is just:

```json
{
  "path": "somepath",
  "old_text": "old",
  "new_text": "new"
}
```

For now we are keeping `EditSession` as is, in case we want to go back to letting the model specify multiple edits.

Release Notes:

- N/A


bennetbo:
Closing for now as there are other issues with this format. We want to experiment further with other formats

PR #56155, opened 2026-05-08
editor: Extract `git` and `input` out of `editor.rs`

GoldStrikeArch:
cc @SomeoneToIgnore

## Summary

- `crates/editor/src/input.rs`
- `crates/editor/src/git.rs`

The git extraction is intentionally partial for now. I left a lot of related parts because otherwise the diff was super huge (over 9K lines) in the Github, so we can move those parts later in the follow-up PRs
Self-Review Checklist:

Release Notes:

- N/A


SomeoneToIgnore (approved):
Thanks!

PR #56157, opened 2026-05-08
Do not send "finished" agent notification when about to send a queued message

SomeoneToIgnore:
I wait for an agent it takes time/thinks, but I notice some odd error in its work that I want to correct: I write a message, knowing that e.g. Claude can read that later at some point and continue normally, and focus away from Zed.

At some point, I get the notification I need to react to, only to understand that the generation is ongoing still due to that delayed message sent.

Release Notes:

- Fixed agent "finished" notification appearing right before the queued message is sent


benbrandt (approved):
thanks! ❤️

PR #56168, opened 2026-05-08
gpui: Defer platform callbacks while app is borrowed

agu-z:
This is just meant to show a possible direction to fix these reentrancy crashes, not ready to merge.

### Summary

Fixes crashes caused by platform callbacks reentering GPUI while the app is already borrowed.

On Windows, reading from the clipboard can reenter our window procedure. In Sentry, this happened while GPUI was rendering: `GetClipboardData` reentered wndproc, delivered `WM_DESTROY`, and ran the window close callback. That callback tried to update GPUI while the app was already borrowed, causing `RefCell already borrowed`.

This adds `AsyncApp::update_or_defer`, which runs an app update immediately when possible and queues it when the app is currently borrowed. Void platform/window callbacks that mutate GPUI now use this path.

### Self-Review Checklist

### Release Notes:

- Fixed crashes caused by platform callbacks reentering GPUI while the app is already updating.

### DIFF
--- a/crates/gpui/src/app.rs
+++ b/crates/gpui/src/app.rs
@@ -70,11 +70,14 @@ mod visual_test_context;
 /// The duration for which futures returned from [Context::on_app_quit] can run before the application fully quits.
 pub const SHUTDOWN_TIMEOUT: Duration = Duration::from_millis(100);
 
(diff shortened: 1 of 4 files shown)


agu-z:
We should probably make the clipboard API async

PR #53465, opened 2026-04-09
markdown: Improve table cell alignment

davidalecrim1:
## Summary

Markdown preview tables kept text pinned to the top of a row when a neighboring cell contained a taller image. This made mixed text-and-image tables look unbalanced and inconsistent with common editor behavior. This change makes table text stay visually centered within taller rows so Markdown tables are easier to scan and match expected rendering more closely.

## Before / After

| Before | After |
| --- | --- |
| Screenshot 2026-04-08 at 19 55 50 | Screenshot 2026-04-08 at 21 47 31 |

## References
Inspired by comparing this with VS Code preview
Screenshot 2026-04-08 at 21 54 06

Release Notes:

- Improved Markdown preview table cells to vertically center content in tall rows and respect column alignment from the table header.

PR #56179, opened 2026-05-08
markdown: Fix table header alignment and apply alignment to cell content

smitbarmase:
Follow up: https://github.com/zed-industries/zed/pull/53465

For Markdown tables, headers are now always centered (ignoring column alignment), matching standard Markdown rendering behavior. For HTML tables, headers default to center but respect explicit `align` attributes.

This also propagates alignment to paragraphs and headings inside table cells, not just the cell container itself.

Release Notes:

- N/A


agu-z (approved):
nice

PR #56185, opened 2026-05-08
client: Rename `UserId` to `LegacyUserId`

maxdeviant:
This PR renames the `UserId` type in the `client` crate to `LegacyUserId`.

The `id` field on the `User` has also been renamed to `legacy_id`.

This is strictly a rename, no change in behavior.

Release Notes:

- N/A


ConradIrwin:
@zed-zippy approved

PR #56188, opened 2026-05-08
workspace: Open new windows on the focused display

haxhibekaj:
## summary

when you have zed fullscreen on a secondary display and hit cmd-shift-n, the new window pops up on your primary display instead of the actively focused display.

the cause is that `Workspace::new_local`'s new-window path only consults the last-saved default bounds when picking a display. it never looks at where the user currently is.

this change adds an "active display" branch into the priority chain, right after the serialized-workspace branch:

1. `ZED_WINDOW_BOUNDS` env override
2. saved bounds for a serialized workspace being reopened
3. **the focused window's display** *(new)*
4. last-saved global default bounds
5. gpui's `default_bounds()` cascade

when the active-display branch fires we drop the saved size/position and let gpui cascade defaults on the new display, since the old bounds were sized for whatever display the user was on before.

## test plan

- [x] `cargo check -p workspace`
- [x] `./script/clippy -p workspace`
- [x] manual: with zed fullscreen on a secondary display, cmd-shift-n opens the new window on that display
- [ ] reopening a serialized workspace still restores its saved bounds/display
- [ ] first window after a fresh launch (no focused window) still falls back to the saved default

Release Notes:

- fixed new windows opening on the primary display when the focused window was on a secondary display

### DIFF
--- a/crates/workspace/src/workspace.rs
+++ b/crates/workspace/src/workspace.rs
@@ -1995,6 +1995,16 @@ impl Workspace {
                 } else {
                     let window_bounds_override = window_bounds_env_override();
 
+                    let active_display = cx.update(|cx| {
+                        cx.active_window().and_then(|handle| {
+                            cx.update_window(handle, |_, window, app| {
+                                window.display(app).and_then(|d| d.uuid().ok())
+                            })
+                            .ok()
+                            .flatten()
+                        })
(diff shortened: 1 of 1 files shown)


haxhibekaj:
@cla-bot check

PR #56203, opened 2026-05-08
agent: Allow specifying which model is used for subagents

bennetbo:
Closes #52042 

Release Notes:

- agent: Added setting `subagent_model` to specify which model is used when subagent is spawned

### DIFF
--- a/crates/agent/src/thread.rs
+++ b/crates/agent/src/thread.rs
@@ -984,7 +984,7 @@ pub struct Thread {
     ui_scroll_position: Option<gpui::ListOffset>,
     /// Weak references to running subagent threads for cancellation propagation
     running_subagents: Vec<WeakEntity<Thread>>,
-    inherits_parent_settings: bool,
+    inherits_parent_model_settings: bool,
 }
 
 impl Thread {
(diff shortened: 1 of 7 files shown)


crates/agent/src/thread.rs
```diff
@@ -1016,7 +1017,12 @@ impl Thread {
             parent_thread_id: parent_thread.read(cx).id().clone(),
             depth: parent_thread.read(cx).depth() + 1,
         });
-        thread.inherit_parent_settings(parent_thread, cx);
+        if let Some(subagent_model) = AgentSettings::get_global(cx).subagent_model.clone() {
+            thread.inherits_parent_settings = false;
+            thread.apply_model_selection(&subagent_model, cx);
+        } else {
+            thread.inherit_parent_settings(parent_thread, cx);
```
benbrandt: Can we make sure this is only model settings? Or would we miss stuff like profiles?

benbrandt (approved):

docs/src/ai/agent-settings.md
```diff
@@ -31,6 +31,7 @@ You can assign distinct and specific models for the following AI-powered feature
 - Thread summary model: Used for generating thread summaries
 - Inline assistant model: Used for the inline assistant feature
 - Commit message model: Used for generating Git commit messages
+- Subagent model: Used for subagents spawned by the Agent Panel
```
benbrandt:
```suggestion
- Subagent model: Used for subagents spawned by the Agent Panel. If not set, the subagent will inherit the model settings from the parent thread.
```

PR #56222, opened 2026-05-08
extension_api: Add LSP client commands

vitallium:
Hi! This pull request allows extension-backed language servers to map editor-owned LSP commands, such as show references and run test code lenses.

## Problem

Some language servers emit code lenses or commands that are meant to be handled by the editor, not by the language server through `workspace/executeCommand`. Zed already supports this for built-in language adapters (Rust&Go), but extensions have no equivalent hook.
This means extension-backed language servers cannot implement common editor-owned command flows such as showing references or running tests from code lenses without relying on server-side command execution, even when the action should be owned by Zed.

So I took this PR https://github.com/zed-industries/zed/pull/54100 and prototyped the new extension API support for editor-owned LSP client commands. Extensions can now map specific LSP commands from code lenses/code actions into Zed-handled client behavior instead of always forwarding them to `workspace/executeCommand`.

The initial supported client commands are:
- `ShowLocations` show references/locations from the original LSP command arguments.
- `ScheduleTask` schedule a Zed task from a command, enabling extension-backed “run test” style code lenses.

The editor now resolves client commands asynchronously through the language server adapter before falling back to normal LSP command execution. If a client command cannot be handled, Zed falls back to the existing `workspace/executeCommand` path aka server-based commands.

This prototype currently adds the new export to v0.8.0 since this version is not published to crates.io yet but I am happy to move it to v0.9.0 if the prototype looks good enough.

## How to validate

- Build Zed with changes from this PR
- Build the Ruby extension by using the `vs/handle-ruby-lsp-commands` branch.
- Start Zed, install the dev version of the Ruby extension
- Ensure you have configured the Ruby lang to use `ruby-lsp` LS. `solargraph` does not support code lenses.
- Open any Ruby project
- Go to a test file
- Enable code lenses
- Click `Run test`
- See it in action

## Demo

https://github.com/user-attachments/assets/a94dd8e4-243d-4453-a7cd-5139f6ac0d22

## Notes

- NO support for remote worktrees. I would like to get your review on the approach taken here before proceeding to adding support for remote worktrees.

Happy to answer your questions or take a different direction. Thanks!

Self-Review Checklist:

Release Notes:

- Added support for handling client commands in the Extensions API


ChristopherBiscardi:
Thanks for the PR! This is going to require some staff discussion, since it is a high impact change. I'll forward it to the right people but I wanted to communicate that this might need some patience as that happens.

vitallium:
> Thanks for the PR! This is going to require some staff discussion, since it is a high impact change. I'll forward it to the right people but I wanted to communicate that this might need some patience as that happens.

Hi @ChristopherBiscardi! No rush. I totally understand that every change to the Extensions API can have a high impact and requires additional discussion. Thanks for the ping!

maxbrunsfeld:
I reviewed this and think it makes perfect sense. @vitallium What work would be needed to support remote worktrees?

maxbrunsfeld (changes requested):
Everything here looks good, but let's talk about what we'd need to do to support remote worktrees before we merge. I would think there wouldn't be much change needed, because the new behavior pertains to how the client handles a click on the code lens, but I may be missing some details.

vitallium:
> I reviewed this and think it makes perfect sense. @vitallium What work would be needed to support remote worktrees?

This work was mostly done, just needed some manual testing. Pushed more changes to support remote worktrees.

Here is a short demo with a small Ruby on Rails app that runs in a devcontainer (I hope this use case is OK):


https://github.com/user-attachments/assets/c2daf25a-89d4-4d2b-96f0-c5a2e013cd26


@maxbrunsfeld Thanks for reviewing this! I would like to clarify one point: this PR was created before I joined the Guild. AFAIK the Zed team does not prioritize Extension API at this point, so I'd totally understand if you leave this PR unmerged. Thanks!

crates/extension_api/src/extension_api.rs
```diff
@@ -142,6 +150,16 @@ pub trait Extension: Send + Sync {
         Ok(None)
     }
 
+    /// Maps an LSP command to a Zed client command.
+    fn language_server_client_command(
+        &mut self,
+        _language_server_id: &LanguageServerId,
+        _command: String,
+        _arguments: Vec<serde_json::Value>,
```
MrSubidubi, line 158: For the surrounding APIs, we provide wrapper types for the commands. Now I see that there is not that much to gain from it here perhaps, but I'd still be curios about whether you thought about that, since I assume the command itself will be only matched on in the API and should not be modified, should it?

crates/extension_api/wit/since_v0.8.0/extension.wit
```diff
@@ -11,7 +11,7 @@ world extension {
 
     use common.{env-vars, range};
     use context-server.{context-server-configuration};
-    use dap.{attach-request, build-task-template, debug-config, debug-adapter-binary, debug-task-definition, debug-request, debug-scenario, launch-request, resolved-task, start-debugging-request-arguments-request};
+    use dap.{attach-request, build-task-template, debug-config, debug-adapter-binary, debug-task-definition, debug-request, debug-scenario, launch-request, resolved-task, start-debugging-request-arguments-request, task-template};
```
MrSubidubi, line 14: Do we want to move this record into `common`, perhaps? Since in this context it does not seem to be related to DAP at all.

crates/project/src/lsp_store.rs
```diff
@@ -4467,6 +4467,41 @@ fn should_log_lsp_request_failure(message: &str) -> bool {
...
+}
+
+fn is_builtin_client_command(command_name: &str) -> bool {
+    matches!(
+        command_name,
+        "editor.action.showReferences"
+            | "editor.action.goToLocations"
+            | "editor.action.peekLocations"
+    )
+}
```
MrSubidubi, line 4597: Can we instead scope and expose this on a per-language-server-basis through the `extension.toml` so that we do not have to guess here and can also dispatch this to the extension only if the language server declared this? 

That way, we can for one ensure that actually only the proper commands are passed as well as improve the error message, since we actually know what the extension handles upfront.

crates/extension_api/src/extension_api.rs
```diff
@@ -142,6 +150,16 @@ pub trait Extension: Send + Sync {
...
+    ) -> Result<Option<ClientCommand>> {
```
MrSubidubi, line 159: What do we expect to fail here? Not insisting on removing this, but curious nonetheless.

MrSubidubi (changes requested):

crates/extension_api/wit/since_v0.8.0/extension.wit
vitallium: @MrSubidubi good idea! What do you think about moving `task-template` to `process` instead of `common`? Like, `task-template` is about providing a template for a task that will be executed by a language server in this context but I see opportunities where this can be called by any other thing like extension process command. I don't have a strong pref here tho.

crates/extension_api/src/extension_api.rs
vitallium: I see scenarios when an extension fails to process the command like: a) invalid command name or arguments b) some kind of an internal error.

crates/extension_api/src/extension_api.rs
vitallium: I just used raw types here as a proof-of-concept but now I do see benefits in terms of readability and flexibility. For example, if we declare for `_command` new wrapper type like `LspCommandName` that wil provide clear understanding of what we want.

crates/project/src/lsp_store.rs
vitallium: @MrSubidubi that will require extension schema changes AFAIK but I consider this is a great improvement here, extensions declare upfront which commands they handle, and Zed knows exactly which extension should handle a command. Do you want me to land such thing in this PR or shall this be discussed before?

vitallium:
Also found this PR https://github.com/zed-industries/zed/pull/63607 from @SomeoneToIgnore that looks related to this PR.

vitallium:
> Also found this PR #63607 from @SomeoneToIgnore that looks related to this PR.

The mentioned PR has been merged. I tried to resolve merge conflicts with an agent but failed. I will do that manually.

PR #56230, opened 2026-05-08
agent: Fix tool paths preferring files in subdirectories named after the project root

### LINKED ISSUE #56225: read_file tool makes some files shadowed in subdirectories impossible to read
### Reproduction steps

1. Create a directory "foo" with the following structure:

```
$ find foo
foo
foo/foo
foo/foo/test.txt
foo/test.txt
foo/root.txt
$ cat foo/test.txt
asd
$ cat foo/foo/test.txt
bar
```

A real example of a directory structure like this would be if you have a project containing a Rust workspace with a crate of the same name ("foo" in this case). Both would have a `Cargo.toml` (equivalent to `test.txt` in the example).

2. Open it in zed: `zed foo`
3. Open Zed agent and ask the model to read the file "test.txt" at the project root

**Current behavior:**

1. Model issues a read_file tool call with `{"path": "foo/test.txt"}`. This is faithful to the tool definition: "the first component of the path should always be a root directory in a project."
2. Zed returns the contents of the file `test.txt` in the subdirectory foo, instead of the project root, likely a heuristic to workaround invalid tool calls which exclude the root directory name
3. Because of this "shadowing", the model is in fact unable to read the corresponding file in the project root

Image

**Expected behavior:**

It is expected that the tool behaves consistently. If `read_file({"path":"foo/root.txt"})` returns the contents of the file `root.txt` in the project root, then replacing `root.txt` with `test.txt` should cause it to read the sibling file `test.txt` and not something in a subdirectory of the same name.

OS: macOS 26.4.1
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

procr1337:
The tool definition is very clearly contradicting the previous behavior. Performance impact is unclear to me, we increase the work in a potentially expensive loop, but it seems necessary to have both the specified behavior from the tool definition, as well as the heuristic/fallback for misbehaving models that seems to be intended.

Self-Review Checklist:

Closes #56225

Release Notes:

- Fixed tool paths preferring files in subdirectories named after the project root

### DIFF
--- a/crates/agent/src/tools/read_file_tool.rs
+++ b/crates/agent/src/tools/read_file_tool.rs
@@ -536,7 +536,6 @@ mod test {
         );
     }
 
-    // Regression test for https://github.com/zed-industries/zed/issues/56225
     // When a worktree is named "foo" and contains a subdirectory also named "foo",
     // read_file({"path": "foo/test.txt"}) should return the file at the worktree
     // root (as the tool schema promises), not the one inside the foo/ subdirectory.
--- a/crates/agent/src/tools/read_file_tool.rs
+++ b/crates/agent/src/tools/read_file_tool.rs
@@ -536,6 +536,48 @@ mod test {
         );
     }
 
+    // Regression test for https://github.com/zed-industries/zed/issues/56225
+    // When a worktree is named "foo" and contains a subdirectory also named "foo",
+    // read_file({"path": "foo/test.txt"}) should return the file at the worktree
+    // root (as the tool schema promises), not the one inside the foo/ subdirectory.
+    #[gpui::test]
+    async fn test_read_file_worktree_root_not_shadowed_by_subdir(cx: &mut TestAppContext) {
+        init_test(cx);
+
+        let fs = FakeFs::new(cx.executor());
+        fs.insert_tree(
+            path!("/foo"),
+            json!({
+                "test.txt": "root content",
+                "foo": {
+                    "test.txt": "subdir content"
+                }
+            }),
+        )
+        .await;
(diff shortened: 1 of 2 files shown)


crates/agent/src/tools/read_file_tool.rs
```diff
@@ -536,6 +536,48 @@ mod test {
         );
     }
 
+    // Regression test for https://github.com/zed-industries/zed/issues/56225
```
bennetbo:
```suggestion
```

crates/project/src/project.rs
```diff
@@ -4977,18 +4977,37 @@ impl Project {
...
             }
         } else {
+            // First pass: for each worktree, try two interpretations of the path and
+            // return whichever finds an existing entry first:
+            //   (a) Strip the worktree root name as a prefix. Tool definitions instruct
+            //       the model that "the first component of the path should always be a
+            //       root directory in a project", so e.g. "foo/file.txt" means file.txt
+            //       at the root of the "foo" worktree. This takes priority so that a
+            //       subdirectory that shares the worktree's name doesn't shadow root files.
+            //   (b) Treat the path as a literal worktree-relative path.
```
bennetbo, line 4983:
```suggestion
            // First pass: for each worktree, try two interpretations of the path and
            // return whichever finds an existing entry first:
            //   (a) Strip the worktree root name as a prefix.
            //   (b) Treat the path as a literal worktree-relative path.
```

bennetbo (commented):
Thank you, looks reasonable. We should get rid of some of the verbose comments (GitHub does not let me push to your branch)

bennetbo (approved):

PR #56237, opened 2026-05-08
Attempt to make macOS test runs faster

ConradIrwin:
Claude's analysis shows we have a roughyl 1.2s overhead per test on macOS runners only

Release Notes:

- N/A


SomeoneToIgnore:
Will close to keep the queue cleaner: CI fails for over 3 weeks.

PR #56247, opened 2026-05-09
search: set pending_external_query in use_selection_for_find when search bar is dismissed

### LINKED ISSUE #55619: cmd-e/cmd-g behaviour is broken
### Reproduction steps

1. Select word

OS: macOS 26.4.1
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

_No response_

None

Dev-X25874:
Fixes cmd-e / cmd-g regression where using cmd-e (UseSelectionForFind) 
while the search bar is dismissed would update the find pasteboard but 
not set `pending_external_query`, causing cmd-g (SelectNextMatch) to 
silently no-op via `WithResultsOrExternalQuery` (which gates on 
`pending_external_query.is_some() || active_match_index.is_some()`).

When the search bar is dismissed, `use_selection_for_find` now also 
sets `pending_external_query` with the selected text and current search 
options, mirroring the same pattern used in the `on_focus` handler 
(line ~715). The next cmd-g then picks up the pending query, runs the 
search, and activates the first match as expected.

Self-Review Checklist:

Closes #55619

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -1249,14 +1249,18 @@ impl BufferSearchBar {
         self.query_editor.update(cx, |query_editor, cx| {
             query_editor.buffer().update(cx, |query_buffer, cx| {
                 let len = query_buffer.len(cx);
-                query_buffer.edit([(MultiBufferOffset(0)..len, search_text)], None, cx);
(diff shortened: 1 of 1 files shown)


Dev-X25874:
@cla-bot check

PR #56256, opened 2026-05-09
terminal_view: Hide agent actions in embedded terminal menus

### LINKED ISSUE #56138: Invoking inline assist in terminal tool output ends up in current editor buffer
### Reproduction steps

1. Open a file in the editor
2. Run prompt: We're debugging Zed. Run the command `ls -al`.
3. Right click in the tool output
4. Choose Inline Assist

Inline assist pops up in the editor window, unrelated to the tool output

https://github.com/user-attachments/assets/e6cd97f2-7abc-4c3d-b124-63ffab8f47e3

Expected: no "Inline assist" context menu option in tool outputs?

Zed Dev 1.3.0 (debug)
Commit: 6aa90e750ded06c684393ab4cccf159c731f58b6
Version: 1.3.0+dev.6aa90e750ded06c684393ab4cccf159c731f58b6

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

terror:
Resolves https://github.com/zed-industries/zed/issues/56138

This diff hides terminal agent actions from context menus shown by embedded terminal views, such as terminal tool output in agent threads. Those surfaces render through `TerminalView`, but they are not normal interactive terminal panes, and actions like Inline Assist and Add to Agent Thread do not currently resolve the embedded terminal as their target.

Previously, choosing Inline Assist from terminal tool output could target the workspace’s active editor instead of the terminal output. The context menu now limits these agent actions to non-embedded terminal views, so regular terminal panes keep their existing behavior while terminal tool output no longer offers actions that cannot resolve to it.

Add to Agent Thread _could_ make sense for embedded terminal selections in the future, but it would need explicit target-resolution support for focused embedded terminals. This change keeps that as separate follow-up work instead of leaving a misleading menu item that does not reliably act on the selected tool output.

Release Notes:

- Fixed embedded terminal context menus showing agent actions that could target the wrong item.

### DIFF
--- a/crates/terminal_view/src/terminal_view.rs
+++ b/crates/terminal_view/src/terminal_view.rs
@@ -511,13 +511,16 @@ impl TerminalView {
                 .action("Paste Text", Box::new(PasteText))
                 .action("Select All", Box::new(SelectAll))
                 .action("Clear", Box::new(Clear))
-                .when(assistant_enabled, |menu| {
-                    menu.separator()
-                        .action("Inline Assist", Box::new(InlineAssist::default()))
-                        .when(has_selection, |menu| {
-                            menu.action("Add to Agent Thread", Box::new(AddSelectionToThread))
-                        })
-                })
+                .when(
+                    assistant_enabled && !matches!(self.mode, TerminalMode::Embedded { .. }),
+                    |menu| {
+                        menu.separator()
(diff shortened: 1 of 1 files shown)


bennetbo (approved):
Thank you!

Issue #56019, opened 2026-05-07
git_ui: Show parent directory in repository selector for worktrees


josiaranda:
## Summary

When using git worktrees, the repository selector in the git panel footer only shows the immediate directory name (e.g. `wt-1`), which can be ambiguous when working with multiple repositories that use similar worktree naming conventions.

**1. Current behavior:**
`wt-1 / develop`

<img width="363" height="1413" alt="Image" src="https://github.com/user-attachments/assets/710255a0-7d1f-459c-be47-75df9fce68b2" />

<img width="363" height="1413" alt="Image" src="https://github.com/user-attachments/assets/3b69bb76-7e14-4542-9d32-0b4eaeebbbd3" />

**2. Proposed behavior:**
`my-project/wt-1 / develop`

Including the parent directory in the display name provides better context about which repository a worktree belongs to. This is useful in any multi-repo workflow where worktree directories share common names:

```
workspace/
  project-alpha/
    wt-1/   ← currently shows "wt-1"
    wt-2/
  project-beta/
    wt-1/   ← also shows "wt-1", ambiguous
```

With the proposed change, these would show as `project-alpha/wt-1` and `project-beta/wt-1` respectively.

## Proposed change

<img width="363" height="1413" alt="Image" src="https://github.com/user-attachments/assets/04fe006d-f898-498c-a14b-553c10b8ad72" />

<img width="363" height="1413" alt="Image" src="https://github.com/user-attachments/assets/c93bf496-13bc-4701-978b-c4891621db6a" />

Modify `Repository::display_name()` in `crates/project/src/git_store.rs` to include the parent directory component in addition to the leaf directory name.

This is a small, targeted change — the `display_name()` method is already the single source for the repo label in the git panel footer.

The `MAX_REPO_LEN` truncation budget in `git_panel.rs` would also need a bump to accommodate the longer string.

PR #56265, opened 2026-05-09
git_ui: Include parent directory in repository display name

josiaranda:
Relates to https://github.com/zed-industries/zed/discussions/56255
Originally opened as https://github.com/zed-industries/zed/issues/56019

When using git worktrees, the repository selector in the git panel footer only shows the leaf directory name (e.g. `wt-1`), which can be ambiguous when working with multiple repositories that share similar worktree naming conventions.

This change modifies `Repository::display_name()` to include the parent directory, so the label becomes `project-alpha/wt-1` instead of just `wt-1`.

Also bumps `MAX_REPO_LEN` from 16 to 32 to accommodate the longer display string.

Self-Review Checklist:

Release Notes:

- Improved git panel repository selector to show parent directory for better disambiguation of worktrees


dinocosta (changes requested):
Hi @josiaranda ! Thank you for taking the time to contribute to Zed :)

Reviewing the original discussion, I understand why this can be confusing. However, I'm not sure if we'll always want to display the parent directory name . For example, if I have a project that has both `Developer/foo` and `Developer/bar`, we'll end up showing those two names in the Git Panel, even though, in that situation, we wouldn't need to, as it's obvious which repository you're viewing just by reading `foo` or `bar` alone. You can see how this would look like in the screenshots below:

| | |
|-|-|
| Image | Image |

I'm also unsure whether this is actually needed. Are you running the latest stable release? If so, the original repository name should already be visible in the title bar, making it much easier to differentiate between repositories, even if the worktree's folder name is the same, as can be seen below:

| | |
|-|-|
| Image | Image |

In both cases I'm working on a worktree whose folder is named `mellow-crest` but, looking at the title bar, I'm able to disambiguate between the `dotfiles` and `bar` repositories.

Let me know what you think!

BirknerAlex:
Hi @dinocosta — thanks for the thorough reply, and sorry for jumping into someone else's PR. I'm not the author, just another user who runs into this, so I wanted to add a case that I think the title-bar argument doesn't cover.

Our setup is a meta-repository that pulls in five services as submodules, and every submodule checks out into a directory called `src` — that's baked into the build paths, so renaming them isn't really on the table. The selector ends up looking like this:

src-submodule

Nothing on any of those rows tells you which project it is, and the filter field matches the same name, so typing the service name finds nothing either.

The difference from your worktree screenshots is that there the title bar shows the outer repository, which does disambiguate. With submodules it shows the workspace root — the same string for every row above — so the selector is the only place this could show up.

That said, I do think your point about `Developer/foo` and `Developer/bar` is fair; always prepending the parent adds noise where there's nothing to disambiguate. Would a middle ground work — keep the leaf name, and only prepend parent components for names that collide within the workspace? `foo` and `bar` would be untouched, while the list above would become `projectA/src`, `projectB/src`, `projectC/src`, and so on.

If that direction sounds reasonable to you I'd be glad to help out, whether that's testing or taking a shot at the implementation.

dinocosta:
> Hi @dinocosta — thanks for the thorough reply, and sorry for jumping into someone else's PR. I'm not the author, just another user who runs into this, so I wanted to add a case that I think the title-bar argument doesn't cover.

No need to say sorry, you're more than welcome to chime in and share your use case, it's the only way for us to be aware of some of these scenarios that we, internally, end up not using and finding on our own  🙂

> With submodules it shows the workspace root — the same string for every row above — so the selector is the only place this could show up.

Thank you for sharing that detailed breakdown, this is very valid feedback!

> That said, I do think your point about `Developer/foo` and `Developer/bar` is fair; always prepending the parent adds noise where there's nothing to disambiguate. Would a middle ground work — keep the leaf name, and only prepend parent components for names that collide within the workspace? `foo` and `bar` would be untouched, while the list above would become `projectA/src`, `projectB/src`, `projectC/src`, and so on.
> 
> If that direction sounds reasonable to you I'd be glad to help out, whether that's testing or taking a shot at the implementation.

I believe we'd first need to see how it looks, but this seems like the most interesting approach, ensuring that we only display this information when necessary. I'd be down to seeing an implementation and testing it out!

@josiaranda Do you have an interest in continuing this work, and possibly having @BirknerAlex test your changes as you go? If not, would you mind if @BirknerAlex takes a stab at it?

Thanks!

BirknerAlex:
@josiaranda @dinocosta — quick update from my side.

I had some free tim last evening and ended up building the collision-only variant we discussed above, so I've opened it as #62800. It keeps the leaf name wherever it's already unique and only grows the colliding ones, so `Developer/foo` and `Developer/bar` stay exactly as they are.

@josiaranda, this is absolutely not meant to take the work off you — you opened this first and did the thinking that got us here. If you'd like to carry it forward yourself, please just say so and I'll close mine straight away, no hard feelings at all. I'm honestly happy either way; I'd just love to see it land in some form, because I run into it every day in my company's project and it makes the repository selector quite hard to use.

Either way, happy to help test or review whatever ends up going in.

Thanks both!

josiaranda:
@BirknerAlex @dinocosta yeap! feel free to take this over, I ended up building it locally and use that for myself on my machine!

@BirknerAlex Perhaps to reduce duplicates you can just continue on this PR? So feel free to do what suits youi best

dinocosta:
Thank you both for aligning on this! 🙂

> @BirknerAlex Perhaps to reduce duplicates you can just continue on this PR? So feel free to do what suits youi best

I believe that would only be possible if you invited @BirknerAlex to collaborate on your fork. It might be easier to simply close this one in favor of #62800 .

@BirknerAlex Would you mind adding a commit, or editing an existing one, to mark @josiaranda as a [co-author](https://docs.github.com/en/pull-requests/how-tos/commit-changes/creating-a-commit-with-multiple-authors) on your Pull Request? Thanks!

BirknerAlex:
Sure, I added @josiaranda as co-author and pushed it to #62800

PR #56267, opened 2026-05-09
editor: Show file icon in breadcrumbs when tab bar is hidden

hron:
[Screencast_20260509_113456.webm](https://github.com/user-attachments/assets/8165246f-6266-4e23-bede-0755e8636a19)

Release Notes: 

- Made file icons be displayed when tab bar is hidden and file icons are turned on for tabs.


danilo-leal (approved):
Thank you!

Issue #56067, opened 2026-05-07
Agent file links still show macOS -50 dialog in remote sessions after #53229


dl-alexandre:
### Reproduction steps

1. Start Zed on macOS.
2. Open a remote project/session.
3. Use the agent panel/editor and ask the agent to reference or summarize files in the remote workspace.
4. Click a file/path link rendered in the agent response.

### Current vs. Expected behavior

Current behavior:

Clicking the file/path link does not open the remote file in Zed. Instead, macOS shows a Finder-style dialog:

```text
The application can't be opened.
-50
```

Expected behavior:

Clicking a file/path link in an agent response should open the corresponding file in the current Zed editor, including when the workspace is remote. If the link includes a line number, Zed should navigate to that line.

This looks closely related to #49978 and #50107, which were expected to be fixed by #53229, but it still reproduces on a newer build.

### Zed version and system specs

Zed: 1.1.6
Build: 20260506.212021
OS: macOS
Architecture: aarch64

### Attach Zed log file

No relevant agent/link parsing errors were found in `~/Library/Logs/Zed/Zed.log` after checking for entries like `relative URL without a base`, `open_url`, and `MentionUri`.

### (for AI issues) Model provider details

Agent panel/editor in a remote session. The issue is triggered by clicking file/path links rendered in the agent output.

### If you are using WSL on Windows, what flavor of Linux are you using?

None

### Additional context

Prior related issues:

- #49978
- #50107
- #50350

Related merged fix:

- #53229

The visible symptom matches the older reports exactly: the file link appears to be routed through the OS/application opener instead of being resolved against the active remote Zed workspace.

yeskunall:
If you see the PR you linked, it has only been merged to `main` so far, and in the issues you linked, you can clearly see it was only pushed to preview:

<img width="357" height="125" alt="Image" src="https://github.com/user-attachments/assets/d2d1dea9-4d8b-48eb-9e1f-d69136a98092" />


If you’re eager for a fix, grab the [latest preview](https://zed.dev/releases/preview) release.

PR #56283, opened 2026-05-09
Fix agent hyperlinks do not open files

di404:
## Summary

Fix agent message hyperlinks that point to Windows file paths but are not parsed as openable project paths.

This covers links like:

```md
[Cargo.toml](</C:/Projects/Example Workspace/Cargo.toml:2>)
[filename.ext](C:\Projects\Example%20Workspace\path\to\filename.ext:42)
[AGENTS.md](</c/Projects/Example Workspace/AGENTS.md>)
```

## Problem

Agent responses can emit Markdown hyperlinks whose targets are Windows paths rather than `file://` URLs. Some of those targets include a leading slash before the drive (`/C:/...`), Git Bash/MSYS-style drive prefixes (`/c/...`), percent-escaped spaces, or line suffixes. These were not normalized before mention parsing, so clicking the hyperlink could do nothing instead of opening the file.

## Solution

- Normalize hyperlink path targets before parsing them as `MentionUri` paths.
- Decode percent escapes in bare path targets so `%20` becomes a literal space before path/line parsing.
- Convert Windows-compatible hyperlink paths such as `/C:/...` and `/c/...` into native Windows paths.
- Generate file resource links from `find_path_tool` through `MentionUri::to_uri()` instead of hand-building `file://` strings.

## Result

Before: clicking agent path hyperlinks did not open the referenced file.

After:  

https://github.com/user-attachments/assets/6c7fad77-4a1e-4497-a4f9-4a4fdf86d527

## Validation

- `cargo test -p acp_thread test_parse_windows --features test-support`
- `cargo fmt --check`

## Follow-up changes

Additions on top of the original work above:

- Moved the hyperlink heuristics into a dedicated `MentionUri::parse_hyperlink` entry point. `MentionUri::parse` stays strict, so canonical mention URIs round-trip verbatim and other callers (message editor, thread deserialization, resource links) are unaffected.
- Percent escapes that decode to path separators (`%2F`, `%5C`) are left encoded, so decoding can never change which directories a path traverses.
- Bare paths with escapes are ambiguous (a file may literally be named `a%20b.rs`): `open_link` prefers the decoded interpretation and falls back to `MentionUri::parse_hyperlink_literal` when the decoded path doesn't resolve in the project but the literal one does.
- Links to files outside the project's worktrees now open, gated by an async existence check through the project `Fs` (correct for remote projects; broken links no longer create empty buffers or add worktrees). `open_link` and the mention-crease open path are unified into one `open_abs_path_at_point`, which now also places the cursor for out-of-project selection/symbol links.
- `grep_tool` resource links also go through `MentionUri::to_uri()` now, fixing malformed `file://C:\...` URIs and unencoded spaces.
- Added tests: percent-escape disambiguation, out-of-project link opening, drive-letter normalization, UNC paths, and literal-path parsing (`cargo test -p acp_thread mention`, `cargo test -p agent_ui open_link`, `cargo test -p agent grep_tool`); manually verified the link spellings above on Windows.

Release Notes:

- Fixed agent path hyperlinks on Windows when paths contain spaces or shell-style drive prefixes.

### DIFF
--- a/crates/acp_thread/src/mention.rs
+++ b/crates/acp_thread/src/mention.rs
@@ -284,21 +284,11 @@ impl MentionUri {
 
     /// Parses a hyperlink target from agent-authored Markdown.
     ///
-    /// Agents spell out file hyperlinks in ways [`MentionUri::parse`]
-    /// intentionally does not accept: bare paths with percent-encoded
-    /// characters (`My%20File.rs`) and Windows-compatible spellings such as
-    /// `/C:/foo`, `/c/foo`, or forward slashes. This entry point normalizes
-    /// bare-path targets before parsing them; `parse` stays strict so that
-    /// canonical mention URIs round-trip verbatim and other callers are
-    /// unaffected by these hyperlink heuristics.
-    ///
-    /// Percent escapes that decode to path separators (`%2F`, `%5C`) are
-    /// left encoded, since decoding them would change which directories the
-    /// path traverses rather than a character within one component. A file
-    /// whose name literally contains a decodable escape sequence (e.g.
-    /// `a%20b.rs`) is misread by this default; callers with access to the
-    /// project can use [`MentionUri::parse_hyperlink_literal`] to
-    /// disambiguate.
+    /// Unlike [`MentionUri::parse`] — which stays strict so canonical mention
+    /// URIs round-trip verbatim — bare path targets are normalized first:
+    /// percent escapes are decoded (see [`decode_path_escapes`]) and
+    /// Windows-compatible spellings like `/C:/foo` or `/c/foo` become native
+    /// paths (see [`to_native_windows_path`]).
     pub fn parse_hyperlink(input: &str, path_style: PathStyle) -> Result<Self> {
         if let Some(target) = bare_path_target(input, path_style) {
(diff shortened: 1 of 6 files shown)


svenit:
Does this MR work on macOS? I’m having the same issue

CleanShot 2026-05-10 at 13 45 54@2x

di404:
> Does this MR work on macOS? I’m having the same issue这个合并请求在 macOS 上能用吗？我也遇到了同样的问题。
> 
> CleanShot 2026-05-10 at 13 45 54@2x

I don't think that will work; this path format fix is designed for Windows. Could you show me the origin Markdown text from the error chat? If the issue is caused by wrong path format, I can add a fix.

---edit---
I saw they already fix it in a pre version https://github.com/zed-industries/zed/issues/56067#issuecomment-4398525707

svenit:
@di404 
`[server.ts](/Users/macos/Code/work/server.ts:306)`

di404:
> @di404 `[server.ts](/Users/macos/Code/work/server.ts:306)`

I see, it was caused by some other problem IMO, you could try v1.20 to see  if it solves your problem.

PR #56299, opened 2026-05-09
workspace: Skip closed items that cannot be reopened

### LINKED ISSUE #55600: pane::ReopenClosedItem doesn't skip non-file tabs
### Reproduction steps

1. Open a file and then close it
2. Open Project Search and then close it
3. Open a new buffer and then close it
4. Open Default Settings and then close it
5. Press cmd-shift-t 4 times

cmd-shift-t doesn't reopen Project Search, the unsaved buffer, or Default Settings. The first 3 presses are no-ops, and then the 4th press reopens the file.

cmd-shift-t should skip unsaved buffers, because they can't be reopened. For Default Settings and Project Search, they should either support reopening or be skipped.

Zed: v1.1.6+stable.265.96962914296a22c22a7f226b5ae7de4dd0e34729 (Zed) 
OS: macOS 15.7.5
Memory: 16 GiB
Architecture: x86_64

terror:
Resolves https://github.com/zed-industries/zed/issues/55600

This diff fixes `pane::ReopenClosedItem` getting stuck when the closed-item stack contains entries that cannot be reopened, such as Project Search, untitled buffers, or Default Settings. Previously, `Workspace::navigate_history_impl` would pop the newest closed entry and stop if that item was no longer present in the pane and had no path recorded for reopening. That made `cmd- shift-t` appear to do nothing until enough attempts had consumed those unreopenable entries.

With this change, closed-item navigation keeps scanning when it encounters an entry that cannot be activated or reopened by path. This preserves the current path-based reopening behavior for normal files, while avoiding no-op shortcuts caused by non-file items in the closed stack.

This made me wonder whether or not we'd eventually want full reopen support for non-traditional items like Project Search or bundled settings editors. Supporting that properly would require storing item-specific restoration state, such as search query/options for Project Search or a bundled-file descriptor for Default Settings, and teaching closed-item navigation how to recreate those items from that state. Something definitely out of scope for this PR.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/c7044423-4531-4857-84f2-4e9651826c6a" controls width="500" title="Before"></video> | <video src="https://github.com/user-attachments/assets/c89dcefb-1796-4cdf-bb21-f165145e678e" controls width="500" title="After"></video> |

Release Notes:
- Fixed reopening closed tabs getting stuck on closed items that cannot be reopened.


Veykril (approved):
Thanks!

PR #56308, opened 2026-05-10
agent_servers: Allow multiple registry agent instances via agent_id

marshallswain:
## Summary

- Adds an optional `agent_id` field to the `Registry` variant of `agent_servers` settings.
- When set, it overrides the registry agent lookup; the `HashMap` key continues to drive the per-instance cache directory and is also used as the picker display name (so each alias gets a distinct label).
- Enables side-by-side instances of the same registry agent (e.g. two Claude Code accounts with different `CLAUDE_CONFIG_DIR`s in the same project).

Existing settings are unaffected: when `agent_id` is `None` (the default), behavior is identical to today, and the picker label still comes from the registry's metadata.

Here's what it looks like built.  With the "Example" config, below.  This lets me run two separate, logged-in Claude Code accounts at the same time in the agent panel.

Screenshot 2026-05-09 at 6 45 14 PM

## Related

Discussion #55461 raises the same use case (Claude work/personal accounts via `CLAUDE_CONFIG_DIR`). That discussion proposes per-project `agent_servers` config; this PR proposes a smaller, alternative shape:

- No new project-scoped config plumbing.
- Multiple accounts coexist in the same agent picker (visible side-by-side rather than one-per-project).
- Existing settings keep working unchanged; no migration.

The two approaches aren't mutually exclusive — happy to defer if maintainers prefer the per-project shape, but wanted to put a concrete patch up since the use case is real and the change is contained.

## Example

```jsonc
"agent_servers": {
  "claude-acp": { "type": "registry" },
  "Claude Agent (Work)": {
    "type": "registry",
    "agent_id": "claude-acp",
    "env": { "CLAUDE_CONFIG_DIR": "/Users/me/.claude-work" }
  }
}
```

The canonical `claude-acp` entry continues to show its existing label ("Claude Agent" — sourced from the registry's metadata, unchanged from today). The aliased `Claude Agent (Work)` entry shows the settings key as its label, with its own cache directory and env.

## Test plan

- [x] `cargo check -p settings_content -p project -p agent_servers -p onboarding -p agent_ui`
- [x] `cargo test -p project --lib agent_server_store::tests::` — added `test_registry_agent_id_alias_creates_distinct_instance` covering both the registry lookup override and the display-name fallback
- [x] `./script/clippy` — clean
- [x] Manually verify two `claude-acp` instances appear in the agent picker with isolated envs

Release Notes:

- Added an optional `agent_id` field to `agent_servers` registry settings, allowing multiple instances of the same registry agent (e.g. side-by-side Claude Code accounts with isolated `CLAUDE_CONFIG_DIR`s)


marshallswain:
@cla-bot check

benbrandt:
hey yeah I think given other aspects of how the registry agents work, I think I'd rather have per project settings. Thanks for taking a look though!

marshallswain:
Mind if I push back gently on the direction to make sure I understand it?

I had been thinking of the alias shape as a foundation for the per-project case, not an alternative to it. The reason: `.zed/settings.json` is team-shared via git, so a per-project `agent_servers` block can't carry `CLAUDE_CONFIG_DIR: "/Users/marshall/.claude-bitovi"` — that path is different for every dev on the team. The OP in #55461 is a solo contractor where this isn't an issue, but if per-project is meant to scale to teams, "personal credentials in a checked-in file" doesn't compose.

The two team-friendly shapes I can see both need something like aliases underneath:

- Project declares agent structure (no env); user settings supply env, composed via merge
- Project picks an agent by name; user settings define what that name means

Is there a third model where per-project `agent_servers` solves the env-per-user problem without a per-user-per-project file? Or is the intent to scope this to solo devs for now? Happy to take it whichever direction you'd prefer — just want to make sure I'm building toward the right shape next time.

PR #56309, opened 2026-05-10
Add `code_fence_block_aliases` for highlighting markdown code blocks tagged with alternative language names

tdanner:
Currently code blocks in markdown are highlighted if the tag matches the name of a registered language or any associated file extension. However, there are some languages that have other names that are sometimes used for this purpose, like `protobuf` for Proto or `shell` for Bash. GitHub highlights code blocks in markdown using the Linguist library, which has [a long list of languages and alternative names](https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml).

Previous attempt at this: https://github.com/zed-industries/zed/pull/54354. This version is better because it allows for multiple aliases.

Self-Review Checklist:

Closes #51932 

Release Notes:

- Adds `code_fence_block_aliases` to language metadata to highlight markdown code blocks tagged with alternative language names (such as `shell` instead of `bash` or `protobuf` instead of `proto`)


crates/markdown/src/markdown.rs
```diff
@@ -809,7 +809,7 @@ impl Markdown {
             if let Some(registry) = language_registry.as_ref() {
                 for name in language_names {
                     let language = if !name.is_empty() {
-                        registry.language_for_name_or_extension(&name).left_future()
+                        registry.language_for_code_fence_name(&name).left_future()
```
MrSubidubi, line 812: A small comment here or change of name for `language_names` would be nice to indicate these come from code blocks.

crates/language/src/syntax_map.rs
MrSubidubi, line 1: Why are these changes required?

crates/language/src/language_registry.rs
```diff
@@ -599,6 +599,44 @@ impl LanguageRegistry {
         async move { rx.await? }
     }
 
+    /// Look up a language by the info string of a Markdown fenced code block.
+    ///
+    /// This performs a case-insensitive match against:
+    /// 1. The language name
+    /// 2. The language's path suffixes
+    /// 3. Explicit code fence aliases defined in the language config
```
MrSubidubi: Are we sure about this order? I'd think that if we know what is passed here is a code fence block name, why do we not match against these first and give them a higher priority?

MrSubidubi (changes requested):
Thanks for this!

I think we should reuse code fence block name here instead of adding a new config entry and rework this to also allow taking lists. I think you can see how that works for snippets and we can `serde(alias = ...)` this to have a name like `code_fence_block_names` in code but also take the old `code_fence_block_name` as an alias

Also, see the other comments

tdanner:
> reuse code fence block name here instead of adding a new config entry and rework this to also allow taking lists

I like this idea. Because `code_fence_block_name` used when generating markdown output, we will need to adjust slightly if we make it take a list. Simplest option would be to just use the first item in the list as the block name for output.

crates/language/src/syntax_map.rs
tdanner: This is where Zed does language lookups for injections, including markdown code fence blocks but also C++ raw string literals and others. `language_for_code_fence_name` matches a superset of what `language_for_name_or_extension` matches, so there should be no regression here.

`markdown.rs` handles language lookup for the markdown preview.

crates/markdown/src/markdown.rs
tdanner: Renaming `language_names` to `code_block_language_names` should make this more clear.

crates/language/src/language_registry.rs
tdanner: I don't think there is actually a priority order enforced here. If multiple languages claim the same string (as a name, path suffix, or code fence name), then it looks like it would just be arbitrary which one wins. The precedence (copied from the existing `language_for_name_or_extension`) is just the length of `string` - but that's a function parameter here, not something that actually differentiates candidate languages.

PR #56315, opened 2026-05-10
gpui_macos: Drop legacy pre-macOS 12 blur path and its private CGS APIs

codeitlikemiley:
## Summary

`gpui_macos` links two private CoreGraphics Services APIs unconditionally:

```rust
#[link(name = "CoreGraphics", kind = "framework")]
unsafe extern "C" {
    fn CGSMainConnectionID() -> id;
    fn CGSSetWindowBackgroundBlurRadius(...) -> i32;
}
```

Apple rejects App Store submissions that reference private APIs, so any
application built on GPUI is ineligible for the Mac App Store — regardless of
whether it ever sets `WindowBackgroundAppearance::Blurred`, since the symbols are
linked either way.

Their only call site is the `NSAppKitVersionNumber < NSAppKitVersionNumber12_0`
branch: macOS 11 and earlier. On macOS 12+ the same effect already goes through
`NSVisualEffectView`, which the existing comment notes is the better path anyway
(it downsamples the backdrop and gives more control over the effect layer).

This deletes the legacy branch and the extern block.

## Changed from the previous revision of this PR

This originally added a `macos_app_store` feature gate. @reflectronic suggested
deleting the legacy `WindowBackgroundAppearance::Blurred` support instead of
carrying a flag — that is the better call and this PR now does that.

It is the stronger fix on the merits, not just the smaller one: deleting makes
every GPUI consumer App Store eligible, whereas a feature flag only helps those
who opt in and leaves the default build ineligible. It also removes a
conditional-compilation path rather than adding one. The diff went from
3 files / +13 −7 to **1 file / +23 −52**.

## Notes for review

Two details that are easy to miss:

- `this.native_window.windowNumber()` resolved through cocoa's **`NSEvent`**
  trait rather than `NSWindow` — `id` is untyped, so it picks up every trait in
  scope. Removing that call makes the `NSEvent` import genuinely unused, so it is
  dropped. Verified against a build of unmodified `main`, which does not emit
  that warning, to confirm this is a consequence of the change and not
  pre-existing.
- `NSAppKitVersionNumber` and `NSAppKitVersionNumber12_0` had no other users and
  are dropped with it.

The `NSVisualEffectView` branch is preserved verbatim; the only change to it is
one level of de-indentation.

## Behaviour change

On macOS 11 and earlier, `WindowBackgroundAppearance::Blurred` no longer applies
a blur — the window falls back to the `NSVisualEffectView` path, which those
releases handle via `CAProxyLayer` rather than concrete sublayers. macOS 11 is
past end of life. Every other platform and macOS 12+ are unaffected.

## Verification

`cargo check -p gpui_macos` on macOS — clean, no errors and no new warnings.

The identical change was also applied to a downstream GPUI fork and built against
a ~60k-line application using `gpui-component`, where CI passed on macOS, Ubuntu
and Windows.

Release Notes:

- N/A

### DIFF
--- a/crates/gpui_macos/src/window.rs
+++ b/crates/gpui_macos/src/window.rs
@@ -9,12 +9,11 @@ use anyhow::Result;
 use block::ConcreteBlock;
 use cocoa::{
     appkit::{
-        NSAppKitVersionNumber, NSAppKitVersionNumber12_0, NSApplication, NSBackingStoreBuffered,
-        NSColor, NSEvent, NSEventModifierFlags, NSFilenamesPboardType, NSPasteboard,
-        NSRequestUserAttentionType, NSScreen, NSView, NSViewHeightSizable, NSViewWidthSizable,
-        NSVisualEffectMaterial, NSVisualEffectState, NSVisualEffectView, NSWindow,
-        NSWindowCollectionBehavior, NSWindowOcclusionState, NSWindowOrderingMode,
-        NSWindowStyleMask, NSWindowTitleVisibility,
+        NSApplication, NSBackingStoreBuffered, NSColor, NSEventModifierFlags,
+        NSFilenamesPboardType, NSPasteboard, NSRequestUserAttentionType, NSScreen, NSView,
+        NSViewHeightSizable, NSViewWidthSizable, NSVisualEffectMaterial, NSVisualEffectState,
+        NSVisualEffectView, NSWindow, NSWindowCollectionBehavior, NSWindowOcclusionState,
+        NSWindowOrderingMode, NSWindowStyleMask, NSWindowTitleVisibility,
     },
     base::{id, nil},
     foundation::{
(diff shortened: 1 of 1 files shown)


codeitlikemiley:
@cla-bot check

reflectronic:
Thank you for the pull request. I'm inclined to just delete the support for `WindowBackgroundAppearance::Blurred` on these legacy versions of macOS. I don't think we need the feature flag.

codeitlikemiley:
@reflectronic thanks — you were right, and I've reworked this as the deletion you
suggested rather than the feature flag.

Agreed on the merits, not just the size: deleting makes every GPUI consumer App
Store eligible, whereas a flag only helps the ones who opt in and leaves the
default build ineligible. It also removes a conditional-compilation path instead
of adding one. 3 files / +13 −7 became 1 file / +23 −52.

Two things surfaced while doing it that are worth a look during review:

- `this.native_window.windowNumber()` was resolving through cocoa's `NSEvent`
  trait rather than `NSWindow` — `id` is untyped, so it picks up every trait in
  scope. Removing that call makes the `NSEvent` import dead, so it's dropped
  here. I confirmed against a build of unmodified `main` that the resulting
  unused-import warning is caused by this change and isn't pre-existing.
- `NSAppKitVersionNumber` / `NSAppKitVersionNumber12_0` had no other users.

The `NSVisualEffectView` branch is otherwise untouched — the only change to it is
de-indentation.

Verified with `cargo check -p gpui_macos` on macOS: no errors, no new warnings.
The same change applied to a downstream GPUI fork also built against a ~60k-line
`gpui-component` application with CI green on macOS, Ubuntu and Windows.

Rebased onto current `main` as well; the branch was three months stale.

reflectronic (approved):
Thank you!

Issue #55358, opened 2026-04-30
current git worktree state isn't preserved across recent project picker


willlma:
### Reproduction steps

1. Open project A on main worktree.
2. Switch to worktree "foo"
3. Open project B.
4. Open project switcher
5. Select project A

### Current behavior

At step 4, projects are not sorted by recency. Project C is on top.

After step 5, we're on the main worktree.

### Expected behavior

At step 4, project A should be on top. Projects should be sorted by recency, like an app switcher or a window switcher. The most common flow is to go back and forth between projects A and B and never touch C.
 
After step 5, I should be back on worktree "foo", the last opened worktree for the project, not main. 

### Zed version and system specs

Zed: v1.0.0+stable.257.5ec84a926ef83865afb92d2a3d1ca3b419572cf9 (Zed) 
OS: macOS 15.7.5
Memory: 64 GiB
Architecture: aarch64

### Attach Zed log file

Nothing added to Zed log. 

### Relevant Zed settings

<details><summary>settings.json</summary>

```json
// Zed settings
//
// For information on how to configure Zed, see the Zed
// documentation: https://zed.dev/docs/configuring-zed
//
// To see all of Zed's default settings without changing your
// custom settings, run `zed: open default settings` from the
// command palette
{
  "cli_default_open_behavior": "existing_window",
  "outline_panel": {
    "dock": "left",
  },
  "collaboration_panel": {
    "dock": "right",
  },
  "project_panel": {
    "dock": "left",
    "hide_hidden": false,
  },
  "search": {
    "regex": true,
  },
  "buffer_line_height": "comfortable",
  "agent": {
    "dock": "right",
    "play_sound_when_agent_done": "always",
  },
  "buffer_font_size": 15.0,
  "diff_view_style": "unified",
  "agent_servers": {
    "claude-acp": {
      "default_config_options": {
        "model": "default",
      },
      "type": "registry",
    },
    "codex-acp": {
      "type": "registry",
    },
    "gemini": {
      "type": "registry",
    },
  },
  "git_panel": {
    "dock": "left",
    "tree_view": true,
    "sort_by_path": true,
  },
  "edit_predictions": {
    "provider": "zed",
    "mode": "subtle",
  },
  "languages": {
    "Git Commit": {
      "soft_wrap": "preferred_line_length",
      "preferred_line_length": 72,
    },
    "JavaScript": {
      "prettier": {
        "singleQuote": true,
        "allowed": true,
      },
    },
    "TypeScript": {
[...]

PR #56316, opened 2026-05-10
recent_projects: Preserve git worktree when switching projects

terror:
Resolves (partially, worktree preservation only) https://github.com/zed-industries/zed/issues/55358

This diff fixes project switcher activation for projects backed by git worktrees. When selecting an already-open project group, Zed now reactivates the group’s last active workspace instead of reopening the group’s normalized main repository path. This preserves the user’s current linked worktree, such as a feature branch checkout, when switching away from a project and back through the project switcher.

I added a test that covers the case where a `main` checkout and linked worktree share the same project group identity, but the linked worktree was the last active workspace.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/5971ca23-74ad-40e3-afa8-a69f12db2627" controls></video> | <video src="https://github.com/user-attachments/assets/87beea22-dc6f-4052-aef1-47978742937c" controls></video> |

Release Notes:
- Fixed project switching to preserve the last active git worktree for already-open projects.


agu-z (changes requested):
Thanks for your contribution. I may be missing something, but it looks like this adds an inherent method that is only called from the test. Where is it wired into the project-switcher flow?

agu-z:
I'm closing this since it doesn't seem to change behavior, and I haven't gotten a response.

terror:
> I'm closing this since it doesn't seem to change behavior, and I haven't gotten a response.

Hi @agu-z!

I apologize I haven't had time to tend to this PR. You're right about the current diff. The method was originally wired into the project switcher, but an equivalent fix landed in #57321 and was later made remote-aware in #59272. I kept the upstream implementation during conflict resolution and missed removing the unused method and test (and then subsequently closing this PR). Thanks for catching it :)

PR #54481, opened 2026-04-22
fs: Polling backend support

probably-neb:
Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [ ] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #51340
Release Notes:

- Linux/WSL: Added polling filesystem watching support in order to support file watching on file systems that do emit events through `inotify`, for example `/mnt/c` in WSL, network filesystems, and FUSE mounts. Polling should be automatically chosen over `inotify` when necessary, however it can be manually chosen by setting the `ZED_FILE_WATCHER_MODE=poll` env var

PR #56345, opened 2026-05-10
Limit recursive poll file watching

karimknaebel:
Draft PR to highlight a file watcher issue seen on NFS-backed paths.

Zed switches these paths to the polling watcher. The poll watcher can then recurse from the worktree root and repeatedly walk large generated or ignored directories. In my case this kept the Zed server process busy and completely prevented the file tree and search from populating.

The important issue is that recursive polling appears too broad for this filesystem case.

### DIFF
--- a/crates/fs/src/fs_watcher.rs
+++ b/crates/fs/src/fs_watcher.rs
@@ -18,6 +18,15 @@ pub enum WatcherMode {
     Poll,
 }
 
+impl WatcherMode {
+    fn is_recursive(self) -> bool {
+        match self {
+            WatcherMode::Native => cfg!(any(target_os = "windows", target_os = "macos")),
+            WatcherMode::Poll => false,
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


karimknaebel:
Possibly related:
https://github.com/zed-industries/zed/pull/54481
https://github.com/zed-industries/zed/issues/38109

karimknaebel:
@probably-neb i think this might be related to recent fs watch changes from your side. let me know if i can help you in any way to reproduce this problem :)

karimknaebel:
@kubkon tagging you here since you reviewed https://github.com/zed-industries/zed/pull/54481

the main issue seems to be that recursively watching a large dir with the poll watcher (especially on NFS) completely overwhelms the remote server process to the point where the file tree doesn't populate, search doesn't work, etc.

ChristopherBiscardi:
closing since this never made it out of draft status, and the referenced issues/prs in comments have been merged/closed

PR #56350, opened 2026-05-10
agent: Add setting to auto-resolve external ACP permission requests

LivioGama:
Auto-resolves external ACP permission popups when enabled via setting.

Release Notes:

- Added auto-resolve setting for external ACP permissions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -278,6 +278,7 @@ dependencies = [
  "acp_thread",
  "action_log",
  "agent-client-protocol",
+ "agent_settings",
  "anyhow",
  "async-channel 2.5.0",
  "chrono",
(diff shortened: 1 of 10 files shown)


benbrandt:
Hi given that each agent manages their own permission settings, I think this setting will cause confusion, as we can't encforce the agent does a particular method

LivioGama:
Hi, Cursor does not.
It's currently impossible to have YOLO mode with Cursor without this pull request.
Unless I am missing something?

LivioGama:
CleanShot 2026-05-18 at 03 22 53
Well I guess one could argue that the PR should be addressed to Cursor ACP instead...
What do you think?
I didn't check the other CLI.

EDIT: yeah definitely. Let's go!
CleanShot 2026-05-18 at 03 26 33

EDIT: Oh. That's proprietary. I guess I will just use [this custom then](https://github.com/raphaelluethy/cursor-acp)

PR #56354, opened 2026-05-10
Add support for custom git commands in the git graph context menu

JosephTLyons:
This PR adds support to create custom git task that are triggered from the git graph context menu. Both Sublime Merge and Fork have custom command support, and I use custom commands I've built out frequently to speed up some of my tasks at Zed. Thus, I'd like to have support in Zed's git graph so I can use Zed even more!

It offers initial support for the following env variables:

- `ZED_GIT_SHA`
- `ZED_GIT_SHA_SHORT`
- `ZED_GIT_REPOSITORY_NAME`
- `ZED_GIT_REPOSITORY_PATH`

These are only populated in the git graph context.

This PR also introduces a the `git-command` tag, which is needed so the git graph can filter down to these custom git commands, and so other tasks aren't polluting the context menu.

An example would be (in the global `tasks.json`):

```json
{
  "label": "Branches containing commit: $ZED_GIT_SHA_SHORT",
  "command": "git",
  "args": ["branch", "-a", "--contains", "$ZED_GIT_SHA"],
  "tags": ["git-command"],
},
```

And then in the context menu:

SCR-20260511-mnfa

And the output in the terminal:

SCR-20260511-mnks

The awesome thing about using tasks is we get all the task infrastructure for free, such as history!

SCR-20260511-mnud

SCR-20260511-moch

And we have all the task configuration options too out of the box.

---

Right now, this only works with global tasks. It isn't clear how to shoe in support for worktree-specific tasks (`.zed/tasks.json`) in a clear way, as not all invocations of the git graph are in an area that has a clear worktree id, and so we sort of have to guess or fallback to something else. That can be a followup once it's more clear how we should cover that. Or maybe someone else has a better solution.

I chose to only ship with these 4 git-specific variables for now. The git graph also currently ONLY resolve those variables. We can adjust / add in more in follow up PRs.

Self-Review Checklist:

Release Notes:

- Added support for running global custom Git command tasks from the Git Graph commit context menu.


crates/git_graph/src/git_graph.rs
```diff
@@ -2031,6 +2032,107 @@ impl GitGraph {
...
+        let workspace = self.workspace.upgrade()?;
+        let project = workspace.read(cx).project().clone();
+
+        worktree_ids.into_iter().find(|worktree_id| {
+            project
+                .read(cx)
+                .worktree_for_id(*worktree_id, cx)
+                .is_some_and(|worktree| worktree.read(cx).is_visible())
+        })
+    }
```
JosephTLyons: I'm not sure the best way to handle the fact we could have multiple `worktree_ids` per repository

JosephTLyons: I cut scope from this for now, the feature will go out with global task support only. We can add support for worktree-based git commands in follow ups, as well as add in support for other environment variables we might want to have access to.

crates/git_graph/src/git_graph.rs
```diff
@@ -2031,6 +2032,82 @@ impl GitGraph {
...
+
+        if let Some(repository_name) = repository_name {
+            task_variables.insert(VariableName::GitRepositoryName, repository_name);
+        }
+
+        Some(TaskContext {
+            cwd: Some(repository_path),
+            task_variables,
+            ..TaskContext::default()
+        })
```
Anthony-Eid, line 2064: this is missing other task variables that could be useful

Anthony-Eid: We should really unify the task context code, so the task inventory can create it in one spot instead of having a lot of different task context creation functions. That's something we should address in a follow up PR.

docs/src/tasks.md
```diff
@@ -255,6 +255,34 @@ Hook tasks are resolved from the same global and worktree-local `tasks.json` fil
...
+```json [tasks]
+[
+  {
+    "label": "Branches containing commit: $ZED_GIT_SHA_SHORT",
+    "command": "git",
+    "args": ["branch", "-a", "--contains", "$ZED_GIT_SHA"],
+    "tags": ["git-command"]
+  }
+]
+```
```
Anthony-Eid, line 284: should this example use. the repository path env variable somewhere?

Anthony-Eid (approved):
Approved with some potential changes

PR #56363, opened 2026-05-11
editor: Auto-refresh Bookmarks tab when bookmarks change

### LINKED ISSUE #56268: New bookmarks do not live reload in "Bookmarks" tab
### Reproduction steps

New bookmarks do not appear in already opened "Bookmarks" tab.
Steps to reproduce:
1. add bookmark;
2. editor: view bookmarks (and leave "Bookmarks" tab opened);
3. go to file and add bookmark;
4. return to "Bookmarks" tab, there will be no new bookmark until you call "editor: view bookmarks" again. 

Current behavior: new bookmarks created after your opened "Bookmarks" tab do not appear in "Bookmarks" tab until you call "editor: view bookmarks" again.

Zed: 1.1.7
OS: macOS Sequoia 15.7.3
Memory: 32 GB

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

saberoueslati:
[## Context

Fixes a bug where the Bookmarks tab, opened via `editor: view bookmarks`, showed a stale snapshot and had to be reopened to display bookmarks added or removed while it was open.

Closes #56268

`view_bookmarks` built a one-time multibuffer and never subscribed to `BookmarkStore`. The store also calls `cx.notify()` while resolving serialized bookmarks into anchors, even though that bookkeeping does not represent a user-visible bookmark change. Observing all notifications would therefore feed resolution notifications back into the refresh mechanism and cause redundant cancellation and requeue work. The store now emits a dedicated `BookmarkStoreEvent::BookmarksChanged` from bookmark mutation paths, while `resolve_all` deliberately does not emit it.

Refresh state lives on a `BookmarksView` entity that holds the tab's multibuffer. Split editors already share that multibuffer, so they share the view as well. Each store event drives one reconciliation regardless of how many split editors display it. The view remains alive as long as any split retains it, so closing the original tab does not stop the remaining split from refreshing.

Each refresh reconciles excerpts per path instead of rebuilding the multibuffer: `set_excerpts_for_path` updates files that still contain bookmarks, while `remove_excerpts` removes files whose final bookmark was deleted. This follows the per-path reconciliation pattern used by project diff views.

`view_bookmarks` creates the tab synchronously and populates it asynchronously, following the existing patterns in `ProjectDiagnosticsEditor::deploy` and `ProjectSearchView::existing_or_new_search`. Repeated invocations focus the existing Bookmarks tab instead of opening duplicates.

Video of the manual test:

[Screencast from 2026-05-11 02-03-38.webm](https://github.com/user-attachments/assets/8ac97245-09ff-4a79-9634-20cb8dfbb7eb)

## How to Review

**`crates/project/src/bookmark_store.rs`**

Adds `BookmarkStoreEvent::BookmarksChanged`, emitted from bookmark mutation paths such as toggle, label edit, file rename or deletion, clear, and deserialization, but not from anchor resolution. It also adds `BookmarkStore::is_empty`.

**`crates/editor/src/bookmarks.rs`**

Adds `BookmarksView`, which holds the multibuffer, store subscription, and in-flight refresh task. It reconciles excerpts and emits `BookmarksViewEvent::Populated(ranges)` for per-editor highlights and cursor placement. It also caches the latest populated ranges and replays them when an editor attaches, allowing a split created after population to display bookmark highlights immediately.

**`crates/editor/src/editor.rs`**

Adds the shared `bookmark_view`, a per-editor subscription to its population events, and `bookmark_initial_selection_pending` for initial cursor placement. `Editor::clone` attaches the split to the shared view instead of creating another store subscription and refresh task.

**`crates/editor/src/editor_tests.rs`**

Adds five tests covering automatic addition and removal across multiple files, removal of a file's path after its final bookmark is deleted, initial cursor placement, a bookmark change racing initial population, splitting while population is pending, cached highlights in a late-attaching split, and continued refreshing after the original tab releases its view handle.

## Self-Review Checklist

Release Notes:

- Fixed Bookmarks tab not updating when bookmarks are added or removed while the tab is open](https://github.com/zed-industries/zed/pull/56363)


saberoueslati:
@ChristopherBiscardi I have a question regarding the PR I made, I also want to add the bookmarks tab closing automatically when the last bookmark is deleted to this PR, is it a sensible choice here?

I believe it is because it doesn't make sense to leave the bookmarks tab open empty

Can you get back to me with feedback or point me towards someone else I can tag who can give me feedback or should I open a discussion on this?

Thank you very much

crates/editor/src/bookmarks.rs
```diff
@@ -198,13 +215,53 @@ impl Editor {
...
         .detach();
     }
 
+    fn schedule_bookmark_refresh(
+        &mut self,
+        bookmark_store: Entity<BookmarkStore>,
+        cx: &mut Context<Self>,
+    ) {
+        self.bookmark_refresh_task = Some(cx.spawn(async move |this, cx| {
+            let Some(locations) = BookmarkStore::all_bookmark_locations(bookmark_store, cx)
```
SomeoneToIgnore: `all_bookmark_locations` → `resolve_all` → `cx.notify()` (`bookmark_store.rs:408`) fires the very observer that spawned the refresh, cancelling the in-flight task and scheduling a new one.

Seems wasteful to do that loop even though it seems to converge, can we do better?

crates/editor/src/bookmarks.rs
```diff
@@ -178,9 +178,26 @@ impl Editor {
         window: &mut Window,
         cx: &mut Context<Workspace>,
     ) {
+        let existing = workspace.panes().iter().find_map(|pane| {
```
SomeoneToIgnore: The subscription is only set after the async spawn completes, so N rapid `view_bookmarks` invocations before the first resolves open N tabs.

Seems now what we want?

crates/editor/src/bookmarks.rs
```diff
@@ -198,13 +215,53 @@ impl Editor {
                         MultibufferSelectionMode::First,
                         window,
                         cx,
-                    );
+                    ) else {
+                        return;
+                    };
+
+                    editor.update(cx, |editor, cx| {
+                        editor.bookmark_view_subscription =
```
SomeoneToIgnore: `clone_on_split` (`items.rs:827` → `Editor::clone`, `editor.rs:1820`) shares the multibuffer but not `bookmark_view_subscription`; split the Bookmarks tab, close the original, and the clone silently stops refreshing while `view_bookmarks` opens a fresh duplicate.

crates/editor/src/bookmarks.rs
```diff
@@ -198,13 +215,53 @@ impl Editor {
...
+            let Some(locations) = BookmarkStore::all_bookmark_locations(bookmark_store, cx)
+                .await
+                .log_err()
+            else {
+                return;
+            };
+
+            this.update(cx, |editor, cx| {
+                editor.buffer.update(cx, |multibuffer, cx| {
+                    multibuffer.clear(cx);
```
SomeoneToIgnore: Every store notify rebuilds all excerpts, O(total bookmarks) per toggle anywhere in the project.

Acceptable, but switching to `set_excerpts_for_path` per path + `remove_excerpts_for_path` for stale paths (the `ProjectDiff` pattern) would make it incremental and match precedent.

crates/editor/src/editor_tests.rs
```diff
@@ -28998,6 +28998,115 @@ async fn test_go_to_bookmark_with_out_of_order_bookmarks(cx: &mut TestAppContext
     assert_eq!(ctx.cursor_row(), 1, "Prev from row 5 should go to row 1");
 }
 
+#[gpui::test]
+async fn test_bookmarks_tab_auto_refresh(cx: &mut TestAppContext) {
```
SomeoneToIgnore, line 34189: No removal seem to be tested.

SomeoneToIgnore (commented):
Hey, sorry that no one came back earlier, I've finally crawled to this PR range...

Looks nice overall, I do not think there's any architectural guidance we need (overall the render-based approach seems odd as causes quite an indirection, but let's fix the bug and collect tests in this PR first, refactoring that needs another PR and stepping back...)

crates/editor/src/bookmarks.rs
saberoueslati: I replaced the generic `cx.observe()` subscription with a dedicated `BookmarkStoreEvent::BookmarksChanged` event.

`resolve_all` still calls `cx.notify()` for its internal anchor-resolution bookkeeping, but it no longer emits the bookmark-change event. The Bookmarks tab now refreshes only after an actual bookmark-store mutation, so a refresh cannot cancel and requeue itself.

crates/editor/src/bookmarks.rs
saberoueslati: This is now handled by creating and inserting the Bookmarks editor synchronously, with its subscription attached immediately. There is no `await` between checking for an existing tab and insertion, so repeated invocations focus the same tab instead of opening duplicates.

The async work now only populates that tab, following the pattern in `ProjectDiagnosticsEditor::deploy` and `ProjectSearchView::existing_or_new_search`.

crates/editor/src/bookmarks.rs
saberoueslati: I replaced the `clear`/rebuild approach with per-path reconciliation.

Each refresh now updates bookmarked paths through `set_excerpts_for_path`, which preserves unchanged excerpts, and removes paths that are no longer bookmarked with `remove_excerpts`. This avoids tearing down and recreating the whole multibuffer on every bookmark change, following the suggested incremental pattern.

crates/editor/src/editor_tests.rs
saberoueslati: Added removal coverage: the auto-refresh test now removes the second bookmark and verifies that the Bookmarks tab updates from two excerpts back to one.

saberoueslati:
@SomeoneToIgnore tried responding to every review comment, I'm obviously here for anything else needed to land this PR, thank you

saberoueslati:
@SomeoneToIgnore CI issue should be fixed now, can you run the CI pipeline again, thank you

crates/editor/src/editor_tests.rs
```diff
@@ -33791,6 +33791,556 @@ async fn test_empty_rename_is_no_op(cx: &mut TestAppContext) {
...
+            "main.rs": "line0\nline1\nline2\nline3\nline4\nline5\nline6\nline7\nline8\nline9\n",
+        }),
+    )
+    .await;
+    let project = Project::test(fs, [path!("/a").as_ref()], cx).await;
+    let window = cx.add_window(|window, cx| MultiWorkspace::test_new(project.clone(), window, cx));
+    let workspace = window
+        .read_with(cx, |mw, _| mw.workspace().clone())
+        .expect("window should still be open");
+    let mut visual_cx = VisualTestContext::from_window(*window, cx);
```
SomeoneToIgnore:
```suggestion
    let mut cx = VisualTestContext::from_window(*window, cx);
```

crates/editor/src/editor_tests.rs
```diff
@@ -33791,6 +33791,556 @@ async fn test_empty_rename_is_no_op(cx: &mut TestAppContext) {
...
+                    .items()
+                    .filter_map(|item| item.downcast::<Editor>())
+                    .find(|editor| editor.read(cx).bookmark_view_subscription.is_some())
+            })
+        })
+        .expect("bookmarks editor should exist after view_bookmarks");
+
+    let split_editor = bookmarks_editor.update_in(&mut visual_cx, |editor, window, cx| {
+        cx.new(|cx| editor.clone(window, cx))
+    });
```
SomeoneToIgnore: This test still passes with the subscription callback neutered, so it does not yet prove what its name claims.

The clone's initial `schedule_bookmark_refresh` from `Editor::clone` is still pending when the second toggle lands, so that initial refresh alone delivers the 2nd excerpt and the subscription is never exercised.
Draining the initial refresh first makes the test fail against a dead subscription and still pass against the real code (verified locally both ways).

```suggestion
    });
    visual_cx.run_until_parked();
```

crates/editor/src/editor_tests.rs
```diff
@@ -33791,6 +33791,556 @@ async fn test_empty_rename_is_no_op(cx: &mut TestAppContext) {
...
+    visual_cx.run_until_parked();
+
+    let excerpt_count = bookmarks_editor.update_in(&mut visual_cx, |editor, _window, cx| {
+        editor.buffer.read(cx).snapshot(cx).excerpts().count()
+    });
+    assert_eq!(
+        excerpt_count, 1,
+        "bookmarks editor should auto-refresh back to 1 excerpt after removing a bookmark"
+    );
+}
```
SomeoneToIgnore, line 34379: Optional: the `stale_paths` → `remove_excerpts` branch in `apply_bookmark_locations` is never exercised, because this file always keeps at least one bookmark.
Removing the last bookmark of a file is the only path that drops a `PathKey` entirely, let's update the test to actually verify this.

crates/editor/src/editor.rs
```diff
@@ -1829,6 +1835,14 @@ impl Editor {
         clone.needs_initial_data_update = self.enable_lsp_data;
         clone.enable_runnables = self.enable_runnables;
         clone.enable_code_lens = self.enable_code_lens;
+        if self.bookmark_view_subscription.is_some()
+            && let Some(bookmark_store) = clone.bookmark_store.clone()
+        {
+            clone.bookmark_view_subscription =
+                Some(clone.subscribe_to_bookmark_store(bookmark_store.clone(), window, cx));
+            clone.bookmark_initial_selection_pending = self.bookmark_initial_selection_pending;
+            clone.schedule_bookmark_refresh(bookmark_store, window, cx);
```
SomeoneToIgnore: The clone shares the original's `MultiBuffer` entity (`self.buffer.clone()` above), so N split Bookmarks tabs re-run `apply_bookmark_locations` N times per store event against the same multibuffer.

Wonder if we can do better somehow?

SomeoneToIgnore (commented):
Thanks, almost there!
A separate event is definitely way better than the previous approach, but we now need to update the PR description.

We need to make the tests better and ideally fix multiple stale refreshes (but if it takes a lot of changes maybe fine for now?).

crates/editor/src/editor_tests.rs
saberoueslati: Applied here and to the other four Bookmarks tab tests that used the same naming.

crates/editor/src/editor_tests.rs
saberoueslati: I confirmed it, the clone previously had its own `schedule_bookmark_refresh` pending, which delivered the excerpt without the subscription firing.

I addressed this more broadly by moving refresh ownership into a shared `BookmarksView` (see the `editor.rs` thread). Clones no longer schedule refreshes, and the initial population is drained before cloning. The second toggle therefore depends on the shared store subscription; stubbing its handler makes this test and `test_bookmarks_tab_auto_refresh` fail.

crates/editor/src/editor.rs
saberoueslati: I moved the subscription, the refresh task, and the excerpt reconciliation onto a new `BookmarksView` entity that owns the multibuffer, and clones share that handle the same way they already share the multibuffer. A store event now runs `apply_bookmark_locations` once, however many split tabs are open. `Editor` keeps only the per-tab pieces (line highlights, initial cursor), which come from a `Populated(ranges)` event.

The view also keeps the last ranges and replays them when an editor attaches, so a tab split after population gets its highlights right away instead of waiting for the next bookmark change.

crates/editor/src/editor_tests.rs
saberoueslati: The original single-file fixture could never exercise removal of a `PathKey`, so I added `other.rs` to `test_bookmarks_tab_auto_refresh`. The test bookmarks a line there, then removes the final bookmark from `main.rs` and verifies that the multibuffer paths change from `["main.rs", "other.rs"]` to `["other.rs"]`.

saberoueslati:
@SomeoneToIgnore updated the PR description and tried my best to act on the latest review, thank you

SomeoneToIgnore (approved):
Thank you.

Had to change a bunch of other things, as many corner cases (e.g. bookmarks tab split) were not covered yet.

That bloated a diff quite a while — would appreciate if you can give it a round of testing and check whether it works for you or any bugs worth fixing before the merge.

saberoueslati:
@SomeoneToIgnore did tests, FWIW it works as expected, resolved a conflict just now, can you run the CI pipeline one more time for a last verification ?, thank you for your time working with me on this, highly appreciated.

PR #56355, opened 2026-05-10
Add Auto-Resolve for non-conflicting merge regions

giovannipapini:
## Summary

- Adds Auto-Resolve for merge conflicts: handles every region where one side matches the diff3 merge base in a single, undoable transaction
- File-level banner at the top of conflicted buffers, with three states (resolvable / none / no diff3 base)
- Two new actions: `git::AutoResolveNonConflicting` (file scope) and `git::AutoResolveNonConflictingInProject` (workspace scope)

For each conflict region the algorithm compares ours, theirs, and base using strict equality. If only one side diverged from base, the other side is taken; if both sides made the identical change, that change is taken; otherwise the region is left for manual resolution. Inspired by IntelliJ's "Apply All Non-Conflicting Changes" and kdiff3's auto-merge.

Tracked in #51541.

### Out of scope (possible follow-ups)

- One-click "Enable diff3" affordance that runs `git config merge.conflictStyle zdiff3` (would expand the `GitRepository` trait surface; deferred to keep this PR narrow)
- Sub-hunk decomposition inside a single conflict region (kdiff3-style)
- Whitespace-only / regex-based auto-merge

## Test plan

- [x] Unit tests covering all four `AutoResolution` outcomes plus a mixed-buffer case (`crates/project/tests/integration/git_store.rs`)
- [x] Integration test verifying single-transaction undo
- [x] Manual smoke test on macOS — all three banner states, Cmd+Z reverts the batch
- [ ] Manual smoke test on Linux

Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments — N/A
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Release Notes:

- Added an Auto-Resolve action for merge conflicts that resolves regions where one side matches the diff3 base, leaving genuine conflicts for manual resolution

PR #56364, opened 2026-05-11
Auto-Resolve follow-ups: line-level decomposition, regex patterns, language-aware merge

giovannipapini:
Stacks on #56355. The bottom seven commits in this PR are the ones already under review there; the remaining commits implement four layered follow-ups to Auto-Resolve. When #56355 merges, the bottom commits will drop out and this PR will rebase down to the follow-ups alone.

### Layer 1: line-level sub-hunk decomposition (commit `eeee1a7696`)

Auto-Resolve previously gave up on a whole region whenever both sides touched it, even if their changes were on disjoint lines. With line-level decomposition, we re-diff each side against the diff3 base, cluster overlapping hunks together, and emit the resolvable lines as plain text while wrapping any remaining clusters in smaller `<<<<<<<` / `|||||||` / `=======` / `>>>>>>>` blocks. The banner and toast now distinguish fully-resolvable regions from partial simplifications.

### Layer 2: regex Auto-Resolve patterns (commit `2cf3c46fc7`)

New `git_panel.auto_resolve_patterns` setting:

```json
{
  "git_panel": {
    "auto_resolve_patterns": [
      { "pattern": "^version = \".*\"$", "take": "theirs" }
    ]
  }
}
```

Fires only when a sub-conflict is exactly one line on each side and both lines match the pattern. Multi-line sub-conflicts are intentionally skipped so a too-broad pattern can't silently discard real changes.

### Layer 3: language-aware structural merge (commits `de19aaf99b` → `ec13dfb48c` → `b8ecfe6c36`)

Re-parses the buffer with the language's tree-sitter grammar in three reconstructed views (with-base / with-ours / with-theirs) and runs a per-language `merges.scm` query file. When a conflict sits inside an `@merge.set` (unordered) or `@merge.ordered_list` (ordered) container, the engine merges items by `@merge.key` identity:

- Same item changed by only one side → take the change
- Both sides changed it identically → take it
- Both sides changed it differently → defer
- Both sides removed the same item → take the deletion
- One side removed an item the other modified → defer
- Both sides added an item with the same key → defer if texts differ, else emit once
- Cross-region key collisions (ours adds `X` in region A, theirs adds `X` in region B) → defer both regions to avoid duplicate definitions

`merges.scm` queries ship for **Rust, TypeScript, Python, Go, JSON, and YAML**. New languages opt in by dropping a query file alongside `highlights.scm` (mirrors the existing extension surface). When a structural merge defers, `DeferReason` carries a precise diagnostic (`BothModifiedDifferently`, `DeleteVsModify`, `BothAddedDifferently`, `CrossRegionKeyCollision`, etc.).

`@merge.key.normalized` exists as a fallback identity that collapses whitespace before comparison.

Three full parses per file → one full parse + two incremental reparses with `Tree::edit()` `InputEdit` deltas.

### Layer 4: inline UI hints + toast breakdown (commit `b8ecfe6c36`, simplified in `39d6810f46`)

After Auto-Resolve runs, each resolved region gets a one-line inline annotation describing the method (`structural set merge` / `structural ordered merge` / `line-level merge` / `partially simplified`) with a per-hint ✕ dismiss button. The toast now reads e.g.:

> Auto-resolved 5 of 7 conflicts (3 structural, 2 line); simplified 1 more; 2 remain — both branches modified `process_event`

## Test coverage

36 integration tests in `crates/project/tests/integration/git_store.rs` covering:
- Region-level fast paths (take ours / take theirs / identical / no base / genuine)
- Line-level decomposition (partial / interleaved / leaves smaller sub-conflict / insertion conflict)
- Regex patterns (version string resolution / multi-line skip)
- Structural set merge (disjoint imports / same import added differently / unilateral deletion / both delete the same item / cross-region collision)
- `@merge.key` cases (one-side modification / delete-vs-modify defer / both-modify defer)
- Structural ordered_list (disjoint match arms)
- One smoke test per shipped language (Rust, Python, Go, JSON, YAML)
- `@merge.key.normalized` whitespace fallback

## Discussion

Original discussion: #51541

Release Notes:

- Improved Auto-Resolve to handle disjoint line-level changes, regex-based version-string conflicts, and language-aware structural merges of import lists, named items, and ordered containers; resolved regions now get an inline hint describing how they were merged


smitbarmase:
Closing since this is follow up of https://github.com/zed-industries/zed/pull/56355 which is closed.

PR #56386, opened 2026-05-11
ep: Generate request IDs on the client

osyvokon:
This change lays the groundwork for canceling in-flight requests. Without it, we always wait for a request to complete, even when we already know its result won't be used.

Release Notes:

- N/A


probably-neb:
@zed-industries/approved

PR #56418, opened 2026-05-11
editor: Show jump indicator for nearby edit predictions in multi-buffers

probably-neb:
When an edit prediction's edits fall outside (or partially outside) the visible excerpts of a multi-buffer, we previously discarded the prediction entirely. This PR classifies such predictions as jumps and shows the jump indicator with a directional arrow. On accept:

- If every hidden edit is within 10 rows of an existing excerpt for that buffer, the excerpts are expanded permanently and additively (existing excerpts and path keys are preserved), the cursor moves to the edit, and the prediction then renders normally (inline preview in eager mode, subtle indicator in subtle mode) so a second accept applies it.
- Otherwise, the target buffer is opened in a separate editor at the edit position.

<details>
<summary>Screenshots</summary>

A prediction targets a hidden edit above the excerpt, so the jump pill appears:

edit_prediction_multibuffer_top_excerpt_expands_upward_before_accept

Accepting expands the excerpt permanently, moves the cursor, and the prediction renders as a normal inline preview:

edit_prediction_multibuffer_top_excerpt_expands_upward_after_jump

Accepting again applies the edit:

edit_prediction_multibuffer_top_excerpt_expands_upward_after_accept

</details>

Also included:

- A `MultiBuffer Jumps` component preview (new `Edit Predictions` component scope) rendering real editors for each state: fully visible hunk, top/bottom/middle excerpt expansion, boundary-line jump, far-from-excerpt open-on-accept, adjacent multi-edit cluster, and different-file jump.
- Visual regression tests covering the same scenarios plus the jump-accept flow (before / after jump / after apply) via `zed_visual_test_runner`, sharing the preview case setup.
- Jump indicator fixes surfaced by the visual tests: end-of-line jump pills now always carry an up/down arrow computed in display coordinates (previously arrows were driven by horizontal overflow and buffer-point rows, which are wrong across excerpts, and could point the wrong way when the cursor sat on the boundary-snapped line), the `MoveOutside` pill no longer overlaps the excerpt header/scrollbar, and pills are vertically centered on their line.

Release Notes:

- Improved edit predictions in multi-buffers: predictions that make edits just outside of an excerpt are no longer discarded. Instead a jump indicator is shown, which expands the excerpt to preview the prediction upon hitting `tab`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5743,6 +5743,7 @@ dependencies = [
  "client",
  "clock",
  "collections",
+ "component",
  "convert_case 0.11.0",
  "ctor",
  "dap",
--- a/crates/component/src/component.rs
+++ b/crates/component/src/component.rs
@@ -302,6 +302,8 @@ pub enum ComponentScope {
     #[strum(serialize = "Data Display")]
     DataDisplay,
     Editor,
+    #[strum(serialize = "Edit Predictions")]
+    EditPredictions,
     #[strum(serialize = "Images & Icons")]
     Images,
     #[strum(serialize = "Forms & Input")]
--- a/crates/editor/Cargo.toml
+++ b/crates/editor/Cargo.toml
@@ -39,6 +39,7 @@ breadcrumbs.workspace = true
 client.workspace = true
 clock.workspace = true
 collections.workspace = true
+component.workspace = true
 convert_case.workspace = true
 dap.workspace = true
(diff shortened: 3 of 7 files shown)


SomeoneToIgnore:
To keep the PR queue cleaner, will close this PR for now, as it's a draft with no changes over 3 weeks.

PR #56420, opened 2026-05-11
Enable CLI socket for LocalPath builds on macOS

bbarker:
# enable CLI socket for LocalPath builds on macOS

## Summary

When building Zed from source on macOS and using the CLI shim with `--zed /path/to/binary`, the `-a` (add to workspace) flag always spawns a new process instead of connecting to the running instance. This is because `listen_for_cli_connections` was gated to Linux/FreeBSD only, and the macOS `LocalPath` launch path had no socket check.

This extends the existing Unix socket mechanism to macOS so the CLI can connect to a running LocalPath instance, matching the Linux behavior.

## Changes

- `open_listener.rs`: Expand `listen_for_cli_connections` cfg gate to include `target_os = "macos"`
- `main.rs` (zed): Call `listen_for_cli_connections` in the macOS single-instance block
- `main.rs` (cli): Add socket check in macOS `Bundle::LocalPath::launch` before spawning

## Self-Review Checklist

- [ ] ~~The content is consistent with the UI/UX checklist~~ (no UI changes)
- [ ] Tests cover the new/changed behavior — the existing Linux socket path is also untested; happy to try and add an integration test if requested. I did test locally and observed that doing `zed -a ...` no longer starts a new process/window.
- [x] Performance impact has been considered and is acceptable (single socket connect attempt on launch)

Release Notes:

- Fixed CLI `-a`/`--add` flag spawning a new window instead of adding to the existing instance on macOS when using `--zed` with a local binary path.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1357,6 +1357,20 @@ mod mac_os {
                 }
 
                 Self::LocalPath { executable, .. } => {
+                    // Check if a running instance is listening on the socket (like Linux does).
+                    let data_dir = user_data_dir
+                        .map(PathBuf::from)
+                        .unwrap_or_else(|| paths::data_dir().clone());
+                    let sock_path = data_dir.join(format!(
+                        "zed-{}.sock",
+                        *release_channel::RELEASE_CHANNEL_NAME
+                    ));
+                    let sock = std::os::unix::net::UnixDatagram::unbound()?;
+                    if sock.connect(&sock_path).is_ok() {
+                        sock.send(url.as_bytes())?;
+                        return Ok(());
+                    }
+
                     let executable_parent = executable
                         .parent()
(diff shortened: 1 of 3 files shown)


crates/cli/src/main.rs
```diff
@@ -1336,6 +1336,20 @@ mod mac_os {
...
                 Self::LocalPath { executable, .. } => {
+                    // Check if a running instance is listening on the socket (like Linux does).
+                    let data_dir = user_data_dir
+                        .map(PathBuf::from)
+                        .unwrap_or_else(|| paths::data_dir().clone());
+                    let sock_path = data_dir.join(format!(
+                        "zed-{}.sock",
+                        *release_channel::RELEASE_CHANNEL_NAME
+                    ));
+                    let sock = std::os::unix::net::UnixDatagram::unbound()?;
```
bbarker: may be safer to fall through to launching a new process instead of propagating the error?

PR #56429, opened 2026-05-11
GitHub board automation: remove redundant event handling

zelenenka:
We shouldn't trigger this workflow on pull_request_review.submitted because (1) two out of three possible cases here are already handled by the build-in workflow automation of the project board itself, and (2) this takes us into the territory of workflow runs that require approval.

Release Notes:

- N/A


probably-neb:
@zed-zippy approve

PR #8437, opened 2024-02-26
linux: fix invalid cross-device link error

sjet47:
This PR fix the "invalid cross-device link" error occurred in linux when trying to write the settings file atomically, like when click the "Enable vim mode" checkbox at first start.

```plain
[2024-02-26T22:59:25+08:00 ERROR util] .../zed/crates/settings/src/settings_file.rs:135: Failed to write settings to file "/home/$USER/.config/zed/settings.json"

Caused by:
0: failed to persist temporary file: Invalid cross-device link (os error 18)
1: Invalid cross-device link (os error 18)
```

Currently the `fs::RealFs::atomic_write()` method write to a temp file created with `NamedTempFile::new()` and then call `persist()` method to write to the config file path, which actually do a `rename` syscall under the hood. As the [issue](https://github.com/Stebalien/tempfile/issues/245) said

> `NamedTempFile::new()` will create a temporary file in your system's temporary file directory. You need `NamedTempFile::new_in()`.

The temporary file directory in linux is in `/tmp`, which is mounted to `tmpfs` filesystem, and in most case(all case I guess) `$HOME/.config/zed` is mounted to a different filesystem. And the `rename` syscall between different filesystems will return a `EXDEV` errno, as described in the man page [rename(2)](https://man7.org/linux/man-pages/man2/renameat2.2.html):

```plain
       EXDEV  oldpath and newpath are not on the same mounted
              filesystem.  (Linux permits a filesystem to be mounted at
              multiple points, but rename() does not work across
              different mount points, even if the same filesystem is
              mounted on both.)
```

And as the issue above said, use a different temp dir with `NamedTempFile::new_in()` for linux platform might be a solution, since the `rename` syscall provides atomicity.

Release Notes:
 - Fix `settings.json` save failed with invalid cross-device link error in linux

PR #56437, opened 2026-05-11
extension_host: Fix cross-device extension installs

PiasekDev:
This fixes extension installation when Zed's cache directory and data directory are on different filesystems or mount points. On my Linux system, `~/.cache` is a separate btrfs subvolume so it can be excluded from system snapshots, while `~/.local/share` is on the main home subvolume. Installing extensions failed because extension archives were unpacked under `paths::temp_dir()` and then renamed into the installed extensions directory.

The error showed up as:
```text
2026-05-11T17:08:57+02:00 INFO  [extension_host] installing extension dockerfile latest version
2026-05-11T17:08:58+02:00 ERROR [crates/extension_host/src/extension_host.rs:842] Invalid cross-device link (os error 18)
```

Linux rename(2) returns EXDEV (os error 18) across filesystem boundaries. This is the same class of issue addressed for settings writes in #8437 and later generalized in 2b3e453d2f3abaaa47c9def4ce5fc63fc017af03. The regression was introduced by #54355, which made extension updates safer by unpacking into a temporary directory before renaming into place.

This change creates the temporary unpack directory inside the installed extensions directory instead. That keeps the final rename on the same filesystem as the destination while preserving the existing direct-unpack fallback if creating the temporary directory fails. A possible alternative would be to create a dedicated temporary install directory under `extensions/`, alongside `extensions/installed/`. I chose, however, to keep the temporary directory directly beside the final destination because that matches the destination-local temp-file approach used in the earlier fixes referenced above.

Self-Review Checklist:

No separate issue. The change is small and includes the reproduction context here. I did not add a regression test because the failure depends on cache and data directories being on different filesystems/subvolumes; this was verified manually by reproducing EXDEV when installing extensions with `~/.cache` on a separate btrfs subvolume, then confirming the patch fixes extension installation.

Release Notes:

- Fixed installing extensions when Zed's cache and data directories are on different filesystems.

### DIFF
--- a/crates/extension_host/src/extension_host.rs
+++ b/crates/extension_host/src/extension_host.rs
@@ -117,6 +117,7 @@ pub struct ExtensionStore {
     pub reload_tx: UnboundedSender<Option<Arc<str>>>,
     pub reload_complete_senders: Vec<oneshot::Sender<()>>,
     pub installed_dir: PathBuf,
+    pub staging_dir: PathBuf,
     pub outstanding_operations: BTreeMap<Arc<str>, ExtensionOperation>,
     pub index_path: PathBuf,
     pub modified_extensions: HashSet<Arc<str>>,
@@ -246,6 +247,7 @@ impl ExtensionStore {
         let work_dir = extensions_dir.join("work");
         let build_dir = build_dir.unwrap_or_else(|| extensions_dir.join("build"));
         let installed_dir = extensions_dir.join("installed");
+        let staging_dir = extensions_dir.join("staging");
         let index_path = extensions_dir.join("index.json");
 
         let (reload_tx, mut reload_rx) = unbounded();
@@ -254,6 +256,7 @@ impl ExtensionStore {
             proxy: extension_host_proxy.clone(),
             extension_index: Default::default(),
             installed_dir,
+            staging_dir,
             index_path,
             builder: Arc::new(ExtensionBuilder::new(builder_client, build_dir)),
             outstanding_operations: Default::default(),
@@ -708,6 +711,7 @@ impl ExtensionStore {
         cx: &mut Context<Self>,
     ) -> Task<Result<()>> {
         let extension_dir = self.installed_dir.join(extension_id.as_ref());
+        let staging_dir = self.staging_dir.clone();
         let http_client = self.http_client.clone();
         let fs = self.fs.clone();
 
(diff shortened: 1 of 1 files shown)


MrSubidubi (changes requested):
Thanks for catching and looking into this that thorougly!

`installed_dir` is watched and regularly used to build the index of installed extensions, so I do not think this is a good place for this to be unpacked in. Lets go with the

> A possible alternative would be to create a dedicated temporary install directory under extensions/, alongside extensions/installed/ 

please, as I think this will be better in this scenario

PiasekDev:
Okay, in that case would you prefer this to become an explicit extension-store directory, e.g. `extensions/install_tmp/`, stored on `ExtensionStore` alongside `installed_dir`, or should we just derive it locally from `installed_dir.parent()`?

My current thinking is to use `extensions/install_tmp/<random-temp-dir>`: `install_tmp` would be a stable staging root next to `installed/`, while each install still gets its own `TempDir` inside it so concurrent installs do not share a directory. On success the inner temp dir is renamed into `installed/<extension-id>`, and on failure it is cleaned up by `TempDir`.

I would probably lean towards adding an explicit field for this path to `ExtensionStore` to avoid the awkward parent lookup. Also, since naming things is famously the hard part, does `install_tmp` sound right to you, or would you prefer something else?

MrSubidubi:
> Okay, in that case would you prefer this to become an explicit extension-store directory, e.g. extensions/install_tmp/, stored on ExtensionStore alongside installed_dir, or should we just derive it locally from installed_dir.parent()?

First sounds good!

> My current thinking is to use extensions/install_tmp/<random-temp-dir>: install_tmp would be a stable staging root next to installed/, while each install still gets its own TempDir inside it so concurrent installs do not share a directory. On success the inner temp dir is renamed into installed/<extension-id>, and on failure it is cleaned up by TempDir.

Yep that too sounds reasonable!

> I would probably lean towards adding an explicit field for this path to ExtensionStore to avoid the awkward parent lookup. Also, since naming things is famously the hard part, does install_tmp sound right to you, or would you prefer something else?

Makes sense, although I am not too big a fan of the name. Something that goes more into the direction that this is more a directory for staging the downloads (e.g. `cache` or something more in that direction with `_tmp` or `tmp_`) I'd prefer. Not insisting on any of these and not too bad in any case, since most users should never get to see this, but yeah

PiasekDev:
PR updated. I went with `staging`, since it lines up with the existing single-word extension-store directories (`work`, `build`, `installed`) and still describes the directory’s role before the final rename.

MrSubidubi (approved):
Thank you again for the find and the quick follow-ups. Awesome first contribution! 🎉

MrSubidubi:
... and catching this before it even went to preview - absolute hero, thank you so much! ❤️

PR #56452, opened 2026-05-11
markdown: Copy selection as rich text (HTML + plain text)

aahlijia:
When copying a selection from the Markdown Preview, Zed now writes both plain text and HTML to the system clipboard. Paste targets that understand rich text (Notion, Google Docs, Apple Notes, email clients) will receive formatted output; plain-text targets are unaffected.

The implementation adds a new ClipboardEntry::Html(String) variant to GPUI's clipboard abstraction and wires it up across all three platforms:

- macOS: writes public.html to the NSPasteboard after plain text
- Linux X11: writes a second ClipboardData entry with the text/html atom (which was already stubbed out but commented in the codebase)
- Linux Wayland: advertises text/html as an offered MIME type and serves HTML bytes when that type is requested
- Windows: registers the HTML Format clipboard format and writes data with the required CF_HTML header

HTML is generated from the selected markdown source using pulldown_cmark, which was already a workspace dependency. The html feature flag is enabled on the workspace-level dependency to make the html module available.

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Added rich text (HTML) copy support to the Markdown Preview, formatted content is now preserved when pasting into Notion, Google Docs, Apple Notes, and other rich-text targets

https://github.com/user-attachments/assets/c1fd3e63-33f4-4293-86c6-36cd40e93422


jeorb:
- I built and tested Zed with this PR on Fedora 44 under Cosmic & KDE
- What works
  - Basic rich-text copy-paste from the Markdown preview to LibreOffice Writer & Chrome
  - Headings
  - Bold
  - Italic
  - Lists
  - Links (some types)
- What doesn't work
  - Select-all
  - Links (some types)
  - Images
  - Strike-through (often)
  - Underscore
  - Headings at the beginning of a selection
  - Tables
  - Code & syntax highlighting
  - Sometimes markdown source is pasted instead of rich text when selecting near or including code blocks

[zed-rich-copy-paste.webm](https://github.com/user-attachments/assets/0474e71c-b2b3-4a85-9129-ec9246c8c2ad)

jeorb:
I built and tested Zed with this PR on Fedora 44 under KDE.

- What works
  - Rich-text copy-paste from the Markdown preview to LibreOffice Writer & Chrome
  - Almost everything copies perfectly to rich text editors in Chrome
  - Most things copy well to LibreOffice Writer
  - Select-all
  - Drag to select
  - Select partial elements
  - Headings
  - Bold
  - Italic
  - Lists
  - Links (all types except bare unmarked links like http://www.example.com)
  - Images
  - Strike-through
  - Tables
  - Code blocks & syntax highlighting
- What doesn't work
  - Some images in LibreOffice writer
  - Bare unmarked links (for example: http://www.example.com)

The code changes are minimal and idiomatic and I hope to see this feature in Zed soon.

[zed-rich-copy-paste-to-libre-office.webm](https://github.com/user-attachments/assets/79b35373-d502-4dbb-84b5-feefdf8ae1eb)

aahlijia:
Thanks for the thorough test pass! Really appreciate your test coverage @jeorb 

Both items you mentioned under "what doesn't work" are addressed in the latest changes:

- **Bare unmarked links** (e.g. 'http://www.example.com'): the preview already linkifies these visually via a separate pass over pulldown's parser output, but 'selection_html' (the copy path) wasn't applying that same pass, so bare URLs copied as plain text. It now runs the same linkify-based autolinking on copy, so they paste as real '<a href>' links.

- **Some images not appearing in LibreOffice Writer**: this was because copy previously emitted a bare img tag for remote images. Chrome fetches that URL live when you paste, but Writer generally doesn't fetch remote resources on paste, so the image just doesn't show up. Copy now kicks off a background fetch of any 'http(s)' images in the selection and upgrades the clipboard to inline them as base64 'data:' URIs once downloaded, so Writer (and any other paste target that doesn't fetch remote 'src's) gets the actual image bytes instead of a dead link. A generation counter guards against that background upgrade clobbering the clipboard if you've made a newer copy in the meantime.

Everything else you listed as working should be unaffected. 

Built and tested on Windows 11 with the same scenarios.

PR #56454, opened 2026-05-11
git_panel: Improve empty states

danilo-leal:
This PR makes the empty states of the Git panel a little bit more polished by making font sizes and button placement more consistent. Also ended up cleaning up the `panel_button` abstraction as that felt a bit unnecessary. We're just using the buttons directly like we do in other places in the codebase.

Release Notes:

- N/A


probably-neb:
@zed-zippy approve

PR #56462, opened 2026-05-11
ci: Improve nextest config

MrSubidubi:
As noted in https://nexte.st/docs/configuration/reference/?h=list+wrapper#default-configuration, `immediate-final` is the recommended setting for `failure-output` for large test suites, as is definitely the case for this repository.

Furthermore, allow one flaky run for the ocasionally flaky tests that we do have. We will add some tracking for which tests are flaky at a later point. 

Release Notes:

- N/A

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,5 +1,7 @@
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 1 }
+failure-output = "immediate-final"
+retries = 1
 
 [test-groups]
 sequential-db-tests = { max-threads = 1 }


osiewicz (changes requested):
I recall that @SomeoneToIgnore had strong objections to `immediate-final`, as we've tried it before: https://github.com/zed-industries/zed/pull/39643

(from private convo):
> ...I do not see the list of what's failed, and have to scroll for eternity for another failure.

MrSubidubi:
Yeah I think that is a valid point, hadn't thought of it like that. Optimally, there would be another variant.

Will close for the time being.

PR #56469, opened 2026-05-11
Fix various activity indicator issues

MrSubidubi:
Follow-up to #54791 with some fixes to the activity indicator which resulted in

- the button no longer being clickable when it should
- the indicator showing a spinner instead of the proper symbol when downloading or when it should show a warning symbol

Release Notes:

- Fixed some issues where the activity indicator would show the wrong icon and not be clickable on errors present.


danilo-leal (approved):
🎉

SomeoneToIgnore:
@zed-industries/approved

PR #56474, opened 2026-05-11
bedrock: Add system-prompt cache anchor on caching-capable models

rtfeldman:
The Bedrock Converse API supports placing `CachePoint` blocks inside the `system` field, but we were sending the system prompt as a single `SystemContentBlock::Text`, which leaves the system tokens dependent on whatever message-level breakpoint happens to fall within the 20-block lookback window.

This widens `bedrock::Request.system` from `Option<String>` to `Vec<BedrockSystemContentBlock>` and has `into_bedrock` emit `[Text(system), CachePoint(Default)]` whenever the model supports prompt caching. The system prompt now anchors its own cache prefix, on top of the existing tool-list anchor and per-message breakpoint, so a stable system prompt keeps producing cache hits even when earlier conversation turns change.

Bedrock does not support automatic caching or the 1-hour TTL, so the default 5-minute ephemeral cache is the only option for this provider.

Release Notes:

- Improved Bedrock prompt cache utilization by anchoring the system prompt as its own cache prefix

### DIFF
--- a/crates/bedrock/src/bedrock.rs
+++ b/crates/bedrock/src/bedrock.rs
@@ -84,10 +84,8 @@ pub async fn stream_completion(
 
     response = response.inference_config(inference_config);
 
-    if let Some(system) = request.system {
-        if !system.is_empty() {
-            response = response.system(BedrockSystemContentBlock::Text(system));
-        }
+    for system_block in request.system {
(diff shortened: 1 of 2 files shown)


MartinYe1234 (approved):
Tested

Issue #55872, opened 2026-05-06
Tracking issue for better fuzzy search


yara-blue:
This is a tracking issue for everything related to improving our fuzzy search feature.
Previous discussion: https://github.com/zed-industries/zed/issues/7586

### About tracking issues

Tracking issues are something new we are trying at Zed. They are used to record the overall progress of implementation. They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions. 

A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature. Instead open a new issue for a bug report or a [github discussion](https://github.com/zed-industries/zed/discussions) for large scale discussions and design. Then post a comment here with the link. Discussion comments here will get marked as off-topic or deleted. Repeated discussions on the tracking issue may lead to the tracking issue getting locked.

The process and text above are adapted from the [rust project](https://github.com/rust-lang/rust/tree/main) and are therefore covered under Apache-2.0 and MIT.

### Steps

- [x] Add a new crate `nucleo_fuzzy` for fuzzy path search
- [x] Extend `nucleo_fuzzy` to support strings 
- [x] Use nucleo fuzzy for a few pickers
- [ ] Use nucleo_fuzzy for all pickers
- [ ] Remove the fuzzy crate

Note: We will extend nucleo_fuzzy's use slowly, PR's switching over a few more pickers are very welcome!

### Unresolved Questions
None

### Bugs/Issues
- https://github.com/zed-industries/zed/issues/55195

### Implementation history 
- [x] https://github.com/zed-industries/zed/pull/51164
- [x] https://github.com/zed-industries/zed/pull/54123
- [x] https://github.com/zed-industries/zed/pull/54371
- [x] https://github.com/zed-industries/zed/pull/54112
- [x] https://github.com/zed-industries/zed/pull/55264
- [x] https://github.com/zed-industries/zed/pull/54143
- [x] https://github.com/zed-industries/zed/pull/54944

### In progress
- [ ] https://github.com/zed-industries/zed/pull/55923
- [ ] https://github.com/zed-industries/zed/pull/56477

yara-blue:
If anyone wants to make this more organized by figuring out all the pickers that still need to be made fuzzy and grouping them so we can divide an conquer. Go ahead and list your groups here then ill add them to the steps.

yara-blue:
[...]

PR #56477, opened 2026-05-11
outline: Switch the outline search to use fuzzy_nucleo

feitreim:
Nucleo tracking issue: https://github.com/zed-industries/zed/issues/55872

This PR switches the outline picker to use `fuzzy_nucleo`. 

with that I refactored the outline fuzzy picking to take advantage of nucleo and its multi-atom query support. The previous implementation had a lot going on specifically to work around the lack of multiple atoms, basically taking matters into its own hands to accomplish the same goals within the constraints it had. 

Instead of having two lists of candidates that we have to run the query against, we just run a single query and take advantage of the fact that nucleo chooses the matches that are the furtherest towards the end of the haystack to implement the same ancestor/leaf filtering. It retains the leaf only matching on single atom queries.

video:

https://github.com/user-attachments/assets/64baa8d7-fd77-452c-86d1-e08561422d85

Self-Review Checklist:

Release Notes:

- outline: switch to fuzzy_nucleo


crates/language/src/outline.rs
```diff
@@ -1,31 +1,66 @@
...
 #[derive(Debug)]
 pub struct Outline<T> {
     pub items: Vec<OutlineItem<T>>,
     candidates: Vec<StringMatchCandidate>,
-    pub path_candidates: Vec<StringMatchCandidate>,
-    path_candidate_prefixes: Vec<usize>,
+    /// For each item, the byte offset within `candidates[i].string` where the
+    /// item's own leaf text starts. Anything before this offset is ancestor
+    /// path text used purely as match context.
+    leaf_offsets: Vec<usize>,
```
yara-blue, line 15: What is in the StringMatchCandidate? Is it the full path to each item? Might help future readers and me if we document that.

crates/language/src/outline.rs
```diff
@@ -160,89 +185,101 @@ impl<T> Outline<T> {
...
-                            positions: Default::default(),
-                            string: Default::default(),
-                        },
-                    );
-                    cur_depth -= 1;
-                }
+/// Interleaves synthetic [`OutlineSearchEntry::Ancestor`] rows before each match so callers
+/// can render the parent chain as tree context above the match.
+///
+/// `matches` must be sorted ascending by `candidate_id` (which is what [`Outline::search`]
```
yara-blue: Let's add a debug_assert for this condition

crates/outline/src/outline.rs
```diff
@@ -346,10 +348,14 @@ impl PickerDelegate for OutlineViewDelegate {
...
+                            entry
+                                .as_match()
+                                .filter(|m| !m.positions.is_empty())
+                                .map(|m| (ix, m.score))
+                        })
                         .max_by(|(ix_a, a), (ix_b, b)| {
-                            OrderedFloat(a.score)
-                                .cmp(&OrderedFloat(b.score))
-                                .then(ix_b.cmp(ix_a))
+                            OrderedFloat(*a).cmp(&OrderedFloat(*b)).then(ix_b.cmp(ix_a))
```
yara-blue: We now have total_cmp: https://doc.rust-lang.org/std/primitive.f32.html#method.total_cmp

crates/language/src/outline.rs
```diff
@@ -1,31 +1,66 @@
...
     pub annotation_range: Option<Range<T>>,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub struct SymbolPath(pub String);
+pub struct SymbolPath(pub SharedString);
+
+/// Result of [`Outline::search`]. Real fuzzy matches are `Match`; `Ancestor`
+/// rows are synthetic entries pointing at parent items, included so callers
+/// can render tree context above each match without those rows competing for
```
yara-blue: This last bit is unclear to me, could you expand on that? What does competing look like? Are `Ancestors` there just to render the tree leading to an item?

yara-blue (changes requested):
I was out for a week and it has taken a bit to catch up with everything, apologies. 
I really appreciate you putting all this effort into fuzzy search!!

Now on to the review, I'm struggling a little in fully understanding all the needed changes. I've left some comments explaining my confusion, do you see a way to rework this (renames/comments) to make it easier to read? I'm not saying it is bad code! I'm saying search is hard, especially when trees get involved :)

crates/outline/src/outline.rs
feitreim: switched this over

crates/language/src/outline.rs
feitreim: Added a comment just explaining that candidate contains the whole path of an object, then slightly reworded the leaf_offsets comment to explain it better with that context.

crates/language/src/outline.rs
feitreim: Reworded it to better explain the purpose: Show the ancestor even when the ancestor was otherwise filtered out due to not matching anything in the query.
and the benefit: makes downstream tasks easier, like avoiding the auto-selection.

and yes their purpose is to render the tree leading to an item.

feitreim:
Okay so I fixed the typo and the unused dependency thing. but the clippy failure/test failure case

```
error[E0308]: mismatched types
   --> crates/language/src/outline.rs:319:19
    |
319 |             text: "completion".to_string(),
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^ expected `SharedString`, found `String`
    |
help: call `Into::into` on this expression to convert `std::string::String` into `gpui::SharedString`
    |
319 |             text: "completion".to_string().into(),
    |                                           +++++++
```

straight up does not exist. the line where I have something very similar happening, I already use `.into()`. so uh, hopefully that goes away?

yara-blue:
> straight up does not exist. the line where I have something very similar happening, I already use `.into()`. so uh, hopefully that goes away?

we'll see :)

yara-blue:
yeah this is kinda crazy, lets see if I get this locally

yara-blue (approved):

PR #56483, opened 2026-05-12
settings: Apply VS Code minimap default when importing settings

### LINKED ISSUE #56297: Minimap setting not imported from VSCode if it was on by default
### Reproduction steps

1. Have VSCode installed with minimap related settings as default (`"editor.minimap.autohide": "none"` and `"editor.minimap.enabled": true`), **but not explicitly set in the user settings.json**
2. In Zed, execute "Import Settings from VS Code" (either on first install, or in the command palette)
3. Observe that Zed's `minimap.show` remains set to "never".

Current behavior: The setting is not imported.
Expected behavior: The setting should be imported, as [stated in the docs](https://zed.dev/docs/migrate/vs-code#settings-imported-from-vs-code).

The defaults for the minimap are opposite between the two editors. In VSCode it's on by default, in Zed it's off. Because it's the default the VSCode user settings.json does not mention the minimap at all, which might be why Zed ignores this on import. I still think Zed should turn the minimap on if a user imports the default settings. First, because a user clicking the "import" expects an experience close to what they've had before. Second, because even toggling the minimap off and on in the VSCode settings UI will still not explicitly add it to the json (because it's the default).

OS: Linux Wayland fedora 44
Architecture: x86_64

cppcoffee:
When importing settings from VS Code, the minimap is now enabled by default (matching VS Code's behavior) even if the user hasn't explicitly set minimap-related options in their VS Code settings.json.

Previously, if `editor.minimap.enabled` and `editor.minimap.autohide` were absent from the VS Code config, Zed would leave minimap at its own default ("never"). Since VS Code defaults to minimap on, this caused a mismatch for users expecting their VS Code experience to carry over.

Closes #56297

Release Notes:

- Fixed apply VS Code minimap default when importing settings

### DIFF
--- a/crates/settings/src/settings_store.rs
+++ b/crates/settings/src/settings_store.rs
@@ -2177,6 +2177,9 @@ mod tests {
             r#" { "editor.tabSize": 37 } "#.to_owned(),
             r#"{
               "base_keymap": "VSCode",
+              "minimap": {
+                "show": "always"
+              },
               "tab_size": 37
             }
             "#
@@ -2195,6 +2198,9 @@ mod tests {
             r#"{ "editor.tabSize": 42 }"#.to_owned(),
             r#"{
                 "base_keymap": "VSCode",
+                "minimap": {
+                    "show": "always"
+                },
                 "tab_size": 42,
                 "preferred_line_length": 99,
             }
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
Looks good. Thank you for the contribution!

PR #56485, opened 2026-05-12
agent_panel: Make thread title editing more discoverable

danilo-leal:
This PR adds an edit icon button in the thread title to better promote/communicate that that's an editable label.

https://github.com/user-attachments/assets/efd69efc-ad9b-4c48-9640-e21b09d70226

Release Notes:

- N/A


bennetbo:
@zed-industries/approved

PR #56487, opened 2026-05-12
extension_host: Sync language extensions to remotes

### LINKED ISSUE #56068: Zed tells that TOML is not installed and the language server doesn't start in a remote session
### Reproduction steps

1. Open a remote directory
2. Confirm `TOML` and `Tombi` extensions are installed.
3. Create a project settings file `.zed/settings.json`.
4. Write like this into the setting file like this:
```json
{
    "languages": {
        "TOML": {},
    },
}
```
5. A warning message `No language with this name is installed.` appears at `"TOML"`:
Image

Additionally, the `tombi` language server doesn't start on remote sessions.
Image

Additional context:
`tombi` command is installed at `~/.local/bin/tombi` on the remote server and works fine.

Current behavior: "TOML" field is not recognized as an installed language and the language server doesn't start.
Expected behavior: "TOML" field is recognized as an installed language and the language server starts.

Zed: v1.1.6+stable.265.96962914296a22c22a7f226b5ae7de4dd0e34729 (Zed) 
OS: Windows 10.0.26200
Memory: 61.7 GiB
Architecture: x86_64
GPU: AMD Radeon RX 9060 XT || AMD Corporation || 26.5.1 (26.10.07.01-260428a-200434C-AMD-Software-Adrenalin-Edition)

### Attach Zed log file

```log
2026-05-08T00:09:06+09:00 INFO  [zed] ========== starting zed version 1.1.6+stable.265.96962914296a22c22a7f226b5ae7de4dd0e34729, sha 9696291 ==========
2026-05-08T00:09:07+09:00 INFO  [gpui_windows::directx_devices] Using GPU: AMD Radeon RX 9060 XT
2026-05-08T00:09:07+09:00 INFO  [gpui_windows::directx_devices] Created device with Direct3D 11.1 feature level.
2026-05-08T00:09:07+09:00 INFO  [gpui_windows::direct_write] Use Yu Gothic UI as UI font.
2026-05-08T00:09:07+09:00 INFO  [crashes] crash signal handlers installed
2026-05-08T00:09:07+09:00 INFO  [crashes] spawning crash handler process
2026-05-08T00:09:07+09:00 INFO  [zed::zed] Using GPU: GpuSpecs { is_software_emulated: false, device_name: "AMD Radeon RX 9060 XT", driver_name: "AMD Corporation", driver_info: "26.5.1 (26.10.07.01-260428a-200434C-AMD-Software-Adrenalin-Edition)" }
2026-05-08T00:09:07+09:00 ERROR [crates/gpui_windows/src/platform.rs:280] アクセスが拒否されました。 (0x80070005)
2026-05-08T00:09:13+09:00 INFO  [recent_projects::remote_servers] SSH: Watching User Config at: "C:\\Users\\*****\\.ssh/config"
2026-05-08T00:09:17+09:00 INFO  [remote::transport::ssh] Remote is windows: false
2026-05-08T00:09:17+09:00 INFO  [remote::transport::ssh] Remote shell discovered: /bin/bash
2026-05-08T00:09:17+09:00 INFO  [remote::transport::ssh] Remote platform discovered: RemotePlatform { os: Linux, arch: X86_64 }
2026-05-08T00:09:18+09:00 INFO  [remote_server] (remote server) starting up with PID 454828:
pid_file: "/home/*****/.local/share/zed/server_state/workspace-15/server.pid", log_file: "/home/*****/.local/share/zed/logs/server-workspace-15.log", stdin_socket: "/home/*****/.local/share/zed/server_state/workspace-15/stdin.sock", stdout_socket: "/home/*****/.local/share/zed/server_state/workspace-15/stdout.sock", stderr_socket: "/home/*****/.local/share/zed/server_state/workspace-15/stderr.sock"
2026-05-08T00:09:18+09:00 INFO  [crashes] (remote server) spawning crash handler process
2026-05-08T00:09:18+09:00 INFO  [remote_server] (remote server) gpui app started, initializing server
2026-05-08T00:09:18+09:00 INFO  [remote_server] (remote server) accepting new connections
2026-05-08T00:09:18+09:00 INFO  [remote_server] (remote server) accepted new connections
2026-05-08T00:09:18+09:00 ERROR [remote::remote_client] client:unhandled remote message name:ExternalAgentsUpdated
2026-05-08T00:09:18+09:00 INFO  [remote_server] (remote server) Got new node settings: NodeBinarySettings { path: None, npm_path: None, ignore_system_version: false }
2026-05-08T00:09:18+09:00 INFO  [zed::zed] Using GPU: GpuSpecs { is_software_emulated: false, device_name: "AMD Radeon RX 9060 XT", driver_name: "AMD Corporation", driver_info: "26.5.1 (26.10.07.01-260428a-200434C-AMD-Software-Adrenalin-Edition)" }
2026-05-08T00:09:18+09:00 INFO  [util] (remote server) set environment variables from shell:/bin/bash, path:/home/*****/.local/bin:/home/*****/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games:/home/*****/.local/bin
2026-05-08T00:09:18+09:00 INFO  [crashes] (remote server) connected to crash handler process after 100ms
2026-05-08T00:09:18+09:00 INFO  [crashes] (remote server) crash handler registered
2026-05-08T00:09:18+09:00 INFO  [node_runtime] (remote server) Node runtime install_if_needed
2026-05-08T00:09:18+09:00 INFO  [node_runtime] (remote server) using Zed managed Node.js at /home/*****/.local/share/zed/node/node-v24.11.0-linux-x64 since system Node.js wasn't found on PATH: cannot find binary path
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: docker-compose
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: docker-language-server
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: dockerfile-language-server
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded debug adapter: buildx-dockerfile
2026-05-08T00:09:18+09:00 INFO  [lsp] (remote server) starting language server process. binary path: "/home/*****/.local/share/zed/node/node-v24.11.0-linux-x64/bin/node", working directory: "/home/*****/code/wallclock", args: ["/home/*****/.local/share/zed/languages/json-language-server/node_modules/vscode-langservers-extracted/bin/vscode-json-language-server", "--stdio"]
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: vscode-html-language-server
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: nginx
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: svelte-language-server
2026-05-08T00:09:18+09:00 INFO  [extension_host::headless_host] (remote server) Loaded language server: tombi
2026-05-08T00:09:26+09:00 INFO  [alacritty_terminal::tty::windows::conpty] Using conpty.dll for pseudoconsole

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

Ubuntu

terror:
Resolves #56068

Remote extension sync previously only sent extensions that can run remotely: language servers, debug adapters, and debug locators. That excluded language-only extensions such as TOML, so a remote session could have Tombi synced and loaded while the remote language registry still lacked TOML's language definition.

This diff keeps the remote sync set narrow: it starts from extensions that can run remotely, then adds the extension that defines each language referenced by those extensions' language servers. This installs the TOML language config alongside Tombi without syncing unrelated language-only extensions, allowing project settings schemas to recognize language names like TOML and allowing remote language servers to attach to buffers whose language is provided by a separate extension.

Release Notes:

- Fixed remote sessions failing to recognize languages provided by language-only extensions.


MrSubidubi (changes requested):
Thanks for finding the issue here! 

I think however that we should and need to improve how we determine an extension needs to be synced: 

Technically speaking, I think it would be somewhat better if we were to have a small dependency graph for such things and thus know what another extension needs in order to function properly. We definitely do not have to sync all language extensions, do we? I think we could probably go as far and only sync the config at some point, because if I am not mistaken, we would not need the grammar on the remote at all, would we? 

Happy to discuss this here with you further before proceeding, but think that is a better path forward. 

Similarly, I think we could use that within Zed aswell, as I do think we have some issues with the order we invoke/start extensions on a reload due to not having such a graph

terror:
> Thanks for finding the issue here!
> 
> I think however that we should and need to improve how we determine an extension needs to be synced:
> 
> Technically speaking, I think it would be somewhat better if we were to have a small dependency graph for such things and thus know what another extension needs in order to function properly. We definitely do not have to sync all language extensions, do we? I think we could probably go as far and only sync the config at some point, because if I am not mistaken, we would not need the grammar on the remote at all, would we?
> 
> Happy to discuss this here with you further before proceeding, but think that is a better path forward.
> 
> Similarly, I think we could use that within Zed aswell, as I do think we have some issues with the order we invoke/start extensions on a reload due to not having such a graph

Thanks! That makes sense.

One note from going over this path again: we already don’t upload grammar artifacts here. [`prepare_remote_extension`](https://github.com/zed-industries/zed/blob/8bccd3a72427c10d48e3d21ea1e02a9410d42eda/crates/extension_host/src/extension_host.rs#L1702-L1779) only uploads `extension.toml`, `extension.wasm` when present, language `config.toml`, and debug adapter schemas. The headless side also drops `config.grammar` before registering the language, so for this case the remote only consumes the language config for language registration.

I agree that syncing every language extension is broader than necessary though. I think the narrower version for this PR would be to build the sync set from extensions that can run remotely, then add the language-providing extension for each language referenced by their `language_servers`. That should cover the Tombi/TOML case without syncing unrelated language-only extensions 🤔 .

A more general dependency graph/toposort for extension reload ordering sounds useful too, but that feels
like a larger follow-up than this fix. But, I'm willing to explore this option if that is the direction approved and/or sought after here.

MrSubidubi:
> One note from going over this path again: we already don’t upload grammar artifacts here. [prepare_remote_extension](https://github.com/zed-industries/zed/blob/8bccd3a72427c10d48e3d21ea1e02a9410d42eda/crates/extension_host/src/extension_host.rs#L1702-L1779) only uploads extension.toml, extension.wasm when present, language config.toml, and debug adapter schemas. The headless side also drops config.grammar before registering the language, so for this case the remote only consumes the language config for language registration.

Very nice find and you are indeed correct, thank you for spotting this!

I think all of the three options you say here (current approach, intermediate approach and full fix) sound like good ways forward. I'd also be happy to proceed with this one as an interim if we want to move the proper "fix" into a distinct PR since yes, this will involve more work. Which option are you most keen about?

terror:
> > One note from going over this path again: we already don’t upload grammar artifacts here. [prepare_remote_extension](https://github.com/zed-industries/zed/blob/8bccd3a72427c10d48e3d21ea1e02a9410d42eda/crates/extension_host/src/extension_host.rs#L1702-L1779) only uploads extension.toml, extension.wasm when present, language config.toml, and debug adapter schemas. The headless side also drops config.grammar before registering the language, so for this case the remote only consumes the language config for language registration.
> 
> Very nice find and you are indeed correct, thank you for spotting this!
> 
> I think all of the three options you say here (current approach, intermediate approach and full fix) sound like good ways forward. I'd also be happy to proceed with this one as an interim if we want to move the proper "fix" into a distinct PR since yes, this will involve more work. Which option are you most keen about?

I’d be happy to go either way between the current approach and the intermediate one in this PR (whatever works for the team, and how high priority this currently is). If we end up merging this as-is, I can follow up with the language-aware sync behavior separately.

I do think the full dependency graph/toposort should happen in a separate PR either way (and we would probably want a more detailed issue for it), since that also touches extension reload ordering and will likely need more dedicated coverage around load/unload behavior.

MrSubidubi:
> I do think the full dependency graph/toposort should happen in a separate PR either way (and we would probably want a more detailed issue for it), since that also touches extension reload ordering and will likely need more dedicated coverage around load/unload behavior.

That very much makes sense.

If its fine with you, let's go with the intermediate approach so we only upload what really is needed here, then follow up with the more thorough fix later.

terror:
Gently bumping this @MrSubidubi :)

MrSubidubi (changes requested):
Sorry for the delay, will also get back to your mail-request shortly 🙂 

Few notes:
- While we are at it, do we want to fix the same for debug adapters too? Can we perhaps look into a solution which makes this less error-prone, e.g. by changing the return type of `allow_remote_load` and instead getting some sort of dependencies there? Or do we want to get to this later? Happy to keep it contained to language servers for now.
- I am a big fan of the fold, but am unsure about the type you use for storing: Can we not just use a single HashMap instead and `into_iter` this at the end instead? I am also leaning towards having a newtype for that HashMap and implementing the methods on that type instead, but not insisting in any way

terror:
> Sorry for the delay, will also get back to your mail-request shortly 🙂
> 
> Few notes:
> 
> * While we are at it, do we want to fix the same for debug adapters too? Can we perhaps look into a solution which makes this less error-prone, e.g. by changing the return type of `allow_remote_load` and instead getting some sort of dependencies there? Or do we want to get to this later? Happy to keep it contained to language servers for now.
> * I am a big fan of the fold, but am unsure about the type you use for storing: Can we not just use a single HashMap instead and `into_iter` this at the end instead? I am also leaning towards having a newtype for that HashMap and implementing the methods on that type instead, but not insisting in any way

Thanks! I took a stab at both points. I kept this scoped but made `allow_remote_load` expose a `remote_load` view that carries dependency information, so the sync logic no longer has to know which manifest fields imply remote loading.

I also switched the collector to a `RemoteSyncExtensions` newtype over a single `HashMap`, and added debug adapter language dependencies via an optional languages field on `[debug_adapters.*]`. Existing debug adapters without dependencies still sync as before.

crates/extension_host/src/extension_host.rs
```diff
@@ -178,6 +178,49 @@ pub struct ExtensionIndex {
...
+
+impl RemoteSyncExtensions {
+    fn insert(&mut self, index: &ExtensionIndex, id: &Arc<str>) {
+        let Some(entry) = index.extensions.get(id) else {
+            return;
+        };
+
+        if self.0.insert(id.clone(), entry.clone()).is_some() {
+            return;
+        }
```
MrSubidubi: Can we use `contains_key` or something similar here to not need to clone immediately? Not insisting, since both are Arcs, but still curious, as it look s a bit off on first glance

crates/extension_host/src/extension_host.rs
```diff
@@ -178,6 +178,49 @@ pub struct ExtensionIndex {
...
+
+        extensions
+    }
+}
+
+#[derive(Default)]
+struct RemoteSyncExtensions(HashMap<Arc<str>, ExtensionIndexEntry>);
+
+impl RemoteSyncExtensions {
+    fn insert(&mut self, index: &ExtensionIndex, id: &Arc<str>) {
```
MrSubidubi: Can we find a better name for this method? I see why you chose it, but I'd llike if we could convey a bit better here that this inserts the index entry and all it's dependencies.

MrSubidubi (changes requested):
Sorry for another delay, got sick and am still recovering. Two minor nits, approach looks good overall though!

Also, I synced with someone on the team who added support for debuggers and was informed there was a good reason for them to not be scoped to languages - sorry for the back and forth on this, but please remove the languages field there again. My bad! 😅

terror:
> Sorry for another delay, got sick and am still recovering. Two minor nits, approach looks good overall though!
> 
> Also, I synced with someone on the team who added support for debuggers and was informed there was a good reason for them to not be scoped to languages - sorry for the back and forth on this, but please remove the languages field there again. My bad! 😅

Thanks Finn, addressed the last comments.

I removed the debug-adapter `languages` manifest field again, so language dependencies are only derived from language servers now. I also renamed the recursive sync helper to `insert_extension_and_language_dependencies` (a bit verbose but I think it gets the point across 🫠) and changed it to check `contains_key` before cloning/inserting the entry.

Get well soon!

crates/extension/src/extension_manifest.rs
```diff
@@ -345,8 +363,9 @@ pub struct SlashCommandManifestEntry {
     pub requires_argument: bool,
 }
 
-#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
+#[derive(Clone, Default, PartialEq, Eq, Debug, Deserialize, Serialize)]
 pub struct DebugAdapterManifestEntry {
+    #[serde(default)]
```
MrSubidubi:
```suggestion
```

crates/extension/src/extension_manifest.rs
```diff
@@ -475,6 +494,7 @@ mod tests {
         let adapter_name = Arc::from("my_adapter");
         let entry = DebugAdapterManifestEntry {
             schema_path: Some(rel_path_buf("foo/bar")),
+            ..Default::default()
```
MrSubidubi:
```suggestion
```

MrSubidubi (changes requested):

MrSubidubi (approved):
Sorry for the delay, looks good now, thank you for bearing with me!

Am finally better, so also thank you for that!

Happy to merge, thanks a lot, and feel free to pick up the deeper fix here if you feel like it!

PR #56499, opened 2026-05-12
editor: Fix inlay-hint hover targeting and popover position

aviatesk:
When two type hints sit one source character apart (e.g. a `::T` hint just before a closing token and another `::U` hint right after it), `clip_point(_, Bias::Right)` skips through the lone source character so both hints fall inside the buffer-position window used to find a hovered hint. The previous `max_by_key(|hint| hint.id)` heuristic then picked an arbitrary one, which often was not the inlay actually under the mouse. Filter the candidate hints by the actual mouse `hovered_offset` so the hint whose rendered text is under the cursor is the one resolved for hover and link navigation.

Additionally, the popover previously used the inlay's buffer anchor directly, which in display space always lands at the inlay's leading edge — so for long or multi-part inlays the popover would appear at the start of the inlay regardless of which character was hovered. Resolve the actual display position by adding the in-inlay byte offset onto the inlay's start.

> Before

https://github.com/user-attachments/assets/820289ec-eb8f-4052-af9d-af44cd645662

> After

https://github.com/user-attachments/assets/fee86939-a5d3-443e-a364-8dfb18c0b78c

Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Fixed incorrect inlay hint resolution and popover positioning when hovering inlay hints.

### DIFF
--- a/crates/editor/src/hover_popover.rs
+++ b/crates/editor/src/hover_popover.rs
@@ -2326,71 +2326,98 @@ mod tests {
             );
         });
 
-        // Hover in the middle of the first inlay's rendered text.
-        let hover_position = cx.update_editor(|editor, window, cx| {
-            let snapshot = editor.snapshot(window, cx);
-            let inlay_start = MultiBufferOffset(first_hint_offset).to_display_point(&snapshot);
-            let exact_unclipped = DisplayPoint::new(
-                inlay_start.row(),
-                inlay_start.column() + (first_label.len() as u32 / 2),
-            );
-            let previous_valid = snapshot.clip_point(exact_unclipped, Bias::Left);
-            let next_valid = snapshot.clip_point(exact_unclipped, Bias::Right);
-            let nearest_valid = if previous_valid == next_valid {
-                previous_valid
-            } else {
-                match snapshot.inlay_bias_at(exact_unclipped) {
-                    Some(Bias::Left) => next_valid,
-                    Some(Bias::Right) => previous_valid,
-                    None => previous_valid,
+        async fn assert_hovered_inlay(
+            cx: &mut EditorLspTestContext,
+            hint_offset: usize,
+            label_len: u32,
+            expected_tooltip: &str,
+            expected_inlay_id: InlayId,
+        ) {
+            let hover_position = cx.update_editor(|editor, window, cx| {
+                let snapshot = editor.snapshot(window, cx);
+                let inlay_start = MultiBufferOffset(hint_offset).to_display_point(&snapshot);
+                let exact_unclipped =
+                    DisplayPoint::new(inlay_start.row(), inlay_start.column() + label_len / 2);
+                let previous_valid = snapshot.clip_point(exact_unclipped, Bias::Left);
+                let next_valid = snapshot.clip_point(exact_unclipped, Bias::Right);
+                let nearest_valid = if previous_valid == next_valid {
+                    previous_valid
+                } else {
+                    match snapshot.inlay_bias_at(exact_unclipped) {
+                        Some(Bias::Left) => next_valid,
+                        Some(Bias::Right) => previous_valid,
+                        None => previous_valid,
+                    }
+                };
+                PointForPosition {
+                    previous_valid,
+                    next_valid,
+                    nearest_valid,
+                    exact_unclipped,
+                    column_overshoot_after_line_end: 0,
                 }
-            };
-            PointForPosition {
-                previous_valid,
-                next_valid,
-                nearest_valid,
-                exact_unclipped,
-                column_overshoot_after_line_end: 0,
-            }
-        });
-        cx.update_editor(|editor, window, cx| {
-            editor.update_inlay_link_and_hover_points(
-                &editor.snapshot(window, cx),
-                hover_position,
-                None,
-                true,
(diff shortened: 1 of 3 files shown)


crates/editor/src/hover_popover.rs
```diff
@@ -2116,6 +2151,184 @@ mod tests {
...
+            );
+        });
+    }
+
+    /// When two type hints sit one source character apart (e.g. a `::T` hint
+    /// just before a closing token and another `::U` hint right after it),
+    /// `clip_point` skips through the lone source character with `Bias::Right`,
+    /// so both hints fall inside the buffer-position window used to find a
+    /// hovered hint. The hovered-offset filter must then disambiguate them by
+    /// the inlay text actually under the mouse.
```
agu-z: This comment is somewhat misleading. The filter is really range check in inlay-offset space; the inlay's text content isn't actually consulted. Mind rewording?

crates/editor/src/inlays/inlay_hints.rs
```diff
@@ -604,7 +604,13 @@ impl Editor {
                 point_for_position.next_valid.to_point(snapshot),
                 Bias::Right,
             );
-            if let Some(hovered_hint) = Self::visible_inlay_hints(self.display_map.read(cx))
+            // The buffer-position window can include neighbouring inlays
+            // (e.g. a `::Type` hint after a closing paren and another hint just
+            // past it both pass the take_while when `clip_point` skips through
+            // a one-character source span between them). Use the actual mouse
+            // `hovered_offset` to pick the inlay whose rendered text is under
+            // the cursor.
```
agu-z: nit: I'm not sure this comment adds much value. The code is pretty self-descriptive and not surprising.

crates/editor/src/hover_popover.rs
```diff
@@ -882,6 +883,48 @@ impl HoverState {
         px((dx.powi(2) + dy.powi(2)).sqrt())
     }
 
+    /// Resolves the display point that the active popover should anchor to.
+    ///
+    /// Diagnostic and text-symbol hovers anchor at the start of the hover range.
+    /// For inlay-hint hovers, however, the buffer anchor alone always maps to the
+    /// inlay's leading edge in display space, so the popover would pin to the start
+    /// of the inlay regardless of which character was hovered. Add the in-inlay
+    /// byte offset onto the inlay's start so the popover lands at the hovered position.
```
agu-z: nit: The last sentence reads as a narration of the implementation. The "why" above it already justifies what needs to happen. I'd drop it.

crates/editor/src/hover_popover.rs
```diff
@@ -2116,6 +2151,184 @@ mod tests {
...
+                1,
+                "expected exactly one popover",
+            );
+            let popover = hover_state.info_popovers.first().unwrap();
+            assert_eq!(
+                popover.get_rendered_text(cx),
+                "Foo",
+                "popover content should match the first inlay (the one under the cursor), \
+                 not the adjacent inlay one source character past it",
+            );
```
agu-z: The disambiguation is only tested in one direction — hovering `::Foo` and getting `Foo`. Could you add a symmetric assertion that hovers `::Bar` and gets `Bar`?

agu-z (changes requested):
Thanks for your contribution! The direction makes sense overall. I still need to look deeper into this change, but leaving some initial thoughts here.

crates/editor/src/hover_popover.rs
aviatesk: Thanks for your feedback - agreed on all. Fixed up the inappropriate comments in b4f3ea7871f8968afede33d7d1b2dd215507546c.

crates/editor/src/hover_popover.rs
aviatesk: Good idea. Applied: 5f371d806d404a087fef614c8914b00a0b783f55

aviatesk:
Friendly ping.

agu-z:
Thanks for addressing the feedback. Before approving I tried to reproduce both bugs manually and haven't managed to, so I'd like to confirm the behavior against a real setup. Could you share the exact one from your videos?

What I tried:

- rust-analyzer on stable, with two type hints one source character apart: `let pass = |v| v;` renders as `|v: Vec<u32>| -> Vec<u32>` with only the `|` between the two hint anchors (closure return-type hints enabled). Cmd-hovering the first hint underlined the part under the cursor as expected, and I couldn't get a popover at all for the long-hint position case.
- JETLS, since the videos look like Julia: I installed your zed-julia extension with the `jetls` app, but didn't get any inlay hints in Zed. I'm not familiar with Julia, so I may be missing something — how do I get inlay hints to show up?

A minimal repro — language server, code snippet, and where exactly to hover — would help a lot. Also worth noting which commit you saw the broken behavior on, in case main and stable differ here.

Sorry for the long delay on this review — I'll be paying attention next week so we can get it moving again.

agu-z:
@aviatesk I'd really like to be able to reproduce this bug and fix before merging. Can you provide the reproduction steps?

SomeoneToIgnore:
Will close to keep the queue cleaner, as it's been waiting on the answer for over 3 weeks.

PR #56500, opened 2026-05-12
git_panel: Add commit history view

danilo-leal:
This PR adds a history view to the git panel. It draws from similar implementations of commit listing in other places like the file history view and the git graph. 

| Changes View | History View |
|--------|--------|
| Screenshot 2026-05-12 at 3  38@2x | Screenshot 2026-05-12 at 3  39@2x | 

Release Notes:

- Git: Added a history view to the Git panel that allows to quickly see in a list all the commits for a given branch.

### DIFF
--- a/crates/git_ui/src/git_panel.rs
+++ b/crates/git_ui/src/git_panel.rs
@@ -1,6 +1,6 @@
 use crate::askpass_modal::AskPassModal;
 use crate::commit_modal::CommitModal;
-use crate::commit_tooltip::CommitTooltip;
+use crate::commit_tooltip::{CommitAvatar, CommitTooltip};
 use crate::commit_view::CommitView;
 use crate::git_panel_settings::GitPanelScrollbarAccessor;
 use crate::project_diff::{self, BranchDiff, Diff, ProjectDiff};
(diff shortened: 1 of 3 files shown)


Anthony-Eid (approved):
I wonder if we can unify this code and the git graph

danilo-leal:
Yeah, great call, will give that a pass in a follow up! Landing this one for now 😄 appreciate the review!

crates/git_ui/src/git_panel.rs
```diff
@@ -4888,6 +4924,426 @@ impl GitPanel {
...
+                            this.fetch_commit_history_shas(cx);
+                        }
+                    }
+                },
+            ));
+            self._repo_subscriptions
+                .push(cx.observe(active_repository, |_this, _repo, cx| {
+                    cx.notify();
+                }));
+        }
```
Anthony-Eid, line 5108: I think this is wrong, we should clear the data depending on the type of graph event. 

Also, should we have a way to switch the repo when the active repository changes

pantheraleo-7:
hey, sorry to comment on a merged PR, but since this pertains to UI I wanted get an opinion by Zed design staff:

the Git Graph button sitting next to the Uncommit button is risky in case of a mis-click (the buttons being [rightfully] small makes it more susceptible), and [IMO] it looks out of place conceptually too (since the things they do are unrelated) 

can I ask the design team to consider moving the Git Graph button somewhere else on the pane?

I have an alternative place in mind:

the graph icon preceding the branch name can be converted into the Git Graph button. what say?

image


(I can open a discussion if you want but I thought this was too trivial...)

shawnwall:
just wanted to swing by and say THANK YOU for this!!!

PR #56506, opened 2026-05-12
rules_library: Clarify Default Rules tooltips: attached to every *turn*

### LINKED ISSUE #56503: rules_library: Default Rules tooltips imply per-thread attachment, but rules are re-applied per turn
### Reproduction steps

1. Open the Rules Library
2. Mouse over the Default Rules ℹ️ icon
3. Create a new rule
4. Mouse over the "Add to Default Rules" 📎  button

Image

The tooltips for "Default Rules" in the rules library imply that rules are baked into a thread at creation time, but in reality they're re-injected into the system prompt on every turn (as revealed by testing and confirmed in the [system prompt template](https://github.com/zed-industries/zed/blob/main/crates/agent/src/templates/system_prompt.hbs#L221-L223)).

### Expected behavior

Both tooltips should answer 3 questions a user would actually ask:

1. *Where does the rule go?* → Into the system prompt of every thread (not just new ones).
2. *When does an edit take effect?* → On the next turn (no thread restart needed).
3. Given ^, *How is the token count affected?*

Zed Dev 1.3.0 (debug)
Commit: bf423dfc45beccd0f9b8a7a794a99e5d8ab7421c

tredondo:
Two tooltips in the rules library implied that default rules are baked into a thread's system prompt at thread creation, which is wrong.

The "Add to Default Rules" button (two locations - [separate bug](https://github.com/zed-industries/zed/issues/56507)) said:

  Always included in every thread.

The (i) icon next to the "Default Rules" header said:

  Default Rules are attached by default with every new thread.

Both phrasings answer the scope question (which threads is this attached to?) but invite the reading that an edit won't reach an already-open thread, and that the attachment is a one-time event at thread creation. Neither is true: ProjectContext is a per-project, live entity that's rebuilt on PromptsUpdatedEvent and re-rendered into the system prompt on every turn (see [crates/agent/src/agent.rs::build_project_context](https://github.com/zed-industries/zed/blob/a5bf443ef68382222a5bd64f3f10ffdcf37246a7/crates/agent/src/agent.rs#L520-L583) and [crates/agent/src/thread.rs::build_request_messages](https://github.com/zed-industries/zed/blob/a5bf443ef68382222a5bd64f3f10ffdcf37246a7/crates/agent/src/thread.rs#L3055-L3094)). Existing threads pick up rule edits on their next turn.

Reword both tooltips to the same two-line meta string:
- where it goes ("Included in every thread's system prompt.") and
- when edits take effect ("Edits take effect on the next turn.")

The \n renders as a hard line break, consistent with existing multi-line tooltips elsewhere in the codebase (recent_projects, remote_servers, tasks_ui/modal).

The (i) icon tooltip is migrated from Tooltip::text to Tooltip::with_meta to accommodate the two-line format and to match the "Add to Default Rules" tooltips' style. The Built-in Rules branch is migrated alongside it for consistency, keeping the same content rephrased as a title + meta pair.

Self-Review Checklist:

Closes #56503 (mostly)

Release Notes:
- Improved Rules library tooltips for Default rules

### On caching

The proposed tooltips deliberately don't mention caching, because that varies per provider. However, their phrasing does beg the question.

Fortunately, re-sending the system prompt every turn isn't wasteful for most users because the major providers cache it:

| Provider | Zed-side caching wiring |
|---|---|
| Anthropic (direct + cloud) | `cache_control: { type: "ephemeral" }` on the last content segment (`crates/anthropic/src/completion.rs`) |
| OpenAI (direct + cloud) | `prompt_cache_key = thread_id`; all OpenAI models opt in via `supports_prompt_cache_key() -> true` |
| Bedrock | `CachePoint` blocks for Claude models with a `cache_configuration` (Sonnet 4, Haiku 4.5, …) |
| Google Gemini | Implicit server-side; Zed observes `cached_content_token_count` in usage metadata |
| OpenAI-compatible / Vercel AI Gateway | Per-model `prompt_cache_key` capability detection |
| xAI, OpenRouter, DeepSeek, Mistral, Ollama, LM Studio, Copilot Chat, OpenCode | None wired up Zed-side (some may cache server-side regardless) |

### DIFF
--- a/crates/rules_library/src/rules_library.rs
+++ b/crates/rules_library/src/rules_library.rs
@@ -338,11 +338,18 @@ impl PickerDelegate for RulePickerDelegate {
     ) -> Option<Self::ListItem> {
         match self.filtered_entries.get(ix)? {
             RulePickerEntry::Header(title) => {
-                let tooltip_text = if title.as_ref() == "Built-in Rules" {
-                    "Built-in rules are those included out of the box with Zed."
-                } else {
-                    "Default Rules are attached by default with every new thread."
-                };
+                let (tooltip_title, tooltip_meta): (&'static str, &'static str) =
+                    if title.as_ref() == "Built-in Rules" {
+                        (
+                            "Built-in Rules",
+                            "These rules are included out of the box with Zed.",
+                        )
+                    } else {
+                        (
+                            "Default Rules",
+                            "Included in every thread's system prompt.\nEdits take effect on the next turn."
+                        )
+                    };
 
                 Some(
                     ListSubHeader::new(title.clone())
@@ -351,7 +358,9 @@ impl PickerDelegate for RulePickerDelegate {
                                 .style(ButtonStyle::Transparent)
                                 .icon_size(IconSize::Small)
                                 .icon_color(Color::Muted)
-                                .tooltip(Tooltip::text(tooltip_text))
+                                .tooltip(move |_window, cx| {
+                                    Tooltip::with_meta(tooltip_title, None, tooltip_meta, cx)
+                                })
                                 .into_any_element(),
                         )
                         .inset(true)
@@ -420,7 +429,7 @@ impl PickerDelegate for RulePickerDelegate {
                                                         Tooltip::with_meta(
                                                             "Add to Default Rules",
                                                             None,
-                                                            "Always included in every thread.",
+                                                            "Included in every thread's system prompt.\nEdits take effect on the next turn.",
                                                             cx,
                                                         )
                                                     })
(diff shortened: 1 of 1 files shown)


danilo-leal:
Hey @tredondo, thanks for this PR! We're actually removing the rules library from the product in favor of skills. We've introduced a migration, though, so all the functionality rules provided should be smoothly ported over to skills. That's being released today in Preview ([you can read the docs here](https://zed.dev/docs/preview/ai/rules#migrating-to-skills)) and next week in Stable. Therefore, closing this one now. Thanks again!

tredondo:
@danilo-leal No worries, makes total sense, and I'm happy to see the switch to skills!

Probably #56512 should be as well?

If so, maybe I can draw your attention to hopefully a more current PR - #57094

danilo-leal:
Yes, thanks for pointing me to that! I'm probably not the best suited person to review that other PR though, I'll let other colleagues chime in time permitting; thanks on working on these things!

PR #56521, opened 2026-05-12
project_panel: Persist worktree and directory collapsed state

### LINKED ISSUE #19898: Save folded directories in project
### Check for existing issues

- [X] Completed

### Describe the feature

I'd love to see the option to save and restore the folded state of directories when reloading or switching projects.

In some of my projects I have a few extra folders added to reference related code but don't always need them so when opening a project I have to fold the ones I don't want right now (but want to keep around) each time I reload or switch projects.

I'd also suggest adding a setting to toggle this feature, allowing users control over it.

_No response_

eth0net:
Self-Review Checklist:

Adds SQLite-backed persistence so the worktrees and directories you've collapsed in the project panel stay that way across sessions.

State is keyed off the workspace's existing `database_id` in a new `project_panel_collapse_state` table (`ProjectPanelDb` domain, FK to `workspaces` with `ON DELETE CASCADE`). One row per worktree stores a JSON array of expanded relative paths — this representation lets us distinguish "never saved" from "saved with nothing expanded" (e.g. the user explicitly collapsed the worktree root). Saves are per-worktree UPSERTs (so a save firing while some worktrees are still loading async can't drop their saved state) debounced by 100ms, and the debounced task is flushed through `Panel::flush_persistence`, which `Workspace::flush_serialization` joins, so quit, workspace close and project-group replacement all persist the most recent change.

One new project panel setting, `restore_collapse_state` (default `true`), toggles the persistence. It's also wired into the Settings UI page.

Fresh-workspace default behaviour is preserved: worktree roots still auto-expand the first time they're seen. Once the user explicitly collapses one and the state is saved, it stays collapsed on reload.

Tested with new project panel tests covering load, save, the persistence-disabled path, the workspace-close flush, renames, restoring inside a gitignored directory, and a save landing mid-scan, plus persistence-module tests for the empty-set sentinel, partial-save UPSERT semantics, and a malformed row. `cargo fmt` and `./script/clippy` are clean.

Closes #19898.

## Suggested .rules additions

Under a testing heading in the root `.rules`, next to "Timers in tests":

* Tests that write to a table with a `workspace_id` foreign key must call `cx.set_global(db::AppDatabase::test_new())` in their setup:
  - On the shared fallback connection, another test's `save_workspace` deletes the `workspaces` row and cascades yours away. It shows up only under parallel runs, as flakiness far from its cause.

Release Notes:

- Added the ability for the project panel to remember which worktree roots and directories you've collapsed or expanded across sessions, with a new \`project_panel.restore_collapse_state\` setting (default \`true\`)


crates/project_panel/src/persistence.rs
```diff
@@ -0,0 +1,163 @@
...
+                        VALUES (?1, ?2, ?3)
+                        ON CONFLICT(workspace_id, worktree_root_path)
+                        DO UPDATE SET expanded_paths = excluded.expanded_paths;
+                    ))?((
+                        workspace_id,
+                        worktree_root_str.as_ref(),
+                        serialized.as_str(),
+                    ))
+                    .log_err();
+                }
```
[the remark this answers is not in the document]
eth0net: Good catch — fixed in 6e59000. Replaced the per-row `.log_err()` with `.with_context(...)?` so a failing INSERT rolls back the savepoint and propagates to the caller, where the awaited save task's existing `.log_err()` provides visibility without committing a partial write.

crates/project_panel/src/project_panel.rs
```diff
@@ -4262,6 +4320,91 @@ impl ProjectPanel {
...
+            return;
+        };
+        let db = persistence::ProjectPanelDb::global(cx);
+        let executor = cx.background_executor().clone();
+        self._save_collapse_state_task = cx.background_executor().spawn(async move {
+            executor.timer(Duration::from_millis(100)).await;
+            db.save_expanded_entries(workspace_id, entries)
+                .await
+                .log_err();
+        });
```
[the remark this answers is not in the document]
eth0net: I don't think this is a problem in the current structure — happy to be corrected if I've missed a mutation site.

`schedule_save_collapse_state` is called from the `update_in` completion callback at the end of `update_visible_entries` (project_panel.rs:4309), not from the toggle action directly. Every mutation site you list ends up running `update_visible_entries`:

- `expand_selected_entry` (1316), `collapse_entry` (1365), `toggle_expanded` (1492), `toggle_expand_all` (1515), `collapse_all_entries` (1468), `collapse_selected_entry_and_children` (1400): call it directly.
- `expand_all_for_entry` / `collapse_all_for_entry`: only called from `toggle_expand_all` / `collapse_selected_entry_and_children`, both of which call it.
- `ExpandedAllForEntry` (723) and `WorktreeRemoved` (682) event handlers: call it directly.
- `reveal_entry` → `expand_entry` (6463/6464): calls it directly.
- Cursor-move helpers (2778/2817/2855/2947) and drop targets (4542, 5727): call it directly.
- The paste and rename helpers call `expand_entry` without an immediate `update_visible_entries`, but the subsequent `WorktreeUpdatedEntries` event from the project triggers it via the subscription at 685.

The "stale snapshot" half is also handled: `self._save_collapse_state_task = cx.background_executor().spawn(...)` drops the previous task on assignment, cancelling it. Only the most recently scheduled task fires, with the latest snapshot captured at that point.

crates/project_panel/src/project_panel.rs
```diff
@@ -4249,6 +4306,7 @@ impl ProjectPanel {
                     this.update_visible_entries_task.autoscroll = false;
                     this.autoscroll(cx);
                 }
+                this.schedule_save_collapse_state(cx);
                 cx.notify();
```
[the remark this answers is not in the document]
eth0net: Good point — fixed in ebfd261. Added a `last_scheduled_expanded_dir_ids` snapshot on `ProjectPanel` and check it at the top of `schedule_save_collapse_state`; if the current `expanded_dir_ids` is equal we skip the reschedule entirely, so git/worktree/settings updates that don't touch expansion state no longer churn the debounce task or trigger redundant writes.

crates/project_panel/src/project_panel.rs
```diff
@@ -4262,6 +4326,97 @@ impl ProjectPanel {
...
+    /// don't churn the database with redundant writes).
+    fn schedule_save_collapse_state(&mut self, cx: &mut Context<Self>) {
+        if self.last_scheduled_expanded_dir_ids.as_ref() == Some(&self.state.expanded_dir_ids) {
+            return;
+        }
+        let Some((workspace_id, entries)) = self.collapse_state_save_payload(cx) else {
+            return;
+        };
+        self.last_scheduled_expanded_dir_ids = Some(self.state.expanded_dir_ids.clone());
+        let db = persistence::ProjectPanelDb::global(cx);
```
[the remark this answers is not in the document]
eth0net: Excellent catch — fixed in 9b494b6. Switched the guard from comparing entry-id maps to comparing the actual path payload (`HashMap<Arc<Path>, Vec<String>>`) that gets persisted. A rename keeps the `ProjectEntryId` but mutates the entry's `RelPath`, so the new payload differs from the cached one and the save reschedules correctly. Added `test_collapse_state_save_follows_rename` that expands a directory, renames it, and asserts the persisted path follows the rename.

crates/project_panel/src/project_panel.rs
```diff
@@ -4262,6 +4326,98 @@ impl ProjectPanel {
...
+            let mut paths: Vec<String> = Vec::with_capacity(expanded_ids.len());
+            for entry_id in expanded_ids {
+                if let Some(entry) = worktree.entry_for_id(*entry_id) {
+                    paths.push(entry.path.as_unix_str().to_owned());
+                }
+            }
+            paths.sort();
+            result.insert(abs_path, paths);
+        }
+        Some(result)
```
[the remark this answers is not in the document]
eth0net: Fair point — fixed in 8eea75a. Added a cheap first-level guard using an entry-id snapshot (`HashMap<WorktreeId, Vec<ProjectEntryId>>`) alongside a `worktree_entries_changed_since_last_save: bool` flag set in the `WorktreeUpdatedEntries` event arm. The slow path (`current_expanded_paths` + path-payload comparison) only runs when either the entry-id snapshot changes or a worktree-updated event has fired since the last save — so git events, settings updates, and worktree-order changes are now skipped without rebuilding the payload, while renames still go through the slow path and are detected correctly (covered by the regression test from the previous round).

crates/project_panel/src/project_panel.rs
```diff
@@ -800,6 +827,20 @@ impl ProjectPanel {
...
+            let restore_collapse_state =
+                ProjectPanelSettings::get_global(cx).restore_collapse_state;
+            let pending_expanded_paths = if restore_collapse_state
+                && let Some(workspace_id) = workspace.database_id()
+            {
+                let db = persistence::ProjectPanelDb::global(cx);
+                db.expanded_entries(workspace_id)
+                    .log_err()
+                    .unwrap_or_default()
+            } else {
```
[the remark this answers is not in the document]
eth0net: I think keeping this synchronous is the right call. Sync reads from `WorkspaceDb`/`EditorDb`/`ProjectPanelDb` at workspace and panel construction is the established pattern in the codebase — see `EditorDb::get_serialized_editor` (`crates/editor/src/persistence.rs`), `WorkspaceDb::workspace_for_roots` and `WorkspaceDb::workspace_for_id` (`crates/workspace/src/persistence.rs`), and the various `WorkspaceDb::global(cx)` callers in `recent_projects` — all called sync on the UI thread during initialization. The `query!` macro itself generates sync functions.

For this query specifically, the cost is one indexed SELECT plus a small JSON decode per worktree row — sub-millisecond in practice even for large workspaces. Making it async would require either:

- spawning the load after construction and triggering a second `update_visible_entries` once it completes, which causes the panel to render with default state first and then visibly re-render once the loaded state arrives (jank that doesn't exist today), or
- holding off the panel's render entirely until the load completes, which is more invasive than the underlying read cost warrants.

Happy to revisit if profiling surfaces this as an actual bottleneck, but I'd prefer not to add the complexity speculatively when the precedent is sync.

crates/project_panel/src/project_panel.rs
```diff
@@ -666,8 +697,16 @@ impl ProjectPanel {
...
                         cx.notify();
                     }
-                    project::Event::WorktreeUpdatedEntries(_, _)
-                    | project::Event::WorktreeAdded(_)
+                    project::Event::WorktreeUpdatedEntries(_, _) => {
+                        // A rename keeps `ProjectEntryId` stable while changing
+                        // the entry's path, which the cheap entry-id guard in
+                        // `schedule_save_collapse_state` cannot detect. Flag
+                        // it so the next schedule recomputes the path payload.
+                        this.worktree_entries_changed_since_last_save = true;
```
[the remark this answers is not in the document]
eth0net: Good refinement — fixed in dc561cb. The handler now inspects the `UpdatedEntriesSet` and only flags `worktree_entries_changed_since_last_save` when the diff contains a path-structure change (`Added`/`Removed`/`AddedOrUpdated`/`Loaded`) or is empty. Pure `PathChange::Updated` (metadata-only) events now stay on the fast path, and renames are still caught because `worktree::build_diff` represents them as `Removed`+`Added`.

crates/project_panel/src/project_panel.rs
```diff
@@ -4441,13 +4521,32 @@ impl ProjectPanel {
                                 match new_state.expanded_dir_ids.entry(worktree_id) {
                                     hash_map::Entry::Occupied(e) => e.into_mut(),
                                     hash_map::Entry::Vacant(e) => {
```
smitbarmase, line 4648: 2. Here, when the first time a worktree shows up, we would hit this path only once. On a big repo the initial scan is still running at this point, so deeper saved paths don't resolve in `entry_for_path`, and the next save will write that smaller set back to the DB?

I think tests miss this because `Project::test` waits for the scan to finish before the panel exists.

smitbarmase (changes requested):
Thanks, I made an initial pass. Two main concerns before I could dive into the rest:

1. Gitignored dirs are never scanned unless something asks for them, so a saved expansion inside an ignored dir can never resolve no matter how long we wait. We would need to need some way to opt scanning for dirs we care about more like how it's handled in `expand_entry`.

eth0net:
Both fixed, rebased onto main.

Resolution ran exactly once, the first time a worktree appeared, and dropped whatever didn't resolve. It now runs on every update: paths that don't resolve stay pending, and one sitting under an unloaded ancestor asks the project to expand it, the same opt-in to scanning `expand_entry` does. `current_expanded_paths` folds pending paths back into the save payload, so a save landing mid-scan can't write back a smaller set than it read.

Two regression tests, both failing on the previous commit.

One for you to call: restoring an expansion inside a gitignored directory forces a scan of it at startup. Same work the user's manual expand would have done, and bounded to what they expanded last session, but say the word if you'd rather those paths were dropped instead.

crates/project_panel/src/project_panel_settings.rs
```diff
@@ -25,7 +25,9 @@ pub struct ProjectPanelSettings {
     pub indent_guides: IndentGuidesSettings,
     pub sticky_scroll: bool,
     pub auto_reveal_entries: bool,
+    pub auto_reveal_ignored_entries: bool,
```
smitbarmase: Can we move `auto_reveal_ignored_entries` to a separate PR? Collapse-state restoration does not depend on it.

crates/project_panel/src/project_panel.rs
```diff
@@ -4580,6 +4643,241 @@ impl ProjectPanel {
...
+    ///    last save, return without rebuilding paths.
+    /// 2. Otherwise build the path payload and compare it (this catches
+    ///    renames, where entry ids stay the same but paths change).
+    fn schedule_save_collapse_state(&mut self, cx: &mut Context<Self>) {
+        let ids_unchanged =
+            self.last_scheduled_expanded_dir_ids.as_ref() == Some(&self.state.expanded_dir_ids);
+        if ids_unchanged && !self.worktree_entries_changed_since_last_save {
+            return;
+        }
+        let Some((workspace_id, entries)) = self.collapse_state_save_payload(cx) else {
```
smitbarmase: Can we restart the debounce first, then build `collapse_state_save_payload` after the timer fires? `Expand All` currently rebuilds the full path payload on the foreground thread for every intermediate update.

crates/project_panel/src/project_panel.rs
```diff
@@ -4580,6 +4643,241 @@ impl ProjectPanel {
...
+
+        // Update the cheap-guard caches even when the path payload turns out
+        // to be unchanged, so we don't redo this work on the next call.
+        self.last_scheduled_expanded_dir_ids = Some(self.state.expanded_dir_ids.clone());
+        self.worktree_entries_changed_since_last_save = false;
+
+        if self.last_scheduled_expanded_paths.as_ref() == Some(&entries) {
+            return;
+        }
+        self.last_scheduled_expanded_paths = Some(entries.clone());
```
smitbarmase: A failed or cancelled save is still recorded in `last_scheduled_expanded_paths`. Later updates can then skip it as if it was persisted. Can we update the cache only after `save_expanded_entries` succeeds?

crates/project_panel/src/project_panel.rs
```diff
@@ -166,6 +173,27 @@ pub struct ProjectPanel {
...
+    /// the slow path on the next scheduled save in case a rename changed a
+    /// persisted path without changing any entry id. Cleared on each
+    /// scheduled save.
+    worktree_entries_changed_since_last_save: bool,
+    _save_collapse_state_task: Task<()>,
+    /// Scans requested on behalf of a saved expanded path, keyed by the
+    /// directory being loaded. Keeping the tasks here both keeps them running
+    /// and records which directories have already been asked for, so one that
+    /// takes several updates to load isn't requested once per update.
+    pending_dir_loads: HashMap<ProjectEntryId, Task<()>>,
```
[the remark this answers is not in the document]
eth0net: `next_entry_id` is a single `Arc<AtomicUsize>` owned by `WorktreeStore` and cloned into every worktree it creates (`worktree_store.rs:927`), so `ProjectEntryId` is unique across worktrees within a project. Remote worktrees don't allocate locally at all. No collision to guard against here.

crates/project_panel/src/project_panel.rs
```diff
@@ -4652,6 +4748,241 @@ impl ProjectPanel {
...
+        if ids_unchanged && !self.worktree_entries_changed_since_last_save {
+            return;
+        }
+        let Some((workspace_id, entries)) = self.collapse_state_save_payload(cx) else {
+            return;
+        };
+
+        // Update the cheap-guard caches even when the path payload turns out
+        // to be unchanged, so we don't redo this work on the next call.
+        self.last_scheduled_expanded_dir_ids = Some(self.state.expanded_dir_ids.clone());
```
smitbarmase: Let’s cache the last successfully written state, not the last scheduled state. I’d also build the path payload after the debounce so rapid toggles only build it once.

crates/project_panel/src/project_panel.rs
```diff
@@ -809,6 +863,12 @@ impl ProjectPanel {
             })
             .detach();
 
+            // Flush any pending debounced save before the app exits so the
+            // most recent collapse state isn't lost when the user quits Zed
+            // within the debounce window.
+            cx.on_app_quit(|this, cx| this.flush_collapse_state(cx))
```
smitbarmase: This needs to join the workspace close and replacement flush path. on_app_quit alone loses the latest toggle if a workspace closes inside the debounce window.

crates/project_panel/src/project_panel.rs
```diff
@@ -840,6 +900,19 @@ impl ProjectPanel {
...
             let weak_project_panel = cx.weak_entity();
+
+            let restore_collapse_state =
+                ProjectPanelSettings::get_global(cx).restore_collapse_state;
+            let pending_expanded_paths =
+                if restore_collapse_state && let Some(workspace_id) = workspace.database_id() {
+                    let db = persistence::ProjectPanelDb::global(cx);
+                    db.expanded_entries(workspace_id)
+                        .log_err()
+                        .unwrap_or_default()
```
smitbarmase: Don’t turn a database read failure into empty saved state. The next save can then overwrite good rows. Keep load failure separate and skip saving until the load succeeds.

crates/project_panel/src/persistence.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// entries" rather than treated as "never saved".
+    pub async fn save_expanded_entries(
+        &self,
+        workspace_id: WorkspaceId,
+        entries: HashMap<Arc<Path>, Vec<String>>,
+    ) -> Result<()> {
+        self.write(move |conn| {
+            conn.with_savepoint("project_panel_save_expanded_entries", || {
+                for (worktree_root_path, paths) in &entries {
+                    let worktree_root_str = worktree_root_path.to_string_lossy();
```
smitbarmase: `to_string_lossy()` isn’t safe for a database key. Store the path losslessly so two non-UTF worktree paths can’t collide.

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -11621,3 +11621,581 @@ async fn test_restore_file_prompt_escapes_markdown_in_file_name(cx: &mut gpui::T
 
     assert_eq!(message, "Discard changes to `__init__.py`?");
 }
+
+#[gpui::test]
+async fn test_restores_saved_collapse_state(cx: &mut TestAppContext) {
+    init_test_with_editor(cx);
```
smitbarmase, line 11995: These tests need their own AppDatabase::test_new() in setup. The shared fallback already makes the collapse-state tests interfere when the runner executes them in parallel.

crates/project_panel/src/project_panel.rs
```diff
@@ -695,8 +724,10 @@ impl ProjectPanel {
                 window,
                 |this, project, event, window, cx| match event {
                     project::Event::ActiveEntryChanged(Some(entry_id)) => {
-                        if ProjectPanelSettings::get_global(cx).auto_reveal_entries {
-                            this.reveal_entry(project.clone(), *entry_id, true, window, cx)
+                        let settings = ProjectPanelSettings::get_global(cx);
+                        if settings.auto_reveal_entries {
+                            let skip_ignored = !settings.auto_reveal_ignored_entries;
```
smitbarmase: Can we move `auto_reveal_ignored_entries` to a separate PR? It has nothing to do with collapse-state persistence.

crates/project_panel/src/persistence.rs
```diff
@@ -0,0 +1,167 @@
...
+                SELECT worktree_root_path, expanded_paths
+                FROM project_panel_collapse_state
+                WHERE workspace_id = ?
+            })
+            .and_then(|mut statement| statement(workspace_id))?;
+
+        let mut result: HashMap<Arc<Path>, Vec<String>> = HashMap::default();
+        for (worktree_root_path, expanded_paths) in rows {
+            let paths: Vec<String> = serde_json::from_str(&expanded_paths)
+                .with_context(|| format!("decoding expanded_paths for {worktree_root_path}"))?;
```
smitbarmase: Could we keep malformed JSON local to this worktree? The `?` here drops all saved collapse state. Log and skip this row, while still returning SQL and query failures.

crates/project_panel/src/project_panel.rs
```diff
@@ -166,6 +174,27 @@ pub struct ProjectPanel {
...
+    /// Path payload from the most recent scheduled save. Used as the
+    /// second-level guard so a `WorktreeUpdatedEntries` event that doesn't
+    /// actually change persisted paths (e.g. file content modified) doesn't
+    /// trigger a redundant DB write.
+    last_scheduled_expanded_paths: Option<collections::HashMap<Arc<Path>, Vec<String>>>,
+    /// Set when a `WorktreeUpdatedEntries` event fires, so we know to force
+    /// the slow path on the next scheduled save in case a rename changed a
+    /// persisted path without changing any entry id. Cleared on each
+    /// scheduled save.
+    worktree_entries_changed_since_last_save: bool,
```
smitbarmase: I don’t think we need all three caches. This can be consolidated to `last_persisted_collapse_state.` if you simply build path payload after debounce, and only update it after successful write. The single persisted-payload cache is reasonable if we want to avoid redundant writes.

smitbarmase (changes requested):
Thanks, this works well. I still haven't dived into the remote and some of the pending dir-load parts, but I think this unblocks you for the rest of the parts we should fix.

eth0net:
All eight addressed. `auto_reveal_ignored_entries` is out of this branch and will follow as its own PR.

The flush needed a no-op-default `Panel::flush_persistence` for `Workspace::flush_serialization` to join, since nothing that path does is observable from a panel, so there's a small change in `workspace` too. Writing its test turned up a panic on every workspace close: the hook runs while the workspace entity is already checked out, and the payload builder was reading it.

Test isolation was the `workspace_id` foreign key rather than shared rows. One test's `save_workspace` deleted another's `workspaces` row and took its collapse state with it, which is what made the rename test flaky.

crates/project_panel/src/project_panel.rs
eth0net: Hooked into `Workspace::flush_serialization` through a no-op-default `flush_persistence` on `Panel`, so removal and project-group replacement get it too. `on_app_quit` is gone.

The hook runs while the workspace entity is checked out, so a panel can't read the `Workspace` from inside it. This one caches `database_id` when a save is scheduled.

crates/project_panel/src/project_panel.rs
eth0net: Moved into a separate PR which I'll post after this one lands.

crates/project_panel/src/project_panel.rs
```diff
@@ -4668,6 +4735,233 @@ impl ProjectPanel {
...
+            let abs_path: Arc<Path> = worktree.abs_path();
+            let mut paths: Vec<String> = Vec::with_capacity(expanded_ids.len());
+            for entry_id in expanded_ids {
+                if let Some(entry) = worktree.entry_for_id(*entry_id) {
+                    paths.push(entry.path.as_unix_str().to_owned());
+                }
+            }
+            if let Some(pending) = self.state.pending_expanded_paths.get(&abs_path) {
+                paths.extend(pending.iter().cloned());
+            }
```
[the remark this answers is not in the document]
eth0net: Right, and Collapse All was undone entirely rather than just for the ignored subtree. Both recursive paths now drop pending paths under what they collapse, with a regression test.

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -11967,3 +11990,541 @@ async fn test_file_rows_reserve_the_chevron_slot(cx: &mut gpui::TestAppContext)
...
+    let root_entry = find_project_entry(&panel, "root", cx).expect("root entry exists");
+    panel.update_in(cx, |panel, window, cx| {
+        panel.toggle_expanded(root_entry, window, cx);
+    });
+    cx.run_until_parked();
+
+    // Drive the debounced save to completion.
+    cx.executor().advance_clock(Duration::from_millis(200));
+    cx.run_until_parked();
+
```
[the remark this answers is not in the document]
eth0net: The save is now awaited at every read that follows the debounce.

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -11967,3 +11990,612 @@ async fn test_file_rows_reserve_the_chevron_slot(cx: &mut gpui::TestAppContext)
...
+    // Drive the panel far enough to update and save, but deliberately not far
+    // enough to guarantee root2 has finished scanning.
+    for _ in 0..20 {
+        cx.executor().tick();
+    }
+    cx.executor().advance_clock(Duration::from_millis(200));
+    for _ in 0..20 {
+        cx.executor().tick();
+    }
```
smitbarmase, line 12829: Here, `FakeFs` finishes `root2`'s three entries before the panel's first background pass sees the worktree, so the save you await here is already post-scan. This would still pass if you remove `!initial_scan_complete` arm and the pending fold-in in `current_expanded_paths`.

How about we grab the real worktree snapshot, put `"does/not/exist"` in `state.pending_expanded_paths`, call it with `initial_scan_complete = false`, and assert the path is still pending. Then call it with `true` and assert it's gone.

crates/project_panel/src/project_panel_tests.rs
```diff
@@ -11967,3 +11990,612 @@ async fn test_file_rows_reserve_the_chevron_slot(cx: &mut gpui::TestAppContext)
...
+        .unwrap_or_default();
+    saved_paths.sort();
+    assert_eq!(
+        saved_paths,
+        vec![
+            "".to_string(),
+            "ignored".to_string(),
+            "ignored/nested".to_string()
+        ],
+        "restoring must not save a truncated set back over the state it just read",
```
smitbarmase, line 12420: I think this also passes with `paths.extend(pending...)` deleted.

After, panel is built and settled, we can do `panel.update(cx, |p, _| p.state.pending_expanded_paths.insert(root, vec!["ghost".into()]))`, then call `p.current_expanded_paths(cx)` and assert `"ghost"` is in the result. That proves the fold-in.

smitbarmase (changes requested):

crates/project_panel/src/project_panel_tests.rs
eth0net: You're right, the whole suite passed with that arm deleted. It's the direct call now: pending at `initial_scan_complete = false`, gone at `true`.

crates/project_panel/src/project_panel_tests.rs
eth0net: Deleting the fold-in changed nothing either. A ghost path goes into `pending_expanded_paths` now, and the payload is asserted to still carry it.

eth0net:
Mutating the rest of the collapse-state logic turned up three more the suite couldn't distinguish: the load request for a saved path whose ignored ancestor is collapsed, the guard that stops saving after a failed load, and the cache only being set once a write lands. All covered now.

One still isn't. The `PendingDir` arm only matters once the initial scan is done and a subtree is mid-scan, and I couldn't build that snapshot deterministically with `FakeFs`.

crates/project_panel/src/project_panel.rs
```diff
@@ -4668,6 +4746,261 @@ impl ProjectPanel {
...
+                continue;
+            };
+            let Some(task) = self.project.update(cx, |project, cx| {
+                project.expand_entry(worktree_id, entry_id, cx)
+            }) else {
+                continue;
+            };
+            slot.insert(cx.background_spawn(async move {
+                task.await.log_err();
+            }));
```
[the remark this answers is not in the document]
eth0net: Local `expand_entry` can't fail, but the remote one is an RPC and can. The marker is dropped when the load fails now, so the next resolution pass asks again.

No test for it: the panel's tests only build local worktrees, where `expand_entry` always returns `Ok`.

crates/project_panel/src/project_panel.rs
```diff
@@ -4372,16 +4433,29 @@ impl ProjectPanel {
...
 
         let visible_entries_task = cx.spawn_in(window, async move |this, cx| {
-            let new_state = cx
+            let (new_state, dirs_to_load) = cx
                 .background_spawn(async move {
-                    for worktree_snapshot in visible_worktrees {
+                    let mut dirs_to_load = Vec::new();
+                    for (worktree_snapshot, initial_scan_complete) in visible_worktrees {
                         let worktree_id = worktree_snapshot.id();
+                        Self::resolve_pending_expanded_paths(
```
smitbarmase, line 4453: I think remote restoration is still not correctly handled. Here, `resolve_pending_expanded_paths` can drop a saved expansion while a remote update is still arriving. For example, suppose we saved `ignored/child` as expanded. One update chunk marks the parent `ignored` as loaded, but `child` arrives in a later chunk. If restoration runs in between, it sees a loaded parent and a missing child. Since the initial scan has already completed, it discards `ignored/child` from the pending paths. When the child finally arrives, we've forgotten that it should be expanded.

crates/project_panel/src/project_panel.rs
```diff
@@ -4621,11 +4697,12 @@ impl ProjectPanel {
                             new_state.max_width_item_index = Some(visited_worktrees_length + index);
                         }
                     }
-                    new_state
+                    (new_state, dirs_to_load)
                 })
                 .await;
             this.update_in(cx, |this, window, cx| {
                 this.state = new_state;
+                this.load_dirs_for_pending_expanded_paths(dirs_to_load, cx);
```
smitbarmase, line 4705: One more case for remote that is, `load_dirs_for_pending_expanded_paths` calls `expand_entry` for the nearest known unloaded ancestor. For a saved path like `ignored/a/b`, it has to load `ignored` before discovering `a`, then repeat for `b`. That adds a remote round trip at each unloaded level.

crates/project_panel/src/project_panel.rs
```diff
@@ -4668,6 +4746,269 @@ impl ProjectPanel {
...
+        }
+        let workspace_id = self.collapse_state_workspace_id?;
+        let entries = self.current_expanded_paths(cx)?;
+        Some((workspace_id, entries))
+    }
+
+    /// Debounced save of the expanded directory state to the database. The
+    /// payload is built after the debounce elapses, so a burst of toggles
+    /// builds it once.
+    fn schedule_save_collapse_state(&mut self, cx: &mut Context<Self>) {
```
smitbarmase, line 4964: Also, saving mechanism needs bit more thought. Here, `schedule_save_collapse_state` cancels the previous task, but an already queued database write can still finish without updating `last_persisted_collapse_state`. Expanding and then collapsing a folder can therefore leave the database saying expanded while the cache says collapsed. `write_collapse_state` then skips saving the collapsed state because it matches the cache.

crates/project_panel/src/project_panel.rs
```diff
@@ -111,6 +113,12 @@ struct State {
     temporarily_unfolded_pending_state: Option<TemporaryUnfoldedPendingState>,
     unfolded_dir_ids: HashSet<ProjectEntryId>,
     expanded_dir_ids: HashMap<WorktreeId, Vec<ProjectEntryId>>,
+    /// Saved set of expanded relative paths per worktree absolute path, loaded
+    /// from the database on panel construction. Paths are resolved to entry
+    /// IDs by `resolve_pending_expanded_paths` on every update, and drop out
+    /// of this map once they resolve (or are found to no longer exist), so
+    /// what remains is the set still waiting on the worktree to scan them.
+    pending_expanded_paths: HashMap<Arc<Path>, Vec<String>>,
```
smitbarmase, line 121: This is pretty trivial. Turning off `restore_collapse_state` prevents new database writes, but doesn't clear `pending_expanded_paths` or cancel pending directory loads. Folders can continue expanding after the setting is disabled.

smitbarmase (commented):
I've taken another look, and I still have concerns.

We're still uncovering cases after several rounds, and I'm worried we cannot finish this without me diving in fully and working through the implementation. As you can see from the concerns above, I can't catch all these interactions in each review, and another round of comments doesn't feel like enough at this point.

I think this needs a deeper look, but I can't give it that time alongside my other work.

Given all that, I'm sorry, but I'm going to close the PR. I don't think it meets the effort/reward bar right now. Thanks for sticking with this and working through the feedback. I'm sorry we couldn't get it over the line.

eth0net:
Thank you for the response, I totally understand you don't have the time to invest in this at the moment, and I appreciate your ongoing effort!

I'll take a look over it again myself over the next few weeks and see if I can get it into shape properly. 

If I can get it fully nailed down, I will submit another PR. I'll make sure to link it all and provide any new reviewers with the full context.

Thanks again!

PR #56538, opened 2026-05-12
collab: Replace `TransitionalUserService` with `CloudUserService`

maxdeviant:
This PR replaces the `TransitionalUserService` with the `CloudUserService`, as all of the calls are now all going through Cloud.

This allows us to delete the `TransitionalUserService`, the `DatabaseUserService`, as well as the backing database queries that are no longer used.

Closes CLO-758.

Release Notes:

- N/A

### DIFF
--- a/crates/collab/src/db/queries/channels.rs
+++ b/crates/collab/src/db/queries/channels.rs
@@ -4,7 +4,7 @@ use rpc::{
     ErrorCode, ErrorCodeExt,
     proto::{ChannelBufferVersion, VectorClockEntry},
 };
-use sea_orm::{ActiveValue, DbBackend, TryGetableMany};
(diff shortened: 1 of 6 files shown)


crates/collab/src/db/queries/channels.rs
```diff
@@ -704,57 +704,29 @@ impl Database {
         .await
     }
 
-    /// Returns the details for the specified channel member.
-    pub async fn get_channel_participant_details(
+    /// Returns the members for the given channel.
+    #[cfg(feature = "test-support")]
+    pub async fn get_channel_members(
```
maxdeviant, line 709: `get_channel_participant_details` was being used in some tests, but just for retrieving the members.

We've replaced it with the test-only `get_channel_members`.

tomhoule (approved):
Exciting!

PR #56548, opened 2026-05-12
collab: Remove unused `From` impl for `User` model to `proto::User`

maxdeviant:
This PR removes a `From` impl from the database `User` model to `proto::User`, as it is no longer used.

Release Notes:

- N/A


tomhoule (approved):
🔪

PR #56553, opened 2026-05-12
Windows: Fix files opened by UNC path getting no language servers

Joblen:
Opening a file like `zed "\\server\share\file.txt"` leaves it with no language servers attached, and the tab shows the ugly `\\?\UNC\...` path.

The problem is the verbatim `\\?\UNC\` prefix Windows adds during path canonicalization. We rely on dunce to strip these prefixes, but it doesn't handle the UNC case, so the path stays in a form that breaks LSP URL conversion.

Fixed by stripping the prefix ourselves in `SanitizedPath` before handing off to dunce.

Before:
Screenshot 2026-06-12 075128

After: (note the correct tooltip and the syntax error highlighting)
Screenshot 2026-06-12 083651

Self-Review Checklist:

- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) — N/A, no UI changes
- [x] Performance impact has been considered and is acceptable — one extra string-prefix check per worktree creation

Release Notes:

- Fixed: language servers now run when opening files via UNC paths on Windows (e.g. `\\server\share\file.txt`).

### DIFF
--- a/crates/util/src/paths.rs
+++ b/crates/util/src/paths.rs
@@ -256,6 +256,10 @@ impl SanitizedPath {
 
         #[cfg(target_os = "windows")]
         {
+            let path = match path.to_str().and_then(|s| s.strip_prefix(r"\\?\UNC\")) {
+                Some(rest) => PathBuf::from(format!(r"\\{rest}")).into(),
+                None => path,
+            };
             let simplified = dunce::simplified(path.as_ref());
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi (approved):
thanks!

PR #56559, opened 2026-05-12
Fix token refresh for HTTP requests

ConradIrwin:
Code had been assuming (erroneously, but understandably) that
LlmApiToken::acquire would give them a valid token.

This is not true, as those tokens expire and you must call refresh explicitly.

Add some helpers to do the retry for you, and rename acquire to cached to be
clearer about the intent.

Closes #ISSUE

Release Notes:

- Fixed some rare cases where API requests would fail with Unauthorized

### DIFF
--- a/crates/edit_prediction/src/zeta.rs
+++ b/crates/edit_prediction/src/zeta.rs
@@ -7,10 +7,15 @@ use crate::{
 };
 use anyhow::Result;
 use cloud_llm_client::{
-    AcceptEditPredictionBody, EditPredictionRejectReason, predict_edits_v3::RawCompletionRequest,
+    AcceptEditPredictionBody, EditPredictionRejectReason, ZED_VERSION_HEADER_NAME,
+    predict_edits_v3::RawCompletionRequest,
 };
 use edit_prediction_types::PredictedCursorPosition;
-use gpui::{App, AppContext as _, Entity, Task, TaskExt, WeakEntity, prelude::*};
+use gpui::{
+    App, AppContext as _, Entity, Task, TaskExt, WeakEntity,
+    http_client::{self, Method},
+    prelude::*,
+};
 use language::{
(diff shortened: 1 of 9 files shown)


maxdeviant:
@probably-neb Could you look over the edit prediction changes? Especially how it might impact using local Zeta.

I see that we still have `ZED_ACCEPT_PREDICTION_URL`, but seems like we no longer have the environment variable to set the actual prediction endpoint (I seem to recall this might have gotten removed when we made local Zeta a first-class citizen)?

crates/edit_prediction/src/zeta.rs
```diff
@@ -651,35 +607,47 @@ pub(crate) fn edit_prediction_accepted(
...
+                .header("Content-Type", "application/json")
+                .header(ZED_VERSION_HEADER_NAME, app_version.to_string());
+            if let Some(token) = token.as_deref() {
+                builder = builder.header("Authorization", format!("Bearer {token}"));
+            }
+            let response = client
+                .http_client()
+                .send(builder.body(body.into())?)
+                .await?;
+            EditPredictionStore::process_api_response::<()>(response, &app_version).await?;
```
probably-neb: I don't remember the last time we used this custom accept URL. The only potential users are people self hosting and tracking their own acceptance data? It's probably safe to remove, but this is fine for now.

probably-neb (approved):

probably-neb:
Yep @maxdeviant, the custom URL was superseded by the OpenAI-compatible edit prediction URL setting

maxdeviant (approved):

crates/edit_prediction/src/zeta.rs
ConradIrwin: ok; I've gone ahead and removed it. (with an agent it's easy to bring back when it's next needed)

PR #56569, opened 2026-05-12
git: Allow choosing branch diff target in branch diff view

Anthony-Eid:
### Summary

Added a branch diff base branch picker that shows the selected diff base and prioritizes branches on the active branch’s remote. This also generalizes the branch picker’s checkout vs. select only behavior so its UI/search/rendering can be reused by a future git graph picker, and makes branch diff drop stale in progress base update tasks when the base or repository changes for faster response.

Self-Review Checklist:

Closes  #54050

Release Notes:

- git: Allow choosing branch diff base in branch diff view


JosephTLyons (approved):

JosephTLyons:
Code seems fine, one UI thing to note, we are showing the checkmark next to the checked out branch in the new picker, but I think we want the checkmark to be the selected base branch, for these two to agree:

image

Anthony-Eid:
> Code seems fine, one UI thing to note, we are showing the checkmark next to the checked out branch in the new picker, but I think we want the checkmark to be the selected base branch, for these two to agree:


What's a really good point. I'm going to hold off on merging this until I fix that

JosephTLyons:
Hopefully it doesn't make the code too ugly to fix that. 😅

Anthony-Eid:
I'm going to generalize the pickers internals that way we could use it for other stuff like the git graph as well. It will be more complicated, but it was something I wanted to do anyway.

PR #56577, opened 2026-05-12
agent_ui: Require an open project for agent panel

benbrandt:
A bit brute force, but it works.

image

Self-Review Checklist:

Release Notes:

- N/A


danilo-leal (approved):
Just slightly increased the scope by adding a component that's used in all relevant panels. There's maybe a solution out there that doesn't repeat the same message in various places, but I think this should work for now!

PR #56588, opened 2026-05-12
vim: Fix c{N}w eating trailing whitespace when count > 1

### LINKED ISSUE #56563: vim `c 2 w` deletes the trailing whitespace
### Reproduction steps

1. create a new file with the contents "hello world :)"
2. put the cursor on the h
3. in normal mode, press `c 2 w`

Current behavior: "hello world " including the whitespace after "world" is deleted.
Expected behavior: the whitespace after "world" is not deleted. This is the behavior in vim 9.1 and neovim 0.12.1. In zed this behavior is implemented only if count is 1.

OS: macOS 26.4.1
Memory: 16 GiB
Architecture: aarch64

### Attach Zed log file

_No response_

### Relevant Zed settings

_No response_

### Relevant Keymap

_No response_

_No response_

None

davidalecrim1:
## Summary

- `c{N}w` with count > 1 was incorrectly consuming trailing whitespace after the Nth word, while `cw` (count = 1) correctly preserved it
- The vim docs state `cw`/`cW` are treated like `ce`/`cE` on a non-blank; this rule applies for any count, not just 1
- Fix extends `expand_changed_word_selection` to use `next_word_end(N)` for count > 1, instead of falling through to `NextWordStart` which eats trailing whitespace

Fixes #56563

Release Notes:

- Fixed `c{N}w` incorrectly deleting trailing whitespace after the Nth word (e.g. `c2w` now matches Vim/Neovim behavior)

### DIFF
--- a/crates/vim/src/normal/change.rs
+++ b/crates/vim/src/normal/change.rs
@@ -165,30 +165,58 @@ fn expand_changed_word_selection(
             .map(|(c, _)| !classifier.is_whitespace(c))
             .unwrap_or_default()
     };
-    if (times.is_none() || times.unwrap() == 1) && is_in_word() {
-        let next_char = map
-            .buffer_chars_at(
-                motion::next_char(map, selection.end, false).to_offset(map, Bias::Left),
-            )
-            .next();
-        match next_char {
-            Some((' ', _)) => selection.end = motion::next_char(map, selection.end, false),
-            _ => {
-                if use_subword {
-                    selection.end =
-                        motion::next_subword_end(map, selection.end, ignore_punctuation, 1, false);
-                } else {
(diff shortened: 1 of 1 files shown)


davidhi7:
This does not quite replicate the vim behavior if the cursor is on the last character of the word and the count is greater than 1. In this case your implementation would delete one word too many. The reason is that this PR calls `motion::next_word_end` with `always_advance = true`, causing it to advance out of the current word before searching word endings for N times.

davidalecrim1:
@davidhi7 thanks for letting me know. I see you have a PR out as well, want me to drop this one?

davidhi7:
Since mine was submitted first and is ready for review, I think it make sense to let that one move forward. Thanks!

PR #56591, opened 2026-05-13
search: Fix UseSelectionForFind not refreshing matches for cmd-g

### LINKED ISSUE #55619: cmd-e/cmd-g behaviour is broken

Dev-X25874:
Self-Review Checklist:

Closes #55619

Release Notes:

- Fixed cmd-e / cmd-g (UseSelectionForFind / SelectNextMatch) not
  navigating to the correct match when the search bar is either
  dismissed or open.

### DIFF
--- a/crates/search/src/buffer_search.rs
+++ b/crates/search/src/buffer_search.rs
@@ -1246,23 +1246,18 @@ impl BufferSearchBar {
         let Some(search_text) = self.query_suggestion(true, window, cx) else {
             return;
         };
-        self.query_editor.update(cx, |query_editor, cx| {
-            query_editor.buffer().update(cx, |query_buffer, cx| {
-                let len = query_buffer.len(cx);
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Similar to #56249, this PR also doesn't seem to fix the issue when the search bar is hidden.

Please include details about how you're testing and verifying these fixes in the PR description so reviewers can understand what cases you're attempting to cover.

Dev-X25874:
Thanks for the review @ChristopherBiscardi. You're right — I missed that `update_matches` immediately calls `pending_external_query.take()`, which wipes the query before cmd-g ever fires. The correct pattern is already in the codebase at the `on_focus` handler: set `pending_external_query` when dismissed, call `self.search()` directly when open. I'll push a corrected commit shortly.

PR #56595, opened 2026-05-13
Fix active pane border and pane opacity dimming showing simultaneously on editor panes and terminal panes

usizu:
When `active_pane_modifiers.border_size` is set in `settings.json`, both the active editor pane and the active terminal pane would display a border simultaneously, making it impossible to tell which pane actually has focus. The pane opacity dimming also renders simultaneously before this fix.

1. Before fix: impossible to tell which pane has focus
image

2. After fix: only one pane has focus
image

The root cause is that both the workspace center and the terminal panel pass their own `active_pane` to `PaneRenderContext`, so each pane group unconditionally marks its active pane for border rendering. The fix adds a focus check so `active_pane_ix` is only set when the pane actually has keyboard focus, ensuring only the truly focused pane gets the border (and the inactive_opacity dimming applies correctly to unfocused groups).

Self-Review Checklist:

Release Notes:

- Fixed active pane border being drawn on both the editor pane and terminal pane simultaneously

### DIFF
--- a/crates/workspace/src/pane_group.rs
+++ b/crates/workspace/src/pane_group.rs
@@ -35,6 +35,8 @@ pub struct PaneGroup {
 pub struct PaneRenderResult {
     pub element: gpui::AnyElement,
     pub contains_active_pane: bool,
+    #[cfg(any(test, feature = "test-support"))]
+    pub decorated_pane_ix: Option<usize>,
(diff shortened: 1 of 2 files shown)


smitbarmase (approved):
Thanks!

Issue #56336, opened 2026-05-10
“Restore checkpoint” does not always work


JohnXu22786:
### Reproduction steps

I haven't figured out how it triggers 

### Current vs. Expected behavior

Always shows the button

### Zed version and system specs

1.1.7

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### If you are using WSL on Windows, what flavor of Linux are you using?

None

yeskunall:
Thanks for the report @JohnXu22786. This has been discussed internally, and we’ve noticed the same. Here’s a PR that aims to fix it: https://github.com/zed-industries/zed/pull/54817.

tredondo:
@JohnXu22786: a "Restore checkpoint" will be displayed only if ALL of these conditions are satisfied:
1. Your workspace needs to contain a git repo.
2. The agent changed at least one non-ignored file inside a git repo during that turn.
3. The checkpoint succeeded across **every** repo in the workspace (one failure nullifies the whole checkpoint - #59100)

yeskunall:
Looks like this was fixed in https://github.com/zed-industries/zed/pull/56611 instead.

PR #56611, opened 2026-05-13
agent: Show checkpoints as soon as the model edits files

bennetbo:
Makes it so that "Restore checkpoint" shows up as soon as the model performs an edit, instead of waiting until the turn ends

Release Notes:

- Fixed the agent panel to show "Restore checkpoint" as soon as the agent starts editing instead of waiting until the turn ends


benbrandt (approved):

tredondo:
@bennetbo Re. the *conditions in which* (rather than the moment at which) the checkpoint is displayed - would solving #59100 then log when the checkpoint was *not* displayed, thus shedding light on #56336?

PR #56619, opened 2026-05-13
csv_preview: #3 Implement data_table columns pining

HalavicH:
## Summary

Adds column pinning (freeze) capability to data tables, allowing the first N columns to remain visible while scrolling horizontally through the rest of the table content.

Common spreadsheet/data table UX pattern. When viewing wide tables with many columns, users need to see identifying information (row labels, IDs) while scrolling right to explore data.

## Demo
Idle state:
image
Scrolled horizontally (name column dissapeared, line numbers column stayed pinned):
image

## Implementation

- New `pin_cols(n: usize)` builder method on `Table` to specify how many columns to pin
- Pinned columns render in a fixed section that doesn't scroll horizontally
- Scrollable columns render separately with independent scroll state
- Horizontal scroll offset adjustments for proper column resize handle positioning with pinned sections
- Pinned section stays at viewport left edge while scrollable section scrolls independently
- Supports 0 < pinned_cols < total_cols (partial pinning)
- Applied to CSV preview for better UX with wide datasets

## Context

Part of CSV preview feature series, following PR #53496 (settings UI).

- [ ] ~~Aligned any UI changes with the [UI checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)~~ no ui changes besides pinning itself. UI improvements out of scope of this PR

Release Notes:

- Improved CSV preview with column pinning to keep identifiers visible while scrolling


Anthony-Eid:
I think for a feature like this it's important to have tests asserting the behavior we want 

cc: @smitbarmase

crates/csv_preview/src/renderer/render_table.rs
```diff
@@ -55,6 +55,7 @@ impl CsvPreviewView {
             .width_config(ColumnWidthConfig::Resizable(current_widths.clone()))
             .header(headers)
```
smitbarmase, line 56: I noticed there are a few Clippy errors that we should resolve.

```
error: redundant clone
error: redundant clone
error: redundant clone
error: redundant clone
error: could not compile `ui` (lib) due to 4 previous errors
warning: build failed, waiting for other jobs to finish...
error: could not compile `ui` (lib test) due to 4 previous errors
```

crates/ui/src/components/data_table.rs
```diff
@@ -561,27 +578,100 @@ pub fn render_table_row(
...
+        let scrollable_items: Vec<AnyElement> = items_vec.drain(pinned_cols..).collect();
+        let scrollable_widths: Vec<Option<Length>> = widths_vec.drain(pinned_cols..).collect();
+
+        // Pinned section: cells in normal flow with flex_shrink_0 to maintain their widths
+        let pinned_section = {
+            let disable = table_context.disable_base_cell_style;
+            let use_ui_font = table_context.use_ui_font;
+            div().flex().flex_row().flex_shrink_0().children(
+                items_vec.into_iter().zip(widths_vec).map(|(cell, width)| {
+                    if disable {
```
smitbarmase: Looks like we have this "disable if" code block in three different places now, it could use some refactoring.

crates/csv_preview/src/renderer/render_table.rs
```diff
@@ -55,6 +55,7 @@ impl CsvPreviewView {
...
             .disable_base_style()
+            .pin_cols(1)
```
smitbarmase, line 58: Thanks, since this is a public API, can we make this not panic?

Right now `.pin_cols(cols + 1)` would panic the whole Zed, which feels a bit too risky for something callers can invoke directly. I think we should handle this case gracefully instead, maybe by clamping to the valid range or otherwise returning early.

```rs
state.widths.as_slice()[..pinned_cols]
state.widths.as_slice()[pinned_cols..]
```

crates/ui/src/components/data_table.rs
```diff
@@ -769,15 +1104,49 @@ impl RenderOnce for Table {
...
+
         let no_rows_rendered = self.rows.is_empty();
         let variable_list_state = if let TableContents::VariableRowHeightList(data) = &self.rows {
             Some(data.list_state.clone())
         } else {
             None
         };
 
+        // Compute pinned_width and total_scrollable_width for the resize handles overlay.
+        let (pinned_width, total_scrollable_width) = if pinned_cols > 0 {
```
smitbarmase: I wonder if we can move the `pinned_width` and `total_scrollable_width` computation into `render_resize_handles_resizable` itself, since that seems to be the only place using them and we're already walking widths there.

smitbarmase (changes requested):
Thanks, I found a few nits, but there is one panic here that we should address first.

smitbarmase (approved):
Thanks, I refactored the code a bit and added a few basic tests. I think this is much cleaner now.

PR #56629, opened 2026-05-13
ep: Allow setting expected patch in rate completions modal

probably-neb:
Self-Review Checklist:

Closes #ISSUE

Release Notes:

- Added the ability to input the "expected patch", i.e. what the model should have predicted in the `edit prediction: rate completions` modal


osyvokon:
That's a cool change!

One question: Do we really need to display editable regions? My concern is that the ideal completion shouldn't really depend on technicalities like editable region. Also, editable region is a potential subject to change with other prompt formats.

probably-neb:
It's more of a nicety. Shouldn't matter for the most part. Fine to remove for other formats as necessary. I just noticed immediately when testing that I had no idea where the editable region boundaries were

danilo-leal:
@zed-industries/approved

PR #56632, opened 2026-05-13
copilot: Fix cache control error

bennetbo:
#56472 broke Copilot chat

> Failed to connect to API: 400 Bad Request {"message":"cache_control: Extra inputs are not permitted"}

This PR makes it so that we still use the legacy caching approach for Copilot

Release Notes:

- N/A


benbrandt (approved):

benbrandt:
@zed-industries/approved

PR #56647, opened 2026-05-13
Improve update button version display

danilo-leal:
This PR adds a small refinement to the update button, removing truncation from the version we display. Instead of just showing the short version of a SHA, we now display it in full, providing more detailed information that sometimes was lost. Also am making the button disabled when it's in the checking, downloading, and installing states, given clicking on it at these moments doesn't do anything.

Release Notes:

- N/A


probably-neb:
@zed-zippy approve

PR #56663, opened 2026-05-13
agent: Pull diagnostics

cameron1024:
Currently, the `diagnostics` tool queries the current state of the workspace. However, if the LSP is slow to update diagnostics, this can lead to stale diagnostics being given to the agent.

This PR changes the `diagnostics` tool to attempt to pull fresh diagnostics. If it fails, it informs the agent and falls back to the old behaviour (not all LSPs support pull-based diagnostics).

Release Notes:


benbrandt (approved):
Looks reasonable to me!

PR #56669, opened 2026-05-13
x11: Fix window not drawing after workspace switch in xmonad / i3

### LINKED ISSUE #56668: Window transparent after switching workspaces on X11
I have a fix for this, just creating the bug for tracking purposes.

### Reproduction steps

Use XMonad or I3 and Zed.  May need to use Mesa driver.  Switch workspaces, and see that the Zed window is sometimes transparent until focus enters it or there is some other source of render dirtiness.

Current behavior: zed window is sometimes transparent after it gets re-mapped but not resized.  In XMonad this can happen on workspace switch or layout changes within a that cause the window to switch visibility.  Also reported on reddit for the i3 tiling window manager [here](https://www.reddit.com/r/ZedEditor/comments/1tbvur2/zed_window_disappears_when_switching_i3_workspaces/).

Expected behavior: window is never durably transparent

Zed: v1.1.8+stable.269.7eb6bc7bad1afa0c026077d3a12e43f8e7c045f9 (Zed) 
OS: Linux X11 debian unknown
Memory: 30.8 GiB
Architecture: x86_64
GPU: Intel(R) Graphics (MTL) || Intel open-source Mesa driver || Mesa 26.0.6-1

mgsloan:
Closes #56668

It isn't clear exactly what is going wrong here.  In theory it should be fine to continue using the old buffers / pixmaps, and just re-present, but that isn't working.  Forcing a redraw seems to reliably fix it.  Since re-mapping a window is typically infrequent, seems fine to force a redraw.

This bug certainly wasn't around a few months ago.  Two theories:

1. Consequence of the work on GPU failure recovery

2. Perhaps issues with spurious dirtiness causing redraw were fixed, so the issue is no longer hidden.

Self-Review Checklist:

Release Notes:

- GPUI: Fixed an issue on Linux/X11 where the window is sometimes transparent after switching tiling window manager workspace

### DIFF
--- a/crates/gpui_linux/src/linux/x11/client.rs
+++ b/crates/gpui_linux/src/linux/x11/client.rs
@@ -795,6 +795,8 @@ impl X11Client {
                 let mut state = self.0.borrow_mut();
                 if let Some(window_ref) = state.windows.get_mut(&event.window) {
                     window_ref.is_mapped = true;
+                    // see #56668 - without this, window may be transparent after workspace switch
+                    window_ref.window.state.borrow_mut().force_render = true;
                 }
                 state.update_refresh_loop(event.window);
             }
@@ -1980,9 +1982,8 @@ impl X11ClientState {
                         if let Some(window) = state.windows.get_mut(&x_window) {
                             let expose_event_received = window.expose_event_received;
                             window.expose_event_received = false;
-                            let force_render = std::mem::take(
-                                &mut window.window.state.borrow_mut().force_render_after_recovery,
-                            );
+                            let force_render =
+                                std::mem::take(&mut window.window.state.borrow_mut().force_render);
                             let window = window.window.clone();
                             drop(state);
                             window.refresh(RequestFrameOptions {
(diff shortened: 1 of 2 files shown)


theurzil:
Hey @mgsloan, I'm actually the OP from [this Reddit post](https://www.reddit.com/r/ZedEditor/comments/1tbvur2/zed_window_disappears_when_switching_i3_workspaces/) that prompted this fix, really appreciate you taking the time to dig into this and open a PR, I really appreciate it.

I built from your branch and tested it on my setup (Arch Linux, i3, AMD Radeon Vega) but the issue is still occurring. Happy to provide logs or test any further changes if that helps. Thanks again!

intrd:
I'm experiencing the exact same problem. When switching away from and back to the workspace containing Zed in i3wm, the window fails to redraw. The render is only triggered by a mouseover event.

yeskunall:
Based on https://github.com/zed-industries/zed/issues/56668#issuecomment-5178202413, it looks like https://github.com/zed-industries/zed/pull/61162 fixed this issue and was reviewed before this one. Thanks for taking a stab at this @mgsloan, really appreciate the initial bug report and the fix here -- and great to see you contributing to Zed (also in general!) once again.

PR #56681, opened 2026-05-13
Fix macOS find query seeding

### LINKED ISSUE #55619: cmd-e/cmd-g behaviour is broken

nathansobo:
Closes #55619

### Summary

- Route `buffer_search::UseSelectionForFind` through `BufferSearchBar::deploy` instead of updating the query editor directly.
- Add an explicit seed-query override to `deploy`, so the Cmd-E action can force `SeedQuerySetting::Always` while regular deploy callers continue to pass `None` and respect the user’s `seed_search_query_from_cursor` setting.
- By going through `deploy`, Cmd-E now also runs the search path that keeps buffer-search navigation state in sync:
  - shows/initializes the search bar for the active searchable item
  - applies the seeded query via `search_suggested`
  - calls `search`, which updates the query editor, search options, active search query, search history, and macOS find pasteboard
  - refreshes `searchable_items_with_matches` and `active_match_index`
  - activates the current match after the search completes
- This ensures the subsequent Cmd-G action has the expected active query, match list, search token, and active match index to select the next result.
- Add a macOS-only end-to-end regression test using the default macOS keymap with `simulate_keystrokes("cmd-e")` and `simulate_keystrokes("cmd-g")`.

### Validation

- `cargo test -p search test_cmd_e_then_cmd_g_uses_selection_for_find`
- `cargo fmt --check --package search --package zed_actions`
- `./script/check-keymaps`
- `cargo check -p search`
- `cargo check -p workspace`
- `cargo check -p vim`

Release Notes:

- Fixed macOS Cmd-E/Cmd-G find behavior so Cmd-E seeds find from the cursor or selection and Cmd-G advances through the newly seeded matches.

### DIFF
--- a/crates/debugger_tools/src/dap_log.rs
+++ b/crates/debugger_tools/src/dap_log.rs
@@ -19,7 +19,7 @@ use project::{
     debugger::{dap_store, session::Session},
     search::SearchQuery,
 };
-use settings::Settings as _;
+use settings::{SeedQuerySetting, Settings as _};
 use std::{
     borrow::Cow,
     collections::{BTreeMap, HashMap, VecDeque},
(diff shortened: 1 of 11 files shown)


ChristopherBiscardi (approved):
works as intended. Search bar now shows every time when using Cmd+E, which seems to be an intentional change.

PR #56689, opened 2026-05-13
agent_ui: Render skills as creases

danilo-leal:
Closes AI-230

This PR makes skills, added as /-mentions, be rendered in the agent panel as creases, like anything you'd @-mention. Naturally, clicking on the crease button opens the corresponding skill file in a buffer.

It turned out to be quite a bit of plumbing to make this work, particularly as I am also introducing an interface to display dividers and headers in the completion menu. This was relevant to me to add because it sets a good foundation to convert many agent panel-related actions as slash commands.

Release Notes:

- N/A


MartinYe1234 (approved):
Need to add support for multiple skill creases, to be handled in a new PR:
https://github.com/zed-industries/zed/pull/56801

PR #56698, opened 2026-05-14
agent_ui: Hide "Start New Thread" when on free tier of Zed subscription

David3u:
Self-Review Checklist:

Release Notes:

- Hide "Start New Thread" when on free tier of Zed subscription

### DIFF
--- a/crates/agent_ui/src/agent_configuration.rs
+++ b/crates/agent_ui/src/agent_configuration.rs
@@ -329,7 +329,7 @@ impl AgentConfiguration {
                             "No configuration view for {provider_name}",
                         ))),
                     })
-                    .when(is_expanded && provider.is_authenticated(cx), |parent| {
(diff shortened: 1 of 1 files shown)


bennetbo:
Hey, can you clarify the reason for making this change? Zed free users can still start new threads?

David3u:
Hi Bennet. Looking at the pricing page, https://zed.dev/pricing and through my own experience trying to create a thread as a free user without the trial I thought free users couldn't create threads.

bennetbo:
That is correct, but AFAIC they will be prompted to start a trial once they start a new thread, so just hiding that action from the settings is not the right approach

David3u:
Ah I see. Thanks for clarifying. I still think it's bad ux to leave the button there for free users who have already been offered the free trial since the button not only does not start a new thread, it gives no indication that anything happened and the user has to leave the settings page to check. Perhaps renaming the button or some other change would be better than straight up removing it. Thanks for the replies

PR #56715, opened 2026-05-14
terminal: Fix terminal jitter on resize

### LINKED ISSUE #55934: Terminal jitter on resize starting with Zed 1.1.5
### Reproduction steps

1. Open a terminal
2. Resize it and observe the prompt

Zed: v1.1.5+stable.029747be6178a9d0db70d13e28da8c86640caa5e (Zed) 
OS: Linux X11 arch unknown
Memory: 93.9 GiB
Architecture: x86_64
GPU: NVIDIA RTX A1000 || NVIDIA || 595.71.05

---

Before Zed 1.1.x, the terminal content is pinned to the top, so when it resizes and the content isn't longer than the terminal's height it resizes smoothly:

https://github.com/user-attachments/assets/ca4ae36e-be76-4311-9d99-561641f4018e

After Zed 1.1.x, perhaps PR #47195, The terminal content is pinned to the bottom, so now when the terminal resizes the prompt is a bit jittery:

https://github.com/user-attachments/assets/54f3269c-ff53-4df6-bcc5-096b6a1b600f

The problem is, if we do pin the content to the top like Zed used to do, then resizing when the content IS longer than the terminal's height, it'll look jittery:

https://github.com/user-attachments/assets/56d9d66c-66d5-4824-8af7-b81b3753df99

Thus, I wonder if there's a strategy that can make it look smoother in both cases, like:

- pinning the content to the top when the last line is empty (content height < terminal height); pinning the content to the bottom when the last line isn't empty (content height  > terminal height); or
- rendering partial lines just like in the editor buffer.

I'm not sure if either is viable or a good idea, but just starting a thread first.

pigletfly:
Self-Review Checklist:

Closes #55934

Release Notes:

- Fixed terminal jitter when resizing vertically with content that doesn't fill the available height.


nathansobo:
I pushed a cleanup commit to this PR branch (`8f63ae9aaa`):

- Precompute whether the viewport bottom row is occupied while building terminal content, instead of rescanning all visible cells during prepaint.
- Keep `scrolled_to_bottom` and `bottom_row_occupied` as separate facts and combine them inline at the anchoring call site.
- Deduplicate the two resize anchoring tests behind a shared setup helper.
- Update the implementation for the terminal API/type changes introduced by the merge from `main`.

The targeted anchoring tests, full `terminal` and `terminal_view` test suites, and clippy all passed before the branch was externally reset; I reapplied the same cleanup before pushing. A final redundant test rerun was interrupted during compilation.

nathansobo:
Thanks for your contribution!

PR #56722, opened 2026-05-14
agent: Add setting to auto-approve external ACP agent tool calls

beardfaceguy:
Self-Review Checklist:

## Summary

External agents that talk to Zed over the Agent Client Protocol — Cursor,
Claude Code, Gemini CLI, and friends — own their own permission decisions
and delegate the confirmation prompt UI to Zed via
`session/request_permission`. The existing `agent.tool_permissions` rules
only apply to Zed's *native* agent, so today there is no way to globally
bypass external prompts: every call re-prompts even after the user picks
"Allow always", because the agent's "always" scope is per-call (e.g.
each web-search query is a different call from Cursor's perspective).

This PR adds an opt-in `agent.always_allow_external_agent_tools` boolean
(default `false`). When enabled, `handle_request_permission` immediately
responds with the agent's `AllowAlways` option (falling back to
`AllowOnce` when the agent doesn't offer `AllowAlways`) instead of
routing through the UI. This mirrors the "yolo" / "always allow" mode
those agents already expose in standalone CLI use.

## Why a separate setting (not extending ` + "`tool_permissions`" + `)

` + "`tool_permissions`" + ` matches against tool input text (commands, paths, URLs,
search queries) and it's the *agent process itself* that decides which
calls to ask about. With external ACP agents, Zed never sees the tool
input the way it does for native tools — it just sees the agent's prompt
options. A single boolean is the right shape for "trust the agent
end-to-end". Future work could extend this to per-agent allowlists once
the ACP surface lets us do that meaningfully.

## Tests

Unit tests in ` + "`crates/agent_servers/src/acp.rs`" + ` cover the pure helper:

- ` + "`auto_approve_outcome_prefers_allow_always`" + `
- ` + "`auto_approve_outcome_falls_back_to_allow_once`" + `
- ` + "`auto_approve_outcome_returns_none_when_only_reject_options`" + `
- ` + "`auto_approve_outcome_returns_none_for_empty_options`" + `
- ` + "`auto_approve_outcome_picks_first_allow_always_when_multiple_present`" + `

Implementation was developed test-first (red → green).

## Manual verification

Set:

` + "```json" + `
{ "agent": { "always_allow_external_agent_tools": true } }
` + "```" + `

…then run an external ACP agent (Cursor, Claude, Gemini) and trigger a
tool call that previously prompted (e.g. ` + "`web_search`" + `, ` + "`terminal`" + `).
Zed no longer shows the permission card; the call proceeds immediately.

## Suggested .rules additions

None — this is a one-off Settings extension and doesn't surface a
recurring trap that future agent sessions need to be warned about.

Release Notes:

- Added an opt-in ` + "`agent.always_allow_external_agent_tools`" + ` setting that auto-approves every tool-call permission request raised by external ACP agents (Cursor, Claude Code, Gemini CLI, etc.).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -279,6 +279,7 @@ dependencies = [
  "acp_thread",
  "action_log",
  "agent-client-protocol",
+ "agent_settings",
  "anyhow",
  "async-channel 2.5.0",
  "chrono",
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1102,6 +1102,13 @@
         // },
       },
     },
+    // When `true`, automatically approve every `session/request_permission`
+    // request raised by external ACP agents (e.g. Cursor, Claude Code,
+    // Gemini CLI) without showing a confirmation prompt. The `tool_permissions`
+    // rules above only apply to Zed's native agent — external agents own their
+    // own permission decisions. Use with care: this is the equivalent of
+    // running the external agent in a "yolo" / "always allow" mode.
+    "always_allow_external_agent_tools": false,
     // When enabled, agent edits will be displayed in single-file editors for review
     "single_file_review": false,
     // When enabled, show voting thumbs for feedback on agent edits.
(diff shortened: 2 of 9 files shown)


beardfaceguy:
@cla-bot check

ChristopherBiscardi:
There is no associated issue for this PR, and the PR description was not even reviewed for basic formatting, which doesn't make me confident in the contents of the PR.

image

PR #56446, opened 2026-05-11
Support renaming ACP threads

bennetbo:
Self-Review Checklist:

- [x] I've reviewed my own diff for quality, security, and reliability
- [x] Unsafe blocks (if any) have justifying comments
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist)
- [x] Tests cover the new/changed behavior
- [x] Performance impact has been considered and is acceptable

Closes #54689

Release Notes:

- Support editing the thread title for external agent threads

PR #56731, opened 2026-05-14
agent_ui: Make 'Edit Thread Title' pencil button functional

figtracer:
The "Edit Thread Title" pencil `IconButton` next to the thread title in the agent panel header (`render_title_view`) shows on hover with the tooltip `"Edit Thread Title"` but had no `on_click` handler — clicking it does nothing. The title editor is only reachable by clicking directly on the title text, which is not visually obvious because the title looks like a plain label.

This particularly bites users in AI Agent mode who try the pencil affordance, observe nothing happens, and conclude that thread titles can't be renamed (#56446 made ACP titles editable, but the pencil button itself remained inert).

This PR wires the pencil button to a new helper, `focus_active_title_editor`, that focuses the active title editor (agent thread or terminal) and selects all of its text — mirroring the pattern used by `edit_terminal_title`.

Behavior:

- Agent thread: focuses the existing `title_editor` from the active `ThreadView` and selects all text.
- Terminal: if a title editor already exists, focuses it and selects all; otherwise calls `edit_terminal_title` (which itself creates, focuses, and selects all).
- Other surfaces: no-op (the pencil is already gated behind `has_open_project` and the existing focus check).

Self-Review Checklist:

Release Notes:

- Fixed the "Edit Thread Title" pencil button in the agent panel header not doing anything when clicked.

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -3921,6 +3921,47 @@ impl AgentPanel {
         }
     }
 
+    fn focus_active_title_editor(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        match self.visible_surface() {
+            VisibleSurface::AgentThread(conversation_view) => {
+                if let Some(title_editor) = conversation_view
+                    .read(cx)
+                    .root_thread_view()
+                    .map(|view| view.read(cx).title_editor.clone())
+                {
+                    title_editor.update(cx, |editor, cx| {
+                        editor.focus_handle(cx).focus(window, cx);
+                        editor.select_all(&editor::actions::SelectAll, window, cx);
+                    });
+                }
+            }
+            VisibleSurface::Terminal(_) => {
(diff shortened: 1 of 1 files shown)


figtracer:
@cla-bot check

danilo-leal:
Heya, thanks for the PR! I had intentionally left it without any click handlers because the editor below it was already automatically getting focused as you clicked the icon button... but it seems like you weren't seeing this happen?

figtracer:
Ah, that makes sense — I missed that `ButtonLike` only attaches the `on_mouse_down`/`on_click`/`stop_propagation` chain when `on_click` is set ([button_like.rs#L767-L772](https://github.com/zed-industries/zed/blob/main/crates/ui/src/components/button/button_like.rs#L767-L772)), so without a handler the click falls straight through to the editor underneath. Sorry for the noise!

The reporter was on the latest stable and observed "can select but not edit" in AI agent mode, which I now believe is the read-only ACP title editor that was already fixed by #56446 — just not yet in a release. Closing this.

Issue #56734, opened 2026-05-14
External ACP agent tool calls (e.g. Cursor web_search) hang indefinitely with no auto-recovery


beardfaceguy:
### Reproduction steps

1. Start Zed.
2. Open the agent panel and select an external ACP agent (Cursor in my case; the same gap likely affects Codex / Claude Code / Gemini CLI).
3. Send a prompt that causes the agent to invoke its `web_search` tool (e.g. "search the web for the latest cargo release notes").
4. Wait — eventually one of these searches will not return.

### Current vs. Expected behavior

**Current behavior:** the tool-call card sits in the `InProgress` state forever. The `cursor-agent` (or other ACP agent) child process is stuck on a network call inside its own `web_search` implementation and never sends the matching `session/update` to transition the call out of `InProgress`. Zed has no inactivity or absolute timeout on ACP tool calls, so the spinner spins indefinitely until the user manually presses Escape (which fires `session/cancel`). If the user doesn't notice, the agent is stuck forever.

This is reproducible enough to leave orphaned `cursor-agent` processes on the system — `ps -ef | grep cursor-agent` on my machine shows multiple wedged children, including one whose original parent Zed process has long exited.

**Expected behavior:** Zed should give up on tool calls that haven't made any progress for a configurable amount of time. When the timeout fires it should:

1. Send `session/cancel` (the same path Escape uses today).
2. Mark the affected tool call with a status that's distinguishable from "user cancelled" (something like "Timed out after Ns") so it's obvious in the UI why generation stopped.
3. Allow the conversation to proceed — i.e. the prompt turn should resolve, not stay pending.

### Why this matters

External ACP agents own the tool execution, so Zed can't influence the network behavior of, e.g., Cursor's `web_search`. But Zed can detect that no `session/update` has arrived for that tool call in a long time and recover. Today the only recovery is manual cancellation, which is easy to miss because the spinner is the only signal — there's no visible deadline.

### Proposed shape

Add a `Option<u64>` setting (working name `agent.external_agent_tool_timeout_seconds`, default `None` to preserve current behavior). When set, schedule a per-tool-call timer at the moment the call enters `InProgress`. If the call is still `InProgress` when the timer fires, route through `AcpThread::cancel` and tag the call accordingly.

I'm happy to send a PR for this — wanted to file it first in case the team has a preferred design (per-tool inactivity vs. absolute, default value, where to surface the "timed out" status, etc.).

### Zed version and system specs

[...]

PR #56736, opened 2026-05-14
acp_thread: Auto-cancel hung external agent tool calls

beardfaceguy:
## Summary

Fixes part of #56734.

External ACP agents (Cursor, Claude Code, Gemini CLI, etc.) own their own tool execution. When one of their tools — most commonly `web_search` — stalls inside the agent process, Zed has no way to recover: the only signal that a tool finished is a `session/update` from the agent, and none ever arrives. The user is left staring at a spinner with no deadline, and the orphaned ACP child process keeps running long after the conversation has been abandoned.

This PR adds a single opt-in setting:

```json
{
  "agent": {
    "external_agent_tool_timeout_seconds": 120
  }
}
```

When set, any tool call that has been in `InProgress` for that many seconds without a status change is automatically cancelled via `session/cancel` (the same primitive Escape uses) and reported as `Canceled` with a `log::warn!` explaining why. The default is `None`, preserving today's behavior of waiting indefinitely.

## Design notes

- The setting is read at the moment the call enters `InProgress`, not when the timer fires, so the deadline reflects the user's intent at that point and isn't perturbed by later setting edits during the same turn.
- Per-call timer tasks are kept in a `HashMap<ToolCallId, Task<()>>` on `AcpThread`. Transitioning out of `InProgress` drops the entry and the GPUI task is cancelled with it. No external bookkeeping required.
- ACP only exposes a session-level cancel primitive (no per-call cancel), so when the timer fires it cancels the entire current turn — every other in-flight tool call on that session is marked `Canceled` along with the timed-out one. This is the same trade-off that pressing Escape makes, and reflects the protocol's actual capabilities.
- `0` is treated as "disabled" along with `None` to avoid a footgun where someone sets it to `0` expecting "cancel immediately".

## Known limitation / follow-up

Timed-out calls reuse `ToolCallStatus::Canceled` and surface only via `log::warn!`. Adding a distinct `ToolCallStatus::TimedOut` variant — with UI affordances to explain the cause inline in the thread — touches roughly a dozen call sites that currently match against `Canceled`, so I left it for a focused follow-up. The log line is enough to confirm the cause when triaging.

## Test plan

Added 4 GPUI tests in `acp_thread::tests` covering:

- `external_agent_tool_timeout_cancels_after_threshold` — call stays InProgress past the deadline; `connection.cancel` is invoked exactly once and the call is marked `Canceled`.
- `external_agent_tool_timeout_does_not_fire_when_call_completes` — call completes before the deadline; no cancel.
- `external_agent_tool_timeout_disabled_by_default` — setting unset; advancing the clock by an hour produces no cancel.
- `external_agent_tool_timeout_respects_subsequent_status_change` — call transitions `InProgress` → `Failed` before the deadline; the in-flight timer task must be dropped so it cannot subsequently call `cancel`.

`FakeAgentConnection::cancel` was extended to record invocations so these tests can assert on it.

```
cargo test -p acp_thread external_agent_tool_timeout
running 4 tests
test tests::external_agent_tool_timeout_cancels_after_threshold ... ok
test tests::external_agent_tool_timeout_does_not_fire_when_call_completes ... ok
test tests::external_agent_tool_timeout_respects_subsequent_status_change ... ok
test tests::external_agent_tool_timeout_disabled_by_default ... ok
```

`./script/clippy --package acp_thread --package agent_settings --package settings_content --package agent --package agent_ui` is also clean.

## Suggested .rules additions

None — nothing in this change felt non-obvious enough to be a repeated trap. The `Task<()>`-as-cancellation-handle pattern was already established in this crate.

Release Notes:

- Added `agent.external_agent_tool_timeout_seconds` to bound how long Zed waits on an external ACP agent (Cursor, Claude Code, Gemini CLI, etc.) tool call before sending `session/cancel`. Defaults to no timeout.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,6 +8,7 @@ version = "0.1.0"
 dependencies = [
  "action_log",
  "agent-client-protocol",
+ "agent_settings",
  "anyhow",
  "async-channel 2.5.0",
  "base64 0.22.1",
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -1102,6 +1102,13 @@
         // },
       },
     },
+    // When set, automatically cancel an external ACP agent (e.g. Cursor,
+    // Claude Code, Gemini CLI) tool call that has been running without
+    // progress for more than this many seconds. The cancellation is sent
+    // via `session/cancel` (the same path as Escape), so it cancels the
+    // entire current turn — ACP has no per-call cancel primitive. Leave
+    // null to wait indefinitely.
+    "external_agent_tool_timeout_seconds": null,
     // When enabled, agent edits will be displayed in single-file editors for review
     "single_file_review": false,
     // When enabled, show voting thumbs for feedback on agent edits.
--- a/crates/acp_thread/Cargo.toml
+++ b/crates/acp_thread/Cargo.toml
@@ -18,6 +18,7 @@ test-support = ["gpui/test-support", "project/test-support", "dep:parking_lot", 
 [dependencies]
 action_log.workspace = true
 agent-client-protocol.workspace = true
+agent_settings.workspace = true
 async-channel.workspace = true
 base64.workspace = true
 anyhow.workspace = true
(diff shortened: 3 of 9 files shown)


beardfaceguy:
@cla-bot check

benbrandt:
Looked into this, seems to be an agent issue

PR #56742, opened 2026-05-14
project: Resolve $keychain references in MCP context_servers settings

wock9000:
Self-Review Checklist:

 

- [x]  Performance impact has been considered and is acceptable

## Summary

Adds support for `{ "$keychain": "<name>" }` references in `context_servers.<id>.settings`, resolving each to a secret in the system keychain so MCP API tokens no longer have to live in cleartext in `settings.json`.

Previously, integrating an MCP server that needs an API token (Dune, Trongrid, etc.) required writing the token plaintext into `settings.json`. This makes settings backup/version-control awkward and risks accidental commit of secrets. The keychain reference syntax routes through the same `CredentialsProvider` already used for AI provider keys.

## Behavior

- A settings JSON object that has exactly the single key `$keychain` whose value is a string is treated as a reference. The reference walker replaces such objects with the secret at `zed://context_servers/<name>` via the existing `CredentialsProvider` used for AI provider keys.
- Cleartext values keep working — user JSON that happens to include a `$keychain` key alongside other keys is **not** consumed (the strict shape check).
- Recurses into nested objects and arrays, so references work anywhere in the settings tree.

## Failure modes

- Missing keychain entry produces a typed `MissingKeychainEntry` error wrapped in `anyhow::Error`. The caller logs the exact platform-native command that would provision the entry (e.g. `security add-generic-password -s 'zed://context_servers/<name>' -a 'zed' -w <secret>` on macOS) and leaves the server stopped rather than starting it with a partial config.

## Tests

Six unit tests in `crates/project/src/keychain_refs.rs` covering:
- single-ref replacement
- recursion into nested objects
- recursion into arrays
- no-refs pass-through path
- missing-entry error including `downcast_ref::<MissingKeychainEntry>()`
- "object with extra keys is not a ref" case

## Docs

`docs/src/ai/mcp.md` gets a section on the syntax with provisioning commands for macOS / Linux / Windows.

## Release Notes

- Added support for `{ "$keychain": "<name>" }` references in `context_servers.<id>.settings`, resolving each to a secret in the system keychain so MCP API tokens no longer have to live in cleartext in `settings.json`.


wock9000:
@cla-bot check

wock9000:
Opened #56870 to discuss the design (scope, syntax, write-path UX, naming). Converting this PR to draft until a maintainer weighs in — happy to scope down or split if helpful. Polish in progress: extending resolution to `Http.headers`, fixing platform-specific log message, and adding the `$keychain` shape to the settings JSON schema.

PR #56744, opened 2026-05-14
markdown_preview: Independent code font family

wln:
Adds a `markdown_preview_code_font_family` setting that overrides the font used in the markdown preview for code — inline `code` spans and fenced code blocks.
This is the counterpart to `markdown_preview_font_family` (#54003), which already does this for body text. Together they let you choose a typographically matched body + code font pair for the preview without forcing the code font onto editor buffers, where it may not be a good coding font:

```json
{
  "markdown_preview_font_family": "Noto Serif",
  "markdown_preview_code_font_family": "Noto Sans Mono"
}
```

Behavior mirrors `markdown_preview_font_family`:
- Scoped to the markdown preview only (`MarkdownFont::Preview`). The agent panel, notifications, hover popovers, and REPL output are unaffected — they keep using the buffer font for code.
- Falls back to the buffer font family when unset, so existing previews are unchanged.
- Overrides the font family only; fallbacks and features still come from the buffer font.

Before (uses buffer font, here Iosevka):

Screenshot 2026-05-14 at 1 39 51 PM

After (uses specified font, here Noto Sans Mono):

Screenshot 2026-05-14 at 1 40 51 PM

Release Notes:

- Added `markdown_preview_code_font_family` to override the code font in the markdown preview

### DIFF
--- a/crates/markdown/src/markdown.rs
+++ b/crates/markdown/src/markdown.rs
@@ -179,6 +179,11 @@ impl MarkdownStyle {
         } else {
             theme_settings.ui_font.family.clone()
         };
+        let code_font_family = if is_preview {
+            theme_settings.markdown_preview_code_font_family().clone()
+        } else {
+            theme_settings.buffer_font.family.clone()
+        };
 
         let text_color = colors.text;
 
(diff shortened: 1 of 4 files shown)


ChristopherBiscardi (approved):
thanks!

Issue #21054, opened 2024-11-22
CamelHump treatment for alt-[arrows] keys


MihaelBercic:
### Describe the feature

When navigating code, navigating using option-arrow command is a big part of the process.
I'm coming from JetBrains' IDEs and having the editor treat `_` (underscores) as word separators is a very important part when it comes to languages where snake_case is the syntax.

In JetBrains editors, the setting is called `Settings|Editor|Smart keys|Use "CamelHumps" words` **CamelHump**. Which treats names such as `thisFeature` as two words: `this|Feature` and `this_feature` as two words: `this|_|feature` (| indicating where the cursor stops when navigating with opt-arrow).

Thank you.

### Environment

Zed: v0.162.5 (Zed)
OS: macOS 15.1.1
Memory: 64 GiB
Architecture: aarch64

notpeter:
Gotcha, so basically you'd like to have a setting which alters the word boundary behavior such that you can do word navigation (`alt-left`, `alt-right`) which considers `CamelCaseWords` or `snake_case_words` as word boundaries.

MihaelBercic:
Yes, that is correct.

Link to another (more popular issue) https://github.com/zed-industries/zed/issues/12816

velixor:
I'm not sure, but maybe what you're looking for is Subwords. Try searching for that word in the file [default-macos.json](https://github.com/zed-industries/zed/blob/main/assets/keymaps/default-macos.json) or in [vim.json](https://github.com/zed-industries/zed/blob/main/assets/keymaps/vim.json).  

If, for example, someone wants to bring [**JetBrains** `[w`, `]w`, `[b`, `]b` word motions](https://ideavim.sourceforge.net/vim/motion.html#:~:text=%5Bw*-,%5Bw%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%5Bcount%5D%20camel%20words%20forward.%20%20%7Cexclusive%7C%20motion.%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7Bnot,camel%20word%20%5Bcount%5D%20(inclusive).%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7Bnot%20in%20Vim%7D.,-These%20commands%20move) into Zed, then these mappings are for you

```json

{
[...]

PR #56756, opened 2026-05-14
editor: Add camelCase subword expand selection with opt-in setting

jstano:
I switched to Zed from IntelliJ and this is one of the few things keeping me from going all-in. VSCode and Sublime have the same gap: `SelectLargerSyntaxNode` jumps from cursor to full word, skipping camelCase humps entirely. In IntelliJ, Ctrl+W steps through `camel` → `camelCase` → `camelCaseVar` → syntax node, which is the motion I reach for dozens of times a day when editing identifiers.

This PR adds that behavior behind an opt-in setting `use_expand_selection_by_subwords` (default `false`), so it's zero-risk for existing users.

Related to #21054.

When enabled, `SelectLargerSyntaxNode` follows the sequence:

1. cursor → camelCase/underscore subword (e.g. `camel` in `camelCaseVar`)
2. subword → full word (`camelCaseVar`)
3. word → syntax node (existing behavior)

The subword boundaries are detected using the existing `is_subword_start`/`is_subword_end` logic already in `movement.rs`. The new `surrounding_subword` function walks backward and forward from the cursor position, stopping at camelCase transitions, underscore boundaries, and non-word characters.

**To enable:**
```json
{
  "use_expand_selection_by_subwords": true
}
```

Self-Review Checklist:

- [X] I verified via a local build that the feature works as expected, with the flag enabled and disabled.

Release Notes:

- Added opt-in `use_expand_selection_by_subwords` setting that makes `SelectLargerSyntaxNode` step through camelCase/underscore subword boundaries before expanding to the full word, mirroring IntelliJ's Ctrl+W behavior


jstano:
@cla-bot check

jstano:
@ChristopherBiscardi Any thoughts on this PR?

jstano:
@ChristopherBiscardi Just checking if this direction makes sense before I continue polishing it. Happy to adjust the implementation if there is a preferred approach for selection expansion behavior.

ChristopherBiscardi (changes requested):
@jstano This PR description doesn't make it clear why we would want this when the functionality already exists with C+alt+arrow keys, optionally with shift to expand selections.

ChristopherBiscardi:
this functionality already exists, and the PR doesn't make it clear how this would be different from that existing functionality right now, so I'm going to close for now. If this is meaningfully different we can consider reopening

jstano:
@ChristopherBiscardi The goal is to enhance the editor::SelectLargerSyntaxNode to match the approach IntelliJ uses. I've mapped "ctrl-w" to "editor::SelectLargerSyntaxNode". Here's an example, in the XML file when I have my cursor in the enumType="GenderType" text, say on the "e"in GenderType. With the current behavior when I hit ctrl-w the first time, it expands the selection to include the quotes. This is not ideal when I want to cut/copy/modify the text in the quotes. In IntelliJ with their "enable camel humps" flag enabled, ctrl-w would first expand the selection to Gender, 2nd time expands to GenderType, 3rd time expands to "GenderType", 4th time expands to enumType="GenderType", 5th time expands to <column name="Gender" type="enum" enumType="GenderType"/> and so on.

```
<columns>
  <column name="ID" type="sequence" required="true"/>
  <column name="Name" type="varchar" length="100" required="true"/>
  <column name="Extra" type="varchar" length="200"/>
  <column name="Gender" type="enum" enumType="GenderType"/>
</columns>
```

The same behavior works great in Java, Rust, etc with their identifiers.

I you apply my patch you'll get a better feel for the tweak to the behavior. Users coming from Jetbrains tools rely on this a lot, it's been there forever.

the ctrl+alt+shift+arrow key is not nearly as smooth of an experience. I feel this one change makes the editor have a much smoother experience, especially from developers using the Jetbrains tools.

PR #56762, opened 2026-05-14
Migrate Default Rules to global AGENTS.md

rtfeldman:
Stacked on top of #56672 (AI-227).

Extends the rules-to-skills migration so that, in the same migration pass, every Default user Rule is appended to the global `AGENTS.md` (at `paths::agents_file()` — e.g. `~/.config/zed/AGENTS.md` on macOS/Linux, `%APPDATA%\Zed\AGENTS.md` on Windows). Each rule lands under an `## H2` heading containing the rule's title, followed by the rule body, with rules separated by blank lines.

* If `AGENTS.md` doesn't exist yet, it's created with just the migrated rules.
* If it exists, existing content is preserved verbatim and the migrated rules are appended after a blank-line separator.
* Whitespace-only existing files are treated as empty.

The split mirrors the original semantics of each Rule type:

* **Default Rules** were auto-included in every conversation → global `AGENTS.md` (loaded into the system prompt of every conversation) preserves the behavior.
* **Non-Default Rules** (handled in #56672) were only included when explicitly invoked → global Skills with `disable-model-invocation: true` preserves the behavior.

Same `skills` feature flag and same `rules_to_skills_migration_done` KVP flag as the parent PR — both migrations run in the same pass, never separately. The "Skills have replaced Rules" modal copy is updated to mention both destinations.

7 new tests cover heading/body rendering, body trimming, empty input, file creation when missing, append-to-existing, whitespace-only existing, and no-op for empty rules.

Closes AI-232

Release Notes:

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12273,6 +12273,7 @@ dependencies = [
 name = "paths"
 version = "0.1.0"
 dependencies = [
+ "const_format",
  "dirs",
  "ignore",
  "util",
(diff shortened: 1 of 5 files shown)


cole-miller:
@zed-industries/approved

PR #56773, opened 2026-05-14
context_server: Reset MCP request idle timer on inbound activity

### LINKED ISSUE #56774: MCP context server requests cancelled mid-flight despite progress notifications
### Check for existing issues

- [x] Completed

Long-running MCP tool calls get cancelled the moment they cross `context_server_timeout` (default 60s) even when the server is actively working and streaming `notifications/progress` back to Zed. On HTTP-transport servers this surfaces in the logs as:

```
http/2 stream closed with error code CANCEL (0x8)
```

The cancellation is initiated client-side: Zed's request future hits its fixed `executor.timer(timeout)` and tears down the channel, which on HTTP/2 manifests as a CANCEL stream reset.

**Repro**

1. Configure any MCP server that emits `notifications/progress` during long tool calls. (Concrete examples: a server that runs `cargo test` on a large workspace, a code-review server that does multi-file analysis, anything that legitimately runs longer than 60s.)
2. Leave `context.servers.<name>.timeout` at the default (or any value shorter than the tool's real runtime).
3. Invoke a tool whose work exceeds the timeout.

**Expected:** the request stays alive as long as the server keeps emitting frames (progress, partial results, anything). Only a server that has actually gone silent past `timeout` should be cancelled.

**Actual:** the request is cancelled at exactly `timeout` regardless of how much progress the server has streamed in the meantime. On HTTP transports this also produces a `CANCEL (0x8)` on the wire.

### Root cause

`Client::request_with` in `crates/context_server/src/client.rs` arms a single `executor.timer(timeout)` at request time and races the response against it in a `select!`. Inbound progress notifications arrive in `handle_input` and get dispatched through `notification_handlers`, but they have no path to bump the per-request timer. The timer fires on wall-clock elapsed since the request was sent, not on the server going silent.

### Fix

PR #56773 tracks the timestamp of the last inbound JSON-RPC frame (any kind — response, request, OR notification) on the shared `Client` and re-arms the request timer as `last_activity + timeout`. Any inbound frame extends the deadline; a server that genuinely stops talking still bails after `timeout` of true silence.

This is the stdio half of the broader \"requests cancelled mid-flight\" problem. The HTTP-transport CANCEL errors users see when laptops sleep / stream connections drop are a related but separate failure mode (the SSE long-poll dies and is not resurrected); that will be addressed in a follow-up.

### Environment

Zed `main` (reproduced against today's tip). Affects all MCP transports, but HTTP/2 is where it shows up most loudly because the cancellation surfaces as a visible `CANCEL (0x8)` in the logs.

N/A

### Zed.log

Representative line from the user-affected logs:

```
http/2 stream closed with error code CANCEL (0x8)
```

beardfaceguy:
Closes #56774.

## Problem

`context_server_timeout` (default 60s) is enforced in `Client::request_with` as a single `executor.timer(timeout)` armed at request time. There is no way for an MCP server to signal "still alive, just busy."

This means any long-running tool call — `cargo test`, large refactors, codebase reviews, anything that legitimately runs beyond the timeout — gets cancelled even when the server is actively working and emitting `notifications/progress`. Users see the request fail; depending on the transport, the server may also see an `http/2 stream closed with error code CANCEL (0x8)` because the cancellation tears the stream down.

## Fix

Track an `inbound_activity_at: Arc<Mutex<Instant>>` on the shared `Client`. `handle_input` updates it on every received JSON-RPC frame — response, request, OR notification. `request_with` computes its deadline as `last_activity + timeout` and re-arms the timer each iteration of a small loop, so any inbound activity (including `notifications/progress`) extends the deadline. A genuinely silent server still bails after `timeout` of true silence, preserving the protective behavior for misbehaving servers.

The timestamp uses `BackgroundExecutor::now()` so it participates correctly in GPUI test executor's fake clock.

## Why this approach

- **No protocol plumbing.** We don't have to thread MCP progress tokens through to individual request futures or hold per-request bookkeeping for which server-emitted notification belongs to which request. Any frame counts as liveness, which matches the underlying invariant we actually care about (the server is still writing to the pipe).
- **Backward compatible.** A server that emits no progress and finishes within `timeout` behaves identically. A misbehaving server that genuinely stops talking still gets cut off. The only behavior change is the win case: a busy server can hold the request open as long as it keeps writing.
- **Minimal surface area.** All changes are inside `crates/context_server`. No settings change, no UI change, no protocol change.

## Tests

Three new tests in `crates/context_server/src/client.rs` driven via a new `FakeTransport::inject_message` helper (so tests can deliver unsolicited notifications without going through `on_request`):

- `request_times_out_when_server_is_silent` — preserves old behavior on truly silent servers.
- `progress_notifications_keep_request_alive_past_timeout` — request survives well past the silence threshold as long as progress notifications keep arriving.
- `request_times_out_after_progress_stops` — once the server stops emitting frames, the timer fires within `~timeout` of the last frame.

All 97 `context_server` tests pass; clippy clean.

Release Notes:

- Fixed MCP context server requests being cancelled mid-flight when a long-running tool call exceeded `context_server_timeout` despite the server emitting `notifications/progress`. The idle timer is now reset by any inbound JSON-RPC frame, so progress-emitting servers can run as long as they keep talking.

### DIFF
--- a/crates/context_server/src/client.rs
+++ b/crates/context_server/src/client.rs
@@ -62,6 +62,12 @@ pub(crate) struct Client {
     #[allow(dead_code)]
     transport: Arc<dyn Transport>,
     request_timeout: Option<Duration>,
+    /// Wall-clock-equivalent timestamp of the most recent inbound JSON-RPC
+    /// frame from the server (response, request, OR notification — anything).
+    /// `request_with` consults this to decide whether the server is genuinely
+    /// silent or just slow-but-talking. Stored as the executor's clock so
+    /// `cx.executor().advance_clock()` works in tests.
+    inbound_activity_at: Arc<Mutex<Instant>>,
     /// Single-slot side channel for the last transport-level error. When the
     /// output task encounters a send failure it stashes the error here and
     /// exits; the next request to observe cancellation `.take()`s it so it can
@@ -205,18 +211,21 @@ impl Client {
         let response_handlers =
             Arc::new(Mutex::new(Some(HashMap::<_, ResponseHandler>::default())));
         let request_handlers = Arc::new(Mutex::new(HashMap::<_, RequestHandler>::default()));
+        let inbound_activity_at = Arc::new(Mutex::new(cx.background_executor().now()));
 
         let receive_input_task = cx.spawn({
             let subscription_set = subscription_set.clone();
             let response_handlers = response_handlers.clone();
             let request_handlers = request_handlers.clone();
             let transport = transport.clone();
+            let inbound_activity_at = inbound_activity_at.clone();
             async move |cx| {
                 Self::handle_input(
                     transport,
                     subscription_set,
                     request_handlers,
                     response_handlers,
+                    inbound_activity_at,
                     cx,
                 )
                 .log_err()
@@ -259,6 +268,7 @@ impl Client {
             transport,
             request_timeout,
             last_transport_error,
+            inbound_activity_at,
         })
     }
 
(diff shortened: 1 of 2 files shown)


ChristopherBiscardi:
Thanks for the batch of PRs in this area. We'll need a reproduction case for #56774 before reviewing any PRs that would apply to fix it.

PR #56776, opened 2026-05-14
context_server: Resume MCP SSE streams after abrupt termination

### LINKED ISSUE #56775: MCP HTTP transport: SSE stream death silently hangs in-flight requests (no resumption)
### Check for existing issues

- [x] Completed

When using an MCP context server over HTTP/streamable-HTTP transport, the SSE response stream associated with a POST can be torn down mid-flight by:

- HTTP/2 `RST_STREAM` with `CANCEL (0x8)` from the server or an intermediary
- Network disconnect / interface change / VPN reconnect
- Idle keepalive on a load balancer (CloudFront, ALB, Cloudflare, etc.)
- Laptop sleep/resume

When this happens the in-flight request does not fail and does not retry. The agent UI sits on a spinner forever (or, after #56773, fails after `context_server_timeout` of silence). The Zed log shows:

```
http/2 stream closed with error code CANCEL (0x8)
```

The MCP streamable-HTTP transport spec §4.5 ("Resumability and Redelivery") describes exactly this scenario and how clients should recover: the server emits `id:` fields on SSE events, and clients SHOULD reconnect via a GET to the MCP endpoint with `Last-Event-ID` to resume. Today's transport does neither — it explicitly ignores `id:` lines and never issues a GET.

**Repro**

1. Connect to any MCP server speaking the streamable-HTTP transport (e.g. anything fronted by an HTTP/2 server with idle stream timeout < `context_server_timeout`).
2. Invoke a tool whose work runs long enough for the connection to be severed by an intermediary, or just disconnect/reconnect the network mid-call.
3. Observe the request hangs.

**Expected:** the transport detects the abnormal termination and resumes via GET + `Last-Event-ID`, transparently delivering the final response.

**Actual:** the spawned SSE reader task dies, an error is forwarded to an internal channel that nothing actually consumes for recovery, and the request hangs.

### Root cause

`HttpTransport::setup_sse_stream` in `crates/context_server/src/transport/http.rs`:

- Treats any `Err` from the body reader as terminal (sends to `error_tx`, breaks the loop).
- Has no notion of `Last-Event-ID` — explicitly skips `id:` lines alongside `event:` and `retry:`.
- Never issues a GET to the MCP endpoint, so even servers that fully support resumption (most production MCP-over-HTTP servers do) get no chance to replay.

### Fix

PR #56776 implements the spec's resume path: capture `id:` from SSE events, and on abnormal stream termination issue a single GET reconnect with `Last-Event-ID`. Clean EOF (the normal end of a per-POST stream) is left alone. 404/405 from the resume GET is treated as "server doesn't support resumability" and surfaces a clean error rather than retrying.

Pairs with #56774 — that one prevents Zed from being the *cause* of the cancellation; this one recovers when the cancellation comes from outside.

### Environment

Zed `main` (today's tip). HTTP/streamable-HTTP MCP transport only — stdio transport is unaffected.

N/A

### Zed.log

```
http/2 stream closed with error code CANCEL (0x8)
```

beardfaceguy:
Closes #56775.

## Problem

The HTTP transport's SSE reader treats any read error as terminal: it forwards `"SSE stream error: ..."` to `error_tx` (which nothing consumes for recovery), breaks the loop, and lets the spawned task die. Anything waiting on the in-flight request hangs until `context_server_timeout` expires — visible in the wild as `http/2 stream closed with error code CANCEL (0x8)` followed by a request that never completes. The dominant triggers are HTTP/2 keepalive on intermediaries (CloudFront, ALB, Cloudflare, etc.), VPN reconnects, and laptop sleep/resume.

The MCP streamable-HTTP transport spec §4.5 (\"Resumability and Redelivery\") describes exactly this scenario and the recovery mechanism: servers emit `id:` fields on SSE events, and clients SHOULD reconnect via GET to the MCP endpoint with `Last-Event-ID` to resume. Today's transport does neither — it explicitly ignores `id:` lines.

## Fix

Implement the spec's resume path:

* Capture `id:` lines instead of ignoring them, holding the most recent value in the SSE-pump's local state.
* On `Err` from the body reader, if we have a session ID and saw at least one event ID, issue a GET to the MCP endpoint with `Last-Event-ID` and continue pumping from the resumed stream.
* On clean EOF, do nothing — the server intentionally closed the per-POST stream after delivering its final response. Reissuing a GET would either re-execute work or pester servers that don't support resumption.
* Cap blast radius: if the GET returns 404/405 the server explicitly doesn't support resumability and we give up cleanly; any other failure is surfaced verbatim to `error_tx` alongside the original error. There is **no** exponential retry loop — one attempt is enough to handle the transient-disconnect case without risking hot-loops against a permanently-broken endpoint.

The SSE line parsing is refactored into `pump_sse_body` so the resume codepath drives a fresh body through the same parser, and the GET construction is extracted into `attempt_sse_resume` so it has a single testable surface.

## Why one retry, not many

The empirical failure mode is a single transient disconnect (HTTP/2 idle keepalive, sleep/resume, VPN flap). Once recovered, the server reliably keeps the connection up for the rest of the call. Adding exponential backoff would increase blast radius against pathologically broken servers — and a permanently-broken endpoint should fail fast so the user sees it, not retry-storm in the background.

## Tests

Two new tests in `crates/context_server/src/transport/http.rs` driven by a tiny `BrokenAfter` `AsyncRead` helper that yields a fixed slice then returns `io::Error::ConnectionReset` (simulating HTTP/2 CANCEL):

* `sse_stream_error_triggers_get_reconnect_with_last_event_id` — POST returns SSE that emits `id: 1` then errors. Asserts a follow-up GET is issued with `Last-Event-ID: 1` and the resumed stream's response reaches `response_rx`.
* `sse_clean_eof_does_not_trigger_reconnect` — POST returns a complete SSE response and EOFs cleanly. Asserts no follow-up GET — confirms we don't pester servers on the normal terminal path.

All 96 `context_server` tests pass; clippy clean.

## Related

Pairs with #56773 / #56774 — that PR prevents Zed from being the *cause* of cancellations on long-running tool calls (silence-timeout reset by progress notifications); this PR recovers when the cancellation comes from outside (network, intermediary, sleep/resume).

Release Notes:

- Fixed MCP context server requests over HTTP transport hanging when the SSE response stream was severed mid-flight (HTTP/2 `CANCEL`, network blip, intermediary keepalive, laptop sleep). Per the streamable-HTTP spec, the transport now resumes via GET with `Last-Event-ID` after abnormal termination.

### DIFF
--- a/crates/context_server/src/transport/http.rs
+++ b/crates/context_server/src/transport/http.rs
@@ -251,63 +251,99 @@ impl HttpTransport {
     }
 
     /// Set up SSE streaming from the response
-    async fn setup_sse_stream(&self, mut response: Response<AsyncBody>) -> Result<()> {
+    async fn setup_sse_stream(&self, response: Response<AsyncBody>) -> Result<()> {
         let response_tx = self.response_tx.clone();
         let error_tx = self.error_tx.clone();
+        let http_client = self.http_client.clone();
+        let endpoint = self.endpoint.clone();
+        let session_id = self.session_id.clone();
+        let protocol_version = self.protocol_version.clone();
+        let headers = self.headers.clone();
+        let token_provider = self.token_provider.clone();
 
-        // Spawn a task to handle the SSE stream
         self.executor
             .spawn(async move {
-                let reader = futures::io::BufReader::new(response.body_mut());
-                let mut lines = futures::AsyncBufReadExt::lines(reader);
+                let mut current = response;
+                let mut last_event_id: Option<String> = None;
 
-                let mut data_buffer = Vec::new();
-                let mut in_message = false;
+                loop {
+                    let outcome = pump_sse_body(
+                        current.body_mut(),
+                        &response_tx,
+                        &mut last_event_id,
+                    )
+                    .await;
 
-                while let Some(line_result) = lines.next().await {
-                    match line_result {
-                        Ok(line) => {
-                            if line.is_empty() {
-                                // Empty line signals end of event
-                                if !data_buffer.is_empty() {
-                                    let message = data_buffer.join("\n");
-
-                                    // Filter out ping messages and empty data
-                                    if !message.trim().is_empty() && message != "ping" {
-                                        if let Err(e) = response_tx.send(message).await {
-                                            log::error!("Failed to send SSE message: {}", e);
-                                            break;
-                                        }
-                                    }
-                                    data_buffer.clear();
-                                }
-                                in_message = false;
-                            } else if let Some(data) = line.strip_prefix("data: ") {
-                                // Handle data lines
-                                let data = data.trim();
-                                if !data.is_empty() {
-                                    // Check if this is a ping message
-                                    if data == "ping" {
-                                        log::trace!("Received SSE ping");
-                                        continue;
-                                    }
-                                    data_buffer.push(data.to_string());
-                                    in_message = true;
-                                }
-                            } else if line.starts_with("event:")
-                                || line.starts_with("id:")
-                                || line.starts_with("retry:")
-                            {
-                                // Ignore other SSE fields
-                                continue;
-                            } else if in_message {
-                                // Continuation of data
-                                data_buffer.push(line);
-                            }
-                        }
-                        Err(e) => {
-                            let _ = error_tx.send(format!("SSE stream error: {}", e)).await;
+                    match outcome {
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Thanks for the batch of PRs in this area. We'll need a reproduction case for #56774 before reviewing any PRs that would apply to fix it.

PR #56779, opened 2026-05-14
Prefix read_file tool output with line numbers

MartinYe1234:
Format `read_file` tool output in `cat -n` style: each line is prefixed with its line number right-aligned in a 6-character field, followed by a single tab, followed by the line's original content (newlines preserved, including CRLF). Numbering reflects the actual file lines, so a ranged read starting at line 42 emits `42` for its first line, not `1`.

For large files, content returned by `get_buffer_content_or_outline` is **not** prefixed:

- The symbol outline path already conveys structure via its `[L100-150]` annotations.
- The truncated first-1KB fallback (used when a file exceeds `AUTO_OUTLINE_SIZE` and has no parseable outline) is wrapped in a synthetic `# First 1KB of …` header, so its lines don't correspond to real file line numbers.

Both cases are reported via `BufferContent::is_synthetic` (renamed from `is_outline`).

Also updates the `edit_file` tool's input doc to describe the prefix format and tell the model to strip it before constructing `old_text` / `new_text`, preserving the original indentation that appears after the tab.

Updates how we render `read_file` tool call outputs in the UI (screenshots included in comments below).
Also fixes an existing bug where `read_file` tool call outputs would not re-render their content code block when an older thread was restored (the tool's `replay` hook was missing).

Closes AI-226

Release Notes:

- Improved how `read_file` tool output renders in the agent panel, with a line-number gutter, and fixed it not re-rendering on restored threads


MartinYe1234:
Need to fix:
Tool output rendering
Run Evals to ensure no regressions

MartinYe1234:
Currently, we do not handle Read File Output properly for file outlines.
image

MartinYe1234:
This PR also fixes the issue where the Read File tool output doesn't render properly when opening a tool call ui item on an older message thread:
Before:
image


Now:
image

crates/agent/src/tools/read_file_tool.rs
```diff
@@ -18,6 +18,118 @@ fn tool_content_err(e: impl std::fmt::Display) -> LanguageModelToolResultContent
...
+    use std::fmt::Write as _;
+
+    let mut line_number = start_line;
+    let mut at_line_start = true;
+    for chunk in chunks {
+        let mut rest = chunk;
+        while !rest.is_empty() {
+            if at_line_start {
+                // Writes to a `String` are infallible, so the `Result` can be ignored.
+                let _ = write!(output, "{line_number:>6}\t");
```
benbrandt, line 75: Would it make more sense to have rather than a hard-coded 6, this be the length of the greatest line number?

crates/agent/src/tools/read_file_tool.rs
```diff
@@ -110,26 +218,60 @@ pub struct ReadFileToolInput {
...
+    pub fn new(
+        project: Entity<Project>,
+        action_log: Entity<ActionLog>,
+        update_agent_location: bool,
+    ) -> Self {
+        Self(ReadFileTool::with_output_format(
+            project,
+            action_log,
+            update_agent_location,
+            ReadFileOutputFormat::LineNumbered,
```
benbrandt: Thinking about this some more, maybe the feature flag is overkill. 
The eval showed now loss of quality, and we'll have time to test it before it lands on stable and do any tweaks.

crates/agent/src/tools/read_file_tool.rs
benbrandt: I guess this is what cat -n does.. so maybe it is best to leave it at that. It just is more tokens :D

benbrandt (approved):
I know i said to feature flag it... but given that the models are all passing `-n` to every tool already, and evals looked good, I'm actually fine just rolling with it since it effects multiple parts of the system. We'll have time to test it out.

Thanks 💪🏻

crates/agent/src/tools/read_file_tool.rs
MartinYe1234: Yeah Ill undo the commit for the feature flag

crates/agent/src/tools/read_file_tool.rs
MartinYe1234: I can make it so that if the line number gets over 6 digits (into the millions) we switch over to the length of the greatest line number but I think at that point there may be other issues

benbrandt: Yeah we can leave as is

dandv:
@MartinYe1234 related? #57230

MartinYe1234:
@dandv Not related as this only affects how we render output from the read file tool. https://github.com/zed-industries/zed/issues/57230 seems to be an issue with the terminal tool not persisting output on restart.

PR #56782, opened 2026-05-14
Hide Rules UI when Skills feature flag is enabled

MartinYe1234:
Hides the legacy Rules surfaces when the `skills` feature flag is enabled, since Rules are surfaced through the Skills UI in that case:

- The **Rules** entry in the agent panel's ellipsis (triple-dot) dropdown menu is hidden.
- The `assistant: open rules library` command palette action is hidden.

The `@rules` autocomplete in the chat input is intentionally left alone for now.

### Implementation notes

`cx.has_flag::<SkillsFeatureFlag>()` reads from a global feature-flag store that is populated asynchronously from the server:

- **Command palette**: `update_command_palette_filter` is already re-run via `cx.on_flags_ready` once flags arrive, so the filter is reapplied with the correct value. No new wiring needed.
- **Dropdown menu**: the popover's menu closure runs on every open, so the latest flag value is reflected at click time.

Release Notes:

- N/A

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -4149,13 +4149,6 @@ impl AgentPanel {
             _ => false,
         };
 
-        // When the Skills feature flag is on, Rules are surfaced through
-        // the Skills UI instead, so we hide the legacy Rules menu entry.
-        // The flag is read from a global store populated asynchronously,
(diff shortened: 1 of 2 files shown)


MartinYe1234:
Short cut to open Rules (default: Command Option L) should be remapped to the new Skill UI

rtfeldman (approved):

PR #56785, opened 2026-05-14
python: Stop falling back to `conda activate base` for nameless toolchains

rgbkrk:
Stops the `conda activate base` fallback in `PythonToolchainProvider::activation_script` for two cases:

1. A Conda toolchain whose `environment.name` is `None`.
2. A Conda toolchain whose name is `Some(...)` but `shell.try_quote` rejects it.

Previously both cases pushed `conda activate base` into the activation script. With miniforge installed and no env explicitly selected, every terminal silently activated `(base)`. The user didn't pick `base`; injecting it is contamination.

Both branches now log a warning and emit no activation line. Users who want `(base)` can select it as the toolchain explicitly.

Two unit tests cover the new behavior: `test_conda_activation_skips_when_name_missing` (name=None) and `test_conda_activation_skips_unquotable_name` (name=Some but unquotable, exercised with an embedded NUL byte that `shlex::try_quote` rejects). The existing `test_conda_activation_script_injection` still verifies that quotable names get activated.

Release Notes:

- Fixed Zed silently injecting `conda activate base` into terminals when a Conda manager (miniforge/miniconda) was installed but no specific environment was selected.

### DIFF
--- a/crates/languages/src/python.rs
+++ b/crates/languages/src/python.rs
@@ -2878,6 +2878,76 @@ mod tests {
     }
 
     #[gpui::test]
+    async fn test_conda_activation_skips_unquotable_name(cx: &mut TestAppContext) {
+        use language::{LanguageName, Toolchain, ToolchainLister};
+        use settings::{CondaManager, VenvSettings};
+        use task::ShellKind;
+
+        use crate::python::PythonToolchainProvider;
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

PR #56792, opened 2026-05-14
agent_panel: Surface prompt navigation in a thread toolbar dropdown

thetaurean:
Adds a "Prompt History" dropdown to the active thread toolbar in the agent
panel. The dropdown lists every user prompt in the current thread; selecting
an entry scrolls that prompt to the top of the viewport.

https://github.com/user-attachments/assets/7adf97d1-ad56-4656-9ca0-1f52fd11f826

## Summary

- New `IconName::HistoryRerun` button in the right-aligned button group of
  `render_toolbar`, on the active-thread branch only. Disabled when there
  is no active thread; clicking on an active thread with zero prompts
  shows a disabled "No prompts yet" entry.
- The menu is built lazily on open from two new `pub(crate)` helpers on
  `ThreadView`:
  - `user_prompt_previews(&App) -> Vec<(usize, SharedString)>` — walks
    `thread.entries()`, picks out `UserMessage` variants, collapses
    multi-line prompts to the first non-empty line, truncates to 80
    chars with `…`.
  - `scroll_to_user_prompt(entry_ix, &mut Context<Self>)` — wraps the
    same `ListState::scroll_to(ListOffset { ... })` call already used
    by `scroll_to_most_recent_user_prompt`. Out-of-range index is a
    no-op.
- The chat list (`ListState`) is already lazy-loaded; only the loaded
  `entries()` slice is iterated, so no extra work was needed for
  pagination.

## Why

The agent panel already supports navigating between prior prompts via
the `ScrollOutputToPreviousMessage` / `ScrollOutputToNextMessage` actions
(default keybindings `ctrl-alt-shift-pageup` / `pagedown`, documented at
[agent-panel#navigating-the-thread][docs]). In practice almost nobody
discovers a four-modifier keybinding through the docs.

This is the gap users keep hitting in long threads — see the third ask
in discussion #55911 (originally filed as #55902, converted to a
discussion). Existing capability, missing visible affordance.

The dropdown gives that capability a one-click surface:

- **Discoverable** — an icon next to the existing toolbar controls,
  with a tooltip.
- **All prompts at once** — useful when you remember roughly *which*
  earlier prompt you want, but not how many steps back it is. The
  existing keybindings step one prompt at a time.

Open PR #55945 covers a different shape of the same problem: repeated
clicks on the existing "scroll to recent prompt" button cycle backward
through prior prompts. That's a nice keyboard-style stepper. The
dropdown here is complementary — direct jump vs. sequential step — and
the two compose cleanly. Happy to coordinate with that PR's author if
the maintainers prefer a single approach.

[docs]: https://zed.dev/docs/ai/agent-panel#navigating-the-thread

## Test plan

- New unit test `test_user_prompt_previews_and_scroll` in
  `crates/agent_ui/src/conversation_view.rs` covers:
  - Multi-line prompts collapsing to the first line.
  - Empty prompts being skipped.
  - `scroll_to_user_prompt` no-op on out-of-range indices.
- `cargo test -p agent_ui` is green.
- `./script/clippy -p agent_ui` reports no new warnings.
- Manual verification on macOS:
  - No active thread → button visible but disabled.
  - Active thread, zero prompts → menu shows disabled "No prompts yet".
  - 3+ prompts including a multi-line and an 80+ char prompt → entries
    appear in submission order, multi-line collapses to first line,
    long prompt truncates with `…`.
  - Selecting an entry from any scroll position scrolls the chosen
    prompt to the top of the viewport.

## Files

- `crates/agent_ui/src/agent_panel.rs` — new
  `prompt_history_menu_handle` field, `render_prompt_history_menu`
  helper, and the wiring in `render_toolbar`.
- `crates/agent_ui/src/conversation_view/thread_view.rs` — two new
  `pub(crate)` helpers next to `scroll_to_most_recent_user_prompt`.
- `crates/agent_ui/src/conversation_view.rs` — the new test plus a
  small `send_message` helper for the existing `tests` module.

## Out of scope

- A dedicated action / Command Palette entry to open the menu. Easy
  follow-up if desired — left out to keep this PR small.
- Pre-pagination history (only currently-loaded entries appear, matching
  the existing list semantics).
- Search-within-thread (a separate ask in #55911).

## Related

- Discussion #55911 (third ask).
- #55945 — overlapping intent, different UX (cycling on the existing
  scroll-to-recent-prompt button).

CLA: I'll accept the CLA prompt on this PR.

Release Notes:

- Agent: Added a prompt history dropdown to the thread toolbar for jumping to any prior user prompt in the active thread.


thetaurean:
@cla-bot check

benbrandt:
Thanks but I don't think we want to add this. cc @danilo-leal if you have any other ideas for how to solve this, if we want to

thetaurean:
No worries! Thanks for the consideration and time. I'm finding it really painful to scroll and navigate between prompts, especially with the lazy load. But obviously a lot of stuff is still maturing, I'm seeing more tool calls and long output getting collapses. The tool call output is really excessive and I would prefer if it defaulted to collapsed at least once it scrolls out of view (or cap the height and make it scrollable).

I also guess a lot of this is going to be shifting into the terminal view with Anthropic "axing" ACP.

PR #56803, opened 2026-05-14
gpui: Add minimalistic devtools

GoldStrikeArch:
cc @Anthony-Eid 

This is a minimalistic setup of GPUI devtools from this branch #55825

You can ignore the `crates/gpui/src/devtools/overlay.rs` file as it is the UI, everything else are the "profiler hooks" and the logic how we collect the frame stats, scene stats, etc

Self-Review Checklist:

Release Notes:

- N/A


MrSubidubi:
Curious though - we have the inspector already, why can this not be tied to that feature?

Anthony-Eid:
> Curious though - we have the inspector already, why can this not be tied to that feature?

I can look into that, but I feel like this feature is different enough that it makes sense to have them separate. Maybe I'm wrong tho

MrSubidubi:
I think I disagree there - while it definitely is different from the inspector we currently have, it definitely would be something I'd expect an inspector to have. If anything, I'd expect a feature that introduces some overhead to be gated behind a feature that I can opt out of - again, I think that would be the `inspector` feature already available for GPUI. Having some sort of inspection enabled/builtin by default, yet other similar tools behind a feature seems like a weird mixture of concerns to me, hence the question above.

Anthony-Eid:
@MrSubidubi You've convinced me for the most part, it should be under the same feature flag as the inspector

UI wise it could be a tab but I like having a separator window. Maybe this is something we should discuss more after I start my in depth review

GoldStrikeArch:
@MrSubidubi 
> Curious though - we have the inspector already, why can this not be tied to that feature?

The main reason is that I was not aware what is "inspector", otherwise it makes sense to have all "devtools" in one place. I wonder why "zed: open performance profiler" command is wired up differently (I also didn't see it in my production build, so I assume it is debug build only as well)

@Anthony-Eid 

> UI wise it could be a tab but I like having a separator window. 

Yeah, separate window works better as this can take a lot of space. But usually in such devtools setups you have the option to have them like a sidebar/or a separate window (like in the chrome devtools for example) and we can add something like this in the future

MrSubidubi:
> I wonder why "zed: open performance profiler" command is wired up differently (I also didn't see it in my production build, so I assume it is debug build only as well)

I think that probably is because the performance profiler is not as inefficient as the inspector is.

GoldStrikeArch:
@MrSubidubi 

> because the performance profiler is not as inefficient as the inspector is.

What do you mean by that? Can you please elaborate more on the "inefficient" part (I am just curious)

Anthony-Eid:
Hooking into the frame rendering code path adds overhead to a very hot code path (render loop). So it's "inefficient" just by the nature of what it's doing.

GoldStrikeArch:
@Anthony-Eid 

> Hooking into the frame rendering code path adds overhead to a very hot code path (render loop). So it's "inefficient" just by the nature of what it's doing.

I see. Well, this "re-render devtools" are doing the same but the overhead itself is tiny from my own measurements.

ChristopherBiscardi:
It seems like @Anthony-Eid has some overlapping work that will impact the ability of this PR to merge and we won't be able to review this PR in the near term. So I'm going to close it so that you're not waiting on a review/merging main to prep for a review that isn't going to come.

That said, after Anthony's work lands, we can revisit this.

PR #56812, opened 2026-05-14
ci: Prepare for Node.js 20 deprecation

MrSubidubi:
https://github.com/zed-industries/zed/actions/runs/25881329418 complains about the deprecation of Node.js version 20, which will soon be force-bumped to 24 in the runner contexts.

Thus, this PR bumps all the jobs mentioned there to versions that run on Node.js 24 by default. I checked all the release note section for the three actions and this should not change any behaviour according to their release notes and get us ready for the deprecation.

Release Notes:

- N/A


ConradIrwin (approved):
Thanks for tidying up the rust code a bit too

PR #56813, opened 2026-05-14
git_ui: Improve history tab focus display

danilo-leal:
Follow-up to https://github.com/zed-industries/zed/pull/56743 where we only show the focused styles in history tab items if you're navigating it with the keyboard. We were previously flashing the focused border styles on click and they're not necessary if you're pointer-based.

Release Notes:

- N/A


cole-miller:
@zed-industries/approved

PR #56818, opened 2026-05-15
Dismiss Settings overlay when invoking New Thread

rtfeldman:
When the Agent Panel had the Settings overlay open on top of the
ephemeral new-draft thread, pressing **Cmd-N** (or clicking *Start New
Thread* in Settings) silently did nothing — the action ran, but the
Settings overlay stayed painted on top.

The bug was in `AgentPanel::activate_draft`. Every other `activate_*`
path goes through `set_base_view`, which clears the overlay. But
`activate_draft` has an early return for the case where the base view
already holds the requested draft, and that early return only refocused
— it never touched `overlay_view`. So when the user was already on the
draft and the Settings overlay was on top, "new thread" had nothing
visible to do.

Fix: in that early-return branch, if an overlay is currently open, clear
it (which also handles focusing and emits `ActiveViewChanged`).

Added a regression test that puts the panel on the draft, sets the
Settings overlay, dispatches `NewThread`, and asserts the overlay is
gone. The test directly toggles `set_overlay(OverlayView::Configuration,
…)` rather than calling `open_configuration`, because the latter builds
provider configuration views and panics on the fake LM provider's
`unimplemented!()` `configuration_view` — and the bug being exercised
lives entirely in the overlay/base-view bookkeeping.

Closes AI-237

Release Notes:

- Fixed pressing "New Thread" (or Cmd-N) doing nothing when the Agent Panel's Settings view was open

### DIFF
--- a/crates/agent_ui/src/agent_panel.rs
+++ b/crates/agent_ui/src/agent_panel.rs
@@ -2094,7 +2094,15 @@ impl AgentPanel {
         let draft = self.ensure_draft(source, window, cx);
         if let BaseView::AgentThread { conversation_view } = &self.base_view {
             if conversation_view.entity_id() == draft.entity_id() {
-                if focus {
+                // If we're already viewing the draft as the base view but an
+                // overlay (e.g. Settings) is covering it, clear the overlay
+                // so the user actually sees the draft they asked for.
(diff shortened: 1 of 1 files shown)


cole-miller:
@zed-industries/approved

PR #56843, opened 2026-05-15
cloud_api_client: Add update_system_settings

tomhoule:
Wire up the client side of the new `PATCH /client/system_settings` endpoint added in zed-industries/cloud#2444, so we can persist the currently selected organization on a per-system basis. This PR only adds the request types and the client method; hooking it up to the actual organization switcher in the editor will come next.

The endpoint requires the `x-zed-system-id` header (the server returns 400 without it), so the method does not take an Option like the other client calls.

Release Notes:

- N/A


tomhoule:
Re-drafting this, I see there's been some changes to CloudApiClient I need to read up on to make sure this is still right.

maxdeviant (approved):

PR #56859, opened 2026-05-15
debugger_ui: Enable search in debugger terminal panel

isushank:
## Summary

pressing `cmd-f` (or using **Edit → Find**) while focused in the debugger terminal panel had no effect. The Edit menu item was not greyed out but nothing happened.

### Root cause

Three things were missing:

1. **No `BufferSearchBar` on toolbar** — debugger panes never registered a search bar, so there was no UI to deploy.
2. **`SubView::as_searchable` returned `None`** — the pane item wrapping the terminal had no searchable delegate, so even with a search bar present it could not find a target.
3. **No search action handlers on the render div** — `RunningState::render` did not wrap its output with `wrap_div_with_search_actions`, so `buffer_search::Deploy` was never caught.

### Changes

- **`new_debugger_pane`**: calls `PaneSearchBarCallbacks::setup_search_bar` to register `BufferSearchBar` on each debugger pane's toolbar (same pattern as `terminal_view::new_terminal_pane`).
- **`SubView::as_searchable`**: for `DebuggerPaneItem::Terminal`, delegates to the inner `TerminalView` which already implements `SearchableItem` (regex search, highlight matches).
- **`RunningState::render`**: wraps the outer div with `wrap_div_with_search_actions(active_pane)` so `buffer_search::Deploy` reaches the pane's toolbar.
- **`refresh_terminal_toolbars`**: re-fires `toolbar.set_active_item` when `DebugTerminal.terminal` is set, covering the edge case where the Terminal tab was already active before the terminal entity existed. Only refreshes panes where Terminal is the current active item (not background tabs).
- **`debugger_ui/Cargo.toml`**: adds `search` dependency.
- **Test**: `test_debugger_terminal_pane_has_search_bar` verifies the active pane toolbar has a `BufferSearchBar` after a debug session starts.

Release Notes:

- Fixed `cmd-f` search not working in the debugger terminal panel.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4768,6 +4768,7 @@ dependencies = [
  "project",
  "rpc",
  "schemars 1.0.4",
+ "search",
  "serde",
  "serde_json",
  "serde_json_lenient",
--- a/crates/debugger_ui/Cargo.toml
+++ b/crates/debugger_ui/Cargo.toml
@@ -57,6 +57,7 @@ pretty_assertions.workspace = true
 project.workspace = true
 rpc.workspace = true
 schemars.workspace = true
+search.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 serde_json_lenient.workspace = true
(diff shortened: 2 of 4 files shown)


isushank:
Closing — resubmitting from correct account.

PR #56861, opened 2026-05-15
editor: Fix hit testing when the editor is clipped

timvermeulen:
Fixes editor hit testing when the editor is clipped, most easily noticeable in a partially off-screen markdown block in the agent panel.

`PositionMap::point_for_position` was indexing into `self.line_layouts` relative to the editor's scroll position (which in this example doesn't apply because these editors don't scroll), rather than relative to `self.visible_row_range.start`. This caused the reported column to not always be computed using the correct row, which especially impacted rows closer to the bottom of the editor. Notice how closer to the bottom, `point_for_position` reports the correct row, but reports the column as 0:

https://github.com/user-attachments/assets/2a5b2c8b-2994-4983-a33b-ede209ba00e6

Self-Review Checklist:

Release Notes:

- Fixes selection not always matching the mouse position in a markdown block in the agent panel

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -12193,9 +12193,9 @@ impl PositionMap {
         let x = position.x + (scroll_position.x as f32 * self.em_layout_width);
         let row = ((y / self.line_height) as f64 + scroll_position.y) as u32;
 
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Thank you so much! This has bothered me in forever and appreciate this finally being fixed!

PR #56863, opened 2026-05-15
debugger_ui: Enable search in debugger terminal panel

### LINKED ISSUE #38409: Debugger: Can't search in debug output terminal
### Summary

When debugging a C++ application using lldb, I don't seem to be able to search in the terminal panel that contains the program output.

### Description
Steps to trigger the problem:
1. Start debugging a C++ application that dumps text to STDOUT.
2. In the debugger terminal panel, attempt to search using the Edit menu or CMD-F 
3. No search input appears.

**Expected Behavior**:
To be able to search in the terminal output. This is useful/essential to inspect log output etc.

**Actual Behavior**:
Nothing happens when pressing CMD-F. The Edit menu is not greyed out, but it doesn't do anything.

Zed: v0.203.5 (Zed) 
OS: macOS 26.0.0
Memory: 16 GiB
Architecture: aarch64

iamsushank:
When a debug session is running, pressing cmd-f (or using Edit → Find) in the terminal panel had no effect. Three things were missing:

- The debugger panes never had a BufferSearchBar added to their toolbar, so there was no search UI to deploy.
- SubView::as_searchable returned None for all item kinds, so even if a search bar were present it could not find a search target.
- RunningState::render did not wrap its output with the pane search action handlers, so buffer_search::Deploy was never handled.

Fix all three:
- Register BufferSearchBar on each debugger pane toolbar via the PaneSearchBarCallbacks global (same pattern as terminal_panel).
- Implement SubView::as_searchable for Terminal kind, delegating to the inner TerminalView which already implements SearchableItem.
- Wrap RunningState::render with wrap_div_with_search_actions so the Deploy action reaches the active pane's toolbar.
- Add refresh_terminal_toolbars to re-fire toolbar.set_active_item when the terminal entity is set, covering the case where the Terminal tab was already active before the terminal was created. Only refreshes panes where the Terminal subview is currently the active item.

Add a test that verifies the active debugger pane toolbar has a BufferSearchBar after a session starts.

Self-Review Checklist:

Fixes #38409

Release Notes:

- Fixed `cmd-f` search not working in the debugger terminal panel.


joneepenk:
why this change not merged for so long?
currently i copy content in debugger terminal to a editor as workaround.
it is trivial and ineffective and have other side effet.
do you guys already delegated all code debugging tasks to claude?

PR #56865, opened 2026-05-15
Fix activity indicator menu opening when actions are running

danilo-leal:
The fix comes based on that if the button is already wrapped in the popover menu, it doesn't need to have a click handler for the menu to open!

Release Notes:

- FIxed bug where the activity indicator in the status bar wouldn't open a menu with menu items to cancel running actions.


MrSubidubi (approved):
Nice!

PR #56866, opened 2026-05-15
agent: Do not decode images during render

bennetbo:
Turns out we were creating an ImageDecoder on every frame (added in #46167) when a tool returned an image as output, because we were trying to get its dimensions. That is now cached on `ContentBlock::Image`.

Release Notes:

- N/A


crates/agent_ui/src/conversation_view/thread_view.rs
```diff
@@ -7778,30 +7775,26 @@ impl ThreadView {
...
-                gpui::ImageFormat::Jpeg => "JPEG",
-                gpui::ImageFormat::Webp => "WebP",
-                gpui::ImageFormat::Gif => "GIF",
-                gpui::ImageFormat::Svg => "SVG",
-                gpui::ImageFormat::Bmp => "BMP",
-                gpui::ImageFormat::Tiff => "TIFF",
-                gpui::ImageFormat::Ico => "ICO",
-                gpui::ImageFormat::Pnm => "PNM",
-            };
-            let dimensions = image::ImageReader::new(std::io::Cursor::new(image.bytes()))
```
bennetbo, line 7798: This is the place where we were creating the decoder during rendering

benbrandt (approved):

PR #56879, opened 2026-05-15
agent_ui: Adjust settings UI for some providers

danilo-leal:
Namely, makes the ChatGPT subscription settings item a bit tidier, as well as more consistent with the Copilot one.

Release Notes:

- N/A


bennetbo:
@zed-industries/approved

PR #56890, opened 2026-05-15
git: Kill git blame process on task drop

benbrandt:
We replace these tasks often, so we should clean up after ourselves
here.

Release Notes:

- git: Fix git blame processes not getting dropped properly.


cole-miller:
@zed-industries/approved

PR #56915, opened 2026-05-15
repl: Improve native notebook editing and output rendering

MedSlimane:
## Summary

This adds the next native notebook slice in core Zed:

- preserve notebook outputs losslessly when opening and saving `.ipynb` files
- consolidate notebook controls into a top toolbar and improve per-cell run state
- render notebook markdown math plus richer execution outputs, including SVG, LaTeX, and table-friendly markdown output

## Motivation

The existing notebook path could already open notebooks and execute cells, but it had a few gaps that made normal notebook use risky or awkward:

- saving a notebook could drop rich outputs because rendered outputs were reconstructed instead of preserving the original nbformat payload
- execution controls were split across different surfaces, and running cells did not always reflect kernel lifecycle changes cleanly
- notebook markdown and kernel output did not render common math output correctly, and HTML-derived markdown tables were easy to collapse or clip

These issues make real notebooks less trustworthy to edit in Zed, especially when moving between Zed, JupyterLab, and VS Code.

## What Changed

- Store notebook code-cell outputs as both raw nbformat data and rendered UI output so save operations preserve MIME bundles, metadata, streams, errors, and unsupported output payloads.
- Keep notebook execution state tied to actual kernel requests and clear stale running indicators when kernels restart, change, or fail to launch.
- Replace the side/bottom notebook controls with a single top toolbar, and improve gutter run buttons with run, rerun, disabled, and running states.
- Add opt-in markdown math parsing/rendering for notebook callers without changing default markdown behavior elsewhere.
- Add native notebook/output support for SVG and `text/latex`, enable math rendering in markdown execution output, and keep wide markdown tables scrollable with min-content columns.
- Preserve unsupported MIME bundles on save while still showing a safe fallback in the UI.

## User Impact

Notebook users can save `.ipynb` files without losing rich outputs, see clearer kernel/cell execution state, and read common notebook math and table outputs directly in Zed without relying on a WebView renderer.

## Validation

- `cargo test -p markdown math`
- `cargo test -p repl test_rank_mime_type`
- `cargo test -p repl output_markdown`
- `cargo test -p repl latex_output`
- `cargo test -p repl html_table`
- `cargo test -p repl loose_math`
- `cargo test -p repl notebook`
- `cargo build -p repl`

## Suggested .rules additions

- N/A

Release Notes:

- Improved native notebook output preservation, controls, and math/table rendering


MedSlimane:
@cla-bot check

PR #56924, opened 2026-05-15
Replace Rules UI with Skills creation UI

MartinYe1234:
Replaces the legacy Rules creation UI with a focused "New Skill" creation window when the `SkillsFeatureFlag` is on.

The new `skills_library` crate provides a single-skill form with:
- Name, Description, and Body fields (placeholders double as labels)
- Live validation matching the SKILL.md spec (name: 1–64 lowercase + digits + hyphens, no leading/trailing hyphen; description: non-empty, ≤1024 chars)
- A scope dropdown listing every local worktree in the originating workspace plus "Global"
- An Optional Parameters card with the `disable-model-invocation` checkbox
- Save writes `<scope>/.agents/skills/<name>/SKILL.md` via `serde_yaml_ng` (YAML-safe escaping), refuses to overwrite existing skills, shows a success toast in the originating workspace, then closes the window

The "Rules" entry in the agent panel triple-dot menu is renamed to "Skills" when the flag is on. The existing `cmd-alt-l` keyboard shortcut still works: `AgentPanel::deploy_rules_library` reroutes to `deploy_skills_library` when the flag is on, so any persisted keymaps and the menu's automatic shortcut lookup keep functioning without changes to the default keymap files.

The old `rules_library` crate is intentionally left in place for this PR — once the flag rolls out and the few remaining `OpenRulesLibrary { prompt_to_select }` call sites (in thread_view link handlers and @-mention crease) are migrated to an "open existing skill" flow, a follow-up can delete the crate entirely.

15 unit tests cover the validation rules, frontmatter formatting (including YAML-special-character round-tripping), and disk write behavior (creates directory, refuses overwrite).

Closes AI-247

Release Notes:

- Added a new Skills creation UI that replaces the old Rules library when the Skills feature flag is enabled.


MartinYe1234:
We may want to show existing skills and allow editting/deletion of them via the UI as well

danilo-leal (approved):

PR #56926, opened 2026-05-15
agent_ui: Add mention disambiguation

danilo-leal:
Closes AI-261

This PR adds mention disambiguation in the agent panel, which works both for regular @-mentions as well as for skills. Effectively, when you mention files with the same name, the mention crease displays the next path parent name, following a similar approach to common tabs in the editor. For skills, the skill source is displayed (either global or from some project).

Screenshot 2026-05-15 at 6  32@2x

Release Notes:

- Agent: Improved file and skill mention disambiguation in the agent panel.


crates/acp_thread/src/mention.rs
```diff
@@ -342,13 +342,34 @@ impl MentionUri {
...
+
+    pub fn disambiguated_name(&self, needs_disambiguation: bool) -> String {
+        if !needs_disambiguation {
+            return self.name();
+        }
+
+        match self {
             MentionUri::Skill { name, source, .. } => {
                 if source.is_empty() {
                     format!("{} (global)", name)
```
rtfeldman, line 362: I'm still not a fan of `✨ code-review (global)` compared to like `✨ code-review ~/.agents/skills` just because the latter is so much clearer, but I don't feel super strongly about it. 😄

danilo-leal: I think now that we're showing the prefix only to disambiguate, we could go ahead and show the path; I was less bothered by it because clicking on the crease opens the file, which would make you learn the location

rtfeldman (approved):
Looks good! I opened https://github.com/zed-industries/zed/pull/57061 as an optional improvement, but this is fine to merge as-is. 👍

crates/acp_thread/src/mention.rs
rtfeldman: Ah that's a good point! Let's try it this way then. 👍

PR #56929, opened 2026-05-15
compliance: Loosen ignore list check

MrSubidubi:
Follow-up to https://github.com/zed-industries/zed/pull/56814 to fix the notice properly on the release branch. 

We can loosen this here since we only allow explicitly specified comments anyway.

Release Notes:

- N/A


probably-neb:
@zed-zippy approve

PR #56989, opened 2026-05-17
Refresh previews after file reloads

jiechen257:
## Summary
- refresh SVG previews when their source buffer reloads from disk
- emit a dedicated editor reload event and use it to refresh Markdown previews
- add regression coverage for editor reload events and Markdown preview refreshes

## Root Cause
Markdown and SVG previews already refreshed after in-editor edits, but external file reloads followed different event paths. SVG preview ignored `BufferEvent::Reloaded`, and Markdown preview had no editor-level content event for `multi_buffer::Event::Reloaded`.

## Tests
- `cargo test -p editor test_edit_events --lib`
- `cargo test -p markdown_preview refreshes_preview_when_source_editor_buffer_reloads --lib`
- `cargo fmt --check`
- `cargo check -p svg_preview --all-targets`


smitbarmase:
Hey, @tomhoule and I tried reproducing this on the latest main, but we couldn't really figure out a way to do it. This either seems fixed, or we haven't really understood what this PR is trying to fix. In both cases, we'd have to close this PR. If you think this is still the issue, please let us know with a reproducible example by creating a new issue and tag me here so I'm notified.

PR #56993, opened 2026-05-17
project_panel: Reduce cloning when rendering entries

cppcoffee:
Share marked selections across rendered project panel entries and defer drag-related path cloning until drag and drop rendering is enabled. Also defer symlink path string conversion until tooltip creation.

Release Notes:


smitbarmase (approved):
Thanks!

PR #56998, opened 2026-05-17
Improve robustness of lib finding in linux bundle script

fu5ha:
Parses the output of `ldd` in a slightly more robust way using `awk` to give a better error message if someone doesn't have one or more of the required shared libraries on their system.

Self-Review Checklist:

- [x] Unsafe blocks (if any) have justifying comments (N/A)
- [x] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) (N/A)
- [x] Tests cover the new/changed behavior (N/A)
- [x] Performance impact has been considered and is acceptable (N/A)

Release Notes:

- N/A

### DIFF
--- a/script/bundle-linux
+++ b/script/bundle-linux
@@ -151,15 +151,39 @@ cp "${target_dir}/${target_triple}/release/zed" "${zed_dir}/libexec/zed-editor"
 cp "${target_dir}/${target_triple}/release/cli" "${zed_dir}/bin/zed"
 
 # Libs
+# parse ldd output, in case the user failed to run script/linux or someone forgets
+# to add a dep there
 find_libs() {
-    ldd ${target_dir}/${target_triple}/release/zed |\
-        cut -d' ' -f3 |\
+    ldd "${target_dir}/${target_triple}/release/zed" |
+        awk '
+            /=> not found/ {
+                print "Missing shared library: " $1 > "/dev/stderr"
+                missing = 1
+                next
+            }
+
+            /=>/ && $3 ~ /^\// {
+                print $3
+                next
+            }
+
+            $1 ~ /^\// {
+                print $1
+                next
+            }
+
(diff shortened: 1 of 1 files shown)


script/bundle-linux
```diff
@@ -151,17 +151,39 @@ cp "${target_dir}/${target_triple}/release/zed" "${zed_dir}/libexec/zed-editor"
...
-        grep -v '\<\(libc.so\|libgcc_s.so\|libm.so\|libpthread.so\|libdl.so\|libasound.so\)'
+    ldd "${target_dir}/${target_triple}/release/zed" |
+        awk '
+            /=> not found/ {
+                print "Missing shared library: " $1 > "/dev/stderr"
+                missing = 1
+                next
+            }
+
+            /=>/ && $3 ~ /^\// {
```
SomeoneToIgnore, line 165: With a relative `LD_LIBRARY_PATH`, glibc can report a resolved dependency as:

```text
libcustom.so.1 => ./local/libcustom.so.1 (0x...)
```

This follows glibc’s [path construction](https://github.com/bminor/glibc/blob/glibc-2.42/elf/dl-load.c#L1771-L1793) and [trace output](https://github.com/bminor/glibc/blob/glibc-2.42/elf/rtld.c#L2058-L2072).
Neither new absolute-path check accepts this entry, so the library disappears from the copy list.
If another absolute dependency remains, the pipeline still succeeds, allowing an incomplete bundle.
The old parser preserves this dependency.

SomeoneToIgnore (commented):
Hey, thank you for the fix — if I've understood it right, we still have one thing to fix on top?

SomeoneToIgnore:
Will close as stale as no activity had happened for a while, and first conflicts started to appear.

PR #57019, opened 2026-05-17
git: Extract `canvas` and `graph` out of `git_graph.rs`

GoldStrikeArch:
cc @Anthony-Eid 

Self-Review Checklist:

Release Notes:

- N/A


Anthony-Eid:
Maybe I'm a lunatic but I kinda like this all being one massive file. 

If we do want to push this change forward I would prefer we do so after we move `git_graph.rs` into the `git_ui` crate

GoldStrikeArch:
@Anthony-Eid 

> Maybe I'm a lunatic but I kinda like this all being one massive file.
> I would prefer we do so after we move `git_graph.rs` into the `git_ui` crate

Okay, I will close this PR for now then. Overall I did 3 things:
- splitted that big `git_graph.rs` into smaller pieces (this PR was just a first step)
- refactored some code a bit (mostly around `add_commits` function and I also wanted it to be a separate PR)
- added a feature so we always use main/master branch as the most left branch (at index 0), because right now it is kinda confusing as your main branch can be second/third depends on the graph history.

Do you need the refactoring or I can just open the PR with the feature? IMO it is good one and usually git graphs are working this way (main branch is the first one). I also wanted to add more colors (as a separate PR) as right now there are too few of them IMO

Anthony-Eid:
How does the main/master branch always being to the left feature work? The graph right now matches the behavior of using the git log --graph cli so I would prefer we pass an argument to git for that functionality

Anthony-Eid:
Also, I changed my mind, we should probably go ahead with an extraction like this just so it's easier to onboard new people to the codebase. 

Could we do this after the git graph mod is moved to git ui please?

GoldStrikeArch:
@Anthony-Eid 

> How does the main/master branch always being to the left feature work?

We already have the info about the default origin branch (from `Repository::default_branch`), I did the fallback to "main"/"master" in case we don't have default branch (but should not really be a popular case), so basically git graph was fetching that info and then using it here 

> Could we do this after the git graph mod is moved to git ui please?

Yes. I can move it in a separate PR and close this one and then after merging open a third PR with just the feature (no refactoring), wdyt? Or do you prefer extraction first, then moving this into `git ui` and then adding the feature? 

You can see the whole thing here -> https://github.com/zed-industries/zed/compare/main...GoldStrikeArch:zed:feat/refactor-graph (extraction + feature)

GoldStrikeArch:
@Anthony-Eid Hi, any update on this one?

PR #57025, opened 2026-05-18
git_ui: Count commit title length by characters instead of bytes

koxya:
# What

- Replace `title.len()` with `title.chars().count()` in the commit title length check, so the limit is measured in Unicode characters instead of UTF-8 bytes.
- Applied in two places that share the same logic:
  - `crates/git_ui/src/git_panel.rs` — Git panel's inline warning
  - `crates/git_ui/src/commit_modal.rs` — commit modal's inline warning

# Why

The commit title length check used str::len(), which returns UTF-8 byte length rather than character count. As a result, titles containing multi-byte characters (Japanese, Chinese, emoji, etc.) triggered the warning far below the configured commit_title_max_length — around 24 characters instead of the default 72.

Self-Review Checklist:

  - Before

    https://github.com/user-attachments/assets/a895530c-2f73-470c-97fa-29d9467c14e1

  - After

    https://github.com/user-attachments/assets/ffbe1ba2-0ccc-4b02-87f5-836da7841dd9

Closes #ISSUE

Release Notes:

- Fixed commit title length check miscounting multi-byte characters as multiple characters.

### DIFF
--- a/crates/git_ui/src/commit_modal.rs
+++ b/crates/git_ui/src/commit_modal.rs
@@ -1,5 +1,7 @@
 use crate::branch_picker::{self, BranchList};
-use crate::git_panel::{GitPanel, commit_message_editor, panel_editor_style};
+use crate::git_panel::{
+    GitPanel, commit_message_editor, commit_title_exceeds_limit, panel_editor_style,
+};
 use crate::git_panel_settings::GitPanelSettings;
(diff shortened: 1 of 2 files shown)


MrSubidubi (approved):
NIce find and fix, even better given that you added a test for this. Thank you and congratulations to your first contribution! 🎉

PR #57032, opened 2026-05-18
git_graph: Show full commit message in tooltip on hover

### LINKED ISSUE #56903: Cannot view full title of commit since it breaks on newlines
### Reproduction steps

Just name commit in a few lines and try check it in Git Graph.

Image

Image

But this full name

Image

Zed 1.2.6 
Commit: 20b7f31e7dbe8233a198728ccf3c8aa1180c13e4
Version: 1.2.6+stable.280.20b7f31e7dbe8233a198728ccf3c8aa1180c13e4

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json

```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

Ubuntu

aaronang:
Self-Review Checklist:

Closes #56903 

Release Notes:

- Improved commit tooltip in Git Graph.

## Screenshots

Showing full commit message:

CleanShot 2026-05-17 at 21 59 35@2x

Truncating messages with more than 800 characters:

CleanShot 2026-05-17 at 22 12 07

## Alternative

One thing I was considering is whether we should be using a `HoverPopover` instead so that the preview is wider and we don't have to truncate the commit message body. I would love to hear your thoughts on this. For now, I stuck with the `Tooltip` for simplicity reasons.

CleanShot 2026-05-17 at 22 04 39@2x

### DIFF
--- a/crates/git_graph/src/git_graph.rs
+++ b/crates/git_graph/src/git_graph.rs
@@ -1538,13 +1538,16 @@ impl GitGraph {
                 let mut formatted_time = String::new();
                 let subject: SharedString;
                 let author_name: SharedString;
+                let message: SharedString;
 
-                if let CommitDataState::Loaded(data) = data {
+                if let CommitDataState::Loaded(data) = &data {
                     subject = data.subject.clone();
+                    message = data.message.clone();
                     author_name = data.author_name.clone();
                     formatted_time = format_timestamp(data.commit_timestamp);
                 } else {
                     subject = "Loading…".into();
+                    message = "Loading…".into();
                     author_name = "".into();
                 }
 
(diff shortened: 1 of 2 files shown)


aaronang:
@cla-bot check

smitbarmase (changes requested):
Thanks, I think a tooltip is just not the right component for this. I agree with your idea for our popover.

You can check how we handle Git inline blame in the editor:

image

Can we reuse it?

aaronang:
@smitbarmase I updated the tooltip to use `CommitTooltip`.

CleanShot 2026-05-18 at 08 26 48

I also updated the styling to match the git blame popover. This removes some funky padding in the commit message. This also means that the commit tooltip renders cleaner in the git panel:

CleanShot 2026-05-18 at 08 30 17

## `CommitTooltip` without styling changes

This is what `CommitTooltip` looked like before the changes. Note that the commit message is clipped at the top and bottom due to some extra padding.

CleanShot 2026-05-18 at 07 42 21

## Open questions

1. One other possible improvement is to have the git blame popover user `CommitTooltip` too but perhaps we can work on that in a different pull request?
2. I didn't write any tests for these changes. Your guidance here would be appreciated.
3. Playing around with this rich commit tooltip, I wonder if we should have a tooltip at all. What if we show the full commit message in the sidebar instead? I have a feeling the tooltip can get quite annoying. It's definitely staying longer than necessary. For example, when you hover over a commit and scroll without moving the mouse, it shows longer than expected before disappearing. At least that's the default behavior of the current `CommitTooltip`.

CleanShot 2026-05-18 at 09 54 50

smitbarmase:
This looks great to me. I'm going to test drive it and see if we can ship it. I do think it would be nice to have more details in the right panel, but for now, the hover popup shouldn't hurt.

smitbarmase (approved):
Thanks! Pushed one small thing on top, we were building the `CommitDetails` and calling `ParsedCommitMessage::parse` upfront for every commit row in the graph during render, that's not good.

I moved that work inside the `hoverable_tooltip` closure so we only pay for it on hover. Thanks again!

aaronang:
@smitbarmase Thanks for catching that!

PR #57034, opened 2026-05-18
editor: Remove redundant allocation when merging highlight ranges

cppcoffee:
Reduce one Vec clone + collect allocation per frame

Release Notes:

### DIFF
--- a/crates/editor/src/element.rs
+++ b/crates/editor/src/element.rs
@@ -10410,20 +10410,13 @@ impl Element for EditorElement {
                         cx,
                     );
 
-                    let merged_highlighted_ranges =
-                        if let Some((_, colors)) = document_colors.as_ref() {
-                            &highlighted_ranges
-                                .clone()
-                                .into_iter()
-                                .chain(colors.clone())
-                                .collect()
-                        } else {
-                            &highlighted_ranges
-                        };
+                    if let Some((_, colors)) = document_colors.as_ref() {
+                        highlighted_ranges.extend(colors.iter().cloned());
+                    }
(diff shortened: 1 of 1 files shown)


smitbarmase (changes requested):
Thanks! 

I think this introduces a rendering regression. Before, we were just combining this for `bg_segments_per_row`, but now we are actually mutating `highlighted_ranges`, which means we'll paint the highlight ranges as well as the document colors that already had a separate paint path. So document color ranges end up getting drawn twice?

Can we keep the merge local to `bg_segments_per_row`? e.g. change it to take `impl IntoIterator` and chain at the call site, so we avoid the clone + collect without mutating `highlighted_ranges`.

cppcoffee:
before:
```
let merged_highlighted_ranges =
    if let Some((_, colors)) = document_colors.as_ref() {
        &highlighted_ranges
            .clone()          // allocation 1: clone highlighted_ranges
            .into_iter()
            .chain(colors.clone())  // allocation 2: clone colors
            .collect()        // allocation 3: collect into new Vec
    } else {
        &highlighted_ranges
    };
```

after:


`iter().cloned()` is lazy, `chain()` and `flat_map()` only combine iterators, and `bg_segments_per_row` directly consumes this combined iterator in the for loop, with no intermediate Vec.

smitbarmase (approved):
Thanks!

smitbarmase:
Can you run `cargo fmt --all -- --check`?

cppcoffee:
Sorry, the formatting has been completed.

PR #57051, opened 2026-05-18
acp: Support additional session directories

benbrandt:
Still behind a feature flag for now for testing with various agents.

Self-Review Checklist:

Release Notes:

- N/A


bennetbo (commented):
Looks good, one thing though is that `work_dirs_from_session_info` will now return additional directories even without the feature flag present. Wondering if that causes issues.

bennetbo (approved):

PR #57054, opened 2026-05-18
editor: Fix code lens showing above folded regions

### LINKED ISSUE #57053: Code Lens does not respect folding
### Reproduction steps

1. Start Zed, ideally with both the JS/TS LSP enabled (bundled with Zed and auto-downloaded) and the C# LSP enabled (bundled with the C# Zed extension and auto-downloaded)
2. Make sure Code Lens is enabled
3. Open the below files and wait for code lens results to appear
4. In the TS file, fold the `Functions` region on line 32 and observe the 5 code lenses above it
5. In the CS file, fold the `Numbers` region on line 9 and observe the 15 code lenses above it

<details>
  <summary>TypeScript example file</summary>
  
```ts
// #region Numbers
// #region Ones
const ONE   = 1;
const TWO   = 2;
const THREE = 3;
const FOUR  = 4;
const FIVE  = 5;
// #endregion Ones

// #region Tens
// #region TensA
const TEN    = 10;
const TWENTY = 20;
const THIRTY = 30;
// #endregion TensA

// #region TensB
const FORTY = 40;
const FIFTY = 50;
const SIXTY = 60;
// #endregion TensB
// #endregion Tens

// #region Hundreds
const HUNDRED       = 100;
const TWO_HUNDRED   = 200;
const THREE_HUNDRED = 300;
const FOUR_HUNDRED  = 400;
// #endregion Hundreds
// #endregion Numbers

// #region Functions

function addOnes(): number {
    return ONE + TWO + THREE;
}

function addTens(): number {
    return TEN
         + TWENTY
         + THIRTY
         + FORTY
         + FIFTY
         + SIXTY;
}

function addHundreds(): number {
    return HUNDRED + TWO_HUNDRED + THREE_HUNDRED + FOUR_HUNDRED;
}

function sumAll(): number {
    return addOnes()
         + addTens()
         + addHundreds()
         + ONE
         + TEN
         + HUNDRED;
}

function multiplyMixed(): number {
    return (ONE + TWO) * (TEN + TWENTY) * (HUNDRED + TWO_HUNDRED);
}

// #endregion Functions


// Use everything so code_lens shows high counts
function useThings(): number {
  return addOnes()
    + addTens()
    + addHundreds()
    + sumAll()
    + multiplyMixed();
}

console.log(`addOnes:       ${addOnes()}`);
console.log(`addTens:       ${addTens()}`);
console.log(`sumAll:        ${sumAll()}`);
console.log(`useThings:     ${useThings()}`);
```
</details>

<details>
  <summary>C# example file</summary>
  
```cs
#!/usr/bin/env dotnet
#:sdk Microsoft.NET.Sdk
#:property TargetFramework=net10.0
#:property Nullable=enable
#:property ImplicitUsings=enable

static class Calc
{
    #region Numbers
    #region SmallNumbers
    public const int One   = 1;
    public const int Two   = 2;
    public const int Three = 3;
    public const int Four  = 4;
    public const int Five  = 5;
    #endregion SmallNumbers

    #region BigNumbers
    #region BigNumbersOne
    public const int Ten     = 10;
    public const int Twenty  = 20;
    public const int Thirty  = 30;
    #endregion BigNumbersOne

    #region BigNumbersTwo
    public const int Forty   = 40;
    public const int Fifty   = 50;
    public const int Sixty   = 60;
    #endregion BigNumbersTwo
    #endregion BigNumbers

    #region SuperBigNumbers
    public const int Hundred      = 100;
    public const int TwoHundred   = 200;
    public const int ThreeHundred = 300;
    public const int FourHundred  = 400;
    #endregion SuperBigNumbers
    #endregion Numbers


    #region Functions

    public static int AddOnes() => One + Two + Three;

    public static int AddTens()
    {
        return Ten
             + Twenty
             + Thirty
             + Forty
             + Fifty
             + Sixty;
    }

    public static int AddHundreds() => Hundred + TwoHundred + ThreeHundred + FourHundred;

    public static int SumAll()
    {
        return AddOnes()
             + AddTens()
             + AddHundreds()
             + One
             + Ten
             + Hundred;
    }

    public static int MultiplyMixed()
    {
        return (One + Two) * (Ten + Twenty) * (Hundred + TwoHundred);
    }

    #endregion Functions

    // Use everything so code_lens shows high counts
    public static int UseThings()
    {
        return AddOnes()
             + AddTens()
             + AddHundreds()
             + SumAll()
             + MultiplyMixed();
    }

    public static void Main()
    {
        Console.WriteLine($"AddOnes:       {AddOnes()}");
        Console.WriteLine($"AddTens:       {AddTens()}");
        Console.WriteLine($"SumAll:        {SumAll()}");
        Console.WriteLine($"UseThings:     {UseThings()}");
    }
}
```
</details>

**Current behavior**: if some code is folded, all the Code Lens data is shown above the fold. For large folds this is a *huge* list.
**Expected behavior**: if some code is folded, all Code Lens data for that fold is hidden.

For the TS example (see line 32):
Screenshot showing the 5 code lenses that should not be shown for the example TS code

For the CS example (see line 9):
Screenshot showing the 15 code lenses that should not be shown for the example dotnet code

Example from some code of mine that actually triggered this whole thing:
Screenshot showing 20+ Code Lens lines before a folded Stucts

**Zed**: Zed Preview 1.3.3 (1.3.3+preview.279.98131bbb4289a73d2f63b0175a65ee72c27493a9)
**OS**: macOS 26.5
**Memory**: 64GB
**Architecture**: amd64

### Attach Zed log file

```log

```

</details>

### Relevant Zed settings

```json
{
    "code_lens": "on",
}
```

</details>

### Relevant Keymap

```json

```

</details>

_No response_

None

Vlaaaaaaad:
Closes https://github.com/zed-industries/zed/issues/57053
> [!WARNING]
> This code was written by LLMs, under the supervision of a so-called developer that never really wrote Rust profesionally and that spends more time in Pages&Keynote than in an IDE.

Code lens blocks were displayed above folded regions because their `BlockPlacement::Above` anchors all mapped to the same fold placeholder row, stacking every lens for lines inside the fold into a single list above the fold marker. That led to chaos and a ton of code lens results above large folded regions, as shown on https://github.com/zed-industries/zed/issues/57053.

This LLM-generated but human-validated (keep in mind: **dumb human**) PR fixes the issue while not ruining performance by implementing several changes:
- Filter folded rows in `apply_lens_actions_for_buffer` so code lens blocks are never inserted for lines that are currently folded.
- When folds are toggled, re-evaluate cached lens actions (zero-clone via `std::mem::take`) to insert/remove blocks without LSP round-trips.
- Refactor `fold_buffer`/`unfold_buffer` into `fold_buffers`/`unfold_buffers` to avoid redundant per-buffer work and ensure lens reapplication happens once.
- Deduplicate in-flight resolve requests using a `resolving` HashSet to prevent redundant LSP round-trips for the same lens.

<details>
  <summary>TypeScript before and after screenshots (note line 32)</summary>
Before
After
</details>

<details>
  <summary>C# before and after screenshots (note line 9)</summary>
Before
After
</details>

**Beware**: I have very little Rust experience, little Zed experience, and absolutely no experience with Zed's editor code or Zed's Code Lens implementation! The code looks correct to me and I did a bunch of validation but, again, I have no idea what I am doing. **I will not feel at all offended if y'all reject this PR** — my only hope is that it helps y'all in finding the correct fix for https://github.com/zed-industries/zed/issues/57053!

----

Self-Review Checklist:

Release Notes:
- Editor: Folded regions now correctly hide code lens results


crates/editor/src/code_lens.rs
```diff
@@ -292,10 +321,13 @@ impl Editor {
             }
         }
 
-        let mut new_lines_by_row = group_lenses_by_row(all_lenses, snapshot)
-            .map(|line| (MultiBufferRow(line.position.to_point(snapshot).row), line))
+        let mut new_lines_by_row = group_lenses_by_row(all_lenses, &snapshot)
+            .map(|line| (MultiBufferRow(line.position.to_point(&snapshot).row), line))
             .collect::<HashMap<_, _>>();
 
+        let display_snapshot = self.display_map.update(cx, |map, cx| map.snapshot(cx));
```
SomeoneToIgnore: MultiBufferSnapshot is a part of the DisplaySnapshot, so let's move this line up instead and use the `.buffer()` instead of creating 2 snapshots separately.

crates/editor/src/fold.rs
```diff
@@ -597,6 +589,7 @@ impl Editor {
         self.scrollbar_marker_state.dirty = true;
         self.update_data_on_scroll(false, window, cx);
         self.folds_did_change(cx);
+        self.reapply_code_lenses_for_fold_change(cx);
```
SomeoneToIgnore: This all is odd, as we do have `update_data_on_scroll` right above and that contains a call to the code lens update.

What is different in the new method so we need that and why cannot the `update_data_on_scroll` be used instead?

SomeoneToIgnore: All other `reapply_code_lenses_for_fold_change` usages are equally questionable: we need to have a single point of truth for that, as not only the code lens may have issues after [un]folds.

crates/editor/src/code_lens.rs
```diff
@@ -50,6 +54,7 @@ impl Default for CodeLensState {
             blocks: HashMap::default(),
             actions: HashMap::default(),
             resolve_task: Task::ready(()),
+            resolving: HashSet::default(),
```
SomeoneToIgnore: Why is the code that tries to deal with "update more on fake scroll/fold" scenario touches the way we resolve things?

And why, if it's so needed, it's not unit tested?

SomeoneToIgnore (changes requested):
Thank you, nice catch, but the impl is somewhat odd: seems though, we're not too far away from having it normal?

SomeoneToIgnore:
Also, taking a step back, I wonder whether this all is needed?
Blocks are top-level transforms and folds are very close to the top of the chain: https://github.com/zed-industries/zed/blob/eb2223c0802c4e864caac1687631ac0c45af4dee/crates/editor/src/display_map.rs#L381-L388

Which means that normally, blocks should not appear on the folded regions?
I wonder if there's a special block kind we could use (or add and add more logic there) instead of all this data re-fetching.

Vlaaaaaaad:
> Also, taking a step back, I wonder whether this all is needed?
> Blocks are top-level transforms and folds are very close to the top of the chain: [...]
> 
> Which means that normally, blocks should not appear on the folded regions?
> **I wonder if there's a special block kind we could use (or add and add more logic there) instead of all this data re-fetching.**

Hm... there might be? As I initially said, this is not a part of the codebase I am familiar with and I didn't have the bandwidth to do a deep-dive and understand it properly. I'll try to do that next week and address all the review comments (**thank you** for the super-kind review❣️) but no promises because I have a chaotic week coming up!

For **context**: my first approach was to filter out code lenses (which are stored as `Block::Custom` with `style: BlockStyle::Spacer`) and just not render them if there's a fold by adding an `|| block.style() == BlockStyle::Spacer` in [this condition](https://github.com/zed-industries/zed/blob/eb2223c0802c4e864caac1687631ac0c45af4dee/crates/editor/src/element.rs#L3972-L3974) but that did not work because the block's height was apparently reserved in the display map and that resulted in a bunch of blank space that I could not figure out how to eliminate.
After that I went to the the more complex approach you see here which initially was less code that had horrible performance and that required a bunch of weird optimizations — expanding nested folds had visible ~2s lag on slower machines, for example (there's a reason I do some of my development on an Intel i9 MacBook!).

Vlaaaaaaad:
**FYI**: I had my schedule full, but I should be able to tackle this either this weekend or by EOD Wednesday next week. Apologies about the delay!

Vlaaaaaaad:
I, annoyingly, did not manage to find the time needed to do the deep-dive I wanted 😞 Fortunately, I did have enough time to investigate _a tiny bit_ and test stuff so I changed the implementation based on @SomeoneToIgnore's advice and implemented filtering at the `BlockMap::sync()` level. I confirmed this works well and there's no performance impact. 

I did look at other alternatives (like adding a `should_be_hidden_by_folds` property to blocks, adding a whole new tracking/origin system for blocks, and so on) but they all seemed way more complex for little benefit. Still, that's coming from somebody that still doesn't know this part of the code well enough so 🤷

crates/editor/src/code_lens.rs
Vlaaaaaaad: This whole code was removed, so marking this as resolved

Vlaaaaaaad:
Closing this one due to a lack of bandwidth on my part 😞 @SomeoneToIgnore I apologize for wasting your time with this!

----

It seems that my _"you know what, no, I should filter all blocks out because folds should hide everything"_ logic was incorrect and/or incomplete. 
**Allegedly**, the `diagnostics_tests::test_diagnostics_with_folds` test is failing because, while Diagnostic blocks are already hidden when folded (renderer skips them using `is_line_folded` at `element.rs:3128`), removing them from the block map is ruining the transform chain somehow(?). This could be fixed by adding back the restriction removed by the ["_All blocks inside folds should be hidden_" commit](https://github.com/zed-industries/zed/pull/57054/commits/4c145663954401798641f8704451ce764a53a1c8) or by adding an alternative `&& matches!(block.placement, BlockPlacement::Above(_))` restriction buuuut I can't do that without understanding this whole thing properly and, frustratingly, I do not have the necessary time to allocate to this right now (and it doesn't look like I'll have enough time in July either) so I am closing this PR. I am sorry!

PR #55215, opened 2026-04-29
Disable fuzzy-ruby-server by default in Ruby language servers

candidosales:
## Summary

Add `!fuzzy-ruby-server` to the Ruby language servers list in the default settings so it is opt-in rather than enabled for all users by default.

This is a prerequisite for merging the Fuzzy Ruby Server support in the Ruby extension: https://github.com/zed-extensions/ruby/pull/283

As requested by @vitallium in the review comment: https://github.com/zed-extensions/ruby/pull/283\#issuecomment-4294888519

## Change

```diff
- "language_servers": ["solargraph", "!ruby-lsp", "!rubocop", "!sorbet", "!steep", "!kanayago", "..."],
+ "language_servers": ["solargraph", "!ruby-lsp", "!rubocop", "!sorbet", "!steep", "!kanayago", "!fuzzy-ruby-server", "..."],
```

Release Notes:

- Disabled `fuzzy-ruby-server` by default for Ruby files.

PR #57055, opened 2026-05-18
Disable rubydex by default in Ruby language servers

candidosales:
## Summary

Add `!rubydex` to the Ruby language servers list in the default settings so it is opt-in rather than enabled by default for all users.

This is a prerequisite for working to integrate this new Ruby library for indexing: https://github.com/Shopify/rubydex

Release Notes:

- Disabled `rubydex` by default for Ruby files.

Reference: https://github.com/zed-industries/zed/pull/55215

### DIFF
--- a/assets/settings/default.json
+++ b/assets/settings/default.json
@@ -2271,6 +2271,7 @@
         "!steep",
         "!kanayago",
         "!fuzzy-ruby-server",
+        "!rubydex",
         "...",
       ],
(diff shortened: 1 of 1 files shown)


MrSubidubi:
Not opposed but can you link to a PR/content where that language server support is about to be added?

MrSubidubi:
Closing as stake

PR #57068, opened 2026-05-18
agent: Correct `single_file_review` default in docs and schema comment

cachebag:
The default was flipped to `false` in #48619, but the doc comment in `settings_content` and the agent-panel docs still implied it was on by default. Update both to match the shipped default and explain that the setting is opt-in because it temporarily overrides the buffers git diff while review is active.

Self-Review Checklist:

- [ ] Unsafe blocks (if any) have justifying comments - N/A
- [ ] The content is consistent with the [UI/UX checklist](https://github.com/zed-industries/zed/blob/main/CONTRIBUTING.md#uiux-checklist) - N/A
- [ ] Tests cover the new/changed behavior - N/A
- [ ] Performance impact has been considered and is acceptable - N/A

Closes #ISSUE: N/A

Release Notes:

- N/A

### DIFF
--- a/crates/settings_content/src/agent.rs
+++ b/crates/settings_content/src/agent.rs
@@ -151,7 +151,11 @@ pub struct AgentSettingsContent {
     pub play_sound_when_agent_done: Option<PlaySoundWhenAgentDone>,
     /// Whether to display agent edits in single-file editors in addition to the review multibuffer pane.
     ///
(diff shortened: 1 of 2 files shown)


bennetbo:
Hey, thanks for this. I can't push to your PR and wanted to tweak some of the wording, so I made another PR here #57375

PR #57078, opened 2026-05-18
client: Reconnect cloud websocket after it disconnects

tomhoule:
The cloud websocket was established once during sign-in and never re-established. On any server restart or transient network drop the connection task exited. yawc itself does not reconnect.

This wraps `connect_to_cloud` in a long-lived task that re-establishes the websocket with exponential backoff and jitter, reusing `INITIAL_RECONNECTION_DELAY` and `MAX_RECONNECTION_DELAY` so the behavior matches the Collab reconnect loop in the same module.

Part of CLO-713.

Release Notes:

- N/A

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -334,6 +334,7 @@ struct ClientState {
     credentials: Option<Credentials>,
     status: (watch::Sender<Status>, watch::Receiver<Status>),
     _reconnect_task: Option<Task<()>>,
+    _cloud_connection_task: Option<Task<()>>,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -435,6 +436,7 @@ impl Default for ClientState {
             credentials: None,
             status: watch::channel_with(Status::SignedOut),
             _reconnect_task: None,
+            _cloud_connection_task: None,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


tomhoule:
I aligned the retry implementation with the collab client retry implementation, but I'm wondering if we wouldn't benefit from something different. We don't care about reconnecting asap as much, but we do care about reconnecting after the user has been offline for a few hours, and I don't think this implementation works well in that case.

crates/client/src/client.rs
```diff
@@ -957,28 +961,56 @@ impl Client {
...
+                cx.background_executor().timer(delay + jitter).await;
+                delay = cmp::min(delay * 2, MAX_RECONNECTION_DELAY);
+            }
+        });
+        self.state.write()._cloud_connection_task = Some(task);
+    }
+
+    /// Runs a single attempt of the cloud websocket connection, returning once the connection
+    /// closes (cleanly or otherwise) or fails to establish.
+    async fn run_cloud_connection(this: &Arc<Self>, cx: &mut AsyncApp) -> Result<()> {
```
maxdeviant: Think we can still call this `self`?

```suggestion
    async fn run_cloud_connection(self: &Arc<Self>, cx: &mut AsyncApp) -> Result<()> {
```

tomhoule:
> I aligned the retry implementation with the collab client retry implementation, but I'm wondering if we wouldn't benefit from something different. We don't care about reconnecting asap as much, but we do care about reconnecting after the user has been offline for a few hours, and I don't think this implementation works well in that case.

I looked into this again and my concern was misplaced. We'll try reconnecting indefinitely.

maxdeviant (approved):

PR #57091, opened 2026-05-18
git_ui: Preserve subdirectory paths when switching worktrees

mattkranzler5:
## Summary

- Previously, creating or switching a git worktree from a workspace opened at a subdirectory of the repo would reopen the new workspace at the worktree root, losing the subdirectory context. Now we collect the project's visible paths that fall under any source git repo and remap them through the create/switch path mapping so the same subdirectory opens under the new worktree root.
- Reworks worktree trust propagation to enumerate the new workspace's visible worktrees directly instead of looking up paths in its `worktree_store`, so trust transfers reliably regardless of which paths were used to open the workspace.

## Test plan

- [ ] Open a workspace at a subdirectory of a repo (e.g. `repo/crates/foo`), create a new worktree, and verify the new workspace opens at the equivalent subdirectory under the worktree root.
- [ ] Switch between existing worktrees from a subdirectory-rooted workspace and confirm the subdirectory persists.
- [ ] Verify worktree trust transfers to the new workspace.

## Demo
https://drive.google.com/file/d/1XDkQhdAvwrz1lUnaIITLsDLVHATZ_rD-/view?usp=sharing

Release Notes:

- Fixed git worktree creation/switching from a workspace opened at a subdirectory so the new workspace opens at the equivalent subdirectory instead of jumping to the worktree root.

### DIFF
--- a/crates/git_ui/src/worktree_service.rs
+++ b/crates/git_ui/src/worktree_service.rs
@@ -1,6 +1,6 @@
 use std::error::Error;
 use std::fmt;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use anyhow::anyhow;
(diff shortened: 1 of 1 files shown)


mattkranzler5:
@cla-bot check

Issue #49776, opened 2026-02-21
Severe file descriptor leak: 450k+ open /proc/[PID]/stat files after 21 hours


iridite:
# Zed Memory Leak Bug Report

## Summary
Severe file descriptor leak causing excessive memory consumption after ~21 hours of runtime. Zed accumulates 450,000+ open file descriptors, primarily `/proc/[PID]/stat` files that are never closed.

## Environment
- **Zed Version**: 0.224.8-1.1
- **OS**: CachyOS Linux (Arch-based)
- **Kernel**: 6.19.2-2-cachyos
- **Desktop**: Niri (Wayland compositor)
- **Installation Method**: Package manager (pacman/yay)

## System Resources
- **Total RAM**: 13GB
- **Swap**: 45GB (12GB used)

## Problem Description

After running Zed for approximately 21 hours, the editor exhibits severe resource leaks:

### File Descriptor Leak
```
Total file descriptors: 451,967
- Regular files (REG): 414,039
- /proc/[PID]/stat files: 94,710+ (never closed)
- eventfd: 979
- sockets/pipes: 375
- Font files: 269
```

### Memory Consumption
```
Physical Memory (RSS): 1.69 GB (11.9% of system RAM)
Virtual Memory (VmSize): 93.7 GB
Data Segment (VmData): 5.8 GB
Threads: 55
Runtime: 21:18:42
```

### Process Details
```bash
$ ps -p 667633 -o pid,ppid,cmd,%mem,%cpu,etime,rss,vsz
PID    PPID CMD                         %MEM %CPU     ELAPSED   RSS    VSZ
667633  905 /usr/lib/zed/zed-editor    11.9 17.6    21:18:42 1682596 95952480

$ lsof -p 667633 2>/dev/null | wc -l
451967

$ lsof -p 667633 2>/dev/null | grep "/proc/" | wc -l
94710
```

## Root Cause Analysis

Zed is repeatedly opening `/proc/[PID]/stat` files for system process monitoring but **never closing them**:

```bash
$ lsof -p 667633 2>/dev/null | grep "/proc/" | awk '{print $NF}' | sed 's|/proc/[0-9]*/|/proc/PID/|g' | sort | uniq -c | sort -rn | head -5
94710 /proc/PID/stat
  193 /proc/PID/task/993/stat
  193 /proc/PID/task/992/stat
  193 /proc/PID/task/982/stat
  193 /proc/PID/task/981/stat
```

[...]

Issue #20806, opened 2024-11-18
Zed fails with `Too many open files (os error 24)`


alexkirsz:
# Describe the bug / provide steps to reproduce it

When opening new files, Zed sometimes fails with the error `Too many open files (os error 24)`. This can happen from the command line by running `zed [path]`, or in editor by attempting to open a file from the tree view or from `Go to file`.

Example:

```
> zed ~/Library/Logs/Zed/Zed.log
error opening PathWithPosition { path: "/Users/alexandrekirszenberg/Library/Logs/Zed/Zed.log", row: None, column: None }: Too many open files (os error 24)
```

I work on a very large repository, so I'm not surprised to hit this error as the default macOS opened file descriptors limit is 256, but it seems counter-intuitive that I'd be the first to report this issue as this is a very common setup. So maybe something else is going on.

For now, I'll try increasing this with the following:
```
launchctl limit maxfiles 4096 unlimited
```

and report back.

### Environment

Zed: v0.162.0 (Zed Preview)
OS: macOS 15.1.0
Memory: 32 GiB
Architecture: aarch64

```
> ulimit -n
256
> launchctl limit maxfiles
	maxfiles    256      unlimited
> sysctl kern.maxfiles
kern.maxfiles: 2147483647
```

### If applicable, attach your Zed.log file to this issue.

<details><summary>Zed.log</summary>

```

```
</details>

TimDiekmann:
I encounter the same issue (quite frequently recently). My `ulimit` is `2560` (I did not change that) and I encounter the issue even with only 20 files open.
I cannot tell exactly when this happens but I don't think that 20 open files should be the issue. Worth mentioning is that `lsof` reports more than 2560 open files, though.

notpeter:
The default soft-limit `ulimit -Sn` is `256` on macOS, but the default hard-limit `ulimit -Hn` is `unlimited`.  Interestingly, my `sysctl kern.maxfiles` is much lower than yours (491520 vs 2147483647), I wonder if that's because I'm still on Sonoma (macOS 14.7)

On macOS I don't know of an easy way to introspect the limits for a given running process like `/proc/$PID/limits` on Linux. 

1. I'd be curious what was happening prior to it throwing that error.  Could you share some log lines before this triggers?
2. Can you share some details about your project: language(s), language-servers, relative size, number of files, source control (git), etc.
[...]

PR #57094, opened 2026-05-18
terminal: Stop caching FDs for every system process per PTY

tredondo:
Generated by Claude Opus 4.7 max. Not tested yet.

------

`PtyProcessInfo::new` initializes its `sysinfo::System` via
`System::new_with_specifics`, which performs a `ProcessesToUpdate::All`
refresh on construction. Combined with `ProcessRefreshKind::nothing()`
defaulting to `tasks: true` (a sysinfo gotcha — see [the
docstring](https://docs.rs/sysinfo/0.37.2/sysinfo/struct.ProcessRefreshKind.html#impl-Default-for-ProcessRefreshKind)),
sysinfo opens and caches a file descriptor for every `/proc/<pid>/stat`
and every `/proc/<pid>/task/<tid>/stat` on the host, up to its internal
budget of `RLIMIT_NOFILE / 2`.

The `System` is never consulted with `ProcessesToUpdate::All` after
construction; only the PTY's own PID is refreshed via
`refresh_processes_specifics(Some(&[pid]), ...)`. The system-wide scan
is pure waste, but the cached FDs persist for the lifetime of the PTY,
and **each new PTY creates an independent `System` with its own cache**.

In an agent-heavy session that opens many real PTYs — every
`Project::create_terminal_task` / `create_terminal_shell` call — this
accumulates hundreds of thousands of stat FDs. Once sysinfo's budget is
hit, subsequent `open()`s on `/proc/<pid>/stat` silently fail (see
`FileCounter::new` in `sysinfo/src/unix/linux/process.rs`), and the
editor reports "Too many open files" when spawning subprocesses.

The other three `sysinfo::System` call sites in Zed already call
`.without_tasks()` (`lsp_button.rs:94`, `attach_modal.rs:390`,
`headless_project.rs:1226`); `pty_info.rs` is the one that didn't.

This change:

1. Adds `.without_tasks()` to `process_refresh_kind`, eliminating the
   per-thread FD cache.
2. Replaces `System::new_with_specifics(...)` with `System::new()` plus
   a targeted refresh of just the PTY's PID, so a single `System`
   caches one FD instead of thousands.

### Empirical impact

In a <2h session on Linux that hit this bug, the editor was holding
**263,371 FDs**, of which **262,144 (99.5%)** were `/proc/<pid>/stat`
and `/proc/<pid>/task/<tid>/stat` — exactly sysinfo's
`RLIMIT_NOFILE / 2` cap. 696 of the 1,940 tracked PIDs were already
dead. The same PID was opened ~195 times across separate `System`
instances (one per PTY ever created in the session).

After this change, each `PtyProcessInfo` opens 1 FD (for its target
PID) instead of approximately `n_processes + n_threads_across_them`.

### Refs

- #49776 — exact same `/proc/[PID]/stat` accumulation pattern, Linux
- #47064 / #20806 — related macOS variant fixed by #47322
  (`MacWatcher` → `notify::FsEventWatcher`); same shape of bug
  ("per-resource subsystem creates a thing that holds many FDs"),
  different leaking subsystem
- sysinfo
  [#242](https://github.com/GuillaumeGomez/sysinfo/issues/242) —
  upstream confirmation that the FD-caching behavior is intentional;
  `set_open_files_limit` is the provided knob

Investigation write-up: [zed-industries/zed#49776 (comment).](https://github.com/zed-industries/zed/issues/49776#issuecomment-4481612281).

Release Notes:

- Fixed a file descriptor leak on Linux where opening a terminal would
  cause Zed to hold open `/proc/<pid>/stat` for every process on the
  system, eventually breaking subprocess spawning after long sessions

### DIFF
--- a/crates/terminal/src/pty_info.rs
+++ b/crates/terminal/src/pty_info.rs
@@ -10,7 +10,7 @@ use std::{path::PathBuf, sync::Arc};
 #[cfg(target_os = "windows")]
 use windows::Win32::{Foundation::HANDLE, System::Threading::GetProcessId};
 
-use sysinfo::{Pid, Process, ProcessRefreshKind, RefreshKind, System, UpdateKind};
+use sysinfo::{Pid, Process, ProcessRefreshKind, ProcessesToUpdate, System, UpdateKind};
 
 use crate::{Event, Terminal};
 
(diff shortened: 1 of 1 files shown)


cole-miller:
Thanks! It looks like the underlying issue was fixed in #58683, so I'm going to close this, but feel free to open a new PR if there are additional issues here!

